PR #20175, opened 2024-04-22
chore: make document service api more consistent

alexandrebodin:
### What does it do?

chore: make document service api more consistent

### Why is it needed?

v5

### DIFF
--- a/packages/core/content-manager/server/src/services/document-manager.ts
+++ b/packages/core/content-manager/server/src/services/document-manager.ts
@@ -138,7 +138,7 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
       return strapi
         .documents(uid)
         .clone({ ...params, documentId: id })
-        .then((result) => result?.versions.at(0));
+        .then((result) => result?.entries.at(0));
     },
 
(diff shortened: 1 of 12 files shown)


Marc-Roig (approved):
nice! 🕺 

this change might affect content history middleware too

alexandrebodin:
@Marc-Roig I'm wondering if we should return entries on update & create too at some point (when/if we allow creating multiple locales at once)

Marc-Roig:
do we plan to allow it? I find the scenario of updating multiple locales with the same values a bit rare. And probably non localized fields might cover many of those use cases.

It's true that it would standarize the response between the other methods, but at the same time it would make the migration a bit more complex. So I don't know either 🤔

alexandrebodin:
> do we plan to allow it? I find the scenario of updating multiple locales with the same values a bit rare. And probably non localized fields might cover many of those use cases.
> 
> It's true that it would standarize the response between the other methods, but at the same time it would make the migration a bit more complex. So I don't know either 🤔

For now I don't see a usecase either, just not a fan of the response differences even if a bit more consistent with this PR

Issue #19660, opened 2024-03-04
TypeError: Cannot read properties of undefined (reading 'length') after upgrade to 4.20.3


vsamofal:
### Required System information

STRAPI_ENFORCE_SOURCEMAPS=true in your .env --> (sorry I tried it, but it's not working)

- Node.js version: v18.14.2
- NPM version: 9.5.0
- Strapi version: 4.20.3
- Database: postgres
- Operating system: darwin-arm64
- Is your project Javascript or Typescript: js

### Describe the bug

After to upgrade to latest version, frontend is failing with exception on reading a length property, on one particular collection, and lead to white screen in both dev and prod modes. Refreshing the page helps, and it loading entity and allows me to add/edit and do everything that I want. But once I go to another collection and moving back to this entity - white screen again. Refresh and it's working just fine. 

If I'm switching menu and getting back to content manager, I can open this page fine, so the problem appears only when I'm switching between collections. 

If I'm switching to single types, I can back to collection just fine as well. So again the problem appears only 

### Expected behavior

Switching between collections should work fine 

### Code snippets

Sorry for this... but I can't make source map works

```js
main.0f63a8e2.js:805 TypeError: Cannot read properties of undefined (reading 'length')
```

Error happens here, f is undefined
![image](https://github.com/strapi/strapi/assets/27850115/eb7878ea-2945-434b-9d36-035a941826ab)

```json
field definition

  "faq": {
      "displayName": "faq",
      "type": "component",
      "repeatable": true,
      "component": "general.faq"
    }
```

but when I'm refreshing the page, this field is there, seems like some async processing happen: 

![image](https://github.com/strapi/strapi/assets/27850115/b7fce1d1-7561-4d54-9fc9-144f9ce5e9a2)

hopefully it helpful and enough, if not let me know what else to share

brechtpallemans:
Same issue here, I'm on version 4.20.2

coderdix24:
same issue

vsamofal:
4.21.0 still here

vsamofal:
Found that it happen also with a backend error, where limit passed to the query is undefined 
![image](https://github.com/strapi/strapi/assets/27850115/3041e6ab-0c84-45b6-adac-d5a9082a9c83)

ioanaasurdu:
Any updates on this?

anurodhsahu:
I was also facing the same issue. Deleting and re-creating the collection worked for me.

anil1712:
[...]

PR #20180, opened 2024-04-23
fix: content could be undefined

alexandrebodin:
### What does it do?

content can be undefined because of some timing issues. the Cell is rendered before the new data is loaded when navigating between content types in the CM list views. 

There is a depper issue to fix for v5 but it it's still valid to check the value is defined and is an array before accessing that length prop regardless

### Why is it needed?

Fix a bug

### How to test it?

Navigating in the CM between content-types with repeatable components displayed in the list view should not reproduce the error anymore

### Related issue(s)/PR(s)

Fix https://github.com/strapi/strapi/issues/19660


joshuaellis (approved):
does this issue exist on v5? I'd hope it doesnt since in V4 it looks like race conditions with data-fetching.

PR #20161, opened 2024-04-19
chore: migrate bulkDelete to v5

Feranchz:
### What does it do?

Migrates BulkDelete to works with v5 and enabled some API tests related to BulkDelete

PR #20129, opened 2024-04-16
feat: webhooks

Marc-Roig:
### What does it do?

Trigger webhooks on document service.

Only if using the document service, the webhooks will be triggered. So, using the query engine will not trigger them.

A schema of what triggers what:

image

PR #20184, opened 2024-04-23
chore: migrate bulk publish & unpublish to v5

Feranchz:
### What does it do?

Migrates Bulk Publish & unpublish to v5

### DIFF
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -411,7 +411,7 @@ export default {
     const { userAbility } = ctx.state;
     const { model } = ctx.params;
     const { body } = ctx.request;
-    const { ids } = body;
+    const { documentIds } = body;
 
     await validateBulkActionInput(body);
 
@@ -422,28 +422,27 @@ export default {
       return ctx.forbidden();
     }
 
-    const permissionQuery = await permissionChecker.sanitizedQuery.publish(ctx.query);
-    const populate = await getService('populate-builder')(model)
-      .populateFromQuery(permissionQuery)
-      .populateDeep(Infinity)
-      .countRelations()
-      .build();
+    const { locale } = getDocumentLocaleAndStatus(body);
 
-    const entityPromises = ids.map((id: any) => documentManager.findOne(id, model, { populate }));
-    const entities = await Promise.all(entityPromises);
+    const documentLocales = await documentManager.findLocales(documentIds, model, {
+      locale,
+    });
 
-    for (const entity of entities) {
-      if (!entity) {
-        return ctx.notFound();
-      }
+    if (documentLocales.length === 0) {
+      return ctx.notFound();
+    }
 
-      if (permissionChecker.cannot.publish(entity)) {
+    for (const document of documentLocales) {
+      if (permissionChecker.cannot.publish(document)) {
         return ctx.forbidden();
       }
     }
 
-    // @ts-expect-error - publish many should not return null
-    const { count } = await documentManager.publishMany(entities, model);
+    const localeDocumentsIds = documentLocales
+      .filter((document) => !document.publishedAt)
+      .map((document) => document.documentId);
+
+    const { count } = await documentManager.publishMany(localeDocumentsIds, model, { locale });
     ctx.body = { count };
   },
 
@@ -451,7 +450,7 @@ export default {
     const { userAbility } = ctx.state;
     const { model } = ctx.params;
     const { body } = ctx.request;
-    const { ids } = body;
+    const { documentIds } = body;
 
     await validateBulkActionInput(body);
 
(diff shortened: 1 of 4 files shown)


packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -422,36 +422,35 @@ export default {
...
-    const populate = await getService('populate-builder')(model)
-      .populateFromQuery(permissionQuery)
-      .populateDeep(Infinity)
-      .countRelations()
-      .build();
+    const { locale } = getDocumentLocaleAndStatus(body);
 
-    const entityPromises = ids.map((id: any) => documentManager.findOne(id, model, { populate }));
-    const entities = await Promise.all(entityPromises);
+    const documentLocales = await documentManager.findLocales(documentIds, model, {
```
jhoward1994, line 427: Can you call findLocales this way? I dont think it currently supports multiple documentIds

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -196,85 +196,30 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
...
-      // Everything is valid, publish
-      const publishedEntitiesCount = await strapi.db.query(uid).updateMany({
-        where: filters,
-        data,
-      });
-      // Get the updated entities since updateMany only returns the count
-      const publishedEntities = await strapi.db.query(uid).findMany({
-        where: filters,
-        populate,
+    async publishMany(
```
jhoward1994, line 199: I think we'd be losing the event emission with this change?

Also FYI (I think I can deal with the conflict) with bulk locale actions this is
something like:

`async publishMany(uid: UID.ContentType, documentIds: string[], locale?: string| string[])`

e.g. it can take multiple locales. You can see it here

https://github.com/strapi/strapi/blob/65c408613ecd61094a501a8537ed6c87e1498793/packages/core/content-manager/server/src/services/document-manager.ts#L200

packages/core/content-manager/server/src/controllers/collection-types.ts
Feranchz: It works, but the function's type are not expecting an array, on the [bulkDelete PR](https://github.com/strapi/strapi/pull/20161/files#diff-6d03b8fb4e523d8c3333b4225c5c685874884b96e4b0d35f8f4e878671b32087) I added, I will do the same here

packages/core/content-manager/server/src/services/document-manager.ts
Feranchz: webhooks events are going to be triggered at document service level, Marc is currently working on that [here](https://github.com/strapi/strapi/pull/20129)

jhoward1994: Thanks, I think I can remove that from my PR too then

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -422,36 +422,35 @@ export default {
...
+    if (documentLocales.length === 0) {
+      return ctx.notFound();
+    }
 
-      if (permissionChecker.cannot.publish(entity)) {
+    for (const document of documentLocales) {
+      if (permissionChecker.cannot.publish(document)) {
         return ctx.forbidden();
       }
     }
```
Marc-Roig, line 440: I think the .cannot.publish check should only be done with the draft versions and I believe you might have draft and published versions here right?

Marc-Roig: you might as well not query those one when looking for the find locales 🤔

Marc-Roig: and same for the unpublish action, you should only check published documents

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -196,85 +178,30 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
...
-      const unpublishedEntities = await strapi.db.query(uid).findMany({
-        where: filters,
-        populate,
+    async unpublishMany(
+      documentIds: Modules.Documents.ID[],
+      uid: UID.CollectionType,
+      opts: Omit<DocServiceParams<'unpublish'>, 'documentId'> = {} as any
+    ) {
+      const unpublishedEntries = await strapi.db.transaction(async () => {
+        return Promise.all(documentIds.map((id) => this.unpublish(id, uid, opts)));
```
Marc-Roig, line 200: you might as well call the document service unpublish directly, it will save a bit of time by not having a deep populate

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -50,11 +30,12 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
      * Find multiple (or all) locales for a document
      */
     async findLocales(
-      id: string | undefined,
+      id: string | string[] | undefined,
       uid: UID.CollectionType,
       opts: {
         populate?: Modules.Documents.Params.Pick<any, 'populate'>;
         locale?: string | string[] | '*';
+        isPublished?: boolean;
```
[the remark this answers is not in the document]
Marc-Roig: This method was originally intended to get the locales of a single document, and you are right that that name doesn't fit it anymore.

Any of the two options seems fitting to me, If I had to choose I would directly use a db query on the controller.

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -191,85 +173,44 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
...
+        );
       });
-      // Emit the publish event for all updated entities
-      await Promise.all(
-        publishedEntities!.map((doc: Document) => emitEvent(uid, ENTRY_PUBLISH, doc))
-      );
 
       // Return the number of published entities
-      return publishedEntitiesCount;
+      return { count: publishedEntries.length };
```
jhoward1994, line 193: I am also using this for publishing multiple locales. I think the way you have
this written we would only get the count of published documents and not the
count of published entities? (as you're doing .at(0))

Would it be okay if we did something like

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -422,36 +422,42 @@ export default {
...
+    }
+
+    const documentLocales = await strapi.db.query(model).findMany({
+      where,
+    });
 
-      if (permissionChecker.cannot.publish(entity)) {
+    if (documentLocales.length === 0) {
+      return ctx.notFound();
+    }
```
jhoward1994, line 442: I think here we can use the

Marc-Roig (approved):

jhoward1994 (approved):
Thanks, looks good!

PR #20185, opened 2024-04-23
fix(content-manager): history service

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -167,7 +167,12 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
             : { documentId: context.params.documentId, locale: context.params?.locale };
 
         const defaultLocale = await getDefaultLocale();
-        const locale = documentContext.locale || defaultLocale;
+
+        // TODO
```
jhoward1994, line 171: Hey @markkaylor 

I was having trouble with content history during the bulk locale publishing stuff I'm working on. This bit of code would cause an error to be thrown from

https://github.com/strapi/strapi/blob/b9f98c5dd7d398642b83d37a9024084add4c0af3/packages/core/core/src/services/document-service/internationalization.ts#L47

I'm not 100% what the best solution is here. I think the problem is that now publishing can accept multiple locales, history will intercept those calls and try and proceed with the logic below using an array of locales from the params. docService.findOne then throws as this is an invalid param.

What do you think the best idea here is?

markkaylor: Maybe we can just check if it's an array and skip it for now 🤷‍♂️ ? Then we can circle back to adjust the logic here to handle all bulk actions. What do you think? cc @remidej

remidej: I think I agree with you @markkaylor. this is a place where we need to eventually write proper tests, and we should probably be the ones to write it, so I think opting out for now in this branch should be okay

jhoward1994:
@remidej @markkaylor 

Is it okay with you both if I merge this into our feature branch and you could track this TODO in the content studio team?

I need this change to test the whole feature from the FE

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -167,7 +167,18 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+        if (documentContext.locale) {
+          if (Array.isArray(documentContext.locale)) {
+            // TODO calls picked from the middleware could contain an array of
+            // locales. This is incompatible with our call to findOne below.
+
+            return next();
+          }
+
+          locale = documentContext.locale;
+        }
```
markkaylor, line 181: Could we avoid creating a mutable variable for locale? Wouldn't this be enough?

```suggestion
if (Array.isArray(documentContext.locale)) {
  // TODO calls picked from the middleware could contain an array of
  // locales. This is incompatible with our call to findOne below.

  return next();
}

const locale = documentContext.locale || defaultLocale;
```

markkaylor (approved):

PR #20192, opened 2024-04-24
chore: update actions deps

alexandrebodin:
### What does it do?

chore: update actions deps

### Why is it needed?

updating major version deps when possible

### How to test it?

if the CI passes we are all good

### Related issue(s)/PR(s)


Convly (commented):
The changes LGTM. Some e2e tests are failing but I don't think it comes from your PR.

alexandrebodin:
Yeah I'll merge in the deps branch as It's not sth I can fix here :)

PR #20179, opened 2024-04-23
chore(history): add api tests

markkaylor:
### What does it do?

- Refactors restore relations, and restore media into their own functions
- Moves all history util functions to their own util factory
- Adds api tests for restoring a history version with missing relations or media

### Why is it needed?

- To clean up the code
- To have more test coverage

### How to test it?

- Everything should behave as expected
- The api tests should pass

TODO:

test media

PR #20199, opened 2024-04-24
fix(tests): update history api tests to created ids

markkaylor:
### What does it do?

Use the created versions to restore


tests/api/core/content-manager/content-manager/history/history.test.api.ts
```diff
@@ -177,7 +178,7 @@ describeOnCondition(edition === 'EE')('History API', () => {
           description: 'Hello',
         },
       }),
-      updateEntry({
+      await updateEntry({
```
remidej, line 181: await doesn't seem needed

markkaylor: good catch I'll remove it in https://github.com/strapi/strapi/pull/20179

PR #20201, opened 2024-04-24
feat: show hidden fields in history frontend

remidej:
### What does it do?

Makes hidden fields visible in the History page. This is not about internal fields like id, but about the ones the user may have deleted from the layout via the configure the view page. I called them "remaining fields" in the code, to match what was already done in the code for that page.

It applies the fix in two places:

- At the root of the history page layout. This one was the most straight forward.
- Inside the components. We didn't have direct access to the component's rendering logic, so I had to refactor it to pass a new `renderLayout`. It has the benefit of avoiding duplicating logic between repeatable and non-repeatable components, and avoids some prop drilling for `renderInput`.

I want to write e2e tests for this, they have issues we need to fix first.

### Why is it needed?

Because even though they're not visible in the edit view, the fields that aren't in the layout would still be restored if the history version is restored. It's therefore important to show them regardless, so that users know what they're restoring.

### How to test it?

Go to the configure the view page for a content type. Delete a root field there (not in the CTB), as well as a field that's inside a component.

Now open the history page for an entry of that content type. You should see the deleted root field on a new line at the bottom of the page, but above the unknown fields.

Also check that component inputs work as expected in the Edit View page, since their code was refactored.


packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -328,7 +333,25 @@ const VersionInputRenderer = ({
           {...props}
           hint={hint}
           disabled={fieldIsDisabled}
-          renderInput={(props) => <VersionInputRenderer {...props} shouldIgnoreRBAC={true} />}
+          renderLayout={(layoutProps) => {
+            // Components can only have one panel, so only save the first layout item
+            const [remainingFieldsLayout] = getRemaingFieldsLayout({
+              layout: [layoutProps.layout],
```
markkaylor, line 339: The use of layoutProps.layout here feels circular. Isn't this the layout you are about to pass to layoutProps?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
```diff
@@ -10,17 +10,17 @@ import { EditFieldLayout } from '../../../../../hooks/useDocumentLayout';
...
+import { ComponentLayout, type ComponentLayoutProps } from './Layout';
 import { NonRepeatableComponent } from './NonRepeatable';
 import { RepeatableComponent } from './Repeatable';
 
 interface ComponentInputProps
   extends Omit<Extract<EditFieldLayout, { type: 'component' }>, 'size' | 'hint'>,
     Pick<InputProps, 'hint'> {
   labelAction?: React.ReactNode;
-  renderInput?: (props: InputRendererProps) => React.ReactNode;
+  renderLayout: (props: ComponentLayoutProps) => React.ReactNode;
```
markkaylor, line 23: I think it's still optional with a default provided...right?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
```diff
@@ -30,6 +30,7 @@ const ComponentInput = ({
   attribute,
   disabled,
   labelAction,
+  renderLayout = ComponentLayout,
```
markkaylor, line 33: I think you should be setting the default at the same level it is being used and it can be undefined until then

joshuaellis (commented):
If you need to change the entire layout, which is basically the child of the strapi component field, doesn't it just make more sense to enable `children` to be passed? We'd then export the component _without_ children applied, it could be moved up to a more generic domain e.g. `components` and in the EditView we add the current layout and in history we add your revised layout requirements? 

How did we come to the conclusion this was the right solution? I'm just trying to weigh up the right direction in the long term... 🤔

markkaylor (commented):
I had this error:

Screenshot 2024-04-25 at 12 13 03

I'm not sure exactly how it would work with you remaining fields function but I'm curious what you think of using children instead of two render props. I believe this is working and accomplishes the same thing: https://github.com/strapi/strapi/pull/20207/files

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
remidej: I don't think I see the circular aspect here. I'm only passing a layout to the util so that it can know which fields are _not_ in that layout (e.g. the remaining ones)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
remidej: sure I can change it, I just wonder why you prefer that?

markkaylor: My previous comment is maybe no good. I had an error yesterday when running this branch and I believe it was related to this being undefined. So I guess I mean just make sure it's defined where needed :)

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
markkaylor: Isn't `remainingFieldsLayout` what you pass to the `layout` prop on `ComponentLayout` and then therefore the argument you receive in `renderLayout`?

remidej: Not sure I 100% understand but it is normal if there is some sort of recursion here, the layout renders fields, some fields are components, therefore they render their own layout, etc

markkaylor: 👌

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -323,13 +332,23 @@ const VersionInputRenderer = ({
...
       return (
-        <ComponentInput
-          {...props}
-          hint={hint}
-          disabled={fieldIsDisabled}
-          renderInput={(props) => <VersionInputRenderer {...props} shouldIgnoreRBAC={true} />}
-        />
+        <ComponentInput {...props} hint={hint} disabled={fieldIsDisabled}>
+          {(name) => (
+            <ComponentLayout name={name} layout={[...layout, ...remainingFieldsLayout]}>
```
markkaylor, line 347: Why does ComponentLayout need to be a function receiving name instead of a component receiving name? Isn't it receiving name via the props passed to ComponentInput? Couldn't you just pass it directly here: `<ComponentLayout name={props.name} ... />`

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Repeatable.tsx
```diff
@@ -211,6 +207,8 @@ const RepeatableComponent = ({
       <AccordionGroup error={error}>
         <AccordionContent aria-describedby={ariaDescriptionId}>
           {value.map(({ __temp_key__: key, id }, index) => {
+            const nameWithIndex = `${name}.${index}`;
+            const safeKey = key ?? nameWithIndex;
```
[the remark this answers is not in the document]
joshuaellis: I'd like to avoid creating keys like this if possible, we had this in V4 and it got very messy and quite buggy to follow, especially if it leaks into the EditView, the re-ordering will crash.

We create tmp keys when we fetch the document: 

https://github.com/strapi/strapi/blob/4300e8f16c86f11fe187f56b72eabee2f5d739b9/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx#L122

https://github.com/strapi/strapi/blob/4300e8f16c86f11fe187f56b72eabee2f5d739b9/packages/core/content-manager/admin/src/pages/EditView/utils/data.ts#L127

You could use the same pipe funcs perhaps? or maybe just the tmp key one.

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
joshuaellis: Do we need to re-export the Component layout? it's basically just a grid and instead of trying to be DRY it might be a lot clearer to just re-do the rendering logic here with the raw grid components, then you're only concerened with the inputs?

joshuaellis (commented):
I think you'd benefit from something similar to what we've done in the edit view - `useDocumentHistoryLayout`, you can basically isolate all the building of your layout to one space, it'll be neat and concise, you can call the hooks you need to (remember we have caching now so you can recall the same hook multiple times and it wont make more requests) then your react components are super dumb and just render what you give them. You could have a separate object for the components that are basically ready to "just use" to get their layouts too, like we do in the EditView, it'll avoid recalculating at different steps.

Let me know what you think, i'm happy to work with you on this if it's something you're interested in but in clear how it'd look :)

i.e. i think we're struggling to follow the flow of how it all comes together, these layouts are complicated, to be clear its not criticism of your ability 😄

remidej:
@joshuaellis I think I see what you're saying, but with the current state solution I think I'm happy where things stand:

- HistoryPage fetches the layout data. I like that because that component centralizes all the loading and error states, so we have one single Loader
- VersionContent transforms the layout and renders it. It's the only purpose of the component so I feel like it's pretty readable
- VersionInputRenderer does a little big of logic to compute the layout for components

For components I agree it could be nice to not have some layout logic in the input renderer for components. But I feel like extracting things to a hook will add more complexity rather than simplify things 🤔

markkaylor (commented):
Thanks for trying several approaches to find what will work best for us 🙏 . I like what you have here. 

- I have this error when trying to access history on a content type with a component
Screenshot 2024-04-30 at 13 34 23
- I find it a bit strange that hidden fields don't have a section title like unknown fields. Isn't it a bit confusing for the user to just see some fields in a different panel? Maybe we should ask Lucas?

remidej:
@markkaylor thanks for bug report! I was only checking with content types that had hidden fields.

About the hidden fields section. I thought about that but I'm not sure it's a good idea. I think it's important that unkown fields remain clearly different from the rest, because they're the ones that actually behave differently, as they can't be restored. If we have three different sections, we lose a bit of the clear page distinction between restorable and non-restorable. So I'm afraid trying to over-explain things by would actually create confusion.

There's also the issue of hidden fields inside of components. You can't extract them into their own section of the page, since they would lose their context. So you'd have to split each component into a known and an unknown section. It's not something I'm excited about.

We can also check with Lucas when he's back, but he's OOO until the Barcelona

markkaylor (commented):
> There's also the issue of hidden fields inside of components. You can't extract them into their own section of the page, since they would lose their context. So you'd have to split each component into a known and an unknown section. It's not something I'm excited about

I think I see what you mean here and that does sound like a bummer. I noticed if I hide a field in a component via configure the view, then in the EditView it's hidden but in the history version it's just still there in the component.  On the other hand if I hid the entire component via configure the view it shows up in the panel with the other hidden fields as expected. I guess that's the kind of thing you are talking about here and that's tomorrow's problem?

- I found one last error when finishing my QA when trying to access history of a content type that does not have a component:
Screenshot 2024-04-30 at 16 08 01

Besides that I'm good to approve 👍

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
```diff
@@ -20,7 +20,8 @@ interface ComponentInputProps
   extends Omit<Extract<EditFieldLayout, { type: 'component' }>, 'size' | 'hint'>,
     Pick<InputProps, 'hint'> {
   labelAction?: React.ReactNode;
-  renderInput?: (props: InputRendererProps) => React.ReactNode;
+  children: (name: InputRendererProps) => React.ReactNode;
+  layout: EditFieldLayout[][];
```
joshuaellis, line 24: Just because it's not clear from *only* looking at this file why we pass layout instead of finding it here, can we just write why we've done it? i.e. `to enable content history to do XYZ`

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
```diff
@@ -20,7 +20,8 @@ interface ComponentInputProps
   extends Omit<Extract<EditFieldLayout, { type: 'component' }>, 'size' | 'hint'>,
     Pick<InputProps, 'hint'> {
   labelAction?: React.ReactNode;
-  renderInput?: (props: InputRendererProps) => React.ReactNode;
+  children: (name: InputRendererProps) => React.ReactNode;
```
joshuaellis, line 23:
```suggestion
  children: (props: InputRendererProps) => React.ReactNode;
```

packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
```diff
@@ -10,6 +10,7 @@ import type { Schema, UID } from '@strapi/types';
  * traverseData
  * -----------------------------------------------------------------------------------------------*/
 
+type PartialSchema = Partial<Schema.Schema> & Pick<Schema.Schema, 'attributes'>;
```
joshuaellis, line 13: Worth a comment to explain why we want to support a Partial Schema? I assume because you don't store the whole schema in CH? 🤔

joshuaellis (commented):
Great job sticking with it to find the best API we can for the component stuff 👏🏼 I've not QAd but code makes sense ⭐

packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
remidej: Yes CH only saves attributes, and only attributes are used in those functions

remidej:
> I found one last error when finishing my QA when trying to access history of a content type that does not have a component:

@markkaylor I can't reproduce this, could you tell me which content type you tested with? I think tried content types with and without components, with and without data in them, with and without hidden fields

>I noticed if I hide a field in a component via configure the view, then in the EditView it's hidden but in the history version it's just still there in the component. On the other hand if I hid the entire component via configure the view it shows up in the panel with the other hidden fields as expected. I guess that's the kind of thing you are talking about here and that's tomorrow's problem?

I believe both cases are what we want, what is the issue you're seeing? Seeing a hidden field within a component is the reason I'm doing this whole refactoring. And for the hidden component there's no other choice but to show it at the bottom

markkaylor:
So for components, I can hide a component and it works as expected:

https://github.com/strapi/strapi/assets/26598053/89bf46f4-c15d-4216-8d02-f0fe81e09fe5

But if I try and hide a field inside a component, then in the EditView it's hidden but in history it is not represented as a hidden field.

https://github.com/strapi/strapi/assets/26598053/3a037999-3ddd-4a4e-9429-74db40a44c14

Let me know if it's clear or if you prefer to take a look on discord or something.

For the error, I had it on the Tag content type but I am unable to reproduce now too. Maybe it was stale data or something 🤷‍♂️

PR #20203, opened 2024-04-25
fix(scheduling): update error message

madhurisandbhor:
### What does it do?

Updates the scheduling required field error message to use more generic required field message.

### Why is it needed?

Asked by UX, https://www.notion.so/strapi/Releases-Scheduling-Alpha-Blitz-Session-14ae56786d5a41e2a0dbf98b7e4b55e2?pvs=4#8916a056719f4dbb862c479bf1eab087

### How to test it?

When user tries to schedule a release if any of the required field is not filled, on submit user should get error message `This value is required.`


packages/core/content-releases/shared/validation-schemas.ts
```diff
@@ -3,22 +3,22 @@ import * as yup from 'yup';
 export const RELEASE_SCHEMA = yup
   .object()
   .shape({
-    name: yup.string().trim().required(),
+    name: yup.string().trim().required('Required field'),
```
joshuaellis, line 6: we have generic translations for this sort of thing coming from the helper-plugin:

https://github.com/strapi/strapi/blob/a76a2ee7a9a66bc0233e678ca062086c34855a5c/packages/core/helper-plugin/src/utils/translatedErrors.ts#L10

madhurisandbhor:
I am closing this PR to create changes on V5 instead as importing from helper plugin will need to refactor again.

PR #20204, opened 2024-04-25
Update deps

alexandrebodin:
### What does it do?

Update dependencies major versions when possible


Convly:
Tested the gql one, seems like it's working well

packages/core/strapi/package.json
```diff
@@ -181,10 +176,6 @@
   },
   "devDependencies": {
     "@strapi/ts-zen": "^0.2.0",
-    "@types/bcryptjs": "2.4.3",
-    "@types/configstore": "5.0.1",
-    "@types/delegates": "1.0.0",
-    "@types/find-root": "1.1.4",
```
remidej, line 187: were configstore, delegate and bcrypt just forgotten when they were removed?

packages/core/data-transfer/package.json
```diff
@@ -61,7 +61,7 @@
   "devDependencies": {
     "@strapi/database": "workspace:*",
     "@strapi/pack-up": "5.0.0",
-    "@types/fs-extra": "9.0.13",
+    "@types/fs-extra": "11.0.4",
```
remidej, line 64: Other packages use fs-extra but don't have the @types dependency, should we add it?

packages/plugins/graphql/package.json
```diff
@@ -50,8 +50,8 @@
   "dependencies": {
     "@apollo/server": "4.10.0",
     "@as-integrations/koa": "1.1.1",
-    "@graphql-tools/schema": "8.5.1",
-    "@graphql-tools/utils": "^8.13.1",
+    "@graphql-tools/schema": "10.0.3",
+    "@graphql-tools/utils": "^10.1.3",
```
remidej, line 54: More general question: how to you pick between using a pinned version or a range?

packages/core/strapi/package.json
alexandrebodin: we moved most of the code into the core package and they didn't get cleaned up :)

packages/core/data-transfer/package.json
alexandrebodin: Yeah good idea I'll add those

Marc-Roig (approved):

packages/plugins/graphql/package.json
unrevised6419: Hey folks! I encountered an issue, and arrived here 😄 

It seems this needs to be `~`.
In a repo I have, graphql codegen fails because `@graphql-tools/schema` resolves to this fixed version that has bugs fixed in newer patches. Current version is 10.0.31

unrevised6419: Why in the first place was decided to have the version fixed instead of ~ or ^ ? 👀

PR #20206, opened 2024-04-25
chore: upgrade mailgun provider and remove deprecated options

alexandrebodin:
### What does it do?

Upgrade mailgun provider and remove deprecated options


innerdvations (approved):
LGTM, should this be documented as a breaking change?

alexandrebodin:
We can definitely add sth as we didn't log the deprecation bfore

PR #20208, opened 2024-04-25
fix(scheduling): update error message

madhurisandbhor:
### What does it do?

Updates the scheduling required field error message to use more generic required field message.

### Why is it needed?

Asked by UX, https://www.notion.so/strapi/Releases-Scheduling-Alpha-Blitz-Session-14ae56786d5a41e2a0dbf98b7e4b55e2?pvs=4#8916a056719f4dbb862c479bf1eab087

### How to test it?

When user tries to schedule a release if any of the required field is not filled, on submit user should get error message This value is required.


Marc-Roig (approved):

packages/core/content-releases/admin/src/validation/schemas.ts
remidej, line 1: I guess we're moving the validations because the backend wasn't using them?

Feranchz: The backend use some of them but with different fields/formats, also in the front we want to throw translated errors and we can't do that on back so one for both don't work anymore

PR #20212, opened 2024-04-25
fix(history): handle sanitization when fetching versions

markkaylor:
### What does it do?

- Adjust history sanitization so it doesn't break the e2e tests
- Build media and relations object in their own dedicated functions

### How to test?

- The e2e tests for history should pass (locally I have them passing as before they were broken, here in the CI they are sometimes flaky but I think it is unrelated)
- History versions and their relations should not return sensitive data
- Verify everything is working as normal (missing relations, restoration, etc...)
- Admin user relations should not be completely sanitized out of the response


packages/core/content-manager/server/src/history/controllers/history-version.ts
```diff
@@ -68,13 +68,28 @@ const createHistoryVersionController = ({ strapi }: { strapi: Core.Strapi }) =>
         ...getValidPagination({ page: params.page, pageSize: params.pageSize }),
       });
 
+      const sanitizedResults = await async.map(results, async (version: any) => {
```
remidej, line 71: Why do we need any? results does have a type, I'm surprised it's not inferred

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -354,9 +354,43 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                      if (userToPopulate == null) {
+                        return null;
+                      }
+
+                      return strapi
+                        .query('admin::user')
+                        .findOne({ where: { id: userToPopulate.id } });
+                    })
+                  )
+                ).filter((user) => user != null),
```
remidej, line 386: I moved this filtering logic from frontend to backend

packages/core/content-manager/server/src/history/controllers/history-version.ts
markkaylor: I'm not sure maybe @Marc-Roig knows how the inference/typings could work with the async.map

Marc-Roig: We are using a curried version of pMap, for some reason the lodash/fp `curry` function doesn't work too well on types :sad-pepe:

markkaylor: Bummer, well I can at least replace the any here with a type from the contract for `version`.

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -349,24 +330,115 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                  ...(await currentDataWithRelations),
+                  /**
+                   * TODO:
+                   *
+                   * Ideally we would return the same "{results: [], meta: {}}" shape, however,
+                   * when sanitizing the data as a whole in the controller before sending to the client,
+                   * the data for admin relation user is completely sanitized if we return an object here as opposed to an array.
+                   */
+                  [attributeKey]: sanitizedAdminUser,
+                };
```
markkaylor, line 436: ℹ️ For now I think this workaround is acceptable to fix the issue and get the tests passing. 

I'd like to investigate this more in another PR like: https://github.com/strapi/strapi/pull/20179 or another refactor PR where I'd also like to move `buildMediaResponse` and `buildRelationResponse` outside of this function.

remidej: while having to do this is a bummer, I doubt there would be a much cleaner way to do it

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -349,24 +330,115 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-                attributeSchema
-              );
+            if (
+              attributeSchema.type === 'relation' &&
+              attributeSchema.relation !== 'morphToOne' &&
+              attributeSchema.relation !== 'morphToMany'
+            ) {
+              /**
+               * Don't build the relations response object for relations to admin users,
+               * because pickAllowedAdminUserFields sanitizati
```
remidej, line 401: unfinished comment here

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -349,24 +330,115 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                  attributeValues.map(async (userToPopulate) => {
+                    if (userToPopulate == null) {
+                      return null;
+                    }
+
+                    const user = await strapi
+                      .query('admin::user')
+                      .findOne({ where: { id: userToPopulate.id } });
+
+                    const permissionChecker = getContentManagerService('permission-checker').create(
```
remidej, line 414: You should move this outside of the loop

markkaylor: Good thinking!

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -349,24 +330,115 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                      {
+                        userAbility: params.state.userAbility,
+                        model: 'admin::user',
+                      }
+                    );
+                    const sanitizedUser = await permissionChecker.sanitizeOutput(user);
+
+                    return sanitizedUser;
```
Marc-Roig, line 421: can't you sanitize at the end? I mean, sanitizing the entire entry at once

Marc-Roig: just saw remis comment , he was asking the same

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -349,24 +330,115 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                } else {
+                  // The related content has been deleted
+                  currentRelationData.meta.missingCount += 1;
+                }
+
+                return currentRelationData;
+              },
+              Promise.resolve({
+                results: [] as any[],
+                meta: { missingCount: 0 },
```
Marc-Roig, line 370: probably same question as the other attributes, are media sanitized if using this format?

packages/core/content-manager/server/src/history/services/history.ts
Marc-Roig: the one about moving the user sanitization logic outside the loop

markkaylor: ahh I had understood that he wanted me to move the creation of the permission checker outside of the loop since it only needs to be created once. Is that what you meant here as well?

markkaylor:
> I would rather not tie data population with sanitization if possible, and looking at this situation, would it be possible to keep the entry structure the same (without the results & meta object wrapping it) , and keeping a context of missing relations?

I would be very interested in refactoring to find a better solution in an immediate follow up PR, but I was thinking of this PR as a patch to fix failing e2e tests currently on `v5/main`.  

I also have this PR open that refactors history quite a bit: https://github.com/strapi/strapi/pull/20179 as well, so it might be nice to have that before making big changes here. 

wdyt? Would you prefer I handle the refactor now on this PR?

Marc-Roig:
sounds good to me to move this forward to fix the e2e 🥳

packages/core/content-manager/server/src/history/controllers/history-version.ts
```diff
@@ -60,21 +61,35 @@ const createHistoryVersionController = ({ strapi }: { strapi: Core.Strapi }) =>
...
       });
 
+      const sanitizedResults = await async.map(
+        results,
+        async (version: HistoryVersions.HistoryVersionDataResponse & { locale: string }) => {
+          return {
+            ...version,
+            data: await permissionChecker.sanitizeOutput({
+              ...version.data,
+              locale: version.locale.code,
```
Marc-Roig, line 82: I ahve checked the sanitization code and I don't believe there is anything specific about needing the locale code here 🤔 

In my opinion all fields should be available in version.data already, and its format should be enough for the sanitizer to take care of any field that should be filtered out

packages/core/content-manager/server/src/history/controllers/history-version.ts
```diff
@@ -60,21 +61,35 @@ const createHistoryVersionController = ({ strapi }: { strapi: Core.Strapi }) =>
...
+            }),
+            createdBy: version.createdBy
+              ? pick(['id', 'firstname', 'lastname', 'username', 'email'], version.createdBy)
+              : undefined,
```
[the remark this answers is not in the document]
markkaylor: We are omitting them as well 🤔 : https://github.com/strapi/strapi/blob/b5974b3a1297d5342d2e76cd6dc273c387265085/packages/core/content-manager/server/src/history/constants.ts#L2

packages/core/content-manager/server/src/history/services/history.ts
markkaylor: I'm not sure...I don't think so. But I was actually under the impression there was nothing to sanitize for media. Are there fields on a media asset that could contain sensitive data (createdBy is not populated in the response)?

Marc-Roig: Can not think of a concrete attribute that needs to be sanitized, but I would try to follow the rule of sanitizing everything by default, just in case

Marc-Roig: derrick might haunt us, hitting us with a stick, during our dreams if we dont do so

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -352,12 +351,19 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+                  model: 'plugin::upload.file',
+                });
+
                 const relatedEntry = await strapi.db
                   .query('plugin::upload.file')
                   .findOne({ where: { id: entry.id } });
 
-                if (relatedEntry) {
-                  currentRelationData.results.push(relatedEntry);
+                const sanitizedEntry = await permissionChecker.sanitizeOutput(relatedEntry);
```
remidej, line 363: I wonder why you picked this in the end and not mainfield only population

packages/core/content-manager/server/src/history/controllers/history-version.ts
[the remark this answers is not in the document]
Marc-Roig: do we need to omit it if we are readding it again here?

Marc-Roig (approved):

remidej (approved):

packages/core/content-manager/server/src/history/controllers/history-version.ts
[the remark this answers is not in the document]
markkaylor: @remidej do you remember why we omit this in `data` and then return it here in the response?

packages/core/content-manager/server/src/history/services/history.ts
markkaylor: Here for upload I don't think a mainField is available 🤔 but maybe it is. For the other instance with relations I did test it out but I think I abandoned it quite quickly when values weren't showing up as expected. I'll try it again real quick.

remidej: as you want I'm just curious, don't want to block you. And I doubt sanitization is needed on media assets

markkaylor: Sounds good, I might just merge this now then and circle back to the idea in another PR with some of the refactor ideas.

PR #20214, opened 2024-04-26
chore(tests): update snapshots

markkaylor:
### What does it do?

- Updates snapshots

### Why is it needed?

- There was a DS update...I think 🤔 
- It's causing FE tests to break

### How to test it?

- The tests in the CI should pass


markkaylor:
> I think this is flakey because of an effect in the header i wonder if we should `waitFor` the header to have the style `height`? 🤔

I had it failing consistently on https://github.com/strapi/strapi/pull/20212 but looks like it's passing now after pulling in v5/main again 🤷‍♂️ closing this one.

Issue #20141, opened 2024-04-17
A duplicated entry can be saved even if a unique field shares the same value as the original one


lucasboilly:
### Required System information

- Node.js version: 18.17.1
- NPM version: 9.6.7
- Strapi version: 4.23.1
- Database: sqlite
- Operating system: macOS 14.4.1
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When an entry is duplicated, it's still possible to save it even if a unique field have the same value as in the original one. However, if the value is changed, saved and that we write the identical value again, the error is displayed.

### Steps to reproduce the behavior

1. Create a Content-Type with a unique field in it
2. Create an entry and fill the unique field
3. Go to the List View
4. Duplicate the entry
5. Click "Save"

### Expected behavior

I'm supposed to see an error on the unique field an not being able to save.

### Screenshots

https://github.com/strapi/strapi/assets/45385696/20a1908b-d642-4d03-96fa-00e35c381bbb

### Additional context

Maybe is it linked to https://github.com/strapi/strapi/issues/20001 ?

Sruthi-Palle:
hi @lucasboilly this is sruthi a frontend developer , i can resolve this issue, can you assign this to me.

joshuaellis:
We don't assign issues. Just feel free to open a PR 😄

dplumptre:
i have a PR for this issue: https://github.com/strapi/strapi/pull/20221

hanpaine:
This has been fixed on Strapi 5 and a fix will be ported to V4 in coming weeks

hanpaine:
Fixed on V4 here: https://github.com/strapi/strapi/pull/20963

PR #20221, opened 2024-04-27
fixed issue 20141

dplumptre:
### What does it do?
It fixes the issue #20141 which says A duplicated entry can be saved even if a unique field shares the same value as the original one #20141

Basically updated the function for the clone to validate a check if there is a unique field and stop it from saving.

### Why is it needed?
If it is not done people will be submitting duplicated data of what is supposed to be unique

### How to test it?

Steps to reproduce the behavior
Create a Content-Type with a unique field in it
Create an entry and fill the unique field
Go to the List View
Duplicate the entry
Click "Save"
Expected behavior
I'm supposed to see an error on the unique field an not being able to save.

### Related issue(s)/PR(s)


alexandrebodin:
This issue has been fixed know. Closing :) 

Thank you for your contribution 🚀

PR #20223, opened 2024-04-28
tests(e2e): improve login util stability

innerdvations:
### What does it do?

- Adds a wait for network idle everywhere that we do a click action that causes network activity
- changes the login utility to clear cookies and navigate to /admin -- this was because rarely there would be an issue where sessions got messed up and a successful login was not being detected and returning the user to the login page
- updates playwright version
- fixes some issues on tokens tests
- use separate page objects for login tests
- converts some files to typescript
- converts some hardcoded strings to constants

### Why is it needed?

For some reason, almost consistently near the end of a test run, the login will fail to load the page and gets an infinitely spinning loader instead. We really need to fix that, but until we can solve it, this will get our e2e CI working again since right now it's nearly impossible to get e2e to pass.

### How to test it?

CI e2e tests should pass

### Related issue(s)/PR(s)

Once I'm sure this passes consistently on the CI, I'll cherry pick this and do the same thing on develop

DX-1393

[v4 version of this PR](https://github.com/strapi/strapi/pull/20224)


tests/e2e/tests/admin/login.spec.ts
```diff
@@ -14,26 +14,27 @@ test.describe('Login', () => {
...
-      await expect(page).toHaveTitle('Homepage | Strapi');
+      await expect(page).toHaveTitle(TITLE_HOME);
 
+      // Close the page and open a new one
       await page.close();
-
-      page = await context.newPage();
-      await page.goto('/admin');
-      await expect(page).toHaveTitle('Strapi Admin');
+      const nonPersistentPage = await context.newPage();
```
innerdvations, line 25: Re-using the page object was rarely but sometimes causing a "page already closed" when attempting to navigate after

tests/e2e/tests/admin/api-tokens.spec.ts
```diff
@@ -37,16 +40,25 @@ test.describe('API Tokens', () => {
...
+
     await navToHeader(page, ['Settings', 'API Tokens'], 'API Tokens');
 
-    const row = page.getByRole('gridcell', { name: 'my test token', exact: true });
-    await expect(row).toBeVisible();
+    const nameCell = page.getByRole('gridcell', { name: 'my test token', exact: true });
+    await expect(nameCell).toBeVisible();
+
+    // Locate the parent of nameCell and then search within it for the timestamp
+    const parentRow = nameCell.locator('xpath=..');
```
innerdvations, line 61: adds check present on transfer tokens that was missing from api tokens

tests/e2e/tests/admin/signup.spec.ts
```diff
@@ -96,8 +96,11 @@ test.describe('Sign Up', () => {
   test('a user should be able to signup when the strapi instance starts fresh', async ({
     page,
   }) => {
-    await page.getByRole('button', { name: "Let's start" }).click();
+    await Promise.all([
```
innerdvations, line 99: Using Promise.all with waitForLoadState and a click action works better to give clicks the time necessary to load the next page

tests/e2e/tests/admin/transfer-tokens.spec.ts
```diff
@@ -57,9 +60,13 @@ test.describe('Transfer Tokens', () => {
...
 
-    const row = page.getByRole('gridcell', { name: 'my test token', exact: true });
-    await expect(row).toBeVisible();
-    await expect(page.getByText(/\d+ (second|minute)s? ago/)).toBeVisible();
+    const nameCell = page.getByRole('gridcell', { name: 'my test token', exact: true });
+    await expect(nameCell).toBeVisible();
+
+    // Locate the parent of nameCell and then search within it for the timestamp
+    const parentRow = nameCell.locator('xpath=..');
+    await expect(parentRow).toContainText(/\d+ (second|minute)s? ago/);
```
innerdvations, line 68: Since we aren't clearing the db after each run anymore and have multiple tokens, we need to be more specific with what we're looking for

innerdvations:
So far this has run four times locally and twice on the CI with only the content history failing, so I think this helps a lot compared to before

tests/e2e/constants.js
```diff
@@ -41,9 +43,17 @@ const ALLOWED_CONTENT_TYPES = [
 const ADMIN_EMAIL_ADDRESS = 'test@testing.com';
 const ADMIN_PASSWORD = 'Testing123!';
 
+const TITLE_LOGIN = 'Strapi Admin';
+const TITLE_HOME = 'Homepage | Strapi';
```
innerdvations, line 47: This is so we stop having so many conflicts merging `develop` <-> `v5/main`

tests/e2e/tests/admin/signup.spec.ts
```diff
@@ -96,8 +96,11 @@ test.describe('Sign Up', () => {
...
+      page.waitForLoadState('networkidle'),
```
jhoward1994, line 100: Maybe we should make networkidle a const somewhere, it gets used quite a bit

innerdvations: Since it's a typed value for waitForLoadState I'll leave it for now, but I actually think the next time I'm in here messing with things I'm going to add a util for navigating + waiting so you can just call one line instead of 2+, nobody should have to remember to do this to write a stable test

innerdvations:
login.spec.ts just failed 😭 I'll take one more look at it, but at least it's the only e2e test that has failed in the last 10 runs or so.

joshuaellis:
> login.spec.ts just failed 😭 I'll take one more look at it, but at least it's the only e2e test that has failed in the last 10 runs or so.

can you share a trace?

innerdvations:
> > login.spec.ts just failed 😭 I'll take one more look at it, but at least it's the only e2e test that has failed in the last 10 runs or so.
> 
> can you share a trace?

Just the one in CI: https://github.com/strapi/strapi/actions/runs/8880591655/job/24382167567?pr=20223

Most of the fixing I've done in there is the same flaky login in general, compounded by creating multiple browsers. Whatever is going wrong here is probably another case I missed where we're not waiting for something.

tests/e2e/tests/admin/transfer-tokens.spec.ts
```diff
@@ -26,12 +26,19 @@ const createTransferToken = async (page, tokenName, duration, type) => {
 };
 
 test.describe('Transfer Tokens', () => {
-  test.beforeEach(async ({ page }) => {
+  let page: Page;
```
joshuaellis, line 29: Is this a pattern we want to promote? you effectively make the last test dependent on the previous because you're not starting with a "fresh" slate every time. 

I'd be cautious about promoting this pattern as I don't think it's something we want to get into the habit of doing in the long term?

joshuaellis: > Since we aren't clearing the db after each run anymore and have multiple tokens, we need to be more specific with what we're looking for

My point exactly.

joshuaellis (commented):
I dont think there are any issues with the code apart from what i've highlighted which im concerned is not a pattern we want to start doing 👀

tests/e2e/tests/admin/transfer-tokens.spec.ts
innerdvations: Unfortunately I think the bad practice is the core of the fix here. I will take a look and see if there's an alternative I can find, but otherwise I may just add a warning not to repeat that pattern in other tests.

PR #20233, opened 2024-04-30
Fix flaky unit test for the CMReleasesContainer component

simotae14:
### What does it do?

I tried to find a way to fix the flaky unit test in the content releases

### Why is it needed?

To pass the CI step

### DIFF
--- a/packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
+++ b/packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
@@ -64,11 +64,8 @@ describe('CMReleasesContainer', () => {
       expect(informationBox).toBeInTheDocument();
     });
 
-    // const informationBox = await screen.findByRole('complementary', { name: 'Releases' });
-    // expect(informationBox).toBeInTheDocument();
-
-    // const addToReleaseButton = await screen.findByRole('button', { name: 'Add to release' });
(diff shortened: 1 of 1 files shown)


packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
```diff
@@ -59,8 +59,10 @@ describe('CMReleasesContainer', () => {
   it('should render the container', async () => {
     render();
 
-    const informationBox = await screen.findByRole('complementary', { name: 'Releases' });
-    expect(informationBox).toBeInTheDocument();
+    await waitFor(async () => {
```
markkaylor, line 62: I think findBy queries already use a waitFor https://testing-library.com/docs/dom-testing-library/api-async#findby-queries. I think what you end up doing here with an additional waitFor is increase the timeout / retries. But It looks like you can pass waitFor options to the findBy query to do that as well.

joshuaellis: Mark is right, it'll also fail if it's not in the document. There's probably a different `async` fetch happening elsewhere that updates the component, you might needed to await something else.

joshuaellis:
since @remidej merged https://github.com/strapi/strapi/pull/20250, can we close this?

simotae14:
Already fixed in https://github.com/strapi/strapi/pull/20250

Issue #20170, opened 2024-04-22
bulk publish only works for the first selected item


cdreier:
### Required System information

- Node.js version: v20.12.0
- NPM version: 10.5.0
- Strapi version: v4.23.1
- Database: sqlite
- Operating system: fedora
- Is your project Javascript or Typescript: js

### Describe the bug

the bulk publish is only working for the very first click - if you click the "select all", everything is fine. 
if you select entries one by one, only the first is showing up - all other are ignored

### Steps to reproduce the behavior

1. Go to a list view
2. Click on first unbublished item
3. Click on second unbublished item
4. Click publish - only the first items shows up

### Expected behavior

All selected items are available to bulk publish

### Screenshots

[strapi-bulk-publish.webm](https://github.com/strapi/strapi/assets/731608/2bd6107b-92a4-4a3f-ad54-52b0a0b1e141)

### Code snippets

If applicable, add code samples to help explain your problem.

### Additional context

i just tested this with a brand new quickstarted strapi to make sure, no plugins etc are messing things up - but this is also happening on our production instance

PR #20234, opened 2024-04-30
fix(content-manager): bulk publish would only ever show first entry to be published

Feranchz:
### What does it do?

~This pull request fixes a bug with the Content Manager's Bulk Publish button based on two premises:~

1. ~All logic related to displaying data in the Bulk Publish modal should reside in the `SelectedEntriesModalContent` and not outside of it. `PublishAction` should contain only logic related to when to show or hide the Bulk Publish button.~
2. ~Since all data related to the ListView is already present in the listView table, it is unnecessary to make another request.~

* Simplifies the Publish Modal by reducing the amount of times we use `setState`, the state is derived originally from the ListView selected items that the action receives. It then queries for *just* these items from the API, so everything is updated originally from the ListView reducing the amount of different times we set what should be viewed which required synchronising different states.

### How to test it?

* As a user I want to be able to select and publish more than one entry from the list view

### Related issues

* resolves https://github.com/strapi/strapi/issues/20170

### DIFF
--- a/packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
+++ b/packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
@@ -19,6 +19,7 @@ import {
   Table as HelperPluginTable,
   getYupInnerErrors,
   useFetchClient,
+  useQueryParams,
   useNotification,
   TranslationMessage,
   useAPIErrorHandler,
@@ -28,7 +29,7 @@ import { Contracts } from '@strapi/plugin-content-manager/_internal/shared';
 import { Entity } from '@strapi/types';
 import { AxiosError, AxiosResponse } from 'axios';
 import { MessageDescriptor, useIntl } from 'react-intl';
-import { useMutation, useQueryClient } from 'react-query';
+import { UseQueryResult, useMutation, useQuery, useQueryClient } from 'react-query';
 import { Link, useHistory } from 'react-router-dom';
 import styled from 'styled-components';
 import { ValidationError } from 'yup';
@@ -39,7 +40,7 @@ import { createYupSchema } from '../../../../utils/validation';
 import { useAllowedActions } from '../../hooks/useAllowedActions';
 import { Table } from '../Table';
 
-import { ConfirmDialogPublishAll } from './ConfirmBulkActionDialog';
+import { ConfirmDialogPublishAll, ConfirmDialogPublishAllProps } from './ConfirmBulkActionDialog';
 
 import type { BulkActionComponent } from '../../../../../core/apis/content-manager';
 
@@ -117,14 +118,14 @@ const EntryValidationText = ({
 
 interface SelectedEntriesTableContentProps {
   isPublishing?: boolean;
-  rowsToDisplay?: TableRow[];
+  rows?: TableRow[];
   entriesToPublish?: Entity.ID[];
   validationErrors: Record<string, EntryValidationTextProps['validationErrors']>;
 }
 
 const SelectedEntriesTableContent = ({
   isPublishing,
-  rowsToDisplay = [],
+  rows = [],
   entriesToPublish = [],
   validationErrors = {},
 }: SelectedEntriesTableContentProps) => {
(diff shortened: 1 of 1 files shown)


joshuaellis (commented):
Code looks fine, i think we're re-writing it a bit in v5. I've not QAd though.

packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -386,7 +411,7 @@ const SelectedEntriesModalContent = ({
         }
         endActions={
           <Flex gap={2}>
-            <Button onClick={() => refetchModalData()} variant="tertiary" loading={isFetching}>
+            <Button onClick={refetchList} variant="tertiary" loading={isFetching}>
```
madhurisandbhor, line 414: `refetch` is broken as you are invalidating list view not the modal content

madhurisandbhor (changes requested):

packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -241,26 +237,55 @@ interface TableRow {
 }
 
 const SelectedEntriesModalContent = ({
-  toggleModal,
-  refetchModalData,
-  setEntriesToFetch,
   setSelectedListViewEntries,
```
madhurisandbhor, line 240: why this needs to be passed from Publish action? you can get it from [here](https://github.com/strapi/strapi/pull/20234/files#diff-fc5c3b6fd53b68ac6c9556235411783d8d6e6c25cd807ce2a6d8ad474428dc8fR245)

packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -241,26 +237,55 @@ interface TableRow {
...
+
+  const validationErrors = React.useMemo(() => {
+    if (rows.length === 0 || !contentType) return {};
+
+    const schema = createYupSchema(
+      contentType,
+      { components },
+      { isDraft: false, isJSONTestDisabled: true }
+    );
+    const validationErrors: Record<Entity.ID, Record<string, TranslationMessage>> = rows.reduce(
```
madhurisandbhor, line 263: this is same name as parent

packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
Feranchz: Do you say is broken for something you see? 🤔  Modal content is the same that listView content, just filtered to show only selected entries.

In my opinion the last behaviour of fetch entries by separated doesn't have sense because I don't want to make the same request twice, just use the information that I already have

packages/core/admin/admin/src/content-manager/pages/ListView/components/BulkActions/PublishAction.tsx
Feranchz: Needs to be passed because we have two different Table Context.

1. (Parent Table) List View Table
2. (Child Table) Modal Table

when we use `useTableContext` hook in the modal, we are accessing the context of the child table, not the parent, so we pass this function to update the selectedEntries in the parent table

joshuaellis (approved):
QA'd ✅

Feranchz:
@joshuaellis maybe I'm missing something but I still don't understand why we need to make a request every time user select a row on the list view's table and not when they open the publish modal

joshuaellis:
> @joshuaellis maybe I'm missing something but I still don't understand why we need to make a request every time user select a row on the list view's table and not when they open the publish modal

What would your alternative be? :) it feels like the alt is refresh the entire list and derive data from the redux store? Which could be equally problematic because you're over-fetching most times if I understand correctly? i.e. your list contains 20 items and you want to bulk publish 3, you update one because it's not ready, you refresh and you fetch 20 instead of 3?

Feranchz:
@joshuaellis I think is a good idea to have a separate request, what I don't like is that we make the request every time a row is checked/unchecked, I think we could save all those requests and only made one when the modal is opened, then if user clicks refresh we can make the request again 

In any case, really thank you for handle this yesterday so the fix could be ship in the next patch release. If you are okay we could explore this in the next days or for v5 🤓

joshuaellis:
> I think is a good idea to have a separate request, what I don't like is that we make the request every time a row is checked/unchecked, I think we could save all those requests and only made one when the modal is opened, then if user clicks refresh we can make the request again

I see what you're saying yeah, you could `enable` the query if the dialog is showing, im not sure expose that state, but i'd love to see the proposal in V5 💪🏼 if you make any changes to the renderer API I would be sure to make those same changes to the document / header actions because they're one of the same.

PR #20237, opened 2024-04-30
fix(content-manager):  return default locale

### LINKED ISSUE #20209: Returning to Default Locale (EN) does not Reset the Filter

- Node.js version: v20.10.0
- NPM version: 10.2.3
- Strapi version: 4.24.0
- Database: SQLite
- Operating system: Ubuntu 20.4

### Describe the bug

I have attached video with English audio. Its really hard to explain in writing. In short, the locale filter in the drop down continue to show the previously selected locale and not resetting to EN. 

1. `npx create-strapi-app@latest my-project --quickstart`
2. log into Strapi V4 
3. Select Resources collection from Content Manager 
4. on the right hand side top of the resource table/ index, select a locale from the locale drop down 
5. then click Resources collection on the left nav list 

### Expected behavior

Locale filter should have correct locale loaded when user click on collection type again

### Screenshots

https://github.com/strapi/strapi/assets/93124968/3282ebc7-a767-429f-9b86-3637301a1d45

### Code snippets

yurimutti:
### What does it do?

Added key on InjectionZone to reset internal states. 

Now with **params.locale** as key. Default locale can be persisted after user click on collection. 

### Why is it needed?

Locale filter doesn't have correct locale loaded when user click on collection type again.

### How to test it?

- npx create-strapi-app@latest my-project --quickstart
- log into Strapi V4
- Select Resources collection from Content Manager
- on the right hand side top of the resource table/ index, select a locale from the locale drop down
- then click Resources collection on the left nav list

### Related issue(s)/PR(s)

Fix #20209


yurimutti:
ezgif-6-0cf907f3f0

joshuaellis:
Can you rebase off develop please, your PR should not have this many commits.

Also I'm not sure this is the appropriate solution, I think we should be looking at how the injected component works in i18n rather than trying to solve it in the CM.

This also assumes i18n is installed and adds locale which isn't necessary true and pollutes the CM domain with i18n concepts, something we try to avoid where possible.

yurimutti:
> Can you rebase off develop please, your PR should not have this many commits.
> 
> Also I'm not sure this is the appropriate solution, I think we should be looking at how the injected component works in i18n rather than trying to solve it in the CM.
> 
> This also assumes i18n is installed and adds locale which isn't necessary true and pollutes the CM domain with i18n concepts, something we try to avoid where possible.

Makes total sense!

alexandrebodin:
Hello @yurimutti I see this PR is still in Draft, are you still planning on finish it ?

alexandrebodin:
Closing this PR as staled. If you want to rework it for v5 let us know 🙏

PR #20238, opened 2024-05-01
fix(admin): if were in EE mode wait for the EE routes to be loaded before rendering

### LINKED ISSUE #20219: Enterprise SSO will not redirect properly and allow users to login

- Node.js version: 20.11.1
- Yarn version: 1.22.22
- Strapi version: 4.24.0
- Database: SQLite

Strapi report info:

```
$ strapi report --all
Launched In: 282 ms
Environment: development
OS: linux-x64
Strapi Version: 4.24.0
Node/Yarn Version: yarn/1.22.22 npm/? node/v20.11.1 linux x64
Edition: Enterprise
Database: sqlite
UUID: 02f807bc-204f-43ae-963d-8b3be826d65d
Dependencies: {
  "@strapi/plugin-cloud": "4.24.0",
  "@strapi/plugin-i18n": "4.24.0",
  "@strapi/plugin-users-permissions": "4.24.0",
  "@strapi/strapi": "4.24.0",
  "better-sqlite3": "8.6.0",
  "passport-discord": "^0.1.4",
  "react": "^18.0.0",
  "react-dom": "^18.0.0",
  "react-router-dom": "5.3.4",
  "styled-components": "5.3.3"
}
Dev Dependencies: {}
```

### Describe the bug

Redirects on admin auth page are broken causing the error page to not show and/or not redirecting after successful login. Refreshing page after successful login does not work either. Effectively: SSO is broken completely for most users, randomly will work for a few others after several tries.

1. Configure SSO with any provider
2. Enable SSO
3. Try to login with SSO

Reproduction video:

https://www.loom.com/share/0e77ba46bfee4805921a8b6c15795a59?sid=3bf8cd13-8985-4510-9cc6-e147e65a3579

### Expected behavior

Should allow login or show error page and handle redirects properly

### Screenshots

N/A

### Code snippets

```js
// path: ./config/admin.js

const DiscordStrategy = require("passport-discord");

  auth: {
    secret: env("ADMIN_JWT_SECRET"),
    providers: [
      {
        uid: "discord",
        displayName: "Discord",
        icon: "https://cdn0.iconfinder.com/data/icons/free-social-media-set/24/discord-512.png",
        createStrategy: (strapi) =>
          new DiscordStrategy(
            {
              clientID: env("DISCORD_CLIENT_ID"),
              clientSecret: env("DISCORD_SECRET"),
              callbackURL:
                strapi.admin.services.passport.getStrategyCallbackURL(
                  "discord"
                ),
              scope: ["identify", "email"],
            },
            (accessToken, refreshToken, profile, done) => {
              done(null, {
                email: profile.email,
                username: `${profile.username}#${profile.discriminator}`,
              });
            }
          ),
      },
    ],
  },
  apiToken: {
    salt: env("API_TOKEN_SALT"),
  },
  transfer: {
    token: {
      salt: env("TRANSFER_TOKEN_SALT"),
    },
  },
  flags: {
    nps: env.bool("FLAG_NPS", false),
    promoteEE: env.bool("FLAG_PROMOTE_EE", false),
  },
});
```

```js
// path: ./config/middlewares.js

module.exports = [
  "strapi::logger",
  "strapi::errors",
  {
    name: "strapi::security",
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          "connect-src": ["'self'", "https:"],
          "img-src": [
            "'self'",
            "data:",
            "blob:",
            "market-assets.strapi.io",
            "cdn0.iconfinder.com",
          ],
          "media-src": [
            "'self'",
            "data:",
            "blob:",
            "market-assets.strapi.io",
            "upload.wikimedia.org",
            "cdn0.iconfinder.com",
          ],
          upgradeInsecureRequests: null,
        },
      },
    },
  },
  "strapi::cors",
  "strapi::poweredBy",
  "strapi::query",
  "strapi::body",
  "strapi::session",
  "strapi::favicon",
  "strapi::public",
];
```

### Additional context

Related to ticket ID 6597, would normally be high but raised to critical as part of escalation process

joshuaellis:
### What does it do?

* if we're in EE the `routes` variable will be an `array` not `null`, if that's the case we wait for the routes to be loaded (which is not feature specific) before trying to re-render the app after being redirected from `admin/connect/PROVIDER` to `admin/login/success` because the component would not be ready to handle it & therefore the redirect will never complete

### Why is it needed?

* You can't login with SSO at the moment

### How to test it?

Experimental: `0.0.0-experimental.ccba4ea6043b6bab83888f85cc9a3f07f61dc262`

* Follow steps here https://docs.strapi.io/dev-docs/configurations/sso (I used Github, it was v easy)
* login with SSO
* be redirected to the homepage!

### Related issue(s)/PR(s)

* resolves #20219
* resolves DX-1392

### DIFF
--- a/packages/core/admin/admin/src/App.tsx
+++ b/packages/core/admin/admin/src/App.tsx
@@ -137,7 +137,13 @@ export const App = ({ authLogo, menuLogo, showReleaseNotification, showTutorials
     [uuid, telemetryPropertiesQuery.data]
   );
 
-  if (initQuery.isLoading) {
+  /**
+   * `routes` will only be an array if EE mode is activated,
+   * therefore we need to ensure the lazy loading of EE routes
+   * has been complete otherwise the SSO repsonse URL will not be
+   * caught by the component and you won't be able to login with it.
+   */
+  if (initQuery.isLoading || (Array.isArray(routes) && routes.length === 0)) {
     return <LoadingIndicatorPage />;
   }
 


jhoward1994 (approved):
Works for me with GH 🚀

derrickmehaffy (approved):
Tested locally and it works for both the error screen and the normal login. It won't redirect to a specific page if it was opened but that has been that way for a while since we use session tokens

packages/core/admin/admin/src/App.tsx
```diff
@@ -137,7 +137,13 @@ export const App = ({ authLogo, menuLogo, showReleaseNotification, showTutorials
...
   );
 
-  if (initQuery.isLoading) {
+  /**
+   * `routes` will only be an array if EE mode is activated,
+   * therefore we need to ensure the lazy loading of EE routes
+   * has been complete otherwise the SSO repsonse URL will not be
+   * caught by the component and you won't be able to login with it.
+   */
+  if (initQuery.isLoading || (Array.isArray(routes) && routes.length === 0)) {
```
innerdvations, line 146: Is there any case where routes could be 0 on purpose? Like if all ee features are disabled but it's still EE?

joshuaellis: We could make `useEnterprise` better for sure e.g. adding "state" like `isLoading` etc.

I think if we wanted to do that we should revisit the hook entirely, the types aren't great etc. So it'd be another ticket 🤔

joshuaellis: This also isn't an issue in V5 because of how the router works there.

PR #20240, opened 2024-05-01
fix: issue 20138

alexandrebodin:
### What does it do?

Apply https://github.com/strapi/strapi/pull/20231 to 4.24.1


joshuaellis (approved):
Since this is a carbon copy of the other PR

PR #20243, opened 2024-05-02
chore: rename core packages

remidej:
### What does it do?

Renames the packages that were "core plugins" in v4, that are now core packages, so that they no longer have `plugin-` in their name. Those are the CM, CTB, email, upload and i18n.


packages/utils/upgrade/resources/codemods/5.0.0/dependency-remove-strapi-plugin-i18n.json.ts
joshuaellis, line 1: i think this needs to be kept as the old name because people won't have installed the new version :)

innerdvations (commented):
LGTM, aside from seconding Josh's comment that we want to leave the codemod alone

packages/utils/upgrade/resources/codemods/5.0.0/dependency-remove-strapi-plugin-i18n.json.ts
remidej: great catch!

joshuaellis (approved):
This looks good, do we need to release these packages manually to avoid NPM errors with them not being public? 🤔 might wanna talk to @alexandrebodin or @Marc-Roig first to double check that plan...

remidej:
@joshuaellis yes I checked with Alex, he needs to do some manual steps before publication, so it won't be in today's release

PR #20245, opened 2024-05-02
feat(main-nav): Main nav refactoring, refactor the main nav container and all the links and user profile

simotae14:
What does it do?
It shows the new main nav. Now all the links are shown with just the icon, the divider between the links are removed and we are using new components in the admin to render the Main nav

Why is it needed?
In the refactoring of the Main Nav we want a new narrow nav to add more space for the important content, also we removed the Dividers (hr elements) between the links

How to test it?
Just check the new main nav in the different themes, using a CE license and EE license, all the states of the links (active, hover) and the Profile button

You can find all the info and link to the figma file in the Jira ticket

Related issue(s)/PR(s)
CS-758 and CS-766


simotae14:
It is still in draft because I need to refine some stuff and add tests but the ui is implemented so I need some comments just to understand if the main work is ok or not, thanks

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const NavUser = React.forwardRef<HTMLButtonElement, NavUserProps>(
+  ({ children, initials, ...props }, ref) => {
+    return (
+      <Flex paddingTop={3} paddingBottom={3} ref={ref} justifyContent="center" {...props}>
+        <Flex as="button" justifyContent="center">
+          <Initials>{initials}</Initials>
+          <VisuallyHidden>
+            <span>{children}</span>
```
joshuaellis, line 19: I'm not sure there's a benefit to wrapping `children` in a `span` within `VisuallyHidden` is there?

packages/core/admin/admin/src/components/MainNav/NavLink.tsx
```diff
@@ -38,7 +39,11 @@ const MainNavLinkWrapper = styled(RouterLink)`
 `;
 
 const LinkImpl = ({ children, ...props }: LinkProps) => {
-  return <MainNavLinkWrapper {...props}>{children}</MainNavLinkWrapper>;
+  return (
+    <Flex as="li">
```
joshuaellis, line 43: Why is a link now a list litem? this doesn't feel named right imo.

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -103,180 +86,121 @@ const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) =
...
-      <NavFooter>
-        <NavUser
-          id="main-nav-user-button"
-          ref={navUserRef}
-          onClick={handleToggleUserLinks}
-          initials={initials}
+            })
+          : null}
+      </NavListWrapper>
+      <NavUser
```
joshuaellis, line 165: Why wouldnt we have the menu & the User section all in the same component? Is there a benefit to the menu live outside the declaration for it's trigger?

joshuaellis: Also reviewing the semantics this should be a `menu`. I'd look at using the `Menu` component instead, this will handle stuff we expect from a popover. It probably means we can remove the `onBlur` handler, which im not totally clear the point of.

joshuaellis (changes requested):
some initial feedback, it's main the user's part im a bit concerned about the approach we have.

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
simotae14: Do you mean completely removing the Visually hidden content?

joshuaellis: no i mean i would just do:

```
<VisuallyHidden>{children}</VisuallyHidden>
```

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: ok I refactored the component, now all the logic of the NavUser is inside the component

derrickmehaffy:
Small bit of feedback, can we push the right side info-blocks over a bit more and make it smaller to leave more room for the main editing part?

image

joshuaellis:
I think this is OOS for this PR, but worth sharing with @lucasboilly 👍🏻

derrickmehaffy:
> I think this is OOS for this PR, but worth sharing with @lucasboilly 👍🏻

No biggie I understand, just seems like we are wasting a ton of space there.

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
```diff
@@ -0,0 +1,88 @@
...
+          </LinkUser>
+          <LinkUser onClick={logout} to="/auth/login">
+            <Typography textColor="danger600">
+              {formatMessage({
+                id: 'app.components.LeftMenu.logout',
+                defaultMessage: 'Logout',
+              })}
+            </Typography>
+            <SignOut />
+          </LinkUser>
```
joshuaellis, line 83: you need to be using the MenuItem here so that the accessibility of focus & keyboard navigation works correctly :)

packages/core/admin/admin/src/components/MainNav/tests/NavUser.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  it('shows the initials of the user', async () => {
+    render(<NavUser initials="JD">John Doe</NavUser>);
+    expect(screen.getByText('JD')).toBeInTheDocument();
+  });
+
+  it('contains the user name', async () => {
+    render(<NavUser initials="JD">John Doe</NavUser>);
+    const userName = screen.getByText('John Doe');
+    expect(userName).toBeInTheDocument();
+  });
```
joshuaellis, line 15: for these you should be able to do `getByRole("button", {name: NAME })`, if you prefer.

joshuaellis (approved):
Not QAd but code looks good ✅ lets squash merge this when we do merge though, not sure we need all these commits!

packages/core/admin/admin/src/components/MainNav/tests/NavUser.test.tsx
simotae14: I tried but it doesn't work, but in NavUser there is just one button

joshuaellis: interesting................ note to self.

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -1,82 +1,38 @@
...
-  FocusTrap,
-  Typography,
-  MainNav,
-  NavBrand,
-  NavCondense,
-  NavFooter,
-  NavLink,
-  NavSection,
-  NavSections,
-  NavUser,
```
remidej, line 16: Does this mean we can remove all these from the design system?

remidej (approved):
this works great!

✅ matches the designs
✅ only shows menu items available to the logged in user
✅ shows the EE features with a lock when on CE (though I couldn't find that in figma)
✅ able to access profile and logout
✅ the middle section is scrollable when the viewport is too short
✅ light and dark theme display as expected

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: Yeah they are not used anymore in the CMS, I am gonna create a jira in our board to remove them from the DS

lucasboilly:
Hey @simotae14 congrats, that's great! I just reported minor issue in the Blitz document :)

lucasboilly:
> Small bit of feedback, can we push the right side info-blocks over a bit more and make it smaller to leave more room for the main editing part?

Feedback noted @derrickmehaffy, thanks 👍

PR #20250, opened 2024-05-02
fix: CMReleasesContainer test

remidej:
### What does it do?

Fixes "update not wrapped in act()" error coming from CMReleasesContainer. I found that the async request at the root of the issue seemed to be [the useRBAC call in ReleaseActionMenu](https://github.com/strapi/strapi/blob/301d2027a59183bfa76012748eaeef6b5ffae9d1/packages/core/content-releases/admin/src/components/ReleaseActionMenu.tsx#L224).

Also fixes an unrelated TS error

### Related

same bug as #20233

### DIFF
--- a/packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
+++ b/packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
@@ -30,9 +30,6 @@ const render = (
       ),
     },
     initialEntries,
-    userEventOptions: {
-      skipHover: true,
-    },
   });
 
 describe('CMReleasesContainer', () => {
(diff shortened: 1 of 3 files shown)


packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
```diff
@@ -155,6 +150,7 @@ describe('CMReleasesContainer', () => {
     render();
 
     await screen.findByRole('complementary', { name: 'Releases' });
+    await screen.findByRole('button', { name: /release action options/i });
```
markkaylor, line 153: how come we need this?

markkaylor (approved):
Nice lgtm

packages/core/content-releases/admin/src/components/tests/CMReleasesContainer.test.tsx
remidej: [ReleaseActionMenu](https://github.com/strapi/strapi/blob/301d2027a59183bfa76012748eaeef6b5ffae9d1/packages/core/content-releases/admin/src/components/ReleaseActionMenu.tsx#L224) calls useRBAC, which makes an API call. This call would be resolved after the test is done, causing the act error.

When the API call resolves, the "release action options" button appears, so I wait for that inside the test

PR #20254, opened 2024-05-02
feat(graphql): allow to pass cache hints to graphql fields

iamandrewluca:
### What does it do?

https://github.com/strapi/strapi/assets/1881266/ac7f1b06-79e5-4c9a-9478-25f3bf9ba49d

### Why is it needed?

https://www.apollographql.com/docs/apollo-server/performance/caching/

Note: Strapi uses a previous major version of Apollo Server, which works similarly.

### How to test it?

### To Do

- [x] Allow to add cache hints to type fields
- [x] Allow to add cache hints to top-level Queries
- [ ] Upgrade to `nexus@next` and use directive as static instead of resolve
- [ ] Check Authenticated/Public roles cache Authenticated should be private
- [ ] Test if the cache works fine for all field types

### DIFF
--- a/packages/core/content-manager/server/src/services/data-mapper.ts
+++ b/packages/core/content-manager/server/src/services/data-mapper.ts
@@ -12,7 +12,6 @@ const dtoFields = [
   'options',
   'pluginOptions',
   'attributes',
-  'pluginOptions',
 ];
 
 export default () => ({
--- a/packages/core/types/src/types/core/schemas/index.ts
+++ b/packages/core/types/src/types/core/schemas/index.ts
@@ -102,7 +102,7 @@ export interface PluginOptions {
     cacheHint?: {
       find?: CacheHint;
       findOne?: CacheHint;
-    }
+    };
   };
 }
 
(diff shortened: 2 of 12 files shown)


packages/core/content-manager/server/src/services/data-mapper.ts
```diff
@@ -12,7 +12,6 @@ const dtoFields = [
   'options',
   'pluginOptions',
   'attributes',
-  'pluginOptions',
```
iamandrewluca, line 15: Note: Removed duplicated value.

packages/core/types/src/types/core/schemas/index.ts
```diff
@@ -97,6 +98,12 @@ export interface Options {
 
 export interface PluginOptions {
   i18n?: { localized: boolean };
+  graphql?: {
+    cacheHint?: {
+      find?: CacheHint;
+      findOne?: CacheHint;
+    }
+  };
```
iamandrewluca, line 106: Question: Why is this field not visible in find/findOne resolvers? 🤔

iamandrewluca:
I will close this experiment PR for now. Might come to it at a later time or in v5.

PR #20235, opened 2024-04-30
enhancement(CM): bulk delete and unpublish migration to v5

madhurisandbhor:
### What does it do?

Migrating bulk delete and unpublish to V5

### How to test it?

`Bulk Delete` and `Bulk Unpublish` are in action on V5, no V5 design changes are included.

### Related issue(s)/PR(s)
⚠️ Merge after this related BE PR, https://github.com/strapi/strapi/pull/20184

PR #20255, opened 2024-05-03
enhancement: bulk publish migration to v5

madhurisandbhor:
### What does it do?

Bulk Publish action migrated to v5 without any v5 design changes.

### Related issue(s)/PR(s)

⚠️ Merge after https://github.com/strapi/strapi/pull/20235 as this is created on top of it


Marc-Roig (approved):

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -0,0 +1,509 @@
...
+  const { pathname } = useLocation();
+  const { formatMessage } = useIntl();
+
+  const {
+    list: {
+      settings: { mainField },
+    },
+  } = useDocLayout();
+
+  const shouldDisplayMainField = mainField != null && mainField !== 'id';
```
simotae14, line 152: Do we really need this variable?

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -0,0 +1,509 @@
...
+      </Table.Body>
+    </Table.Content>
+  );
+};
+
+/* -------------------------------------------------------------------------------------------------
+ * BoldChunk
+ * -----------------------------------------------------------------------------------------------*/
+
+const BoldChunk = (chunks: React.ReactNode) => <Typography fontWeight="bold">{chunks}</Typography>;
```
simotae14, line 235: at the end the chunks prop is the children prop, isn't it?

Feranchz: Yes, but I think formatMessage pass chunks as the children

simotae14:
I found two strange behaviours, not sure are related to this PR to be honest.
The first one is the alignment of the confirmation message, as you can see the text in red is not aligned with the other message
Schermata 2024-05-27 alle 11 15 42
the second weird behaviour are some warning with message "Not found" that are shown when we switch from a document with entries to another one without entries as you can see in this video
 

https://github.com/strapi/strapi/assets/2589748/5dc3ee9e-6bd6-4fe4-8f1a-4a7daa2f4796

simotae14:
And also some validation error messages have no useful meaning, for example
Schermata 2024-05-27 alle 11 25 29

Feranchz:
Thanks for the errors @simotae14 !

I discovered a couple of things:

1. "Not Found" error is on main too, looks like it's related on how we use the `useDoc` hook, probably we can go with this PR and create a ticket for that error, but before lets take a look at other teams to know if someone else is already taking a look at this 
2. The validation error message is actually two errors:
a. Error formatting the error message: **Solved** ✅ 
b. The entry object we are using to validate is not nested populated, this means that for this specific case `closingPeriod.dish` is always throwing an error

For this last one we can create a ticket because it's a more complicated thing and I think is a v4 error too (I remember something similar in Releases which works the same way) 🤔

madhurisandbhor:
New changes on fetching entries on opening modal LGTM, thanks!! 
Noticed 2 issues, 
1. modal confirm message for internalisation is not correct, it just needs to be aligned at center.
2. On refresh selected/ ready to publish entries are not updated.
https://github.com/strapi/strapi/assets/7886244/15cb4a80-32d5-418b-8c2a-e3610b1abced

Feranchz:
> Would like to know why don't we use publishedAt value?

@madhurisandbhor because right now we have two versions of the entry (the published one and the in draft one). We want to use the one in draft (to validate it) and in this version publishedAt is null (because is in draft), but we want the "status" (which takes in consideration all the versions of the entry). 

Actually this change is interesting, probably we want to discuss it more with Yannis 

Could you add these two issues you found to the blitz docs? 🙏

PR #20256, opened 2024-05-03
chore: add type to bulk action actions

joshuaellis:
### What does it do?

* adds `type` to the action signature (no the description interface)

### Why is it needed?

* users can use the reducer pattern for adding bulk actions and remove ones like `publish` or `delete` or swap them out, similar to how document actions work

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/Actions.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/Actions.tsx
@@ -109,7 +109,7 @@ const BulkActionsRenderer = () => {
  * BulkActionAction
  * -----------------------------------------------------------------------------------------------*/
 
-interface Action extends BulkActionDescription, Pick<BulkActionComponent, 'actionType'> {
+interface Action extends BulkActionDescription {
   id: string;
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/Actions.tsx
```diff
@@ -138,7 +138,6 @@ const BulkActionAction = (action: Action) => {
           break;
         case 'dialog':
         case 'modal': {
-          if (action.actionType === 'delete') trackUsage('willBulkDeleteEntries');
```
madhurisandbhor, line 141: I changed `type` to `actionType` because I needed to do this 😥 and I polluted it. Thanks for the changes!!

madhurisandbhor (approved):
Merci!!

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/Actions.tsx
joshuaellis: You could add this to the action call of `delete` before the API request is sent?

PR #20257, opened 2024-05-03
feat: discard drafts when enabling draft and publish on ct

Marc-Roig:
### What does it do?

On V5, content types without draft and publish have a single version , where the publishedAt value is filled with a date.
Content Types with draft and publish have two versions, a draft and a published version, where the publishedAt value is null on drafts and filled with a date on published versions.

When enabling draft and publish on a content type, we were copying the single version with the publishedAt value filled and updating the publishedAt value to be null. 

This process did not copy components or relations, and now all documents are processed in batch, to call the `documents.discard`method, which takes care of that, at the cost of processing one version at a time.


alexandrebodin (approved):
Logic seems fine, didn't QA

Marc-Roig:
Will ask marion to include QA in a blitz session

PR #20069, opened 2024-04-09
feat(i18n): bulk locale publish modal in CM edit view

jhoward1994:
### What does it do?

Adds bulk locale publish modal to the CM edit view

### Why is it needed?

V5 core feature

### How to test it?

Added E2E test

### Related issue(s)/PR(s)

DX-1364

TODO

- [x] Errors cases. Handle API errors and display them in the modal
- [x] Tests
- [x] ~~What tracking events should we send from this action~~ Emit event to be handled by the doc service

PR #20046, opened 2024-04-05
feat(core): document service (un)publish multiple locales

jhoward1994:
### What does it do?

Allows arrays of locales to be passed to publish and unpublish in the document service

### Why is it needed?

So we can (un)publish multiple locales at once from the CM edit view

### How to test it?

Added API tests

### Related issue(s)/PR(s)

CONTENT-1903
DX-1335

PR #20258, opened 2024-05-03
Publish multiple locales 

jhoward1994:
### What does it do?

Adds bulk locale publish modal to the CM edit view

### Why is it needed?

V5 core feature

### Related issue(s)/PR(s)

CONTENT-1903
DX-1335
DX-1364

See:
https://github.com/strapi/strapi/pull/20046 (backend)
https://github.com/strapi/strapi/pull/20069 (frontend)


packages/core/content-manager/server/src/services/document-metadata.ts
```diff
@@ -79,26 +89,57 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
...
-    return (
-      Object.values(versionsByLocale)
-        .map((localeVersions: DocumentVersion[]) => {
-          // There will not be a draft and a version counterpart if the content type does not have draft and publish
-          if (!contentTypes.hasDraftAndPublish(strapi.getModel(uid))) {
-            return pick(AVAILABLE_LOCALES_FIELDS, localeVersions[0]);
+    // There will not be a draft and a version counterpart if the content
+    // type does not have draft and publish
+    const model = strapi.getModel(uid);
+    const keysToKeep = [...AVAILABLE_LOCALES_FIELDS, ...validatableFields];
```
Marc-Roig, line 95: how does the validatableFields string[] works on nested schemas? what happens if the root has a "name" field, but a component inside also has a name field?

Marc-Roig:
As discussed, I still feel uncomfortable by coupling bulk actions frontend validation into the document metadata. And I still think we would benefit from loading the document metadata asynchronously on the CM, and not returning it on the main entry retrieval requests.

But that might be a bit of work, so let's explore first this solution and we can refactor later (famous last words)

packages/core/content-manager/server/src/services/document-metadata.ts
jhoward1994: I think this is covered in this test

https://github.com/strapi/strapi/blob/483791373ac11e4d33672122e16626f3104957e8/tests/api/core/content-manager/api/document-metadata.test.api.js#L266-L267

Both the product CT and features components have a 'name' string field

https://github.com/strapi/strapi/blob/483791373ac11e4d33672122e16626f3104957e8/packages/core/content-manager/server/src/services/document-metadata.ts#L230-L231
The populate here becomes 

```
	{
	  name: true,
	  category: true,
	  price: true,
	  features: {
	    populate: {
	    },
	  },
	}
```

The keys of which are used to build `keysToKeep`. So `getAvailableLocales` returns

```
	{
	  id: 2,
	  documentId: "bpd5tcr2mngn2j0e15jjakq9",
	  name: "prod-fr-draft",
	  category: "Cat-1-fr",
	  price: 1,
	  createdAt: "2024-05-10T11:23:53.441Z",
	  updatedAt: "2024-05-10T11:23:53.441Z",
	  publishedAt: null,
	  locale: "fr",
	  features: {
	    id: 2,
	    name: "Feature 1 fr",
	  },
	  status: "draft",
	}
```

Marc-Roig: Probably not a big deal, but this might return more atteibutes than necessary. But I see!

Marc-Roig (approved):
Code looks good, didnt QA though

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -324,7 +324,8 @@ const SelectedEntriesModalContent = ({
         }),
       });
 
-      setPublishedCount(res.data.count);
+      // TODO fix for V5
```
joshuaellis, line 327: .... this is v5?

packages/plugins/i18n/admin/src/components/BulkLocaleActionModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+type Status = Modules.Documents.Params.PublicationStatus.Kind | 'modified';
+
+/* -------------------------------------------------------------------------------------------------
+ * EntryValidationText
+ * -----------------------------------------------------------------------------------------------*/
+
+const TypographyMaxWidth = styled(Typography)`
+  max-width: 300px;
+`;
```
joshuaellis, line 26: max width can be applied to the component, you dont need this.

packages/plugins/i18n/admin/src/components/BulkLocaleActionModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+  if (validationErrors) {
+    const validationErrorsMessages = Object.entries(validationErrors)
+      .map(([key, value]) => `${key}: ${formatMessage(value)}`)
+      .join(' ');
+
+    return (
+      <Flex gap={2}>
+        <CrossCircle fill="danger600" />
+        <Tooltip description={validationErrorsMessages}>
```
joshuaellis, line 44:
```suggestion
        <Tooltip label={validationErrorsMessages}>
```

packages/plugins/i18n/admin/src/components/BulkLocaleActionModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+  headers,
+  rows,
+  localesMetadata,
+  onClose,
+  validationErrors = {},
+}: BulkLocaleActionModalProps) => {
+  const { formatMessage } = useIntl();
+  const navigate = useNavigate();
+
+  const selectedRows: LocaleStatus[] = useTable<LocaleStatus[]>(
```
joshuaellis, line 128:
```suggestion
  const selectedRows = useTable<LocaleStatus[]>(
```

packages/plugins/i18n/admin/src/components/BulkLocaleActionModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+                      label={formatMessage(
+                        {
+                          id: getTranslation('Settings.list.actions.edit'),
+                          defaultMessage: 'Edit {name} locale',
+                        },
+                        {
+                          name: locale,
+                        }
+                      )}
+                      icon={<Pencil />}
```
joshuaellis, line 238: i'd make this `children` instead. we remove the `icon` prop in the next DS beta

packages/plugins/i18n/admin/src/components/BulkLocaleActionModal.tsx
```diff
@@ -0,0 +1,253 @@
...
+                      </Status>
+                    </Box>
+                  </Table.Cell>
+                  <Table.Cell>
+                    <EntryValidationText validationErrors={error} status={status} />
+                  </Table.Cell>
+                  <Table.Cell>
+                    <IconButton
+                      onClick={() => {
+                        navigateToLocale(locale);
```
joshuaellis, line 227: why a button not an `anchor`? if we're navigating somewhere.

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+    {
+      label: formatMessage({
+        id: getTranslation('CMEditViewBulkLocale.publication-status'),
+        defaultMessage: 'Publication Status',
+      }),
+      name: 'publication-status',
+    },
+  ];
+
+  const [rows, validationErrors] = React.useMemo(() => {
```
joshuaellis, line 311: i find this whole block really hard to follow, can we add some comments to explain why we're doing things?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+    }
+
+    const rowsFromMeta: LocaleStatus[] = documentMeta?.availableLocales.map((doc) => {
+      const { locale, status } = doc;
+
+      return { locale, status };
+    });
+    rowsFromMeta.unshift({
+      locale: document?.locale ?? '',
+      status: document?.status ?? 'draft',
```
joshuaellis, line 323: is document undefined here? I think above you check it, so you probs dont need the optional chaining and fallbacks?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+        return document;
+      }
+
+      const validation = validate(document as Modules.Documents.AnyDocument);
+      if (validation !== null) {
+        errors[document.locale] = validation;
+      }
+
+      return document;
+    });
```
joshuaellis, line 344:
```suggestion
    const errors = allDocuments.reduce<Record<
      Modules.Documents.Params.Locale.StringNotation,
      Record<string, MessageDescriptor>
    >>((errs, document) => {
      if (!document) {
        return errs;
      }

      const validation = validate(document as Modules.Documents.AnyDocument);
      if (validation !== null) {
        errs[document.locale] = validation;
      }

      return errs;
    }, {});
```

I think this would be a lot clearer as a reduce function:

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+
+    return [rowsFromMeta, errors];
+  }, [document, documentMeta?.availableLocales, validate]);
+
+  const localesToPublish = selectedRows
+    .filter(
+      (selectedRow) =>
+        selectedRow.status !== 'published' &&
+        !Object.keys(validationErrors || {}).includes(selectedRow.locale)
```
joshuaellis, line 353:
```suggestion
        !Object.keys(validationErrors).includes(selectedRow.locale)
```

validationErrors is always an object i think?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+        errors[document.locale] = validation;
+      }
+
+      return document;
+    });
+
+    return [rowsFromMeta, errors];
+  }, [document, documentMeta?.availableLocales, validate]);
+
+  const localesToPublish = selectedRows
```
joshuaellis, line 349: this is optional, but instead of doing a filter, then a map, you could just do a `reduce`. I don't think it'll be an issue here, because no ones gonna have 1000s of locales....

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+        },
+      });
+
+      if ('error' in bulkPublishRes) {
+        toggleNotification({
+          type: 'danger',
+          message: formatAPIError(bulkPublishRes.error),
+        });
+        return;
+      }
```
joshuaellis, line 413: Is this not handled in the `usedocumentActions` hook?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -226,6 +237,299 @@ const DeleteLocaleAction: DocumentActionComponent = ({
...
+          setIsConfirmationOpen(false);
+        },
+        type: 'dialog',
+        title: formatMessage({
+          id: getTranslation('actions.publish.dialog.title'),
+          defaultMessage: 'Confirmation',
+        }),
+        content: (
+          <Flex direction="column" alignItems="center" gap={2}>
+            <WarningCircle width="24px" height="24px" fill="danger600" />
```
joshuaellis, line 464:
```suggestion
            <WarningCircle width="2.4rem" height="2.4rem" fill="danger600" />
```

rem, not px 👍🏼

joshuaellis (commented):
some refinement points 🧽

joshuaellis (approved):

PR #20259, opened 2024-05-03
chore(content-releases): releases migration to v5

Feranchz:
### What does it do?

Migrates and apply new changes on the Content Releases plugin for v5.

### Work to do

- [x] Fix test cases
- [x] Validate releases with entries with multiple relations and if validation is still working
- [x] Validate db lifecycles and how affected are releases (entry is deleted, discarded, etc)


packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -29,50 +28,112 @@ const releaseController = {
...
-        return {
-          ...releaseData,
+    if (isEntryAttachedBoolean) {
+      const releases = await releaseService.findMany({
+        where: {
           actions: {
-            meta: {
-              count: actions.count,
+            target_type: contentTypeUid,
+            target_id: entryId,
```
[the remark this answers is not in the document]
Feranchz: That's a good point and probably is going to be a problem 🤔

I'm using the entryId because I think polymorphic relations with documentId are not implemented yet, right? So I don't see a solution 🤔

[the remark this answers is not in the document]
Marc-Roig: we might need to find alternatives, relations will allways be stored using entry ids, it's just that we expose docId relations to make things consistent on the api.

I guess, at the moment, discarding a draft would result in removing that one from the release 🤔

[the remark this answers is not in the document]
Feranchz: Yes, we are using a db lifecycle to listen to that "event" and remove deleted entries from releases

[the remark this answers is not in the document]
Marc-Roig: well , I guess discarding the documetn would just unlist it from the release then.
It might be a good discussion with Yannis to check if this is an issue and if we want to work on it on the future , wdyt?

packages/core/content-releases/server/src/utils/index.ts
```diff
@@ -25,14 +25,14 @@ export const getPopulatedEntry = async (
 };
 
 export const getEntryValidStatus = async (
```
Marc-Roig, line 27: cc @jhoward1994 just in case, just in case you want to handle publishing validations backend side this might be useful.

joshuaellis (commented):
some minor feedback on the FE tweaks, you might not have know those things were available :)

packages/core/content-releases/server/src/controllers/release-action.ts
```diff
@@ -81,9 +81,11 @@ const releaseActionController = {
     });
     const query = await permissionsManager.sanitizeQuery(ctx.query);
 
+    query.sort = query.groupBy === 'action' ? 'type' : query.groupBy;
+    delete query.groupBy;
```
Marc-Roig, line 85: Probably not something related to this PR, but this should contain a yup validation. I don't think the sanitizeQuery is sanitizing non existing filters like "groupBy"

packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -9,17 +9,16 @@ import type {
...
 } from '../../../shared/contracts/releases';
 import type { UserInfo } from '../../../shared/types';
-import { getService } from '../utils';
+import { getService, getEntryId } from '../utils';
 
 type ReleaseWithPopulatedActions = Release & { actions: { count: number } };
 
 const releaseController = {
-  async findMany(ctx: Koa.Context) {
+  async findByDocumentAttached(ctx: Koa.Context) {
```
Marc-Roig, line 21: could you add some comment or explanation of what's the responsability of this controller? 🤔 
After 2 weeks of being of, I forgot a bit the logic, and I don't quite see what `findByDocumentAttached`  means.

Maybe what confuses me is the "Attached" word

packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -29,50 +28,110 @@ const releaseController = {
...
+                $notIn: relatedReleases.map((release: any) => release.id),
+              },
+            },
+            {
+              actions: null,
+            },
+          ],
+          releasedAt: {
+            $null: true,
+          },
```
Marc-Roig, line 92: I believe these queries could be inside the service itself.

releases.findEntryReleases({id, populate...})
releases.findNotInEntryReleases({id, ...))

You could also reduce logic here if you do something like:
releases.findDocumentReleases({documentId, locale, populate...})
releases.findNotInDocumentReleases({documentId, locale, ...))

packages/core/content-releases/server/src/controllers/release.ts
Marc-Roig: Also, it feels off that the endpoint is called findByDocumentAttached, but you can send a parameter to find by documents not attached 🤔

packages/core/content-releases/server/src/utils/index.ts
```diff
@@ -44,3 +44,27 @@ export const getEntryValidStatus = async (
...
+  if (documentId) {
+    const document = await strapi.documents(contentTypeUid).findOne({ documentId, locale });
+
+    return document?.id;
+  }
+
+  // documentId is not defined in singleTypes, so we want to get the first document
+  const documents = await strapi.documents(contentTypeUid).findMany({ locale });
+
+  return documents[0].id;
```
Marc-Roig, line 69: Use the `.findFirst` instead, this might load 1M documents if they exist in the database.

Marc-Roig:
Don't know if it's something already fixed in the v5 branch , but it's a bit difficult for me to test strapi using multiple users atm. It's not possible for me to refresh the page when I have two tabs (one in incognito) with two different users.

packages/core/content-releases/server/src/controllers/release.ts
Feranchz: I'm not 100% sure about this but we can discuss it when you're back

It's true that we reduce the controller complexity but we increase the service complexity and we need to maintain and look up for breaking changes in the service but not in the controller. What I mean is: Users can use the releases service in plugins or other places so in my opinion service's API should be as simple as possible and all these functions are just different way of using the findMany that already exists in the service

packages/core/content-releases/admin/src/components/AddToReleaseAction.tsx
markkaylor, line 1: Could this file name match the convention and maybe describe what is and not what it does, something like `ReleaseActionModal`

packages/core/content-releases/admin/src/components/AddToReleaseAction.tsx
```diff
@@ -0,0 +1,299 @@
...
+      />
+    </Flex>
+  );
+};
+
+/* -------------------------------------------------------------------------------------------------
+ * AddToReleaseAction
+ * -----------------------------------------------------------------------------------------------*/
+
+const AddToReleaseAction: DocumentActionComponent = ({
```
markkaylor, line 164: Would it make more sense to call this `ReleaseActionForm` or with the other suggestion `ReleaseActionModalForm`?

packages/core/content-releases/admin/src/components/AddToReleaseAction.tsx
```diff
@@ -0,0 +1,299 @@
...
+  EmptyStateLayout,
+  LinkButton,
+  Field,
+} from '@strapi/design-system';
+import { Cursor } from '@strapi/icons';
+import { EmptyDocuments } from '@strapi/icons/symbols';
+import { isAxiosError } from 'axios';
+import { useFormik } from 'formik';
+import { useIntl } from 'react-intl';
+import { Link as ReactRouterLink, useParams } from 'react-router-dom';
```
markkaylor, line 25: Heads up! `useParams` is not used

packages/core/content-releases/admin/src/components/AddToReleaseAction.tsx
```diff
@@ -0,0 +1,299 @@
...
+            required
+            placeholder={formatMessage({
+              id: 'content-releases.content-manager-edit-view.add-to-release.select-placeholder',
+              defaultMessage: 'Select',
+            })}
+            name="releaseId"
+            onChange={(value) => onInputChange('releaseId', value)}
+            value={values.releaseId}
+          >
+            {releases?.map((release: any) => (
```
markkaylor, line 137: Doesn't the type come throught? I think you might be able to get rid of this `any`

packages/core/content-releases/admin/src/components/ReleasesPanel.tsx
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  return {
+    title: formatMessage({
+      id: 'content-releases.plugin.name',
+      defaultMessage: 'Releases',
+    }),
+    content: (
+      <Flex direction="column" alignItems="stretch" gap={3} width="100%">
+        {releases?.map((release: any) => (
```
markkaylor, line 63: I think something is off here. The type comes through so we shouldn't need `any`. Removing it then gives an error and looking at the type, `actions` is not an array:

https://github.com/strapi/strapi/blob/c9e680cf69a00f2ccbe5984c3618e5cb7f8b9da8/packages/core/content-releases/shared/contracts/releases.ts#L24-L27

packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
```diff
@@ -122,17 +122,19 @@ interface EntryValidationTextProps {
...
-  const { validate } = unstable_useDocument(
+
+  const { validate, isLoading } = unstable_useDocument(
     {
       collectionType: schema?.kind ?? '',
       model: schema?.uid ?? '',
     },
     {
-      skip: !schema,
+      skip: true,
```
markkaylor, line 132: 🤔 what does skip do? Should it always be true?

packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
```diff
@@ -122,17 +122,19 @@ interface EntryValidationTextProps {
...
     }
   );
 
-  const errors = validate(entry) ?? {};
+  // Validate needs to make a init query, so we need to wait for it to be ready
+  const errors = !isLoading ? validate(entry) ?? {} : {};
```
markkaylor, line 137: It's a code style thing but I think it might be more readable with something like:

```tsx
  if (isLoading) {
    return null;
  }

  const errors = validate(entry) ?? {};
  ```
or
```tsx
const errors = (!isLoading && validate(entry)) ?? {};
```

packages/core/content-releases/server/src/bootstrap.ts
```diff
@@ -52,10 +52,10 @@ export const bootstrap = async ({ strapi }: { strapi: Core.Strapi }) => {
        * so we need to fetch them before deleting the entries to save the ids on our state
        */
       async beforeDeleteMany(event) {
-        const { model, params } = event;
+        const model = strapi.getModel(event.model.uid as any);
```
markkaylor, line 55: Is typecasting as the expected type better than any 🤔 ? Maybe this should be as UID.Schema or whatever it is. Also wondering if this type is accurate or if it itself should be using UID.Schema

packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -29,50 +33,110 @@ const releaseController = {
...
-    } else {
-      const query: GetReleases.Request['query'] = await permissionsManager.sanitizeQuery(ctx.query);
-      const { results, pagination } = await releaseService.findPage(query);
+    if (!entryId) {
+      throw new errors.NotFoundError('Entry not found');
+    }
 
-      const data = results.map((release: ReleaseWithPopulatedActions) => {
-        const { actions, ...releaseData } = release;
+    const isEntryAttachedBoolean =
```
markkaylor, line 44: Personally `isXXX` convention is sufficient for indicating a boolean, so the Boolean part of this variable name seems redundant. 

The type that comes through ends up being any... Maybe we could wrap JSON.parse in Boolean 🤷‍♂️

packages/core/content-releases/server/src/controllers/validation/release-action.ts
```diff
@@ -4,8 +4,9 @@ const RELEASE_ACTION_SCHEMA = yup.object().shape({
   entry: yup
     .object()
     .shape({
-      id: yup.strapiID().required(),
+      documentId: yup.strapiID(),
```
markkaylor, line 7: It's not required?

packages/core/content-releases/server/src/utils/index.ts
```diff
@@ -44,3 +54,27 @@ export const getEntryValidStatus = async (
...
+  if (documentId) {
+    const document = await strapi.documents(contentTypeUid).findOne({ documentId, locale });
+
+    return document?.id;
+  }
+
+  // documentId is not defined in singleTypes, so we want to get the first document
+  const document = await strapi.documents(contentTypeUid).findFirst({ locale });
+
+  return document?.id;
```
markkaylor, line 79: Not really sure how this function is used but just curious if it is useful to return undefined or would it be more useful to throw an error?

packages/core/content-releases/server/src/controllers/validation/release-action.ts
Feranchz: Is not required because with SingleTypes you can create a ReleaseAction only with the ContentTypeUID and the locale

packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
Feranchz: This is because how `useDocument` hook works. You can use this hook with documentId to get more details about an specific entry, but for this case we only want to get the validate function so you pass `skip: true` so it doesn't try to make any request to get more details about the document

packages/core/content-releases/admin/src/components/ReleasesPanel.tsx
Feranchz: Thank you, I was using the wrong type for this query response

packages/core/content-releases/server/src/controllers/release-action.ts
```diff
@@ -81,9 +81,17 @@ const releaseActionController = {
     });
     const query = await permissionsManager.sanitizeQuery(ctx.query);
```
Marc-Roig, line 82: is the groupBy actually sanitized? as it's not a query parameter that we normally use in strapi.

Would it work the same if the wsanitization is done after replacing the `groupBy` with  the `sort`?

packages/core/content-releases/server/src/controllers/release.ts
Marc-Roig: I think it's a bit confusing for me given the name of the controller, calling a findByDocumentAttached, sending a `hasEntryAttached: false` looks odd.

Also, its mixing document & entry language in here which makes it a bit more complicated imo

packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -29,50 +33,106 @@ const releaseController = {
...
+            },
+            {
+              actions: null,
+            },
+          ],
+          releasedAt: {
+            $null: true,
+          },
         },
       });
```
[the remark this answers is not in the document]
Feranchz: This is because of the logic expected by the query engine.

Let's say you want to get the releases where documentId `abcd` is not attached 

If I say something like:
```
strapi.db("releases").query({
  where: {
    actions: {
      entryDocumentId: not abcd,
    }
  }
})
```

This could return releases where the documentId `abcd` is attached, because if you have two documents in a release, one is `abcd` and the other is anything, that other makes the condition pass

packages/core/content-releases/server/src/bootstrap.ts
```diff
@@ -103,46 +54,11 @@ export const bootstrap = async ({ strapi }: { strapi: Core.Strapi }) => {
           });
```
Marc-Roig, line 54: this lifecycle is blocking , is it necessary? Maybe we can make this asynchronous so it doesnt slow down deleeteMany ops

packages/core/content-releases/server/src/bootstrap.ts
```diff
@@ -1,93 +1,44 @@
...
 
             await strapi.db.query(RELEASE_ACTION_MODEL_UID).deleteMany({
               where: {
-                target_type: model.uid,
-                target_id: {
-                  $in: (entriesToDelete as Array<{ id: Data.ID }>).map((entry) => entry.id),
-                },
+                contentType: model.uid,
+                locale: where.locale ?? null,
+                ...(where.documentId && { entryDocumentId: where.documentId }),
```
Marc-Roig, line 41: could this create a recursive infinite loop? (deleteMany triggering this same lifecycle)

packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -1,88 +1,137 @@
...
+      ability: ctx.state.userAbility,
+      model: RELEASE_MODEL_UID,
+    });
+
+    await permissionsManager.validateQuery(ctx.query);
+
+    const releaseService = getService('release', { strapi });
+
+    const query: GetReleases.Request['query'] = await permissionsManager.sanitizeQuery(ctx.query);
+    const { results, pagination } = await releaseService.findPage(query);
```
Marc-Roig, line 100: shoulndt this endpoint  also need a yup validation?

packages/core/content-releases/server/src/middlewares/documents.ts
```diff
@@ -0,0 +1,122 @@
...
+    getService('release', { strapi }).updateReleaseStatus(release.id);
+  }
+};
+
+const deleteActionsOnDelete: Middleware = async (ctx, next) => {
+  if (ctx.action !== 'delete') {
+    return next();
+  }
+
+  if (!ctx.contentType.options?.draftAndPublish) {
```
Marc-Roig, line 66: could you use the following `@ strapi/utils ` utility to check this?
```suggestion
  import { contentTypes } from "@ strapi/utils"

  if (contentTypes.hasDraftAndPublish(ctx.contentType) {
```

Marc-Roig (approved):
overall approach lgtm ! :) only added minor comments

packages/core/content-releases/server/src/bootstrap.ts
Feranchz: It's a good point, honestly never thought about that. But you can see at line 15 that this subscriber is added only for D&P ContentTypes, and ReleaseAction doesn't has D&P, so the subscriber is not registered for it

packages/core/content-releases/server/src/controllers/release.ts
Feranchz: I don't think is necessary because this query doesn't handle any custom logic and the ctx.query is already sanitized with the permissionsManager, I tried some basis cases and it's working

joshuaellis (changes requested):
There's some strange decisions in the FE i think we need to discuss.

packages/core/content-manager/admin/src/services/api.ts
```diff
@@ -9,6 +9,8 @@ const contentManagerApi = adminApi.enhanceEndpoints({
     'InitialData',
     'HistoryVersion',
     'Relations',
+    'Release',
+    'ReleaseAction',
```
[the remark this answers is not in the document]
Feranchz: The problem is that changes to documents could cause stale release data. For example, if you have an invalid entry that causes a release to have a "Blocked" status, and then you go to the Edit View and update the entry to fit the validation rules, this changes the release status to "Ready". This is the scenario I want to address here.

packages/core/content-releases/admin/src/index.ts
```diff
@@ -1,11 +1,14 @@
 import { PaperPlane } from '@strapi/icons';
 
-import { CMReleasesContainer } from './components/CMReleasesContainer';
+import { ReleaseActionModalForm } from './components/ReleaseActionModal';
+import { Panel as ReleasesPanel } from './components/ReleasesPanel';
+// import { addColumnToTableHook } from './components/ReleaseListCell';
```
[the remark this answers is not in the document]
Feranchz: This is for the "Releases" column in the list view, is not implemented yet and I would like to do it in an upcoming PR. I left the code commented here because I think it should not change too much at this file.

packages/core/content-manager/admin/src/services/api.ts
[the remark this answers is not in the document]
joshuaellis: you should be doing it this way in that case i think? https://redux-toolkit.js.org/rtk-query/api/created-api/code-splitting#enhanceendpoints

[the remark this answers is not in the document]
Feranchz: I think I didn't explain the first question 😅 I'm doing that I think, for example I invalidate the releases tags [here](https://github.com/strapi/strapi/blob/62ce06180fe9a772eaeb3d43d238b26644f39f7c/packages/core/content-manager/admin/src/services/documents.ts#L338) on the update document.

Adding the types here is to avoid a Typescript error that not allowed me to invalidate the releases tags because they are not known. [Here](https://github.com/reduxjs/redux-toolkit/issues/1510#issuecomment-1546751373) is a discussion about this on the RTK's repo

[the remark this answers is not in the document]
joshuaellis: you're invalidating it in the wrong place, because thats the endpoint definition in the domain of the content-manager, instead your `enhanceEndpoints` object on the CR service should look like this:

```ts
  .enhanceEndpoints({
    addTagTypes: ['Release', 'ReleaseAction', 'EntriesInRelease'],
    endpoints: {
      updateDocument: {
        invalidatesTags: [
          { type: 'Release', id: 'LIST' },
          { type: 'ReleaseAction', id: 'LIST' },
        ],
      },
      deleteDocument: {
        invalidatesTags: [
          { type: 'Release', id: 'LIST' },
          { type: 'ReleaseAction', id: 'LIST' },
        ],
      },
      deleteManyDocuments: {
        invalidatesTags: [
          { type: 'Release', id: 'LIST' },
          { type: 'ReleaseAction', id: 'LIST' },
        ],
      },
      discardDocument: {
        invalidatesTags: [
          { type: 'Release', id: 'LIST' },
          { type: 'ReleaseAction', id: 'LIST' },
        ],
      },
    },
  })
```

where you add the invalidate tags from this side, keeping the release logic locked away. This is the way i'd expect plugin authors to have their data invalidated when a document is published, for example.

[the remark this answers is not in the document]
Feranchz: thanks! This works way better

joshuaellis (approved):
⭐  nice work Fernando!

PR #20261, opened 2024-05-03
fix: create locale column when migrating from v4

Marc-Roig:
### What does it do?
Migrating from v4 content types with i18n disabled resulted in the following error:
image

The cause was the `locale` column not being in the database for those content types. It is assumed , in v5, that the locale column will be present even if i18n is disabled on a content type.

This fix proposes adding a locale column in the discard-draft migration,  similarly to how we add the document_id in another migration.

### DIFF
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -50,9 +50,11 @@ async function* getBatchToDiscard({
 }
 
 /**
- * This migration makes use of the document service.
- * This one assumes the `locale` column exists. In v4, that was not the case
- * for content types with i18n disabled, so we need to create it for those cases.
+ * This migration makes use of the document service,
+ * which assumes the `locale` column always exists.
+ * In v4, that was not the case, for content types with i18n disabled.
+ *
+ * This function creates the `locale` column if it doesn't exist.
  */
 const createLocaleColumn = async (db: Knex, tableName: string) => {
   await db.schema.alterTable(tableName, (table) => {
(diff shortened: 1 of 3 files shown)


innerdvations (approved):
I don't like it because in theory we should be letting the db sync handle this, but the whole point is we have to do these migrations before the db sync, so... I don't see any alternative except to manually add it like this. Thanks! :)

Marc-Roig:
same.. we did something similar for the document id, but at this point if we want to use the document service we need to include the locale column somehow.

I also was not fully convinced this migration was the right place to do this 🤔 but as it's the only place it's necessary I put it there.

Will ask Marion to include this in a blitz session too, I think it's important we validate it works.

alexandrebodin (commented):
Looks good, but should we even migrate if the column doesn't exist yet ?

Marc-Roig:
@alexandrebodin  In this case we need to migrate if the content type has D&P enabled, regardless of its i18n configuration. And by using the discardDraft of the document service we are tied to have content types with an available locale column

alexandrebodin:
> @alexandrebodin In this case we need to migrate if the content type has D&P enabled, regardless of its i18n configuration. And by using the discardDraft of the document service we are tied to have content types with an available locale column

Ho yes ok , Should we just have a locale migration to add that column everywhere and order it earlier in the list ?

Marc-Roig:
> Ho yes ok , Should we just have a locale migration to add that column everywhere and order it earlier in the list ?
I thought something similar and yes.. makes sense, will add one in the database after the document id one

Marc-Roig:
@markkaylor needs this change asp, @alexandrebodin  let me know if you would like to add additional changes and I can create a new pr!

PR #20263, opened 2024-05-03
fix back button on ListSettingsView

xdeelord24:
### What does it do?

image

### Why is it needed?

The back button is not working, and I need it to facilitate easy back history.

### How to test it?

The back button does not work when configuring the view of any collection or single type on ListSettingsView.

### Related issue(s)/PR(s)

There is no related issue

### DIFF
--- a/packages/core/admin/admin/src/content-manager/pages/ListSettingsView/index.jsx
+++ b/packages/core/admin/admin/src/content-manager/pages/ListSettingsView/index.jsx
@@ -9,18 +9,11 @@ import {
   Layout,
   Main,
 } from '@strapi/design-system';
-import {
-  Link,
-  useFetchClient,
-  useNotification,
-  useQueryParams,
-  useTracking,
-} from '@strapi/helper-plugin';
+import { Link, useFetchClient, useNotification, useTracking } from '@strapi/helper-plugin';
 import { ArrowLeft, Check } from '@strapi/icons';
 import isEqual from 'lodash/isEqual';
 import upperFirst from 'lodash/upperFirst';
 import PropTypes from 'prop-types';
-import { stringify } from 'qs';
 import { useIntl } from 'react-intl';
 import { useMutation } from 'react-query';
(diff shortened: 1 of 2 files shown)


joshuaellis:
Can you elaborate on why the back-button doesn't currently work for you?

xdeelord24:
> Can you elaborate on why the back-button doesn't currently work for you?

When I create new project on strapi@latest as V4, I was editing some collections and I was gonna change the view of my viewList data using "Configure the view" , after editing the fields I want to click the back button and suddenly it is not working, I also tried the previous version of V4 and it is all not working. You can try the latest V4.
This is the button that I'm fixing. 
image

It works when I change the code into this PR

joshuaellis (changes requested):
I think the solution is actually to change L187 from `to` to `pathname` as per the types of `Link` from react-router-dom. If we can do that change, which is a smaller surface area to change, we can accept this :)

packages/core/admin/admin/src/content-manager/pages/ListSettingsView/index.jsx
```diff
@@ -183,23 +171,16 @@ export const ListSettingsView = ({ layout, slug }) => {
             navigationAction={
               <Link
                 startIcon={<ArrowLeft />}
-                to={{
-                  to: `/content-manager/${kind}/${uid}`,
```
joshuaellis, line 187: Sorry there was some confusion, if you dont make the changes you've made and just change this property to `pathname` it should work as you expect it too?

xdeelord24: it works for me

xdeelord24: it works, I'll PR it

joshuaellis (approved):

joshuaellis:
Looks like prettier is failing. You need to format that file 👍🏻

PR #20266, opened 2024-05-06
feat: unlist core packages from plugin list

Marc-Roig:
### What does it do?

We were listing all internally installed plugins on the admin plugins page.

This PR removes the ones that are pre-installed for the user (core plugins), it was confusing as users were not able to uninstall them, or even see them listed in their "package.json" files.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
 <td>

</table>

Resolves: Content-722


packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -154,12 +154,18 @@ export default {
...
 
-    const plugins = Object.entries(enabledPlugins).map(([key, plugin]: any) => ({
-      name: plugin.info.name || key,
-      displayName: plugin.info.displayName || plugin.info.name || key,
-      description: plugin.info.description || '',
-      packageName: plugin.info.packageName,
-    }));
+    // List of core plugins that are always enabled,
+    // and so it's not necessary to display them in the plugins list
+    const CORE_PLUGINS = ['content-manager', 'content-type-builder', 'email', 'upload', 'i18n'];
```
remidej, line 159: I think the official term now is core packages

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -154,12 +154,18 @@ export default {
...
+
+    const plugins = Object.entries(enabledPlugins)
```
remidej, line 161: It may not be feasible but I wonder if core packages should be enabled in `strapi.config.get('enabledPlugins')` at all?

derrickmehaffy:
Shouldn't review workflows and releases be delisted also?

Feranchz:
> Shouldn't review workflows and releases be delisted also?

I was thinking the same, and releases is also only for EE editions... I know technically is always "installed" but doing nothing for CE edition but always felt weird for me show it when is really not doing anything

derrickmehaffy:
> > Shouldn't review workflows and releases be delisted also?
> 
> I was thinking the same, and releases is also only for EE editions... I know technically is always "installed" but doing nothing for CE edition but always felt weird for me show it when is really not doing anything

Indeed yeah I agree, that code shouldn't even be loaded on CE to begin with.

Marc-Roig:
good point! will add those to the list

packages/core/admin/server/src/controllers/admin.ts
Marc-Roig: Originally I was going this route, of not including them in the enabled plugins config,  but those are also used to load the pacakges itself, removing them from that config would mean refactoring how we load plugins, which felt like a lot of work

packages/core/admin/server/src/controllers/admin.ts
Marc-Roig: shouldn't we rename everything in here "packages"? even the configuration references plugins, and it feels "off" to reference packages in this const variable but reference plugins in the rest of the controller.

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -154,12 +154,26 @@ export default {
...
 
-    const plugins = Object.entries(enabledPlugins).map(([key, plugin]: any) => ({
-      name: plugin.info.name || key,
-      displayName: plugin.info.displayName || plugin.info.name || key,
-      description: plugin.info.description || '',
-      packageName: plugin.info.packageName,
-    }));
+    // List of core plugins that are always enabled,
+    // and so it's not necessary to display them in the plugins list
+    const CORE_PLUGINS = [
```
joshuaellis, line 159: This list will need to be maintained with the same list as https://github.com/strapi/strapi/blob/v5/main/packages/core/core/src/loaders/plugins/get-enabled-plugins.ts shall we write that?

Marc-Roig: good point, will add a comment

PR #20267, opened 2024-05-06
fix(history): adjust condition to display missing relations

markkaylor:
### What does it do?

- Fix condition from return "No relations" for missing relations

### Why is it needed?

- Missing relations are always represented as no relations

### How to test it?

- Go to a content-type with relations like kitchensink
- Create some entries to relate to like Tag
- Attach the relations to kitchensink and create some history versions
- Now go delete some relations
- Access previous versions where the relations existed
- Make sure the correct message is displayed

Relation should display:
- No relations (including no missing relations) should display "No relatins"
- Relations + missing relations should display the existing relations + missing relation message
- Missing relation messages should display the number of missing relations if there are several that are missing.


remidej (approved):
tested, works as expected

PR #20268, opened 2024-05-06
Rework create-strapi-xxx

alexandrebodin:
### What does it do?

- Rework the cli UX and rework prompt
- Add package manager detection
- Support pnpm
- Remove create-strapi-starter

### Why is it needed?

- v5

### How to test it?

./node_modules/.bin/create-strapi-app


packages/cli/create-strapi-starter/src/create-strapi-starter.ts
```diff
@@ -31,7 +31,6 @@ program
   .version(packageJson.version)
   .arguments('[directory], [starter]')
   .option('--use-npm', 'Force usage of npm instead of yarn to create the project')
-  .option('--debug', 'Display database connection error')
```
Marc-Roig, line 34: I assume a database error would still be displayed without the debug tag right? But the debug logs would not appear.

packages/generators/app/src/index.ts
```diff
@@ -80,17 +149,27 @@ export const generateNewApp = (projectDirectory: string, options: Partial<NewOpt
...
+
+    const scopeExtension = await parseDatabaseArguments(options);
+    merge(scope, scopeExtension);
+
+    // create a project with full list of questions
+    return await createProject(scope);
+  } catch (error) {
+    if (!(error instanceof Error)) {
+      throw error;
+    }
```
Marc-Roig, line 165: why not tracking errors that are not of Error type?

packages/cli/create-strapi-starter/src/create-strapi-starter.ts
alexandrebodin: we removed the db connection test a long time ago this is just dead code it's not doign anythign :)

packages/generators/app/src/index.ts
alexandrebodin: ts is complaining if the type isn't and instance of error, it's just a safety measure we should always receive errors anyway but still need a runtime check to make ts happy

Marc-Roig:
rework looks good to me! let me know when this is ready to be reviewed and I can QA :) 

Also, does it make sense to do some tests on different options to validate it works as intended?

Marc-Roig (approved):
Did QA with yarn and pnpm, nice ✨ 

- Could you document in the generators README this can be tested by using `./node_modules/.bin/create-strapi-app` ?


Also, I am going to create a ticket, once creating a new strapi app, I can only see these pages:
image

I need to reload the server to see the rest of pages.

packages/generators/app/src/utils/git.ts
```diff
@@ -27,8 +27,15 @@ export default async function tryGitInit(rootDir: string) {
 
     await execa('git', ['init'], { stdio: 'ignore', cwd: rootDir });
 
+    await execa('git', ['add', '.'], { stdio: 'ignore', cwd: rootDir });
+    await execa('git', ['commit', '-m', 'Initial commit from Strapi'], {
```
Convly, line 31: What if the users' git server expect signed commits or other configs? (we used to required that on Github, hence the question).

Having to rewrite the commits history would create some friction, wouldn't it? Or do we expect this to be a niche scenario, thus only supporting the main usecase by default?

packages/generators/app/src/index.ts
```diff
@@ -47,50 +43,50 @@ export const generateNewApp = (projectDirectory: string, options: Partial<NewOpt
...
+    await trackUsage({ event: 'willCreateProject', scope });
+
+    // create a project with full list of questions
+    return await createProject(scope);
+  } catch (error) {
+    if (!(error instanceof Error)) {
+      throw error;
+    }
+
+    console.log(`\n${error.message}\n`);
```
Convly, line 84: If we don't want to ignore errors thrown as strings, we could do the following

```suggestion
    let e: Error;
 
    if (typeof error === 'string') {
        e = new Error(e);
    }
    else if (error instanceof Error) {
        e = error;
    }
    else {
        throw error
    }
    
    console.log(`\n${e.message}\n`);
```
but that might be a bit too much, as you prefer

packages/generators/app/src/utils/usage.ts
```diff
@@ -12,43 +11,6 @@ function addPackageJsonStrapiMetadata(metadata: Record<string, unknown>, scope:
...
-    await sentry.flush();
-  } catch (err) {
-    /** ignore errors */
-    return Promise.resolve();
-  }
-}
-
-async function captureError(message: string) {
-  try {
-    sentry.captureMessage(message, 'error');
```
Convly, line 27: Are we removing sentry events completely? Or did I miss them being included somehere else?

Convly (commented):
Some questions, but it works well :)

README.md
```diff
@@ -59,18 +59,18 @@ Install Strapi with this **Quickstart** command to create a Strapi project insta
...
 ```
 
 **or**
 
-- (Use npm/npx to install the Strapi project.)
+- (Using npx to install the Strapi project.)
 
 ```bash
-npx create-strapi-app my-project --quickstart
+npx create-strapi@latest
```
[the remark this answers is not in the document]
alexandrebodin: yeah that avoids the caching issues

packages/cli/create-strapi-app/src/create-strapi-app.ts
```diff
@@ -1,32 +1,33 @@
...
-  'dbusername',
-  'dbpassword',
-  'dbssl',
-  'dbfile',
-];
-
 command
   .version(packageJson.version)
   .arguments('[directory]')
-  .option('--no-run', 'Do not start the application after it is created')
```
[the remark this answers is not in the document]
alexandrebodin: it's to separate the concerns better. before it ran the app everytime now it's only on --quickstart (which we will remove eventually)  I don't really see the point of having --quickstart --no-run, just don't use --quickstart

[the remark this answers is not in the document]
Convly: It was useful when bootstraping app for testing purposes (when you only want to create the app + do some changes in the code). 

But it's not the end of the world, just a bit annoying

packages/generators/app/src/utils/git.ts
alexandrebodin: we were doing it for the starters and it's done in most tools nowdays I think. no strong opinions honestly

packages/generators/app/src/index.ts
alexandrebodin: If you throw a string you are a bad person :D more seriously it means those are probably not our own errors so I let thme bubble up.

packages/generators/app/src/utils/git.ts
Convly: Yep, after thinking a bit about this, I don't think it's worth taking this specific usecase into consideration. It's a nice to have that even github does 🤷 

Let's keep it

packages/generators/app/src/index.ts
Convly: make sense

packages/generators/app/src/utils/usage.ts
alexandrebodin: yes :) we don't use it anymore

Convly (approved):

PR #20273, opened 2024-05-07
chore(ui): make auth form RTL based on locale

zaaakher:
### What does it do?

This sets the auth form direction to RTL if the locale is RTL (Arabic, Persian, Urdu, etc)

image

### Why is it needed?

Better UI for RTL languages

### How to test it?

Run the getStarted example and see the login form

### Related issue(s)/PR(s)

n/a


zaaakher:
I realized I should do the same with the other auth forms so I moved `RTL_LOCALES` array into `constants.ts`

zaaakher:
And if you notice the lack of spacing in the "remember me" checkbox label when in RTL. I created this [issue](https://github.com/strapi/design-system/issues/1714#issue-2283566117) in the design-system repo

joshuaellis:
Unfortunately we don't currently support RTL and the proposed solution doesn't really extend to the rest of the codebase so it's not entirely practical to make the entire app support it this way. 

V2 of the design-system uses logical spacing which will aid us moving to RTL so in V5 we could look at officially supporting it, you can see the canny request here – https://feedback.strapi.io/developer-experience/p/support-rtl

PR #20274, opened 2024-05-07
chore(cm): changed the Configure the view link icon in the edit view

lucasboilly:
### What does it do?

It changes the Cog icon of the Configure the view link of the edit view to a ListPlus icon. 

### Why is it needed?

The edit view is currently the only page where we use the Cog icon to illustrate the Configure the view feature. In every other page we we use the ListPlus icon (my bad, the error comes from my mockups).

### How to test it?

It can be seen in the edit view of any entry.

### Related issue(s)/PR(s)

Not related to any issue or pull request.

https://github.com/strapi/strapi/assets/45385696/a13d90ab-fcbb-4533-a338-8222ccfa4a56

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
@@ -9,7 +9,7 @@ import {
   useQueryParams,
 } from '@strapi/admin/strapi-admin';
 import { Flex, SingleSelect, SingleSelectOption, Typography } from '@strapi/design-system';
-import { Cog, Pencil, Trash, WarningCircle } from '@strapi/icons';
+import { ListPlus, Pencil, Trash, WarningCircle } from '@strapi/icons';
 import { useIntl } from 'react-intl';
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -390,7 +390,7 @@ ConfigureTheViewAction.type = 'configure-the-view';
  * Because the icon system is completely broken, we have to do
  * this to remove the fill from the cog.
  */
-const StyledCog = styled(Cog)`
+const StyledListPlus = styled(ListPlus)`
```
joshuaellis, line 393: this can probably be removed and you directly call `ListPlus` because the svg default colors are `currentColor`

lucasboilly:
@joshuaellis I just dit it, I also removed the functions for the other icons and it seems to work well

joshuaellis:
I think your IDE is not configured correctly with the lint & prettier, you need to resolve that before we can merge :)

joshuaellis (approved):

PR #20275, opened 2024-05-07
feat(main-nav): Main nav refactoring, change links order

simotae14:
### What does it do?

It changes the main nav links order based on the specifications on the figma file

I added a position property, which is a numeric and can help us to reorder the plugins.
The links in the main nav are all sorted by position starting from 0 to the last one, the links with the same position number are sorted by link label
If you don't specify the link position then its position is in the middle of the main nav, before the plugin list

### Why is it needed?

To give more visibility to the most used links

### How to test it?

Just check the new main nav and check the figma file (you can find it in the jira ticket)

### Related issue(s)/PR(s)

CS-768


joshuaellis:
We're gonna look for a better, more complete & extensible way to manage all this.

joshuaellis:
> The behaviour is quite simple, if the plugin position number is low (lower than -3 for example) then the link appears as the first link in the main nav, if the plugin position number is high (higher than 5) then the link appears ad the last link in the main nav

I find this quite confusing, are you using the value to on all MenuItems? or just Plugins? IMO it should be all of them for clarity, which then feels like we should just have the items number 0-X, if you don't specify it gets rendered where it would normally?

If you set the CM plugin to position `0` it would see if it's alphabetically higher than `home` which it is, so it'd be first, you could apply `-1` to get the same result though.

Maybe i've misunderstood you and this is what you've implemented though?

simotae14:
You are right, my previous solution was a little bit confusing

We are sorting ALL the menu items (not just the plugins ), and the logic is this one
if two plugins has the same position then we sort them by alphabetic order of their names.
In our case to follow the ordering requested:
 - home
 - content manager
 - releases
 - media
 - content type builder
 - external installed plugins
 - plugin list
 - marketplace
 - documentation
 - settings
 I added the position property to every single link in this way
  - content manager => 1
 - content type builder => 5
 - documentation => 9
 - home => 0
 - marketplace => 8
 - media => 4
 - external plugins
 - plugin list => 7
 - releases => 2
 - settings => 10
 In this way if we want to create a new plugin and add it in the main nav we can't avoid to add the position property, in this way it will be placed before the plugin list (because by default I give the position number 6) otherwise you can specify a different position to place it wherever you want

Let me know if it seems better now, thanks

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -1,27 +1,47 @@
...
 
-const NewNavLinkBadge = styled(NavLink.Badge)`
+const sortLinks = (links: MenuItem[]) => {
+  return links.sort((a, b) => {
+    // if no position is defined, we put the link in the position of the external plugins, before the plugins list
+    const positionA = a.position ?? 6;
+    const positionB = b.position ?? 6;
+
+    if (positionA === positionB) {
+      return (a.intlLabel.defaultMessage ?? '') < (b.intlLabel.defaultMessage ?? '') ? -1 : 1;
```
joshuaellis, line 26: theres a couple of issues with this, I would firstly run `formatMessage` on all these links, you could do a map before hand, secondly you want to use `useCollator` because this won't account for certain characters in certain languages 👍🏼

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -1,27 +1,47 @@
...
 
 import { MainNav } from './MainNav/MainNav';
 import { NavBrand } from './MainNav/NavBrand';
 import { NavLink } from './MainNav/NavLink';
 import { NavUser } from './MainNav/NavUser';
 
-const NewNavLinkBadge = styled(NavLink.Badge)`
+const sortLinks = (links: MenuItem[]) => {
+  return links.sort((a, b) => {
+    // if no position is defined, we put the link in the position of the external plugins, before the plugins list
```
joshuaellis, line 21: great comment writing ⭐

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -45,115 +65,56 @@ const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) =
...
+                      ) : badgeContentNumeric ? (
+                        <NavLinkBadgeCounter
+                          label={badgeContentNumeric}
+                          backgroundColor="primary600"
+                          width="2.3rem"
+                          color="neutral0"
+                        >
+                          {badgeContentNumeric}
+                        </NavLinkBadgeCounter>
+                      ) : null}
```
joshuaellis, line 117:
```suggestion
                      {link?.lockIcon ? (
                        <NavLinkBadgeLock
                          label="locked"
                          textColor="neutral500"
                          paddingLeft={0}
                          paddingRight={0}
                        >
                          <Lock />
                        </NavLinkBadgeLock>
                      ) : null}
                      {badgeContentNumeric ? (
                        <NavLinkBadgeCounter
                          label={notificationCount}
                          backgroundColor="primary600"
                          width="2.3rem"
                          color="neutral0"
                        >
                          {notificationCount}
                        </NavLinkBadgeCounter>
                      ) : null}
```

joshuaellis: if people really want to add them both, let them, easier to read.

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -45,115 +65,56 @@ const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) =
...
-                </Flex>
-              );
-            })
-          : null}
-        {generalSectionLinks.length > 0
-          ? generalSectionLinks.map((link) => {
+        {listLinks.length > 0
+          ? listLinks.map((link) => {
               const LinkIcon = link.icon;
+              const badgeContentLock = link?.lockIcon ? <Lock /> : undefined;
```
joshuaellis, line 80:
```suggestion
```

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -45,115 +65,56 @@ const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) =
...
 
-              const badgeContent =
+              const badgeContentNumeric =
```
joshuaellis, line 82:
```suggestion
              const notificationCount =
```

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -1,27 +1,47 @@
...
+      return (a.intlLabel.defaultMessage ?? '') < (b.intlLabel.defaultMessage ?? '') ? -1 : 1;
+    } else if (positionA < positionB) {
+      return -1;
+    } else {
+      return 1;
+    }
+  });
+};
+
+const NavLinkBadgeCounter = styled(NavLink.Badge)`
```
joshuaellis, line 35: You should be able to inline this by passing the `textColor` prop?

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -1,27 +1,47 @@
...
   span {
     color: ${({ theme }) => theme.colors.neutral0};
   }
 `;
 
+const NavLinkBadgeLock = styled(NavLink.Badge)`
+  background-color: transparent;
+`;
```
joshuaellis, line 43: You should be able to inline this by passing the `backgroundColor` prop?

joshuaellis (changes requested):
Logic makes loads more sense ⭐ just need to tidy it up a bit

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: transparent is not in the possible backgroundColor for the Badge component, I have a type error

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: Same here, we have some issue passing the textColor (it doesn't overwrite the defined color of the Badge) because of that I needed to create a styled component

packages/core/admin/admin/src/components/LeftMenu.tsx
joshuaellis: does it work though? cause if so, we can fix the type 🤔

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: they occupy the same place, if we let them to add both then we will have some ui issues, and also the lock icon will be removed from the Main nav in future because we don t want to show the Releases purchase page in the main nav instead we are gonna move it to the Settings (I added the lock icon logic in past just for Releases)

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: same reason as before

packages/core/admin/admin/src/components/LeftMenu.tsx
simotae14: same reason as before

joshuaellis (approved):
I think theres improvements on the Badge component to be made if you want to add bug reports to the DS 👍🏼

joshuaellis:
@simotae14 can you double check if the docs need an update? I think we document adding plugin links quite well, and this is very valuable to individuals!

simotae14:
Yeah I suppose they need an update, I am gonna talk about it with Piwi on Monday

simotae14:
We need to add the position parameter here https://docs.strapi.io/dev-docs/api/plugins/admin-panel-api#menu-api

PR #20287, opened 2024-05-09
fix(chore): replace use-context-selector context with React Context

simotae14:
### What does it do?

Fix the issue we had after registration because the token was not updated correctly when we are redirected to the usecase page

### Why is it needed?

With use-context-selector context when we save the token in the registration handler, using the setToken coming from the useAuth hook, the value is not immediately updated to be used in the PrivateRoute component so the UseCasePage page is not served

### How to test it?

- clean the db
- create a new admin and flag and enable telemetry
- check if the user is redirected to the /usecase page instead of the /admin

### Related issue(s)/PR(s)

Related issue:  https://github.com/strapi/strapi/issues/20345


joshuaellis:
From your PR description, it's not clear how you came to this change, can you explain it in detail? It's a very "big" change to make across the app because we loose selection performance, so good to understand it thoroughly :)

joshuaellis (commented):
Great investigation. This would have been very tricky to debug, to do this properly we have two options:

1. remove `use-context-selector` as you have done, but also remove the selector function since it's pointless and verbose now.
2. use a unique version with the `Auth` feature i.e. `@radix-ui/react-context` that doesn't have the selector used and would also solve the issue, but we do risk this quirk coming back for us in the future, although I don't think it will considering how unique this is to the router acting before react has updated.
3. We could potentially pass the `redirect` query param a bit more and in `login` check if it's there and just do it if we dont actually need the user to login, https://github.com/strapi/strapi/blob/22f3123e946861ca798a41335ec80df6e68352e2/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx#L54

Let me know what you think!

simotae14:
Hey @joshuaellis , thanks a lot for your suggestions... In the end I decided to continue with solution 1 because it lets me use most of your code and just change some small parts in the code (remove the selectors).
If you are ok with this solution I can then create a new PR also in v5 to solve the same issue

joshuaellis:
> If you are ok with this solution I can then create a new PR also in v5 to solve the same issue

we'll need to check there aren't performance regressions, but i'm sure it'll be fine :)

joshuaellis (approved):
code changes look fine ✅

remidej (approved):
Did QA, merging now as Simone is OOO

PR #20290, opened 2024-05-09
feat(main-nav): Fix the tooltip position of the links

simotae14:
### What does it do?

Reverts the order of the NavLink.Tooltip and the NavLink.Link

### Why is it needed?

Because we had a strange behaviour in Safari, the position of the tooltip was starting in the middle of the icon link, wrapping the NavLink.Link with the NavLink.Tooltip solved the issue

### How to test it?

Open the CMS in Safare and check the tooltips position


Feranchz (approved):
🎉

PR #20295, opened 2024-05-10
chore: global backbutton on history page and fix e2e tests

jhoward1994:
### What does it do?

Uses the global app Back Button  on history pages
Fixes history E2E tests according to this change (I think they have been failing on any PR into v5/main)

### Why is it needed?

CI stability and consistent back button behaviour

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
+++ b/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
@@ -6,13 +6,14 @@ import {
   useQueryParams,
   useRBAC,
   Layouts,
+  BackButton,
 } from '@strapi/admin/strapi-admin';
-import { Button, Typography, Flex, Link } from '@strapi/design-system';
-import { ArrowLeft, WarningCircle } from '@strapi/icons';
+import { Button, Typography, Flex } from '@strapi/design-system';
+import { WarningCircle } from '@strapi/icons';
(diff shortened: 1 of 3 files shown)


jhoward1994:
Hey @markkaylor & @remidej 

Let me know what you think of this change? It seems something like this is definitely needed to improve the history e2e tests. It also seems nice to keep the back button behaviour consistent wherever we can but I don't have the full context on this feature and maybe theres something I missed?

joshuaellis:
I think if we want to use the global back button we need to add a `fallback` prop, the issue on CH is it's a complete overlay of the UI so if you refresh your page, you're stuck there. But it would be nice to make it consistent :)

remidej:
Hey @jhoward1994, thanks for looking into this. I don't think I can merge this because it's not the only issue with the history e2e tests, so I would rather bundle this with the other fixes I'm working on, to hopefully get to the point where the CI goes green on the fix PR.

I agree with Josh, we would need a fallback. I added one on my end and played with it, but in my experience the fact that clicking "back" after toggling through history versions only changes the active history version is a real problem. We could customize BackButton further to let it know if it should ignore query params, but I feel like that's too much complexity for a simple abstraction, when the alternative it to keep using a simple link.

jhoward1994:
Thanks @remidej , closing this in favour of your PR

PR #20303, opened 2024-05-13
chore: replace plugin cli with external strapi/sdk-plugin package

innerdvations:
### What does it do?

Removes the plugin:* commands and their contributor docs, because they have been moved to [@strapi/sdk-plugin
](https://github.com/strapi/sdk-plugin)

### Why is it needed?

It is moving out of the monorepo into its own project

### How to test it?

`yarn build` should still build all the plugins

### Related issue(s)/PR(s)

Fixes have already been made to solve the "Tracking Issue: Strapi 5 plugin template currently out-of-date/incorrect", now found at https://github.com/strapi/sdk-plugin/issues/1 they just need to be tested and released

DX-1406


Convly (changes requested):
Overall LGTM, but there are conflicts probably due the latest release

Convly (approved):

PR #20316, opened 2024-05-16
chore: update react&react-dom to 18.3

joshuaellis:
### What does it do?

* updates react & react-dom to latest version
* updates `@testing-library/X` packages to latest versions too

### Why is it needed?

* We now have visibility on deprecations coming in react19

### Related issue(s)/PR(s)

* resolves DX-1415


innerdvations (approved):
LGTM, the only testing I did was that Strapi builds and the admin generally works but not thouroughly

PR #20318, opened 2024-05-16
fix(admin): reset all redux toolkit cache on logout

joshuaellis:
### What does it do?

* refactors the admin to only use one createApi function for the entire app (plugins can use this too) so we can call `resetApiState` once on log out to clear the whole redux cache. When we migrate the rest of the app to use redux/toolkit it'll "just work"
* refactored some exports from the admin to satisfy typescript and tidy up the fetch work

### Why is it needed?

* If someone logs out we might not necessarily refetch data if all the params we pass to hooks match. This ensures that things are wiped clean.

### Related issue(s)/PR(s)

* resolves DX-1416


packages/core/admin/admin/src/services/api.ts
```diff
@@ -1,37 +1,22 @@
 import { createApi } from '@reduxjs/toolkit/query/react';
 
 import { TokenRegenerate } from '../../../shared/contracts/transfer';
```
Bassel17, line 3: this an be removed now

packages/core/admin/admin/src/services/api.ts
```diff
@@ -1,37 +1,22 @@
...
-import { fetchBaseQuery, type UnknownApiError } from '../utils/baseQuery';
+import { fetchBaseQuery } from '../utils/baseQuery';
 
+/**
+ * @public
+ * @description This is the reducl toolkit api for the admin panel, users
```
Bassel17, line 8:
```suggestion
 * @description This is the redux toolkit api for the admin panel, users
```

Bassel17 (changes requested):
Great job, thanks for tiding things up,
I left a few suggestions/comments, otherwise all looks good

packages/core/admin/admin/src/services/api.ts
joshuaellis: you dont want to use reducl? 😛

Bassel17 (approved):
looks great

Issue #20188, opened 2024-04-23
Unable to Publish Single Type in Non-Default Locale if It's Not Created in Default Locale


hiba-ijjig-nw:
### Required System Information

- Node.js version: v20
- Yarn version: v3.1.1
- Strapi version: v4.20.0
- Database: Postgres
- Operating System: Windows
- Project Language: Typescript

### Describe the Bug

A user cannot publish a single type in a specific locale if that single type hasn't been created in the default locale, or if the user lacks access to the default locale.

### Steps to Reproduce

1. Create a new single type.
2. Fill out and save the content in this non-default locale.
3. Try to publish the single type.

### Expected Behavior

The user should be able to publish a single type in a specific locale, even if it's not created in the default locale, or if the user lacks access to the default locale.

### Screenshots
In this example, `fr-FR` is the default locale.  
![image](https://github.com/strapi/strapi/assets/125470635/111d3876-639f-439f-ae93-56ddeba774d2)

1. Creating a single type and attempting to publish it in `en-FR`:
![image](https://github.com/strapi/strapi/assets/125470635/b9e0b256-c558-451c-882e-1728bb799dff)

publishing the single type gives a countDraftRelations error with a status code of 404 because it hasn't been created in the default locale (fr-FR).

2. If the single type is created in the default locale but the user lacks access to it:
![image](https://github.com/strapi/strapi/assets/125470635/bb52a485-32ab-41d9-b48f-0ae8b8bf529e)
In this case, `countDraftRelations` failed with a 403 because the user didn't have access to the default locale.

I belive that these errors occur because `countDraftRelations` checks the default locale instead of the current one.

alexandermirzoyan:
Same issue for me.
I have investigated that somehow when saving Content with non-default language in db it stores with default locale.
That's why during publish the system can't find non-default content from db.

alexandermirzoyan:
Seems this [issue](https://github.com/strapi/strapi/issues/19901) is about the same thing.

PR #20324, opened 2024-05-17
fix: fixing countDraftRelations to get locale as a prop

hiba-ijjig-nw:
### What does it do?

Modified countDraftRelations to retrieve the locale from parameters, allowing it to verify the single type for the specified query locale instead of defaulting to the locale

### Why is it needed?

Currently, countDraftRelations does not retrieve a specified locale. The locale is not provided in the parameters that's why the default locale is used to fetch the entity from the database. If the entity has not been created in the default locale, users will be unable to publish the single type in their specified locale.

### How to test it?

1. Choose a single type that hasn't been created for the default locale yet.
2. Switch to a locale other than the default, and fill in the content.
3. Try to publish it, and the content will be published successfully.

### Related issue(s)/PR(s)

This PR fixes this issue [#20188](https://github.com/strapi/strapi/issues/20188)

### DIFF
--- a/packages/core/content-manager/server/src/controllers/single-types.ts
+++ b/packages/core/content-manager/server/src/controllers/single-types.ts
@@ -192,7 +192,7 @@ export default {
   async countDraftRelations(ctx: any) {
     const { userAbility } = ctx.state;
     const { model } = ctx.params;
-
+    const { query = {} } = ctx.request;
     const entityManager = getService('entity-manager');
     const permissionChecker = getService('permission-checker').create({ userAbility, model });
 
@@ -200,7 +200,7 @@ export default {
       return ctx.forbidden();
     }
 
-    const entity = await findEntity({}, model);
+    const entity = await findEntity(query, model);
     if (!entity) {
       return ctx.notFound();
     }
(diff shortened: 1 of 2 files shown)


alexandrebodin:
> [!IMPORTANT] 
> With the release of Strapi 5, we have lifted the code freeze. To ensure we can effectively review and assist with contributions, we are now only accepting PRs targeting version 5 (the develop branch).<br/> 
> As your PR is targeting some of the most recent v4 version, We we would greatly appreciate it if you could update your contribution to align with version 5 and resubmit!

markkaylor:
Note: confirm the issue isn't still present in v5

packages/core/content-manager/server/src/controllers/single-types.ts
```diff
@@ -192,15 +192,15 @@ export default {
...
     const entityManager = getService('entity-manager');
     const permissionChecker = getService('permission-checker').create({ userAbility, model });
 
     if (permissionChecker.cannot.read()) {
       return ctx.forbidden();
     }
 
-    const entity = await findEntity({}, model);
+    const entity = await findEntity(query, model);
     if (!entity) {
```
jhoward1994, line 204:
```suggestion
    const sanitizedQuery = await permissionChecker.sanitizedQuery.read(query);

    const entity = await findEntity(sanitizedQuery, model);
```

I think we should sanitize the query as we do in the find

markkaylor (commented):
Hi @hiba-ijjig-nw I've tested it and it's working nicely. I agree with @jhoward1994, we need to sanitize the query. Could you please take care of that and then we can merge 🚀

hanpaine:
Closed due to inactivity

PR #20327, opened 2024-05-17
chore: make experimental-dev example app

joshuaellis:
### What does it do?

* adds `experimental-dev` application set up to work with react-rc & the compiler
* adds `USE_EXPERIMENTAL_DEPENDENCIES` flag to the strapi develop/build commands so we can test RCs / Betas of deps we require users to install in their strapi projects

### Why is it needed?

* so we can actively work and test against react19 for a smoother upgrade experience in the future

### Related issue(s)/PR(s)

* spawned from work #20314


examples/experimental-dev/.env.template
```diff
@@ -0,0 +1,4 @@
+# You'll need this to run the project, most likely.
+USE_EXPERIMENTAL_DEPENDENCIES=true
+# This is very experimental and isn't required to use with react@19
+USE_REACT_COMPILER=false
```
innerdvations, line 4:
```suggestion
USE_REACT_COMPILER=false

```

innerdvations (commented):
We need a bit more discussion on this.

The current "getstarted" should really be called "kitchensink" since it's got almost everything in it
The current "kitchensink" should be called something like "empty" or "clean" or almost anything else.

And then there should be a -ts version of those.

joshuaellis:
> We need a bit more discussion on this.
> 
> The current "getstarted" should really be called "kitchensink" since it's got almost everything in it The current "kitchensink" should be called something like "empty" or "clean" or almost anything else.
> 
> And then there should be a -ts version of those.

Should it be called `kitchensink`? we use these for development, they're not examples at all imho, they're full of "we need to test this" so in reality, i'd say `dev` is potentially more accurate??

Do we need TS versions of both, couldn't we just use the TS versions on their own? I don't think anyone realistically tests against the JS/TS versions? 🤔  at most we could do a `dev-js` version? but `experimental-dev` i'd just keep as ts to be honest, not to over complicate it?

innerdvations:
I'm also wondering, how temporary is this? Do we actually want to add something like this to main or would we be better off keeping a POC branch where we can QA against it until releasing it.

jhoward1994:
It all makes sense to me. I guess this kind of error is caused by the React 19 DS compatibility issues you mentioned?

```
react-router-dom.js?v=f90c7ab0:3547 React Router caught the following error during render TypeError: Cannot read properties of null (reading 'querySelectorAll')
    at getFocusableNodes (@strapi_design-system.js?v=f90c7ab0:40348:13)
    at @strapi_design-system.js?v=f90c7ab0:43307:28
    at commitHookEffectListMount (react-dom_client.js?v=f90c7ab0:15881:31)
    at commitHookLayoutEffects (react-dom_client.js?v=f90c7ab0:15963:15)
    at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16114:17)
    at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
    at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16112:15)
    at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
    at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16112:15)
    at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
```

joshuaellis:
> It all makes sense to me. I guess this kind of error is caused by the React 19 DS compatibility issues you mentioned?
> 
> ```
> react-router-dom.js?v=f90c7ab0:3547 React Router caught the following error during render TypeError: Cannot read properties of null (reading 'querySelectorAll')
>     at getFocusableNodes (@strapi_design-system.js?v=f90c7ab0:40348:13)
>     at @strapi_design-system.js?v=f90c7ab0:43307:28
>     at commitHookEffectListMount (react-dom_client.js?v=f90c7ab0:15881:31)
>     at commitHookLayoutEffects (react-dom_client.js?v=f90c7ab0:15963:15)
>     at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16114:17)
>     at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
>     at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16112:15)
>     at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
>     at commitLayoutEffectOnFiber (react-dom_client.js?v=f90c7ab0:16112:15)
>     at recursivelyTraverseLayoutEffects (react-dom_client.js?v=f90c7ab0:17173:15)
> ```

yep!

joshuaellis:
> I agree with Ben we should have JS and TS versions as we support both and thus should test both.

theres an RFC internally on this if you want to comment :) I think Ben shared it in engineering, the comments are out of date with what the PR does

innerdvations:
@derrickmehaffy (and for general transparency)

The short term goal is to reduce this down to just `dev` and `dev-js` which will have the same schemas and custom code. There's not a common need for the empty project for quick tests on the monorepo so we'll remove those. And then we'll have this `experimental-dev` as well for experimental features.

The long term goal would be to also add some scripts and configuration so that it's simple (or even automatic) to generate the desired test apps, and then maybe delete the examples altogether.

innerdvations (approved):
LGTM

PR #20335, opened 2024-05-20
fix: data transfer logging

innerdvations:
### What does it do?

- only create one log file per cli command run
- change log level to 'info' for file transport

### Why is it needed?

Data transfer commands are currently creating a log file for every individual warning that is sent to the console, and it's an empty file because the file transport was set to 'error'

This way we get only one single file for every data transfer command that is run

### How to test it?

unit test has been added

when you run a data transfer, you should only see a single log file, and it should only be created if there was a reason to create it like a warning or error.

### Related issue(s)/PR(s)

DX-1408


Bassel17 (approved):
LGTM

Convly (approved):

PR #20342, opened 2024-05-20
Fix transfer document id in DTS

Bassel17:
### What does it do?

- filtering out documentId key from data, so it's not removed when consolidating destination and source schemas

### Why is it needed?

- documentId property was not being transferred with the entities 

### How to test it?

- attempt a transfer ( import, export, or a transfer between two projects)
- documentId should not change, translations should be preserved and publishing status should be preserved as well

### DIFF
--- a/packages/core/data-transfer/src/engine/index.ts
+++ b/packages/core/data-transfer/src/engine/index.ts
@@ -841,7 +841,11 @@ class TransferEngine<
           const { type, data } = entity;
           const attributes = schemas[type].attributes;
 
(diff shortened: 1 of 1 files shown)


innerdvations (commented):
LGTM, but one minor annoying comment

It might be clearer logically if we add it to the other side, basically change `attributes` into `attributesToKeep` and make sure documentId is in there.

innerdvations (approved):

Convly (approved):

Issue #19750, opened 2024-03-13
Conflict Error with API ID (Plural) When Enabling i18n on New Single Type or Collection


Eventyret:
### Required System information

- Node.js version: 20.11.1
- NPM version: 10.2.4
- Strapi version: 4.20.4
- Database: sqlite
- Operating system: Mac
- Is your project Javascript or Typescript: javascript

### Describe the bug

When editing a single-type or collection in Strapi to enable i18n, an error is encountered. The API ID (plural) throws an error message indicating that the value is already in use by another content-type, even though this is not the case as it was newly created.

### Steps to reproduce the behavior
- Create a new single type or collection (without i18n)
- Save the single type or collection.
- Edit this single-type or collection by enabling i18n.
- Save the changes.
- An error is displayed in the Basic Settings regarding the API ID (plural) stating “this value is already in use by another content-type”.

### Expected behavior

The expectation is that once i18n is enabled for a single type or collection, it should save without any errors regarding the API ID (plural), as it should not conflict with other content types if it is unique and newly created.

### Screenshots

![image](https://github.com/strapi/strapi/assets/977232/a75d4423-c8ac-4140-80ac-c15f786a3210)

rGillyazanov:
If there is a hyphen in the API ID name, an error is displayed.

AdityaBanik:
Facing the same bug

zirho:
You can edit `schema.json` as a workaround.

Tsun4mii:
Facing the same bug. It appears not only on single types but on some collection types in my case

PR #20347, opened 2024-05-21
fix(ctb): validation compares pluralName and collectionName correctly

innerdvations:
### What does it do?

Fixes the comparison in the validator so that takenCollectionNames only includes collectionNames and not also collectionNames equal to the current pluralName (which meant that it would then compare its pluralName against its own collectionName and fail validation because they were equal)

### Why is it needed?

To allow saving content types after changing the value of i18n or d&p

### How to test it?

See https://github.com/strapi/strapi/issues/19750 for details. It should now allow saving.

e2e tests have also been added, which fail without the change and pass with the change

### Related issue(s)/PR(s)

DX-1160

Fixes https://github.com/strapi/strapi/issues/19750

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/FormModal/forms/forms.ts
+++ b/packages/core/content-type-builder/admin/src/components/FormModal/forms/forms.ts
@@ -251,10 +251,9 @@ export const forms = {
       const takenCollectionNames = isEditing
         ? collectionNames.filter((collectionName) => {
             const { schema } = contentTypes[ctUid];
-            const currentPluralName = schema.pluralName;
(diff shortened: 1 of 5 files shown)


Convly (approved):
Any chance we could add a test for this one (to avoid regression)? Or is it not setup yet?

innerdvations:
> Any chance we could add a test for this one (to avoid regression)? Or is it not setup yet?

We didn't have any ctb edit tests, but I've added some 🚀

Bassel17 (approved):
LGTM!
I like the new test utils very useful

PR #20348, opened 2024-05-21
enhancement: improve gitignore in generated projects

innerdvations:
### What does it do?

Adds package manager files, .tsbuildinfo, and .eslintcache to generated project gitignore and the root gitignore

I did not do the individual strapi package gitignore files because tbh they shouldn't even exist unless they are overriding the root level one, but we can deal with that later.

### Why is it needed?

To ignore more files, but specifically it is needed for .tsbuildinfo because that is now created when building typescript projects in v5

### How to test it?

Generate a new typescript project, build it, and see that it ignores those files.

### Related issue(s)/PR(s)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,6 +108,8 @@ packages/generators/app/files/public/
 schema.graphql
 dist
 .nx
+.tsbuildinfo
+.eslintcache
 
 ############################
 # Example app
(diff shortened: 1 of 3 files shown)


Convly (approved):
Overall LGTM

PR #20354, opened 2024-05-23
feat: releases settings

Marc-Roig:
### What does it do?

We are introducing a Settings page for releases, where Users can set their Default Timezone for release scheduling. That is so they don't have to manually change it every time they want to release.

image

This PR includes the backend endpoints for that feature.

- `GET /content-releases/settings` - Gets the stored settings , returns null by default.
- `PUT /content-releases/settings` - Updates the settings.


packages/core/content-releases/server/src/utils/index.ts
```diff
@@ -1,9 +1,19 @@
...
-export const getService = (
-  name: 'release' | 'release-validation' | 'scheduling' | 'release-action' | 'event-manager',
+type Services = {
+  release: any;
+  'release-validation': any;
+  scheduling: any;
+  'release-action': any;
+  'event-manager': any;
+  settings: SettingsService;
+};
```
Marc-Roig, line 11: @Feranchz I updated the `getService` method to be typed, but we might as well want to add the rest of services type. Do you prefer to fill them in another PR?

packages/core/content-releases/server/src/routes/settings.ts
```diff
@@ -0,0 +1,38 @@
...
+      handler: 'settings.find',
+      config: {
+        policies: [
+          'admin::isAuthenticatedAdmin',
+          // {
+          //   name: 'admin::hasPermissions',
+          //   config: {
+          // actions: ['plugin::content-releases.settings-read'],
+          //   },
+          // },
```
remidej, line 16: looks like this was forgotten (same below). I would assume we do want that custom action?

packages/core/content-releases/shared/contracts/settings.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Update the stored settings
+ */
+export declare namespace UpdateSettings {
+  export interface Request {
+    body: Settings;
+  }
+
+  export interface Response {
+    data: Settings;
+    error?: errors.ApplicationError | errors.ValidationError;
```
remidej, line 40: isn't it possible to tie the two together? To use a union to let TS know that we either have data and no error, or an error and no data.

remidej (commented):
Looks good! I

packages/core/content-releases/server/src/routes/settings.ts
Marc-Roig: you got it! forgot to remove the comment in this pr but indeed it should be covered by the other PR!

packages/core/content-releases/shared/contracts/settings.ts
Marc-Roig: sure :) good point, will create a util for this!

packages/core/content-releases/shared/types.ts
```diff
@@ -24,3 +24,26 @@ export interface Entity {
...
+ * type Response = OneOf<
+ *   { data: Data },
+ *   { error: ApplicationError | ValidationError }
+ * >;
+ *
+ * // Is equivalent to:
+ * type Response = { data: Data, error: never } | { data: never, error: ApplicationError | ValidationError };
+ *
+ */
+export type OneOf<T, U> = (T & { [K in keyof U]?: never }) | (U & { [K in keyof T]?: never });
```
Marc-Roig, line 49: @Convly Do you think I could fit a utility like this into our types package? This is something I could start using a lot in our type contract files

Convly: It could actually replace the usage we're having of [XOR](https://github.com/strapi/strapi/blob/v5/main/packages/core/types/src/utils/index.ts#L99-L126) in a lot of places.

Go for it :slightly_smiling_face:

packages/core/content-releases/shared/contracts/settings.ts
Marc-Roig: created an utility to make it easier on the other scenarios? wdyt? 
We can use this everywhere else from now on !

Feranchz (approved):

PR #20377, opened 2024-05-27
fix(webhooks): ensure id sent to client is always a string

markkaylor:
### What does it do?

Frontend
- Fix create action not redirecting to the created entry
- Fix delete action in the list view not closing the confirm dialog
- Fix update action not opening the EditPage

Backend
- Refactor to use the Webhook type exported from @strapi/types
- Ensure the webhook store returns webhooks with string ids as the type dictates

I set it up this way since the Webhook type id is set as string, so I assume that's what we want to be dealing with even if db.query returns a number... Open to discussion not sure it's the best solution 🤷 

### Why is it needed?

Otherwise the client expects a string but receives a number this breaks client side redirects when navigate receives a number instead of a string

### How to test it?

- Go to the webhooks page, settings -> webhooks
- Create a webhook, on create the url should change from /create to /:id
- Go to the list view and try to click the row or the pencil icon to update, it should redirect to the webhook edit page
- Go the list view and try to delete one or many webhooks, the modal should close on delete

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/EditPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/EditPage.tsx
@@ -116,7 +116,7 @@ const EditPage = () => {
           message: formatMessage({ id: 'Settings.webhooks.created' }),
         });
 
-        navigate(res.data.id, { replace: true });
+        navigate(`/settings/webhooks/${res.data.id}`, { replace: true });
       } else {
         const res = await updateWebhook({ id: id!, ...cleanData(data) });
 
--- a/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/EditPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/EditPage.tsx
@@ -116,7 +116,7 @@ const EditPage = () => {
           message: formatMessage({ id: 'Settings.webhooks.created' }),
         });
 
-        navigate(`/settings/webhooks/${res.data.id}`, { replace: true });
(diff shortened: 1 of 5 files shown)


packages/core/admin/admin/src/pages/Settings/pages/Webhooks/EditPage.tsx
```diff
@@ -116,7 +116,7 @@ const EditPage = () => {
           message: formatMessage({ id: 'Settings.webhooks.created' }),
         });
 
-        navigate(res.data.id, { replace: true });
+        navigate(`/settings/webhooks/${res.data.id}`, { replace: true });
```
remidej, line 119: tiny nitpick but I think a relative link (e.g. `../${res.data.id}`) gives us more flexibility if the parent routes ever change

packages/core/core/src/services/webhook-store.ts
```diff
@@ -33,31 +34,14 @@ const webhookModel: Model = {
...
-}
+type Webhook = Modules.WebhookStore.Webhook;
 
-interface DBOutput {
-  id: string;
-  name: string;
-  url: string;
-  headers: Record<string, string>;
-  events: string[];
+interface DBOutput extends Omit<Webhook, 'id' | 'isEnabled'> {
```
Convly, line 39: That's just a detail but I'm wondering why we're mixing interface (for output) and types (for input).

Is this to allow certain behavior? Otherwise, I would suggest keeping only one syntax (probably type if we don't need the interface features).

remidej (approved):
Tested ✅

PR #16348, opened 2023-04-12
chore: release peer version constraint using ^

chentsulin:
### What does it do?

Using `^` range to release the version constraints of the peerDependencies.

### Why is it needed?

The original peer constraints are way too strict. We have `styled-components@5.3.6` in our project and the version can't be accepted because it's not `5.3.3`.

PR #20380, opened 2024-05-28
chore: release peer version constraint using ^

chentsulin:
### What does it do?

Using `^` range to release the version constraints of the `peerDependencies`.

### Why is it needed?

The original peer constraints are way too strict. We have `styled-components@5.3.6` in our project and the version can't be accepted because it's not `5.3.3`.

### Related issue(s)/PR(s)

There is a similar pr I submitted last year: https://github.com/strapi/strapi/pull/16348


alexandrebodin:
Hi @chentsulin We release v5 that changes how this is done. I'm going to close as we do not accept v4 PRs anymore for non critical or security issues.

If you still see some issues on v5 let us know 🙌

PR #20385, opened 2024-05-28
Fix transfer DZ and repeatable components

### LINKED ISSUE #20394: DTS Import Unique Constraint Failed

- Environment: development
- OS: darwin-arm64
- Strapi Version: 5.0.0-beta.8
- Node/Yarn Version: npm/10.2.4 node/v20.11.1 darwin arm64 workspaces/false
- Edition: Community
- Database: sqlite

### Describe the bug

Get  `UNIQUE constraint failed` error when doing a DTS import on SQlite Database.

1. Run `strapi import`

https://github.com/strapi/strapi/assets/6153188/ee200c62-fccd-4be1-bdd5-010241f2486f

### Expected behavior

Should not fail with `Unique constrain` error.

Bassel17:
### What does it do?

filters out links related to components so we don't export them

### Why is it needed?

- components and dz are already populated at the entity stage 
- we face a unique constraint error otherwise by attempting to insert duplicate ids

### How to test it?

- export a strapi app with repeatable components and dz data it should work if you try to import it back

fix #20394


packages/core/data-transfer/src/strapi/queries/link.ts
```diff
@@ -338,7 +338,16 @@ const filterValidRelationalAttributes = (attributes: Record<string, any>) => {
...
   const isComponentLike = (attribute: any) => {
-    return attribute.component || attribute.components;
+    return (
+      /**
+       * probably can remove attribute.component & attribute.components
+       * to be checked
+       *  */
+      attribute.component ||
+      attribute.components ||
+      attribute.joinTable?.name.includes('components') ||
```
Convly, line 348: How is that possible to have a relation to components/cmps that is not a component/dynamic zone?

What is the shape of the attribute you get here?

Bassel17: it's not different, I just think the object we are receiving doesn't contain this anymore attribute.component  or
      attribute.components

Convly: I see, could you make it so that we're checking the suffix instead of a loose `includes`? This should prevent having false positives.

Also, you can indeed remove the previous check.

innerdvations: Just curious what causes this bug on v5 but not on v4. Are we just covering up another issue?

innerdvations: Basically: can you see if this issue exists or not in v4?

If the issue also exists in v4, this fix seems ok (but will also need to be merged in v4 as well).

If this issue doesn't exist in v4, it means that something unexpected has changed in the objects sent, and it could be an underlying problem we need to look into.

Bassel17: it's only a v5 issue, the v4 object contains the component/components attribute, it seems it's been removed in v5
at the same time it doesn't look like it's been used anywhere else expect here previously

innerdvations:
Could you add a unit test for `filterValidRelationalAttributes` that takes attributes and filters them correctly?

Convly (commented):
Given the following structure

`CT1 > (comp) CompA > (rel) CT2`

Would `CompA > CT2` link be preserved with this solution?

Bassel17:
> Given the following structure
> 
> `CT1 > (comp) CompA > (rel) CT2`
> 
> Would `CompA > CT2` link be preserved with this solution?

it works, the links are preserved, this only changes what we are checking against "table name" but it still the same behavior as before, we are ignoring the same links we used to ignore

Convly (approved):

Convly (changes requested):
Looks like some tests (including lint) are failing

Bassel17:
> Looks like some tests (including lint) are failing

indeed seems i forgot to commit one of the files

Convly (approved):
LGTM

innerdvations (approved):
Looks great now, thanks!

PR #20386, opened 2024-05-28
Fix: bulk publish modal and related issues

madhurisandbhor:
### What does it do?

Fixes following bugs: 
1. Publish confirm modal message for internalization is misaligned, aligned at center.
2. On refresh published/ ready to publish entries are not updated.
3. On locale picker change reset entry selection, https://strapi-inc.atlassian.net/browse/CS-844 : I have a temporary fix but I am looking for a better way to reset selected entries on locale picker change across all table entries.


packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -333,15 +333,16 @@ const SelectedEntriesModalContent = ({
...
+  // Filter selected entries from the updated modal table rows
+  const selectedEntries = rows.filter((entry) =>
+    selectedRows.some((selectedEntry) => selectedEntry.documentId === entry.documentId)
+  );
 
-    return acc;
-  }, []);
+  const entriesToPublish = selectedEntries
+    .filter((entry) => !validationErrors[entry.documentId])
+    .map((entry) => entry.documentId);
```
remidej, line 345: Do you know what the max number of entries this code will have to manage at once will be? Something like 10 or 20?

I'm asking because your change increases readability but is a bit worse for performance, so it's only worth it if it doesn't deal with lots of data

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -330,6 +329,30 @@ const ActionsCell = styled(Table.Cell)`
...
+ * TableActionsBar
+ * -----------------------------------------------------------------------------------------------*/
+
+const TableActionsBar = () => {
+  const selectRow = useTable('TableActionsBar', (state) => state.selectRow);
+  const [{ query }] = useQueryParams<{ plugins: { i18n: { locale: string } } }>();
+  const locale = query?.plugins?.i18n?.locale;
+  const prevLocale = usePrev(locale);
+
+  // TODO: find a better way to reset the selected rows when the locale changes across all the app
```
remidej, line 342: Have you looked into handling the event when the user clicks on the locale dropdown?

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
madhurisandbhor: its not for readability, previous change was done by me only but its throwing TS error for accumulators `documentId` type `never` not matching with selectedEntries.documentId type `string`. so I updated it with simpler code. I could use your help to resolve the issue.
For the performance as its just selectedEntries its going to be max the pagination size so its not really impacting performance alot IMO.

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
madhurisandbhor: Yes its in i18n plugin, so making changes from this plugin inside Table is not doable.
I thought of making this change inside `Table.tsx` but I am not sure if it will pollute the Table functionality.

remidej: I see, in that case in think your solution is fine 👍

remidej (approved):
Tested ✅

PR #20387, opened 2024-05-28
fix: add clone actions to history

remidej:
### What does it do?

Adds clone and auto-clone actions to content history

### Why is it needed?

When you create an entry by cloning an existing one, before, the created entry would have 0 history versions. Now we keep track of who created it, and users no longer land on a empty history page.

### How to test it?

Create an entry via the auto-clone feature. You can do it on the "temp" content type in getstarted since it's simple enough. You can find the duplicate button in the list view in the popover. Go to the history page for the created entry, you should see one history item.

Create an entry via the clone feature on a content type like kitchensink. Click the same duplicate button, then remember to save to actually create the new entry. Same thing, you should see one history item for the new entry.

No new entries should be created for the original entries you used to clone content.

### DIFF
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -38,10 +38,11 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
           return next();
         }
 
-        // NOTE: can do type narrowing with array includes
+        // NOTE: cannot do type narrowing with array includes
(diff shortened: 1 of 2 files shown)


innerdvations (approved):
LGTM!

markkaylor (approved):
I tested locally for temp and it works ✨ , I'm currently not able to create a kitchensink entry to duplicate since I get a 500 error. I confirmed this is on v5/main and as best I can tell it is not related to your work here.

PR #18519, opened 2023-10-20
Adds roles to user allowed fields

rolangom:
### What does it do?

This added field 'roles' allows user's "Media Library" permissions to be validated, when "Same role as creator" is checked.

### Why is it needed?

When an admin user is acceding the Media Library section, with a role action that was previously marked with "Same role as creator", Strapi is throwing a { status: 400, name: "ValidationError", message: "Invalid parameter roles", … } error.

### How to test it?

Node.js version: v20.8.0
NPM version: 10.1.0
Strapi version: v4.14.4
Database: sqlite
Operating system: darwin-x64
Is your project Javascript or Typescript: Javascript

Steps to reproduce:
0. Using the getstarted example.
1. The Super-Admin has created some foldres and uploaded some files in the root folder.
2. Then create two roles: 'XReader' and 'XWritter' which can access (readonly read) and create(upload, update) in media library respectively with "same role as creator" checked.
3. Create 2 users, user 1 whose is roles are 'XWritter' and 'XReader', and user 2 whose is just 'XReader'.
4. Try to access to the Media Library with any of the user 1 or user 2
5. Strapi should display the error

### Related issue(s)/PR(s)

Issue #18297

PR #20389, opened 2024-05-29
Fix: Query Validation in Files' Admin API

### LINKED ISSUE #18297: Problem with privileges for accessing "Media Library" files

- Node.js version: 18.17.1
- NPM version: 9.6.7
- Strapi version: 4.10.1
- Database: postgres
- Operating system: OS: linux-x64

### Describe the bug

I'm having issues accessing files within the "Media Library".

Access privileges to the media library via the "Has the same role as the author" setting do not work

Screenshot 2023-10-04 alle 16 53 48

Convly:
### What does it do?

Fixes the query validation for the files' admin API.

It now only validates/sanitizes the request query, not the one generated with permissions/conditions. 

### Why is it needed?

See #18297

### How to test it?

`yarn test:api tests/api/core/upload/admin/file.test.api.js`

or see reproduction steps on #18519

### Related issue(s)/PR(s)

fix #18297


packages/core/upload/server/controllers/admin-file.js
```diff
@@ -24,10 +24,16 @@ module.exports = {
       return ctx.forbidden();
     }
 
-    const pmQuery = pm.addPermissionsQueryTo(merge(defaultQuery, ctx.query));
+    await pm.validateQuery(ctx.query);
```
innerdvations, line 27:
```suggestion
    // validate the incoming user query params
    await pm.validateQuery(ctx.query);
```

packages/core/upload/server/controllers/admin-file.js
```diff
@@ -24,10 +24,16 @@ module.exports = {
...
 
-    await pm.validateQuery(pmQuery);
-    const query = await pm.sanitizeQuery(pmQuery);
+    const query = await pipeAsync(
+      // Start by sanitizing the incoming query
+      (q) => pm.sanitizeQuery(q),
+      // Add the default query
```
innerdvations, line 32:
```suggestion
      // Add the default query which should not be validated or sanitized
```

tests/api/core/upload/admin/file.test.api.js
```diff
@@ -53,6 +56,81 @@ describe('Upload', () => {
...
+        },
+        {
+          action: 'plugin::upload.assets.update',
+          subject: null,
+          conditions: ['admin::has-same-role-as-creator'],
+          properties: {},
+        },
+      ]);
+
+      // TODO: We create 2 users in order to be able to test the condition itself (same role as creator)
```
innerdvations, line 95:
```suggestion
      // Create 2 users in order to be able to test the condition itself (same role as creator)
```

innerdvations:
I think the failing CI could be because our mysql:latest container has upgraded to 8.4 from 8.3.x

tests/api/core/upload/admin/file.test.api.js
Convly: We're not testing the condition for now, hence the todo, why do you want to remove it?

Convly:
> I think the failing CI could be because our mysql:latest container has upgraded to 8.4 from 8.3.x

It is, I've tried to investigate that in another PR, but no solution yet

innerdvations:
Initial guess is that it's an issue with the `bitnami/mysql:latest` container that they'll fix, but it looks like if we switch to `mysql:latest` [it works](https://github.com/strapi/strapi/pull/20400). But I don't have context for why we're using bitnami.

tests/api/core/upload/admin/file.test.api.js
Convly: The todo is to say that we'll need additional tests to check the condition itself (not just that you can send the query, but also that the condition is respected)

innerdvations (approved):
LGTM!

jhoward1994 (approved):

PR #20393, opened 2024-05-29
Fix: disable bulk selection conditionally

madhurisandbhor:
### What does it do?
Fixes the bug where if user doesn't have delete or publish permission, `select`/ `select all` is still enabled which is misleading as nothing happens on selecting muliple entries. So if user doesnt have delete/publish permissions disabling  entries selection.

### DIFF
--- a/packages/core/admin/admin/src/components/Table.tsx
+++ b/packages/core/admin/admin/src/components/Table.tsx
@@ -254,7 +254,7 @@ const ActionBar = ({ children }: Table.ActionBarProps) => {
  * HeaderCheckboxCell
  * -----------------------------------------------------------------------------------------------*/
 
-const HeaderCheckboxCell = () => {
+const HeaderCheckboxCell = ({ disabled = false }: { disabled?: boolean }) => {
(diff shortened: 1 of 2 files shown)


simotae14 (approved):
👍🏽

Feranchz:
I'm not completely sure if this is the right solution.

We don't want to show the checkboxes if you don't have publish/delete permissions because we are not showing those buttons, but plugins can add Bulk Buttons if I'm not wrong, so maybe we could have (or already have) plugins that add one bulk button that is not related to delete/publish permissions

madhurisandbhor:
> ut plugins could add Bulk Buttons if I'm not wrong, so maybe we could have (or already have) plugins that add one bulk button that is not related to delete/publish permissions

Yes you are right, I missed it. It could be add to release and relates to release permission so we cant disable it then 😲

joshuaellis:
Fernando makes a really great point, I think we just have to leave it as is?

madhurisandbhor:
Yes I think so, so closing it, thanks for the inputs.

PR #20395, opened 2024-05-30
Fix: bulk publish validation error

madhurisandbhor:
### What does it do?

Fixes validation error for simple error message (without nested error fields)

### How to test it?

When user is on bulk publish modal, if there are any validation errors on the content it should be displayed correctly.


Feranchz (approved):
🚀

PR #20400, opened 2024-05-30
chore(ci): remove mysql_native_password in test container

innerdvations:
### What does it do?

Removes `MYSQL_AUTHENTICATION_PLUGIN=mysql_native_password`

### Why is it needed?

8.4 (which mysql:latest now points to) throws an error instead of deprecation warning when using mysql_native_password

### Related issue(s)/PR(s)

DX-1439

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -289,7 +289,7 @@ jobs:
         shard: [1/5, 2/5, 3/5, 4/5, 5/5]
     services:
       mysql:
-        image: bitnami/mysql:latest
+        image: mysql:latest
         env:
           MYSQL_ROOT_PASSWORD: strapi
           MYSQL_USER: strapi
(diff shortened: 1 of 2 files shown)


Convly (commented):
Silly question but should we remove it entirely or enable it with the new option?

`--mysql-native-password=ON`

innerdvations:
I'd opt for just using the defaults, it's CI and if it fails because a default changed I'd rather know and fix it, but no strong opinion either way.

Convly (commented):
Should we lock mysql to a certain version or do we want to keep fixing upcoming issues on as they arrive to be sure we're always compatible with latest?

alexandrebodin (approved):

innerdvations:
Example projects: I would use a range to let it stay updated, easier to notice when something goes wrong
Generated projects: I lean towards pinning it to a specific version, but if we trust mysql2 enough we could use `~3.10.0` for those

PR #20401, opened 2024-05-30
chore: test e2e history version can be restored

markkaylor:
### What does it do?

- Add e2e test to confirm history version can be restored

### Why is it needed?

- To test the feature works as expected

### How to test it?

- All history tests should pass here in the CI


markkaylor:
Forgot to add the test for Single types...I'll do that now

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -281,6 +280,35 @@ describeOnCondition(edition === 'EE')('History', () => {
...
+      await page.waitForURL(ARTICLE_CREATE_URL);
+
+      const titleInput = await page.getByRole('textbox', { name: 'title' });
+      // Create an initial entry to also create an initial version
+      await titleInput.fill('Being from Kansas');
+      await page.getByRole('button', { name: 'Save' }).click();
+      await page.waitForURL(ARTICLE_EDIT_URL);
+      // Update to create another version
+      await titleInput.fill('Being from Florida');
+      await page.getByRole('button', { name: 'Publish' }).click();
```
remidej, line 296: I believe you mean to click the save button here, not publish

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -281,6 +280,35 @@ describeOnCondition(edition === 'EE')('History', () => {
...
+
+      await goToHistoryPage(page);
```
remidej, line 298: Based on my experience with the flaky tests, I recommend awaiting tons of stuff to be extra safe. Here I think you can use the function to check that the success notification has appeared before opening history.

You may get an error because there could be two success notifications at once, after the create action. To prevent it you could replace the `await page.waitForURL(ARTICLE_EDIT_URL)` above by `closeSuccessNotification` so that it gets dismissed

remidej (commented):
I wonder if there's any point in duplicating this for single types?

tests/e2e/tests/content-manager/history.spec.ts
markkaylor: Yeah... I did this intentionally 🙃 should have mentioned it. The save button for some reason is not being enabled when I fill the text input a second time. I saw this [issue on playwright](https://github.com/microsoft/playwright/issues/27564) where they indicate it is related to [hydration](https://playwright.dev/python/docs/navigations#hydration). I'm not sure if that's the cause for us but also not sure it's an issue with playwright. Wondering too if this is maybe caused by a similar issue:

https://github.com/strapi/strapi/blob/955bc0269eee58bb62229db838870577196ded81/tests/e2e/tests/content-manager/history.spec.ts#L31-L37

Happy to investigate just not sure if it should be done here on this PR...

markkaylor:
> I wonder if there's any point in duplicating this for single types?

Yeah I think you might be right, a single test for collection-types should be enough.

PR #20407, opened 2024-05-31
feat: customize localization label actions in history

remidej:
### What does it do?

Changes the i18n label action wording on the history page. It does it by mutating the labelAction component at the field level in history when it detects that it comes from i18n.

There's a completely different way to solve the problem that I also explored. It requires adding another CM hook to let the i18n plugin. But it prevents us from directly using the edit view layout in history, so it ends up requiring a lot more changes. Even though this current solution feels more hacky, I personally prefer it.

### Why is it needed?

Instead of:

>This value is the same across all locales

When in history we want to show:

>This value is common to all locales. If you restore this version, the content will be replaced for all locales.

So that users are aware that their actions on one locale can affect the others.

### How to test it?

The i18n tooltip is actually broken 😅 but that's unrelated to this PR, it already happens in the edit view on v5/main. So to test this current change, inspect the icons in the devtools. You can use the `category` content type on kitchen sink which now has both localized and non localized fields.

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
@@ -219,6 +219,39 @@ type VersionInputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
 };
 
 /**
+ * Checks if the i18n plugin added a label action to the field and modifies it
+ * to adapt the wording for the history page.
+ */
+const getLabelAction = (labelAction: VersionInputRendererProps['labelAction']) => {
+  if (!labelAction || typeof labelAction !== 'object' || !('props' in labelAction)) {
+    return labelAction;
+  }
+
+  const customLabelAction =
+    typeof labelAction === 'object' && labelAction != null && 'props' in labelAction
+      ? { ...labelAction, props: { ...labelAction.props } }
+      : labelAction;
+  const labelActionTitleId =
+    labelAction && typeof labelAction === 'object' && labelAction.props.title.id;
+
+  if (labelActionTitleId === 'i18n.Field.localized') {
+    customLabelAction.props.title = {
+      id: 'history.content.localized',
+      defaultMessage:
+        'This value is specific to this locale. If you restore this version, the content will not replaced for other locales.',
+    };
+  } else if (labelActionTitleId === 'i18n.Field.not-localized') {
+    customLabelAction.props.title = {
+      id: 'history.content.not-localized',
+      defaultMessage:
+        'This value is common to all locales. If you restore this version and save the changes, the content will be replaced for all locales.',
+    };
+  }
+
+  return customLabelAction;
+};
+
+/**
  * @internal
  *
  * @description An abstraction around the regular form input renderer designed specifically
@@ -229,8 +262,11 @@ const VersionInputRenderer = ({
   visible,
   hint: providedHint,
   shouldIgnoreRBAC = false,
+  labelAction,
   ...props
 }: VersionInputRendererProps) => {
+  const customLabelAction = getLabelAction(labelAction);
+
   const { formatMessage } = useIntl();
   const version = useHistoryContext('VersionContent', (state) => state.selectedVersion);
   const configuration = useHistoryContext('VersionContent', (state) => state.configuration);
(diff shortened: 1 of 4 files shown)


packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -218,6 +218,39 @@ type VersionInputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
...
+  } else if (labelActionTitleId === 'i18n.Field.not-localized') {
+    customLabelAction.props.title = {
+      id: 'history.content.not-localized',
+      defaultMessage:
+        'This value is common to all locales. If you restore this version and save the changes, the content will be replaced for all locales.',
+    };
+  }
+
+  return customLabelAction;
+};
```
joshuaellis, line 252: Is `labelAction` a `ReactNode`? IIRC it is, it would be better to do a `cloneElement` rather than accessing the react internals directly... you can also use `isValidElement` as well, since it's not a component reference but at this point we've rendered it too that should work to tell you you can do `cloneElement`. Otherwise i think the solution isn't as hacky as you might think! 😄

remidej: I'm happy to make the hack less hacky! let me look into this. thanks

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -218,6 +218,39 @@ type VersionInputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
...
+  if (!labelAction || typeof labelAction !== 'object' || !('props' in labelAction)) {
+    return labelAction;
+  }
+
+  const customLabelAction =
+    typeof labelAction === 'object' && labelAction != null && 'props' in labelAction
+      ? { ...labelAction, props: { ...labelAction.props } }
+      : labelAction;
+  const labelActionTitleId =
+    labelAction && typeof labelAction === 'object' && labelAction.props.title.id;
```
remidej, line 235: I realize those checks should be removed with the early return, will update

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -218,6 +218,43 @@ type VersionInputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
...
+/**
+ * Checks if the i18n plugin added a label action to the field and modifies it
+ * to adapt the wording for the history page.
+ */
+const getLabelAction = (labelAction: VersionInputRendererProps['labelAction']) => {
+  if (!React.isValidElement(labelAction)) {
+    return labelAction;
+  }
+
+  const labelActionTitleId = labelAction.props.title.id;
```
joshuaellis, line 230:
```suggestion
	// TODO: find a better way to do this rather than access internals
  const labelActionTitleId = labelAction.props.title.id;
```

joshuaellis: just feel like it's one step away from being perfect, but this would take ages to figure out.

joshuaellis (approved):
Nice job ✅

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -218,6 +218,44 @@ type VersionInputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
...
+	// TODO: find a better way to do this rather than access internals
+  const labelActionTitleId = labelAction.props.title.id;
+
+  if (labelActionTitleId === 'i18n.Field.localized') {
+    return React.cloneElement(labelAction, {
+      ...labelAction.props,
+      title: {
+        id: 'history.content.localized',
+        defaultMessage:
+          'This value is specific to this locale. If you restore this version, the content will not replaced for other locales.',
```
markkaylor, line 239: Looks like you might have a typo here:

```suggestion
          'This value is specific to this locale. If you restore this version, the content will not be replaced for other locales.',
```

markkaylor (approved):
Tested looks good to me once the typo is adjusted

PR #20413, opened 2024-05-31
Replace lock icon with lightning icon in the Settings and Navbar links

simotae14:
### What does it do?

Replace the Lock icon used for the features not available without EE license with the Lightning icon

### Why is it needed?

We want to replace the Lock icon that shows users they don’t have access to Releases and other features in the Community Edition by a lightning flash icon so that these features are perceived as boosters if users upgrade their app (rather than blockers).

### How to test it?

Open a Strapi App with Community Edition

### Related issue(s)/PR(s)

CS-808


joshuaellis (changes requested):
changing `lockIcon` to `locked` constitutes a breaking change, whilst I doubt third parties are using it, I don't think it's right to make the change. Lets keep that property the same name with a `TODO` to rename it in V5 👍🏼

joshuaellis (approved):

PR #20415, opened 2024-05-31
fix: prevent duplicate history versions on publish

remidej:
### What does it do?

Prevents the creation of a duplicate history item when publishing a document.

Because of how D&P works in v5, when publishing a document, both the draft and the published document get updated, which until now created two items in content history. This was confusing to users. Now they should only see the published entry in history when they publish a document.

### How to test it?

Publish a document, check history.


markkaylor:
Not positive but I think this will have an impact on the e2e tests 🤔

markkaylor (approved):
Tested. Looks good, definitely think this is more intuitive 👍

Issue #17317, opened 2023-07-14
Passport SSO providers do not use the globalProxy when fetching user information


derrickmehaffy:
### Required System information

- Node.js version: 18
- NPM version: 8
- Strapi version: 4.11.5
- Database: SQLite
- Operating system: Linux Mint 21.1
- Is your project Javascript or Typescript: Javascript

### Describe the bug

All of the passport.js providers we recommend in our documentation are not getting the proxy agent constructed from the `globalProxy` config option as we don't pass in the `strapi.fetch.agent` but even if we did/could it appears that a nested package `passport-oauth2` doesn't even have a mechanism for passing in an http proxy agent anyway.

### Steps to reproduce the behavior

Setup for testing is quite complex as you will need to install something like squid and configure it in your Strapi application. However if you want to take a look at a sample repo I constructed as a workaround for a customer and work backwards from there you can see how I got around it.

https://github.com/strapi-support-demo-apps/strapi-example-v4-sso-proxy

### Expected behavior

Honestly I don't know what we can do if the downstream libs don't support passing in a proxy other than going with a forced global proxy to force all traffic over the proxy however this will break implementations where a user only wants some traffic to go over the proxy and not custom http calls that say need access to resources on the local machine or in the same internal network.

### Screenshots

see example repo

### Code snippets

see example repo

### Additional context

Related to internal TID 4233

derrickmehaffy:
Patch given to the EE customer was this one to override the nested provider to support proxies: https://github.com/strapi-support-demo-apps/strapi-example-v4-sso-proxy

PR #20416, opened 2024-05-31
enhancement!: use server.globalProxy value for creating a global http/https proxy

innerdvations:
### What does it do?

- Move `server.proxy` to `server.proxy.koa`
- Move `server.globalProxy` to `server.proxy.global`
- add `server.proxy.fetch` to act only on Fetch requests  (defaults to proxy.global if set)
- add `server.proxy.http` to act on `http` requests (defaults to proxy.global if set)
- add `server.proxy.https` to act on `https` requests (defaults to proxy.global if set)

This

### Why is it needed?

Allows proxy of all http and https requests globally (including anything that uses http/https internally, such as axios, passport-oauth2, etc, giving Strapi an actual global proxy to solve issues like [proxied SSO](https://github.com/strapi/strapi/issues/17317)

### How to test it?

- set server.proxy.global (or ad individual value) in your config
- use something like `npx http-echo-server` to create an echo proxy
- set up SSO (or whatever feature you want to test proxied)
- when you use SSO, you should see it proxied through your echo server
- fetch requests (such as telemetry) should still come through to the echo proxy

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/17317

DX-989

Documentation PR: https://github.com/strapi/documentation/pull/2124

TODO:

- [ ] write a codemod to migrate server.proxy -> server.proxy.koa and server.globalProxy -> server.proxy.global which will come in a later PR


derrickmehaffy:
Building exp release: https://github.com/strapi/strapi/actions/runs/9322440851

derrickmehaffy:
@innerdvations looks like we are getting build failures with `@strapi/core`:

image

derrickmehaffy:
Can you run a new exp version build Ben and I'll at in the morning

derrickmehaffy:
new exp: https://github.com/strapi/strapi/actions/runs/9354606706

derrickmehaffy:
exp version: `0.0.0-experimental.c3e9d4b26f9fd3d9eb530b5c11f9baa1d09b13ad`

derrickmehaffy:
Something I noticed so far is that image thumbails loaded by the media lib are not passing through the proxy:

image

Will continue to test for patching of SSO

innerdvations:
> Something I noticed so far is that image thumbails loaded by the media lib are not passing through the proxy:
> 
> image


I haven't looked into how they're loade. If they're essentially an img src so it would make sense not to pass through the proxy. If we have something where the server is pulling them from the remote bucket and then sending them to the user, that should go through the proxy, but I'll consider it a separate issue and address it later.

derrickmehaffy:
> > Something I noticed so far is that image thumbails loaded by the media lib are not passing through the proxy:
> > 
> > image
> 
> 
> I haven't looked into how they're loaded. If they're essentially an img src so it would make sense not to pass through the proxy. If we have something where the server is pulling them from the remote bucket and then sending them to the user, that should go through the proxy, but I'll consider it a separate issue and address it later.

Ah yeah your probably right ignore me

derrickmehaffy (approved):
Looking good! these are the logs from my squid proxy:

image

packages/core/core/src/services/server/index.ts
```diff
@@ -17,7 +17,7 @@ const healthCheck: Core.MiddlewareHandler = async (ctx) => {
 
 const createServer = (strapi: Core.Strapi): Modules.Server.Server => {
   const app = createKoaApp({
-    proxy: strapi.config.get('server.proxy'),
```
Convly, line 20: Shouldn't this change have a dedicated codemod?

If it's meant for another PR, could we add a TODO so that we don't forget?

Convly (commented):
The code LGTM, one small comment.

I couldn't test the actual behavior as I don't have the correct setup.

packages/core/core/src/services/server/index.ts
innerdvations: yep, already created a jira ticket but I'll mention it here too

innerdvations:
> I couldn't test the actual behavior as I don't have the correct setup.

You can test it generally using `http-proxy-server`, for example that the `fetch` option still works. Also, if you write some custom code and send a request using the `http` or `https` modules, those should also get proxied.

For the more advanced stuff like testing that SSO works through a proxy, I'm hoping @derrickmehaffy will be able to give it one final QA once the code has settled before we merge.

packages/core/core/src/utils/fetch.ts
```diff
@@ -14,11 +14,13 @@ export const createStrapiFetch = (strapi: Core.Strapi): Modules.Fetch.Fetch => {
     return fetch(url, fetchOptions);
   }
 
-  const globalProxy =
-    strapi.config.get<ConstructorParameters<typeof ProxyAgent>[0]>('server.globalProxy');
+  const proxy =
+    strapi.config.get<ConstructorParameters<typeof ProxyAgent>[0]>('server.proxy.fetch') ||
```
Bassel17, line 18: I was wondering if it will be clearer to add a type alias for this `ConstructorParameters<typeof ProxyAgent>[0]`
will leave it for you to decided otherwise looks good

Bassel17 (approved):
LGTM code wise, just left a small suggestion

packages/core/core/src/services/server/index.ts
alexandrebodin: Hum if we just have strapi.proxy set can't we consider that it's should be the proxy applied for everything like a shorthand version ? that could be super convenient 🤔

innerdvations: I don't have time to write tests for it before I'm off, but I did this as a follow-up convenience feature, and someone can hopefully finish it and update the docs: https://github.com/strapi/strapi/pull/20537

innerdvations: Oh wait, I think I misunderstood. You mean the strapi.proxy could be like `string | { fetch, http, https }` and if it's a string it is treated as global?

Issue #20225, opened 2024-04-28
core::5.0.0-discard-drafts migration fails when not using i18n


laurenskling:
### Required System information

Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0-beta.5
Node/Yarn Version: yarn/1.22.19 npm/? node/v18.20.2 darwin arm64
Edition: Community
Database: sqlite

### Describe the bug

Starting beta.5 runs migration `core::5.0.0-discard-drafts` thats crashes on a missing `locale` field

### Steps to reproduce the behavior

1. have a strapi 4 without i18n
2. install beta 5
3. start
4. crash

### Expected behavior

No crash

n-alonso:
Hi @laurenskling ,

I am not able to reproduce the in `5.0.0-beta.5`.
Im also in a mac, using yarn 1.22, node 18, SQLite and upgraded from v4@latest without i18. `yarn develop` still works just fine and I am able to create and delete content.

Also your reproduction steps are not very descriptive, please in the future try to be as explicit as possible.

I will mark this issue as 'unable to reproduce' until you provide further steps.

laurenskling:
Hi Nicolas, 
thanks for taking the time to debug this.
Sorry for my limited bug report. Somewhere I was sure this migration just forgot about v4 working without having i18n turned on and it just failed to work with that use case. But you mention that you are able to migration from a v4 without i18n, that changes the situation.
My error:

```
⠙ Loading Strapi{ event: 'migrating', name: 'core::5.0.0-discard-drafts' }
[ERROR]  There seems to be an unexpected error, try again with --debug for more information 

MigrationError: Migration core::5.0.0-discard-drafts (up) failed: Original error: select `t0`.`id`, `t0`.`document_id`, `t0`.`locale` from                  
`documents_document` as `t0` where (`t0`.`published_at` is not null) order by `t0`.`id` asc limit 1000 - no such column: t0.locale                          
at /Users/laurenskling/sites/strapi5/node_modules/umzug/lib/umzug.js:151:27                                                                                
at async Umzug.runCommand (/Users/laurenskling/sites/strapi5/node_modules/umzug/lib/umzug.js:107:20)                                                       
at async Object.up (/Users/laurenskling/sites/strapi5/node_modules/@strapi/database/dist/index.js:6239:7)                                                  
at async Object.up (/Users/laurenskling/sites/strapi5/node_modules/@strapi/database/dist/index.js:6263:11)                                                 
[...]

PR #20422, opened 2024-06-03
Handle non-localized content-types in discard-draft migration

Convly:
### What does it do?

Adds a check to conditionally set locale in the `discardDraft`'s params.

### Why is it needed?

This can break app with no i18n
See https://github.com/strapi/strapi/issues/20225

### How to test it?

See reproduction steps in https://github.com/strapi/strapi/issues/20225

### Related issue(s)/PR(s)

fix https://github.com/strapi/strapi/issues/20225

### DIFF
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -18,21 +18,29 @@ export async function* getBatchToDiscard({
   db,
   trx,
   uid,
+  isLocalized,
   batchSize = 1000,
 }: {
   db: Database;
   trx: Knex;
   uid: string;
+  isLocalized: boolean;
   batchSize?: number;
 }) {
   let offset = 0;
   let hasMore = true;
 
+  const fields = ['id', 'documentId'];
+
+  if (isLocalized) {
+    fields.push('locale');
+  }
+
   while (hasMore) {
     // Look for the published entries to discard
     const batch: DocumentVersion[] = await db
       .queryBuilder(uid)
-      .select(['id', 'documentId', 'locale'])
+      .select(fields)
       .where({ publishedAt: { $ne: null } })
       .limit(batchSize)
       .offset(offset)
@@ -74,7 +82,6 @@ const migrateUp = async (trx: Knex, db: Database) => {
       const params: DiscardDraftParams<typeof uid> = { documentId: entry.documentId };
 
       // Only add the locale param if the model is localized
-      // ref. https://github.com/strapi/strapi/issues/20225
       if (isLocalized) {
         params.locale = entry.locale;
       }
@@ -89,7 +96,7 @@ const migrateUp = async (trx: Knex, db: Database) => {
      * Load a batch of entries (batched to prevent loading millions of rows at once ),
      * and discard them using the document service.
      */
-    for await (const batch of getBatchToDiscard({ db, trx, uid: meta.uid })) {
+    for await (const batch of getBatchToDiscard({ db, trx, uid: meta.uid, isLocalized })) {
       await async.map(batch, discardDraft, { concurrency: 10 });
     }
   }
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -1,8 +1,10 @@
 /* eslint-disable no-continue */
-import type { UID } from '@strapi/types';
+import type { UID, Modules, Utils } from '@strapi/types';
 import type { Database, Migration } from '@strapi/database';
 import { async, contentTypes } from '@strapi/utils';
 
+type DiscardDraftParams<TSchemaUID extends UID.ContentType> =
+  Modules.Documents.ServiceParams<TSchemaUID>['discardDraft'];
 type DocumentVersion = { documentId: string; locale: string };
 type Knex = Parameters<Migration['up']>[0];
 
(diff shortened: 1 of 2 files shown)


packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -57,16 +59,31 @@ const migrateUp = async (trx: Knex, db: Database) => {
...
+      if (isLocalized) {
+        params.locale = entry.locale;
+      }
+
       strapi
         .documents(uid)
-        // Discard draft by referencing the documentId and locale
-        .discardDraft({ documentId: entry.documentId, locale: entry.locale });
+        // Discard draft by referencing the documentId (and locale if the model is localized)
+        .discardDraft(params);
```
alexandrebodin, line 85: Shouldn't discardDraft support ignoring the locale instead 🤔  cc @Marc-Roig ?

alexandrebodin (commented):
the PR looks good but sounds like we might want to fix this inside discardDraft (or both)

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
Convly: In the end, we'll probably have to fix it in both. I just saw [that](https://github.com/strapi/strapi/blob/v5/main/packages/core/core/src/services/document-service/repository.ts#L298-L299) in the method itself.

Convly: (even though it appears the first methods are checking if the CT is localized)

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -57,16 +59,31 @@ const migrateUp = async (trx: Knex, db: Database) => {
...
     if (!hasDP) {
       continue;
     }
 
-    const discardDraft = async (entry: DocumentVersion) =>
+    const discardDraft = async (entry: DocumentVersion) => {
+      const params: DiscardDraftParams<typeof uid> = { documentId: entry.documentId };
+
+      // Only add the locale param if the model is localized
+      // ref. https://github.com/strapi/strapi/issues/20225
```
alexandrebodin, line 77: Not sure this is worth mentionning

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
alexandrebodin: Yes, can you create an issue to also fix the fn 🙏

Marc-Roig: True, the discardDraft should be smart enough to ignore it :/

Convly: Actually, it seems like it is for the most part. There are checks for isLocalized inside each internationalization method used for the params.

The only bit I'm uncertain about is https://github.com/strapi/strapi/blob/v5/main/packages/core/core/src/services/document-service/repository.ts#L298-L299
Should we keep that as is? Are we assuming relations have locale in any case :thinking:

Marc-Roig: Mmm we might need to not include locale if the relation target is not localized.
We are populating the target locale & document id fields, as that information is necessary in some scenarios (non localized CT -> localized CT)

We could manually ignore the "locale" field in that populate utility if the CT does not have I18n enabled. Wdyt?

Convly: Yeah, indeed. In any case, we'll have to work a bit on those generated populate queries. We also have another issue where it doesn't respect morph relations and sends regular queries instead of fragments.

Convly: Not sure it's the responsibility of the `getDeepPopulate` to sanitize the incoming selected fields though :thinking: 

I'll think a bit about it

alexandrebodin (approved):

innerdvations (commented):
There's a build error from the discardDraft typings, but other than that LGTM

Marc-Roig (approved):

packages/core/core/src/migrations/draft-publish.ts
```diff
@@ -31,22 +33,37 @@ const enableDraftAndPublish = async ({ oldContentTypes, contentTypes }: Input) =
       const oldContentType = oldContentTypes[uid];
       const contentType = contentTypes[uid];
 
+      const isLocalized = strapi
+        .plugin('i18n')
+        .service('content-types')
+        .isLocalizedContentType(oldContentType);
```
Convly, line 39: I wasn't sure whether we wanted to target the old CT or the new one since I'm not sure about what we're trying to achieve exactly. 

cc @Marc-Roig for confirmation

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
Marc-Roig: This is a problem we have faced in many other places, where we have to omit the "locale" or "publishedAt" field.
In any case, the document service assumes all content types include locale and publishedAt, even if those features are not enabled on the CT, we do include those columns.

Marc-Roig: I believe this migration should include locale for all content type shttps://github.com/strapi/strapi/blob/eb316acc5c08cd34e14e22ce36c6e9806e701b70/packages/core/database/src/migrations/internal-migrations/5.0.0-03-locale.ts

shouldn't this fix the issue for v4->v5 migrations? 🤔

Convly: Won't this be run before the one you've mentioned?

Convly: Yeah, if we change the order of execution, the other migration might make this PR completely useless :slightly_smiling_face:

maccomaccomaccomacco:
You would make a PM happy by resolving this PR, it's preventing me to have a data migration done

Convly:
> You would make a PM happy by resolving this PR, it's preventing me to have a data migration done

Even on the latest beta (14) ?

ildfreelancer:
@Convly yes, the issue is still there while migrating from v4 to v5 (using beta(14))
image

Convly:
> @Convly yes, the issue is still there while migrating from v4 to v5 (using beta(14)) image

Wait, how is it possible to expect localized admin users, is it the wanted behavior in the new i18n implementation @Marc-Roig?

Marc-Roig:
>Wait, how is it possible to expect localized admin users, is it the wanted behavior in the new i18n implementation
@Convly all content types are expected to have the locale column atm

ildfreelancer:
image

The issue above is gone away with v5 rc (latest version), however, I got another error.
@Marc-Roig

Marc-Roig:
@ildfreelancer hey! thank you for the report, is that the full error message?

If not, could you share a bit more of the schemas you have inyour proejct?

Marc-Roig:
@ildfreelancer It would be extremely helpful to have a bit more context on the content types you have, would it be possible to share those schemas (private if you prefer)

Issue #17604, opened 2023-08-08
Cannot Save component even if the fields are unique.


vaidikMirae:
### Required System information

- Node.js version: 18.12
- NPM version: 9.6
- Strapi version: v4.12.1
- Database: Postgresql
- Operating system: Linux
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When I try to save the single content type, which has a links component with a unique field that throws an error anytime I try to save the content, it throws an error even if I provide the unique content.
### Steps to reproduce the behavior

1. Go to single type content
2.  Create a component with a unique field inside it.
3.  Fill the content in single type
4.  Try to save the content.

### Expected behavior

Save the single type content with the unique fields inside it.

### Screenshots

![image](https://github.com/strapi/strapi/assets/140403673/a6076370-c885-4eb3-9036-929922eb2c72)

pjr94:
I have also experienced this issue. I have a ParticipantID field which relates to an ID provided by an external system, so I cannot simply change the ID in order to save the Entry as it would be desynced with the external system. I've not run into the issue yet, but if the client asks me to update some data in the collection (eg phone number or whatever), I will not be able to do this without changing the ParticipantID, which would be a big problem and break the relationship with the external system unless they can update it there too. Either way, its a bad work around and unproffessional.

TL:DR Yes it is very important to fix this for business reasons!

The Issue:
I think the issue is more than the unique field is comparing its new content with the existing old content and recognising it as not unique. 

Solution: There needs to be an exemption that unique fields can be the same in the same when it relates to the same Entry, so you can edit the Entry without changing the unique field.

jhoward1994:
This issue has been fixed in v5 from 5.0.0-beta.13 onwards via https://github.com/strapi/strapi/pull/20427

Issue #16918, opened 2023-06-05
A unique field on a component of a single-type leads to a faulty 'uniqueness check'


maxbethke:
### Required System information

- Node.js version: v16.17.1
- NPM version: 9.1.2
- Strapi version:
```
"dependencies": {
    "@strapi/plugin-i18n": "4.10.7",
    "@strapi/plugin-users-permissions": "4.10.7",
    "@strapi/strapi": "4.10.7",
    "mysql": "2.18.1"
  },
```
- Database: MySQL Server 5.7.41-1.el7
- Operating system: MacOS 12.6
- Is your project Javascript or Typescript: TS

### Describe the bug

A unique field on a component of a single-type leads to a faulty 'uniqueness check'

### Steps to reproduce the behaviour

- Create a new single-type with
   - a repeatable component on it
   - some other field, e.g. short text
- Create a text field on the component (short text, **unique**)
- Add a component to the single-type
- Save the single-type
- Put a value in the unique field of the component
- Save the single-type
- Edit any other field (just so you can hit save again)
- Save the single type

At this point error gets displayed saying the value in the unique field of the component is not unique, even though it is.

### Expected behavior
After the steps above, the single-type should be able save, because the unique field's value is unique.

### Idea
I might be very well wrong, but to me it looks like on the update() of the single-type, the 'uniqueness check' for the unique field does not only look at array of components that would result from update, but at the old data merged with the new data. Thus resulting in the error, because the unique value now exists twice, once in the old record of the edited component and once in the new record.

rbsam176:
If anyone has any workarounds (other than not making the field unique) while this is being fixed it would be great to hear them, it's affecting our app significantly.

jhoward1994:
Hi, thank you for the bug report. We are aware of this issue with unique fields and are actively working on it.

We also plan to extend the capabilities of unique fields in future, please see https://github.com/strapi/strapi/issues/12541#issuecomment-1503127208

Kartik4152:
Any updates on this?

DenuxPlays:
Also happens on collection-type.
Please update title!

jhoward1994:
This issue has been fixed in v5 from 5.0.0-beta.13 onwards via https://github.com/strapi/strapi/pull/20427

Issue #12541, opened 2022-02-16
Unique attribute in a repeatable component field leads save error when insert more component entries.


smoothdvd:
### Describe the bug

If a repeatable component has a field, which has unique attribute, insert this component multiple times will lead failure on unique validation.

### Steps to reproduce the behavior

1. Create a repeatable component name "foo"
2. Create a field named "title" in foo component, and enable "unique" validation on it
3. Add component foo to a content-type TEST
4. Add first entry with component  foo ( title => "test1") and save entry
5. Add second entry with component foo ( title => "test2" and save entry
6. See error: Failed to load resource: the server responded with a status of 400 (Bad Request)

```
{
    "data": null,
    "error": {
        "status": 400,
        "name": "ValidationError",
        "message": "14 errors occurred",
        "details": {
            "errors": [
                {
                    "path": [
                        "categories",
                        "0",
                        "name"
                    ],
                    "message": "This attribute must be unique",
                    "name": "ValidationError"
                },
...
```

### Expected behavior

Correct save multiple repeatable component which has contain unique validation field.

### Code snippets

If applicable, add code samples to help explain your problem.

### System

- Node.js version: 14.x
- NPM version: 6.14.15
- Strapi version: 4.0.7
- Database: PostgreSQL 13
- Operating system: macOS 12.2.1

the94air:
I am running into the same issue with the same procedures as @mmarseu. For now the only way around is to disable the field unique option.

mmarseu:
@gu-stav The restaurant example was only for illustration. My real use case is a different one. I was evaluating strapi to use as a backend for a product information site we'll be building in future. Take that into account, when reading my description below: _I don't actually know strapi very well and might have used it in unintended or inefficient ways._

We have a product portfolio with lots of _models_ belonging to different _classes_ and different _series_. That concept seems to map rather well to strapi collections. 
However, each model can also come in different _variants_. The variants of a model share a lot of common information but also have some pieces of information which are different. Among those is the _part number_ (or order number, if you like), which is unique.

My idea was to have _model_ be a collection type and _variant_ be a repeatable component on each _model_. That way, I can put common information in the fields of _model_ and the rest into the component. _Part number_ would have been part of the component and would have been unique.
Obviously, I would still have liked to be able to edit common information on the _model_, even if there are already any _variants_.

In more generic terms, personally, I would love to have the choice to have unique fields inside components be either:
- unqiue within the context of the parent content-type
- globally unique.
[...]

Issue #12248, opened 2022-01-19
Unique and localisation - checks in all languages


HannesOberreiter:
### Describe the bug

Unique fields are checked in all languages and therefore you cannot same value in different languages. To be honest, I don't know if that is expected.

### Steps to reproduce the behavior

1. Create a Content-Type with a text-field with attributes "**Enable Unique**" and "**Enable localization for this field**" enabled.
2. Create a "New Entry" for this Content-Type and write for example the text value "**Tourist**" for English
3. Now create for example the "German" localisation which would be the same: "**Tourist**"
4. You get an error that the value is not "Unique"

### Expected behavior

The "unique" check should only apply to one language.

### System

- Docker: strapi/base:latest (tag 14)
- Strapi version: 4.0.4

### Additional context

Somewhat similar issue: https://github.com/strapi/strapi/issues/10694

Cheers
Hannes

yanniskadiri:
This is an issue related to this one too: #10107
It's not the same use case but it's the same cause. It's a feature request we will address in the future. Thanks for sharing your use case :)

jhoward1994:
This issue has been fixed in v5 from 5.0.0-beta.13 onwards via https://github.com/strapi/strapi/pull/20427

PR #20427, opened 2024-06-04
Component unique fields

jhoward1994:
### What does it do?

Re implements component unique fields in V5

According to the logic:

- Component unique fields are validated against values in other entries of their parent content type in the same dimensions (locale and publication state)
- All unique values in a repeatable must be unique in the current entity
- The order of components in a repeatable does not matter. Validation will fail if any other entry (of content type + dimensions) holds the same value in the same component field anywhere in the repeatable

### Why is it needed?

Fix existing V4 bugs and implement core v5 functionality

### How to test it?

API tests
E2E tests

### Related issue(s)/PR(s)

DX-1348

WIP 

- [x] E2E tests
- [x] Update types for new CTs in API tests
- [x] Add repeatable cases to E2E


jhoward1994:
Hey @innerdvations & @Bassel17 the WIPs are complete now so this is ready for review 🙏🏻

tests/api/core/strapi/document-service/resources/types/contentTypes.d.ts
```diff
@@ -672,6 +672,8 @@ export interface ApiArticleArticle extends Struct.CollectionTypeSchema {
       'api::article.article'
     >;
     locale: Schema.Attribute.String;
+    idenitifers: Schema.Attribute.Component;
```
innerdvations, line 675:
```suggestion
    identifiers: Schema.Attribute.Component;
```

tests/api/core/strapi/document-service/resources/types/contentTypes.d.ts
```diff
@@ -672,6 +672,8 @@ export interface ApiArticleArticle extends Struct.CollectionTypeSchema {
...
+    repeatableIdenitifers: Schema.Attribute.Component;
```
innerdvations, line 676:
```suggestion
    repeatableIdentifiers: Schema.Attribute.Component;
```

packages/core/types/src/modules/entity-validator.ts
```diff
@@ -4,12 +4,10 @@ import type * as UID from '../uid';
 import type * as EntityService from './entity-service';
 
 type Entity = {
-  id: ID;
+  id: number;
```
innerdvations, line 7: Are we sure this can't ever be string?

innerdvations (commented):
Overall looks great! Just a few small comments

innerdvations:
And API tests are failing, but that may be the issue on v5/main. Once we know they're fixed we can run them again.

packages/core/types/src/modules/entity-validator.ts
jhoward1994: Yes I think so. The entity validator works with entities not documents everywhere

innerdvations (approved):
LGTM!

innerdvations (approved):
I think we're basically there!

jhoward1994:
Known issue with dynamic zones tracked here https://github.com/strapi/strapi/issues/20546

jhoward1994:
This PR fixes the following issues in Strapi v5

closes https://github.com/strapi/strapi/issues/12541 
closes https://github.com/strapi/strapi/issues/12248
closes https://github.com/strapi/strapi/issues/10107
closes https://github.com/strapi/strapi/issues/17604
closes https://github.com/strapi/strapi/issues/16918

PR #20433, opened 2024-06-05
test: backport cli tests

innerdvations:
### What does it do?

- copies the `v5/main` CLI tests to the `develop` branch
- rename package.json script `test:clean` to be `test:api:clean`
- makes a new `test:clean` that runs all the test cleaning scripts

### Why is it needed?

- CLI tests were missing from `develop`
- there is often a yalc issue that will cause incomplete CLI or e2e test setups to break each other, so there needs to be a universal test clean script
- Cloud CLI for v4 needs the infrastructure available so it can be tested

### How to test it?

all tests should pass

### Related issue(s)/PR(s)

DX-1445


Convly (commented):
The code itself looks good, waiting for the tests to actually pass in the CI

joshuaellis (commented):
I think you need to update the workflows to include the test suite?

packages/utils/typescript/tsconfigs/.tsbuildinfo
joshuaellis, line 1: are we including this?

joshuaellis (approved):

packages/utils/typescript/tsconfigs/.tsbuildinfo
innerdvations: Nope. Just added it to gitignore too so that this doesn't happen again, it's a leftover from when you switch from a v5 branch to a v4 branch but might as well ignore it to save this headache.

PR #20453, opened 2024-06-06
Expand nested components

Bassel17:
### What does it do?

- Expands the limit of nesting components to 6 levels of nesting
- calculate the Depth of components and there child components

### Why is it needed?

- to have the possibility to put a number limit on nesting components

### How to test it?

- you should only be able to nest up to 6 components from the CTB
- you should not be able to add a component field to a component at the 6th level
image

- you should only be shown components that you can add within the limit in the list of components

image


packages/core/content-type-builder/admin/src/utils/getMaxDepth.ts
joshuaellis, line 1: I think you could improve your comment writing here, at the moment you're describing what we can currently see below. But what about explaining when you'd use these functions, what it means to have children depth, what it is component depth etc.

packages/core/content-type-builder/admin/src/utils/getMaxDepth.ts
```diff
@@ -0,0 +1,79 @@
+import type { NestedComponent } from '../components/DataManagerProvider/utils/retrieveNestedComponents';
+import type { Internal } from '@strapi/types';
+
+type ChildComponent = {
+  repeatable: boolean;
+  component: Internal.UID.Component;
+  name?: string;
+  required?: boolean;
+  min?: number;
+};
```
joshuaellis, line 10: Why is a child component different to a regular component? This is the sort of context you should be writing about :)

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveComponentsThatHaveComponents.ts
```diff
@@ -3,16 +3,15 @@ import get from 'lodash/get';
 import { makeUnique } from '../../../utils/makeUnique';
 
 import type { Component, AttributeType, Components } from '../../../types';
-import type { Internal } from '@strapi/types';
 
 const retrieveComponentsThatHaveComponents = (allComponents: Components) => {
   const componentsThatHaveNestedComponents = Object.keys(allComponents).reduce(
-    (acc: Internal.UID.Component[], current) => {
+    (acc: any, current) => {
```
joshuaellis, line 9: I think this should be typed accurately, not loosely 😄

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveNestedComponents.ts
```diff
@@ -1,24 +1,55 @@
...
-  const nestedComponents = Object.keys(appComponents).reduce((acc: any, current) => {
-    const componentAttributes = appComponents?.[current]?.schema?.attributes ?? [];
-    const currentComponentNestedCompos = getComponentsFromComponent(componentAttributes);
+export type NestedComponent = {
+  component: Internal.UID.Component;
+  parentCompoUid?: Internal.UID.Component[];
+};
 
+export const retrieveNestedComponents = (appComponents: any): NestedComponent[] => {
+  const nestedComponents = Object.keys(appComponents).reduce((acc: any, current: any) => {
```
joshuaellis, line 9: Better typing please.

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/tests/retrieveNestedComponents.test.ts
```diff
@@ -49,7 +49,83 @@ describe('CONTENT TYPE BUILDER | COMPONENTS | DataManagerProvider | utils | retr
...
+          collectionName: 'components_openingperiods',
+          attributes: [{ component: 'default.dish', type: 'component', name: 'dish' }],
+        },
+      },
+    };
+
+    const expected = [
+      {
+        component: 'default.dish',
+        parentCompoUid: ['default.closingperiod', 'default.openingperiod'],
```
joshuaellis, line 126: I find this confusing, why is `parentCompoUID` an array if it's a UID?

packages/core/content-type-builder/admin/src/components/FormModal/utils/getAttributesToDisplay.ts
```diff
@@ -34,8 +35,15 @@ export const getAttributesToDisplay = (
     ];
   }
 
-  if (canAddComponentInAnotherComponent) {
-    return [defaultAttributes, ['component']];
+  // this will only run when adding attributes to components
```
joshuaellis, line 38: this is a good example of a well written comment :)

joshuaellis (changes requested):
Some initial feedback, overall i find the functions confusing and it's unclear why you've done what you've done at the moment, but I think thats a lack of explanation rather than a criticism of your solution :)

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/tests/retrieveNestedComponents.test.ts
Bassel17: it's multiple component UIDs, a component can have multiple different parents at the same level

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveNestedComponents.ts
```diff
@@ -1,24 +1,55 @@
...
+        component,
+        parentCompoUid,
+      });
     }
 
     return acc;
   }, []);
 };
+
+// merge components different parents if they exist
```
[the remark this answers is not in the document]
Bassel17: if we have a component with multiple different parents, we merge the object into one with an array of it's parent components uids

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/tests/retrieveNestedComponents.test.ts
joshuaellis: I think you should write an explanation and maybe change the name to be clearer, the way you've written it at the moment gives the illusion it's singular or nested i.e. dish is inside openingperiod, which is inside closingperiod 🤔

joshuaellis (commented):
I've not QAd but the code looks much better.

packages/core/content-type-builder/admin/src/utils/getMaxDepth.ts
```diff
@@ -0,0 +1,95 @@
+import type { ComponentWithChildren } from '../components/DataManagerProvider/utils/retrieveComponentsThatHaveComponents';
+import type { NestedComponent } from '../components/DataManagerProvider/utils/retrieveNestedComponents';
+import type { Internal } from '@strapi/types';
+
+const findComponent = <T extends { component: Internal.UID.Component }>(
+  componentName: Internal.UID.Schema,
```
jhoward1994, line 6:
```suggestion
  componentUID: Internal.UID.Schema,
```

packages/core/content-type-builder/admin/src/utils/getMaxDepth.ts
```diff
@@ -0,0 +1,95 @@
...
+  components: Array<T>
+) => {
+  return components.find((c) => c.component === componentName);
```
jhoward1994, line 9: Could there be multiple components matching the UID in this array?

packages/core/content-type-builder/admin/src/utils/tests/getMaxDepth.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    expect(componentsMaxDepth).toEqual(0);
+  });
+
+  test('should accurately give the max depth of components children', () => {
+    const componentsMaxDepth = getChildrenMaxDepth('default.openingtimes', components);
+
+    expect(componentsMaxDepth).toEqual(4);
+  });
+});
```
jhoward1994, line 95: Is this possible to write tests for getComponentDepth too ?

The tests helped me understand

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveNestedComponents.ts
```diff
@@ -1,24 +1,60 @@
...
     return [...acc, ...currentComponentNestedCompos];
   }, []);
 
-  return makeUnique(nestedComponents);
+  return mergeComponents(nestedComponents);
 };
 
-const getComponentsFromComponent = (componentAttributes: any) => {
-  return componentAttributes.reduce((acc: any, current: any) => {
+const getComponentsFromComponent = (
```
jhoward1994, line 23: Some docs / a new name for this fucntion would be nice I think. Naming is tricky
though if I understand it right it's kinda:

getNestedComponents? getComponentsNestedWithinComponent? 😄

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveNestedComponents.ts
```diff
@@ -1,24 +1,60 @@
...
+        component,
+        parentCompoUid,
+      });
     }
 
     return acc;
   }, []);
 };
+
+// merge duplicate components
```
jhoward1994, line 40:
```suggestion
// Merge duplicate components
```

sorry 🫣

packages/core/content-type-builder/admin/src/components/SelectComponent.tsx
```diff
@@ -66,8 +72,11 @@ export const SelectComponent = ({
   );
 
   if (isAddingAComponentToAnotherComponent) {
-    options = options.filter((option) => {
-      return !componentsThatHaveOtherComponentInTheirAttributes.includes(option.uid);
+    options = options.filter(({ uid }: any) => {
+      const maxDepth = getChildrenMaxDepth(uid, componentsThatHaveOtherComponentInTheirAttributes);
+      const componentDepth = getComponentDepth(targetUid, nestedComponents);
+      const totalDepth = maxDepth + componentDepth;
+      return totalDepth <= MAX_COMPONENT_DEPTH;
```
jhoward1994, line 79: I'm just wondering if we should provide feedback to the user about why
components have been excluded from the list?

I guess we had the same lack of feedback problem before but it was probably more
obvious what was going on as any nesting was filtered out (is that right?)

Now the max depth is set at 6 it could be harder to work out why the filtering
is happening. wdyt?

packages/core/content-type-builder/admin/src/utils/getMaxDepth.ts
Bassel17: nah it's unique, only one will exist

packages/core/content-type-builder/admin/src/components/SelectComponent.tsx
Bassel17: yeah some form of feedback would've been nice,
however it's unclear to me how we could give feedback either, the current way things look
there isn't much space or place to display useful, info, I mean maybe I could put a simple general message such as `components that would exceed the nested limit of 6 are excluded from the list`

packages/core/content-type-builder/admin/src/components/DataManagerProvider/utils/retrieveNestedComponents.ts
Bassel17: yeah `getComponentsNestedWithinComponent` could work well

packages/core/content-type-builder/admin/src/components/SelectComponent.tsx
jhoward1994: Yeah it's a bit tricky maybe we should run this by Marco and track it as an improvement after this is merged ?

jhoward1994:
It looks like there is a bug in the CTB where I'm unable to scroll on the list of components when adding one to another. Do you see this too?

jhoward1994:
Sorry for the long comment. I'm seeing some strange behaviour in the CTB.

I've let up Level 1 - Level 6 components (level 1 has level 2 nested, 2 has 3 etc etc). I was able to set this up by creating all levels with no component fields and then starting from Level 1 I added Level 2 and on and on.

My test was going to be to try and add Level 1 to another component and see that it is excluded from the list (as we've now exceeded 6 level, does that make sense?)

Instead I'm able to see all levels in the component drop down but when I add any I get an error from the backend like

```
http://localhost:1337/content-type-builder/components
POST

{
    "error": {
        "name": "ValidationError",
        "details": {
            "errors": [
                {
                    "path": [
                        "component",
                        "attributes",
                        "L5",
                        "component"
                    ],
                    "message": "level5 already is a nested component. You cannot have more than one level of nesting inside your components.",
                    "name": "ValidationError"
                }
            ]
        },
        "message": "level5 already is a nested component. You cannot have more than one level of nesting inside your components."
    }
}

```

Bassel17:
> Sorry for the long comment. I'm seeing some strange behaviour in the CTB.
> 
> I've let up Level 1 - Level 6 components (level 1 has level 2 nested, 2 has 3 etc etc). I was able to set this up by creating all levels with no component fields and then starting from Level 1 I added Level 2 and on and on.
> 
> My test was going to be to try and add Level 1 to another component and see that it is excluded from the list (as we've now exceeded 6 level, does that make sense?)
> 
> Instead I'm able to see all levels in the component drop down but when I add any I get an error from the backend like
> 
> ```
> http://localhost:1337/content-type-builder/components
> POST
> 
> {
>     "error": {
>         "name": "ValidationError",
>         "details": {
>             "errors": [
>                 {
>                     "path": [
>                         "component",
>                         "attributes",
>                         "L5",
>                         "component"
>                     ],
>                     "message": "level5 already is a nested component. You cannot have more than one level of nesting inside your components.",
>                     "name": "ValidationError"
>                 }
>             ]
>         },
>         "message": "level5 already is a nested component. You cannot have more than one level of nesting inside your components."
>     }
> }
> ```

well interesting, well in terms of being viewable that's accurate cause that's how i defined level 6, you'd have to check the parent of level 1 and see, anyways the server error is strange to me I didn't face it, perhaps it's the method things have been added, I wasn't aware of any server limitations on nesting components

Bassel17:
> It looks like there is a bug in the CTB where I'm unable to scroll on the list of components when adding one to another. Do you see this too?

yep seems to be new didn't exist before I merge `v5/main` into the PR, more things to fix, I'll create some tickets with all the CTB front issues

packages/core/content-type-builder/server/src/controllers/validation/types.ts
```diff
@@ -219,24 +219,8 @@ const getTypeShape = (
...
-                  path: this.path,
-                  message: `${targetCompo.modelName} already is a nested component. You cannot have more than one level of nesting inside your components.`,
-                });
-              }
-              return true;
-            },
-          })
-          .required(),
+        // TODO: Add correct server validation for nested components
+        component: yup.string().required(),
```
jhoward1994, line 219: Maybe you had plans to do this but I think if we're saying that the max level of nesting is 6 that should be enforced on the frontend and backend. wdyt?

Bassel17: we actually never planned to enforce on the backend, and whatever was previously here was not working from all places anyways, so yeah would leave it to a different point in time since this will just add more to the complexity and  no one should be hitting this point anyways

jhoward1994: I still think the validation here should match our intentions for the feature
and prevent incorrect setups being made through the backend. Could we somehow
count the deepest level the component is being used at in a util & use it
similarly to `hasComponent` ?

wdyt? @innerdvations

jhoward1994 (approved):
QA'd and everything is working 👍🏻

I do think we should track this in JIRA and add some more specific BE validation for the components https://github.com/strapi/strapi/pull/20453#discussion_r1646467255

packages/core/content-type-builder/server/src/controllers/validation/types.ts
Convly: What would it take to add such validation, is it only about checking the number of nested components in the received data?

Anyway, we should try to have the same rules in the admin and server. So I would add this check here too if it's possible and doesn't deviate from the original PR goal.

Bassel17: will do that at a second iteration, we would be adding the limitation rules to the server as well
but the goal here was to add the limitations only on the frontend considering the server already accepted any amount of nested components, and too also move with quickly.
we would basically have to duplicate the same depth counting algorithm on the frontend to add the rule, it's a little complex but should get the job done

PR #20454, opened 2024-06-06
fix: update remaining monorepo dependencies after beta.9 update

Marc-Roig:
### What does it do?
Some api tests were randomnly failing, looks like the cause was some project dependencies were not properly updated by lerna during the release.


innerdvations (approved):
LGTM, thanks!

markkaylor (approved):

PR #20466, opened 2024-06-07
fix(content-type-builder): draft and publish toggle

jhoward1994:
### What does it do?

Fixes buggy behaviour of the draft and publish confirmation pop up in the CTB

### Related issue(s)/PR(s)

DX-1419


innerdvations (approved):
LGTM, tested and it works.

I will work today to get the e2e tests enabled again so that this will be tested.

joshuaellis (approved):

PR #20467, opened 2024-06-07
fix(chore): increase specificity of the Lightning icon color

simotae14:
### What does it do?

Fix the color issue that we have in the Left menu with the Lightning icon color

### Why is it needed?

Because we want to use this color for the icon
Schermata 2024-06-07 alle 16 37 08
So to override the style with higher specificity I used this trick recommended in the styled components doc
https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

### How to test it?

Open a Strapi App with Community Edition license


remidej (approved):
didn't know about this technique, but it seems appropriate here 👍

PR #20468, opened 2024-06-07
chore: put ee LICENSE in correct place

innerdvations:
** Still determining the best way to handle this internally so that it doesn't happen again **

### What does it do?

Includes the ee/LICENSE file

### Why is it needed?

It should be there, it was missing after switching to releasing typescript dist dirs

### How to test it?

`npm pack` in the admin package and see that the license file now exists where it didn't before

### Related issue(s)/PR(s)

DX-1453

### DIFF
--- a/packages/core/admin/package.json
+++ b/packages/core/admin/package.json
@@ -49,6 +49,7 @@
   },
   "files": [
     "./dist",
+    "ee/LICENSE",
     "strapi-server.js",
     "index.js",
     "index.d.ts"


joshuaellis:
do you want to do an experimental and we can see what was published to npm?

Boegie19:
Um there also is an EE folder in here.
https://github.com/strapi/strapi/blob/f4ff842a3cb7b83db540bee67554b704e042b042/packages/core/strapi/src/ee/LICENSE.txt#L1

innerdvations:
> Um there also is an EE folder in strapi it self.
> 
> https://github.com/strapi/strapi/blob/f4ff842a3cb7b83db540bee67554b704e042b042/packages/core/strapi/src/ee/LICENSE.txt#L1

Thanks, I missed that one!

innerdvations:
Closing this until we decide how to best handle it

PR #20469, opened 2024-06-07
feat: add language selector to the blocks code editor

remidej:
### What does it do?

- Adds a floating select to specify the language of all code blocks in Blocks.
- Sets up the first end to end test for Blocks (because it's impossible to test in unit tests, since JSDom doesn't support contenteditable)
### Why is it needed?

- Users need to know which language a code block uses, so they can use tools like Prism to do syntax highlighting in their frontend
- We'll be able to do syntax highlighting in blocks pretty soon

### How to test

Create a code block in a Blocks editor. The select should only appear when your code block is focused. The default value should be plain text. You should be able to change it. When you do, you should be able to save the change


SalahAdDin:
Could you add a screenshot for it?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/Code.tsx
```diff
@@ -1,20 +1,240 @@
...
+import { useSelected, type RenderElementProps, useFocused, ReactEditor } from 'slate-react';
 import { styled } from 'styled-components';
 
-import { type BlocksStore } from '../BlocksEditor';
+import { useBlocksEditorContext, type BlocksStore } from '../BlocksEditor';
 import { baseHandleConvert } from '../utils/conversions';
 import { pressEnterTwiceToExit } from '../utils/enterKey';
 import { type Block } from '../utils/types';
 
+const languages: { value: string; label: string }[] = [
```
Feranchz, line 15: I know we are using it only here but considering its extension what do you think about moving this to a constants file? We do the same, for example, with permissions constants on plugins 🤔

Feranchz (approved):
This is such a great work 🚀

tests/e2e/tests/content-manager/blocks.spec.ts
joshuaellis, line 1: Whats your rationale behind doing an e2e test for this? :) We normally use e2e to do whole user journeys as opposed to an integration test which is what you're trying to do. The reason being is e2e tests are slowwwwww (and soon will be required).

remidej: I wish I could do just unit tests, but unfortunately it wouldn't work, because .

This is why so far all Blocks tests have been testing our functions directly, rather than simulating user interactions (e.g. we test `block.handleEnterKey` instead of `user.keyboard.press('Enter')`. But in this case I don't think it makes sense to expose a function, and we have the e2e tooling easily available now

joshuaellis: Aha, that makes perfect sense! I knew there'd be a reason. I was just curious, thank you 😄

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/tests/Code.test.tsx
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable testing-library/no-node-access */
-import { render, screen } from '@testing-library/react';
+import { act } from 'react';
```
simotae14, line 2: do you need this one?

remidej: good catch, i don't

simotae14 (approved):
Overall it looks good, well done

PR #20479, opened 2024-06-10
feat(cli): use project name from package json by default

abdallahmz:
### What does it do?

When prompting the user for the name of the project during the deployment process, the setDefaultName function sets the default project name to the name of the existing Strapi project (retrieved from package.json), instead of using a generic default name like "my-strapi-project".

- Also added unit testing config to the cli/cloud package.
To run: `yarn test:unit` from the cloud folder.

### Why is it needed?

This change is needed to improve the user experience during the deployment process

### How to test it?

- you need to have Strapi Cloud services running locally (main server + cli-api)
- run `yarn build` in the strapi/strapi folder containing this code's repository
- create a new strapi CMS project
- from the strapi CMS project: call the built `strapi.js` command to run the `login` command
- Then, use the `deploy` command:
When asked for the name of the project, check that the default value is the name of your existing project (as specified in package.json), not "my-strapi-project".

Result:

Screenshot 2024-06-14 at 11 28 35

### Related issue(s)/PR(s)


packages/cli/cloud/src/create-project/utils/set-default-name.ts
```diff
@@ -0,0 +1,14 @@
...
+export function setDefaultName(
+  newDefaultName: string,
+  questions: ReadonlyArray<DistinctQuestion<ProjectAnswers>>,
+  defaultValues: Partial<ProjectAnswers>
+) {
+  defaultValues.name = newDefaultName;
+  const nameQuestion = questions.find((q) => q.name === 'name');
+  if (nameQuestion) {
+    nameQuestion.default = newDefaultName;
+  }
```
nathan-pichon, line 13: Can you avoid mutating parameters, please?
I believe we want to go towards functional programming and pure functions. 👍

abdallahmz: Nice thought, function updated!

packages/cli/cloud/src/create-project/utils/apply-default-name.ts
```diff
@@ -0,0 +1,22 @@
...
+  newQuestions: ReadonlyArray<DistinctQuestion<ProjectAnswers>>;
+  newDefaultValues: Partial<ProjectAnswers>;
+} {
+  const newDefaultValues = { ...defaultValues, name: newDefaultName };
+
+  const newQuestions = questions.map((question) => {
+    if (question.name === 'name') {
+      return { ...question, default: newDefaultName };
+    }
+    return { ...question };
```
nathan-pichon, line 18: Do we really need to create a fresh object?
```suggestion
    return question;
```

If the point was to copy the object to not keep the reference, then you might want to do a deepClone instead of destructuring. 🤔

abdallahmz: Went for the JSON parse/stringify to do the copy, although it'll only work with simple objects/array (The "...question" would've at least kept the functions if they were added to the question)

The only solution without any "effective" drawback would be to add a proper class or factory to instantiate the "question" and "default values" objects and use that instead. You think it's worth refactoring now?

nathan-pichon: I think you didn't responded to my first question 😄 

But in any case, if we really want to copy, we can use https://lodash.com/docs/4.17.15#cloneDeep

abdallahmz: Indeed, I want to avoid having the same references. Initial code works for the current code since we only have scalar values so far.
Using deepClone now, should be better!

nathan-pichon (approved):

PR #20481, opened 2024-06-10
test: temporarily disable broken edit ctb tests

innerdvations:
### What does it do?

Skip two tests on the CI that work locally but always fail on github

### Why is it needed?

A compromise to allow CI to pass while we work on finding the issue

### How to test it?

Tests should pass on the CI, and the "broken" tests should still run locally

### Related issue(s)/PR(s)


Marc-Roig (approved):
looks good to me to unblock the CI.

Do you know who will be working on the fix?

innerdvations:
> looks good to me to unblock the CI.
> 
> Do you know who will be working on the fix?

I'll probably be the one to fix it, but I already spent too much time looking at it last week and I'd rather unblock people in the meantime. I'm hoping to have some time to look at it again later this week and solve it.

It has to do with the server restart after the file reset; the browser ends up stuck in the "waiting for restart" loop even after the server comes back up, which doesn't happen locally.

PR #20486, opened 2024-06-10
Replace lock icon with lightning icon in the Settings and Navbar links

simotae14:
### What does it do?

Replace the Lock icon used for the features not available without EE license with the Lightning icon

### Why is it needed?

We want to replace the Lock icon that shows users they don’t have access to Releases and other features in the Community Edition by a lightning flash icon so that these features are perceived as boosters if users upgrade their app (rather than blockers).

### How to test it?

Open a Strapi App with Community Edition

### Related issue(s)/PR(s)

CS-808

### DIFF
--- a/docs/docs/docs/01-core/admin/01-ee/00-intro.md
+++ b/docs/docs/docs/01-core/admin/01-ee/00-intro.md
@@ -35,7 +35,7 @@ Everytime a new EE feature is added in Strapi, in the settings menu, you should 
             to: '/settings/purchase-new-ee-feature',
             id: 'new-ee-feature',
             // TODO: to replace with another name in v5
-            lockIcon: true,
+            toBuy: true,
           },
         ]
       : []),
--- a/docs/docs/docs/01-core/admin/01-ee/00-intro.md
+++ b/docs/docs/docs/01-core/admin/01-ee/00-intro.md
@@ -34,8 +34,7 @@ Everytime a new EE feature is added in Strapi, in the settings menu, you should 
             },
             to: '/settings/purchase-new-ee-feature',
             id: 'new-ee-feature',
-            // TODO: to replace with another name in v5
-            toBuy: true,
+            licenseOnly: true,
           },
         ]
       : []),
--- a/packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
@@ -83,7 +83,7 @@ const SettingsNav = ({ menu }: SettingsNavProps) => {
                   position="relative"
                 >
                   {formatMessage(link.intlLabel)}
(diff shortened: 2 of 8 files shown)


docs/docs/docs/01-core/admin/01-ee/00-intro.md
```diff
@@ -35,7 +35,7 @@ Everytime a new EE feature is added in Strapi, in the settings menu, you should
             to: '/settings/purchase-new-ee-feature',
             id: 'new-ee-feature',
             // TODO: to replace with another name in v5
-            lockIcon: true,
+            toBuy: true,
```
joshuaellis, line 38: wouldn't `enterpriseOnly` be more appropriate?

simotae14: it could be a good name, tbh it can be served also with other cloud licenses, maybe we can use something like `paidLicenseOnly` or just `licenseOnly`, wdyt?

remidej: >tbh it can be served also with other cloud licenses

That's true but we don't really make that distinction inside the codebase, I don't think we need to start worrying about it now. For example `admin/ee` hosts most of the code that's protected by a license, no matter if it's in the cloud plans or not

joshuaellis: `licenseOnly` makes sense.

simotae14: ok thanks

remidej (approved):
Tested ✅

packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
```diff
@@ -5,19 +5,23 @@ import {
...
-const CustomIcon = styled(Lock)`
+const CustomIcon = styled(Lightning)`
   right: 15px;
   position: absolute;
   bottom: 50%;
   transform: translateY(50%);
+
+  path {
+    fill: ${({ theme }) => theme.colors.warning500};
+  }
```
joshuaellis, line 24:
```suggestion
fill: ${({ theme }) => theme.colors.warning500};
```

nitpick, but with new SVGs you can just apply a normal fill, no need to target the path.

joshuaellis (commented):
I know it's super minor, but please fill out the BC database for @pwizla just so it's recorded, no one is probably using it, but we should hold ourselves to a high standard 👍🏼

packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
simotae14: without the path I lose the specificity that I need
Schermata 2024-06-11 alle 16 20 50
If it is ok for you I am gonna stay with my style

PR #20492, opened 2024-06-11
NPS: Update frequency of the NPS

simotae14:
### What does it do?

It changes the delays to display the NPS for the first time (from 5 minutes to 30 minutes) and after the first dismissal (from 7 days to 14 days)

### Why is it needed?

Many users complain about the frequency of NPS

### How to test it?

 - Register for the first time on Strapi and on the Registration page accept the Newsletter checkbox
 - then wait after 30 minutes of activity and you will be able to see the NPS survey at the bottom of the page
 - when the NPS survey is displayed, close it by closing the modal
 - then after 14 days the survey will be proposed again
### Related issue(s)/PR(s)

CS-795

### DIFF
--- a/packages/core/admin/admin/src/components/NpsSurvey.tsx
+++ b/packages/core/admin/admin/src/components/NpsSurvey.tsx
@@ -59,9 +59,9 @@ const FieldWrapper = styled(Field)`
 
 const delays = {
   postResponse: 90 * 24 * 60 * 60 * 1000, // 90 days in ms
-  postFirstDismissal: 7 * 24 * 60 * 60 * 1000, // 7 days in ms
+  postFirstDismissal: 14 * 24 * 60 * 60 * 1000, // 14 days in ms
   postSubsequentDismissal: 90 * 24 * 60 * 60 * 1000, // 90 days in ms
-  display: 5 * 60 * 1000, // 5 minutes in ms
+  display: 10 * 60 * 60 * 1000, // 10 hours in ms
 };
 
 const ratingArray = [...Array(11).keys()];
--- a/packages/core/admin/admin/src/components/NpsSurvey.tsx
+++ b/packages/core/admin/admin/src/components/NpsSurvey.tsx
@@ -61,7 +61,7 @@ const delays = {
   postResponse: 90 * 24 * 60 * 60 * 1000, // 90 days in ms
   postFirstDismissal: 14 * 24 * 60 * 60 * 1000, // 14 days in ms
   postSubsequentDismissal: 90 * 24 * 60 * 60 * 1000, // 90 days in ms
-  display: 10 * 60 * 60 * 1000, // 10 hours in ms
+  display: 30 * 60 * 1000, // 30 minutes in ms
 };
 
 const ratingArray = [...Array(11).keys()];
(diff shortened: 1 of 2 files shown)


joshuaellis:
10 hours? How do you expect this to show? Does anyone actually leave an instance of Strapi open for 10 hours......?

simotae14:
I know it is a lot, we received a lot of complaints from the users for the frequency of the nps survey, the first request of the Product was to show the survey after 3 days of activity. This delay is not stored in the local storage, instead is counted from the first rendering of the component, product is aware of it, I'll let @yanniskadiri explain his reasons

remidej:
This should be regardless of people closing their tabs or reloading right? Since we're using [usePersistentState](https://github.com/strapi/strapi/blob/de5f8d28b56ef8e90099f64d5f04e330910617b6/packages/core/admin/admin/src/components/NpsSurvey.tsx#L392)

joshuaellis:
the time isn't cumulative from what i can see on the code?  we set a timeout, now 10 hours and when the component unmounts we don't add "how long its been open for" we just say, bye bye?

https://github.com/strapi/strapi/pull/20492/files#diff-f53ed4891968fac5b0e452d38c0fbbeb478121d6b46395c5a2e467f49654dcadL157-L165

simotae14:
yeah we are not using usePersistantState, so it is like @joshuaellis says...maybe we can keep track of the time passed since the first registration like we do for the other delays

joshuaellis:
> yeah we are not using usePersistantState for this timeout, so it is like @joshuaellis says...maybe we can keep track of the time passed since the first registration like we do for the other delays
> 
> I mean store the value in the local storage

yeah that would make more sense, if we've not shown the NPS keep track of "spent delay" & then when someone dismisses, you reset that spent time so next time to shows we count again.

remidej:
I don't think you're wrong @joshuaellis 💀 

So the problem is the use of a long setTimeout, right?

Maybe instead we can do some sort of polling solution with setInterval? Something that runs every minute or so to check if the modal should be displayed based on the data in localstorage

simotae14:
yeah, it can be a solution. We need just take into consideration another aspect, the 5 minutes timeout now it is used also in conjuction with the other timeouts, I mean, for example, if 90 days are passed since the first response then before showing the nps survey we wait 5 minutes of activity (that means if we change it to 10 hours we need to wait 10 hours of activity)... I think that we can avoid this behaviour and just show immediately the NPS survey when 90 days are passed (this value is calculated at the first rendering of the NPS component)

joshuaellis:
> yes, it can be a solution. Just take another aspect into consideration, the 5 minute timeout is now also used in conjunction with the other timeouts, I mean for example if 90 days have passed since the first response then before showing the nps survey we wait 5 minutes of activity (this it means if we change it to 10 hours we have to wait for 10 hours of activity)... I think we can avoid this behavior and immediately show the NPS survey when 90 days have passed (this value is calculated at the first rendering of the NPS component)

This should be discussed with Yannis, IMO do it in a public channel so everyone has some visibility & we're all on the same page 😄

simotae14:
As a quick win we decided to try with 30 minutes just to see if we can stop the complaints, otherwise we can discuss and plan a better approach for the issue, thanks for the contribution team

joshuaellis:
good idea, 30 mins is probably reasonable. Someone should be monitoring the repsonses though, if we're not getting any then we know it's too long

remidej (approved):

PR #20493, opened 2024-06-11
Fix: Cloud left image not following the sidenav

Mcastres:
### What does it do?

Fix the left Cloud image CSS when you expand/shrink the side nav.
Due to the previous `left: ...;` CSS rule, the cloud image was not following the side nav "movement".

Now it does with `float: left;` :)


joshuaellis (approved):

alexandrebodin:
@Mcastres is this fixed in v5?

Mcastres:
Yes it has been fixed

PR #20494, opened 2024-06-11
feat: add client router API

joshuaellis:
### What does it do?

* adds first iteration of the client router API (marked as alpha)

### Why is it needed?

* enables users to add route objects to the react-router i.e. to use `loaders` and `actions`.
* enables users to completely modify routes e.g replace the homepage

### Related issue(s)/PR(s)

* resolves DX-1177

### DIFF
--- a/packages/core/admin/admin/src/features/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/features/StrapiApp.tsx
@@ -1,5 +1,6 @@
 import { createContext } from '../components/Context';
 import { RBAC } from '../core/apis/rbac';
+import { Router } from '../core/apis/router';
 
 import type { StrapiApp } from '../StrapiApp';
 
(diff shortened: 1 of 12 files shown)


packages/core/admin/admin/src/features/StrapiApp.tsx
```diff
@@ -8,16 +9,15 @@ import type { StrapiApp } from '../StrapiApp';
...
-  > {
+      StrapiApp,
+      | 'customFields'
+      | 'getPlugin'
+      | 'getAdminInjectedComponents'
+      | 'plugins'
+      | 'runHookParallel'
+      | 'runHookSeries'
+    >,
+    Pick<Router, 'menu' | 'settings'> {
```
innerdvations, line 20: Why do we want to leave menu and settings at the root level here if they're actually in router and addressed that way everywhere?

packages/core/content-manager/admin/src/layout.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import { Page, useGuidedTour, Layouts } from '@strapi/admin/strapi-admin';
 import { useIntl } from 'react-intl';
-import { Navigate, Route, Routes, useLocation, useMatch } from 'react-router-dom';
+import { Navigate, Outlet, Route, Routes, useLocation, useMatch } from 'react-router-dom';
```
innerdvations, line 6:
```suggestion
import { Navigate, Outlet, useLocation, useMatch } from 'react-router-dom';
```
You can also remove `routes` from the line down below too

packages/core/admin/admin/src/features/StrapiApp.tsx
joshuaellis: to avoid more BC. I'd rather deprecate, which i can do now actually – than add more BCs just for the fun of it lol wdyt?

jhoward1994 (approved):
I've QA'd navigation in the app and everything is working

PR #20496, opened 2024-06-11
Fix/nested-form-errors

jhoward1994:
### What does it do?

Change error extraction logic to correctly surface validation errors on nested fields.
Such as repeatable components on unique field validation

### Why is it needed?

Found in QA on the target branch
So the user can see the errors in their CM form

### How to test it?

Modified existing uniqueness E2E test to ensure correct errors are visible

### Related issue(s)/PR(s)

DX-1348


packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -669,12 +669,48 @@ const useField = <TValue = any,>(path: string): FieldValue<TValue | undefined> =
 
   const handleChange = useForm('useField', (state) => state.onChange);
 
-  const error = useForm('useField', (state) => getIn(state.errors, path));
+  const convertErrorFormat = (stateErrors: FormErrors<FormValues>) => {
```
joshuaellis, line 672: `convertErrorFormat` to what? 😄

joshuaellis: on the flip side, this function code is really clear so well done 👏🏼

joshuaellis (commented):
Do you think it's possible to add a few tests for the error message stuff? 😄 be nice for no one to break it and have our assumptions verified

packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -778,7 +820,7 @@ const Blocker = ({ onProceed = () => {}, onCancel = () => {} }: BlockerProps) =>
   return null;
 };
 
-export { Form, Blocker, useField, useForm, getYupValidationErrors };
+export { Form, Blocker, useField, useForm, getYupValidationErrors, FormProvider };
```
joshuaellis, line 823: How confident are you that you're not exporting this provider from the admin package when you do this?

jhoward1994: Oops I see https://github.com/strapi/strapi/blob/0c5ac9b4a0965db69613ed48a167c99c6127d9e0/packages/core/admin/admin/src/index.ts#L15-L16

🤔 wdyt the best idea is? should we do named form exports from the package index?

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -69,7 +69,14 @@ const formatErrorMessages = (errors: FormErrors, parentKey: string, formatMessag
           )
         );
       } else {
-        messages.push(...formatErrorMessages(value, currentKey, formatMessage));
+        messages.push(
+          ...formatErrorMessages(
+            // @ts-expect-error TODO: check why value is not compatible with FormErrors
```
jhoward1994, line 74: @joshuaellis 

Providing 'initialerrors' to the Form seems like the best way to go. The only problem I'm having is with the FormError types here.

I've modified them to accept TranslationMessages but I can't work out why that has broken TS here in the CM. The type for values seems like it should be compatible.

(parameter) value: FormErrors<any> | TranslationMessage

any ideas?

joshuaellis: You could mark it as `todo`? it's probably just a bit of a bad recursive function 🤔

joshuaellis (approved):

PR #20502, opened 2024-06-12
feat(cli): add browser logout step

abdallahmz:
### What does it do?

We perform a proper OIDC logout by opening the logout endpoint in the browser. We provide the client ID and token. We do nothing if an error happens during the logout process: the logout happens at least on the app level, this just adds a logout on the Auth0 SSO level.

"chore" changes (unrelated to the initial task):
- Handle errors while fetching the config

### Why is it needed?

This change is needed to address the issue where users are unable to switch between different login providers. By logging out the user from the OIDC provider, the user's session is cleared. This allows the user to choose a different login provider when they attempt to log in again.

### How to test it?

First, use the login command:
Log in in the CLI using one of the login providers (e.g., Google), 

Then, log out (using the logout command):
The CLI should open the OIDC provider's logout URL in a browser. 

Try to log in again. You should be able to choose a different login provider.

7a95c833-d2d8-4060-b5b4-35d24b2f911d

### DIFF
--- a/packages/cli/cloud/src/logout/action.ts
+++ b/packages/cli/cloud/src/logout/action.ts
@@ -1,6 +1,8 @@
 import type { CLIContext } from '../types';
 import { tokenServiceFactory, cloudApiFactory } from '../services';
 
+const openModule = import('open');
+
 export default async (ctx: CLIContext) => {
   const { logger } = ctx;
   const { retrieveToken, eraseToken } = await tokenServiceFactory(ctx);
@@ -10,10 +12,27 @@ export default async (ctx: CLIContext) => {
     logger.log("You're already logged out.");
     return;
   }
-  const cloudApiService = await cloudApiFactory(token);
+  const cloudApiService = await cloudApiFactory(ctx, token);
+  const config = await cloudApiService.config();
+  const cliConfig = config.data;
+
   try {
-    // we might want also to perform extra actions like logging out from the auth0 tenant
     await eraseToken();
+
+    openModule.then((open) => {
+      open
+        .default(
+          `${cliConfig.baseUrl}oidc/logout?client_id=${encodeURIComponent(
+            cliConfig.clientId
+          )}&logout_hint=${encodeURIComponent(token)}
+          `
+        )
+        .catch((e: Error) => {
+          // Failing to open the logout URL is not a critical error, so we just log it
+          logger.debug(e.message, e);
+        });
+    });
+
     logger.log(
       '🔌 You have been logged out from the CLI. If you are on a shared computer, please make sure to log out from the Strapi Cloud Dashboard as well.'
     );
--- a/packages/cli/cloud/src/logout/action.ts
+++ b/packages/cli/cloud/src/logout/action.ts
@@ -22,7 +22,7 @@ export default async (ctx: CLIContext) => {
     openModule.then((open) => {
       open
         .default(
-          `${cliConfig.baseUrl}oidc/logout?client_id=${encodeURIComponent(
+          `${cliConfig.baseUrl}/oidc/logout?client_id=${encodeURIComponent(
             cliConfig.clientId
           )}&logout_hint=${encodeURIComponent(token)}
           `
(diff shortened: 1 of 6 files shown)


packages/cli/cloud/src/logout/action.ts
```diff
@@ -11,9 +13,26 @@ export default async (ctx: CLIContext) => {
...
   try {
-    // we might want also to perform extra actions like logging out from the auth0 tenant
     await eraseToken();
+
+    openModule.then((open) => {
+      open
+        .default(
+          `${cliConfig.baseUrl}oidc/logout?client_id=${encodeURIComponent(
+            cliConfig.clientId
+          )}&logout_hint=${encodeURIComponent(token)}
```
nathan-pichon, line 27: Do you have a preview of the logout page?

abdallahmz: 7a95c833-d2d8-4060-b5b4-35d24b2f911d
Yes, there you go! Will add it to the description

packages/cli/cloud/src/logout/action.ts
```diff
@@ -11,9 +13,26 @@ export default async (ctx: CLIContext) => {
     return;
   }
   const cloudApiService = await cloudApiFactory(token);
+  const config = await cloudApiService.config();
```
nathan-pichon, line 16: I think we missed it in the original dev of this action, but could you add a try catch around all awaits to not end up with an uncatched failed promise? 🤔

abdallahmz: We could have the error within cloudApiService's `config` function? we could even extract .data to simplify the returned value.

Current code:

```
    config(): Promise<AxiosResponse<CloudCliConfig>> {
      return axiosCloudAPI.get('/config');
    },
```

Something like the following:
```
async config(): Promise<CloudCliConfig> {
  try {
    const response = await axiosCloudAPI.get('/config');
    const cloudCliConfig = response.data;

    if (!cloudCliConfig) {
      throw new Error('Undefined cloudCliConfig');
    }

    return cloudCliConfig;
  } catch (error) {
    throw new Error(`Couldn't retrieve the cloud CLI config - ${error.message}`);
  }
},
```
I'll go with that, I think it'll be more convenient and it's still acceptable logic for an API service. I don't think we need to whole Axios response anywhere.

nathan-pichon: I think it's better to stick with the full Axios response, as it is used in other API functions calls (e.g. project creation), and staying coherent.
It's ok to still handle the error in the action, I believe.

nathan-pichon: Also we want to hide technical errors to our users and only display business error messages + debug log the technical one.

abdallahmz: Modifications done!

I added the try catch within config() and kept the AxiosResponse signature.
(Passed ctx through the factory. The whole axios response will be displayed in debug mode if an error occurs)

packages/cli/cloud/src/services/cli-api.ts
```diff
@@ -111,8 +114,22 @@ export async function cloudApiFactory(token?: string): Promise<CloudApiService>
...
+          throw new Error('Error fetching cloud CLI config from the server.');
+        }
+
+        return response;
+      } catch (error) {
+        logger.error(
+          "🥲 Oops! Couldn't retrieve the cloud CLI config from the server. Please try again."
+        );
+
+        throw error;
```
nathan-pichon, line 131: wouldn't this log two errors in the user's terminal if the action is try..catching this ?

abdallahmz: Yes, it could output something like:

- "🥲 Oops! Couldn't retrieve the cloud CLI config from the server. Please try again." 
- "Login action failed."

nathan-pichon: Could we only have one log for UX ?

abdallahmz: Sure, if it's too much details for the `error` log level (or if we have a "one error line max per command" rule), I'll output it to `debug` instead

nathan-pichon (approved):
lgtm

PR #20503, opened 2024-06-12
fix(admin): date picker jumps around due to timezones

### LINKED ISSUE #20321: Typeof - date (ex: 01/01/2024) field decrements from the selected date

Node.js version: v20.6.1
NPM version: 9.8.1
Strapi version: 5.0.0-beta.3
Database: sqlite
Operating system: Ubuntu 22.04.4 LTS
Is your project Javascript or Typescript: TypeScript

### Describe the bug

A user selects a date through the date picker which is a field in Strapi.
When the user selects a date and proceeds to the next field or just clicks outside of the calendar field the date decrements by one.
For example: If a user were to select 2nd March 2019 on the date picker and proceeds to click outside the calendar or another field the date then decrements by one count and displays 1st March 2019.

1. Create a Collection Type with any name
2. Add a "Date" field to the Collection Type.
3. Navigate to Content Manager and select a date through the newly created "Date" field for the Collection Type.
4. User selects 5th Feb. 2020 
5. User clicks or tabs to the next field
6. The date is now decremented by one and now displays 4th Feb. 2020.

### Expected behavior

When a user chooses a date and clicks on it then that date should remain and not be decremented or incremented.

joshuaellis:
### What does it do?

* handles timezones with the date picker

### Why is it needed?

* the date picker is timezone aware (should we want to start working with them) but the date it converts it too doesn't handle it well, making it think it should be X time prior

### How to test it?

* add a date
* blur the input
* date should stay what you set

### Related issue(s)/PR(s)

* resolves #20321
* resolves DX-1462

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Date.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Date.tsx
@@ -1,5 +1,6 @@
 import { forwardRef, memo } from 'react';
 
+import { getLocalTimeZone, parseAbsolute, toCalendarDate } from '@internationalized/date';
 import { DatePicker, useComposedRefs, Field } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
 
@@ -27,7 +28,7 @@ const DateInput = forwardRef<HTMLInputElement, InputProps>(
             field.onChange(name, date);
           }}
           onClear={() => field.onChange(name, undefined)}
-          value={value}
+          value={value ? convertLocalDateToUTCDate(value) : undefined}
           {...props}
         />
         <Field.Hint />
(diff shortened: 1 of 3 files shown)


packages/core/admin/admin/src/components/FormInputs/Date.tsx
```diff
@@ -27,7 +28,7 @@ const DateInput = forwardRef<HTMLInputElement, InputProps>(
             field.onChange(name, date);
           }}
           onClear={() => field.onChange(name, undefined)}
-          value={value}
+          value={value ? convertLocalDateToUTCDate(value) : undefined}
```
Bassel17, line 31: A small suggestion to use `??` operator
```suggestion
          value={value ?? convertLocalDateToUTCDate(value)}
```

Bassel17 (approved):
Overall looks good!

packages/core/admin/admin/src/components/FormInputs/Date.tsx
joshuaellis: thats not what this code is doing, its checking theres a value before we pass it to the `convertLocalDatetoUTCDate` func, so i'll probably keep it the same :)

PR #20507, opened 2024-06-13
test(e2e): fix releases-page and listview

innerdvations:
### What does it do?

- use the right locator to click the date on releases page
- update various text labels that have changed
- make `findAndClose` close all instead of throwing when more than one item is found

### Why is it needed?

test was broken

### How to test it?

test should pass

except the following that are still broken:

- editview.spec.ts:255 - as a user I want to unpublish a document
- blocks.spec.ts:32 - adds a code block and specifies the language : appears to be an actual bug, on webkit there is no language selector
- releases.spect.ts - a few failing in here


jhoward1994:
https://github.com/strapi/strapi/blob/3e99153e0ba79213de4b2e27d70329c156ecb5e7/tests/e2e/tests/content-manager/editview.spec.ts#L87-L88

I think this is failing because the add to release button enabled in this menu

innerdvations:
> https://github.com/strapi/strapi/blob/3e99153e0ba79213de4b2e27d70329c156ecb5e7/tests/e2e/tests/content-manager/editview.spec.ts#L87-L88
> 
> I think this is failing because the add to release button enabled in this menu

So you mean that line can be removed from the test?

Feranchz:
> > https://github.com/strapi/strapi/blob/3e99153e0ba79213de4b2e27d70329c156ecb5e7/tests/e2e/tests/content-manager/editview.spec.ts#L87-L88
> > 
> > I think this is failing because the add to release button enabled in this menu
> 
> So you mean that line can be removed from the test?

@innerdvations I think you can remove it. This is my bad that I didn't catch this on my PR.

You can add published entries to a release for unpublis, so the more options button is not disabled, so I guess this line doesn't makes sense anymore.

jhoward1994 (approved):

PR #20526, opened 2024-06-14
fix(content-releases): release validation and dropdown menu endIcon

Feranchz:
### What does it do?

Fix 2 small issues on Content Releases:
1. When adding an entry to a release we need to wait for the entry status validation
2. Remove endIcon from edit release button to match designs

### DIFF
--- a/packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
+++ b/packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
@@ -29,8 +29,9 @@ import {
   Tooltip,
   EmptyStateLayout,
   LinkButton,
-  Menu,
   Dialog,
+  SimpleMenu,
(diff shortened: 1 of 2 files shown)


packages/core/content-releases/admin/src/pages/ReleaseDetailsPage.tsx
```diff
@@ -422,6 +422,7 @@ const ReleaseDetailsLayout = ({
                     defaultMessage: 'Release edit and delete menu',
                   })}
                   variant="tertiary"
+                  endIcon={null}
```
remidej, line 425: Should the comment above the updated?

Feranchz: Thanks! take the opportunity to change this to the SimpleMenu component that has what we needed in v2

PR #20532, opened 2024-06-17
feat: Add Release Settings page

simotae14:
### What does it do?

Add the Release Settings page to select the timezone to use
Schermata 2024-06-17 alle 10 22 06

### Why is it needed?

We want to give the ability to the user to select the default timezone to use in Releases

Page should consume these two endpoints:

Get settings
GET `/releases/settings`

Update settings
PUT `/releases/settings`

body:
```
{
  defaultTimezone: string | null
}
```

### How to test it?

Open the Strapi CMS with a EE license
- Go to the "Settings" and open the Release link under the "Global Settings"
- Select a Timezone from the dropdown and save it
  - then after the Successful message you can check if the value is saved just refreshing the page

### Related issue(s)/PR(s)

CS-811


simotae14:
Closed in favour of https://github.com/strapi/strapi/pull/20589

PR #20535, opened 2024-06-17
fix(content-manager): add isFetching to loading check

Feranchz:
### What does it do?

Add a new variable to the isLoading check on the add column to table hook.

### Why is it needed?

When rendering the additional columns added by plugins, we have two values that are separated: the model and collectionType information (found in the URL and accessible using the useDoc hook) and the list view data.

When a user moves from one ContentType list view to another, there is a moment when the model changes to the new one, but the data still reflects the old one because we are in the process of fetching the new data. To avoid this, we use an isLoading check. However, this is insufficient because isLoading is true only the first time a query is performed. Instead, we need to check both isLoading and isFetching to ensure that whenever the user is fetching different data, nothing is displayed.

This approach helps us avoid the "Not found" error in the list view.

### How to test it?

1. Go to the list view and move between different content types.
2. Should not have any error.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -97,7 +97,7 @@ const ListViewPage = () => {
   });
 
   const params = React.useMemo(() => buildValidParams(query), [query]);
-  const { data, error, isLoading } = useGetAllDocumentsQuery({
+  const { data, error, isLoading, isFetching } = useGetAllDocumentsQuery({
     model,
     params,
   });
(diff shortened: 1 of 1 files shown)


simotae14 (approved):
it looks good, 👏🏽

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -170,7 +170,7 @@ const ListViewPage = () => {
     return formattedHeaders;
   }, [displayedHeaders, formatMessage, list, runHookWaterfall, schema?.options?.draftAndPublish]);
 
-  if (isLoading) {
+  if (isLoading || isFetching) {
```
remidej, line 173: I think the isLoading check is now redundant

remidej (approved):

Issue #20504, opened 2024-06-13
NPX / CLI Broken in beta 10


Eventyret:
### Required System information

- Node.js version: 20.13.1
- NPM version: 10.0.1
- Strapi version: v5.0.0-beta10
- Database: sqlite
- Operating system: Windows / Mac
- Is your project Javascript or Typescript: Javascript / Typescript

### Describe the bug

INSERT HERE

### Steps to reproduce the behavior

1. run `npx create-strapi-app@v5.0.0-beta.10 myproject`
2. Let the CLI finish
4. Try run `npm run develop`
Sometimes here you can see Error 1 (Please see code snippets below) 
5. If Error 1 (Please see code snippet below) run `npm install` and it goes away.
6. Run `npm run develop`
The outcome now would be Error 2 (Please see code snippet and screenshot below)

This seems to be because of the changes in `.env`

```.env
# Server
HOST=0.0.0.0
PORT=1337

# Secrets
APP_KEYS=XYZ
API_TOKEN_SALT=XYZ
ADMIN_JWT_SECRET=XYZ
TRANSFER_TOKEN_SALT=XYZ
# Database
DATABASE_CLIENT=sqlite
DATABASE_FILENAME=
```
Note that the DATABASE_FILENAME is blank and does not point to `.tmp/data.db`
Also, there is no `JWT_SECRET` 

All of this was in `Beta-9`

### Expected behavior

When using the CLI the DATABASE_FILENAME= should be filled out to allow strapi to be installed.

### Screenshots

![image](https://github.com/strapi/strapi/assets/977232/c436beb4-2f50-41d4-8573-022a2b0875ce)

### Code snippets

If applicable, add code samples to help explain your problem.

Error 1)
```bash
 npm run develop

> beta-10@0.1.0 develop
> strapi develop

node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module 'ajv/dist/core'
Require stack:
- /Users/eventyret/Playground/beta10/node_modules/ajv-draft-04/dist/index.js
- /Users/eventyret/Playground/beta10/node_modules/@rushstack/node-core-library/lib/JsonSchema.js
- /Users/eventyret/Playground/beta10/node_modules/@rushstack/node-core-library/lib/index.js
- /Users/eventyret/Playground/beta10/node_modules/@rushstack/terminal/lib/NormalizeNewlinesTextRewriter.js
[...]

PR #20536, opened 2024-06-17
chore: upgrade umzug & improve db migration logging

alexandrebodin:
### What does it do?

- Add ajv as a direct dep to fix https://github.com/strapi/strapi/issues/20504 for now.
- Improve migration logging so internal migration only appear in debug mode & improve logging

### Why is it needed?

v5 RC

### How to test it?

- delete some rows in strapi_migrations & restart the getstarted app, logs should appear in debug for internal migration
- try creating a user migration in database/migrations (a simple sql file with SELECT 1) and restart and see the info log

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -243,8 +243,16 @@ class Strapi extends Container implements Core.Strapi {
 
   // TODO: split into more providers
   registerInternalServices() {
+    const config = createConfigProvider(this.internal_config, this);
+
+    const logger = createLogger({
+      level: 'http', // Strapi defaults to level 'http'
+      ...config.get('logger'), // DEPRECATED
+      ...config.get('server.logger.config'),
+    });
+
     // Instantiate the Strapi container
-    this.add('config', () => createConfigProvider(this.internal_config, this))
+    this.add('config', () => config)
       .add('query-params', createQueryParamService(this))
       .add('content-api', createContentAPI(this))
       .add('auth', createAuth())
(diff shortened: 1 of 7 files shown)


alexandrebodin:
Trying an experimental to see if npm is happy with this version

packages/core/core/src/Strapi.ts
```diff
@@ -243,22 +243,24 @@ class Strapi extends Container implements Core.Strapi {
 
   // TODO: split into more providers
   registerInternalServices() {
+    const config = createConfigProvider(this.internal_config, this);
+
+    const logger = createLogger({
+      level: 'http', // Strapi defaults to level 'http'
+      ...config.get('logger'), // DEPRECATED
```
Convly, line 250: Do we want to soft deprecate it and remove it in v6, or do we want to create a codemod for that and remove this line?

Convly: (Wondering if it's even possible to create a codemod for it :thinking: )

packages/core/database/src/migrations/users.ts
```diff
@@ -46,7 +48,21 @@ export const createUserMigrationProvider = (db: Database): UserMigrationProvider
 
   const umzugProvider = new Umzug({
     storage: createStorage({ db, tableName: 'strapi_migrations' }),
-    logger: console,
+    logger: {
+      info(message) {
```
Convly, line 52: It's kind of annoying and error-prone to have to redefine the same logger again and again for each migration. Is there a way to factorize/DRY that a bit, or is it on purpose?

packages/core/core/src/Strapi.ts
alexandrebodin: I just moved the code I didn't think about it at all :D

packages/core/database/src/migrations/users.ts
alexandrebodin: it's the kind of stuff you don't want to factorize too early. the to info funcs are different and it's enough to duplicate the rest instead of making a factory with options honestly

Convly (approved):

PR #20537, opened 2024-06-17
enhancement: set app.proxy automatically if proxy is set

innerdvations:
### What does it do?

Set koa app.proxy automatically if server.proxy.global, server.proxy.http, or server.proxy.https is set

TODO:

- [ ] add a test
- [ ] update documentation

### Why is it needed?

If a proxy is used, this value must be true or else koa won't work. So it saves the user the extra config of adding `proxy.koa:true` to their server config

### How to test it?

- Set server.proxy.global (or http or https) and koa should automatically set app.proxy to true and work
- server.koa should still be able to override that value so a user can turn it on without explicitly setting a proxy in strapi (for example, if they have their own custom method)
- technically setting server.koa:false should also override and keep it as false, but I don't really see a use-case for it; we may just want to make it true and output a warning that the option is in conflict

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/20416


packages/core/core/src/services/server/index.ts
```diff
@@ -15,9 +16,25 @@ const healthCheck: Core.MiddlewareHandler = async (ctx) => {
   ctx.status = 204;
 };
 
+const shouldTrustProxy = (strapi: Core.Strapi) => {
+  const explicit = strapi.config.get('server.proxy.koa', undefined);
+
+  if (isBoolean(explicit)) {
```
innerdvations, line 22: We might want to just return when explicit===true, but when `explicit === false && implict` we output a warning that there is a conflict with the settings and that we're ignoring it because it would break the koa server.

derrickmehaffy:
If we are going to do this it might be worth checking if the ./config/server.js contains a `url` key that has an `https://` protocol too because in a case like that you also need the koa proxy set to true.

Issue #20083, opened 2024-04-10
Cannot select Multiple files in a  Multiple files Media field in Strapi V5


creative-ae:
## Bug report
Cannot select Multiple files in a  Multiple files Media field in Strapi V5

### Required System information
- Node.js version:
- NPM version:
- Strapi version: V5 BETA 3
- Database:
- Operating system:
- Is your project Javascript or Typescript:

### Describe the bug

Cannot select Multiple files in a  Multiple files Media field in Strapi V5. The field is only limited to select 1 files only

### Steps to reproduce the behavior

1. login to admin
2. create a new collection type
3. add Media field
4. select multiple media
5. save
6. go to add a content
7. add content with image
8. i can select only 1 file

### Expected behavior
i should be able to select multiple files

creative-ae:
![Screenshot 2024-04-10 at 22 50 02](https://github.com/strapi/strapi/assets/46525762/33a036ca-7d2d-4595-b8a6-2c13f2743b94)
![Screenshot 2024-04-10 at 22 47 24](https://github.com/strapi/strapi/assets/46525762/1a140bdb-4748-41f7-9914-9b17133a84c1)

Eventyret:
Confirmed `5.0.0-beta.8`

alexandrebodin:
Merged the fix

PR #20541, opened 2024-06-17
fix: multiple is only available in attribute now

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/20083


simotae14 (approved):
it looks good and just checked, it works 👏🏽

PR #20125, opened 2024-04-16
fix(database): Add groupBy after distinct queries (option 2)

innerdvations:
Alternative solution: https://github.com/strapi/strapi/pull/20124

### What does it do?

After performing a query that is supposed to be `distinct` but should actually probably be a `groupBy`, we apply the same groupBy on the resultset received, in Node rather than the db query.

NOTE: on further review and discussion, this will probably break pagination with a resultset that exceeds the size of one page.

### Why is it needed?

Fixes https://github.com/strapi/strapi/issues/11892

### How to test it?

See [the original issue](https://github.com/strapi/strapi/issues/11892).

A test of `sort` has been added to prevent regressions, but this may have unintended side effects that are not caught by existing tests, so a pretty thorough test of nested populates and queries with joins and distincts that aren't the typical path we use.

### Related issue(s)/PR(s)

Alternative solution: https://github.com/strapi/strapi/pull/20124

Fixes https://github.com/strapi/strapi/issues/11892

DX-117

PR #20124, opened 2024-04-16
fix(database): Add groupBy to queries with joins (option 1)

innerdvations:
Alternative solution: https://github.com/strapi/strapi/pull/20125

### What does it do?

Removes duplicate results from sorts on relation fields by applying a groupBy on all select statements that don't have one

TODO:

- [ ] fix that it breaks components queries

### Why is it needed?

Fixes https://github.com/strapi/strapi/issues/11892

### How to test it?

See [the original issue](https://github.com/strapi/strapi/issues/11892).

A test of `sort` has been added to prevent regressions, but this may have unintended side effects that are not caught by existing tests, so a pretty thorough test of the content API, admin API, pretty much anything that makes queries.

### Related issue(s)/PR(s)

Alternative solution: https://github.com/strapi/strapi/pull/20125

Fixes https://github.com/strapi/strapi/issues/11892

DX-117

PR #13387, opened 2022-05-24
Fix duplicate sort results

Bassel17:
### What does it do?

In query builder when we query for data and we have joins, we run `qb.distinct` which should remove duplicate data from the query but after some testing it doesn't seem to affect queries with multiple joins and duplicate data was queried when trying nested sorting. So I opted to use `qb.groupBy` instead when we have multiple joins and sorting

### Why is it needed?

When doing nested sorting, duplicate data is being returned

### How to test it?

run a strapi application and create content types with a relation, try to sort the parent content type based on a relation field there should be no duplicates, you can follow the same method in the issue I will reference below.

### Related issue(s)/PR(s)

fixes #11892

PR #20543, opened 2024-06-18
Fix Duplicate Rows on Deep Sorting

### LINKED ISSUE #11892: [v4] sort by graphql duplicating rows
### Describe the bug

Graphql sort on nested relations is basically duplicating rows as a join would do.
This was working in v3.

Create your graphql query on a resource that has a relation. This is a one to many in this case, this should work for any relation i think
image

Call this in your sort parameters:
image

### Expected behavior

Should only have one row per item but this is duplicated:
image

As you can see there i have multiples rows with the same id depending on the number of nested articles linked to the category.

### System

- Node.js version: 14.18.1
- NPM version: 6.14.15
- Strapi version: 4.0.0
- Database: MySQL
- Operating system: Windows 11

Convly:
**TL;DR: Add support for deep sorting using Strapi queries.**

**Initial report:**  [https://github.com/strapi/strapi/issues/11892](https://github.com/strapi/strapi/issues/11892)

### What does it do?
- Add the support for deep sorting in the Strapi query builder
    - [If an attempt to deep sort is detected](https://github.com/strapi/strapi/blob/fix/duplicates-on-deep-sorts/packages/core/database/src/query/query-builder.ts#L449-L475), then [wrap the current query with one allowing for deep sorting capabilities](https://github.com/strapi/strapi/blob/fix/duplicates-on-deep-sorts/packages/core/database/src/query/query-builder.ts#L626).
    - Detecting whether to use deep sort is done by checking the `orderBy` clauses.
        - Only look at those having at least one `.` character.
        - If, when divided into groups, one of them (*expect for the last*) matches one of the join table aliases or a relational attribute in the current metadata, then it means we’re facing a deep sort.
    - How the wrapping works (code available [here](https://github.com/strapi/strapi/blob/fix/duplicates-on-deep-sorts/packages/core/database/src/query/helpers/order-by.ts#L86-L216)):
        1. Initiate a new Knex query instance (referenced as `Q`) using the DB connection.
        2. Clone the original query to create the sub-query (referenced as `R`) and avoid any mutation on the initial object
            1. Remove pagination statements
            2. Remove select statement
        3. Create a sub-query callback from `R` to extract and sort partitions using row number, known as `T` 
            1. Partition by `R.id`
            2. Order by the original clauses
        4. Create the final `Q` query, that selects and sorts the wanted data using `T`
            1. Re-apply the pagination
            2. Re-apply a clean select
            3. Re-apply the order by using data returned by `T`
        
        Here’s the difference in the generated queries and returned rows (before/after)
        
        - Before
        
        ```sql
        SELECT DISTINCT "t2"."name",
                        "t0".*,
                        "t0"."id"
        FROM "public"."articles" AS "t0"
        LEFT JOIN "public"."articles_categories_lnk" AS "t1" ON "t0"."id" = "t1"."article_id"
        LEFT JOIN "public"."categories" AS "t2" ON "t1"."category_id" = "t2"."id"
        ORDER BY "t2"."name" ASC
        LIMIT ?
        ```
        
        | name | id | document_id | title |
        | --- | --- | --- | --- |
        | Category A | 3 | czdpldaa2eynfjd95pbh2rf5 | Article D |
        | Category B | 1 | ls4i55fi1zdkiwiwq427qutw | Article A |
        | Category C | 4 | eb3wgthmh8kjwua5t2rqgahz | Article B |
        | Category D | 2 | efppopmjiovgck66adwpvntv | Article C |
        | Category D | 3 | czdpldaa2eynfjd95pbh2rf5 | Article D |
        | Category D | 4 | eb3wgthmh8kjwua5t2rqgahz | Article B |
        - After
        
        ```sql
        SELECT "t3".*,
               "t3"."id"
        FROM "public"."articles" AS "t3"
        INNER JOIN
          (SELECT "t4"."id",
                  "t4"."__strapi_order_by__t2_name",
                  row_number() OVER (PARTITION BY "t4"."id"
                                     ORDER BY "t4"."__strapi_order_by__t2_name" ASC) AS __strapi_row_number
           FROM
             (SELECT "t0"."id",
                     "t2"."name" AS "__strapi_order_by__t2_name"
              FROM "public"."articles" AS "t0"
              LEFT JOIN "public"."articles_categories_lnk" AS "t1" ON "t0"."id" = "t1"."article_id"
              LEFT JOIN "public"."categories" AS "t2" ON "t1"."category_id" = "t2"."id") AS "t4") AS "t5" ON "t5"."id" = "t3"."id"
        AND "t5"."__strapi_row_number" = 1
        ORDER BY "t5"."__strapi_order_by__t2_name" ASC,
                 "t5"."id" ASC
        LIMIT ?
        ```
        
        | id | document_id | title |
        | --- | --- | --- |
        | 3 | czdpldaa2eynfjd95pbh2rf5 | Article D |
        | 1 | ls4i55fi1zdkiwiwq427qutw | Article A |
        | 4 | eb3wgthmh8kjwua5t2rqgahz | Article B |
        | 2 | efppopmjiovgck66adwpvntv | Article C |
    - **Performances** ⚡
        - Before: 6 rows retrieved starting from 1 in 61 ms (execution: 5 ms, fetching: 56 ms)
        - After: 4 rows retrieved starting from 1 in 51 ms (execution: 4 ms, fetching: 47 ms)
        - ⇒ **No noticeable changes in performances**

*(other changes)*

- Add API tests for sorting and deep sorting:
    - **Regular Sort**: This tests the basic sorting of articles by title.
    - **Deep Sort (1st level)**: This tests sorting of articles when they are nested one level deep, specifically sorting articles based on category names.
    - **Deep Sort (2nd level)**: This is a more complex sort, checking sorting of articles based on tags' names which are two levels deep.
    - **Deep Sort (2nd level) + Regular Sort**: This test checks the correct sequence of sorting when multiple sorts (a deep sort and a regular sort) are applied.
    - **Two Deep Sorts (2nd and 1st level)**: This test checks the correct sequence of sorting when multiple sorts (two deep sorts) are applied.
    - **Deep Sort (2nd Level) + Pagination**: Tests sorting on paginated results.
    - **Deep Sort (1st Level) + Filters**: Tests sorting on filtered results.
    - **Update + Deep Sort**: Ignore sorting on update operations
    - **Delete + Deep Sort**: Ignore sorting on delete operations
- Fix flaky API tests on PostgreSQL
    - `deep-filtering.test.api.js` was incorrectly querying collectors by using `$contains` instead of `$contains` to filter by cards’ names.
    - `number-of-draft-relations.test.api.js` was running create operations concurrently, which was causing an issue with transactions when updating the relation attached to the entities (e.g. transferring ownership)
    - Fixed fixtures being JSON files with hardcoded ID, causing the sequences for PK (in PostgreSQL) to be out-of-sync and generate uniqueness violation. They’ve been moved to JS modules and use callbacks to reference other fixtures.
- Fix `createDocumentSiblingsIDSubQuery` in `regular-relation.ts` (database)
    - Use sub-queries instead of re-using the same connection (and mutating it)
    - Support attaching the same transaction to the sub-queries.

### Why is it needed?

Deep sorting causes duplicate results in the returned rows.

This is due to the conflict between needing to select the column used in the order by statement and applying the distinct to it, creating duplicate rows.

### How to test it?

Run the `sort.test.api.ts` suite or try running Strapi queries (REST or programmatically) containing deep sorts (e.g. `categories.tags.name`)

### Related issue(s)/PR(s)

fix #11892

Take over #20125, #20124, #13387

---
Shoutout to @innerdvations and @MorganClt for their help :slightly_smiling_face:


innerdvations (commented):
From a quick review this looks great! I haven't tested it yet though.

Do you think we need to flag it as a breaking change since anyone using getKnexQuery with a deep sort will now get a very different query than they would have in v4?

packages/core/database/src/entity-manager/regular-relations.ts
```diff
@@ -24,8 +24,8 @@ declare module 'knex' {
 }
 
 //  TODO: This is a short term solution, to not steal relations from the same document.
-const getDocumentSiblingIdsQuery = (con: Knex, tableName: string, id: ID) => {
-  // Find if the model is a content type or something else (e.g component)
+const createDocumentSiblingsIDSubQuery = (tableName: string, id: ID, trx?: Knex.Transaction) => {
```
alexandrebodin, line 27: @Marc-Roig js changed this code too, that will probably conflict with your change but using the trx looks nice too

packages/core/database/src/query/helpers/order-by.ts
```diff
@@ -8,7 +9,10 @@ import type { Ctx } from '../types';
 
 type OrderByCtx = Ctx & { alias?: string };
```
alexandrebodin, line 10: I thought Ctx already contained an alias, if it's the case there might be conflicts between the 2, worth using a diff name if that's the case 🤔  (

packages/core/database/src/query/helpers/order-by.ts
```diff
@@ -67,3 +71,151 @@ export const processOrderBy = (orderBy: OrderBy, ctx: OrderByCtx): OrderByValue[
...
+
+  // The orderBy is cloned to avoid unwanted mutations of the original object
+  const orderBy = _.cloneDeep<OrderByValue[]>(qb.state.orderBy);
+
+  // 0. Init a new Knex query instance (referenced as Q) using the DB connection
+  //    The connection reuse the original table name (aliased if needed)
+  const qAlias = qb.getAlias();
+  const aliasedTableName = qb.mustUseAlias() ? alias(qAlias, tableName) : tableName;
+
+  const Q = db.getConnection(aliasedTableName);
```
alexandrebodin, line 107: Let's avoid single letter variables :)

packages/core/database/src/query/query-builder.ts
```diff
@@ -408,6 +446,34 @@ const createQueryBuilder = (
       return state.joins.length > 0 && _.isEmpty(state.groupBy);
     },
 
+    shouldUseDeepSort() {
+      return (
```
alexandrebodin, line 450: we can move that into the sort helpers maybe ?

tests/api/core/strapi/api/sort/sort.test.api.ts
```diff
@@ -0,0 +1,403 @@
...
+  test('Regular sort', async () => {
+    const res = await rq.get(`/${schemas.contentTypes.article.pluralName}`, {
+      qs: { sort: 'title' },
+    });
+
+    expect(res.status).toBe(200);
+    expect(res.body.data.length).toBe(4);
+
+    expect(res.body.data).toMatchObject([
+      { id: 1, title: 'Article A', ...BASIC_ARTICLE_FIELDS },
```
alexandrebodin, line 201: to avoid unpredictability you can remove the hardcoded ids from the matchers

tests/api/core/strapi/api/sort/sort.test.api.ts
```diff
@@ -0,0 +1,403 @@
...
+    expect(res.body.data).toMatchObject([
+      { id: 3, title: 'Article D', ...BASIC_ARTICLE_FIELDS },
+      { id: 1, title: 'Article A', ...BASIC_ARTICLE_FIELDS },
+      { id: 2, title: 'Article C', ...BASIC_ARTICLE_FIELDS },
+    ]);
+  });
+
+  // This one will fail because we don't support deep sort with deep filter
+  // The where and orderBy will be applied to two different joins (instead of a single one), impacting the results
+  test.skip('Deep sort (1st level) + Deep filters (1st level)', async () => {
```
alexandrebodin, line 345: can you create a ticket for this to be supported 🙏

alexandrebodin (commented):
just a few things but LGTM. Well commented

packages/core/database/src/entity-manager/regular-relations.ts
Marc-Roig: oops! We can merge the changes

Aren't transactions automatically applied to subqueries? :O

Convly: I initially thought the same but had a concurrency issue, added transaction support and it worked.

Also asked some AI assistants (Copilot, jetbrain's) and they both said to me that knex sub queries don't inherit from the parent transactions. I'm having a hard time finding official documentation on this. I'll try digging a bit in Knex's codebase to confirm

Marc-Roig: interesting!

packages/core/database/src/query/helpers/order-by.ts
Convly: I didn't touch this one, I'll leave it for now :)

tests/api/core/strapi/api/sort/sort.test.api.ts
Convly: Good catch

tests/api/core/strapi/api/sort/sort.test.api.ts
Convly: Yep, done in `DX-1473`

Convly:
> Do you think we need to flag it as a breaking change since anyone using getKnexQuery with a deep sort will now get a very different query than they would have in v4?

Deep sorting was not even supported before this. So I wouldn't qualify this as a breaking change

alexandrebodin (approved):

tests/api/core/strapi/document-service/resources/fixtures/article.js
```diff
@@ -0,0 +1,59 @@
+module.exports = (fixtures) => {
```
jhoward1994, line 1: Do we need 'use strict' in these files?

tests/api/core/content-manager/api/number-of-draft-relations.test.api.js
```diff
@@ -465,32 +465,34 @@ describe('CM API - Number of draft relations', () => {
...
+    });
+
+    const p2 = await rq({
+      method: 'POST',
+      url: `/content-manager/collection-types/${UID_PRODUCT}`,
+      body: {
+        name: 'New Product 2',
+        categories: [categoryIds[1], categoryIds[0]],
+      },
+    });
```
jhoward1994, line 484:
```suggestion
    // Given products have a oneToMany relation with categories, adding categoryIds[0] here // steals the relation from p1. Hence why we are expecting totalDraftRelations to equal 3 below
    
    const p2 = await rq({
      method: 'POST',
      url: `/content-manager/collection-types/${UID_PRODUCT}`,
      body: {
        name: 'New Product 2',
        categories: [categoryIds[1], categoryIds[0]],
      },
    });
```

packages/core/database/src/query/helpers/order-by.ts
```diff
@@ -67,3 +71,155 @@ export const processOrderBy = (orderBy: OrderBy, ctx: OrderByCtx): OrderByValue[
...
+    partitionedQuery
+      .select(
+        // Always select baseQuery.id
+        prefix(baseQueryAlias, 'id'),
+        // Sort columns
+        ...orderByColumns
+      )
+      // The row number is used to assign an index to every row in every partition
+      .rowNumber(COL_STRAPI_ROW_NUMBER, function (subQuery) {
+        // Each row number is assigned using the sort columns depending on their partition
```
jhoward1994, line 158: ?

```suggestion
      .rowNumber(COL_STRAPI_ROW_NUMBER, (subQuery) => {
```

jhoward1994 (approved):

jhoward1994:
QAed through the content API and and everything seems to be working for me 👍🏻

Bassel17 (approved):
LGTM

Issue #20508, opened 2024-06-13
CTB : user is initially directed to Advanced settings tab instead of Basic settings tab when creating a field


yanniskadiri:
Reported by Marion Lemaire during the Releases Blitz session: https://app.dashcam.io/replay/666ab36b12bbe900729758c7?share=1iHLpq7qlQ8050o8lJU0Q

When creating a CT in CTB, with many fields : 

I noticed that, when I last was on Advanced settings when creating a field, when I start creating another field, I'm directed to Advanced settings tab instead of Basic settings tab

it can be confusing as I many times filled up "default value" instead of "name of the field"

hanpaine:
Duplicate of #20584

PR #20545, opened 2024-06-18
fix: ctb ui & state issues

alexandrebodin:
### What does it do?

Fix several UI issues in CTB

- field type color
- component icon picker 

Closes https://github.com/strapi/strapi/issues/20508


simotae14 (approved):
👍🏽

Issue #20483, opened 2024-06-10
Filters not always working in v5


simotae14:
### Required System information

- Node.js version: v20.10
- NPM version: v10.2.3
- Strapi version: 5.0.0-beta.9
- Database: sqlite
- Operating system: macOS Monterey v12.6.3
- Is your project Javascript or Typescript: Typescript

### Describe the bug

The filters in v5 are not always working like in v5

For example in the Users Settings in v4 you can filter a field for a null value
![337593634-389dbef5-4945-4c43-87b8-2cedd0b9af0c](https://github.com/strapi/strapi/assets/2589748/1110a925-7a7f-469f-b34a-1c3ed92cce65)
![337593752-90e467a0-ad5b-4b7c-ae6e-0939c58c0095](https://github.com/strapi/strapi/assets/2589748/f370f86a-fb64-4a4c-800f-0a5f8a3515f7)

instead in v5 you can see how this filter doesn't work

https://github.com/strapi/strapi/assets/2589748/3ae28f7e-8776-4943-8b48-7d48ac1b1096

Issue #20058, opened 2024-04-08
500 when adding filter in content-type list on a many2many relation


verdverm:
### Required System information

- Node.js version: 20.x
- Strapi version: v5.0.0-beta.2
- Database: Postgres
- Operating system: linux
- Is your project Javascript or Typescript:

### Describe the bug

500 error when filtering on relation column where other side has no entries in the database yet

`function lower(integer) does not exist` seems to be the root of the error. Looks like the generated query is trying to call `lower` on an `id` field which is an int type

### Steps to reproduce the behavior

1. Blog content-type
2. Tags content-type
3. Add many2many relation between
4. Have some blogs, no tags yet
5. Try to add tag filter, `contains (insensitive)`

### Expected behavior

Filter should be added, probably no results?

It looks like the filter on a relation cannot specify a subfield. I'd like to be able to filter based on the value of the tag, not the id

### Error logs

```
[2024-04-08 02:42:25.785] error: select count(distinct "t0"."id") as "count" from "public"."blogs" as "t0" left join "public"."blogs_tags_lnk" as "t1" on "t0"."id" = "t1"."blog_id" left join "public"."tags" as "t2" on "t1"."tag_id" = "t2"."id" where ("t0"."published_at" is null and ((LOWER("t2"."id") LIKE LOWER($1))) and "t0"."published_at" is null) limit $2 - function lower(integer) does not exist
error: select count(distinct "t0"."id") as "count" from "public"."blogs" as "t0" left join "public"."blogs_tags_lnk" as "t1" on "t0"."id" = "t1"."blog_id" left join "public"."tags" as "t2" on "t1"."tag_id" = "t2"."id" where ("t0"."published_at" is null and ((LOWER("t2"."id") LIKE LOWER($1))) and "t0"."published_at" is null) limit $2 - function lower(integer) does not exist
    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/dist/parser.js:283:98)
    at Parser.handlePacket (/app/node_modules/pg-protocol/dist/parser.js:122:29)
    at Parser.parse (/app/node_modules/pg-protocol/dist/parser.js:35:38)
    at Socket.<anonymous> (/app/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:518:28)
    at Socket.emit (node:domain:488:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
[...]

PR #20548, opened 2024-06-18
fix(content-manager): fix filters bugs with relations, enums and uids

Feranchz:
### What does it do?

Fix multiple issues related to filters on v5:
* Relations filter are using the mainField instead of always the id
* UID filtering is supported 
* enum filtering has the right options

### Related issue(s)/PR(s)

#20483 
#20058


madhurisandbhor (approved):
LGTM 👍  
Is it possible to add tests to check filters passed on submit for different types?

PR #20553, opened 2024-06-19
fix: self referencing relations

Marc-Roig:
### What does it do?

Fixes an issue only reproducible in Postgres and on new Strapi apps (not in the monorepo itself)

The database relation stealing was failing, and was due to the knex connection being "reused" to create subqueries.

```ts
// Before
connection
	.from(tableName)
	.where(
		'column', 
	     connection.select(...)  // Reused same connection
) 

// After
connection
	.from(tableName)
	.where(
		'column', 
	     (con) => con.select(...)  // New connection reference
) 

```

Aparently the former created a self reference inside the connection query

### DIFF
--- a/packages/core/database/src/entity-manager/regular-relations.ts
+++ b/packages/core/database/src/entity-manager/regular-relations.ts
@@ -24,7 +24,7 @@ declare module 'knex' {
 }
 
 //  TODO: This is a short term solution, to not steal relations from the same document.
-const getDocumentSiblingIdsQuery = (con: Knex, tableName: string, id: ID) => {
+const getDocumentSiblingIdsQuery = (tableName: string, id: ID) => {
   // Find if the model is a content type or something else (e.g component)
   // to only get the documentId if it's a content type
   const models: Model[] = Array.from(strapi.db.metadata.values());
@@ -37,20 +37,17 @@ const getDocumentSiblingIdsQuery = (con: Knex, tableName: string, id: ID) => {
     return [id];
   }
 
-  return (
+  // NOTE: SubQueries are wrapped in a function to not reuse the same connection,
+  // which causes infinite self references
+  return (con: Knex) => {
     con
       .from(tableName)
       // Get all child ids of the document id
       .select('id')
-      .where(
-        'document_id',
-        con
-          .from(tableName)
-          // get document id related to the current id
-          .select('document_id')
-          .where('id', id)
-      )
-  );
+      .where('document_id', (con2: Knex) =>
+        con2.select('document_id').from(tableName).where('id', id)
+      );
+  };
 };
 
 /**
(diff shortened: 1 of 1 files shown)


packages/core/database/src/entity-manager/regular-relations.ts
```diff
@@ -83,7 +80,11 @@ const deletePreviousOneToAnyRelations = async ({
     .delete()
     .from(joinTable.name)
     // Exclude the ids of the current document
-    .whereNotIn(joinColumn.name, getDocumentSiblingIdsQuery(con, joinColumn.referencedTable!, id))
+    .whereNotIn(
+      // @ts-expect-error - knex incorrectly expects a string array
+      joinColumn.name,
```
Marc-Roig, line 85: @alexandrebodin  only when the second argument is a function, knex expects the first one to be an array of strings, do you know if that should be the case? This specific setup seems to work 🤷‍♂️  and the documentation doesn't mention anything about that either

Marc-Roig:
I am going to create an experimental to validate the fix

Convly (commented):
What a nice timing :smiley: 

https://github.com/strapi/strapi/pull/20543/files#r1645584094

Depending on whether you want to include the transactions support, we could merge our work here or in the other PR

packages/core/database/src/entity-manager/regular-relations.ts
Convly: That's weird, I didn't have the issue in my PR :thinking:

Marc-Roig:
@Convly oops, indeed. Lets see who can merge first, if I merge this first I can fix the conflicts on your branch 🚀

packages/core/database/src/entity-manager/regular-relations.ts
Marc-Roig: this only happened to me if the second argument was a function (qb) => qb.subquery
When using a regular subquery , it worked fine 😞

alexandrebodin: super weird 🤔  

Another options is to build the query this where. not sure the typings would work better but who knows
```
 .where((builder) =>
    builder.whereIn('id', [1, 11, 15]).whereNotIn('id', [17, 19])
  )
```

Marc-Roig:
@alexandrebodin I did require the func expressions, just that I made the `getDocumentSiblingIdsQuery` return a function instead, which makes it a bit cleaner

packages/core/database/src/entity-manager/regular-relations.ts
Marc-Roig: fixes the type issue I do see the original max stack size error 🙃

Marc-Roig:
ah!! missunderstood your comment :) yes! It didn't require using `this`

PR #20557, opened 2024-06-19
fix(core): disallow required on relation fields on schemas

Feranchz:
### What does it do?

* Check required is not present on relation fields on schemas and throw an error asking user to change it if it's present. 
* Remove required check from relations fields

### DIFF
--- a/packages/core/core/src/utils/transform-content-types-to-models.ts
+++ b/packages/core/core/src/utils/transform-content-types-to-models.ts
@@ -158,13 +158,6 @@ export const transformAttribute = (
         },
       };
     }
-    case 'relation': {
-      if (attribute.required) {
-        throw new Error(
-          `Required relations are not supported, please remove it from ${contentType.collectionName} schema on the attribute ${name}`
-        );
-      }
-    }
     default: {
       return attribute;
     }
--- a/packages/core/core/src/utils/transform-content-types-to-models.ts
+++ b/packages/core/core/src/utils/transform-content-types-to-models.ts
@@ -158,6 +158,13 @@ export const transformAttribute = (
         },
       };
     }
+    case 'relation': {
+      if (attribute.required) {
+        throw new Error(
+          `Required relations are not supported, please remove it from ${contentType.collectionName} schema on the attribute ${name}`
+        );
+      }
+    }
     default: {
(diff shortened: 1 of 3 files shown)


packages/core/core/src/utils/transform-content-types-to-models.ts
```diff
@@ -158,6 +158,13 @@ export const transformAttribute = (
         },
       };
     }
+    case 'relation': {
+      if (attribute.required) {
+        throw new Error(
+          `Required relations are not supported, please remove it from ${contentType.collectionName} schema on the attribute ${name}`
+        );
+      }
+    }
```
Marc-Roig, line 167: If we did allow this setup in the content type schema in v4, we should notify the docs team to include this in the list of v5 breaking changes

edporras: I recall quite a few comments by people who added the required attribute by hand so the UI would reflect it so, yes, a huge notice alerting this as a potential breaking change should be included in the release notes.

packages/core/core/src/utils/transform-content-types-to-models.ts
```diff
@@ -158,6 +158,13 @@ export const transformAttribute = (
         },
       };
     }
+    case 'relation': {
+      if (attribute.required) {
```
alexandrebodin, line 162: I don't think we should prevent this here, (not the right place and considering we want to support it a bit later during RC if possible I don't see the point of make it a schema breaking change)

Feranchz: I thought we wanted it to be not allowed 🤔 

In that case, instead of throwing an error I guess makes sense remove the required attribute in this same place in case if exists so it's not considered in the model.

Worth mention that in case you make a change in the CTB the required disappear because is not sent to the backend

alexandrebodin: I think it's fine just not applying a required in the validator only for now until we can support it (target is GA for that)

Marc-Roig (approved):

alexandrebodin:
Can you make sure we have a ticket to support it again for GA 🙏  cc @oiorain

Issue #20555, opened 2024-06-19
[V5] Relations are not removed after saving


simotae14:
### Required System information

 - Node.js version: 20
 - NPM version: 10.1.0
 - Strapi version: v5.0.0-beta.9
 - Database: sqlite
 - Operating system: MacOs
 - Is your project Javascript or Typescript: TS

### Describe the bug

When you try to remove one relation from an entry and then click save the relation just deleted comes back again

### Steps to reproduce the behavior

1. Go to a Collection with a relation field and create some entries (with some relations)
2. then from the edit view try to remove the relation and save
3. after the successful notification the relation just deleted appears again

### Expected behavior

I want to see the relation removed

### Screenshots

https://github.com/strapi/strapi/assets/2589748/8e00e72d-dc6d-41d1-8a08-e9968340839d

### Ticket related 

Probably similar to this one https://github.com/strapi/strapi/issues/20217

Issue #20554, opened 2024-06-19
V5: When we delete the entries from the List view the UI is not updated without the deleted entries


simotae14:
### Required System information

 - Node.js version: 20
 - NPM version: 10.1.0
 - Strapi version: v5.0.0-beta.9
 - Database: sqlite
 - Operating system: MacOs
 - Is your project Javascript or Typescript: TS

### Describe the bug

When you delete some entries in the list view, with or without Bulk publish then, after saving, the entries remain in place, the list is updated only after a Refresh

### Steps to reproduce the behavior

1. Go to a Collection and create some entries
2. then from the list view try to delete some of them
3. after the successful notification the entries remain in place

### Expected behavior

I want to see the updated list

### Screenshots

https://github.com/strapi/strapi/assets/2589748/eae2efd9-89ac-441f-bbcc-a768979bf5c8

PR #20558, opened 2024-06-19
Fix multiple CM/i18N/admin state loading issues

alexandrebodin:
### What does it do?

- Prevent over fetching unecessary information
- Avoid using outdated RTK data when fetching new data

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -97,7 +97,11 @@ const ListViewPage = () => {
   });
 
   const params = React.useMemo(() => buildValidParams(query), [query]);
-  const { data, error, isFetching } = useGetAllDocumentsQuery({
+  const {
+    currentData: data,
+    error,
+    isLoading,
+  } = useGetAllDocumentsQuery({
     model,
     params,
   });
@@ -170,7 +174,7 @@ const ListViewPage = () => {
     return formattedHeaders;
   }, [displayedHeaders, formatMessage, list, runHookWaterfall, schema?.options?.draftAndPublish]);
 
-  if (isFetching) {
+  if (isLoading) {
     return <Page.Loading />;
   }
 
@@ -239,7 +243,7 @@ const ListViewPage = () => {
       />
       <Layouts.Content>
         <Flex gap={4} direction="column" alignItems="stretch">
-          <Table.Root rows={results} headers={tableHeaders} isLoading={isFetching}>
+          <Table.Root rows={results} headers={tableHeaders} isLoading={isLoading}>
             <TableActionsBar />
             <Table.Content>
               <Table.Head>
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -97,11 +97,7 @@ const ListViewPage = () => {
   });
 
   const params = React.useMemo(() => buildValidParams(query), [query]);
-  const {
-    currentData: data,
-    error,
-    isLoading,
-  } = useGetAllDocumentsQuery({
+  const { data, error, isFetching } = useGetAllDocumentsQuery({
     model,
     params,
   });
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -170,7 +174,7 @@ const ListViewPage = () => {
     return formattedHeaders;
   }, [displayedHeaders, formatMessage, list, runHookWaterfall, schema?.options?.draftAndPublish]);
 
-  if (isFetching) {
+  if (isLoading) {
```
Feranchz, line 177: Is this right? I thought we could have refetching on listView and for that reason start using it

alexandrebodin: I prefered using `currentData` from rtk so I get and empty array instead of a loading page everytime we refetch.

packages/plugins/i18n/admin/src/hooks/useI18n.ts
```diff
@@ -43,14 +43,15 @@ const useI18n: UseI18n = () => {
...
+  // TODO: use specific hook to get schema only
   const { schema } = useDocument(
     {
       // We can non-null assert these because below we skip the query if they are not present
       collectionType: params.collectionType!,
       model: params.slug!,
     },
     {
-      skip: !params.slug || !params.collectionType,
+      skip: true,
```
remidej, line 54: Why do we now always skip? Also you can use selectFromResult if you want more specific

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
remidej: I think it's more clear for the user to show him a loading screen than an empty view for a second. We definitely have too many spinners in Strapi, but IMO it should be fixed elsewhere (maybe react router loaders)

remidej (commented):
it does fix the edit view issue I was looking at :)

simotae14 (changes requested):
It seems not optimal 100% we have some weird ui effects

Feranchz:
> it seems working but there is a small visual issue, it seems like when we click save for a small amount of time the cached value is shown and after some milliseconds the correct one is shown, as you can see in this video

@simotae14 I think to fix this we gonna need to implement an optimistic update here, but I think it's okay like this for now and we work on that later as improvement 🤔

packages/plugins/i18n/admin/src/hooks/useI18n.ts
alexandrebodin: that skip option doesn't allow skipping the schema loading but only skipping a document data here there is never a reason to load a document data. 

Overall that hook should not even be used everywhere like it is right now it creates so much confusion on what it does it's very unsafe

packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
```diff
@@ -20,6 +20,7 @@ const LocaleListCell = ({
   collectionType,
   model,
 }: LocaleListCellProps) => {
+  // TODO: avoid loading availableLocales for each row but get that from the BE
```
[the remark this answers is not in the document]
alexandrebodin: This should be handled BE side so the data is just there no point doing 10 queries even if optimized

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
alexandrebodin: it should just keep the old data until it can change the view but there are too many other issues to fix to get there for now. I don't have a preference I'll put back the full page loading if you prefer

simotae14:
ok, there are some lint errors to fix

alexandrebodin:
> it seems working but there is a small visual issue, it seems like when we click save for a small amount of time the cached value is shown and after some milliseconds the correct one is shown, as you can see in this video
> 
>  Registrazione.schermo.2024-06-19.alle.16.21.39.mov

I wasn't planning on fixing that many pbl so I'll stop there we can do more fixes later

simotae14 (commented):
it solves also this issue https://github.com/strapi/strapi/issues/20555 and https://github.com/strapi/strapi/issues/20554 👏🏽

PR #20559, opened 2024-06-19
Fix Can add two attributes with the same name and save 

### LINKED ISSUE #20364: Can add two attributes with the same name and save

- Node.js version: 20
- NPM version: 10
- Strapi version: 5.0.0-beta.7
- Database: sqlite
- Operating system: mac

### Describe the bug

The content type builder is not preventing from saving when it has two fields with the same name 

1. edit a content type
2. add a field of any type with the same name as an existing field
3.  click save

### Expected behavior

The input validation should check existing fields and I should be prevented from adding the second field.

Bassel17:
### Why is it needed?

- CTB allows adding an attribute with the same name causing issues when saving

### How to test it?

- Try to add an attribute to a contentType with an attribute name that already exist it should show and error like this
image

### Related issue(s)/PR(s)

fix #20364


Convly:
QAed and works well. Thanks!
image

PR #20561, opened 2024-06-19
feat(create-strapi-app): remove the cloud project creation part

nathan-pichon:
### What does it do?

Removing the cloud project creation part on create-strapi-app.

### Why is it needed?

As we experienced some issues with the actual flow, we decided to remove that part.

Issues:
- Delays the strapi project creation by ~10s
- Active the free trial when the project is generated
- If the user remove his project and creates another, he cannot re-use the Cloud CLI to create a project as he as already used his free trial.

### How to test it?

 The create-strapi-app command should not create a project on Cloud anymore.
The project should be created on Cloud during the first `strapi deploy` command (this is already handled in the deploy command).

### Related issue(s)/PR(s)

### DIFF
--- a/packages/cli/create-strapi-app/src/cloud.ts
+++ b/packages/cli/create-strapi-app/src/cloud.ts
@@ -1,5 +1,4 @@
 import inquirer from 'inquirer';
-import { resolve } from 'node:path';
 import { cli as cloudCli, services as cloudServices } from '@strapi/cloud-cli';
 import parseToChalk from './utils/parse-to-chalk';
 
@@ -16,13 +15,13 @@ function assertCloudError(e: unknown): asserts e is CloudError {
   }
 }
 
-export async function handleCloudProject(projectName: string): Promise<void> {
+export async function handleCloudLogin(): Promise<void> {
   const logger = cloudServices.createLogger({
     silent: false,
     debug: process.argv.includes('--debug'),
     timestamp: false,
   });
-  let cloudApiService = await cloudServices.cloudApiFactory();
+  const cloudApiService = await cloudServices.cloudApiFactory();
   const defaultErrorMessage =
     'An error occurred while trying to interact with Strapi Cloud. Use strapi deploy command once the project is generated.';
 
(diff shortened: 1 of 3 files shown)


packages/cli/create-strapi-app/src/cloud.ts
```diff
@@ -86,22 +58,14 @@ export async function handleCloudProject(projectName: string): Promise<void> {
           const message =
             typeof e.response.data === 'string'
               ? e.response.data
-              : 'We are sorry, but we are not able to create a Strapi Cloud project for you at the moment.';
-          if (projectCreationSpinner.isSpinning) {
-            projectCreationSpinner.fail(message);
-          } else {
-            logger.warn(message);
-          }
+              : 'We are sorry, but we are not able to login you to Strapi Cloud.';
```
Jimimimi, line 61:
```suggestion
              : 'We are sorry, but we are not able to log you into Strapi Cloud at the moment.';
```

packages/cli/create-strapi-app/src/cloud.ts
```diff
@@ -16,13 +15,13 @@ function assertCloudError(e: unknown): asserts e is CloudError {
   }
 }
 
-export async function handleCloudProject(projectName: string): Promise<void> {
+export async function handleCloudProject(): Promise<void> {
```
Jimimimi, line 18:
```suggestion
export async function handleCloudLogin(): Promise<void> {
```
Probably better to rename the function to something more representative

packages/cli/create-strapi-app/src/create-strapi-app.ts
```diff
@@ -61,8 +56,7 @@ async function generateApp(
   }
 
   if (!options.skipCloud) {
-    checkRequirements();
-    await handleCloudProject(projectName);
+    await handleCloudProject();
```
Jimimimi, line 59:
```suggestion
    await handleCloudLogin();
```

Probably better to rename the function to something more representative

Jimimimi (changes requested):
Some small recommendations

Jimimimi (approved):

abdallahmz (approved):
Tested locally, looks all good!

Issue #20217, opened 2024-04-26
CM : Back arrow issue >> Relation field cache is not invalidated when going "back"


MarionLemaire:
## Bug report

- Node.js version: 20.9.0
- Strapi version: v5.0.0-beta.5
- Operating system: Windows

### Describe the bug

In CM, when creating a new entry for a collection type with Relation field, after saving, if I click the Back arrow to create a new entry, the Relation field is still filled-in with the infos from previous entry (all other fields are empty).

Loom here : https://www.loom.com/share/2a740320420f4e81a09927c614412530?sid=1cf33509-4824-4ad4-8a35-973ce62fef8b

### Steps to reproduce the behavior

1. Create a Collection type with a Relation field
2.Go to Content Manager
3. Create a first entry with this CT, with one or many relations in the Relation field
4. Save this entry (and/or publish it)
5. Click Back arrow
6. Check Relation field : still filled in with previous info

### Expected behavior

Relation field should be cleared up and empty

### Screenshots

Entry #1 saved and published : 
![image](https://github.com/strapi/strapi/assets/97893822/2c2f5bcf-b233-4063-ac72-66722b7f177d)

Entry #2 after clicking on Back arrow : 
![image](https://github.com/strapi/strapi/assets/97893822/117243b5-13ce-4c87-aaff-d8005ce82f9e)

Dashcam info here : https://app.dashcam.io/share/FU79RvPPj6PcV2obpIO9w

derrickmehaffy:
I see you @MarionLemaire 

![Image](https://github.com/strapi/strapi/assets/8593673/a06b1813-fd4d-488e-88c0-a0ae41fe276f)

And I raise you

![Image](https://github.com/strapi/strapi/assets/8593673/ff79af90-4237-496c-bee1-358cb95da461)

joshuaellis:
This could be either local state in the relation input or it might be the form state not correctly clearing itself. I don't think it'll be a redux cache, but if those formers seem to be normal then check out redux?

PR #20562, opened 2024-06-19
Fix CM : Back arrow issue >> Relation field cache is not invalidated when going "back"

simotae14:
### What does it do?

It fixes the cache issue we had in the CM when we create an entry with Relations and after saving we click the Back button, we are redirected to the create entry page with the form not all empty, the relations are still in place

### Why is it needed?

We want to show all the fields empty in the Create entry page

### How to test it?

1. Create an entry in a collection with a Relation field and add a relation
2. Save the entry and then click the Back Button
3. you need to see the Form empty, even for the Relation field

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20217

### DIFF
--- a/packages/core/content-manager/admin/src/services/relations.ts
+++ b/packages/core/content-manager/admin/src/services/relations.ts
@@ -41,6 +41,9 @@ const relationsApi = contentManagerApi.injectEndpoints({
       }
     >({
       query: ({ model, id, targetField, params }) => {
+        if (id === undefined) {
+          return { url: '', method: 'GET', config: { params: {} } }; // Return a dummy URL to prevent the query
+        }
         return {
           url: `/content-manager/relations/${model}/${id}/${targetField}`,
           method: 'GET',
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/services/relations.ts
```diff
@@ -41,6 +41,9 @@ const relationsApi = contentManagerApi.injectEndpoints({
       }
     >({
       query: ({ model, id, targetField, params }) => {
+        if (id === undefined) {
+          return { url: '', method: 'GET', config: { params: {} } }; // Return a dummy URL to prevent the query
+        }
```
remidej, line 46: I'm not 100% sure about this pattern. Isn't there a place we can invalidate some cache? Or where we could use isFetching to ensure we don't show stale data?

simotae14: The problem is before we were calling the GET api only when you have an ID, otherwise the API is not called, when you create a new entry with a Relation the GET api is called and the response coming from the API is assigned to the relations variable. Then when you click on the Back button, the API is not called and the value of relations remains the same...it is because of that I tried to mock the call to an API when you don't have an Id to respond with an empty array and then populate the relations variable with this array. unfortunately isFetching in this case has no effect

simotae14: the only way to do that is to invalidate the cache when you click on the BackButton

Feranchz:
This change is still needed after this #20565 ? @simotae14 @remidej

remidej:
Ah I wasn't sure it would 100% override this one, I should have mentioned it before sorry @simotae14

Issue #20569, opened 2024-06-20
CM : Date field issues in Components


simotae14:
### Required System information

- Node.js version: v20.10
- NPM version: v10.2.3
- Strapi version: 5.0.0-beta.9
- Database: sqlite
- Operating system: macOS Monterey v12.6.3
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When saving an entry in the Content manager with a date sometimes it happens that the date is saved with a different value and if you try to delete the date it is not possible

Try the different date fields: Date, Datetime and Time, and their use in a Collection or inside a component

You can see the error in this video
https://app.dashcam.io/replay/667409a910c6b5007053a434?share=7vYxdwRmQlZ1VC1QxiqkA

simotae14:
I am gonna close it because it is fixed with this PR https://github.com/strapi/strapi/pull/20571

PR #20571, opened 2024-06-20
V5 fix: wrong date saved

simotae14:
### What does it do?

Fix the issue we had with Date fields, everytime we change a date and save we were saving the locale date and not the utc one resulting that the value shown is the day before

### Why is it needed?

Otherwise after saving the date we show the day before

### How to test it?

1. Create an entry on a collection with a date field, add a date value and save
2. After save you need to see the date saved and not the date with the day before

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20569

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
@@ -11,7 +11,7 @@ import { InputProps } from './types';
 const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
     const { formatMessage } = useIntl();
-    const field = useField(name);
+    const field = useField<Date | null>(name);
     const fieldRef = useFocusInputField<HTMLInputElement>(name);
 
     const composedRefs = useComposedRefs(ref, fieldRef);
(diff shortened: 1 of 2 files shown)


madhurisandbhor:
I think this is a datepicker issue and is fixed recently [here](https://github.com/strapi/strapi/pull/20503). Could you please check if you have updated version in strapi.

simotae14:
I am using v5/main and we have the same issue in the last beta version

packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
```diff
@@ -11,7 +11,7 @@ import { InputProps } from './types';
 const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
     const { formatMessage } = useIntl();
-    const field = useField<Date>(name);
+    const field = useField(name);
```
remidej, line 14: How about `Date | null`?

remidej (approved):
qa looks good 👍

packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
simotae14: yeah, I used the same instruction we used in the Date.tsx but for sure I can change both lines

PR #20572, opened 2024-06-20
feat(cli): trigger login sequence when token is missing or invalid 

gonbaum:
### What does it do?

When a user who is not logged in tries to execute a command that requires authentication, the system prompts the user with an option to log in. If the user confirms, the login process is initiated, allowing the user to authenticate before proceeding with the command.

### Why is it needed?

Currently, when a user who is not logged in or has a non-existent, invalid, or expired token tries to execute a command that requires authentication, the process fails, and the command execution is halted.

### How to test it?

Scenario: Using the deploy command without being logged in
- Expected Behavior:
	- The user is informed that they need to log in to use the command and is asked if they want to log in.
	- If the user selects login confirmation, the login flow is launched.
	- After successful authentication, the deploy or project creation flow continues.
	- If the user declines to log in, nothing else happens.

Scenario: Using the deploy command with an expired token
- Expected Behavior:
	- The user is informed that their token has expired and that they need to log in again.
	- The user is asked if they want to log in.
	- If the user selects login confirmation, the login flow is launched.
	- After successful authentication, the deploy or project creation flow continues.
	- If the user declines to log in, nothing else happens.

### Related issue(s)/PR(s)


packages/cli/cloud/src/create-project/action.ts
```diff
@@ -67,7 +61,24 @@ export default async (ctx: CLIContext) => {
...
     return data;
   } catch (e: Error | unknown) {
-    spinner.fail('Failed to create project on Strapi Cloud.');
-    await handleError(ctx, e as Error);
+    if (e instanceof AxiosError && e.response?.status === 401) {
+      spinner.fail('Oops! Your session has expired. Please log in again.');
+      await eraseToken();
+      if (await promptLogin(ctx)) {
+        try {
+          const spinner = logger.spinner('Setting up your project...').start();
```
abdallahmz, line 69: The way it's done means
- we have duplicate code
- it's not try catching & retrying for things outside the try/catch
- (subjective) it's a bit more complex to understand what's happening

My suggestion:
- Have a function "validateTokenValidityOrLogin", which any process can call before proceeding
- `await validateTokenValidityOrLogin()`
- This function would ensure beforehand that no 401 error can happen during our processes


So the PR would consist of:
- 1 new function added
- 1 call of this function wherever needed

I feel like this way the code would be clearer/simpler. WDYT?

gonbaum: Thanks for the feedback! Actually, I don't feel like adding a new function since we already have 3 functions in the token service:

- getValidToken()
- validateToken()
- isTokenValid() 

We somehow ended up with 3 functions for the beta but I believe we can merge some of these (maybe get and validate) into one. 

The reason I included a check in the catch block is that the current validation of the token is only done against public keys. However, the token could still be expired or rejected in `platform-api `(now less likely since I check for expiration in the token headers). But we can include a API call (wondering if there should be a check against our table of users 🤔 ) 

But yes, overall I think it could be simpler and if we're not rushed we can do this better.

nathan-pichon: As the token expiration is already checked on CLI-client side, I believe `ensureValidToken` would just add another call and another endpoint that is already handled by secured API endpoints (like the one we use to call platform-api).
I don't see any benefit in doing that extra call, the only one would be "there is no more 401", but I don't think it's a benefit. 😅 4xx are legit errors. 🤷 
Or did I misunderstood what you said @abdallahmz ?

nathan-pichon: I believe there are some use cases, like when we ask users to enter information. To avoid them to get a 401 and redo everything once connected 😬

abdallahmz: Since we don't have a proper "no-side-effect" endpoint to validate the token, and it adds an extra call we might not want, we can drop the idea 🤔 
We'll find something else to improve the solution, otherwise we'll still have the current implementation as backup! (I think we can at least have a "createProject" function to avoid the copy pasted code, that'd be the minimum)

packages/cli/cloud/src/services/token.ts
```diff
@@ -72,16 +71,16 @@ export async function tokenServiceFactory({ logger }: { logger: CLIContext['logg
...
     return new Promise<void>((resolve, reject) => {
       jwt.verify(idToken, getKey, (err: VerifyErrors | null) => {
-        if (err) {
-          reject(err);
-        } else {
-          resolve();
-        }
+        if (err) reject(err);
+        if (decodedToken.payload.exp < Math.floor(Date.now() / 1000))
+          reject(new Error('Token is expired'));
```
nathan-pichon, line 82: I don't think block-less conditions should be allowed 🤔

abdallahmz (approved):
Looks good to me 👍

abdallahmz (approved):

PR #20576, opened 2024-06-21
Update LICENSE packing for packages for v4

Convly:
### What does it do?

Move EE terms directly into the packages' root `LICENSE` file.

Note: Two packages didn't have any `LICENSE` file, added it for both (`packages/plugins/cloud` and `packages/plugins/color-picker`).


alexandrebodin (approved):
So the only thing I see is the content-releases go a full license without the /ee path and all the rest got ee mentioned ? LGTM

derrickmehaffy:
LGTM

PR #20582, opened 2024-06-21
fix: incorrect populate when using morph relations

Marc-Roig:
### What does it do?

Polymorphic relations are theoretically not supported, but we do allow them to be defined in the schema.
the Kitchensink Content Type had one, and it was not possible to create or update one in v5/main.

This fix prevents content history and the document service to incorrectly populate morph relations, which need a different kind of syntax than regular relations.
Once we work on polymorphic relations, we will need to fix this.


remidej (approved):
I can use kitchesink again ❤️❤️

PR #20588, opened 2024-06-21
fix(admin): not authenticated upon registration

### LINKED ISSUE #20332: On Initial Install Get "You don't have permission to access content error"  when looking at User collection type until you refresh.

- Node.js version: v20.11.1
- NPM version: 10.2.4 
- Strapi version: 5.0.0-beta.7
- Database: sqlite
- Is your project Javascript or Typescript:

### Describe the bug

After initial install when you navigate to `COLLECTION TYPES` and click on `User` you get the following message.
```
You don't have permission to access content error
```

as well as don't see the rest of the menu options until you reload the app.

1. Install Strapi
2. Click on the content manager link
3. Select `User` collection type

### Expected behavior

Should not see the message and have all the available items in menu.

### Screenshots

Video showing the steps to reproduce

https://github.com/strapi/strapi/assets/6153188/5605c4d3-f4a9-4aee-9977-f0534dc514f7

### Code snippets

### Additional context

Not a breaking issue but can hurt initial perception when trying out Strapi.  Due to most users being frustrated before figuring out that restarting or refreshing fixes the issue.

remidej:
### What does it do?

Fixes the issue where Strapi thought you didn't have Super Admin permissions immediately after creating the first admin user (until you refreshed)

Stops relying on useEffect to store the JWT updates in the localStorage, as it would run after the page redirection, so too late to be picked up by the fetch client

Also deletes a type that had been written twice in the same file

### How to test it?

Delete your admin users in your database, it will prompt Strapi to register a first admin user. After signing up, make sure you see all the expected options in the main nav.

### Related issue(s)/PR(s)

fixes #20332

### DIFF
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -50,7 +50,7 @@ interface AuthContextValue {
   isLoading: boolean;
   permissions: Permission[];
   refetchPermissions: () => Promise<void>;
-  setToken: (token: string | null) => void;
+  setToken: (token: string) => void;
   token: string | null;
   user?: User;
 }
@@ -144,11 +144,10 @@ const AuthProvider = ({ children, _defaultPermissions = [] }: AuthProviderProps)
     }
   }, [dispatch, user]);
 
-  React.useEffect(() => {
-    if (token) {
-      storeToken(token, false);
-    }
-  }, [token]);
+  const updateToken = React.useCallback((newToken: string) => {
+    setToken(newToken);
+    storeToken(newToken);
+  }, []);
 
   React.useEffect(() => {
     /**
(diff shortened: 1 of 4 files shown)


jhoward1994:
Works in QA for me 👍🏻

I'm wondering could this be replicated in one of our e2e specs? (login?) and therefore this fix could be figured by extending that test a bit?

jhoward1994:
Also the failing FE tests in `nx run @strapi/plugin-users-permissions:"test:front"` seem to be related to this change. Could it be because of the useEffect that was removed?

remidej:
@jhoward1994 thanks for testing!

- the frontend test failing was a real bug, I had forgotten a bit of code from a first fix attempt, the fetch client no longer dependends on useAuth now
- good idea for e2e tests! I added a tiny check to the register test (since the issue only occured upon registration). It looks for the CTB, since the CTB won't be displayed if you have no permissions loaded. I checked the before/after locally

remidej:
Looks like this will be made obsolete by #20609, will confirm

PR #20589, opened 2024-06-21
feat: Release settings page

simotae14:
### What does it do?

Add the Release Settings page to select the timezone to use
Schermata 2024-06-17 alle 10 22 06

### Why is it needed?

We want to give the ability to the user to select the default timezone to use in Releases

Page should consume these two endpoints:

Get settings
GET `/releases/settings`

Update settings
PUT `/releases/settings`

body:
```
{
  defaultTimezone: string | null
}
```

### How to test it?

Open the Strapi CMS with a EE license
- Go to the "Settings" and open the Release link under the "Global Settings"
- Select a Timezone from the dropdown and save it
  - then after the Successful message you can check if the value is saved just refreshing the page

### Related issue(s)/PR(s)

CS-811


packages/core/content-releases/admin/src/pages/SettingsPage.tsx
```diff
@@ -15,31 +82,112 @@ export const SettingsPage = () => {
...
+            return (
+              <>
+                <Layouts.Header
+                  primaryAction={
+                    <Button
+                      disabled={!modified || isSubmittingForm}
+                      loading={isSubmitting}
+                      startIcon={<Check />}
+                      type="submit"
+                      size="L"
```
remidej, line 108: I don't think size L is needed, the button doesn't look this big to me in the designs

packages/core/content-releases/admin/src/components/ReleaseModal.tsx
```diff
@@ -248,7 +248,7 @@ interface ITimezoneOption {
   value: string;
 }
 
-const getTimezones = (selectedDate: Date) => {
+export const getTimezones = (selectedDate: Date) => {
```
remidej, line 251: Maybe this belongs in the utils folder?

remidej (commented):
Tested ✅

packages/core/content-releases/admin/src/pages/SettingsPage.tsx
simotae14: ok

packages/core/content-releases/admin/src/components/ReleaseModal.tsx
```diff
@@ -22,7 +22,7 @@ import { useLocation } from 'react-router-dom';
 
 import { RELEASE_SCHEMA } from '../../../shared/validation-schemas';
 import { pluginId } from '../pluginId';
-import { getTimezoneOffset } from '../utils/time';
+import { getTimezones } from '../utils/getTimezones';
```
remidej, line 26: sorry to be annoying but I think getTimezones should be inside that utils/time, they're closely related

remidej (approved):
Tested again ⭐

PR #20603, opened 2024-06-21
fix(content-manager): blocks issues

Feranchz:
### What does it do?

Fix two bugs on Blocks inputs:
* Link selector was 100% broken because of changes on how the Popover works. With these changes should work as expected.
* Lists selector didn't work when multiple paragraphs are selected.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/utils/conversions.ts
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/utils/conversions.ts
@@ -1,4 +1,4 @@
-import { type Element, type Path, Editor, Transforms } from 'slate';
+import { type Element, Editor, Transforms } from 'slate';
 
 /**
  * Extracts some logic that is common to most blocks' handleConvert functions.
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/utils/conversions.ts
remidej, line 1: this will need lots of qa :) it runs the conversions of all types of blocks

remidej (commented):
You have failing unit tests fyi

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/utils/conversions.ts
Feranchz: I agree. But looks like the current implementation of the conversion fn didn't considerer selecting multiple blocks to apply a selector. This bug exists also in v4. 

We could just disable applying list selectors when multiple blocks are selected for now and then fix this? In my opinion is something that we need to take care soon or later, so I prefer to do it right now 

wdyt?

remidej (approved):
🚀

PR #20607, opened 2024-06-22
chore: bump glob from 7.2.0 to 10.4.2 in core/strapi

chrismuiruriz:
### What does it do?

This pull request updates the `glob` dependency in Strapi from version 7.2.0 to 10.4.2. The reason for this change is that versions of `glob` prior to 9 are deprecated and no longer supported (https://www.npmjs.com/package/glob/v/7.2.0?activeTab=code). Additionally, the older version includes the `inflight` package, which has a known vulnerability (https://github.com/isaacs/inflight-DEPRECATED-DO-NOT-USE/issues/18, https://security.snyk.io/package/npm/inflight/1.0.6).

### Why is it needed?

Updating `glob` to version 10.4.2 resolves the deprecated status and eliminates the vulnerability associated with the `inflight` package, thereby enhancing the security and maintainability of the Strapi package.

### How to test it?

To verify the behavior:
- Ensure you have set up a testing environment similar to the one specified in Strapi's documentation.
- Execute existing tests and ensure they pass without errors related to the `glob` functionality.
- Verify that the `glob` package functions correctly in scenarios where it is utilized within Strapi.

### Related issue(s)/PR(s)

This pull request is related to fixing the issues with deprecated `glob` versions and the `inflight` vulnerability.

(https://github.com/isaacs/inflight-DEPRECATED-DO-NOT-USE/issues/18, https://security.snyk.io/package/npm/inflight/1.0.6)


alexandrebodin:
Hi @chrismuiruriz any reason to upgrade to 9 and not 10 directly ?

chrismuiruriz:
> Hi @chrismuiruriz any reason to upgrade to 9 and not 10 directly ?

Hi @alexandrebodin I wanted to mimize the number of breaking changes any considering we were moving from v7 to v9 which is a full rewrite, with significant API and algorithm changes.

alexandrebodin:
Considering it is only used internally without user facing changes it's fine to upgrade to the latest as long as our internals work :) FYI Strapi v5 code is already on 10.3.10 and we didn't have any pbl moving to it there

chrismuiruriz:
> Considering it is only used internally without user facing changes it's fine to upgrade to the latest as long as our internals work :) FYI Strapi v5 code is already on 10.3.10 and we didn't have any pbl moving to it there

Got you, I can update the version and update the PR, is that okay?

chrismuiruriz:
> Hi @chrismuiruriz any reason to upgrade to 9 and not 10 directly ?

Hi @alexandrebodin I've pushed the change with Glob version 10.4.2.

alexandrebodin:
Hey, the tests aren't passing because the yarn.lock is outdated :) make sure to update it too 👍

chrismuiruriz:
> Hey, the tests aren't passing because the yarn.lock is outdated :) make sure to update it too 👍

Hi @alexandrebodin, seems all the tests are passing apart from `Vercel – contributor-docs` any idea why?

packages/core/strapi/src/load/glob.ts
```diff
@@ -1,15 +1,15 @@
...
  * Promise based glob
  */
-function promiseGlob(...args: [string, IOptions]): Promise<string[]> {
-  return new Promise((resolve, reject) => {
-    glob(...args, (err, files) => {
-      if (err) return reject(err);
-      resolve(files);
-    });
-  });
+async function promiseGlob(...args: [string, any]): Promise<string[]> {
```
alexandrebodin, line 6: This function has become unnecessary as glob is now async, you can remove this file and update `packages/core/strapi/src/load/load-files.ts` to use glob directly I think

alexandrebodin (changes requested):

packages/core/strapi/src/load/glob.ts
chrismuiruriz: Makes sense, on it

chrismuiruriz: @alexandrebodin I've updated

alexandrebodin (approved):

alexandrebodin:
Thank you for your contribution 💯

PR #20610, opened 2024-06-24
feat(content-manager): warn when publishing an entry with draft relations

jhoward1994:
### What does it do?

Shows a pop up when you try to publish an entry that has draft relations.

### Why is it needed?

V5 features and prevent accidental publishing with draft relations.

### How to test it?

### Related issue(s)/PR(s)

DX-1336


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -503,15 +506,113 @@ const PublishAction: DocumentActionComponent = ({
...
+          id: getTranslation('error.records.fetch-draft-relatons'),
+          defaultMessage: 'An error occurred while fetching draft relations on this document.',
+        }),
+      });
+    }
+  }, [isErrorDraftRelations, toggleNotification, formatMessage]);
+
+  React.useEffect(() => {
+    let totalDraftRelations = 0;
+    const localDraftRelations = new Set();
```
jhoward1994, line 539: @Marc-Roig 👋🏻 this PR is FE stuff but this is the part I'd like to get your opinion on as it relates to the relations connecting syntax.

It's used to count any draft relations that are only stored in the FE edit view form based on 'connect' relation arrays

Marc-Roig: >FE edit view form based on 'connect' relation arrays

should this take into account the FE local relations + the ones stored in database?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -503,15 +506,113 @@ const PublishAction: DocumentActionComponent = ({
...
+     */
+    const traverseAndExtract = (data: { [field: string]: any }) => {
+      Object.entries(data).forEach(([key, value]) => {
+        if (key === 'connect' && Array.isArray(value)) {
+          extractDraftRelations({ connect: value });
+        } else if (typeof value === 'object' && value !== null) {
+          traverseAndExtract(value);
+        }
+      });
+    };
```
Marc-Roig, line 568: not aware of what the data object has, but does it work with nested relations inside compos and dynamic zones?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -503,15 +506,113 @@ const PublishAction: DocumentActionComponent = ({
...
+        }
+
+        if (data) {
+          totalDraftRelations += data.data;
+          setCountOfDraftRelations(totalDraftRelations);
+        }
+      };
+
+      checkForDraftRelations();
+    }
```
Marc-Roig, line 604: when does this effect run? Is it every time the form values change? or when the user tries to publish an entry?

Even with this if (!isLoadingDraftRelations) check, if this runs on every keypress on the edit view form, we might be spamming the server with many requests that migth not be necessary

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
jhoward1994: Yep I've just tested those cases. I'll look at the e2e app template and see if that can be covered

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -503,15 +506,96 @@ const PublishAction: DocumentActionComponent = ({
...
+      traverseAndExtract(formValues);
+      setLocalCountOfDraftRelations(localDraftRelations.size);
+    }
+  }, [documentId, modified, formValues, setLocalCountOfDraftRelations]);
+
+  React.useEffect(() => {
+    if (documentId && !isLoadingDraftRelations) {
+      const fetchDraftRelationsCount = async () => {
+        const { data, error } = await countDraftRelations({
+          collectionType,
```
Marc-Roig, line 580: is it possible for this to create any kind of infinite loop?

`countDraftRelations` updating `isLoadingDraftRelations` and retriggering the `coundDraftRelations` again?

jhoward1994: Depending on the loading state of the query wasn't causing and infinite loops or
unecessary backend calls that I could see. I think you're right that we don't
need this dependency though. I've removed it

Marc-Roig (approved):
logic looks good , did not qa!

Bassel17 (approved):
LGTM! and QAed seems to be working

PR #20615, opened 2024-06-24
fix: several UI issues a typos

alexandrebodin:
### What does it do?

Fixes several UI issues & the CM validations

### DIFF
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -619,7 +619,7 @@ const reducer = <TFormValues extends FormValues = FormValues>(
         draft.values = setIn(
           state.values,
           action.payload.field,
-          newValue.length > 0 ? newValue : undefined
+          newValue.length > 0 ? newValue : []
         );
 
         break;
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -670,45 +670,16 @@ const useField = <TValue = any,>(path: string): FieldValue<TValue | undefined> =
 
   const handleChange = useForm('useField', (state) => state.onChange);
 
-  const formatNestedErrorMessages = (stateErrors: FormErrors<FormValues>) => {
-    const nestedErrors: Record<string, any> = {};
+  const error = useForm('useField', (state) => {
+    const error = getIn(state.errors, path);
 
-    Object.entries(stateErrors).forEach(([key, value]) => {
-      let current = nestedErrors;
+    if (isErrorMessageDescriptor(error)) {
+      const { values, ...message } = error;
+      return formatMessage(message, values);
+    }
 
-      const pathParts = key.split('.');
-      pathParts.forEach((part, index) => {
-        const isLastPart = index === pathParts.length - 1;
-
-        if (isLastPart) {
-          if (typeof value === 'string') {
-            // If the value is a translation message object or a string, it should be nested as is
-            current[part] = value;
-          } else if (isErrorMessageDescriptor(value)) {
-            // If the value is a plain object, it should be converted to a string message
-            current[part] = formatMessage(value);
-          } else {
-            // If the value is not an object, it may be an array or a message
-            setIn(current, part, value);
-          }
-        } else {
-          // Ensure nested structure exists
-          if (!current[part]) {
-            const isArray = !isNaN(Number(pathParts[index + 1]));
-            current[part] = isArray ? [] : {};
-          }
-
-          current = current[part];
-        }
-      });
-    });
-
-    return nestedErrors;
-  };
-
-  const error = useForm('useField', (state) =>
-    getIn(formatNestedErrorMessages(state.errors), path)
-  );
+    return error;
+  });
 
   return {
     initialValue,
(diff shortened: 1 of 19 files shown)


packages/core/content-type-builder/admin/src/components/FormModal/attributes/validation/common.ts
```diff
@@ -118,7 +118,11 @@ const createTextShape = (usedAttributeNames: Array<string>, reservedNames: Array
         name: 'isValidRegExpPattern',
         message: getTrad('error.validation.regex'),
         test(value) {
-          return new RegExp(value || '') !== null;
+          try {
+            return new RegExp(value || '') !== null;
```
Convly, line 122: When can this fail?

Convly: Also shouldn't this use ?? to avoid replacing falsy values?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Initializer.tsx
```diff
@@ -46,11 +46,6 @@ const Initializer = ({ disabled, name, onClick }: InitializerProps) => {
           </Flex>
         </Flex>
       </Box>
-      {field.error && (
```
Convly, line 49: Why are we removing the error logging?

packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -670,45 +670,9 @@ const useField = <TValue = any,>(path: string): FieldValue<TValue | undefined> =
...
-      });
-    });
-
-    return nestedErrors;
-  };
-
-  const error = useForm('useField', (state) =>
-    getIn(formatNestedErrorMessages(state.errors), path)
-  );
+  const error = useForm('useField', (state) => {
```
Convly, line 673: Could you explain quickly why we're removing all this logic?

packages/plugins/users-permissions/admin/src/pages/Providers/index.jsx
```diff
@@ -63,7 +63,7 @@ export const ProvidersPage = () => {
 
   const submitMutation = useMutation((body) => put('/users-permissions/providers', body), {
     async onSuccess() {
-      await queryClient.invalidateQueries(['users-permissions', 'providers']);
+      await queryClient.invalidateQueries(['users-permissions', 'get-providers']);
```
Convly, line 66: Was that a typo? Don't we have a type for the invalid queries to help up with that?

alexandrebodin: Yes, this is old queries that didn't get migrated and if you look at the query above in the code the cache tag is `get-providers` :)

packages/core/content-type-builder/admin/src/components/FormModal/attributes/validation/common.ts
alexandrebodin: if you pass an invalid regexp like '*' it throws I noticed that while testing so just wrapping in a try catch makes the error much better. 

value is a string so the only falsy value is empty string which is the default anyway

remidej (commented):
CleanShot 2024-06-25 at 10 19 05@2x

I still see a validation issue: here on the address content type I have validation errors on both DZs (one is required, one has a min of 2), but the form is only able to show one error at a time

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Initializer.tsx
```diff
@@ -46,11 +46,6 @@ const Initializer = ({ disabled, name, onClick }: InitializerProps) => {
...
-        <Typography textColor="danger600" variant="pi">
-          {field.error}
-        </Typography>
-      )}
     </>
```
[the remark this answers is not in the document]
alexandrebodin: the error is displayed in the parent component already there is a Field.Error now  it was being displayed twice before the fix

packages/core/content-type-builder/admin/src/translations/en.json
```diff
@@ -64,7 +64,7 @@
   "error.validation.enum-duplicate": "Duplicate values are not allowed (only alphanumeric characters are taken into account).",
   "error.validation.enum-empty-string": "Empty strings are not allowed",
   "error.validation.enum-regex": "At least one value is invalid. Values should have at least one alphabetical character preceding the first occurence of a number.",
-  "error.validation.minSupMax": "Can't be superior",
+  "error.validation.minSupMax": "min can't be superior to max",
```
jhoward1994, line 67:
```suggestion
  "error.validation.minSupMax": "Min can't be superior to max",
```

Should this still be title case?

packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -670,45 +670,16 @@ const useField = <TValue = any,>(path: string): FieldValue<TValue | undefined> =
 
   const handleChange = useForm('useField', (state) => state.onChange);
 
-  const formatNestedErrorMessages = (stateErrors: FormErrors<FormValues>) => {
```
jhoward1994, line 673: I don't think we can remove this logic. The problem is for cases like this:


```
 state.errors =
	{
	    "Compo.Nested.Unique": "This attribute must be unique"
	}
```

When you just do `const error = getIn(state.errors, path);` we can't extract and highlight these nested errors correctly

examples/getstarted/src/api/address/content-types/address/schema.json
remidej, line 1: I think in the end we should actually remove those changes because the validations are very hard to fulfill, it makes the content type quite hard to use

remidej (approved):
QA wise looks good 👍

PR #20617, opened 2024-06-25
Fix Vite HMR Server Restart 

### LINKED ISSUE #20220: [BUG] Strapi loses the connection to the server after saving changes on entities

- Environment: development
- OS: linux-x64
- Strapi Version: 5.0.0-beta.5
- Node/Yarn Version: yarn/4.1.1 npm/? node/v20.12.2 linux x64
- Edition: Community
- Database: sqlite
- Typescript

### Describe the bug
Every time we save changes on any entity, either creating or updating it, the admin panel disconnects from the backend and we have to manually refresh the browser to get the admin back.

1. Go to the **Content Manager**.
2. Click on the Collection Type you will work with.
3. Create/Update any entity from the Collection Type.
4. Fill the fields.
5. Save the changes.
6. Wait until the animation is finished.
7. See the error.
8. Click on Try Again.
9. See the admin back.

### Expected behavior
It should refresh itself instead of losing connection to the backend.

### Screenshots
StrapiDoesNotRefreshAfterSavingEntities

### Code snippets

### Additional context
It is a basic project, with not third-party plugins on it.

Convly:
### What does it do?

The HMR configuration in config.ts has been tweaked; now the server property is set to the instance of the HTTP server that Strapi uses, so that Vite HMR server and Strapi start accepting requests at the same time.

### Why is it needed?

The Vite server was restarting faster than the Strapi HTTP server, hence triggering vite to reload the page before Strapi is ready to accept requests (hence the "Unable to connect" message).

### How to test it?

- Got to the CTB
- Modify a content-type
- The server shouldn't display the "Unable to connect" page and instead only reload when the server is ready

### Related issue(s)/PR(s)

fix #20220

### DIFF
--- a/examples/getstarted/src/admin/app.jsx
+++ b/examples/getstarted/src/admin/app.jsx
@@ -5,8 +5,6 @@ const config = {
   locales: ['it', 'es', 'en'],
 };
 const bootstrap = (app) => {
-  console.log('I AM  BOOTSTRAPPED');
-
   app.getPlugin('content-manager').injectComponent('editView', 'right-links', {
     name: 'PreviewButton',
     Component: () => (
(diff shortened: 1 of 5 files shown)


alexandrebodin (approved):
code looks good, not more browser shutdowns is nice

Issue #20616, opened 2024-06-25
Content History: Crash when viewing versions with added media or relation fields


simotae14:
### Required System information

- Node.js version: v20.10
- NPM version: v10.2.3
- Strapi version: 5.0.0-beta.16
- Database: sqlite
- Operating system: macOS Monterey v12.6.3
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When you open the History of a collection where you added some new fields (media field or relation field) after the creation of an entry and click on the versions before the new fields added and after you receive an error

### Steps to reproduce the behavior

1. Create a new collection with some fields on it
2. then create an entry of this collection and save
3. go back to the Content Type Builder and add some new fields to the collection (one image or one relation or both)
4. Update the entry previously created with some values for the new fields and save
5. Go to the Content History of the Document and click the different version from one without the new fields and one with the new fields
6. See error

### Expected behavior

I don't want to see errors if I switch between versions

### Screenshots

https://github.com/strapi/strapi/assets/2589748/18c7b169-319b-4352-8f15-5cf470b53b34

PR #20619, opened 2024-06-25
Fix Content History: Crash when viewing versions with added media or relation fields

simotae14:
### What does it do?

It fixes the error that we had when you opened the History of a collection where you added some new fields (media field or relation field) after the creation of an entry and click on the versions before the new fields were added and after you received an error

### Why is it needed?

To avoid the app crashing if you switch between the versions

### How to test it?

1. Create a new collection with some fields on it
2. then create an entry of this collection and save
3. go back to the Content Type Builder and add some new fields to the collection (one image or one relation or both)
4. Update the entry previously created with some values for the new fields and save
5. Go to the Content History of the Document and click the different version from one without the new fields and one with the new fields
6. You don't have to see the error

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20616


packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -168,9 +168,9 @@ const CustomRelationInput = (props: RelationsFieldProps) => {
  * -----------------------------------------------------------------------------------------------*/
 
 const CustomMediaInput = (props: VersionInputRendererProps) => {
-  const {
-    value: { results, meta },
-  } = useField(props.name);
+  const { value } = useField(props.name);
+  const results = value ? value.results : [];
+  const meta = value ? value.meta : { missingCount: 0 };
```
remidej, line 173: nice, should we maybe do the same for CustomRelationInput to be safe?

simotae14: yeah, I tried to replicate the error for the relations and there is not error but for sure we can add a preventive check

remidej (approved):
tested 👍

PR #20622, opened 2024-06-25
Experimental internal template

Mcastres:
### What does it do?

This PR includes an `example-app` template that behaves the same way as a template except that the app, is living in the generators folder. When you create a project with the CLI, users will have the choice to start with a project containing data or not.

Right now, I try to not change a lot of things by attaching the new option to the template option (merging function) but it might not be the best way. I'm opening this Draft pull Request to received feedback on the implementation and make sure I'm moving in the right direction.

### Why is it needed?

We might give a better user experience by delivering non-empty projects starting with Strapi. This is an experiment and might not be permanent.

### How to test it?

```
cd /packages/cli/create-strapi-app 
node bin/index.js /tmp/my-project
```


packages/generators/app/src/utils/merge-template.ts
```diff
@@ -38,11 +39,16 @@ export default async function mergeTemplate(scope: Scope, rootPath: string) {
     scope.template?.startsWith(filePrefix)
   );
 
+  // Force templatePath to point to the internal example template
+  if (scope.useExampleApp) {
+    templatePath = path.resolve(__dirname, '..', 'resources', 'templates', 'example-app', scope.useTypescript ? 'ts' : 'js');
```
alexandrebodin, line 44: You don't have the ts version yet from what I can see.

alexandrebodin (commented):
considering the changes we will make to the template system it looks fine to do this experiment for a few weeks

packages/generators/app/src/utils/merge-template.ts
Mcastres: Yes, right now the PR is not over, I'm just looking for feedback in terms of the way I implemented the system, not the template in itself :)

alexandrebodin: Yes, not much to say it's ok considering we will move everything somewhere else soonish

Mcastres:
It successfully build on my side 🤔 
Wondering why it doesn't pass the test

alexandrebodin:
> It successfully builds on my side 🤔 Wondering why it doesn't pass the test

it's try to build the template file and it shouldn't, make sure to add them in the ignore of the tsconfig like the other resource files

Mcastres:
All good then!

alexandrebodin:
@Mcastres Can you squash all your commits into a single one please :)

alexandrebodin (changes requested):
For now I only test the ts version. It works fine. 

A few notes before I review it again: 

- In the media library I see 11 times the coffee-art image  & 5 times the cofee-beans. I think you can avoid this and only upload them once and then link to them in the conten types. 

I can create an experimental for you to test it in real life if you want. just ping me in DM

alexandrebodin:
Closing in favor of [the ](https://github.com/strapi/strapi/pull/20913)

PR #20626, opened 2024-06-25
fix: main nav fixes

remidej:
### What does it do?

Fixes on the main nav:

- empty space at the bottom
- logo not vertically aligned
- show tooltips instantly
- fix a defaultMessage not matching the value in the translation files

### Why is it needed?

Polishing for RC


simotae14 (approved):
👍🏽

Issue #20490, opened 2024-06-11
Duplicating entries allows records to bypass uniqueness checks


Nniol:
### Required System information

- Node.js version: 18.12.1
- NPM version: 8.19.2
- Strapi version: 4.22.1
- Database: Postgres
- Operating system: Dev: Windows, iOS  Prod: Linux
- Is your project Javascript or Typescript: JavaScript

### Describe the bug

We have created a schema which is unique fields set to true for two fields. Is we choose Duplicate Item Line x then we get the warning message saying we have unique checks in place for some fields.
However if we do nothing and just click Save then the record saves and can be published bypassing any uniqueness checks hence causing issues in the solution.

### Steps to reproduce the behavior

1. Create an pickem_group , save and publish
2. Then choose Duplicate Item on line 2 (Picture 1)
3. You get a warning message, good (Picture 2)
4. Click Create
5. Click Save 
6. Go to list view and see duplicate values for both pickemGroupId and name which are unique according to the schema (Picture 3)
7. Click back on item
8. Click Publish
9. Still no warning or stop

### Expected behaviour

At step 5 a warning should appear preventing the creation of the record

### Screenshots
Picture 1
![image](https://github.com/strapi/strapi/assets/1842296/8ababbbd-903e-4297-bdba-6febd89c04e1)

Picture 2
![image](https://github.com/strapi/strapi/assets/1842296/ab02b78b-3324-48d6-8b6c-ce2267de1dbe)

Picture 3
![image](https://github.com/strapi/strapi/assets/1842296/e4895243-5b65-4b85-b00b-eb1766816176)

### Code snippets

pickem_group schema:

{
  "kind": "collectionType",
  "collectionName": "pickem_groups",
  "info": {
    "singularName": "pickem-group",
    "pluralName": "pickem-groups",
    "displayName": "Pickem Group",
    "description": ""
  },
  "options": {
    "draftAndPublish": true
  },
  "attributes": {
    "pickemGroupId": {
      "type": "string",
      "required": true,
      "unique": true,
      "minLength": 5,
      "regex": "^[a-zA-Z0-9-]+$"
    },
    "name": {
      "type": "string",
      "unique": true,
      "required": true,
      "minLength": 5
    },
    "pickemOpens": {
      "type": "datetime",
      "required": true
    },
    "pickemCloses": {
      "type": "datetime",
      "required": true
    },
    "event": {
      "type": "relation",
      "relation": "manyToOne",
      "target": "api::event.event",
[...]

Issue #17656, opened 2023-08-12
Unique fields validation does not work when saving a cloned entry


itsmohmans:
### Required System information

- Node.js version: `v18.16.1`
- NPM version: `v9.5.1`
- **Yarn** version: `1.22.19`
- Strapi version: `4.12.0`
- Database: sqlite
- Operating system: Manjaro Linux
- Is your project Javascript or Typescript: JS

### Describe the bug

I have a collection with 3 unique values, [here's the schema](https://github.com/Marje3PSUT/psutarchive-api/blob/dev/src/api/course/content-types/course/schema.json)
```json
"attributes": {
    "name": {
      "pluginOptions": {},
      "type": "string",
      "required": true,
      "unique": true
    },
    "course_id": {
      "pluginOptions": {},
      "type": "string",
      "required": true,
      "unique": true
    },
    "name_ar": {
      "type": "string",
      "required": true,
      "unique": true
    }
}
```

Creating a new item and attempting to save it with any duplicate value for one of these unique attributes works as expected, it returns a validation error and prevents me from saving the item.
The problem appears when I attempt to clone any item, it shows the expected warning
```
Warning:
Entity could not be cloned as it has unique and/or relational fields.
Please edit those fields manually and save to complete the cloning.
```
but still allows me to save it without validation and end up with duplicate values.

### Steps to reproduce the behavior
Here's the repo to reproduce (`dev` branch) https://github.com/itsmohmans/psutarchive-api/tree/dev

1. Run the project with yarn
2. Go to content manager and create a new course, fill in `name` and `course_id` (both are unique)
3. Save, then go back to courses list and click on `duplicate item` button
4. See the expected error 
![image](https://github.com/strapi/strapi/assets/55551984/5ba85504-fbb5-40dd-ae0d-463963b4a425)
Now ignore this warning, don't edit any field and click on Save, it'll save successfully, and you'll end up with duplicate values that are supposed to be unique.

### Expected behavior

It should validate before saving a cloned item and prevents from saving it if any of the unique fields is still the same as the one we cloned from, the same way it validates when creating and saving a new item.

### Screenshots

Cloned items with the same _unique_ values, either draft or published
![image](https://github.com/strapi/strapi/assets/55551984/881db7db-bac0-453f-bd6c-31e85faf53eb)

### Code snippets

If applicable, add code samples to help explain your problem.

prokopsimek:
[...]

PR #20629, opened 2024-06-25
[Fix] Auto clone modal and ensure we validate correct statuses

jhoward1994:
### What does it do?

- Makes it so we can successfully navigate to create a manual new entry when auto cloning fails
- Modifies backend validator logic to only treat content types with draft and publish enabled as drafts during cloning

### Why is it needed?

Bug fixes

### How to test it?

### Related issue(s)/PR(s)

DX-1478

fixes https://github.com/strapi/strapi/issues/20141
fixes https://github.com/strapi/strapi/issues/20490
fixes https://github.com/strapi/strapi/issues/17656


packages/core/core/src/services/document-service/entries.ts
```diff
@@ -26,7 +26,7 @@ const createEntriesService = (uid: UID.ContentType) => {
 
     const validData = await entityValidator.validateEntityCreation(contentType, data, {
       // Note: publishedAt value will always be set when DP is disabled
-      isDraft: !params?.data?.publishedAt,
+      isDraft: params.status ? params.status === 'draft' : !params?.data?.publishedAt,
```
Marc-Roig, line 29: is there a situation where the status is draft and publishedAt is set? If that's the case, I feel we should fix that instead of adding this double check here 🤔

Marc-Roig: publishedAt should be enough to know if we are dealing with a draft or not imo

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -85,7 +85,7 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
 
     async create(uid: UID.CollectionType, opts: DocServiceParams<'create'> = {} as any) {
       const populate = opts.populate ?? (await buildDeepPopulate(uid));
-      const params = { ...opts, status: 'draft' as const, populate };
+      const params = { ...opts, status: opts?.status ?? ('draft' as const), populate };
```
Marc-Roig, line 88: Probably the same as :

```
const params = { ...opts, status: opts?.status, populate };
```

While this is valid, we are allowint to create a draft and publish it, which I don't think is what the document manager is intended to do atm.

This might have other security implications, e.g user with only create permissions (not publish) could modify the request to add "status": "publish" and publish a document

jhoward1994:
Thanks @Marc-Roig I got a bit confused there I think.

I though I saw this case during the auto clone call

_"is there a situation where the status is draft and publishedAt is set? If that's the case, I feel we should fix that instead of adding this double check here 🤔"_

But now having reverted the changes I cant replicate. I think the important change was in the validator the whole time

```
    const hasPathToComponent = componentContext?.pathToComponent?.length > 0;
    if (hasPathToComponent) {
```

So most of this has been reverted as the fix for this PR is working. Are you happy with the changes in

- packages/core/core/src/services/document-service/repository.ts
- packages/core/content-manager/server/src/controllers/collection-types.ts

To not always default to 'draft' during creation?

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -148,7 +148,12 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
         assoc('documentId', createDocumentId()),
         // Merge new data into it
         (data) => merge(data, queryParams.data),
-        (data) => entries.create({ ...queryParams, data, status: 'draft' })
+        (data) =>
+          entries.create({
+            ...queryParams,
+            data,
+            status: hasDraftAndPublish ? 'draft' : 'published',
+          })
```
Marc-Roig, line 156: is the status even necessary here? I believe entries.create totally ignores that attribute, so maybe we can just ignore it.

The clone will copy the publishedAt value of the original entry , and as all entries have that column (even if draft and publish is disabled) I believe it should work, wdyt?

Marc-Roig:
other than the last comment, everything else looks good to me :)

alexandrebodin:
Leaving that PR to both of you I don't have enough details on the expected behavior and the code looks fine from a high level 👍

packages/core/core/src/services/document-service/repository.ts
jhoward1994: Hey, sorry for getting to this slowly, I see what you mean.

I think the problem was in `packages/core/content-manager/server/src/services/document-manager.ts` here:
[`b356982` (#20629)](https://github.com/strapi/strapi/pull/20629/commits/b35698277f72b0594d0eb7b4f4799edce11592c2)

With `publishedAt` always being set to null this line (150) was always merging publishedAt as null into data, regardless of D&P being enabled
https://github.com/strapi/strapi/blob/b35698277f72b0594d0eb7b4f4799edce11592c2/packages/core/core/src/services/document-service/repository.ts#L150-L151

With that removed we can rely on the publlishedAt retrieved from the above query:
https://github.com/strapi/strapi/blob/b35698277f72b0594d0eb7b4f4799edce11592c2/packages/core/core/src/services/document-service/repository.ts#L132-L141

Marc-Roig: ah! 💡 that sounds like it

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -35,7 +35,10 @@ const createDocument = async (ctx: any, opts?: Options) => {
   const sanitizeFn = async.pipe(pickPermittedFields, setCreator as any);
   const sanitizedBody = await sanitizeFn(body);
 
-  const { locale, status = 'draft' } = await getDocumentLocaleAndStatus(body);
+  const {
+    locale,
+    status = contentTypes.hasDraftAndPublish(strapi.getModel(model)) ? 'draft' : 'published',
```
Marc-Roig, line 40: does that make sense that this goes into the getDocumentLocaleAndStatus? (Setting published or undefined by default if it doesnt have draft and publish)

Marc-Roig (approved):

packages/core/content-manager/server/src/controllers/collection-types.ts
Marc-Roig: As you see fit :) 
My point is also, all the other controllers should set this value to published by default if D&P is disabled (maybe I am wrong?), as this applies to all methods of the document service model. Also, an undefined status should also work in this situation. (Document Service defaults to draft when necessary) so maybe we dont even need the default value

PR #20630, opened 2024-06-25
fix: split single & bulk delete

alexandrebodin:
### What does it do?

Separates concerns between bulk & single webhook delete. the UX was pretty strange (when clicking on the trash icon it opens a confirm but also selects in the bulk delete in the background)

### DIFF
--- a/packages/core/admin/admin/src/components/ConfirmDialog.tsx
+++ b/packages/core/admin/admin/src/components/ConfirmDialog.tsx
@@ -93,7 +93,13 @@ const ConfirmDialog = ({
       <Dialog.Footer>
         {startAction || (
           <Dialog.Cancel>
-            <Button fullWidth variant="tertiary">
+            <Button
+              fullWidth
+              variant="tertiary"
+              onClick={(e) => {
+                e.stopPropagation();
+              }}
+            >
               {formatMessage({
                 id: 'app.components.Button.cancel',
                 defaultMessage: 'Cancel',
--- a/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/ListPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/ListPage.tsx
@@ -103,7 +103,32 @@ const ListPage = () => {
     }
   };
 
-  const confirmDelete = async () => {
+  const deleteWebhook = async (id: string) => {
+    try {
+      const res = await deleteManyWebhooks({
+        ids: [id],
+      });
+
+      if ('error' in res) {
+        toggleNotification({
+          type: 'danger',
+          message: formatAPIError(res.error),
+        });
+
+        return;
+      }
+    } catch {
+      toggleNotification({
+        type: 'danger',
+        message: formatMessage({
+          id: 'notification.error',
+          defaultMessage: 'An error occurred',
+        }),
+      });
+    }
+  };
+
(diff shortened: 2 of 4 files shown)


jhoward1994:
It looks like we need to do this at https://github.com/strapi/strapi/blob/2d79c58816fc439db3b9eb9ae4f2bdbb6848d3f5/packages/core/admin/admin/src/components/ConfirmDialog.tsx#L95-L96

(added a click handler with stopPropagation )

```
            <Button
              fullWidth
              variant="tertiary"
              onClick={(e) => {
                e.stopPropagation();
              }}
            >
```

Otherwise cancelling the single delete modal still navigates to the webhook edit page

remidej (commented):
Besides what Jamie mentioned I just found one way to kind of break it:

- select a webhook
- delete it with the individual delete button
- the selection is not cleared even though the webhook doesn't exist anymore

https://github.com/strapi/strapi/assets/8087692/a5f7d1c8-d92d-4510-b32d-f407f3864805

alexandrebodin:
@

> Besides what Jamie mentioned I just found one way to kind of break it:
> 
> * select a webhook
> * delete it with the individual delete button
> * the selection is not cleared even though the webhook doesn't exist anymore
> 
>  CleanShot.2024-06-26.at.10.49.40.mp4

Yeah I have noticed but considered it was not a pbl as it won't fail anywhere for now.

remidej:
@alexandrebodin I agree it's not very problematic, but wouldn't it be fixed by writing `setWebhooksToDelete([])` in deleteWebhook?

alexandrebodin:
> @alexandrebodin I agree it's not very problematic, but wouldn't it be fixed by writing `setWebhooksToDelete([])` in deleteWebhook?

I would only remove the id we just deleted from the list in case you started select multiple things and then deleted a single one. Feel free to push to the PR I have to finish sth a bit more urgent for the RC. We can always merge this another time

remidej:
I pushed, LGTM now

packages/core/admin/admin/src/components/ConfirmDialog.tsx
```diff
@@ -93,7 +93,13 @@ const ConfirmDialog = ({
       <Dialog.Footer>
         {startAction || (
           <Dialog.Cancel>
-            <Button fullWidth variant="tertiary">
+            <Button
```
alexandrebodin, line 96: is this fine to do it for all the confirms ?

jhoward1994:
I just checked every use of <ConfirmDialog> and everything was working with this change 

[`9983fd4` (#20630)](https://github.com/strapi/strapi/pull/20630/commits/9983fd4c2a0871330913ed75696f3c610bee4ec3)

I spotted a couple of other things but can track them separately:
- Cannot close bulk CM list view publish confirm modal (it uses a different custom modal)
- Upload bulk select doesn't work

wdyt?

alexandrebodin:
> I just checked every use of and everything was working with this change
> 
> [`9983fd4` (#20630)](https://github.com/strapi/strapi/pull/20630/commits/9983fd4c2a0871330913ed75696f3c610bee4ec3)
> 
> I spotted a couple of other things but can track them separately:
> 
> * Cannot close bulk CM list view publish confirm modal (it uses a different custom modal)
> * Upload bulk select doesn't work
> 
> wdyt?

Sounds good I'll merge this one and you can create gh issues for the others 👍

PR #20633, opened 2024-06-26
fix: drop slug fields index

Marc-Roig:
### What does it do?

Adds a migration to drop the unique index of slug fields, those are automatically removed by the strapi schema diff, but that diff is executed after the migrations, and  `discard-draft` migration will fail if there are such indexes.

### DIFF
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-05-drop-slug-unique-index.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-05-drop-slug-unique-index.ts
@@ -9,10 +9,12 @@ import type { Knex } from 'knex';
 
 import type { Migration } from '../common';
 
-const dropIndex = async (knex: Knex, tableName: string, columName: string) => {
+const dropIndex = async (knex: Knex, tableName: string, columnName: string) => {
   try {
     await knex.schema.alterTable(tableName, (table) => {
-      table.dropUnique([columName]);
+      // NOTE: Can not use "identifiers" utility, as the 5.0.0-01 migration does not rename this particular index
+      // to `tableName_columnName_uq`.
+      table.dropUnique([columnName], `${tableName}_${columnName}_unique`);
     });
   } catch (error) {
     // If unique index does not exist, do nothing
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-05-drop-slug-unique-index.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-05-drop-slug-unique-index.ts
@@ -1,0 +1,41 @@
+/**
+ * In V4 slug fields contained a unique index.
+ * In V5 slug fields should not have a unique index.
+ *
+ * This migration drops existing unique indexes from slug fields so downstream migrations
+ * can work on the data without violating the unique index.
+ */
+import type { Knex } from 'knex';
+
+import type { Migration } from '../common';
+
+const dropIndex = async (knex: Knex, tableName: string, columName: string) => {
+  try {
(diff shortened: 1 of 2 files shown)


packages/core/database/src/migrations/internal-migrations/5.0.0-05-drop-slug-unique-index.ts
```diff
@@ -0,0 +1,41 @@
...
+  async up(knex, db) {
+    for (const meta of db.metadata.values()) {
+      const hasTable = await knex.schema.hasTable(meta.tableName);
+      if (!hasTable) {
+        continue;
+      }
+
+      for (const attribute of Object.values(meta.attributes)) {
+        if (attribute.type === 'uid' && attribute.columnName) {
+          await dropIndex(knex, meta.tableName, attribute.columnName);
```
alexandrebodin, line 33: we don't used a prefix or suffix for the index name ?

Marc-Roig: the table.dropUnique(columnName) removes any unique index tied to that column

Convly: Does it work with long DB identifiers that are transformed?

Marc-Roig: I assumed (maybe wrongly) the columnName  was the database column name, will check thats the case 🤔

Marc-Roig: I believe `column_name` should be the shortened identifier, and the `convert-identifiers-long-than-max-lenght` should run before this PR migration, so that should be good

alexandrebodin: dropUnique takes a second argument in case the indexName is not the default. Do we actually use the default name 🤔

Marc-Roig: hurm , we do not use the default name, but let me check

Marc-Roig: image
Unless this is the default name , I assume we are not using one

but still the migration is running ok 🤔 I will just use the index name just in case

Marc-Roig: looks like knex index unique key matches ours (tableName_columnName_unique)
I created a commit making the index name explicit

alexandrebodin: better safe than sorry 👍

alexandrebodin (approved):
didn't do QA on it but LGTM

kaue-delmonte-acn:
Hi everyone,

I'm encountering an error while starting my Strapi server after migrating from version 5.0.0-beta2 to 5.0.1. The error message is as follows:


`MigrationError: Migration 5.0.0-05-drop-slug-fields-index (up) failed: Original error: select * from information_schema.tables where table_name = $1 and table_schema = current_schema() - current transaction is aborted, commands ignored until end of transaction block`

Could someone assist me with this issue? @Marc-Roig 

Thanks in advance!

Marc-Roig:
@kaue-delmonte-acn could you open an additional issue tocket for your matter?
Ping me in the issue and I can provide some help :) 

Was this a project migrated from v4?

PR #20636, opened 2024-06-26
fix: components count in ctb menu

remidej:
### What does it do?

- The CTB menu counted the amount of folders, not of components. Fixes it
- Fixes a defaultMessage that didn't match the translation file values (hence the snapshot update)

### How to test it?

Open the CTB, make sure the count of components, single types and collection types is accurate. Use the search to filter the result, the counts should be updated.

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/ContentTypeBuilderNav.tsx
+++ b/packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/ContentTypeBuilderNav.tsx
@@ -23,22 +23,19 @@ export const ContentTypeBuilderNav = () => {
   const { menu, searchValue, onSearchChange } = useContentTypeBuilderMenu();
   const { formatMessage } = useIntl();
 
+  const pluginName = formatMessage({
+    id: getTrad('plugin.name'),
+    defaultMessage: 'Content-Type Builder',
+  });
+
   return (
-    <SubNav
-      aria-label={formatMessage({
-        id: `${getTrad('plugin.name')}`,
-        defaultMessage: 'Content-Types Builder',
-      })}
-    >
+    <SubNav aria-label={pluginName}>
       <SubNavHeader
         searchable
         value={searchValue}
         onClear={() => onSearchChange('')}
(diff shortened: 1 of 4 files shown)


lucasboilly:
It works perfectly on my side 👌

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/useContentTypeBuilderMenu.ts
```diff
@@ -185,20 +187,26 @@ export const useContentTypeBuilderMenu = () => {
...
               ...link,
               links: filteredLinks.sort((a: any, b: any) => formatter.compare(a.title, b.title)),
             };
           })
           .filter(Boolean),
+        linksCount: filteredLinksCount,
       };
     }
 
+    const fileteredLinks = section.links
```
f1shear, line 202: `filteredLinks`

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/ContentTypeBuilderNav.tsx
```diff
@@ -53,7 +50,7 @@ export const ContentTypeBuilderNav = () => {
                 defaultMessage: section.title.defaultMessage,
               })}
               collapsable
-              badgeLabel={section.links.length.toString()}
+              badgeLabel={section.linksCount.toString()}
             >
```
jhoward1994, line 54: Completely up to you but we could remove the hook changes and do this. I think
it achieves the same thing?

```suggestion
              badgeLabel={section.links
                .reduce((acc, link) => acc + (link.links ? link.links.length : 1), 0)
                .toString()}
```

remidej: Yes it does the same so a matter of personal preference. I think I prefer extending the hook like I did because there was already a `if (hasChild) {` condition, so I can put the logic in the same place

jhoward1994 (approved):

PR #20640, opened 2024-06-26
fix: history action position in menu

remidej:
### What does it do?

Fixes the position of the history document action in the edit view header. It used to show up last, but the designs specify it has to show up in the middle, right before the "danger" actions.

I'm doing it by using the public addDocumentAction, instead of setting it in the default array. It's the [same approach as done in i18n](https://github.com/strapi/strapi/blob/93fad71560acac3452f5fbd9b63fe0490e9eaedd/packages/plugins/i18n/admin/src/index.ts#L72-L76) to control the action's position in the array.


Marc-Roig:
looks like there is a build error in strapi/strapi, but maybe it's not related to this pr? 🤔

remidej:
@Marc-Roig nice spot, it actually looks like a type error got fixed and that caused the error because of ts-expect-error

simotae14 (approved):
looks good and it works, 👏🏽

PR #20646, opened 2024-06-27
fix(content-manager): dismiss bulk list view publish

jhoward1994:
### Why is it needed?

Clicking cancel at the confirmation stage of a bulk list view publish does nothing.
The modal is being toggled multiple times and therefore staying visible

### Related issue(s)/PR(s)

DX-1482


Feranchz (approved):
Thanks! There is one UI error with the Bulk Publish modal that I will take care on other PR 🚀

PR #16768, opened 2023-05-18
feature: onCommit , onRollback transaction methods

Marc-Roig:
## What does it do?

This is another iteration to improve transaction handling on async operations like audit logs.
It is a follow up of https://github.com/strapi/strapi/pull/16341.

During the development of RW we have seen two issues regarding how Audit Logs & Transactions interact with each other. And I will try to explain our setup and the problems we have seen.

We were doing something like this:
```js
    await strapi.db.transaction(async (trx) => {
	  const stages = await createWorkflowStages()//  -> Stores in DB new stages & Triggers audit log
      await createWorkflow(stages); // -> Stores in DB new workflow & Triggers audit log
    });
```

### Storing audit log event when data is rollbacked
In some cases, `createWorkflow` can throw an error and so, stages data is rollbacked. But... Even if we rollback the data, the audit log event is stored in DB.

### Saving events on committed transactions
In some cases the execution of the workflow logic and audit logs was like so:
```
RW ====> createsWorkflow ...................commit................................
AL  ====>  .......................savesEvent.................DB Access & commit 
```
With  https://github.com/strapi/strapi/pull/16341 seemed like we had fixed this, but it was just luck and you can see this error happening now, if you go to Settings -> Review Workflows and save some changes.

### Proposed solution
Save audit log events when the transaction is actually commited
```js
      await strapi.db.transaction(({ onCommit }) => {
        onCommit(() => state.provider.saveEvent(processedEvent));
      });
```

This way if data is rollbacked the event will not be stored.

PR #20647, opened 2024-06-27
fix: duplicated audit log events

Marc-Roig:
### What does it do?

Publishing an entry resulted in duplicated "entry.update" logs in the audit logs settings page. 

The cause is a bug in how we stored events in audit logs:

```ts
await strapi.db.transaction(({ onCommit }) => {
        onCommit(() => auditLogsService.saveEvent(processedEvent));
      });
```
See the explanation of why this was introduced in https://github.com/strapi/strapi/pull/16768

This bit of code is not necessary anymore, as the document service emits the events using this same api.
**To dig more: For some reason, nested onCommit calls (a callback inside an oncommit opening another transaction with an onCommit call inside) caused the callbacks to be called twice**

### DIFF
--- a/packages/core/admin/ee/server/src/audit-logs/services/lifecycles.ts
+++ b/packages/core/admin/ee/server/src/audit-logs/services/lifecycles.ts
@@ -112,11 +112,7 @@ const createAuditLogsLifecycleService = (strapi: Core.Strapi) => {
     const processedEvent = processEvent(name, ...args);
 
     if (processedEvent) {
-      // This stores the event when after the transaction is committed,
-      // so it's not stored if the transaction is rolled back
-      await strapi.db.transaction(({ onCommit }) => {
-        onCommit(() => auditLogsService.saveEvent(processedEvent));
-      });
+      await auditLogsService.saveEvent(processedEvent);
     }
   };
 


Marc-Roig:
@alexandrebodin Instead of relying on such api (onCommit) , do you think we could extend the current strapi.db.transaction api, to not use any shared transaction, but rather create a new one?

A couple of times, we have seen async jobs hitting already closed transactions just because they started using the transaction in the Async Context. 
Forcing that async job to take a new transaction would make the onCommit api not necessary, and probably would be more readable.

alexandrebodin:
> @alexandrebodin Instead of relying on such api (onCommit) , do you think we could extend the current strapi.db.transaction api, to not use any shared transaction, but rather create a new one?
> 
> A couple of times, we have seen async jobs hitting already closed transactions just because they started using the transaction in the Async Context. Forcing that async job to take a new transaction would make the onCommit api not necessary, and probably would be more readable.

Yes I would expect that fn to create a new transaction everytime

alexandrebodin (approved):

Marc-Roig:
We do share the transaction if we nest them, which I think makes sense, but for this kind of situations creating a new one would be ideal imo

alexandrebodin:
> We do share the transaction if we nest them, which I think makes sense, but for this kind of situations creating a new one would be ideal imo

Yeah nesting strapi.db.transaction calls should result in just using the same transaction. We could also consider logging a warning saying that's unecessary and could be unwanted 🤔

Marc-Roig:
Alright, lets wait to get that idea cooked in our minds and I will merge this in the meantime 🚋

PR #20648, opened 2024-06-27
feat(release-settings): apply rbac permission in the release settings page

simotae14:
### What does it do?

Apply the Read and Update permissions in the Release Settings page

Here the different scenarios:
- NO READ NO UPDATE : we show the No Permissions body content in the page
- READ NO UPDATE: we show the Release Settings Page with the Save button not shown and the Default timezone combobox disabled
- READ UPDATE: we show the normal Release Settings page with all the functionalities enabled

### Why is it needed?

To show the proper content to the different kind of users

### How to test it?

- Open the CMS with a EE license (if you want you can try to check if we show the Purchase page in the settings with CE license)
- As an admin create a new user with the Editor Role (by default the permissions are set to false)
- Go to the settings page, select the Editor Role and in the Settings Tab change the permissions for the Content Releases
- then open a new Tab in the browser and login as an Editor and check the Release Settings page with the different permissions selected

### Related issue(s)/PR(s)

CS-807

### DIFF
--- a/packages/core/admin/ee/admin/src/constants.ts
+++ b/packages/core/admin/ee/admin/src/constants.ts
@@ -21,9 +21,23 @@ export const ADMIN_PERMISSIONS_EE = {
       read: [{ action: 'admin::provider-login.read', subject: null }],
       update: [{ action: 'admin::provider-login.update', subject: null }],
     },
+    releases: {
+      read: [
+        {
+          action: 'plugin::content-releases.settings.read',
+          subject: null,
+        },
+      ],
+      update: [
+        {
+          action: 'plugin::content-releases.settings.update',
+          subject: null,
+        },
+      ],
+    },
   },
 } satisfies {
-  settings: Pick<PermissionMap['settings'], 'auditLogs' | 'review-workflows' | 'sso'>;
+  settings: Pick<PermissionMap['settings'], 'auditLogs' | 'review-workflows' | 'sso' | 'releases'>;
 };
 
 /**
--- a/packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
+++ b/packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
@@ -96,7 +96,7 @@ export const ReleasesSettingsPage = () => {
         <Form
           method="PUT"
           initialValues={{
-            defaultTimezone: data?.data.defaultTimezone || '',
+            defaultTimezone: data?.data.defaultTimezone,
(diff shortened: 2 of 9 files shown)


packages/core/admin/ee/admin/src/constants.ts
```diff
@@ -21,9 +21,23 @@ export const ADMIN_PERMISSIONS_EE = {
...
+        },
+      ],
+      update: [
+        {
+          action: 'plugin::content-releases.settings.update',
+          subject: null,
+        },
+      ],
+    },
   },
```
Marc-Roig, line 38: just curious, why do we need this set of constant permissions in the admin package ? 
PRobably same as for review workflows, is it not possible to define this permissions only inside the content-releases packaeg?

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
```diff
@@ -85,13 +95,9 @@ export const SettingsPage = () => {
...
           method="PUT"
-          initialValues={
-            data?.data.defaultTimezone
-              ? data.data
-              : {
-                  defaultTimezone: null,
-                }
-          }
+          initialValues={{
+            defaultTimezone: data?.data.defaultTimezone || '',
```
Marc-Roig, line 99: If the API defaultTimezone is null, and then we set the defaultTimezone form value to '', wouldn't we be overriding the value to be an empty string after saving the settings?

packages/core/admin/ee/admin/src/constants.ts
Marc-Roig: nvm, I just realized this is necessary to make use of the useTypedSelector api

Feranchz (approved):
QAed and works well!

packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+          ctx.json({
+            data: {
+              defaultTimezone: null,
+            },
+          })
+        )
+      )
+    );
+
+    render(<ReleasesSettingsPage />);
```
remidej, line 52: I don't think it's required to mock RBAC if you're rendering the non-protected component. I would recommend to always render the protected version since you already have the mock set up. You wouldn't even need to export ReleasesSettingsPage directly

remidej (approved):

Issue #20544, opened 2024-06-18
This changes seems to break previous cors configuration and I couldn't find documentation about how to upgrade to the new changes.


Pablo-Aldana:
This changes seems to break previous cors configuration and I couldn't find documentation about how to upgrade to the new changes.

Right now I have my CORS like this:

```
{
    name: 'strapi::cors',
    config: {
      //enabled: true,
      headers: '*',
      origin: env('CORS'),
    },
  },
  ```

And in the env(CORS) like this but with PROD values:
CORS=http://localhost:1338, http://localhost:8000, http://192.168.68.114:8000, http://192.168.68.114:3000, http://localhost:3000, http://192.168.68.114:1338

As soon as I update to this the strapi version 4.24.0 (https://github.com/strapi/strapi/releases/tag/v4.24.0)or higher, the web stop working due to cors issues, even if the URL is in the cors.

![Screenshot_2024-06-13 23 40 57_studUG](https://github.com/strapi/strapi/assets/9969839/6b1b817f-fa54-45e5-a4ec-66d7037bdb82)

Can anyone point me to the documentation or  update guide for this change?
I can create a new issue if needed

Many thanks

_Originally posted by @Pablo-Aldana in https://github.com/strapi/strapi/issues/20044#issuecomment-2166888703_

alexandrebodin:
Coping the answer from the PR comment here :)

With the cors fix we had to do for security we removed the string list support as we don't know if it makes sense to process it this way or not. to fix this for your usecase you should split the string yourself

env('CORS').split(', ')

alexandrebodin:
If we can support a string list again we will add it back when the issue is triaged & prioritize 🙏

Pablo-Aldana:
Hi @alexandrebodin  thanks a lot for your supper I will definitely try this. 

Not sure what would be the best approach, but I think the migration guide and the documents should get updated as from my side we got a massive issue in production because of this even tho every time we upgrade we check the migration and upgrade guide.

alexandrebodin:
cc @pwizla were do you think we could add some guidance after that security release ?

pwizla:
[...]

PR #20658, opened 2024-06-28
fix: support string array

alexandrebodin:
### What does it do?

Put string array support back into cors for compatbility 

Closes https://github.com/strapi/strapi/issues/20544


Convly (approved):

Convly:
Tested it and it seems to be working now :slightly_smiling_face:

PR #20659, opened 2024-06-28
Fix Koa Query Type

### LINKED ISSUE #20230: Types for ctx.query in custom controller appear to be broken

- Node.js version: v20
- Yarn version: v1
- Strapi version: 5.0.0-beta.5
- Database: SQLite

### Describe the bug

Attempting to play around with custom controllers in Strapi 5 I was attempting to set a custom `ctx.query` and noticed that type errors are being thrown around the structure of ctx.query

1. Create a custom route
2. Create a custom controller
3. Attempt to manually set the value of `ctx.query`
4. Request the custom controller
5. See error

### Expected behavior

No breaking changes were recorded for the object based populate syntax thus this was valid in Strapi 4 and should also be valid in Strapi 5

### Screenshots

image

image

### Code snippets

Route:

```js
module.exports = {
  routes: [
    {
      method: "GET",
      path: "/tests/custom/example",
      handler: "api::test.test.exampleAction",
    },
  ],
};
```

Controller:

```js
"use strict";

/**
 * test controller
 */

const { createCoreController } = require("@strapi/strapi").factories;

module.exports = createCoreController("api::test.test", ({ strapi }) => ({
  async exampleAction(ctx) {
    await this.validateQuery(ctx);
    const defaultQuery = {
      populate: {
        createdBy: {
          fields: ["id", "username"],
        },
        updatedBy: {
          fields: ["id", "username"],
        },
      },
    };

    if (!ctx.query) {
      ctx.query = defaultQuery;
    }

    const sanitizedQuery = await this.sanitizeQuery(ctx);
    const entities = await strapi
      .documents("api::test.test")
      .findMany(sanitizedQuery);

    const sanitizedData = await this.sanitizeOutput(entities, ctx);

    try {
      return this.transformResponse(sanitizedData);
    } catch (err) {
      return err;
    }
  },
}));
```

Schema:

```json
{
  "kind": "collectionType",
  "collectionName": "tests",
  "info": {
    "singularName": "test",
    "pluralName": "tests",
    "displayName": "test"
  },
  "options": {
    "comment": "",
    "populateCreatorFields": true
  },
  "attributes": {
    "test": {
      "type": "string"
    }
  }
}
```

### Additional context

N/A

Convly:
### What does it do?

Augment koa query type in `BaseRequest` and `BaseContext`

### Why is it needed?

We're modifying the default query type to be able to use `qs` instead of `node:querystring`. This change reflects that in the types.

### How to test it?

See #20230 for reproduction steps

### Related issue(s)/PR(s)

fixes #20230

### DIFF
--- a/packages/core/core/src/index.ts
+++ b/packages/core/core/src/index.ts
@@ -20,3 +20,25 @@ export const createStrapi = (options: Partial<StrapiOptions> = {}): Core.Strapi 
 
   return strapi;
 };
+
+// Augment Koa query type based on Strapi query middleware
+
+import * as qs from 'qs';
+
+declare module 'koa' {
+  type ParsedQuery = ReturnType<typeof qs.parse>;
+
+  export interface BaseRequest {
+    _querycache?: ParsedQuery;
+
+    get query(): ParsedQuery;
+    set query(obj: any);
+  }
+
+  export interface BaseContext {
+    _querycache?: ParsedQuery;
+
+    get query(): ParsedQuery;
+    set query(obj: any);
+  }
+}
--- a/packages/core/core/src/middlewares/query.ts
+++ b/packages/core/core/src/middlewares/query.ts
@@ -1,8 +1,8 @@
-import qs, { IParseOptions } from 'qs';
+import qs from 'qs';
 import type Koa from 'koa';
 import type { Core } from '@strapi/types';
 
-type Config = IParseOptions;
+type Config = Parameters<typeof qs.parse>[1];
 
 const defaults: Config = {
   strictNullHandling: true,
@@ -22,6 +22,7 @@ const addQsParser = (app: Koa, settings: Config) => {
      */
     get() {
       const qstr = this.querystring;
+
       this._querycache = this._querycache || {};
       const cache = this._querycache;
 
@@ -38,7 +39,7 @@ const addQsParser = (app: Koa, settings: Config) => {
     set(obj) {
       this.querystring = qs.stringify(obj);
     },
-  });
+  } satisfies PropertyDescriptor & ThisType<Koa.BaseRequest>);
 
   return app;
 };


packages/core/core/src/index.ts
```diff
@@ -20,3 +20,25 @@ export const createStrapi = (options: Partial<StrapiOptions> = {}): Core.Strapi
 
   return strapi;
 };
+
+// Augment Koa query type based on Strapi query middleware
+
+import * as qs from 'qs';
+
+declare module 'koa' {
```
alexandrebodin, line 28: can't we reuse the IParseOptions from qs ?

alexandrebodin (commented):
Just a general question

packages/core/core/src/index.ts
Convly: Where would we use that in this file? :thinking:

Convly: If this is related to the middlewares/query.ts change

`qs.parse` has 2 definitions, one that returns `{ [key: string]: unknown }` and the other returns `ParsedQs`.
The return type will be inferred based on the params passed to the function.
- If `IParseOptions & { decoder?: never | undefined }` is passed, then it'll resolve to `{ [key: string]: unknown }`
- If only `IParseOptions` is passed, then it'll resolve to `ParsedQs`

So here we had the choice of setting Config to `IParseOptions & { decoder?: never | undefined } | IParseOptions` or `Parameters<typeof qs.parse>[1]`. The first one is more atomic/readable, the other one is more maintainable as it'll automatically adapt to future signatures.

I'm OK with both :shrug:

Convly: It'll default to that through inference in some case such as here https://github.com/strapi/strapi/pull/20659/files#diff-3fcd641ac6072f7f3c6cde350394972f7998e6bc022c38e0edf320e81ef78e38R30

It's kinda needed since `{ [key: string]: unknown ? }` is wider than `ParsedQs`. All in all it's just to be covered, but it's minor

alexandrebodin (approved):

PR #20662, opened 2024-07-01
fix: do not populate morph relations on count draft relations

Marc-Roig:
### What does it do?

Morph relations were populated as regular relation on the count draft relations, it throws an error in v5.
We do not officially support morph relations, so when the time comes we can correctly populate the morph relation if necessary.

Another similar PR was opened, but we missed this spot.


remidej (approved):
Even less error alerts ❤️

PR #20664, opened 2024-07-01
fix(content-manager): bulk actions modal don't force to use modal

Feranchz:
### What does it do?

Fix the UI error on Bulk Publish modal. Bulk Publish API shouldn't force user to always wrap render in `Modal.Body` and `Modal.Footer` so users can build custom modals as they want (Bulk publish use case).

We can discuss later if we want to have a different API to add context when adding a Bulk Action modal. 

### How to test it?

Check all the modals from Bulk Actions (currently we have Bulk Publish, Add To Release, Publish multiple locales)


remidej (approved):
works well 👍

simotae14 (approved):

PR #20666, opened 2024-07-02
feat(releases-settings): use the default timezone in the Release modal

simotae14:
### What does it do?

It uses the Default Timezone saved in the Release Settings page as initial value in the Release Modal (when we create and edit a Release)

### Why is it needed?

To use a common default Timezone set by the admin, it is still possible to change the Release timezone for every release

### How to test it?

1. First of all as an Admin go to the Release Settings page and change the default timezone
2. then go to the Release page and try to create a Release, the new default timezone should be applied in the dropdown
3. Create a new Release without Scheduling and, after save, open it in Edit mode and enable Scheduling, the initial value of the timezone dropdown should be the default timezone

### Related issue(s)/PR(s)

CS-806


packages/core/content-releases/admin/src/pages/ReleasesPage.tsx
```diff
@@ -220,7 +222,7 @@ const ReleasesPage = () => {
     setReleaseModalShown((prev) => !prev);
   };
 
-  if (isLoading) {
+  if (isLoading || isLoadingSettings) {
```
remidej, line 225: I recommend renaming isLoading to something more specific now that there are several things loading

remidej (approved):

Feranchz (approved):

PR #20667, opened 2024-07-02
fix(e2e): dont expect document actions to be disabled because of releases

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/tests/e2e/tests/content-manager/editview.spec.ts
+++ b/tests/e2e/tests/content-manager/editview.spec.ts
@@ -129,7 +129,6 @@ test.describe('Edit View', () => {
       );
       await expect(page.getByRole('button', { name: 'Save' })).toBeDisabled();
       await expect(page.getByRole('button', { name: 'Publish' })).toBeDisabled();
-      await expect(page.getByRole('button', { name: 'More document actions' })).toBeDisabled();
 
       await page.getByRole('tab', { name: 'Draft' }).click();
       await expect(page.getByRole('button', { name: 'Save' })).toBeDisabled();


jhoward1994:
hey @simotae14 

wdyt of this? is it expected that "More document actions" is enabled at all of these points because of the releases action?

simotae14:
Yeah I think so, otherwise you don't have the possibility to add the document to the release.

tests/e2e/tests/content-manager/editview.spec.ts
```diff
@@ -129,7 +129,6 @@ test.describe('Edit View', () => {
       );
       await expect(page.getByRole('button', { name: 'Save' })).toBeDisabled();
       await expect(page.getByRole('button', { name: 'Publish' })).toBeDisabled();
-      await expect(page.getByRole('button', { name: 'More document actions' })).toBeDisabled();
```
simotae14, line 132: Why removing the assertion instead of checking if the button is enabled and just add a comment to explain why this one is enabled

jhoward1994: imo I don't see the point of checking that it's enabled, I think it makes more sense to keep the set of expectations consistent

but I'm still wondering should the releases button be enabled when we're on the published tab?

simotae14: yeah you can unpublish stuff with release

jhoward1994: @simotae14 I see thanks. To me removing these expectations makes sense then. Could you approve if you're okay with the change?

simotae14 (approved):

PR #20668, opened 2024-07-02
fix: inconsistent publish permissions

Marc-Roig:
### What does it do?

There is an inconsistency on the content manager publish / update permisions.

Scenario:
- User has permissions to publish and not update
- Expected behaviour: User can publish but not update draft
- Current behaviour: User can not publish

And also, there was a workaround about this, bulk publishing multiple locales.
See more [CS-838](https://strapi-inc.atlassian.net/browse/CS-838)

This is only the first step to fix the inconsistency, some FE work will be required (enabling the publish button on this scenario)

[CS-838]: https://strapi-inc.atlassian.net/browse/CS-838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -666,8 +666,6 @@ const PublishAction: DocumentActionComponent = ({
      *  - the document is already published & not modified
      *  - the document is being created & not modified
      *  - the user doesn't have the permission to publish
-     *  - the user doesn't have the permission to create a new document
-     *  - the user doesn't have the permission to update the document
      */
     disabled:
       isCloning ||
@@ -676,8 +674,7 @@ const PublishAction: DocumentActionComponent = ({
       activeTab === 'published' ||
       (!modified && isDocumentPublished) ||
       (!modified && !document?.documentId) ||
-      !canPublish ||
-      Boolean((!document?.documentId && !canCreate) || (document?.documentId && !canUpdate)),
+      !canPublish,
     label: formatMessage({
       id: 'app.utils.publish',
       defaultMessage: 'Publish',
@@ -754,14 +751,8 @@ const UpdateAction: DocumentActionComponent = ({
      * - the form is submitting
      * - the document is not modified & we're not cloning (you can save a clone entity straight away)
      * - the active tab is the published tab
-     * - the user doesn't have the permission to create a new document
-     * - the user doesn't have the permission to update the document
      */
-    disabled:
-      isSubmitting ||
-      (!modified && !isCloning) ||
-      activeTab === 'published' ||
-      Boolean((!documentId && !canCreate) || (documentId && !canUpdate)),
+    disabled: isSubmitting || (!modified && !isCloning) || activeTab === 'published',
     label: formatMessage({
       id: 'content-manager.containers.Edit.save',
       defaultMessage: 'Save',
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -410,7 +410,7 @@ export default {
           throw new errors.ForbiddenError();
         }
 
-        document = await createDocument(ctx);
+        document = await createDocument(ctx, { populate });
       }
 
       const isUpdate = !isCreate;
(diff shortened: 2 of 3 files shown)


Marc-Roig:
[CS-942](https://strapi-inc.atlassian.net/browse/CS-942) will be the follow up of this work

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -390,14 +390,39 @@ export default {
...
+      }
+
+      if (isCreate) {
+        await createDocument(ctx);
+      }
+
+      // Update if user has update permissions
+      if (isUpdate && permissionChecker.can.update(document)) {
+        await updateDocument(ctx);
+      }
```
remidej, line 424: What if the user doesn't have the permission? Wouldn't we want to throw?

Marc-Roig: In this scenario we don't want to throw,

If the user doesn't have any permission , we want both Publish and Save buttons to be disabled 
image

If the user does have all permissions, the "Publish" button will save and publish

But if the user only has publish permissions, we want the "Publish" button only to publish

Marc-Roig: In the scenario where user does not have update permissions, it will not be able to update the draft either as all attribute fields will be disabled.

remidej: Ah yes makes sense, and I see it's explained in the comment above ✨

remidej (approved):

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -390,14 +390,39 @@ export default {
...
+       * 2. Update an existing document and publish it in one request
+       *
+       * Based on user permissions:
+       * 1. User cannot create a document, but can publish
+       *    Action will be forbidden as user cannot create a document
+       * 2. User can update and publish a document
+       *    Action will be allowed, but document will not be updated, only published with the latest draft
+       */
+      const isCreate = !id;
+      const isUpdate = id;
```
Convly, line 411: WHAT IF `id` is `0`? :eyes: 
(it shouldn't happen but maybe being more accurate in our checks could prevent bugs in weird setups in the future?)

Convly: Also, `isXXX` often reference boolean, but the inference will probably won't consider your variables as such, you might want to enforce some type here

Convly (commented):
Logic is fine for me, some API tests are failing. probably due to the changes

_(I'm also being a bit annoying on checks)_

packages/core/content-manager/server/src/controllers/collection-types.ts
Marc-Roig: good catch! will update that 🚀

simotae14:
Hey do you need help for the FE part? because I can still see the Publish button disabled

https://github.com/strapi/strapi/assets/2589748/cefe5222-ed87-44e1-8822-66c91e963208

Marc-Roig:
@simotae14 yes 🙏 I was going to ask yannnnis, but if we have time we could do the FE in this PR too

remidej (approved):
It's working well! Just 1 thing I don't understand in the code

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -754,14 +751,8 @@ const UpdateAction: DocumentActionComponent = ({
...
      * - the active tab is the published tab
-     * - the user doesn't have the permission to create a new document
-     * - the user doesn't have the permission to update the document
      */
-    disabled:
-      isSubmitting ||
-      (!modified && !isCloning) ||
-      activeTab === 'published' ||
-      Boolean((!documentId && !canCreate) || (documentId && !canUpdate)),
+    disabled: isSubmitting || (!modified && !isCloning) || activeTab === 'published',
```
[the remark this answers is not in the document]
simotae14: Otherwise if you don't have create or update permission you can't publish, I did it to follow the same logic we have decided in the BE

PR #20670, opened 2024-07-02
fix(e2e): new data set to fix shop single types

jhoward1994:
### What does it do?

In order to fix: 

https://github.com/strapi/strapi/blob/25a6bed81bfc1f859e5fe283df9508d1d7033b7d/tests/e2e/tests/content-manager/editview.spec.ts#L625-L626

The problem seems to be that our seed data `with-admin.tar` had 2 documents for the shop single type in the same locale

One was being loaded into the cm edit view but the other was being published during the e2e spec. It was failing at this point because the document loaded into the CM edit view actually hadn't had it's status changed by the publish action

https://github.com/strapi/strapi/blob/25a6bed81bfc1f859e5fe283df9508d1d7033b7d/tests/e2e/tests/content-manager/editview.spec.ts#L642-L656

### Why is it needed?

### How to test it?

Run `rm -rf test-apps; yarn test:e2e --setup --domains=content-manager -- tests/e2e/tests/content-manager/editview.spec.ts`

### Related issue(s)/PR(s)


Marc-Roig (approved):
I wonder how we got to the point of having two instances of the same locale, and if we could somehow "regenerate" invalid data like this in the future 🤔

PR #20674, opened 2024-07-03
chore: run relation permissions test

Marc-Roig:
### What does it do?

Spotted an api tests that was skipped. 
Adapted the response formats to make tests pass

Note: Users & Permissions Content Manager endpoints are not returning the content wrapped inside a `data` object. We might want to take a look at that in the future.


remidej (approved):
The tests are passing, lgtm

PR #20679, opened 2024-07-03
fix(content-releases): check time has not passed if date is today

Feranchz:
### What does it do?

Fix bugs on Releases modal when creating a release with scheduling & also show again translated error messages in release modal. 

### How to test it?

Try to create a new release with scheduling enabled trying with different times.


simotae14:
Hey @Feranchz I think there is a problem with timezones, for example if I set the timezone to London and I try to schedule the release for today London time at 10:45 I receive an error even if the time is not in the past, as you can see in this video

https://github.com/strapi/strapi/assets/2589748/85d91a27-ea1d-4230-bc8a-d82dc0c24efb

simotae14 (approved):
Tested and it works, well done 👏🏽

PR #20687, opened 2024-07-04
Fix/dynamic-zone-unique-validation

jhoward1994:
feat(core): api test for dynamic zone unique fields

### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -153,6 +154,175 @@ const addUniqueValidator = <T extends yup.AnySchema>(
...
+      },
+      0
+    );
+
+    if (countOfValueInThisEntity > 1) {
+      // If the value is repeated in the current entity, the validation fails.
+      return false;
+    }
+
+    // When validating a unique field within a dynamiczone, we cannot
```
Convly, line 247: What's concretely preventing us from using the query builder directly? :thinking: 

Using populate fragments & nested filters, it _should_ be possible to reach a valid query

jhoward1994:
Hey @Marc-Roig 

We have been discussing the best way to build the query for this validation

https://github.com/strapi/strapi/blob/d4bab65e8beac1c5e5f1ea19028059fb514b8534/packages/core/core/src/services/entity-validator/validators.ts#L231-L423

You can see the 2 options in the code

- Option 1 uses the query builder, is more readable and easier to maintain

- Option 2 is what I went with initially as I didn't know about this populate notation that we are using in Option 1

https://github.com/strapi/strapi/blob/d4bab65e8beac1c5e5f1ea19028059fb514b8534/packages/core/core/src/services/entity-validator/validators.ts#L266-L272

It uses Knex directly to first select all the entity IDs of all components of the target type that are used within the parent content type in this dynamic zone. Before using the `strapi.db` for the final uniqueness check

Mainly because of this join Option 2 will select fewer rows from the DB and therefore be more performant

https://github.com/strapi/strapi/blob/d4bab65e8beac1c5e5f1ea19028059fb514b8534/packages/core/core/src/services/entity-validator/validators.ts#L353-L359

So I think it's a trade off between readability and performance. we were hoping to get your opinion on which approach you think is best? 🙏🏻

Marc-Roig:
@jhoward1994 sorry for the delay :) this is a tricky one

In this scenario, which is only used on "mutations" I would prioritize readability over performance (if it's not a gigantic difference)

As we support many types of databases, doing knex queries can be troublesome, for example, the table names might require the database schema prefixed in some situations.

So yeah! as long as the difference is not a lot, I would vote for the populate option :)

jhoward1994:
@Marc-Roig & @Convly 

This is ready for review using the query builder implementation of the DZ unique validation. Let me know if you have any q's 🙂

Convly (approved):

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -153,6 +156,168 @@ const addUniqueValidator = <T extends yup.AnySchema>(
...
+
+    // Populate the dynamic zone for any components that share the same value
+    // as the updated attribute.
+    const query: QueryType = {
+      select: ['id'],
+      where: {},
+      populate: {
+        [startOfPath]: {
+          on: {
+            [targetComponentUID]: { where: { [updatedAttribute.name]: updatedAttribute.value } },
```
Convly, line 283: Just a side thought, but we could probably optimize a bit the query by only querying the id column from populated components

jhoward1994: Thanks! Would this do the trick [`31dcab9` (#20687)](https://github.com/strapi/strapi/pull/20687/commits/31dcab9056435b19a3ac7f4891cc1781d480ed1f)

We need access to the `__component` here 
https://github.com/strapi/strapi/blob/31dcab9056435b19a3ac7f4891cc1781d480ed1f/packages/core/core/src/services/entity-validator/validators.ts#L318-L319

so I thought I would have to do:
`select: ['id', '__component'],`

But it seems to be working with just id

PR #20691, opened 2024-07-05
Fix Failing E2E Tests

Convly:
### What does it do?

- fix several bugs causing tests to fail
- fix several broken tests
- skip some i18n tests that are failing from a bug; will re-enable when bug is fixed

### Why is it needed?

Too many tests were failing, and without the e2e test requirement on in the CI more were being added. This gets us back to a state where we can rely on e2e tests again 🎉

### How to test it?

All tests should pass.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -778,6 +778,7 @@ const UpdateAction: DocumentActionComponent = ({
       setSubmitting(true);
 
       try {
+        // TODO: This is not what we should do. We should just run some validation and skip others instead.
         // Don't run the validation for drafts
         if (activeTab !== 'draft') {
           const { errors } = await validate();
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -778,19 +778,22 @@ const UpdateAction: DocumentActionComponent = ({
       setSubmitting(true);
 
       try {
-        const { errors } = await validate();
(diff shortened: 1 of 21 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -770,19 +770,22 @@ const UpdateAction: DocumentActionComponent = ({
...
-            message: formatMessage({
-              id: 'content-manager.validation.error',
-              defaultMessage:
-                'There are validation errors in your document. Please fix them before saving.',
-            }),
-          });
-
-          return;
+        // Don't run the validation for drafts
+        if (activeTab !== 'draft') {
```
Convly, line 774: This is not what we should do. We should just run some validation and skip others instead.

packages/core/content-manager/admin/src/utils/validation.ts
```diff
@@ -287,7 +302,7 @@ const addMinValidation: ValidationFn =
         (attribute.type === 'component' && attribute.repeatable) ||
         attribute.type === 'dynamiczone'
       ) {
-        if (!attribute.required && 'test' in schema && min) {
+        if (options.status !== 'draft' && !attribute.required && 'test' in schema && min) {
```
Convly, line 305: Should we also handle https://github.com/strapi/strapi/pull/20691/files#diff-c186be7a068bde5fa18b001d2973bfe8a3981eb96e405ef4b87d0a80be5a7dedR330 the same way?

innerdvations: Yeah, but let's handle it in another PR, we need to get this merged ASAP to get tests fixed

Bassel17 (approved):
LGTM !

remidej (approved):

PR #20704, opened 2024-07-08
fix(documentation): add check to password input

Feranchz:
### What does it do?

Add right validations to password input on documentation plugin

### How to test it?

Go to the documentation plugin in the settings page and try to setup a new password. It should have:
1. At least one uppercase letter
2. At least one lowercase letter
3. At least one number


remidej (commented):
The validation is indeed needed, but if I recall correctly the frontend wasn't showing that the backend was erroring. Can you let me know if that's the case? Because if so we should also throw an error notification on the frontend when something goes wrong

Feranchz:
@remidej Yes, you're right. I change some logic to show the notification when there is an error from the backend

remidej (approved):
Works nicely 💯

Issue #20643, opened 2024-06-27
Content filter filters only on 'Starts with' but should filter on 'contains'


ariadna-aldeguer:
### Required System information

- Node.js version: v18.20.2
- NPM version: 10.5.0
- Strapi version: 4.24.5
- Database: MySQL
- Operating system: win32-x64
- Is your project Javascript or Typescript: Typescript

### Describe the bug

Content filter filters only on 'Starts with' but should filter on 'contains'

### Steps to reproduce the behavior

1. Go to 'Content Manager'
2. Click on Search buttón next to Content
<img width="95" alt="image" src="https://github.com/strapi/strapi/assets/71792438/154a7db9-2a62-4ff4-bc23-db90ee004d07">

3. Search for a word that does not appear in the first position

### Expected behavior

If I search for 'aviso', 'Single Type' appears but if I search for 'legal' it does not appear

<img width="277" alt="image" src="https://github.com/strapi/strapi/assets/71792438/997af4ef-e744-4ad0-b409-1389b0e9a484">

remidej:
fixed by #20706

PR #20706, opened 2024-07-08
[feat] content manager filter logic

noobCode-69:
### What does it do?

Changes the filter logic for content manager , instead of using `startsWith` it uses `includes`


noobCode-69:
solves #20643

alexandrebodin:
Hi @noobCode-69 We have just released v5. 

v4 is now in maintenance mode but we would really like to get PRs merged into v5 if possible/still needed.

Could you check if this issue still exists there and resubmit your PR on `develop`?

remidej (approved):
I tested and confirm this does fix the bug. Thanks!

Issue #20475, opened 2024-06-10
[Design] Inconsistent height of UI elements (e.g. search and sort)


jhoward1994:
### Required System information

- Node.js version:
- NPM version:
- Strapi version: v5.0.0-beta.9
- Database:
- Operating system:
- Is your project Javascript or Typescript:

### Describe the bug

Inconsistent height of UI elements - search bars and sort drop downs

### Steps to reproduce the behavior

1. Go to Upload plugin
### Screenshots

Example from upload plugin

<img width="1397" alt="Screenshot 2024-06-10 at 11 07 42" src="https://github.com/strapi/strapi/assets/48524071/039f11ee-9b73-4d90-b63e-922432c98d46">

Eventyret:
Confirmed `v5.0.0-beta.9` seems filters and other minor buttons are smaller while search and dropdown box is larger.

lucasboilly:
@remidej the select height was fixed but the issue remains for the search. What we need to do is update the design of the search bar in the design system.

lucasboilly:
Should we close the issue?

jhoward1994:
Yes I would say so. Closing now, I will double check and open again in the DS repo if it's still relevant 🙏🏻

lucasboilly:
@jhoward1994 the sort is already fixed and the search bar will be fixed by https://github.com/strapi/design-system/pull/1833.

We can keep it closed 👍

PR #20708, opened 2024-07-08
fix: media library action container alignment issue

noobCode-69:
### What does it do?

removes the un-necessary padding-y which causes the alignment issue 

Screenshot 2024-07-09 at 12 14 12 AM

`perfectly aligned as all things should be`


noobCode-69:
closes  #20475

alexandrebodin:
Hi @noobCode-69 We have just released v5. 

v4 is now in maintenance mode but we would really like to get PRs merged into v5 if possible/still needed.

Could you check if this issue still exists there and resubmit your PR on `develop`?

remidej (approved):
Nice 👍 thank you

I won't close the issue you linked because that one is about Strapi 5

PR #20710, opened 2024-07-08
[chore] component name casing bug 

noobCode-69:
### What does it do?

Remove the logic which apply upper casing to component name in media library


remidej (changes requested):
I like the change but we would need to change the formatting [in the page's heading too](https://github.com/strapi/strapi/blob/696bc91c729eb24b0c29d8518e473e1381864108/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx#L169):

CleanShot 2024-09-20 at 09 57 27@2x

alexandrebodin:
Hi, I'm going to close as it's stale and we are working on a refactor current and these views will change.  Thank you for your contribution :)

PR #20713, opened 2024-07-09
feat(cm): add optimistic document updates

### LINKED ISSUE #20028: v5: rich-text blocks slate editor crashes when hit `Enter` key to new line and hit `save`

- Node.js version: v20.12.0
- NPM version: 10.1.0
- Strapi version: v5.0.0-beta.2
- Database: sqlite
- Operating system: ubuntu 22.04
- Is your project Javascript or Typescript: ts

### Describe the bug

- content manager edit view page with rich-text blocks slate editor crashes when hit `Enter` key to new line and hit `save`

1. npx create-strapi-app@beta test-page11 --quickstart  --ts
2. npm run develop
3. create a new collection type, with a plain text field `title11` and rich text blocks field `rich11`
4. in content manager, type in the `rich11` slate editor, hit `Enter` key to new line and hit `save`, the page will crash
5. if no crash happens, try to reproduce by hitting `Enter` key twice and hit `save`, then hitting `Enter` key and hit `save`

### Expected behavior

- in rich text slate editor, hit `Enter` key to new line and hit `save`, the content should be saved normally

### Screenshots

image

### Code snippets

### Additional context

- I've test on v4. v4 doesn't have this issue

remidej:
### What does it do?

Adds optimistic behavior to the RTK updateDocument API. We now assume that the backend won't throw errors when the user saves content, so that we can immediately update the UI with the new content when the user clicks update.

If the backend _does_ throw however, then the local state update is rollbacked, as you can see here:

https://github.com/strapi/strapi/assets/8087692/4a79c1d5-a1f7-47cd-a1df-caea9761e21d

### Why is it needed?

- It prevents flickering in the edit view: when you save an update, you would see the old version of your content for a split second.
- Fixes an issue in Blocks: #20028 (the Slate state wasn't kept properly in sync with the RTK state during that "flickering" phase)

### How to test it?

- Update content in the edit view, and save. The content in the inputs should not change, not even for a split second. Try both collection and single types, and try a variety of different fields.
- Add a new block to the end of a blocks field and save. The app shouldn't crash

### Related issue(s)/PR(s)

- need for [optimistic updates mentioned before there](https://github.com/strapi/strapi/pull/20558#issuecomment-2178849160)
- fixes #20028

### DIFF
--- a/packages/core/content-manager/admin/src/services/documents.ts
+++ b/packages/core/content-manager/admin/src/services/documents.ts
@@ -335,6 +335,20 @@ const documentApi = contentManagerApi.injectEndpoints({
           'Relations',
         ];
       },
+      async onQueryStarted({ data, ...patch }, { dispatch, queryFulfilled }) {
+        // Optimistically update the cache with the new data
+        const patchResult = dispatch(
+          documentApi.util.updateQueryData('getDocument', patch, (draft) => {
+            Object.assign(draft.data, data);
+          })
+        );
+        try {
+          await queryFulfilled;
+        } catch {
+          // Rollback the optimistic update if there's an error
+          patchResult.undo();
+        }
+      },
     }),
     unpublishDocument: builder.mutation<
       Unpublish.Response,


Feranchz (approved):
👍

simotae14 (approved):
wow, it works!!! great job

PR #20714, opened 2024-07-09
feat(cli): add project link command to cloud cli

gonbaum:
What does it do?
This PR introduces the project `link` command into the Strapi Cloud CLI

Why is it needed?
It gives users a chance to link their local projects to existing projects in Strapi Cloud from the CLI

How to test it?

1. Verify that the projects command is listed among the available commands.
2. Ensure that the link command executes correctly.
3. If the user is not logged in, they should be prompted to log in.
4. If the user is not in a valid project folder, the command should exit with a message.
5. If there is a project already linked to the current folder, it should ask the user if they want to update the current link and correctly display the name of the linked project.
6. If the user selects "no", the command should exit.
7. If the user selects "yes", the command should continue with the project fetching and display a list of projects for which the user is the owner (not maintainer).
8. When the user selects a project, the project config should be created, replacing the existing previous config.
9. A message for successful linking should be displayed.
10. Confirm that any 500 error returned from the server is handled gracefully, without causing the CLI to crash.
11. Verify that the rest of the commands function correctly.

### DIFF
--- a/packages/cli/cloud/src/link/action.ts
+++ b/packages/cli/cloud/src/link/action.ts
@@ -1,0 +1,83 @@
+import inquirer from 'inquirer';
+import chalk from 'chalk';
+
+import type { CLIContext } from '../types';
+import { cloudApiFactory, tokenServiceFactory, local } from '../services';
+import { promptLogin } from '../login/action';
+
+export default async (ctx: CLIContext) => {
+  const { getValidToken } = await tokenServiceFactory(ctx);
+  const token = await getValidToken(ctx, promptLogin);
+  const { logger } = ctx;
+
+  if (!token) {
+    return;
+  }
+
+  try {
+    const existingConfig = await local.retrieve();
+    if (existingConfig.project) {
+      const { shouldRelink } = await inquirer.prompt([
+        {
+          type: 'confirm',
+          name: 'shouldRelink',
+          message: `A project named ${chalk.cyan(
+            existingConfig.project.displayName
+              ? existingConfig.project.displayName
+              : existingConfig.project.name
+          )} is already linked to this local folder. Do you want to update the link?`,
+          default: false,
+        },
+      ]);
+
+      if (!shouldRelink) {
+        return;
+      }
+    }
+  } catch (e) {
+    logger.debug('Failed to check project config', e);
+    logger.error('An error occurred while trying to link the project.');
+  }
+
+  const cloudApiService = await cloudApiFactory(ctx, token);
+  const spinner = logger.spinner('Fetching your projects...\n').start();
+  try {
+    const {
+      data: { data: projectList },
+    } = await cloudApiService.listLinkProjects();
+    spinner.succeed();
+    const projects = Object.values(projectList)
+      .filter((project: any) => !project.isMaintainer)
+      .map((project: any) => {
+        return {
+          name: project.displayName,
+          value: { name: project.name, displayName: project.displayName },
+        };
+      });
+    if (projects.length === 0) {
+      logger.log("We couldn't find any projects available for linking in Strapi Cloud");
+      return;
+    }
+    await inquirer
+      .prompt([
+        {
+          type: 'list',
+          name: 'linkProject',
+          message: 'Which project do you want to link?',
+          choices: [...projects],
+        },
+      ])
+      .then(async (answer) => {
+        try {
+          await local.save({ project: { ...answer.linkProject } }, { override: true });
+          logger.log(`Project ${chalk.cyan(answer.linkProject.displayName)} linked successfully.`);
+        } catch (e) {
+          logger.debug('Failed to save project', e);
+          logger.error('An error occurred while linking the project.');
+        }
+      });
+  } catch (e) {
+    ctx.logger.debug('Failed to list projects', e);
+    spinner.fail('An error occurred while fetching your projects from Strapi Cloud.');
+  }
+};
--- a/packages/cli/cloud/src/link/action.ts
+++ b/packages/cli/cloud/src/link/action.ts
@@ -1,10 +1,21 @@
 import inquirer from 'inquirer';
 import chalk from 'chalk';
 
+import type { Answers } from 'inquirer';
 import type { CLIContext } from '../types';
 import { LocalSave } from '../services/strapi-info-save';
 import { cloudApiFactory, tokenServiceFactory, local } from '../services';
 import { promptLogin } from '../login/action';
+import { trackEvent } from '../utils/analytics';
+
+interface LinkProjectValue {
+  name: string;
+  displayName: string;
+}
+
+interface LinkProjectAnswer extends Answers {
+  linkProject: LinkProjectValue | 'quit';
+}
 
 export default async (ctx: CLIContext) => {
   const { getValidToken } = await tokenServiceFactory(ctx);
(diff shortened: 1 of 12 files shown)


packages/cli/cloud/src/list-projects/command.ts
```diff
@@ -12,7 +12,7 @@ const command: StrapiCloudCommand = ({ command, ctx }) => {
     .description('List Strapi Cloud projects')
     .option('-d, --debug', 'Enable debugging mode with verbose logs')
     .option('-s, --silent', "Don't log anything")
-    .action(() => runAction('listProjects', action)(ctx));
+    .action(() => runAction('projects', action)(ctx));
```
gonbaum, line 15: This is the name of the command that is displayed when there is an error message like when a user tries to run a command in a non strapi project folder: 'You need to run strapi <command> in a Strapi project...'

packages/cli/cloud/src/services/cli-api.ts
```diff
@@ -10,13 +10,15 @@ import packageJson from '../../package.json';
...
-  nodeVersion: string;
-  region: string;
+  displayName?: string;
+  nodeVersion?: string;
+  region?: string;
   plan?: string;
   url?: string;
 };
-export type ProjectInput = Omit<ProjectInfos, 'id'>;
+export type CreateProjectInput = Omit<ProjectInfos, 'id'>;
```
gonbaum, line 21: renamed this because it's the type we use for the input of the project creation

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -5,12 +5,14 @@ import type { ProjectInfos } from './cli-api';
...
 };
 
-export async function save(data: LocalSave, { directoryPath }: { directoryPath?: string } = {}) {
-  const alreadyInFileData = await retrieve({ directoryPath });
-  const storedData = { ...alreadyInFileData, ...data };
+export async function save(
+  data: LocalSave,
+  { directoryPath, override }: { directoryPath?: string; override?: boolean } = {}
+) {
+  const storedData = override ? data : { ...(await retrieve({ directoryPath })), ...data };
```
nathan-pichon, line 15: I believe it was more readable before, having the `await retrieve(...)` in another variable. 😅

nathan-pichon: Why do we need the override parameter? Its already overwritten with how the object is created 🤔

packages/cli/cloud/src/services/cli-api.ts
nathan-pichon: I believe the project would have the same input everywhere it is modified, not only in create? 🤔

packages/cli/cloud/src/link/index.ts
```diff
@@ -0,0 +1,12 @@
+import action from './action';
+import command from './command';
+import type { StrapiCloudCommandInfo } from '../types';
+
+export { action, command };
+
+export default {
+  name: 'link-project',
+  description: 'link a local directory to a Strapi Cloud project',
```
nathan-pichon, line 9:
```suggestion
  description: 'Link a local directory to a Strapi Cloud project',
```

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,134 @@
...
+              projectInternalName: answer.linkProject,
+            });
+          } catch (e) {
+            ctx.logger.debug('Failed to track didNotLinkProject', e);
+          }
+          return;
+        }
+
+        try {
+          await local.save({ project: { ...answer.linkProject } }, { override: true });
```
nathan-pichon, line 107: Why do you destructure `answer.linkProject` here?

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,134 @@
...
+    await inquirer
+      .prompt([
+        {
+          type: 'list',
+          name: 'linkProject',
+          message: 'Which project do you want to link?',
+          choices: [...projects, { name: chalk.grey('(Quit)'), value: 'quit' }],
+        },
+      ])
+      .then(async (answer) => {
```
nathan-pichon, line 81: Why do you use a `.then` here instead of `const answer = await inquirer.prompt(...);` ?

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,134 @@
...
+      ]);
+
+      if (!shouldRelink) {
+        return;
+      }
+    }
+  } catch (e) {
+    logger.debug('Failed to check project config', e);
+    logger.error('An error occurred while trying to link the project.');
+  }
```
nathan-pichon, line 43: I believe you can put this part into another function

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const cloudApiService = await cloudApiFactory(ctx, token);
+
+  try {
+    await cloudApiService.track('willLinkProject', {});
+  } catch (e) {
+    ctx.logger.debug('Failed to track willLinkProject', e);
+  }
```
nathan-pichon, line 51: Maybe we can have an util function to do this, as we do it everywhere.
Like
```javascript
function trackEvent({ logger, cloudApiService }, event, data = {}) {
  try {
    await cloudApiService.track(event, data);
  } catch (e) {
    logger.debug(`Failed to track ${event}`, e);
  }
}
```
WDYT?

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,134 @@
...
+            message:
+              'Warning: Once linked, deploying from CLI will replace the existing project and its data. Confirm to proceed:',
+            default: false,
+          },
+        ]);
+
+        if (!confirmLink) {
+          try {
+            await cloudApiService.track('didNotLinkProject', {
+              projectInternalName: answer.linkProject,
```
nathan-pichon, line 98: shouldn't projectInternalName be the current project linked? 🤔

packages/cli/cloud/src/services/strapi-info-save.ts
gonbaum: The save function in the original function merges the new data with the existing data. I don't want data from a previous config to coexist with new linked project data, so I've added an override parameter to re-use the function with the option of creating a brand new config object.

nathan-pichon: Yes but it will overwrite any added data in future.
Also, the project data, as it is an object, will be replaced by the new one, no?

packages/cli/cloud/src/link/action.ts
gonbaum: Silly mistake, I guess

packages/cli/cloud/src/services/strapi-info-save.ts
gonbaum: 1- Only if you pass the override parameter as true. Also for linking it works like that, you would like to have a brand new config when you link a new project to a folder with an existing config.
2- With the current implementation, the new project data will, but not the other keys of the object like node version and any other key you might have or want to add in the future.

gonbaum: I see the point now, we are using shallow merges so the override is not needed in this case. Good catch, thanks.

packages/cli/cloud/src/link/action.ts
gonbaum: I guess I was following inquirer examples in docs.

packages/cli/cloud/src/link/action.ts
gonbaum: I guess it depends on what you want to track:
1- The action to re-link the current folder to a new project named X failed
2- The action to re-link the current folder linked to project X failed

packages/cli/cloud/src/link/action.ts
gonbaum: I though about it, yes. We can have that.

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,196 @@
...
+  const spinner = ctx.logger.spinner('Fetching your projects...\n').start();
+
+  try {
+    const {
+      data: { data: projectList },
+    } = await cloudApiService.listLinkProjects();
+    spinner.succeed();
+    const projects: ProjectsList = Object.values(projectList)
+      .filter(
+        (project: any) => !(project.isMaintainer || project.name === existingConfig?.project?.name)
```
nathan-pichon, line 84: I believe any should be prohibited

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,196 @@
...
+  } catch (e) {
+    spinner.fail('An error occurred while fetching your projects from Strapi Cloud.');
+    ctx.logger.debug('Failed to list projects', e);
+    return null;
+  }
+}
+
+async function getUserSelection(
+  projects: ProjectsList,
+  logger: any
```
nathan-pichon, line 106: same here
Do eslint run on this file? 🤔

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,205 @@
...
+      data: { data: projectList },
+    } = await cloudApiService.listLinkProjects();
+    spinner.succeed();
+    const projects: ProjectsList = (projectList as unknown as Project[])
+      .filter(
+        (project: Project) =>
+          !(project.isMaintainer || project.name === existingConfig?.project?.name)
+      )
+      .map((project: Project) => {
+        console.log(project);
```
nathan-pichon, line 94: To be deleted

nathan-pichon (commented):
I'll do some QA this noon

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,204 @@
...
+        return null;
+      }
+    }
+  } catch (e) {
+    ctx.logger.debug('Failed to get project config', e);
+    ctx.logger.error('An error occurred while retrieving config data from your local project.');
+    return null;
+  }
+
+  return existingConfig;
```
nathan-pichon, line 73: I believe early return and const are preferred, but it's just an opinion.
```suggestion
  try {
    const existingConfig = await local.retrieve();
    if (existingConfig.project) {
      const { shouldRelink } = await inquirer.prompt([
        {
          type: 'confirm',
          name: 'shouldRelink',
          message: `A project named ${chalk.cyan(
            existingConfig.project.displayName
              ? existingConfig.project.displayName
              : existingConfig.project.name
          )} is already linked to this local folder. Do you want to update the link?`,
          default: false,
        },
      ]);

      if (!shouldRelink) {
        await trackEvent(ctx, cloudApiService, 'didNotLinkProject', {
          currentProjectName: existingConfig.project?.name,
        });
        return null;
      }
      return existingConfig;
    }
  } catch (e) {
    ctx.logger.debug('Failed to get project config', e);
    ctx.logger.error('An error occurred while retrieving config data from your local project.');
  }

  return null;
```

packages/cli/cloud/src/link/action.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  if (!token) {
+    return;
+  }
+
+  const cloudApiService = await cloudApiFactory(ctx, token);
+
+  const existingConfig: LocalSave | null = await getExistingConfig(ctx, cloudApiService);
+  // If user selects not to relink, return
+  if (!existingConfig) {
```
nathan-pichon, line 151: I believe `getExistingConfig` function doesn't really do what it says, as it also query User to know if we should link or not.

nathan-pichon (approved):

packages/cli/cloud/src/link/action.ts
gonbaum: Well, I've asked that same question myself, but I think the purpose is to decompose the main function into subfunctions that make the code easier to read, maintain, and test. We can extract the query outside of it, but that would leave getExistingConfig as merely a wrapper for local.retrieve(). Other options include:

- Don’t abstract away getConfig and prompting; keep it in the main function.
- Create a function only for prompting for re-link, though it would be quite small to justify being outside.
- Rename `getExistingConfig` to something like `getConfigAndPrompt`.

gonbaum: Resolved to refactor into single purpose functions

nathan-pichon (approved):

packages/cli/cloud/src/services/cli-api.ts
```diff
@@ -2,35 +2,42 @@ import axios, { type AxiosResponse } from 'axios';
...
-
 export type ListProjectsResponse = {
   data: {
     data: string;
   };
 };
 
+export type ListLinkProjectsResponse = {
+  data: {
+    data: ProjectInfos[];
```
abdallahmz, line 37: data can either be `ProjectInfos[]` or an empty object `{}` in case there is no project.

Currently we get an error if we have no project:

```
✖ An error occurred while fetching your projects from Strapi Cloud.
[DEBUG] Failed to list projects TypeError: projectList.filter is not a function
    at getProjectsList (/Users/abdallah/Documents/git/strapi/packages/cli/cloud/dist/index.js:1282:34)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async action$2 (/Users/abdallah/Documents/git/strapi/packages/cli/cloud/dist/index.js:1335:20)

```

gonbaum: Weird, tested that 🤔 maybe introduced in last commits

PR #20721, opened 2024-07-10
enhancement: New homepage with dynamic data

simotae14:
### What does it do?

Define a new homepage with statistics and dynamic data

### Why is it needed?

We want to show some precious data to our users

### How to test it?

Just open the cms and go to the admin page to check the different statistics


derrickmehaffy:
And you can bet I'm generating an exp release to see what this looks like, excited :star_struck:

derrickmehaffy:
https://github.com/strapi/strapi/actions/runs/9880427024

packages/core/admin/server/src/services/statistics.ts
```diff
@@ -0,0 +1,93 @@
...
+  await Promise.all(results);
+
+  return statistics;
+}
+
+const defaultProviders: {[name: string]: () => Promise<number>} = {
+  async admin_users() {
+    return strapi.query('admin::user').count();
+  },
+  async contentTypes() {
```
derrickmehaffy, line 42: Would be awesome to break this into an object with collection vs single types

packages/core/admin/server/src/services/statistics.ts
```diff
@@ -0,0 +1,93 @@
...
+    });
+  });
+
+  await Promise.all(results);
+
+  return statistics;
+}
+
+const defaultProviders: {[name: string]: () => Promise<number>} = {
+  async admin_users() {
```
derrickmehaffy, line 39: Same here, would be nice to get a breakdown of users and roles like:

```
admin: {
  users: x
  roles: y
}
```

packages/core/admin/server/src/services/statistics.ts
```diff
@@ -0,0 +1,93 @@
...
+    return strapi.query('admin::user').count();
+  },
+  async contentTypes() {
+    return Object.keys(strapi.contentTypes).filter((type) => type.startsWith('api::')).length;
+  },
+  assets: count('plugin::upload.file'),
```
derrickmehaffy, line 45: Possibly mime type breakdown too? :thinking:

packages/core/admin/server/src/services/statistics.ts
```diff
@@ -0,0 +1,93 @@
...
+  locales: count('plugin::i18n.locale'),
+  releases: count('plugin::content-releases.release'),
+  apiTokens: count('admin::api-token'),
+  async components() {
+    return Object.entries(strapi.components).length;
+  },
+  async webhooks() {
+    return strapi.query('strapi::webhook').count();
+  },
+  async entries() {
```
derrickmehaffy, line 55: If we could match this against the same filters used by cloud (eg detect if the user is running cloud and show entries based on that)

Mcastres:
SHIP IT!

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -177,4 +177,170 @@ export default {
...
+}[];
+
+async function assignedEntries(userId: number, locales: Record<string, string>): Promise<AssignedEntriesResult> {
+  const workflows = await strapi.documents('plugin::review-workflows.workflow').findMany();
+
+  const contentTypes = [...new Set(workflows.flatMap((workflow) => workflow.contentTypes))]; // Set to remove duplicates
+
+  const result: AssignedEntriesResult = [];
+
+  await Promise.all(
```
Baldinof, line 290: This would perform poorly in case the app has a lot content-types / entries. I think for a first iteration it would be better to not include this part.

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -177,4 +177,170 @@ export default {
 
     ctx.send({ plugins }) satisfies Plugins.Response;
   },
+
+  async dashboardData(ctx: Context) {
+    const stats = await getService('statistics').getStatistics();
```
Baldinof, line 182: I think most of the code should be moved to a dedicated `dashboard` service.

Mcastres:
Hello there 👋 

I submitted a commit on this PR. Please find below what it contains:

### CE Users

#### Project statistics 
- Entries count
- Assets count
- CT count
- Component count
- Locales count
- Admins count
- Webhooks count
- API tokens count

#### Collection Type overview
- Dropdown button to select an existing Collection-Type
    - Entries count
    - Pie Chart of entries count broken down by status
    - Locales count
    - Pie Chart of entries count broken down by locales
    - Stacked Area Chart of entries count created in time broken down by locales
    - Lists
        - Top 10 contributors (createBy/updateBy)
        - 10 Latest Draft entries
        - 10 Latest Published entries
  
### EE Users
#### Collection Type overview
- Stacked Area Chart of entries count created in time broken down by review workflow stages

#### Assigned Entries
- Assigned Entries count
- Pie Chart of assigned entries count broken down by review workflow stage
- Stacked Area Chart of assigned entries count created in time broken down by review workflow stage
- Lists of assigned entries broken down by review workflow stage (tabs)

#### Advanced features
- Upcoming release(s) count
- Number of days before the next release
- Overview of audit logs (last 10 actions)

## What's missing
While I'm proposing a dashboard foundation idea with this commit, some aspects are missing and will need discussions to move forward:

- When do we display the dashboard? Should it become the default homepage? Should it appear when there is the first Collection Type created? @Aurelsicoko @alexandrebodin 
- Permissions. This is more a technical question but based on what I proposed, how can I actually make sure that users can't see the collection-type data that they don't have permission to see? 
- I'm executing a single query using the Query Engine API. From what @derrickmehaffy said, it's not a good practice, but based on what I proposed, is there a better way to fetch the data and not making too much API requests using the Document Service API?
- Maybe there is some interesting data we can display in the dashboard that I'm missing

Thanks for your time!

This is how it looks like for an Enterprise User (Stacked Area Charts entries are all from the same day, so no area lol)
screen

Aurelsicoko:
Thank you @Mcastres for this contribution, it looks good! To answer your questions:

- We should display this dashboard once at least one content type and one entry are created. Otherwise, we should encourage the first user to follow the traditional onboarding process.
- Regarding the Permissions, it was something we identified during the first iteration, that's why we thought it was a good idea to keep it simple at first to ensure we only list items that the user should be able to see.
- We need a review from @lucasboilly to ensure it follows all the guidelines and best practices.

Baldinof:
Great work!

Some feedback:
1. I struggled to find the assigned entry list
1. Assigned entry list could include the content-type and maybe an extract of the entry based on some euristic like ``entry.title || entry.name || `${firstFieldName(entry)}: ${firstFieldValue(entry)}` ``
1. Maybe a title on the audit table?
1. The history could contain useful titles instead of `entry.update`, like `Updated 'Article' entry #documentId`

packages/core/admin/server/src/services/statistics.ts
```diff
@@ -0,0 +1,425 @@
...
+  const assignedEntries: Entry[] = [];
+
+  const contributors: Record<string, any> = {};
+  const stackedAreaChartByLocaleData: Record<string, any> = {};
+  const stackedAreaChartByStageData: Record<string, any> = {};
+  const stackedAreaChartByAssignedStageData: Record<string, any> = {};
+  const publishedDocumentIds = new Set<string>();
+
+  // Process each entry
+  entries.forEach(entry => {
```
Baldinof, line 174: I think this may take a while on large databases, we need to ensure it's behaving well in such case like:

- Add a timeout, to not waste server resources in case it's very slow (like 10s)
- Ensure the frontend is not blocked when the data is loading
- Caching could help too, but it seems complex to achieve the right way

alexandrebodin:
Hey @remidej @jhoward1994 & @Marc-Roig I assigned you the review for this so it doesn't stay out of the team loop

alexandrebodin:
@Mcastres Actually if you are taking this topic over please make a new PR so you can put it as ready etc and not depend other people please 🙏 and please assign Marc, Rémi, Jamie & Simone 🙏

simotae14:
great work done @Mcastres !! I really like the Charts. I saw the code and some stuff can be implemented better, but for a demo, it is ok. I think there is room for improvement mostly on the design of the page because, in my personal opinion, we need to show just a summary of the statistics to the user (for example the last table with a lot of rows, for me it is better to show just the last 3 activities) because at the end this one is a dashboard in the home page with a summary of the statistics. I agree we need to add some title to the different sections

Mcastres:
Thanks for the feeback everyone!
I created another PR as Alex suggested: https://github.com/strapi/strapi/pull/20924

alexandrebodin:
Going to close this in favor of the clean new PR. See you there

PR #20732, opened 2024-07-12
fix: allow saving ai generated content type

Marc-Roig:
### What does it do?

- Just hardcodes some schema values to ensure the schema is saveable

### DIFF
--- a/packages/core/content-type-builder/server/src/services/architect.ts
+++ b/packages/core/content-type-builder/server/src/services/architect.ts
@@ -31,11 +31,12 @@ const systemPrompt = `
             </schema_kind>
 
             <info_properties>
+              You MUST provide all the following properties:
+              - name: Same as pluralName
               - displayName: Default name to use in the admin panel
               - singularName: Singular form of the content-type name. Used to generate the API routes and databases/tables collection. Must be kebab-case.
               - pluralName: Plural form of the content-type name. Used to generate the API routes and databases/tables collection. Must be kebab-case.
               - description: Description of the model, it should describe the general purpose of the content-type, why it's used for, what it represents.
-              - name: Same as pluralName
             </info_properties>
 
             <field_types>
@@ -116,11 +117,11 @@ const systemPrompt = `
                   "kind": "collectionType",
                   "collectionName": "posts",
                   "info": {
-                    "displayName": "Blog Posts",
-                    "singularName": "Blog Post",
-                    "pluralName": "Blog Posts",
-                    "description": "A collection of blog posts",
                     "name": "posts"
+                    "displayName": "Blog Posts",
+                    "singularName": "blog-post",
+                    "pluralName": "blog-posts",
+                    "description": "A collection of blog posts",
                   },
                   "options": {
                     "draftAndPublish": true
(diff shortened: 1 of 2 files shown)


packages/core/content-type-builder/server/src/services/architect.ts
```diff
@@ -151,12 +152,46 @@ const systemPrompt = `
           </examples>
     `;
 
+const applyDefaultValues = (schema: any) => {
```
Marc-Roig, line 155: I am trying to fine tune the prompt but this seems like a good failsafe

packages/core/content-type-builder/server/src/services/architect.ts
```diff
@@ -31,11 +31,12 @@ const systemPrompt = `
             </schema_kind>
 
             <info_properties>
+              You MUST provide all the following properties:
+              - name: Same as pluralName
```
Convly, line 35: Does it actually work? I've tried similar prompts but didn't manage to fix the issue

packages/core/content-type-builder/server/src/services/architect.ts
Convly: I hate having to do this :grimacing: 
But if we don't have a choice, why not

packages/core/content-type-builder/server/src/services/architect.ts
```diff
@@ -151,12 +152,46 @@ const systemPrompt = `
           </examples>
     `;
 
+const applyDefaultValues = (schema: any) => {
+  // Transform info properties to be kebab-case always
+  schema.info.singularName = schema.info.singularName.toLowerCase().replace(/\s+/g, '-');
+  schema.info.pluralName = schema.info.pluralName.toLowerCase().replace(/\s+/g, '-');
+  if (!schema.info.name) schema.info.name = schema.info.pluralName;
```
Convly, line 159: :eyes:

packages/core/content-type-builder/server/src/services/architect.ts
```diff
@@ -151,12 +152,46 @@ const systemPrompt = `
...
+  return schema;
+};
+
 export const create = async (prompt: string, previousSchema?: object) => {
   console.log({ previousSchema });
 
   const { object } = await generateObject({
     model: anthropic('claude-3-sonnet-20240229', {}),
-    maxRetries: 3,
+    maxRetries: 10,
```
Convly, line 194: We should probably make this configurable :sweat_smile:

packages/core/content-type-builder/server/src/services/architect.ts
```diff
@@ -166,7 +201,7 @@ export const create = async (prompt: string, previousSchema?: object) => {
         singularName: z.string(),
         pluralName: z.string(),
         description: z.string().optional(),
-        name: z.string(),
+        name: z.string().optional(), // Model forgets to add this sometimes
```
Convly, line 204: LLMs are dumb :man_shrugging:

Marc-Roig: hahaha yes they are,
we need to adapt the prompt for it to be reliable, and the best approach imo is to just ask for dumb things, and make it spit dumb plain jsons, and avoid nested keys if possible

packages/core/content-type-builder/server/src/services/architect.ts
Marc-Roig: hehe

Marc-Roig:
Same, this is an elegant solution but also woulnd't be the final solution, 
a much better solution would have specific repairs for every part of the schema (repair fields, info, relations, ...)

packages/core/content-type-builder/server/src/services/architect.ts
Eventyret: Can't we cast it to unknown as that's what it is ?

Bassel17: we do what we have for a successful demo

packages/core/content-type-builder/server/src/services/architect.ts
Eventyret: I agree could cast it to process.env

packages/core/content-type-builder/server/src/services/architect.ts
Marc-Roig: 🫡 amen

packages/core/content-type-builder/server/src/services/architect.ts
Marc-Roig: sure!

Convly (approved):

PR #20742, opened 2024-07-15
test(e2e): fix broken listview and cloning tests

innerdvations:
### What does it do?

fixes some tests that were looking for old incorrect text values

### Why is it needed?

So the tests pass

### How to test it?

n/a

### Related issue(s)/PR(s)


jhoward1994:
This line is being temperamental for me locally. If I run in debug it passes fine but when I run the tests normally it fails. Do you see the same thing?

Convly:
We also have this one with some fixes. We might want to sync to see what we both learned?

https://github.com/strapi/strapi/pull/20691

innerdvations:
> We also have this one with some fixes. We might want to sync to see what we both learned?
> 
> #20691

Just checked, looks like yours includes everything in this one!

PR #15755, opened 2023-02-09
add ability to upload buffer

cpaczek:
### What does it do?

Ability to upload a nodejs file buffer. 

### Why is it needed?

Currently, the upload plugin wants a FormData object only available on the front end. this adds the ability to upload a node js buffer to the media library

### How to test it?

```js        
const file = fs.readFileSync('./test.png')
return strapi.plugin('upload').service('upload').uploadBuffer({file, fileName: crypto.randomUUID(), mime: 'image/png', ext: 'png', alternativeText: "test"})
```

PR #20769, opened 2024-07-16
enhancement: create a simple uploadFile method to programmatically upload files

alexandrebodin:
> ⚠️  Work in progress

### What does it do?

- Adds a `uploadFile` method to the upload service to make uploading file programmaticaly easier

### Why is it needed?

- A lot of users have been using internal methods because of the lack of public APIs. Use private code means your own code will be brittle and break anytime Strapi changes its internals. 

### How to test it?

It support file path, Buffers & read stream

```ts
strapi.plugin('upload').service('upload').uploadFile('path', { 
  name: '',
  alternativeText: '',
  caption: '',
})
```

### Related issue(s)/PR(s)

- https://github.com/strapi/strapi/pull/15755

### Todo

- [ ] Testing
- [ ] Add some more options to opt out of processing

### DIFF
--- a/packages/core/upload/server/src/services/upload.ts
+++ b/packages/core/upload/server/src/services/upload.ts
@@ -4,7 +4,7 @@ import crypto from 'crypto';
 import fs from 'fs';
 import fse from 'fs-extra';
 import _ from 'lodash';
-import { extension } from 'mime-types';
+import { extension, lookup } from 'mime-types';
 import {
   sanitize,
   strings,
@@ -553,9 +553,48 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
     });
   }
 
+  /**
+   * Uploads a file by path, buffer or stream programmaticaly
+   * @public
+   */
+  async function uploadFile(file: string | Buffer | fs.ReadStream, options: FileInfo) {
+    let filepath: string;
+
+    if (Buffer.isBuffer(file)) {
+      filepath = path.resolve(os.tmpdir(), crypto.randomBytes(10).toString('hex'));
+      await fse.writeFile(filepath, file);
+    } else if (file instanceof fs.ReadStream) {
+      filepath = path.resolve(os.tmpdir(), crypto.randomBytes(10).toString('hex'));
+      await fse.createWriteStream(filepath).write(file);
+    } else {
+      filepath = path.resolve(file);
+    }
+
+    const fileObject = {
+      filepath,
+      newFilename: options.name ?? path.basename(filepath),
+      originalFilename: options.name ?? path.basename(filepath),
+      mimetype: lookup(filepath) || 'application/octet-stream',
+      size: fs.statSync(filepath).size,
+    } as InputFile;
+
+    const tmpDir = await createAndAssignTmpWorkingDirectoryToFiles(fileObject);
+
+    try {
+      const fileData = await enhanceAndValidateFile(fileObject, options, {
+        tmpWorkingDirectory: tmpDir,
+      });
+      return await uploadFileAndPersist(fileData);
+    } catch (error) {
+      // delete temporary folder
+      await fse.remove(tmpDir);
+    }
+  }
+
   return {
     formatFileInfo,
     upload,
+    uploadFile,
     updateFileInfo,
     replace,
     findOne,


packages/core/upload/server/src/services/upload.ts
```diff
@@ -553,9 +553,48 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
+   */
+  async function uploadFile(file: string | Buffer | fs.ReadStream, options: FileInfo) {
+    let filepath: string;
+
+    if (Buffer.isBuffer(file)) {
+      filepath = path.resolve(os.tmpdir(), crypto.randomBytes(10).toString('hex'));
+      await fse.writeFile(filepath, file);
+    } else if (file instanceof fs.ReadStream) {
+      filepath = path.resolve(os.tmpdir(), crypto.randomBytes(10).toString('hex'));
+      await fse.createWriteStream(filepath).write(file);
```
cpaczek, line 568: On my phone rn so can't test it thoroughly but iirc you need to wrap streams in Pipeline() to prevent mem leaks. (May FSE does this under the hood)

cpaczek (commented):
Looks a lot cleaner than my original implementation in v4

derrickmehaffy:
Haven't tested yet but does this send it through sharp to do optimizations and such? If possible I'd love to have an option to skip that entirely maybe something like `optimize: false`? 🤔

alexandrebodin:
> Haven't tested yet but does this send it through sharp to do optimizations and such? If possible I'd love to have an option to skip that entirely maybe something like `optimize: false`? 🤔

Indeed right now it does go through all the normal processing

packages/core/upload/server/src/services/upload.ts
alexandrebodin: actually the way to avoid leaks is to store the stream on disk and then pass a path to the pipelining. this is what we recently did everywhere and it fixed both meme leaks & perf issues

packages/core/upload/server/src/services/upload.ts
```diff
@@ -553,9 +553,48 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
+    const tmpDir = await createAndAssignTmpWorkingDirectoryToFiles(fileObject);
+
+    try {
+      const fileData = await enhanceAndValidateFile(fileObject, options, {
+        tmpWorkingDirectory: tmpDir,
+      });
+      return await uploadFileAndPersist(fileData);
+    } catch (error) {
+      // delete temporary folder
+      await fse.remove(tmpDir);
```
Convly, line 590: Shouldn't we also delete the tmp file (if created)?

packages/core/upload/server/src/services/upload.ts
```diff
@@ -553,9 +553,48 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
     });
   }
 
+  /**
+   * Uploads a file by path, buffer or stream programmaticaly
+   * @public
+   */
+  async function uploadFile(file: string | Buffer | fs.ReadStream, options: FileInfo) {
```
Convly, line 560: Do we specifically want to only handle fs streams? Or should we allow any kind of readable stream? (`NodeJS.ReadableStream`)

packages/core/upload/server/src/services/upload.ts
Convly, line 568: You can't really "write" a readable stream to a writable stream though :thinking: 

Did you mean to pipe the readable into the writable?

Convly, line 568:
```suggestion
      filepath = path.resolve(os.tmpdir(), crypto.randomBytes(10).toString('hex'));
      const out = fse.createWriteStream(filepath);

      await new Promise((resolve, reject) => {
        file.pipe(out).on('finish', resolve).on('error', reject);
      });
```

Convly (commented):
I've tested it, it seems that both the buffer and stream method (fixed manually for testing purpose) are not resolving the correct mime type, which prevent the media library from displaying the images

Convly:
> I've tested it, it seems that both the buffer and stream method (fixed manually for testing purpose) are not resolving the correct mime type, which prevent the media library from displaying the images

Maybe giving the option to manually passe the mime type could help for such case? Or is there a way to automatically resolve the correct one?

Issue #20366, opened 2024-05-23
Long titled Collection Type breaks buttons display


maccomaccomaccomacco:
### Required System information

- Node.js version: 20
- NPM version: 10
- Strapi version: 5.0.0-beta.7
- Database: sqlite
- Operating system: mac
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When the collection type name is too long, the UI gets broken 

### Steps to reproduce the behavior

1. Create a collection type with a very long name
2. Save the content type
The content type name breaks the Ui and the nearest buttons style.

### Expected behavior

The content type name should be truncated with ellipses if it's too long and not break the style of the surrounding elements

### Screenshots

![U`ntitled](https://github.com/strapi/strapi/assets/2686869/3cc91d46-14ca-4b1d-a0bd-dcea7faa38da)
![Untitled](https://github.com/strapi/strapi/assets/2686869/2980b45d-bed8-4799-85c2-9d90855b9c0e)

hanpaine:
@derrickmehaffy what's the reasoning on this being a High? Do we know the maximum content length and that it is frequently exceeded?

simotae14:
Closed because it was fixed with this PR https://github.com/strapi/strapi/pull/20770

PR #20770, opened 2024-07-16
Fix Long Collection title breaks ui

simotae14:
### What does it do?

It solves the UI issues when you have a collection with a long title

### Why is it needed?

Because it breaks the UI in the Content manager, Content-Type builder, the Relation field and the Left menu

### How to test it?

 - create a new collection with a long name (use a long big word)
 - check if the Edit view of the new collection in the CTB has no UI issues and also its Left menu
 - check if the List view of the Collection has no UI issues in the content manager and its left menu
 - check if a relation with this collection has UI issues

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20366

### DIFF
--- a/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
+++ b/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
@@ -7,6 +7,7 @@ import isEqual from 'lodash/isEqual';
 import upperFirst from 'lodash/upperFirst';
 import { useIntl } from 'react-intl';
 import { unstable_usePrompt as usePrompt, useMatch } from 'react-router-dom';
+import { styled } from 'styled-components';
 
 import { List } from '../../components/List';
 import { ListRow } from '../../components/ListRow';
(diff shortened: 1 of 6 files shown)


packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
```diff
@@ -19,6 +20,10 @@ import { LinkToCMSettingsView } from './LinkToCMSettingsView';
 
 /* eslint-disable indent */
 
+const LayoutsHeaderCustom = styled(Layouts.Header)`
+  overflow-wrap: anywhere;
```
remidej, line 24: I wonder why anywhere and not break-word? Looking at the MDN docs it looks like it could affect margin/padding sizes

simotae14: Because we want to break long title with just one word, the break-word in this case doesn't work

remidej (approved):

Issue #20655, opened 2024-06-28
Strapi v5 rc: admin ui read entry issue (config2.doc || "").split is not a function


nasraldin:
### Required System information

- Node.js version: 20
- NPM version: 10
- Strapi version: 5.0.0-rc.0
- Database: Any db provider (SQLite, MySQL, Postgres)
- Operating system: Mac
- Is your project Javascript or Typescript: yes also happens in js

### Describe the bug

The problem began when we migrated from Strapi v4 to v5. Our data, stored as JSON, was easily handled in the old Strapi dashboard UI. However, after creating a new Strapi project in v5 and setting up a test API with a 'name' attribute of type JSON, the data stored in this column is now stringified. This causes an issue when trying to read and render it in the admin dashboard of the new version, resulting in exceptions as shown in the screenshot. Interestingly, the REST API operations continue to function correctly; the problem lies specifically in the admin dashboard's handling of these values.

### Steps to reproduce the behavior

1. Create a new v5 project with any database provider
2. from admin add a new collection type like test and have one attribute name type JSON
3. add some JSON values and save and update this from the database to meet our one old data as is JSON not stringified because Strapi now when you save it will be stringified in DB. From the database update this new entry to have a pure JSON object (as we already have our data as JSON not stringified so we need to read with any modification from the old system db) imaging you have the data seeded

### Expected behavior

Admin UI can render value as in old v4

### Code snippets

`
{
  "kind": "collectionType",
  "collectionName": "tests",
  "info": {
    "singularName": "test",
    "pluralName": "tests",
    "displayName": "test"
  },
  "options": {
    "draftAndPublish": false
  },
  "pluginOptions": {},
  "attributes": {
    "name": {
      "type": "json"
    }
  }
}`

DB
Saved from Admin UI
row sql `1	u9n4ruiare28wxgcoh1sc6yx	"{\"test\": 134}"	2024-06-28 09:38:32.286	2024-06-28 09:38:32.286	2024-06-28 09:38:32.281	1	1	`

Update the name column value by converting it to a JSON  object to meet our existing data in the old system
`1	u9n4ruiare28wxgcoh1sc6yx	{"test": 134}	2024-06-28 09:38:32.286	2024-06-28 09:38:32.286	2024-06-28 09:38:32.281	1	1`	

### Screenshots

[...]

Issue #16672, opened 2023-05-10
Issue with JSON field that has value as string


NotAhmedNasr:
## Bug report
A json field with value as json string gets turned into object after upgrading from 4.9.0 to 4.10.1

### Required System information

- Node.js version: v16.13.2
- NPM version: v8.1.2
- Strapi version: v4.10.1
- Database: PostgreSQL v12.0
- Operating system: Ubuntu 20.04.6 LTS
- Is your project Javascript or Typescript: Typescript

### Describe the bug
A json field that has a value being string is turned into an object. This shouldn't happen because string is indeed a valid json value regardless of what that string contains. But now if that string contains a json object it will explicitly be turned into an object. If that was the wanted behavior, I would just entered the value as an object not string.
This behavior appeared after [#16310](https://github.com/strapi/strapi/pull/16310)

### Steps to reproduce the behavior

1. Create a json field.
2. Set the value of the field to "{\"a\": 123}"
3. Save the entry.

### Expected behavior

The value of the field should remain "{\"a\": 123}". Instead it is turned into { "a": 123 }

#### Before saving
![1](https://github.com/strapi/strapi/assets/26029266/eb91f93c-2c0c-4422-8013-8fe924ea1a7b)
![2](https://github.com/strapi/strapi/assets/26029266/b3dc5ebc-f237-4eec-bd93-05a65d30c7b0)
#### After saving
![3](https://github.com/strapi/strapi/assets/26029266/8335e8a4-098d-4943-b0ed-bed9085cf945)
![4](https://github.com/strapi/strapi/assets/26029266/f588106e-cbf8-4384-bfbb-aa40ada66d15)

### Additional context

I know that it's not likely that anyone would enter the value of json field as json string explicitly from the admin UI but this examples shows the issue better.
Our real case is that we have a custom field that has a type json and the value we store in that field is a json string. so when the value of the filed is fetched from the server it tries to parse it then crashes because it finds the value to be an object not a string.

#### Before saving:
![Field](https://github.com/strapi/strapi/assets/26029266/62ec35f8-c378-4f69-8aeb-89ff7d1d10bd)
![create](https://github.com/strapi/strapi/assets/26029266/c06d26dd-2dfe-4c6f-9d7c-de047531bdcc)

### After saving
![find](https://github.com/strapi/strapi/assets/26029266/35c0b89d-eeab-490c-a26d-0a5ed7fea837)
[...]

PR #20829, opened 2024-07-17
fix: json input stringifying value when it shouldn't

remidej:
### What does it do?

Ensures that even though the JSON input from the design system deals with string data, at the Content Manager, we manipulate JSON objects instead.

In v5 users noticed the data type in the database for JSON inputs had become string instead of json objects. This was because in v5 we were sending stringified JSON data from the admin, instead of objects like in v4.

so what this does:
- ensures all saved data for JSON inputs is objects
- prevents crashed if the data in the database was a string instead of an object
- auto-formats json data in the input

### How to test it?

- try to write invalid json in an input. You should see your update, but a validation should stop you from saving it
- write and save valid content in an input. Open a database explorer. You should see the content saved as JSON (not in a string)

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20655

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Json.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Json.tsx
@@ -16,20 +16,34 @@ const JsonInput = forwardRef<JSONInputRef, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
     const field = useField(name);
     const fieldRef = useFocusInputField(name);
-
     const composedRefs = useComposedRefs(ref, fieldRef);
+
+    const handleChange = (json: string) => {
+      if (required && !json.length) {
+        // Default to null when the field is not required and there is no input value
+        field.onChange(name, null);
+      } else {
+        try {
+          field.onChange(name, JSON.parse(json));
+        } catch (e) {
+          /**
+           * Save the string value even though it's invalid to keep providing visual feedback.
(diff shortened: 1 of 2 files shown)


nasraldin:
Thanks, @remidej, for the fix. When can we expect this to be merged so we can test it in our application?

remidej:
@nasraldin it should land in the next RC on Wednesday

Feranchz:
I think this limit the json field to only accept objects, and I think that is fine but that's probably a breaking change to document. This because strings and numbers are valid JSON values, you can see one example in this [issue](https://github.com/strapi/strapi/issues/16672). 

Also looks like there is a bug in the json input when you only write a number, but I think is in the DS side

nasraldin:
Dear @remidej , @Feranchz   , @Is fix released, it seems still exist I have updated to 5.0.0-rc.6 and test but got same issue

remidej:
@nasraldin sorry the last RC shipped just before I merged #20866, but another one will be released today with the fix

nasraldin:
@remidej  thanks for update

remidej:
@nasraldin it should be up now in rc.7, could you test if it fixes your issue?

nasraldin:
thanks @remidej, it works as expected now

Screenshot 2024-07-24 at 6 33 16 PM
Screenshot 2024-07-24 at 6 33 45 PM

PR #20830, opened 2024-07-17
Use session cookie storage for login

Bassel17:
### What does it do?

- uses cookie storage (expiring when browser closes) for the jwt token instead of `window.sessionStorage`

### Why is it needed?

- so we can preserve login state across different tabs or windows unless the browser was closed

### How to test it?

- login and open a new tab or window you should login automatically to strapi


innerdvations (commented):
First thought is that this is exactly what we need to do, we just need to verify a few things next week.

- GDPR regarding session cookies (afaik it's fine)
- XSRF implications?
- we need some tests for every scenario

derrickmehaffy:
This will probably break split admin/server and cross domain cookies. Given split admin is still broken on v4 I haven't tested it on v5 yet.

innerdvations:
> This will probably break split admin/server and cross domain cookies. Given split admin is still broken on v4 I haven't tested it on v5 yet.

Yeah, we will need to do some testing there for sure and probably some futher enhancements. This is more of a POC to demo a potential solution to the problem, that we will refine.

I'll add a "do not merge" just to be safe until we know we've accounted for everything.

alexandrebodin:
Switching to cookies will sadly be a breaking change and prevent the admin from being hosted on another domain unless we support cross domain login (sso would need to also support it)

Convly (commented):
LGTM codewise, but it indeed causes issues for admin on a different host.

Should we put it in draft/close it while we think about how to address this? Or should we just let it rest here?

PR #20831, opened 2024-07-17
chore(cloud-cli): update dependencies

nathan-pichon:
### What does it do?

Updating axios and fs-extra dependency to use the same as other packages in the monorepo.


abdallahmz (approved):
Tested login, logout, deploy (on an existing project), all good ✅

PR #20846, opened 2024-07-17
fix: show media dialog in blocks and markdown editors

### LINKED ISSUE #20743: Adding an image to blocks from Toolbar list doesn’t work
https://www.notion.so/strapi/Adding-an-image-to-blocks-from-Toolbar-list-doesn-t-work-15be551b8cc5425fb4d76af9cd002c41?pvs=4

remidej:
### What does it do?

The media dialog did not display in the rich text editors (both Blocks and Markdown). This fixes it

The open/closed state is actually managed outside of the MediaLibraryDialog, by conditionally rendering it. So when it is rendered, it should always display its content, hence the `open={true}`

Also removes some unnecessary onClose code, because since the component is now remounted when we reopen it, the state will be reinitialized, no need to do it manually

### How to test it?

- Add an image to a blocks field. Create a folder too from within the dialog.
- Do the same in a markdown field

### Related issue(s)/PR(s)

fix #20743


simotae14 (approved):
it works 🥇

PR #20847, opened 2024-07-18
feat: history version on bulk publish locales

Marc-Roig:
### What does it do?

- Creates a history version when bulk publishing multiple locales of the same document.
- Splitted the history document service middleware as it was getting a bit long. 

### How to test it?

- Go to the Edit View of an entry
- Bulk publish one or more locales
- See that each locale has a "publish" history version created

### DIFF
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -143,14 +143,14 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
         });
 
         await strapi.db.transaction(async ({ onCommit }) => {
+          // .createVersion() is executed asynchronously,
+          // onCommit prevents creating a history version
+          // when the transaction has already been committed
           onCommit(async () => {
             for (const entry of localeEntries) {
               const status = await serviceUtils.getVersionStatus(uid, entry);
 
-              // .createVersion is executed asynchronously,
-              // onCommit prevents creating a history version
-              // when the transaction has already been committed
-              getService(strapi, 'history').createVersion({
+              await getService(strapi, 'history').createVersion({
                 contentType: uid,
                 data: omit(FIELDS_TO_IGNORE, entry) as Modules.Documents.AnyDocument,
                 relatedDocumentId: documentId,
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -1,6 +1,6 @@
-import type { Core, Modules } from '@strapi/types';
+import type { Core, Modules, UID } from '@strapi/types';
 
-import { omit } from 'lodash/fp';
+import { omit, castArray } from 'lodash/fp';
 
 import { scheduleJob } from 'node-schedule';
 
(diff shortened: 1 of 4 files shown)


packages/core/content-manager/server/src/services/document-metadata.ts
```diff
@@ -268,7 +268,13 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
...
   ) {
     if (!document) {
-      return document;
+      return {
+        data: document,
+        meta: {
+          availableLocales: [],
+          availableStatus: [],
+        },
+      };
```
Marc-Roig, line 277: Saw that if a middleware incorrectly returns an empty result we throw a non descriptive error in here.

packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -27,110 +107,58 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+
+        // Find all affected entries
+        const localeEntries = await strapi.db.query(uid).findMany({
+          where: {
+            documentId,
+            locale: { $in: locales },
+            publishedAt: null,
+          },
+          populate: serviceUtils.getDeepPopulate(uid, true /* use database syntax */),
         });
```
remidej, line 143: Why do you use the query service and not the document one?

packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -10,6 +10,86 @@ import { FIELDS_TO_IGNORE, HISTORY_VERSION_UID } from '../constants';
...
+   * It creates confusion for users because they see two history versions each publish action.
+   * To avoid this, we silence the update action during a publish request,
+   * so that they only see the published version of the document in the history.
+   */
+  if (
+    context.action === 'update' &&
+    strapi.requestContext.get()?.request.url.endsWith('/actions/publish')
+  ) {
+    return false;
+  }
```
remidej, line 50: Will this work for bulk publish too?

packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -27,110 +107,58 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-              status,
-            });
+          onCommit(async () => {
+            for (const entry of localeEntries) {
+              const status = await serviceUtils.getVersionStatus(uid, entry);
+
+              // .createVersion is executed asynchronously,
+              // onCommit prevents creating a history version
+              // when the transaction has already been committed
+              getService(strapi, 'history').createVersion({
```
remidej, line 153: You probably need to await here

remidej (commented):
Thank you for breaking up that chunky middleware ❤️ and for the tests

packages/core/content-manager/server/src/history/services/lifecycles.ts
Marc-Roig: atm we don't have a way to get multiple locales of the same document, so this way we can save some queries we send to the db

packages/core/content-manager/server/src/history/services/lifecycles.ts
Marc-Roig: Good point, I forgot about that.

This case doesn't work on bulk publish, but also do not update entries, so it's not necessary either!

packages/core/content-manager/server/src/history/services/lifecycles.ts
Marc-Roig: yes I do 🏃

remidej (approved):
QAEed 🔥

PR #20848, opened 2024-07-18
fix(content-releases): add bulk add to release to v5

Feranchz:
### What does it do?

Add again the bulk add to release feature to Strapi v5.

### How to test it?

1. On EE edition, go the the list view of an CT with draft & publish enabled
2. Select multiple entries and try to add them to one release
3. You should see in the list view table the releases column


packages/core/content-releases/server/src/controllers/release.ts
```diff
@@ -156,43 +156,58 @@ const releaseController = {
...
+    const releasesWithActions = await releaseService.findMany({
+      where: {
+        releasedAt: null,
+        actions: {
+          contentType: contentTypeUid,
+          entryDocumentId: {
+            $in: documentIds,
+          },
+          locale,
+        },
```
Marc-Roig, line 177: just in case, it's not possible to make a release to publish and unpublish the same entry right? Asking because the publishedAt value is not specified here, but I assume that's the case

Marc-Roig:
code looks good, could you add some api tests? 🙏

packages/core/content-releases/server/src/controllers/release.ts
Feranchz: Yes, is not possible. When you try to add the same entry a second time it should throw an error

madhurisandbhor (commented):
Styling issue for the `add to release` button, also carret icon is missing in front of release count. 

Screenshot 2024-07-22 at 11 07 13

One more thing, may be its not part od this PR, but in edit view there is no `add to release` button.
Screenshot 2024-07-22 at 11 07 55

Feranchz:
@Marc-Roig Done! I just migrate the v4 API tests that were not migrated yet. 

@madhurisandbhor Add to release button is now in the "More actions" menu (the button with 3 dots)

Marc-Roig (approved):

Marc-Roig:
unit tests seem to be failing

remidej:
@Marc-Roig looks like it's a typo causing the error, I'll push a patch now and hopefully merge

PR #20850, opened 2024-07-18
fix(cloud-cli): avoid blocking event loop in deploy command

nathan-pichon:
### What does it do?

This PR adds some logs and handle a timeout that was not properly cleared before ending the deploy action.

### Why is it needed?

Avoid our users waiting for unnecessary time when something wrong happens with the logs from the build engine.

### How to test it?

You may need to edit the code to try to get logs from a not found URL (I added 123 in the URL for example), in the build engine service.
Then run the deploy command on a strapi project. You will reach the max retries and the command should gracefully quit the node process.

### Related issue(s)/PR(s)


gonbaum (approved):
LGTM

Issue #20759, opened 2024-07-16
[Design] Missing disabled aspect for delete locale and delete document


MarionLemaire:
https://www.notion.so/strapi/Missing-disabled-aspect-for-delete-locale-and-delete-document-cd91baa7a6e649d598af958315345432?pvs=4

https://github.com/user-attachments/assets/6fbbf5a6-ca7a-431d-b0ff-e2038c96d6a7

simotae14:
fixed with https://github.com/strapi/strapi/pull/20851

PR #20851, opened 2024-07-19
fix color in disabled buttons

simotae14:
### What does it do?

It fixes a UI problem we had in the Document actions when some of them are disabled, instead of using the variant color it is better to inherit the disabled color

### Why is it needed?

Because otherwise, the UI is wrong, we show some disabled buttons with the enabled color

### How to test it?

- Go to the CM and add a new document to a collection
- before saving, click on the More button in the Header, you can see the Delete document and Delete locale buttons disabled with the correct color

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20759

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -298,14 +298,7 @@ const DocumentActionsMenu = ({
                   gap={2}
                   tag="span"
                 >
-                  <Box
-                    tag="span"
-                    color={
-                      !action.disabled ? convertActionVariantToColor(action.variant) : 'inherit'
-                    }
-                  >
-                    {action.icon}
-                  </Box>
+                  <span>{action.icon}</span>
                   {action.label}
                 </Flex>
                 {/* TODO: remove this in 5.1 release */}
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -293,8 +293,17 @@ const DocumentActionsMenu = ({
...
+                <Flex
+                  color={!action.disabled ? convertActionVariantToColor(action.variant) : 'inherit'}
+                  gap={2}
+                  tag="span"
+                >
+                  <Box
+                    tag="span"
+                    color={
+                      !action.disabled ? convertActionVariantToColor(action.variant) : 'inherit'
+                    }
```
remidej, line 305: could you check if just "color: inherit" would work here? (or even not setting color at all). We could maybe avoid having to write the color logic twice

simotae14: I can't set the color to undefined if it is that your comment, I tried if inherit work and it works because it inherit color from the variant of the wrapper

remidej: if inherit works here that means you only need to call convertActionVariantToColor(action.variant) once right? (in the parent Flex)

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -293,8 +293,17 @@ const DocumentActionsMenu = ({
...
+                  >
                     {action.icon}
                   </Box>
```
remidej, line 308: I ran it locally and it looks like just this would work:

```suggestion
                  <span>{action.icon}</span>
```

remidej (approved):

PR #20852, opened 2024-07-19
fix(review-workflows): invalidate document queries when stage & assignee is changed

### LINKED ISSUE #20745: RW - Stage just passed is not immediately displayed in Edit view or in list View (user has to refresh)

Feranchz:
### What does it do?

Make sure to invalidate document queries when RW stage or assignee is changed so we can see the change on the Edit View.

### How to test it?

1. Go to an entry with RW.
2. Change the stage of the entry.

Closes #20745

### DIFF
--- a/packages/core/content-manager/admin/src/services/documents.ts
+++ b/packages/core/content-manager/admin/src/services/documents.ts
@@ -169,6 +169,7 @@ const documentApi = contentManagerApi.injectEndpoints({
       }),
       providesTags: (result, _error, arg) => {
         return [
+          { type: 'Document', id: `ALL_LIST` },
           { type: 'Document', id: `${arg.model}_LIST` },
           ...(result?.results.map(({ documentId }) => ({
             type: 'Document' as const,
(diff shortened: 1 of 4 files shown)


simotae14 (commented):
I found another possible scenario, if you change the name of the review workflow step is not directly reflected on the document until you don't refresh the page, here an example

https://github.com/user-attachments/assets/2e555fcb-ddb4-454b-9966-5752bc98624a

Feranchz:
> I found another possible scenario, if you change the name of the review workflow step is not directly reflected on the document until you don't refresh the page, here an example
> 
>  Registrazione.schermo.2024-07-22.alle.10.33.46.mov

Great catch @simotae14, I needed to add a general tag to the Documents service so I can invalidate all the content types queries from review workflows. This because in some cases if you remove a Content Type from a Review Workflow you don't know the uid of the content type removed, so you need to invalidate all, same for the content type configuration (because it's used for the table headers)

simotae14 (approved):
The code seems ok, I haven't tested it

Issue #20764, opened 2024-07-16
[Design] IconButtons of the edit view are not squares and their icons are not centered


MarionLemaire:
https://www.notion.so/strapi/IconButtons-of-the-edit-view-are-not-squares-and-their-icons-are-not-centered-4d4429405d4f40f685205547bfec7c46?pvs=4

![CleanShot 2024-07-04 at 16 59 53@2x](https://github.com/user-attachments/assets/1294a5b7-6e29-41a4-a303-e604bb0c7389)

simotae14:
fixed with https://github.com/strapi/strapi/pull/20855

PR #20855, opened 2024-07-19
Fix the dimensions of the More button

simotae14:
### What does it do?

it fixes the dimensions of the More button in the Header, in the Document menu and in the Edit view

### Why is it needed?

Because the button needs to be a square with the More icon inside centered

### How to test it?

Open an entry and check
- the header
Schermata 2024-07-19 alle 17 34 51

- the Document menu
Schermata 2024-07-19 alle 17 35 09

- and the Release section in the Edit view
Schermata 2024-07-19 alle 17 35 40

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20764


madhurisandbhor:
I also made a change for this, but as its not from DS I let you merge

madhurisandbhor (approved):

Issue #20824, opened 2024-07-16
[Design] A cloud is cutted in half in Cloud’s plugin


MarionLemaire:
https://www.notion.so/strapi/A-cloud-is-cutted-in-half-in-Cloud-s-plugin-07fd9726bd664a768947f8c85c547a60?pvs=4

![CleanShot 2024-07-04 at 17 32 53@2x](https://github.com/user-attachments/assets/03680e52-e736-4318-bef7-8e429a9242e8)

simotae14:
fixed with the https://github.com/strapi/strapi/pull/20862 PR

PR #20862, opened 2024-07-22
fix: the left offset of the cloud image

simotae14:
### What does it do?

Fix a UI issue we had in the cloud plugin

### Why is it needed?

The Cloud background left image had the wrong offset so it seems cut in the layout

### How to test it?

- add the cloud plugin to your Strapi instance with an EE license and check the plugin home page

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20824


Mcastres (commented):
This is only working for Strapi 5. It won't work on Strapi 4 with the expandable sidenav 🤔

lucasboilly:
That's all good!

remidej (approved):

simotae14:
The issue opened is for v5 not v4, @Mcastres can you please open another GH issue for v4?

PR #20864, opened 2024-07-22
fix(upload): mock usefolderstructure to fix flaky frontend test

jhoward1994:
### What does it do?

Mock usefolderstructure to fix flaky frontend test

The isLoading state here 
`packages/core/upload/admin/src/components/BulkMoveDialog/BulkMoveDialog.jsx`

was causing a state update that failed this test case
https://github.com/strapi/strapi/blob/47fed6c2b2716a33bc5b44c7465d4be77be27a84/packages/core/upload/admin/src/pages/App/MediaLibrary/tests/MediaLibrary.test.jsx#L361-L362

### How to test it?

Run the tests. To check flaky tests like this I've been using a (pretty hacky) script that looks like

```
#!/bin/zsh
# multi-run.zsh

# COMMAND='npx nx run @strapi/upload:test:front --runInBand --skip-nx-cache'
COMMAND='yarn test:front packages/core/upload/admin/src/pages/App/MediaLibrary/tests/MediaLibrary.test.jsx'

TIMES=10
fail_count=0

for i in {1..$TIMES}; do
  echo "Attempt $i of $TIMES"
  if ! eval $COMMAND; then
    echo "Command failed on attempt $i"
    ((fail_count++))
  fi
done

echo "Command failed $fail_count times out of $TIMES attempts."
```

### DIFF
--- a/packages/core/upload/admin/src/pages/App/MediaLibrary/tests/MediaLibrary.test.jsx
+++ b/packages/core/upload/admin/src/pages/App/MediaLibrary/tests/MediaLibrary.test.jsx
@@ -67,6 +67,11 @@ jest.mock('../../../../hooks/useMediaLibraryPermissions');
 jest.mock('../../../../hooks/useFolders');
 jest.mock('../../../../hooks/useFolder');
 jest.mock('../../../../hooks/useAssets');
+jest.mock('../../../../hooks/useFolderStructure', () => ({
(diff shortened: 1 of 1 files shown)


simotae14 (approved):
It looks good to me, it is the unit test always failing. Thanks Jamie

PR #20866, opened 2024-07-23
fix: json fields not parsed properly

### LINKED ISSUE #20655: Strapi v5 rc: admin ui read entry issue (config2.doc || "").split is not a function

remidej:
### What does it do?

Makes sure the JSON fields are parsed properly when read from the database. It turns out sometimes, probably because of v4 to v5 migrations, they were stringified twice, so [they needed to be parsed twice](https://stackoverflow.com/questions/31662849/parse-json-object-that-stringify-twice).

I also made a fix on the frontend to support both string and object types, as both could be sent for users migrating.

### How to test it?

- Write JSON in a JSON field in the edit view, the experience should be normal
- Save and refresh, your changes should be retrieved
- Query the content using the content API, you should see your JSON input in object form, not as a string

### Related issue(s)/PR(s)

replaces #20829
fixes #20655

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Json.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Json.tsx
@@ -24,7 +24,9 @@ const JsonInput = forwardRef<JSONInputRef, InputProps>(
         <Field.Label action={labelAction}>{label}</Field.Label>
         <JSONInputImpl
           ref={composedRefs}
-          value={field.value}
+          value={
+            typeof field.value == 'object' ? JSON.stringify(field.value, null, 2) : field.value
+          }
           onChange={(json) => {
             // Default to null when the field is not required and there is no input value
             const value = required && !json.length ? null : json;
--- a/packages/core/database/src/fields/json.ts
+++ b/packages/core/database/src/fields/json.ts
@@ -8,12 +8,19 @@ export default class JSONField extends Field {
   fromDB(value: unknown) {
     try {
       if (typeof value === 'string') {
-        return JSON.parse(value);
+        const parsedValue = JSON.parse(value);
+
+        if (typeof parsedValue === 'string') {
+          return JSON.parse(parsedValue);
+        }
+
+        return parsedValue;
       }
     } catch (error) {
       // Just return the value if it's not a valid JSON string
       return value;
     }
+
     return value;
   }
 }
--- a/packages/core/database/src/fields/json.ts
+++ b/packages/core/database/src/fields/json.ts
@@ -2,7 +2,15 @@ import Field from './field';
 
 export default class JSONField extends Field {
   toDB(value: unknown) {
-    return JSON.stringify(value);
+    if (value == null) {
+      return null;
+    }
+
+    if (typeof value === 'object') {
+      return JSON.stringify(value);
+    }
+
+    return value;
   }
 
   fromDB(value: unknown) {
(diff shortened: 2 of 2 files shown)


Marc-Roig:
I am also curious to find out what caused this too, should we test the migration to avoid this data issue in db?

remidej:
@Marc-Roig you're right I looked more into the root cause, and think I patched it now. The frontend is now sending the data as a string, but we were stringifying at the db level regardless when saving

Marc-Roig (approved):
lets goo!! 🚀

shamglam:
@remidej we have a custom field of type json, and the input component of the custom field isn't receiving anything in the value property. 
Could this be related? Currently running on rc7

remidej:
@shamglam could you share the code of your custom field input? I don't think it's related to this PR but we can check

remidej:
It looks like this is a breaking change we failed to document. Value, setValue and error can now be retrieved from a hook, [like it's done there](https://github.com/strapi/strapi/blob/baabf9edef4d0079bdbe0f891dcba743afc5a571/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx#L79).

I think we should be able to provide backward compatibility though. I'll share that with the team. Thanks for your report!

shamglam:
@remidej - while we're at this - I've updated the code according to the link you provided, had a small question 
when calling onChange(name, val); - do i need to call JSON.stringify for the val object if the field type is json.

remidej:
@shamglam both stringified and non-stringified forms should be supported, so you can keep it as an object if that's what you're working with

Issue #19001, opened 2023-12-07
User ID shown in relation dropdown instead of the configured mainField


ekdennisek:
### Required System information

- Node.js version: v18.17.1
- NPM version: 9.6.7
- Strapi version: 4.15.5
- Database: PostgreSQL
- Operating system: Ubuntu 22.04.3 LTS
- Is your project Javascript or Typescript: TypeScript

### Describe the bug

Adding a relation to Users from a collection type causes the ID of the user to be rendered in the RelationInput input field rather than displaying what's set to be the `mainField` when editing an entry or creating new entries.

![image](https://github.com/strapi/strapi/assets/10733631/b45a6fa3-2d42-4569-a063-af07353e2d15)

### Steps to reproduce the behavior

1. Create a collection type

2. Add a relation to Users

![image](https://github.com/strapi/strapi/assets/10733631/4e8da570-e7c7-4381-8f82-5b18c855d799)

3. 'Configure the view' -> Set the entry title to the correct field

![image](https://github.com/strapi/strapi/assets/10733631/ee27ba3c-b99f-4f4e-8468-6eae0f3e95a0)

4. Create a new entry

![image](https://github.com/strapi/strapi/assets/10733631/b45a6fa3-2d42-4569-a063-af07353e2d15)

### Expected behavior

I expected a list of names to appear. 

* Jane
* John
* Micael
* Liza

And so on.

### Additional context

I've troubleshot this a little bit. The problem seems to lie in the `sanitizeMainField` function (whether it's because of the function itself or the permission system I haven't managed to figure out). Even though I'm logged in as a super admin, it doesn't think I have the proper permissions. 

[This check](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/controllers/relations.ts#L30) always return true, then 'id' is returned [here](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/controllers/relations.ts#L43) instead of the intended field `mainField`.

AndyCamicci:
Hello ! Same issue here, and same troubleshooting results as @ekdennisek (`permissionChecker.cannot.read(null, mainField)` returns true, even if i'm logged as super admin), tested on a fresh install, same results. Works fine with any other data type, but not for `admin::user`.

Adam-rGz:
Hello, same problem with relations, Instead of showing selected value, there are only ID's

sdaoudi:
Hello ! Same issue here with admin::user

[...]

PR #20878, opened 2024-07-25
RBAC Action Aliases - v5

Convly:
:warning:  **This PR targets v5. This will also be ported to v4.**

#### Action Aliases

Introduce a new `aliases` property on admin actions (`RBAC`).

This property is defined as an optional array of action alias.

```ts
interface ActionAlias {
  /**
   * The action ID to alias
   */
  actionId: string;

  /**
   * An optional array of subject to restrict the alias usage
   */
  subjects?: string[];
}

type Action = {
  // other properties ...

  /**
   * An optional array of @see {@link ActionAlias}.
   *
   * It represents the possible aliases for the current action.
   *
   * Aliases are unidirectionals.
   *
   * Note: This is an internal property and probably shouldn't be used outside of Strapi core features.
   *       Its behavior might change at any time without notice.
   *
   * @private
   */
  aliases?: ActionAlias[];
};
```

#### Action Provider

A new method `aliases` is added to the admin action provider (`RBAC`).

```ts
declare function aliases(actionId: string, subject?: string | null): string[];
```

It returns an array of `actionId` that can be used as aliases for the given `actionId` and `subject`.

- When an alias doesn't declare a `subjects` property, it only checks actions ID
- When an alias has an array of subjects declared, it ensures both the actions ID and subjects matches   

#### New Aliases

A default set of aliases for common use cases have been defined:

- `admin::users.read` is an alias of `plugin::content-manager.explorer.read`(`admin::user`)
- `admin::users.roles` is an alias of `plugin::content-manager.explorer.read`(`admin::role`)
- `plugin::i18n.locale.read` is an alias of `plugin::content-manager.explorer.read`(`plugin::i18n.locale`)

### Why is it needed?

The initial idea came when looking at #19001. The issue being hidden content-types (`admin::user`, `plugin::i18n.locale`, etc...) not having a content manager action being registered for them (only CTs being displayed in the CM have one).

Since it is possible to create a relation to `admin::user` (manual exception), this is why even super admin can't see the main field for the relation (because the needed permission doesn't even exist).

**Note:** It's also already possible to declare manual relations to any internal content-type from the schema file (e.g. `locale`)

This is a limitation in how permissions are defined. However, simpler solutions like only adding the hidden content-types (mostly internal ones like `upload::file` or `upload::folder`) to the list of allowed subjects for the `plugin::content-manager.explorer.read` action might lead to unwanted side effects (like adding them to the roles' edit permission table).

The proposed (internal) `aliases` feature mitigates the issue by adding more flexible options when checking permissions while limiting side effects on already existing features and behaviors. 

### How to test it?

1. Create a content-type with a relation to `admin::user` (`CT`)
2. Create a non super-admin user (`A`), go the content manager
3. From the super-admin account (`B`), set the following permissions for (`A`)'s role
  - In `Collection Types`: allow every CRUD operation on `CT`
  - In `Settings`: make sure `Read` is **not** checked in `Users & Roles`
4. As (`A`) Go to `Content-Manager` > `CT` > `Create new entry`
5. When opening the select input for the `admin::user` relation, only document IDs should be visible
6. As (`B`), update (`A`)'s role permissions, and check the `Read` box in `Settings` > `Users & Roles
7. As (`A`), try opening the select input again, the proper main field should be displayed (by default `firstname`)

### Related issue(s)/PR(s)

#19001

### Notes

- Aliases are unidirectional to allow setting more granular permissions
- The aliases defined in this PR are just meant to address common scenarios, but this is meant to be extended based on future use cases
- While this is set as an internal feature (for now), it could be interesting to imagine the potential use cases for plugin & user app if made public

### Screenshots

Screenshot from 2024-07-25 00-43-21
Screenshot from 2024-07-25 00-44-03


packages/core/content-manager/server/src/policies/hasPermissions.ts
```diff
@@ -7,6 +7,10 @@ const { createPolicy } = policy;
 export default createPolicy({
   name: 'plugin::content-manager.hasPermissions',
   validator: validateHasPermissionsInput,
+  /**
+   * NOTE: Action aliases are currently not checked at this level (policy).
+   *       If you want to add support for it, please create a dedicated RFC with a list of potential side effect this could have.
+   */
```
innerdvations, line 13: Should we say that it is actually intended? That is, because aliases are internal only, they will not be visible to policies.

innerdvations (approved):

innerdvations (commented):
Just a few questions, but overall works for me.

packages/core/utils/src/provider-factory.ts
```diff
@@ -31,36 +31,35 @@ export interface Options {
...
 type Item = Record<string, unknown>;
 
-export interface Provider {
+export interface Provider<T = unknown> {
   hooks: ProviderHooksMap;
-  register(key: string, item: Item): Promise<Provider>;
+  register(key: string, item: T): Promise<Provider>;
   delete(key: string): Promise<Provider>;
-  get(key: string): Item | undefined;
-  getWhere(filters?: Record<string, unknown>): Item[];
```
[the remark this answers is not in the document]
Convly: I don't think it has ever been considered public and isn't used internally. However, that's a point we can discuss further indeed.

Marc-Roig (approved):
great! 
This reminded me to this https://github.com/strapi/strapi/blob/7e6f107daf918a80b04e10fca1627c114f1d09ad/packages/core/content-manager/server/src/controllers/relations.ts#L36

We had hardcoded logic to show the u&p role if user had permissions to se the u&p user, but maybe we could use an alias instead 🤔

packages/core/admin/server/src/domain/action/index.ts
```diff
@@ -2,18 +2,77 @@ import type { Utils } from '@strapi/types';
...
+   * An optional array of @see {@link ActionAlias}.
+   *
+   * It represents the possible aliases for the current action.
+   *
+   * Aliases are unidirectionals.
+   *
+   * Note: This is an internal property and probably shouldn't be used outside of Strapi core features.
+   *       Its behavior might change at any time without notice.
+   *
+   * @private
```
alexandrebodin, line 73: isn't it @internal ?

alexandrebodin (changes requested):
Overall looks fine. 

To avoid making this into a stable feature we could prefix all of this with unstable_xxx for a litlle while to see if this approach makes sense for a bit longer. 

Will we safely be able to remove the prefix when we are happy with this ?

packages/core/admin/server/src/validation/common-validators.ts
```diff
@@ -110,7 +110,7 @@ export const permission = yup
         // @ts-expect-error yup types
         const action = getActionFromProvider(this.options.parent.action);
 
-        if (!action) {
+        if (!action || !subject) {
```
alexandrebodin, line 113: was this a bug ?

packages/core/utils/src/provider-factory.ts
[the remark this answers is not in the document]
alexandrebodin: Worth adding it to the list of BC just in case

alexandrebodin (approved):

packages/core/admin/server/src/domain/action/index.ts
Convly: Yep

packages/core/admin/server/src/validation/common-validators.ts
Convly: It was actually just a TS issue, but adding the initial check might just introduce a bug, I've reworked it

Convly:
> strapi/packages/core/content-manager/server/src/controllers/relations.ts

@Marc-Roig What do you think about adding an alias from `plugin::users-permissions.roles.read` to `plugin::content-manager.explorer.read` (`plugin::users-permissions.role`)?

packages/core/admin/server/.eslintrc
```diff
@@ -2,6 +2,6 @@
   "root": true,
   "extends": ["custom/back/typescript"],
   "parserOptions": {
-    "project": ["./server/tsconfig.eslint.json"]
```
alexandrebodin, line 5: isn't this necessary as the config is based on the root folder of the package ? Can you check what we do elsewhere ?

Convly: Only 4 of them were doing this like that and it was breaking the auto parsing in the IDE, so I pushed that to see if it still works in the CI. I'm following this closely :slightly_smiling_face:

Convly: Yep, it seems it's failing in the CI now. I'll have to go without eslint locally, I guess :upside_down_face: 

I'll revert the changes

Convly:
What do you want to do about the changes in the permissions behavior that break this test @Marc-Roig  (as per discussed internally)

https://github.com/strapi/strapi/blob/8ad1ccbb99f2d91df1b816e62f98652faeaf3bf3/tests/api/core/content-manager/api/relations-permissions.test.api.js#L183-L197

Marc-Roig:
@Convly the test mod looks good to me :)

alexandrebodin (approved):

innerdvations (approved):
Still LGTM!

Issue #20298, opened 2024-05-11
strapi-5@0.1.0 error: Cannot find module '@strapi/email/package.json' on first start 


rbjs:
### Required System information
- Node.js version: v20.12.2
- NPM version: 10.5.2
- Strapi version: 5.0.0-beta.7
- Database: n/a
- Operating system: macOS 13.6.4
- Is your project Javascript or Typescript: n/a
### Describe the bug

'npm run develop' returns fatal error on first run after fresh installation.

> strapi-5@0.1.0 develop
> strapi develop

⠋ Loading Strapi⠋ Building build context
[INFO] Including the following ENV variables as part of the JS bundle:
    - ADMIN_PATH
    - STRAPI_ADMIN_BACKEND_URL
    - STRAPI_TELEMETRY_DISABLED
✔ Building build context (47ms)
✔ Creating admin (378ms)
[ERROR]  There seems to be an unexpected error, try again with --debug for more information
`                                                                               
│   Error: Cannot find module '@strapi/email/package.json' 
│   Require stack:  
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/loaders/plugins/get-enabled-plugins.js 
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/loaders/plugins/index.js
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/loaders/index.js 
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/providers/registries.js 
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/providers/index.js 
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/Strapi.js 
│   - /Users/dev/strapi5/node_modules/@strapi/core/dist/index.js 
│   - /Users/dev/strapi5/node_modules/@strapi/strapi/dist/cli/commands/admin/create-user.js 
│   - /Users/dev/strapi5/node_modules/@strapi/strapi/dist/cli/commands/index.js 
│   - /Users/dev/strapi5/node_modules/@strapi/strapi/dist/cli/index.js 
│   - /Users/dev/strapi5/node_modules/@strapi/strapi/bin/strapi.js 
│   at Module._resolveFilename (node:internal/modules/cjs/loader:1143:15)           
│   at Module._load (node:internal/modules/cjs/loader:984:27)                       
│   at Module.require (node:internal/modules/cjs/loader:1231:19)                    
│   at require (node:internal/modules/helpers:179:18)
[...]

PR #20879, opened 2024-07-25
fix: resolve internal plugins from the strapi pkg

alexandrebodin:
### What does it do?

Tentative fix for https://github.com/strapi/strapi/issues/20298

### How to test it?

Create a new app with the experiemental linked below

### Related issue(s)/PR(s)

Closes https://github.com/strapi/strapi/issues/20298

### DIFF
--- a/packages/core/core/src/loaders/plugins/get-enabled-plugins.ts
+++ b/packages/core/core/src/loaders/plugins/get-enabled-plugins.ts
@@ -90,7 +90,13 @@ export const getEnabledPlugins = async (strapi: Core.Strapi, { client } = { clie
 
   for (const dep of INTERNAL_PLUGINS) {
     const packagePath = join(dep, 'package.json');
-    const packageInfo = require(packagePath);
+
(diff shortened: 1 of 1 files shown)


alexandrebodin:
Prerelease job: https://github.com/strapi/strapi/actions/runs/10091469460

innerdvations (commented):
I haven't been able to recreate the actual bug in a real world situation, so let's wait to see what the results are from users reporting it before merging.

Code looks good. I don't like the implicit/circular dependency that this leaves in place, so I'll still keep a task open to later restructure how the plugin loading is handled.

alexandrebodin:
> I haven't been able to recreate the actual bug in a real world situation, so let's wait to see what the results are from users reporting it before merging.
> 
> Code looks good. I don't like the implicit/circular dependency that this leaves in place, so I'll still keep a task open to later restructure how the plugin loading is handled.

Yeah I have this as v6 refactorings happy to give you the details

innerdvations (approved):
Anyway, LGTM!

Issue #20585, opened 2024-06-21
[Review Workflows] User needs to refresh when changing stage to see the new one


MarionLemaire:
### Required System information

Node.js version: 20.9.0
Strapi version: 5.0.0-beta.13
Operating system: Windows WSL and OS

### Describe the bug

When you have a RW associated to a CT, when changing stage for an entry of this CT, even if you have a confirmation message, the stage displayed is still the initial one. You have yo refresh the page to see the updated stage.

### Steps to reproduce the behavior

1. Go to Settings > Review Workflows
2. Associate a workflow to a CT
3. Go to CM > create an entry for this CT
4. Try to change stage : the displayed stage is still the initial one even if you get a confirmation message

### Expected behavior

The displayed stage should be the last selected one.

### Screenshots

https://github.com/strapi/strapi/assets/97893822/80031b95-35f9-4027-ad41-f256c019c00b

PR #20880, opened 2024-07-25
fix: review workflows cache issues

remidej:
### What does it do?

Fixes some review workflow issues:

- The cache is now updates when changing an entry's stage. I'm using RTK's invalitateTags method for it to be more declarative. Unfortunately I need to silence a TS issue, but I think it's acceptable considering the previous solution called dispatch directly, which provided no type safety
- Clears document cache when a workflow is updated or created, so that the workflow changes are visible on the edit view and the list view
- Fixes an issue in the modal warning you that creating new workflow means some content types will lose their original workflow. It got the count of affected content types wrong (always said 0)
- Fixes that same modal which never worked when updating a workflow, because it relied on data from the other workflows, and other workflows were not actually fetched.

### How to test it?

- create a review workflow, assign it to a content type
- go to that content type, you should see the workflow information, both in edit and list views
- edit the review workflow (e.g. change the stage name)
- go the the CM again, the edit and list views should reflect the changes
- create a new workflow and assign it the same content type
- when saving, you should see a warning that 1 content type will be unlinked from its previous workflow
- go back to the first workflow, assign it the same content type again
- you should see the same modal

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20585

### DIFF
--- a/packages/core/review-workflows/admin/src/routes/settings/:id.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/:id.tsx
@@ -119,7 +119,7 @@ const EditPage = () => {
     error,
     update,
     create,
-  } = useReviewWorkflows({ id: isCreatingWorkflow ? undefined : id });
+  } = useReviewWorkflows();
   const permissions = useTypedSelector(
     (state) => state.admin_app.permissions['settings']?.['review-workflows']
   );
@@ -188,7 +188,7 @@ const EditPage = () => {
         if ('error' in res && isBaseQueryError(res.error) && res.error.name === 'ValidationError') {
           helpers.setErrors(formatValidationErrors(res.error));
         } else if ('data' in res) {
-          navigate(`../${res.data.id}`);
+          navigate(`../${res.data.id}`, { replace: true });
         }
       }
     } catch (error) {
@@ -397,7 +397,7 @@ const EditPage = () => {
                         {
                           count:
                             contentTypesFromOtherWorkflows?.filter((contentType) =>
-                              currentWorkflow?.contentTypes?.includes(contentType)
+                              values.contentTypes.includes(contentType)
                             ).length ?? 0,
                         }
                       )}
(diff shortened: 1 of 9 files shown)


packages/core/review-workflows/admin/src/routes/settings/:id.tsx
```diff
@@ -113,15 +112,14 @@ const EditPage = () => {
...
-  const dispatch = useDispatch();
   const {
     isLoading: isLoadingWorkflow,
     meta,
     workflows,
     error,
     update,
     create,
-  } = useReviewWorkflows({ id: isCreatingWorkflow ? undefined : id });
+  } = useReviewWorkflows();
```
jhoward1994, line 122: Now we've removed ID from here doesn't it make this whole route redundant?

https://github.com/strapi/strapi/blob/ee7979c6d5a886a44ba3e11aebb41c417f30caa7/packages/core/review-workflows/server/src/routes/review-workflows.ts#L75-L91

If in every case in the FE we need access to all workflows maybe we should
remove this route?

jhoward1994:
I think we can remove errors here? https://github.com/strapi/strapi/blob/ee7979c6d5a886a44ba3e11aebb41c417f30caa7/packages/core/review-workflows/server/src/controllers/workflows.ts#L5-L6

jhoward1994:
https://github.com/user-attachments/assets/45fef0ea-f66f-45a6-871b-499b9687d611

When I update the stage the content of the stages drop down component doesn't reflect the change. So when I only have 2 stages in the workflow I can't actually move from the stage until I refresh

remidej:
@jhoward1994 thx for the video, should be fixed now

jhoward1994 (approved):

jhoward1994:
Just one other thing but unrelated to this PR I think

When a workflow only has one stage the UI says "no permission to change the stage"

I think we should update that to something more relevant or insist that workflows have > 1 stage

image

remidej:
Agreed, but I do think it's another issue, I'll share that with the team. Thanks!

PR #20883, opened 2024-07-25
FIX - Email not found on forgot password

hnrqsss:
### What does it do?

Returns an error when email not found on forgotPassword lifecycle.

### Why is it needed?

Because the strapi response when email is finded as the same when it not find

### How to test it?

Just send a POST request to /api/auth/forgot-password with a registered user and one not registered user.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/plugins/users-permissions/server/controllers/auth.js
+++ b/packages/plugins/users-permissions/server/controllers/auth.js
@@ -26,7 +26,7 @@ const {
 } = require('./validation/auth');
 
 const { getAbsoluteAdminUrl, getAbsoluteServerUrl, sanitize } = utils;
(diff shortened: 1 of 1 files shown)


hnrqsss:
> Describe the technical changes you did.

Returns an error when email not found on forgotPassword lifecycle.

Ohiekkar:
It is a security feature to always return a consistent message for both existent and non-existent accounts in order to prevent user enumeration attacks.

derrickmehaffy:
> It is a security feature to always return a consistent message for both existent and non-existent accounts in order to prevent user enumeration attacks.

Completely agree, we won't merge this as it would be a security risk.

Marking as closed

PR #20892, opened 2024-07-29
fix: draft validations

Marc-Roig:
### What does it do?

Modifies backend validators so, only max validators are applied on draft entries.

This was identified and discussed with the product team during our QA sessions, and we concluded it makes sense to exclude all validations on drafts except the max length/value ones.

This PR only includes de BE side, but will require adapting the FE to not trigger the validations on draft entries.

### DIFF
--- a/packages/core/core/src/services/entity-validator/validators.ts
+++ b/packages/core/core/src/services/entity-validator/validators.ts
@@ -1,3 +1,13 @@
+/**
+ * Validators check if the entry data meets specific criteria before saving or publishing.
+ * (e.g., length, range, format).
+ *
+ * Drafts have limited validations (mainly max constraints),
+ * while published content undergoes full validation.
+ *
+ * The system also takes locales into account when validating data.
+ * E.g, unique fields must be unique within the same locale.
+ */
 import _ from 'lodash';
 import { yup } from '@strapi/utils';
 import type { Schema, Struct, Modules } from '@strapi/types';
@@ -377,7 +387,9 @@ const addUniqueValidator = <T extends yup.AnySchema>(
     }
 
     // The validation should pass if there is no other record found from the query
-    return !(await strapi.db.query(model.uid).findOne({ where: scalarAttributeWhere }));
+    return !(await strapi.db
+      .query(model.uid)
+      .findOne({ where: scalarAttributeWhere, select: ['id'] }));
   });
 };
 
--- a/packages/core/core/src/services/entity-validator/validators.ts
+++ b/packages/core/core/src/services/entity-validator/validators.ts
@@ -80,8 +80,9 @@ const addMinIntegerValidator = (
     attr,
   }: {
     attr: Schema.Attribute.Integer | Schema.Attribute.BigInteger;
-  }
-) => (_.isNumber(attr.min) ? validator.min(_.toInteger(attr.min)) : validator);
+  },
+  { isDraft }: ValidatorOptions
+) => (_.isNumber(attr.min) && !isDraft ? validator.min(_.toInteger(attr.min)) : validator);
 
 /**
  * Adds max integer validator
@@ -104,8 +105,9 @@ const addMinFloatValidator = (
     attr,
   }: {
     attr: Schema.Attribute.Decimal | Schema.Attribute.Float;
-  }
-) => (_.isNumber(attr.min) ? validator.min(attr.min) : validator);
+  },
+  { isDraft }: ValidatorOptions
+) => (_.isNumber(attr.min) && !isDraft ? validator.min(attr.min) : validator);
 
 /**
  * Adds max float/decimal validator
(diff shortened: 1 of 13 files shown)


Marc-Roig:
Small note, this is not testable from the FE side yet, we need to remove those validations on the UI side too

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -80,8 +80,9 @@ const addMinIntegerValidator = (
     attr,
   }: {
     attr: Schema.Attribute.Integer | Schema.Attribute.BigInteger;
-  }
-) => (_.isNumber(attr.min) ? validator.min(_.toInteger(attr.min)) : validator);
+  },
+  { isDraft }: ValidatorOptions
+) => (_.isNumber(attr.min) && !isDraft ? validator.min(_.toInteger(attr.min)) : validator);
```
remidej, line 85: Why don't you do the same isDraft check for addMaxIntegerValidator below?

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -104,8 +105,9 @@ const addMinFloatValidator = (
     attr,
   }: {
     attr: Schema.Attribute.Decimal | Schema.Attribute.Float;
-  }
-) => (_.isNumber(attr.min) ? validator.min(attr.min) : validator);
+  },
+  { isDraft }: ValidatorOptions
+) => (_.isNumber(attr.min) && !isDraft ? validator.min(attr.min) : validator);
```
remidej, line 110: same question, why do it for min and not max?

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -380,8 +377,7 @@ const addUniqueValidator = <T extends yup.AnySchema>(
     }
 
     // The validation should pass if there is no other record found from the query
-    const queryUid = model.uid;
-    return !(await strapi.db.query(queryUid).findOne({ where: scalarAttributeWhere }));
+    return !(await strapi.db.query(model.uid).findOne({ where: scalarAttributeWhere }));
```
remidej, line 380: It probably doesn't make much of a difference, but maybe selecting only the id field could make the query faster?

remidej (commented):
It would be nice to add some comments in the code explaining the expected behavior. Just looking at the code it's not clear _why_ we're doing these draft/published checks

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -336,12 +337,9 @@ const addUniqueValidator = <T extends yup.AnySchema>(
...
 
     /**
-     * If we are updating a draft and the value is unchanged we skip the unique verification. This will
-     * prevent the validator to be triggered in case the user activated the
-     * unique constraint after already creating multiple entries with
-     * the same attribute value for that field.
+     * We don't validate any unique constraint for draft entries.
      */
-    if (!isPublish && value === entity?.[updatedAttribute.name]) {
+    if (options.isDraft) {
```
jhoward1994, line 342: Are we agreed that uniqueness doesn't apply on drafts now? I'd be okay with that
if it's the way we want to go but when we build uniqueness for v5 we agreed on
the behaviour you can see in JIRA: DX-1422, CONTENT-2117

packages/core/core/src/services/entity-validator/validators.ts
Marc-Roig: will add some comments explaining the behaviour 🙌 
only max validations will apply on drafts (for number and text fields)

packages/core/core/src/services/entity-validator/validators.ts
Marc-Roig: sure!

packages/core/core/src/services/entity-validator/validators.ts
Marc-Roig: yep! Our original design included draft unique validations, after doing a blitz session the PMs & Aurelien concluded drafts should not have those validations anymore.

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -1,3 +1,13 @@
+/**
+ * Validators check if the entry data meets specific criteria before saving or publishing.
+ * (e.g., length, range, format).
+ *
+ * Drafts have limited validations (mainly max constraints),
```
remidej, line 5: I guess this is more of a product question just for my curiosity, but do you know why we decided to only check max for drafts?

remidej (approved):

packages/core/core/src/services/entity-validator/validators.ts
Marc-Roig: I believe the idea is for the user to save an "unfinished" version of an entry, that means being able to save with everything empty and with the minimum friction. Max validations should not be an intrusive validation while drafting an entry, compared to the rest that is

amirhosseinNouri:
Hi guys.
Our marketing team uses Strapi for content management. Sometimes, they when they prepare the content, before publishing it, they need to check the overall result on FE. We've implemented draft mode using NextJS draft mode feature in this regard. 
The problem is the validation logic is not executed in draft state. As a result, when they add some components and due to invalid/incomplete data, they are not able to see the result.

Is there any way to have those validations in draft state too?
Thanks for your support

Issue #17067, opened 2023-06-22
Discord Auth not assigning username correctly


mallowsc:
- Node.js version: 18.15.0
- NPM version: 9.5.0
- Strapi version: 4.11.2
- Database: PostgresSQL
- Operating system: Ubuntu Server
- Is your project Javascript or Typescript: Javascript

### Describe the bug

Discord is removing discriminator soon and Strapi does not update to the new username.  In addition, users that have changed their Discord name have a #0 at the end of their name.  This would not be needed. Another issue that would help is adding the Discord ID of a user to the user table.

### Code snippets

https://github.com/strapi/strapi/blob/553217bd1cd9b63a2c144e1aee7543810a478fc1/packages/plugins/users-permissions/server/services/providers-registry.js#L58

mallowsc:
I added these lines of code to a branch I have but did not know the proper way to continue so I talked to Boegie and Boegie mentioned to submit an issue.  The below is what I added to start.
![image](https://github.com/strapi/strapi/assets/96073136/0c6fe283-6e87-4896-b46c-bd3481920ccf)
![image](https://github.com/strapi/strapi/assets/96073136/b083eef8-be5c-45bc-ab19-79482e60b9aa)

ahzia:
@Boegie19 @derrickmehaffy I want to work on this issue and create a PR, Is it valid?

derrickmehaffy:
Sure, please feel free

Aditya8840:
I want to work on this issue, if still open

mallowsc:
Hi @derrickmehaffy .  I wanted to follow up on this.  I know it's been awhile and fix 18115 was abandoned due to the previous issues with 4.15.  Do you think we could get this implemented soon?

derrickmehaffy:
I requested some changes to the PR author as is we can't merge that but there are few things that need to be changed.

mallowsc:
It looks like we are back to square one.  When I get a chance, is it possible that I can look at correcting this?

derrickmehaffy:
> It looks like we are back to square one.  When I get a chance, is it possible that I can look at correcting this?

Sure please feel free to submit a PR

mallowsc:
10-4.  I did this.  Hopefully I did this correctly.

PR #20900, opened 2024-07-29
Fix: #17067 Discord Auth not assigning username correctly

mallowsc:
What does it do?
Combine username and discriminator only if discriminator exists and isn't 0. A discriminator of 0 means the user has converted their username.

Why is it needed?
Discord made a system change to remove the discriminator and now has unique usernames.  Due to this change, the discriminator is not needed unless a user has not changed their username.  In the future, the discriminator check will not be needed at all.

How to test it?
Login with Discord and you will only get your Discord username if you have it set.

Related issue(s)/PR(s)
fixes https://github.com/strapi/strapi/issues/17067


derrickmehaffy (approved):
LGTM

PR #20911, opened 2024-07-31
shard E2E tests

Boegie19:
### What does it do?

Shard e2e tests 

### Why is it needed?

To lower the time it takes for CI to run

### How to test it?

Run the CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -161,6 +161,7 @@ jobs:
       fail-fast: false
       matrix:
         project: ['chromium', 'webkit', 'firefox']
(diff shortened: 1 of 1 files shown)


Boegie19:
Just found out that what you guys did with --project did not work so you ran all tests 3 times.

Boegie19:
https://github.com/strapi/strapi/actions/runs/10180562773 time lowered back down to 15 minutes

Boegie19:
Also can we run unit tests and all other tests at the same time that way we are not waiting 5min just for unit frontend to be done. if not I want to shard the unit tests.

PR #20914, opened 2024-07-31
shard ee tests+ fix broken platform

Boegie19:
### What does it do?

### Why is it needed?

Fixes that platform works and add sharding

### How to test it?

See CI logs

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -161,6 +161,7 @@ jobs:
       fail-fast: false
       matrix:
         project: ['chromium', 'webkit', 'firefox']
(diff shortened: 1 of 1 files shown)


innerdvations:
Nice catch, thanks! I will take a look at it this week but seems straightforward

Boegie19:
@innerdvations I think it would be worth not waiting for the unit tests to do the api and end to end tests if we do it all in parallel it would be 10minutes instead of 15min since we are waiting for 5 minutes on the unit tests.

innerdvations (approved):
LGTM!

If you want to make another PR to flatten the workflow deps, we can take a look to see if it speed things up

Issue #20903, opened 2024-07-30
You cannot uncheck the box to select or unselect an audio file


yanniskadiri:

MarionLemaire:
Tested in 5.0.0 

Actually I could check the boxes for audio files even if they were half hidden (see video)

https://github.com/user-attachments/assets/e4efc12f-7407-49c1-8950-68e4f32e7269


Steps to reproduce : 

1. Go to MediaLib
2. Upload an audio file
3. Try to select it with the checkbox

archanaagivale30:
Suggestion: can we place audio play at the bottom of the card to solve the overlap issue of the checkbox and the edit button(which occurs on hover).  Screenshots for reference,

![Image](https://github.com/user-attachments/assets/bd815d20-2ec4-4fda-9cbc-50cdb1272270)

Padding Will solve the overlap issue for a particular screen.

![Image](https://github.com/user-attachments/assets/1522ce9b-e8d6-4240-8d1c-ffd7a96e4c67)

@alexandrebodin can you please let me know your thoughts?

alexandrebodin:
@lucasboilly Do you mind sharing what this should look like ? @archanaagivale30 if you are interested in contributing let us know 🙏

archanaagivale30:
@alexandrebodin yes I am interested in contributing to this fix just need a final approach from you.

yanniskadiri:
@lucasboilly I think the proposal from @archanaagivale30, ie placing the audio play at the bottom of the card, is good. Can you confirm if you are aligned?

archanaagivale30:
@lucasboilly @yanniskadiri can you please confirm this?

yanniskadiri:
Hey @archanaagivale30, yes it's confirmed. Good news there is someone else who already submitted a PR to fix it here: https://github.com/strapi/strapi/pull/22098#issuecomment-2471213236

lucasboilly:
@yanniskadiri the PR you mentioned fix this issue because the checkbox becomes accessible, but it would be indeed even better if we were moving the player, as I explained in [this comment](https://github.com/strapi/strapi/pull/22098#issuecomment-2470468516).

> Regarding the audio file, that's great but I think it would be even better to center it vertically because the edit button and the checkbox (in the Media Library) are hiding parts of it.

@archanaagivale30 would it be possible to center it instead of moving it to the bottom? Or do we really need to have it to the bottom to be 100% sure it won't hide the checkbox and the edit button?

[...]

PR #20921, opened 2024-08-01
[Bug]: fix#20903 by decreasing the width of a container

KISMAT1234:
### What does it do?

Changes the width of the container and the left padding too

### Why is it needed?

To make able to uncheck the select button 

### Related issue(s)/PR(s)

issue no #20903


alexandrebodin:
Hello @KISMAT1234  Your pr is targeting v4 but you were originally trying to solve a v5 issue. If you want to you can resubmit your pr against `develop` to target v5.

alexandrebodin:
Hi, closing this PR for now for inactivity. It doesn't seem to solve the issue so we will take another approach

Thank you for your contribution 🙏

Issue #20598, opened 2024-06-21
[Content Manager] Wrong behaviour of UID field status


MarionLemaire:
### Required System information

Node.js version: 20.9.0
Strapi version: 5.0.0-beta.13
Operating system: Windows WSL

### Describe the bug

UID field displays “Unavailable” in a saved entry.

### Steps to reproduce the behavior

1. Create a CT with UID field
2. Go to CM, create an entry for this CT
3. Fill in UID field
4. Save this entry
5. Check the entry : UID status says "unavailable"

### Expected behavior

UID status shouldn't display unavailable (must not display anything when entry is already saved maybe ?)

### Screenshots

My only entry for this CT is saved : 
![image](https://github.com/strapi/strapi/assets/97893822/675d11c0-b345-4c67-b544-0f5bddbf75e6)

Still the status of UID field says it's unavailable : 
![image](https://github.com/strapi/strapi/assets/97893822/ce8a762e-aeae-4756-98f6-703844153416)

oiorain:
Confirmed in rc.7.  
But it seems to work at first when you create the entry and then not work anymore. i'm not sure but there is still a problem with uid showed as unavailable when it should be available.

remidej:
done in #20922 and #20985

PR #20922, opened 2024-08-01
fix: uid availability check

remidej:
### What does it do?

Fixes about the UID field availibility check:
- the check should not run if you haven't changed the value of the uid field
- the availability cache should be cleared when you save or update a content type

Also fixes the initialValue from the useField hook which was broken

### How to test it?

- Create a content type with a UID field
- Save an entry with a value in the UID field
- You should no longer see "available" or "unavailable" when nothing was changed
- Create a new entry in the same content type
- If you enter the same UID as the first entry, it should say unavailable. Anything else should be available

### Related issue(s)/PR(s)

fixes [#20598](https://github.com/strapi/strapi/issues/20598)


madhurisandbhor (commented):
https://github.com/user-attachments/assets/7f158c9e-b122-404a-b8f0-39d1552dc839

Saving empty value doesn't clear unavailability check

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/tests/UID.test.tsx
```diff
@@ -92,61 +102,56 @@ describe('UIDInput', () => {
...
-      timeout: 10000,
+    await waitFor(() => expect(screen.queryByTestId('loading-wrapper')).not.toBeInTheDocument());
+    expect(await screen.findByText(/^Available$/)).toBeInTheDocument();
+
+    // Change the value to make it unavailable
+    await user.type(input, 'unavailable');
+    act(() => {
+      jest.advanceTimersByTime(4000);
     });
+    expect(await screen.findByText(/^Unavailable$/)).toBeInTheDocument();
```
madhurisandbhor, line 141: Could you please use different uid than `available` and `unavailable`, I am not sure why your unavailability check is on here?

remidej: I can see how the uid is confusing so I updated it.

Regarding the check itself, I'm testing the general availability check behavior, so it makes sense to me to check both outcomes, the positive and negative results

madhurisandbhor (commented):
I am still unable to save empty uid value(if its not required)

https://github.com/user-attachments/assets/f5a9e000-c738-4821-bba1-0e82507b8061

remidej:
I'm opening a separate PR for the unique check (comes from the backend and also affects normal strings, not just UIDs)

PR #20926, opened 2024-08-01
chore: delete api tests for deleted review workflows route

remidej:
### What does it do?

Fixes API tests. Hopefully the CI passes now.

My earlier PR (#20880) deleted a review workflows endpoint as it was no longer used by the frontend (GET individual workflow). I didn't see that it had associated API tests. This PR deletes those left over tests


innerdvations (approved):
LGTM, let's see if tests all pass!

PR #20929, opened 2024-08-02
chore: use "workspace:*" version for all internal dependencies

remidej:
### What does it do?

Uses [a Yarn v3 cross reference](https://yarnpkg.com/features/workspaces#cross-references) for all dependencies to packages within the monorepo.

Also sets the version of our example apps to 0.0.0, so we don't feel like we need to keep them in sync with the rest.

For context I used a vscode find an replace with a regex: `"((?!(?:version))[^"]+)": "(5\.0\.0-rc\.8)"` => `"$1": "workspace:*"`
### Why is it needed?

It should make our release process smoother. We frequently get conflicts when merging the release branch because of package.json updates. Now that the dependencies won't change, this shouldn't happen anymore.

### How to test it?

However you want, nothing should change


examples/experimental-dev/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "experimental-dev",
-  "version": "5.0.0-rc.8",
```
alexandrebodin, line 3: usually this get auto bumped during release, with the version being 0.0.0 isn't it breaking anything. Can you try running an experimental release on this branch to see if everything works ?

remidej: In the couple of releases I did, those weren't upgraded automatically, but I felt the need to fix that manually because they were referencing previous v5 RC versions.

I just did an experimental release (locally, not github actions) and the result is odd, it didn't error but only upgraded the package.json for lerna.json and 6 package.jsons 🤔

remidej: Releasing via github actions seems to work as expected

alexandrebodin: Ok that's odd, Let's merge it and try it in real condition with the next v5 rc, worst case is we revert and do another rc

alexandrebodin (approved):

PR #20941, opened 2024-08-05
fix: policies registry returns same data types as other registries

### LINKED ISSUE #20893: CLI command `strapi policies:list` is broken
In a Strapi v5 application, the `strapi policies:list` throws an error:

```txt
TypeError: list.forEach is not a function
    at action (/mystrapi/packages/core/strapi/dist/cli/commands/policies/list.js:18:8)
```

You can test in the getstarted directory with this Yarn:

```sh
yarn strapi policies:list
```

innerdvations:
### What does it do?

- fixes policies registry to return the same data types as controllers, services, etc

### Why is it needed?

fixes the `strapi policies:list` command and makes the policies `keys()` return type consistent with other registries

### How to test it?

Policies should all work fine
especially check that `strapi policies:list` should return policies

### Related issue(s)/PR(s)

DX-1505 

Fixes #20893

### DIFF
--- a/packages/core/core/src/registries/policies.ts
+++ b/packages/core/core/src/registries/policies.ts
@@ -1,5 +1,5 @@
 import { pickBy, has, castArray } from 'lodash/fp';
-import type { Core, UID } from '@strapi/types';
+import type { Core } from '@strapi/types';
 import { addNamespace, hasNamespace } from './namespace';
 
 const PLUGIN_PREFIX = 'plugin::';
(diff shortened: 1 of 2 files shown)


packages/core/core/src/registries/policies.ts
```diff
@@ -108,7 +109,22 @@ const policiesRegistry = () => {
      * Returns a map with all the policies in a namespace
      */
     getAll(namespace: string) {
-      return pickBy((_, uid) => hasNamespace(uid, namespace))(Object.fromEntries(policies));
+      const filteredPolicies = pickBy((_, uid) => hasNamespace(uid, namespace))(
+        Object.fromEntries(policies)
+      );
+
+      // Return an enumerable object of the same format as controllers, services, etc
```
alexandrebodin, line 116: I don't see why this is necessary ? the picKBy should already return a enumerable object 🤔

innerdvations: Oops, you're right, that one can be left alone I think.

alexandrebodin:
I did not QA anything but the code looks fine. I'm not sure we actually use policies.keys() anywhere ?

innerdvations:
We only use it one place in our own code

Convly (approved):
LGTM

PR #20942, opened 2024-08-05
chore: aggregate test result to add a single required status check an…

alexandrebodin:
### What does it do?

Add an aggregated test result check to be able to make just this one required in github rulesets

### Why is it needed?

To not have to manually add each new check in the required lists. Also supports matrixes

### How to test it?

Break tests or not to see the final result change

### Related issue(s)/PR(s)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -523,3 +523,34 @@ jobs:
           dbOptions: '--dbclient=sqlite --dbfile=./tmp/data.db'
           runEE: true
           jestOptions: '--shard=${{ matrix.shard }}'
+
+  test_result:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    name: aggregate_test_result
+    needs:
+      [
+        pretty,
+        lint,
+        build,
+        typescript,
+        unit_back,
+        unit_front,
+        e2e_ce,
+        e2e_ee,
+        cli,
+        api_ce_pg,
+        api_ce_mysql,
+        api_ce_mysql_5,
+        api_ce_sqlite,
+        api_ee_pg,
+        api_ee_mysql,
+        api_ee_sqlite,
+      ]
+    steps:
+      - run: exit 1
+        if: >-
+          ${{
+                contains(needs.*.result, 'failure')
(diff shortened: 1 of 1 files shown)


alexandrebodin:
> Not an expert on workflows but it looks good to me. I assume it will work fine on workflows that don't exist on a branch (like api_ce_mysql_5 on v5/main) as long as they don't report as error.

I'll have to change the code to remove the job of the CI will not even start saying I'm referencing a unknown job :)

.github/workflows/tests.yml
```diff
@@ -523,3 +523,34 @@ jobs:
...
+        build,
+        typescript,
+        unit_back,
+        unit_front,
+        e2e_ce,
+        e2e_ee,
+        cli,
+        api_ce_pg,
+        api_ce_mysql,
+        api_ce_mysql_5,
```
Convly, line 544: Why don't we run the mysql_5 on ee too by the way?

.github/workflows/tests.yml
```diff
@@ -523,3 +523,34 @@ jobs:
...
+        api_ce_sqlite,
+        api_ee_pg,
+        api_ee_mysql,
+        api_ee_sqlite,
+      ]
+    steps:
+      - run: exit 1
+        if: >-
+          ${{
+                contains(needs.*.result, 'failure')
```
Convly, line 554: Is that the official way of checking for failure on previous tests? :sweat_smile: 
Is there a way to get exit code or status instead?

.github/workflows/tests.yml
alexandrebodin: I don't know. considering we have dropped it in v5 and it will soon become the main branch I don't think it's worth doing anything about that :)

.github/workflows/tests.yml
alexandrebodin: nope you only get a status 🤷

Convly (approved):

Convly:
Just a quick thought, but it could be nice to have a page dedicated to PRs in the contributor documentation. Something explaining which tests are expected to pass and what's the point of this check/how to add one to the "required" list

PR #20949, opened 2024-08-06
test(cli): add list command tests

innerdvations:
### What does it do?

- adds strapi *:list tests
- fix cli test script to allow running individual tests (same as e2e tests)
- upgrades cli-table3

### Why is it needed?

Adds tests and test functionality

### How to test it?

`yarn test:cli`

### Related issue(s)/PR(s)

DX-1524

adds additional test for https://github.com/strapi/strapi/pull/20941


tests/scripts/run-cli-tests.js
```diff
@@ -20,7 +20,7 @@ yargs
      * This lets us pass any other arguments to the test runner
      * e.g. the name of a specific test or the project we want to run
      */
-    'unknown-options-as-args': true,
+    'unknown-options-as-args': false,
```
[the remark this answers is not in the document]
innerdvations: It's to allow passing arguments to jest with `--`, if this is true it considers `--` itself to be an argument

tests/cli/tests/strapi/strapi/content-types-list.test.cli.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const coffee = require('coffee');
+const semver = require('semver');
+const assert = require('assert');
```
jhoward1994, line 5: Can we remove these imports?

innerdvations: oops, yes, good catch!

jhoward1994:
To get the tests to pass I have to run it as:
`rm -rf test-apps; STRAPI_DISABLE_EE=true yarn test:cli`

Otherwise I see errors like this because of EE features

image

innerdvations:
> To get the tests to pass I have to run it as: `rm -rf test-apps; STRAPI_DISABLE_EE=true yarn test:cli`
> 
> Otherwise I see errors like this because of EE features
> 
> image

What if you run a `yarn test:clean` in between CE / EE runs?

Anyway, the point isn't to test the actual content types that show up, so I think it's fine to update the tests themselves to only run on CE so we don't have to worry about it

jhoward1994:
Nope `yarn test:clean; yarn test:cli` still doesn't work. It's services, controllers, content types and routes that fail in EE

I think running on CE makes sense 👍🏻

innerdvations:
> Nope `yarn test:clean; yarn test:cli` still doesn't work. It's services, controllers, content types and routes that fail in EE
> 
> I think running on CE makes sense 👍🏻

Oh oops, it's actually because I don't have a database reset in the `beforeAll` 🙈

That would be a better option for sure.

innerdvations:
Since implementing a reset is a lot more work (although something we absolutely need to implement soon, I'll add a task for it in the backlog) for now I've changed the expectation from "equal" to "includes" so that additional list items won't cause a problem.

Convly (approved):
Nice addition :100:

PR #20954, opened 2024-08-06
RBAC Action Aliases - v4

### LINKED ISSUE #19001: User ID shown in relation dropdown instead of the configured mainField

Convly:
:warning:  **This PR is a v4 port of https://github.com/strapi/strapi/pull/20878.**

#### Action Aliases

Introduce a new `aliases` property on admin actions (`RBAC`).

This property is defined as an optional array of action alias.

```ts
interface ActionAlias {
  /**
   * The action ID to alias
   */
  actionId: string;

  /**
   * An optional array of subject to restrict the alias usage
   */
  subjects?: string[];
}

type Action = {
  // other properties ...

  /**
   * An optional array of @see {@link ActionAlias}.
   *
   * It represents the possible aliases for the current action.
   *
   * Aliases are unidirectionals.
   *
   * Note: This is an internal property and probably shouldn't be used outside of Strapi core features.
   *       Its behavior might change at any time without notice.
   *
   * @private
   */
  aliases?: ActionAlias[];
};
```

#### Action Provider

A new method `unstable_aliases` is added to the admin action provider (`RBAC`).

```ts
declare function unstable_aliases(actionId: string, subject?: string | null): string[];
```

It returns an array of `actionId` that can be used as aliases for the given `actionId` and `subject`.

- When an alias doesn't declare a `subjects` property, it only checks actions ID
- When an alias has an array of subjects declared, it ensures both the actions ID and subjects matches   

#### New Aliases

A default set of aliases for common use cases have been defined:

- `admin::users.read` is an alias of `plugin::content-manager.explorer.read`(`admin::user`)
- `admin::users.roles` is an alias of `plugin::content-manager.explorer.read`(`admin::role`)
- `plugin::i18n.locale.read` is an alias of `plugin::content-manager.explorer.read`(`plugin::i18n.locale`)

### Why is it needed?

The initial idea came when looking at #19001. The issue being hidden content-types (`admin::user`, `plugin::i18n.locale`, etc...) not having a content manager action being registered for them (only CTs being displayed in the CM have one).

Since it is possible to create a relation to `admin::user` (manual exception), this is why even super admin can't see the main field for the relation (because the needed permission doesn't even exist).

**Note:** It's also already possible to declare manual relations to any internal content-type from the schema file (e.g. `locale`)

This is a limitation in how permissions are defined. However, simpler solutions like only adding the hidden content-types (mostly internal ones like `upload::file` or `upload::folder`) to the list of allowed subjects for the `plugin::content-manager.explorer.read` action might lead to unwanted side effects (like adding them to the roles' edit permission table).

The proposed (internal) `aliases` feature mitigates the issue by adding more flexible options when checking permissions while limiting side effects on already existing features and behaviors. 

### How to test it?

1. Create a content-type with a relation to `admin::user` (`CT`)
2. Create a non super-admin user (`A`), go the content manager
3. From the super-admin account (`B`), set the following permissions for (`A`)'s role
  - In `Collection Types`: allow every CRUD operation on `CT`
  - In `Settings`: make sure `Read` is **not** checked in `Users & Roles`
4. As (`A`) Go to `Content-Manager` > `CT` > `Create new entry`
5. When opening the select input for the `admin::user` relation, only document IDs should be visible
6. As (`B`), update (`A`)'s role permissions, and check the `Read` box in `Settings` > `Users & Roles
7. As (`A`), try opening the select input again, the proper main field should be displayed (by default `firstname`)

### Related issue(s)/PR(s)

#19001

### Notes

- Aliases are unidirectional to allow setting more granular permissions
- The aliases defined in this PR are just meant to address common scenarios, but this is meant to be extended based on future use cases
- While this is set as an internal feature (for now), it could be interesting to imagine the potential use cases for plugin & user app if made public

### Screenshots

Screenshot from 2024-07-25 00-43-21
Screenshot from 2024-07-25 00-44-03

fixes #19001

### DIFF
--- a/packages/core/admin/server/src/config/admin-actions.ts
+++ b/packages/core/admin/server/src/config/admin-actions.ts
@@ -50,6 +50,12 @@ export const actions = [
     section: 'settings',
     category: 'users and roles',
     subCategory: 'users',
+    aliases: [
+      {
+        actionId: 'plugin::content-manager.explorer.read',
+        subjects: ['admin::user'],
+      },
+    ],
   },
   {
     uid: 'users.update',
@@ -82,6 +88,12 @@ export const actions = [
     section: 'settings',
     category: 'users and roles',
     subCategory: 'roles',
+    aliases: [
+      {
+        actionId: 'plugin::content-manager.explorer.read',
+        subjects: ['admin::role'],
+      },
+    ],
   },
   {
     uid: 'roles.update',
--- a/packages/core/admin/server/src/domain/action/__tests__/action-provider.test.ts
+++ b/packages/core/admin/server/src/domain/action/__tests__/action-provider.test.ts
@@ -8,12 +8,12 @@ const providerMethods = [
   'appliesToProperty',
   'delete',
   'get',
-  'getWhere',
   'values',
   'keys',
   'has',
   'size',
   'clear',
+  'unstable_aliases',
 ];
 
 describe('Action Provider', () => {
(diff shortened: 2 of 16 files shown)


innerdvations:
Got some failing unit tests

innerdvations (approved):
LGTM! Tested and works fine for me!

PR #20958, opened 2024-08-07
fix(content-manager): dont perform unnecessary requests

### LINKED ISSUE #20736: v5: `countDraftRelations` is called for every entry in the list view

- Node.js version: 20
- Yarn version: 1.22
- Strapi version: 5 rc 3
- Database: sqlite
- Operating system: macOS

### Describe the bug

Just gave the RC3 a spin. Something I notice is that the list view is significantly slower because it calls `countDraftRelations` for every entry in view. If you set it to 100 entries, it will DDoS it's own backend with 100 tiny API calls. Is this really necessary?

1. Go to a list view
2. Set viewing entries to 100
3. Wait until the UI stops crashing
4. See the logs go crazy with API calls

### Expected behavior

I think this could be solved smarter than calling the API so many times, all of the time.

### Additional context

As discussed in the v5 beta discord channel

Feranchz:
### What does it do?

`DescriptionComponentRenderer` receive all the possible buttons and all the components are rendered even if at the end they are not showed. So we explicitly remove the publish button as it is not showed in the ListView and because the PublishAction button performs requests that we don't want to make at the list view.

### How to test it?

1. Go to any CT on the Content Manager. 
2. It should not have any request to countDraftRelations endpoint

### Related issue(s)/PR(s)

closes #20736

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/components/TableActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/components/TableActions.tsx
@@ -52,9 +52,10 @@ const TableActions = ({ document }: TableActionsProps) => {
   return (
     <DescriptionComponentRenderer
       props={props}
-      descriptions={(
-        plugins['content-manager'].apis as ContentManagerPlugin['config']['apis']
-      ).getDocumentActions()}
+      descriptions={(plugins['content-manager'].apis as ContentManagerPlugin['config']['apis'])
+        .getDocumentActions()
+        // We explicitly remove the PublishAction from description so we never render it and we don't make unnecessary requests.
+        .filter((action) => action.name !== 'PublishAction')}
     >
       {(actions) => {
         const tableRowActions = actions.filter((action) => {


packages/core/content-manager/admin/src/pages/ListView/components/TableActions.tsx
```diff
@@ -52,9 +52,10 @@ const TableActions = ({ document }: TableActionsProps) => {
   return (
     <DescriptionComponentRenderer
       props={props}
-      descriptions={(
-        plugins['content-manager'].apis as ContentManagerPlugin['config']['apis']
-      ).getDocumentActions()}
+      descriptions={(plugins['content-manager'].apis as ContentManagerPlugin['config']['apis'])
+        .getDocumentActions()
+        // We explicitly remove the PublishAction from description so we never render it and we don't make unnecessary requests.
+        .filter((action) => action.name !== 'PublishAction')}
```
madhurisandbhor, line 58: why we don't need bulk publish?

Feranchz: This is not related to Bulk Publish. Is about the individual publish button which is not showed on the ListView, so here we avoid run the component

madhurisandbhor (approved):

simotae14 (approved):

PR #20959, opened 2024-08-07
feat(cloud-cli): preselect node-version based on local node-version

nathan-pichon:
### What does it do?

Adding a preselection on the currently running NodeJS version while selecting the Node Version on project creation in Cloud CLI.

### Why is it needed?

Better UX

### How to test it?

- Try to deploy a new project to the Cloud using `strapi deploy` command.
- Setup your local to either run on Node 18 or 20
- See the node version preselected when asked to select a node version

### Related issue(s)/PR(s)


packages/cli/cloud/src/create-project/utils/project-questions.ts
abdallahmz, line 1: suggestion/optional:

- call this project-questions.utils.ts
- return `projectQuestionsUtils = {
...your functions here
}`

(IMO it would improve discovery)

nathan-pichon: I don't get why `return projectQuestionsUtils = { ...your functions here }` would improve discovery versus exporting individually all functions?

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export function questionDefaultValuesMapper(
+  QuestionsMapper: Partial<{
+    [K in keyof ProjectAnswers]:
+      | ((question: DistinctQuestion<ProjectAnswers>) => ProjectAnswers[K])
+      | ProjectAnswers[K];
+  }>
+) {
+  return (
+    questions: ReadonlyArray<DistinctQuestion<ProjectAnswers>>
```
abdallahmz, line 16: Are we really expecting to apply the default values to various sets of questions?
If not, it might be argued that a pure function would be more suitable (for simplicity & readability).

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
...
+  };
+}
+
+/**
+ * Get default values from questions
+ * @param questions - An array of questions for project creation
+ */
+export function getDefaultsFromQuestions(
+  questions: ReadonlyArray<DistinctQuestion<ProjectAnswers>>
+) {
```
abdallahmz, line 49: (suggestion) Returned type would help to understand what's happening

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return acc;
+  }, {});
+}
+
+/**
+ * Get the default node version based on the current node version if it is in the list of choices
+ * @param question - The question for the node version in project creation
+ */
+export function getProjectNodeVersionDefault(question: DistinctQuestion<ProjectAnswers>) {
```
abdallahmz, line 62: (suggestion) Returned type would help here (I think it's simply a string?)

abdallahmz:
LGTM overall 👍 ping me for testing!

packages/cli/cloud/src/create-project/utils/project-questions.ts
abdallahmz: just by reading the file name, I would be able to use it:
- I import projectQuestionsUtils in the file I'm working on (I know it's there since the file is called project-questions.utils.ts)
- then, IDE suggests everything (correct import, available functions signatures etc.) and I can start using it right away.

Versus, in the current state, I have to open the file and scroll through the content, copy paste the names, etc.

It's a dev workflow thing so I understand it's a bit subjective, but I find myself more efficient this way :D

packages/cli/cloud/src/create-project/utils/project-questions.ts
nathan-pichon: > Are we really expecting to apply the default values to various sets of questions?
As I was there, I made a generic function that could handle all the questions.
It's the second time we have had this need (the first was the project's name), so we are not excluding having this need again in future.

I don't really understand why a "simple function" (I guess you mean only one level function that is getting everything in parameters) would enhance readability. Wouldn't it be the same if we curry the function you talk about?
For me, splitting into multiple functions (having a context for each, first for setup, second for mapping) is more readable (and simpler, with fewer parameters and a clear purpose) than having too many parameters for a single function.

packages/cli/cloud/src/create-project/utils/project-questions.ts
nathan-pichon: I don't mind to add this export, but it wouldn't be used and leave unused code in the codebase 😅

abdallahmz: I know projectQuestionsUtils exists because it is the filename. So, just by looking at the filename, I can use it.

I don't need to look at other parts of the code. And I am very likely to stumble upon the filename, because I'd likely be working in the same folders.

(also, I use filename search very often)

packages/cli/cloud/src/create-project/utils/project-questions.ts
abdallahmz: Hmm then what's the limit? If there are 2 parameter, should we go for higher order function systematically?
We could end up with a bit of overengineering I feel like ^^' 

But I understand your point, let's keep it as is 👍

packages/cli/cloud/src/create-project/utils/project-questions.ts
nathan-pichon: Why would you use projectQuestionsUtils file if you don't know its content? 😅

abdallahmz: Various reasons:
- maybe the name would be explicit enough.
- maybe I vaguely remember the content (maybe I wrote it a few months ago, or made the review, or used it before) In those cases I am very likely to "know" what's inside without knowing the exact function names.

Applied to the whole codebase, I guarantee you it makes my workflow much smoother.

Again it's optional, the code won't break either way so I'll let you take the final decision 🫡

nathan-pichon: I'm not strongly opposed to that, but I wanted to understand how it can ease your day-to-day 🤔 
I believe it won't improve mine as I will never import things I don't know of, but if it can helps you, maybe it can help others 🤔

nathan-pichon: I just saw why I would not use this method in my IDE.
I get this while typing the start of a function
image

abdallahmz: We do have that in VSCode as well, so it does answer part of my concerns/usecase indeed 👍 

So I only have the "discover by reading the filename" but it's more of a general thing (maybe for this utils there is less value)

About the god object, IMO seeing `projectQuestionsUtils.doSomething`(instead of `doSomething` ) is more explicit becausee I know where `doSomething` comes from (it's an Utils, so I know there won't be third party calls, etc). There's more "information" within the line (so it helps for readability), instead of having to scroll back up or remember stuff from lines before.

Endless debate ^^' but yeah if you're not going to import `projectQuestionsUtils` it limits some of the (imo) "benefits", so we can keep as is.

nathan-pichon: I'm not sure it's an endless debate, but for sure an interesting one! Improving our DX is important as well as understanding how others work too. 👍

packages/cli/cloud/src/create-project/utils/project-questions.ts
nathan-pichon: Should be good now 👍

packages/cli/cloud/src/create-project/utils/project-questions.ts
nathan-pichon: Should be good now 👍

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
+import { DistinctQuestion } from 'inquirer';
+import type { ProjectAnswers } from '../../types';
+
+/**
+ * Apply default values to questions based on the provided mapper
+ * @param QuestionsMapper - A partial object with keys matching the ProjectAnswers keys and values being the default value or a function to get the default value
+ */
+export function questionDefaultValuesMapper(
+  QuestionsMapper: Partial<{
```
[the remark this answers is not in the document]
nathan-pichon: I opted for questionsMap, is it ok?

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
...
+  }>
+) {
+  return (
+    questions: ReadonlyArray<DistinctQuestion<ProjectAnswers>>
+  ): ReadonlyArray<DistinctQuestion<ProjectAnswers>> => {
+    return questions.map((question) => {
+      const questionName = question.name as keyof ProjectAnswers;
+
+      // If the question is part of the mapper, apply the default value
+      if (question.name && QuestionsMapper[questionName]) {
```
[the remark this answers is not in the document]
nathan-pichon: Should be good now 👍

packages/cli/cloud/src/create-project/utils/project-questions.ts
```diff
@@ -0,0 +1,73 @@
...
+            default: questionDefault(question),
+          };
+        }
+        // else we consider it as a static value
+        return {
+          ...question,
+          default: questionDefault,
+        };
+      }
+      return question;
```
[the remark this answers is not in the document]
nathan-pichon: Should be good now 👍

abdallahmz (approved):
QA passed ✅ (see ticket for test details)

PR #20960, opened 2024-08-07
Add codemod for react and react-dom

Bassel17:
### What does it do?

Codemod to upgrade react and react-dom to `^18.0.0`, and keeps both under the same version

### How to test it?

run following command from examples project in the repo

`../../packages/utils/upgrade/bin/upgrade.js codemods run 5.0.0-dependency-upgrade-react-and-react-dom-json`

play around with the versions in package.json, should upgrade both to `^18.0.0`

### DIFF
--- a/packages/utils/upgrade/resources/codemods/5.0.0/dependency-upgrade-react-and-react-dom.json.ts
+++ b/packages/utils/upgrade/resources/codemods/5.0.0/dependency-upgrade-react-and-react-dom.json.ts
@@ -1,0 +1,67 @@
+import path from 'node:path';
+import semver from 'semver';
+
+import type { modules } from '../../../dist';
+
+const REACT_DEP_NAME = 'react';
+const REACT_DEP_PATH = `dependencies.${REACT_DEP_NAME}`;
+
+const REACT_DOM_DEP_NAME = 'react-dom';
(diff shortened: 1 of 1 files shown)


jhoward1994 (approved):

packages/utils/upgrade/resources/codemods/5.0.0/dependency-upgrade-react-and-react-dom.json.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const rootPackageJsonPath = path.join(cwd, 'package.json');
+
+  if (file.path !== rootPackageJsonPath) {
+    return file.json;
+  }
+
+  const j = json(file.json);
+
+  if (j.has(REACT_DEP_PATH) && j.has(REACT_DOM_DEP_PATH)) {
```
Convly, line 35: What if it only has one of them :thinking:

Bassel17: doesn't matter since they have to be both present, if they have only one in the strapi project something is wrong, so  we set both to `18.0.0` tried to make this logic simple no need to overthink it, both need to be present and on the same version

Convly (approved):

PR #20963, opened 2024-08-07
[Fix] Validate cloning as entity creation so field validations aren't skipped e.g. unique constraints

jhoward1994:
### What does it do?

Treats cloning operations in V4 as entity creation.

### Why is it needed?

So that unique field constraints are applied correctly when cloning

I have tested other field validations during cloning since this change and for me they are still working.

### How to test it?

Run new E2E test

### Related issue(s)/PR(s)

DX-1502


Marc-Roig (approved):
good catch :)

PR #20969, opened 2024-08-08
flatten tests

Boegie19:
### What does it do?

flatten the test structure

### Why is it needed?

To decrease wait time on tests with this down from 17 min to [11 min](https://github.com/strapi/strapi/actions/runs/10296893677)

### How to test it?

see test pipeline

### Related issue(s)/PR(s)


innerdvations (approved):
Ok, let's try it and see, we can easily revert it if we run into issues like limits.

Thanks!

PR #20973, opened 2024-08-08
fix(content-manager): dont count draft relations on the list view

jhoward1994:
### What does it do?

Don't call the `countDraftRelations` from the content manager list view

### Why is it needed?

- There is no use for the response from these calls in the list view
- When bulk publishing from the list view we make a call to `API/countManyEntriesDraftRelations` which provides the data we need
- It was causing performance issues as a call was made for every entry in the list view table

### How to test it?

Navigate to the CM list view and inspect the network tab

### Related issue(s)/PR(s)

DX-1501

Fixes https://github.com/strapi/strapi/issues/20736


jhoward1994:
Don't merge for now. I've just realised this implementation is probably wrong as we can bulk publish when multiple entries in the list are selected

remidej (approved):
👍

PR #20976, opened 2024-08-08
Chore(CLI): Add validation to prevent deployment of suspended projects

gonbaum:
### What does it do?

This PR adds a new validation that prevents deployment if the project is in a suspended status.

### Why is it needed?

Currently, the deployment process continues even if the project is suspended, only failing after the user completes the upload without the CLI providing any prior warning.

### How to test it?

- Try the deploy command on a suspended project:
    - 	The deployment should not initiate.
    -   A message should be displayed explaining why the deployment didn’t start, along with a link to the dashboard where you can view the suspension reason and take action.

### DIFF
--- a/packages/cli/cloud/src/deploy-project/action.ts
+++ b/packages/cli/cloud/src/deploy-project/action.ts
@@ -100,18 +100,7 @@ async function upload(
       return data.build_id;
     } catch (e: any) {
       progressBar.stop();
-      if (e instanceof AxiosError && e.response?.data) {
-        if (e.response.status === 404) {
-          ctx.logger.warn(
-            `The project does not exist. Please link your local project to a Strapi Cloud project using the link command.`
-          );
-        } else {
-          ctx.logger.error(e.response.data);
-        }
-      } else {
-        ctx.logger.error('An error occurred while deploying the project. Please try again later.');
-      }
-
+      ctx.logger.error('An error occurred while deploying the project. Please try again later.');
       ctx.logger.debug(e);
     } finally {
       await fse.remove(tarFilePath);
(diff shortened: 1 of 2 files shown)


packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -166,7 +155,38 @@ export default async (ctx: CLIContext) => {
...
+    const { data: projectInfo } = await cloudApiService.getProject({ name: project.name });
+    const isProjectSuspended = projectInfo.suspendedAt;
+    if (isProjectSuspended) {
+      ctx.logger.warn(
+        'Oops! This project has been suspended. \n Please reactivate it from the dashboard to continue deploying: '
+      );
+      ctx.logger.log(chalk.underline(`${apiConfig.dashboardBaseUrl}/projects/${project.name}`));
+      return;
+    }
+    console.log('paso por checkeo de info');
```
nathan-pichon, line 170: 👀

packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -166,7 +155,38 @@ export default async (ctx: CLIContext) => {
     return;
   }
 
-  const cloudApiService = await cloudApiFactory(ctx);
+  const cloudApiService = await cloudApiFactory(ctx, token);
+
+  try {
+    const { data: projectInfo } = await cloudApiService.getProject({ name: project.name });
+    const isProjectSuspended = projectInfo.suspendedAt;
+    if (isProjectSuspended) {
```
nathan-pichon, line 163:
```suggestion
    const isProjectSuspended = projectInfo.suspendedAt;

    if (isProjectSuspended) {
```

packages/cli/cloud/src/services/cli-api.ts
```diff
@@ -181,6 +183,23 @@ export async function cloudApiFactory(
       }
     },
 
+    async getProject({ name }): Promise<AxiosResponse<unknown>> {
+      try {
+        const response = await axiosCloudAPI.get(`/projects/${name}`);
+
+        if (response.status !== 200) {
+          throw new Error("Error fetching project's info from the server.");
```
nathan-pichon, line 191: Might be a bit too confusing for users to talk about "the server". 🤔 
```suggestion
          throw new Error("There was a problem retrieving your project details. Please try again later.");
```

gonbaum: I was being consistent with the rest of the error messages which also use this wording. But I agree.

packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -166,7 +155,38 @@ export default async (ctx: CLIContext) => {
...
+
+  try {
+    const { data: projectInfo } = await cloudApiService.getProject({ name: project.name });
+    const isProjectSuspended = projectInfo.suspendedAt;
+
+    if (isProjectSuspended) {
+      ctx.logger.warn(
+        'Oops! This project has been suspended. \n Please reactivate it from the dashboard to continue deploying: '
+      );
+      ctx.logger.log(chalk.underline(`${apiConfig.dashboardBaseUrl}/projects/${project.name}`));
```
nathan-pichon, line 168: Can we get this url from project info ?
As we discovered this morning, we might have changes in URLs in the future.

gonbaum: Unfortunately, there is no dashboard url (or base url) in the project info. We would have to construct it in the same way. However, we could build it and return it on API side.

nathan-pichon: Yes, that's what I meant 🙌

nathan-pichon: I think I will do the same to fix the issue with the deployment link too.

nathan-pichon (approved):
LGTM ! Thanks

PR #20984, opened 2024-08-09
Now User Can Create Users based on role field in Payload, Also role information in response

anildalar:
### What does it do?

This pull request introduces changes to the sanitizeUser function to include detailed role information in the user response. The following technical changes were made:

Enhancement of sanitizeUser function:

The sanitizeUser function now fetches and includes the user's role details, ensuring that the role object is part of the user response.
The findOne method is used with populate: ['role'] to ensure that role data is fetched and included in the sanitized output.
The sanitize.contentAPI.output function is utilized to sanitize user data while allowing the inclusion of the role object.
Modification of User Creation Logic:
Default Role Assignment: If the role field is not specified in the payload, the system assigns the default role automatically, ensuring consistency and reducing the likelihood of errors.

Adjusted the user creation logic in the controller to utilize the enhanced sanitizeUser function.
Updated the API response to include the full role object alongside other user details.

### Why is it needed?

The changes are necessary to resolve the following issues:

Incomplete User Information: The previous implementation only returned a partial user object without role information, which was insufficient for certain application functionalities.
Improved User Experience: By including role details, clients can display comprehensive user profiles, enhancing user management and UI/UX.
Enhanced Security and Customization: Ensuring only the necessary and relevant user information is shared, improving data security and application customization.

### How to test it?

To verify the new behaviour, follow these steps:

Environment Setup:

Use the latest version of the code from this branch.
Testing Steps:

Create a User: Use the API to create a new user with a specific role.
Fetch User Data: Call the user API endpoint that uses the sanitizeUser function.
Verify Response: Check the API response to ensure it includes the complete role object with fields like id, name, description, type, createdAt, and updatedAt.

### Related issue(s)/PR(s)
WhatsApp Image 2024-08-09 at 12 37 35 PM
WhatsApp Image 2024-08-09 at 12 47 02 PM
WhatsApp Image 2024-08-09 at 12 48 58 PM

### DIFF
--- a/packages/plugins/users-permissions/server/controllers/auth.js
+++ b/packages/plugins/users-permissions/server/controllers/auth.js
@@ -28,11 +28,28 @@ const {
 const { getAbsoluteAdminUrl, getAbsoluteServerUrl, sanitize } = utils;
 const { ApplicationError, ValidationError, ForbiddenError } = utils.errors;
 
-const sanitizeUser = (user, ctx) => {
+const sanitizeUser = async(user, ctx) => {
   const { auth } = ctx.state;
+  // Ensure the role relation is populated
+  const populatedUser = await strapi.query('plugin::users-permissions.user').findOne({
+    where: { id: user.id },
+    populate: ['role'], // Populate the role relation
+  });
+  
   const userSchema = strapi.getModel('plugin::users-permissions.user');
 
-  return sanitize.contentAPI.output(user, userSchema, { auth });
+  const sanitizedData = await sanitize.contentAPI.output(user, userSchema, { auth });
+  // Attach role information
+  sanitizedData.role = {
+    id: populatedUser.role.id,
+    name: populatedUser.role.name,
+    description: populatedUser.role.description,
+    type: populatedUser.role.type,
+    createdAt: populatedUser.role.createdAt,
+    updatedAt: populatedUser.role.updatedAt,
+  };
+
+  return sanitizedData;
 };
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -347,12 +364,24 @@ module.exports = {
 
     await validateRegisterBody(params);
 
-    const role = await strapi
-      .query('plugin::users-permissions.role')
-      .findOne({ where: { type: settings.default_role } });
+     // Determine role to assign
+    let role;
+    if (ctx.request.body.role) {
```
derrickmehaffy, line 369: Role should be sent during registration that would be a huge security vulnerability

packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -347,12 +364,24 @@ module.exports = {
...
+      role = await strapi.query('plugin::users-permissions.role').findOne({
+        where: { id: ctx.request.body.role },
+      });
+
+      if (!role) {
```
derrickmehaffy, line 374: This would be a breaking change

derrickmehaffy (commented):
Sorry but we cannot accept this PR, we specifically do not allow setting the role during register as it would a major security concern. Likewise forcing people to specify the role would be a breaking change.

packages/plugins/users-permissions/server/controllers/auth.js
anildalar: Then what is the way to create user based on different role from API ?

PR #20985, opened 2024-08-09
fix: don't check uniqueness for empty strings

remidej:
### What does it do?

Skips the uniqueness validation when the value is an empty string. This applies to short text, long text, and uid fields.

### Why is it needed?

After checking with @innerdvations, we never want to check uniqueness when no value is provided. Uniqueness was already skipped for null and undefined, but not for empty strings. This was revealed when working on the UID field, but was also an issue for text fields.

### How to test it?

- create a content type with a slug field
- save 2 entries with different slug values, you should be able to save
- try to have the same value in both slugs, you should get an error
- clear the value in both slugs (make them empty), **you should be able to save**

Repeat the same steps for a text value with uniqueness turned on

### DIFF
--- a/packages/core/core/src/services/entity-validator/validators.ts
+++ b/packages/core/core/src/services/entity-validator/validators.ts
@@ -339,10 +339,10 @@ const addUniqueValidator = <T extends yup.AnySchema>(
 
   return validator.test('unique', 'This attribute must be unique', async (value) => {
     /**
-     * If the attribute value is `null` we want to skip the unique validation.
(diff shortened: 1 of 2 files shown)


Marc-Roig (approved):
lgtm, would prefer if we had an additional api test to be sure !

innerdvations (commented):
Change looks good, but can we add an API test? It looks like we don't have any tests for which values should bypass uniqueness yet unfortunately.

tests/api/core/strapi/document-service/uniqueness.test.api.ts

Marc-Roig (approved):

PR #20987, opened 2024-08-09
test: enable CTB tests on CI

innerdvations:
### What does it do?

- Re-enables to CTB e2e tests on github
- Skips them only on webkit due to a known bug that will be fixed later
- Increases the timeout waiting for server restart (that was one reason they were failing before)
- does db reset before file reset (another cause of them failing)
- updates playwright version

### Why is it needed?

They were being skipped

### How to test it?

CTB e2e tests should run and pass (only being skipped on webkit)

### Related issue(s)/PR(s)

DX-1516

### DIFF
--- a/playwright.base.config.js
+++ b/playwright.base.config.js
@@ -37,8 +37,8 @@ const getEnvBool = (envVar, defaultValue) => {
 const createConfig = ({ port, testDir, appDir }) => ({
   testDir,
 
-  /* default timeout for a jest test to 30s */
-  timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 30 * 1000),
+  /* default timeout for a test to 30s */
+  timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 90 * 1000),
 
   expect: {
     /**
--- a/playwright.base.config.js
+++ b/playwright.base.config.js
@@ -52,7 +52,7 @@ const createConfig = ({ port, testDir, appDir }) => ({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 3 : 0,
+  retries: process.env.CI ? 4 : 2,
   /* Opt out of parallel tests on CI. */
   workers: 1,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
--- a/playwright.base.config.js
+++ b/playwright.base.config.js
@@ -37,7 +37,7 @@ const getEnvBool = (envVar, defaultValue) => {
 const createConfig = ({ port, testDir, appDir }) => ({
   testDir,
 
-  /* default timeout for a test to 30s */
+  /* default timeout for a test */
   timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 90 * 1000),
 
   expect: {
--- a/tests/e2e/tests/content-manager/history.spec.ts
+++ b/tests/e2e/tests/content-manager/history.spec.ts
@@ -48,6 +48,7 @@ describeOnCondition(edition === 'EE')('History', () => {
   });
 
   test.afterAll(async () => {
+    await resetDatabaseAndImportDataFromPath('with-admin.tar');
     await resetFiles();
   });
 
--- a/tests/e2e/utils/file-reset.ts
+++ b/tests/e2e/utils/file-reset.ts
@@ -33,6 +33,8 @@ export const pollHealthCheck = async (interval = 1_000, timeout = 30_000) => {
   throw new Error('Timeout reached, service did not become available in time.');
 };
 
+// resetFiles resets the filesystem state back to the original state, removing uploaded files, schemas, etc
+// NOTE: in some cases, call resetFiles without first using resetDatabase* can cause the restart to fail
 export const resetFiles = async () => {
   if (process.env.TEST_APP_PATH) {
     console.log('Restoring filesystem');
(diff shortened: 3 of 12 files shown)


innerdvations:
Tests passed first try, trying them again to get some reassurance they're not flaky.

tests/e2e/utils/file-reset.ts
```diff
@@ -50,6 +50,6 @@ export const resetFiles = async () => {
   // wait for server to restart after modifying files
   console.log('Waiting for Strapi to restart...');
   // TODO: this is both a waste of time and flaky. We need to find a way to access playwright server output and watch for the "up" log to appear
-  await delay(3); // give it time to detect file changes and begin its restart
+  await delay(4); // give it time to detect file changes and begin its restart
```
Convly, line 53: Seems arbitrary; do we have any idea if this could cause problems (flaky tests) in the future? Is this impacted by servers' performances?

playwright.base.config.js
```diff
@@ -52,7 +52,7 @@ const createConfig = ({ port, testDir, appDir }) => ({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 3 : 0,
```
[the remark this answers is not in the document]
innerdvations: Less flakiness on the CI, but also our playwright tests (in general, not the CTB being enabled here) are flaky enough that having a couple retries are pretty much required to get them all passing consistently even locally, so I added that so we don't have to constantly add the retries manually.

tests/e2e/utils/file-reset.ts
innerdvations: It is arbitrary. It is based on the server performance, and the CI is currently much much slower than running tests locally, so this just adds 1s to every fileReset to give Strapi server more time to realize that files have changed and resart itself.

Maybe a future improvement could just be to shut down Strapi, reset the files, then start up Strapi again, but I don't know if that would help.

Essentially, I found that the primary reason that the CTB tests were failing was just timing out, with a secondary cause being that in some cases (I couldn't determine when) having a fileReset before a resetDatabase was causing tests to fail.

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -48,6 +48,7 @@ describeOnCondition(edition === 'EE')('History', () => {
   });
 
   test.afterAll(async () => {
+    await resetDatabaseAndImportDataFromPath('with-admin.tar');
```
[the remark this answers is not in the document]
innerdvations: The main reason is that `resetFiles()` seems to require a `resetDatabase` before it -- one of the changes I just discovered this morning, and I will add a comment explaining that.

The reason for having a `resetFiles()` in `afterAll` is already commented in a couple other tests, and the reason for that is simply time-saving. Yes, every test should be responsible for setting itself up properly, but if we did that they would all be triggering a 10+ second fileReset that is only necessary if the filesystem has actually changed. So until we have a chance to fix that, we just have the tests that mess up the filesystem clean up after themselves.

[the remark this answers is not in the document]
innerdvations: (added comment to resetFiles)

tests/e2e/utils/restart.ts
```diff
@@ -1,19 +1,25 @@
...
+    page.waitForSelector('text="is taking longer"', { state: 'visible' }),
+  ])
+    .then(() => true)
+    .catch(() => false);
 };
 
 /**
- * Wait for a restart modal to appear, but instead of failing if it doesn't, attempt to
- * refresh the page and see if it comes back up
+ * Wait for a restart modal to appear and then disappear
```
innerdvations, line 14: The reload was originally introduced to reduce flakiness, but now it is causing tests to fail; it shouldn't have been needed to begin with so I've removed it

tests/e2e/utils/shared.ts
```diff
@@ -71,6 +71,11 @@ export const findAndClose = async (
 };
 
 export const createSingleType = async (page, data) => {
+  // TODO: fix webkit bug with selecting ctb attributes
```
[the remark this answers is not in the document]
innerdvations: I'm not exactly sure what the issue is, in the trace viewer it looks like the page is slightly offset but I really have no idea why it's failing consistently and only on webkit. Re-enabling the tests for webkit and running them will provide more info that I'd be able to.

Convly (changes requested):
Chromium isn't happy :disappointed:

innerdvations:
> Chromium isn't happy 😞

For some reason it's "consistently flaky" on chromium and almost always fails on the first run. If I run it a second time, it works. I'll keep trying until it looks like it will pass consistently.

playwright.base.config.js
```diff
@@ -37,8 +37,8 @@ const getEnvBool = (envVar, defaultValue) => {
 const createConfig = ({ port, testDir, appDir }) => ({
   testDir,
 
-  /* default timeout for a jest test to 30s */
-  timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 30 * 1000),
+  /* default timeout for a test to 30s */
```
jhoward1994, line 40:
```suggestion
  /* default timeout for a test to 90s */
```
?

tests/e2e/utils/shared.ts
```diff
@@ -137,6 +142,11 @@ export const createSingleType = async (page, data) => {
 };
 
 export const createCollectionType = async (page, data) => {
+  // TODO: fix webkit bug with selecting ctb attributes
+  if (page.context()._browser._name === 'webkit') {
+    return test.fixme();
+  }
```
jhoward1994, line 148:
```suggestion
  if (page.context()._browser._name === 'webkit') {
    return test.fixme();
  }
```

const browserName = page.context().browser().browserType().name();
if (browserName === 'webkit') {
    return test.fixme();
}

innerdvations: This block is temporary and copy-pasted several places so I'll leave this one alone

jhoward1994 (approved):

Issue #20967, opened 2024-08-07
Clicking a filter deletes the filter instead of allowing you to edit it


derrickmehaffy:
### Required System information

- Node.js version: v20
- NPM version: yarn v1
- Strapi version: 4.25.7
- Database: SQLite
- Operating system: Linux Mint 21.2
- Is your project Javascript or Typescript: Javascript

### Describe the bug

Clicking a filter anywhere deletes it rather than allowing you to edit it

### Steps to reproduce the behavior

Reproduction video: https://www.loom.com/share/3e8a5a13b0cb4d2fadbc2ba33f507292?sid=1851f53c-a73c-47fd-94d6-52ec21f6e8b9

### Expected behavior

Clicking a filter other than the "x" should allow you to edit that filter

### Screenshots

N/A

### Code snippets

N/A

### Additional context

Related to TID7764

moeez506:
Hey @derrickmehaffy can I work on this issue ?

PR #20997, opened 2024-08-12
Fix(content-manager): Clicking a filter deletes the filter instead of allowing to edit it.

moeez506:
### What does it do?

Filter edit functionality added. Clicking a filter other than the "x" will allow you to edit that filter.

### Why is it needed?

Clicking a filter anywhere deletes it rather than allowing you to edit it

### How to test it?

Navigate to content manager and add any filter then click on it, will allow you to edit.

### Related issue(s)/PR(s)
#20967 

### Need Assitance. 
 I have solved 90% issue but the design part is ambigous, right now the edit modal is stick to the top left corner which should be exactly below the filter being edited. Needed some guidance regarding that. 
 Attached recording for refrence 
 https://www.loom.com/share/07f37672aaeb4860b3eefe14c53cdf03?sid=40084628-af47-4d91-a237-0c201da496f5
 
 PS: New to strapi codebase, it's my first contribution here.

### DIFF
--- a/packages/core/helper-plugin/src/components/FilterListURLQuery.tsx
+++ b/packages/core/helper-plugin/src/components/FilterListURLQuery.tsx
@@ -7,6 +7,7 @@ import { useIntl } from 'react-intl';
 import { useQueryParams } from '../hooks/useQueryParams';
 
 import type { FilterData, Filter } from '../types';
+import { FilterPopoverURLQuery } from './FilterPopoverURLQuery';
 export interface FilterListURLQueryProps {
   filtersSchema: FilterData[];
 }
@@ -18,7 +19,6 @@ export const FilterListURLQuery = ({ filtersSchema = [] }: FilterListURLQueryPro
     };
     page?: number;
   }>();
-
   const handleClick = (filter: Filter) => {
     const nextFilters = (query?.filters?.$and || []).filter((prevFilter) => {
       const name = Object.keys(filter)[0] as keyof Filter;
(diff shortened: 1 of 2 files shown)


packages/core/helper-plugin/src/components/FilterListURLQuery.tsx
```diff
@@ -162,8 +186,17 @@ const AttributeTag = ({ attribute, filter, onClick, operator, value }: Attribute
   return (
     <Box padding={1}>
       <Tag onClick={handleClick} icon={<Cross />}>
-        {content}
+        <span onClick={handleContentClick}>{content}</span>
```
simotae14, line 189: please, span is a text, its behaviour is not to have an onClick, if you want to handle the onClick in different way you can add a condition in the handleClick handler, there is not need of two handlers for the same thing, you can differentiate the behaviour base on the event target element that triggered the handleClick for example

packages/core/helper-plugin/src/components/FilterListURLQuery.tsx
```diff
@@ -162,8 +186,17 @@ const AttributeTag = ({ attribute, filter, onClick, operator, value }: Attribute
...
       </Tag>
+      {isVisible && (
+        <FilterPopoverURLQuery
+          displayedFilters={filtersSchema}
+          isVisible={true}
+          onToggle={handleToggle}
+          source={buttonRef}
+          initialFilter={filter}
+        />
+      )}
```
simotae14, line 199: to change the position of this dropdown you need to assign the buttonRef to a button but in your case you are passing it just to the FilterPopoverURLQuery component
Please take a look how the FilterPopoverURLQuery is used in other places for example in this file packages/core/admin/admin/src/pages/Settings/components/Filters.tsx

simotae14 (changes requested):
Hello @moeez506, thanks a lot for your contribution,
please take a look at my comments and before submitting your code check locally if it passed all the necessary steps (run prettier command, unit and e2e tests, build the package you changed), thanks in advance

alexandrebodin:
HI @moeez506 we will be closing this PR as staled but if you eventually have time to work on it if we haven't done it in the future feel free to ping us to reopen 🙏  

Thank you for contributing!

PR #20999, opened 2024-08-12
enhancement!: restrict strapi prefixes in CTB

innerdvations:
### What does it do?

- adds support for prefixes in the ctb reserved names
- reserves `strapi*`, `_strapi*`, and `__strapi*` for model and attribute names in the CTB

### Why is it needed?

So that no further breaking changes are required to add new Strapi content types or attributes

### How to test it?

You should now receive an error if you try to create content types or attributes within a content type that start with 'strapi', '_strapi', or '__strapi'

### Related issue(s)/PR(s)

DX-1370

### DIFF
--- a/packages/core/content-type-builder/server/src/services/builder.ts
+++ b/packages/core/content-type-builder/server/src/services/builder.ts
@@ -1,44 +1,95 @@
+import { snakeCase } from 'lodash/fp';
+
+// use snake_case
+export const reservedAttributes = [
+  // TODO: these need to come from a centralized place so we don't break things accidentally in the future and can share them outside the CTB, for example on Strapi bootstrap prior to schema db sync
+
+  // ID fields
+  'id',
+  'document_id',
+
+  // Creator fields
+  'created_at',
+  'updated_at',
+  'published_at',
+  'created_by_id',
+  'updated_by_id',
+  // does not actually conflict because the fields are called *_by_id but we'll leave it to avoid confusion
+  'created_by',
+  'updated_by',
+
+  // Used for Strapi functionality
+  'entry_id',
+  'status',
+  'localizations',
+  'meta',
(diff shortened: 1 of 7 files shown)


packages/core/content-type-builder/server/src/services/builder.ts
```diff
@@ -1,44 +1,95 @@
...
+    return true;
+  }
+
+  if (
+    reservedModels
+      .filter((key) => key.endsWith('*'))
+      .map((key) => key.slice(0, -1))
+      .some((prefix) => snakeCaseName.startsWith(prefix))
+  ) {
+    return true;
```
jhoward1994, line 72:
```suggestion
  const wildcardPatterns = reservedModels
    .filter((key) => key.endsWith('*'))
    .map((key) => `^${key.slice(0, -1)}`)
    .join('|');

  const regex = new RegExp(wildcardPatterns);

  if (regex.test(snakeCaseName)) {
    return true;
  }
```

Would it be better to make this test using regex?

innerdvations: If we reused it or did anything else with it maybe, but I feel like in this case it's easier to understand and update the simple version. But I have no problem changing it if anyone else prefers it as a regex :)

jhoward1994 (approved):
The way the errors show in the CTB FE could be improved but I think this PR is all good!

https://github.com/user-attachments/assets/b8d93746-1bea-4d5a-af55-31f074b55d8e

innerdvations:
> The way the errors show in the CTB FE could be improved but I think this PR is all good!
> 
>  Screen.Recording.2024-08-19.at.10.28.04.mov

We definitely need to improve front-end reporting of errors. I'll add a ticket for it!

PR #21001, opened 2024-08-13
fix: trigger review worklfow webhooks

Marc-Roig:
### What does it do?

Review workflows `entryUpdate` webhook was not being triggered. 

Webhook is only triggered when using the document service, it listens to document updates. So the change involved using the document service instead of the used strapi.db.query

### DIFF
--- a/packages/core/review-workflows/server/src/services/document-service-middleware.ts
+++ b/packages/core/review-workflows/server/src/services/document-service-middleware.ts
@@ -94,6 +94,7 @@ const handleStageOnUpdate: Middleware = async (ctx, next) => {
     strapi.eventHub.emit(WORKFLOW_UPDATE_STAGE, {
       model: model.modelName,
       uid: model.uid,
+      // TODO v6: Rename to "entry", which is what it is used for regular CRUD updates
       entity: {
         // @ts-expect-error
         id: result?.id,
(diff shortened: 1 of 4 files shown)


packages/core/review-workflows/server/src/services/document-service-middleware.ts
```diff
@@ -94,6 +94,7 @@ const handleStageOnUpdate: Middleware = async (ctx, next) => {
     strapi.eventHub.emit(WORKFLOW_UPDATE_STAGE, {
       model: model.modelName,
       uid: model.uid,
+      // TODO v6: Rename to "entry", which is what it is used for regular CRUD updates
```
remidej, line 97:
```suggestion
      // TODO v6: Rename to "entry", which is what is used for regular CRUD updates
```

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -196,13 +197,8 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
     /**
      * Update the stage of an entity
-     *
-     * @param {object} entityInfo
-     * @param {number} entityInfo.id - Entity id
-     * @param {string} entityInfo.modelUID - the content-type of the entity
-     * @param {number} stageId - The id of the stage to assign to the entity
      */
-    async updateEntity(entityInfo: any, stageId: any) {
+    async updateEntity(documentId: string, locale: string, model: UID.ContentType, stageId: any) {
```
remidej, line 201: I wonder why you prefer prefer several params over one object?

remidej (commented):
Good catch! And thanks for the test. Just 1 question

remidej (approved):

PR #21007, opened 2024-08-13
V5 lifecycles codemods

Bassel17:
### What does it do?

Comments out lifecycles files, and add a comment describing why

### Why is it needed?

- Lifecycles will probably behave in unexpected ways for users in V5, they probably need to migrate to document service middlewares

### How to test it?

- run codemod from any example project
`../../packages/utils/upgrade/bin/upgrade.js codemods run 5.0.0-comment-out-lifecycle-files-code`
- all lifecycles files  will be commented out with a description at the top

### DIFF
--- a/packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
+++ b/packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
@@ -1,0 +1,62 @@
+import { Transform } from 'jscodeshift';
+
+/**
+ * comments out lifecycles.js/ts files and adds a description for the reason at the top
+ */
+const transform: Transform = (file, api) => {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  if (file.path.endsWith('lifecycles.js') || file.path.endsWith('lifecycles.ts')) {
+    // Get the entire source code as a string
+    const sourceCode = root.toSource();
+
+    // Split the source code into lines and prepend // to each line
+    // we are use line comments instead of block comments so we don't face issues with existing block comments
+    const commentedCode = sourceCode
+      .split('\n')
+      .map((line) => `// ${line}`)
+      .join('\n');
+
+    // Add a header comment at the top to explain why the file is commented out
+    const headerComment = `
+/*
+ *
+ * ============================================================
+ * WARNING: THIS FILE HAS BEEN COMMENTED OUT
+ * ============================================================
+ *
+ * CONTEXT:
+ *
+ * The lifecycles.js file has been commented out to prevent unintended side effects when starting Strapi 5 for the first time after migrating to the document service.
+ *
+ * STRAPI 5 introduces a new document service that handles lifecycles differently compared to previous versions. Without migrating your lifecycles to document service middlewares, you may experience issues such as:
+ *
+ * - \`unpublish\` actions triggering \`delete\` lifecycles for every locale with a published entity, which differs from the expected behavior in v4.
+ * - \`discardDraft\` actions triggering both \`create\` and \`delete\` lifecycles, leading to potential confusion.
+ *
+ * MIGRATION GUIDE:
+ *
+ * For a thorough guide on migrating your lifecycles to document service middlewares, please refer to the following link:
+ * [Document Services Middlewares Migration Guide](https://docs-next.strapi.io/dev-docs/migration/v4-to-v5/breaking-changes/lifecycle-hooks-document-service)
+ *
+ * IMPORTANT:
+ *
+ * Simply uncommenting this file without following the migration guide may result in unexpected behavior and inconsistencies. Ensure that you have completed the migration process before re-enabling this file.
+ *
+ * ============================================================
+ */
+`;
+
+    // Combine the header comment with the commented-out code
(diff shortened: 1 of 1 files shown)


packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
```diff
@@ -0,0 +1,62 @@
+import { Transform } from 'jscodeshift';
+
+/**
+ * comments out lifecycles.js/ts files and adds a description for the reason at the top
+ */
+const transform: Transform = (file, api) => {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  if (file.path.endsWith('lifecycles.js') || file.path.endsWith('lifecycles.ts')) {
```
Convly, line 10: Do you think we could be a bit more specific/precise when targeting the files? What if apps have `lifecycles` files dedicated to other purposes? :thinking:

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Get the entire source code as a string
+    const sourceCode = root.toSource();
+
+    // Split the source code into lines and prepend // to each line
+    // we are use line comments instead of block comments so we don't face issues with existing block comments
```
Convly, line 15:
```suggestion
    // we are using line comments instead of block comments so we don't face issues with existing block comments
```

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * STRAPI 5 introduces a new document service that handles lifecycles differently compared to previous versions. Without migrating your lifecycles to document service middlewares, you may experience issues such as:
+ *
+ * - \`unpublish\` actions triggering \`delete\` lifecycles for every locale with a published entity, which differs from the expected behavior in v4.
+ * - \`discardDraft\` actions triggering both \`create\` and \`delete\` lifecycles, leading to potential confusion.
+ *
+ * MIGRATION GUIDE:
+ *
+ * For a thorough guide on migrating your lifecycles to document service middlewares, please refer to the following link:
+ * [Document Services Middlewares Migration Guide](https://docs-next.strapi.io/dev-docs/migration/v4-to-v5/breaking-changes/lifecycle-hooks-document-service)
```
Convly, line 41: We'll need to update that with `docs.strapi.io` instead of `docs-next`. Do we want to do it now or right before the release?

Convly: _(I would personally vote to link the stable doc right now so that we don't forget)_

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
Bassel17: sure I can make it look for `lifecycles` files that are under `content-types` directory only

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
Bassel17: we certainly can do it now, It'll just be not working for now, the only concern would be people who are migrating now

Convly: I think it's fine if the doc link isn't available during RC, I prefer to be sure we have a valid link upon release so that the vast majority of users don't have issues :slightly_smiling_face:

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * comments out lifecycles.js/ts files and adds a description for the reason at the top
+ */
+const transform: Transform = (file, api) => {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  // check if file path follows this pattern `content-types/[content-type-name]/lifecycles`
+  if (/content-types\/[^/]+\/lifecycles\.(js|ts)$/.test(file.path)) {
```
Convly, line 11: It could've been easier to maintain to assert the path validity using glob, but that'll do. Does the path has to start with content-types? Or can it be `src/api/foo/content-types`? Will the regular expression work for this?

Convly (commented):
Just one question, the rest LGTM

packages/utils/upgrade/resources/codemods/5.0.0/comment-out-lifecycle-files.code.ts
Bassel17: `src/api/foo/content-types/foo/lifecycles.js/ts` will work for this yes, it'll match the regex, this is already the case if you test the codemod

Convly (approved):

innerdvations (approved):
LGTM!

PR #21009, opened 2024-08-14
feat(i18n): fill in from another locale

Feranchz:
### What does it do?

Add the "Fill in from another locale" button to v5.

### How to test it?

1. Go to a entry with i18n enabled.
2. Click on the button to the right of the locale picker 
3. On the dialog, select one locale to copy content


simotae14 (commented):
Hey Fernando, I found some issues

First of all when I switch the locale I have some notification errors (I am not sure they are related to your code)

Second issue, when I try to fill the locale from English to the other locale in the Confirmation modal the dropdown is empty

you can see these issues in this video


https://github.com/user-attachments/assets/8d2327bd-893e-4a47-bd75-f7f73cfce28f

Feranchz:
Good catch @simotae14 thank you! now should be ready again. I will add @Marc-Roig as reviewer as well because now there are some changes on the backend

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -209,7 +209,7 @@ export default {
         permissionChecker,
         model,
         // @ts-expect-error TODO: fix
-        { id, locale, publishedAt: null },
+        { documentId: id, locale, publishedAt: null },
```
Marc-Roig, line 212: was this working without the document id 😮

simotae14 (approved):
Overall, the code looks good, and I've tested the branch—it now works as expected. However, I have one concern, not related to your work but regarding the UI solution chosen for this feature. I'm not entirely convinced that the button with its current icon is easily recognizable as the trigger for filling from another locale. Maybe we can involve @lucasboilly  on this

packages/core/content-manager/server/src/controllers/collection-types.ts
Feranchz: Not as expected, for this reason it was returning always empty `availableLocales`. Probably we don't noticed it because this case only happen when you are creating a new entry and we don't need the availableLocales when creating for other thing different to fill in from another locale 🤔

Marc-Roig: dont we need it to list the available locales in the list of the document locales? I thought that was working before

Marc-Roig: could we add some tests for that ? 🙏 in the content manager api tests

lucasboilly:
Hello guys, it seems our users were not disturbed by this icon during the research phase, but it's not the first time I hear this feedback internally so I tried to create a custom icon, what do you think of it?

CleanShot 2024-08-22 at 11 47 31@2x

Also I tested the branch and I noticed 2 UI issues:
- The focused state of the Select is kind of broken, like if it was "blocked" by its container
- The IconButton is not a square, it's like "crushed" on the right by the Select

CleanShot 2024-08-22 at 11 42 09@2x

packages/plugins/i18n/admin/src/translations/en.json
```diff
@@ -6,7 +6,12 @@
   "CMEditViewCopyLocale.copy-failure": "Failed to copy locale",
   "CMEditViewCopyLocale.copy-success": "Locale copied",
   "CMEditViewCopyLocale.copy-text": "Fill in from another locale",
+  "CMEditViewCopyLocale.cancel-text": "No, cancel",
   "CMEditViewCopyLocale.submit-text": "Yes, fill in",
+  "CMEditViewCopyLocale.dialog.title": "Confirmation",
+  "CMEditViewCopyLocale.dialog.body": "Your current content will be erased and filled by the content of the selected locale:",
+  "CMEditViewCopyLocale.dialog.field.label": "Locale",
```
jhoward1994, line 13:
```suggestion
  "CMEditViewCopyLocale.dialog.body": "Your current content will be erased and filled by the content of the selected locale",
```

Is the colon intended it reads a bit weirdly to me

jhoward1994:
For me I think we could just use the globe icon from the DS for this - I agree the download icon doesn't look quite right

jhoward1994:
I'm seeing a bug that I think is related to repeatable components. I thought it was only if the component had a relation field type but it seems to happen regardless of component schema

https://github.com/user-attachments/assets/a3d0a36a-5444-4c46-9b55-6c3e616ba70e

Should components also be excluded from the "Fill from" logic?

derrickmehaffy:
> I'm seeing a bug that I think is related to repeatable components. I thought it was only if the component had a relation field type but it seems to happen regardless of component schema
> 
>  Screen.Recording.2024-08-22.at.12.08.46.mov 
> Should components also be excluded from the "Fill from" logic?

I think that depends if the component is localized or not? (I'm not even sure if you can set that) or if the component has localized fields within it (also not sure if that's possible).

My gut says we should copy the component and dz data personally

simotae14:
You need to solve a lint error

Feranchz:
> > I'm seeing a bug that I think is related to repeatable components. I thought it was only if the component had a relation field type but it seems to happen regardless of component schema
> > Screen.Recording.2024-08-22.at.12.08.46.mov
> > Should components also be excluded from the "Fill from" logic?
> 
> I think that depends if the component is localized or not? (I'm not even sure if you can set that) or if the component has localized fields within it (also not sure if that's possible).
> 
> My gut says we should copy the component and dz data personally

@derrickmehaffy That's a good point, components can't be localised but they could have relations and these relation could be with a localised Content Type. 

in my opinion the best for now is ignore components and dz as well. We could discuss then if copy components & dz is useful and in that case find a way to do it ignoring localised fields inside

Marc-Roig:
@Feranchz are we copying regular relations already?
I believe in v4 we were copying compos and dz, just not its relations. In a similar way as top relations. 
In my opinion, if possible, we should also be copying compos when filling from another locale

Feranchz:
> @Feranchz are we copying regular relations already? I believe in v4 we were copying compos and dz, just not its relations. In a similar way as top relations. In my opinion, if possible, we should copy compos and dz when filling from another locale , without its relations

We are not copying regular relations, but we are trying to copy them inside components, I'm working on fix that like you said

oiorain:
It seems I still have the empty dropdown 

Screenshot 2024-08-29 at 15 20 08


- I filled in the english draft
- selected french in dropdown
- clicked the button
- this popup shows with no locale to select from

Marc-Roig:
I do have an empty dropdown when there are no available locales to fill in from.

Specially, during entry creation, I don't think it makes sense to keep in the fill from other locales, should the button at least be disabled?

Feranchz:
> I do have an empty dropdown when there are no available locales to fill in from.
> 
> Specially, during entry creation, I don't think it makes sense to keep in the fill from other locales, should the button at least be disabled?

@Marc-Roig That's a great point, I just added a check for that

@oiorain I think latest change would also fix your case, be sure to run `yarn setup` because there are some changes also in the content-manager & APIs 🤔 Let me know if it's still not working

Marc-Roig (approved):
you are a super ⭐

oiorain:
yes! thanks Fernando. it works now

Issue #20604, opened 2024-06-21
Filtering on relations does not populate data like the unfiltered list does in Admin UI


verdverm:
### Required System information

- Node.js version: 20
- NPM version:
- Strapi version: v5.0.0-beta.14
- Database:
- Operating system:
- Is your project Javascript or Typescript:

### Describe the bug

Following up on #20058, while this no longer errors, the result is missing the relation data for the field that is filtered on

### Steps to reproduce the behavior

Same as #20058

### Expected behavior

Should still be able to see the data in the relation. In this case, we are setting multiple tags on documents. We want to be able to search by tag and have the results

1. show the tags that match the filter
2. show all tags, so that we might refine our filter(s)

Generally, I would expect the same relation population to occur in the filtered list as the unfiltered list.
The relation population does work if filtering by a direct field on the content-type

It would be great if the UI colored the matching entries different from the non-matching.

### Screenshots

Relation filter, not populated

<img width="1071" alt="Screenshot 2024-06-21 at 12 55 23 PM" src="https://github.com/strapi/strapi/assets/1390600/46c8af67-a092-4125-b70e-46d5c3c6358c">

Field filter, does populate

<img width="1088" alt="Screenshot 2024-06-21 at 1 03 37 PM" src="https://github.com/strapi/strapi/assets/1390600/71da4e44-08c7-4114-8332-032406a28c21">

derrickmehaffy:
Confirmed on v5.0.0-rc.6

alexandrebodin:
FIxed in rc.13 :)

PR #21010, opened 2024-08-14
fix: show relations to draft entries in list view

remidej:
### What does it do?

- Fixes an issue where the relations were not populated in the list view when the related entries were drafts.
- Also fixes an issue where the Relation Input component would display a relation to a published entry twice (once as draft, once as published)
- Re-enables an API test suite for the CM

### How to test it?

- create an Address in getstarted
- link it to some draft Categories
- you should see the relations count in the addresses list view
- publish some categories
- the relations count in the addresses list view should not change
- opening the address edit view, you should only see one Category per actual category in the relation input

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20604

### DIFF
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -137,10 +137,10 @@ export default {
       .countRelations({ toOne: false, toMany: true })
       .build();
 
-    const { locale, status } = await getDocumentLocaleAndStatus(query, model);
+    const { locale } = await getDocumentLocaleAndStatus(query, model);
 
     const { results: documents, pagination } = await documentManager.findPage(
-      { ...permissionQuery, populate, locale, status },
+      { ...permissionQuery, populate, locale },
       model
     );
 
--- a/packages/core/content-manager/server/src/controllers/relations.ts
+++ b/packages/core/content-manager/server/src/controllers/relations.ts
@@ -454,7 +454,7 @@ export default {
       ordering: 'desc',
     });
 
-    const relationsUnion = uniqBy('id', concat(sanitizedRes.results, res.results));
+    const relationsUnion = uniqBy('documentId', concat(sanitizedRes.results, res.results));
 
     ctx.body = {
       pagination: res.pagination || {
--- a/packages/core/content-manager/server/src/controllers/relations.ts
+++ b/packages/core/content-manager/server/src/controllers/relations.ts
@@ -454,7 +454,7 @@ export default {
       ordering: 'desc',
     });
 
-    const relationsUnion = uniqBy('documentId', concat(sanitizedRes.results, res.results));
+    const relationsUnion = uniqBy('id', concat(sanitizedRes.results, res.results));
 
     ctx.body = {
       pagination: res.pagination || {
--- a/tests/api/core/content-manager/index.test.api.js
+++ b/tests/api/core/content-manager/index.test.api.js
@@ -281,7 +281,7 @@ describe('Relations', () => {
 
       expect(body.results).toMatchObject([
         { title: 'Article 1', tags: { count: 3 } },
-        { title: 'Article 2', tags: { count: 0 } },
+        { title: 'Article 2', tags: { count: 1 } },
       ]);
     });
 
--- a/tests/api/core/content-manager/index.test.api.js
+++ b/tests/api/core/content-manager/index.test.api.js
@@ -273,6 +273,18 @@ describe('Relations', () => {
       expect(tags.length).toBe(3);
     });
 
+    test('Get tags count in articles list', async () => {
+      const { body } = await rq({
+        url: '/content-manager/collection-types/api::article.article',
+        method: 'GET',
+      });
+
+      expect(body.results).toMatchObject([
+        { title: 'Article 1', tags: { count: 3 } },
+        { title: 'Article 2', tags: { count: 0 } },
+      ]);
+    });
+
     test('Update article1 remove one tag', async () => {
       const { body } = await rq({
         url: `/content-manager/collection-types/api::article.article/${data.articles[0].documentId}`,
--- a/tests/api/core/content-manager/index.test.api.js
+++ b/tests/api/core/content-manager/index.test.api.js
@@ -23,7 +23,7 @@ const getRelations = async (modelName, field, id) => {
     url: `/content-manager/relations/api::${modelName}.${modelName}/${id}/${field}`,
   });
 
-  return res.body.data;
+  return res.body;
 };
 
 const deleteFixtures = async () => {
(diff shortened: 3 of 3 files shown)


tests/api/core/content-manager/index.test.api.js
```diff
@@ -273,6 +273,18 @@ describe.skip('Relations', () => {
...
+    test('Get tags count in articles list', async () => {
+      const { body } = await rq({
+        url: '/content-manager/collection-types/api::article.article',
+        method: 'GET',
+      });
+
+      expect(body.results).toMatchObject([
+        { title: 'Article 1', tags: { count: 3 } },
+        { title: 'Article 2', tags: { count: 0 } },
+      ]);
```
remidej, line 285: Ideally I'd want to test that the count doesn't change when publishing an unpublishing Tags. But Strapi doesn't seem to pick up when I enable D&P in the Tag schema in `packages/utils/api-tests/generators.js`

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -137,10 +137,10 @@ export default {
       .countRelations({ toOne: false, toMany: true })
       .build();
 
-    const { locale, status } = await getDocumentLocaleAndStatus(query, model);
+    const { locale } = await getDocumentLocaleAndStatus(query, model);
 
     const { results: documents, pagination } = await documentManager.findPage(
-      { ...permissionQuery, populate, locale, status },
+      { ...permissionQuery, populate, locale },
```
Marc-Roig, line 143: isn't the status necessary to list published versions?

packages/core/content-manager/server/src/controllers/relations.ts
```diff
@@ -454,7 +454,7 @@ export default {
       ordering: 'desc',
     });
 
-    const relationsUnion = uniqBy('id', concat(sanitizedRes.results, res.results));
+    const relationsUnion = uniqBy('documentId', concat(sanitizedRes.results, res.results));
```
Marc-Roig, line 457: this might not work as expected if an entry has a relation to multiple locales of the same documentId.

Why was it necessary to make this change?

remidej: Ah I didn't think this was possible, thanks for raising the flag.

I found it to be needed as it was returning published content in duplicates: it returned both the draft and the published version.

So I guess I need to check uniqueness based the combination of documentId and locale

packages/core/content-manager/server/src/controllers/collection-types.ts
remidej: I don't think so, without it I'm seeing it both published and draft/modified entries in the list view

The reason for my change is that requiring `status: published` explicitly here filtered out the draft relations of the entry, wish should be visible in the CM

Marc-Roig: I don't see how the published data version would be listed if we don't specify the status 🤔 if you filter by published entries, you should see the published data on the list view. But after this change would assume users would see the "draft" entry

packages/core/content-manager/server/src/controllers/relations.ts
remidej: Oh okay, I'll stop making assumptions then 😄. But if I know that I assigned only one relation, then the endpoint should only return one entry, right? It's currently not the right fix, but I feel like the fix should still be in the backend

Marc-Roig: Yep! That should be the case, if the endpoint returns more than one we need to see why is that.
Which type of relation are you using? Do both sides have D&P enabled?

Marc-Roig: This might be the scenario you are facing:

Category (No DP) -> Article (DP)

You create a relation to a draft article, you publish the article, and then a new relation would be created from the category to the published version of the article.

Thats an intended behaviour, as the published version of the article should also contain that relation, but probably the way it's displayed is undesired. Is that the case?

Marc-Roig: totally! 
Then I also raise the question of, when trying to conect a relation, should you be able to connect to the draft Test and the  published Test, or only the Test document in itself?

Marc-Roig: Something similar happens with locales, but it makes more sense to show different "locales" of the same document. Indeed it feels confusing to show different publication states.

Regardless, should we show they belong to the same document somehow? Should we discuss this with design?

remidej: I thought you could only connect to the document (or at least that you couldn't pick between draft/publish version of the related content). That's how the input currently works. Otherwise I feel like it would put a lot of complexity on the users, and probably lead to behaviors they don't expect

But if there's ambiguity here, I think we need to check with Alex or a PM to not make mistakes since it's quite important

Marc-Roig: Indeed,
technically we are connecting entries, both the published and draft versions of the other side of the relation need a different relation, so effectively there need to be two relations in the database.

We can sort how things are displayed by only showing draft or published versions though, your approach (only showing the draft version) sounds like the most sensible approach imo, but lets discuss this with a PM

packages/core/content-manager/server/src/controllers/collection-types.ts
remidej: Currently it's not possible to filter by published entries in the list view, though that may be a bug. In the filter select, there's on status option, no publishedAt either.

>Even if marked as "status" = "published" the displayed data would be the draft one

if the status if actually published, that would make no difference right? I think it would only change for a draft/modified entry. And in that case I thought we always wanted to show the more recent draft version, and leave the published one only to the published tab

Marc-Roig: The displayed data would be different, as the draft & published data might differ.
Even if the feature is not enabled, I don't think we should remove that filter though, as I don't think that is the "cause" of the issue, but rather a patch to make it work.

Should we not filter by "status" if D&P is not enabled? We could keep that value as undefined by default

packages/core/content-manager/server/src/controllers/relations.ts
yanniskadiri: I would only show the published version of the relation as it reflects the finalized content that has been reviewed, approved and published. The draft version may contain incomplete information.

Marc-Roig: but then you could not connect to any document that's in draft, or see if it's connected to the draft version of the document. 
It could lead to a situation where you see a draft Article connected to a Category, but not see that article from the category side, not until you publish the article.

Marc-Roig: This could be an oportunity to start discussing about relation fallbacks (show published version by default, or fallback to the draft verssion). But this would require alex

Issue #19306, opened 2024-01-23
Strapi v4.17.1 fresh install yields 13 high risk vulnerabilities


ProfXponent:
### Required System information
- Is your project Javascript or Typescript: ts

Environment: development
OS: linux-arm64
Strapi Version: 4.17.1
Node/Yarn Version: npm/10.2.3 node/v18.19.0 linux arm64 workspaces/false
Edition: Community
Database: unknown

### Describe the bug

Clean install of strapi results in 13 high vulnerability issues present in `npm audit`.

### Steps to reproduce the behavior

1. Fresh install in docker container
2. run `npm audit`

### Expected behavior

A fresh install of a cms intended for production usage should not introduce security vulnerabilities into the target applications stack.

### Additional context

```shell
1302a5923ddc:/opt/app$ npm audit
# npm audit report

@koa/cors  <5.0.0
Severity: high
Overly permissive origin policy - https://github.com/advisories/GHSA-qxrj-hx23-xp82
No fix available
node_modules/@koa/cors
  @strapi/strapi  *
  Depends on vulnerable versions of @koa/cors
  Depends on vulnerable versions of @strapi/admin
  Depends on vulnerable versions of @strapi/content-releases
  Depends on vulnerable versions of @strapi/data-transfer
  Depends on vulnerable versions of @strapi/pack-up
  Depends on vulnerable versions of @strapi/plugin-content-manager
  Depends on vulnerable versions of @strapi/plugin-content-type-builder
  Depends on vulnerable versions of @strapi/plugin-upload
  Depends on vulnerable versions of @strapi/types
  node_modules/@strapi/strapi
    @strapi/admin  <=0.0.0-next.ff946d2c25a3e577b47132a357cac2932eb8e635 || >=4.3.5
    Depends on vulnerable versions of @strapi/data-transfer
    Depends on vulnerable versions of @strapi/strapi
    Depends on vulnerable versions of @strapi/types
    node_modules/@strapi/admin
    @strapi/content-releases  *
    Depends on vulnerable versions of @strapi/strapi
    Depends on vulnerable versions of @strapi/types
    node_modules/@strapi/content-releases
    @strapi/data-transfer  *
    Depends on vulnerable versions of @strapi/strapi
    Depends on vulnerable versions of @strapi/types
    node_modules/@strapi/data-transfer
    @strapi/plugin-cloud  *
    Depends on vulnerable versions of @strapi/strapi
    node_modules/@strapi/plugin-cloud
[...]

PR #21014, opened 2024-08-15
chore(deps): update axios to 1.7.4

Ohiekkar:
### What does it do?

Bump axios from 1.6.0 to 1.7.4

### Why is it needed?

To solve https://github.com/advisories/GHSA-8hc4-vh64-cxmj

### How to test it?

-

### Related issue(s)/PR(s)

-

### DIFF
--- a/packages/cli/cloud/package.json
+++ b/packages/cli/cloud/package.json
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@strapi/utils": "4.25.8",
-    "axios": "1.6.0",
+    "axios": "1.7.4",
     "chalk": "4.1.2",
     "cli-progress": "3.12.0",
     "commander": "8.3.0",
--- a/packages/core/admin/package.json
+++ b/packages/core/admin/package.json
@@ -81,7 +81,7 @@
     "@strapi/typescript-utils": "4.25.8",
     "@strapi/utils": "4.25.8",
     "@vitejs/plugin-react-swc": "3.5.0",
-    "axios": "1.6.0",
+    "axios": "1.7.4",
     "bcryptjs": "2.4.3",
     "boxen": "5.1.2",
     "browserslist": "^4.22.2",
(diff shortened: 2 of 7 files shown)


packages/cli/cloud/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@strapi/utils": "4.25.8",
-    "axios": "1.6.0",
+    "axios": "1.7.4",
```
gusdecool, line 49: Should we use version `^1.7.4` which will handle the update automatically for us in the future?

usamawork61:
When is this likely to get merged?

HonzaTuron:
Would be great to merge & release as soon as possible as vulnerability has high severity.

jbartadev:
Guys, can we please prioritize merging this (for [security reasons](https://github.com/advisories/GHSA-8hc4-vh64-cxmj))?

CC @soupette @alexandrebodin

HeikkiYlipaavalniemi:
There seems to be a really long wait time in this project regarding security issues. I opened a security issue related to this same axios npm audit report a week ago and haven't received any reply from the project maintainers.

To me a response time of a week to a high security risk in NPM is not good enough and there should be more prompt response from the core maintainers.

It also seems that there are situations when high risk vulnerabilities are not treated as such like mentioned in https://github.com/strapi/strapi/issues/19306 if there is no proof of actual security risk to Strapi. But this does not take into account for example zero day security issues or actual project risks where client environment could already be compromised but without a proper response from Strapi team where the issue would be taken into account and a fix available.

Marc-Roig (approved):

packages/cli/cloud/package.json
gusdecool: Hi @Marc-Roig thank you for your reply.
Can you please clarify the issue you experienced with Axios versioning. 

Is it the issue on Axios end when we do minor upgrade? Because if they follow Semver, it should not causing issue. But who knows, could be something that I missing.

Marc-Roig: Absolutely! We have never encountered any issue in particular with Axios, but unfortunately and in practice, some packages do introduce breaking changes even in patch updates, so we prefer to stay on the safe side an apply this rule for all packages.

PR #21023, opened 2024-08-19
Add codemod for useRBAC import statement

Bassel17:
### What does it do?

- Adds a codemod that changes useRBAC imports from `@strapi/helper-plugin` to `@strapi/strapi/admin`

### How to test it?

you can test in any of the example folders using:
`../../packages/utils/upgrade/bin/upgrade.js codemods run 5.0.0-useRBAC-hook-import-change-code`


jhoward1994:
I can't work out exactly how this works but it isn't changing the imports for me.

I think it's because `this.paths = []` here when I run the codemod 
https://github.com/strapi/strapi/blob/975a81213f3debcd5543feca619b4a2d2d017a2e/packages/utils/upgrade/src/modules/runner/runner.ts#L19-L30

For example when I run `packages/utils/upgrade/resources/codemods/5.0.0/sqlite3-to-better-sqlite3.json.ts` this.paths includes the package JSON of the plugin and that file is affected.

https://github.com/user-attachments/assets/3c4d9eea-9cdd-4b3f-83d9-37ba5a66f843

Bassel17:
yeah code codemods for some reason don't seem to be running in plugins, you just uncovered a new issue @jhoward1994 
I'll look into it

packages/utils/upgrade/resources/codemods/5.0.0/useRBAC-hook-import-change.code.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (methodImportedFromOldDependency) {
+    const dependencies = root
+      .find(j.ImportDeclaration)
+      .filter((path) => path.node.source.value === newDependency);
+
+    dependencies.forEach((path) => {
+      const importDeclaration: ImportDeclaration = path.node;
+
+      const newSpecifier = j.importSpecifier(j.identifier(methodName));
+      importDeclaration.specifiers?.push(newSpecifier);
```
innerdvations, line 54:
```suggestion
      const importDeclaration: ImportDeclaration = path.node;

      const newSpecifier = j.importSpecifier(j.identifier(methodName));
      const specifiersArray = importDeclaration.specifiers || [];
      j(path).replaceWith(
        j.importDeclaration([...specifiersArray, newSpecifier], j.literal(newDependency))
      );
```
We can avoid mutating the AST by using `replaceWith`

packages/utils/upgrade/resources/codemods/5.0.0/useRBAC-hook-import-change.code.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+    // Add the new import declaration if it doesn't already exist
+    if (dependencies.length === 0) {
+      const newImportDeclaration = j.importDeclaration(
+        [j.importSpecifier(j.identifier(methodName))],
+        j.literal(newDependency)
+      );
+      root.get().node.program.body.unshift(newImportDeclaration);
```
innerdvations, line 62:
```suggestion
      // Find the index of the first non-import declaration
      const body = root.get().node.program.body;
      const lastImportIndex = body.findIndex(node => node.type !== 'ImportDeclaration');

      // Check if there are any import declarations
      if (lastImportIndex > -1) {
        // Insert the new import declaration just before the first non-import node
        body.splice(lastImportIndex, 0, newImportDeclaration);
      } else {
        // Check if 'use strict' exists at the beginning
        const hasUseStrict = body[0]?.type === 'ExpressionStatement' && body[0]?.expression?.value === 'use strict';

        // Add the new import after 'use strict' if it exists, otherwise at the beginning
        body.splice(hasUseStrict ? 1 : 0, 0, newImportDeclaration);
      }
```

This will add the new import directly after the import it was removed from, rather than adding it at the beginning. That will keep all the strapi imports together instead of separating them.

To see an example of the problem, start with this code:

```
import React from 'react';
import { Button } from '@strapi/design-system';
import { useRBAC, useOtherMethod } from "@strapi/helper-plugin";
```

the original codemod transforms it to:
```
import { useRBAC } from "@strapi/strapi/admin";
import React from 'react';
import { Button } from '@strapi/design-system';
import { useOtherMethod } from "@strapi/helper-plugin";
```

the updated code transforms it to:
```
import React from 'react';
import { Button } from '@strapi/design-system';
import { useOtherMethod } from "@strapi/helper-plugin";
import { useRBAC } from "@strapi/strapi/admin";
```

innerdvations (changes requested):

innerdvations:
One more issue: it doesn't handle aliases.

```
import { useRBAC as handleAPIError } from '@strapi/helper-plugin';
```

will become

```
import { useRBAC } from '@strapi/admin';
```

which will break user code.

When I discovered it, I just asked chatGPT to solve it and it came up with a solution pretty easily 😆 but it requires changes in a few places so I can't do a code suggestion. If you need help just let me know and I'll show you the solution.

Bassel17:
we have an issue with code codemods and plugins, they will not work on plugins because we have set of allowed directories, I added `admin` and `server` to make it work for plugins, but technically users can do anything they want so that's not a flexible solution, I was wondering why we even have it, we can just make it run all directories or files in the root, we can exclude directories we don't want instead 

@innerdvations @Convly @jhoward1994  Opinions please

jhoward1994:
> we have an issue with code codemods and plugins, they will not work on plugins because we have set of allowed directories, I added `admin` and `server` to make it work for plugins, but technically users can do anything they want so that's not a flexible solution, I was wondering why we even have it, we can just make it run all directories or files in the root, we can exclude directories we don't want instead

Ah I see in most cases this should cover everything for plugin developers right?

Could we add an option to the CLI if users wanted to cover other directories?

Something like ?
`.addOption(new Option('-a, --additional-paths <paths>', 'Additional allowed root paths').split(','))`
in packages/utils/upgrade/src/cli/commands/codemods.ts

Bassel17:
@jhoward1994 yep will move with the current fix in this PR, we can think of a better solution and change things later
after all plugin devs can technically rename the directories it doesn't really matter as long as you are registering the plugin properly

innerdvations (approved):
Awesome! Great utility, and works fine for everything I tried including multiple aliases across multiple exports.

Convly (approved):

jhoward1994 (approved):

derrickmehaffy:
Good idea, plugins are the wild west indeed

PR #21025, opened 2024-08-19
change useAPIErrorHandler import to `@strapi/strapi/admin` codemod

Bassel17:
### What does it do?

Codemod to change useAPIErrorHandler import to `@strapi/strapi/admin`

### How to test it?

you can test in any of the example folders using:
`../../packages/utils/upgrade/bin/upgrade.js codemods run 5.0.0-change-useAPIErrorHandler-import-code`


innerdvations (changes requested):
Same three issues as https://github.com/strapi/strapi/pull/21023, but otherwise looks good

innerdvations:
Considering this does almost the exact same thing as the other PR, and we will have more coming (and might have some older ones with the same bugs?), I wonder if you could abstract this to a utility so that we don't have to repeat ourselves and have to fix bugs across multiple codemods every time we find one.

Issue #20945, opened 2024-08-06
Content History is never found


remidej:
On the latest v5 RC, it's impossible to access history, it always says "Content not found".

![2024-08-05_09-49-26 (1)](https://github.com/user-attachments/assets/c3beb06d-3673-4601-b243-c51aba8a0163)

PR #21026, opened 2024-08-19
fix(content-manager): error when content type has no i18n enabled

Feranchz:
### What does it do?

Fix a bug related to the history page of content type without localisation enabled. Now we considered these type of content types and we allow saving history of content types with no i18n.

### How to test it?

1. Go to a Content Type with no i18n enabled and create and edit one entry.
2. Go to the history page of that entry
3. It should show the history without errors

### Related issue(s)/PR(s)

closes https://github.com/strapi/strapi/issues/20945

### DIFF
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -1,4 +1,5 @@
 import type { Core, Modules, UID } from '@strapi/types';
+import { contentTypes } from '@strapi/utils';
 
 import { omit, castArray } from 'lodash/fp';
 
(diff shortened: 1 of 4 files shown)


packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -131,13 +131,20 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
...
           where: {
             documentId,
-            locale: { $in: locales },
-            publishedAt: null,
+            ...(isLocalizedContentType
+              ? {
+                  locale: { $in: locales },
+                  publishedAt: null,
+                }
+              : {}),
```
Marc-Roig, line 147: the published at filter should always apply, wether the content type is localized or not

Marc-Roig (changes requested):

packages/core/content-manager/server/src/history/services/lifecycles.ts
Feranchz: But entries without draft & publish are always published, so they have publishedAt with a date value (same as created I think)

Marc-Roig: true, then that would require another check, wether the content type has D&P enabled

Marc-Roig (approved):
amazing! ✨

Marc-Roig (approved):

PR #19676, opened 2024-03-05
Fix: issue when end received before previous data finishes writing

chrisli-03:
### What does it do?

Describe the technical changes you did.

This will add a counter to each payload and only close the stream after all payloads are written to disk.

### Why is it needed?

On rare occasions end action will happen before write finishes, causing error at await this.writeAsync(assets[item.assetID].stream, chunk);

### How to test it?

Run transfer with many files to increase likelyhood of error happening

### Related issue(s)/PR(s)

N/A

PR #21032, opened 2024-08-20
fix(data-transfer): sequential payload consumption

meeehdi-dev:
Hello,

Here's a fix PR for data transfers.

Thank you.

### What does it do?

It fixes an issue with data transfer when two stream payloads are received at the same time, and one of them contains the "end" action for an asset transfer, while the other still has "stream" actions.
There is a 100ms sleep looping while waiting for the previous payload to finish being read, this delay can be changed if you prefer another value.

### Why is it needed?

It solves data transfer in some cases with bigger files.

### How to test it?

The repo I tested this on is private and owned by my client. But I guess you could create a remote strapi instance with big media (videos > 50MB) and check if that happens for you.
Without this fix, it's a dice roll, it sometimes works and sometimes not, but with this, it's working in all cases.

### Related issue(s)/PR(s)

I'm guessing a lot of data transfer issues could be related to this fix, but I didn't check.


derrickmehaffy:
@innerdvations I say we merge this and test it in the wild. Like we discussed it'll either do something to assist or nothing but it shouldn't hurt anything

innerdvations:
> @innerdvations I say we merge this and test it in the wild. Like we discussed it'll either do something to assist or nothing but it shouldn't hurt anything

Agreed, this looks like a very similar alternative to the [same issue this solves](https://github.com/strapi/strapi/pull/19676), and now I actually have a clear idea of WHY the other one worked. I'll think about it and discuss it tomorrow and we'll get one of them merged ASAP. Even if we can't reproduce it internally, there are now PRs from two different people who have solved their problem in the same basic way, so that's good enough for me.

derrickmehaffy:
> > @innerdvations I say we merge this and test it in the wild. Like we discussed it'll either do something to assist or nothing but it shouldn't hurt anything
> 
> Agreed, this looks like a very similar alternative to the [same issue this solves](https://github.com/strapi/strapi/pull/19676), and now I actually have a clear picture of what is actually going on that caused the other PR to work. I'll think about it and discuss it tomorrow and we'll get one of them merged ASAP. Even if we can't reproduce it internally, there are now PRs from two different people who have solved their own problem in the same basic way, so that's good enough for me.

Ah I want paying attention I thought it was the same one but your right

meeehdi-dev:
@derrickmehaffy @innerdvations 

Hey guys, thank you for looking at this PR.
I didn't see the issue you mentioned when I was searching to avoid any duplicates, so sorry for that.

But I was wondering, do you think there could be a way to avoid this issue directly from the source ? Why does the stream process send a second package even tho the first one hasn't finished loading ?

image

innerdvations:
@meeehdi-dev No worries, and thank you very much for your PR! You provided some additional information that finally made the issue clear, plus I've been waiting for someone to confirm it since I haven't been able to reproduce on my end.

You're right that we should fix this at the source, we just haven't had the time to debug it. But since these PRs offer a functional fix for an issue many users are facing, we can start here with plans to fix it the right way (as written in the [comment](https://github.com/strapi/strapi/pull/19676/files/64badd64d72c6c823a6ce772656aaaf48d1c72c4#r1687779203) I added in the other PR) 

If you're able to find the cause of the multiple streams that would be fantastic, but admittedly that part of the code is overly complicated right now so it may be a bigger task than anticipated.

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
```diff
@@ -168,6 +180,8 @@ class RemoteStrapiSourceProvider implements ISourceProvider {
             });
           }
         }
+
+        currentPayloadIndex++;
```
innerdvations, line 184:
```suggestion
        currentPayloadIndex = currentPayloadIndex + 1;
```
for the linting

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
```diff
@@ -129,8 +129,20 @@ class RemoteStrapiSourceProvider implements ISourceProvider {
     const stream = await this.#createStageReadStream('assets');
     const pass = new PassThrough({ objectMode: true });
 
+    let payloadIndexer = 0;
+    let currentPayloadIndex = 0;
+
     stream
       .on('data', async (payload: Protocol.Client.TransferAssetFlow[]) => {
+        const payloadIndex = payloadIndexer;
```
innerdvations, line 137:
```suggestion
        // TODO: fix the race condition causing multiple streams to open at once so that this check can be removed
        // Keep track of streams and wait to prevent bugs with parallel processing
        const payloadIndex = payloadIndexer;
```

Convly (commented):
Hey, thanks for submitting this PR. I read through the discussion, but in the end I think keeping the status of each file independently would allow more concurrence in the future, rather than having a global lock (a bit like the other PR does). What do you think?

meeehdi-dev:
@Convly I was also thinking that keeping some parallelism would be better, but you still need to make sure that for the same asset is written in the correct order (I think).

If you look again at the drawing I did, let's say:
- You receive `data1` that contains chunks from `assetID1`
- You receive `data2` that also contains chunks from `assetID1`, while still writing the file with `data1`'s chunks

If the `writeAsync` function needs to receive the chunks in the correct order, then you have to make sure to delay `data2` chunks for `assetID1` while finishing to write the first chunks contained in `data1`.

My PR is lazier but feels safer to me, but if you prefer the other one, as long as real world scenarios are fixed, that's the main concern.

### At the end of the day, what's the final goal ? Should multiple streams arrive at the same time or no ?
If **yes**, then the other PR should be merged with the improvement I talked about, as it would allow parallelism and correct chunk order.
If **no**, choose whatever PR right now, and we'll focus on finding the root cause as to why this happens in the first place.

derrickmehaffy:
@meeehdi-dev I just wanted to pop in here and say thank you from a previous community member now turned Strapi team member.

I miss the early days of Strapi when we had really interesting PRs and discussions like this.

🥰

Issue #20944, opened 2024-08-05
Strapi 5 Upgrade tool does not properly detect Old plugin structure to execute proper codemods (es to ds)


derrickmehaffy:
### Required System information

- Node.js version: node/v20.11.1
- NPM version: yarn/1.22.22
- Strapi version: 5.0.0-rc.8
- Database: SQLite
- Operating system: Linux Mint 21.3
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When running the upgrade tool against a plugin directly, the upgrade tool does not properly detect the plugin structure to run the codemods properly. In this example running the upgrade tool against a newly generated Strapi plugin from v4 (with some minor modification to add the entityService.

Ex: `npx @strapi/upgrade@rc codemods run --project-path ./src/plugins/test-plugin`

### Steps to reproduce the behavior

1. Create a plugin in v4 (using the legacy `yarn strapi generate plugin` command
2. Copy that plugin to a Strapi RC project
3. Setup the plugin properly as normal (install modules, ect)
4. Try to run the upgrade tool
5. See that only 2 of the codemods actually did anything since Strapi 4 plugins have an old structure

### Expected behavior

We should be doing all required modifications to the plugin that we can such as:

- Migrating plugins to the new structure if required (moving to ./src and removing the no-longer required `strapi-admin.js` file)
- Have codemods look through all files within the directory tree for things we can migrate (es to ds, and config namespace)
- Warn the user on things we cannot migrate

### Screenshots

![image](https://github.com/user-attachments/assets/1bf0bf84-08bc-4346-8a1a-f13f7531c98d)

### Code snippets

```
"use strict";

module.exports = ({ strapi }) => ({
  index(ctx) {
    let data = strapi.entityService.find("admin::user");

    console.log(data);

    ctx.body = strapi
      .plugin("test-plugin")
      .service("myService")
      .getWelcomeMessage();
  },
});
```

### Additional context

Related to @PaulBratslavsky and my work with Mux on migrating their plugin

hanpaine:
@derrickmehaffy I'm not sure we are supporting the creation of new plugins on v4 so is this use case valid? Or is the context the same whether it's a new v4 plugin vs and existing one?

hanpaine:
@derrickmehaffy confirmed with Product that new v4 plugin creation will be supported until end of 2024 btw 🙏

derrickmehaffy:
[...]

PR #21033, opened 2024-08-20
Fix Upgrade Tool Targeted Files For Plugins

Convly:
### What does it do?

Refactors the handling of project files by differentiating between application and plugin projects.

It introduces new constants for allowable root paths and specific root files for both project types.

The `Project` class is updated to accept a configuration object, which includes paths specific to either applications or plugins.

`AppProject` and `PluginProject` constructors are updated to define their specific paths.

### Why is it needed?

The previous approach used generic constants that led to inconsistencies and (most importantly) errors when dealing with different types of projects (app/plugin).

### How to test it?

1. Create a sample application and plugin projects in their respective directories.
2. Run the upgrade tasks (e.g., `list-codemods`, `run-codemods`, `upgrade`) with the `debug` flag enabled to ensure project details, including paths, are correctly logged.
3. Verify that the correct paths are being used for scanning project files in both application and plugin contexts.
4. Perform various codemod and upgrade operations on both types of projects to ensure that the file handling works as expected.

### Related issue(s)/PR(s)

#20944

### DIFF
--- a/packages/utils/upgrade/src/modules/format/formats.ts
+++ b/packages/utils/upgrade/src/modules/format/formats.ts
@@ -19,7 +19,7 @@ export const codemodUID = (uid: string) => {
 };
 
 export const projectDetails = (project: AppProject | PluginProject) => {
-  return `Project: TYPE=${projectType(project.type)}, CWD=${path(project.cwd)}, PATHS=${project.paths.map(path)}`;
+  return `Project: TYPE=${projectType(project.type)}; CWD=${path(project.cwd)}; PATHS=${project.paths.map(path)}`;
 };
 
 export const projectType = (type: ProjectType) => chalk.cyan(type);
--- a/packages/utils/upgrade/src/modules/format/formats.ts
+++ b/packages/utils/upgrade/src/modules/format/formats.ts
@@ -3,7 +3,7 @@ import chalk from 'chalk';
 
 import { constants as timerConstants } from '../timer';
 
-import type { ProjectType } from '../project';
+import type { AppProject, PluginProject, ProjectType } from '../project';
 import type { Codemod } from '../codemod';
 import type { Version } from '../version';
 import type { Report } from '../report';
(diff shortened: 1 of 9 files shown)


packages/utils/upgrade/src/modules/format/formats.ts
```diff
@@ -18,6 +18,10 @@ export const codemodUID = (uid: string) => {
   return chalk.bold.cyan(uid);
 };
 
+export const projectDetails = (project: AppProject | PluginProject) => {
+  return `Project: TYPE=${projectType(project.type)}, CWD=${path(project.cwd)}, PATHS=${project.paths.map(path)}`;
```
innerdvations, line 22:
```suggestion
  return `Project Details
Type: ${projectType(project.type)}
Cwd: ${path(project.cwd)}
Paths: ${project.paths.map(path)}
`;
```
What do you think about making this newline delimited, since PATHS is also comma delimited?

innerdvations (commented):
Code LGTM, I'll test it tomorrow

packages/utils/upgrade/src/modules/format/formats.ts
Convly: I can maybe use a `;` instead? I was trying to avoid newlines so that one line of log = a particular context (easier to copy paste etc...)

innerdvations: Ok, that makes sense, let's go with that. With all the new info we log it's tricky to keep it from feeling cluttered.

Convly: Indeed. My goal here is to make sure running the upgrade with --debug displays enough information to fix potential issues quickly. The normal output won't be changed anyway

Convly, line 22:
```suggestion
  return `Project: TYPE=${projectType(project.type)}; CWD=${path(project.cwd)}; PATHS=${project.paths.map(path)}`;
```

jhoward1994 (approved):
Tested with a v5 plugin and `create-strapi-app@rc`, its working well for me 👍🏻

innerdvations:
Already merged, but wanted to add that I've also tested it and it works for me!

PR #21034, opened 2024-08-20
feat!: document service query param validation

innerdvations:
### What does it do?

- allows for custom validators to be used instead of only having the content API defaults
- adds validation to every document service method so that it will only accept valid parameters for `filters`, `fields`, `sort`, and `populate`

Specifically, it also adds new validations (inc. for content API) for:
- populate: unpopulatable fields
- populate: full validation for nested dynamic zones and components

Not included in this PR but should be coming at a later date are `status` `locale` and `pagination` validation

### Why is it needed?

To improve DX by throwing errors when invalid parameters are passed in to document service calls.

In Typescript those errors should already be caught in build, but this adds another layer of protection at runtime and will prevent any unexpected behaviour

### How to test it?

Try making document service calls with invalid query params (filters, fields, sort, populate) such as non-existant fields, sorting or filtering on morph relations, etc.

Full test coverage is included, and some additional content API tests have been added to ensure we aren't accidentally allowing anything we were previously validating

### Related issue(s)/PR(s)

DX-861


derrickmehaffy:
This looks awesome

Marc-Roig:
Are this same validations done in the content api? If so, does it make sense to remove thhem and only keep this ones?

innerdvations:
> Are this same validations done in the content api? If so, does it make sense to remove thhem and only keep this ones?

Do you mean to avoid doing them twice? From a performance standpoind, they're fast, so I don't worry about it there.

However 🤔 if we do that and move the validation to the very end of every doc service call, it would kind of allow middlewares to extend the content API or doc service parameters (although they would need to modify the typings as well) because they would have a chance to mutate the params before the validation happens.

But I think for now we should start as restrictive as possible, and if we want to ease up on it later we can. If we do the other way around it'll be a breaking change.

tests/api/core/strapi/api/sort/sort.test.api.ts
```diff
@@ -211,6 +211,17 @@ describe('Sort', () => {
     ]);
   });
 
+  test('Sort on invalid key returns ValidationError', async () => {
```
innerdvations, line 214: Note: this was a very false start on where to put the tests. It's a case we don't actually have a test for, but it actually belong in tests/api/core/strapi/api/validate/validate-query.test.api.js so I'll move it before opening the PR

packages/plugins/i18n/server/src/bootstrap.ts
```diff
@@ -37,7 +37,8 @@ const registerModelsHooks = () => {
     // Use the id and populate built from non localized fields to get the full
     // result
     let resultID;
-    if (Array.isArray(result?.entries)) {
```
innerdvations, line 40: See https://github.com/strapi/strapi/issues/21066

tests/api/core/strapi/api/validate/validate-query.test.api.js
```diff
@@ -444,12 +455,12 @@ describe('Core API - Validate', () => {
       });
 
       // TODO: Nested sort returns duplicate results. Add back those tests when the issue will be fixed
-      describe.skip('Relation', () => {
+      describe('Relation', () => {
         describe('Scalar', () => {
-          describe('Basic (no modifiers)', () => {
```
innerdvations, line 449: Here I'm just adding back a portion of the skipped tests; adding back all of them didn't work unfortunately, so I left the referenced TODO about a bug

tests/api/core/strapi/api/validate/validate-query.test.api.js
innerdvations, line 1: In this file I'm just adding checks for invalid attribute names because we didn't have them before. There are a lot more that could be added, but it's not totally relevant to this PR since it's all targeting the content API rather than document service API

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -389,7 +389,7 @@ const BulkLocalePublishAction: DocumentActionComponent = ({
     }
   }, [isDraftRelationsError, toggleNotification, formatAPIError]);
 
-  if (!schema?.options?.draftAndPublish ?? false) {
```
innerdvations, line 392: fixes a TS warning I was getting, `?? false` doesn't do anything because left-hand will always return

tests/api/core/strapi/document-service/validation/validation.test.api.ts
```diff
@@ -0,0 +1,242 @@
...
+        });
+      });
+
+      it('should not throw on password attribute', async () => {
+        await strapi.documents(ARTICLE_UID)[methodName]({
+          filters: {
+            password: 'Hello World',
+          },
+        });
+      });
```
jhoward1994, line 90: This is outside the scope of filter validation I think and maybe I'm missing something but here if I do      

it('should not throw on password attribute', async () => {
    const res = await strapi.documents(ARTICLE_UID)[methodName]({
        filters: {
        password: 'Hello World',
        },
    });

    console.log(res);
});

I get this for `findMany`. Should we be ablke to filter on passwords like this?

```
    [
      {
        id: 1,
        documentId: 'Article1',
        title: 'Article1-Draft-EN',
        password: '$2a$10$4EP.Y8crWYRpfk2rVmmoce8raDqGYVwMlcI//GDQdO6z06bO50igu',
        private: 'private',
        createdAt: '2024-08-27T10:56:51.245Z',
        updatedAt: '2024-08-27T10:56:51.245Z',
        publishedAt: null,
        locale: 'en'
      },
      {
        id: 2,
        documentId: 'Article2',
        title: 'Article2-Draft-EN',
        password: null,
        private: 'private',
        createdAt: '2024-08-27T10:56:51.249Z',
        updatedAt: '2024-08-27T10:56:51.249Z',
        publishedAt: null,
        locale: 'en'
      }
    ]
```

innerdvations: On the doc service, yes, since it's the programmatic part and it should allow things that the content API doesn't (passwords, private fields, etc), it's more like a helper for writing db queries.

If the content API allows passwords or private fields, that would be a problem (and should be double-checked here just to be sure I haven't somehow allowed it under some circumstances)

innerdvations: Also, even if someone uses the doc service to retrieve this data and then returns it directly to a controller, the controller _should_ be sanitizing the response to the user which will remove things like passwords based on the user permissions.

tests/api/core/strapi/document-service/validation/validation.test.api.ts
```diff
@@ -0,0 +1,242 @@
...
+    'findMany',
+    'findFirst',
+    'findOne',
+    'publish',
+    'delete',
+    'create',
+    'unpublish',
+    'clone',
+    'update',
+  ];
```
Marc-Roig, line 31: should we add discard here?

innerdvations: oops, I missed that one! Yep, I'll add it

Marc-Roig: also, we support sending those params to `.count()` , so maybe we could test that one too? 🤔

innerdvations: Done!

Marc-Roig (approved):
Code looks fine to me! Let me know if you need some QA

jhoward1994 (approved):

packages/core/utils/src/validate/validators.ts
```diff
@@ -21,184 +22,366 @@ interface Context {
...
   return traverseEntity(throwPassword, ctx, entity);
 };
 
-const defaultValidateFilters = curry((ctx: Context, filters: unknown) => {
-  // TODO: schema checks should check that it is a validate schema with yup
-  if (!ctx.schema) {
-    throw new Error('Missing schema in defaultValidateFilters');
+type AnyFunc = (...args: any[]) => any;
+
+// TODO: move this to a utility
```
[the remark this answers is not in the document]
innerdvations: Ahhh, I probably should, yes. I'll update that

Convly (approved):

NEDDL:
The problem with this feature, now I cannot use the deep populate plugin because it is stuck in the validation.
Screenshot 2024-09-06 at 13 30 10

Any suggestions for a work around?

innerdvations:
> The problem with this feature, now I cannot use the deep populate plugin because it is stuck in the validation. Screenshot 2024-09-06 at 13 30 10
> 
> Any suggestions for a work around?

Thanks for the report! 

We're looking into the best solution and will post an update as soon as possible.

innerdvations:
Ok, after conferring internally, here is what we recommend.

In Strapi 5, plugins should handle cases like this with document service middlewares. 

So in this specific case, Content API queries for  deep-populate could be sent in a parameter such as `?deeppopulate=5`

The deep populate plugin would have a document service middleware such as:

```
documents.use((ctx, next) => {
   ctx.params.populate = getDeepPopulateQuery(ctx.params.populatedeep);
   return next();
})
```

to fill populate with a valid populate query. I haven't looked at how deep-populate works internally, but I assume it is already doing something very similar and would need to update to handle it at the right layer.

NEDDL:
@innerdvations I am thinking it might take a while if they update the plugin at all any time soon. Any way, I can disable the validation for populate in the meantime?

NEDDL:
I figured i need to do it myself ASAP: https://www.npmjs.com/package/strapi-v5-plugin-populate-deep

PR #21048, opened 2024-08-22
fix(content-manager): configure the view edit button should be type button

### LINKED ISSUE #21006: Can't open the edition modal in the configure the view page

- Node.js version: 18.17.1
- Strapi version: 5.0.0-rc.9

### Screenshots
https://github.com/user-attachments/assets/685fabc1-1143-4766-bbdb-745690dbd09a

Feranchz:
### What does it do?

Buttons inside forms are by default "submit" type, for the edit button on each field of the configure the view page we don't want them to be "submit" type, instead we mark them as "button" type so they don't have a default behaviour. 

### How to test it?

1. Go to the Edit View of any entry
2. Go to the Configure the view page
3. Try to click the edit button of one field
4. It should open the modal and nothing else

### Related issue(s)/PR(s)

Closes #21006


simotae14 (approved):
good catch!! it looks good and it works! well done 👏🏽

Issue #21037, opened 2024-08-21
The document ID should have been the same for both translations


vinda2401:
### Required System information

Node.js version: 18+
NPM version: 18.6
Strapi version: 5.0.0-rc.8
Plugin version:
Database: mysql
Operating system: Ubuntu 22.04.4 LTS

### Describe the bug

After following migration methods from v4 to v5, we discovered that the translations were not properly linked.

According to our understanding, the document id for both versions should be the same, yet the migrated records have different document ids for each entry. 
Thus, the translations are not correctly connected.

### Steps to reproduce the behavior

Prerequisite: Should have data in strapi v4 which is multilingual. Run though the migration process and upgrade to strapi v5. 

1. Open a collection containing multilingual data.
2. Click to open a record in editing mode. Attempt to swap between locales using the top right dropdown. You will notice an error.
3. Note down the document ID for this record.
4. Return to the collection’s listing page.
5. Filter data for another locale.
6. Choose the translated version of the record you opened before. Click to open it.
7. Write down the document ID.

In this scenario, the document IDs are different, hence the translations are not linked.

### Expected behavior

Both entries should have the same document ID, and switching between translations should be possible via a dropdown menu.

derrickmehaffy:
I've confirmed this issue and already informed our engineering team via Slack ([Internal Thread for Engineers](https://strapihq.slack.com/archives/C02CB44NJMV/p1724238274195689))

derrickmehaffy:
Source of the issue (I think) is here, since it's not finding the related IDs it's creating a new document ID:

https://github.com/strapi/strapi/blob/a2eac9d89132daee95ef7f194ed97e3b88b3650c/packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts#L96-L116

derrickmehaffy:
I think the issue is simple in that the wrong table name is being used in the join it should be:

```js
identifiers.getJoinTableName(meta.tableName, `localization_links`),
```

![image](https://github.com/user-attachments/assets/edb303a2-886b-4469-a130-08368cfa2a56)

derrickmehaffy:
fixed by: https://github.com/strapi/strapi/pull/21050

PR #21050, opened 2024-08-22
fix: v5 migration does not link locales

Marc-Roig:
### What does it do?

If you had a content-type with i18n enabled and multiple locales in v4, when you use the upgrade tool to migrate to v5 the locales are no longer properly linked (different document ID)

image

We renamed some of our table names in v5, mostly to fit long content types /field names into the database. The migration to migrate to document ids expected the locale join table to be renamed to the new format, but it wasn't.

This PR adds the necessary changes so the locales share the same document id.

Ideally , we would have some automated tests.

### How to test it?

- Using a v4 project, create multiple locales of the same entry in a content type with i18n.
- Migrate to v5
- You should see a single document with multiple locales

fixes: https://github.com/strapi/strapi/issues/21037

### DIFF
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
@@ -1,10 +1,21 @@
+/**
+ * NOTE: This migration avoids using the `identifiers` utility.
+ * As the `5.0.0-01-convert-identifiers-long-than-max-length`
+ * migration does not convert the `localizations` join tables, as they are not
+ * tables that exist anymore in v5.
+ *
+ * Hence the `identifiers` utility would return the wrong join table name.
+ *
+ * Database join table name: `categories_localizations_links`
+ * Actual `identifiers` returned join table name: `categories_localizations_lnk`
+ */
 import { createId } from '@paralleldrive/cuid2';
+import { snakeCase } from 'lodash/fp';
 import type { Knex } from 'knex';
 
 import type { Migration } from '../common';
 import type { Database } from '../..';
 import type { Meta } from '../../metadata';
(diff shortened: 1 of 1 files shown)


packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
```diff
@@ -1,10 +1,21 @@
...
+ * NOTE: This migration avoids using the `identifiers` utility.
+ * As the `5.0.0-01-convert-identifiers-long-than-max-length`
+ * migration does not convert the `localizations` join tables, as they are not
+ * tables that exist anymore in v5.
+ *
+ * Hence the `identifiers` utility would return the wrong join table name.
+ *
+ * Database join table name: `categories_localizations_links`
+ * Actual `identifiers` returned join table name: `categories_localizations_lnk`
+ */
```
innerdvations, line 11:
```suggestion
/**
 * NOTE: This migration avoids using the `identifiers` utility.
 * As the `5.0.0-01-convert-identifiers-long-than-max-length`
 * migration does not convert the `localizations` join tables, as they are not
 * tables that exist anymore in v5 and are not in the db metadata.
 *
 * This migration therefore relies on the fact that those tables still exist, and
 * references them directly.
 *
 * Database join table name: `categories_localizations_links`
 * Actual `identifiers` returned join table name: `categories_localizations_lnk`
 * 
 */
```

innerdvations (commented):
Code LGTM! will test it out later today

innerdvations (approved):
Works for me!

derrickmehaffy:
exp: `0.0.0-experimental.5691f94a497e119f564a7f60ce0830bf4c875eb9`

derrickmehaffy (approved):
Tested the exp and works perfectly, nice job!

PR #21072, opened 2024-08-26
feat: sync unidirectional relations

Marc-Roig:
### What does it do?

Updates uni directional relations to target the right entries when overriding published or draft entries.

Based on the following [notion RFC](https://www.notion.so/strapi/Unidirectional-relations-on-publish-discard-29d0133709c646b088f377727c9c7ff1)

### DIFF
--- a/packages/core/core/src/services/document-service/repository.ts
+++ b/packages/core/core/src/services/document-service/repository.ts
@@ -14,7 +14,7 @@ import { getDeepPopulate } from './utils/populate';
 import { transformParamsToQuery } from './transform/query';
 import { transformParamsDocumentId } from './transform/id-transform';
 import { createEventManager } from './events';
-import { syncUnidirectionalRelations } from './utils/unidirectional-relations';
+import * as unidirectionalRelations from './utils/unidirectional-relations';
 
 export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
   const contentType = strapi.contentType(uid);
@@ -251,15 +251,19 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
       }),
     ]);
 
+    // Load any unidirectional relation targetting the old published entries
+    const relationsToSync = await unidirectionalRelations.load(uid, oldPublishedVersions);
+
+    // Delete old published versions
+    await async.map(oldPublishedVersions, (entry: any) => entries.delete(entry.id));
+
     // Transform draft entry data and create published versions
     const publishedEntries = await async.map(draftsToPublish, (draft: any) =>
       entries.publish(draft, queryParams)
     );
 
-    await syncUnidirectionalRelations(uid, oldPublishedVersions, publishedEntries);
-
-    // Delete old published versions
-    await async.map(oldPublishedVersions, (entry: any) => entries.delete(entry.id));
+    // Sync unidirectional relations with the new published entries
+    await unidirectionalRelations.sync(oldPublishedVersions, publishedEntries, relationsToSync);
 
     publishedEntries.forEach(emitEvent('entry.publish'));
 
@@ -312,15 +316,19 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
       }),
     ]);
 
+    // Load any unidirectional relation targeting the old drafts
+    const relationsToSync = await unidirectionalRelations.load(uid, oldDrafts);
+
+    // Delete old drafts
+    await async.map(oldDrafts, (entry: any) => entries.delete(entry.id));
+
     // Transform published entry data and create draft versions
     const draftEntries = await async.map(versionsToDraft, (entry: any) =>
       entries.discardDraft(entry, queryParams)
     );
 
-    await syncUnidirectionalRelations(uid, oldDrafts, draftEntries);
-
-    // Delete old drafts
-    await async.map(oldDrafts, (entry: any) => entries.delete(entry.id));
+    // Sync unidirectional relations with the new draft entries
+    await unidirectionalRelations.sync(oldDrafts, draftEntries, relationsToSync);
 
     draftEntries.forEach(emitEvent('entry.draft-discard'));
     return { documentId, entries: draftEntries };
--- a/packages/core/core/src/services/document-service/repository.ts
+++ b/packages/core/core/src/services/document-service/repository.ts
@@ -14,6 +14,7 @@ import { getDeepPopulate } from './utils/populate';
 import { transformParamsToQuery } from './transform/query';
 import { transformParamsDocumentId } from './transform/id-transform';
 import { createEventManager } from './events';
+import { syncUnidirectionalRelations } from './utils/unidirectional-relations';
 
 export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
   const contentType = strapi.contentType(uid);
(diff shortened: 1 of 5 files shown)


packages/core/core/src/services/document-service/repository.ts
```diff
@@ -304,18 +308,20 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
...
+        select: ['id', 'locale'],
       }),
     ]);
 
-    // Delete all drafts
-    await async.map(versionsToDelete, (entry: any) => entries.delete(entry.id));
-
     // Transform published entry data and create draft versions
-    const draftEntries = await async.map(versionsToDraft, (entry: any) =>
+    const draftEntries = await async.map(versionsToDraft, async (entry: any) =>
```
alexandrebodin, line 316: Just wondering if it's actually needed ?

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -0,0 +1,70 @@
...
+  oldEntries: { id: string; locale: string }[],
+  newEntries: { id: string; locale: string }[]
+) => {
+  const relationsToSync: RelationToSync[] = [];
+
+  // Iterate all components and content types
+  const contentTypes = Object.values(strapi.contentTypes) as Schema.ContentType[];
+  const components = Object.values(strapi.components) as Schema.Component[];
+
+  for (const contentType of [...contentTypes, ...components]) {
```
alexandrebodin, line 30: Maybe use another name than contentType

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -0,0 +1,70 @@
...
+    // If the content type has a relation to the current content type
+    for (const [name, attribute] of Object.entries(contentType.attributes) as any) {
+      if (attribute.type !== 'relation') continue;
```
alexandrebodin, line 33: Would it be safer to target the specific oneWay & manyWay relational types directly ?

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -250,15 +251,18 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (uid) => {
...
 
+    await syncUnidirectionalRelations(uid, oldPublishedVersions, publishedEntries);
+
+    // Delete old published versions
+    await async.map(oldPublishedVersions, (entry: any) => entries.delete(entry.id));
+
     publishedEntries.forEach(emitEvent('entry.publish'));
+
     return { documentId, entries: publishedEntries };
   }
```
alexandrebodin, line 268: More of a general thought remark, when deleting data we do not clear the unidirectional relations pointing to nothing anymore. Sth we need to think about as a "cleanup task done in the bg"

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (!oldEntry) return;
+
+      // Update all relations to the new entry
+      for (const { uid, attribute: name } of relationsToSync) {
+        const model = strapi.db.metadata.get(uid);
+        const attribute = model.attributes[name];
+        if (attribute?.type === 'relation') {
+          // @ts-expect-error - FIX
+          const joinTable = attribute.joinTable;
+          if (!joinTable) continue;
```
alexandrebodin, line 55: if there is a joinColumn we should also handle it I think

We def should look into adding some higher level db queries to make this simpler 🤔

alexandrebodin (commented):
The logic looks sane and well encapsulated. Still those low level queries I think we have to offer an api for instead but not an issue for this PR

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
Marc-Roig: I don't fully understand the question, isn't this bit only selecting the one way/many way relations ? (by checking if the attr has inversedBy & mappedBy)

packages/core/core/src/services/document-service/repository.ts
Marc-Roig: can we rely on the cascade delete we currently have?

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
Marc-Roig: good point, will implement this for joinColumns too! 

do we currently have any db api to more easily update joinTables atm? ideally I would prefer to have a batch update on relations but that would be too much to implement atm

packages/core/core/src/services/document-service/repository.ts
alexandrebodin: in some cases yes but not for all db or polymorphics)

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
alexandrebodin: we don't but you got the idea right for later ;)

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
Marc-Roig: do you know of other ways of filtering by one way relations? I was checking the available metadata on the attribute and we only store the following:
```
{
  type: "relation",
  relation: "oneToOne",
  target: "api::tag.tag",
}
```

alexandrebodin: Ok that's my bad, we call them oneWay & manyWay only inside the ctb 🤦

derrickmehaffy: Trust me I do that too 😂

Marc-Roig:
@alexandrebodin we still do not support publishing entries with join columns, it's something I can work next, what do you think if we merge this one first and I work on that on a follow PR?

alexandrebodin (approved):
Code LGTM. I didn't QA, let me know if you need help with that

Marc-Roig:
@alexandrebodin a bit less compact, but I have updated the code to load+update unidirectional relations instead of updating them in a single query.

The original approach made the unique validations fail, we were not deleting the old published versions before creating the new ones. Hence triggering those validation errors

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -0,0 +1,106 @@
...
+  oldEntries: { id: string; locale: string }[],
+  newEntries: { id: string; locale: string }[],
+  oldRelations: { joinTable: any; relations: any[] }[]
+) => {
+  /**
+   * Create a map of old entry ids to new entry ids
+   *
+   * Will be used to update the relation target ids
+   */
+  const newEntryByLocale = keyBy('locale', newEntries);
```
alexandrebodin, line 77: We shouldn't need to handle the case with no locale or locale being null but maybe worth some check 🤔

Marc-Roig: will make it a bit more robust to that & will set up a test, last time I checked the locale value was always null on that scenario

alexandrebodin (approved):
This looks good 👍

PR #21073, opened 2024-08-26
chore: update prettier to 3.3.3

innerdvations:
### What does it do?

- updates prettier to 3.3.3 to match v5/main
- runs  `yarn prettier:write`

### Why is it needed?

fix an issue with the github CI using the wrong version and reporting errors that can't be seen locally

### How to test it?

### Related issue(s)/PR(s)

DX-1566

`v5/main` branch PR: https://github.com/strapi/strapi/pull/21071


innerdvations:
This doesn't seem necessary, still has some issues, and could cause a lot more conflicts merging with v5 later, so I'm closing for now.

Issue #20748, opened 2024-07-15
I can’t uncheck files on Media Library Modal on CM


yanniskadiri:
https://www.notion.so/strapi/I-can-t-uncheck-files-on-Media-Library-Modal-on-CM-c56b5297620144a5bf77e1cddcd33f24?pvs=4

yanniskadiri:
Could not reproduce in rc.7

PR #21077, opened 2024-08-26
fix: uncheck files on Media Library

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/20748


Marc-Roig (approved):
Looks good!
The "selected files" tab looks odd though 
image

But will create another ticket for that one

Issue #20747, opened 2024-07-15
Undefined status for an entry in Releases (might be related to the permissions)


yanniskadiri:
https://www.notion.so/strapi/Undefined-status-for-an-entry-in-Releases-might-be-related-to-the-permissions-2c96454dccf748b9bfaf82e45b24fbcf?pvs=4

yanniskadiri:
Could not reproduce in rc.7

PR #21080, opened 2024-08-26
fix: release errors not parsing nested errors

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/20747


Feranchz (approved):
ty 🚀

PR #21065, opened 2024-08-23
chore: remove helper plugin migration guide from monorepo

jhoward1994:
### What does it do?

Removes HELPER_PLUGIN_MIGRATION_GUIDE.md from the monorepo

### Why is it needed?

It exists in the dev docs. Recently modified here https://github.com/strapi/documentation/pull/2182
I think it's easier to only maintain this information in one place

### Related issue(s)/PR(s)

DX-1546

PR #21084, opened 2024-08-27
chore: add e2e and typescript to skipped_tests workflow

innerdvations:
### What does it do?

adds typescript, e2e_ce, and e2e_ee to skipped_tests

### Why is it needed?

Allows `aggregate_test_result` to complete even when tests are skipped 

### How to test it?

CI should pass

### Related issue(s)/PR(s)

I think this is why https://github.com/strapi/strapi/pull/21065 is being blocked

### DIFF
--- a/.github/workflows/skipped_tests.yml
+++ b/.github/workflows/skipped_tests.yml
@@ -36,6 +36,13 @@ jobs:
     steps:
       - run: echo "Skipped"
 
+  typescript:
+    runs-on: ubuntu-latest
+    needs: [build]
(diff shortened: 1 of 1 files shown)


alexandrebodin:
don't we need to also add the aggregate_test_result job in this file ?

alexandrebodin:
Nope sorry it's not what I meant, I'm not sure the aggregate_test_result can access the skipped jobs results anyway. So I wonder if we need to duplicate it (and not skip it)

PR #21085, opened 2024-08-27
feat(core): loading local plugins admin from dist

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/strapi/src/node/core/plugins.ts
+++ b/packages/core/strapi/src/node/core/plugins.ts
@@ -157,8 +157,16 @@ const loadUserPluginsFile = async (root: string): Promise<UserPluginConfigFile> 
   return {};
 };
 
-const getMapOfPluginsWithAdmin = (plugins: Record<string, PluginMeta>) =>
-  Object.values(plugins)
+const getMapOfPluginsWithAdmin = (plugins: Record<string, PluginMeta>) => {
+  /**
+   * This variable stores the import paths for plugins.
+   * The keys are the module paths of the plugins, and the values are the paths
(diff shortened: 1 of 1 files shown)


jhoward1994:
I'm not 100% sure I like this solution. It complicates the logic in `packages/core/strapi/src/node/core/plugins.ts` & makes it harder to follow. I'd appreciate any ideas for improvements or perhaps we can discuss creating the `strapi-admin` file with the plugin SDK again?

derrickmehaffy:
Do we think that's needed? Idc personally but @joshuaellis seemed adamant that he didn't want it.

Every Strapi plugin in v4 had it, if we keep it that's one less breaking change that honestly just feels like semantics at this point.

If I had to vote I'd say let's keep it if it doesn't hurt anything and generate it by default.

innerdvations:
> I'm not 100% sure I like this solution. It complicates the logic in `packages/core/strapi/src/node/core/plugins.ts` & makes it harder to follow. I'd appreciate any ideas for improvements or perhaps we can discuss creating the `strapi-admin` file with the plugin SDK again?

I think in this case it's better to have this slightly more complicated approach to support both ways. The file isn't required, but like Derrick said if we can avoid another breaking change, that would be great, especially since we can't have a codemod delete a file and it would have to be a manual process.

innerdvations:
e2e tests are failing because the documentation plugin is not loading. If we see it failing on other PRs we know it's not your problem, but there's a chance this change has caused the plugin loading to break for the documentation plugin

innerdvations (approved):
LGTM!

Issue #20899, opened 2024-07-29
Impossible to use URL to add a new asset (media library)


nathan-pichon:
### Required System information

- Environment: development
- OS: darwin-arm64
- Strapi Version: 5.0.0-rc.3
- Node/Yarn Version: yarn/1.22.19 npm/? node/v20.15.0 darwin arm64
- Edition: Community
- Database: sqlite
- Is your project Javascript or Typescript: Typescript

### Describe the bug

The URL input "FROM URL" on the modal to add a new asset cannot be changed (not possible to write in it or copy/paste)

### Steps to reproduce the behavior

1. Go to Media Library
2. Click on Add new assets
3. Click on FROM URL
3. Try to input text in the URL field

### Expected behavior

I should be able to write or paste my urls in this input.

### Screenshots

![image](https://github.com/user-attachments/assets/510a1600-5c35-46c6-a8e4-6786984c4b0d)

alexandrebodin:
cc @oiorain

Issue #20749, opened 2024-07-15
App crash when removing all fields in Configure The View


yanniskadiri:
https://www.notion.so/strapi/App-crash-when-removing-all-fields-in-Configure-The-View-b253a6dd11c64d6d83bf1fc7302cbda1?pvs=4

yanniskadiri:
Confirmed in rc.7

alexandrebodin:
cc @oiorain

Issue #20591, opened 2024-06-21
[Releases] Going back from Content Manager to Content Releases doesn't work as expected


MarionLemaire:
### Required System information

Node.js version: 20.9.0
Strapi version: 5.0.0-beta.13
Operating system: Windows WSL and OS

### Describe the bug

If I click the “Go to Content Manager” button in a release, the go back button doesn't work as expected, it is not redirecting me to the release page.

### Steps to reproduce the behavior

1. Go to Releases
2. Create a new release
3. Click on [Open the Content Manager]
4. From CM page, click Back arrow button

### Expected behavior

Go back button should redirect me to my release page.

### Screenshots
https://app.dashcam.io/replay/667589bebcbf70007514e420?share=wjlj2PxoYlWfcizWJ7F9ww

yanniskadiri:
Confirmed

alexandrebodin:
cc @oiorain

PR #21088, opened 2024-08-27
fix: multiple v5 issues

alexandrebodin:
### What does it do?

- Fix https://github.com/strapi/strapi/issues/20749 
- Fix the edit button didn't work and the modal would not open
- Fix https://github.com/strapi/strapi/issues/20591
- Fix https://github.com/strapi/strapi/issues/20899


Marc-Roig (approved):
🔥

Issue #20584, opened 2024-06-21
CTB - Modal retains previous tab status


MarionLemaire:
### Required System information

Node.js version: 20.9.0
Strapi version: 5.0.0-beta.13
Operating system: Windows WSL and OS

### Describe the bug

when you were on an Advanced settings tab when validating a new field creation, then for the next field creation, you arrive directly on Advanced settings tab instead of Basic settings tab (shouldn't be the case)

### Steps to reproduce the behavior

1. Go to CTB
2. Create a new CT
3. Create your first field, go to Advanced settings tab, validate
4. Create your second field : you're directly on Advanced settings tab instead of Basic settings tab

### Expected behavior

For every field creation, user should first arrive on Basic settings tab.

### Screenshots

https://app.dashcam.io/replay/6673edb810c6b5007053a42d?share=sGQ8Fg1JoBHunuehwMh9Gw

PR #21101, opened 2024-08-28
fix: ctb configure view invalid permission check

alexandrebodin:
### What does it do?

- Fix ctb missing configure view button because of invalid permission check
- Fix https://github.com/strapi/strapi/issues/20584

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/FormModalNavigationProvider/FormModalNavigationProvider.tsx
+++ b/packages/core/content-type-builder/admin/src/components/FormModalNavigationProvider/FormModalNavigationProvider.tsx
@@ -41,6 +41,7 @@ export const FormModalNavigationProvider = ({ children }: FormModalNavigationPro
         modalType: 'customField',
         attributeType,
         customFieldUid,
+        activeTab: 'basic',
       };
     });
   };
@@ -58,6 +59,7 @@ export const FormModalNavigationProvider = ({ children }: FormModalNavigationPro
         step,
         attributeType,
         showBackLink: true,
+        activeTab: 'basic',
       };
     });
   };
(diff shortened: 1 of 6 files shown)


packages/core/content-type-builder/admin/src/components/FormModalNavigationProvider/FormModalNavigationProvider.tsx
```diff
@@ -178,6 +184,7 @@ export const FormModalNavigationProvider = ({ children }: FormModalNavigationPro
         forTarget,
         targetUid,
         modalType: 'chooseAttribute',
+        activeTab: 'basic',
       };
     });
   };
```
jhoward1994, line 190: I think it's only at this point `onNavigateToChooseAttributeModal` that we need
to ensure activeTab is reset to 'basic'?

jhoward1994:
CTB -> CM navigation fix works well for me 👍🏻

packages/core/content-type-builder/admin/src/components/FormModalNavigationProvider/FormModalNavigationProvider.tsx
alexandrebodin: I prefered to always reset it just in case (thinking about when you cancel or go back a step what happens) 
but happy to change if you think it's unnecessary

jhoward1994:
I was trying to comment through vscode but it wasn't working - sorry if this appears twice


Could we do this in FormModalNavigationProvider/constants.ts
```suggestion
    import type { ModalEventProps } from './FormModalNavigationProvider';

    const INITIAL_STATE_DATA: ModalEventProps = {
    actionType: undefined,
    attributeName: undefined,
    attributeType: undefined,
    categoryName: undefined,
    dynamicZoneTarget: undefined,
    forTarget: undefined,
    modalType: undefined,
    isOpen: false,
    showBackLink: false,
    kind: undefined,
    step: undefined,
    targetUid: undefined,
    customFieldUid: undefined,
    activeTab: 'basic',
    };

    export { INITIAL_STATE_DATA };
```

And change ModalEventProps in
components/FormModalNavigationProvider/FormModalNavigationProvider.tsx to

```
export type ModalEventProps = {
  attributeType?: string;
  customFieldUid?: string;
  dynamicZoneTarget?: string;
  forTarget?: string;
  targetUid?: string;
  attributeName?: string;
  step?: string | null;
  kind?: string;
  categoryName?: string;
  modalType?: string;
  actionType?: string;
  isOpen?: boolean;
  showBackLink?: boolean;
  activeTab?: string;
};

```

alexandrebodin:
Not sure this makes sense @jhoward1994, I think the `State` type should be updated and used for initial state but I would rather avoid doing that in a bug fix

PR #21107, opened 2024-08-29
chore(e2e): cover ctb tab state management

jhoward1994:
### What does it do?

Covers CTB modal state management with e2e test


jhoward1994:
@alexandrebodin do you think your fixes are worth covering in e2e?

alexandrebodin:
Sure why not !

PR #21110, opened 2024-08-29
Fix Document Service Typing Issues

### LINKED ISSUE #20735: v5: typing errors in various places

- Node.js version: 20.9.0
- NPM version:
- Strapi version: 5.0.0-rc.3
- Database: SQLite
- Operating system: macOS

### Describe the bug

There are typing errors in various places in the application. Among others:

- On an entity returned by the Document Service API, the populated fields (via the populate option) do not appear, causing a typing error.
- The parameters of the Document Service API's middlewares are also incorrectly typed, which necessitates the use of `ts-ignore` in multiple places.

1. Use the Document Service API.
2. Use the populate option.
3. Observe that the populated fields do not appear, causing a typing error.
4. Check the parameters of the Document Service API's middlewares and see the typing issues.

### Expected behavior

The populated fields should appear correctly when using the populate option, without causing typing errors. The parameters of the middlewares in the Document Service API's middlewares should be correctly typed, eliminating the need for `ts-ignore`.

### Screenshots
image

Convly:
### What does it do?

- Fix document service return types
- Add various type documentation

### Why is it needed?

- Currently, the document service is not dynamically reusing the methods’ parameters in their return types, resulting in wrong types.

### How to test it?

- Go to a typescript application with generated types
- Use the document service to make several queries (making use of properties like `fields` or `populate`)
- Example
```typescript
  // restaurant has draft & publish enabled, the publish method is available
  await strapi.documents('api::restaurant.restaurant').publish({
    documentId: 'iuhzf-hbzefyg-567vzef',
    locale: 'fr',
    filters: { name: { $startsWith: 'Restaurant' } },
  });

  const entries = await strapi
    .documents('api::restaurant.restaurant')
    .findMany({ fields: 'name,shortDescription', populate: 'categories' });

  for (const restaurant of entries) {
    // Only name, shortdescription and categories are select in the query. They should be the only ones available in the types
   // (+ documentId and id that are always there)
    const { name, shortDescription, categories } = restaurant;

    console.log(
      name,
      shortDescription,
      categories.map((cat) => cat.name)
    );
  }
```
- Make sure the results make sense and that the types are correctly inferred.

### Related issue(s)/PR(s)

Partially fix #20735


packages/core/types/src/modules/documents/service-instance.ts
```diff
@@ -1,76 +1,145 @@
...
+  ): Result.Create<TContentTypeUID, TParams>;
+
+  update<const TParams extends Params.Update<TContentTypeUID>>(
+    params: TParams
+  ): Result.Update<TContentTypeUID, TParams>;
+
+  count<const TParams extends Params.Count<TContentTypeUID>>(params: TParams): Result.Count;
+
+  /**
+   * @internal
```
alexandrebodin, line 36: Can you also activate the skipInternal during build ?

packages/core/types/src/modules/entity-service/params/data.ts
```diff
@@ -1,4 +1,72 @@
...
+ *
+ * If so, it returns a partial version of `Input<TSchemaUID>`, otherwise it returns the default `Input` type.
+ *
+ * @template TSchemaUID - Represents a unique identifier for a schema, either content-type or component, extending {@link UID.Schema}.
+ *
+ * @remark
+ * This type is particularly useful for ensuring backward compatibility within input data structures,
+ * enabling flexibility during schema transitions or extensions without breaking existing data contracts.
+ */
+export type PartialInput<TSchemaUID extends UID.Schema> = Utils.If<
```
alexandrebodin, line 68: Should this be moved directly to the document service types instead ?

alexandrebodin (approved):

packages/core/types/src/modules/entity-service/params/data.ts
Convly: Yeah it should have been there in the first place :man_facepalming:

packages/core/types/src/modules/documents/service-instance.ts
alexandrebodin: we pbly need to figure a way to use the src code instead of the built types locally. we can do that later 👍

PR #21111, opened 2024-08-29
fix: (v4) marketplace hide install button when incompatible

innerdvations:
### What does it do?

Hides the install button for incompatible versions

### Why is it needed?

v5 plugins are entering the marketplace, and this will be the initial approach to guiding users to the correct plugin version

more updates will be coming after v5 launch to improve the experience

### How to test it?

Only compatible plugins should show the install button.

You can set up a proxy to mock the data from market-api.strapi.io or verify with some existing plugins that are out of the current version range. Which there currently aren't any, but in a few hours the marketplace cache should pick up that we've upgraded the documentation plugin to 5.0.0.

### Related issue(s)/PR(s)

DX-1531

### DIFF
--- a/packages/core/admin/admin/src/pages/Marketplace/tests/MarketplacePlugins.test.tsx
+++ b/packages/core/admin/admin/src/pages/Marketplace/tests/MarketplacePlugins.test.tsx
@@ -101,13 +101,6 @@ describe('Marketplace page - plugins tab', () => {
     )!;
     const alreadyInstalledText = within(alreadyInstalledCard).queryByText(/installed/i);
     expect(alreadyInstalledText).toBeVisible();
-
-    // Plugin that's not installed
-    const notInstalledCard = getAllByTestId('npm-package-card').find((div) =>
-      div.innerHTML.includes('Comments')
-    )!;
-    const notInstalledText = within(notInstalledCard).queryByText(/copy install command/i);
-    expect(notInstalledText).toBeVisible();
   });
 
   it('shows plugins filters popover', async () => {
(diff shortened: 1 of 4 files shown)


jhoward1994:
Full stack unlocked! 👏🏻

Just want to check, does this mean that a v4 app wont show install buttons for any plugins that have been updated to support v5? Even if theres an earlier version of the plugin that would be compatible?

innerdvations:
> Full stack unlocked! 👏🏻
> 
> Just want to check, does this mean that a v4 app wont show install buttons for any plugins that have been updated to support v5? Even if theres an earlier version of the plugin that would be compatible?

Yeah, it was decided that it's better to just direct them to the plugin author's page in that case.

In the future, we may add support in the marketplace for multiple versions to be listed.

packages/core/admin/admin/src/pages/Marketplace/tests/MarketplacePlugins.test.tsx
```diff
@@ -101,13 +101,6 @@ describe('Marketplace page - plugins tab', () => {
     )!;
     const alreadyInstalledText = within(alreadyInstalledCard).queryByText(/installed/i);
     expect(alreadyInstalledText).toBeVisible();
-
-    // Plugin that's not installed
```
innerdvations, line 105: These should be (and are, afaict) tested on their own in other places.

remidej (commented):
You may want to remove the pre-existing compatibility check from the button itself: https://github.com/strapi/strapi/blob/de05ede882bfb336c3573322cde42e66db3f918b/packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx#L258

jhoward1994 (approved):

Bassel17 (approved):
LGTM

Convly (approved):
Works well

PR #21113, opened 2024-08-30
fix(content-releases): await release status update after createMany actions

Feranchz:
### What does it do?

On createAction function we run the updateReleaseStatus function without awaiting for it. This is on purpose because updateReleaseStatus function could be expensive to run, so we want to return an answer and let this run on the background.

For bulk add to release we create a transaction so we can catch a possible error when adding one entry to a release and rollback everything, so we call `createAction` for each entry user is adding, this create two problems:
* We are calling `updateReleaseStatus` for each entry added to the release, this is inefficient.
* Depending on how long takes `updateReleaseStatus`, and considering this is inside a transaction, we could end up running queries when transaction is ended, causing crash errors.

To fix all this, I add a new option to `createAction` so we can disable calling `updateReleaseStatus`, and then we can run it after all entries are added to the release.

### How to test it?

1. Go to any Content Type with draft & publish enabled
2. Bulk add entries to a release
3. It should work without errors on console

### DIFF
--- a/packages/core/content-releases/server/src/controllers/__tests__/release-action.test.ts
+++ b/packages/core/content-releases/server/src/controllers/__tests__/release-action.test.ts
@@ -5,6 +5,7 @@ const mockSanitizedQueryRead = jest.fn().mockResolvedValue({});
 const mockFindActions = jest.fn().mockResolvedValue({ results: [], pagination: {} });
 const mockSanitizeOutput = jest.fn((entry: { id: number; name: string }) => ({ id: entry.id }));
 const mockCreateAction = jest.fn();
+const mockUpdateReleaseStatus = jest.fn();
 
 jest.mock('../../utils', () => ({
   getService: jest.fn(() => ({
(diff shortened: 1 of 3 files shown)


Feranchz:
@Marc-Roig I was sure this bug was also on v5 but I wasn't able to reproduce it on v5 too 🤔 

Do you know if we change something in transactions on v5 maybe?

In any case this check is good to have it also in v5 for performance reasons

Marc-Roig:
we have not modified transactions in v5, so it might be luck or the order things are executed are different 🤔 

We really need to improve the transaction api to execute things in a different transaction,

For now we normally rely on this hack:
```ts
strapi.db.transaction(({onCommit}) => {
  onCommit(() => // your code )
  })
```
This will make sure your code is executed after the current transaction is finished

Feranchz:
> we have not modified transactions in v5, so it might be luck or the order things are executed are different 🤔
> 
> We really need to improve the transaction api to execute things in a different transaction,
> 
> For now we normally rely on this hack:
> 
> ```ts
> strapi.db.transaction(({onCommit}) => {
>   onCommit(() => // your code )
>   })
> ```
> 
> This will make sure your code is executed after the current transaction is finished

@Marc-Roig Got it, but I think for this case is still better to run the `updateReleaseStatus` after the transaction, so I can check if added entries are greater than 0 to call the function

Marc-Roig (approved):

PR #21120, opened 2024-08-30
fix: return only drafts on relations from no DP to DP

Marc-Roig:
### What does it do?

It was noticed during the development of https://github.com/strapi/strapi/pull/21010 (specifically [here](https://github.com/strapi/strapi/pull/21010#discussion_r1716882293)) That a relation from a Content Type with disabled DP to another one with enabled DP can show both the relation targetting the draft and the published version, see:

image

This is at least confusing, and we should only show one relation.

We have debated multiple solutions, and atm product and engineering proposals are contradictory and we need to get to an agreement.

Product:
- From the content type without DP, show only the published version of the relation, and fallback to the draft

Engineering:
- From the content type without DP, show only the draft relation. 
That way, modifying the relation from that side does not interfere with live data from the other side.

This PR implements the engineering proposal, which is just a one liner. And can serve to start a discussion with @yanniskadiri

### DIFF
--- a/packages/core/content-manager/server/src/controllers/relations.ts
+++ b/packages/core/content-manager/server/src/controllers/relations.ts
@@ -1,4 +1,4 @@
-import { prop, uniq, uniqBy, concat, flow } from 'lodash/fp';
+import { prop, uniq, uniqBy, concat, flow, isEmpty } from 'lodash/fp';
 
 import { isOperatorOfType, contentTypes, relations, errors } from '@strapi/utils';
 import type { Data, Modules, UID } from '@strapi/types';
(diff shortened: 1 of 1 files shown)


Marc-Roig:
we agreed with yannis that this will be our approach (only showing the draft version of the relation) and that probably we could work on the way to explain this to the users (a warning, or something else)

packages/core/content-manager/server/src/controllers/relations.ts
```diff
@@ -328,8 +328,9 @@ export default {
       }
 
       // Add the status and locale filters if they are provided
-      if (status) {
-        where[`${alias}.published_at`] = getPublishedAtClause(status, targetUid);
+      const publishedAt = getPublishedAtClause(status, targetUid);
```
alexandrebodin, line 331: so published -> published / draft -> draft

`getPublishedAtClause` always returns an object so the condition will always be truthy is that a pbl ? Sounds a bit odd to allow that and let the db layer skip it 🤔 

```
where: {
  alias.published_at: {}
}
```

Marc-Roig: good point! will set it so status has to a filled object

alexandrebodin (approved):

Issue #21116, opened 2024-08-30
[Migration] Relations in Published entries are not correctly migrated


MarionLemaire:
### Required System information

- Node.js version: v20.9.0
- Strapi version: from 4.25.9 to 5.0.0-rc.12
- Operating system: Windows using WSL

### Describe the bug

WHen having a project in Strapi V4, migrating to V5 is erasing some existing relations for Published entries, displaying "No relations available" instead of the initially selected relation in the dropdown.

Checked with @Marc-Roig already.

### Steps to reproduce the behavior

1. In a V4 Strapi project, create 2 collection types with a many-to-many relations between them.
2. Create entries for both collection types, including relations between them.
3. Publish at least on of these entries.
4. Migrate to V5 using npx @strapi/upgrade@rc to 5.0.0-rc.12 -c 5.0.0
5. Check your data when migration is done : for published enties, relations that were saved and published are not saved anymore, instead "no relations available" is displayed in the dropdown instead of the initially selected relation.

### Expected behavior

Relations should remain the same for every entry when project is migrated to V5.

### Screenshots

https://app.dashcam.io/replay/66d197d5d5ed02ddb9e9501e?share=2f91bIA15A6fz6Dccu1Wsg

Published tab : 
<img width="605" alt="image" src="https://github.com/user-attachments/assets/e15983b7-e62b-4cca-8ca9-0fde80a82bc0">

Draft tab : 
<img width="545" alt="image" src="https://github.com/user-attachments/assets/7f49ede5-c71d-4101-bdb1-099c54fef15a">

alexandrebodin:
Fix PR was merged. Closing :)

alexandrebodin:
cc @Marc-Roig & @oiorain

PR #21121, opened 2024-08-30
fix: draft relations not being copied over when migrating to v5

Marc-Roig:
### What does it do?

V5 Migration  is erasing some existing relations on the draft versions, but they are kept on the published ones.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/21116

### DIFF
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -20,6 +20,26 @@ type DocumentVersion = { documentId: string; locale: string };
 type Knex = Parameters<Migration['up']>[0];
 
 /**
+ * Check if the model has draft and publish enabled.
+ */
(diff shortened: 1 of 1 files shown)


packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -62,13 +133,20 @@ const migrateUp = async (trx: Knex, db: Database) => {
       continue;
     }
 
+    /**
+     * Create plain draft entries for all the entries that were published.
+     */
+    await copyPublishedEntriesToDraft({ db, trx, uid: meta.uid });
+
```
alexandrebodin, line 140: you will need to do a 1st pass on all contenttypes to do the copyPublishedEntries 1st before running discardDraft on them as a 2nd pass :)

alexandrebodin (approved):
did not do testing so make sure to QA it but code lgtm

PR #21122, opened 2024-08-30
chore(upgrade): codemod for replacing nocontent with emptystatelayout

jhoward1994:
### What does it do?

Adds a codemod to change NonContent to EmptyStateLayout

### Why is it needed?

V5 plugin upgrade path

### How to test it?

- `npx create-strapi-app@rc v5-rc`
- `npx @strapi/sdk-plugin@latest init sdk-plugin` in the project
- cd PATH_TO_PLUGIN
- `PATH_TO_THIS_BRANCH/packages/utils/upgrade/bin/upgrade.js codemods run`

### Related issue(s)/PR(s)

DX-1532


Bassel17 (approved):
Looks great, I like the new utility
I think there are similar components that need changing like `nocontent`
once we get this through it should be easy to get more codemods done
looking at the list in the documentation here 
https://docs-next.strapi.io/dev-docs/migration/v4-to-v5/additional-resources/helper-plugin

Issue #17536, opened 2023-08-01
AWS Cognito redirect adding “https” twice in URL


khushvir:
### Required System information

- Node.js version: 18.13.0
- NPM version: 8.15.0
- Strapi version: 4.12.0
- Database: SQLite
- Operating system: Windows 11
- Is your project Javascript or Typescript: Javascript

### Describe the bug

I am using the below URL to log in to AWS Cognito.
http://localhost:1337/api/connect/cognito/

It’s redirecting me to my Cognito URL, but it’s adding “https” twice in the URL, e.g. "https://https//strapi.auth.ca-central-1.amazoncognito.com/oauth2/authorize?client_id=xxx"

### Expected behavior

It should redirect  to the AWS Cognito sign-in page without adding "https" twice.

### Screenshots

Redirect 
![aaa](https://github.com/strapi/strapi/assets/25701322/054c02ed-23f6-4573-9bb6-3fce0351f436)

Strapi Cognito config
![image](https://github.com/strapi/strapi/assets/25701322/956c4a50-18c3-4cfc-bd3e-93bffefa4e98)

### Code snippets

Server.js
module.exports = ({ env }) => ({
  host: env('HOST', '0.0.0.0'),
  port: env.int('PORT', 1337),
  url: env('', 'http://localhost:1337/'),
  app: {
    keys: env.array('APP_KEYS'),
  },
  webhooks: {
    populateRelations: env.bool('WEBHOOKS_POPULATE_RELATIONS', false),
  },
});

### Additional context

After removing the second "https" manually from the URL, it works.

Abhi1234-tech:
@derrickmehaffy, I've implemented the requested changes in this pull request. Please provide an update on the status. I've also closed the previous pull request due to complications

NmN03jain:
can you pls suggest in which file i have to make changes ?

Puneet-Amrutkar:
can you please assign this issue to me

ASCE-D:
Hey @derrickmehaffy I would like to contribute to this issue, please assign me.

derrickmehaffy:
> Hey @derrickmehaffy I would like to contribute to this issue, please assign me.

Please go read https://github.com/strapi/strapi/issues/18298

arg-adarsh:
Hey @derrickmehaffy, I would like to contribute to this issue. Can i work on this issue too?

Aditya8840:
Hey @derrickmehaffy I would like to work of this issue, please assign me

[...]

PR #21127, opened 2024-09-02
 fix: validation issues in the Users & Permissions plugin

sm0483:
### What does it do?

- Fixed issue #17536 related to validation issues in the provider schema
- Added unit tests for the email template schema and provider schema
- Fixed error message for the advanced settings

### Why is it needed?

- Fixes issues with validation in the provider, which was causing the subdomain and URL fields to be filled with invalid values
- Removes displaying the regex directly on the UI

### How to test it?

- Log in to Strapi
- Navigate to the Settings > Users & Permissions plugin
- Test the new fixes by filling out the fields in the advanced settings and providers sections

### Related issue(s)/PR(s)

- Issue #17536


Bassel17:
Hello, thanks for the PR, it looks good to me
I would like it if your re-target the PR to the develop branch (v5 version)

Thank you for the contribution

Issue #20948, opened 2024-08-06
Relation to self loses data of the relation when migrating from v4 to v5


Bassel17:
### Required System information

- Node.js version: 20.13.1
- NPM version: 10.5.2
- Strapi version: latest V5 (v5/main)
- Database: sqlite
- Operating system: Macos
- Is your project Javascript or Typescript: JS

### Describe the bug

- we lose relation data when we have a contentType with a circular relation to itself, when we migrate from v4 to v5

### Steps to reproduce the behavior

1- On v4 strapi project (or develop branch) add a Many to many relation to contenttype Country to itself
2- populate it with some test data
3- Upgrade to V5 (or switch to v5/main) run the project (migrations will run)
4- if you check the countries contentType the relations are lost

### Expected behavior

- relations should be migrated successfully

### Screenshots

- On V4
<img width="2029" alt="image" src="https://github.com/user-attachments/assets/bfb90845-cd81-47e7-bc6b-81437df0c888">

- After upgrading to V5
<img width="2235" alt="image" src="https://github.com/user-attachments/assets/7ca3ccd3-fd23-4f28-a1c0-12ca1ad4a028">

innerdvations:
After some investigation, we've found this is not self-relations, it is a result of two bugs.

1) The CTB in v4 generates manyToMany self-relations with an `inversedBy` on both attributes, rather than creating one with an `inversedBy` and one with a `mappedBy`  
1) In v4, Strapi generates two tables for the case where two sides of a relation have `inversedBy` rather than both `inversedBy` and `mappedBy` but only uses the first one listed in the schema
1) In v5, Strapi only keeps the last one listed in the Schema, and drops the first one (which is where v4 was storing the relation data)

PR #21129, opened 2024-09-02
fix: reproduce v4 behavior for data compatibility

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/20948

- This only ensure the v4 behavior still exists to avoid any data loss. The app is already logging a warning for the user to change their config. 
- It also fixes the ctb behavior for new relations

- This will need another refactoring at a later time

### Related issue(s)/PR(s)

Implements DX-1512

#20948

### DIFF
--- a/tests/api/core/database/v4-self-ref-compat.test.api.js
+++ b/tests/api/core/database/v4-self-ref-compat.test.api.js
@@ -1,0 +1,50 @@
+'use strict';
+
+// Test an API with all the possible filed types and simple filterings (no deep filtering, no relations)
+const { createStrapiInstance } = require('api-tests/strapi');
+const { createTestBuilder } = require('api-tests/builder');
+
+const builder = createTestBuilder();
+let strapi;
+
+const testCT = {
+  displayName: 'test',
+  singularName: 'test',
+  pluralName: 'tests',
+  kind: 'collectionType',
+  attributes: {
+    children: {
+      type: 'relation',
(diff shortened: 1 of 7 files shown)


innerdvations:
Failing test just because it's trying to match metadata and doesn't recognize the new attribute

Marc-Roig (approved):
approach looks good to me 👍

innerdvations (commented):
Makes sense, and it worked for me.

I'm not sure what the best way would be, but it would be nice to have an explicit test for this. Maybe an API test with a "bad" schema, and check that the expected tables exist?

alexandrebodin:
@innerdvations I added a test :)

innerdvations (approved):

tests/api/core/database/v4-self-ref-compat.test.api.js
```diff
@@ -0,0 +1,50 @@
...
+      type: 'relation',
+      relation: 'manyToMany',
+      target: 'api::test.test',
+      inversedBy: 'parents',
+    },
+    parents: {
+      type: 'relation',
+      relation: 'manyToMany',
+      target: 'api::test.test',
+      inversedBy: 'children', // intentionnaly wrong to validate retro compatibility
```
innerdvations, line 26:
```suggestion
      inversedBy: 'children', // intentionally wrong to validate retro compatibility
```

PR #21130, opened 2024-09-02
fix: (v5) marketplace hide install button when incompatible

innerdvations:
### What does it do?

Hides the install button for incompatible versions

### Why is it needed?

v5 plugins are entering the marketplace, and this will be the initial approach to guiding users to the correct plugin version

more updates will be coming after v5 launch to improve the experience

### How to test it?

Only compatible plugins should show the install button.

You can set up a proxy to mock the data from market-api.strapi.io or verify with some existing plugins that are out of the current version range. Which there currently aren't any, but in a few hours the marketplace cache should pick up that we've upgraded the documentation plugin to 5.0.0.

However, it's also being tested in front-end tests

### Related issue(s)/PR(s)

DX-1586

[v4 version](https://github.com/strapi/strapi/pull/21111)

### DIFF
--- a/packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx
+++ b/packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx
@@ -65,6 +65,9 @@ const NpmPackageCard = ({
     attributes.slug
   }`;
 
+  const versionRange = semver.validRange(attributes.strapiVersion);
+  const isCompatible = semver.satisfies(strapiAppVersion ?? '', versionRange ?? '');
+
   return (
     <Flex
       direction="column"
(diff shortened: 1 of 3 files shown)


reslear:
also maybe need implement in store https://feedback.strapi.io/feature-requests/p/marketplace-category-selection-with-v4-v5-plugin-support

innerdvations:
> also maybe need implement in store [feedback.strapi.io/feature-requests/p/marketplace-category-selection-with-v4-v5-plugin-support](https://feedback.strapi.io/feature-requests/p/marketplace-category-selection-with-v4-v5-plugin-support)

This feature is on the roadmap, coming soon :)

Bassel17 (approved):

reslear:
@innerdvations send pls link for track?

Convly (approved):
QAed

PR #21131, opened 2024-09-02
fix(admin): don't redirect if token is set

Feranchz:
### Current problem

When user mark "news" checkbox during registration they should be redirected to `/usecase` page before going to the admin panel. Right now they are being redirected to the admin panel instead.

This was happening because on `Register` component, on Register function (exactly [here](https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/pages/Auth/components/Register.tsx#L156)) we set the user's token. This cause `AuthPage` ([here](https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx)) to rerender because token has changed, that cause users to being redirected to the admin panel instead of `/usecase` because now token is not empty

### Solution

Solution is very simple, we really don't need to redirect users to admin if token is set at AuthPage level. I removed that check and instead I added it to the Login component.

### DIFF
--- a/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx
+++ b/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx
@@ -37,24 +37,17 @@ const AuthPage = () => {
 
   const { token } = useAuth('AuthPage', (auth) => auth);
 
-  const searchParams = new URLSearchParams(search);
-  const redirectTo = searchParams.get('redirectTo');
-
   if (!authType || !forms) {
     return <Navigate to="/" />;
   }
 
   const Component = forms[authType as keyof FormDictionary];
 
-  if (token && redirectTo) {
-    return <Navigate to={redirectTo} />;
-  }
-
   // Redirect the user to the login page if
   // the endpoint does not exist or
   // there is already an admin user oo
   // the user is already logged in
-  if (!Component || (hasAdmin && authType === 'register-admin') || (token && !redirectTo)) {
+  if (!Component || (hasAdmin && authType === 'register-admin' && token)) {
     return <Navigate to="/" />;
   }
 
--- a/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx
+++ b/packages/core/admin/admin/src/pages/Auth/AuthPage.tsx
@@ -37,17 +37,24 @@ const AuthPage = () => {
 
   const { token } = useAuth('AuthPage', (auth) => auth);
 
+  const searchParams = new URLSearchParams(search);
+  const redirectTo = searchParams.get('redirectTo');
+
   if (!authType || !forms) {
     return <Navigate to="/" />;
   }
 
   const Component = forms[authType as keyof FormDictionary];
 
+  if (token && redirectTo) {
+    return <Navigate to={redirectTo} />;
+  }
+
   // Redirect the user to the login page if
   // the endpoint does not exist or
   // there is already an admin user oo
   // the user is already logged in
-  if (!Component || (hasAdmin && authType === 'register-admin') || token) {
+  if (!Component || (hasAdmin && authType === 'register-admin') || (token && !redirectTo)) {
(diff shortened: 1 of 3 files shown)


alexandrebodin (commented):
A few comments/ideas 

- This approach will allow going and staying on the forgot-password & others when you are logged in. Is it ok ?
- Have you considered doing the redirection to /usecase in the AuthPage instead ?

Feranchz:
> * This approach will allow going and staying on the forgot-password & others when you are logged in. Is it ok ?

@alexandrebodin I don't see a problem. We could add the same check on those pages but that is what I don't like about this, we are doing an auth check in multiples pages instead of only one

> * Have you considered doing the redirection to /usecase in the AuthPage instead ?

The problem is that we end with some kind of race condition when an user is registered. We first set the token state and then navigate to `/usecase` if news is true, but at the same time we check the token on `AuthPage` and navigate to `/` if token is set. 
Screenshot 2024-09-03 at 10 32 41

(I was so confused yesterday that I made a little draw 😅 )

This morning I was thinking about adding a new state with the token when user is register, something like:
```
// ...
const res = await registerAdmin(body);

dispatch(login({
  token: res.data.token,
  isAuthFinish: !body.news
}))
```
Then I could use `isAuthFinish` on `AuthPage` to redirect or not to `/`, this could be also useful if in the future we decide to add some kind of second step to the register process

alexandrebodin:
> > * This approach will allow going and staying on the forgot-password & others when you are logged in. Is it ok ?
> 
> @alexandrebodin I don't see a problem. We could add the same check on those pages but that is what I don't like about this, we are doing an auth check in multiples pages instead of only one
> 
> > * Have you considered doing the redirection to /usecase in the AuthPage instead ?
> 
> The problem is that we end with some kind of race condition when an user is registered. We first set the token state and then navigate to `/usecase` if news is true, but at the same time we check the token on `AuthPage` and navigate to `/` if token is set. Screenshot 2024-09-03 at 10 32 41
> 
> (I was so confused yesterday that I made a little draw 😅 )
> 
> This morning I was thinking about adding a new state with the token when user is register, something like:
> 
> ```
> // ...
> const res = await registerAdmin(body);
> 
> dispatch(login({
>   token: res.data.token,
>   isAuthFinish: !body.news
> }))
> ```
> 
> Then I could use `isAuthFinish` on `AuthPage` to redirect or not to `/`, this could be also useful if in the future we decide to add some kind of second step to the register process

What about using the same logic as the redirectTo query param

login(token, { redirectTo: new ? '/usecase', '/' }) ? 

It is a bit more flexible

packages/core/admin/admin/src/pages/Auth/AuthPage.tsx
```diff
@@ -37,17 +37,24 @@ const AuthPage = () => {
...
 
+  if (token && redirectTo) {
+    return <Navigate to={redirectTo} />;
+  }
+
   // Redirect the user to the login page if
   // the endpoint does not exist or
   // there is already an admin user oo
   // the user is already logged in
-  if (!Component || (hasAdmin && authType === 'register-admin') || token) {
```
alexandrebodin, line 50: Wouldn't all of this be fixed if we changed that condition to 

`
if (!Component || (hasAdmin && authType === 'register-admin' && token)) 
`

otherwise we leave it to each page to redirect to where it should redirect at the end of the process 🤔

Feranchz: 🤯

alexandrebodin (approved):
LGTM. Can you ensure you do some hefty QA on all the login flows please 🙏  cc @oiorain

oiorain:
QA LGTM

Created a new user **with** the checkbox checked: 
- got redirected to a page where i was asked my role 
- then to the homepage 

Created a new user **without** the checkbox checked: 
- got redirected to the homepage

Issue #20430, opened 2024-06-04
[strapi v5 Beta] Custom controller cannot get body


iamobj:
### Required System information

- Node.js version: 20.11.1
- NPM version: pnpm/9.0.6
- Strapi version: 5.0.0-beta.8
- Database: sqlite
- Operating system: darwin-arm64
- Is your project Javascript or Typescript: Typescript

### Describe the bug

Custom controller, cannot get the body of `POST` request in the controller, I use `ctx.request.body`, ts prompt type "Request" does not have attribute "body", use `ctx.body`, ts does not report an error, but what I get is undefined

### Code snippets

```js
createOrder(ctx) {
      console.log("body", ctx.body); // body undefined
      return {};
}
```
if use `ctx.request.body`, Program error and terminated, error message `Property 'body' does not exist on type 'Request'`

iamobj:
I tested on v4, `ctx.body` also has no value, the correct usage is `ctx.request.body`, but v5 will type error: error message `Property 'body' does not exist on type 'Request'`, I currently use the following way to get it normally
```js
createOrder(ctx) {
      console.log("body", (ctx.request as any).body);
}
```

gregg-cbs:
Somewhat related forum:
https://forum.strapi.io/t/how-to-type-the-context-object-in-strapi-api-with-typescript/28935/6

Eventyret:
Confirmed in `5.0.0-beta.9`

```bash
curl --location 'http://localhost:1337/api/test/createOrder' \
--header 'Content-Type: application/json' \
--data '{
    "hello": "World"
}'
```

```typescript
import { factories } from "@strapi/strapi";
import { Context } from "koa";

export default factories.createCoreController(
  "api::test.test",
  ({ strapi }) => ({
    async createOrder(ctx: Context) {
      console.log("ctx", ctx.request); // has no body
      console.log("body", ctx.body); // body undefined
      return {};
    },
  })
);

```

```json
ctx {
  method: 'POST',
  url: '/api/test/createOrder',
  header: {
    'content-type': 'application/json',
    'user-agent': 'PostmanRuntime/7.39.0',
    accept: '*/*',
    'postman-token': '666c3c50-ee4b-4841-8e30-4e54b59ad3d2',
    host: 'localhost:1337',
    'accept-encoding': 'gzip, deflate, br',
    connection: 'keep-alive',
    'content-length': '24'
  }
}
body undefined
```

body object does not exist in ctx.

[...]

PR #21132, opened 2024-09-02
chore: import koa-body type to have request.body work

alexandrebodin:
### What does it do?

Tentative fix for https://github.com/strapi/strapi/issues/20430

### DIFF
--- a/packages/core/types/src/core/controller.ts
+++ b/packages/core/types/src/core/controller.ts
@@ -1,7 +1,7 @@
+import type {} from 'koa-body';
 import type { Context, Next } from 'koa';
 
 export type Controller = Record<string, ControllerHandler>;
-
 export interface ControllerHandler<TResponse = unknown> {
   (context: Context, next: Next): Promise<TResponse | void> | TResponse | void;
 }
--- a/packages/core/types/src/core/controller.ts
+++ b/packages/core/types/src/core/controller.ts
@@ -1,4 +1,4 @@
-import type {} from 'koa-body';
+import 'koa-body';
 import type { Context, Next } from 'koa';
 
 export type Controller = Record<string, ControllerHandler>;
--- a/packages/core/types/package.json
+++ b/packages/core/types/package.json
@@ -52,6 +52,7 @@
     "@strapi/utils": "workspace:*",
     "commander": "8.3.0",
     "koa": "2.15.2",
+    "koa-body": "4.2.0",
     "node-schedule": "2.1.1",
     "typedoc": "0.25.10",
     "typedoc-github-wiki-theme": "1.1.0",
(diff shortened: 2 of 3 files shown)


Convly (commented):
Were you able to reproduce the issue in the first place? 

I'm trying to test your fix, but I get the correct output with or without your change.

alexandrebodin:
Yes creating a new app with the latest RC I could reproduce but you have to be outside the monorepo

Convly:
Tested on v5-rc.12
image

Convly:
Published an experimental to test it: `0.0.0-experimental.9a23f06c0b3b13548dbd12feb3dc92d3c10bebb1`

packages/core/types/src/core/controller.ts
```diff
@@ -1,7 +1,7 @@
+import type {} from 'koa-body';
```
Convly, line 1: It doesn't work yet because empty import statements are stripped from the generated definition files.
Use the following instead to keep the import in the final build. (this does fix the issue)

```suggestion
import 'koa-body';
```

Convly (changes requested):

Convly (approved):

innerdvations (commented):
imports with side effects, my least favorite thing 😆 I know `koa-bodyparser` does the same thing, so this makes sense

LGTM, @Bassel17 is giving it some further QA

innerdvations (approved):

innerdvations:
There seem to be some build errors though

alexandrebodin:
I'll release another experimental I was using the wrong version

alexandrebodin:
here is the version `0.0.0-experimental.9c5778c9200404fcee94d5a0fc88da5e4376b82c`

alexandrebodin:
FYI, Looks like it doesn't work with this import either, We might need to create our custom ctx or sth like that

innerdvations (approved):

Bassel17:
`0.0.0-experimental.9c5778c9200404fcee94d5a0fc88da5e4376b82c` this version did not seem to fix the issue, still don't get a body

alexandrebodin:
> `0.0.0-experimental.9c5778c9200404fcee94d5a0fc88da5e4376b82c` this version did not seem to fix the issue, still don't get a body

Yes there was a pbl on main that prevents the experimental from being used correctly I think you are still using rc.11 actually. I'm doing a new one here https://github.com/strapi/strapi/actions/runs/10682310009/job/29607843337 you can try when it's done

alexandrebodin:
New exp: 0.0.0-experimental.65ad956ae83ac65d866b9bb3e8c022c34b85468f

Bassel17 (approved):
Looks good, it works

PR #21135, opened 2024-09-03
fix: connect and disconnect for morphtomany relations

### LINKED ISSUE #20961: Using `disconnect` on one file relation removes all file relations

- Node.js version: v18.17.1
- NPM version: 9.6.7
- Strapi version: 4.25.3
- Database: sqlite

### Describe the bug

Using `disconnect` in the body of a `PUT` request for an entity to disconnect a related file upload will remove all file relations.

If an upload ID is used that the entity isn't related to, then the request correctly doesn't make any changes. If an upload ID is used that the entity IS related to, then all upload relations are removed.

1. Have a content type, Listings, that has a media field, that allows multiple uploads and only image files
2. Upload two files when you create a Listing entity [like these docs state](https://docs.strapi.io/dev-docs/plugins/upload#upload-files-at-entry-creation). The file uploads have IDs 1 and 2.
3. Make a `PUT` request to update the entity, and try to disconnect file 1 like this:
```js
const response = await fetch(`my-strapi-url/api/listings/${id}`, {
  method: "PUT",
  body: JSON.stringify({
    data: {
      images: {
        disconnect: [{ id: 1 }],
      },
    },
}),
```
4. See that file 1 AND file 2 are not related to the Listing entity any longer

### Expected behavior

Given when is in the docs I would expect that the file with ID 1 would no longer be related to this Listing entity, and file 2 would still be related. I don't expect file 1 to have been deleted from the CMS or my external storage location, just no longer associated with this Listing.

I see the docs state that `connect` [can't be used to add media relations](https://docs.strapi.io/dev-docs/api/rest/relations#connect). It's unclear if that means I can't upload new files that way (not my goal here), but I wonder if that means we cannot manage media relations at all this way because of how the upload plugin works.

### Code snippets

### Additional context

My ultimate goal is to allow end users to remove images from the entities that they've created, via that entity's API endpoints. I need ownership enforcement on these entities and their related images, and it would be great to not have to open the `upload` endpoints directly to users and have to manage ownership issues there as well.

innerdvations:
### What does it do?

 - Adds correct support for `connect`, `disconnect`, and `set` with reordering for morphToMany relations (fixes #20961 )
- Also adds support in the sanitization and validation for polymorphic relations

**NOTE / WARNING**: I am not adding support at this time for the `morphMany` side of the relation, only on the `morphToMany` side. It's likely that `morphMany` has the same issue(s) that `morphToMany` had. We will add support for that in v5 at a later time

### Why is it needed?

They weren't working because they were deleting ALL relations even when connect and disconnect were being used

### How to test it?

Tests for media relations have been provided

### Related issue(s)/PR(s)

fixes #20961

DX-1545


packages/core/database/src/entity-manager/index.ts
```diff
@@ -834,16 +834,63 @@ export const createEntityManager = (db: Database): EntityManager => {
...
+                  .execute();
+              }
+              if (!isEmpty(cleanRelationData.connect)) {
+                const rows = cleanRelationData.connect?.map((data, idx) => ({
+                  [joinColumn.name]: data.id,
+                  [idColumn.name]: id,
+                  [typeColumn.name]: uid,
+                  ...(joinTable.on || {}),
+                  ...(data.__pivot || {}),
+                  order: idx + 1,
```
Convly, line 871: What happens if you already have relations with a certain order in the table and try to connect more? Shouldn't you start after the last one instead of 1?

innerdvations: Ahh, I think you're right. I'll check on that

packages/core/database/src/entity-manager/index.ts
```diff
@@ -834,16 +834,63 @@ export const createEntityManager = (db: Database): EntityManager => {
...
-                [typeColumn.name]: uid,
-                ...(joinTable.on || {}),
-                field: attributeName,
-              })
-              .transacting(trx)
-              .execute();
+            // for connect/disconnect, only modify those relations
+            if (!isEmpty(cleanRelationData.disconnect) || !isEmpty(cleanRelationData.connect)) {
+              // delete disonnects, and connects (to prevent duplicates when we add them later)
+              const idsToDelete = [
```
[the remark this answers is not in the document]
innerdvations: I didn't actually test it, but I think it would either throw a unique restraint error, or it would work and insert a new entry for the same relation

packages/core/database/src/entity-manager/index.ts
```diff
@@ -834,16 +834,63 @@ export const createEntityManager = (db: Database): EntityManager => {
...
-              .delete()
-              .where({
-                [idColumn.name]: id,
-                [typeColumn.name]: uid,
-                ...(joinTable.on || {}),
-                field: attributeName,
-              })
-              .transacting(trx)
-              .execute();
+            // for connect/disconnect, only modify those relations
```
alexandrebodin, line 837: I think you are missing the `set` usecase too

alexandrebodin (changes requested):
@innerdvations there are a few utils like `toAssocs` that could help uou in some places I believe but overall the logic seems fine

I think you need to look at how the ordering was managed in the other types of relations and the relations-ordered utils to support the reordering params (before, after, end, start) etc

packages/core/database/src/entity-manager/index.ts
innerdvations: For `set` they all need to be deleted and then recreated. I could actually check if `set` exists, and then skip this entire step, because I would consider that a `set` overrides any connect/disconnect that is used (and ideally we would throw an error if you tried to combine them, I don't know if that's the case or not)

innerdvations:
Yeah, I totally forgot about ordering here, I will add that

packages/core/database/src/entity-manager/index.ts
alexandrebodin: yes if set is defined the connect & disconnect shouldn't be. 

Right now the code will just use the data to set but not access the data inside the set key so it would break

Marc-Roig (commented):
other than what alex said of using existing utils , everything else looks good to me!

innerdvations:
So that it doesn't get forgotten, this will need the same fix as here: https://github.com/strapi/strapi/pull/22736

PR #21136, opened 2024-09-03
fix: list view populated data not managed  

alexandrebodin:
### What does it do?

Fixes https://github.com/strapi/strapi/issues/20604


remidej (approved):
I didn't think the fix would be in the frontend :grin:

alexandrebodin:
> I didn't think the fix would be in the frontend 😁

I think we need to refactor a bit more but it's enough for stable :)

oiorain:
Can I close #20604, @alexandrebodin ?

Issue #21098, opened 2024-08-28
[V5] Cannot choose default value for date field in Content-Type Builder


butcherZ:
### Required System information

- Node.js version: v20.17.0
- NPM version: 10.8.2
- Strapi version: v5.0.0-rc.11
- Database: sqlite
- Operating system: darwin-arm64
- Is your project Javascript or Typescript: TypeScript

### Steps to reproduce the behavior
https://www.loom.com/share/b42f7cb66f9b440b8c431bba37ebdba1?sid=9a92fcc4-b294-47a7-8ac2-700a3ad5e9f6
1. Go to Content-Type Builder
2. Click on any collection, add a date field.
3. Add a name, chose type as either `date` or `date time` (they both trigger the bug)
4. Go to Advanced Settings, try to select a default value.
5. See error.

### Expected behavior

 being able to select a default value with the format chosen from step 3.

### Screenshots
<img width="516" alt="Screenshot 2024-08-28 at 16 28 22" src="https://github.com/user-attachments/assets/75a6d5d1-6c7a-4525-a978-6efed9fda27c">

### Additional context
Might be related to #20875 but not exactly the same behaviour.
#

PR #21140, opened 2024-09-04
fix: fix default date value bug in content type builder(#21098)

butcherZ:
### What does it do?

Added a helper function `parseDateValue` to convert input types into valid Date objects or undefined for invalid inputs passed to DateTimePicker or DatePicker components.

### Why is it needed?

unexpected input types (eg. boolean) had been passed as value, causing errors in date handling components.

### How to test it?

- Go to content type builder
- Create a new date field in any collection
- Select type date or datetime
- Go to advanced setting tab, select a default value from date picker
- Tada! (it works!)

### Related issue(s)/PR(s)

DX-1588

#21098


packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
+import { Schema } from '@strapi/types';
+
+export const parseDateValue = <TAttribute extends Schema.Attribute.AnyAttribute>(
+  value: Schema.Attribute.Value<TAttribute>
+): Date | undefined => {
```
Bassel17, line 5: you don't really need to specify the return type here, typescript will infer it correctly

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (value instanceof Date) {
+    return isValidDate(value) ? value : undefined;
+  }
+
+  if (typeof value === 'string' || typeof value === 'number') {
+    const date = new Date(value);
+    return isValidDate(date) ? date : undefined;
+  }
+
+  return undefined;
```
Bassel17, line 15: no need for the `return undefined` this will be the default behavior of the method

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
+import { Schema } from '@strapi/types';
+
+export const parseDateValue = <TAttribute extends Schema.Attribute.AnyAttribute>(
+  value: Schema.Attribute.Value<TAttribute>
+): Date | undefined => {
+  if (value instanceof Date) {
```
Bassel17, line 6:
```
  if (value instanceof Date && isValidate(value)) {
```
you can do this here instead and then just return the value, looks a little neater not having to type `undefined`

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
...
+    return isValidDate(value) ? value : undefined;
+  }
+
+  if (typeof value === 'string' || typeof value === 'number') {
+    const date = new Date(value);
+    return isValidDate(date) ? date : undefined;
```
Bassel17, line 12: same idea here, you can do 
```
if ( isValidate(date)) {
    return date;
}
```
no need to explicitly return undefined this way

Bassel17 (changes requested):
looks great, just a few refactors you can do
Thanks for adding tests

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
+import { Schema } from '@strapi/types';
```
Convly, line 1:
```suggestion
import type { Schema } from '@strapi/types';
```

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const parseDateValue = <TAttribute extends Schema.Attribute.AnyAttribute>(
+  value: Schema.Attribute.Value<TAttribute>
```
Convly, line 4: To be honest, I think the input type can be simplified to `unknown` here since `Schema.Attribute.Value<TAttribute>` will resolve to `any` most of the time anyway (impossible intersection between too many different types).

The function signature would then look like
```ts
declare function parseDateValue(value: unknown): Date | undefined;
```

As for https://github.com/strapi/strapi/pull/21140#discussion_r1743862547, I would argue manually defining the return type in that kind of utils ensure we don't deviate from what is expected to be returned in the future, but that's a matter of preferences and it's not that important, whatever you think is best here.

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (value instanceof Date && isValidDate(value)) {
+    return value;
+  }
+
+  if (typeof value === 'string' || typeof value === 'number') {
+    const date = new Date(value);
+    if (isValidDate(date)) {
+      return date;
+    }
+  }
```
Convly, line 15: Couldn't this be simplified to the following?
```suggestion
  const date = new Date(value);
  
  return isValidDate(date) ? date : undefined;
```


Since the Date constructor don't throw with invalid data types (and instead returns an invalid date instance)

Convly: oops, didn't see Bassel comments, could also be

```typescript
  const date = new Date(value);
  
  if (isValidDate(date)) {
    return date;
  }
```

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
butcherZ: I'll update the input type then :)
And for the return type, if it were up to me, I'd prefer to declare it  manually just for better readability :D 
Do we have a norm for typescript usage?

Bassel17 (approved):
LGTM

packages/core/content-type-builder/admin/src/utils/parseDateValue.ts
Convly: We're in the ongoing process of writing them :upside_down_face:

Convly (approved):

PR #21141, opened 2024-09-04
enhancement: prevent committing tests with only

innerdvations:
### What does it do?

adds a pre-commit hook for staged changes that include `(test|it|describe).only`

### Why is it needed?

prevent us from accidentally committing a `.only()` in a test, which has happened several times in the past; we should only allow `.skip()` because then it is obvious something was intentionally skipped

### How to test it?

- after running `yarn prepare`
- try staging and committing a change with a test that includes .only left in it; you should be prevented
- having .only in your code that isn't staged should still be ok

### Related issue(s)/PR(s)

DX-1593

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,7 +2,7 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 # Check for .only in the staged changes of test files
-if git diff --cached | grep -E '^\+\s*(test|it|describe)\.only'; then
+if git --no-pager diff --cached | grep -E '^\+\s*(test|it|describe)\.only'; then
   echo "Error: .only found in staged changes. Please remove it before committing."
   exit 1
 fi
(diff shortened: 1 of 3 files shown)


Marc-Roig (approved):
💯

.husky/pre-commit
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+# Check for .only in the staged changes of test files
+if git diff --cached | grep -E '^\+\s*(test|it|describe)\.only'; then
```
alexandrebodin, line 5: I think the CI will be detected but just in case

```suggestion
if git --no-pager diff --cached | grep -E '^\+\s*(test|it|describe)\.only'; then
```

jhoward1994 (approved):

.husky/pre-commit
innerdvations: oh, good catch, thanks!

PR #21150, opened 2024-09-05
feat: Mongolian admin translations

ermuunn:
### What does it do?

This PR adds the Mongolian translation for Strapi's Admin, Upload (Content Manager) and Content-Type-Builder plugins.

### Why is it needed?

Web development has grown a lot in recent years in Mongolia, and many developers and web admins use Strapi admin dashboard. Adding Mongolian translation makes it easier for them to work in their own language, improving usability and accessibility.

### How to test it?

- Add `mn` into your app config.
- Select `Монгол` as your preferred language in the Profile settings or on the Login page.


alexandrebodin:
Hi @ermuunn we are not merging new language translations at the moment because we are unable to validate them correctly. We are currently rethinking the way we can offer this without having to make PRs to Strapi directly. 

Thank you anyway. We will close this PR for now but the content can be reused whenever we find a long term solution 🙏

PR #21152, opened 2024-09-05
fix: remove entity service decorator

Marc-Roig:
### What does it do?

Removes unused i18n entity service decorator


jhoward1994 (approved):
Thanks! looks good other than the linting error

alexandrebodin (approved):
ci is complaining but lgtm

PR #21155, opened 2024-09-05
fix: publishing correctly updates creator fields

Marc-Roig:
### What does it do?

Creator fields were not correctly copied when Publishing/Discarding, resulting in the `createdBy` and `updatedBy` fields being empty.

This change:
- makes it so those relations are copied
- allows using long hand format ( `{id: number}` ) to create/update join column relations


packages/core/database/src/entity-manager/index.ts
```diff
@@ -192,7 +192,7 @@ const processData = (
           : data[joinColumnName];
 
         if (!isUndefined(attrValue)) {
-          obj[joinColumnName] = attrValue;
+          obj[joinColumnName] = toId(attrValue);
```
Marc-Roig, line 195: this change allows connecting a join table using a long hand format, which was not supported before

innerdvations:
Linting issue, but the change lgtm! Testing it now

tests/api/core/strapi/document-service/publish.test.api.ts
```diff
@@ -24,6 +24,19 @@ describe('Document Service', () => {
   });
 
   describe('Publish', () => {
+    testInTransaction('Publishing keeps creator fields from the draft', async () => {
+      const articleDb = await findArticleDb({ title: 'Article1-Draft-EN' });
+
+      const publishedArticle = await publishArticle({
```
alexandrebodin, line 30: Just so we separate testing concerns could you do a publish without populate & then a find with populate. 

This way we "ensure" it doesn't work by side-effect of the manual populate

tests/api/core/strapi/document-service/publish.test.api.ts
```diff
@@ -24,6 +24,22 @@ describe('Document Service', () => {
   });
 
   describe('Publish', () => {
+    testInTransaction('Publishing keeps creator fields from the draft', async () => {
+      const articleDb = await findArticleDb({ title: 'Article1-Draft-EN' });
+
+      await publishArticle({ documentId: articleDb.documentId });
+
+      const publishedArticle = await findArticleDb({
```
alexandrebodin, line 32: do you have the populateCreatorFields set to true in the ct & don't you need to add them into the populate of the find ?

Marc-Roig: :O what ! didnt know that existed

Marc-Roig: ah 🤦 it was not being populated, but the validation was expecting the value to not be null, and it was undefined.

should be fixed now :)

packages/core/core/src/services/document-service/utils/populate.ts
```diff
@@ -22,12 +22,7 @@ export const getDeepPopulate = (uid: UID.Schema, opts: Options = {}) => {
           break;
         }
 
-        // TODO: Should this just be a plain list?
-        // Ignore createdBy, updatedBy, ...
-        const isVisible = contentTypes.isVisibleAttribute(model, attributeName);
```
alexandrebodin, line 27: So with the populateCreatorFields I think you can keep the isVisible, otherwise it's looses it's purpose 🤔

Marc-Roig: could you elaborate a bit more? 🤔 doesn't populateCreatorFields has to be set on the content type level? 
This bit ensures the creator fields are populated for all content types, regardless of that config

alexandrebodin: So populateCreatorFields defines if they are "allowed to be populated" what it does is set them to private or not but it doesn't actually populate them if you don't request them to avoid loading stuff when you do'nt want to see it. the name of the option could be better :D

Marc-Roig: It seems a bit odd to me, that to have the creator fields on the published data, you need to activate a "populateCreatorFields" in your content type. Shouldn't we always have that data available?

Maybe we could keep the visible attribute check, but always include the creator fields, wdyt?

alexandrebodin: I think there is a confusion. 

the populateCreatorFields just trigger private/not private on the fields. This is done to avoid leaking user information so devs have to opt-in to allow fetching creator fields.

You definitly can remove the isVisibleAttribute here as the purpose is to hide things in the admin ui not in the API response :)

Marc-Roig: Ah I see :) thank you! You were saying that in that scenario, the isVisible check lost its purpose 🤦‍♂️  mb

I gave it another thought, and there are specific fields we might not want to copy here, such as the review workflow assignee or stage. I might add those as excluded , without the isVisibleAttribute field.

alexandrebodin: true, you have 2 options

- Specificaly populate creator fields & exclude visible
- Specifically exclude some visible and select all 

Sounds like 1. is the safer approach 🤔

alexandrebodin (approved):
Logic looks good to me

PR #21156, opened 2024-09-05
chore: update quickstart behavior to disable example app

alexandrebodin:
### What does it do?

Changing the default behavior for the example app

cc @pwizla can you update the cli doc to say the example app will not be included with the quickstart ?


alexandrebodin:
Pre-release `0.0.0-experimental.f6c00790e260ea5a9b6b86abac5fea02b05d569c`

alexandrebodin:
> Not caused by this PR, but I just noticed that when I used --quickstart, we do not create a `.gitignore`, so node_modules are being included when we initialize git. Do you want to quickly fix that here, or should we create a separate PR for it?

I'll add the gitignore in the template here

innerdvations (approved):
LGTM! We just need to check it with an experimental tomorrow to be sure

alexandrebodin:
New experimental `0.0.0-experimental.626cf5324f21d318dee435c11cb3e08bb4c414b7`

PR #21158, opened 2024-09-05
fix: webhook button submitting and remove row clearing

alexandrebodin:
### What does it do?

Fix webhook header input submitting the form when removing a row and not allowing to clear the 1st row easily

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
@@ -36,7 +36,17 @@ const HeadersInput = () => {
 
   const addFieldRow = useForm('HeadersInput', (state) => state.addFieldRow);
   const removeFieldRow = useForm('HeadersInput', (state) => state.removeFieldRow);
+  const setFieldValue = useForm('HeadersInput', (state) => state.onChange);
   const { value = [] } = useField<Header[]>('headers');
+
+  const removeRow = (index: number) => {
+    // if we are removing the last row, simply clear it
+    if (value.length === 1) {
+      setFieldValue('headers', [{ key: '', value: '' }]);
+    } else {
+      removeFieldRow('headers', index);
+    }
+  };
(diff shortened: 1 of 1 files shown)


jhoward1994 (commented):
--

packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
```diff
@@ -81,8 +91,7 @@ const HeadersInput = () => {
                       padding="0.4rem"
                       alignItems="center"
                       justifyContent="center"
-                      disabled={value.length === 1}
-                      onClick={() => removeFieldRow('headers', index)}
+                      onClick={() => removeRow(index)}
```
jhoward1994, line 94: The styling is a bit strange for the minus button. If we replace this second <Grid.Item> with this it looks better

```
              <Grid.Item col={6} direction="column" alignItems="stretch">
                <Flex alignItems="flex-end" gap={2}>
                  <Box style={{ flex: 1 }}>
                    <StringInput
                      name={`headers.${index}.value`}
                      aria-label={`row ${index + 1} value`}
                      label={formatMessage({
                        id: 'Settings.webhooks.value',
                        defaultMessage: 'Value',
                      })}
                      type="string"
                    />
                  </Box>
                  <IconButton
                    width="3rem"
                    height="4rem"
                    onClick={() => removeRow(index)}
                    color="primary600"
                    label={formatMessage(
                      {
                        id: 'Settings.webhooks.headers.remove',
                        defaultMessage: 'Remove header row {number}',
                      },
                      { number: index + 1 }
                    )}
                    type="button"
                  >
                    <Minus width="0.8rem" />
                  </IconButton>
                </Flex>
              </Grid.Item>
```

alexandrebodin: I have a hard time know what change you are suggesting :D

jhoward1994 (approved):

PR #21160, opened 2024-09-06
chore(admin): improve webhook delete button style

jhoward1994:
### What does it do?

Improves the styling of the header delete button in the webhook edit view

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
@@ -83,28 +83,22 @@ const HeadersInput = () => {
                       type="string"
                     />
                   </Box>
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/pages/Settings/pages/Webhooks/components/HeadersInput.tsx
```diff
@@ -83,28 +83,22 @@ const HeadersInput = () => {
...
-                        },
-                        { number: index + 1 }
-                      )}
-                      type="button"
-                    >
-                      <Minus width="0.8rem" />
-                    </IconButton>
-                  </Flex>
+                  <IconButton
+                    width="3rem"
```
alexandrebodin, line 87: Just to make it a square, visually it feels strange to have more height than width ? 

```suggestion
                    width="4rem"
```

alexandrebodin (approved):

Issue #20915, opened 2024-07-31
Custom field input components don't receive the expected props


remidej:
Comes from a report and conversation with @shamglam on #20866.

In v5, custom field input components currently are not given the `value` and `onChange`.

Those could be retrieved using the useField hook. But in my opinion there's no need to introduce a breaking change here. We could call useField from the parent component, and pass value and onChange to the custom field's props.

This way we ensure backwards compatibility, and avoid breaking the existing custom field ecosystem in v5.

AndreiMrakov:
@remidej any updates or possible "hotfix" u did find? Coz i'm trying to update CKEditor and face with the same issue: input type "richtext" and almost half props is missing including `value` and `onChange`. But `useField` also doen't work coz returns empty object

brianbento:
@remidej Any idea when this will be fixed? Working on building my own custom fields and plugins and this is a major blocker.

Is there a workaround for now?

Thank you

brianbento:
@remidej Not a bug. This has been removed and replaced.

View code here as in #20866 
https://github.com/strapi/strapi/blob/baabf9edef4d0079bdbe0f891dcba743afc5a571/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx#L79

This should be added to the breaking changes documentation.

AndreiMrakov:
> @remidej Not a bug. This has been removed and replaced.
> 
> View code here as in #20866
> 
> https://github.com/strapi/strapi/blob/baabf9edef4d0079bdbe0f891dcba743afc5a571/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx#L79
> 
> This should be added to the breaking changes documentation.

@brianbento 
It returns empty object in this case, that’s the core problem

PR #21163, opened 2024-09-06
fix(content-manager): pass field to CustomField input to have backward compatibility

Feranchz:
### What does it do?

Pass field values to CustomField Input so we can have backward compatibility and don't introduce a breaking change.

### How to test it?

Use a Custom Field without using `useField`, instead use `value` and `onChange` that you should receive as param in your Input CustomField 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20915

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -4,6 +4,7 @@ import {
   useStrapiApp,
   useForm,
   InputRenderer as FormInputRenderer,
+  useField,
 } from '@strapi/admin/strapi-admin';
 import { useIntl } from 'react-intl';
 
@@ -63,6 +64,9 @@ const InputRenderer = ({ visible, hint: providedHint, ...props }: InputRendererP
     edit: { components },
   } = useDocLayout();
 
+  // We pass field in case of Custom Fields to keep backward compatibility
+  const field = useField(props.name);
+
   if (!visible) {
     return null;
   }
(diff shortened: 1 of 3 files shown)


remidej (commented):
Could you update the color picker custom field to use this? This way our example custom fields would match the docs, and we can test this PR more easily

Feranchz:
> Could you update the color picker custom field to use this? This way our example custom fields would match the docs, and we can test this PR more easily

@remidej Color Picker custom field is using `useField` currently, which I think is what we want users to do, and keep this as a backward, right? 🤔

remidej:
@Feranchz It's a question of point of view IMO: which is the recommended approach? both are fine, but I tend to prefer putting the props approach as the recommended one because:
- if you build a plugin like this, it works with v4 and v5, while using useField breaks v4
- it's the method the docs currently show, so no need to change them
- there's a lot of info that's available in the props and not the hook (name, required, hint, label, labelAction, etc.). So I like getting my data from 1 place instead of an arbitrary mix of props and useField

What do you think?

Feranchz:
> @Feranchz It's a question of point of view IMO: which is the recommended approach? both are fine, but I tend to prefer putting the props approach as the recommended one because:
> 
> * if you build a plugin like this, it works with v4 and v5, while using useField breaks v4
> * it's the method the docs currently show, so no need to change them
> * there's a lot of info that's available in the props and not the hook (name, required, hint, label, labelAction, etc.). So I like getting my data from 1 place instead of an arbitrary mix of props and useField
> 
> What do you think?

@remidej I agree with all you mentioned. So I change the current ColorPickerInput

remidej:
@Feranchz looks like there's a TS issue

remidej (approved):
🎯

PR #21167, opened 2024-09-06
chore: improve admin panel DX & add more advance plugin example

alexandrebodin:
### What does it do?

- Adds a more advance example plugin based on @remidej's todo plugin migrated to v5
- Exposed the PluginDefinition type and use it in the example plugin to show how to get typings & autocomplete in the admin api
- Exports `useDoc` to offer the needed equivalent to `useCMEditViewDataManager ` from v4 as `unstabled` for now & demo how to use it in the plugin
- Added back some of the properties from v4 in useDoc to ease the migrate experience as `deprecated`


examples/getstarted/src/plugins/todo-example/admin/src/hooks/useRelatedTasks.js
remidej, line 1: Since I understand the goal is to show plugin development best practices, maybe we could show how to use redux toolkit, instead of fetching in an effect?

packages/core/admin/admin/src/index.ts
```diff
@@ -76,3 +78,5 @@ export type { CMAdminConfiguration } from './types/adminConfiguration';
  */
 
 export { Layouts, type LayoutProps } from './components/Layouts/Layout';
+
+export type PluginDefinition = StrapiAppPlugin;
```
remidej, line 82: Love this change, maybe we can update the plugin boilerplates to use it? Both for TS and JS with the jsdoc just like you did

packages/core/content-manager/admin/src/exports.ts
```diff
@@ -6,7 +6,10 @@
 
 export { buildValidParams } from './utils/api';
 
-export { useDocument as unstable_useDocument } from './hooks/useDocument';
+export {
+  useDocument as unstable_useDocument,
+  useDoc as unstable_useDoc,
```
remidej, line 11: there is a comment where useDoc is defined that indicate it's for internal use only. we should probably update the comment if the plan has changed

remidej (commented):
I wonder why not use typescript?

examples/getstarted/src/plugins/todo-example/admin/src/hooks/useRelatedTasks.js
alexandrebodin: I migrated the example with the minimum of changes to make it comprable. doing more changes would confuse people. We should create a complexe plugin example but not in our day to day getstarted project

packages/core/admin/admin/src/index.ts
alexandrebodin: Yes next step is add a `definePlugin` function to make it even better but I want to separate some of the concerns :)

packages/core/content-manager/admin/src/exports.ts
alexandrebodin: I missed that comment. I'll do a mix of the 2 by export a wrapper as we definitely need to expose sth of that flavor regardless 👍

remidej: nice, maybe this is a chance to find a better hook name then 😁

alexandrebodin: any suggestion ? :D

remidej: tough one! something like `useEditViewData`? It's pretty close to the v4 hook it replaces, and makes it clear that it only works on the edit view page

innerdvations:
Works for me, but it now requires the plugin to be manually built before running getstarted or else it throws an error, which could be a big pain for everyone working on the monorepo (unless it gets added to the workspace and is built with `yarn setup`)

I'm wondering if we could have `strapi develop` in a project look for plugins and build them as well?

jhoward1994 (commented):
I'm wondering if we should we hold a plugin that needs building in the repo (examples/getstarted/src/plugins/todo-example)

It adds a step to running the example as the plugin needs to be built first

Also the watch script `"watch": "strapi-plugin watch" doesn't seem to be working correctly. If I watch the TODO plugin I get this when starting the example

```
││   Error: Could not load js config file                                                                                              ││   /Users/jamie/scr/strapi/code/strapi-2/examples/getstarted/src/plugins/todo-example/strapi-server.js: Cannot find module           ││   './content-types'                                                                                                                 

```

alexandrebodin:
> I'm wondering if we should we hold a plugin that needs building in the repo (examples/getstarted/src/plugins/todo-example)
> 
> It adds a step to running the example as the plugin needs to be built first
> 
> Also the watch script `"watch": "strapi-plugin watch" doesn't seem to be working correctly. If I watch the TODO plugin I get this when starting the example
> 
> ```
> │Error: Could not load js config  /Users/jamie/scr/strapi/code/strapi-2/examples/getstarted/src/plugins/todo example/strapi-server.js: Cannot find module  './content-types
> ```

If we can't have a pluign that will be built in our own repo we can't hope for user to be able to user them.

All of this shows the sdk-plugin is not working properly (watch mode in particular is messing up the server side)

I'll automate things and see if I move it in another folder too

alexandrebodin:
> I'm wondering if we could have `strapi develop` in a project look for plugins and build them as well?

We could but we force people to have a build command ? I think we have to clarify the path from local plugin to external plugin. I'll write up sth

maccomaccomaccomacco (approved):

packages/core/content-manager/admin/src/exports.ts
alexandrebodin: the pbl is it's completly missing the data 🤔

remidej: what is it returning then?

alexandrebodin: I updated this to include usefull context for plugin developers. let me know what you think :)

packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -196,5 +205,61 @@ const useDoc = () => {
...
 };
 
-export { useDocument, useDoc };
+/**
+ * @public
+ * @experimental
+ * Content manager context hooks for plugin development.
+ * Make sure to use this hook inside the content manager.
+ */
+const useContentManagerContext = () => {
```
remidej, line 214: Maybe we should remove the useDoc hook now?

- It feels weird to have 3 levels of abstraction (useContentManagerContext calls useDoc calls useDocument)
- The useDoc / useDocument naming is confusing
- Everything it returns is returned from this hook

So I think we could move its contents into this hook

alexandrebodin: Honestly I don't want to change any of the internal or refactor anything right now it's really not the goal. 
I Just want to expose a convenient hook for plugin developers. We can change all our internals afterwards for sure. I agree with your points :)

remidej: yes it doesn't need to be now, we can simplify it later

examples/plugins/todo-example/admin/src/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+      name,
+    });
+  },
+
+  bootstrap(app) {
+    app.getPlugin('content-manager').injectComponent('editView', 'right-links', {
+      name: 'my-plugin-my-compo',
+      Component: TodoCard,
+    });
+  },
```
remidej, line 20: CleanShot 2024-09-09 at 14 24 19@2x

The UI looks a bit weird. I think it would look much nicer using the `contentManagerPluginApis.addEditViewSidePanel` so that it gets its own card. Or if the goal is to stay as close as possible to v4, just removing the padding would be an improvement

alexandrebodin: UI update done

remidej (approved):

Issue #21177, opened 2024-09-06
[Releases] Single Types are not published/unpublished with i18n enabled


yanniskadiri:
https://www.notion.so/strapi/Releases-Single-Types-are-not-published-unpublished-with-i18n-enabled-17bb80ed851b47aab0ff3129e4c1b339?pvs=4

PR #21192, opened 2024-09-09
fix(content-releases): get entryDocumentId when entry is singleType

Feranchz:
### What does it do?

This PR fix multiple errors on Releases when you add a Single Type entry. SingleTypes have documentId but we don't have them on the frontend so we don't send it to the releases' endpoints. With this PR we append the documentId when content type is a Single Type. 

### How to test it?

Follow all the Releases flow (adding an entry, edit the entry, remove it, publish, etc...) with singleType entries.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21177

### DIFF
--- a/packages/core/content-releases/admin/src/components/ReleasesPanel.tsx
+++ b/packages/core/content-releases/admin/src/components/ReleasesPanel.tsx
@@ -11,7 +11,12 @@ import { ReleaseActionMenu } from './ReleaseActionMenu';
 
 import type { PanelComponent, PanelComponentProps } from '@strapi/content-manager/strapi-admin';
 
-const Panel: PanelComponent = ({ model, documentId, collectionType }: PanelComponentProps) => {
+const Panel: PanelComponent = ({
+  model,
+  document,
+  documentId,
+  collectionType,
+}: PanelComponentProps) => {
   const [{ query }] = useQueryParams<{ plugins: { i18n: { locale: string } } }>();
   const locale = query.plugins?.i18n?.locale;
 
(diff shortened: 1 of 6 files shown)


packages/core/content-releases/server/src/controllers/release-action.ts
```diff
@@ -24,6 +24,20 @@ const releaseActionController = {
...
 
+    // If we are adding a singleType, we need to append the documentId of that singleType
+    const model = strapi.getModel(releaseActionArgs.contentType);
+    if (model.kind === 'singleType') {
+      const document = await strapi.db.query(model.uid).findOne({ select: ['documentId'] });
+
+      if (!document) {
+        throw new errors.NotFoundError(
+          `No entry found for contentType ${releaseActionArgs.contentType}`
+        );
```
Marc-Roig, line 35: another option could be 
```
ctx.notFound(...)
```

Marc-Roig (approved):

remidej (approved):
It would be nice to have API tests for adding an entry to a release. But it can be post stable release

Feranchz:
thanks @Marc-Roig @remidej I also added an additional check so "Add to release" button is disabled when entry is creating

packages/core/content-releases/admin/src/components/ReleasesPanel.tsx
```diff
@@ -51,7 +61,7 @@ const Panel: PanelComponent = ({ model, documentId, collectionType }: PanelCompo
     return null;
   }
 
-  if (releases && releases.length === 0) {
+  if (!releases || (releases && releases.length === 0)) {
```
Marc-Roig, line 64: I think this might be the same:
```
if (!releases || releases.length === 0) {
```

If the first condition is met, the second is not computed, so the only way it can get there is by releases not being undefined

PR #21194, opened 2024-09-09
enhancement: resolve strapi-server with exports map too

alexandrebodin:
### What does it do?

Make `strapi-server` resolution use either "exports" key if define or look for the file at the root of the package to align with the `strapi-admin` loading. 

### Why is it needed?

The goal is to avoid the need for this file in the sdk-plugin generated plugin and remove the mix up


innerdvations (approved):
Nice, thanks for taking care of this!

Will you make the change in sdk-plugin too or should I create a task for it?

(Looks like there's a lint error though)

alexandrebodin:
I already have a PR for it in the sdk :) https://github.com/strapi/sdk-plugin/pull/50

alexandrebodin:
I'll QA with an experimental before merging

alexandrebodin:
Experimental `0.0.0-experimental.afa3b513b8f95459043f33fb94f4bac03af1474f`

innerdvations (approved):
Code lgtm!

Marc-Roig (approved):

PR #21195, opened 2024-09-09
fix: upgrade tool ignores node modules and dist

### LINKED ISSUE #21165: upgrade tool parses node_modules of plugins, too

- Node.js version: 20
- NPM version: 6
- Strapi version: 5-rc.15
- Database: sqlite
- Operating system: mac

### Describe the bug

When the upgrade tool is run and it parses a local plugin folder, it parses also its node_modules folder content.

1. Create a local plugin inside a project
2. Install its dependencies and so make a node_modules folder exist
3. run the upgrade tool against the whole project

you will get several errors stating files in the node_modules folders, giving evidence that the upgrade tool is parsing them, too

### Expected behavior

node_modules folders of local plugins shouldn’t be parsed

innerdvations:
### What does it do?

- switches from 'glob' to 'fast-glob' to support negations
- ignore node_modules and dist paths when running codemods

### Why is it needed?

The upgrade tool was running on node_modules in plugin paths

### How to test it?

codemods should never run on files within a node_modules or dist folder
they should run everywhere else as intended

### Related issue(s)/PR(s)

DX-1599

Fixes #21165


jhoward1994 (approved):

packages/utils/upgrade/src/modules/file-scanner/scanner.ts
```diff
@@ -11,7 +11,10 @@ export class FileScanner implements FileScannerInterface {
   }
 
   scan(patterns: string[]) {
-    const filenames = glob.sync(patterns, { cwd: this.cwd });
+    // we use fastglob instead of glob because it supports negation patterns
```
alexandrebodin, line 14: did you know you can pass an ignore param to glob directly ?

alexandrebodin (approved):
Makes sense

packages/utils/upgrade/src/modules/file-scanner/scanner.ts
innerdvations: Yeah, I wanted to keep it all in one place in `paths` so that we know what to expect

PR #21196, opened 2024-09-09
Adding codemods for helper plugin migrations

Bassel17:
### What does it do?

- Adds as many codemods as possible to automate all the changes we can automate that are presented here 
https://docs-next.strapi.io/dev-docs/migration/v4-to-v5/additional-resources/helper-plugin

### How to test it?

**This a list of all the changes that those codemods handle**

1. **Change `AnErrorOccurred` import:**
   - From: `@strapi/helper-plugin`
   - To: `Page` from `@strapi/strapi/admin`
   - Replace all uses of `AnErrorOccurred` with `Page.Error`

2. **Change `ConfirmDialog` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/strapi/admin`

3. **Change `DateTimePicker` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/design-system`

4. **Change `getFetchClient` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/strapi/admin`

5. **Change `LoadingIndicatorPage` import:**
   - From: `@strapi/helper-plugin`
   - To: `Page` from `@strapi/strapi/admin`
   - Replace all uses of `LoadingIndicatorPage` with `Page.Loading`

6. **Change `NoPermissions` import:**
   - From: `@strapi/helper-plugin`
   - To: `Page` from `@strapi/strapi/admin`
   - Replace all uses of `NoPermissions` with `Page.NoPermissions`

7. **Change `Status` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/design-system`

8. **Change `translatedErrors` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/strapi/admin`

9. **Change `useCallbackRef` import:**
   - From: `@strapi/helper-plugin`
   - To: `@strapi/design-system`

10. **Change `useCollator` import:**
    - From: `@strapi/helper-plugin`
    - To: `@strapi/design-system`

11. **Change `useFetchClient` import:**
    - From: `@strapi/helper-plugin`
    - To: `@strapi/strapi/admin`

12. **Change `useFilter` import:**
    - From: `@strapi/helper-plugin`
    - To: `@strapi/design-system`

13. **Change `useQueryParams` import:**
    - From: `@strapi/helper-plugin`
    - To: `@strapi/strapi/admin`

14. **Change `SearchURLQuery` import:**
    - From: `@strapi/helper-plugin`
    - To: `@strapi/strapi/admin`

### Related issue(s)/PR(s)


innerdvations (commented):
I wonder if we want to combine all of these into one deprecate-helper-plugin codemod so there aren't so many to scroll through.

@maccomaccomaccomacco what do you think?

maccomaccomaccomacco:
I have the feeling that it makes more sense to keep them separated to allow a more granular approach and to better handle eventual issues with the specific single codemods

@alexandrebodin what are your thoughts?

alexandrebodin:
Considering we are migrating a single package, running all it's codemods at once makes sense as you can't really run them granularly anyway. once you have remove the helper-plugin you need all the migrations to run. 

Would it be possible to offer both with an "alias" codemod that would run all the other ones ? (I guess worst case is there run twice ? )

innerdvations:
After discussing with @maccomaccomaccomacco , for now let's combine them all into one, because 
a) the helper plugin migration only makes sense to remove completely
b) this would double the amount of codemods we have, pushing the list (and results table) well into multiple pages of results
c) don't want to risk adding support for an alias (or folders?) at this point, although we may add something like that later to improve the granularity

innerdvations (commented):
I double-checked all of these against the migration guide and looks good! Just let me know when it's combined into one codemod

packages/utils/upgrade/resources/codemods/5.0.0/deprecate-helper-plugin.code.ts
```diff
@@ -0,0 +1,236 @@
...
+  changeImportSpecifier(root, j, {
+    oldMethodName: 'SearchURLQuery',
+    oldDependency: '@strapi/helper-plugin',
+    newDependency: '@strapi/strapi/admin',
+  });
+
+  return root.toSource();
+};
+
+export default transform;
```
jhoward1994, line 236: What do you think of using this kind of approach to remove duplicate code?
```suggestion
import type { Transform } from 'jscodeshift';
import { changeImportSpecifier } from '../../utils/change-import';
import { replaceJSXElement } from '../../utils/replace-jsx';

/**
 * This codemods automates all the imports and naming changes
 * for methods or components that used to be imported from '@strapi/helper-plugin'
 */
const transform: Transform = (file, api) => {
  const { j } = api;

  const root = j.withParser('tsx')(file.source);

  type Replacement = {
    oldName: string;
    oldDependency: string;
    toReplace: boolean;
    toChangeImportSpecifier: boolean;
    newDependency?: string;
    newName?: string;
  };

  const replacements: Replacement[] = [
    {
      oldName: 'AnErrorOccurred',
      newName: 'Page.Error',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toReplace: true,
      toChangeImportSpecifier: true,
    },
    {
      oldName: 'CheckPagePermissions',
      newName: 'Page.Protect',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toReplace: true,
      toChangeImportSpecifier: true,
    },
    {
      oldName: 'ConfirmDialog',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'DateTimePicker',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'getFetchClient',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'LoadingIndicatorPage',
      newName: 'Page.Loading',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toReplace: true,
      toChangeImportSpecifier: true,
    },
    {
      oldName: 'NoContent',
      newName: 'EmptyStateLayout',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toReplace: true,
      toChangeImportSpecifier: true,
    },
    {
      oldName: 'NoPermissions',
      newName: 'Page.NoPermissions',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toReplace: true,
      toChangeImportSpecifier: true,
    },
    {
      oldName: 'Status',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'translatedErrors',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useAPIErrorHandler',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useCallbackRef',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useCollator',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useFetchClient',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useFilter',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/design-system',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useQueryParams',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useRBAC',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'SearchURLQuery',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
    {
      oldName: 'useSettingsForm',
      oldDependency: '@strapi/helper-plugin',
      newDependency: '@strapi/strapi/admin',
      toChangeImportSpecifier: true,
      toReplace: false,
    },
  ];

  replacements.forEach((replacement) => {
    if (replacement.toReplace && replacement.newName) {
      replaceJSXElement(root, j, {
        oldElementName: replacement.oldName,
        newElementName: replacement.newName,
        oldDependency: replacement.oldDependency,
      });
    }

    if (replacement.toChangeImportSpecifier && replacement.newDependency) {
      changeImportSpecifier(root, j, {
        oldMethodName: replacement.oldName,
        newMethodName: replacement.newName,
        oldDependency: replacement.oldDependency,
        newDependency: replacement.newDependency,
      });
    }
  });

  return root.toSource();
};

export default transform;
```

packages/utils/upgrade/resources/codemods/5.0.0/deprecate-helper-plugin.code.ts
```diff
@@ -0,0 +1,236 @@
...
+    newElementName: 'Page.Loading',
+    oldDependency: '@strapi/helper-plugin',
+  });
+
+  changeImportSpecifier(root, j, {
+    oldMethodName: 'LoadingIndicatorPage',
+    newMethodName: 'Page',
+    oldDependency: '@strapi/helper-plugin',
+    newDependency: '@strapi/strapi/admin',
+  });
```
jhoward1994, line 96:
```suggestion
  changeImportSpecifier(root, j, {
    oldMethodName: 'LoadingIndicatorPage',
    newMethodName: 'Page.Loading',
    oldDependency: '@strapi/helper-plugin',
    newDependency: '@strapi/strapi/admin',
  });
```

??

innerdvations (approved):
LGTM, I did find some minor issues where ugly starting code will break. If you can fix them by tomorrow that would be great, but otherwise let's merge this so we can get it in before the code freeze because I don't think those issues are important enough to block this.

For example, if it had multiple statements for one import we add it to both which breaks:

```typescript
import { Button } from '@strapi/design-system';
import { useFilter } from '@strapi/helper-plugin';
import { useSomethingElse } from '@strapi/design-system';

// becomes

import { Button, useFilter } from '@strapi/design-system';
import { useSomethingElse, useFilter } from '@strapi/design-system';
```
and spits out a bunch of errors during the process

alexandrebodin:
Make sure to inform @pwizla  about the list of codemods once merged to update this section of the doc 🙏 
https://docs-next.strapi.io/dev-docs/migration/v4-to-v5/step-by-step#step-2-run-automated-migrations

innerdvations (approved):
LGTM, works great!

Issue #21157, opened 2024-09-05
Edit date and save in content manager causing infinite loop in browser


butcherZ:
### Required System information

Node.js version: v20.17.0
NPM version: 10.8.2
Strapi version: v5.0.0-rc.14
Database: sqlite
Operating system: MacOS 14.6.1 (23G93)
Is your project Javascript or Typescript: TypeScript

### Describe the bug

When creating a new entry in a collection with `date` or `datetime` type, editing the date will cause an infinite loop in the browser and can cause browser freeze.

### Steps to reproduce the behavior

1. Go to 'Content Manager'
2. Click on Any collection that has a date field with `date` or `datetime` type, create a new entry.
3. entry is now correctly created,
4.  right click to check the console in browser, try to update the date or time value
5. See error

### Expected behavior

document being updated without causing browser freeze

### Screenshots

<img width="390" alt="Screenshot 2024-09-05 at 14 34 28" src="https://github.com/user-attachments/assets/f6bbd59a-1dff-44b2-a08f-da83a77cb10b">

### Additional context
this warning is also shown in the terminal.
<img width="603" alt="Screenshot 2024-09-05 at 14 37 51" src="https://github.com/user-attachments/assets/c74f7106-9491-4c10-9c50-006a77f713ce">

However, even with the errors in the console, the updated value seemed to be saved correctly.
Also, field with only `time` type works perfectly fine here.

oiorain:
Hey @simotae14 , can i close this issue?

simotae14:
nope, the PR is still in review and we need to QA it carefully

PR #21197, opened 2024-09-09
fix: change the value format on change for date and datetime

simotae14:
### What does it do?

It solves the infinite loop in the devTool when we save a date or a date time on the CM Edit view

### Why is it needed?

Because RTK doesn't allow non serializable values so we want to serialize the value when we change it
In addiction I have fixed a small issue we had when we call the review workflow controller and we try to destructure stages even if the value is null

### How to test it?

1. Go to 'Content Manager'
2. Click on Any collection that has a date field with `date` or `datetime` type, create a new entry.
3. entry is now correctly created,
4.  right click to check the console in browser, try to update the date or time value
5. we don't have errors anymore

Please test carefully every single value with different locales and timezones, thanks

### Related issue(s)/PR(s)

CS-1088
https://github.com/strapi/strapi/issues/21157

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Date.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Date.tsx
@@ -25,7 +25,7 @@ const DateInput = forwardRef<HTMLInputElement, InputProps>(
           ref={composedRefs}
           clearLabel={formatMessage({ id: 'clearLabel', defaultMessage: 'Clear' })}
           onChange={(date) => {
-            field.onChange(name, date ? convertLocalDateToUTCDate(date) : null);
+            field.onChange(name, date ? convertLocalDateToUTCDate(date).toISOString() : null);
           }}
           onClear={() => field.onChange(name, null)}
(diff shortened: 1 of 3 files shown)


Feranchz (approved):

jhoward1994 (approved):

oiorain:
following the steps, I got this error when i set the date and datetime but im not sure it's related.
Screenshot 2024-09-10 at 10 49 00

butcherZ (approved):

butcherZ:
> following the steps, I got this error when i set the date and datetime but im not sure it's related. Screenshot 2024-09-10 at 10 49 00
> 
> The other error originally reported is not happening anymore

I performed QA on my side, and I don’t see the error shown in the screenshot in my console. 

Additionally, I can confirm that the original issue has been fixed :)

simotae14:
I am not able to reproduce your console error @oiorain so I am gonna merge this fix and if you want you can create another GH issue with the Syntax error

Issue #19960, opened 2024-03-28
When enabling i18n on an existing CT with existing published versions on by default locale, it creates a Draft in the new locale automatically 


MarionLemaire:
Version : 5.0.0-beta.0

Description

When enabling i18n on an existing CT with existing published versions on by default locale, it creates a Draft in the new locale automatically 

Steps To reproduce

- Create a Content type without i18N
- Create documents with this Content Type, save and publish them
- Go to settings, in Internationalization, add a new locale,
- Go to CTB, enable i18n for the Content Type above (Edit icon next to CT title > Advanced settings tab > check Internationalization checkbox), Finish
- Go to CM, go to an existing document for this CT in by default locale
- In the right upper corner, check the dropdown : there’s already a Draft version for the new locale you just created

Expected behaviours instead

Draft version label shouldn’t already exist

Discussed with @Marc Roig Campos : already exists in V4

Marc-Roig:
Checked, this is not the case in v4.
Also, I do think this is a cosmetic issue, but one we should fix regardless (if its not intended), empty locales are displayed as Draft in the locale dropdown 

(Second locale does not exist)
![image](https://github.com/strapi/strapi/assets/20578351/9f977a45-d228-4ac3-9550-b5453c0df4ce)

@lucasboilly was this intended?

yanniskadiri:
Lucas worked on some designs, let's go with [this option](https://www.figma.com/proto/Tt8uHuQezEP0CsWunbulb5/Draft-%26-Publish?page-id=3390%3A70176&node-id=6748-38808&viewport=544%2C224%2C0.1&t=7wnkIa6TjjvQNJCG-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6748%3A38808&show-proto-sidebar=1)

simotae14:
I am gonna close it because it was fixed by @Feranchz with this Pr https://github.com/strapi/strapi/pull/21199 👏🏽

PR #21199, opened 2024-09-09
fix: change status on locale picker

Feranchz:
### What does it do?

Change LocalePicker to match style and give user more information about creating an entry in a new locale.

Screenshot 2024-09-09 at 18 07 59

### How to test it?

(Try for both: content types with d&p) 

1. Enter a entry with i18n enabled.
2. Try to create, publish, etc, same entry in multiple locales and see results in the locale picker.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/19960


packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -50,6 +50,59 @@ import type { I18nBaseQuery } from '../types';
...
+  return (
+    <Flex width="100%" gap={1} justifyContent="space-between">
+      <Typography>{locale.name}</Typography>
+      {isDraftAndPublishEnabled ? (
+        <Status
+          display="flex"
+          paddingLeft="6px"
+          paddingRight="6px"
+          paddingTop="2px"
+          paddingBottom="2px"
```
remidej, line 90: is it not possible to use the design system values here (1, 2, 3 etc) instead of absolute values?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -50,6 +50,59 @@ import type { I18nBaseQuery } from '../types';
...
+          showBullet={false}
+          size={'S'}
+          variant={
+            status === 'draft' ? 'secondary' : status === 'published' ? 'success' : 'alternative'
```
remidej, line 94: I find nested ternaries quite hard to read. I think you could try either using a simple `getVariantColor` function, or using an object to map each status to a color and do `statusColors[status]`

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
Feranchz: Couldn't make it work without big changes on DS and I think we need to revisit how Select component from DS is working currently 🤔

Feranchz:
@remidej This version is not 100% exactly like Figma designs because the Header API we have now restricted me on some points 🤔 I think for now it's okay because is working and is clear the status of each entry but I would like to discuss about this Header API vs old Injection zones soon

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
remidej: let's not bother if it's complex 👍

remidej (commented):
CleanShot 2024-09-13 at 10 19 26@2x

code LGTM but i see an issue with the vertical alignment

Feranchz:
> CleanShot 2024-09-13 at 10 19 26@2x
> 
> code LGTM but i see an issue with the vertical alignment

@remidej Yes! It's a small issue on the DS, I will create a PR for that. Here I will solve a problem with e2e tests and go

remidej:
@Feranchz okay, we can merge after the tests are fixed and fix the alignment later then

Do you think you'll make the DS PR today? I'm planning to release it and bump it in the CMS this afternoon, before the code freeze

Feranchz:
> @Feranchz okay, we can merge after the tests are fixed and fix the alignment later then
> 
> Do you think you'll make the DS PR today? I'm planning to release it and bump it in the CMS this afternoon, before the code freeze

@remidej Sure! It's already done, this PR is ready also!

PR #21200, opened 2024-09-09
fix: single type fill from locale disabled

### LINKED ISSUE #21178: [CM] Fill in from another locale doesn’t work on ST
https://www.notion.so/strapi/CM-Fill-in-from-another-locale-doesn-t-work-on-ST-71e913ee1ea3437f903aabe3bfc59f1b?pvs=4

Bassel17:
### What does it do?

maps property to `documentId` instead of `id` so the document query is consistent for versions that are being created and the ones already created

similar issue fixed already in collection types in a different PR before, check https://github.com/strapi/strapi/pull/21009/files#r1726619948

### Why is it needed?

- fill from locale in single types is not working

### How to test it?

- fill from locale for singletypes should be working now and the button should not be disabled

### Related issue(s)/PR(s)

fixes #21178

### DIFF
--- a/packages/core/content-manager/server/src/controllers/single-types.ts
+++ b/packages/core/content-manager/server/src/controllers/single-types.ts
@@ -123,7 +123,7 @@ export default {
         permissionChecker,
         model,
         // @ts-expect-error - fix types
-        { id: document.documentId, locale, publishedAt: null },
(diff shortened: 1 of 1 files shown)


Marc-Roig (approved):

Feranchz (approved):
👍

PR #21204, opened 2024-09-10
fix: template vite config example

alexandrebodin:
### What does it do?

Updates vite config example template for ts to be actual ts

### DIFF
--- a/packages/cli/create-strapi-app/templates/example/src/admin/vite.config.example.ts
+++ b/packages/cli/create-strapi-app/templates/example/src/admin/vite.config.example.ts
@@ -1,6 +1,6 @@
-const { mergeConfig } = require('vite');
+import { mergeConfig, type UserConfig } from 'vite';
 
-module.exports = (config) => {
+export default (config: UserConfig) => {
   // Important: always return the modified config
   return mergeConfig(config, {
     resolve: {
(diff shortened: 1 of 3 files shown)


packages/cli/create-strapi-app/templates/example/src/admin/vite.config.example.ts
```diff
@@ -1,6 +1,6 @@
-const { mergeConfig } = require('vite');
+import { mergeConfig, UserConfig } from 'vite';
```
innerdvations, line 1:
```suggestion
import { mergeConfig, type UserConfig } from 'vite';
```

packages/cli/create-strapi-app/templates/vanilla/src/admin/vite.config.example.ts
```diff
@@ -1,6 +1,6 @@
-const { mergeConfig } = require('vite');
+import { mergeConfig, UserConfig } from 'vite';
```
innerdvations, line 1:
```suggestion
import { mergeConfig, type UserConfig } from 'vite';
```

templates/website/src/admin/vite.config.example.ts
```diff
@@ -1,6 +1,6 @@
-const { mergeConfig } = require('vite');
+import { mergeConfig, UserConfig } from 'vite';
```
innerdvations, line 1:
```suggestion
import { mergeConfig, type UserConfig } from 'vite';
```

innerdvations (approved):
LGTM!

Issue #20875, opened 2024-07-24
When trying to save a Date field with the default value it returns a error


EP-KHoeper:
### Required System information
- Node.js version: v20.5.0
- NPM version: 9.8.0
- Strapi version: 5.0.0-rc.6
- Database: sqlite
- Operating system: Ubuntu 22.04
- Is your project Javascript or Typescript: Typescript

### Steps to reproduce the behavior

Create a new Date field (type time) with a Default value.
Try to save without setting the time value.

[Screencast from 24.07.2024 16:09:46.webm](https://github.com/user-attachments/assets/e6a9d74a-6f6e-4ce6-ab1a-9e8b9ff2f56e)

### Expected behavior

When saving the date should be saved with its default value. 

### Screenshots

![image](https://github.com/user-attachments/assets/ae87d7e1-182a-474e-a4ac-e203f9dce681)

ymohit1603:
Assign me this issue i can fix this.

punitive1729:
@derrickmehaffy i have raised a PR for this issue to merge into v5/main since develop branch is still on v4 i hope its fine let me know if for v5 you are maintaining a separate base branch

butcherZ:
Hey there, this issue has been fixed, and @punitive1729 thank you for your contribution :)

PR #21206, opened 2024-09-10
fix: fix default time formatting issue

butcherZ:
### What does it do?

add `handleTimeChange` function to format time when needed.

### Why is it needed?

Input (Display) Format:

The`TimePicker` only supports the time format "HH:MM" (hours and minutes).
The default value set in the Content-Type Builder (CTB) comes in the format `HH:mm:ss.SSS`
To display the default value correctly, we need to convert it from `HH:mm:ss.SSS` to `HH:MM`.

Output (Submission) Format:

When a user selects or changes a time in the `TimePicker`, it provides the time in `HH:MM` format.
However, the backend expects to receive the time in the full `HH:mm:ss.SSS` format.
Therefore, we need to convert the user-selected time from `HH:MM` back to `HH:mm:ss.SSS` before sending it to the backend.

In summary:

For display: Convert `HH:mm:ss.SSS` → `HH:MM`
For submission: Convert `HH:MM` → `HH:mm:ss.SSS`

### How to test it?

Manually: 
- Create a `Date` field with type `time` in any collection in the Content type builder
- Go to advanced settings and select a `default value`
- Save
- Go to Content Manager, create a new entry with the default value
- There should be no errors

Unit test:

`cd strapi/packages/core/content-type-builder`
run `yarn test:front `

### Related issue(s)/PR(s)

#20875 

DX-1592

### DIFF
--- a/packages/core/content-type-builder/admin/src/utils/timeFormat.ts
+++ b/packages/core/content-type-builder/admin/src/utils/timeFormat.ts
@@ -3,14 +3,14 @@ type TimeChangeHandler = (params: {
 }) => void;
 
 type TimeChangeParams = {
-  value: string | undefined;
+  value?: string;
   onChange: TimeChangeHandler;
   name: string;
   type: string;
 };
 
-// The backend send a value which has the following format: '00:45:00.000'
-// or the time picker only supports hours & minutes so we need to mutate the value
+// The backend sends a value which has the following format: '00:45:00.000'
+// but the time picker only supports hours & minutes so we need to mutate the value
 const removeSeconds = (time: string): string => {
   const [hours, minutes] = time.split(':');
   return `${hours}:${minutes}`;
@@ -21,12 +21,12 @@ const addSecondsAndMilliseconds = (time: string): string => {
   return time.split(':').length === 2 ? `${time}:00.000` : time;
 };
 
-const formatTimeForInput = (value: string | undefined): string | undefined => {
-  if (!value) return undefined;
+const formatTimeForInput = (value?: string): string | undefined => {
+  if (!value) return;
   return value.split(':').length > 2 ? removeSeconds(value) : value;
 };
 
-const formatTimeForOutput = (value: string | undefined): string | undefined => {
+const formatTimeForOutput = (value?: string): string | undefined => {
   if (!value) return undefined;
   return addSecondsAndMilliseconds(value);
 };
@@ -38,10 +38,10 @@ export const handleTimeChange = ({ value }: TimeChangeParams): string | undefine
 };
 
 export const handleTimeChangeEvent = (
-  time: string | undefined,
   onChange: TimeChangeHandler,
   name: string,
-  type: string
+  type: string,
+  time?: string
 ): void => {
   const formattedOutputTime = formatTimeForOutput(time);
 
--- a/packages/core/content-type-builder/admin/src/utils/timeFormat.ts
+++ b/packages/core/content-type-builder/admin/src/utils/timeFormat.ts
@@ -1,0 +1,55 @@
+type TimeChangeHandler = (params: {
+  target: { name: string; value: string | undefined; type: string };
+}) => void;
+
+type TimeChangeParams = {
+  value: string | undefined;
+  onChange: TimeChangeHandler;
+  name: string;
+  type: string;
+};
+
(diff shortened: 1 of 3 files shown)


packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
...
+
+type TimeChangeParams = {
+  value: string | undefined;
+  onChange: TimeChangeHandler;
+  name: string;
+  type: string;
+};
+
+// The backend send a value which has the following format: '00:45:00.000'
+// or the time picker only supports hours & minutes so we need to mutate the value
```
innerdvations, line 13:
```suggestion
// The backend sends a value which has the following format: '00:45:00.000'
// but the time picker only supports hours & minutes so we need to mutate the value
```

innerdvations (approved):
Looks good to me, tested and works! Wait for a front-end review before merging though just in case I'm missing something :)

packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
+type TimeChangeHandler = (params: {
+  target: { name: string; value: string | undefined; type: string };
+}) => void;
+
+type TimeChangeParams = {
+  value: string | undefined;
```
Bassel17, line 6:
```suggestion
  value?: string;
```
you can do this instead just a little neater

packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
...
+  return `${hours}:${minutes}`;
+};
+
+// we need to send back the value with the same '00:45:00.000' format
+const addSecondsAndMilliseconds = (time: string): string => {
+  return time.split(':').length === 2 ? `${time}:00.000` : time;
+};
+
+const formatTimeForInput = (value: string | undefined): string | undefined => {
+  if (!value) return undefined;
```
Bassel17, line 25: you can choose to do this  you don't have to explicitly return undefined
```suggestion
  if (!value) return;
```
or just use 

```
if (value) {
  return value.split(':').length > 2 ? removeSeconds(value) : value;
}
```

packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
...
+  const [hours, minutes] = time.split(':');
+  return `${hours}:${minutes}`;
+};
+
+// we need to send back the value with the same '00:45:00.000' format
+const addSecondsAndMilliseconds = (time: string): string => {
+  return time.split(':').length === 2 ? `${time}:00.000` : time;
+};
+
+const formatTimeForInput = (value: string | undefined): string | undefined => {
```
Bassel17, line 24:
```suggestion
const formatTimeForInput = (value?: string): string | undefined => {
```

packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (!value) return undefined;
+  return value.split(':').length > 2 ? removeSeconds(value) : value;
+};
+
+const formatTimeForOutput = (value: string | undefined): string | undefined => {
```
Bassel17, line 29: Same ideas for this one as `formatTimeForInput`

packages/core/content-type-builder/admin/src/utils/timeFormat.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const handleTimeChange = ({ value }: TimeChangeParams): string | undefined => {
+  const formattedInputTime = formatTimeForInput(value);
+
+  return formattedInputTime;
+};
+
+export const handleTimeChangeEvent = (
+  time: string | undefined,
```
Bassel17, line 41:
```suggestion
  time?: string,
```

Bassel17 (changes requested):
Looks great! and seems to be working 
just left a few small comments

Bassel17 (approved):
LGTM!

there seems to be a failing e2e test suite, am betting it's a flaky one, you can attempt to re-run it, it will probably pass

PR #21211, opened 2024-09-10
chore: revert previous chante

Marc-Roig:
### What does it do?

Revert of https://github.com/strapi/strapi/pull/21050/files PR and add a different fix (snakecase on singularName)


innerdvations:
Closing this because the primary problem appears to be the new 'localizations' field.

I am a little worried about the migration fix we did a few weeks ago, because it is snakeCasing **everything**, and I seem to remember that tables created by i18n in v4 had a bug where they did not snakeCase the "i18n" part of the name, so we need to investigate that we're not losing any data that should be migrated because the migration is now looking for `i_18_n` instead of `i18n`

PR #21229, opened 2024-09-11
fix: add status to removed fields when cleaning data

### LINKED ISSUE #21179: [CM] Fill in from another locale doesn’t work properly on CT

Bassel17:
### What does it do?

Remove status field from added fields fields when filling from another locale

### Why is it needed?

look at issue #21179 
status field shouldn't be filled from another locale

### How to test it?

Try to fill from another locale (when the other locale is in `modified` status) and then save it should work with no errors
you can check #21179 for reproduction

### Related issue(s)/PR(s)

fixes #21179


alexandrebodin (approved):
Tested and it works well 👌 Found another i8n issue in the meantime I'll open an issue

PR #21230, opened 2024-09-11
fix: do not generate localizations links table

### LINKED ISSUE #21186: Migration core::5.0.0-discard-drafts (up) no such table: forms_form_localizations_lnk

- Node.js version: 18
- NPM version:
- Strapi version: 5 rc 17
- Database: sqlite
- Operating system: macos

### Describe the bug

The migration crashes on a select on a table that doesn't exist. I cannot find this code so I cannot debug it. I think it is somewhere in the `document().discardDraft()` code, which I cannot find.

1. Create a v4 without i18n
2. create a content type in a plugin
3. Migrate to v5
4. see MigrationError

### Expected behavior

no Error.

### Screenshots

Scherm­afbeelding 2024-09-07 om 13 09 49

### Additional context

https://github.com/strapi/strapi/issues/20225

https://github.com/strapi/strapi/issues/21185

### LINKED ISSUE #21189: strapi transfer Command Fails with Integer Syntax Error in *_localizations_lnk Tables After Upgrading to rc.17

- Environment: development
- OS: darwin-arm64
- Strapi Version: 5.0.0-rc.17
- Node/Yarn Version: yarn/1.22.22 npm/? node/v20.17.0 darwin arm64
- Edition: Community
- Database: sqlite

### Describe the bug

Running the `strapi transfer` command failed due to an attempt in insert data into *_localizations_lnk tables. The transfer is trying to insert a record using the `document_id` **string** where the table is expecting an **integer**, probably the `id` of that record.

I wasn't able to track down which release candidate introduced these tables, but this error started after upgrading from rc.13 to rc.17. I've rolled back to rc.13 for now to avoid the transfer error.

Error:
```error: [ERROR] insert into "i_18_n_locale_localizations_lnk" ("inv_locale_id", "locale_id", "locale_ord") values ($1, $2, $3) - invalid input syntax for type integer: "fzgtea87itb49x4u767jav50"```

1. run `strapi transfer` CLI command
2. See error

### Expected behavior

Transfer data without errors

### Additional context

- I am only using English
- All of my *_localizations_lnk tables are empty locally, so not sure why the transfer is trying to insert records into them

### PR DESCRIPTION
**Pull request by innerdvations**
### What does it do?

Prevents generating localizations link tables in the database, because they are a virtual relation

### Why is it needed?

Those tables shouldn't exist in Strapi 5 and it's currently wreaking havoc in a lot of areas of the code

### How to test it?

simple test: migrating from v4 to v5, migration should run fine and db should not include any `*localizations_links` or `*localizations_lnk` tables

long test: see the linked issues, this should solve all of them

### Related issue(s)/PR(s)

I believe it will fix all of the following, but I haven't confirmed yet:

fix #21189
fix #21186
another migration issue that was reported internally by @PaulBratslavsky

initial misdirected poc to fix some of the issues: https://github.com/strapi/strapi/pull/21211

DX-1604

### DIFF
--- a/packages/core/database/src/metadata/relations.ts
+++ b/packages/core/database/src/metadata/relations.ts
@@ -59,9 +59,6 @@ const isOwner = (
 ): attribute is RelationalAttribute & Relation.Owner =>
   !isBidirectional(attribute) || hasInversedBy(attribute);
 
-const isVirtual = (attribute: RelationalAttribute) =>
-  'virtual' in attribute && attribute.virtual === true;
-
 const shouldUseJoinTable = (attribute: RelationalAttribute) =>
   !('useJoinTable' in attribute) || attribute.useJoinTable !== false;
 
@@ -129,7 +126,7 @@ const createOneToMany = (
       attributeName,
       meta,
     });
-  } else if (isOwner(attribute) && !isVirtual(attribute)) {
+  } else if (isOwner(attribute)) {
     throw new Error('one side of a oneToMany cannot be the owner side in a bidirectional relation');
   }
 };
@@ -382,10 +379,6 @@ const createMorphMany = (
  * Creates a join column info and add them to the attribute meta
  */
 const createJoinColumn = (metadata: Metadata, { attribute, attributeName }: JoinColumnOptions) => {
-  if (isVirtual(attribute)) {
-    return;
-  }
-
   const targetMeta = metadata.get(attribute.target);
 
   if (!targetMeta) {
@@ -424,10 +417,6 @@ const createJoinTable = (
   metadata: Metadata,
   { attributeName, attribute, meta }: JoinTableOptions
 ) => {
-  if (isVirtual(attribute)) {
-    return;
-  }
-
   if (!shouldUseJoinTable(attribute)) {
     throw new Error('Attempted to create join table when useJoinTable is false');
   }
--- a/packages/core/database/src/metadata/relations.ts
+++ b/packages/core/database/src/metadata/relations.ts
@@ -419,6 +419,12 @@ const createJoinTable = (
 ) => {
   const targetMeta = metadata.get(attribute.target);
 
+  // Localizations are a "calculated" relation set in packages/plugins/i18n/server/src/register.ts
+  // we do not create an actual join table for localizations
+  if (attributeName === 'localizations') {
+    return;
+  }
+
   if (!targetMeta) {
     throw new Error(`Unknown target ${attribute.target}`);
   }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by alexandrebodin — APPROVED** [review]
logic looks good. Let's make sure it's not doing anything unexpected. Reading the code it should enter into a noop but to be verified

File: packages/core/database/src/metadata/relations.ts [hunk 1]
```diff
@@ -417,6 +425,10 @@ const createJoinTable = (
   metadata: Metadata,
   { attributeName, attribute, meta }: JoinTableOptions
 ) => {
+  if (!shouldUseJoinTable(attribute)) {
```
innerdvations, line 428: Just a note, every place in the codebase the calls createJoinTable has a check for this before it now. And I don't think we expose it. So this should help catch future mistakes.

packages/core/database/src/metadata/relations.ts
```diff
@@ -120,6 +120,10 @@ const createOneToMany = (
   meta: Meta,
   metadata: Metadata
 ) => {
+  if (!shouldUseJoinTable(attribute)) {
```
alexandrebodin, line 123: I would move that into the isBidirectional & throw if it's falsy as it's an error

packages/core/database/src/metadata/relations.ts
```diff
@@ -190,6 +194,10 @@ const createManyToMany = (
...
```
alexandrebodin, line 197: I think you can throw actually. ManyToMany can only work with joinTables

alexandrebodin (changes requested):

packages/core/database/src/metadata/relations.ts
innerdvations: Actually this is the one that is necessary to suppress for our localizations hack. The localizations attribute tries to create a oneToMany, but we've set joinTable to false, so this allows it to skip it. I'll leave them both as returns but comment it so we remember why it was done.

alexandrebodin: Would it make sense to add an option to disable database create instead in that case like `unstable__virtual: true`

innerdvations:
@alexandrebodin  I added a virtual field so that we have some additional context, but I'm starting to confuse myself on exactly how this should work. If you still see issues, maybe we can have a quick session together and figure out the best solution.

packages/plugins/i18n/server/src/register.ts
```diff
@@ -70,6 +70,8 @@ const extendContentTypes = (strapi: Core.Strapi) => {
       private: false,
       configurable: false,
       visible: false,
+      unstable_virtual: true,
+      useJoinColumn: false,
```
alexandrebodin, line 74: you can remove the useJoinColumn too now

innerdvations: oops, that was a last minute typo trying to add back useJoinTable specifically to denote that we wouldn't want to use a join table even if it weren't virtual. But I'll just remove it completely.

alexandrebodin (approved):
LGTM, good thing to add as unstable so nobody starts using it

Issue #20817, opened 2024-07-16
[Design] After clicking on Review Workflow, all icons are becoming black


MarionLemaire:
https://www.notion.so/strapi/After-clicking-on-Review-Workflow-all-icons-are-becoming-black-9b2725f727924ac3836b68cc1767e50d?pvs=4

https://github.com/user-attachments/assets/020790d5-6cf1-4c7a-a605-d54a3e2165e5

yanniskadiri:
Confirmed in rc.7

PR #21237, opened 2024-09-11
fix: review workflow settings icons

remidej:
### What does it do?

Fixes issue where the review workflows settings page would mess up the colors of the Admin main menu's icons (because of an odd styled-components scoping issue)

Fixed by using IconButton from the design system instead of a custom component

### How to test it?

- Go to the RW settings page
- The main nav shouldn't look any different
- Make sure the edit and delete buttons still work as expected

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20817


jhoward1994 (approved):
Looks good! 🙏🏻

PR #21247, opened 2024-09-11
fix: escape raw queries

### LINKED ISSUE #21185: Migration core::5.0.0-discard-drafts (up) crashes on restricted keywords (like: 'to')

- Node.js version: 18
- Strapi version: 5 rc 17
- Database: sqlite
- Operating system: macos

### Describe the bug

After https://github.com/strapi/strapi/issues/20225, this migration crashes again for me.

Migrating from v4 to v5 crashes core::5.0.0-discard-drafts, because it doesn't escape the values on this line: https://github.com/strapi/strapi/blob/46d57caa24827fde44e3cce8d2300acddae1c288/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts#L80
I have a form model with a `to` field. `to` is a restricted keyword in an SQL statement.

1. create a v4 with a model that has a `to` field.
2. migrate to 5.
3. MigrationError.

### Expected behavior

no Error.

### Screenshots

Scherm­afbeelding 2024-09-07 om 13 05 43

innerdvations:
### What does it do?

- escapes raw queries in db migrations
- no longer skips indexes that contain '_lnk_'
- fix sqlite package names

### Why is it needed?

to prevent errors when attribute names are sql keywords

### How to test it?

see #21185

### Related issue(s)/PR(s)

Fixes #21185 

DX-1606


packages/core/database/src/migrations/internal-migrations/5.0.0-01-convert-identifiers-long-than-max-length.ts
```diff
@@ -76,15 +76,10 @@ const renameIndex = async (knex: Knex, db: Database, diff: IndexDiff) => {
     debug(`not renaming index ${full.indexName} because name hasn't changed`);
     return;
   }
-  // link tables are new and do not need to be migrated
+
   // fk indexes can't be easily renamed, and will be recreated by db sync
   // if this misses something due to the loose string matching, it's not critical, it just means index will be rebuilt in db sync
-  if (
```
innerdvations, line 82: This was ridiculous and nonsensical (I can call it that because it was my own code) and came from a misunderstanding of the localizations_links tables and a bug that has since been fixed

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -77,15 +77,15 @@ async function copyPublishedEntriesToDraft({
    */
   await trx
     // INSERT INTO tableName (columnName1, columnName2, columnName3, ...)
-    .into(trx.raw(`${meta.tableName} (${scalarAttributes.join(', ')})`))
+    .into(trx.raw(`?? (??)`, [meta.tableName, scalarAttributes]))
```
alexandrebodin, line 80: I'm not sure you can spread an array into a single ?? identifier 🤔 you might need to do sth like 

```
trx.raw(`?? (${scalarAttributes.map(() => `??`).join(', ')})`, [meta.tableName, scalarAttributes])
```

alexandrebodin (changes requested):
Just one doubt for a query.

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
innerdvations: Ahhh, you're right. I wonder why it actually works though? Undocumented feature? 😆

I'll fix it just to be safe.

alexandrebodin (approved):
Looks much safer 👍

packages/core/database/src/migrations/internal-migrations/5.0.0-01-convert-identifiers-long-than-max-length.ts
```diff
@@ -99,18 +94,21 @@ const renameIndex = async (knex: Knex, db: Database, diff: IndexDiff) => {
...
+            short.indexName,
+          ])
           .transacting(trx);
       } else if (client === 'pg' || client === 'postgres') {
         await knex
-          .raw(`ALTER INDEX "${full.indexName}" RENAME TO "${short.indexName}"`)
+          .raw('ALTER INDEX ?? RENAME TO ??', [full.indexName, short.indexName])
           .transacting(trx);
-      } else if (client === 'sqlite' || client === 'better') {
+      } else if (['sqlite', 'sqlite3', 'better-sqlite3'].includes(client as any)) {
```
innerdvations, line 107: I actually think this can only be 'sqlite' but it makes zero difference here and even if it did it's only a debug log, so I'm leaving it to be safe

Issue #21249, opened 2024-09-11
[CM][I18N] Cannot publish new locale directly


alexandrebodin:
**Steps**

- Create a new document
- Save or publish in any locale for the 1st time
- Switch locale, add some content. Try to "publish" directly without saving. See the error.

**Expected behavior**
- Should publish correctly

<img width="1800" alt="Screenshot 2024-09-11 at 22 00 03" src="https://github.com/user-attachments/assets/e9062f8d-5b65-46c3-a1d9-98b7089c1c32">

simotae14:
The error is "Document not found"

simotae14:
fixed with PR https://github.com/strapi/strapi/pull/21251

PR #21251, opened 2024-09-12
fix: allow publish new locale directly

simotae14:
### What does it do?

It handles the case when you try to publish a document in a different locale directly (it was not possible before because in the controller we try to find the document for the new locale but not checking if the document exist)

### Why is it needed?

Because otherwise everytime we try to publish directly a document new locale we receive the error "Document not found"

### How to test it?

- Create a new document
 - Save or publish in any locale for the 1st time
 - Switch locale, add some content. Try to "publish" directly without saving. It has to publish the document

### Related issue(s)/PR(s)

CS-1116
https://github.com/strapi/strapi/issues/21249


packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -417,14 +417,22 @@ export default {
...
-        if (!document) {
+        if (!document && !documentExists) {
           throw new errors.NotFoundError('Document not found');
         }
 
+        // change the locale of the document if it's different from the current locale to check the permissions
+        const documentCopy = { ...document };
+        if (documentCopy.locale !== locale) {
+          documentCopy.locale = locale;
+        }
```
remidej, line 433: why not mutate document directly?

simotae14: because we use it afterwards for other checks

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -417,14 +417,22 @@ export default {
 
       const isUpdate = !isCreate;
       if (isUpdate) {
-        document = await documentManager.findOne(id!, model, { populate, locale });
+        // check the document version
+        document = await documentManager.findOne(id!, model, { populate });
```
alexandrebodin, line 421: I would recommend changing a bit the logic to avoid the permission copy you have to make 

pseudo code

```
if documentExists does not exist then
  throw error
if document with locale doesn't exist  then
 create and publish that new version  // runs the create permission check 1st for the new locale and then the publish permission which is more accurate
else just publish the existing one
```

simotae14: If document with locale doesn't exist but the documentExists is true I need to update the document not to create a new one, because otherwise I have different document Ids for the different locales

alexandrebodin: Yes exactly you have to create the new locale (not a new document 👍)

simotae14: I added a new logic, trying to follow your suggestions, let me know if it works for you, thanks

alexandrebodin (changes requested):

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -417,14 +417,22 @@ export default {
...
-        if (!document) {
+        if (!documentExists) {
           throw new errors.NotFoundError('Document not found');
         }
 
-        // Only Update if user has update permissions
-        if (permissionChecker.can.update(document)) {
+        if (!document) {
+          // update and publish the new version
+          if (permissionChecker.cannot.update() || permissionChecker.cannot.publish()) {
```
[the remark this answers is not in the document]
simotae14: I checked and we don't have this issue in single types

alexandrebodin (approved):
The code looks good to me. Just requires some QA and testing with different RBAC setup (on locale permissions in particular)

alexandrebodin:
@simotae14 If you can add a ticket to add some tests afterwards if you can't do it right now ;)

simotae14:
Ticket created CS-1123

oiorain:
- Create a new document
- Save or publish in any locale for the 1st time
- Switch locale, add some content. Try to "publish" directly without saving. See the error.

I can confirm those steps but when i see the error, the page reloads and im loosing the content. is this supposed to happen?

oiorain:
LGTM!

Issue #21227, opened 2024-09-11
[Releases x RBAC] Weird behaviour on releases with specific permissions


MarionLemaire:
## Bug report

- Node.js version: v20.9.0
- Strapi version: v5.0.0-rc.19
- Operating system: WSL

### Describe the bug

I was testing RBAC and permissions on V5 stable and faced a weird behaviour through releases : my Admin user was not able to proceed the release (even with all permissions), only the user with the role I've created was able to proceed the release.

### Steps to reproduce the behavior

1. With a super admin profile, create 2 collection types (here : Owner and Dog)
2. Create a role called "MyRole", allowing all permissions for Owner and Dog collection types, and allowing permissions with custom conditions as in the screenshots for releases
3. Invite a new user with the role "MyRole"
4. Sign up with this new user
5. Create entries with this new user for the 2 collection types Dog and Owner
6. Add these entries to a new release
7. Go back to Admin user profile, try to proceed the release 
8. Go back to MyRole user profile, try to proceed the release

First video is with Super admin user, who can't Publish an entry through Release because of an error (which is not relevant as Relation field seems OK) (+ user is able to publish this item through Edit View, but not through Releases).
https://www.notion.so/strapi/Release-x-RBAC-weird-behaviour-7ad09ea794f7455c8b4cd2cfbae3c0c5?pvs=4#f69c71a897ef45a6983d0dee9b1b00c7

In second video, it's another user (with role "MyRole" I've created) who has access to Releases and to the CT included in this release, but who can only update and publish releases through these permissions (screenshots below) :
https://www.notion.so/strapi/Release-x-RBAC-weird-behaviour-7ad09ea794f7455c8b4cd2cfbae3c0c5?pvs=4#46f3fa1c1abe41e8a1f4f8b0065faa36
![image (4)](https://github.com/user-attachments/assets/a5ea91ce-d5ad-48b8-bf49-dd4f04a2a7a6)

Custom conditions here are "edit", "remove entry" and "add entry" if "is creator" of the release)
![image (5)](https://github.com/user-attachments/assets/b15bb0cc-3613-432b-9d67-3d3162cad671)

### Expected behavior

Both user should be allowed to proceed the release.

PR #21256, opened 2024-09-12
fix(content-releases): validation and cache releases

Feranchz:
### What does it do?

Fix two bugs that were affecting content releases.
* When validating relations to many entries that are required we want to validate id is a number, not a string
* We need to invalidate cache data of release when entries are added to a release

### How to test it?

1. Add one entry to a release with "hasMany" relation, validation should work 

1. Create a release
2. Add one entry to that release and move to the release page
3. Release status should be ready and publish button enabled

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21227


remidej (approved):
Perfect 🚂

PR #21260, opened 2024-09-13
fix: allow morphToMany relation ordering in sanitize and validate

innerdvations:
### What does it do?

- allows morph relations set, connect, disconnect, and options properties to validate/sanitize correctly

### Why is it needed?

This PR is necessary for v5 stable to allow polymorphic relations input to validate/sanitize properly

It came from https://github.com/strapi/strapi/pull/21135 but that PR has too many potential side effects to merge into v5 this close to release, so only this critical part is being added for now

### How to test it?

- verify that polymorphic relations can set/disconnect/connect through validation -- beware that they may not actually work correctly due to [this bug](https://github.com/strapi/strapi/issues/20961), the thing to test is that it makes it through validation
- ensure that it properly sanitizes and validates connect, disconnect, set, and options and doesn't allow bad values to pass through

Automated tests have been added for every positive and negative case (though it's missing results testing on the api tests because results would be wrong)

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/21135

### DIFF
--- a/packages/core/utils/src/__tests__/morph.test.ts
+++ b/packages/core/utils/src/__tests__/morph.test.ts
@@ -15,7 +15,7 @@ describe('restricted relations', () => {
       target: 'admin::user',
     };
 
-    // we don't care about actual relation permissions here, only that the properties are not removed
+    // we don't care about actual relation permissions here, only that the properties are sanitized and validated correctly
     global.strapi = {
       auth: {
         verify(_config, auth) {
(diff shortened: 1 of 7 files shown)


alexandrebodin (approved):
code LGTM. Can you add an api test that send those params ?

jhoward1994:
Makes sense to me - just so I've understood correctly the actual attachment of morphToMany relations is broken in some cases still. But as we will be fixing and supporting this in future we need the validation to be correct for the first v5 stable?

innerdvations:
Yep, exactly. That's why the media tests are weird and the api tests are only partially implemented; many of them would fail. So for now we just test that validation and sanitization are working properly on them, so people can still make use of them even if their behaviour is a bit wonky right now.

jhoward1994 (approved):

PR #21262, opened 2024-09-13
enhancement: add version param when send request to marketplace

butcherZ:
### What does it do?

added a `version` parameter when sending request to market place

### Why is it needed?

Allow us to filter plugins depends on Strapi App version

more detailed description [here](https://strapi-inc.atlassian.net/browse/DX-1607)

### How to test it?

Inspect in dev tool under network tab, the payload (for both Plugins and Providers) should contain a `version` parameter when on the Marketplace page

### Related issue(s)/PR(s)

DX-1607

### DIFF
--- a/packages/core/admin/admin/src/pages/Marketplace/hooks/useMarketplaceData.ts
+++ b/packages/core/admin/admin/src/pages/Marketplace/hooks/useMarketplaceData.ts
@@ -14,6 +14,7 @@ interface UseMarketplaceDataParams {
   debouncedSearch: string;
   query?: MarketplacePageQuery;
   tabQuery: TabQuery;
+  strapiVersion?: string | null;
 }
 
 type Collections =
@@ -101,6 +102,7 @@ function useMarketplaceData({
   debouncedSearch,
   query,
   tabQuery,
+  strapiVersion,
 }: UseMarketplaceDataParams) {
   const { notifyStatus } = useNotifyAT();
   const { formatMessage } = useIntl();
@@ -131,6 +133,7 @@ function useMarketplaceData({
     ...tabQuery.plugin,
     pagination: paginationParams,
     search: debouncedSearch,
+    version: strapiVersion,
   };
 
   const { data: pluginsResponse, status: pluginsStatus } = useQuery(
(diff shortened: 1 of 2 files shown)


examples/getstarted/src/api/kitchensink/content-types/kitchensink/schema.json
```diff
@@ -152,10 +152,6 @@
     "cats": {
       "type": "dynamiczone",
       "components": ["basic.relation", "basic.simple"]
-    },
-    "slug": {
-      "type": "uid",
-      "targetField": "short_text"
```
innerdvations, line 158: I think this made it in by mistake :)

Bassel17:
looks good, you just need to remove the example folder changes

packages/core/admin/admin/src/pages/Marketplace/hooks/useMarketplaceData.ts
```diff
@@ -14,6 +14,7 @@ interface UseMarketplaceDataParams {
   debouncedSearch: string;
   query?: MarketplacePageQuery;
   tabQuery: TabQuery;
+  strapiVersion?: string | null;
```
jhoward1994, line 17: Should it be optional? I think in v5 onwards we can expect it to be there every time?

butcherZ: yeah I wasn't really sure about that, the `strapiVersion` originally coming from the `AppInfo` context where it was typed like this

jhoward1994: Ah okay, makes sense if it matches the types there

jhoward1994:
I think you need a `source:` label top make this check pass

`Checks / check-pr-status (pull_request) Failing after 6s`

jhoward1994 (approved):
👏🏻👏🏻👏🏻

butcherZ:
> I think you need a `source:` label top make this check pass
> 
> `Checks / check-pr-status (pull_request) Failing after 6s`

Ahhh thank you! totally forgot about that one

Issue #21240, opened 2024-09-11
[Relations] Incoherent behaviour of relations field at selecting and saving + no drag & drop


MarionLemaire:
### Required System information

Node.js version: v18.20.3
Strapi version: v5.0.0-rc.20
Operating system: IOS

### Describe the bug

At selecting relations in Relations field for an entry, behaviour is weird : for a 1to1 relation field, it's possible to select many relations.
Then at saving, only 1 remains. For a many to 1 relation, it's possible to select many relations (that sounds logical) but at saving only 1 remains. Also, no drag & drop available for relations items.

### Steps to reproduce the behavior

Create a CT with at least 2 relations fields : 1 field 1 to 1, 1 field 1 to many.
![Screenshot 2024-09-11 at 14 59 39](https://github.com/user-attachments/assets/4816e234-fa1b-4bc7-ab77-9c26a505706e)
![Screenshot 2024-09-11 at 14 59 55](https://github.com/user-attachments/assets/a11fa5a8-80cb-4a79-98e0-d5cca8fe7f50)

Save and go to CM
Create an entry for this CT
In Edit view, fill in the 2 relations field : 
- 1 to 1 : try to add many items
- 1 to many : try to add many items
Try to drag and drop the different relations to reorder them.
Save

### Expected behavior

1 to 1 relation field should select only 1 item
many to 1 relation field should select many items and propose drag & drop
after saving, all data must be correctly saved

### Screenshots

[If applicable, add screenshots to help explain your problem.](https://www.loom.com/share/7cba13c5a60242c092cc1348595f4bef)

Discussed with @meganelacheny

PR #21263, opened 2024-09-13
fix: single relation input allowing multiple value

remidej:
### What does it do?

Prevents the relation input from having multiple relations selected for toOne relations.

This was only a frontend issue, the backend always only accepted 1 relation.
### Why is it needed?

### How to test it?

- Create a relation field of type oneWay, oneToOne, or manyToOne
- Add one relation in that field. Then another. It should replace the 1st one, never have the 2 relations at the same time
- Save the content
- Try to add another relation. It should replace the previous one. (it's at this step that we had the bug)
- Create a relation field of type oneToMany or manyToMany. You should be free to select several relations for it.
### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21240


remidej:
The ❌ icon button change was validated by @lucasboilly

simotae14 (approved):
The code looks good, let me try to QA it

lucasboilly:
@remidej All good!

simotae14:
Tested and it works 👏🏽

jhoward1994 (approved):

PR #21266, opened 2024-09-13
fix: wrong jobs value after removing a task

jotafilaj:
### What does it do?
In cron service remove method, the array.filter method is replaced with array.splice method so the jobs property would point to the correct value.

### Why is it needed?
In cron service remove method, array.filter method was used which created a shallow copy of a jobsSpec array. This meant that the jobs property pointing to the jobsSpec array was not updated. So after removing a task, jobs returned the old array with the removed task.

### How to test it?
Add a cron task, log the jobs array, remove a cron task, log the jobs array. After the fix, removed task is no longer present in the jobs array.


hanpaine:
> [!IMPORTANT] 
> With the release of Strapi 5, we have lifted the code freeze. To ensure we can effectively review and assist with contributions, we are now only accepting PRs targeting version 5 (the develop branch).<br/> 
> As your PR is targeting some of the most recent v4 version, we we would greatly appreciate it if you could update your contribution to align with version 5 and resubmit! Thanks in advance 🙏

jhoward1994:
Hi @jotafilaj, the fix looks good to me. I think it's something we can apply to both v4 and v5

I've made another PR to include this fix in v5, I'll wait for that to be approved and merge these together.

jhoward1994 (approved):

Issue #20955, opened 2024-08-06
[Strapi v5 RC] Missing schema in defaultSanitizeOutput


komat1:
### Required System information
Node: 20.15.1
System: MacOS

Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0-rc.9
Node/Yarn Version: npm/10.7.0 node/v20.15.1 darwin arm64 workspaces/false
Edition: Community
Database: mysql

### Describe the bug

When I try to add a user to the strapi (from REST, GraphQL or admin panel), this error comes up: Missing schema in defaultSanitizeOutput.
But it adds the user anyway.
Also, it seems to only bug out when "Enable email confirmation" in the advanced settings is set to **TRUE**

### Steps to reproduce the behavior

1. Go to Strapi
2. Click on Create user using add
4. See error in the console

### Additional context

Also, strapi doesnt send emails after registering, but sending emails work.
**when using GraphQL**
{
  "errors": [
    {
      "message": "Email or Username are already taken",
      "path": [
        "register"
      ],
      "extensions": {
        "code": "STRAPI_APPLICATION_ERROR",
        "stacktrace": [
          "ApplicationError: Email or Username are already taken",
          "    at Object.register (/Users/matejkolestik/Desktop/gabnoc/gabnoc/node_modules/@strapi/plugin-users-permissions/server/controllers/auth.js:344:13)",
          "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)",
          "    at async resolve (/Users/matejkolestik/Desktop/gabnoc/gabnoc/node_modules/@strapi/plugin-users-permissions/server/graphql/mutations/auth/register.js:24:7)"
        ],
        "error": {
          "name": "ApplicationError",
          "message": "Email or Username are already taken",
          "details": {}
        }
      }
    }
  ],
  "data": null
}

dazzafact7:
same issue when trying to create an user

```
const registerResponse = await axios.post(`${strapiUrl}/api/auth/local/register`, {
        username: user.username,
        email: user.email,
        password: user.password
      }, {
        headers: { 'Authorization': `Bearer ${strapiToken}` }
      });
```
User will be created, but with error and not getting an ID back as response

alexandrebodin:
Testing this on rc.12 I cannot reproduce. @dazzafact7 do you still have this if you upgrade to rc.12 ?

xXnikosXx:
Stumbled upon this error myself just now, using v5.0.0-rc.19.

[...]

PR #21268, opened 2024-09-16
fix: emailConfirmation broken

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/20955


innerdvations (approved):
LGTM

PR #21271, opened 2024-09-16
fix: yarn 4 support for create-strapi-app install

innerdvations:
### What does it do?

when using create-strapi-app with yarn 4, use env YARN_NETWORK_TIMEOUT instead of `--network-timeout` command line option

### Why is it needed?

yarn 4 has removed --network-timeout as a cli option and it was failing with an error when it tried to run `yarn install --network-timeout 1000000`

### How to test it?

create-strapi-app should work with yarn 4 as well as previous versions of yarn, and still work with `pnpm` and `npm`

### Related issue(s)/PR(s)

DX-1611

### DIFF
--- a/packages/cli/create-strapi-app/src/create-strapi.ts
+++ b/packages/cli/create-strapi-app/src/create-strapi.ts
@@ -214,27 +214,84 @@ async function createApp(scope: Scope) {
 
 const installArguments = ['install'];
 
-const installArgumentsMap = {
+type VersionedArgumentsMap = {
+  [key: number]: string[]; // Maps version numbers to argument arrays
+};
+
+type VersionedEnvMap = {
+  [key: number]: Record<string, string>; // Maps version numbers to environment variables
+};
+
+// set command line options for specific package managers
+const installArgumentsMap: {
+  [key: string]: string[] | VersionedArgumentsMap;
+} = {
   npm: ['--legacy-peer-deps'],
-  yarn: ['--network-timeout 1000000'],
+  yarn: {
+    1: ['--network-timeout', '1000000'],
+    2: ['--network-timeout', '1000000'],
+    3: ['--network-timeout', '1000000'],
+  },
   pnpm: [],
 };
 
-function runInstall({ rootPath, packageManager }: Scope) {
+// set env vars for specific package managers
+const installEnvMap: {
+  [key: string]: Record<string, string> | VersionedEnvMap;
+} = {
+  yarn: {
+    4: { YARN_NETWORK_TIMEOUT: '1000000' },
+  },
+  npm: {},
+  pnpm: {},
+};
+
+async function getPackageManagerVersion(packageManager: string): Promise<number | null> {
+  try {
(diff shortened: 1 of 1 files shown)


alexandrebodin (commented):
A few question

- Isn't the env var compatible with prvious version of yarn ? we could make the code much simpler by just setting the env var. 
- does yarn v4 fails with the --network-timeout option or just ignores it ?

innerdvations:
Unfortunately it fails, and it was added in yarn 4

I made it a bit more complicated so that we can generically support the same things (versioned env and options) in any package manager, but I could just simplify it to a specific check for yarn 4.

packages/cli/create-strapi-app/src/create-strapi.ts
```diff
@@ -214,27 +214,84 @@ async function createApp(scope: Scope) {
...
+  },
+  npm: {},
+  pnpm: {},
+};
+
+async function getPackageManagerVersion(packageManager: string): Promise<number | null> {
+  try {
+    const { stdout } = await execa(packageManager, ['--version']);
+    const version = stdout.trim();
+    const majorVersion = parseInt(version.split('.')[0], 10);
```
alexandrebodin, line 253: Should we use semver to parsde so we could have more precise check if needed ?

packages/cli/create-strapi-app/src/create-strapi.ts
```diff
@@ -214,27 +214,84 @@ async function createApp(scope: Scope) {
...
+  }
+
+  // Get install arguments
+  let packageOptions: string[] = [];
+  const argsMap = installArgumentsMap[packageManager];
+  if (packageManagerVersion && typeof argsMap === 'object' && packageManagerVersion in argsMap) {
+    packageOptions = (argsMap as VersionedArgumentsMap)[packageManagerVersion] || [];
+  } else if (Array.isArray(argsMap)) {
+    packageOptions = argsMap;
+  }
```
alexandrebodin, line 281: Can you move this into its own func `const { envArgs, cmdArgs } = getPackageManagerInstallationArgs(...)`

Another suggestion to avoid all the repeating 

```js
const argsMap = {
	 yarn: {
	    default: { cmdArgs: [--network-timeout, 10000] }
	    versionOverwrites: {
	      4: { envAgs: [NETWORK_TIMEOUT....], cmdArgs: [] }
	     }
	  }
 }
```

Also you could check if there are versionOverwrites and only parse the version if needed (one less cmd to run 🎉 )

packages/cli/create-strapi-app/src/create-strapi.ts
innerdvations: yeah, I thought about that but worried it might complicate it even more than I did 😆
but I'll do that so we can (for example) make the assumption that yarn 5 will also need the env var with a range rather than adding in another major version

alexandrebodin: 👍

PR #21282, opened 2024-09-16
Use a script instead of a bootstrap function for example data

alexandrebodin:
### What does it do?

- Runs seed script only in development env & auto add a skip_seed env var to .env for local use to avoid rerunning it if the DB is being cleared


alexandrebodin:
Experimental `0.0.0-experimental.f74ae50eea1ce95176f088dba837e95b60fa2a4d`

alexandrebodin:
Creating a new exp here https://github.com/strapi/strapi/actions/runs/10886863023

packages/cli/create-strapi-app/templates/example-js/scripts/seed.js
```diff
@@ -235,17 +254,27 @@ async function importSeedData() {
...
-      console.error(error);
-    }
-  }
-};
+  const appContext = await compileStrapi();
+  const app = await createStrapi(appContext).load();
+
+  // seedExampleApp().catch((error) => {
+  //   console.error(error);
+  //   process.exit(1);
```
innerdvations: is this here on purpose?

packages/cli/create-strapi-app/templates/example/package.json
```diff
@@ -4,7 +4,8 @@
     "start": "strapi start",
     "build": "strapi build",
     "strapi": "strapi",
-    "deploy": "strapi deploy"
+    "deploy": "strapi deploy",
+    "seed-example": "node ./scripts/seed.js"
```
innerdvations, line 8:
```suggestion
    "seed:example": "node ./scripts/seed.js"
```
to follow our normal pattern, could add other seed data in the future

packages/cli/create-strapi-app/src/create-strapi.ts
```diff
@@ -152,6 +154,22 @@ async function createApp(scope: Scope) {
 
   logger.title('Strapi', `Your application was created!`);
 
+  if (shouldRunSeed) {
+    if (await fse.exists(join(rootPath, 'scripts/seed.js'))) {
+      logger.title('Seed', 'Seeding your database with sample data');
+
+      try {
+        await execa(packageManager, ['run', 'seed-example'], {
```
innerdvations, line 162:
```suggestion
        await execa(packageManager, ['run', 'seed:example'], {
```

packages/cli/create-strapi-app/templates/example/scripts/seed.js
```diff
@@ -248,3 +253,28 @@ async function importSeedData() {
...
+if (require.main === module) {
+  main().catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });
+} else {
+  module.exports = {
+    seedExampleApp,
+  };
+}
```
innerdvations, line 280: Do we use this pattern anywhere else? Might want to add a comment about it since it's a bit unusual.

innerdvations (commented):
Code LGTM!

@maccomaccomaccomacco do you have some time to test it out?

packages/cli/create-strapi-app/templates/example-js/scripts/seed.js
alexandrebodin: nono already cleaned up

packages/cli/create-strapi-app/templates/example/package.json
alexandrebodin: why not

packages/cli/create-strapi-app/templates/example/scripts/seed.js
alexandrebodin: it's actually very common in node :) I'll remove it I just used it for testing

alexandrebodin:
@maccomaccomaccomacco Experimental to test `0.0.0-experimental.30e5d19d03d07d2b73ac7ed4b376a833d4270bde`

maccomaccomaccomacco (approved):
Looks and works good to me.
Thank you <3

Issue #21269, opened 2024-09-16
Enabling D&P after creating entries without D&P makes entries disappear


yanniskadiri:
Steps:
- Create a Content Type without D&P
- Create some entries for it
- Enable D&P in the CTB
- All the entries disappeared

**→ In DB they still exist and are created as published but they are missing there draft versions that’s why they are not showing up**

Notion link: https://www.notion.so/strapi/Enabling-D-P-after-creating-entries-without-D-P-makes-entries-disappear-240ec7132c2e42c5af423ba449b255a4

PR #21288, opened 2024-09-16
fix(i18n): call d&p & i18n migrations in the same file

Feranchz:
### What does it do?

Now that i18n is core we can have d&p & i18n migrations together to avoid race conditions when running migrations.

### How to test it?

1. Create a Content Type without draft & publish and i18n.
2. Create some entries
3. Add draft & publish and i18n to the content type
4. It should keep the entries created

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21269


packages/core/core/src/migrations/index.ts
```diff
@@ -0,0 +1,14 @@
+import draftPublishMigrations, { type Input } from './draft-publish';
+import i18nMigrations from './i18n';
+
+const enable = async ({ oldContentTypes, contentTypes }: Input) => {
+  await i18nMigrations.enable({ oldContentTypes, contentTypes });
+  await draftPublishMigrations.enable({ oldContentTypes, contentTypes });
+};
+
+const disable = async ({ oldContentTypes, contentTypes }: Input) => {
+  await i18nMigrations.disable({ oldContentTypes, contentTypes });
```
alexandrebodin, line 10: should this be inverted ? 

usually you do a-b<->b-a  ?

packages/core/core/src/migrations/index.ts
```diff
@@ -0,0 +1,14 @@
+import draftPublishMigrations, { type Input } from './draft-publish';
```
alexandrebodin, line 1: prefer import with * and remove default exports :)

alexandrebodin (changes requested):
Code commented. Once change I'll do some QA. Overall looks good

packages/core/core/src/migrations/index.ts
Feranchz: When disabling is not relevant I think (It's working in this way). discardDraft is going to use the locale defined on the entry because Content Type has not i18n (In this case is going to be always null because i18n migration run first, but otherwise is going to use the old locale)

On enable it should be i18n migration first, because discardDraft adds the default locale because content type has i18n enabled.

alexandrebodin (approved):
Code looks good, and QA validated on my side.

I'll merge later tonight

Issue #21294, opened 2024-09-16
Can create a ST in a specific locale while NOT having the permissions to do so


hanpaine:
### Describe the bug
I shouldn’t be able to create a single type with D&P enabled in French when I do not have the rights to CREATE or UPDATE it, but it is possible

### Steps to reproduce the behavior

1. Create Author without CREATE or UPDATE permissions in Admin for 1 of the locales (e.g. Author_EN without FR perms)
2. As Author_EN (or whatever you're called) create a new single type and edit an existing one
![Image](https://github.com/user-attachments/assets/b9c8d849-b0a0-42e7-8e8d-95c6912d9018)

![Image](https://github.com/user-attachments/assets/2dbe7ab6-7069-4240-8ce2-e4834b4d7deb)

### Expected behavior

RBAC rules are adhered to

PR #21295, opened 2024-09-16
[Fix/i18n] Single type deletion and locale nav

jhoward1994:
### What does it do?

Fix https://github.com/strapi/strapi/issues/21294
Fix https://github.com/strapi/strapi/issues/21296

### Why is it needed?

- Fix for locale deletion in single types (for single types we should not require a documentId to proceed with the deletion as the CM url doesn't contain one)
- Fix to disable CM navigation to a locale you don't have permission for (you need the `canCreate` permission if the locale doesnt exist yet & `canRead` if it does)

DX-1612
DX-1613


innerdvations:
LGTM! Can you add an e2e test for it?

jhoward1994:
I have to go now but will add this test next https://github.com/strapi/strapi/blob/6819dcbd7ca367f785b9f33786b3190eec1c21dc/tests/e2e/tests/i18n/editview.spec.ts#L306-L308

hanpaine:
QAed by @butcherZ ✅

innerdvations (approved):
LGTM!

PR #21299, opened 2024-09-16
fix: manyToOne relation input

remidej:
### What does it do?

Fixes this issue:

https://www.loom.com/share/01b04be39380436a8befe4f79b27ba11?sid=f945ffb3-55e9-4bb0-a510-59152a602421

It's a follow-up to my previous PR because there was a case we didn't check:

- Have a relation saved already in a toOne relation field
- Select another related entry instead. The previous value of the relation should be removed 
- **do that again**: this is what we didn't check. It makes sure that relations that haven't been saved yet are also removed to make room for the new ones

### How to test it?

Follow the steps above, also save at various points and refresh. Also sanity checks:
- play with other relation types
- make sure the "remove" icon button works for each related entry


oiorain:
LGTM! nice job

simotae14 (approved):
The code looks good to me

Issue #21287, opened 2024-09-16
[RBAC/CTM] Missing relational entries, lost of relations when creating an entry in several locales


yanniskadiri:
When giving full CRUD access + admin users access to an admin user with this role. 

I can create an entry with two options for my relation. 

However, after creating the entry in another locale, the data starts to be inconsistent with the value being applied to two locales, then listing only one of the two options available in the relational field. See video for more.

https://github.com/user-attachments/assets/1b509a02-28f3-43db-bfae-7f75b556ad77

Issue #21286, opened 2024-09-16
[Content Manager/i18n] The relations are lost when adding a new language version of an entity


yanniskadiri:
Reproduction steps:
- Enable the i18n internalization on a Content Type already existing with several entities and relations in it (didn’t tested with a fresh Content Type)
- Try to add a new language to an entity
- Save some content on the newly added language
- Go back to original entity
- See relations disappeared

![Image](https://github.com/user-attachments/assets/8264dca8-2a89-4b92-af97-ca8cd836f616)
![Image](https://github.com/user-attachments/assets/4babf860-5611-4c65-8063-5d7bfc8e7299)
![Image](https://github.com/user-attachments/assets/77a9695c-b9da-4e9d-9566-25628ce1c24d)

PR #21301, opened 2024-09-16
fix: relations filtering on locales

alexandrebodin:
### What does it do?

Fix relation findAvalaible 

Fix https://github.com/strapi/strapi/issues/21287
Fix https://github.com/strapi/strapi/issues/21286

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/server/src/controllers/relations.ts
+++ b/packages/core/content-manager/server/src/controllers/relations.ts
@@ -91,12 +91,8 @@ const validateLocale = (sourceUid: UID.Schema, targetUid: UID.ContentType, local
   const isSourceLocalized = isLocalized(sourceModel);
   const isTargetLocalized = isLocalized(targetModel);
 
-  let validatedLocale = locale;
-
-  if (!targetModel || !isTargetLocalized) validatedLocale = undefined;
-
   return {
-    locale: validatedLocale,
+    locale,
     isSourceLocalized,
     isTargetLocalized,
   };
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/server/src/controllers/relations.ts
```diff
@@ -93,7 +93,7 @@ const validateLocale = (sourceUid: UID.Schema, targetUid: UID.ContentType, local
 
   let validatedLocale = locale;
 
-  if (!targetModel || !isTargetLocalized) validatedLocale = undefined;
+  if (!targetModel && !isTargetLocalized) validatedLocale = undefined;
```
Feranchz, line 96: I think we can safely remove this line... I tried it locally and it was working.

We only use locale to filter if `isTargetLocalized` or `isSourceLocalized` is true. (depending on the case) So this looks like not necessary

alexandrebodin: agreed

alexandrebodin:
Making an experimental here for QA @Feranchz cc @oiorain 
https://github.com/strapi/strapi/actions/runs/10898914294/job/30243144943

oiorain:
@alexandrebodin regarding https://github.com/strapi/strapi/issues/21287
Im not sure this case is covered ny this fix but there is still this bug: 
- Create a CT in english
- create the french version: at first the relation is showing empty but after you save it appears with the same value as the english verison 
- same for a third japanese version

oiorain:
confirmed this is not happening on the experimental! 
good job! 🚀

Issue #21293, opened 2024-09-16
[RBAC/RW] RW isn’t visible/working for ST


yanniskadiri:
I enabled the Default RW on the ST “Cookie Banner”, it isn’t visible.

![Image](https://github.com/user-attachments/assets/34ad3444-0d51-4019-962b-3271ef36e08a)
![Image](https://github.com/user-attachments/assets/dec1de51-8fbf-4584-9f9e-e73c1677ad01)

alexandrebodin:
Fixed

oiorain:
@alexandrebodin 
I can see the RX box now but it seems i dont have the rights to change the status 
https://www.loom.com/share/fa4a271b02404a81987539c697adf392?sid=651eceb1-279e-466d-8bda-6ca4cf7bde6e

Issue #21292, opened 2024-09-16
[RBAC/RW] I cannot assign someone to an entry while having the permissions to do so


yanniskadiri:
I cannot assign someone to an entry with RW whereas I’m at a stage when I have the permissions.

Steps to reproduce:

- Create a CT “Article” with whatever fields
- Create an entry “Test” with the Admin Role
- Create a new role “Writer” with full access to the “Article”
- Create a new user with the Writer role assigned
- Update the Default workflow and let the new role “Writer” update the stage when reaching “Ready to Review”

- Open a new private tab, connect to the second user
- Go to the entry “Test”, see that you cannot edit the stage, you cannot assign either.
- Use the Admin to move to “Ready to Review” stage, and try to assign again with the second user, it still doesn’t work

![Image](https://github.com/user-attachments/assets/fe89828b-be0e-4d68-b6d9-0b6139dba407)

remidej:
I'm currently testing if Alex's PR fixed this

PR #21303, opened 2024-09-16
fix: review workflow issues

alexandrebodin:
### What does it do?

- Fix https://github.com/strapi/strapi/issues/21293
- Might have fixed https://github.com/strapi/strapi/issues/21292 as I cannot reproduce it on my branch

### DIFF
--- a/packages/core/review-workflows/admin/src/routes/content-manager/[model]/[id]/components/AssigneeSelect.tsx
+++ b/packages/core/review-workflows/admin/src/routes/content-manager/[model]/[id]/components/AssigneeSelect.tsx
@@ -45,7 +45,7 @@ const AssigneeSelect = () => {
       documentId: id,
     },
     {
-      skip: !id,
+      skip: !id && collectionType !== 'single-types',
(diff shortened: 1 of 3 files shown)


packages/core/review-workflows/admin/src/routes/content-manager/[model]/[id]/components/AssigneeSelect.tsx
```diff
@@ -55,7 +55,9 @@ const AssigneeSelect = () => {
 
   const [updateAssignee, { error, isLoading: isMutating }] = useUpdateAssigneeMutation();
 
-  if (!collectionType || !model || !id) {
+  console.log(users);
```
Comment (simotae14) [inline packages/core/review-workflows/admin/src/routes/content-manager/[model]/[id]/components/AssigneeSelect.tsx:58]: you need to remove the console llog

simotae14 (commented):
The code looks good to me, there is a console log to remove...I need to QA it

simotae14 (approved):
QA and it works for the first case, I was not able to replicate the second bug so I assume it solves it

Issue #21270, opened 2024-09-16
CreatedBy is empty when creating new locale. Only the updatedBy is set


yanniskadiri:
Steps
- When create a new locale the createdBy is not set in DB and not disaplyed in the document info
- The UpdatedBy works fine

![Image](https://github.com/user-attachments/assets/6dc42dca-07f2-4b4a-b492-8016a572b72f)

Notion link: https://www.notion.so/strapi/CreatedBy-is-empty-when-creating-new-locale-Only-the-updatedBy-is-set-1038f359807480d3aaa9f875b2425e54

PR #21304, opened 2024-09-16
fix: createdBy missing when you create a variant of a document in a new locale

simotae14:
### What does it do?

Fixes this issue:

https://github.com/strapi/strapi/issues/21270

Just on collection Type because in Single type the ternary option was already in place

### How to test it?

Follow the steps above:
- enable i18n in a collection and add a new locale in the settings
- create a document on the default locale in the collection and save, check the string "Created n days ago by ..." in the "More" dropdown in the top right of the edit view
- change the locale and update the document, check the string "Created n days ago by ..." in the "More" dropdown in the top right of the edit view


alexandrebodin (approved):
LGTM and tested and works well. I also took the chance to fix the copy to follow @lucasboilly design

Issue #21291, opened 2024-09-16
[Releases/i18n] single type / content type delete bug


yanniskadiri:
- Create a release with a few articles + one single type in different languages
- Go to the single type in english edit view
- Delete from release.
- If you go back to the release view, you ll see an article has been deleted instead.

Deleting a single type from the release edit view works correctly.

https://www.loom.com/share/0c1f02ba29f24cdf914b3916c474fe53?sid=ee3de17d-0f18-44d3-95e0-4565c0ef4a40

PR #21306, opened 2024-09-17
fix(content-releases): delete from release

Feranchz:
### What does it do?

Find releases with one entry attached was returning all the actions on the release and not only the action we are. With this change we only return the action with the entry.

### How to test it?

1. Create a release and add multiple entries to it
2. Try to delete from the release one of the last entries
3. It should delete the right entry

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21291

### DIFF
--- a/packages/core/content-releases/server/src/controllers/release.ts
+++ b/packages/core/content-releases/server/src/controllers/release.ts
@@ -64,9 +64,15 @@ const releaseController = {
         populate: {
           actions: {
             fields: ['type'],
+            filters: {
+              contentType,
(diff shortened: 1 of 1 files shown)


alexandrebodin (commented):
Code looks good, testing it right now

alexandrebodin (approved):
This is working 👍  

for another PR : @Feranchz I think we need to also invalidate the CM cache when publishing a release. When I go back to the CM I don't see the document status updated.

simotae14 (approved):
I tested the PR and it is working

PR #21308, opened 2024-09-17
fix(content-releases): invalidate documents when release is published

Feranchz:
### What does it do?

Invalidate documents list from CM when a Release is published.

### How to test it?

1. Create a release and add some entries in-draft to it.
2. Publish the release and go to the content-manager
3. Entries status should be updated


simotae14 (approved):

remidej (approved):
Haven't done QA but code looks good

simotae14:
I tested the PR and it works, also with entries in different locales, when you go back to the CM the status is "Published"

PR #21310, opened 2024-09-17
Use correct user permissions when switching between locales

jhoward1994:
### What does it do?

Fix permission resolution when switching between locales in the CM 

### Related issue(s)/PR(s)

DX-1614


jhoward1994:
@innerdvations I will clean up the code before doing anything with this but could you test this as a fix for the CM locale switching issue we saw?

packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -202,11 +202,8 @@ const useDoc = () => {
     }
   );
 
-  const singleTypeId = collectionType === SINGLE_TYPES ? document.document?.documentId : undefined;
+  const returnId = origin || id === 'create' ? undefined : id;
```
alexandrebodin, line 205: I'm not sure I understand thise change ? for a single type this will return undefined ? 

did you remove it because it's spread below and will replace the id (document has an id) ? if so I would suggestion removing this line entirely and just spread the document below 🤔

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -35,7 +36,7 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'>;
  * components such as Blocks / Relations.
  */
 const InputRenderer = ({ visible, hint: providedHint, ...props }: InputRendererProps) => {
-  const { id } = useDoc();
+  const { id, document, collectionType } = useDoc();
```
alexandrebodin, line 39: document is spread above it's not a property so it will be undefined here

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -45,8 +46,14 @@ const InputRenderer = ({ visible, hint: providedHint, ...props }: InputRendererP
   const canUpdateFields = useDocumentRBAC('InputRenderer', (rbac) => rbac.canUpdateFields);
   const canUserAction = useDocumentRBAC('InputRenderer', (rbac) => rbac.canUserAction);
 
-  const editableFields = id ? canUpdateFields : canCreateFields;
-  const readableFields = id ? canReadFields : canCreateFields;
+  let idToCheck = id;
```
alexandrebodin, line 49: based on the comments above this would be reworked

alexandrebodin (changes requested):

packages/core/content-manager/admin/src/hooks/useDocument.ts
jhoward1994: If you compare to `releases/5` you'll see the logic for the return id here hasn't actually changed 

What we did with `const singleTypeId = collectionType === SINGLE_TYPES ? document.document?.documentId : undefined;` was an attempt to fix permissions with single types but we found it had too many side effects

These InputRenderer changes are what we've done instead https://github.com/strapi/strapi/blob/ca278734871c5ea02ab7d1bbe932089ac2bd16be/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx#L49-L52

alexandrebodin (approved):
LGTM

alexandrebodin (approved):

PR #21311, opened 2024-09-17
fix(rw): check transition permission to change assignee

remidej:
### What does it do?

fixes https://github.com/strapi/strapi/issues/21292

Basically we were checking for the "view review workflows settings" permission to update the stage, which was wrong.

So I removed it, and after [checking with Yannis](https://strapihq.slack.com/archives/C051PLC9CF7/p1726562903262239), I added a check to only allow a user to change the assignee if they have the permission to change the current stage

Note that this is only a backend fix, the field is not disabled in the frontend, but you should see the error message in the field if you don't have the permission.

### How to test it?

Follow the steps listed in Yannis' issue: https://github.com/strapi/strapi/issues/21292

### DIFF
--- a/packages/core/review-workflows/server/src/controllers/assignees.ts
+++ b/packages/core/review-workflows/server/src/controllers/assignees.ts
@@ -2,6 +2,7 @@ import type { Context } from 'koa';
 
 import { getService } from '../utils';
 import { validateUpdateAssigneeOnEntity, validateLocale } from '../validation/review-workflows';
+import { ENTITY_STAGE_ATTRIBUTE, STAGE_TRANSITION_UID } from '../constants/workflows';
 
 export default {
   /**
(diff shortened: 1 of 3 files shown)


oiorain:
Did some QA with the instructions that Yannis gave: 

Please confirm this is the behaviour we want: 
- my writer role cant change the status as first 
- admin sets status to `Ready to review` 
- Writer can then move to `in progress` but then is unable to update status again 
- Admin can change to whatever

I did observe that only Admin can set a user. and the user is not visible as set on the Writer's side:  
Left side is Writer
Screenshot 2024-09-17 at 16 37 56

remidej:
>I did observe that only Admin can set a user. and the user is not visible as set on the Writer's side:
Left side is Writer

I believe it's because as Writer you don't have the permission to view other admin users. If you add that permission to the role, then the assignee select should display the info for your Writer, and the Writer user should be able to edit the assignee if he has stage permissions

simotae14:
I did QA and it works (with the help of @remidej ) 👏🏽

Feranchz (approved):
code looks good 👍

oiorain:
QA LGTM! Lezgo

Issue #21285, opened 2024-09-16
[Content Manager] Required field validations


yanniskadiri:
Issue 1:
- Create an entry of a CT with a required field
- Fill in content other than the required field
- Save (this works fine)
- Publish - see the error message is overly technical where it should be `this value is required`

Issue 2:
- Save and publish content with the required field filled - this works as expected
- From the published entry clear the required field
- Directly publish - This succeeds when it should fail due to required validation

https://github.com/user-attachments/assets/17b5ab1c-a13f-4289-9f1d-c2860fe157ee

Agam1997:
Hi, is this ticket picked up by the strapi team?

I was looking to contribute to Strapi. Is there any bug that the team would recommend me to start off with?

oiorain:
Hi @Agam1997! 

Yes, this has been picked up already. 

If you want to find little bugs to fix to start you can check the Low and Medium bugs in the [Strapi 5 - Bug Tracker](https://github.com/orgs/strapi/projects/34/views/1?sliceBy%5Bvalue%5D=Low) board. The ones in the Backlog column have been confirmed. If there is someone assigned they're already picked up. 

Cheers! 😊

PR #21312, opened 2024-09-17
fix: fix null value bypassing required fields

butcherZ:
### What does it do?

Removed the code that allowed null values to be accepted for fields marked as required in the validation schema

### Why is it needed?

check description of issue 1 and issue 2 in #21285 

The root cause of the issues lies in the `createYupSchema` function:

- The function generates the `validationSchema` passed to the form component.
- When the `status` value is 'draft', the generated schema allows null values for all fields, including required ones.

As a result:

1. When a required text field is cleared, it becomes null.
2. The client-side validation, using the generated `ValidationSchema`, doesn't flag null value as an error.
3. The form submission bypasses client-side validation. (issue 2)
4. The null value is submitted directly to the API for the cleared required field, so the API returns the server side error directly (issue 1)

### How to test it?

For Issue 1:

Create an entry of a CT with a required field
Fill in content other than the required field
Save (this works fine)
Publish - on the required field, see the error message `this field is required`

For Issue 2:

Save and publish content with the required field filled - this works as expected
From the published entry clear the required field
Directly publish - should see the error message `this field is required` again.
### Related issue(s)/PR(s)

#21285 

DX-1616

### DIFF
--- a/packages/core/content-manager/admin/src/utils/validation.ts
+++ b/packages/core/content-manager/admin/src/utils/validation.ts
@@ -241,11 +241,7 @@ type ValidationFn = (
   options: ValidationOptions
 ) => <TSchema extends AnySchema>(schema: TSchema) => TSchema;
 
-const addRequiredValidation: ValidationFn = (attribute, options) => (schema) => {
-  if (options.status === 'draft') {
-    return nullableSchema(schema);
-  }
-
+const addRequiredValidation: ValidationFn = (attribute) => (schema) => {
   if (
     ((attribute.type === 'component' && attribute.repeatable) ||
       attribute.type === 'dynamiczone') &&


innerdvations:
I think required fields shouldn't be required to save, only on publish.

So for issue 1, to make the error "less technical" we could send an empty string for text fields instead of a `null` when the user hasn't entered anything.

Then for issue 2, we probably want to validate required fields to not accept empty strings.

@maccomaccomaccomacco does that make sense for issue 2?

oiorain:
For Issue 1:

Create an entry of a CT with a required field
Fill in content other than the required field
Save (this works fine)
Publish - on the required field, see the error message this field is required

Confirmed: ✅ 
Screenshot 2024-09-17 at 16 46 04


For Issue 2:

Save and publish content with the required field filled - this works as expected
From the published entry clear the required field
Directly publish - should see the error message this field is required again.

Confirmed ✅ 

Screenshot 2024-09-17 at 16 47 30

LGTM!

Bassel17:
> I think required fields shouldn't be required to save, only on publish.
> 
> So for issue 1, to make the error "less technical" we could send an empty string for text fields instead of a `null` when the user hasn't entered anything.
> 
> Then for issue 2, we probably want to validate required fields to not accept empty strings (might also want to .trim() it so only whitespace is also considered empty?)
> 
> @maccomaccomaccomacco does that make sense for issue 2?

the PR seems to fix the issues mentioned, and am still able to save required fields in Draft, not sure how removing this line of code fixed all this though, it might need a little more investigation

innerdvations:
> > I think required fields shouldn't be required to save, only on publish.
> > So for issue 1, to make the error "less technical" we could send an empty string for text fields instead of a `null` when the user hasn't entered anything.
> > Then for issue 2, we probably want to validate required fields to not accept empty strings (might also want to .trim() it so only whitespace is also considered empty?)
> > @maccomaccomaccomacco does that make sense for issue 2?
> 
> the PR seems to fix the issues mentioned, and am still able to save required fields in Draft, not sure how removing this line of code fixed all this though, it might need a little more investigation

That's really weird, in my test it's not working right 🤔 I'll keep testing and see if I can see what the difference is and report back.

Bassel17:
I was testing with a simple required string field, didn't check other required stuff like components or dz , so am not sure what's happening there if thats what you are testing

innerdvations (commented):
Ok, did some further testing and now it works, maybe it was just something wrong on my end.

LGTM! Did you find out exactly why this fixes it? Might be worth adding a comment about it, and an e2e test to prevent a regression

butcherZ:
> Ok, did some further testing and now it works, maybe it was just something wrong on my end.
> 
> LGTM! Did you find out exactly why this fixes it? Might be worth adding a comment about it, and an e2e test to prevent a regression

I'm not so sure, that's why I added JS as reviewer as well, it seems like he originally wrote those lines to fix an e2e test issue (I can't tell which one)and I'm hoping removing them is not going to break anything else :joy:

I will dig a bit more into it and add the test for it :)

innerdvations:
Closing in favor of https://github.com/strapi/strapi/pull/21314

PR #21313, opened 2024-09-17
fix: history fields width

remidej:
### What does it do?

Prevents content history fields from shrinking. It copied the method from the main edit view

Instead of this:

CleanShot 2024-09-17 at 17 43 31@2x

We want this:
CleanShot 2024-09-17 at 17 57 52@2x

I also removed the shadow from the relation boxes that we don't want

### Why is it needed?

### How to test it?

Open history, check and check normal fields, fields in components, and fields in dynamic zones

### DIFF
--- a/packages/core/content-manager/server/src/history/services/history.ts
+++ b/packages/core/content-manager/server/src/history/services/history.ts
@@ -64,6 +64,7 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
         const entryWithRelations = await Object.entries(entry.schema).reduce(
           async (currentDataWithRelations, [attributeKey, attributeSchema]) => {
             const attributeValue = entry.data[attributeKey];
+
             const attributeValues = Array.isArray(attributeValue)
               ? attributeValue
               : [attributeValue];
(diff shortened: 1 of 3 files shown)


Feranchz (approved):

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -105,9 +106,14 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-                      .query('admin::user')
-                      .findOne({ where: { id: userToPopulate.id } });
+                    return strapi.query('admin::user').findOne({
+                      where: {
+                        ...(userToPopulate.id ? { id: userToPopulate.id } : {}),
+                        ...(userToPopulate.documentId
+                          ? { documentId: userToPopulate.documentId }
+                          : {}),
+                      },
+                    });
```
remidej, line 116: @alexandrebodin I don't mind the change but I need some context:
- what's the bug you were seeing and how can I test it?
- are admin users supposed to just have id, or just document id, or both? Why is it not the same every time?

alexandrebodin: The history code is buggy. It's only storing in db the documentId, locale and published information and not the id. So When this code is executed (create a relation to an admin user in a content type to test it) it doesn't have any id in the userToPopulate and just breaks. 

I'm just adding documentId and not removing id in case some data was saved with an id previsouly. I don't know when this was broken

alexandrebodin: Discussed live. UI fix incoming from Rémi 👍

alexandrebodin (approved):

PR #21314, opened 2024-09-17
fix: cm validation logic is broken

alexandrebodin:
### What does it do?

Fix https://github.com/strapi/strapi/issues/21285

More general fix based on https://github.com/strapi/strapi/pull/21312

### DIFF
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -345,7 +345,13 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
           .filter((el) => el.selected)
           .map((el) => el.value);
       } else {
-        val = value;
+        // NOTE: reset value to null so it failes required checks.
+        // The API only considers a required field invalid if the value is null|undefined, to differentiate from min 1
+        if (value === '') {
+          val = null;
+        } else {
+          val = value;
+        }
       }
 
       if (field) {
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -53,7 +53,10 @@ interface FormContextValue<TFormValues extends FormValues = FormValues>
   setErrors: (errors: FormErrors<TFormValues>) => void;
   setSubmitting: (isSubmitting: boolean) => void;
   setValues: (values: TFormValues) => void;
-  validate: () => Promise<
+  validate: (
+    shouldSetErrors?: boolean,
+    options?: Record<string, string>
+  ) => Promise<
     { data: TFormValues; errors?: never } | { data?: never; errors: FormErrors<TFormValues> }
   >;
 }
@@ -132,6 +135,8 @@ interface FormProps<TFormValues extends FormValues = FormValues>
   // TODO: type the return value for a validation schema func from Yup.
   validationSchema?: Yup.AnySchema;
   initialErrors?: FormErrors<TFormValues>;
+  // NOTE: we don't know what return type it can be here
+  validate?: (values: TFormValues, options: Record<string, string>) => Promise<any>;
 }
 
 /**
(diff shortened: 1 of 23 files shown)


packages/core/content-manager/admin/src/utils/validation.ts
```diff
@@ -317,7 +317,7 @@ const addMinValidation: ValidationFn =
         (attribute.type === 'component' && attribute.repeatable) ||
         attribute.type === 'dynamiczone'
       ) {
-        if (options.status !== 'draft' && !attribute.required && 'test' in schema && min) {
+        if (options.status === 'draft' && !attribute.required && 'test' in schema && min) {
```
jhoward1994, line 320: I'm a bit confused by this, are we saying minimum validations of repeatable components and dzs happen on drafts too? 

If so should we remove the status check all together so it runs on both draft and publish?

alexandrebodin: There is the empty array condition that we need only for drafts. I have another issue to fix still though so I'll change this again

- null/undefined => ok -> Value when creating
- empty array => ok -> Value after creation returned by the api

innerdvations (approved):
Very nice compromise between the minimal fix and ripping the whole thing out and replacing it 😆

Tested and works for me, although we definitely need more QA from more people in case I've missed something critical.

butcherZ (approved):
Just did QA on my side, LGTM! 😄

alexandrebodin:
Experimental `0.0.0-experimental.0faf8967355707d2684ecb8579b03979ad5df042`

innerdvations:
This has caused a lot of e2e tests to fail, but it could actually be that they were written incorrectly to get around the way the validation worked previously

jhoward1994 (commented):
I think this is intended but just in case, max & maxLength validations are still run on draft entries

jhoward1994:
Uploading rep-bug.mov…


The delete buttons on repeatable component accordion headers are still enabled in the CM edit view published tab

I would call this minor as you can't actually perform any action from that tab but I think you shouldn't be able to modify the form in this way

simotae14 (approved):
I tried the experimental and it fixes my issue, thanks Alex

packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -333,7 +345,11 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
           .filter((el) => el.selected)
           .map((el) => el.value);
       } else {
-        val = value;
+        if (value === '') {
+          val = null;
+        } else {
+          val = value;
+        }
```
jhoward1994, line 352: What case was this a fix for? Maybe we could add a comment explaining it?

jhoward1994 (approved):

packages/core/admin/admin/src/components/Form.tsx
alexandrebodin: when you empty a field it needs to be set to null to be considered "cleared" instead of just an empty string. For the API it's an important difference

alexandrebodin: I'll add a note

jhoward1994:
Should this be nullable as it has `required` ?

https://github.com/strapi/strapi/blob/2ee3865f9d96228baa4c49eb4f2fb14332d0e2ea/packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/constants.ts#L6-L10

alexandrebodin:
> Should this be nullable as it has `required` ?
> 
> https://github.com/strapi/strapi/blob/2ee3865f9d96228baa4c49eb4f2fb14332d0e2ea/packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/constants.ts#L6-L10

Since it's a select it can't be set to an empty string so it's not necessary but I can if yu prefer

PR #21329, opened 2024-09-18
fix(create-strapi): yarn 4 support

innerdvations:
### What does it do?

adds support for env vars and version ranges for both env vars and command line args for each package manager

### Why is it needed?

yarn 4 needs to use env YARN_HTTP_TIMEOUT because it has removed --network-timeout as a cli option and it was failing with an error when it tried to run `yarn install --network-timeout 1000000`

and I didn't feel like adding yet another conditional for a specific case, so I made it super general 😆 since I'm sure there will be more things like this necessary in the future

### How to test it?

create-strapi-app should work with yarn 4 as well as previous versions of yarn, and still work with `pnpm` and `npm`

### Related issue(s)/PR(s)

DX-1611
https://github.com/strapi/strapi/pull/21271


packages/cli/create-strapi-app/src/utils/get-package-manager-args.ts
```diff
@@ -0,0 +1,126 @@
...
+
+// Set environment variables for specific package managers, with full semver ranges
+const installEnvMap: {
+  [key: string]: VersionedEnvMap;
+} = {
+  yarn: {
+    '>=4': { YARN_NETWORK_TIMEOUT: '1000000' },
+    '*': {},
+  },
+  npm: {
```
alexandrebodin, line 38: Looks like these could be omitted

alexandrebodin (approved):
I won't have time to QA today. the code LGTM if you can make an experimental for testing that would help 👍

Bassel17:
experimental: `0.0.0-experimental.117b0ba4c41638592858ef7c29fec09396cf01be`

Bassel17 (changes requested):
encountered a different error while using the experimental for this 

image

innerdvations:
Oops, which version of yarn is that with? I might have broken something in one of my last commits

Bassel17:
> Oops, which version of yarn is that with? I might have broken something in one of my last commits

this is `Yarn 4.5.0`

innerdvations:
Yeah, my bad, I must have done something last minute and broke the one thing this was attempting to fix 😆

innerdvations:
Generating another experimental, the command run by execa for `yarn version`  and `yarn install` were running in different locations with differnt env options, which was causing version mismatches where it thought the system installed pm version was the same version that would run the install, but it wasn't necessarily

innerdvations:
Ok, finally got it I think: 0.0.0-experimental.d3243594aea3e6fa0ef08580192bb0df29c9162d

Bassel17 (approved):
LGTM ! tested with the latest experimental seems to be working now

PR #21330, opened 2024-09-18
fix: use strapi v5 for template

innerdvations:
### What does it do?

website template uses strapi 5

### Why is it needed?

it's using strapi rc

### How to test it?

template should be generated with correct dependendies

### Related issue(s)/PR(s)

DX-1620

### DIFF
--- a/templates/website/package.json
+++ b/templates/website/package.json
@@ -9,10 +9,10 @@
     "deploy": "strapi deploy"
   },
   "dependencies": {
-    "@strapi/plugin-cloud": "^5.0.0-rc",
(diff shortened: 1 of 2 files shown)


alexandrebodin (approved):

innerdvations:
@alexandrebodin does the release script that updates version numbers also target this from now on? Or is it something we need to add?

alexandrebodin:
@innerdvations no it doesn't. templates are independent intentionnaly

However the script auto updates the package.json automatically 

for all the deps in defined in packages/cli/create-strapi-app/src/index.ts

jhoward1994 (approved):

PR #21373, opened 2024-09-19
fix: add importName to helper-plugin codemods

Bassel17:
### What does it do?

- Add a `newImport` to helper plugin-codemods to distinguish  between how a component is imported and how it's used

### Why is it needed?

- We have issues with some of the transforms examples:
transforming to this `import { Page.Protect } from "@strapi/strapi/admin";`
 instead of this `import { Page } from "@strapi/strapi/admin";`

### How to test it?
 run codemods against 
 `import { CheckPagePermissions } from '@strapi/helper-plugin` should become `import { Page } from "@strapi/strapi/admin";`


innerdvations (dismissed):

innerdvations (approved):

jhoward1994 (approved):
Thanks 🙏🏻

PR #18010, opened 2023-09-11
Find database migrations dir in typescript project

DevDengChao:
### What does it do?

Recongnize /dist/database/migrations in ts project

### Why is it needed?

https://github.com/strapi/strapi/issues/17820

### How to test it?

+ There are unit tests.
+ Or you can create an empty strapi project and write some typescript migrations.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/17820

Issue #17820, opened 2023-08-26
use migrations dont work with file typescript


tuan-zigvy:
if i write migrations by file ts , it not working so i convert to js  it work

PR #21374, opened 2024-09-19
enhancement: add support for user database migrations in typescript

DevDengChao:
### What does it do?

Recongnize /dist/database/migrations in ts project

### Why is it needed?

https://github.com/strapi/strapi/issues/17820

### How to test it?

Setup a typescript project with ts migrations, verify those migrations are executed.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/18010


hanpaine:
Hi @DevDengChao thanks for your contribution and patience! We'll take a look soon 🙏

packages/core/core/src/Strapi.ts
```diff
@@ -269,20 +270,20 @@ class Strapi extends Container implements Core.Strapi {
...
-          new Database(
-            _.merge(this.config.get('database'), {
-              logger,
-              settings: {
-                migrations: {
-                  dir: path.join(this.dirs.app.root, 'database/migrations'),
-                },
+      .add('db', () => {
+        const outDir = tsUtils.resolveOutDirSync(this.dirs.app.root) as string;
+        const root = outDir === undefined ? this.dirs.app.root : outDir;
```
Convly, line 275:
```suggestion
        const outDir = tsUtils.resolveOutDirSync(this.dirs.app.root) ?? this.dirs.app.root;
```

Convly (changes requested):
Hi, thanks for opening this PR, I've released an experimental version to test it.

I only have one small comment regarding the code itself.

:test_tube: `0.0.0-experimental.62ebf57bb35846666febb5ee2078fe02fe990f57`

packages/core/core/src/Strapi.ts
```diff
@@ -269,20 +270,19 @@ class Strapi extends Container implements Core.Strapi {
...
-                  dir: path.join(this.dirs.app.root, 'database/migrations'),
-                },
+      .add('db', () => {
+        const outDir = tsUtils.resolveOutDirSync(this.dirs.app.root) ?? this.dirs.app.root;
+        return new Database(
+          _.merge(this.config.get('database'), {
+            logger,
+            settings: {
+              migrations: {
+                dir: path.join(root, 'database/migrations'),
```
Convly, line 280: Did you mean `outDir` instead? :thinking:

Convly: This is breaking the build

Convly (changes requested):

packages/core/core/src/Strapi.ts
DevDengChao: Fixed.

Convly (approved):

innerdvations:
This seems to prevent .js migrations from being found by typescript projects, we need to be able to support both, especially since someone may already have .js files in their migrations prior to having this feature

Convly:
Also, unsure why the e2e are failing on this one :thinking:

innerdvations:
> Also, unsure why the e2e are failing on this one 🤔

I think it's a tooling issue, I deleted the artifacts and I think it'll pass next time they run.

hanpaine:
@innerdvations they did not pass... 😬

innerdvations:
They haven't run again yet. If it still happens when a fix is pushed for .js migrations in typescript projects I'll look into it, but I suspect it'll work

DevDengChao:
Maybe just [set "allowJs" to "true"](https://www.typescriptlang.org/tsconfig/#allowJs) ?

packages/core/core/src/Strapi.ts
```diff
@@ -269,20 +270,19 @@ class Strapi extends Container implements Core.Strapi {
...
-                  dir: path.join(this.dirs.app.root, 'database/migrations'),
-                },
+      .add('db', () => {
+        const outDir = tsUtils.resolveOutDirSync(this.dirs.app.root) ?? this.dirs.app.root;
+        return new Database(
+          _.merge(this.config.get('database'), {
+            logger,
+            settings: {
+              migrations: {
+                dir: path.join(outDir, 'database/migrations'),
```
innerdvations, line 280:
```suggestion
        const tsDir = tsUtils.resolveOutDirSync(this.dirs.app.root);
        const projectDir = this.config.get('database.useTypescriptMigrations') === true && tsDir ? tsDir : this.dirs.app.root;
        return new Database(
          _.merge(this.config.get('database'), {
            logger,
            settings: {
              migrations: {
                dir: path.join(projectDir, 'database/migrations'),
```

Something like this might be the best option for now to avoid breaking anything (note: I didn't test it, just wrote it here quickly to show the idea)

As a follow-up, we will start generating new typescript projects with `database.useTypescriptMigrations:true` already set, and in Strapi 6 we can make it the default behaviour.

Let's get some opinions on if this is the best place to put it, and then we will need to document it.

cc @remidej @Marc-Roig @butcherZ

innerdvations:
> @innerdvations they did not pass... 😬

It's because of the playwright update. If that part is reverted it should work @DevDengChao 

If playwright needs to be updated for some reason for this PR to work, I've got a PR here that includes a playwright upgrade that also fixes the CI: https://github.com/strapi/strapi/pull/22033/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fL198-L257

so after that's merged this PR could be updated with it and everything should be fine

packages/core/core/src/Strapi.ts
innerdvations: Also tagging @alexandrebodin in case you have any comment on it

Database config seems the right place for it, my only concern is that we've never put an option other than `connection` there before afaik.

DevDengChao:
@innerdvations This PR doesn't require the update of playwright, let me revert it tomorrow morning (+8).

packages/core/core/src/Strapi.ts
innerdvations: @DevDengChao could you please have it only use the new behaviour when config `database.settings.useTypescriptMigrations` is true? I'll create a docs PR for it

innerdvations:
Documentation: https://github.com/strapi/documentation/pull/2283

hanpaine:
Hey @DevDengChao did you have a chance to look at this yet? Thanks 🙏

DevDengChao:
@innerdvations @hanpaine Updated.

innerdvations:
Looks like there is a linting issue to fix, but other than that this works for me! As soon as it's fixed we can merge

DevDengChao:
@innerdvations Updated, but I'm not sure if that fixs the lint issue because the `lint:fix` did nothing.

innerdvations (approved):

DevDengChao:
🎉

innerdvations:
For the future, there's also a `yarn prettier:write` script that may help.

Thank you so much for your contribution!

PR #21376, opened 2024-09-19
enhancement: add en-GB and en-US locales

alexandrebodin:
### What does it do?

Add en-GB support and allows js lang files for dynamic overrides


Christian-Schwarz2003:
Nice, that solution is way neater, also for the future if there are words added that actually differ between BE and AE.

remidej (approved):
🚀

alexandrebodin:
Thank you @Christian-Schwarz2003 This will be included in the next release 🚀

PR #21380, opened 2024-09-19
feat(review-workflows): adding required stage for publishing

Feranchz:
### What does it do?

Adds "Required stage for publishing" to the Review Workflow settings page. This PR **don't** add publishing validation yet. Only is responsible to add the new field to Workflows in both backend/frontend.

### How to test it?

Go to an existing workflow or create a new one.
1. Try to set one of the stages as the required one for publishing
2. It should keep the value without problems

### DIFF
--- a/packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
@@ -82,7 +82,7 @@ const Stages = ({ canDelete = true, canUpdate = true, isCreating }: StagesProps)
                   canReorder={stages.length > 1}
                   canUpdate={canUpdate}
                   stagesCount={stages.length}
-                  defaultOpen={isCreating}
+                  defaultOpen={!stage.id}
                   {...stage}
                 />
               </Box>
--- a/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
@@ -216,10 +216,13 @@ const StageSelector = ({ disabled }: StageSelectorProps) => {
 
   const { error, value, onChange } = useField('stageRequiredToPublish');
 
+  // stages with empty names are not valid, so we avoid them from being used to avoid errors
+  const validStages = stages.filter((stage) => stage.name);
+
   return (
     <Field.Root
       error={error}
-      name={'stageRequiredToPublish'}
+      name="stageRequiredToPublish"
       hint={formatMessage({
         id: 'settings.review-workflows.workflow.stageRequiredToPublish.hint',
         defaultMessage:
@@ -245,8 +248,11 @@ const StageSelector = ({ disabled }: StageSelectorProps) => {
             defaultMessage: 'Any stage',
           })}
         </SingleSelectOption>
-        {stages.map((stage, i) => (
-          <SingleSelectOption key={`requiredForPublishStage-${i}`} value={stage.name}>
+        {validStages.map((stage, i) => (
+          <SingleSelectOption
+            key={`requiredToPublishStage-${stage.id || stage.__temp_key__}`}
+            value={stage.id?.toString() || stage.__temp_key__}
+          >
             {stage.name}
           </SingleSelectOption>
         ))}
--- a/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
@@ -214,33 +214,34 @@ const StageSelector = ({ disabled }: StageSelectorProps) => {
   const { value: stages = [] } = useField<WorkflowStage[]>('stages');
   const { formatMessage } = useIntl();
 
-  const { error, value, onChange } = useField('stageRequiredForPublish');
+  const { error, value, onChange } = useField('stageRequiredToPublish');
 
   return (
     <Field.Root
       error={error}
-      name={'stageRequiredForPublish'}
+      name={'stageRequiredToPublish'}
       hint={formatMessage({
-        id: 'settings.review-workflows.workflow.stageRequiredForPublish.hint',
-        defaultMessage: 'Prevents entries to be published if they are not at the required stage.',
+        id: 'settings.review-workflows.workflow.stageRequiredToPublish.hint',
+        defaultMessage:
+          'Prevents entries from being published if they are not at the required stage.',
       })}
     >
       <Field.Label>
         {formatMessage({
-          id: 'settings.review-workflows.workflow.stageRequiredForPublish.label',
+          id: 'settings.review-workflows.workflow.stageRequiredToPublish.label',
           defaultMessage: 'Required stage for publishing',
         })}
       </Field.Label>
       <SingleSelect
         disabled={disabled}
         onChange={(value) => {
-          onChange('stageRequiredForPublish', value);
+          onChange('stageRequiredToPublish', value);
         }}
         value={value}
       >
         <SingleSelectOption value={''}>
           {formatMessage({
-            id: 'settings.review-workflows.workflow.stageRequiredForPublish.any',
+            id: 'settings.review-workflows.workflow.stageRequiredToPublish.any',
             defaultMessage: 'Any stage',
           })}
         </SingleSelectOption>
--- a/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
@@ -7,12 +7,16 @@ import {
   MultiSelectOption,
   Typography,
   useCollator,
+  SingleSelect,
+  SingleSelectOption,
 } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
 import { styled } from 'styled-components';
 
 import { useGetContentTypesQuery } from '../../../services/content-manager';
 import { useReviewWorkflows } from '../hooks/useReviewWorkflows';
+
+import type { WorkflowStage } from './Stages';
 
 /* -------------------------------------------------------------------------------------------------
  * WorkflowAttributes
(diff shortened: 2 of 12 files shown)


packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
```diff
@@ -82,7 +82,7 @@ const Stages = ({ canDelete = true, canUpdate = true, isCreating }: StagesProps)
                   canReorder={stages.length > 1}
                   canUpdate={canUpdate}
                   stagesCount={stages.length}
-                  defaultOpen={isCreating}
+                  defaultOpen={!stage.id}
```
remidej, line 85: Can you explain this change?

packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
```diff
@@ -196,5 +203,57 @@ const ContentTypeTakeNotice = styled(Typography)`
...
+
+  const { error, value, onChange } = useField('stageRequiredForPublish');
+
+  return (
+    <Field.Root
+      error={error}
+      name={'stageRequiredForPublish'}
+      hint={formatMessage({
+        id: 'settings.review-workflows.workflow.stageRequiredForPublish.hint',
+        defaultMessage: 'Prevents entries to be published if they are not at the required stage.',
```
remidej, line 225: I think it should be "from being" instead of "to be"? Could a native speaker check? @hanpaine @cache-your-dreams

packages/core/review-workflows/admin/src/routes/settings/hooks/useReviewWorkflows.ts
```diff
@@ -31,7 +31,7 @@ const useReviewWorkflows = (params: UseReviewWorkflowsArgs = {}) => {
 
   const { data, isLoading, error } = useGetWorkflowsQuery(
     {
-      populate: 'stages',
+      populate: ['stages', 'stageRequiredForPublish'],
```
remidej, line 34: Again just wording, but I think grammatically it makes more sense to have "stageRequiredToPublish"

packages/core/review-workflows/server/src/services/workflows.ts
```diff
@@ -87,6 +87,17 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 
         createOpts = set('data.stages', mapIds(stages), createOpts);
 
+        if (opts.data.stageRequiredForPublish) {
+          const stageRequiredForPublish = stages.find(
+            (stage: any) => stage.name === opts.data.stageRequiredForPublish
+          );
+          if (!stageRequiredForPublish) {
+            throw new errors.ApplicationError('Stage required for publish does not exist');
```
remidej, line 95: Grammatically it should be "stage required to publish" or "stage required for publication"

packages/core/review-workflows/server/src/services/workflows.ts
```diff
@@ -128,11 +140,36 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 
           updatedStageIds = await stageService
             .replaceStages(workflow.stages, opts.data.stages, workflow.contentTypes)
-            .then((stages: any) => stages.map((stage: any) => stage.id));
+            .then((stages: any) => {
+              updatedStages = stages;
+              return stages.map((stage: any) => stage.id);
+            });
```
remidej, line 146: I find the mix of functional and non-functional code a bit weird here (since you mutate updated stages). I would recommend doing it in two steps instead:

- `const updatedStages = await stageService.replaceStages(...)`
- `const updatedStagesIds = updatedStages.map(stage => stage.id)`

packages/core/review-workflows/shared/contracts/review-workflows.ts
```diff
@@ -115,7 +116,9 @@ namespace GetAll {
 namespace Update {
   export interface Request {
     body: {
-      data: Partial<Workflow>;
+      data: Partial<Omit<Workflow, 'stageRequiredForPublish'>> & {
+        stageRequiredForPublish?: Stage['name'] | null;
+      };
     };
```
remidej, line 122: Just to clarify, how does this differ from the original `Partial<Workflow>`?

packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
```diff
@@ -196,5 +203,57 @@ const ContentTypeTakeNotice = styled(Typography)`
...
+        onChange={(value) => {
+          onChange('stageRequiredForPublish', value);
+        }}
+        value={value}
+      >
+        <SingleSelectOption value={''}>
+          {formatMessage({
+            id: 'settings.review-workflows.workflow.stageRequiredForPublish.any',
+            defaultMessage: 'Any stage',
+          })}
```
remidej, line 245: Have you checked if a normal stage could also have an empty string as a name?

packages/core/review-workflows/shared/contracts/review-workflows.ts
```diff
@@ -133,7 +136,11 @@ namespace Update {
 namespace Create {
   export interface Request {
     body: {
-      data: Omit<Workflow, 'id' | 'createdAt' | 'updatedAt'>;
+      data: Omit<Workflow, 'id' | 'createdAt' | 'updatedAt'> & {
+        // We cannot use id because the stage could be not created yet
+        // And considering name is unique inside the workflow, name is a better option
+        stageRequiredForPublish?: Stage['name'] | null;
```
remidej, line 142: It wouldn't change anything here because it's a string too, but I would find it more explicit if you added an empty string in this union: it would explain what you use to store the "any" value, since it's not exactly a stage name

remidej (commented):
Beautiful code! and thanks for including the e2e test

packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
Feranchz: I was expecting that every time you click at "Add new stage" the accordion is opened in the new item. That is what I thought when using it because I don't see many sense on opening when creating a new stage **only** when is a new workflow but if you are editing and you add a new stage is not opened. (This also simplified a lot the e2e test)

But I would check with Lucas about this 🤔

packages/core/review-workflows/shared/contracts/review-workflows.ts
Feranchz: On Workflow interface, stageRequiredForPublish is of type Stage, this is okay because we return that and this field is a relation. 

But for Create/Update endpoins, we don't receive the complete Stage, only the name, so I omit that type and add this stageRequiredForPublish that is only name

packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
Feranchz: Is not possible, at least at backend we have a validation check for that. While testing I discovered an error when leaving the input completely empty but not sure if it's only on this PR or is in v5. I will check and in case if on v5 I will create a PR only with that to fix it quickly 👍

packages/core/review-workflows/shared/contracts/review-workflows.ts
Feranchz: The thing is that I don't want users to send empty string when "any" value is selected. I want them to send null and on frontend I catch that case to replace "" for null (only because SingleSelect component doesn't accept null as value of one option)

packages/core/review-workflows/admin/src/routes/settings/components/WorkflowAttributes.tsx
```diff
@@ -196,5 +203,58 @@ const ContentTypeTakeNotice = styled(Typography)`
...
+const StageSelector = ({ disabled }: StageSelectorProps) => {
+  const { value: stages = [] } = useField<WorkflowStage[]>('stages');
+  const { formatMessage } = useIntl();
+
+  const { error, value, onChange } = useField('stageRequiredToPublish');
+
+  return (
+    <Field.Root
+      error={error}
+      name={'stageRequiredToPublish'}
```
simotae14, line 222:
```suggestion
      name='stageRequiredToPublish'
```

packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
remidej: That sounds great to me 👍

Feranchz:
@remidej @simotae14 I made some changes following a bug I discovered when changing the name of one stage selected as required to publish. 

Also changed names of some variables to make it more clear, I prefer RW endpoint to receive `stageRequiredToPublishName` so is more clear we are sending the name and not the stage or its id (I'm sending name because depending on the case stage maybe is not created yet so it doesn't have an id)

packages/core/review-workflows/admin/src/translations/en.json
```diff
@@ -4,5 +4,8 @@
   "settings.page.isLoading": "Workflow is loading",
   "settings.page.delete.confirm.body": "All entries assigned to deleted stages will be moved to the previous stage. Are you sure you want to save?",
   "settings.stage.name.label": "Stage name",
-  "settings.not-available": "Review Workflows is only available as part of the Enterprise Edition. Upgrade to create and manage workflows."
+  "settings.not-available": "Review Workflows is only available as part of the Enterprise Edition. Upgrade to create and manage workflows.",
+  "settings.review-workflows.workflow.stageRequiredToPublish.label": "Required stage for publishing",
+  "settings.review-workflows.workflow.stageRequiredToPublish.any": "Any stage",
+  "settings.review-workflows.workflow.stageRequiredToPublish.hint": "Prevents entries to be published if they are not at the required stage."
```
remidej, line 10: This translation should be updated too

remidej:
I doesn't seem to actually save:
https://github.com/user-attachments/assets/89a9ff83-56e6-4b33-8fba-67b0db0e6ee8

remidej:
The app also crashes when I clear the name field, whether it's a new stage or an existing one


https://github.com/user-attachments/assets/0b7ecbd4-69e2-4cfe-9c24-fa4825483181

remidej (approved):
All good 👌

PR #21386, opened 2024-09-19
chore: update yarn to 4.5.0

smorimoto:
### What does it do?

As the title says.

### Why is it needed?

There are some notable fixes.

### How to test it?

CI.

### Related issue(s)/PR(s)

None.


alexandrebodin:
Thanks for the update @smorimoto

Issue #21355, opened 2024-09-18
REST API doesn't correctly return relations that are not inversed


laurenskling:
### Required System information

- Node.js version: 20
- NPM version: yarn 1.22
- Strapi version: 5.0.0 (and rc 17-29 at least)
- Database: sqlite
- Operating system: macos
- Is your project Javascript or Typescript: js

### Describe the bug + Steps to reproduce the behavior

I have a Homepage singleType. It doesnt have draftAndPublish, because the homepage never goes unpublished (v4 logic of DP). 

Now say i have a form contentType i want to relate. I pick it in the UI. I set all permission. I go to:
`http://0.0.0.0:1337/api/homepage/homepage?&populate[form][fields]=*` (/homepage/homepage is my custom endpoint), i see `form: null`.
if I then request: `http://0.0.0.0:1337/api/homepage/homepage?&populate[form][fields]=*&status=draft`, my form is returned. This is weird. I believe that without setting status, it should always work. I think the system assumes it will bubble down from the top contentType (in this case homepage), but that doesn't have a status.

### Expected behavior

A non-DP contentType should always return all of the requested relations.

### Status

- we found it has to do something with relations without `mappedBy` and `inversedBy` fields
- probably nothing to do with non-DP

alexandrebodin:
Hi, 

Could you provide the content-type schema.json for reproduction ?

I'm sorry but you lost me on the component explanation. Do you mind recording a loom or writing a list of specific steps to reproduce 🙏  ? 

A few infos

- `production` is not a valid status it's either `draft` or `published`
- You should see the published versions of your relations (when you target a D&P content-type) from your homepage when you populate. We are setting the status as published by default unless you manually overwrite it (might be where there is a bug)

laurenskling:
Yeah im sorry about the component explanation. It's very specific. It took me a while to understand what was happening. I'll try to make a repro a.s.a.p.

turns out, every &status=x just works (probably draft). So I can change my story to just `draft` returning my relation, whilst nothing or `published` doesn't.

laurenskling:
Okay I got it about the direct relation being `null`. (and probably also why components return `null`).

I've created a one way relation without `mappedBy` and `inversedBy` on the other side.

Steps:
1. npx create-strapi my-strapi-project with the test data installed
2. in the article schema.json, remove the author block
3. in the author schema.json, remove the `mappedBy`
4. start the app
5. make articles
6. make an author and relate some articles
7. set permissions to public rest endpoints
[...]

PR #21390, opened 2024-09-19
fix: proof of concept for the fix

alexandrebodin:
### What does it do?

Fix attempt for https://github.com/strapi/strapi/issues/21355

### Issues

- Frontend sending UI metadata to the API (id and status instead of documentId)
- When creating a unidirectional relation, only the published or draft relation were selected instead of both for Non D&P sources

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -643,7 +643,7 @@ const PublishAction: DocumentActionComponent = ({
           documentId,
           params,
         },
-        formValues
+        formDocumentToData(formValues)
       );
 
       if ('data' in res && collectionType !== SINGLE_TYPES) {
@@ -733,6 +733,40 @@ const PublishAction: DocumentActionComponent = ({
 
 PublishAction.type = 'publish';
 
+const formDocumentToData = (document: any) => {
+  return Object.keys(document).reduce((acc, key) => {
+    if (document[key] !== undefined) {
+      acc[key] = document[key];
+    }
+
+    if (typeof document[key] === 'object' && document[key] !== null) {
+      if ('connect' in document[key] && document[key].connect !== null) {
+        acc[key] = {
+          ...acc[key],
+          connect: document[key].connect.map((item: any) => {
+            return {
+              documentId: item.documentId,
+            };
+          }),
+        };
+      }
+
+      if ('disconnect' in document[key] && document[key].disconnect !== null) {
+        acc[key] = {
+          ...acc[key],
+          disconnect: document[key].disconnect.map((item: any) => {
+            return {
+              documentId: item.documentId,
+            };
+          }),
+        };
+      }
+    }
+
+    return acc;
+  }, {} as any);
+};
+
 const UpdateAction: DocumentActionComponent = ({
   activeTab,
   documentId,
@@ -797,7 +831,7 @@ const UpdateAction: DocumentActionComponent = ({
               documentId: cloneMatch.params.origin!,
               params,
             },
-            document
+            formDocumentToData(document)
           );
 
           if ('data' in res) {
@@ -823,7 +857,7 @@ const UpdateAction: DocumentActionComponent = ({
               documentId,
               params,
             },
-            document
+            formDocumentToData(document)
           );
 
           if (
@@ -841,7 +875,7 @@ const UpdateAction: DocumentActionComponent = ({
               model,
               params,
             },
-            document
+            formDocumentToData(document)
           );
 
           if ('data' in res && collectionType !== SINGLE_TYPES) {
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -84,7 +84,7 @@ function useHandleDisconnect(fieldName: string, consumerName: string) {
       }
     }
 
-    addFieldRow(`${fieldName}.disconnect`, { id: relation.id });
+    addFieldRow(`${fieldName}.disconnect`, { id: relation.id, documentId: relation.documentId });
   };
 
   return handleDisconnect;
@@ -259,6 +259,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
 
       const item = {
         id: relation.id,
+        documentId: relation.documentId,
         status: relation.status,
         /**
          * If there's a last item, that's the first key we use to generate out next one.
--- a/packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-continue */
-import { keyBy } from 'lodash/fp';
+import { omit, keyBy } from 'lodash/fp';
 
 import { UID, Schema } from '@strapi/types';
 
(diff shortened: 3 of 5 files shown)


oiorain:
Before i start testing, here's a test plan. Do you see other cases to add @alexandrebodin ?
cc @MarionLemaire Can you give me a hand with testing, please?

# Test cases

- [ ] Regular entry to regular entry 
- [ ] D&P entry to regular entry 
- [ ] regular entry to D&P entry 
- [ ] Regular entry to i18n 
- [ ] i18n to regular entry 
- [ ] regular entry to d&p + i18n 
- [ ] d&p + i18n to regular 
- [ ] D&p to d&p + i18n 
- [ ] D&p + i18n  to d&p
- [ ] D&p + i18n  to d&p + i18n

# Possible tests to perform in each case:
## Many to many relation
- [ ] Add a Many to many relation
- [ ] Add it from left side + check presence
- [ ] Remove it from left side + check presence
- [ ] Add it from right side + check presence
- [ ] Remove it from right side + check presence
**When D&P is enabled**
- [ ] Test publishing left side
- [ ] Test publishing right side 
- [ ] Test editing the relationship on one side then the other
- [ ] Test removing the relationship on one side then the other
- [ ] test unpublishing left side
- [ ] test unpublishing right side
**When i18n is enabled** 
- [ ] Test to enable i18n on a content type containing a relation on left side then right side
- [ ] Test editing the relationship or entry
- [ ] Test to disable i18n 
**When D&P and i18n is enabled**
- [ ] Test to publish a content with i18n on one side then right side
- [ ] Test to unpublish a content with i18n on one side then right side

## Possible test cases for a one way relation
- [ ] Add a Many way relation + check presence
- [ ] Remove it + check presence
**When D&P is enabled**
- [ ] Test publishing
- [ ] Test editing the relationship
- [ ] Test removing the relationship
- [ ] test unpublishing
**When i18n is enabled** 
- [ ] Test to enable i18n on a content type containing a relation
- [ ] Test editing the relationship or entry
- [ ] Test to disable i18n 
**When D&P and i18n is enabled**
- [ ] Test to publish a content with i18n
- [ ] Test to unpublish a content with i18n

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
```diff
@@ -259,6 +259,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
 
       const item = {
         id: relation.id,
+        documentId: relation.documentId,
```
remidej, line 262: Is the normal ID still being used?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -733,6 +733,40 @@ const PublishAction: DocumentActionComponent = ({
 
 PublishAction.type = 'publish';
 
+const formDocumentToData = (document: any) => {
+  return Object.keys(document).reduce((acc, key) => {
```
remidej, line 737: Don't you need to apply the logic for relations inside components too? You may need to traverse the document tree. There's traverseAndExtract in this file that does something similar

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -12,7 +12,11 @@ import { UID, Schema } from '@strapi/types';
  * @param oldEntries The old entries that are being overridden
  * @returns An array of relations that need to be updated with the join table reference.
  */
-const load = async (uid: UID.ContentType, oldEntries: { id: string; locale: string }[]) => {
+const load = async (
+  uid: UID.ContentType,
+  oldEntries: { id: string; locale: string }[],
+  draftEntries: { id: string; locale: string }[]
```
remidej, line 18: I think we should either add this to the JSdoc types above or remove them entirely

remidej (commented):
thanks for taking this one! fyi Marc is out until wednesday, so we may want to  get a review by another backend engineer

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
alexandrebodin: Yeah I'll redo the code once we validate the behavior ;)

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
alexandrebodin: Yes definitly needed for components

oiorain:
Small bug found: 
Article (D&P ) has one way link to Category. 
One article is published and linked to `Dogs` category -> `Dogs` doesnt appear in the list of the category field on the draft version of article.

oiorain:
@lucasboilly maybe you'll have an opinion on the i18n language being displayed in the dropdowb?

MarionLemaire:
I'll take for now (and check boxes directly in test plan when done) :

Test cases

 _Regular entry to regular entry
 D&P entry to regular entry
 regular entry to D&P entry_
 **Regular entry to i18n
 i18n to regular entry**
 _regular entry to d&p + i18n_
 _d&p + i18n to regular
 D&p to d&p + i18n_
 **D&p + i18n to d&p
 D&p + i18n to d&p + i18n**

Many to many relation

 **Add a Many to many relation
 Add it from left side + check presence
 Remove it from left side + check presence
 Add it from right side + check presence
 Remove it from right side + check presence**

oiorain:
So @alexandrebodin Good job! your fix is working 🎉  
Bonus: we QAed the whole relations ...

lucasboilly:
Hey @oiorain

It would be a great addition indeed, however we'll need to find a cleverer way to display the locales names or giving a max width to its container. If we don't, they could completely hide the entries names as they are pretty long by default and could be even longer if users change them in the i18n settings.

Since 2 entries with i18n can only be linked if they are in the same locale, it would be only useful to link an entry without i18n to entries with i18n right? Or is there other use cases?

Also, what would be even better is to group all locales from a same entry together, and separate them from the others with a divider. In this screenshot, if I don't know that "Boulangerie" is the french translation of "Bakery", I have no way to know that it's the same entry but in 2 different languages. 

CleanShot 2024-09-20 at 10 23 14@2x

lucasboilly:
@oiorain regarding the fact that we don't understand that there's more than 3 entries, it's because the lines are too big, but they are too big because the draft/published/modified badge is too big. I use a smaller version in the mockups but it wasn't developed yet. If we do it, the lines will have the same height as in v4 and the last line should go beyond the dropdown.

There's also the fact that the scrollbar is missing, but that's a general issue in our DS.

MarionLemaire:
Test plan completed, LGTM, +1 for @oiorain small bugs above

lucasboilly:
@oiorain sorry for spamming but it makes me think that it would be great to be able to bulk add relations. Some clients have 45+ locales, meaning that if they want to link an entry without i18n to all locales of another entry they have to do it 45+ times. To add relations, users have to make 2 clicks and 1 scroll in the popover, so 3 actions x 45 locales = 135 actions to perform (multiplied by the number of entries they want to link) 😅

oiorain:
> Since 2 entries with i18n can only be linked if they are in the same locale, it would be only useful to link an entry without i18n to entries with i18n right? Or is there other use cases? 

yes, this is the usecase. I agree it's probably not super common. 

> regarding the fact that we don't understand that there's more than 3 entries, it's because the lines are too big, but they are too big because the draft/published/modified badge is too big. I use a smaller version in the mockups but it wasn't developed yet. If we do it, the lines will have the same height as in v4 and the last line should go beyond the dropdown.

> There's also the fact that the scrollbar is missing, but that's a general issue in our DS.

Thank you for clarifying!

> @oiorain sorry for spamming but it makes me think that it would be great to be able to bulk add relations. Some clients have 45+ locales, meaning that if they want to link an entry without i18n to all locales of another entry they have to do it 45+ times. To add relations, users have to make 2 clicks and 1 scroll in the popover, so 3 actions x 45 locales = 135 actions to perform (multiplied by the number of entries they want to link) 😅

 Yeah.. there might be a better UX to navigate relations. but i think the broader usecase is having i18n activited on both content types. Having an option to propagate the value to all the other languages may be interesting. no idea if this is a big need to users.

lucasboilly:
> Having an option to propagate the value to all the other languages may be interesting. no idea if this is a big need to users.

Can't remember anyone asking for this, however I like the idea!

Issue #21391, opened 2024-09-20
Typo in default keycloak provider


Throttle:
### Required System information

- Node.js version: 18.17.1
- NPM version: 9.6.7
- Strapi version: 5.0.0
- Database: Postgres
- Operating system: macOS 14.6.1 (23G93)
- Is your project Javascript or Typescript: Typescript

### Describe the bug

There is a typo in the providers registry:
keycloack. Should be **keycloak**.

### Steps to reproduce the behavior

1. Enable Keycloack provider with real parameters
2. Send a request to /api/auth/keycloack/callback?access_token=XXX where XXX is a real access token
3. See error "Cannot read properties of undefined (reading 'subdomain')"

### Expected behavior

Successful response must be retrieved.

### Screenshots

Not applicable

### Code snippets

Not applicable

### Additional context

Grant and Purest libraries are used to process a request to specified provider.
Because of this typo correct provider could not be retrieved.

derrickmehaffy:
Fixed in https://github.com/strapi/strapi/pull/21392

PR #21392, opened 2024-09-20
fix: correct spelling of keycloak provider

Throttle:
Corrected TYPO in keycloak provider

### What does it do?

Fixes issue #21391 

### Why is it needed?

Default configuration of keycloak provider does not work.

### How to test it?

Configure the provider according to https://docs.strapi.io/dev-docs/plugins/users-permissions#keycloak-config

### Related issue(s)/PR(s)

No related issues


alexandrebodin (approved):
Hi @Throttle thank you for the fix 💯

PR #21400, opened 2024-09-20
feat(review-workflows): check entry stage before publish

Feranchz:
### What does it do?

Add validation when ContentType is related to one RW with required stage for publish. Entry only will be published if the entry is at the right stage.

### How to test it?

1. Create one RW with a required stage for publish and assign it to one Content Type.
2. Create or go to an already existing entry of that Content Type and put it in a wrong stage
3. Try to publish, it should throw an error
4. Change entry's stage for the required one and try to publish
5. It should publish with no errors

### DIFF
--- a/packages/core/review-workflows/server/src/services/document-service-middleware.ts
+++ b/packages/core/review-workflows/server/src/services/document-service-middleware.ts
@@ -1,4 +1,5 @@
 import { Modules } from '@strapi/types';
+import { errors } from '@strapi/utils';
 
 import { isNil } from 'lodash/fp';
 import { ENTITY_STAGE_ATTRIBUTE } from '../constants/workflows';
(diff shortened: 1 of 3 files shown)


packages/core/review-workflows/server/src/services/document-service-middleware.ts
```diff
@@ -122,7 +123,34 @@ const handleStageOnUpdate: Middleware = async (ctx, next) => {
   return next();
 };
 
+/**
+ * Check if the entity is at the required stage before publish
+ */
+const checkStageBeforePublish: Middleware = async (ctx, next) => {
+  if (ctx.action !== 'publish') {
+    return next();
+  }
```
remidej, line 132: Did you check if action === publish is enough to handle bulk publish and publishing via releases?

Feranchz: Yes! It's enough for both cases so we are safe!

remidej (approved):
Tested ✅

PR #21402, opened 2024-09-20
fix: drop and create columns and indexes in the correct order

### LINKED ISSUE #21398: [CTB] When changing the type of relation in the CTB, my app crashes

- Node.js version: v20.9.0
- Strapi version: 5.0.0 (and on this fix's experimental : https://github.com/strapi/strapi/pull/21390)
- Database: Sqlite
- Operating system: WSL (reproduced on IOS)

### Describe the bug

In a fresh app with zero data, when I'm creating my scheme in the CTB, while changing the type of relation field in a collection type, then my app crashes.

1. Create a fresh app with no example data.
2. Go to CTB, create a first Collection Type (here : "Owner" with only a Text field "OwnerName", i18n and D&P both disabled)
3. Create a second Collection Type (here : "Dog" with only a Text field "DogName", i18n and D&P both enabled)
4. In Dog CT, create a Relation field to Owner (here : "Multidirectional" with Owner having many dogs). Save.
5. Still in Dog CT, update the type of Relation field to Owner (here : turn it to unidirectional with dog having only one owner)
6. Save

### Expected behavior

App shouldn't crash, relation field update should be saved.

### Sceenshots

https://github.com/user-attachments/assets/4dd9ef02-1c5d-4ba0-b69e-d4aa92799876

innerdvations:
### What does it do?

fixes #21398 

### Why is it needed?

When we were dropping a column in the database, we were still trying to delete indexes on that column, but those indexes were already being dropped along with the column, so we got an error

NOTE: specifically for indexes, I think we should consider in the future adding a conditional check for if the index exists, and if so log a warning rather than throwing an error. Bugs like this one would still be caught and fixed, but it would become a 'low' rather than a 'critical' or 'high' because Strapi would still be able to start up, and the worst case scenario would be a few extra indexes lingering when they shouldn't be there

### How to test it?

With all supported databases:

- Specifically, test the linked issue
- More generally, test all kinds of creating and deleting content types and relations in the CTB
- To be 100% safe, please test a migration of a complex dataset (ie, one with every relation type) from v4 to v5 and ensure this hasn't broken anything there

I will also add a CTB e2e test for this asap before we merge, but it may not be able to be enabled on the CI due to issues we haven't yet been able to solve with github and the server restarts required by the CTB

### Related issue(s)/PR(s)


packages/core/database/src/schema/builder.ts
```diff
@@ -287,6 +282,11 @@ const createHelpers = (db: Database) => {
         }
       }
 
+      for (const removedColumn of table.columns.removed) {
```
alexandrebodin, line 285: I thought we had to drop the columns before dropping certain indexes 🤔  (unique, multi col indexes etc) 

Would it be safer to add the exists check instead directly and not change the order ?

innerdvations: AFAIK dropping indexes can always be done, it's the column dropping that can fail due to a fk or multi column index.  So they should always happen first.

My initial attempt was actually to check if index exists, but it's more complicated because it's different for each dialect, and requires a raw query for each index, which could potentially introduce connection pool issues. And it can't all be done at once up front, because something out of order like this would still cause a problem. I'll upload that version of the fix as a poc as well so you can see.

But this makes sense to me as the "real" solution, whereas that would be a much more complicated band-aid sort of fix... which I still think we should do to reduce the severity of any similar issues in the future, but only with a lot more testing.

innerdvations:
Also, I should note that moving "create column" ahead of "create indexes" was NOT part of the bug fix in the linked issue, just another logic error that seemed obvious and important to avoid potential issues in the future. If anyone feels it's unsafe or unneccessary since there aren't any known bugs around it, I can revert that part.

packages/core/database/src/schema/builder.ts
alexandrebodin: We could also consider split the different steps in multiple transactions that we consider more or less "required" to succeed 🤔 

Let's go with that for now. Can you make an experimental to get the user testing ?

innerdvations: I'm putting together a second PR for exactly how I think this part of the code should work, but it's a somewhat bigger change so I don't want to merge it as the fix for this issue and potentially make it worse. It'll also take a lot of discussion we don't want to wait for since this is a high bordering on critical

alexandrebodin (approved):
LGTM. 

Need to add some CTB tests that update a CTB structure in all kinds of directions. Right after

innerdvations:
Experimental: 0.0.0-experimental.8f5736f39974781dca58cef1d1dea9d54580fbf1

packages/core/database/src/schema/builder.ts
innerdvations: Here it is: https://github.com/strapi/strapi/pull/21403

I'll come back to that later once we actually have tests to test it against :)

innerdvations:
Added e2e test (confirmed it fails on `develop`)

I'm trying to enable it on the CI as well, if it fails I'll add the condition back.

MarionLemaire:
Tested with data and with zero data : no app crashing anymore, relation field update saved, well done :)

butcherZ (approved):
tested, LGTM :D

PR #21403, opened 2024-09-20
fix: [POC] refactor alterTable and check if indexes exist

innerdvations:
### What does it do?

- removes custom logic for mysql by deleting regular indexes before fk indexes
- checks if indexes exist before attempting to delete them

### Why is it needed?

See comment thread in https://github.com/strapi/strapi/pull/21402

### How to test it?

### Related issue(s)/PR(s)

Possible fix for #21583


innerdvations:
After seeing https://github.com/strapi/strapi/issues/21583 , this might need the reverse logic too, where it doesn't attempt to create and index if it already exists

innerdvations:
Closing this for now because it has the potential to cause some problems to go unnoticed and most of the issues we had in this area are fixed. Still might be worth revisiting later.

PR #21407, opened 2024-09-21
enhancement: change all internal cron tasks to named tasks

derrickmehaffy:
## What this does

Previously all of our built-in metrics based cron tasks were using the simplified format:

```js
cron.add({
  '*/1 * * * *': () => console.log('A minute has passed.'),
});
```

And they have been changed to use the named format

```js
cron.add({
  myTask: {
    task,
    options: '*/5 * * * *', // Executes every 5 minutes.
  },
});
```

## Why this change

As the maintainer of the Strapi Redis Plugin and with the new changes coming in my plugin for Strapi 5 I'm integrating the Redlock package which automatically forces all cron-tasks to use red-locks to prevent clustered environments from running the same task at the same time.

In essence what this could do is help prevent duplicate data being sent to the telemetry system from the same project running in a horizontal or vertical cluster mode.

## How to test it

Simply run Strapi in console mode and review the output of `strapi.cron.jobs` and see that all of the built in cron-tasks now have a specified name instead of being anonymous tasks

### DIFF
--- a/packages/core/admin/ee/server/src/services/metrics.ts
+++ b/packages/core/admin/ee/server/src/services/metrics.ts
@@ -49,7 +49,10 @@ const sendUpdateProjectInformation = async (strapi: Core.Strapi) => {
 
 const startCron = (strapi: Core.Strapi) => {
   strapi.cron.add({
-    '0 0 0 * * *': () => sendUpdateProjectInformation(strapi),
+    sendProjectInformation: {
+      task: () => sendUpdateProjectInformation(strapi),
+      options: '0 0 0 * * *',
+    },
   });
 };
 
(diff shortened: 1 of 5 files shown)


derrickmehaffy:
I'm setting this as community as it's coming from me as a "community plugin maintainer" rather than being a team member :)

alexandrebodin:
Can you explain why you need the tasks to be named ? Your explanation doesn't say why it is required 🤔

derrickmehaffy:
When the distributed lock is created in redis it uses the name as an identifier so if they aren't named then each of the instances doesn't see the lock and they still try and execute them.

See this part of the code from my plugin: 

https://github.com/strapi-community/plugin-redis/blob/0c375ca642ad2b7d426fa286385a100e22c7e9f8/server/register.js#L69

There is some fail over to attempt to inject red lock into the cron task but in testing with our internal ones we have some that overlap and the locks aren't created properly (multiple instances still fire them in parallel when they shouldn't).

alexandrebodin (approved):

alexandrebodin:
Did some QA and it's working just fine

PR #21409, opened 2024-09-21
Fix unidirectional-relations.js missing await on getConnection promise

sbastias:
### What does it do?

Added "await" before strapi.db.getConnection() 

image

### Why is it needed?

Fixes Publish internal error on unidirectional relationship entry
```
TypeError: con.batchInsert is not a function
    at /.../node_modules/@strapi/core/dist/services/document-service/utils/unidirectional-relations.js:53:17
```

### How to test it?

Not sure how to test - but it fixed the bug.


sbastias:
Never mind. This didn't fix the issue - it just failed silently and skipped the rest of the function making it seem like the bug was fixed.

Issue #21395, opened 2024-09-20
MigrationError: Migration core::5.0.0-discard-drafts (up) failed: Original error: `con.batchInsert` is not a function


jackbuehner:
### Required System information

- Node.js version: v18.20.4
- NPM version: 10.7.0
- Strapi version: 5.0.0
- Database: PostgreSQL
- Operating system: Windows 10 21H2 (LTSC)
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When I upgrade to Strapi 5.0.0 and run Strapi for the first time, I get a migration error:

```
[2024-09-20 00:00:01.604] error: Migration core::5.0.0-discard-drafts (up) failed: Original error: con.batchInsert is not a function
MigrationError: Migration core::5.0.0-discard-drafts (up) failed: Original error: con.batchInsert is not a function
    at C:\Users\jackb\Workspace\content-manager\node_modules\umzug\lib\umzug.js:169:27
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Umzug.runCommand (C:\Users\jackb\Workspace\content-manager\node_modules\umzug\lib\umzug.js:125:20)
    at async Object.up (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\database\dist\index.js:6531:7)
    at async Object.up (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\database\dist\index.js:6555:11)
    at async Object.sync (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\database\dist\index.js:2246:9)
    at async Strapi.bootstrap (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\core\dist\Strapi.js:351:5)
    at async Strapi.load (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\core\dist\Strapi.js:315:5)
    at async Strapi.start (C:\Users\jackb\Workspace\content-manager\node_modules\@strapi\core\dist\Strapi.js:207:9)
[2024-09-20 00:00:01.606] info: Shutting down Strapi
```

### Steps to reproduce the behavior

1. Upgrade from Strapi 4.25.10.
2. Make a testing/development copy of the database from 4.25.10 and configure Strapi to use the copy.
3. Start Strapi.
4. After several minutes, the migration fails.

### Expected behavior

The migration should not fail.

### Additional context

This is probably related to https://github.com/strapi/strapi/issues/21352.

bwcr:
[...]

PR #18857, opened 2023-11-20
fix: db connection fails when providing a function as the db connecti…

evan-thinksync:
…on config

### What does it do?

Fixes this issue: [#18474](https://github.com/strapi/strapi/issues/18474)

### Why is it needed?

To be able to use dynamic database connections supported by Knex.

### How to test it?

Provide a function for the db connection config

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

Issue #18474, opened 2023-10-17
Dynamic database connection config


evan-thinksync:
### Required System information

- Node.js version: v18.14.2
- Yarn version: 1.22.19
- Strapi version: 4.14.4
- Database: postgres
- Operating system: darwin-arm64
- Is your project Javascript or Typescript: Typescript

### Describe the bug
[Knex supports dynamic database connection](https://knexjs.org/guide/#configuration-options) by providing a function for the database connection config.  It's been noted that the data inside of the strapi db config is passed directly to knex however when I use a function for the strapi db connection config, I get the following error 
`TypeError: Cannot read properties of undefined (reading 'schema')`
from the [getSchemaName() function](https://github.com/strapi/strapi/blob/7c56d94b0c60d76f26a6b3532475f001d5414b42/packages/core/database/src/index.ts#L141-L143) resulting in a failed startup.

The issue is that `this.connection.client.connectionSettings` is undefined when providing a function as the connection config. 

### Steps to reproduce the behavior

1. Provide a function as the strapi db connection config. i.e.
```
export default ({ env }) => ({
  connection: {
    client: 'postgres',
    connection: async () => {
      const {
        token,
        tokenExpiration
      } = await someCallToGetTheToken();

      return {
        host : 'your_host',
        port : 5432,
        user : 'your_database_user',
        password : token,
        database : 'myapp_test',
        expirationChecker: () => {
          return tokenExpiration <= Date.now();
        }
      };
    }
  }
})
```
2. Start strapi app

### Expected behavior
Strapi to support dynamic database connection via db connection config function.

### Screenshots
N/A

### Code snippets
N/A

### Additional context
N/A

PR #21412, opened 2024-09-21
Revise database/index to skip .withSchema when no tableName passed to strapi.db.getConnection()

sbastias:
This was breaking in unidrectional-relations.js, returning a QueryBuilder object instead of a Knex object.

### What does it do?

Skips the connection.withSchema directive if tableName is NOT provided. Otherwise, strapi.db.getConnection() returns a QueryBuilder instance instead of a Knex instance.

### Why is it needed?

unidrectional-relations.js was breaking on Publish of a saved entry with:
```
TypeError: con.batchInsert is not a function
    at .../server/node_modules/@strapi/core/dist/services/document-service/utils/unidirectional-relations.js:69:17
```

### How to test it?

Save an entry with unidrectional relationship to another object. Publish it. Then try to update, save, Publish.

**Result:**
image

**After revision** to return Knex object instead of QueryBuilder, the Publish function works:
image

### DIFF
--- a/packages/core/database/src/index.ts
+++ b/packages/core/database/src/index.ts
@@ -166,6 +166,7 @@ class Database {
   getConnection(): Knex;
   getConnection(tableName?: string): Knex.QueryBuilder;
   getConnection(tableName?: string): Knex | Knex.QueryBuilder {
+    if (!tableName) return this.connection;
     const schema = this.getSchemaName();
     const connection = tableName ? this.connection(tableName) : this.connection;
     return schema ? connection.withSchema(schema) : connection;


alexandrebodin:
Fix https://github.com/strapi/strapi/issues/21395

innerdvations:
This bypasses the schema check which we need. I've got a fix that instead updates unidirectional-relations to use queryBuilder instead of needing Knex.

Now that I've seen this though, I feel like we should deprecate and rename the tablename overload, because it feels like it could lead to some other subtle bugs. I'll keep it in mind while we investigate https://github.com/strapi/strapi/issues/18474 https://github.com/strapi/strapi/pull/18857

PR #21419, opened 2024-09-23
skipping uniquness e2e test on webkit

Bassel17:
### What does it do?

- Skipping uniqueness e2e test on webkit

### Why is it needed?

- The test is only failing on webkit on the github CI, we already have ticket to fix the webkit issues
- it's blocking PRs, so until we get around to fix this we will skip only on webkit 

### How to test it?

wait for CI tests to pass


innerdvations:
Looks like a `prettier` issue still, other than that lgtm as long as the e2e tests now pass

innerdvations (approved):

PR #21421, opened 2024-09-23
fix: cannot create component from dynamic zone

### LINKED ISSUE #21370: Can’t create a new component from the Dynamic Zone because the category folder is disabled
https://github.com/user-attachments/assets/c1d2063e-256c-4368-b95f-28b4020d2231

Bassel17:
### What does it do?

- Make the category field enabled when trying to create component from a Dynamic zone
- Adding an extra check to see if we are creating from a DZ

### Why is it needed?

check #21370 

### How to test it?

create component from a dynamic zone in the CTB

### Related issue(s)/PR(s)

fixes #21370


butcherZ (approved):
Did QA on my side,
- Can create component from dynamic zone correctly now ✅ 
- The category name cannot be changed when attempting to edit a component. ✅ 

LGTM!  :D

jhoward1994 (approved):

PR #21422, opened 2024-09-23
chore: update path-to-regexp to 8.1.0

innerdvations:
### What does it do?

updates path-to-regexp

### Why is it needed?

Security issue in body-parser subdependency: https://nvd.nist.gov/vuln/detail/CVE-2024-45590

### How to test it?

documentation plugin should work, including api endpoints

### Related issue(s)/PR(s)

DX-1632


innerdvations:
This doesn't actually work; closing until I have time to fix it

Issue #21382, opened 2024-09-19
DP-ed Relations in components in non-DP types show up `null` in REST API


laurenskling:
### Required System information

- Node.js version: 20
- NPM version: yarn 1.22
- Strapi version: 5.0.0 (and rc 17-29 at least)
- Database: sqlite
- Operating system: macos
- Is your project Javascript or Typescript: js

### Describe the bug + Steps to reproduce the behavior

follow up bug report to https://github.com/strapi/strapi/issues/21355

I have a Homepage singleType. It doesnt have draftAndPublish, because the homepage never goes unpublished (v4 logic of DP).

 have a repeatable component with relations inside. (see https://github.com/strapi/strapi/issues/21354)
After migration, when I request the REST API all return correctly. (on `http://0.0.0.0:1337/api/homepage/homepage?populate[sections][populate][section]=*`). When I change one of the relations in the UI, that relation goes `null` in the REST API result. So after changing all the relations in the UI, all of them will return f.e. `{ "id": 2, "section": null },`. To make things confusing, when I add `&status=draft` to the REST call, the _last saved relation_ is returned. So if I change component 2 to another relation, with `&status=draft`, it will now return the relation, but component 2 will be `null` again when i change component 3.

Make it make sense.

### Expected behavior

A non-DP contentType should always return all of the requested relations.

### Aditional context

will create a reproduction codebase a.s.a.p.

laurenskling:
I'm pretty sure, as https://github.com/strapi/strapi/issues/21355 looks like a unidirectional bug, this is the same problem. As relations in components are always unidirectional.

laurenskling:
okay I did some more testing and this really does seem to be related to non-DP. I've tried collectionType/singleType, no difference. But when I change `draftAndPublish` from true to false, relations are showing up as `null`

laurenskling:
Found the combination!
Repo: https://github.com/laurenskling/strapi-5-tests

The exact situation is (based on the content types in this repo):
1. have DP-ed Authors (non-DP authors work)
2. have a non-DP Articles (DP will work)
3. have a (repeatable) component to the author(s) in articles (components in DZ's _do_ work! also in the repo)
4. add 2 authors
5. add an article with 2 quotes, relate them to the authors

<img width="844" alt="Scherm­afbeelding 2024-09-19 om 21 39 54" src="https://github.com/user-attachments/assets/b2a4b10c-f868-4add-962f-1fc2205cb80b">

```
[...]

PR #21428, opened 2024-09-23
fix: cm relations

alexandrebodin:
### What does it do?

- Fix https://github.com/strapi/strapi/issues/21355
- Fix https://github.com/strapi/strapi/issues/21382

Replacement for https://github.com/strapi/strapi/pull/21390

### Details

- Fixes findExisting returning both draft & published entries when non DP relations are present
- Fixes UI not sending the documentId & locale as part of relations infos
- Fixes Relation links not including the locale & redirecting to the defaultLocale instaed (will need refactoring later)
- When a DP doc is published and is targeted by a non DP source with a unidirectional relation, the relation should point to both Draft & Published
- v4 to v5 migration doesn't create links to both published & drafts when running discardDraft

### Others issues to fix in another PR

- RTK query cache is messing up with the UX

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -208,6 +208,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
 
     const realServerRelationsCount =
       'pagination' in data && data.pagination ? data.pagination.total : 0;
+
     /**
      * Items that are already connected, but reordered would be in
      * this list, so to get an accurate figure, we remove them.
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -84,7 +84,12 @@ function useHandleDisconnect(fieldName: string, consumerName: string) {
       }
     }
 
-    addFieldRow(`${fieldName}.disconnect`, { id: relation.id });
+    addFieldRow(`${fieldName}.disconnect`, {
+      id: relation.id,
+      api_data: {
+        documentId: relation.documentId,
+      },
+    });
   };
 
   return handleDisconnect;
@@ -145,14 +150,17 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
     const isMorph = props.attribute.relation.toLowerCase().includes('morph');
     const isDisabled = isMorph || disabled;
 
-    const { id: componentId, uid } = useComponent('RelationsField', ({ uid, id }) => ({ id, uid }));
+    const { id: componentId, uid: componentUID } = useComponent(
+      'RelationsField',
+      ({ uid, id }) => ({ id, uid })
+    );
 
     /**
      * We'll always have a documentId in a created entry, so we look for a componentId first.
      * Same with `uid` and `documentModel`.
      */
     const id = componentId ? componentId.toString() : documentId;
-    const model = uid ?? documentModel;
+    const model = componentUID ?? documentModel;
 
     /**
      * The `name` prop is a complete path to the field, e.g. `field1.field2.field3`.
@@ -259,6 +267,10 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
 
       const item = {
         id: relation.id,
+        api_data: {
+          documentId: relation.documentId,
+          locale: relation.locale,
+        },
         status: relation.status,
         /**
          * If there's a last item, that's the first key we use to generate out next one.
@@ -268,7 +280,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
         [props.mainField?.name ?? 'documentId']: relation[props.mainField?.name ?? 'documentId'],
         label: getRelationLabel(relation, props.mainField),
         // @ts-expect-error – targetModel does exist on the attribute, but it's not typed.
-        href: `../${COLLECTION_TYPES}/${props.attribute.targetModel}/${relation.documentId}`,
+        href: `../${COLLECTION_TYPES}/${props.attribute.targetModel}/${relation.documentId}?${relation.locale ? `plugins[i18n][locale]=${relation.locale}` : ''}`,
       };
 
       if (ONE_WAY_RELATIONS.includes(props.attribute.relation)) {
@@ -294,7 +306,8 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
         <StyledFlex direction="column" alignItems="start" gap={2} width="100%">
           <RelationsInput
             disabled={isDisabled}
-            id={id}
+            // NOTE: we should not default to using the documentId if the component is being created (componentUID is undefined)
+            id={componentUID ? (componentId ? `${componentId}` : '') : documentId}
             label={`${label} ${relationsCount > 0 ? `(${relationsCount})` : ''}`}
             model={model}
             onChange={handleConnect}
@@ -389,7 +402,7 @@ const addLabelAndHref =
         // Fallback to `id` if there is no `mainField` value, which will overwrite the above `documentId` property with the exact same data.
         [mainField?.name ?? 'documentId']: relation[mainField?.name ?? 'documentId'],
         label: getRelationLabel(relation, mainField),
-        href: `${href}/${relation.documentId}`,
+        href: `${href}/${relation.documentId}?${relation.locale ? `plugins[i18n][locale]=${relation.locale}` : ''}`,
       };
     });
 
@@ -454,6 +467,7 @@ const RelationsInput = ({
      * individual components. Therefore we split the string and take the last item.
      */
     const [targetField] = name.split('.').slice(-1);
+
     searchForTrigger({
       model,
       targetField,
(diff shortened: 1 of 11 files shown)


alexandrebodin:
Experimental `0.0.0-experimental.727a0f9b729da91886ca93e88ec1d6d811d1beba`

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
```diff
@@ -84,7 +84,12 @@ function useHandleDisconnect(fieldName: string, consumerName: string) {
       }
     }
 
-    addFieldRow(`${fieldName}.disconnect`, { id: relation.id });
+    addFieldRow(`${fieldName}.disconnect`, {
+      id: relation.id,
+      api_data: {
```
remidej, line 89: just a style question but I wonder why snake case?

packages/core/content-manager/server/src/controllers/relations.ts
```diff
@@ -458,7 +458,7 @@ export default {
       ordering: 'desc',
     });
 
-    const relationsUnion = uniqBy('id', concat(sanitizedRes.results, res.results));
+    const relationsUnion = uniqBy('documentId', concat(sanitizedRes.results, res.results));
```
Convly, line 461: Would it be possible to easily add a test (unit? API?) to make sure we're doing the unique transform on `documentId` and not `id`? Seems like a critical path given the issue this fix

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
```diff
@@ -145,14 +150,17 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
     const isMorph = props.attribute.relation.toLowerCase().includes('morph');
     const isDisabled = isMorph || disabled;
 
-    const { id: componentId, uid } = useComponent('RelationsField', ({ uid, id }) => ({ id, uid }));
+    const { id: componentId, uid: componentUID } = useComponent(
+      'RelationsField',
+      ({ uid, id }) => ({ id, uid })
```
remidej, line 155: you could also rename the values here in the selector

packages/core/content-manager/server/src/controllers/relations.ts
alexandrebodin: Yes I'll add some tests once the QA validates it fixes the issues as there are many :/

remidej, line 461: I wanted to make this change last month but Marc warned me it could break some scenarios. Could you check [his comments in that thread](https://github.com/strapi/strapi/pull/21010/files/f8c21235d3e01a5fa5b2dd7753ea53d7eb395bb7#r1716846564)? I don't know if they're use cases we want to support

remidej (commented):
I like the api_data solution :)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
alexandrebodin: just doing too much rust right now :D I'll rename it

alexandrebodin:
New eperimental `0.0.0-experimental.9bd63a0f51283779d88aa9d86aa78c40ed628a15`

packages/core/content-manager/server/src/controllers/relations.ts
alexandrebodin: Nice catch 🎉  I updated the condition to check by docId and locale as we need the status to be ignored

remidej (commented):
Code changes LGTM, though ideally we'd get Marc to review it too as he has way more knowledge of this backend part.

There are tests failing, should they be updated?

> RTK query cache is messing up with the UX

Could you give more details? We could try to fix that for 5.0.2

alexandrebodin:
Yet another exp `0.0.0-experimental.f8a68bc03b751aa0a66c4dcfaf83553c9b6adf2b`

MarionLemaire:
Re-using test plan from @oiorain and adding some stuff for relations in components. 
**Test cases**

**Unidirectional relations**

- [ ]  Regular entry to regular entry
- [ ]  D&P entry to regular entry
- [ ]  regular entry to D&P entry
- [ ]  Regular entry to i18n
- [ ]  i18n to regular entry
- [ ]  regular entry to d&p + i18n
- [ ]  d&p + i18n to regular
- [ ]  D&p to d&p + i18n
- [ ]  D&p + i18n to d&p
- [ ]  D&p + i18n to d&p + i18n

**Possible tests to perform in each case:

Many to many relation**

- [ ]  Add a Many to many relation
- [ ]  Add it from left side + check presence
- [ ]  Remove it from left side + check presence
- [ ]  Add it from right side + check presence
- [ ]  Remove it from right side + check presence

**When D&P is enabled**

- [ ]  Test publishing left side
- [ ]  Test publishing right side
- [ ]  Test editing the relationship on one side then the other
- [ ]  Test removing the relationship on one side then the other
- [ ]  test unpublishing left side
- [ ]  test unpublishing right side

**When i18n is enabled**

- [ ]  Test to enable i18n on a content type containing a relation on left side then right side
- [ ]  Test editing the relationship or entry
- [ ]  Test to disable i18n

**When D&P and i18n is enabled**

- [ ]  Test to publish a content with i18n on one side then right side
- [ ]  Test to unpublish a content with i18n on one side then right side

**Possible test cases for a one way relation**

- [ ]  Add a Many way relation + check presence
- [ ]  Remove it + check presence

**When D&P is enabled**

- [ ]  Test publishing
- [ ]  Test editing the relationship
- [ ]  Test removing the relationship
- [ ]  test unpublishing

**When i18n is enabled**

- [ ]  Test to enable i18n on a content type containing a relation
- [ ]  Test editing the relationship or entry
- [ ]  Test to disable i18n

**When D&P and i18n is enabled**

- [ ]  Test to publish a content with i18n
- [ ]  Test to unpublish a content with i18n

**Relations in components**

- [ ] For D&P enabled collection type entries
- [ ] For D&P disabled collection type entries

remidej:
@alexandrebodin I'm seeing duplicate relations after saving when relations are in components. This happens only after saving, but seems to be in all scenarios: inside a component or not, from a D&P content type or not, to a D&P content type or not

https://github.com/user-attachments/assets/487abbd3-cea2-4e2d-8add-5efef90c5d2d

alexandrebodin:
> @alexandrebodin I'm seeing duplicate relations after saving when relations are in components. This happens only after saving, but seems to be in all scenarios: inside a component or not, from a D&P content type or not, to a D&P content type or not
> 
>  CleanShot.2024-09-24.at.16.11.25.mp4

Yeah I just found out. I think it's because they don't have a docId, I'll also include id in the rule

remidej:
https://github.com/user-attachments/assets/4dc7d956-3189-44ff-997d-e3b56d915bd2

@alexandrebodin I can't remove a relation if it's from a D&P content type to a D&P content type and inside of a relation

MarionLemaire:
@alexandrebodin status issue here :
When I have an undirectional multiple relation from a CT (with D&P) - here Cat- to another CT (with D&P and i18n) - here Dog, it seems that the status displayed in the relation field of the first CT (Cat) displays the wrong status of the entry of the second one if in multiple locales. In my case, if one of the localized entry is published, then every entry of this CT has a Published status even if it's still in Draft.


https://github.com/user-attachments/assets/dae66213-36a1-44f1-ac8a-6df08c2dd4e7

MarionLemaire:
@alexandrebodin 
On many-to-many relation field between 2 Collection types having D&P enabled, I randomly have this error of only 1 entry remains when saving even if many are selected (even after refreshing the page) : 

https://github.com/user-attachments/assets/e283b18e-9655-40f5-8be8-44c2ff5d4716

MarionLemaire:
@alexandrebodin 
On many-to-many relation field between 2 collection types being enabled (here : Kid and Mom), if I enable i18n on one of these content types (here : Kid) and create localized entries for this CT, the status of the localized entries of the relation field in non-localized CT doesn't display the right status (here : Kid/Merlin is published in ENG by default, but draft in FR, though in Mom/Marion content type relation field Merlin & Merlin FR both have Published status even after refreshing)

https://github.com/user-attachments/assets/7430666d-ceb2-4d76-84b9-f4901be13d7b

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -27,29 +28,61 @@ const load = async (uid: UID.ContentType, oldEntries: { id: string; locale: stri
...
+
+          const draftRelations = await strapi.db
+            .getConnection()
+            .select('*')
+            .from(joinTable.name)
+            .whereIn(name, oldEntriesIds)
+            .transacting(trx);
+
+          if (draftRelations.length > 0) {
+            updates.push({ joinTable, relations: draftRelations.map(omit('id')) });
```
Marc-Roig, line 83: is it necessary to omit the id?

alexandrebodin:
latest exp `0.0.0-experimental.b1d7921ddb1b36c84b58a3946e1dc9dbda91b2dc`

Marc-Roig:
I believe there are a couple of problem with relation pagination
image

This is a unidirectional (NoDP NoI18n -> DP I18n) relation 
I have 6 connected relations,
- After clicking the load more button I don't see any page being loaded
- When displaying more than 5 relations (more than one page) , deleting relations and saving doesnt update the field cache  and all of the deleted relations pop up again.
- When saving, the relation pagination is reseted and  you go back to one page. even if you had more than one loaded

oiorain:
Tested: 
- D&P one way to D&P 
- D&P one way to D&P + i18n
- D&P + i18n  one way to D&P + i18n
- D&P + i18n  many to many to D&P + i18n

I didnt find any bugs! 
Good job guys :)

Marc-Roig (approved):
Did some final QA on relations, custom permisions, reordering, findExisting and findAvailable endpoints

other than some rtk issues you already mention everything looks ok

Marc-Roig (approved):
Did some final QA on relations, custom permisions, reordering, findExisting and findAvailable endpoints

other than some rtk issues you already mention everything looks ok

Marc-Roig:
I would prioritize adding more tests (both api and e2e) as its supppper tedious to test all cases

PR #21440, opened 2024-09-24
chore: fix strip-hash

innerdvations:
### What does it do?

fix the regexp for stripping the hash on the size comparison action

### Why is it needed?

the matching is wrong so it's considering every single file to be removed or added, instead of comparing accurately

### How to test it?

it's CI, I'm not sure if it will affect this PR or only work after it's merged

### Related issue(s)/PR(s)


alexandrebodin (approved):
I think there was a discussion to remove it entirely. but nice if it's at least fixed.

innerdvations:
Yeah, I thought this might allow us to keep it if anyone finds it useful, but removing it altogether would work too. As long as we stop getting giant lists of a hundred files every PR 😆

remidej (approved):
I'm all for keeping it if it stops spamming PRs 😁

innerdvations:
Ok, let's hope my regexp is right, won't know for sure until the next front-end PR 😆

PR #19883, opened 2024-03-21
add export {} declaration at the end of the generated typescript files

n1xx1:
### What does it do?

This changes how the `.d.ts` files are generated in `types/generated` to prevent an issue where if you have no components or no content-types then it's mistaken as an [Ambient Module](https://www.typescriptlang.org/docs/handbook/modules/reference.html#ambient-modules) declaration. An empty export declaration is sufficient to make the file always be a module.

### Why is it needed?

Without this all the types declared in `@strapi/types` (and `@strapi/strapi`, since they are used by it) aren't available unless you create both a component and a content-type. This doesn't happen when after you create components or content-types because an import statement is added outside the `declare module` block, so the file becomes a module and the issue goes away.

### How to test it?

In a fresh ts project, after you started the development server or generated the typescript declaration, then attempt to `import { Strapi } from "@strapi/strapi"` in any .ts file. With this fix it should work just fine, without it you will be met with a `Module '"@strapi/strapi"' has no exported member 'Strapi'.` error.

### Related issue(s)/PR(s)

I haven't found any issues or PRs mentioning this issue.

PR #21441, opened 2024-09-24
Add Placeholders for Empty Artifacts in Type Generation

Convly:
### What does it do?

Do not generate the module augmentation for empty artifact collections (content-types, components).

Instead, add a comment to explain why the file is empty.

### Why is it needed?

When no components (or content-types) exist in an app, we still try to do the module augmentation which can lead to unexpected behavior in the type system.

### How to test it?

- `cd examples/kitchensink-ts`
- `yarn strapi ts:generate-types -d`
- open `types/generated/components.d.ts`
- The file should only contain a single comment

- `cd examples/kitchensink-ts`
- `yarn strapi ts:generate-types -d`
- open `types/generated/components.d.ts`
- The file should contain all the components' type definitions + the module augmentation

### Related issue(s)/PR(s)

take over #19883

### DIFF
--- a/packages/utils/typescript/lib/generators/components/index.js
+++ b/packages/utils/typescript/lib/generators/components/index.js
@@ -6,7 +6,7 @@ const { models } = require('../common');
 const { emitDefinitions, format, generateSharedExtensionDefinition } = require('../utils');
 
 const NO_COMPONENT_PLACEHOLDER_COMMENT = `/*
- * The app doesn't have any component yet.
+ * The app doesn't have any components yet.
  */
 `;
 
--- a/packages/utils/typescript/lib/generators/components/index.js
+++ b/packages/utils/typescript/lib/generators/components/index.js
@@ -5,6 +5,11 @@ const { factory } = require('typescript');
 const { models } = require('../common');
 const { emitDefinitions, format, generateSharedExtensionDefinition } = require('../utils');
 
+const NO_COMPONENT_PLACEHOLDER_COMMENT = `/*
+ * The app doesn't have any component yet.
+ */
+`;
+
 /**
  * Generate type definitions for Strapi Components
  *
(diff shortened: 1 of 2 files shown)


packages/utils/typescript/lib/generators/components/index.js
```diff
@@ -5,6 +5,11 @@ const { factory } = require('typescript');
 const { models } = require('../common');
 const { emitDefinitions, format, generateSharedExtensionDefinition } = require('../utils');
 
+const NO_COMPONENT_PLACEHOLDER_COMMENT = `/*
+ * The app doesn't have any component yet.
```
innerdvations, line 9:
```suggestion
 * The app doesn't have any components yet.
```

packages/utils/typescript/lib/generators/content-types/index.js
```diff
@@ -5,6 +5,11 @@ const { factory } = require('typescript');
 const { models } = require('../common');
 const { emitDefinitions, format, generateSharedExtensionDefinition } = require('../utils');
 
+const NO_CONTENT_TYPE_PLACEHOLDER_COMMENT = `/*
+ * The app doesn't have any content-type yet.
```
innerdvations, line 9:
```suggestion
 * The app doesn't have any content-types yet.
```

packages/utils/typescript/lib/generators/content-types/index.js
```diff
@@ -23,6 +28,12 @@ const generateContentTypesDefinitions = async (options = {}) => {
     definition: models.schema.generateSchemaDefinition(contentType),
   }));
 
+  options.logger.debug(`Found ${contentTypesDefinitions.length} content-types.`);
```
innerdvations, line 31:
```suggestion
  options.logger.debug(`Found ${contentTypesDefinitions.length} content-type${contentTypesDefinitions.length === 1 ? '' : 's'}.`);
```
not sure if we care or not

packages/utils/typescript/lib/generators/content-types/index.js
innerdvations: Is it even possible to not have any content types though, since we combine it with our internal models?

innerdvations (dismissed):
Tested and works for me!

jhoward1994 (approved):

packages/utils/typescript/lib/generators/content-types/index.js
Convly: Shouldn't be possible in the context of a normal application, but better be on the safe side

packages/utils/typescript/lib/generators/content-types/index.js
Convly: I think it adds a lot of verbosity for very little benefit (those are debug logs anyway)

PR #21443, opened 2024-09-24
chore: set up preview feature boilerplate

remidej:
### What does it do?

- Sets up boilerplate so we can start working on preview
- Only loads feature when a `preview` feature flag is passed

I copied the system we used for history that worked pretty well: we have one folder for the feature in the CM admin, and one in the CM server. This way it's pretty isolated, which would make it easy to turn it into a plugin in the future. Each folder contains the EE license because they're outside of the EE directory.

### Why is it needed?

- so we can start working on preview
- so we don't need to maintain a feature branch
### How to test it?

- Start Strapi. You should see no Preview-related logs, whether it's in the browser console or terminal.
- Set a `STRAPI_FUTURE_PREVIEW` env var to true in your getstarted app
- Start Strapi. You should see "Bootstrapping preview" logs in both the browser console and the terminal


simotae14 (commented):
please fix the build issues you have in the content manager, thanks

packages/core/content-manager/admin/src/preview/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { PluginDefinition } from '@strapi/admin/strapi-admin';
+
+const previewAdmin = {
+  bootstrap(app) {
+    // TODO: Add license registry check when it's available
+    if (!window.strapi.future.isEnabled(FEATURE_ID)) {
+      return {};
+    }
+    // eslint-disable-next-line no-console -- TODO remove when we have real functionality
+    console.log('Bootstrapping preview admin');
```
Marc-Roig, line 14: should this be a dev log instead? this might be confusing for users if we merge into develop

Marc-Roig (approved):

packages/core/content-manager/admin/src/preview/index.ts
remidej: what do you mean by dev log? Checking node_env before logging?

Marc-Roig: sorry I meant the `strapi.log.debug` , but I just saw this is under a feature flag so I don't think it matters !

Issue #21450, opened 2024-09-24
[BUG] Forgot Password respond with "ok" for non-existent users or blocked users 


damancodes:
### Required System information

- Node.js version: v20.0.0
- NPM version:10.2.4
- Strapi version:5.0.0
- Database:
- Operating system:MacOs 14.5
- Is your project Javascript or Typescript:

### Describe the bug

Forgot password always returns ok even for blocked or non-existent users, causing unintended behavior

### Screenshots

<img width="721" alt="image" src="https://github.com/user-attachments/assets/a97158bc-9527-4d5c-9b15-60ba943bfd38">

derrickmehaffy:
This is intended so that we do not leak if the user exists or is blocked: https://github.com/strapi/strapi/blob/146a31b564bc8232686331f6b28f7ff966817963/packages/plugins/users-permissions/server/controllers/auth.js#L247-L249

PR #21451, opened 2024-09-24
fix: forgot password respond with ok for non-existent users or blocke…

damancodes:
Fixed 👍  #21450

Added
image


packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -237,7 +237,7 @@ module.exports = {
       .findOne({ where: { email: email.toLowerCase() } });
 
     if (!user || user.blocked) {
-      return ctx.send({ ok: true });
+      return ctx.throw(404, 'User not found or access blocked');
```
alexandrebodin, line 240: Hi, this is intentional to avoid brute force attacks being able to test for all existing users in the DB.

alexandrebodin (changes requested):
Sadly we cannot merge this PR

alexandrebodin:
I'm goign to close this PR as we don't want to leak user information and it's standar practice to just say "we are trying if the user exists"

Issue #20778, opened 2024-07-16
[Design] The description text of Content History should be Neutral600


MarionLemaire:
Current design
![CleanShot 2024-07-04 at 17 29 23@2x](https://github.com/user-attachments/assets/7af7d475-5b2d-4da1-82eb-69c970cbc6c3)

Expected design
![Image](https://github.com/user-attachments/assets/1d90feb6-2f4d-46e0-a433-3ac9a3a69911)

Figma file here: https://www.figma.com/design/XYJUrJn9i9ius3bOUSL4fn/Content-History?node-id=234-4385&node-type=frame&t=VXrhyi0M8NWqKrQl-0

PR #21453, opened 2024-09-24
fix: Changed the color of the subtitle

lucasboilly:
### What does it do?

I changed the color of the subtitle of the Content History page.

### Why is it needed?

It has to be a little less visible as it contains secondary informations.

### How to test it?

Open any entry and check its content history.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20778

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
+++ b/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
@@ -118,7 +118,7 @@ export const VersionHeader = ({ headerId }: VersionHeaderProps) => {
           minute: 'numeric',
         })}
         subtitle={
-          <Typography variant="epsilon">
+          <Typography variant="epsilon" textColor={"neutral600"}>
             {formatMessage(
               {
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
```diff
@@ -118,7 +118,7 @@ export const VersionHeader = ({ headerId }: VersionHeaderProps) => {
           minute: 'numeric',
         })}
         subtitle={
-          <Typography variant="epsilon">
+          <Typography variant="epsilon" textColor={"neutral600"}>
```
simotae14, line 121:
```suggestion
          <Typography variant="epsilon" textColor="neutral600">
```

simotae14 (dismissed):
you have a commitlint issue because of the format of your commit, you need to follow this convention https://github.com/conventional-changelog/commitlint/#what-is-commitlint and also a lint problem that will be solved with my suggested code, so close 👏🏽

Issue #20808, opened 2024-07-16
[Design] In the “Publish multiple locales” modal, the text style of the name is not the right one and there’s unexpected capital letters


MarionLemaire:
### Steps to reproduce
- Go to the edit view of an entry with at least 2 locales
- Click on the 3 dots next to the Publish button
- Click on "Publish multiple locales"
- See that the locale is written in capital letters and the second word starts with a capital letter

### Expected behavior

We should use small letters and not capital letters for the locale's name.
We should not use a capital letter for the second word in the status column.

### Screenshots

**Current design:**
![CleanShot 2024-07-05 at 11 21 06@2x](https://github.com/user-attachments/assets/6e98a1cd-4100-4217-a232-d678d4413411)

**Expected design:**
![Image](https://github.com/user-attachments/assets/18d45076-5289-4522-b0bd-4d8d1bb2127b)

**Figma file here:** https://www.figma.com/design/Tt8uHuQezEP0CsWunbulb5/Draft-%26-Publish?node-id=6865-43455&t=Bde7KX17urNDkv9p-0

yanniskadiri:
Confirmed in rc.7

PR #21454, opened 2024-09-24
fix: Removed the unexpected capital letters

lucasboilly:
### What does it do?

I removed the capital letters that were at the beginning of the words "Multiple" and "Locales".

### Why is it needed?

Publish and unpublish multiple locales were the only items in the popover with a capital letters for each word.

### How to test it?

Open an entry that has i18n and click the "More" button next to the "Publish" button.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20808


simotae14 (changes requested):
you have a commitlint issue because of the format of your commit, you need to follow this convention https://github.com/conventional-changelog/commitlint/#what-is-commitlint

remidej (commented):
We can bypass the commitlink check this time, but you need to fix some end to end tests. Look for `getByRole('menuitem', { name: 'Publish Multiple Locales', exact: true })` in your repo. It no longer finds it, you need to update the casing there

lucasboilly:
@simotae14 @remidej should be good now (except the commitlint issue).

simotae14 (approved):
Well done @lucasboilly 👏🏽

PR #21455, opened 2024-09-24
Add an environment variable for disabling /_health log

derrickmehaffy:
### What does it do?

Adds a conditional statement to look for an environment variable called `LOG_HEALTH` and specifically looks for this value to be set false (it will obviously be set to undefined by default) and checks to see if the route to be logged is the `/_health`

If both of these conditionals pass then the request won't be logged to the console

### Why is it needed?

Since we transitioned Strapi Cloud to use an HTTP healthcheck rather and a TCP port check all Strapi Cloud runtime logs are constantly spammed by us checking the status of the running project as you can see below:

image

This is obviously quite annoying to users that are looking at the runtime logs within the Strapi Cloud dashboard but also increases the size of the log file and slows down the real time logs rendering within Strapi Cloud

### How to test it?

Using one of the example projects within the repo:

- Run Strapi normally with `yarn develop`
- Execute a `GET /_health`
- See that the request is properly logged to the server console

- Run Strapi with the following environment variable `LOG_HEALTH=false yarn develop`
- Execute a `GET /_health`
- See that the request is no longer logged to the server console

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/packages/core/core/src/middlewares/logger.ts
+++ b/packages/core/core/src/middlewares/logger.ts
@@ -6,6 +6,10 @@ export const logger: Core.MiddlewareFactory = (_, { strapi }) => {
     await next();
     const delta = Math.ceil(Date.now() - start);
 
-    strapi.log.http(`${ctx.method} ${ctx.url} (${delta} ms) ${ctx.status}`);
+    if (process.env.LOG_HEALTH === 'false' && ctx.url === '/_health') {
+      return;
+    } else {
+      strapi.log.http(`${ctx.method} ${ctx.url} (${delta} ms) ${ctx.status}`);
+    }
   };
 };


alexandrebodin (commented):
This is not something we will merge right away. We need a more advanced discussion on the topic or we are going to be adding exceptions on top of exceptions in the codebase. 

Overall the best practice is to leave the source of truth alone and change the log processing instead of tweaking how logs are emited.

innerdvations:
Yeah, on second thought, this *should* be possible by adding a user logging config

See, server.js `logger.config` is merged in with the winston config on startup: 

Strapi.ts: https://github.com/strapi/strapi/blob/7736a2810c38676bfd5205a96ebc616349c1bb7f/packages/core/core/src/Strapi.ts#L249
Logger: https://github.com/strapi/strapi/blob/7736a2810c38676bfd5205a96ebc616349c1bb7f/packages/utils/logger/src/index.ts#L8

If for some reason we're missing something that prevents you from passing in a custom logger config to filter it out, let us know here and I will make sure we get it fixed.

alexandrebodin:
Hey, I'll close this as winston offer everything needed to filter out events directly etc. @Jimimimi could be interesting for you to explore using a custom log config to get json output and then do all the processing you need. It will definitly be necessary if we want to offer more options in the UI (log level filters ...)

PR #21457, opened 2024-09-24
chore: Harmonized all button sizes in settings

lucasboilly:
### What does it do?

I removed the `size="L"` and `size="M"` properties of the buttons that had them in the settings. I also added a `<Check />` icon to the "Save" button of the admin role page.

### Why is it needed?

A few pages in the settings had buttons that had a `size="L"` or a `size="M"`, but it's not supposed to be the case. All the other pages from the settings have a `size="S"`. I changed it so we have a consistent design across all pages of the settings. I added the `<Check />` icon for the same reasons.

### How to test it?

You'll need to go to the following pages:
- Settings > Review Workflow > Open any workflow
- Settings > Review Workflow > Create new workflow
- Settings > Single Sign-On
- Settings > Webhooks > Open any webhook
- Settings > Webhooks > Create new webhook
- Settings > Roles (admin panel) > Open any role
- Settings > Roles (admin panel) > Create new role
- Settings > Users > Open any user

### Related issue(s)/PR(s)

No related issue.

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Roles/CreatePage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Roles/CreatePage.tsx
@@ -11,6 +11,7 @@ import {
   TextInput,
   Typography,
 } from '@strapi/design-system';
+import { Check } from '@strapi/icons';
 import { format } from 'date-fns';
 import { Formik, Form, FormikHelpers } from 'formik';
 import { useIntl } from 'react-intl';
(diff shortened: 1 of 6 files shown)


lucasboilly:
@remidej since Madhuri recently made `size="S"` the default button size in the DS I simply removed the property on each button. Is it good or should I keep it and write "S" instead of "M" or "L"?

remidej (approved):
Looks good to me, I'll bypass the commitlint check this time and change the commit name during squash and merge

PR #21464, opened 2024-09-25
test(e2e): refactor ctb tests to enable on CI

innerdvations:
### What does it do?

- increases test timeout for all CTB tests and i18n/editview
- creates a utility for a shared test setup similar to beforeAll
  - ...but with page context which playwright beforeAll doesn't provide
  - ...which allows us to minimize the number of restarts required
  - ...which is what leads to timeouts and extra-flakiness
  - utility provides several common test setup operations like logging in, skipping ctb tour, resetting files and data
- enables all CTB tests on CI

### Why is it needed?

these tests weren't able to run on the CI because they always failed on chromium on the first run

it appears to be that they were simply timing out because they required multiple server restarts before each test

### How to test it?

See this PR pass the CI checks on the first try
- 10:15 CEST: all passed except i18, so I updated that
- 13:00 CEST: all passed
- 15:00 CEST: all passed
- 16:15 CEST: all passed

### Related issue(s)/PR(s)

DX-1516


Convly (approved):
Nice

PR #21465, opened 2024-09-25
feat(cli): add environment list command to cloud cli

gonbaum:
### What does it do?
This PR introduces the environment list command to the Strapi Cloud CLI.

Additionally, it aligns with the General CLI Guidelines RFC by removing colons as namespace separators for commands and subcommands. To support this, two new base commands, cloud and environment, have been added to enable clearer command composition.

A future refactor of existing commands to follow this updated structure should be considered.

### Why is it needed?

With the introduction of Strapi Cloud multi environments feature, we want to allow users to list their available environments from a linked project in their CLI.

### How to test it?

An endpoint is available in CLI-API. The command can be tested against staging env checking the following scenarios:

1. Non-Strapi Project:
Running the command in a non-Strapi project should fail and return an appropriate error message.
2. Unauthenticated User:
Running the command without authentication should prompt the user to log in.
3. No Available Environments:
If no environments are available, the command should return a clear message indicating that there are no environments to list.
4. Available Environments:
When environments are present, the command should display the correct information.
5. Deleted Project:
If the command is used with a deleted project, it should return a valid error and prompt the user to link the project.
6. Corrupted strapi-config.json:
If the strapi-config.json is corrupted, the command should display an error and prompt the user to link the project.
7. Command Alias:
The alias cloud environments should function identically to the full command cloud environment list.
8. Help command
Commands and subcommands suggestions should appear running `cloud help` or `cloud environment help` commands

### DIFF
--- a/packages/cli/cloud/src/environment/list/action.ts
+++ b/packages/cli/cloud/src/environment/list/action.ts
@@ -1,0 +1,52 @@
+import chalk from 'chalk';
+import type { CLIContext } from '../../types';
+import { cloudApiFactory, local, tokenServiceFactory } from '../../services';
+import { promptLogin } from '../../login/action';
+
+async function getProject(ctx: CLIContext) {
+  const { project } = await local.retrieve();
+  if (!project) {
+      ctx.logger.warn(`\nWe couldn't find a valid local project config.\nPlease link your local project to an existing Strapi Cloud project using the ${chalk.cyan(
+    'link'
+  )} command`);
+      process.exit(1);
+  }
+  return project;
+}
+
+export default async (ctx: CLIContext) => {
+  const { getValidToken } = await tokenServiceFactory(ctx);
+  const token = await getValidToken(ctx, promptLogin);
+  const { logger } = ctx;
+
+  if (!token) {
+    return;
+  }
+
+  const project = await getProject(ctx);
+  if (!project) {
+    return;
+  }
+
+  const cloudApiService = await cloudApiFactory(ctx, token);
+  const spinner = logger.spinner('Fetching environments...').start();
+
+  try {
+    const {
+      data: { data: environmentsList },
+    } = await cloudApiService.listEnvironments({ name: project.name });
+    spinner.succeed();
+    logger.log(environmentsList);
+  } catch (e: any) {
+    if (e.response && e.response.status === 404) {
+      spinner.succeed();
+      logger.warn(`\nThe project associated with this folder does not exist in Strapi Cloud. \nPlease link your local project to an existing Strapi Cloud project using the ${chalk.cyan(
+        'link'
+      )} command`);
+
+    } else {
+    ctx.logger.debug('Failed to list environments', e);
+    spinner.fail('An error occurred while fetching environments data from Strapi Cloud.');
+    }
+  }
+};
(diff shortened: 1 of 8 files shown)


packages/cli/cloud/src/environment/list/action.ts
```diff
@@ -0,0 +1,55 @@
...
+  } catch (e: any) {
+    if (e.response && e.response.status === 404) {
+      spinner.succeed();
+      logger.warn(
+        `\nThe project associated with this folder does not exist in Strapi Cloud. \nPlease link your local project to an existing Strapi Cloud project using the ${chalk.cyan(
+          'link'
+        )} command`
+      );
+    } else {
+      ctx.logger.debug('Failed to list environments', e);
```
nathan-pichon, line 51:
```suggestion
      logger.debug('Failed to list environments', e);
```

packages/cli/cloud/src/environment/list/action.ts
```diff
@@ -0,0 +1,55 @@
...
+  return project;
+}
+
+export default async (ctx: CLIContext) => {
+  const { getValidToken } = await tokenServiceFactory(ctx);
+  const token = await getValidToken(ctx, promptLogin);
+  const { logger } = ctx;
+
+  if (!token) {
+    return;
```
nathan-pichon, line 25: WDYT of adding a debug log?

packages/cli/cloud/src/environment/list/action.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const project = await getProject(ctx);
+  if (!project) {
+    return;
```
nathan-pichon, line 30: WDYT of adding a debug log?

packages/cli/cloud/src/environment/list/action.ts
nathan-pichon: I would also call the spinner.fail before the debug log, as it would mess a bit the display (not a big issue tho)

packages/cli/cloud/src/services/cli-api.ts
```diff
@@ -197,6 +207,23 @@ export async function cloudApiFactory(
...
+
+        if (response.status !== 200) {
+          throw new Error('Error fetching cloud environments from the server.');
+        }
+
+        return response;
+      } catch (error) {
+        logger.debug(
+          "🥲 Oops! Couldn't retrieve your project's environments from the server. Please try again."
+        );
```
nathan-pichon, line 222: Debug logs are for debugging, not displaying to the user there was an error. Could be great to add more technical information in this log, like error message, API endpoint, name.

packages/cli/cloud/src/environment/list/action.ts
gonbaum: We have a warning log in the getProject. This is a 'safety net' in case there is no project even after the getProject() function run (which seems very unlikely). A debug log would be correct but maybe it would not add much value in this case. Same for the token one mentioned in the comment above. Do you think that we should still add it? If we add it, we should also do it for the other command's actions since we're following the same pattern.

packages/cli/cloud/src/environment/list/action.ts
gonbaum: same answer as below

packages/cli/cloud/src/environment/list/action.ts
gonbaum: It makes sense to add logs in unusual end-of functions, but in this case, I think the log created in getProject() function already serves that purpose. I would say we either:

1- Discard the redundant check
2- Add a log before returning

For any of both decisions, we should refactor all the functions that follow the same pattern.

nathan-pichon (approved):
LGTM
Thanks

Issue #21267, opened 2024-09-16
Error `Cannot read properties of undefined (reading 'push')` when adding component to deeply nested content type in v5


tammyhart:
### Required System information

- Environment: development
- OS: darwin-arm64
- Strapi Version: 5.0.0-rc.23
- Node/Yarn Version: yarn/1.22.22 npm/? node/v20.17.0 darwin arm64
- Edition: Community
- Database: sqlite

### Describe the bug

I have a collection content type called `page`. It has a few fields and a dynamic zone called `content`.
- One of its components is deeply nested, called `block`. 
- Block has some fields as well as a repeatable component called `section`.
- Section has some fields and a repeatable component called `lists`. 
- Lists has a single field and a repeatable component called `items`. 
- Items have a single field.

This has been working beautifully and is the main reason I decided to switch this brand new project to v5.

I have tried a few times to add a component directly to `page` called `seo`. When I load the content manager after adding the component, I get an error: `Cannot read properties of undefined (reading 'push')`

The function the error is happening in is called `formatEditLayout`. The line of code with the push:
```
panels[currentPanelIndex].push(row);
```

I have added `STRAPI_ENFORCE_SOURCEMAPS=true` to my .env, but this is still the best I get.

### Expected behavior

The `seo` component should be available on the `page` content type just as it is working correctly on other content types. The Content Manager should load the `page` admin index without error.

### Additional context

I first attempted to add the `seo` component to `page` at rc.13 and got this identical error. It has continued until the current rc.23

tammyhart:
Upon further investigation, I have learned that the error isn't happening because of the deeply nested `block` component in the `content` dynamic zone. It has something to do with the order the view tries to add my fields.

The view is configured like so:
1. text field
2. UID field
3. non-repeatable component: heading
  - text field
  - text field
  - custom field
4. dynamic zone: content
5. dynamic zone: ctas
6. non-repeatable component: seo

If I move `seo` to position 4, the error no longer happens. I'm not familiar with how the view manages to output each field, component, and dynamic zone, but I'm guessing that it behaves fine when it can simply group my fields and components together and then do the dynamic zones, but behaves badly when it tries to build an interface that mixes the order of dynamic zones and components.

laurenskling:
Related to: https://github.com/strapi/strapi/issues/20938

This happens when you do _something_ in the view layout. I only get it on one of my contentTypes after migration. Does it happen to you after migration or can you reproduce it in a clean install?

[...]

PR #21483, opened 2024-09-26
fix: dynamic zones breaking layout logic

remidej:
### What does it do?

Fixes https://github.com/strapi/strapi/issues/21267

I narrowed down the problem to when a dynamic zone is the first field in the edit view layout, and there's one or more fields after.

I updated the tests so they would get the bug with the previous version of the `useDocumentLayout` code (by adding a dynamic zone at the start of the layout).

### How to test it?

Test the flow that would break before:
- create a new content type
- add a dynamic zone field first
- then add a text field (or anything else)
- then open the edit view for that content type, or the configure the view page

I think we also need some random QA to try various combinations: DZ after DZ, DZ then field then DZ, field then DZ...

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21267

### DIFF
--- a/packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
+++ b/packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
@@ -202,6 +202,12 @@ describe('useDocument', () => {
             "visible": false,
             "writable": false,
           },
+          "dz": {
+            "components": [
+              "blog.test-como",
+            ],
+            "type": "dynamiczone",
+          },
           "id": {
             "type": "string",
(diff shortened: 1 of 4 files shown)


Feranchz (approved):
Code looks good. Also did QA and everything was working but lets wait for some confirmation from community users 🚀

Magiczne:
@remidej 
According to changelogs it was releases yesterday in 5.0.2 - but when installing 5.0.2 the old code is still appearing in the library

Magiczne:
Working in 5.0.3

PR #21493, opened 2024-09-27
Config password rules users-permissions

Bassel17:
### What does it do?

- Adds a custom password validation

### Why is it needed?

- To have validations on the api level

### How to test it?

Add the config below and test if it validates

```
'users-permissions': {
    config: {
      validationRules: {
        validatePassword(value) {
          if (value.length < 8) {
            // custom error message
            throw new Error('password should be more than 8 letters');
          }

          if (value.length > 24) {
            // throws default error message 
            return false;
          }

          return true; // Validation passed
        },
      },
    },
  },
  ```

### CODE REVIEW

File: packages/plugins/users-permissions/server/controllers/validation/auth.js [hunk 1]
```diff
@@ -27,31 +32,46 @@ const forgotPasswordSchema = yup
...
-    code: yup.string().required(),
-  })
-  .noUnknown();
+const resetPasswordSchema = (config) =>
+  yup
+    .object({
+      password: yup
+        .string()
+        .required()
+        .min(config?.password?.min || 6)
```
alexandrebodin, line 41: adding this as a default would be a Breaking change

alexandrebodin (changes requested):
Could we just accept a custom yup schema or function instead of having to maintain our own abstraction of validation rules ?

packages/plugins/users-permissions/server/controllers/validation/auth.js
Bassel17: hmmmm, sure we can make not the default
we can use a custom yup schema but that would make it a breaking change if we change yup later on

alexandrebodin: Yeaha agreed, I was thinking more about a function so we can just pass it the password and the user can return true / false or throw an Error for ex

Bassel17: oh yeah even better i was thinking to add a method like this specifically if they wanted to add a specific password pattern, but yeah it would work to maintain also the min, max value too
would you suggest removing the min max, and just providing them a method ? or keeping both options

alexandrebodin: I would do just a function so we can wrap it in a yup.test()

packages/plugins/users-permissions/server/controllers/validation/auth.js
```diff
@@ -7,11 +7,22 @@ const callbackSchema = yup.object({
...
+    email: yup.string().email().required(),
+    username: yup.string().required(),
+    password: yup
+      .string()
+      .required()
+      .when([], {
+        is: () => !!config?.validatePassword,
+        then: yup
+          .string()
+          .test('password-validation', 'Password invalid', (value) =>
```
innerdvations, line 21: Should we make it async just in case someone needs it?

packages/plugins/users-permissions/server/controllers/validation/auth.js
```diff
@@ -7,11 +7,22 @@ const callbackSchema = yup.object({
...
+      .string()
+      .required()
+      .when([], {
+        is: () => !!config?.validatePassword,
+        then: yup
+          .string()
+          .test('password-validation', 'Password invalid', async (value) =>
+            config.validatePassword(value)
+          ),
+      }),
```
alexandrebodin, line 24:
```suggestion
      .test(() => {
      	if (type of config?.validatePassword) {
      		// you probably need to try catch to return a custom error message 
         	 config.validatePassword(value)
      	}
      	
      	return true
      })
```

alexandrebodin: Just made this example, but overall you can simplify and avoid the when I think

innerdvations:
@pwizla Could you please add documention for this? It's a method to add a custom validation function to users and permissions user registration. You can use the example as written in the "how to test it" here

pwizla:
Thanks for the heads-up, @innerdvations :)

I created [a small PR](https://github.com/strapi/documentation/pull/2250); did my best because I don't have a lot of context on this. It would be nice if devs could create these small PRs when little additions are made to the Strapi codebase, esp. given the Documentation team's busy roadmap ([see Slack](https://strapihq.slack.com/archives/C013U9JUX36/p1727354101645439) for details). 🙏

packages/plugins/users-permissions/server/controllers/validation/auth.js
```diff
@@ -27,31 +43,68 @@ const forgotPasswordSchema = yup
...
+        .required()
+        .oneOf([yup.ref('password')], 'Passwords do not match'),
+      currentPassword: yup.string().required(),
+    })
+    .noUnknown();
 
 module.exports = {
   validateCallbackBody: validateYupSchema(callbackSchema),
-  validateRegisterBody: validateYupSchema(registerSchema),
+  validateRegisterBody: (config) => validateYupSchema(registerSchema(config)),
```
alexandrebodin, line 104: just a semantic note

I would suggest you change the signatures following this pattern

```
validateRegisterBody: (payload, config) => validateYupSchema(createRegisterSchema(config))(payload),
```

so the `validateRegisterBody` is actually a validation function and not a factory. 
Also note the renaming of `registerSchema` to `createRegisterSchema`

alexandrebodin (approved):

alexandrebodin:
Nice job 🤝

PR #21495, opened 2024-09-27
Make locale/localizations private for non-localized cts

### LINKED ISSUE #21401: [Strapi V5] Localisations required when not enabled on collection type with GraphQL
## Bug Report

We use GraphQL in our Next.js application along with codegen to generate types based on the GraphQL schema and queries. In Strapi v5, we have a collection type where localization is disabled, but in the GraphQL schema, the `localizations` field is still marked as required.

When I include the `localizations` field in my query, it triggers the same error on the subsequent `localizations` field (possibly due to a circular dependency).

### System Information

- **Node.js version**: 18.20.4
- **NPM version**: 10.7.0
- **Strapi version**: 5.0.0
- **Database**: MySQL
- **Operating System**: macOS (darwin-arm64)
- **Project Language**: TypeScript

### Steps to Reproduce

1. Enable the GraphQL plugin in Strapi v5.
2. Create a collection type with localization disabled.

While the following steps might not be necessary to trigger the issue, they reflect our specific setup:

3. Create a component that uses the collection type as a relation.
4. Add this component to a Dynamic Zone.
5. Create a GraphQL query that queries the component in the dynamic zone
6. Run codegen in Next.js to generate `types.ts`.
7. Use the generated type in your application and build the Next.js project.
8. You will encounter an error stating that `localizations` is required, despite being disabled.

### Expected Behavior

When localization is disabled, the `localizations` field should not be required in the query.

### Screenshots

In the screenshot below, the `localizations` field is marked as required, even though it is disabled for blogs:

image

Convly:
### What does it do?

Make `locale` and `localizations` attributes private if a content-type is not localized. 

### Why is it needed?

Prevent those fields from appearing in the GraphQL schema or REST API response

### How to test it?

- Create a CT with and without i18n enabled. 
- Check the graphql queries

### Related issue(s)/PR(s)

fix #21401

### DIFF
--- a/packages/plugins/users-permissions/server/controllers/role.js
+++ b/packages/plugins/users-permissions/server/controllers/role.js
@@ -1,9 +1,18 @@
 'use strict';
 
 const _ = require('lodash');
-const { ApplicationError, ValidationError } = require('@strapi/utils').errors;
+const { async, errors } = require('@strapi/utils');
 const { getService } = require('../utils');
 const { validateDeleteRoleBody } = require('./validation/user');
+
+const { ApplicationError, ValidationError } = errors;
+
+const sanitizeOutput = async (role) => {
+  const { sanitizeLocalizationFields } = strapi.plugin('i18n').service('sanitize');
+  const schema = strapi.getModel('plugin::users-permissions.role');
+
+  return async.pipe(sanitizeLocalizationFields(schema))(role);
+};
 
 module.exports = {
   /**
@@ -30,13 +39,17 @@ module.exports = {
       return ctx.notFound();
     }
 
-    ctx.send({ role });
+    const safeRole = await sanitizeOutput(role);
+
+    ctx.send({ role: safeRole });
   },
 
   async find(ctx) {
     const roles = await getService('role').find();
 
-    ctx.send({ roles });
+    const safeRoles = await Promise.all(roles.map(sanitizeOutput));
+
+    ctx.send({ roles: safeRoles });
   },
 
   async updateRole(ctx) {
--- a/tests/api/core/content-manager/search.test.api.js
+++ b/tests/api/core/content-manager/search.test.api.js
@@ -191,7 +191,6 @@ describe('Search query', () => {
 
       expect(Array.isArray(res.body.results)).toBe(true);
       expect(res.body.results.length).toBe(data.beds.length);
-      // TODO V5: Filter out i18n fields if content type is not localized
       expect(
         res.body.results.map(omit([...CREATOR_FIELDS, 'locale', 'localizations', 'status']))
       ).toEqual(expect.arrayContaining(data.beds));
--- a/tests/api/core/content-manager/search.test.api.js
+++ b/tests/api/core/content-manager/search.test.api.js
@@ -192,9 +192,9 @@ describe('Search query', () => {
       expect(Array.isArray(res.body.results)).toBe(true);
       expect(res.body.results.length).toBe(data.beds.length);
       // TODO V5: Filter out i18n fields if content type is not localized
-      expect(res.body.results.map(omit([...CREATOR_FIELDS, 'localizations', 'status']))).toEqual(
-        expect.arrayContaining(data.beds)
-      );
+      expect(
+        res.body.results.map(omit([...CREATOR_FIELDS, 'locale', 'localizations', 'status']))
+      ).toEqual(expect.arrayContaining(data.beds));
     });
 
     test('search with special characters', async () => {
--- a/packages/plugins/i18n/server/src/register.ts
+++ b/packages/plugins/i18n/server/src/register.ts
@@ -3,6 +3,7 @@ import type { Core } from '@strapi/types';
 
 import validateLocaleCreation from './controllers/validate-locale-creation';
 import graphqlProvider from './graphql';
+import { getService } from './utils';
 
 export default ({ strapi }: { strapi: Core.Strapi }) => {
   extendContentTypes(strapi);
@@ -38,12 +39,16 @@ const addContentManagerLocaleMiddleware = (strapi: Core.Strapi) => {
  * @param {Strapi} strapi
  */
 const extendContentTypes = (strapi: Core.Strapi) => {
+  const { isLocalizedContentType } = getService('content-types');
+
   Object.values(strapi.contentTypes).forEach((contentType) => {
     const { attributes } = contentType;
 
+    const isLocalized = isLocalizedContentType(contentType);
+
     _.set(attributes, 'locale', {
       writable: true,
-      private: false,
+      private: !isLocalized,
       configurable: false,
       visible: false,
       type: 'string',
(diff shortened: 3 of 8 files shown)


alexandrebodin (commented):
I'm not sure this is what we should do here. 

Always having the fields means we have a consistent and more predictable API response & GraphQL schema and the only thing that changes is the data 🤔

Convly:
The thing is that it makes no sense to use i18n-related fields on non-localized CTs, the locale will always be null and localizations an empty array.

I just feel like it's more confusing than anything

Convly:
When D&P is disabled, the publishedAt doesn't exist, does it?

Marc-Roig:
@Convly publishedAt does exist even if D&P is disabled

Convly:
Damn, that's true, forgot about that :slightly_frowning_face:

innerdvations:
> I'm not sure this is what we should do here.
> 
> Always having the fields means we have a consistent and more predictable API response & GraphQL schema and the only thing that changes is the data 🤔

If a content type doesn't have localization, shouldn't it should throw an error if you try to access it by locale or populate localizations from the content API?

If anything, I would say the consistency should go a step deeper, and the document service should also sanitize those attributes out.

innerdvations:
> All content-types have both published at and locale so we have structural consistency.
> 
> The structure (ct, db model) should remained unchanged but the data sanitized out / or removed from the schema rather than made private. Using private to remove it from the schema is a "lucky" but very brittle consequence. It's a matter of encapsulation here. Private has a meaning that is different from "this feature should not be enabled so I will not add it to the schema or sanitize it out"

For localizations, that works fine, but for locale won't that cause us to delete locale data on startup if it gets temporarily disabled? Or do we keep that column anyway?

alexandrebodin:
What I meant with 

> Private has a meaning that is different from "this feature should not be enabled so I will not add it to the schema or sanitize it out"

is that you need to add some dedicated logic for it instead of using private :)

alexandrebodin:
> > > The structure (ct, db model) should remained unchanged but the data sanitized out / or removed from the schema rather than made private.
> > 
> > 
> > > is that you need to add some dedicated logic for it instead of using private :)
> > 
> > 
> > `private` is defined as "sanitized from the content API" so I'm not sure why we would need to create custom logic for these two fields. I think I'm misinterpreting something. At what level would you remove the attributes?
> 
> I guess that technically, we could add a custom visitor to sanitize/validate the incoming data based on i18n rules, directly from i18n, just like we do in U&P ([`develop`/packages/plugins/users-permissions/server/utils/sanitize](https://github.com/strapi/strapi/tree/develop/packages/plugins/users-permissions/server/utils/sanitize?rgh-link-date=2024-10-01T13%3A25%3A06Z)).
> 
> But then, the fields would still appear in the generated type definitions (next/etc...) since the structure hasn't changed

I was only thinking about the graphql issue with the schema generation that should check if i18n is enabled or not instead of rellying on private attributes 🤔 

It's ok to use private for the time being with a good comment for the sanitization at least. mid term this should be addressed with the split between content-types & models

packages/core/upload/admin/src/components/AssetDialog/BrowseStep/index.jsx
```diff
@@ -374,7 +374,7 @@ export const BrowseStep = ({
         </>
       )}
 
-      {pagination.pageCount > 0 && (
+      {pagination.total > pagination.pageSize && (
```
innerdvations, line 377: what's happening here (and with the other pagination changes)?

tests/api/core/content-manager/search.test.api.js
```diff
@@ -192,9 +192,9 @@ describe('Search query', () => {
       expect(Array.isArray(res.body.results)).toBe(true);
       expect(res.body.results.length).toBe(data.beds.length);
       // TODO V5: Filter out i18n fields if content type is not localized
```
innerdvations, line 194: I think you can remove this comment now

packages/core/upload/admin/src/components/AssetDialog/BrowseStep/index.jsx
Convly: uh? I just updated the branch with develop and it brought this, no idea where this comes from :thinking:

Convly: https://github.com/strapi/strapi/pull/21495/commits

innerdvations: weird, maybe you need to rebase

tests/api/core/content-manager/search.test.api.js
```diff
@@ -191,10 +191,9 @@ describe('Search query', () => {
 
       expect(Array.isArray(res.body.results)).toBe(true);
       expect(res.body.results.length).toBe(data.beds.length);
-      // TODO V5: Filter out i18n fields if content type is not localized
-      expect(res.body.results.map(omit([...CREATOR_FIELDS, 'localizations', 'status']))).toEqual(
-        expect.arrayContaining(data.beds)
-      );
+      expect(
```
innerdvations, line 194: Sorry to be annoying on this, but could we add a test for if i18n is enabled on roles and see that the fields still show up? That's the one case where this would potentially become a breaking change, since we have our own custom sanitization on this I could see someone potentially breaking it by accident in the future.

innerdvations (commented):
Are we totally sure a user wouldn't intentionally want locale to be in the results?

Do we allow someone to enable i18n on he role, for example if they want to translate their role names?

tests/api/core/content-manager/search.test.api.js
Convly: How do you want to enable i18n on the roles content type though? Or are you talking about the relation? :thinking:

Convly:
@innerdvations I've rewritten the branch history, should be cleaner now

tests/api/core/content-manager/search.test.api.js
innerdvations: My bad, I thought we let it be customized like the User ct. It looks like we don't allow it, so it's fine, someone would have to be messing with things they shouldn't

innerdvations (approved):

packages/plugins/users-permissions/server/controllers/role.js
```diff
@@ -30,13 +39,17 @@ module.exports = {
       return ctx.notFound();
     }
 
-    ctx.send({ role });
+    const safeRole = await sanitizeOutput(role);
```
alexandrebodin, line 42: Can't this use the defaultSanitizers instead of manually requiring i18n? I don't see how we need this specifically here ?

alexandrebodin (changes requested):

packages/plugins/users-permissions/server/controllers/role.js
Convly: I didn't want to change the current behavior of returning all the fields (including potentially private ones) since it wasn't present before, hence the custom sanitization.

If you have more context on what should/shouldn't be returned from this API and why, I'm all ears :sweat_smile:

alexandrebodin: I don't know, just strange to see this done here only 🤔

PR #21497, opened 2024-09-27
feat(content-releases): add stage check to releases details page

Feranchz:
### What does it do?

This PR refactors the Status column on Releases details page to add a Popover where we show both: fields validations & stage validation of each entry inside a release.

### How to test it?

1. Create a release and a Review Workflow with a required stage to publish
2. Add multiple entries to your release.
3. Check the status column is showing the expected information

### Known issue

- Icon colors are not okay. I will add a following PR on the DS for this. 
Screenshot 2024-09-27 at 13 53 43

### DIFF
--- a/packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
+++ b/packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
@@ -31,7 +31,7 @@ const StyledPopoverFlex = styled(Flex)`
 
 interface EntryValidationPopoverProps {
   action: ReleaseAction['type'];
-  schema?: Struct.ContentTypeSchema & { requiredStageToPublish?: Stage };
+  schema?: Struct.ContentTypeSchema & { stageRequiredToPublish?: Stage };
   entry: ReleaseActionEntry;
   status: ReleaseAction['status'];
 }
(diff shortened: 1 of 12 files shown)


packages/core/content-releases/server/src/services/release-action.ts
```diff
@@ -243,16 +243,32 @@ const createReleaseActionService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-        (
-          acc: { [key: ReleaseAction['contentType']]: Struct.ContentTypeSchema },
+      const workflowsService = strapi.plugin('review-workflows').service('workflows');
+
+      const contentTypeModelsMap = await contentTypeUids.reduce(
+        async (
+          accPromise: Promise<{
+            [key: ReleaseAction['contentType']]: Struct.ContentTypeSchema & {
+              requiredStageToPublish?: number;
+            };
```
Marc-Roig, line 253: could we use the type in the contract here? the one that merges the content type schema and the requriedStageToPublish, that will be less types to maintain

packages/core/content-releases/server/src/services/release-action.ts
```diff
@@ -243,16 +243,32 @@ const createReleaseActionService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+          const acc = await accPromise;
+          const contentTypeModel = strapi.getModel(contentTypeUid);
+
+          const workflow = await workflowsService.getAssignedWorkflow(contentTypeUid, {
+            populate: 'stageRequiredToPublish',
+          });
+
+          acc[contentTypeUid] = {
+            ...contentTypeModel,
+            requiredStageToPublish: workflow?.stageRequiredToPublish,
```
Marc-Roig, line 266: couldnt be a more minor comment, but could we unify the naming of `stageRequiredToPublish` and `requiredStageToPublish` ?

packages/core/content-releases/server/src/services/release-action.ts
```diff
@@ -243,16 +243,32 @@ const createReleaseActionService = ({ strapi }: { strapi: Core.Strapi }) => {
         return acc;
       }, []);
 
-      const contentTypeModelsMap = contentTypeUids.reduce(
-        (
-          acc: { [key: ReleaseAction['contentType']]: Struct.ContentTypeSchema },
+      const workflowsService = strapi.plugin('review-workflows').service('workflows');
+
+      const contentTypeModelsMap = await contentTypeUids.reduce(
+        async (
```
Marc-Roig, line 249: Just a comment,  we have the async.reduce utility that can be imported from:
```ts
import { async } from "@strapi/utils"

async.reduce(contentTypeUids)(func, {})
```

packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
```diff
@@ -0,0 +1,376 @@
...
+}: EntryValidationPopoverProps) => {
+  const { validate, isLoading } = unstable_useDocument(
+    {
+      collectionType: schema?.kind ?? '',
+      model: schema?.uid ?? '',
+    },
+    {
+      // useDocument makes a request to get more data about the entry, but we only want to have the validation function so we skip the request
+      skip: true,
+    }
```
Marc-Roig, line 332: non related question, would we benefit of discussing refactoring the useDocument hook? the skip flag looks like a shortcut that makes the hook a bit more difficult to understand

Marc-Roig:
good job on the feature 💪 did some light QA and it is working fine!

Could you add some api tests to reflect and validate this change? 🙏

packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+  return (
+    <Popover.Root>
+      <EntryStatusTrigger
+        action={action}
+        status={status}
+        hasErrors={hasErrors}
+        requiredStage={requiredStage}
+        entryStage={entryStage}
+      />
```
remidej, line 356: Why do you forward the ref in the EntryStatusTrigger definition if you don't use it?

yarn.lock
remidej, line 1: How come you have a lockfile difference 🤔

remidej (commented):
Nice job keeping the frontend status logic readable :)

packages/core/content-releases/server/src/services/release-action.ts
Feranchz: You're right, my mistake 😅

packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
Feranchz: I agree we should find what is wrong in this hook, because there are multiple places where you need to use it but is more complicated than other thing. In my opinion it should be split into two separate hooks: `useDocument` to get data about the document and `useDocumentMetadata` (or something like that) to get data about the document or utils functions about the document (for example the schema, the validation fn, etc...)

yarn.lock
Feranchz: This is because I installed qs for stringify on the content-releases' plugin. We use this same library in multiple places to handle sending the locale info on the query params

remidej: nevermind i didn't see the package.json change

PR #21500, opened 2024-09-27
test(api): enable graphl and upload tests

innerdvations:
### What does it do?

updates and re-enables tests that were skipped on v5

### Why is it needed?

they were skipped

### How to test it?

tests should pass

### Related issue(s)/PR(s)

DX-1658


Marc-Roig (dismissed):

tests/api/plugins/graphql/crud.test.api.js
```diff
@@ -346,34 +346,6 @@ describe('Test Graphql API End to End', () => {
       });
     });
 
-    test.skip('List posts with `createdBy` and `updatedBy`', async () => {
```
innerdvations, line 349: Removed because those fields aren't accessible via graphql api

tests/api/plugins/i18n/content-api/content-api.test.api.js
```diff
@@ -187,18 +171,5 @@ describe('i18n - Content API', () => {
       expect(statusCode).toBe(200);
       expect(body.data).toMatchObject(transformToRESTResource(data.homepages[1]));
     });
-
-    // TODO V5: Decide if we want to support locale=all
```
innerdvations, line 191: Removed this and the previous test because the answer to the comment is "no, only for certain methods like delete"

tests/api/plugins/graphql/crud.test.api.js
alexandrebodin: if you enable the populateCreatorFields options they should be I believe 🤔

innerdvations: ahh, I'll see if I can keep and fix that one then

innerdvations:
Sigh. All the tests pass individually but when run together they fail, so there must be a test not cleaning up after itself. I'll fix that too.

tests/api/plugins/graphql/crud.test.api.js
innerdvations: Ok, I checked and apparently it's not available in Graphql: https://docs.strapi.io/dev-docs/api/rest/guides/populate-creator-fields

Convly: I don't think we're excluding them anywhere in the GQL code though. Did you try to manually enable the "populate creator fields" and check the GQL definitions?

Convly (dismissed):
LGTM, only one concern with https://github.com/strapi/strapi/pull/21500#discussion_r1782293513

tests/api/plugins/graphql/crud.test.api.js
innerdvations: Yeah, I added populateCreatorFields to the content type in the tests and it didn't add them. So I checked the docs and that's what it said.

I can add a task to add the feature back because it does seem odd we don't have it.

innerdvations: I added a `test.todo` and created an item in the DX backlog to add (back) the feature

alexandrebodin: If it was documented, it definitely means it was intentionnal. now we need to remember why 🤔

alexandrebodin (approved):

Issue #21166, opened 2024-09-06
Error When Modifying Records Post-Migration to Strapi 5 


vinda2401:
Required System information
Node.js version: 18+
NPM version: 18.6
Strapi version: 5.0.0-rc.15
Plugin version:
Database: mysql
Operating system: Ubuntu 22.04.4 LTS

Describe the bug
We migrated  to Release Candidate 15. The migration went smoothly, and strapi is now operational.

We are preparing for the V5 migration and need to make some adjustments to the old data. But when I try to modify a record, I get this error.

To determine what is causing the problem, we searched for the document ID displayed in the warning pop up, which corresponds to the database table 'upload_folders'.

Steps to reproduce the behavior
Prerequisite: Perform migration from v4 to rc.15.

- Open a collection containing images.
- Click to open a record in editing mode. 
- Try changing any of the details, then save or publish.
- A warning pop-up will appear, as illustrated in the screenshot below.

Expected behavior
It should have saved the record, without showing error.

Screenshots

![image](https://github.com/user-attachments/assets/572c0841-9fb5-46d7-bca0-321c269007ff)

laurenskling:
Just had the same error when changing or adding a image

vinda2401:
@innerdvations @laurenskling 
This seems like an blocker for us. Is this scheduled to be resolved soon?

alexandrebodin:
@vinda2401 There hae been multiple new rc release that fixed the localization migrations. 

Make sure to migrade your app again from the original database backup when testing a new rc.

laurenskling:
@alexandrebodin just tested with rc 29, still happening.

alexandrebodin:
We are going to need some more information to reproduce this issue. Could you share a reproduction repo with a v4 dataset so we could replicate your process ?

laurenskling:
Now tested with 5.0.0 ;) 
It's happening on media _in folders_ after a migration. Outside of folders it doesn't happen. I'm trying my best to get a steady reproduction happening.

vinda2401:
@alexandrebodin 
is this getting picked up soon?

laurenskling:
seems to be fixed in 5.0.4! <3

PR #21501, opened 2024-09-27
fix: loading relations without DP fails after migrating to v5

Marc-Roig:
### What does it do?

A migration to add the publishedAt column was not setting a default value for content types without DP. 

Resulted in non DP entries to have null values on the publishedAt column. It failed when trying to load relations to such content types with an unexpected NULL publishedAt column. 

This PR:
- Fixes the migration
- Prevents this being an issue at all. Loading relations to nonDP content types do not require filtering by the publishedAt attribute

Fixes https://github.com/strapi/strapi/issues/21166

### DIFF
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-04-published-at.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-04-published-at.ts
@@ -11,8 +11,11 @@ import type { Migration } from '../common';
  */
 const createPublishedAtColumn = async (db: Knex, tableName: string) => {
   await db.schema.alterTable(tableName, (table) => {
-    table.string('published_at').defaultTo(db.fn.now());
+    table.string('published_at');
   });
+
+  // Non DP content types should have their `published_at` column set to a date
+  await db(tableName).update({ published_at: new Date() });
 };
 
 export const createdPublishedAt: Migration = {
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-04-published-at.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-04-published-at.ts
@@ -11,7 +11,7 @@ import type { Migration } from '../common';
  */
 const createPublishedAtColumn = async (db: Knex, tableName: string) => {
   await db.schema.alterTable(tableName, (table) => {
-    table.string('published_at').defaultTo(new Date().getTime());
+    table.string('published_at').defaultTo(db.fn.now());
   });
 };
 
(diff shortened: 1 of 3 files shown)


packages/core/database/src/migrations/internal-migrations/5.0.0-04-published-at.ts
```diff
@@ -11,7 +11,7 @@ import type { Migration } from '../common';
  */
 const createPublishedAtColumn = async (db: Knex, tableName: string) => {
   await db.schema.alterTable(tableName, (table) => {
-    table.string('published_at');
+    table.string('published_at').defaultTo(new Date().getTime());
```
Marc-Roig, line 14: I am not completely sure this is how I should format a date, @alexandrebodin  do you remember if we format it differently for each db?

alexandrebodin: you should be ok passing in a new Date()  directly

alexandrebodin (commented):
I'm a bit lost, 

If you don't have D&P why should the publishedAt be sth else than null ? 

Setting the defaultTo means everytime we create a row in db the publishedAt will be set to now. Is this really want we want ?

Marc-Roig:
@alexandrebodin maybe I missunderstood the knex api, what I assumed was:

- The migration supposed to add the `publishedAt` for non DP content types was not setting its value to a date and was left to null
- Hence there are users noticing issues with non DP entries with invalid publishedAt values (e.g. folders, files)
- My approach was:
	- To fix the migration by adding the default value, and I assume this default value only applies to the column creation, but not a default value that will forever be used on row creation 
	- Then, for entries with invalid publication states, I just modified the filtering to ignore `publishedAt` if DP is disabled for a content type

alexandrebodin:
> @alexandrebodin maybe I missunderstood the knex api, what I assumed was:
> 
> * The migration supposed to add the `publishedAt` for non DP content types was not setting its value to a date and was left to null

when creating the column if you set defaultTo it will set it forever so not just for the migration. You will have to create the column without defaultTo and then do an updated query instead

> * Hence there are users noticing issues with non DP entries with invalid publishedAt values (e.g. folders, files)
> * My approach was:
>   
>   * To fix the migration by adding the default value, and I assume this default value only applies to the column creation, but not a default value that will forever be used on row creation
>   * Then, for entries with invalid publication states, I just modified the filtering to ignore `publishedAt` if DP is disabled for a content type

alexandrebodin (approved):

Issue #21479, opened 2024-09-26
Strapi 5 with nginx proxy pass not loading admin panel


giorgiocolombo:
Environment: development
OS: linux-arm64
Strapi Version: 5.0.1
Node/Yarn Version: npm/10.8.2 node/v20.17.0 linux arm64 workspaces/false
Edition: Community
Database: postgres

admint.ts
```
export default ({ env }) => ({
  auth: {
    secret: env("ADMIN_JWT_SECRET"),
  },
  apiToken: {
    salt: env("API_TOKEN_SALT"),
  },
  transfer: {
    token: {
      salt: env("TRANSFER_TOKEN_SALT"),
    },
  },
  flags: {
    nps: env.bool("FLAG_NPS", true),
    promoteEE: env.bool("FLAG_PROMOTE_EE", true),
  },
  url: `${env("STRAPI_BASE_URL", "http://localhost:1337")}/admin`,
  host: env("HOST", "0.0.0.0"),
  port: env.int("PORT", 1337),
});

```
servert.ts
```
export default ({ env }) => ({
  host: env("HOST", "0.0.0.0"),
  port: env.int("PORT", 1337),
  url: env("STRAPI_BASE_URL", "http://localhost:1337"),
  app: {
    keys: env.array("APP_KEYS"),
  },
});

```

.env
`STRAPI_BASE_URL= http://localhost/strapi
`

nginx.conf
```
...
location /strapi/ {
            rewrite ^/strapi/?(.*)$ /$1 break;
            proxy_pass http://strapi:1337;
            proxy_http_version 1.1;
            proxy_set_header X-Forwarded-Host $host;
            proxy_set_header X-Forwarded-Server $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
            proxy_set_header Host $http_host;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "Upgrade";
            proxy_pass_request_headers on;
        }
...
```

with this configuration i'm only able to fetch api in my app and enter admin panel directly at localhost:1337, but **following the url at proxy the admin panel tries to reach some files at the root directory and not the proxy directory**:
![image](https://github.com/user-attachments/assets/da0f582c-e851-4cc0-8287-7e7bba3e10ca)

raphaelbadawi:
[...]

Issue #21452, opened 2024-09-24
url parameter in config/server.ts not working


Matb85:
### Required System information

- Node.js version: v20.11.0
- NPM version: yarn 1.22.2 npm 10.2.4
- Strapi version: 5.0.0
- Database: postgres
- Operating system: MacOS Sonoma 14.5
- Is your project Javascript or Typescript: Typescript

### Describe the bug

After building for production, the frontend has incorrect urls for script files

### Steps to reproduce the behavior

1. Go to set a custom url with a subpath (e.g. https:/example.com/api)in config/server.ts
2. Build and deploy the project
3. See error - the admin panel does not load, paths to scripts in the generated index.html file does not start with /api 

<img width="414" alt="Screenshot 2024-08-20 at 12 40 20" src="https://github.com/user-attachments/assets/710ca975-13a6-4d98-b606-3a7346896a25">
ff5">

### Expected behavior

The admin login view should show up

### Screenshots

<img width="414" alt="Screenshot 2024-08-20 at 12 40 20" src="https://github.com/user-attachments/assets/710ca975-13a6-4d98-b606-3a7346896a25">
ff5">

### Code snippets

config/server.ts

<img width="513" alt="Screenshot 2024-09-24 at 23 05 05" src="https://github.com/user-attachments/assets/28b87f9a-453e-4b0d-9604-2bbd128c5d64">

### Additional context

probably something to do with Vite's config

the error:

<img width="1470" alt="1" src="https://github.com/user-attachments/assets/75a1c04a-0a9d-4228-9520-9a1d2c61da3d">
<img width="402" alt="2" src="https://github.com/user-attachments/assets/0b91be7a-3851-46b3-acfb-e960d868d64c">

the path should be /api/admin instead of /admin

raphaelbadawi:
I had a similar issue after migrating to v5, but after [digging in to the documentation, it seems that you can fill your admin subpath directly in ./config/admin.ts](https://docs.strapi.io/dev-docs/admin-panel-customization/host-port-path). It seems the admin path is built on the host rather than on the server url, which makes you having to fill the subpath directly there.

raphaelbadawi:
I reproduced the same issue. The thing is, to workaround this, I added a nginx rewrite rule so all requests to /admin are redirected to to /api/admin. After this, the .js files end up being loaded properly, but the react app will still not load. No JS error nor network failure.

wladyslawczyzewski:
I think the issue is here:

[...]

Issue #21184, opened 2024-09-07
Install Strapi v5 in Sub-Path on Server not working


JWandscheer:
### Required System information

- Node.js version: v18
- Strapi version: 5.0.0-rc.17
- Database: postgres
- Operating system: Linux
- Is your project Javascript or Typescript: Typescript

### Describe the bug

Installing Strapi in a Sub-Path on a Server (f.e. https://rooturl.tld/backend) seems nearly unpossible. In a fresh install with server config url: https://rooturl.tld/backend/ and admin config admin-url: /admin (at build-time and runtime) the admin page generally loads (http 200) in /backend/admin, but the resources are loaded from the root path (https://rooturl.tld/strapi-BA4UeQmR.js). This leads to 404 and a blank page.

### Steps to reproduce the behavior

1. Build Strapi with ENV URL=https:/rooturl.tld/backend/, ENV ADMIN_URL=/admin
2. Run Container
3. Admin Page is blank because js resources could not be loaded

### Expected behavior

The resources should not ignore server-paths.

### Code snippets

We are running strapi in a kubernetes cluster and have a rewrite in our virtual service:
```
- name: "backend-routes"
    match:
    - uri:
        prefix: "/backend/"
    rewrite:
      uri: "/"
    route:
    - destination:
        host: strapi-backend
        port:
          number: 1337
```

We already tried nearly all permutations in rewriting, ENV URL/ADMIN URL setting (like ADMIN_URL=/backend/admin) etc., but the initial javascript is always requested from the root url.

menflx:
We encountered the same problem when trying to host Strapi v5.0.3 under the subpath "/cms". Tried all possible combinations of absolute und relative URLs, but the admin page stays blank because the resources are still loaded from the root url.

`config/admin.ts`:
``` typescript
export default ({ env }) => ({
  auth: {
    secret: env('ADMIN_JWT_SECRET'),
  },
  apiToken: {
    salt: env('API_TOKEN_SALT'),
  },
  transfer: {
    token: {
      salt: env('TRANSFER_TOKEN_SALT'),
    },
  },
  url: env('ADMIN_URL'),
});

```

`config/server.ts`:
``` typescript
export default ({ env }) => ({
  host: env('HOST', '0.0.0.0'),
  port: env.int('PORT', 1337),
  app: {
    keys: env.array('APP_KEYS'),
  },
  url: env('PUBLIC_URL'),
  proxy: true,
  webhooks: {
[...]

PR #21507, opened 2024-09-27
fix: server and admin url configuration

wladyslawczyzewski:
Hi. While trying to launch Strapi 5 for my new website in subpath configuration I noticed that isn't working. Turns out that others also have issues with this. I did some investigation and found that configuration is not loaded properly causing the issues when launching Strapi in subpath configuration (e.g. https://domain.com/cms).

### What does it do?

1. update the logic to load the correct `admin.path` based on `admin.url` (`admin.path` is required in build context to set the `ctx.basePath` which is used to put the resources for admin panel)
2. fix the `prefix` value for admin panel's APIs
3. fixed configuration loading, as `config` object is mutable, so using it for setting  `server.absoluteUrl` and `admin.absoluteUrl` should be done before mutating the `config` object in `loadConfiguration` function

### Why is it needed?

Strapi 5 when running in subpath of the main website domain is not working.

### How to test it?

Tested on local machine by checking the admin panel load correctly with different pair of configuration for `server.url` and `admin.url`:

| `server.url` | `admin.url` | working? |
|-------------|------------|----------|
| http://localhost:1337 | undefined | ✅ |
| http://localhost:1337 | http://localhost:1337/admin | ✅ |
| http://localhost:1337 | /admin | ✅ |
| http://localhost:1337 | /dashboard | ✅ |
| http://localhost:1337 | http://localhost:1337/dashboard | ✅ |
| http://localhost:1337 | http://localhost:1337/cms/dashboard | ❌ (to fix this case I think we need to introduce new config key `admin.path`) |
| http://localhost:1337/cms | undefined | ✅ |
| http://localhost:1337/cms | http://localhost:1337/cms/admin | ✅ |
| http://localhost:1337/cms | /admin | ✅ |
| http://localhost:1337/cms | /dashboard | ✅ |
| http://localhost:1337/cms | http://localhost:1337/cms/dashboard | ✅ |
| /cms | undefined | ✅ |
| /cms | http://localhost:1337/cms/admin | ✅ |
| /cms | /admin | ✅ |
| /cms | /dashboard | ✅ |
| /cms | http://localhost:1337/cms/dashboard | ✅ |
| /cms | http://localhost:1337/dashboard | ❌ |
| undefined | undefined | ✅ |
| undefined | http://localhost:1337/admin | ✅ |
| undefined | /admin | ✅ |
| undefined | /dashboard | ✅ |
| undefined | http://localhost:1337/dashboard | ✅ |
| undefined | http://localhost:1337/cms/dashboard  | ❌ |

### Related issue(s)/PR(s)

Related issues:

#21452 #21479 

NB. For Strapi 5 when it's run in subpath configuration (http://domain.com/cms) if you used reverse proxy like NGINX - you don't need to rewrite the request uri, just pass the request to the Strapi's instance origin/upstream.


wladyslawczyzewski:
Updated the test cases in the PR description. Looks like at this moment there's no chance to run the Strapi admin panel in different subpath not related to the `server.url`.

hanpaine:
Hi @wladyslawczyzewski thanks for your contribution - we'll take a look very soon 🙏

packages/core/core/src/configuration/urls.ts
```diff
@@ -43,20 +43,16 @@ export const getConfigUrls = (config: Record<string, unknown>, forAdminBuild = f
...
+  // this could be just /admin, or custom one like /dashboard, or subpath like /strapi/dashboard
   let adminPath = adminUrl;
-  if (
-    serverUrl.startsWith('http') &&
-    adminUrl.startsWith('http') &&
-    new URL(adminUrl).origin === new URL(serverUrl).origin &&
-    !forAdminBuild
-  ) {
-    adminPath = adminUrl.replace(strings.getCommonPath(serverUrl, adminUrl), '');
-    adminPath = `/${_.trim(adminPath, '/')}`;
```
Convly, line 58: What's the main reason behind removing this bit?

Convly (commented):
Only one question, the rest LGTM

packages/core/core/src/services/server/admin-api.ts
```diff
@@ -3,7 +3,11 @@ import { createAPI } from './api';
 
 const createAdminAPI = (strapi: Core.Strapi) => {
   const opts = {
-    prefix: '', // '/admin';
+    // prefix is the base URL of the server api
+    // it's relative to the server host
+    // if we run the server as / (root) and admin panel on /admin it should be / (empty string)
+    // if we run the server as subpath like /strapi and admin panel as a subpath like /strapi/admin it should be /strapi
+    prefix: strapi.config.get<string>('admin.path').replace(/\/[^\/]+\/?$/, ''), // remove last part of the path
```
Convly, line 10: Not so sure about that though.

Shouldn't this be the responsibility of the nginx (& co) to strip the `/strapi` (sub domain) part from requests & keep the `server.url` pointing to the API root endpoint?

innerdvations:
Thanks for the PR and even moreso for the detailed information! I'll check this out and hopefully we will get it merged soon.

In the meantime, could you fix the reported linting errors?

```
  /home/runner/work/strapi/strapi/packages/core/core/src/configuration/urls.ts
    2:10  error  'strings' is defined but never used  @typescript-eslint/no-unused-vars

  /home/runner/work/strapi/strapi/packages/core/core/src/services/server/admin-api.ts
    10:66  error  Unnecessary escape character: \/  no-useless-escape
```

packages/core/core/src/configuration/urls.ts
wladyslawczyzewski: @Convly thanks for this question! this is related to the issue of loading admin panel's resources, when we setup the configuration is such way that admin panel is in subpath to the strapi instance (ie. `https://domain.fake/strapi` and `https://domain.fake/strapi/admin`) the code above will remove the `https://domain.fake/strapi` from the `adminPath`. Later in the code, this variable is used to set the `ctx.basePath` which is used to put the links to the resources in admin panel causing the issue loading it because it will be looking for resources at `https://domain.fake` instead of `https://domain.fake/strapi`

packages/core/core/src/services/server/admin-api.ts
wladyslawczyzewski: @Convly this is similar issue to the one I answered above – when running strapi instance as subpath of main domain, the admin panel will be sending requests to admin api to correct place but admin api itself will not be bonded to correct url, so using `admin.path` configuration here we can bind the admin api to correct address mitigating the issue (ie. admin api should listen `/strapi` instead of `/`)

wladyslawczyzewski:
hi @innerdvations , thanks for your review – I've fixed the lint errors, could you please check one more time?

packages/core/core/src/services/server/admin-api.ts
alexandrebodin: hi @wladyslawczyzewski we never allowed doing this in the past so that the proxy servers were responsible for url rewriting (basically removing the /strapi in your example) from the calls. this allowed for more flexibility in v4 and worked fine. Could you start with just the changes that make it work like in v4 I would like to consider the consequences of manually changing the server prefixes 🙏

wladyslawczyzewski: hi @alexandrebodin I'm not familiar with strapi v4 implementation details but I don't think that reverting this change is possible as all these three issues are tightly coupled to run the strapi with subpath configuration. I will do more investigation early next week as currently I don't have time to work on this and will get back to you.

in the meantime if you get thoughts or will be able to do the investigation by yourself, please post the update here so we will be on the same page. thanks in advance!

hanpaine:
@innerdvations looks like this could be an issue on V4 too - can you check if so, and dupe to v4 please?

innerdvations:
> @innerdvations looks like this could be an issue on V4 too - can you check if so, and dupe to v4 please?

I think @alexandrebodin and @Convly are sorting out the discrepencies between the two, but if not then I'll do it

raphaelbadawi:
The fix works fine on production for me, but I noticed that the media library assets links are broken in the admin panel context. Not sure why it does that, for now I workaround this with a nginx rewrite rule (which is weird since I don't need rewrite rules for the admin panel itself).

derrickmehaffy:
> The fix works fine on production for me, but I noticed that the media library assets links are broken in the admin panel context. Not sure why it does that, for now I workaround this with a nginx rewrite rule (which is weird since I don't need rewrite rules for the admin panel itself).

That's probably the security middleware csp settings, you'll need to set those.

mad-zephyr:
Hello, colleagues. Sorry for bother you. Could someone kindly give updates related to this bug fix, any plans to merge it? As far as I know it almost ready for merging.

Matb85:
Hello, any updates on when this will be merged?

alexandrebodin:
Hello all, 

To share some context, 

Right now this PR cannot be merged. It does seem to work but it does so by adding internal prefixing, which creates unexpected behaviors. The goal in the short term is to offer a way to make strapi work behind a reverse-proxy like we supported in v4. 

To be fully transparent we have a dedicated workshop tomorrow on it 🙏

Moonlight63:
I wonder how that's going.

Convly:
Hey, we took some time to go over this with @alexandrebodin and test the different scenario + decide where we wanted to go.

1st, you can find below a table you shared in your original PR that we completed (on top of the result for your PR) with the behavior in v4 (pre vite update) + the behavior when using Alex's PR.

| `server.url` | `admin.url` | Community PR | v4.16 (❌ <=> requires reverse proxy) | PR Alex |
| --- | --- | --- | --- | --- |
| [http://localhost:1337](http://localhost:1337/) | undefined | ✅ | ✅ | ✅ |
| [http://localhost:1337](http://localhost:1337/) | http://localhost:1337/admin | ✅ | ✅ | ✅ |
| [http://localhost:1337](http://localhost:1337/) | /admin | ✅ | ✅ | ✅ |
| [http://localhost:1337](http://localhost:1337/) | /dashboard | ✅ | ✅ | ✅ |
| [http://localhost:1337](http://localhost:1337/) | http://localhost:1337/dashboard | ✅ | ✅ | ✅ |
| [http://localhost:1337](http://localhost:1337/) | http://localhost:1337/cms/dashboard | ❌  | ✅ | ✅ |
| http://localhost:1337/cms | undefined | ✅ | ❌ | ❌ (vite serves too many things / TBF later) |
| http://localhost:1337/cms | http://localhost:1337/cms/admin | ✅ | ❌ | same as above |
| http://localhost:1337/cms | /admin | ✅ | ❌ | same as above |
| http://localhost:1337/cms | /dashboard | ✅ | ❌ | same as above |
| http://localhost:1337/cms | http://localhost:1337/cms/dashboard | ✅ | ❌ | same as above |
| /cms | undefined | ✅ | ❌ | same as above |
| /cms | http://localhost:1337/cms/admin | ✅ | ❌ | same as above |
| /cms | /admin | ✅ | ❌ | same as above |
| /cms | /dashboard | ✅ | ❌ | same as above |
| /cms | http://localhost:1337/cms/dashboard | ✅ | ❌ | same as above |
| /cms | http://localhost:1337/dashboard | ❌ | ❌ | same as above |
| undefined | undefined | ✅ | ✅ | ✅ |
| undefined | http://localhost:1337/admin | ✅ | ✅ | ✅ |
| undefined | /admin | ✅ | ✅ | ✅ |
| undefined | /dashboard | ✅ | ✅ | ✅ |
| undefined | http://localhost:1337/dashboard | ✅ | ✅ | ✅ |
| undefined | http://localhost:1337/cms/dashboard | ❌ | ✅ | ✅ |

As you can see, your PR allows to handle far more possibilities while Alex's make it ISO with what we had in v4.

After discussing it, we came to the conclusion that our goal with setting server URLs is to match reverse proxies configurations rather than automatically modifying how we serve our APIs.

Handling the proxying ourselves could cause breaking change issues for plugin maintainers depending on constant endpoints as well as for users, whereas delegating the server URL redirection to a reverse proxy allows for more flexibility.

To summarize, we will move forward with Alex's PR for now to keep consistency & flexibility with v4, but we're open to discuss a feature request around setting the server URL & handling the paths directly in Strapi.

Below I'm sharing a very simple caddy file that we used for our tests & that create a reverse proxy on localhost:1338/cms -> localhost:1337.

```text
:1338

handle_path /cms/* {
	uri strip_prefix /cms
	reverse_proxy localhost:1337
}

respond "Not Found" 404

log logger {
	output stdout
	format console
}

```

wladyslawczyzewski:
hi @Convly thanks for detailed explanation. I see your point and I totally trust you in such decisions as you're both the maintainers of this project and you know better this project then me (it was actually my first time I ran strapi).

Using reverse proxy is a quite good solution and I think it's good option for the most use cases, however, there're situations where the reverse proxy can't do url rewrites, for instance, AWS' Load Balancers - so there's no option to run the strapi in subpath and subdomain installation is only way to get that work.

Feel free to close this PR as soon as you decide the issue is solved.

Convly:
Interesting, I haven't used AWS ALBs in a long time, but isn't it possible to place a reverse proxy in between your ALB & your application server?
Otherwise, I believe it's possible to do full redirects for simple use cases

In the meantime, I'll close this PR just so that it's removed from our backlog. Thanks for your contribution, greatly appreciated!

alexandrebodin:
@wladyslawczyzewski definitly make a feature request with the explanation of your usecase that won't work with the reverse proxy so the product team can see it 🙏

Moonlight63:
I guess I am confused about how to make a correct configuration for serving the admin panel on a different subdomain from the api server given the double / bug that this PR seems to have partially fixed. I have been trying for a few days to off and on to get my strapi instance to server at api.example.com, and put the admin panel on admin.example.com, but it seems like this is not possible because that requires the admin panel to be pointing to root /, which doesn't work. If I host the admin panel at admin.example.com/admin, then everything is fine of course, but that is not the goal. Are we just saying flat out that this use case or setup is not and will not be supported/possible? I am doing this using a "different servers" config, hosting the built admin SPA in a simple nginx server.

alexandrebodin:
> I guess I am confused about how to make a correct configuration for serving the admin panel on a different subdomain from the api server given the double / bug that this PR seems to have partially fixed. I have been trying for a few days to off and on to get my strapi instance to server at api.example.com, and put the admin panel on admin.example.com, but it seems like this is not possible because that requires the admin panel to be pointing to root /, which doesn't work. If I host the admin panel at admin.example.com/admin, then everything is fine of course, but that is not the goal. Are we just saying flat out that this use case or setup is not and will not be supported/possible? I am doing this using a "different servers" config, hosting the built admin SPA in a simple nginx server. Edit: It's even more irritating given that I could just set a full domain name as the url for the admin panel, which would solve the issue entirely, but the fqdn gets parsed out to a relative path anyway so it doesn't work.

We are only prioritizing another PR not leaving the issue open. We shared the experimental version in the original issue you can find it here for testing. It should work like the v4 with a reverse proxy configuration. https://github.com/strapi/strapi/issues/21184#issuecomment-2490644833

alexandrebodin:
For any further discussion please go to https://github.com/strapi/strapi/issues/21184 so we have a single source of truth for the issue 🙏

PR #21533, opened 2024-09-30
Fix Various Document Service Type Issues

### LINKED ISSUE #20735: v5: typing errors in various places

Convly:
### What does it do?

- Add the possibility to filter on `documentId` (root and nested filters)
- Added back the support for the relation re-ordering API in create/update data payloads

_I've also split the attribute utils (document service's params) into different files to make it less bloated._

### Why is it needed?

Fix incorrect behaviors in the document service types

### How to test it?

Reproduce the queries from https://github.com/strapi/strapi/issues/20735#issuecomment-2380452010 using any application and similar content types

### Related issue(s)/PR(s)

fix #20735

### DIFF
--- a/packages/core/types/src/modules/documents/params/attributes/index.ts
+++ b/packages/core/types/src/modules/documents/params/attributes/index.ts
@@ -1,8 +1,18 @@
-import type * as Schema from '../../../schema';
-import type * as Data from '../../../data';
+import type * as Schema from '../../../../schema';
 
-import type * as UID from '../../../uid';
-import type { Array, Constants, Object, If, Extends, MatchFirst, IsNotNever } from '../../../utils';
+import type * as UID from '../../../../uid';
+import type {
+  Array,
+  Constants,
+  Object,
+  If,
+  Extends,
+  MatchFirst,
+  IsNotNever,
+} from '../../../../utils';
+
+import type { ID, DocumentID } from './id';
+import type { OmitRelationsWithoutTarget, RelationInputValue } from './relations';
 
 export type NonFilterableKind = Extract<Schema.Attribute.Kind, 'password' | 'dynamiczone'>;
 export type FilterableKind = Exclude<Schema.Attribute.Kind, NonFilterableKind>;
@@ -22,8 +32,6 @@ export type GetNestedKeys<TSchemaUID extends UID.Schema> = Exclude<
   Schema.AttributeNamesWithTarget<TSchemaUID>,
   GetNonFilterableKeys<TSchemaUID>
 >;
-
-export type ID = Data.ID;
 
 export type BooleanValue = boolean | 'true' | 'false' | 't' | 'f' | '1' | '0' | 1 | 0;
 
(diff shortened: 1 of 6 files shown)


packages/core/types/src/modules/documents/params/attributes/index.ts
```diff
@@ -82,14 +91,6 @@ export type GetValues<TSchemaUID extends UID.Schema> = {
   }
```
Marc-Roig, line 91: Just so I understand, Schema attributes only contain attributes strictly listed in the schema.json file ? and not attributes such as the id or document id?

packages/core/types/src/modules/documents/params/attributes/relations.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type WithPositionArguments<T> = T & { position?: PositionalArguments };
+
+type Set = { set: ShortHand[] | LongHand[] | null };
+type Connect = { connect: ShortHand[] | WithPositionArguments<LongHand>[] };
+type Disconnect = { disconnect: ShortHand[] | LongHand[] };
+
+type FullUpdate = Set;
+type PartialUpdate = Partial<Connect & Disconnect>;
```
Marc-Roig, line 27: I believe this is a copy of what we have in the entity-service types too, just so I know for the future, is the plan to maintain both at the same time? or we plan to unify them?

packages/core/types/src/modules/documents/params/attributes/index.ts
Convly: Schema attributes are all the attributes present in a schema once loaded in Strapi, which doesn't include `id` or `documentId`

packages/core/types/src/modules/documents/params/attributes/relations.ts
Convly: Yep it is, the plan is to only maintain the DS one, the ES is deprecated and won't receive updates.

Marc-Roig (approved):

innerdvations (approved):
LGTM, tested and fixes the issue!

Issue #21471, opened 2024-09-25
New Strapi 5 project with EE and mysql/mariadb fails to start


innerdvations:
### Required System information

Launched In: 666 ms
Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0
Node/Yarn Version: yarn/4.5.0 npm/? node/v20.9.0 darwin arm64
Edition: Community
Database: mysql

### Describe the bug

Creating a new Strapi project with EE and mariadb or mysql fails to start the project with this error:

![image](https://github.com/user-attachments/assets/30940428-ab71-4d46-b272-4be65711f08b)

### Steps to reproduce the behavior

1. create a strapi 5 app
2. set an EE license key before starting it up the first time
3. start up strapi

### Expected behavior

Strapi should start up with no issues

### Workaround

- Disable EE
- start up strapi
- it creates all the initial tables it needs
- shut down strapi
- Enable EE

### Screenshots

![image](https://github.com/user-attachments/assets/30940428-ab71-4d46-b272-4be65711f08b)

### Code snippets

If applicable, add code samples to help explain your problem.

derrickmehaffy:
Ben and I both confirmed this on MariaDB only when using an EE license

yukimasaki:
I am using Strapi's Comunity Edition and I get the same message with MariaDB 10.6.

> Transaction was implicitly committed, do not mix transactions and DDL with MySQL (#805)

I can access the admin panel with `yarn develop`, but not with `yarn start`.
Chrome's developer tools shows the following error message.

![image](https://github.com/user-attachments/assets/d97cc2be-fb20-422c-8c88-52c8c672fa52)

Should I create a new Issue on this issue?

hanpaine:
Hi @yukimasaki thank you for flagging - it would be great if you could raise another issue with as much further information as possible, yes please 🙏 cheers!

PR #21542, opened 2024-10-01
fix: Create table if it doesn't exist

butcherZ:
### What does it do?
- check if table exists
- create table if it doesn't exist already

### Why is it needed?

issues described here https://github.com/strapi/strapi/issues/21471

### How to test it?

- create a strapi 5 app with mysql or mariadb

- set an EE license key before starting it up the first time

- start up strapi, project should be launched without seeing errors.

### Related issue(s)/PR(s)

#21471 

DX-1651

### DIFF
--- a/packages/core/content-releases/server/src/migrations/database/5.0.0-document-id-in-actions.ts
+++ b/packages/core/content-releases/server/src/migrations/database/5.0.0-document-id-in-actions.ts
@@ -14,19 +14,10 @@ export const addEntryDocumentToReleaseActions: Migration = {
   name: 'content-releases::5.0.0-add-entry-document-id-to-release-actions',
   async up(trx: Knex, db: Database) {
     // Check if the table exists
-    const tableExists = await trx.schema.hasTable('strapi_release_actions');
+    const hasTable = await trx.schema.hasTable('strapi_release_actions');
 
-    if (!tableExists) {
-      // Create the table if it doesn't exist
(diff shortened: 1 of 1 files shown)


packages/core/content-releases/server/src/migrations/database/5.0.0-document-id-in-actions.ts
```diff
@@ -13,6 +13,22 @@ type Knex = Parameters<Migration['up']>[0];
...
+        table.increments('id').primary();
+        table.string('type');
+        table.string('content_type');
+        table.string('entry_document_id');
+        table.string('locale');
+        table.boolean('is_entry_valid');
+        table.integer('release_id').unsigned();
+      });
+    }
+
```
innerdvations, line 31: I think you can actually just have it return rather than creating the table. If the table doesn't exist, there is nothing to migrate. And then Strapi will create the correct table structure on startup.

butcherZ: oh I see! will do! :D

innerdvations (approved):
LGTM, tested and works fine. @Marc-Roig do you see any potential issues with doing this? AFAIK there's nothing to migrate and Strapi will create the table correctly on normal startup, but I might be missing something.

Marc-Roig:
looks completely fine to me!

Convly (approved):

Issue #21338, opened 2024-09-18
countDraftRelations crashes 400 when containing relations to non-DP types


laurenskling:
### Required System information

- Node.js version: 20
- NPM version: yarn 1.22
- Strapi version: 5.0.0 (and rc 29)
- Database: sqlite
- Operating system: macos
- Is your project Javascript or Typescript: js

### Describe the bug

When going to a entry which has relations to content types without draftAndPublish, the countDraftRelations API call crashes and throws errors.

### Steps to reproduce the behavior

1. create a type with a relation to a non-DP type (in my screenshots, searchResultsPage is a relation to a non-DP type)
2. create and save, or open an existing entry
3. See error

### Expected behavior

no error.

### Screenshots

<img width="584" alt="Scherm­afbeelding 2024-09-18 om 19 24 13" src="https://github.com/user-attachments/assets/c221eb1a-2d20-4a05-82a6-7862fadd65cb">
<img width="1020" alt="Scherm­afbeelding 2024-09-18 om 19 24 24" src="https://github.com/user-attachments/assets/6ec79c98-77b7-48d4-abe2-52c598783e2b">
<img width="829" alt="Scherm­afbeelding 2024-09-18 om 19 24 35" src="https://github.com/user-attachments/assets/3b0ff09c-2819-4b61-8fe2-2981d81c75d5">

laurenskling:
If you want to see it in action:

https://github.com/strapi/strapi/issues/21354
 https://discord.com/channels/811989166782021633/1095091586452426824/1286021240842223678

alexandrebodin:
Hey @laurenskling I think we fixed that before stable. Do you confirm ?

laurenskling:
@alexandrebodin this happened to me when I installed 5.0.0 on my actual codebase. Even with clean DB.
Will try to make a good reproduction a.s.a.p., because it does not seem to happen on my clean installed example project. It also doesn't happen on all relation. Will need to look into it why it happens on my searchResultsPage...

alexandrebodin:
Thank you 🙏

laurenskling:
@alexandrebodin Step closer! Just realized, these content types doing this have a field called `filters`!

If I rename them, the error is gone. Is `filters` a key we should not use? I think it's conflicting in the code of countDraftRelations on actual document filters.

alexandrebodin:
Wow nice find. it will help see if we can fix the code or if we need to prevent the use of this name 🙏  You rock!

[...]

PR #21547, opened 2024-10-01
feat: fix filters attribute name

Marc-Roig:
### What does it do?

As seen in https://github.com/strapi/strapi/issues/21338 , attribute names called `filters` broke the populate validation.
That's cause it was confusing a "filter" inside a populate with the filters attribute name:
```ts
{
  populate: {
    foo: {
      filters: { bar: { $eq: 1 } }  // Should filter entries where bar is 1
    }
  }
}

{
  populate: {
    foo: {
      populate: {
        filters: true  //  Should populate the "filters" attribute
      }
    }
  }
}
```

The `traverseQueryPopulate` was assuming `filters` was an attribute, as it was an actual attribute of the schema. Hence the first query was always invalid.

This PR solves this by modifying the `traverseQueryPopulate` and tracking each traversed node parent.

### DIFF
--- a/packages/core/utils/src/__tests__/query-populate.test.ts
+++ b/packages/core/utils/src/__tests__/query-populate.test.ts
@@ -152,7 +152,7 @@ describe('traverseQueryPopulate', () => {
       getModel: strapi.getModel,
     } as const;
 
-    await traverseQueryPopulate(({ key, schema, path, parent, attribute }) => {
+    await traverseQueryPopulate(({ key, parent }) => {
       switch (key) {
         case 'address':
           // top level populate should not have parent
--- a/packages/core/utils/src/__tests__/query-populate.test.ts
+++ b/packages/core/utils/src/__tests__/query-populate.test.ts
@@ -152,17 +152,18 @@ describe('traverseQueryPopulate', () => {
       getModel: strapi.getModel,
     } as const;
 
-    await traverseQueryPopulate(({ key, parent }) => {
+    await traverseQueryPopulate(({ key, parent, attribute }) => {
       switch (key) {
         case 'address':
           // top level populate should not have parent
           expect(parent).toBeUndefined();
+          expect(attribute).toBeDefined();
           break;
-
         case 'filters':
           // Parent information should be available
           expect(parent.key).toBe('address');
           expect(parent.attribute).not.toBeUndefined();
+          expect(attribute).toBeDefined();
           break;
         default:
           break;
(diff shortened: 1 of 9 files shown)


Marc-Roig:
@alexandrebodin @innerdvations  @Convly I think this might be easier if we spend 15 minutes on discord discussing this ticket together 🤞

Convly:
Agree, lots of things to say :joy:

Marc-Roig:
@Convly @alexandrebodin if we agree this is what we want to merge, could we have a last look and proceed with it? 🙏

packages/core/utils/src/traverse/query-populate.ts
```diff
@@ -24,53 +24,63 @@ const isKeyword = (keyword: string) => {
...
   isArray(value) && value.every(isString);
 
 const isObj = (value: unknown): value is Record<string, unknown> => isObject(value);
 
 const populate = traverseFactory()
-  // Array of strings ['foo', 'foo.bar'] => map(recurse), then filter out empty items
+  .intercept(isPopulateString, async (visitor, options, populate, { recurse }) => {
+    const populateObject = pathsToObjectPopulate([populate]);
+    const traversedPopulate = (await recurse(visitor, options, populateObject)) as PopulateObject;
+    const [result] = objectPopulateToPaths(traversedPopulate);
```
Convly, line 42: Could we add some context to this bit to make sure it's not altered by future contributors? Basically explaining why we have to transform in & out the strings

packages/core/utils/src/__tests__/query-populate.test.ts
```diff
@@ -104,4 +104,75 @@ describe('traverseQueryPopulate', () => {
...
+          break;
+
+        case 'filters':
+          // Parent information should be available
+          expect(parent.key).toBe('address');
+          expect(parent.attribute).not.toBeUndefined();
+          break;
+        default:
+          break;
+      }
```
Convly, line 169: Could we also add an expect on the attribute itself? To make sure both have the correct values?

```suggestion
    await traverseQueryPopulate(({ key, parent, attribute }) => {
      switch (key) {
        case 'address':
          // top level populate should not have parent
          expect(parent).toBeUndefined();
          expect(attribute).toBeDefined();
          break;

        case 'filters':
          // Parent information should be available
          expect(parent.key).toBe('address');
          expect(parent.attribute).not.toBeUndefined();
          expect(attribute).toBeDefined();
          break;
        default:
          break;
      }
```

Convly (commented):
Seems like unit tests are failing, the rest lgtm

Convly (approved):

PR #21549, opened 2024-10-01
Add Back the Deprecated Shared Populate Option for Morph Links

Convly:
### What does it do?
Introduces (add back) a deprecated method for populating second-level links in polymorphic structures using a shared `populate`.

It also includes:
- the addition of the `CountClause` type
- updates for handling nested population queries more robustly
- introduces the `Pretty` type for enhancing readability of complex type intersections

### Why is it needed?

Ease the migration to v5 while we work on a better solution to allow pattern matching during populate ops.

Allow developers to populate polymorphic links' relations using a wildcard (`'*'`), although this method will be deprecated in the next major.

### How to test it?
- Verify that the deprecated `populate` method works with polymorphic links using a wildcard.
- Ensure that the `CountClause` type is applied correctly in populate clauses (should only be there for regular relations or medias, not components & morphs)
- Check that nested population queries within polymorphic attributes are validated properly.

### DIFF
--- a/packages/core/types/src/modules/documents/params/populate.ts
+++ b/packages/core/types/src/modules/documents/params/populate.ts
@@ -1,7 +1,17 @@
 import type * as Schema from '../../../schema';
 
 import type * as UID from '../../../uid';
-import type { Constants, Guard, If, And, DoesNotExtends, IsNotNever, XOR } from '../../../utils';
+import type {
+  Constants,
+  Guard,
+  If,
+  And,
+  DoesNotExtends,
+  IsNotNever,
+  XOR,
+  Intersect,
+  Extends,
+} from '../../../utils';
 
 import type { Params } from '..';
 
@@ -67,9 +77,31 @@ type PopulateClause<
   TSchemaUID extends UID.Schema,
   TKeys extends Schema.PopulatableAttributeNames<TSchemaUID>,
 > = {
-  [TKey in TKeys]?:
-    | boolean
-    | NestedParams<Schema.Attribute.Target<Schema.AttributeByName<TSchemaUID, TKey>>>;
+  [TKey in TKeys]?: Schema.Attribute.Target<
+    Schema.AttributeByName<TSchemaUID, TKey>
+  > extends infer TTarget extends UID.Schema
+    ?
+        | boolean
+        | Intersect<
+            [
+              NestedParams<TTarget>,
+              // Only add the count clause to content types links, ignore components
+              If<Extends<TTarget, UID.ContentType>, CountClause, unknown>,
+            ]
+          >
+    : never;
+};
+
+type CountClause = { count?: boolean };
+
+// TODO: Remove support in v6
+type DeprecatedSharedPopulateClauseForPolymorphicLinks = {
+  /**
+   * Allows the population of all first level links in a polymorphic link by using a wildcard.
+   *
+   * @deprecated The support is going to be dropped in Strapi v6
+   */
+  populate?: WildcardNotation;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


Convly:
**Notes**

@alexandrebodin / @innerdvations this is the first step we talked about regarding easing the populate for the fragment API. We'll need to create an RFC before moving forward on the more complete solution

@Marc-Roig I've added you just because it touches to the document API type surface too (notably the count clause). You're welcome to make sure this is the correct implementation.

Convly:
Somehow the build fails because of a data transfer (build) error, although the IDE (vscode, webstorm) can't catch it :melting_face:

packages/core/types/src/modules/documents/params/populate.ts
```diff
@@ -92,31 +124,39 @@ export type ObjectNotation<TSchemaUID extends UID.Schema> = [
...
         Constants.AreSchemaRegistriesExtended,
         // If TSchemaUID === UID.Schema, then ignore it and move to loose types
-        // Note: Currently, this only ignores TSchemaUID when it is equal to UID.Schema, it won't work if UID.(ContentType|Component) is passed as parameter
+        // Note: Currently, this only ignores TSchemaUID when it's equal to UID.Schema, it won't work if UID.(ContentType|Component) is passed as parameter
         DoesNotExtends<UID.Schema, TSchemaUID>
       >,
       // Populatable keys with a target
       | If<IsNotNever<TKeysWithTarget>, PopulateClause<TSchemaUID, TKeysWithTarget>>
-      // Populatable keys with either zero or multiple target(s)
+      // Populatable keys with either zero or multiple target
```
innerdvations, line 132:
```suggestion
      // Populatable keys with either zero or multiple targets
```

packages/core/types/src/utils/index.ts
```diff
@@ -205,3 +205,32 @@ export type PartialWithThis<T> = Partial<T> & ThisType<T>;
...
+ *
+ * type C = A & B;
+ * ^ {
+ *     foo: string;
+ *     bar: number;
+ *     baz: boolean;
+ *   }
+ * ```
+ */
+export type Pretty<T> = { [K in keyof T]: T[K] } & unknown;
```
innerdvations, line 236: Could this have any unintended side effects? Is the `& unknown` necessary?

packages/core/utils/src/convert-query-params.ts
```diff
@@ -354,7 +358,7 @@ const createTransformer = ({ getModel }: TransformerOptions) => {
         try {
           const subPopulateAsBoolean = parseType({ type: 'boolean', value: subPopulate });
           // Only true is accepted as a boolean populate value
-          return subPopulateAsBoolean === true ? { ...acc, [key]: true } : acc;
+          return subPopulateAsBoolean ? { ...acc, [key]: true } : acc;
```
innerdvations, line 361: Is this necessary? Shouldn't parseType() always return a boolean? And if it's '*' won't it evaluate to false (assuming that's the reason for the change)

innerdvations (commented):
LGTM, haven't tested it yet but will try to do that later this morning

packages/core/types/src/utils/index.ts
Convly: https://www.totaltypescript.com/concepts/the-prettify-helper

packages/core/utils/src/convert-query-params.ts
Convly: The reason for the change was to make lints happy

At this point it should always be a boolean (as it would throw if the string wasn't a valid bool representation, thus not reaching this point)

packages/core/types/src/utils/index.ts
alexandrebodin: haaa it finally makes its appearance :D

Convly: Incrementally trying it out in small areas to make sure it doesn't cause regression on other types/features

packages/core/utils/src/convert-query-params.ts
```diff
@@ -386,11 +390,33 @@ const createTransformer = ({ getModel }: TransformerOptions) => {
           );
         }
 
+        /**
+         * Validate nested population queries in the context of a polymorphic attribute (dynamic zone, morph relation).
+         *
+         * If 'populate' exists in subPopulate, its value should be constrained to a wildcard ('*').
+         *
+         * TODO: This approach should be marked as deprecated and replaced with a more robust solution in future versions.
```
alexandrebodin, line 398: why a todo ? can't you add a log now ?

packages/core/utils/src/convert-query-params.ts
```diff
@@ -386,11 +390,33 @@ const createTransformer = ({ getModel }: TransformerOptions) => {
...
+            `Invalid nested population query detected. When using 'populate' within polymorphic structures, ` +
+              `its value must be '*' to indicate all second level links. Specific field targeting is not supported here. ` +
+              `Consider using the fragment API for more granular population control.`
+          );
+        }
+
         const newSubPopulate = {};
 
-        // { on: { ... } }
+        // case: { populate: '*' }
```
alexandrebodin, line 411: Should we prevent a mix of the 3 params ? so it's only a single one ?

packages/core/types/src/schema/attribute/definitions/component.ts
```diff
@@ -47,4 +47,4 @@ export type GetComponentValue<TAttribute extends Attribute.Attribute> =
     : never;
 
 export type ComponentTarget<TAttribute extends Attribute.Attribute> =
-  TAttribute extends Component<infer TComponentUID> ? TComponentUID : never;
+  TAttribute extends Component<infer TComponentUID, Constants.BooleanValue> ? TComponentUID : never;
```
alexandrebodin, line 50: can you remind me what that second param does ?

packages/core/utils/src/convert-query-params.ts
Convly: By log you mean adding a server log every time a `populate='*' `is used for a DZ? I fear it could spam the logs too much :thinking:

packages/core/types/src/schema/attribute/definitions/component.ts
Convly: It's used to indicate whether the component attribute is repeatable or not.
(https://github.com/strapi/strapi/blob/develop/packages/core/types/src/schema/attribute/definitions/component.ts#L18)

In this case, `TAttribute` was not matching the condition for repeatable components because the default value for `TRepeatable` is `false`, thus comparing `Component<UID, true>` and `Component<UID, false>` (always false).

By adding the `Constants.BooleanValue`, we're forcing the comparison between `Component<UID, true>` and `Component<UID, boolean>` (the first does extend the second)

packages/core/utils/src/convert-query-params.ts
alexandrebodin: True, then should you just remove the TODO ? you already set it as deprecated in the typings

Convly: The todo was more about the replacing part, but I can remove it altogether

packages/core/utils/src/convert-query-params.ts
Convly: Even though it doesn't make sense to have multiple params, I believe it is used by the CM for some reason.

I would advise keeping the logic as it is right now, and adjust that when we'll look at the next behavior

alexandrebodin: Ok, can you add a // TODO to prevent this and discuss it with the content squad ?

Bassel17 (approved):
Looks good, works

PR #21555, opened 2024-10-02
feat: base files for static preview

Marc-Roig:
### What does it do?

Create basic outline of controllers/services/routes for static preview.

Next PR will introduce the preview config validation

### DIFF
--- a/packages/core/content-manager/server/src/preview/controllers/index.ts
+++ b/packages/core/content-manager/server/src/preview/controllers/index.ts
@@ -3,4 +3,8 @@ import { createPreviewController } from './preview';
 
 export const controllers = {
   preview: createPreviewController,
+  /**
+   * Casting is needed because the types aren't aware that Strapi supports
+   * passing a controller factory as the value, instead of a controller object directly
+   */
 } as unknown as Plugin.LoadedPlugin['controllers'];
--- a/packages/core/content-manager/server/src/preview/controllers/index.ts
+++ b/packages/core/content-manager/server/src/preview/controllers/index.ts
@@ -1,0 +1,6 @@
+import type { Plugin } from '@strapi/types';
+import { createPreviewController } from './preview';
+
+export const controllers = {
+  preview: createPreviewController,
+} as unknown as Plugin.LoadedPlugin['controllers'];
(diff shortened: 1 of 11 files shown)


packages/core/content-manager/server/src/preview/controllers/index.ts
```diff
@@ -0,0 +1,6 @@
+import type { Plugin } from '@strapi/types';
+import { createPreviewController } from './preview';
+
+export const controllers = {
+  preview: createPreviewController,
+} as unknown as Plugin.LoadedPlugin['controllers'];
```
remidej, line 6: Why is unknown needed?

and I thought you would need to call the function, I didn't know this style was allowed

packages/core/content-manager/server/src/preview/routes/index.ts
```diff
@@ -0,0 +1,10 @@
+import type { Plugin } from '@strapi/types';
+import { previewRouter } from './preview';
+
+/**
+ * The routes will me merged with the other Content Manager routers,
```
remidej, line 5:
```suggestion
 * The routes will be merged with the other Content Manager routers,
```

packages/core/content-manager/server/src/preview/controllers/index.ts
Marc-Roig: apparently the controller types do not account for using the factory itself, but it works (copied it from content history)

I will add a comment!

packages/core/content-manager/server/src/preview/index.ts
```diff
@@ -1,21 +1,38 @@
...
   return {
     bootstrap() {
       // eslint-disable-next-line no-console -- TODO remove when we have real functionality
       console.log('Bootstrapping preview server');
+
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- TODO: Remove when implemented
+      const config = strapi.config.get('admin.preview');
+
+      // TODO: Validation
+      // TODO: Disable feature if config is not valid
```
remidej, line 31: Maybe it's easier to just crash the app?

remidej (approved):

packages/core/content-manager/server/src/preview/index.ts
Marc-Roig: totally! I will do that in the [next](https://github.com/strapi/strapi/pull/21561) pr

PR #21558, opened 2024-10-02
feat: adding cutom validation method for password field

Bassel17:
### What does it do?

Same as #21493 but for V4

### Why is it needed?

- so people on v4 can also validate passwords on the api level
- it's considered more like a security fix than just an additional enhancement


jhoward1994 (approved):
Works well for me 👍🏻

PR #21559, opened 2024-10-02
fix: history trying to restore components that were deleted

### LINKED ISSUE #21405: Content History shows error when editing components in a dynamic zone

- Node.js version: v20.15.0
- NPM version:10.8.2
- Strapi version: 5.0.0
- Database:sqlite
- Operating system: Darwin-arm64

### Describe the bug

When attempting to restore piece of content that has a dynamic zone, occasionally Strapi is unable to restore. 

### Expected behavior

The content should restore with not issue.

### Screenshots

Loom issue below

https://www.loom.com/share/1ce57833d6a14669bedc59df16a0e350

remidej:
### What does it do?

Fix: https://github.com/strapi/strapi/issues/21405

The issue was that we made references to the IDs of components that were deleted while restoring content. Now I omit the ID of all components while restoring, forcing them to be recreated to avoid any issue

I also added an API test that isolates this behavior (it would fail before my change)

~~And I restored the history e2e tests, let's see if the CI is nice to them~~

### How to test it?

Have a content type that has a component field. 

- create content with components content
- delete the component content
- go to history, restore the history version that has components content
- it should be restored successfully

Do the same for:
- repeatable components
- components in dynamic zones
### Related issue(s)/PR(s)

fixes #21405

### DIFF
--- a/packages/core/content-manager/server/src/history/services/history.ts
+++ b/packages/core/content-manager/server/src/history/services/history.ts
@@ -1,5 +1,5 @@
 import type { Core, Data, Schema, Struct } from '@strapi/types';
-import { async, errors, traverseEntity } from '@strapi/utils';
+import { async, errors } from '@strapi/utils';
 import { omit } from 'lodash/fp';
 
 import { FIELDS_TO_IGNORE, HISTORY_VERSION_UID } from '../constants';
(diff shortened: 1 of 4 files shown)


tests/api/core/content-manager/content-manager/history/history.test.api.ts
```diff
@@ -670,6 +688,38 @@ describeOnCondition(edition === 'EE')('History API', () => {
       expect(restoredDocument['image']).toBe(null);
       expect(restoredDocument['images']).toHaveLength(1);
     });
+
+    test.only('Restores a version with deleted components', async () => {
```
Marc-Roig, line 692:
```suggestion
    test('Restores a version with deleted components', async () => {
```

packages/core/content-manager/server/src/history/services/history.ts
```diff
@@ -236,7 +236,25 @@ const createHistoryService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+          }
+        },
+        {
+          schema: strapi.getModel(version.contentType),
+          getModel: strapi.getModel.bind(strapi),
+        },
+        dataWithoutMissingRelations
+      );
+
+      const data = omit(['id', ...Object.keys(schemaDiff.removed)], dataWithoutComponentIds);
```
Marc-Roig, line 257: I was wondering if it would be easier to not select the id field from compos and dz's from the beggining, but that would not work for history versions created before this fix.

tests/api/core/content-manager/content-manager/history/history.test.api.ts
remidej: https://www.youtube.com/watch?v=r49hLfjjxpM

packages/core/content-manager/server/src/history/services/history.ts
remidej: You're right! A big plus of your approach is not having to apply that logic each time history is fetched.

It won't work for history versions from 5.0.0 to 5.0.2, but I don't think it's a big deal, history versions get auto-deleted after the retention period anyway

remidej:
@Marc-Roig i think I may need a design system fix to re-enable the e2e tests, so I removed that part from this PR to not block it

Marc-Roig (approved):
nice!

PR #21573, opened 2024-10-03
enhancement: allow db connection functions

innerdvations:
### What does it do?

Allow database connection functions in user configs without any breaking changes

### Why is it needed?

https://github.com/strapi/strapi/pull/18857

### How to test it?

Use a function in your database config connection instead of an object

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/18857

### DIFF
--- a/packages/core/database/src/index.ts
+++ b/packages/core/database/src/index.ts
@@ -1,5 +1,6 @@
 import type { Knex } from 'knex';
 
+import createDebugger from 'debug';
 import { Dialect, getDialect } from './dialects';
 import { createSchemaProvider, SchemaProvider } from './schema';
 import { createMetadata, Metadata } from './metadata';
@@ -15,6 +16,8 @@ import type { Migration } from './migrations';
 import { type Identifiers } from './utils/identifiers';
 
 export { isKnexQuery } from './utils/knex';
+
+const debug = createDebugger('strapi::database');
 
 interface Settings {
   forceMigration?: boolean;
@@ -97,8 +100,30 @@ class Database {
   }
 
   async init({ models }: { models: Model[] }) {
+    /**
+     * TODO v6: implement a cleaner solution that will involve breaking changes to db constructor
+     *
+     * if the user passed in a connection function, we need to resolve it first
+     * and recreate the connection, otherwise accessing this.connection.anyProperty will not work
+     *
+     * Note that even if it was already created in the constructor, the connection method will
+     * not be called at this point by Strapi or Knex, so it shouldn't have any effects to do this
+     *  */
+    if (typeof this.config.connection.connection === 'function') {
+      debug('User config.connection is a function; resolving and recreating knex connection');
+      this.config.connection.connection = await this.config.connection.connection();
+
+      await this.connection.destroy();
+      this.connection = createConnection(this.config.connection, {
+        pool: { afterCreate: afterCreate(this) },
+      });
+    }
+
     this.metadata.loadModels(models);
     await validateDatabase(this);
+
+    debug('Database initialized');
+
     return this;
   }
 


packages/core/database/src/index.ts
```diff
@@ -97,8 +100,30 @@ class Database {
...
+     *
+     * if the user passed in a connection function, we need to resolve it first
+     * and recreate the connection, otherwise accessing this.connection.anyProperty will not work
+     *
+     * Note that even if it was already created in the constructor, the connection method will
+     * not be called at this point by Strapi or Knex, so it shouldn't have any effects to do this
+     *  */
+    if (typeof this.config.connection.connection === 'function') {
+      debug('User config.connection is a function; resolving and recreating knex connection');
+      this.config.connection.connection = await this.config.connection.connection();
```
innerdvations, line 114: Final version we would clone or deconstruct the user config instead of rewriting it

innerdvations:
I think there may be issues with migrations, investigating now

innerdvations:
Nope, the issue was this: https://github.com/strapi/strapi/pull/21576

innerdvations:
If we like this idea, I will give @evan-thinksync some time to update their PR https://github.com/strapi/strapi/pull/18857 with this code since they started the effort to get credit :)

packages/core/database/src/index.ts
```diff
@@ -49,7 +52,7 @@ const afterCreate =
   };
 
 class Database {
-  connection: Knex;
+  #connection?: Knex;
```
alexandrebodin, line 55: Maybe someday they will have consistency between # and ts private keyword 🙌

alexandrebodin (dismissed):

packages/core/database/src/index.ts
innerdvations: Do we have a preference for Strapi?

alexandrebodin:
> If we like this idea, I will give @evan-thinksync some time to update their PR #18857 with this code since they started the effort so they can get credit :)

You can ammend your commit and add him as co-author to give credits ;)

innerdvations:
> > If we like this idea, I will give @evan-thinksync some time to update their PR #18857 with this code since they started the effort so they can get credit :)
> 
> You can amend your commit and add him as co-author to give credits ;)

Does that work in the auto-generated release notes on github?

packages/core/database/src/index.ts
Marc-Roig: I would vote for # as its enforced by javascript itself

packages/core/database/src/index.ts
```diff
@@ -82,9 +85,11 @@ class Database {
 
     this.metadata = createMetadata([]);
 
-    this.connection = createConnection(this.config.connection, {
-      pool: { afterCreate: afterCreate(this) },
-    });
+    // for object connections, create Knex connection synchronously
```
Marc-Roig, line 88: could you specify a bit more why we would need a function? 🙏 
A note explaining that we support async conections by deffining the connection as a function

packages/core/database/src/index.ts
alexandrebodin: private in ts is just a type check thing and # is actually runtime enforced so # for actual strictness is better. hopefully the behaviour will converge because I really dislike this syntax 😂

packages/core/database/src/index.ts
innerdvations: If you check the linked issue you can find some use cases directly from users, but it's necessary for some database authentication methods

Marc-Roig: ah , I meant on the code itself, so its easy to understand why we do this 🙈

packages/core/database/src/index.ts
```diff
@@ -96,12 +101,39 @@ class Database {
...
+
     this.metadata.loadModels(models);
     await validateDatabase(this);
+
+    debug('Database initialized');
+
     return this;
   }
 
+  get connection() {
```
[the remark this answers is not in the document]
Boegie19: Just a check for this. In register the whole db part is not initialized. so what you would NOT be a breaking change. But if it is not loaded in by bootstrap It WILL be a breaking change.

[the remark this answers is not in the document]
innerdvations: For connection object format, it would already be initialized because we do it in the constructor 100% of the time, synchronously, and Strapi creates the Database in its own constructor before any user code is run.

For connection function format, db is the very first thing we load in Strapi bootstrap before running any user bootstrap functions, so it should still be fine for the case you mention with seeding, although if you have a case you can test where it would fail, please post an example. But there are still potentially cases where it could fail, like if user code tries to access `connection[someProperty]` during register rather than waiting for bootstrap, which is why we have to consider the connection function format to be experimental.

Boegie19 (changes requested):
https://github.com/strapi/strapi/blob/006a21d39433b14a52c58dc2600abacaf3e551ea/packages/core/database/src/dialects/sqlite/index.ts#L22-L29

Using the function on a sqlite database won't work because of the above configure statement. and will stop strapi from starting

innerdvations:
> https://github.com/strapi/strapi/blob/006a21d39433b14a52c58dc2600abacaf3e551ea/packages/core/database/src/dialects/sqlite/index.ts#L22-L29
> 
> Using the function on a sqlite database won't work because of the above configure statement. and will stop strapi from starting

Ahhh, good catch, thanks! I will fix that.

packages/core/database/src/index.ts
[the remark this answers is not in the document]
Boegie19: During register the whole strapi.db object does not exist.

Also just tested and on bootstrap it will always work so you are right this is not a breaking change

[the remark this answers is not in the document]
innerdvations: 🤔 It should, we do this in the Strapi constructor: https://github.com/strapi/strapi/blob/8ed749ecd9fe2851f95d5f48e9a4ec95327ff57a/packages/core/core/src/Strapi.ts#L272

Maybe it's not available for some reason though

innerdvations:
> Using the function on a sqlite database won't work because of the above configure statement. and will stop strapi from starting

🫠 knex sqlite3 dialect doesn't take into account that connection functions exist, so it has a bug where it warns about the wrong filename if you use a connection function. I may put in a PR over there to fix it #hacktoberfest

innerdvations:
@fardarter I believe this version should work for you, although I'm not familiar enough with how expirationChecker works to know how to test it 😆 but we're letting knex handle it all now so it should be fine.

packages/core/database/src/dialects/dialect.ts
```diff
@@ -17,7 +17,8 @@ export default class Dialect {
     this.client = client;
   }
 
-  configure() {}
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
alexandrebodin, line 20: time to use an abstract class & methods ?

innerdvations: Almost time 😆 I would just wait until we do some investigation on supporting dialects via user code

packages/core/database/src/index.ts
```diff
@@ -92,16 +107,37 @@ class Database {
...
-
-    this.logger = config.logger ?? console;
   }
 
   async init({ models }: { models: Model[] }) {
+    if (typeof this.config.connection.connection === 'function') {
+      // for function connections, we must configure the dialect in init
+      // Note: due to a bug in Knex, sqlite will warn about a missing filename for function connections because
+      // it checks in the constructor rather than on connect
+      const conn = await this.config.connection.connection();
```
fardarter, line 117: It would probably be better from a performance perspective to only run the token acquisition when needed. I suspect (though am not sure) that getting it twice won't invalidate the first one, but it's still a startup and code cost.

Given that all you're doing with the connection from that point seems to be getting a string value, why not just get the string from the client string from the config object rather than the connection?

I don't see it as inherently more reliable to get it off the connection function -- it's coming from the config value anyway. Unless I'm reading the code wrong.

innerdvations: it's a bit of a catch-22 that requires refactoring of our dialect code to get right, but I was worried about breaking something. Basically, dialect needs access to the connection config before we can use the knex connection. But we don't have access to the connection config object until the connection is used.

I might be able to do some safe refactoring to get it a bit closer, but I don't think we'll be able to get it perfect until we make some breaking changes. And even with breaking changes, it may be impossible in cases like sqlite where we want to verify a filename exists before we create the knex connection, but I think it would be fine there to just disable the check and have it be the responsibility of the user to ensure the file exists if they're using a connection function.

fardarter: Maybe it's worth lifting the config object up the initialisation path? Anything lower down the path would still be able to access it, and you'd be able to make use of it much earlier. 

Obviously don't know the code well, but that seems like something that could be done without breakage?

Just feels (from the outside, without full context) like you're taking on complexity to avoid a larger piece of work which, let me be absolutely clear about, I have 100% done in my life, so I completely get it, but I have also regretted it more times than I have not.

innerdvations: The main problem is that we can't, until we are able to make a breaking change (the one I missed the deadline on making for v5) just due to how everything that was sync has to remain sync, and we need to be able to modify the user knex config before passing it to knex.

However, I just pushed an update that should solve everything. I tested it out locally with expiring connections and it seems to work great. I will generate a new experimental and post it in a few minutes.

innerdvations: In v6 we will just declare "you cannot rely on connection.connection to be an object and you must wait until db init before accessing it" and update our code internally as well, and can remove any added complexity here

alexandrebodin (approved):

Issue #19964, opened 2024-03-29
core database possible bug and performance improvement


binarybase:
### Required System information

- Node.js version: 18.19.0
- NPM version: 9.9.2
- Strapi version: tested from 4.11.1 to 4.21.1 current
- Database: mysql (8.0.36)
- Operating system: Windows 10 LTSC
- Is your project Javascript or Typescript: Typescript

### Steps to reproduce the behavior

1. Run strapi with 200+ Content-Types
2. Created new Collection Type and restarted the server

### Expected behavior

After creating new Collection-Type, during the server restart - core database module should compare database schema, skip unmodified fields, update existing or create new ones.

Instead of that it forces deleting foreign keys and recreates new ones even if they was not modified.
Restarting server takes tooo long because Strapi is reindexing whole database.

### Screenshots

Without patched database
![image](https://github.com/strapi/strapi/assets/87497154/fc345068-f789-481d-8c9f-35c66b5fab1e)
![image](https://github.com/strapi/strapi/assets/87497154/45b2f982-2307-4ed2-8a0b-2b01081aa62e)
![image](https://github.com/strapi/strapi/assets/87497154/e9afc1f0-e688-47c0-8b64-22db790a5008)

With patched database
it will just only create new table, but not recreating existing foreign keys for other tables
as you can see the launch time was 14 seconds instead of 141 seconds

![image](https://github.com/strapi/strapi/assets/87497154/11a8721a-6829-4738-842a-c90664a3abe7)
![image](https://github.com/strapi/strapi/assets/87497154/3e2db16d-102c-4a45-97c1-711c69eab671)

### Code snippets

The first issue I have found in mysql/schema-inspector.js that produces unwanted indexes update based on bad value comparator.
knex queries database with SHOW INDEX command and the result will return raw values from database. The result query column called "Non_unique" is not an number or boolean, so the JS comparator !index.Non_unique will fail and the result will be NULL value instead of string 'unique'. I have modified the comparator from `!index.Non_unique` into `index.Non_unique === '0'`.

The second issue is with the foreign key `onUpdate` comparing. Its weird that in the stored database schema (table strapi_database_schema) is this value not included, but database query it includes. So it always looks like this value has been changed.

[...]

PR #21576, opened 2024-10-03
fix: perform index diff ignoring array order

innerdvations:
### What does it do?

sorts the columns when running a diff on indexes

### Why is it needed?

if new columns/indexes are registered asynchronously in strapi register, they appear in a different order

then when we run a db sync, we see them as changed and recreate the indexes

there's currently a race condition (no github issue for it) where a handful of indexes on internal strapi columns are being recreated somewhat randomly on every startup

### How to test it?

Unit tests have been added to confirm that we don't have any more ordering issues

#### Manual testing
without this PR:

run `DEBUG=strapi::database yarn develop` a few times in a row and see that some indexes get recreated for no reason

switch to this PR:

it will not recreate the indexes unless something actually changes in your schema

Then in general, dropping, adding and updating attributes in a schame or CTB should still work fine.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/19964 is NOT fixed by this but may be mitigated

### DIFF
--- a/packages/core/database/src/schema/__tests__/schema-diff.test.ts
+++ b/packages/core/database/src/schema/__tests__/schema-diff.test.ts
@@ -120,6 +120,55 @@ describe('diffSchemas', () => {
     });
   });
 
+  test('Table order ignored', async () => {
+    const databaseSchema: Schema = {
+      tables: [
+        {
+          name: 'table_a',
+          columns: [],
+          indexes: [],
+          foreignKeys: [],
+        },
+        {
+          name: 'table_b',
+          columns: [],
+          indexes: [],
+          foreignKeys: [],
+        },
+      ],
+    };
+
+    const userSchema: Schema = {
+      tables: [
+        {
+          name: 'table_b',
+          columns: [],
+          indexes: [],
+          foreignKeys: [],
+        },
+        {
+          name: 'table_a',
+          columns: [],
+          indexes: [],
+          foreignKeys: [],
+        },
+      ],
+    };
+
+    const result = await diffSchemas({ databaseSchema, userSchema, previousSchema: userSchema });
+
+    expect(result.status).toBe('UNCHANGED');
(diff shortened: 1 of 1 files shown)


packages/core/database/src/schema/diff.ts
```diff
@@ -272,7 +276,8 @@ export default (db: Database) => {
         const { status, diff } = diffIndexes(databaseIndex, userSchemaIndex);
 
         if (status === statuses.CHANGED) {
-          updatedIndexes.push(diff);
+          console.log('CHANGED INDEX', diff, databaseIndex, userSchemaIndex);
```
alexandrebodin, line 279: hum :D

packages/core/database/src/schema/diff.ts
```diff
@@ -89,7 +89,11 @@ export default (db: Database) => {
   const diffIndexes = (oldIndex: Index, index: Index): IndexDiff => {
     const changes: string[] = [];
 
-    if (!_.isEqual(oldIndex.columns, index.columns)) {
+    // Sort columns before comparing
+    const sortedOldColumns = [...oldIndex.columns].sort();
+    const sortedNewColumns = [...index.columns].sort();
+
+    if (!_.isEqual(sortedOldColumns, sortedNewColumns)) {
```
alexandrebodin, line 96: you could leave all of this to lodash using the _.xor if you don't want to bother with the spreading & sort

alexandrebodin (dismissed):

alexandrebodin (dismissed):

derrickmehaffy:
Building exp to have a user test this fix: https://github.com/strapi/strapi/actions/runs/11171100942

Marc-Roig (dismissed):
nice catch

Marc-Roig (approved):

Convly (approved):

PR #21578, opened 2024-10-03
fix: copy button shown on v4 plugins

Bassel17:
### What does it do?

insures that  a plugin is considered incompatible if versionRange is not valid

### Why is it needed?

copy install button is showing on invalid plugins for v5

### How to test it?

- plugins like “BigBluebutton” or “Cloudflare pages” shouldn’t be shown with the “Copy install command” button

### DIFF
--- a/packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx
+++ b/packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx
@@ -275,28 +275,17 @@ const CardButton = ({
       return (
         <Tooltip
           data-testid={`tooltip-${pluginName}`}
-          label={
-            !versionRange
-              ? formatMessage(
-                  {
-                    id: 'admin.pages.MarketPlacePage.plugin.version.null',
-                    defaultMessage:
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/pages/Marketplace/components/NpmPackageCard.tsx
```diff
@@ -272,28 +275,17 @@ const CardButton = ({
       return (
         <Tooltip
           data-testid={`tooltip-${pluginName}`}
-          label={
-            !versionRange
-              ? formatMessage(
-                  {
-                    id: 'admin.pages.MarketPlacePage.plugin.version.null',
```
innerdvations, line 279: We should delete this string from the translations as well, unless we plan to use it again in the near future and reference it in a ticket so we don't forget it

Bassel17: no I don't think we need it, I think the strategy is to be more strict anyways, instead of saying we don't know

jhoward1994 (approved):

PR #21610, opened 2024-10-07
fix: update review workflow

Marc-Roig:
### What does it do?

There is an ongoing [PR](https://github.com/strapi/strapi/pull/21501) to fix non DP entries that were migrated with an empty "publishedAt" column, this issue also was breaking RW, making it imposible to change a stage. 

This will fix it asap before we merge the other PR

### DIFF
--- a/packages/core/review-workflows/server/src/services/stages.ts
+++ b/packages/core/review-workflows/server/src/services/stages.ts
@@ -210,7 +210,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       const entity = await strapi.documents(model).update({
         documentId,
         locale,
-        data: { [ENTITY_STAGE_ATTRIBUTE]: stage },
+        data: { [ENTITY_STAGE_ATTRIBUTE]: pick(['id'], stage) },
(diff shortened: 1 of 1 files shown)


packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -210,7 +210,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       const entity = await strapi.documents(model).update({
         documentId,
         locale,
-        data: { [ENTITY_STAGE_ATTRIBUTE]: stage },
+        data: { [ENTITY_STAGE_ATTRIBUTE]: pick(['id'], stage) },
```
Marc-Roig, line 213: when connecting a relation with an "id" , it skips the document id+locale+status transformation. As the review workflow stage doesnt have i18n nor DP, its ok to skip it

remidej: maybe this could be a comment in the code?

Marc-Roig: yes!! good point
just added a comment

PR #21612, opened 2024-10-07
feat(content-releases): considered review stage when getting the action status

Feranchz:
### What does it do?

Releases' status depends on the validation status of attached entries. Now these status could change based on changes on Review Workflows (if one content type is added to a Review Workflow for example)

### How to test it?

Tests that all the following actions cause the required changes on releases's status:

- Changing the review stage of one entry inside a release.
- Creating a new Workflow with one content type that has entries inside a release.
- Update a Workflow with content types that has entries inside a release.
- Remove a workflow with content types that has entries inside a release.

### DIFF
--- a/packages/core/content-releases/server/src/middlewares/documents.ts
+++ b/packages/core/content-releases/server/src/middlewares/documents.ts
@@ -17,6 +17,7 @@ const updateActionsStatusAndUpdateReleaseStatus = async (
 ) => {
   const releases = await strapi.db.query(RELEASE_MODEL_UID).findMany({
     where: {
+      releasedAt: null,
       actions: {
         contentType,
         entryDocumentId: entry.documentId,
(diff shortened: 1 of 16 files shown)


packages/core/content-releases/server/src/middlewares/documents.ts
```diff
@@ -27,7 +28,7 @@ const updateActionsStatusAndUpdateReleaseStatus = async (
 
   const entryStatus = await isEntryValid(contentType, entry, { strapi });
 
-  await strapi.db.query(RELEASE_ACTION_MODEL_UID).update({
+  await strapi.db.query(RELEASE_ACTION_MODEL_UID).updateMany({
```
Marc-Roig, line 31: you can have two actions refering to the published and draft version of a document?

Marc-Roig:
note: there is an issue that prevents changing the RW stage of an entry if it has i18n enabled, but that should be fixed in `develop`

packages/core/content-releases/server/src/middlewares/documents.ts
Feranchz: No, but you could have the same document in multiple releases (so you have multiple actions)

Marc-Roig: Ah! Good point

Marc-Roig (approved):

PR #21614, opened 2024-10-07
fix: types for custom http response APIs

dhoko:
Hello 👋 , a small fix

### What does it do?

Fix the type definitions for HTTP methods added on top of the `Koa.Context`.

Today if we add details tsc won't be happy

```sh
server/src/controllers/sources.ts:14:60 - error TS2554: Expected 0-1 arguments, but got 2.

14       ctx.internalServerError('Cannot fetch source pages', parseError(e))
                                                              ~~~~~~~~~~~~~
```

because of:

```
(property) ExtendableContext.internalServerError: (response?: string | object) => Koa.Context
```

### Why is it needed?

Code behind these methods:

```ts
      app.response[camelCasedName] = function responseCode(message = name, details = {}) {
        const httpError = createError(code, message, { details });
        const { status, body } = formatHttpError(httpError);
        this.status = status;
        this.body = body;
      };
```
> cf https://github.com/strapi/strapi/blob/develop/packages/core/core/src/services/server/koa.ts#L28

And confirmed by a test:
```sh
ROBERT 500 {
  data: null,
  error: {
    status: 500,
    name: 'InternalServerError',
    message: 'Cannot fetch source pages',
    details: [ [Object], [Object] ]
  }
} 500
```

```
    app.response[camelCasedName] = function responseCode(message = name, details = {}) {
      const httpError = createError__default.default(code, message, { details });
      const { status, body } = errors.formatHttpError(httpError);
      console.log("ROBERT", status, body, code)
      this.status = status;
      this.body = body;
    };
```


hanpaine:
Hey @dhoko thanks for submitting this and sorry we missed this last month! 🙏 we'll take a look soon 🚀

Bassel17 (approved):
LGTM!
thanks for the contribution

PR #21616, opened 2024-10-07
fix: history cron job breaking on mysql

### LINKED ISSUE #21613: Content History scheduled Job is using the wrong time format and wrong cron schedule format

- Node.js version: v18.20.2
- NPM version: Unknown
- Strapi version: 5.x.x (unknown exactly)
- Database: MySQL (exact version unknown)
- Operating system: Docker Node container 
- Is your project Javascript or Typescript: Unknown

### Describe the bug

Every evening when the cron kicks off the user encounters the following error because the time format given to the database is incorrect:

```
/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/packets/packet.js:738
    const err = new Error(message);
                ^
Error: delete from `strapi_history_versions` where (`created_at` < '2024-07-07T21:00:00.024Z') - Incorrect datetime value: '2024-07-07T21:00:00.024Z' for column 'created_at' at row 1
    at Packet.asError (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/packets/packet.js:738:17)
    at Query.execute (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/commands/command.js:29:26)
    at Connection.handlePacket (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/connection.js:481:34)
    at PacketParser.onPacket (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/connection.js:97:12)
    at PacketParser.executeStart (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/packet_parser.js:75:16)
    at Socket.<anonymous> (/app/work/strapi/build_images/-snip-/node_modules/mysql2/lib/connection.js:104:25)
    at Socket.emit (node:events:517:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:368:12)
    at readableAddChunk (node:internal/streams/readable:341:9) {
  code: 'ER_TRUNCATED_WRONG_VALUE',
  errno: 1292,
  sqlState: '22007',
  sqlMessage: "Incorrect datetime value: '2024-07-07T21:00:00.024Z' for column 'created_at' at row 1",
  sql: "delete from `strapi_history_versions` where (`created_at` < '2024-07-07T21:00:00.024Z')"
}
Node.js v18.20.2
error Command failed with exit code 1.
```

1. Enable enterprise mode
2. Wait for cron to kick off
3. See error

### Expected behavior

Proper datetime format used

### Screenshots

N/A

### Code snippets

Appears to come from

https://github.com/strapi/strapi/blob/bbe5b7f1f46249a3154695d5ecc04fc4f74793b4/packages/core/content-manager/server/src/history/services/lifecycles.ts#L174-L189

Specifically this line:

https://github.com/strapi/strapi/blob/bbe5b7f1f46249a3154695d5ecc04fc4f74793b4/packages/core/content-manager/server/src/history/services/lifecycles.ts#L182

### Additional context

Related TID: 8440

Also the syntax of this cron task is incorrect, it's using the older format and should be changed to the new format see: https://github.com/strapi/strapi/pull/21407

remidej:
### What does it do?

- stops using an ISO 8601 string in the filter, because it breaks on mysql
- gives a name to the cron job (we don't use strapi.cron because we don't want the job to be disable-able)
- adds a catch clause to prevent crashing the app if other errors occur

### How to test it?

- Use different types of databases, starting with mysql
- Update the code to make the job run every second: `* * * * * *`, add a log to make sure it works, and build the content manager (or run yarn setup)
- The app should not crash
- Old history version entries should be deleted (you can make some to test with a db explorer)

### Related issue(s)/PR(s)

fixes #21613

### DIFF
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -181,8 +181,7 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
           .deleteMany({
             where: {
               created_at: {
-                // Only keep YYYY-MM-DD from the string because mysql doesn't support ISO 8601
-                $lt: expirationDate.toISOString().split('T').at(0),
+                $lt: expirationDate,
               },
             },
           })
--- a/packages/core/content-manager/server/src/history/services/lifecycles.ts
+++ b/packages/core/content-manager/server/src/history/services/lifecycles.ts
@@ -172,17 +172,25 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
       });
 
       // Schedule a job to delete expired history versions every day at midnight
-      state.deleteExpiredJob = scheduleJob('0 0 * * *', () => {
+      state.deleteExpiredJob = scheduleJob('historyDaily', '0 0 * * *', () => {
         const retentionDaysInMilliseconds = serviceUtils.getRetentionDays() * 24 * 60 * 60 * 1000;
         const expirationDate = new Date(Date.now() - retentionDaysInMilliseconds);
 
-        strapi.db.query(HISTORY_VERSION_UID).deleteMany({
-          where: {
-            created_at: {
-              $lt: expirationDate.toISOString(),
+        strapi.db
+          .query(HISTORY_VERSION_UID)
+          .deleteMany({
+            where: {
+              created_at: {
+                // Only keep YYYY-MM-DD from the string because mysql doesn't support ISO 8601
+                $lt: expirationDate.toISOString().split('T').at(0),
+              },
             },
-          },
-        });
+          })
+          .catch((error) => {
+            if (error instanceof Error) {
+              strapi.log.error('Error deleting expired history versions', error.message);
+            }
+          });
       });
 
       state.isInitialized = true;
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -172,17 +172,25 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-          where: {
-            created_at: {
-              $lt: expirationDate.toISOString(),
+        strapi.db
+          .query(HISTORY_VERSION_UID)
+          .deleteMany({
+            where: {
+              created_at: {
+                // Only keep YYYY-MM-DD from the string because mysql doesn't support ISO 8601
+                $lt: expirationDate.toISOString().split('T').at(0),
```
remidej, line 185: This is basically the fix. There may be a more elegant one. @Marc-Roig do you think the db engine should be the one to handle that?

Marc-Roig (approved):

packages/core/content-manager/server/src/history/services/lifecycles.ts
innerdvations: Agreed, I also worry introducing timezones like that could cause some subtle bugs unless we have tests explicitly verifying they'll work correctly

oiorain:
QA done with Remi. LGTM! 🎸

packages/core/content-manager/server/src/history/services/lifecycles.ts
remidej: @Marc-Roig you're right Knex handles the date object directly. I tested manually with mysql postgres and sqlite, so I updated the code

Convly (approved):
Changes LGTM, haven't tested it yet

PR #21726, opened 2024-10-09
fix: send ids when connecting relations to update u&p user role

### LINKED ISSUE #21653: Error during assigning/re-assigning roles for user via admin panel

- Node.js version:v20.10.0
- NPM version:10.2.3
- Strapi version:5.4.0
- Database:postgres
- Operating system:darwin-x64
- Is your project Javascript or Typescript:Typescript

### Describe the bug
Error appears if role is changed for user.

1. Go to any user page in admin panel
2. Modify role from authenticated to public
3. Press save button

### Expected behavior

Role must be changed and saved.

### Screenshots

image

Marc-Roig:
### What does it do?
It was not possible to create a new U&P user, or update its roles. 

U&P was expecting the role attribute to be set using the `id` attribute, but the CM was only sending the document ID.

By default, when connecting a relation, the document service [ will prioritize the documentId ](https://github.com/strapi/strapi/blob/7a6d9a2b0e029dd25100e68f71a35f20a134d3a2/packages/core/core/src/services/document-service/transform/relations/extract/data-ids.ts#L60) and transform it to the appropiate entry id.

fixes #21653

### DIFF
--- a/packages/plugins/users-permissions/server/controllers/validation/user.js
+++ b/packages/plugins/users-permissions/server/controllers/validation/user.js
@@ -17,12 +17,13 @@ const createUserBodySchema = yup.object().shape({
           .shape({
             connect: yup
               .array()
-              .of(yup.object().shape({ id: yup.strapiID().required() }))
+              // NOTE: Content Manager uses document ID to connect relation
+              .of(yup.object().shape({ documentId: yup.string().required() }))
               .min(1, 'Users must have a role')
               .required(),
           })
           .required()
-      : yup.strapiID().required()
+      : yup.string().required()
   ),
 });
 
--- a/packages/plugins/users-permissions/server/controllers/validation/user.js
+++ b/packages/plugins/users-permissions/server/controllers/validation/user.js
@@ -17,13 +17,12 @@ const createUserBodySchema = yup.object().shape({
           .shape({
             connect: yup
               .array()
-              // NOTE: Content Manager uses document ID to connect relation
-              .of(yup.object().shape({ documentId: yup.string().required() }))
+              .of(yup.object().shape({ id: yup.strapiID().required() }))
               .min(1, 'Users must have a role')
               .required(),
           })
           .required()
-      : yup.string().required()
+      : yup.strapiID().required()
   ),
 });
 
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -87,6 +87,7 @@ function useHandleDisconnect(fieldName: string, consumerName: string) {
     addFieldRow(`${fieldName}.disconnect`, {
       id: relation.id,
       apiData: {
+        id: relation.id,
         documentId: relation.documentId,
         locale: relation.locale,
       },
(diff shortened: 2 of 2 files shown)


packages/plugins/users-permissions/server/controllers/validation/user.js
```diff
@@ -17,12 +17,13 @@ const createUserBodySchema = yup.object().shape({
...
               .array()
-              .of(yup.object().shape({ id: yup.strapiID().required() }))
+              // NOTE: Content Manager uses document ID to connect relation
+              .of(yup.object().shape({ documentId: yup.string().required() }))
               .min(1, 'Users must have a role')
               .required(),
           })
           .required()
-      : yup.strapiID().required()
+      : yup.string().required()
```
innerdvations, line 26: String is fine for now, but I wonder if we should have a `strapiDocId()` in the future.

innerdvations (commented):
Could this be considered a breaking change? I mean, I know it's already broken, but did v5.0.0 support it as a strapi numeric ID, and now we're changing it to document id?

If not, ignore me, but if so, can we support both?

Marc-Roig:
@innerdvations good point, I will allow both just in case, I believe there are many users with custom u&p logic

alexandrebodin:
To follow on @innerdvations comment, We could consider that this is a BC just because the FE is sending different data than before 🤔 

Could we do this instead ?

- Should we readd the id in the CM requests
- Make sure that if a documentId is sent in the docService it's used 1st 
- leave the U&P code use the ids for now

alexandrebodin:
> @alexandrebodin
> 
> > Should we readd the id in the CM requests
> 
> agree , I assume you mean reverting this change [#21428 (files)](https://github.com/strapi/strapi/pull/21428/files#diff-b83b1eab85f1f8b8ea26078e27af271156187197505b5f99ba4d5259bd49717aR87) and using id's to connect relations (will also add a note if we do that so we dont forget about u&p)
> 

No reverting but also sending the id 😄 

> > Make sure that if a documentId is sent in the docService it's used 1st
> 
> I believe thats the case now! documentId takes precedence when connecting a relation

Ok let's make sure but good to know 🙏

alexandrebodin (approved):
lgtm. woth doing a more global QA on the CM just to be on the safe side 🙏

MarionLemaire:
Test plan discussed with @Marc-Roig  : 

Create a User in CM
Update a User in CM
Check every type of relations for different collection types (with i18n/D&P enabled and disabled) in CM

MarionLemaire:
@Marc-Roig : Tested today using the experimental : 

- New user can be created using the CM.
- User can be updated using the CM.

At user creation (not for updating), there's an UI issue with a error message that shouldn't be displayed : 

https://github.com/user-attachments/assets/2ad322ce-715b-4f29-a271-2c3012b1a246

Non-blocking but annoying.

I'll now focus on relations in CM.

MarionLemaire:
@Marc-Roig : Checking relations, here is an issue about the title displayed in Relations field : 
Owner (D&P + i18n enabled) has many-to-many relation field with Dog (D&P and i18n disabled).
On the Dog side, the Owner relation field is wrongly displayed (should be the DogName title)

https://github.com/user-attachments/assets/be1102e2-12d1-4214-8401-2372cf3a4c3c

PR #21737, opened 2024-10-09
fix: do not display fill from locales if i18n no enabled

### LINKED ISSUE #21203: “Fill from another locale” button should not appear when i18n is not activated
### **Steps to reproduce**
- Go an entry without i18n enabled
- See that the fill from another locale button is there and disabled

### **Expected behavior**

We should not show the disabled icon when i18n is disabled

### **Screenshots**

Image

Marc-Roig:
### What does it do?

Do not display the "Fill from locale" button if i18n is not enabled or there is not more than one locale.

fixes #21203

### DIFF
--- a/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
+++ b/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
@@ -279,7 +279,7 @@ const FillFromAnotherLocaleAction = ({
     onClose();
   };
 
-  if (!hasI18n || !Array.isArray(locales) || locales.length === 0) {
+  if (!hasI18n) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -278,6 +279,10 @@ const FillFromAnotherLocaleAction = ({
     onClose();
   };
 
+  if (!hasI18n || !Array.isArray(locales) || locales.length === 0) {
```
remidej, line 282: if i18n is enable but the locales aren't loaded, should we fall back to the disabled state defined below? otherwise I don't see what it's for

Marc-Roig: good point, will do that!

remidej (approved):
simple and effective 👌

tested locally

PR #21467, opened 2024-09-25
enhancement: add support for Node 22

innerdvations:
### What does it do?

Add Node 22 to workflows:

- allow matrices on every test type for consistency
- add node 22 to several test runs
- flatten runs to only two levels of dependency:
  - first: short checks (lint, pretty, build, changes)
  - second: all tests (cli, typescript, e2e, api, unit:front, unit:back)
- removes Node 18 from CI
  
Fixes required for Strapi to run with Node 22:

- update project template `better-sqlite3` package to a version that supports Node 22
- add 5.1.0 codemod to upgrade better-sqlite3 package in existing projects

### Why is it needed?

- Node 22 becomes active version October 22, 2024

### How to test it?

- the CI workflow should pass here
- Experimental: 0.0.0-experimental.9c7de54d8ce77e13047fc15ada33f937a068f1f2
  - `npx create-strapi-app@0.0.0-experimental.9c7de54d8ce77e13047fc15ada33f937a068f1f2`

### Related issue(s)/PR(s)

DX-1510

PR #21740, opened 2024-10-09
fix: test CI/CD building repo 2 times

Boegie19:
### What does it do?
Moved lint to second strage since it requires build stap
Removed pretty as a requirement to run second stage tests (Final test will still fail but we atleast want to know if the code passes) 
increase E2E shards 
### Why is it needed?
1  to not build 2 times
2 To make the tests run if prettier fails so that we can still check if there are any code issues or only prettier issues
3 so that the time it takes for the tests to run goos down from 21 to 15
### How to test it?

If all test pass it should be fine 

### Related issue(s)/PR(s)

 https://github.com/strapi/strapi/pull/21467
 
 
 2 shards https://github.com/strapi/strapi/actions/runs/11256802614
 5 shards https://github.com/strapi/strapi/actions/runs/11257246445/job/31301300364?pr=21740
 7 shards https://github.com/strapi/strapi/actions/runs/11257753134?pr=21740
 
 aka sharding here does not help since all the log tests happen in shard 1 aka to make it go down form taking 10 minutes we need Playwright Orchestration so that it is split more fairly
 
 From my research only 2 exist 
 https://github.com/alyaothman14/playwright-smart-orchestration-js/tree/main
 Second one is propertery
 https://docs.currents.dev/guides/parallelization-guide/pw-parallelization/playwright-orchestration

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -177,7 +177,7 @@ jobs:
       fail-fast: false # remove once tests aren't flaky
       matrix:
         project: ['chromium', 'webkit', 'firefox']
-        shard: [1/7, 2/7, 3/7, 4/7, 5/7, 6/7, 7/7]
+        shard: [1/2, 2/2]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


Boegie19:
@innerdvations 
Question for you is the tests taking 18 minutes a issue for you guys? Or just an enoying and not worth fixing yet. if it is the second then the question becomes how long can tests take before it is an issue.  if it is the first I think I can write a tool in a couple of days like playwright-smart-orchestration-js only using github cache + a custom reporter

innerdvations (approved):

PR #21754, opened 2024-10-09
Add <latest> Upgrade Command

Convly:
### What does it do?

Introduces a new `latest` release type and command for version upgrades.

When using:

`upgrade latest`

The upgrade tool will try to upgrade to the latest available version of Strapi (no other rule applies unlike `major`, `minor`, and `patch`)

### Why is it needed?

Provide an option for users to easily upgrade to the latest available version of Strapi without having to specify a specific release type.

It simplifies the upgrade process, making it more intuitive and flexible, thus hopefully improving the overall developer experience.

### How to test it?

Set up a Strapi environment and use the 'upgrade' command with the new 'latest' command.

Verify that the system correctly chooses and upgrades to the latest version.


Convly:
> Does the system skip the checks on the latest minor.patch of the current major?
> 
> The risk here is high, especially with the future major versions.
> 
> I would rather do `latest` an alias of `major`

Ideas from internal discussions:
- Skip checks on latest minor.patch for current major
- Adding a confirmation prompt if the upgrade triggers a major version change

Convly:
@maccomaccomaccomacco here's a behavioral proposal for the prompt

image

innerdvations (dismissed):
LGTM, will test when it gets down to the full upgrade change PR

Convly:
Related doc PR: https://github.com/strapi/documentation/pull/2259

maccomaccomaccomacco (dismissed):

PR #21739, opened 2024-10-09
chore: re-enable history e2e tests

remidej:
### What does it do?

Re-enables e2e tests, and makes them not flaky anymore

PR #21770, opened 2024-10-10
chore: add test workflow filters for each test type

innerdvations:
### What does it do?

implements filters specific for each type of integration/e2e tests so we run exactly what is necessary

### Why is it needed?

e2e tests are currently not running when e2e tests are changed, because we were piggybacking on the setup for api tests. now all the test types should just track themselves

### How to test it?

CI should run exactly what it needs to. For example:
- a change to a backend file will trigger runs for api, cli, e2e, and unit:back
- a change to a frontend file will trigger runs for unit:front and e2e
- a change to an api test or utility will trigger runs only for api
- a change to an e2e test or utility will trigger runs only for e2e
- a change to a cli test or utility will trigger runs only for cli
- a change to github workflows will trigger all tests (this is a trade-off; the test results themselves will never be affected by a workflow change, however, workflow changes are difficult to test if they don't trigger tests to run)
- any skipped runs should not prevent test_aggregate from passing

I'm not sure, but I think if you branch from this PR and target it in github it will follow the workflows from this PR, that would be one way to test it if that's true

### Related issue(s)/PR(s)

DX-1699

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -6,7 +6,6 @@ backend:
   - 'packages/**/strapi-server.js'
   - 'packages/{utils,generators,cli,providers}/**'
   - 'packages/core/*/{lib,bin,ee,src}/**'
-  - 'tests/api/**'
   - 'packages/core/database/**'
 frontend:
   - '.github/actions/yarn-nm-install/*.yml'
@@ -16,3 +15,9 @@ frontend:
   - 'packages/**/admin/ee/admin/**'
   - 'packages/**/strapi-admin.js'
   - 'packages/admin-test-utils/**'
+api:
+  - 'tests/api/**'
+e2e:
+  - 'tests/e2e/**'
+cli:
+  - 'tests/cli/**'
--- a/.github/workflows/skipped_tests.yml
+++ b/.github/workflows/skipped_tests.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [20]
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [20]
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -49,6 +49,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build]
     name: 'typescript'
+    strategy:
+      matrix:
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -58,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [18, 20]
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -78,7 +81,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [18, 20]
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -86,6 +89,9 @@ jobs:
     name: 'CLI Tests'
     needs: [changes, build]
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
@@ -115,7 +121,7 @@ jobs:
     name: '[CE] API Integration (postgres, node: ${{ matrix.node }})'
     strategy:
       matrix:
-        node: [18, 20]
+        node: [20, 22]
     steps:
       - run: echo "Skipped"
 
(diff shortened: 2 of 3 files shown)


remidej:
Thank you I'll use my e2e PR to test this

alexandrebodin (changes requested):
Make sure to update the skipped_test workflow accordingly :)

remidej:
I think it should have run the e2e tests for my PRs, since I merged these changes into my branch: https://github.com/strapi/strapi/pull/21739

innerdvations:
I updated skipped_tests, it had the incorrect node versions

alexandrebodin (changes requested):
Not sure the skipped test will work the way they are intended. and with the recent change of using the aggregate test I wonder if it even makes sense anymore 🤔 

I think we would be much better off having a more dynamic aggregate that will pass only if the test that should have run have passed (so taking into accounts the different "when") 

Does that make sense ?

innerdvations:
The way it seemed to be working the last time I updated it is that if a workflow in tests.yml is skipped with an `if`, it uses the result from skipped_tests instead. So every tests workflow needs to have a duplicate with the same name in skipped_tests, or else aggregate stalls forever waiting for something with that name to complete successfully.

It would definitely be better to have a more dnyamic aggregate though. Or could we delete skipped_tests.yml entirely and have tests.yml include both conditions? basically `if conditions run test; if !conditions echo 'skipped'`?

alexandrebodin:
> The way it seemed to be working the last time I updated it is that if a workflow in tests.yml is skipped with an `if`, it uses the result from skipped_tests instead. So every tests workflow needs to have a duplicate with the same name in skipped_tests, or else aggregate stalls forever waiting for something with that name to complete successfully.
> 
> It would definitely be better to have a more dnyamic aggregate though. Or could we delete skipped_tests.yml entirely and have tests.yml include both conditions? basically `if conditions run test; if !conditions echo 'skipped'`?

According to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context it looks like a skipped test will still have an output and we are only checking for cancelled or failure states so the skipped_tests files should already be useless

innerdvations:
I had to do [this](https://github.com/strapi/strapi/pull/21084) a little while back because without it when e2e tests were skipped, aggregate was stalling.

But now I'm playing around with the idea of a single check for all the tests that reduces the file down to like 10% of the current size and makes it much easier to manage. If only it were easier to test this stuff 😆

innerdvations:
> According to [docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context) it looks like a skipped test will still have an output and we are only checking for cancelled or failure states so the skipped_tests file should already be useless

I think [this](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context:~:text=the%20job%20is%20only%20sent%20to%20the%20runner%20if%20the%20result%20is%20true) is why.

Because we're using "needs" it waits for a result, but github doesn't send the task to the runner at all. So it's not that it completes with a "skipped" status, it's that it's just not there. skipped_tests is allowing us to provide a fallback value -- I've verified it's working like that but I have no idea how or why it works like that.

It seems like the current workflow fixes the immediate problem, but I will post a refactor in the next week or so, because I think we can reduce this to something much simpler and delete skipped_tests altogether.

remidej:
I'll let others review the code part but I confirm this does fix the issue on my e2e-only PR ✅

alexandrebodin:
> > According to [docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context) it looks like a skipped test will still have an output and we are only checking for cancelled or failure states so the skipped_tests file should already be useless
> 
> I think [this](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context:~:text=the%20job%20is%20only%20sent%20to%20the%20runner%20if%20the%20result%20is%20true) is why.
> 
> Because we're using "needs" it waits for a result, but github doesn't send the task to the runner at all. So it's not that it completes with a "skipped" status, it's that it's just not there. skipped_tests is allowing us to provide a fallback value -- I've verified it's working like that but I have no idea how or why it works like that.
> 
> It seems like the current workflow fixes the immediate problem, but I will post a refactor in the next week or so, because I think we can reduce this to something much simpler and delete skipped_tests altogether.

> so the 'skipped' result documented can't actually be seen in there. that's too bad 🤷  Having a custom action for the aggregate test will probably be the best way to make things dynamic at this point. it shouldn't be too difficult :) 👍

alexandrebodin (approved):

PR #21771, opened 2024-10-10
chore: ignore vscode settings file

innerdvations:
### What does it do?

removes .vscode/settings.json from the repo

TODO:

- [x] find an alternative solution to the parsing error

### Why is it needed?

- so that strapi devs can have their own user config without it tracking as changed in git
- this was an old file that was needed to help vscode understand lint in the monorepo but doesn't seem necessary anymore

### How to test it?

- using vscode, try making code changes in various parts of the monorepo and your code should still do live linting properly


Convly (approved):
Are we sure it's not useful/required now?

alexandrebodin (changes requested):
I'm seeing this warning now. I can re-add the file locally for my usecase but it might be a problem for everyone ? 

Screenshot 2024-10-10 at 10 13 09

innerdvations:
Spent a few minutes looking at this today, since Cursor insists on updating settings.json every time it loads. I didn't completely solve it but I'll leave a note to my future self (or whoever else is annoyed by this issue):

It seems to be caused by the way we extend the base config file. When we remove the vscode config, it tries looking for each project config ('tsconfig.eslint.json') in that base file dir rather than the project. Solution is probably for each project to explicitly declare its own absolute path to config file rather than trying to do it globally in the base file, but didn't have time to update them all and confirm.

PR #21611, opened 2024-10-07
fix(e2e): remove webkit skip conditions

jhoward1994:
### What does it do?

Describe the technical changes you did.

### Why is it needed?

Describe the issue you are solving.

### How to test it?

Provide information about the environment and the path to verify the behaviour.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #21781, opened 2024-10-10
Chore/test-workflow-filters-e2e

jhoward1994:
Using this to test that https://github.com/strapi/strapi/pull/21770 runs the correct CI based on this change

Commits lifted from https://github.com/strapi/strapi/pull/21611


innerdvations:
nice, looks like CLI, API, and unit front and back are skipped as expected 🎉

PR #21795, opened 2024-10-11
chore: upgrade mysql2 package

innerdvations:
### What does it do?

generates new projects with mysql 3.9.8 instead of 3.9.4

### Why is it needed?

CVE-2024-21511
CVE-2024-21512

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/examples/getstarted/package.json
+++ b/examples/getstarted/package.json
@@ -25,7 +25,7 @@
     "@strapi/strapi": "workspace:*",
     "better-sqlite3": "11.3.0",
     "lodash": "4.17.21",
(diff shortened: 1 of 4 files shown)


Convly (approved):
I'm wondering whether we should upgrade the already generated app's dependencies (codemod) if they're in the 3.9.x range :thinking: 

Thoughts?

innerdvations:
That would be a great feature in the upgrade tool, but I wouldn't do it on a case-by-case basis.

@maccomaccomaccomacco what do you think about codemods for upgrading third party packages with security issues in a strapi project?

PR #21798, opened 2024-10-11
chore: upgrade plop to 4.0.1

innerdvations:
### What does it do?

- updates `plop` to 4.1.0
- fixes code to handle the breaking changes

Note that `node-plop` is being held back because of breaking changes that affected unit and api test setup, but since it doesn't include any vulnerable packages it doesn't need to be upgraded anyway

### Why is it needed?

the old version of plop was years out of date and had too many security vulnerabilities in subdependencies to try to list accurately here (none of which caused Strapi to be vulnerable aside from potentially from attacks from admin accounts, but needed to be fixed regardless)

the big one is that vulnerable versions of micromatch are finally completely removed here

### How to test it?

Experimental for QA: 0.0.0-experimental.9a7831c04ba07c03200a7d90706d832f63c0f3a9

We have:
- e2e tests for ctb
- cli test for very basic generators command

 you can manually test file generation for creating and editing content types in the CTB and running all the generators from the CLI

### Related issue(s)/PR(s)

DX-1574


packages/generators/generators/src/index.ts
```diff
@@ -1,28 +1,33 @@
 import { join } from 'node:path';
 import { Plop, run } from 'plop';
```
alexandrebodin, line 2: isn't it ESM only now ?

alexandrebodin (commented):
Would be a nice timing to add some tests on the generators ;)

packages/generators/generators/src/index.ts
innerdvations: Yeah, not sure why this was working. I'm getting a unit test error that seems related, will figure it out and update

innerdvations:
> Would be a nice timing to add some tests on the generators ;)

for unit tests, we probably could, but I don't know where to start 😆 do you notice anything in particular that would be worth adding here?

for the cli tests, we don't have the proper setup to test this yet, but it's in the backlog and coming up soon
for the e2e tests, we already have CTB tests but are also expanding on them soon

Convly (approved):

alexandrebodin (approved):

innerdvations:
Experimental for QA: 0.0.0-experimental.9a7831c04ba07c03200a7d90706d832f63c0f3a9

Convly (approved):
Tested and works well

PR #21406, opened 2024-09-20
fix: Filter by text field with "&" symbol doesn't work #20271

chhedadhruv:
What does it do?
Fixes a bug where filtering by text fields containing the "&" symbol fails due to improper encoding and decoding of query parameters.

Why is it needed?
Without this fix, users cannot correctly filter records by text fields containing special characters such as "&", which significantly hampers the usability of the filtering functionality.

How to test it?
Create an entity in a collection type with a text field containing the "&" symbol.
Example: Create an article titled "Test article & issue reproduce".
Go to the list of the collection type.
Use the "Filters" button to select the text field, "is" operand, and text with the "&" symbol.
Click "Add filter".
Verify that the filter is applied correctly and the appropriate records are displayed.

image

Related issue(s)/PR(s)
#20271

PR #21800, opened 2024-10-11
fix: broken list view pagination

### LINKED ISSUE #21789: Content Manager pagination is broken as of 5.0.3

- Node.js version: v20
- NPM version: yarn v1
- Strapi version: 5.0.4 and 5.0.3
- Database: SQLite
- Operating system: Linux Mint 21.2

### Describe the bug

When paginating within the content-manager, changes pages yields an invalid admin REST Query:

`http://localhost:1337/content-manager/collection-types/api::article.article?[object%20Object]?page=2&pageSize=10&sort=title%3AASC`

Specifically the problem is here: `[object%20Object]?page=2`

1. Create a content-type with any field
2. Create more than 10 entries
3. Try to paginate

### Expected behavior

Should paginate properly :) 

### Screenshots

Reproduction video:

https://www.loom.com/share/c5a773d26cb74308b920e6df0235a9c7?sid=50da458d-2463-4147-a048-661efeb8eab4

### Code snippets

N/A

### Additional context

N/A

remidej:
### What does it do?

Fixes the way query params are written into the URL of the list

It seems like when passing a params object in the config, RTK query stringifies it properly, but here an object was put directly into the string.

### How to test it?

Have 2+ pages of content. Navigate between pages. Then content should reflect the page you're on
### Related issue(s)/PR(s)

fixes #21789
introduced in #21406

### DIFF
--- a/packages/core/content-manager/admin/src/services/documents.ts
+++ b/packages/core/content-manager/admin/src/services/documents.ts
@@ -166,7 +166,7 @@ const documentApi = contentManagerApi.injectEndpoints({
       }
     >({
       query: ({ model, params }) => ({
-        url: `/content-manager/collection-types/${model}${params ? `?${params}` : ''}`,
+        url: `/content-manager/collection-types/${model}`,
         method: 'GET',
         config: {
           params,
--- a/tests/e2e/tests/content-manager/listview.spec.ts
+++ b/tests/e2e/tests/content-manager/listview.spec.ts
@@ -19,6 +19,33 @@ test.describe('List View', () => {
     await expect(page.getByRole('link', { name: /Create new entry/ }).first()).toBeVisible();
   });
 
+  test('Entries should be paginated', async ({ page }) => {
+    // Go to the authors list view, all three entries should be visible
+    await page.getByRole('link', { name: 'Content Manager' }).click();
+    await page.getByRole('link', { name: 'Author' }).click();
+    await expect(page.getByRole('gridcell', { name: 'Draft' })).toHaveCount(3);
+    await expect(page.getByRole('link', { name: 'Next page' })).not.toBeVisible();
+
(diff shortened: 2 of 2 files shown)


alexandrebodin (commented):
so this was borken way before 5.0.3 ? 

Can you add some test on RTK ?

remidej:
According to our release notes the would have been released in 5.0.2

oiorain:
QA LGTM.
Nice job! ⚡

jhoward1994:
Thanks this looks good, a test to cover it would be good. Maybe `tests/e2e/tests/content-manager/listview.spec.ts` is the place to add it?

remidej:
@jhoward1994 completely agree, I'm working on adding a pagination e2e test to this PR

packages/core/content-manager/admin/src/services/documents.ts
```diff
@@ -166,7 +166,7 @@ const documentApi = contentManagerApi.injectEndpoints({
       }
     >({
       query: ({ model, params }) => ({
-        url: `/content-manager/collection-types/${model}${params ? `?${params}` : ''}`,
+        url: `/content-manager/collection-types/${model}`,
```
alexandrebodin, line 169: Can you confirm the original fix still works ?

remidej: yes i tested this manually

oiorain:
QA LGTM

alexandrebodin (approved):
code LGTM 👍

neilsmith:
Just in case it's helpful: as well as breaking pagination, I believe the same issue also broke searching for multiple words (i.e. searching for "foo" would find a record "foo bar baz" - but searching for "foo bar" would yield no results.)

This fix also seems to resolve that issue!

Don't know if that would be worth covering in an e2e test too?

PR #21801, opened 2024-10-11
chore: fix mysql workflow

innerdvations:
### What does it do?

use mysql for dbclient in api_ce_mysql workflow

### Why is it needed?

api_ce_mysql tests are not receiving a dbclient value and are probably currently running with sqlite

### How to test it?

CI should work

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -361,7 +361,7 @@ jobs:
         uses: ./.github/actions/run-build
       - uses: ./.github/actions/run-api-tests
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -361,7 +361,7 @@ jobs:
         uses: ./.github/actions/run-build
       - uses: ./.github/actions/run-api-tests
         with:
-          dbOptions: '--dbclient=${{ matrix.db_client }} --dbhost=localhost --dbport=3306 --dbname=strapi_test --dbusername=strapi --dbpassword=strapi'
```
Convly, line 364: Where is db_client supposed to come from though? Is it set somewhere else?

Convly (approved):

Issue #21346, opened 2024-09-18
[Design] Display issue when scrolling roles permissions


yanniskadiri:
See issue at the top, with the header 👇

![Image](https://github.com/user-attachments/assets/171b509a-097e-4258-b329-dba8fb424db0)

Adzouz:
Hi @yanniskadiri, thanks for the issue. Fixed in 5.15.0 with this PR: https://github.com/strapi/strapi/pull/23634.
Closing 🚀

PR #21826, opened 2024-10-13
fix: display issue while scrolling on roles view

normydev:
### What does it do?

It updates the z-index of the fixed header to 2.

### Why is it needed?

To make the scroll better without hindering the top header view.

### How to test it?

- Navigate to Roles in admin
- Try to edit any role
- Scroll through the rows

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/21346


Adzouz:
Hi @normydev, thank you so much for the fix and so sorry I haven't see it earlier because I already pushed the fix... It's available from v5.15.0.
Closing this PR but thanks again for your time and dedication, I appreciate 🙏
Have a great day 🚀

PR #21617, opened 2024-10-07
chore: updated msw and migrated from rest to http and HttpResponse

nilsingwersen:
### What does it do?

I updated msw to 2 the newest version and migrated the tests.

### Why is it needed?

In the msw 1 version there is a peer dependency to an older typescript version which throws a warning when installing strapi.

### How to test it?

I tested it by running all frontend tests.

### Related issue(s)/PR(s)

refers to #21201

PR #21833, opened 2024-10-14
chore: upgrade cookie to 0.7.1

innerdvations:
### What does it do?

upgrades locked versions of:
- cookie to 0.7.1
- express from 4.21.0 to 4.21.1

### Why is it needed?

CVE-2024-47764

cookie is only a subdependency but it's used by:

- @apollo/server (via express)
- @docusaurus/core

but we're already on the latest version of those packages and they use a version range that includes the updated cookie, so we just need the yarn.lock bump

### How to test it?

- @apollo/server (via express) -- check graphql
- @docusaurus/core -- check contributor docs

### Related issue(s)/PR(s)

msw still has a dependency on cookie that might need to be upgraded, PR for that can be found in https://github.com/strapi/strapi/pull/21617

DX-1709

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -7,6 +7,7 @@ backend:
   - 'packages/{utils,generators,cli,providers}/**'
   - 'packages/core/*/{lib,bin,ee,src}/**'
   - 'packages/core/database/**'
+  - 'yarn.lock'
 frontend:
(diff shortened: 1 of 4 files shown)


Convly (approved):

Convly (approved):
Both GraphQL queries and the contrib doc worked for me. nice.

Issue #21808, opened 2024-10-11
Content Manager: With >10 relations, display order is wrong


CodeToTee:
## Bug report
When a one-to-many relationship is populated with more than ten items, the UI loses the ordering and displays them out of order.  The order is correct in the database and REST API queries return the proper order.

### Required System information
- Node.js version: v20
- NPM version: Yarn 1.22.22
- Strapi version: 5.0.6
- Database: SQLite
- Operating system: MacOS
- Is your project Javascript or Typescript: Typescript

### Describe the bug
In the Content Manager, as you `Load More` relations to view them, ordering is lost when the third page is loaded.

### Steps to reproduce the behavior
1. Instantiate a Strapi project via `npx create-strapi@latest` and install the example data.  The Categories and Articles collections have a one-to-many relationship (one category to many articles).
2. Create 11 test Articles (I named mine `test1` through `test11`).
3. Create a `test` Category.
4. In the `test` Category, add a relation for each of the 11 test articles.  Add them in-order (`test1`-`test11`)
5. Refresh the page to reload the `test` Category.  The last 5 relations are shown (`test7`-`test11`).  (IMHO, the first five should be displayed, but that's not the issue with this bug.)
6. Click `Load More` and the next five relations are loaded (`test2`-`test6`).  The displayed relations are in the proper order.
7. Click `Load More` to load the last relation.  The last relation (`test1`) is loaded in the wrong place.  For me, it's loaded between `test6` and `test7`. (see last screenshot below)

### Expected behavior
As you `Load More` relations, the UI should maintain the proper relationship ordering.

### Screenshots
<img width="1601" alt="Screenshot 2024-10-11 at 4 12 03 PM" src="https://github.com/user-attachments/assets/cafc10cd-75e3-4acd-b47b-d66803357b2d">
<img width="1585" alt="Screenshot 2024-10-11 at 4 12 46 PM" src="https://github.com/user-attachments/assets/40a9b6ec-6ef5-4a1c-b462-0c6a51a9215e">
<img width="1609" alt="Screenshot 2024-10-11 at 4 13 22 PM" src="https://github.com/user-attachments/assets/3009a25a-6137-4682-88a6-4f523e952eb8">

### Code snippets
N/A

### Additional context
I ran a SQL query on the database to confirm the ordering of the relationships is correct (i.e. the order I added them).

PR #21855, opened 2024-10-15
fix: order when loading more than 10 relations 

Marc-Roig:
### What does it do?
Fix loading more than 2 pages of relations in the content manager. 

Relations are assigned a __temp_key__ value based on their order (a0, a1, ..., z1, z2) . Each time a page was loaded the new temp keys were based from the first temp key of the list (a0), but that same value was used to compute the temp keys for the 3rd , ... pages. So the order was messed up.

### How to test it?

Load more than 10 relations in a relational field.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/21808

### DIFF
--- a/packages/core/content-manager/admin/src/services/relations.ts
+++ b/packages/core/content-manager/admin/src/services/relations.ts
@@ -67,8 +67,7 @@ const relationsApi = contentManagerApi.injectEndpoints({
              * Relations will always have unique IDs, so we can therefore assume
              * that we only need to push the new items to the cache.
              */
-
-            currentCache.results.push(...prepareTempKeys(newItems.results, currentCache.results));
+            currentCache.results = prepareTempKeys([...newItems.results, ...currentCache.results]);
             currentCache.pagination = newItems.pagination;
           } else if (newItems.pagination.page === 1) {
             /**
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/services/relations.ts
```diff
@@ -178,11 +177,8 @@ const relationsApi = contentManagerApi.injectEndpoints({
  * @description Adds a `__temp_key__` to each relation item. This gives us
  * a stable identifier regardless of it's ids etc. that we can then use for drag and drop.
  */
-const prepareTempKeys = (relations: RelResult[], existingRelations: RelationResult[] = []) => {
-  const [firstItem] = existingRelations.slice(0);
-
-  const keys = generateNKeysBetween(null, firstItem?.__temp_key__ ?? null, relations.length);
-
+const prepareTempKeys = (relations: RelResult[]) => {
+  const keys = generateNKeysBetween(null, null, relations.length);
```
remidej, line 181: I have no doubt this works, but does it mean that when loading that appending new pages, the keys for the previously loaded pages get regenerated?

Asking because if the keys change it triggers re-renders that could be avoided (though it's not a huge deal)

Marc-Roig: good point :) ! I will only regenerate the keys for new ones

remidej (commented):
I see an issue when trying to save after reordering _unsaved_ relations


https://github.com/user-attachments/assets/81b4a0f9-d5df-48a5-b8f9-0ae51a677b40

Marc-Roig:
will take a look! good catch

Marc-Roig:
@remidej I am not being able to replicate your issue (using the same content types and relation) , do you have any API error or cli error?

remidej (approved):
I'm not seeing my error again after doing the same actions. Weird! Looks like I messed up something last time. Looks all good now

Issue #21577, opened 2024-10-03
v4: Missing type for Attribute.column


iamandrewluca:
### Required System information

- Node.js version: v20.17.0
- NPM version: 10.8.3
- Strapi version: 4.24.2
- Database: postgres
- Operating system: darwin-arm64
- Is your project Javascript or Typescript: JavaScript with JSDoc

### Describe the bug

In a project, I started converting the API `schema.json` files to `schema.js` to be able to type them using JSDoc and types provided by Strapi

e.g.

```js
/** @type {import('@strapi/types').Schema.CollectionType} */
module.exports = {
	kind: 'collectionType',
	// ...
```

Everything works fine, even Strapi types can be generated. The problem is when having attributes that use the `column` property to customize the DB, it is not typed

e.g.

```js
pin_sent: {
	type: 'boolean',
	default: false,
	column: {
		defaultTo: false,
	},
},
```

At the moment to get autocompletion, I manually type the field, that gives the autocompletion.

```js
pin_sent: {
	type: 'boolean',
	default: false,
	/** @type {Partial<import('@strapi/database/dist/schema').Column>} */
	column: {
		defaultTo: false,
	},
},
```

### Expected behavior

When using the `CollectionType` type, the `column` field of an attribute should be type correctly.

### Screenshots

| Without Typing | With Typing |
|--------|--------|
| <img width="661" alt="image" src="https://github.com/user-attachments/assets/a22997c4-ee3a-43a6-b174-b7be148b1437"> | <img width="688" alt="image" src="https://github.com/user-attachments/assets/fcef16be-7eb2-4e43-8611-6725435485ce"> |

iamandrewluca:
I made an attempt in the linked PR.

https://github.com/strapi/strapi/issues/21577

Convly:
Closed by #21580, thanks for your contribution

PR #21858, opened 2024-10-15
fix(strapi/types): add Attribute column type

iamandrewluca:
### What does it do?

Types the `Attribute.column` for Strapi v5

### Why is it needed?

Missing `column` type when create schema attributes

### Related issue(s)/PR(s)

#21577 #21580

cc @Convly


Convly (approved):

Convly (changes requested):
It seems the prettier check fails on this. Could you format the code to make it green? :slightly_smiling_face:

iamandrewluca:
Fixed

Convly (approved):

Issue #21852, opened 2024-10-15
Mismatch between docs and Strapi UI for findOne endpoint


pjosxyz:
### Describe the bug

For findOne in (admin/settings/users-permissions/roles/) the endpoint displayed in the UI is /api/collection/:id (screenshot attached)
But this endpoint returns a 404 error.

In the v5 documentation [docs](https://docs.strapi.io/dev-docs/api/rest#endpoints) the endpoint given is /api/collection/:documentId, which works and returns data.

### Expected behavior

I think the Strapi UI should show the same working endpoint described in the documentation: /api/collection/:documentId

### Screenshots

<img width="1031" alt="image" src="https://github.com/user-attachments/assets/495197f9-f687-4f7e-9baf-3d85ce2f58e9">

ishsarin:
Hi, could you please assign this to me. I would like to work on it.

PR #21869, opened 2024-10-16
fix: change name of id param in content api routes

Marc-Roig:
### What does it do?

Rename content api route param name to :documentId (was :id before), this was specially confusing in the content api permissions UI

Before: 
image

After:
image

### Why is it needed?

To make it clear which parameter users should use.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/21852

### DIFF
--- a/packages/core/core/src/core-api/routes/index.ts
+++ b/packages/core/core/src/core-api/routes/index.ts
@@ -42,7 +42,7 @@ const getCollectionTypeRoutes = ({ uid, info }: Schema.ContentType) => {
     },
     findOne: {
       method: 'GET',
-      path: `/${info.pluralName}/:id`,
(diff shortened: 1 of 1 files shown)


alexandrebodin (approved):
TY !

Marc-Roig:
@alexandrebodin just realized this is a breaking change for the content api. as the controller context parameter is named the same as the one in the route

alexandrebodin:
> @alexandrebodin just realized this is a breaking change for the content api. as the controller context parameter is named the same as the one in the route. I will think of something else, as I don't think its wise to break this atm

Yes you are right ! I thought this was only the "UI" config 🤦

Marc-Roig:
I will check a way to do this only from the UI so its not breaking 🤞

PR #21876, opened 2024-10-16
enhancement: add db info to startup logger

innerdvations:
### What does it do?

adds a "database name" line to the startup logger

mysql:
image

postgres with schema name:
image

sqlite showing filename:
image

### Why is it needed?

just some code I use locally a lot when I'm developing with various dbs but I thought it might be nice to have it as a standard piece of startup info

useful info for debugging

also good to help catch issues where you might be targeting the wrong db like prod instead of dev

### How to test it?

try various db configurations and make sure it works


maccomaccomaccomacco (approved):
Why not>

maccomaccomaccomacco (dismissed):

packages/core/core/src/utils/startup-logger.ts
```diff
@@ -1,9 +1,36 @@
 import chalk from 'chalk';
 import CLITable from 'cli-table3';
 import _ from 'lodash/fp';
+import path from 'node:path';
 
 import type { Core } from '@strapi/types';
 
+const formatDbInfo = (app: Core.Strapi) => {
```
alexandrebodin, line 8: Could you expose a db.getDatabaseInfo() or sth like that to abstract the DB implementation details and just return the basinc info (also thinking about the PR you just made, the less we use the connection object directly outside the better)
Even if those are already accessed somewherese else I think it would be nice to start having better encapsulation :)

alexandrebodin (dismissed):

derrickmehaffy (dismissed):

Convly (commented):
What Alex already said + should we also update the[ `report` command](https://github.com/strapi/strapi/blob/develop/packages/core/strapi/src/cli/commands/report.ts#L28) to streamline the DB data displayed?

Otherwise, LGTM

innerdvations:
> What Alex already said + should we also update the[ `report` command](https://github.com/strapi/strapi/blob/develop/packages/core/strapi/src/cli/commands/report.ts#L28) to streamline the DB data displayed?
> 

I wouldn't update report because it's not really relevant and could leak sensitive information, although something like "(with schema)" after "postgres" could be useful there

packages/core/core/src/utils/startup-logger.ts
```diff
@@ -17,16 +17,23 @@ export const createStartupLogger = (app: Core.Strapi) => {
...
         [chalk.blue('Process PID'), process.pid],
         [chalk.blue('Version'), `${app.config.info.strapi} (node ${process.version})`],
         [chalk.blue('Edition'), app.EE ? 'Enterprise' : 'Community'],
-        [chalk.blue('Database'), app.db?.dialect.client]
+        [chalk.blue('Database'), dbInfo?.client],
+        [chalk.blue('Database name'), dbInfo?.displayName]
       );
 
+      if (dbInfo?.schema) {
+        infoTable.push([chalk.blue('Database Schema'), dbInfo.schema]);
```
innerdvations, line 34:
```suggestion
        infoTable.push([chalk.blue('Database schema'), dbInfo.schema]);
```

jhoward1994 (approved):

Convly (approved):
QAed and works well, thanks :slightly_smiling_face:

PR #21878, opened 2024-10-16
fix: validation message errors

Feranchz:
### What does it do?

If status of entry is published and the action is not publish (means that actions is unpublish) we want to show the right message: "Ready to publish", this PR handle that case.

### How to test it?

1. Create an entry and publish it
2. Assign it to a Release for unpublish
3. The message should say "Ready to unpublish", before was showing "Ready to publish"


remidej (approved):
✅ 
CleanShot 2024-10-16 at 14 52 29@2x

Marc-Roig (approved):

Issue #21367, opened 2024-09-18
Caching issue media-lib new folder


yanniskadiri:
Steps to reproduce:
- Start with 0 folders
- Upload a few images
- Try to move a file from the list view and see there is no sub folder to move to
- click “creat folder” fill and save the form
- try to move the same file again
    - The select is still empty
- Refresh the page
    - do the same operation and the select is now filled correctly

Sam-Phillemon9493:
@yanniskadiri, I can see the newly created folder in the dropdown when we try to move it without refreshing. I’ve also noticed that the dropdown is collapsed by default. 

**Reference Video:**

https://github.com/user-attachments/assets/22d04170-7a31-46eb-b132-246dad03661c


Suggestion: Maybe we should make the dropdown expanded by default, while allowing any subfolders inside it to be collapsed.

Please let me know if I’ve missed anything regarding this issue.

yanniskadiri:
Hey @Sam-Phillemon9493 thanks for checking, I managed to reproduce the issue. You have to refresh to be able to select the sub-folder you just created (see the screen recording).

The issue with expanding the full dropdown is when you have many sub-folders and nested sub-folders. If we do it, the users might have to scroll for a very long time before finding the right one.

https://github.com/user-attachments/assets/d945839c-ad4a-4fd5-84b6-c3d6fef13857

Sam-Phillemon9493:
@yanniskadiri , Thanks for the explanation! I understand the issue now and will work on it.

PR #21893, opened 2024-10-17
fix: fixed the caching issue in media-lib new folder

Sam-Phillemon9493:
### What does it do?
Fixed the new folder not getting displayed when we try to move the asset in media-lib.

### Why is it needed?
To fix issue [21367](https://github.com/strapi/strapi/issues/21367) : Caching issue media-lib new folder

### How to test it?

- Start with 0 folders
- Upload a few images
- Try to move a file from the list view and see there is no sub folder to move to
- click “creat folder” fill and save the form
- try to move the same file again
- The select is still empty
- Refresh the page
- do the same operation and the select is now filled correctly

### Related issue(s)/PR(s)
fixes [21367](https://github.com/strapi/strapi/issues/21367)


markkaylor (approved):
Looks good to me. Tested locally.

Thanks for the contribution @Sam-Phillemon9493 🎈 !

PR #21898, opened 2024-10-17
Feat(cloud-cli): deploy command allows environment to be passed

gonbaum:
### What does it do?

This PR introduces a feature to allow the cloud CLI deploy command to deploy to a specific cloud project environment by using a prompt selection or a flag.

- When the user runs `strapi deploy`, they would be asked to select an environment to deploy
- When the user runs `strapi deploy -e <environment> `, the deployment will be done in the specified environment

### Why is it needed?

With the introductions of Strapi Cloud multi environments, a feature to allow users to deploy to a specific environment from the CLI is needed.

### How to test it?

Pre-test: new options should be displayed on running `strapi deploy -h`:

`  -e, --environment <name>  Specify the environment to deploy`

Test Case 1: NO prompt for Environment Selection

Steps: Run strapi deploy with only 1 environment in a cloud project.
Expected: deploy should proceed as usual

Test Case 2: Prompt for Environment Selection

Steps: Run strapi deploy with multiple environments configured for the project.
Expected: 
- The user is prompted to select an environment from the available options
- All the available environments should be listed
- A quit option allows the user to cancel the deployment action

Test Case 3: Environment name invalid

Steps: Run strapi deploy -e <env-name> with an invalid environment name
Expected: command should reject with a proper log message

Test Case 4: Environment name valid

Steps: Run strapi deploy -e <env-name> with a valid environment name
Expected: 
- Deployment should start right away
- Deployment should be done to the passed environment

### DIFF
--- a/packages/cli/cloud/src/deploy-project/action.ts
+++ b/packages/cli/cloud/src/deploy-project/action.ts
@@ -1,4 +1,5 @@
 import fse from 'fs-extra';
+import inquirer from 'inquirer';
 import path from 'path';
 import chalk from 'chalk';
 import { AxiosError } from 'axios';
@@ -21,6 +22,40 @@ type PackageJson = {
     uuid: string;
   };
 };
+
+interface CmdOptions {
+  environment?: string;
+}
+
+const QUIT_OPTION = 'Quit';
+
+async function promptForEnvironment(environments: string[]): Promise<string> {
+  const choices = environments.map((env) => ({ name: env, value: env }));
+  const { selectedEnvironment } = await inquirer.prompt([
+    {
+      type: 'list',
+      name: 'selectedEnvironment',
+      message: 'Select the environment to deploy:',
+      choices: [...choices, { name: chalk.grey(`(${QUIT_OPTION})`), value: null }],
+    },
+  ]);
+  if (selectedEnvironment === null) {
+    process.exit(1);
+  }
+
+  const { confirm } = await inquirer.prompt([
+    {
+      type: 'confirm',
+      name: 'confirm',
+      message: `Do you want to proceed with deployment to ${chalk.cyan(selectedEnvironment)}?`,
+    },
+  ]);
+
+  if (!confirm) {
+    process.exit(1);
+  }
+  return selectedEnvironment;
+}
 
 async function upload(
   ctx: CLIContext,
@@ -143,7 +178,7 @@ async function getConfig({
   }
 }
 
-export default async (ctx: CLIContext) => {
+export default async (ctx: CLIContext, opts: CmdOptions) => {
   const { getValidToken } = await tokenServiceFactory(ctx);
   const token = await getValidToken(ctx, promptLogin);
   if (!token) {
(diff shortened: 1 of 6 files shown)


packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -22,6 +23,40 @@ type PackageJson = {
   };
 };
 
+interface CmdOptions {
+  environment?: string;
+}
+
+const QUIT_OPTION = 'Quit';
+
+async function promptForEnvironment(environments: string[]): Promise<string> {
```
abdallahmz, line 32: nice function split!

packages/cli/cloud/src/environment/command.ts
```diff
@@ -1,7 +1,12 @@
 import { Command } from 'commander';
 import { defineCloudNamespace } from '../cloud/command';
 
-export function createEnvironmentCommand(command: Command): Command {
-  const cloud = defineCloudNamespace(command);
-  return cloud.command('environment').description('Manage environments for a Strapi Cloud project');
-}
+let environmentCmd: Command | null = null;
+
+export const getEnvironmentCommand = (command: Command): Command => {
```
abdallahmz, line 6: "get" or "create" are okay, if you're looking for other suggestions: `setup` or `initialize` could be more explicit (at least to me)

abdallahmz:
Code reads well, looks nice so far 👍

packages/cli/cloud/src/environment/command.ts
gonbaum: The function will be used in other environment subcommands. Using `create` suggests that we are creating a new command each time it's used, which may be misleading. I believe `get` better conveys the idea of retrieving an existing instance, but I’m leaning now toward `initialize`.

packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -214,6 +251,21 @@ export default async (ctx: CLIContext) => {
     maxSize = 100000000;
   }
 
+  let targetEnvironment: string;
+
+  if (opts.environment) {
+    if (!environments.includes(opts.environment)) {
+      ctx.logger.error(`Environment ${opts.environment} does not exist.`);
```
nathan-pichon, line 258: Maybe we should warn them about not being able to match the environment and proceed to the promptForEnvironment func? 🤔 
Although not sure it would please ppl wanting to use this in CI, but I doubt we have many CI users

gonbaum: The idea was that the user can use the flag (opts) to automate the deployment to an environment. Adding a step that requires human intervention defeats this purpose. We could solve that with a timeout, but I would rather make the "--flag way" as straightforward as possible: IMO the user is supposed to know what they are doing when using flags, if they pass the wrong env value, it should simply fail. The interactive path should be reserved for the regular way of using the command.

nathan-pichon: I'm ok with this approach 👍

nathan-pichon (approved):
lgtm to be merged like that
I'll try something to avoid relying on the `let environmentCmd: Command | null = null;` but it can always be changed later if we want to 👍

PR #21900, opened 2024-10-17
feat: DTS assert count of items sent and received is equal

Bassel17:
### What does it do?

- Adds counters to the remote transfers, so we know how much data is sent, received and successfully added
- does a comparison and throws and error incase of a mismatch

### Why is it needed?

- Throw and error and rollback to prevent data loss during transfers

### How to test it?

- Transfers should work just fine

### DIFF
--- a/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
+++ b/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
@@ -12,12 +12,12 @@ import type {
   ProviderType,
   IAsset,
   TransferStage,
+  Protocol,
 } from '../../../../types';
 import type { Client, Server, Auth } from '../../../../types/remote/protocol';
 import type { ILocalStrapiDestinationProviderOptions } from '../local-destination';
 import { TRANSFER_PATH } from '../../remote/constants';
 import { ProviderTransferError, ProviderValidationError } from '../../../errors/providers';
-import { TransferAssetFlow } from '../../../../types/remote/protocol/client';
 
 export interface IRemoteStrapiDestinationProviderOptions
   extends Pick<ILocalStrapiDestinationProviderOptions, 'restore' | 'strategy'> {
@@ -44,7 +44,7 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
 
   transferID: string | null;
 
-  stats!: { [TStage in Exclude<TransferStage, 'schemas'>]: { count: 0 } };
+  stats!: { [TStage in Exclude<TransferStage, 'schemas'>]: { count: number } };
 
   constructor(options: IRemoteStrapiDestinationProviderOptions) {
     this.options = options;
@@ -110,7 +110,7 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
     try {
       const res = await this.dispatcher?.dispatchTransferStep<{
         ok: boolean;
-        stats: { started: number; finished: number };
+        stats: Protocol.Client.Stats;
       }>({
         action: 'end',
         step,
@@ -136,7 +136,7 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
   ) {
     try {
       if (step === 'assets') {
-        const assetMessage = message as TransferAssetFlow[];
+        const assetMessage = message as Protocol.Client.TransferAssetFlow[];
         this.stats[step].count += assetMessage.filter((data) => data.action === 'start').length;
       } else {
         this.stats[step].count += message.length;
@@ -183,13 +183,12 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
         }
         const { error, stats } = await this.#endStep(step);
 
-        if (
-          stats &&
-          (stats.started !== this.stats[step].count || stats.finished !== this.stats[step].count)
-        ) {
+        const { count } = this.stats[step];
+
+        if (stats && (stats.started !== count || stats.finished !== count)) {
           callback(
             new Error(
-              `Data missing: sent ${this.stats[step].count} ${step}, recieved ${stats.started} and added ${stats.finished} ${step}`
+              `Data missing: sent ${this.stats[step].count} ${step}, recieved ${stats.started} and saved ${stats.finished} ${step}`
             )
           );
         }
--- a/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
+++ b/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
@@ -6,11 +6,18 @@ import type { Struct, Utils } from '@strapi/types';
 
 import { createDispatcher, connectToWebsocket, trimTrailingSlash } from '../utils';
 
-import type { IDestinationProvider, IMetadata, ProviderType, IAsset } from '../../../../types';
+import type {
+  IDestinationProvider,
+  IMetadata,
+  ProviderType,
+  IAsset,
+  TransferStage,
+} from '../../../../types';
 import type { Client, Server, Auth } from '../../../../types/remote/protocol';
 import type { ILocalStrapiDestinationProviderOptions } from '../local-destination';
 import { TRANSFER_PATH } from '../../remote/constants';
 import { ProviderTransferError, ProviderValidationError } from '../../../errors/providers';
+import { TransferAssetFlow } from '../../../../types/remote/protocol/client';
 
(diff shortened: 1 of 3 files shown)


packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -6,11 +6,18 @@ import type { Struct, Utils } from '@strapi/types';
...
+  IMetadata,
+  ProviderType,
+  IAsset,
+  TransferStage,
+} from '../../../../types';
 import type { Client, Server, Auth } from '../../../../types/remote/protocol';
 import type { ILocalStrapiDestinationProviderOptions } from '../local-destination';
 import { TRANSFER_PATH } from '../../remote/constants';
 import { ProviderTransferError, ProviderValidationError } from '../../../errors/providers';
+import { TransferAssetFlow } from '../../../../types/remote/protocol/client';
```
Convly, line 20:
```suggestion
import type { TransferAssetFlow } from '../../../../types/remote/protocol/client';
```

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -37,11 +44,24 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
 
   transferID: string | null;
 
+  stats!: { [TStage in Exclude<TransferStage, 'schemas'>]: { count: 0 } };
```
Convly, line 47:
```suggestion
  stats!: { [TStage in Exclude<TransferStage, 'schemas'>]: { count: number } };
```

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -78,33 +101,48 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
+
     return null;
   }
 
   async #endStep<T extends Client.TransferPushStep>(step: T) {
     try {
-      await this.dispatcher?.dispatchTransferStep({ action: 'end', step });
+      const res = await this.dispatcher?.dispatchTransferStep<{
+        ok: boolean;
+        stats: { started: number; finished: number };
```
Convly, line 113: We're using the `{ started: number; finished: number }` type in a lot of places, it might be worth creating a dedicated type for it, what do you think?

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -78,33 +101,48 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
+    this.stats[step] = { count: 0 };
+
     return null;
   }
 
   async #endStep<T extends Client.TransferPushStep>(step: T) {
     try {
-      await this.dispatcher?.dispatchTransferStep({ action: 'end', step });
+      const res = await this.dispatcher?.dispatchTransferStep<{
+        ok: boolean;
```
Convly, line 112: I think we even have a type for the return value of those dispatch calls (in the protocol types). Could you just verify we can't re-use/modify them instead of writing it raw here

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -78,33 +101,48 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
+        stats: { started: number; finished: number };
+      }>({
+        action: 'end',
+        step,
+      });
+
+      return { stats: res?.stats, error: null };
```
Convly, line 119: To make it consistent with the rest of the possible response (goes from `Stats | null | undefined` to `Stats | null`)

```suggestion
      return { stats: res?.stats ?? null, error: null };
```

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -78,33 +101,48 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
 
   async #streamStep<T extends Client.TransferPushStep>(
     step: T,
-    data: Client.GetTransferPushStreamData<T>
+    message: Client.GetTransferPushStreamData<T>
   ) {
     try {
-      await this.dispatcher?.dispatchTransferStep({ action: 'stream', step, data });
+      if (step === 'assets') {
+        const assetMessage = message as TransferAssetFlow[];
```
Convly, line 139: So weird that it doesn't narrow it down automatically :disappointed:

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -78,33 +101,48 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
+        this.stats[step].count += assetMessage.filter((data) => data.action === 'start').length;
+      } else {
+        this.stats[step].count += message.length;
```
Convly, line 142: Does it treat `message` as an array or any here? (just to know if the inference somewhat works or not)

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -143,9 +181,20 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
             return callback(streamError);
           }
         }
-        const e = await this.#endStep(step);
+        const { error, stats } = await this.#endStep(step);
+
+        if (
+          stats &&
+          (stats.started !== this.stats[step].count || stats.finished !== this.stats[step].count)
+        ) {
```
Convly, line 189: Just to improve readability

```suggestion
        const { count } = this.stats[step];
        
        if (stats && (stats.started !== count || stats.finished !== count)) {
```

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
```diff
@@ -143,9 +181,20 @@ class RemoteStrapiDestinationProvider implements IDestinationProvider {
...
+          callback(
+            new Error(
+              `Data missing: sent ${this.stats[step].count} ${step}, recieved ${stats.started} and added ${stats.finished} ${step}`
```
Convly, line 192:
```suggestion
              `Data missing: sent ${this.stats[step].count} ${step}, received ${stats.started} and added ${stats.finished} ${step}`
```

Not sure about the meaning of `received` and `added` here? :thinking:

packages/core/data-transfer/types/remote/protocol/client/transfer/push.d.ts
```diff
@@ -15,7 +15,9 @@ export type GetTransferPushStreamData<T extends TransferPushStep> = {
     step: key;
   } & TransferPushMessage;
 }[T] extends { data: infer U }
-  ? U
+  ? U extends any[] // Check if U is already an array
```
Convly, line 18: Did that actually fix the inference issue? :thinking:

Convly (changes requested):
nice :slightly_smiling_face:

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
Bassel17: recieved as in arrived to remote, added as in successfuly transfered /inserted to the DB

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
Bassel17: message is an array

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
Bassel17: yep I tried but my typescript guru skills failed to make that happen

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
Convly: Alright, then `received` makes sense but maybe "saved" could be easier to understand for the user?

packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts
Convly: Yeah just saw the change in the type below

Convly (approved):

PR #21912, opened 2024-10-18
chore: set up preview e2e tests

remidej:
### What does it do?

Adds the first e2e tests for preview. They'll need to evolve as the feature grows.

### DIFF
--- a/tests/e2e/app-template/config/admin.js
+++ b/tests/e2e/app-template/config/admin.js
@@ -14,4 +14,16 @@ module.exports = ({ env }) => ({
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
   },
+  preview: {
+    enabled: true,
+    config: {
+      handler(uid, { documentId, locale, status }) {
+        if (uid === 'api::product.product') {
+          return null;
+        }
+
+        return `https://strapi.io/preview/${uid}/${documentId}/${locale}/${status}`;
+      },
+    },
+  },
 });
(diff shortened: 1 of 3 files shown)


tests/e2e/app-template/config/admin.js
```diff
@@ -14,4 +14,16 @@ module.exports = ({ env }) => ({
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
   },
+  preview: {
```
markkaylor, line 17: Is this new? Should we document it somewhere?

markkaylor (approved):
Looks like a good start to me

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    page,
+    context,
+    browser,
+  }) => {
+    // Open an edit view for a content type that has preview
+    await clickAndWait(page, page.getByRole('link', { name: 'Content Manager' }));
+    await clickAndWait(page, page.getByRole('link', { name: 'Article' }));
+    await clickAndWait(page, page.getByRole('gridcell', { name: /west ham post match/i }));
+
+    // Check that the preview link can be copied. Webkit doesn't support clipboard-write
```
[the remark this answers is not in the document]
remidej: yeah it just throws in the context of playwright :/

tests/e2e/app-template/config/admin.js
remidej: No it's not documented yet but we're in contact with @pwizla to have that ready for launch. Currently it's also under an undocumented future flag until it's ready

Marc-Roig (dismissed):

remidej:
I removed the clipboard check because the playwright-specific issues outweigh the confidence it gives imo

markkaylor (approved):

PR #21914, opened 2024-10-18
fix: Reverse proxy config 

### LINKED ISSUE #21184: Install Strapi v5 in Sub-Path on Server not working

alexandrebodin:
### What does it do?

Fixes reverse proxy compatibility without involved prefixing strapi routes. Make it iso with v4 behavior

Alternative to https://github.com/strapi/strapi/pull/21507
fixes #21184


Convly (commented):
Dumping the raw output of my local test, haven't tested to use a reverse proxy yet, I wanted to make sure the rest is working fine and do some first testing from the train.

```json
[
   {
      "serverUrl":"",
      "serverAbsoluteUrl":"http://localhost:1337",
      "adminUrl":"/admin",
      "adminAbsoluteUrl":"http://localhost:1337/admin",
      "adminPath":"/admin",
      "adminBaseName":"/admin",
      "serverPath":"/",
      "note": "Working fine"
   },
   {
      "serverUrl":"",
      "serverAbsoluteUrl":"http://localhost:1337",
      "adminUrl":"/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/dashboard",
      "adminPath":"/dashboard",
      "adminBaseName":"/dashboard",
      "serverPath":"/",
      "note": "Working fine"
   },
   {
      "serverUrl":"",
      "serverAbsoluteUrl":"http://localhost:1337",
      "adminUrl":"http://localhost:1337/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/dashboard",
      "adminPath":"http://localhost:1337/dashboard",
      "adminBaseName":"/dashboard",
      "serverPath":"/",
      "note": "Working fine"
   },
   {
      "serverUrl":"http://localhost:1337/cms",
      "serverAbsoluteUrl":"http://localhost:1337/cms",
      "adminUrl":"http://localhost:1337/cms/admin",
      "adminAbsoluteUrl":"http://localhost:1337/cms/admin",
      "adminPath":"/admin",
      "adminBaseName":"/cms/admin",
      "serverPath":"/cms",
      "note":"localhost:1337/admin don't return 404, but only a blank page (network requests are fine), (vite vs requests issue?), both /api and /cms/api works"
   },
   {
      "serverUrl":"http://localhost:1337/cms",
      "serverAbsoluteUrl":"http://localhost:1337/cms",
      "adminUrl":"http://localhost:1337/cms/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/cms/dashboard",
      "adminPath":"/dashboard",
      "adminBaseName":"/cms/dashboard",
      "serverPath":"/cms",
      "note":"same as above"
   },
   {
      "serverUrl":"/cms",
      "serverAbsoluteUrl":"http://localhost:1337/cms",
      "adminUrl":"/cms/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/cms/cms/dashboard",
      "adminPath":"/dashboard",
      "adminBaseName":"/cms/cms/dashboard",
      "serverPath":"/cms",
      "note":"/cms/cms/dashboard ??? (doesn't even load using this url, and vite fails on /cms/dashboard but tries to load anyway)"
   },
   {
      "serverUrl":"/cms",
      "serverAbsoluteUrl":"http://localhost:1337/cms",
      "adminUrl":"/cms/cms/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/cms/cms/cms/dashboard",
      "adminPath":"/cms/dashboard",
      "adminBaseName":"/cms/cms/cms/dashboard",
      "serverPath":"/cms",
      "note":"doesn't detect well common subpaths"
   },
   {
      "serverUrl":"/cms/server",
      "serverAbsoluteUrl":"http://localhost:1337/cms/server",
      "adminUrl":"/cms/server/cms/dashboard",
      "adminAbsoluteUrl":"http://localhost:1337/cms/server/cms/server/cms/dashboard",
      "adminPath":"/cms/dashboard",
      "adminBaseName":"/cms/server/cms/server/cms/dashboard",
      "serverPath":"/cms/server",
      "note":"wtf"
   }
]
```

wladyslawczyzewski:
hey @Convly, I think the issues you get with the paths are due to mutation of config object and generating the paths on mutated config instead of using the config before mutation - see the change from the original PR to fix this issue https://github.com/strapi/strapi/pull/21507/files#diff-39132cb22fba5c04fba6f5fe28ee2a991f962c84eefc1dffdee2618cb7091ac1

Convly:
> hey @Convly, I think the issues you get with the paths are due to mutation of config object and generating the paths on mutated config instead of using the config before mutation - see the change from the original PR to fix this issue [#21507 (files)](https://github.com/strapi/strapi/pull/21507/files#diff-39132cb22fba5c04fba6f5fe28ee2a991f962c84eefc1dffdee2618cb7091ac1)
> 
> should I continue investigation with my original PR to reduce the number of changes or you prefer to go forward with these changes?

Hey, sorry for the late response, I was out of office for the past 2 weeks.

I'm gonna meet with Alexandre today to talk more about our options and where we want to go, so that we can move forward on this topic.

I'll keep you updated

packages/core/strapi/src/node/create-build-context.ts
```diff
@@ -79,15 +80,25 @@ const createBuildContext = async <TOptions extends BaseOptions>({
     });
 
   const serverUrl = strapiInstance.config.get<string>('server.url');
-  const adminPath = strapiInstance.config.get<string>('admin.path');
+  const serverAbsoluteUrl = strapiInstance.config.get<string>('server.absoluteUrl');
+  const adminUrl = strapiInstance.config.get<string>('admin.url');
+  const adminAbsoluteUrl = strapiInstance.config.get<string>('admin.absoluteUrl');
+
+  const sameOrigin = new URL(adminAbsoluteUrl).origin === new URL(serverAbsoluteUrl).origin;
```
Convly, line 87: Maybe consider adding a quick explanation of what we're testing.

e.g.

```typescript
// Check if both the server and admin URLs share the same origin
// Example: Comparing origins 'http://localhost:1337/admin' and 'http://localhost:1337/cms' will return true
```

packages/core/strapi/src/node/create-build-context.ts
```diff
@@ -79,15 +80,25 @@ const createBuildContext = async <TOptions extends BaseOptions>({
...
+
+  const adminPublicPath = new URL(adminAbsoluteUrl).pathname;
+
+  const adminPath = sameOrigin
+    ? strings.getCommonPath(serverUrl, adminUrl)
```
Convly, line 92: If both path share the same origin, shouldn't we take what's after the common path in the admin URL? Here it'll return the common origin and/or path (or am I missing something?)

packages/core/strapi/src/node/create-build-context.ts
```diff
@@ -79,15 +80,25 @@ const createBuildContext = async <TOptions extends BaseOptions>({
...
+    : new URL(adminUrl).pathname;
 
   const appDir = strapiInstance.dirs.app.root;
 
   await loadEnv(cwd);
 
   const env = getStrapiAdminEnvVars({
-    ADMIN_PATH: adminPath,
-    STRAPI_ADMIN_BACKEND_URL: serverUrl,
+    ADMIN_PATH: adminPublicPath,
```
Convly, line 100: Just to be sure, why are we using adminPublicPath instead of the computed adminPath here ?

Convly: (might be worth adding a comment to explain the reasoning anyway)

packages/core/strapi/src/node/vite/config.ts
```diff
@@ -85,6 +85,7 @@ const resolveDevelopmentConfig = async (ctx: BuildContext): Promise<InlineConfig
       },
     },
     server: {
+      cors: false,
```
Convly, line 88: Is it needed for this PR? Does serving from different paths causes cors issues?

alexandrebodin: It was causing some issues in the past, and I figure it doesn't make sense to have that conflict with our own cors middleware basically. I don't see where there would be an issue considering it's only in dev mode

packages/core/strapi/src/node/create-build-context.ts
alexandrebodin: it's going to be the url used in on the browser side so you need to have the full path including the potential proxied parts

Convly (approved):
LGTM, tested with the different usecases and works well. The PR makes the behavior ISO with v4

innerdvations:
Do we need to document these as new config options?

Convly:
> Do we need to document these as new config options?

We're making sure everything is ISO with v4, nothing new or undocumented should have been added :thinking:

Convly (changes requested):
Code LGTM

Can you fix the failing tasks (lint, ...) ?

Convly (approved):

packages/core/strapi/src/node/create-build-context.ts
ashisherc: this doesnt seem to work, if the admin config URL is set to `/` and the server URL is set to a backend URL. this still results in true and hence admin panel is built with relative path. essentially deployment on different server doesn't work.

It will be good to utilise the serverAdminPanel config option itself to determine the backend URL.

raphaelbadawi:
I may be missing something but upgrading Strapi to 5.5.1 (which should include this PR), putting back the rewrite rule which actually worked in Strapi 4 and the "url" conf in server.ts pointing to the absolute URL including the subpath, this which worked perfectly in Strapi 4, the exact same configuration and proxying, still doesn't work at all with Strapi 5.5.1 since assets are still loaded from the root domain.

Issue #20931, opened 2024-08-02
Deleting or renaming a component does not delete/update the relational linking data for the component


derrickmehaffy:
### Required System information

- Node.js version: node/v20.11.1 linux x64
- NPM version: yarn/1.22.22 npm/? 
- Strapi version: 4.25.6
- Database: SQLite
- Operating system: Linux Mint 21.3
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When a component is deleted or rename the related component data is not properly deleted or updated (updating may not be easily possible).

In certain cases or certain database types this can lead to an error similar to:

```
error: Metadata for "category.component" not found
```

I was not able to reproduce this error but I can see the remaining artifacts within the database

### Steps to reproduce the behavior

1. Create a content-type containing several components or dynamic zones with components
2. Create some test data on the content-type including some component data
3. Delete or rename the component (renaming requires doing it manually through the schema files)
4. Check the database component linking tables (using something like dbeaver)
5. See that the component data still exists

### Expected behavior

When a component is deleted all references in the database to that component should also be deleted. In the case of renaming we may not be able to do that automatically until we refactor the CTB (similar to renaming a normal field)

### Screenshots

Content Structure prior to modification:
![image](https://github.com/user-attachments/assets/2e994003-e6e1-4058-9d04-3ef0f2fa9322)

Example entry containing the data:
![image](https://github.com/user-attachments/assets/f50aeae1-3776-4c7b-84c8-8e783dec7404)

Database Prior to deleting the component:
![image](https://github.com/user-attachments/assets/a9eac3d3-76d8-48e1-8a02-815597529565)

Deleting a component and the data within the database:
![image](https://github.com/user-attachments/assets/0e4647c2-3105-4230-a9b6-9acf07b8bcbd)

### Code snippets

N/A

### Additional context

Related to TID7662

smoove:
Same issue, after deleting a component that was still assigned to a dynamic zone in a page.
Stack trace:

```
strapi | [2024-08-05 09:16:20.721] error: Metadata for "dev.test-delete" not found
strapi | Error: Metadata for "dev.test-delete" not found
strapi |     at Metadata.get (/opt/node_modules/@strapi/database/dist/index.js:2676:13)
strapi |     at morphToMany (/opt/node_modules/@strapi/database/dist/index.js:3704:22)
strapi |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #21928, opened 2024-10-21
feat: reordering for polymorphic relations in Content API

innerdvations:
### What does it do?

- adds support for set, connect, and disconnect through the Content API and document service for polymorphic relations
  - includes support for re-ordering

In a later PR we should refactor the massive relations test files into multiple smaller files

### Why is it needed?

They were buggy and incomplete on v4 and absent from v5; this adds support for them in v5, including reordering

### How to test it?

API tests have been added

Manually, ensure that submitting an API request to set, connect, and disconnect polymorphic relations works as expected, especially the case described in issue 20961

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/20961

Should fix? https://github.com/strapi/strapi/issues/20931

v4 PR: https://github.com/strapi/strapi/pull/21135

DX-1545

DX-1726


packages/core/core/src/services/document-service/transform/relations/extract/data-ids.ts
```diff
@@ -47,31 +47,34 @@ const addRelationDocId = curry(
...
 
       return mapRelation((relation) => {
         if (!relation || !relation.documentId) {
           return relation;
         }
 
-        addDocId(relation);
+        // Regular relations will always target the same target
+        // if its a polymorphic relation we need to get it from the data itself
+        const targetUid = attribute.target || (relation.__type as string);
```
Convly, line 65: Might be secondary, but instead of assuming there will always be a string `__type` property if `attribute.target` is not defined, could we make assertions and raise errors if it's not of the expected type?

Convly: This way we let TypeScript handles the type narrowing and we keep runtime safety

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
```diff
@@ -16,11 +16,14 @@ interface Options {
   allowMissingId?: boolean; // Whether to ignore missing ids and not throw any error
 }
 
+export const isPolymorphicRelation = (attribute: any): any =>
```
Convly, line 19: Don't we already have something similar in `@strapi/utils+content-types`?

Convly: (If not, maybe we could move this to the content types utils for the next person who need it?)

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
```diff
@@ -70,39 +73,47 @@ const transformDataIdsVisitor = (idMap: IdMap, data: Record<string, any>, source
...
       // Transform the relation documentId to entity id
       const newRelation = await mapRelation((relation) => {
         if (!relation || !relation.documentId) {
           return relation;
         }
 
-        const ids = getIds(relation);
+        // Find relational attributes, and return the document ids
+        // if its a polymorphic relation we need to get it from the data itself
+        const targetUid = (attribute.target || relation.__type) as UID.Schema;
```
Convly, line 87: Same as above

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
```diff
@@ -70,39 +73,47 @@ const transformDataIdsVisitor = (idMap: IdMap, data: Record<string, any>, source
...
+        const ids = getIds(targetUid, relation);
 
         // Handle positional arguments
         const position = { ...relation.position };
 
         if (position.before) {
           const beforeRelation = { ...relation, ...position, documentId: position.before };
-          position.before = getIds(beforeRelation).at(0);
+          // TODO: Reordering on polymorphic relations
```
Convly, line 95: Might be silly, but could you be a bit more precise on what's left to do regarding the reordering (or why isn't it working yet)?

packages/core/database/src/entity-manager/index.ts
```diff
@@ -554,35 +559,73 @@ export const createEntityManager = (db: Database): EntityManager => {
...
+          //     uid,
+          //     attributeName,
+          //     joinTable,
+          //     db,
+          //     transaction: trx,
+          //   });
+          // }
+
+          // set has priority over connect
+          // TODO: should we throw an error when both are passed in?
```
Convly, line 603: Couldn't that be considered a breaking change if we start refusing requests we were previously accepting? (both in the prog and rest API)

Convly (commented):
Just had some comments about the overall changes, didn't comment on the core logic.

packages/core/database/src/entity-manager/index.ts
innerdvations: yeah, more of a general question probably for v6, especially since I think right now combining the two is ambiguous and unpredictable between types of relations anyway. I think ideally it would accept all three at once with a fixed priority like set, then connect, then disconnect, that

packages/core/core/src/services/document-service/transform/relations/extract/data-ids.ts
innerdvations: Yeah, I'll clean this up over the next couple days with stuff like that

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
innerdvations: I think this was copy-pasted from there to get it working, can probably just delete it and use the shared util

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
innerdvations: I think this can be removed actually. @Marc-Roig do you know what would still be left here if anything? All the tests for re-ordering are implemented and passing I believe

Marc-Roig: this is still the case, we could make it work instead of keeping the todo

if connecting before / after a relation (polymorphic), we should also specify the __type uid of the specified before/after relation, atm we are using the uid of the relation being reordered.

e.g
```ts
{
  "connect": {
    "documentId": "X",
    "__type": "api::category.category",
    "position": {
      "before": {
        "documentId": "Y", 
        "type": "api::article.article" // This is what we are not using here
      }
    }
  }
}```

File: packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts (hunk 2 above)
Comment (Marc-Roig) [reply packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts:19]: absolutely! if we have a shared function for this we should reuse it, this was copy pasted from the database plugin, which does not export it

File: packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts (hunk 4 above)
Comment (innerdvations) [reply packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts:95]: Ahhh, so the reason the tests are passing is because the uids happen to be the same, that makes sense.

It seems like we might not need it at all though, the document id alone should be enough to find it in the relations table. We can take a look tomorrow

File: packages/core/utils/src/sanitize/visitors/remove-restricted-relations.ts [hunk 6]
```diff
@@ -63,15 +63,16 @@ export default (auth: unknown): Visitor =>
...
       } else {
         const newMorphValue = await handleMorphElements(elements);
 
-        // If the new value is empty, remove the relation completely
         if (newMorphValue.length === 0) {
-          remove(key);
+          // keep empty arrays
         } else {
           set(key, newMorphValue);
         }
```
Marc-Roig, line 78: should we just keep the set without the if else?

tests/api/core/upload/content-api/upload.test.api.js
```diff
@@ -438,25 +438,25 @@ describe('Upload plugin', () => {
...
+                connect: [imageIds[1]],
               },
             },
           }),
         });
 
         expect(connectRes.status).toBe(200);
 
-        // TODO: once Github issue 20961 is fixed, uncomment these lines to test results and update this test name to "works"
-        // expect(connectRes.body.data.relatedMedia).toHaveLength(1);
```
Marc-Roig, line 450: glad that we are finally closing that one 🥳

butcherZ (dismissed):
LGTM! Did QA, can confirm case described in #20961 have been fixed.

packages/core/database/src/entity-manager/relations-orderer.ts
```diff
@@ -51,8 +52,16 @@ const sortConnectArray = (connectArr: Link[], initialArr: Link[] = [], strictSor
...
 
-      // If the relation is already in the array, throw an error
-      if (mapper[relation.id]) {
+      // Check if the relation already exists in the mapper
+      const existingRelation = mapper[relation.id];
+
+      // Determine if the existing relation has no component or the same component
+      const hasNoComponent = existingRelation && !('__component' in existingRelation);
+      const hasSameComponent =
+        existingRelation && existingRelation.__component === relation.__component;
```
Marc-Roig, line 61: could you elaborate this note a bit more? it might be confusing if the dev reading this doesnt know about the __component attribute

Marc-Roig (dismissed):
LGTM! added a minor comment and did some light QA. Lets go 🚀

Marc-Roig (approved):

innerdvations:
I'm going to go ahead and merge this to fix 20961 for v5, because it's been open a long time and it's got tests and been tested for that.

For 20931, I will consider it complete once I add a test to verify it in a later PR.

Issue #21846, opened 2024-10-15
V5: Headers missing in OPTIONS request


HRRED:
### Required System information

- Node.js version: v20.15.0
- NPM version: 10.7.0
- Strapi version: 5.0.5
- Database: MySQL
- Operating system: MacOS Sonoma 14.5
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When sending an `OPTIONS` request to Strapi V5, none of the headers configured in the `strapi::cors` or the `strapi::security` middleware are set, unlike Strapi V4 as shown in the screenshots section. This causes CORS issues when sending requests with `withCredentials` set to `true`.

### Steps to reproduce the behavior

1. Create a Strapi V5 project
2. Configure the CORS middleware with origin (Example in the Code Snippets section)
3. Send an `OPTIONS` request to an endpoint with an `Origin` header configured and an `Access-Control-Request-Method` header set to `POST`.
4. None of the headers from the Strapi middlewares are set.

### Expected behavior

The response should include Access-Control headers as well as any headers configured by the `strapi::security` middleware. 

### Screenshots

The response in Strapi V4:
<img width="1297" alt="image" src="https://github.com/user-attachments/assets/f19bfacd-2827-4ae0-89d7-2680111ed42e">

The response in Strapi V5:
<img width="1296" alt="image" src="https://github.com/user-attachments/assets/643250aa-5cee-45a8-8482-357795e85122">

### Code snippets

The `strapi::cors` middleware configuration:
```Javascript
{
    name: 'strapi::cors',
    config: {
      headers: [
        'Content-Type',
        'Authorization',
        'Origin',
        'Accept',
      ],
      origin: [
        // Local API
        'http://localhost:1337',
        'http://localhost:1338',

        // Local Frontend
        'http://localhost:3000',
        'http://localhost:5173',
      ],
      keepHeaderOnError: true,
    }
  }
  ```

praharshaAdhikari:
Same issue here. It broke my auth that set refreshToken and accessToken as server-side-cookie.

derrickmehaffy:
Possible related cloud ticket: TID8590

derrickmehaffy:
> Reported a similar issue yesterday: #21836 Sending a request to auth/local sends back a cors error. I too tried to configure the cors middleware with no success

Looks quite similar but that one is touching the security middleware rather than the CORs one :thinking:

BayBreezy:
[...]

PR #21931, opened 2024-10-21
fix: load strapi before watch

alexandrebodin:
### What does it do?

- Load strapi before watch to avoid installing vite middleware too early

### Why is it needed?

Fix https://github.com/strapi/strapi/issues/21846

### How to test it?

- See that runing yarn develop works exactly the same with vite & webpack
- check the cors headers in the issue are work for both yarn develop & yarn start

### Related issue(s)/PR(s)


innerdvations:
This seems like the right thing to do in general, but it didn't exactly work for me to solve the issue.


My cors middleware:
```
 {
    name: 'strapi::cors',
    config: {
      headers: [
        'Origin',
        'Accept',
        'Content-Type',
        'Authorization',
        'X-Frame-Options',
        'X-Forwarded-For',
        'Access-Control-Allow-Credentials',
        'Access-Control-Allow-Origin',
      ],
      credentials: true,
      keepHeaderOnError: true,
      origin: [
        // Local API
        'http://localhost:1337',
        'http://localhost:1338',

        // Local Frontend
        'http://localhost:3000',
        'http://localhost:5173',
      ],
    },
  },
```

On `develop` I get access-control-origin as an empty line
```
"

"
```

With this PR I get access-control origin as `"*"`

derrickmehaffy:
Localhost won't set the origin header so it returning * on localhost is normal. The origin header only is sent on actual domains or cross site requests

alexandrebodin:
You have to manually set the origin to test this. In Strapi this won't do anything because it's the same domain

example 
Screenshot 2024-10-22 at 06 29 20

innerdvations:
Ahhh, ok, I thought it was always set. I will try again with that in mind later!

PR #21934, opened 2024-10-22
fix: Remove Private Attributes from GraphQL Filters and Inputs

### LINKED ISSUE #16325: Private fields still show in inputs and filters inputs

- Node.js version:v18.12.1
- NPM version: 8.19.2
- Strapi version: 4.8.2
- Database: Postgres
- Operating system: MacOS

### Describe the bug

Following up on #16169 and #16172

Since 4.8.1 private fields can't be filtered by anymore, but they still show up in GraphQL's Inputs and FiltersInputs.

1. Add a private field
2. See that they still show in GraphQL's schema (not on the entity, but on the input and filters input)

### Expected behavior

Private fields shouldn't show up on the schema at all.

Convly:
### What does it do?

Prevent attributes tagged as private to appear in the GraphQL Input and Filters schema definitions.

TODO:

- [x] u&p registration for graphql should manually add password as an exception to prevent a breaking change and to mirror content API register behaviour
- [x] add test that private fields aren't visible in schema anymore

### Why is it needed?

Private attributes shouldn't be leaked through the GraphQL schemas, even if they're filtered out anyway.

Also this could cause issues due to required private fields in the inputs.

### How to test it?

- Create a content-type with public and private attributes
- Head to http://localhost:1337/graphql
- Navigate to the schemas inputs and filters types and make sure there is no mention of the private attribute

### Related issue(s)/PR(s)

fix #16325


Convly:
Note: Two API tests are now failing (Users & Permissions create and update user mutations)

This is most likely because `password` is now stripped out of the valid attributes in the User schema.

Is it something we want to address in the tests or the schema type? Do we want to allow creating/updating users (including their passwords)

Convly:
Experimental version for testing: `0.0.0-experimental.bf539fcb8054f8f95f3fd89fb24321df84d0a9f8`

innerdvations:
> Note: Two API tests are now failing (Users & Permissions create and update user mutations)
> 
> This is most likely because `password` is now stripped out of the valid attributes in the User schema.
> 
> Is it something we want to address in the tests or the schema type? Do we want to allow creating/updating users (including their passwords)

It should mirror what the content API accepts, which I think does accept password -- although I haven't verified that

innerdvations:
Looks like some tests are failing :(

alexandrebodin (commented):
code lgtm but did not do any QA. I'll let you handle it 👍

PR #21938, opened 2024-10-22
fix: releases page e2e

Marc-Roig:
### What does it do?

Fix `releases-page.spec.ts` e2e tests


tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -136,10 +134,9 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
     });
 
     await test.step('releases should be updated in the release column of list view', async () => {
-      const releaseColumn = page.getByRole('button', { name: '2 releases' });
+      const releaseColumn = page.getByRole('button', { name: '1 release' });
       await releaseColumn.first().click();
       await expect(page.getByText('The Diamond Dogs')).toBeVisible();
-      await expect(page.getByText('Trent Crimm: The Independent')).toBeVisible();
     });
```
markkaylor, line 140: Just curious why we only check for 1 now instead of 2? I can't remember but I think before the data set seeded the database with an initial release...

markkaylor (dismissed):
Looks good to me if they all pass here in the CI

tests/e2e/tests/content-releases/releases-page.spec.ts
Marc-Roig: Indeed there are two releases in the initial data, but the entry is only added into one of them in this e2e test, so I assume it was added into two of them when this test was created

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -43,57 +43,65 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
...
-  test.fixme(
-    'A user should be able to create a release with scheduling info and view their pending and done releases',
-    async ({ page }) => {
-      // Navigate to the releases page
-      await page.getByRole('link', { name: 'Releases' }).click();
+  test('A user should be able to create a release with scheduling info and view their pending and done releases', async ({
+    page,
+  }) => {
+    // Navigate to the releases page
+    await page.getByRole('link', { name: 'Releases' }).click();
```
remidej, line 50: I recommend using our clickAndWait util, it makes sure to wait that the page is loaded before proceeding, which helps with flakiness

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -43,57 +43,65 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
...
-      await page.getByLabel(formattedDate).click();
+    await page.getByRole('combobox', { name: /^time$/i }).click();
 
-      await page
-        .getByRole('combobox', {
-          name: /^time$/i,
-        })
-        .click();
+    const optionIsVisible = await page.getByRole('option', { name: '08:00' }).isVisible();
+    console.log('08:00 option visible:', optionIsVisible);
```
remidej, line 81: 🧹

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -43,57 +43,65 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
...
 
-      await page.getByRole('option', { name: '14:00' }).click();
-      await page.getByRole('button', { name: 'Continue' }).click();
-      // Wait for client side redirect to created release
-      await page.waitForURL('/admin/plugins/content-releases/*');
-      await expect(page.getByRole('heading', { name: newReleaseName })).toBeVisible();
+    const buttonVisible = await page.getByRole('button', { name: 'Continue' }).isVisible();
+    console.log('Continue button visible:', buttonVisible);
```
remidej, line 84: 🧹

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -43,57 +43,65 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
...
-      await page.getByRole('link', { name: 'Releases' }).click();
-      await expect(page.getByRole('link', { name: `${newReleaseName}` })).toBeVisible();
-    }
-  );
+    // Log URL changes
+    page.on('framenavigated', async (frame) => {
+      if (frame === page.mainFrame()) {
+        console.log('Navigation occurred to:', frame.url());
+      }
+    });
```
remidej, line 91: looks like this can be removed as you're just logging

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -43,57 +43,65 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
...
+    await page.getByRole('option', { name: '08:00' }).click();
+    await page.getByRole('button', { name: 'Continue' }).click();
+
+    // Wait for client side redirect to created release
+    await page.waitForURL('/admin/plugins/content-releases/*');
+    await expect(page.getByRole('heading', { name: newReleaseName })).toBeVisible();
+
+    // Navigate back to the release page to see the newly created release
+    await page.getByRole('link', { name: 'Releases' }).click();
```
remidej, line 100: Looks like a good spot for clickAndWait (I know await toBeVisible below should make this redundant, but I have flakiness ptsd 🙃)

tests/e2e/tests/content-releases/releases-page.spec.ts
```diff
@@ -111,9 +119,8 @@ describeOnCondition(edition === 'EE')('Releases page', () => {
 
       // Navigate to the content manager
       await page.getByRole('link', { name: 'Open the Content Manager' }).click();
-      await expect(page).toHaveTitle('Content Manager');
+      await page.waitForURL('/admin/content-manager/collection-types/*');
```
remidej, line 122: I think we can clickAndWait again, we just need to make sure we arrived on the page, we don't need to explicitly check that the CM URL is the right one

Marc-Roig:
@remidej there was no default timezone set for e2e tests, and the timezone field for a release was epmty, making it fail on creation 😄

remidej (approved):
Great fix 🚀

playwright.base.config.js
```diff
@@ -62,6 +62,9 @@ const createConfig = ({ port, testDir, appDir }) => ({
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: `http://127.0.0.1:${port}`,
 
+    /** Set timezone for consistency across any machine*/
+    timezoneId: 'Europe/Paris',
```
markkaylor, line 66: 🚀

markkaylor (approved):

PR #20609, opened 2024-06-24
fix: ui

alexandrebodin:
### What does it do?

Fixes the following issues

- log user out on unauthorized requests
- Broken icons in users & permissions
- missing confirm modal on user delete
- app logo customization broken
- RBAC permissions matrix broken
- Infinite loading because of unecessary i18N initializer
- Rbac conditions modal broken
- Fix menu not showing EE links before refresh

PR #21947, opened 2024-10-23
fix: api state not cleared on logout

### LINKED ISSUE #20344: User cache is not cleared when logging out and logging in as a different user

- Node.js version: v20
- Yarn version: v1
- Strapi version: v4.24.2
- Database: SQLite

### Describe the bug

When logging out and logging in as a different user with different permissions, the new login will see some things related to the previous users permissions. This is corrected by refreshing the page (and doesn't leak any sensitive information) but any data fetching caches should be cleared when logging out

Bug report created simply to mirror DX-1416: https://strapi-inc.atlassian.net/browse/DX-1416

### Expected behavior

Data caches should be cleared

### Screenshots

N/A

### Code snippets

N/A

### Additional context

Related to TID: 6795

remidej:
### What does it do?

Fixes an issue where the app's local cache wasn't cleared when you logged out.

### Why is it needed?

Prevent one user's data from leaking to the next user to log in via the same browser

### How to test it?

Have 2 users users:
- one is a super admin with full permissions
- one is an "author" without permissions to read a specific field

You should be able to do this in the same tab without refreshing:
- login as the author user
- open the edit/create view where you removed the permissions for a field. You shouldn't see that field.
- logout
- login as super admin
- go to the same edit/create view. You should see the field

### Related issue(s)/PR(s)

fixes #20344

### DIFF
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -7,6 +7,7 @@ import { Login } from '../../../shared/contracts/authentication';
 import { createContext } from '../components/Context';
 import { useTypedDispatch } from '../core/store/hooks';
 import { setLocale } from '../reducer';
+import { adminApi } from '../services/api';
 import {
   useGetMeQuery,
   useLoginMutation,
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/features/Auth.tsx
```diff
@@ -154,10 +155,11 @@ const AuthProvider = ({ children }: AuthProviderProps) => {
   );
 
   const logout = React.useCallback(async () => {
+    dispatch(adminApi.util.resetApiState());
     await logoutMutation();
     clearStorage();
     push('/auth/login');
-  }, [clearStorage, logoutMutation, push]);
+  }, [clearStorage, logoutMutation, push, dispatch]);
```
remidej, line 162: fun fact, if resetApiState isn't the first thing in that function, it triggers a window reload on the login page. Not sure why

jhoward1994 (approved):
Works for me 👍🏻

I think the v5 equivalent was added here https://github.com/strapi/strapi/pull/20609

PR #21950, opened 2024-10-23
fix: Documentation plugin makes git changes in production

dkochetkov:
### What does it do?

Added check <code>process.env.NODE_ENV !== "production"</code> before generating documentation.

### Why is it needed?

Generating documentation leads to changes in git every time the application is launched using <code>NODE_ENV=production</code>.

We don't need to change documentation in production because we can't change API schemas in production, we can change only content. We can and should use generated and committed in dev-mode documentation.

### How to test it?

- Create git repo
- Create Strapi project
- Add and configure Documentation plugin
- Run project in <code>dev</code> mode — documentation plugin works perfectly
- Stop project
- Commit changes
- Go to production server
- Fetch project from origin repo
- Set <code>NODE_ENV=production</code>
- Run project with <code>npm start</code>
- Check <code>git status</code>

git


Bassel17 (approved):

Bassel17:
Looks good! Thank you for the contribution

molund:
Note: you can override this change in `src/index.js` to get the old behaviour back.
```
  async bootstrap({ strapi }) {
    if (process.env.NODE_ENV === "production") {
      const svc = strapi.plugin("documentation")?.service("documentation");
      if (svc) {
        await svc.generateFullDoc();
      }
    }
  },
  ```
Our team uses a `.gitignore` entry to get around the issue described in the PR description.

Issue #20461, opened 2024-06-07
Translating content literals for admin does not work as before 


SalahAdDin:
### Required System information

- Environment: development
- OS: Linux-x64
- Strapi Version: 5.0.0-beta.5
- Node/Yarn Version: yarn/4.1.1 npm/? node/v20.12.2 linux x64
- Edition: Community
- Database: SQLite
- Operating system: 6.6.32-1-MANJARO
- Typescript

```
Launched In: 382 ms
Environment: development
OS: linux-x64
Strapi Version: 5.0.0-beta.5
Node/Yarn Version: yarn/4.1.1 npm/? node/v20.12.2 linux x64
Edition: Community
Database: SQLite
[2024-06-07 00:15:05.490] info: Shutting down Strapi
[2024-06-07 00:15:05.491] info: Strapi has been shut down
```

### Describe the bug
We are working on localizing all the content types we create for our project, we followed the existing [documentation](https://docs-next.strapi.io/dev-docs/admin-panel-customization#locales), but it seems to be deprecated.

When checking the literals on the admin panel, only the content type names are translated, but the rest of the literals are in English yet.

This does not happen in Strapi 4.

### Steps to reproduce the behavior

1. Add the localization files to the folder.
2. Write the function for loading the translation files on the admin.
3. Load the translations on the admin app file.
4. Run the server.
5. Go to the admin panel.
6. Check the content manager.
7. Click on a content type.
8. Check the missing translations.

### Expected behavior

All the translation literals should be shown properly as it happens in Strapi 4.

### Screenshots

Strapi 5:
![image](https://github.com/strapi/strapi/assets/5034215/8292c87c-2989-4d80-93dc-758a0a13a7e9)

![image](https://github.com/strapi/strapi/assets/5034215/5aa9fdef-5cf5-4f34-8781-d9f4623084a6)

Strapi 4:
![image](https://github.com/strapi/strapi/assets/5034215/2c1c9c7d-bdfa-468f-8674-49b0b86ae8a5)

![image](https://github.com/strapi/strapi/assets/5034215/c6497588-4992-4524-980b-55daf37ce42e)

### Code snippets

You can check the files in [this commit](https://github.com/SalahAdDin/strapi-template/commit/c6a4777b0848e939f5fc2998a089d32708381e38).

### Additional context

It also has a lot of reports of missing translation literals:
```
[...]

PR #21951, opened 2024-10-23
fix(content-manager): translate content-type fields in List and Edit …

markkaylor:
### What does it do?

- Adds ability to translate a content-types' attributes in both the list view (table headers) and edit view (form labels)
- Adds ability to translate components' attributes wherever the component is used

The translation ids are computed with:

`content-manager.{content-types | components}.{contentTypeUid | componentUid}.{attributeName}`

### Why is it needed?

- It was possible in Strapi 4

### How to test it?

In `examples/getstarted/src/admin/app.jsx` provide a config like the one below.
Then change the language of your app in your profile to one of the available locales

```js
const config = {
  locales: ['it', 'es', 'en', 'en-GB', 'fr'],
  translations: {
    fr: {
      // Keys scoped to a content-type, tranlsates the attribute for that content-type
      'content-manager.content-types.api::kitchensink.kitchensink.short_text': 'translated short text',
      'content-manager.content-types.api::kitchensink.kitchensink.single_compo':
        'translated single compo',
      // Keys scoped to a component, translates the component attribute for any content-type that uses it
      'content-manager.components.basic.relation.categories': 'translated dynamic zone compo attribute',
      'content-manager.components.default.temp.name': 'translated default.temp name',
      'content-manager.components.basic.simple.name': 'translated basic.simple name',
      'content-manager.components.basic.simple.nested': 'translated nested compo',
    },
  },
};
```
- Providing a translation for a root level attribute (text, component, dynamic zone, integer, etc...) should only translate within the scope of its content-type
- Providing a translation for a component (non-repeatable, repeatable, and dynamic zone) should translate within the scope of the component and therefore in _any_ content-type that is using it.

### Related issue(s)/PR(s)

resolves: https://github.com/strapi/strapi/issues/20461


packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -3,20 +3,36 @@ import { Box, Flex, Grid } from '@strapi/design-system';
...
+import { useIntl } from 'react-intl';
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {}
 
 const FormLayout = ({ layout }: FormLayoutProps) => {
+  const { formatMessage } = useIntl();
+
+  const getTranslatedLabel = (label: string) => {
+    return formatMessage({
+      id: `content-manager.containers.EditPage.form.${label}`,
```
markkaylor, line 15: ❓  I'm wondering if we should scope the translations to a content-type and not worry about the page we are on? `content-manager.container.{content-type-name}.{attribute-name}`.

remidej: That sounds nice, but I the same time I think it means breaking backwards compatibility? If so I think we can keep the current ids

remidej (approved):
haven't done QA but LGTM :)

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
remidej: if we need a BC anyway then let's make the best api then 🚀 I agree the label should be the same between pages

markkaylor:
@remidej I pushed to use a single key for both pages: `content-manager.{contentTypeUid}.{attributeName}`

```
const config = {
  locales: ['it', 'es', 'en', 'en-GB', 'fr'],
  translations: {
    fr: {
      'content-manager.api::kitchensink.kitchensink.short_text': 'texte_court'
    }
  }
};
```

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -1,22 +1,40 @@
...
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {}
 
 const FormLayout = ({ layout }: FormLayoutProps) => {
+  const { formatMessage } = useIntl();
+  const { model } = useDoc();
+
+  const getTranslatedLabel = (label: string) => {
+    return formatMessage({
+      id: `content-manager.${model}.${label}`,
```
remidej, line 17: This is much better but I think we could still have collisions 🤔 how about `content-manager.[something].${model}.${label}`, with something like field in the middle?

markkaylor: `content-manager.content-types.${model}.${label}`? wdyt? All ears for another suggestion, before it was containers but I don't like that

remidej: sounds good!

markkaylor:
@remidej updated

remidej (approved):
Works nicely! Just a heads up you're missing `.content-types` in the PR description key (I think Marion uses them for QA around release time, that's why it's good to keep it updated)

Also I wonder if this should handle components too? If so, how? And should it be another PR?

markkaylor:
> Works nicely! Just a heads up you're missing `.content-types` in the PR description key (I think Marion uses them for QA around release time, that's why it's good to keep it updated)
> 


My bad thanks for the heads up. I updated.

> Also I wonder if this should handle components too? If so, how? And should it be another PR?

I'll take a look at components now to see if it makes sense to handle it here.

kauedelmonte:
Please release this 🙏

markkaylor:
@remidej I don't think there is anything to be done here for components actually. Maybe something in the CTB? but it looks like there isn't any translation over there and the attribute name is used directly. I think I'll go ahead and merge this.

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -1,22 +1,40 @@
...
   return (
     <Flex direction="column" alignItems="stretch" gap={6}>
       {layout.map((panel, index) => {
         if (panel.some((row) => row.some((field) => field.type === 'dynamiczone'))) {
           const [row] = panel;
           const [field] = row;
+
+          const fieldWithTranslatedLabel = {
+            ...field,
+            label: getTranslatedLabel(field.label),
```
markkaylor, line 31: This is wrong I should be passing field.name

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -1,22 +1,40 @@
...
 
 import { InputRenderer } from './InputRenderer';
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {}
 
 const FormLayout = ({ layout }: FormLayoutProps) => {
+  const { formatMessage } = useIntl();
+  const { model } = useDoc();
+
+  const getTranslatedLabel = (label: string) => {
```
markkaylor, line 15: the argument should be name, not label

remidej (approved):
Tested and works well!

I'm noticing the labels don't apply to history, but I think that makes sense, since we don't track the history of those labels over time

PR #21963, opened 2024-10-24
release: main to develop

Marc-Roig:


Convly (approved):

innerdvations (approved):
I didn't realize we could safely do this from github 😆

Marc-Roig:
@Convly @innerdvations do you know if we did this merge main->develop before? there seems to be some diffs that are not from the release (and prettier is failing too)

innerdvations:
There seems to be a `prettier` issue though 🤔 we might want to fix that before merging to develop

Marc-Roig:
I plan to fix those issues during the day, but I wonder if I messed up something during the merge

innerdvations:
As far as I know we've never merged it this way, only manually. I'm a bit worried about that "Update Branch" button on github getting pushed.

PR #21965, opened 2024-10-24
enhancement: create dedicated preview page

remidej:
### What does it do?

Sets up the admin routes for Preview. The route the edit view URL + `/preview`. So I two routes are technically created, one for single types and one for collection types.

Nothing is currently displayed on the page, I'm breaking it into several PRs.

I set up the page very similarly to what we did in history as it worked pretty well. The page component does all the data fetching that the components will need. This way we can have centralized loading and error states. Then the data is made available to components via a context hook.

Most e2e tests are deleted for now, we'll restore/update them as we introduce the bits of UI in the next PRs.

### Why is it needed?

It's what we need to then display the header and iframe.

### How to test it?

- open the edit view for a saved content type
- the "copy preview url" button should not appear
- click on open preview, you should be redirect to the `/preview` route and see "preview will go here"
- the query params should be kept on the preview route. Try that by clicking on preview from different locales and from both draft and published tabs


packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+  if (
+    !previewUrlResponse.data?.data?.url ||
+    previewUrlResponse.error ||
+    documentLayoutResponse.error ||
+    !documentResponse.document ||
+    !documentResponse.meta ||
+    !documentResponse.schema
+  ) {
+    return <Page.NoData />;
+  }
```
Marc-Roig, line 104: should we show an error in case there is an issue fetching the previewUrl?

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  if (!model) {
+    throw new Error('Could not find model in url params');
+  }
+
+  // Only collection types must have a documentId
+  if (collectionType === COLLECTION_TYPES && !documentId) {
+    throw new Error('Could not find documentId in url params');
+  }
```
markkaylor, line 65: Isn't impossible to access this route without these values? I know we want to ensure we have a value, but I'm not sure these errors ever throw since if you don't have these params then I guess you are on another route...right?

markkaylor (dismissed):

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: yes they should never throw, it's just for type safety because everything from useParams is `string | undefined`. I copied the approach from `useDoc`, this way I don't have to use `!`

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: oh yes I think I got confused here, I wanted Page.NoData only is no preview URL is provided, and an error screen for the rest. I'll update

Marc-Roig (approved):
nicee 🚀

markkaylor (approved):

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: Also with throw you can keep using hooks below, which is not the case when returning something as it breaks the rules of hooks

PR #21970, opened 2024-10-24
enhancement: add back button fallback support

remidej:
### What does it do?

Adds a fallback prop to the BackButton from `@strapi/admin/strapi-admin`.

This has been approved by Yannis [on slack (plus video for more context)](https://strapihq.slack.com/archives/C051PLC9CF7/p1729774440081529)

### Why is it needed?

I want to avoid the case where the back button is disabled, as that UX isn't great imo. It happens when you reload your browser window, or if you land on a page via a bookmark or a link someone sent you.

In a lot of cases, the page . For example, CM list view > CM edit view. It doesn't work everywhere, but this PR adds the API so we can tell where the back button should lead when no history is found. In a lot of cases it looks like this with react router:

```jsx
<BackButton fallback=".." />
```

Note that when relying on the fallback prop, it just adds to the history stack, it doesn't emulate the browser's back button. IMO this is an okay tradeoff.

### How to test it?

In each of these pages, reload the page before clicking on the back button. Check that you are redirect to the page that you expected (the "parent" page).

- cm > edit view for a collection type
- cm > edit view for a single type (should be disabled)
- cm > list view > configure the view
- settings > admin roles > create
- settings > admin roles > edit
- settings > admin users > edit
- settings > webhooks > create/edit
- settings > review workflows > workflow
- settings > users & permissions > role

- releases > release details page


packages/core/admin/admin/src/features/BackButton.tsx
```diff
@@ -188,43 +188,54 @@ const reducer = (state: HistoryState, action: HistoryActions) =>
...
-    </Link>
-  );
-});
+    return (
+      <Link
+        ref={ref}
+        tag={NavLink}
+        to={history.at(-1) ?? fallback}
+        onClick={handleClick}
+        disabled={disabled || (!canGoBack && !hasFallback)}
```
simotae14, line 227: Maybe to avoid repetition you can define a variable for the disable value, but it is up to you

simotae14 (dismissed):
The code looks good, I need to test it

simotae14:
I am testing the solution:
- cm > edit view > configure the view

https://github.com/user-attachments/assets/4cf7c530-1ba6-4d04-b956-190bc2a09b31

in the video I can see the back button go to the cm page instead of the edit view, is it the correct "parent" page?

simotae14:
Same for the case:
 - cm > list view > configure the view

https://github.com/user-attachments/assets/74c395f8-041a-4bb7-be81-7abc47940d1e

 
 we land on the cm instead of the list view, is it correct?

remidej:
> cm > edit view > configure the view

@simotae14 for that one we can't guide to an edit view because we don't know which entry the user came from. Also you can get to this page from the CTB injection zone button. So since there's no clear parent I think I'll just remove the fallback there. Thanks for highlighting this

> cm > list view > configure the view

yeah that's just a bug, nice catch!

packages/core/admin/admin/src/features/BackButton.tsx
```diff
@@ -188,43 +188,54 @@ const reducer = (state: HistoryState, action: HistoryActions) =>
...
  * BackButton
  * -----------------------------------------------------------------------------------------------*/
-interface BackButtonProps extends Pick<LinkProps, 'disabled'> {}
+interface BackButtonProps extends Pick<LinkProps, 'disabled'> {
+  fallback?: To;
+}
 
 /**
  * @beta
  * @description The universal back button for the Strapi application. This uses the internal history
```
jhoward1994, line 197: This looks good overall. Maybe a comment covering what the fallback prop is for would be useful?

simotae14 (approved):
Code looks great 💯 and also I tested the solution and it works, thanks a lot Rémi

jhoward1994 (approved):

Issue #21915, opened 2024-10-18
Strapi 5: Apollo Sandbox inaccessible


maxep:
### Required System information

- Environment: development
- OS: linux-x64
- Strapi Version: 5.1.0
- Node/Yarn Version: yarn/4.5.0 npm/? node/v20.10.0 linux x64
- Edition: Community
- Database: sqlite

### Describe the bug

Apollo Sandbox is inaccessible when project is **not** using the Users & Permissions plugin.
It was previously [fixed in v4](https://github.com/strapi/strapi/issues/12289) and previous Apollo version. So it's most likely a regression.

### Steps to reproduce the behavior

1. Create vanilla project
2. Remove the `@strapi/plugin-users-permissions` from dependencies
3. Add `@strapi/plugin-graphql` and its plugin configuration
4. Start project and open the graphql endpoint in browser
5. See error:

```json
{
  "data": null,
  "error": {
    "status": 401,
    "name": "UnauthorizedError",
    "message": "Missing or invalid credentials",
    "details": {}
  }
}
```

### Expected behavior

Apollo Sandbox should not require users-permissions

PR #21977, opened 2024-10-25
fix graphql playground

kaioduarte:
### What does it do?

Makes the GraphQL playground accessible on V5 by postponing the auth middleware check. With the current implementation, the auth middleware is called before we can check the route config (`auth: false`), causing it to throw 401.

### Why is it needed?

GraphQL playground is not accessible at all without the user-permissions plugin.

### How to test it?

1. Create vanilla project
2. Remove the @strapi/plugin-users-permissions from dependencies
3. Add @strapi/plugin-graphql and its plugin configuration
4. Start project and open the graphql endpoint in browser, it should display the Apollo playground;

### Sanity checks

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/1ba31c7b-9a51-40f2-bcb6-3b2bad346fde"> | <video src="https://github.com/user-attachments/assets/2bbccecc-a48d-4284-b0c9-c5427e8ce7cb"> |

### Related issue(s)/PR(s)

Closes https://github.com/strapi/strapi/issues/21915

### DIFF
--- a/packages/plugins/graphql/package.json
+++ b/packages/plugins/graphql/package.json
@@ -49,7 +49,6 @@
   },
   "dependencies": {
     "@apollo/server": "4.11.0",
-    "@as-integrations/koa": "1.1.1",
     "@graphql-tools/schema": "10.0.3",
     "@graphql-tools/utils": "^10.1.3",
     "@koa/cors": "5.0.0",
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -4,7 +4,6 @@ import {
   ApolloServerPluginLandingPageLocalDefault,
   ApolloServerPluginLandingPageProductionDefault,
 } from '@apollo/server/plugin/landingPage/default';
-import { koaMiddleware } from '@as-integrations/koa';
 import depthLimit from 'graphql-depth-limit';
 import bodyParser from 'koa-bodyparser';
 import cors from '@koa/cors';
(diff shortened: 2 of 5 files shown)


kaioduarte:
fixed the commit message :)

Bassel17:
Hello thanks for the contribution, this does indeed fix your issue when you remove users-permissions plugin, but it does break the authentication with users-permissions, so if you can take a look an revise the solution.
you should be able to test this by running the api tests

kaioduarte:
These changes break the other tests because this Apollo Server middleware doesn't call the `next()` function and the middleware that authenticates is never called.

https://github.com/strapi/strapi/blob/ca8b9b60c71455ebc2e2ddbdca85a95ac434eea2/packages/plugins/graphql/server/src/bootstrap.ts#L125-L133

Source code
https://github.com/apollo-server-integrations/apollo-server-integration-koa/blob/2a6a9c55b23efa297c9ea42915ec3a1f5146b0d1/src/index.ts#L72

I patched this middleware locally and everything seems to work well without touching the existing code. I can try to make the change in the AS repo, but I'm afraid it'll take a lot to merge as the repo seems inactive (last release >1y ago).

What do you think? @Bassel17

kaioduarte:
I opened the PR in the AS repo, let's see how it goes.

https://github.com/apollo-server-integrations/apollo-server-integration-koa/pull/232

Bassel17:
Hey thanks for the contribution and work on this will wait until your PR in the AS repo is merged,
but for now we went with a one line solution in another PR

PR #21980, opened 2024-10-25
fix(e2e): enable e2e bulk action tests in listview.spec.ts

markkaylor:
### What does it do?

Enables the bulk action tests in listview.spec.ts

### Why is it needed?

To test the feature

### How to test it?

The tests should pass here in the CI


remidej (approved):
I guess that works 💪

PR #21983, opened 2024-10-25
enhancement: add preview page header

remidej:
### What does it do?

Adds a header to the preview page. It displays the main field, the document's status, and a button to copy the preview page link (within Strapi Admin, not the direct preview URL).

I found that the CM's document status wasn't using translations, so I changed it. The modified key is new but for draft and publish it should be compatible with translations from v4

The draft/publish toggle will be in another ticket/PR

### How to test it?

Open preview for content types where it's enabled. Check the mainfield and document status, try changing them too. And copy the page URL


tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -24,7 +24,16 @@ describeOnCondition(edition === 'EE')('Preview', () => {
 
     // Check that preview opens in its own page
     await clickAndWait(page, page.getByRole('link', { name: /open preview/i }));
-    await expect(page.getByText('Preview will go here!')).toBeVisible();
+    await expect(page.getByText(/^draft$/i)).toBeVisible();
```
remidej, line 27:
```suggestion
    await expect(page.getByText(/^Draft$/)).toBeVisible();
```

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -0,0 +1,208 @@
...
+        border: 'success200',
+        text: 'success700',
+        message: {
+          id: 'content-manager.containers.List.published',
+          defaultMessage: 'Published',
+        },
+      };
+  }
+};
+
```
markkaylor, line 112: How come we have to do this here specifically for preview? Isn't this similar to what we are doing in history and maybe elsewhere in the app?

https://github.com/strapi/strapi/blob/11331d7572f8d2da397e6268b4f2d82f836e0998/packages/core/content-manager/admin/src/history/components/VersionsList.tsx#L45

Is there a difference? Or would it make sense to share this function instead of duplicating it?

remidej: Yes that's the code I copied. In the Content Manager list and edit view it uses [a DocumentStatus component](https://github.com/strapi/strapi/blob/2f7b9ae55adbbf4416af874fb13b222b851c4432/packages/core/content-manager/admin/src/pages/EditView/components/DocumentStatus.tsx#L19). It relies on [Status from the design system](https://design-system.strapi.io/?path=/docs/design-system-components-status--docs) to apply the right text/border/bg colors.

I'd like reuse Document status across the app, but Status from the DS doesn't have a variant small enough to match our designs. So we need to open a PR in the DS, ship it, bump it in Strapi, update DocumentStatus, then we can delete the extraneous code both here and in history.

In the meantime we're left with this 🤷

markkaylor (approved):

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
remidej: I'm working on opening the DS PR now so we don't forget

PR #21984, opened 2024-10-25
fix: review workflow persist data

Marc-Roig:
### What does it do?

Review workflows adds two additional attributes to content types, strapi_assignee and strapi_stage . Those are the values displayed in the content manager edit view.

As RW is an EE feature, it was decided for those attributes to not exist on the Comunity Edition. But at the same time if a user filled those values in EE and they downgrade, we want to keep the data there.

In v4 we were using a persist_table logic, to persist those join tables even if the attribute didn't exist. At some point we discussed doing otherwise, but eventually this was forgotten.

This PR reintroduces the persisted table logic we had in in v4.

### DIFF
--- a/packages/core/admin/ee/server/src/services/persist-tables.ts
+++ b/packages/core/admin/ee/server/src/services/persist-tables.ts
@@ -1,7 +1,7 @@
 import type { Core } from '@strapi/types';
 import { differenceWith, isEqual } from 'lodash/fp';
 
-interface PersistedTable {
+export interface PersistedTable {
   name: string;
   dependsOn?: Array<{ name: string }>;
 }
--- a/packages/core/review-workflows/server/src/register.ts
+++ b/packages/core/review-workflows/server/src/register.ts
@@ -1,6 +1,6 @@
 import { defaultsDeep, filter, pipe, map } from 'lodash/fp';
 
-import type { Core } from '@strapi/types';
+import type { Core, UID } from '@strapi/types';
 
 import { getService, getAdminService } from './utils';
 import migrateStageAttribute from './migrations/shorten-stage-attribute';
@@ -63,8 +63,8 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
 function persistRWOnDowngrade({ strapi }: { strapi: Core.Strapi }) {
   const { removePersistedTablesWithSuffix, persistTables } = getAdminService('persist-tables');
 
-  return async ({ contentTypes }: any) => {
-    const getStageTableToPersist = (contentTypeUID: any) => {
+  return async ({ contentTypes }: { contentTypes: Record<UID.ContentType, any> }) => {
+    const getStageTableToPersist = (contentTypeUID: UID.ContentType) => {
       // Persist the stage join table
       const { attributes, tableName } = strapi.db.metadata.get(contentTypeUID) as any;
       const joinTableName = attributes[ENTITY_STAGE_ATTRIBUTE].joinTable.name;
@@ -74,7 +74,7 @@ function persistRWOnDowngrade({ strapi }: { strapi: Core.Strapi }) {
       };
     };
 
-    const getAssigneeTableToPersist = (contentTypeUID: any) => {
+    const getAssigneeTableToPersist = (contentTypeUID: UID.ContentType) => {
       // Persist the assignee join table
       const { attributes, tableName } = strapi.db.metadata.get(contentTypeUID) as any;
       const joinTableName = attributes[ENTITY_ASSIGNEE_ATTRIBUTE].joinTable.name;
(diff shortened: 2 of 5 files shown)


packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
+/**
+ * Persist the stage & assignee attributes so they are not removed when downgrading to CE.
+ *
+ * TODO: V6 - Instead of persisting the join tables, always create the stage & assignee attributes, even in CE mode
+ *            It was decided in V4 & V5 to not expose them in CE (as they pollute the CTs) but it's not worth given the complexity this needs
+ */
+function persistRWOnDowngrade({ strapi }: { strapi: Core.Strapi }) {
+  const { removePersistedTablesWithSuffix, persistTables } = getAdminService('persist-tables');
+
+  return async ({ contentTypes }: any) => {
```
markkaylor, line 66: We don't have any types for contentTypes?

packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
+    const getStageTableToPersist = (contentTypeUID: any) => {
```
markkaylor, line 67: Don't we know that the uid is a string?

packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
+      const joinTableName = attributes[ENTITY_ASSIGNEE_ATTRIBUTE].joinTable.name;
+      return {
+        name: joinTableName,
+        dependsOn: [{ name: tableName }],
+      };
+    };
+
+    const stageJoinTablesToPersist = pipe([
+      getVisibleContentTypesUID,
+      filter((uid: any) => hasStageAttribute(contentTypes[uid])),
```
markkaylor, line 89: string? 🤔

packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
+
+    // Remove previously created join tables and persist the new ones
+    await removePersistedTablesWithSuffix('_strapi_stage_lnk');
+    await persistTables(stageJoinTablesToPersist);
+
+    const assigneeJoinTablesToPersist = pipe([
+      getVisibleContentTypesUID,
+      filter((uid: any) => hasStageAttribute(contentTypes[uid])),
+      map(getAssigneeTableToPersist),
+    ])(contentTypes);
```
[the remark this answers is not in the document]
markkaylor: I may not understand exactly what you have in mind @remidej , but for what it's worth, I am personally not too bothered by the repetition here. It's all in one place, readable, and easy to follow. I feel like an abstraction might take away from that just to be more concise.

packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
+
+    const stageJoinTablesToPersist = pipe([
+      getVisibleContentTypesUID,
+      filter((uid: any) => hasStageAttribute(contentTypes[uid])),
+      map(getStageTableToPersist),
+    ])(contentTypes);
+
+    // Remove previously created join tables and persist the new ones
+    await removePersistedTablesWithSuffix('_strapi_stage_lnk');
+    await persistTables(stageJoinTablesToPersist);
```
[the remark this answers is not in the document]
Marc-Roig: indeed, will add a comment to make it clearer.
There might be content types that have been removed or modified, hence the persisted table list has to be recreated again.

Its inefficient, and hacky. But its the only way we found to make this work at the time

packages/core/review-workflows/server/src/register.ts
```diff
@@ -56,33 +54,62 @@ function extendReviewWorkflowContentTypes({ strapi }: { strapi: Core.Strapi }) {
...
-//     };
-
-//     const joinTablesToPersist = pipe([
-//       getVisibleContentTypesUID,
-//       filter((uid: any) => hasStageAttribute(contentTypes[uid])),
-//       map(getStageTableToPersist),
-//     ])(contentTypes);
-
-// await removePersistedTablesWithSuffix('_strapi_stage_links');
-// await persistTables(joinTablesToPersist);
```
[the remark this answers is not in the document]
Marc-Roig: nope, looks like we were not doing that in v4 either

packages/core/review-workflows/server/src/register.ts
Marc-Roig: we do! will add it :)

packages/core/review-workflows/server/src/register.ts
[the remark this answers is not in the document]
Marc-Roig: :cheffkiss:

packages/core/admin/ee/server/src/services/persist-tables.ts
```diff
@@ -1,7 +1,7 @@
 import type { Core } from '@strapi/types';
 import { differenceWith, isEqual } from 'lodash/fp';
 
-interface PersistedTable {
+export interface PersistedTable {
```
markkaylor, line 4: I just don't see where you are using this outside this file 🤔

markkaylor (approved):
Nice, looks good to me.

packages/core/admin/ee/server/src/services/persist-tables.ts
Marc-Roig: typescript was complaining about it, if another file is importing a function that references this type, it also needs to be exported to be visible (or thats what I was able to understand)

PR #21986, opened 2024-10-25
fix(e2e): enable e2e tests in editview.spec.ts

markkaylor:
### What does it do?

- Turns on the e2e tests for content-manager editview
- Turns on the draft relations check when publishing

NOTE:

It doesn't look like we get the draft relation warning on the list page with bulk publish. I think it should be another PR...maybe there is already an issue or ticket about it 🤷‍♂️ 

### Why is it needed?

- To test and make things work as expected

### How to test it?

- The tests should pass in the CI


packages/core/content-manager/admin/src/router.tsx
```diff
@@ -55,7 +55,7 @@ const CollectionTypePages = () => {
 const CLONE_RELATIVE_PATH = ':collectionType/:slug/clone/:origin';
 const CLONE_PATH = `/content-manager/${CLONE_RELATIVE_PATH}`;
 const LIST_RELATIVE_PATH = ':collectionType/:slug';
-const LIST_PATH = `/content-manager/${LIST_RELATIVE_PATH}`;
+const LIST_PATH = `/content-manager/collection-types/:slug`;
```
markkaylor, line 58: ℹ️ Single types do not have a list page so I think this makes sense...

tests/e2e/tests/content-manager/editview.spec.ts
```diff
@@ -15,51 +15,50 @@ test.describe('Edit View', () => {
...
+      await page.getByLabel('Coach BeardDraft').click();
+      // Attempt to publish the entry
+      await page.getByRole('button', { name: 'Publish' }).click();
 
-        // Verify that a warning about a single draft relation is displayed
-        await expect(page.getByText('This entry is related to 1')).toBeVisible();
-        await page.getByRole('button', { name: 'Cancel' }).click();
+      // Verify that a warning about a single draft relation is displayed
+      await expect(page.getByText('This entry is related to 1')).toBeVisible();
+      await page.getByRole('button', { name: 'Cancel' }).click();
```
Marc-Roig, line 36: I thought we removed the count to draft relations, as we had a couple of issues with that

Marc-Roig (dismissed):

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -685,11 +666,7 @@ const PublishAction: DocumentActionComponent = ({
   };
 
   const totalDraftRelations = localCountOfDraftRelations + serverCountOfDraftRelations;
-
-  // TODO skipping this for now as there is a bug with the draft relation count that will be worked on separately
-  // see RFC "Count draft relations" in Notion
-  const enableDraftRelationsCount = false;
-  const hasDraftRelations = enableDraftRelationsCount && totalDraftRelations > 0;
```
markkaylor, line 692: @Marc-Roig Is this what your are talking about? The check was turned off but I turned it on and it seemed to be working. Are you aware of a bug here or is this safe to merge?

tests/e2e/tests/content-manager/editview.spec.ts
markkaylor: See other comment. I turned them on but maybe I should be commenting this check here instead. Lmk what you think.

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
Marc-Roig: I believe the draft relation count is not working properly (logic) and we decided to disable it. It always count relations as draft if I recall correctly. 

But if we are still showing that message, lets keep the e2e test as well 🤷‍♂️

markkaylor: The message only shows after removing / setting `enableDraftRelationsCount` to true. So if I think I'll need to set it to false again here and then comment out the check in the e2e tests since the dialog message won't show up anymore after that.

Marc-Roig: ah ! I thought this flag was in the e2e tests. Yes, we should keep it as false here !
The backend logic needs a bit of love

markkaylor: Got it! Thanks and good catch!

markkaylor:
@Marc-Roig I disabled the draft relation check for now and skipped the relevant tests. Could you give this one another review?

Marc-Roig (approved):

Issue #21499, opened 2024-09-27
Copying API token to clipboard triggers "save" action


innerdvations:
### Required System information

Environment: development
OS: darwin-arm64
Strapi Version: 5.0.2-beta.0
Node/Yarn Version: yarn/4.5.0 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite
### Describe the bug

When clicking the "copy to clipboard" icon after creating an API token, I get a "Success: API Token successfully edited" toast message

It actually sends a 200 PUT  network request as well to update the token, which could potentially cause some issues

Note: I noticed similar behaviour on the role permissions editor as well, where clicking on the configure button triggered a save, so the bug is probably at a higher level than api tokens

### Steps to reproduce the behavior

1. Create an API token
2. Click the copy to clipboard icon
3. the "edited" popup appears

### Expected behavior

Should not see "API Token successfully edited"

### Screenshots

![image](https://github.com/user-attachments/assets/65f546d8-e05b-4e7c-b9bb-627d7f0a1d19)

### Code snippets

If applicable, add code samples to help explain your problem.

sah246ir:
do i need to get this issue assigned to me so that i can work on it ?

innerdvations:
Hi @sah246ir , we don't assign issues but you're free to work on it if you'd like to contribute!

sah246ir:
understood!

Issue #21901, opened 2024-10-17
[V5] Admin Panel -> Roles, data is saved when clicking on settings for contidions


butcherZ:
### Required System information

Node.js version: v20.17.0
NPM version: 10.8.2
Strapi version: v5.1.0
Database: sqlite
Operating system: MacOS 14.6.1 (23G93)
Is your project Javascript or Typescript: TypeScript

### Describe the bug
 
Admin Panel -> Roles, data is saved when clicking on settings button
### Steps to reproduce the behavior

1. Go to Admin Panel
2. Click on Roles
3. Add a new Role or selecting an exsisting Role
4. Click on `settings` button (mouse hover the current row)
5. data will be saved before displaying the desired settings pop up

### Expected behavior

data should only be saved when clicking the save button, setting shouldn't trigger a save action.

### Screenshots

https://www.loom.com/share/f770cb11645c4980b760b48d5ab22fac?sid=0cb837d3-57d8-4673-a2b1-42ed8e4393a4

### Additional context

don't know if it could be related to https://github.com/strapi/strapi/issues/21499

oiorain:
This has been fixed.

PR #21988, opened 2024-10-25
fix: prevent submitting form on button clicks

kaioduarte:
### What does it do?

It prevents some forms from being submitted when a button is clicked, such as the Copy token button on the API Token page or the Settings button on the Roles form.

### Why is it needed?

The `Button` and `IconButton` components from the `@strapi/design-system` package don't specify a type for the buttons, so by default, they have `type="submit"`, which submits the form on click.

### How to test it?

This PR covers two scenarios, there might be others with the same issue.

- API Token
1. Create an API Token
2. Click on the copy button
3. It should only copy the token and nothing else

- Roles form
1. Open the Roles form
2. Click to edit an existing role
3. Open the Settings tab
4. Expand any row and click on the Settings button
5. It should open a confirm modal and nothing else

### Sanity checks

| Case | Before | After |
| -- | -- | -- |
| API Token | <video src="https://github.com/user-attachments/assets/b1c6ab05-9a88-40f5-bde0-825cac682751"> | <video src="https://github.com/user-attachments/assets/b91826ea-6e89-4450-93f6-dec642d87e14"> |
| Roles form | <video src="https://github.com/user-attachments/assets/9a0d3be4-d9c7-435a-89e4-7f1bcc8f32ed"> | <video src="https://github.com/user-attachments/assets/5e666a6a-8f41-429a-9241-76fcc78f7c8f"> |

### Related issue(s)/PR(s)

Closes https://github.com/strapi/strapi/issues/21499 and https://github.com/strapi/strapi/issues/21901

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/components/Tokens/TokenBox.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/components/Tokens/TokenBox.tsx
@@ -47,6 +47,7 @@ export const TokenBox = ({ token, tokenType }: TokenBoxProps) => {
               })}
               onClick={handleClick(token)}
               variant="ghost"
+              type="button"
               style={{ padding: 0, height: '1.6rem' }}
             >
               <Duplicate />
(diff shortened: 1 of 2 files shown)


hanpaine:
Hey @kaioduarte thanks for your contribution, it's much appreciated - we'll take a look asap 🙏 🚀

jhoward1994 (approved):
Thank you, looks like a good fix!

Issue #21358, opened 2024-09-18
Different behavior in the search between CTM menu and search


yanniskadiri:
I have to press enter to search in the CTM list view, I don’t have to do this in the CTM menu.

https://github.com/user-attachments/assets/2ff642c6-669a-4a0d-878a-e81f2757bc72

PR #21989, opened 2024-10-25
chore: match search between CTM menu and search

kaioduarte:
### What does it do?

Removes the need to press Enter to search on CTM by searching by typing with 500ms debounce.

### Why is it needed?

It'll match the behavior between the CTM menu and the search input, it's also a better UX IMO.

### How to test it?

1. Open CTM
2. Type something in the search input

https://github.com/user-attachments/assets/cd41b0bd-d8e8-4d6e-8880-e656bf7a2af2

### Related issue(s)/PR(s)

Closes https://github.com/strapi/strapi/issues/21358

### DIFF
--- a/packages/core/admin/admin/src/components/SearchInput.tsx
+++ b/packages/core/admin/admin/src/components/SearchInput.tsx
@@ -5,6 +5,7 @@ import { Search as SearchIcon } from '@strapi/icons';
 import { useIntl } from 'react-intl';
 
 import { TrackingEvent, useTracking } from '../features/Tracking';
+import { useDebounce } from '../hooks/useDebounce';
 import { useQueryParams } from '../hooks/useQueryParams';
 
 interface SearchInputProps {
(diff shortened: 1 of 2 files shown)


markkaylor:
> I'm good with the proposed solution, only question is for the debounce time. 500ms seems fair since we're doing some requests and want to avoid doing too much of them but could also be perceived as a bit too much for some people. What do you think @markkaylor?

I haven't tested it out but 500ms seems fine to me and I believe that's what we are using in other places for debounced search 👍

markkaylor:
It looks like the debounce is breaking the e2e tests, @kaioduarte if you have a chance to take a look that'd be great otherwise I'll come back later to fix them

Adzouz:
> It looks like the debounce is breaking the e2e tests, @kaioduarte if you have a chance to take a look that'd be great otherwise I'll come back later to fix them

Hey @markkaylor, I fixed the test but for some reason it appears that the front unit tests are stalled. Do you have any idea on what's happening?

PR #22008, opened 2024-10-29
feat: preview content in an iframe

Marc-Roig:
### What does it do?

Renders the preview url content inside an iframe.

> [!WARNING]  
> To render external URLs users will need to set that url into the CSP directives
```ts
// config/middlewares.js
module.exports = [
  // ...
  {
    name: 'strapi::security',
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          'frame-src': ["'self'", 'https://docs.strapi.io'], // e.g render strapi.io docs
        },
      },
    },
  },
  // ...
];

```

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
@@ -1,19 +1,28 @@
 import * as React from 'react';
 
+import { Box } from '@strapi/design-system';
+import { useIntl } from 'react-intl';
+
 import { usePreviewContext } from '../pages/Preview';
 
 const PreviewContent = () => {
   const previewUrl = usePreviewContext('PreviewContent', (state) => state.url);
 
+  const { formatMessage } = useIntl();
+
   return (
-    <iframe
+    <Box
       src={previewUrl}
-      title="Content Preview"
+      title={formatMessage({
+        id: 'content-manager.preview.panel.title',
+        defaultMessage: 'Preview',
+      })}
+      width="100%"
+      height="100%"
       style={{
-        width: '100%',
-        height: '100%',
         border: 'none',
       }}
+      tag="iframe"
     />
   );
 };
--- a/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
@@ -19,9 +19,7 @@ const PreviewContent = () => {
       })}
       width="100%"
       height="100%"
-      style={{
-        border: 'none',
-      }}
+      borderWidth={0}
       tag="iframe"
     />
   );
--- a/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react';
+
+import { Page } from '@strapi/admin/strapi-admin';
+
+import { usePreviewContext } from '../pages/Preview';
+
+const PreviewContent = () => {
+  const previewUrl = usePreviewContext('PreviewIframe', (state) => state.url);
+
+  if (!previewUrl) {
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Page } from '@strapi/admin/strapi-admin';
+
+import { usePreviewContext } from '../pages/Preview';
+
+const PreviewContent = () => {
+  const previewUrl = usePreviewContext('PreviewIframe', (state) => state.url);
+
+  if (!previewUrl) {
+    return <Page.Loading />;
+  }
```
remidej, line 12: Does typescript tell you previewUrl is nullable? Because I believe by the time this component renders it should always be available already. If not we need to fix the context's type. But the loading state should be handled at the page level, no need to add it here

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+
+import { Page } from '@strapi/admin/strapi-admin';
+
+import { usePreviewContext } from '../pages/Preview';
+
+const PreviewContent = () => {
+  const previewUrl = usePreviewContext('PreviewIframe', (state) => state.url);
```
remidej, line 8:
```suggestion
  const previewUrl = usePreviewContext('PreviewContent', (state) => state.url);
```

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
Marc-Roig: shouldn't be null! will remove this :)

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <iframe
+      src={previewUrl}
+      title="Content Preview"
+      style={{
+        width: '100%',
+        height: '100%',
+        border: 'none',
+      }}
+    />
```
remidej, line 17: - the title needs to be translated. Not sure if there's a json key that can be reused
- we try to use the design system as much as possible for styles. Could you try implementing this with `<Box tag="iframe"`?

remidej:
Should we add the CSP change to the getstarted application? For testing convenience, so that it works out of the box for us

(sorry for the comment spam)

remidej:
Should we add the CSP change to the getstarted application? For testing convenience, so that it works out of the box for us

(sorry for the comment spam)

remidej (approved):
Tested ✅ let's goooo

thematrixdev:
Excuse me, where can I find this feature on admin panel?

Marc-Roig:
@thematrixdev this is still a WIP feature, but if you are interested these are the steps you should follow:

1. (`src/features.js`) ensure preview feature flag is enabled
```js
  future: {
    preview: true
  },
});
```

2. (`src/admin.js`) Update your preview config to return different urls for draft and published content, e.g:
```ts
  preview: {
    enabled: true,
    config: {
		// This is just a dumb example
        if (status === 'published') {
        }

      },
    },
  },
```

3. (`src/middlewares`) Update your security policy to allow rendering your website inside strapi
```
[
  // ...
  {
    name: 'strapi::security',
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          'frame-src': ["'self'", 'https://docs.strapi.io/'], // e.g render strapi.io docs, but here you should put your website
        },
      },
    },
  },
  // ...
];
```

thematrixdev:
@Marc-Roig Thank you very much! It would be great if the URLs could be specified in Content-Type Builder.

oiorain:
Very happy to see the excitement around this feature!! 
We're planning to release an alpha on November 20th, a beta on December 4th and the stable should be out on the 18th 🚀

Issue #22015, opened 2024-10-29
Strapi v5 Fractional Indexing Issue with Drag & Drop


glenthompson-sng:
### Required System information
- Node.js version: v20.11.1
- NPM version: 10.2.4
- Yarn version: 1.22.21
- Strapi version: 5.1.1
- Database: postgres
- Operating system: Mac
- Is your project Javascript or Typescript: TS

### Describe the bug

Unable to save order of one to many relationships when pagination + drag and drop from Strapi admin UI.

<img width="385" alt="image" src="https://github.com/user-attachments/assets/912c9f95-816b-49b0-b9ec-0ae5a92cb71e">

### Steps to reproduce the behavior

I use sticky notes to explain the issue as I progress through the recording.

https://github.com/user-attachments/assets/fbfb7ef6-0bf3-42b3-9d1c-87a144eea184

### Expected behavior

Drag and drop order saved with no key space exhaustion error thrown.

### Screenshots

![image](https://github.com/user-attachments/assets/b41109f5-2e6e-4a28-b7f2-99e7ca621e1c)

#### 1. **Key Space Exhaustion**
   - The space between beforeKey and afterKey can be too narrow, making it impossible to generate unique keys between them.
   - This often happens when paginated results are added and the temporary keys overlap or are close to existing keys, especially near the upper boundary like "Zv".

#### 2. **Boundary Clustering from Reordering**
   - If initial keys are set near high boundary values like "Zv", subsequent key generation may quickly cluster around these limits, limiting available space.
   - During pagination or reordering, inadequate key adjustments can result in both beforeKey and afterKey becoming identical ("Zv"), leading to collisions.

#### 3. **Impact of Minified Code Optimization**
   - Minified code, optimized by esbuild, can alter the execution order, variable scope, and expression evaluation, which impacts the key generation logic.
   - These optimizations can cause key conflicts by skipping checks or changing recursion behavior, resulting in unexpected duplicates like "Zv" in production.
   - Local system linking from the non-minified `strapi/node_modules/@strapi` to my project `node_modules` folder allows for the pagination + drag & drop to **work correctly** (as seen in above recording).

Adzouz:
Hey @glenthompson-sng,

Closing this issue since I couldn't reproduce. I think it was fixed in newer versions of Strapi.

Here's a video of my test:

https://github.com/user-attachments/assets/27f999cb-1de9-4ba5-98fb-d7c4c54b65c2

Don't hesitate to reopen if you reproduce it.
[...]

PR #22019, opened 2024-10-30
Prevent key collisions and or exhaustion

particledream:
### What does it do?

The update enhances the `prepareTempKeys` with the ability to manage temporary keys across multiple operations, making it more reliable and preventing duplicate keys during pagination, drag-and-drop, and reordering scenarios, even when minified in production.

### Why is it needed?

When reordering occurs, keys need to be generated between existing keys. If the key space is narrow (e.g., close to "Zv"), adding new items can result in overlapping keys, triggering key collisions or errors like a >= b.

### How to test it?

Create a one to many relation, paginate through relations in admin UI, reorder and save. Return to validate position is as expected after save.

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/22015

### DIFF
--- a/packages/core/content-manager/admin/src/services/relations.ts
+++ b/packages/core/content-manager/admin/src/services/relations.ts
@@ -1,4 +1,4 @@
-import { generateNKeysBetween } from 'fractional-indexing';
+import { generateNKeysBetween, generateKeyBetween } from 'fractional-indexing';
 
 import {
   RelationResult as RelResult,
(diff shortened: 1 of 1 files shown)


hanpaine:
Hey @particledream thanks for your contribution! 🚀 we'll be taking a look soon 🙏

Adzouz:
Hi @particledream,
Thanks so much for your contribution and so sorry it took so much time 😢
After testing again on my side, I couldn't reproduce the issue (I put a video of testing in the original issue ticket). I think it has been resolved in newer versions of Strapi. I'll let you reopen the PR again if anything.
Closing it in the meantime, have a great day 🙏

PR #22025, opened 2024-10-30
wip: static preview draft & publish tabs

Marc-Roig:
### What does it do?

Add Draft / Publish tabs to switch between draft and live content in Static Preview.

image

Requirements:
- Clicking on draft tab will return the status=draft url from the config
- Clicking on published tab will return the status=published url from the config
- If document is not published, only the draft tab will be available
- If the content type doesn't have draft and publish, tabs will not be visible

### Why is it needed?

To preview both draft and published data.

### How to test it?
1. `.env` set environment variable
`STRAPI_FUTURE_PREVIEW=true`

2. (`src/features.js`) ensure preview feature flag is configured as:
```js
  future: {
    preview: env.bool('STRAPI_FUTURE_PREVIEW', false),
  },
});
```

2. (`src/admin.js`) Update your preview config to return different urls for draft and published content, e.g:
```ts
  preview: {
    config: {
        if (status === 'published') {
        }

      },
    },
  },
```

3. (`config/middlewares.js`) update CSP policies:
```ts
module.exports = [
  // ...
  {
    name: 'strapi::security',
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          'frame-src': ["'self'", 'https://docs.strapi.io/'], // e.g render strapi.io docs
        },
      },
    },
  },
  // ...
];
```


Marc-Roig:
I am going to do a PR on top of this one to fix the close button

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -151,6 +140,56 @@ const DocumentStatus = () => {
...
+            {formatMessage({
+              id: 'preview.tabs.draft',
+              defaultMessage: 'draft',
+            })}
+          </StatusTab>
+          <StatusTab value="published" disabled={documentStatus === 'draft'}>
+            {formatMessage({
+              id: 'preview.tabs.published',
+              defaultMessage: 'published',
+            })}
```
remidej, line 185: You should use `content-manager.containers.List.(modified|draft|published)` to avoid adding a new key

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -177,32 +216,50 @@ const PreviewHeader = () => {
...
+        <Flex gap={3}>
+          <ClosePreviewButton />
+          <Typography tag="h1" fontWeight={600} fontSize={2}>
+            {title}
+          </Typography>
+          <DocumentStatus />
+        </Flex>
+      </Grid.Item>
+
+      {/* Tabs */}
```
remidej, line 239: Tiny nitpick but we don't usually put blank lines within our JSX

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -177,32 +216,50 @@ const PreviewHeader = () => {
...
+          <LinkIcon />
+        </IconButton>
+      </Grid.Item>
+    </Grid.Root>
   );
 };
 
+const StatusTab = styled(Tabs.Trigger)`
+  text-transform: uppercase;
+`;
```
remidej, line 263: This isn't necessary with the right translation keys

Marc-Roig: At some point this compo needs to  uppercase  it, as it compares the document status with the Status tab value. Do you think it would be preferable to upercase it [here](https://github.com/strapi/strapi/pull/22025/files#diff-4483fd0983d09b3ed6d90b4f62c82777caad79bc82ade8b242fbb3abe652f1d0R168)

markkaylor:
Not necessarily related to this PR but It looks like some padding is lost when draft and published is disabled on the content-type

Screenshot 2024-11-04 at 11 07 23


Thanks for updating the how to test. I think you are missing one step still:

```
// features.js
  future: {
    preview: env.bool('STRAPI_FUTURE_PREVIEW', false),
  },
});

```

Otherwise I've tested it and it's working nicely. For the code, I don't have anything to add to what Rémi already suggested

Marc-Roig:
@markkaylor aagh , good catch. will do something about that 🤔

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
remidej: I believe it's only the `value` props (+ `onValueChange`) that need to match, what you render inside the trigger should be purely visual and disconnected from the logic

Marc-Roig: true! alright :) last questoin then, do I create a new key for this?  as it looks like content-manager.containers.List.(published|draft) are not upper cased

remidej: The value seems to be uppercased, for english at least: https://github.com/strapi/strapi/blob/5c463a4f3349c33b8034dcda66a04758d0efc542/packages/core/content-manager/admin/src/translations/en.json#L104-L106

remidej: Oooh sorry I thought your CSS was only capitalizing the first letter! (because that's done in a few places in the CM and I don't like it 😁)

But yes for this your code is perfect 💯

Marc-Roig:
cc @markkaylor @remidej just pushed some changes, thank you Mark for the help 🤟

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -221,6 +221,7 @@ const PreviewHeader = () => {
       gridCols={3}
       paddingLeft={2}
       paddingRight={2}
+      style={{ gridTemplateRows: '47px' }}
```
remidej, line 224: Where does 47px come from? Is this technique used elsewhere? I don't think we should be resorting to custom styles here, I think the DS should be able to handle this padding on its own

Marc-Roig: lets discuss this tomorrow!  I agree we should find a way to make this be handled by the DS :) 

There were two options here, add a paddiing to the Header component (which was the orignal approach before this PR), or add no padding and use the Tabs component (if you check the design, it doesnt have any [padding](https://www.figma.com/design/BDCSSHnOZutJ4aaYTs41Bg/Preview?node-id=586-15081&node-type=canvas&t=G5Xjb3o7hmzuoyq5-0) )

Problem is D&P can be disabled and those tabs are not rendered, leavng the header without any padding (see Mark's [comment](https://github.com/strapi/strapi/pull/22025#issuecomment-2454293305))

To avoid using absolute positioining or negative paddings, we opted for this solution. But I would be glad to find something else!

remidej: How about setting padding on the left and on the right separately?

Marc-Roig: that we are already doing, but we also need vertical padding when there is no D&P tabs

remidej: That's what I mean too, setting the vertical padding once for each section:

```jsx
<Grid paddingLeft={2} paddingRight={2}>
  <Flex paddingTop={1} paddingBottom={1}>
     // close btn + title + status
  </Flex>
  <MaybeTabs />
  <Flex paddingTop={1} paddingBottom={1}>
    // icon button
  </Flex>
  ```

Comment (Marc-Roig) [reply packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx:224]: ah! 😄 that sounds like a good solution 🏃

Comment (markkaylor) [reply packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx:224]: I think actually it might be 48px. It's in the design file for the height of the header. 

Since we are using grid here, I think grid-template-rows default value is auto and will take the height of its children. We only have one row and we want to set a fixed height.

Not sure why the prop isn't there as it is for gridCols, but we could easily do the same. My guess is we don't often need to explicitly set the height of rows in a grid.

https://github.com/strapi/design-system/blob/107f3b86f96a6dfb080a0d4fedca366ef895bd5c/packages/design-system/src/primitives/Grid/Grid.tsx#L32

I'm not as bothered by the custom style as I am about the style tag vs a styled component that could get the value from the theme. `sizes[9]` I think. 

https://github.com/strapi/design-system/blob/107f3b86f96a6dfb080a0d4fedca366ef895bd5c/packages/design-system/src/themes/common-theme.ts#L18

Comment (markkaylor) [reply packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx:224]: > That's what I mean too, setting the vertical padding once for each section:
> 
> ```js
> <Grid paddingLeft={2} paddingRight={2}>
>   <Flex paddingTop={1} paddingBottom={1}>
>      // close btn + title + status
>   </Flex>
>   <MaybeTabs />
>   <Flex paddingTop={1} paddingBottom={1}>
>     // icon button
>   </Flex>
> ```

If this works, great, do that!

Comment (Marc-Roig) [reply packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx]: Done! That seems to work ok

We are still not handling long title names, I was thinking of using a max width and overflowing the text with an ellipsis, I asked Lucas and we can handle that in another PR
image

File: packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx [hunk 5]
```diff
@@ -177,32 +216,48 @@ const PreviewHeader = () => {
...
+          </Typography>
+          <DocumentStatus />
+        </Flex>
+      </Grid.Item>
+      {/* Tabs */}
+      <Grid.Item xs={1}>
+        <PreviewTabs />
+      </Grid.Item>
+      {/* Copy link */}
+      <Grid.Item xs={1} justifyContent={'end'} paddingTop={2} paddingBottom={2}>
```
markkaylor, line 243:
```suggestion
      <Grid.Item xs={1} justifyContent="end" paddingTop={2} paddingBottom={2}>
```

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -177,32 +216,48 @@ const PreviewHeader = () => {
...
-          defaultMessage: 'Copy preview link',
-        })}
-        onClick={handleCopyLink}
-      >
-        <LinkIcon />
-      </IconButton>
-    </Flex>
+      {/* Title and status */}
+      <Grid.Item xs={1} paddingTop={2} paddingBottom={2}>
+        <Flex gap={3}>
```
markkaylor, line 230: Can you not put the gap on the Grid item and avoid the Flex component?

markkaylor:
Heads up there is a small gap beneath the tabs

Screenshot 2024-11-05 at 10 16 50

remidej (commented):
I think we should update the e2e tests to check that the tabs appear when the content type has D&P

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
remidej: Okay let's go ahead with this 🚂  Lucas gave me more context

lucasboilly: Hey guys, I just discussed with @Marc-Roig and we decided to simply set a 200px maxWidth with an ellipsis. As he said, no need to spend too much time on this for now 👍

> if they are not at the center of the screen, switiching between draft and publish would make the tabs move (if the title is different)

That's exactly why the tabs were originally centered, but I had too change this for the side editor iteration (that's a long story) 😄

markkaylor:
Didn't see this was up for review again. I'm on it 👀

Marc-Roig:
I am still missing the e2e tests but yes! :)

markkaylor:
Idk what's up but the build is failing on this branch for me locally 😢 

Screenshot 2024-11-07 at 14 18 36

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -228,13 +228,14 @@ const PreviewHeader = () => {
...
-        <Typography tag="h1" fontWeight={600} fontSize={2}>
+        <Typography tag="h1" fontWeight={600} fontSize={2} maxWidth={'200px'}>
           {title}
         </Typography>
         <DocumentStatus />
       </Grid.Item>
       {/* Tabs */}
-      <Grid.Item xs={1}>
+      {/* <Grid.Item xs={1}> */}
+      <Grid.Item xs={1} style={{ alignSelf: 'end' }}>
```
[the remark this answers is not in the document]
Marc-Roig: doesnt look like alignSelf is part of the api exposed to the grid / flex compo, do you know any alternative?  Icould just align everything to the bottom but that seems off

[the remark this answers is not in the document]
Marc-Roig: > I'm still seeing the gap, I think it's the border of the header that's causing the issue. Could we just remove the border?

I added a negative margin to remove the gap! I would rather not remove the border, as it's in the design and its a great way to separate the spaces imo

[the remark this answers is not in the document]
Marc-Roig: nice! I figured the api! will try not to use the style tag anymore 🫡 thx!

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -46,4 +47,26 @@ describeOnCondition(edition === 'EE')('Preview', () => {
...
+
+    // Check that preview opens in its own page
+    await clickAndWait(page, page.getByRole('link', { name: /open preview/i }));
+    // Draft status is visible (second Draft text is the draft tab)
+    await expect(page.getByText(/^Draft$/).nth(0)).toBeVisible();
+    await expect(page.getByRole('heading', { name: /west ham post match/i })).toBeVisible();
+
+    // Verify that Draft and Publish tabs are visible
+    await expect(page.getByText(/^Draft$/).nth(1)).toBeVisible();
+    await expect(page.getByText(/^Published$/)).toBeVisible();
```
remidej, line 67: You can select by role here, which is better as it makes sure code is accessible too. There are tab, tablist, and tabpanel roles

remidej (commented):
Test and works well! Just 1 small e2e thing

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -63,8 +63,9 @@ describeOnCondition(edition === 'EE')('Preview', () => {
     await expect(page.getByRole('heading', { name: /west ham post match/i })).toBeVisible();
 
     // Verify that Draft and Publish tabs are visible
-    await expect(page.getByText(/^Draft$/).nth(1)).toBeVisible();
-    await expect(page.getByText(/^Published$/)).toBeVisible();
+    // await expect(page.getByText(/^Draft$/).nth(1)).toBeVisible();
```
remidej, line 66: oops

remidej (approved):
Just a rogue comment otherwise lgtm!

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -46,4 +47,27 @@ describeOnCondition(edition === 'EE')('Preview', () => {
...
+    await clickAndWait(page, page.getByRole('gridcell', { name: /west ham post match/i }));
+
+    // Check that preview opens in its own page
+    await clickAndWait(page, page.getByRole('link', { name: /open preview/i }));
+    // Draft status is visible (second Draft text is the draft tab)
+    await expect(page.getByText(/^Draft$/).nth(0)).toBeVisible();
+    await expect(page.getByRole('heading', { name: /west ham post match/i })).toBeVisible();
+
+    // Verify that Draft and Publish tabs are visible
+    // await expect(page.getByText(/^Draft$/).nth(1)).toBeVisible();
```
Marc-Roig, line 66:
```suggestion
```

PR #22027, opened 2024-10-30
fix: preview close button

Marc-Roig:
### What does it do?

Fix preview close button redirecting to the edit view regardless of having switched to Draft & publish.

### How to test it?

Open preview, switch to publish & draft multiple times and close preview.

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
@@ -32,18 +32,14 @@ const ClosePreviewButton = () => {
   const [{ query }] = useQueryParams();
   const { formatMessage } = useIntl();
 
-  const closePreview = () => {
-    return {
-      pathname: '..',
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -30,38 +30,20 @@ import { usePreviewContext } from '../pages/Preview';
...
+    };
   };
 
   return (
     <IconButton
       tag={Link}
-      to={history.at(-1) ?? fallbackUrl}
-      onClick={handleClick}
+      relative="path"
+      to={closePreview()}
```
remidej, line 46: I don't think the function is needed anymore, you can inline the object imo

remidej (commented):
Yeah that was overly complex, good to keep it simple :)

remidej (approved):

remidej:
looks like there's a TS error

PR #21882, opened 2024-10-16
feat(content-manager): integrate review workflows with releases

Feranchz:
### What does it do?

Introduces new feature Releases x Review Workflows. All the commits on this PR have been already approved before.

PR #22028, opened 2024-10-30
fix(e2e): update accessible name for flaky webkit test in history.spec.ts 

markkaylor:
### What does it do?

Update the accessible name to be the same as what is in the Dom

### Why is it needed?

The webkit test is flaky

### How to test it?

The e2e tests should pass

### DIFF
--- a/tests/e2e/tests/content-manager/history.spec.ts
+++ b/tests/e2e/tests/content-manager/history.spec.ts
@@ -252,7 +252,7 @@ describeOnCondition(edition === 'EE')('History', () => {
       await clickAndWait(page, page.getByRole('link', { name: /Create new entry/, exact: true }));
       await page.waitForURL(ARTICLE_CREATE_URL);
       await page.getByRole('textbox', { name: 'title' }).fill('Being from Kansas');
-      await page.getByRole('textbox', { name: 'slug' }).fill('being-from-kansas');
+      await page.getByRole('textbox', { name: 'slug This value is unique for the selected locale' }).fill('being-from-kansas');
       await page.getByRole('button', { name: 'Save' }).click();
       await page.waitForURL(ARTICLE_EDIT_URL);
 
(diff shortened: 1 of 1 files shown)


tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -252,7 +252,9 @@ describeOnCondition(edition === 'EE')('History', () => {
       await clickAndWait(page, page.getByRole('link', { name: /Create new entry/, exact: true }));
       await page.waitForURL(ARTICLE_CREATE_URL);
       await page.getByRole('textbox', { name: 'title' }).fill('Being from Kansas');
-      await page.getByRole('textbox', { name: 'slug' }).fill('being-from-kansas');
+      await page
+        .getByRole('textbox', { name: 'slug This value is unique for the selected locale' })
+        .fill('being-from-kansas');
```
remidej, line 257: I'm all for it if the tests pass reliably, but I admit I don't understand it. Do you know what the error in the CI currently is? Does it match several items? because it should be able to find a match even if we don't provide the whole dom text

remidej (approved):

tests/e2e/tests/content-manager/history.spec.ts
markkaylor: The test wasn't even failing here. It's failing later when trying to find a value in the listview gridcell, but it can't because it was never filled. But I took another look and it does look like this is always filled even though sometimes on webkit there is no value in the listview.

> because it should be able to find a match even if we don't provide the whole dom text

Yeah I had the same thought but since it's only failing for webkit I was assuming it was a limitation of that browser 🤷‍♂️ 

This does seem to fix this issue consistently (I've run the same fix on the review workflow pr as well) but I can't really explain why. 

In other news it looks like there are other issues with the history tests. So looks like I'll be digging deeper.

simotae14 (approved):

markkaylor:
This was merged through https://github.com/strapi/strapi/pull/21882, but still investigating flaky e2e tests

PR #22062, opened 2024-11-04
feat(cloud-cli): add command to link project with environment

gonbaum:
### What does it do?

This PR adds a command for Cloud CLI users to link their local project to a specific cloud environment.

With strapi environment link, users can select an environment from their available options to set as the default for deploying their local project.

If a user runs strapi deploy without a linked environment or an `-e / --env` flag, they will be prompted to choose an environment.
Once an environment is linked, the strapi deploy command will use it automatically, without further prompts.

**Additional changes**

- Duplicated configuration retrieval functions have been consolidated in a utils file.
- Type names have been updated for consistency.

### Why is it needed?

With the introduction of Strapi Cloud multi-environments, users need a way to deploy to specific environments via the CLI. This command completes the essential environment features, allowing users to deploy to a selected environment by either choosing it manually, using a flag, or setting a default environment.

### How to test it?

---

**Pretest 1** 
- New options should be displayed on running `strapi cloud environment -h:`

```
  Manage environments

Options:
  -h, --help      Display help for command

Commands:
  link [options]  Link project to a specific Strapi Cloud project environment
  list [options]  List Strapi Cloud project environments
  help [command]  Display help for command
```

**Pretest 2**  
- Using any command without being logged in should prompt to login.

----
**Test Case 1: No Project in Local Config** 

- Steps: Clear the project from the local config (`$ rm .strapi-cloud.json`) and run `strapi cloud environment link`.
- Expected: Reject with an error about no project configured:
`[WARN] `
`We couldn't find a valid local project config.`
`Please link your local project to an existing Strapi Cloud project using the link command.`

**Test Case 2: Invalid Project in Cloud API** 

- Steps: Set an invalid project name (change it in .strapi-cloud.json file), run `strapi cloud environment link`.
- Expected: Reject with an error about an invalid project: 
`[WARN]`
`The project associated with this folder does not exist in Strapi Cloud.`
`Please link your local project to an existing Strapi Cloud project using the link command.`

**Test Case 3: Environment Selection** 

- Steps: Running `strapi cloud environment link` with >1 environment should prompt to select one environment:
- Expected: 
 	- All environments available in strapi cloud project should be displayed
 	- Selecting `Quit` option should cancel the operation.
 	- Selecting one env save the selected environment to the local config.
 	- Succesful operation should log:
 	`You have successfully linked your project to <environment>, on <project>. You are now able to deploy your project`

**Test Case 4: Default Environment in Deploy** 

- Steps: Set a default environment, run the deploy command.
- Expected: CLI should use the default environment to deploy and no question about the environment should be displayed.

**Test Case 5: Default Environment in and flag** 

- Steps: Set a default environment, run the deploy command with a flag `--env` to deploy to a different env than the default environment.
- Expected: CLI should NOT use the default environment to deploy but the environment passed with the flag
- repeat the test with using the short flag `-e`

**Test Case 6: Clear Environment on Relink** 

- Steps: Set a default environment and then re-link the project to a different strapi cloud project.
- Expected: target environment should be removed from `.strapi-config.json` file

**Test Case 7: Only environment linked** 

- Steps: In a project with only production environment, link production environment and then use the link command again
- Expected: Rejected operation with log ` The only available environment is already linked. You can add a new one from your project settings on the Strapi Cloud dashboard.`

### DIFF
--- a/packages/cli/cloud/src/environment/link/action.ts
+++ b/packages/cli/cloud/src/environment/link/action.ts
@@ -43,6 +43,14 @@ export default async (ctx: CLIContext) => {
   const environments = await getEnvironmentsList(ctx, cloudApiService, project);
 
   if (!environments) {
+    logger.debug(`Fetching environments failed.`);
+    return;
+  }
+
+  if (environments.length === 0) {
+    logger.log(
+      `The only available environment is already linked. You can add a new one from your project settings on the Strapi Cloud dashboard.`
+    );
     return;
   }
 
--- a/packages/cli/cloud/src/environment/link/action.ts
+++ b/packages/cli/cloud/src/environment/link/action.ts
@@ -65,12 +65,14 @@ export default async (ctx: CLIContext) => {
       environmentName: answer.targetEnvironment,
     });
     logger.debug('Failed to link environment', e);
-    logger.error('An error occurred while trying to link the environment.');
+    logger.error(
+      'Failed to link the environment. If this issue persists, try re-linking your project or contact support.'
+    );
     process.exit(1);
   }
 
   logger.log(
-    `You have successfully linked your project to ${chalk.cyan(answer.targetEnvironment)}, on ${chalk.cyan(project.displayName)}. You are now able to deploy your project.`
+    ` You have successfully linked your project to ${chalk.cyan(answer.targetEnvironment)}, on ${chalk.cyan(project.displayName)}. You are now able to deploy your project.`
   );
   await trackEvent(ctx, cloudApiService, 'didLinkEnvironment', {
     projectName: project.name,
@@ -117,13 +119,14 @@ async function getEnvironmentsList(
     const {
       data: { data: environmentsList },
     } = await cloudApiService.listLinkEnvironments({ name: project.name });
-    spinner.succeed();
 
-    return (environmentsList as unknown as Environment[])
-      .filter((environment: Environment) => environment.name !== project.targetEnvironment)
-      .map((environment: Environment) => {
-        return environment;
-      });
+    if (!Array.isArray(environmentsList) || environmentsList.length === 0) {
+      throw new Error('Environments not found in server response');
+    }
+    spinner.succeed();
+    return environmentsList.filter(
+      (environment: Environment) => environment.name !== project.targetEnvironment
+    );
   } catch (e: any) {
     if (e.response && e.response.status === 404) {
       spinner.succeed();
--- a/packages/cli/cloud/src/environment/link/action.ts
+++ b/packages/cli/cloud/src/environment/link/action.ts
@@ -133,7 +133,7 @@ async function getEnvironmentsList(
       ctx.logger.warn(
         `\nThe project associated with this folder does not exist in Strapi Cloud. \nPlease link your local project to an existing Strapi Cloud project using the ${chalk.cyan(
           'link'
-        )} command`
+        )} command.`
       );
     } else {
       spinner.fail('An error occurred while fetching environments data from Strapi Cloud.');
--- a/packages/cli/cloud/src/environment/link/action.ts
+++ b/packages/cli/cloud/src/environment/link/action.ts
@@ -66,7 +66,7 @@ export default async (ctx: CLIContext) => {
   });
 
   try {
-    await local.addEnvironment(answer.targetEnvironment);
+    await local.patch({ project: { targetEnvironment: answer.targetEnvironment } });
   } catch (e) {
     await trackEvent(ctx, cloudApiService, 'didNotLinkEnvironment', {
       projectName: project.name,
--- a/packages/cli/cloud/src/environment/link/action.ts
+++ b/packages/cli/cloud/src/environment/link/action.ts
@@ -1,5 +1,6 @@
 import chalk from 'chalk';
 import inquirer, { type Answers } from 'inquirer';
+import { ProjectInput } from '../../services/cli-api';
 import type { CLIContext, CloudApiService } from '../../types';
 import { cloudApiFactory, tokenServiceFactory, local } from '../../services';
 import { promptLogin } from '../../login/action';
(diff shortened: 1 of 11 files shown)


packages/cli/cloud/src/environment/link/action.ts
```diff
@@ -0,0 +1,140 @@
...
+) {
+  const spinner = ctx.logger.spinner('Fetching environments...\n').start();
+
+  try {
+    const {
+      data: { data: environmentsList },
+    } = await cloudApiService.listLinkEnvironments({ name: project.name });
+    spinner.succeed();
+
+    return (environmentsList as unknown as Environment[])
```
abdallahmz, line 122: Use type inference to avoid casting. Try to add this before the line:

```
    if (!Array.isArray(environmentsList)) {
      return [] or undefined or null, or throw error
    }
```

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -1,22 +1,36 @@
...
+  const storedData = {
+    ...alreadyInFileData,
+    project: {
+      ...alreadyInFileData.project,
+      targetEnvironment,
+    },
+  };
   const pathToFile = path.join(directoryPath || process.cwd(), LOCAL_SAVE_FILENAME);
-  // Ensure the directory exists
   await fse.ensureDir(path.dirname(pathToFile));
```
abdallahmz, line 24: I believe we should re-use the `save` function? (benefits: typings check on the data we write in the file)

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -1,22 +1,36 @@
...
-  project?: Omit<ProjectInfos, 'id'>;
+  project?: Omit<ProjectInfo, 'id'>;
 };
 
-export async function save(data: LocalSave, { directoryPath }: { directoryPath?: string } = {}) {
+export async function addEnvironment(
+  targetEnvironment: string,
+  { directoryPath }: { directoryPath?: string } = {}
+) {
   const alreadyInFileData = await retrieve({ directoryPath });
```
abdallahmz, line 15: What if "alreadyInFileData" is undefined or empty ({})? I think that's an error case that we need to add, otherwise we will just have the "targetEnvironment" in the save file (= incomplete / unexpected data).

gonbaum: It shouldn't be empty cause this function wouldn't be called otherwise. However, a check should be done in case the reading operation fails. Good catch!

packages/cli/cloud/src/services/strapi-info-save.ts
gonbaum: Yes, we can do that.

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -1,20 +1,35 @@
...
+      ...alreadyInFileData.project,
+      targetEnvironment,
+    },
+  };
+  await save(dataToSave, { directoryPath });
+}
+
+export async function save(data: LocalSave, { directoryPath }: { directoryPath?: string } = {}) {
   const pathToFile = path.join(directoryPath || process.cwd(), LOCAL_SAVE_FILENAME);
-  // Ensure the directory exists
```
nathan-pichon, line 15: You don't like my comments? 🥲
Maybe a more explicit one could be better like

// Ensure the directory exists and creates it if not.

As not everyone is knowing the functions use in fs-extra 🤔

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -1,20 +1,35 @@
...
   await fse.ensureDir(path.dirname(pathToFile));
-  await fse.writeJson(pathToFile, storedData, { encoding: 'utf8' });
```
nathan-pichon, line 17: Why did you removed the 
```
const storedData = { ...alreadyInFileData, ...data };
```

part?

gonbaum: I think we should use `save` to store the configuration when linking a project initially. It’s safer to overwrite all config data as this avoids carrying over any outdated data from a previously linked project. Then use a separate function to add specific details as needed—as I do with addEnvironment. Since addEnvironment is our only use case so far, I think we can avoid creating a fully abstract function, for now, to keep the code simpler.

nathan-pichon: Isn't this more safer to keep everything avoiding removing all data by mistake? I builded this function with this exact purpose in mind, if we add later something more than project, we would want to add/update this new property without worrying about any project property already existing inside the file.
If we want to erase (outdated or not) data from the config file, we can still pass { project: null } as data. 🤔

packages/cli/cloud/src/services/strapi-info-save.ts
```diff
@@ -1,20 +1,35 @@
...
 
 export const LOCAL_SAVE_FILENAME = '.strapi-cloud.json';
 
 export type LocalSave = {
-  project?: Omit<ProjectInfos, 'id'>;
+  project?: Omit<ProjectInfo, 'id'>;
 };
 
-export async function save(data: LocalSave, { directoryPath }: { directoryPath?: string } = {}) {
+export async function addEnvironment(
```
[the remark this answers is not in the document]
gonbaum: answer below

packages/cli/cloud/src/services/strapi-info-save.ts
gonbaum: [Update after discussion carried out of this thread]: we concluded that both approaches could be prone to errors (both replacing and preserving existing data). We resolved to rework the service with Create, Retrieve, Update, and Patch functions.

packages/cli/cloud/src/services/strapi-info-save.ts
gonbaum: Re-added comment

abdallahmz (commented):
LGTM, starting QA 👍

Issue #21429, opened 2024-09-23
The entry state is changed from Draft to Modified when the Review Workflow stage is updated


yanniskadiri:
### **Steps to reproduce**
- Go to the Content Manager
- Go to a content type with a review workflow assigned (you will need a paid license)
- Go to an entry with a draft state
- Change the stage of this entry
- You will see the state change from Draft to Modified

### **Expected behaviour**

Changing a stage should not change the entry state (Draft or Modified).

### **Screenshots**

https://github.com/user-attachments/assets/f7ed03cb-15de-41c4-a308-33a9e1e15f2e

Issue #20956, opened 2024-08-06
[v5] Unable to set Review Workflow Assignee


phil-janeapp:
## Bug report

Unable to set an "Assignee" for the **Review Workflows** section.  The server log has the following for that request:

```
[2024-08-06 23:01:41.181] error: select "t0"."id" from "public"."blog_posts" as "t0" where ("t0"."id" = $1) limit $2 - invalid input syntax for type integer: "isj1gyoqryyiyw6bnlvx8h3f"
error: select "t0"."id" from "public"."blog_posts" as "t0" where ("t0"."id" = $1) limit $2 - invalid input syntax for type integer: "isj1gyoqryyiyw6bnlvx8h3f"
    at Parser.parseErrorMessage (/strapi/app/tmp/cache/yarn/berry/cache/pg-protocol-npm-1.6.1-c0b26dda9d-10c0.zip/node_modules/pg-protocol/src/parser.ts:369:69)
    at Parser.handlePacket (/strapi/app/tmp/cache/yarn/berry/cache/pg-protocol-npm-1.6.1-c0b26dda9d-10c0.zip/node_modules/pg-protocol/src/parser.ts:188:21)
    at Parser.parse (/strapi/app/tmp/cache/yarn/berry/cache/pg-protocol-npm-1.6.1-c0b26dda9d-10c0.zip/node_modules/pg-protocol/src/parser.ts:103:30)
    at Socket.<anonymous> (/strapi/app/tmp/cache/yarn/berry/cache/pg-protocol-npm-1.6.1-c0b26dda9d-10c0.zip/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:519:28)
    at Socket.emit (node:domain:488:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:191:23)
```

### Required System information

- Node.js version: node/v20.15.1
- NPM version: N/A (yarn)
- Strapi version: 5.0.0-rc.8
- Database: postgres
- Operating system: linux-arm64
- Is your project Javascript or Typescript: TS

### Describe the bug

Unable to set an "Assignee" for the **Review Workflows** section.

### Steps to reproduce the behavior

1. Go to any custom content type
2. Click on **Assignee** in the **Review Workflows** card
3. Select any user

[...]

PR #22069, opened 2024-11-05
fix: stage change should not update modified status

Marc-Roig:
### What does it do?
Changing the stage/assignee attribute made the document status go from publshed to modifed, confusing the users.

This PR ensures changhing those values will not change the publicaton status of the document, by keeping the same updated at value when making the update on the entry.

### Why is it needed?

### How to test it?

Go to the Content Manager
Go to a content type with a review workflow assigned (you will need a paid license)
Go to an entry with a published state
Change the stage of this entry
You will see the state not changed from Published to Modified

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/21429

### DIFF
--- a/packages/core/review-workflows/server/src/services/assignees.ts
+++ b/packages/core/review-workflows/server/src/services/assignees.ts
@@ -1,4 +1,4 @@
-import type { Core, UID } from '@strapi/types';
+import type { Core, UID, Modules } from '@strapi/types';
 import { errors } from '@strapi/utils';
 import { isNil } from 'lodash/fp';
 import { ENTITY_ASSIGNEE_ATTRIBUTE } from '../constants/workflows';
@@ -10,7 +10,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
   const metrics = getService('workflow-metrics', { strapi });
 
   return {
-    async findEntityAssigneeId(id: string, model: UID.ContentType) {
+    async findEntityAssigneeId(id: string | number, model: UID.ContentType) {
       const entity = await strapi.db.query(model).findOne({
         where: { id },
         populate: [ENTITY_ASSIGNEE_ATTRIBUTE],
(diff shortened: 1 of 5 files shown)


Marc-Roig:
I will create an api test tomorrow to ensure the behaviour!

remidej (changes requested):
Clearing the assignee still changes the status
CleanShot 2024-11-07 at 10 16 44

packages/core/review-workflows/server/src/services/assignees.ts
```diff
@@ -24,42 +24,49 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
 
-      if (!userExists) {
-        throw new ApplicationError(`Selected user does not exist`);
+        if (!userExists) {
+          throw new ApplicationError(`Selected user does not exist`);
+        }
       }
 
-      metrics.sendDidEditAssignee(await this.findEntityAssigneeId(documentId, model), assigneeId);
+      const oldAssigneeId = await this.findEntityAssigneeId(entityToUpdate.id, model);
```
Marc-Roig, line 46: @remidej @markkaylor I will use this PR to also fix https://github.com/strapi/strapi/issues/20956 , we were sending the documentId to find the assignee id, but this method requires the entry id

Marc-Roig: We dont have any assignee api test, so if thats fine by you I will create another pr with new review workflow tests so we can merge this one and we can do a hot fix (after the review :D )

remidej: works for me 👍

markkaylor: Sounds good 👍

remidej (approved):
Tested 👌 thanks for the new tests!

PR #22071, opened 2024-11-05
fix(e2e): update data import and export scripts

markkaylor:
### What does it do?

- Migrates the export script to TypeScript
- Adds an import script 
- Updates the documentation so it is more clear how to properly update the data set.
- Removes two article entries in the database that were unnecessary

### Why is it needed?

- When following the current documentation the export script doesn't work
- The documentation suggests using the dts import with `yarn strapi import` which won't work since it won't include admin users and content-types such as `releases`.

### How to test it?

- Follow the documentation to import a data set
- Follow the documentation to export a data set

### DIFF
--- a/docs/docs/guides/e2e/02-data-transfer.md
+++ b/docs/docs/guides/e2e/02-data-transfer.md
@@ -9,83 +9,82 @@ tags:
 
 ## Overview
 
-This document looks particularly at the `@strapi/data-transfer` package framed around how we use it with our test instances in the e2e test suite. It is not a comprehensive explanation of how to use `@strapi/data-transfer`. For that information you should [view the documentation](https://docs.strapi.io/developer-docs/latest/developer-resources/data-management.html) surrounding it.
+This document explains how and why we use `@strapi/data-transfer` as means to reset and seed the database for end-to-end tests. It is not a comprehensive explanation of how to use `@strapi/data-transfer`. See the [Strapi documentation](https://docs.strapi.io/developer-docs/latest/developer-resources/data-management.html) to learn more about the feature.
 
-## Why use Data Transfer?
+### Why use Data Transfer?
 
-Each test should be isolated and contained e.g. if you edit an entity in one test, the next test shouldn't know or care about it otherwise your tests need to run in a specific order and become flakey quite quickly.
+We could use custom API endpoints of the application, and whilst this isn't a poor solution, it would _most likely_ require some code writing to set up the data for the schema entries. However, in `4.6.0` Strapi released the `DTS` feature (DTS – Data Transfer System). This means any member of Strapi can export the data of their instance producing a `.tar` that we can then import programatically restoring the database to this point in time and ensuring a "pure" test environment.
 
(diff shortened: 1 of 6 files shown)


tests/e2e/tests/i18n/editview.spec.ts
```diff
@@ -69,23 +69,7 @@ test.describe('Edit view', () => {
...
-      .toMatchObject({
-        contentTypeUID: 'api::product.product',
-        data: {
-          id: '',
-          isAvailable: true,
-          name: 'Camiseta de fuera 23/24 de Nike para hombres',
-          slug: 'product',
-        },
-        field: 'slug',
-      });
```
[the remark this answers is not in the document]
markkaylor: opened: https://github.com/strapi/strapi/actions/runs/11744784403/job/32720958990?pr=22110

innerdvations (dismissed):
Ahhh, thank you so much for this! All the changes lgtm, as long as the tests pass

docs/docs/guides/e2e/02-data-transfer.md
```diff
@@ -9,83 +9,79 @@ tags:
...
-    "test-apps/e2e/*",
-  ]
-```
+You should be able to login with the test app instance credentials.
+
+| Email            | Password    |
+| ---------------- | ----------- |
+| test@testing.com | Testing123! |
 
+Now that you have a Strapi instance witht he same data that each e2e starts with, you can modify the data in the CMS to prepare for a new data export.
```
remidej, line 54:
```suggestion
Now that you have a Strapi instance with the same data that each e2e starts with, you can modify the data in the CMS to prepare for a new data export.
```

tests/e2e/data/with-admin.tar
remidej, line 1: Can you explain the removal?

tests/e2e/scripts/dts-export.ts
```diff
@@ -1,4 +1,4 @@
-const { exportData } = require('../utils/dts-export.ts');
+import { exportData } from '../utils/dts-export';
 
 // TODO: make an actual yargs command and pass common options to exportData so it's easier to build the test data
```
remidej, line 3: no idea what this comment implies, but do you have any thoughts about it?

remidej (commented):
Thanks for diving into this 🙌

tests/e2e/data/with-admin.tar
markkaylor: The file is still there, I just removed two article entries that were duplicates of existing entries.

markkaylor:
sorry had to force push to make commit lint happy

remidej (approved):

tests/e2e/scripts/dts-export.ts
markkaylor: I think it implies that we could use `yargs` to make the arguments passed to the command a little more explicit. ie `node ./dts-export.ts` --file=my-file. Since we only have 1 argument for both scripts I'm not sure it's necessary right, but probably is a good DX improvement so I'm ok with leaving the comment. 

See how we use yargs for the yarn test:e2e command for example:

https://github.com/strapi/strapi/blob/c4a7837bc375b8e2fed4a3d2677d9f11115c348a/tests/scripts/run-e2e-tests.js#L67-L95

Issue #21848, opened 2024-10-15
Migration issue in Postgres


vedbhanushal1:
## Bug report

MigrationError: Migration 5.0.0-05-drop-slug-fields-index (up) failed: Original error: select * from information_schema.tables where table_name = $1 and table_schema = current_schema() - current transaction is aborted, commands ignored until end of transaction block

### Required System information

- Node.js version: V18.20.3
- NPM version:
- Strapi version: V5.0.6 (have tried with all patches of V5 getting in all)
- Database: Postgres
- Operating system: Ubuntu 22.04.3 LTS
- Is your project Javascript or Typescript: Typescript

### Describe the bug

earlier I was on Strapi Version : 5.0.0-beta.17 and when I have upgraded the version to 5.0.6
and building my application ```npm run build``` facing database migration issue.

kaue-delmonte-acn:
Same issue here.

kauedelmonte:
Same issue here.

kauedelmonte:
@Marc-Roig can you help us?

My and @vedbhanushal1 case was migrating from 5.x.x-beta.xx to 5.0.x.

Some prints:

Strapi log:
![image](https://github.com/user-attachments/assets/6cfbf147-b95a-4147-9d34-48a5c653543c)

Postgres log:
<img width="1123" alt="image" src="https://github.com/user-attachments/assets/a141f759-ae71-4b82-ac60-a412dda4f5d8">

joojp17:
Did you manage to find a solution?

innerdvations:
It appears this issue is only caused by migrating from a v5 beta to a release v5, so we are unable to support it. However, I can guess at the workaround based on the error message to get you up and running again :)

If you create unique indexes on based on the errors shown, each time you add one, it will give an error with the new one to create. Eventually (once you've created all the missing indexes) it should work.

That is, with the example shown by @kauedelmonte you would need to create a unique index called "blogs_slug_unique" on your "blogs" table

innerdvations:
Closing because it only appears to affect beta; if anyone faces this issue upgrading from a 4.x version, we can reopen

fmlukaszmatusik:
My solution

1. yarn add patch-package
2. create /patches directory (e.g. backend/patches)
3. create file `@strapi+database+5.26.0.patch` (with your strapi version in name)
4. add content below to this file

```
[...]

PR #22074, opened 2024-11-05
fix: run discardDrafts without concurrency

innerdvations:
### What does it do?

Removes concurrency from the discardDrafts migration

Unfortunately that means it will run slower, but we can work on safer performance improvements later and unblock those who are stuck upgrading to Strapi 5.

### Why is it needed?

When discardDrafts is run in parallel, there is something akin to a race condition where it tries to create relations to (or from?) an entity that is being changed at the same time. When run one at a time, the problem cannot occur.

Solves https://github.com/strapi/strapi/issues/21607

### How to test it?

Experimental: 0.0.0-experimental.a13c58eec89ab119f0e381fb79c0252979e9c125

In Strapi v4, with postgres:

- Create a `destination` schema like so:
```
{
  "kind": "collectionType",
  "collectionName": "destinations",
  "info": {
    "singularName": "destination",
    "pluralName": "destinations",
    "displayName": "Destinations",
    "description": ""
  },
  "options": {
    "draftAndPublish": true
  },
  "pluginOptions": {
    "i18n": {
      "localized": true
    },
    "preview-button": {
      "listViewColumn": true
    }
  },
  "attributes": {
    "children": {
      "displayName": "children",
      "type": "component",
      "repeatable": false,
      "pluginOptions": {
        "i18n": {
          "localized": true
        }
      },
      "component": "destinations.children"
    }
  }
}
```

and children component like:

```
{
  "collectionName": "components_destinations_children",
  "info": {
    "displayName": "children",
    "description": ""
  },
  "options": {},
  "attributes": {
    "children": {
      "type": "relation",
      "relation": "oneToMany",
      "target": "api::destination.destination"
    }
  }
}
```

- add a lot of data with self-references (I don't know the exact way to cause it, but the error happens when discardDrafts is run in parallel in a transaction and references an id that is being discarded at the same time I think)
- attempt to migrate to Strapi 5 _with `develop` branch_ and and you get an error about transaction
- attempt to migrate to Strapi 5 _with this branch_, and it works

All your data should be migrated properly

### Related issue(s)/PR(s)

Solves https://github.com/strapi/strapi/issues/21607

Might relate to https://github.com/strapi/strapi/issues/21848

DX-1717

### DIFF
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -194,7 +194,9 @@ const migrateUp = async (trx: Knex, db: Database) => {
       });
 
     for await (const batch of getBatchToDiscard({ db, trx, uid: model.uid })) {
-      await async.map(batch, discardDraft, { concurrency: 10 });
+      // NOTE: concurrency had to be disabled to prevent a race condition with self-references
+      // TODO: improve performance in a safe way
+      await async.map(batch, discardDraft, { concurrency: 1 });
     }
   }
 };


Marc-Roig (approved):
a shame that we can not keep the concurrency , but doesnt look like there is any other inmediate solution

Marc-Roig:
Maybe an ipmrovement could be to increase the concurrency if there are not self referencing relations in the schema

pandeyprafull:
I am still running into the issue @hanpaine ...
Screenshot from 2025-02-07 17-15-32

PR #22085, opened 2024-11-06
feat: adding info diagnostics to data-transfer providers

Bassel17:
### What does it do?

- Adds diagnostics to the providers of data-transfer so we can have logs and a log file with information for debugging issues

### Why is it needed?

- To make debugging DTS easier

### DIFF
--- a/packages/core/data-transfer/src/file/providers/destination/index.ts
+++ b/packages/core/data-transfer/src/file/providers/destination/index.ts
@@ -15,6 +15,7 @@ import type {
   ProviderType,
   Stream,
 } from '../../../../types';
+import type { IDiagnosticReporter } from '../../../engine/diagnostic';
 import { createFilePathFactory, createTarEntryStream } from './utils';
 import { ProviderTransferError } from '../../../errors/providers';
 
@@ -61,8 +62,20 @@ class LocalFileDestinationProvider implements IDestinationProvider {
 
   #archive: { stream?: tar.Pack; pipeline?: Stream } = {};
 
+  #diagnostics?: IDiagnosticReporter;
+
   constructor(options: ILocalFileDestinationProviderOptions) {
     this.options = options;
+  }
+
+  #reportInfo(message: string) {
+    this.#diagnostics?.report({
+      details: {
+        createdAt: new Date(),
+        message: `[file-destination-provider] ${message}`,
+      },
+      kind: 'info',
+    });
   }
 
   get #archivePath() {
@@ -91,7 +104,8 @@ class LocalFileDestinationProvider implements IDestinationProvider {
     return zlib.createGzip();
   }
 
-  bootstrap(): void | Promise<void> {
+  bootstrap(diagnostics: IDiagnosticReporter): void | Promise<void> {
+    this.#diagnostics = diagnostics;
     const { compression, encryption } = this.options;
 
     if (encryption.enabled && !encryption.key) {
@@ -153,6 +167,7 @@ class LocalFileDestinationProvider implements IDestinationProvider {
   }
 
   async #writeMetadata(): Promise<void> {
+    this.#reportInfo('writing metadata');
     const metadata = this.#providersMetadata.source;
 
     if (metadata) {
--- a/packages/core/data-transfer/src/file/providers/source/index.ts
+++ b/packages/core/data-transfer/src/file/providers/source/index.ts
@@ -11,6 +11,7 @@ import { parser } from 'stream-json/jsonl/Parser';
 import type { Struct } from '@strapi/types';
 
 import type { IAsset, IMetadata, ISourceProvider, ProviderType, IFile } from '../../../../types';
+import type { IDiagnosticReporter } from '../../../engine/diagnostic';
 
 import * as utils from '../../../utils';
 import { ProviderInitializationError, ProviderTransferError } from '../../../errors/providers';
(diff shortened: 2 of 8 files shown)


packages/core/data-transfer/src/file/providers/destination/index.ts
```diff
@@ -61,10 +63,22 @@ class LocalFileDestinationProvider implements IDestinationProvider {
...
+
   constructor(options: ILocalFileDestinationProviderOptions) {
     this.options = options;
   }
 
+  #reportInfo(message: string) {
+    this.#diagnostics?.report({
+      details: {
+        createdAt: new Date(),
+        message: `${chalk.cyan(`[file-destination-provider]`)} ${message}`,
```
innerdvations, line 76: Instead of embedding the source into the messages, could we add a `source` identifier and have the logger add it in? That way we don't have to manage it in every provider, we can add filters for it, etc

packages/core/data-transfer/src/file/providers/source/index.ts
```diff
@@ -143,7 +163,7 @@ class LocalFileSourceProvider implements ISourceProvider {
     const inStream = this.#getBackupStream();
     const outStream = new PassThrough({ objectMode: true });
     const loadAssetMetadata = this.#loadAssetMetadata.bind(this);
-
+    this.#reportInfo('creating assets read stream');
     pipeline(
```
innerdvations, line 167:
```suggestion
    this.#reportInfo('creating assets read stream');
    
    pipeline(
```

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -32,9 +32,11 @@ const command = () => {
           `URL of the remote Strapi instance to send data to`
         ).argParser(parseURL)
       )
+
       .addOption(
         new Option('--to-token <token>', `Transfer token for the remote Strapi destination`)
       )
+      .addOption(new Option('-i, --info', 'log information about the export process'))
```
innerdvations, line 39: Shouldn't this be `--log-level` or `verbosity` to choose the level? That way warnings could be disabled if desired, and we can add a "debug" level later that includes even more detailed information

packages/utils/logger/src/formats/detaliled-log.ts
innerdvations, line 1: typo in filename

packages/core/strapi/src/cli/utils/data-transfer.ts
```diff
@@ -210,14 +210,18 @@ const errorColors = {
 } as const;
 
 const formatDiagnostic = (
-  operation: string
+  operation: string,
+  info?: boolean
```
innerdvations, line 214: Same comment as CLI command, should this accept an actual level rather than a boolean?

innerdvations (changes requested):

packages/core/strapi/src/cli/commands/transfer/command.ts
butcherZ: Bassel and I had a discussion on that. We decided to make info a boolean because we think that, either way, the warnings and errors will be displayed by default. So we are only determining if info will be displayed here or not. What do you think?

innerdvations: Yeah, less so for warn/error, but I'm afraid we'll want to add 'debug' level soon that logs -everything- and we will have limited ourselves and will be forced to add another option like --debug

Bassel17: yeah I have been thinking about this, in terms of changing it to `--verbose` or something nothing prevents us from changing things later if we need to, or even just adding `--debug` option, which is also pretty standard.
but adding log levels now when I know there is only one level and probably will be the case for a very long while I think is unnecessary and will only cause confusion

packages/core/strapi/src/cli/utils/data-transfer.ts
Bassel17: I still think it should be a boolean, just `--verbose` rather than info

packages/core/strapi/src/cli/commands/transfer/command.ts
maccomaccomaccomacco: in the future we could have multiple level of logs, though

so, even if it's just one level for now, probably we should already support this scenario

innerdvations (approved):
This all lgtm and works well, as long as you've worked out the option name with @maccomaccomaccomacco I'm happy :)

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
```diff
@@ -348,7 +351,20 @@ class RemoteStrapiSourceProvider implements ISourceProvider {
     return res.transferID;
   }
 
-  async bootstrap(): Promise<void> {
+  #reportInfo(message: string) {
+    console.log(this.#diagnostics);
```
innerdvations, line 355: Is this intentional?

innerdvations (changes requested):
Works for me! Only change is the console.log that seems like it might be a mistake

(update: and -v issue, see below)

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
Bassel17: shit nooo

innerdvations (approved):
LGTM!

Issue #21546, opened 2024-10-01
Media information not populated in webhook events for content


cekpowell:
### Required System information

- Node.js version: v20.16.0
- NPM version: 10.8.1
- Strapi version: 5.0.0
- Database: sqlite
- Operating system: MacOS (darwin-arm64)
- Is your project Javascript or Typescript: Typescript

### Describe the bug

- Media information/properties are not populated inside webhook events for content (e.g., when creating/updating content).
- In Strapi V4, for media fields, we used to receive the full record for the media within the webhook event when the content is was attached to was created/updated. However, in Strapi V5, we only recieve the media's Asset ID and nothing else.
- This screenshot shows what we receieve in webhook events for content with media (in this case a create event). As you can see, we only receive the media ID:

<img width="400" alt="Screenshot 2024-10-01 at 14 09 03" src="https://github.com/user-attachments/assets/10e474e0-ab87-4702-bc32-e7ef1fc5b664">

- Interestingly, the media information is populated for `entry.delete` events:

<img width="400" alt="Screenshot 2024-10-01 at 14 10 33" src="https://github.com/user-attachments/assets/e3909831-472d-4203-8836-5bd70f02c443">

We use S3 as our upload provider, and want to perform some actions on the media in response to it being attached to content. However, as the media is stored using the filename, unless this information is in the webhook, its not possible to know which piece of media we need to work with.

### Steps to reproduce the behavior

- Setup a new webhook for content create/update/delete
- Create a content type with a media field
- Create a new piece of content with a piece of media attached to it
- Inspect the payload on the webhook event - there will be no informaiton on the media other than the media ID.

### Expected behavior

- The full set of media information should be populated on all content webhook events, as can be seen in this example:

<img width="400" alt="Screenshot 2024-10-01 at 14 10 33" src="https://github.com/user-attachments/assets/e3909831-472d-4203-8836-5bd70f02c443">

derrickmehaffy:
related to: https://docs.strapi.io/dev-docs/migration/v4-to-v5/breaking-changes/remove-webhook-populate-relations

Need to discuss with engineering what was intended here

cekpowell:
Thanks - Yup relates to that change. Weirdly enough for non-media relations (i.e., other content), the data is properly populated, but it seems like thats now broken the population for media relations.

In case anyone else is facing this issue, as a work around, I have defined by 'own' webhooks within lifecycle hooks for the DB  👍

PR #22086, opened 2024-11-06
fix: populate media in document service

jhoward1994:
### What does it do?

Change Media population logic in the document service so webhooks receive full media info

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/21546

### DIFF
--- a/packages/core/core/src/services/document-service/utils/populate.ts
+++ b/packages/core/core/src/services/document-service/utils/populate.ts
@@ -36,6 +36,8 @@ export const getDeepPopulate = (uid: UID.Schema, opts: Options = {}) => {
       }
 
       case 'media': {
+        // We populate all media fields for completeness of webhook responses
+        // see https://github.com/strapi/strapi/issues/21546
         acc[attributeName] = { select: ['*'] };
         break;
       }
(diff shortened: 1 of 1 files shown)


jhoward1994:
Alternative idea

```
selectAllMedia?: boolean;
...
acc[attributeName] = { select: opts.selectAllMedia ? ['*'] : ['id'] };
```

Marc-Roig (dismissed):
lets go with this for now as it looks simpler :) ! nice catch

packages/core/core/src/services/document-service/utils/populate.ts
```diff
@@ -36,7 +36,7 @@ export const getDeepPopulate = (uid: UID.Schema, opts: Options = {}) => {
       }
 
       case 'media': {
-        acc[attributeName] = { select: ['id'] };
+        acc[attributeName] = { select: ['*'] };
```
Marc-Roig, line 39: could you add a note thats important to populate all data so the webhooks contain that informatoin? maybe linking the issue

jhoward1994:
Thanks, will do!

innerdvations (dismissed):
LGTM!

innerdvations (approved):

PR #22096, opened 2024-11-07
test(e2e): change playwright timeouts

innerdvations:
### What does it do?

- increase overall test timeout
- decrease action and expect timeouts
- adds missing outputs to skipped_tests.yml (update: nope, this won't change anything, but for now I'll leave it to keep it in sync with test.yml)
 
### Why is it needed?

- to reduce incidence of flaky tests while also allowing them to fail faster
- I'm hoping that fixing the outputs also fixes handling of PRs that don't trigger any tests but don't have evidence yet; if this doesn't work then I'll follow up with something else

There are currently several tests that fail because they take too long. Efforts like [this](https://github.com/strapi/strapi/pull/22079) improve them, and we should continue to write shorter and more performant tests, but right now just having a couple of servers restarts is enough to make a test flaky because the CI is so slow

### How to test it?

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22079

### DIFF
--- a/playwright.base.config.js
+++ b/playwright.base.config.js
@@ -37,15 +37,15 @@ const getEnvBool = (envVar, defaultValue) => {
 const createConfig = ({ port, testDir, appDir }) => ({
   testDir,
 
-  /* default timeout for a jest test to 30s */
-  timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 30 * 1000),
+  /* default timeout for a jest test */
+  timeout: getEnvNum(process.env.PLAYWRIGHT_TIMEOUT, 90 * 1000),
 
   expect: {
     /**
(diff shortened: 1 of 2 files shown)


playwright.base.config.js
markkaylor, line 1: I'm cool with this but is it really the solution to the problem? Won't this potentially increase the amount of time tests can run in the CI?

markkaylor (approved):

playwright.base.config.js
innerdvations: That is, I think at 60-90 seconds we can safely insitute a policy that any flaky test MUST be rewritten, at 30 we're already running into roadblocks that make writing tests harder

markkaylor: Cool works for me and they're all green 🚀

Issue #22097, opened 2024-11-07
Media Upload UX problems


ar53n:
### Required System information

- Node.js version: v20.12.1
- NPM version: 10.5.0
- Strapi version: 5.2.0
- Database: sqlite
- Operating system: MacOS 12.3.1
- Is your project Javascript or Typescript: Typescript

### Describe the bug

Bug about UX in media field. We can sort media files but cursor don't show this action. But i think this component should be reworked for better UX/UI

### Steps to reproduce the behavior

1. Create "media" field
2. Go to Content Media and add media files
3. Select "Selected Files" tab
4. Upload audio and some images
5. Move files for needed order

### Expected behavior

* We should see correct cursor and draggable effect on item. 
* Also we should see checkbox for audio

### Screenshots

<img width="1280" alt="image" src="https://github.com/user-attachments/assets/114a2591-de93-4b50-8c09-b9206b6df831">

https://github.com/user-attachments/assets/b4ea7411-5b52-44b0-b1a7-95fbd4bf4644

### Additional context

It's just UX problem with upload and assets management

ar53n:
https://github.com/strapi/strapi/issues/22002#issuecomment-2443953545

@frontBOI I have fixed some UI problems

ar53n:
#22098

lucasboilly:
Hi @ar53n, thanks! I just gave my feedback in the PR :)

Adzouz:
I just merged @ar53n PR, thanks a lot 🚀
I'm closing the issue!

PR #22098, opened 2024-11-07
fix(upload): draggable element style

ar53n:
### What does it do?

I have fixed some issues with upload media files for example:
* Checkbox z-index with audio file
* Edit button with audio file
* Draggable media file style

### Why is it needed?

The main problem is that it is not obvious that you can change the order of files in the field

### How to test it?

1. Create media field with multiple files
2. Upload files
3. Select "Selected files"
4. Move files and change order.
5. Check action buttons position

Fixed issue
https://github.com/user-attachments/assets/826c887a-8a5e-4b94-a599-a4b72b40ba80

Expected behavior in browse tab
https://github.com/user-attachments/assets/55d7b3c9-ae04-4ffe-ad3c-3dff22a8709c

### Related issue(s)/PR(s)

#22097

### DIFF
--- a/packages/core/upload/admin/src/components/AssetCard/AssetCard.tsx
+++ b/packages/core/upload/admin/src/components/AssetCard/AssetCard.tsx
@@ -22,6 +22,7 @@ interface AssetCardProps {
   size?: 'S' | 'M';
   allowedTypes?: AllowedTypes[];
   alt?: string;
+  className?: string;
 }
 
 export const AssetCard = ({
@@ -32,6 +33,7 @@ export const AssetCard = ({
   onRemove,
   size = 'M',
   local = false,
+  className,
 }: AssetCardProps) => {
   const handleSelect = onSelect ? () => onSelect(asset) : undefined;
 
@@ -47,6 +49,7 @@ export const AssetCard = ({
     onRemove: onRemove ? () => onRemove(asset) : undefined,
     selected: isSelected,
     size,
+    className,
   };
 
   if (asset.mime?.includes(AssetType.Video)) {
--- a/packages/core/upload/admin/src/components/AssetCard/AssetCardBase.tsx
+++ b/packages/core/upload/admin/src/components/AssetCard/AssetCardBase.tsx
@@ -63,6 +63,7 @@ export interface AssetCardBaseProps {
   selected?: boolean;
   subtitle?: string;
   variant: 'Image' | 'Video' | 'Audio' | 'Doc';
+  className?: string;
 }
 
 export const AssetCardBase = ({
@@ -76,6 +77,7 @@ export const AssetCardBase = ({
   selected = false,
   subtitle = '',
   variant = 'Image',
+  className,
 }: AssetCardBaseProps) => {
   const { formatMessage } = useIntl();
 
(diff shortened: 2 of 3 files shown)


hanpaine:
Hi @ar53n thanks for the fix, we'll take a look soon 🚀

lucasboilly:
Hello @ar53n, thanks for this PR :)

There's a few things I would like to change from a design point of view:
- Could we change the cursor to `grab` when hovering a card?
CleanShot 2024-11-12 at 13 46 57@2x
- Could we change the cursor to `grabbing` when moving a card?
CleanShot 2024-11-12 at 13 48 00@2x
- I don't think we should make the cards grow on hover. That's not a bad idea at all but I'd prefer working on a global hover state for all cards in Strapi instead of using this right now.

Regarding the audio file, that's great but I think it would be even better to center it vertically because the edit button and the checkbox (in the Media Library) are hiding parts of it.

ar53n:
@lucasboilly thanks for review. 
* About audio file I wanted to make a small edit because I saw an issue about redesigning the card preview.

* Don't you think the cursors `grab` `grabbing` `pointer` are look similar? And the move cursor is more suitable in this situation? I think we can remove hover effect but save `move` сursor.
* I added hover effect because the same cards in other upload parts are not draggable. And this can be very confusing. What do you think?

yanniskadiri:
This PR would fix this issue as well: #20903

FYI: I just referred it in the first comment. Thanks for working on this issue @ar53n.

lucasboilly:
Hi, @ar53n thanks for your answers :)
- As @yanniskadiri said it will be fixed in another PR, no need to worry about this.
- They do, but I think it's more natural (so more user friendly) to show a hand that literally grabs the elements to move them somewhere else. It's also a more modern approach, for example Notion and Figma are also doing it (see recording of Notion's drag and drop interaction).
- Do you know which cases a card is not draggable? Also the cursor change should be enough for now to show that a card is draggable. But as I said I'll work on an iteration that we could generalize to the whole admin later.

https://github.com/user-attachments/assets/04751d8b-fa23-4b62-9337-693cdfab2f0f

ar53n:
@lucasboilly Thanks for you point. Okey, we can change cursor, but look at upload modal in "browse" tab. There you can't move the cards but the cursor type is `pointer` and it very similiar. I just find it a little confusing. If you think it's good, I'll fix it.

lucasboilly:
@ar53n to me it's another issue linked to the modal's design itself, but it's a good feedback. We could indeed find a way to make it clearer than one tab is for selection and the other is for ordering. It would require extra work and research from the product team but I'm gonna tell them about it :)

In the meantime we can just apply why we discussed together, it's already a good improvement compared to what we have today!

ar53n:
@lucasboilly i agree with you. I will do changes.
For upload design i can suggest you to see on other CMS implementations. For more information i can write you in discord

lucasboilly:
@ar53n sure! Are you part of Strapi's Discord server?

ar53n:
@lucasboilly yeah! You can write me. My nickname `arskdev`. I couldn't find you

lucasboilly:
@ar53n thanks! I just sent you a message :)

lucasboilly:
@ar53n I just checked the last version and there's a last issue. The `grabbing` pointer only appears while clicking, but after moving  the card it becomes an arrow again and then `grab` when it has exchanged is place with another card. The `grabbing` cursor should stay until the user removes his finger from his mouse/trackpad.

https://github.com/user-attachments/assets/f8fdd92b-491c-4486-87c8-510e7a482e94

ar53n:
@lucasboilly yeah, i will try, but i think it's react-dnd problem. And we gonna do more changes.
You can look examples https://react-dnd.github.io/react-dnd/examples/dustbin/single-target. In example have the same problem

lucasboilly:
@ar53n okay, let's do this 👍

Adzouz:
Thanks for your contribution @ar53n, merged 🚀

PR #22103, opened 2024-11-07
 chore: make ctb tests no longer rely on test order

remidej:
### What does it do?

Updates the code in the CTB e2e tests so that it can work regardless of test order.

Our tests often rely on retries (especially on webkit), and on retries, [the execution order is no longer respected](https://playwright.dev/docs/test-retries#failures), because Playwright will only resume at the specific test that failed, the previous ones won't re-run.

For context, the PR that introduced the logic that relies on retries is #21464. I understand why, as it makes tests a lot faster. But I believe our priority should be the flakiness right now, and the extended timeouts should allow us to have slow tests that don't fail. Then I think when the tests pass reliably we can discuss the different ways to speed them up. And a test that passes means a faster CI, because fewer retries are triggered

Another possible approach would be to check if we're in a retry and only adapt the logic in that scenario, so the perf wouldn't be impacted when all goes well. But that sounds more complex to me.

Happy to discuss it!

### DIFF
--- a/tests/e2e/tests/content-type-builder/collection-type/edit-collection-type.spec.ts
+++ b/tests/e2e/tests/content-type-builder/collection-type/edit-collection-type.spec.ts
@@ -14,27 +14,27 @@ test.describe('Edit collection type', () => {
   const attributes = [{ type: 'text', name: 'testtext' }];
 
   test.beforeEach(async ({ page }) => {
+    await resetFiles();
     await sharedSetup('ctb-edit-ct', page, {
-      resetFiles: true,
       importData: 'with-admin.tar',
       login: true,
       skipTour: true,
-      afterSetup: async () => {
-        // create a collection type to be used
-        await createCollectionType(page, {
-          name: ctName,
-          attributes,
-        });
+      // Don't reset files here as it would only run once for the whole suite
+      resetFiles: false,
+    });
 
-        await createCollectionType(page, {
-          name: 'dog',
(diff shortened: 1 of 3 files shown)


innerdvations:
As long as they still pass, this is fine with me.

However, I was seeing it less that tests -depend- on previous tests, and more that tests which use this method need to be aware that the side effects of previous tests will not reset. Since we're keeping the functionality itself in the utils, maybe improve the comment there to make that more clear at the same time?

remidej:
@innerdvations yeah I think `sharedSetup` itself is a neat tool, we just need to make sure we use it in a safe way.

So you're right I'll update to remove that recommendation:

https://github.com/strapi/strapi/blob/9e6d64696bde7a68396e3e0d567347b610fb5232/tests/e2e/utils/setup.ts#L38-L41

innerdvations:
> @innerdvations yeah I think `sharedSetup` itself is a neat tool, we just need to make sure we use it in a safe way.
> 
> So you're right I'll update to remove that recommendation:
> 
> https://github.com/strapi/strapi/blob/9e6d64696bde7a68396e3e0d567347b610fb5232/tests/e2e/utils/setup.ts#L38-L41

Yeah, unfortunately I think there are several e2e tests that depend on previous steps :(  But at least there are fewer of them in e2e than API (where I run into it on a regular basis)

markkaylor (approved):
lgtm

PR #22112, opened 2024-11-08
fix: cleanup data after deleting a component

### LINKED ISSUE #20931: Deleting or renaming a component does not delete/update the relational linking data for the component

innerdvations:
### What does it do?

Cleans up the data in the database properly after deleting a component

TODO in follow-up PRs:

- [ ] add a migration and/or database repair script to remove this data from databases that are already affected
- [ ] [Graceful renaming/deletion of everything](https://feedback.strapi.io/developer-experience/p/gracefully-handle-renaming-of-content-types-and-fields-in-the-ctb)

### Why is it needed?

We aren't cleaning up all the data in the database after deleting a component, which also can cause issues when components with the same name are recreated

### How to test it?

Tests are included; see linked issue to test manually

### Related issue(s)/PR(s)

Mitigates #20931 for v5 -- full solution for affected production systems will be much more complex

DX-1726


packages/core/content-type-builder/server/src/services/components.ts
```diff
@@ -86,13 +86,38 @@ export const editComponent = async (
...
+        (attr: any) => attr.target === deleteUid && attr.joinTable?.name
+      );
+
+      // Delete entries in each join table associated with matching attributes
+      for (const attr of matchingAttributes) {
+        if (!('joinTable' in attr && attr.joinTable && attr.joinTable.name)) {
+          continue;
+        }
+
+        await trx(attr.joinTable.name).where('component_type', deleteUid).delete();
```
innerdvations, line 109: Any opinions on if this should be allowed to throw, or if this should be treated as a non-critical part of the delete and just warn? I already have a stashed change that just logs a warning that the data couldn't be deleted if there is a problem, which would make this change a lot safer in the case that we've missed any edge cases where we target the wrong table for example.

alexandrebodin: I think this will rarely be usefull. the CTB is used locally so most likely not connected to a production database and as such will not change anything when the deployment happen. Considering this I would reco a warning

innerdvations: True, this will just be preventing immediate errors. The migration or repair will be needed to fix it in production.

jhoward1994:
```
  if (failedDeletes.length > 0) {
    strapi.log.warn(
      `Failed to delete component data for ${deleteUid} from ${failedDeletes.map((attr) => attr.table)}`
    );
  }
```

Should we make this warning message clearer that the component itself has been deleted but some references could be left in the DB. I'm just thinking that the way it's currently written could be ambiguous.

e.g. `Component '${deleteUid}' was deleted, but references remain in the following database tables: ${failedDeletes.map((attr) => attr.table)}. Consider cleaning up these tables to avoid orphaned data.`

butcherZ (approved):
QAed with `v0.0.0-experimental.702fcf7aec9be82c4485a8bd49e235293c37f022`

- created a component, added the component to a content type, created test data with the content type, delete the component
1.  test data is gone ✅
2. relations in database is gone ✅
3. content type is still available ✅
4. able to re-create component with the same name ✅

I have also tested the edge case mentioned in DX-1726
1. prod server starts up  ✅
2. no errors in CTB  ✅
3. no errors in component editor  ✅
4. no problems adding / editing those components  ✅

considering the edge case is a bit tricky, better to have someone to confirm with me on it :D

packages/core/content-type-builder/server/src/services/components.ts
```diff
@@ -86,13 +86,60 @@ export const editComponent = async (
   return updatedComponent;
 };
 
-export const deleteComponent = async (uid: Internal.UID.Component) => {
-  const builder = createBuilder();
+export const deleteComponent = async (deleteUid: Internal.UID.Component) => {
+  const failedDeletes: any[] = [];
```
Convly, line 90: Can we have a real type here pls :)

packages/core/content-type-builder/server/src/services/components.ts
```diff
@@ -86,13 +86,60 @@ export const editComponent = async (
...
+    const builder = createBuilder();
+
+    const models = [...builder.contentTypes.entries(), ...builder.components.entries()];
+
+    // Find models that include an attribute with this component as a target and delete references to it from the db
+    for (const [modelUid] of models) {
+      const metadata = strapi.db.metadata.get(modelUid);
+
+      const matchingAttributes = Object.values(metadata.attributes).filter(
+        (attr: any) => attr.target === deleteUid && attr.joinTable?.name
```
Convly, line 102: Just for good measure, should we also check that it is a relation? (also, don't we have an attribute type that we can use that is exported from the DB? If not, not a big deal)

innerdvations:
Closing in favor of https://github.com/strapi/strapi/pull/22267

Issue #21838, opened 2024-10-14
Color Picker does not allow you to manually paste in hex code


derrickmehaffy:
### Required System information

- Node.js version: v20
- NPM version: Yarn v1
- Strapi version: 5.0.4
- Database: SQLite
- Operating system: Linux Mint 21.2
- Is your project Javascript or Typescript: Javascript

### Describe the bug

The color picker custom field in Strapi 5 does not allow editing or pasting a hex code in the text box (this worked on Strapi 4)

### Steps to reproduce the behavior

1. Install the color picker plugin
2. Create a content-type with the color picker field
3. Create a new entry and select a color
4. Save the entry
5. Edit the entry and try to paste/edit the hex code with something else (without picking the color)
6. See that you can't edit it

### Expected behavior

Should be able to edit the hex code (same as v4)

### Screenshots

N/A

### Code snippets

N/A

### Additional context

N/A

ksb86:
Also having this same issue.
Node 20.16.0
@strapi/strapi 5.4.0
@strapi/plugin-color-picker 5.4.0

Jhonaiker2309:
Hey, could you please assign me this issue?

ksb86:
Issue still persists

Node v20.18.1
@strapi/strapi 5.5.0
@strapi/plugin-color-picker 5.5.0

samhenderson511:
Also having this issue

Node v22.11.0
@strapi/strapi 5.5.0
@strapi/plugin-color-picker 5.5.0

PR #22122, opened 2024-11-10
fix: unable to type color value using keyboard

KamilPieronczyk:
### What does it do?

I updated the onChange function call in the color picker plugin to pass the name as the argument

### Why is it needed?

User was unable to type the hex color value using keyboard

### How to test it?

Assign a color picker component to some single type component and try to edit the color value using keyboard

### Related issue(s)/PR(s)

fixes: https://github.com/strapi/strapi/issues/21838


OssprO (commented):
Everything looks right!

simotae14:
Hello @KamilPieronczyk  ,
Thanks a lot for your contribution. We have just merged another fix for the issue, so I will close your Pr in favor of the one merged.
But thanks again for your help 🙏🏽

PR #22133, opened 2024-11-12
enhancement: add didRestoreHistoryVersion and willNavigate to history

simotae14:
### What does it do?

We added the track for:
 - the number of times users go to the Content History page: we’ll need to implement the willNavigate event for that
 - the number of times users restore a history version: we’ll need to implement the didRestoreHistoryVersion event

### Why is it needed?

To have some metrics regarding the use of the Content History feature

### How to test it?

- add some entries to a Collection and try to modify them and save
- click on the "Content History" button in the Edit view 
- check the Network tab in the DevTool, a new POST track call will be triggered with the event "willNavigate"
- now in the History page try to select an old version and click on the Restore button (it needs to be enabled)
- in the Confirmation dialog click Restore
- check the Network tab in the DevTool, a new POST track call will be triggered with the event "didRestoreHistoryVersion"

### Related issue(s)/PR(s)

CS-1206

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/HistoryAction.tsx
+++ b/packages/core/content-manager/admin/src/history/components/HistoryAction.tsx
@@ -1,8 +1,8 @@
-import { useQueryParams } from '@strapi/admin/strapi-admin';
+import { useQueryParams, useTracking } from '@strapi/admin/strapi-admin';
 import { ClockCounterClockwise } from '@strapi/icons';
 import { stringify } from 'qs';
 import { useIntl } from 'react-intl';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useLocation } from 'react-router-dom';
 
 import type { DocumentActionComponent } from '../../content-manager';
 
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/history/components/HistoryAction.tsx
```diff
@@ -1,28 +1,39 @@
...
 
   if (!window.strapi.features.isEnabled('cms-content-history')) {
     return null;
   }
 
+  const handleOnClick = () => {
+    const destination = { pathname: 'history', search: pluginsQueryParams };
+    trackUsage('willNavigate', {
+      from: pathname,
+      to: `${pathname}/${destination.pathname}?${destination.search}`,
```
Marc-Roig, line 25: just wondering (cause I might do the same in another pr) do we normally track the search params?

simotae14: Hello @Marc-Roig thanks a lot for your question...I discussed the topic with @yanniskadiri and we tried to check if there are other "willNavigate" where we pass the search params, there is an example when you click on the API tokens in the Setting navbar
Screenshot 2024-11-12 alle 16 11 30
That said we think there is no real value passing the search param in the case of history, because we want to measure how many people click on the button so we decided to remove the search params part...I hope it can be useful in your case too

Marc-Roig: noice!! thank you for sharing, I will keep the search params out of the preview event too! ty

PR #22138, opened 2024-11-12
feat: preview allowed origins config

Marc-Roig:
### What does it do?

Introduces a new `allowedOrigins` preview config, so all the necessary configuration for Preview is in the same place.

Until now, it was required to set up the preview URL in the `config/middlewares.js` security middleware. It was tedious and just another step that made preview config more complicated. 

Now you can do the same by configuring it in `config/admin.js`:

```ts
  preview: {
    config: {
      allowedOrigins: ['docs.strapi.io'],
        if (status === 'published') {
        }

      },
    },
  },
``` 

There is no need to update the `config/middlewares.js` file anymore.

### DIFF
--- a/examples/getstarted/config/admin.js
+++ b/examples/getstarted/config/admin.js
@@ -22,7 +22,7 @@ module.exports = ({ env }) => ({
     enabled: env.bool('PREVIEW_ENABLED', true),
     config: {
       handler: (uid, { documentId, locale, status }) => {
-        return `/preview/${uid}/${documentId}/${locale}/${status}`;
+        return `/admin/preview/${uid}/${documentId}/${locale}/${status}`;
       },
     },
   },
--- a/packages/core/content-manager/server/src/preview/index.ts
+++ b/packages/core/content-manager/server/src/preview/index.ts
@@ -21,13 +21,12 @@ const getFeature = (): Partial<Plugin.LoadedPlugin> => {
   // }
 
   return {
-    bootstrap() {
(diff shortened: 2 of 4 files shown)


remidej (approved):
🚀

examples/getstarted/config/admin.js
```diff
@@ -22,7 +22,7 @@ module.exports = ({ env }) => ({
     enabled: env.bool('PREVIEW_ENABLED', true),
```
remidej, line 22: Unrelated but I'm noticing this exists in addition to the feature check. We'll need to clarify how we want to do these things

remidej (commented):
I'm actually not managing to get this to work. I'm getting to trying to load youtube embeds:

```
config: {
      allowedOrigins: ['youtube.com'],
        return 'https://youtube.com/embed/GMgsFZ4rkEI';
      },
    },
```

> Refused to frame 'https://youtube.com/' because it violates the following Content Security Policy directive: "frame-src 'self'".

I think it may work with the Strapi Docs because they're on the same domain

Marc-Roig:
@remidej for some reason I have to use `https://youtube.com *.youtube.com` to make it work with youtube, I will investigate tomorrow

remidej:
I can't get it to work with it myself with those too, that's weird

Marc-Roig:
@remidej can you try now? there was an issue with the middleware extension logic.

remidej (approved):
🚀🚀🚀!

Issue #22026, opened 2024-10-30
Enum difference v4/v5 - "Choose here" as null value


alex-strapi:
### Required System information

- Node.js version: 20
- Yarn version: 4.3.1
- Strapi version: 5.0.6 - 5.1.1
- Database: N/A
- Operating system: Mac
- Is your project Javascript or Typescript: N/A

### Describe the bug

In Strapi v4 a user could create an enum - a content editor could select values but could also select "Choose here" as a an empty string value.

![image](https://github.com/user-attachments/assets/7646f8ec-b9b7-43c2-a2ea-97fb5caeaca1)

In Strapi v5 a user can create an enum - a content editor cannot select anything other than a value, is this expected? 
Should users have to create an enum with their own empty string value option?

![image](https://github.com/user-attachments/assets/9059722e-ef74-422b-8d3c-1df056996c72)

### Steps to reproduce the behaviour

1. Create an enum on Strapi v4 - check content editor experience (choose here option exists)
2. Create an enum on Strapi v5 - check content editor experience (no choose here option)

### Expected behaviour

If we're basing expected behaviour on what v4 had, we should have a 'Choose here' option. If it's expected in v5 that a empty string value option needs to be added by the user we should update the documentation to reflect this. 

https://docs.strapi.io/user-docs/content-type-builder/configuring-fields-content-type#-enumeration

Edit: Changed null to empty string

alex-strapi:
> just an annotation, in v4 the value passed when we selected the label "Choose here" was an empty string and not a null value, is it ok for you replicate the old behaviour?

Updated the issue to reflect this.

simotae14:
I am gonna close the issue because it is fixed with this PR https://github.com/strapi/strapi/pull/22139 thanks

PR #22139, opened 2024-11-12
fix: Enumeration add "Choose here" as empty value

simotae14:
### What does it do?

It adds again in the Enumeration field the "Choose here" as the empty option, to reflect the behavior we had in v4 

### Why is it needed?

Because otherwise there is no possibility to set to empty an enumeration field when filled. In the case the field is required the "Choose here" option is not selectable like it was in v4

### How to test it?

- Create an enum on Strapi v5 
-  set the field as not required
- then create a new instance of the enum field and try to select an option
- Save
- then edit the entry and choose the "Choose here" option to save the empty value and save

test the required field case
- go to a collection and add an enum field required
- then create a new instance of the Collection and try to save the enum empty
- then try to publish the entry (an error message needs to show up for the enumeration field)
- then select an option in the enumeration and save the instance
- try to select the "Choose here" option (you can't do it)

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/22026

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Enumeration.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Enumeration.tsx
@@ -1,6 +1,7 @@
 import { forwardRef, memo } from 'react';
 
 import { SingleSelect, SingleSelectOption, useComposedRefs, Field } from '@strapi/design-system';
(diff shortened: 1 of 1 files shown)


remidej (approved):
Works well 👍

PR #22141, opened 2024-11-13
fix 22026 : Enum difference v4/v5 - "Choose here" as null value

### LINKED ISSUE #22026: Enum difference v4/v5 - "Choose here" as null value

archanaagivale30:
### What does it do?

Fix #22026 : Enum difference v4/v5 - "Choose here" as null value

### Why is it needed?

In Strapi v4 a user could create an enum - a content editor could select values but could also select "Choose here" as a null value.

### How to test it?

Create an enum on Strapi v5 - check content editor experience (no choose here option)

### Related issue(s)/PR(s)

#22026

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Enumeration.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Enumeration.tsx
@@ -11,7 +11,7 @@ const EnumerationInput = forwardRef<HTMLDivElement, EnumerationProps>(
   ({ name, required, label, hint, labelAction, options = [], ...props }, ref) => {
     const field = useField(name);
     const fieldRef = useFocusInputField<HTMLDivElement>(name);
-
+    options = [{ label: 'Choose here', value: '' }, ...options];
     const composedRefs = useComposedRefs(ref, fieldRef);
 
     return (
@@ -20,7 +20,7 @@ const EnumerationInput = forwardRef<HTMLDivElement, EnumerationProps>(
         <SingleSelect
           ref={composedRefs}
           onChange={(value) => {
-            field.onChange(name, value);
+            field.onChange(name, value || null);
           }}
           value={field.value}
           {...props}


simotae14:
Hello @archanaagivale30 
Thanks again for your contribution 🚀 ... 
we currently have an open PR ready for review that fixes the bug.
I'll let you know if it gets merged or not
I will let you know if it will be merged or not

Thanks

simotae14:
Hey @archanaagivale30 thanks again for the contribution, unfortunately, I need to close your fix because we already merged a bug fix https://github.com/strapi/strapi/pull/22139 thanks again

archanaagivale30:
Hi @simotae14 

I just checked  the merged PR  seeing below two issues which I have encountered while resolving the issue yesterday:

1) When the enum field is required, it doesn't allow me to reselect the "Choose here" option.

https://github.com/user-attachments/assets/2d7d3ace-5697-4dbc-a718-58b68fa6699e

2) When I try to save with the "Choose here" option selected, it gives a validation error.
Screenshot 2024-11-13 at 5 33 51 PM

simotae14:
Hello, @archanaagivale30 I replicated the v4 behavior where if the field is requested you cannot select the Choose here option and if you save an empty value (the initial one) and try to publish you have the validation error, so to me what happens in your two videos is correct

archanaagivale30:
hey @simotae14 Thank you for confirming !

PR #22147, opened 2024-11-13
test(unit): sort order of type members before checking

innerdvations:
### What does it do?

Sorts types members so that tests can't fail in different environments

While the order of attributes in types shouldn't matter, I do wonder if we should do this, or instead try to pin down exactly what slight difference in environment is causing this

### Why is it needed?

when I run unit tests locally, they fail because the typescript generation tests depend on the order that certain members of the types are generated in

### How to test it?

run `test:unit` locally and on the CI and they should pass

### Related issue(s)/PR(s)

I have the same problem on v5, so I will open another PR for that


innerdvations:
> The attributes' modifiers should always be generated in the same order. The fact they don't in some instance is kinda worrying. I'll investigate this further this week and come back to you 🙂

If you don't have the same problem locally, we can debug on my system later

Convly:
It's been happening on my side too. Let's bump the priority on this and issue a fix asap to avoid blocking the releases.

innerdvations:
Ok. Shall we merge this to unblock?  The other tests are still in place so that if anything that would actually be a breaking change happens, we will be alerted to it.

Convly:
Before merging this, I want to make sure we're fixing the issue and not adapting the tests to silence a problem in our code.

Maybe the first step would be to understand why it might generate the structures with different orders ? Does it impact the generated types ? Does it break the type system implementation when trying to -for example- use the document service in a TS project with generated types ?

innerdvations:
Just a stab in the dark, we'll see if this sheds any light on the problem: https://github.com/strapi/strapi/pull/22233

Convly:
Tried something in https://github.com/strapi/strapi/pull/22246

PR #22150, opened 2024-11-13
feat: add missing Ukrainian translations for various plugins and comp…

IMalaniak:
### What does it do?

This PR adds missing Ukrainian translations and fixes some existing translations.

### Why is it needed?

Because it was missing or not correct.

### How to test it?

Switch locale to Ukrainian.


IMalaniak:
Is there something I am missing or maybe I can do something to speed up merging this request?

Bassel17 (approved):
LGTM!
thanks for the added translations

PR #22153, opened 2024-11-14
test(front): run front-end tests with runInBand

innerdvations:
### What does it do?

use `--run-in-band` to run test:front

### Why is it needed?

front-end tests quite consistently fail without it, probably due to race conditions

it's essentially a requirement for running front end test, CI was already using it, this just makes it so it's also used locally

### How to test it?

try running front-end unit tests locally and they should pass

without this PR it's very likely that some of them fail

### Related issue(s)/PR(s)

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -8,6 +8,7 @@ backend:
   - 'packages/core/*/{lib,bin,ee,src}/**'
   - 'packages/core/database/**'
   - 'yarn.lock'
+  - 'package.json'
 frontend:
   - '.github/actions/yarn-nm-install/*.yml'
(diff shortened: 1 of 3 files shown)


markkaylor (approved):

innerdvations:
Since front-end aren't running here, I'm worried this could break the CI runs because it will have `--run-in-band` listed twice 😆

I'll try removing that and see that everything still runs fine

innerdvations:
CI didn't like receiving both `--runInBand` and `--run-in-band`, nor did it like it only being included in the package.json script, but apparently having `--runInBand` included twice is fine

remidej (approved):
nice!

PR #22161, opened 2024-11-14
ci: fix docs workflow jobs

innerdvations:
### What does it do?

- removes the `skipped_tests.yml` which duplicated+mimicked the real workflow and instead rely on conditionals in `tests.yml`
- fixes the issue that docs-only workflows do not ever complete the aggregate
- now tests docs build process instead of docs-only changes always passing
- unify test conditionals

TODO in a followup PR:

- combine install+build and possibly conditionals into a single action (see https://github.com/strapi/strapi/pull/21783 ) which should reduce the amount of code needed
- find a way to split this up into multiple files without complicating it too much; the issue is that workflows can only communicate through artifacts, so the amount of setup and output necessary in each file might make it not worth it unless they can somehow rely on something like a single action to handle everything

### Why is it needed?

Currently without this change:

- We had to set up a duplicate 'aggregate' in the docs-only workflow, and in order for it to still show "skipped" in github we had to duplicate all existing jobs
- because of that, it regularly got out-of-sync or broken
- When we have a docs-only PR, aggregate doesn't run and tests never pass/fail, they just stall
- When we have PRs that mix docs changes and code changes, there is ambiguous behaviour since both workflows run with identical job names inside
- We never test that docs can build after docs changes
- Our conditional statements are long and confusing

### How to test it?

Create PRs that want to merge into this branch with the following cases of changed files:

docs: should only run docs_build 
front-end: should only run front-end and e2e tests
back-end: should only run back-end, e2e, typescript, cli
github workflow, package.json, jest file: should run all tests

in all cases, aggregate_test_result should pass once the run is complete and build, lint, and pretty should still run

### Related issue(s)/PR(s)


.github/filters.yaml
```diff
@@ -1,26 +1,37 @@
...
+  - 'fileTransformer.js'
 api:
   - 'tests/api/**'
 e2e:
   - 'tests/e2e/**'
   - 'playwright.base.config.js'
 cli:
   - 'tests/cli/**'
+
+# TODO: this needs a lot more in it
```
innerdvations, line 25: but it's an improvement over what we had, so we can build it up over time if we notice tests are skipped when they shouldn't be

.github/workflows/tests.yml
```diff
@@ -75,31 +94,36 @@ jobs:
...
+  lint:
+    name: 'lint (node: ${{ matrix.node }})'
+    needs: [build] # build is needed so that imports can be verified
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node: [20, 22]
     steps:
       - uses: actions/checkout@v4
+        with:
```
alexandrebodin, line 109: isn't this the default behavior ?

.github/filters.yaml
```diff
@@ -1,28 +1,37 @@
...
+all_tests:
```
alexandrebodin, line 26: should we name this global_sth ? so it makes it more explicit it's global files ? 🤔

alexandrebodin (approved):
Just questions. the config LGTM

.github/filters.yaml
innerdvations: I originally didn't want it to affect docs_build or if we potentially added filters for lint/prettier, but thinking about it maybe it should, since package changes can affect pretty much anything anywhere.

I'll just call it global and have it trigger docs_build too

.github/workflows/tests.yml
innerdvations: Oops, no, but it should be removed anyway

innerdvations: `0` meant we were fetching the entire git history, and we were using it in every job 😆

removing this everywhere could actually make a noticeable difference in setup speed, trying it out now

PR #22169, opened 2024-11-15
fix: 20903 You cannot uncheck the box to select or unselect an audio file in media library

archanaagivale30:
### What does it do?

fix: 20903 You cannot uncheck the box to select or unselect an audio file in media library
### Why is it needed?

You cannot uncheck the box to select or unselect an audio file in the media library

### How to test it?

- Go to MediaLib
- Upload an audio file
- Try to select it with the checkbox

### Related issue(s)/PR(s)

Ref : #20903


hanpaine:
🙌 Hey @archanaagivale30 nice one! Thanks for another fix 🙏 we'll take a look soon - in the meantime, would you mind adding "fixes" or similar before the issue ref so that it will auto-close please? [More info about how that works on GH here ](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)😁

Adzouz:
Hi @archanaagivale30 and thanks a lot for your fix and the time you took to handle this issue 🙏
Unfortunately I'll have to close the PR since it was already fixed.
Have a great day 🚀

PR #22171, opened 2024-11-15
feat(cloud-cli): display warning when using deploy command

gonbaum:
### What does it do?

We add a warning message to be displayed when the deploy command is used and the target environment has already a deployment done or is in progress.

At the same time, a `-f / --force` flag to skip the warning and confirmation prompt has been introduced.

### Why is it needed?

Deploying a local project with a different db state will incur in data loss. We need to warn users that deploying is a also a destructive operation. 

### How to test it?
----
**Pretest** 

**Verify Deployment Commands**

- Steps: Run the `strapi deploy -h` command.
Confirm the help output includes the following commands and options:

```
Deploy a Strapi Cloud project

Options:
  -d, --debug       Enable debugging mode with verbose logs
  -s, --silent      Don't log anything
  -f, --force       Skip confirmation to deploy
  -e, --env <name>  Specify the environment to deploy
  -h, --help        display help for command
```

**Tests**

You would need an environment with a deployment done and another without. To achieve the second, you can create a new environment and cancel the build process.

You should test all ways to deploy a project:

- `deploy` command
- `deploy -e <environment>`
- `deploy --env <environment>`
- deploy with default environment set (`cloud environment link `→ select an environment)

You should at least test one case with a deployment in course and one with a finished successful deployment.

**1. Project with no deployments:**

A project with no deployments should not display a warning on deploy in any case

**2. Project with deployments in course or done:**

- A project with a deployment should always display warning and ask for confirmation to deploy

- When using `deploy -f` / `deploy --force` the warning should not be displayed and the deployment should be done straightaway

- `deploy -e <environment> -f` should deploy straightaway without warning.

- Warning should look like in the screenshot:

Screenshot 2024-11-14 at 11 52 01

### DIFF
--- a/packages/cli/cloud/package.json
+++ b/packages/cli/cloud/package.json
@@ -47,6 +47,7 @@
   "dependencies": {
     "@strapi/utils": "5.4.0",
     "axios": "1.7.4",
+    "boxen": "5.1.2",
     "chalk": "4.1.2",
     "cli-progress": "3.12.0",
     "commander": "8.3.0",
--- a/packages/cli/cloud/src/deploy-project/action.ts
+++ b/packages/cli/cloud/src/deploy-project/action.ts
@@ -1,5 +1,6 @@
 import fse from 'fs-extra';
 import inquirer from 'inquirer';
+import boxen from 'boxen';
 import path from 'path';
 import chalk from 'chalk';
 import { AxiosError } from 'axios';
@@ -7,7 +8,13 @@ import * as crypto from 'node:crypto';
 import { apiConfig } from '../config/api';
 import { compressFilesToTar } from '../utils/compress-files';
 import createProjectAction from '../create-project/action';
-import type { CLIContext, CloudApiService, CloudCliConfig, ProjectInfo } from '../types';
+import type {
+  CLIContext,
+  CloudApiService,
+  CloudCliConfig,
+  EnvironmentDetails,
+  ProjectInfo,
+} from '../types';
 import { getTmpStoragePath } from '../config/local';
 import { cloudApiFactory, tokenServiceFactory, local } from '../services';
 import { notificationServiceFactory } from '../services/notification';
@@ -25,7 +32,16 @@ type PackageJson = {
 
 interface CmdOptions {
   env?: string;
+  force?: boolean;
 }
+
+const boxenOptions: boxen.Options = {
+  padding: 1,
+  margin: 1,
+  align: 'center',
+  borderColor: 'yellow',
+  borderStyle: 'round',
+};
 
 const QUIT_OPTION = 'Quit';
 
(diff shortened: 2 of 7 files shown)


packages/cli/cloud/package.json
```diff
@@ -47,6 +47,7 @@
   "dependencies": {
     "@strapi/utils": "5.4.0",
     "axios": "1.7.4",
+    "boxen": "5.1.2",
```
kibwashere, line 50: nice one! do you confirm it's ok to add more modules? not sure what are the rules of this repo ^^'

packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -281,6 +300,26 @@ export default async (ctx: CLIContext, opts: CmdOptions) => {
 
   project.targetEnvironment = await getTargetEnvironment(ctx, opts, project, environments);
 
+  if (!opts.force) {
+    const shouldDisplayWarning = await hasPendingOrLiveDeployment(
```
kibwashere, line 304: does it need to be await/async?

kibwashere (commented):
just few thoughts! looks good :)

packages/cli/cloud/src/deploy-project/action.ts
```diff
@@ -206,6 +211,17 @@ async function getTargetEnvironment(
...
 
+async function hasPendingOrLiveDeployment(
+  environments: EnvironmentDetails[],
+  targetEnvironment: string
+): Promise<boolean> {
+  const environment = environments.find((env) => env.name === targetEnvironment);
+  if (!environment) {
+    throw new Error(`Environment details ${targetEnvironment} not found.`);
+  }
+  return environment.hasPendingDeployment || environment.hasLiveDeployment || false;
```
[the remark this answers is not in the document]
gonbaum: Being extra defensive, in the case that these two properties are not yet defined (which could but shouldn't happen in a very freshly created project), having false would ensure that deploy continues normally.

packages/cli/cloud/src/deploy-project/action.ts
gonbaum: Good catch, it doesn't seem needed since we're not performing async operations in the function. Correcting.

packages/cli/cloud/src/environment/link/action.ts
```diff
@@ -17,8 +17,10 @@ interface LinkEnvironmentInput extends Answers {
   targetEnvironment: string;
 }
 
-type Environment = {
+export type Environment = {
   name: string;
+  hasLiveDeployment: boolean;
+  hasPendingDeployment: boolean;
```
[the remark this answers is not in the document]
gonbaum: I think it's not needed. Good catch.

kibwashere:
QA looks good to me!
good job @gonbaum!

kibwashere (approved):

PR #22175, opened 2024-11-15
enhancement(data-transfer): add directory provider

innerdvations:
### What does it do?

Adds a directory file provider to data-transfer

TODO:
 
- [ ] update prettier config to ignore the data dumps
- [ ] fix import/export so that e2e tests aren't all failing
- [ ] make sure it's flagged as internal use only
- [ ] add a test or three of the provider itself?

### Why is it needed?

To allow us to use raw files in our e2e test setup (and potentially open it up as an option to users in the future)

Destination isn't needed at this time, because we can just extract our tar export, but it could be a nice addition later

### How to test it?

Extract a data-transfer tar file
set up a programmatic import using the path of that directory instead of the tar file

### Related issue(s)/PR(s)

Add the destination provider: https://github.com/strapi/strapi/pull/22176
Use this PR in tests: https://github.com/strapi/strapi/pull/22180


packages/core/data-transfer/src/directory/providers/source/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  async #loadMetadata() {
+    const metadataPath = path.join(this.options.directory.path, METADATA_FILE_PATH);
+    this.#metadata = await this.#readJSONFile<IMetadata>(metadataPath);
+  }
+
+  async #loadAssetMetadata(filePath: string) {
+    return this.#readJSONFile<IFile>(filePath);
```
Convly, line 80: This isn't being called by the `createAssetsReadStream` method, is it intentional?

packages/core/data-transfer/src/directory/providers/source/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  async getMetadata() {
+    if (!this.#metadata) {
+      await this.#loadMetadata();
+    }
+
+    return this.#metadata ?? null;
+  }
+
+  async getSchemas() {
+    const schemaCollection = await utils.stream.collect<Struct.Schema>(
```
Convly, line 92: Kind of annoying we have to do this (read => convert to stream => convert to a collection) since we could basically do a simple
`concat(files.map(f => fs.read(f)).split('\n').map(l => JSON.parse(l))` (kinda)

But your impl keeps consistency with everything else, so let's stick with it

packages/core/data-transfer/src/directory/providers/source/index.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  createConfigurationReadStream(): Readable {
+    return this.#streamJsonlDirectory('configuration');
+  }
+
+  createAssetsReadStream(): Readable {
+    const outStream = new PassThrough({ objectMode: true });
+    const assetsDir = path.join(this.options.directory.path, 'assets/uploads');
+
+    fs.readdir(assetsDir, { withFileTypes: true }, (err, files) => {
```
Convly, line 127: Could be simpler to create a stream from an async generator and iterate through its content (e.g. using `fs.opendir`) rather than relying on a custom impl, what do you think ?

packages/core/data-transfer/src/directory/providers/source/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    });
+
+    return outStream;
+  }
+
+  #streamJsonlDirectory(subDirectory: string): Readable {
+    const fullPath = path.join(this.options.directory.path, subDirectory);
+    const outStream = new PassThrough({ objectMode: true });
+
+    fs.readdir(fullPath, { withFileTypes: true }, (err, files) => {
```
Convly, line 172: Same than for the assets read stream, maybe we could try to create a stream using an async iterator to avoid introducing potential issues (like oversights in our custom impl) + keep it easy to read?

Convly (changes requested):
Overall looks nice, haven't tested everything yet

Marc-Roig:
couldnt be more thankful to have this! 
Could you update the contributor docs with a mention of this source provider? It would be helpful to have a refernce of how to set it up 🙏

innerdvations:
> couldnt be more thankful to have this! Could you update the contributor docs with a mention of this source provider? It would be helpful to have a refernce of how to set it up 🙏

I won't be able to work on this very much in the next few weeks :( if you'll have any time though, feel free to push changes directly to this or the other 2 related PRs

Convly:
Converting for draft as Ben won't have the time to finish it for now.

If someone wants to take it over the finish line, feel free to ping me for any help with the implementation/review.

innerdvations:
> Small update made on the source one + dest implementation: [#22180 (comment)](https://github.com/strapi/strapi/pull/22180#issuecomment-2498270157)

I merged it here, we'll just handle the feature as a whole. I'll keep a TODO list in the description so anyone who wants to can help finish this

innerdvations:
Closing in favor of https://github.com/strapi/strapi/pull/25867

PR #22176, opened 2024-11-15
enhancement(data-transfer): add directory destination provider

innerdvations:
### What does it do?

adds a directory destination provider

### Why is it needed?

to save a step for anyone who needs to script exporting data as directories instead of tar

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
Closing in favor of #22180

PR #22180, opened 2024-11-15
chore(e2e): use directory source provider for data sets

innerdvations:
### What does it do?

use the new directory source for e2e test data sets

### Why is it needed?

to view text files on changes to the data set instead of one big binary tar that can't be easily viewed or tracked in git history

### How to test it?

tests should pass

try updating the data set json/jsonl and it should work

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22175


markkaylor:
> Hmm. Can you tell what has actually changed? When I visually compare the diff lines from the video they look identical.

The majority of the diff is probably identical since the _every_ file was replaced. Maybe [`git mv`](https://git-scm.com/docs/git-mv) could help? I'll give it a try in a little bit. 

> you can just commit specifically what changed instead of all the other records as well.

I think this would be hard to identify and/or error prone

innerdvations:
The media metadata files being difficult to read when something is updated I think is fine, if you're modifying or adding an image it's going to be a long block of JSON describing all the metadata of the media item. If they're being modified every export (which I'm not sure if that's what you mean) can you see what exactly is changing? 

If the entities exports are being completely modified every export, that's a bigger concern to solve I think than the media metadata. We could easily add the destination provider though with modifications to keep the exports more stable (like sorting or fixing timestamps to not update or whatever the difference is)

Marc-Roig:
This looks like a improvement 🚀 
Would it be possible to export different content types over different entity files? in order to improve readability and make reviews easier, it could help if we had an "articles.json" file to take a look

innerdvations:
I think what we should do is finish the directory destination provider and fix it up to export in a "cleaner" manner, like running prettier on the json. 

Then we could also standardizing any data that might change incidentally between exports, like timestamps, filtering out anything that might change unexpected in the core_store, etc in the transforms in the transfer engine.

markkaylor:
In my example case I would like to have a diff after export that only reflects the changes I made. I only added an article with title test. So I would expect a single-line diff like:

```
27 {"type":"api::article.article","id":4,"data":{"documentId":"godf1ydha4qvrc8zh4mj7b5o","title":" Pourquoi je préfère le football au soccer","content":[{"type":"paragraph","children":[{"type":"text","text":"Salut tout le monde !"}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"Parlons de pourquoi j'aime le football—certains pourraient l'appeler soccer, et c'est bien. Le football, pour moi, est plus qu'un jeu ; c'est un langage universel qui unit les gens à travers le monde. Sa simplicité est magnifique—juste un ballon, un bout de gazon, et la joie de le faire circuler. Contrairement à l'action stop-and-start du football américain, le flux continu, le rythme et le travail d'équipe du football résonnent avec mon âme."}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"La danse sur le terrain, l'anticipation et l'effort collectif d'une équipe font du football quelque chose de spécial. L'énergie dans un stade de football est inégalée—les chants, les chansons, le rugissement de la foule—c'est électrisant. Bien que j'apprécie la compétition dans tous les sports, la camaraderie du football réchauffe mon cœur. Les joueurs des équipes adverses échangent des maillots, reconnaissant un amour partagé pour le jeu."}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"Alors, que ce soit du football ou du soccer, concentrons-nous sur la magie, la joie et les liens que le jeu apporte. À mes yeux, il n'y a pas de sport plus grand, et je n'échangerais le monde du football pour rien d'autre."}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"À la beauté du jeu !"}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"Le vôtre dans le football (ou le soccer, si vous préférez),"}]},{"type":"paragraph","children":[{"type":"text","text":""}]},{"type":"paragraph","children":[{"type":"text","text":"Ted Lasso"}]}],"slug":null,"createdAt":"2024-01-31T09:01:13.326Z","updatedAt":"2024-01-31T09:01:13.326Z","publishedAt":null,"locale":"fr","seo":null}}
---added
28 {"type":"api::article.article","id":5,"data":{"documentId":"sy5ulkgfo7axtarqc5ee7sw8","title":"test","content":null,"slug":null,"createdAt":"2024-11-18T08:25:36.247Z","updatedAt":"2024-11-18T08:25:36.247Z","publishedAt":null,"locale":"en","seo":null}}
---
```
Screenshot 2024-11-18 at 17 17 52

The only way I can get a diff like this is if I copy paste the changes from the export to the existing data set, but I really don't like that. 

I've tried git mv which I've used in the past to keep the git history when migrating a js file to ts, but I haven't been able to get that working. 

If the process to update the data set is to just replace the existing data set files with the new one I don't think we will even end up with a real git history. Instead I think we would just always have a single commit per file showing the file creation. It would also mean every time we open a PR to update the data set we will have a diff with like 100 files. 

Am I missing something? Do you see any way we can get a meaningful diff after export besides just copy/paste? What did you have in mind for the flow of replacing the old data set with the new one?

Convly (commented):
> I think what we should do is finish the directory destination provider and fix it up to export in a "cleaner" manner, like running prettier on the json, sorting, all those things you mentioned

The exported files' format is JSONL, I'm afraid running prettier on them won't be possible, except if we change it to JSON


Also, I'm quite interested in why there are so many diffs at all. Like, what are those changes except for the intended ones? Is it because of sorting?

>Not dramatic, but, would it be possible to export different content types over different entity files? in order to improve readability and make reviews easier, it could help if we had an "articles.json" file to take a look.
That might be simpler than keeping the export format constant on every export, entries within the same content type file would be sorted by ID, and that would always keep the same order

That's doable

Convly:
Small update, I ended up taking an hour to implement the destination provider + fix the source one.

Here's the output of the destination one, taking into account @Marc-Roig request to have entries in different files based on their types.

image

I've kept `jsonl` formatting for now, but these could very well be regular `json` files if we want to apply prettier formatting to them later.

I've tested the `LocalDirectoryDestinationProvider` behavior by manually modifying the strapi export command so that it uses it instead of the `LocalFileDestinationProvider`.

PR #22191, opened 2024-11-18
test(api): check cleanup of data after deleting a component

innerdvations:
### What does it do?

Proof-of-concept for migration builder

### Why is it needed?

This feature would enable graceful renaming and correct deletion of components, attributes, and content types while ensuring that database integrity and data consistency are maintained. Renaming entities, in particular, is challenging without a system to handle the underlying database changes seamlessly. The migration builder ensures that these operations can occur without accidental data loss or orphaned relationships.

These improvements are essential for projects with complex schemas or in production environments, where schema modifications need to be handled with precision and reproducibility. By introducing a migration builder, Strapi can:
	•	Preserve existing data during operations like renaming attributes or content types.
	•	Ensure predictable outcomes for deletions and schema updates.
	•	Enable reproducible environments by version-controlling schema changes.
	•	Improve the developer experience by simplifying safe content-type and component modifications.

In the future, this feature could evolve to completely replace the startup database sync process for new projects by auto-generating migrations for any schema-altering action. This aligns Strapi’s database management with other schema-driven systems such as:
	•	Django ORM with its built-in migrations (makemigrations and migrate) to track and execute schema changes.
	•	Laravel with its migration scaffolding (php artisan make:migration) to version and control database schema updates.
	•	Prisma with its prisma migrate tool for generating and applying migration scripts.
	•	Ruby on Rails with its database migration system that enables both schema definition and data transformation in sync with application changes.

Adopting a migration-based workflow would provide developers with fine-grained control over database evolution and the ability to manage their database changes seamlessly across environments.

### How to test it?

This proof of concept only supports deleting components references from content types in which it is used, essentially solving https://github.com/strapi/strapi/pull/22112

So to test that:
- add a component to a few content types
- delete the component
- see that you now have a migration file that runs on strapi startup and deletes that component data from the table (solving the linked issue even without my dev solution in https://github.com/strapi/strapi/pull/22112 )

### Related issue(s)/PR(s)


packages/core/content-type-builder/server/src/controllers/components.ts
```diff
@@ -135,6 +138,10 @@ export default {
 
       const component = await componentService.deleteComponent(uid);
 
+      const migrationBuilder = new MigrationBuilder();
+      migrationBuilder.addDeleteComponent(uid);
+      await migrationBuilder.writeFiles();
```
innerdvations, line 143: This isn't actually necessary anymore, but provides an example of how we could begin adding this feature everywhere for users who wish to use `database.settings.forceMigrations:false`

packages/core/content-type-builder/server/src/controllers/content-types.ts
```diff
@@ -113,6 +117,29 @@ export default {
 
       const contentTypeService = getService('content-types');
 
+      // Find attributes with a `renamed` field that contains the original name
```
innerdvations, line 120: as mentioned, this isn't how the final version will be implemented; we will probably send it as a separate object rather than tacking it on to the schema itself

innerdvations:
Closing this because it proved its point and we now have more information to track this in the updated CTB so the approach would be slightly different.

PR #22195, opened 2024-11-18
chore: added french translations for blocks

lucasboilly:
### What does it do?

Added french translations for Blocks.

### Why is it needed?

Blocks currently exists only in english.

### How to test it?

Open any entry containing a Blocks field.


packages/core/admin/admin/src/translations/fr.json
```diff
@@ -526,5 +526,32 @@
...
+  "components.Blocks.modifiers.strikethrough": "Barré",
+  "components.Blocks.modifiers.code": "Code",
+  "components.Blocks.link": "Lien",
+  "components.Blocks.expand": "Agrandir",
+  "components.Blocks.collapse": "Rétrécir",
+  "components.Blocks.popover.text": "Texte",
+  "components.Blocks.popover.text.placeholder": "Enter link text",
+  "components.Blocks.popover.link": "Lien",
+  "components.Blocks.popover.link.placeholder": "Coller un lien",
+  "components.Blocks.popover.link.error": "Please enter valid link",
```
markkaylor, line 542: Should this still be in english?

packages/core/admin/admin/src/translations/fr.json
```diff
@@ -526,5 +526,32 @@
...
+  "components.Blocks.modifiers.bold": "Gras",
+  "components.Blocks.modifiers.italic": "Italique",
+  "components.Blocks.modifiers.underline": "Souligné",
+  "components.Blocks.modifiers.strikethrough": "Barré",
+  "components.Blocks.modifiers.code": "Code",
+  "components.Blocks.link": "Lien",
+  "components.Blocks.expand": "Agrandir",
+  "components.Blocks.collapse": "Rétrécir",
+  "components.Blocks.popover.text": "Texte",
+  "components.Blocks.popover.text.placeholder": "Enter link text",
```
markkaylor, line 539: Should this still be in english?

packages/core/admin/admin/src/translations/fr.json
lucasboilly: I missed this one, but I'm not sure we use it since the "Save" button is disabled until we paste/type a valid link. I translated it anyway, just in case 👍

packages/core/admin/admin/src/translations/fr.json
lucasboilly: No, I missed this one too!

markkaylor (approved):

Issue #20792, opened 2024-07-16
[Design] Review Workflows : pointer is not the right one when clicking on a workflow or on [Create new workflow]


MarionLemaire:
https://www.notion.so/strapi/Review-Workflows-pointer-is-not-the-right-one-when-clicking-on-a-workflow-or-on-Create-new-workfl-b9a8e684b5a14c0eb35982c9a34ab576?pvs=4

yanniskadiri:
Could not reproduce in rc.7 but there is a design issue with the "+" icon in "+ Create a new workflow" (see screenshot)

![Image](https://github.com/user-attachments/assets/41e98bdd-c02e-49c6-bdfb-ec18af43b8c4)

yanniskadiri:
Discussed with @lucasboilly, we want the cursor when we hover on the entire row.

yanniskadiri:
And we need to put a real icon there, not a "+" icon in a circle.

![image](https://github.com/user-attachments/assets/bfbca058-0555-442e-a20d-384ae23a3edb)

yanniskadiri:
Here are the specs:
- 32px for the icon
- 8px of padding with the text

sg1asgard:
@yanniskadiri  To replace the icon we need to update the design-system.
I have a meeting with @HichamELBSI on Monday to introduce me to the design-system and see if I can start doing some work in there / fixes.

PR #22197, opened 2024-11-18
fix: added the prop for pointer cursor

sg1asgard:
### What does it do?
I had to pass a prop to display the cursor as pointer to Create new workflow button inside Settings/Review Workflows.
It is just cosmetics fix.

### Why is it needed?
The cursor had to show as pointer on mouse over.

### How to test it?
- Go to Settings
- Go to Review Worfklows (you will need a paid license)
- Hover over Create new workflow below the list of workflows
- The cursor should now be a hand like pointer.

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/20792

### DIFF
--- a/packages/core/review-workflows/admin/src/routes/settings/index.tsx
+++ b/packages/core/review-workflows/admin/src/routes/settings/index.tsx
@@ -162,7 +162,7 @@ export const ReviewWorkflowsListView = () => {
           rows={workflows}
           footer={
             canCreate ? (
-              <TFooter icon={<Plus />} onClick={handleCreateClick}>
+              <TFooter cursor={'pointer'} icon={<Plus />} onClick={handleCreateClick}>
                 {formatMessage({
                   id: 'Settings.review-workflows.list.page.create',
                   defaultMessage: 'Create new workflow',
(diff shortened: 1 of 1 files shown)


sg1asgard:
Pardon me, I forgot all the labels I need to add. If I have missed a few please let me know.

packages/core/review-workflows/admin/src/routes/settings/index.tsx
```diff
@@ -162,7 +162,7 @@ export const ReviewWorkflowsListView = () => {
           rows={workflows}
           footer={
             canCreate ? (
-              <TFooter icon={<Plus />} onClick={handleCreateClick}>
+              <TFooter cursor={'pointer'} icon={<Plus />} onClick={handleCreateClick}>
```
remidej, line 165:
```suggestion
              <TFooter cursor="pointer" icon={<Plus />} onClick={handleCreateClick}>
```

just a code style thing but we use plain strings if there's no logic (I wish the linter would catch that)

remidej:
About our tags:

- the source: tag should be the package in which you're making changes, so in this case it would be review workflows
- severity is more intended for issues, you don't need to set it for your PRs

packages/core/review-workflows/admin/src/routes/settings/index.tsx
sg1asgard: Done.

remidej (approved):
Looks good to me, let's just make sure the tests pass :)

derrickmehaffy:
(Came in to help George to help fix the commit lint)

sg1asgard:
> (Came in to help George to fix the commit lint)

Thanks Derrick!

PR #22206, opened 2024-11-19
fix: admin build path when serverAdminPanel is false

### LINKED ISSUE #19330: Deployment on different servers not working

Launched In: 166 ms
Environment: development
OS: darwin-arm64
Strapi Version: 4.19.0
Node/Yarn Version: npm/10.2.3 node/v20.10.0 darwin arm64 workspaces/false
Edition: Community
Database: sqlite

### Describe the bug

I'm following the [deployment guide](https://docs.strapi.io/dev-docs/admin-panel-customization#different-servers) for different servers (I want to deploy the admin panel - react app - to Netlify and the server to a droplet or similar). Still, the generated index.html file contains invalid paths to the generated js files.

1. Create a Strapi example project using the quick start guide - https://docs.strapi.io/dev-docs/quick-start
2. Configure the server.js and admin.js files as described in the different server configuration guide - https://docs.strapi.io/dev-docs/admin-panel-customization#different-servers
3. Run NODE_ENV=production npm run build
4. The generated paths to the js files contain an extra slash, generating 404 errors:

src="//runtime~main.35ff75f1.js"
src="//main.19186bea.js"

### Expected behavior

Create the index.html with the correct paths with only one slash to access the admin site using the http://yourfrontend.com URL. I think in this context the proxy recommendation not apply because this is a front-end app with the index.html file and js files, must work with any static hosting provider.

butcherZ:
### What does it do?

determine `basePath` based on the `serveAdminPanel` value

### Why is it needed?

By default, `serverAdminPanel` is true and admin panel is served by Strapi back end at `/admin`, therefore the assets are loaded from `/admin/runtime~main.js`

when `serverAdminPanel` is false, it generates relative path (`runtime~main.js`) to support deploying the admin panel separately from the backend.

### How to test it?

-  create a new strapi app with ~~`npx create-strapi-app@0.0.0-experimental.d10cfc2d4947d3dcbf166b29dc168cb4a7123772`~~ ~~`npx create-strapi-app@0.0.0-experimental.4184db001fb3249acae738f57aab40d83602c0e9`~~ `npx create-strapi-app@0.0.0-experimental.6e4965146398480201de1e5fa870efbead50da6a`
-   add `url: "http://blahblah.com"`,  to `./config/server.js` 
-  add `url: "/",` and `serverAdminPanel: false` to `./config/admin.js`([more info here](https://docs-v4.strapi.io/dev-docs/admin-panel-customization#different-servers))
- run `yarn build`
- the `index.html` file located in the `build` folder should have relative path like `src="runtime~main.xxxx.js"` instead of `src="//runtime~main.2e50647e.js"` reported in the issue

### Related issue(s)/PR(s)

fixes #19330

DX-1655


packages/core/admin/_internal/node/createBuildContext.ts
```diff
@@ -170,11 +170,14 @@ const createBuildContext = async ({
...
 
+  const serveAdminPanel = strapiInstance.config.get('admin.serveAdminPanel', true);
+
   const { bundler = 'webpack', ...restOptions } = options;
 
   const buildContext = {
     appDir,
-    basePath: `${adminPath}/`,
+    // If serveAdminPanel is false, use relative paths for separate deployment
+    basePath: serveAdminPanel ? adminPath : '',
```
jhoward1994, line 180: I haven't tested this yet but given the description of the issue I would expect us to still need the trailing '/' on `adminPath` in the `serveAdminPanel = true` case. Did you test that or have I missed something?

I'll test out the experimental now 🙏🏻

butcherZ: I have tested it before opening the PR, but it'd be nice if you could confirm the behaviour with me :D many thanks!

jhoward1994, line 180:
```suggestion
    basePath: serveAdminPanel ? `${adminPath}/` : '',
```

This is the change that works for me. In both cases (true and false) of `serveAdminPanel`

Convly:
Note: somewhat related discussions in https://github.com/strapi/strapi/pull/21914, https://github.com/strapi/strapi/pull/21507 and https://github.com/strapi/strapi/issues/21184

alexandrebodin:
IS that fix necessary with the changes in https://github.com/strapi/strapi/pull/21914. ?

butcherZ:
@alexandrebodin I blocked it for the time being, I wasn't aware of #21914 until yesterday, but this PR is intended to solve the mentioned v4 issue in https://github.com/strapi/strapi/issues/19330, Should I update it accordingly to #21914? @Convly @innerdvations what do you guys think? :D

packages/core/admin/_internal/node/createBuildContext.ts
```diff
@@ -170,11 +170,14 @@ const createBuildContext = async ({
...
 
+  const serveAdminPanel = strapiInstance.config.get('admin.serveAdminPanel', true);
+
   const { bundler = 'webpack', ...restOptions } = options;
 
   const buildContext = {
     appDir,
-    basePath: `${adminPath}/`,
+    // If serveAdminPanel is false, use relative paths for separate deployment
+    basePath: serveAdminPanel ? `${adminPath}/` : '',
```
innerdvations, line 180:
```suggestion
    basePath: serveAdminPanel ? `${adminPath}/` : strapiInstance.config.get('admin.url', ''),
```

That seems like it would match [the documentation](https://docs-v4.strapi.io/dev-docs/admin-panel-customization#different-servers) closer, but I might be missing something.

If we make use of admin.url it would allow users to set whatever prefix they want

innerdvations (approved):
Works for me!

derrickmehaffy (changes requested):
So I have a couple of questions/concerns when testing this for prod.

I've set the server and admin URL as well as the serve admin and when I run the build I noticed something we have a `.strapi/client/index.html` and the `build/index.html` I assumed that I needed to statically serve the `build/index.html` but it's not working (giving me a white screen) so I assume I need to use `.strapi/client/index.html` but it has no references to loading the build (also not sure how I would setup the folder structure copying this into a static load directory like `/var/www` or something).

Does anyone have any clarity on if the `.strapi/client/index.html` should be used?

packages/core/admin/_internal/node/createBuildContext.ts
```diff
@@ -170,11 +170,14 @@ const createBuildContext = async ({
...
 
+  const serveAdminPanel = strapiInstance.config.get('admin.serveAdminPanel', true);
+
   const { bundler = 'webpack', ...restOptions } = options;
 
   const buildContext = {
     appDir,
-    basePath: `${adminPath}/`,
+    // If serveAdminPanel is false, use relative paths for separate deployment
+    basePath: serveAdminPanel ? `${adminPath}/` : strapiInstance.config.get('admin.url', ''),
```
innerdvations, line 180:
```suggestion
    basePath: serveAdminPanel ? `${adminPath}/` : strapiInstance.config.get('admin.url', `${adminPath}/`),
```
I'm concerned this will be a breaking change for someone if we change the default values. This will allow it to continue working exactly as before if admin.url isn't set, but now fixes the code to match the documentation, where it says to use admin.url to set this value.

jhoward1994:
I agree with your suggestion @innerdvations the docs say it should fall back to `/admin`

I've tested with serveAdminPanel true using a default and custom adminUrl and it is working for me 👍🏻

With the config

```
  serveAdminPanel: false,
  url: '/',
```

I see this in `examples/getstarted/build/index.html`


```
      }</style><script defer="defer" src="/runtime~main.cbbf53b3.js"></script><script defer="defer" src="/main.f2ef5f00.js"></script></head><body><div id="strapi"></div><noscript><div class="strapi--root"><div class="strapi--no-js"><style>.strapi--root {
```

imo this is correct and we should expect to see the slash prefix here `src="/main.f2ef5f00.js` ? but it doesnt match the PR desc exactly

jhoward1994 (approved):

PR #22207, opened 2024-11-19
fix: cleanup component data after delete

innerdvations:
### What does it do?

- remove component data from the *_components tables after a component is deleted

### Why is it needed?

mitigates https://github.com/strapi/strapi/issues/20931 for v4

an improved fix that also cleans affected production systems may come in the future

### How to test it?

test is included

- create a component
- add it to several content types
- delete the component
- the data should now be removed from the _components table
- you should now be able to recreate a component with the same name as the deleted one and it should work

### Related issue(s)/PR(s)

mitigates https://github.com/strapi/strapi/issues/20931 for v4

v5 version: https://github.com/strapi/strapi/pull/22112

DX-1731


Convly (approved):
Same comments as V5, but overall LGTM

innerdvations:
I have one last thing I want to test here before merging, I'm worried it could still leave some traces on non-dev environments that could make the issue worse there

innerdvations:
Closing in favor of the future v4 version of https://github.com/strapi/strapi/pull/22267

Issue #21791, opened 2024-10-10
_q search broken due to query param encoding


derrickmehaffy:
### Required System information

- Node.js version: v20
- NPM version: yarn v1
- Strapi version: 5.0.4
- Database: SQLite
- Operating system: Linux Mint 21.2
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When searching in the content-manager the encoding of the search request is wrong.

Looking at the database query logs it appears the encoded characters are being escaped and are being searched for literally.

### Steps to reproduce the behavior

1. Create some entries in a content-type like `test 1` and `test 2`
2. Try to search specifically for `test 1`
3. Look at REST query logs and database query logs and see that the space is being encoded into `%20`
4. Within the database query see that the `%` is being escaped: `'%test\\%202%'` 
5. See error

### Expected behavior

We should decode the request before handing it off to the database

### Screenshots

![image](https://github.com/user-attachments/assets/36f14d8f-8ceb-4e13-acc1-f3199565f80b)

### Code snippets

N/A

### Additional context

N/A

SANGET:
I have same issue when the parameter is not English.

<img width="1405" alt="image" src="https://github.com/user-attachments/assets/a43d4525-447b-4bb0-a3c7-111c6faeaa45">

Required System information
Node.js version: v20
NPM version: yarn v1
Strapi version: 5.1.0
Database: Postgre

artsiompeshko:
+ 1.

derrickmehaffy:
Bumping sev due to a customer ticket, reference is TID8982

PR #22212, opened 2024-11-19
fix: search param being encoded twice

Marc-Roig:
### What does it do?

Searching in the List View with special characters ( * & / ...) was not working.
Characters were being encoded twice.

The solution proposes removing all encodings we had for `_q` (or any other param) , and handling the encoding inside the injected content manager endpoints.

Ideally, this should be done by the api interceptor, but doing so would be a breaking change at the moment.

### How to test it?
- Create some entries in a content-type named like `A & B` and `C & D`
- Try to search specifically for `A & B`
- The `A & B` entry should appear in the list

### Related issue(s)/PR(s)
Fixes https://github.com/strapi/strapi/issues/21791

### DIFF
--- a/packages/core/content-manager/admin/src/utils/api.ts
+++ b/packages/core/content-manager/admin/src/utils/api.ts
@@ -32,12 +32,6 @@ const buildValidParams = <TQuery extends Query>(query: TQuery): TransformedQuery
     ),
   };
 
-  if ('_q' in validQueryParams) {
-    // Encode the search query here since the paramsSerializer will not
-    // @ts-expect-error – TODO: fix this type error
-    validQueryParams._q = encodeURIComponent(validQueryParams._q);
-  }
-
   return validQueryParams;
 };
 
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/utils/api.ts
```diff
@@ -32,12 +32,6 @@ const buildValidParams = <TQuery extends Query>(query: TQuery): TransformedQuery
     ),
   };
 
-  if ('_q' in validQueryParams) {
-    // Encode the search query here since the paramsSerializer will not
-    // @ts-expect-error – TODO: fix this type error
-    validQueryParams._q = encodeURIComponent(validQueryParams._q);
-  }
```
markkaylor, line 39: Isn't buildValidParams used in a lot of places? Did you check the impact removing this has?

Marc-Roig: not yet! I will do the QA today

derrickmehaffy:
Building exp for testing: https://github.com/strapi/strapi/actions/runs/11942041215

Marc-Roig:
I have done some QA on this PR, no other part of the app makes use of the _q param and the buildValidParams utility at the same time.

If this is a documented function, we might as well mention this in the changelog

markkaylor (approved):
LGTM!

Not sure it matters for tracking or whatever but maybe add both related PRs? 🤷‍♂️

PR #22214, opened 2024-11-20
feat: send diagnostic messages from remote strapi

Bassel17:
### What does it do?

- listen to diagnostic messages from the server, and report them so we can log them on the cli and to a file
when using `--verbose`

### Why is it needed?

- So we can receive information about whats happening on the remote strapi during a transfer

### How to test it?

try this transfer, you should see logs from the `push-handler` and `local-destination-provider`

`yarn strapi transfer --to {STRAPI_APP_ADMIN_URL} --to-token {TRANSFER_TOKEN}  --verbose`


innerdvations (commented):
LGTM, I haven't tested it thoroughly yet, will try to do that later today

packages/core/data-transfer/src/strapi/providers/utils.ts
```diff
@@ -221,6 +226,32 @@ export const connectToWebsocket = (address: Address, options?: Options): Promise
...
+            details: response.diagnostic.details,
+            kind: response.diagnostic.kind,
+          });
+        }
+
+        if (response.diagnostic.kind === 'error') {
+          diagnostics?.report({
+            details: response.diagnostic.details,
+            kind: response.diagnostic.kind,
+          });
```
Convly, line 250: Since we're doing the same thing for info, warning and error, couldn't we factorize this a bit ?

Convly (commented):
Overall LGTM, only one suggestion

packages/core/data-transfer/src/strapi/remote/handlers/pull.ts
```diff
@@ -100,6 +121,7 @@ export const createPullController = handlerControllerFactory<Partial<PullHandler
 
     // Transfer message (the transfer must be init first)
     else if (type === 'transfer') {
+      this.onInfo(`recieved transfer action:${msg.action} step:${msg.kind} uuid:${uuid}`);
```
innerdvations, line 124:
```suggestion
      this.onInfo(`received transfer action:${msg.action} step:${msg.kind} uuid:${uuid}`);
```

tests/e2e/utils/dts-export.ts
```diff
@@ -51,7 +51,9 @@ export const exportData = async (): Promise<void> => {
     },
   });
 
-  engine.diagnostics.onDiagnostic(console.log);
+  engine.diagnostics.onDiagnostic((diagnostic) => {
```
innerdvations, line 54: Could you add this info filter to dts-import too?

innerdvations (approved):
Tested and LGTM!

PR #22215, opened 2024-11-20
chore: removing support for assets without metadata

Bassel17:
### What does it do?

- Remove unused assets code from DTS, cleaning up old stuff

### Why is it needed?

- Tiding things up 

### How to test it?

- Attempt an import or a transfer it should work as usual


packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
```diff
@@ -332,41 +331,6 @@ class LocalStrapiDestinationProvider implements IDestinationProvider {
       },
       async write(chunk: IAsset, _encoding, callback) {
         await transaction?.attach(async () => {
-          // TODO: Remove this logic in V5
```
Convly, line 335: Should we add a warning with a clear description that if the transfer fails due to missing metadata, it might be because they're trying to use an outdated dataset?

Convly: Might be worth thinking about (for a future task/PR) a small script to transform old datasets into new ones (basically creating metadata out of the file entities)

packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
```diff
@@ -406,7 +370,7 @@ class LocalStrapiDestinationProvider implements IDestinationProvider {
               return callback();
             }
             const entry: IFile = await strapi.db.query('plugin::upload.file').findOne({
-              where: { id: fileEntitiesMapper[uploadData.id] },
+              where: { id: fileEntitiesMapper?.[uploadData.id] },
```
Convly, line 373: What happens if it's indeed undefined? Should we throw an error? Allow continuing?

packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
innerdvations: Just to be sure this couldn't be considered a breaking change, did v5 every export without metadata?

I think considering we don't officially support v4 backups imported into v5, it's fine to break those cases.

Bassel17: yep every single v5 exports metadata as well as every latest v4, this was a very old fix

packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
innerdvations: Same question. Maybe we should have two checks, one before this query to ensure the id exists, and then after the query to ensure it actually returned the expected result. Does that make sense?

Convly (approved):

PR #22216, opened 2024-11-20
fix(e2e): use clickAndWait to ensure the page is ready

markkaylor:
### What does it do?

- Removes condition in order to `waitForLoadState` on every browser
- Uses clickAndWait when navigating from page to page

### Why is it needed?

- The test is flaky often failing to find the first locator after page navigation
- It fails for webkit and firefox 

### How to test it?

The CI should pass

### DIFF
--- a/tests/e2e/tests/content-manager/history.spec.ts
+++ b/tests/e2e/tests/content-manager/history.spec.ts
@@ -30,7 +30,7 @@ const goToHistoryPage = async (page: Page) => {
 
   // TODO find out why the history button sometimes doesn't appear. Reloading shouldn't be necessary
   if (await historyButton.isVisible()) {
-    await historyButton.click();
+    await clickAndWait(page, historyButton);
   } else {
     await page.reload();
     await goToHistoryPage(page);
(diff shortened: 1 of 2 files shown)


markkaylor:
All tests have passed once, running them again

tests/e2e/utils/shared.ts
```diff
@@ -80,10 +80,7 @@ export const skipCtbTour = async (page: Page) => {
  */
 export const clickAndWait = async (page: Page, locator: Locator) => {
   await locator.click();
-
-  if (page.context().browser()?.browserType().name() === 'webkit') {
-    await page.waitForLoadState('networkidle');
-  }
+  await page.waitForLoadState('networkidle');
```
remidej, line 83: do you know why it was originally webkit only?

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -432,9 +432,7 @@ describeOnCondition(edition === 'EE')('History', () => {
 
       // Create new author
       await clickAndWait(page, page.getByRole('link', { name: 'Content Manager' }));
-      // await page.waitForSelector('text=Author');
       await clickAndWait(page, page.getByRole('link', { name: 'Author' }));
-      // await page.waitForSelector('text=Create new entry');
```
remidej, line 437: 🧹🧹🧹

remidej (commented):
good idea! i didn't realize we weren't using that API in history

simotae14 (approved):
well done 💪🏽

tests/e2e/utils/shared.ts
markkaylor: Not sure, maybe @innerdvations knows more?

innerdvations: webkit is the only browser that actually requires clickAndWait, it doesn't render instantly like the other browsers so you have to have a slight delay in there (we use waitfor `networkidle`) to allow the render to finish before the next step can proceed

innerdvations: I can't say 100%, but I think clickAndWait is necessary for anything that opens a modal

innerdvations: but running it for all browsers is fine with me, at worst I think it will add a few milliseconds to test runs

Issue #22058, opened 2024-11-04
The icon for "Code block" and "Inline code" are the same in Blocks


lucasboilly:
### Why should we do it?
Users may confuse the them, but they are very different. One is a block and the other is simply a text modifier. 

![CleanShot 2024-11-04 at 11 54 26@2x](https://github.com/user-attachments/assets/5711d96e-11a5-468b-94c1-ab3471ba772a)

### Solution
We should use the [code-block](https://phosphoricons.com/?q=%22code+block%22&weight=%22fill%22) icon from Phosphor (our new icon library implemented for Strapi 5) to illustrate the "Code block" block. In order to do it, we'll have to add the icon to our design system as explained in [this issue](https://github.com/strapi/design-system/issues/1804) in the [Strapi Design System](https://github.com/strapi/design-system) repository.

![CleanShot 2024-11-04 at 11 56 26@2x](https://github.com/user-attachments/assets/c1b3fd95-c0e2-4162-b1ac-24d453783350)

sg1asgard:
It is fixed with the following PR:
https://github.com/strapi/strapi/pull/22218

PR #22218, opened 2024-11-20
fix: replaced the Code icon with CodeBlock icon

sg1asgard:
### What does it do?
Replaced the Code icon with CodeBlock icon inside the block editor.

### Why is it needed?
Users were confusing them. One is a code block modifier and the other is simply a inline block modifier.

### How to test it?
Create a collection with a block editor field. Click the dropdown and see at the bottom where it says Code Block.
The icon should be now different than the inline code on the Toolbar.

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/22058


simotae14 (approved):
well done @sg1asgard 👏🏽

Issue #21175, opened 2024-09-06
The confirmation message doesn’t make sense if D&P is not enabled


yanniskadiri:
### **Steps to reproduce**
- Edit an entry of a content-type without D&P enabled
- Go to the content history of this entry (you will need a paid license)
- Restore a past version of the entry
- See the notification saying that the version was restored on top of a published one

### **Expected behavior**

The notification should just say that a past version was restored and not say that it's not published yet.

### **Screenshots**

![Image](https://github.com/user-attachments/assets/c67f8b96-d7cb-47e1-b50c-b3e53e9b4dbf)

sg1asgard:
@yanniskadiri  The issue is closed now.
https://github.com/strapi/strapi/pull/22223

PR #22223, opened 2024-11-20
fix: the confirmation message of restoring a version was updated

sg1asgard:
### What does it do?
It updates the message for the notification box after an user restores a document version.

### Why is it needed?
The message was reported as incorrect.

### How to test it?
- Edit an entry of a content-type without D&P enabled
- Go to the content history of this entry (you will need a paid license)
- Restore a past version of the entry
- See the notification saying that the version was restored on top of a published one

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/21175

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
+++ b/packages/core/content-manager/admin/src/history/components/VersionHeader.tsx
@@ -78,7 +78,7 @@ export const VersionHeader = ({ headerId }: VersionHeaderProps) => {
           }),
           message: formatMessage({
             id: 'content-manager.restore.success.message',
-            defaultMessage: 'The content of the restored version is not published yet.',
+            defaultMessage: 'A past version of the content was restored.',
           }),
         });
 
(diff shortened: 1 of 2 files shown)


simotae14 (approved):
Well done @sg1asgard 🏅

sg1asgard:
@simotae14 @markkaylor 
Well, some tests are failing and I am unsure to what's the problem, if I caused it or it is something else.
If any of you have the time maybe today in the afternoon to assist me? Much appreciated.

markkaylor:
I think the failing test is flaky. Re-running it now to see if it passes. It's not related to your work here. I've see it fail in another PR and it has already been added to a backlog. I plan to take a look today.

sg1asgard:
> I think the failing test is flaky. Re-running it now to see if it passes. It's not related to your work here. I've seen it fail in another PR and it has already been added to a backlog. I plan to take a look today.

Ah, good to know. I've been triple checking my side and pocking around to see what is it all about to no avail : D

markkaylor:
Looks like it could actually be a broken test on develop, I'll take a look in a second

Issue #20838, opened 2024-07-17
CM edit view - given the option of other locales in the drop down even when I don’t have access


yanniskadiri:

yanniskadiri:
Confirmed in rc.7

The locale is showed but the user cannot click on it (it should not be shown at all)

PR #22229, opened 2024-11-21
fix: hide the locale option in the Edit view dropdown if you don't have permission to read it

simotae14:
### What does it do?

To avoid weird behaviors in the Edit view, we have removed the option in the locale dropdown to filter the locale where the user has not read permission

### Why is it needed?

To solve a bug we had in v5, because the option was shown and if you click on it you have an error page

### How to test it?

 - Remove locale A permissions from Editor Role
 - Go to the Content Manager
 - Open an entry that has at least 2 locales
 - See that you have access to locale A as User A even without the permissions

### Related issue(s)/PR(s)
fixes https://github.com/strapi/strapi/issues/20838


markkaylor (approved):
Looks good to me, tested it and it's working

remidej (approved):
I haven't tested but the code looks good 👍

remidej:
I just started a rerun of a failing e2e test. It was the failing iframe query selector, something I think @markkaylor just fixed in develop

Issue #22158, opened 2024-11-14
[Preview] Long title of an entry is not truncated and extend the top bar


MarionLemaire:
### Required System information

- Strapi version: v5
- Operating system: IOS

### Describe the bug

When you have a long title for your entry, the title is not truncated in Preview mode.

### Steps to reproduce the behavior

1. Add a long title to your entry
2. Open the preview
3. See the extended top bar

### Expected behavior

Title should be truncated and shown entirely on hover

### Screenshots

https://github.com/user-attachments/assets/65730ce5-0c01-4216-9dcc-77eca3ff006b

PR #22232, opened 2024-11-21
fix: preview title overflow

Marc-Roig:
## What does it do?

Handle long title names for static preview:

image

Before , the title was being wrapped and made the header higher.

### How to test it?

- Create an entry with a long title
- Open Preview and see the title being cropped to fit maximum 200px

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/22158


markkaylor (approved):
✨

PR #22233, opened 2024-11-21
chore: fix inconsistent test:unit results between environments

innerdvations:
### What does it do?

upgrades lerna to allow us to have only a single version of 'typescript' in the monorepo yarn.lock

### Why is it needed?

- to fix issue mentioned https://github.com/strapi/strapi/pull/22147
- why lerna? it's the only package that was requiring typescript:^4 preventing us from having a consistent 5.x version pinned

### How to test it?

- if github test:unit job fails, this was a success and we can update the unit tests with a fixed order (and investigate if upgrading lerna will break anything in our publish scripts...)
- if github test:unit job succeeds, this doesn't actually fix the problem

### Related issue(s)/PR(s)

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "jest-cli": "29.6.0",
     "jest-environment-jsdom": "29.6.1",
     "jest-watch-typeahead": "2.2.2",
-    "lerna": "6.6.2",
+    "lerna": "8.1.9",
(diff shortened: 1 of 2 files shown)


innerdvations:
Nope, tests passed, so this wasn't the problem 🤔

Issue #22235, opened 2024-11-21
Sharp below v0.33 requires a build environment


LunaticMuch:
### Required System information

- Node.js version: v20.16.0
- NPM version: 10.9.0
- Strapi version: 5.4.0
- Database: SQLIte
- Operating system: MacOS Sequoia
- Is your project Javascript or Typescript: Typescript

### Describe the bug

In the current `strapi` version, the package `@strapi/upload` uses `sharp`. Sharp requires some additional binaries which are either built locally or pulled from github.
This causes troubles when strapi is being install in a corporate environment where a private registry and there's no direct access to internet and/or the proxy blocks the download of binaries.

### Steps to reproduce the behavior

In a corporate environment just make a `yarn install`

### Expected behavior

The install should work smoothly without requiring extra pull of binaries from github.
This is fixed using sharp `v0.33.5`

### Additional context

[There's a thread on sharp](https://github.com/lovell/sharp/pull/2217)

derrickmehaffy:
Hey @LunaticMuch would you be willing to open a PR on this one?

LunaticMuch:
> Hey @LunaticMuch would you be willing to open a PR on this one?

Done already 😄  https://github.com/strapi/strapi/pull/22237

VatsalJain2004:
Hello @LunaticMuch Sir,
If you've already resolved the original issue, kindly request the admins to remove this one as well, since both issues are connected and this one is no longer needed.

Thank you!

tysoncung:
I'm interested in working on this issue. Could you provide more context about what you're looking for? Any additional details about requirements or constraints would be helpful.

ayhid:
Hey, looks like this was already fixed in #22237 (merged Apr 4, 2025) — sharp is at `0.33.5` in `packages/core/upload/package.json` now.

@markkaylor can we close this one out?

markkaylor:
> Hey, looks like this was already fixed in [#22237](https://github.com/strapi/strapi/pull/22237) (merged Apr 4, 2025) — sharp is at `0.33.5` in `packages/core/upload/package.json` now.
> 
> [@markkaylor](https://github.com/markkaylor) can we close this one out?

Yup I think we can thanks!

PR #20146, opened 2024-04-18
chore: bump sharp and pass ignore-engines for yarn v1.x.x

markkaylor:
### What does it do?

- Bump sharp to 0.33.3
- Pass the --ignore-engines to the install for yarn v1.x.x users

Experimental: `0.0.0-experimental.1932e6b506cbbdb53c0b8b9ef0f6b5799ebfe7ca`

PR #22237, opened 2024-11-21
feat(upload): sharp v0.33.5

LunaticMuch:
### What does it do?

Update `sharp` to the current version

### Why is it needed?

New version of sharp does not require any complicated setup for building locally libvips or pulling prebuild binaries from github.

### How to test it?

The change is not breaking in term of functionalities. All tests are passing.

### Related issue(s)/PR(s)

-  https://github.com/strapi/strapi/issues/22235


markkaylor:
In the past we had issues with the build in Strapi cloud when bumping sharp: https://github.com/strapi/strapi/pull/20146. I'll create an experimental to test this in the cloud later today but we can't move forward with this upgrade until the build issue in strapi cloud is resolved.

markkaylor:
> In the past we had issues with the build in Strapi cloud when bumping sharp: #20146. I'll create an experimental to test this in the cloud later today but we can't move forward with this upgrade until the build issue in strapi cloud is resolved.

I tested deploying a project to strapi cloud with the sharp upgrade and it looks like it is working. As for v4 I'm not sure if we have the same issue as before but we can probably move forward with this PR for v5.

markkaylor (approved):

PR #22242, opened 2024-11-22
Remove EarthStriked icon when i18n isn't enabled on a field

Aurelsicoko:
### What does it do?

The difference between the localized vs non-localized icon is almost invisible, making it difficult for content editors to determine what are the localized fields vs the others. As an experiment, we would like to remove the icon when the field isn't localized. Only the localized fields would get an icon to identify them quickly.

### How to test it?

Create a project using a template, enable i18n on some fields for a specific content type, open the Content Manager, try to create a document, and see the icon only visible on some fields.


ancashoria:
Ooooh, this will be such a great improvement! Thank you!

remidej:
replaced by #22774

PR #22245, opened 2024-11-22
fix(e2e): wait for relation response before navigating

markkaylor:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


markkaylor:
The tests have a passed a few times now, pulling develop to run again

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -218,14 +218,25 @@ describeOnCondition(edition === 'EE')('History', () => {
...
       await page.getByRole('button', { name: 'Save' }).click();
-      await page.waitForURL(ARTICLE_EDIT_URL);
+      // Confirm the save was succesful before proceeding, otherwise we may end up on the related page before the relation is established
+      await findAndClose(page, 'Saved Document');
 
       // Delete one of the authors, leaving only Coach Beard
       await clickAndWait(page, page.getByRole('link', { name: 'Will Kitman' }));
       await page.waitForURL(AUTHOR_EDIT_URL);
-      await page.getByRole('button', { name: /more actions/i }).click();
+      await page.getByRole('button', { name: 'More actions' }).click();
```
markkaylor, line 239: I don't think this change is necessary but I'm not too bothered by unless y'all are

tests/e2e/tests/content-manager/history.spec.ts
```diff
@@ -218,14 +218,25 @@ describeOnCondition(edition === 'EE')('History', () => {
...
+
+      // Wait for the relation response before saving, otherwise the document might save without the added relation
+      const relationResponsePromise = page.waitForResponse(async (response) => {
+        return (
+          response.status() === 200 &&
+          response.url().includes('content-manager/relations/api::article.article/authors')
+        );
+      });
       await page.getByText('Coach Beard').click();
+      await relationResponsePromise;
```
remidej, line 230: I don't love this because it's testing an implementation detail instead of something the user will actually see. Of course I don't mind if it's needed to stop the flakiness, but I want to make sure I understand better.

This request is to populate the options of the relations picker? If so how does the test ever move past `await page.getByText('Coach Beard').click()` if the request isn't ready yet?

markkaylor: You know you're right this just doesn't make sense. 

I think I had understood that when you add a relation, the draft was updated immediately with that relation. I thought we just weren't triggering a notification for that and there wasn't a way to test it using something the user sees.....but that't not the case. This was just buying us some time and probably not much different then a setTimeout sorta deal. 

I believe the problem is a relation is added but sometimes it looks like the save button is clicked so quickly that the document is saved before the relation is included in the update.

I've updated now to test the relation's link is visible before proceeding to save. Still not the most intuitive thing when reading the test but maybe it's acceptable for now 🤷‍♂️

remidej: This makes perfect sense now 👍

remidej (approved):

PR #22249, opened 2024-11-22
ci: balance sharding to reduce total workflow runtime

innerdvations:
### What does it do?

- reduces shards from each api job, they run around 6 minutes
- adds shards to each e2e job, they run around 10-20 minutes

### Why is it needed?

This should balance them, reducing the runners needed for API tests and slightly increasing their time, while adding a shard to e2e tests and reducing the time they run. We don't care if API take a bit longer since we're generally waiting 10 minutes on e2e anyway.

### How to test it?

Check the workflow runtimes on this PR vs other PRs targeting develop. If this reduces the total runtime by more than a couple minutes it's probably worth it.

The only thing that changes between these runs are the workflows themselves, all are running the same tests and codebase

2 e2e shards, 5 api shards: 26min 28sec https://github.com/strapi/strapi/actions/runs/11974897789?pr=22161
3 e2e shards, 4 api shards : 21min 42sec https://github.com/strapi/strapi/actions/runs/11975737667?pr=22249
4 e2e shards, 4 api shards: 19min 47sec https://github.com/strapi/strapi/actions/runs/11976261244?pr=22249

Thanks to @Boegie19 for the inspiration+idea even though it's been months finally getting around to it (will also credit you in a commit when I merge!)

An alternative idea would be to run the tests with a domain matrix, the only issue there is we would have to manually update the workflows to run the tests when we add new e2e test domains, which is a bit of a hassle

### Related issue(s)/PR(s)


Convly (approved):
Overall looks good to me

.github/workflows/tests.yml
```diff
@@ -193,7 +193,7 @@ jobs:
       fail-fast: false # remove once tests aren't flaky
       matrix:
         project: ['chromium', 'webkit', 'firefox']
-        shard: [1/2, 2/2]
+        shard: [1/4, 2/4, 3/4, 4/4]
```
Marc-Roig, line 196: time out is set to 60 minutes, does it makes sense to reduce it now that we use more shards?

innerdvations: If we ever see it get stuck and hit the timeout we can reduce it to avoid wasting that time, but so far I haven't seen it happen before, I don't want to reduce it and possibly introduce more flakiness

Marc-Roig (approved):

PR #22264, opened 2024-11-25
Chore: setting up nx releases

Bassel17:
### What does it do?

- Add `@nx/js` packages
- Add release script and configs for nx releases

### Why is it needed?

- we are integrating nx releases and we need it for our auto releases workflow

### How to test it?

- every thing should build `yarn setup` should complete successfully
- if your run `yarn release --dry-run`  you should be able to test a dry run of a release


nx.json
```diff
@@ -45,5 +45,53 @@
...
+        "file": false,
+        "createRelease": "github"
+      }
+    },
+    "version": {
+      "conventionalCommits": false
+    },
+    "conventionalCommits": {
+      "types": {
+        "docs": {
```
innerdvations, line 67: Looking in commitlintrc, we also have `chore` that could show up in a log. And perhaps `revert` although it would be quite unlikely

I think it's fine to leave out `test` `ci` and `release` because those shouldn't affect users at all and don't need to be listed in changelogs.

.commitlintrc.ts
```diff
@@ -18,6 +18,7 @@ const config: UserConfig = {
         'revert',
         'security',
         'test',
+        'future',
```
innerdvations, line 21: What will future be used for?

innerdvations:
There seems to be an issue with the github checks, if it happens again after updating the branch I can take a look, it could be related to one of the changes I made last week

innerdvations (approved):

PR #22267, opened 2024-11-25
enhancement(database): add repair manager with component deletion fix

innerdvations:
### What does it do?

- Adds a database repair manager 
- add a polymorph cleanup script for it
- call the repair function to clean up polymorph component_type when schema diff has changed

This PR is also serving as a proof-of-concept for the database repair function. Once decisions have been made about how it should work, I will add tests before merging, but I want it to stabilize first.

TODO: will add documentation in a later PR

### Why is it needed?

Cleans up orphaned data in deleted component tables on all environments without requiring a migration script

NOTE: anyone with an existing issue will still need to run `strapi.db.repair.removeOrphanTypes('component_type')` one time to fix it OR push a schema change to production to trigger this to run

### How to test it?

See https://github.com/strapi/strapi/issues/20931

- add components and dzs using components to your content types
- add data for those things
- delete the component used in the ct and dz
- everything should still work
- the content type _cmps table should not contain data for the deleted content type

There should not be any case where there is unintended data loss

### Related issue(s)/PR(s)

Alternative to: [v4 fix](https://github.com/strapi/strapi/pull/22207) [v5 fix](https://github.com/strapi/strapi/pull/22112)

Fixes https://github.com/strapi/strapi/issues/20931


packages/core/database/src/index.ts
```diff
@@ -119,6 +124,15 @@ class Database {
     this.entityManager = createEntityManager(this);
   }
 
+  async repair(options: RepairOptions) {
```
innerdvations, line 127: An alternative would be do pass in these options when calling `sync` and sync could do the normal db sync, then follow up with this

innerdvations: But this method _should_ (untested) work for any polymorphic relations table, so it could be useful for also cleaning up deleted content types.

I didn't attempt to clean everything because it would have been morecomplex to write the code to define what needs to be deleted, and I was also worried about potentially causing data loss or a performance impact.

packages/core/core/src/Strapi.ts
```diff
@@ -441,6 +441,11 @@ class Strapi extends Container implements Core.Strapi {
 
     await this.db.schema.sync();
 
+    // Remove orphaned components (entire types, not ids)
+    await this.db.repair({
```
alexandrebodin, line 445: Just a quick 1st comment we should avoid as much as possible to run this at startup time. We have to reduce the startup work as much as possible :)

Main question is what happens if it doesn't run until an action triggers it ? (CTB CRUD/CLI cmd/...)

packages/core/database/src/index.ts
```diff
@@ -119,6 +124,15 @@ class Database {
...
+    // We only clean up components on startup
+    if (options.removeOrphanMorphTypes) {
```
alexandrebodin, line 129: I would suggest to create a database repair service where we will implement different logic overtime instead of just have utils :)

packages/core/database/src/utils/remove-orphan-morph-types.ts
```diff
@@ -0,0 +1,70 @@
...
+ * - Morph types without corresponding metadata.
+ * - Morph types pointing to tables that no longer exist in the database.
+ *
+ * Note: This function does not check for orphaned IDs, only orphaned morph types.
+ *
+ * @param db - The database object containing metadata and a Knex connection.
+ * @param pivot - The name of the column in the join table representing the morph type.
+ */
+export const removeOrphanMorphTypes = async (db: Database, pivot: string) => {
+  const allTables = await db.dialect.schemaInspector.getTables();
```
alexandrebodin, line 18: Could this depend on the memory schema directly instead of calling the getTables again ?

packages/core/database/src/utils/remove-orphan-morph-types.ts
```diff
@@ -0,0 +1,70 @@
...
+        attribute.type === 'relation' &&
+        'joinTable' in attribute &&
+        'name' in attribute.joinTable &&
+        'target' in attribute &&
+        'pivotColumns' in attribute.joinTable &&
+        attribute.joinTable.pivotColumns.includes(pivot)
+    );
+
+    for (const attribute of attributes) {
+      if (!('joinTable' in attribute)) {
```
alexandrebodin, line 32: I supposed it's just to fix the typings ?

packages/core/database/src/utils/remove-orphan-morph-types.ts
```diff
@@ -0,0 +1,70 @@
...
+        continue;
+      }
+      const joinTableName = attribute.joinTable.name;
+
+      // Query distinct morph types from the join table
+      const morphTypes = await db.connection(joinTableName).distinct(pivot).pluck(pivot);
```
alexandrebodin, line 38: don't we need to add the schema here if user has configured one ? db.getConnection()

packages/core/database/src/utils/remove-orphan-morph-types.ts
```diff
@@ -0,0 +1,70 @@
...
+            return true;
+          }
+        })();
+
+        if (deleteComponentType) {
+          db.logger.warn(
+            `Removing invalid morph type "${morphType}" from table "${joinTableName}".`
+          );
+          try {
+            await db.connection(joinTableName).where(pivot, morphType).del();
```
alexandrebodin, line 59: same for the db.getConnection ?

alexandrebodin (changes requested):
A few high level comments :)

packages/core/database/src/utils/remove-orphan-morph-types.ts
innerdvations: TBH, it's not even fully necessary to do this; if something is not in the metadata, that's enough to know it should be deleted. I could remove this part.

packages/core/database/src/utils/remove-orphan-morph-types.ts
innerdvations: yeah, it narrows down the attribute type

packages/core/database/src/utils/remove-orphan-morph-types.ts
innerdvations: schema is added automatically to all non-raw queries

packages/core/database/src/utils/remove-orphan-morph-types.ts
alexandrebodin: yeah it should already be narrowed down above. I recommend you just loop over the attributes once and do the filter in the for loop cleanly instead

packages/core/database/src/utils/remove-orphan-morph-types.ts
alexandrebodin: I'm pretty sur it's not 🤔

packages/core/database/src/utils/remove-orphan-morph-types.ts
innerdvations: I think it's a typescript parser issue, it wasn't recognizing it from above. But I can keep playing with it and probably get that removed

alexandrebodin: You shoul use a type assertion for this 👍  helps the typescript parsing & also makes the code quite clean

packages/core/database/src/utils/remove-orphan-morph-types.ts
alexandrebodin: 👍

packages/core/database/src/index.ts
alexandrebodin: each function can be in it's own file but I would expose them from a single repair service to match with what is already done (schema, inspector, migration...)

packages/core/core/src/Strapi.ts
innerdvations: I'll keep looking for a more specific point we could run this, since it does add one query for every content type that uses components on startup which would be good to avoid.

I'm not sure exactly when this error occurs, but I assume startup: https://github.com/strapi/strapi/issues/20931#issuecomment-2268589861

If I can pinpoint where that happens, maybe it could go in a try/catch and only run this repair when it fails.

alexandrebodin: Looks like in this particular case it mostly related to CTB actions ?

I would consider that if someone manualy changes a schema they should run the function themselves (via CLI eventually) though

innerdvations: Updated to only run on schema diff changes rather than every startup

packages/core/database/src/repairs/index.ts
```diff
@@ -0,0 +1,52 @@
+import type { Database } from '..';
+import { removeOrphanMorphTypes } from './operations/remove-orphan-morph-types';
+import type { Operation, RepairManager } from './types';
+
+export type * from './types';
+
+export const createRepairManager = (db: Database): RepairManager => {
```
alexandrebodin, line 7: I'm not sure I see why we need the complexity of doing a generic list of operation when we are still statically importing them at the top of the file 🤔  

Can't this be simpler and avoid all the manual typings ? 

```
import * as operations from './operations';

createRepairManager = () =>  {

  // some internals to the service like stats, loggering, debugging ...

 return {
   removeOprhanMorphTypes:  operation.removeOprhanMorphTypes 
  
   // or just this if you need to wrap. 
   removeOprhanMorphTypes: {
      // some log
			return operations.removeOrphanMorphTypes()
			// some log
   ...
 }
}
```

alexandrebodin (commented):
Looks good, just trying to avoid adding premature complexity maybe

packages/core/database/src/repairs/index.ts
innerdvations: Sounds good. Much simpler now :)

packages/core/database/src/repairs/index.ts
```diff
@@ -0,0 +1,9 @@
+import { curry } from 'lodash/fp';
+import type { Database } from '..';
+import { removeOrphanMorphType as removeOrphanMorphTypeFunc } from './operations/remove-orphan-morph-types';
+
+export const createRepairManager = (db: Database) => {
+  return {
+    removeOrphanMorphType: curry(removeOrphanMorphTypeFunc)(db),
```
alexandrebodin, line 7: just a thought I believe lodash curry messes up the typings in som ecases (might be when there are many params I don't recall. Did you notice anything ?

packages/core/database/src/repairs/operations/remove-orphan-morph-types.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @param db - The database object containing metadata and a Knex connection.
+ * @param pivot - The name of the column in the join table representing the morph type.
+ */
+export const removeOrphanMorphType = async (
+  db: Database,
+  { pivot }: RemoveOrphanMorphTypeOptions
+) => {
+  db.logger.debug(`Removing orphaned morph type: ${JSON.stringify(pivot)}`);
+  const isRelationWithJoinTable = (
```
alexandrebodin, line 25: Can this pulled upward ?

alexandrebodin (approved):
Just super minor stuff, it looks good. Let me know if you need some help on QA.

packages/core/database/src/repairs/index.ts
innerdvations: I just checked where I added an asyncCurry elsewhere, and the problems are that it isn't aware of async functions, and it also provides ugly typings in general.

For now I'll add an internal util for asyncCurry similar to the other one, and then later on I think we should either add a @strapi/util to share between packages, or just start using Ramda 😆

Convly (approved):

Convly (approved):

PR #22273, opened 2024-11-26
fix: multiple requests locale

Marc-Roig:
### What does it do?

List View page was doing individual API requests for every entry.

i18n bulkActions and i18n localizations Cell was making use of `useDocument` hook, which was triggering those requests.
This PR proposes a fix to prevent those calls.

Before this PR, this was the network tab:
image

After this PR:
image

### Why is it needed?

### How to test it?

- Have more than 1 locale configured 
- Access ListView of a Content type with i18n enabled
- See individual api requests made for each entry

### Related issue(s)/PR(s)


packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
```diff
@@ -9,37 +9,24 @@ import { useGetLocalesQuery } from '../services/locales';
 interface LocaleListCellProps {
   documentId: string;
   collectionType: string;
+  localizations: { locale: string }[];
   locale: string;
   model: string;
```
remidej, line 14: Should you remove those other unused props from this type? Or will they be passed upstream even if we don't use them?

remidej (commented):
Less code, fewer requests, love it. I guess it was like this because we had initially removed the localizations array in v5?

haven't tested because it's a draft but it looks good to me

packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
Marc-Roig: absolutely!

Marc-Roig:
@remidej I have uncovered a bigger issue in the content manager, I will probably need another day to fix it

packages/core/content-manager/server/src/services/utils/__tests__/validatable-fields-populate.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      const result = getPopulateForValidation(uid as any);
+
+      expect(result).toEqual({
+        populate: {
+          componentAttrName: {
+            fields: ['subfield1'], // Only required fields in the component
+          },
+        },
+      });
+    });
```
Marc-Roig, line 84: Sooo, I am trying to create a function to select fields inside components, but it turns out there is a bug that prevents this. Until we fix that this PR will be blocked 😢 cc @remidej @markkaylor 

We use those fields populated inside the "localizations" attribute when bulk publishing locales

remidej: Hey Marc. So to merge your improvement we'd also need to update the bulk publish functionality? What's the bug exactly?

Marc-Roig: 1st issue that I am trying to fix: 
- the bulk publish UI needs required (and others) fields available inside `localizations` to know if a locale is publisheable or not

2nd bug I found while doing the 1st:
- we can not select fields inside component, I am not sure how we made it work in the first place (cc @jhoward1994 , maybe you can give us some hints to unblock this)
e.g of whats not possible:
```ts
populate: {
  compo: {
    fields: ['name'] // This fails
  },
}
```

Marc-Roig: I am going to take a look at the failing API tests too

Marc-Roig: For @jhoward1994 

the current getValidatableFieldsPopulate() function doesnt give a right output atm, for the address content type:

image

It tries to populate non populatable fields (city, postal_code), which makes the document service validation fail.

I have tried to propose a different approach, but fields can not be selected in components at the moment

Marc-Roig:
I opened another PR with an additional fix https://github.com/strapi/strapi/pull/22330
Didnt want to polute this pr even more

packages/core/content-manager/server/src/services/document-metadata.ts
```diff
@@ -227,13 +205,12 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
     version: DocumentVersion,
     { availableLocales = true, availableStatus = true }: GetMetadataOptions = {}
   ) {
-    // TODO: Ignore publishedAt if availableStatus=false, and ignore locale if
+    // TODO: Ignore publishedAt aif availableStatus=false, and ignore locale if
```
jhoward1994, line 208:
```suggestion
    // TODO: Ignore publishedAt if availableStatus=false, and ignore locale if
```

remidej:
@Marc-Roig your screenshot show 1 less mapEntriesToReleases call, does your change also have impact on that?

remidej (approved):
Tested (the cross locale validations too) and looks all good! Thanks for walking me through it

PR #22275, opened 2024-11-26
enhancement: disable preview link when changes are not saved

remidej:
### What does it do?

Disables preview link when there are unsaved changes in the edit view. Also adds a tooltip to explain why it's disabled

### Why is it needed?

Help users avoid losing content
### How to test it?

Make changes, don't save, button should be disabled and you should see the tooltip.

Save changes, the button should be enabled, and the tooltip should be gone

### Related issue(s)/PR(s)

replaces #22244


Marc-Roig (approved):

Marc-Roig:
much simpler than before 💯

PR #22283, opened 2024-11-27
fix: document actions being rendered multiple times

remidej:
### What does it do?

Prevents document actions from being rendered multiple times, by introducing a static optional "position" property on the component function, and using it to pre-filter the actions to render when possible.

### Why is it needed?

The issue is that the document action "components" do several things. They return static information, like their label, position, or icon. But they also do logic in the same component, to check if they should be enabled for example, or to define on onClick behavior. And that logic relies on hooks, like accessing context.

We keep a single document actions store, but they're split in 3 places: edit view side panel, edit view header, and list view table row. To get the right subset of actions in each location, we filter by the returned 

So here's the problem: on the edit view, we call every action component just do display the header ones, and we call every action component just to display the panel ones. So they all get called multiple times, no matter how many times they're actually rendered. This creates an issue when we add a useEffect to an action, because that effect will do everything twice. And that blocks us for what we're about to build.

Ideally we would split the static params from the dynamic React code. But this is the easiest fix that doesn't require a breaking change.

### How to test it?

Go into the code for UpdateAction, and add this code above the return:

```jsx
  const randomId = React.useId();
  console.log('randomId', randomId);
```

See the log in your browser: you should only see a single ID. You can see the previous bug if you remove the `UpdateAction.position = 'panel'` line, there would be two different IDs in the console.

You can do the same checks for header actions (e.g. history) and table row actions. Also check that the delete action is visible in both the list view rows and the edit view header.

### DIFF
--- a/packages/core/content-manager/admin/src/content-manager.ts
+++ b/packages/core/content-manager/admin/src/content-manager.ts
@@ -3,6 +3,7 @@ import { INJECTION_ZONES } from './components/InjectionZone';
 import { PLUGIN_ID } from './constants/plugin';
 import {
   DEFAULT_ACTIONS,
+  type DocumentActionPosition,
   type DocumentActionDescription,
 } from './pages/EditView/components/DocumentActions';
 import {
@@ -94,6 +95,7 @@ interface DocumentActionComponent
     | 'publish'
     | 'unpublish'
     | 'update';
+  position?: DocumentActionDescription['position'];
 }
 
 interface HeaderActionProps extends EditViewContext {}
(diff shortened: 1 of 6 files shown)


markkaylor:
Hey I haven't pulled to try and understand the issue yet but I'm a bit confused if something is broken or not. 

> This creates an issue when we add a useEffect to an action, because that effect will do everything twice. 

Aren't some actions already using useEffect. Just to clarify the effect runs twice but does that cause any real issue? (not saying we should keep it that way just curious)

> And that blocks us for what we're about to build.

How does it block us from building what?

remidej:
>Aren't some actions already using useEffect?

@markkaylor I don't think they're using useEffect directly, but many use RTK query hooks, so they probably have effects under the hood. But for those queries it's not an issue, since RTK acts as the central cache and won't trigger the requests multiple times.

>How does it block us from building what?

We want to add the cmd+enter shortcut to update an entry. We do it in the UpdateAction, so it's colocated with the update logic. But because it's rendered twice (or at least _called_ twice), we have 2 listeners for each cmd+enter event, so that would trigger updates twice. I guess we could live with that but it doesn't seem right to me

packages/core/content-manager/admin/src/content-manager.ts
```diff
@@ -209,7 +211,23 @@ class ContentManagerPlugin {
...
+           * which wasn't visible but created issues with useEffect for instance.
+           * The response should still be filtered by the position, as the static property is new
+           * and not mandatory to avoid a breaking change.
+           */
+          if (position) {
+            return this.documentActions.filter(
+              (action) =>
+                action.position == undefined || [action.position].flat().includes(position)
+            );
+          }
```
Marc-Roig, line 227: Can end users register actions? (still unfamiliar with the api and how actions are supposed to be registered :D )

remidej: Yes they can, that's why I have to make the .position key on the component optional, because we can't make it a requirement now

remidej: so this is just about fixing the bug when possible, but it will still occur for community registered document actions that don't have it

Marc-Roig (approved):

Marc-Roig:
logging a random id in the update action seems to work ok from my side now! 🚀

Issue #21018, opened 2024-08-15
Cannot disable GraphQL playground outside of production environment


derrickmehaffy:
### Required System information

- Node.js version: v20
- NPM version: Yarn v1
- Strapi version: 4.25.8
- Database: SQLite
- Operating system: Linux Mint 21.2
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When running Strapi with a different `NODE_ENV` besides production it's not possible to disable the GraphQL playground using the `playgroundAlways: false`. Likewise when running in production it should be possible to enable the playground should the user choose too.

### Steps to reproduce the behavior

1. Create a Strapi application
2. Install the GraphQL plugin
3. Configure the GraphQL plugin with `playgroundAlways: false`
4. Start the application in development mode
5. See that the application is still offering the playground on the /graphql endpoint

### Expected behavior

Four states should be possible:

- Production Environment without the Playground (Default)
- Production Environment with the Playground (`playgroundAlways: true`)
- Non-Prod Environment with the Playground (Default)
- Non-Prod Environment without the Playground (`playgroundAlways: false`)

### Screenshots

N/A

### Code snippets

N/A

### Additional context

Related to TID7673

innerdvations:
I tested v4 and that doesn't seem to be how it worked before. I added a quick fix that I didn't have time for before releasing Strapi 5, let me know if it makes sense and I'll try to get it pushed through: https://github.com/strapi/strapi/pull/22285

PR #22285, opened 2024-11-27
enhancement: apollo landing page config

innerdvations:
### What does it do?

- marks `playgroundAlways` as deprecated, to be removed in Strapi v6
- adds graphql plugin `landingPage` config value that accepts either
  -  a boolean value
  - a user-defined function that returns either
    - a boolean
    - an `ApolloServerPlugin` that implements `renderLandingPage` (see https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages )

with the meanings:
- `true` means show the "local landing page" (formerly playground)
- `false` means to show the "production landing page" (the generic page with the graphql url but no 'playground')
- An `ApolloServerPlugin` means to use that directly instead of either of the built-in options
- if both `playgroundAlways` and `landingPage` values are left undefined, use `false` for production environments and `true` for all other environments

### Why is it needed?

- Apollo removed 'playground' in Apollo v4 so it's a bad name for it
- We need a way to disable it everywhere, and `playgroundAlways: false` doesn't logically make sense, it would just mean to not always enable it
- This adds the ability for users to provide whatever Apollo landing page plugin they wish, or dynamically determine whether to show it or not, so in theory they can put whatever they want there

QUESTION: Should this be called a new `feat` so it triggers a minor version since it flags playgroundAlways as deprecated? (although we won't change how it actually works until Strapi 6)

### How to test it?

- play around with the `playgroundAlways` values and `landingPage` values and see that it works as expected
- note that `playgroundAlways:true` will always take precedence to avoid a breaking change

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/21018

Docs PR: https://github.com/strapi/documentation/pull/2301


derrickmehaffy:
EXP version: `0.0.0-experimental.cb74730ce5154c26404d4dccca14976a22319002`

derrickmehaffy:
Testing with the experimental it looks like we have a CSP issue not being able to load the landing page from apollo's CDN: 
image

Similar to the suggestion I made to the content squad we should probably dynamically inject the required CSP setting such as what @Marc-Roig did here: https://github.com/strapi/strapi/pull/22138

---

Likewise even though I am in development mode with the configuration example shown in the docs preview, it's giving me the production landing page (after I properly, manually, configured the CSP settings for script-src and manifest-src:

```ts
// Path: ./config/plugins.ts

export default {
  graphql: {
    config: {
      endpoint: "/graphql",
      shadowCRUD: true,
      depthLimit: 7,
      amountLimit: 100,
      landingPage: (ctx) => ctx.env === "development",
      apolloServer: {
        tracing: true,
      },
    },
  },
};
```

```ts
export default [
  "strapi::logger",
  "strapi::errors",
  {
    name: "strapi::security",
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          "connect-src": ["'self'", "https:"],
          "img-src": [
            "'self'",
            "data:",
            "blob:",
            "market-assets.strapi.io",
            "apollo-server-landing-page.cdn.apollographql.com",
          ],
          "media-src": [
            "'self'",
            "data:",
            "blob:",
            "market-assets.strapi.io",
            "apollo-server-landing-page.cdn.apollographql.com",
          ],
          "script-src": [
            "'self'",
            "'unsafe-inline'",
            "cdn.jsdelivr.net",
            "apollo-server-landing-page.cdn.apollographql.com",
          ],
          "manifest-src": ["apollo-server-landing-page.cdn.apollographql.com"],
          upgradeInsecureRequests: null,
        },
      },
    },
  },
  "strapi::cors",
  "strapi::poweredBy",
  "strapi::query",
  "strapi::body",
  "strapi::session",
  "strapi::favicon",
  "strapi::public",
];
```

According to this: https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages I should be seeing the non-prod landing page, if further configuration is required we should probably give a few more examples in the docs.

derrickmehaffy (changes requested):
see my larger note above

innerdvations:
For the CORS settings, they are set dynamically in the security exceptions when landingPage is enabled, so for production they are not set and I believe they weren't set previously either in those cases

For the `landingPage: (ctx) => ctx.env === "development",` it's actually a strapi object being passed in, so that resolves to false, which is why it shows the production page. If you use `process.env.NODE_ENV === 'development';` it should work as expected to only show it on development but no other environments.

Can you confirm?

I will make sure when I add documentation that it explains the method for easily enabling the CORS settings for users who want to add the exceptions even in production.

innerdvations:
> Code looks fine, I tested it a bit, just to confirm a few behaviors I noticed, 1 the user needs to add security middleware to see the default production landing page.

I'm 95% sure this is correct, but I didn't confirm that's how it was working on v4. My thought was that on a production environment, we shouldn't be adding security exceptions unless the user does so explicitly, but since it's limited to /graphql it might be ok.

> when the landingPage is set to `false` we are seeing the default production landing page even when it's a development environment?
Yep, that's the main reason for this change, to provide a way to disable it even in development or other environments. Previously there was no way to do so.

Bassel17 (approved):

packages/plugins/graphql/server/src/services/utils/playground.ts
```diff
@@ -1,12 +1,17 @@
 import { Context } from '../types';
 
+// TODO: deprecate and remove this because it is only used to determine if we need helmet security exceptions
+// stores the state of the Apollo landingPage (playground)
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export default (ctx: Context) => {
+  let enabled = false;
+
   return {
+    setEnabled(val: boolean) {
```
Convly, line 10: Should we mark those functions as deprecated/experimental if they're meant to be removed later on?

Convly: (I'm referring to the TODO comment above)

packages/plugins/graphql/server/src/bootstrap.ts
```diff
@@ -20,6 +20,86 @@ const merge = mergeWith((a, b) => {
...
+      return localLanding();
+    }
+    if (result === false) {
+      return prodLanding();
+    }
+    strapi.log.debug('Apollo landing page: user-defined');
+    return result;
+  };
+
+  // DEPRECATED, remove in Strapi v6
```
Convly, line 64: Should we update the documentation somewhere to add a callout about this?

Convly: Or is the warning enough?

innerdvations: I have a docs PR too: https://github.com/strapi/documentation/pull/2301

packages/plugins/graphql/server/src/services/utils/playground.ts
innerdvations: We don't have a replacement for them at the moment, it's probably going to take a refactoring to be able to inject exceptions into the security middleware.  So for now I'm just going to leave it alone even though it's a bit ugly, it would be a lot of change for this one little problem.

derrickmehaffy:
Building exp to test with: https://github.com/strapi/strapi/actions/runs/12601132947

derrickmehaffy:
exp: `0.0.0-experimental.7bef52cb5066b2d3918afde3c5a60ded8754fdf9`

derrickmehaffy:
I still have a few major concerns about this implementation though I may just be lacking some context here. At worst I fear some of these changes will be breaking changes since they require config changes to work as it did before.

- Since we require modifications to the middlewares for this to work properly it's going to break any current application
- Using the example from the documentation does not work as intended (not sure where ctx comes from in the config but it's undefined)

I would advise we automatically inject the proper security middleware exceptions using the same method that @Marc-Roig did whenever the landing page is enabled: https://github.com/strapi/strapi/pull/22138

derrickmehaffy (approved):

innerdvations:
Just tested again with @derrickmehaffy , this doesn't cause any breaking changes with the original config value `playgroundAlways` and adds support for `landingPage` correctly. The only issue is that security exceptions do not happen automatically when in production for some reason, but that was also the case with `playgroundAlways` so I will go ahead and merge this, then document how to add the security exceptions manually on production, and later we can revisit the idea of adding them automatically, which is what happens on other environments.

PR #22290, opened 2024-11-27
fix: add condition to basePath (v5)

### LINKED ISSUE #22284: [V5:19330] Deployment on different servers not working
**Dupe of #19330 for tracking V5 fix**

Required System information
Launched In: 166 ms
Environment: development
OS: darwin-arm64
Strapi Version: 4.19.0
Node/Yarn Version: npm/10.2.3 node/v20.10.0 darwin arm64 workspaces/false
Edition: Community
Database: sqlite

Describe the bug
I'm following the [deployment guide](https://docs.strapi.io/dev-docs/admin-panel-customization#different-servers) for different servers (I want to deploy the admin panel - react app - to Netlify and the server to a droplet or similar). Still, the generated index.html file contains invalid paths to the generated js files.

Steps to reproduce the behavior
Create a Strapi example project using the quick start guide - https://docs.strapi.io/dev-docs/quick-start
Configure the server.js and admin.js files as described in the different server configuration guide - https://docs.strapi.io/dev-docs/admin-panel-customization#different-servers
Run NODE_ENV=production npm run build
The generated paths to the js files contain an extra slash, generating 404 errors:
src="//runtime~main.35ff75f1.js"
src="//main.19186bea.js"

Expected behavior
Create the index.html with the correct paths with only one slash to access the admin site using the http://yourfrontend.com/ URL. I think in this context the proxy recommendation not apply because this is a front-end app with the index.html file and js files, must work with any static hosting provider.

butcherZ:
### What does it do?

setting root path to `/` when serve the admin build folder on a separate server

### Why is it needed?

fixes #22284 

### How to test it?

- create a new strapi app with `npx create-strapi-app@0.0.0-experimental.0588b9a4effba287cf26bd50e23baa14cdaf1616`
- add `url: 'http://localhost:1337'`, to ./config/server.js
- add `url: 'http://localhost:3000'`, and `serverAdminPanel: false` to ./config/admin.js([more info here](https://docs.strapi.io/dev-docs/admin-panel-customization/deployment))
- run yarn build
- copy and move the build folder outside of the strapi project
- `cd` to build folder, run ` npx serve -l 3000`
- admin panel should be accessible at `http://localhost:3000` without any errors shown 

### Related issue(s)/PR(s)

#22284 
DX-1768

### DIFF
--- a/packages/core/strapi/src/node/create-build-context.ts
+++ b/packages/core/strapi/src/node/create-build-context.ts
@@ -135,10 +135,11 @@ const createBuildContext = async <TOptions extends BaseOptions>({
   const features = strapiInstance.config.get('features', undefined);
 
   const { bundler = 'vite', ...restOptions } = options;
+  const serveAdminPanel = strapiInstance.config.get('admin.serveAdminPanel', true);
 
   const buildContext = {
     appDir,
-    basePath: `${adminPath}/`,
+    basePath: serveAdminPanel ? `${adminPath}/` : '/',
     bundler,
     customisations,
     cwd,


packages/core/strapi/src/node/create-build-context.ts
```diff
@@ -135,10 +135,11 @@ const createBuildContext = async <TOptions extends BaseOptions>({
   const features = strapiInstance.config.get('features', undefined);
 
   const { bundler = 'vite', ...restOptions } = options;
+  const serveAdminPanel = strapiInstance.config.get('admin.serveAdminPanel', true);
 
   const buildContext = {
     appDir,
-    basePath: `${adminPath}/`,
+    basePath: serveAdminPanel ? `${adminPath}/` : '/',
```
innerdvations, line 142: Should this be the same as your v4 version so users can set the admin.url to use?

alexandrebodin: I completly change the way this is created in my reverse proxy PR can you check if the issue is already fixed there instead ?

alexandrebodin: I won't port my PR to v4 so you can def go ahead. for this one it's fine too, I was more concerned that my PR was breaking the fix. you can merge it if it fixes what it says it does 💯  I don't know when my fix will be merged we need more testers

butcherZ: @alexandrebodin I will test your PR throughly along with mine tomorrow!

butcherZ:
closing this PR as #21914 fixed the problem mentioned in #22284

PR #22304, opened 2024-11-28
ci: fix docs_build

innerdvations:
### What does it do?

allows docs_build to run by making it dependant on build and installing its deps

### Why is it needed?

docs_build isn't actually able to run because it fails to install dependencies and build the monorepo

### How to test it?

it should run and pass here

### Related issue(s)/PR(s)


markkaylor (approved):
oh man, thank you!

PR #22309, opened 2024-11-29
ci: enable dependabot for security upgrades

innerdvations:
### What does it do?

- allows dependabot to start opening PRs again (every Sunday)
- updates package groups
- for now, ignore all updates except security updates
  - later we can revisit if we want to allow non-security upgrades, but generally in the past it has caused more effort and instability than usefulness

### Why is it needed?

- saves time and effort and improves response time in patching security issues in dependencies

### How to test it?

Can't easily test it until it's merged and we see that dependabot starts submitting PR

### Related issue(s)/PR(s)

DX-1798

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,7 @@ updates:
 
       jest:
         patterns:
+          - '@swc/jest'
           - 'jest'
           - 'jest-*'
           - 'supertest'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,8 +8,9 @@ updates:
           - '@babel/*'
           - 'babel-*'
 
-      eslint:
+      eslint-prettier:
         patterns:
+          - 'prettier'
           - 'eslint'
           - 'eslint-*'
           - '@eslint/*'
@@ -23,12 +24,6 @@ updates:
       playwright:
         patterns:
           - '@playwright/*'
-
-      prettier:
-        patterns:
-          - 'prettier'
-          - 'eslint-config-prettier'
-          - 'eslint-plugin-prettier'
 
       react:
         patterns:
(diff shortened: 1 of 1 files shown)


innerdvations:
@alexandrebodin As far as I know it wasn't disabled at the github level, we just set PRs to 0 here to temporarily disable it. But could you check that it's still enabled at https://github.com/strapi/strapi/settings/security_analysis ?

.github/dependabot.yml
```diff
@@ -61,26 +85,22 @@ updates:
...
-      - dependency-name: '@babel/*'
-        update-types:
-          - 'version-update:semver-major'
-          - 'version-update:semver-minor'
-
+      # only accept security updates
       - dependency-name: '*'
         update-types:
           - 'version-update:semver-patch'
+          - 'version-update:semver-minor'
```
alexandrebodin, line 95: you are literally saying to ingore all types of upgrades here I think :)

alexandrebodin: Ok in that case we should entirely remove snyk so we don't have 2 tools working on security dependencies. 

It's definitely unclear that this config will still allow security bumps

innerdvations: I did some more research and it seems like this doesn't work anymore and dependabot v2 made "security updates only" only available through the github ui :(

So I will remove this ignore statement and put back the one for babel, and we'll just have to manage security vs non-security updates in the github settings. I personally prefer that for now, internally upgrading packages for non-security updates doesn't offer us anything; if we need a feature from a newer package we'll just upgrade it manually.

innerdvations: And in that case, this PR doesn't appear to do anything except add some groupings; once security updates are enabled in the github UI, in theory they ignore the PR limit set in dependabot.yml

innerdvations:
Closing this for now since updates have been re-enabled in the github settings. I will revisit this later to enable updates for specific tooling packages that we regularly update in sync, such as playwright, jest, babel, etc

PR #22311, opened 2024-11-29
enhancement: cmd+enter and ctrl+enter to save entry

remidej:
### What does it do?

Adds command shortcuts that trigger a save on the edit view:
- cmd+enter
- ctrl+enter

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
@@ -121,18 +121,6 @@ const EditViewPage = () => {
     return transformDocument(schema, components)(form);
   }, [document, isCreatingDocument, isSingleType, schema, components]);
 
-  const validate = React.useCallback(
-    (values: Record<string, unknown>, options: Record<string, string>) => {
-      const yupSchema = createYupSchema(schema?.attributes, components, {
-        status,
-        ...options,
-      });
-
-      return yupSchema.validate(values, { abortEarly: false });
-    },
-    [components, schema?.attributes, status]
-  );
-
   if (hasError) {
     return <Page.Error />;
   }
@@ -180,7 +168,14 @@ const EditViewPage = () => {
         disabled={hasDraftAndPublished && status === 'published'}
         initialValues={initialValues}
         method={isCreatingDocument ? 'POST' : 'PUT'}
-        validate={validate}
+        validate={(values: Record<string, unknown>, options: Record<string, string>) => {
+          const yupSchema = createYupSchema(schema?.attributes, components, {
+            status,
+            ...options,
+          });
+
+          return yupSchema.validate(values, { abortEarly: false });
+        }}
         initialErrors={location?.state?.forceValidation ? validateSync(initialValues, {}) : {}}
       >
         {({ resetForm }) => (
--- a/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
@@ -121,6 +121,18 @@ const EditViewPage = () => {
     return transformDocument(schema, components)(form);
   }, [document, isCreatingDocument, isSingleType, schema, components]);
 
+  const validate = React.useCallback(
+    (values: Record<string, unknown>, options: Record<string, string>) => {
+      const yupSchema = createYupSchema(schema?.attributes, components, {
+        status,
+        ...options,
+      });
+
+      return yupSchema.validate(values, { abortEarly: false });
+    },
+    [components, schema?.attributes, status]
+  );
+
   if (hasError) {
     return <Page.Error />;
   }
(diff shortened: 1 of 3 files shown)


Convly:
cool

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -121,6 +121,18 @@ const EditViewPage = () => {
...
+    (values: Record<string, unknown>, options: Record<string, string>) => {
+      const yupSchema = createYupSchema(schema?.attributes, components, {
+        status,
+        ...options,
+      });
+
+      return yupSchema.validate(values, { abortEarly: false });
+    },
+    [components, schema?.attributes, status]
+  );
```
markkaylor, line 134: Doesn't useDoc already create and return the `validate` function?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -768,113 +768,147 @@ const UpdateAction: DocumentActionComponent = ({
...
+    navigate,
+    params,
+    rawQuery,
+    resetForm,
+    setErrors,
+    setSubmitting,
+    toggleNotification,
+    update,
+    validate,
+  ]);
```
markkaylor, line 881: 😳

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
remidej: True we could investigate if they can be shared. We need validation at the form level but we could try to plug the useDocument 🤔 

I think I'll remove this change from this PR and look into it separately

markkaylor (commented):
Cool looks good to me, I can test it this afternoon

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -180,7 +168,14 @@ const EditViewPage = () => {
...
         method={isCreatingDocument ? 'POST' : 'PUT'}
-        validate={validate}
+        validate={(values: Record<string, unknown>, options: Record<string, string>) => {
+          const yupSchema = createYupSchema(schema?.attributes, components, {
+            status,
+            ...options,
+          });
+
+          return yupSchema.validate(values, { abortEarly: false });
+        }}
```
markkaylor, line 178: Sorry, previously I was asking if it is possible to use the validate function useDocument exposes:

https://github.com/strapi/strapi/blob/46c3b79e02bbbbb309f7420bcace00a271149d2e/packages/core/content-manager/admin/src/hooks/useDocument.ts#L127-L168

If not, do you know what the difference between the too is?

remidej:
I just pushed 2 fixes:
- don't register cmd+s or ctrl+s when using Arc
- don't trigger the update is the form doesn't have unsaved changes

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
remidej: I think here we should use the validate function coming from the Form context, because it's the one that is tied to the form inputs, in order to display validation errors within the fields.

So I think the change would need to be done higher up, because it comes from something like that:

```js
<Form validation={someValidateFn} />
```

and I agree with you that it sounds like we could use the useDocument fn there. But it's also not related to cmd+enter to save, so I don't want to do it in this PR. It sounds like it could be part of a larger topic to avoid rerenders and reduce tech debt

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -768,113 +768,166 @@ const UpdateAction: DocumentActionComponent = ({
...
+     * Detect the Arc browser specifically, because it registers CMD-S/CTRL-S for the quite common
+     * toggle sidebar action. So ignore that shortcut for it to avoid conflicts.
+     */
+    const isArcBrowser =
+      getComputedStyle(window.document.documentElement).getPropertyValue('--arc-palette-title')
+        .length > 0;
+
+    if (!isArcBrowser) {
+      keysToListenTo.push('s');
+    }
```
markkaylor, line 902: I can get on board with merging this..._but_, what if we just don't include the ctrl+s / cmd+s part for now and do it in a follow up _if_ that's what people want.

We can maintain one command for now that works for all browsers and all flavors of browsers. My original comment about cmd+s was just that it was the most intuitive thing. I think if users learn what cmd+enter does they will be happy with that.

Adding cmd+s later is easy. If we add it now, we are stuck with maintaining both commands for the same action until a major release.

I'm particularly not crazy about the special treatment of the hype train that is Arc.

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
markkaylor: 👍  That makes sense. Thanks.

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
remidej: Let me check with Yannis 😁 As I said on Slack, I'm not a fan of cmd+s on the web, so I'm just trying to please everyone but I'd be happy to remove it. I'm not sure we need to worry about backward compatibility for this though

markkaylor: >I'm not sure we need to worry about backward compatibility

markkaylor: > trying to please everyone

Nahhhh let's do what's best for the codebase and the product 🎈

yanniskadiri: From a product perspective, it makes sense to start with Cmd+Enter as the primary save shortcut. It’s straightforward, works across all browsers, and avoids potential conflicts. We can gather user feedback after release and, if there’s significant demand, add Cmd+S later in a follow-up.

Issue #22295, opened 2024-11-28
Failed to turn off `Restricted Access` for Documentation Plugin


martin-luo:
### Required System information

- Node.js version: 20.10.0
- NPM version: 10.5.0
- Strapi version: 5.4.2
- Database:Postgresql
- Operating system: macOS 15.1.1 (24B91)
- Is your project Javascript or Typescript: TypeScript

### Steps to reproduce the behavior

1. Go to http://localhost:1337/admin/settings/documentation
2. Click on `ON` in `Settings -> Restricted Access`
3. Type in a valid password
4. Click 'Save'
5. Click on `OFF` in `Settings -> Restricted Access`
6. Click 'Save'
7. See error

### Expected behavior

Restricted Access is turned off and user can visit documentations, e.g. http://localhost:1337/documentation/v1.0.0 without entering the password

### Screenshots
![Documentation setting page with error](https://github.com/user-attachments/assets/1014b958-859a-4152-b9b3-385aa1d5444b)
![Documentation setting page with error and network tab details](https://github.com/user-attachments/assets/76177f96-37ad-4dfb-a74c-66200f290e0c)

cwilso03:
A workaround is to manually (through curl, postman, etc.) submit the `PUT` request to update the settings without including the empty `password` property, like so:

```
PUT http://localhost:1337/documentation/updateSettings
Content-Type: application/json
Authorization: bearer <token value>

{
  "restrictedAccess": false
}
```

To get the JWT, send a `POST` to the admin login URL:

```
POST http://localhost:1337/admin/login
Content-Type: application/json

{
  "email": "youremail@yourdomain.com"
  "password": "yourpassword"
}
```

The JWT is in the response body's `data.token` property.

jigyarth-searce:
Yes I also faced this issue. My workaround was to use a chrome extension and intercept the request, inserting password to the request body. I used a chrome extension `[Temper dev] for intercepting the request(https://chromewebstore.google.com/detail/tamper-dev/mdemppnhjflbejfbnlddahjbpdbeejnn)`

Your new curl should be 
`curl 'http://localhost:1337/documentation/updateSettings' \
  -X 'PUT' \
  -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8,gu;q=0.7' \
  -H 'Connection: keep-alive' \
[...]

PR #22313, opened 2024-11-29
fix: add conditional rendering of password prompt

AltayAkkus:
### What does it do?

In the original version the password prompt is only rendered when the state is equal to ON.
This leads to the problem described in #22295, when turning the restricted OFF no password is being passed to the endpoint, which results in restricted access being ON forever.
### Why is it needed?

To solve #22295

### How to test it?

1. Open http://localhost:1337/admin/settings/documentation
2. Turn the Restricted access ON
3. Save
4. Turn the Restricted access OFF
4. Save
5. Be happy!

### Related issue(s)/PR(s)

#22295


AltayAkkus:
found a better way, sorry

Issue #22281, opened 2024-11-27
Boolean type should use translations


dominikjasek:
## Bug report

![image](https://github.com/user-attachments/assets/54f367a6-e7e9-4cb4-ac42-bcffaf4e8c8b)

My guess is that this [piece of code](https://github.com/strapi/strapi/blob/d65d44102fd32871728c0d74ec4f2519b7cc0a16/packages/core/content-manager/admin/src/pages/ListView/components/TableCells/CellValue.tsx#L16) should be extended for `'boolean'` content type

### Required System information

Launched In: 754 ms
Environment: development
OS: darwin-arm64
Strapi Version: 5.2.0
Node/Yarn Version: pnpm/9.6.0 npm/? node/v20.8.0 darwin arm64
Edition: Community
Database: postgres

### Steps to reproduce the behavior

1. Use admin language different than english
2. Create boolean type
3. Go to content manager
4. See that in table view translations of boolean are not working

### Expected behavior

Values should be using translations like [here](https://github.com/strapi/strapi/blob/d65d44102fd32871728c0d74ec4f2519b7cc0a16/packages/core/admin/admin/src/components/FormInputs/Boolean.tsx#L26).

![image](https://github.com/user-attachments/assets/9a638911-a012-4604-b865-04ec840b0937)

PR #22314, opened 2024-11-29
fix: Add i18n for boolean cell values

AltayAkkus:
### What does it do?

Add a conditional boolean field types, which uses i18n.

### Why is it needed?

So users with a non-default language can see read the boolean field in their chosen language.
### How to test it?

1. Create a new content type with a boolean
2. Create an entry for that content type
3. Look at the list of entries

### Related issue(s)/PR(s)

#22281


hanpaine:
Hey @AltayAkkus cheers for the contribution 🙌  we'll take a look soon 🙏

PR #22324, opened 2024-12-02
fix(preview): page title and disabled link

### LINKED ISSUE #22298: [Preview] We should match the title of the entry and the title in the preview

- Edition: Community
- OS: Mac
- Environment: Local

### Describe the bug

When opening the preview of an entry untitled, we should display “untitled”

1. Go to an entry untitled
2. Open the preview
3. See that nothing is displayed

### Expected behavior

We should display “Untitled”

### Screenshots

https://github.com/user-attachments/assets/5e36c133-0f19-4221-bf6a-1405b1a4c8ae

### LINKED ISSUE #22300: [Preview] We should see the whole title when hovering on the truncated entry title in preview mode

- Edition: Community
- OS: Mac
- Environment: Local

### Describe the bug

We should see the whole title of an entry when hovering on the truncated entry title in preview mode

1. Go to an entry with a long title
2. Open the preview and see the truncated entry title
3. Hover on the title and you won’t see anything

### Expected behavior

The whole title should be shown on hover

### Screenshots

https://github.com/user-attachments/assets/b9fe0465-ff9b-44da-9d81-37c7a7016654

remidej:
### What does it do?

Adds 3 fixes related the the preview page header title, reported in the blitz:

- adds a tooltip, so we have a way to see the full title when it's cropped because it's too long. I rely on the native HTML title to trigger the browser's default tooltip. I checked with @lucasboilly and he approved it, our DS tooltip isn't great at handling long content over multiple lines. It's also much simpler to implement.
- makes sure the title is the same as the one from the edit view by copying the logic from there
- properly disables the link. TIL the disabled tag does nothing for anchors! so I combined [this workaround](https://stackoverflow.com/questions/13955667/disabled-href-tag#comment119178426_27057625) with [this one](https://stackoverflow.com/a/46665946/3661792) for the not allowed cursor

### How to test it?

- write long content in the main field, open preview, the title is truncated. Hover it and you'll see the full title in the browser's tooltip
- make sure the preview page shows the same header title as the edit view:
  - collection type with non-id main field: main field
  - collection type with id main field: "Untitled"
  - single type with non-id main field: main field
  - single type with id main field: content type display name
- make a change in the edit view and don't save. This preview button should be disabled with 🚫 cursor. Click it, nothing should happen. Save and it's enabled again
 
### Related issue(s)/PR(s)

- fixes #22300
- fixes #22298

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
@@ -58,8 +58,6 @@ const PreviewSidePanel: PanelComponent = ({ model, documentId, document }) => {
     trackUsage('willNavigate', { from: pathname, to: destinationPathname });
   };
 
-  const isDisabled = isModified;
-
   return {
     title: formatMessage({ id: 'content-manager.preview.panel.title', defaultMessage: 'Preview' }),
     content: (
@@ -78,9 +76,9 @@ const PreviewSidePanel: PanelComponent = ({ model, documentId, document }) => {
               to={{ pathname: 'preview', search: stringify(query, { encode: false }) }}
               onClick={trackNavigation}
               width="100%"
-              disabled={isDisabled}
-              pointerEvents={isDisabled ? 'none' : undefined}
-              tabIndex={isDisabled ? -1 : undefined}
+              disabled={isModified}
+              pointerEvents={isModified ? 'none' : undefined}
+              tabIndex={isModified ? -1 : undefined}
             >
               {formatMessage({
                 id: 'content-manager.preview.panel.button',
--- a/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
@@ -61,33 +61,31 @@ const PreviewSidePanel: PanelComponent = ({ model, documentId, document }) => {
   return {
     title: formatMessage({ id: 'content-manager.preview.panel.title', defaultMessage: 'Preview' }),
     content: (
-      <Flex gap={2} width="100%">
-        <ConditionalTooltip
-          label={formatMessage({
-            id: 'content-manager.preview.panel.button-disabled-tooltip',
-            defaultMessage: 'Please save to open the preview',
-          })}
-          isShown={isModified}
-        >
-          <Box cursor="not-allowed" flex="auto">
-            <Button
-              variant="tertiary"
-              tag={Link}
-              to={{ pathname: 'preview', search: stringify(query, { encode: false }) }}
-              onClick={trackNavigation}
-              width="100%"
-              disabled={isModified}
-              pointerEvents={isModified ? 'none' : undefined}
-              tabIndex={isModified ? -1 : undefined}
-            >
-              {formatMessage({
-                id: 'content-manager.preview.panel.button',
-                defaultMessage: 'Open preview',
-              })}
-            </Button>
-          </Box>
-        </ConditionalTooltip>
-      </Flex>
+      <ConditionalTooltip
+        label={formatMessage({
+          id: 'content-manager.preview.panel.button-disabled-tooltip',
+          defaultMessage: 'Please save to open the preview',
+        })}
+        isShown={isModified}
+      >
+        <Box cursor="not-allowed" width="100%">
+          <Button
+            variant="tertiary"
+            tag={Link}
+            to={{ pathname: 'preview', search: stringify(query, { encode: false }) }}
+            onClick={trackNavigation}
+            width="100%"
+            disabled={isModified}
+            pointerEvents={isModified ? 'none' : undefined}
+            tabIndex={isModified ? -1 : undefined}
+          >
+            {formatMessage({
+              id: 'content-manager.preview.panel.button',
+              defaultMessage: 'Open preview',
+            })}
+          </Button>
+        </Box>
+      </ConditionalTooltip>
     ),
   };
 };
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
```diff
@@ -58,6 +58,8 @@ const PreviewSidePanel: PanelComponent = ({ model, documentId, document }) => {
     trackUsage('willNavigate', { from: pathname, to: destinationPathname });
   };
 
+  const isDisabled = isModified;
```
markkaylor, line 61: why do you need to do create a new variable? Couldn't you just use isModified wherever you need things disabled?

remidej: I did that because I check in [3 different places](https://github.com/strapi/strapi/pull/22324/files#diff-a3f6f0d2c32a2510e10592dc1904c0d9b8770cb2855aff5efd2375e233bac15fR81-R83) if the button is disabled, so if in the future there are more reasons to disable then we'd need to change the check everywhere.

We don't have that problem for now though, so I'll update to simplify 👍

markkaylor: Gotcha 👍 , clever. But I agree with removing it for now or leaving a comment to explain your intention (otherwise next time I'm in that file 🧹 )

markkaylor (commented):
Tested it out and all looks good for the fixes described here:

✅ write long content in the main field, open preview, the title is truncated. Hover it and you'll see the full title in the browser's tooltip
✅ make sure the preview page shows the same header title as the edit view:
collection type with non-id main field: main field
collection type with id main field: "Untitled"
single type with non-id main field: main field
single type with id main field: content type display name
✅ make a change in the edit view and don't save. This preview button should be disabled with 🚫 cursor. Click it, nothing should happen. Save and it's enabled again

### Other comments
I noticed the sidebar is showing up now? Is that intentional? I don't see it in the designs but not sure if a decision was made after a blitz or something?

preview-sidebar

The truncated text + tooltip works really nicely for preview, but in the EditViewPage it sorta breaks the header and you aren't able to access the actions menu.

I don't think that should block the PR though, what should we do? Create a ticket in JIRA?

preview-cm-long-text

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -93,7 +94,7 @@ describeOnCondition(edition === 'EE')('Preview', () => {
     await clickAndWait(page, page.getByRole('link', { name: /open preview/i }));
 
     // Check if the iframe is present
-    const iframe = await page.getByTitle('Preview');
+    const iframe = page.getByTitle('Preview');
     expect(iframe).not.toBeNull();
```
markkaylor, line 98: Shouldn't you keep the await or await the expect? Per the example: https://playwright.dev/docs/api/class-page#page-get-by-title

tests/e2e/tests/content-manager/preview.spec.ts
```diff
@@ -104,7 +105,7 @@ describeOnCondition(edition === 'EE')('Preview', () => {
     // Navigate to the published tab
     await clickAndWait(page, page.getByRole('tab', { name: /^Published$/ }));
 
-    const updatedIframe = await page.getByTitle('Preview');
+    const updatedIframe = page.getByTitle('Preview');
     const srcPublished = await updatedIframe.getAttribute('src');
```
markkaylor, line 109: I'm a little confused by what is async and what is not in the playwright API. Is it clear for you? Are you confident in these changes because missing an await can very easily lead to flaky tests

remidej:
Thanks for the QA @markkaylor!

- the sidebar you see is _inside_ the iframe. It's because we set up a dummy preview route inside the getstarted app, just for testing convenience. Users won't see it when they set up a real preview
- for the edit view header: yeah that's a bug, worth its own issue imo

markkaylor:
> the sidebar you see is inside the iframe. It's because we set up a dummy preview route inside the getstarted app, just for testing convenience. Users won't see it when they set up a real preview 

🤣  got it my bad!

> for the edit view header: yeah that's a bug, worth its own issue imo

You think it should be a GH issue or JIRA or both? Would you mind creating it... 🙏 

Just want to be confident about the changes to the tests and then we can approve...promise! 🏁

tests/e2e/tests/content-manager/preview.spec.ts
remidej: `page.getByTitle` does not return a promise, so I'm confident it wasn't useful, vscode was warning me about it.

But you're right that we should await an expect here

markkaylor: Oh I think I get it. It looks like getting a locator is not async but the actions on that locator are async. I think there are many places where we await when we don't need. But does that have any sort of real impact 🤔 or just ugly?

I did just notice getAttribute is discouraged, could you please try what they recommend with `toHaveAttribute` to avoid a flaky test 🙏 :
Screenshot 2024-12-02 at 17 25 32

remidej: > It looks like getting a locator is not async but the actions on that locator are async

that's my understanding too. But not actions return promises. For example I tried adding a step to make sure it's there but it also doesn't return a promise, so not useful:

screenshot 2024-12-02 at 17 24 30@2x

Extra awaits don't create issues except they make me sad when I see warnings in my editor, so I like to 🧹🧹🧹 when I find them

markkaylor: Yes I agree good to clean them and only use them where needed, thanks for doing that! I don't get those errors though I wonder why 😞

markkaylor (approved):

PR #22330, opened 2024-12-03
fix: do not select document ids on components

Marc-Roig:
### What does it do?

When querying a CT with a compo, trying to select a field from the compo errored out.

Issue was in convertQueryParams utility, it always injected a `documentId` field inside the `fields`attribute, regardless if the populated model was a content type or a component.

### Why is it needed?

To fix the following scenario
e.g. query:
```ts
    const result = await strapi.documents('plugin::myplugin.test').findMany({
      populate: {
        compo: {
          fields: ['name']
        },
      }
    })
```

e.g error
image

This will hopefully unblock https://github.com/strapi/strapi/pull/22273

### DIFF
--- a/packages/core/utils/src/__tests__/convert-query-params.test.ts
+++ b/packages/core/utils/src/__tests__/convert-query-params.test.ts
@@ -18,6 +18,9 @@ const models = {
       title: {
         type: 'string',
       },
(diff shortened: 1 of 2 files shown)


alexandrebodin (commented):
looks pretty straigth forward, worth add quick unit tests ;)

alexandrebodin (approved):
I'll leave the QA to you but LGTM

PR #22334, opened 2024-12-03
fix: preview back button behavior

### LINKED ISSUE #22299: [Preview] Edit view : The Back button redirects to the Preview

- Edition: Community
- OS: Mac and WSL
- Environment: Local

### Describe the bug

When opening the preview of an entry, then closing the preview and clicking on the back button, I’m redirected to the preview

1. Go to the list view and go to an entry
2. Open the preview, then close it
3. Click on the back button and you will be redirected to the preview

### Expected behavior

It should redirect to the list view (or at least the previous page in the CMS)

### Screenshots

https://github.com/user-attachments/assets/a407821b-4717-43b7-bc6d-568da993863c

remidej:
### What does it do?

Updates the preview back button so it uses navigate(-1) from React Router, simulating a click on the browser's back button. It's needed because otherwise, if you go to preview, click back, and click and the edit view's back button, you're take to the preview page again.

I also fixed an issue in the shared BackButton component where the href of the link was the wrong one, it would show the URL of the current page instead of the previous one.

### How to test it?

I think it's better to let you try it and see if it does what you expect. Just avoid starting a session (or reloading) directly on the preview page, because that will rely on the fallback URL, in which case the back/previous logic cannot be managed.

FYI I noticed a bug where if you toggle the draft and publish tabs multiple times on the preview page, clicking on the back button will not do the navigation on the first try. But I'm pretty sure this is a React Router bug because of us embedding the admin inside itself. Try changing the preview URL in the config to any other URL and I think the bug should disappear.

### Related issue(s)/PR(s)

fixes #22299

### DIFF
--- a/packages/core/admin/admin/src/features/BackButton.tsx
+++ b/packages/core/admin/admin/src/features/BackButton.tsx
@@ -241,7 +241,6 @@ const BackButton = React.forwardRef<HTMLAnchorElement, BackButtonProps>(
     };
 
     // The link destination from the history. Undefined if there is only 1 location in the history.
-    // const historyTo = canGoBack ? history.at(-2) : undefined;
     const historyTo = canGoBack ? history.at(currentLocationIndex - 2) : undefined;
     // If no link destination from the history, use the fallback.
     const toWithFallback = historyTo ?? fallback;
--- a/packages/core/admin/admin/src/features/BackButton.tsx
+++ b/packages/core/admin/admin/src/features/BackButton.tsx
@@ -4,7 +4,7 @@ import { Link, LinkProps } from '@strapi/design-system';
 import { ArrowLeft } from '@strapi/icons';
 import { produce } from 'immer';
 import { useIntl } from 'react-intl';
-import { NavLink, type To, useLocation, useNavigate } from 'react-router-dom';
+import { NavLink, type To, useLocation, useNavigate, useNavigationType } from 'react-router-dom';
 
 import { createContext } from '../components/Context';
 
@@ -70,6 +70,7 @@ interface HistoryProviderProps {
 
 const HistoryProvider = ({ children }: HistoryProviderProps) => {
   const location = useLocation();
+  const navigationType = useNavigationType();
   const navigate = useNavigate();
   const [state, dispatch] = React.useReducer(reducer, {
     history: [],
@@ -89,8 +90,7 @@ const HistoryProvider = ({ children }: HistoryProviderProps) => {
 
   const goBack: HistoryContextValue['goBack'] = React.useCallback(() => {
     /**
-     * Perform the browser back action
-     * dispatch the goBack action to keep redux in sync
+     * Perform the browser back action, dispatch the goBack action to keep the state in sync
      * and set the ref to avoid an infinite loop and incorrect state pushing
      */
     navigate(-1);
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/features/BackButton.tsx
```diff
@@ -221,7 +241,8 @@ const BackButton = React.forwardRef<HTMLAnchorElement, BackButtonProps>(
     };
 
     // The link destination from the history. Undefined if there is only 1 location in the history.
-    const historyTo = canGoBack ? history.at(-1) : undefined;
+    // const historyTo = canGoBack ? history.at(-2) : undefined;
```
markkaylor, line 244: Did you wwant to get rid of this?

packages/core/admin/admin/src/features/BackButton.tsx
```diff
@@ -221,7 +241,8 @@ const BackButton = React.forwardRef<HTMLAnchorElement, BackButtonProps>(
...
+    const historyTo = canGoBack ? history.at(currentLocationIndex - 2) : undefined;
```
remidej, line 245:
```suggestion
    const historyTo = canGoBack ? history.at(currentLocationIndex - 2) : undefined;
```

markkaylor (approved):
Just tested it out.  Looks good to me.

simotae14:
I tested it locally and it works, I have the same error when I toggle draft and publish tabs multiple times, and then when I click on the Close button I am redirected to this page 
Screenshot 2024-12-04 alle 11 35 29

remidej:
@simotae14 can you tell me if you get the same issue if you change the preview URL in the getstarted's config/admin.js to any external domain?

simotae14:
@remidej in case I change the preview URL with an external url I need to click multiple times to the close button before being redirected to the edit view page

remidej:
@simotae14 interesting, can you show the config you set up and share a video?

simotae14:
This is my admin file
`module.exports = ({ env }) => ({
  // autoOpen: false,
  auth: {
    secret: env('ADMIN_JWT_SECRET', 'example-token'),
  },
  apiToken: {
    salt: env('API_TOKEN_SALT', 'example-salt'),
  },
  auditLogs: {
    enabled: env.bool('AUDIT_LOGS_ENABLED', true),
  },
  transfer: {
    token: {
      salt: env('TRANSFER_TOKEN_SALT', 'example-salt'),
    },
  },
  flags: {
    nps: env.bool('FLAG_NPS', true),
    promoteEE: env.bool('FLAG_PROMOTE_EE', true),
  },
  preview: {
    config: {
        const kind =
          strapi.contentType(uid).kind === 'collectionType' ? 'collection-types' : 'single-types';
        // return `/admin/preview/${kind}/${uid}/${documentId}/${locale}/${status}`;
        return 'http://localhost:3000/en';
      },
    },
  },
});
`

and here a video


https://github.com/user-attachments/assets/f4aaee61-abb8-4807-a4f0-a6966d20c232

now I don't have the issue

remidej:
@simotae14 thanks! Your video is what we expect indeed. So maybe it was an issue with your environment, or maybe there's a tricky edge case here that we haven't identified. I'll merge for now and ask people to pay special attention to this during the next blitz.

PR #22336, opened 2024-12-04
feat: add static preview (remove future flag)

remidej:
### What does it do?

Removes the future flag to make Static Preview available to everyone

### Why is it needed?

So folks can use it! Docs should be live in time for next week's release

### DIFF
--- a/packages/core/content-manager/admin/src/preview/index.ts
+++ b/packages/core/content-manager/admin/src/preview/index.ts
@@ -1,18 +1,12 @@
 /* eslint-disable check-file/no-index */
 
 import { PreviewSidePanel } from './components/PreviewSidePanel';
-import { FEATURE_ID } from './constants';
 
 import type { ContentManagerPlugin } from '../content-manager';
 import type { PluginDefinition } from '@strapi/admin/strapi-admin';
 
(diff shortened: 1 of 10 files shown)


Marc-Roig (approved):
can you also update `preview.test.api.ts` and `preview.spec.ts`? 🙏  we had a describeOnCondition on both tests

packages/core/content-manager/admin/src/preview/index.ts
```diff
@@ -1,18 +1,12 @@
...
 
 import { PreviewSidePanel } from './components/PreviewSidePanel';
-import { FEATURE_ID } from './constants';
 
 import type { ContentManagerPlugin } from '../content-manager';
 import type { PluginDefinition } from '@strapi/admin/strapi-admin';
 
 const previewAdmin = {
   bootstrap(app) {
-    // TODO: Add license registry check when it's available
```
markkaylor, line 11: What about this part 👀 ?

markkaylor: 👍 oh yeah. That sounds fun....

Marc-Roig (approved):

PR #22338, opened 2024-12-04
chore(homepage): move guided tour

markkaylor:
### What does it do?

Create pages/comonents/GuidedTour.tsx which now handles the logic to return the existing GuidedTourHomepage component.

### Why is it needed?

- For the homepage feature

### How to test it?

- Go to the new homepage
- You should see the gudied tour
- Skip the tour, you should now longer see it (reset local storage to get it back)
- Otherwise the GuidedTour should behave as usual

### DIFF
--- a/packages/core/admin/admin/src/pages/Home/components/GuidedTour.tsx
+++ b/packages/core/admin/admin/src/pages/Home/components/GuidedTour.tsx
@@ -1,0 +1,20 @@
+import { GuidedTourHomepage } from '../../../components/GuidedTour/Homepage';
+import { useGuidedTour } from '../../../components/GuidedTour/Provider';
+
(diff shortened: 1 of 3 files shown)


packages/core/admin/admin/src/pages/Home/components/GuidedTour.tsx
markkaylor, line 1: @remidej I followed the tech specs but now I'm wondering why we are creating a new GuidedTour component and importing `components/GuidedTour/Homepage` and not moving `components/GuidedTour/Homepage` to `pages/Home/components/GuidedTour` and updating it their to return null 🤔

remidej (commented):
could you update the e2e test?

Issue #21337, opened 2024-09-18
[Design] Missing the cursor pointer in the list field when switching from one locale to another, it might apply elsewhere


yanniskadiri:
![Image](https://github.com/user-attachments/assets/0ea171f8-a8f2-4ba9-847c-e2490ab126f5)

Adzouz:
Hi @yanniskadiri,
I couldn't reproduce on version 5.15.0. I'm closing the issue but feel free to reopen if there's anything.
Thanks 🚀

PR #22345, opened 2024-12-05
fix: Fix Cursor issue hover over local select list

pandeyprafull:
### Why is it needed?
To fix issue [21337](https://github.com/strapi/strapi/issues/21337) : Missing the cursor pointer in the list field when switching from one locale to another.

### How to test it?

1. Enable localization for any collection type you want and choose the no. of locales as your need from global settings.
2. Now go to content-manager and open a record from list of any collection-type.
3. You will be able to see the locales option to choose at the top (above save/preview menu options).
4. If you click and hover over the locales options list, you will not get cursor pointer there while hover.  

### Related issue(s)/PR(s)
fixes [21337](https://github.com/strapi/strapi/issues/21337)


pandeyprafull:
@markkaylor Can you please raise the points if it's there any issue here on checklist , if yes let me know also please review the PR and try to merge if possible ...

markkaylor:
@pandeyprafull thank you for your contribution. I will do my best to review it as soon as I can.

markkaylor:
@lucasboilly do we want all select options to have a cursor pointer? If so I think this update should probably be made in the design system.

lucasboilly:
You are right @markkaylor! 

By the way I just checked the `Select` component in the [DS](https://design-system.strapi.io/?path=/docs/inputs-select--docs) and the cursor changes. I don't know why that's not the case in the CMS.

markkaylor:
> You are right @markkaylor!
> 
> By the way I just checked the `Select` component in the [DS](https://design-system.strapi.io/?path=/docs/inputs-select--docs) and the cursor changes. I don't know why that's not the case in the CMS.

Good thinking and good catch! Yup looks like the style is [there](https://github.com/strapi/design-system/blob/0693ce92f711bbff280dcd21d952bdc8a423892c/packages/design-system/src/components/Select/SelectParts.tsx#L270-L272)

I'm not sure either but I think we should investigate that and understand what's going on instead of re-applying the style here. @pandeyprafull could you look into that?

markkaylor:
Closing this PR due to inactivity and the fact that these styles should be applied though the DS

PR #22354, opened 2024-12-06
fix: accessing graphql playground without users-permissions plugin

### LINKED ISSUE #21915: Strapi 5: Apollo Sandbox inaccessible

Bassel17:
### What does it do?

- skips the authenticate request for the graphql playground

### Why is it needed?

- So we are able to load graphql playground without users-permissions

### How to test it?

- remove users-permissions and try to access graphql playground the UI should work
- `GET` queries on graphql should be authenticated

### Related issue(s)/PR(s)

#21977 
fixes #21915

### DIFF
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -118,6 +118,17 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
       },
     };
 
+    const isPlaygroundRequest =
+      ctx.request.method === 'GET' &&
+      ctx.request.url === path && // Matches the GraphQL endpoint
+      playgroundEnabled && // Only allow if the Playground is enabled
+      ctx.request.header.accept?.includes('text/html'); // Specific to Playground UI loading
+
+    // Skip authentication for the GraphQL Playground UI
+    if (isPlaygroundRequest) {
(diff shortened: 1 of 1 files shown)


innerdvations (approved):
LGTM, tested and works as expected!

Could you add a line in the documentation explaining that the landing page will show schemas even without authorization? I feel like that's something we should still make very clear, with or without u&p.

PR #22355, opened 2024-12-06
enhancement(homepage): create recently edited documents endpoint

remidej:
### What does it do?

Adds a `/admin/homepage/recent-documents` endpoint in the admin package.

I haven't extracted many utils or anything, because this PR blocks the frontend work, so ideally we'd merge it soon. I think we can look at the abstractions to create when we work on the recently published documents ticket.

### Why is it needed?

For the "recently edited" widget on the new homepage

### How to test it?

Make updates in Strapi. Then make a GET `/admin/homepage/recent-documents?action=update`, while being authenticated. You should see the most recent documents you have updated across all content types. Each should have a "title" field that is the main field (or the documentId if there's no main field)

### DIFF
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -1,0 +1,11 @@
+import type { Core } from '@strapi/types';
+
+const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
+  return {
+    async getRecentUpdates() {
+      return 'recent updates';
+    },
+  };
+};
+
+export { createHomepageService };
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -1,6 +1,6 @@
 import type { Core, UID } from '@strapi/types';
 import { contentTypes } from '@strapi/utils';
-import { omit } from 'lodash/fp';
+import type { GetRecentDocuments } from '../../../shared/contracts/homepage';
 
 const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
   const MAX_DOCUMENTS = 4;
@@ -15,7 +15,7 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
   const coreStore = strapi.db.query('strapi::core-store');
 
   return {
-    async getRecentUpdates() {
+    async getRecentUpdates(): Promise<GetRecentDocuments.Response['data']> {
       // Get all the content types the user has permissions to read
       const readPermissions = await permissionService.findMany({
         where: {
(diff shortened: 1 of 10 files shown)


packages/core/admin/server/src/services/homepage.ts
```diff
@@ -0,0 +1,131 @@
...
+                kind: contentType.kind,
+                hasDraftAndPublish,
+              },
+            };
+          });
+        })
+      );
+
+      const overallRecentDocuments = recentDocuments
+        .flat()
```
markkaylor, line 100: I think the nested loop and flattening could be avoided by splitting into two arrays. One to compute the info you need with fields. A second one that maps the first one to do the actual async stuff.

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -0,0 +1,131 @@
...
+        .sort((a, b) => b.data.updatedAt.valueOf() - a.data.updatedAt.valueOf())
+        .slice(0, MAX_DOCUMENTS);
```
markkaylor, line 102: Shouldn't you slice first to only sort on the length of MAX_DOCUMENTS

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -0,0 +1,131 @@
...
+
+          // Only add locale if it's localized
+          const isLocalized = (contentType.pluginOptions?.i18n as any)?.localized;
+          if (isLocalized) {
+            fields.push('locale');
+          }
+
+          const documents = await strapi.documents(contentTypeName).findMany({
+            limit: MAX_DOCUMENTS,
+            sort: 'updatedAt:desc',
```
markkaylor, line 70: Does this really do anything for us or how come you also have to sort updatedAt on 101?

markkaylor (commented):
Here come the nitpicks!

Nice tests!

Tested it and it's working.

packages/core/admin/server/src/services/homepage.ts
markkaylor: Not sure why Github didn't include my entire selection but I'm talking about everything inside the Promise.all above

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -0,0 +1,131 @@
...
+             */
+            const mainFieldValue = document[mainField ?? 'documentId'];
+            delete document[mainField];
+
+            return {
+              data: {
+                ...document,
+                updatedAt: new Date(document.updatedAt),
+                title: mainFieldValue,
+              },
```
markkaylor, line 88: Could the delete just be avoided with something like this? 

```suggestion
            const { mainField, ...restDcoument } = document;

            return {
              data: {
                ...restDocument,
                updatedAt: new Date(document.updatedAt),
                title: mainField ?? 'documentId',
              },
```

packages/core/admin/server/src/services/homepage.ts
remidej: I don't think I can because before the sort all the documents are one content type after the other. So if I slice before I sort then I'll only keep content type A docs even if content type Z docs are more recent

packages/core/admin/server/src/services/homepage.ts
remidej: This makes sure we only fetch the most recent stuff for each content type. Line 101 compares a blend of all content types to only keep the most recent stuff overall

packages/core/admin/server/src/services/homepage.ts
remidej: lemme try again but I think the types weren't happy when I tried that

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -0,0 +1,131 @@
...
+            ),
+          },
+        },
+      });
+      const configurations = rawConfigurations.map((rawConfiguration) => {
+        return JSON.parse(rawConfiguration.value);
+      });
+
+      const recentDocuments = await Promise.all(
+        allowedContentTypeNames.map(async (contentTypeName) => {
```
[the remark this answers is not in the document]
remidej: I can if you want, but I'm personally not a fan. With it I lose type inference on `contentTypeName` so I need to cast it. I also don't love reaching for util functions when native JS seems fine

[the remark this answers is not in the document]
markkaylor: ok

packages/core/admin/server/src/services/homepage.ts
remidej: Actually sorry this does break type safety. The reason is that mainField is dynamic, so when I destructure `[mainField]: mainFieldValue`, typescript thinks that because mainField is dynamic then nothing is safe anymore. Not 100% sure why I don't have the issue with `delete` though 🤷

markkaylor: No worries, I'm cool with what you have too just thought I'd suggest this as an option, but if TS isn't happy let's stick with what you have

remidej: Yeah I'm sad it doesn't work as well because it's very clean

markkaylor (approved):

PR #22369, opened 2024-12-09
future(homepage): add generic widget component

remidej:
### What does it do?

Adds a widget component that will be the shell for each homepage widget.

### How to test it?

- Go to the admin homepage and see what it looks like. I made the widget children red so it's obvious what part is the widget and what part is the children. The children should be scrollable
- To test the error and loading states, comment the children and uncomment each one of them


packages/core/admin/admin/src/pages/Home/components/Widget.tsx
```diff
@@ -0,0 +1,52 @@
...
+      hasRadius
+    >
+      <Flex direction="row" alignItems="center" gap={2}>
+        <Icon fill="neutral500" />
+        <Typography textColor="neutral500" variant="sigma">
+          {formatMessage(title)}
+        </Typography>
+      </Flex>
+      <Box width="100%" height="256px" overflow="auto">
+        {isLoading ? (
```
markkaylor, line 35: what about error? 🤔

remidej: I'm not sure if the error state will be common between all widgets or not. We don't have error state designs so I just asked @lucasboilly for them

markkaylor: Just a thought, couldn't we have children here and create a similar API to [Page Helpers](https://github.com/strapi/strapi/blob/775e90d48ce6fe0aa3fbb16e78059d805e15178d/packages/core/admin/admin/src/components/PageHelpers.tsx#L1)? Like Widget.Loading, Widget.Empty, Widget.Error? I wonder if we could just use the page ones actually 🤔

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
```
simotae14, line 1: this line is useless, you don't need to import React

simotae14 (approved):
Hello @remidej I tested your widget and all the different options in case it is loading, there is an error or there are no data and it works.
I added just one small comment, by the way the code looks great and it works

packages/core/admin/admin/src/pages/Home/components/Widget.tsx
```diff
@@ -0,0 +1,117 @@
+import * as React from 'react';
+
+import { Box, Flex, Loader, Typography } from '@strapi/design-system';
+import { PuzzlePiece, WarningCircle } from '@strapi/icons';
+import { EmptyDocuments } from '@strapi/icons/symbols';
+import { type MessageDescriptor, useIntl } from 'react-intl';
+
+interface RootProps {
+  title: MessageDescriptor;
```
markkaylor, line 9: What's your reasoning for requiring the translation object vs just a string? The argument I see for the string is the Widget consumer decides whether or not the title should be translated.

markkaylor (commented):
I think this looks really nice, thanks for taking the time. I haven't tested but happy to approve to move forward. Otherwise I'll do the QA this afternoon.

markkaylor (approved):

packages/core/admin/admin/src/pages/Home/components/Widget.tsx
remidej: I'm thinking that the future API will likely look something like this:

```js
bootstrap() {
  app.registerWidget({ title, icon, component })
}
```

and in that context, from a lifecycle function, I'm not sure how to access the react-intl utils. Custom fields are a quite similar and we pass the translation object for them

packages/core/admin/admin/src/pages/Home/HomePage.tsx
remidej: will remove it in my next PR thanks

packages/core/admin/admin/src/pages/Home/components/Widget.tsx
markkaylor: What do you need title and icon for when registering a widget? Are they used elsewhere besides the component itself?

remidej: not for the mvp, but we will need them in the future for a "configure the view" style feature

PR #22377, opened 2024-12-09
enhancement(homepage): get recent publish activity

markkaylor:
### What does it do?

- Refactors the service to extract functions to be shared 
- Updates homepage service to get recent publishing activity

### Why is it needed?

- To display publish activity

### How to test it?

- Publish some stuff, update a published document to get a modified state, and make some updates to documents that don't have draft and publish
- Make an authenticated GET to `http://localhost:1337/admin/homepage/recent-documents?action=publish` 
- You should see only documents with draft and published, only the published data, and the correct status (published or modified)

### DIFF
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -94,20 +94,14 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
 
   const formatDocuments = (documents: Modules.Documents.AnyDocument[], meta: DocumentMeta) => {
     return documents.map((document) => {
-      /**
-       * Save the main field value before deleting it so we can use the common
-       * title key instead across all content types. Use the delete operator instead of
-       * destructuring or lodash omit for better type inference.
-       */
-      const mainFieldValue = document[meta.mainField ?? 'documentId'];
-      delete document[meta.mainField];
+      const { mainField, ...restDocument } = document;
 
       return {
         data: {
-          ...document,
-          ...(document.publishedAt && { publishedAt: new Date(document.publishedAt) }),
+          ...restDocument,
+          ...(restDocument.publishedAt && { publishedAt: new Date(restDocument.publishedAt) }),
           updatedAt: new Date(document.updatedAt),
-          title: mainFieldValue,
+          title: document[meta.mainField ?? 'documentId'],
         },
         meta: {
           model: meta.uid,
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -171,6 +171,7 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
 
       return addStatusToDocuments(overallRecentDocuments);
     },
+
     async getRecentlyUpdatedDocuments(): Promise<GetRecentDocuments.Response['data']> {
       const allowedContentTypeUids = await getPermittedContentTypes();
       // Fetch the configuration for each content type in a single query
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -15,7 +15,9 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
   const coreStore = strapi.db.query('strapi::core-store');
 
   return {
-    async getRecentUpdates(): Promise<GetRecentDocuments.Response['data']> {
+    async getActivityForAction(
+      action: 'update' | 'publish'
+    ): Promise<GetRecentDocuments.Response['data']> {
       // Get all the content types the user has permissions to read
       const readPermissions = await permissionService.findMany({
         where: {
@@ -23,9 +25,18 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
           action: 'plugin::content-manager.explorer.read',
         },
       });
-      const allowedContentTypeNames = readPermissions
+      const permittedContentTypeNames = readPermissions
         .map((permission) => permission.subject)
         .filter(Boolean) as UID.ContentType[];
+
+      // Setup for the provided action
+      const allowedContentTypeNames =
+        action === 'publish'
+          ? permittedContentTypeNames.filter((contentType) => {
+              return contentTypes.hasDraftAndPublish(strapi.contentType(contentType));
+            })
+          : permittedContentTypeNames;
+      const actionColumn = action === 'publish' ? 'publishedAt' : 'updatedAt';
 
       // Fetch the configuration for each content type in a single query
       const rawConfigurations = await coreStore.findMany({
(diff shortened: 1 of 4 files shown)


packages/core/admin/server/src/services/homepage.ts
```diff
@@ -1,129 +1,218 @@
...
+            availableLocales: false,
           }
+        );
+        const status = metadataService.getStatus(document.data, availableStatus);
+
+        return {
+          ...document.data,
+          ...document.meta,
+          status,
+        } as unknown as RecentDocument;
```
markkaylor, line 151: ℹ️ I just started getting frustrated. If you have ideas to improve the TS let me know.

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -1,129 +1,212 @@
...
+    return Promise.all(
+      documents.map(async (document) => {
+        /**
+         * Tries to query the other version of the document if draft and publish is enabled,
+         * so that we know when to give the "modified" status.
+         */
+        const { availableStatus } = await metadataService.getMetadata(
+          document.meta.model,
+          document.data,
+          {
```
remidej, line 134: Maybe you can extend your DocumentMeta type here?

packages/core/admin/server/src/services/homepage.ts
remidej: this makes me a bit sad, what was the error?

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -1,129 +1,212 @@
...
+      const configurations = await getConfiguration(allowedContentTypeNames);
+      // Get the necessary metadata for the documents
+      const documentsMeta = await getDocumentsMetaData(allowedContentTypeNames, configurations);
+      // Now actually fetch and format the documents
+      const recentDocuments = await Promise.all(
+        documentsMeta.map(async (meta) => {
+          const docs = await strapi.documents(meta.uid).findMany({
+            limit: MAX_DOCUMENTS,
+            sort: 'updatedAt:desc',
+            fields: meta.fields,
```
remidej, line 195: we could use an empty line in between

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -1,129 +1,212 @@
...
-          if (hasDraftAndPublish) {
-            fields.push('publishedAt');
-          }
+  const getPermittedContentTypes = async () => {
+    const readPermissions = await permissionService.findMany({
+      where: {
+        role: { users: { id: strapi.requestContext.get()?.state?.user.id } },
+        action: 'plugin::content-manager.explorer.read',
+      },
+    });
```
Marc-Roig, line 44: how does this work for super admins? do we also store super admin permissions in the db?

packages/core/admin/shared/contracts/homepage.ts
remidej, line 1: I think you need to add publishedAt to RecentDocument. Could this be why you have the TS issue?

tests/api/core/admin/admin-homepage.test.api.ts
```diff
@@ -157,4 +183,60 @@ describe('Homepage API', () => {
...
+      data: {
+        title: 'The Paperback Writer',
+      },
+    });
+    const tag = await strapi.documents(tagUid).create({
+      data: {
+        slug: 'Tag 1',
+      },
+    });
+    // Create non raft and publish document
```
remidej, line 199:
```suggestion
    // Create non draft and publish document
```

remidej (commented):
I think you find the right balance between extracting functions and repetitive code. This makes it easier to understand what each service method does at a glance 🎈

packages/core/admin/server/src/services/homepage.ts
remidej: I think so. I've been testing with my super admin account JWT

markkaylor: Not sure 🤔 @remidej any idea? Otherwise I'll look into it.

markkaylor: oops didn't see your comment 🙃 ✅

packages/core/admin/server/src/services/homepage.ts
remidej: this ended up in the wrong place I meant there: https://github.com/strapi/strapi/pull/22377/files#diff-eb8ec54830d7c9a5050f556298d317736043775b05e26620d260fde437253b79R118-R122

packages/core/admin/server/src/services/homepage.ts
markkaylor: The problem TypeScript isn't smart enough to know what's in the spread. So it thinks all the properties are missing.

remidej: I see you removed the delete keyword to use destructuring, I think this is the issue I had without that delete

markkaylor: I think 🤔 , idk I'm getting turned around with the 2D array. 

Here's the error:

```
Type '{ status: any; model: ContentType; kind: ContentTypeKind; hasDraftAndPublish: boolean; length: number; toString(): string; toLocaleString(): string; toLocaleString(locales: string | string[], options?: (NumberFormatOptions & DateTimeFormatOptions) | undefined): string; ... 37 more ...; [Symbol.unscopables]: { ...; }...' is not assignable to type 'RecentDocument[]'.
  Type '{ status: any; model: ContentType; kind: ContentTypeKind; hasDraftAndPublish: boolean; length: number; toString(): string; toLocaleString(): string; toLocaleString(locales: string | string[], options?: (NumberFormatOptions & DateTimeFormatOptions) | undefined): string; ... 37 more ...; [Symbol.unscopables]: { ...; }...' is missing the following properties from type 'RecentDocument': documentId, title, updatedAt
```

markkaylor: > I see you removed the delete keyword to use destructuring, I think this is the issue I had without that delete

We do lose type inference and I will have to revert that change but it doesn't completely resolve the issue here. I find that so frustrating that we can't use better JavaScript features because TypeScript doesn't know what to do with them.

packages/core/admin/server/src/services/homepage.ts
markkaylor: Not I see where you mean? I think I have one after the findMany

packages/core/admin/server/src/services/homepage.ts
remidej: nice, this is simpler and makes sense. testing now

packages/core/admin/server/src/services/homepage.ts
remidej: My comment is again in the wrong spot for some reason. I mean in between the two methods in the service

markkaylor:
@remidej 

I think among the files changed I saw what I would call a content type UID referred to as `model`, `contentTypeName` and I think also sometimes `contentType`. But my question really is what's your reasoning for calling it `model` on the response? Are you opposed to `contentTypeUid`?  I think it is more explicit.

```
export interface RecentDocument {
  kind: Struct.ContentTypeKind;
  model: UID.ContentType;
  documentId: string;
  locale?: string;
  status?: 'draft' | 'published' | 'modified';
  title: string;
  updatedAt: Date;
  publishedAt?: Date | null;
}
```

packages/core/admin/server/src/services/homepage.ts
markkaylor: oh right, yeah I'll do that

remidej:
@markkaylor To be honest I've seen this thing have so many names in different places of strapi that I gave up. But yeah if it could at least be consistent within the homepage code that would be nice

remidej (approved):

PR #22378, opened 2024-12-09
fix: Cleaner fallback for documents and audio files in the Media Libr…

### LINKED ISSUE #22211: Cleaner fallback for documents and audio files in the Media Library list view
### Why should we do it?

There's no real fallback for the preview documents and audio files in the list view of the Media library. Instead, we have a blue square with the extension name in it but it looks buggy.

CleanShot 2024-11-19 at 16 33 45@2x

### Solution

We should display the right icons for the right files depending on what's available in our icon library.

- For documents, the default icon should be `File`, but it should change to `FilePdf`, `FileCsv`, `FileZip` or `FileXls` for PDFs, CSVs, ZIPs and XLSs, exactly like in the grid view (see https://github.com/strapi/strapi/issues/22100 and  https://github.com/strapi/strapi/pull/22151).
- For audio files, we should always use `VolumeUp`. The reason is that there's no specific icons for supported audio files (ex: MP3, WAV) in [Phosphor](https://phosphoricons.com/), our icon library.
- We don't need any fallback for images and videos as we directly display a short preview of it.

Specifications:
- The icon should have a 16px size, use `Neutral500` for the color and be centered
- The container of the icon should use `Neutral100` for the color and have a 100% border radius

Frame 31152

dilshad-knk:
resolves #22211 

**Before** 

image

**Now**

Screenshot 2024-12-09 231747
Screenshot 2024-12-09 231806


lucasboilly:
Hi @dilshad-knk 🙂

All good on the design side, thanks a lot!

I'm gonna share the PR to the team 👍

packages/core/upload/admin/src/components/TableList/PreviewCell.tsx
```diff
@@ -86,9 +96,51 @@ export const PreviewCell = ({ type, content }: PreviewCellProps) => {
...
+        justifyContent="center"
+      >
+        <VolumeUp width={16} height={16} />
+      </Flex>
+    );
+  }
+
+  let DocIcon;
+
+  switch (fileExtension) {
```
simotae14, line 116: Instead of using a switch you can try to create a map like this one
`const DOC_ICON_MAP: Record<string, React.ComponentType<any>> = {
  pdf: FilePdf,
  csv: FileCsv,
  xls: FileXls,
  zip: FileZip,
} as const;`
and handle the document types in this way 
`const fileExtension = getFileExtension(ext);
  const DocIcon = DOC_ICON_MAP[fileExtension] ?? FileIcon;
  return (
    <Flex
      justifyContent="center"
      borderRadius="100%"
      background="neutral100"
      color="neutral500"
      width="3.2rem"
      height="3.2rem"
    >
      <DocIcon width={16} height={16} />
    </Flex>
  );`

simotae14 (changes requested):
Hello @dilshad-knk thanks a lot for your contribution
Please take a look at my comments and also there is a FE unit test failing, can you please give it a look? thanks

simotae14:
hello @dilshad-knk , thanks again for your contribution, can you please take a look at my comments and fix the issues with unit tests and e2e tests, thanks

dilshad-knk:
image

why iam getting this error in set up....its wsl

Adzouz:
Hey @simotae14, I made the changes inspired by your comment. Let me know if it's good for you so we can merge! 🚀

simotae14 (changes requested):
Could we avoid having snapshot tests and instead replace them with a meaningful unit test? we are trying to remove all the useless snapshot tests

Adzouz:
> Could we avoid having snapshot tests and instead replace them with a meaningful unit test? we are trying to remove all the useless snapshot tests

Hey @simotae14, you're right, this snapshop is useless and we are already testing the functionality so I removed all match to snapshot in order to get rid of it.

simotae14 (approved):
well done

PR #22380, opened 2024-12-10
fix: adjust polymorphic relation encoding and index mapping

### LINKED ISSUE #22170: Not Saving Order of Components in Dynamic Zone When Published

- Node.js version: v18.20.4
- Yarn version: 1.22.22
- Strapi version: 5.4.0
- Database: MySQL
- Operating system: MacOS (darwin-arm64)

### Describe the bug

Strapi Not Saving Order of Components in Dynamic Zone When Published

1. Go to Article edit page
2. Drag and Drop components in dynamic zone
3. Click on Save
4. Switch to Published tab
5. See error: Component order is not save like expect
(in DB, column `article_cmps.order` is not saved for Published document, only saved for Draft document)
#### Expected Behavior:

The order of components in the dynamic zone should be preserved after publishing the content.

#### Actual Behavior:

The order of components is not preserved, and they appear in a different order after publishing.

#### Additional Information:

I have ensured that the dynamic zone is correctly configured in the Content-Type Builder.
I am using the latest version of Strapi.
I have not implemented any custom save logic for the dynamic zone.

https://github.com/user-attachments/assets/8c9fee63-c8d0-4ab7-acc6-eeeb5c24ba22

### LINKED ISSUE #22228: Relation field does not keep order in published version
## Bug report

- Node.js version: 18.18
- NPM version: 10.8.2
- Strapi version: 5.3.0
- Database: mysql
- Operating system: Ubuntu

### Describe the bug

Field which is Relation and you select items in what order you want and it keeps it when returning from API. But in published version it is returning sorted Relation field items by "id".

1. Go to draft version, select in relation field items for example: item A with id 3, item B with id 10, and item C with id 1
2. Save it and publish it
3. In draft version it returns in same order you saved items: A (id 3), B (id 10), C (id 1)
4. In published it returns: C (id 1), A (id 3), B (id 10)

### Expected behavior

In published version it should return: A (id 3), B (id 10), C (id 1)

### LINKED ISSUE #22250: Issue with Component Order Mismatch in Dynamic Zone After upgrade to v5.4.0

- Node.js version: 18.18.2
- NPM version: 10.9.0
- Strapi version: 5.4.1
- Database: mysql
- Operating system: windows 10

### Description of the Issue

The order of components displayed in the dynamic zone on localhost does not match the order configured in the Strapi admin panel. This discrepancy makes it difficult to ensure consistent rendering between the admin interface and the local environment.
This problem occured after upgrade from v5.3.0 to v5.4.0. 
Previously I used v5.3.0 without any problems.

1. Log in to the Strapi admin panel.  
2. Navigate to a content type that includes a **dynamic zone**.  
3. Add multiple components to the dynamic zone and arrange them in a specific order.  
4. Save the changes.  
5. Query the content type through the API or render it in the frontend on localhost.  
6. Observe that the order of the components displayed does not match the order configured in the admin panel.  

### Expected behavior

The order of components displayed on localhost should match the order set in the Strapi admin panel for the dynamic zone.

### Actual behavior

The order of components displayed on localhost is incorrect and does not align with the order set in the Strapi admin panel.

### Screenshots

Screenshot 2024-11-24 145648

Screenshot 2024-11-24 145652

Screenshot 2024-11-23 152736

Screenshot 2024-11-23 152344

image

### Code snippets

image

Convly:
### What does it do?

Fix the components' reordering by passing the relation's type field (_`__type` for relations and `__component` for dynamic zones' components_) instead of hard-coding `__type`.

Additionally, start the value of the order property at `1` instead of `0` to maintain consistency with the rest of the code.

### Why is it needed?

This was causing issues on entries' creation as components would not be ordered correctly (_`encodePolymorphicRelation` was returning `1::undefined`-like IDs since `__type` doesn't exist in component data_)

### How to test it?

See reproduction steps in related issues and PRs.

### Related issue(s)/PR(s)

fix #22170
fix #22228
fix #22250

### TODO

- [ ] Add e2e tests

### DIFF
--- a/packages/core/database/src/entity-manager/index.ts
+++ b/packages/core/database/src/entity-manager/index.ts
@@ -601,11 +601,11 @@ export const createEntityManager = (db: Database): EntityManager => {
           )
             .connect(
               // Merge id & __type to get a single id key
-              dataset.map(encodePolymorphicRelation({ idColumn: 'id', typeColumn: '__type' }))
+              dataset.map(encodePolymorphicRelation({ idColumn: 'id', typeColumn: typeField }))
             )
             .get()
             // set the order based on the order of the ids
-            .reduce((acc, rel, idx) => ({ ...acc, [rel.id]: idx }), {} as Record<ID, number>);
+            .reduce((acc, rel, idx) => ({ ...acc, [rel.id]: idx + 1 }), {} as Record<ID, number>);
 
           rows.forEach((row: Record<string, unknown>) => {
             const rowId = row[morphColumn.idColumn.name] as ID;


innerdvations:
Nice! Fix makes sense, will test it out now.

But this fix is just for create, do we need to do the same thing for update?

Convly:
> Nice! Fix makes sense, will test it out now.
> 
> But this fix is just for create, do we need to do the same thing for update?

Seems like `typeField` is used correctly in the update but don't hesitate to check again

innerdvations (approved):
Tested and works for me!

derrickmehaffy:
I'm sending a comment to users on all 3 issues to see if anyone can test the exp for confirmation.

otsolap:
Hi,

I am using version 5.6.0 but I still get this error? My blocks are rendered by their ID and not by how they are published.

PR #22402, opened 2024-12-12
Fixed an issue where PluginsInitializer could not exit the loading screen

### LINKED ISSUE #21811: Admin Page endless waiting after hitting F5

- Node.js version: v20.11.0
- NPM version: 10.2.4
- Strapi version: 5.0.6
- Database: SQLite
- Operating system: Windows

### Describe the bug
The following bug only happens in production mode when the project is deployed to azure. I was unable to reproduce it locally.
If you press F5 to get the entire site reloaded, sometimes you end up with an infinite loading (the blue loading spinner never ends).
There're no errors in the console.

1. Go to the admin page url in Chrome in incognito mode
2. If it loads, press F5 a few times. Sometimes the page loads, sometimes it doesn't and you'll end up with infinite loading.
3. If it doesn't load, press the Back button on the browser bar to get rid of the infinite loading. You can see the error either on the login screen or after getting logged in on the Home screen, or even in Content Manager.
4. See attached screenshot.

### Expected behavior

No infinite loading.

### Screenshots
F5_timeout

## package.json
"dependencies": {
    "@_sh/strapi-plugin-ckeditor": "^4.0.6",
    "@strapi/email": "5.0.6",
    "@strapi/plugin-cloud": "5.0.6",
    "@strapi/plugin-users-permissions": "5.0.6",
    "@strapi/strapi": "5.0.6",
    "better-sqlite3": "9.4.3",
    "moment": "^2.30.1",
    "mysql2": "^3.11.3",
    "react": "^18.0.0",
    "react-dom": "^18.0.0",
    "react-router-dom": "^6.0.0",
    "styled-components": "^6.0.0"
  }
  
  ## config
  All default in middlewares.ts except "frame-src" in "strapi::security" for strapi-plugin-ckeditor.

pksr:
Fixed a bug that prevented PluginsInitializer from re-rendering to exit the loading screen when a plugin was not in the ready state.

Fix #21811

### DIFF
--- a/packages/core/admin/admin/src/components/PluginsInitializer.tsx
+++ b/packages/core/admin/admin/src/components/PluginsInitializer.tsx
@@ -1,8 +1,5 @@
 import * as React from 'react';
 
-import { produce } from 'immer';
-import set from 'lodash/set';
-
 import { Page } from '../components/PageHelpers';
 import { StrapiAppContextValue, useStrapiApp } from '../features/StrapiApp';
 
@@ -90,17 +87,16 @@ type SetPluginReadyAction = {
 
 type Action = SetPluginReadyAction;
 
-const reducer: React.Reducer<State, Action> = (state = initialState, action: Action): State =>
-  produce(state, (draftState) => {
-    switch (action.type) {
-      case 'SET_PLUGIN_READY': {
-        set(draftState, ['plugins', action.pluginId, 'isReady'], true);
-        break;
-      }
-      default:
-        return draftState;
+const reducer: React.Reducer<State, Action> = (state = initialState, action: Action): State => {
+  switch (action.type) {
+    case 'SET_PLUGIN_READY': {
+      const newState = { ...state };
+      newState.plugins[action.pluginId].isReady = true;
+
+      return newState;
     }
-  });
+  }
+};
 
 /* -------------------------------------------------------------------------------------------------
  * Init state


hanpaine:
Thanks for the PR @pksr 🚀  we'll take a look after the holiday period 🙏

julien-maurel:
@hanpaine any news about that ? 
bug about loading screen is very painfull (#21811)

packages/core/admin/admin/src/components/PluginsInitializer.tsx
```diff
@@ -90,17 +87,16 @@ type SetPluginReadyAction = {
 
 type Action = SetPluginReadyAction;
 
-const reducer: React.Reducer<State, Action> = (state = initialState, action: Action): State =>
-  produce(state, (draftState) => {
```
markkaylor, line 94: It does seem like produce is part of the problem but I don't understand why and I'd like to. Could you share more how you tracked it down to here and ultimately decided to remove it?

markkaylor (commented):
Thanks for the PR as far as I can tell this does resolve the issue but I'd like to better understand why.

packages/core/admin/admin/src/components/PluginsInitializer.tsx
jorrit: According to the docs of `immer`, there is no need to return anything. `produce` should return a new object when it detects modification. But not needing `immer` and `lodash` for this simple operation is perhaps a good change in its own right.

The thing that is lacking from this change is handling `action.type != 'SET_PLUGIN_READY'`

markkaylor: Thanks to both of you for your input

> But not needing immer and lodash for this simple operation is perhaps a good change in its own right

I agree. Lodash we definitely want to get rid of. But using `immer.produce` is a convention being followed in all other reducers. If we break that convention here I'd like to be certain why and leave a comment so someone doesn't come back later and add it back trying to keep things consistent. 

> I cannot say for sure if it is really due to "produce" since others in the ticket have used my code without removing "produce".

I tried what you have here and also with "produce" and without lodash, but I had the loading state again. 

I did another test to [log children](https://github.com/strapi/strapi/blob/0edd480c667859a1853e55e881b26d2cf9edeef7/packages/core/admin/admin/src/components/PluginsInitializer.tsx#L71) to see if we make it there, and we do reach it but we still have the loader too 🤷‍♂️ 

> But an entry without "isReady" would not the exit the loading state:

All plugins should have an ` isReady` property set when they are instantiated with `new Plugin`

Screenshot 2025-01-29 at 17 50 44

> But of course I can add the "default" again if that makes sense

I think it does make sense to keep the default.

jorrit: I have found the cause of this problem: the `Plugin` class is not `immerable`. `immer` requires custom classes to be processed using `immerable`, otherwise changes on those classes are not detected by `produce()`.

See https://immerjs.github.io/immer/complex-objects

packages/core/admin/admin/src/components/PluginsInitializer.tsx
```diff
@@ -90,17 +87,16 @@ type SetPluginReadyAction = {
...
-    switch (action.type) {
-      case 'SET_PLUGIN_READY': {
-        set(draftState, ['plugins', action.pluginId, 'isReady'], true);
-        break;
-      }
-      default:
-        return draftState;
+const reducer: React.Reducer<State, Action> = (state = initialState, action: Action): State => {
```
markkaylor, line 90: I think it could be good to leave a comment like:

> immer produce is intentionally not used in this reducer since the Plugin class is not "immerable". See: https://immerjs.github.io/immer/complex-objects/

packages/core/admin/admin/src/components/PluginsInitializer.tsx
```diff
@@ -90,17 +87,16 @@ type SetPluginReadyAction = {
...
+  switch (action.type) {
+    case 'SET_PLUGIN_READY': {
+      const newState = { ...state };
+      newState.plugins[action.pluginId].isReady = true;
```
markkaylor, line 94: According to the react docs shouldn't we be spreading at every level since spread is a shallow copy: https://react.dev/reference/react/useReducer#ive-dispatched-an-action-but-the-screen-doesnt-update

```js
return {
    ...state,
    plugins: {
      ...state.plugins,
      [action.pluginId]: {
        ...state.plugins[action.pluginId],
        isReady: true,
      },
    },
  };
```

packages/core/admin/admin/src/components/PluginsInitializer.tsx
markkaylor: Great, thanks @jorrit let's move forward with this PR then!

jorrit: @markkaylor Either that, or this code is left alone and the `Plugin` class is modified.

markkaylor: That's true we could update it over there. I considered it for a second but I don't think I really have a strong preference either way. I guess it makes sense to do it over there in case anywhere else in the app we try to update state similarly for a Plugin. I'll also ask others on the team and I'll go ahead and test it this morning.

jorrit: It depends if you like to keep your convention of using produce in every reducer. Your call :)

packages/core/admin/admin/src/components/PluginsInitializer.tsx
markkaylor: Yeah no worries, I'll handle it from here. Thanks.

markkaylor:
Thanks again to both of you for your help. I'm closing this one in favor of https://github.com/strapi/strapi/pull/22763.

PR #22405, opened 2024-12-12
fix(home): make locale defined or null to get the correct status

markkaylor:
### What does it do?

Ensures locale is defined or null when getting a status

### Why is it needed?

Otherwise the status is incorrect

### How to test it?

Edit some content in draft and published 
Verify the correct status show up on the homepage

NOTE: A content-type without d&p will still show draft. This is another bug....I think?

### DIFF
--- a/packages/core/admin/server/src/services/homepage.ts
+++ b/packages/core/admin/server/src/services/homepage.ts
@@ -96,7 +96,7 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
     return documents.map((document) => {
       return {
         documentId: document.documentId,
-        locale: document.locale,
+        locale: document.locale ?? null,
         updatedAt: new Date(document.updatedAt),
         title: document[meta.mainField ?? 'documentId'],
         publishedAt:
--- a/tests/api/core/admin/admin-homepage.test.api.ts
+++ b/tests/api/core/admin/admin-homepage.test.api.ts
@@ -149,6 +149,7 @@ describe('Homepage API', () => {
      **/
     for (let i = 0; i < 9; i++) {
       if (i % 3 === 0) {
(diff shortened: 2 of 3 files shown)


packages/core/admin/server/src/services/homepage.ts
```diff
@@ -92,11 +92,11 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
     });
   };
 
-  const formatDocuments = (documents: Modules.Documents.AnyDocument[], meta: DocumentMeta) => {
+  const formatDocuments = (documents: Modules.Documents.AnyDocument[], meta: DocumentMeta): RecentDocument[] => {
     return documents.map((document) => {
       return {
         documentId: document.documentId,
-        locale: document.locale,
+        locale: document.locale ?? null,
```
markkaylor, line 99: To debug this I queried the document in the addStatusToDoucment function:

```
const document = strapi.documents(uid).findOne({documentId: recentDocument.documentId})
```

and then picked the fields one at a time to see what `metadataService.getMetadata`` is expecting. I don't think that thing is typed properly if at all.

tests/api/core/admin/admin-homepage.test.api.ts
```diff
@@ -183,7 +186,8 @@ describe('Homepage API', () => {
     expect(response.body.data[0].contentTypeUid).toBe('api::tag.tag');
     expect(response.body.data[1].title).toBe('global-7');
     expect(response.body.data[1].contentTypeUid).toBe('api::global.global');
-    expect(response.body.data[1].status).toBe('draft');
+    // Document with globalUid was published and then updated, assert the modified status
+    expect(response.body.data[1].status).toBe('modified');
```
remidej, line 190: Looks like I had hardcoded the bug 💀

PR #22407, opened 2024-12-12
fix(e2e): remove unused published product from the database

markkaylor:
### What does it do?

- Removes a published 'products' document from the DB that is unused

On develop you can see the unused document with a database viewer. Look for products and you will see two entries. One is published. In the CMS you never see this entry. 

### Why is it needed?

It's useless. It causes tests to fail trying to publish the actual document. For example in https://github.com/strapi/strapi/pull/22409, I am trying to publish the only products document and it fails with uniqueness validation. 

### How to test it?

The tests should pass here in the CI. I've tested the updated data set in #22409 and it resolves the issue

PR #22409, opened 2024-12-12
fix: hide doc status when no d&p

markkaylor:
### What does it do?

- [Backend] Return undefined locale if the CT doesn't have D&P
- [Backend] Update API tests 
- [Frontend] Hide the status for documents undefined locale
- [e2e] TODO: Looking into it in the related PR, but there is already a published product in the DB that isn't present in the CMS in the e2e test instance.

### Why is it needed?

Only d&p documents should have a status

### How to test it?

Make changes to documents with and without draft and publish
Check the statuses are correct for draft and published (draft, published, modified)
Check there is no status for no draft and published (empty grid cell)

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22407

### DIFF
--- a/packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
+++ b/packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
@@ -81,7 +81,11 @@ const WidgetContent = ({ document }: { document: RecentDocument }) => {
       </Td>
       <Td>
         <Box display="inline-block">
-          {document.status && <DocumentStatus status={document.status} />}
+          {document.status ? (
+            <DocumentStatus status={document.status} />
+          ) : (
+            <Typography aria-hidden>-</Typography>
+          )}
         </Box>
       </Td>
       <Td>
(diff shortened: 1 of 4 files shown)


remidej:
Haven't test tested since it's a draft but LGTM so far :)

markkaylor:
@remidej I'm close to fixing the issues with the e2e test, but if you agree I think we could review and merge this and wrap up the e2e test in a follow up to be sure this gets in before the release branch lock on Monday. We still get good coverage from the api test.

remidej:
@markkaylor we're not affected by the release branch lock since this is a feature branch and will be a separate exp release. Happy to get it merged now if you want though, your call!

remidej (approved):
I just checked with @lucasboilly and we agree it would look nicer with a dash instead of the empty space. But we can merge like this first too!

remidej:
I added the dash thingy
screenshot 2024-12-18 at 11 33 22@2x

Marc-Roig (approved):

simotae14 (approved):
lgtm

Issue #22023, opened 2024-10-30
Date Field Changing values according to Locale while editing on CMS


sulavsssmc:
- Node.js version: 20.18.0
- NPM version: 10.8.2
- Strapi version: 5.1.0
- Database: postgresql
- Operating system: Mac os
- Is your project Javascript or Typescript: Typescript

I'm using the Date field in Strapi CMS using the Calendar Date Picker. The problem is, when the date is saved let's say for example "02/05/1980" is mm/dd/yyyy format, it is showing as a different date as "01/06/1980" to another editor who is from different country with wide time zone gap when editing from the cms. However when i check the postgres database and at the collections table list, the date is correctly shown as "02/05/1980". 

This is a serious problem because the nature of date being captured is a date that should not change based on locale such as "Date of a certain Historic Event" or "Date of Birth". 

Is there any option to prevent the automatic date conversion because it would create a massive confusion when date of birth is showing a day earlier or ahead on cms.

Also how do i change the "MM/DD/YYYY" format to "DD/MM/YYYY" and force it to be the standard for any user who edits the content regardless of where they are from?

bolg55:
Related to EE Internal TID8980


https://github.com/user-attachments/assets/367403d3-6900-46d0-832e-aa3631e8ba7f

leosantacruz:
+1 Same here

- Strapi v5.4.1
- No plugins installed

lucasrcorreia23:
+1 Same here

Strapi v5.4.2
No plugins installed

devincrandall:
+ 1 5.5.0

PR #22410, opened 2024-12-12
fix: remove double conversion that causes date shift

butcherZ:
### What does it do?

removed the `convertLocalDateToUTCDate` function which caused double conversion and date shift from the client side (check the video posted in #22023 ), the component now preserves the actual calendar date the user selected and eliminates timezone-related shifts (same behaviours as in V4)

### Why is it needed?

by default, `Date` object will be created in the local time zone of the user's browser, but dates should be kept consistent in UTC at all times

### How to test it?

- launch `examples/getStarted`
- change your device's timezone to PST (Los Angeles for example )
- go to content type builder, create a collection with a `date` field
- go to content manager, create a new entry
  please check:
   1. When selecting a date from a date picker, the date displayed in the input is the selected value.
   2. Without selecting the date, if the datePicker loses focus (onBlur), the date shown in the input is not changing.
 - save the new entry, and change the timezone on your device again (to Singapore for example)
 - the displayed date doesn't change its value
-  change back the timezone and verify the value again.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/22023

DX-1788


packages/core/admin/admin/src/components/FormInputs/Date.tsx
```diff
@@ -1,6 +1,5 @@
-import { forwardRef, memo } from 'react';
+import { forwardRef, memo, useState } from 'react';
```
remidej, line 1: Could you update this import to `import * as React`, then use React.useState etc? it's not enforced by ESLint yet but it's the style we try to apply everywhere for consistency

packages/core/admin/admin/src/components/FormInputs/Date.tsx
```diff
@@ -14,21 +13,45 @@ const DateInput = forwardRef<HTMLInputElement, InputProps>(
...
+          onClear={() => {
+            field.onChange(name, null);
+            setLastValidDate(null);
+            return;
+          }}
+          onBlur={() => {
+            // When the input is blurred, ensure that the value doesn't change unexpectedly
+            if (field.value && !value) {
+              field.onChange(name, lastValidDate?.toISOString() ?? null);
+            }
```
remidej, line 52: >ensure that the value doesn't change unexpectedly

How would the value change unexpectedly? reading your unit test, aren't you more trying to prevent invalid dates from sticking around?

butcherZ: My bad, I forgot to update the change. In fact, those tests are not as effective as I wanted them to be in preventing this bug from happening. We will write an E2E test for it before trying to get this PR merged.

Marc-Roig (approved):

PR #22401, opened 2024-12-12
Fix error with onChange on input of color picker

Jhonaiker2309:
### What does it do?

The onChange function of the input to change the hex colors wasn't getting the right parameters

### Why is it needed?

The color picker custom field in Strapi 5 does not allow editing or pasting a hex code in the text box

### How to test it?

In can be done in any hex input, I tested using the hex input of the route "/admin/content-manager/collection-types/api::kitchensink.kitchensink/create" that is in examples/getstarted

### Related issue(s)/PR(s)

Fixed #21838

PR #22413, opened 2024-12-13
fix the date field changing values according to locale while editing …

### LINKED ISSUE #22023: Date Field Changing values according to Locale while editing on CMS

EliezerMCR:
This PR fixes the issue where dates displayed in the Strapi CMS admin panel would shift by one day depending on the editor’s time zone. The change ensures that stored dates (e.g., dates of birth or historic events) remain consistent for all users, regardless of their locale or time zone.

Why is it needed?
Currently, when a date is stored in the database, Strapi’s DatePicker interprets and displays it based on the local time zone of the viewing user. This causes confusion as dates such as “02/05/1980” can appear as “01/06/1980” for users in a different country/time zone. Such sensitive dates (birthdates, historical event dates, etc.) should remain the same universally.

How to test it?
Run Strapi in development mode.
Create or edit a content-type that includes a date field.
Set a date, for example “02/05/1980”.
Test viewing and editing the entry from different locales/time zones.
Confirm that the date remains “02/05/1980” in the UI, without shifting a day.
Tested Environment:

Time: Thu Dec 12 2024 23:47:27 GMT-0400 (Venezuela Time)
Launched in: ~2480 ms
Environment: development
Process PID: 188358
Strapi Version: 5.4.2 (node v20.17.0)
Edition: Community
Database: sqlite
Database name: .tmp/data.db
Related issue(s)/PR(s)
Fix #22023

This ensures that date formatting and display are no longer impacted by user locale differences, preserving a consistent date representation in the CMS for all editors.


hanpaine:
Thanks for your contribution @EliezerMCR - we've actually already fixed this in #22401 🙏

PR #22419, opened 2024-12-13
enhancement: update purchase page links

Mcastres:
### What does it do?

Updates the purchase page links for:
- Content Releases
- Review Workflows

### Why is it needed?

Because we now have dedicated pages for these features on the website.


simotae14 (approved):
LGTM 🌟

Issue #20781, opened 2024-07-16
[Design] There’s a border around the drag icon in Blocks (it should not)


MarionLemaire:
**Current design**
![CleanShot 2024-07-09 at 11 10 42@2x](https://github.com/user-attachments/assets/e391f4dc-0be8-4f08-b029-ba82dc19c5c7)

**Expected design**
![Image](https://github.com/user-attachments/assets/dcace0c3-e5b8-41c4-9b8f-a75340a45ad1)

PapatMayuri:
@MarionLemaire Can you add some steps to check this issue.

pandeyprafull:
@PapatMayuri Did you start working on it or should I pick this up ?

PR #22423, opened 2024-12-13
fix(blocks): various ui issues

lucasboilly:
### What does it do?

- Removes the border of the `DragIconButton` in both default and disabled mode
- Changes the color and the size of the `Drag` icon because it wasn't visible enough
- Hides the `DragIconButton` when the input is disabled
- Changes the cursor to `not-allowed` for disabled modifiers
- Adds a shadow to `ExpandIconButton` to make it understandable that it is sticky
- Set the with of `ToolbarWrapper` to 100% and added a background color for its disabled state
- Thanks to the new width of `ToolbarWrapper` the input has a disabled style in expanded mode as well
- Removed `dragHandleTopMargin` for `Code` and the `Quote`
- Fixes padding issues

### Why is it needed?

Blocks is currently not aligned with the designs and its disabled states are not understandable.

### How to test it?

Open any entry that has a Blocks field. Don't forget to check the expanded mode.

### Related issue(s)/PR(s)

fix https://github.com/strapi/strapi/issues/20781

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
@@ -26,7 +26,7 @@ import { insertLink } from './utils/links';
 import { type Block, getEntries, getKeys } from './utils/types';
 
 const ToolbarWrapper = styled<FlexComponent>(Flex)`
-width: 100%;
+  width: 100%;
   &[aria-disabled='true'] {
     cursor: not-allowed;
     background: ${({ theme }) => theme.colors.neutral150};
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
@@ -26,7 +26,6 @@ import { insertLink } from './utils/links';
 import { type Block, getEntries, getKeys } from './utils/types';
 
 const ToolbarWrapper = styled<FlexComponent>(Flex)`
-  width: 100%;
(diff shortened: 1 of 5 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -41,6 +43,7 @@ const FlexButton = styled<FlexComponent<'button'>>(Flex)`
   // Inherit the not-allowed cursor from ToolbarWrapper when disabled
   &[aria-disabled] {
     cursor: inherit;
+    cursor: not-allowed;
```
remidej, line 46: you should replace the line above instead of adding this

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -26,8 +26,10 @@ import { insertLink } from './utils/links';
 import { type Block, getEntries, getKeys } from './utils/types';
 
 const ToolbarWrapper = styled<FlexComponent>(Flex)`
+  width: 100%;
```
remidej, line 29: can you pass width="100%" as a prop on the component instead? we try to use the DS components props as much as possible for simple things

remidej (commented):
thanks! haven't tested yet

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
lucasboilly: @remidej does it mean doing this?

CleanShot 2024-12-13 at 14 16 02@2x

remidej (approved):

Issue #22379, opened 2024-12-09
Strapi 5 upgrade error from v4 to v5.4+


kasonde:
### Required System information

- Strapi version: 4.25.x to 5.4+, 
- Database: Postgres
- Operating system: Linux
- Is your project Javascript or Typescript: Javascript

### Describe the bug

When upgrading Strapi from v4 to Strapi 5, users encountered issues with part of the process, presumably when database tables were being modified from v4 schema to Strapi 5 schema and the error below was thrown:

```bash
[2024-12-09 12:55:48] [2024-12-09 12:55:48.957] debug: ⛔️ Server wasn't able to start properly.
[2024-12-09 12:55:48] [2024-12-09 12:55:48.968] error: create index "admin_permissions_role_links_fk" on "public"."admin_permissions_role_links" ("permission_id") - relation "admin_permissions_role_links_fk" already exists
[2024-12-09 12:55:48] error: create index "admin_permissions_role_links_fk" on "public"."admin_permissions_role_links" ("permission_id") - relation "admin_permissions_role_links_fk" already exists
[2024-12-09 12:55:48]     at Parser.parseErrorMessage (/home/app/docroot/node_modules/pg-protocol/dist/parser.js:283:98)
[2024-12-09 12:55:48]     at Parser.handlePacket (/home/app/docroot/node_modules/pg-protocol/dist/parser.js:122:29)
[2024-12-09 12:55:48]     at Parser.parse (/home/app/docroot/node_modules/pg-protocol/dist/parser.js:35:38)
[2024-12-09 12:55:48]     at TLSSocket.<anonymous> (/home/app/docroot/node_modules/pg-protocol/dist/index.js:11:42)
[2024-12-09 12:55:48]     at TLSSocket.emit (node:events:517:28)
[2024-12-09 12:55:48]     at TLSSocket.emit (node:domain:489:12)
[2024-12-09 12:55:48]     at addChunk (node:internal/streams/readable:368:12)
[2024-12-09 12:55:48]     at readableAddChunk (node:internal/streams/readable:341:9)
[2024-12-09 12:55:48]     at Readable.push (node:internal/streams/readable:278:10)
[2024-12-09 12:55:48]     at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23)
```

### Steps to reproduce the behavior (according to what we know currently)

1. Upgrade code base to Strapi 5
2. Connect to a v4 database (with existing data)
3. Attempt to start your app

### Expected behavior

[...]

PR #22424, opened 2024-12-13
fix(database): delete indexes before creating if they already exist

innerdvations:
### What does it do?

In `SchemaInspector`, make `getIndexes` and `getForeignKeys` required dialect methods.

In `SchemaBuilder`, before creating an index or foreign key, drop any existing one with the same name. This works because extra drops in a `Knex.TableBuilder` do not throw an error _as long as the index existed to begin with_

### Why is it needed?

Should fix https://github.com/strapi/strapi/issues/22379 and probably any other migration errors that involve indexes, as well as any other cases where an app may have an extra index leftover from some unknown cause that prevents startup.

Note that in a sense it will "cover up" potential migration issues around dropping/creating indexes, but we will be avoiding critical migration errors in exchange for potentially not catching a bug that could cause a one-time rebuilt index during v4->v5 migration, which is not a big deal since that is already considered a possibility during migration. And in general, an unexpected index should not be a cause for Strapi to fail to start up.

### How to test it?

experimental: 0.0.0-experimental.ee7402bacc4656d268ab76aa9c334a7b7a951201

Existing CTB tests should cover most basic cases but

- Test migrations from v4 -> v5 on complex projects and data
- Test complex cases with the CTB creating/removing/updating relations, components, and dzs

### Related issue(s)/PR(s)


packages/core/database/src/schema/builder.ts
```diff
@@ -140,15 +153,27 @@ const createHelpers = (db: Database) => {
...
+    const existingIndex = existingIndexes?.find((existing) => existing.name === name);
+    const forceMigration = db.config.settings?.forceMigration;
+    if (forceMigration && existingIndex) {
+      dropIndex(tableBuilder, index);
+    }
+
     switch (type) {
       case 'primary': {
-        return tableBuilder.primary(columns, name);
+        return tableBuilder.primary(columns, { constraintName: name });
```
innerdvations, line 173: this change is just because we were using a deprecated format

packages/core/database/src/schema/builder.ts
```diff
@@ -259,30 +302,21 @@ const createHelpers = (db: Database) => {
         dropForeignKey(tableBuilder, updatedForeignKey.object);
       }
 
-      // for mysql only, dropForeignKey also removes the index, so don't drop it twice
-      const isMySQL = db.config.connection.client === 'mysql';
-      const ignoreForeignKeyNames = isMySQL
```
innerdvations, line 264: I'm still investigating why this was necessary before, now that I've discovered multiple dropIndexes are fine. I might need to put this back.

innerdvations: I'm pretty sure it is necessary

Convly (commented):
Code LGTM, I have tested with simple scenario and using the CTB. Everything has been working well so far.

gorgsenegger:
Hello, could somebody please complete the review? I cannot run my local setup anymore :-/

Bassel17 (approved):
Looks good

gorgsenegger:
Hi @innerdvations, 

I've not spent much time with the strapi code base, so maybe you could give me some insight whether the code changes in this pull request would also be applicable to version 4 (4.25.18)? I'm (suddenly) running into the same problems locally. Upgrading to version 5 does work, but then I cannot use the import/export plugin that I used to use with version 4.

Any help would be greatly appreciated, thank you!

derrickmehaffy:
> Hi @innerdvations, 
> 
> I've not spent much time with the strapi code base, so maybe you could give me some insight whether the code changes in this pull request would also be applicable to version 4 (4.25.18)? I'm (suddenly) running into the same problems locally. Upgrading to version 5 does work, but then I cannot use the import/export plugin that I used to use with version 4.
> 
> Any help would be greatly appreciated, thank you!

No sadly not, this PR will only be released on the upcoming Strapi 5 version.

PR #22425, opened 2024-12-13
fix: add component button style

lucasboilly:
### What does it do?

- Simplifies `AddComponentButton` by using its styles inherited from `Button`
- Removes the unnecessary styles change on hover when disabled
- Changes the padding left to better fit the icon shape

### Why is it needed?

The current version is not in sync with the design and makes users think that they can click on the button even if it's disabled. It will improve the UX of the disabled Dynamic Zones in the "Published" tab and in Content History.

#### Before

https://github.com/user-attachments/assets/c9d987d1-e66d-492d-939e-28ba27cfd34f

#### After

https://github.com/user-attachments/assets/e19ad3c6-859a-47fb-a629-c35acc1e98c9

### How to test it?

Open any entry with a Dynamic Zone and check its disabled version.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/AddComponentButton.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/AddComponentButton.tsx
@@ -30,13 +30,13 @@ const AddComponentButton = ({
     >
       <Flex tag="span" gap={2}>
         <StyledAddIcon aria-hidden $isOpen={isOpen} $hasError={hasError && !isOpen} />
-        <AddComponentTitle
+        <Typography
           variant="pi"
           fontWeight="bold"
-          textColor={hasError && !isOpen ? 'danger600' : 'neutral500'}
+          textColor={hasError && !isOpen ? 'danger600' : 'neutral600'}
         >
           {children}
-        </AddComponentTitle>
+        </Typography>
       </Flex>
     </StyledButton>
   );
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/AddComponentButton.tsx
```diff
@@ -53,45 +53,19 @@ const StyledAddIcon = styled(PlusCircle)<{ $isOpen?: boolean; $hasError?: boolea
...
     fill: ${({ theme, $hasError }) =>
-      $hasError ? theme.colors.danger600 : theme.colors.neutral600};
+      $hasError ? theme.colors.danger600 : theme.colors.neutral500};
   }
 `;
 
-const AddComponentTitle = styled<TypographyComponent>(Typography)``;
+const AddComponentTitle = styled<TypographyComponent>(Typography)`
+  color: ${({ theme }) => theme.colors.neutral600};
+`;
```
remidej, line 62: Looks like it was useless to have a custom component here. Could you try deleting it and and calling `<Typography textColor="neutral600">` instead where it's used?

lucasboilly: It works but the style is defined next to the style of the error state. Is it fine?

CleanShot 2024-12-13 at 16 03 26@2x

remidej: i think that's fine if it works for all cases

remidej (approved):

PR #22432, opened 2024-12-13
fix: preview back button behavior

remidej:
### What does it do?

fixes the bug where after toggling tabs multiple times in the preview page, you would need to click multiple times on the back button to get to the edit view.

### How to test it

Start from the edit view, open preview, toggle tabs a bunch of times, click on the back icon button once, you should be taken back to the preview immediately. Clicking on your browser's forward button should lead you to the preview page again.

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
@@ -13,6 +13,14 @@ const PreviewContent = () => {
   return (
     <Box
       src={previewUrl}
+      /**
+       * For some reason, changing an iframe's src tag causes the browser to add a new item in the
+       * history stack. This is an issue for us as it means clicking the back button will not let us
+       * go back to the edit view. To fix it, we need to trick the browser into thinking this is a
+       * different iframe when the preview URL changes. So we set a key prop to force React
+       * to mount a different node when the src changes.
+       */
+      key={previewUrl}
       title={formatMessage({
         id: 'content-manager.preview.panel.title',
         defaultMessage: 'Preview',


packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
```diff
@@ -13,6 +13,14 @@ const PreviewContent = () => {
...
     <Box
       src={previewUrl}
+      /**
+       * For some reason, changing an iframe's src tag causes the browser to add a new item in the
+       * history stack. This is an issue for us as it means clicking the back button will not let us
+       * go back to the edit view. To fix it, we need to trick the browser into thinking this is a
+       * different iframe when the preview URL changes. So we set a key prop to force React
+       * to mount a different node when the src changes.
+       */
+      key={previewUrl}
```
markkaylor, line 23: What a pain! This is simple enough and I'm cool with it but it's a bit hacky.

What does it add to the history stack exactly? I mean is there a way we could filter out something we don't want or detect that we need to go back -2 instead -1? Just out of curiosity did you have any other ideas before you settled on this one?

Marc-Roig (approved):
that sounds like you had fun debugging this one 😄

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
remidej: It pushes state to the browser's history, so the URL is the same but it creates a new entry. I found the issue [there](https://arc.net/l/quote/fcgnvciv).

With that in mind there are gonna be several possible fixes.
- call an imperative location.replace API using a ref and an effect
- we don't have access to the browser's history stack, so we can't filter on it. We could keep track of clicks on the tabs to go back to -N. But then we'll still have extra entries in the history stack, so the ability to click "forward" when it's not gonna change anything

i feel like it's gonna be hacky either way 🤷

markkaylor: 👍

markkaylor (approved):

Issue #21349, opened 2024-09-18
[Design] Loader isn’t a circle when generating new entry


yanniskadiri:
![Image](https://github.com/user-attachments/assets/d7041020-3231-44b0-ae19-feb46582b175)

sheeshhhhhh:
I'd like to work on this issue, can I be assigned? im a first timer btw :)

dilshad-knk:
Is there design file for this

PR #22433, opened 2024-12-13
fix: make ctb loader truly round

lucasboilly:
### What does it do?

- Makes the CTB loader round
- Harmonizes the icons sizes between the 2 states

### Why is it needed?

It's currently not the case.

#### Before

https://github.com/user-attachments/assets/451077b7-6ca9-4e5b-921d-adf55c20205b

#### After

https://github.com/user-attachments/assets/e6a0d05a-cf35-4612-a8e1-d4a57766cd35

### How to test it?

Create or update any content type.

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/21349


remidej (approved):
❤️❤️❤️

PR #22434, opened 2024-12-13
Fix: logging data-transfer and adding verbose

Bassel17:
### What does it do?

Backporting the dts enhancement recently added for data-transfer in V5 to have `--verbose` option
fixing logging a file during a transfer

### Why is it needed?

- `--verbose` option is need for v4 projects

### How to test it?

- Add `--verbose` to transfer or import or export commands it should log a file with information about the transfer

### Related issue(s)/PR(s)

#22085 #22214


Convly (approved):
LGTM, tested with a local & remote apps

Issue #22426, opened 2024-12-13
[Homepage] Floating widgets when no data to display


MarionLemaire:
### Required System information
- Community Edition
- Mac
- Local

### Describe the bug

When there are no data to display in the widgets, the “No edited entries” and "No published entries" logos are floating,  don’t have any entry created yet and I just see No entries edited and No entries published without the boxes they should be in.

### Steps to reproduce the behavior

1. Login with Super Admin without any entry
2. Check the Homepage
3. Widgets are not displayed under the Guided tour but “No edited entries” logo are floating

### Expected behavior

 logos should be in a card

### Screenshots

![image](https://github.com/user-attachments/assets/0f124b0a-3c3a-4318-8a4d-4e0aab9dcc41)
![image](https://github.com/user-attachments/assets/78e44b55-7516-4fb5-b673-683e18f5e4d8)

PR #22448, opened 2024-12-16
fix(homepage): add missing styles for error, loading, and empty states

markkaylor:
### What does it do?

- Refactors the root component to handle every state

### Why is it needed?

To match the designs

### How to test it?

Check the designs

Screenshot 2024-12-16 at 11 49 47
Screenshot 2024-12-16 at 11 49 33
Screenshot 2024-12-16 at 11 49 24

### Related issue(s)/PR(s)

resolves https://github.com/strapi/strapi/issues/22426

### DIFF
--- a/packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
+++ b/packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
@@ -48,10 +48,15 @@ const DocumentStatus = ({ status = 'draft' }: DocumentStatusProps) => {
   );
 };
 
-const WidgetContent = ({ document }: { document: RecentDocument }) => {
+/* -------------------------------------------------------------------------------------------------
+ * LastEditedWidget
+ * -----------------------------------------------------------------------------------------------*/
+
+const LastEditedWidgetContent = () => {
   const { formatMessage } = useIntl();
   const { trackUsage } = useTracking();
   const navigate = useNavigate();
+  const { data, isLoading, error } = useGetRecentDocumentsQuery({ action: 'update' });
 
   const handleRowClick = (document: RecentDocument) => () => {
     trackUsage('willEditEntryFromHome');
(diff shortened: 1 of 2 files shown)


remidej (commented):
I'm not sure this is the design spec. I think the loading/error/nodata state show be shown inside the shared widget UI below the header. Here it looks like they're replacing the header entirely

markkaylor:
> I'm not sure this is the design spec. I think the loading/error/nodata state show be shown inside the shared widget UI below the header. Here it looks like they're replacing the header entirely

oops yeah you are right thanks!

packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
```diff
@@ -118,21 +118,49 @@ const LastEditedWidget = () => {
...
+        title={{
+          id: 'content-manager.widget.last-edited.title',
+          defaultMessage: 'Last edited entries',
+        }}
+        icon={Pencil}
+      >
+        <Widget.Loading />
+      </Widget.Root>
+    );
   }
```
remidej, line 132: I don't really like this approach for 2 reasons:
- this means recreating the root component everytime the state changes, even though that part of the ui is constant. it means some rerenders/repaints that can be avoided
- it's a bit further from the way homepage will work when we add the API. It's Strapi that's gonna render the Root component, not the widget developers, they'll only have access to the content box inside of it.

So I would recommend something closer to what I had on the 1st widget PR:
- a LastEditedContent + LastPublishedContent component that manages the different states (this is the part that will later get registered in the API)
- a LastEditedWidget that just wraps the Content in the Root (this will be the job of the admin in the future)

so I would recommend something

markkaylor: >this means recreating the root component everytime the state changes, even though that part of the ui is constant. it means some rerenders/repaints that can be avoided

I see what you mean makes sense.

>a LastEditedContent + LastPublishedContent component that manages the different states (this is the part that will later get registered in the API)

I find it unfortunate to create and maintain two components that are pretty much the exact same thing but if that's how you want to proceed, works for me

remidej: I'm happy to discuss it!

> pretty much the exact same thing

Are you referring to LastEditedContent + LastPublishedContent being very similar? or the split between the Widget and the Content?

markkaylor: LastEditedContent + LastPublishedContent <-- this one

I started working on it and I think your suggestion has us down the right path, I'll implement it that way then maybe we can take a quick look together tomorrow to see if we can refactor anything 👍

remidej: sounds good!

packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
```diff
@@ -49,84 +41,98 @@ const DocumentStatus = ({ status = 'draft' }: DocumentStatusProps) => {
...
   const { formatMessage } = useIntl();
   const { trackUsage } = useTracking();
   const navigate = useNavigate();
 
+  const getEditViewLink = (document: RecentDocument): string => {
+    // TODO: import the constants for this once the code is moved to the CM package
+    const kindPath = document.kind === 'singleType' ? 'single-types' : 'collection-types';
+    const queryParams = document.locale ? `?plugins[i18n][locale]=${document.locale}` : '';
+
+    return `/content-manager/${kindPath}/${document.contentTypeUid}/${document.documentId}${queryParams}`;
```
Marc-Roig, line 54: should the documentId be in there for single types?

simotae14 (approved):
Code looks good to me, I let you address @Marc-Roig comment.
I need to test the solution once is fully approved

packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
remidej: you're right! it somehow didn't 404 🤷

Marc-Roig (approved):

Issue #18526, opened 2023-10-20
Koa types


mbhaskar98:
### Required System information

- Node.js version: 18.16.1
- NPM version: 9.8.0
- Strapi version: 4.13.7
- Database: mysql2
- Operating system: darwin-x64
- Is your project Javascript or Typescript: Typescript

### Describe the bug

To get auto complete for context object (type Koa context) in controllers or middlewares I installed @types/Koa. The body object according to Koa Context type is available directly with the context object but the at runtime it is undefined. At runtime the body is found under ctx.request.body["data"] but accoeding to Koa Context there is no body object under request object. 

### Expected behavior

Is TypeScript with Koa available? Is the context object actualy Koa context? 

### Screenshots

<img width="842" alt="Strapi-context-issue" src="https://github.com/strapi/strapi/assets/46974232/db5d3d51-8d27-47e0-aa8b-19bb3de50b44">

Madanpurohit:
If this is issue i will be happy to contribute.

nick-aj:
Hello @Madanpurohit,

Thank for submitting this issue with us, we can confirm that we were able to reproduce it.
_Here are my details:_
For the following extended core controller:
```javascript
module.exports = createCoreController('api::order.order', ({ strapi }) => ({

        async createCompleteOrder(ctx, next) {

            // Order Fields: products, owner, cofirmed, confirmation_data
            // TODO (Implement isOwner policy)
            const user = ctx.state.user;
            const order = await strapi.entityService.create("api::order.order", {
                data: {
                    products: ctx.request.body.products,
                    owner: user.id
                }
            });
```
The following error is thrown:
```javascript
Property 'body' does not exist on type 'Request'.ts(2339)
```

hekmatnajafi:
hey @nick-aj is there any update?
today I faced the same issue!

hekmatnajafi:
a quick fix for me was `npm install --save-dev @types/koa-bodyparser`

iamandrewluca:
> a quick fix for me was `npm install --save-dev @types/koa-bodyparser`

Strapi v4 is using `koa-body` instead of `koa-bodyparser`.

To fix that, you need to reference `koa-body`

```ts
/// <reference types="koa-body" />
```

For some reason, Strapi v4 has `koa-bodyparser` as a dependency in some places, probably also because of types, but in reality is using `koa-body`

https://github.com/strapi/strapi/blob/v4/packages/core/strapi/src/middlewares/body.ts#L3

PR #22453, opened 2024-12-17
fix(ts): use correct koa-body types for request.body

iamandrewluca:
### What does it do?

Uses the correct type definitions for `Koa.Context['request']['body']`

### Why is it needed?

`koa-bodyparser` is used only by `apollo-server-koa`

### How to test it?

```bash
npm run test:ts
```

### Related issue(s)/PR(s)

- https://github.com/strapi/strapi/issues/18526#issuecomment-2547788666
- https://github.com/strapi/strapi/blob/v4/packages/core/strapi/src/middlewares/body.ts#L3
- https://github.com/koajs/koa-body/blob/v4.2.0/index.d.ts#L4-L9
- https://github.com/apollographql/apollo-server/blob/release-3.10.0/packages/apollo-server-koa/src/ApolloServer.ts#L4C25-L4C39

### DIFF
--- a/packages/core/admin/ee/server/src/controllers/index.ts
+++ b/packages/core/admin/ee/server/src/controllers/index.ts
@@ -1,4 +1,4 @@
-import 'koa-bodyparser';
+import 'koa-body';
 
 import authentication from './authentication';
 import role from './role';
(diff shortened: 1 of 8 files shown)


hanpaine:
hey @iamandrewluca we are only merging changes addressing critical and high bugs for V4 now I'm afraid - thanks for your contribution 🙏

PR #22454, opened 2024-12-17
fix: remove lookup param from document service

innerdvations:
### What does it do?

remove lookup param from document service

### Why is it needed?

it's for internal use only

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/services/document-service/draft-and-publish.ts
+++ b/packages/core/core/src/services/document-service/draft-and-publish.ts
@@ -70,6 +70,7 @@ const statusToLookup: TransformWithContentType = (contentType, params) => {
     return params;
   }
 
+  // @ts-expect-error: we need to create a different typing for internal params
   const lookup = params.lookup || {};
 
   switch (params?.status) {
--- a/packages/core/core/src/services/document-service/draft-and-publish.ts
+++ b/packages/core/core/src/services/document-service/draft-and-publish.ts
@@ -70,7 +70,6 @@ const statusToLookup: TransformWithContentType = (contentType, params) => {
     return params;
   }
 
-  // @ts-expect-error: we need to create a different typing for internal params
   const lookup = params.lookup || {};
 
   switch (params?.status) {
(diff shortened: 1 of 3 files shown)


Boegie19:
@innerdvations Are we sure we want to throw an error here and not just ignore the input like we do with any other invalid input?

alexandrebodin (approved):
LGTM

innerdvations:
> @innerdvations Are we sure we want to throw an error here and not just ignore the input like we do with any other invalid input? makes no sence to error for this if for all other invalid input we just ignore it

For other unknown fields, they are ignored completely, but since this one is actually a valid internal field, we will throw to make it visible that it is not allowed

Marc-Roig (approved):

Boegie19 (approved):
LGTM.

I see one other alternative we add a check in https://github.com/strapi/strapi/blob/b6435ada233136a0d0b14fba67961ff8f16cdac2/packages/core/core/src/middlewares/query.ts

that way we can block it higer what means that the you can use lookup in the document service to bypass private fields but you can't use it from the api

innerdvations:
> LGTM.
> 
> I see one other alternative we add a check in [`b6435ad`/packages/core/core/src/middlewares/query.ts](https://github.com/strapi/strapi/blob/b6435ada233136a0d0b14fba67961ff8f16cdac2/packages/core/core/src/middlewares/query.ts)
> 
> that way we can block it higer what means that the you can use lookup in the document service to bypass private fields but you can't use it from the api

For now we want `lookup` to be internal-only, so this will keep it that way

Boegie19:
@innerdvations One final thing then
https://github.com/strapi/strapi/blob/b6435ada233136a0d0b14fba67961ff8f16cdac2/packages/core/types/src/modules/documents/params/document-engine.ts

in the file we see that document service types say that lookup is a valid input since it will always error when used we should probebly rmeove it from the ts types.

alexandrebodin:
@Boegie19 we need some refactor to fix the typigns cleanly we will do it right after

Issue #20383, opened 2024-05-28
[v5] Using pagination query params at root level of query returns incorrect error


innerdvations:
### Required System information

Launched In: 646 ms
Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0-beta.8
Node/Yarn Version: yarn/4.0.1 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite

### Describe the bug

Sending a query like `?page=1&pageSize=10` instead of `?pagination[page]=1&pagination[pageSize]=10` returns the error:

```json
{
    "data": null,
    "error": {
        "status": 400,
        "name": "PaginationError",
        "message": "Invalid pagination attributes. You cannot use page and offset pagination in the same query",
        "details": {}
    }
}
```

The error is incorrect and very misleading.

### Steps to reproduce the behavior

See description of bug above

### Expected behavior

I would expect one of the following (haven't put too much thought into it):
- a validation error saying that page parameters are incorrect and must be in the pagination object
OR
- for the parameters to work anyway
OR
- for nothing to happen with pagination; since page and pageSize are not in the pagination object they should be ignored and the defaults should be used

hanpaine:
Lowered as this is about messaging not functionality

PR #22455, opened 2024-12-17
fix: pagination query params at root level of query return incorrect …

pandeyprafull:
### What does it do?
pagination query params at root level of query return incorrect error message

### Why is it needed?
To fix issue [20383](https://github.com/strapi/strapi/issues/20383) : pagination query params at root level of query return incorrect error message

### How to test it?

1. Create multiple entries of any collection.
2. Fetch entries of that collection using REST api using query `?page=1&pageSize=10` instead of `?pagination[page]=1&pagination[pageSize]=10`
3. Getting improper error message in the response.

### Related issue(s)/PR(s)
fixes [20383](https://github.com/strapi/strapi/issues/20383)


Bassel17 (approved):
LGTM!

thanks for the contribution

Issue #20782, opened 2024-07-16
[Design] A border appear when clicking on the avatar and the popover is not aligned with it


MarionLemaire:
https://github.com/user-attachments/assets/d4c41f0a-6cc8-44cb-81e8-fe8f67eab898

PapatMayuri:
@MarionLemaire can you confirm like how that popover needs to be shown when we click on avatar.
adding 2 screenshots for your reference, can you please check and confirm the correct popover UI.
![Image](https://github.com/user-attachments/assets/b9af4f06-0e39-47d6-8751-2e9458380808)
![Image](https://github.com/user-attachments/assets/ace00ffa-5939-4170-87a0-ce2f252d18e8)

Please add if you have any other UI for popover.

Adzouz:
Hey @MarionLemaire, this one was fixed during my work on navigation responsiveness https://github.com/strapi/strapi/pull/22460. Closing it! 🚀

PR #22457, opened 2024-12-17
fix: ui issues of the avatar button

lucasboilly:
### What does it do?

- Gives a fixed width and height to `MenuTrigger`.
- Makes `MenuTrigger` being the same size as the `Avatar`.
- Moves the border styles and paddings from `MenuTrigger` to `Flex`.
- Adjusts `MenuContent` left margin to make it aligned with the `Avatar`.

#### Before

https://github.com/user-attachments/assets/be01b260-6b4a-4cc1-8c6f-fa01e13ef0c9

#### After

https://github.com/user-attachments/assets/baf03cac-ff48-4741-92be-85d2e606a091

### Why is it needed?

It was looking buggy.

### How to test it?

Login to Strapi and check the MainNav.

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20782


lucasboilly:
I had to merge latest changes from `develop` before publishing my branch but it seems it created issues in the tests.

CleanShot 2024-12-17 at 18 47 25@2x

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
```diff
@@ -60,7 +59,14 @@ export const NavUser = ({ children, initials, ...props }: NavUserProps) => {
   };
 
   return (
-    <Flex justifyContent="center" {...props}>
+    <Flex
+      justifyContent="center"
+      {...props}
```
remidej, line 64: I think the spread props should probably be last. It's usually a way to let the place calling the component (the place that does <NavUser />) override the default styles of the component

remidej (commented):
You need to fix the logout end to end test, not sure why your changes affect it though. It's probably a good idea to replace the failing selector by getByRole at least to ensure better accessibility:
```ts
// replace this:
await page.getByText('test testing').click();

// likely by this:
await page.getByRole('button', {name: 'test testing').click();
```

also are you sure you wanna make the touch target smaller?

lucasboilly:
@remidej do I need to do it for the 3 lines? Or only the one you mentioned?

```ts
await page.getByText('test testing').click();
await page.getByText('Logout').click();

await expect(page.getByText('Log in to your Strapi account')).toBeVisible();
```
Regarding the smaller touch target, since the `Avatar` size is as big as any `IconButton` that's not an issue.

remidej:
@lucasboilly you should do it for the first 2, for the elements that are clicked. It's a way to ensure anything that needs to be clicked has the right HTML tag, otherwise it won't be "clickable" with a keyboard or screen reader

lucasboilly:
@remidej it still blocked 🤔

tests/e2e/tests/admin/logout.spec.ts
```diff
@@ -11,8 +11,8 @@ test.describe('Log Out', () => {
   });
 
   test('a user should be able to logout', async ({ page }) => {
-    await page.getByText('test testing').click();
-    await page.getByText('Logout').click();
+    await page.getByRole('button', { name: 'test testing' }).click();
+    await page.getByRole('button', { name: 'Logout' }).click();
```
simotae14, line 15: hey @lucasboilly to pass the e2e test you need to change this line
`await page.getByRole('button', { name: 'Logout' }).click();`
with this one
`await page.getByText(/logout/i).click();`

simotae14 (dismissed):
Hey
to pass the e2e tests you need to modify the locator
Well done btw

remidej (approved):

PR #22464, opened 2024-12-18
fix: renable the basic relations test

simotae14:
### What does it do?

Remove the skip to the api test and replace the id with documentId and the correct response object

### Why is it needed?

The test was disabled during the implementation of v5

### How to test it?

- from the root run the api test
`yarn test:api tests/api/core/content-manager/api/basic-relations.test.api.js`

### Related issue(s)/PR(s)

CS-1193


remidej (approved):
lgtm if the ci passes :)

PR #22465, opened 2024-12-18
chore(telemetry): smooth event rate limiting

Baldinof:
### What does it do?

Rate limited events now resets 24h after the last event (or process start), instead of each day at midnight.

### Why is it needed?

It will smooth the sending of events, and avoid receiving a burst of events at midnight on the telemetry server.

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/services/metrics/__tests__/middleware.test.ts
+++ b/packages/core/core/src/services/metrics/__tests__/middleware.test.ts
@@ -1,6 +1,12 @@
 import createMiddleware from '../middleware';
 
 describe('Metrics middleware', () => {
+  const originalDateNow = Date.now;
+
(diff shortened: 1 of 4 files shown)


Convly (changes requested):
Code and behavior LGTM.

Could you run `yarn format:code` to solve the prettier issue?

Convly (approved):

innerdvations (approved):
LGTM, but I'm also curious why it's only once per day and not something like every 6 hours or 12 hours to smooth it out more

Baldinof:
Mainly to keep the rate the same, if it is reset every 12h, rate limited events will send more events.

Issue #21359, opened 2024-09-18
[Relations] Both Modified and Published entries are displayed in my Relation field


yanniskadiri:
https://github.com/user-attachments/assets/1bf3fd6d-43d8-4446-9d51-901f53882a6d

Marc-Roig:
This one should be fixed as of the latest v5 version

Marc-Roig:
It was noted by a community user that this issue might still be happening https://github.com/strapi/strapi/pull/22478#issuecomment-2780601419

PR #22478, opened 2024-12-19
fix: modified and published both entries are displayed in relation fi…

pandeyprafull:
### What does it do?
Fixes the modified and published both entries are displayed in relation field selection dropdown

### Why is it needed?
To fix issue [21359](https://github.com/strapi/strapi/issues/21359) : Both Modified and Published entries are displayed in my Relation field

### How to test it?

1. Create multiple entries with Daft/Modified/Published statuses of a collection type
2. Map that collection type into another collection type as 1:1 or 1:N relation
3. Now if you open the relation field dropdown ton select you will get all states of same entries

### Related issue(s)/PR(s)
fixes [21359](https://github.com/strapi/strapi/issues/21359)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -507,7 +507,7 @@ const RelationsInput = ({
 
   const hasNextPage = data?.pagination ? data.pagination.page < data.pagination.pageCount : false;
 
-  const options = data?.results ?? [];
+  const options = data?.results?.filter((opt) => opt.status === 'published') ?? [];
 
   const handleChange = (relationId?: string) => {
     if (!relationId) {


hanpaine:
Thanks for another fix @pandeyprafull ! We'll take a look in the new year ✨

Marc-Roig:
thank you for the contribution! just realized this has already been fixed as the latest version of strapi v5

marabouq:
@Marc-Roig thank you so much for your contribution of closing this task, but could you provide more details when you plan to drop this "latest version of strapi v5" with this absolutely required functionality fixed? It's not fixed in latest published release [v5.12.3](https://github.com/strapi/strapi/releases/tag/v5.12.3). Is it in your 2025 roadmap to release this fix?
Thank you in advance!

Marc-Roig:
@marabouq were you able to reproduce it in the latest releaese? I will reopen if thats the case

Marc-Roig:
looking at the fix I am not sure this is exactly what we want, I will reopen the issue, and close this one

oleza:
@Marc-Roig, it's still reproducible, but only with v4 compatibility enabled. So, it's not as high a priority as I thought at first, but thanks!

marabouq:
@Marc-Roig, it's still reproducible, but only with the v4 compatibility flag enabled, so it's not as high a priority as I thought at first. But thanks!

Marc-Roig:
aaaah! 💡 completely missed that one, that has to be fixed for sure then too.

Issue #20809, opened 2024-07-16
[Design] Missing overlay and oversized “Cancel” button for the Dialog modal of Bulk Publish


MarionLemaire:
**Current design**
![CleanShot 2024-07-05 at 11 09 18@2x](https://github.com/user-attachments/assets/774f6094-5a59-4e3b-a0d0-4bb774f6506a)

**Expected design**
![Image](https://github.com/user-attachments/assets/c98daa3e-9745-4c1b-8f64-b6016730022d)

Figma file here: https://www.figma.com/design/Tt8uHuQezEP0CsWunbulb5/Draft-%26-Publish?node-id=6933-37751&node-type=instance&t=VVDKqUnnTDyHSSup-0

Adzouz:
Thanks a lot for your fix @PapatMayuri 🚀

PR #22481, opened 2024-12-19
fix: fixed cancel and publish button from dialog modal of bulk publish

PapatMayuri:
### What does it do?
Fixed cancel and publish button from dialog modal of bulk publish
### Why is it needed?
To fix issue [20809](https://github.com/strapi/strapi/issues/20809): [Design] Missing overlay and oversized “Cancel” button for the Dialog modal of Bulk Publish
### How to test it?
- go to CM
- select any one collection type then
- add more than one entries then select all and click on publish button
- one popup will be open with details then click on Publish we will get confirmation popup 
- in that popup cancel button is in fullwidth 
### Related issue(s)/PR(s)
fixes [20809](https://github.com/strapi/strapi/issues/20809)


Adzouz (approved):
Tested locally, thanks a lot! 🚀

Issue #21341, opened 2024-09-18
[Design] Missing the pointer cursor when on top of a possible relation


yanniskadiri:
![Image](https://github.com/user-attachments/assets/7b029322-98e1-4d3c-8d97-b3c0a9722f93)

PR #22511, opened 2024-12-26
fix: missing pointer cursor when hover top of possible relations

pandeyprafull:
### What does it do?
Fix missing pointer cursor when hover top of possible relations

### Why is it needed?
To fix issue [21341](https://github.com/strapi/strapi/issues/21341) : Missing the pointer cursor when on top of a possible

### How to test it?

1. Go to CTB
2. Create relation field and hover over the possible relations.
3. Now you will not be able to see the pointer cursor.

### Related issue(s)/PR(s)
fixes [21341](https://github.com/strapi/strapi/issues/21341)


butcherZ (approved):
Tested and it fixes the cursor issue, thanks for the contribution :)

PR #22521, opened 2024-12-28
enhancement: add customField parameter to extendFields

giollord:
enhancement: add customField parameter to args in app.getPlugin('content-type-builder').apis.forms.extendFields(..., {form: {advanced(args) => {...}}})

### What does it do?

Adds customField argument that can be used in plugin bootstrap in `app.getPlugin('content-type-builder').apis.forms.extendFields(..., {form: {advanced(args) => {...}}})`. Example with slightly modified i18n/admin/src/index.ts :
<details>
<summary>Example with slightly modified i18n/admin/src/index.ts</summary>

</details>

### Why is it needed?

The solution allows plugin developers to have more flexibility when designing custom fields. When writing code in `advanced(args) => {...}`, at the moment there is no way to know whether current field is custom or built-in, even name of the field is unknown. Using customField argument along with contentTypeSchema (already there) allows to know what the current field exactly is.\
It is useful in case if there are few custom 'json' field types created and when creating content type, each field needs it's own configuration.

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/FormModal/forms/forms.ts
+++ b/packages/core/content-type-builder/admin/src/components/FormModal/forms/forms.ts
@@ -98,6 +98,7 @@ export const forms = {
           data,
           type: customField.type,
           step,
+          customField,
           ...rest,
         });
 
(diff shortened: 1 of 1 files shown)


hanpaine:
Thanks @giollord , we'll take a look soon 🙏

PR #22525, opened 2024-12-29
enhancement: added Latvian machine translation

giollord:
### What does it do?
Added machine translation to Latvian language for some language jsons using GitHub Copilot (based on GPT 4o). I did sanity check and it looks very good, even format.js masks and words like 'webhook' are intact. English and sometimes German (where English was missing from json) were used as base for the translation.

### Why is it needed?
So that there are more languages available in user interface


hanpaine:
Hi @giollord, thanks for your time on this. I'm afraid that we are not able to accept new languages as we are not equipped to review and validate the quality of them suitably nor maintain them.

You are very welcome to publish it for the community in your own repo and then using
https://docs.strapi.io/dev-docs/admin-panel-customization/options#extending-translations someone could use them 🙏

Thank you again for your contribution !!

giollord:
Hi, @hanpaine ! Thank you for reviewing! Yes, I understand that it might be risky to merge such PRs :)
But would it be possible to merge if I create PR with a single file [languageNativeNames.ts](https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/translations/languageNativeNames.ts) that adds Latvian language with `lv: 'Latviešu',` ? The problem is that without modifying this file there seem to be no way to add **new** interface language from plugin or configuration - it is only possible to extend languages that already exist in the file.

giollord:
update: made https://github.com/strapi/strapi/pull/22596 instead

PR #22537, opened 2025-01-01
fix: handle clearing of number field to send null instead of undefined

### LINKED ISSUE #22533: Clearing the value of a number field sends the value as undefined rather than null

- Node.js version: v20
- NPM version: v10
- Strapi version: 4.6.0
- Database: SQLite
- Operating system: Linux Mint 22

### Describe the bug

When setting any other normal field (string, ect) to empty it will send that information as `null` rather than `undefined` however with number fields (in this case a decimal field) it sends the value as undefined which is interrupted as nothing has been changed and causing the field to keep the same value.

1. Create a content-type with a decimal field
2. Create an entry with a value in that field
3. Try to update the value and remove it
4. See that the previous data remains

### Expected behavior

Clearing the value should send the update as `null` rather than `undefined`

### Screenshots

https://www.loom.com/share/86695a415171483aadb200a0a8169e40?sid=dcde0622-452a-44f9-a472-1270d907394a

### Code snippets

If needed here is the schema used and the document service middleware for easier debugging:

Schema:

```json
{
  "kind": "collectionType",
  "collectionName": "tests",
  "info": {
    "singularName": "test",
    "pluralName": "tests",
    "displayName": "test"
  },
  "options": {
    "draftAndPublish": true
  },
  "pluginOptions": {},
  "attributes": {
    "string": {
      "type": "string"
    },
    "number": {
      "type": "decimal",
      "default": 0
    }
  }
}
```

document service middleware

```js
"use strict";

module.exports = {
  /**
   * An asynchronous register function that runs before
   * your application is initialized.
   *
   * This gives you an opportunity to extend code.
   */
  async register(/*{ strapi }*/) {
    const applyTo = ["api::test.test"];

      // Only run for certain content types
      if (!applyTo.includes(context.uid)) {
        return next();
      }

      // Only run for certain actions
      if (["create", "update"].includes(context.action)) {
        console.log(context);
      }

      const result = await next();

      // do something with the result before returning it
      return result;
    });
  },

  /**
   * An asynchronous bootstrap function that runs before
   * your application gets started.
   *
   * This gives you an opportunity to set up your data model,
   * run jobs, or perform some special logic.
   */
  bootstrap(/*{ strapi }*/) {},
};
```

### Additional context

Related to TID10590

Dhruv-Maradiya:
Ensure that when a number field is cleared, its value is explicitly set to null instead of undefined to guarantee the update is processed correctly.

### What does it do?

Fixes #22533 by explicitly setting the value to null when a number field is cleared. Previously, the value was set to undefined, which was interpreted as no change, preventing the number field from being cleared.

### Why is it needed?

This fix is necessary to enable clearing a number field. Without it, passing undefined results in no change, leaving the field uncleared.

### How to test it?

1. Navigate to the content-manager section.
2. Clear the value of a number input field.
3. Reload the page.
4. Verify that the value of the cleared field is null.

### Related issue(s)/PR(s)

Related to #22533.

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Number.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Number.tsx
@@ -9,7 +9,7 @@ import { InputProps } from './types';
 
 const NumberInputImpl = forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, type, ...props }, ref) => {
-    const field = useField<number>(name);
+    const field = useField<number | null>(name);
     const fieldRef = useFocusInputField<HTMLInputElement>(name);
 
     const composedRefs = useComposedRefs(ref, fieldRef);
@@ -20,10 +20,12 @@ const NumberInputImpl = forwardRef<HTMLInputElement, InputProps>(
         <NumberInput
           ref={composedRefs}
           onValueChange={(value) => {
-            field.onChange(name, value);
+            // Convert undefined to null to store it in the form state
+            // See https://github.com/strapi/strapi/issues/22533
+            field.onChange(name, value ?? null);
           }}
           step={type === 'float' || type == 'decimal' ? 0.01 : 1}
-          value={field.value}
+          value={field.value ?? undefined}
           {...props}
         />
         <Field.Hint />
--- a/tests/api/core/content-manager/fields/integer.test.api.js
+++ b/tests/api/core/content-manager/fields/integer.test.api.js
@@ -92,4 +92,27 @@ describe('Test type integer', () => {
       field: 543,
     });
   });
+
+  test('Clearing the value should set it to null', async () => {
+    const res = await rq.post('/content-manager/collection-types/api::withinteger.withinteger', {
+      body: {
+        field: 123,
+      },
+    });
+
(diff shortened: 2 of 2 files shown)


Dhruv-Maradiya:
Hello 👋,

This is my first contribution to Strapi! I noticed the issue labeled as `severity: high` and decided to work on it because I believe the fix was straightforward and I could address it quickly. 

I didn’t announce beforehand that I was working on this issue—my apologies if that caused any overlap. I’ll ensure better communication in the future. I hope this PR helps resolve the issue effectively!

Looking forward to feedback and contributing more to Strapi. 😊

hanpaine:
hey @Dhruv-Maradiya thanks for contributing! Kudos on going in big on your first PR 😁 the team will try and take a look next week 🙏

remidej (approved):
Thanks for the PR! Code looks good and I confirm it does fix the issue. Not sure why github isn't running the tests, I'll look into it

Dhruv-Maradiya:
Thank you so much for reviewing the PR and for your kind words! 😊 I really appreciate the feedback and support. Let me know if there’s anything else I can help with.

Issue #21354, opened 2024-09-18
Relation in a component starts bugging after adding more components


laurenskling:
### Required System information

- Node.js version: 20
- NPM version: yarn 1.22
- Strapi version: 5.0.0 (or rc 29)
- Database: sqlite
- Operating system: macos
- Is your project Javascript or Typescript: js

### Describe the bug

I've made a video trying to explain. It's very weird. It happens when you start adding more components into your repeatable. It will start 404-ing on loading the relation options. instead of  fetching `http://0.0.0.0:1337/content-manager/relations/pagebuilder.section/section?id=8&pageSize=10&_q=&page=1`, where 8 is the component number, it will start fetching `http://0.0.0.0:1337/content-manager/relations/pagebuilder.section/section?id=d0szffkl8hqp92ye7cdqk42e&pageSize=10&_q=&page=1` where `d0szffkl8hqp92ye7cdqk42e` is the current page i'm on.

I'm not sure why the `id` is added in the fetch call, because both seem wrong... component 8 could be anywhere. `section` does not know about components relating to it, so it shouldn't be able to handle this `8`. 
Also, adding my current `page` (contentType) documentId doesn't make sense either, because it is not the page, but the component relating to it. Thats probably why it's 404ing. It doesn't know what to do with this documentId.

### Steps to reproduce the behavior

1. Create a contentType, containing a repeatable component, that contains a relation.
2. add an entry with some components
3. save, refresh the page to get rid of cached data
4. see it all bug out. In my video, I've made an 11th entry, which i can never find again.

### Screenshots

https://github.com/user-attachments/assets/4d81d013-8778-4e22-b92f-60398eba7a92

sorry for the low res video, im only allowed 10mb.
I've added the original on discord: https://discord.com/channels/811989166782021633/1095091586452426824/1286021240842223678

Marc-Roig:
@laurenskling hey! I believe this issue might be fixed after https://github.com/strapi/strapi/pull/21428

I am checking the queries made and they are looking good, relations in components should refernce the id of the component, and relations in the top level should reference the document id.

Let me know if you are still seeing this issue and its a content type setup difference, and I will gladly reopen the issue!

laurenskling:
Hey @Marc-Roig , 
almost. I have a repeatable component with a relation inside. 
- when editing an existing entry
- relations in component repeatable load by their id
- adding a new row into the repeatable, loads with the parent documentId

in this screenshot i opened 2 existing component and pressed "Add an entry"
[...]

PR #22549, opened 2025-01-03
fix: repeatable components should not load relations when creating a new entry

laurenskling:
### What does it do?

When pressing "add an entry" on a repeatable component, the RelationsFields starts looking for the selected relation with the component id. Which doesn't exist yet. So it falls back to the documentId of the parent document this component is inside of. This is wrong and returns a `404`.

### Why is it needed?

Making sure the incorrect API call is not triggered.

### How to test it?

The example project in the Strapi repo contains a `Menu Section` contentType that contains a repeatable component with a relation inside called `categories`. If you press "add an entry" here, an incorrect API call is triggered. This PR fixes this. It should not trigger the `useGetRelationsQuery` . (note: it does trigger `useLazySearchRelationsQuery` which is correct and looks alike)

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/21354


hanpaine:
Hey @laurenskling thanks for another contrib - I'll ping the team on this directly as I see it's been long running/was hoped resolved but wasn't quite all there 🙏

laurenskling:
hm, I tried to sync my PR now this happened.

laurenskling:
Successfully updated

PR #22278, opened 2024-11-26
fix(history): handle nested component images

markkaylor:
### What does it do?

- Handle possible null values on the frontend
- Update the backend to handle images inside components

These changes resolve the related issue and should display images at an infinite level of nesting

### Why is it needed?

Describe the issue you are solving.

### How to test it?

- Create a component that has an image, and another component inside it that also has an image. (note: I think self nesting is broken so please don't try to use the same component inside itself)
- Add this component to a content type
- Also add a sibling component that has an image
- Now go to your content type and add images to all your components
- Save, go to content history

You should see all your images at every level at nesting:

Screenshot 2024-11-26 at 14 23 25

### Related issue(s)/PR(s)

resolves https://github.com/strapi/strapi/issues/22060

--- 

## Help Wanted!

Unfortunately I am unable to restore a history version with components that have images.

The document service is throwing the following errors when [it calls update](https://github.com/strapi/strapi/blob/7d575f66bf2575073780538336ccfcdbe1eaca19/packages/core/content-manager/server/src/history/services/history.ts#L273-L278)

Screenshot 2024-11-26 at 14 29 55

PR #22573, opened 2025-01-06
chore(docs): document traverseEntity util

markkaylor:
### What does it do?

Adds documentation for the `traverseEntity` utility

### Why is it needed?

So it is easier for everyone to understand and use the utility

### How to test it?

Read it 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22278

p.s.

ChatGPT did a good job IMO

### DIFF
--- a/docs/docs/docs/01-core/utils/traverse-entity.md
+++ b/docs/docs/docs/01-core/utils/traverse-entity.md
@@ -4,7 +4,6 @@ tags:
   - utils
 ---
 
-
 # Traverse Entity Utility
 
 The `traverseEntity` function is designed to recursively traverse a data entity based on its schema definition. It provides hooks to execute custom logic at each level of the traversal through the `Visitor` function.
@@ -12,11 +11,13 @@ The `traverseEntity` function is designed to recursively traverse a data entity 
 ## Overview
 
 ### Purpose
+
 The utility is used for structured data manipulation. It supports various types of attributes, such as relations, components, dynamic zones, and media, allowing seamless traversal and transformation of deeply nested entities.
 
 ## Function: `traverseEntity`
 
 ### Signature
+
 ```typescript
 async function traverseEntity(
   visitor: [Visitor](#visitoroptions),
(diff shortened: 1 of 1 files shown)


docs/docs/docs/01-core/utils/traverse-entity.md
```diff
@@ -0,0 +1,142 @@
...
+  - `parent` *(Parent, optional)*: Information about the parent attribute.
+  - `getModel` *(Function)*: A function to retrieve the schema of a model by its UID.
+
+- **`entity`**:  
+  The data entity to be traversed.
+
+### Returns
+A `Promise<Data>` that resolves with the transformed entity.
+
+---
```
remidej, line 49: I'm curious why you use these separators? If it's for styling purposes I feel like it's better to update the headings' CSS than to add these

docs/docs/docs/01-core/utils/traverse-entity.md
```diff
@@ -0,0 +1,142 @@
...
+### Path
+```typescript
+interface Path {
+  raw: string | null;
+  attribute: string | null;
+}
+```
+Tracks the traversal path:
+- **`raw`**: The raw path of the current attribute in the entity.  
+- **`attribute`**: The schema-defined attribute path.
```
remidej, line 112: these 2 are not clear to me, can you add examples?

remidej (commented):
Thanks a lot for this!

docs/docs/docs/01-core/utils/traverse-entity.md
markkaylor: Good catch, I don't completely understand the difference. They both create the key path (eg a.b.c) to the current property being traversed. As far as I can tell the values are always the same. The only way I see they could be different is if the attribute doesn't exist on the schema. In that case the `path.attribute` will not be update whereas `path.raw` will always be updated. Maybe @Convly could clarify the difference?

remidej (approved):

docs/docs/docs/01-core/utils/traverse-entity.md
Convly: >The only way I see they could be different is if the attribute doesn't exist on the schema. In that case the path.attribute will not be updated whereas path.raw will always be updated

That's basically it. It's made so that you're always sure that path.attribute corresponds to a valid schema attribute whereas path.raw is used to get access to a specific value from the base object. They'll almost always be identical but there are some situation (with nested structures like DZ) where you can have small differences.

tbf, this is mainly used in the other traversal (queries, filters, etc...) when dealing with extra operators or when manually modifying the base object to inject new properties.

PR #22596, opened 2025-01-08
enhancement: allow to add custom languages and replace existing langu…

giollord:
### What does it do?
Allows to add custom languages and replace existing language native names in app.ts

### Why is it needed?
It allows to add custom translations for not yet supported languages. At the moment it is not possible to add new language. It also allows to change existing languages' native names displayed in language switch dropdown in admin panel.

### How to test it?
Put the following to app.ts or app.js:
```
const config = {
  locales: ['it', 'es', 'en', 'en-GB', 'lv'],
  languageNativeNames: {
    lv: 'Latviešu',
    es: "Spanish"
  }
};
```
It will add `Latviešu` language option and replace `Español` with `Spanish`. Translations for added languages could be provided in same way as extensions for existing languages by adding file `lv.json` or `config.translations` https://docs.strapi.io/dev-docs/admin-panel-customization/options#extending-translations

### DIFF
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -42,6 +42,7 @@ interface StrapiAppConstructorArgs extends Partial<Pick<StrapiApp, 'appPlugins'>
     auth?: { logo: string };
     head?: { favicon: string };
     locales?: string[];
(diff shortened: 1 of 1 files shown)


oiorain:
Hi there, 

Thank you for opening a PR. We will not be merging it because languages are standardized and we are using the standard versions.

PR #22606, opened 2025-01-09
Release 4.25.19

innerdvations:
### What does it do?

- update package versions for 4.25.19 release
- updates playwright to get github CI to work; doing it here because it has no impact on the code itself, only tooling, and this way we have validation from the CI that tests still pass

### Why is it needed?

release

### How to test it?

n/a


remidej (commented):
the build is failing ⚠️

remidej (approved):

PR #22594, opened 2025-01-08
fix: remove link to open a new issue directly

derrickmehaffy:
### What does it do?

Alters the link to open up a new bug report issue when encountering the error page to just point to issues directly

### Why is it needed?

Too many people opening bug reports and not filling out the template properly
https://github.com/strapi/strapi/issues?q=is%3Aissue+label%3A%22flag%3A+invalid+template%22+is%3Aclosed

### How to test it?

Hit the error and click the button

### Related issue(s)/PR(s)

N/A

PR #22608, opened 2025-01-09
chore: switch to the new github bug report template system

derrickmehaffy:
### What does it do?

Removes the old bug report template and implements the new form system for bug reports

### Why is it needed?

Hopefully to get better bug reports submitted by users that provide enough information

### How to test it?

Sadly we have to merge this to test it

### Related issue(s)/PR(s)

related to feedback on https://github.com/strapi/strapi/pull/22594


.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,117 @@
...
+      description: Provide a detailed explanation of what you expected to happen instead of the observed issue. Be as specific as possible.
+    validations:
+      required: true
+
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logs
+      description: If available, please provide the logs generated either in the Browser or Server Side console.
+      render: shell
```
innerdvations, line 92: Considering there might be more than one log, I wonder if we might want a text area for "additional information" that can be used for whatever the reporter feels like adding

derrickmehaffy: added another block for that, I added the logs one because it has a special renderer I want to take a look at

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,117 @@
...
+  - type: input
+    id: nodeVersion
+    attributes:
+      label: Node Version
+      description: The version of Node.js you are using (only LTS versions are supported).
+      placeholder: 20.0.0
+    validations:
+      required: true
+
+  - type: input
```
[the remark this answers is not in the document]
derrickmehaffy: the report command we have I think combines them so if people want to copy paste this might be easier

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -4,14 +4,20 @@ contact_links:
     url: https://feedback.strapi.io/feature-requests
     about: Provide feedback to the Strapi team and ask for new features or enhancements!
   - name: Documentation Bug Report
-    url: https://github.com/strapi/documentation/issues/new?template=BUG_REPORT.md&title%5B%5D=BUG
+    url: https://github.com/strapi/documentation/issues/new?assignees=&labels=type%3A+bug&projects=&template=BUG_REPORT.yml&title=%5BBug%5D%3A+
```
innerdvations, line 7:
```suggestion
    url: https://github.com/strapi/documentation/issues/new?assignees=&labels=type%3A+bug&projects=&template=BUG_REPORT.yml&title=%5BBug%5D%3A%20
```

innerdvations (approved):
LGTM, if there are any issues we can check and fix after

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,150 @@
...
+      description: Select the operating system where you are experiencing the issue.
+      multiple: false
+      options:
+        - Strapi Cloud
+        - MacOS
+        - Linux (Debian/Ubuntu)
+        - Linux (RedHat/CentOS)
+        - Linux (Other)
+        - Windows 10
+        - Windows 11
```
Convly, line 44: Why are we asking the version for Windows but not for the other OS? Should we just ask "Windows"?

Convly (approved):
Overall LGTM

Maybe one last thing we could ask is if the app is run in a container (like Docker) or not?

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
innerdvations: It kind of makes sense just because trying to provide a list for the others would be unmanageable and windows only has 2 😆

But also, maybe make it a multiselect?  For things like strapi cloud + macOs,  or docker + linux. Or split it into OS + OS Version fields?

For now I think it's fine and we can tweak it later though

innerdvations (approved):

Issue #22266, opened 2024-11-25
Better bulk publish overview


lucasboilly:
#### 1. Missing data
The current bulk publish overview doesn't include a counter of the entries that have a `Ready to publish changes` status despite the fact that we introduced the changes in Strapi 5 through the `Modified` entry status.

#### 2. Readability issue
The current overview is pretty hard to read as it's only text and there's nothing to differentiate each counter. 

![CleanShot 2024-11-25 at 11 16 20@2x](https://github.com/user-attachments/assets/778b0cf7-0694-41ac-ae72-c635fb5a56e9)

### Solution

![CleanShot 2024-11-25 at 11 46 36@2x](https://github.com/user-attachments/assets/b00d4c73-cb82-4957-bdcf-4f4712bc2952)

#### Step 1

Display the number of entries that have a `Ready to publish changes` status.

#### Step 2

Change the displayed texts. It should match the statuses texts in the table except for `Waiting for action`, because we display different info in the table status depending on the error.

#### Step 3

Add the french wording in the [french translations file](https://github.com/strapi/strapi/blob/185e16fdc3e6f1895610ab73932bc3f044bb59e9/packages/core/admin/admin/src/translations/fr.json). You'll find the english wording in the [english translation file](https://github.com/strapi/strapi/blob/185e16fdc3e6f1895610ab73932bc3f044bb59e9/packages/core/admin/admin/src/translations/en.json) (ex: `app.utils.ready-to-publish`).
- Ready to publish > Prêt à publier
- Ready to publish changes > Prêt à publier les changements
- Already published > Déjà publié
- Waiting for action > En attente d'une action

#### Step 4

Add a specific icon for each counter. It should match the statuses icons in the table.

- Ready to publish > `CheckCircle`, the color should be `Success600`
- Ready to publish changes > `ArrowsCounterClockwise`, the color should be `Alternative600`
- Already published > `CheckCircle`, the color should be `Success600`
- Waiting for action > `CrossCircle`, the color should be `Danger600`

#### Step 5

Apply the rest of the UI changes
- Put the counters on the right
- Split the overview into 4 cells with borders (the borders color should be `Neutral150`)
- Add a border radius (4px)
- Use the right spacings

![CleanShot 2024-11-25 at 12 03 32@2x](https://github.com/user-attachments/assets/6fe25bfb-0018-41e7-8efd-0163b2fdb04b)

abhilov23:
calm

abhilov23:
[...]

PR #22618, opened 2025-01-12
feat: better bulk publish overview

Dhruv-Maradiya:
### What does it do?

- Displays the `Ready to publish changes` status in publish overview
- Enhances readability with a table view.

### Why is it needed?

1. Resolves missing data issues.
2. Improves readability.

### How to test it?

1. Open the bulk publish view.
2. Verify that the updated UI uses a table format instead of plain text.

### Related issue(s)/PR(s)

Related to #22266


Dhruv-Maradiya:
@lucasboilly Thank you for your review, I will make mentioned changes. 😊

lucasboilly:
Nice, thanks again!

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -273,6 +284,93 @@ const SelectedEntriesTableContent = ({
...
+}
+
+const PublicationStatusSummary = ({ count, icon, message }: PublicationStatusSummaryProps) => {
+  return (
+    <Flex justifyContent="space-between" flex={1} gap={3}>
+      <Flex gap={2}>
+        {icon}
+        <Typography>{message}</Typography>
+      </Flex>
+      {BoldChunk(count)}
```
markkaylor, line 304: I think the `BoldChunk` was originally for passing to a translation object and I don't think it was used anywhere else. I think we should delete the component and just add the Typography bold here:
```suggestion
      <Typography fontWeight="bold">{chunks}</Typography>
```

tests/e2e/tests/content-manager/listview.spec.ts
```diff
@@ -58,10 +58,43 @@ test.describe('List View', () => {
...
+        const waitingForActionText = await waitingForActionElement.textContent();
+        const readyToPublishChangesText = await readyToPublishChangesElement.textContent();
+
+        expect(publishedText).toBe(publishedCount.toString());
+        expect(readyToPublishText).toBe(readyToPublishCount.toString());
+        expect(waitingForActionText).toBe(waitingForActionCount.toString());
+        expect(readyToPublishChangesText).toBe(readyToPublishChangesCount.toString());
+      };
+
+      await verifyPublicationStatus(0, 2, 0, 0);
```
markkaylor, line 97: Your e2e test should be using real app data, otherwise this should be a unit test.

tests/e2e/tests/content-manager/listview.spec.ts
```diff
@@ -58,10 +58,43 @@ test.describe('List View', () => {
...
+      await page.waitForSelector('text=Ready to publish');
+
+      const verifyPublicationStatus = async (
+        publishedCount: number,
+        readyToPublishCount: number,
+        waitingForActionCount: number,
+        readyToPublishChangesCount: number
+      ) => {
+        const alreadyPublishedElement = await page
+          .locator('text=Already Published')
```
markkaylor, line 70: We should try to use the recommended built in locators: https://playwright.dev/docs/locators#quick-guide. Have you tried getByText?

lucasboilly:
Hey @Dhruv-Maradiya :)

I just noticed a small bug, in some cases I get a value of -1. Do you know how to fix this?

CleanShot 2025-01-17 at 14 12 15@2x

Dhruv-Maradiya:
I think the issue is with how these values are calculated. I will fix it, but can you clear my doubt?  

Should the total of all status fields match the number of selected entries, or can they overlap?

Dhruv-Maradiya:
> Hey @Dhruv-Maradiya :)
> 
> I just noticed a small bug, in some cases I get a value of -1. Do you know how to fix this?
> 
> CleanShot 2025-01-17 at 14 12 15@2x

In this image value of `Ready to publish changes` and `Waiting for action` is overlapped. Is it the intended behavior or should I change it?

lucasboilly:
@Dhruv-Maradiya

The counters should only be based on the publication status, not the status.

The publication statuses are:
- Ready to publish
- Already published
- Ready to publish changes
- Waiting for action

The statuses are:
- Draft
- Published
- Modified

For example, in the screenshot I shared we should have:
- Ready to publish: 0
- Already published: 2
- Ready to publish changes: 1
- Waiting for action: 1


Regarding your question, the values shouldn't be overlapped. If an entry has an error, it automatically gets the Waiting for action status and can't be published.
> In this image value of Ready to publish changes and Waiting for action is overlapped. Is it the intended behavior or should I change it?

I'm available anytime if you have any other question :)

Dhruv-Maradiya:
Thank you, @lucasboilly, for your quick response and clarification.

tests/e2e/tests/content-manager/listview.spec.ts
Dhruv-Maradiya: I'm not entirely sure what I need to do here. I think it's using app data. Previously, it matched a string containing all count of publication statuses, but with tabular view, we now need to match each publication status separately.

Let me know if I’m missing something or if there’s a specific approach I should follow.

markkaylor: No sorry that's my mistake 🤦‍♂️ , but do we need that function?

tests/e2e/tests/content-manager/listview.spec.ts
```diff
@@ -67,19 +67,19 @@ test.describe('List View', () => {
         readyToPublishChangesCount: number
       ) => {
         const alreadyPublishedElement = await page
-          .locator('text=Already Published')
+          .getByText('Already Published')
           .first()
           .locator('xpath=../following-sibling::*');
```
markkaylor, line 72: Sorry, I don't really understand what this is. Could you explain and why it is necessary? Shouldn't `.first()` be enough?

tests/e2e/tests/content-manager/listview.spec.ts
Dhruv-Maradiya: We can remove the function, but that'll mean repeating these lines twice, so I thought keeping it would be better. Also, we can take out the locating part since we don't need to do it twice.

```typescript
let publishedText = await alreadyPublishedElement.textContent();
let readyToPublishText = await readyToPublishElement.textContent();
let waitingForActionText = await waitingForActionElement.textContent();
let readyToPublishChangesText = await readyToPublishChangesElement.textContent();

expect(publishedText).toBe('0');
expect(readyToPublishText).toBe('2');
expect(waitingForActionText).toBe('0');
expect(readyToPublishChangesText).toBe('0');
```

Let me know your thoughts!

tests/e2e/tests/content-manager/listview.spec.ts
Dhruv-Maradiya: We're actually targeting the publication status count, not the text. In the DOM, the text's parent is a sibling to the count, so after finding the text, we need to go to its parent and then to its next sibling. I tried looking for a better way to do this but couldn’t find one.

markkaylor: Ok I think I understand it better, but I think this is a sign that we are dealing with elements that are not accessible. It doesn't look like we can navigate to the status info summaries with the keyboard and the screen reader isn't able to read it. Maybe we should be using a table without in PublicationStatusSummary? wdyt you have any other idea how we could make this accessible?

markkaylor (commented):
I found this behavior a bit strange and I don't think it is expected. Ready to publish is tied to the selected entries. For example if I deselect an entry that is "Ready to publish" the Ready to publish count decreases. 

https://github.com/user-attachments/assets/d1cbc9ae-49f1-483d-aca4-ce7e59aae455

Dhruv-Maradiya:
> I found this behavior a bit strange and I don't think it is expected. Ready to publish is tied to the selected entries. For example if I deselect an entry that is "Ready to publish" the Ready to publish count decreases.

I think it makes sense this way because only the selected entries will be published. If the count didn't update based on selection, there would be a mismatch between the entries that were actually published and what we saw in the publication status summary. Let me know how you'd like me to proceed.

tests/e2e/tests/content-manager/listview.spec.ts
Dhruv-Maradiya: You're right, we can't navigate the status info summaries using the keyboard. I originally tried using a table but faced some issues, so I switched to a grid instead. I'll try using a table again. Since I don't have much experience with Playwright, I'm not completely sure about the best approach.

markkaylor: I think the table will probably be the most straightforward and give us the accessibility we need. On the playwright side that is going to make it much easier to get the element we want and ensure the app remains accessible with something that might look like `page.getByRole('gridcell', {name: 'Ready to publish 1'})`

Dhruv-Maradiya: That makes sense. I'll update it to use a table instead.

markkaylor:
> > I found this behavior a bit strange and I don't think it is expected. Ready to publish is tied to the selected entries. For example if I deselect an entry that is "Ready to publish" the Ready to publish count decreases.
> 
> I think it makes sense this way because only the selected entries will be published. If the count didn't update based on selection, there would be a mismatch between the entries that were actually published and what we saw in the publication status summary. Let me know how you'd like me to proceed.

Yup you're right! And that's how it was before.

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -47,6 +51,17 @@ const TypographyMaxWidth = styled<TypographyComponent>(Typography)`
...
+const TableComponent = styled(RawTable)`
+  width: 100%;
+  table-layout: fixed;
+  td:first-child {
+    border-right: 1px solid ${({ theme }) => theme.colors.neutral150};
+  }
+`;
+const TableCell = styled(Td)`
+  padding: 16px !important;
+`;
```
markkaylor, line 63: I prefer we avoid the important by overriding the `td:first-of-type` rule that is set on each `tr`. I believe that should give the same result. You won't need the `TableCell` styled component anymore and you can use `Td` directly

```suggestion
  td:first-child {
    border-right: 1px solid ${({ theme }) => theme.colors.neutral150};
  }
  td:first-of-type {
    padding: ${({ theme }) => theme.spaces[4]};
  }
`;
```

markkaylor (commented):
Really Nice work @Dhruv-Maradiya ! I just have a couple small comments 🙏 then I think we will be good to go!

tests/e2e/tests/content-manager/listview.spec.ts
```diff
@@ -60,41 +60,26 @@ test.describe('List View', () => {
...
-        expect(waitingForActionText).toBe(waitingForActionCount.toString());
-        expect(readyToPublishChangesText).toBe(readyToPublishChangesCount.toString());
-      };
-
-      await verifyPublicationStatus(0, 2, 0, 0);
+      expect(
+        await page.getByRole('gridcell', {
+          name: 'Already published 0',
+        })
+      ).toBeVisible();
```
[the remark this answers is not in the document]
Dhruv-Maradiya: Hmm, makes sense! Thanks for pointing it out.

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
Dhruv-Maradiya: Ohh right, I see what went wrong earlier! I had mistakenly applied padding to `first-child` instead of `first-of-type`, which is why it didn’t work as expected so I used `!important` for padding. Thanks for the suggestion!

markkaylor (approved):
@Dhruv-Maradiya, this is a really nice improvement. Thanks for your contribution hope to see more PRs from you soon :) 👏👏👏👏👏👏👏 

@lucasboilly I've tested it locally but could you confirm all is good on your side before I merge?

lucasboilly:
@markkaylor I checked one last time and that's still all good 👌

Thanks again @Dhruv-Maradiya!

MarionLemaire:
Just 3 small remarks here after testing : 
- Already Published : the P is still capitalized
- "Ready to publish changes" lets no margin  on the table
- For waiting for action items, there's a dot at the end of the missing field (here "cover is required.")
Overall it's really a great improvement, nice job !!
Screenshot 2025-01-28 at 11 32 34

PR #22619, opened 2025-01-12
fix(core/utils): return correct type when providing defaultValue

iamandrewluca:
### What does it do?

Previously, when any of the `env` helpers were used with a provided `defaultValue` argument, it still included `undefined` in the return type. Now, when a `defaultValue` is provided, `undefined` is excluded from the return type.

### Why is it needed?

Whenever the `env` helper is used in config files, even if the return value is correct, you need to typecast it to work in TS.

```ts
// config/database.ts
import type { ConfigParams } from "../src/utils/types";

export default ({ env }: ConfigParams) => {
	const connections = { postgres: {}, sqlite: {} };

	const client = env("DATABASE_CLIENT", "sqlite");
	//    ^ string | undefined

	return {};
};
```

### How to test it?

Use any `env` helper with a provided default value. The return type will include `undefined`.

> [!NOTE]
> The function override where `defaultValue` is explicitly `undefined`, I kept it to be compatible with previous behavior, where the user could provide `undefined` as `defaultValue`.


packages/core/utils/src/env-helper.ts
```diff
@@ -10,90 +13,130 @@ function getKey(key: string) {
...
-      throw new Error(`env.oneOf requires defaultValue to be included in expectedValues`);
-    }
+/**
+ * Gets a value from env that matches oneOf provided values
+ */
+function envOneOf<T extends string>(key: string, expectedValues: ReadonlyArray<T>): T | undefined;
+function envOneOf<T extends string>(
+  key: string,
+  expectedValues: ReadonlyArray<T>,
+  defaultValue: NoInfer<T>
```
iamandrewluca, line 108: Note: `NoInfer` is a `typescript@5.4` feature.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type

packages/core/utils/src/env-helper.ts
```diff
@@ -10,90 +13,130 @@ function getKey(key: string) {
...
-    }
-  },
+  return getKey(key) === 'true';
+}
 
-  array(key: string, defaultValue?: string[]): string[] | undefined {
-    if (!_.has(process.env, key)) {
-      return defaultValue;
+function envJson(key: string): object | undefined;
+function envJson(key: string, defaultValue: object): object;
```
iamandrewluca, line 50: Note: To allow any primitive value, `{}` can be used instead of `object`

unrevised6419:
Closing because I think the changes are too exaggerated.
I'll consider creating another PR with less changes, and same effect.

PR #21859, opened 2024-10-15
feat(strapi/create): type strapi configs

unrevised6419:
### What does it do?

- Exports the Core.Config.Shared
- Types the Strapi Configs

### Why is it needed?

Want to have the ability to type configurations out of the box.

PR #22620, opened 2025-01-12
feat(core/config): export Shared types

unrevised6419:
### What does it do?

Export the `Shared` types namespace for Core.Config

### Why is it needed?

With this exported, Strapi users are able to type their config files.

```ts
import type { Core } from "@strapi/strapi";

type AdminConfig = Core.Config.Shared.ConfigFunction<Core.Config.Admin>;

const config: AdminConfig = ({ env }) => {
	return {
		apiToken: { salt: env("API_TOKEN_SALT") },
		transfer: { token: { salt: env("TRANSFER_TOKEN_SALT") } },
		auth: { secret: env("ADMIN_JWT_SECRET") },
	};
};

export default config;
```

> [!NOTE]
> This `shared.ts` file seems to have been created for this but was never exported and is not used inside Strapi code also.

### DIFF
--- a/packages/core/types/src/core/config/index.ts
+++ b/packages/core/types/src/core/config/index.ts
@@ -1,3 +1,4 @@
+export type * as Shared from './shared';
 export type { Server } from './server';
 export type { Admin } from './admin';
 export type { Api } from './api';


innerdvations:
@Convly @maccomaccomaccomacco It was always the intention to export this and it just got overlooked with v5, right?

packages/core/types/src/core/config/index.ts
```diff
@@ -1,3 +1,4 @@
+export type * as Shared from './shared'
```
innerdvations:
```suggestion
export type * as Shared from './shared';
```

Convly (approved):

unrevised6419:
Sorry, did a rebase and it dismissed Convly's Review

innerdvations:
Thanks for the contribution! Once all the CI tests pass I'll merge

unrevised6419:
After a new release of Strapi, I will reopen this PR, which adds the config types for `create-strapi-app`

https://github.com/strapi/strapi/pull/21859

innerdvations:
@pwizla It looks like we don't have any documentation on typing configuration files, but now would be a good time to add it. Would you be able to add the example from the PR description, maybe at the bottom of https://docs.strapi.io/cms/configurations ?

unrevised6419:
I actually reopened the PR and rebased. It depends on changes from this PR

https://github.com/strapi/strapi/pull/21859

> Would you be able to add the example from the PR description, maybe at the bottom of [docs.strapi.io/cms/configurations](https://docs.strapi.io/cms/configurations) ?

This is a question to me or @pwizla? 👀

innerdvations:
> I actually reopened the PR and rebased. It depends on changes from this PR
> 
> #21859
> 
> > Would you be able to add the example from the PR description, maybe at the bottom of [docs.strapi.io/cms/configurations](https://docs.strapi.io/cms/configurations) ?
> 
> This is a question to me or @pwizla? 👀

That was for @pwizla (unless you want to try your hand at writing documentation) 😆

unrevised6419:
I could try that. But that would be more for users who use TypeScript and don't have yet the configs typed.
The second PR I opened does it make sense to have it in this case?

IMO the default created Strapi App with TypeScript should already include the types for the configs.

innerdvations:
> I could try that. But that would be more for users who use TypeScript and don't have yet the configs typed. The second PR I opened does it make sense to have it in this case?
> 
> IMO the default created Strapi App with TypeScript should already include the types for the configs.

Yeah, I agree it should be, it got left behind for higher priority things during v5, so I'm very happy to see it happening.

I'll try to get your other PR reviewed next week, thanks again!

pwizla:
> @pwizla It looks like we don't have any documentation on typing configuration files, but now would be a good time to add it. Would you be able to add the example from the PR description, maybe at the bottom of https://docs.strapi.io/cms/configurations ?

Sure, I can do it next week!

PR #22622, opened 2025-01-12
fix(plugins/users): remove redundant grant-koa dependency in favor of grant.koa

iamandrewluca:
### What does it do?

Remove `grant-koa` dependency that only re-exports grant koa.

The only thing `grant-koa` does is re-export `require('grant').koa()`

https://unpkg.com/browse/grant-koa@5.4.8/index.js

`grant-koa` was not published for 4 years, and it seems is not needed anymore.

### Why is it needed?

Faster download times. Smaller bundles/builds.


hanpaine:
Hi @iamandrewluca thanks for your fixes/improvement suggestions, it's much appreciated 🚀 we'll take a look soon 🙏

jhoward1994 (approved):

jhoward1994:
Approved but just flagging to not merge while the CI runs

PR #22634, opened 2025-01-13
fix(i18n): allow unique + un localized fields on i18n content types

jhoward1994:
### What does it do?

This PR fixes an issue in i18n where it was not possible to define unique and non-localized fields for i18n content types. It modifies the handling of uniqueness constraints to allow these fields to function correctly in i18n models.

### How to test it?

1. Create an i18n content type with at least one unique and non-localized field.
2. Add multiple records to the content type in different locales.
3. Verify that the unique constraints are respected across locales for the non-localized field.
4. Confirm that the application does not throw errors when managing such content types.

### Related issue(s)/PR(s)

DX-1834

### DIFF
--- a/packages/plugins/i18n/admin/src/index.ts
+++ b/packages/plugins/i18n/admin/src/index.ts
@@ -139,28 +139,6 @@ export default {
       });
 
       ctbFormsAPI.extendFields(LOCALIZED_FIELDS, {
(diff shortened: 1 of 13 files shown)


Bassel17 (approved):
LGTM!
and working

PR #22635, opened 2025-01-13
test(e2e): text fields advanced settings

innerdvations:
### What does it do?

Adds testing for text field "required" and "regexp" advanced settings, from creating it in the CTB to entering data for them in basic fields, components, and dzs.

More specifically:

- adds util for to ensure a checkbox value
- retain logs on failure locally (helps with debugging flaky tests)
- findAndClose allows to skip in case of expected error
- add test that error messages show up when a required text field in a dynamic zone is not entered

TODO:

- [ ] add test that basic required field shows error
- [ ] add test that required field within component (NOT in a dz) shows error
- [ ] check that aria-invalid is set
- [ ] add regexp and test it

### Why is it needed?

gap in testing

### How to test it?

run them

### Related issue(s)/PR(s)

DX-1843


butcherZ (approved):
tested locally, LGTM!

innerdvations:
Ok, let's merge this and then follow up with the remaining TODOs

PR #22637, opened 2025-01-13
Small architectural change proposal

AbdelrahmanElmughrabi:
### What does it do?
It improves storing rateLimit values, and it makes the rateLimit values shared between instances/applications ensuring consistent rate limiting behavior.

I integrated Redis with "rate-limiter-flexible" to replace the in-memory rate limiting.

### Why is it needed?
Improves scalability, it addresses the issue of different counters on different servers, and provides a robust solution for distributed rate limiting.

### How to test it?
I'm not sure

### Related issue(s)/PR(s)
No


alexandrebodin (changes requested):
Sorry but we cannot accept such a PR. we do not want to enforce using Redis, this would be breaking change

PR #22647, opened 2025-01-15
ci: passing github token to publishing script

Bassel17:
### What does it do?

- setup the github actor to do the commits 
- ensures the github token is passed to the script

### Why is it needed?

- make the workflow work

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -30,3 +30,5 @@ jobs:
         env:
           VERSION: ${{ github.event.inputs.version }}
           DIST_TAG: 'latest'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_ACTOR: ${{ github.actor }}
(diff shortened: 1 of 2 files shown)


remidej (commented):
I'm happy to try this, just curious if you know if we have ["permissive" or "restricted" access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)? if it's restricted I think this will fail

Bassel17:
> I'm happy to try this, just curious if you know if we have ["permissive" or "restricted" access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)? if it's restricted I think this will fail

looking at other actions permissions, I think we do have the permissions. (to commit and release)
@alexandrebodin if you can confirm

alexandrebodin:
> > I'm happy to try this, just curious if you know if we have ["permissive" or "restricted" access](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)? if it's restricted I think this will fail
> 
> looking at other actions permissions, I think we do have the permissions. (to commit and release) @alexandrebodin if you can confirm

yes this should work

PR #22650, opened 2025-01-15
fix: disable lifecycle hooks on transfers [v4]

### LINKED ISSUE #19383: DTS Transfer/export/import Does Not Disable Lifecycle Hook

- Node.js version:   18.18.0
- NPM version:        9.8.1
- Strapi version:      4.19.0
- Database:             Sqlite
- Operating system:  Mac OS
- Is your project Javascript or Typescript: No

### Describe the bug
DTS Transfer Does Not Disable Lifecycle Hook when importing data leading to errors depending on how those lifecycles are written

1. Create a content-type
2. Create a lifecycle that does something unique such as setting a relationship that may not exist yet during transfer
3. See error

### Expected behavior

Much like we do with HTTP timeouts we should probably disable lifecycles during the transfer then re-enable them once transfer is complete

### Screenshots

N/A

### Code snippets

Will add if we find some good examples

### Additional context

Reported via community office hours in Discord

Bassel17:
### What does it do?
- disable lifecycle hooks during a transfer
enable them back again after a transfer

### Why is it needed?
- prevent lifecycle hooks from being triggered during transfer operations

### How to test it?
- during a transfer or import or export lifecycle hooks should not be triggered

### Related issue(s)/PR(s)

fixes #19383

### DIFF
--- a/packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
@@ -68,7 +68,7 @@ class LocalStrapiDestinationProvider implements IDestinationProvider {
     if (!this.strapi) {
       throw new ProviderInitializationError('Could not access local strapi');
     }
-
+    this.strapi.db.lifecycles.disable();
     this.transaction = utils.transaction.createTransaction(this.strapi);
   }
 
@@ -102,7 +102,8 @@ class LocalStrapiDestinationProvider implements IDestinationProvider {
   async close(): Promise<void> {
     const { autoDestroy } = this.options;
     this.transaction?.end();
-
+    assertValidStrapi(this.strapi);
+    this.strapi.db.lifecycles.enable();
     // Basically `!== false` but more deterministic
     if (autoDestroy === undefined || autoDestroy === true) {
       await this.strapi?.destroy();
--- a/packages/core/data-transfer/src/strapi/providers/local-destination/__tests__/assets.test.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-destination/__tests__/assets.test.ts
@@ -30,7 +30,13 @@ const strapiFactory = getStrapiFactory({
       public: 'static/public/assets',
     },
   },
-  db: { transaction },
+  db: {
+    transaction,
+    lifecycles: {
+      enable: jest.fn(),
+      disable: jest.fn(),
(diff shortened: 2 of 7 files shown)


packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
```diff
@@ -102,7 +102,8 @@ class LocalStrapiDestinationProvider implements IDestinationProvider {
   async close(): Promise<void> {
     const { autoDestroy } = this.options;
     this.transaction?.end();
-
+    assertValidStrapi(this.strapi);
+    this.strapi.db.lifecycles.enable();
```
innerdvations, line 106: I haven't checked the code, but does this actually change something that could persist if this fails to run? Or is it just a nice thing to re-enable before shutting down the instance?

innerdvations (commented):
Could end up in the situation where other users are relying on them running. What do you think about having a config setting that is checked?

packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
Bassel17: it does have to run, or else strapi has to restart to have lifecycle hooks enabled

Bassel17: or maybe not actually

jhoward1994:
Not very familiar with this part of the code but I'm wondering if you need to make the same changes for remote destinations and sources?

e.g.
https://github.com/strapi/strapi/blob/c44172f3a1ca2b435f87478ad430530079c84443/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts#L188-L189

innerdvations:
> I'm not very familiar with this part of the code but I'm wondering if you need to make the same changes for remote destinations and sources?
> 
> e.g.
> 
> https://github.com/strapi/strapi/blob/c44172f3a1ca2b435f87478ad430530079c84443/packages/core/data-transfer/src/strapi/providers/remote-destination/index.ts#L188-L189

I think the remote providers load up the local provider internally, so it should be fine. Is that right @Bassel17 ?

Bassel17:
> LGTM, works well!

If you can take a look at the v5 one too, it's exactly the same @innerdvations https://github.com/strapi/strapi/pull/22651

PR #22654, opened 2025-01-16
fix: webpack public path was missing a slash at the end

axe312ger:
### What does it do?

It fixes the public path in the generated webpack config to include a slash as suffix

### Why is it needed?

The browser can not load the react backend when building with webpack.

Webpack was generating chunk paths like this `http://localhost:1337/admin1400.641ba92c.chunk.js` which results in a blank page as these files don't exist with that name. 

### How to test it?

I guess it should happen for anyone on current versions (I tried 5.6 and 5.8).

```
npx strapi build --bundler=webpack --debug
npx strapi start
```

### Workaround

We set this in our custom webpack config to keep the project running for now:

```
config.output.publicPath = "/admin/";
```

### DIFF
--- a/packages/core/strapi/src/node/webpack/config.ts
+++ b/packages/core/strapi/src/node/webpack/config.ts
@@ -167,7 +167,7 @@ const resolveDevelopmentConfig = async (ctx: BuildContext): Promise<Configuratio
     output: {
       filename: '[name].js',
       path: ctx.distPath,
-      publicPath: ctx.basePath,
+      publicPath: `${ctx.basePath}/`,
     },
     infrastructureLogging: {
       level: 'error',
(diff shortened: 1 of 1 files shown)


hanpaine:
Thanks for the contribution @axe312ger ! 🚀  we'll take a look soon 🙏

innerdvations:
It appears this is still a valid issue; if you need this fixed, could you please update it so that it can safely handle anyone working around it with an extra trailing slash like in your workaround? That is, normalize the path before adding a slash so that basePath values like "/admin" and "/admin/" both become "/admin/" and never "/admin//". For example:

`publicPath: ctx.basePath.replace(/\/+$/, '') + '/',`

That way the fix works for the default case and doesn’t break anyone who already uses a trailing slash in their admin URL. Thanks!

Adzouz:
I've made the change @innerdvations and also done it for resolve production config. Let me know if it looks good to you so we can merge it 🙏

innerdvations (approved):

PR #22655, opened 2025-01-16
feat: updating Plan label now that we have a Growth plan

cache-your-dreams:
### What does it do?

Simply updates the label in Settings -> Overview to Current Edition

### Why is it needed?

When buying a Growth Plan, Enterprise Edition is currently shown under a label saying Current Plan.

### How to test it?

Start Strapi and navigate to Settings -> Overview. The label should say Current Edition


alexandrebodin (changes requested):
Did ou check for his translation ke in other languages ?

HichamELBSI (approved):

HichamELBSI (changes requested):
Indeed, you'll have to update the other languages as well.

alexandrebodin (approved):

PR #22656, opened 2025-01-16
fix: adding component to dynamic zone overwrites data

### LINKED ISSUE #22399: Dynamic zone not inserting items correctly when using "Add component below/above" button

- Node.js version: 22.12.0
- NPM version: 10.9.0
- Strapi version: 5.5.1
- Database: MySql
- Operating system: MacOs

### Describe the bug

When adding new items to a dynamic zone list, clicking 'Add Component Below' replaces the item below the selected element instead of inserting a new item underneath it as expected. Similarly, clicking 'Add Component Above' replaces the current item with the new one, rather than adding it above the selected element.

1. Create components to add to the dynamicZone list
2. Create a collection type with a dynamicZone attribute
3. Add those components in the dynamicZone.
4. Create a new item on the new collection.
5. Upon creation, add components to the dynamicZone
6. Try to add components above or below on that dynamicZone list using the buttons provided by the admin panel (3 dots)
7. (To reproduce the issue, the component you are referencing to insert a new item above or below should already have items below and above)

### Expected behavior

- New components added via "Add component above" should be inserted **above** the referenced component where the action was triggered.
- New components added via "Add component below" should be inserted **below** the referenced component where the action was triggered.

### Screenshots

https://github.com/user-attachments/assets/ff49eec5-54c6-4b88-89b7-69ba00cb541e

### Code snippets

remidej:
### What does it do?

- Fixes #22399. I do the fix by not relying on setIn anymore, because it's designed to overwrite content at the index it's given. I use toSpliced instead (the FP version of splice) which just inserts at the the given position
- Adds new e2e test case to catch the issue
- Fixes the alignment of the "more" action button on components

### Why is it needed?

To avoid accidental data loss

### How to test it?

Create content in dynamic zones, add components to the top, to the bottom, in the middle, maybe delete and recreate some.

Also do sanity check changing review workflows stages, because it seems to use the same addFieldRow util.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
@@ -149,7 +149,7 @@ const DynamicComponent = ({
         <Drag />
       </IconButton>
       <Menu.Root>
-        <Menu.Trigger size="S" endIcon={null} paddingTop="4px" paddingLeft="7px" paddingRight="7px">
+        <Menu.Trigger size="S" endIcon={null} paddingTop={1} paddingLeft="7px" paddingRight="7px">
           <More aria-hidden focusable={false} />
           <VisuallyHidden tag="span">
             {formatMessage({
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
@@ -7,11 +7,11 @@ import {
   Flex,
   Grid,
   IconButton,
-  VisuallyHidden,
   useComposedRefs,
   Menu,
   MenuItem,
   BoxComponent,
+  VisuallyHidden,
 } from '@strapi/design-system';
 import { Drag, More, Trash } from '@strapi/icons';
 import { getEmptyImage } from 'react-dnd-html5-backend';
@@ -156,6 +156,7 @@ const DynamicComponent = ({
               id: getTranslation('components.DynamicZone.more-actions'),
               defaultMessage: 'More actions',
             })}
+            tag="span"
           >
             <More aria-hidden focusable={false} />
           </IconButton>
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
@@ -149,14 +149,16 @@ const DynamicComponent = ({
         <Drag />
       </IconButton>
       <Menu.Root>
-        <Menu.Trigger size="S" endIcon={null} paddingTop={1} paddingLeft="7px" paddingRight="7px">
-          <More aria-hidden focusable={false} />
-          <VisuallyHidden tag="span">
-            {formatMessage({
+        <Menu.Trigger size="S" endIcon={null} paddingLeft={0} paddingRight={0}>
+          <IconButton
+            variant="ghost"
+            label={formatMessage({
               id: getTranslation('components.DynamicZone.more-actions'),
               defaultMessage: 'More actions',
             })}
-          </VisuallyHidden>
+          >
+            <More aria-hidden focusable={false} />
+          </IconButton>
         </Menu.Trigger>
         <Menu.Content>
           <Menu.SubRoot>
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
@@ -11,7 +11,6 @@ import {
   Menu,
   MenuItem,
   BoxComponent,
-  VisuallyHidden,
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
```diff
@@ -149,7 +149,7 @@ const DynamicComponent = ({
         <Drag />
       </IconButton>
       <Menu.Root>
-        <Menu.Trigger size="S" endIcon={null} paddingLeft={2} paddingRight={2}>
+        <Menu.Trigger size="S" endIcon={null} paddingTop="4px" paddingLeft="7px" paddingRight="7px">
```
markkaylor, line 152: Could you explain why we are no longer using theme values?

markkaylor (commented):
The code looks good to me but I won't have time to do the QA today

remidej:
@markkaylor looks like I have to investigate the failing tests before it's ready for QA

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
remidej: 4px can just be 1, I updated that 👍

for the 7px, it's the padding given to the siblings via IconButton, not sure why (@lucasboilly do you know?). So I'm setting 7px myself so the sizes and gaps match. It's also done [there](https://github.com/strapi/strapi/blob/52ef733c58d68d1fbeb03775edcde7dc1ebd490b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx#L282-L284)

screenshot 2025-01-20 at 10 24 06@2x

lucasboilly: @remidej the paddings should be inherited from `size`, I don't know why we need to add styles on top of it. Maybe it comes from the `Accordion` component in the DS? And what happens if you only keep the `size` property?

remidej: screenshot 2025-01-20 at 10 49 53@2x
it gives it a huge padding

lucasboilly: @remidej ok, only for the last one from what I see? 

It seems specific styles were added to it somewhere.

remidej: The first 2 are IconButtons. The last one is a Menu.Trigger

lucasboilly: Can't we nest an `IconButton` in the `Menu.Trigger`? That's what we do in the nav, a menu is displayed when clicking on the `Avatar` and the `Menu.Trigger` has `padding: 0`.

If it's not possible let's just use your workaround. What's important here is to fix the DZ bug.

remidej: That works! I just had to give tag="span" to the icon button, otherwise the nested buttons messed with the accessibility

tests/e2e/tests/content-manager/editview.spec.ts
```diff
@@ -678,5 +678,47 @@ test.describe('Edit View', () => {
...
+        .filter({ has: page.getByRole('heading') })
+        .all();
+      expect(components).toHaveLength(2);
+      expect(components[0]).toHaveText(/product carousel/i);
+      expect(components[1]).toHaveText(/content and image/i);
+
+      // Add components add specific locations:
+      // - very last position
+      await components[1].getByRole('button', { name: /more actions/i }).click();
+      await page.getByRole('menuitem', { name: /add component below/i }).dispatchEvent('click');
```
remidej, line 700: I hate that I have to do those dispatchEvent, but a normal click just would not pass in the CI, even though it always passed locally. See [this section](https://playwright.dev/docs/input#forcing-the-click) of the Playwright docs (force: true did not work either in the CI 🤷)

Marc-Roig (approved):

markkaylor:
Hey just tested this but I have a couple issues:
- Components are always added below
- If there is only one component already and I try to add above I get an error


https://github.com/user-attachments/assets/9040ca07-7827-4285-8e2d-e0f1b02c39fb

tests/e2e/tests/content-manager/editview.spec.ts
```diff
@@ -678,5 +678,47 @@ test.describe('Edit View', () => {
...
+      // There should be a dynamic zone with two components
+      const components = await page
+        .getByRole('listitem')
+        .filter({ has: page.getByRole('heading') })
+        .all();
+      expect(components).toHaveLength(2);
+      expect(components[0]).toHaveText(/product carousel/i);
+      expect(components[1]).toHaveText(/content and image/i);
+
+      // Add components add specific locations:
```
markkaylor, line 697:
```suggestion
      // Add components at specific locations:
```

remidej:
@markkaylor up until the crash isn't the order what you're expecting?

markkaylor:
> @markkaylor up until the crash isn't the order what you're expecting?

oh yeah you're right it looks like it is! My bad

markkaylor (approved):
Nice quick fix!

Issue #14627, opened 2022-10-14
Email Confirmation: Invalid token on the second request


zockerwurf:
### Required System information

- Node.js version: >=12.x.x <=16.x.x
- NPM version: >=6.0.0
- Strapi version: 4.3.6
- Database: sqlite
- Operating system: Windows

### Describe the bug

When a user creates a new account, he gets a email with a confirmation-link (with a token inside). If the user clicks this link twice, he gets an error the second time, looking on an json-response (no styled interface). The message is:
`{"data":null,"error":{"status":400,"name":"ValidationError","message":"Invalid token","details":{}}}`

### Steps to reproduce the behavior

1. Register a new user using the users-permissions-plugin
2. Click on the email-confirmation-link in the received email.
3. Click another time on the same link.

### Expected behavior

The first time the user is redirected:
`ctx.redirect(settings.email_confirmation_redirection || '/');`

I wish the user could be redirected too (p.e. to the login-form) if he clicks the confirmation-link the second time.

### Code snippets

/config/sync/core-store.plugin_users-permissions_email.json:
`
    "email_confirmation": {
      "display": "Email.template.email_confirmation",
      "icon": "check-square",
      "options": {
        "from": {
          "name": "XXX",
          "email": "XXX"
        },
        "response_email": "XXX",
        "object": "Account confirmation",
        "message": "<p>Thank you for registering!</p>\n\n<p>You have to confirm your email address. Please click on the link below.</p>\n\n<p><%= URL %>?confirmation=<%= CODE %></p>\n\n<p>Thanks.</p>"
      }
`

IJsLauw:
I would like to redirect the user as well, to a specified url would be great, thanks!

derrickmehaffy:
Hello,

I see you are wanting to ask a question that is not really a bug report, 

- questions should be directed to [our forum](https://forum.strapi.io) or our [Discord](https://discord.strapi.io). 
- feature requests should be directed to our [feedback and feature request database](https://feedback.strapi.io)

Please see the following contributing guidelines for asking a question [here](https://github.com/strapi/strapi/blob/master/CONTRIBUTING.md#reporting-an-issue).

Lauwisme:
Leaving this here for future reference:

https://github.com/strapi/strapi/blob/97e6bb4346343dc2ba76e12eb124d43d8ba7b17f/packages/plugins/users-permissions/server/controllers/auth.js#L368

[...]

PR #22674, opened 2025-01-19
feat: add email confirmation error redirection settings

jb-thery:
**What does it do?**

Capture d’écran 2025-01-19 à 19 10 46

---

This PR adds an optional error redirection URL in the admin panel for email confirmation. If a user clicks their confirmation link more than once, they are redirected to a custom page instead of seeing a raw 403 error. If no error URL is set, the default 403 message still appears, preserving the existing behavior.

```js
if (!user) {
  const redirectUrl = settings.email_confirmation_error_redirection || null;

  if (redirectUrl) {
    return ctx.redirect(redirectUrl);
  }

  throw new ValidationError('Invalid token');
}
```
**Why is it needed?**

Several developers have reported that end users often click the confirmation link multiple times and run into the unformatted 403 error. In all cases, offering a custom error page leads to a better user experience for end users by providing a more friendly and informative message in case of errors.

see:

- [https://feedback.strapi.io/feature-requests/p/redirect-a-user-after-confirmation-link-failed](https://feedback.strapi.io/feature-requests/p/redirect-a-user-after-confirmation-link-failed)

- [https://forum.strapi.io/search?q=email%20confirmation%20link](https://forum.strapi.io/search?q=email%20confirmation%20link)

Capture d’écran 2025-01-18 à 23 11 44

Allowing a custom redirect URL provides a smoother experience and avoids confusion.

Some email providers scan or visit links for security reasons before the user clicks them, which can automatically validate the account and cause an error when the real user tries to confirm. This solution gives developers the option to display a custom page if that happens, preserving a smoother experience for end users.

**How to test it?**

	1.	Enable email confirmation in your Strapi settings.
	2.	Set a custom error redirection URL in the admin panel.
	3.	Sign up a new test user and confirm their email.
	4.	Click the confirmation link to validate account and click to the link a second time.
	5.	Verify that you are redirected to your custom page rather than seeing the 403 error.
	6.	Remove the custom URL to confirm the default 403 behavior remains intact.

**Additional Note**
- I’m not entirely sure how to implement this directly, but ideally, the confirmation link would include the user’s email as a parameter. That way, in the async emailConfirmation(ctx, next, returnUser) function, we could use the userService to check if the user is already validated and return a success response.

- I have added the default text to the translations, but I did not edit all the translation files since I only speak English and French.

Related issue(s)/PR(s)

User and dev documentation : https://github.com/strapi/documentation/pull/2364

Related:
https://github.com/strapi/strapi/issues/14627

### DIFF
--- a/packages/plugins/users-permissions/admin/src/pages/AdvancedSettings/utils/layout.js
+++ b/packages/plugins/users-permissions/admin/src/pages/AdvancedSettings/utils/layout.js
@@ -1,98 +1,62 @@
 import { getTrad } from '../../../utils';
 
+const createField = ({ id, defaultMessage, name, type = 'string', size = 12, placeholder }) => ({
+  label: {
+    id: getTrad(`EditForm.inputToggle.label.${id}`),
+    defaultMessage,
+  },
+  hint: {
+    id: getTrad(`EditForm.inputToggle.description.${id}`),
+    defaultMessage: `Description for ${defaultMessage}`,
+  },
+  name,
+  type,
+  size,
+  ...(placeholder && {
+    placeholder: {
+      id: getTrad(`EditForm.inputToggle.placeholder.${id}`),
+      defaultMessage: placeholder,
+    },
+  }),
+});
+
 const layout = [
-  {
-    label: {
-      id: getTrad('EditForm.inputToggle.label.email'),
-      defaultMessage: 'One account per email address',
-    },
-    hint: {
-      id: getTrad('EditForm.inputToggle.description.email'),
-      defaultMessage:
-        'Disallow the user to create multiple accounts using the same email address with different authentication providers.',
-    },
+  createField({
+    id: 'email',
+    defaultMessage: 'One account per email address',
     name: 'unique_email',
     type: 'boolean',
-    size: 12,
-  },
-  {
-    label: {
-      id: getTrad('EditForm.inputToggle.label.sign-up'),
-      defaultMessage: 'Enable sign-ups',
-    },
-    hint: {
-      id: getTrad('EditForm.inputToggle.description.sign-up'),
-      defaultMessage:
-        'When disabled (OFF), the registration process is forbidden. No one can subscribe anymore no matter the used provider.',
-    },
+  }),
+  createField({
+    id: 'sign-up',
+    defaultMessage: 'Enable sign-ups',
     name: 'allow_register',
     type: 'boolean',
-    size: 12,
-  },
-  {
-    label: {
-      id: getTrad('EditForm.inputToggle.label.email-reset-password'),
-      defaultMessage: 'Reset password page',
-    },
-    hint: {
-      id: getTrad('EditForm.inputToggle.description.email-reset-password'),
-      defaultMessage: "URL of your application's reset password page.",
-    },
-    placeholder: {
-      id: getTrad('EditForm.inputToggle.placeholder.email-reset-password'),
-      defaultMessage: 'ex: https://youtfrontend.com/reset-password',
-    },
+  }),
+  createField({
+    id: 'email-reset-password',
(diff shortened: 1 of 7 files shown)


hanpaine:
Thanks for the contribution @jb-thery ! 🚀 U&P plugin improvement work is currently planned for Q3 at the latest so we may hold off on adding changes for now. I'll leave it with @maccomaccomaccomacco to take a look 🙏

jb-thery:
@hanpaine @maccomaccomaccomacco Hello, any update ?

packages/plugins/users-permissions/admin/src/pages/AdvancedSettings/utils/schema.js
```diff
@@ -10,6 +10,11 @@ const schema = yup.object().shape({
     then: yup.string().matches(URL_REGEX).required(),
     otherwise: yup.string().nullable(),
   }),
+  email_confirmation_error_redirection: yup.mixed().when('email_confirmation_error', {
```
innerdvations, line 13: The yup rule uses `when('email_confirmation_error')`, but that field doesn't exist in the form, so URL validation never runs and admins can save invalid redirect URLs.

Could you align this with `email_reset_password` (nullable + `URL_REGEX`) and add a schema test that rejects invalid values?

```js
email_confirmation_error_redirection: yup
  .string(translatedErrors.string)
  .matches(URL_REGEX, {
    id: translatedErrors.regex.id,
    defaultMessage: 'This is not a valid URL',
  })
  .nullable(),
```

_AI-assisted comment._

packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -403,7 +403,17 @@ module.exports = ({ strapi }) => ({
 
     const [user] = await userService.fetchAll({ filters: { confirmationToken } });
 
+    const settings = await strapi
+      .store({ type: 'plugin', name: 'users-permissions', key: 'advanced' })
+      .get();
+
     if (!user) {
```
innerdvations, line 410: Could you add API integration tests for this new redirect path? Suggested cases in `tests/api/plugins/users-permissions/content-api/auth.test.api.js`:

- User confirms once → second request with the same token + error URL configured → 302 to the custom URL
- Invalid token + error URL configured → 302
- Invalid token + no error URL → 400 `ValidationError` (existing behaviour)

_AI-assisted comment._

packages/plugins/users-permissions/admin/src/pages/AdvancedSettings/utils/layout.js
```diff
@@ -1,81 +1,62 @@
...
+    id: 'email-confirmation-redirection',
+    defaultMessage: 'Redirection url',
     name: 'email_confirmation_redirection',
-    type: 'string',
-    size: 12,
-  },
+    placeholder: 'ex: https://youtfrontend.com/email-confirmation',
+  }),
+  createField({
+    id: 'email-confirmation-error-redirection',
```
innerdvations, line 55: The new field needs `EditForm.inputToggle.label/description/placeholder.email-confirmation-error-redirection` in `en.json` (and ideally `fr.json`). Without them the admin UI shows the generic `createField` fallback hint (`Description for …`).

_AI-assisted comment._

innerdvations (changes requested):
Thanks for this — the optional error redirect is a sensible UX improvement and the `auth.js` logic matches the existing success-redirect pattern.

A few items to address before merge (blockers inline; follow-ups below):

**Follow-up (UX):** In `AdvancedSettings/index.jsx`, consider disabling `email_confirmation_error_redirection` when email confirmation is off, same as the success redirect field — it only applies when confirmation is enabled.

**Note:** The branch needs a rebase onto `develop` (merge conflicts at the moment).

PR #22678, opened 2025-01-20
chore: upgrade jscodeshift to v17

innerdvations:
### What does it do?

- upgrade jscodeshift to eliminate sub-dependency of unmaintained `rimraf` v2

### Why is it needed?

rimraf v2 is unmaintained and we are trying to eliminate it

### How to test it?

upgrade tool should still work as before (it's the only thing using jscodeshift)

### Related issue(s)/PR(s)

DX-1850


butcherZ (approved):
Tested upgrade tool locally, LGTM!

PR #22683, opened 2025-01-20
feat(content-manager): add getters to useDocument

markkaylor:
### What does it do?

Add two getters to the useDocument hook
- getTitle => get's the document's title (mainField value, singleType displayName, or Untitled)
- getIntiialFormValues => get's the document's form values

### Why is it needed?

Currently these values are only used in one place but soon we will need to access them in the EditView Preview and on-the-fly Relations

### How to test it?

You should see the correct document title on EditViewPage and Preview
You should still have initial values when editing an existing entry


packages/core/content-manager/admin/tests/mockData.ts
```diff
@@ -371,6 +371,9 @@ const CM_CONTENT_TYPE_MOCK_DATA = [
       id: {
         type: 'string',
       },
+      name: {
+        type: 'string',
+      },
```
markkaylor, line 376: ℹ️  name is returned in the mock server response but it is not defined on the schema so I added it

packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -107,13 +117,31 @@ const useDocument: UseDocument = (args, opts) => {
...
+      return document[mainField];
+    }
+
+    if (isSingleType && schema?.info.displayName) {
+      // When it's a singleType without a mainField, use the contentType displayName
+      return schema.info.displayName;
+    }
+
+    // Otherwise, use a fallback
+    return 'Untitled';
```
remidej, line 143: Unrelated but does this mean "Untitled" is never localized?

packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
```diff
@@ -21,7 +21,7 @@ type Transform = <TAttribute extends Schema.Attribute.AnyAttribute>(
   value: any,
   attribute: TAttribute
 ) => any;
-type AnyData = Omit<Document, 'id'>;
+export type AnyData = Omit<Document, 'id'>;
```
remidej, line 24: The export should likely be at the end of the files with the others

remidej (commented):
Haven't tested yet but this looks great :)

packages/core/content-manager/admin/src/hooks/useDocument.ts
markkaylor: Good catch, I'll take a look

packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
```diff
@@ -329,6 +332,40 @@ describe('useDocument', () => {
...
+      useDocument({
+        collectionType: 'collection-types',
+        model: mockData.contentManager.contentType,
+        documentId: '12345',
+      })
+    );
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(result.current.getTitle).toBeInstanceOf(Function);
```
simotae14, line 347: is this assertion needed to check the next assertion?

packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -107,13 +117,31 @@ const useDocument: UseDocument = (args, opts) => {
...
   const {
     components,
     schema,
     schemas,
     isLoading: isLoadingSchema,
   } = useContentTypeSchema(args.model);
+  const isSingleType = schema?.kind === 'singleType';
+
+  const getTitle = (mainField: string) => {
+    if (mainField !== 'id' && document?.[mainField]) {
```
simotae14, line 132: I think it is better to move the comment before the conditions just to explain better the different scenarios, for example

`
// Use always the mainField if it's not an id
if (mainField !== 'id' && document?.[mainField]) {

`

packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
markkaylor: No it's not needed I just saw some other tests also checked this, so stayed consistent 🤷‍♂️

packages/core/content-manager/admin/src/hooks/useDocument.ts
markkaylor: I don't really have a preference, if you prefer it this way I can update.

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -203,7 +203,12 @@ const ListViewPage = () => {
     return <Page.Error />;
   }
 
-  const contentTypeTitle = schema?.info.displayName ?? 'Untitled';
+  const contentTypeTitle = schema?.info.displayName
+    ? formatMessage({ id: schema.info.displayName, defaultMessage: schema.info.displayName })
```
markkaylor, line 207: This fixes the issue described here: https://github.com/strapi/strapi/issues/20461#issuecomment-2606591970

simotae14 (approved):
Haven't tested it yet but the code looks great, 🌟 well done!

remidej (approved):
Tested, all good

PR #22689, opened 2025-01-21
Feat/sdk-content-api-generation

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/strapi/src/cli/commands/generate-schema.ts
+++ b/packages/core/strapi/src/cli/commands/generate-schema.ts
@@ -10,21 +10,17 @@ import type { StrapiCommand } from '../types';
 interface CmdOptions {
   silent?: boolean;
   outDir?: string;
-  typesDir?: string;
 }
 
 // Filter to only include content types where the UID begins with "api::"
 // TODO: temporary restriction to API content types
-const restrictToApiContentTypes = (uid: string) => {
-  let result;
-
+const restrictToApiContentTypes = (uid: string): boolean => {
   try {
-    result = uid.startsWith('api::');
+    return uid.startsWith('api::');
   } catch (e) {
-    console.error('error', e);
+    console.error('Error checking UID:', e);
+    return false;
   }
-
-  return result;
 };
 
 const allCrudOperationNames = ['create', 'update', 'delete', 'findOne', 'find'];
@@ -69,13 +65,12 @@ const action = async ({ silent, outDir }: CmdOptions) => {
       contentTypes(contentType: Struct.CollectionTypeSchema) {
         // Strip private attributes
         const attributes = Object.fromEntries(
-          Object.entries(contentType.attributes).filter(([, attr]) => {
-            return (
+          Object.entries(contentType.attributes).filter(
+            ([, attr]) =>
               attr.type !== 'password' &&
               // @ts-expect-error TODO: fix this
               !attr.private
-            );
-          })
+          )
         );
         return { ...contentType, attributes };
       },
@@ -89,25 +84,25 @@ const action = async ({ silent, outDir }: CmdOptions) => {
       Struct.CollectionTypeSchema | Struct.SingleTypeSchema,
     ][];
 
-  // Generate a single import statement for generated types
+  // Map UIDs to generated type names
   const uidToGeneratedTypeName = new Map<string, string>();
-  const allTypeNames = apiContentTypes.map(
-    ([uid, contentType]: [string, Struct.CollectionTypeSchema | Struct.SingleTypeSchema]) => {
-      const {
-        info: { singularName },
-      } = contentType;
+  const allTypeNames = apiContentTypes.map(([uid, contentType]) => {
+    const {
+      info: { singularName },
+    } = contentType;
 
-      const formattedSingularName = singularName
-        .split('-')
-        .map((part: string) => part.charAt(0).toUpperCase() + part.slice(1))
-        .join('');
+    const formattedSingularName = singularName
+      .split('-')
+      .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+      .join('');
 
-      const typeName = `Api${formattedSingularName}${formattedSingularName}`;
-      uidToGeneratedTypeName.set(uid, typeName);
+    const typeName = `Api${formattedSingularName}${formattedSingularName}`;
+    uidToGeneratedTypeName.set(uid, typeName);
 
-      return typeName;
-    }
-  );
+    return typeName;
+  });
+
+  // Generate a single import statement for generated types
   const combinedImportStatement = `import { ${allTypeNames.join(', ')} } from './generated/contentTypes';`;
 
   // Helper function to generate TypeScript interface definitions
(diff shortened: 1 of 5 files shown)


derrickmehaffy:
This looks a little confusing with the generate:api command  🤔

jhoward1994:
@derrickmehaffy do you mean naming wise it conflicts with `strapi generate`

'generate:api' doesn't exist in v5, afaik that command is all interactive through the CLI. That makes the separation clear enough to me. Would you suggest a different naming?

derrickmehaffy:
> @derrickmehaffy do you mean naming wise it conflicts with `strapi generate`
> 
> `generate:api` doesn't exist in v5, afaik that command is all interactive through the CLI. That makes the separation clear enough to me. Would you suggest a different naming?

Confusing I think more than conflicting. I'll double check but I was fairly certain that command existed.

If it's for types though I think we should piggy back on the existing command for that: https://docs.strapi.io/dev-docs/cli#ts-generate

innerdvations:
For more context, this command is generating a dump of project data which includes public schemas in a javascript format, and the existing types files, and may eventually include routes and other public information about the project. The file will not be used by Strapi itself, it will be needed in your external project in conjunction with the javascript SDK to provide your project context to the SDK so that it's not completely generic.

We decided to put up this PR without a final decision on the name specifically to request feedback from others, because there are so many options and none of them seem great, so we welcome any community comments on it.

So thanks for the comments @derrickmehaffy and if you know anyone else who would be interested, please point them to this PR/comment!

And on that note, I agree, I don't like using 'content-api' in this context. To me something like 'app' or 'project' sounds better. We wanted to avoid 'sdk' because we don't necessarily want this to always be sdk-specific, although it's a decent option too.

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+  const uidToGeneratedTypeName = new Map<string, string>();
+  const allTypeNames = apiContentTypes.map(([uid, contentType]) => {
+    const {
+      info: { singularName },
+    } = contentType;
+
+    const formattedSingularName = singularName
+      .split('-')
+      .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+      .join('');
```
innerdvations, line 97: Can we use something like lodash/fp `camelCase` ?

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+        ([, contentType]) =>
+          `${contentType.info.singularName.replace(/-/g, '')}: ${contentType.info.singularName.replace(/-/g, '')}`
+      )
+      .join(';\n')}\n}
+  `;
+
+  // Determine output directory
+  const outputPath = `${outputDir}/schema.ts`;
+
+  const formattedOutput = await prettier.format(
```
innerdvations, line 180: Can we include a "version" that is not necessarily the strapi version (in fact, that could be a security risk since this file is public) but a file-format version, just in case we need to make structural changes without it breaking?  For now we can just fix it at "0.0.1" and hopefully never change it.

Convly (commented):
Apart from the code itself or the CLI naming, what I'm a bit worried about, is that we're focusing on generating the types only, as opposed to a format that could export both the types and actual schema (endpoints, params, etc...). Do we envision that this command will also generate the other artifacts in the future?

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+        // Strip private attributes
+        const attributes = Object.fromEntries(
+          Object.entries(contentType.attributes).filter(
+            ([, attr]) =>
+              attr.type !== 'password' &&
+              // @ts-expect-error TODO: fix this
+              !attr.private
+          )
+        );
+        return { ...contentType, attributes };
```
Convly, line 75: To align the behavior with other APIs, we could use the content types utils (@strapi/utils) to whether an attribute is private or not

```suggestion
        const attributes = Object.fromEntries(
          Object.entries(contentType.attributes)
            // Skip passwords
            .filter(([, attribute]) => attribute.type !== 'password')
            // Strip private attributes
            .filter(([name]) => !contentTypesUtils.isPrivateAttribute(contentType, name))
        );

        return { ...contentType, attributes };
```

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+      },
+    },
+  });
+
+  const contentTypes = app.get('content-types').getAll();
+  const apiContentTypes: [string, Struct.CollectionTypeSchema | Struct.SingleTypeSchema][] =
+    Object.entries(contentTypes).filter(([uid]) => restrictToApiContentTypes(uid)) as [
+      string,
+      Struct.CollectionTypeSchema | Struct.SingleTypeSchema,
+    ][];
```
Convly, line 85:
```suggestion
  const contentTypes = app.get('content-types').getAll() as Record<
    UID.ContentType,
    Struct.ContentTypeSchema
  >;

  const apiContentTypes = Object.entries(contentTypes).filter(([uid]) =>
    restrictToApiContentTypes(uid)
  );
```

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+    uidToGeneratedTypeName.set(uid, typeName);
+
+    return typeName;
+  });
+
+  // Generate a single import statement for generated types
+  const combinedImportStatement = `import { ${allTypeNames.join(', ')} } from './generated/contentTypes';`;
+
+  // Helper function to generate TypeScript interface definitions
+  const generateInterfaceDefinition = (
```
Convly, line 109: I believe it could be easier to use the TS factory API to generate the interfaces rather than trying to generate raw strings, what do you think?

packages/core/strapi/src/cli/commands/generate-schema.ts
```diff
@@ -0,0 +1,210 @@
...
+    console.error('Error checking UID:', e);
+    return false;
+  }
+};
+
+const allCrudOperationNames = ['create', 'update', 'delete', 'findOne', 'find'];
+
+const crudOperationNames = {
+  collectionType: allCrudOperationNames,
+  singleType: allCrudOperationNames.filter((op) => op !== 'find'),
```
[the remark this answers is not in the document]
Convly: Also, do we actually have a create + an update for single types? Isn't it only find/update/delete?

packages/core/strapi/src/cli/commands/generate-schema.ts
Convly: Also, maybe we could consider moving the schema generation part to the typescript generators? Or do we want to keep the logic in the action on purpose?

jhoward1994:
Closing for now as we are revisiting the approach to building the typed SDK which this command was in support of. I will reopen this PR if we need it in future :)

PR #22693, opened 2025-01-21
feat(content-manager): add container queries to form fields

markkaylor:
### What does it do?

- Adds two styled components to create a responsive grid container. I'm exporting them from the FormLayout which I think makes sense, but open to better ideas
- Uses the responsive grid to make all fields (primitives, components, and dynamic zones) responsive
- It does NOT handle the rich text editor or blocks

### Why is it needed?

To prepare for the preview side editor

### How to test it?

To make the container the form is in really small you can just swap the FormLayout grid item col values like this:

```
  <Grid.Item col={3} s={12} direction="column" alignItems="stretch">
    <Tabs.Content value="draft">
      <FormLayout layout={layout} />
    </Tabs.Content>
    <Tabs.Content value="published">
      <FormLayout layout={layout} />
    </Tabs.Content>
  </Grid.Item>
  <Grid.Item col={9} s={12} direction="column" alignItems="stretch">
    <Panels />
  </Grid.Item>
```
- Check all types (text, datetime, number, etc... )
- Check the component picker buttons
- Check components in dynamic zones
- Check repeatable and non-repeatable component

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
@@ -2,23 +2,11 @@ import { forwardRef, memo } from 'react';
 
 import { DateTimePicker, useComposedRefs, Field } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
-import { styled } from 'styled-components';
 
 import { useFocusInputField } from '../../hooks/useFocusInputField';
 import { useField } from '../Form';
 
 import { InputProps } from './types';
-
-const ResponsiveFieldRoot = styled(Field.Root)`
-  > div {
-    flex-direction: column;
-    align-items: flex-start;
-
-    div {
-      width: 100%;
-    }
-  }
-`;
 
 const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
@@ -30,7 +18,7 @@ const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
     const value = typeof field.value === 'string' ? new Date(field.value) : field.value;
 
     return (
-      <ResponsiveFieldRoot error={field.error} name={name} hint={hint} required={required}>
+      <Field.Root error={field.error} name={name} hint={hint} required={required}>
         <Field.Label action={labelAction}>{label}</Field.Label>
         <DateTimePicker
           ref={composedRefs}
@@ -45,7 +33,7 @@ const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
         />
         <Field.Hint />
         <Field.Error />
-      </ResponsiveFieldRoot>
+      </Field.Root>
     );
   }
 );
--- a/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
@@ -2,11 +2,23 @@ import { forwardRef, memo } from 'react';
 
 import { DateTimePicker, useComposedRefs, Field } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
 
 import { useFocusInputField } from '../../hooks/useFocusInputField';
 import { useField } from '../Form';
 
 import { InputProps } from './types';
+
+const ResponsiveFieldRoot = styled(Field.Root)`
+  > div {
+    flex-direction: column;
+    align-items: flex-start;
+
+    div {
+      width: 100%;
+    }
+  }
+`;
 
 const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
@@ -18,7 +30,7 @@ const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
     const value = typeof field.value === 'string' ? new Date(field.value) : field.value;
 
     return (
-      <Field.Root error={field.error} name={name} hint={hint} required={required}>
+      <ResponsiveFieldRoot error={field.error} name={name} hint={hint} required={required}>
         <Field.Label action={labelAction}>{label}</Field.Label>
         <DateTimePicker
           ref={composedRefs}
@@ -33,7 +45,7 @@ const DateTimeInput = forwardRef<HTMLInputElement, InputProps>(
         />
         <Field.Hint />
         <Field.Error />
-      </Field.Root>
+      </ResponsiveFieldRoot>
     );
   }
 );
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
@@ -5,6 +5,7 @@ import { useIntl } from 'react-intl';
 import { styled } from 'styled-components';
 
 import { ComponentIcon } from '../../../../../components/ComponentIcon';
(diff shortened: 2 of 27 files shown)


packages/core/admin/admin/src/components/FormInputs/DateTime.tsx
```diff
@@ -2,12 +2,24 @@ import { forwardRef, memo } from 'react';
...
+const ResponsiveFieldRoot = styled(Field.Root)`
+  > div {
+    flex-direction: column;
+    align-items: flex-start;
+
+    div {
+      width: 100%;
+    }
+  }
+`;
```
markkaylor, line 21: ⚠️ not sure this is working as expected, looking into it

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
```diff
@@ -59,15 +60,23 @@ const ComponentCategory = ({
...
+  container-type: inline-size;
+`;
+
 const Grid = styled(Box)`
   display: grid;
-  grid-template-columns: repeat(auto-fit, 14rem);
+  grid-template-columns: repeat(auto-fill, 100%);
   grid-gap: ${({ theme }) => theme.spaces[1]};
+
+  @container (min-width: ${() => RESPONSIVE_CONTAINER_BREAKPOINTS.sm}) {
```
remidej, line 77: no need for a function

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -1,11 +1,27 @@
...
 import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
 
 import { useDoc } from '../../../hooks/useDocument';
 import { EditLayout } from '../../../hooks/useDocumentLayout';
 
 import { InputRenderer } from './InputRenderer';
 
+export const RESPONSIVE_CONTAINER_BREAKPOINTS = {
+  sm: '440px',
```
remidej, line 11: maybe use rem so it scales with the user's browser settings

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
```diff
@@ -59,15 +60,23 @@ const ComponentCategory = ({
...
           ))}
         </Grid>
-      </Accordion.Content>
+      </ResponsiveAccordionContent>
     </Accordion.Item>
   );
 };
 
+const ResponsiveAccordionContent = styled(Accordion.Content)`
+  container-type: inline-size;
```
remidej, line 69: Looks like we need to edit our eslint setup because ts-styled-plugin warns about this being an unknown rule. That package is deprecated so we need to replace it with @styled/typescript-styled-plugin

remidej (commented):
screenshot 2025-01-22 at 10 49 16@2x
the date time picker is the only one that seems broken on a super narrow form, but I think it's fine as the side editor's min width will be larger

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -1,11 +1,27 @@
...
+};
+
+export const ResponsiveGridRoot = styled(Grid.Root)`
+  container-type: inline-size;
+`;
+
+export const ResponsiveGridItem = styled(Grid.Item)`
+  grid-column: span 12;
+  @container (min-width: ${() => RESPONSIVE_CONTAINER_BREAKPOINTS.sm}) {
```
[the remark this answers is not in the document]
markkaylor: weird yesterday it was not happy when I didn't provide a function 🤷

markkaylor:
Keep going smaller and they all start to overflow

>  but I think it's fine as the side editor's min width will be larger

but yeah that's what I concluded too

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
markkaylor: I get an error here without a function 🤔 

```
Cannot access 'RESPONSIVE_CONTAINER_BREAKPOINTS' before initialization
```

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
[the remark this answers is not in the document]
remidej: oh yeah? I thought functions were only needed to access theme values, didn't think you'd need it for string interpolation. but keep it then :)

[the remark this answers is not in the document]
markkaylor: I thought the same as you, and it looks like it is working here without the function, but the other file has an issue 😠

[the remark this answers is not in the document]
remidej: it's gotta be a scope or hoisting issue I guess? probably not worth investigating

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
markkaylor: I've gotten rid of the warning but I'm a bit confused. I couldn't find any dependency for `typescript-styled-plugin` (which is deprecated). I tried installing https://github.com/styled-components/typescript-styled-plugin, but it doesn't seem necessary to install it for it to work...

Adding the following config to `packages/core/content-manager/admin/tsconfig.json` does the trick but I don't understand how it's working if the plugin ins't installed 🤔 

```
    "plugins": [
      {
        "name": "@styled/typescript-styled-plugin",
        "lint": {
          "validProperties": ["container-type"],
          "unknownAtRules": "ignore"
        }
      }
    ]
```

.vscode/settings.json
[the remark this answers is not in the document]
markkaylor: sorry had to force push to fix this

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/ComponentCategory.tsx
remidej: Huh that is a bit of a mystery indeed. Maybe the warning just comes from something with our vscode configs? Anyway I wouldn't recommend writing a reference to a package we know is not there 🤔

markkaylor: Man it took me forever but I finally figured out that `@styled/typescript-styled-plugin` is installed through the vscode styled components extension, that's why this config is working. I also learned the vscode extension is no longer maintained :( https://github.com/styled-components/vscode-styled-components/issues/275. 

What do you think should I leave the config and add a comment explaining where the dep comes from? Or should I remove it and we live with the warning 🤔

markkaylor: And just for the record...AI was no help...not that I'm keeping score or anything

remidej: I think living with the warning is safer, I'm afraid referring that rule may break for people using other IDEs

markkaylor: Sounds good, updated!

remidej (approved):

PR #22694, opened 2025-01-21
fix(upload): folder subtitles for folders without assets or subfolders

jorrit:
### What does it do?

Before:

image

After:

image

It also fixes the Spanish language, the translation seemed inaccurate.

### Why is it needed?

The subtitle of folders is incorrect for zero subfolders and / or zero assets.

### How to test it?

Create a folder, observe the subtitle in the Media Library.


hanpaine:
Thanks for the contribution @jorrit  ! 🚀  we'll take a look soon 🙏

PR #22702, opened 2025-01-22
fix(content-manager): renable relations api tests

simotae14:
### What does it do?

Renable some skipped test in the relations.test.api.js file related to the Cloning action

### Why is it needed?

Because before v5 release some tests were skipped

### How to test it?

Run the api test from the root
`yarn test:api tests/api/core/content-manager/api/relations.test.api.js`

### Related issue(s)/PR(s)

CS-1194

### DIFF
--- a/tests/api/core/content-manager/api/relations.test.api.js
+++ b/tests/api/core/content-manager/api/relations.test.api.js
@@ -1061,7 +1061,7 @@ describe('Relations', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


MarionLemaire:
Tested today on PR branch : 
I've duplicated entries with existing relations, and connect/disconnect other relations to these duplicated entries, and I don't face any issue so far.
I don't get why API tests are failing on theses scenarios.
image

PR #22728, opened 2025-01-24
Fix 'woops' spelling disagreement 

Link2Twenty:
### What does it do?

I've added the missing `h` to whoops.

### Why is it needed?

To correct the spelling.

### How to test it?

N/A

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/packages/core/admin/admin/src/components/PageHelpers.tsx
+++ b/packages/core/admin/admin/src/components/PageHelpers.tsx
@@ -79,7 +79,7 @@ const Error = (props: ErrorProps) => {
           icon={<WarningCircle width="16rem" />}
           content={formatMessage({
             id: 'anErrorOccurred',
-            defaultMessage: 'Woops! Something went wrong. Please, try again.',
+            defaultMessage: 'Whoops! Something went wrong. Please, try again.',
           })}
           {...props}
         />
--- a/packages/core/admin/admin/src/components/tests/PageHelpers.test.tsx
+++ b/packages/core/admin/admin/src/components/tests/PageHelpers.test.tsx
@@ -23,7 +23,7 @@ describe('PageHelpers', () => {
       render(<Page.Error />);
 
       expect(
-        screen.getByText('Woops! Something went wrong. Please, try again.')
+        screen.getByText('Whoops! Something went wrong. Please, try again.')
       ).toBeInTheDocument();
     });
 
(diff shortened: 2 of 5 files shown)


innerdvations:
@soupette I know it's been a while ( @maccomaccomaccomacco still misses you ) but could you explain  if your original intention was for this line to be ["oops" or "whoops"](https://www.difference.wiki/whoops-vs-oops/)? We should have more context before accepting this change.

hanpaine:
Thanks for the contribution @Link2Twenty! 🚀  It's a fair challenge from @innerdvations; "whoops" would be the correct spelling when linking to the word of origin, but as we know, languages are ever-changing and living. The chaotic nature of English means the 'h' is absent from the pronunciation of "whoops" (unlike "whip" as in "cool-whip") and so "woops" is also an acceptable spelling variant, as per [OED](https://www.oed.com/search/dictionary/?scope=Entries&q=woops), [Merriam-Webster ](https://www.merriam-webster.com/dictionary/woops)and [Collins](https://www.collinsdictionary.com/dictionary/english/woops) 🤷 🤓 I'd argue this is a spelling disagreement, rather than spelling mistake @Link2Twenty 😁

innerdvations:
@hanpaine That sheds more confusion on it unfortunately. Merriam-Webster considers it an alternative to "oops" but Collins claims it to be an alternative to "whoops"

However, I would still consider "woops" as either a mistake or a colloquialism, as over the past century, and especially prior to the internet era, it has been almost exclusively published as a proper name.

soupette:
> @soupette I know it's been a while ( @maccomaccomaccomacco still misses you ) but could you explain if your original intention was for this line to be ["oops" or "whoops"](https://www.difference.wiki/whoops-vs-oops/)? We should have more context before accepting this change.

I believe it was Oops, I do also miss @maccomaccomaccomacco

derrickmehaffy:
> > @soupette I know it's been a while ( @maccomaccomaccomacco still misses you ) but could you explain if your original intention was for this line to be ["oops" or "whoops"](https://www.difference.wiki/whoops-vs-oops/)? We should have more context before accepting this change.
> 
> I believe it was Oops, I do also miss @maccomaccomaccomacco 

And we miss you sir 🥹

innerdvations:
It looks like the [Cloud CLI uses "Oops"](https://github.com/strapi/strapi/blob/develop/packages/cli/cloud/src/login/action.ts#L61) as well, although they use a 🥲 emoji with it which I wouldn't recommend in this situation to avoid introducing potential encoding issues.

innerdvations (approved):

Issue #22437, opened 2024-12-13
Strapi admin panel freezes and stops working when searching in a closed dropdown menu


paulovitorsd:
### Required System information

- Node.js version: v18.20.4
- NPM version: 10.7.0
- Strapi version: v5.5.1
- Database: sqlite (default configuration)
- Operating system: Ubuntu 22.04.4 LTS
- Is your project Javascript or Typescript: Typescript

### Describe the bug

The Strapi admin panel becomes unresponsive when performing a search in a relational field while the dropdown displaying relation items is closed. This issue prevents further interaction with the page and forces the user to refresh the browser.

### Steps to reproduce the behavior

1. Open the **Content Manager** in the Strapi admin panel.
2. Navigate to a collection type that includes a relation field.
   - For example, the **Article** entity created during the Strapi setup (`npx create-strapi@latest project`).
3. Ensure the relation dropdown is **closed but focused**.
   - Double-click on the dropdown to focus it and begin typing.
4. Use the search bar to filter relation items.
5. Observe that the admin panel freezes, and no further interactions are possible.

### Expected behavior

The search should filter relation items smoothly, without causing the admin panel to freeze, regardless of whether the dropdown is open or closed.

### Screenshots
- In the video the problem occurs the second time I start typing in the input.

https://github.com/user-attachments/assets/2cfd9c77-926a-4b73-9d8e-4f21ec617a87

dongnguyenbca:
I have the same problem; I'm stuck in production.

paulovitorsd:
@dongnguyenbca

I was able to temporarily fix this by removing this `selectFromResult` function [here](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx#L196-L204), so instead of dealing with formatting the result inside the function you can do it outside the `useGetRelationsQuery` hook.

To modify this piece of code in your project while Strapi doesn't solve this, you can detect this file within your `node_modules` and modify it with [patch-package](https://www.npmjs.com/package/patch-package). Of course this is a temporary solution, but it works for now.

Dupflo:
Same !

Node.js version: v20.18.0
NPM version: 10.8.2
Strapi version: v5.6.0
Database: Postgres
Operating system: MacOs
Is your project Javascript or Typescript: Typescript

derrickmehaffy:
was able to reproduce in v5.6.0 as I was rapidly trying to type in relation entries

[...]

PR #22741, opened 2025-01-27
fix: debounce relation search

Marc-Roig:
### What does it do?
EditView relation search did not have any debouncing, every written character triggered an API request.
This PR introduces such behaviour.

Note: If this is not enough, we could abort the rtk query by using the .abort() method, but it feels less responsive if using it, so lets try if this works.

### How to test it?

See https://github.com/strapi/strapi/issues/22437 for reproduction
Continuously type in a relational field search and at some point your client will stop responding

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/22437

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
@@ -36,6 +36,7 @@ import { styled } from 'styled-components';
 import { RelationDragPreviewProps } from '../../../../components/DragPreviews/RelationDragPreview';
 import { COLLECTION_TYPES } from '../../../../constants/collections';
 import { ItemTypes } from '../../../../constants/dragAndDrop';
+import { useDebounce } from '../../../../hooks/useDebounce';
 import { useDoc } from '../../../../hooks/useDocument';
 import { type EditFieldLayout } from '../../../../hooks/useDocumentLayout';
 import {
@@ -461,6 +462,7 @@ const RelationsInput = ({
   const fieldRef = useFocusInputField<HTMLInputElement>(name);
   const field = useField<RelationsFormValue>(name);
 
+  const searchParamsDebounced = useDebounce(searchParams, 300);
   const [searchForTrigger, { data, isLoading }] = useLazySearchRelationsQuery();
 
   /**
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
```diff
@@ -478,18 +478,22 @@ const RelationsInput = ({
...
-      params: {
-        ...buildValidParams(query),
-        id: id ?? '',
-        pageSize: 10,
-        idsToInclude: field.value?.disconnect?.map((rel) => rel.id.toString()) ?? [],
-        idsToOmit: field.value?.connect?.map((rel) => rel.id.toString()) ?? [],
-        ...searchParams,
-      },
-    });
+    const timeout = setTimeout(() => {
```
remidej, line 481: Have you checked out the useDebounce hook in the CM?

Marc-Roig: yes but I didn't quite see a way to use it here, and prevent the actual api request. Do you think we could use it here?

remidej: I would guess something like this pseudocode:

```jsx
const debouncedFieldValue = useDebounce(field.value, 300);

React.useEffect(() => {
  searchForTrigger(debouncedFieldValue);
}, [debouncedFieldValue]);
```

Also I haven't looked into it in detail but I'm not sure why we need a lazy query. A non-lazy one would simplify things:

```
const debouncedFieldValue = useDebounce(field.value, 300);

const { data, isLoading } = useLazySearchRelationsQuery(debouncedFieldValue);
```

Marc-Roig: ah I see! thank you <3

Marc-Roig: that works wonders! great :)

markkaylor: I noticed this hook is not documented in the [contributor docs](https://contributor.strapi.io/docs/core/content-manager/hooks/use-content-types). Do you think we should document it now so it is easier to find and use?

remidej (approved):
Tested, nice improvement!

PR #22747, opened 2025-01-27
enhancement: pre-generated api tokens

Convly:
### What does it do?

- When Strapi starts, check if there are no admin users + no API tokens
	- If this is the case, create two default API tokens: one read-only and one full-access
	- Else, do nothing

### Why is it needed?

This is the first step is to ease the product on-boarding process and remove friction for new applications.

Basically, as a developer, I want API tokens to be automatically generated after creating a project so that I don’t have to create them manually.

### Note

- Since API Tokens can't be viewed after being created, this still forces the tokens to be regenerated to be used the first time
- We had to introduce a not-so-ideal way of restarting Strapi manually by creating a file then removing it. This is supposed to be temporary while we think about a better solution

### How to test it?

- Create a new application
- Sign up a first admin user
- Go to the API token settings page
- There should be 2 pre generated tokens: 1 read only, 1 full acess


Convly:
Experimental: `0.0.0-experimental.5bd54472327d69041855d9b38a3d4e445535c9a3`

packages/core/admin/server/src/bootstrap.ts
```diff
@@ -69,6 +69,43 @@ const syncAPITokensPermissions = async () => {
...
+  const apiTokenService = getService('api-token');
+
+  const usersCount = await userService.count();
+  const apiTokenCount = await apiTokenService.count();
+
+  if (usersCount === 0 && apiTokenCount === 0) {
+    await apiTokenService.create({
+      name: 'Read Only',
+      description:
+        'A default API token with read-only permissions, only used for accessing resources',
```
innerdvations, line 94: Not necessarily in this PR, but ideally we should look up a string in the default locale so this can be translated, otherwise non-english users might have very little idea what this is or why it's here.

tests/e2e/utils/file-reset.ts
```diff
@@ -45,11 +16,13 @@ export const resetFiles = async () => {
...
-  await delay(3); // give it time to detect file changes and begin its restart
-  await pollHealthCheck(); // give it time to come back up
+    // wait for the server to restart after modifying files
+    console.log('Waiting for Strapi to restart...');
+
+    // TODO:  This is both a waste of time and flaky.
+    //        We need to find a way to access playwright server output and watch for the "up" log to appear
+    await delay(3); // give it time to detect file changes and begin its restart
+    await pollHealthCheck(); // give it time to come back up
+  }
```
innerdvations, line 27: Since we're fixing this, can we just do:

```
if (!process.env.TEST_APP_PATH) {
    console.warn("Could not reset files because TEST_APP_PATH not set");
    return;
}
```

innerdvations (approved):
LGTM, just a suggestion

packages/core/admin/server/src/bootstrap.ts
```diff
@@ -69,6 +69,43 @@ const syncAPITokensPermissions = async () => {
...
+ */
+
+const createDefaultAPITokensIfNeeded = async () => {
+  const userService = getService('user');
+  const apiTokenService = getService('api-token');
+
+  const usersCount = await userService.count();
+  const apiTokenCount = await apiTokenService.count();
+
+  if (usersCount === 0 && apiTokenCount === 0) {
```
innerdvations, line 90: There might be a better way to check if this is the first startup without adding two queries on every reboot. Like maybe something in core store?

Convly: What if you manually wipe out your database to reset the data? What about partial data transfer, should the core store flag be transferred as well?

I tried coming out with a better solution but couldn't, if you have an idea please let me know :confused:

innerdvations (approved):
Tested and works for me

packages/core/admin/server/src/bootstrap.ts
Convly: I'll add a todo about moving the logic to first run completion status and create a task to implement this (or similar) in another PR. Does that make sense?

innerdvations: Sounds good to me!

PR #22750, opened 2025-01-28
ci: use release app token permissions for release workflow

Bassel17:
### What does it do?

Use github app token for permissions in release workflow

### Why is it needed?

Normal github token cannot bypass protected branches


innerdvations (approved):
Seems right to me, but I don't know how to QA it beyond trying it out after it's merged

innerdvations (approved):

PR #22753, opened 2025-01-28
feat: responsive blocks editor toolbar

remidej:
### What does it do?

- exports the useElementOnScreen hook in the admin package, and import it in HeaderLayout where it was previously defined
- uses that hook in the Blocks editor to make it responsive. I use it in 2 places:
  - in the more menu trigger (the three dots button): the trigger is displayed at the very end, if it's not 100% visible, it will hide one more item in the toolbar, until it's visible.
  - in each item in the toolbar: to check check they become 100% visible again if they are already hidden

What it does not do:
- Show an active style in the more menu when appropriate
- Show the icons in the more menu

Both are because of limitations of the Menu in the design system. I could either ignore it, force custom styles, or create issues/PRs in the design system. Not sure yet, but it shouldn't block the work on the markdown side.

### Why is it needed?

Responsive blocks editor will be required for the Preview side editor

### How to test it?

Make your browser smaller, make it larger, see the options in the toolbar and in the more menu. Make sure they actually work.

Also sanity check the HeaderLayout sticky behavior. You can see it when editing a users and permissions role in the settings.
### Related issue(s)/PR(s)


packages/core/admin/admin/src/hooks/useElementOnScreen.ts
```diff
@@ -0,0 +1,32 @@
...
+const useElementOnScreen = <TElement extends HTMLElement = HTMLElement>(
+  callback: IntersectionObserverCallback,
+  options?: IntersectionObserverInit
+): React.RefObject<TElement> => {
+  const containerRef = React.useRef<TElement>(null);
+  const id = React.useId();
+
+  React.useEffect(() => {
+    const containerEl = containerRef.current;
+    const observer = new IntersectionObserver(callback, options);
```
remidej, line 16: I'm considering abstracting the callback further, so the hook would take onIsVisibleChange. Less flexible but easier to use and understand 🤔

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -341,10 +344,12 @@ const isListNode = (node: unknown): node is Block<'list'> => {
 interface ListButtonProps {
   block: BlocksStore['list-ordered'] | BlocksStore['list-unordered'];
   format: Block<'list'>['format'];
+  location?: 'toolbar' | 'menu';
 }
 
-const ListButton = ({ block, format }: ListButtonProps) => {
+const ListButton = ({ block, format, location = 'toolbar' }: ListButtonProps) => {
```
remidej, line 350: I'm really not a fan of this 'location' pattern, but it's a quick way to adapt the UI while benefiting from the existing logic. But the logic should be moved into the blocks store in my opinion, just like for the modifiers.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -538,32 +667,104 @@ const BlocksToolbar = () => {
...
+        />,
+        <ListButton
+          block={blocks['list-ordered']}
+          format="ordered"
+          location="menu"
+          key="ordered"
+        />,
+      ],
+    },
+  ] as Array<ObservedItem>;
```
remidej, line 727: This makes the component quite long, maybe we could extract it into a useFlexibleItems hook so it's easier to read

remidej (commented):
some thoughts on the stuff I'm unsure of

packages/core/admin/admin/src/hooks/useElementOnScreen.ts
markkaylor: I'd be interested to hear more of your reasoning and discuss it but I think the way you have it makes sense.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -496,23 +516,132 @@ const LinkButton = ({ disabled }: { disabled: boolean }) => {
...
 
+interface ObservedToolbarItemProps {
+  index: number;
+  lastVisibleIndex: number;
+  setLastVisibleIndex: React.Dispatch<React.SetStateAction<number>>;
+  rootRef: React.RefObject<HTMLElement>;
+  children: React.ReactNode;
+}
+
+const ObservedToolbarItem = ({
```
markkaylor, line 552: Would this be something we can abstract to share with the markdown editor as well?

markkaylor (commented):
Gave it a quick look, nice work this looks promising ✨ I'll test it out and take another look later today

packages/core/admin/admin/src/hooks/useElementOnScreen.ts
remidej: When writing the callback I realized that entry.isIntersecting is weird, it's not clear if it means it's visible or not. Since the hook is useElementOnScreen and not useIntersectionObserver I find it weird to deal with something low-level like this.

So I think I'll change it, if we need more control somewhere else in the future we could always call IntersectionObserver directly

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -538,32 +667,104 @@ const BlocksToolbar = () => {
...
+        </Menu.Item>
+      ),
+    })),
+    {
+      renderInToolbar: () => <LinkButton disabled={isButtonDisabled} location="toolbar" />,
+      renderInMenu: () => <LinkButton disabled={isButtonDisabled} location="menu" />,
+    },
+    {
+      // List buttons can only be rendered together when in the toolbar
+      renderInToolbar: () => (
```
markkaylor, line 704: do these need to be functions? Couldn't they just be object keys? That would also match the location prop thing you got going. 

```
toolbar: <ToolbarComponent />
menu: <MenuComponent />
```
Then used like:
```tsx
 {observedItems
              .map((item, index) => (
                <ObservedToolbarItem
                  lastVisibleIndex={lastVisibleIndex}
                  setLastVisibleIndex={setLastVisibleIndex}
                  index={index}
                  rootRef={toolbarRef}
                  key={index}
                >
                  {item['toolbar']} <---------here!
                </ObservedToolbarItem>
              ))
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -538,32 +667,104 @@ const BlocksToolbar = () => {
 
   const isButtonDisabled = checkButtonDisabled();
 
+  /**
+   * Observed items are ones that may or may not be visible in the toolbar, depending on the
+   * available space. They provide two render props:
+   * - renderInToolbar: for when we try to render the item in the toolbar (may be hidden)
+   * - renderInMenu: for when the item didn't fit in the toolbar and is relegated to the "more" menu
+   */
+  const observedItems = [
```
markkaylor, line 677: What about `observedComponents` since it's an array of components?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -538,32 +667,104 @@ const BlocksToolbar = () => {
...
+               */
+              .toSpliced(
+                lastVisibleIndex + 1,
+                0,
+                <MoreMenu
+                  setLastVisibleIndex={setLastVisibleIndex}
+                  hasHiddenItems={hasHiddenItems}
+                  rootRef={toolbarRef}
+                >
+                  {observedItems.slice(lastVisibleIndex + 1).map((item) => item.renderInMenu())}
```
markkaylor: Any value in storing `lastVisibleIndex + 1` in a variable with the other derived values since it's used twice?

markkaylor (commented):
When you select an item from the toolbar or the menu with headings the focus goes to the editor. When you select from the more menu the focus goes back to the menu trigger. The active styles on the trigger look strange.

https://github.com/user-attachments/assets/582b4a45-c139-4e05-9ea1-07fb3321c9a4

> What it does not do:
> - Show an active style in the more menu when appropriate
> - Show the icons in the more menu

Probably you are aware, but also need to add the section separators

Screenshot 2025-01-30 at 11 35 23

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
markkaylor: Sure sounds good, not sure about the name though 🤔. I think I prefer the notion of `observed` since they are designed to be used with the intersection observer.

remidej: yes observed is better. But actually if we end up abstracting the logic further I'm not sure it will be needed since the component code will be a lot shorter

remidej:
Thanks for highlighting the focus & separator issues. For the focus I should be able to fix it here. For the separator I think we should export Radix's DropdownMenu.Separator from the DS. So I'll open an DS PR to fix this and the icon issue, which is that startIcon and endIcon are ignored (unless you make the menu item a link, since the implementation make it inherit a different component)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -496,23 +524,170 @@ const LinkButton = ({ disabled }: { disabled: boolean }) => {
...
+        paddingLeft={0}
+        paddingRight={0}
+        ref={containerRef}
+        style={{ visibility: hasHiddenItems ? 'visible' : 'hidden' }}
+      >
+        <IconButton
+          variant="ghost"
+          label={formatMessage({ id: 'global.more', defaultMessage: 'More' })}
+          tag="span"
+        >
```
markkaylor, line 643: I think we will need the disabled state, and unfortunately I think you have to add it to both Menu.Trigger _and_ IconButotn

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -431,6 +437,21 @@ const ListButton = ({ block, format }: ListButtonProps) => {
     }
   };
 
+  if (location === 'menu') {
+    const Icon = block.icon;
+
+    return (
+      <StyledMenuItem
+        onClick={() => toggleList(format)}
```
markkaylor, line 445: Heads up Menu.Item should use onSelect. onClick will still fire with disabled=true

markkaylor (approved):
Really nice work Rémi 👏 👏 

I tested it out, all looks good:

PR #22763, opened 2025-01-30
fix: class Plugin is not immerable

markkaylor:
### What does it do?

Make the Plugin class immerable 

https://immerjs.github.io/immer/complex-objects/

### Why is it needed?

It fixes an infinite loading state

### How to test it?

The bug is only reproducible in production. It can be reproduced on strapi cloud.

The work here is published on an experimental: `0.0.0-experimental.a4df7ad5ff6a4366442059f7c067a0424ba1e2f9`

Deploy a project to strapi cloud to reproduce the bug
Login, you should be stuck with infinite loading

Update your project with the experimental and redploy
Login, the homepage should load

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/21811
https://github.com/strapi/strapi/pull/22402

### DIFF
--- a/packages/core/admin/admin/src/components/PluginsInitializer.tsx
+++ b/packages/core/admin/admin/src/components/PluginsInitializer.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 
 import { produce } from 'immer';
-import set from 'lodash/set';
 
 import { Page } from '../components/PageHelpers';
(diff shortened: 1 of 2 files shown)


remidej (approved):
I didn't know immerability was a thing 😳 looks like a good fix!

I haven't tested, I guess the best way is to way for users feedback in the original issue

simotae14 (approved):
Wow thanks a lot for this PR Mark

PR #22767, opened 2025-01-30
chore(preview): prepare layout for side editor

markkaylor:
### What does it do?

- Setup future flag
- Split the screen so half the viewport is the space for the side editor and the other half the preview content
- Add the new header layout

Still a weird thing in the e2e tests where its using the future flag for the space for the side editor but not for the header 🤔 

### How to test it?

The future flag should be set to true by default in examples/getstarted now
Go to a preview page
You should see the new layout:
Screenshot 2025-01-30 at 17 18 59

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
@@ -57,7 +57,7 @@ const ClosePreviewButton = () => {
 
   return (
     <IconButton
-      style={{ border: 'none' }}
+      variant="ghost"
       tag={Link}
       relative="path"
       to={toWithFallback}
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -57,6 +57,7 @@ const ClosePreviewButton = () => {
 
   return (
     <IconButton
+      style={{ border: 'none' }}
```
markkaylor, line 60: I thought it was overkill to created a styled component for this, but happy to do it if y'all think it's better

remidej: I don't mind a quick style tag here and there. But have you tried variant="ghost"?

markkaylor: Yup good call thanks! That's me just being a dumb dumb

remidej (approved):
Love this unstable export pattern

simotae14 (approved):
Thanks for your PR, I took the inspiration for my Pr where I setup the future flag for Relations on the fly

PR #22776, opened 2025-01-31
chore: fix missing dependencies and update outdated ones

Convly:
### What does it do?

- Upgrade TypeScript across all packages to the latest version (5.7.3)
- Add missing (needed) dependencies to various packages
- Security upgrade on various dependencies
- Remove outdated and unused example project

### Why is it needed?

Cleanup technical debt around dependencies’ management

### How to test it?

- `yarn install` should not list any missing dependency
- tests should still pass
- everything else should work as usual


innerdvations:
Build isn't building :(

I'll take a closer look monday, but for now, do we have any better alternatives to all the `any` being added? I understand we probably can't add a real type guard because we don't want to cause a breaking change if it throws on some unexpected data, so maybe not, but it always makes me uncomfortable to add those.

Convly:
> Build isn't building :(

Yeah I'm currently looking why

> but for now, do we have any better alternatives to all the `any` being added

I hesitated between fixing everything by adding better types or just adding any and delegating that to another PR

The initial scope was not to fix the types but rather to upgrade the dependencies hence the decision to add any rather than fixing

I plan on working on all those any and types soon

innerdvations:
Yeah, we should break this down into only fixing the typescript version and missing dependencies, I think there are several issues with upgrading the deprecated ones that will probably need some additional refactoring

Convly:
Closing for the time being, I don't have the time necessary to fix all of this at the same time, will open incremental PRs instead

PR #22248, opened 2024-11-22
feat: trunk tests

Marc-Roig:
### What does it do?

Describe the technical changes you did.

### Why is it needed?

Describe the issue you are solving.

### How to test it?

Provide information about the environment and the path to verify the behaviour.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #22801, opened 2025-02-04
chore(tooling): add Trunk Integration to CE e2e tests

samgutentag:
### What does it do?

Extension of work from PR #22248 This adds uploads to Trunk for E2E tests on Community Edition, rather than only for the Enterprise Edition.


innerdvations (approved):

Marc-Roig (approved):
this is amazing, thank you for taking care of it 🙌 seems to be a TS issue in develop, we will handle that

PR #22804, opened 2025-02-05
feat(content-manager): make the markdown editor responsive

markkaylor:
### What does it do?

Uses the intersection observer to progressively hide toolbar items in a more menu
Abstracts shared logic to use between markdown and blocks editors

### Why is it needed?

To prepare for the side editor and otf relations

### How to test it?

Go to a content type with the markdown editor
Resize your screen smaller and smaller
Toolbar items should progressively be added to the more menu 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22753

## TODO
- [x] Update Blocks to use new EditorToolbarObserver


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations.tsx
```diff
@@ -308,8 +308,8 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
     return (
       <Flex
         ref={ref}
-        direction="column"
         gap={3}
+        direction="column"
         justifyContent="space-between"
```
markkaylor, line 313: No idea how this happened but I will revert this change

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/tests/Field.test.tsx
```diff
@@ -246,19 +243,19 @@ describe('Wysiwyg render and actions buttons', () => {
...
-    const { user, getByRole } = render();
+  // it('should disabled the more button when editor is on preview mode', async () => {
+  //   const { user, getByRole } = render();
 
-    await user.click(getByRole('button', { name: /Preview/ }));
-    expect(getByRole('button', { name: 'More' })).toHaveAttribute('aria-disabled', 'true');
-  });
+  //   await user.click(getByRole('button', { name: /Preview/ }));
+  //   expect(getByRole('button', { name: 'More' })).toHaveAttribute('aria-disabled', 'true');
+  // });
```
markkaylor, line 251: I will uncomment these

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/WysiwygNav.tsx
```diff
@@ -48,199 +51,425 @@ interface WysiwygNavProps {
...
+          </Menu.Item>
+          <Menu.Item onSelect={() => handleActionClick('Italic', editorRef)} disabled={isDisabled}>
+            <Flex tag="span" gap={2}>
+              <Italic aria-hidden fill="neutral600" />
+              {formatMessage({
+                id: 'components.Blocks.modifiers.italic',
+                defaultMessage: 'Italic',
+              })}
+            </Flex>
+          </Menu.Item>
```
markkaylor, line 183: Menu.Item doesn't have a `startIcon` prop which I think would be very useful. But @remidej blocks will need to do this exact same thing. We could either update the DS to let MenuItem handle a startIcon, create a shared MoreMenutItem component here in the CM, or just handle it like this. wdyt?

remidej: Ideally we'd have that in the design system, I agree. Menu.Items does accept a startIcon already, but only if it's a link, not a button like here. Links and buttons seem to be separate in the DS implementation, which is odd.

If you wanna fix it there, that would be great, otherwise I think this is fine. I don't really like the local shared abstraction though

markkaylor: Ok sounds good I'll take a look in the DS but we can keep moving forward like this.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -609,9 +610,16 @@ interface MoreMenuProps {
...
 }
 
-const MoreMenu = ({ setLastVisibleIndex, hasHiddenItems, rootRef, children }: MoreMenuProps) => {
+export const MoreMenu = ({
+  setLastVisibleIndex,
+  hasHiddenItems,
+  rootRef,
+  children,
+  triggerVariant = 'tertiary',
+}: MoreMenuProps) => {
```
remidej, line 622: in my opinion this should be moved into your EditorToolbarObserver file. Same for ObservedToolbarComponent and the ObservedComponent type

packages/core/content-manager/admin/src/pages/EditView/components/EditorToolbarObserver.tsx
remidej, line 1: just thinking about the name. i don't think there's anything actually specific to editors in that implementation. maybe we could move this up into the features folder so it can be reused anywhere? and maybe call it something like ResponsiveToolbarSection?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
markkaylor: Yes it is in there, I haven't update blocks

packages/core/content-manager/admin/src/pages/EditView/components/EditorToolbarObserver.tsx
markkaylor: > i don't think there's anything actually specific to editors in that implementation. 

I prefixed Editor since the component is designed to be used with Markdown and Blocks _Editors_, not sure if I've misunderstood or we see it differently but IMO the component itself is very specific to these editors. 

> maybe we could move this up into the features folder so it can be reused anywhere? and maybe call it something like ResponsiveToolbarSection?

To be completely honest I don't think I understand the exact intention of the features folder. In my opinion this component isn't built to be used anywhere, it is built to be used with the markdown and blocks editors and I think they are both only being used in the EditView. I'm fine with moving it and renaming it but if we move it to the features folder shouldn't it be organized into a folder that references the features it is intended for (ie Editor)?

markkaylor:
this work was moved to: https://github.com/strapi/strapi/pull/22820

packages/core/content-manager/admin/src/pages/EditView/components/EditorToolbarObserver.tsx
remidej: Very true, let's not optimize too early, we could try to rename + move it up into the features folder when a need for it elsewhere comes up, until then it's fine where it is

PR #22805, opened 2025-02-05
fix: ctb reload state after save & delete

alexandrebodin:
### What does it do?

Ensures that when we reload data we update the modified state for post-creation/deletion states to be reflecting what has been saved

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/DataManagerProvider/DataManagerProvider.tsx
+++ b/packages/core/content-type-builder/admin/src/components/DataManagerProvider/DataManagerProvider.tsx
@@ -306,6 +306,7 @@ const DataManagerProvider = ({ children }: DataManagerProviderProps) => {
     try {
       const requestURL = `/${pluginId}/${endPoint}/${currentUid}`;
       const isTemporary = get(modifiedData, [firstKeyToMainSchema, 'isTemporary'], false);
+
       // eslint-disable-next-line no-alert
       const userConfirm = window.confirm(
         formatMessage({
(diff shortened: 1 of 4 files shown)


innerdvations:
Nice, I think this actually solves several other long-standing issues.

Here's an experimental: 0.0.0-experimental.28b545d58abbc57b054aa474c1d98ecfdfca9f18

innerdvations:
I confirmed by removing a bunch of `.reload` workarounds we had in e2e tests, want me to push them to this branch so we have it tested?

alexandrebodin:
> I confirmed by removing a bunch of `.reload` workarounds we had in e2e tests, want me to push them to this branch so we have it tested?

yeah definitely

simotae14 (dismissed):
I tested the Prism error and now it is fixed, also the highlighting of the Code in the Block Editor works as expected, thanks a lot for the fix @alexandrebodin 🙏🏽

innerdvations (dismissed):
Works for me, you can merge this into it if you want tests: https://github.com/strapi/strapi/pull/22818

MarionLemaire (approved):
Tested on experimental and PR branch on the monorepo today : LGTM

- [x] Reach CTB
- [x] Create a new Content type / Single type
- [x] Update this CT/ST
- [x] Delete this CT/ST
- [x] Check updates are reflected real-time in the sidebar
- [x] Reach CM
- [x] Reach Listview of the new CT/ST
- [x] Create an entry for this CT/ST and check all fields are present
- [x] Create a new component 

(Note : using the monorepo on the PR branch, I faced some random issues such as "useEffect" but can't reproduce, already discussed this with @innerdvations )

innerdvations (approved):

PR #22806, opened 2025-02-05
fix(upload): window title in media library

jorrit:
### What does it do?

The window title is not set when opening the Media Library.

### Why is it needed?

Other Strapi components correctly set the Media Library. Without this change, a tab with the Media Library open shows a wrong title.

### How to test it?

Open the Media Library, look at the browsers window title.

### DIFF
--- a/packages/core/upload/admin/src/pages/App/App.tsx
+++ b/packages/core/upload/admin/src/pages/App/App.tsx
@@ -37,16 +37,12 @@ export const Upload = () => {
   }, [isLoading, isError, config, rawQuery, setQuery]);
 
   if (isLoading) {
-    return (
-      <>
-        <Page.Title>{title}</Page.Title>
-        <Page.Loading />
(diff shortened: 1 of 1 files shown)


markkaylor (approved):
Thanks for the PR, looks good to me!

I just noticed that the `Page.Title` is being set during the loading state. Which is kind of weird and not how we handle it anywhere else. For the sake of consistency do you mind cleaning that up to just return `Page.Loading` without the fragment and `Page.Title`? I would comment a suggestion but I can't review unchanged code 😞 

Screenshot 2025-02-06 at 12 29 28

markkaylor:
> Done. In other places the window title creation is inlined. Here it is stored in a variable first. Would you also like me to change this?

We can leave it as is for now, I'm gonna go ahead and merge this one. Thanks for the PR!

markkaylor (approved):

Issue #22117, opened 2024-11-09
[Strapi 5] Admin panel show message "Upgrade your admin panel" for the latest Strapi release


teinett:
### Required System information

- Node.js version: 20.16.0
- NPM version: 10.8.1
- Strapi version: 5.3.0
- Database: sqlite
- Operating system: macOS Sonoma 14.6.1 
- Is your project Javascript or Typescript: Typescript

### Describe the bug

I just installed the Strapi from the very beginning: `npx create-strapi-app@latest`. 
Strapi version installed: 5.3.0

I go to http://localhost:1337/admin/settings/application-infos and see the message: "Upgrade your admin panel". 

### Steps to reproduce the behavior

1. Install Strapi
2. Go to http://localhost:1337/admin/settings/application-infos
3. Observe the message: "Upgrade your admin panel"

### Expected behavior

It's already the LATEST version, that you just gave me. There is nothing to upgrade. So I expect no to see the message of upgrading.

### Screenshots

<img width="843" alt="2024-11-09 13-26-52" src="https://github.com/user-attachments/assets/ba31e0a8-4871-4834-aaf3-4c0a3e67d734">

NicoHof:
Note: This also happens to Strapi V4!

derrickmehaffy:
tested on v5.4.0 but cannot reproduce: 

![image](https://github.com/user-attachments/assets/01122242-06f6-4406-803c-89a40be34f65)

zjjtm:
![1116A7BD-7B43-4ca9-8C69-CCE5EDD9B804](https://github.com/user-attachments/assets/7e73bb19-9fcc-4039-8c99-e50a2867ce36)
This does not appear in version v5.4.0, but it appears again in v5.4.1

pawelhoros:
This also happen in 5.4.2. 
I think I found out why.

AuthenticatedLayout.tsx fetches data from 'https://api.github.com/repos/strapi/strapi/releases/latest'
and save **tagName** as "**v5.4.2**"

On the same file, we have:
const strapiVersion = packageJSON.version;
and it's "**5.4.2**" - without this prefix "v".

then is goes down:
shouldUpdateStrapi={checkLatestStrapiVersion(strapiVersion, tagName)}

Finally checkLatestStrapiVersion returns lt(currentPackageVersion, latestPublishedVersion);

and this "**v**" in front of tagName gives us this nice badge and "Upgrade your admin panel" link :)

teinett:
[...]

PR #22810, opened 2025-02-05
fix(admin): display correct version on admin notification

jorrit:
### What does it do?

Defers loading package.json until `@strapi/admin` is built in the end application, instead of when preparing for a release.

### Why is it needed?

As explained in bug #22117 strapi always notifies the user that a new version is available. This is caused by the fact that the version of strapi is embedded in the `@strapi/admin` when the package is built prior to increasing the version.

### How to test it?

Create a new release, go to the Strapi admin page. The notification should be gone.

### Related issue(s)/PR(s)

#22117

### DIFF
--- a/packages/core/admin/admin/src/layouts/AuthenticatedLayout.tsx
+++ b/packages/core/admin/admin/src/layouts/AuthenticatedLayout.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import { version as strapiVersion } from '@strapi/admin/package.json';
 import { Box, Flex, SkipToContent } from '@strapi/design-system';
 import { DndProvider } from 'react-dnd';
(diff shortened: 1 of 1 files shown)


innerdvations (approved):
This makes sense, thanks for your time finding the problem!

jorrit:
You're welcome. Is there any way you can confirm that the fix works before publishing the next release?

innerdvations:
I did a somewhat fake test with an experimental build, but considering the version number there won't match up it's not a perfect test.  Once we create the release branch next week, I'll make a beta version release and check for sure.

jorrit:
Thanks, looking forward to it!

lucasboilly:
@innerdvations @jorrit I just created a new instance and it seems like it's fixed!

PR #22811, opened 2025-02-05
fix: removed unexpected padding around accordions

lucasboilly:
### What does it do?

- Removes unexpected padding around the accordions list
- Removed unused background property
- Added a gap between the subtitle and the accordions list

### Why is it needed?

It currently doesn't look good.

### Before

CleanShot 2025-02-05 at 17 49 30@2x

### After

CleanShot 2025-02-05 at 17 51 12@2x

### How to test it?

Go to Settings, API Tokens and check or create a token.


innerdvations (approved):
LGTM!

PR #22820, opened 2025-02-06
feat(content-manager): make markdown editor responsive

markkaylor:
### What does it do?

Uses the intersection observer to progressively hide toolbar items in a more menu
Abstracts shared logic to use between markdown and blocks editors

### Why is it needed?

To prepare for the side editor and otf relations

### How to test it?

Go to a content type with the markdown editor
Resize your screen smaller and smaller
Toolbar items should progressively be added to the more menu


packages/core/content-manager/admin/src/pages/EditView/components/EditorToolbarObserver.tsx
```diff
@@ -0,0 +1,144 @@
...
+  menu: React.ReactNode;
+  key: string;
+}
+
+export const EditorToolbarObserver = ({
+  observedComponents,
+  editor = 'blocks',
+}: {
+  observedComponents: ObservedComponent[];
+  editor: 'blocks' | 'markdown';
```
remidej, line 75: This editor prop is only used for the menu trigger variant right?

If so I have a slight preference for a triggerVariant prop over this. I feel a bit weird about an abstraction needing to specific from where it's being called (although yes that is exactly what I did with my location prop 🙃)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/tests/Field.test.tsx
```diff
@@ -148,82 +145,81 @@ describe('Wysiwyg render and actions buttons', () => {
   it('should render quote markdown when clicking the quote button', async () => {
     const { user, getByText, getByRole } = render();
 
-    await user.click(getByRole('button', { name: 'More' }));
+    // await user.click(getByRole('button', { name: 'More' }));
```
remidej, line 148: should this just be deleted?

remidej (commented):
Tested, this is :chefskiss:

i'm happy to approve just have 2 questions

remidej (approved):
Let's go 🚀🚀🚀

remidej (approved):

PR #22821, opened 2025-02-06
fix(cli): add --no-open option

innerdvations:
### What does it do?

adds the --no-open option to `strapi develop`

### Why is it needed?

`--open` is a commander boolean which means it doesn't accept a value; so there's currently no way to turn it off. `--open=false` throws an error

### How to test it?

- run `strapi develop` and it should open an admin window if you don't already have one open
- and `strapi develop --no-open` and it should not open an admin window at all

### Related issue(s)/PR(s)


alexandrebodin (approved):
LGTM, do you know what happens when you use the 2 flags in the cmd I'm just curious :D

innerdvations:
with `yarn strapi --watch-admin --no-watch-admin` the false value wins regardless of order

alexandrebodin:
are you gonna merge this ?

innerdvations:
no, it doesn't actually work, I think the code to supress it was removed at some point so it turned out not to be trivial. I'll close this for now and come back to it when there's time to investigate

PR #22823, opened 2025-02-06
future: add edit form to preview page

remidej:
### What does it do?

Renders the edit view form on the preview page. aka the side editor

This does NOT handle saving content updates on that form.

### Why is it needed?

Next iteration of preview

### How to test it?

Needs a future flag:

```
// config/features.js
  future: {
    unstablePreviewSideEditor: true,
    unstableRelationsOnTheFly: true,
  },
});
```

Open the preview for an entry that has preview set up. You should see the edit view form for that entry on the left half of the page next to the preview iframe. The form should be editable.

Open that same page while authenticated as a user who has permissions to read but not update that content type. You should see the preview page, but the fields should be disabled.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
@@ -23,9 +23,11 @@ export const ResponsiveGridItem = styled(Grid.Item)`
   }
 `;
 
-interface FormLayoutProps extends Pick<EditLayout, 'layout'> {}
+interface FormLayoutProps extends Pick<EditLayout, 'layout'> {
+  hasBackground?: boolean;
+}
 
-const FormLayout = ({ layout }: FormLayoutProps) => {
+const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
   const { formatMessage } = useIntl();
   const { model } = useDoc();
 
@@ -56,14 +58,13 @@ const FormLayout = ({ layout }: FormLayoutProps) => {
         return (
           <Box
             key={index}
-            hasRadius
-            background="neutral0"
-            shadow="tableShadow"
-            paddingLeft={6}
-            paddingRight={6}
-            paddingTop={6}
-            paddingBottom={6}
-            borderColor="neutral150"
+            {...(!hasBackground && {
+              padding: 6,
+              borderColor: 'neutral150',
+              background: 'neutral0',
+              hasRadius: true,
+              shadow: 'tableShadow',
+            })}
           >
             <Flex direction="column" alignItems="stretch" gap={6}>
               {panel.map((row, gridRowIndex) => (
--- a/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
+++ b/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
@@ -1,6 +1,12 @@
 import * as React from 'react';
 
-import { Page, useQueryParams, useRBAC, createContext } from '@strapi/admin/strapi-admin';
+import {
+  Page,
+  useQueryParams,
+  useRBAC,
+  createContext,
+  Form as FormContext,
+} from '@strapi/admin/strapi-admin';
 import { Box, Flex, FocusTrap, Portal } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
 import { useParams } from 'react-router-dom';
@@ -9,7 +15,7 @@ import { GetPreviewUrl } from '../../../../shared/contracts/preview';
 import { COLLECTION_TYPES } from '../../constants/collections';
 import { DocumentRBAC } from '../../features/DocumentRBAC';
 import { type UseDocument, useDocument } from '../../hooks/useDocument';
-import { useDocumentLayout } from '../../hooks/useDocumentLayout';
+import { type EditLayout, useDocumentLayout } from '../../hooks/useDocumentLayout';
 import { buildValidParams } from '../../utils/api';
 import { PreviewContent, UnstablePreviewContent } from '../components/PreviewContent';
 import { PreviewHeader, UnstablePreviewHeader } from '../components/PreviewHeader';
@@ -27,6 +33,7 @@ interface PreviewContextValue {
   document: NonNullable<ReturnType<UseDocument>['document']>;
   meta: NonNullable<ReturnType<UseDocument>['meta']>;
   schema: NonNullable<ReturnType<UseDocument>['schema']>;
+  layout: EditLayout;
 }
 
 const [PreviewProvider, usePreviewContext] = createContext<PreviewContextValue>('PreviewPage');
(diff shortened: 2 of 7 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -23,9 +23,11 @@ export const ResponsiveGridItem = styled(Grid.Item)`
   }
 `;
 
-interface FormLayoutProps extends Pick<EditLayout, 'layout'> {}
+interface FormLayoutProps extends Pick<EditLayout, 'layout'> {
+  hasBackground?: boolean;
```
remidej, line 27: i'm open to suggestions for better names. the issue is that the form layout is normally rendered on top of a greyish background (for light mode), so it has borders and shadows to separate each layout pane from it. We don't want those styles for preview though, because the layout is rendered on a pure white card.

markkaylor: Maybe a crazy idea but what if instead of a prop we used the path params to determine whether or not it is a the preview page and if it is then adjust the styles

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -193,7 +206,11 @@ const ProtectedPreviewPageImpl = () => {
       zIndex={2}
       background="neutral0"
     >
-      <Page.Protect permissions={permissions}>
+      <Page.Protect
+        permissions={permissions.filter((permission) =>
+          permission.action.includes('explorer.read')
+        )}
+      >
```
markkaylor, line 213: I'm not able to edit the form and I think it's caused by this part. Aren't you filtering out all permissions except read?

markkaylor (commented):
Man so cool to see the side editor!!

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: Are the fields disabled for you? they don't look disabled on your slack screenshot

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -23,9 +23,11 @@ export const ResponsiveGridItem = styled(Grid.Item)`
...
+}
 
-const FormLayout = ({ layout }: FormLayoutProps) => {
+const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
```
simotae14, line 30: I needed the same stuff

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
remidej: I don't think it works because of relations on the fly. We don't want the borders etc on the relation modal, and those modals can be displayed both in the edit view path and the preview path

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
markkaylor: yeah because I adjusted this line to pass all the permissions 🙃 


https://github.com/user-attachments/assets/30c274de-18dc-4e2a-835a-3a91d9efe2c5

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
markkaylor: Bad idea, won't work for relations otf.

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: My thinking behind this: Page.Protect is the thing that prevents the page from being rendered at all. I only want that when users don't have read permissions, hence the filter

Disabling the fields though is done in InputRenderer, which relies on the DocumentRBAC context just below this. I do want to give this one the update permissions data so it can know whether to disable the fields

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
remidej: but again, I'm sure there must be a better name for that prop 😁

simotae14: I need the same stuff for the modal, so for me hasBackground can be the best option because it is the need we have in both the features

simotae14: btw in my case I don't have path params because the relation's modal is inside the edit view page

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
markkaylor: Thanks I understand it better now and I see what you mean, but I think Page.Protect forwards the permissions it received to its children. So DocumenRBAC is only going to end up with read permissions

markkaylor: https://github.com/strapi/strapi/blob/743192159941b6524bfa234b13e9d58b3d111710/packages/core/admin/admin/src/components/PageHelpers.tsx#L237-L241

remidej: ah you're right, I didn't realize it was using the render prop. will fix now

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
simotae14: btw to avoid regression errors I think we need to set as default value hasBackground to true

remidej: what we have in develop (so what the edit view uses) is the equivalent of hasBackground=false here 🤔 because on the edit view it's not rendered on top of a white background. but this shows that the name is pretty bad because it's confusing

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -211,11 +211,9 @@ const ProtectedPreviewPageImpl = () => {
           permission.action.includes('explorer.read')
         )}
       >
-        {({ permissions }) => (
-          <DocumentRBAC permissions={permissions}>
-            <PreviewPage />
-          </DocumentRBAC>
-        )}
+        <DocumentRBAC permissions={permissions}>
```
markkaylor, line 214: I'm cool with this idea, but a bit bothered that the permissions are not handled the same here vs the EditViewPage. What are your thoughts on that. I'm cool to move forward as is. Just wondering if we should leave a comment to update the EditViewPage, update the EditViewPage now, or decide that it's fine if they handle it differently.

remidej: tbh i didn't check the edit view page implementation, i did what made sense to me here. i'm gonna look into how that one works now

remidej: update: there's actually an issue with Page.Protect that the edit view page relies on, I'll bring this up at the frontend sync

markkaylor (approved):

Issue #22607, opened 2025-01-09
Problem with order when updating a record linked through a has-many relationship


FrancescoMeneghini:
### Required System information

- Node.js version: 20.16.0
- NPM version: 10.8.1
- Strapi version: 5.7.0 and previows
- Database: postgresql

### Describe the bug

In my case, I have two collections: faq_categories and faqs. faq_categories has a has_many relationship with faqs. When I update an faq linked to an faq_category and publish it, the record is destroyed and a new one is created. However, in the join table, the faq_order value for the new record changes it moves to the end of the list.

### Steps to reproduce the behavior

1. Go to related record (ex: faq foo)
2. Change a field save and publish (ex: answer from 'foo' to 'updated foo')
3. Call the APIs to check the order in which they are returned
4. See that the position are changed

### Expected behavior

I expect that updating a record does not change its position and that the sorting remains the same.

### Screenshots

![image](https://github.com/user-attachments/assets/621fc4e6-6f4f-4961-b338-88a96078737f)

![image](https://github.com/user-attachments/assets/a5e85300-c97c-4926-b23a-01202e4284fa)

![image](https://github.com/user-attachments/assets/5939d366-2636-4077-a331-6a9974fd1ed9)

![image](https://github.com/user-attachments/assets/0c66c1c6-2c43-4dfb-9956-7edc2da3f736)

![image](https://github.com/user-attachments/assets/804cb99a-14de-4efb-9f15-cc8adb17841e)

![image](https://github.com/user-attachments/assets/cef7fabd-e6e2-4d86-a378-53c38a837aaa)

laurenskling:
That is what felt so odd about my test project! I tested it and:
I can confirm that in published mode, the relation order is ignored and the last edited entry is sent to the end of the array. 
In `status=draft` the order is correct.

If you ask me, this is a pretty high prio bug..

obafemitayor:
@lucasboilly @derrickmehaffy  I am a first time contributor. Is it ok for me to work on this issue?

ignissak:
Can replicate. This is a very annoying bug, I had to create a workaround by adding another field that forcibly specifies order and sort by that field. This is not applicable to every scenario because I fetch REST with multiple level nesting where I cannot sort properly by nested fields.

lucasboilly:
Hello everyone, and thanks for raising the issue!

@obafemitayor of course, you can directly submit a PR and our engineers are gonna review it :)

laurenskling:
[...]

Issue #21828, opened 2024-10-14
Internal Server Error in Content Manager - Dropdown add new item, change order and hit Save


cslegany-synerinsoft:
### Required System information

- Node.js version: v20.11.0
- NPM version: 10.2.4
- Strapi version: 5.0.6
- Database: MySQL (doesn't matter)
- Operating system: Windows
- Is your project Javascript or Typescript: Typescript

### Describe the bug
Dropdown throws and Internal Server Error if you select and add a new item, reorder items and save it. No error is thrown if you just add a new item and save it without changing the order.

### Steps to reproduce the behavior

1. Go to Content Manager
2. We've added a Single Type called Highlight Settings which contains a component called as featured_news
3. Go to the featured_news, select right_side_articles which is a dropdown listing articles (i.e. a one-to-many relation to articles)
4. Click on Add relation and add a new article
5. Change the order of the articles by using drag-drop
6. Click Save button
7. Error is thrown and Strapi won't be able to save the changes you made (new item added and order changed)

### Expected behavior
Adding a new item to the dropdown list and reordering works together properly.

### Screenshots
![dropdown_eror](https://github.com/user-attachments/assets/67208da0-c731-4632-a916-63b29153fe07)

### Code snippets
Highlight Settings json (just relevant information is kept)
`{
  "kind": "singleType",
  "collectionName": "highlight_settings",
  "info": {
    "singularName": "highlight-setting",
    "pluralName": "highlight-settings",
    "displayName": "Highlight Settings",
    "description": ""
  },
  "options": {
    "draftAndPublish": true
  },
  "pluginOptions": {},
  "attributes": {
    "featured_news": {
      "type": "component",
      "repeatable": false,
      "component": "highlighted.highlighted-articles"
    },
    "name": {
      "type": "string"
    }
  }
}`

featured_news
`{
  "collectionName": "components_highlighted_highlighted_articles",
  "info": {
    "displayName": "Highlighted Articles",
    "icon": "dashboard"
  },
  "options": {},
  "attributes": {
    "right_side_articles": {
      "type": "relation",
      "relation": "oneToMany",
      "target": "api::article.article"
    }
  }
}`

cslegany-synerinsoft:
Is there any improvement on this issue? It would be important for us to get this fixed soon.

[...]

PR #22834, opened 2025-02-07
fix: inverse relation ordering

Marc-Roig:
### What does it do?

This PR fixes a bug in bidirectional relations where the order of relations was being lost when entries were published. The fix involves:

1. Capturing the original relation order from both sides before an entry's state changes
2. After the entry is republished and relations are recreated, restoring the correct order in the join tables
3. Using transaction-safe database operations to ensure data consistency

TODO:
- Add api tests

### Why is it needed?

In Strapi, when an entry is published, the following occurs:
1. The old published entry is deleted
2. A new entry is created with all its relations

This process was breaking the relation order from the inverse perspective. For example:
- If Entry A had relations to X, Y, Z in that order
- And X had Entry A as its 2nd relation, Y as 1st, and Z as 3rd
- After publishing Entry A, while A→X→Y→Z order was preserved
- From X's perspective, the relation to Entry A would incorrectly appear last

This was particularly noticeable in the admin UI and API responses where related entries would suddenly change their display order after publishing, even though no reordering was intended.

### How to test it?

1. Create two collections with a bidirectional relation (e.g., FAQs and FAQ Categories)
2. Create an FAQ category and multiple FAQs
3. Order the FAQs within the category
4. Order the categories within each FAQ
5. Update and publish an FAQ
6. Verify that:
   - The order of FAQs in the category remains unchanged
   - The order of categories in the FAQ remains unchanged

### Related issue(s)/PR(s)

- Fixes: https://github.com/strapi/strapi/issues/22607


packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
```diff
@@ -0,0 +1,174 @@
...
+ *      └──(3)→ Z            └──(3)→ Z
+ *
+ *   X's perspective:         Y's perspective:         Z's perspective:
+ *      └──(3)→ Entry A         └──(3)→ Entry A         └──(3)→ Entry A
+ *                           (all relations appear last in order)
+ * ```
+ *
+ * This module preserves the original ordering from both perspectives by:
+ * 1. Capturing the relation order before the entry state changes
+ * 2. Restoring this order after the new relations are created
```
markkaylor, line 50: Does this belong here or in the contributor docs?

markkaylor (commented):
Tested locally, looks like it's working. Code looks good to me ✨

packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
Marc-Roig: open for any option, I personally prefer having this collocated with the code as the only moment a user will read this is when modifying the code

alexandrebodin: Why not just include the JSdoc documentation into the contributor doc so all code level documentation is automatically added in the contributor doc ? 

If the comment explains how the function works specificaly it should stay in the code. if it explains how the overall sync system works this should be in a .md file

alexandrebodin (commented):
The logic looks sane overall. Howeer have you considered managing this directly during the relation creation by allowing passing the inv order info somhow ? Don't we already support some of this in the DB logic ?

alexandrebodin:
Exprimental release https://github.com/strapi/strapi/actions/runs/13283073601

alexandrebodin (approved):
Playing around I got this error once randomly and could not reproduce by clicking the save or publish buttons a second time so it looks like a race condition somewhere that send the wrong info to the BE. In any case worth double checking if there is not a piece of code where we pass documentId to an id where clause or sth like that 🙏 

I played around with some combinations of D&P & i18N and the ordering was well maintained 👍  Some strange flickering that still persist in the relation UI reloading but just an improvement to add to the list

Screenshot 2025-02-13 at 08 39 24

Marc-Roig:
@alexandrebodin I will check but I remember seeing that issue when adding multiple relations and reordering one before another that was already created.

alexandrebodin:
> @alexandrebodin I will check but I remember seeing that issue when adding multiple relations and reordering one before another that was already created.

this rings a bell, I added and reordered at the same time

laurenskling:
> @alexandrebodin I will check but I remember seeing that issue when adding multiple relations and reordering one before another that was already created.

https://github.com/strapi/strapi/issues/21828

laurenskling:
Is there a release I can test?

laurenskling:
Publishing a selected relation will put it at the end of the array, not at the chosen order.

laurenskling:
> Might be worth also testing relations order deeply nested in components too as a user mentioned it yesterday in office hours.

I've created a Many relation within a component. 
Publishing a selected relation will not show up in the API response. 
Only when I reorder, it starts showing up in the result.
The order of the response does become out of sync with the admin UI. As if that previously draft item has a broken index number.

Marc-Roig:
>Publishing a selected relation will put it at the end of the array, not at the chosen order.
will check that one

>I've created a Many relation within a component.
Publishing a selected relation will not show up in the API response.
Only when I reorder, it starts showing up in the result.
The order of the response does become out of sync with the admin UI. As if that previously draft item has a broken index number.

I will work on this one in a separate issue, for now I will merge this one as it seems tof ix the original issue.

alexandrebodin:
@Marc-Roig should we merge ?

laurenskling:
> Publishing a selected relation will put it at the end of the array, not at the chosen order.

This is still a thing in 5.11 / 5.12.0-beta.0 🥲

alexandrebodin:
@laurenskling can you create a specific issue for it cc @oiorain for tracking during our Relations focused week

Marc-Roig:
@laurenskling could you specify the exact schema setup you have in that issue? I was not able to replicate it

laurenskling:
I've created the issue with a repo and a video

PR #22852, opened 2025-02-10
fix: unable to customize cors settings in graphql plugin

dmaklygin:
### What does it do?

This PR reinstates the ability to customize CORS settings specifically for the graphql plugin in Strapi. Previously, configuring CORS under the apolloServer attribute inside the graphql plugin settings was effective. However, this functionality was removed in [this commit](https://github.com/strapi/strapi/commit/13a2f8b2468b7fa0ac27fd4f08ac47ad4e36ee31#diff-c689097227a8c063449c5940e4f97b13f193e20f6f2c2b38b2b60a29b7b8b601L127).

This PR restores the ability to define CORS settings directly in the graphql plugin configuration.

### Why is it needed?

Setting CORS in middlewares.js is not sufficient because @koa/cors is executed twice:

- Globally – as part of the application's middleware stack.
- Specifically for the GraphQL plugin – when processing GraphQL requests.

The second invocation, which occurs within the graphql plugin, does not inherit the CORS settings from middlewares.js. This leads to unintended behavior where custom GraphQL CORS configurations are ignored.

### How to test it?

Specify two configs:
1. middlewares.js

```
{
    name: "strapi::cors",
    config: {
      maxAge: 31536000,
      origin: ["http://my-domain.tt"],
      methods: ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
    },
  },
```

2. in the plugins.js

```
{
...

graphql: {
    config: {
      endpoint: '/graphql',
      apolloServer: {
        cors: {
          origin: ["http://my-domain.tt"],
          methods: ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
        },
      },
    },
  },
...

}
```

try to perform a request from the application which is based on `my-domain-2.tt`

Expected:
Application should return a header:
```
access-control-allow-origin: http://my-domain.tt
```

Actual:
Application returns header:
```
access-control-allow-origin: *
```

### Related issue(s)/PR(s)


hanpaine:
Hey @dmaklygin thanks for your PR - we'll take a look soon 🙏 🚀

packages/plugins/graphql/server/src/bootstrap.ts
```diff
@@ -167,7 +168,7 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
 
   // add cors middleware
   if (cors) {
-    handler.push(cors());
+    handler.push(cors(serverConfig.cors));
```
innerdvations, line 208: I don't think we listed this as a breaking change from Strapi 4-> Strapi 5, we just... accidentally removed it completely 🤦

So in order to keep this working as expected for someone upgrading from Strapi 4 (which would have worked if not for this bug), can we keep the support for booleans?

It would be something like this:

```
  if (serverConfig.cors === false) {
    // Explicitly disabled - don't add middleware
  } else if (serverConfig.cors === undefined || serverConfig.cors === true) {
    // enable with defaults (backwards compatible)
    handler.push(cors());
  else {
    handler.push(serverConfig.cors);
  }
```

Bassel17 (approved):
Thanks for the contribution, it looks good now !

PR #22858, opened 2025-02-11
feat(relations-on-the-fly): add the Relation Modal

simotae14:
### What does it do?

 - Opens the Edit Modal when you click on a Relation
 - inside the modal you can open nested relations and show them in the modal

### Why is it needed?

Retrieve the Modal data using the documentId and Open the modal instead of showing a link to the relation edit view

### How to test it?

The future flag should be set to true by default in examples/getstarted now
Go to a document with a relation field and add a new relation
Then if you click on the relations it opens a modal with all the relation's fields

### Related issue(s)/PR(s)

CS-1304


packages/core/content-manager/admin/src/features/ModelRBAC.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * @internal This component is not meant to be used outside of the Content Manager plugin.
+ * It depends on knowing the model of the content-type using an argument passed.
+ * If you do use the hook outside of the context, we default to `false` for all actions.
+ *
+ * It then creates an list of `can{Action}` that are passed to the context for consumption
+ * within the app to enforce RBAC.
+ */
+const ModelRBAC = ({ children, permissions, model }: ModelRBACProps) => {
```
markkaylor, line 64: As far as I can tell the only real difference between DocumentRBAC and this file is the additon of model prop and removal of `slug` from useParams.

Couldn't you avoid duplicating all this code by passing model as a prop, make it optional, if it's not passed fall back to the slug from useParams?

```tsx
const DocumentRBAC = ({ children, permissions, model }: DocumentRBACProps) => {
  const { slug: modelSlug } = useParams<{ slug: string }>();

  const slug = model ?? modelSlug

  if (!slug) {
    throw new Error('Cannot find the slug param in the URL');
  }
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/RelationModal.tsx
markkaylor, line 1: I think we should create a Relations fold for this and Relations.tsx

packages/core/content-manager/admin/src/pages/EditView/components/FormLayoutModal.tsx
```diff
@@ -0,0 +1,108 @@
+/*
+This one is a copy of the FormLayout component.
+But instead of using the useDoc to retrieve the data from the url we use the
+useRelationModalContext to get the current relation model. It is used inside the Relation on the fly modal.
+*/
```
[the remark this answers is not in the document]
markkaylor: >Why not modify the original FormLayout component to take the model as a prop, so it doesn't call useDoc anymore?

Yes I think we should try this or explore other options to avoid the code duplication.

packages/core/content-manager/admin/src/features/ModelRBAC.tsx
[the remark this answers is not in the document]
markkaylor: I agree, could you please tell us more in your PR description what you did and why?

markkaylor (commented):
I think we are duplicating a lot of code for _very_ small changes. I'm not sure I agree with this approach. I'd rather find a way to make the concerned components more flexible to handle this case. If you want we can discuss that in a PP session with @remidej

packages/core/content-manager/admin/src/features/ModelRBAC.tsx
remidej: i don't think that specific solution works because on the edit view we'll always get a param match from the URL, but it won't be the right one when dealing with a relation to another content type in a modal

packages/core/content-manager/admin/src/pages/EditView/components/FormLayoutModal.tsx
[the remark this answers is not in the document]
simotae14: Hey, thanks for the suggestions, the PR is in draft and I duplicated the code only to have a working initial solution, not as a final solution. I wanted to understand from you first if the solution works as expected

packages/core/content-manager/admin/src/pages/EditView/components/InputRendererModal.tsx
```diff
@@ -0,0 +1,277 @@
+/*
+This one is a copy of the InputRenderer component.
+But instead of using the useDoc to retrieve the data from the url we use the
+useRelationModalContext to get the current relation model, id, document and collectionType. It is used inside the Relation on the fly modal.
+*/
```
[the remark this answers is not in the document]
simotae14: Hey, thanks for the suggestions, the PR is in draft and I duplicated the code only to have a working initial solution, not as a final solution. I wanted to understand from you first if the solution works as expected

packages/core/content-manager/admin/src/features/ModelRBAC.tsx
simotae14: Hey, thanks for the suggestions, the PR is in draft and I duplicated the code only to have a working initial solution, not as a final solution. I wanted to understand from you first if the solution works as expected

packages/core/content-manager/admin/src/pages/EditView/components/InputRendererModal.tsx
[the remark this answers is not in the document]
simotae14: I removed this component and created a wrapper on top of InputRender to pass props from the RelationModalContext

packages/core/content-manager/admin/src/features/ModelRBAC.tsx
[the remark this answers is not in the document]
simotae14: Removed

packages/core/content-manager/admin/src/pages/EditView/components/FormLayoutModal.tsx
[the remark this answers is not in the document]
simotae14: I removed the FormLayoutModal component and added the logic in the FormLayout component

packages/core/content-manager/admin/src/features/ModelRBAC.tsx
simotae14: I removed the ModelRBAC and used the DocumentRBAC instead

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -15,21 +15,29 @@ export const ResponsiveGridRoot = styled(Grid.Root)`
...
+        grid-column: span 12;
+      `;
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {
   hasBackground?: boolean;
+  model?: string;
 }
 
-const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
+const FormLayout = ({ layout, hasBackground = true, model: modelContext }: FormLayoutProps) => {
```
remidej, line 37: Not sure why name it context if it's from a prop?

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -15,21 +15,29 @@ export const ResponsiveGridRoot = styled(Grid.Root)`
...
   const { formatMessage } = useIntl();
-  const { model } = useDoc();
+  const { model: modelHook } = useDoc();
+  const model = modelContext || modelHook;
```
remidej, line 40: To be sure I understand what's going on. For a relation modal, useDoc will get a match, but not the one from the relation modal, instead a match from the preview or edit view page underneath.

But if we have the model passed in the props, which is the case for relations modal, we use that in priority.

Is that right? It would be nice to have a comment somewhere to explain why we're doing this

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -58,7 +70,7 @@ const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
         return (
           <Box
             key={index}
-            {...(!hasBackground && {
+            {...(hasBackground && {
```
remidej, line 73: again i really wish we found a better name for this, as neither true or false is really clear

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -35,8 +45,54 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'>;
...
+      contextDocument={document}
+      contextCollectionType={collectionType}
+      contextModel={model}
+      changeCurrentRelation={changeCurrentRelation}
+      isModalOpen={isModalOpen}
+      contextComponents={components}
+      {...props}
+    />
+  );
+};
```
markkaylor, line 77: This jumps out at me as strange. I feel like an issue with this current implementation is we have a context provider that is deep down the tree (in the modal). Wouldn't it be better if we had the relation context higher up the tree? Could we design something that provides the context for editing relations from the root entry down. That way no matter where we are we have access to whatever relation is being edited?

That's just an idea, not saying you should try to implement that right now. Happy to have a PP session with you and @remidej to come up with an API to support this feature.

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -35,8 +45,54 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'>;
...
+  contextId,
+  contextDocument,
+  contextCollectionType,
+  contextModel,
+  changeCurrentRelation,
+  isModalOpen,
+  contextComponents,
+  ...props
+}: InputRendererProps) => {
+  const { id: hookId, document: hookDocument, collectionType: hookCollectionType } = useDoc();
```
remidej, line 91: I don't think hook is the right diferenciator here. Maybe url? urlId, urlDocument, etc.

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -86,7 +98,11 @@ const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
                         direction="column"
                         alignItems="stretch"
                       >
-                        <InputRenderer {...fieldWithTranslatedLabel} />
+                        {modelContext ? (
+                          <InputRendererWithContext {...fieldWithTranslatedLabel} />
+                        ) : (
+                          <InputRenderer {...fieldWithTranslatedLabel} />
+                        )}
```
remidej, line 105: What's stopping us from always rendering the component that has the context?

For the "root" relation (not in a modal) I guess the context could be initialized with the values from the URL (useDoc stuff)

remidej (commented):
I'm really liking the changes :)

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
remidej: I agree that the Thing + ThingWithContext pattern is odd, I'd rather unify everything. Moving the context at the input level rather than the modal would be really nice indeed. And it's probably simpler to debug since there would be a single context rather than nested ones

packages/core/content-manager/admin/src/features/DocumentRBAC.tsx
```diff
@@ -54,35 +54,39 @@ const [DocumentRBACProvider, useDocumentRBAC] = createContext<DocumentRBACContex
...
+const DocumentRBAC = ({ children, permissions, model }: DocumentRBACProps) => {
   const { slug } = useParams<{ slug: string }>();
 
-  if (!slug) {
-    throw new Error('Cannot find the slug param in the URL');
+  if (!slug && !model) {
+    throw new Error('Cannot find the slug param in the URL or the model prop is not provided.');
   }
+
+  const documentIdentifier = model ?? slug;
```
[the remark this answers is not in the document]
markkaylor: Personally I would go with `contentTypeUid` since the model and the slug are both referring to that. Which is super confusing.

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
markkaylor: I agree. I think when the values are coming from `useDoc` we know it is the `root` document. And likewise if it's passed as a prop then it is a _related_ document. Maybe the variable names could delineate this difference?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/RelationModal.test.tsx
```diff
@@ -0,0 +1,213 @@
+import { render, screen } from '@tests/utils';
+
+import { RelationModal } from '../RelationModal';
+
+jest.mock('../../../../../../hooks/useDocument', () => ({
```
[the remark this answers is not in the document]
simotae14: I tried to do it but I have the same problem we have with Relations test https://github.com/strapi/strapi/blob/02c84481842cb3cd37b9f5242e376bbb18e921b7/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/Relations.test.tsx#L59 so I tried this workaround

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -15,21 +15,29 @@ export const ResponsiveGridRoot = styled(Grid.Root)`
...
+  process.env.NODE_ENV !== 'test'
+    ? styled(Grid.Item)<{ col: number }>`
+        grid-column: span 12;
+        @container (min-width: ${RESPONSIVE_CONTAINER_BREAKPOINTS.sm}) {
+          ${({ col }) => col && `grid-column: span ${col};`}
+        }
+      `
+    : styled(Grid.Item)<{ col: number }>`
+        grid-column: span 12;
+      `;
```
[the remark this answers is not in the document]
simotae14: I understand your point but the line gave me problems was this one https://github.com/strapi/strapi/blob/02c84481842cb3cd37b9f5242e376bbb18e921b7/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx#L25 I tried different solutions but this one is the only one working but yeah I can try to reiterate on this

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
simotae14: so your suggestion is to move everything in the Relations file?

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
simotae14: yes that's right, I can add a comment, sure

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
remidej: maybe hasSinglePane? the biggest difference is that we're not splitting the different layout panes, they're all in the same box with a common background

simotae14: ah ok thanks

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
simotae14: To give you some context, here is a representation of the flow of Components for the Relation field
`InputRenderer 
   -> UnstableRelationsInput aka UnstableRelationsField
	-> RelationsInput
	-> UnstableRelationsList
		-> UnstableListItem
			-> RelationModal
				-> RelationModalBody (CTX)
					-> FormLayout (usa il model)
						-> InputRendererWithContext
							-> InputRenderer
						-> InputRenderer`

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -116,11 +117,25 @@ interface Relation extends Pick<RelationResult, 'documentId' | 'id' | 'locale' |
   label: string;
   position?: RelationPosition;
   __temp_key__: string;
+  apiData?: {
+    documentId: RelationResult['documentId'];
+    id: RelationResult['id'];
+    locale?: RelationResult['locale'];
+    position: RelationPosition;
+  };
```
markkaylor, line 125: Why is this added?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -116,11 +117,25 @@ interface Relation extends Pick<RelationResult, 'documentId' | 'id' | 'locale' |
...
   extends Omit<Extract<EditFieldLayout, { type: 'relation' }>, 'size' | 'hint'>,
-    Pick<InputProps, 'hint'> {}
+    Pick<InputProps, 'hint'> {
+  documentModel?: string;
+  document?: ReturnType<UseDocument>['document'];
+  setCurrentDocument?: (newDocument: {
+    documentId: string;
+    model: string;
+    collectionType: string;
+  }) => void;
```
markkaylor, line 137: I don't think we use these props anymore, or we should since we can get the values from context

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -15,21 +15,32 @@ export const ResponsiveGridRoot = styled(Grid.Root)`
...
```
markkaylor, line 30: I'm really not a fan of this, can you at least put a TODO to fix it.

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
```diff
@@ -15,21 +15,32 @@ export const ResponsiveGridRoot = styled(Grid.Root)`
...
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {
   hasBackground?: boolean;
+  model?: string;
 }
 
-const FormLayout = ({ layout, hasBackground = false }: FormLayoutProps) => {
+/**
+ * Component that renders the form layout, in case the model is passed as a prompt we give priority to it instead of using the one from useDoc, which is the default behavior. This is useful for Relation modal where the model is not the one in the url but the one from the relation.
+ */
```
markkaylor, line 39: I think this comment is outdated now that we are using useDocumentContext

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
markkaylor: Actually on my branch I already removed the unnecessary props

markkaylor:
Ignore my comments in Relations.tsx my PR will merge the refactor

tests/e2e/tests/content-manager/relations.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    });
+    test.describe('Edit Modal', () => {
+      test('as a user I want to open a relation modal inside a collection', async ({ page }) => {
+        await page.getByRole('link', { name: 'Content Manager' }).click();
+        await page.getByRole('link', { name: 'Author' }).click();
+        await page.getByRole('gridcell', { name: 'Ted Lasso' }).click();
+
+        await expect(page.getByRole('heading', { name: 'Ted Lasso' })).toBeVisible();
+
+        await page.getByRole('button', { name: 'Pourquoi je préfère le' }).click();
```
markkaylor, line 22: why is this check in french but the rest are in english?

tests/e2e/tests/content-manager/relations.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { test, expect } from '@playwright/test';
+import { login } from '../../utils/login';
+import { resetDatabaseAndImportDataFromPath } from '../../utils/dts-import';
+import { describeOnCondition } from '../../utils/shared';
+
+describeOnCondition(process.env.STRAPI_FEATURES_UNSTABLE_RELATIONS_ON_THE_FLY === 'true')(
+  'Relations on the fly tests',
```
markkaylor, line 7: I would prefix the unstable so it's clear the test and feature are unstable

```suggestion
  'Unstable Relations on the fly',
```

tests/e2e/tests/content-manager/relations.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  () => {
+    test.beforeEach(async ({ page }) => {
```
markkaylor, line 9: I don't think we need a nested describe

tests/e2e/tests/content-manager/relations.spec.ts
markkaylor: oh it's the relation

packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -13,6 +13,7 @@ import {
   FormErrors,
   getYupValidationErrors,
   useForm,
+  createContext,
```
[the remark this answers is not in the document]
simotae14: nope, I am gonna remove it

tests/e2e/tests/content-manager/relations.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      await resetDatabaseAndImportDataFromPath('with-admin.tar');
+      await page.goto('/admin');
+      await login({ page });
+    });
+    test.describe('Edit Modal', () => {
+      test('as a user I want to open a relation modal inside a collection', async ({ page }) => {
+        await page.getByRole('link', { name: 'Content Manager' }).click();
+        await page.getByRole('link', { name: 'Author' }).click();
+        await page.getByRole('gridcell', { name: 'Ted Lasso' }).click();
```
markkaylor, line 18: I'd recommend clickAndWait util for all of these

tests/e2e/tests/content-manager/relations.spec.ts
markkaylor: You can probably use clickAndWait here too

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
simotae14: we need the apiData in this handler handleChangeModalContent https://github.com/strapi/strapi/blob/18663acd7bf8a6b72833694b0257609103a7a4c5/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx#L1472

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
simotae14: ok I am gonna remove them

packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
simotae14: I can remove this fix and add a todo to find a solution to remove warnings when we launch the FE unit tests, is it ok for you?

markkaylor: Wait are they warnings or errors? If it doesn't make the test fail then I'd say remove this

markkaylor (approved):

PR #22869, opened 2025-02-12
fix: issue in 5.10.0

alexandrebodin:
### What does it do?

- Fix wrong dep version breaking validation instance of checks
- Fix invalid resource path after rollup build chunking


Marc-Roig (approved):

innerdvations:
Noticed this experimental: 0.0.0-experimental.dd200315442827f7887dd7c49ae0d196b801865d

Testing it now

innerdvations (approved):

PR #22878, opened 2025-02-13
fix(core-utils): validate child data correctly

### LINKED ISSUE #22680: Content PUT API not supporting child component upserts
### Node Version

22.13.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.8.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/lucasheim/strapi-v5-upsert-bug

### Bug Description

When you have a Collection type with an attribute as a Repeatable Component, there's no way of updating the data on the components without creating a new one.

I have a very small reproduction repo here: https://github.com/lucasheim/strapi-v5-upsert-bug

There, I have one collection called `Parent Collection`, it has two properties:
- `name` - Short Text
- `children` - Component (repeatable)

The component used above is called `Child Component` and it contains one single property:
- `childName` - Short Text

### Steps to Reproduce

1. Create a new entry for the collection in the UI. It's possible to fetch it through the API

Image

2. Add a new component to the `children` property via API

Image

3. Fetch again the data through API, it should be updated with the new child

Image

4. Try to update the existing child entry created by providing the ID on the API call

Image

Error comes up: `Invalid key id`

### Expected Behavior

It should allow the `id` property to be passed when updating a child component via a PUT API call. This is possible on Strapi v4 and I believe it's not working because of a wrong validation here: https://github.com/strapi/strapi/blob/develop/packages/core/utils/src/validate/visitors/throw-unrecognized-fields.ts#L20

First, it seems a bit weird to me that this array of allowed properties is just repeating the same property twice, so it seems like a typo: https://github.com/strapi/strapi/blob/develop/packages/core/utils/src/validate/visitors/throw-unrecognized-fields.ts#L14

Locally I tried to change this and it still wasn't enough to fix this bug. So what I did as a workaround was just commenting out the `throw` at the end of the function: https://github.com/strapi/strapi/blob/develop/packages/core/utils/src/validate/visitors/throw-unrecognized-fields.ts#L61

When I removed this, the validation error didn't come up (obviously) and the feature actually worked, I was able to update my component data without having to recreate it. So I believe we need one more condition on this function that allows ID to be passed in this cases.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

innerdvations:
### What does it do?

allows component children ids to validate

TODO:

- [ ] check that this doesn't accidentally allow setting the id on create
- [ ] add api test(s) for the test case and others (because there are clearly some other issues untested)

### Why is it needed?

See #22680

### How to test it?

See #22680

### Related issue(s)/PR(s)

fixes #22680


innerdvations:
Added `don't merge` because I need to be 100% sure this won't re-open the security issue the validation is preventing

packages/core/utils/src/validate/visitors/throw-unrecognized-fields.ts
```diff
@@ -52,8 +53,13 @@ const throwUnrecognizedFields: Visitor = ({ key, attribute, path, schema, parent
...
   // allow id fields where it is needed for setting a relational id rather than trying to create with a given id
-  const canUseID = isRelationalAttribute(parent?.attribute) || isMediaAttribute(parent?.attribute);
-  if (canUseID && !ID_FIELDS.includes(key)) {
+  const canUseID =
+    parent?.attribute &&
+    (isRelationalAttribute(parent.attribute) ||
+      isMediaAttribute(parent.attribute) ||
+      isComponentAttribute(parent.attribute));
+
+  if (canUseID && ID_FIELDS.includes(key)) {
```
innerdvations, line 62: I have no idea why the logic was inverted before or how it worked

packages/core/utils/src/validate/visitors/throw-unrecognized-fields.ts
```diff
@@ -6,13 +6,14 @@ import {
...
   hasRelationReordering,
+  isComponentAttribute,
 } from '../../content-types';
 import type { Visitor } from '../../traverse-entity';
 import { throwInvalidKey } from '../utils';
 
 // TODO these should all be centralized somewhere instead of maintaining a list
-const ID_FIELDS = [constants.DOC_ID_ATTRIBUTE, constants.DOC_ID_ATTRIBUTE];
-const ALLOWED_ROOT_LEVEL_FIELDS = [...ID_FIELDS];
+const ID_FIELDS = [constants.DOC_ID_ATTRIBUTE];
```
innerdvations, line 15:
```suggestion
const ID_FIELDS = [constants.ID_ATTRIBUTE, constants.DOC_ID_ATTRIBUTE];
```

innerdvations:
Closing this because @Convly is working on a better solution

PR #22890, opened 2025-02-14
chore: fix api tests for cloning with relations

remidej:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/services/document-service/repository.ts
+++ b/packages/core/core/src/services/document-service/repository.ts
@@ -17,6 +17,7 @@ import { transformParamsDocumentId } from './transform/id-transform';
 import { createEventManager } from './events';
 import * as unidirectionalRelations from './utils/unidirectional-relations';
 import entityValidator from '../entity-validator';
+import { standarizeRelations } from './transform/relations/transform/data-ids';
 
 const { validators } = validate;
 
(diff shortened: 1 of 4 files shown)


packages/core/core/src/services/document-service/repository.ts
```diff
@@ -193,7 +194,24 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
...
         // Merge new data into it
-        (data) => merge(data, queryParams.data),
+        async (data) =>
+          /**
+           * Merge receives [{ connect: [{ id: 1 }] }] and { connect: [{ documentId }] }]
+           * but it does not merge arrays, so only keeps one of them.
+           * We want to not do connect for everything, but use both connect (for the new one) and
+           * set (for the old one) at the same time. This requires a change at the database level.
+           * Can this be done?
+           */
```
Marc-Roig, line 204: If I recall correctly, the limitation here was that using "set" for the old could stole relations, and if the user disconnected some relations to prevent that, they would still be cloned.

packages/core/core/src/services/document-service/transform/relations/transform/data-ids.ts
```diff
@@ -63,6 +63,24 @@ const getRelationIds = curry(
   }
 );
 
+/**
+ * Ensure set/connect/disconnect syntax for all relations in the data object
+ */
+const standarizeRelations = (data: Record<string, any>, uid: string) => {
```
markkaylor, line 69: typo?

```suggestion
const standardizeRelations = (data: Record<string, any>, uid: string) => {
```

remidej: Typo that's been here for ages indeed :)

markkaylor: oh I thought you were adding this function 🙃

remidej: I don't even remember but I'll fix it anyway

tests/api/core/content-manager/api/relations.test.api.ts
```diff
@@ -1236,7 +1243,7 @@ describe('Relations', () => {
       expect(populatedCreatedShop.products_oo).toMatchObject({ documentId: docid1 });
     });
 
-    test('Clone entity with relations and set data should not steal relations', async () => {
+    test.skip('Clone entity with relations and set data should not steal relations', async () => {
```
Marc-Roig, line 1246: are these two failing now?

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -186,19 +193,60 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
...
+        uid
+      );
+
+      const newEntry = await entries.create({
+        ...queryParams,
+        data: newData,
+        status: 'draft',
+      });
+
+      // Step 2 (optional): Update with queryParams.data if it exists, after some transformations
```
Marc-Roig, line 212: I think we should explain in a more abstract way whats happening here, what is "standarizing" relations, and

remidej:
closing in favor of #23553

PR #22904, opened 2025-02-17
fix: password validation tests

innerdvations:
### What does it do?

- fix the validation on front end password reset and registration because `TextEncoder` is not available there
- fix the message itself which was incorrectly wrapped in an object
- adds additional tests for too short / too long
- fixes an api test that was not expecting correctly

### Why is it needed?

- the front end tests are failing because of the above reasons

### How to test it?

tests should pass

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/utils/strings.ts
+++ b/packages/core/admin/admin/src/utils/strings.ts
@@ -5,8 +5,34 @@ function getByteSize(value: string) {
     return new TextEncoder().encode(value).length;
   }
 
-  // Fallback: Count bytes manually (assuming most non-ASCII chars are 2+ bytes)
-  return value.split('').reduce((sum, char) => sum + (char.charCodeAt(0) > 127 ? 2 : 1), 0);
+  // Manual UTF-8 encoding (fully reliable fallback)
+  let bytes = 0;
+  for (let i = 0; i < value.length; i++) {
+    const code = value.charCodeAt(i);
+
+    if (code < 0x80) {
+      bytes += 1; // ASCII (1 byte)
+    } else if (code < 0x800) {
+      bytes += 2; // 2-byte UTF-8
+    } else if (code >= 0xd800 && code <= 0xdbff) {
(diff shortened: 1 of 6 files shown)


packages/core/admin/admin/src/utils/strings.ts
```diff
@@ -1,3 +1,38 @@
 const capitalise = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
 
-export { capitalise };
+function getByteSize(value: string) {
+  if (typeof TextEncoder !== 'undefined') {
```
markkaylor, line 4: When would `TextEncoder` ever be undefined? Is this just for the testing environment?

innerdvations: AFAIK the test environment or a very old browser

markkaylor: I see a lot of suggestions for [providing the implementation](https://github.com/inrupt/solid-client-authn-js/issues/1676#issuecomment-917016646) to the jest environment. Curious if you tried that? I'm not sure the manual implementation below is worth it to support such outdated browsers. Could just be me and my little brain but I struggle to understand what's going on down there 🤯

innerdvations: I'll try that and see!

innerdvations: Ok, seems to work, we can go with that instead.

packages/core/admin/admin/src/utils/strings.ts
```diff
@@ -1,3 +1,7 @@
 const capitalise = (str: string): string => str.charAt(0).toUpperCase() + str.slice(1);
 
-export { capitalise };
+function getByteSize(value: string) {
+  return new globalThis.TextEncoder().encode(value).length;
```
markkaylor, line 4: I think this should be:

```suggestion
  return new TextEncoder().encode(value).length <= 72;
```

PR #22905, opened 2025-02-17
fix(data-transfer): surface errors from createassetsreadstream

jhoward1994:
### What does it do?

Ensures errors are caught and displayed from `createAssetsReadStream`

### How to test it?

Updated unit test

A manual test requires a bit of set up:
- Modify `packages/core/data-transfer/src/strapi/providers/local-source/assets.ts` to simulate an error being thrown. e.g. add `throw new Error('Testing DTS');` at L14
- Run `examples/getstarted` with `yarn run build; yarn run start` (port 1337)
- Using another example project run `yarn strapi transfer --from http://localhost:1337/admin --from-token=TRANSFER_TOKEN ` (this will try to transfer data from the app at port 1337 to the app where you are running this command)
- See the error is surfaced in the users data transfer CLI and in the console of the running strapi app.

### Related issue(s)/PR(s)

https://strapi-inc.atlassian.net/browse/DX-1857


Bassel17:
Looks good,
we also need to surface errors from the write streams, especially on the remote providers
I would say the remote destination provider, and local destination provider

innerdvations (approved):
LGTM! This reports the type of errors that were hidden in the support tickets but like Bassel said, we will need to add this to a lot more places. Do you want to do that here or in a follow-up?

jhoward1994:
Given that this covers the support tickets we received do you agree to merge this and manage the rest in a follow up?

Bassel17 (approved):

PR #22908, opened 2025-02-17
chore: remove files from the old homepage

lucasboilly:
### What does it do?

I removed files from the old homepage.

### Why is it needed?

They are not used anymore in the admin.

### How to test it?

Nothing to test.

### Related issue(s)/PR(s)

No related issue/pull request.


remidej (approved):
good catch

PR #22913, opened 2025-02-17
feat: improve package registry URL determination during Strapi upgrade

mottet:
### What does it do?

During the run of Strapi's upgrade script, when fetching information about the @strapi/strapi package, instead of forcing the use of the official NPM registry https://registry.npmjs.org/, the registry is chosen from:

1. NPM_REGISTRY_URL environment variable
2. If not found, detect the used package manager (npm or yarn) and get the registry setup for it
3. If it cannot find it, fallback to the default value https://registry.npmjs.org/

### Why is it needed?

If the official NPM registry is not accessible or the user does not want to use it for any reason, the user should be able to choose another registry.

For example, this is needed in organisation that block access to the offical registry and require the use of a controlled mirror registry.

### How to test it?

Build @strapi/upgrade and check unit tests
``` bash
yarn nx build @strapi/upgrade
yarn nx test:unit @strapi/upgrade
```

Create a Strapi project that doesn't use the lastest version. 
``` bash
npx create-strapi@5.0.5 my-strapi-project-to-upgarde
```

Run the upgrade script on the created Strapi project using the locally build @strapi/upgrade. Use the debug option to display information about used registry as well as the dry option to run multiple time the command and so try different ways to setup the registry.
``` bash
# Used the registry of the project's package manager
npx /path/to/strapi/packages/utils/upgrade/bin/upgrade.js latest --dry --debug
# Used the registry set in NPM_REGISTRY_URL
NPM_REGISTRY_URL=https://registry.yarnpkg.com npx /path/to/strapi/packages/utils/upgrade/bin/upgrade.js latest --dry --debug
```

The update fails but what we are looking for here is the line `[DEBUG] [2025-02-17T23:37:23.605Z] Using package manager registry: https://registry.npmjs.org/` that give us the used registry. Also the next few lines should show that it was able to fetch @strapi/strapi and get the last version to upgrade to.

If you got this message with the first command: 
```
[WARN]  [2025-02-17T23:47:13.204Z] Failed to determine registry URL from package manager
[DEBUG] [2025-02-17T23:47:13.205Z] Using default registry: https://registry.npmjs.org
```
Run a package dependency installation and rerun the upgrade script.

If you do not have the `Failed to determine registry`, to try the fallback to default registry, delete the lock file and the node_modules folder. 

For the test with the environment variable, you should have the log `[DEBUG] [2025-02-17T23:44:24.401Z] Using NPM_REGISTRY_URL: https://registry.yarnpkg.com`


mottet:
Could you please add the source tag "source: utils:upgrade" ? 😃

Bassel17:
Thank for your contribution !
can you attempt to resolve the conflicts please 🙏

PR #22914, opened 2025-02-17
fix(media-library): Fix broken breadcrumbs for deeply nested folders (404 Error)

### LINKED ISSUE #22885: Media Library Breadcrumbs Broken for Deeply Nested Folders (404 Error)
### Node Version

20.18.0

### NPM/Yarn/PNPM Version

npm 10

### Strapi Version

5.10.2

### Operating System

Linux (Debian/Ubuntu)

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

There's a problem with the Media Library's breadcrumb navigation when you have folders inside other folders (nested folders). If you create a series of folders within each other, like Folder1 > Folder2 > Folder3, and you're looking at the contents of Folder3, trying to go back to Folder1 using the breadcrumb links at the top leads to a 404 error.

### Steps to Reproduce

1. Open Strapi's Media Library.
2. Create a folder named Folder1.
3. Inside Folder2, create a folder named Folder3.
4. Go into the Folder3 folder.
5. At the top, in the breadcrumb trail, click on Folder1.

### Expected Behavior

Clicking on Folder1 should take you back to the Folder1 folder, showing its contents.

### Logs

```shell
{"data":null,"error":{"status":404,"name":"NotFoundError","message":"Not Found","details":{}}}
```

### Code Snippets

_No response_

### Media

Image

### Additional information

_No response_

### Confirmation Checklist

0bhi:
### What does it do?

This commit fixes the issue where breadcrumbs were broken for deeply nested folders in the Media Library, resulting in a 404 error. The fix ensures that breadcrumbs are correctly generated for all folder levels.

### Why is it needed?

The current implementation of breadcrumbs in the Media Library does not handle deeply nested folders correctly, leading to a 404 error. This fix addresses the issue and improves the user experience by ensuring breadcrumbs work as expected.

### How to test it?

1. Go to the Media Library.
2. Create multiple nested folders.
3. Navigate through the nested folders and observe the breadcrumbs.
4. Ensure that the breadcrumbs are correctly displayed and no 404 errors occur.

### Related issue(s)/PR(s)

Fixes #22885

### DIFF
--- a/packages/core/upload/admin/src/components/Breadcrumbs/CrumbSimpleMenuAsync.tsx
+++ b/packages/core/upload/admin/src/components/Breadcrumbs/CrumbSimpleMenuAsync.tsx
@@ -22,6 +22,7 @@ export const CrumbSimpleMenuAsync = ({
   const [shouldFetch, setShouldFetch] = React.useState(false);
   const { data, isLoading } = useFolderStructure({ enabled: shouldFetch });
   const { pathname } = useLocation();
+  const fullPathname = `/admin${pathname}`;
   const [{ query }] = useQueryParams();
   const { formatMessage } = useIntl();
 
@@ -70,8 +71,8 @@ export const CrumbSimpleMenuAsync = ({
             );
           }
 
-          const url = getFolderURL(pathname, query, {
-            folder: typeof ascendant?.id === 'string' ? ascendant.id : undefined,
+          const url = getFolderURL(fullPathname, query, {
+            folder: String(ascendant.id),
             folderPath: ascendant?.path,
           });
 
--- a/packages/core/upload/admin/src/components/SelectTree/utils/flattenTree.ts
+++ b/packages/core/upload/admin/src/components/SelectTree/utils/flattenTree.ts
@@ -2,6 +2,7 @@ type TreeNode<T> = {
   value: T;
   children?: TreeNode<T>[];
   label?: string;
+  path?: string;
 };
 
 export type FlattenedNode<T> = {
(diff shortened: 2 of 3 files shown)


0bhi:
"Hey  @soupette , I’ve submitted this PR, and I’d love your feedback when you get a chance. Let me know if anything needs improvement!"

soupette:
> Hey @soupette , I’ve submitted this PR, and I’d love your feedback when you get a chance. Let me know if anything needs improvement!

Thanks for pinging me I'll let @simotae14 validate your PR.

0bhi:
Hey @simotae14, could you please review the PR and let me know if there's anything that needs improvement? Thank you!

simotae14 (commented):
Hello @0bhi thanks for the PR and the patience, Can I ask you to add a test for the CrumbSimpleMenuAsync? Thanks

packages/core/upload/admin/src/components/Breadcrumbs/CrumbSimpleMenuAsync.tsx
```diff
@@ -22,6 +22,7 @@ export const CrumbSimpleMenuAsync = ({
   const [shouldFetch, setShouldFetch] = React.useState(false);
   const { data, isLoading } = useFolderStructure({ enabled: shouldFetch });
   const { pathname } = useLocation();
+  const fullPathname = `/admin${pathname}`;
```
Adzouz: Instead of hardcoding the /admin directly in the url, we would want to use `react-router-dom` hooks to add the prefix automatically.
Unfortunately, I don't think NavLink is compatible with MenuItem so I would maybe suggest to change the link for a button by removing isLink from MenuItem and handle the redirection like so: `<MenuItem key={ascendant.id} onSelect={() => navigate(url)}>` (where navigate is coming from useNavigate.
What do you think?

Adzouz (changes requested):
Thank you for the contribution 🙏 Added a comment that we would need to address before merging.

packages/core/upload/admin/src/components/Breadcrumbs/CrumbSimpleMenuAsync.tsx
remidej: I agree we can't rely on `/admin`, as the admin path can be changed by users. We should be able to make it work using some sort of relative link.

Aren't we just setting a search param when navigating from folder to folder? If so we can tell react router to only update the search with something like this. This way we don't need to build the whole url

```jsx
<Link to={{ search: newSearchParams.toString() }}>
```

(maybe it's not a Link in this context, but the `To` object is supported in other contexts as well)

Adzouz: I updated the logic + the test. Can you tell me if it's ok for you @remidej?

remidej: the change lgtm, I'm testing now 👍

PR #22919, opened 2025-02-18
future: toggle preview side editor

remidej:
### What does it do?

- [x] adds a floating icon button
- [x] animates the transition

### How to test it?

open a preview page, toggle the thing


markkaylor (commented):
Looks good to me, works nicely. Just one question:

Isn't the arrow currently backwards? This is the arrow when it's closed, which looks like a back button:

Screenshot 2025-02-20 at 11 32 31

Even though we will update the icons to match the designs still seems confusing. Shouldn't the arrow point towards the way it will move?

The designs show the back direction for when the editor is open:

Screenshot 2025-02-20 at 11 33 27

remidej:
@markkaylor you're very much right. I'll update that (and open that icon DS PR) now

markkaylor (approved):
Nice ✨

PR #22942, opened 2025-02-19
feat(preview): add review workflow injection zone

markkaylor:
### What does it do?

- Create a preview.actions injection zone
- Uses that injection zone for review workflows
- Updates the InjectionZone component to pass the area to the injected component. We need something like this to conditionally render things in the ReviewWorkflows components. It think it could be useful in other cases where components are injected in different places that require different behaviors. 

### Why is it needed?

To be able to update stage and assignee from preview

### How to test it?

Go to the preview page and update the stage and assignee

### Related issue(s)/PR(s)

This PR should be merged after https://github.com/strapi/strapi/pull/22902

### DIFF
--- a/packages/core/review-workflows/admin/src/index.ts
+++ b/packages/core/review-workflows/admin/src/index.ts
@@ -1,5 +1,7 @@
 import { PLUGIN_ID, FEATURE_ID } from './constants';
+import { AssigneeSelect } from './routes/content-manager/model/id/components/AssigneeSelect';
 import { Panel } from './routes/content-manager/model/id/components/Panel';
+import { StageSelect } from './routes/content-manager/model/id/components/StageSelect';
 import { addColumnToTableHook } from './utils/cm-hooks';
 import { prefixPluginTranslations } from './utils/translations';
 
@@ -47,6 +49,19 @@ const admin: Plugin.Config.AdminInput = {
           const { PurchaseReviewWorkflows } = await import('./routes/purchase-review-workflows');
           return { default: PurchaseReviewWorkflows };
         },
+      });
+    }
+  },
+  bootstrap(app: StrapiApp) {
+    if (window.strapi.features.isEnabled(FEATURE_ID)) {
+      app.getPlugin('content-manager').injectComponent('preview', 'actions', {
+        name: 'review-workflows-assignee',
+        Component: AssigneeSelect,
+      });
+
+      app.getPlugin('content-manager').injectComponent('preview', 'actions', {
+        name: 'review-workflows-stage',
+        Component: StageSelect,
       });
     }
   },
--- a/packages/core/review-workflows/admin/src/index.ts
+++ b/packages/core/review-workflows/admin/src/index.ts
@@ -1,5 +1,5 @@
 import { PLUGIN_ID, FEATURE_ID } from './constants';
-import { AssigneeSelect } from './routes/content-manager/model/id/components/AssigneeSelect';
+import { Header } from './routes/content-manager/model/id/components/Header';
 import { Panel } from './routes/content-manager/model/id/components/Panel';
 import { StageSelect } from './routes/content-manager/model/id/components/StageSelect';
 import { addColumnToTableHook } from './utils/cm-hooks';
@@ -56,12 +56,7 @@ const admin: Plugin.Config.AdminInput = {
     if (window.strapi.features.isEnabled(FEATURE_ID)) {
       app.getPlugin('content-manager').injectComponent('preview', 'actions', {
         name: 'review-workflows-assignee',
-        Component: AssigneeSelect,
-      });
-
-      app.getPlugin('content-manager').injectComponent('preview', 'actions', {
-        name: 'review-workflows-stage',
-        Component: StageSelect,
+        Component: Header,
       });
     }
   },
--- a/packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
+++ b/packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
@@ -89,6 +89,11 @@ const AssigneeSelect = ({ area }: { area?: InjectionZoneArea }) => {
     }
   };
 
+  const assigneeLabel = formatMessage({
+    id: 'content-manager.reviewWorkflows.assignee.label',
+    defaultMessage: 'Assignee',
+  });
+
   return (
     <Field.Root
       name={ASSIGNEE_ATTRIBUTE_NAME}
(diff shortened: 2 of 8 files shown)


packages/core/review-workflows/admin/src/index.ts
```diff
@@ -50,6 +52,19 @@ const admin: Plugin.Config.AdminInput = {
       });
     }
   },
+  bootstrap(app: StrapiApp) {
+    if (window.strapi.features.isEnabled(FEATURE_ID)) {
```
remidej, line 56: would there be danger in injecting the component regardless of the flag? since we do the check in the header itself?

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
```diff
@@ -19,7 +19,10 @@ import { getDisplayName } from '../../../../../utils/users';
 
 import { ASSIGNEE_ATTRIBUTE_NAME } from './constants';
 
-const AssigneeSelect = () => {
+type InjectionZoneArea = 'preview.actions';
+
+const AssigneeSelect = ({ area }: { area?: InjectionZoneArea }) => {
+  const isCompact = area === 'preview.actions';
```
remidej, line 25: I assume there's enough complexity in there that you'd rather do this + pass the area prop to every injected component rather than have 2 versions of this component?

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/StageSelect.tsx
```diff
@@ -160,13 +164,26 @@ export const StageSelect = () => {
...
-          })}
-        </Field.Label>
+        {isCompact ? (
+          <VisuallyHidden>
+            <Field.Label>
+              {formatMessage({
+                id: 'content-manager.reviewWorkflows.stage.label',
+                defaultMessage: 'Review stage',
+              })}
+            </Field.Label>
```
remidej, line 174: Great a11y stuff! I'd just recommend storing the formattedMessage in a const above to avoid writing it twice

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/StageSelect.tsx
```diff
@@ -192,7 +209,7 @@ export const StageSelect = () => {
           customizeContent={() => {
             return (
               <Flex tag="span" justifyContent="space-between" alignItems="center" width="100%">
-                <Typography textColor="neutral800" ellipsis>
+                <Typography textColor="neutral800" ellipsis style={{ lineHeight: 'inherit' }}>
```
remidej, line 212: We don't have a prop for lineHeight?

packages/core/review-workflows/admin/src/index.ts
markkaylor: Where do we do the check in the header? Maybe I missed that. I didn't have this but the components were still showing without the license. I saw review-workflows panel api and content-releases injection zone handle it this way.

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/StageSelect.tsx
markkaylor: yeah I think we do my bad

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
remidej: Why not indeed. It would be easy easy to fix anyway if we get the report that it somehow broke something

packages/core/review-workflows/admin/src/index.ts
remidej: Oh this is a feature flag, not a future one, gotcha. Makes sense then

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
remidej, line 1: Do you know why changing the stage or the assignee seems to rerender the whole thing?

https://github.com/user-attachments/assets/8ee3abec-31aa-4de4-9ee9-d59e23e91d3a

markkaylor: Not sure, but yeah that is annoying, I'll take a look to see if I can improve it

remidej: Doesn't seem like it's related to review workflows in particular actually. I get that full refresh when I update the content and save too

remidej: It could be related to us dispatching the re-initialize form event on every update for no valid reason

markkaylor: Damn that's true, not sure how I didn't notice that before 🤦‍♂️ . I guess I figured that was the same behavior on editviewpage. My bad.

markkaylor: Unfortunately it doesn't look like resetForm is the culprit. I had to refactor to handle hints, errors, and returning null when content-type doesn't have a review workflows so I haven't had time to investigate further yet.

markkaylor:
Thanks for the reviews @remidej, unfortunately I had to refactor quite a bit. Let me know what you think. We don't need to rely on a prop from the InjectionZone anymore.

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/AssigneeSelect.tsx
markkaylor: Let's keep moving forward, I can take another look at this when I add it for review-workflows

simotae14:
It is not related to your PR but in the Review Workflow dropdown we need to fix the alignment of the X icon and the arrow down icons as you can see here 
Screenshot 2025-02-24 alle 09 47 09

tests/e2e/tests/review-workflows/content-manager.spec.ts
simotae14, line 1: I don't know if it is doable but maybe it could be worth checking if the Review Workflow combobox is disabled in case a role is not able to change the stage (a minor check is to be sure the tooltip is also shown)

simotae14 (commented):
Overall the code looks great (I liked the way you handled all the labels). I tried to test the solution and it works (even if there is the refresh problem)
Let me know if it is doable to test also the cases where the current role can't change the stage otherwise for me the PR is ready to be approved

simotae14:
There is a snapshot test failing in the FE unit tests inside the Content manager

tests/e2e/tests/review-workflows/content-manager.spec.ts
markkaylor: That should already be tested through the existing unit tests 👍

simotae14 (approved):
thanks for the update of the failing snapshot test. The looks is ok, well done

PR #22952, opened 2025-02-20
feat(relations-on-the-fly): refactoring context

simotae14:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -197,7 +191,7 @@ const UnstableRelationsField = React.forwardRef<HTMLDivElement, RelationsFieldPr
      * Same with `uid` and `documentModel`.
      */
     const id = componentId ? componentId.toString() : documentId;
-    const model = isModalOpen ? (documentModel ?? componentUID) : (componentUID ?? documentModel);
+    const model = documentModel ?? componentUID ?? componentUID ?? documentModel;
```
markkaylor, line 194: Aren't the last two unnecessary?

```suggestion
    const model = documentModel ?? componentUID
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -1450,11 +1439,12 @@ const UnstableListItem = ({ data, index, style }: ListItemProps) => {
...
     relations,
     targetModel,
-    isModalOpen,
     changeCurrentRelation,
   } = data;
-  const { formatMessage } = useIntl();
   const [showModal, setShowModal] = React.useState(false);
+  const { formatMessage } = useIntl();
+  const isModalOpen = useRelationContext('RelationContext', (state) => state.isModalOpen);
+  const setIsModalOpen = useRelationContext('RelationContext', (state) => state.setIsModalOpen);
```
markkaylor, line 1447: Why do we have an internal state and the state being managed by the context? Do we need to have the modal state in the context?

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -25,6 +28,22 @@ import { FormLayout } from './components/FormLayout';
 import { Header } from './components/Header';
 import { Panels } from './components/Panels';
 
+export interface CurrentRelation {
+  documentId: string;
+  model: string;
+  collectionType: string;
+}
+
+interface RelationModalContextValue {
```
markkaylor, line 37:
```suggestion
interface RelationContextValue {
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
simotae14: we need the modal state in the context in other part of the application because I need to distinguish if it is the modal is already opened or not (for example it will be useful for the back button history)
If, at the end of the implementation the context var is useless I can remove it

markkaylor: Do we need it in other places now? Do we need both the local state and the context state here in this component?

simotae14: I need the local state because otherwise the modal has the rerendering issue as you know but I need to keep track in the context if the modal is open or not.

Issue #22951, opened 2025-02-20
Missing Await in Asynchronous Function Causing Incorrect Conditional Branching


AtsushiMomoyama:
At [collection-types.ts#L423](https://github.com/strapi/strapi/blob/8264ee203574958f5f79b39d6365b90d99074674/packages/core/content-manager/server/src/controllers/collection-types.ts#L423), it appears that an asynchronous function call is missing the await keyword. As a result, the conditional branch that depends on its return value always evaluates to false, leading to behavior that deviates from the intended functionality.

Please note that this report is solely for informational purposes, and no fixes will be implemented on our end. It would be greatly appreciated if you could address this issue at your earliest convenience.

PR #22956, opened 2025-02-20
fix: await document exists

Marc-Roig:
### What does it do?

At [collection-types.ts#L423](https://github.com/strapi/strapi/blob/8264ee203574958f5f79b39d6365b90d99074674/packages/core/content-manager/server/src/controllers/collection-types.ts#L423), it appears that an asynchronous function call is missing the await keyword. As a result, the conditional branch that depends on its return value always evaluates to false, leading to behavior that deviates from the intended functionality.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/22951

### DIFF
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -420,7 +420,7 @@ export default {
       const isUpdate = !isCreate;
       if (isUpdate) {
         // check if the document exists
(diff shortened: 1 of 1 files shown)


alexandrebodin (approved):
LGTM. This might be fixing a lot of D&P issues cc @oiorain

remidej (approved):
I'm not familiar with the related bugs but this looks good

PR #22964, opened 2025-02-21
fix(data-transfer): race condition and backpressure in large asset transfers

innerdvations:
### What does it do?

- [**remote source**] fixes race condition where receiving the second chunk payload of a media item before the first one has finished writing results in an attempt to write the chunks out of order
- [**remote source**] removes some cases where errors were ignored; previously it was crashing when that would happen, but with this fix there would have been a few edge cases which could have resulted in corrupted data
- [**transfer engine**] refactor to fix a backpressure issue and abort control
- [**local destination**] ensure return when callbacks are called
- adds more debug reporting

TODO:

- [x] refactor the payload processor to make more sense
- [x] fix the backpressure issue with assets
- [x] add back the timeout so that it doesn't hang if an 'end' is never received
- [x] fix issue with interrupting a transfer with ctl+c
- [x] add unit tests where possible

### Why is it needed?

It was nearly impossible to complete a pull transfer when a large number of large media assets were present without crashing

### How to test it?

- create a strapi project with a large number (>100) of large media (>10mb) called ProjectA
  - I suggest using https://github.com/MaksZhukov/strapi-plugin-generate-data
- clone the project schemas into a ProjectB so that a transfer can be done
- in the ProjectB dir CLI, run a transfer `from` ProjectA
- testing from a local project -> local project almost guarantees the error; testing on a remote server may fail less often because the race condition occurs when the second chunk of a file is received before the very first one has finished writing to its stream

Without this PR, there's an extremely high liklihood that it will fail -- odds increase for every asset over ~2mb and on faster connections, so realistically this should always fail especially if you're trying it locally.

With this PR, it should succeed.

Please also confirm that all assets have been transferred, and are not corrupted -- although this PR adds additional protections for that compared with the previous version

Here is a helpful shell script to compare the files in two directories:

```
diff -qr ./getstarted/public/uploads ./getstarted-2/public/uploads | while read -r line; do
    if [[ $line == Files* && $line == *differ ]]; then
        # Handle differing files
        file1=$(echo "$line" | awk '{print $2}')
        file2=$(echo "$line" | awk '{print $4}')
        size1=$(stat -f%z "$file1")
        size2=$(stat -f%z "$file2")
        echo "Different: $file1 ($size1 bytes) vs $file2 ($size2 bytes)"
    elif [[ $line == Only* ]]; then
        # Handle files only in one of the directories
        dir=$(echo "$line" | awk '{print $3}')
        file=$(echo "$line" | awk '{print $4}')
        full_path="$dir/$file"
        size=$(stat -f%z "$full_path" 2>/dev/null || echo "N/A")
        echo "Only in $dir: $file ($size bytes)"
    fi
done
```

ALSO:

Please test interrupting transfers at various stages with ctrl+c. I had to rewrite the signal listeners due to the changes.

### Related issue(s)/PR(s)

- several internal support ticket as well as several reported issues on github, but impossible to say for sure because the errors from data-transfer were not being reported properly and could be anything -- once this is released, we will go through and have the issues tested and confirm if they are fixed


derrickmehaffy:
building experimental to test with: https://github.com/strapi/strapi/actions/runs/13461042920

derrickmehaffy:
New exp with latest changes: https://github.com/strapi/strapi/actions/runs/13507842660

derrickmehaffy:
Running another to see if I can get rid of an error I think was introduced while we were fixing the CTB: https://github.com/strapi/strapi/actions/runs/13508199877

image

derrickmehaffy:
exp version: `0.0.0-experimental.f0ae22a92ff2c56127a7f843a3e973c934c1aab4`

derrickmehaffy:
So running some tests, first off a few things:

- My images were super large as I used that plugin mentioned in OP to generate a ton of data
- I was testing using strapi cloud 
- The transfer was successful

Something that easier to see in cloud that you can't see locally is the performance impact it has on resource usage, I noticed looking at the cloud resources that the CPU was spiking:

image

With the transfer complete it did take a bit to transmit this relatively small amount of data (as long as it works for now we can work on speed later)

image

Some stats (and for clarification this going over my 1gbps up/35mbps local connection from Las Vegas to NYC

- Total transfer time is 352.23 seconds reported by the command which is roughly `1,256.14 KB/s` or `1.23 MB/s`
- Just the asset transfer is `547.55 KB/s` or `0.535 MB/s`

---

I'll have to do a bit more testing with much larger assets (may do some testing with some 30mb video assets to see) and I'll do some local to local testing but it's frustrating with our stupid vite port bug since we can't run two apps in dev mode at the same time.

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
```diff
@@ -167,57 +172,29 @@ class RemoteStrapiSourceProvider implements ISourceProvider {
           }
 
           // Writes the asset's data chunks to their corresponding stream
-          else if (action === 'stream') {
-            // If the asset hasn't been registered, or if it's been closed already, then ignore the message
+          // "end" is considered a chunk, but it's not a data chunk, it's a control message
```
innerdvations, line 175: The refactoring in this file is fixing a case where, when two payloads were sent at once, the first part of the first payload started writing immediately, the second payload pushed to the queue, then the second part of the first payload is pushed to the queue after.

Now instead of trying to manage locking ourself, we just -always- push to the queue in the order it comes in, and then notify the queue processor that more pieces of an asset have come in. This required switching from the idea of queuing only data to instead queueing control messages, so that we can easily ensure 'end' actions happen at the correct time, again because we do not want to have to attempt to maintain our own locking system.

packages/core/data-transfer/src/strapi/remote/handlers/pull.ts
```diff
@@ -175,6 +175,7 @@ export const createPullController = handlerControllerFactory<Partial<PullHandler
         error: null,
         id,
       });
+      batch = [];
```
innerdvations, line 178: This is just enforcing something that is necessary rather than allowing room for potentially bugs.

packages/core/data-transfer/src/strapi/queries/link.ts
```diff
@@ -302,9 +302,7 @@ export const createLinkQuery = (strapi: Core.Strapi, trx?: Knex.Transaction) =>
         assignOrderColumns();
         const qb = connection.insert(payload).into(addSchema(joinTable.name));
         if (trx) {
-          await trx.transaction(async (nestedTrx) => {
-            await qb.transacting(nestedTrx);
-          });
+          await qb.transacting(trx);
```
innerdvations, line 305: We don't need to create a new transaction, we're already in one for the data transfer. I think this may have been causing some connection pool issues.

This isn't a strictly necessary fix for this PR, but I've included it here because the amount of manual testing on data transfer issues is high, and I'd rather do it just once.

innerdvations:
> memory spiking

I think I've just fixed that, could you please try again?

> stupid vite port bug since

yeah, I hope we can fix that soon, I usually just keep them running in production mode with `yarn start`

Bassel17 (dismissed):
Looks, great works!
nice work on handling the remote-source assets it looks much better

packages/core/data-transfer/src/strapi/queries/link.ts
Bassel17: yeah indeed not needed, it's shouldn't have been causing issues though, if am not wrong the logic of the transaction will already ignore this nested one because it's inside one already

Bassel17:
we do need, to follow this up with some performance improvements, maybe the batch sizes although am not sure if  we change it will have lot's of positive impact.

innerdvations:
Bigger chunks will be the "easy" improvement, but I think the biggest true improvement is if we can send assets as binary rather than as json sent through websocket messages.

Although another quick win there is to not send prettified JSON at any point... during my debugging I noticed the data being sent through the system has probably 5% whitespace which is completely unnecessary

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
```diff
@@ -128,13 +140,27 @@ class RemoteStrapiSourceProvider implements ISourceProvider {
 
     // Init the asset map
     const assets: {
-      [filename: string]: IAsset & {
+      // TODO: could we include filename in this for improved logging?
```
Convly, line 143: IAsset does contain the filename, no?

Convly (commented):
Overall LGTM, very nice work :pray: 
It would be awesome to have some unit tests if possible though

packages/core/data-transfer/src/strapi/providers/remote-source/index.ts
innerdvations: Ahhh, maybe it does, I completely missed that it was an IAsset. I might add a log in there in that case

innerdvations:
> It would be awesome to have some unit tests if possible though

Yep, will try to get those in tomorrow or Monday

derrickmehaffy:
> Bigger chunks will be the "easy" improvement, but I think the biggest true improvement is if we can send assets as binary rather than as json sent through websocket messages.
> 
> Although another quick win there is to not send prettified JSON at any point... during my debugging I noticed the data being sent through the system has probably 5% whitespace which is completely unnecessary

While I don't think we are anywhere close to implementing something like this (nor am I sure if it's even a good idea/possible + we are dipping into c or c++ levels of not picking)

But

I think it would actually be cool to do dynamic chunk sizes depending on runtime conditions, I know in node or at least through v8 we can fetch the maximum memory size and current usage  + if we can get current CPU utilization + if we can dynamically ramp up chunk size based on previous chunk transfer performance.

Using pseudo logic here but say chunk size is a simple integer between 1 and 100

- we fire the first chunk with a size 1, wait for it to finish and communicate performance back to the origin
- if it falls within acceptable increase parameters we dynamically update the chunk size 2
- repeat performance calculation (but double using exponential increase so 1, 2, 4, 8, 16, 32, 64, ect)
- eventually we will either hit the "sweet spot" for this specific operation or we back off one increase if we cross a negative performance report

^ the key thing here is that we will do this continually as the sweet spot may change depending on the applications load (since we don't have a maintenance mode)

I know it sounds complex and to be honest I don't even know how difficult this would be but many data transfer systems operate with this ramping methodology.

(Perfect world we should also be monitoring the database load also but that won't always be available depending the database and DB permissions)

innerdvations:
Yeah, I agree, it shouldn't be too hard to implement some sort of rudimentary variable chunk sizes. I feel like connection speed is going to be the determining factor rather than memory; as long as backpressure is working (which has admittedly been broken several times in the past) I haven't seen any memory issues.

packages/core/data-transfer/src/engine/__tests__/engine.test.ts
```diff
@@ -1116,7 +1116,7 @@ describe('Transfer engine', () => {
 
       // Source that detects when it is paused
       const source = {
-        ...createSource({ entities: sourceData }),
+        ...createSource({ entities: sourceData as any }),
```
Convly, line 1119: Why do we need this?

innerdvations: test data wasn't implemented fully, I just updated it

PR #22971, opened 2025-02-21
proof-of-concept: transfer remote <-> file

innerdvations:
### What does it do?

Adds `--to-file` and `--from-file` along with the same options import/export has, such as `--to-key` to provide a key, `--no-to-encrypt` to disable encryption, etc

**Requires also adding `--experimental` to be able to use them**

Note that remote <-> remote is still disallowed because it would launch an unrelated local Strapi in order to run and behaves unpredictably, but it would be quite simple to get that working too

Also note that even remote->file will still technically invoke the local strapi instance because that's how it was originally implemented, but that limitation could be removed along with allowing remote-remote

### Why is it needed?

Needed to do it for debugging some stuff, thought I'd clean it up and put it here to share

### How to test it?

run remote transfers with `--from https://your-source --to-file exportfile` or `--from-file importfile --to https://your-destination`

### Related issue(s)/PR(s)


derrickmehaffy:
Hoooo this is interesting and could be really useful

innerdvations:
Closing because this is very old and served its purpose

Issue #22969, opened 2025-02-21
Component icons too small and missing accessibility


Da-nie-elT:
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

NPM 11.1.0

### Strapi Version

5.8.1

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

The icons in the icon selection for a component are too small and have no accessibility like icon value in a tooltip.

Strapi v5:
![Image](https://github.com/user-attachments/assets/cb5129a8-5316-4cca-971e-1d4451c40d4f)

Strapi v4.14.4:
![Image](https://github.com/user-attachments/assets/3452b605-34aa-4768-bc1b-7b801d3dd7f9)

These are both at browser zoom 110% & browser font size set to "Very large"

### Steps to Reproduce

1. create new component.
2. see icons.

### Expected Behavior

The icons should be bigger and should have their value/name be displayed in a Design-System Tooltip on hover.

And maybe the icons list could be expanded to include all possible icons available by the Design-System.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #22973, opened 2025-02-21
Fix:#22969 Resolve small icons issue and added tooltip for icons

Anandoganiya:
Fixed small icons issue by adding width the height to icon component.
Added tooltip component  to icon to enhance accessibility .


Bassel17 (approved):
LGTM! thanks for adding this

PR #22990, opened 2025-02-25
fix(preview): updating the document causes full page reload

markkaylor:
### What does it do?

- Remove the full page loading state triggered by the document action
- Handled the same way as the edit view

### Why is it needed?

- To not have a full page reload when we update or publish

### How to test it?


simotae14 (approved):
The code looks good, and I tried your solution and there is no more the full page reload when I save some changes in the document, well done 💯

PR #22993, opened 2025-02-25
fix(data-transfer): server crash when remote data transfer is interrupted by client

innerdvations:
### What does it do?

Log but suppress the error when attempting to send a ws confirmation message

_This is not the ideal fix_, we should catch _all_ of these at a higher level, but it would be complex to find the correct solution. In the meantime, this will fix probably 95% of all cases where this is happening

### Why is it needed?

If ws connection has been closed unexpectedly by the other end, we can't send a confirmation message, so it throws an error and crashes the strapi server

### How to test it?

- run a bunch of data transfers and interrupt them

Without this PR
- most of the time (but not always) you get an unhandled websocket connection closed error on the server and it shuts down

With this PR
- it should log an error and not crash the server

### Related issue(s)/PR(s)

DX-1900


packages/core/data-transfer/src/strapi/remote/handlers/pull.ts
```diff
@@ -167,6 +167,16 @@ export const createPullController = handlerControllerFactory<Partial<PullHandler
     const stream = this.streams?.[stage];
 
     const batchLength = () => Buffer.byteLength(JSON.stringify(batch));
+
+    const maybeConfirm = async (data: any) => {
+      try {
+        await this.confirm(data);
+      } catch (error) {
+        // Handle the error, log it, or take other appropriate actions
+        console.error('Error during confirmation:', error);
```
Convly, line 176: Could we make this a bit more specific? Like the fact it comes from DTS and killed it or smth?
Otherwise it might be hard for users to understand the source of this log

innerdvations: Oops, yeah, and it should also try to send it through strapi.log rather than console. Good catch!

innerdvations: Ok, added that!

Next step for the backlog: we need to make the diagnostics reporter also log warnings and errors to strapi logger whenever it's available; right now it doesn't, which is why I'm now logging to it directly like we do elsewhere

Convly (approved):

packages/core/data-transfer/src/strapi/remote/handlers/pull.ts
```diff
@@ -167,6 +180,20 @@ export const createPullController = handlerControllerFactory<Partial<PullHandler
...
 
     const batchLength = () => Buffer.byteLength(JSON.stringify(batch));
+
+    const maybeConfirm = async (data: any) => {
+      try {
+        await this.confirm(data);
+      } catch (error) {
+        // Handle the error, log it, or take other appropriate actions
+
+        strapi?.log.error(
```
jhoward1994, line 190: Does this need to be optional like this?
```suggestion
        strapi.log.error(
```

innerdvations: I wanted to be safe in case it's used in some context where it's not globally defined, but yeah, in the typical case it works

Convly (approved):

PR #22998, opened 2025-02-26
feat(relations): add save and publish actions to relations modal

markkaylor:
### What does it do?

- Adds the save and publish actions

### Why is it needed?

To save and publish

### How to test it?

Open the relation modal
Make a change
Save or publish

## TODO

- [x] Update e2e tests
- [x] Fetch the correct locale
- [x] Save the correct locale

### Known issue
- There is an issue where save reverts back to the original data, then updates with the newly saved data. Removing `resetForm` in UpdateAction fixes the issue but causes issues on the editview form. Still investigating. 

https://github.com/user-attachments/assets/76d1440e-bce0-4fc1-9a71-67392c887416

- A relation that has not been saved does not refresh its data when edited

### DIFF
--- a/packages/core/content-manager/admin/src/features/DocumentContext.tsx
+++ b/packages/core/content-manager/admin/src/features/DocumentContext.tsx
@@ -29,6 +29,7 @@ interface DocumentMeta {
 }
 
 interface DocumentContextValue {
+  rootDocumentMeta: DocumentMeta;
   document: ReturnType<UseDocument>;
   meta: DocumentMeta;
   changeDocument: (newRelation: DocumentMeta) => void;
@@ -70,6 +71,12 @@ const DocumentContextProvider = ({
     <DocumentProvider
       changeDocument={changeDocument}
       document={document}
+      rootDocumentMeta={{
+        documentId: initialDocument.documentId,
+        model: initialDocument.model,
+        collectionType: initialDocument.collectionType,
+        params: initialDocument.params,
+      }}
       meta={currentDocumentMeta}
     >
       {children}
--- a/packages/core/content-manager/admin/src/features/DocumentContext.tsx
+++ b/packages/core/content-manager/admin/src/features/DocumentContext.tsx
@@ -22,8 +22,8 @@ interface DocumentMeta {
    */
   collectionType: string;
   /**
-   * The parsed query params
-   * i.e. ?i18n=en parsed to { i18n: 'en' }
+   * Query params object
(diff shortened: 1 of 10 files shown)


markkaylor:
The database for the e2e tests was corrupt again. I fixed it to address the issues but some other tests are likely to fail on this run.

packages/core/content-manager/admin/src/features/DocumentContext.tsx
```diff
@@ -54,7 +60,11 @@ const DocumentContextProvider = ({
    * one of the root level document's relations.
    */
   const [currentDocumentMeta, changeDocument] = React.useState<DocumentMeta>(initialDocument);
-  const document = useDocument(currentDocumentMeta);
+  const params = React.useMemo(
```
markkaylor, line 63: I don't know if the useMemo is necessary or not

simotae14: If you remember we discussed regarding the removal of useMemo and useCallback if not strictly necessary so feel free to remove it

simotae14 (commented):
I think you already mentioned that on the original edit view the changes are not applied I mean you have not invalidated the cache

simotae14:
As you already said there is a problem invalidating the cache

https://github.com/user-attachments/assets/beaca5b5-9f07-4753-9e3e-6b68d6da32a0

By the way if you open, inside the modal, the document in the Edit view and you save changes these changes are reflected in the Edit view, so at least this part is automatically correct

simotae14:
Overall it seems working, there is the little refresh when you save that you already know

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -148,23 +146,92 @@ const RelationModalBody = ({ id }: RelationModalBodyProps) => {
...
+              documentResponse.components,
+              {
+                status: documentResponse.document?.status,
+                ...options,
+              }
+            );
+
+            return yupSchema.validate(values, { abortEarly: false });
+          }}
+          method="PUT"
```
simotae14, line 177: maybe you can add a TODO to change this value when we introduce the creation modal

simotae14 (commented):
overall the code looks good, there are some small behaviors you need to fix but the functionality seems almost ready

PR #23001, opened 2025-02-26
[V4] feat(upload): reduce load on strapi backend during multi image upload + validate file size before optimisations

Tyki:
Not sure where to put the target branch, picking v4 as it was the most obvious to me.

### What does it do?

This PR ensure that Strapi keeps working under low memory environment (such as 256 mb and lower) by uploading images sequentially.

It also fix a potential bug where the plugin start optimizing the image before checking the size of the image.

Some QoL is added in order to customize the transform quality level (not sure where to add some documentation about a new config parameter

### /!\ This is a UX modification /!\  

I would rather add a configuration for the plugin that is propagated to the admin UI but I prefer discuss it with you before going further.

### Why is it needed?

On low memory servers (such as K8S pods with 256mb), Strapi goes OOM during multi image upload due to multiple requests sent concurrently and sharp start optimizing all of the uploaded images

### How to test it?

Start Strapi in a docker with a memory limit such as 

```
version: "3"

services:
  strapi:
    build:
      dockerfile: Dockerfile
    command: yarn start
    ports:
      - "1337:1337"
    deploy:
      resources:
        limits:
          memory: 256M
```

and upload multiple images that are simple 5mb samples (took them from https://sample-videos.com/download-sample-jpg-image.php)

Constat that Strapi goes SIGKILL due to OOM

### Related issue(s)/PR(s)

This is technically not an issue so I didn't raised any issue about it. But if needed, I'm ok with it.

### DIFF
--- a/packages/core/upload/admin/src/components/AssetCard/UploadingAssetCard.jsx
+++ b/packages/core/upload/admin/src/components/AssetCard/UploadingAssetCard.jsx
@@ -36,6 +36,8 @@ export const UploadingAssetCard = ({
   onStatusChange,
   addUploadedFiles,
   folderId,
+  performUpload,
+  setCurrentAssetIndexUploading,
 }) => {
   const { upload, cancel, error, progress, status } = useUpload();
   const { formatMessage } = useIntl();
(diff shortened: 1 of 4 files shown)


Tyki:
WIll port this to V5

PR #23007, opened 2025-02-26
chore: add GitHub Action to list merged PRs between branches

Convly:
### What does it do?

Add an action and a workflow that consumes it, they allow listing the diff (in PRs) between two branches easily

### Why is it needed?

Remove the dependency on milestones

### How to test it?

We'll need to merge it before being able to test it. This PR is here mainly to validate the idea


innerdvations:
So the idea would be that instead of the auto-generated release notes, we would run this action and paste in the results?

.github/workflows/diff_prs.yml
```diff
@@ -0,0 +1,53 @@
+name: Diff as PRs
+
+on:
+  workflow_dispatch:
+    inputs:
+      base_branch:
```
alexandrebodin, line 6: Do we really need 2 inputs ? 

We could use the branch the job runs on (like we do for experimentals) as the ref. here we would always select a release branch.

And then we need the base branch that would be the previous version tag. 

do we have a way to make this into a select ? and side not do you support tags the same way you support branches (it's supposed to just be refs but just in case you need to test it)

.github/workflows/diff_prs.yml
```diff
@@ -0,0 +1,53 @@
...
+permissions:
+  contents: read
+
+jobs:
+  fetch-prs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
alexandrebodin, line 23: Is it a shallow checkout by default ?

.github/workflows/diff_prs.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Fetch Merged PRs Between Branches
+        uses: ./.github/actions/pr_diff
+        with:
+          base_branch: ${{ github.event.inputs.base_branch }}
+          target_branch: ${{ github.event.inputs.target_branch }}
+
+      - name: Format output
```
alexandrebodin, line 31: Why not do this part in the action itself ? or as a pr_diff_logger action

.github/actions/pr_diff/script.sh
```diff
@@ -0,0 +1,174 @@
...
+TARGET_BRANCH=""
+
+# GitHub repository owner
+OWNER="strapi"
+
+# GitHub repository name
+REPO="strapi"
+
+# Parse command-line arguments
+while getopts "b:t:o:r:" arg; do
```
alexandrebodin, line 16: is it necessary considering it's always passed as env vars ?

alexandrebodin (commented):
Mostly comments. My main concern is about the security/safety of using a shell script (input sanitization & co)  🤔

Convly:
> So the idea would be that instead of the auto-generated release notes, we would run this action and paste in the results?

Yeah, it would allow the QA team (and anyone) to automatically generate a report of every PRs that were included in the release (or between two branches, which can be helpful for tracking down a bug)

.github/workflows/diff_prs.yml
Convly: The goal of having two inputs was to allow for more precise diffs if needed (e.g. to track down where a bug is originating), I was hoping the input would be a select by default just like the publish/pre-release actions. (I've tried looking them up for specific options but haven't found anything.) If you have tips on that, they're welcomed :smiley: 

We could pass a default though

Yes it supports tags (which is super useful), I've been testing the script between `v5.10.3` and `releases/5.10.4` and it was working fine

.github/workflows/diff_prs.yml
Convly: I have no idea, the doc don't seem to mention it
https://github.com/actions/checkout
What difference does this make except for the execution time and resource used?

.github/actions/pr_diff/script.sh
Convly: It was initially a script I wrote to be run manually, and then thought about adapting it for GitHub actions. I can remove it, but it seemed harmless to keep it here just in case.

I can remove it if you want?

.github/workflows/diff_prs.yml
Convly: I wanted to keep the other action very generic (outputs json) and make the formatting in another step.

Putting all of this in a dedicated action seemed a bit overkill though, especially for a workflow that little.
Note: this is using the `$GITHUB_STEP_SUMMARY` variable which generates (supposedly :crossed_fingers: ) a standalone report for the workflow execution

Convly:
> My main concern is about the security/safety of using a shell script (input sanitization & co)

Could you be more specific, I'm not sure to get exactly what attack surface could be exposed here? :thinking: 
Technically the only inputs available are the refs (base/target) and are sent by the workflow dispatcher itself

.github/workflows/diff_prs.yml
alexandrebodin: Yeah that's what I was expecting sadly I don't know a way to dynamically display a list of branch in a select :/

.github/workflows/diff_prs.yml
alexandrebodin: Here is the relevant info in their doc and it's the good default. the diff is it's not gonna load the entire git history

  # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''

Convly: > ```
># Default: 1```
We should be fine by default then :slightly_smiling_face:

.github/workflows/diff_prs.yml
alexandrebodin: Ok, the 2 sentences seem a bit contradictory (generic action but not generic output) but it's fine 👍

.github/actions/pr_diff/script.sh
alexandrebodin: Yeah I don't know if it could have any side effect on how it's run in the CI so I would remove it. let's only keep code that runs

alexandrebodin:
> > So the idea would be that instead of the auto-generated release notes, we would run this action and paste in the results?
> 
> Yeah, it would allow the QA team (and anyone) to automatically generate a report of every PRs that were included in the release (or between two branches, which can be helpful for tracking down a bug)

-> But we are not going to replace the generated changelog with this output. it's purely for internal purposes to know in advance what will be in a release

Convly:
> > > So the idea would be that instead of the auto-generated release notes, we would run this action and paste in the results?
> > 
> > 
> > Yeah, it would allow the QA team (and anyone) to automatically generate a report of every PRs that were included in the release (or between two branches, which can be helpful for tracking down a bug)
> 
> -> But we are not going to replace the generated changelog with this output. it's purely for internal purposes to know in advance what will be in a release

Oh, yeah of course

alexandrebodin:
> > My main concern is about the security/safety of using a shell script (input sanitization & co)
> 
> Could you be more specific? I'm not sure to get exactly what attack surface could be exposed here 🤔 Technically the only inputs available are the refs (base/target) and are sent by the workflow dispatcher itself

we are allowing passing branches &  using a github token. how much of this could be used to fabricate invalid github urls that would expose sth 🤷

.github/actions/pr_diff/script.sh
Convly: I've removed the CLI arguments and only rely on env variables now

alexandrebodin (approved):
I can't say I love shell always feels unsafe but I can't find anything obvious

PR #20663, opened 2024-07-01
Prevent Vite hmr from Hijacking WS Upgrade Events

Convly:
### What does it do?

Introduce a dedicated HTTP server used by vite hmr, and reference it in the related vite configuration.

### Why is it needed?

[Vite hmr hijacks any WS upgrade events](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/ws.ts#L149). 

Since the current hmr config is re-using Strapi HTTP server, this behavior prevents Strapi data transfer requests from triggering remote transfers (handling custom WS connections (with policies, middleware, etc...)).

### How to test it?

Both remote data transfers and vite hmr should be working fine.

### Related issue(s)/PR(s)

fixes #20649

PR #23008, opened 2025-02-26
enhancement: allow multiple projects in develop mode by using next available port for vite

innerdvations:
### What does it do?

Finds an available port for vite to use when using `yarn develop` so that two develop mode strapi projects can be run on the same server.

I tried everything I could to make vite auto-select its own port and couldn't get it to work. I'm sure there has to be a way to do it, but for now this should do.

### Why is it needed?

currently you can't run `yarn develop` on two different strapi projects at once

### How to test it?

run two different strapi servers with `yarn develop`
make changes to admin files on each server and make sure the appropriate project hot-reloads

### Related issue(s)/PR(s)


innerdvations:
@derrickmehaffy happy birthday

derrickmehaffy:
> @derrickmehaffy happy birthday

Your about a month and a half late but I won't complain.

THANK YOU SO MUCH!

packages/core/strapi/src/node/vite/watch.ts
```diff
@@ -12,6 +13,33 @@ interface ViteWatcher {
 }
 
 const HMR_DEFAULT_PORT = 5173;
+const MAX_PORT_ATTEMPTS = 30;
+
+const findAvailablePort = (
```
Convly, line 18: I kinda hate we have to do this. But if we're sure we don't have any solution then why not

Out of curiosity, wouldn't that create less noise and consume less resources to use net.Socket and try establishing connections rather than create actual servers?

innerdvations: Yeah, I hate it too. If it affected production I wouldn't do it, but since it's just dev mode I think it's acceptable.

This create server method seemed the most reliable since it actually binds to a port it's certain it's free instead of just being somewhat sure. And slightly less complex because trying to connect to a port requires more listeners. But yeah, it is slightly more "noisy" as you said.

Convly: It would be nice to start unit testing all this stuff with server mocks etc... even though for now we don't have any :thinking:
This is the kind of behavior that I don't want anyone to break because they don't have context or smth :confused:

innerdvations: Next week is a testing blitz... maybe you can come up with something 😆

jhoward1994:
I guess you have tried this but if we added a port option to the server portion of the config in `resolveDevelopmentConfig` vite should automatically try the next ports?

https://vite.dev/config/server-options#server-port

innerdvations:
> I guess you've tried this but if we added a port option to the server portion of the config in `resolveDevelopmentConfig` vite should automatically try the next ports?
> 
> [vite.dev/config/server-options#server-port](https://vite.dev/config/server-options#server-port)

Yeah, that's what I was trying to get to work, but there is an issue with the way we're creating this external server first that we pass to vite to bind to. I actually don't really understand why this server is necessary though, maybe it's possible to remove it.

Convly (approved):

innerdvations:
Just noticed where the hmr server was introduced and why: https://github.com/strapi/strapi/pull/20663

PR #23021, opened 2025-02-27
ci: match v4 tooling packages to v5

innerdvations:
### What does it do?

- use the same versions of lerna and @actions/github as the `develop` branch
- ignore .nx in prettier

### Why is it needed?

to eliminate several packages with vulnerabilities from the yarn.lock; none of them actually affect anything, but it triggers dependabot alerts and vulnerability warnings 

### How to test it?

### Related issue(s)/PR(s)


Convly (commented):
Is there any breaking change impacting us in this?

innerdvations:
> Is there any breaking change impacting us in this?

There shouldn't be, as long as the monorepo builds. But there does seem to be some sort of conflict with nx in github. I'll take a look later.

Convly (approved):

PR #23023, opened 2025-02-27
security(deps): upgrade vite to 5.4.11

innerdvations:
### What does it do?

- upgrades vite to 5.4.11
- bump /docs pinned version of dompurify to 3.2.4

### Why is it needed?

CVE-2025-24010

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -3244,6 +3244,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@types/trusted-types@npm:^2.0.7":
+  version: 2.0.7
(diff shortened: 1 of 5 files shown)


innerdvations:
The CI keeps randomly encountering a weird npm install error with the nx package, seems like https://github.com/nrwl/nx/issues/29481 could be related. I'll see if it has to do with something in this PR or a general problem.

innerdvations:
Oops, sorry, not sure how that happened! I will see about releasing a hotfix today, although it's a fairly low vulnerability in our case since it only affects develop mode and not production, so it may not happen until next week.

PR #23024, opened 2025-02-27
feat(preview): refresh iframe when strapi updates

markkaylor:
### What does it do?

- Uses [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to inform the preview iframe that strapi was updated

### Why is it needed?

- To refresh the preview content without having to reload the page

### How to test it?

Go to a preview page. Make an update and click save or publish. You should see the data updated in the preview iframe. It should the correct data based on the document status.

### TODO

- [ ] Update the test env app to have an iframe that display content to run tests against. Not sure how that will go with testing in an iframe 🤔

### DIFF
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 import { useParams } from 'react-router-dom';
 
 // @ts-ignore
@@ -14,6 +14,20 @@ const PreviewComponent = () => {
     params: { locale, status },
     collectionType,
   });
+
+  React.useEffect(() => {
+    const handleStrapiUpdate = (event) => {
+      if (event.data?.type === 'strapiUpdate') {
+        window.location.reload();
+      }
+    };
+
+    window.addEventListener('message', handleStrapiUpdate);
+
+    return () => {
+      window.removeEventListener('message', handleStrapiUpdate);
+    };
+  }, []);
 
   return (
     <Layouts.Root>
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -8,7 +8,7 @@ import { Grid, Flex, Typography, JSONInput } from '@strapi/design-system';
 
 const PreviewComponent = () => {
   const { uid: model, documentId, locale, status, collectionType } = useParams();
-  const { document } = useDocument({
+  const { document, refetch } = useDocument({
     model,
     documentId,
     params: { locale, status },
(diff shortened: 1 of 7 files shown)


examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -15,6 +15,20 @@ const PreviewComponent = () => {
     collectionType,
   });
 
+  React.useEffect(() => {
+    const handleStrapiUpdate = (event) => {
```
markkaylor, line 19: Add type

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
remidej, line 1: I'm guessing you're moving this code because you think the side editor isn't really preview content. But aren't you removing the future flag behavior by enforcing the side editor here?

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -15,6 +15,20 @@ const PreviewComponent = () => {
...
+      if (event.data?.type === 'strapiUpdate') {
+        window.location.reload();
+      }
```
remidej, line 22: this works but it would be nice to demo the cache invalidation method here to avoid a slow hard refresh.

I think we could update useDocument to expose the refetch function and call it here

packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -183,6 +184,9 @@ const UnstablePreviewHeader = () => {
     documentId: document.documentId,
     document,
     meta,
+    onPreview: () => {
+      iframeRef?.current?.contentWindow?.postMessage({ type: 'strapiUpdate' });
+    },
   } satisfies DocumentActionProps;
```
remidej, line 190: Could you rename this object something like documentActionProps? Otherwise just reading props I'd think it's the component's props (UnstablePreviewHeader)

remidej (commented):
I really like your onPreview solution

packages/core/content-manager/admin/src/preview/components/PreviewContent.tsx
markkaylor: I did the `UnstablePreviewContent` inline so it was easier to pass the iframe ref. I think I'm still in the future flag condition though and otherwise should be returning the old `PreviewContent` component. I'll double check to make sure I didn't do something stupid.

markkaylor: It looks like everything is still working as expected with the future flag

remidej (approved):
It's beautiful. Maybe we should dispatch the event on publication as well?

markkaylor:
> It's beautiful. Maybe we should dispatch the event on publication as well?

It's not that clear in the diff but it should be dispatched for both UpdateAction and PublishAction. Did you not see publishing changes working the same?

Screenshot 2025-03-03 at 16 21 23

remidej:
@markkaylor you're right everything's good, the test i did was dumb

PR #23028, opened 2025-02-27
feat(content-manager): support right click open in content list

jorrit:
### What does it do?

This change allows users to right click a cell in the content list and open the edit page in a new browser window / tab.

### Why is it needed?

It's sometimes handy to open a bunch of content items in browser tabs to edit them.

### How to test it?

Open a content list page, right click an item and observe that the browser offers open in new tab etc.

### Related issue(s)/PR(s)

It was mentioned somewhere in an issue but I can't find it anymore.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -210,11 +210,13 @@ const ListViewPage = () => {
         defaultMessage: 'Untitled',
       });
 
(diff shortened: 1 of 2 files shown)


mathildeleg:
Hi there, thanks so much for your contribution ✨
I took at look at your PR but I think a better UX enhancement would be if the full row was right-clickable, rather than just if the content-type has a textfield, as some content-types may not have some and thus the right-clicking feature wouldn't be available, making it a bit confusing to the user!
I created a PR for this and thus will close yours, but thanks so much for bringing this up 🤗

jorrit:
That's also great, of course. Thanks for considering.

PR #23029, opened 2025-02-27
chore(content-manager): fix deprecation in Tooltip usage

jorrit:
### What does it do?

Replace references to the deprecated `description` property of the `Tooltip` component with `label`, as suggested.

### Why is it needed?

Forward compatibility with newer design-system releases that remove the deprecated `description` prop.

### How to test it?

Visit any page that uses a tooltip and observe that they still work.


Adzouz (approved):
LGTM, thanks a lot 🙏 I updated the branch to resolve conflicts. Will merge as soon as the pipelines pass!

PR #23035, opened 2025-02-28
fix: marketplace remove incorrect query param

Bassel17:
### What does it do?

Remove version query, that send the current strapi version to the marketplace api,

### Why is it needed?

The implementation of this version query is not correct

### How to test it?

- everything should be working the same when loading the marketplace
-check the requested marketplace api it should not contain `version=CURRENT_STRAPI_VERSION` or something like this 
`https://market-api.strapi.io/providers?pagination%5Bpage%5D=1&pagination%5BpageSize%5D=24&search=&version=5.10.3`

### DIFF
--- a/packages/core/admin/admin/src/pages/Marketplace/MarketplacePage.tsx
+++ b/packages/core/admin/admin/src/pages/Marketplace/MarketplacePage.tsx
@@ -90,7 +90,7 @@ const MarketplacePage = () => {
     possibleCollections,
     possibleCategories,
     pagination,
-  } = useMarketplaceData({ npmPackageType, debouncedSearch, query, tabQuery, strapiVersion });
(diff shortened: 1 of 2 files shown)


jhoward1994:
You mention on JIRA that the expected format is:

`https://market-api.strapi.io/plugins?pagination%5Bpage%5D=1&pagination%5BpageSize%5D=24&search=&version[]=v5`

Just to check should we be adjusting the query string to this rather than removing the whole thing?

Bassel17:
> You mention on JIRA that the expected format is:
> 
> `https://market-api.strapi.io/plugins?pagination%5Bpage%5D=1&pagination%5BpageSize%5D=24&search=&version[]=v5`
> 
> Just to check should we be adjusting the query string to this rather than removing the whole thing?

that would filter only the v5 plugins, am still waiting for Marco's plan on that

PR #23047, opened 2025-03-03
chore: remove or upgrade incorrect packages

innerdvations:
### What does it do?

- removes the `@strapi/pack-up` devDep in `packages/core/upload` because it's no longer needed
- upgrades `@actions/core` to 1.11.1
- upgrade and standardize `nx` and `nx/js` to 20.4.6 including the pinned version used by `lerna` so that they all use the same version

Doing them all together here because they are quite simple and they would cause merge conflicts if done separately since they affect several of the same subdeps

### Why is it needed?

- pack-up wasn't used, it was just a line we forgot to move when switching to rollup
- bumps the remaining vulnarble pinned usage of `serialize-javascript` to a non-vulnerable version
- partial fix for removing all versions of undici that were vulnerable to security issues
- fixed the flaky github CI caused by mismatched version of `nx`

NOTE: there are not currently any user-reaching usages of vulnerable versions of `undici`, this is just in tooling, but we're trying to get rid of those in order to have a clean yarn.lock
 
### How to test it?

if everything builds and the CI runs fine, it should work, these are all tooling packages

### Related issue(s)/PR(s)

DX-1917


Bassel17 (approved):
LGTM

PR #23052, opened 2025-03-03
chore: add documentId to tracking events

remidej:
### What does it do?

adds documentId to some tracking events
### Why is it needed?

telemetry for the product team to understand which features lead to more content being published
### How to test it?

- publish content
- update content
- create content

for each action inspect the devtools, see the tracking request, you should see the documentId in the eventProperties object

### DIFF
--- a/packages/core/admin/admin/src/features/Tracking.tsx
+++ b/packages/core/admin/admin/src/features/Tracking.tsx
@@ -131,7 +131,6 @@ interface EventWithoutProperties {
     | 'didNotCreateFirstAdmin'
     | 'didNotSaveComponent'
     | 'didPluginLearnMore'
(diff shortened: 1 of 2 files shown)


Marc-Roig:
how this will be used? is the intention to track how many diferent documents are published?

remidej:
@Marc-Roig I believe the intention is to check which features lead to more content being published. For example, does opening preview on an entry lead to it being published quicker

Marc-Roig:
@remidej how does the documentId enable such thing?

remidej:
@Marc-Roig we're tracking preview opens via willNavigate, it sends the whole URL, then we can parse the documentId from that URL on the other side

simotae14 (approved):
👏🏽

PR #23059, opened 2025-03-04
fix: add origin to preview message

remidej:
### What does it do?

- adds origin to preview postMessage events
- also uses the new icon for the side editor toggle
### Why is it needed?

fix cross-origin errors


markkaylor (dismissed):
Woah, great catch and a clever fix 🚀

markkaylor (approved):

PR #23068, opened 2025-03-05
chore(deps): upgrade esbuild-loader to 4.3.0

innerdvations:
### What does it do?

- removes `esbuild` dependency because we don't actually use it for anything
- upgrades `esbuild-loader` to 4.3.0 to have it depend on esbuild 0.25.0 which fixes GHSA-67mh-4wv8-2f99

### Why is it needed?

Partial resolution to warnings from GHSA-67mh-4wv8-2f99 although Strapi does not appear to be vulnerable

Example plugin template still includes reference to @strapi/pack-up which includes a vulnerable version, but the vulnerable functionality does not appear to be used in @strapi/pack-up either

### How to test it?

### Related issue(s)/PR(s)


Convly:
What areas should be tested in priority ? @innerdvations

innerdvations:
> What areas should be manually tested in priority in your opinion ? @innerdvations

The only thing that would have potentially changed is the webpack build for production, which I believe already has some issues with it, so I would say confirm that it still builds the same as without this PR.

Other than that, I really don't know why "esbuild" was in the deps list, maybe we used it at some point and forgot to remove it. I can't think of any way it could have side effects, so as long as everything still builds ok (`build`, `develop`, etc) it should be fine.

Convly (approved):

faduxw:
@innerdvations could you please also remove this dependency on v4.25

PR #23075, opened 2025-03-06
test(e2e): rbac roles

Convly:
### What does it do?

Add RBAC-related e2e tests to
- Create a role
- See a role
- Delete a role
- Edit a role
- Assign a role to a user
- All of the above in one scenario

### Why is it needed?

Improve the product stability

### How to test it?

```bash
yarn run test:e2e --domains=settings -- rbac --project=chromium
```


innerdvations (approved):
LGTM!

PR #23076, opened 2025-03-06
enhancement(ctb): add aria data to relations buttons

innerdvations:
### What does it do?

Adds aria attributes to the relations buttons in the CTB

NOTE: I accidentally merged [some tests](https://github.com/strapi/strapi/pull/23078) into this that were dependant on this PR and I had intended to wait until this was merged. But they're just tests and technically do test this functionality 😆 and they've already been reviewed [here](https://github.com/strapi/strapi/pull/23078)

### Why is it needed?

There was no way for screen readers (or Playwright) to identify which button was which, nor to detect which button was currently active. This way we tell screen readers to ignore the svg and read the new label instead.

### How to test it?

inspect element on the ctb relations buttons as you use them and see the labels

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/Relation/RelationNaturePicker/RelationNaturePicker.tsx
+++ b/packages/core/content-type-builder/admin/src/components/Relation/RelationNaturePicker/RelationNaturePicker.tsx
@@ -108,8 +108,11 @@ export const RelationNaturePicker = ({
                     }}
                     padding={2}
                     type="button"
+                    aria-label={formatMessage({ id: getTrad(`relation.${relation}`) })}
+                    aria-pressed={relationType === relation}
+                    data-relation-type={relation}
                   >
-                    <Asset key={relation} />
+                    <Asset key={relation} aria-hidden="true" />
                   </IconWrapper>
                 );
               })}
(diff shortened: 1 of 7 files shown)


packages/core/content-type-builder/admin/src/components/Relation/RelationNaturePicker/RelationNaturePicker.tsx
```diff
@@ -108,8 +108,11 @@ export const RelationNaturePicker = ({
                     }}
                     padding={2}
                     type="button"
+                    aria-label={formatMessage({ id: getTrad(`relation.${relation}`) })}
```
innerdvations, line 111: This actually leads to a duplicate of "has many" where one is the inverted has many, but I didn't want to add a new translation for it. If someone has a good idea let me know -- like if we have another string somewhere that we can use that explains it better

jhoward1994 (approved):
This look good, thanks!

Just wondering if we want to do anything about the sonar cloud duplicated lines flag?

It looks like we could lift stuff out as common in
- tests/e2e/tests/content-type-builder/collection-type/create-collection-type.spec.ts
- tests/e2e/tests/content-type-builder/components/create-components.spec.ts
- tests/e2e/tests/content-type-builder/single-type/create-single-type.spec.ts

innerdvations:
> This look good, thanks!
> 
> Just wondering if we want to do anything about the sonar cloud duplicated lines flag?
> 
> It looks like we could lift stuff out as common in
> 
> * tests/e2e/tests/content-type-builder/collection-type/create-collection-type.spec.ts
> * tests/e2e/tests/content-type-builder/components/create-components.spec.ts
> * tests/e2e/tests/content-type-builder/single-type/create-single-type.spec.ts

Since they're tests, I was going to leave them repetitive like that. Maybe after the CTB refactor we can see if they're still similar enough to DRY them a bit

PR #23074, opened 2025-03-06
feat(relations): add back button history

markkaylor:
### What does it do?

Maintains document history in DocumentContext

### Why is it needed?

So we can use the back button in the relation modal

### How to test it?

Go to an entry with relation, that has a relation, that has a relation...or something like that
Click the relation, then its relation, then its relation.
The go back, back, back.
You should be where you started.
Going back with with unsaved changes should trigger a confirm dialog

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #23077, opened 2025-03-06
feat(relations): add back button history

simotae14:
### What does it do?

Maintains document history in DocumentContext

### Why is it needed?

So we can use the back button in the relation modal

### How to test it?

Go to an entry with relation, that has a relation, that has a relation...or something like that
Click the relation, then its relation, then its relation.
The go back, back, back.
You should be where you started.
Going back with with unsaved changes should trigger a confirm dialog

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
@@ -57,6 +57,19 @@ const CustomModalContent = styled(Modal.Content)`
   max-height: 100%;
 `;
 
+interface RelationModalContextValue {
+  parentModified: boolean;
+  depth: number;
+}
+
+const [RelationModalProvider, useRelationModal] = createContext<RelationModalContextValue>(
+  'RelationModal',
+  {
+    parentModified: false,
+    depth: 0,
+  }
+);
+
 const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalProps) => {
   const navigate = useNavigate();
   const { pathname, search } = useLocation();
@@ -71,16 +84,38 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
   );
   const currentDocumentMeta = useDocumentContext('RelationModalBody', (state) => state.meta);
   const changeDocument = useDocumentContext('RelationModalBody', (state) => state.changeDocument);
+  const documentHistory = useDocumentContext('RelationModalBody', (state) => state.documentHistory);
+  const setDocumentHistory = useDocumentContext(
+    'RelationModalBody',
+    (state) => state.setDocumentHistory
+  );
 
   const [isConfirmationOpen, setIsConfirmationOpen] = React.useState(false);
-  const [isNavigating, setIsNavigating] = React.useState(false);
+  const [actionPosition, setActionPosition] = React.useState<'cancel' | 'back' | 'navigate'>(
+    'cancel'
+  );
 
   const [isModalOpen, setIsModalOpen] = React.useState(false);
+  // NOTE: Not sure about this relation modal context, maybe we should move this to DocumentContext?
   // Get parent modal context if it exists
   const parentContext = useRelationModal('RelationModalWrapper', (state) => state);
-  const isNested = parentContext.depth > 0;
-  // Track this modal's depth
-  const depth = isNested ? parentContext.depth + 1 : 1;
+  // Get depth of nested modals
+  const depth = parentContext ? parentContext.depth + 1 : 0;
+  // Check if this is a nested modal
+  const isNested = depth > 0;
+
+  const addDocumentToHistory = (document: DocumentMeta) =>
+    setDocumentHistory([...documentHistory, document]);
+
+  const getPreviousDocument = () => {
+    if (documentHistory.length === 0) return undefined;
+
+    const lastDocument = documentHistory[documentHistory.length - 1];
+
+    setDocumentHistory([...documentHistory].slice(0, documentHistory.length - 1));
+
+    return lastDocument;
+  };
 
   const handleToggleModal = () => {
     if (isModalOpen) {
@@ -92,6 +127,10 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
       };
       // Change back to the root document
       changeDocument(document);
+      // Reset the document history
+      setDocumentHistory([]);
+      // Reset action position
+      setActionPosition('cancel');
       // Read from cache or refetch root document
       triggerRefetchDocument(
         document,
@@ -125,8 +164,13 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
   };
 
   const handleConfirm = () => {
-    if (isNavigating) {
+    if (actionPosition === 'navigate') {
       handleRedirection();
+    } else if (actionPosition === 'back') {
+      const previousRelation = getPreviousDocument();
+      if (previousRelation) {
+        changeDocument(previousRelation);
+      }
     } else {
       handleToggleModal();
     }
@@ -150,6 +194,9 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
       }}
     >
       {({ modified, isSubmitting, resetForm }) => {
+        // We don't count the root document, so history starts after 1
+        const hasHistory = documentHistory.length > 1;
+
         return (
           <RelationModalProvider parentModified={modified} depth={depth}>
             <Modal.Root
@@ -177,6 +224,10 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
                         if (modified && !isSubmitting) {
                           setIsConfirmationOpen(true);
                         } else {
+                          // Add current relation to history before opening a new one
+                          if (currentDocumentMeta && Object.keys(currentDocumentMeta).length > 0) {
+                            addDocumentToHistory(currentDocumentMeta);
+                          }
                           handleToggleModal();
                         }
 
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -177,6 +224,10 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
                         if (modified && !isSubmitting) {
                           setIsConfirmationOpen(true);
                         } else {
+                          // Add current relation to history before opening a new one
+                          if (currentDocumentMeta && Object.keys(currentDocumentMeta).length > 0) {
```
simotae14, line 228: Probably we can avoid to check if the currentDocumentMeta length is undefined because it needs to be always defined (https://github.com/strapi/strapi/pull/23074#discussion_r1982983064)
```suggestion
                          if (currentDocumentMeta) {
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -71,16 +84,38 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
...
+  const setDocumentHistory = useDocumentContext(
+    'RelationModalBody',
+    (state) => state.setDocumentHistory
+  );
 
   const [isConfirmationOpen, setIsConfirmationOpen] = React.useState(false);
-  const [isNavigating, setIsNavigating] = React.useState(false);
+  const [actionPosition, setActionPosition] = React.useState<'cancel' | 'back' | 'navigate'>(
+    'cancel'
+  );
```
simotae14, line 96: We can discuss a different naming (https://github.com/strapi/strapi/pull/23074#discussion_r1982983572)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -198,8 +249,18 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
...
+                        disabled={!hasHistory}
+                        onClick={() => {
+                          setActionPosition('back');
+                          if (modified && !isSubmitting) {
+                            setIsConfirmationOpen(true);
+                          } else {
+                            const previousRelation = getPreviousDocument();
+                            if (previousRelation) {
+                              changeDocument(previousRelation);
+                            }
```
simotae14, line 261: this code is used in different places so it could be a function (https://github.com/strapi/strapi/pull/23074#discussion_r1982984288)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -71,16 +84,38 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
...
+  );
 
   const [isConfirmationOpen, setIsConfirmationOpen] = React.useState(false);
-  const [isNavigating, setIsNavigating] = React.useState(false);
+  const [actionPosition, setActionPosition] = React.useState<'cancel' | 'back' | 'navigate'>(
+    'cancel'
+  );
 
   const [isModalOpen, setIsModalOpen] = React.useState(false);
+  // NOTE: Not sure about this relation modal context, maybe we should move this to DocumentContext?
```
simotae14, line 99: We can discuss if we want to move it or not (https://github.com/strapi/strapi/pull/23074#discussion_r1982985551)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -71,16 +84,38 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
...
-  const isNested = parentContext.depth > 0;
-  // Track this modal's depth
-  const depth = isNested ? parentContext.depth + 1 : 1;
+  // Get depth of nested modals
+  const depth = parentContext ? parentContext.depth + 1 : 0;
+  // Check if this is a nested modal
+  const isNested = depth > 0;
+
+  const addDocumentToHistory = (document: DocumentMeta) =>
+    setDocumentHistory([...documentHistory, document]);
```
remidej, line 108: I always prefer using the callback here to avoid potential stale closure issues
```suggestion
    setDocumentHistory(prev => [...prev, document]);
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -71,16 +84,38 @@ const RelationModalWrapper = ({ relation, triggerButtonLabel }: RelationModalPro
...
+
+  const getPreviousDocument = () => {
+    if (documentHistory.length === 0) return undefined;
+
+    const lastDocument = documentHistory[documentHistory.length - 1];
+
+    setDocumentHistory([...documentHistory].slice(0, documentHistory.length - 1));
```
remidej, line 115: same think about the callback. But also why does a getter function set some state? This isn't what I would expect

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
remidej: I'm having a hard time understanding this prop. It's to adjust the behavior of the modal's back button depending on the circumstances, right? If so I understand cancel (close the modal), but I don't understand the difference between back and navigate

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
remidej, line 1: The state management with a documentHistory and its array slicing + the "actionPosition" thing is a bit complex. I think this could be a use case for useReducer instead of useState, similar to the admin's global BackButton component. I see 2 benefits:

- Abstract the complexity away from the component code. The handlers here are getting messy, with a reducer we could just dispatch actions
- Combine actionPosition and documentHistory into the same reducer state. I feel like the two belong together, so modifying both at the same time via a single action dispatch could be neat, and they'd always be in sync

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
markkaylor: > It's to adjust the behavior of the modal's back button depending on the circumstances, right?
Not exactly. Depending on what action is performed (clicked cancel or close, clicked open full page, clicked back button) we need to provide the correct function to the dialogs onConfirm handler. This system isn't great I agree, but the handleConfirm does this to get the right function:

```tsx
  const handleConfirm = () => {
    if (actionPosition === 'navigate') {
      handleRedirection();
    } else if (actionPosition === 'back') {
      const previousRelation = getPreviousDocument();
      if (previousRelation) {
        changeDocument(previousRelation);
      }
    } else {
      handleToggleModal();
    }
  };
```

cancel => close the modal
navigate => change route
back => change document but leave the modal open

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
markkaylor: Good question, we should improve this

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
markkaylor: Sounds like a lot of good ideas maybe we can work on it together

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
remidej: that makes more sense, how about something dialogConfirmAction then?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
remidej: I would like that! tomorrow morning?

markkaylor: Yep I'll ping y'all around 10:30ish. (cc @simotae14)

simotae14:
https://github.com/strapi/strapi/pull/23077#discussion_r1983673398
It handles the actions to trigger in case you click the Confirm button in the Confirmation dialog, so in case you "navigate" basically you are gonna open the relation in full page (its edit view page), in case of "cancel" you are closing the modal so the confirm action toggle the modal and in case of "back" you are clicking the back button and when you confirm then you are going back to the previous relation you have opened in the modal

Probably we need to add a comment to explain better this state

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
simotae14: I removed setDocumentHistory from the getter because it doesn't reflect what we want the getter to do, return the last element in history and I have created a single line function to remove the last item from history (I preffered to define a function instead of simply call setDocumentHistory([...documentHistory].slice(0, documentHistory.length - 1)); because the name of he function explains what we are trying to do)

markkaylor (approved):

remidej (approved):

PR #23078, opened 2025-03-06
test(e2e): add test for relations in ctb

innerdvations:
### What does it do?

Adds e2e tests for collection type, component, and single type:
- for every attribute, test advanced settings (required)
- add all possible relation attributes for the type
  - confirm names are pre-filled correctly

TODO in a later PR:

- [ ] check error cases for relations
- [ ] test the content-manager functionality for created relations

### Why is it needed?

we didn't have these tests

we'd previously had the advanced settings tests but they were misplaced into content-manager tests and got refactored out, so I've added them back in here where they belong

### How to test it?

they should all pass

### Related issue(s)/PR(s)

DX-1817


tests/e2e/utils/content-types.ts
```diff
@@ -16,8 +16,30 @@ export interface AddAttribute {
...
+    target: {
+      name?: string;
+      select?: string;
+    };
+  };
 }
 
+// keys are the relation types used by the RelationNaturePicker component
+// locatorText is the text that should be displayed for the relation type
+// inverse is used to identify the inverse relation type(s)
```
innerdvations, line 30:
```suggestion
// invered is used to identify the inverse relation type(s)
```

Convly (approved):

tests/e2e/tests/content-type-builder/single-type/edit-single-type.spec.ts
```diff
@@ -84,6 +107,53 @@ test.describe('Edit single type', () => {
     await expect(page.getByRole('heading', { name: ctName })).toBeVisible();
   });
 
+  /**
+   * TODO: This test is flaky likely due to an actual display bug
+   * where specific circumstances (demonstrated here) cause a modal to close/reopen on the first click
+   * instead of triggering the submit
```
Convly, line 113: That's weird, any workaround?

innerdvations: I just copied it from the edit-collection-type test, I'm not sure if it's still valid. And it wasn't actually flaky as-written because I think the utilities take it into account and ensure that the finish button does actually get clicked.
 
I can investigate later and remove the comments if it looks like it's not an issue anymore.

Issue #20853, opened 2024-07-19
[Strapi v5 RC] Assets Cannot be Moved to Folders


burakozdemir32:
### Required System information

- **Node.js version**: 18.17.0
- **NPM version**: yarn/1.22.19
- **Strapi version**: 5.0.0-rc.2
- **Database**: sqlite
- **Operating system**: darwin-arm64
- **Is your project Javascript or Typescript**: Typescript

### Describe the bug

Assets cannot be moved to folders. When attempting to move an asset, the destination folder options are not displayed correctly.

### Steps to reproduce the behavior

1. Open the Media Library.
2. Select an asset to move.
3. Click on the "Move" button.
4. Observe that the destination folder options are not correctly displayed.

### Expected behavior

The destination folder options should be displayed correctly, allowing for the asset to be moved to the selected folder.

### Screenshots

<img width="1557" alt="Screenshot 2024-07-19 at 16 25 33" src="https://github.com/user-attachments/assets/0d595656-18f8-42e8-9618-3a0c954af6f0">

### Code snippets

N/A

### Additional context

N/A

derrickmehaffy:
unable to reproduce on rc.6

PR #23101, opened 2025-03-10
fix(upload): empty folder list in root view

noreiller:
In the media library, the dropdown used to move assets was not displaying the children folders. This was due to mixed values (`null` and `''`) while the check was using a strict comparison.

This PR fixes the issue by only using an empty string. I chose to not use the `null` value since it's also used in react-select which does not accept it.

### What does it do?

It changes the default value of a folder list to an empty string instead of `null`.

### Why is it needed?

In the root view of the media library, it's not possible to use the bulk move feature since no folder structure is displayed on the dropdown.

### How to test it?

- In DEV env, run the get started example
- Go to the media library
- Add an asset (if not yet added)
- Add a folder
- Select the asset and click on the "move" button
- See the modal with the bulk move options
- PREVIOUS behaviour: the dropdown was not proposing the newly added folder
- NEW behaviour: the dropdown is proposing the newly added folder

### Related issue(s)/PR(s)

#20853

### DIFF
--- a/packages/core/upload/admin/src/components/SelectTree/SelectTree.tsx
+++ b/packages/core/upload/admin/src/components/SelectTree/SelectTree.tsx
@@ -18,7 +18,7 @@ import { getValuesToClose } from './utils/getValuesToClose';
 const hasParent = (option: FlattenedNode<string | number | null>) => !option.parent;
 
 export type OptionSelectTree = {
-  value: string | number | null;
+  value: string | number;
   label?: string;
   children?: OptionSelectTree[];
 };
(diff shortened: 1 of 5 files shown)


Adzouz:
Hi @noreiller, thanks a lot for your contribution 🚀
The issue related to this PR was closed so I'm closing this PR too.
Don't hesitate to let us know if anything or if the issue is still happening 🙏

noreiller:
@Adzouz I added the relation issue since it seemed to be a regression.

Adzouz:
Hey @noreiller, if you're talking about https://github.com/strapi/strapi/issues/20853, it was closed. Are you still able to reproduce the bug is the latest version of Strapi (v5.27.0)?
Thanks 🙏

PR #23105, opened 2025-03-10
security(deps): upgrade axios to 1.8.2 for CWE-918

innerdvations:
### What does it do?

upgrades axios to 1.8.2

### Why is it needed?

CWE-918

### How to test it?

used in too many places to list

The [changelog](https://github.com/axios/axios/releases) lists some breaking changes in 1.8.0 but they don't seem to affect us; but if someone who knows more about where and how axios is used can confirm, please do!

### Related issue(s)/PR(s)

DX-1927

### DIFF
--- a/packages/cli/cloud/package.json
+++ b/packages/cli/cloud/package.json
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@strapi/utils": "5.11.0",
-    "axios": "1.7.4",
+    "axios": "1.8.2",
     "boxen": "5.1.2",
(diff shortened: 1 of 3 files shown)


markkaylor (commented):
Looking at the breaking changes listed for 1.8.0 I think we should be ok, but just wondering should we merge this after the release branch lock and live with it for a week? That way we have a week to ensure there is no unexpected behavior? Otherwise happy to approve if that feels like playing it way too safe.

markkaylor (approved):

PR #23106, opened 2025-03-10
security(deps): upgrade axios to 1.8.2 for CVE-2025-27152

innerdvations:
### What does it do?

upgrades axios to 1.8.2

### Why is it needed?

CVE-2025-27152

### How to test it?

used in too many places to list

The [changelog](https://github.com/axios/axios/releases) lists some breaking changes in 1.8.0 but they don't seem to affect us; but if someone who knows more about where and how axios is used can confirm, please do!

### Related issue(s)/PR(s)

DX-1927


innerdvations:
@Marc-Roig it was impossible to run the CI on this without fixing nx, this is related to the issue we had to do on the release branch

Marc-Roig (approved):

PR #23120, opened 2025-03-12
chore: improve uniqueness of deviceId + userId

Mcastres:
### What does it do?

This PR enhances the uniqueness of `deviceId + userId` in Strapi-created projects. While analytics remain fully **anonymous**, they play a crucial role in understanding usage trends and improving the product. By refining this tracking, we can gain better insights into adoption, feature engagement, and overall growth—helping us make more informed decisions for the Strapi ecosystem.

If you have any concerns or suggestions, feel free to share your thoughts!

### Why is it needed?

This PR solves two **anonymous** tracking issues that we are facing:

- The deviceId we compute from the machine id are not 100% unique.
- The userId we generate are not really associated to a single project which is an issue.

#### deviceId

We use the [node-machine-id](https://github.com/automation-stack/node-machine-id) package to generate the deviceId of the first anonymous events we send. **However**, as the package mentions: `Image-based environments have usually the same machine-id`. This is problematic when trying to match the _completely_ anonymous events and the authenticated but still anonymous events later on.

**Solution**: Generate a hash of machineId + projectId (`strapi.uuid`) to get a 100% unique deviceId. In case the machineId can’t be found or if the projectId doesn’t exists in the package.json, we create a random UUID for the deviceId but, we would need to keep it for later, at least, until we send the first authenticated _(again, still anonymous)_ event so that we can match the original random deviceId to the user events later. The idea proposed in this PR for this scenario is to save it in the package.json so it can be used by the telemetry instance later.

#### userId

(EDIT: Our tracking system is actually already doing that so I removed the implementation below)

Currently, the same user on different projects will be represented by the same userId (hashed email). We want to change that, we would like to get a different userId for each project the user belongs to.

**Solution:**  Add the projectId (`strapi.uuid`) to the current hash that we create for generating the userId to get a 100% unique userId if it exists of course. Otherwise, nothing can really help us making it unique, so we leave it like it is right now.

### How to test it?

I intentionally left some console.logs to make testing easier for reviewers.

- Create a Strapi project using `cli/create-strapi/bin/index.js`
  - A deviceId will be printed during the creation. Save it somewhere. Depending on the machineID() function, it will be either generated using the strapi.uuid + machineId, or random. In the last case, it will be saved in the package.json. Please verify that. So you can test two things, regular process where the machineID() function doesn't throw (deviceId not random), and where the machineID() function throws and creates a random deviceId that will be saved in package.json
- Then, trigger any event from the telemetry instance (ex: `didChangeInterfaceLanguage`, `didUpdateRolePermissions`, `didInviteUser`)
  - A deviceId will be printed in your terminal. **It should match the one from the CLI**.


maccomaccomaccomacco:
@maximeMD I wonder if the change on the userId on strapi package side is enough.

[On the tracking system side](https://github.com/strapi/tracker/blob/44d351d1b627f1b6b63079f745be448ed3cd4be9/src/services/track-v2.ts#L64C40-L64C46), the userId is already ashed comnining the uuid and the projectId, which should be enough

innerdvations:
instead of `deviceId` and `isDeviceIdUsingProjectId`, would it be possible to leave `deviceId` alone and add a new property like `installId` that always combines the two?

Mcastres:
> @maximeMD I wonder if the change on the userId on strapi package side is enough.
> 
> [On the tracking system side](https://github.com/strapi/tracker/blob/44d351d1b627f1b6b63079f745be448ed3cd4be9/src/services/track-v2.ts#L64C40-L64C46), the userId is already ashed combining the uuid and the projectId, which should be enough

You are right, I forgot that we were doing that already, I'll probably revert my changes then

Mcastres:
> instead of `deviceId` and `isDeviceIdUsingProjectId`, would it be possible to leave `deviceId` alone and add a new property like `installId` that always combines the two? That way you'd still always have an accurate deviceId that worked the same through all versions, while also knowing that the new property was always device+project
> 
> I'm not very familiar with the back-end analytics configuration so maybe that won't work, but if you're already separating with `isDeviceIdUsingProjectId` it seems like that should be possible.

Humm I see what you mean but from my knowledge, I think that we are just using the deviceId for tracking purpose. Since the current value it can have is causing issues later, I'm wondering if it makes sense to keep it and have a new one that will basically replace it 🤔

maccomaccomaccomacco (dismissed):
QAed, looks good to me

innerdvations:
> > instead of `deviceId` and `isDeviceIdUsingProjectId`, would it be possible to leave `deviceId` alone and add a new property like `installId` that always combines the two? That way you'd still always have an accurate deviceId that worked the same through all versions, while also knowing that the new property was always device+project
> > I'm not very familiar with the back-end analytics configuration so maybe that won't work, but if you're already separating with `isDeviceIdUsingProjectId` it seems like that should be possible.
> 
> Humm I see what you mean but from my knowledge, I think that we are just using the deviceId for a tracking purpose. Since the current value it can have is causing issues in our system, I'm wondering if it makes sense to keep it and have a new one that will basically replace it 🤔
> 
> So yeah, having a deviceId which would be an actual deviceId + projectId is no longer a deviceId but it is the one we need. Actually, I would just rename deviceId in the code by installId, and then give `deviceId: installId` to our tracking calls.

Yeah, my main concern was just that having `isDeviceIdUsingProjectId` adds some extra logic we have to be aware of in multiple places, whereas if we just rename it we know exactly what it is and don't have to check which "version" of deviceId it is. Would it be a problem to instead rename it to something like `installId`?

Mcastres:
> Yeah, my main concern was just that having `isDeviceIdUsingProjectId` adds some extra logic we have to be aware of in multiple places, whereas if we just rename it we know exactly what it is and don't have to check which "version" of deviceId it is. Would it be a problem to instead rename it to something like `installId`? Or would that just be shifting the complexity around?

No, I'm up to rename it!

innerdvations (commented):
One last check: it won't be a problem that all `deviceId` will now actually be `installId`? It seems like that will "disconnect" all new records from any previous ones. Would it be better to continue sending `deviceId`but also send `installId` along with it and then update the telemetry server?

I'm not 100% up to speed on what's going on on the back-end so if you're sure on that part, I'll approve it after you remove the console logs!

innerdvations:
By the way, if you want to test it out yourself to be 100% sure the events come through correctly on the telemetry server, I made an experimental: 0.0.0-experimental.eb2d102c03f0984bece7258f6e0e8ab9ee39c82d

innerdvations (approved):
Approving this with the caveat I wrote above about the telemetry data; I'll leave that up to you to confirm. As far as the CMS side, looks good though.

packages/core/core/src/services/metrics/sender.ts
```diff
@@ -70,12 +71,15 @@ export default (strapi: Core.Strapi): Sender => {
   return async (event: string, payload: Payload = {}, opts = {}) => {
     const userId = generateAdminUserHash(strapi);
 
+    // TO REMOVE
+    console.log(`[TELEMETRY INSTANCE] installId: ${installId}, userId: ${userId}`);
+
```
innerdvations, line 76:
```suggestion
```

Mcastres:
> One last check: it won't be a problem that all `deviceId` will now actually be `installId`? It seems like that will "disconnect" all new records from any previous ones. Would it be better to continue sending `deviceId` but also send `installId` along with it and then update the telemetry server?

Could you elaborate more on what you mean? From what I understood, it's not an issue of the deviceId will change for the same user/project with this change. It's not important in terms of data.

Mcastres:
@nathan-pichon pinging you on this since I brought changes to the Cloud CLI part. Just want to confirm it's all right

packages/core/utils/src/install-id.ts
```diff
@@ -0,0 +1,14 @@
+import crypto from 'crypto';
+import { machineIdSync } from 'node-machine-id';
+
+export default (projectId: string, installId: string) => {
+  if (installId) return installId;
```
nathan-pichon, line 5: I don't get why we have the installId in parameters if we return it if it's not a nullable. Shouldn't the function be not invoked if the caller already has the information?

nathan-pichon (commented):
Some comments, overall LGTM.

packages/core/utils/src/install-id.ts
nathan-pichon: Wouldn't it be less obfuscated to do `installId ?? getInstallId(projectId)` where you need the installId? 

Another one: Could you do a named export? using `export default` reduce the discoverability of the code 👍

Mcastres: What do you think @innerdvations?

Mcastres: Yes, I can make a named export!

nathan-pichon (approved):
On my side, lgtm
I'll let you discuss with Ben about the request changes that needs to be changed.

innerdvations:
My changes were just removing the debug logging

Mcastres:
Alright! Should anyone else review this?

Mcastres:
@innerdvations

innerdvations (approved):

Mcastres:
@innerdvations could we merge it?

PR #23121, opened 2025-03-12
fix: preview should not have a background

markkaylor:
### What does it do?

Fixes incorrect boolean has background passed to layout

### Why is it needed?

To not see the background on the preview page

### How to test it?

Go to the preview page you should see this:

Screenshot 2025-03-12 at 10 37 23
Not this:
Screenshot 2025-03-12 at 10 37 40

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
+++ b/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
@@ -224,7 +224,7 @@ const PreviewPage = () => {
                         <FormLayout
                           layout={documentLayoutResponse.edit.layout}
                           document={documentResponse}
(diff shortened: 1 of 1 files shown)


simotae14 (approved):

markkaylor:
I'm a bit confused though now, it looks like we changed the hasBackground prop around at some point. @simotae14 do you know why?

simotae14:
I think we had a problem with naming, which was confusing at the beginning, we had a discussion in the past with @remidej because we needed to introduce the prop for both preview and ROTF, I think we can discuss it together and find a better name

remidej (approved):

PR #23122, opened 2025-03-12
fix: onPreview is not called when updating a relation

markkaylor:
### What does it do?

- Passes onPreview to the DocumentContextProvider instead of the PreviewProvider

Side quests:
- Rename RelationModalWrapper => RelationModalForm
- Update incorrect component string names in the context providers

### Why is it needed?

- So the callback can be called in the RelationModal and on the PreviewPage

### How to test it?

Go to preview, update a relation, you should see the change immediately without having to refresh

https://github.com/user-attachments/assets/2f0029f8-1380-48f0-90be-dbe129ca5417

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
@@ -156,7 +156,7 @@ const UnstablePreviewHeader = () => {
   const schema = usePreviewContext('PreviewHeader', (state) => state.schema);
   const meta = usePreviewContext('PreviewHeader', (state) => state.meta);
   const plugins = useStrapiApp('PreviewHeader', (state) => state.plugins);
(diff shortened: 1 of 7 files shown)


packages/core/content-manager/admin/src/preview/components/PreviewHeader.tsx
```diff
@@ -156,7 +156,7 @@ const UnstablePreviewHeader = () => {
   const schema = usePreviewContext('PreviewHeader', (state) => state.schema);
   const meta = usePreviewContext('PreviewHeader', (state) => state.meta);
   const plugins = useStrapiApp('PreviewHeader', (state) => state.plugins);
-  const iframeRef = usePreviewContext('PreviewHeader', (state) => state.iframeRef);
+  const onPreview = usePreviewContext('PreviewHeader', (state) => state.onPreview);
```
markkaylor, line 159: Actually we could just get this from the DocumentContext instead, it doesn't need to live in both places.

Issue #23070, opened 2025-03-05
Bug with PUT 'Empty .update() call detected'


SontorsM:
### Node Version

20.18.2

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.11.0

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

Added a new record to content-type(added relation field) and updated strapi from 5.10.4 to 5.11.0 (my fault, shouldn't many changes in one step)

Stopped working PUT for old records of content type . 
After creating a new record - works ok. Send PUT and update without problem.  
With old records: 

Error in strapi:
`Empty .update() call detected! Update data does not contain any values to update. This will result in a faulty query.` 

Rolled back strapi version to 5.10.4, saving the changes to content type  and problem was solved. 

### Steps to Reproduce

Cannot reproduce cause work in producution db. 

### Expected Behavior

Should not get 'Empty .update() call detected! Update data does not contain any values to update. This will result in a faulty query.' 
Should update record

### Logs

```shell

```

### Code Snippets

stock. no additions to code. 

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

jorrit:
I guess 6fdadbae0a1232299a094878b5ae4d81575d58af or 4d4e78457ae42752910f16fdfc60632c3499e914 are the culprit. For me it is reported on a 1 to many relationship with the same content type (parent_page <-> child_pages).

cdreier:
same here - happens even with plain document service usage

Nijhazer:
I'm getting this same error when running a v4 -> v5 upgrade, because `.sync()` is invoked on a 1-to-many relationship, looks for `orderColumnName` on `existingRelations.relations`, and finds nothing, because there is a `inverseOrderColumnName` property but there is not a `orderColumnName` property.

iberflow:
Fresh `5.11.0` install with a belongs to many relationship. Downgraded to 5.10.4 as OP has and everything seems to be working.

amjadniazi48:
same here with 5.11.0, I m getting this error in case of updating the collection type having relation

magicfr3d:
Same problem with 5.11.0

kinkinava:
same here

[...]

PR #23123, opened 2025-03-12
fix: fails to update self referencing relations on publish

Marc-Roig:
### What does it do?
When publishing, we were re we were syncing bidirectional relations order
https://github.com/strapi/strapi/pull/22834

This PR introduced a rather manual process in the document service, to handle the inverse order of relations when publishing.

It was achieved by retrieving all relations connected to a soon to be published entry, and overwritting the order of new relations.

This introduced an issue for bidirectional relations targeting itself (same content type). Those do not store any inverse order, and the code didn't account for null order values.

The current PR addresses the issue by NOT syncing any bidirectional relation that targets itself (the same content type)

### How to test it?

- Create a content type (e.g Category) with Draft and Publish enabled
- Create a bidirectional relation (e.g oneToMany) that targets itself  (categories -> category)
- Create a draft entry and connect a category relation
- Publish the entry
- Update the draft and try to publish

Before, this resulted in the error described in  https://github.com/strapi/strapi/issues/23070

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/23070

### DIFF
--- a/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
@@ -73,6 +73,11 @@ const load = async (uid: UID.ContentType, { oldVersions }: LoadContext) => {
           continue;
         }
 
+        // If it's a self referencing relation, there is no need to sync any relation
+        if (model.uid === uid) {
+          continue;
+        }
+
         const joinTable = attribute.joinTable;
         if (!joinTable) {
           continue;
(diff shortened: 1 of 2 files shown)


markkaylor (approved):
As far as I can tell this fixes the issue. I had one case where existing data still threw the error. I wasn't able to reproduce it though, the next time I tried it was fine.

jorrit:
Gracias

PR #23125, opened 2025-03-12
fix: relation label fix overflow behaviour

simotae14:
### What does it do?

Fixes the style of the Modal.Trigger to occupy just one line even with long relation names

### Why is it needed?

To respect the precedent behavior and show relations in the edit view, relation modal and preview page with the correct ui

### How to test it?

Go to the edit view page and open an entry that contains a relation with a long name
it should appear like this
Screenshot 2025-03-13 alle 11 31 51

same in the Preview page of the entry
Screenshot 2025-03-13 alle 11 32 27

### Related issue(s)/PR(s)

CS-1364


remidej:
@simotae14 your screenshot shows the text truncated, don't we want to show the ellipsis? (...)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -1064,7 +1064,7 @@ const ListItem = ({ data, index, style }: ListItemProps) => {
               </IconButton>
             ) : null}
             <Flex width="100%" minWidth={0} justifyContent="space-between">
-              <Box minWidth={0} paddingTop={1} paddingBottom={1} paddingRight={4}>
+              <Box minWidth={0} paddingTop={1} paddingBottom={1} marginRight={4} overflow="hidden">
```
markkaylor, line 1067: Is overflow necessary?  I think paddingRight still gives you the same result and I think it would be better since we are applying padding everywhere. 

The more I look at this though I'm thinking we shouldn't have this Box wrapper here. Instead shouldn't we handle the styling of the trigger (including the ellipsis) inside the RelationModalForm where we have the trigger?

markkaylor (approved):

PR #23128, opened 2025-03-12
chore: drop side editor future flag

remidej:


tests/e2e/tests/content-manager/preview.spec.ts
remidej, line 1: the diff looks wild but i just changed the describeOnCondition condition

markkaylor (approved):
Nice, tested and it looks like its working

PR #23141, opened 2025-03-13
feat(uid): add regex attribute

noreiller:
The UID field can be used to generate a slug but it misses the opportunity to include a slash (`/`).

This PR implements a new "regex" attribute based on the one from the Text field.

### What does it do?

It adds a `regex` attribute on the UID field type.

### Why is it needed?

When using the UID field type to generate a slug, we may need to include a slash (`/`).

### How to test it?

- In the "get started" example, update the homepage content type to set the regex attribute of the `slug` property with the following pattern `^[A-Za-z0-9-_.~/]*$`. It's the same than the UID_REGEX used everywhere in the codebase, except it allows the `/` character.
- Edit the `slug` property in homepage content to replace the value with `home/page`.

### Screenshots

BEFORE

Capture d’écran du 2025-03-13 09-15-32

AFTER

Capture d’écran du 2025-03-13 11-31-42
Capture d’écran du 2025-03-13 11-30-55
Capture d’écran du 2025-03-13 11-31-26

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/UID.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/UID.tsx
@@ -42,11 +42,12 @@ import type { Schema } from '@strapi/types';
 const UID_REGEX = /^[A-Za-z0-9-_.~]*$/;
 
 interface UIDInputProps extends Omit<InputProps, 'type'> {
+  attribute?: Pick<Schema.Attribute.UIDProperties, 'regex'>;
   type: Schema.Attribute.TypeOf<Schema.Attribute.UID>;
 }
 
 const UIDInput = React.forwardRef<any, UIDInputProps>(
-  ({ hint, label, labelAction, name, required, ...props }, ref) => {
+  ({ hint, label, labelAction, name, required, attribute = {}, ...props }, ref) => {
     const { currentDocumentMeta } = useDocumentContext('UIDInput');
     const allFormValues = useForm('InputUID', (form) => form.values);
     const [availability, setAvailability] = React.useState<CheckUIDAvailability.Response>();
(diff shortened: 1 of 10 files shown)


noreiller:
Hello @oiorain @hanpaine , do you know if this PR could be accepted/reviewed? This is a critical need for our migration to Strapi. Any investigation on another solution would be more complex.

innerdvations:
Two things:

- Could you please update the branch to get the latest CTB changes? I don't think there will be any issue but considering the major refactoring we recently did I want to be sure
- I get a generic server error when attempting to save `^[A-Za-z0-9-_.~/]*$` with an empty default value, which appears to be two issues. When I use the same regex with text attributes it works, but that seems to be because we're not even validating that :( and I'm glad you've done so here so if we can keep it that would be great.
  1. `""` should pass that regex. My first guess (didn't verify) is that it's a case like `default` is being sent as null or undefined rather than `""` so the check fails.
  2. no error message is making it to the user, so (especially in the context of the new CTB) if it causes the save to fail it will be almost impossible to figure out why. Not sure it will actually be an issue at all after updating the branch -- maybe it will automatically start working then :)

And thank you very much for this contribution, it's a nice addition!

noreiller:
> Two things:
> 
>     * Could you please update the branch to get the latest CTB changes? I don't think there will be any issue but considering the major refactoring we recently did I want to be sure
> 
>     * I get a generic server error when attempting to save `^[A-Za-z0-9-_.~/]*$` with an empty default value, which appears to be two issues. When I use the same regex with text attributes it works, but that seems to be because we're not even validating that :( and I'm glad you've done so here so if we can keep it that would be great.
>       
>       1. `""` should pass that regex. My first guess (didn't verify) is that it's a case like `default` is being sent as null or undefined rather than `""` so the check fails.
>       2. no error message is making it to the user, so (especially in the context of the new CTB) if it causes the save to fail it will be almost impossible to figure out why. Not sure it will actually be an issue at all after updating the branch -- maybe it will automatically start working then :)
> 
> 
> And thank you very much for this contribution, it's a nice addition!

Hi, thanks for checking the changes. I'll do that asap. :slightly_smiling_face:

noreiller:
@innerdvations You spot an issue when the regexp is updated through the content type builder, I missed to add the new property in the zod schema. It's now fixed and working.

innerdvations (approved):
LGTM, tested it and it works as expected now. @Bassel17 @remidej Could one of you give the front-end code a quick final review in case I missed something?

PR #23145, opened 2025-03-13
chore: refactor relation input modal

remidej:
Refactoring of the relations on the fly solution. The main goal is to simplify the logic to make it less error-prone, and to remove the need for the DocumentContextProvider for ease of use.

The key to make it work was to update our useContext hook to not throw when it's called outside the context (with a specific param). This is what allowed me to get rid of the DocumentProvider. I'm keeping a useDocumentContext hook though, which tries to read from the relation modal if available, and falls back to URL-based state. It also allowed me to get the onPreview handler from the preview context directly.

I also removed the component unit test. It relied on many context providers and API queries, so it required lots of mocking and boilerplate, making it hard to manage and hard to trust. This is the type of logic that is much better covered in e2e tests in my opinion. Instead I wrote unit tests for the reducer managing the modal state.


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -57,91 +59,193 @@ const CustomModalContent = styled(Modal.Content)`
...
-  const addDocumentToHistory = (document: DocumentMeta) =>
-    setDocumentHistory((prev) => [...prev, document]);
+  // Return children directly if the context is already in the tree
+  try {
+    useRelationModal('RelationModalWrapper', () => null);
+    // If we reach that point without an error, the context is already in the tree
+    return memoizedChildren;
+  } catch (e) {
+    // Silently catch the error. We will wrap children in the context provider below
+  }
```
remidej, line 203: This isn't quite working yet. I'm getting an endless rerender issue that crashes my browser

remidej: I had another idea on my run earlier. If this current approach turns out to be tricky, we could also decide to just recursively wrap the context inside itself anyway. I think it should work if we always forward the context value from the parent

```js
const context = useRelationModal(state => state);

return <RelationModalProvider {...state}>{children}</RelationModalProvider>
```

useContext just matches the closest parent, it doesn't complain about being nested inside itself, so I believe eventually it should resolve to the same parent context so everything is still managed in the same reducer

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -57,91 +61,292 @@ const CustomModalContent = styled(Modal.Content)`
...
-    (state) => state.rootDocumentMeta
-  );
-  const currentDocumentMeta = useDocumentContext('RelationModalForm', (state) => state.meta);
-  const changeDocument = useDocumentContext('RelationModalForm', (state) => state.changeDocument);
-  const documentHistory = useDocumentContext('RelationModalForm', (state) => state.documentHistory);
-  const setDocumentHistory = useDocumentContext(
-    'RelationModalForm',
-    (state) => state.setDocumentHistory
-  );
+const FormWatcher = () => {
```
remidej, line 172: Need to check if I can move this inside another existing component that has access to the form context. Because I don't love having a component that doesn't return any UI

simotae14 (approved):
well done @remidej  and 🙏🏽 thanks for your refactoring

PR #23150, opened 2025-03-14
fix: add z index property to the menu content in blocks and markdown

simotae14:
### What does it do?

Fix the visibility problem we had in preview with the blocks and markdowns when you reduce the width of the viewport and show the three dots menu

### Why is it needed?

Currently the menu is not shown

### How to test it?

Open an entry with a markdown field or a block field and open the preview page of it, then reduce the width of the page and click on the three dot menu in the header of the block field and markdown field and check it is working

### Related issue(s)/PR(s)

CS-1361


markkaylor (approved):
Nice catch

PR #23152, opened 2025-03-14
fix: add the correct locale and document Id when we reorder the relations

simotae14:
### What does it do?

Pass the correct data on reordering

### Why is it needed?

Because otherwise we have this issue in the Relations on the fly modal

https://github.com/user-attachments/assets/757e85f5-9caa-4dad-a602-5cedcf129ff4

### How to test it?

adding a new relation, reorder, click to open the modal

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
@@ -802,9 +802,7 @@ const RelationsList = ({
             ...{
               apiData: {
                 id: relation.id,
-                documentId: relation.documentId
-                  ? relation.documentId
-                  : relation.apiData?.documentId || '',
+                documentId: relation.documentId ?? relation.apiData?.documentId ?? '',
                 locale: relation.locale || relation.apiData?.locale,
                 position,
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -802,8 +802,10 @@ const RelationsList = ({
             ...{
               apiData: {
                 id: relation.id,
-                documentId: relation.documentId,
-                locale: relation.locale,
+                documentId: relation.documentId
+                  ? relation.documentId
+                  : relation.apiData?.documentId || '',
```
markkaylor, line 807: It's a bummer because the TypeScript isn't correct. I kind of prefer this syntax for this if it's all the same to you. 

```suggestion
                documentId: relation.documentId ?? relation.apiData?.documentId ?? ''
```

markkaylor (approved):
Nice it's working. I did notice you can't reorder when creating an entry. If that is the same on develop then we don't need to fix it here in my opinion. Here's an example creating an entry and reordering


https://github.com/user-attachments/assets/eedee986-b18f-4fc0-aea1-09ea72429683

Issue #22512, opened 2024-12-26
publishedAt is updated when an already published entry is updated


Aylay:
### Required System information

- Node.js version: 20
- Yarn version: v1.22.19
- Strapi version: 5.6.0
- Database: sqlite
- Operating system: MacOs Sequoia 15.2 - darwin-x64
- Is your project Javascript or Typescript: Typescript

### Describe the bug

I updated an entry from a collection and the publishedAt is updated same as updatedAt.
So publishedAt and updatedAt are the same.

### Expected behavior

publishedAt don't have to change
only updatedAt

### Screenshots
<img width="508" alt="Capture d’écran 2024-12-26 à 15 35 04" src="https://github.com/user-attachments/assets/eda8ab7d-0e31-456f-96da-3dcb1d024b40" />

derrickmehaffy:
@innerdvations / @Marc-Roig can I get your input on this if it's intended or not?

AdrianoCahete:
> can I get your input on this if it's intended or not?

Even if is intended, isn't conceptually wrong?

derrickmehaffy:
> > can I get your input on this if it's intended or not?
> 
> Even if is intended, isn't conceptually wrong?

Not entirely because you could create a new draft and publish that draft 2 months later and it would be considered a "published date".

jackbuehner:
> > > can I get your input on this if it's intended or not?
> > 
> > 
> > Even if is intended, isn't conceptually wrong?
> 
> Not entirely because you could create a new draft and publish that draft 2 months later and it would be considered a "published date".

Back in v4, if we edited in published version, `publishedAt` would remain the same, but `updatedAt` would change. In the case where an entry is already published, it would be nice to be able to keep `publishedAt` as the original publish date.

derrickmehaffy:
It's a bit different now since we have draft & published v2 so you always have two different "entities" one draft and one published.

Aylay:
Ok, I understand but how can I get the `publishedAt` as I compare with the `updatedAt` ?

derrickmehaffy:
From what I can see in the code we expressly forbid allowing users to change it manually via:

https://github.com/strapi/strapi/blob/5d0cd5a1a7d80ae3da90f2331d25161b28f06910/packages/core/core/src/services/document-service/draft-and-publish.ts#L54-L63

I've asked our engineers in our internal Slack what is intended.

cc @innerdvations / @maccomaccomaccomacco / @Marc-Roig

[...]

PR #23160, opened 2025-03-15
 Added firstPublishedAt field which sets the datetime when a content is first published. (#22512)

ankit7201:
### What does it do?

Added firstPublishedAt field which sets the datetime when a content is first published.

### Why is it needed?

See: https://github.com/strapi/strapi/issues/22512

### How to test it?

Create a content-type and query it using the API. You should see the firstPublishedAt field. 
Update the content and publishedAt field would change but firstPublishedAt field will not. 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/22512


alexandrebodin:
Hi @ankit7201  in order to accept this contribute that looks great we will need the follow things:

- Making it opt-in (you can look at the `populateCreatorField` option for example to avoid making it a Breaking change
- Add unit & api integration tests.

ankit7201:
Hi @alexandrebodin 

Thank you for the review. 

Had a doubt regarding the first point that you mentioned and I was hoping to get some clarity on it. 

When you mentioned to make this opt-in, did you mean its visibility will be controlled by a flag in schema (similar to populateCreatorField) 

Meaning, it needs to be "private" unless it is enabled in the schema?

alexandrebodin:
> Hi @alexandrebodin
> 
> Thank you for the review.
> 
> Had a doubt regarding the first point that you mentioned and I was hoping to get some clarity on it.
> 
> When you mentioned to make this opt-in, did you mean its visibility will be controlled by a flag in schema (similar to populateCreatorField)
> 
> Meaning, it needs to be "private" unless it is enabled in the schema?

We need to opt-in adding the field completely not just making it visible. It shouldn't be there by default or it would be a breaking change. It's about wrapping the code where you add the attribute with a condition checking the user has opted in adding it.

ankit7201:
> We need to opt-in adding the field completely not just making it visible. It shouldn't be there by default or it would be a breaking change. It's about wrapping the code where you add the attribute with a condition checking the user has opted in adding it.

Got it. 
But, wouldn't this cause problem if a user publishes a content and then enables this field, since in that case, the field would be created but its value won't be the first time this was published or is that the expected behaviour?

alexandrebodin:
> > We need to opt-in adding the field completely not just making it visible. It shouldn't be there by default or it would be a breaking change. It's about wrapping the code where you add the attribute with a condition checking the user has opted in adding it.
> 
> @alexandrebodin Got it. But, wouldn't this cause problem if a user publishes a content and then enables this field, since in that case, the field would be created but its value won't be the first time this was published or is that the expected behaviour?
> 
> What I mean is, since we are not recording the first published at datetime if it is not enabled by default, and some content is published and then we are enabling it, we don't have any value to populate this field.
> 
> Or, is the expectation that if this is enabled after the content is published, a user would have to manually update this field to add a value to it.

Good questions

-> For new content-types created we would generate them with the option enabled so new users would have the field enabled.
-> For existing projects that want this field they would need to add the option. we can create a enable/disable migration. like we have for enabling/disabling Draft & Publish that will set the value when enabling it. you can look at packages/core/core/src/providers/registries.ts and the ../migrations import in there

ankit7201:
>  that will set the value when enabling it
@alexandrebodin What will we set the value to when it is enabled for existing project? 

One idea would be to set it same as publishedAt date but that would be inaccurate since a content might be published several times and thereby changing the published date and time when it was first published?

alexandrebodin:
> > that will set the value when enabling it
> 
> @alexandrebodin What will we set the value to when it is enabled for existing project?
> 
> One idea would be to set it same as publishedAt date but that would be inaccurate since a content might be published several times and thereby changing the published date and time when it was first published?

Yes it's the best we can do for existing data anyway

ankit7201:
@alexandrebodin 
And for new apps, since firstPublishedAt date would be true by default, do we need to update the UI as well (similar to draftAndPublish)? 

Screenshot 2025-03-19 at 5 22 05 PM

alexandrebodin:
> @alexandrebodin And for new apps, since firstPublishedAt date would be true by default, do we need to update the UI as well (similar to draftAndPublish)?
> 
> Screenshot 2025-03-19 at 5 22 05 PM

I think we should just make it an option in the json as the goal isn't to easily switch back and forth it's just a way to introduce it in a backward compat way :)

ankit7201:
@alexandrebodin 

Got it. Thank you for all the help. 

Here is the final action plan: 
1. Add code to include firstPublishedAtField as part of the content-type schema json with default value as true anytime a new content type is created
2. Add code to update the schema attribute to include firstPublishedAt field if it is enabled in the schema json (enabled by default)
3. For existing apps, add DB migrations using publishedAt date as the value (similar to draftAndPublish) 

For point 1, if possible, can you point me to some code sample where some schema options are being set when a new content type is created?

alexandrebodin:
> @alexandrebodin
> 
> Got it. Thank you for all the help.
> 
> Here is the final action plan:
> 
> 1. Add code to include firstPublishedAtField as part of the content-type schema json with default value as true anytime a new content type is created
> 2. Add code to update the schema attribute to include firstPublishedAt field if it is enabled in the schema json (enabled by default)
> 3. For existing apps, add DB migrations using publishedAt date as the value (similar to draftAndPublish)
> 
> For point 1, if possible, can you point me to some code sample where some schema options are being set when a new content type is created?

If you look in packages/core/content-type-builder/server/src/services/schema-builder/content-type-builder.ts you will see where we set the options that will be saved to file :)

ankit7201:
@alexandrebodin 
Thanks. 

Also, just realized one more thing while working on this. 

Wouldn't it be better if we create the firstPublishedAt field by default and keep it private/not private since by default we are enabling it anyway. 

This way, if someone switches off the feature and later switches it on again, it wouldn't cause problem with the firstPublishedAt value disappearing since while turning it off would not include it in the schema.

alexandrebodin:
> @alexandrebodin Thanks.
> 
> Also, just realized one more thing while working on this.
> 
> Wouldn't it be better if we create the firstPublishedAt field by default and keep it private/not private since by default we are enabling it anyway.
> 
> This way, if someone switches off the feature and later switches it on again, it wouldn't cause problem with the firstPublishedAt value disappearing since while turning it off would not include it in the schema.

This would be a breaking change. we need to make sure someone upgrading to the latest Strapi version doesn't get this field added at all. it is not in the reserved names so someone could already have a field call that way

ankit7201:
> > @alexandrebodin Thanks.
> > Also, just realized one more thing while working on this.
> > Wouldn't it be better if we create the firstPublishedAt field by default and keep it private/not private since by default we are enabling it anyway.
> > This way, if someone switches off the feature and later switches it on again, it wouldn't cause problem with the firstPublishedAt value disappearing since while turning it off would not include it in the schema.
> 
> This would be a breaking change. we need to make sure someone upgrading to the latest Strapi version doesn't get this field added at all. it is not in the reserved names so someone could already have a field call that way

Makes sense. 
So if someone disables it and reenabled it, is it acceptable that the value is gone and it needs to be updated by the user after enabling it again?

ankit7201:
Hey @alexandrebodin 

Just wanted to check back for the above query in case you missed my previous comment :-)

Here is the scenario that I would need some more information on: 
1. User creates a new content type
2. This would create that content type schema with firstPublishedAt as true 
3. If firstPublishedAt is true, during strapi start, firstPublishedAt key would be added to the DB schema 
4. User fetches the data using the API and gets this firstPublishedAt in the response
5. User makes firstPublishedAt as false in the content type schema and restarts strapi 
6. Since, firstPublishedAt is now false, this would remove this from the DB (Correct me if I am wrong here but in my limited knowledge, this is how it should work)
7. User adds some new entries. 
8. User enables firstPublishedAt in the content type schema
9. Because it is now enabled and when strapi restarts, the migration would pick it up and populate firstPublishedAt key using publishedAt key (this is required since in step 6, firstPublishedAt was deleted when it was made false in the schema) 
10. User sees this new firstPublishedAt value 

Is this scenario expected?

alexandrebodin:
> Hey @alexandrebodin
> 
> Just wanted to check back for the above query in case you missed my previous comment :-)
> 
> Here is the scenario that I would need some more information on:
> 
> 1. User creates a new content type
> 2. This would create that content type schema with firstPublishedAt as true
> 3. If firstPublishedAt is true, during strapi start, firstPublishedAt key would be added to the DB schema
> 4. User fetches the data using the API and gets this firstPublishedAt in the response
> 5. User makes firstPublishedAt as false in the content type schema and restarts strapi
> 6. Since, firstPublishedAt is now false, this would remove this from the DB (Correct me if I am wrong here but in my limited knowledge, this is how it should work)
> 7. User adds some new entries.
> 8. User enables firstPublishedAt in the content type schema
> 9. Because it is now enabled and when strapi restarts, the migration would pick it up and populate firstPublishedAt key using publishedAt key (this is required since in step 6, firstPublishedAt was deleted when it was made false in the schema)
> 10. User sees this new firstPublishedAt value
> 
> Is this scenario expected?

Yes this looks good. Objectively someone wouldn't be playing around with the option in production with actual usefull data. once you decide to enable it (for old projects) or disable it for new projects it most likely will stay that way. 

One last thing I think you need to check with this though is that the condition to add it should be that `firstPublishedAt` & `draftAndPublish` are both true. We can't add it for type that disable D&P :) 

In terms of the logic to set the value the default() { new Date() } won't be sufficient, it would add the date on the draft data. you will need more logic to be added in the document service.

ankit7201:
Hi @alexandrebodin 

I have updated the code wherever applicable. I am yet to add the test cases but just wanted to get a quick review to make sure if this is what is expected.  Would you mind taking a look at the changes? :)

Changes: 
- Added firstPublishedAt: true for content-type schema builder
-  If firstPublishedAtField is enabled and draftAndPublish are enabled, then added the firstPublishedAt attribute to the schema. 
- While publishing new content, added firstPublishedAt but only if it is being added for the first time. (Checked this by checking if there is any old published entry present or not)
- While finally updating firstPublishedAt, also checked if first published at is present as data when request to update this field is coming from the api. This is to make sure that someone can update first published at only when sending as part of the update api otherwise first published at should either be the old one or if not present, it should be same as publishedAt field 
- Also, if someone sends a request to update firstPublishedAt, it should not be part of draft but only for the published entry so removed firstPublishedAt when draft is being updated. 
- Added migration which checks if the firstPublishedAtField has been enabled, copy the value from publishedAt and copy it to firstPublishedAt

Please note: 
1. All existing tests are passing except a couple which needs the snapshots to be updated. 
2. Will be working on adding the test cases next. Just to make sure if the code changes are in accordance to what is expected :)

Thank you.

alexandrebodin:
Hey @ankit7201 Our team managing this scope will take over the review. Something I have been thinking about that could make more sense is set this as an option in the config folder so it's just a flag to trigger the new fields if D&P is enabled and it will apply everywhere. I'm a bit concerned about have an option per content type :) 

The team will be coming back to you soon with some product options :)

alexandrebodin:
Hi @ankit7201 the team is on a few days of workshop & conference, they will be back at it next week 🙏

simotae14:
Hello, @ankit7201 I can't see in your PR unit & api integration tests as requested in another comment.

simotae14:
I tried your solution with a brand new collection and it seems working, `firstPublishedAt` and `publishedAt` have different values as requested
Screenshot 2025-04-08 alle 17 21 31
and in the case of old collections instead the `firstPublishedAt` is not shown

packages/core/core/src/services/document-service/first-published-at.ts
```diff
@@ -0,0 +1,47 @@
+import { Modules } from '@strapi/types';
+import { contentTypes } from '@strapi/utils';
+import { Model } from '@strapi/utils/dist/types';
```
simotae14, line 3: I don't think it is the right way of importing stuff

ankit7201:
> Hello, @ankit7201 I can't see in your PR unit & api integration tests as requested in another comment.

Hi @simotae14 
Thanks for the review. 

I haven't added tests yet because I wanted to make sure my changes are what is expected. 
@alexandrebodin mentioned that you guys are thinking of going in a different direction for this. 

If the current changes are fine, I can start working on the tests. If any change is needed, please let me know so I can add those changes and then write the test once the changes are according to the expectation.

simotae14:
Hi @ankit7201
Thanks again for your contribution, I'm sure many community members will be grateful.
As I said I tested your solution and it works, I let other people in the team with more knowledge on that part give you a proper review
And regarding the tests I think you can try to add some of them in the meantime.
Thanks again

oiorain:
Hello @ankit7201,

Thanks again for your contribution! we really appreciate your work.
I finally got a chance to talk to the team about this. Indeed, as Alex pointed out, we'd like this to be a global setting in config files. Would you be able to make the change, please? 

Thanks a lot!

ankit7201:
> Hello @ankit7201,
> 
> Thanks again for your contribution! we really appreciate your work. I finally got a chance to talk to the team about this. Indeed, as Alex pointed out, we'd like this to be a global setting in config files. Would you be able to make the change, please?
> 
> Thanks a lot!

Hi @oiorain 

Sure!

Had a couple of questions if you can please answer them: 

1. By global config, do we mean the configuration under strapi class? (I see something called "loadConfigurtion" function. Is that what we are talking about?)

2. How exactly will this configuration work? Can it be controlled by end user or will it always be enabled? 

3. What happens to old content types for which there is no first published at field? Once a user upgrades to newer version, do I need to add a migration that adds this field in the database for all the old entries? 

4. If we decide to disable this in the future by either disabling it ourselves or through end user (assuming end user has control over this: see point 2), do we delete all the firstPublishedAt entries from the database for all content types?

ankit7201:
Hi @oiorain 

Just wanted to check back regarding the above query :)

Just wanted to understand how to proceed with this feature as global config.

oiorain:
Hi @ankit7201 

Sorry to keep you waiting on this. it's been a busy week :) 

1. By global config, do we mean the configuration under strapi class? (I see something called "loadConfigurtion" function. Is that what we are talking about?)
Yes, this is what i mean. but let me confirm with the devs on the specific files if you need more info. 

3. How exactly will this configuration work? Can it be controlled by end user or will it always be enabled?
It will be set to `false` by default and the user can turn it on if they want. 

4. What happens to old content types for which there is no first published at field? Once a user upgrades to newer version, do I need to add a migration that adds this field in the database for all the old entries?
Yes, i think there would be a need to migrate the rest of the content so it's everywhere since we're enabling globally.

5. If we decide to disable this in the future by either disabling it ourselves or through end user (assuming end user has control over this: see point 2), do we delete all the firstPublishedAt entries from the database for all content types?
I think we can lease it in DB if it's disabled so we keep the content. 

6. Is the original logic of first published at should only be enabled when draft and publish is enabled, still true? If yes, how would that work? Since, draft and publish will be for each content type but first published at will be a global setting.
> Is the original logic of first published at should only be enabled when draft and publish is enabled, still true?
yes.

> If yes, how would that work? Since, draft and publish will be for each content type but first published at will be a global setting.
It will be a global setting so that each database that has draft and publish turned on will have this field. 
Which means it needs to be be created in DB also when draft and publish is enabled in the admin. if d&p is disabled, it can be left in DB so there is no data loss I think.

Thanks again for your work!

ankit7201:
Hi @oiorain 
Thank you for the clarification. 

Can you please elaborate a little more on the last point. I am not sure if I understood it correctly. 

When you say: 
>It will be a global setting so that each database that has draft and publish turned on will have this field.
Which means it needs to be be created in DB also when draft and publish is enabled in the admin. if d&p is disabled, it can be left in DB so there is no data loss I think.

1. Does this mean that when someone enables D&P for a content type, we need to create firstPublishedAt field (but not serve it in the API response because by default first published at will be false)?

2. Or, we only create first published at field in the database only when both D&P and first published at is enabled?

If it is the 1st one, then during migration, we would be looking at d&p for creation of first published at field but serving it in the api response only when first published at field is true in the global config?

Also, if it is the first one, as far as I know, d&p is enabled by default which means everytime someone creates a new content type we would need to create first published at field in the database (unless d & p is specifically disabled)?

Please let me know if I have misunderstood something here. 

Thanks.

ankit7201:
> Hi @ankit7201
> 
> I don't see a problem sharing firstPublishedAt in the API if it is false or not set yet.
> 
> I think it would make sense to create the field in DB at creation of the Content Type if D&P is selected then or when it is enabled later on. So we don't create the field in the case of a creation of content type without d&P enabled.
> 
> Does that answer your question?

Hey @oiorain 

Yes. This answers my question but raises another one :)

When you say: 
>I don't see a problem sharing firstPublishedAt in the API if it is false or not set yet.

Do you mean we would be sending firstPublishedAt even if it is false but d&p is enabled? 
If yes, then what would be the point of the firstPublishedAt flag? Wouldn't we be looking at the d&p always to check if firstPulishedAt needs to be served or not? 

-----
I have another doubt regarding one of the previous conversation we had. 
You mentioned that we would be using [this](https://github.com/strapi/strapi/blob/7c7ba817d00aeac2e9f750e6693dc8e51a8a2659/packages/core/types/src/core/config/admin.ts#L38) file to make the changes and it would be documented [here](https://docs.strapi.io/cms/configurations/server)
But, the file you mentioned is admin and the url points to the server config. 
I see there is server config present as well. Do we add this firstPublishedAt config in the server config or admin config? 
If it is admin config, it would still be documented in server config?

oiorain:
> Do you mean we would be sending firstPublishedAt even if it is false but d&p is enabled?

yes.

> If yes, then what would be the point of the firstPublishedAt flag? Wouldn't we be looking at the d&p always to check if firstPulishedAt needs to be served or not? 

Isnt the point of this new datapoint to carry the datetime of the first time an entry is published? If it hasnt been published yet, then there is no date in the variable, right? 
Yes, i believe we would check if draft and publish is enabled to server this variable or not.


> Do we add this firstPublishedAt config in the server config or admin config?

It should be in admin config.

ankit7201:
>Isnt the point of this new datapoint to carry the datetime of the first time an entry is published? If it hasnt been published yet, then there is no date in the variable, right?
Yes, i believe we would check if draft and publish is enabled to server this variable or not.

Sorry. I did not get this point. 

Here is what I understand: 
There would be a flag called "firstPublishedAtField" whose value would be either true or false. (We would store this flag in admin config) 

If it is true and d&p is enabled, we would be storing firstPublishedAt field in the DB tables for that content type. 
If is is false (irrespective of d&p enabled or not) we would not be storing firstPublishedAt field in the DB tables for any content type. 

The flag "firstPublishedAtField" is deciding here if we need to store firstPublishedAt value in the DB tables or not. 

If d&p is deciding if we need to store "firstPublishedAt" field in the DB table and it needs to be served if d&p is enabled or not, what would be the point of "firstPublishedAtField" flag. 

Since a new table is created for each content type, we would be creating this new column based on the "firstPublishedAtField" in the admin config. 

Sorry for the confusion. Please let me know if I have misunderstood something here.

oiorain:
Hi @ankit7201 

Sorry to keep you waiting on this. We've been very busy with the Strapiconf and it got out of my head.

Am I understanding right that this is what we are doing:

- add an admin config option to enable the feature (important to avoid a breaking change)
- when that config and d&p are both enabled
   - add firstPublishedAt to every content type
   - have a migration so the first time it’s enabled, we set the value for each draft to the current publishedAt time (if it exists) for their locale published version
   - (didn’t see discussed in the PR) if it finds a field with that name that already exists, we let it be used/converted if it was formerly a different data type
- when that config value is disabled
   - we should probably keep the db column, right? I know we have an array somewhere in the db schema sync (?) to prevent it from deleting known fields, like keeping review workflows even if the feature is briefly disabled
- (only saw discussed briefly) the first time a draft is published, the publish time should be saved on that locale’s draft so that the value gets maintained and copied through each time it’s published
- firstPublishedAt should be available in the content API

Does this clarify the situation? 
I am looking in @innerdvations to help you on this.

ankit7201:
> Hi @ankit7201
> 
> Sorry to keep you waiting on this. We've been very busy with the Strapiconf and it got out of my head.
> 
> Am I understanding right that this is what we are doing:
> 
> * add an admin config option to enable the feature (important to avoid a breaking change)
> * when that config and d&p are both enabled
>   
>   * add firstPublishedAt to every content type
>   * have a migration so the first time it’s enabled, we set the value for each draft to the current publishedAt time (if it exists) for their locale published version
>   * (didn’t see discussed in the PR) if it finds a field with that name that already exists, we let it be used/converted if it was formerly a different data type (db sync handles this automatically, this is just a note so we're aware)
> * when that config value is disabled
>   
>   * we should probably keep the db column, right? I know we have an array somewhere in the db schema sync (?) to prevent it from deleting known fields, like keeping review workflows even if the feature is briefly disabled
> * (only saw discussed briefly) the first time a draft is published, the publish time should be saved on that locale’s draft so that the value gets maintained and copied through each time it’s published
> * firstPublishedAt should be available in the content API
> 
> Does this clarify the situation? I am looking in @innerdvations to help you on this.

Hi @oiorain and @innerdvations 

This clarified a  lot of things. Thank you. 

A couple of follow ups: 
> have a migration so the first time it’s enabled, we set the value for each draft to the current publishedAt time (if it exists) for their locale published version

Wouldn't it make sense to add firstPublishedAt only for published entry because someone can create an entry and not publish for sometime. The firstPublishedAt needs to be populated when the entry was first published, right? 
So, the value stored in the draft won't be useful I think. 

> (didn’t see discussed in the PR) if it finds a field with that name that already exists, we let it be used/converted if it was formerly a different data type (db sync handles this automatically, this is just a note so we're aware)

A little confused in this. If the field already exists and it is of different data type (for example a boolean), what do we do here? We change it to date and update the value in it or we leave it be meaning for that entry we can never have firstPublishedAt since someone as already used that name? 
If we overwrite the field, wouldn't it cause data loss for the user? 

> (only saw discussed briefly) the first time a draft is published, the publish time should be saved on that locale’s draft so that the value gets maintained and copied through each time it’s published

As mentioned in point 1, wouldn't it make sense to only have firstPublishedAt value for published entry. For subsequent publishing, I can just check if the previous published entry has firstPublishedAt field set or not. If it is, I can just use that.

innerdvations:
> Wouldn't it make sense to add firstPublishedAt only for published entry because someone can create an entry and not publish for sometime. The firstPublishedAt needs to be populated when the entry was first published, right?

> As mentioned in point 1, wouldn't it make sense to only have firstPublishedAt value for published entry. For subsequent publishing, I can just check if the previous published entry has firstPublishedAt field set or not. If it is, I can just use that.

This part is still debateable, I'm not 100% sure and am open to feedback on it. I'll write my opinion and reasoning here.

For the migration, we should look for published documents and set the time on
- the published document itself
- also on the draft that it was spawned from (that is, the draft version where locale and documentId both match that published version)

When the first publish happens, we should first set the value on the draft and then it will automatically be in the published version as well.

That is because when we publish, what happens in the background is that we delete the current published version and copy the draft to be a new published version. Published versions are ephemeral, and the draft is actually the more permanent version. So to me it feels like the draft should be the source of truth for where firstPublishedAt is stored. It will automatically be included whenever a new published version is made, without needing any extra code to keep it, and it's harder to lose the data from a bug or improperly deleting a published document.

The remaining question, though, is that if that draft was published but then becomes unpublished, do we keep the firstPublishedAt value? Or do we clear it? Because technically it was published before.

ankit7201:
@innerdvations 

> When the first publish happens, we should first set the value on the draft and then it will automatically be in the published version as well.

>That is because when we publish, what happens in the background is that we delete the current published version and copy the draft to be a new published version. Published versions are ephemeral, and the draft is actually the more permanent version. So to me it feels like the draft should be the source of truth for where firstPublishedAt is stored. It will automatically be included whenever a new published version is made, without needing any extra code to keep it, and it's harder to lose the data from a bug or improperly deleting a published document.

Makes sense. 
Doubt: What happens when someone updates the data using the strapi API? Does it affect draft as well? When someone updates firstPublishedAt field using the API, we would be updating both draft and published version? 

> The remaining question, though, is that if that draft was published but then becomes unpublished, do we keep the firstPublishedAt value? Or do we clear it? Because technically it was published before. So my preference is to keep it (which also reinforces the idea we have to keep it on draft, because otherwise there's nowhere to put it after deleting the published doc)

I guess the answer would depend on how strapi interprets "unpublishing"? When someone unpublishes data, is it as if there is no record of that data being published ever and once unpublished means back to the complete original state where only draft entry exist or is there some record somewhere that a publish happened but now it is unpublished? 

If there is no record anywhere, I think then it would make sense to remove firstPublishedAt as well (why treat firstPublishedAt as a special data?), keep it otherwise.

oiorain:
I'll pitch in here: I think it should be kept personally.

yanniskadiri:
Thanks for asking @ankit7201, I agree that `firstPublishedAt` should be a read-only field in the Content API

I believe we should keep the `firstPublishedAt` value as a record of the first time the content was published, even after it's unpublished - if someone wants to track the current published state, that should be handled via separately via `publishedAt`

ankit7201:
> Thanks for asking @ankit7201, I agree that `firstPublishedAt` should be a read-only field in the Content API
> 
> I believe we should keep the `firstPublishedAt` value as a record of the first time the content was published, even after it's unpublished - if someone wants to track the current published state, that should be handled via separately via `publishedAt`

Thanks @yanniskadiri  for the response.

So, firstPublishedAt field would be present even if someone unpublishes or switches off this feature from admin config, right?
Is there any reference I can see in the code where this is happening for an existing field? Because as I understand currently, by default the field is deleted automatically in the DB if it is not part of the config

Also, just realized something regarding firstPublishedAt field not being updatable. Wouldn't that cause a problem in cases where someone enables it for existing content type and firstPublishedAt field's value would be set as current publishedAt field (due to migration) but what if someone wants to change that value to when it was originally published and not current publishedAt field.

yanniskadiri:
Hey @ankit7201 let's keep things simple for now: firstPublishedAt value remains the same, ie the date when a user first published the document

If there is feedback asking us to update the value, then we'll consider it

ankit7201:
Hi @oiorain @innerdvations @yanniskadiri 

I have made the changes according to the discussion. 

I will be working on the test cases next (A little busy nowdays, will pick them up as soon as I get some free time)
Please note: There is some minor cleanup required like removing a couple of console.log statements and removing some commented out code. I will get to it ASAP. 

Let me know if this is what is expected in case you guys get a chance to review. Thanks. 

Following changes done: 

- Adding firstPublishedAt field to the DB irrespective of it is enabled or not but keeping it private in case it is disabled and making it non private if enabled. (This is to ensure that if someone disables the field, firstPublished at data remains intact) 
- During the update, filtering out the firstPublishedAt data (if it is present) since this field should not be updated. 
- Adding firstPublishedAt data to the draft data for the first time when entry is getting published. 
- During migration, I am picking content types in pairs (one for draft and one for published) for each locale.  If firstPublishedAt is already present, migration does nothing for those content types. If there is only one present (instead of a pair), that means this is not published yet so we can ignore it. Otherwise, pick the value from publishedAt field and populate firstPublishedAt for that locale and document.

oiorain:
Hi @ankit7201 

Thank you for the changes and apologies to keep you waiting on this. I missed the notification. 
We will get this reviewed very soon. 🙏

sergejpapper:
Hello,

maybe it is clear when this functionality will be released in stable version (this month, next month, ...)?

This change is very important for our organization, and we think between waiting your (Strapi) solution or making our custom.

Thank you.

oiorain:
Hi @sergejpapper !

This should be soon but I can't say for sure. I think we should be good for next month. The person who's done the reviews so far is off so they should be able to get it reviewed one last time next week. 
Sorry for the delay. We're activitely working on making community PR reviewed faster. :)

packages/core/core/src/domain/content-type/index.ts
```diff
@@ -50,6 +51,8 @@ const createContentType = (uid: string, definition: ContentTypeDefinition) => {
 
   addCreatorFields(schema);
 
+  addFirstPublishedAt(schema);
```
innerdvations, line 54:
```suggestion
  if (strapi.config.get('admin.firstPublishedAtField.enabled', false)) {
    addFirstPublishedAt(schema);
  }
```

Although it's probably better suited inside `addDraftAndPublish` since it's dependent on that

innerdvations: For now this is ok, but I've started a discussion internally about adding a "preserve_attributes" similar to our "preserve_tables" in order to prevent this data from being deleted when the config value is not set.

Essentially, we won't add the attribute until the config value is set BUT once it has been enabled, we _probably_ want it to be kept and not deleted even if disabled. But we will resolve that internally and make another PR later, just want to leave a note here for future reference :)

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -219,6 +221,10 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
       i18n.localeToData(contentType)
     )(params);
 
+    console.log('===============');
+    console.log(queryParams);
+    console.log('===============');
+
```
innerdvations:
```suggestion
```

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -314,20 +320,26 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
     // Delete old published versions
     await async.map(oldPublishedVersions, (entry: any) => entries.delete(entry.id));
 
+    // Add firstPublishedAt to draft if it doesn't exist
+    // await async.map(draftsToPublish, (draft: any) => entries.update(draft))
```
innerdvations:
```suggestion
```

packages/core/core/src/domain/content-type/index.ts
```diff
@@ -82,6 +85,20 @@ const addDraftAndPublish = (schema: Schema.ContentType) => {
...
+    strapi.config.get('admin.firstPublishedAtField.enabled', false) &&
+    _.get(schema, 'options.draftAndPublish', false);
+
+  schema.attributes[FIRST_PUBLISHED_AT_ATTRIBUTE] = {
+    type: 'datetime',
+    configurable: false,
+    writable: true,
+    visible: false,
+    private: !isNonPrivate,
+  };
```
innerdvations:
```suggestion
  const isEnabled =
    strapi.config.get('admin.firstPublishedAtField.enabled', false) &&
    _.get(schema, 'options.draftAndPublish', false);

  schema.attributes[FIRST_PUBLISHED_AT_ATTRIBUTE] = {
    type: 'datetime',
    configurable: false,
    writable: true,
    visible: false,
    private: !isEnabled,
  };
```

tests/api/plugins/users-permissions/content-api/users-api.test.api.js
innerdvations, line 1: You can revert this file once you fix the enabled check.

tests/api/core/content-manager/search.test.api.js
innerdvations, line 1: You can revert this file once you fix the enabled check, and in fact this will ensure that we don't accidentally enable the feature without the config.

packages/core/core/src/domain/content-type/index.ts
ankit7201: Hi @innerdvations 
Thanks for the review. 

A little confused....
So, we won't be adding the if condition to check if firstPublishedAtField is enabled or not (as per your suggestion above) for now, right? 

Because if we add this if condition and then someone switches off this feature and restarts the app, the firstPublishedAt data will disappear (But the expectation is that it should be preserved so that if someone re-enables it, it should show the old data)

So, the id condition wouldn't make sense as of now, correct?

innerdvations: Sorry, I wasn't very clear. What I mean is:

**For now**: we need to add the `if` condition that enables/disables it and will accept that data gets removed if it becomes disabled

**For later**: We will add a  mechanism to ensure the data gets kept

Of course, if you'd like to go ahead and do it here now that would be great 😆

ankit7201: > Of course, if you'd like to go ahead and do it here now that would be great

Sure. Are we talking about "preserve_attributes"? 
If you could let me know more about this, I can take a look into this. 

Or, do you think we should get this PR merged and I can pick it up as a separate PR. 

Let me know what works. 
Thanks.

ankit7201:
Hi @innerdvations 

I have updated the code with the review suggestions. Please take a look whenever you get a chance. 

No tests added yet (Based on your comment, I guess we will be releasing this an experimental feature for now)

Thank you. 

CC: @oiorain

packages/core/core/src/domain/content-type/index.ts
innerdvations: While it's experimental, it's ok if the data just gets dropped when the feature is disabled. That's another part we will have to implement before we can call it stable though.

BUT, in case you or anyone else is interested, here's some context for a later PR:

In the strapi/database package, there is a sync method that ensures only data that in the strapi maintained tables, only current valid schema data exists, and it deletes anything that isn't recognized. So when this feature is disabled, the firstPublishedAt column will be dropped.

We have a solution already implemented for full tables, which we use for EE features to ensure that an expired/invalid/missing license won't cause data loss, by using a "preserve_tables" array of those table names.

We will need to add something just like that but for columns/attributes, so that we can keep an array of attributes that should never be deleted from the database even if they don't appear in a schema anymore.

innerdvations:
@ankit7201 I hope you don't mind, I just pushed some changes to your branch to speed things along so we don't have to keep going back and forth!

Here's what I changed:

- use the content type utils as the source of truth for if the field is enabled or not
- do not export the utils types in favor of using the compatible `Schema.ContentType` that already exists in @strapi/types
- rename the configuration option to `features.future.experimental_firstPublishedAt` after some further internal discussion
- adding a warning log for now to show which fields have it enabled
 
I'm going to have someone else on the team QA it this week and then we'll finally get this merged!

innerdvations:
@pwizla This is finally ready to go. Could you add this to the documentation?

For some quick context so you don't have to read the whole thread:

- in config/features.js setting `future.experimental_firstPublishedAt:true` will add a new attribute to all content types with d&p enabled that will start recording the first time something was published, which will stay constant even if the article is unpublished 
- This is an EXPERIMENTAL feature -- it should be noted that if the feature is ever disabled, you will lose the data in the firstPublishedAt field

innerdvations (approved):
LGTM! Will merge following a final round of manual QA

pwizla:
Perfect, I've just [documented](https://github.com/strapi/documentation/pull/2636) it, @innerdvations. Thank you!

PR #23169, opened 2025-03-17
fix: correctly read asset stream for dts asset remote providers

### LINKED ISSUE #16473: Transfer feature hangs on asset transfer

- Node.js version: 16+
- NPM version: 8+
- Strapi version: 4.9.2

### Describe the bug

Edit: Based on issue #16437 this does not seem to be limited to cloud.

When attempting to transfer from a local Strapi instance to a Strapi Cloud project, using:

`yarn strapi transfer --to <url> --to-token <token> `
or
`npm run strapi transfer --to <url> --to-token <token>`

The transfer hangs when it gets to the `assets` transfer. Only killing the process will stop it, which then produces an error.

1. Go to a local Strapi project, running on 4.8+
2. Login to Strapi Cloud and ensure both local and cloud projects have same schema
3. Try to transfer from local to cloud using either:
`yarn strapi transfer --to <url> --to-token <token> `
or
`npm run strapi transfer --to<url> --to-token <token>`

### Expected behavior

That the transfer would go through successfully. 

### Screenshots

image

image

### LINKED ISSUE #22038: strapi transfer hangs forever on asset transfer (Strapi 5.1)
- Node.js version: 20.14.0
- NPM version: 10.8.3
- Strapi version: 5.1.0
- Database: Postgres
- Operating system: Mac OS Sonoma 14.6.1 (darwin-arm64)

### Describe the bug

When attempting to execute a transfer from a Cloud Strapi instance to a local Strapi project, using:

`yarn strapi transfer --from <cloud-strapi-url>` 

The transfer starts executing correctly after entering the token, downloads the entities, but then hangs forever at 0 when dealing with assets.

This problem was described in a closed issue with Strapi 4 users (issue [#16473](https://github.com/strapi/strapi/issues/16473)).

We reproduced the same error exactly on three different computers (Macbook), reinstalling a local version from the Github repo synced to the cloud instance. The size of our cloud assets (entities, media library) is very reasonable. 

1. Open terminal on local instance of Strapi
2. Run `yarn strapi transfer --from <cloud-strapi-url>`
3. See entities getting transferred 
4. Then it hangs forever 

### Expected behavior

The download of the assets should start, or it should be retried in case of a timeout, or at least a clear error message should be displayed after a reasonable amount of time. 

### Screenshots

screenshot-strapi

Bassel17:
### What does it do?

- res.body is already a readable stream no need to convert it

### Why is it needed?

- reading assets streams from remote providers is not working and getting stuck

### How to test it?

experimental: `0.0.0-experimental.9678892591ed5d28450369e1511b3d21cb7db4f9`

- test a pull transfer against a project hosted on cloud
- or setup a project with aws or cloundinary providers and try to pull

### Related issue(s)/PR(s)

possibly fixes #16473 
fixes #22038

### DIFF
--- a/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
@@ -29,7 +29,7 @@ function getFileStream(
 
       if (res.body) {
         // pipe the image data
-        Readable.fromWeb(new webStream.ReadableStream(res.body)).pipe(readableStream);
+        Readable.fromWeb(res.body as any).pipe(readableStream);
       } else {
         readableStream.emit('error', new Error('Empty data found for file'));
       }
--- a/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
@@ -1,6 +1,5 @@
 import { join } from 'path';
 import { Duplex, PassThrough, Readable } from 'stream';
-import * as webStream from 'stream/web';
 import { stat, createReadStream, ReadStream } from 'fs-extra';
 import type { Core } from '@strapi/types';
 
--- a/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
@@ -1,6 +1,7 @@
 import { join } from 'path';
 import { Duplex, PassThrough, Readable } from 'stream';
 import { stat, createReadStream, ReadStream } from 'fs-extra';
+import * as webStream from 'stream/web';
 import type { Core } from '@strapi/types';
 
 import type { IAsset, IFile } from '../../../../types';
@@ -28,7 +29,7 @@ function getFileStream(
 
       if (res.body) {
         // pipe the image data
-        Readable.fromWeb(res.body as any).pipe(readableStream);
+        Readable.fromWeb(res.body as webStream.ReadableStream<Uint8Array>).pipe(readableStream);
       } else {
         readableStream.emit('error', new Error('Empty data found for file'));
       }


packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
```diff
@@ -29,7 +29,7 @@ function getFileStream(
 
       if (res.body) {
         // pipe the image data
-        Readable.fromWeb(new webStream.ReadableStream(res.body)).pipe(readableStream);
+        Readable.fromWeb(res.body as any).pipe(readableStream);
```
Convly, line 32: The cast to `any` means that `res.body` is not be fully compatible with `Readable.fromWeb`, are we sure we can do this without breaking anything? What's the actual type of `res.body`?

Bassel17: it is `ReadableStream<Uint8Array<ArrayBufferLike>>`
and the method expects `ReadableStream<any>`
yeah I mean it was broken so it can't break anything technically

Bassel17:
experimental for testing:
`0.0.0-experimental.9678892591ed5d28450369e1511b3d21cb7db4f9`

packages/core/data-transfer/src/strapi/providers/local-source/assets.ts
Convly: I've checked some documentation, and it seems it's mainly a type mismatch due to node/dom implementation but they should indeed be compatible.

Could you at least remove the any and replace it with something like
```suggestion
        Readable.fromWeb(res.body as webStream.ReadableStream<Uint8Array>).pipe(readableStream);
```

innerdvations: That doesn't seem to work either. I think I would prefer a `@ts-expect-error` though over `as any`.

From what I can tell this is a known type incompatibility in a usage where the objects themselves are actually compatible. Then at some point hopefully the types will get fixed and we can remove it.

Convly: >That doesn't seem to work either.

What do you mean by it doesn't work either ? :thinking: 
image
image
image

> From what I can tell this is a known type incompatibility in a usage where the objects themselves are actually compatible. Then at some point hopefully the types will get fixed and we can remove it.

yep

Bassel17: It does work I changed the casting to webStream.ReadableStream<Uint8Array> that seems to satisfy it

innerdvations: Ahh, yeah, it does work, nevermind 😆

innerdvations (approved):
Yay! LGTM!

PR #23171, opened 2025-03-17
fix: run cronjobs closer to app start

Marc-Roig:
### What does it do?

Upload cronjob take 5 minutes until it first, and its too much time for cloud environments that shutdown the container if there is no use.

This PR reduces this time.

### DIFF
--- a/packages/core/review-workflows/server/src/services/metrics/weekly-metrics.ts
+++ b/packages/core/review-workflows/server/src/services/metrics/weekly-metrics.ts
@@ -60,7 +60,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       let weeklySchedule = currentSchedule;
 
       if (!currentSchedule || !lastWeeklyUpdate || lastWeeklyUpdate + ONE_WEEK < now.getTime()) {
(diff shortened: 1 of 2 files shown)


packages/core/review-workflows/server/src/services/metrics/weekly-metrics.ts
```diff
@@ -60,7 +60,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       let weeklySchedule = currentSchedule;
 
       if (!currentSchedule || !lastWeeklyUpdate || lastWeeklyUpdate + ONE_WEEK < now.getTime()) {
-        weeklySchedule = getWeeklyCronScheduleAt(add(now, { seconds: 10 }));
+        weeklySchedule = getWeeklyCronScheduleAt(add(now, { seconds: 15 }));
```
simotae14, line 63: why did you change also this value? just for curiosity

simotae14 (approved):
it seems ok to me

packages/core/review-workflows/server/src/services/metrics/weekly-metrics.ts
Marc-Roig: to keep it consistent with the upload metrics, 10 seconds seemed to little

Issue #23099, opened 2025-03-09
`Cannot destructure property 'ProtectedEditPage' of '(intermediate value)' as it is undefined.` On production build webhooks detail page.


laurenskling:
### Node Version

18

### NPM/Yarn/PNPM Version

yarn 1.22

### Strapi Version

5.11.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

It does not happen in dev mode (which is weird) but on my production build (also when i locally build and run) I hit this error when moving to a webhooks detailpage.

```
TypeError: Cannot destructure property 'ProtectedEditPage' of '(intermediate value)' as it is undefined.
    at Object.lazy ....
```

### Steps to Reproduce

1. production build 5.11.0
2. go to webhooks
3. click an existing (or create a new one, it will reload into this error)
4. boom

### Expected Behavior

no boom

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

mheiduk:
I get this as well with v5.11.0 with PSQL, in v5.10.4 everything works as expected.

Johntyb:
Same here. v5.11.0. Although the webhooks still work, you just cant edit them once created.

GoodDD:
Same issue for me. v5.11.0

davimdantas:
Same issue for me. v5.11.2

Decryptu:
I'm experiencing the same issue with Strapi v5.11.2 using Node version 20.18.3. The error occurs when trying to open an existing webhook in the admin panel, showing:

```
TypeError: Cannot destructure property 'ProtectedEditPage' of '(intermediate value)' as it is undefined.
    at Object.lazy (admin/strapi-aS45AdIn.js:2406:5457)
    at async ogt (admin/strapi-aS45AdIn.js:2225:3694)
    at async ugt (admin/strapi-aS45AdIn.js:2225:5353)
    at async Promise.all (index 0)
    at async lgt (admin/strapi-aS45AdIn.js:2225:4195)
    at async cgt (admin/strapi-aS45AdIn.js:2225:4630)
    at async nt (admin/strapi-aS45AdIn.js:2224:28529)
    at async Nt (admin/strapi-aS45AdIn.js:2224:29216)
[...]

PR #23179, opened 2025-03-19
fix: On production build webhooks detail page intermediate value issue.

seenasingh30:
### What does it do?

Updated the import of ProtectedEditPage in the webhook module to prevent failure in the production build. Introduced delayed import by first assigning the module to a const and resolving it afterwards.

### Why is it needed?

It solves the issue : Cannot destructure property 'ProtectedEditPage' of '(intermediate value)' as it is undefined. On production build webhooks detail page.

### How to test it?

I validated this by creating a webhook in an older version and attempting to update it. Before applying this fix, I encountered the error:
"Cannot destructure property 'ProtectedEditPage' of '(intermediate value)' as it is undefined."
After implementing the fix, the issue is resolved.

### Related issue(s)/PR(s)

It is related to #23099.

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/constants.ts
+++ b/packages/core/admin/admin/src/pages/Settings/constants.ts
@@ -73,10 +73,10 @@ export const ROUTES_CE: RouteObject[] = [
   },
   {
     lazy: async () => {
-      const { ProtectedEditPage } = await import('./pages/Webhooks/EditPage');
+      const editWebhook = await import('./pages/Webhooks/EditPage');
 
       return {
-        Component: ProtectedEditPage,
+        Component: editWebhook.ProtectedEditPage,
       };
     },
     path: 'webhooks/:id',


Link2Twenty:
Won't this just move the error to line 79?

Something like this
> TypeError: Cannot read properties of undefined (reading 'ProtectedEditPage')

I've not tested it so can't be certain but that's what I'd expect

packages/core/admin/admin/src/pages/Settings/constants.ts
```diff
@@ -73,10 +73,10 @@ export const ROUTES_CE: RouteObject[] = [
   },
   {
     lazy: async () => {
-      const { ProtectedEditPage } = await import('./pages/Webhooks/EditPage');
+      const editWebhook = await import('./pages/Webhooks/EditPage');
 
       return {
-        Component: ProtectedEditPage,
+        Component: editWebhook.ProtectedEditPage,
```
alexandrebodin, line 79: if the object is undefined, access the property shouldn't fix anything 🤔  you would just get cannot access property of undefined errors instead of destructuring

seenasingh30: This implementation has been thoroughly tested. I explored multiple approaches before finalizing this solution to ensure stability and prevent errors in the production build. Let me know if you have any suggestions! @alexandrebodin

Link2Twenty: Interestingly, it does actually work.. Which suggests a race condition.

I've had a quick look and can't see anything that should make the export not work right. I've even tried loading a different component on that page and that works.

seenasingh30: I’ve tested this implementation across multiple cases, and it consistently resolves the issue. But if there’s a better way to handle this, I’m happy to explore further.

alexandrebodin: The build for this specific route is generating different code than the usual ones. I'm looking into it as I think we should fix it differently.

seenasingh30: Thanks for looking into it! Let me know what you find, and I’ll be happy to adjust the solution accordingly!

alexandrebodin: Ok, so we have a compatbility issue between our rollup build of the package and the vite build of the final app. Vite is basically dismissing a part of the rollup import in charge of resolving the named export. I don't have an obvious fix so we will go with this. thank you !

alexandrebodin (approved):

PR #23183, opened 2025-03-19
fix: issue with layout property in InputRenderer

simotae14:
### What does it do?

Fix the access of layout property from an undefined object.
It happens sometimes when you try to open a relation inside a document

### Why is it needed?

To avoid the error saying "not possible to access layout from undefined"
As you can see in this video

https://github.com/user-attachments/assets/c5797b0b-f160-4657-8586-1bedf2681457

### How to test it?

You can try to follow this video

https://github.com/user-attachments/assets/357c0a0f-2c29-4ff6-8961-e314fdd2e538

### Related issue(s)/PR(s)

CS-1374

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -6,13 +6,12 @@ import {
   InputRenderer as FormInputRenderer,
   useField,
 } from '@strapi/admin/strapi-admin';
-import { ContentTypeKind } from '@strapi/types/dist/struct';
 import { useIntl } from 'react-intl';
 
 import { SINGLE_TYPES } from '../../../constants/collections';
 import { useDocumentRBAC } from '../../../features/DocumentRBAC';
 import { useDoc, UseDocument } from '../../../hooks/useDocument';
-import { useDocumentLayout } from '../../../hooks/useDocumentLayout';
+import { useDocLayout, useDocumentLayout } from '../../../hooks/useDocumentLayout';
 import { useLazyComponents } from '../../../hooks/useLazyComponents';
 
 import { BlocksInput } from './FormInputs/BlocksInput/BlocksInput';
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -41,8 +40,9 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
  */
 const InputRenderer = ({ visible, hint: providedHint, document, ...props }: InputRendererProps) => {
   const { model: rootModel } = useDoc();
+  const docLayout = useDocLayout();
   const documentLayout = useDocumentLayout(document.schema?.uid ?? rootModel);
-  const components = documentLayout.edit.components;
+  const components = { ...documentLayout.edit.components, ...docLayout.edit.components };
```
remidej, line 45: I'm a bit confused by this. Doesn't it mean that we're overwriting the layout specific to the content type in a modal with the layout of the content type from the edit view page underneath?

I also feel like documentLayout will _always_ be overwritten by docLayout, making it not necessary

simotae14: so you mean I can reverse the order of the merging?

simotae14: we did something similar in the Dynamic component btw https://github.com/strapi/strapi/pull/22997/commits/4bbf30d43befd671c5a6201ce6afedc6b9a38fef

remidej: Thanks for the link, it makes it clearer for me. Could you do the same thing as over there when it comes to variable names? and the order shouldn't matter in the end

simotae14: done

remidej (approved):

PR #23188, opened 2025-03-19
enhancement: admin session presists across tabs

Bassel17:
### What does it do?

- replaces `sessionStorage` with session cookies when logging in and saving the jwt token

### Why is it needed?

- so we can preserve login state across different tabs or windows unless the browser was closed

### How to test it?

- follow this acceptance criteria:

- As a user logging into the admin
    - when I log in and have not checked the “Remember me” setting
      - then I want to be logged in
    - when I open the admin panel in another tab or window of the same browser
      - then I want to be automatically logged in and access the admin panel
    - when I close the browser and reaccess the admin
      - then I want to be logged out and be requested to login

### Related issue(s)/PR(s)

DX-1461


jhoward1994 (approved):
Works for me 👍🏻

laurenskling:
Thank you for this :) Opening a new tab and having to log in again was so annoying.

PR #23190, opened 2025-03-19
fix: add headers for email-nodemailer provider

ruslan-zhuba:
### What does it do?

I've added ability to pass headers into the email-nodemailer provider. 

### Why is it needed?

I'm having a SMTP server. I need to pass headers as part of Business requirements for that SMTP server. 

### How to test it?

Pass custom headers into `send` function as described into `README.md`. 

### Related issue(s)/PR(s)

N/A


ruslan-zhuba:
Any updates?

PR #23191, opened 2025-03-20
chore: upgrade prismjs to 1.30.0

innerdvations:
### What does it do?

Upgrades prismjs to 1.30.0

### Why is it needed?

Remove warnings for CVE-2024-53382 -- in Strapi's usage there does not appear to be an actual vulnerability

### How to test it?

syntax highlighting should still work

### Related issue(s)/PR(s)

DX-1951


Bassel17 (approved):
LGTM!

PR #23192, opened 2025-03-20
security: bump locked babel versions to 7.26.10

innerdvations:
### What does it do?

IMPORTANT: This PR could technically present a breaking change for a small number of users, because it causes changes in some subtle javascript behaviours, like async run order, and build changes like white-spacing differences; things which are unsafe to rely on but nonetheless sometimes might be used by user code. This should be released in a minor version of Strapi along with a release note.

- Updates all babel related packages in yarn.lock
- Updates doc service api tests that broke from it, due to assertions on async events that weren't being awaited; without these changes, the assertions don't happen until after the test completes, causing it to fail
- Updates upgrade tool unit tests that broke from it
- Updates front-end test snapshots because there were several minor changes like whitespace

### Why is it needed?

Resolves CVE-2025-27789 -- Strapi itself does not appear to be vulnerable, although project code that directly passes user input strings to the second param of a regexp are potentially vulnerable

### How to test it?

### Related issue(s)/PR(s)

DX-1952


innerdvations:
For some reason, this bump is causing document service events tests to fail, investigating it now

Marc-Roig (approved):
looks good!

innerdvations:
front-end tests are failing from this as well

Bassel17:
to be resolved in #23391

PR #23198, opened 2025-03-20
enhancement: add tracking for editing relations on-the-fly

simotae14:
### What does it do?

Add the property fromRelationModal to true when we trigger the events didEditEntry and didPublishEntry

### Why is it needed?

to Distinguish between the Save and Publish events in the Relation modal and the ones outside it

### How to test it?

Try to save and modify an entry in the Relation modal and do the same in the Edit View and check if the event property fromRelationModal is set to true in the first case and to false in the other cases

### Related issue(s)/PR(s)

CS-1313


simotae14:
Closed in favour of https://github.com/strapi/strapi/pull/23206 that contains event for the ROTF and Preview page

PR #23202, opened 2025-03-20
enhancing data transfer CLI command

maccomaccomaccomacco:
### What does it do?
Adding support for the `strapi transfer` command; running this command the user can chose the kind of operation to run and insert the needed data (url and token). Furthermore, the URL and token values fall back to dedicated environment variables value.

Kapture 2025-03-20 at 23 33 55

### Why is it needed?
Inserting the many and long options needed to transfer/push or transfer/pull data can be confusing and difficult; and it's so even more when the transfer is run iteratively.

### How to test it?
Just run the `strapi transfer` command and enjoy the journey.
Set the `STRAPI_TRANSFER_URL` and `STRAPI_TRANSFER_TOKEN` env variables to enjoy it even more.


maccomaccomaccomacco:
wrong branch created from main, not from develop

PR #23206, opened 2025-03-21
enhancement: add tracking for editing relations on-the-fly and preview

simotae14:
### What does it do?

Add the property fromRelationModal to true when we trigger the events didEditEntry and didPublishEntry
Add the property fromPreview to true when we trigger the events didEditEntry and didPublishEntry in the Preview page
and Add the properties fromRelationModal and fromPreview to true when we trigger the events didEditEntry and didPublishEntry in a relation modal inside the Preview page

### Why is it needed?

to Distinguish between the Save and Publish events in the Relation modal and in the Preview page

### How to test it?

Try to save and modify an entry in the Relation modal and do the same in the Edit View and check if the event property fromRelationModal is set to true in the first case and to false in the other cases
Try to save and modify an entry in the Relation modal inside a Preview page and do the same in the Edit View and check if the event property fromRelationModal and fromPreview are set to true in the first case and to false in the other cases
Try to save and modify an entry in the Preview page and do the same in the Edit View and check if the event property fromPreview is set to true in the first case and to false in the other cases

### Related issue(s)/PR(s)

CS-1313
CS-1368

### DIFF
--- a/packages/core/content-manager/admin/src/hooks/useDocumentActions.ts
+++ b/packages/core/content-manager/admin/src/hooks/useDocumentActions.ts
@@ -58,7 +58,7 @@ type BulkOperationResponse<TResponse extends { data: any; error?: any }> =
   | Pick<TResponse, 'data'>
   | { error: BaseQueryError | SerializedError };
 
-type UseDocumentActions = () => {
+type UseDocumentActions = (fromPreview?: boolean) => {
   /**
    * @description Attempts to clone a document based on the provided sourceId.
    * This will return a list of the fields as an error if it's unable to clone.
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/hooks/useDocumentActions.ts
```diff
@@ -189,7 +192,7 @@ type IUseDocumentActs = ReturnType<UseDocumentActions>;
  *
  * @see {@link https://contributor.strapi.io/docs/core/content-manager/hooks/use-document-operations} for more information
  */
-const useDocumentActions: UseDocumentActions = () => {
+const useDocumentActions: UseDocumentActions = (fromPreview = false, fromRelationModal = false) => {
```
remidej, line 195: Once my refactoring PR is merged we could avoid both of these by relying to context to infer them. Do you think this tracking PR could wait until then?

Marc-Roig: is this hook only used in preview/relation modal?

remidej: sure we can merge then. no need for todos i won't forget

remidej (approved):

packages/core/content-manager/admin/src/hooks/useDocumentActions.ts
simotae14: ok thanks

PR #17160, opened 2023-06-30
Show localhost in URL for IPv6 addresses

mirkonasato:
### What does it do?

Treat `::` like `0.0.0.0` and `::1` like `127.0.0.1` when showing the local Strapi URL.

### Why is it needed?

By default Strapi listens on `0.0.0.0` i.e. only on IPv4. This is a problem because Node v18 (current LTS) [defaults to IPv6](https://github.com/nodejs/node/issues/40537). So if you're running another Node process, say a Next.js app, that tries to connect to Strapi it will probably fail with `ECONNREFUSED ::1:1337`. (_Probably_ because it depends on how exactly the network is configured on your machine.)

That's why I set `HOST=::` in `.env`. `::` the IPv6 equivalent of `0.0.0.0`.

But that way Strapi shows an invalid URL in the logs, which won't work if somebody tries to open it in the browser. Either `strapi develop` or `strapi start` print:
```
To manage your project 🚀, go to the administration panel at:
http://:::1337/admin

To access the server ⚡️, go to:
http://:::1337
```

This change should make it show `http://localhost:1337`, like it already does for `0.0.0.0`.

### How to test it?

Start a Strapi app and see what URL it prints to the console.

### Related issue(s)/PR(s)

#12285

Issue #12285, opened 2022-01-24
Strapi doesn't work with IPv6 addresses properly


mkg20001:
### Describe the bug

Using IPv6 address as server address makes things fail left and right

### Steps to reproduce the behavior

1. Edit 'config/server.js'
2. Set host to `::1`
3. `strapi develop` and `strapi develop --watch-admin`

strapi develop:
The message printed contains 
```
Welcome back!
To access the server ⚡️, go to:
http://::1:1337
```

`http://::1:1337` is an invalid address, it should be `http://[::1]:1337`

strapi develop --watch-admin:
It opens `http://localhost:8080/` which eventually loads only to give the error `Warning: An error occurred while requesting the API` (why is a fatal error a warning?)

It keeps loading ad-infinitum. 

Console shows
```
main.bundle.js:40325 DOMException: Failed to execute 'open' on 'XMLHttpRequest': Invalid URL
    at dispatchXhrRequest (http://localhost:8080/admin/main.bundle.js:36835:13)
    at new Promise (<anonymous>)
    at xhrAdapter (http://localhost:8080/admin/main.bundle.js:36806:10)
    at dispatchRequest (http://localhost:8080/admin/main.bundle.js:37600:10)
```

The URL it tries to load is `http://::1:1337/admin/project-type`. Again, improper IPv6 handling.

### Expected behavior

It just works with IPv6 addresses

### Screenshots

![image](https://user-images.githubusercontent.com/7735145/150779552-7437b462-9979-4d38-9108-5975b48a298d.png)

### Code snippets

this is a good stopgap, no idea if that's going to break at some edge case

```js
if (ip.indexOf(':') != -1) ip = `[${ip}]`
```

### System

- Node.js version: v17.3.0
- NPM version: 8.3.0
- Strapi version: 4.0.4
- Database: sqlite3
- Operating system: nixos 22.05beta

derrickmehaffy:
> Node.js version: v17.3.0

While I don't think this will matter, can you please test on a valid supported version of node? We don't support Node 17 or any odd version.

mkg20001:
v16.13.1, same errors

mirkonasato:
Similarly, if you set `HOST=::` in `.env` then `strapi develop` or `strapi start` will print

```
Create your first administrator 💻 by going to the administration panel at:

┌──────────────────────┐
│ http://:::1337/admin │
└──────────────────────┘
```
which won't work.

[...]

PR #23212, opened 2025-03-22
fix: add support for IPv6 addresses

chsanch:
### What does it do?

Treat `::` like `0.0.0.0 `and `::1` like `127.0.0.1` when showing the local Strapi URL. This PR is basically the same as #17160 

### Why is it needed?

By default Strapi listens on `0.0.0.0` i.e. only on IPv4. This is a problem because Node v18 (current LTS) https://github.com/nodejs/node/issues/40537. So if you're running another Node process, say a Next.js app, that tries to connect to Strapi it will probably fail with `ECONNREFUSED ::1:1337`. (Probably because it depends on how exactly the network is configured on your machine.)

That's why I set `HOST=::` in .env. `::` the IPv6 equivalent of `0.0.0.0`.

But that way Strapi shows an invalid URL in the logs, which won't work if somebody tries to open it in the browser. Either strapi develop or strapi start print:

```
To manage your project 🚀, go to the administration panel at:
http://:::1337/admin

To access the server ⚡️, go to:
http://:::1337
```
This change should make it show `http://localhost:1337`, like it already does for `0.0.0.0`.
### How to test it?

Start a Strapi app and see what URL it prints to the console.

### Related issue(s)/PR(s)

#12285

### DIFF
--- a/packages/core/core/src/configuration/urls.ts
+++ b/packages/core/core/src/configuration/urls.ts
@@ -79,7 +79,8 @@ const getAbsoluteUrl =
 
     const serverConfig = config.server as ServerConfig;
     const hostname =
-      config.environment === 'development' && ['127.0.0.1', '0.0.0.0'].includes(serverConfig.host)
+      config.environment === 'development' &&
+      ['127.0.0.1', '0.0.0.0', '::1', '::'].includes(serverConfig.host)
         ? 'localhost'
         : serverConfig.host;
 


packages/core/core/src/configuration/urls.ts
```diff
@@ -79,7 +79,8 @@ const getAbsoluteUrl =
 
     const serverConfig = config.server as ServerConfig;
     const hostname =
-      config.environment === 'development' && ['127.0.0.1', '0.0.0.0'].includes(serverConfig.host)
+      config.environment === 'development' &&
```
innerdvations, line 82:
```suggestion
      ['127.0.0.1', '0.0.0.0', '::1'].includes(serverConfig.host)
```
Not sure if we should include `::` since it's not technically localhost. It may cause confusion somewhere.

Maybe we could do something like:

```
import { isIp } from 'net';

// getAbsoluteUrl
const getAbsoluteUrl =
  (adminOrServer: 'admin' | 'server') =>
  (config: Record<string, unknown>, forAdminBuild = false) => {
    const { serverUrl, adminUrl } = getConfigUrls(config, forAdminBuild);
    const url = adminOrServer === 'server' ? serverUrl : adminUrl;

    if (url.startsWith('http')) {
      return url;
    }

    const serverConfig = config.server as ServerConfig;

    const isLocalhost =
      config.environment === 'development' &&
      ['127.0.0.1', '0.0.0.0', '::1'].includes(serverConfig.host);

    if (isLocalhost) {
      return `http://localhost:${serverConfig.port}${url}`;
    }

    const isIPv6 = isIP(serverConfig.host) === 6;

    // For raw IPv6 addresses, we need to use brackets
    if (isIPv6) {
      return `http://[${serverConfig.host}]:${serverConfig.port}${url}`;
    }

    // For hostnames and IPv4 addresses, we don't need to use brackets
    return `http://${serverConfig.host}:${serverConfig.port}${url}`;
  };
```

I tested it with `::1` and `::`, could you please try it out and make sure it still solves the problem for you too?

chsanch: @innerdvations I have added the changes you have suggested, but I'm not really sure how to test this in my local environment, I've followed the steps in the `CONTRIBUTING.md` file, I have started the `getstarted` app and it seems to run fine.

davimdantas: > @innerdvations I have added the changes you have suggested, but I'm not really sure how to test this in my local environment, I've followed the steps in the `CONTRIBUTING.md` file, I have started the `getstarted` app and it seems to run fine.

Wouldn't this be enough?

`curl -g -6 http://\[::\]:1337/`

`curl -g -6 http://\[::1\]:1337/ `

innerdvations: If you've tried the getstarted app that should be fine, I just wanted to make sure it still worked for you as well. Thanks for the contribution!

innerdvations:
> [CLA assistant check](https://cla.strapi.io/strapi/strapi?pullRequest=23212) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.strapi.io/strapi/strapi?pullRequest=23212) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.strapi.io/check/strapi/strapi?pullRequest=23212) it.

@chsanch Could you please sign the contributor license agreement shown here so I can approve and merge this?

packages/core/core/src/configuration/urls.ts
hsource: Thanks for this patch! This resolves an issue in our upgraded codebase too.

One tiny nit: I think we should include `::` for the same reason that `0.0.0.0` is included. If we're binding to `::`, it means to all addresses, but we choose to show localhost since it's more natural than typing in 0.0.0.0.

Just my 2 cents - honestly OK either way! I just think we should either include both `0.0.0.0` and `::` or exclude both of them.

innerdvations: Yeah, I think you're right actually, I didn't consider that we already include `0.0.0.0`. And if `::` is used localhost will work.

innerdvations: I just updated it and will merge as soon as CI checks pass

PR #23221, opened 2025-03-24
enhancement: add advanced preview license check

remidej:
### What does it do?

adds a license check to display the preview side editor only if cms-advanced-preview is true

### How to test it?

Without a license, you shouldn't see the side editor.

With an EE license you should


packages/core/core/src/ee/license.ts
```diff
@@ -111,6 +112,7 @@ const fetchLicense = async (
         throwError();
     }
   } else {
+    console.log('validate');
```
simotae14, line 115: you forgot this log

PR #23223, opened 2025-03-24
enhancement: create injection zones for edit view components

davimdantas:
### What does it do?

- Creates injection zones for the Dynamic Zone and Repeatable Components action areas.
- Adds a usable example demonstrating these injection zones.

### Why is it needed?

Enable developers to create plugins/inject logic into those components

### How to test it?

```bash
cd ./examples/getstarted
yarn develop
```

this collections has examples of booth components
- http://localhost:1337/admin/content-manager/collection-types/api::address.address?page=1&pageSize=10&sort=postal_code:ASC
- http://localhost:1337/admin/content-manager/collection-types/api::kitchensink.kitchensink?page=1&pageSize=10&sort=short_text%3AASC

repeatable

dynamic


rafael-couto-p (commented):
Great work!

mathildeleg:
Thank you for the work on this 
We are closing the PR because the change set mixes new injection zones with a getstarted-only duplicate example, and the duplicate flow does not match how the edit view already behaves: for example, when a repeatable array grows, we only auto-expand the last row, so a duplicate inserted earlier stays collapsed and the wrong row can stay open. Dynamic zone blocks do not get a supported “open the new item” path from core either, so the UX feels incomplete. 
We also do not want to ship the example as the main reference: it uses unstable admin APIs, wraps a single control in `DesignSystemProvider`, and relies on a narrow `Jest` preset workaround. Injection zones here are still a good idea, but we would prefer them in a small, focused PR with a clear, typed contract for injected props, and any duplicate behavior either first-party in content-manager or documented separately—not bundled as the primary deliverable of this branch.
Feel free to open another PR for this if you'd like to 🙏🏻 I'll check with Product but I don't think it's on the roadmap for us for now :)
Thanks again

Issue #22205, opened 2024-11-19
Register User Endpoint needs id for relations instead of documentId


pmcroOhler:
### Required System information

Node.js version: 18.20.4
NPM version: 10.7.0
Strapi version: 5.3.0
Database: sqlite
Operating system: Mac OS / homebrew
Is your project Javascript or Typescript: JS

### Describe the bug

Using auth.register on Rest API we passed a documentId of a relation to connect ist while user is registered.
We get a sqLite Database Error. When we pass Id instead of documentId everything works fine.

### Steps to reproduce the behavior

1. Create Collection Type
2. Create an Entry in collection type
3. Create a ManyToOne Relation inside the collection type with user
4. Register a user via REST API and try to connect the entry (POST /api/auth/local/register)

```
{
...
  collectionTypePluralName : [ "CollectionTypesEntryDocumentId" ],
}
```

### Expected behavior

We expected that the documentId is needed for connecting relations on register user. Instead the Id is needed for now.

nevil2:
My experience is that the user permissions plugin still requires Id rather than documentId.

To change it now would be a **breaking change** unless it could work with either by evaluating whether the id is an integer or a string (e.g. `parseInt(id) === +id`)

PR #23226, opened 2025-03-25
fix(users-permissions): updating user throws error

### LINKED ISSUE #23201: Users-Permission Admin error when updating a user
### Node Version

v20.18.2

### NPM/Yarn/PNPM Version

Yarn 1.22.22

### Strapi Version

5.11.3

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When updating a user in the admin panel (username, password, ect) and you save an error is shown in the admin console but the data is still updated. Looks like the admin is trying to refetch the data using the Document ID rather than the integer ID

### Steps to Reproduce

1. Create an app
2. Create a user for users-permissions
3. Try to update the user's username and save
4. See Error in Admin
5. Check API response and see that the data was properly updated

### Expected Behavior

Should not throw an error (more specifically it shouldn't be doing any kind of documentID valication on users-permissions since it doesn't use the document Service

### Logs

```shell

```

### Code Snippets

_No response_

### Media

Reproduction video

https://www.loom.com/share/df1e50020c394fffa98d17a4dd0474bf?sid=20d6b168-d721-4eab-8f53-7eb4bee58f5f

### Additional information

_No response_

### Confirmation Checklist

ankit7201:
… user permissions (fixes #23201)

### What does it do?

Fixes the error coming while saving user details

### Why is it needed?

See #23201

### How to test it?

Update username and notice the error doesn't appear anymore 

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/23201
Fixes https://github.com/strapi/strapi/issues/22205


markkaylor (approved):
Great looks good to me thanks for the PR

derrickmehaffy:
Wouldn't this be a breaking change since it shifts the response format to the normal one? (I'd love that but it's still a breaking change no?)

markkaylor:
> Wouldn't this be a breaking change since it shifts the response format to the normal one? (I'd love that but it's still a breaking change no?)

Hey @derrickmehaffy this shouldn't be a breaking change since the content manager is altering the response after it receives it from the API. The actual response shape on the server is left untouched. We already did this for the create action and that's why there is the comment to update this in v6 to alter the shape on the server.

PR #23227, opened 2025-03-25
Chore/api requests tracker

maccomaccomaccomacco:
### What does it do?

### Why is it needed?

### How to test it?

### DIFF
--- a/packages/core/core/src/services/metrics/middleware.ts
+++ b/packages/core/core/src/services/metrics/middleware.ts
@@ -19,7 +19,7 @@ const createMiddleware = ({ sendEvent }: { sendEvent: Sender }) => {
   const middleware: Core.MiddlewareHandler = async (ctx, next) => {
     const { url, method } = ctx.request;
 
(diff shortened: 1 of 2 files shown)


Mcastres:
There might be something wrong with the PR @maccomaccomaccomacco 🤔

maccomaccomaccomacco:
> There might be something wrong with the PR @maccomaccomaccomacco 🤔 

Why? @Mcastres

Mcastres:
> > There might be something wrong with the PR @maccomaccomaccomacco 🤔
> 
> Why? @Mcastres

You added several commits yet I only see on line update in one file in this PR 🤔

Mcastres:
Nevermind you reverted everything.
How could I test this?

maccomaccomaccomacco:
you're right; most probably it happened because i created the branch not from develop

I will close this PR and do a new one

maccomaccomaccomacco:
@innerdvations @jhoward1994 just let me know if I need to repoen the PR on a cleaner branch, I probably made a bit of a mess here

PR #23231, opened 2025-03-25
Fix report generation and upload

gewenyu99:
### What does it do?

There are some problems with JUnit report generation. 
- Look like all the files are called junit, which collides
- Looks like the `test-apps/e2e/test-results` was getting wiped between runs so only the last run ever gets saved there
- Updates the upload job to Trunk, which should point now to `test-apps/junit-reports/*.xml`

This instead outputs different test domains as separate files in a path that isn't wiped between runs
Screenshot 2025-03-25 at 1 45 57 PM

Existing HTML reports aren't changed.

### Why is it needed?

Currently, uploads to Trunk is incorrect, this affects flaky tests detection.

Further more, this will allow accurate display if you choose to enable PR comments: https://docs.trunk.io/flaky-tests/github-pull-request-comments

### How to test it?

Run tests and observe the outputted XML files. Also view upload succeeds in CI and uploads are found in Trunk.

### Related issue(s)/PR(s)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -244,7 +244,7 @@ jobs:
         if: '!cancelled()'
         uses: trunk-io/analytics-uploader@main
         with:
-          junit-paths: test-apps/e2e/test-results/**/junit.xml
+          junit-paths: test-apps/junit-reports/*.xml
           org-slug: strapi
           token: ${{ secrets.TRUNK_API_TOKEN }}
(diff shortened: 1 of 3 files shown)


Marc-Roig (approved):
thank you ❤️

Marc-Roig (approved):
thank you ❤️

Marc-Roig (approved):
thank you ❤️

Issue #21378, opened 2024-09-19
After updating a relation and publishing, content is immediately treated as modified


innerdvations:
### Required System information

$ strapi report
Launched In: 240 ms
Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0
Node/Yarn Version: yarn/1.22.22 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite

### Describe the bug

After updating  a relation (one to many in this case) and then clicking "publish", the "save" and "publish" buttons immediately become available again, and if I try to navigate away from the page it opens the "are you sure" confirmation modal

### Steps to reproduce the behavior

1. Have an "article" and "category" type, with article having one category
2. (in my case I also have localized content and a rich text field which may or may not be related)
3. Create an article
4. Publish it
5. Go back and edit the article, changing the category
6. Click publish
7. Publish and save immediately become available again because the content is detected as modified

The problem is resolved if you make another change, such as updating the rich text field and clicking publish. Then the page does not behave as modified. It seems to have something to do with the relation selection input.

### Expected behavior

The front end should not detect changes when there were none

### Code snippets

If applicable, add code samples to help explain your problem.

PR #23239, opened 2025-03-26
fix: disable publish button when publishing a relation change

Marc-Roig:
### What does it do?

Disables the 'publish' / 'save' button when publishing an entry that only updated a relation.

We were not reseting the form after publishing.

### Why is it needed?

After updating a relation (one to many in this case) and then clicking "publish", the "save" and "publish" buttons immediately became available again, and when trying to navigate away from the page it opened  the "are you sure" confirmation modal.

### How to test it?

1. Have an "article" and "category" type, with article having one category
2. Create an article
3. Publish it
4. Go back and edit the article, changing the category
5. Click publish
6. Publish and save should be disabled, as there is not any modified content.

### Related issue(s)/PR(s)
Fixes https://github.com/strapi/strapi/issues/21378


simotae14 (approved):
It works for me, I tested it because it fixes also this bug I was working on CS-1379

markkaylor (approved):

PR #23243, opened 2025-03-27
fix: relation modal nested form issue

remidej:
### What does it do?

Moves the FormProvider inside Modal.Body, because inside Modal.Root wasn't enough for it to be inside the react portal. This should fix the nested form error

Also fixes the ellipsis issue on the relation modal trigger. The actual fix is max-width: 100% on StyledTextButton, I just made some housekeeping updates around it too

### How to test it?

- check the errors in your console
- try having a relation with a super long name to make sure the ellipsis shows


markkaylor (approved):
Looks good to me. Tested locally.

PR #23245, opened 2025-03-27
fix: relation modal not closing when opening self in full page

remidej:
From entry A, open relation B in a modal, then on it click on relation A. Then click the open full page button. Before this fix the relation modal wouldn't close. Also make sure the open full page button still works in other cases


markkaylor (approved):
works.

PR #23247, opened 2025-03-27
fix: SSO login JWT formatting issues

### LINKED ISSUE #23246: Strapi v5.12.0 broke JWT-based SSO auth
### Node Version

v20.19.0

### NPM/Yarn/PNPM Version

Yarn 1.22.22

### Strapi Version

 v5.12.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

After upgrading to Strapi v5.12.0 strapi admin UI is failing to load with error when logging in with JWT-based SSO auth:

```
Uncaught (in promise) SyntaxError: Unexpected token 'e', "<redacted>"... is not valid JSON
    at JSON.parse (<anonymous>)
    at MOt (strapi-CxOHA5nJ.js:2309:11361)
    at jz (strapi-CxOHA5nJ.js:2309:11554)
    at mSe (strapi-CxOHA5nJ.js:2414:12064)
    at efn (strapi-CxOHA5nJ.js:6020:10823)
    at strapi-CxOHA5nJ.js:6020:11102
```

UI loads fine using Username + Password. After inspecting `jwtToken` cookie value I found that:
- When logging in with username+password jwtToken value is wrapped in quotes `%22<redacted>%22`
- When logging in with SSO jwtToken value is not wrapped in quotes `<redacted>`

This is likely related to the latest change in strapi v5.12.0:
- https://github.com/strapi/strapi/pull/23188

### Steps to Reproduce

Attempt to login to Strapi v5.12.0 using JWT-based SSO provider.

### Expected Behavior

Logging in with SSO provider doesn't fail with `is not valid JSON` error.

### Logs

```shell

```

### Code Snippets

Failing JS code:
```js
const jO = t => t instanceof wF
  , MOt = () => JSON.parse(localStorage.getItem(Iie.TOKEN) ?? SEe(Iie.TOKEN) ?? '""')
```

### Media

_No response_

### Additional information

Internal ticket with more details: 11796

### Confirmation Checklist

Bassel17:
### What does it do?

removes re-setting cookie of `jwtToken` in SSO,
cookies is already set from the server
 
### Why is it needed?

Login with SSO doesn't work because of session cookies enhancement

### How to test it?
Try to login with SSO it should work

### Related issue(s)/PR(s)
fixes #23246

### DIFF
--- a/packages/core/admin/admin/src/reducer.ts
+++ b/packages/core/admin/admin/src/reducer.ts
@@ -29,13 +29,13 @@ const THEME_LOCAL_STORAGE_KEY = 'STRAPI_THEME';
 const LANGUAGE_LOCAL_STORAGE_KEY = 'strapi-admin-language';
 
 export const getStoredToken = (): string | null => {
-  const token = localStorage.getItem(STORAGE_KEYS.TOKEN) ?? getCookieValue(STORAGE_KEYS.TOKEN);
-
-  if (typeof token === 'string') {
-    return JSON.parse(token);
+  const fromLocalStorage = localStorage.getItem(STORAGE_KEYS.TOKEN);
+  if (fromLocalStorage) {
+    return JSON.parse(fromLocalStorage);
   }
 
-  return null;
+  const fromCookie = getCookieValue(STORAGE_KEYS.TOKEN);
+  return fromCookie ?? null;
 };
 
 const adminSlice = createSlice({
@@ -75,7 +75,7 @@ const adminSlice = createSlice({
       const { token, persist } = action.payload;
 
       if (!persist) {
-        setCookie(STORAGE_KEYS.TOKEN, JSON.stringify(token));
+        setCookie(STORAGE_KEYS.TOKEN, token);
       } else {
         window.localStorage.setItem(STORAGE_KEYS.TOKEN, JSON.stringify(token));
       }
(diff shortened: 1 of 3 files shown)


packages/core/admin/admin/src/reducer.ts
```diff
@@ -29,7 +29,8 @@ const THEME_LOCAL_STORAGE_KEY = 'STRAPI_THEME';
 const LANGUAGE_LOCAL_STORAGE_KEY = 'strapi-admin-language';
 
 export const getStoredToken = (): string | null => {
-  const token = localStorage.getItem(STORAGE_KEYS.TOKEN) ?? getCookieValue(STORAGE_KEYS.TOKEN);
+  const token =
+    localStorage.getItem(STORAGE_KEYS.TOKEN) ?? JSON.stringify(getCookieValue(STORAGE_KEYS.TOKEN));
```
jhoward1994, line 33: I think this could result in the token being the string "null" if getCookieValue return null. Is that intended / could cause issues?

Bassel17: it shouldn't be null, and if it is then the users operation wouldn't work and they would be logged out which removes the cookie

innerdvations (commented):
LGTM but I have not QA'd it

jhoward1994 (approved):
Works for me! Tested with github oauth application SSO

innerdvations (approved):

PR #23250, opened 2025-03-27
chore: remove sdk-plugin dependencies from user deps

innerdvations:
### What does it do?

- removes @strapi/sdk-plugin devDependency from our plugin packages

### Why is it needed?

Those plugins no longer require sdk-plugin, they use rollup directly. So this should remove @strapi/sdk-plugin from all strapi project yarn.lock files and remove a bunch of (untrue) vulnerability warnings for our users.

The only remaining reference is in an example project plugin, so a `yarn audit` on the monorepo will still alert on it, but for that we will fix sdk-plugin itself.

### How to test it?

cloud plugin should still work fine
color-picker plugin should still work fine
@strapi/sdk-plugin (and @strapi/pack-up?) should no longer be used anywhere except the examples plugin project 

### Related issue(s)/PR(s)


markkaylor (dismissed):
Tested, lgtm

Convly (approved):
LGTM, some e2e tests are still failing though

PR #23251, opened 2025-03-27
fix: uid fields broken inside relation modals

remidej:
### What does it do?

Avoids calling useDoc directly in the UID field. Because it would not work when inside a relation modal. Use useDocumentContext instead.

### How to test it?

Normal UID field behavior:
- open kitchensink editview
- regenerate the slug field, it should not throw errors and should match the shortField value

Then do the same inside of a relation modal. You can find a relation to kitchensink via a tag entry.


simotae14 (approved):
The code looks good and it works, well done

PR #23253, opened 2025-03-27
feat(upload): add new setting to limit concurrent upload on low ram envs

Tyki:
### What does it do?

This pull request add a new settings to limit the concurrency of uploads to 1 in the media library.

It also fix a small issue when the file size is checked only after that images optimizations are applied. But it means that, whatever the image size is, sharp will try to optimize it. On low ram environments, it leads to OOM.

### How to test it?

Add the new setting in the configuration for the upload plugin: limitConcurrentUpload

```
  upload: {
    config: {
      limitConcurrentUploads: true
    }
  },
```

and then try to upload multiple assets in the same time

### Related issue(s)/PR(s)

https://github.com/strapi/documentation/pull/2445


markkaylor:
I believe this is closed by https://github.com/strapi/strapi/pull/26053, but thank you for contributing!

PR #23262, opened 2025-03-28
fix(upload): update snapshot

jhoward1994:
WIP for fixing FE tests in CI


jhoward1994:
The upload FE tests had a failing snapshot. Locally I couldn't get `npx nx run @strapi/admin:"test:front"` to fail as it is on the CI

PR #23264, opened 2025-03-28
fix: use starticon for each menu item

lucasboilly:
### What does it do?

- Use `startIcon` for each menu item
- Add a `fill` property (to remove when [#1811](https://github.com/strapi/design-system/issues/1811) will be fixed in the DS)

### Why is it needed?

There's currently 2 icons displayed for Image and Link.

CleanShot 2025-03-28 at 16 30 59@2x

### How to test it?

- Find an entry with a Rich text (Markdown) field
- Reduce the window size until the Image and Link are hidden in the menu
- Open the menu

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/1811

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/WysiwygNav.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/WysiwygNav.tsx
@@ -163,18 +163,24 @@ const WysiwygNav = ({
       menu: (
         <>
           <Menu.Separator />
-          <Menu.Item onSelect={() => handleActionClick('Bold', editorRef)} disabled={isDisabled}>
+          <Menu.Item
+            startIcon={<Bold fill="neutral500" />}
+            onSelect={() => handleActionClick('Bold', editorRef)}
+            disabled={isDisabled}
+          >
             <Flex tag="span" gap={2}>
-              <Bold aria-hidden fill="neutral600" />
               {formatMessage({
                 id: 'components.Blocks.modifiers.bold',
                 defaultMessage: 'Bold',
               })}
             </Flex>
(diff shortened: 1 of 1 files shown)


markkaylor (dismissed):
Nice catch and thanks for fixing it. Looks good to me.

lucasboilly:
Thanks @markkaylor! I just added a new commit to put the right color for the headings :)

markkaylor (approved):

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/ContentTypeBuilderNav.tsx
```diff
@@ -136,19 +136,26 @@ export const ContentTypeBuilderNav = () => {
...
+              <Menu.Item
+                disabled={!history.canUndo || !isInDevelopmentMode}
+                onSelect={undoHandler}
+                startIcon={<ArrowCounterClockwise />}
+              >
+                {formatMessage({
+                  id: 'global.last-change.undo',
+                  defaultMessage: 'Undo last change',
+                })}
+                Undo last change
```
simotae14: wait, why do you need both these stuff, I would assume the formatMessage part is enough

simotae14 (changes requested):
I found just one problem

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/ContentTypeBuilderNav.tsx
Adzouz: Oh wow, it's just a mistake sorry 😅 I fixed it, good catch!

simotae14 (approved):
perfect 🏅

Issue #22381, opened 2024-12-10
Roles assigned to review workflow stages went missing after save


a6i-palo:
### Required System information

- Node.js version: 20.17
- NPM version: 10.8.3
- Strapi version: 5.5.0
- Database: sqlite
- Operating system: osx
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When saving the roles assigned to review workflow stages, the assigned roles immediately "disappear" from the admin interface.

### Steps to reproduce the behavior

1. Go to a workflow in review workflow
2. Assign a role to the stages in the workflow
3. Click save
4. There's no error but the assigned roles is gone.

### Expected behavior

The assigned role should remain visible in the interface.

### Screen capture

https://github.com/user-attachments/assets/fab42149-2c0b-4a01-ab08-fad4d67044ce

StrapiBartosh:
I am getting the same behavior running LaunchPad with an EE license on v5.5.1

fatihtelis:
I'm encountering the same error in Strapi v5.10.2 Enterprise version. After saving the review workflow, roles disappear from the screen and do not persist after reloading the page.

PR #23273, opened 2025-03-31
fix: review workflow show roles

simotae14:
### What does it do?

It fixes a problem in the findAll query where the roles are not shown because of an error on the populate param
For example in the update CALL we pass an empty object in the populate and then we use the default constant value in the be to populate the data, https://github.com/strapi/strapi/blob/3df0fc0021c81650cb2d0ba4b433f378327a75b8/packages/core/review-workflows/server/src/constants/workflows.ts#L26 
Instead in the case of the GET we used to pass an array with two strings in the populate
`populate: ['stages', 'stageRequiredToPublish'],`
but with this value the permissions are not populated so I decided to remove the populate in the RTK query definition to avoid the issue

### Why is it needed?

Because otherwise in the Review Workflow page in the Settings the roles are not shown

### How to test it?

- Go to the Review Workflow page in the Settings and try to create or update a review workflow
- you need to see the roles associated to every single stage even after creation or modification

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/22381


packages/core/review-workflows/admin/src/routes/settings/hooks/useReviewWorkflows.ts
```diff
@@ -31,7 +31,6 @@ const useReviewWorkflows = (params: UseReviewWorkflowsArgs = {}) => {
 
   const { data, isLoading, error } = useGetWorkflowsQuery(
     {
-      populate: ['stages', 'stageRequiredToPublish'],
```
Marc-Roig, line 34: not tested yet, but why were we populating these fields in the first place? arent they necessary?

simotae14: Tbh I don't know the original reason, also because we are not doing it for the create and update methods, btw with this array we don't have the correct returning values because what we need to populate the stages with permissions is something like this
`{
  "stages": {
    "populate": {
      "permissions": {
        "fields": [
          "action",
          "actionParameters"
        ],
        "populate": {
          "role": {
            "fields": [
              "id",
              "name"
            ]
          }
        }
      }
    }
  },
  "stageRequiredToPublish": true
}`

remidej (approved):
Tested, all good

Issue #17391, opened 2023-07-20
Relation filed search not working as before + wrong layout


radoslavzeman:
### Required System information

- Node.js version: v16.20.1
- NPM version: 7.24.2
- Strapi version: 4.11.7
- Database: postgres
- Operating system: macOS Ventura
- Is your project Javascript or Typescript: ts

### Describe the bug

When searching in relation field, it returns results matching only beginning of the "relation display field". Before, it matched all results.

E.g. we have childPages, using `path` as main relations display field. Typing `ubytovanie` should return `bratislava-pre-ukrajinu/ubytovanie` but it doesn't.

Choosing some relations manually works. Search matching the beginning works.

But the layout is somehow broken.

### Expected behavior

Search in relations field should work correctly. The layout should display correctly.
It would be nice if it also ignores accents (áéšťýö etc...)

### Screenshots

![image](https://github.com/strapi/strapi/assets/19418224/05a92b5e-ce83-4d09-99bf-093f7e9ba5b8)
![image](https://github.com/strapi/strapi/assets/19418224/a647168e-8c6a-41e9-b24a-099bb4f86bb7)

![image](https://github.com/strapi/strapi/assets/19418224/d1bf61bb-70c3-4d78-95e5-80ebc10506aa)

dernetzjaeger:
I can confirm the change in searching behaviour when searching for a relations.

The code what changes it, seems to be in the v4.11.4 release.
Before v4.11.4 the search worked like "contains" filter, after updating to v4.11.4 it works like a "starts with" filter.

Maybe it is caused by #16671

nicoXXXXXtitan:
Hello

I can confirm that we have the same problem. We are currently on 4.13.6 
We preferred to do all the latest updates for security reasons but it's true that it really complicates the work of our editors.

vadym-lohvynovych:
Same here. I'm the one who created the topic on Strapi Community Forum. We had to decrease our strapi version because this issue makes work for our content manager much more complicated.

Seyolas:
> Same here. We had to decrease our strapi version because this issue makes work for our content manager much more complicated.

we are currently on 4.12.5. Which version are you decreased ? Look's like it works correctly in your decreased version.

vadym-lohvynovych:
> > Same here. We had to decrease our strapi version because this issue makes work for our content manager much more complicated.
> 
[...]

PR #23276, opened 2025-03-31
fix: relation field search to find all strings, not only the first ch…

apfmelo:
### What does it do?

Added a filter in autocomplete prop inside Combobox component equals to `contains`

### Why is it needed?

Fixed the relation field search to find all strings, not only the first characters

### How to test it?

Using the `examples/getstarted`:
Add an entry to the Categories collection and try to search for it in the relation field of the Address collection. 
Without the fix, it's only possible to find if you search for the string's beginning. 

| Before                                               | After                                                |
| ---------------------------------------------------- | ---------------------------------------------------- |
| Captura de Tela 2025-03-25 às 17 01 14 | Captura de Tela 2025-03-25 às 16 19 53 |

### Related issue(s)/PR(s)

This old issue was solved on Strapi v4 but not on Strapi v5
https://github.com/strapi/strapi/issues/17391#issuecomment-2462439423

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
@@ -615,7 +615,7 @@ const RelationsInput = ({
       <Combobox
         ref={fieldRef}
         name={name}
-        autocomplete="list"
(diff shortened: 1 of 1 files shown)


mathildeleg:
Hi, I believe this has since been fixed so I'll close the PR, but thanks for your contribution :)

Issue #23286, opened 2025-04-01
[ROTF] When publishing a relation during new entry creation, it redirects to this relation Edit view


MarionLemaire:
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

yarn

### Strapi Version

v0.0.0-experimental.3df0fc0021c81650cb2d0ba4b433f378327a75b8

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

When you create a new entry including a relation field, and you open the relation entry through the relation modal, update it and publish it, you get an error message and then when you close, you get redirected to this relation entry Edit view instead of initial entry Edit view.

### Steps to Reproduce

1. Create an entry with a relation field
2. Fill in the relation field
3. Click on the relation and see the relation modal is displayed
4. Update the relation and publish it
5. Check the error message but data is saved
6. Close relation modal
7. Check you're redirected to the relation Edit view instead of initial entry Edit view

### Expected Behavior

You should be redirected to initial entry Edit view

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/2ac7fa52-71b5-4b9d-9bb2-ef4d0a2a8521

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #23287, opened 2025-04-02
fix: OTF relations wrong redirect when publishing a nested relation during new entry creation

simotae14:
### What does it do?

It handles the case you publish, in the Relation on the fly modal, a relation starting from the creation entry form, because if you try to publish an entry from the creation view you have a redirect to the edit view of the just created document, but in the case of a relation modal, we need to skip this redirection

### Why is it needed?

We had a wrong redirection when we tried to Publish a nested relation inside a creation entry form

### How to test it?

- Go to a Collection and click "Create an entry"
- in the form you need to have a relation field, try to select a value for this relation field and click on it to open the Relation modal
- in the relation modal change a value and Publish
- you need to see a successful message and no redirections but instead remain in the modal

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/23286

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -650,7 +650,6 @@ const PublishAction: DocumentActionComponent = ({
       const { errors } = await validate(true, {
         status: 'published',
       });
-
       if (errors) {
         toggleNotification({
           type: 'danger',
@@ -660,10 +659,8 @@ const PublishAction: DocumentActionComponent = ({
               'There are validation errors in your document. Please fix them before saving.',
           }),
         });
-
         return;
       }
-
       const res = await publish(
         {
           collectionType,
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -673,12 +670,11 @@ const PublishAction: DocumentActionComponent = ({
         },
         transformData(formValues)
       );
-
       if ('data' in res && collectionType !== SINGLE_TYPES) {
         /**
          * TODO: refactor the router so we can just do `../${res.data.documentId}` instead of this.
          */
-        if (id === 'create') {
+        if (id === 'create' && !currentDocumentMeta.documentId) {
```
remidej, line 677: should we maybe retrieve the id from useDocumentContext instead of from useParams? The less we need complex conditions the better

remidej (approved):
Works as advertised. Thanks for the quick fix

PR #23288, opened 2025-04-02
CTB 💅

alexandrebodin:
### What does it do?

Revamp the CTB 💅

#### Details

- Adds undo/redo/discard
- Adds Drag & Drop of attributes
- Adds collaspe of components & dynamic zones
- Adds Edition status management (new,modified,removed)
- Adds concurrent editing of types
- Changes the SubNav in the entire App (CM/Settings/CTB)
- Adds Shortcuts in the CTB
	- Add ctrl+enter to save
	- Add ctrl+z / ctrl+y to undo/redo
- Ensures saved changes before leaving
- Improve Empty States in CTB & CM
- Disable Action when field or types are deleted
- Improve reload watcher to avoid false positives
- Fix modal closing on ESC / Enter
- Improvement to responsive
- Adds links to realted component & content types to navigate faster
- Adds required field info directly from the list view

TODO
- [x] Disable more actions when thigns are deleted
- [ ] i18N localization for undo/redo
- [ ] Make undo/redo configurable ? (hisotry is fixed at 50 right now)

#### Testings

Latest experimental (15/4/2025 12:35pm GMT+1): 0.0.0-experimental.7d94d6f1299d54c2aa90675f15b27b5c463aacc7

Here I'm listing extra things that are less obvious but need to be tested
- Test Fix done in https://github.com/strapi/strapi/pull/21310 as I changed the code
- Check Settings + CM side navs
- Check CM & CTB Empty states 
- Test status consistency (new/modified/delete)
- Test undo/redo behavior is not missing some edge cases
- Test relations creation & *update* are valid
  - Changing names or target, removing on one side and recreating ...
- Test UI in multiple browsers (drag & drop / collapse / responsive)


tests/e2e/utils/content-types.ts
```diff
@@ -3,6 +3,7 @@ import { waitForRestart } from './restart';
 import pluralize from 'pluralize';
 import { expect, type Page } from '@playwright/test';
 import { clickAndWait, ensureCheckbox, findByRowColumn, navToHeader } from './shared';
+import { rowHeight } from '@strapi/admin/admin/src/pages/Settings/pages/Roles/utils/constants';
```
innerdvations, line 6:
```suggestion
```

innerdvations (commented):
Awesome! First round of review and testing, nothing major. Really excited for this!

innerdvations:
Adding the dnd package seems to have broken the admin bundle size check 🤔

innerdvations:
Note on the editview conflict: I just split the file into 3 files and cleaned it up in general yesterday because it was so flaky. You should be able to just accept all the changes coming from develop and then reapply the one or two fixes you had to make there (and in the two new files)

alexandrebodin:
@innerdvations Thanks for the tests. The way the admin bundle check is down always breaks when we add new packages sadly. 

I'll try to update the branch today !

markkaylor:
This is awesome 👏 nice work. I played around with it a bit and found some things to nitpick about:

- I can still drag a deleted field even though its drag handle displays a disabled state


https://github.com/user-attachments/assets/3bd3e513-80f1-4cc9-8997-8ef956e132e6

- Deleting a root level field on a content-type adds the M (modified) indicator next to the content-type, deleting a field inside a component or DZ of a content-type does not add the M indicator to the content-type, but it does add it to the component. I guess that makes sense but it is a bit confusing. 


https://github.com/user-attachments/assets/f79acfc1-dde7-4793-b802-9d6527d6e2e0

- Modifying an attribute adds the M indicator. Adding an attribute also adds the M indicator. Deleting an attribute also adds the M indicator. Shouldn’t we see multiple indicators like M D N with the the respective Modified, Deleted, and New colors?


https://github.com/user-attachments/assets/911ef769-0d8a-4279-9530-50f2627b492d

- There are shortcuts to undo and redo but the UI does not help us learn what they are. Shouldn't there be an icon or a tooltip. See github for example.

Screenshot 2025-04-14 at 17 40 18
Screenshot 2025-04-14 at 17 41 31

- If I make a change and then manually revert that change. A diff is still recognized and the save button is still enabled. If I make a change and then use the undo feature there is no diff. Should the XP be the same for a manual undo and using the undo feature?


https://github.com/user-attachments/assets/3e78609a-679d-49a1-b2b6-36408649e383

MarionLemaire:
Second pair-testing session just done with @meganelacheny on latest experimental @alexandrebodin, we updated the Notion doc issues table accordingly.

Convly (approved):

Issue #21534, opened 2024-09-30
useFetchClient only handles json responses


cdreier:
### Describe the bug

we are migrating several local admin plugins to strapi5 - which is pretty awesome btw, so many great new apis to work with :heart: 

we want to download a zip file, which is created on the server. obviously zip files cannot be parsed as json - therfore the new `useFetchClient` with its responseInterceptor takes over, and just returns an empty list without any error

https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/utils/getFetchClient.ts#L123

### Steps to reproduce the behavior

1. use the new useFetchClient
2. return anything than json from a plugin api
3. wonder why you get any empty file

### Expected behavior

depending on the content-type, return `text()` / `blob()` or `json()`

cdreier:
@Boegie19 from my pov this is more a regression as this worked pretty well in v4 for example 

```jsx
const { get } = useFetchClient();
//[...]
get(url, {
  responseType: 'blob',
})
```

anyway, i would take care of it to implement a similar api in the new fetchClient?

alexandrebodin:
Hi @cdreier if you are interesting in contributing I think this wouldn't hurt to allow the fetchClient to support other responseTypes. We never really considered someone would be using it for this but why not :)

cdreier:
Hi @alexandrebodin i fiddled around, and tbh i am not able to create appropriate typings  - we solved the main problem in another way, using the fetch client with the jwt. wdyt, is it worth to give it another try? from my pov we also can close this issue

alexandrebodin:
Let's close it for now and we will see if we get more request like that before investing too much time :) 
Thanks for looking into it ! I'll flag it as feature request so it gets the right treatment 👍

ericbf:
This is still required, and shouldn’t be closed. The workaround shouldn’t be to just not use `useFetchClient`, which handles other things like authorization.

NawfelBengherbiaSfeir:
We stumbled on the same issue when updating our plugin to work on Strapi 5.

The current limited API does not let us request non JSON data from the server. This is quite frustrating as we need to go through workarounds to do basic things.

Excuse me, but the current design is half-arsed and needs fixing.

yunusyuksel:
That issue still exists

yunusyuksel:
[...]

PR #23291, opened 2025-04-02
fix: extend GET getFetchClient to get a raw response

dhoko:
Hello 👋

Use case: You want do generate a file server-side and download it as a user.

### What does it do?

When we use `useFetchClient` the `get` will auto convert everything to JSON. The idea is to introduce an option so we cat get the raw `Response` object and manipulate it ex to download a blob.

### Why is it needed?

You want to download a file from the API using an authenticated user -> ex you generate a file inside a plugin.

### How to test it?

Hard to test, I tested it with my own app:

Screenshot 2025-04-02 at 14 44 30

### Related issue(s)/PR(s)

cf https://github.com/strapi/strapi/issues/21534

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -29,6 +29,7 @@ type FetchResponse<TData = any> = {
 };
 
 type FetchOptions = {
+  raw?: boolean;
   params?: any;
   signal?: AbortSignal;
   headers?: Record<string, string>;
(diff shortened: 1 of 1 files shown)


Adzouz:
Hi @dhoko, thanks a lot for your contribution 🙏
Unfortunately, the solution provided wasn't exactly what we were looking for. We just merged another PR that fix the original issue https://github.com/strapi/strapi/pull/25974 so I'll have to close this PR.
Have a great day ☀️

PR #23310, opened 2025-04-03
fix: update the values of new relations added and not yet saved when we modify their name or status in the modal

simotae14:
### What does it do?

It solves an issue we had when we added new relations in an entry or in a new entry without saving the new relation and then we opened the modal to change the name of the relation, we saved or published and then we closed the modal, the changes were not reflected in the initial entry

### Why is it needed?

Because we need to show always to the user the updated value of the relations, to prevent these kind of errors

https://github.com/user-attachments/assets/c7ac23f2-fa8e-4d9b-b38a-98c7bcc3f999

https://github.com/user-attachments/assets/94b39bd9-bd06-42d8-971b-79b6f1bf4a11

### How to test it?

To test it you can create a new entry or edit an old one
- in the Edit/Create view add a new relation and click on it (without saving)
- then the relation modal try to change the name of the relations and Save (or Publish) to change also its status
- close the relation modal and check if the changes are reflected in the relations field

### Related issue(s)/PR(s)

CS-1390 and CS-1391

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
@@ -34,6 +34,7 @@ import { RelationDragPreviewProps } from '../../../../../components/DragPreviews
 import { COLLECTION_TYPES } from '../../../../../constants/collections';
 import { ItemTypes } from '../../../../../constants/dragAndDrop';
 import { useDebounce } from '../../../../../hooks/useDebounce';
+import { useDocument } from '../../../../../hooks/useDocument';
 import { type DocumentMeta, useDocumentContext } from '../../../../../hooks/useDocumentContext';
 import { type EditFieldLayout } from '../../../../../hooks/useDocumentLayout';
 import {
@@ -46,6 +47,7 @@ import {
   useLazySearchRelationsQuery,
   RelationResult,
 } from '../../../../../services/relations';
+import { type MainField } from '../../../../../utils/attributes';
 import { getRelationLabel } from '../../../../../utils/relations';
 import { getTranslation } from '../../../../../utils/translations';
 import { DocumentStatus } from '../../DocumentStatus';
@@ -118,6 +120,8 @@ interface Relation extends Pick<RelationResult, 'documentId' | 'id' | 'locale' |
     id: RelationResult['id'];
     locale?: RelationResult['locale'];
     position: RelationPosition;
+    // Added this property to prevent call useDocument with a not temporary relation (one already saved in the database)
+    temporary?: boolean;
   };
 }
 
@@ -305,6 +309,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
           id: relation.id,
           documentId: relation.documentId,
           locale: relation.locale,
+          temporary: true,
         },
         status: relation.status,
         /**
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -118,6 +120,8 @@ interface Relation extends Pick<RelationResult, 'documentId' | 'id' | 'locale' |
     id: RelationResult['id'];
     locale?: RelationResult['locale'];
     position: RelationPosition;
+    // Added this property to prevent call useDocument with a not temporary relation (one already saved in the database)
+    temporary?: boolean;
```
simotae14, line 124: probably the apiData was enough to discriminate between a new relation connected (and not save) and one already in memory but I wanted to be sure so I added this boolean prop

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -971,11 +981,34 @@ const ListItem = ({ data, index, style }: ListItemProps) => {
...
+    {
+      collectionType,
+      model: targetModel,
+      documentId: documentId ?? apiData?.documentId,
+      params: currentDocumentMeta.params,
+    },
+    { skip: !apiData?.temporary }
+  );
+  const label = document != null ? getRelationLabel(document, mainField) : originalLabel;
+  const status = document !== null ? document?.status : originalStatus;
```
remidej, line 1011: It would make more sense to have that ternary depend on the temporary thing imo, so it matches the skip param above

remidej (commented):
this looks really nice 👍

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
remidej, line 1: could you add comments explaining _why_ we're doing this? we have the context fresh in mind so it makes sense to us, but someone discovering this code would have a hard time figuring it out

remidej (approved):
🚀🚀🚀

markkaylor (commented):
Looks good to me, tested it locally and looks like it's working. Just a couple comments about the e2e tests updated here.

tests/e2e/tests/content-manager/create-content.spec.ts
```diff
@@ -224,13 +224,18 @@ test.describe('Adding content', () => {
...
+          .filter({ hasText: 'The value does not match the regex' })
+          .first(),
     },
   ] satisfies { description: string; fields: FieldValue[]; expectedError: string }[];
 
   for (const { description, fields, expectedError } of testCases) {
     test(`when I publish ${description} I see an error`, async ({ page }) => {
       await createContent(page, 'Match', fields, { save: false, publish: true, verify: false });
-      expect(page.getByText(expectedError)).toBeVisible();
+      expect(page.getByText(expectedError).first()).toBeVisible({ timeout: 20000 });
```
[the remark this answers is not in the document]
simotae14: I can remove the timeout, the problem with the flaky test was that we have multiple text with the expected error content

tests/e2e/tests/content-manager/create-content.spec.ts
```diff
@@ -224,13 +224,18 @@ test.describe('Adding content', () => {
         { name: 'opponent*', type: 'text', value: 'West Ham' },
       ],
       expectedError: 'The value does not match the regex',
+      locator: (page) =>
+        page
+          .locator('[data-strapi-field-error="true"]')
+          .filter({ hasText: 'The value does not match the regex' })
+          .first(),
```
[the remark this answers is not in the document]
simotae14: nope, it was an attempt to fix the flaky test

markkaylor (approved):

PR #23311, opened 2025-04-03
fix: blocks ui

lucasboilly:
### What does it do?

- Harmonize spacing between modifiers groups
- Change icons colors
- Use the `startIcon` property
- Add a shadow the the collapse `IconButton` in the expanded mode
- Remove some CSS properties to use the ones inherited from `Menu.Item`

### Why is it needed?

It's currently not 100% aligned with the designs.

### How to test it?

You can see the changes in an entry that has a Rich text (Blocks) field.

### Related issue(s)/PR(s)

None.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
@@ -41,6 +41,8 @@ const ToolbarSeparator = styled(Toolbar.Separator)`
   background: ${({ theme }) => theme.colors.neutral150};
   width: 1px;
   height: 2.4rem;
+  margin-left: 0.8rem;
+  margin-right: 0.8rem;
 `;
 
 const FlexButton = styled<FlexComponent<'button'>>(Flex)`
@@ -331,7 +333,7 @@ const BlockOption = ({ value, icon: Icon, label, blockSelected }: BlockOptionPro
 
   return (
     <SingleSelectOption
-      startIcon={<Icon fill={isSelected ? 'primary600' : 'neutral600'} />}
+      startIcon={<Icon fill={isSelected ? 'primary600' : 'neutral500'} />}
       value={value}
     >
       {formatMessage(label)}
@@ -443,11 +445,11 @@ const ListButton = ({ block, format, location = 'toolbar' }: ListButtonProps) =>
 
     return (
       <StyledMenuItem
+        startIcon={<Icon />}
         onSelect={() => toggleList(format)}
         isActive={isListActive()}
         disabled={isListDisabled()}
       >
-        <Icon />
         {formatMessage(block.label)}
       </StyledMenuItem>
     );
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -746,8 +739,8 @@ const BlocksToolbar = () => {
     {
       // List buttons can only be rendered together when in the toolbar
       toolbar: (
-        <Flex direction="row" gap={1}>
-          <ToolbarSeparator />
+        <Flex direction="row">
+          <ToolbarSeparator style={{ marginLeft: '0.4rem' }} />
```
markkaylor, line 743: I think you can define this in the styled component on line 44 instead of overriding it here.

lucasboilly: I feel like the way elements are nested is forcing me to override stuff. It's hard to explain with text, I'm gonna record a Loom :)

remidej: maybe set a paddingLeft on the Flex above instead?

lucasboilly: @remidej what I need to do is to lower the `margin-left` of `ToolbarSeparator` in this specific place because of the `gap` property of the `Flex` above. But if I remove the `gap` I'm gonna have to add margins manually everywhere else. 

Or maybe it's possible to have negative padding? Like `gap={-1}`?

remidej: that's not a thing, only margins can be negative.

I think generally you want to avoid setting margins internally and prefer relying on gap on parents instead. Maybe it doesn't work here though, not sure.

You may also be interested in CSS selectors like `:first-child` to adjust the CSS based on an element's siblings

lucasboilly: The layout looks like this:

CleanShot 2025-04-04 at 13 21 39@2x

My issue is a gap, but only the one that is before the divider.

remidej: There may be a more elegant solution somewhere but this is not a standard layout scenario, so we'd need to dig into it for a bit. I personally don't think it's worth holding up the PR. WDYT @markkaylor?

markkaylor: Yeah fine by me to move forward

remidej (dismissed):

remidej (approved):

PR #23315, opened 2025-04-04
fix: dragging fields over components breaks the view

### LINKED ISSUE #22056: Drag moves incorrect field when configuring a view

- Node.js version: 20.16.0
- NPM version: 10.8.2
- Strapi version: 5.2.0
- Database: 
- Operating system: MacOS Sonoma 14.5

### Describe the bug

When configuring the view of a content type, I am dragging a text field up the page but when it moves over a component, it then starts dragging the component rather than the text field.

1. Configure the view of a content type which includes a component.
2. Drag a text field up the page and over the component.
3. Once you move past the component, it then starts moving the component rather than the original text field.

### Expected behavior

I should be able to drag the text field past the component to move it further up the page. You can workaround the issue by dragging around the component (see the second video).

### Screenshots

https://github.com/user-attachments/assets/f8f758fe-7c19-4032-8b34-9146a51f06ae

https://github.com/user-attachments/assets/908b53d7-e28c-4324-8d1d-79853e421335

markkaylor:
### What does it do?

- Fix dragging field over components breaks the view
- Migrate configure the view drag and drop from react-dnd to dnd-kit

### Why is it needed?

To fix the issue
To start the migration from react-dnd to dnd-kit

### How to test it?

Go to the configure the view for any content-type with components and dynamic zones
Drag and drop things around

### Related issue(s)/PR(s)

fixes #22056

### TODO

- Handle drag preview on the y axis to see where the item will be dropped
- [Bug] moving a field in a container of multiple fields over a full width field sometimes move the entire container

### DIFF
--- a/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
+++ b/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
@@ -1,5 +1,9 @@
 import * as React from 'react';
 
+import { DndContext, DragOverlay, UniqueIdentifier } from '@dnd-kit/core';
+import { SyntheticListenerMap } from '@dnd-kit/core/dist/hooks/utilities';
+import { arrayMove, rectSortingStrategy, SortableContext, useSortable } from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
 import { useField, useForm } from '@strapi/admin/strapi-admin';
 import {
   Modal,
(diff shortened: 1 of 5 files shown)


packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -47,17 +104,24 @@ interface FieldsProps extends Pick<EditLayout, 'metadatas'>, Pick<FieldProps, 'c
...
     'Fields',
     (state) => state.values.layout ?? []
   );
+  const containers = layout.reduce(
+    (acc, row, index) => ({
+      ...acc,
+      [`container-${index}`]: row,
+    }),
+    {} as Record<string, ConfigurationFormData['layout'][number]>
+  );
```
markkaylor, line 119: This is probably unnecessary, in the beginning I just found it easier to work with an object. I'll look at removing and relying on the layout array directly.

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -370,15 +497,14 @@ const Field = ({ attribute, components, name, index, onMoveField, onRemoveField
         borderColor="neutral150"
         background="neutral100"
         hasRadius
-        style={{ opacity: isDragging ? 0.5 : 1 }}
```
markkaylor, line 373: This should be aded back

PR #23328, opened 2025-04-07
Experiment adding doc links in HeaderLayout

Mcastres:
### What does it do?

Include a documentation helper link on every pages of the CMS.

### Why is it needed?

We need to understand if the users feel already disoriented in the admin panel and on which sections. By introducing a tracked documentation link through an `IconButton` component next the the page HeaderLayout, we could get a the number of click = users looking for information about that particular page.

### How to test it?

- Go to any page of the CMS
- If there is an `IconButton` next to the page title, it should redirect you, when clicking on it, on the associated documentation page.
- Same should happen on the sticky version of the Header of the page.

### DIFF
--- a/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
+++ b/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
@@ -58,7 +58,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
     const { trackUsage } = useTracking();
 
     const docLinkButton = docLink ? (
-      <Flex paddingLeft={2}>
+      <Flex>
         <IconButton
           onClick={() =>
             trackUsage('didClickOnDocLink', { from: docLink.pathname, to: docLink.link })
@@ -93,7 +93,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
           zIndex={1}
           data-strapi-header-sticky
         >
-          <Flex justifyContent="space-between">
+          <Flex gap={2} justifyContent="space-between">
             <Flex>
               {navigationAction && <Box paddingRight={3}>{navigationAction}</Box>}
               <Box>
@@ -112,7 +112,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
             </Flex>
             <Flex>
               {primaryAction ? (
-                <Flex gap={2}>
+                <Flex>
                   {docLinkButton}
                   {primaryAction}
                 </Flex>
@@ -123,6 +123,8 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
       );
     }
 
+    console.log(primaryAction ? true : false);
+    
     return (
       <Box
         ref={ref}
@@ -134,7 +136,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
         data-strapi-header
       >
         {navigationAction ? <Box paddingBottom={2}>{navigationAction}</Box> : null}
-        <Flex justifyContent="space-between">
+        <Flex gap={2} justifyContent="space-between">
           <Flex minWidth={0}>
             <Typography tag="h1" variant="alpha" {...props}>
               {title}
@@ -142,7 +144,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
             {secondaryAction ? <Box paddingLeft={4}>{secondaryAction}</Box> : null}
           </Flex>
           {/* Experiment */}
-          <Flex gap={2}>
+          <Flex>
             {docLinkButton}
             {primaryAction}
           </Flex>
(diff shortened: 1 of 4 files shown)


Mcastres:
@lucasboilly sorry I edited your message instead of quote replying to it 😅 
I made the changes you requested

lucasboilly:
No worries @Mcastres :D

I just checked and that's good 👍

maccomaccomaccomacco (dismissed):
LGTM

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
```diff
@@ -1,29 +1,82 @@
...
+    },
     ref
   ) => {
     const isSubtitleString = typeof subtitle === 'string';
 
+    const { formatMessage } = useIntl();
+    const { trackUsage } = useTracking();
+
+    const docLinkButton = docLink ? (
+      <Flex paddingLeft={2}>
```
remidej, line 61: It's a nitpick but I don't think it makes sense for the the doc link button to have its own padding left. It has more to do with where it's being used and rendered, so I'd recommend putting the padding (or relying on gap) there

packages/core/admin/admin/src/components/Layouts/utils/getMatchingDocLink.ts
```diff
@@ -0,0 +1,163 @@
...
+export async function getMatchingDocLink(pathname: string) {
+  const result =
+    matchingLinks.find((item) => pathname === item.pathname) ||
+    matchingLinks.find((item) => pathname.includes(item.pathname));
+
+  if (!result) {
+    return null;
+  }
+  return { ...result };
+}
```
remidej, line 163: This doesn't need a promise, you can remove the async. Also I don't think you need to spread result when you return it

packages/core/admin/admin/src/components/Layouts/utils/getMatchingDocLink.ts
```diff
@@ -0,0 +1,163 @@
...
+    pathname: '/me',
+    link: 'https://docs.strapi.io/user-docs/getting-started/setting-up-admin-panel#setting-up-your-administrator-profile',
+    title: 'Administrator profile',
+  },
+];
+
+export async function getMatchingDocLink(pathname: string) {
+  const result =
+    matchingLinks.find((item) => pathname === item.pathname) ||
+    matchingLinks.find((item) => pathname.includes(item.pathname));
```
remidej, line 157: You can avoid looping on the array twice by checking for both conditions inside the same find instead

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
```diff
@@ -124,13 +195,24 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
...
   }, [baseHeaderLayoutRef]);
 
+  React.useEffect(() => {
+    const fetchDocLink = async () => {
+      const result = await getMatchingDocLink(pathname);
+      setDocLink(result);
+    };
+
+    fetchDocLink();
+  }, [pathname]);
```
remidej, line 205: Since getMatchingDocLink doesn't need to be async, you can get the docLink by deriving it from the pathname directly in the component, no need for an effect or any state:

```
const { pathname } = useLocation();
const docLink = getMatchingDocLink(pathname);
```

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
Mcastres: Where? 😅

packages/core/admin/admin/src/components/Layouts/utils/getMatchingDocLink.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!result) {
+    return null;
+  }
```
jhoward1994, line 161: Would a fallback default doc link make sense in this case?

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
Mcastres: Not sure to know the best practices if the button can't have its own padding. Where should I put he gap?

packages/core/admin/admin/src/components/Layouts/utils/getMatchingDocLink.ts
Mcastres: Sure!

remidej: Then as a user, it would mean that I no longer know if the button I'm clicking on is specific to the page I'm on or not. I'd personally prefer only showing the doc link when it points to the specific area I'm on, so I know it'll always be useful

Also if we can't find a doc link for a given page, then we likely don't have docs for that page, so I don't think the user would find what they're looking for, even manually

derrickmehaffy:
As always, we need an option in the admin config to disable this, many users and/or customers want to white label and this remove all references to official Strapi resources.

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
remidej: I think you want to have a gap higher up, [here](https://github.com/strapi/strapi/blob/dac464e299ef4e07dfb5522447e833c2e104c182/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx#L96) and [here](https://github.com/strapi/strapi/blob/dac464e299ef4e07dfb5522447e833c2e104c182/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx#L137)

They're the Flex parents that make their first child render on the left, and their second on the right. Putting a gap there will ensure those 2 children won't collide, which I believe is what your paddingLeft was meant for

innerdvations:
Superseded by https://github.com/strapi/strapi/pull/26422

Issue #20312, opened 2024-05-15
select * from `strapi_database_schema` order by `time` DESC limit 1 - Out of sort memory, consider increasing server sort buffer size


claudiograssia:
## Bug report
Hi, I found this error after the schema column in the strapi_database_schema table reached 262kb.

### Required System information
Strapi Version: You can reproduce this bug using Strapi versions from 4.5.0 to 4.24.2.
Database: Mysql 8
Node Versione: 16,18,20
NPM: 6.0
OS: Ubuntu 20.04
Project: Javascript

### Describe the bug
I encountered this error when I added many collections, resulting in the schema column in the strapi_database_schema table reaching a size of 262kb. When I attempted to start the server using the command:

```shell
yarn develop
```

I receive the error:

```sql
Error: select * from `strapi_database_schema` order by `time` DESC limit 1 - Out of sort memory, consider increasing server sort buffer size
```

I attempted to resolve the issue by adding a simple index key to the schema column, but it did not work in any case.

### Steps to reproduce the behavior

1. Create a Strapi instance with a MySQL database.
2. Create many collections until the schema column in the strapi_database_schema table reaches 262kb.
3. Attempt to start Strapi using the command yarn develop.
4. Observe the error in the console.

### Expected behavior

The error:
```sql
Error: select * from `strapi_database_schema` order by `time` DESC limit 1 - Out of sort memory, consider increasing server sort.
```
should not appear; otherwise, the Strapi instance will not start.

claudiograssia:
"I also tried this query directly in the database; if I try to insert 'id' instead of 'time' in the sorting, the error does not appear."

Lau08:
Is there any update of this issue? Or a way to partial fix?

abaksha-sc:
Faced the same issue on MySQL 8.0.35 and Strapi 4.11.5.

Hot fixed by increasing `sort_buffer_size` from default `262144` to `500000` (execute SQL `SET GLOBAL sort_buffer_size = 500000;`), but documentation says 
```
On Linux, there are thresholds of 256KB and 2MB where **larger values may significantly slow down memory allocation**, so you should consider staying below one of those values. Experiment to find the best value for your workload
```

So maybe better fix is necessary.

This happens when you have many collections and strapi stores all DB schema in a single field. In my case the field `schema` of table `strapi_database_schema` has size 257112 characters what is almost a limit for default database buffer size 262144. And when strapi reads it with additional fields then the limit exceeds.

[...]

PR #23331, opened 2025-04-07
fix: optimize schema validation lookup for mysql/mariadb

derrickmehaffy:
### What does it do?

Optimizes the schema lookup during the schema validation for MySQL/MariaDB to prevent "Out of sort memory" errors caused by the very large JSON structure we are storing in the database for the schema

### Why is it needed?

Fixes a problem during the schema sync with very large projects or projects with a complex schema

### How to test it?

Run it on a very complex project where you could easily reproduce #20312

### Related issue(s)/PR(s)

fixes: #20312

### DIFF
--- a/examples/experimental-dev/package.json
+++ b/examples/experimental-dev/package.json
@@ -29,6 +29,6 @@
   "devDependencies": {
     "@vitejs/plugin-react": "4.2.1",
     "babel-plugin-react-compiler": "0.0.0-experimental-c23de8d-20240515",
-    "vite": "5.4.15"
+    "vite": "5.4.17"
   }
 }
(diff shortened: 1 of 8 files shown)


derrickmehaffy:
building experimental: https://github.com/strapi/strapi/actions/runs/14316789403

derrickmehaffy:
exp version: `0.0.0-experimental.2d0a5401e9f3f5c797448e48890c9dd0c05ebad2`

Marc-Roig (approved):
a bit of a bummer that we have to do this in two queries, but do not see any other way around it. Thank you

alexandrebodin:
it's a known issue in MYSQL that is not fix when select json / blob & binary columns 👍

innerdvations (approved):

PR #23335, opened 2025-04-08
Poc: OTF relations create modal

simotae14:
### What does it do?

Add a raw solution to implement the create relation modal, and once the new relation is saved or published add it to the parent relation

### Why is it needed?

To start the implementation of the OTF Relation create modal

### How to test it?

Go to the content manager, select a collection that contains a relation and start creating new relation values


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -154,6 +157,17 @@ function reducer(state: State, action: Action): State {
...
+      const lastIndex = state.documentHistory.length - 1;
+      const lastDocument = state.documentHistory[lastIndex];
+      const isLastDocumentCreate =
+        !lastDocument.documentId && lastDocument.collectionType !== SINGLE_TYPES;
+      return {
+        ...state,
+        documentHistory: isLastDocumentCreate
+          ? state.documentHistory.slice(0, lastIndex)
+          : state.documentHistory,
+      };
```
remidej, line 170: I think you can rely on GO_BACK instead for this instead

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -215,6 +229,8 @@ const RelationModalRenderer = ({
 
   const currentDocumentMeta = state.documentHistory.at(-1) ?? rootDocumentMeta;
   const currentDocument = useDocument(currentDocumentMeta);
+  const isSingleType = currentDocumentMeta.collectionType === SINGLE_TYPES;
+  const isCreatingDocument = !currentDocumentMeta.documentId && !isSingleType;
```
remidej, line 233: I see you need to define this in 3 places. I would recommend adding it to the relation context instead. Not in the reducer, but in the provider, next to all the other data that is derived from the reducer state

packages/core/content-manager/admin/src/hooks/useDocumentContext.ts
```diff
@@ -27,6 +27,10 @@ interface DocumentMeta {
    * i.e. { locale: 'fr' }
    */
   params?: Record<string, string | string[] | null>;
+  /**
+   * Field to connect, is the field name to add the newly created relation
+   */
+  fieldToConnect?: string;
```
remidej, line 33: I haven't taken the time to explore this but this is the part I'm a bit uncertain of. There may be a chance to find something more flexible for us in the future

remidej (commented):
Very cool that you got something working already 👏

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
simotae14: yeah but the GO BACK change also the document, in my case I just need to clean up the history and replace the object with documentId undefined (basically the not saved new relation) with the object containing the documentID (and model and collection type) coming from the create/publish API response

packages/core/content-manager/admin/src/hooks/useDocumentContext.ts
simotae14: yes the part regarding adding the new relation to the parent relation field is the one where I have more doubts because right now when I save a new relation at the same time I update the parent relation using the field name I passed with fieldToConnect and the connect property, we can discuss about it during the real implementation of the feature in the next sprint if we want

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
remidej: What I mean is that conceptually, the intent is still to go back, and it's triggered on the back button. With reducers we to think about events first, that's why I think it would be nice for it to be the same event.

The logic in terms of how to react and adapt state from this event can be figured out in the GO_BACK reducer implementation. You have access to the previous state, so you can infer if the user was creating when he clicked back, so you can add an if clause based on that

simotae14:
I am gonna close the POC because we are releasing the feature

PR #23337, opened 2025-04-09
test(e2e): use afterAll instead of afterEach to reset files to reduce flakiness

innerdvations:
### What does it do?

Resets files with afterAll instead of afterEach

### Why is it needed?

Resetting files in the e2e test suites is very expensive (time-wise) and also a minor source of flakiness on its own. Even though the ideal would be to reset files after each test to avoid cross-contamination, for efficiency and reduced flakiness we will switch to `afterAll` and in cases where tests need to avoid interfering with each other a reset can be done in the specific tests that require it.

### How to test it?

They should all pass

### Related issue(s)/PR(s)

### DIFF
--- a/tests/e2e/tests/content-type-builder/collection-type/create-collection-type.spec.ts
+++ b/tests/e2e/tests/content-type-builder/collection-type/create-collection-type.spec.ts
@@ -19,8 +19,6 @@ test.describe('Create collection type with all field types', () => {
     await clickAndWait(page, page.getByRole('link', { name: 'Content-Type Builder' }));
   });
 
-  // TODO: each test should have a beforeAll that does this, maybe combine all the setup into one util to simplify it
-  // to keep other suites that don't modify files from needing to reset files, clean up after ourselves at the end
   test.afterAll(async () => {
     await resetFiles();
   });
--- a/tests/e2e/tests/content-type-builder/collection-type/edit-collection-type.spec.ts
+++ b/tests/e2e/tests/content-type-builder/collection-type/edit-collection-type.spec.ts
@@ -23,9 +23,7 @@ test.describe('Edit collection type', () => {
     await navToHeader(page, ['Content-Type Builder', ctName], ctName);
   });
 
-  // TODO: each test should have a beforeAll that does this, maybe combine all the setup into one util to simplify it
-  // to keep other suites that don't modify files from needing to reset files, clean up after ourselves at the end
-  test.afterEach(async () => {
+  test.afterAll(async () => {
     await resetFiles();
   });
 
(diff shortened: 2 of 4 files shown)


Convly (commented):
Code LGTM, but the overall idea seems a bit scary to me. This is likely going to be harder to chase down bugs due to DB pollution from other tests :thinking: 

Do you have a mini benchmark to test the performance gained from this change to better understand the compromise we are making?

Convly:
> > Code LGTM, but the overall idea seems a bit scary to me. This is likely going to be harder to chase down bugs due to DB pollution from other tests 🤔
> > Do you have a mini benchmark to test the performance gained from this change to better understand the compromise we are making?
> 
> Do you mean the `afterAll` change? That's already how it's done in almost all of the other tests, these ones just hadn't been updated to follow. And it only causes pollution between tests in the same file, not other files.
> 
> Benchmark is basically that without this PR, tests on github yesterday had to be re-run 5+ times to get them to pass while this one passed on the first try three times in a row. But this [fourth try](https://github.com/strapi/strapi/actions/runs/14374179442/job/40303040106?pr=23337) has now failed, so I'm going to investigate editview and see if there's something else that can be done
> 
> As for the initial file generation commits, that may actually be a minor slow down, but it will ultimately allow us to enforce/ensure a clean environment between test files without a significant performance loss.

Yeah I was talking about the afterAll, wasn't aware it was already like that in other tests. If the gain is having tests passing most of the time & removing flakiness I like it, but... We'll need to carefully monitor whether we end up in weird situations in the future

innerdvations:
Ok, I've split editview into three tests and updated it to use the more modern/stable utilities we've added in the time since it was originally written. Seems to be a lot more stable now, although there aren't any speed improvements unfortunately -- probably because any gains were negated by the additional server start I added. But now we have more room for future improvement.

innerdvations:
> Looks good!
> 
> I am wondering at some point once the e2e tests grow even bigger and slower, it might be worth investigating a solution other than the Github runners

Agreed. For future reference, I think some big potential improvements are:

- caching test-apps, like if we could only do the project file creation, yalc install, initial startup file generation, and git stuff only once, and then essentially copy/paste it
- reduce admin build time; it's only like 3-5s for me locally but the CI takes like 20+ seconds. But if we could just globally get it faster somehow, that would be great
- more efficient and reliable way to do file and data resets and server reboots; committing binary file data with data transfer and listening for health checks is pretty ugly, we just haven't had time to find a better way
- API test time could probably be cut in less than half; I did some benchmarks last year and generally 80-90% of the runtime in each test file is just starting up the mock strapi

PR #23342, opened 2025-04-10
future(admin): add widgets api

markkaylor:
### What does it do?

Adds the widget API as described in the RFC found in notion

### Why is it needed?

To register custom widgets 

### How to test it?

In `examples/getstarted/src/admin/app.jsx` register a widget

```
app.widgets.register({
      icon: Pencil,
      title: {
        id: 'content-manager.widget.last-edited.title',
        defaultMessage: 'Last edited entries',
      },
      component: async () => <div>test</div>,
      pluginId: 'plop',
      id: 'last-edited',
      permissions: [{ action: 'plugin::content-manager.explorer.read' }],
    });

    console.log(app.widgets.getAll());
```


packages/core/admin/admin/src/StrapiApp.tsx
```diff
@@ -133,6 +135,15 @@ class StrapiApp {
     this.createHook(MUTATE_EDIT_VIEW_LAYOUT);
 
     this.router = new Router(getInitialRoutes());
+
+    if (window.strapi.future.isEnabled('unstableWidgetsApi')) {
+      /**
+       * TODO:
+       * Instantiate this above with the other properties
+       * whe removing the future flag
```
markkaylor, line 143:
```suggestion
       * when removing the future flag
```

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable check-file/filename-naming-convention */
+
+import invariant from 'invariant';
+import { To } from 'react-router-dom';
+
+import { Permission } from '../../../../shared/contracts/shared';
+
+import type { MessageDescriptor } from 'react-intl';
+
+type WidgetUID = `plugin::${string}.${string}` | `global::${string}`;
```
Convly, line 10: Might be a bit more verbose, but if you want to be future proof & rely on existing utils, you could use the following
```suggestion
import type { Internal, Utils } from '@strapi/types';

type WidgetUID = Utils.String.Suffix<
  | Internal.Namespace.WithSeparator<Internal.Namespace.Plugin>
  | Internal.Namespace.WithSeparator<Internal.Namespace.Global>,
  string
>;
```

Or even add it to the `Internal.uid` namespace in https://github.com/strapi/strapi/blob/develop/packages/core/types/src/internal/uid.ts (if that make sense to have it defined for broader usage)

markkaylor: Nice, we could add it to that file. Custom fields uses the same pattern. Just not sure what the name could be so it could work for both or anything else in the future. wdyt @remidej ?

packages/core/admin/admin/src/StrapiApp.tsx
remidej: I think we could instantiate it already, as long as we don't query the widgets store in the frontend I don't think it's an issue to have "future" code not under future flag as long it doesn't impact real app usage

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      const uid: WidgetUID = pluginId ? `plugin::${pluginId}.${id}` : `global::${id}`;
+
+      this.widgets[uid] = widget;
+    }
+  };
+
+  getAll = () => {
+    return this.widgets;
+  };
```
remidej, line 53: I think it would be more convenient to have getAll return an array, so `Object.entries(this.widgets)`

If so then it would be helpful to do some transformations on the widgets object before storing it:
- having the uid as an object property would be neat, since otherwise we don't have it anymore via getAll in array form
- since id and pluginId are only used to generate the uid, I think we could not store them

markkaylor: Makes sense, sounds good for the transformations I'll update

markkaylor: > having the uid as an object property would be neat, since otherwise we don't have it anymore via getAll in array form

I'm doing this when storing

> since id and pluginId are only used to generate the uid, I think we could not store them

I thought it made more sense to do this transformation only in getAll, I thought [uid]: `{...widget, uid}` was kind of strange. But now I'm thinking it makes more sense to do all the transformations in one place. wdyt?

markkaylor: I decided to do this: But now I'm thinking it makes more sense to do all the transformations in one place

which I believe is what you were suggesting anyway 🙃

PR #23345, opened 2025-04-10
future(homepage): export widget helper components

markkaylor:
### What does it do?

- Exports the widget helper components
- Does NOT export the root since all widgets will be wrapped in this component
- Adds a NoPermissions helper component
- Refactors the current homepage to use the Root directly

### Why is it needed?

Prepare for incoming widgets

### How to test it?

No visible change. The homepage should look and function exactly the same. 

You should be able to import Widget from a plugin like:

```
import { Widget } from '@strapi/strapi/admin';
```

### DIFF
--- a/packages/core/admin/admin/src/components/tests/Widget.test.tsx
+++ b/packages/core/admin/admin/src/components/tests/Widget.test.tsx
@@ -1,14 +1,15 @@
 import { Cog } from '@strapi/icons';
 import { render, screen } from '@tests/utils';
 
-import { Widget } from '../../../../components/WidgetHelpers';
+import { WidgetRoot } from '../../pages/Home/HomePage';
+import { Widget } from '../WidgetHelpers';
 
 describe('Homepage Widget component', () => {
   it('should render the widget with info from props', () => {
     render(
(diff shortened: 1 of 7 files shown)


remidej (approved):
Tested ✅ good idea to rename the file, it matches nicely the PageHelpers

packages/core/admin/admin/src/components/tests/Widget.test.tsx
```diff
@@ -1,14 +1,15 @@
 import { Cog } from '@strapi/icons';
 import { render, screen } from '@tests/utils';
 
-import { Widget } from '../Widget';
+import { WidgetRoot } from '../../pages/Home/HomePage';
```
markkaylor, line 4: @remidej I forgot to update the tests. I also moved them to be in the components folder with the WidgetHelper. I don't like that I have to export the WidgetRoot from the homepage just for this test but currently don't have a better solution. Maybe it's nbd 🤷 

p.s.

sorry had to force push along the way

remidej: Probably not a big deal, thanks for the heads up

remidej (approved):

PR #23352, opened 2025-04-11
future(homepage): refactor internal widgets to use api

markkaylor:
### What does it do?

- Move admin/server homepage services from core/admin to core/content-manager
- Move the internal widget to content-manager and register them using the api
- Add the UnstableHomePageCe in the Admin and render all registered widgets

### Why is it needed?

- It makes the homepage really cool instead of just sorta cool

### How to test it?

Go to the homepage, everything should work the same as before. No visible change.


packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -11,14 +12,15 @@ import { useAuth } from '../../features/Auth';
...
 /* -------------------------------------------------------------------------------------------------
  * Root
  * -----------------------------------------------------------------------------------------------*/
 
 interface RootProps {
   title: MessageDescriptor;
-  icon?: typeof import('@strapi/icons').PuzzlePiece;
+  icon?:
+    | typeof import('@strapi/icons').PuzzlePiece
+    | React.ComponentType<{ fill: string; 'aria-hidden': boolean }>;
```
markkaylor, line 23: I think I need help with the TS this feels wrong

packages/core/content-manager/admin/src/index.ts
```diff
@@ -44,6 +44,38 @@ export default {
...
+    app.widgets.register([
+      {
+        icon: Pencil,
+        title: {
+          id: `${PLUGIN_ID}.widget.last-edited.title`,
+          defaultMessage: 'Last edited entries',
+        },
+        component: async () => {
+          const { LastEditedWidget } = await import('./components/Widgets');
+          return LastEditedWidget;
```
markkaylor, line 58: I think we can just expect the component to be returned (not {default}), I updated the type but lmk if it doesn't make sense  (see the registration below). I figure if it's a default export you just do Compo.default 🤷‍♂️ is that confusing?

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -0,0 +1,185 @@
...
+};
+
+const RecentDocumentsTable = ({ documents }: { documents: RecentDocument[] }) => {
+  const { formatMessage } = useIntl();
+  const { trackUsage } = useTracking();
+  const navigate = useNavigate();
+
+  const getEditViewLink = (document: RecentDocument): string => {
+    // TODO: import the constants for this once the code is moved to the CM package
+    const isSingleType = document.kind === 'singleType';
```
markkaylor, line 50: Are we talking about the COLLECTION_TYPES/SINGLE_TYPES constant? Because that returns kebab case. I'm cool with this as is but lmk.

```suggestion
    const isSingleType = document.kind === 'singleType';
```

remidej: oh yes i probably didn't realize it wasn't the right casing

remidej:
@markkaylor could you make this target your closest other widget branch, so there are only the real changes in the diff?

markkaylor:
> @markkaylor could you make this target your closest other widget branch, so there are only the real changes in the diff?

Updated the diff should be accurate now

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -55,14 +57,88 @@ export const WidgetRoot = ({ title, icon = PuzzlePiece, children }: RootProps) =
...
+          const LazyComponent = await widget.component();
+
+          return {
+            ...widget,
+            component: LazyComponent,
+          };
+        })
+      );
+
+      setLoadedWidgets(loaded);
```
markkaylor, line 90: hmm thinking now I should just lazy load the components I think this blocks the widgets from rendering things like their title until the component is ready

markkaylor: nvm it doesn't block anything

packages/core/content-manager/admin/src/index.ts
remidej: I think I like that as we've moved more towards named exports, at least internally. It creates an inconsistency with custom fields but probably no big deal

packages/core/admin/admin/src/pages/Home/HomePage.tsx
remidej: should we actually allow non-DS icons? wdyt @lucasboilly?

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -55,14 +57,88 @@ export const WidgetRoot = ({ title, icon = PuzzlePiece, children }: RootProps) =
...
+        subtitle={formatMessage({
+          id: 'HomePage.header.subtitle',
+          defaultMessage: 'Welcome to your administration panel',
+        })}
+      />
+      <Layouts.Content>
+        <Flex direction="column" alignItems="stretch" gap={8} paddingBottom={10}>
+          <GuidedTour />
+          <Grid.Root gap={5}>
+            {loadedWidgets.map((widget) => {
```
remidej, line 115: You're mapping on an array that's only populated after a Promise.all, so I don't see how this wouldn't block all rendering?

I assume it's not very noticeable because it's fast, but I think we should provide lazy loading still, because some widgets could have big bundle sizes, if they require dataviz libraries for example. Also I think our ability to render Widget.Loading while waiting is neat, we should make the most of it

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -0,0 +1,185 @@
...
+    <Status variant={statusVariant} size="XS">
+      <Typography tag="span" variant="omega" fontWeight="bold">
+        {formatMessage({
+          id: `content-manager.containers.List.${status}`,
+          defaultMessage: capitalise(status),
+        })}
+      </Typography>
+    </Status>
+  );
+};
```
remidej, line 41: Looks like I forgot to delete this code. Could you update to import the component from DocumentStatus.tsx instead? It has an XS variant now

packages/core/content-manager/admin/src/index.ts
```diff
@@ -44,6 +44,38 @@ export default {
...
+        },
+        pluginId: PLUGIN_ID,
+        id: 'last-edited-entries',
+        permissions: [{ action: 'plugin::content-manager.explorer.read' }],
+      },
+      {
+        icon: CheckCircle,
+        title: {
+          id: `${PLUGIN_ID}.widget.last-published.title`,
```
remidej, line 67: did you check if translation files need to be updated?

packages/core/admin/admin/src/pages/Home/HomePage.tsx
remidej: I think optional with PuzzlePiece as the default is good, it's a lesson we learnt with custom fields.

and yes typeof PuzzlePiece sounds great 👍

packages/core/admin/admin/src/pages/Home/HomePage.tsx
markkaylor: I had [the same thought](https://github.com/strapi/strapi/pull/23352#discussion_r2039836719), I tested it this morning thought with a slow network connection. We see the title and the loading state for the content. But probably its just more readable/understandable to handle the components separately. I'll update.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
lucasboilly: @remidej let's make them use the DS icons only, at least for now.

packages/core/content-manager/admin/src/index.ts
markkaylor: The translations were already coming from the content-manager package so I think we should be good.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -65,33 +64,28 @@ const UnstableHomePageCe = () => {
...
-      const loaded = await Promise.all(
-        widgets.map(async (widget) => {
-          const LazyComponent = await widget.component();
-
-          return {
-            ...widget,
-            component: LazyComponent,
-          };
-        })
+      const resolvedComponents = await Promise.all(widgets.map((widget) => widget.component()));
```
remidej, line 74: This is gonna load them all at once right? I don't think it's ideal, it will be costly if one widget is longer to load

I'd recommend to create a new component for each widget instance and to manage the loading individually inside. It should also be simpler, no reduce and store needed

markkaylor: 100000000%

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -54,6 +56,73 @@ export const WidgetRoot = ({ title, icon = PuzzlePiece, children }: RootProps) =
...
+ * -----------------------------------------------------------------------------------------------*/
+
+const WidgetComponent = ({ component }: { component: () => Promise<React.ComponentType> }) => {
+  const [loadedComponent, setLoadedComponent] = React.useState<React.ComponentType | null>(null);
+
+  React.useEffect(() => {
+    const loadComponent = async () => {
+      const resolvedComponent = await component();
+
+      setLoadedComponent(() => resolvedComponent);
```
markkaylor, line 70: I was a bit confused why I needed to provide a new function but without it the resolvedComponent function is invoked immediately throws warnings for violating the rules of hooks and then a bunch of errors

remidej: storing components is always a mess, I guess the alternative may have been to use the ReactNode type and to display it with `{Component}` instead of `<Component />`

what you have is fine though!

remidej (approved):
🚀

packages/core/admin/admin/src/pages/Home/HomePage.tsx
markkaylor: If you think this is acceptable I kinda prefer the <Component /> syntax even it is completely an aesthetic thing

PR #23357, opened 2025-04-11
chore: add new locales

devcshort:
### What does it do?

Adds additional locales

### Why is it needed?

It's a requirement/is needed for a current application we are building

### How to test it?

N/A

### Related issue(s)/PR(s)

N/A


hanpaine:
Hi @devcshort I'm afraid we are not adding further functionality to V4

Issue #23265, opened 2025-03-28
Texts with no spaces (such as links) are changing the width of the Rich text (Markdown) field in expanded mode


lucasboilly:
### Node Version

22.11.0

### NPM/Yarn/PNPM Version

Yarn 4.5.0

### Strapi Version

5.12.0

### Operating System

MacOS

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

Texts with no spaces (such as links) are changing the width of the Rich text (Markdown) field in expanded mode, pushing the preview out of the modal.

### Steps to Reproduce

1. Go to an entry with a Rich text (Markdown) field
2. Type a text with no spaces in it

### Expected Behavior

The text should continue one the next line if it reaches the edge of the field.

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/b3bb88c5-e716-4e9a-8db2-6bd76859b5c9

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #23378, opened 2025-04-16
fix: rich text editors focus mode modals

remidej:
### What does it do?

Changes the implementation of the blocks and markdown editor "focus mode" modals, in order to use the Modal component from the design system instead of a DIY React portal

### Why is it needed?

Without it the modal can't be rendered on top of the ROTF relation modal. It also removes some old imperative code
### How to test it?
Use the markdown and blocks editors in the edit view, try the focus mode, preview mode for markdown, etc.

Also try it in the context of a relation modal: Tag entry => open Kitchensink relation modal => see rich text fields

### Related issue(s)/PR(s)

Replaces #23296

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
@@ -114,13 +114,6 @@ const EditorDivider = styled(Divider)`
   background: ${({ theme }) => theme.colors.neutral200};
 `;
 
-const ExpandIconButton = styled(IconButton)`
-  position: absolute;
-  bottom: 1.2rem;
-  right: 1.2rem;
-  box-shadow: ${({ theme }) => theme.shadows.filterShadow};
-`;
-
 /**
  * Forces an update of the Slate editor when the value prop changes from outside of Slate.
  * The root cause is that Slate is not a controlled component: https://github.com/ianstormtaylor/slate/issues/4612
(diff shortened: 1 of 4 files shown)


remidej:
@lucasboilly can you check the modals UI? Especially the dimensions

lucasboilly:
@remidej the size of the modals is good, was there something else to check?

There's still a few problems in the UI as you can see in these pull requests and issues
- https://github.com/strapi/strapi/pull/23264
- https://github.com/strapi/strapi/pull/23311
- https://github.com/strapi/strapi/issues/23265

simotae14 (approved):
I really like the new implementation, I tested it with normal entries and entries inside the Relation modal and it works, well done @remidej 
maybe to prevent having again this issues you can add a e2e test to check the visibility of the expanded block markdown

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
```diff
@@ -177,11 +170,7 @@ const BlocksEditor = React.forwardRef<{ focus: () => void }, BlocksEditorProps>(
     );
     const [liveText, setLiveText] = React.useState('');
     const ariaDescriptionId = React.useId();
-    const [isExpandedMode, setIsExpandedMode] = React.useState(false);
-
-    const handleToggleExpand = () => {
-      setIsExpandedMode((prev) => !prev);
-    };
+    const [isExpandedMode, handleToggleExpand] = React.useReducer((prev) => !prev, false);
```
markkaylor, line 173: What's your reasoning for the reducer? I thought a reducer was about accumulating a value...not sure I see how that makes sense for a boolean.

remidej: This is just a little trick to get a toggle function directly instead of a setter function

remidej:
@simotae14 I added the e2e test

simotae14 (approved):
well done Remi, btw in my save and publish PR for the create relation I created a folder for ROTF e2e tests and two files for the edit relation and the create relation

PR #23382, opened 2025-04-16
feat: OTF relations create modal add save and publish

simotae14:
### What does it do?

Add the logic in the create relation modal to save and publish a new entry, add the new relation to the parent relation field, track the didCreateEntry with fromRelationModal and fromPreview and handle the back button history in the case you try to open or create a new modal from the create relation modal without saving.
Also the fullpage click is handled in this PR

### Why is it needed?

We need to implement the logic to save and publish a new relation in the create relation modal

### How to test it?

Just try to save and publish relations in the create relation modal starting from collection types, single types document, DZ, components ...

### Related issue(s)/PR(s)

CS-1338
Fixes CS-1407


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -734,7 +735,10 @@ const PublishAction: DocumentActionComponent = ({
           // Update, if needed, the parent relation with the newly published document
           // check if in history we have the parent relation otherwise use the rootDocument
           if (fieldToConnect && documentHistory && parentDocumentMetaToUpdate.documentId) {
-            const parentDataToUpdate = parentDocumentData.getInitialFormValues();
+            const parentDataToUpdate =
+              parentDocumentMetaToUpdate.collectionType === SINGLE_TYPES
+                ? getInitialFormValues()
+                : parentDocumentData.getInitialFormValues();
```
remidej: I wonder why `parentDocumentData.getInitialFormValues` doesn't work for single types? Asking because ideally I'd prefer to avoid needing useDoc since it's not flexible

simotae14: I receive an empty object with single types

simotae14: and then I realized that I can't have a single type document in the modal but only in the Edit view page so I can use useDoc

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -788,16 +892,46 @@ const UpdateAction: DocumentActionComponent = ({
...
+  const documentHistory = useRelationModal(
+    'UpdateAction',
+    (state) => state.state.documentHistory,
+    false
+  );
+  const rootDocumentMeta = useRelationModal('UpdateAction', (state) => state.rootDocumentMeta);
+  const fromRelationModal = relationContext != undefined;
 
   const { currentDocumentMeta } = useDocumentContext('UpdateAction');
+  const [updateDocumentMutation] = useUpdateDocumentMutation();
```
markkaylor, line 923: Why do we do this directly instead of getting`update` from `useDocumentActions`? That's where we get the publish mutation as well on line 540. The isLoading destructured as well and that includes loading state for publish || update. Not sure if that is beneficial or not but just fyi

simotae14: We did it in this way with Remi because otherwise we would have Two notification success messages for the create and the update

markkaylor: ok what about the loading part, should the loading state returned in PublishAction consider the loading state of this mutation? See line 818

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -352,8 +394,12 @@ const RelationModal = ({ children }: { children: React.ReactNode }) => {
         </Modal.Header>
         <Modal.Body>
           <FormContext
-            method="PUT"
-            initialValues={currentDocument.getInitialFormValues()}
+            method={isCreating ? 'POST' : 'PUT'}
+            initialValues={
+              isCreating
+                ? currentDocument.getInitialFormValues(isCreating)
+                : currentDocument.getInitialFormValues()
```
remidej, line 401: Why not always call getInitialFormValues with isCreating? I assume it can handle isCreating being false

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -522,7 +579,9 @@ const RelationModalForm = () => {
   const documentLayoutResponse = useDocumentLayout(currentDocumentMeta.model);
   const plugins = useStrapiApp('RelationModalForm', (state) => state.plugins);
 
-  const initialValues = currentDocument.getInitialFormValues(isCreating);
+  const initialValues = isCreating
+    ? currentDocument.getInitialFormValues(isCreating)
+    : currentDocument.getInitialFormValues();
```
remidej, line 584: same question

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
simotae14: for sure we have an issue with loading and we can open a GH issue for that, and refactor a little bit this code, before merging the future branch

remidej: >We did it in this way with Remi because otherwise we would have Two notification success messages for the create and the update

also otherwise it would send 2 tracking events for the same action

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
simotae14: I tried and I received an empty object

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
simotae14: same as before

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
markkaylor: I get that we have issues to resolve around loading in the relation modal. Just want to clarify my point that here you are introducing a loading state that is not handled. It is not an existing issue. For example the publish button in the relation modal is only going to show a loader for the publish action, even though an update to the parent document could still be in progress after the publish is finished. 

Unless I'm missing something, and I'll leave it up to you if you want to handle it in another PR, but I would recommend going ahead and doing this in the relevant places:

```
const [updateDocumentMutation, { isLoading: isLoadingUpdate }] = useUpdateDocumentMutation();
// ....
return {
  loading: isLoading || isLoadingUpdate
}
```

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -683,11 +725,108 @@ const PublishAction: DocumentActionComponent = ({
...
+                        id: res.data.documentId,
+                        documentId: res.data.documentId,
+                        locale: res.data.locale,
+                      },
+                    ],
+                    disconnect: [],
+                  },
+                  __component: fieldToConnectUID,
+                };
+            // In case the object was not present you need to pass the componentUID of the parent document
```
markkaylor, line 791: I think this comment might be in the wrong place now...Maybe it's supposed to be above line 789?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -683,11 +725,108 @@ const PublishAction: DocumentActionComponent = ({
...
+                collectionType: metaDocumentToUpdate.collectionType,
+                model: metaDocumentToUpdate.model,
+                documentId:
+                  metaDocumentToUpdate.collectionType !== SINGLE_TYPES
+                    ? metaDocumentToUpdate.documentId
+                    : undefined,
+                params: metaDocumentToUpdate.params,
+                data: {
+                  ...dataToUpdate,
+                },
```
markkaylor, line 806: I don't think the spread is necessary:

```suggestion
                data: dataToUpdate,
```

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -875,13 +1049,109 @@ const UpdateAction: DocumentActionComponent = ({
...
+                        },
+                      ],
+                      disconnect: [],
+                    },
+                    __component: fieldToConnectUID,
+                  };
+              // In case the object was not present you need to pass the componentUID of the parent document
+              const objectToConnect = set({}, fieldToConnectPath, fieldToConnectValue);
+              const dataToUpdate = merge(parentDataToUpdate, objectToConnect);
+
```
markkaylor, line 1111: This same logic is used in two places. I think it should be a function with a signature like:

`connectRelationToParent(parentDataToUpdate, fieldToConnect, { data: res.data })`

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -337,7 +375,17 @@ const RelationModal = ({ children }: { children: React.ReactNode }) => {
...
                 });
                 if (!state.hasUnsavedChanges) {
-                  navigate(getFullPageUrl(currentDocumentMeta));
+                  if (isCreating) {
+                    navigate(
+                      `/content-manager/${currentDocumentMeta.collectionType}/${currentDocumentMeta.model}/create${
+                        currentDocumentMeta.params?.locale
+                          ? `?plugins[i18n][locale]=${currentDocumentMeta.params.locale}`
+                          : ''
+                      }`
```
markkaylor, line 384: It looks like you are computing this same value on link 484 could it be a function?

PR #23391, opened 2025-04-17
chore: resolve babel

Bassel17:
### What does it do?

resolve babel dependencies to a newer version

### Why is it needed?

trying to resolve  https://github.com/advisories/GHSA-968p-4wvh-cqc8

### How to test it?

tests to should be successful


docs/package.json
```diff
@@ -25,6 +25,11 @@
       "last 1 safari version"
     ]
   },
+  "resolutions": {
+    "@babel/helpers": "7.27.0",
+    "@babel/runtime": "^7.26.10",
+    "@babel/runtime-corejs3": "^7.26.10"
```
innerdvations, line 31: Should we use a fixed version here to keep them in sync? I know babel is picky with that (and might be part of the reason for the problems in the other PR)

jhoward1994: I think so yeah, it should match the root. I just pushed it

jhoward1994 (approved):

innerdvations (approved):

PR #23392, opened 2025-04-17
chore: update cli cloud

Mcastres:
This PR modifies the Cloud CLI to allow different prompts originating from the Cloud CLI API

### DIFF
--- a/packages/cli/cloud/src/login/action.ts
+++ b/packages/cli/cloud/src/login/action.ts
@@ -63,7 +63,7 @@ export default async function loginAction(ctx: CLIContext): Promise<boolean> {
     return false;
   }
   await trackEvent(ctx, cloudApiService, 'willLoginAttempt', {
-    ...(promptExperiment && { promptExperiment })
+    ...(promptExperiment && { promptExperiment }),
   });
 
   logger.debug('🔐 Creating device authentication request...', {
@@ -158,7 +158,8 @@ export default async function loginAction(ctx: CLIContext): Promise<boolean> {
           );
           spinnerFail();
           await trackEvent(ctx, cloudApiService, 'didNotLogin', {
-            loginMethod: 'cli', ...(promptExperiment && { promptExperiment })
+            loginMethod: 'cli',
+            ...(promptExperiment && { promptExperiment }),
           });
           return false;
         }
@@ -169,7 +170,8 @@ export default async function loginAction(ctx: CLIContext): Promise<boolean> {
           logger.debug(e);
           spinnerFail();
           await trackEvent(ctx, cloudApiService, 'didNotLogin', {
-            loginMethod: 'cli', ...(promptExperiment && { promptExperiment })
+            loginMethod: 'cli',
+            ...(promptExperiment && { promptExperiment }),
           });
           return false;
         }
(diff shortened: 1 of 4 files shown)


packages/cli/cloud/src/login/action.ts
```diff
@@ -62,7 +62,7 @@ export default async function loginAction(ctx: CLIContext): Promise<boolean> {
     logger.debug(e);
     return false;
   }
-  await trackEvent(ctx, cloudApiService, 'willLoginAttempt', {});
+  await trackEvent(ctx, cloudApiService, 'willLoginAttempt', { promptExperiment });
```
kibwashere, line 65: so here since `promptExperiment` will be `undefined` for the control group right?
it means we send to the tracker something like `{ promptExperiment: undefined }`, is it all good?

kibwashere: and it's "already" sent via `ctx`, since `promptExperiment` is inside the first parameter here - still all good?

kibwashere (dismissed):
let's 🐛 Ride the Sandworm

packages/cli/cloud/src/login/action.ts
Mcastres: You are right, I will conditionnaly include the property since I don't want `undefined` value

Jimimimi (approved):

packages/cli/create-strapi-app/src/cloud.ts
```diff
@@ -42,10 +44,11 @@ export async function handleCloudLogin(): Promise<void> {
     },
   ]);
 
-  if (userChoice !== 'Skip') {
+  if (!userChoice.includes('Skip')) {
     const cliContext = {
       logger,
       cwd: process.cwd(),
+      ...(cloudApiConfig.projectCreation?.reference && { promptExperiment: cloudApiConfig.projectCreation?.reference }),
```
kibwashere, line 51: hey @Mcastres @Jimimimi, I just realized something (started to learn the CLIs last week) here:
because of the update in the CMS:
- before (old version will still have `userChoice !== 'Skip'` right?): we check the exact word `Skip`
- after: `!userChoice.includes('Skip')`

Now the Cloud CLI the answerrs are ([here](https://github.com/strapi/cloud/pull/2814/files)):
- before: `Skip`
- now (for all since it's not versioned): `Walk the Dunes (Skip)`

So here for old version CMS CLI users, we will have an issue, don't you think?
because we will try to match the new skip answer, with the operator `!==`

What do you think?

gonbaum: This is a valid remark. Using `.include()` should fix it. We can use: 
`.toLowerCase().includes("skip")`

gonbaum (commented):
Looks good! 🦾 
I think you only need to fix the double -> single quotes for the linter.

gonbaum (dismissed):

Jimimimi (dismissed):

Jimimimi (approved):

packages/cli/create-strapi-app/src/cloud.ts
kibwashere: @gonbaum, maybe I'm not familiar enough but not sure how your solution would fix things,
I feel the problem is more that old version users will always have the `!==` operator, so what we are trying to do we the new copies is:
`"Walk the Dunes (Skip)" !== 'Skip'` -> which is impossible to solve :(

kibwashere: it should be all goodish :) because of:
image

since message/choices are "fixed", then into the cms-cli old user version will only see the old possibilities!
(but then it means we will have a mix of copies/answers!)

So up to date users:
image

Users on an older version:
- `Choose your journey` -> `Please login or sign in`
- `Ride the Sandworm (Login/Sign up)` -> `Login/Sign up`
- `Walk the Dunes (Skip)` -> `Skip`

PR #23397, opened 2025-04-18
future: check permissions before rendering widgets

remidej:
### What does it do?

Uses the permissions array provided by each plugin. Only display each widget to the user if they have all the permissions in the array

### Why is it needed?

Security / content safety
### How to test it?

- Have the widgets API future flag (it's enabled in getstarted)
- You should see the CM widgets when logged in as super admin since you have the CM explorer.read permission
- Log in as a user that does not have the CM eplorer.read permission (aka someone who has no read permissions at all, both collection types and single types). You should see the CM widgets in their "no permissions" mode

### DIFF
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -234,7 +234,7 @@ const AuthProvider = ({
             (perm) =>
               perm.action === permission.action &&
               // Only check the subject if it's provided
-              (perm.subject == undefined || permission.subject === permission.subject)
+              (perm.subject == undefined || perm.subject === permission.subject)
           ) >= 0
       );
 
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -231,7 +231,10 @@ const AuthProvider = ({
       const matchingPermissions = actualUserPermissions.filter(
         (permission) =>
           permissions.findIndex(
-            (perm) => perm.action === permission.action && perm.subject === permission.subject
+            (perm) =>
+              perm.action === permission.action &&
+              // Only check the subject if it's provided
+              (perm.subject == undefined || permission.subject === permission.subject)
           ) >= 0
       );
 
--- a/packages/core/admin/admin/src/pages/Home/HomePage.tsx
+++ b/packages/core/admin/admin/src/pages/Home/HomePage.tsx
@@ -48,7 +48,12 @@ export const WidgetRoot = ({
       setPermissionStatus(shouldGrant ? 'granted' : 'forbidden');
     };
 
-    if (!permissions || permissions.length === 0) {
+    if (
+      // TODO: remove unstable check once widgets API is stable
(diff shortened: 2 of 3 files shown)


markkaylor (commented):
Hey I tested it:

- If I have permissions to at least one collection type or single type I see this:
Screenshot 2025-04-22 at 09 38 23 in this case shouldn't we see the Widget.NoPermissions?
- If I don't have permissions to any single types or collection types I just have a blank homepage. Do we want to do that instead of displaying Widget.NoPermissions for all installed widgets?

remidej:
Thanks @markkaylor. On your 2 points:
- I believe that's what we want. If I can only see articles, and no article was edited for some reason, then I think it makes sense to show me an empty view. If instead I got a forbidden view, then I'd ask my app admin to look into my permissions, because I know I should be able to see articles.
- Good point, something to ask @lucasboilly: is it better to show nothing at all? Or a bunch of "no permissions" widgets?

lucasboilly:
@remidej a bunch of "no permission" widgets. In the future, a widget will be hidden only if the user decides to not add it to the homepage. It's different from not having access to its content.

packages/core/admin/admin/src/features/Auth.tsx
```diff
@@ -231,7 +231,10 @@ const AuthProvider = ({
       const matchingPermissions = actualUserPermissions.filter(
         (permission) =>
           permissions.findIndex(
-            (perm) => perm.action === permission.action && perm.subject === permission.subject
+            (perm) =>
+              perm.action === permission.action &&
+              // Only check the subject if it's provided
+              (perm.subject == undefined || permission.subject === permission.subject)
```
markkaylor, line 237: Shouldn't this be:

```suggestion
              (perm.subject == undefined || perm.subject === permission.subject)
```

If so I wonder if the undefined check is still necessary 🤔

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -14,21 +14,47 @@ import { useStrapiApp } from '../../features/StrapiApp';
...
 
 interface RootProps {
   title: MessageDescriptor;
   icon?: typeof import('@strapi/icons').PuzzlePiece;
+  permissions?: WidgetType['permissions'];
   children: React.ReactNode;
 }
 
-export const WidgetRoot = ({ title, icon = PuzzlePiece, children }: RootProps) => {
+export const WidgetRoot = ({
```
markkaylor, line 30: This component is used in the stable version and the future version. Do the changes here have an impact on the stable version?

packages/core/admin/admin/src/features/Auth.tsx
remidej: Oof that would have been bad, thanks for spotting it

packages/core/admin/admin/src/pages/Home/HomePage.tsx
remidej: I made a small tweak to always grant the permissions if the future flag isn't on 👍

markkaylor (approved):
🎈

PR #23400, opened 2025-04-19
fix(upload-aws-s3): use baseUrl even if upload location lacks protocol

### LINKED ISSUE #23399: AWS-S3 provider does not use `baseUrl` when server returns Location without the protocol
### Node Version

20.14.0

### NPM/Yarn/PNPM Version

pnpm 9.15.5

### Strapi Version

5.12.3

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When an S3-compatible provider does not return the protocol (`https`) as part of the `Upload.Location` property, Strapi is not adding `baseUrl` to the final URL. This is a problem when using a CDN url that is different from that of the bucket.

Strapi already contains code that handles S3 providers that don't include the protocol as part of the Location prop, as introduced by https://github.com/strapi/strapi/pull/14654:

https://github.com/strapi/strapi/blob/7c7ba817d00aeac2e9f750e6693dc8e51a8a2659/packages/providers/upload-aws-s3/src/index.ts#L120-L124

When the `baseUrl` property was introduced, this else block was not updated to handle it properly.

### Steps to Reproduce

1. Add the `@strapi/provider-upload-aws-s3` plugin to a Strapi instance
2. Set up a provider that does not return the protocol in the `Location` prop (I have been using Hetzner Object Storage, but DigitalOcean Spaces is reported to also have an issue according to https://github.com/strapi/strapi/pull/14654)
3. Upload a file
4. Resulting URL of the file does not consider `baseUrl`.

### Expected Behavior

If a `baseUrl` is provided, even if the S3 provider does not return the protocol/https as part of the `Location` prop, it should use it correctly in combination with the `fileKey` to create a valid accessible URL. If a `baseUrl` is not provided, then it should fall back to the current behavior.

### Logs
This is a log of the UploadCommandOutput. Notice the Location value without a protocol.

```shell
{
  '$metadata': {
    //
  },
  Bucket: 'bucketName',
  ETag: 'x',
  Key: 'file.png',
  Location: 'region.your-objectstorage.com/bucketName/file.png'
}
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

EnriqCG:
### What does it do?

I have extended the use `baseUrl` in combination with the `fileKey` for providers that do not return a protocol in the `Location` prop. When the the `baseUrl` is not provided, it defaults to the previous behavior: use `Location` if it has the protocol, or append the protocol to `Location` if it's missing.

### Why is it needed?

The combination of using a `baseUrl` with a provider that does not return the protocol is broken. It never uses `baseUrl` in that case.

### How to test it?

1. Add the @strapi/provider-upload-aws-s3 plugin to a Strapi instance
2. Set up a provider that does not return the protocol in the Location prop (I have been using Hetzner Object Storage - I am happy to provide some access keys for you to test).
3. Upload a file
4. Resulting URL of the file should consider `baseUrl`

### Related issue(s)/PR(s)

Fixes #23399.

### DIFF
--- a/packages/providers/upload-aws-s3/src/__tests__/upload-aws-s3.test.ts
+++ b/packages/providers/upload-aws-s3/src/__tests__/upload-aws-s3.test.ts
@@ -39,6 +39,7 @@ describe('AWS-S3 provider', () => {
       const file: File = {
         name: 'test',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
         path: 'tmp',
         hash: 'test',
@@ -66,6 +67,7 @@ describe('AWS-S3 provider', () => {
       const file: File = {
         name: 'test',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
         path: 'tmp',
         hash: 'test',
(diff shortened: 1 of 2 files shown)


packages/providers/upload-aws-s3/src/__tests__/upload-aws-s3.test.ts
```diff
@@ -39,6 +39,7 @@ describe('AWS-S3 provider', () => {
       const file: File = {
         name: 'test',
         size: 100,
+        sizeInBytes: 100 * 1024,
```
EnriqCG, line 42: This prop was missing in this object but was defined in the `File` type. I have taken the opportunity to add here to remove a type error in this file.

packages/providers/upload-aws-s3/src/index.ts
```diff
@@ -117,8 +129,11 @@ export default {
 
       const upload = (await uploadObj.done()) as UploadCommandOutput;
 
-      if (assertUrlProtocol(upload.Location)) {
-        file.url = baseUrl ? `${baseUrl}/${fileKey}` : upload.Location;
+      if (baseUrl) {
+        file.url = `${baseUrl}/${fileKey}`;
+      } else if (assertUrlProtocol(upload.Location)) {
```
nclsndr, line 134: fixed linting `else { if {` => `else if {`

EnriqCG:
Great to see this has been merged!

PR #23407, opened 2025-04-22
chore(homepage): move widget controllers, services, and routes from admin to content-manager

markkaylor:
### What does it do?

- Move the routes, controllers, services into a homepage feature domain folder in the content-manager
- Remove all the homepage apis in the admin

### Why is it needed?

- To complete the migration

### How to test it?

- The feature should work as expected with and without the future flag
- There should be no server side code referencing homepage in the admin package

### Related issue(s)/PR(s)


markkaylor:
Looking into the failing tests

remidej:
@markkaylor Maybe just widgets as the name would work? As you want though, I'm not super troubled by homepage references here but I see your point

tests/api/core/content-manager/permissions-conditions.test.api.ts
remidej, line 1: TS conversion at the same time. Chapeau

packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
remidej, line 1: So just to be sure, we'll still need to delete this when we drop the future flag?

remidej (approved):
Tested, looks good

packages/core/admin/admin/src/pages/Home/components/ContentManagerWidgets.tsx
markkaylor: Correct

markkaylor:
Just gonna do any experimental before merging for my own sanity

remidej:
@markkaylor we're about to do some testing on it this afternoon with the marions so you should be safe

PR #23410, opened 2025-04-23
feat: add cloud CLI user info integration

jhoward1994:
## Description

This PR enhances the admin registration experience by integrating with the Strapi Cloud CLI to pre-fill user information when creating the first admin user. When users have already authenticated with the Strapi Cloud CLI during project creation, their email is automatically pre-filled in the admin registration form.

### What does it do?

- Introduced a new service (cloud-user-info.ts) to manage cloud user information
- Added functionality to store cloud user email in a temporary file during project creation
- Implemented a mechanism to read the stored email and pass it to the admin frontend through the Strapi features config
- Updated the admin registration form to utilize cloud user data for pre-filling the email field
- Added cleanup by deleting the temporary user info file after it's used

### Why is it needed?

When users have already authenticated with the Strapi Cloud CLI, they currently need to re-enter the same information when creating their first admin user. This creates unnecessary friction in the onboarding process.

### How to test it?

1. **Prerequisites:**
   - Authenticate with the Strapi Cloud CLI

2. **Testing steps:**
   - Create a new Strapi project
   - Visit the admin registration page (first-time setup)
   - Verify that your email from the Cloud CLI are pre-filled

3. **Testing edge cases:**
   - Test without cloud CLI (should work normally without pre-filling)

### Related issue(s)/PR(s)

DX-1881


packages/core/admin/server/src/routes/admin.ts
```diff
@@ -66,4 +66,10 @@ export default [
       ],
     },
   },
+  {
+    method: 'GET',
+    path: '/cloud-user-info',
+    handler: 'admin.getCloudAuthInfo',
+    config: { auth: false },
```
Convly, line 73: I'm a bit concerned by this. Does this mean anyone with the admin URL can get access to the cloud CLI data? :thinking:

jhoward1994: Yeah it would be a public endpoint but only return the info of the user who authenticated through the cloud CLI

Convly (commented):
So if I've followed this right, if there is no admin user, it'll add the cloud info of who created the app to the window object so that it can be reused, right?

jhoward1994:
Yeah that's right, if there is no admin user it will load the user info from the `config.json` created by the cloud CLI if it's present during bootstrap. It then puts the email of the cloud user on the window object which is loaded into the form in `packages/core/admin/admin/src/pages/Auth/components/Register.tsx`

packages/cli/create-strapi-app/src/cloud.ts
```diff
@@ -49,7 +49,30 @@ export async function handleCloudLogin(): Promise<void> {
     };
 
     try {
-      await cloudCli.login.action(cliContext);
+      const loginSuccessful = await cloudCli.login.action(cliContext);
+
+      // If login was successful, try to get the user email
```
gonbaum, line 54: I think the comments are nice-to-have, but the code itself is quite self-evident. Feel free to keep them though!

packages/cli/create-strapi-app/src/cloud.ts
```diff
@@ -49,7 +49,30 @@ export async function handleCloudLogin(): Promise<void> {
...
+      if (loginSuccessful) {
+        try {
+          // Get token and user info from API
+          const { getValidToken } = await cloudServices.tokenServiceFactory(cliContext);
+          const token = await getValidToken(cliContext, async () => false);
+
+          if (token) {
+            const api = await cloudServices.cloudApiFactory(cliContext, token);
```
gonbaum, line 62: Perhaps the variable name 'api' could be more descriptive, like 'cloudApiClient ' or something like that.

packages/core/admin/server/src/bootstrap.ts
```diff
@@ -137,4 +138,30 @@ export default async ({ strapi }: { strapi: Core.Strapi }) => {
...
+
+        strapi.config.set('features', {
+          ...currentFeatures,
+          initialUserInfo: {
+            email: cloudUserInfo.email,
+          },
+        });
+
+        // Delete the temporary file after we've used it to prevent reuse
+        cloudUserInfoService.deleteCloudUserInfo();
```
gonbaum, line 161: Don't we need to await this async operation?

gonbaum: Disregard this comment. I didn't notice you were using `fs.unlinkSync()`

gonbaum (approved):

Convly:
Based on internal discussions, we might want to re-use the `.cloud-info` file and store information during project creating rather than creating a brand new one. We'll update the PR soon with the related changes. Flagging this in the meantime :rocket:

Convly:
Hey, what's the status on this? Should it be rewritten in a separate PR? Updated? Should we just close this one? cc @jhoward1994

PR #23412, opened 2025-04-23
future: track registered widgets

remidej:
### What does it do?

Adds information about registered widgets to the didInitializeAdministration tracking event

### Why is it needed?

So we can watch usage of the widgets API
### How to test it?

Open your admin. The first tracking event, that should only appear once, should be didInitializeAdministration. Look at the event payload. In the groupProperties object, you should see a registeredWidgets property that should be an array of UID strings (e.g. `global::mywidget`, or `plugin::myplugin.mywidget`)


oiorain (approved):
LGTM

simotae14 (approved):

simotae14:
it works
Screenshot 2025-04-24 alle 10 14 26

PR #23416, opened 2025-04-24
fix: widget type not exported from @strapi/admin/strapi-admin

markkaylor:
### What does it do?

Export the Widget type from `@strapi/admin/strapi-admin`

### Why is it needed?

To make TS happy

### How to test it?

Run yarn test:ts:front the tests should pass

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/23397

It looks like it's causing an error on the release branch https://github.com/strapi/strapi/pull/23413

but I don't get it the code isn't on that branch 🤔

### DIFF
--- a/packages/core/admin/admin/src/index.ts
+++ b/packages/core/admin/admin/src/index.ts
@@ -68,6 +68,7 @@ export type {
   Entity,
 } from '../../shared/contracts/shared';
 export type { RBACContext, RBACMiddleware } from './core/apis/rbac';
+export type { Widget as WidgetType } from './core/apis/Widgets';
 
 /**
  * Utils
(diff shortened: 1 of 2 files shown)


remidej (approved):
I guess TS doesn't care about future flags, as long as it's in the code it will be checked. Weird that we didn't get this in our own PRs though 🤔

markkaylor:
> I guess TS doesn't care about future flags, as long as it's in the code it will be checked. Weird that we didn't get this in our own PRs though 🤔

I also think it's weird that it didn't fail on any other PR. But I think it was introduced in the last permissions PR which is on develop but shouldn't be on 5.12.6 release branch. Yet it's failing on the release branch but not develop 🤦‍♂️. Very confused.

PR #23420, opened 2025-04-24
Limiting the api tracker to api requests

maccomaccomaccomacco:
What does it do?

Why is it needed?

How to test it?

### DIFF
--- a/packages/core/core/src/services/metrics/middleware.ts
+++ b/packages/core/core/src/services/metrics/middleware.ts
@@ -19,8 +19,7 @@ const createMiddleware = ({ sendEvent }: { sendEvent: Sender }) => {
   const middleware: Core.MiddlewareHandler = async (ctx, next) => {
     const { url, method } = ctx.request;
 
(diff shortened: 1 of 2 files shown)


packages/core/core/src/services/metrics/middleware.ts
```diff
@@ -19,7 +19,11 @@ const createMiddleware = ({ sendEvent }: { sendEvent: Sender }) => {
   const middleware: Core.MiddlewareHandler = async (ctx, next) => {
     const { url, method } = ctx.request;
 
-    if (!url.includes('.') && ['GET', 'PUT', 'POST', 'DELETE'].includes(method)) {
+    if (
+      !url.includes('.') &&
+      url.includes('/api') &&
```
Convly, line 24: Instead of hard-coding the `'/api'` prefix, you can get it dynamically from the REST API config.

You can access it like this

```typescript
interface APIConfig {
  rest: {
    prefix: string;
    // ... other props
  }
}

const { rest } = strapi.config.get('api');

console.log(rest.prefix);
```

PR #23421, opened 2025-04-24
enhancement: data transfer CLI command prompts

maccomaccomaccomacco:
What does it do?
Adding support for the strapi transfer command; running this command the user can chose the kind of operation to run and insert the needed data (url and token). Furthermore, the URL and token values fall back to dedicated environment variables value.
425207109-2f6070c0-1053-4945-bd16-4f5baf4ef1f4

Why is it needed?
Inserting the many and long options needed to transfer/push or transfer/pull data can be confusing and difficult; and it's so even more when the transfer is run iteratively.

How to test it?
Just run the strapi transfer command and enjoy the journey.
Set the STRAPI_TRANSFER_URL and STRAPI_TRANSFER_TOKEN env variables to enjoy it even more.

### DIFF
--- a/packages/core/strapi/src/cli/commands/transfer/command.ts
+++ b/packages/core/strapi/src/cli/commands/transfer/command.ts
@@ -53,16 +53,17 @@ const command = () => {
       )
       .hook('preAction', async (thisCommand) => {
         const opts = thisCommand.opts();
+        const hasEnvUrl = process.env.STRAPI_TRANSFER_URL;
+        const hasEnvToken = process.env.STRAPI_TRANSFER_TOKEN;
 
         const logDocumentation = () => {
-          const hasUrl = process.env.STRAPI_TRANSFER_URL;
-          const hasToken = process.env.STRAPI_TRANSFER_TOKEN;
-
           console.info(
             'ℹ️  Data transfer documentation: https://docs.strapi.io/dev-docs/data-management/transfer'
           );
+        };
 
-          if (!hasUrl && !hasToken) {
+        const logEnvironmentVariables = () => {
+          if (!hasEnvUrl && !hasEnvToken) {
             console.info('ℹ️  No transfer configuration found in environment variables');
             console.info(
               '   → Add STRAPI_TRANSFER_URL and STRAPI_TRANSFER_TOKEN environment variables to make the transfer process faster for future runs'
@@ -72,13 +73,13 @@ const command = () => {
 
           console.info('ℹ️  Found transfer configuration in your environment:');
 
-          if (hasUrl) {
+          if (hasEnvUrl) {
             console.info(
-              `   → Environment STRAPI_TRANSFER_URL (${hasUrl}) will be used as the transfer URL`
+              `   → Environment STRAPI_TRANSFER_URL (${hasEnvUrl}) will be used as the transfer URL`
             );
           }
 
-          if (hasToken) {
+          if (hasEnvToken) {
             console.info(
               '   → Environment STRAPI_TRANSFER_TOKEN value will be used as the transfer token'
             );
@@ -88,15 +89,19 @@ const command = () => {
         };
 
         const determineDirection = async () => {
+          // If user has not provided a direction from CLI, log the documentation
+          if (!opts.from && !opts.to) {
+            logDocumentation();
+          }
+
+          logEnvironmentVariables();
+
           if (opts.from) {
             return opts.from;
           }
           if (opts.to) {
             return opts.to;
           }
-
-          // If user has not provided a direction from CLI, log the documentation
-          logDocumentation();
 
           const { dir } = await inquirer.prompt([
             {


innerdvations:
You need to run `yarn lint:fix` on your code to fix the `pretty` tests

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -44,14 +44,128 @@ const command = () => {
...
+                  default: process.env.STRAPI_TRANSFER_URL,
+                  validate(input) {
+                    try {
+                      const url = new URL(input);
+                      if (!['http:', 'https:'].includes(url.protocol)) {
+                        return 'URL must use http: or https: protocol';
+                      }
+                      return true;
+                    } catch (error) {
+                      return 'Please enter a valid URL (e.g., http://localhost:1337 or https://example.com)';
```
innerdvations, line 105:
```suggestion
                      return 'Please enter a valid URL (e.g., http://localhost:1337/admin or https://example.com/admin)';
```
And eventually we need to make it retry with /admin if it doesn't have it and the initial connection fails

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -44,14 +44,128 @@ const command = () => {
...
+            } else if (hasUrl || hasToken) {
+              if (hasUrl) {
+                console.info('ℹ️  Found partial transfer configuration:');
+                console.info('   → STRAPI_TRANSFER_URL environment variable value will be used as default URL');
+                console.info('   → STRAPI_TRANSFER_TOKEN environment variable not found ');
+              } else {
+                console.info('ℹ️  Found partial transfer configuration:');
+                console.info('   → STRAPI_TRANSFER_URL environment variable not found');
+                console.info('   → STRAPI_TRANSFER_TOKEN environment variable value will be used as default token');
+              }
```
innerdvations, line 71:
```suggestion
            if (hasUrl || hasToken) {
              console.info('ℹ️  Found transfer configuration in your environment:');
              if (hasUrl) {
                console.info('   → STRAPI_TRANSFER_URL environment variable value will be used as default URL');
              }
              if(hasToken) {
                console.info('   → STRAPI_TRANSFER_TOKEN environment variable value will be used as default token');
              }
```

I found this a little confusing/distracting when I was testing it. If I only set one of the two, I don't expect to see the other, and my brain doesn't want to read all that but is forced to to find out why there's something I didn't set, it felt like something went wrong 😆 For me, I'd prefer to just see the item that I set.

If we do want to keep it this way, maybe we could make the "found" items green or something and the not-found items yellow or something?

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -44,14 +44,128 @@ const command = () => {
...
+          
+          // Validate that exactly one direction is specified
+          if (opts.from && opts.to) {
             exitWith(
               1,
               'Exactly one remote source (from) or destination (to) option must be provided'
-            )
-        )
+            );
+          }
```
innerdvations, line 167: Can we keep the error check in its own hook?  basically leave the exact code that was already there, and just put it after your new hook?

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -44,14 +44,122 @@ const command = () => {
...
+                console.info('   → STRAPI_TRANSFER_URL environment variable value will be used as default URL');
+              }
+              if(hasToken) {
+                console.info('   → STRAPI_TRANSFER_TOKEN environment variable value will be used as default token');
+              }
+            } else {
+              console.info('ℹ️  No transfer configuration found in environment variables');
+              console.info('   → Add STRAPI_TRANSFER_URL and STRAPI_TRANSFER_TOKEN environment variables to make the transfer process faster for future runs');
+            }
+            console.log(); // Empty line for better readability
```
innerdvations, line 70:
```suggestion
            console.info(); // Empty line for better readability
```
Since we used .info above

innerdvations:
When an env var is given, we shouldn't prompt for it, but other than that LGTM

innerdvations (dismissed):

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -45,14 +45,142 @@ const command = () => {
...
+
+          if (process.env.STRAPI_TRANSFER_URL) {
+            return new URL(process.env.STRAPI_TRANSFER_URL);
+          }
+
+          const answer = await inquirer.prompt([
+            {
+              type: 'input',
+              name: 'remoteUrl',
+              message: `Enter the URL of the remote Strapi instance to ${direction === 'from' ? 'get data from' : 'send data to'}:`,
```
Convly, line 134: Should we make it clear from the beginning that the /admin part of the path is expected? (currently it's only in the error handling)

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -45,14 +45,142 @@ const command = () => {
...
+                }
+              },
+            },
+          ]);
+
+          return new URL(answer.remoteUrl);
+        };
+
+        const determineToken = async (direction: 'from' | 'to') => {
+          if (opts[`${direction}Token`]) {
```
Convly, line 154: So if a token is provided, it takes precedence over the env value?

Convly: Saw that afterward, but yeah https://github.com/strapi/strapi/pull/23421#issuecomment-2850485479

packages/core/strapi/src/cli/commands/transfer/command.ts
innerdvations: @maccomaccomaccomacco what do you think? I think it's a good idea I just don't see how to reasonably fit it into one line.

packages/core/strapi/src/cli/commands/transfer/command.ts
innerdvations: Yeah, that's how I usually expect most tools to work. env values used by default, but provide a mechanism to override it case-by-case using the cli.

maccomaccomaccomacco: lgtm this way

packages/core/strapi/src/cli/commands/transfer/command.ts
maccomaccomaccomacco: I don't think we can do much more than say "Enter the _admin_ URL of the remote Strapi instance to..."

maccomaccomaccomacco:
overall, lgtm

packages/core/strapi/src/cli/commands/transfer/command.ts
innerdvations: For now let's leave it, but I think a future improvement could be that if we do not detect it ends with '/admin' we log a warning message. That would basically solve the problem and we wouldn't need the more complex solutions we've discussed before (like appending it and retrying connection)

innerdvations (approved):

packages/core/strapi/src/cli/commands/transfer/command.ts
derrickmehaffy: It's not always /admin as that can be changed so we need to accommodate for that if we do

maccomaccomaccomacco: Can't we just let the user put the base Url and get the admin path from the admin configuration?

packages/core/strapi/src/cli/commands/transfer/command.ts
```diff
@@ -45,14 +45,142 @@ const command = () => {
...
+          }
+
+          logEnvironmentVariables();
+
+          if (opts.from) {
+            return opts.from;
+          }
+          if (opts.to) {
+            return opts.to;
+          }
```
WDeenik, line 104: Shouldn't this be

```typescript
if (opts.from) {
  return 'from';
}
if (opts.to) {
  return 'to';
}
```

Right now if I do `strapi transfer --from [URL] --from-token [TOKEN]` it still asks me about a URL and token, and the direction message on line 134 also says 'send data to', because the `direction` will be set to my `--from` URL instead of to the string `'from'`.

Issue #23423, opened 2025-04-24
Releases created by roles that do not have super-admin permissions for admin users is limited to 3


derrickmehaffy:
### Node Version

node/v20.18.2

### NPM/Yarn/PNPM Version

yarn/1.22.22

### Strapi Version

5.12.6

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

When you as a user who doesn't have super admin permissions (default author/editor roles with releases permissions) tries to create more than 3 releases you will be blocked by license limits that shouldn't exist.

This is primarily caused because one specific internal route has a default policy that requires the following permission nodes:

- admin::users.create
- admin::users.read
- admin::users.update
- admin::users.delete

Because of these permissions sensitive nature (gives total control over admin users) some users may not want to grant these permissions to more restricted roles but those roles should still be able to have complete control over releases.

### Steps to Reproduce

IMPORTANT: You CANNOT reproduce this as a super admin, you must have a lower role.

1. Create a new application
2. Modify the admin `Author` role to have full permissions on the releases feature (under the plugins section)
3. Create a new user as the `Author` role
4. Try to create more than 3 releases
5. See license restriction warning
6. Log into the super admin account
7. See that you can create more than 3 releases

### Expected Behavior

Lower roles should be able to use their paid features up to their limits regardless of role if the proper permissions are set

### Logs

```shell

```

### Code Snippets

The problem largely comes from this route:

https://github.com/strapi/strapi/blob/167a9bf54fe8f7f1746540f004c1e9d9b5ab21c6/packages/core/admin/ee/server/src/routes/license-limit.ts

Specifically the `admin::hasPermissions` policy which requires the elevated permission nodes, the releases feature calls this every time the releases main page is loaded to see how many releases can be created

This policy requires that the user have access to all of these permissions:

https://github.com/strapi/strapi/blob/167a9bf54fe8f7f1746540f004c1e9d9b5ab21c6/packages/core/admin/server/src/policies/hasPermissions.ts

### Media

Reproduction with a seperate lower level user: https://www.loom.com/share/02466c55f3b74ff28ff57083a12011e7?sid=1bb13f14-7730-4c19-9844-043c00270809

Here is another reproduction I did as a super user (albeit the license I was using actually had a hard limit of 3, make sure you use a license that has a higher limit to test)

https://www.loom.com/share/848ed7b35cb342049c22a33e99dbdaaf

[...]

PR #23424, opened 2025-04-24
fix: remove policy from license-limit-info that breaks releases

derrickmehaffy:
### What does it do?

Removes the `admin::hasPermissions` policy from the `/license-limit-information` route to allow lower level users without the ability to CRUD admin users to create more than 3 workflows (up to their license limit)

### Why is it needed?

Currently only super-admins or roles with the following permissions could create more than 3 releases

- admin::users.create
- admin::users.read
- admin::users.update
- admin::users.delete

### How to test it?

See reproduction in https://github.com/strapi/strapi/issues/23423

### Related issue(s)/PR(s)

fixes: #23423

### DIFF
--- a/packages/core/admin/ee/server/src/routes/license-limit.ts
+++ b/packages/core/admin/ee/server/src/routes/license-limit.ts
@@ -7,20 +7,7 @@ export default {
       path: '/license-limit-information',
       handler: 'admin.licenseLimitInformation',
       config: {
-        policies: [
-          'admin::isAuthenticatedAdmin',
-          {
-            name: 'admin::hasPermissions',
-            config: {
-              actions: [
-                'admin::users.create',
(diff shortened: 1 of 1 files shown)


derrickmehaffy:
Don't Merge yet, pending review on why that policy even was added in the first place.
See Internal Slack Thread: https://strapihq.slack.com/archives/C01J6DEK1RS/p1745528911887619

remidej:
Hey @derrickmehaffy, I see you removed the don't merge flag, do you think it's safe in the end security-wise? the thread you linked to doesn't seem to have a definitive answer

derrickmehaffy:
I believe so yes, at the end of the day we require an admin panel user to be logged in to see this and the data in this endpoint isn't sensitive

Issue #23305, opened 2025-04-03
Repeatable components can be deleted under the published tab


lucasboilly:
### Node Version

22.11.0

### NPM/Yarn/PNPM Version

Yarn 4.5.0

### Strapi Version

5.12.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

Repeatable components can be deleted under the published tab, but it should be only possible under the draft tab.

### Steps to Reproduce

1. Find an entry with a repeatable component
2. Add entries in the list
3. Publish
4. Go to the published tab
5. Click the delete `IconButton` on some components 

### Expected Behavior

The delete and drag icon buttons should be disabled.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #23427, opened 2025-04-25
fix: updated icon button to include disabled flag for repeatable comp

ankit7201:
### What does it do?

Disables the icon button for repeatable component in published tab 

### Why is it needed?

See #23305 

### How to test it?

Go to a repeatable component and try clicking on delete icon when in Published tab. 
Nothing will happen as it is disabled. 

Go to draft tab and click the delete button and it will work. 

### Related issue(s)/PR(s)

#23305

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Repeatable.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Repeatable.tsx
@@ -439,6 +439,7 @@ const Component = ({
             <Accordion.Trigger>{displayValue}</Accordion.Trigger>
             <Accordion.Actions>
               <IconButton
+                disabled={disabled}
                 variant="ghost"
                 onClick={onDeleteComponent}
                 label={formatMessage({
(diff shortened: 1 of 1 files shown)


ankit7201:
Hi @lucasboilly 

Can you please review this?

lucasboilly:
Hi @ankit7201, really cool! 

Could you disable the drag `IconButton` that's just next to it as well? And add a 4px gap between the 2 `IconButtons` so they're not sticked together?

Thanks for your help 🙌

ankit7201:
Hi @lucasboilly 

I have updated it and now drag button is also disabled. 

Regarding the 4px gap task, I have added the condition to add the gap only is disabled mode since in non disabled mode, the button spacing looks good. 
Please correct me if I am wrong, but wouldn't it make more sense to fix the gap issue in strapi design system (as a separate PR) rather than here since this seems like a general issue and accordion should control how its icons should look like.

lucasboilly:
@ankit7201 you are right about the gap, let's forget this and we'll fix it later in the DS 👍

So you just need to remove the condition and the `Flex` around the buttons if I'm not wrong?

It's all good regarding the `Trash` and `Drag` icon buttons 👌

ankit7201:
@lucasboilly 
Reverted the gap code. 
Thanks.

lucasboilly:
All good, thanks again @ankit7201!

I'm gonna share it to our engineers for the code review :)

remidej (approved):

lucasboilly:
Hi @ankit7201, your code was shipped in the last Strapi version ([5.12.7](https://github.com/strapi/strapi/releases/tag/v5.12.7)) :)

ankit7201:
@lucasboilly 
Awesome. Thanks.

PR #23428, opened 2025-04-25
future(homepage): missing addTagTypes results in stale data

markkaylor:
### What does it do?

Fixes the cache invalidation for the homepage widgets in the content manager

### Why is it needed?

The cache becomes stale and requires a refresh

### How to test it?

Make sure you have the future flag set for unstableWidgetsApi (set by default in examples/getstarted)
Go edit an entry
Go to the homepage
You should see your update without refreshing
Go publish and entry
Go to the hompage
You should see your published entry without refreshing

https://github.com/user-attachments/assets/5a4d236a-0049-4b35-b799-893371facd6c


remidej (approved):
works perfectly

PR #23439, opened 2025-04-28
Feat: make api token viewable

Bassel17:
### What does it do?

- Make it possible to view api tokens after creating them
- it store and encrypts the tokens 
- decrypts them for viewing

### How to test it?

1. Display the "View Token" Button
Given I am logged into the Strapi Admin Panel and I navigate to the API Tokens section,
When I click on an API token,
Then I should see a button labeled "View token",
And when I select it, the actual token should be revealed.

2. Copy the Token
Given the token is visible,
When I click on the "Copy" button,
Then the token should be copied to my clipboard,
And a confirmation message should appear (e.g., "Token copied successfully!").

3. Token Auto-Hides After a Time Interval
Given I have revealed the token,
When a certain period (e.g., 30 seconds) passes,
Then the token should automatically be masked again.

4. Role-Based Access Control for Viewing Tokens
Given I am an admin user with the read permission,
When I try to view a token,
Then I should be able to see and copy it.


packages/core/admin/admin/src/translations/en.json
```diff
@@ -265,7 +265,7 @@
   "Settings.tokens.RegenerateDialog.title": "Regenerate token",
   "Settings.tokens.copy.editMessage": "For security reasons, you can only see your token once.",
   "Settings.tokens.copy.editTitle": "This token isn’t accessible anymore.",
-  "Settings.tokens.copy.lastWarning": "Make sure to copy this token, you won’t be able to see it again!",
+  "Settings.tokens.copy.lastWarning": "Make sure to copy this token.",
   "Settings.tokens.duration.30-days": "30 days",
```
jhoward1994, line 269: Maybe this doesn't need to be as much of a warning now?

```suggestion
  "Settings.tokens.copy.message": "Copy this token for use elsewhere.",
```
I think "Settings.tokens.copy.editMessage" could be worded differently now too ?

packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/EditViewPage.tsx
```diff
@@ -173,6 +175,23 @@ export const EditView = () => {
     }
   }, [data]);
 
+  React.useEffect(() => {
+    // Only set up timer when token is shown
+    if (showToken) {
+      hideTimerRef.current = setTimeout(() => {
+        setShowToken(false);
+      }, 30000); // 30 seconds
```
jhoward1994, line 183: So it auto hides after 30 seconds and that's the only way of hiding the token?

For me being able to manually close/hide it seems like a better experience? Or
maybe having both so we still have the added security of auto hiding ?

Bassel17: we can have both, just following the acceptance criteria
@maccomaccomaccomacco can make the decision on that

packages/core/admin/admin/src/translations/en.json
Bassel17: yep this just follows the design changes
@maccomaccomaccomacco it's your time to shine

packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/EditViewPage.tsx
innerdvations: Yeah, could the "View" button toggle to "Hide"?

packages/core/admin/admin/src/translations/en.json
innerdvations: Another more general product question: what should we do in the case of a translation string change? In this case I think it makes sense to leave it slightly incorrect rather than removing them, but technically all the translations will be incorrect now.

packages/core/admin/server/src/content-types/api-token.ts
```diff
@@ -47,6 +47,13 @@ export default {
       required: true,
       searchable: false,
     },
+    encryptedKey: {
+      type: 'string',
+      minLength: 1,
+      configurable: false,
+      required: true,
+      searchable: true,
```
innerdvations, line 55:
```suggestion
      searchable: false,
```

packages/core/admin/server/src/services/encryption.ts
```diff
@@ -0,0 +1,57 @@
+import crypto from 'crypto';
+
+const IV_LENGTH = 16; // 16 bytes for AES-GCM IV
+/**
+ * Encrypts a value string using AES-256-GCM.
+ * Returns a string containing IV, encrypted content, and auth tag (all hex-encoded).
+ */
+const encrypt = (value: string) => {
+  const encryptionKey: string = strapi.config.get('admin.secrets.encryptionKey');
```
innerdvations, line 9: This causes a breaking change, because `regenerate` no longer works until the user adds that config value.

Maybe we should use the existing `auth.secret` as a fallback and document that if you want to set a separate encryption key you are able to set it in `secrets.encryptionKey`? Or, if we don't want to do it globally, we could allow `encrypt` here to optionally accept a key, so that API tokens (which are technically part of auth) could explicitly pass in `auth.secret`

We definitely need a better way to manage all these secrets, we should have a single base key we use universally, unless a user explicitly wants to be more granular with it for key rotations.

@Convly @jhoward1994 @maccomaccomaccomacco what do you think?

packages/core/admin/server/src/services/encryption.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (!encryptionKey) {
+    strapi.log.warn('Encryption key is missing from config');
+    return null;
+  }
+  const iv = crypto.randomBytes(IV_LENGTH);
+  const cipher = crypto.createCipheriv('aes-256-gcm', Buffer.from(encryptionKey, 'hex'), iv);
```
innerdvations, line 15: I don't know if it's only on my end, but I get type errors here and down below

packages/core/admin/server/src/services/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      },
+    } as any;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('encrypt', () => {
+    test('encrypts and returns a colon-separated string', () => {
```
innerdvations, line 28:
```suggestion
    test('encrypts and returns a three-part colon-separated string (iv, auth, enrypted)'), () => {
```
just because I was confused why we had this test 😆

packages/core/admin/server/src/services/encryption.ts
maccomaccomaccomacco: Well, I stopped at the first line 

> This causes a breaking change, because regenerate no longer works until the user adds that config value.

we can't provide breaking changes, we need a workaround

Bassel17: it's not a breaking change, regenerate will work, it just won't be a viewable token if no key is provided, same for creating new tokens, this is the way we avoided it being a breaking change for users, once they set an encryption key, then they will be able to view tokens that they create or regenerate

packages/core/admin/server/src/services/encryption.ts
Bassel17: no type errors on my side or the linter, anyways this section did change a bit with a new thing I added

packages/core/admin/server/src/services/encryption.ts
maccomaccomaccomacco: newly created projects have this key by default, right?

We could then make the viewable token disabled if this key is not provided and show a tooltip saying that they need to insert the key in their configuration. Wdyt?

Bassel17: currently it will log a warning in the server, but if we want to add something visual, yeah why not 
a tooltip could work, we currently don't show the `view token` button, i can have it be shown but disabled with a tooltip message

packages/core/admin/admin/src/translations/en.json
Bassel17: yeah I was thinking about translations too, we'll need to translate to all the languages if we need it to be correct
I wish we had a script or something we can just run and it corrects the translations

packages/core/admin/server/src/services/encryption.ts
maccomaccomaccomacco: I think it would be better to display it disabled, with the tooltip on hover stating "In order to view the token, you need an encryption key in the admin configuration" (@innerdvations help me with the wording :D )

Sorry if it wasn't described in the acceptance criteria in the first place! :)

innerdvations: > it's not a breaking change, regenerate will work

I tested by trying to regenerate in the admin without that key set, I got an error and it failed to regenerate. Switched back to `develop` and it worked. That was my concern. Maybe regenerate just needs a try/catch around when it generates the encryptionKey though and that would prevent it from being breaking.

> newly created projects have this key by default, right?

I don't think projects have this key added by default, it's a new one we're adding here.

>  "In order to view the token, you need an encryption key in the admin configuration" (@innerdvations help me with the wording :D )

I think your message is fine, I just wonder if we even need it. We've been adding way too many keys. If we can possibly use something that already exists like auth.secret I think it would be better

maccomaccomaccomacco: > I don't think projects have this key added by default, it's a new one we're adding here.
we need it then

> We've been adding way too many keys. If we can possibly use something that already exists like auth.secret I think it would be better
🤔

Bassel17: yes new projects will have the encryption key generated for them after we merge this of course

innerdvations: Ahh, sorry, I see that it's being added to new projects.

I'll check the regenerate again!

packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/EditViewPage.tsx
Bassel17: yeah it currently auto hides, the question was if we want the button to also hide on/off, basically a show and hide button, it's a very simple addition to add for me if we choose to go that way

innerdvations: I don't think it's important, it could come as a later improvement. The user can always navigate away if they want it hidden quickly

packages/core/admin/server/src/content-types/api-token.ts
```diff
@@ -47,6 +47,13 @@ export default {
       required: true,
       searchable: false,
     },
+    encryptedKey: {
+      type: 'string',
+      minLength: 1,
+      configurable: false,
+      required: true,
```
innerdvations, line 54:
```suggestion
      required: false,
```
Just wondering if we actually need this to be required, since it's not really, and in fact all current tokens won't have it.

packages/core/admin/server/src/services/encryption.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (!key) return null;
+  const iv = crypto.randomBytes(IV_LENGTH);
+  const cipher = crypto.createCipheriv('aes-256-gcm', key, iv);
+
+  let encrypted = cipher.update(value, 'utf8', 'hex');
+  encrypted += cipher.final('hex');
+
+  const authTag = cipher.getAuthTag();
+
+  return `${iv.toString('hex')}:${encrypted}:${authTag.toString('hex')}`;
```
innerdvations, line 30:
```suggestion
  return `v1:${iv.toString('hex')}:${encrypted}:${authTag.toString('hex')}`;
```
We may want to version this so we don't run into problems in the future if we have to change anything

packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/EditViewPage.tsx
maccomaccomaccomacco: let me check with Louis if he can come out with something very quick; but if you need to wait for this thing only in order to approve the PR, let's skip it and let's handle it later

packages/core/admin/server/src/services/encryption.ts
Bassel17: yeah and about the keys, having `auth.secret` as the key instead, am not so sure about that, I think the problem would be that auth.secret is currently for JWT, the api token and transfer token salts are scoped for their purpose too, having one single key for all is interesting but it does bring a bigger discussion of where and which one.
and for this case `auth` wouldn't make sense since encryption is not `auth`, also if people want to rotate keys in a global setup it forces them to rotate for all things affected. This just brings it's own future headache, if we want to do that I would create a fallback key for all in like a `secrets` object

packages/core/admin/server/src/services/encryption.ts
Bassel17: that's a very nice idea I like it, will add it

packages/core/admin/admin/src/pages/Settings/pages/ApiTokens/EditView/EditViewPage.tsx
Bassel17: talked to Louis, we'll have it also hide, and have a hide eye icon

maccomaccomaccomacco: wait, what do you mean with "we'll have it also hide"?

packages/core/admin/admin/src/pages/Settings/components/Tokens/FormHead.tsx
```diff
@@ -151,6 +158,31 @@ export const FormHead = <TToken extends Token | null>({
...
             )}
+            {token?.id && toggleToken && (
+              <Tooltip
+                label={
+                  !canShowToken &&
+                  formatMessage({
+                    id: 'Settings.tokens.encryptionKeyMissing',
+                    defaultMessage:
+                      'In order to view the token, you need an encryption key in the admin configuration',
+                  })
```
jhoward1994, line 169: I wonder if here we should distinguish between the 2 cases:
- no encryption key present at all
- encryption key available but not one that can display this token

?

packages/core/admin/admin/src/pages/Settings/pages/TransferTokens/EditView.tsx
```diff
@@ -263,6 +263,7 @@ const EditView = () => {
                 }}
                 token={transferToken}
                 setToken={setTransferToken}
+                canShowToken={false}
                 canEditInputs={canEditInputs}
```
jhoward1994, line 267: And maybe the message should be different for transfer tokens given that you
wont be able to see them again ?

We could revert back to "Make sure to copy this token, you won’t be able to see it again!"

jhoward1994 (dismissed):
It works well for me, I only had a couple of minor suggestions

packages/core/admin/admin/src/pages/Settings/components/Tokens/FormHead.tsx
Bassel17: I could say a "valid encryption key"

packages/core/admin/admin/src/pages/Settings/pages/TransferTokens/EditView.tsx
Bassel17: indeed

packages/core/admin/admin/src/pages/Settings/components/Tokens/FormHead.tsx
Bassel17: the frontend can't tell really if exists or not, it just knows "i don't have a key to display"

jhoward1994 (dismissed):

jhoward1994 (approved):

PR #23470, opened 2025-05-05
feat: make widgets api stable

remidej:
### What does it do?

drops the future flag for the widgets api, allowing users to register custom widgets on their admin homepage. and removes dead code that was the result of moving the CM widgets from the admin to the CM package during the future flag phase

### Why is it needed?

homepage customization!

### How to test it?

- register a widget, you should see it on the admin homepage
- the pre existing content manager widgets should keep working like before


markkaylor (approved):
Great it's working 🚀

Issue #23255, opened 2025-03-28
Strapi 5.12.0 and translations


sergejpapper:
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

YARN 1.22.22

### Strapi Version

5.12.0

### Operating System

Windows 11

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

There are some problems (please check attached image) with Strapi translations after updating to 5.12.0 version.

![Image](https://github.com/user-attachments/assets/0554e711-55ae-474d-bde5-64685cee80c0)

As you can see there is "undefined" word in translate path.

### Steps to Reproduce

1. create new collection type (ex. form) and field inside (ex. successMessage);
2. in src/admin/extensions/translations/en/en.ts file add row
  'content-manager.content-types.api::form.form.successMessage': 'Success message'

### Expected Behavior

Field should be translated from "successMessage" to "Success message", but because of undefined in path nothing is happened.

![Image](https://github.com/user-attachments/assets/698e8827-7b6d-4032-b8c2-26160db26d5d)

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

sergejpapper:
Can you please give an update on it?

After three weeks and couple of new Strapi versions problem still exists. We lose all our translations in this way.

imryck:
Hi, we have encountered the same issue since upgrading to version 5.12.0.

With a quick look, it seems related to this line: https://github.com/strapi/strapi/blob/d8f8da8f5b333be4a20563a1a15cd61350139956/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx#L45

Previously, it was retrieved from the path `documentMeta.model`. This was changed in the related commit: https://github.com/strapi/strapi/commit/7c06b73932ae059464ff3abb5fe2d130dee6960f.

A naive fix would be to change it to `document.model`, which provides the model name correctly. I do not see any other impact at first glance. Do you think this is a good way to fix this issue?

Thanks.

imryck:
The fix seems much safer on relying on the schema property (`document.schema?.uid). The model property is present on the JavaScript objet but does not appears on the type `ReturnType<UseDocument>`. I'm preparing a PR on the subject.

PR #23471, opened 2025-05-05
fix(content-manager): use UID instead of “undefined” for missing translations

maximpostnikov:
### What does it do?

- To fix the "undefined" string, use `uid` instead of `modelName`.
- Remove the now-unused model?: string prop from the FormLayoutProps interface.

### Why is it needed?

In development mode, schema.modelName is always undefined, so every translation key fell back to:
```
Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "content-manager.content-types.undefined.content"
```

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
@@ -36,13 +36,12 @@ export const ResponsiveGridItem =
 
 interface FormLayoutProps extends Pick<EditLayout, 'layout'> {
   hasBackground?: boolean;
-  model?: string;
   document: ReturnType<UseDocument>;
 }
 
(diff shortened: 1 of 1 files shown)


markkaylor (approved):
Tested, LGTM. Thank you for handling this fix 🙏

markkaylor:
fixes: #23255

markkaylor:
How to test:

In examples/getstarted:

In `examples/getstarted/src/admin/app.jsx` add the following:
```
const config = {
  locales: ['it', 'es', 'en', 'en-GB'],
  translations: {
    en: {
      'content-manager.content-types.api::kitchensink.kitchensink.short_text': 'TEST!!'
    }
  }
};
```
Start the server, make sure your app is in english
Go to kitchensink content type in the content manager
The field short_text should be translated to "Test!!" in the listview table and editview field

PR #23472, opened 2025-05-05
fix: define an initial ComponentProvider on top of the Relation Modal

simotae14:
### What does it do?

It avoids having issues of passing the ComponentProvider from the Edit view document to a new document in the create relaton

### Why is it needed?

Because otherwise if you try to create a document starting from a root document that contains a component the uid and id of the component are passed to the children

### How to test it?

Try to add a collection that contains a component with a relation and then try to create a new document from the parent DZ component, the new Document needs to contains another DZ

### Related issue(s)/PR(s)

CS-1419

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
@@ -333,6 +333,11 @@ const RelationModal = ({ children }: { children: React.ReactNode }) => {
   const currentDocument = useRelationModal('RelationModalForm', (state) => state.currentDocument);
   const isCreating = useRelationModal('RelationModalForm', (state) => state.isCreating);
 
+  /*
+   * We must wrap the modal window with Component Provider with reset values
+   * to avoid inheriting id and uid from the root document and having weird
+   * behaviors with simple relationships..
+   */
   return (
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -333,94 +334,96 @@ const RelationModal = ({ children }: { children: React.ReactNode }) => {
...
-          });
-        }
-      }}
-    >
-      {children}
-      <StyledModalContent>
-        <Modal.Header gap={2}>
-          <Flex justifyContent="space-between" alignItems="center" width="100%">
-            <Flex gap={2}>
+    <ComponentProvider id={undefined} level={-1} uid={undefined} type={undefined}>
```
remidej, line 337: We need a comment to explain why this is necessary

remidej (commented):
This looks like a great fix :)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
simotae14: ok sure

remidej (approved):

markkaylor (approved):

PR #23477, opened 2025-05-06
fix: use text for encrypted key of viewable api token

innerdvations:
### What does it do?

Uses 'text' instead of 'string' to store the encrypted key

### Why is it needed?

The size of the string exceeds the length of varchar(255) so mysql and postgres fail to start up

### How to test it?

- Set the encryptedKey config property
- using mysql + postgres
  - start up strapi; it should work now

### Related issue(s)/PR(s)


Convly (approved):
The code LGTM, thanks!

PR #23479, opened 2025-05-06
fix(data-transfer): fix large transfer crashes; show transfer progress

### LINKED ISSUE #23485: [data-transfer] Some critical errors are reported as a successful transfer
### Node Version

22.13.1

### NPM/Yarn/PNPM Version

yarn 4.5.0

### Strapi Version

5.13.0

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When run `yarn strapi transfer --from ...` to the local destination, if the local destination runs out of disk space, the transfer ends but says that it completed successfully.

There are other cases than running out of disk space where the same thing can occur; there are certain places in the transfer code where errors are caught and ignored rather than bubbled up.

I believe that the same issue would also occur in a push transfer but didn't test it.

### Steps to Reproduce

Run `yarn strapi transfer --from ...` to the local destination, if the local destination runs out of disk space, the transfer ends but says that it completed successfully.

### Expected Behavior

The transfer should report a failure instead of a success

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

### LINKED ISSUE #25175: Tracking Issue: Data Transfer fails with large media libraries
### Node Version

20

### NPM/Yarn/PNPM Version

YARN 3

### Strapi Version

5.33.2

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

This is a tracking issue for all data transfer issues that involve large media transfers (larger than ~400mb total)

The problem is caused by a memory leak during transfers that causes media files to be held in memory until the transfer has completed, when it is then finally released. That means that, generally, any transfer where the total media file size is close to or greater than the available memory on the destination will fail, often with inconsistent error messaging

Temporary workarounds until it is fixed:

- run the transfer with `--exclude files` and then uploading them to your upload directory via another method.
- increase memory size on the destination -- the memory leak is temporary and released on completion, so if the transfer can succeed the Strapi instance will still be in a good state afterwards

### Steps to Reproduce

- Run a `transfer --to` to a system with less memory than the size of the media library files
- See that the transfer will generally fail

### Expected Behavior

The transfer should succeed even if the total media size is larger than the available memory on the destination

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

innerdvations:
### `@strapi/data-transfer`

- **Asset bytes on the wire:** stream chunks use **base64** payloads (`transfer-asset-chunk.ts`) instead of Buffer’s default JSON shape (`{ type: 'Buffer', data: […] }`), which explodes memory on `JSON.parse` for large files. Decoders still accept **legacy Buffer JSON** and plain base64 strings so older peers keep working.

- **WebSocket responses:** a shared **`replacerForTransferWebSocket`** (`transfer-websocket-json.ts`) stringifies Buffer / typed-array values as base64 in outgoing frames, so incidental binary in messages is less likely to serialize as huge number arrays.

- **Pull path (`pull.ts`):** builds asset batches with the chunk helper; optional **SHA-256 per asset** when checksums are negotiated; on **assets** step **start**, runs **`estimateAssetTotals`** and returns **`totals`** for progress (ignored by older clients).

- **Push / pull receivers:** shared decode path for asset stream rows; **`remote-destination`** batches outgoing chunks the same way.

- **`remote-source` (`createAssetsReadStream`):** processes each incoming batch through a **writable with `highWaterMark: 1`** so only one batch is processed at a time and backpressure/deadlocks around piping are avoided. **Stall detection** (`streamTimeout`, overridable from the CLI via config) resets on incoming chunks and after **downstream drain** so slow reads do not look like a stall; timers are **cleared** when an asset ends or the stream tears down. **One warning** if the remote still sends legacy Buffer JSON for chunks (detected from the parsed payload). Optional **checksum verification** on pull when enabled.

- **`remote-destination`:** mirrors the above for push, including **checksum negotiation** with the remote.

- **Checksum negotiation:** **`init`** can include **`checksums: true`**; the server answers whether it supports it. When both sides agree, pull/push attach **SHA-256** to each asset’s **`end`** and the receiver validates (or logs and continues if the peer does not support checksums).

- **Engine:** **`#progressTrackerChunks`** wraps each asset stream so **byte progress updates per chunk** (not only per file), keeps a single consumer, and avoids double-counted bytes/speed. **`getStageTotals('assets')`** (when provided) seeds **total bytes / count** before the assets stage so the CLI can show **fractions and ETA**.

- **Local source:** **`estimate-asset-totals.ts`** walks the same upload records as **`createAssetsStream`** but only sums sizes/counts (no open file handles). **`createAssetsStream`** can warn when DB rows point at **missing files**; those warnings go through **diagnostics** so the CLI can show them.

### `@strapi/strapi` CLI

- **`strapi transfer`:** reads optional **`server.transfer.remote.assetIdleTimeoutMs`** for **`--from`** (maps to **`streamTimeout`** on the remote source provider). **`--no-checksums`** turns off end-to-end asset checksum verification for remote source and destination (default is on).

- **Progress:** ora lines refresh on a short interval so **elapsed time** moves during slow byte progress; **readable** durations; for **assets**, **transferred vs total** size and **count / totalCount** when totals exist, plus a rough **ETA** after a short warm-up.

- **Diagnostics:** always writes **info** (and the rest) to the timestamped **log file** and logs the file’s **absolute path**; **console** shows **warn+** by default and **info** when **`--verbose`** (via **`consoleLevel`** on the file logger).

### `@strapi/logger`

- **`createOutputFileConfiguration`:** optional **`consoleLevel`** so the file can stay verbose while the console stays quiet.

### Examples

- **`examples/complex/config/server.ts`:** **`transfer.remote.enabled`** 
- **`examples/complex`** / **`examples/empty`:** **`engines.node`** bumped to **`>=20.0.0 <=24.x.x`**.

### Tests / CI

- Unit and provider tests cover chunks, engine progress, remote streams, checksum negotiation, and CLI behavior. **Remote pull/push** CLI tests and helpers live under **`tests/cli`** (see **`tests/cli/README.md`** for optional heavier media env vars).

**Backward compatibility:** asset **decoding** stays tolerant of legacy wire shapes; **checksums** are opt-in via negotiation and can be disabled with **`--no-checksums`**.

---

## Why is it needed?

Large **remote** transfers (especially **pull**) could **OOM**, look **stuck** (progress only advanced per file), or hit **backpressure/stall** bugs. Asset payloads as Buffer JSON made parsing especially expensive. This branch addresses those issues, improves **visibility** (progress, totals, warnings, logs), and adds **optional integrity checks** without breaking older Strapi versions on the other end.

---

## How to test it?

Latest experimental: ``0.0.0-experimental.11f10e472f3e424651fbe8ee40a844786e8d836f`

- **Unit:** `yarn test:unit` (or target `@strapi/data-transfer` / `@strapi/strapi` CLI tests).
- **CLI:** `yarn test:cli -d strapi --testPathPattern=data-transfer`
- **Remote e2e (slower):** `tests/cli/README.md` — **`pull-remote`**, **`push-remote`**, optional **`TRANSFER_CLI_MEDIA_*`**.
- **Manual:** `strapi transfer --from` / `--to` with large media — smooth byte progress, totals/ETA when both ends support them, **`--no-checksums`** if you need to rule out checksum overhead, and reasonable memory vs a pre-change build.

---

## Documentation needed

- **`server.transfer.remote.assetIdleTimeoutMs`:** optional **`config/server`**, milliseconds without forward progress while **`strapi transfer --from`** streams assets; applies to the **CLI machine** pulling from the remote. When to raise it (very large files, slow links, backpressure).

---

## Related issue(s) / PRs

- Fixes CMS-112
- Fixes #23485
- Fixes #25175


packages/core/data-transfer/src/engine/index.ts
```diff
@@ -316,8 +316,34 @@ class TransferEngine<
     aggregate?: {
       size?(value: unknown): number;
       key?(value: unknown): string;
-    }
+    },
+    perChunk?: boolean
   ) {
+    if (perChunk) {
```
Convly, line 322: I'm not a big fan of always adding more options and branches to the flow, could we try to simplify it a bit?

For instance, since the perChunk option seems to only be useful if coupled with the asset stage, could we remove the option and instead branch based on the stage value?

packages/core/data-transfer/src/engine/index.ts
```diff
@@ -916,10 +942,15 @@ class TransferEngine<
     const destination = await this.destinationProvider.createAssetsWriteStream?.();
 
     const transform = this.#createStageTransformStream(stage);
-    const tracker = this.#progressTracker(stage, {
-      size: (value: IAsset) => value.stats.size,
-      key: (value: IAsset) => extname(value.filename) || 'No extension',
-    });
+    // Use the perChunk tracker for assets
+    const tracker = this.#progressTracker(
```
Convly, line 946: Instead of modifying the generic method, we could also just have a dedicated method to create an assetProgressTracker? Might be easier to maintain

innerdvations: Yeah, I can split it off into a different method instead of using an option

packages/core/data-transfer/src/engine/index.ts
innerdvations: I'd rather have the stage itself decide, since it needs to provide additional info anyway. And even though it's only on assets stage for now, I thought it would be better to keep it generic so that we could use it anywhere else in the future, no reason for it to be tied to assets necessarily.

innerdvations:
Oops, I broke the speed indicator for assets, and also noticed it's double-reporting the filesize. Fixing that now.

packages/core/strapi/src/cli/utils/helpers.ts
```diff
@@ -19,13 +19,40 @@ const readableBytes = (bytes: number, decimals = 1, padStart = 0) => {
...
+  } else if (elapsedTime >= 1000) {
+    value = elapsedTime / 1000;
+    unit = 's';
+  } else {
+    value = elapsedTime;
+    unit = 'ms';
+  }
+
+  const floored = floorToDecimal(value, decimals);
+  const result = `${floored.toFixed(decimals)}${unit}`;
```
Bassel17, line 52: I am a little confused do we actually need `.toFixed` here ?

Bassel17 (dismissed):
Overall looks good !
just left a small comment

packages/core/strapi/src/cli/utils/helpers.ts
innerdvations: Yeah, this is what keeps it from flickering when we move from something like "29.9" to "30" and keeps it as "30.0"

Convly (dismissed):

jhoward1994 (dismissed):
The code looks good to me. I think I'll need a bit of help to set up an asset transfer, let me know if you want me to test that too

innerdvations:
Some more QA revealed that this corrupts files; investigating it now

markkaylor:
Hey I tried to test this but when transferring to the remote instance I keep getting rejected with this error:

```
 error: [FATAL] Failed to initialize the connection: Data transfer is not enabled on the remote host
Transfer process failed.
```

I created the transfer token and explicitly set `STRAPI_DISABLE_REMOTE_DATA_TRANSFER=false` in strapi cloud. Currently in my local instance I have 3 images about 8mb each. Not sure what I've done wrong 🤔

innerdvations:
> error: [FATAL] Failed to initialize the connection: Data transfer is not enabled on the remote host
> Transfer process failed.

Make sure it's all enabled in config

`config/server.js:`

```
transfer: {
  remote: {
    enabled: true,
  },
},
```

`config/admin.js:`

```
transfer: {
  token: {
    salt: 'your-secret-salt-at-least-16-chars',
  },
},
```

markkaylor:
Alright I got it to work! The problem was I was missing `/admin` at the end of my app url.

My first attempt I got this error
Screenshot 2026-03-20 at 08 37 30

I retried and it worked!
Screenshot 2026-03-20 at 08 36 43

As I mentioned I have 3 large images (11mb, 8mb, and 5mb). Is that enough or should I try with more files, bigger files, or both?

innerdvations:
Specifically what we need to confirm is transfers to a remote server where the total of the media library is larger than the memory on the system now succeed.

Additionally, even after success, we should manually confirm with a checksum that the files were transferred intact without any chunks out of order, since an earlier fix I had appeared to work even though it would corrupt files.

derrickmehaffy:
experimental: https://github.com/strapi/strapi/actions/runs/23548451672

derrickmehaffy:
experimental version: `0.0.0-experimental.e5e1e18bb446666fd3d1f255874900dec87d9854`

derrickmehaffy:
new exp: https://github.com/strapi/strapi/actions/runs/23594271010/job/68706879649

derrickmehaffy:
@innerdvations the latest experimental failed, looks like a problem with review workflows?

derrickmehaffy:
New EXP: https://github.com/strapi/strapi/actions/runs/23737134256/job/69144594748

OcneanuVlad:
Exp version: `0.0.0-experimental.163ec76e070cb6bd198adc261e1909aefe09a6a1`

packages/core/data-transfer/src/strapi/remote/handlers/pull.ts
```diff
@@ -247,10 +257,19 @@ export const createPullController = handlerControllerFactory<Partial<PullHandler
 
       const flushUUID = randomUUID();
 
+      let totals: StageTotalsEstimate | undefined;
+      if (step === 'assets') {
+        totals = await estimateAssetTotals(strapi as Core.Strapi);
```
Bassel17, line 262: I have a small concern here, if this takes a long time to accomplish we have logic in the client that times out after a certain amount of retries

innerdvations: Pushed two changes: (1) longer retry window for the assets start message on remote pull so slow totals don’t hit the dispatcher timeout. (2) Faster totals: local still uses stat on disk; remote uses size from the DB for main + formats when all are present, and only falls back to signing + HTTP when something is missing — so we avoid per-file HTTP in the common case.

innerdvations:
Latest experimental: `0.0.0-experimental.0e1799c5b2176fdca775c7ccedfc87c7ad27f60d`

innerdvations:
@claude review

innerdvations:
New experimental to fix some `push` issues found in testing

`0.0.0-experimental.613553bfe58920e602f1eaa7f1e2053733ee2bd3`

innerdvations:
New experimental: `0.0.0-experimental.efc28eaabf4e75159e06914d9f2f8059073b806f`

cc @butcherZ

innerdvations:
`0.0.0-experimental.11f10e472f3e424651fbe8ee40a844786e8d836f`

butcherZ (approved):
QAed, and it works between two strapi instances locally, might still have issue pushing to remote, but let's release this fix first so we can take our time to explore the issue with cloud 😄

PR #23486, opened 2025-05-07
fix(types): add missing flags prop to admin config

unrevised6419:
### What does it do?

Adds the missing config `Admin['flags']` prop

### Why is it needed?

When typing `config/admin.ts`  TS gives an error

```
Object literal may only specify known properties, and 'flags' does not exist in type 'Admin'.ts(2353)
```

https://docs.strapi.io/cms/configurations/admin-panel#available-options

| Parameter | Description | Type | Default |
|--------|--------|--------|--------|
| `flags` | Settings to turn certain features or elements of the admin on or off | `object` | - |
| `flags.nps` | Enable/Disable the Net Promoter Score popup | `boolean` | `true` | 
| `flags.promoteEE` | Enable/Disable the promotion of Strapi Enterprise features | `boolean` | `true` | 

### How to test it?

Add `Core.Config.Admin` return type to the `config/admin.ts`

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/22620

### DIFF
--- a/packages/core/types/src/core/config/admin.ts
+++ b/packages/core/types/src/core/config/admin.ts
@@ -35,6 +35,11 @@ export interface TransferProp {
   token: TransferTokenProp;
 }
 
+export interface FlagsProp {
+  nps?: boolean | undefined;
+  promoteEE?: boolean | undefined;
+}
+
 export interface Admin {
   // required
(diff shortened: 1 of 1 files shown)


innerdvations:
Thanks for the contribution!

PR #23487, opened 2025-05-07
feat: add rules engine for json logic schema validation and tests

butcherZ:
### What does it do?

create rules engine for generate validate schema based on json logic

### How to test it?

test with rules-engine.test.ts

### Related issue(s)/PR(s)

dx-1989


packages/core/utils/src/rules-engine.ts
```diff
@@ -0,0 +1,73 @@
...
+  generate: (condition: Condition) => JsonLogicCondition;
+  validate: (condition: Condition) => void;
+  evaluate: (condition: JsonLogicCondition, data: unknown) => boolean;
+};
+
+export function createRulesEngine(): RulesEngine {
+  /**
+   * Transforms a high-level `Condition` object into a JSON Logic-compatible condition.
+   *
+   * Converts operators like 'is' and 'isNot' into their JSON Logic equivalents ('==' and '!=').
```
innerdvations, line 22: Do we actually want to do this? If we keep it the same format as JSON logic we can state that we use their format, and point to their docs.

innerdvations (approved):
LGTM! Just one comment

packages/core/utils/src/rules-engine.ts
Bassel17: yeah good question, not a 100% sure what we'll end up with in the end, but it serves the purpose of simplifying a bit instead of having to create the structure multiple times in different places, might be something we move to the CTB internally only, will all be clearer once we use it

Issue #21906, opened 2024-10-17
image crop not working on production builds


cdreier:
### Required System information

- Node.js version: v20.12.0
- NPM version: 10.5.0
- Strapi version: 5.1.0
- Database: sqlite
- Is your project Javascript or Typescript: typescript

### Describe the bug

after building strapi and run `npm start` - the image crop functionality is no longer working. in development mode, everything is working well.

as far as i could see, at least the the cropperjs css is missing

### Steps to reproduce the behavior

1. have at least one image in the media library
2. run `npm run build`
3. run `npm start`
4. try to crop the image

### Expected behavior

the crop overlay is visible and we can crop the image

### Screenshots

![image](https://github.com/user-attachments/assets/18ac4ee6-3618-4d5c-9ed3-470ce2f23c0b)

### Code snippets

in the inspector, you can clearly see, that the `cropper-container` div has no styles attached

### Additional context

as i said, during development, everything is working fine

cdreier:
can anyone help me here?

cdreier:
i tested a bit, i am not sure why the styles are removed on a production build, i also explicitly imported the styles in the admin `app.tsx` with `import 'cropperjs/dist/cropper.css';` which is also missing after running `npm run build` and `npm start`

a workaround for me is to manually add the styles in the admin bootstrap function 

```
bootstrap(app: StrapiApp) {
    // console.log(app);
    const head = document.head || document.getElementsByTagName('head')[0];
    const style = document.createElement('style');
    head.appendChild(style);
    style.appendChild(document.createTextNode(`INSERT CROPPERJS-CSS`));
  },
```

after the styles are back in place, the cropping is working again

huapeixuan:
Same problem in my project. The imported css only work on npm run develop.

mrajaeim:
Same problem in my project.

Topman-14:
> const head = document.head || document.getElementsByTagName('head')[0];
>     const style = document.createElement('style');
>     head.appendChild(style);
>     style.appendChild(document.createTextNode(`INSERT CROPPERJS-CSS`));

@cdreier I have this same issue, I'm quite new to the framework, and I really don't understand how you applied the fix directly in the bootstrap function

cdreier:
> > const head = document.head || document.getElementsByTagName('head')[0];
> > const style = document.createElement('style');
> > head.appendChild(style);
> > style.appendChild(document.createTextNode(`INSERT CROPPERJS-CSS`));
> 
[...]

PR #23547, opened 2025-05-19
fix: change rollup config to include the css file in prod mode

simotae14:
### What does it do?

Update the rollup configuration with a post css plugin to handle css files in production mode

### Why is it needed?

We had some issue importing css files in our components in prod mode, the style was not injected giving us some weird behaviours like in the case of the crop feature in the media library

### How to test it?

 - have at least one image in the media library
 - run npm run build
 - run npm start
 - try to crop the image

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/21906

### DIFF
--- a/rollup.utils.mjs
+++ b/rollup.utils.mjs
@@ -9,6 +9,7 @@ import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
 import commonjs from '@rollup/plugin-commonjs';
 import image from '@rollup/plugin-image';
 import html from 'rollup-plugin-html';
+import postcss from 'rollup-plugin-postcss';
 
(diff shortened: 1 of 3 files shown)


rollup.utils.mjs
```diff
@@ -9,15 +9,22 @@ import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
...
 
 const basePlugins = () => [
   image(),
   html(),
   json(),
+  postcss({
+    extensions: ['.css'],
+    inject: true,
+    minimize: true,
+    extract: false,
```
innerdvations, line 24: We might want to use `extract: true` here for better performance

remidej (approved):
Nice fix 👍

PR #23558, opened 2025-05-20
feat: Front-end for Growth Trial

mazzucchelli:
### What does it do?

Implement all the new UI elements for the upcoming growth free trial logic

### How to test it?

Using a growth trial license, start the admin panel in dev mode


packages/core/core/src/ee/index.ts
```diff
@@ -246,6 +249,10 @@ export default Object.freeze({
   get type() {
     return ee.licenseInfo.type;
   },
+  
+  get licenseKey() {
+    return ee.licenseInfo.licenseKey;
+  },
```
Convly, line 255: @alexandrebodin @innerdvations 

There are two options here:

1. We expose the license key through `strapi.ee.licenseKey`, easy retrieval and usage, but it also means that any code can access it (e.g. plugins). Is this something acceptable or does it contain sensitive information that we shouldn't leak programmatically?)

2. We grab the logic responsible for making the HTTP call to the license registry and create a getter for the licenseExpiresAt instead? It would hide the potential sensitive stuff but would increase the area of responsibility for the ee module

What do you think?

alexandrebodin: I don't see it being use except in a controller, we should remove that and just check if strapi.isEE in that code instead and leave this private if it's not needed.

alexandrebodin: I like option 2 of doing strapi.ee.timeLeft() or sth like that

innerdvations: Any plugin or user code is technically able to get it if they really want to the same way we do. But yeah, I'd prefer to keep it as limited as possible rather than making it this easy if we can.

Option 2 would work. It's not the prettiest option but it's a bit safer.

However, I just thought of a different issue:
- If we use `strapi.fetch`, by default it logs the url retrieved, leaking the license key in server logs
- if we don't use `strapi.fetch`, we will not respect proxy settings

So we might want something like logging options in strapiFetch

packages/core/admin/admin/custom.d.ts
```diff
@@ -21,6 +21,7 @@ interface BrowserStrapi {
     REVIEW_WORKFLOWS: 'review-workflows';
     isEnabled: (featureName?: string) => boolean;
   };
+  isTrial: boolean;
```
markkaylor, line 24: Is this not a bit vague? Maybe `isTrialLicense`? Everywhere else it's more obvious since `isTrial` is accessed via `licenseInfo.isTrial` or `license.isTrial`.

markkaylor (commented):
Hey @mazzucchelli this all looks really good. Running the app without a license seems to work fine. If I run the app with my current license (not growth plan) I get the following error:

https://github.com/user-attachments/assets/22086827-4e7b-407d-b4b5-09c5fcb27bcb

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -177,4 +187,31 @@ export default {
...
+
+    if (!licenseKey) {
+      throw new LicenseCheckError(
+        'License key not found. Please make sure you are using a valid license key.',
+        true
+      );
+    }
+
+    const res = await strapi
+      .fetch(`https://license.strapi.io/api/licenses/${licenseKey}/trial-countdown`, {
```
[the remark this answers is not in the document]
mazzucchelli: `LICENSE_REGISTRY_URI` added to `packages/core/core/src/ee/license.ts`, now both requests consume this const instead of the hardcoded url

packages/core/admin/admin/custom.d.ts
mazzucchelli: ✅ renamed in `isTrialLicense`

packages/core/core/src/ee/index.ts
mazzucchelli: option 2 implemented, however I'm having issues preventing the log to appear! Do you have any suggestions on how to achieve that?

markkaylor:
I created a growth trial license but I was expecting to see a banner at the top:

Screenshot 2025-05-26 at 10 10 13

packages/core/core/src/ee/index.ts
```diff
@@ -234,6 +261,7 @@ const get = (featureName: string) => list().find((feature) => feature.name === f
 export default Object.freeze({
   init,
   checkLicense,
+  trialDaysLeft,
```
Convly, line 264: So after the server starts we wouldn't check again for the number of remaining days? What if the servers stays up for weeks or even months?

Should we implement a getter with a throttling mechanism instead?
This way we could control the update rate

packages/core/core/src/ee/index.ts
Convly: One option **could** be to create a local instance of the Strapi fetch client with logs disabled then use it?

Would require modifying `packages/core/core/src/utils/fetch.ts` with something like
```typescript
import type { Core, Modules } from '@strapi/types';
import { ProxyAgent } from 'undici';

// TODO: once core Node exposes a stable way to create a ProxyAgent we will use that instead of undici

interface StrapiFetchOptions {
  logs?: boolean;
}

// Create a wrapper for Node's Fetch API that applies a global proxy
export const createStrapiFetch = (
  strapi: Core.Strapi,
  options?: StrapiFetchOptions
): Modules.Fetch.Fetch => {
  const { logs = true } = options ?? {};

  function strapiFetch(url: RequestInfo | URL, options?: RequestInit) {
    const fetchOptions = {
      ...(strapiFetch.dispatcher ? { dispatcher: strapiFetch.dispatcher } : {}),
      ...options,
    };

    if (logs) {
      strapi.log.debug(`Making request for ${url}`);
    }

  }

  const proxy =
    strapi.config.get<ConstructorParameters<typeof ProxyAgent>[0]>('server.proxy.fetch') ||
    strapi.config.get<string>('server.proxy.global');

  if (proxy) {
    if (logs) {
      strapi.log.info(`Using proxy for Fetch requests: ${proxy}`);
    }

    strapiFetch.dispatcher = new ProxyAgent(proxy);
  }

  return strapiFetch;
};

export type Fetch = Modules.Fetch.Fetch;
```

and then instantiate it 

```typescript
import { createStrapiFetch } from '../utils'

const silentStrapiFetch = createStrapiFetch(strapi, { logs: false });

// ...
```

What do you think?

packages/core/core/src/ee/index.ts
Convly: oh, my bad! I thought you were computing the number of days left once & then exporting this value only (and not the function)

It's all good then!

mazzucchelli: is `getTrialDaysLeft` better?

packages/core/core/src/utils/fetch.ts
```diff
@@ -20,9 +20,7 @@ export const createStrapiFetch = (
       ...options,
     };
 
-    if (logs) {
-      strapi.log.debug(`Making request for ${url}`);
-    }
+    if (logs) strapi.log.debug(`Making request for ${url}`);
```
Convly, line 23: We tend to avoid single-line ifs in https://github.com/strapi/strapi :eyes: 
```suggestion 
    if (logs) {
      strapi.log.debug(`Making request for ${url}`);
    }
```

packages/core/core/src/utils/fetch.ts
```diff
@@ -32,7 +30,7 @@ export const createStrapiFetch = (
     strapi.config.get<string>('server.proxy.global');
 
   if (proxy) {
-    strapi.log.info(`Using proxy for Fetch requests: ${proxy}`);
+    if (logs) strapi.log.info(`Using proxy for Fetch requests: ${proxy}`);
```
Convly, line 33: Same
```suggestion
    if (logs) {
      strapi.log.info(`Using proxy for Fetch requests: ${proxy}`);
    }
```

Convly (commented):
Small style comment, but the fetch logic seems better now!

innerdvations (approved):
LGTM now with the latest changes! Make sure someone else takes a look at the front-end in case I missed something important there.

Caveat: I did not test it with trial keys, but we can QA this more thoroughly in the feature branch itself.

PR #23559, opened 2025-05-20
fix(data-transfer): abort on errors

### LINKED ISSUE #23485: [data-transfer] Some critical errors are reported as a successful transfer

innerdvations:
### What does it do?

If there was an error in the transfer, don't treat it as a success

### Why is it needed?

fixes #23485

### How to test it?

Test that errors during the transfer process cause the whole transfer to fail and not report success

In particular, test that -handled- errors such as schema mismatch are allowed to be bypassed, and only unhandled errors cause the failure.

### Related issue(s)/PR(s)

fixes #23485

### DIFF
--- a/packages/core/data-transfer/src/engine/index.ts
+++ b/packages/core/data-transfer/src/engine/index.ts
@@ -772,7 +772,20 @@ class TransferEngine<
       // Note: This will be configurable in the future
       await this.destinationProvider.rollback?.(e as Error);
 
+      // Ensure we close providers even on error
+      try {
+        await this.close();
+      } catch (closeError) {
+        this.reportError(closeError as Error, 'error');
+      }
+
       throw e;
+    }
+
+    // Check if there were any errors in the diagnostics
+    const hasErrors = this.diagnostics.stack.items.some((item) => item.kind === 'error');
+    if (hasErrors) {
+      throw new Error('Transfer completed with errors. Check the logs for details.');
     }
 
     return {


innerdvations:
@Convly I wanted to get your opinion on if you think this is too naive an approach or not, and if you think an alternative is better, since you wrote most of the diagnostics and error handling

Convly:
https://github.com/strapi/strapi/pull/23559/files#diff-89c36c71c67a8c5f5ef9fc0ccb4aa4667e838b65d613e1d49d0e04a3de1547ceR786

This is indeed quite aggressive but I'm not opposed to the idea. I'm just worried about what would happen in the case of soft errors like auto schema merging (in case of discrepencies). For now we merge & proceed (and it happens quite a lot between EE & CE env) but with that we would just abort, right?

Convly:
Hey, what's the status on this? Is it still planned or should we close it? cc @innerdvations

innerdvations:
It will be coming back up soon during our upcoming focus on bug fixes!

innerdvations:
This should be fine, I'll reopen it in case anyone has a chance to review even though it's not on the roadmap

innerdvations:
Closing in favor of #23479 which has a similar fix

PR #23562, opened 2025-05-21
fix: avoid history data loss when license is missing

remidej:
### What does it do?

persists the history table

### Why is it needed?

We want to avoid deleting history data when users no longer have a valid license. They should be able to re-add a license and still have access to their old history.

Same thing is done for audit logs already

### How to test it?

User facing:
- Start the app with an EE license
- Make on an entry
- Open the history page, see the history versions
- Restart the app without the license
- Restart the app with the license again
- Check the history page for your entry. You should see the history versions

Or checking the implementation detail: open a database explorer, in the `core_store` table, find the `core_persisted_tables` table. The json content should include `strapi_history_versions`


markkaylor (approved):
Nice, looks good. Tested, works.

Issue #23519, opened 2025-05-14
upload assets error


im-manaosoftware:
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

11.3.0

### Strapi Version

5.12.4-5.13.0

### Operating System

Windows 11

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

1. upload new assets error

![Image](https://github.com/user-attachments/assets/cf8368d3-4632-4be6-91c9-168512eba6bd)

### Steps to Reproduce

1. Media Library
2. Add new assets
3. Browse files

![Image](https://github.com/user-attachments/assets/cf8368d3-4632-4be6-91c9-168512eba6bd)

### Expected Behavior

do not show error when upload image

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/591170ce-fec9-4ddd-a555-8df6724b31c4

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

rajeshrc12:
Hey @im-manaosoftware  can I work on this ?

rajeshrc12:
https://github.com/user-attachments/assets/513b5fee-cc9b-44c9-923b-94f9598da1fc

@im-manaosoftware Could you please confirm this behavior? The error only appears for .jpg or .jpeg formats, while .png files work correctly. However, all files are still getting uploaded regardless of the error.

joaoprcosta:
In my case, when removing the breakpoints object from the `configs/plugins.ts` it started working again correctly. 
I think it's not related with the extension because I hade some jpg and png that worked and others not.

rajeshrc12:
@joaoprcosta Its causing due to permission error, i am looking into it

node:internal/process/promises:391                                                                               
    triggerUncaughtException(err, true /* fromPromise */);                                                       
    ^                                                                                                            
                                                                                                                 
[Error: EPERM: operation not permitted, unlink 'C:\Users\rajesh\AppData\Local\Temp\6b2c145a17729fd9aad529600'] { 
  errno: -4048,                                                                                                  
  code: 'EPERM',                                                                                                 
  syscall: 'unlink',                                                                                             
  path: 'C:\\Users\\rajesh\\AppData\\Local\\Temp\\6b2c145a17729fd9aad529600'                                     
}

[...]

PR #23567, opened 2025-05-21
Upload assets error

rajeshrc12:
### What does it do?

It fixes error message "Internal server error" while upload images
443495746-cf8368d3-4632-4be6-91c9-168512eba6bd

### How to test it?

1.goto media library tab
2.click on add new assets button
3.select image and click on upload button

### Related issue(s)/PR(s)

[23519](https://github.com/strapi/strapi/issues/23519)


rajeshrc12:
@reaksa-maii , it's showing on Local

jhoward1994:
Thanks for this but afaik I don't think this is the right fix.

The change wraps fse.remove() calls with .catch() handlers that only log to console and continue. This masks the underlying error rather than fixing it, e.g. swallowing failures during file cleanup.

I'll close the PR with this comment but feel free to reopen the discussion if needed

PR #23581, opened 2025-05-23
fix(content-type): improve field validation logic in permission settings

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/server/src/services/content-type.ts
+++ b/packages/core/admin/server/src/services/content-type.ts
@@ -178,18 +178,15 @@ const cleanPermissionFields = (
       nestingLevel,
     });
 
-    const requiredFields = getNestedFields(strapi.contentTypes[subject], {
-      components: strapi.components,
-      requiredOnly: true,
-      nestingLevel,
-      existingFields: fields,
-    });
+    const currentFields: string[] = fields || [];
+    const validUserFields: string[] = uniq(intersection(currentFields, possibleFields));
 
-    // @ts-expect-error lodash types
-    const badNestedFields = uniq([...intersection(fields, possibleFields), ...requiredFields]);
-
-    const newFields = badNestedFields.filter(
-      (field) => !badNestedFields.some(startsWith(`${field}.`))
+    const newFields = validUserFields.filter(
+      (field: string) =>
+        !validUserFields.some(
+          (validUserField: string) =>
+            validUserField !== field && startsWith(validUserField, `${field}.`)
+        )
     );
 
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -742,14 +745,35 @@ const PublishAction: DocumentActionComponent = ({
...
       if (errors) {
-        toggleNotification({
-          type: 'danger',
-          message: formatMessage({
-            id: 'content-manager.validation.error',
-            defaultMessage:
-              'There are validation errors in your document. Please fix them before saving.',
-          }),
-        });
+        let hasUnreadableRequiredField = false;
```
jhoward1994, line 748: Without a more significant refactor of this component, **I think the only scenario we can cover with a new specific error message here is when the user cannot read a required field.**

All the publishAction logic is based off "formValues". That is whats seen in the frontend form and what is sent to the publish action in the end.

Even if the draft in the DB has a value in the required field, if the user cannot read that field then it's value in form values will be null/undefined.

So we should always error with a message that says something like "There is a required field without a value". Even if in the DB this field does hold a value (this component doesn't know about that DB value). From the admin users perspective it doesn't have a value (they can't read the field).

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -742,14 +745,35 @@ const PublishAction: DocumentActionComponent = ({
...
+              hasUnreadableRequiredField = true;
+              break;
+            }
+          }
+        }
+
+        if (hasUnreadableRequiredField) {
+          toggleNotification({
+            type: 'danger',
+            message: 'Test error about unreadable required field', // TODO: the new error message ??
```
jhoward1994, line 765: @maccomaccomaccomacco please could you help with the new error message?

maccomaccomaccomacco: "Your current permissions prevent access to certain required fields. Please request access from an administrator to proceed."
or
"You don’t have the necessary permissions to fill out all the required fields. Please request access from an administrator to proceed."

Convly (approved):
Code LGTM, I'll need to test it manually next

innerdvations (approved):
Nice fix! LGTM, tested and works as expected

PR #23602, opened 2025-05-27
chore(content-manager): use dndkit for configure the view drag and drop

markkaylor:
### What does it do?

Disclaimer: it's not perfect...but I think it's an improvement

- Uses dndkit inspired by the new content-type builder
- Adjusts the drag and drop logic. Mainly a new row isn't created when a dragged item doesn't fit in the container its over. This means a user has to resize the item to be <= to the space where they want to drop it. This makes it less smart but in my opinion it improves the experience.
- It should fix the bug where dragging an item over a component starts to drag the component instead of the initial item

### Why is it needed?

The current drag and drop is close to unusable 

### How to test it?

Try dragging, dropping, resizing, and saving. Try dragging and dropping full width items. Try sorting items within a row. 

### Related issue(s)/PR(s)

resolves https://github.com/strapi/strapi/issues/22056


packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -33,6 +39,130 @@ import type { EditLayout } from '../../hooks/useDocumentLayout';
...
+  };
+
+  return (
+    <div ref={setNodeRef} style={style} {...attributes}>
+      {children}
+    </div>
+  );
+};
+
+const useDragAndDrop = (layout: ConfigurationFormData['layout']) => {
```
markkaylor, line 71: ℹ️ The hook isn't really necessary I just found it helpful to isolate as much of the dnd logic as possible

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -198,120 +240,231 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
+                    <DroppableContainer id={container.dndId}>
+                      <Grid.Root gap={2}>
+                        {container.children.map((child, childIndex) => (
+                          <Grid.Item
+                            col={child.size}
+                            key={child.dndId}
+                            direction="column"
+                            alignItems="stretch"
+                          >
+                            <>
```
remidej, line 403: looks like the fragment isn't necessary

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -33,6 +39,130 @@ import type { EditLayout } from '../../hooks/useDocumentLayout';
...
+    return layout
+      .map((row) => ({
+        ...row,
+        children: row.children.filter((field) => field.name !== TEMP_FIELD_NAME),
+      }))
+      .filter((row) => row.children.length > 0)
+      .map((row) => {
+        const totalSpaceTaken = row.children.reduce((acc, curr) => acc + curr.size, 0);
+
+        if (totalSpaceTaken < 12) {
```
Adzouz, line 123: C: We could maybe store this value in a constant to reuse it across the file so that it would be easier to change its value if needed at some point. What do you think?

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -81,71 +120,77 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
-        1,
-        ...recalculatedRows.map((row, index) => ({
-          __temp_key__: rowKeys[index],
-          children: row,
-        }))
-      );
+  /**
+   * Compute uids and formName for drag and drop items for the incoming layout
+   */
+  const createDragAndDropContainersFromLayout = React.useCallback(
```
remidej: i believe you wouldn't need to useCallback here if you moved the function outside the component

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -81,71 +120,77 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
+      }));
+    },
+    []
+  );
+
+  const [containers, setContainers] = React.useState(() =>
+    createDragAndDropContainersFromLayout(layout)
+  );
+  const [activeDragItem, setActiveDragItem] = React.useState<
+    (typeof containers)[number]['children'][number] | null
```
remidej: you're doing this same TS inference in a few places, maybe worth extracting a `type Container = (typeof containers)[number]`?

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -176,131 +221,246 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
+        const activeItem = getItemFromContainer(
+          event.active.id,
+          containersAsDictionary[activeContainer]
+        );
+
+        if (activeItem) {
+          setActiveDragItem(activeItem);
+        }
+      }}
+      onDragOver={({ active, over }) => {
```
remidej, line 259: did you see any performance issues with this? there are a lot of loops and it will re-run on every render. No need to change it if it doesn't have a visible impact though

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -176,131 +221,246 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
+
+        setContainers(update);
+      }}
+      onDragEnd={(event) => {
+        const { active, over } = event;
+        const { id } = active;
+        const overId = over?.id;
+        const containersAsDictionary = Object.fromEntries(
+          containers.map((container) => [container.dndId, container])
+        );
```
remidej: looks like this could be shared between start/over/end

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
simotae14: Yes I agree with Remi

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -81,71 +120,77 @@ const Fields = ({ attributes, fieldSizes, components, metadatas = {} }: FieldsPr
...
+  const [containers, setContainers] = React.useState(() =>
+    createDragAndDropContainersFromLayout(layout)
+  );
+  const [activeDragItem, setActiveDragItem] = React.useState<
+    (typeof containers)[number]['children'][number] | null
+  >(null);
+
+  function findContainer(
+    id: UniqueIdentifier,
+    containersAsDictionary: Record<string, (typeof containers)[number]>
```
simotae14: I can understand the aim of this function but it doesn't seem to me super readable...also I am sure the code to use the new dependency is super difficult to implement (even if we need to adjust just one file) and thanks a lot for your comments to explain every single step, I am just wondering if you can add some more variables or function to abstract a little bit the complexity

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
markkaylor: I understand the concern but so far didn't notice anything. Even on content-types with a lot of fields like kitchensink.

remidej: good to know, let's ignore then 👍

Adzouz:
I pushed a little fix for the full rows because before it was switching the complete row and we needed to only switch the children between rows to avoid creating a weird behaviour of changing indexes.
Also added a small condition to be sure that swapping elements within a row will only happen in the same row.

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
markkaylor: I've added another comment above the function to help explain otherwise I'm not sure what more I could do. If you have specific suggestions happy to update.

oiorain:
Good job! it seems like a good improvement UX wise. 
The 33% or 50% rows get a little stuck after a while when you move the 100% fields about. Otherwise lgtm

packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
simotae14: Thanks a lot for the comments, the code seems more readable now (kudos to you because it is really complex).

simotae14:
We have some e2e tests in the relations e2e branch with some checks about reordering and such, I hope, with your code, they are still gonna work

markkaylor:
> We have some e2e tests in the relations e2e branch with some checks about reordering and such, I hope, with your code, they are still gonna work

The changes here are specific to the editviewpage configure the view, the drag and drop hasn't changed for relations reordering so I don't think it should have an impact. Let me know if you want to take a look together if you have any concerns.

Adzouz (approved):
Good job @markkaylor! 🚀 This one was rough 🫡

Issue #22053, opened 2024-11-02
Cannot show documentIds in Content Manager


Matb85:
### Required System information

- Node.js version: v22.9.0
- NPM version: v10.8.3
- Strapi version: v5.2.0
- Database: postgres
- Operating system: Macos 15.0.1
- Is your project Javascript or Typescript: Typescript

### Describe the bug

Cannot display documentId in the content manager 
### Steps to reproduce the behavior

1. Go to Content Manager
2. View a collection
3. Try to show the documentId field

### Expected behavior

I should be able to see the "documentId" field of the entry. 

AFAIK, each language version of an entry has a different "documentId" then please provide a way to view them all.

Alternatively, allow to query documents by the old "id" field so that knowing "documentId" is not necessary

### Screenshots

See? no "documentId" option

![Screenshot 2024-11-02 at 22 47 08](https://github.com/user-attachments/assets/5e53b8dd-6f2f-4666-9890-e745ad8ebc54)

### Additional context

Currently, it is very difficult to work with the api, it is harder to check the documentIds but at the same time I caanot use the old "id" field to query an entry

surfjedi:
Yeah its redoculous we can't see documentId in the manager. I second this

verancegold:
Agreed. Would make my work a lot more streamlined!

TJBachorz:
This is making life difficult, a fix would be greatly appreciated 🙏

allison-hamelin:
I am seeing this issue as well.

Matb85:
Hi, any updates on this?

derrickmehaffy:
Related to TID 11591

justnils:
push! why don't they show the document id?!

PR #23604, opened 2025-05-27
fix(content-manager): documentId missing from fields for the view and…

Adzouz:
… filters

### What does it do?

Added documentId in the list of "Displayed fields" and filtering items in the listing view in the Content Manager.

Displayed fields:
Screenshot 2025-05-27 at 15 41 35

Filter:
Screenshot 2025-05-27 at 11 00 16

### Why is it needed?

Now the documentId is the new standard of defining unique identifiers to entities, it's required by the community to be able to display this information directly on the list view (as it was the case for the old id).

### How to test it?

Add / remove the field from the list view:
- Go to the Strapi admin interface
- Go to the Content Manager
- Select a collection
- Click on the little cog icon on the right (next to the language selector)

then:

1. First way to test it:
  - The documentId should be proposed in the list of displayed fields
  - Check and uncheck the option to see it appearing and disappearing from the view

2. Another way to test it:
  - Click on Configure the view
  - In the View section, if the documentId is not listed, click on the + button to add it (it should be in the list of available options)

Filter the results by documentId:
- Go to the Strapi admin interface
- Go to the Content Manager
- Select a collection
- Click on the Filters button on the left (next to the magnifying glass icon)
- Select documentId in the first field
- Select the condition
- Enter a value in the third field (matching the query you want to perform)
- Click on + Add filter button and see the list of results updated with the query

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/22053 raised by [Matb85](https://github.com/Matb85) (thanks a lot ✨)
🚀

### DIFF
--- a/tests/e2e/tests/content-manager/listview.spec.ts
+++ b/tests/e2e/tests/content-manager/listview.spec.ts
@@ -9,11 +9,25 @@ test.describe('List View', () => {
     await login({ page });
   });
 
+  test('A user can filter entries', async ({ page }) => {
+    await page.getByRole('link', { name: 'Content Manager' }).click();
+    await page.getByRole('link', { name: 'Article' }).click();
+
+    await page.getByRole('button', { name: 'Filters' }).click();
+    await page.getByRole('combobox', { name: 'Select field' }).click();
+    await page.getByRole('option', { name: 'documentId' }).click();
+    // va0x2nt206hluydibmsoiquc => documentId for article "Why I prefer football over soccer"
+    await page.getByRole('textbox', { name: 'documentId' }).fill('va0x2nt206hluydibmsoiquc');
+    await page.getByRole('button', { name: 'Add filter' }).click();
+    await expect(page.getByText('documentId is va0x2nt206hluydibmsoiquc')).toBeVisible();
+    // There should be 2 rows, 1 for the header and 1 for the Article entry
+    await expect(page.getByRole('row')).toHaveCount(2);
+  });
+
   test('A user should be able to navigate to the ListView of the content manager and see some entries', async ({
     page,
   }) => {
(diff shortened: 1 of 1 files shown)


markkaylor (commented):
Great, looks good to me. I tested it. We can take a look at the tests together tomorrow if you want.

packages/core/content-manager/server/src/services/utils/configuration/metadatas.ts
```diff
@@ -103,11 +103,6 @@ async function syncMetadatas(configuration: any, schema: any) {
     const attr = schema.attributes[key];
 
     const updatedMeta = { edit, list };
-    if (key === 'documentId') {
-      _.set(updatedMeta, ['edit'], {});
-      _.set(acc, [key], updatedMeta);
-      return acc;
-    }
```
markkaylor: ℹ️ Locally this change had the tests passing, but we shouldn't need this anyway since we set it in the `createDefaultMetadatas` above.  Which we merge with updatedMetas.

innerdvations (commented):
I didn't test this, but as far as the backend change is concerned it makes sense to me!

Could you add an e2e test for this? If you're lucky you might just be able to add a few lines to `content-manager/listview.spec.ts`

markkaylor:
> I didn't test this, but as far as the backend change is concerned it makes sense to me!
> 
> Could you add an e2e test for this? If you're lucky you might just be able to add a few lines to `content-manager/listview.spec.ts`

Sounds good thanks giving it a review! I'll look into adding an e2e test

markkaylor (approved):
I added the test @innerdvations asked for. I think we are good to merge this one now!

Adzouz:
Woot woot 🚀

PR #23609, opened 2025-05-27
chore: update OpenAPI types to OpenAPIV3_1 across the codebase

Convly:
### What does it do?

Update the codebase usage of `openapi-types` to target 3_1 instead of 3.

Note: A single instead of `OpenAPIV3` has been kept because `OpenAPIV3_1` doesn't export the needed enum.

See https://github.com/kogosoftwarellc/open-api/blob/main/packages/openapi-types/index.ts#L99 for more info

### Why is it needed?

Make sure the specification is compliant with OpenAPI 3.1 since this is the version we're targeting

### How to test it?

- `yarn setup`
- `cd examples/getstarted`
- `yarn strapi openapi:generate`


innerdvations:
> > > Note: A single instead of OpenAPIV3 has been kept because OpenAPIV3_1 doesn't export the needed enum.
> > 
> > 
> > Can you add a comment whever the v3 is used so that we know that it was kept on purpose and the reason?
> 
> Already there [#23609 (files)](https://github.com/strapi/strapi/pull/23609/files#diff-69de2b3f6c51f4304203022804e7c1e26fa40083497f737954c5e9abc89d44eeR65) Or maybe something more specific?

Nope, I just missed it, sorry! :)

PR #23612, opened 2025-05-27
Fix types of `env` helper

### LINKED ISSUE #23611: [bug]: env helper types incorrectly nullable even with default value provided
### Node Version

22.15.0

### NPM/Yarn/PNPM Version

PNPM 9.12.1

### Strapi Version

5.13.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

The type for the `env` helper utility (from `ConfigParams`) has it returning a nullable value even when a default is provided, which is not correct.

```ts
export type Env = typeof envFn & typeof utils;
declare function envFn<T>(key: string, defaultValue?: T): string | T | undefined;
declare const utils: {
    int(key: string, defaultValue?: number): number | undefined;
    float(key: string, defaultValue?: number): number | undefined;
    bool(key: string, defaultValue?: boolean): boolean | undefined;
    json(key: string, defaultValue?: object): any;
    array(key: string, defaultValue?: string[]): string[] | undefined;
    date(key: string, defaultValue?: Date): Date | undefined;
    /**
     * Gets a value from env that matches oneOf provided values
     * @param {string} key
     * @param {string[]} expectedValues
     * @param {string|undefined} defaultValue
     * @returns {string|undefined}
     */
    oneOf(key: string, expectedValues?: unknown[], defaultValue?: unknown): unknown;
};
declare const env: Env;
export default env;
//# sourceMappingURL=env-helper.d.ts.map
```

### Steps to Reproduce

- Edit any of the config files
- Import `ConfigParams` and set the parameters to that type.
- Use `env` to get an environment variable, passing a default value.
- See that the type is `string | undefined` (assuming a string value).

### Expected Behavior

Since a default was passed, I’d expect the return type to just be string (also applies to the typed helpers).

### Code Snippets

I’d expect to see some sort of function overloading, for example:

```ts
declare function envFn(key: string): string | undefined
declare function envFn(key: string, defaultValue: string): string
```

### Media

Image

### Confirmation Checklist

ericbf:
The env helper should not return a nullable value when a default is passed. This commit fixes all functions in the env helper to conform to the correct types.

### What does it do?

This fixes the typescript types of the `env` helper so that if you pass a default value, it does not return a nullable value.

### Why is it needed?

When using the env helper with a default value, you would have to cast it as non-nullable when using to a prop that was required, or you’d get a typescript error.

### How to test it?

Check the returned type of the `env` helper if passing a default. Functionality wasn’t changed, just types.

### Related issue(s)/PR(s)

Fixes #23611


packages/core/utils/src/env-helper.ts
```diff
@@ -2,98 +2,127 @@ import _ from 'lodash';
...
+function int(key: string, defaultValue: number): number;
+function int(key: string, defaultValue?: number | null | undefined): number | undefined;
+function int(key: string, defaultValue?: number | null | undefined): number | undefined {
+  if (!_.has(process.env, key)) {
+    return defaultValue ?? undefined;
+  }
 
-    return parseInt(getKey(key), 10);
-  },
+  return parseInt(getKey(key), 10);
```
unrevised6419: Thought: Maybe out of scope for this PR, but `parseInt` can return `NaN`

```suggestion
  const maybeNumber = parseInt(getKey(key), 10);
  return Number.isNaN(maybeNumber) ? undefined : maybeNumber;
```

packages/core/utils/src/env-helper.ts
```diff
@@ -2,98 +2,127 @@ import _ from 'lodash';
...
+function float(key: string, defaultValue: number): number;
+function float(key: string, defaultValue?: number | null | undefined): number | undefined;
+function float(key: string, defaultValue?: number | null | undefined): number | undefined {
+  if (!_.has(process.env, key)) {
+    return defaultValue ?? undefined;
+  }
 
-    return parseFloat(getKey(key));
-  },
+  return parseFloat(getKey(key));
```
unrevised6419: Note: Similar to `int` comment

packages/core/utils/src/env-helper.ts
```diff
@@ -2,98 +2,127 @@ import _ from 'lodash';
...
-   * @param {string} key
-   * @param {string[]} expectedValues
-   * @param {string|undefined} defaultValue
-   * @returns {string|undefined}
-   */
-  oneOf(key: string, expectedValues?: unknown[], defaultValue?: unknown) {
-    if (!expectedValues) {
-      throw new Error(`env.oneOf requires expectedValues`);
-    }
+  return new Date(getKey(key));
```
unrevised6419: Note: Similar to `int` comment. Date might be `Invalid Date`

packages/core/utils/src/env-helper.ts
ericbf: I’m fine with that. I’ll update the PR title and description accordingly.

ericbf: What are your thoughts then, if a default is passed and the resolved value is invalid, should we return the default or throw an error? The point was that if a default is passed, you won’t ever get undefined back. I am leaning towards throwing an error, as we shouldn’t invisibly ignore invalid values, especially if a default is passed.

ericbf: Since `env.json` already throws on invalid value, I’ll just do that here too.

packages/core/utils/src/env-helper.ts
```diff
@@ -2,98 +2,147 @@ import _ from 'lodash';
...
 
-    return parseInt(getKey(key), 10);
-  },
+  const maybeInt = parseInt(getKey(key), 10);
 
-  float(key: string, defaultValue?: number): number | undefined {
-    if (!_.has(process.env, key)) {
-      return defaultValue;
-    }
+  if (Number.isFinite(maybeInt)) {
```
unrevised6419: Suggestion: `Number.isNaN` should be enough. `Infinity/-Infinity` are impossible in this case, because `parseInt` cannot return such values (same applies to `date` util)

ericbf: I just prefer the affirmative `Number.isFinite` versus `!Number.isNaN` haha. It also adds a bit more robustness, _just in case_ future versions of `parseInt`/`parseFloat` support `Infinity`/`-Infinity`. It’s not a deal breaker for me, so I’m good either way. Would you like me to switch it out?

ericbf: `parseFloat` does support `Infinity`. I suppose, should that value be treated as valid? This may be a more substantive question than I had thought.

ericbf: Ok, resolved to only check `isNaN`, throw only if not explicitly set to `"NaN"` in `int` and `float`.

Bassel17:
Hey @ericbf  Thanks for this PR - the type improvements and validation additions are definitely valuable improvements to the `env` helper. 👍

I wanted to discuss two things about this PR:

This PR introduces **breaking changes** to the runtime behavior. Specifically, `env.int()`, `env.float()`, and `env.date()` will now throw errors for invalid values instead of returning `NaN` or `Invalid Date`. 

**Before:**
```typescript
process.env.INVALID = '';
const value = env.int('INVALID'); // Returns NaN
```

**After:**
```typescript
process.env.INVALID = '';
const value = env.int('INVALID'); // Throws: Error: env.int encountered an invalid value for INVALID
```

While I agree this is the correct behavior (silently returning `NaN` can lead to hard-to-debug issues), any existing code that was handling these invalid returns will now break with exceptions. This should be:
1. Clearly documented as a breaking change
2. Included in a major version release
3. Added to the migration guide


I think it would be beneficial to split this PR into two separate PRs, one with the Type improvements and one with the validation changes (we can keep that to track it later).
but a breaking change is something we can't merge immediately, the type fixes however is something we can merge

ericbf:
> Hey @ericbf Thanks for this PR - the type improvements and validation additions are definitely valuable improvements to the `env` helper. 👍
> 
> I wanted to discuss two things about this PR:
> 
> This PR introduces **breaking changes** to the runtime behavior. Specifically, `env.int()`, `env.float()`, and `env.date()` will now throw errors for invalid values instead of returning `NaN` or `Invalid Date`.
> 
> **Before:**
> 
> ```ts
> process.env.INVALID = '';
> const value = env.int('INVALID'); // Returns NaN
> ```
> 
> **After:**
> 
> ```ts
> process.env.INVALID = '';
> const value = env.int('INVALID'); // Throws: Error: env.int encountered an invalid value for INVALID
> ```
> 
> While I agree this is the correct behavior (silently returning `NaN` can lead to hard-to-debug issues), any existing code that was handling these invalid returns will now break with exceptions. This should be:
> 
> 1. Clearly documented as a breaking change
> 2. Included in a major version release
> 3. Added to the migration guide
> 
> I think it would be beneficial to split this PR into two separate PRs, one with the Type improvements and one with the validation changes (we can keep that to track it later). but a breaking change is something we can't merge immediately, the type fixes however is something we can merge

That was always my intention but others requested additional changes, unfortunately. Idc as long as it gets fixed.

ericbf:
@Bassel17 I’ve reduced the scope of this PR to just the type fixes. Can you re-open? Thanks! Or let me know and I can open a new PR with the same changes.

ericbf:
I’ve also created a new PR with the same changes (#24881). We can close that one if this one gets re-opened. But I don’t want this to get missed as we are narrowing the scope.

PR #23616, opened 2025-05-28
feat(content-manager): conditional fields

Bassel17:
### What does it do?

Adds conditional field logic to the Content Manager

### How to test it?

Add the following conditions attribute to in any fields schema as such:

```json
"code": {
      "type": "string",
      "required":true,
      "maxLength": 3,
      "unique": true,
      "minLength": 2,
      "pluginOptions": {
        "i18n": {
          "localized": true
        }
      },
      "conditions": {
        "visible": {
          "==": [
            {
              "var": "visible"
            },
            true
          ]
        }
      }
    },
"visible": {
      "type": "boolean"
    }
```

the condition should target a specific field in the schema (checkbox or enumeration fields)

example behavior: 

image

image

#### Conditional Field Visibility

- [x] Given a boolean field **X**, when **X = true**, the conditional field **Y** becomes **visible** in the content editor.
- [x] When **X = false**, the conditional field **Y** becomes **hidden**.
- [x] The visibility of field **Y** updates **immediately** when field **X** is toggled — without page reloads or form refresh.
- [x] If field **Y** is marked as `required`, then the required validation is **only enforced when Y is visible**.
- [x] If field **Y** is hidden, it should **not be submitted or validated** as part of the form.
- [x] If the user **lacks permission** for field **Y**, it should remain **hidden**, regardless of conditional logic.
- [x] RBAC rules **override conditional visibility** — field **Y** should never appear or be editable if the role disallows it.

---

#### Version History Navigation

- [x] Given a content type with conditional fields, when navigating through previous versions, the UI should reflect **visibility and value state** at the time of each version.
- [x] When navigating to an earlier version where a conditionally visible field was shown (e.g., **Type = Physical Product** → shows **Shipping Weight**), both the **field and its value should be displayed**.
- [x] When navigating to a later version where that field is hidden (e.g., **Type = Digital Product**), the UI should **hide the field**, even if it had a value in a prior version.
- [x] The version view should render the conditionally visible fields **as they were** at the time of that version.

✅ *Example:*
- A writer selects **Type = Physical Product**, fills in **Shipping Weight**, and saves → *Version 1*
- Later, changes **Type = Digital Product**, which hides **Shipping Weight**, and saves → *Version 2*
- When viewing **Version 1**, the UI shows:
  - **Type = Physical Product**
  - **Shipping Weight = previously entered value**
- When viewing **Version 2**, the UI shows:
  - **Type = Digital Product**
  - **Shipping Weight = hidden**

---

#### Version Restoration Behavior

- [x] When restoring a version where a conditional field was **visible**, the field should become **visible again** and restore its value.
- [x] When restoring a version where a conditional field was **hidden**, the field should be **hidden again** and its previous value should be discarded or suppressed.
- [x] Restoration should update both the **conditional logic** and the **field values**, reflecting exactly what was saved in the restored version.
- [x] Restoration must not rely on current conditional logic state — it must reproduce the **state as saved** in the chosen version.

✅ *Example:*
- Version 1: **Type = Physical Product**, **Shipping Weight = 3kg**
- Version 2: **Type = Digital Product**, **Shipping Weight = hidden**
- On restoring **Version 1**, the UI should:
  - Set **Type = Physical Product**
  - Show and prefill **Shipping Weight = 3kg**

---


derrickmehaffy:
> If field Y is marked as required, then the required validation is only enforced when Y is visible.

If conditional fields MVP isn't going to wrap into the content API how are we going to manage this from the API side?

innerdvations (commented):
Looks good so far, and working as expected (with the exception of the remaining items you still have unchecked in TODO)!

packages/core/content-manager/admin/src/history/components/VersionInputRenderer.tsx
```diff
@@ -345,6 +347,17 @@ const VersionInputRenderer = ({
     edit: { components: componentsLayout },
   } = useDocLayout();
 
+  const rulesEngine = createRulesEngine();
+  const attribute = props.attribute;
+
+  if (attribute?.conditions && attribute?.conditions.visible) {
+    const isVisible = rulesEngine.evaluate(attribute?.conditions.visible, fieldValues);
```
jhoward1994, line 354: A little confusing to read with `isVisible` here and `visible` below. Maybe some better variable names would make it clearer that the field is hidden because of a condition?

Bassel17:
> If conditional fields MVP isn't going to wrap into the content API how are we going to manage this from the API side?

it needed is a problem, I am in the process of trying to currently get this behavior to work and make sense

packages/core/content-manager/admin/src/utils/validation.ts
```diff
@@ -22,6 +22,7 @@ type AnySchema =
 
 interface ValidationOptions {
   status: 'draft' | 'published' | null;
+  removedAttributes?: string[];
```
innerdvations, line 25:
```suggestion
  ignoreAttributes?: string[];
```
Maybe more accurate?

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -132,12 +133,22 @@ const EditViewPage = () => {
         initialValues={initialValues}
         method={isCreatingDocument ? 'POST' : 'PUT'}
         validate={(values: Record<string, unknown>, options: Record<string, string>) => {
+          // removes hidden fields from the validation
+          // this is necessary because the yup schema doesn't know about the visibility conditions
```
innerdvations, line 148: Why is this preferred over giving the yup schema awareness of conditions?

tests/e2e/tests/content-manager/conditional-fields.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+test.describe('Conditional Fields', () => {
+  test.beforeEach(async ({ page }) => {
+    await resetDatabaseAndImportDataFromPath('with-admin.tar');
+    await page.goto('/admin');
+    await login({ page });
+
+    await navToHeader(page, ['Content Manager'], 'Content Manager');
+  });
+
+  test('As a user if I toggle a boolean field that affects a conditional field, the field should be hidden and the value should not be filled', async ({
```
innerdvations, line 16: Can you please add a test case for the actual submission of the form, and one for the "required" fields too?

innerdvations: Also, if you're able to put it all in a `test.each` to try out all the other types that would be great, but maybe not strictly necessary since they all evaluate the exact same way

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
Bassel17: we are essentially reusing a method that we use to also remove this data from being submitted, and it also made sense considering we are acting as if the field doesn't exist meaning it doesn't need to be validated or submitted for anything.

innerdvations:
@Bassel17 should be fine now, I reverted the accidental beta commit

ChristopheCVB:
Hi there!

Super excited to see this feature is coming soon.

I have a question related to Custom Fields registered by plugins and how could we (as plugin developers), expose that configuration of Conditional Fields to the developers using our Custom Fields.
Right now, as a developer I can edit the schema and add the conditions myself and everything works like a charm. Is there a plan to be able to configure the conditions through the CTB for Custom Fields?

Again, amazing work, can't wait for the release, keep pushing!

Thank you!

maccomaccomaccomacco:
> Hi there!
> 
> Super excited to see this feature is coming soon.
> 
> I have a question related to Custom Fields registered by plugins and how could we (as plugin developers), expose that configuration of Conditional Fields to the developers using our Custom Fields. Right now, as a developer I can edit the schema and add the conditions myself and everything works like a charm. Is there a plan to be able to configure the conditions through the CTB for Custom Fields?
> 
> Again, amazing work, can't wait for the release, keep pushing!
> 
> Thank you!

It's one of the very next enhancements.

We wanted to ship it fast, so conditional fields have a few rules and act on a few fields' values. Over time, we aim to expand the feature's capabilities, and we plan to support conditions on custom fields.

I can't give an exact timeline, but it could even come in July.

innerdvations (dismissed):
Did another review and QA this morning; works great, let's go! 🚀

innerdvations (approved):

Issue #20754, opened 2024-07-15
Release settings weird ux flow


yanniskadiri:

Adzouz:
Hi @yanniskadiri, thanks a lot for reporting the issue 🙏
I pushed the fix and it will be part of version 5.15.1 🚀

PR #23629, opened 2025-05-30
fix: default timezone releases issue

Adzouz:
### What does it do?

Fix the default timezone message when a wrong value is saved in the Releases settings page.

### Why is it needed?

It was displaying a success message even if the value wasn't correct. The value wasn't stored but the message was saying that the default timezone was successfully updated.

Before:
https://github.com/user-attachments/assets/3b94bbe8-c290-40e1-a64d-a72b59446b72

After:
https://github.com/user-attachments/assets/622134ba-7024-4f4e-b0a4-b445838affca

### How to test it?

- In the admin interface, go to Settings > Releases
- Try to put an incorrect value in the text combobox field
- Save
- You should see an error message

### Related issue(s)/PR(s)

Resolves [#20754](https://github.com/strapi/strapi/issues/20754)

### DIFF
--- a/packages/core/admin/admin/src/translations/en.json
+++ b/packages/core/admin/admin/src/translations/en.json
@@ -618,7 +618,7 @@
   "components.Input.error.contentTypeName.taken": "This name already exists",
   "components.Input.error.custom-error": "{errorMessage} ",
   "components.Input.error.password.noMatch": "Passwords do not match",
-  "components.Input.error.validation.combobox.invalid": "The value provided is not valid.",
+  "components.Input.error.validation.combobox.invalid": "The value provided is not valid",
   "components.Input.error.validation.email": "This is not a valid email",
   "components.Input.error.validation.json": "This doesn't match the JSON format",
   "components.Input.error.validation.lowercase": "The value must be a lowercase string",
--- a/packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
+++ b/packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
@@ -54,19 +54,15 @@ const ReleasesSettingsPage = () => {
     { setErrors }: FormHelpers<UpdateDefaultTimezone>
   ) => {
     const { defaultTimezone } = body;
-    const formattedDefaultTimezone = defaultTimezone?.replace(/ /, '&');
+    const formattedDefaultTimezone = defaultTimezone;
     const isBodyTimezoneValid = timezoneList.some(
       (timezone) => timezone.value === formattedDefaultTimezone
     );
-    const newBody =
-      !defaultTimezone || !isBodyTimezoneValid
-        ? { defaultTimezone: null }
-        : { defaultTimezone: formattedDefaultTimezone };
 
     if (!isBodyTimezoneValid && defaultTimezone) {
       const errorMessage = formatMessage({
         id: 'components.Input.error.validation.combobox.invalid',
-        defaultMessage: 'The value provided is not valid.',
+        defaultMessage: 'The value provided is not valid',
       });
       setErrors({
         defaultTimezone: errorMessage,
@@ -77,6 +73,11 @@ const ReleasesSettingsPage = () => {
       });
       return;
     }
+
+    const newBody =
+      !defaultTimezone || !isBodyTimezoneValid
+        ? { defaultTimezone: null }
+        : { defaultTimezone: formattedDefaultTimezone };
 
     try {
       const response = await updateReleaseSettings(newBody);
@@ -235,6 +236,11 @@ const TimezoneDropdown = () => {
       <Combobox
         autocomplete={{ type: 'list', filter: 'contains' }}
         onTextValueChange={(value) => field.onChange('defaultTimezone', value)}
+        onChange={(value) => {
+          if ((field.value && value) || !field.value) {
+            field.onChange('defaultTimezone', value);
+          }
+        }}
         onClear={() => field.onChange('defaultTimezone', '')}
         value={field.value}
         disabled={!canUpdate}
--- a/packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
+++ b/packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
@@ -208,7 +208,7 @@ describe('Releases Settings page', () => {
     fireEvent.click(saveButton);
 
     // Wait for the error message
-    const errorMessage = await screen.findAllByText('The value provided is not valid.');
+    const errorMessage = await screen.findAllByText('The value provided is not valid');
     expect(errorMessage).toHaveLength(2);
   });
 });


packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
```diff
@@ -103,4 +133,82 @@ describe('Releases Settings page', () => {
...
+
+    // Wait for the combobox to be rendered
+    const combobox = await screen.findByRole('combobox', { name: 'Default timezone' });
+    await user.click(combobox);
+    await user.type(combobox, 'Invalid timezone');
+    await user.keyboard('{Enter}');
+
+    // Try to save the value
+    const saveButton = await screen.findByText('Save');
+    fireEvent.click(saveButton);
```
markkaylor, line 208: How come we need to use fireEvent instead user.click?

packages/core/content-releases/admin/src/translations/en.json
```diff
@@ -13,7 +13,6 @@
   "content-manager-edit-view.scheduled.date": "{date} at {time} ({offset})",
   "content-manager-edit-view.edit-release": "Edit release",
   "content-releases.content-manager-edit-view.edit-entry": "Edit entry",
-  "content-releases.pages.Releases.title": "Releases",
```
markkaylor, line 16: Isn't this still used did you mean to delete it?

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
```diff
@@ -43,11 +48,35 @@ const ReleasesSettingsPage = () => {
...
-    const isBodyTimezoneValid = timezoneList.some((timezone) => timezone.value === defaultTimezone);
+    const formattedDefaultTimezone = defaultTimezone?.replace(/ /, '&');
+    const isBodyTimezoneValid = timezoneList.some(
+      (timezone) => timezone.value === formattedDefaultTimezone
+    );
     const newBody =
-      !defaultTimezone || !isBodyTimezoneValid ? { defaultTimezone: null } : { ...body };
+      !defaultTimezone || !isBodyTimezoneValid
+        ? { defaultTimezone: null }
+        : { defaultTimezone: formattedDefaultTimezone };
```
markkaylor: total nitpick but I'd put this closer to where it is used when passed to `updateReleaseSettings` on line 81

packages/core/admin/admin/src/translations/en.json
```diff
@@ -600,6 +600,7 @@
   "components.Input.error.contentTypeName.taken": "This name already exists",
   "components.Input.error.custom-error": "{errorMessage} ",
   "components.Input.error.password.noMatch": "Passwords do not match",
+  "components.Input.error.validation.combobox.invalid": "The value provided is not valid.",
```
markkaylor: Is the fullstop `.` in the designs? Personally I don't think we should have it

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
```diff
@@ -43,11 +48,35 @@ const ReleasesSettingsPage = () => {
...
   const { timezoneList } = getTimezones(new Date());
 
-  const handleSubmit = async (body: UpdateSettings.Request['body']) => {
+  const handleSubmit = async (
+    body: UpdateSettings.Request['body'],
+    { setErrors }: FormHelpers<UpdateDefaultTimezone>
+  ) => {
     const { defaultTimezone } = body;
-    const isBodyTimezoneValid = timezoneList.some((timezone) => timezone.value === defaultTimezone);
+    const formattedDefaultTimezone = defaultTimezone?.replace(/ /, '&');
```
markkaylor:
```suggestion
    // Replace first space with '&'
    const formattedDefaultTimezone = defaultTimezone?.replace(/ /, '&');
```

markkaylor: I find this a bit confusing since it looks like in the ComboxboxOption we do the exact opposite. Why don't we get that value passed to ComboxboxOption, wouldn't that be what we wanted without any formatting here?

packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
Adzouz: Using the user.click wasn't triggering the handleSubmit for some reason. We can review that together if you want, maybe I misunderstood something.

packages/core/content-releases/admin/src/translations/en.json
Adzouz: This was a duplicate, the value already exists in the file.

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
Adzouz: Well this one was tricky because the value provided by the api is not really readable by the user because there's a & so that's why we were formatting visually by replacing & with a space. But by doing that, the value sent is also changed (again I don't know exactly why) so we need to put back the & to store it correctly in the request.

markkaylor: But shouldn't `value=timezone.value` here have the value we want to send to the server? I'm just confused why we aren't receiving that value instead of what we are displaying to the user. 

Screenshot 2025-06-02 at 11 41 07

Adzouz: I'm also confused because even with that it's sending what's inside the ComboboxOption and not the value provided as a parameter...

packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
```diff
@@ -1,10 +1,28 @@
 import { useRBAC } from '@strapi/admin/strapi-admin';
-import { render, server, screen } from '@tests/utils';
+import { render, server, screen, waitFor, fireEvent } from '@tests/utils';
```
remidej, line 2: userEvent is the recommended API instead of fireEvent: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

Adzouz: I know, but unfortunately, as said in the comment of Mark, using user.click doesn't trigger the handleSubmit

packages/core/content-releases/admin/src/pages/tests/ReleasesSettingsPage.test.tsx
remidej: No big deal to leave it as is of course. But you you try going into `packages/core/admin/admin/tests/utils.tsx` and moving the user event stuff above the `...renderRTL`? Because packages/core/admin/admin/tests/utils.tsx the user object should be defined before rendering

Adzouz: I tried but it's not helping unfortunately 😢 We'll keep fireEvent for the moment to not waste too much time

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
Adzouz: Done!

packages/core/admin/admin/src/translations/en.json
Adzouz: I removed the . but we'll open a ticket to update all sentences for consistency

packages/core/content-releases/admin/src/pages/ReleasesSettingsPage.tsx
Adzouz: I found a little better way to do it. I removed the formatting in handleSubmit and put back the onChange just to update the value is case we select a value or clear the field.

remidej (approved):
🚀

PR #23630, opened 2025-05-30
fix: add the disabled state of the creatable button with no permissions

simotae14:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
@@ -6,6 +6,7 @@ import {
   useForm,
   useNotification,
   useFocusInputField,
+  useRBAC,
 } from '@strapi/admin/strapi-admin';
 import {
   Box,
@@ -21,8 +22,10 @@ import {
   Field,
   FlexComponent,
   BoxComponent,
+  Loader,
+  EmptyStateLayout,
 } from '@strapi/design-system';
-import { Cross, Drag, ArrowClockwise, Link as LinkIcon, Plus } from '@strapi/icons';
+import { Cross, Drag, ArrowClockwise, Link as LinkIcon, Plus, WarningCircle } from '@strapi/icons';
 import { generateNKeysBetween } from 'fractional-indexing';
 import pipe from 'lodash/fp/pipe';
 import { getEmptyImage } from 'react-dnd-html5-backend';
@@ -33,6 +36,8 @@ import { styled } from 'styled-components';
 import { RelationDragPreviewProps } from '../../../../../components/DragPreviews/RelationDragPreview';
 import { COLLECTION_TYPES } from '../../../../../constants/collections';
 import { ItemTypes } from '../../../../../constants/dragAndDrop';
+import { PERMISSIONS } from '../../../../../constants/plugin';
+import { DocumentRBAC, useDocumentRBAC } from '../../../../../features/DocumentRBAC';
 import { useDebounce } from '../../../../../hooks/useDebounce';
 import { useDocument } from '../../../../../hooks/useDocument';
 import { type DocumentMeta, useDocumentContext } from '../../../../../hooks/useDocumentContext';
@@ -54,6 +59,7 @@ import { DocumentStatus } from '../../DocumentStatus';
 import { useComponent } from '../ComponentContext';
 import { RelationModalRenderer, getCollectionType } from '../Relations/RelationModal';
 
+import type { FindAvailable } from '../../../../../../../shared/contracts/relations';
 import type { Schema } from '@strapi/types';
 
 /**
@@ -483,7 +489,6 @@ const RelationsInput = ({
   isRelatedToCurrentDocument,
   ...props
 }: RelationsInputProps) => {
-  const [textValue, setTextValue] = React.useState<string | undefined>('');
   const [searchParams, setSearchParams] = React.useState({
     _q: '',
     page: 1,
@@ -491,7 +496,7 @@ const RelationsInput = ({
   const { toggleNotification } = useNotification();
   const { currentDocumentMeta } = useDocumentContext('RelationsInput');
   const { formatMessage } = useIntl();
-  const fieldRef = useFocusInputField<HTMLInputElement>(name);
+
   const field = useField<RelationsFormValue>(name);
 
   const searchParamsDebounced = useDebounce(searchParams, 300);
@@ -539,10 +544,6 @@ const RelationsInput = ({
     currentDocumentMeta.params,
   ]);
 
-  const handleSearch = async (search: string) => {
-    setSearchParams((s) => ({ ...s, _q: search, page: 1 }));
-  };
-
   const hasNextPage = data?.pagination ? data.pagination.page < data.pagination.pageCount : false;
 
   const options = data?.results ?? [];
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -601,25 +590,137 @@ const RelationsInput = ({
...
+          {...props}
+        />
+      </DocumentRBAC>
+      <Field.Error />
+      <Field.Hint />
+    </Field.Root>
+  );
+};
+
+interface RelationModalWithContextProps {
```
remidej: maybe you can extend RelationInputProps?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -601,25 +590,137 @@ const RelationsInput = ({
...
+  if (isLoadingPermissions) {
+    return (
+      <Loader small>
+        {formatMessage({
+          id: 'content-manager.ListViewTable.relation-loading',
+          defaultMessage: 'Relations are loading',
+        })}
+      </Loader>
+    );
+  }
```
remidej: do you think you could move the loading state inside the combobox? Otherwise I would assume this creates a layout shift when permissions load

Adzouz (approved):
Good job, it works well 🚀
Nice testing too 👌

tests/e2e/tests/content-manager/relations-on-the-fly/edit-relation.spec.ts
```diff
@@ -8,7 +8,7 @@ const AUTHOR_EDIT_URL =
 
 test.describe('Relations on the fly - Edit a Relation', () => {
   test.beforeEach(async ({ page }) => {
-    await resetDatabaseAndImportDataFromPath('with-admin.tar');
+    await resetDatabaseAndImportDataFromPath('rotf.tar');
```
markkaylor: Is this intentional or was this for testing? Idk if we want to introduce a new dataset

simotae14: It is intentional because I didn't wanted to broke the other test cases, basically I changed the permission for the Editor regarding a collection, if you think it is not the case I need to do it in the old with-admin tar file and then try to fix the rbac failing tests if needed

simotae14: Btw we did something similar in the brach with the e2e tests regarding relations, I generated a new tar just for the purpose of these use cases

simotae14: I changed the with-admin dataset, I added a new Author user with the permission I needed to check the creatable disabled button, let me know if it is ok now, thanks

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/Relations.test.tsx
```diff
@@ -55,13 +57,31 @@ describe('Relations', () => {
...
 
+    // Wait for the list items to be rendered
+    await waitFor(() => {
+      expect(screen.getAllByRole('listitem')).toHaveLength(3);
+    });
+
+    // Wait for the combobox to be updated with the count
+    await waitFor(() => {
+      expect(screen.getByLabelText(/relations \(3\)/)).toBeInTheDocument();
+    });
```
markkaylor: Can we not use finBy queries instead of waitFor?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -602,25 +591,127 @@ const RelationsInput = ({
...
+    } else if (data.pagination.page < data.pagination.pageCount) {
+      setSearchParams((s) => ({ ...s, page: s.page + 1 }));
+    }
+  };
+
+  const options = data?.results ?? [];
+
+  React.useLayoutEffect(() => {
+    setTextValue('');
+  }, [fieldValue]);
```
markkaylor, line 695: I know this was already in place but I don't understand what it is doing or why it is necessary?

Louis-strap: Hello, regarding the disabled state UI, what do you find weird?

markkaylor:
> Hello, regarding the disabled state UI, what do you find weird?

@Louis-strap I thought this border radius looked strange:

Screenshot 2025-06-11 at 14 24 00

Also surprised to see the hover state without a border radius

Screenshot 2025-06-11 at 14 42 57

But if all that is cool with you then works for me

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
simotae14: I suppose it was added by Rémi

simotae14: basically it reset the value of textValue whenever fieldValue changes, basically I suppose it clears the search input when a relation is selected

Louis-strap:
I agree that the top left and right corner could have no border radius, however I can't find other examples where we handle this situation differently so I suggest we leave it like this, WDYT?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/Relations.test.tsx
simotae14: Not for every single case, the first waitFor is checking for the absence of an element (not.toBeInTheDocument()), which findBy can't do. I replaced, where possible, waitFor with findBy

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/Relations.test.tsx
```diff
@@ -92,56 +122,84 @@ describe('Relations', () => {
   it.todo('should disconnect a relation');
 
   describe.skip('Accessibility', () => {
-    it('should have have description text', () => {
-      const { getByText } = render();
+    it('should have have description text', async () => {
```
markkaylor, line 125: This describe block appears to be skipped but I see you still updated the tests. Were you running these locally? Should we turn them on?

simotae14: yeah my bad, I was fighting with e2e tests because I modified the tar file in two PRs, one I have already merged and this one

simotae14: I was wrong, these tests are already skipped in develop https://github.com/strapi/strapi/blob/89c5758919abbc4571b0b24bc64a60f6b92aa12b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/tests/Relations.test.tsx#L124 I can try to fix them but, to be honest, I think this PR needs to be reviewed for its scope, because it is taking too much time, if you are ok I can add a jira task to work on reactivate the accessibility tests in the future, wdyt?

markkaylor: I don't think this PR needs to turn these tests on but I find it weird that we are updating skipped tests

PR #23634, opened 2025-05-30
fix: header behind other elements

### LINKED ISSUE #22208: Z-index issue in settings/accordion

- Node.js version: 20
- Strapi version: 5.4.0
- Database: N/A
- Operating system: Mac OS
- Is your project Javascript or Typescript: JS/TS

### Describe the bug
There is a z-index issue with the header and an open accordion, you can easily see this when using the user/permissions plugin. 

1. Start a new Strapi project
2. Navigate to any settings section that has an accordion
3. Scroll up

### Expected behavior

The header is always above the content of the page

### Screenshots

image

image

Adzouz:
### What does it do?

It fixes a z-index issue showing the content of an accordion above the sticky header in the admin interface.

image
image

_ℹ️ I picked 3 as zIndex because I saw that the modal of the guided tour starts at 4 and the accordions are 1, it gives a tiny flexibility to have other elements at 2 while keeping the sticky header above._

### Why is it needed?

To improve the user experience on the site and improve readability when displaying accordions in the admin.

### How to test it?

- Go to the admin interface
- Go to Settings > Roles
- Click on any role
- Scroll to an accordion and expand it
- The header appear behind the content of the accordion

### Related issue(s)/PR(s)

Resolves #22208 

🚀

### DIFF
--- a/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
+++ b/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
@@ -37,7 +37,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
           background="neutral0"
           shadow="tableShadow"
           width={`${width}px`}
-          zIndex={1}
+          zIndex={3}
           data-strapi-header-sticky
         >
           <Flex justifyContent="space-between">


packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
```diff
@@ -37,7 +37,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
           background="neutral0"
           shadow="tableShadow"
           width={`${width}px`}
-          zIndex={1}
+          zIndex={3}
```
markkaylor, line 40: Wouldn't we want a lower zIndex so that the content remains behind the sticky header?

Adzouz: That's the sticky header, this one should be above the rest of the content, that's what caused the current issue and needed to be bumped higher.

markkaylor: ahh gotcha my bad, 👍

markkaylor (approved):

markkaylor:
I tested it locally

PR #23648, opened 2025-06-03
fix(content-manager): ui issue when too many elements in dz in config…

### LINKED ISSUE #21056: [V5] UI Problem: Configure view screen doesn't show more components because it doesn't fit in the box

Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0-rc.11
Node/Yarn Version: npm/10.1.0 node/v20.9.0 darwin arm64 workspaces/false
Edition: Community
Database: postgres
### Describe the bug

On the configure view screen, when there are more components than what it fits on the screen, there is no vertical scroll. Which makes it impossible to navigate to the ones that are not visible on the list

_General note: There are many bugs with configure screen._

1. Create multiple components
2. Create a collection with a relation with multiple components
3. Go to Configure View screen
4. Narrow the width of your screen if you don't have enough components.
5. The components that doesn't fit on the view, won't be there and you cannot scroll vertically.

### Expected behavior

As a user, I need to be able to scroll vertically on the components relation so I can navigate to the all components.

### Screenshots

Screenshot 2024-08-22 at 17 26 45 copy

Adzouz:
…ure the view

### What does it do?

Fix the UI of the dynamic zone in configure the view to display all the elements correctly when there is a lot of them. Copying the same behaviour as the create / edit view of a content to fix the issue and placing elements on a new line.

Before:
image

After:
Screenshot 2025-06-03 at 09 19 11

### Why is it needed?

UI is broken when there are too many elements in the dynamic zone field in configure the view and the user can't see the rest of the elements + the elements are shrinking and the visual is not ideal.

### How to test it?

- Go to the content manager in the admin interface
- Go to a collection type or a single type that has a dynamic zone with a lot of children elements (if not, add them before)
- Create or edit an entry
- Click on the three dots to access the settings menu and select "Configure the view"

### Related issue(s)/PR(s)

Resolves #21056

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
+++ b/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
@@ -487,6 +487,7 @@ const Field = ({ attribute, components, name, index, onMoveField, onRemoveField 
               alignItems="flex-start"
               gap={2}
               width="100%"
+              wrap="wrap"
             >
               {attribute?.components.map((uid) => (
                 <ComponentLink


packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
```diff
@@ -487,6 +487,7 @@ const Field = ({ attribute, components, name, index, onMoveField, onRemoveField
               alignItems="flex-start"
               gap={2}
               width="100%"
+              wrap="wrap"
```
remidej, line 490: When the row wraps they aren't aligned in the center (they should likely occupy the full width). This is likely due to ComponentLink having an arbitrary value of 14 rems. Could you address this?

Adzouz: You want me to also update the create / edit view? Because it's the same behaviour

remidej: Probably then yes.

Just to clarify, I'm not saying if there's a single component it should be full width. Just that each component should be a clean fifth of the width so they form a clean grid when wrapping

Adzouz: Discussed together that it will be hard to do this with only CSS without causing blocks to have different sizes from a line to another (depending on how many items can fit on each line). Resolving the comment to move forward.

markkaylor (approved):

PR #23650, opened 2025-06-03
chore: update trial ended modal logic

mazzucchelli:
### What does it do?

Fix 2 bugs found during QA

- The trial ended dialog not appearing when trial ends -> the `FreeTrialEndedModal` component now use the same logic as the `UpsellBanner` component

- LocalStorage is not updated when `FreeTrialEndedModal` and `FreeTrialWelcomeModal` modals are closed by clicking the modal background layer -> logic updated for the `onOpenChange` prop


innerdvations (approved):
LGTM but I didn't test it; can test it in the release branch later

PR #23651, opened 2025-06-03
Feat/ctb-tracking

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/features/Tracking.tsx
+++ b/packages/core/admin/admin/src/features/Tracking.tsx
@@ -443,12 +443,6 @@ export interface UseTrackingReturn {
  * ```
  */
 
-const logger = (...args: unknown[]) => {
-  console.log('Tracking', ...args);
-};
-
-const logForEvents = ['didUpdateCTBSchema'];
-
 const useTracking = (): UseTrackingReturn => {
   const { uuid, telemetryProperties } = React.useContext(TrackingContext);
   const userId = useAppInfo('useTracking', (state) => state.userId);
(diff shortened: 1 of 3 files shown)


jhoward1994:
@maccomaccomaccomacco 
This differs slightly from the JIRA ticket. I didn't think it was necessary to include the property 


```
	Properties: successfully
	Values
	true if the save operation is done successfully
	false in case of errors when the content structure is saved.
```

As it seems implicit that it was successful if the event fires? Let me know if I'm missing something?

innerdvations:
@maccomaccomaccomacco Do we want to keep `didEditFieldNameOnContentType`? I notice it is still sent as soon as I edit the field, not when I save.

Also noticed a `didCreateGuidedTourCollectionType` but that only gets sent on save so I assume that's working as intended but just want to confirm.

innerdvations:
> @maccomaccomaccomacco This differs slightly from the JIRA ticket. I didn't think it was necessary to include the property
> 
> ```
> 	Properties: successfully
> 	Values
> 	true if the save operation is done successfully
> 	false in case of errors when the content structure is saved.
> ```
> 
> As it seems implicit that it was successful if the event fires? Let me know if I'm missing something?

Does it mean having a 'success' that is set to`true` at the end of the try and `false` at the end of the catch?

innerdvations (approved):
LGTM and seems to work, just waiting on the pending questions from @maccomaccomaccomacco before approving!

innerdvations (commented):
LGTM and seems to work, just waiting on the pending questions from @maccomaccomaccomacco before approving!

maccomaccomaccomacco:
> @maccomaccomaccomacco Do we want to keep `didEditFieldNameOnContentType`? I notice it is still sent as soon as I edit the field, not when I save.
> 
> Also noticed a `didCreateGuidedTourCollectionType` but that only gets sent on save so I assume that's working as intended but just want to confirm.

yep, it's fine to keep it

maccomaccomaccomacco:
> @maccomaccomaccomacco This differs slightly from the JIRA ticket. I didn't think it was necessary to include the property
> 
> ```
> 	Properties: successfully
> 	Values
> 	true if the save operation is done successfully
> 	false in case of errors when the content structure is saved.
> ```
> 
> As it seems implicit that it was successful if the event fires? Let me know if I'm missing something?

Same consideration as Ben's one for me

Issue #23358, opened 2025-04-13
Content Manager Media items per page and pagination broken


ockiphertweck:
### Node Version

v20.18.0

### NPM/Yarn/PNPM Version

10.9.0

### Strapi Version

5.12.4

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When using a media field in a collection type to select an image, the media selection modal opens, but the pagination controls and the items-per-page dropdown do not function correctly. This makes it impossible to browse through uploaded images beyond the first page or adjust how many media items are shown per view.

### Steps to Reproduce

1. Go to Content-Type Builder and create a collection with a Media field (set to "one image" or "multiple images").
2. Naviagte to the Media Library and Upload multiple images (i did it in a seperate folder)
3. Navigate to the collection in Content Manager and create or edit an entry.
4. Click the Add media button to select an image.
In the media library modal, try
  - Clicking the pagination buttons to go to the next page
  - Opening the items-per-page dropdown (e.g., 10, 20, 50)

### Expected Behavior

The pagination controls should allow navigating between pages, and the items-per-page dropdown should open and allow the user to select a new value.

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/b7e32b7e-d9ee-4657-9cb9-f324dc6f20bd

### Additional information

In the video you see that there is also a small horizontal scrolling

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

jsc-zyres:
We have the same issue and I can add that there is some React error in the console:

![Image](https://github.com/user-attachments/assets/dc1d249e-1495-4f01-b9fd-73f90183d39e)

This seems to break some z-index cause I can press a button again when I increase its index with my browser devtools

zovanni:
Same here. As a temporary workaround, you can use your tab (or shift + tab) on your keyboard to navigate trough those buttons, then use spacebar to click

genai-work:
Same for me

sayedbs:
I found a solution that resolves this issue by applying a z-index style. This ensures that the element is properly layered and no longer hidden behind other components.

Here’s a screenshot for reference:

[...]

PR #23655, opened 2025-06-03
fix(upload): switching tab when updating pagination

Adzouz:
### What does it do?

Fixes a tab switch triggered when changing page of number of items in the assets dialog.
ℹ️ It was happening if one or multiple items were already selected.

### Why is it needed?

It's not the expected behaviour and we should stay on the current tab instead of being redirected on the selected files tab.

### How to test it?

- Go to the admin interface > Content Manager
- Select a collection or a single type that has an image field
- Add / Edit an entry
- Click on the image field to open the assets modal
- Select an image
- Try to change the number of items displayed or the page in the pagination

### Related issue(s)/PR(s)

Resolves [@GemaMesasVelazquez](https://github.com/GemaMesasVelazquez) comment on [#23358](https://github.com/strapi/strapi/issues/23358#issuecomment-2918738995)

🚀


remidej (approved):
Thanks for fixing this quick!

PR #23657, opened 2025-06-03
Fix transfer command with proxy

dhoko:
Hello 👋 

### What does it do?

The current version of the command is unable to run inside a CI env with a proxy.

ex:
```
> strapi transfer --from https://xxxx/admin --exclude files --from-token [MASKED] --force --verbose
? The transfer will delete all the local Strapi assets and its database. Are you sure you want to proceed? Yes
Config file not loaded, extension must be one of .js,.json): admin.js.map
Config file not loaded, extension must be one of .js,.json): api.js.map
Config file not loaded, extension must be one of .js,.json): database.js.map
Config file not loaded, extension must be one of .js,.json): features.js.map
Config file not loaded, extension must be one of .js,.json): middlewares.js.map
Config file not loaded, extension must be one of .js,.json): plugins.js.map
Config file not loaded, extension must be one of .js,.json): server.js.map
[2025-06-03 09:56:57.796] info: [remote-source-provider] establishing websocket connection
[2025-06-03 09:56:57.799] info: [local-destination-provider] validating options
[2025-06-03 09:56:58.314] info: [local-destination-provider] Rolling back transaction
[2025-06-03 09:56:58.315] info: [local-destination-provider] Rolled back transaction
Transfer process failed.
```
That's it, nothing else.

### Why is it needed?

So we can execute the command inside an env with a proxy, ex a CI job.

### How to test it?

Set a proxy, run the transfer it fails. Fixed via `sed` inside my CI to validate everything

### Related issue(s)/PR(s)


Bassel17:
Hello thank you for this PR,
I have a small question for you, can you confirm for me how you setup strapi to use your proxy ?
did you pass it in the strapi server config as such 
```
proxy: {
    global:'your_proxy'
  },
  ```

dhoko:
Nope we don't have this inside the config.

Bassel17:
okay thanks, I would encourage to setup the config with the proxy
https://docs.strapi.io/cms/configurations/server
that should make the transfer work with a proxy.
I'll still look into your PR and see if it's something we want to implement.

Bassel17:
Thanks for the time effort but I don't think we will with this solution,
in the meantime you can configure the proxy as I mentioned above

PR #23660, opened 2025-06-03
chore: useScopedPersistentState hook

mazzucchelli:
### What does it do?

Introduce a new custom hook, to store values on localstorage that are only available on the current project

This hook is a wrapper for the existing `usePersistentState`, the difference is

```
const [data, setData] = usePersistentState('KEY', 'value');
localstorage.getItem('KEY') // -> value
```
```
// projectID -> test-uuid
const [data, setData] = useScopedPersistentState('KEY', 'value');
localstorage.getItem('KEY') // -> undefined
localstorage.getItem('KEY:test-uuid') // -> value
```

### Why is it needed?

Certain values, such as trial information, should be restricted to the current project only. When multiple Strapi instances run on localhost, they no longer access values that are relevant to other instances.

### DIFF
--- a/packages/core/admin/admin/src/pages/Home/components/FreeTrialEndedModal.tsx
+++ b/packages/core/admin/admin/src/pages/Home/components/FreeTrialEndedModal.tsx
@@ -7,7 +7,7 @@ import { useIntl } from 'react-intl';
 import styled from 'styled-components';
 
 import { useLicenseLimits } from '../../../../../ee/admin/src/hooks/useLicenseLimits';
-import { usePersistentState } from '../../../hooks/usePersistentState';
+import { useScopedPersistentState } from '../../../hooks/usePersistentState';
 
 const StyledModalContent = styled(Modal.Content)`
   max-width: 51.6rem;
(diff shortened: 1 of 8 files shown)


simotae14 (approved):
it looks good to me, I let @Adzouz give his review

Adzouz (approved):
Very nice, good job 👏🚀

packages/core/admin/admin/src/pages/Home/components/FreeTrialEndedModal.tsx
```diff
@@ -7,7 +7,7 @@ import { useIntl } from 'react-intl';
 import styled from 'styled-components';
 
 import { useLicenseLimits } from '../../../../../ee/admin/src/hooks/useLicenseLimits';
-import { usePersistentState } from '../../../hooks/usePersistentState';
+import { useScopedPersistentState } from '../../../hooks/usePersistentState';
```
soupette, line 10: **question**: Maybe you could introduce this lib in strapi https://usehooks-ts.com/react-hook/use-local-storage this way you can sync the localStorage accross all tabs and keep the scope state as well.

PR #23661, opened 2025-06-03
chore: Update settings nav UI

mazzucchelli:
### What does it do?

Update the lighting icons in the settings left navigation column. Now all links related to premium features always show the icon, gray when the feature is disabled, and blue when the feature is available

Screenshot 2025-06-03 alle 19 21 21

### How to test it?

Using different type of licenses, make sure the icons are correctly rendered

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
@@ -4,6 +4,7 @@ import { useIntl } from 'react-intl';
 import { useLocation } from 'react-router-dom';
 import { styled } from 'styled-components';
 
+import { useLicenseLimits } from '../../../../../ee/admin/src/hooks/useLicenseLimits';
 import { SubNav } from '../../../components/SubNav';
 import { useTracking } from '../../../features/Tracking';
 import { SettingsMenu } from '../../../hooks/useSettingsMenu';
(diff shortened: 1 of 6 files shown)


packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
```diff
@@ -67,7 +79,19 @@ const SettingsNav = ({ menu }: SettingsNavProps) => {
                   endAction={
                     <>
                       {link?.licenseOnly && (
-                        <Lightning fill="primary600" width="1.5rem" height="1.5rem" />
+                        <Lightning
+                          fill={
+                            (availableFeatureNames || []).includes(
+                              linksIdsToLicenseFeaturesNames[
+                                link.id as keyof typeof linksIdsToLicenseFeaturesNames
```
simotae14: just in terms of readability can you create a type definition for this type on top?

packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
```diff
@@ -67,7 +79,19 @@ const SettingsNav = ({ menu }: SettingsNavProps) => {
...
+                              ] as keyof typeof availableFeatureNames
```
simotae14: just in terms of readability can you create a type definition for this type on top?

simotae14 (changes requested):
overall it seems ok to me, please just address the two comments I added. Thanks

simotae14:
Just a little reminder, every time you open a new PR in the CMS to pass the check-pr-status you need to provide a source and a pr labels, like I did right now
Screenshot 2025-06-19 alle 10 13 29
and also assign the Pr to yourself, thanks a lot for the contribution @mazzucchelli

packages/core/admin/admin/src/pages/Settings/components/SettingsNav.tsx
mazzucchelli: Thanks for the feedback, let me know if the last commit improves types enough!

simotae14 (approved):
Thanks for the patience and this enhancement, the code looks good and I tested the solution with an EE license and it works, well done

PR #23662, opened 2025-06-03
feat: conditional form

butcherZ:
### What does it do?

- add conditional form in CTB
- validate schema using rulesEngine
- add badge for field that has conditions

### Why is it needed?

to add conditioinal field

### How to test it?

have fun in ctb

### Related issue(s)/PR(s)

DX-1973
DX-1974


packages/core/content-type-builder/admin/src/components/FormModal/attributes/ConditionForm.tsx
```diff
@@ -0,0 +1,413 @@
...
+}
+
+interface LocalValue {
+  dependsOn: string;
+  operator: 'is' | 'isNot';
+  value: string | boolean;
+  action: 'show' | 'hide';
+}
+
+const convertFromJsonLogic = (jsonLogic: JsonLogicValue): LocalValue => {
```
innerdvations, line 63: Can we add comments explaining all the conversions?

PR #23669, opened 2025-06-04
move generateFileName to image-manipulation to allow overriding

laurenskling:
### What does it do?

This is an alternative to: https://github.com/strapi/strapi/pull/22370
This moves the `generateFileName` function to a service which I could override in my `register`. Now I can have full control over what the generated filename will be to achieve goals presented in the comments in https://github.com/strapi/strapi/pull/22370

### Why is it needed?

I want to have control over the filenames. I do not want the random suffix in my case.

### How to test it?

Upload a file, see that it still works. Override the generateFileName function on the `image-manipulation` service and see if that function is being used on a upload


alexandrebodin (dismissed):
Hey @laurenskling I think this is the way to go. the code LGTM. I'll do some QA

alexandrebodin:
@laurenskling it just broke a backend test because it's missing a mock for this. Can you fix that and we can merge afterwards it's working just fine. 

Also feel free to close the other PR ;)

laurenskling:
@alexandrebodin I think I fixed it :)

alexandrebodin:
@laurenskling Thank you 🤝

Issue #23524, opened 2025-05-14
Duplicate relations with identical documentId returned


channyeintun:
### Node Version

20

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.13.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/channyeintun/strapi-starter-duplicate-relations

### Bug Description

With default sample data for articles and categories,
I used `"@strapi/plugin-graphql": "^5.13.0"` to query categories together with its articles. The reponse includes articles duplicating.
Query is as follows: 
```
query Categories {
  categories {
    name
    articles {
      documentId
      title
    }
  }
}
```
The response is as follows:
```
{
  "data": {
    "categories": [
      {
        "name": "test",
        "articles": [
          {
            "documentId": "zb5p1h2ct1ays90gypvy28jb",
            "title": "test"
          },
          {
            "documentId": "zb5p1h2ct1ays90gypvy28jb",
            "title": "test"
          }
        ]
      }
    ]
  }
}
```

You can see two articles with same documentId.
I have not enabled internationalization.
I only created a single article and a single category with default sample content types and relationships.

### Steps to Reproduce

1. Clone reproduced repo and go to the location.
2. Run docker command to run postgres container as described in README.md
3. Run `npm install` 
4. Run `npm run dev`
5. Add sample data for category and article in the admin panel
6. Go to `localhost:1337/graphql` and run the query described above and see the result

### Expected Behavior

Relations data should not have items duplicated.

### Logs

```shell

```

### Code Snippets

```
query Categories {
  categories {
    name
    articles {
      documentId
      title
    }
  }
}
```

### Media

<img width="957" alt="Image" src="https://github.com/user-attachments/assets/f0d11e06-224a-48e9-9304-c2c74965ec8b" />

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

amarinsam:
I'm running into this as well.  For each single piece of content in Strapi CMS, the Postgres database returns two entry.  However, when I make a request to the relevant Strapi deployed endpoint via the REST API, it returns the correct content, deduplicated.  

I also tested that, saving a document as a draft, and not publishing it only creates 1 row in Postgres.

[...]

Issue #23460, opened 2025-05-02
Relation field returns null in API after related item is unpublished and republished


tommarshall:
### Node Version

v20.18.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.12.7

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

We’re experiencing a critical issue with relation fields in Strapi where valid references appear correctly in the Admin UI but return `null` when queried via the API under specific conditions.

### Steps to Reproduce

1. Create two content types (e.g., Product and Category) with a `has_one` or `has_many` relation (e.g., Product → Category).
2. Create a Category entry and publish.
3. Create a Product entry, linked it to the Category and publish.
4. Fetch the Product via the Strapi API (we've been using REST), with the Category populated. Note that the category data is present. This is expected.
5. Unpublish the Category.
6. View the Product and note that the Category entry is still linked showing as a Draft.
7. Refetch the Product via the Strapi API, with the Category populated. Note that the category is `null`. This is expected.
8. Republish the Category.
9. View the Product and note that the Category entry is still linked showing as Published - identical to how it was after Step 3.
10. Refetch the Product via the Strapi API, with the Category populated. Note that the category is still `null`. This is **not** expected.
11. Republish the Product.
12. Refetch the Product via the Strapi API, with the Category populated. Note that the category data has returned.

### Expected Behavior

The API should return the Category relation as expected (since it’s republished and still appears linked in the Admin UI).

OR

The Admin UI should show the relationship to the Category on the Product item as broken so that content editors are aware action is required.

--

Personally, I think the repairing the relationship is the better solution here as this requires less manual action from content editors.

### Actual Behaviour

The API response for the Product shows the Category relation as `null` after the Category is republished.

This persists until the Product itself is manually republished — despite the Admin UI continuing to display the link as valid.

### Logs

```shell

```

### Confirmed Scope

- Issue occurs in `has_one` and `has_many` relation field types.
- Does **not** occur when the relation is bidirectional.
- Impacts the relation fields in content types and components.
- This data is corrupted broken in the `_lnk` tables, so is not API interface (e.g. REST) specific.
- Reproducible in a clean install using the official Strapi starter project and starter sample data, on the latest Strapi version of v5.12.7.

### Potential Workarounds and Limitations

[...]

PR #23673, opened 2025-06-05
fix: one to many relations returns published and draft versions

markkaylor:
### What does it do?

- Forwards the filters to dynamic zones that contain the 'on' keyword
- Filters a graphql association to return the correct dp version

### Why is it needed?

Dynamic zones and containing a component with a oneToMany relation returns both draft and published versions

Graphql oneToMany relations (at root level) return draft and published versions

### How to test it?

**REST with DZ**
Create a dynamic zone with a component containing a one to many relation
Make a request example query `(/api/tester-rels/g3ihtqhceeog7yq8az0k84xe?populate[dz][on][default.dz-compo-rel][populate]=*)`
It should only return published relations

**GraphQL**
Create a oneToMany relation between two content types
Make a request like:
```
query TesterRels {
  testerRels {
    documentId
    testers {
      title
      documentId
    }
  }
}
```
It should only return published relations

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/23524#issuecomment-2933458701

### DIFF
--- a/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
+++ b/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
@@ -1,4 +1,4 @@
-import { get } from 'lodash/fp';
+import { get, merge } from 'lodash/fp';
 import { async, errors } from '@strapi/utils';
 import type { Internal } from '@strapi/types';
 
@@ -58,8 +58,14 @@ export default ({ strapi }: Context) => {
           }
         );
 
-        const dbQuery = strapi.get('query-params').transform(targetUID, sanitizedQuery);
-
+        const transformedQuery = strapi.get('query-params').transform(targetUID, sanitizedQuery);
+        const defaultFilters = {
+          where: {
+            // Return the same draft and publish version as the parent
+            publishedAt: { $notNull: parent.publishedAt !== null },
+          },
+        };
+        const dbQuery = merge(defaultFilters, transformedQuery);
         const data = await strapi.db?.query(contentTypeUID).load(parent, attributeName, dbQuery);
 
         const info = {
--- a/packages/core/database/src/query/helpers/populate/apply.ts
+++ b/packages/core/database/src/query/helpers/populate/apply.ts
@@ -671,6 +671,28 @@ const pickPopulateParams = (populate: Record<string, unknown>) => {
   return _.pick(fieldsToPick, populate);
 };
 
+const getPopulateValue = (populate: Record<string, any>, filters: Record<string, any>) => {
+  const populateValue = {
+    filters,
+    ...pickPopulateParams(populate),
+  };
+
+  if ('on' in populateValue) {
+    populateValue.on = _.mapValues(
+      (value) => {
+        if (_.isPlainObject(value)) {
+          value.filters = filters;
+        }
+
+        return value;
+      },
+      populateValue.on as Record<string, any>
+    );
+  }
+
+  return populateValue;
+};
+
 const applyPopulate = async (results: Row[], populate: Record<string, any>, ctx: Context) => {
   const { db, uid, qb } = ctx;
   const meta = db.metadata.get(uid);
@@ -686,10 +708,7 @@ const applyPopulate = async (results: Row[], populate: Record<string, any>, ctx:
       throw new Error(`Invalid populate attribute ${attributeName}`);
     }
 
-    const populateValue = {
-      filters: qb.state.filters,
-      ...pickPopulateParams(populate[attributeName]),
-    };
+    const populateValue = getPopulateValue(populate[attributeName], qb.state.filters);
 
     const isCount = 'count' in populateValue && populateValue.count === true;
 
--- a/tests/api/core/strapi/document-service/dp/dz-component-relation.test.api.ts
+++ b/tests/api/core/strapi/document-service/dp/dz-component-relation.test.api.ts
@@ -1,0 +1,111 @@
+'use strict';
+
+import { Core } from '@strapi/types';
+import { createTestBuilder } from 'api-tests/builder';
+import { createStrapiInstance } from 'api-tests/strapi';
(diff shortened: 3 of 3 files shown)


markkaylor:
I forgot the tests for graphql. I'll add them now.

packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
```diff
@@ -58,8 +58,14 @@ export default ({ strapi }: Context) => {
           }
         );
 
-        const dbQuery = strapi.get('query-params').transform(targetUID, sanitizedQuery);
-
+        const transformedQuery = strapi.get('query-params').transform(targetUID, sanitizedQuery);
+        const defaultFilters = {
+          where: {
+            // Return the same draft and publish version as the parent
+            publishedAt: { $notNull: parent.publishedAt !== null },
```
alexandrebodin: what happens if the parent doesn't have any ? should we make sure we return the published data in the case of a parent without draft & publish ?

markkaylor: You mean something like:

```
'publishedAt' in parent ? parent.publishedAt !== null : true
```

I tested a parent without dp and a parent with dp. In all cases there was always a publishedAt attribute. But happy to play it safe 😄

alexandrebodin: you are right with v5 the column is always there. I meant more as a business logic point of view if we should have a rule for non d&p content types; but I can't come up with one so that's fine

alexandrebodin (approved):
Only checked the code make sure to QA this thouroughly 🙏 

 LGTM

markkaylor:
@MarionLemaire 

Would you be able available to QA this one? The setup is kind of tricky. I'll explain it in more detail here but we could also do the QA together. Feel free to ping me.

Create a strapi app with the experimental:
```
npx create-strapi-app@0.0.0-experimental.74ebdf402d90857bc7282394855d43aad6d283e6
```
Choose the option to create an app with data

### GraphQL
Install graphql (I had to do this with the force option)
```
npm install -f @strapi/plugin-graphql": "0.0.0-experimental.74ebdf402d90857bc7282394855d43aad6d283e6
```
Go to the graphql playground `localhost:1337/graphql`
Query categories with articles

```
query Categories {
  categories {
    documentId
    articles {
      title
      documentId
    }
  }
}
```
It should return all categories and all articles on each category should be an empty array (none are published)
Go publish one and run the query again, you should now see an article on the concerned category
Categories does not have draft and publish so you could also test adding it and then querying with DRAFT or PUBLISHED. You should always get only draft or only published entries.

```
query Categories {
  categories(status: DRAFT) {
    documentId
    articles {
      title
      documentId
    }
  }
}
```

### REST DZ

Add a dynamic zone to category
Create a component called `relations` in `shared` with a oneToMany relation to articles
Go to a category and add some articles to the dynamic zone component with articles relations
Then query the REST api with `/api/categories?populate[dz][on][shared.relations][populate]=*`
Same as the graphql tests we should only see published articles in the dynamic zone

MarionLemaire:
Tested on the experimental :

- [x] GraphQL query with D&P disabled on Categories and only Draft articles
- [x] GraphQL query with D&P disabled on Categories and 1 Published article
- [x] GraphQL query with D&P enabled on Categories
- [x] Rest DZ (pair tested with @markkaylor) so far : If the relation is added and then you publish it isn't returned /
if the relation is published and then added it is returned --> issue not related to this PR ✅

markkaylor:
QA'd internally and by those reporting the issue

sami-shukerov:
@markkaylor After this fix (v5.15.1), when I try to query only the DRAFT state of an entry with Dynamic zones containing a component with a oneToMany relation, the component in the dynamic zone is always empty. For PUBLISHED state its working and its returning only the published data.

markkaylor:
> @markkaylor After this fix (v5.15.1), when I try to query only the DRAFT state of an entry with Dynamic zones containing a component with a oneToMany relation, the component in the dynamic zone is always empty. For PUBLISHED state its working and its returning only the published data.

@sami-shukerov Are you using graphql? There is another PR up to fix graphql https://github.com/strapi/strapi/pull/23737 otherwise it could be related to https://github.com/strapi/strapi/issues/23460 which I will investigate next

sami-shukerov:
@markkaylor Yes its in graphql. I will wait for the PR to be deployed and will test it again :)

markkaylor:
> @markkaylor Yes its in graphql. I will wait for the PR to be deployed and will test it again :)

Sounds good. If you have the time you could test the experimental release that has the fix:

`npx create-strapi-app@0.0.0-experimental.cf7b5567d5aec466707a9d9685676df38c5b7c5c`
`npm install -f @strapi/plugin-graphql@0.0.0-experimental.cf7b5567d5aec466707a9d9685676df38c5b7c5c`

This could help us QA and ensure the fix covers all edge cases.

sami-shukerov:
@markkaylor I updated my package.json like 
```
    "@strapi/plugin-graphql": "0.0.0-experimental.cf7b5567d5aec466707a9d9685676df38c5b7c5c",
    "@strapi/strapi": "0.0.0-experimental.cf7b5567d5aec466707a9d9685676df38c5b7c5c",
```

And now it's even worse :( with 5.16.0 I get an empty array for a component in a dynamic zone with relations inside when I want the DRAFT entries, and now all relation fields are empty when I want the DRAFT entries.

markkaylor:
@sami-shukerov could you please provide more information to help me reproduce. For example, the steps in the CMS, your schema, the queries you are making. A video is always helpful as well. 

Please continue the discussion on this issue https://github.com/strapi/strapi/issues/23524 so we centralize the conversation. 

Also for reference there is a [video on my PR](https://github.com/strapi/strapi/pull/23737#issuecomment-2980635007) demonstrating the fix.

PR #23707, opened 2025-06-10
fix(upload): import cropperjs css by styled components

simotae14:
### What does it do?
SOLUTION PROVIDED by [ben39053372](https://github.com/ben39053372) 👏🏽 

fix crop image function not work on production mode

BTW I think this related with pack-up or vite.

### Why is it needed?

crop image page crashed on production mode

### How to test it?

test on production build (create a new strapi project with the EXPERIMENTAL VERSION 0.0.0-experimental.267375bdc00aa57de5e78acd345a838d21a63676)
1. go to media library
2. upload a image
3. edit image
4. click crop image

### Related issue(s)/PR(s)

Solves #21906


packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/PreviewBox.tsx
```diff
@@ -60,6 +60,7 @@ export const PreviewBox = ({
   replacementFile,
   trackedLocation,
 }: PreviewBoxProps) => {
+  const CropperjsStyle = createGlobalStyle`${cropperjscss}`;
```
markkaylor, line 63: The global style thing worries me a bit. Looking at the css (node_modules/cropperjs/dist/cropper.css) it's probably safe 🤷 wdyt @Adzouz ?

Adzouz: Yeah I don't see that being an issue since it's pretty scoped and light css

Adzouz:
I'm good with the solution but the only thing is that is breaking the tests now 😕

Adzouz (approved):

Adzouz:
Thank you for fixing the tests that were failing @simotae14 🙌
We could consider finding another package for the revamp of the media library later.
🚀

Issue #23530, opened 2025-05-16
Page duplication for wrong locale


jsc-zyres:
### Node Version

22.15.0

### NPM/Yarn/PNPM Version

yarn@4.8.1

### Strapi Version

5.13.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

When duplicating the version of a page which exists for different locales, that is not for the default locale
a copy is created for the default locale and not the selected one.

### Steps to Reproduce

1. Create Page in default locale and add version for a different locale
2. Switch to the other locale on the Pages tab and duplicate the page
3. You will now see a duplicate of the page from the default locale and not the one you selected

### Expected Behavior

The page in the selected locale should be duplicated

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

michalkos:
Can confirm, this is since v5.

clone function is not passing any locale information document-manger clone

https://github.com/strapi/strapi/blob/b445ac2d35ee8f4b2d2f16f390c6d16b24de29f9/packages/core/content-manager/server/src/controllers/collection-types.ts#L305
https://github.com/strapi/strapi/blob/b445ac2d35ee8f4b2d2f16f390c6d16b24de29f9/packages/core/content-manager/server/src/services/document-manager.ts#L105

PR #23709, opened 2025-06-10
fix: serve the correct document when we duplicate a document with a locale different than the default one

simotae14:
### What does it do?

Fixes this bug
"When duplicating the version of a page which exists for different locales, that is not for the default locale
a copy is created for the default locale and not the selected one."

### Why is it needed?

Because we were duplicating just the documents with default locale

### How to test it?

1. Create a document in default locale and add version for a different locale
2. Switch to the other locale on the document list view and duplicate the document
3. You will now see a duplicate of the document with the selected locale instead of the default locale

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/23530


packages/core/content-manager/admin/src/hooks/useDocumentActions.ts
```diff
@@ -645,11 +646,13 @@ const useDocumentActions: UseDocumentActions = () => {
 
   const [autoCloneDocument] = useAutoCloneDocumentMutation();
   const autoClone: IUseDocumentActs['autoClone'] = React.useCallback(
-    async ({ model, sourceId }) => {
+    async ({ model, sourceId, locale }) => {
       try {
         const res = await autoCloneDocument({
           model,
           sourceId,
+          data: {},
```
markkaylor: How come empty data? 🤔

simotae14: I have already removed that part

markkaylor (approved):
Nice work, looks good to me. Tested locally.

Adzouz (approved):
Also tested on my side, working great! Very good job 👏

PR #23719, opened 2025-06-11
Fix/design issues

### LINKED ISSUE #21340: [Design] Missing gray background when a relation isn’t possible
Image

Adzouz:
### What does it do?

Fixing a few design issues:

1. Resolves #21340: Added a background for the disable state when a relation isn't possible in Content Type Builder.
Steps to reproduce:
- Go to admin interface > Content Type Builder
- Choose / Create any content type
- Add a new field and choose "Relation"
- On the right relation, select User (from: admin)
- The four items in the middle should be disabled

Before fix:
image

After fix:
Screenshot 2025-06-11 at 11 56 14

2. Fix a glitch when resizing the window making the border disappearing on radio groups in modals
Steps to reproduce:
- Go to admin interface > Content Type Builder
- Choose / Create any content type
- Add a new field and choose "Text"
- Resize the window to see the glitch happening

Video of the glitch:
https://github.com/user-attachments/assets/f5b4a0b7-fe35-4739-be11-3c6a962a6534\

3. Asset is not refreshing when replacing a media by another
Steps to reproduce:
- Go to admin interface > Content Manager
- Choose / Create any single or collection type item that has an image attribute
- If no media already uploaded, upload one and save
- Click on the pencil icon on the media to open the edit asset modal
- Click on "Replace media" and select a new image
- Click on "Finish"

Video of the glitch:
https://github.com/user-attachments/assets/bef9d140-b06b-4541-a2cb-d831148b7c44

4. List of components in Dynamic Zones in Content Type Builder
Steps to reproduce:
- Go to admin interface > Content Type Builder
- Choose / Create any content type
- Create a dynamic zone and add a few components in it
- Resize the window
- 
Before:
image

After:
Screenshot 2025-06-11 at 11 37 40

### Why is it needed?

Improving the experience on the CMS.

### How to test it?

See each fix above.

### Related issue(s)/PR(s)

Resolving #21340 and issues raised by @MarionLemaire 🙏

🚀


packages/core/content-type-builder/admin/src/components/CustomRadioGroup/Styles.tsx
```diff
@@ -27,6 +24,8 @@ const Wrapper = styled(Flex)`
     border-radius: 4px;
     will-change: transform, opacity;
     background: ${({ theme }) => theme.colors.neutral0};
+    border: 1px solid ${({ theme }) => theme.colors.neutral200};
+    border-radius: 4px;
```
markkaylor: I think there is a borderRadius value on theme and otherwise there is also spaces? theme.spaces[0]. This is probably fine too though.

packages/core/upload/admin/src/components/MediaLibraryInput/Carousel/CarouselAsset.tsx
```diff
@@ -53,12 +53,18 @@ export const CarouselAsset = ({ asset }: { asset: FileAsset }) => {
   }
 
   if (asset.mime?.includes(AssetType.Image)) {
+    const assetUrl = createAssetUrl(asset, true);
+    if (!assetUrl) return null;
+
+    // Adding a param to the url to bust the cache and force the refresh of the image when replaced
+    const cacheBustedUrl = `${assetUrl}${assetUrl.includes('?') ? '&' : '?'}updatedAt=${asset.updatedAt}`;
```
markkaylor, line 60: Is there not a way to invalidate the cache with react-query?

markkaylor (commented):
Looks good to me. Couple quick questions. Haven't tested yet

packages/core/upload/admin/src/components/MediaLibraryInput/Carousel/CarouselAsset.tsx
Adzouz: I don't think so, I think it's on browser side 😕 I'll try

packages/core/content-type-builder/admin/src/components/CustomRadioGroup/Styles.tsx
Adzouz: Good catch, updated!

packages/core/upload/admin/src/components/MediaLibraryInput/Carousel/CarouselAsset.tsx
Adzouz: So no, since it's on browser side and that we already use react-query for refreshing the assets, it's the only solution I came with (not the most elegant but the most simple and efficient).

markkaylor:
Heads up on the commit message if you squash and merge fix: not Fix/

PR #23726, opened 2025-06-11
fix: validation error code

innerdvations:
### What does it do?

- Loosens the check for error types to only compare constructor name rather than actually using `instanceof`
- Changes the CI workflow to run API tests when the middlewares files are modified so this change can be tested

### Why is it needed?

This solves an issue currently causing tests to fail where some ValidationErrors are being sent to the content API as 500 errors instead of their correct status code because despite being a ValidationError, when we check how to format the error we use `instanceof` which fails because wherever the ValidationError is coming from it's a different instance of the class.

I couldn't quickly figure out where the problem was actually coming from, and this workaround should be fine considering v4 is in maintenance mode; it actually makes it more resilient in exchange for potential problems if someone creates incompatible error classes to our own with identical names but nobody should be doing that anyway and that's an extreme edge case where the worst that would happen is an invalid status code from what I can tell.

### How to test it?

### Related issue(s)/PR(s)

DX-2049

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -8,6 +8,7 @@ backend:
   - 'packages/core/*/{lib,bin,ee}/**'
   - 'tests/api/**'
   - 'packages/core/database/**'
+  - 'packages/**/**/middlewares/**/*.(js|ts)'
 frontend:
   - '.github/actions/yarn-nm-install/*.yml'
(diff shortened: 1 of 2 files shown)


innerdvations:
Fixed elsewhere

PR #23728, opened 2025-06-11
Bug fix: Enum values named true prevent server from starting

pritam-ago:
### What does it do?

This PR adds validation to prevent using GraphQL reserved keywords (`true`, `false`, `null`, `undefined`) as enum values in the Content-Type Builder. The changes include:

1. Frontend validation in the CTB to prevent users from entering invalid values
2. Backend validation to ensure data integrity
3. Clear error messages for both frontend and backend validation

### Why is it needed?

Currently, the Content-Type Builder allows saving enum fields with values "true" and "false", which causes GraphQL to fail on server startup with the error:

This happens because GraphQL has specific requirements for enum values and doesn't allow using reserved keywords. The fix prevents this issue by validating enum values before they can be saved.

### How to test it?

1. Start the Strapi development server:
```bash
yarn develop
```

2. Test the Content-Type Builder:
   - Go to Content-Type Builder
   - Create a new content type or edit an existing one
   - Add an enumeration field
   - Try to add the following values:
     - "true" (should be rejected)
     - "false" (should be rejected)
     - "null" (should be rejected)
     - "undefined" (should be rejected)
     - "valid_enum" (should be accepted)

3. Test the API:
   - Try to create a content type with an enum field containing reserved keywords
   - The request should be rejected with a validation error

4. Run the tests:
```bash
# Run all tests except upload tests (which have Windows permission issues)
yarn test:api --testPathIgnorePatterns="tests/api/core/upload"

# Run specific enum validation tests
yarn test:api tests/api/core/content-type-builder/collection-type.test.api.js
```

### Related issue(s)/PR(s)

This PR fixes the issue where the CTB allows saving enum fields with values "true" and "false", causing GraphQL to fail on server startup.


Bassel17 (changes requested):
Thanks for looking into this. I agree this startup failure should be fixed, but I don’t think we should solve it by blocking these enum values in CTB/core validation.

`true`, `false`, and `null` are valid domain values for Strapi string-backed enum fields, and this is not a REST/content API issue. The failure is specific to GraphQL schema generation, where the stored enum value is reused as the GraphQL enum literal name.

Could you rework the PR to fix this in the GraphQL enum builder instead? The main place to check is:

`packages/plugins/graphql/server/src/services/builders/enums.ts`

That builder already maps GraphQL enum member names to internal stored values, so we should be able to escape/remap only the GraphQL enum literal names while keeping the stored Strapi enum values unchanged. The implementation should also handle collisions after mapping, e.g. if both `true` and the escaped alias are present.

I’d also avoid blocking `undefined`, since it does not appear to be forbidden as a GraphQL enum value.

Bassel17:
I’m going to close this PR since the current approach blocks valid Strapi enum values globally. Happy to review a new PR, or this one can be reopened, if you want to tackle it in the GraphQL enum mapping layer instead.

PR #23729, opened 2025-06-11
enhancement: replace lodash cloneDeep with structuredClone behind a type-checked helper

unrevised6419:
### What does it do?

Replaces lodash [`cloneDeep`](https://lodash.com/docs/4.17.15#cloneDeep) with the native [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone), called through a small typed helper.

- Adds `structuredCloneSafe` to `@strapi/utils` (`packages/core/utils/src/structured-clone-safe.ts`) and exports it from the package root. It is `structuredClone` with a compile-time check: `structuredCloneSafe<T>(value: T & CloneCheck<T>): T`, where the check walks `T` and replaces every position the structured clone algorithm cannot serialize — functions, symbols, promises, weak collections — with a marker type, so a bad value is a type error naming the offending property instead of a runtime `DataCloneError`.
- The accepted set follows the [MDN supported types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) list. Web/API types (`Blob`, `File`, `ImageData`, `DOMRect`, …) are resolved off `globalThis`, so each one is recognised where the runtime declares it and ignored where it does not — the same file compiles in the DOM-less server packages and in the admin.
- 56 call sites across 35 files now go through the helper. `scripts/front/utils/search-files-content.js` calls `structuredClone` directly, since it is a build script outside the workspace packages.
- `tests/setup/unit.setup.js` rebinds `globalThis.structuredClone` to `core-js-pure/actual/structured-clone`. `jest-environment-node` injects the host realm's `structuredClone` into the sandbox, so its results carry the host realm's prototypes and `toStrictEqual` fails with "serializes to the same string" ([jestjs/jest#2549](https://github.com/jestjs/jest/issues/2549)). The root dev dependency moves from `core-js` to `core-js-pure` for that reason.
- `packages/plugins/documentation/server/src/services/__tests__/build-component-schema.test.ts` builds its mocks from a factory: the shared mock held a function, which is not cloneable.

### Why is it needed?

Drops a third-party dependency in favour of a platform API.

It is not a like-for-like swap, and that is the point of the helper. `structuredClone` throws on functions and symbols where `cloneDeep` copies them, drops symbol-keyed properties, does not walk the prototype chain (class instances come back as plain objects, private fields and accessors are lost), and does not preserve `RegExp.lastIndex`. Doing this by hand hides those differences until runtime; the type surfaces them at the call site.

### How to test it?

`yarn test:unit`, `yarn test:front` and `yarn test:ts`.

Still a draft: `yarn test:ts` currently reports 11 errors in 6 packages, each one a value the new type rejects. Seven are genuine cloneability problems the swap would otherwise have hidden:

- `packages/core/database/src/metadata/metadata.ts:54` — `JoinColumn.on` may be a function.
- `packages/core/core/src/domain/content-type/index.ts:32,155` and `packages/core/content-type-builder/server/src/services/schema-builder/schema-handler.ts:40,88,94` — schema attribute fields accept `() => string`.
- `packages/core/admin/admin/src/hooks/useMenu.ts:166` — `MenuItem.icon` is a React component.

The remaining four are not cloneability failures. `packages/core/utils/src/provider-factory.ts:71,80,84` clone a value of a still-unresolved generic parameter, which no conditional type can validate; they need a widening cast. `packages/utils/upgrade/src/modules/json/transform-api.ts:22` needs an explicit type argument.

### Related issue(s)/PR(s)

Part of the effort to reduce lodash usage across the monorepo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,7 @@
     "chokidar": "3.6.0",
     "coffee": "5.5.1",
     "commander": "8.3.0",
-    "core-js": "3.36.0",
+    "core-js-pure": "3.36.0",
     "create-strapi-app": "workspace:*",
     "cross-env": "7.0.3",
     "dotenv": "16.6.1",
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "chokidar": "3.6.0",
     "coffee": "5.5.1",
     "commander": "8.3.0",
-    "core-js": "3.36.0",
+    "core-js-pure": "3.36.0",
     "create-strapi-app": "workspace:*",
     "cross-env": "7.0.3",
     "dotenv": "16.6.1",
--- a/packages/admin-test-utils/src/patches/node-globals.ts
+++ b/packages/admin-test-utils/src/patches/node-globals.ts
@@ -69,7 +69,7 @@ const NODE_GLOBALS = Object.freeze({
   fetch,
   AbortController,
   AbortSignal,
-  structuredClone,
+  structuredClone: globalThis.structuredClone,
   BroadcastChannel,
 } as const);
 
--- a/packages/core/admin/admin/src/hooks/useMenu.ts
+++ b/packages/core/admin/admin/src/hooks/useMenu.ts
@@ -1,7 +1,6 @@
 import * as React from 'react';
 
 import { Cog, ShoppingCart, House } from '@strapi/icons';
-import cloneDeep from 'lodash/cloneDeep';
 
 import { useTypedSelector } from '../core/store/hooks';
 import { useAuth, AuthContextValue } from '../features/Auth';
@@ -163,7 +162,7 @@ const getGeneralLinks = async (
     return [];
   }
 
-  const authorizedGeneralLinksClone = cloneDeep(authorizedGeneralSectionLinks);
+  const authorizedGeneralLinksClone = globalThis.structuredClone(authorizedGeneralSectionLinks);
 
   authorizedGeneralLinksClone[settingsLinkIndex].notificationsCount = shouldUpdateStrapi ? 1 : 0;
 
--- a/packages/core/admin/admin/src/hooks/useMenu.ts
+++ b/packages/core/admin/admin/src/hooks/useMenu.ts
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { Cog, ShoppingCart, House } from '@strapi/icons';
-import cloneDeep from 'lodash/cloneDeep';
+import { structuredCloneSafe } from '@strapi/utils';
 
 import { useTypedSelector } from '../core/store/hooks';
 import { useAuth, AuthContextValue } from '../features/Auth';
@@ -163,7 +163,7 @@ const getGeneralLinks = async (
     return [];
   }
 
-  const authorizedGeneralLinksClone = cloneDeep(authorizedGeneralSectionLinks);
+  const authorizedGeneralLinksClone = structuredCloneSafe(authorizedGeneralSectionLinks);
 
   authorizedGeneralLinksClone[settingsLinkIndex].notificationsCount = shouldUpdateStrapi ? 1 : 0;
 
--- a/packages/core/admin/admin/src/pages/Settings/pages/Roles/components/Permissions.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Roles/components/Permissions.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 
 import { Tabs } from '@strapi/design-system';
 import { produce } from 'immer';
-import cloneDeep from 'lodash/cloneDeep';
 import get from 'lodash/get';
 import has from 'lodash/has';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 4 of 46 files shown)


unrevised6419:
> cloneDeep is loosely based on the [structured clone algorithm](https://mdn.io/Structured_clone_algorithm)
> — https://lodash.com/docs/4.18.1#clone

> The prototype chain is not walked or duplicated.
> — https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

Note: This means `cloneDeep` is not exactly identical with `structuredClone` and some tests fail because they use `toStrictEqual`

package.json
```diff
@@ -151,7 +151,7 @@
     "chokidar": "3.6.0",
     "coffee": "5.5.1",
     "commander": "8.3.0",
-    "core-js": "3.36.0",
+    "core-js-pure": "3.36.0",
```
unrevised6419, line 154: > core-js patches the global environment. When you import it, Array.prototype.includes, Promise, globalThis.Symbol, etc. are modified/added directly on the global objects and prototypes. So after importing, plain code like [1,2,3].includes(2) just works everywhere.

> core-js-pure never touches globals or prototypes. Instead, it exports polyfilled functionality as standalone modules that you import and use explicitly

`core-js` it seems not to be used at all since https://github.com/strapi/strapi/commit/6e46e601548f990b83880ea84ee2f416a85b56f7

unrevised6419:
Hey @innerdvations, please trigger again the CI for this one, pushed some fixes.

unrevised6419:
## ⚠️ This change breaks app boot — `structuredClone` is not a drop-in for `_.cloneDeep`

I audited all **39 production `structuredClone` call sites** introduced here (4 parallel passes, then verified against a real boot). **5 of them throw `DataCloneError` at runtime, 4 on the normal startup path.** A fresh `examples/getstarted` boot on this branch crashes before the server starts:

```
[ERROR] There seems to be an unexpected error…

DataCloneError: beforeUpdate() {
  const ctx = strapi.requestContext.get();
} could not be cloned.
    at new DOMException (node:internal/per_context/domexception:76:18)
    at structuredClone (node:internal/worker/js_transferable:126:26)
    at Object.createContentType (packages/core/core/dist/domain/content-type/index.js:21:56)
    at Object.add (packages/core/core/dist/registries/content-types.js:48:43)
    at loadAPIs (packages/core/core/dist/loaders/apis.js:41:28)
    at async Object.loadApplicationContext (packages/core/core/dist/loaders/index.js:13:5)
```

The root cause: `_.cloneDeep` copies functions by reference and preserves prototypes; `structuredClone` throws on functions, symbols, and class instances (the [structured-clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)). Strapi's content-type definitions, DB models, condition providers, and admin menu links **all carry functions**, so replacing the clone primitive globally is unsafe. Unit tests stayed green only because their fixtures are plain data — nothing exercised a real content-type graph, which is why the first real boot dies.

### 🔴 Boot-blocking (server won't start)

| Site | Cloned value | Non-cloneable content |
|------|--------------|-----------------------|
| `core/core/src/domain/content-type/index.ts:32` | `structuredClone(definition)` | `lifecycles` = user hook **functions** (`beforeUpdate`, …). **Confirmed crash above** — fires first, during `loadAPIs`. |
| `core/database/src/metadata/metadata.ts:53` | `structuredClone(models)` | attribute `default` **functions** — `documentId`'s `createDocumentId`, `publishedAt`'s `default() { return new Date() }`. Fatal on essentially every app (not reached in this run only because the site above dies first). |
| `core/utils/src/hooks.ts:103` | `structuredClone(context)` in `content-types.beforeSync`/`afterSync` | context carries content-types whose `publishedAt.default` is a **function** — startup DB sync. |
| `core/utils/src/provider-factory.ts:70,79,83` | `structuredClone(item)` | condition `handler` **functions** (`admin-conditions`) — registered at bootstrap. |

### 🟠 Admin runtime crash

| Site | Cloned value | Non-cloneable content |
|------|--------------|-----------------------|
| `core/admin/admin/src/hooks/useMenu.ts:165` | `structuredClone(authorizedGeneralSectionLinks)` | each link's `icon` is a `@strapi/icons` `forwardRef` (**function + `Symbol.for('react.forward_ref')`**) → throws whenever the authorized general menu is computed. |

### 🟢 Safe (the other 34 call locations)

Verified plain, JSON-ish data with no prototype reliance: the query/filter clones (`permissions-manager` ×2, `convert-query-params`, `utils/sanitize`), `history` (JSON columns), `schema-handler`, `upload` file entities, all i18n sites, `upgrade` JSON transforms, `scripts`, `content-type/index.ts:154` (picks whitelisted schema keys only), `loaders/components.ts:52`, permissions-engine `permission` (conditions are string names, resolved to handlers later), `users-permissions` actionMap, `Permissions.tsx modifiedData` (booleans), `EditView data`, `content-type-builder/formAPI`. No `PROTOTYPE-LOSS` cases anywhere.

### Suggested path

`structuredClone` is fine for the 34 proven-plain sites but wrong for the 5 above. Two options:

1. **Keep `_.cloneDeep` for the 5 flagged sites**, `structuredClone` everywhere else — smallest safe diff, preserves the PR's intent for the data-only clones.
2. **Revert the migration** if a single uniform clone primitive is preferred over a mixed strategy.

Either way, please add a real boot check (e.g. `examples/getstarted`) to the verification for this PR — the unit suite alone can't catch this class of regression.

<sub>Audit methodology: enumerate all non-test `structuredClone(` sites → trace each cloned value to its runtime origin for function/symbol/class-instance/non-cloneable content → confirm on a live `yarn develop` boot.</sub>

PR #23731, opened 2025-06-12
feat(ctb): Tracking events for AI Chat interaction

innerdvations:
### What does it do?

Adds analytics events for the AI CTB chat

- [x] `didStartNewChat` 
Description: Triggered when a new conversation is initiated in the widget.
Properties: None.

- [x] `didUsePresetPrompt`
Description: Triggered when a user selects one of the preset questions.

Properties:

promptType: Specifies the type of preset question used.
Possible values:
generate-product-schema
tell-me-about-the-content-type-builder
tell-me-about-strapi
(Note: These values are lowercase with dashes replacing spaces.)

- [x] `didUserSendMessage`
Description: Triggered when a message is sent in the chat.

Properties:

attachment-type: Indicates the type of attachment sent with the message.
Possible values:
code — File uploaded from the local computer
figma — Figma file
image — Image file
none – No attachment

number-of-input-tokens: Indicates the number of input tokens
Possible values:
number values

- [x] `didAnswerMessage`
Description: Triggered when the AI provides a response to a message.

Properties:
successful: Indicates whether the response was delivered successfully.
Possible values:
true
false (e.g., in case of an error)

- [x] `didVoteAnswer`
Description: Triggered when the user gives feedback on an AI answer via thumbs up or thumbs down.

Properties:
value: Represents the type of feedback given.
Possible values:
positive
negative

- [x] Adds `ai_license_key` property at the project level
Description: Tracks the AI license key that is stored as value of STRAPI_ADMIN_AI_LICENSE.
Possible values:
the STRAPI_ADMIN_AI_LICENSE license key value
NOTE: This will be replaced after beta

### Why is it needed?

improving user experience

### How to test it?

TODO

### Related issue(s)/PR(s)

DX-2026

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,7 +163,6 @@
     "rimraf": "5.0.5",
     "rollup": "4.27.4",
     "rollup-plugin-html": "0.2.1",
-    "rollup-plugin-postcss": "4.0.2",
     "semver": "7.5.4",
     "stream-chain": "2.2.5",
     "stream-json": "1.8.0",
--- a/packages/core/admin/admin/src/components/SearchInput.tsx
+++ b/packages/core/admin/admin/src/components/SearchInput.tsx
@@ -72,6 +72,11 @@ const SearchInput = ({
           clearLabel={formatMessage({ id: 'clearLabel', defaultMessage: 'Clear' })}
           onClear={handleClear}
           placeholder={placeholder}
+          onBlur={(e) => {
+            if (!e.currentTarget.contains(e.relatedTarget) && e.currentTarget.value === '') {
+              setIsOpen(false);
+            }
+          }}
         >
           {label}
         </Searchbar>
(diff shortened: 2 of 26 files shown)


packages/core/content-type-builder/admin/src/components/AIChat/providers/ChatProvider.tsx
```diff
@@ -99,7 +111,26 @@ export const BaseChatProvider = ({
...
+  const getTokenCount = (text: string): number => {
+    // TODO: Implement token counting
+    return text.length;
+  };
+
   const handleSubmit = async (event: Parameters<typeof chat.handleSubmit>[0]) => {
+    const tokenCount = getTokenCount(chat.input || '');
+
+    trackUsage('didUserSendMessage', {
+      'attachment-type': currentAttachmentType,
```
Mcastres, line 130: Why these event_properties are not camelCase like we use to have them?

innerdvations: Oops, that's probably an issue.

Could you move this comment to https://github.com/strapi/strapi/pull/23730 and we can be sure to investigate before merging the feature?

PR #23737, opened 2025-06-13
fix(graphql): draft and publish document does not forward status arg …

markkaylor:
### What does it do?

- Stores the root query args in context so they can be accessed in the association resolver

### Why is it needed?

- parent doesn't refer to the root, in the context of a component parent is the component not the root document so it doesn't have a publishedAt attribute. 
- associations should use the same status as the root query for any relations with draft and publish

### How to test it?

GIVEN a content-type WITHOUT draft and publish that has relations, components with relations, dynamic zones with components with relations
OR 
GIVEN a content-type WITH draft and publish that has relations, components with relations, dynamic zones with components with relations

WHEN 

```
query Articles {
  articles {
    documentId
    publishedAt
    categories {
      documentId
      name
      locale
      publishedAt
    }
    compo {
      categories {
        documentId
        name
        publishedAt
      }
    }
    dz {
      ... on ComponentDefaultRelations {
        categories {
          documentId
          name
          locale
          publishedAt
        }
      }
    }
  }
}
```

THEN

- Should return all published entries all the way down the tree (NEVER both draft and published)
- The same query with articles(status: DRAFT) should return all draft entries all the way down the tree
- The same query with articles(status: PUBLISH) should return all published entries all the way down the tree

See video below where I have relations, component with relations, dz with compoonent with relations for one content-type with draft and publish and one content-type without draft and publish. 

https://github.com/user-attachments/assets/5a6544d2-550f-40c0-aaa4-dee23e009720

Test the extensions api

In your app's index file's register function add a custom resolver

```
const extension = {
  resolversConfig: {
    'Query.articleDpViaCat': {
      auth: false,
    },
  },
  typeDefs: `
    type Query {
      articleDpViaCat(catName: String!, status: String): Articledp
    }
  `,
  resolvers: {
    Query: {
      articleDpViaCat: {
        resolve: async (parent, args, context, info) => {
          const entry = await strapi.documents('api::article.article').findFirst({
            filters: {
              category: { name: args.catName },
            },
            status: args.status || 'published',
          });

          return entry;
        },
      },
    },
  },
};
// ...
register({ strapi }) {
    strapi.plugin('graphql').service('extension').use(extension);
  },
//...
```

Then query it in the playground. The expectations should be the same as above.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/23524#issuecomment-2966669833
https://github.com/strapi/strapi/pull/23673

### DIFF
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -121,6 +121,15 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
   const path: string = config('endpoint');
 
   const landingPage = determineLandingPage(strapi);
+  /**
+   * We need the arguments passed to the root query to be available in the association resolver
+   * so we can forward those arguments along to any relations.
+   *
+   * In order to do that we are currently storing the arguments in context.
+   * There is likely a better solution, but for now this is the simplest fix we could find.
+   *
+   * @see https://github.com/strapi/strapi/issues/23524
+   */
   const pluginAddRootQueryArgs: ApolloServerPlugin<StrapiGraphQLContext> = {
     async requestDidStart() {
       return {
@@ -128,6 +137,7 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
           return {
             willResolveField({ source, args, contextValue, info }) {
               if (!source && info.operation.operation === 'query') {
+                // NOTE: context.rootQueryArgs is intended for internal use only
                 contextValue.rootQueryArgs = args;
               }
             },
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -20,7 +20,15 @@ const merge = mergeWith((a, b) => {
   }
 });
 
-export const determineLandingPage = (strapi: Core.Strapi) => {
+type StrapiGraphQLContext = BaseContext & {
+  rootQueryArgs?: Record<string, unknown>;
+  state: any; // if you want to be stricter, replace 'any' with your app-specific state shape
+  koaContext: any;
+};
+
+export const determineLandingPage = (
+  strapi: Core.Strapi
+): ApolloServerPlugin<StrapiGraphQLContext> => {
   const { config } = strapi.plugin('graphql');
   const utils = strapi.plugin('graphql').service('utils');
 
@@ -115,6 +123,21 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
   const path: string = config('endpoint');
 
   const landingPage = determineLandingPage(strapi);
+  const pluginAddRootQueryArgs: ApolloServerPlugin<StrapiGraphQLContext> = {
+    async requestDidStart() {
+      return {
+        async executionDidStart() {
+          return {
+            willResolveField({ source, args, contextValue, info }) {
+              if (!source && info.operation.operation === 'query') {
+                contextValue.rootQueryArgs = args;
+              }
+            },
+          };
+        },
+      };
+    },
+  };
 
   type CustomOptions = {
     cors: boolean;
@@ -122,7 +145,7 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
     bodyParserConfig: boolean;
   };
 
-  const defaultServerConfig: ApolloServerOptions<BaseContext> & CustomOptions = {
+  const defaultServerConfig: ApolloServerOptions<StrapiGraphQLContext> & CustomOptions = {
     // Schema
     schema,
 
@@ -138,7 +161,7 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
     bodyParserConfig: true,
     // send 400 http status instead of 200 for input validation errors
     status400ForVariableCoercionErrors: true,
-    plugins: [landingPage],
+    plugins: [landingPage, pluginAddRootQueryArgs],
 
     cache: 'bounded' as const,
   };
@@ -146,7 +169,7 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
   const serverConfig = merge(
     defaultServerConfig,
     config('apolloServer')
-  ) as ApolloServerOptions<BaseContext> & CustomOptions;
+  ) as ApolloServerOptions<StrapiGraphQLContext> & CustomOptions;
 
   // Create a new Apollo server
   const server = new ApolloServer(serverConfig);
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -22,8 +22,6 @@ const merge = mergeWith((a, b) => {
 
 type StrapiGraphQLContext = BaseContext & {
   rootQueryArgs?: Record<string, unknown>;
-  state: any; // if you want to be stricter, replace 'any' with your app-specific state shape
-  koaContext: any;
 };
 
 export const determineLandingPage = (
--- a/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
+++ b/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
@@ -57,16 +57,21 @@ export default ({ strapi }: Context) => {
             auth,
           }
         );
-
         const transformedQuery = strapi.get('query-params').transform(targetUID, sanitizedQuery);
 
         const defaultFilters = {
           where: {
-            // Return the same draft and publish version as the parent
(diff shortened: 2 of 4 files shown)


packages/plugins/graphql/server/src/services/builders/resolvers/query.ts
```diff
@@ -16,6 +16,9 @@ export default ({ strapi }: Context) => ({
           auth: ctx?.state?.auth,
         });
 
+        // Store the root query args in context so they can be accessed deeply down the tree (dz, components, etc..)
+        ctx.rootQueryArgs = args;
```
markkaylor: ℹ️ I'm not sure this is the most elegant solution but I was unable to find another way, lmk what you think or if you have other suggestions

markkaylor:
~.....and I broke the tests 🤦‍♂️~

broksonic21:
thanks for jumping on this one - it's blocking our upgrade, so fully appreciate it!

packages/plugins/graphql/server/src/services/builders/resolvers/query.ts
innerdvations: Nothing comes to me; but maybe add a comment that this is for internal use only to dissuade anyone from using it so we can change it later if necessary.

packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
```diff
@@ -57,16 +57,21 @@ export default ({ strapi }: Context) => {
...
         const defaultFilters = {
           where: {
-            // Return the same draft and publish version as the parent
-            publishedAt: { $notNull: 'publishedAt' in parent ? parent.publishedAt !== null : true },
+            publishedAt: {
+              $notNull: context.rootQueryArgs?.status
+                ? // Filter by the same status as the root query if the argument is present
+                  context.rootQueryArgs?.status === 'published'
+                : // Otherwise fallback to the published version
+                  true,
```
innerdvations: I can see some users wanting to only select published items at the top level, but include the draft relations in children.

So I think we should use:

local arg if present -> rootArg if present -> published by default

innerdvations (commented):
I think this is a reasonable solution.

We also need to make sure we document that this param carries through and others don't and the reasons why. I have seen issues in the past where users sometimes expect all params to be passed down, and now that one does and others don't that could be extra confusing.

markkaylor:
> We also need to make sure we document that this param carries through and others don't and the reasons why. I have seen issues in the past where users sometimes expect all params to be passed down, and now that one does and others don't that could be extra confusing.

Thanks for the review. I think I am passing all params (args) through with 

```
ctx.rootQueryArgs = args;
```

it's just that I'm only accessing `status` where I need it in the associations file. Let me know if I misunderstood something 🤔

packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
markkaylor: I would have thought this was already the case since I merge the arguments with the defaults I set up. However if you query PUBLISHED at the root and then try to filter with something like 

```
    dz {
      ... on ComponentDefaultRelations {
        categories(filters: { publishedAt: { notNull: false } }) {
          documentId
          name
          locale
          publishedAt
        }
      }
    }
```

It does override the default filter but the response is always an empty array. 

Screenshot 2025-06-16 at 15 44 11

A similar query with REST api returns the same results:

```
localhost:1337/api/articledps?populate[dz][on][default.relations][populate][categories][filters][publishedAt][$notNull]=false&status=published
```

Screenshot 2025-06-16 at 15 43 44

So I want to say this is the expected behavior... 🤷‍♂️

innerdvations: Ahh, sorry, I missed that down below! I think you're right, no problem here then.

innerdvations:
> > We also need to make sure we document that this param carries through and others don't and the reasons why. I have seen issues in the past where users sometimes expect all params to be passed down, and now that one does and others don't that could be extra confusing.
> 
> Thanks for the review. I think I am passing all params (args) through with
> 
> ```
> ctx.rootQueryArgs = args;
> ```
> 
> it's just that I'm only accessing `status` where I need it in the associations file. Let me know if I misunderstood something 🤔

I mean that we're only carrying down status as a default, right? So if someone applies one filter at the top level, it doesn't apply to all the children. It's only status that gets applied to the defaults.

markkaylor:
> > > We also need to make sure we document that this param carries through and others don't and the reasons why. I have seen issues in the past where users sometimes expect all params to be passed down, and now that one does and others don't that could be extra confusing.
> > 
> > 
> > Thanks for the review. I think I am passing all params (args) through with
> > ```
> > ctx.rootQueryArgs = args;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it's just that I'm only accessing `status` where I need it in the associations file. Let me know if I misunderstood something 🤔
> 
> I mean that we're only carrying down status as a default, right? So if someone applies one filter at the top level, it doesn't apply to all the children. It's only status that gets applied to the defaults.

No we are passing any and all arguments on the root to ctx.rootQueryArgs. `status` may not even be defined on those args and in that case we fallback to only returning published entries

markkaylor:
@MarionLemaire could you QA this one 🙏 experimental: `0.0.0-experimental.ae2010c73ea60e398b726874a9b7b2bf3faada51`

markkaylor:
@innerdvations I pushed a fix to handle an issue with nondp relations with a dp parent


Draft and publish parent:
- Relations that also have draft and publish should return the same status as the parent (request draft parent only get draft relations). This should work at all levels  (components, dz)
- Relations that do NOT have draft and publish should ALWAYS be returned (request draft parent get nondp relation, request published parent get nondp relations) 



https://github.com/user-attachments/assets/4c61c045-5052-4926-8ded-f7c49e05cbf0

markkaylor:
New experimental: `0.0.0-experimental.cf7b5567d5aec466707a9d9685676df38c5b7c5c`. I'll add a test for the latest commit

michalkos:
hello, does this fix https://github.com/strapi/strapi/issues/23460 ?

markkaylor:
> hello, does this fix #23460 ?

No I don't think so. This PR is specific to graphql.

markkaylor:
@innerdvations the previous fix did not take into account [schema extensions](https://docs.strapi.io/cms/plugins/graphql#extending-the-schema) which allows users to provider custom resolvers. I've pushed a new fix to handle that case as well.

MarionLemaire:
Tested today on `0.0.0-experimental.f3ec579d505563eb88e01308248766a76e49d9b3` : LGTM 🎉

packages/plugins/graphql/server/src/bootstrap.ts
```diff
@@ -115,14 +121,29 @@ export async function bootstrap({ strapi }: { strapi: Core.Strapi }) {
   const path: string = config('endpoint');
 
   const landingPage = determineLandingPage(strapi);
+  const pluginAddRootQueryArgs: ApolloServerPlugin<StrapiGraphQLContext> = {
```
innerdvations, line 133: Can you add a comment for this? I'm worried all of this will end up getting forgotten about over time.

innerdvations (approved):
LGTM! My only concern is documenting what we did and why, I suspect at some point we'll find a better way even though nothing is apparent right now.

packages/plugins/graphql/server/src/bootstrap.ts
markkaylor: yes sorry I thought I had one but must have deleted it on accident at some point

sami-shukerov:
@markkaylor I’ve just noticed that starting with Strapi v5.16.1 (or possibly as a side effect of this fix), there seems to be a new issue affecting custom GraphQL queries.
Specifically, when querying a collection type that includes a relation field, the related data is returned as null instead of the expected fields. I’ve attached a minimal reproducible example to demonstrate the issue.
{82342AC6-93D0-4EC2-B65E-6803C0AE53D7}
{22B5A7FD-D7CB-455C-88C8-B5347A513E84}


https://github.com/user-attachments/assets/fceb0cc6-1734-4813-85a8-55cc184851a9

markkaylor:
> @markkaylor I’ve just noticed that starting with Strapi v5.16.1 (or possibly as a side effect of this fix), there seems to be a new issue affecting custom GraphQL queries. Specifically, when querying a collection type that includes a relation field, the related data is returned as null instead of the expected fields. I’ve attached a minimal reproducible example to demonstrate the issue. {82342AC6-93D0-4EC2-B65E-6803C0AE53D7} {22B5A7FD-D7CB-455C-88C8-B5347A513E84}
> 
>  2025-07-10_12h37_20.mp4

Can you please create a new bug report and I will investigate 🙏

PR #23739, opened 2025-06-13
fix: display name of upload plugin in roles settings

### LINKED ISSUE #20842: We should rename the Upload plugin to Media Library in the Permissions view both for the Admin Roles and Users & Permissions Roles
Image

Adzouz:
### What does it do?

Changes the name of the upload plugin in Roles and Permissions from Upload to Media Library.

Before:

image

After:

Screenshot 2025-06-13 at 11 39 05

### Why is it needed?

It's confusing to the user since the plugin is called Media Library on the CMS. Even if the plugin is called upload, on the frontend it's displayed and identified as Media Library.
In a second time, we should consider moving the plugin itself but for now we just changed the displayed name in the Roles and Permissions.

### How to test it?

- Go to the admin interface > Settings > Roles
- Select any role
- Go to Plugins tab (if in Admin Roles) or by just scrolling to the accordion (if in Users & Permissions)

### Related issue(s)/PR(s)

Resolves #20842

🚀

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Roles/components/PluginsAndSettings.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Roles/components/PluginsAndSettings.tsx
@@ -91,6 +91,8 @@ const Row = ({
   const { formatMessage } = useIntl();
 
   const categoryName = name.split('::').pop() ?? '';
+  const categoryDisplayName =
+    categoryName === 'upload' ? 'Media Library' : capitalise(categoryName.replace(/-/g, ' '));
 
   return (
     <Accordion.Item value={name}>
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/pages/Settings/pages/Roles/components/PluginsAndSettings.tsx
```diff
@@ -91,6 +91,8 @@ const Row = ({
   const { formatMessage } = useIntl();
 
   const categoryName = name.split('::').pop() ?? '';
+  const categoryDisplayName =
+    categoryName === 'upload' ? 'Media Library' : capitalise(categoryName.replace(/-/g, ' '));
```
markkaylor, line 95: Surprised to see that none of these are translated

Adzouz: I was surprised too... I can do it if you want? For the moment it's only using the plugin name extracted from `plugin::{NAME}` which is not ideal. Wdyt?

markkaylor: I think merge this PR for now we can create another issue/pr for the translations

PR #23741, opened 2025-06-13
fix: min/max value for components not clear

### LINKED ISSUE #20495: [v5] Repeatable component Min/Max values inaccurate

Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0-beta.8
Node/Yarn Version: yarn/4.0.1 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite

### Describe the bug

When creating a repeatable component, on the Advanced Settings page, there are settings called "Maximum Value" and "Minimum Value" that do not make sense.

I tested, and it actually means the maximum and minimum number of components allowed to be created

(although, perhaps another bug is that when you drop below "minimum" by deleting components, it allows you to save but ignores all the deletions, but I'll wait to see what decision is made here before filing that one)

See description of bug

### Expected behavior

The text should match what the field actually means

### Screenshots

image

### Code snippets

Adzouz:
### What does it do?

Changed the title of the min / max value for repeatable components since the default meaning is confusing in that specific case.
Kept all logic in place and overridden only the title wording.

Screenshot 2025-06-13 at 14 44 10

### Why is it needed?

It will reduce the confusion of the user about the purpose of the min / max value in the case of a repeatable component.

### How to test it?

- Go to the admin interface > Content Type Builder
- Choose or create a type that has a repeatable component
- Create or edit the repeatable component 
- Go to Advanced settings tab

### Related issue(s)/PR(s)

Resolves #20495 by @innerdvations 

🚀


tests/e2e/tests/review-workflows/settings.spec.ts
```diff
@@ -139,11 +139,11 @@ describeOnCondition(edition === 'EE')('settings', () => {
...
     await expect(requiredStageCombobox.locator('span').first()).toHaveText('Done');
 
-    // Navigate to Content Manager to check validation is working
-    await page.getByRole('link', { name: 'Content Manager' }).click();
-
     // Go to the entry
     await page.getByRole('link', { name: 'Content Manager' }).click();
+    await page.waitForLoadState('networkidle');
+
+    await page.getByRole('link', { name: 'Author' }).waitFor({ state: 'visible' });
```
markkaylor: You might be able to replace all this with the clickAndWait util

markkaylor (dismissed):
Tested, try the clickAndWait util otherwise lgtm

tests/e2e/tests/review-workflows/settings.spec.ts
Adzouz: Done! I don't know if it will fix this test but I always see it failing and I wanted to improve that. 🤞

markkaylor (approved):

PR #23749, opened 2025-06-16
fix: update of logo

### LINKED ISSUE #20755: Menu logo not applied
When you try to update the logo in the Overview settings the logo is not applied even if you have a successful message. It appears only after a refresh.

https://github.com/user-attachments/assets/ba9e6cac-7e82-46e6-9d74-0606a7b4713f

Adzouz:
### What does it do?

Fix the update of the logo after sending the request due to wrong value of permission canRead in configuration.

https://github.com/user-attachments/assets/821019e2-27f0-442d-ba26-a3e3d504e5d9

### Why is it needed?

When clicking on save, the old logo is displayed giving the illusion to the user that the logo wasn't updated (even if it was the case).

### How to test it?

- Go to admin interface > Settings
- Scroll to Customization section
- Click on the + icon to change the logo
- Load a new image from your computer
- Click on Save

### Related issue(s)/PR(s)

Resolves #20755

🚀

### DIFF
--- a/examples/getstarted/config/features.js
+++ b/examples/getstarted/config/features.js
@@ -1,3 +1,5 @@
 module.exports = ({ env }) => ({
-  future: {},
+  future: {
+    unstableGuidedTour: false,
+  },
 });
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
@@ -97,9 +97,12 @@ const UnstableGuidedTourTooltip = ({
     requiredActions?.every((action) => {
       return guidedTourMeta?.data?.completedActions.includes(action);
     }) ?? true;
-
+  const hasFutureFlag = window.strapi.future.isEnabled('unstableGuidedTour');
   const isEnabled =
-    guidedTourMeta?.data?.isFirstSuperAdminUser && !state.tours[tourName].isCompleted;
+    guidedTourMeta?.data?.isFirstSuperAdminUser &&
+    !state.tours[tourName].isCompleted &&
+    hasFutureFlag;
+
   const isPopoverOpen = isEnabled && isCurrentStep && hasCompletedRequiredActions;
 
   // Lock the scroll
--- a/packages/core/types/src/modules/features.ts
+++ b/packages/core/types/src/modules/features.ts
@@ -1,5 +1,7 @@
 export interface FeaturesConfig {
-  future?: object;
+  future?: {
+    unstableGuidedTour?: boolean;
+  };
 }
 
 export interface FeaturesService {
@@ -9,5 +11,6 @@ export interface FeaturesService {
   config: FeaturesConfig | undefined;
   future: {
     isEnabled: (futureFlagName: string) => boolean;
+    unstableGuidedTour?: boolean;
   };
 }


simotae14 (changes requested):
I still have the issue

https://github.com/user-attachments/assets/6f2f2e75-d97d-4e69-a655-ef6b8d39693a

Also it is not clear to me why the destructuring of canRead doesn't work...and when you find a solution please add a test to prevent having regression errors in the future, thanks

packages/core/admin/admin/src/features/Configuration.tsx
```diff
@@ -76,9 +76,7 @@ const ConfigurationProvider = ({
   );
   const token = useAuth('ConfigurationProvider', (state) => state.token);
 
-  const {
-    allowedActions: { canRead },
-  } = useRBAC(permissions);
+  const { allowedActions } = useRBAC(permissions);
```
remidej: I haven't tested the fix but this looks like the only change is you made is to no longer destructure canRead right? I don't see how that would change the runtime behavior

Adzouz: I don't either but when I logged allowedActions, canRead was true but if I logged directly canRead, it was false for some reason 🤷‍♂️ I'll continue investigating

Adzouz:
I put back the logic of permissions since the issue was elsewhere. I stored the result of the updated image and use it in the rendering. Can you please tell me if it's better now @simotae14?

packages/core/admin/admin/src/features/Configuration.tsx
```diff
@@ -159,19 +163,23 @@ const ConfigurationProvider = ({
...
-            : {
-                url: customAuthLogo ?? '',
-              },
+          custom: authLogoUpdated
+            ? authLogoUpdated
+            : isSuccess
+              ? data?.authLogo
+              : {
+                  url: customAuthLogo ?? '',
+                },
```
remidej: we try to avoid nested ternaries as they're not super readable

packages/core/admin/admin/src/features/Configuration.tsx
```diff
@@ -67,6 +67,8 @@ const ConfigurationProvider = ({
   defaultMenuLogo,
   showReleaseNotification = false,
 }: ConfigurationProviderProps) => {
+  const [authLogoUpdated, setAuthLogoUpdated] = React.useState<Partial<Logo> | null>(null);
+  const [menuLogoUpdated, setMenuLogoUpdated] = React.useState<Partial<Logo> | null>(null);
```
remidej: I understand what you're doing, but I think it's worth taking a step back here and considering a refactor.

I don't really like the way the existing code works by fetching data with RTK and exposing it via context. It adds complexity and it's error prone. The app logo is server state, so we should probably rely on the RTK Query cache to manage it just like we manage the rest of the app's server state.

in my opinion, the preferred workflow would be:
- the left menu (or NavBrand directly) accesses the logo via useInitQuery. It should be available in the cache by the time the menu renders so no loading state required
- invalidate the init query cache tag when the user uploads a new logo

Adzouz: Good call, I updated the query and provided tags so that it could be invalidated on image update. It should work as expected. I also updated the image from the NavBrand to use the one from the same query.

packages/core/admin/admin/src/components/MainNav/NavBrand.tsx
```diff
@@ -16,21 +16,21 @@ const BrandIconWrapper = styled<FlexComponent>(Flex)`
 
 export const NavBrand = () => {
   const { formatMessage } = useIntl();
-  const {
-    logos: { menu },
-  } = useConfiguration('LeftMenu');
+  const { data: { menuLogo } = {} } = useInitQuery();
```
remidej: Thanks for the change!

I think you should now be able to remove the menuLogo code from the Configuration file. Ideally we'd remove the logos key from the configuration context altogether, but this requires changes on the login page so it doesn't need to be in this PR

Adzouz: Ok so after we discussed, I revert to what we had before because even if it's not ideal, configuration is already using RTK query. The only thing is that the init query wasn't invalidated because it was missing the providesTags. Let me know if anything!

remidej (approved):
love a simple fix!

simotae14 (approved):

Issue #23056, opened 2025-03-04
Publish multiple locales broken with PostgreSQL - reports present required value as missing


breezertwo:
### Node Version

22.14.0

### NPM/Yarn/PNPM Version

npm 10.9.2

### Strapi Version

5.6.0 - latest 

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

After updating to any version beyond 5.5.2 the "Publish multiple locales" feature appears to be broken for us. 
This seems to be happing if a field is required, for optional fields this does not happen.

The variable (e.g thumbnail) is present in all locales, and already published, but Strapi shows an error, that I can't publish one locale because of a missing required field:
![Image](https://github.com/user-attachments/assets/e5889f32-38f4-459d-b29c-72ec6bbed7d0)

The thumbnail IS set for all locales.

### Steps to Reproduce

1. Update to any version greater then 5.5.2 of strapi
2. Go into details of a existing content type
3. Change something
4. Select publish multiple locales 
5. See can't publish as a required variable is marked as missing despite being present

### Expected Behavior

1. Update to any version greater then 5.5.2 of strapi 
2. Go into details of a existing content type
3. Change something
4. Select publish multiple locales 
5. Be able to publish all locales that do have the required variables

### Additional information

I tried several versions. (e.g 5.6, 5.8) and all seem to have this issue. If I downgrade to 5.5.2 again, I am able to use the multiple publish button again. 

Publishing locales by them self directly does work, with the same content that does not work with multiple locales.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

martin-luo:
I experienced the same issue using postgresql too.

Strapi: 5.11.0

markkaylor:
I'm able to reproduce the issue with sqlite as well. Currently only able to reproduce with a required media attribute (required text for example does not have the issue)

breezertwo:
@markkaylor So happy to see this ticket in progress 🙏 

If I can help with anything just let me know.

PR #23750, opened 2025-06-16
fix: bulk publish mutliple locales does not validate required media

markkaylor:
### What does it do?

Ensures required media are populated for validation across locales

### Why is it needed?

Bulk publish in multiple locales does not validate since the localization attribute is not populated correctly

### How to test it?

Use a content-type with i18n, draft and publish, and a required media attribute
Create a document with a draft version in two languages (don't add the image)
Click the 3 dots and then Publish in multiple locales
In the modal you should see validation failing for both since there is no image
Go update the drafts for both to include an image
Now try to publish multiple locales again
There should be no validation errors and you should be able to publish in both locales

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/23056

### DIFF
--- a/packages/core/content-manager/server/src/services/utils/__tests__/validatable-fields-populate.test.ts
+++ b/packages/core/content-manager/server/src/services/utils/__tests__/validatable-fields-populate.test.ts
@@ -38,6 +38,15 @@ describe('getPopulateForValidation', () => {
         },
       },
     },
+    media: {
+      modelName: 'Fake media model',
+      attributes: {
+        mediaAttrName: {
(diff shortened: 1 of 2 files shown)


innerdvations (approved):
Code LGTM, I didn't test it though

markkaylor:
@simotae14 @Adzouz either of you available to QA this one?

simotae14:
I tested your PR and it works!!! thanks a lot for this bugfix @markkaylor

simotae14 (approved):
🥇

PR #23770, opened 2025-06-18
future: add guided tour provider and tour factory

markkaylor:
### What does it do?

Add the tour factory function and guided tour context provider for the new guided tour

### Why is it needed?

To make it possible to create guided tours in Strapi

### How to test it?

Read the documentation added here for to create tours and then consume them in the CMS
Do something like this anywhere you want in the CMS: 
```
<unstable_tours.contentManager.TEST>ANCHOR</unstable_tours.contentManager.TEST>
```
You should see this displayed:

Screenshot 2025-06-18 at 15 13 29

Otherwise there are unit tests for the tour reducer

NOTE:

We can safely merge this without a future flag. There is no impact on the user unless they try to use the exported `unstable_tours` but it is clearly marked as `unstable`

### DIFF
--- a/docs/docs/guides/06-guided-tour.md
+++ b/docs/docs/guides/06-guided-tour.md
@@ -1,0 +1,63 @@
+---
+title: Guided Tour
+---
+
+This document explains how to create and use Guided Tours in the Strapi CMS.
+
+## Creating tours
+
+To create a tour use the `createTour` factory function.
+
+```tsx
+const tours = {
+  contentManager: createTour('contentManager', [
+    {
+      name: 'TheFeatureStepName',
+      content: () => (
+        <>
+          <div>This is the content for Step 1 of some feature</div>
+        </>
+      ),
+    },
+  ]),
+} as const;
+```
+
+Tours for the CMS are defined in the `packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx` file.
+
+The tours are then passed to the `UnstableGuidedTourContext` provider.
+
+```tsx
+import { tours } from '../UnstableGuidedTour/Tours';
+import { UnstableGuidedTourContext } from '../UnstableGuidedTour/Context';
+
+function App() {
+  return (
+    <UnstableGuidedTourContext tours={tours}>
+      <Outlet />
+    </UnstableGuidedTourContext>
+  );
+}
+```
+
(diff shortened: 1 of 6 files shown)


docs/docs/guides/06-guided-tour.md
```diff
@@ -0,0 +1,63 @@
...
+The tours are then passed to the `UnstableGuidedTourContext` provider.
+
+```tsx
+import { tours } from '../UnstableGuidedTour/Tours';
+import { UnstableGuidedTourContext } from '../UnstableGuidedTour/Context';
+
+function App() {
+  return (
+    <UnstableGuidedTourContext tours={tours}>
+      <Outlet />
```
simotae14, line 37: Hey, what is this Outlet component?

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -0,0 +1,78 @@
...
+/* -------------------------------------------------------------------------------------------------
+ * Tour factory
+ * -----------------------------------------------------------------------------------------------*/
+
+type TourStep<P extends string> = {
+  name: P;
+  content: Content;
+};
+
+type Content = ({
```
simotae14, line 31: just for a sense of readability, isn't it better to declare the Content type before the TourStep type?

docs/docs/guides/06-guided-tour.md
markkaylor: https://api.reactrouter.com/v7/functions/react_router.Outlet.html but here it could be App or anything really since this is just an example

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
markkaylor: I think things are going to get moved around a lot so personally wouldn't worry about it right now but if you prefer I move it I will just let me know

simotae14: I think it it a personal opinion, we can stay like this atm

simotae14 (approved):
Overall the code looks good and easy to follow, thanks for this first step @markkaylor

PR #23771, opened 2025-06-18
fix: design of user status

### LINKED ISSUE #21343: [Design] Unnecessary green border in the Admin Users list
It’s a DS issue, meaning we’ll have to override its style (it’s certainly how we did it everywhere else in the admin). Overall we have too many different tags and badges, it would be nice to harmonize them someday.

Image

Adzouz:
### What does it do?

Fixing the design of the user's status in the list in Settings > Administration Panel > Users to match other badges in the CMS.

Before:

Screenshot 2025-06-18 at 16 09 54
Screenshot 2025-06-18 at 16 10 23

After:

Screenshot 2025-06-17 at 16 06 56
Screenshot 2025-06-17 at 16 07 12

_There's also an initiative to revamp the badges in the design system to unify them on the CMS here: https://github.com/strapi/design-system/issues/1935_

### Why is it needed?

To unify the experience of the badges and their meaning.

### How to test it?

- Go to the admin interface > Settings > Administration panel > Users
- The tags are displayed in each element of the list (Active / Inactive)

### Related issue(s)/PR(s)

Resolves #21343

🚀

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
@@ -288,14 +288,10 @@ const TABLE_HEADERS: Array<
     cellFormatter({ isActive }) {
       return (
         <Flex>
-          <Status
-            size="S"
-            borderWidth={0}
-            background="transparent"
-            color="neutral800"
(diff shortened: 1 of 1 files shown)


simotae14 (approved):
well done

PR #23776, opened 2025-06-19
future: add guided tour tooltip and step composable components

markkaylor:
### What does it do?

- Adds the GuidedTourTooltip
- Adds the Step factory and passes Step to GuidedTourTooltip content prop

### Why is it needed?

- To display the tooltip for the current step
- To add the content for each step with the correct styles and actions

### How to test it?

Import the `unstable_tours` and wrap any element. For example in the content manager list view edit page EmptyStateLayout

```tsx
<unstable_tours.contentManager.Introduction>
  <EmptyStateLayout
    action={canCreate ? <CreateButton variant="secondary" /> : null}
    content={formatMessage({
      id: 'app.components.EmptyStateLayout.content-document',
      defaultMessage: 'No content found',
    })}
    hasRadius
    icon={<EmptyDocuments width="16rem" />}
  />
</unstable_tours.contentManager.Introduction>
```

You should see

Screenshot 2025-06-19 at 13 51 31

NOTE: 

This PR can be safely merged without a future flag. There is no impact on the user unless they choose to import and use the `unstable_tours` which are clearly marked as `unstable`

### DIFF
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
@@ -69,7 +69,6 @@ const UnstableGuidedTourTooltip = ({
   tourName: ValidTourName;
   step: number;
 }) => {
-  useAuth;
   const state = unstableUseGuidedTour('UnstableGuidedTourTooltip', (s) => s.state);
   const dispatch = unstableUseGuidedTour('UnstableGuidedTourTooltip', (s) => s.dispatch);
   const Step = React.useMemo(() => createStepComponents(tourName), [tourName]);


packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
```diff
@@ -10,49 +10,66 @@ import type { Tours } from './Tours';
...
   });
 }
 
 const UnstableGuidedTourContext = ({
   children,
-  tours,
+  tours: registeredTours,
 }: {
   children: React.ReactNode;
+  // NOTE: Maybe we just import this directly instead of a prop?
```
simotae14, line 55: if the content inside tours is static I agree with you, we can directly import it

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -7,18 +15,91 @@ import type { State, Action } from './Context';
...
+  content,
+  tourName,
+  step,
+}: {
+  children: React.ReactNode;
+  content: Content;
+  tourName: ValidTourName;
+  step: number;
+}) => {
+  useAuth;
```
simotae14: I don't understand this line of code

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -7,18 +15,91 @@ import type { State, Action } from './Context';
...
+    state: State;
+    dispatch: React.Dispatch<Action>;
+  }
+) => React.ReactNode;
+
+export const GuidedTourOverlay = styled(Box)`
+  position: fixed;
+  inset: 0;
+  background-color: rgba(50, 50, 77, 0.2);
+  z-index: 10;
```
simotae14, line 57: I am curios, do you have already checked if this z-index is in conflict with others element with the same value?

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
markkaylor: 😄 good catch that's a mistake

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
markkaylor: I've only been testing on the content manage list page so far and there is no issue. The overlay zIndex should be higher than every thing except popover.

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
simotae14: ok so you can remove also the import

simotae14:
It works!!! well done
Screenshot 2025-06-20 alle 15 01 42

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
markkaylor: Gonna leave it for now but we will probably change it later

PR #23786, opened 2025-06-20
future: add guided tour tooltip and step composable components

markkaylor:
### What does it do?

- Adds an admin endpoint to fetch meta data for the guided tour

Currently this is only fetching whether or not the user isFirstAdminUser but we will likely have more in the future for the auto-detect feature which determines whether or not the guided tour should run based on the app state. This is currently postponed and may be handled in another iteration. 

### Why is it needed?

To ensure we only display the guided tour for the correct users

### How to test it?

Import the `unstable_tours` and wrap any element. For example in the content manager list view edit page EmptyStateLayout

```tsx
<unstable_tours.contentManager.Introduction>
  <EmptyStateLayout
    action={canCreate ? <CreateButton variant="secondary" /> : null}
    content={formatMessage({
      id: 'app.components.EmptyStateLayout.content-document',
      defaultMessage: 'No content found',
    })}
    hasRadius
    icon={<EmptyDocuments width="16rem" />}
  />
</unstable_tours.contentManager.Introduction>
```

If you are the only super admin you should see
Go create another super admin and log in as that one, you should not see the tooltip
Create another user that isn't a super admin and login as that user, you show not see the tooltip

NOTE: 

This PR can be safely merged without a future flag. It has not impact on the user unless the import and consume the `unstable_tours`


packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -185,4 +186,46 @@ export default {
...
+      (token) => token.name
+    );
+    const didCreateApiToken = apiTokenNames.some((name) => !DEFAULT_API_TOKENS.includes(name));
+
+    return {
+      data: {
+        isFirstSuperAdminUser,
+        didCreateContentTypeSchema,
+        didCreateContent,
+        didCreateApiToken,
```
markkaylor: ℹ️  Maybe we should store these on a key: `requiredActions`?

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -185,4 +186,46 @@ export default {
...
+        .count({});
+      if (docs > 0) {
+        hasContent = true;
+        break;
+      }
+    }
+    const didCreateContent = didCreateContentTypeSchema && hasContent;
+
+    // Check if any api tokens have been created besides the default ones
+    const DEFAULT_API_TOKENS = ['Read Only', 'Full Access'];
```
markkaylor: ℹ️ I couldn't really find a better way to determine default api tokens unless I add a property to the api tokens model

packages/core/admin/server/src/services/user.ts
```diff
@@ -161,6 +161,30 @@ const isLastSuperAdminUser = async (userId: Data.ID): Promise<boolean> => {
   return superAdminRole.usersCount === 1 && hasSuperAdminRole(user);
 };
 
+/**
+ * Check if a user is the last super admin
```
markkaylor:
```suggestion
 * Check if a user is the first super admin
```

markkaylor:
Looking into the failing tests

packages/core/admin/server/src/services/user.ts
```diff
@@ -161,6 +161,30 @@ const isLastSuperAdminUser = async (userId: Data.ID): Promise<boolean> => {
...
+  const [oldestUser] = await strapi.db.query('admin::user').findMany({
+    populate: {
+      roles: {
+        where: {
+          code: { $eq: SUPER_ADMIN_CODE },
+        },
+      },
+    },
+    orderBy: { createdAt: 'asc' },
+    limit: 1,
```
remidej, line 182: you could select just the id field

packages/core/admin/shared/contracts/admin.ts
```diff
@@ -218,3 +219,24 @@ export declare namespace GetLicenseLimitInformation {
...
   }
 }
+
+/**
+ * Meta data for the guided tour
+ */
+export declare namespace GetGuidedTourMeta {
+  export interface Request {
+    query: {
+      id?: Data.ID;
```
remidej: that's the user id right? I don't understand why it's needed. Don't you have it via the koa context?

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -185,4 +186,46 @@ export default {
...
+    let hasContent = false;
+    for (const contentTypeSchemaName of contentTypeSchemaNames) {
+      const docs = await strapi
+        .documents(contentTypeSchemaName as Internal.UID.ContentType)
+        .count({});
+      if (docs > 0) {
+        hasContent = true;
+        break;
+      }
+    }
```
remidej: I think we should try to follow our more functional style habites. Maybe use Array.some instead? (You may have picked your current approach to be able to break out of the loop early but I wonder if you tried different things)

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -185,4 +186,46 @@ export default {
...
     return data;
   },
+
+  async getGuidedTourMeta(ctx: Context) {
+    const userService = getService('user');
+    const { id } = ctx.query as GetGuidedTourMeta.Request['query'];
+
+    const isFirstSuperAdminUser = id ? await userService.isFirstSuperAdminUser(id) : false;
+
+    // Check if any content-type schemas have been create on the api:: namespace
```
remidej: i think most of this controller logic belongs in a service

packages/core/admin/server/src/controllers/admin.ts
remidej: i don't mind the simplicity of what you currently have 👍

packages/core/admin/admin/src/services/apiTokens.ts
```diff
@@ -4,7 +4,7 @@ import { adminApi } from './api';
 
 const apiTokensService = adminApi
   .enhanceEndpoints({
-    addTagTypes: ['ApiToken'],
+    addTagTypes: ['ApiToken', 'GuidedTourMeta'],
```
remidej, line 7: I'm not super clear on how adding tags work, but this doesn't feel like it belongs here. Also you have it in services/admin.ts too

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
```diff
@@ -55,23 +79,44 @@ const UnstableGuidedTourContext = ({
...
+
   const [state, dispatch] = React.useReducer(reducer, {
     tours,
+    completedActions: [],
+    meta: guidedTourMeta?.data ?? {
+      isFirstSuperAdminUser: false,
+      didCreateContentTypeSchema: false,
+      didCreateContent: false,
+      didCreateApiToken: false,
+    },
```
remidej: could we merge meta and completedActions? is the goal of having them separated to have the modified data?

packages/core/admin/shared/contracts/admin.ts
markkaylor: Oh yeah that's right I forgot about that 🤦‍♂️, thanks I'll update

packages/core/admin/server/src/controllers/admin.ts
markkaylor: I didn't try `some` but yes my goal was to break out of the loop as soon as possible, other than what I have now I was working with `reduce`. I agree I would favor the functional approach but I couldn't find a way to do to exit early that way. I can give it another look.

packages/core/admin/admin/src/services/apiTokens.ts
markkaylor: I'm not really clear on it either, but without it TypeScript wasn't happy. I added it sort of blindly to satisfy TS, I'll look into how it actually works.

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
markkaylor: I did consider it but I think I just liked meta in the context being meta that's received from the server 🤷‍♂️ . I don't have a strong opinion if you prefer it in the meta object. What do you mean by "modified data"?

packages/core/admin/admin/src/services/apiTokens.ts
markkaylor: The documentation doesn't have an example showing separate api's like we have: https://redux-toolkit.js.org/rtk-query/api/createApi#tagtypes

But I believe this is the correct usage. We aren't adding tags here, the tagTypes refers to all the tag strings that can be passed to providesTags or invalidatesTags in the given API.

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
remidej: when do you think you'll use the distinction of a completed (or not) action coming from the server or not?

packages/core/admin/server/src/controllers/admin.ts
markkaylor: I considered it and I will create the service to follow convention but just to share my thinking:
- Seemed like overkill to create a service to have one function for this logic
- The service doesn't have much benefit beyond scoping/isolating the code. We won't write a unit test for it since we have the api test, and we don't need to expose it on the strapi instance.

remidej: I really wish we had a quicker way to colocate everything related to an endpoint. It's just that we're forced into extracting routes and controllers, so it feels weird to not extract the service too, but it's probably just me not being used to it.

So I definitely hear you and you can skip it if this is gonna be the only method in the controller that has new logic

markkaylor: I did it and I don't regret it. Still probably the cleanest thing to do and follows convention 🫡

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -55,26 +70,37 @@ export const GuidedTourOverlay = styled(Box)`
...
   step,
+  requiredActions,
 }: {
   children: React.ReactNode;
   content: Content;
   tourName: ValidTourName;
   step: number;
+  requiredActions?: GetGuidedTourMeta.Response['data']['completedActions'];
 }) => {
+  const { data: guidedTourMeta } = useGetGuidedTourMetaQuery();
```
markkaylor, line 88: I think we can move forward with this here but I did just realize that eventually we will want to determine the current step of each tour based on the completedActions

tests/api/core/admin/admin-guided-tour-meta.test.api.ts
```diff
@@ -0,0 +1,145 @@
...
+      });
+
+      expect(res.status).toBe(200);
+      expect(res.body.data).toMatchObject({
+        isFirstSuperAdminUser: true,
+        completedActions: [],
+      });
+    });
+
+    test.only('Detects first super admin user', async () => {
```
markkaylor: oops!

markkaylor:
Not sure what's up with the tests they pass locally for all databases

markkaylor:
Tests are failing due to leaked data from another test

markkaylor:
@remidej I skipped the test that is failing due to leaked data from tests/api/core/strapi/components/cleanup-after-delete.test.api.ts. I couldn't find a quick fix to properly cleanup and it feels like that test isn't testing what it says it's testing since it is testing deletion and cleanup. I've asked @innerdvations for help. 

I'd like to move forward with this PR and investigate that separately to not hold up the guided tour dev.

packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -185,4 +185,16 @@ export default {
 
     return data;
   },
+
+  async getGuidedTourMeta(ctx: Context) {
+    const isFirstSuperAdminUser = await getService('user').isFirstSuperAdminUser(ctx.state.user.id);
+    const completedActions = await getService('guided-tour').getCompletedActions();
```
remidej: optional but these could be run in parallel

innerdvations:
> it feels like that test isn't testing what it says it's testing since it is testing deletion 

It's testing the deletion of certain data in those tables, not the deletion of the tables themselves

I think the problem is in the test cleanup, either a bug in builder.cleanup, or that it's being used incorrectly. Can probably be worked around for now with something like:

```typescript
afterAll() {
  const knex = strapi.db.connection;
  await knex.schema.dropTableIfExists('contenttypewithcomponents');
  await knex.schema.dropTableIfExists('contenttypewithcomponents_cmps');

  await strapi.destroy();
  await builder.cleanup();
}
```

packages/core/admin/server/src/services/guided-tour.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        if (count > 0) return true;
+      }
+
+      return false;
+    })();
+    const didCreateContent = didCreateContentTypeSchema && hasContent;
+
+    // Check if any api tokens have been created besides the default ones
+    const DEFAULT_API_TOKENS = ['Read Only', 'Full Access'];
```
remidej: did you check if these are localised? also is there a way to import these constants to avoid matching errors?

markkaylor: The api token content-type is not localized. As far as I can tell there is no existing constant for the names used to create the default api tokens.

markkaylor: Gonna create a shared constant for the API tokens

remidej: if it's possible to extract the constant that would be neat, if it's annoying let's not bother

packages/core/admin/server/src/bootstrap.ts
```diff
@@ -88,21 +89,8 @@ const createDefaultAPITokensIfNeeded = async () => {
...
-
-    await apiTokenService.create({
-      name: 'Full Access',
-      description:
-        'A default API token with full access permissions, used for accessing or modifying resources',
-      type: 'full-access',
-      lifespan: null,
-    });
+    await apiTokenService.create(constants.DEFAULT_API_TOKENS[0]);
+    await apiTokenService.create(constants.DEFAULT_API_TOKENS[1]);
```
remidej: nitpick but I'd prefer to loop on default tokens here in case we add/remove some in the future

packages/core/admin/server/src/bootstrap.ts
```diff
@@ -89,8 +89,11 @@ const createDefaultAPITokensIfNeeded = async () => {
   const apiTokenCount = await apiTokenService.count();
 
   if (usersCount === 0 && apiTokenCount === 0) {
-    await apiTokenService.create(constants.DEFAULT_API_TOKENS[0]);
-    await apiTokenService.create(constants.DEFAULT_API_TOKENS[1]);
+    await Promise.all(
+      Object.values(constants.DEFAULT_API_TOKENS).map((token) => {
+        return apiTokenService.create(token);
+      })
+    );
```
remidej: Does Object values do anything? Isn't DEFAULT_API_TOKENS an array already?

I'm also not sure we should Promise.all here as it means the tokens could be created in any order, so it will affect the UI

markkaylor: Oh yeah shit the Object values thing is a mistake. For the promise.all happy to remove it but I don't think order really matters.

remidej: definitely not a big deal but it could swap the order in the ui, so for example the [docs](https://docs.strapi.io/cms/features/api-tokens#creating-a-new-api-token) screenshot would not be accurate

remidej (approved):

Issue #21964, opened 2024-10-24
No relations available


danchistyakov:
### Required System information

- Node.js version: 18
- Strapi version: 5.1.1
- Database: Postgres
- Operating system: Ubuntu 22
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When you try to search relations by numeric IDs I'm not getting them ("No relations available"). Moreover search suggestions are documentIds instead of numeric ids. Searching by documentId is not working too. This is a critical bug.

### Steps to reproduce the behavior

1. Go to content manager
2. Click on any relation
3. Try to type numeric ID
4. See "No relations available"

### Expected behavior
I see the list of relations with numeric ids

### Screenshots
<img width="1462" alt="Снимок экрана 2024-10-24 в 11 52 34" src="https://github.com/user-attachments/assets/3d154207-7e9f-408c-9cb2-62ee9cf5c0d6">
<img width="1470" alt="Снимок экрана 2024-10-24 в 11 54 42" src="https://github.com/user-attachments/assets/6d2ccffb-a6ec-4bd9-a3ed-bd93f60d8510">
<img width="1469" alt="Снимок экрана 2024-10-24 в 11 50 53" src="https://github.com/user-attachments/assets/5aabba94-f1f1-4aa6-89e6-27d408674791">

derrickmehaffy:
We have a very strict definition of critical and this isn't it, but I do agree it would be high

Flowrome:
What is the status of this? also the documentation plugin also is broken as you can't use documentid in the openapi generated swagger, and this is critical due to the fact we're using openapi for generating client api requests

kasonde:
Confirming this issue as well as referencing Internal TID 8656

PR #23796, opened 2025-06-23
fix: not searchable relation by document Id

simotae14:
### What does it do?

It fixes the issue when you try to search relations by document Id you receive always the message "No relations available"

### Why is it needed?

It is impossible to search for a relation when the mainField is the documentId

### How to test it?

- Go to content manager
- Open a document that contains a Relation with the document Id as main field
- Try to filter the relation options by document id  and check if you are able to do it
- and the message "No relations available" is not always shown

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/21964


remidej (approved):
It's working 👍 
screenshot 2025-07-01 at 16 19 58@2x

PR #23798, opened 2025-06-23
future: add api tokens guided tour

simotae14:
### What does it do?

- Adds the GuidedTour ApiTokens tour

### Why is it needed?

- To display the various steps
-  and trigger the redirect on the final step

### How to test it?

Access the content managers with the oldest super admin credentials
- Go to the Settings page
- Click the Api Tokens link
- follow the steps for the API Tokens guided tour

take a look at the figma file

### DIFF
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
@@ -2,12 +2,19 @@ import * as React from 'react';
 
 import { Box, Popover, Portal, Flex, LinkButton } from '@strapi/design-system';
 import { FormattedMessage } from 'react-intl';
+import { NavLink } from 'react-router-dom';
 import { styled } from 'styled-components';
 
 import { type GetGuidedTourMeta } from '../../../../shared/contracts/admin';
 import { useGetGuidedTourMetaQuery } from '../../services/admin';
 
-import { type State, type Action, unstableUseGuidedTour, ValidTourName } from './Context';
+import {
+  type State,
+  type Action,
+  unstableUseGuidedTour,
+  ValidTourName,
+  ExtendedCompletedActions,
+} from './Context';
 import { Step, createStepComponents } from './Step';
 
 /* -------------------------------------------------------------------------------------------------
@@ -108,6 +115,7 @@ const tours = {
           <Step.Actions showSkip />
         </Step.Root>
       ),
+      when: (completedActions) => !completedActions.includes('didCreateApiToken'),
     },
     {
       name: 'CreateAnAPIToken',
@@ -128,7 +136,7 @@ const tours = {
     {
       name: 'CopyAPIToken',
       content: (Step) => (
-        <Step.Root side="bottom" align="start" sideOffset={-10}>
+        <Step.Root side="bottom" align="start">
           <Step.Title
             id="tours.apiTokens.CopyAPIToken.title"
             defaultMessage="Copy your new API token"
@@ -140,6 +148,7 @@ const tours = {
           <Step.Actions />
         </Step.Root>
       ),
+      when: (completedActions) => completedActions.includes('didCreateApiToken'),
     },
     {
       name: 'Finish',
@@ -161,7 +170,8 @@ const tours = {
                   onClick={() => {
                     dispatch({ type: 'next_step', payload: 'apiTokens' });
                   }}
-                  href="/"
+                  tag={NavLink}
+                  to="/"
                 >
                   <FormattedMessage id="tours.gotIt" defaultMessage="Got it" />
                 </LinkButton>
@@ -170,7 +180,7 @@ const tours = {
           </Step.Root>
         );
       },
-      when: (completedActions) => completedActions.includes('didCreateContentTypeSchema'),
+      when: (completedActions) => completedActions.includes('didCopyApiToken'),
     },
   ]),
 } as const;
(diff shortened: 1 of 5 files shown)


packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -92,15 +170,15 @@ const UnstableGuidedTourTooltip = ({
 
   const Step = React.useMemo(() => createStepComponents(tourName), [tourName]);
 
-  const isCurrentStep = state.tours[tourName].currentStep === step;
+  const isCurrentStep = state?.tours[tourName].currentStep === step;
```
markkaylor: Why the optional chaining?

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -92,15 +170,15 @@ const UnstableGuidedTourTooltip = ({
...
   const hasCompletedRequiredActions =
     requiredActions?.every((action) => {
       return guidedTourMeta?.data?.completedActions.includes(action);
     }) ?? true;
   const hasFutureFlag = window.strapi.future.isEnabled('unstableGuidedTour');
   const isEnabled =
     guidedTourMeta?.data?.isFirstSuperAdminUser &&
-    !state.tours[tourName].isCompleted &&
+    !state?.tours[tourName].isCompleted &&
```
markkaylor: Same question

packages/core/admin/admin/src/hooks/useMenu.ts
```diff
@@ -34,6 +37,7 @@ const useMenu = (shouldUpdateStrapi: boolean) => {
         to: '/',
         permissions: [],
         position: 0,
+        tourComponent: unstable_tours.apiTokens.Done,
```
markkaylor: I think I'm handling this on my PR fro the CTB: https://github.com/strapi/strapi/pull/23844/files#diff-6d7b9072c08214ee19a418aa4d023a5ae89f130470adcddef50671cbcbfd3866

would it work for you?

packages/core/admin/admin/src/pages/Settings/components/Tokens/TokenBox.tsx
```diff
@@ -85,4 +86,10 @@ export const TokenBox = ({ token, tokenType }: TokenBoxProps) => {
       iconBackground="neutral100"
     />
   );
+
+  return tokenType === 'api-token' ? (
```
markkaylor: See same comment with CTB link to my CTB PR. I think we could use a getter that returns a fragment by default

packages/core/admin/server/src/services/guided-tour.ts
```diff
@@ -32,7 +32,7 @@ export const createGuidedTourService = ({ strapi }: { strapi: Core.Strapi }) =>
     const createdApiTokens = await strapi
       .documents('admin::api-token')
       .findMany({ fields: ['name', 'description'] });
-    const didCreateApiToken = createdApiTokens.some((doc) =>
+    const didCopyApiToken = createdApiTokens.some((doc) =>
```
markkaylor: I don't understand? We changed the name but it has nothing to do with copying? The check is for creating. I think we will have to change some things in the context. I don't think this can come from the server. We will have to push to an array on the frontend.

markkaylor: We can look at this together

simotae14: yeah you are right, in my idea we need to save the action in the local storage (I mean didCreateApiToken) to be sure it start from the final step if we refresh the page

Btw I am gonna remove this logic because it is useless

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
simotae14: otherwise I have an error, you can try to disable the future flag and go to the Api tokens page

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
simotae14: otherwise I have an error, you can try to disable the future flag and go to the Api tokens page

packages/core/admin/server/src/services/guided-tour.ts
markkaylor: Let's discuss first

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
markkaylor: We should look together. I don't have any errors on the CTB PR for this

packages/core/admin/admin/src/pages/Settings/components/Tokens/TokenBox.tsx
```diff
@@ -35,54 +38,67 @@ export const TokenBox = ({ token, tokenType }: TokenBoxProps) => {
     }
   };
 
+  const getGuidedTourTooltip = (tokenType: string) => {
+    switch (tokenType) {
+      case 'api-token':
+        return unstable_tours.apiTokens.CopyAPIToken;
```
markkaylor: Do you need the getter function here? Couldn't you just use unstable_tours.apiToken.copyAPIToken directly?

simotae14: ahahah I did it because of a comment I received from you https://github.com/strapi/strapi/pull/23798#discussion_r2179233893, I need to differentiate between api-token and trasnfer-token

markkaylor: oh my bad, If it's possible to wrap it directly then that is the best option but I don't think that's what you were doing before either

simotae14 (commented):
I have just a doubt regarding the required action for the api tokens because you trigger it when you copy an api token not when you create an api token

Other than that to me it seems everything correct

PR #23809, opened 2025-06-25
chore: track widgets usage

remidej:
### What does it do?

- remove the didInitializeAdministration event as be block it anyway on amplitude as it created too many events
- move the widgets tracking to didAccessAuthenticatedAdministration instead

### How to test it?

check your networks tab, find didAccessAuthenticatedAdministration, see the registeredWidgets property with an array of string uids


markkaylor (approved):
Code looks. Tested and it's working.

PR #23820, opened 2025-06-25
fix(core/admin): no warn if StrapiApp link no component, use correct useRBAC

unrevised6419:
### What does it do?

- No warn on `StrapiApp#addMenuLink` and `StrapiApp#addSettingsLink` missing component
- Remove some deprecated `useRBAC` calls that produce warnings
- Adjust `useRBAC` to report deprecated usage in IDE

### Why is it needed?

image

image

image

PS: After all `Object.values({}).flat()`, I'm not sure if it was worth deprecating object notation.

PS2: I'm thinking to split this PR in 2 more simple PRs

- [ ] no warn if StrapiApp link no component
- [ ] type useRBAC to show deprecated use of object notation

`useRBAC` changes to use array notation will be dropped, and let the devs stop using it.

### DIFF
--- a/packages/core/admin/admin/src/pages/Settings/pages/Users/EditPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Users/EditPage.tsx
@@ -18,7 +18,6 @@ import { useNotification } from '../../../../features/Notifications';
 import { useAPIErrorHandler } from '../../../../hooks/useAPIErrorHandler';
 import { useEnterprise } from '../../../../hooks/useEnterprise';
 import { useRBAC } from '../../../../hooks/useRBAC';
-import { selectAdminPermissions } from '../../../../selectors';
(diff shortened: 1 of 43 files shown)


packages/core/admin/admin/src/pages/Settings/pages/Users/EditPage.tsx
```diff
@@ -18,7 +18,6 @@ import { useNotification } from '../../../../features/Notifications';
 import { useAPIErrorHandler } from '../../../../hooks/useAPIErrorHandler';
 import { useEnterprise } from '../../../../hooks/useEnterprise';
 import { useRBAC } from '../../../../hooks/useRBAC';
-import { selectAdminPermissions } from '../../../../selectors';
```
unrevised6419, line 21: Note: This is a deprecated selector, and suggests to use inline it.

PR #23681, opened 2025-06-06
enhancement: debounce blocks editor update

remidej:
### What does it do?

Debounces the updates that keep the Strapi form state in sync with Slate's internal state.

>Slate handles the state of the editor internally. We just need to keep Strapi's form state in sync with it in order to make sure that things like the "modified" state isn't broken. Updating the whole state on every change is very expensive however, so we debounce calls to onChange to mitigate input lag.
### Why is it needed?

The blocks editor has way too much input lag. This helps a lot

### How to test it?

Type in the blocks editor. Things should feel snappier. Try on a content type with a lot of fields and a lot of content in particular.

You can use the devtools performance tab too, which gives you an Interaction to Next Paint (INP) metric

Also make sure the "discard changes" button still works for updates made in a blocks field.

### Related issue(s)/PR(s)

fixes #23208

Issue #22949, opened 2025-02-20
Error when discarding changes in rich text editor.


Hakim-Lakrout:
### Node Version

20.17.0

### NPM/Yarn/PNPM Version

NPM 10.8.2

### Strapi Version

5.8.1 and 5.10.3

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When writing text in `Rich text (Blocks)`  in a component and saving the changes then discard them and the component is open,an error occur.
More info in the screenshots and the video.
Also reproduced on MACOS Sequoya 15.3.1. Also with PostgreSQL 14 database.

### Steps to Reproduce

To reproduce

1. Create a component Text in Strapi with Type Rich text (Blocks).
2. Create a collection type that uses this component.
3. Create an instance of this collection type write some text in the text component and publish it.
4. After publishing it, modify the content by adding some text and save it.
5. discard the changes with the accordion (in which there is the text editor) uncollapsed.
**Note**
The bug don't appear when  accordion (in which there is the text editor) collapsed.

### Expected Behavior

Changes will be discarded without the error.

### Logs

```shell

```

### Media

Please see the video and the screenhhots for more informations.
- Reproduction of the Bug : 
https://github.com/user-attachments/assets/0848288a-02da-4acb-a5c2-701f2cdcc3a0
- Text component :  ![Image](https://github.com/user-attachments/assets/8635e46a-26c6-46aa-ae57-f7ee2f9afa0a)
- Collectiont type component :  ![Image](https://github.com/user-attachments/assets/005254d7-008f-4f27-8ce0-1390903e0f2b)
- Error message in Chrome browser (version : 133.0.6943.98) : ![Image](https://github.com/user-attachments/assets/19857cc7-3bc3-4f52-aefb-6857ce46df1b)
- Error message in Firefox  (version :135.0.1 ): ![Image](https://github.com/user-attachments/assets/40a48f11-11b9-4872-9d8e-cf2acecfac68)

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Hakim-Lakrout:
Hello, 
Here are some more informations.
I cloned the official GitHub repo of Strapi and i lunched the example app to debug the error.
[...]

PR #23822, opened 2025-06-26
Fix : Discarding changes error when text editor is collapsed

Hakim-Lakrout:
When adding text to rich text editor, and then discard the changes in Strapi an error occurs. The problem is in the anchor when the text is deleted after discarding it but the rendering still uses the old text position. An error is raised from different functions that verify the state of editor buttons. This issue was fixed in  #23681 but this PR actually suggests another solutions with adequate unit tests.

### What does it do?

Fix the issue #22949 by verifying the anchor index when discarding changes and replace it to last valid index of the text editor.

### Why is it needed?

Described in issue #22949

### How to test it?

Described in issue #22949, this changes fix the issue and does not raise the error.

### Related issue(s)/PR(s)

Issue #22949


markkaylor:
>This issue was fixed in https://github.com/strapi/strapi/pull/23681 but this PR actually suggests another solutions with adequate unit tests.

If the issue is fixed is what your are proposing here and alternate solution to the same problem or an addition for specific edge case?

Hakim-Lakrout:
> > This issue was fixed in #23681 but this PR actually suggests another solutions with adequate unit tests.
> 
> If the issue is fixed is what your are proposing here and alternate solution to the same problem or an addition for specific edge case?

Hello, the fix that I am proposing is an alternative solution for what made in the previously mentioned PR and also my work add automated tests that reproduce the error and verify that was fixed. In the mentioned PR there was no tests on this behavior.

markkaylor:
Thanks for the PR and proposing this alternate solution. The fix #23681 addresses the the problem at the source (BlocksEditor), when a discard causes the form value to diverge from the slate's internal state we clear the selection entirely. Without good reason for why an additional fix is necessary I don't think we will move forward with the changes suggested on this PR.

As for the tests, the tests your propose here test only your solution. I acknowledge that the current fix lacks tests and a standalone PR to test that implementation would be welcome.

I'll close this PR but thanks again for your contribution and discussion.

PR #23825, opened 2025-06-26
test(api): fix issues with cleanup-after-delete

innerdvations:
### What does it do?

fix a test that doesn't clean up after itself and has issues running locally

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

Hoping it fixes the issues encountered in https://github.com/strapi/strapi/pull/23786


hanpaine:
@markkaylor did this solve the issues you found on #23786?

markkaylor:
@hanpaine  Last I tested it did not and I believe I still have the step skipped. I'll give it another try later today.

markkaylor:
Unfortunately it wasn't working, I pushed another fix that passed locally for this test and admin-guided-tour-meta, let's see if it passes here in the CI

PR #23828, opened 2025-06-26
feat: add tracker for conditional field count

butcherZ:
### What does it do?

add conditional field count tracker

### Why is it needed?

because

### How to test it?

1. in `packages/core/core/src/services/metrics/sender.ts` OR do a global find/replace (necessary if you are testing some events that happen outside of normal Strapi usage) 
2. replace the `https://analytics.strapi.io/`string to be `http://localhost`
You may have to replace it in several places, and when starting your Strapi server, you should see your localhost (http://localhost/) URL in the logs rather than analytics.strapi.io
3. Build Strapi again so that it builds with that URL
4. Run the command `npx http-echo-server 80` in a terminal window
NOTE: if you don't have port 80 available for this purpose, you’ll need to add a port to the previous line such as 'http://localhost:1234'
5.In another terminal window, start Strapi (or run the CLI command desired, whatever is sending the analytics you are testing) and do whatever action desired to send telemetry events

You should see those events sent by Strapi in the terminal window where the http-echo-server is running

You should see `"numberOfConditionalFields` and its value changes as you add / delete conditional field

### Related issue(s)/PR(s)

DX-2087

### DIFF
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -29,6 +29,7 @@ import createAuth from './services/auth';
 import createCustomFields from './services/custom-fields';
 import createContentAPI from './services/content-api';
 import getNumberOfDynamicZones from './services/utils/dynamic-zones';
+import getNumberOfConditionalFields from './services/utils/conditional-fields';
 import { FeaturesService, createFeaturesService } from './services/features';
 import { createDocumentService } from './services/document-service';
 
@@ -301,6 +302,7 @@ class Strapi extends Container implements Core.Strapi {
           numberOfAllContentTypes: _.size(this.contentTypes), // TODO: V5: This event should be renamed numberOfContentTypes in V5 as the name is already taken to describe the number of content types using i18n.
           numberOfComponents: _.size(this.components),
           numberOfDynamicZones: getNumberOfDynamicZones(),
+          numberOfConditionalFields: getNumberOfConditionalFields(),
           numberOfCustomControllers: Object.values<Core.Controller>(this.controllers).filter(
             // TODO: Fix this at the content API loader level to prevent future types issues
             (controller) => controller !== undefined && factories.isCustomController(controller)
--- a/packages/core/core/src/services/utils/__tests__/conditional-fields.test.ts
+++ b/packages/core/core/src/services/utils/__tests__/conditional-fields.test.ts
@@ -1,0 +1,309 @@
+import type { Schema, UID } from '@strapi/types';
+import getNumberOfConditionalFields from '../conditional-fields';
+
+const mockStrapi = {
+  contentTypes: {} as Record<UID.ContentType, Schema.ContentType>,
+  components: {} as Record<UID.Component, Schema.Component>,
+};
+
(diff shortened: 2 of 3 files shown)


jhoward1994:
Works well for me! We could add a simple unit test like this??

packages/core/core/src/services/utils/__tests__/conditional-fields.test.ts
```
import type { Schema, UID } from '@strapi/types';
import getNumberOfConditionalFields from '../conditional-fields';

const mockStrapi = {
  contentTypes: {} as Record<UID.ContentType, Schema.ContentType>,
  components: {} as Record<UID.Component, Schema.Component>,
};

(global as any).strapi = mockStrapi;

describe('getNumberOfConditionalFields', () => {
  beforeEach(() => {
    mockStrapi.contentTypes = {};
    mockStrapi.components = {};
  });

  it('should return 0 when there are no content types or components', () => {
    const result = getNumberOfConditionalFields();
    expect(result).toBe(0);
  });

  it('should return 0 when content types and components have no conditional fields', () => {
    mockStrapi.contentTypes = {
      'api::article.article': {
        attributes: {
          title: {
            type: 'string',
            required: true,
          },
          content: {
            type: 'text',
          },
        },
      } as unknown as Schema.ContentType,
    };

    mockStrapi.components = {
      'default.hero': {
        attributes: {
          title: {
            type: 'string',
          },
        },
      } as unknown as Schema.Component,
    };

    const result = getNumberOfConditionalFields();
    expect(result).toBe(0);
  });

  it('should count conditional fields in both content types and components', () => {
    mockStrapi.contentTypes = {
      'api::article.article': {
        attributes: {
          title: {
            type: 'string',
          },
          conditionalField: {
            type: 'string',
            conditions: {
              visible: true,
            },
          },
          notConditionalField: {
            type: 'string',
          },
        },
      } as unknown as Schema.ContentType,
    };

    mockStrapi.components = {
      'default.hero': {
        attributes: {
          conditionalSubtitle: {
            type: 'string',
            conditions: {
              visible: { field: 'title', operator: 'isNotEmpty' },
            },
          },
          anotherConditionalField: {
            type: 'text',
            conditions: {
              required: { field: 'subtitle', value: 'test' },
            },
          },
        },
      } as unknown as Schema.Component,
    };

    const result = getNumberOfConditionalFields();
    expect(result).toBe(3); // 1 from content type + 2 from component
  });

  it('should ignore fields with conditions that are not objects', () => {
    mockStrapi.contentTypes = {
      'api::article.article': {
        attributes: {
          title: {
            type: 'string',
          },
          fieldWithStringCondition: {
            type: 'string',
            conditions: 'some string',
          } as any,
          fieldWithNullCondition: {
            type: 'string',
            conditions: null,
          } as any,
          validConditionalField: {
            type: 'string',
            conditions: {
              visible: true,
            },
          },
        },
      } as unknown as Schema.ContentType,
    };

    const result = getNumberOfConditionalFields();
    expect(result).toBe(1); // Only the valid conditional field should be counted
  });
});


```

butcherZ:
@jhoward1994 awesome! thank you, I will add them :D

maccomaccomaccomacco:
LGTM

PR #23829, opened 2025-06-26
feat: homepage widget profile

Adzouz:
### What does it do?

Adding a new Profile widget to the homepage dashboard to display current user's information.

Screenshot 2025-06-26 at 17 16 04

### Why is it needed?

To enhance the experience of the user on the homepage.

### How to test it?

-> See the new block displayed with correct information 

🚀


packages/core/admin/admin/src/StrapiApp.tsx
```diff
@@ -317,6 +318,22 @@ class StrapiApp {
...
+        icon: User,
+        title: {
+          id: 'widget.profile.title',
+          defaultMessage: 'Profile',
+        },
+        component: async () => {
+          const { ProfileWidget } = await import('./components/Widgets');
+          return ProfileWidget;
+        },
+        pluginId: 'admin.widget.profile',
```
remidej: i think this should be just "admin". It's the id of the package registering the widget, not the id of the widget itself. The id of the widget is then computed inside of the register fn by putting together the pluginId and the id just below

packages/core/admin/admin/src/components/Widgets.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const user = useAuth('User', (state) => state.user);
+  const userDisplayName = getDisplayName(user);
+  const initials =
+    user?.firstname && user.lastname
+      ? `${user?.firstname.substring(0, 1)}${user.lastname.substring(0, 1)}`
+      : userDisplayName
+          .split(' ')
+          .map((name) => name.substring(0, 1))
+          .join('')
+          .substring(0, 2);
```
remidej: The initials already exist at the bottom of the admin left menu. Did you check how it's implemented there? It could be worth extracting a util, especially as this logic would benefit from unit tests

Adzouz:
I fixed the code based on your comments @remidej, thanks a lot 🙌

packages/core/admin/admin/src/utils/tests/users.test.ts
```diff
@@ -1,11 +1,55 @@
...
+      expect(getInitials({ firstname: 'Alice' })).toBe('A');
+    });
+
+    it('returns first two letters of username if no firstname/lastname', () => {
+      expect(getInitials({ username: 'foobar' })).toBe('F');
+    });
+
+    it('returns first two letters of email if no name fields', () => {
+      expect(getInitials({ email: 'user@example.com' })).toBe('U');
+    });
```
remidej: the assertion doesn't match the test names (1 letter vs 2)

Adzouz: Woops, monday morning am I right 😅 Good catch!

Shbh-9 (commented):
```

```@

PR #23830, opened 2025-06-26
fix: preserve non-localized fields when updating localized content for new locale

### LINKED ISSUE #21594: Updating localization results in data loss for unlocalized fields

- Node.js version: v20
- Strapi version: v5
- Database: PostgreSQL
- Operating system: macOS

### Describe the bug

Updating a content type with localized and unlocalized fields results in data loss if updating the localized fields for one locale, but not including the unlocalized data.

1. Create a type with e.g. name as localized field and image as unlocalized media field
2. Create an entry:  `const dummy = await strapi
          .documents("api::dummy.dummy")
          .create({
            data: {
              name: "nameInEnglish",
              locale: "en",
              image: imageId
            }
          });`
3. Look inside strapi dashboard, see image being linked
4. Update other locale: `await strapi.documents("api::dummy.dummy").update({
          documentId: dummy.documentId,
          locale: "es",
          data: {
            name: "nameInSpanish"
          }
        });`
5. Look inside strapi dashboard, link to image is lost

### Expected behavior

An unlocalized field should not be lost if updating a localized field

Similar issue with Strapi v4 : https://github.com/strapi/strapi/issues/18942

remidej:
## What does it do?

Fixes a bug where non-localized fields get lost when updating localized content for a new locale _via the document service_. Now when you update a document for a different locale, the shared fields are properly copied over from existing entries.

## Why is it needed?

When you have a content type with both localized and non-localized fields, updating just the localized field for a new locale would wipe out the shared field values.

fixes #21594

## How to test it?

Create a content type with mixed localized/non-localized fields, add an entry in English, then update it for Spanish with only the localized field using the document service. The shared fields should stick around.

Added a test case that covers this scenario.

### DIFF
--- a/packages/core/core/src/services/document-service/repository.ts
+++ b/packages/core/core/src/services/document-service/repository.ts
@@ -6,6 +6,7 @@ import type { UID } from '@strapi/types';
 import { wrapInTransaction, type RepositoryFactoryMethod } from './common';
 import * as DP from './draft-and-publish';
 import * as i18n from './internationalization';
+import { copyNonLocalizedFields } from './internationalization';
 import * as components from './components';
 
 import { createEntriesService } from './entries';
@@ -240,29 +241,14 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
         .findOne({ where: { documentId } });
 
       if (documentExists) {
-        const dataToCreate = { ...queryParams.data, documentId };
-
-        // Check if this is a localized content type and if i18n plugin is available
-        const i18nService = strapi.plugin('i18n')?.service('content-types');
-        if (i18nService?.isLocalizedContentType(contentType)) {
-          // Find an existing entry for the same document to copy unlocalized fields from
-          const existingEntry = await strapi.db.query(contentType.uid).findOne({
-            where: { documentId },
-            // Prefer published entry, but fall back to any entry
-            orderBy: { publishedAt: 'desc' },
-          });
-
-          // If an entry exists in another locale, copy its non-localized fields
-          if (existingEntry) {
-            i18nService.fillNonLocalizedAttributes(dataToCreate, existingEntry, {
-              model: contentType.uid,
-            });
-          }
-        }
+        const mergedData = await copyNonLocalizedFields(contentType, documentId, {
+          ...queryParams.data,
+          documentId,
+        });
 
         updatedDraft = await entries.create({
           ...queryParams,
-          data: dataToCreate,
+          data: mergedData,
         });
         emitEvent('entry.create', updatedDraft);
       }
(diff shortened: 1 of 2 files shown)


tests/api/core/strapi/document-service/update.test.api.ts
```diff
@@ -129,5 +129,53 @@ describe('Document Service', () => {
 
       expect(article).toBeNull();
     });
+
+    testInTransaction(
+      'Preserves non-localized fields when updating localized content for new locale',
+      async () => {
+        // This test covers the fix for bug #21594: https://github.com/strapi/strapi/issues/21594
+        // When updating a content type with both localized and non-localized fields,
+        // updating only the localized field for a new locale should preserve the non-localized fields
```
remidej:
```suggestion
        // Covers issue https://github.com/strapi/strapi/issues/21594
```

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -240,9 +240,34 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
         .findOne({ where: { documentId } });
 
       if (documentExists) {
+        const dataToCreate = { ...queryParams.data, documentId };
+
+        // Check if this is a localized content type and if i18n plugin is available
+        const i18nService = strapi.plugin('i18n')?.service('content-types');
```
markkaylor: It looks like there is file of internationalization utils. Most of the utils seem related to params but still wondering if all this is belongs over there?

packages/core/core/src/services/document-service/repository.ts
```diff
@@ -240,9 +240,34 @@ export const createContentTypeRepository: RepositoryFactoryMethod = (
...
+            orderBy: { publishedAt: 'desc' },
+          });
+
+          // If an entry exists in another locale, copy its non-localized fields
+          if (existingEntry) {
+            const nonLocalizedAttributes = i18nService.getNonLocalizedAttributes(contentType);
+
+            // Don't override provided data
+            for (const attr of nonLocalizedAttributes) {
+              if (existingEntry[attr] !== undefined && dataToCreate[attr] === undefined) {
```
markkaylor: The value could never be null?

remidej: i'll make it a loose check 👍

markkaylor (approved):
The code looks good to me, but I haven't tested it locally.

simotae14 (approved):
I tested your stuff in the Country collection where I set the code field as not localized

Here the instructions used in the Strapi console to create and update a country document
Screenshot 2025-07-01 alle 16 38 56
and here you can see the results of my operations in the italian locale
Screenshot 2025-07-01 alle 16 39 04
It is working because I have not lost the english code version when I have updated the document in the italian version

Well done

PR #23833, opened 2025-06-27
feat(core): add uid transformation utility for openapi compliant names

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/core-api/routes/validation/utils.ts
+++ b/packages/core/core/src/core-api/routes/validation/utils.ts
@@ -7,6 +7,39 @@
 import * as z from 'zod/v4';
 
 /**
+ * Transforms a Strapi UID into an OpenAPI-compliant component name.
+ *
+ * @param uid - The Strapi UID to transform (e.g., "basic.seo", "api::category.category", "plugin::upload.file")
+ * @returns The OpenAPI-compliant component name (e.g., "BasicSeoEntry", "ApiCategoryCategoryDocument", "PluginUploadFileDocument")
+ */
+export const transformUidToValidOpenApiName = (uid: string): string => {
+  const capitalize = (str: string): string => {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  };
+
+  const toPascalCase = (str: string): string => {
+    return str.split(/[-_]/).map(capitalize).join('');
+  };
+
+  // Check if it contains double colons (other namespaced UIDs)
+  if (uid.includes('::')) {
+    const [namespace, ...rest] = uid.split('::');
+    const namespacePart = toPascalCase(namespace);
+    const restParts = rest.join('.').split('.').map(toPascalCase).map(capitalize);
+    return `${capitalize(namespacePart)}${restParts.join('')}Document`;
+  }
+
+  if (uid.includes('.')) {
+    // basic.seo -> BasicSeoEntry
+    const parts = uid.split('.');
+    const transformedParts = parts.map(toPascalCase).map(capitalize);
+    return `${transformedParts.join('')}Entry`;
+  }
+
+  return `${toPascalCase(capitalize(uid))}Schema`;
+};
+
+/**
  * Conditionally makes a Zod schema optional based on the `required` parameter.
  *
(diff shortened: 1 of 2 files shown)


packages/core/core/src/core-api/routes/validation/utils.ts
```diff
@@ -6,6 +6,39 @@
 
 import * as z from 'zod/v4';
 
+/**
+ * Transforms a Strapi UID into an OpenAPI-compliant component name.
+ *
+ * @param uid - The Strapi UID to transform (e.g., "basic.seo", "api::category.category", "plugin::upload.file")
+ * @returns The OpenAPI-compliant component name (e.g., "BasicSeoEntry", "ApiCategoryCategoryDocument", "PluginUploadFileDocument")
+ */
+export const transformUidToValidOpenApiName = (uid: string): string => {
```
innerdvations: Just noticed we're weren't validating that uid is actually valid anywhere in the parent callers, maybe we should add a sanity check for empty string, invalid uid like extra ::, etc and log a warning if we find one?

packages/core/core/src/core-api/routes/validation/utils.ts
```diff
@@ -155,10 +190,12 @@ export const safeGlobalRegistrySet = (id: string, schema: z.ZodType) => {
...
   const { _idmap: idMap } = z.globalRegistry;
 
-  const existsInGlobalRegistry = idMap.has(id);
+  const transformedId = transformUidToValidOpenApiName(id);
+
+  const existsInGlobalRegistry = idMap.has(transformedId);
 
   if (existsInGlobalRegistry) {
-    return idMap.get(id) as z.ZodType;
+    return idMap.get(transformedId) as z.ZodType;
```
innerdvations: `.has` doesn't work as a typeguard, we can probably replace it and do this instead to avoid `as`

```suggestion
    const item = idMap.get(transformedId);
    if (!item) {
      throw new Error(`Schema with id ${transformedId} not found in global registry`);
    }
    return item;
```

Maybe log a warning instead of throwing though? That might be better so we don't block the entire file generation from a small error.

innerdvations (approved):
Works for me! Approved even as-is, but two comments that might be nice to add.

packages/core/core/src/core-api/routes/validation/utils.ts
jhoward1994: Thanks this made me think we could just type the parameters for these functions more accurately?  

wdyt see [`eac0751` (#23833)](https://github.com/strapi/strapi/pull/23833/commits/eac0751d4098d311aa2149596861ec082c9483e4)

packages/core/core/src/core-api/routes/validation/utils.ts
jhoward1994: Done in [`4c01fe6` (#23833)](https://github.com/strapi/strapi/pull/23833/commits/4c01fe6b1546b49c95dad41483de1d3d9a8d296a) ??

innerdvations:
Typescript test is failing :(

packages/core/core/src/core-api/routes/validation/utils.ts
innerdvations: Good idea, I think that's good enough here; this will force us to actually check our inputs before calling

packages/core/core/src/core-api/routes/validation/utils.ts
```diff
@@ -193,12 +193,15 @@ export const safeSchemaCreation = (id: Internal.UID.Schema, callback: () => z.Zo
 
   const transformedId = transformUidToValidOpenApiName(id);
 
-  const existsInGlobalRegistry = idMap.has(transformedId);
-
-  if (existsInGlobalRegistry) {
-    return idMap.get(transformedId) as z.ZodType;
+  const mapItem = idMap.get(transformedId);
+  if (mapItem) {
+    return mapItem as z.ZodType;
```
innerdvations: You shouldn't need the `as z.ZodType` anymore, it was only showing an error before because it could possibly be undefined

PR #23834, opened 2025-06-27
fix: sign file URLs in upload service and in graphql

### LINKED ISSUE #21749: Media AWS S3 in Strapi v5 does't have a signed url in API response

- Node.js version: 20
- NPM version: 10
- Strapi version: 5.0.4
- Database: mysql
- Operating system: MacOS

### Describe the bug

From various tests, it seems that the images are not signed only for those linked to the User entity (users-permissions plugin).

If I call any API for other entities, the images work.
Only for media-relations into user-permissions plugin does't work. (images or files)

Example of signed url:
https://----bucket----.s3.eu-north-1.amazonaws.com/Screenshot_2024_10_09_alle_21_13_45_12be30dc56.png?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Credential=----%----%2Feu-north-1%2Fs3%2Faws4_request&X-Amz-Date=20241009T205030Z&X-Amz-Expires=900&X-Amz-Signature=----&X-Amz-SignedHeaders=host&x-id=GetObject)

```

1. Upload a file with s3 in private bucket
2. Retrieve a files via API for User

### Expected behavior

I would have a signedUrl media for User

### Screenshots

### Code snippets

### Additional context

### PR DESCRIPTION
**Pull request by remidej**
### What does it do?

Ensures that all file URLs from the upload plugin actually get signed. This wasn't the case when using the upload service or when using the graphql plugin.

- updates the document middleware so it signs media entries directly (`upload::file.file`) in addition to media attributes
- runs signFileUrl util from the upload service
- fixes the unit testing setup in the graphql plugin (it looks like all the unit tests in there didn't run before)

### How to test it

Configure your app to use a private S3 bucket. DM me for access to the required env variables.

```js
// config/middlewares.js

'use strict';

const responseHandlers = require('./src/response-handlers');

module.exports = ({ env }) => [
  'strapi::logger',
  'strapi::errors',
  {
    name: 'strapi::security',
    config: {
      contentSecurityPolicy: {
        useDefaults: true,
        directives: {
          'frame-src': ["'self'"], // URLs that will be loaded in an iframe (e.g. Content Preview)
          'img-src': [
            "'self'",
            'data:',
            'blob:',
            'market-assets.strapi.io',
            `${env('AWS_BUCKET')}.s3.${env('AWS_REGION')}.amazonaws.com`,
          ],
          'media-src': [
            "'self'",
            'data:',
            'blob:',
            'market-assets.strapi.io',
            `${env('AWS_BUCKET')}.s3.${env('AWS_REGION')}.amazonaws.com`,
          ],
        },
      },
    },
  },
  'strapi::cors',
  'strapi::poweredBy',
  'strapi::query',
  'strapi::body',
  'strapi::session',
  // 'strapi::compression',
  // 'strapi::ip',
  {
    name: 'strapi::responses',
    config: {
      handlers: responseHandlers,
    },
  },
  'strapi::favicon',
  'strapi::public',
  {
    name: 'global::test-middleware',
    config: {
      foo: 'bar',
    },
  },
  {
    resolve: './src/custom/middleware.js',
    config: {},
  },
];
```

```js
//config/plugins.js


  // ...the config for all other plugins...
  upload: {
    config: {
      provider: 'aws-s3',
      providerOptions: {
        // baseUrl: `https://${env('AWS_BUCKET')}.s3.${env('AWS_REGION')}.amazonaws.com`,
        // rootPath: env('CDN_ROOT_PATH'),
        s3Options: {
          credentials: {
            accessKeyId: env('AWS_ACCESS_KEY_ID'),
            secretAccessKey: env('AWS_ACCESS_SECRET'),
          },
          region: env('AWS_REGION'),
          params: {
            ACL: env('AWS_ACL', 'public-read'),
            signedUrlExpires: env('AWS_SIGNED_URL_EXPIRES', 15 * 60),
            Bucket: env('AWS_BUCKET'),
          },
        },
      },
      actionOptions: {
        upload: {},
        uploadStream: {},
        delete: {},
      },
    },
  },
});
```

Then:
- upload assets, check that the URL you get for the assets are signed (they should include a bunch of query params starting in `X-Amz-`
- query content with graphQL (localhost:1337/admin), check that the URLs are signed. In the footer under the Headers tab you'll need to add an `Authorization` header with the value of `Bearer your_jwt_here`. To find your JWT, run this in the console: `JSON.parse(window.localStorage.getItem('jwtToken'))`.  Here's a sample query:
  ```graphql
  query {
    # test media attributes
    addresses {
      documentId
      images { url}
    }
    # test media files directly
    uploadFiles {
      url
    }
  }
   ```

### Related issue(s)/PR(s)

fixes #21749


packages/core/upload/server/src/services/__tests__/upload/formatFileInfo.test.ts
remidej, line 1: the diff looks big but there are very few changes (see it [vscode style instead](https://github.dev/strapi/strapi/pull/23834/files#diff-1b1df1e3c6f18a49e384812a9bab6f376c044351e86390542d86c984d7395d87)), it's just that i removed an unnecessary describe that wrapped everything

packages/core/upload/server/src/services/extensions/utils.ts
```diff
@@ -67,6 +68,17 @@ const signEntityMediaVisitor: SignEntityMediaVisitor = async (
  * @returns
  */
 const signEntityMedia = async (entity: any, uid: UID.Schema) => {
+  if (!entity) {
+    return entity;
+  }
+
+  // Special handling for direct upload file entities
```
markkaylor: what do you mean by "direct"?

packages/core/upload/server/src/services/__tests__/upload/uploadImage.test.ts
```diff
@@ -5,6 +5,18 @@ import _ from 'lodash';
...
+global.strapi = {
+  plugins: {
+    upload: {
+      services: {
+        'image-manipulation': imageManipulation,
+      },
+    },
+  },
+  plugin: (name: string) => global.strapi.plugins[name],
+} as any;
```
markkaylor: There is a mockUploadProvider function below that sets up global.strapi, should this be handled there? Seems a bit strange to be mutating the global in two places.

packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
```diff
@@ -76,7 +76,18 @@ export default ({ strapi }: Context) => {
...
+        // Sign media URLs if upload plugin is available and using private provider
+        if (isMediaAttribute && strapi.plugin('upload')) {
+          const { signFileUrls } = strapi.plugin('upload').service('file');
+
+          if (Array.isArray(data)) {
+            data = await async.map(data, (item: any) => signFileUrls(item));
+          } else if (data) {
+            data = await signFileUrls(data);
+          }
+        }
```
markkaylor: Preference would be to respect the fp paradigm with a function instead of mutable variable

packages/core/upload/server/src/services/__tests__/upload/uploadImage.test.ts
remidej: I still need to mutate some of the global in 2 places, because global.strapi.plugin needs to exist before createUploadService() runs, and then the upload and getSettings need to change from test to test.

but I updated it so we only mutate those functions when needed instead of redefining global.strapi

packages/core/upload/server/src/services/extensions/utils.ts
remidej: bad wording indeed 👍

packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
remidej: I'm with you but I also don't love extracting functions for this kind of stuff, so I made it a self-invoking one. Let me know if it looks readable to you or not

markkaylor: Yep I get it and I agree, I dig the self-invoking functions and did the same thing for the guided tour to avoid extracting a function as well. Looks readable to me. All up to you though for what you think is best :)

PR #23836, opened 2025-06-27
chore: add env tracker

butcherZ:
### What does it do?

replaced hard coded url for tracking events with env variable

### Why is it needed?

make testing tracking easier

### How to test it?

add `STRAPI_ANALYTICS_URL=http://localhost` to the `.env` file at the root of `example/getStarted`
### Related issue(s)/PR(s)

DX-2088


packages/cli/create-strapi-app/src/utils/usage.ts
```diff
@@ -57,8 +57,10 @@ function trackEvent(event: string, payload: Record<string, unknown>) {
     return;
   }
 
+  const analyticsUrl = process.env.STRAPI_ANALYTICS_URL || 'https://analytics.strapi.io';
```
innerdvations, line 60: Do you think it would be possible to have a centralized constant for the string instead of hard-coding it everywhere?

As far as I know we don't currently have anywhere to put constants that are used in multiple packages, and I wouldn't want to add it now since it will take some discussion. So I'm asking because I'm not sure, but I'm curious if you have ideas 😆

butcherZ: hmm in that case maybe we should just remove the hard-coded string at all and use the env variable instead, which will require @maccomaccomaccomacco to manually set it up each time he wants to test  😆

innerdvations: It would still need the default to work in general. Let's just leave it as you've written it for now and deal with that later, this is still an improvement

innerdvations (changes requested):
Looks like this got some junk left over from conditional fields things that changed that we need to clean out before merging

It might be better to just cherry pick the commits you actually want here into a fresh branch to avoid having all that history

PR #23842, opened 2025-06-30
fix: color of release status icon

### LINKED ISSUE #23760: Release's statuses icons colors don't adapt to the text color
### Node Version

22.11.0

### NPM/Yarn/PNPM Version

Yarn 4.5.0

### Strapi Version

5.16.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Release's statuses icons colors don't adapt to the text color.

### Steps to Reproduce

1. Go to releases
2. Open a releases
3. Check the statuses

### Expected Behavior

The Release's statuses icons colors should adapt to the text color.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

Image

### Additional information

_No response_

### Confirmation Checklist

Adzouz:
### What does it do?

Overriding the color of the icon in the button of the status of an entry in Releases.

Screenshot 2025-06-30 at 08 58 05

This specific variant doesn't exist in the design system so we need to override the style of the icon. We can't use an existing variant because it's also using the alternative600 color and there's no button for this one. But it could be a suggestion to provide it at some point @lucasboilly?

### Why is it needed?

To improve consistency with the design system.

### How to test it?

- Go to the admin interface
- You have to create a release (if you don't already have one)
- You have to put some entries to your release
  - Publish action (select Publish action when adding to a release -> the three dots icon next to the Publish button)
    - One entry with some errors on any field => "Not ready to publish"
    - One entry with status Draft => "Ready to publish"
    - One entry with status Modified => "Ready to publish changes"
    - One entry with status Published => "Already published"
  - Unpublish action
    - One entry with status Published => "Ready to unpublish"
    - One entry with status Unpublished => "Already unpublished"
- Then go to Releases and select the release you put all your entries
-> You should see the status of each entry in the list

### Related issue(s)/PR(s)

Resolves #23760

🚀

### DIFF
--- a/packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
+++ b/packages/core/content-releases/admin/src/components/EntryValidationPopover.tsx
@@ -29,6 +29,34 @@ const StyledPopoverFlex = styled(Flex)`
   }
 `;
 
+const ButtonContent = styled(Flex)`
+  svg {
+    fill: currentColor;
+  }
+`;
+
+const CustomStatusButton = ({
+  children,
+  icon,
+  color,
+}: {
+  icon: React.ReactElement;
+  color: string;
+} & React.PropsWithChildren) => {
+  return (
+    <Popover.Trigger>
+      <Button variant="ghost" endIcon={<CaretDown />}>
+        <ButtonContent color={color} gap={2}>
+          {icon}
+          <Typography textColor={color} variant="omega" fontWeight="bold">
+            {children}
+          </Typography>
+        </ButtonContent>
+      </Button>
+    </Popover.Trigger>
+  );
+};
+
 interface EntryValidationPopoverProps {
   action: ReleaseAction['type'];
   schema?: Struct.ContentTypeSchema & {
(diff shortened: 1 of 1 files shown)


lucasboilly:
@Adzouz in that specific case we can't use the badges because we need to be able to click on the status to trigger a dropdown. It's fine for now but it could be optimized later I guess. Anyway thanks for fixing this!

Adzouz:
> @Adzouz in that specific case we can't use the badges because we need to be able to click on the status to trigger a dropdown. It's fine for now but it could be optimized later I guess. Anyway thanks for fixing this!

Yes totally, I was talking about the buttons (https://design-system.strapi.io/?path=/story/components-button--all-variants) if we add the alternative variant, we could use the corresponding variant to each of them (danger-ghost, success-ghost and alternative-ghost). My point was more to provide the correct button on DS side but as you said, for the moment it's enough to do it there since it's the only place it's used (right?).
Thanks!

lucasboilly:
@Adzouz in fact there's another place where it's used, it's in the bulk publish modal. At the moment the only difference is that you can't click on it.

CleanShot 2025-06-30 at 18 04 50@2x

Adzouz:
> @Adzouz in fact there's another place where it's used, it's in the bulk publish modal. At the moment the only difference is that you can't click on it.
> 
> CleanShot 2025-06-30 at 18 04 50@2x

@lucasboilly Yes, there it's because it's not using buttons in that case. I see that it's not consistent on the colors. Do you want me to update them also (changing the color of the text to fit the color of the icon)?

lucasboilly:
I'd like to speak about it with @yanniskadiri first because I wonder if we couldn't simply use the exact same component from releases in the bulk publish modal. But I need to refresh my memory 😅

CleanShot 2025-07-02 at 09 11 32@2x

Adzouz:
> I'd like to speak about it with @yanniskadiri first because I wonder if we couldn't simply use the exact same component from releases in the bulk publish modal. But I need to refresh my memory 😅
> 
> CleanShot 2025-07-02 at 09 11 32@2x

Do we have a status update on this one @lucasboilly and @yanniskadiri? Could be a quickwin to close asap.

yanniskadiri:
Hey @Adzouz, following a discussion with @lucasboilly, I checked and confirmed that we're currently using different components for releases and bulk publishing

If it’s feasible, I think it would make sense to align both and use the same component for consistency. What do you think?

Adzouz:
Hey @yanniskadiri and @lucasboilly,
I can't use the same component per se because one is a button and the other is just some text but I made them look alike to keep consistency.
Is it good for you now?

Screenshot 2025-07-24 at 17 24 59

yanniskadiri:
Got it, and yes it looks good to me - thanks @Adzouz!

markkaylor (approved):

PR #23844, opened 2025-07-01
future: add content-type builder tour

markkaylor:
### What does it do?

- Add the content-type builder guided tour
- Changes the `requiredActions` API to `when`
- Requires a content type schema to have at least one attribute besides the default ones
- Updates some styling issues
- Saves/reads the tour state to/from local storage
- Import tours directly instead of using a prop on the provider

### Why is it needed?

To help users use the product

### How to test it?

You will need an empty strapi project. I created one in examples. I could commit and push it here if it's easier for testing.

Set the future flag:
```
export default () => ({
  future: {
    unstableGuidedTour: true
  }
});

```
Given you have no content-types and no content in your app
Go to the Content-type builder, the tour should being
Go through the steps. 
After components you need to create a content-type with an attribute
Then you should see the final tooltip leading you to the content manager
Click Next here should navigate to the content manager

Ensure the tour state is only stored in local storage when the future flag is set. Be sure to clear local storage every time you test.

https://github.com/user-attachments/assets/1586b53c-3b4b-4cfa-9b6d-4c047be05cbd

NOTE: 

The arrow on the tooltip from the designs is not included here. I'll open a PR in the DS to export it and then add it in another PR

### Related issue(s)/PR(s)


packages/core/admin/admin/src/components/SubNav.tsx
```diff
@@ -123,6 +145,7 @@ const Section = ({
   link?: { label: string; onClik: () => void };
 }) => {
   const listId = useId();
+  const GuidedTourForSection = getGuidedTourForSection(label) ?? Fragment;
```
markkaylor: Need to look at this again, I think this is the translated string so it won't work

markkaylor:
Oops thought I fixed the tests 🤦‍♂️

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/useContentTypeBuilderMenu.ts
```diff
@@ -25,7 +25,7 @@ type SubSection = {
   links: Link[];
 };
 
-type MenuSection = {
+export type MenuSection = {
```
markkaylor: Unintentional

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -148,16 +200,27 @@ function createTour<const T extends ReadonlyArray<TourStep<string>>>(tourName: s
...
-    );
+    acc[step.name as keyof Components] = ({ children }: { children: React.ReactNode }) => {
+      /**
+       * TODO
+       * Find a way to mock the tours and provide a passthrough to children for the tooltip.
+       * I've tried everything I can think of and can't get it to work.
+       */
+      if (process.env.NODE_ENV === 'test') {
+        return children;
+      }
```
markkaylor: I think I found a better soluton

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -117,9 +178,15 @@ const UnstableGuidedTourTooltip = ({
     };
   }, [isPopoverOpen]);
 
+  const Step = React.useMemo(() => createStepComponents(tourName), [tourName]);
+
   return (
     <>
-      {isPopoverOpen && <GuidedTourOverlay />}
+      {isPopoverOpen && (
+        <Portal>
```
markkaylor, line 186: Full transparency: if I do this inside the Popover.Content to share it's portal for some reason the overlay is rendered twice and therefore is twice as dark 🤦‍♂️. I don't dig duplicating the condition but also not thaaat bothered by it.

packages/core/admin/admin/src/components/MainNav/NavLink.tsx
```diff
@@ -39,8 +41,25 @@ const MainNavLinkWrapper = styled(RouterLink)`
   }
 `;
 
+const getGuidedTourTooltip = (to: To) => {
+  const normalizedTo = to.toString().replace(/\//g, '');
+
+  switch (normalizedTo) {
+    case 'content-manager':
```
simotae14, line 48: I suppose you made a typo, is the content type builder case

markkaylor: I don't think so because we want to point to the content-manager at the final step of the content-type builder tour

packages/core/admin/admin/src/components/Providers.tsx
simotae14, line 1: if you are not passing tours as a UnstableGuidedTourContext you need to remove the tours import

simotae14 (approved):
I tested the project with an experimental and it works, well done

PR #23847, opened 2025-07-01
fix: make upgrade tool work with a proxy

### LINKED ISSUE #22944: @strapi/upgrade fails with "fetch failed" behind proxy server
### Node Version

22.12.0

### NPM/Yarn/PNPM Version

10.9.0

### Strapi Version

4.25.20

### Operating System

Linux (Debian/Ubuntu)

### Database

MySQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When running `npx @strapi/upgrade minor` or `npx @strapi/upgrade major` I encounter

```
npx @strapi/upgrade major --debug
[WARN]	[2025-02-19T13:53:31.674Z] Please make sure you've created a backup of your codebase and files before upgrading
[DEBUG]	[2025-02-19T13:53:31.691Z] Project: TYPE=application; CWD=...; PATHS=./{src,config,public}/**/*.{js,mjs,ts,jsx,tsx,json},!./**/node_modules/**/*,!./**/dist/**/*,package.json
[DEBUG]	[2025-02-19T13:53:31.691Z] Application: VERSION=v4.25.20; STRAPI_VERSION=v5.10.2
[ERROR]	[2025-02-19T13:53:42.252Z] fetch failed
```

Using `strace` I see direct connection attempts to IP addresses of `registry.npmjs.org`.

In my `~/.npmrc` and `.npmrc` the proxy settings are correct and even read according to `strace`:

```
proxy=http://proxy...:3128
https-proxy=http://proxy...:3128
noproxy=localhost,...
```

Sometimes the proxy is even used as the trace shows. 

### Steps to Reproduce

Just run `npx @strapi/upgrade minor` behind a proxy server and direct outbound HTTP traffic forbidden.

### Expected Behavior

The proxy settings are accepted when running the command.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

strace output:

```
...
connect(20, {sa_family=AF_INET6, sin6_port=htons(443), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "2606:4700::6810:1e22", &sin6_addr), sin6_scope_id=0}, 28) = -1 EINPROGRESS
...
connect(20, {sa_family=AF_INET, sin_port=htons(443), sin_addr=inet_addr("104.16.30.34")}, 16) = -1 EINPROGRESS
...
connect(20, {sa_family=AF_INET, sin_port=htons(443), sin_addr=inet_addr("104.16.1.35")}, 16) = -1 EINPROGRESS
...
connect(20, {sa_family=AF_INET6, sin6_port=htons(443), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "2606:4700::6810:1d22", &sin6_addr), sin6_scope_id=0}, 28) = -1 EINPROGRESS
...
```

It should connect to the proxy server.

### Confirmation Checklist

Bassel17:
### What does it do?

- Use undici fetch instead of native fetch
- Add the proxy if it's configured

### Why is it needed?

fix #22944 

### How to test it?

- Setup a proxy on your device, and run the upgrade tool it should work

### DIFF
--- a/packages/utils/upgrade/src/modules/npm/__tests__/package.test.ts
+++ b/packages/utils/upgrade/src/modules/npm/__tests__/package.test.ts
@@ -53,7 +53,7 @@ describe('Package class', () => {
   const packageInstance = new Package(packageName);
   it('should fetch package data and update npmPackage', async () => {
     await packageInstance.refresh();
-    expect(global.fetch).toHaveBeenCalledWith(packageInstance.packageURL);
+    expect(global.fetch).toHaveBeenCalledWith(packageInstance.packageURL, expect.anything());
     expect(packageInstance.versionExists).toBeTruthy();
   });
 
@@ -108,7 +108,7 @@ describe('Package class', () => {
 
     await packageInstance.refresh();
 
-    expect(global.fetch).toHaveBeenCalledWith(packageInstance.packageURL);
+    expect(global.fetch).toHaveBeenCalledWith(packageInstance.packageURL, expect.anything());
     expect(packageInstance.isLoaded).toBeTruthy();
     expect(packageInstance.getVersionsAsList().length).toBeGreaterThan(0);
   });
--- a/packages/utils/upgrade/src/modules/npm/package.ts
+++ b/packages/utils/upgrade/src/modules/npm/package.ts
@@ -1,7 +1,7 @@
 import assert from 'node:assert';
 import semver from 'semver';
 
-import { fetch, ProxyAgent } from 'undici';
+import { ProxyAgent } from 'undici';
 import * as constants from './constants';
 import { isLiteralSemVer } from '../version';
 
@@ -65,7 +65,10 @@ export class Package implements PackageInterface {
   }
 
   async refresh() {
-    const response = await fetch(this.packageURL, { dispatcher: agent });
+    const response = await fetch(this.packageURL, {
+      // @ts-expect-error Node.js fetch supports dispatcher (undici extension)
+      dispatcher: agent,
+    });
 
     // TODO: Use a validation library to make sure the response structure is correct
     assert(response.ok, `Request failed for ${this.packageURL}`);


innerdvations (commented):
We already use experimental (in node 18,20) support for dispatchers in native fetch in core: `packages/core/core/src/utils/fetch.ts` and node 22 has stable support for it so it's probably fine to just use that

Bassel17:
> We already use experimental (in node 18,20) support for dispatchers in native fetch in core: `packages/core/core/src/utils/fetch.ts` and node 22 has stable support for it so it's probably fine to just use that

Actually I don't think that's gonna work for this case in the upgrade tool, the strapiFetch expects a server.proxy.fetch or a server.proxy.global to be loaded from a strapi project, that wouldn't work when people are using the upgrade tool since there is no strapi project being loaded

innerdvations:
> > We already use experimental (in node 18,20) support for dispatchers in native fetch in core: `packages/core/core/src/utils/fetch.ts` and node 22 has stable support for it so it's probably fine to just use that
> 
> Actually I don't think that's gonna work for this case in the upgrade tool, the strapiFetch expects a server.proxy.fetch or a server.proxy.global to be loaded from a strapi project, that wouldn't work when people are using the upgrade tool since there is no strapi project being loaded

Ahhh, ok! This works for me then (although wouldn't it be the same thing, just that we respect HTTP_PROXY env vars rather than pulling it from config?)

Bassel17:
> > > We already use experimental (in node 18,20) support for dispatchers in native fetch in core: `packages/core/core/src/utils/fetch.ts` and node 22 has stable support for it so it's probably fine to just use that
> > 
> > 
> > Actually I don't think that's gonna work for this case in the upgrade tool, the strapiFetch expects a server.proxy.fetch or a server.proxy.global to be loaded from a strapi project, that wouldn't work when people are using the upgrade tool since there is no strapi project being loaded
> 
> Ahhh, ok! This works for me then (although wouldn't it be the same thing, just that we respect HTTP_PROXY env vars rather than pulling it from config?)

yeah same concept, its just our fetch require the strapi instance to be running to be able to get the configs, that's not what happens when the upgrade tool is used it will just run the scripts and in this case get the proxy from env

innerdvations (approved):
LGTM!

PR #23849, opened 2025-07-01
fix(strapi): add output option for openapi generation command

jhoward1994:
### What does it do?

- Correctly add the --output option to the command:

`strapi openapi generate --output path/to/output.json`

- Uses `outputFileSync` so the path is created if it doesn't exist

### How to test it?

Run this from `examples/getstarted`

`strapi openapi generate --output path/to/output.json`


jhoward1994:
I think we could merge this even with the build error, it's fixed in https://github.com/strapi/strapi/pull/23833

butcherZ (approved):
QAed, works well! 😸

PR #23853, opened 2025-07-01
fix: auth providers generate unique username on conflict

### LINKED ISSUE #23852: Auth providers create duplicate username entries if username is already registered before
### Node Version

20.0.0

### NPM/Yarn/PNPM Version

10.0.0

### Strapi Version

5.7.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

Registering normally, with a username `joe`, after which trying to register with `joe@gmail.com` or any other provider, due to the current handling of usernames, allowed the user to be created with the username `joe` again. Leaving us with 2 users with the same username. Due to the default behaviour of strapi, usernames are generated from emails differently depending on providers, but mostly it is by getting what's left of the `@` symbol (`email.split('@')[0]`).

This is a breaking issue, that needs to be looked at. I think it is best to append a counter at the end of an already taken username in these scenarios. Initially register `joe` would still stay. Newly registered with provider, `joe@gmail.com` will get the username `joe1` instead of the duplicate `joe`, which fixes the issue.

Basically, we get two users with the same username, even though the emails might've been different. Which is unexpected behaviour and a potential security issue, since the Username is a unique identifier.

### Steps to Reproduce

1. Register a user with a username that you have on another provider (ex. `joe` with email `test@test.com`, and you have a provider account let's say google with email `joe@gmail.com`) 
2. Register with the provider (ex. `joe@gmail.com`)
3. Now you will have two users with the same username `joe`

### Expected Behavior

Expected behaviour is to have some handling for providers that generates a different username if the username is already registered before. For example it could append a number depending on how many users there are with the same username already.

### Logs

```shell

```

### Code Snippets

` async authCallback({ accessToken }) {
      const google = purest({ provider: 'google' });

      return google
        .query('oauth')
        .get('tokeninfo')
        .qs({ accessToken })
        .request()
        .then(({ body }) => ({
          username: body.email.split('@')[0],
          email: body.email,
        }));
    },`

This is an example of how the google provider generates the username.

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

ndinevski:
### What does it do?

This change appends a number at the end of a username if the username generated from the PROVIDER email is already taken instead of making a duplicate username.

### Why is it needed?

fix #23852 

This change is motivated by a bug I incountered in my application. Registering normally, with a username `joe`, after which trying to register with `joe@gmail.com` or any other provider, due to the current handling of usernames, allowed the user to be created with the username `joe` again. Leaving us with 2 users with the same username. This is a breaking issue, that needs to be looked at. I think it is best to append a counter at the end of an already taken username in these scenarios. Initially register `joe` would still stay. Newly registered with provider, `joe@gmail.com` will get the username `joe1` instead of the duplicate `joe`, which fixes the issue.

### How to test it?

Register using a username that you have a username on any provider with, ex. `joe` and you have an email ex. `joe@random.com`. It is important not to use the google account here and use a different email. After this user any auth provider login/register, and register with the provider account ex. `joe@google.com`. A user with username `joe1` will be created, instead of the duplicate `joe`. Therefore fixing the issue.

### DIFF
--- a/packages/plugins/users-permissions/server/services/providers.js
+++ b/packages/plugins/users-permissions/server/services/providers.js
@@ -86,9 +86,28 @@ module.exports = ({ strapi }) => {
       .query('plugin::users-permissions.role')
       .findOne({ where: { type: advancedSettings.default_role } });
 
+    let username = email.split('@')[0];
+
+    // Get all users with the same username
+    const usersWithUsername = await strapi.db
+      .query("plugin::users-permissions.user")
+      .findMany({
+        where: { username },
+      });
+
+    // If there are users with the same username, append a number to the username
+    if (usersWithUsername.length > 0) {
+      let counter = 1;
+      while (usersWithUsername.some(user => user.username === username)) {
+        username = `${email.split('@')[0]}${counter}`;
+        counter++;
+      }
+    }
+
     // Create the new user.
     const newUser = {
       ...profile,
+      username, // use the generated or provided username
       email, // overwrite with lowercased email
       provider,
       role: defaultRole.id,
--- a/packages/plugins/users-permissions/server/services/providers.js
+++ b/packages/plugins/users-permissions/server/services/providers.js
@@ -8,7 +8,7 @@
 const _ = require('lodash');
 const urlJoin = require('url-join');
 
-const { getService } = require('../utils');
+const { getService, findUniqueUsername } = require('../utils');
 
 module.exports = ({ strapi }) => {
   /**
@@ -86,23 +86,8 @@ module.exports = ({ strapi }) => {
       .query('plugin::users-permissions.role')
       .findOne({ where: { type: advancedSettings.default_role } });
 
-    let username = email.split('@')[0];
-
-    // Get all users with the same username
-    const usersWithUsername = await strapi.db
-      .query("plugin::users-permissions.user")
-      .findMany({
-        where: { username },
-      });
-
-    // If there are users with the same username, append a number to the username
-    if (usersWithUsername.length > 0) {
-      let counter = 1;
-      while (usersWithUsername.some(user => user.username === username)) {
-        username = `${email.split('@')[0]}${counter}`;
-        counter++;
-      }
-    }
+    // Generate a unique username based on the email prefix
+    const username = await findUniqueUsername(email.split('@')[0]);
 
     // Create the new user.
     const newUser = {
--- a/packages/plugins/users-permissions/server/services/providers.js
+++ b/packages/plugins/users-permissions/server/services/providers.js
@@ -86,8 +86,9 @@ module.exports = ({ strapi }) => {
       .query('plugin::users-permissions.role')
       .findOne({ where: { type: advancedSettings.default_role } });
 
-    // Generate a unique username based on the email prefix
-    const username = await findUniqueUsername(email.split('@')[0]);
+    // Username: prefer profile, else email prefix; findUniqueUsername ensures valid + unique
+    const base = (profile.username && profile.username.trim()) || email.split('@')[0];
+    const username = await findUniqueUsername(base);
 
     // Create the new user.
     const newUser = {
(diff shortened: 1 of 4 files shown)


packages/plugins/users-permissions/server/services/providers.js
```diff
@@ -86,9 +86,28 @@ module.exports = ({ strapi }) => {
...
+      });
+
+    // If there are users with the same username, append a number to the username
+    if (usersWithUsername.length > 0) {
+      let counter = 1;
+      while (usersWithUsername.some(user => user.username === username)) {
+        username = `${email.split('@')[0]}${counter}`;
+        counter++;
+      }
+    }
```
innerdvations: If we have existing usernames "joe" and "joe1" and we try to register a "joe" this will still have a conflict with "joe1"

It would also be more secure to use a random number so that this doesn't leak how many users have that name.

I would move this out to a `findUniqueUsername(basename)` method and then have it verify the uniqueness in a loop after generating it, with a max attempt counter. And maybe if that counter is exceeded, we just generate a completely random uuid to use?

@FabAdrien what do you think?

ndinevski: Yes, you are totally right, I agree.

The first PR I had a couple of months back had that approach, and I had the issue appear again so I decided to open a PR again, but this time I used a counter for some reason.

ndinevski: @innerdvations I have added the changes suggested

packages/plugins/users-permissions/server/services/providers.js
```diff
@@ -86,9 +86,13 @@ module.exports = ({ strapi }) => {
       .query('plugin::users-permissions.role')
       .findOne({ where: { type: advancedSettings.default_role } });
 
+    // Generate a unique username based on the email prefix
+    const username = await findUniqueUsername(email.split('@')[0]);
```
innerdvations:
```suggestion
    // Username: prefer profile, else email prefix; findUniqueUsername ensures valid + unique
    const base = (profile.username && profile.username.trim()) || email.split('@')[0];
    const username = await findUniqueUsername(base);
```
previously we used profile.username, so we don't want to just switch to email prefix, but I think it's fine to use it as a fallback

packages/plugins/users-permissions/server/utils/index.js
```diff
@@ -1,12 +1,44 @@
...
+      .findOne({ where: { username: candidateUsername } });
+
+    if (!userExists) {
+      return candidateUsername;
+    }
+  }
+
+  // Fallback to UUID if all attempts fail
+  return crypto.randomUUID();
+};
```
innerdvations:
```suggestion
const isUsernameTaken = async (username) => {
  const user = await strapi.db
    .query('plugin::users-permissions.user')
    .findOne({ where: { username } });
  return Boolean(user);
};

const findValidUsername = async (basename) => {
  const minLength =
    strapi.getModel('plugin::users-permissions.user')?.attributes?.username?.minLength ?? 3;
  const tryBasenameFirst = basename.length >= minLength;

  let attempt = 0;
  let candidate;
  let taken;
  do {
    candidate =
      attempt === 0 && tryBasenameFirst ? basename : `${basename}${crypto.randomInt(1000, 9999)}`;
    taken = await isUsernameTaken(candidate);
    attempt += 1;
  } while (taken && attempt <= MAX_USERNAME_ATTEMPTS);

  return taken ? crypto.randomUUID() : candidate;
};
```

This is just a thought; apparently we also have a bug that usernames < 3 are accepted -- not introduced by your PR, but just made obvious here.

We could easily solve that at the same time by making this a 'findValidUsername' method and ensure that it's both padded and unique

ndinevski: Implemented and updated

Adzouz:
Hey @innerdvations, I added some tests and the code looks good to me, all subjects were addressed. Do you think we can move forward and merge? (I'll run the piplelines)

innerdvations:
I haven't had a chance to review this again, but as long as the previous comment about using the same default values as before is addressed, that was my only blocker :)

Adzouz (approved):
Let's merge it 🚀 Thanks a lot @ndinevski and sorry for the delay 🙏

PR #23855, opened 2025-07-02
Fix placeholder/default issues in color picker custom field

### LINKED ISSUE #23854: Color Picker field placeholder should not be hardcoded
### Node Version

.

### NPM/Yarn/PNPM Version

.

### Strapi Version

.

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

.

### Bug Description

The placeholder for the color picker field should not be hardcoded:

https://github.com/strapi/strapi/blob/1fedcce151b1c035980226f28bf3396436d27fde/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx#L143

It should come from the `placeholder` value set in the view configurator:

Image

### Steps to Reproduce

1. Look at code and see hardcoded placeholder.
2. Load a color picker field, set placeholder
3. See the placeholder ignored

### Expected Behavior

Custom placeholder should be displayed.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

ericbf:
### What does it do?

This commit fixes two issues in the color picker plugin.

When a placeholder is set, no value is set, and the popover is hidden, the placeholder was not actually shown (instead it was showing the default placeholder `#000000`). This fixes that by using the value of placeholder before the default placeholder.

The second issue is that when the popover is expanded, it was autofilling the field with the default placeholder value (`#000000`) regardless of whether a field value was actually set, or what the placeholder was.

### Why is it needed?

This makes users think that the field has a value of `#000000` despite the default implemented in code (and hinted by the placeholder) being something else. The configured placeholder should be shown instead.

### How to test it?

1. Add a custom field with the color picker plugin.
2. Set a placeholder in the view configurator.
3. Open the content manager and view the field
4. You should see the placeholder value, not `#000000`.

### Related issue(s)/PR(s)

Fixes #23854

### DIFF
--- a/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx
+++ b/packages/plugins/color-picker/admin/src/components/ColorPickerInput.tsx
@@ -83,7 +83,7 @@ export const ColorPickerInput = React.forwardRef<HTMLButtonElement, ColorPickerI
     const colorPickerButtonRef = React.useRef<HTMLButtonElement>(null!);
     const { formatMessage } = useIntl();
     const field = useField(name);
-    const color = field.value ?? '#000000';
+    const color = field.value ?? props.placeholder ?? '#000000';
 
     const composedRefs = useComposedRefs(forwardedRef, colorPickerButtonRef);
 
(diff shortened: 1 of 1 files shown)


hanpaine:
Hey @ericbf nice one spotting and fixing this 🙏 we'll try and take a look soon!

PR #23861, opened 2025-07-02
enhancement: refactor AI CTB to use AI SDK v5

innerdvations:
### What does it do?

Work in progress / proof of concept to upgrade to SDK v5 (requires AI Server to be upgraded as well).

If the SDK stabilizes before we release the stable AI CTB, we will try to launch with SDK v5 to avoid potential breaking changes in the future. If not, the AI server will need a versioned API before we can upgrade.

### Why is it needed?

We want to launch with the most up-to-date API if possible so we don't miss out on features/enhancements in the future

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
Scrapping this and starting with a new attempt

PR #23866, opened 2025-07-03
feat: add env variable for tracking

butcherZ:
### What does it do?

replaced hard coded url for tracking events with env variable

### Why is it needed?

make testing tracking easier

### How to test it?

add `STRAPI_ANALYTICS_URL=http://localhost` to the `.env` file at the root of `example/getStarted`
### Related issue(s)/PR(s)

DX-2088


jhoward1994:
Right now I think you'd need to add 

```
STRAPI_ANALYTICS_URL=url
STRAPI_ADMIN_ANALYTICS_URL=url
```

For it to work in both backend and frontend packages. Is that right?

If so maybe we could instead add STRAPI_ANALYTICS_URL to the default env here?

https://github.com/strapi/strapi/blob/69e3d1196ba91e28e1a2b3106258dce49cd4de54/packages/core/strapi/src/node/create-build-context.ts#L95-L99

jhoward1994 (approved):

PR #23877, opened 2025-07-03
fix: guided tour test types

markkaylor:
### What does it do?

Update reducer tests 

### Why is it needed?

Other test:ts:front fails

### How to test it?

Run yarn test:ts:front should pass


Adzouz (approved):
LGTM, tested locally and tests pass now. Waiting for the pipelines to pass too, thanks for the reactivity 🙏

PR #23885, opened 2025-07-04
fix: fix condition loss on relation field

butcherZ:
### What does it do?

- Added data preservation for both frontend and backend
- Created new `InternalRelationAttribute` and `InternalAttribute` types for CTB's internal use

### Why is it needed?

1. When applying conditional visibility to relation fields in the Content Type Builder, the conditions would not persist after saving
2. The issue only affected existing content types with relations; new content types worked correctly
3. Conditions would sometimes be saved to both sides of bidirectional relations incorrectly
4. Once saved, conditions could not be deleted - they would persist even after removal

### How to test it?

Please check:
1. Adding conditions to relation fields on existing content types
2.  Saving conditions properly persists them to schema.json
3. Conditions only appear on the intended side of bidirectional relations
4. Deleting conditions properly removes them from schema.json

### Related issue(s)/PR(s)

DX-2096


packages/core/content-type-builder/admin/src/components/FormModal/FormModal.tsx
```diff
@@ -739,8 +739,15 @@ export const FormModal = () => {
             targetUid,
           });
         } else {
+          // Ensure conditions are explicitly set to undefined if they were removed
```
Bassel17, line 742: why? and if you can the answer to that as a comment

packages/core/content-type-builder/server/src/services/schema-builder/index.ts
```diff
@@ -94,15 +94,24 @@ function createSchemaBuilder({ components, contentTypes }: SchemaBuilderOptions)
...
       };
 
       if (attribute.type === 'relation') {
-        const { target, relation, targetAttribute, dominant, ...restOfProperties } = attribute;
+        const {
+          target,
+          relation,
+          targetAttribute,
+          dominant,
+          conditions: _conditions,
```
Bassel17: you can probably remove conditions from here if it's unused

packages/core/content-type-builder/server/src/services/schema-builder/index.ts
```diff
@@ -120,6 +129,9 @@ function createSchemaBuilder({ components, contentTypes }: SchemaBuilderOptions)
         if (_.isNil(targetAttribute)) {
           return attr;
         }
+        // if (targetAttribute === null || targetAttribute === undefined) {
```
Bassel17: if not needed we can remove this commented block

Bassel17 (changes requested):
Tested looks looks good, left a few comments, and it seems there is a prettier issue

Bassel17 (approved):
LGTM!

PR #23886, opened 2025-07-04
Add zod schema validation to content api routes

jhoward1994:
### What does it do?

Add zod schema validation to content API routes

- [x] packages/core/content-type-builder/server/src/routes/content-api.ts 
- [x] packages/core/upload/server/src/routes/content-api.ts 
- [x] packages/core/email/server/src/routes/content-api.ts 
- [x] packages/plugins/i18n/server/src/routes/content-api.ts 
- [x] packages/plugins/users-permissions/server/routes/content-api/**

### Why is it needed?

Accuracy of OpenAPI spec generated by `packages/core/openapi`

### Related issue(s)/PR(s)

DX-2027

### DIFF
--- a/packages/plugins/users-permissions/server/routes/content-api/auth.js
+++ b/packages/plugins/users-permissions/server/routes/content-api/auth.js
@@ -24,7 +24,7 @@ module.exports = (strapi) => {
         prefix: '',
       },
       request: {
-        body: validator.loginBodySchema,
+        body: { 'application/json': validator.loginBodySchema },
       },
       response: validator.authResponseSchema,
     },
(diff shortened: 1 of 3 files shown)


innerdvations:
Still working on reviewing all of it, but overall looks good.

When I try running `yarn strapi openapi generate` in getstarted though, it fails with `Error: Couldn't transform the zod schema into an OpenAPI schema`

butcherZ (approved):
LGTM! QAed, and read generated specification.json using https://editor-next.swagger.io/ all works well for me!

innerdvations (approved):
LGTM!

Two non-blocking comments:

- We should find a way to avoid duplicating all these schemas and have a centralized source of truth instead, but that can hopefully be done during the yup->zod migration
- This also could use some more tests, but let's address that in the feature branch

PR #23896, opened 2025-07-06
feat(provider-email-amazon-ses): migrate to native AWS SDK v3

sjoukedv:
… of 3rd party package

### What does it do?

Currently, `node-ses` is used to send e-mail to the Amazon Simple Email Service. This PR migratie to the SES client provided and maintained by AWS itself.

### Why is it needed?

Reduce 3rd party dependencies. More features like default credential chain usage, etc.

### How to test it?

Send e-mail using SES as the configured e-mail plugin.

Please let me know if you require more changes. Would be great to also backport this into the v4 provider, since it does not conflict with Strapi core.

### DIFF
--- a/packages/providers/email-amazon-ses/package.json
+++ b/packages/providers/email-amazon-ses/package.json
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "@strapi/utils": "5.17.0",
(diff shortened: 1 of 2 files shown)


sjoukedv:
This is already changed in the v5 of the provider

PR #23903, opened 2025-07-07
feat: homepage widget entries chart

Adzouz:
### What does it do?

https://github.com/user-attachments/assets/996bb0c8-f690-455e-8d49-fcf71076823f

New widget for the homepage: Entries chart. It's a donut pie chart that displays the percentage of Draft / Published and Modified documents that the connected account user has access.
Hovering each section will give the exact count of items.
_Note: Non draft-and-publish documents types will be considered as Published documents._

### Why is it needed?

Give more statistics and information to the user about the content they have created.

#### Display the widget
-> You should see the widget displaying current information about documents

#### Create / Add a Draft entry
- Go to the Content Manager
- Pick any collection type
- Create a new entry and fill some information
- Click Save
- Go back to the homepage to see the chart updated

#### Create / Add a Published entry
- Go to the Content Manager
- Pick any collection type
- Create a new entry and fill some information
- Click Publish
- Go back to the homepage to see the chart updated

#### Create / Add a Modified entry
- Go to the Content Manager
- Pick any collection type
- Create a new entry and fill some information
- Click Publish
- Change any of the fields
- Click Save
- Go back to the homepage to see the chart updated

🚀


packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -159,4 +171,224 @@ const LastPublishedWidget = () => {
...
+            })}
+          </TextChart>
+        </text>
+      </svg>
+      {tooltip.visible && tooltip.value && (
+        <Portal
+          style={{
+            position: 'fixed',
+            left: tooltip.x + 16,
+            top: tooltip.y - 16,
```
remidej: I would recommend using transform translateX/Y here. Unlike positioning with x and y, it doesn't trigger browser repaints when it's moved, so it's better for performance

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -159,4 +171,224 @@ const LastPublishedWidget = () => {
...
+          <Box
+            background="neutral900"
+            padding={2}
+            borderRadius={1}
+            textAlign="center"
+            role="tooltip"
+            aria-live="polite"
+          >
+            <Typography textColor="neutral0">
+              {tooltip.value.count} {tooltip.value.label}
```
remidej: I think this whole line should be 1 formatMessage that takes count and label as params

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -301,10 +301,12 @@ const DonutChartSVG = ({ data }: { data: ChartData[] }) => {
         <Portal
           style={{
             position: 'fixed',
-            left: tooltip.x + 16,
-            top: tooltip.y - 16,
+            left: tooltip.x,
+            top: tooltip.y,
             zIndex: 2,
+            transform: 'translate(1.6rem, 1.6rem)',
           }}
```
remidej: my point was that the _dynamic_ part (tooltip.x and tooltip.y) should be the one in the transform, because it's the transform that super cheap for the browser to render when it updates

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -371,17 +382,26 @@ const ChartEntriesWidget = () => {
       <DonutChartSVG
         data={[
           {
-            label: 'Draft',
+            label: formatMessage({
+              id: 'content-manager.containers.List.draft',
+              defaultMessage: 'Draft',
+            }),
```
remidej, line 392: Is this label displayed on its own anywhere, or is it only shown in the string with the count?

Adzouz: Both. It's used for the keys below the chart + the string with the count

remidej (approved):
🚀

UsmanHaider546:
I have a questions regarding strapi use packages. when i install strapi it gives warning on some packages with npm warn. and sometimes those packages that not supported. please use clean and fully supported packages in next time.

romain-hasseveldt:
The widget seemed to work fine on my SQLite database, but it turns out there are errors when using a PostgreSQL database. One or several queries appear to cause issues and raise a PostgreSQL protocol error.

Capture d’écran 2025-07-20 à 15 59 23

Capture d’écran 2025-07-20 à 16 00 02

Capture d’écran 2025-07-20 à 16 00 12

Capture d’écran 2025-07-20 à 16 01 38

PR #23905, opened 2025-07-07
future: add the Popover.Arrow in the guided tour

simotae14:
### What does it do?

Adds the Popover.Arrow to the Guided tour popovers

### Why is it needed?

it is required by design

### How to test it?

- Clean the local storage and create a new Strapi project from scratch without collectiions
- try to create new Collections following the Guided tour
- and add new content using the Content manager
- and create a new Api token

use the experimental 0.0.0-experimental.f37a0ea0333947b6879c4608fae75646d63c07a1  to create the project

### DIFF
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
@@ -62,7 +62,7 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
         <Flex width="360px" direction="column" alignItems="start">
           {props.children}
         </Flex>
-        {currentStep > 1 && <PopoverArrowStandard />}
+        {currentStep > 1 && <PopoverArrowStandard width="20px" height="12px" />}
       </Popover.Content>
     );
   }),
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
@@ -55,12 +55,14 @@ const PopoverArrowStandard = styled(Popover.Arrow)`
 
 const createStepComponents = (tourName: ValidTourName): Step => ({
   Root: React.forwardRef((props, ref) => {
+    const state = unstableUseGuidedTour('GuidedTourPopover', (s) => s.state);
+    const currentStep = state.tours[tourName].currentStep + 1;
     return (
       <Popover.Content ref={ref} side="top" align="center" {...props}>
         <Flex width="360px" direction="column" alignItems="start">
           {props.children}
         </Flex>
-        <PopoverArrowStandard />
+        {currentStep > 1 && <PopoverArrowStandard />}
       </Popover.Content>
     );
   }),
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
@@ -49,7 +49,7 @@ const ActionsContainer = styled(Flex)`
   border-top: ${({ theme }) => `1px solid ${theme.colors.neutral150}`};
 `;
 
-const PopoverArrowStandard = styled(Popover.Arrow)`
+const PopoverArrow = styled(Popover.Arrow)`
   fill: ${({ theme }) => `${theme.colors.neutral0}`};
 `;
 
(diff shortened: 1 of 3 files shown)


packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -49,14 +49,23 @@ const ActionsContainer = styled(Flex)`
...
-  )),
+  Root: React.forwardRef((props, ref) => {
+    const state = unstableUseGuidedTour('GuidedTourPopover', (s) => s.state);
+    const currentStep = state.tours[tourName].currentStep + 1;
+    return (
+      <Popover.Content ref={ref} side="top" align="center" {...props}>
+        <Flex width="360px" direction="column" alignItems="start">
+          {props.children}
+        </Flex>
+        {currentStep > 1 && <PopoverArrowStandard width="20px" height="12px" />}
```
markkaylor: I see what you did there, makes sense for our current setup but I think still would be better to either
- add a prop like `withArrow` that defaults to true since most cases have the arrow, and then set it to false on the ones that don't have it
- rely on composition and add the PopoverArrow directly in the steps that need it

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -49,14 +49,23 @@ const ActionsContainer = styled(Flex)`
   border-top: ${({ theme }) => `1px solid ${theme.colors.neutral150}`};
 `;
 
+const PopoverArrowStandard = styled(Popover.Arrow)`
+  fill: ${({ theme }) => `${theme.colors.neutral0}`};
+`;
```
markkaylor: I think we can just call is PopoverArrow

packages/core/admin/admin/src/components/UnstableGuidedTour/Tours.tsx
```diff
@@ -203,7 +201,7 @@ const tours = {
     {
       name: 'Fields',
       content: (Step) => (
-        <Step.Root side={'top'} align="start" sideOffset={-36}>
+        <Step.Root side={'top'} sideOffset={-36}>
```
markkaylor: I think it should just be the default side top center and I have a hunch the sideOffset is not necessary anymore but haven't tested yet. 

```suggestion
        <Step.Root>
```

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -72,16 +80,23 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
+          <Typography tag="div" variant="omega">
+            <FormattedMessage
+              tagName="p"
+              id={restProps.id}
+              defaultMessage={restProps.defaultMessage}
+            />
+          </Typography>
+        )}
+      </Box>
+      {withArrow && <PopoverArrow />}
```
markkaylor: This is fine, but I think I prefer it in Popover.Root because that is actually the Popover wrapping title, content, and actions. With the props on content it's not that clear where the arrow will be applied.

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -49,14 +51,20 @@ const ActionsContainer = styled(Flex)`
   border-top: ${({ theme }) => `1px solid ${theme.colors.neutral150}`};
 `;
 
+const PopoverArrow = styled(Popover.Arrow)`
+  fill: ${({ theme }) => `${theme.colors.neutral0}`};
+`;
+
 const createStepComponents = (tourName: ValidTourName): Step => ({
-  Root: React.forwardRef((props, ref) => (
-    <Popover.Content ref={ref} side="top" align="center" style={{ border: 'none' }} {...props}>
```
markkaylor, line 54: Please don't remove the border none

simotae14: yeah my fault sorry

markkaylor:
@simotae14 Tested on experimental and all looks good to me but I'll let you test as well: 0.0.0-experimental.8ed7454f698016398d3ac8209f399206cfe8a6bb before we merge

simotae14:
tested and it works!!! yuppie

PR #23906, opened 2025-07-07
fix: yarn build

### LINKED ISSUE #23859: error TS2688: Cannot find type definition file for 'minimatch'
### Node Version

20.0

### NPM/Yarn/PNPM Version

npm 10

### Strapi Version

5.17.0

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Yarn build is failing as: error TS2688: Cannot find type definition file for 'minimatch'.

```
yarn run v1.22.22
$ strapi build
- Compiling TS
error TS2688: Cannot find type definition file for 'minimatch'.
  The file is in the program because:
    Entry point for implicit type library 'minimatch'

Found 1 error(s).

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
Error: building at STEP "RUN yarn build": while running runtime: exit status 1
``


### Steps to Reproduce

Pull latest strapi
run yarn build

### Expected Behavior

Should build fine without errors

### Logs

```shell
yarn run v1.22.22
$ strapi build
- Compiling TS
error TS2688: Cannot find type definition file for 'minimatch'.
  The file is in the program because:
    Entry point for implicit type library 'minimatch'

Found 1 error(s).

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
Error: building at STEP "RUN yarn build": while running runtime: exit status 1
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

Bassel17:
### What does it do?
refactors the generators to use 'plop' only

### Why is it needed?

dependency issues #23859 
see DX-2094

### How to test it?

Use experimental as such to create a project
`npx create-strapi-app@0.0.0-experimental.992051e9ba86a828182a248c3661b3a7e59fe389 testing-build --use-yarn`
in the project run `yarn build` it should work
you need to run with `--use-yarn` and be using yarn v1 to confirm 

### Related issue(s)/PR(s)

fix #23859


packages/generators/generators/src/plops/prompts/ct-names-prompts.ts
```diff
@@ -22,7 +20,7 @@ const questions: Array<PromptQuestion> = [
     name: 'singularName',
     message: 'Content type singular name',
     default: (answers: Answers) => slugify(answers.displayName),
-    validate(input) {
+    validate(input: any) {
```
butcherZ: could this be a string type instead of any? :D

packages/generators/generators/src/plops/prompts/get-attributes-prompts.ts
```diff
@@ -65,13 +64,13 @@ const getAttributesPrompts: DynamicPromptsFunction = async (inquirer) => {
         }),
       },
       {
-        when: (answers) => answers.attributeType === 'enumeration',
+        when: (answers: any) => answers.attributeType === 'enumeration',
```
butcherZ: I feel like for answers since we know the literal types maybe we can create an interface for it? 
```
export interface DestinationAnswers {
    destination: 'root' | 'new' | 'api' | 'plugin';
    api?: string;      only present when destination is 'api'
    plugin?: string;   only present when destination is
  'plugin'
  }
```
something like this.

jhoward1994:
The build works for me ✅

I don't have loads of context on this work but looking at the code my concerns would be:

- added complexity in `packages/generators/generators/src/index.ts` - maybe we could add comments throughout the new process to explain what is happening
- lost type accuracy throughout other files - can we type better than `any` in some of these cases

We've said in the ticket "ideally... refactor out node-plop and get the same functionality from Plop instead" & this has probably been investigated so just a thought but did we look into more recent versions of `node-plop` in https://www.npmjs.com/package/node-plop?activeTab=versions ?

remidej (approved):
I don't have much context to judge the technical solution in depth, but it looks okay to me. I did QA though:
- creating an app ✅
- building an app ✅
- using generators (content type etc.) ✅

Bassel17:
> The build works for me ✅
> 
> I don't have loads of context on this work but looking at the code my concerns would be:
> 
> * added complexity in `packages/generators/generators/src/index.ts` - maybe we could add comments throughout the new process to explain what is happening
> * lost type accuracy throughout other files - can we type better than `any` in some of these cases
> 
> We've said in the ticket "ideally... refactor out node-plop and get the same functionality from Plop instead" & this has probably been investigated so just a thought but did we look into more recent versions of `node-plop` in https://www.npmjs.com/package/node-plop?activeTab=versions ?

yes we were using the latest version from `node-plop` it's more of an issue with a sub package dependency that `node-plop` uses `minimatch` there is no way to resolve it except by users resolving it in their package.json, but we want a seemless way for this to work, and considering it's not that complicated, we are just writing template files, taking `node-plop` out was the best way I think we can go

PR #23908, opened 2025-07-07
fix: add reducer to be able to mutate widgets in register function

Adzouz:
fix: put widgets from content-manager first in list
fix: put review-workflows widget last in list

### What does it do?

Provides a reducer to the register function of the widgets in order to allow mutations on it.
The purpose here is to be able to reorder the widgets by using the same method than for the addDocumentAction (to reorder items in action menus in the content manager -> ex: history action before delete actions).
_The two widgets from the content-manager (last edited entries and last published entries need to always be the two first widgets on the homepage)._

### Why is it needed?

Be able to alter the list of widgets on the homepage by using the register function.

### How to test it?

-> The list of widget should be in the following order:
  - Last edited entries
  - Last published entries
  - Profile
  - Entries (chart) -> when https://github.com/strapi/strapi/pull/23903 will be merged
  - Assigned to me

🚀

### DIFF
--- a/packages/core/admin/admin/src/core/apis/Widgets.ts
+++ b/packages/core/admin/admin/src/core/apis/Widgets.ts
@@ -29,6 +29,8 @@ type WidgetArgs = {
 
 type Widget = Omit<WidgetArgs, 'id' | 'pluginId'> & { uid: WidgetUID };
 
+type DescriptionReducer<T extends object> = (prev: Record<string, T>) => Record<string, T>;
+
 class Widgets {
   widgets: Record<string, Widget>;
 
@@ -36,19 +38,21 @@ class Widgets {
     this.widgets = {};
   }
 
-  register = (widget: WidgetArgs | WidgetArgs[]) => {
-    if (Array.isArray(widget)) {
-      widget.forEach((newWidget) => {
+  register = (widgets: WidgetArgs | WidgetArgs[] | DescriptionReducer<Widget>) => {
+    if (Array.isArray(widgets)) {
+      widgets.forEach((newWidget) => {
         this.register(newWidget);
       });
+    } else if (typeof widgets === 'function') {
+      this.widgets = widgets(this.widgets);
     } else {
-      invariant(widget.id, 'An id must be provided');
-      invariant(widget.component, 'A component must be provided');
-      invariant(widget.title, 'A title must be provided');
-      invariant(widget.icon, 'An icon must be provided');
+      invariant(widgets.id, 'An id must be provided');
+      invariant(widgets.component, 'A component must be provided');
+      invariant(widgets.title, 'A title must be provided');
+      invariant(widgets.icon, 'An icon must be provided');
 
       // Replace id and pluginId with computed uid
-      const { id, pluginId, ...widgetToStore } = widget;
+      const { id, pluginId, ...widgetToStore } = widgets;
       const uid: WidgetUID = pluginId ? `plugin::${pluginId}.${id}` : `global::${id}`;
 
       this.widgets[uid] = { ...widgetToStore, uid };
--- a/packages/core/content-manager/admin/src/index.ts
+++ b/packages/core/content-manager/admin/src/index.ts
@@ -8,6 +8,8 @@ import { previewAdmin } from './preview';
 import { routes } from './router';
 import { prefixPluginTranslations } from './utils/translations';
 
+import type { WidgetType } from '@strapi/admin/strapi-admin';
+
 // NOTE: we have to preload it to ensure chunks will have it available as global
 import 'prismjs';
 
@@ -76,6 +78,19 @@ export default {
         permissions: [{ action: 'plugin::content-manager.explorer.read' }],
       },
     ]);
+
+    // Always put the last-edited-entries and last-published-entries widgets first in the list of widgets
+    app.widgets.register((widgets: Record<string, WidgetType>) => {
+      const desiredOrder = [
+        'plugin::content-manager.last-edited-entries',
+        'plugin::content-manager.last-published-entries',
+      ];
+      const ordered = [
+        ...desiredOrder.map((uid) => widgets[uid]).filter(Boolean),
+        ...Object.values(widgets).filter((w) => !desiredOrder.includes(w.uid)),
+      ];
+      return Object.fromEntries(ordered.map((w) => [w.uid, w]));
+    });
   },
   bootstrap(app: any) {
     if (typeof historyAdmin.bootstrap === 'function') {
--- a/packages/core/review-workflows/admin/src/index.ts
+++ b/packages/core/review-workflows/admin/src/index.ts
@@ -6,7 +6,7 @@ import { Panel } from './routes/content-manager/model/id/components/Panel';
 import { addColumnToTableHook } from './utils/cm-hooks';
 import { prefixPluginTranslations } from './utils/translations';
 
-import type { StrapiApp } from '@strapi/admin/strapi-admin';
+import type { StrapiApp, WidgetType } from '@strapi/admin/strapi-admin';
 import type { Plugin } from '@strapi/types';
 
 const admin: Plugin.Config.AdminInput = {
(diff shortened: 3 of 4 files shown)


packages/core/content-manager/admin/src/index.ts
```diff
@@ -76,6 +78,19 @@ export default {
...
+    app.widgets.register((widgets: Record<string, WidgetType>) => {
+      const desiredOrder = [
+        'plugin::content-manager.last-edited-entries',
+        'plugin::content-manager.last-published-entries',
+      ];
+      const ordered = [
+        ...desiredOrder.map((uid) => widgets[uid]).filter(Boolean),
+        ...Object.values(widgets).filter((w) => !desiredOrder.includes(w.uid)),
+      ];
+      return Object.fromEntries(ordered.map((w) => [w.uid, w]));
```
remidej: I'm just realizing now that our this.widgets store is an object, not an array. So your logic is a bit more complex than the document actions one because of the Object.values + Object.fromEntries.And more importantly, in javascript the order of the entries in an object is not guaranteed, so we shouldn't rely on it.

Could you update apis/Widgets.ts so this.widgets stores an array instead of an object?

packages/core/review-workflows/admin/src/index.ts
```diff
@@ -37,22 +37,30 @@ const admin: Plugin.Config.AdminInput = {
...
+
+      // Always put the assigned to me widget last in the list of widgets
+      app.widgets.register((widgets: Record<string, WidgetType>) => {
+        const assignedPluginUid = 'plugin::review-workflows.assigned';
+        const ordered = [
+          ...Object.values(widgets).filter((w) => !assignedPluginUid.includes(w.uid)),
+          widgets['plugin::review-workflows.assigned'],
+        ];
+        return Object.fromEntries(ordered.map((w) => [w.uid, w]));
+      });
```
remidej: Here you're calling `register()` once to register the widget, and then a second time to change the widgets order. Instead you should call it a single time to add the widget precisely in the position you want it

Adzouz: Yes that's because as you said it's using an object and not an array so it was the most readable / easy way to do it: register them first to create the key + value in the widgets object + mutate it afterwards with the reducer but I get that it's not the best. I could try to update the logic to use an array instead as you proposed. I'll let you know.

packages/core/content-manager/admin/src/index.ts
markkaylor: That's weird I thought we did switch to an array 🤔 . Could also consider a [Map too which does maintain entry order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -29,34 +29,68 @@ type WidgetArgs = {
...
+    if (Array.isArray(widgets)) {
+      this.widgets.push(...this.addUidToWidgets(widgets));
+    } else if (typeof widgets === 'function') {
+      const result = widgets(this.widgets);
+      this.widgets = this.addUidToWidgets(result);
+    } else if (typeof widgets === 'object') {
+      this.widgets.push(this.addUidToWidgets([widgets])[0]);
+    } else {
+      throw new Error('Expected widgets to be an array or a reducer function');
     }
```
markkaylor: Not sure I have the whole context here but I thought eventually we want give the control of the order of the widgets to the user. 

If so I'm not sure this is the API we want. In my opinion we should think about the interface we want to expose to the user and build that first. If the reordering thing is blocking something we want to build then all the more reason to do it now, no? Once we expose the widget register function accepting a function as an argument we are stuck with it until a major release. Personally I'm not really a fan.

Adzouz: I agree but @remidej had a good point that it's another complexity and would be more complicated to design. But maybe we could find kind of an in between @remidej? As you said @markkaylor, maybe prepare at least the structure and not register anything for the moment in the database. But I don't know yet how those kind of things are structured. We could discuss that together.

soupette: I guess the ordering issue also depends on the loading of the plugins order. I'm not sure we have solve the issue already, did we?

markkaylor: I haven't checked but I have a hunch too that right now the ordering depends on plugin load order.

soupette: well, another approach could be to introduce a new widget api next to this one, maybe you could define an export in each plugin package.json

```
exports: {
 "./widget": "path to widget"
}
```

When bundling the app for the first time you retrieve all the widgets based on the package.json exports.widget and you generate an ordered array of widgets that you expose to the end user that he could then reorder as he wishes.

This could also be the opportunity to avoid storing all the widgets in memory but in the file system instead which will improve performances as well as the final bundle.

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -29,34 +29,68 @@ type WidgetArgs = {
...
+      widgetArgs.pluginId
+        ? `plugin::${widgetArgs.pluginId}.${widgetArgs.id}`
+        : `global::${widgetArgs.id}`
+    ) as WidgetUID;
+  };
 
-      // Replace id and pluginId with computed uid
-      const { id, pluginId, ...widgetToStore } = widget;
-      const uid: WidgetUID = pluginId ? `plugin::${pluginId}.${id}` : `global::${id}`;
+  private addUidToWidgets = (widgets: (WidgetArgs | Widget)[]): Widget[] =>
```
remidej: The logic and types would all be simpler if you generated the UID dynamically on getAll() instead of on registration

packages/core/content-manager/admin/src/index.ts
Adzouz: We discussed finally to keep the array.

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -27,38 +27,65 @@ type WidgetArgs = {
   permissions?: Array<Pick<Permission, 'action'> & Partial<Omit<Permission, 'action'>>>;
 };
 
-type Widget = Omit<WidgetArgs, 'id' | 'pluginId'> & { uid: WidgetUID };
+type WidgetWithUID = Omit<WidgetArgs, 'id' | 'pluginId'> & { uid: WidgetUID };
+
+type DescriptionReducer<T extends object> = (prev: T[]) => T[];
```
remidej: I feel like the reducer type could just be `(prev: WidgetArgs) => WidgetArgs[]`, I don't see why we would need a generic

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -27,38 +27,65 @@ type WidgetArgs = {
...
 
-  register = (widget: WidgetArgs | WidgetArgs[]) => {
-    if (Array.isArray(widget)) {
-      widget.forEach((newWidget) => {
-        this.register(newWidget);
-      });
-    } else {
+  private generateUid = (widget: WidgetArgs): WidgetUID => {
+    // At this point, widget must be WidgetArgs (has id and pluginId)
+    const widgetArgs = widget as WidgetArgs;
```
remidej: It looks like the type casting here may be left from what you had before, because widget is already of WidgetArgs type here

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -27,38 +27,65 @@ type WidgetArgs = {
...
 
-      // Replace id and pluginId with computed uid
-      const { id, pluginId, ...widgetToStore } = widget;
-      const uid: WidgetUID = pluginId ? `plugin::${pluginId}.${id}` : `global::${id}`;
-
-      this.widgets[uid] = { ...widgetToStore, uid };
+  register(widgets: WidgetArgs): void;
+  register(widgets: WidgetArgs[]): void;
+  register(widgets: DescriptionReducer<WidgetArgs>): void;
+  register(widgets: WidgetArgs | WidgetArgs[] | DescriptionReducer<WidgetArgs>) {
```
remidej: if you provide the type union here then are the 3 lines above with the different signatures needed?

remidej (commented):
just some TS comments

packages/core/admin/admin/src/core/apis/Widgets.ts
Adzouz: That's totally true haha

packages/core/admin/admin/src/core/apis/Widgets.ts
Adzouz: We decided for the moment to stick with what was done (copying the logic of the addDocumentAction) and in the future, whenever we'll have to implement the configure the view for the user, we'll redesign this api. Thanks a lot for all your ideas, it was really enlightening! 🚀🙏

PR #23924, opened 2025-07-08
feat: add validation for content API routes and upload plugin

butcherZ:
### What does it do?

 - Created `AbstractRouteValidator` base class providing standard query parameter validators, so other plugins can extend from this class
  - Extended `UploadRouteValidator` from the new base class added validation for upload plugin
  - Created schema for content type and component structures
  - test for upload
  
### Why is it needed?

### How to test it?

run `yarn strapi generate` and copy the generated `specification.json` to https://editor.swagger.io/, the following routes should behave like below: 

  1. POST /api/upload
    - Query parameter id (optional, type: number)
    - Response schema showing file or array of files
  2. GET /api/upload/files
    - Query parameters: fields, populate, sort, pagination, filters
    - Each parameter should have proper type definitions
    - Response should be array type
  3. GET /api/upload/files/{id}
    - Path parameter id (required, type: number)
    - Query parameters: fields, populate
    - Response should be object type
  4. DELETE /api/upload/files/{id}
    - Path parameter id (required, type: number)
    - Response schema for deleted file
  5. GET /api/content-type-builder/content-types
    - Query parameter kind (REQUIRED, enum: ['collectionType', 'singleType'])
    - Response: { data: array of content types }
  6. GET /api/content-type-builder/content-types/{uid}
    - Path parameter uid (REQUIRED, string matching regex)
    - Response: { data: single content type }
  7. GET /api/content-type-builder/components
    - NO parameters (no request validation defined)
    - Response: { data: array of components }
  8. GET /api/content-type-builder/components/{uid}
    - Path parameter uid (REQUIRED, string matching regex)
    - Response: { data: single component }

### Related issue(s)/PR(s)

dx-2027

### DIFF
--- a/packages/core/utils/src/convert-query-params.ts
+++ b/packages/core/utils/src/convert-query-params.ts
@@ -693,7 +693,6 @@ const createTransformer = ({ getModel }: TransformerOptions) => {
 
     const query: Query = {};
 
-    // eslint-disable-next-line @typescript-eslint/naming-convention
     const { _q, sort, filters, fields, populate, page, pageSize, start, limit, status, ...rest } =
       params;
 
--- a/packages/plugins/documentation/server/src/controllers/documentation.ts
+++ b/packages/plugins/documentation/server/src/controllers/documentation.ts
@@ -18,9 +18,9 @@ const validation = {
       password: yup
         .string()
         .min(8)
-        .matches(/[a-z]/, `${path} must contain at least one lowercase character`)
-        .matches(/[A-Z]/, `${path} must contain at least one uppercase character`)
-        .matches(/\d/, `${path} must contain at least one number`)
+        .matches(/[a-z]/, '${path} must contain at least one lowercase character')
+        .matches(/[A-Z]/, '${path} must contain at least one uppercase character')
+        .matches(/\d/, '${path} must contain at least one number')
         .when('restrictedAccess', (value, initSchema) => {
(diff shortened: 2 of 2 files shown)


packages/core/content-type-builder/server/src/routes/content-api.ts
```diff
@@ -1,8 +1,95 @@
...
+// Content type schema
+const contentTypeSchemaBase = z.object({
+  displayName: z.string(),
+  singularName: z.string(),
+  pluralName: z.string(),
+  description: z.string(),
+  draftAndPublish: z.boolean(),
+  kind: z.enum(['collectionType', 'singleType']),
+  collectionName: z.string().optional(),
+  attributes: z.record(z.string(), attributeSchema),
```
jhoward1994, line 86: I don't think attributeSchema covers components. e.g. something in the form

```
"notrepeat_req": {
    "type": "component",
    "repeatable": false,
    "pluginOptions": {},
    "component": "blog.test-como",
    "required": true
},

```

butcherZ: yep, you are absolutely right :')

jhoward1994:
Changed base to `feat/openapi-route-schemas`, I think it makes sense to collect the changes in there first

jhoward1994 (approved):

PR #23926, opened 2025-07-09
future: Guided Tour Tracking

simotae14:
### What does it do?

Adds the tracking Events in the new guided tour

### Why is it needed?

We want to keep track of the guided tour usage

### How to test it?

Here the list of the events registered
- `didSkipGuidedTour` with tour name or 'all' as payload
- `didStartGuidedTourFromHomepage` with tour name as payload
- `didCompleteGuidedTour` with tour name as payload


packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -111,13 +115,19 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
     </Box>
   ),
 
-  Actions: ({ showStepCount = true, showSkip = false, to, ...props }) => {
+  Actions: ({ showStepCount = true, showSkip = false, to, trackedEvent, ...props }) => {
```
markkaylor: I don't really like this as a prop here and on the TASK_CONTENT constant. 

What if we made a new shared constant like 
```
GUIDED_TOUR_TRACKING = {
 contentTypeBuilder: { create: 'didCreateGuidedTourContentTypeSchema', start: 'didStartGuidedTourContentTypeBuilder'}
}
```
Then we can look up the tracking with the tourName
```
const trackingEvent = GUIDED_TOUR_TRACKING['contentTypeBuilder'].create
```
I think this way if we need to add or update a tracking event it will only require a change in one place vs two

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -111,13 +115,19 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
+    const { trackUsage } = useTracking();
     const dispatch = unstableUseGuidedTour('GuidedTourPopover', (s) => s.dispatch);
     const state = unstableUseGuidedTour('GuidedTourPopover', (s) => s.state);
     const currentStep = state.tours[tourName].currentStep + 1;
     // TODO: Currently all tours do not count their last step, but we should find a way to make this more smart
     const displayedLength = state.tours[tourName].length - 1;
 
+    const handleSkipAction = () => {
+      trackUsage('didSkipGuidedtour');
+      dispatch({ type: 'skip_tour', payload: tourName });
```
markkaylor, line 150: Is this supposed to be tracking for skipping a single tour or all tours? We should clarify that with product. Maybe we need a new event?

simotae14: Right now a single tour and also the Close guided tour in the homepage

simotae14: we added a new event for the skip in the Popovers

simotae14:
@markkaylor I am not sure about the first event 'didLaunchGuidedtour' implementation, for the rest I am quite confident

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
```diff
@@ -104,6 +106,12 @@ const UnstableGuidedTourContext = ({
     }
   }, [state, setTours]);
 
+  React.useEffect(() => {
+    if (enabled) {
+      trackUsage('didLaunchGuidedtour');
+    }
+  }, [enabled, trackUsage]);
```
markkaylor: I'm gonna check with Yannis but this event doesn't make much sense to me. I think the didClickGuidedTourX tracking should be enough

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -112,12 +129,18 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
   Actions: ({ showStepCount = true, showSkip = false, to, ...props }) => {
+    const { trackUsage } = useTracking();
     const dispatch = unstableUseGuidedTour('GuidedTourPopover', (s) => s.dispatch);
     const state = unstableUseGuidedTour('GuidedTourPopover', (s) => s.state);
     const currentStep = state.tours[tourName].currentStep + 1;
     // TODO: Currently all tours do not count their last step, but we should find a way to make this more smart
     const displayedLength = state.tours[tourName].length - 1;
 
+    const handleSkipAction = () => {
+      trackUsage('didSkipGuidedTourStep', { from: tourName });
```
markkaylor: We need to check with Yannis because we are skipping a feature tour not a step in a feature tour.

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -48,6 +50,21 @@ type Step = {
   Actions: (props: ActionsProps & { to?: string }) => React.ReactNode;
 };
 
+const GUIDED_TOUR_TRACKING_EVENTS = {
+  contentTypeBuilder: {
+    create: 'didCreateGuidedTourCollectionType' as EventWithoutProperties['name'],
```
markkaylor: I would add `start: 'didClickGuidedTourHomepageContentManager'` and then use the constant in the overview so all tracking is centralized

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -135,23 +158,28 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
               )}
               {to ? (
                 <LinkButton
                   tag={NavLink}
                   to={to}
-                  onClick={() => dispatch({ type: 'next_step', payload: tourName })}
+                  onClick={() => {
+                    trackUsage(GUIDED_TOUR_TRACKING_EVENTS[tourName]?.create);
+                    dispatch({ type: 'next_step', payload: tourName });
+                  }}
```
markkaylor: If you are creating a handler for skip maybe create one for this too? Or do them both inline? I think consistency one way or another would be nice

packages/core/admin/admin/src/components/UnstableGuidedTour/Step.tsx
```diff
@@ -135,23 +158,28 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
                 >
                   <FormattedMessage id="tours.next" defaultMessage="Next" />
                 </LinkButton>
               ) : (
-                <Button onClick={() => dispatch({ type: 'next_step', payload: tourName })}>
+                <Button
+                  onClick={() => {
+                    trackUsage(GUIDED_TOUR_TRACKING_EVENTS[tourName]?.create);
+                    dispatch({ type: 'next_step', payload: tourName });
+                  }}
```
markkaylor: same comment as above

packages/core/admin/admin/src/components/UnstableGuidedTour/Overview.tsx
```diff
@@ -77,6 +92,7 @@ const TASK_CONTENT = [
       defaultMessage: 'Create your schema',
     },
     done: DONE_LABEL,
+    trackingEvent: 'didClickGuidedTourHomepageContentTypeBuilder',
```
markkaylor: I'd use the tracking constant instead of adding a property here.

packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
```diff
@@ -79,6 +80,7 @@ const UnstableGuidedTourContext = ({
   children: React.ReactNode;
   enabled?: boolean;
 }) => {
+  const { trackUsage } = useTracking();
```
markkaylor: Could you remove this

markkaylor (approved):

PR #23929, opened 2025-07-09
fix: admin constantly reloading in development

### LINKED ISSUE #23916: Admin keeps reloading the page while running monorepo example apps
### Bug Description

When running getstarted or any of the monorepo example apps, the window keeps reloading upon page navigation. This is quite frustrating for the development experience. Thankfully it does not affect end users, just Strapi contributors.

It happens because of [Vite discovering new dependencies that it needs to optimize](https://vite.dev/config/dep-optimization-options.html#optimizedeps-include). We use code splitting, so dependencies are often found when navigating from page to page. If they don't use ES modules, Vite will need to optimize them as it encounters them. Then it needs to reload.

### Potential fixes

There are 2 main angles to fix this issue.
- We can tell Vite to pre-optimize some dependencies. So they'll be in the initial bundle and won't cause a reload when they're encountered. This bundle size change only makes a difference on the monorepo development experience, it does not affect end Strapi apps.
- We can avoid having to optimize dependencies by ensuring they use ES modules. This would be fixing the problem at its root. But it's more work and not all npm packages support it yet.

The first step is to make a list of the problematic dependencies. When you notice your admin reloading for no reason, check your terminal logs, you'll see a Vite message like this:

```sh
11:32:22 AM [vite] ✨ new dependencies optimized: lodash/fp
11:32:22 AM [vite] ✨ optimized dependencies changed. reloading
```

Make note of the dependency, because we need to build a list to ensure we cover them all. Here's the list I'll keep updating as we encounter more:

```ts
const depsThatReload = [
  '@dnd-kit/core',
  '@dnd-kit/sortable',
  '@dnd-kit/utilities',
  '@dnd-kit/modifiers',
  '@radix-ui/react-toolbar',
  'codemirror5',
  'codemirror5/addon/display/placeholder',
  'date-fns-tz',
  'date-fns/format',
  'date-fns/formatISO',
  'highlight.js',
  'lodash/capitalize',
  'lodash/fp',
  'lodash/groupBy',
  'lodash/has',
  'lodash/isNil',
  'lodash/locale',
  'lodash/map',
  'lodash/mapValues',
  'lodash/pull',
  'lodash/size',
  'lodash/sortBy',
  'lodash/tail',
  'lodash/toLower',
  'lodash/toNumber',
  'lodash/toString',
  'lodash/truncate',
  'lodash/uniq',
  'lodash/upperFirst',
  'markdown-it',
  'markdown-it-abbr',
  'markdown-it-container',
  'markdown-it-deflist',
  'markdown-it-emoji',
  'markdown-it-footnote',
  'markdown-it-ins',
  'markdown-it-mark',
  'markdown-it-sub',
  'markdown-it-sup',
  'prismjs/components/prism-asmatmel',
  'prismjs/components/prism-bash',
  'prismjs/components/prism-basic',
  'prismjs/components/prism-c',
  'prismjs/components/prism-clojure',
  'prismjs/components/prism-cobol',
  'prismjs/components/prism-cpp',
  'prismjs/components/prism-csharp',
  'prismjs/components/prism-dart',
  'prismjs/components/prism-docker',
  'prismjs/components/prism-elixir',
  'prismjs/components/prism-erlang',
  'prismjs/components/prism-fortran',
  'prismjs/components/prism-fsharp',
  'prismjs/components/prism-go',
  'prismjs/components/prism-graphql',
  'prismjs/components/prism-groovy',
  'prismjs/components/prism-haskell',
  'prismjs/components/prism-haxe',
  'prismjs/components/prism-ini',
  'prismjs/components/prism-java',
  'prismjs/components/prism-javascript',
  'prismjs/components/prism-json',
  'prismjs/components/prism-jsx',
  'prismjs/components/prism-julia',
  'prismjs/components/prism-kotlin',
  'prismjs/components/prism-latex',
  'prismjs/components/prism-lua',
  'prismjs/components/prism-makefile',
  'prismjs/components/prism-markdown',
  'prismjs/components/prism-matlab',
  'prismjs/components/prism-objectivec',
  'prismjs/components/prism-perl',
  'prismjs/components/prism-php',
  'prismjs/components/prism-powershell',
  'prismjs/components/prism-python',
  'prismjs/components/prism-r',
  'prismjs/components/prism-ruby',
  'prismjs/components/prism-rust',
  'prismjs/components/prism-sas',
  'prismjs/components/prism-scala',
  'prismjs/components/prism-scheme',
  'prismjs/components/prism-sql',
  'prismjs/components/prism-stata',
  'prismjs/components/prism-swift',
  'prismjs/components/prism-tsx',
  'prismjs/components/prism-typescript',
  'prismjs/components/prism-vbnet',
  'prismjs/components/prism-yaml',
  'react-colorful',
  'react-dnd-html5-backend',
  'react-window',
  'sanitize-html',
  'semver',
  'semver/functions/lt',
  'semver/functions/valid',
  'slate',
  'slate-history',
  'slate-react',
]
```

remidej:
### What does it do?

Fixes the issue where the admin would frequently reload while running the monorepo getstarted apps.

I'm doing the fix in the [simplest and dumbest way](https://vite.dev/config/dep-optimization-options.html#optimizedeps-include) possible because it solves the reload issue. But it's more of a hot patch. The root of the issue is that we have many dependencies that don't use ES Modules. 

**I wrote more about it in the issue: #23916**

### How to test it?

`yarn setup` of course. Then run the getstarted app, navigate from page to page. It should not randomly reload. If you do see the admin reloading, check your terminal for a `✨ new dependencies optimized` log, and let me know the problematic ones.

If needed, running `yarn build` inside the core/strapi repo seems to be the only thing that actually clears Vite's optimized dependencies cache.

### Related issue(s)/PR(s)

fixes #23916


markkaylor (approved):
Tested. This is awesome. Thank you.

PR #23934, opened 2025-07-10
chore: update cloud plugin to add Free plan

mariekirsch:
### What does it do?

This PR updates the Cloud plugin, to better reflect the current functionalities. More specifically:
- Removed mentions of the trials
- Added mentions of the Free plan
- Updated deployment methods to include the CLI
- Added translations for Spanish and Italian

### Why is it needed?

The plugin was outdated and did not reflect new features released on Strapi Cloud


packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+              </Flex>
+            </Box>
+          </Box>
+        </Tabs.Content>
+        <Tabs.Content value={'npm'}>
+          <Box background={'neutral100'}>
+            <Box padding={4}>
+              <Flex direction={'row'} justifyContent={'space-between'} alignItems={'center'}>
+                <Typography tag={'code'} textColor={'neutral800'}>
+                  npm run deploy
```
markkaylor: I think you forgot some translations here

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { formatMessage } = useIntl();
+
+  const documentationLink = formatMessage({
+    id: getTrad('Homepage.deploy.documentation'),
+    defaultMessage: 'Having trouble? Check our documentation',
+  });
+
+  return (
+    <Box paddingBottom={5}>
+      <Flex direction={'column'}>
```
markkaylor: We try to use strings directly where possible but it's not a big deal you can leave it as is

```suggestion
      <Flex direction="column">
```

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
mariekirsch: For the CLI commands? These are universal no?

markkaylor: 🤦‍♂️ yep! my bad 😆

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
mariekirsch: Ok I can fix! Anywhere else aside from the column/row?

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
```diff
@@ -0,0 +1,60 @@
+import { Box, Typography, Tabs, IconButton, Flex, Tooltip } from '@strapi/design-system';
+import { Duplicate } from '@strapi/icons';
+
+const CopyCommandButton = ({ command }: { command: string }) => {
+  const handleCopy = async () => {
+    navigator.clipboard.writeText(command);
+  };
```
markkaylor: There is a hook: for copy `import { useClipboard } from '@strapi/admin/strapi-admin';`. That seems to be the way we are interacting with the clipboard everywhere else in the codebase. Maybe we should do the same here? There is an example in `packages/core/upload/admin/src/components/CopyLinkButton/CopyLinkButton.tsx`

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </IconButton>
+    </Tooltip>
+  );
+};
+
+const CLIDeployTabs = () => {
+  return (
+    <Box minWidth={'28em'} maxWidth={'28em'} paddingTop={8}>
+      <Tabs.Root defaultValue={'yarn'}>
+        <Tabs.List aria-label={'Package manager'}>
```
markkaylor: aria-labels should be translated

mariekirsch: Ok! Do you see other components where it would be good to add them btw?

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
markkaylor: anywhere you are passing a string as a value to a prop. I think it's a lot of places 😬. It's really a style preference thing so up to you.

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Duplicate } from '@strapi/icons';
+
+const CopyCommandButton = ({ command }: { command: string }) => {
+  const handleCopy = async () => {
+    navigator.clipboard.writeText(command);
+  };
+
+  return (
+    <Tooltip label={'Copy'}>
+      <IconButton size={'XS'} variant={'ghost'} label={'Copy'} onClick={handleCopy}>
```
markkaylor: The label is what appears in the tooltip right? I think it should be translated

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
markkaylor: Don't think so, elsewhere accessible name should be inferred 👍

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
mariekirsch: I initially did that but when I included the @strapin/admin as a dependency it broke the build. Any suggestions?

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
mariekirsch: Updated

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
mariekirsch: Updated

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <Tabs.List
+            aria-label={formatMessage({
+              id: getTrad('Homepage.deploy.ariaLabel'),
+              defaultMessage: 'Deployment options',
+            })}
+          >
+            <Tabs.Trigger value="cloud">
+              <Box minWidth={{ initial: '10em', medium: '20em' }}>
+                <Flex direction="row" gap={2} justifyContent="center">
+                  <Typography variant="omega">Cloud</Typography>
```
markkaylor, line 53: translation?

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
markkaylor: Weird I'll take a quick when I test it

packages/plugins/cloud/admin/src/components/CloudDeploy.tsx
mariekirsch: Nuage? 😅 I wouldn't translate Cloud

packages/plugins/cloud/admin/src/components/CLIDeployTabs.tsx
mariekirsch: Updated

PR #23938, opened 2025-07-10
feat: key statistics homepage widget

Adzouz:
### What does it do?

Screenshot 2025-07-10 at 17 29 19

New widget for the homepage for the super admin users. They can see key statistics about various items created in the project.

### Why is it needed?

Helping improving the experience of the user on the platform.

### How to test it?

- Go to the admin homepage as a Super Admin
-> You should see the new widget

_Create / Delete some items across the CMS and come back to the homepage to see the numbers change._

🚀

### DIFF
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { darkTheme, lightTheme } from '@strapi/design-system';
-import { User } from '@strapi/icons';
+import { User, ChartBubble } from '@strapi/icons';
 import invariant from 'invariant';
 import isFunction from 'lodash/isFunction';
 import merge from 'lodash/merge';
@@ -338,6 +338,19 @@ class StrapiApp {
           },
           href: '/me',
         },
+      },
+      {
+        icon: ChartBubble,
+        title: {
+          id: 'widget.key-statistics.title',
+          defaultMessage: 'Project statistics',
+        },
+        component: async () => {
+          const { KeyStatisticsWidget } = await import('./components/Widgets');
+          return KeyStatisticsWidget;
+        },
+        pluginId: 'admin',
+        id: 'key-statistics',
       },
     ]);
 
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -351,6 +351,7 @@ class StrapiApp {
         },
         pluginId: 'admin',
         id: 'key-statistics',
+        superAdminOnly: true,
       },
     ]);
 
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import { darkTheme, lightTheme } from '@strapi/design-system';
-import { User, ChartBubble } from '@strapi/icons';
+import { User, TrendUp } from '@strapi/icons';
 import invariant from 'invariant';
 import isFunction from 'lodash/isFunction';
 import merge from 'lodash/merge';
@@ -340,7 +340,7 @@ class StrapiApp {
         },
       },
       {
-        icon: ChartBubble,
+        icon: TrendUp,
         title: {
           id: 'widget.key-statistics.title',
           defaultMessage: 'Project statistics',
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -351,7 +351,7 @@ class StrapiApp {
         },
         pluginId: 'admin',
         id: 'key-statistics',
-        superAdminOnly: true,
+        roles: ['strapi-super-admin'],
       },
     ]);
 
--- a/packages/core/admin/admin/src/components/Widgets.tsx
+++ b/packages/core/admin/admin/src/components/Widgets.tsx
@@ -71,8 +71,15 @@ const GridCell = styled(Box)`
   }
 `;
 
+const formatNumber = ({ locale, number }: { locale: string; number: number }) => {
+  return new Intl.NumberFormat(locale, {
+    notation: 'compact',
+    maximumFractionDigits: 1,
+  }).format(number);
+};
+
 const KeyStatisticsWidget = () => {
-  const { formatMessage } = useIntl();
+  const { formatMessage, locale } = useIntl();
   const { data: countDocuments, isLoading: isLoadingCountDocuments } = useGetCountDocumentsQuery();
   const { data: countKeyStatistics, isLoading: isLoadingKeyStatistics } =
     useGetKeyStatisticsQuery();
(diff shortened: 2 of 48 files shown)


packages/core/admin/admin/src/StrapiApp.tsx
```diff
@@ -339,6 +339,20 @@ class StrapiApp {
           href: '/me',
         },
       },
+      {
```
soupette, line 342: **question**: since these widgets are always displayed (unless I'm wrong) why not hardcoding them in the page that displays the widget instead? Doing so will bring 2 improvements at least from what I'm seeing:
- bundle size: the bundle will be improved as it will be easier for vite to create chunks even though we're creating a new chunk here it's not the react way to split chunks
- performance improvement: in strapi we have several stores and we do load a lot of JSX in our memory, since these widgets are always displayed I would rather hardcode them than keeping them in memory..

packages/core/admin/admin/src/components/tests/Widgets.test.tsx
```diff
@@ -29,3 +35,136 @@ describe('Homepage Widget Profile component', () => {
     expect(await screen.findByText('Editor')).toBeInTheDocument();
   });
 });
+
+describe('Homepage Widget Key Statistics component', () => {
```
soupette, line 39: **question**: have you thought about setting up storybook and do story testing instead of unit test? it could drastically improve your DX...

packages/core/upload/admin/src/hooks/tests/useRemoveAsset.test.tsx
```diff
@@ -39,17 +55,24 @@ const client = new QueryClient({
...
-      <DesignSystemProvider>
-        <NotificationsProvider>
-          <IntlProvider locale="en" messages={{}}>
-            {children}
-          </IntlProvider>
-        </NotificationsProvider>
-      </DesignSystemProvider>
-    </QueryClientProvider>
+    <Provider store={store}>
+      <QueryClientProvider client={client}>
```
soupette, line 66: **question**: how do you choose when to use RTK query vs tanstack query?

Adzouz: Good question, I saw that upload plugin is using mostly Tanstack Query for internal caching related to the assets but here the purpose of adding this Provider is to invalidate the cache of the adminApi from the admin plugin that uses RTK query.

packages/core/admin/admin/src/components/tests/Widgets.test.tsx
Adzouz: That could be some kind of improvement at some point, but I don't think we already do that for the moment.

packages/core/admin/admin/src/StrapiApp.tsx
Adzouz: If I understand correctly, you want to directly inject Widgets.tsx inside the HomePage.tsx without using the widgets.register method? I get your point but first, the purpose of this API was to be able to inject widgets from other plugins into the homepage so we could loose some consistency and second, there's another side work I started doing that will give the ability to reorder the widgets (at first only in the code but in the future, users would be able to configure their homepage view themselves). So unfortunately, we would loose this flexibility...

packages/core/upload/admin/src/hooks/tests/useRemoveAsset.test.tsx
markkaylor: > question: how do you choose when to use RTK query vs tanstack query?

We should be using RTK query everywhere and not have tanstack query as a dependency. 

The upload plugin was never migrated but we will completely rework the plugin soon. I thought it was the only package in the codebase that hadn't been migrated to RTK query but apparently the CTB is not using it either.

packages/core/admin/admin/src/StrapiApp.tsx
soupette: I'm not sure we would loose it as you could only move the default widgets to the homepage and the ones from the required plugins. The others ones can still consume the api.

Adzouz: We discussed with the other members of the team and we agreed to keep it as it is for the moment and we will see later to refactor if needed (whenever we'll have to implement the user view configuration).

packages/core/admin/admin/src/components/Widgets.tsx
```diff
@@ -39,4 +44,184 @@ const ProfileWidget = () => {
...
+
+  if (!countKeyStatistics || !countDocuments) {
+    return <Widget.Error />;
+  }
+
+  const mapping: {
+    [key: string]: {
+      label: string;
+      icon: {
+        file: React.ReactNode;
```
markkaylor: why is this called file?

packages/core/admin/admin/src/components/Widgets.tsx
```diff
@@ -39,4 +44,184 @@ const ProfileWidget = () => {
...
+        background: string;
+        color: string;
+      };
+    };
+  } = {
+    entries: {
+      label: formatMessage({ id: 'widget.key-statistics.list.entries', defaultMessage: 'Entries' }),
```
markkaylor: Instead of applying formatMessage here you could instead provide the translation object and only use formatMessage once for label.  Also not too crazy about the vague variable name "mapping"

packages/core/admin/admin/src/components/Widgets.tsx
```diff
@@ -39,4 +44,184 @@ const ProfileWidget = () => {
...
+  if (isLoadingKeyStatistics || isLoadingCountDocuments) {
+    return <Widget.Loading />;
+  }
+
+  if (!countKeyStatistics || !countDocuments) {
+    return <Widget.Error />;
+  }
+
+  const mapping: {
+    [key: string]: {
```
markkaylor: Wouldn't it be less work to not apply the type here but typecast the key in the map, ie   `const mapKey = key as keyof typeof mapping;`

packages/core/admin/admin/src/core/apis/Widgets.ts
```diff
@@ -25,6 +25,7 @@ type WidgetArgs = {
   pluginId?: string;
   id: string;
   permissions?: Array<Pick<Permission, 'action'> & Partial<Omit<Permission, 'action'>>>;
+  superAdminOnly?: boolean;
```
markkaylor: Just wondering if it would not be more flexible to have an array of roles instead of a boolean for a single role.

packages/core/content-type-builder/admin/src/components/DataManager/DataManagerProvider.tsx
```diff
@@ -183,8 +183,7 @@ const DataManagerProvider = ({ children }: DataManagerProviderProps) => {
       // Make sure the server has restarted
       await serverRestartWatcher();
       // Invalidate the guided tour meta query cache
```
markkaylor: Maybe update or remove the comment

packages/core/admin/admin/src/components/Widgets.tsx
Adzouz: I didn't know how to call it 😅 Do you have a suggestion?

markkaylor: component?

packages/core/admin/admin/src/components/Widgets.tsx
Adzouz: Like
```
label: {
  key: 'widget.key-statistics.list.entries',
  default: 'Entries'
}
```
and call formatMessage in the template?
I agree for the naming, I wasn't very inspired 😅 Do you also have a suggestion? Something like "keyStasticticsList" maybe?

markkaylor: yep! yeah I think something that mentions `keyStatistics` is good 👍

packages/core/admin/admin/src/components/Widgets.tsx
Adzouz: It actually gave me a better idea. Instead of mapping on the values sent back from the query, I'll loop over the key statistics mapping so we can get rid of this typing here and it's cleaner that way.

markkaylor:
Could you add api tests for this one as well?

markkaylor (approved):
LGTM tested locally

Adzouz:
I added the tests and fixed some failing e2e tests also @markkaylor!

packages/core/admin/admin/src/components/Widgets.tsx
```diff
@@ -39,4 +44,190 @@ const ProfileWidget = () => {
...
+                  color={item.icon.color}
+                >
+                  {item.icon.component}
+                </Flex>
+                <Flex direction="column" alignItems="flex-start">
+                  <Typography variant="pi" fontWeight="bold" textColor="neutral500">
+                    {formatMessage({
+                      id: item.label.key,
+                      defaultMessage: item.label.defaultMessage,
+                    })}
```
markkaylor: I guess you could have done `formatMessage(item.label)` directly but this is fine too

markkaylor (approved):

PR #23941, opened 2025-07-11
future: only show guided tour overview to first super admin

markkaylor:
### What does it do?

Esnures only first super admin can see the guided tour overview

### Why is it needed?

It's a product requirement

### How to test it?

Turn the future flag on
Ensure local storage is clear of STRAPI_GUIDED_TOUR in case you've closed the overview before
Login as the first super admin
Go to the homepage
You should see the guided tour
Login as another user regardless of role
You should not see the guided tour


simotae14 (commented):
The code looks good, I still need to test it

simotae14 (approved):
it works, well done

PR #23943, opened 2025-07-11
fix: add Introduction step in the Single collection Edit view

simotae14:
### What does it do?

Fixes the absence of the Introduction step in the content manager guided tour when you create a new single collection document

### Why is it needed?

In the case you are starting with a single collection the lack of this step can cause confusion

### How to test it?

Use this experimental to create a new strapi project and follow the guided tour steps, create a single collection instead of a collection and then try to add some content to this new collection

### Related issue(s)/PR(s)

fixes https://github.com/strapi/content-squad/issues/35

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
@@ -10,7 +10,7 @@ import {
   useQueryParams,
   unstable_tours,
 } from '@strapi/admin/strapi-admin';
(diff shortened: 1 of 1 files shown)


aymay-170:
Hi, thank you for the PR!

I’m a Strapi user, and in our current development we ran into the exact issue this PR is fixing (plugin name transformation). It would be really helpful if this could be merged or handled in an upcoming release.

May I ask if this PR is planned to be merged, or if this issue will be fixed officially later?

Thanks again for the work!

PR #23944, opened 2025-07-11
fix(content-manager): deduplicate documentIds in bulkDelete

jhi721:
### What does it do?

Deduplicates document IDs before performing bulk deletion in bulkDelete. This ensures only unique documentIds are passed to the delete operation.

### Why is it needed?

When draft & publish is enabled, findLocales can return both draft and published versions of the same document. This results in duplicate documentIds being sent for deletion, which causes the documents middleware to trigger the delete event multiple times for the same data.

### How to test it?

1. Enable draft & publish on a collection type.
2. Create a document and publish it (so both draft and published versions exist).
3. Delete the document using bulk delete.
4. Before the fix, the delete event is logged twice with the same payload when using strapi.documents.use.
5. After the fix, the delete event is logged only once per document.

### Related issue(s)/PR(s)


mathildeleg:
Thanks for your PR! 
Unfortunately doing this means we don't delete the document in the other locales it exists in, which is the expected behaviour of the current bulkDelete implementation. I'll close this PR for now but feel free to raise an issue if you believe we should re-think how the bulkDelete is thought of :)

PR #23954, opened 2025-07-14
chore(cron): replace node-schedule with strapi.cron service

giu1io:
### What does it do?

Replaced `node-schedule` with `strapi.cron` in the packages that were not updated.

### Why is it needed?

This is technical debt. All packages should use the new `strapi.cron` service and not be awere of the underlining implementation of the cron service.

### How to test it?

No new functionality is added in this PR. Regression testing of the following features will be required:
- Audit Logs: logs older than the set retation period should be deleted every hour.
- History: content history older thant the set reation days should be deleted every hour.
- Release Schedule:
  - A release scheduled should be published at the correct date/time
  - A release cancelled should not be published
- Telemetry ping: a telemetry ping should be sent every 12 hours.

### Related issue(s)/PR(s)

Internal Jira issue CLOUD-1520


packages/core/content-releases/server/src/services/scheduling.ts
```diff
@@ -19,29 +18,36 @@ const createSchedulingService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+          }
+        },
+        options: scheduleDate,
+      };
 
-        this.cancel(releaseId);
-      });
+      const tasks: { [key: string]: typeof task } = {};
+      const taskName = `publishRelease_${releaseId}`;
+      tasks[taskName] = task;
```
markkaylor: Just a style thing but I feel like this could be a bit more concise. Something like:

```suggestion
 const taskName = `publishRelease_${releaseId}`;
  const tasks = {
    [taskName]: {
      async task() {
        try {
          await getService('release', { strapi }).publish(releaseId);
          // @TODO: Trigger webhook with success message
        } catch (error) {
          // @TODO: Trigger webhook with error message
        }
      },
      options: scheduleDate,
    },
  };
```

markkaylor (commented):
Looks good to me, I'll test it this afternoon

packages/core/content-releases/server/src/services/scheduling.ts
```diff
@@ -19,29 +18,36 @@ const createSchedulingService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+
+      strapi.cron.add(tasks);
```
remidej: What happens when the user [disables](https://docs.strapi.io/cms/configurations/cron#enabling-cron-jobs) the cron service?

(i think this is covered but i can't remember how)

giu1io: That configuration only affects the cron task added by the user. The service is still running and executing any tasks registered by the core or any plugins.

packages/core/content-releases/server/src/services/scheduling.ts
giu1io: Nice, I pushed a commit where I made it more concise. I also removed the unused dependency from the package.json of the packages.

markkaylor: Cool, testing it now!

markkaylor (approved):
✨ Nice work! Tested and works as expected.

remidej (approved):

MarionLemaire:
Tested using the experimental : 

- [x] Scheduled release
- [x] Cancelled release
- [x] History
- [x] Audit logs

LGTM -> GO :)

hanpaine:
Hey @giu1io just a reminder to merge this if it's ready. Release branch is created this afternoon 👀

MarionLemaire:
@hanpaine @giu1io I'll need more time to test this PR (I need to find the right setup), let's move it to next week's release :)

hanpaine:
Weekly reminder that we will lock the release branch at end of day @MarionLemaire @giu1io :D

PR #23956, opened 2025-07-15
fix: getstarted admin constantly reloading

remidej:
### What does it do?

Re-applies #23929. See that PR for the details.

But also only applies the changes for our internal getstarted application by checking the app UUID

### Why is it needed?

- to fix the reload bug
- to not break production lol

### How to test it?

Internal side: see the instructions on #23929

External side: create a new app with exp. `0.0.0-experimental.072258cbcd42181dea057515b972de0e71a24ebc`

The admin should work

### Related issue(s)/PR(s)

second try after #23929


packages/core/strapi/src/node/vite/config.ts
```diff
@@ -10,6 +10,7 @@ import { buildFilesPlugin } from './plugins';
 
 const resolveBaseConfig = async (ctx: BuildContext): Promise<InlineConfig> => {
   const target = browserslistToEsbuild(ctx.target);
+  const isMonorepoExampleApp = (ctx.strapi as any).internal_config?.uuid === 'getstarted';
```
markkaylor, line 13: Nice idea!

markkaylor (approved):
👏 tested! 🚀

PR #23959, opened 2025-07-16
Resolved error on empty placeholder by replacing saved value

### LINKED ISSUE #22966: Placeholder cannot be made empty after having previously contained a value
### Node Version

20.11

### NPM/Yarn/PNPM Version

10.3

### Strapi Version

5.8.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

In the View mode of a model, when i try to remove a placeholder on a field i get a "Warning: An error occurred.".

This will only happen if there is a placeholder value for that particular field when removing it.

### Steps to Reproduce

1. Create a content type
2. Add a text field
3. Go to view mode, set a placeholder
4. Save
5. Then go back to view mode, remove the placeholder
6. Save
7. Error!

### Expected Behavior

No errors

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

akash-dabhi-qed:
from null to empty string

### What does it do?

When passing an empty placeholder value, the placeholder got null assigned which threw the 400 error, replacing placeholder value with empty string resolves the issue.

https://github.com/user-attachments/assets/f64b2445-8343-4010-97a7-2d722d5f24a0

### Why is it needed?

While trying to clear out the once populated placeholder field it throws an error. On investigating further it shows that it does not accept the null value and so throws the 400 bad request error.

### How to test it?

### Related issue(s)/PR(s)

Fixes #22966

### DIFF
--- a/packages/core/content-manager/admin/src/components/ConfigurationForm/EditFieldForm.tsx
+++ b/packages/core/content-manager/admin/src/components/ConfigurationForm/EditFieldForm.tsx
@@ -111,7 +111,11 @@ const EditFieldForm = ({ attribute, name, onClose }: EditFieldFormProps) => {
         initialValues={value}
         validationSchema={FIELD_SCHEMA}
         onSubmit={(data) => {
-          onChange(name, data);
+          const submitData = { ...data };
+          if (submitData.placeholder === null || submitData.placeholder === undefined) {
+            submitData.placeholder = '';
+          }
+          onChange(name, submitData);
           onClose();
         }}
       >


Adzouz:
Hi @akash-dabhi-qed,
Thanks a lot again for the contribution, you're very efficient 🙌 Unfortunately, I'll have to close this PR because the issue was already resolved.
Keep doing the good work and have a great day ☀️

PR #23961, opened 2025-07-16
enhancement: promote preview configuration docs

remidej:
### What does it do?

Displays a "set up preview" button when preview hasn't been configured.
screenshot 2025-07-16 at 15 37 37@2x

### Why is it needed?

The goal is to make the feature more discoverable, because there's otherwise no way for users to find out about it within the product

### How to test it?

edit your config/admin.js preview configuration multiple times, and see how that impacts the edit view:
- when preview.enabled is false, there should be no preview ui
- when preview is not explicitly disabled but preview.config.handler isn't provided either, you should get the new "set up preview" button. for example:
  ```js
  preview: {
      config: {},
    },
  ``` 
- when the preview.config.handler function returns null, there should be no preview ui

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/PreviewSidePanel.tsx
@@ -31,6 +31,7 @@ const PreviewSidePanel: PanelComponent = ({ model, documentId, document }) => {
   const { pathname } = useLocation();
   const [{ query }] = useQueryParams();
   const isModified = useForm('PreviewSidePanel', (state) => state.modified);
+  const isUnsaved = Boolean(!document || !document.id);
 
   const title = formatMessage({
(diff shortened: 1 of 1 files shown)


Adzouz:
LGTM, the only thing when I was testing is that I can't make it disappear even if I force the PREVIEW_ENABLED option to false.

Adzouz:
LGTM, the only thing when I was testing is that I can't make it disappear even if I force the PREVIEW_ENABLED option to false.

Adzouz:
> LGTM, the only thing when I was testing is that I can't make it disappear even if I force the PREVIEW_ENABLED option to false.

So after investigating with @remidej, we figured that it was in the specific case where a single type was just created without any content (Draft, not saved nor published). In the logic, just don't load the component whenever the configuration is at false.

Adzouz (approved):
Perfect, it's working now! Thanks 🚀

PR #23965, opened 2025-07-16
future: always run the tour regardless of app state

markkaylor:
### What does it do?

- Runs the tours no matter what the app state is

### Why is it needed?

Simplify

### How to test it?

0.0.0-experimental.7281e9ed655cf0093344801f3db6c50a4f6ec20f

- Test with this option: Start with an example structure & data? y
When the overview loads all tours should be marked as not done yet
When you go to the CTB it should propose the tour even if there is already a schema
When you go to the CM it should propose the tour even if there is already content
When you go to the API tokens it should propose the tour even if a new token has been created

Otherwise it should function the same as before. Test an empty app too to ensure this is the case. 

### Related issue(s)/PR(s)


simotae14 (approved):
Tested and it works perfectly

PR #23966, opened 2025-07-16
future: fix Final Step Popover has the LeftMenu link tooltip on top of it

simotae14:
### What does it do?

Fix this undesired behaviour 
tooltip

when we show the Final step of a Guided tour

### Why is it needed?

I want to solve this ui issue

### How to test it?

Try to run the new guided tour and reach the final step of every tour and check if the Popover near the LeftMenu Link is visible without the NavLink tooltip on top of it

### Related issue(s)/PR(s)

fixes https://github.com/strapi/content-squad/issues/39

### DIFF
--- a/packages/core/admin/admin/src/components/LeftMenu.tsx
+++ b/packages/core/admin/admin/src/components/LeftMenu.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { Divider, Flex, FlexComponent, useCollator } from '@strapi/design-system';
 import { Lightning } from '@strapi/icons';
 import { useIntl } from 'react-intl';
-import { useLocation } from 'react-router-dom';
+import { type To, useLocation } from 'react-router-dom';
 import { styled } from 'styled-components';
 
 import { useAuth } from '../features/Auth';
@@ -16,6 +16,7 @@ import { NavBrand } from './MainNav/NavBrand';
 import { NavLink } from './MainNav/NavLink';
 import { NavUser } from './MainNav/NavUser';
 import { TrialCountdown } from './MainNav/TrialCountdown';
+import { tours as unstable_tours } from './UnstableGuidedTour/Tours';
 
 const sortLinks = (links: MenuItem[]) => {
   return links.sort((a, b) => {
@@ -46,6 +47,21 @@ const NavListWrapper = styled<FlexComponent<'ul'>>(Flex)`
 `;
 
 interface LeftMenuProps extends Pick<Menu, 'generalSectionLinks' | 'pluginsSectionLinks'> {}
+
+const getGuidedTourTooltip = (to: To) => {
+  const normalizedTo = to.toString().replace(/\//g, '');
+
+  switch (normalizedTo) {
+    case 'content-manager':
+      return unstable_tours.contentTypeBuilder.Finish;
+    case '':
+      return unstable_tours.apiTokens.Finish;
+    case 'settings':
+      return unstable_tours.contentManager.Finish;
+    default:
+      return React.Fragment;
+  }
+};
 
 const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) => {
   const user = useAuth('AuthenticatedApp', (state) => state.user);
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -88,38 +104,41 @@ const LeftMenu = ({ generalSectionLinks, pluginsSectionLinks }: LeftMenuProps) =
                   : undefined;
 
               const labelValue = formatMessage(link.intlLabel);
+              const GuidedTourTooltip = getGuidedTourTooltip(link.to);
```
remidej: I believe this creates a brand new component on every render, which should be avoided. See the react docs on the topic: https://react.dev/learn/your-first-component#nesting-and-organizing-components

markkaylor: But we aren't defining the function/component inside another component as illustrated as a bad example in the docs. It's defined outside as the docs suggest. Am I missing something? Do you have a suggestion to improve it?

remidej: it looks different, but think these are effectively the same:

```jsx
const Component = getComponent()
// or
const Component = () => <div>component</p>
```

because the identify of the component function will change on every render

There's a way to check though, log React.useId() in your component, it will remain the same on every render if the component identity is the same

markkaylor: Yeah if I log React.useId in `getGuidedTourTooltip` the id is different on every render, it should probably be a component anyway now that I think about it but the result is the same with React.useId when I do this 🤔  

```
const GuidedTourTooltip = ({ to, children }: { to: To; children: React.ReactNode }) => {
  const normalizedTo = to.toString().replace(/\//g, '');
  console.log(React.useId());

  if (normalizedTo === 'content-manager') {
    return (
      <unstable_tours.contentTypeBuilder.Finish>
        {children}
      </unstable_tours.contentTypeBuilder.Finish>
    );
  }

  if (normalizedTo === 'settings') {
    return <unstable_tours.contentManager.Finish>{children}</unstable_tours.contentManager.Finish>;
  }

  if (normalizedTo === '') {
    return <unstable_tours.apiTokens.Finish>{children}</unstable_tours.apiTokens.Finish>;
  }

  return children;
};
```

remidej: >so does this

can you show how you're using it? normally the key tag inside the loop should be what helps react persist the instances

markkaylor: Screenshot 2025-07-17 at 11 06 54

markkaylor: Ok so I should move it to add the key!

markkaylor: Same result, but it does repeat the same 6 ids, not sure I understand that either

Screenshot 2025-07-17 at 11 08 29

remidej: >Ok so I should move it to add the key!

not sure you need to move it! i just meant that this particular problem is why keys were necessary. as long as there's one in a parent you should be good

>Same result, but it does repeat the same 6 ids, not sure I understand that either

isn't that what you want? I guess there are 6 items in the loop. This means each loop item's tooltip has a stable identity

markkaylor: >not sure you need to move it! i just meant that this particular problem is why keys were necessary. as long as there's one in a parent you should be good

ok yeah I was about to ask that actually! I thought it might be the case. 

> isn't that what you want? I guess there are 6 items in the loop. This means each loop item's tooltip has a stable identity

Oh yeah I guess that does make sense, but the function before has the same result it's just I'm a dumb dumb and forgot we are in a loop so I _thought_ the id was different on every render. The component is still the better pattern though.

simotae14: you are right Rémi, I refactored a little bit the code to externalise the GuidedTourTooltip definition to avoid rirendering the component on every LeftMenu rirender

packages/core/admin/admin/src/components/LeftMenu.tsx
```diff
@@ -47,6 +48,27 @@ const NavListWrapper = styled<FlexComponent<'ul'>>(Flex)`
...
+        </unstable_tours.contentTypeBuilder.Finish>
+      );
+    case '':
+      return <unstable_tours.apiTokens.Finish>{children}</unstable_tours.apiTokens.Finish>;
+    case 'settings':
+      return (
+        <unstable_tours.contentManager.Finish>{children}</unstable_tours.contentManager.Finish>
+      );
+    default:
+      return <>{children}</>;
```
markkaylor: Does it have to be in wrapped a fragment?

markkaylor:
@simotae14 I think we were using a getter function somewhere else too in packages/core/admin/admin/src/components/SubNav.tsx, once we agree on the pattern here could you update over there. 

Also this just reminded me, I'm a bit concerned in the SubNav switch we are using a translated value. I checked that it still works when the app is in different languages, and it does as far as I can tell, but if you have a second to see if there is a more reliable solution that would be great. Otherwise I'll create a task to work on it.

markkaylor (approved):
Tested it. Nice fix! So glad to be rid of that other tooltip!

remidej (approved):
I haven't tested but LGTM

PR #23975, opened 2025-07-17
future: add step count to all got it actions

markkaylor:
### What does it do?

Adds the step count to all got it actions

### Why is it needed?

They are supposed to be there

### How to test it?

Go through all tours. You should always see the step count on a "Got it" step

### Related issue(s)/PR(s)


simotae14 (approved):
nice catch

PR #23978, opened 2025-07-17
fix: strapi package version numbers to correct version

innerdvations:
### What does it do?

Fixes mismatched version numbers 

### Why is it needed?

Aside from the obvious, it's causing test private-params.test.api.js to fail in the v4 branch

### How to test it?

- everything should work fine
- tests should now all pass
- experimentals now build again

### Related issue(s)/PR(s)

DX-2049

Should unblock all the v4 PRs that have failing CI (see [milestone](https://github.com/strapi/strapi/milestone/342))


innerdvations:
> any clue why the versions were behind?

I can't say for sure, but the git history shows we went from `.20` straight to `.22` so there may have been a missed release version merge. But that doesn't exactly explain why only a handful of them got missed.

PR #23982, opened 2025-07-18
Ci: release improvements

Bassel17:
### What does it do?

- Moves the release github changelog step to after the npm publishing step
- Moves the clean and build commands to the `releases.js` script

### Why is it needed?

- Just makes more sense for the changelog to be released after npm publish when the packages are actually available
- Adding the the clean and build to the original script makes it more usable and shorter as a release command if used locally

### How to test it?

- release an experimental
- release --dry-runs locally

### things still to be done in this PR:

- [x]  add beta, alpha, rc releases to the github actions

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -9,6 +9,17 @@ on:
         options:
           - patch
           - minor
+          - prepatch
+          - preminor
+      distTag:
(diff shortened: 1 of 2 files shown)


remidej (approved):
changes LGTM and are working. thanks for the details

PR #23984, opened 2025-07-18
future: add reset guided tour to profile

markkaylor:
### What does it do?

- Add the "Reset guided tour" button to the profile page
- Disable all start buttons on the homepage except "Create your schema" if a schema doesn't already exist

### Why is it needed?

So users can redo the tour if they want to

### How to test it?

**Reset**
Do the tours or skip them
Go to the profile page
Click reset guided tour at the bottom
Go to the homepage and you should see all tours are available to start from the beginning

**Disabled state**
When you have an empty project and haven't created a schema yet
Go to the homepage, you should only the "Create your schema" start button enabled
Click it and do the CTB tour
Go back to the homepage and you should see the rest of the tours buttons enabled


packages/core/admin/admin/src/pages/ProfilePage.tsx
```diff
@@ -423,70 +427,106 @@ const UserInfoSection = () => {
...
+  const { toggleNotification } = useNotification();
+  const dispatch = unstableUseGuidedTour('ProfilePage', (s) => s.dispatch);
+
+  const onClickReset = () => {
+    dispatch({ type: 'reset_all_tours' });
+    toggleNotification({
+      type: 'success',
+      message: formatMessage({
+        id: 'notification.success.reset',
+        defaultMessage: 'Guided tour reset',
```
markkaylor: Missing translation

PR #23985, opened 2025-07-18
feat: add preview device selector

remidej:
### What does it do?

- adds a preview device selector ([figma](https://www.figma.com/design/BDCSSHnOZutJ4aaYTs41Bg/Preview?node-id=2017-8393&t=daFNmxPBBIHrKNZZ-0))
- fixes an issue where the no data view was shown for a split second when opening the preview page

### How to test it?

open preview page, toggle between mobile and desktop views


packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -77,6 +85,31 @@ const PreviewPage = () => {
 
   const params = React.useMemo(() => buildValidParams(query), [query]);
 
+  const devices = [
+    {
+      name: 'desktop',
+      label: formatMessage({
+        id: 'content-manager.preview.device.desktop',
+        defaultMessage: 'Desktop',
+      }),
```
markkaylor: Just a thought: you could store the translation object here and only use formatMessage in SelectOption below. Which would mean you could move the devices array outside the component to not pollute that space with a constant. Speaking of which you could consider the `DEVICES` constant casing convention. What you have is just fine too.

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -110,7 +143,7 @@ const PreviewPage = () => {
 
   const isLoading =
     previewUrlResponse.isLoading || documentLayoutResponse.isLoading || documentResponse.isLoading;
-  if (isLoading && !documentResponse.document?.documentId) {
```
markkaylor, line 113: You sure about removing this? It's weird I thought this was there for fixing the no data thing 🤔 , but apparently it's causing it 😆

markkaylor (dismissed):

markkaylor:
p.s. tested

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: good call! i initially want to put that array outside the component 👍

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: it definitely causes the no data flash issue. i'm trying to figure out why it was here in the first place but i can't make sense of it 🤷

markkaylor: Maybe it was that it was removed and then re-added in a bad merge or something 🤷‍♂️. I remember something about this, but anyway doesn't matter fixed now, I tested it. Thanks for taking a second look.

markkaylor (approved):
Nice, tested 🚀

YoannDelpierre:
Amazing! I was bootstrapping a live preview for my company, and this was the missing piece—just in time!

Thanks

PR #23998, opened 2025-07-21
feat: homepage upcoming releases widget

Adzouz:
### What does it do?

Screenshot 2025-07-21 at 17 26 52

Next homepage widget: Upcoming releases.
When creating releases scheduled in the future using the package content-releases (Premium - Enterprise Edition feature), the connected user can see the list of them on the homepage.

### Why is it needed?

To improve visibility and user experience on the homepage.

### How to test it?

- Go to the admin homepage
  -> You should see the new widget

_Create / edit / delete some releases to see the updates in the widget._

🚀


packages/core/content-releases/admin/src/components/Widgets.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { Link, useNavigate } from 'react-router-dom';
+import { styled } from 'styled-components';
+
+import { Release } from '../../../shared/contracts/releases';
+import { getBadgeProps } from '../pages/ReleasesPage';
+import { useGetUpcomingReleasesQuery } from '../services/homepage';
+
+import { RelativeTime } from './RelativeTime';
+
+const CellTypography = styled(Typography).attrs({ maxWidth: '14.4rem', display: 'block' })`
```
markkaylor: It looks like you're styling the component 3 different ways. Maybe these could just be passed as props or written with the rest of the css?

markkaylor (dismissed):
Tested 👏 nice work!

Adzouz:
I fixed the failing test and the typescript issue @markkaylor it's good now!

markkaylor (commented):
Should we add an api test before merging?

Adzouz:
> Should we add an api test before merging?

@markkaylor We don't need to, it's using an existing endpoint already used in the audit logs settings page. It should be already tested (well I hope so).

Adzouz:
> Should we add an api test before merging?

@markkaylor yes you're right, we should this time. I'll do it!

Adzouz:
I added the test @markkaylor!

markkaylor:
> I added the test @markkaylor!

Awesome thanks! I will review and test this afternoon

markkaylor (dismissed):

PR #24001, opened 2025-07-22
fix: guided tour is null

markkaylor:
### What does it do?

Fixes an issue with the FE build

### Why is it needed?

It breaks everything

### How to test it?

Create an app with this experimental:

`npx create-strapi-app@0.0.0-experimental.85d89daad792594faa7ca75fec6a1a1913e3e2cc test-tour-fix`

It should work. You should see the admin load with no errors in the console about `tours` being null or undefined

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
+++ b/packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
@@ -51,17 +51,6 @@ const [GuidedTourProviderImpl, unstableUseGuidedTour] = createContext<{
   dispatch: React.Dispatch<Action>;
 }>('UnstableGuidedTour');
 
-const initialTourState = Object.keys(guidedTours).reduce((acc, tourName) => {
-  const tourLength = Object.keys(guidedTours[tourName as ValidTourName]).length;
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/components/UnstableGuidedTour/Context.tsx
```diff
@@ -84,7 +73,31 @@ function reducer(state: State, action: Action): State {
...
+          currentStep: 0,
+          length: 0,
+          isCompleted: false,
+        },
+        strapiCloud: {
+          currentStep: 0,
+          length: 0,
+          isCompleted: false,
+        },
+      };
```
markkaylor: I'll refactor this but I think we should just merge it for now to unblock the release QA

simotae14: ok I see why you are changing the code in this way, we can reiterate on this part after the stable release

markkaylor:
Looking in to the failing tests

simotae14 (approved):
I tried the experimental with and withour future flag and it doesn't break, well done

PR #24003, opened 2025-07-22
chore: remove old guided tour and set the new one as stable

markkaylor:
### What does it do?

- Removes the old guided tour
- Removes the future flag for the new guided tour
- Sets the new guided tour as stable

### Why is it needed?

To release the feature next week

### How to test it?

Use the experimental: _coming soon_
The tour should work as expected


simotae14 (dismissed):
The code looks good to me, thanks for the patience to search and remove all the logic related to the old Guided tour

PR #24009, opened 2025-07-23
fix uid format

maccomaccomaccomacco:
### What does it do?

It fixes uids creation from the frontend when a content type is created.

### Why is it needed?
It currently mismatches with the server uid.


derrickmehaffy:
Experimental build: https://github.com/strapi/strapi/actions/runs/16473011805

innerdvations:
I'm not so sure about this. While `singularName` does make way more sense here imo, it has been this way for over 4 years so this will change expectations users might have of what the resulting uid will be.

If there's a bug related to this (which I think is why you made this change, but I can't find the related issue) we may want to fix it in `createUid` or the validators for now and save a change like this PR for v6

Or maybe the server side is actually wrong here? I'm not entirely sure how this happened if this bit has been unchanged for over 4 years.

PR #24010, opened 2025-07-23
Ctm debounced search

Adzouz:
### What does it do?

Removes the need to press Enter to search on CTM by searching by typing with 500ms debounce.

### Why is it needed?

It'll match the behavior between the CTM menu and the search input, it's also a better UX IMO.

### How to test it?

1. Open CTM
2. Type something in the search input

https://github.com/user-attachments/assets/cd41b0bd-d8e8-4d6e-8880-e656bf7a2af2

### Related issue(s)/PR(s)

Closes https://github.com/strapi/strapi/issues/21358

### DIFF
--- a/packages/core/admin/admin/src/components/SearchInput.tsx
+++ b/packages/core/admin/admin/src/components/SearchInput.tsx
@@ -5,6 +5,7 @@ import { Search as SearchIcon } from '@strapi/icons';
 import { useIntl } from 'react-intl';
 
 import { TrackingEvent, useTracking } from '../features/Tracking';
+import { useDebounce } from '../hooks/useDebounce';
 import { useQueryParams } from '../hooks/useQueryParams';
 
 interface SearchInputProps {
@@ -29,6 +30,7 @@ const SearchInput = ({
 
   const [value, setValue] = React.useState(query?._q || '');
   const [isOpen, setIsOpen] = React.useState(!!value);
+  const debouncedSearch = useDebounce(value, 500) || '';
 
   const { formatMessage } = useIntl();
   const { trackUsage } = useTracking();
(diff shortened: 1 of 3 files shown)


packages/core/admin/admin/src/components/tests/SearchInput.test.tsx
```diff
@@ -46,13 +46,14 @@ describe('SearchInput', () => {
...
-
-    expect(searchParams.has('_q')).toBe(true);
-    expect(searchParams.get('_q')).toBe('michka');
+    await waitFor(
+      () => {
+        expect(new URLSearchParams(getByRole('listitem').textContent ?? '').get('_q')).toBe(
+          'michka'
+        );
+      },
+      { timeout: 600 }
```
markkaylor: Should we not be using jest timer mocks 🤔  https://jestjs.io/docs/timer-mocks

packages/core/admin/admin/src/components/SearchInput.tsx
```diff
@@ -46,24 +48,27 @@ const SearchInput = ({
...
-      handleToggle();
-      setQuery({ _q: '' }, 'remove');
-    }
-  };
+    },
+    [setQuery, trackUsage, trackedEvent, trackedEventDetails]
+  );
+
+  React.useEffect(() => {
+    handleSearch(debouncedSearch);
```
remidej: why not move the handleSearch implementation in here? you wouldn't need useCallback anymore

Adzouz: Because it would regenerate the method each time the hook is called no?

remidej: True but I don't think that really affects performance, at least compared to the price of memoization with useCallback.

The React team recommends moving function definitions inside the effect when possible: https://react.dev/reference/react/useEffect#removing-unnecessary-function-dependencies

remidej (approved):
Tested, very nice 👍

Adzouz:
> Tested, very nice 👍

@remidej Yes but the issue on this PR is that for some reason the tests are cancelled after 360 min and we don't know why exactly because they still pass. I even tried to comment the whole file but nothing changes. It's like there's a test (I don't which one) that is stalling or keep the thread still opened even if it succeeded 😢

Adzouz:
Closing because we couldn't find the issue with the tests being stucked.

PR #24032, opened 2025-07-25
fix(content-type-builder): fix schema validation when max is equal to…

ulybu:
… min

Validation of the update schema was rejecting when we would  pass a max value equal to the min value. The error message clearly stated the intention: "max must be greater or  EQUAL to min".

### What does it do?

Fixes a validation rule that enforce max being greater or equal than min. This PR allows a max value to be equal to the min value. 

### Why is it needed?

I would say it's the expected behavior when there is no option to choose between a "strictly" and "greater or equal" comparator. Reason is , with a "strictly" you cannot implement the behavior "value must be exactly x".

Also the error messge says `max must be greater or  equal to min`.

### How to test it?

- Create a dynamic zone
- go to `Advanced Settings`
- set  a `Minimum value` and `Maximum value` to `1`
- Hit Save, it should not reject the update 

Screenshot 2025-07-25 at 17 51 07

### DIFF
--- a/packages/core/content-type-builder/server/src/controllers/validation/__tests__/schema.test.ts
+++ b/packages/core/content-type-builder/server/src/controllers/validation/__tests__/schema.test.ts
@@ -1,0 +1,96 @@
+import { z } from 'zod';
+import { maxGreaterThanMin, maxLengthGreaterThanMinLength } from '../schema';
+
+describe('Schema', () => {
+  let ctx: z.RefinementCtx & { addIssue: jest.Mock };
+  let expectAddIssue: (shouldAddIssue: boolean) => void;
+
+  beforeEach(() => {
+    ctx = {
+      addIssue: jest.fn(),
+    } as unknown as z.RefinementCtx & { addIssue: jest.Mock };
+
+    expectAddIssue = (shouldAddIssue: boolean) => {
+      if (shouldAddIssue) {
+        expect(ctx.addIssue).toHaveBeenCalled();
+      } else {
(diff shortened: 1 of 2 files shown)


ulybu:
This bug is currently in prod. It was introduced in c9daf35 wich was released with [v5.14.0](https://github.com/strapi/strapi/releases/tag/v5.14.0)

hanpaine:
Thanks @ulybu, nice catch - just to confirm, did you check if it's been raised as an Issue already (or did you spot it and fix it in one fell swoop? 😁 )

ulybu:
I looked for an issue but could not find it. 
Also initially when i found the commit I thought it was released with 5.19 which is only a ~5 days old. So looking through recent bugs involving `max` was fairly easy. Since then i'm not so sure when was it released. I still cannot find it though. 

Screenshot 2025-07-28 at 14 24 13

innerdvations (approved):
LGTM! Just tested it briefly to confirm it worked, @butcherZ do you want to do a bit more QA and then merge?

butcherZ:
@innerdvations I just did QA locally, we can definitely merge it once all the tests are passed 😄 ! thank you again @ulybu 🍻

ulybu:
You're welcome 👋  , from [Macadam](https://www.macadam.app/es)'s team 👟

edporras:
Does this resolve the issue that prevents forcing a fixed length on a text field by setting the minimum length equal to maximum length ? Trying to do this on (v5) throws an error.

PR #24038, opened 2025-07-28
fix: prevent full screen loader regression in preview page

remidej:
### What does it do

- fixes regression from #23985 that caused a full screen loader when saving an update from the preview side editor. That prevented "live" preview to work because it reloaded the iframe every time
- another attempt at fixing the "no data" flash by adding a more specific condition for the loader

### How to test it

- navigate from the edit view to the preview page. You should not see the "no data" screen, even for a fraction of a second
- make an update in the preview side editor and fix it. There shouldn't be any full screen loader. The preview on the right should directly update to reflect the change instead

### DIFF
--- a/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
+++ b/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
@@ -148,7 +148,7 @@ const PreviewPage = () => {
 
   const isLoading =
     previewUrlResponse.isLoading || documentLayoutResponse.isLoading || documentResponse.isLoading;
-  if (isLoading) {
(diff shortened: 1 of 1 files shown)


markkaylor (approved):
Looks like it's working:
- No "no data" flash when first navigating to preview
- Updates are instant with no refresh

For another PR, I did notice an aggressive reload when switch d&p tabs for the first time, then it works smoothly. 

https://github.com/user-attachments/assets/c43dc9c7-b05f-480f-86ad-b8a0e2d69488

remidej:
@markkaylor I agree this D&P switch is annoying, we should really fix it for the edit view too

PR #24039, opened 2025-07-28
fix: support wildcards in arrays of origins

innerdvations:
### What does it do?

Adds support for wildcards in arrays of origins 

### Why is it needed?

They currently aren't supported

### How to test it?

Create a CORS origin array (either array or csv string) that includes a wildcard *
- Without this PR, the wildcard would not be respected
- With this PR, the wildcard is respected

Automated unit tests have been provided

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/middlewares/cors.ts
+++ b/packages/core/core/src/middlewares/cors.ts
@@ -38,7 +38,9 @@ export const cors: Core.MiddlewareFactory<Config> = (config) => {
 
   return koaCors({
     async origin(ctx) {
-      if (!ctx.get('Origin')) {
+      const requestOrigin = ctx.get('Origin');
+
+      if (!requestOrigin) {
         return '*';
       }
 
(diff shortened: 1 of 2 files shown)


butcherZ (approved):
QAed locally, I can confirm the fix, also somedomain:3000 requesting a server configured for somedomain:1337 is now correctly blocked as well 😄

packages/core/core/src/middlewares/cors.ts
```diff
@@ -22,6 +22,54 @@ const defaults: Config = {
...
+    originList = configuredOrigin;
+  }
+
+  // Normalize originList into an array
+  let normalizedOrigins: string[];
+  if (Array.isArray(originList)) {
+    normalizedOrigins = originList;
+  } else if (originList === undefined || originList === null) {
+    // Handle undefined/null - treat as wildcard
+    normalizedOrigins = ['*'];
```
innerdvations, line 58: Just as a comment for posterity, this seems to be standard practice for undefined origins in settings (I think koa does the same thing), and it seemed better to err on the side of not causing breaking changes.

PR #24041, opened 2025-07-28
fix: create relation and save flaky test

simotae14:
### What does it do?

Fix the crate relation and save e2e test regarding ROTF

### Why is it needed?

The test often was failing and we had to relaunch the CI every time

### How to test it?

just check if the CI in our PRs are still failing or not because of this test now splitted in 3 files


innerdvations (approved):
LGTM as long as the tests still pass! Thanks for splitting this up, we probably need the same treatment in a lot of other places too, it would speed up the overall test runtimes

Adzouz (approved):
I really hope it will help because this one was really annoying! Thanks a lot, approved! 🙏

PR #24049, opened 2025-07-29
chore: add empty project to examples

markkaylor:
### What does it do?

Add an empty project to the examples in the monorepo

### Why is it needed?

To have a brand new project for testing

### How to test it

cd examples/empty
yarn develop

You should see an empty strapi application with not content or structure

### DIFF
--- a/examples/empty/config/features.ts
+++ b/examples/empty/config/features.ts
@@ -1,5 +1,1 @@
-export default () => ({
-  future: {
-    unstableGuidedTour: true,
-  },
-});
+export default () => ({});


examples/empty/config/features.ts
```diff
@@ -0,0 +1,5 @@
+export default () => ({
+  future: {
+    unstableGuidedTour: true,
```
simotae14: why are you adding the future flag?

simotae14 (approved):
Thanks a lot for this PR, super useful to test some scenarios like the guided tour

examples/empty/config/features.ts
markkaylor: good catch, that's old

remidej (approved):

PR #24053, opened 2025-07-30
chore: improve content-type builder tour

markkaylor:
### What does it do?

- Updates the wording 
- Adds steps
- Adds local storage migration to include the new steps if the user already has a local storage entry
- Starts the remove of the getCompletedActions service.

### Why is it needed?

Improves the tour based on feedback from the blitz

### How to test it?

Run the tour.

WIP E2E test updated


packages/core/admin/server/src/services/guided-tour.ts
markkaylor, line 1: I've decided to remove all this. Since I'm updating each tour I'll remove them one by one until I can remove the entire service and the cache invalidations. 

This just doesn't make sense anymore. Originally it was to support the auto detect, then I thought it was nice because we didn't have to imperatively dispatch the updates for completed actions. I just find it kind of weird to be managing this here and on the frontend since some actions are FE only. It adds complexity and is just sort of confusing. I rather keep everything in one place on the frontend even if it does mean imperatively making the updates to state. It's less cool but easier.

Adzouz:
I know it's becoming too complicated and I'm picky but what if someone was at step 2 and we introduce a new step 1 in a tour? They will stay on step 2 and so the new step 1 will never appear to the user? Wouldn't we want to reset the current step of the user in an ongoing tour?

packages/core/admin/admin/src/components/GuidedTour/tests/reducer.test.ts
```diff
@@ -723,9 +723,9 @@ describe('GuidedTour | reducer', () => {
       const initialState = {
         tours: {
           contentTypeBuilder: {
-            currentStep: 5,
+            currentStep: 8,
             isCompleted: true,
-            length: 5,
+            length: 8,
```
Adzouz: Don't we want to make them dynamic (importing Tours, like you did in migrations.test.ts). Otherwise we would have to update it again here if the length is changing again.

markkaylor:
> I know it's becoming too complicated and I'm picky but what if someone was at step 2 and we introduce a new step 1 in a tour? They will stay on step 2 and so the new step 1 will never appear to the user? Wouldn't we want to reset the current step of the user in an ongoing tour?

That's correct. For what changed here in step 1, I didn't think it really mattered we just changed the wording. But I don't think it would be complicated to reset the current step to the first one when the tour length changes. I can try it out.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -72,6 +79,13 @@ function reducer(state: State, action: Action): State {
       draft.tours[action.payload].isCompleted = nextStep === draft.tours[action.payload].length;
     }
 
+    if (action.type === 'previous_step') {
```
Adzouz, line 95: Personal suggestion: Could be written like
```
if (action.type === 'previous_step' && draft.tours[action.payload].currentStep - 1 > 0) {
  draft.tours[action.payload].currentStep -= 1;
}
```
We could maybe also protect the `next_step` action to be sure we won't go above the length of the tour?

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
```diff
@@ -183,6 +192,14 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
                   <FormattedMessage id="tours.skip" defaultMessage="Skip" />
                 </Button>
               )}
+              {!showSkip && showPrevious && (
```
Adzouz, line 195: Question: why previous is shown only if skip isn't? Is it a space issue?

packages/core/admin/admin/src/translations/en.json
```diff
@@ -802,16 +802,22 @@
...
-  "tours.contentTypeBuilder.CollectionTypes.title": "Collection Types",
-  "tours.contentTypeBuilder.CollectionTypes.content": "A content structure that can manage multiple entries, such as articles or products.",
-  "tours.contentTypeBuilder.SingleTypes.title": "Single Types",
-  "tours.contentTypeBuilder.SingleTypes.content": "A content structure that can manage only one entry, such as a homepage or a contact page.",
-  "tours.contentTypeBuilder.Components.title": "Components",
-  "tours.contentTypeBuilder.Components.content": "A content structure that can be used in multiple collection types or single types.",
-  "tours.contentTypeBuilder.Finish.title": "It's time to create content!",
-  "tours.contentTypeBuilder.Finish.content": "Now that you created content types, you'll be able to create content in the content manager.",
+  "tours.contentTypeBuilder.Introduction.title": "Welcome to the Content-Type Builder!",
+  "tours.contentTypeBuilder.Introduction.content": "Ready to shape your content universe? Here's where you create and manage the structure of your app with collection types, single types, and reusable components. Let's dive in!",
```
Adzouz: Not really a fan of the word "universe", it's vague and weird imho. I would remove it.

packages/core/admin/admin/src/translations/en.json
```diff
@@ -802,16 +802,22 @@
...
+  "tours.contentTypeBuilder.CollectionTypes.title": "Meet Collection Types",
+  "tours.contentTypeBuilder.CollectionTypes.content": "These are your go-to for managing multiple entries — think blog posts or products. If you need many, start here.",
+  "tours.contentTypeBuilder.SingleTypes.title": "Say Hi to Single Types",
+  "tours.contentTypeBuilder.SingleTypes.content": "Perfect for one-off entries like your homepage or site settings. You only need one of these in your life.",
```
Adzouz: Same here with "You only need one of these in your life." it feels too much AI generated and too much. I would keep something simpler but fun like "They'll be unique... Like you." 🤣

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
markkaylor: I don't think so. The !showSkip is probably not necessary I'll double check.

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -27,11 +30,42 @@ const GotItAction = ({ onClick }: { onClick: () => void }) => {
...
+  const uid = routeParams?.split('/').pop();
+
+  const contentType = uid ? data?.data?.schemas?.[uid as UID.ContentType] : null;
+  const contentTypeKindDictionary = {
+    collectionType: 'collection-types',
+    singleType: 'single-types',
+  };
+  // Create the to path with fallback to content-manager
+  const to = contentType
+    ? `/content-manager/${contentTypeKindDictionary[contentType.kind as keyof typeof contentTypeKindDictionary]}/${contentType.uid}`
```
markkaylor: remove type casting

packages/core/admin/admin/src/translations/en.json
markkaylor: Maybe the simplest thing to do is just remove "in your life"

To be completely transparent the more I read this stuff the more I don't like it. I know the feedback in the blitz was to make it more "fun" but the guided tour is for the developer persona not the editor. I feel like developers just want a succinct explanation without any fluff 🤷.

packages/core/admin/admin/src/translations/en.json
markkaylor: I don't think I can say "Ready to shape your content" we need a word that represents something like schema, structure, or model. Maybe we could just remove the first part and have:

`Here you can create and manage the structure..."

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
markkaylor: I deliberately avoided that in favor of one umbrella condition for each action. But if you prefer this I can update.  Also happy to add a similar check for next_step

packages/core/admin/admin/src/translations/en.json
Adzouz: You're right, I have the same feeling. Still, you can keep it fun without being too much. Could we ask maybe the AI to be more in that way?

packages/core/admin/admin/src/translations/en.json
Adzouz: Ok with that

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: No, I wasn't sure but that fine by me, just the check for next_step then :)

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
markkaylor: It's really just a convenience since skip and previous are never displayed at the same time. If I remove it I have to add the prop to every step with skip to hide previous. I still want to control showPrevious so a ternary based on showSkip won't work. I think I'll leave it if it's cool with you, it beats having to manage a prop in several places in my opinion.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
markkaylor: I did try your suggestion but in the end I think I still prefer the readability of all the action conditions being uniform, it's a bit more verbose but more my flavor. Hope you don't mind, lmk what you think.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -2,35 +2,47 @@ import * as React from 'react';
...
+type RequiredActionValues = {
+  [K in keyof typeof GUIDED_TOUR_REQUIRED_ACTIONS]: (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K] extends Record<
+    string,
+    string
+  >
+    ? (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K][keyof (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K]]
+    : never;
+}[keyof typeof GUIDED_TOUR_REQUIRED_ACTIONS];
+
+export type CompletedActions = RequiredActionValues[];
```
markkaylor, line 31: this does what I want but I won't pretend like I wrote it.... 

co-authored by: claude 😄

markkaylor:
@Adzouz thanks for the review, I've updated but still happy to discuss some of the things I didn't change. Let me know your thoughts.

packages/core/admin/admin/src/translations/en.json
markkaylor: I tamed it down with my own brain. Hopefully it's still a bit fun and less of the AI generated vibe...

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
Adzouz: If that's the rule I'm good with it!

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: Fine by me!

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: I don't even know what's going on here 🤣

markkaylor: I can add a comment. I think it looks more complicated than it is. We are just deriving the type union of accepted strings from the constant. So it ends up being 'didCreateContentTypeSchema' | 'didAddField' | 'didCreateConte' etc...

The constant I think is nice since you only have to change something there and everything else updates.

Adzouz:
LGTM, just a thing, testing locally I have a weird behaviour that displays the message about creating an API token (the last step of CM tour) before I was able to fill any of the field and publish the entry:
https://github.com/user-attachments/assets/93f11250-a122-4050-9d44-7743a5dca030

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -177,13 +247,14 @@ const tours = {
           <Step.Actions showStepCount={false} to="/settings/api-tokens" />
         </Step.Root>
       ),
-      when: (completedActions) => completedActions.includes('didCreateContent'),
+      when: (completedActions) =>
+        completedActions.includes(GUIDED_TOUR_REQUIRED_ACTIONS.contentTypeBuilder.addField),
```
Adzouz: It doesn't seem to be the right condition, causing the message to be displayed too soon. It should be `GUIDED_TOUR_REQUIRED_ACTIONS.contentManager.didCreateContent`

markkaylor: Yup! thanks I'll update

markkaylor: But it is strange the e2e tests still pass 🤔

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: Totally agree, and yes a little comment would be for visibility :)

Adzouz (approved):
Tested and all seems good now! 🚀 Thanks! (just be sure that the tests are failing just because of flaky ones, otherwise it's good for me)

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -2,35 +2,50 @@ import * as React from 'react';
...
+ * (ie didCreateContentTypeSchema | didCreateContent etc...)
+ */
+type RequiredActionValues = {
+  [K in keyof typeof GUIDED_TOUR_REQUIRED_ACTIONS]: (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K] extends Record<
+    string,
+    string
+  >
+    ? (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K][keyof (typeof GUIDED_TOUR_REQUIRED_ACTIONS)[K]]
+    : never;
+}[keyof typeof GUIDED_TOUR_REQUIRED_ACTIONS];
```
remidej, line 29: I think this would be easier to read if you extract types, something like this:

```ts
type RequiredActions = typeof GUIDED_TOUR_REQUIRED_ACTIONS;
type RequiredActionsKey = keyof RequiredActions
```

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
```diff
@@ -183,6 +192,14 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
+                <Button
+                  variant="tertiary"
+                  onClick={() => dispatch({ type: 'previous_step', payload: tourName })}
+                >
+                  <FormattedMessage id="tours.previous" defaultMessage="Previous" />
```
remidej, line 200: We usually use the hook and not the FormattedMessage component, is there a specific reason for it here?

packages/core/admin/server/src/services/guided-tour.ts
```diff
@@ -2,52 +2,34 @@ import { Core, Internal } from '@strapi/types';
...
-];
-
 export const createGuidedTourService = ({ strapi }: { strapi: Core.Strapi }) => {
+  /**
+   * @internal
+   * TODO:
+   * Remove completed actions from the server and handle it all on the frontend
+   * [x] didCreateContentTypeSchema
+   * [ ] didCreateContent
+   * [ ] didCreateApiToken
```
remidej, line 17: did you create tickets tickets for it? otherwise I'm not sure we'll ever get back to it

packages/core/admin/server/src/services/guided-tour.ts
```diff
@@ -2,52 +2,34 @@ import { Core, Internal } from '@strapi/types';
...
+      )) {
         const count = await strapi.documents(name as Internal.UID.ContentType).count({});
 
         if (count > 0) return true;
       }
 
       return false;
     })();
-    const didCreateContent = didCreateContentTypeSchema && hasContent;
+    const didCreateContent = hasContent;
```
remidej, line 32: why not name is didCreateContent from the beginning?

packages/core/admin/admin/src/components/GuidedTour/Step.tsx
markkaylor: I actually don't understand why we prefer the hook, I think this is cleaner and it's one less step (no pun intended).

packages/core/admin/server/src/services/guided-tour.ts
markkaylor: I'm updating each tour individually and will handle this as I go. This will be gone by the end of the week. So yes there are tickets.

packages/core/admin/server/src/services/guided-tour.ts
markkaylor: This will be gone by the end of the week

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
markkaylor: I'll try it but I think it too shallow, it's an object of objects and we want the values of the nested objects. Yours looks like it's only extracting keys.

packages/core/admin/server/src/services/guided-tour.ts
remidej: oh cool I was just just afraid this would be a _soon™_ situation

markkaylor: 😄 I get the concern

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
markkaylor: I found a more readable/maintable solution:
```
type ValueOf<T> = T[keyof T];
type NonEmptyValueOf<T> = T extends Record<string, never> ? never : ValueOf<T>;
export type CompletedActions = NonEmptyValueOf<ValueOf<typeof GUIDED_TOUR_REQUIRED_ACTIONS>>[];
```
I've made the update on my next PR for API tokens, gonna go ahead and merge this one. All tour updates are going into a temporary branch before merging at once into develop.

PR #24057, opened 2025-07-30
enhancement: resolved filter editability on clicking filter tag

### LINKED ISSUE #20967: Clicking a filter deletes the filter instead of allowing you to edit it

akash-dabhi-qed:
### What does it do?

Added the functionality to edit the filter tag when clicked after it is initially applied.  

https://github.com/user-attachments/assets/07f82995-2a9a-4742-9bd0-449bdd0cd98c

### Why is it needed?

To resolve the issue of not being able to edit filter once it is applied by pressing on the filter tag.

### How to test it?

### Related issue(s)/PR(s)

Fixes #20967

### DIFF
--- a/packages/core/admin/admin/src/components/Filters.tsx
+++ b/packages/core/admin/admin/src/components/Filters.tsx
@@ -43,33 +43,49 @@ interface FitlersContextValue {
 
 const [FiltersProvider, useFilters] = createContext<FitlersContextValue>('Filters');
 
-const isFilterBeingEdited = (
-  filter: Record<string, unknown>,
-  options: Filters.Filter[],
-  editingFilter: FilterFormData
-): boolean => {
-  const [attributeName] = Object.keys(filter);
-  if (attributeName !== editingFilter.name) return false;
+const getFilterDetails = (
+  filterEntry: Record<string, unknown>,
+  options: Filters.Filter[]
+): { name: string; operator: string; value: unknown } | null => {
+  const [name] = Object.keys(filterEntry);
+  const option = options.find((o) => o.name === name);
+  if (!option) {
+    return null;
+  }
 
-  const option = options.find(({ name }) => name === attributeName);
-  if (!option) return false;
-
-  const { type, mainField } = option;
   const operatorObj =
-    type === 'relation'
-      ? (filter[attributeName] as Record<string, unknown>)?.[mainField?.name ?? 'id']
-      : (filter[attributeName] as Record<string, unknown>);
-  if (typeof operatorObj !== 'object' || operatorObj === null) return false;
-  const entries = Object.keys(operatorObj);
-  if (entries.length === 0) return false;
-  const operator = entries[0];
-  const filterValue = (operatorObj as Record<string, unknown>)[operator];
+    option.type === 'relation'
+      ? (filterEntry[name] as Record<string, unknown>)?.[option.mainField?.name ?? 'id']
+      : filterEntry[name];
 
-  if (operator !== editingFilter.filter) return false;
-  if (FILTERS_WITH_NO_VALUE.includes(editingFilter.filter)) return true;
-  const decodedFilterValue =
-    typeof filterValue === 'string' ? decodeURIComponent(filterValue) : filterValue;
-  return decodedFilterValue === editingFilter.value;
+  if (typeof operatorObj !== 'object' || operatorObj === null) {
+    return null;
+  }
+
+  const [operator] = Object.keys(operatorObj as Record<string, unknown>);
+  if (!operator) {
+    return null;
+  }
+
+  return { name, operator, value: (operatorObj as Record<string, unknown>)[operator] };
+};
+
+const isFilterMatch = (
+  filterEntry: Record<string, unknown>,
+  options: Filters.Filter[],
+  target: FilterFormData
+): boolean => {
+  const details = getFilterDetails(filterEntry, options);
+  if (!details || details.name !== target.name || details.operator !== target.filter) {
+    return false;
+  }
+  if (FILTERS_WITH_NO_VALUE.includes(target.filter)) {
+    return true;
+  }
+
+  const decoded =
+    typeof details.value === 'string' ? decodeURIComponent(details.value) : details.value;
+  return decoded === target.value;
 };
 
 interface RootProps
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/components/Filters.tsx
```diff
@@ -434,11 +527,28 @@ const AttributeTag = ({
...
     onClick({ name, value, filter: operator });
   };
 
+  const handleEditClick = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    const decodedValue = FILTERS_WITH_NO_VALUE.includes(operator)
+      ? undefined
+      : decodeURIComponent(value);
+
+    setEditingFilter({
```
mathildeleg, line 475: Unfortunately here editing a "is null" or "is not null" filter does not work, probably because `filterValue === editingFilter.value` is always false for `$null/$notNull`, thus the existing filter is never found, and the “update” path effectively no-ops (no replacement).
The problem may come from storing the value for a filter in two different places, and they don’t use the same value for `$null / $notNull`: 

1. In handleSubmit, when the operator is $null or $notNull, the code always stores the string "true":

```const value = FILTERS_WITH_NO_VALUE.includes(data.filter)  ? 'true'   // ← stored in the query  : encodeURIComponent(data.value ?? '');```

So in the query you have something like: `{ someField: { $null: "true" } }`.

2. In AttributeTag, when you click the tag, we build the object we’re “editing” and set no value for $null/$notNull:

```const decodedValue = FILTERS_WITH_NO_VALUE.includes(operator)  ? undefined   // ← we put undefined in editingFilter  : decodeURIComponent(value);setEditingFilter({  name,  filter: operator,  value: decodedValue,  // undefined for $null / $notNull});```

So when editing a “is null” or “is not null” filter, editingFilter looks like: `{ name: "someField", filter: "$null", value: undefined }.`

File: packages/core/admin/admin/src/components/Filters.tsx [hunk 2]
```diff
@@ -158,37 +188,95 @@ const PopoverImpl = ({ zIndex }: { zIndex?: number }) => {
...
           method="POST"
-          initialValues={
-            {
-              name: options[0]?.name,
-              filter: BASE_FILTERS[0].value,
-            } satisfies FilterFormData
-          }
+          initialValues={initialValues}
           onSubmit={handleSubmit}
+          key={
```
mathildeleg, line 262: When we're in edit mode, `editingFilter.value` can be `undefined` for `$null/$notNull `filters. To match how we handle optional key parts elsewhere (e.g. Breadcrumbs with `?? 'root'`), we should normalize it so the key is always an explicit string:
```suggestion
          key={  editingFilter    ? `edit-${editingFilter.name}-${editingFilter.filter}-${editingFilter.value ?? 'empty'}`    : 'create'}
```
Using 'empty' keeps the key readable and avoids relying on `${undefined}` becoming "undefined".

packages/core/admin/admin/src/components/Filters.tsx
```diff
@@ -158,37 +188,95 @@ const PopoverImpl = ({ zIndex }: { zIndex?: number }) => {
...
-              ? {
-                  [fieldOptions.mainField?.name ?? 'id']: operatorValuePairing,
-                }
-              : operatorValuePairing,
-        },
-      ],
-    };
+    let newFilterQuery;
+
+    if (editingFilter) {
```
mathildeleg: To align with common style in the codebase (avoid let and deep nesting), we could refactor this block by:
* Building the new filter entry once and reusing it for both “replace” and “append”.
* Extracting helpers so the main flow is a single ternary and a .map, with early returns in the helpers instead of nested ifs.
* Using a single const for `newFilterQuery` (no let), derived from that ternary.
This keeps the same behaviour, makes the “edit vs add” logic easier to follow, and the helper that compares the current filter to the one being edited should be a good place to fix the `$null/$notNull` value matching mentioned below.

packages/core/admin/admin/src/components/Filters.tsx
```diff
@@ -489,9 +599,15 @@ const AttributeTag = ({
   })} ${operator !== '$null' && operator !== '$notNull' ? formattedValue : ''}`;
 
   return (
-    <Tag padding={1} onClick={handleClick} icon={<Cross />}>
-      {content}
-    </Tag>
+    <Box style={{ display: 'inline-block' }}>
+      <Tag padding={1} onClick={handleRemoveClick} style={{ cursor: 'pointer' }} icon={<Cross />}>
+        <Flex alignItems="center" gap={1}>
+          <Box onClick={handleEditClick} style={{ flex: 1 }}>
```
mathildeleg: I don't think you need the `<Box> `here, you could probably just do this: 

```suggestion
          <Flex alignItems="center" gap={1} onClick={handleEditClick} >
          {content}
        </Flex>
```

mathildeleg (changes requested):
Thanks so much for your PR, this would be a great UX improvement ✨
I have some comments to improve the code based on our codebase but also to fix a bug (the `$null / $notNull` filters cannot be edited).
Could you also add some tests for this new implementation?
Also could you fix the commit-lint issue :) 
Thank you!

packages/core/admin/admin/src/components/Filters.tsx
```diff
@@ -254,7 +342,12 @@ const PopoverImpl = ({ zIndex }: { zIndex?: number }) => {
                   type="submit"
                   fullWidth
                 >
-                  {formatMessage({ id: 'app.utils.add-filter', defaultMessage: 'Add filter' })}
+                  {editingFilter
+                    ? formatMessage({
+                        id: 'app.utils.update-filter',
```
mathildeleg, line 333: Forgot to say, you'll need to add this in the `translations/en.json` file please :)

mathildeleg:
I've implemented the changes I asked, asking for my colleagues's review since I re-wrote some of the code :)

jhoward1994:
Thanks for this great work @akash-dabhi-qed 

I made some changes, mainly refactors for readability and code organisation. I couldn't push to your fork but would you be able to check out the changes here ?

https://github.com/jhoward1994/strapi/tree/fix/20967-enhancements

mathildeleg:
> Thanks for this great work @akash-dabhi-qed
> 
> I made some changes, mainly refactors for readability and code organisation. I couldn't push to your fork but would you be able to check out the changes here ?
> 
> https://github.com/jhoward1994/strapi/tree/fix/20967-enhancements

Should be addressed [here](https://github.com/strapi/strapi/pull/24057/commits/63853e62efeaffba1dead931a2e884b76546b182) :)

PR #24060, opened 2025-07-30
feat: add a disable config for AI features

Bassel17:
### What does it do?

- Loads an AI config to the window strapi object in the browser
- incase it AI was disabled, the AI Chat is not rendered

### Why is it needed?

- Allows users to disable AI features if they want to

### How to test it?

- Add 
```
ai:{
   disable:true
}
```
to the admin config, navigate to the CTB, no AI chat should be present

### Related issue(s)/PR(s)

DX-2064

### DIFF
--- a/examples/getstarted/config/admin.js
+++ b/examples/getstarted/config/admin.js
@@ -31,4 +31,7 @@ module.exports = ({ env }) => ({
       },
     },
   },
+  ai: {
+    disable: false,
+  },
 });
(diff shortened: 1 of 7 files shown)


examples/getstarted/config/admin.js
```diff
@@ -31,4 +31,7 @@ module.exports = ({ env }) => ({
       },
     },
   },
+  ai: {
+    disable: false,
```
innerdvations: I think it's policy to use "enabled: false" everywhere, but let's confer with @maccomaccomaccomacco

jhoward1994 (approved):
Looks good other than Bens comment about inverting the condition

examples/getstarted/config/admin.js
maccomaccomaccomacco: true, as @innerdvations said

PR #24063, opened 2025-07-31
fix: incorrect config won't run app

markkaylor:
### What does it do?

Adds config with default env vars

### Why is it needed?

The example app won't start without them

### How to test it?

Run `examples/empty` it should work

### DIFF
--- a/examples/getstarted/package.json
+++ b/examples/getstarted/package.json
@@ -40,6 +40,6 @@
     "npm": ">=6.0.0"
   },
   "strapi": {
-    "uuid": "getstarted"
+    "uuid": "strapi_examples_getstarted"
   }
 }
--- a/examples/empty/config/admin.ts
+++ b/examples/empty/config/admin.ts
@@ -1,20 +1,19 @@
-export default ({ env }) => ({
+const adminConfig = ({ env }) => ({
   auth: {
-    secret: env('ADMIN_JWT_SECRET'),
+    secret: env('ADMIN_JWT_SECRET', 'example-token'),
   },
   apiToken: {
-    salt: env('API_TOKEN_SALT'),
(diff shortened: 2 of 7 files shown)


innerdvations (approved):

examples/getstarted/package.json
```diff
@@ -40,6 +40,6 @@
     "npm": ">=6.0.0"
   },
   "strapi": {
-    "uuid": "getstarted"
```
markkaylor, line 43: I think it's safe to change this. I think it's probably better to at least try and have some sort of uid and use a prefix for the vite check. Lmk if y'all have another idea or think this will cause an issue

Adzouz (approved):
All good for me, thanks a lot! 🙌

Adzouz:
> All good for me, thanks a lot! 🙌

Just need to address this tests failing 😬

markkaylor:
Great all the tests are broken 🙄

examples/getstarted/package.json
innerdvations: I think it needs to stay `getstarted` or it will affect our analytics. If it's causing problems as `getstarted` and needs to be something else, we might need to inform @maccomaccomaccomacco so he's aware of the new uuid

markkaylor: shit, ok I'll open a PR to revert

markkaylor: Are y'all aware that kitchensink, kitchensink-ts, and getstarted all had a uuid of `getstarted`? Is that what you want for analytics, all example apps to have the same uuid? Seems weird...

markkaylor: Draft PR here https://github.com/strapi/strapi/pull/24067 if we want to revert the change

PR #24067, opened 2025-07-31
chore: revert uuid in examples

markkaylor:
### What does it do?

Reverts the prefix on uuids

### Why is it needed?

Apparently they should all be getstarted fro analytics

### How to test it?

https://github.com/strapi/strapi/pull/24063

### DIFF
--- a/examples/empty/package.json
+++ b/examples/empty/package.json
@@ -34,6 +34,6 @@
     "npm": ">=6.0.0"
   },
   "strapi": {
-    "uuid": "strapi_examples_empty"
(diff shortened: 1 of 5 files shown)


innerdvations (approved):
LGTM! Although I do like your idea, we just need to check with @maccomaccomaccomacco if using a prefix is fine

markkaylor:
> LGTM! Although I do like your idea, we just need to check with @maccomaccomaccomacco if using a prefix is fine

Sounds good I'll wait for a response in that thread before merging

PR #24068, opened 2025-07-31
feat: last activity homepage widget

Adzouz:
### What does it do?

Screenshot 2025-07-31 at 12 13 52

New homepage widget: Last activity (audit logs enabled - Enterprise Edition premium feature).
Users that have the option Audit Logs enabled in their project will be able to see this new widget on the homepage that will display the last four actions made on the project (like the ones in Settings > Audit Logs) with shortcut to display more information and redirect directly to the settings page.

### Why is it needed?

Enhance the user experience on the platform homepage and provide more useful information to the user when they connect.

### How to test it?

- Login to the admin interface and go to the Homepage (has to be an EE account that has Audit Logs feature enabled, otherwise it won't show up)
  -> You should see the widget with the last 4 actions on the platform (date order DESC - more recent first)

_If you want to see it changing, you can perform any actions you want on the platform and come on the homepage again to see them appearing on the widget (creating, publishing, updating, deleting an entry / content-type, login / logout, create, udpate, delete user, webhooks, API token, locales, etc.)_

🚀


packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -164,22 +149,21 @@ const HomePageCE = () => {
       <Layouts.Content>
         <Flex direction="column" alignItems="stretch" gap={8} paddingBottom={10}>
           <GuidedTourHomepageOverview />
-          <Grid.Root gap={5}>
-            {filteredWidgets.map((widget) => {
-              return (
+          {loading ? (
+            <Box position="absolute">
+              <Widget.Loading />
+            </Box>
```
remidej: this is the loader designed for individual widgets, for page-level loading you should use <Page.Loading />

packages/core/admin/ee/admin/src/components/AuditLogs/Widgets.tsx
```diff
@@ -0,0 +1,126 @@
...
+          const action = formatMessage(
+            {
+              id: `Settings.permissions.auditLogs.${item.action}`,
+              // @ts-expect-error – getDefaultMessage probably doesn't benefit from being so strongly typed unless we just add string at the end.
+              defaultMessage: getDefaultMessage(item.action),
+            },
+            { model: (item.payload?.model as string) ?? '' }
+          );
+          const actionDate = formatTimeStamp(item.date);
+          const userDisplayName = item?.user?.displayName ?? '-';
```
remidej: i'm a bit confused by when you use optional chaining and when you don't. you should only use the ? if TS says the value can be null/undefined

Adzouz: You're right, this line I admit I copied pasted it but it doesn't make sense with the rest so I'll update everything to be consistent.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
Adzouz: I just kept the absolute positioning because otherwise the loader isn't centered in the view.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -150,8 +150,8 @@ const HomePageCE = () => {
         <Flex direction="column" alignItems="stretch" gap={8} paddingBottom={10}>
           <GuidedTourHomepageOverview />
           {loading ? (
-            <Box position="absolute">
-              <Widget.Loading />
+            <Box position="absolute" top={0} left={0} right={0} bottom={0}>
+              <Page.Loading />
             </Box>
```
remidej: I don't think you need the position absolute box, Page.Loading manages this itself

remidej (commented):
looks like there's a frontend unit test failing

packages/core/admin/admin/src/pages/Home/HomePage.tsx
Adzouz: Unfortunately I had to do this because otherwise the loader is not centered in the page (because of the title pushing the loader that has a 100vh in height) so I thought it was better to do it that way to keep it centered.

packages/core/admin/ee/admin/src/components/AuditLogs/Widgets.tsx
```diff
@@ -0,0 +1,126 @@
...
+            >
+              <Td>
+                <Typography title={action} variant="omega" textColor="neutral800">
+                  {action}
+                </Typography>
+              </Td>
+              <Td>
+                <Typography title={actionDate} variant="omega" textColor="neutral800">
+                  {actionDate}
+                </Typography>
```
remidej: just wanna double check that this is what we want and not relative time?

remidej (dismissed):
tested, lgtm :)

packages/core/admin/ee/admin/src/components/AuditLogs/Widgets.tsx
Adzouz: I think we want to display the exact timestamp to the user. That's what already in place in the audit logs page. I'll double check just to be sure!

Adzouz: I put the relative time instead.

remidej (approved):

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -132,17 +107,27 @@ const HomePageCE = () => {
...
   const getAllWidgets = useStrapiApp('UnstableHomepageCe', (state) => state.widgets.getAll);
-  const filteredWidgets = React.useMemo(
-    () =>
-      getAllWidgets().filter(
-        (widget) =>
-          !widget.roles ||
-          user?.roles?.some((userRole) => widget.roles?.find((role) => userRole.code === role))
-      ),
-    [getAllWidgets, user?.roles]
-  );
```
unrevised6419, line 145: Hey @Adzouz, will the widget `roles` logic be brought back anytime?

I was testing with a user that has 0 content type permissions, but they still still see the key statistics widget

Image

Issue #24066, opened 2025-07-31
Changing a field from type "time" to type "datetime" in Strapi Cloud causes deployments to fail


OcneanuVlad:
### Node Version

22.17.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.19.0

### Operating System

MacOS

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

Changing a field from type "time" to type "datetime" in Strapi Cloud causes deployments to fail if you have a minimum of 1 entry with the field filled

### Steps to Reproduce

1. Create a collection type with a "time" field
2. Deploy to Strapi Cloud
3. Create an entry of that specific collection type and fill in the time field
4. On local, modify the collection type's "time" field to a "datetime" type
5. Deploy to Strapi Cloud
6. The deployment fails with a migration error

### Expected Behavior

The migration should handle such changes with a default placeholder date

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

OcneanuVlad:
Internal note: TID 2725

kasonde:
<img width="871" height="81" alt="Image" src="https://github.com/user-attachments/assets/3791f8ce-f4bf-45b3-88ab-4c972f31aefe" />

Failing DB query is here. Possible solution could be either editing every record in the DB with a TZ/appropriate data or adding 'WITHOUT TIME STAMP' to the failing DB query to account for data that doesn't have it.

Bassel17:
fixed by #24073

PR #24073, opened 2025-08-01
fix(core): handle time to dateTime and dateTime to time conversion wi…

butcherZ:
### What does it do?

- add default value for date when converting time to dateTime
- removing date when converting dateTime to time
- added unit tests

### Why is it needed?

PostgreSQL stores different types:
   - `time` → `time without time zone` (e.g., "14:30:00")
   - `datetime` → `timestamp without time zone` (e.g., "2024-01-15 14:30:00")

2. PostgreSQL cannot automatically cast between these types because:
   - A time value has no date information
   - A timestamp requires both date and time

3. our default migration generates:
   ```sql
   ALTER TABLE "my_times" ALTER COLUMN "hour" TYPE timestamp(6)
   ```
   This fails because PostgreSQL doesn't know what date to use.

### How to test it?

- creat a strapi app with experimental `0.0.0-experimental.ef1d1db8f9c879c5db48472a0add235295c17c50`
- On local, create a collection type with a "time" field
- Deploy to Strapi Cloud
- In cloud, create an entry of that specific collection type and fill in the time field
- On local, modify the collection type's "time" field to a "datetime" type
- Deploy to Strapi Cloud
- The deployment should not fail

### Related issue(s)/PR(s)

#24066 

dx-2148

### DIFF
--- a/packages/core/database/src/schema/builder.ts
+++ b/packages/core/database/src/schema/builder.ts
@@ -472,30 +472,29 @@ const createHelpers = (db: Database) => {
   /**
    * Apply column properties after type conversion
    */
-  const applyColumnProperties = async (tableName: string, columnName: string, column: Column) => {
+  const applyColumnProperties = async (
+    trx: Knex.Transaction,
+    tableName: string,
+    columnName: string,
+    column: Column
+  ) => {
     // Apply NOT NULL constraint
     if (column.notNullable) {
-      await db.connection.raw(`ALTER TABLE ?? ALTER COLUMN ?? SET NOT NULL`, [
-        tableName,
-        columnName,
-      ]);
+      await trx.raw(`ALTER TABLE ?? ALTER COLUMN ?? SET NOT NULL`, [tableName, columnName]);
     } else {
-      await db.connection.raw(`ALTER TABLE ?? ALTER COLUMN ?? DROP NOT NULL`, [
-        tableName,
-        columnName,
-      ]);
+      await trx.raw(`ALTER TABLE ?? ALTER COLUMN ?? DROP NOT NULL`, [tableName, columnName]);
     }
 
     // Apply default value
     if (column.defaultTo !== undefined) {
       const [defaultValue, defaultOpts] = castArray(column.defaultTo);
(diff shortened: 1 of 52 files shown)


innerdvations (commented):
This looks ok, but I think we should also document that this will happen on postgres (including that Cloud uses postgres) and advise against changing type from TIME -> DATETIME

And in this PR, I would add a log with warning on the server that this is happening, since it's not really intuitive behaviour.

Let's also talk to @maccomaccomaccomacco about if we should add a warning in the CTB itself when a user attempts this, or possibly even prevent converting from TIME -> DATETIME in the future

innerdvations:
@alexandrebodin @Marc-Roig Could one of you please take a look at this too?

The issue is that postgres does not allow converting time to datetime like we allow in the db sync.

I think in the future we may want to prevent this option altogether, but that will take a lot more discussion and work to implement and I want to move on this quickly to unblock Cloud users encountering the issue (server won't start if you try converting `time` to `datetime`)

So this seems to be a decent workaround for now, I just want to make sure we don't make the problem worse.

alexandrebodin (commented):
Hi, thanks for the contribution. 

here are the main things I'm looking for to approve

- Add a conversion utility function in the dialects themselves that will return the using statements.
- Split the code in functions so it's more digestable :)

innerdvations (commented):
I can't QA it today, but code LGTM!

packages/core/database/src/schema/builder.ts
```diff
@@ -423,11 +427,132 @@ const createHelpers = (db: Database) => {
...
+      }
+    }
+
+    // Apply conversions
+    for (const conversion of conversionsToApply) {
+      const { column, sql, params, currentType, targetType, warning } = conversion;
+
+      // Log warning about type conversion
+      db.logger.warn(
+        `⚠️  Database type conversion: Converting column "${column.name}" in table "${table.name}" from "${currentType}" to "${targetType}".`
```
alexandrebodin: that log plus the specific log for the warning that's a bit much :D Claude is very verbose

Bassel17 (approved):
LGTM!
tested as well

PR #24075, opened 2025-08-01
Fix/22733 make boolean nullable

### LINKED ISSUE #22733: Not nullable Boolean fields
### Node Version

23.1.0

### NPM/Yarn/PNPM Version

10.9.0

### Strapi Version

5.4.0

### Operating System

Windows 11

### Database

MySQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

Boolean fields used to be nullable in version 4, having a "Clear" button to reset it to null, however in version 5 there is no such an option which means that after we set a Boolean value then we cannot make it null again.

Null value could be used as "unknown" in some cases

### Steps to Reproduce

1. Add a Boolean field on a entity
2. Add a value (T/F) on an entry's Boolean field

### Expected Behavior

Having a "Clear" button as an option above Boolean fields so their value could be unset.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

akash-dabhi-qed:
### What does it do?

Made the Boolean toggle null-able.

https://github.com/user-attachments/assets/5cfe1c36-a235-4844-9e3f-585aa745945e

### Why is it needed?

Currently there is no way to change a toggle to an `unset` state, this PR adds a clear button that brings back the functionality to assign null value to toggle.

### How to test it?

### Related issue(s)/PR(s)

Fixes #22733

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
@@ -1,7 +1,6 @@
 import { forwardRef, memo } from 'react';
 
-import { Toggle, useComposedRefs, Field, Flex } from '@strapi/design-system';
-// import { Toggle, useComposedRefs, Field, Button } from '@strapi/design-system';
+import { Toggle, useComposedRefs, Field } from '@strapi/design-system';
 import { useIntl } from 'react-intl';
 
 import { useFocusInputField } from '../../hooks/useFocusInputField';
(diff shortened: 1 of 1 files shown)


akash-dabhi-qed:
I initially thought it would be an easy fix, I made the following changes
changed `checked` of  `Toggle` to `checked={field.value === false ? false : field.value === true ? true : null}`
Adding clear button:
```
{(field.value === true || field.value === false) && (
  <Button variant="tertiary" size="S" onClick={() => field.onChange(name, null)}>
    {formatMessage({
      id: 'app.components.Boolean.clear',
      defaultMessage: 'Clear',
    })}
  </Button>
)}
```
This Immediately gave an unlikely error: (this is also present in older [v4.14.6](https://github.com/user-attachments/assets/3092100c-9a05-46a6-aacb-2a5175bdfae7) of strapi)

https://github.com/user-attachments/assets/15d6783e-aa4d-4f5a-87c6-c93397cd9222

- Using clear from an initial state changes the value of the toggle, triggering the save button. (It goes from `undefined` to `null`)
- Selecting `true` value on the `toggle` is easy, but the `false` value requires 2 clicks, it cannot be set using 1 click.
- The toggle between true and false works as intended.

On changing the Button onclick's onChange to take undefined instead of null gave much graver issues:

https://github.com/user-attachments/assets/98a84f71-3163-4e53-b422-0f22811b1e36


- The Toggle component's `onChange` event was being processed by Strapi's form validation/normalization system
- It was rejecting or reverting non-initial boolean values back to undefined
- This created a race condition where UI updates would be applied then immediately undone

Therefore to resolve these issues of the twice clicking to get to false value and reverting back state from undefined to last state

I added invisible `divs` that circumvent the `toggle` when initially setting value from `null` or `undefined`. When going back to selecting between `true` or `false` value, it reverted back to the toggle.

I also made the old look of the `Clear` button, but also kept a bit modern version available just in case it is needed.

Adzouz:
Hi @akash-dabhi-qed, thanks a lot for your contribution 🙏
Unfortunately I'll close this PR since it was already handled by this PR: https://github.com/strapi/strapi/pull/24836
Keep up the good work 🚀

PR #24091, opened 2025-08-04
chore: improve getstarted dummy preview page

remidej:
### What does it do?

ℹ️ this is only a meta PR about our dev environment, it can be ignored for pre-release QA

Makes changes in the getstarted example app:
- hides the admin sidebar in the dummy-preview
- makes dummy-preview consume the public content API
- makes dummy-preview require a STRAPI_ADMIN_API_TOKEN env variable
- changes the URL pattern of the dummy page, because the content API can't use the model uid, it needs the plural name
- refactors dummy-preview data fetching to use react router loader
- removes the controller override that was in the address content type

### Why is it needed?

To make it easier to test the preview improvements that are coming up.

### How to test it?

Set a read-only API token in the .env of your getstarted app. Then open up the preview page for different content types. Make sure to try both single and collection types
### Related issue(s)/PR(s)

fixes https://github.com/strapi/content-squad/issues/57

### DIFF
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -1,97 +1,138 @@
 import * as React from 'react';
-import { useParams } from 'react-router-dom';
+import { useParams, useLoaderData, useRevalidator } from 'react-router-dom';
 
 // @ts-ignore
 import { Page, Layouts } from '@strapi/admin/strapi-admin';
-import { unstable_useDocument as useDocument } from '@strapi/content-manager/strapi-admin';
-import { Grid, Flex, Typography, JSONInput } from '@strapi/design-system';
+import { Grid, Flex, Typography, JSONInput, Box } from '@strapi/design-system';
+
+const filterAttributes = (item) => {
+  const excludedKeys = ['documentId', 'id', 'createdAt', 'updatedAt', 'publishedAt'];
+  return Object.entries(item).filter(([key]) => !excludedKeys.includes(key));
+};
 
 const PreviewComponent = () => {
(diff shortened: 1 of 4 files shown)


examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -1,97 +1,138 @@
...
+                    Locale
+                  </Typography>
+                  <Typography tag="dd">{locale}</Typography>
+                </Grid.Item>
+              </Grid.Root>
+              <Typography variant="delta" tag="h3">
+                Rest API data
+              </Typography>
+              {revalidator.state === 'loading' && <Typography>Refreshing data...</Typography>}
+              {!data && <Typography textColor="neutral600">No data found</Typography>}
```
Adzouz: You could merge into one condition `!data ? () : ()`

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -1,97 +1,138 @@
...
+  const revalidator = useRevalidator();
 
   React.useEffect(() => {
-    const handleStrapiUpdate = (event) => {
-      if (event.data?.type === 'strapiUpdate') {
-        refetch();
+    const handleMessage = (event) => {
+      const { origin, data } = event;
+
+      if (origin !== 'http://localhost:1337') {
```
Adzouz: What if another port is setup? Or even an url alias?

examples/getstarted/src/admin/preview/dummy-preview.jsx
remidej: this is some dumb ai code indeed 😬 👍

PR #24098, opened 2025-08-05
New workflow to tag issues before 2021 to prepare for cleanup

oiorain:
This PR is the very first step of the [RFC [Github] Proposal of a process to close stale issues](https://www.notion.so/strapi/Github-Proposal-of-a-process-to-close-stale-issues-2088f359807480b7aa03e17f5b40605d).

## What does it do?
This PR introduces a new github workflow that will add a "stale-before-2021" tag to all issues created before 2021 that do not carry any of these labels:
- "severity: high"
- "severity: critical"
- "issue: security"

## Why is it needed?
The original workflow had several issues:

- Incorrect GitHub Actions expression syntax (${{ TAG }}) for shell variables
- Inconsistent variable naming (TAG vs label in different contexts)
- Unclear documentation about which issues are being excluded
- The workflow would fail to properly label issues due to these syntax errors

## How to test it?

1. Manual testing: Trigger the workflow manually via GitHub Actions UI using the workflow_dispatch trigger
2. Verify the changes:
- Check that [issues created before 2021-01-01](https://github.com/strapi/strapi/issues?q=is:issue%20is:open%20created:2010-01-01..2021-01-01) are properly labeled with "stale-before-2021"
- Confirm that issues with "severity: high", "severity: critical", or "issue: security" labels have not been tagged.
- Verify that issues with milestones are excluded.
- Ensure issues already labeled with "stale-before-2021" are not processed again.

## What's next?

Next, I will add a step to this workflow to close these issues with a specific message.

### DIFF
--- a/.github/workflows/cleanup_issues_before_2021.yml
+++ b/.github/workflows/cleanup_issues_before_2021.yml
@@ -1,0 +1,60 @@
+name: Close Pre-2021 Issues
+on:
+  # schedule:
+    # - cron: '35 15 * * *'  # Runs daily at 3:30 PM - stop after first run
+  workflow_dispatch: # Allows manual triggering from the Github UI
+
+permissions:
+  issues: write  # Required to close issues and post comments
+  contents: read  # Needed by most GitHub Actions
+  pull-requests: read  # Needed to access PR data
+
+jobs:
+  close-inactive:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          
+      - name: Label old issues
+        run: |
+          # Set the cutoff date (issues created before this date will be labeled)
+          CUTOFF_DATE="2021-01-01"
+          LABEL="stale-before-2021"
+          
+          echo "Labeling issues created before $CUTOFF_DATE with $LABEL"
+          
+          # Get all open issues created before the cutoff date
+          # Exclude issues with 'severity: high' or 'severity: critical' and 'issue: security' label and issues in milestones
+          curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            "https://api.github.com/repos/${{ github.repository }}/issues?state=open&per_page=100" | \
+            jq -r --arg cutoff "$CUTOFF_DATE" '
+              .[] | 
+              select(.created_at < $cutoff) |
+              select(.pull_request == null) |
+              select(any(.labels[]?; .name == "severity: high") | not) |
+              select(any(.labels[]?; .name == "severity: critical") | not) |
+              select(any(.labels[]?; .name == "issue: security") | not) |
+              select(.milestone == null) |
+              select(any(.labels[]?; .name == $label) | not) |
+              .number
+            ' | while read issue_number; do
+              if [ ! -z "$issue_number" ]; then
+                echo "Labeling issue #$issue_number with $label" 
+                
+                # Add the stale-before-2021 label
+                curl -s -X POST \
+                  -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+                  -H "Accept: application/vnd.github.v3+json" \
+                  "https://api.github.com/repos/${{ github.repository }}/issues/$issue_number/labels" \
+                  -d "{\"labels\": [\"$label\"]}"
+                
+                echo "Labeled issue #$issue_number"
+              fi
+            done


pwizla:
LGTM! 😊 🚀

pwizla:
Re-reading it (and asking a (robotic) friend of mine… 😅  *cough*), I might have overlooked this 👀 

The script uses `LABEL="stale-before-2021"` but in the `jq` command you're doing `select(any(.labels[]?; .name == $label) | not)`, however `$label` is not passed to jq. 
This might result in a condition doing nothing 😅 , and issues that are already labeled could be relabeled.

💡 The suggested fix is to pass `$LABEL` to `jq` as a variable:

```
bashjq -r --arg cutoff "$CUTOFF_DATE" --arg label "$LABEL" ' ... '
```

.github/workflows/cleanup_issues_before_2021.yml
```diff
@@ -0,0 +1,60 @@
...
+          LABEL="stale-before-2021"
+          
+          echo "Labeling issues created before $CUTOFF_DATE with $LABEL"
+          
+          # Get all open issues created before the cutoff date
+          # Exclude issues with 'severity: high' or 'severity: critical' and 'issue: security' label and issues in milestones
+          curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            "https://api.github.com/repos/${{ github.repository }}/issues?state=open&per_page=100" | \
+            jq -r --arg cutoff "$CUTOFF_DATE" '
```
oiorain: `bashjq -r --arg cutoff "$CUTOFF_DATE" --arg label "$LABEL" ' `
@pwizla if I understand well, you're suggesting to replace this line by this , right? 
Indeed i think i had caught it in my draft but didnt carry it over 🤔

oiorain:
Than kyou for the suggestion @pwizla ! I added a fix.

.github/workflows/cleanup_issues_before_2021.yml
```diff
@@ -0,0 +1,60 @@
...
+permissions:
+  issues: write  # Required to close issues and post comments
+  contents: read  # Needed by most GitHub Actions
+  pull-requests: read  # Needed to access PR data
+
+jobs:
+  close-inactive:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
```
derrickmehaffy: Shouldn't need to checkout the repo unless we are loading a script or something from the repo

.github/workflows/cleanup_issues_before_2021.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/checkout@v4
+        
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
```
derrickmehaffy: Same here we shouldn't need node since you are currently only running a bash script

.github/workflows/cleanup_issues_before_2021.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          node-version: '20'
+          
+      - name: Label old issues
+        run: |
+          # Set the cutoff date (issues created before this date will be labeled)
```
derrickmehaffy: I'm wondering if we could save a lot of bash logic here by instead using some prebuilt action runners such as the one we use for other labeling automation: https://github.com/strapi/strapi/blob/2a9b9b61cfdc484a1736b8c7d460c850ecc64608/.github/workflows/issues_handleLabel.yml#L23

We already use issue helper: https://github.com/actions-cool/issues-helper
For example: https://actions-cool.github.io/issues-helper/advanced#close-issues

---

Entirely up to you on this, doing it manually in bash with curl and bash logic is totally fine but might be slightly harder to maintain if people don't know bash

.github/workflows/cleanup_issues_before_2021.yml
```diff
@@ -0,0 +1,60 @@
...
+          CUTOFF_DATE="2021-01-01"
+          LABEL="stale-before-2021"
+          
+          echo "Labeling issues created before $CUTOFF_DATE with $LABEL"
```
derrickmehaffy: Purely subjective opinion here but I would break up the script runner for fetching labels vs closing them (you can hand data off between action steps as variables: https://www.edwardthomson.com/blog/github_actions_15_sharing_data_between_steps)

derrickmehaffy: (this would let us use the issue helper I mentioned above as we could use a script to fetch exactly what we want but then use the helper to do other stuff like set labels/close/lock/ect)

derrickmehaffy (commented):
Should be fine for testing but left some comments

.github/workflows/cleanup_issues_before_2021.yml
oiorain: Yes! I've seen this package indeed and I think I will use it for other things. I'm using `actions/stale@v3` in the next step. This thing is you can apply the action to issues opened _after_ a date but not before so this is the workaround I found. Once the cleanup is done and we work on all issues regardless of date the code will be more straight forward. This is just temporary actions that will not be maintained over time.

.github/workflows/cleanup_issues_before_2021.yml
oiorain: Yes, you're absolutely right. As I said in my previous comment, this is an exploration. I'm only acting on a tiny subset to make sure the logic is sound and I'm not messing up the repo. I agree that this is not well named. I will edit so this is inly labelling. The closing part will come later today once I confirm that this is working well. Then i'll edit to make it work for all issues prior to 2024 like the RFC stated.

oiorain:
I ended up renaming the current action which only takes care of labelling and adding the closing action since it will be run once the labelling is done. This gives a fuller picture of the work.

Bassel17:
you might want to format with prettier and push again

PR #24102, opened 2025-08-05
feat: implement session management system with generaterefreshtoken

jhoward1994:
### What does it do?

- Implements that start of a session management system with JWT-based refresh tokens
- Adds database-backed session storage with a new `admin::session` content type
- Device ID, origin, and expiration management
- Integrates SessionManager service into the core Strapi instance accessible via `strapi.sessionManager`
- Includes automatic cleanup of expired sessions

### Why is it needed?

- Enables secure session management for admin users with proper token lifecycle control
- Creates foundation for enhanced authentication and authorization features
- Provides device-based session tracking

### How to test it?

- Run the new API integration tests: `yarn test:api tests/api/core/admin/session-manager.test.api.ts`
- Test session creation and token generation through the SessionManager service
- Verify database sessions are properly stored in the `admin_sessions` table
- Test automatic cleanup of expired sessions
- Verify the service is accessible via `strapi.sessionManager` in your application

### Related issue(s)/PR(s)

DX-1987
DX-2141

### DIFF
--- a/packages/core/admin/package.json
+++ b/packages/core/admin/package.json
@@ -84,6 +84,7 @@
     "@radix-ui/react-context": "1.0.1",
     "@radix-ui/react-toolbar": "1.0.4",
     "@reduxjs/toolkit": "1.9.7",
+    "@strapi/core": "5.19.0",
     "@strapi/design-system": "2.0.0-rc.29",
     "@strapi/icons": "2.0.0-rc.29",
     "@strapi/permissions": "5.20.0",
--- a/packages/core/admin/server/src/content-types/session.ts
+++ b/packages/core/admin/server/src/content-types/session.ts
@@ -1,0 +1,70 @@
+export default {
+  collectionName: 'admin_sessions',
+  info: {
+    name: 'Session',
+    description: 'Admin user session management',
+    singularName: 'session',
+    pluralName: 'sessions',
+    displayName: 'Session',
+  },
+  pluginOptions: {
+    'content-manager': {
+      visible: false,
+    },
+    'content-type-builder': {
+      visible: false,
+    },
+  },
+  attributes: {
+    user: {
+      type: 'string',
+      required: true,
+      configurable: false,
+      private: true,
+    },
+    sessionId: {
+      type: 'string',
+      unique: true,
+      required: true,
+      configurable: false,
+      private: true,
+    },
+    deviceId: {
+      type: 'string',
+      required: true,
+      configurable: false,
+      private: true,
+    },
+    origin: {
+      type: 'string',
+      required: true,
+      configurable: false,
+      private: true,
+    },
+    expiresAt: {
+      type: 'datetime',
+      required: true,
+      configurable: false,
+      private: true,
+    },
+  },
+  config: {
+    attributes: {
+      user: {
+        hidden: true,
+      },
+      sessionId: {
+        hidden: true,
+      },
+      deviceId: {
+        hidden: true,
+      },
+      origin: {
+        hidden: true,
+      },
+      expiresAt: {
+        hidden: true,
+      },
+    },
+  },
+};
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -138,6 +138,10 @@ class Strapi extends Container implements Core.Strapi {
     return this.get('telemetry');
   }
 
+  get sessionManager(): Modules.SessionManager.SessionManagerService {
+    return this.get('sessionManager');
+  }
+
   get store(): Modules.CoreStore.CoreStore {
     return this.get('coreStore');
   }
--- a/packages/core/core/src/providers/sessionManager.ts
+++ b/packages/core/core/src/providers/sessionManager.ts
@@ -1,0 +1,34 @@
+import { defineProvider } from './provider';
+import { createSessionManager } from '../services/session-manager';
+
+interface AdminAuthConfig {
+  secret?: string;
+  options?: Record<string, unknown>;
+}
+
+export default defineProvider({
+  init(strapi) {
+    // Get JWT secret from admin auth settings (same as admin token service)
+    const adminAuth = strapi.config.get<AdminAuthConfig>('admin.auth', {});
+    const jwtSecret = adminAuth.secret;
+
+    if (!jwtSecret) {
+      throw new Error(
+        'Missing admin.auth.secret configuration. The SessionManager requires a JWT secret'
+      );
+    }
+
+    const config = {
+      jwtSecret,
+      refreshTokenLifespan: 30 * 24 * 60 * 60, // 30 days in seconds
+      accessTokenLifespan: 60 * 60, // 1 hour in seconds
+    };
+
+    strapi.add('sessionManager', () =>
+      createSessionManager({
+        db: strapi.db,
+        config,
+      })
+    );
+  },
+});
--- a/packages/core/core/src/services/session-manager.ts
+++ b/packages/core/core/src/services/session-manager.ts
@@ -1,0 +1,154 @@
+import crypto from 'crypto';
+import jwt from 'jsonwebtoken';
+import type { Database } from '@strapi/database';
+
+export interface SessionProvider {
+  create(session: SessionData): Promise<SessionData>;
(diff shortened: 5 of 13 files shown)


packages/core/core/src/providers/sessionManager.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!jwtSecret) {
+      throw new Error(
+        'Missing admin.auth.secret configuration. The SessionManager requires a JWT secret'
+      );
+    }
+
+    const config = {
+      jwtSecret,
+      refreshTokenLifespan: 30 * 24 * 60 * 60, // 30 days in seconds
```
derrickmehaffy: We should make these configurable, similar to what we do with the rate-limit middleware

packages/core/admin/server/src/content-types/session.ts
```diff
@@ -0,0 +1,70 @@
...
+      private: true,
+    },
+    expiresAt: {
+      type: 'datetime',
+      required: true,
+      configurable: false,
+      private: true,
+    },
+  },
+  config: {
```
derrickmehaffy: We shouldn't need to hide the attributes if the content-type is hidden

packages/core/admin/package.json
```diff
@@ -84,6 +84,7 @@
     "@radix-ui/react-context": "1.0.1",
     "@radix-ui/react-toolbar": "1.0.4",
     "@reduxjs/toolkit": "1.9.7",
+    "@strapi/core": "5.19.0",
```
Bassel17: why did we need to add `@strapi/core` as a dependency here ?

packages/core/core/src/providers/sessionManager.ts
Bassel17: ➕

packages/core/admin/server/src/content-types/session.ts
```diff
@@ -0,0 +1,51 @@
...
+  pluginOptions: {
+    'content-manager': {
+      visible: false,
+    },
+    'content-type-builder': {
+      visible: false,
+    },
+  },
+  attributes: {
+    user: {
```
innerdvations: I would call this `userId` like the other id types. But I'm not sure if it will be confused with a particular user system, since it's actually generic. What do you think?

packages/core/core/src/services/session-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+  generateSessionId(): string {
+    return crypto.randomBytes(16).toString('hex');
+  }
+
+  async generateRefreshToken(
+    userId: string,
+    deviceId: string,
+    origin: string
+  ): Promise<{ token: string; sessionId: string }> {
+    await this.provider.deleteExpired();
```
innerdvations: Good idea, I definitely prefer this over adding another cron job.

But let's make it more efficient to avoid an extra query every run. Like only run it 5% of the time (though I don't like that if it's non-deterministic) or a counter that only runs it once every 50 calls (which might behave a little unexpectedly if scaled horizontally, but I think that's fine here) and/or run it completely asynchronously without await, and/or some other idea you might have

packages/core/core/src/providers/sessionManager.ts
innerdvations: I think it needs to be definable per-origin, basically:

- admin: `admin.auth.refreshTokenLifespan` and `admin.auth.accessTokenLifespan`
- u&p: `plugins['users-permissions'].config.jwt.refreshTokenLifespan` and  `plugins['users-permissions'].config.jwt.refreshTokenLifespan` (note that we already have `expiresIn` that I think we should leave alone and not re-use, so that the legacy option still works the same)

I would say those options are passed in with each call to create the tokens (maximum flexibility), or else add a way to define origins in this file and use those settings based on the origin (simpler)

packages/core/core/src/Strapi.ts
```diff
@@ -138,6 +138,10 @@ class Strapi extends Container implements Core.Strapi {
     return this.get('telemetry');
   }
 
+  get sessionManager(): Modules.SessionManager.SessionManagerService {
```
innerdvations, line 141: Wondering if this should just be `sessions` since we don't use `Manager` anywhere else. Anyone have any thoughts either way?

@maccomaccomaccomacco ?

packages/core/core/src/services/session-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    const sessionId = this.generateSessionId();
+    const expiresAt = new Date(Date.now() + this.config.refreshTokenLifespan * 1000);
+
+    await this.provider.create({
```
innerdvations, line 137: I think once this succeeds, we should delete the previous refresh token.

However, if we do it automatically here there would be a problem with this edge case:

- User refreshes token
- Token is refreshed successfully and we delete the old one
- Something happens and token fails to send back to the user via HTTP, or the user's browser fails to save it correctly
- User no longer has a valid token to refresh and is logged out

Probably just have the front-end wait to be sure it got the new token, then call a more specific `logout` passing an the old refresh token ID rather than an origin+device. or do it automatically server-side but only after it has been successfully returned to the user.

If it's too complex to add here, just add a ticket to add it later before we release it.

packages/core/core/src/services/session-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+
+    const payload: Omit<RefreshTokenPayload, 'iat' | 'exp'> = {
+      userId,
+      sessionId,
+      type: 'refresh',
+    };
+
+    const token = jwt.sign(payload, this.config.jwtSecret, {
+      expiresIn: this.config.refreshTokenLifespan,
```
innerdvations, line 152:
```suggestion
      expiresIn: this.config.refreshTokenLifespan,
      algorithm: 'RS256'
```
It would be better to specify algorithm, but then we should also specify in jwt.verify, which should be fine for admin since we control that, but would probably be incompatible with keeping the legacy option in u&p. But we can deal with that when we add the u&p integration, we just need to make sure it's in the ticket so we don't forget.

I think RS256 is best because then we could potentially allow third party verification of the JWTs in the future but HS256 would probably be fine too.

packages/core/core/src/services/session-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async deleteBySessionId(sessionId: string): Promise<void> {
+    await this.db.query(this.contentType).delete({
+      where: { sessionId },
+    });
+  }
+
+  async deleteByIdentifier(userId: string): Promise<void> {
+    await this.db.query(this.contentType).delete({
```
innerdvations:
```suggestion
    await this.db.query(this.contentType).deleteMany({
```

packages/core/core/src/services/session-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+      where: { user: userId },
+    });
+  }
+
+  async deleteExpired(): Promise<void> {
+    await this.db.query(this.contentType).delete({
```
innerdvations:
```suggestion
    await this.db.query(this.contentType).deleteMany({
```

packages/core/core/src/providers/sessionManager.ts
innerdvations: but if that's coming in a follow-up PR, that's fine, we don't have to address it here

packages/core/core/src/services/session-manager.ts
jhoward1994: wdyt of the approach in [`53fec8f` (#24102)](https://github.com/strapi/strapi/pull/24102/commits/53fec8f17f79b515bec628231c0eec4942ea5dea)
?

jhoward1994: Is the intention to clean up this users expired sessions every time & throttle the deletion of all expired sessions as we are doing?

The current version fails a couple of your API test cases as the users expired sessions aren't removed every time, but I think this is intentional?

innerdvations: yeah, the tests I added don't take that into account, they would have to be updated once we fix this.

I think doing it only once ever 50 runs is fine!

PR #24096, opened 2025-08-05
enhancement: improve api token tour

markkaylor:
### What does it do?

- updates the api token tour to use one of the existing api tokens instead of a creating a token
- updates the wording to be more "fun" and explicit about actions the user has to perform

### Why is it needed?

Feedback

### How to test it?

EXPERIMENTAL: `0.0.0-experimental.4a33056a161ac309a7bfc1ea77527009939b7ffe`
- It's better to QA on the experimental
- Otherwise, currently locally you MUST click "generate new token" before you will be able to click "view token" 
- Otherwise it should work out of the box on the experimental

Go to settings -> api tokens
The tour should start
Do the tour

Looking for feedback on the wording in the tooltips and the flow of the _api token_ tour. There are no designs for this.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #24110, opened 2025-08-06
enhancement: improve content-type builder tour (#24053)

markkaylor:
### What does it do?

- Refactor to move all step related code into a folder and create individual files for each tour's steps
- Update the content manager wording
- Going from CTB tour to CM tour we should always end up on the editview page, this is where the CM tour starts now

### Why is it needed?

### How to test it?
Reset local storage

- First follow the CTB tour and create a collection type
- After save when clicking next you should end up on the create page of the content-type you just created
- You should see the CM tour. Do the tour. 

### Related issue(s)/PR(s)

This PR is based on https://github.com/strapi/strapi/pull/24094 but it should wait for https://github.com/strapi/strapi/pull/24096 to be merged and then update to refactor api token. 24096 should be reviewed first.


packages/core/admin/admin/src/components/GuidedTour/Steps/ApiTokensSteps.tsx
markkaylor, line 1: NOTE: This file just moves the steps definition for the tour to its own file. It's just an organization change. Nothing really to review.

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentTypeBuilderSteps.tsx
```diff
@@ -0,0 +1,191 @@
...
+    <Step.Actions>
+      <StepCount tourName="contentTypeBuilder" />
+      <GotItAction
+        onClick={() => {
+          // Ensure the completed action is removed
+          // in the event the user already has a schema but is still doing the tour
+          dispatch({
+            type: 'remove_completed_action',
+            payload: GUIDED_TOUR_REQUIRED_ACTIONS.contentTypeBuilder.createSchema,
+          });
```
markkaylor, line 109: NOTE: This file just moves the steps definition for the tour to its own file. It's mainly an organization change. This is the only real diff.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -127,7 +146,7 @@ const GuidedTourContext = ({
     enabled,
     completedActions: [],
   });
-  const migratedTourState = migrateTourSteps(storedTours);
+  const migratedTourState = migrateTours(storedTours);
```
markkaylor, line 149: NOTE: migrateTourSteps is no longer necessary since we no longer store the length of the store in state / local storage. 

migrateTours is not _actually_ necessary but it future proofs the guided tour allowing tours to be added and removed between versions. I can remove it from the PR if you feel it should be its own PR

docs/docs/docs/01-core/admin/05-guided-tour.md
```diff
@@ -318,9 +326,9 @@ Tour state is automatically persisted to localStorage using the `usePersistentSt
 
 Tours integrate with the backend through:
 
-- `useGetGuidedTourMetaQuery()` - Fetches tour metadata from the server
-- Completed actions are synchronized between client and server.
-- First-time user detection (`isFirstSuperAdminUser`)
+`useGetGuidedTourMetaQuery()`
```
remidej, line 329: You said in another PR you wanted to get rid of the backend. Is this part here to stay?

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentManagerSteps.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const state = useGuidedTour('ContentManagerActions', (s) => s.state);
+  const dispatch = useGuidedTour('ContentManagerActions', (s) => s.dispatch);
+
+  const isSingleType = collectionType === 'single-types';
+
+  const currentStepOffset = state.tours.contentManager.currentStep + 1;
+  const displayedCurrentStep = (() => {
+    if (isSingleType && currentStepOffset > collectionTypeSpecificSteps.length) {
+      return currentStepOffset - 1;
+    }
```
remidej: * should the `-1` maybe be `- collectionTypeSpecificSteps.length + 1`? with the +1 being for cosmetics
* I'm surprised you always do the substraction. It would make sense to me if the CT specific steps were first, but there's also an Intro step before that. On that Intro do we want the substraction? Since it's before the skipped steps

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
remidej: not necessary at all? why not removing it then?

markkaylor: Not necessary for the changes made on this PR since it doesn't add or remove a tour. In the future if someone adds a tour like `preview: createTour('preview', [...])` there will be nothing to do it will update the users local storage when they upgrade.

markkaylor: > why not removing it then?

Fine with removing it but I'm anticipating tours in the future, for example ai architect, I already wrote the migration so I thought why not merge it now but if you feel it's out of place I can remove it.

docs/docs/docs/01-core/admin/05-guided-tour.md
markkaylor: I didn't want to get rid of the backend entirely, we need it for some things like `isFirstSuperAdmin` but I don't want to tie it tightly to the concept of `completedActions`, so yes this part is here to stay.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
remidej: Thanks! I'm not saying we should remove it, just trying to understand better

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentManagerSteps.tsx
markkaylor: > should the -1 maybe be - collectionTypeSpecificSteps.length + 1? with the +1 being for cosmetics

yes, thanks for the catch. the `-1` is no good

> I'm surprised you always do the substraction. It would make sense to me if the CT specific steps were first, but there's also an Intro step before that. On that Intro do we want the substraction? Since it's before the skipped steps

I'll have to take a look and consider this tomorrow about to take off for the day

markkaylor: The places where you see subtraction is mostly for what total step count is displayed to the user. We need it on all steps displaying step count for single types. So yes for the Introduction step we need it.

markkaylor:
Sorry git and the diff are messed up working on it

markkaylor:
> Sorry git and the diff are messed up working on it

Updated sorry for the force push 😬

Adzouz:
image

The step 3 is a bit cropped because on top of the page. It's happening on non draftAndPublish entities

Adzouz:
Question here: If an entry already exist, the action GUIDED_TOUR_REQUIRED_ACTIONS.contentManager.createContent is true (well I guess so) because I directly see the API tokens message even if I didn't do anything. For this specific tour, won't we want to force the user to create a new entry (even if there are some already) before displaying the API token message?

markkaylor:
> Question here: If an entry already exist, the action GUIDED_TOUR_REQUIRED_ACTIONS.contentManager.createContent is true (well I guess so) because I directly see the API tokens message even if I didn't do anything. For this specific tour, won't we want to force the user to create a new entry (even if there are some already) before displaying the API token message?

This action GUIDED_TOUR_REQUIRED_ACTIONS.contentManager.createContent should only be dispatched when you click the publish button. So you should be forced to publish the entry before you see the message to go to the api tokens tour:

https://github.com/user-attachments/assets/18a4c8ac-3271-448c-9019-f995bf15ac9d


I did notice though if it's not a d&p content type it's not dispatched when clicking save. I'll update to handle that.

markkaylor:
> image
> The step 3 is a bit cropped because on top of the page. It's happening on non draftAndPublish entities

Thanks guess it must have something to do with the number of fields as well since for my simple test it puts it below

Screenshot 2025-08-13 at 16 26 37

Adzouz:
> > image
> > The step 3 is a bit cropped because on top of the page. It's happening on non draftAndPublish entities
> 
> Thanks guess it must have something to do with the number of fields as well since for my simple test it puts it below
> 
> Screenshot 2025-08-13 at 16 26 37

Not to be annoying but what if the user put so many fields that the popover won't be visible because on the bottom of the page. Is there a way to attach the popover to the first field? Or at least put an empty Box element just before the first field.
Also I couldn't reproduce the error I had with the API token appearing without doing any action, so probably some weird behaviour on my side, I hope it's not reproducible.

markkaylor:
> Not to be annoying but what if the user put so many fields that the popover won't be visible because on the bottom of the page. Is there a way to attach the popover to the first field? Or at least put an empty Box element just before the first field.
Also I couldn't reproduce the error I had with the API token appearing without doing any action, so probably some weird behaviour on my side, I hope it's not reproducible.

Oh I'm definitely gonna look for a solution just trying to pin down what causes it 😄

markkaylor:
Updated to handle the collision. Not sure I _love_ it but the only other idea I have right now is to display it in the same spot as the Introduction step. That might be kind of strange for single types since you would have two popovers in the exact same spot. wdyt?

Screenshot 2025-08-13 at 16 54 16

Adzouz:
> Updated to handle the collision. Not sure I _love_ it but the only other idea I have right now is to display it in the same spot as the Introduction step. That might be kind of strange for single types since you would have two popovers in the exact same spot. wdyt?
> 
> Screenshot 2025-08-13 at 16 54 16

Do you have an example of what it would look like with single types in that case?

markkaylor:
> Do you have an example of what it would look like with single types in that case?

https://github.com/user-attachments/assets/abafee04-c3de-4cc4-86fd-e0302a58868b

markkaylor:
I think I prefer what I pushed considering this collision as an edge case. The guided tour is for a fresh strapi project, it encourages content-type with d and p enabled and that's what is set by default for all content types. For that case it works as expected regardless of how many fields

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentManagerSteps.tsx
remidej, line 1: Image

I think the padding issue is back

remidej (commented):
Sorry I forgot to come back to it after my 1st review. It looks really good

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentManagerSteps.tsx
markkaylor: ahh yeah, looks like I only updated the CM then. I'lll pull from develop to update since it should be fixed over there now or just update here directly.

PR #24111, opened 2025-08-06
future: set up preview injected script

remidej:
### What does it do?

sets up the boilerplate needed to get code that lives in the content manager to run on the websites of users.
### Why is it needed?

we need to own a script that runs in the websites of users for the next preview improvements
### How to test it?

open the preview for any entry in getstarted, open your console, you should see a log from within the preview page

### Related issue(s)/PR(s)

fixes https://github.com/strapi/content-squad/issues/59

### DIFF
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -46,10 +46,10 @@ const PreviewComponent = () => {
   }, []);
 
   React.useEffect(() => {
-    if (document) {
+    if (data) {
       window.parent?.postMessage({ type: 'previewReady' }, '*');
     }
-  }, [document]);
+  }, [data]);
 
   return (
     <Box
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -11,7 +11,7 @@ const filterAttributes = (item) => {
 };
 
 const PreviewComponent = () => {
-  const { apiName, documentId, locale, status: documentStatus, collectionType } = useParams();
+  const { apiName, documentId, locale, status: documentStatus } = useParams();
   const data = useLoaderData();
   const revalidator = useRevalidator();
 
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -31,6 +31,10 @@ const PreviewComponent = () => {
       if (data?.type === 'strapiUpdate') {
         // The data is stale, force a refetch
         revalidator.revalidate();
+      } else if (data?.type === 'strapiScript') {
+        const script = window.document.createElement('script');
+        script.textContent = data.payload.script;
+        window.document.head.appendChild(script);
       }
     };
 
@@ -40,6 +44,12 @@ const PreviewComponent = () => {
       window.removeEventListener('message', handleMessage);
     };
   }, []);
+
+  React.useEffect(() => {
+    if (document) {
+      window.parent?.postMessage({ type: 'previewReady' }, '*');
+    }
+  }, [document]);
 
   return (
     <Box
--- a/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
+++ b/packages/core/content-manager/admin/src/preview/pages/Preview.tsx
@@ -33,6 +33,8 @@ import { buildValidParams } from '../../utils/api';
 import { createYupSchema } from '../../utils/validation';
 import { PreviewHeader } from '../components/PreviewHeader';
 import { useGetPreviewUrlQuery } from '../services/preview';
+import { INTERNAL_EVENTS, PUBLIC_EVENTS } from '../utils/constants';
+import { previewScript } from '../utils/previewScript';
 
 import type { UID } from '@strapi/types';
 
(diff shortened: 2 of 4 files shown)


packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -87,6 +89,32 @@ const AnimatedArrow = styled(ArrowLineLeft)<{ $isSideEditorOpen: boolean }>`
...
+function getSendMessage(iframe: React.RefObject<HTMLIFrameElement>) {
+  return (type: MessageType, payload?: unknown) => {
+    if (!iframe.current) return;
+
+    const { origin } = new URL(iframe.current.src);
+
+    iframe.current.contentWindow?.postMessage(
+      {
+        type,
+        ...(payload !== undefined && { payload }),
```
markkaylor, line 111: what if payload is null?

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -31,6 +31,10 @@
       if (data?.type === 'strapiUpdate') {
         // The data is stale, force a refetch
         revalidator.revalidate();
+      } else if (data?.type === 'strapiScript') {
+        const script = window.document.createElement('script');
+        script.textContent = data.payload.script;
```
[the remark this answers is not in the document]
markkaylor: 😳

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -41,6 +45,12 @@ const PreviewComponent = () => {
     };
   }, []);
 
+  React.useEffect(() => {
+    if (document) {
+      window.parent?.postMessage({ type: 'previewReady' }, '*');
+    }
+  }, [document]);
```
markkaylor: I can't find where `document` comes from...

examples/getstarted/src/admin/preview/dummy-preview.jsx
markkaylor, line 1: collectionType destructured from useParams is no longer used

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -116,6 +144,23 @@ const PreviewPage = () => {
   );
   const device = DEVICES.find((d) => d.name === deviceName) ?? DEVICES[0];
 
+  // Listen for ready message from iframe before injecting script
+  React.useEffect(() => {
+    const handleMessage = (event: MessageEvent) => {
+      if (event.data?.type === PUBLIC_EVENTS.PREVIEW_READY) {
+        const script = `(${previewScript.toString()})()`;
```
markkaylor, line 151: Just confused what's happening here. You stringify the script but then what's the `()` about? Does it become a callable function?

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -87,6 +89,32 @@ const AnimatedArrow = styled(ArrowLineLeft)<{ $isSideEditorOpen: boolean }>`
   transition: rotate 0.2s ease-in-out;
 `;
 
+type MessageType =
+  | (typeof INTERNAL_EVENTS)[keyof typeof INTERNAL_EVENTS]
+  | (typeof PUBLIC_EVENTS)[keyof typeof PUBLIC_EVENTS];
+
+/**
+ * A function factory so we can generate a new sendMessage everytime we need one.
+ * We can't store and reuse a single sendMessage because it needs to have a stable identity
```
markkaylor, line 98: Won't it have a stable identity if it's defined outside the react component? Otherwise wouldn't the useEffect complain the getSendMessage is not in its dep array?

I don't really have any issue with any of this just curious...

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: it should be ignored, i used a loose check so it covers undefined and null

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: When I stringify the function, what I get is its internal implementation. I could just dump it as the script content, but then it would require top-level await support, which doesn't have as good compatibility as regular await because it's more recent.

So it's safer to wrap it in a function and call it, which is what I'm doing here via a self invoking function. The long form version would be this, I can change it it you prefer:

```
const script = `
  const runScript = () => {
    ${previewScript.toString()}
  }
  runScript();
`
```

examples/getstarted/src/admin/preview/dummy-preview.jsx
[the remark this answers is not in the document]
remidej: Copilot is definitely not gonna like that. But Alex approved the approach. It should be safe if you check the origin of the even before loading the script.

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
markkaylor: ok thanks, this is fine no need to update

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: >Won't it have a stable identity if it's defined outside the react component

It does have a stable identity, but it also needs a reference to the iframe ref, so it can't be 100% isolated outside the component. it's why I have this compromise of moving it outside but taking the ref as a param

>Otherwise wouldn't the useEffect complain the getSendMessage is not in its dep array?

the getSendMessage is stable indeed because it's outside the component. And refs don't need to be passed to dependency arrays because their identity is stable too

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
remidej: Oh you mean what if we use null intentionally and want to keep it?

PR #24116, opened 2025-08-06
chore: change query to include parameters instead of filtering response in github action

oiorain:
This is a followup to https://github.com/strapi/strapi/pull/24098.

- Changed the API endpoint to use one that enables us to have arguments directly applied instead of filtering the result. 
- Pagination is not handled here. This will only treat 100 issues at a time and may need reruns.


Bassel17 (approved):
Should be good !

PR #24125, opened 2025-08-07
chore: implement a new prompt to enable A/B test and track the event

Mcastres:
### What does it do?

Implement a new prompt in the CLI at then end asking if a user is interested in participating in A/B test.
If yes, we send an event to track the percentage of users interested and then take the decisiong to implement or not A/B tests.

### Why is it needed?

To potentially propose A/B tests on the product so we can easily improve the product.

### How to test it?

- Run the following command: `node packages/cli/create-strapi-app/bin/index.js`
- You should get asked to enable A/B test. By default it is false.
- An event should be sent.

### DIFF
--- a/packages/cli/create-strapi-app/src/create-strapi.ts
+++ b/packages/cli/create-strapi-app/src/create-strapi.ts
@@ -257,7 +257,7 @@ async function createApp(scope: Scope) {
   }
 
   if (isABTestEnabled) {
-    await trackUsage({ event: 'didEnabledABTest', scope });
+    await trackUsage({ event: 'didEnableABTest', scope });
   }
 }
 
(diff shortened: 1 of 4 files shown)


jhoward1994:
Makes sense to me. So we're just tracking this for percentages and not to gather other details from these users?

packages/cli/create-strapi-app/src/create-strapi.ts
```diff
@@ -254,6 +255,10 @@ async function createApp(scope: Scope) {
       logger.fatal('Failed to start your Strapi application');
     }
   }
+
+  if (isABTestEnabled) {
+    await trackUsage({ event: 'didEnabledABTest', scope });
```
Bassel17: I would change the event name here to `didEnableABTest` just makes more sense grammatically
but overall the code looks good

Mcastres: Sure, thanks for the review 🙏

Mcastres:
> Makes sense to me. So we're just tracking this for percentages and not to gather other details from these users?

Yes, the idea is to just see how many users would be willing to participate in A/B testing

jhoward1994 (approved):

PR #24127, opened 2025-08-07
future: create preview overlay div with highlights

remidej:
### What does it do?

⚠️ this doesn't affect production users, it can be ignored for pre-release QA

Adds actual functionality into the preview script from #24111

- creates an overlay div over the whole page that doesn't affect user interactions
- tracks elements that have a data-source-strapi tag
- displays a "highlight" (blue box) on top of these elements. their position is dynamic, will react to scroll and resize events
- unregisters all its logic, listeners and observers so the script can be called multiple times safely
- handles double clicks on higlights (see console log)
- lets users customize the color of highlights via a window property

### Why is it needed?

for live preview / visual editing

### How to test it?

open a preview page in getstarted. hover elements that display strapi content. you should see the highlights. double click them and see the console log. Then resize, scroll, etc., the highlight positioning should remain right
### Related issue(s)/PR(s)

fixes https://github.com/strapi/content-squad/issues/60

### DIFF
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -40,10 +40,7 @@ const previewScript = (shouldRun = true) => {
   const createOverlaySystem = () => {
     // Clean up before creating a new overlay so we can safely call previewScript multiple times
     window.__strapi_previewCleanup?.();
-    const existingOverlay = document.getElementById(OVERLAY_ID);
-    if (existingOverlay) {
-      existingOverlay.remove();
-    }
+    document.getElementById(OVERLAY_ID)?.remove();
 
     const overlay = document.createElement('div');
     overlay.id = OVERLAY_ID;
@@ -241,7 +238,6 @@ const previewScript = (shouldRun = true) => {
   const highlightManager = createHighlightManager(overlay);
   const observers = setupObservers(highlightManager);
   const eventHandlers = setupEventHandlers(highlightManager);
-
   createCleanupSystem(overlay, observers, eventHandlers);
 };
 
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -2,7 +2,7 @@
 declare global {
   interface Window {
     __strapi_previewCleanup?: () => void;
-    __strapi_HIGHLIGHT_HOVER_COLOR?: string;
+    STRAPI_HIGHLIGHT_HOVER_COLOR?: string;
   }
 }
 
@@ -17,7 +17,7 @@ const previewScript = (shouldRun = true) => {
    * Params
    * ---------------------------------------------------------------------------------------------*/
   const HIGHLIGHT_PADDING = 2; // in pixels
-  const HIGHLIGHT_HOVER_COLOR = window.__strapi_HIGHLIGHT_HOVER_COLOR ?? '#4945ff'; // dark primary500
+  const HIGHLIGHT_HOVER_COLOR = window.STRAPI_HIGHLIGHT_HOVER_COLOR ?? '#4945ff'; // dark primary500
   const SOURCE_ATTRIBUTE = 'data-strapi-source';
   const OVERLAY_ID = 'strapi-preview-overlay';
   const INTERNAL_EVENTS = {
@@ -69,7 +69,7 @@ const previewScript = (shouldRun = true) => {
     const highlights: HTMLElement[] = [];
     const eventListeners: EventListenersList = [];
 
-    const drawOverlay = (target: Element, highlight: HTMLElement) => {
+    const drawHighlight = (target: Element, highlight: HTMLElement) => {
       if (!highlight) return;
 
       const rect = target.getBoundingClientRect();
@@ -82,7 +82,7 @@ const previewScript = (shouldRun = true) => {
       highlights.forEach((highlight, index) => {
         const element = elements[index];
         if (element && highlight) {
-          drawOverlay(element, highlight);
+          drawHighlight(element, highlight);
         }
       });
     };
@@ -135,7 +135,7 @@ const previewScript = (shouldRun = true) => {
         highlights.push(highlight);
         overlay.appendChild(highlight);
 
-        drawOverlay(element, highlight);
+        drawHighlight(element, highlight);
       }
     });
 
@@ -163,10 +163,10 @@ const previewScript = (shouldRun = true) => {
       highlightManager.updateAllHighlights();
     };
 
-    // Find all scrollable ancestors for all tracked elements
     const scrollableElements = new Set<Element | Window>();
-    scrollableElements.add(window); // Add window as a special case
+    scrollableElements.add(window);
 
+    // Find all scrollable ancestors for all tracked elements
     highlightManager.elements.forEach((element) => {
       let parent = element.parentElement;
       while (parent) {
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -2,8 +2,7 @@
 declare global {
   interface Window {
     __strapi_previewCleanup?: () => void;
-    __strapi_HIGHLIGHT_COLOR?: string;
-    __strapi_DISABLE_STEGA_DECODING?: boolean;
+    __strapi_HIGHLIGHT_HOVER_COLOR?: string;
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+          (element as Element).removeEventListener('scroll', observers.updateOnScroll);
+        }
+      });
+
+      // Remove highlight event listeners
+      eventHandlers.forEach(({ element, type, handler }) => {
+        element.removeEventListener(type, handler);
+      });
+
+      overlay.remove();
```
remidej, line 229: I know we said the other days that all overlays could be moved up at the body root instead of in a .overlay div. But personally I like the div because I can remove everything at once

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -1,6 +1,9 @@
 // NOTE: This override is for the properties on _user's site_, it's not about Strapi Admin.
 declare global {
-  interface Window {}
+  interface Window {
+    __strapi_previewCleanup?: () => void;
+    __strapi_HIGHLIGHT_HOVER_COLOR?: string;
```
markkaylor: what does the leading __ mean here? I always interpret this as private or internal

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+  /* -----------------------------------------------------------------------------------------------
+   * Functionality pieces
+   * ---------------------------------------------------------------------------------------------*/
+
+  const createOverlaySystem = () => {
+    // Clean up before creating a new overlay so we can safely call previewScript multiple times
+    window.__strapi_previewCleanup?.();
+    document.getElementById(OVERLAY_ID)?.remove();
+
+    const overlay = document.createElement('div');
```
markkaylor, line 45: Idk if it's helpful for what you're doing here for the overlay or highlights but [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) has methods for controlling visibility like  `show` and `close`.

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+    type: keyof HTMLElementEventMap;
+    handler: EventListener;
+  }>;
+
+  const createHighlightManager = (overlay: HTMLElement) => {
+    const elements = window.document.querySelectorAll(`[${SOURCE_ATTRIBUTE}]`);
+    const highlights: HTMLElement[] = [];
+    const eventListeners: EventListenersList = [];
+
+    const drawOverlay = (target: Element, highlight: HTMLElement) => {
```
markkaylor: Are you redrawing the overlay to include a highlight or should this be called `drawHighlight` just a bit turned around what we are actually talking about here highlight or overlay or both

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+          will-change: transform;
+          transition: outline-color 0.1s ease-in-out;
+        `;
+
+        // Move hover detection to the underlying element
+        const mouseEnterHandler = () => {
+          highlight.style.outlineColor = HIGHLIGHT_HOVER_COLOR;
+        };
+        const mouseLeaveHandler = () => {
+          highlight.style.outlineColor = 'transparent';
```
markkaylor, line 108: Regarding my comment above about dialog I would think here you could just highglight.close() instead of using styles but this is fine too

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+        element.addEventListener('mouseleave', mouseLeaveHandler);
+        element.addEventListener('dblclick', doubleClickHandler);
+        element.addEventListener('mousedown', mouseDownHandler);
+
+        // Store event listeners for cleanup
+        eventListeners.push(
+          { element, type: 'mouseenter', handler: mouseEnterHandler },
+          { element, type: 'mouseleave', handler: mouseLeaveHandler },
+          { element, type: 'dblclick', handler: doubleClickHandler },
+          { element, type: 'mousedown', handler: mouseDownHandler as EventListener }
```
markkaylor, line 132: how come only this one has to be type casted?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+
+    resizeObserver.observe(document.documentElement);
+
+    const updateOnScroll = () => {
+      highlightManager.updateAllHighlights();
+    };
+
+    // Find all scrollable ancestors for all tracked elements
+    const scrollableElements = new Set<Element | Window>();
+    scrollableElements.add(window); // Add window as a special case
```
markkaylor: what's the special case?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -23,9 +33,212 @@ const previewScript = (shouldRun = true) => {
...
+      resizeObserver.observe(element);
+    });
+
+    resizeObserver.observe(document.documentElement);
+
+    const updateOnScroll = () => {
+      highlightManager.updateAllHighlights();
+    };
+
+    // Find all scrollable ancestors for all tracked elements
```
markkaylor: You are just creating a new set here and not finding anything should this comment be moved to 169 maybe?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: I was mixing 2 patterns, I'm updating it to:
- __strapi_previewCleanup: unchanged, because it's actually internal, not meant to be called by users
- STRAPI_HIGHLIGHT_HOVER_COLOR: will be documented, users would set it themselves

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: In this one I need the event.detail property, which only exists on MouseEvent, but EventListener is not compatible with MouseEvent. There may be a cleaner fix but I couldn't find it quickly

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: If we implemented dialogs I think it would be for each highlight, not at the overlay level. But I'm not sure it's the best for this case. I see the highlights more as the triggers for the modal than as the modal itself

markkaylor: totally up to whatever you think is best, just saw that was available so I thought I'd share it with you in case you hadn't considered it

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: i removed the comment, it's more confusing than anything 👍

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: I appreciate that! I won't use it this time specifically but it's a good tool to know about

markkaylor (approved):
Screenshot 2025-08-12 at 16 49 18

I had this error and then added a full access api token to my env file. Is that what you should do? Is it in the docs? Maybe update the PR description for whoever QAs this in during the release?

Otherwise the tests check out, nice work 👍

remidej:
> Maybe update the PR description for whoever QAs this in during the release?

image

This is a change from #24091, because for stega support we need the preview to query the content API and not the CM API. And I think it's easier to force the use of an API token than to tell people to open up all their routes to the public role.

@markkaylor I'm happy to document it somewhere, but I'm not sure where? I was hoping the error message would be self explanatory. A read only token should be enough btw

markkaylor:
> > Maybe update the PR description for whoever QAs this in during the release?
> 
> image
> This is a change from #24091, because for stega support we need the preview to query the content API and not the CM API. And I think it's easier to force the use of an API token than to tell people to open up all their routes to the public role.
> 
> @markkaylor I'm happy to document it somewhere, but I'm not sure where? I was hoping the error message would be self explanatory. A read only token should be enough btw

I do think for us developers the error is definitely clear enough to know what to do, maybe it's enough for the QA engineers as well just trying to anticipate a ping on a release week 🙃

markkaylor:
oh I think I understood 😆 you trying to release this this week and the QA yourself?

remidej:
I'm not trying to release this in particular this week, but I am the releaser, and in charge of the QA since Marion is off

Now that I think about it though, release QA (and even blitz sessions) are normally done on fresh applications, so I think only developers should see this page

PR #24129, opened 2025-08-07
fix(content-manager): compute document status per-locale to prevent cross-locale state bleed

LisnicVictor:
### What does it do?

Scopes Content Manager document status computation by locale to prevent cross-locale state bleed.

Technical changes:
- Server
  - document-metadata.getStatus():
    - When a locale is provided, prefer the counterpart (draft/published) of the same locale among localizations.
    - Ignore other locales when deciding if an entry is published/modified/draft for the current locale.
    - Fall back to the first available counterpart only if no same-locale version exists.
  - controllers/collection-types.find():
    - Filter availableStatuses by the requested locale before computing setStatus for each returned item.
- Tests
  - Added locale-aware status unit tests:
    - packages/core/content-manager/server/src/services/__tests__/document-metadata.locale-status.test.ts

### Why is it needed?

Previously, changing/publishing an entry in one locale could inadvertently flip the status of entries with the same documentId in other locales (e.g., editing en-GB made the en version look “modified”). This fix ensures the status reflects only the state of the current locale’s version.

Scenarios addressed:
- Published EN and modified draft EN-GB no longer marks EN as modified.
- Modified draft EN compared to published EN returns “modified” as expected.
- If only another-locale published exists, current-locale draft remains “draft.”
- If a version has no locale metadata, fallback remains preserved.

### How to test it?

Automated
- Run package unit tests for Content Manager:
  - yarn workspace @strapi/content-manager run test:unit
- The new tests are located at:
  - packages/core/content-manager/server/src/services/__tests__/document-metadata.locale-status.test.ts

Manual (Admin UI with i18n and Draft & Publish enabled)
1) Create a collection-type with i18n enabled (e.g., Article).
2) Create an entry in EN, publish it.
3) Switch to EN-GB, create/edit the localized version (leave as draft or modify after publishing).
4) Verify in the list view:
   - EN stays “published” unless its own draft changed in EN.
   - EN-GB shows “modified” only when EN-GB’s draft is newer than its own published version.
   - Changing EN-GB does not alter EN’s status, and vice versa.

### Related issue(s)/PR(s)

N/A (please link an issue if one exists)

### DIFF
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -154,9 +154,18 @@ export default {
 
     const setStatus = (document: any) => {
       // Available status of document
-      const availableStatuses = documentsAvailableStatus.filter(
-        (d: any) => d.documentId === document.documentId
-      );
+      // Scope by documentId and, when present, by the same locale to avoid cross-locale state
+      const availableStatuses = documentsAvailableStatus.filter((d: any) => {
+        if (d.documentId !== document.documentId) {
+          return false;
+        }
+        // If the document has a locale, only consider statuses from the same locale
+        if (document.locale) {
+          return d.locale === document.locale;
+        }
+        // If no locale on the document, keep all
+        return true;
+      });
       // Compute document version status
       document.status = documentMetadata.getStatus(document, availableStatuses);
       return document;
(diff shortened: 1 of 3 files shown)


mathildeleg:
Hi, thanks so much for your contribution! I believe this has been fixed by another PR, which has been included in the release of v5.31.1.
I'll close your PR now but if you still encounter the issue, please re-open it :)

PR #24132, opened 2025-08-08
fix: clear values of conditionally hidden fields to prevent UI inconsistencies

### LINKED ISSUE #24120: Conditional fields retain values when hidden instead of clearing them
### Node Version

22.17.0

### NPM/Yarn/PNPM Version

NPM 10.9.2

### Strapi Version

5.21.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

In the content editor, fields that are conditionally hidden do not clear their existing values. Specifically, when a field becomes hidden due to a condition no longer being met, its previously set value remains stored and causes related fields to stay visible incorrectly.

### Steps to Reproduce

1. Add the following example schema to your project:
`
"type": {
  "type": "enumeration",
  "required": true,
  "default": "internal-link",
  "enum": ["internal-link", "external-link", "simple-download", "multilingual-download"]
},
"linkTo": {
  "type": "enumeration",
  "conditions": {
    "visible": {
      "==": [
        { "var": "type" },
        "internal-link"
      ]
    }
  },
  "required": true,
  "enum": ["Pages", "Posts"]
},
"page": {
  "type": "relation",
  "relation": "oneToOne",
  "target": "api::page.page",
  "conditions": {
    "visible": {
      "==": [
        { "var": "linkTo" },
        "Pages"
      ]
    }
  }
}
`
2. Select "internal-link" for the type field.
3. Select "Pages" for the linkTo field.
4. Change the type field to "external-link".

The page field remains visible even though the linkTo field is now hidden because it still holds a value, causing inconsistent UI behavior.

### Expected Behavior

When a field is conditionally hidden, its value should be cleared to prevent dependent fields from remaining visible or causing unintended effects.

_No response_

### Confirmation Checklist

chhedadhruv:
### What does it do?

This PR fixes an issue where conditionally hidden fields in the Content Manager retain their old values, causing dependent fields to remain visible incorrectly.  

**Technical changes:**
- Added a `maskedValues` memoized object that excludes values of fields currently hidden by visibility conditions.
- Updated all conditional visibility checks (`rulesEngine.evaluate`) to use `maskedValues` instead of the raw `fieldValues`.
- This ensures that dependent field conditions are evaluated without stale hidden field data, without mutating form state (so the modified flag is unaffected).

---

### Why is it needed?

Previously, when a field became hidden due to a condition no longer being met, its stored value persisted in `fieldValues`.  
Because other field visibility conditions could still read this stale value, dependent fields would stay visible even though the triggering field was hidden.  

This fix ensures hidden fields are effectively ignored in visibility evaluations, preventing inconsistent UI behavior in the content editor.  

**Fixes:** #24120

---

### How to test it?
1. Add the example schema from the issue description (with `type`, `linkTo`, and `page` fields, each having visibility conditions).
2. Select `"internal-link"` for `type` and `"Pages"` for `linkTo`.
3. Change `type` to `"external-link"`.
4. **Before fix:** `page` remains visible because `linkTo` still holds `"Pages"`.
5. **After fix:** `page` correctly hides as soon as `linkTo` becomes hidden.

---

#### Before:
https://github.com/user-attachments/assets/bf5167b2-f584-4fcc-97e4-b33226379324

#### After:

https://github.com/user-attachments/assets/f3c60f08-1851-40b3-839a-671f55d2200c

---

### Related issue(s)/PR(s)

Fixes #24120

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormLayout.tsx
@@ -63,6 +63,32 @@ const FormLayout = ({ layout, document, hasBackground = true }: FormLayoutProps)
     });
   };
 
+  // Compute a masked copy of values that excludes fields that are hidden, so
+  // dependent fields don't evaluate against stale values. We do NOT mutate the
+  // form state to avoid toggling the modified flag.
+  const maskedValues = React.useMemo(() => {
+    const schema = document.schema;
+    if (!schema) return fieldValues;
+
+    const hidden = new Set<string>();
+    // First pass: determine hidden fields at the top level
+    for (const [name, attr] of Object.entries(schema.attributes ?? {})) {
+      const condition = (attr as any)?.conditions?.visible;
+      if (condition) {
+        const isVisible = rulesEngine.evaluate(condition, fieldValues);
+        if (!isVisible) hidden.add(name);
+      }
+    }
+
+    if (hidden.size === 0) return fieldValues;
+
+    const copy: Record<string, unknown> = { ...fieldValues };
+    for (const key of hidden) {
+      if (key in copy) delete (copy as any)[key];
+    }
+    return copy;
+  }, [document.schema, fieldValues, rulesEngine]);
+
   return (
     <Flex direction="column" alignItems="stretch" gap={6}>
       {layout.map((panel, index) => {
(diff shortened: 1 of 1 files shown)


sas1sf3:
**Proposed Enhancement:** Configurable Behavior for Conditional Fields

It might be beneficial to make the behavior of conditional fields configurable rather than choosing one fixed approach (either clearing or retaining hidden field values).

Different projects handle conditional visibility differently depending on their business logic. In some cases, automatically clearing hidden field values prevents inconsistent state and avoids confusion in dependent conditions. In others, retaining values is preferable—for example, when a temporary hidden state shouldn’t remove entered data that may become relevant again if conditions change back.

A potential approach could be to introduce a configuration option in the field settings, such as:

```json
"conditions": {
  "visible": { "==": [ { "var": "type" }, "internal-link" ] },
  "clearOnHide": true
}

Bassel17:
hello this is the expected behavior from conditional fields, they do retain their values, and the solution would not be to clear it.
however I will bring this up with product so we can re-evaluate how conditionals fields and the affected fields need to may need to behave depending on the scenario

PR #24133, opened 2025-08-08
[generateAccessToken] implement session management 

jhoward1994:
### What does it do?

- Adds generateAccessToken

### Why is it needed?

- To generate new access tokens

### How to test it?

- Unit tests:

```bash
yarn test:unit
```

- API integration tests:

```bash
yarn test:api --db=sqlite
```

### Related issue(s)/PR(s)

DX-2143

### DIFF
--- a/packages/core/core/src/services/__tests__/session-manager.test.ts
+++ b/packages/core/core/src/services/__tests__/session-manager.test.ts
@@ -361,6 +361,38 @@ describe('SessionManager Factory', () => {
   });
 
   describe('generateAccessToken', () => {
+    it('should use correct algorithm and expiration time for access token', async () => {
+      const refreshToken = 'valid-refresh-token';
+      const userId = 'user123';
+      const sessionId = 'session456';
+      // Mock validateRefreshToken to return success
+      const mockValidateResult = {
+        isValid: true,
+        userId,
+        sessionId,
+      };
+      sessionManager.validateRefreshToken = jest.fn().mockResolvedValue(mockValidateResult);
+
+      mockJwt.sign.mockReturnValue('access-jwt-token' as any);
+      await sessionManager.generateAccessToken(refreshToken);
+
+      // Verify the algorithm is explicitly set to HS256
+      expect(mockJwt.sign).toHaveBeenCalledWith(
+        expect.any(Object),
+        config.jwtSecret,
+        expect.objectContaining({
+          algorithm: 'HS256',
+          expiresIn: config.accessTokenLifespan,
+        })
+      );
+
+      // Verify the expiration time matches the config
+      const signCall = mockJwt.sign.mock.calls[0];
+      const options = signCall[2];
+      expect(options.expiresIn).toBe(config.accessTokenLifespan);
+      expect(options.algorithm).toBe('HS256');
+    });
+
     it('should generate access token for valid refresh token', async () => {
       const refreshToken = 'valid-refresh-token';
       const userId = 'user123';


jhoward1994:
~To be merged after https://github.com/strapi/strapi/pull/24134~

packages/core/core/src/services/session-manager.ts
```diff
@@ -141,6 +148,12 @@ class SessionManager {
     deviceId: string,
     origin: string
   ): Promise<{ token: string; sessionId: string }> {
+    if (typeof this.provider?.deleteExpiredByIdentifier === 'function') {
```
innerdvations: Do we actually need this? We're doing the expired cleanup handling in the next line. It seems like we can remove all of the deleteExpired changes in this PR and make it simpler, unless I'm missing something.

packages/core/core/src/services/session-manager.ts
```diff
@@ -197,13 +210,49 @@ class SessionManager {
...
     } catch (error: any) {
+      // If the token is expired, verify signature ignoring expiration to safely extract payload
+      // and clean up the corresponding session in database.
       if (error instanceof jwt.JsonWebTokenError) {
+        if (error.name === 'TokenExpiredError') {
+          try {
+            const expiredPayload = jwt.verify(token, this.config.jwtSecret, {
+              // Validate signature but ignore exp to retrieve session information
+              ignoreExpiration: true as any,
+            }) as RefreshTokenPayload;
```
innerdvations: Instead of `as` can we add type checking in case `verify` returns a string? (even though we know it shouldn't with how we're using it)

packages/core/core/src/services/__tests__/session-manager.test.ts
```diff
@@ -344,6 +366,75 @@ describe('SessionManager Factory', () => {
     });
   });
 
+  describe('generateAccessToken', () => {
+    it('should generate access token for valid refresh token', async () => {
```
innerdvations, line 396:
```suggestion
    it('should use correct algorithm and expiration time for access token', async () => {
      const refreshToken = 'valid-refresh-token';
      const userId = 'user123';
      const sessionId = 'session456';

      // Mock validateRefreshToken to return success
      const mockValidateResult = {
        isValid: true,
        userId,
        sessionId,
      };

      sessionManager.validateRefreshToken = jest.fn().mockResolvedValue(mockValidateResult);
      mockJwt.sign.mockReturnValue('access-jwt-token');

      await sessionManager.generateAccessToken(refreshToken);

      // Verify the algorithm is explicitly set to HS256
      expect(mockJwt.sign).toHaveBeenCalledWith(
        expect.any(Object),
        config.jwtSecret,
        expect.objectContaining({
          algorithm: 'HS256',
          expiresIn: config.accessTokenLifespan,
        })
      );

      // Verify the expiration time matches the config
      const signCall = mockJwt.sign.mock.calls[0];
      const options = signCall[2];
      expect(options.expiresIn).toBe(config.accessTokenLifespan);
      expect(options.algorithm).toBe('HS256');
    });

    it('should generate access token for valid refresh token', async () => {
```

just an extra test you may or may not want

packages/core/core/src/services/session-manager.ts
```diff
@@ -197,13 +210,49 @@ class SessionManager {
...
+    }
+
+    const payload: Omit<AccessTokenPayload, 'iat' | 'exp'> = {
+      userId: validation.userId!,
+      sessionId: validation.sessionId!,
+      type: 'access',
+    };
+
+    const token = jwt.sign(payload, this.config.jwtSecret, {
+      expiresIn: this.config.accessTokenLifespan,
```
innerdvations, line 266:
```suggestion
      algorithm: 'HS256',
      expiresIn: this.config.accessTokenLifespan,
```

Better if we explicitly specify this everywhere (and then easier to find/replace if we go with a different algorithm before release)

derrickmehaffy: We should make this configurable if possible.

jhoward1994: Done via https://github.com/strapi/strapi/pull/24263

PR #24163, opened 2025-08-12
security(deps): upgrade inquirer from v8 to v12

dargmuesli:
### What does it do?

Upgrade inquirer from v8.2.5 to v12.9.1 (the most recent).

TODO:
- [ ] verify the typing is still accurate, it seems there were some changes

### Why is it needed?

There is a transitive dependency on `tmp` that is removed from inquirer v10 and up. Reading the changelog for the major versions I find no apparent reason to not upgrade to v12 right away though:

- v9: esm only
- v10: cjs added back
- v11: some select visual change at runtime
- v12: `@types/node` removed from package requirements

### How to test it?

Check if CI build including testing is still fine.

### Related issue(s)/PR(s)

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "get-port": "5.1.1",
     "glob": "10.3.10",
     "husky": "8.0.3",
-    "inquirer": "8.2.5",
+    "inquirer": "12.9.1",
     "jest": "29.6.0",
     "jest-circus": "29.6.0",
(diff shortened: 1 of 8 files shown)


innerdvations:
Thank you for the PR! We are in the process of taking care of this in https://github.com/strapi/strapi/pull/25776

PR #24167, opened 2025-08-12
fix: track only the api requests

maccomaccomaccomacco:
### What does it do?
It limits the tracking to the Content api requests

### Why is it needed?
Right now, the tracking system event sends the `didReceiveRequest` event to the admin operations and actions, which is not the purpose of this tracking event.

### How to test it?
With a lot of patience.

### DIFF
--- a/packages/core/core/src/services/metrics/middleware.ts
+++ b/packages/core/core/src/services/metrics/middleware.ts
@@ -10,7 +10,7 @@ function nextResetDate(): number {
   return Date.now() + 24 * 60 * 60 * 1000; // Now + 24 hours.
 }
 
-const createMiddleware = ({ sendEvent }: { sendEvent: Sender }) => {
+const createMiddleware = ({ sendEvent, strapi }: { sendEvent: Sender; strapi: Core.Strapi }) => {
   const state: State = {
     expires: nextResetDate(),
     counter: 0,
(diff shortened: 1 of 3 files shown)


packages/core/core/src/services/metrics/middleware.ts
```diff
@@ -19,7 +19,11 @@ const createMiddleware = ({ sendEvent }: { sendEvent: Sender }) => {
   const middleware: Core.MiddlewareHandler = async (ctx, next) => {
     const { url, method } = ctx.request;
 
-    if (!url.includes('.') && ['GET', 'PUT', 'POST', 'DELETE'].includes(method)) {
+    if (
+      !url.includes('.') &&
+      url.includes('api/') &&
```
innerdvations:
```suggestion
      url.includes(strapi.config.get('api.rest.prefix')) &&
```

innerdvations (approved):
LGTM! Can you and @Mcastres QA it and make sure the correct events are still sending?

Mcastres:
Hey 👋 
I'm not sure to understand the goal of this PR 🤔

maccomaccomaccomacco:
@Mcastres we're currently sending the didReceiveRequest event also from the admin, while this event should only be sent when API requests are made. Hence, I have limited the event logic to be sent only on API requests. That's the purpose of this PR.

PR #24170, opened 2025-08-13
fix: guided tour adds paddings to pages instead of using offset

markkaylor:
### What does it do?

- Removes padding on the invisible anchor for guided tour tooltip
- Uses sideOffset instead 

### Why is it needed?

It causes the whole page to shift down

### How to test it?

It should only be impacting CTB and CM list view page
Ensure both of these pages have no added padding coming from GuidedTourTooltip

### DIFF
--- a/packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
+++ b/packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
@@ -32,7 +32,7 @@ const tours = {
     {
       name: 'Introduction',
       content: (Step) => (
-        <Step.Root side="bottom" withArrow={false}>
+        <Step.Root side="bottom" sideOffset={33} withArrow={false}>
           <Step.Title
             id="tours.contentTypeBuilder.Introduction.title"
             defaultMessage="Content-Type Builder"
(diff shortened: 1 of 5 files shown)


packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -32,7 +32,7 @@ const tours = {
     {
       name: 'Introduction',
       content: (Step) => (
-        <Step.Root side="bottom" withArrow={false}>
+        <Step.Root side="bottom" sideOffset={33} withArrow={false}>
```
remidej, line 35: how do you get to 33?

remidej (commented):
Thanks a lot! I'm making an experimental to QA before I merge

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
markkaylor: the designs have 33px...but now that I think about it maybe it should be 32 

https://github.com/user-attachments/assets/9a53b2f4-36e8-4904-af12-86e6f63aed46

PR #24172, opened 2025-08-13
feat: add getAiToken endpoint

butcherZ:
### What does it do?

new endpoint: GET `/users/me/getAiToken`

### Why is it needed?

This endpoint sends user identifier and the EE license to the AI server,  get the AI JWT and return it to the user

## Configuration Required

  ```- STRAPI_LICENSE=your-EE-license-key ```
 ```- STRAPI_ADMIN_AI_URL=http://localhost:3001```
## How to test
  1. Login to Strapi admin and get admin JWT token
  2. launch [Strapi AI server](https://github.com/strapi/strapi-ai)
  3.  run  
  ```
curl -X GET http://localhost:1337/admin/users/me/ai-token" \
-H "Authorization: Bearer YOUR_JWT_TOKEN
```
  5.   Expected API Response:
  ```{
    "data": {
      "token": "eyJhbGciOiJIUzI1NiIs...",
      "expiresAt": "2025-01-01T12:00:00Z",
      "projectId": "uuid"
    }
  }
```

### Related issue(s)/PR(s)

DX-2129

### DIFF
--- a/packages/core/admin/server/src/controllers/__tests__/authenticated-user.test.ts
+++ b/packages/core/admin/server/src/controllers/__tests__/authenticated-user.test.ts
@@ -4,17 +4,18 @@ import createContext from '../../../../../../../tests/helpers/create-context';
 import authenticatedUserController from '../authenticated-user';
 
 describe('Authenticated User Controller', () => {
+  const ORIGINAL_ENV = process.env;
+
   beforeEach(() => {
     jest.clearAllMocks();
-
-    // Reset environment variables
-    delete process.env.STRAPI_LICENSE;
-    delete process.env.STRAPI_ADMIN_AI_URL;
-    delete process.env.STRAPI_AI_URL;
-    delete process.env.NODE_ENV;
+    process.env = { ...ORIGINAL_ENV }; // fresh copy for each test
 
     // Reset global fetch
     delete (global as any).fetch;
+  });
+
+  afterAll(() => {
+    process.env = ORIGINAL_ENV; // fully restore after suite
   });
 
   describe('getAiToken', () => {
@@ -70,7 +71,6 @@ describe('Authenticated User Controller', () => {
       const defaultResponse = {
         jwt: 'test-jwt-token',
         expiresAt: '2025-01-01T12:00:00Z',
-        projectId: 'test-project-id',
       };
 
       return jest.fn().mockResolvedValueOnce({
@@ -78,7 +78,6 @@ describe('Authenticated User Controller', () => {
         json: jest.fn().mockResolvedValueOnce({ ...defaultResponse, ...responseData }),
       });
     };
-
 
     test('Should return unauthorized when user is not authenticated', async () => {
       const ctx = createMockContext(null);
@@ -93,6 +92,9 @@ describe('Authenticated User Controller', () => {
       const ctx = createMockContext();
       global.strapi = createMockStrapi() as any;
 
+      // Ensure no license is available
+      delete process.env.STRAPI_LICENSE;
+
       await authenticatedUserController.getAiToken(ctx as any);
 
       expect(ctx.badRequest).toHaveBeenCalledWith(
@@ -103,7 +105,10 @@ describe('Authenticated User Controller', () => {
     test('Should return bad request when AI server URL is not configured', async () => {
       const ctx = createMockContext();
       global.strapi = createMockStrapi() as any;
+
       process.env.STRAPI_LICENSE = 'test-license';
+      delete process.env.STRAPI_AI_URL;
+      delete process.env.STRAPI_ADMIN_AI_URL;
 
       await authenticatedUserController.getAiToken(ctx as any);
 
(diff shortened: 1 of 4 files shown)


packages/core/admin/server/src/routes/users.ts
```diff
@@ -23,6 +23,14 @@ export default [
       policies: ['admin::isAuthenticatedAdmin'],
     },
   },
+  {
+    method: 'GET',
+    path: '/users/me/getAiToken',
```
innerdvations:
```suggestion
    path: '/users/me/ai-token',
```

packages/core/admin/shared/contracts/users.ts
```diff
@@ -44,6 +44,25 @@ export declare namespace UpdateMe {
...
+  export interface Request {
+    query: {};
+    body: {};
+  }
+
+  export interface Response {
+    data: {
+      token: string;
+      expiresAt?: string;
+      projectId?: string;
```
innerdvations:
```suggestion
```
I don't think there's a reason to expect projectId (although technically I think it's not considered private, just unneccessary)

packages/core/admin/server/src/controllers/authenticated-user.ts
```diff
@@ -52,4 +55,164 @@ export default {
...
+          'AI server URL not configured. Please set STRAPI_ADMIN_AI_URL or STRAPI_AI_URL environment variable.'
+        );
+      }
+
+      // Get the current user
+      const user = ctx.state.user as AdminUser;
+
+      // Create a secure user identifier
+      // Using a hash of email + id + internal salt for enhanced privacy
+      const internalSalt = strapi.config.get('uuid', 'default-salt');
```
innerdvations: Just a note, but `uuid` here is essentially the strapi `projectId`, not an internal salt.

We shouldn't include email, that actually adds a chance it could be brute-forced out of it, especially since this is hashed and not encrypted. Hashing doesn't really gain us anything in this case.

I think for now we should just use `id` by itself, nothing else, and before release we will run it by Legal to see if it needs to be encrypted (rather than plain text) but afaik the terms of using Strapi AI will include that we transmit the user ids (but no other identifiable info for now)

packages/core/admin/server/src/controllers/authenticated-user.ts
```diff
@@ -52,4 +55,164 @@ export default {
...
+                ? `AI server request failed. Response: ${errorText || 'No response body'}.`
+                : 'AI server request failed.';
+            return ctx.unauthorized(message);
+          }
+          if (response.status === 403) {
+            return ctx.forbidden('License validation failed. Check your EE license.');
+          }
+          if (response.status === 404) {
+            return ctx.notFound(`AI service endpoint not found at ${aiServerUrl}`);
+          }
```
innerdvations, line 167: On all of these (not just these two) we need to review if this is information the server should be telling the user, or if it should be put in the server log and sent to the user as a more generic error.

@maccomaccomaccomacco might have some insight here

packages/core/admin/server/src/controllers/__tests__/authenticated-user.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+describe('Authenticated User Controller', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    // Reset environment variables
+    delete process.env.STRAPI_LICENSE;
+    delete process.env.STRAPI_ADMIN_AI_URL;
+    delete process.env.STRAPI_AI_URL;
+    delete process.env.NODE_ENV;
```
innerdvations: Does this potentially interfere with other tests? Like if I run `NODE_ENV=whatever yarn test:unit` will it reset the NODE_ENV for tests after this one?

I actually don't know the answer to this, but it seems like it would.

innerdvations: You might be able to use something like this:

```typescript
describe('Authenticated User Controller', () => {
  const ORIGINAL_ENV = process.env;

  beforeEach(() => {
    jest.clearAllMocks();
    process.env = { ...ORIGINAL_ENV }; // fresh copy for each test

    // Override only what you need
    delete process.env.STRAPI_LICENSE;
    delete process.env.STRAPI_ADMIN_AI_URL;
    delete process.env.STRAPI_AI_URL;
    delete process.env.NODE_ENV;
  });

  afterAll(() => {
    process.env = ORIGINAL_ENV; // fully restore after suite
  });

  test('example', () => {
    expect(process.env.STRAPI_LICENSE).toBeUndefined();
  });
});
```

But also, I would avoid messing with NODE_ENV at all; it doesn't look like you need it here, and modifying that while running could potentially cause weird effects.

innerdvations (changes requested):

packages/core/admin/server/src/controllers/authenticated-user.ts
Bassel17: I would be in favor of having useful errors, although some info hidden like the server url
something like:
`Your Enterprise license could not be validated with the AI service, Verify that your EE license is active and configured correctly.`
and 
`AI service could not be reached.`

better for us if we receive a screenshot from a user with an error  that gives us some useful hints

Bassel17 (approved):
LGTM!

PR #24175, opened 2025-08-13
feat(CTB): add conditional fields form to custom fields advanced settings

Bassel17:
### What does it do?

- Adds conditional fields form in the CTB to the custom field advanced settings

### Why is it needed?

- Add support for conditional fields in Custom fields

### How to test it?

- Add a custom field in the CTB, you should be able to add a condition from the advanced settings
- check in the content-manager to see if the condition is applied
- the condition should be present in the JSON schema as well


butcherZ (approved):
LGTM! QAed locally, it's wooooorking! 🥳

jhoward1994 (approved):

jhoward1994:
All works for me. Maybe we could cover this in `tests/e2e/tests/content-manager/conditional-fields/enum-conditional-text-field-visibility.spec.ts` ?

Bassel17:
> All works for me. Maybe we could cover this in `tests/e2e/tests/content-manager/conditional-fields/enum-conditional-text-field-visibility.spec.ts` ?

Adding a custom field to the e2e test app, is for some reason not working,
will leave it as is, it's not worth wasting too much time on

PR #23667, opened 2025-06-04
enhancement: prevent edit view unnecessary rerenders

remidej:
### What does it do?

Describe the technical changes you did.

### Why is it needed?

Describe the issue you are solving.

### How to test it?

Provide information about the environment and the path to verify the behaviour.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #24179, opened 2025-08-14
future: sync preview to side editor forms

remidej:
### What does it do?

Adds all the interactions we want to enable live/visual editing on Preview:
- highlighting elements matching the focused input in the preview
- double clicking on a highlight triggers a popover with the input for that field
- editing a field (both in side editor and in popover) reflects the changes in real time in the preview

⚠️ not handling nested fields yet, I'll look at that as part of the stega/encoding task
⚠️ not handling yet inputs that require more space than the popover offers, e.g. markdown
ℹ️ an e2e test suite and some contributor docs will come in separate PRs

### How to test it?

Open any preview page in getstarted, click around, edit stuff

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/Blocker.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/Blocker.tsx
@@ -1,0 +1,15 @@
+// Needs to be wrapped in a component to have access to the form context via a hook.
+
+import { Blocker as BaseBlocker, useForm } from '@strapi/admin/strapi-admin';
+
+/**
+ * Prevents users from leaving the page with unsaved form changes
+ */
+const Blocker = () => {
+  const resetForm = useForm('Blocker', (state) => state.resetForm);
+
+  // We reset the form to the published version to avoid errors like – https://strapi-inc.atlassian.net/browse/CONTENT-2284
+  return <BaseBlocker onProceed={resetForm} />;
+};
+
+export { Blocker };
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -14,6 +14,7 @@ import { type UseDocument } from '../../../hooks/useDocument';
 import { useDocumentContext } from '../../../hooks/useDocumentContext';
 import { useDocumentLayout } from '../../../hooks/useDocumentLayout';
 import { useLazyComponents } from '../../../hooks/useLazyComponents';
+import { usePreviewInputManager } from '../../../preview/hooks/usePreviewInputManager';
 
 import { BlocksInput } from './FormInputs/BlocksInput/BlocksInput';
 import { ComponentInput } from './FormInputs/Component/Input';
@@ -39,7 +40,12 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
  * the complete EditFieldLayout and will handle RBAC conditions and rendering CM specific
  * components such as Blocks / Relations.
  */
-const InputRenderer = ({ visible, hint: providedHint, document, ...props }: InputRendererProps) => {
+const InputRenderer = ({
+  visible,
+  hint: providedHint,
+  document,
+  ...inputProps
+}: InputRendererProps) => {
   const { currentDocumentMeta } = useDocumentContext('DynamicComponent');
   const {
     edit: { components },
@@ -63,6 +69,10 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
 
   const editableFields = idToCheck ? canUpdateFields : canCreateFields;
   const readableFields = idToCheck ? canReadFields : canCreateFields;
+
+  // Everything preview related
+  const previewProps = usePreviewInputManager(inputProps.name);
+  const props = { ...inputProps, ...previewProps };
 
   /**
    * Component fields are always readable and editable,
@@ -103,13 +113,23 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
     const CustomInput = lazyComponentStore[props.attribute.customField];
 
     if (CustomInput) {
-      // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
-      return <CustomInput {...props} {...field} hint={hint} disabled={fieldIsDisabled} />;
+      return (
+        <>
+          <CustomInput
+            {...props}
+            {...field}
+            // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
+            hint={hint}
+            disabled={fieldIsDisabled}
+          />
+        </>
+      );
     }
 
     return (
       <FormInputRenderer
         {...props}
+        {...previewProps}
         hint={hint}
         // @ts-expect-error – this workaround lets us display that the custom field is missing.
         type={props.attribute.customField}
@@ -124,8 +144,16 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
   const addedInputTypes = Object.keys(fields);
   if (!attributeHasCustomFieldProperty(props.attribute) && addedInputTypes.includes(props.type)) {
     const CustomInput = fields[props.type];
-    // @ts-expect-error – TODO: fix this type error in the useLibrary hook.
-    return <CustomInput {...props} hint={hint} disabled={fieldIsDisabled} />;
+    return (
+      <>
+        <CustomInput
+          {...props}
+          // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
+          hint={hint}
+          disabled={fieldIsDisabled}
+        />
+      </>
+    );
   }
 
   /**
@@ -143,7 +171,7 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
           layout={components[props.attribute.component].layout}
           disabled={fieldIsDisabled}
         >
-          {(inputProps) => <InputRenderer {...inputProps} />}
+          {(componentInputProps) => <InputRenderer {...componentInputProps} />}
         </ComponentInput>
       );
     case 'dynamiczone':
@@ -161,6 +189,7 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
       return (
         <FormInputRenderer
           {...props}
+          {...previewProps}
           hint={hint}
           options={props.attribute.enum.map((value) => ({ value }))}
           // @ts-expect-error – Temp workaround so we don't forget custom-fields don't work!
@@ -174,6 +203,7 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
       return (
         <FormInputRenderer
           {...restProps}
+          {...previewProps}
           hint={hint}
           // @ts-expect-error – Temp workaround so we don't forget custom-fields don't work!
           type={props.customField ? 'custom-field' : props.type}
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -114,15 +114,13 @@ const InputRenderer = ({
 
     if (CustomInput) {
       return (
-        <>
-          <CustomInput
-            {...props}
-            {...field}
-            // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
-            hint={hint}
-            disabled={fieldIsDisabled}
-          />
-        </>
+        <CustomInput
+          {...props}
+          {...field}
+          // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
+          hint={hint}
+          disabled={fieldIsDisabled}
+        />
       );
     }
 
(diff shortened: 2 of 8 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/Blocker.tsx
remidej, line 1: I'm extracting this as a component since it's used in the edit view and in the preview page.

And having the Blocker inline using the render prop is bad because it breaks memoization for all Form siblings (see #23667)

packages/core/content-manager/admin/src/preview/utils/getSendMessage.ts
remidej, line 1: this function is from the previous PR, but I'm moving it to a util because it's also used by usePreviewInputManager now, and the Preview page file was getting long

packages/core/content-manager/admin/src/preview/pages/Preview.tsx
```diff
@@ -301,108 +301,107 @@ const PreviewPage = () => {
             return yupSchema.validate(cleanedValues, { abortEarly: false });
           }}
         >
-          {({ resetForm }) => (
```
remidej, line 304: this large diff is just because I removed the render prop

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -103,13 +113,23 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
     const CustomInput = lazyComponentStore[props.attribute.customField];
 
     if (CustomInput) {
-      // @ts-expect-error – TODO: fix this type error in the useLazyComponents hook.
-      return <CustomInput {...props} {...field} hint={hint} disabled={fieldIsDisabled} />;
+      return (
+        <>
```
markkaylor: why the fragment?

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -124,8 +144,16 @@ const InputRenderer = ({ visible, hint: providedHint, document, ...props }: Inpu
   const addedInputTypes = Object.keys(fields);
   if (!attributeHasCustomFieldProperty(props.attribute) && addedInputTypes.includes(props.type)) {
     const CustomInput = fields[props.type];
-    // @ts-expect-error – TODO: fix this type error in the useLibrary hook.
-    return <CustomInput {...props} hint={hint} disabled={fieldIsDisabled} />;
+    return (
+      <>
```
markkaylor: fragment?

packages/core/content-manager/admin/src/preview/hooks/usePreviewInputManager.ts
```diff
@@ -0,0 +1,49 @@
...
+>;
+
+export function usePreviewInputManager(name: string): PreviewInputProps {
+  const iframe = usePreviewContext('usePreviewInputManager', (state) => state.iframeRef, false);
+  const setPopoverField = usePreviewContext(
+    'usePreviewInputManager',
+    (state) => state.setPopoverField,
+    false
+  );
+  const hasInputPopoverParent = useHasInputPopoverParent();
```
markkaylor, line 22: I'm guessing you have bigger plans for this hook? I only see it used here so for the time being it seems unnecessary.

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -199,8 +236,60 @@ const previewScript = (shouldRun = true) => {
...
+        if (field === highlightManager.getFocusedField()) {
+          highlightManager.focusedHighlights.forEach((highlight: HTMLElement) => {
+            highlight.style.outlineColor = 'transparent';
+            highlight.style.outlineWidth = '2px';
+          });
+          highlightManager.focusedHighlights.length = 0;
+          highlightManager.setFocusedField(null);
+        }
+      }
+    };
```
markkaylor: Styling stuff. This is sort of a wall of code that is hard to read. Maybe some spaces could help organize things for the reader? also wondering if the if else if chain is necessary couldn't you just have 3 separate conditions? Did you consider returning early instead of creating new if blocks? like `if (!highlight) return` could help flatten the function a bit. 🤷‍♂️

remidej:
Thanks for reviewing it @markkaylor 🙌 

>If I click the input in the side editor am I supposed to see a highlight in the preview?

Yes you should. Do you still not see it if you refresh the page and focus the input again? I think I've seen issues that were solved by a refresh during dev at some point.

> why we don't just do a scroll into view in the side editor when the user clicks the thing in the preview

That's actually what I built at first: opening a popover if the side editor is closed, and focus+scroll to the right input if it is open. So definitely possible. But when I showed it, the feedback from both @lucasboilly and @Adzouz was that they would prefer to always have popovers. I think I agree with them, because I like how colocated it is with the content, and I think it's more predictable to always have the same behavior. Also the long-term plan is to get rid of the side editor altogether.

>Then there wouldn't need to be any sync...

I don't think I agree with you here. The sync is required for admin-iframe communication. Within the admin, the sync between the popover's input and the side editor is managed by useForm/useField, because they both share the same context, so we get it for free

>Just a bit confused why it's on double click instead of single click

That part is TBD so I'm very keen on feedback and suggestions. I don't have it as a single click because you can have clickable elements (links and buttons) that use Strapi content, and overriding clicks would break that. I like double clicks because they don't break that and they're easy enough (unlike, say, alt+click). But obviously people need to know about that or they'll be stuck. So maybe we can dispatch a "use double click to edit content" notification when users single click a highlight? 🤷

markkaylor:
> Yes you should. Do you still not see it if you refresh the page and focus the input again? I think I've seen issues that were solved by a refresh during dev at some point.

I didn't see it on Friday when I tested but I'll be sure to test that again 👍

packages/core/content-manager/admin/src/preview/hooks/usePreviewInputManager.ts
remidej: No bigger plans actually, I really just need to know if the InputRender is rendered inside of a preview popover.

I like relying on context for it because it avoids having to add a prop on every input component.  Would you see another solution?

I could also export useInputPopoverContext instead and move the `context !== undefined` here. I thought it was neat to export a tailor made hook but I'm happy to change that if you prefer

markkaylor: > I thought it was neat to export a tailor made hook

Sounds good! No strong opinion just caught my eye. This is just fine.

markkaylor:
> That's actually what I built at first: opening a popover if the side editor is closed, and focus+scroll to the right input if it is open. So definitely possible. But when I showed it, the feedback from both @lucasboilly and @Adzouz was that they would prefer to always have popovers. I think I agree with them, because I like how colocated it is with the content, and I think it's more predictable to always have the same behavior. Also the long-term plan is to get rid of the side editor altogether.

Sounds good, what about clicking a thing in the side editor scroll into view in the preview 🤔

remidej:
Thanks for all the feedback @markkaylor. I'm not gonna handle it all here right now because I want to get to the stega part pretty soon, but I'm making note of it.

- Scroll lag: weird I'll check that, we definitely want it to be smooth
- Small bubble when empty: agreed, but I'm not sure how to handle that case. Just remove the highlight? Or give it minimum dimensions?
- Markdown broken: I assumed this would work, I'll check
- Blocks broken: This is expected and will require a complex fix. We'll need to target precisely the leaf node being updated, not the field

Adzouz:
Edit: Oh I didn't see Mark's comments my bad... But yes I also wanted to address this matter. Only thing I'm not fan of about the minimum dimensions is that it will depend on the font-size defined by the frontend + you won't know that this area is linked to a field since it will still be a space with nothing in the page. Maybe always put null?

Original comment: I was playing with the feature and it looks cool but I ran into a special case where the field is empty:
https://www.loom.com/share/104118bb9c3e4f8da49d057c3eecebf5?sid=534f1b6e-d964-4fd6-b5c2-fd2f507768df
If you double click, it will remove completely the field instead of having a "null" default value. It would make it disappear and we won't be able to select it again from the preview view. I have to refresh to put it back again or edit the value from the left side to make it appearing again.

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  const iframeRect = iframeRef.current.getBoundingClientRect();
+
+  return (
+    <>
+      {popoverField && (
+        <Box
+          position={'fixed'}
+          top={iframeRect.top + 'px'}
```
Adzouz: You told me that it's better for performance to use transform instead of positionning. Is it something that can be applied here too?

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -0,0 +1,85 @@
...
+          left={iframeRect.left + 'px'}
+          width={iframeRect.width + 'px'}
+          height={iframeRect.height + 'px'}
+          zIndex={4}
```
Adzouz: What is happening if the user has positioned something in their frontend at a higher zIndex in the page at the same position?

remidej:
@Adzouz thanks for trying it and giving feedback.

>Maybe always put null

I hear you, just 2 things I'm unsure about:
- Is null the right term? It makes sense to us as devs, but this is a content editor feature. Maybe "empty" could work?
- Mutating the DOM content of the preview with our own language feels dangerous to me. Ideally I'd prefer a solution that only involves the overlay so we can be sure we're not breaking anything

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -0,0 +1,85 @@
...
+        />
+      )}
+      <InputPopoverProvider>
+        <Popover.Root
+          open={popoverField != null}
```
Adzouz: Why not using a strict check here?

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -0,0 +1,85 @@
...
+          onOpenChange={(open) => !open && setPopoverField(null)}
+        >
+          <Popover.Trigger>
+            <Box
+              id="popover-trigger"
+              position="fixed"
+              width={popoverField.position.width + 'px'}
```
Adzouz, line 53: Ah maybe my comment with transform applies more on this element

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
remidej: In this case I don't think it can break. The z-index scale only applies within a window. Since the popover is outside of the iframe, in the parent window, I don't think elements inside the iframe could ever get on top of it.

It is a concern for the highlights system though, since those are inside the iframe. That's why I set their z-index as 9999

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
remidej: True 👍 

Although re-reading the code I realize there's an early return higher up if `!popoverField`, so it looks like I can just set open to true here

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -18,10 +19,15 @@ const previewScript = (shouldRun = true) => {
    * ---------------------------------------------------------------------------------------------*/
   const HIGHLIGHT_PADDING = 2; // in pixels
   const HIGHLIGHT_HOVER_COLOR = window.STRAPI_HIGHLIGHT_HOVER_COLOR ?? '#4945ff'; // dark primary500
+  const HIGHLIGHT_ACTIVE_COLOR = window.STRAPI_HIGHLIGHT_ACTIVE_COLOR ?? '#7b79ff'; // dark primary600
```
Adzouz, line 22: Could we use the `darkTheme.colors.primary600` from `import { darkTheme } from '@strapi/design-system';` so if at some point it's updated on the DS side, it will also be here?

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
remidej: See the browser's pixel pipeline: https://web.dev/articles/rendering-performance#the_pixel_pipeline

Transform has much better performance if the element is gonna move, because it only requires a paint once, but after that the repositioning is done in the composite phase. Whereas everything else (except opacity) requires a new paint for every frame, which is a lot more expensive.

In this case this Box is static and will never move. That's because it's only here to cover the iframe while the popover is open, because without it, the popover doesn't detect clicks outside of it, because they happen on the iframe's window. And that would break the click outside to close popover behavior. I'll add a comment to explain that

Adzouz:
> @Adzouz thanks for trying it and giving feedback.
> 
> > Maybe always put null
> 
> I hear you, just 2 things I'm unsure about:
> 
> * Is null the right term? It makes sense to us as devs, but this is a content editor feature. Maybe "empty" could work?
> * Mutating the DOM content of the preview with our own language feels dangerous to me. Ideally I'd prefer a solution that only involves the overlay so we can be sure we're not breaking anything

I have to admit that I don't have a strong opinion on this one but I agree that if the user see null or something else like empty, they will think that there's an issue that could be also there in production but it's not the case it's only in the preview. Maybe we could discuss together to find a better and more suitable solution?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: Good idea, then we could also make it match the light or dark theme depending on what's enabled. I would just need to pass it via the function parameters because I can't refer to any imports in this function directly (because of the function stringify thing)

remidej: I made another ticket for it: https://github.com/strapi/content-squad/issues/70

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
Adzouz: Yeah sorry I wasn't targeting the right element. I wanted to talk about the popover one, the one that moves when scrolling on the page and that is positioned based on the position of the text in the page.

Adzouz (approved):
LGTM, we would still need to find a solution for empty values that causes the frame to disappear but otherwise it looks good thanks a lot 🙏

Issue #24082, opened 2025-08-04
Warn after upgrade to 5.20.0: Schema XXXEntry not found in global registry, creating an any placeholder


hungtcs:
### Node Version

v22.17.1

### NPM/Yarn/PNPM Version

NPM 10.9.2

### Strapi Version

5.20.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

After upgrading to 5.20.0, the console print following warning

```txt
warn: Schema PluginUploadFileDocument not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.204] warn: Schema ShareSocialItemEntry not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.204] warn: Schema ShareContactItemEntry not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.214] warn: Schema HomeSliderBannerEntry not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.214] warn: Schema ShareSeoEntry not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.218] warn: Schema ProductProductVariantEntry not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.219] warn: Schema ApiProductCategoryProductCategoryDocument not found in global registry, creating an any placeholder
[2025-08-04 11:48:26.219] warn: Schema ApiProductProductDocument not found in global registry, creating an any placeholder
```

### Steps to Reproduce

1. Upgrade 5.18.1  to 5.20.0

This warning not displayed in the newly created type, 
only the types created before the upgrade will appear in the warning.

### Expected Behavior

no warn.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Rowan-Paul:
Can confirm this also happens to us after updating from 4.19.1 to 4.20.0, old content types all get that warning every time

nevil2:
Happening on multiple Strapi instances for us too.
Does not seem to cause any problems, other than getting the warning.

oiorain:
👋 Hello, and apologies for the confusion!

As part of an automated issue cleanup, this issue was mistakenly closed. That action was unintentional — we're in the process of reviewing and reopening affected issues like this one.

[...]

PR #24147, opened 2025-08-11
Fix plugin content-api route extensions

jhoward1994:
### What does it do?

- Fixes plugin content-api route extension for `i18n` and `users-permissions` by:
  - Adds `createContentApiRoutes.routes` getter/setter so extensions can safely reference `plugin.routes["content-api"].routes`.
  - Keeps route definitions unchanged while restoring compatibility with existing extension patterns.
  - API test: server still starts when plugins mutate `content-api` routes via extensions.
  - CLI test: `strapi openapi generate` still describes content API routes from `i18n` and `users-permissions` (e.g. `/locales`, `/auth/local`, `/users`).

### Why is it needed?

- Recent changes broke the ability for plugins (and app extensions) to mutate `content-api` routes.

### How to test it?

- API:
  - `yarn test:api tests/api/core/strapi/plugin-routes-extension-bc.test.api.ts`
- CLI:
  - `yarn test:cli tests/cli/tests/strapi/strapi/openapi-generate.test.cli.ts`
- Manual check in a Strapi app:
  - `yarn strapi openapi generate`
  - Verify output contains content API paths:
    - i18n: `/locales`
    - users-permissions: `/auth/local`, `/auth/local/register`, `/auth/{provider}/callback`, `/users`, `/users/{id}`

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/24095

PR #24182, opened 2025-08-14
fix: refactor openAPI log

butcherZ:
### What does it do?

Refactor the warning for openAPI generation
Screenshot 2025-08-14 at 15 01 37

### Why is it needed?

users are scared by warnings :D 

### How to test it?

just run `yarn develop` and you will see the logs

### Related issue(s)/PR(s)

#24082


packages/core/core/src/core-api/routes/validation/utils.ts
```diff
@@ -95,7 +115,16 @@ export const safeSchemaCreation = (id: Internal.UID.Schema, callback: () => z.Zo
...
-    strapi.log.debug(`Schema ${transformedId} successfully created and registered`);
+    // Show completion for user content only
+    if (!isBuiltInSchema) {
+      const fieldCount = Object.keys((schema as any)?._def?.shape || {}).length || 0;
+      const schemaName = transformedId
+        .replace('Document', '')
+        .replace('Entry', '')
+        .replace(/([A-Z])/g, ' $1')
+        .trim();
+      strapi.log.info(`   ✅ "${schemaName}" schema created with ${fieldCount} fields`);
```
innerdvations: I would keep all of these new `info` logs as `debug` too, possibly even lower (`silly`?) since it's weird to see these on every startup. If it only happened once it would make sense, but since it happens every time it's just kind of annoying 😆

packages/core/core/src/core-api/routes/validation/utils.ts
```diff
@@ -24,7 +26,7 @@ export const safeGlobalRegistrySet = (id: Internal.UID.Schema, schema: z.ZodType
 
     if (idMap.has(transformedId)) {
       // Remove existing schema to prevent conflicts
-      strapi.log.debug(`Removing existing schema ${transformedId} from registry`);
+      strapi.log.debug(`Replacing existing schema ${transformedId} in registry`);
```
jhoward1994, line 29: I think the old message was correct ?

packages/core/core/src/core-api/routes/validation/utils.ts
jhoward1994: Makes sense, but maybe we don't need the extra info in the log? Could keep it as ?

strapi.log.debug(`Schema ${schemaName} successfully created and registered`);

butcherZ: yeaaah I'm gonna move it to debug level :D

jhoward1994:
Looks good overall

@innerdvations suggested we get this in the same change as https://github.com/strapi/strapi/pull/24147 - hopefully we can change the target branch without too much trouble ? 🤞🏻

packages/core/core/src/core-api/routes/validation/utils.ts
butcherZ: only a matter of choice xD it removes the existing one then add a new one, so I thought replacing could be more accurate here 😄

Convly (approved):
Why not

packages/core/core/src/core-api/routes/validation/utils.ts
jhoward1994: I see yeah. It's up to you feel free to go with whatever describes it best.

Just one final idea 🫣

Remove this line and L34 becomes something like:
`strapi.log.debug(`${idMap.has(transformedId) ? "Replacing" : "Registering"} schema ${transformedId} in global registry`);`

innerdvations (approved):
Works for me!

@jhoward1994 we can just fast-forward the fix with this PR, hopefully it doesn't cause merge conflicts with yours

jhoward1994 (approved):

PR #24192, opened 2025-08-18
chore: update guided tour tracking events

markkaylor:
### What does it do?

- updates tracking events per the issue description
- adds completedAt to the guided tour state to be saved in local storage
- removes useless assertion in marketplace test causing flaky test

### Why is it needed?

Tracking

### How to test it?

🤷 

### Related issue(s)/PR(s)

resolves https://github.com/strapi/content-squad/issues/33

### DIFF
--- a/packages/core/admin/admin/src/components/GuidedTour/Context.tsx
+++ b/packages/core/admin/admin/src/components/GuidedTour/Context.tsx
@@ -88,7 +88,7 @@ const getCompletedTours = (tours: TourState): ValidTourName[] => {
   ) as ValidTourName[];
 };
 
-const areAllToursComplete = (tours: TourState) => Object.values(tours).every((t) => t.isCompleted);
+const areAllToursCompleted = (tours: TourState) => Object.values(tours).every((t) => t.isCompleted);
 
 function reducer(state: State, action: Action): State {
   return produce(state, (draft) => {
@@ -163,7 +163,7 @@ const GuidedTourContext = ({
   }, [state, setStoredTours]);
 
   // Derive all completed tours from state
-  const currentAllCompletedState = areAllToursComplete(state.tours);
+  const currentAllCompletedState = areAllToursCompleted(state.tours);
   // Store completed state in ref to survive a re-render,
   // when current state changes this will persist and be used for comparison
   const previousAllCompletedStateRef = React.useRef(currentAllCompletedState);


packages/core/admin/admin/src/pages/Marketplace/tests/MarketplacePage.test.tsx
```diff
@@ -34,15 +29,8 @@ const waitForReload = async () => {
...
 describe('Marketplace page - layout', () => {
   it('renders the online layout', async () => {
-    const trackUsage = jest.fn();
-    // @ts-expect-error - mock
-    useTracking.mockImplementationOnce(() => ({ trackUsage }));
-
     const { queryByText, getByRole } = render();
     await waitForReload();
-    // Calls the tracking event
-    expect(trackUsage).toHaveBeenCalledWith('didGoToMarketplace');
```
Adzouz, line 44: Why removing this check of this particular event?

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -154,6 +167,24 @@ const GuidedTourContext = ({
...
+  const completedTours = getCompletedTours(state.tours);
+  React.useEffect(() => {
+    if (
+      completedTours.length === Object.keys(state.tours).length &&
+      state.completedAt &&
+      !hasTrackedCompletion.current
+    ) {
+      const completedAtDate = new Date(state.completedAt);
+      const timeSinceCompletion = Date.now() - completedAtDate.getTime();
+      if (timeSinceCompletion < 2000) {
```
Adzouz: I'm a bit confused by this rule. If the completion has been done more than 2000ms from now we don't send the event? Why exactly?

markkaylor: Yeah that's confusing, I meant to highlight this when I created the PR. I only want to fire this event _once_ and only once the moment the user completes _all_ tours. 

> If the completion has been done more than 2000ms from now we don't send the event?

Yes if the completion happened less than 2s ago (ie _just happened_) ago fire the event. The 2s threshold is there to prohibit duplicate dispatches to the tracking server. 

Looking at it again I hate it and I don't think it's necessary, I mean this should be enough right:
- WHEN completedAt IS NOT null THEN do nothing
- WHEN completedAt changes value THEN fire the event

packages/core/admin/admin/src/pages/Marketplace/tests/MarketplacePage.test.tsx
markkaylor: This caused the tests to fail. I think it's flaky 🤔. I fixed it but then I thought, I'm not sure there is much value in this assertion anyway. Do you? I can add it back if you think it's a valuable check.

Adzouz: I'm ok with removing it. I see that we don't really test tracking events triggering for the moment so it's not a big deal but I don't know if that's something we would like to do at some point.

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: Yes I was more thinking about something simpler maybe, without the need of time calculation. The completedAt could be maybe a value that we store in localStorage in that case?

markkaylor: completedAt will be in local storage, but really I'd like to avoid adding more things to state / local storage. The trick is not allowing it to fire on another render or page refresh. It would be ideal if I could find a way to just call trackUsage in the reducer 🤔

Adzouz: That would make sense, if possible yes, put it where the completedAt is set 👍

markkaylor: I ran this plan by ChatGPT and it raised a good point that the reducer should remain pure and this would technically introduce a side effect. It suggested another approach. I just pushed, lmk what you think.

Adzouz: Not Bad

Adzouz (approved):
I tested and everything looks good! Good job 🚀

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -81,17 +82,23 @@ const getInitialTourState = (tours: Tours) => {
   }, {} as TourState);
 };
 
+const getCompletedTours = (tours: TourState): ValidTourName[] => {
+  return Object.keys(tours).filter(
+    (tourName) => tours[tourName as ValidTourName].isCompleted
+  ) as ValidTourName[];
+};
+
+const areAllToursComplete = (tours: TourState) => Object.values(tours).every((t) => t.isCompleted);
```
remidej: should it be areAllToursCompleted? for naming consistency

packages/core/admin/admin/src/features/Tracking.tsx
```diff
@@ -399,14 +399,15 @@ interface DidSkipGuidedTour {
 interface DidCompleteGuidedTour {
   name: 'didCompleteGuidedTour';
   properties: {
-    name: keyof Tours;
+    name: keyof Tours | 'all';
   };
```
remidej, line 403: I'm a bit confused by this event. Is it meant for the guided tour as a whole, or for each tour?

markkaylor: Both, it dispatches the name of the tour when that tour is finished and 'all' the moment all tours are completed.

remidej: so if I complete the last tour it sends 2 events?

remidej (dismissed):

packages/core/admin/admin/src/features/Tracking.tsx
markkaylor: yes

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
markkaylor: hey sorry I pushed but forgot to commit this change. I jus pushed again, might need another approval 🙏, my bad

remidej (approved):

Issue #21478, opened 2024-09-26
Serious issue, cannot find the Collection route I created in Create API Token


JiveCheng:
- Node.js version: v20.17.0
- NPM version: 10.8.1
- Strapi version: 5.0.1
- Database: sqlite
- Operating system: windows 11
- Is your project Javascript or Typescript: js

### Describe the bug

cannot find the Collection route I created in Create API Token

### Screenshots

![image](https://github.com/user-attachments/assets/ea7037eb-a295-4159-aa0b-b2c1a7bb3a9c)
![image](https://github.com/user-attachments/assets/dd9e7b41-9b76-4e31-abe8-9baffb770752)

### Additional context

[2024-09-26 16:54:26.349] error: Cannot read properties of undefined (reading 'type')
TypeError: Cannot read properties of undefined (reading 'type')
    at filterContentAPI (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\@strapi\core\dist\services\content-api\index.js:15:48)
    at Array.filter (<anonymous>)
    at C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\@strapi\core\dist\services\content-api\index.js:25:10
    at C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\lodash\lodash.js:4967:15
    at baseForOwn (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\lodash\lodash.js:3032:24)
    at C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\lodash\lodash.js:4936:18
    at Function.forEach (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\lodash\lodash.js:9410:14)
    at Object.getRoutesMap (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\@strapi\core\dist\services\content-api\index.js:19:24)
    at Object.getRoutes (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\@strapi\admin\server\src\controllers\content-api.ts:13:47)
    at dispatch (C:\Users\jivec\Projects\buy4esg\goodthings\node_modules\koa-compose\index.js:42:32)
[2024-09-26 16:54:26.352] http: GET /admin/content-api/routes (33 ms) 500

gacek1123:
Having this issue on Strapi v5.10.4, Node v22.14.0

Adzouz:
[...]

PR #24180, opened 2025-08-14
fix: remove circular structure in content-api route

Adzouz:
### What does it do?

When going to API tokens page in settings, we have a 500 error when calling the endpoint /content-api/routes because of a circular structure in the routes objects.
We need to get rid of the `request` and `response` attributes in the data sent to avoid this circular structure.

### Why is it needed?

Routes aren't correctly retrieved in the API tokens creation / edit page.

### How to test it?

- Go to the admin and show the console
- Go to Settings > API Tokens
- Click on Create new API Token
  -> The 500 error while calling /content-api/routes should appear in the console before this fix

### Related issue(s)/PR(s)

Resolves #21478

🚀

PR #24207, opened 2025-08-20
Fix: content api route serialization

jhoward1994:
### What does it do?

- Adds route serialization utilities to ensure content API route objects are safe to JSON-serialize:
  - `packages/core/utils/src/route-serialization.ts`
  - Exported via `packages/core/utils/src/index.ts`
- Applies sanitization where content API routes are exposed to the admin:
  - `packages/core/core/src/services/content-api/index.ts` → `getRoutesMap()` now returns a sanitized routes map
  - `packages/plugins/users-permissions/server/services/users-permissions.js` → `getRoutes()` now returns a sanitized routes map

The sanitization strips `request` and `response` (Zod) properties from route objects to avoid circular/non-serializable data.

### Why is it needed?

Some content API routes are returned to the admin panel (e.g. API Tokens permissions UI, Users & Permissions). Route objects can carry Zod schemas in `request` and `response` which are not JSON-serializable and can cause runtime serialization errors when building the admin UI. Sanitizing these fields ensures the responses sent to the admin are plain JSON.

### How to test it?

- Verify admin endpoints return serializable JSON:
  1. Core admin routes map
     - Perform: `GET /admin/content-api/routes`
     - Expect: 200 with JSON map of routes.
  2. Users & Permissions routes
     - Perform: `GET /users-permissions/permissions/routes`
     - Expect: 200 with JSON map of routes; similarly sanitized.
- UI validation (optional):
  - In Users & Permissions → Roles, ensure permissions/routing data loads correctly.

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/21478


butcherZ (approved):
QAed locally, everything seems to work well, LGTM!

Adzouz:
I made a similar PR to fix the same issue https://github.com/strapi/strapi/pull/24180 but I think you're doing the same thing here so I'm good with your solution :) Only thing is that I'm trying to get rid of lodash as much as possible so do you think you could do it also on your PR? You can get inspired by what I've done on mine and we can sync if you want!

packages/core/utils/src/route-serialization.ts
```diff
@@ -0,0 +1,30 @@
...
+    .filter((route): route is Record<string, unknown> => !!route && typeof route === 'object')
+    .map(sanitizeRouteForSerialization);
+
+export const sanitizeRoutesMapForSerialization = (
+  map: Record<string, unknown[]>
+): Record<string, Record<string, unknown>[]> =>
+  Object.entries(map).reduce(
+    (acc, [key, value]) => ({
+      ...acc,
+      ...(Array.isArray(value) && { [key]: sanitizeRoutesArrayForSerialization(value) }),
```
Convly, line 27: I might misunderstand what this is doing, but won't that function return an empty object if there are no arrays in the input? :thinking:

Convly: Or do we want to only keep arrays?

jhoward1994: Yeah we only want to keep arrays. Where this is used the expected input always has the type Record<string, any[]> 
- packages/core/core/src/services/content-api/index.ts
- packages/plugins/users-permissions/server/services/users-permissions.js

`isArray` just makes sure we're only preserving arrays

innerdvations:
Note: we're following this up with a fix to keep non-array values, to ensure it won't be a breaking change for some weird edge cases

markkaylor:
@jhoward1994 I believe this was merged after the release branch was created and was not released in 5.23.1. Could you confirm and update the milestone 🙏

PR #24211, opened 2025-08-20
fix: wrong count of non draftAndPublish entries

Adzouz:
### What does it do?

The query to count the number of non draftAndPublish entries on the homepage widgets was wrong and was retrieving only 1.

Screenshot 2025-08-20 at 17 06 52

Screenshot 2025-08-20 at 16 57 48

Screenshot 2025-08-20 at 16 57 57

### Why is it needed?

Display the correct count on the homepage widgets.

### How to test it?

- Go to the admin interface
- Go to the Content Manager
- Select a content type that doesn't have draftAndPublish flag enabled (if you don't have one, go the Content Type Builder to create one)
- Create a few entries
- Go back to the homepage
  -> The count in the homepage widgets should be updated accordingly

🚀

### DIFF
--- a/packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
+++ b/packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
@@ -2,7 +2,12 @@ import * as React from 'react';
 
 import { NotificationsProvider } from '@strapi/admin/strapi-admin';
 import { DesignSystemProvider } from '@strapi/design-system';
-import { fireEvent, render as renderRTL, waitForElementToBeRemoved } from '@testing-library/react';
+import {
+  fireEvent,
+  render as renderRTL,
+  waitForElementToBeRemoved,
+  waitFor,
+} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { IntlProvider } from 'react-intl';
 import { QueryClient, QueryClientProvider } from 'react-query';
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -305,8 +305,11 @@ const DonutChartSVG = ({ data }: { data: ChartData[] }) => {
             $textColor="neutral600"
           >
             {formatMessage({
-              id: 'content-manager.widget.chart-entries.title',
-              defaultMessage: 'entries',
+              id:
+                total < 2
+                  ? 'content-manager.widget.chart-entries.title.singular'
+                  : 'content-manager.widget.chart-entries.title',
+              defaultMessage: total < 2 ? 'entry' : 'entries',
```
remidej: You can manage the singular and plural with a single translation, for example: https://github.com/strapi/strapi/blob/f89d3b4448a830a27ac46c5cd9786b9898c9a88d/packages/core/content-manager/admin/src/translations/en.json#L222

I wish it was more readable but it's what we use everywhere

Adzouz: True I was confused at first 😅 But I've done it!

Adzouz:
Sorry I don't know if it will help because it's hard to reproduce flaky tests failing but I pushed something to try to fix it if you can reapprove @remidej 🙏

remidej (approved):

PR #24213, opened 2025-08-20
future(preview): add content source maps service

remidej:
### What does it do?

- Creates a content-source-maps service. It traverses entries to attach field sourcemaps as stega encoded data when possible in the content API. This only runs when the `encode-source-maps` header is passed.
- Adds the logic in our previewScript to detect fields with encoded data, and write that data to the data-attributes so we can show our highlight overlays for them

Known issues:
- not all fields are handled, only the string-based ones. 
- the actual data we encode is quite dumb, improving that will be the next PR

I used 2 references for inspiration:
- Alex's [v4 sourcemaps plugin](https://github.com/strapi/plugin-content-source-map/blob/main/server/src/encoder.ts) (RIP) for the actual encoding logic
- the v4 `strapi-response-format` functionality to see how to transform the content API responses based on a header

### Why is it needed?

This metadata powers our new Live Preview

### How to test it?

The most convenient way is to open the preview page for any entry on getstarted, and scroll down in the preview to the code block. This should let you see the weird invisible characters:
screenshot 2025-08-20 at 18 35 48@2x

### DIFF
--- a/examples/getstarted/config/middlewares.js
+++ b/examples/getstarted/config/middlewares.js
@@ -12,6 +12,8 @@ module.exports = [
         useDefaults: true,
         directives: {
           'frame-src': ["'self'"], // URLs that will be loaded in an iframe (e.g. Content Preview)
+          // Needed to load the `@vercel/stega` lib on the dummy-preview page
+          'script-src': ["'self'", "'unsafe-inline'", 'https://cdn.jsdelivr.net'],
         },
       },
     },
--- a/examples/getstarted/src/admin/app.jsx
+++ b/examples/getstarted/src/admin/app.jsx
@@ -7,19 +7,10 @@ import { registerPreviewRoute } from './preview';
 const config = {
   locales: ['it', 'es', 'en', 'en-GB'],
 };
-const bootstrap = (app) => {
-  app.getPlugin('content-manager').injectComponent('editView', 'right-links', {
-    name: 'PreviewButton',
-    Component: () => (
-      <Button onClick={() => window.alert('Not here, The preview is.')}>Preview</Button>
-    ),
-  });
-};
 
 export default {
   config,
   register: (app) => {
     registerPreviewRoute(app);
   },
-  bootstrap,
 };
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -4,6 +4,7 @@ declare global {
     __strapi_previewCleanup?: () => void;
     STRAPI_HIGHLIGHT_HOVER_COLOR?: string;
     STRAPI_HIGHLIGHT_ACTIVE_COLOR?: string;
+    STRAPI_DISABLE_STEGA_DECODING?: boolean;
   }
 }
 
@@ -21,6 +22,7 @@ const previewScript = (shouldRun = true) => {
   const HIGHLIGHT_HOVER_COLOR = window.STRAPI_HIGHLIGHT_HOVER_COLOR ?? '#4945ff'; // dark primary500
   const HIGHLIGHT_ACTIVE_COLOR = window.STRAPI_HIGHLIGHT_ACTIVE_COLOR ?? '#7b79ff'; // dark primary600
 
+  const DISABLE_STEGA_DECODING = window.STRAPI_DISABLE_STEGA_DECODING ?? false;
   const SOURCE_ATTRIBUTE = 'data-strapi-source';
   const OVERLAY_ID = 'strapi-preview-overlay';
   const INTERNAL_EVENTS = {
@@ -53,6 +55,36 @@ const previewScript = (shouldRun = true) => {
   /* -----------------------------------------------------------------------------------------------
    * Functionality pieces
    * ---------------------------------------------------------------------------------------------*/
+
+  const setupStegaDecoding = async () => {
+    if (DISABLE_STEGA_DECODING) {
+      return;
+    }
+
+    const { vercelStegaDecode: stegaDecode } = await import(
+      // @ts-expect-error it's not a local dependency
+      // eslint-disable-next-line import/no-unresolved
+      'https://cdn.jsdelivr.net/npm/@vercel/stega@0.1.2/+esm'
+    );
+
+    const allElements = document.querySelectorAll('*');
+
+    Array.from(allElements).forEach((element) => {
+      const directTextContent = Array.from(element.childNodes)
+        .filter((node) => node.nodeType === Node.TEXT_NODE)
+        .map((node) => node.textContent || '')
+        .join('');
+
+      if (directTextContent) {
+        try {
+          const result = stegaDecode(directTextContent);
+          if (result) {
+            element.setAttribute(SOURCE_ATTRIBUTE, result.key);
+          }
+        } catch (error) {}
+      }
+    });
+  };
 
   const createOverlaySystem = () => {
     // Clean up before creating a new overlay so we can safely call previewScript multiple times
@@ -121,12 +153,12 @@ const previewScript = (shouldRun = true) => {
 
         // Move hover detection to the underlying element
         const mouseEnterHandler = () => {
-          if (!highlightManager.focusedHighlights.includes(highlight)) {
+          if (!focusedHighlights.includes(highlight)) {
             highlight.style.outlineColor = HIGHLIGHT_HOVER_COLOR;
           }
         };
         const mouseLeaveHandler = () => {
-          if (!highlightManager.focusedHighlights.includes(highlight)) {
+          if (!focusedHighlights.includes(highlight)) {
             highlight.style.outlineColor = 'transparent';
           }
         };
@@ -340,11 +372,13 @@ const previewScript = (shouldRun = true) => {
    * Orchestration
    * ---------------------------------------------------------------------------------------------*/
 
-  const overlay = createOverlaySystem();
-  const highlightManager = createHighlightManager(overlay);
-  const observers = setupObservers(highlightManager);
-  const eventHandlers = setupEventHandlers(highlightManager);
-  createCleanupSystem(overlay, observers, eventHandlers);
+  setupStegaDecoding().then(() => {
+    const overlay = createOverlaySystem();
+    const highlightManager = createHighlightManager(overlay);
+    const observers = setupObservers(highlightManager);
+    const eventHandlers = setupEventHandlers(highlightManager);
+    createCleanupSystem(overlay, observers, eventHandlers);
+  });
 };
 
 export { previewScript };
--- a/packages/core/core/src/core-api/controller/index.ts
+++ b/packages/core/core/src/core-api/controller/index.ts
@@ -32,6 +32,7 @@ function createController({
       return transformResponse(data, meta, {
         contentType,
         useJsonAPIFormat: ctx?.headers?.['strapi-response-format'] === 'v4',
+        encodeSourceMaps: ctx?.query?.encodeSourceMaps === 'true',
       });
     },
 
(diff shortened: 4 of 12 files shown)


examples/getstarted/src/admin/app.jsx
```diff
@@ -7,19 +7,10 @@ import { registerPreviewRoute } from './preview';
...
   locales: ['it', 'es', 'en', 'en-GB'],
 };
-const bootstrap = (app) => {
-  app.getPlugin('content-manager').injectComponent('editView', 'right-links', {
-    name: 'PreviewButton',
-    Component: () => (
-      <Button onClick={() => window.alert('Not here, The preview is.')}>Preview</Button>
-    ),
-  });
-};
```
remidej, line 17: not related to the rest but it annoyed me to have a "preview" button in getstarted that wasn't the actual preview

packages/core/core/src/core-api/controller/index.ts
```diff
@@ -32,6 +32,7 @@ function createController({
       return transformResponse(data, meta, {
         contentType,
         useJsonAPIFormat: ctx?.headers?.['strapi-response-format'] === 'v4',
+        encodeSourceMaps: ctx?.headers?.['encode-source-maps'] === 'true',
```
Marc-Roig: should we prefix the header name with strapi- too?

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,201 @@
...
+  'id',
+  'documentId',
+  'locale',
+  'localizations',
+  'created_by',
+  'updated_by',
+  'created_at',
+  'updated_at',
+  'publishedAt',
+];
```
Marc-Roig, line 41: can we get this list of fields from somewhere else? Are this internal strapi fields? Or the ones not visible in the content manager?

Asking cause its tricky to maintain lists like this over time, and we often forget to update them.

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,201 @@
...
+  'richtext',
+  'uid',
+  'enumeration',
+  'email',
+  'date',
+  'datetime',
+  'time',
+  'timestamp',
+];
+
```
Marc-Roig, line 29: shouldn't we encode all but relations, components & dzs? 
```suggestion
const ENCODABLE_TYPES = [
      "string",
      "text",
      "richtext",
      "integer",
      "blocks",
      "biginteger",
      "float",
      "decimal",
      "date",
      "time",
      "datetime",
      "timestamp",
      "boolean",
      "enumeration",
      "json",
      "media",
      "email",
      "password",
      "uid",
];
```

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,201 @@
...
+      const { entryRootId, entryRootModel, entryData, model } = options;
+
+      if (typeof entryData !== 'object' || entryData === null || entryData === undefined) {
+        return entryData;
+      }
+
+      const encodedData = { ...entryData };
+
+      await Promise.all(
+        Object.keys(entryData).map(async (key) => {
```
Marc-Roig: there is the traverseEntity util you could use to traverse such data, could simplify the logic here

examples/getstarted/config/middlewares.js
```diff
@@ -12,6 +12,8 @@ module.exports = [
         useDefaults: true,
         directives: {
           'frame-src': ["'self'"], // URLs that will be loaded in an iframe (e.g. Content Preview)
+          // Needed to load the `@vercel/stega` lib on the dummy-preview page
+          'script-src': ["'self'", "'unsafe-inline'", 'https://cdn.jsdelivr.net'],
```
Marc-Roig, line 16: could we add ithe cdn.jsdelivr.net  to the defaults so users don't need to add it manually?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -54,6 +56,36 @@ const previewScript = (shouldRun = true) => {
...
 
+  const setupStegaDecoding = async () => {
+    if (DISABLE_STEGA_DECODING) {
+      return;
+    }
+
+    const { vercelStegaDecode: stegaDecode } = await import(
+      // @ts-expect-error it's not a local dependency
+      // eslint-disable-next-line import/no-unresolved
+      'https://cdn.jsdelivr.net/npm/@vercel/stega@0.1.2/+esm'
```
Marc-Roig, line 67: 0 knowledge of available cdn's, but how reliable is this one? 
could we add an optional env var just in case users want to opt out to another one?

Marc-Roig: also, why not making it a local dependency?

packages/core/core/src/services/content-source-maps.ts
remidej: Unfortunately stega encoding only works on strings, so I don't think we can attach metadata on numbers and booleans without changing the types of the response. Unless you have ideas :)

Blocks is intentionally missing because it will require some custom treatment that I'm keeping for later. But I'm probably still missing some. Thanks for your list, I'll check

examples/getstarted/config/middlewares.js
remidej: The CDN will load in the real frontend of the users, so adding it to their Strapi CSP isn't necessary. It's only needed specifically for us in the getstarted app, because we're embedding the admin inside itself, because we don't have a real frontend so we use the dummy-preview for convenience

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: This one and unpkg are the only ones I've heard about frequently. I don't know how to compare them though, the only source I have is [the unpkg marketing page](https://www.jsdelivr.com/unpkg) so slightly biased 😁 

It could be nice to load the unpkg one as a backup if jsdelivr fails, I'll look into it

remidej: >also, why not making it a local dependency?

I wish we could! But all this `previewScript` logic needs to be smuggled into the users frontend. We do this by stringifying the function and passing it via a postMessage event to the iframe. The drawback of this approach is that the script cannot refer to anything outside of the function's scope, so no imports.

packages/core/core/src/services/content-source-maps.ts
remidej: @innerdvations agrees with you :) https://github.com/strapi/strapi/pull/20999/files#diff-fa1aa6b92551dda5664e962046435170c28dac858d50e033eebb60b62035d1d9R5

packages/core/core/src/services/content-source-maps.ts
Marc-Roig: :sad-pepe:

packages/core/core/src/services/content-source-maps.ts
remidej: It's a bit too much of a side quest for this PR imo so I added a tech debt entry to do this: https://www.notion.so/strapi/6619863711f74711a4fa4db73e16c517?v=1268f3598074807eb62b000c96893028&p=25a8f359807480c98c15c7cda8cf79c5&pm=s

Marc-Roig (dismissed):
🚀

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -21,6 +22,7 @@ const previewScript = (shouldRun = true) => {
   const HIGHLIGHT_HOVER_COLOR = window.STRAPI_HIGHLIGHT_HOVER_COLOR ?? '#4945ff'; // dark primary500
   const HIGHLIGHT_ACTIVE_COLOR = window.STRAPI_HIGHLIGHT_ACTIVE_COLOR ?? '#7b79ff'; // dark primary600
 
+  const DISABLE_STEGA_DECODING = window.STRAPI_DISABLE_STEGA_DECODING ?? false;
```
Adzouz, line 25: Just a question, what's happening if you disable it? Would we be able to find and associate the fields in the preview?

remidej: If you disable it then you need to provide the strapi data attributes yourself on your site

Adzouz (approved):
Tested and everything looks good on my side, good job 👏

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,119 @@
...
+  'media',
+  'email',
+  'password',
+  'uid',
+  /**
+   * We cannot modify the response shape, so types that aren't based on string cannot be encoded:
+   * - json: object
+   * - blocks: object, will require a custom implementation in a dedicated PR
+   * - integer, float and decimal: number
+   * - boolean: boolean (believe it or not)
```
markkaylor, line 26: Does that mean these fields will not have live preview?

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,119 @@
...
+      });
+      return res;
+    },
+
+    async encodeEntry(options: {
+      entryRootId: any;
+      entryRootModel: string;
+      entryData: any;
+      model: any;
+    }): Promise<any> {
```
markkaylor: How come all the `any`?

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,119 @@
...
+            data.map((item) => this.encodeSourceMaps({ data: item, contentType }))
+          );
+        }
+
+        if (typeof data !== 'object' || data === null) {
+          return data;
+        }
+
+        const actualRootId = rootId || data.id;
+        const actualRootModel = rootModel || contentType.uid;
```
markkaylor: what is rootId and rootModel? Mainly what does root signify? How do they differ from data.id and contentType.uid?

packages/core/core/src/services/content-source-maps.ts
```diff
@@ -0,0 +1,119 @@
...
+      const { data, contentType, rootId, rootModel } = options;
+
+      try {
+        if (Array.isArray(data)) {
+          return await Promise.all(
+            data.map((item) => this.encodeSourceMaps({ data: item, contentType }))
+          );
+        }
+
+        if (typeof data !== 'object' || data === null) {
```
markkaylor, line 89: how come you only check null here but above for entry data you also checked undefined

markkaylor (commented):
Looks good. I'm sure I missed some config but I just have errors when I access the preview. I added a full access api token

Image

packages/core/core/src/services/content-source-maps.ts
remidej: - blocks -> it will have it after we handle it
- json -> we could make it work too if we want
- numbers and booleans: would only work if the users write the data attribute themselves. Which we could potentially make easier in the future with the client library

packages/core/core/src/services/content-source-maps.ts
remidej: What I really mean in this condition is to return data if it's not an object. I'm checking for null specifically too because of this:
screenshot 2025-08-26 at 09 36 49@2x

packages/core/core/src/services/content-source-maps.ts
remidej: Great catch, they were here to help with the recursive logic but they seem to be dead code since I refactored to use traverseEntity

packages/core/core/src/services/content-source-maps.ts
remidej: I removed a few of them, but for the `data` ones, I don't think we can avoid them, since the data we're receiving [upstream is any already](https://github.com/strapi/strapi/blob/bdc664be21d1c7223c5c71299bd913dd9e6a3a03/packages/core/core/src/core-api/controller/transform.ts#L40)

packages/core/core/src/services/content-source-maps.ts
markkaylor: ohh yeah forgot about that js gotcha 🤦‍♂️

markkaylor:
Tested and got it working, I guess my API token was stale or something.
Screenshot 2025-08-26 at 13 39 11

PR #24216, opened 2025-08-21
enhancement: show nested fields in getstarted's dummy preview

remidej:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -1,14 +1,119 @@
...
+                </Flex>
+              )}
+            </Box>
+          ))}
+        </Flex>
+      </ToggleableContainer>
+    );
+  }
+
+  if (typeof value === 'object' && value !== null) {
```
markkaylor, line 88: idk if it's valuable for you but I see you doing this same check in a lot of places, maybe time for an util?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -114,9 +159,8 @@ const previewScript = (shouldRun = true) => {
   }>;
 
   const createHighlightManager = (overlay: HTMLElement) => {
-    const elements = window.document.querySelectorAll(`[${SOURCE_ATTRIBUTE}]`);
+    const elementToHighlight = new Map<Element, HTMLElement>();
```
markkaylor: should it be plural? `elementsToHighlight`? 


```suggestion
    const elementsToHighlight = new Map<Element, HTMLElement>();
```

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -56,34 +56,79 @@ const previewScript = (shouldRun = true) => {
...
 
-    const { vercelStegaDecode: stegaDecode } = await import(
+    const { vercelStegaDecode: stegaDecode, vercelStegaClean: stegaClean } = await import(
       // @ts-expect-error it's not a local dependency
       // eslint-disable-next-line import/no-unresolved
       'https://cdn.jsdelivr.net/npm/@vercel/stega@0.1.2/+esm'
     );
 
-    const allElements = document.querySelectorAll('*');
+    const processElementForStega = (element: Element) => {
```
markkaylor: Maybe something like: encodeElement or applySteganographyToElement would be more explicit?


```suggestion
    const encodeElement = (element: Element) => {
```

or

```suggestion
    const applySteganographyToElement = (element: Element) => {
```

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -221,53 +299,126 @@ const previewScript = (shouldRun = true) => {
...
     const scrollableElements = new Set<Element | Window>();
     scrollableElements.add(window);
 
-    // Find all scrollable ancestors for all tracked elements
-    highlightManager.elements.forEach((element) => {
-      let parent = element.parentElement;
-      while (parent) {
-        const computedStyle = window.getComputedStyle(parent);
-        const overflow = computedStyle.overflow + computedStyle.overflowX + computedStyle.overflowY;
+    const findScrollableAncestors = () => {
```
markkaylor: Not sure I understand what this function does. I think it's looking for parent elements that are scrollable but I'm not sure why or why we have to add the scroll listeners

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -221,53 +299,126 @@ const previewScript = (shouldRun = true) => {
...
+    // Initial setup of scrollable elements
+    findScrollableAncestors();
 
     return {
       resizeObserver,
+      highlightObserver,
+      stegaObserver,
       updateOnScroll,
       scrollableElements,
+      findScrollableAncestors,
```
markkaylor: It seems kind of strange to have a function setupObservers that returns observers and functions that don't really seem related to the observers but just share a dependency with highlightManager. Maybe they could be outside the setupObservers function and inject their own highglightManager dependency? Could be something I'm missing or overlooking but it caught my eye so just a heads up to re-consider this functions responsibility and ensure it is not mixing concerns.

markkaylor:
I see there is a yarn.lock conflict, I was sure to keep the deps you added when back merging main to develop but let me know if something went wrong 🤔

remidej:
>Are the highlights supposed to work for these values. For example a field inside a component? When recording this video I also hit an error when double clicking a field inside a component in the preview

for now the bindings with the inputs only work for root fields. I'm working on addressing that in another branch. So it's normal that fields inside components break currently

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: This is about positioning the highlights correctly even when the element moves because the user scrolls. If it's a regular page scroll, it's easy enough, but the element can also be inside one or multiple overflow:auto/scroll element parents. So we need to listen to scroll events on them.

The dummy preview, for example, wouldn't work without this, because all the content is rendered inside a scrollable position fixed parent (in order to hide the menu)

This logic is from #24127 but I'll see if I can add some comments

markkaylor: Yeah I remembered the logic was from before and I think you explained it back when we had a call but I just couldn't remember what it was about. Thanks for the explanation.

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: You're (absolutely?) right , I made some tweaks, let me know if it makes more sense now

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -372,12 +525,13 @@ const previewScript = (shouldRun = true) => {
    * Orchestration
    * ---------------------------------------------------------------------------------------------*/
 
-  setupStegaDecoding().then(() => {
+  setupStegaDOMObserver().then((stegaObserver) => {
     const overlay = createOverlaySystem();
     const highlightManager = createHighlightManager(overlay);
-    const observers = setupObservers(highlightManager);
+    const observers = setupObservers(highlightManager, stegaObserver);
+    const scrollManager = setupScrollManagement(highlightManager);
```
markkaylor, line 532: Nitpick. I'd settle on a naming convention that is consistent createXManager vs setupX. But it's whatever and you can also update this down the road if you agree.

markkaylor (approved):
Nice, thanks for the updates. Looks good to me. Tested. 

I left one optional comment but feel free to ignore or make the update in the next PR

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: I'll merge now because it's hard to juggle with several branches, but I agree. I'll add it to the next PR

Issue #23858, opened 2025-07-02
Ghost/Orphaned Relations: Admin Interface Shows Different Count Than API Response


MihaiSandor:
### Node Version

20.15.1

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

v5.17.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

## **Current Behavior**
- **Admin Interface**: Shows 3 relations in a component
- **API Response**: Returns 4 relations including a "ghost" entry
- **Ghost Entry Behavior**: 
  - Appears in API responses
  - Returns 404 when accessed directly (`/api/features/124`)
  - Cannot be seen or managed in admin interface
  - Cannot be removed through normal admin operations

### **Setup**
1. Create a content type with a relation field to another content type
2. Create several entries in the related content type
3. Link multiple relations in a component

### **Trigger the Bug**
1. Delete a related entry directly from the array
2. Observe admin interface shows X relations
3. Check API response shows X+1 relations

### **Verify the Bug**
```bash
# API shows all relations including ghost
GET /api/pages?populate[component][populate][relations]=*
# Returns: 4 items

# Direct access to ghost relation fails  
GET /api/target-content-type/124
# Returns: 404 Not Found

# Admin interface only shows 3 manageable relations
```

### **API Response (showing ghost relation)**
```json
{
  "data": {
    "component": {
      "relations": [
        {"id": 2, "label": "Valid Entry 1"},
        {"id": 117, "label": "Valid Entry 2"}, 
        {"id": 124, "label": "Ghost Entry"},  // ← Ghost relation
        {"id": 254, "label": "Valid Entry 3"}
      ]
    }
  }
}
```

### **Direct Access to Ghost Entry**
```json
{
  "data": null,
  "error": {
    "status": 404,
    "name": "NotFoundError", 
    "message": "Not Found"
  }
}
```

### Expected Behavior

- Admin interface and API should return the same number of relations
- All relations returned by API should be accessible and manageable
- No orphaned/ghost relations should persist after deletion

### Logs

```shell

```

### Media

<img width="1056" alt="Image" src="https://github.com/user-attachments/assets/cfe9c1fc-9a06-42f2-87b9-21b6725fdc73" />

and here is the API response:

```json
{
    "sticky": {
        "id": 3837,
        "title": "Leg Massager Device",
        "cta": "Request price list",
        "orderCta": "Order now",
        "priceLabel": "Price",
[...]

Issue #23607, opened 2025-05-27
For a relation inside a component, the relation values are not show in the "Published" tab


h2m:
### Node Version

20.19.2

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.13.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When a collection element containing a one-to-many relation in a component is published, the selected relations are not displayed on the "Published" tab

https://github.com/user-attachments/assets/362d1356-6a6d-4a69-85e1-99acd16a5b10

### Steps to Reproduce

Define a component with a one-to-many relation
Use the component in a collection.
Create an entry and relate it with published values of the related collection.
Publish the entry
Switch to the Published tab
see the read-only relation (entries are missing)

### Expected Behavior

all the published entries of the relation from the Draft tab are also shown on the Published tab

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

dimitristhebrink:
Any update on this?

laurenskling:
I can confirm this. Makes debugging https://github.com/strapi/strapi/issues/23158 very hard.

laurenskling:
Found the problem, here: https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/controllers/relations.ts#L461

the component does not have draftAndPublish, so it goes to check if the target collection has draftAndPublish. It does, so we filter by `publishedAt = { $null: true };`. Making published relations not show upin the query.

oiorain:
Hi there! 
I just restested this issue on the version we're shipping today. it's fixed.

Issue #23158, opened 2025-03-15
publishing an entry in a many relation within a component breaks ordering


laurenskling:
### Node Version

20

### NPM/Yarn/PNPM Version

yarn 1

### Strapi Version

5.11.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

https://github.com/laurenskling/strapi-5-tests/tree/feature/publish-relation-moves-to-the-end

### Bug Description

Publishing a selected relation (within a (DZ) component) will not show up in the API response.
Only when I reorder, it starts showing up in the result.
The order of the response does become out of sync with the admin UI. As if that (previously draft) item has a broken index number.

http://0.0.0.0:1337/api/articles/f10kfztylqyjrcsp9lvj3u4j?populate[blocks][on][shared.quote][populate][authors][fields]=*

### Steps to Reproduce

1. have a (DZ) component with a oneToMany relation
2. add relations with some draft items
3. publish that item
4. see that it does not show up in the API response
5. save and refresh because ordering newly selected relations is bugged....
6. move it in the relation ordering
7. see that it shows up, but at the incorrect order..

### Expected Behavior

relations always showing correctly in the correct order

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/7a18068f-c90d-430c-b263-c6414af95521

### Additional information

related to 
https://github.com/strapi/strapi/pull/22834
https://github.com/strapi/strapi/issues/23155

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

laurenskling:
Alexandre asked me to ping @oiorain

laurenskling:
No it behaves differently. Also, we can now test 2 cases in 2 issues, right?

laurenskling:
updated: 
- reordering before saving is no longer broken, removed that step.
- added steps to reproduce that relation shows in API output while removed from the relations in the UI...

laurenskling:
updated to 5.15.1

laurenskling:
I've found the issue.
[...]

PR #24227, opened 2025-08-22
fix: syncing components in draft and publish relations

Bassel17:
### What does it do?

- Introduces a new findComponentParent method.

- checks if a component belongs to a draft and publish content-type before proceeding with syncing and filters those components out.

### Why is it needed?

- check issue #23858

### How to reproduce

On develop you can do the following to reproduce the bug and see the created ghost relation

Create A content type with draft and publish that has a component containing a manyWay relation also with draft and publish

Create a draft entry including a relation in the component (also in draft)
It creates an entry in the join table for the component relation

Publish the component relation
It creates another entry (ghost/orphaned relation) in the join table

Publish the parent entry containing the component with relation
It creates another entry in the join table

### How to test the fix?

Experimental for testing: `0.0.0-experimental.0e5c9562e7bee0ab55e14e247b87a12c3b0ecb40`

**Parent and relation with draft and publish**

Use a database viewer to monitor the component <> relation join table (eg components_compo_relation_lnk)

Create A content type with draft and publish that has a component containing a manyWay relation also with draft and publish

Create a draft entry including a relation in the component (also in draft)
It creates an entry in the join table for the component relation

Publish the component relation
It  DOES NOT create another entry (ghost/orphaned relation) in the join table

Publish the parent entry containing the component with relation
It creates another entry in the join table

Check the API response to ensure the relations are returned as expected

**Parent without draft and publish, relation with draft and publish**

Create A content type WITHOUT draft and publish that has a component containing a manyWay relation WITH  a draft and publish content type

Create an entry including a relation in the component in draft
It creates an entry in the join table for the component relation

Publish the relation
It creates another entry in the join table (both entries should have the same id for <compo_name>_id)

see issue #23858

### Related issue(s)/PR(s)

Partial fix (but incomplete) for #23858

Repair tool for existing ghost relations will come in a follow-up

### DIFF
--- a/packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
@@ -108,9 +108,7 @@ const load = async (
             .whereIn(targetColumnName, ids)
             .transacting(trx);
 
-          // TODO: Reconsider this name and remove the mutable variable
-          let filteredRelations = newVersionsRelations;
-          // Apply custom filtering if provided
+          let versionRelations = newVersionsRelations;
           if (options.shouldPropagateRelation) {
             const relationsToPropagate = [];
             for (const relation of newVersionsRelations) {
@@ -118,14 +116,14 @@ const load = async (
                 relationsToPropagate.push(relation);
               }
             }
-            filteredRelations = relationsToPropagate;
+            versionRelations = relationsToPropagate;
           }
 
-          if (filteredRelations.length > 0) {
+          if (versionRelations.length > 0) {
             // when publishing a draft that doesn't have a published version yet,
             // copy the links to the draft over to the published version
             // when discarding a published version, if no drafts exists
-            const discardToAdd = filteredRelations
+            const discardToAdd = versionRelations
               .filter((relation) => {
                 const matchingOldVersion = oldVersionsRelations.find((oldRelation) => {
                   return oldRelation[sourceColumnName] === relation[sourceColumnName];
@@ -187,23 +185,6 @@ const sync = async (
         const newId = oldEntriesMap[relation[column]];
         return { ...relation, [column]: newId };
       });
-
-      // Delete the old relations BEFORE inserting new ones
-      // This prevents orphaned draft relations while being precise about what we delete
-      if (relations.length > 0) {
-        // Delete the old relations by matching their exact foreign key values
-        for (const relation of relations) {
-          const sourceColumn = joinTable.joinColumn.name;
-          const targetColumn = joinTable.inverseJoinColumn.name;
-
-          await trx(joinTable.name)
-            .where({
-              [sourceColumn]: relation[sourceColumn],
-              [targetColumn]: relation[targetColumn],
-            })
-            .del();
-        }
-      }
 
       // Insert those relations into the join table
       await trx.batchInsert(joinTable.name, newRelations, 1000);


laurenskling:
is there a release I can test?

innerdvations:
We're pushing some fixes to this branch today, and then I can release an experimental to test. We're aiming to have this fix in next week's release, which will fix the bug itself and prevent future syncing issues but will not resolve existing cases.  Then we will follow up with a "repair" tool in the next few weeks that can be run to clean up any previously existing problems.

markkaylor:
Experimental for testing: ~`0.0.0-experimental.ed76b5e1c1d6fb3b4435c438af7398a06b511d6f`~ See PR description

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -8,13 +8,115 @@ interface LoadContext {
   newVersions: { id: string; locale: string }[];
 }
 
+interface JoinTable {
```
alexandrebodin: can't you use the types exported from the DB layer ?

markkaylor:
Tested on the experimental:

https://github.com/user-attachments/assets/18d97b53-0e9b-4281-9b2d-71c3ac2cef5f

packages/core/core/src/services/document-service/utils/unidirectional-relations.ts
```diff
@@ -74,28 +172,33 @@ const load = async (uid: UID.ContentType, { oldVersions, newVersions }: LoadCont
          *    if published version link exists & not draft version link
          *       create link to new draft version
          */
-
         if (!model.options?.draftAndPublish) {
           const ids = newVersions.map((entry) => entry.id);
 
+          // This is the step were we query the join table based on the id of the document
```
jhoward1994, line 103:
```suggestion
          // This is the step where we query the join table based on the id of the document
```

jhoward1994:
I followed the QA scenario and everything worked for me - DB tables were changed in the expected ways

Just to check, the tables I was looking at were:
- `testingprs_cmps` - linking my root content type `testingprs` with its components
- `components_default_prcompos_testingwithincompos_lnk` - linking components (named `prcompos`) to their related content type `testingwithincompos`

remidej:
I tested on an experimental similar to your video @markkaylor. It looks good, I don't have the orphaned relation bug. I also tested the rest content API and get what I'm expecting

Adzouz:
I also tested on my side, everything looks good. Also tried a few other use cases to test it even further and no weird behaviour nor ghost creation in the DB 🙌

jhoward1994 (approved):

laurenskling:
I can still get ghost relations.
Here are the steps (as detailed before in https://github.com/strapi/strapi/issues/23158)

1) I have a DZ with a component that has a many relation inside ("labels")
2) i add the component tot the DZ, and chose a label (doesnt matter how many). I now have 2 entries in the DB, one for draft, one for published.
3) I go to that label, make changes and publish. Now the published version of this label is added as a relation to the draft version of the CT with the DZ. This is wrong. This should never happen.
4) at first, nothing happens. Draft will get the draft, published will get the published. But..
5) when I unpublish, both relations are destroyed. Only the draft is now liked, 1 entry in the DB. So far so good.
6) repeat the first steps. Get 3 entries in the DB. If I now remove the relation in my DZ component in the draft tab and publish this change; the draft relation is removed, but as we had a wrong entry in the DB pointing to the published version, this is now published. So I now have 2 entries in my db. the wrong draft to published and the published to published. As there is no draft to draft, it doesnt show up in the UI. As it is linked as published to published, it shows up in the published API result. 
7) so now we have created a ghost relation that can never be removed again (unless by removing the entire component)

I've just tested as well outside of a DZ. Same issue.

laurenskling:
> Tested on the experimental:
> 
>  Screen.Recording.2025-08-29.at.12.20.24.mov

so what is different between my steps and this video; is that the ghost creation happens when you update/publish the related content. This video only links and removes.

markkaylor:
Hi @laurenskling, thank you for testing and reporting this. I am trying to reproduce but I am not seeing any ghost relations in the database after making updates to the related entry. Looking at your steps I am trying to get to the state of your step 3. To do that I did the following (see video).

- In a dz component containing a manyWay relation
- Add a relation and save the parent (1 entry in the lnk table - draft to draft)
- Publish the relation (no entry in the lnk table)
- Publish the parent (Another entry in the lnk table - published to published)
- Go to the relation make updates, save, and publish (no entries added to the link table)
- Go to the parent, update another field, save and publish (no new entries added to the link table)

Let me know if I’ve misunderstood something. If you could provide a video of you reproducing the issue that might be helpful too 🙏. Thanks again for your help. 


https://github.com/user-attachments/assets/c884e9ba-c0fc-4ff1-a9e8-4516da3a84d4

laurenskling:
I don't know what to say, it happens for me:

https://github.com/user-attachments/assets/b1e4ff82-aff8-4897-bfdc-15d10c0566e0

I'm on `0.0.0-experimental.e5be7949a8aed02e6af3c5a8bb244f5512801ec4`

Okay tried again, I found our difference... You started with a draft on the other side, I started with a published. But if I keep updating stuff, I can still get it happening. This time its triggered from the parent, not from the related label:

https://github.com/user-attachments/assets/49f0806f-dda4-474e-9cd6-3fcbac71f29a

sorry for the bad quality, needs to be under 10mb.

markkaylor:
Thanks @laurenskling for the video. I am still unable to reproduce. 

I did exactly what you did in the video:

- Add a published relation in dz component
- Publish the parent (creates two entries in the join table)
- Go to the published relation, make an update, and publish (no entry created in the join table)
(see video below)

I did this test on a _new_ app created with the experimental `0.0.0-experimental.e5be7949a8aed02e6af3c5a8bb244f5512801ec4`

I also created a new 5.23.1 app reproduced the bug, then upgraded to the experimental and tested again and saw the bug fixed.

https://github.com/user-attachments/assets/45fd1676-cf6d-4d86-b77f-6fc585ca6fb8

laurenskling:
The error is swallowed by:
https://github.com/strapi/strapi/pull/24227/files#diff-57cb9671b4c3d7978edae6d68a95dab020df5c7b806b45155fe5f943df9d1b8fR322

`Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [cmp_id] query: select `entity_id` from `site_site_cmps` where `cmp_id` = ? and `component_type` = ? limit ?`

~sqlite?~

`componentId` seems empty for me.

~I'm running both contentTypes from plugins. Thats the reason?~

Got it! 
The singular name for my component is `site-footer`. In this code: https://github.com/strapi/strapi/pull/24227/files#diff-70761d48190d3eeb77d6ce85d4c16590766cabc8a43160615646fd6fb04f411bR48 it creates `site-footer_id`, while it should be looking for `site_footer_id` The snake casing is automated to the DB, but singular names can be different. So there should be some kind of snakeCase conversion happening here similar to how the DB is created

innerdvations:
Great debugging, thank you so much for your support on this!

This points out a refactor we need to do; the database layer should not be aware of any Strapi logic like components, so we'll need to quickly update this to make the new database methods more generic to clean relations, and move the components outside to the document service middleware. There we need to use the database identifier naming functions from `packages/core/core/src/utils/transform-content-types-to-models.ts` to ensure that this fix works correctly with all possible component names

We'll prioritize that tomorrow and still aim to get this fix in Wednesday!

laurenskling:
Thank you so much

jhoward1994:
The code looks good
I just tested against the PR description again - it seems to be behaving as intended 

The only weird behaviour during my test was:
1. Publish relation within a component from the draft entry
2. Then publish this draft entry
3. The published tab for the entry does not show a link to the published component relation from step 1

I'm not sure that's related to this PR though, what do you think from this video?

https://github.com/user-attachments/assets/f8c0fb26-4c28-49e2-90d6-4d21a43c1c56

markkaylor:
> The code looks good I just tested against the PR description again - it seems to be behaving as intended
> 
> The only weird behaviour during my test was:
> 
> 1. Publish relation within a component from the draft entry
> 2. Then publish this draft entry
> 3. The published tab for the entry does not show a link to the published component relation from step 1
> 
> I'm not sure that's related to this PR though, what do you think from this video?
> 
>  Screen.Recording.2025-09-02.at.12.10.21.mov

Thanks for testing again. Yes it looks correct for parent that has draft and publish. The relation not showing in the published tab is not caused by the work here, it's another issue also on develop: https://github.com/strapi/strapi/issues/23607 

I think This [commit](https://github.com/strapi/strapi/pull/24227/commits/928bc0088a1289976efdc6b5b43687057d28beff) introduced a new issue when the parent does NOT have draft and publish. I saw when QAing with Marion. It no longer creates a new entry in the join table when the component relation is published. Looking into it. See below:


https://github.com/user-attachments/assets/033c418b-f152-4bbc-a492-48a95de99cc6

markkaylor:
I believe it's working now. New experimental with this fix: 0.0.0-experimental.0e5c9562e7bee0ab55e14e247b87a12c3b0ecb40


https://github.com/user-attachments/assets/bb039e41-35a1-4664-b366-dfb0af4cc203

markkaylor:
@jhoward1994 would you be able to QA again on this experimental: `0.0.0-experimental.0e5c9562e7bee0ab55e14e247b87a12c3b0ecb40`. I updated the PR testing instructions to include a manual test for non-dp parents 🙏 🙏 🙏

jhoward1994:
Thanks, I tested with a D&P and non-D&P content type using that experimental. 

Assuming whats written in *"Parent without draft and publish, relation with draft and publish"* is the expected fixed behaviour (?) it behaves as you described

https://github.com/user-attachments/assets/0c61e89c-34e8-4e02-bb0b-82e0e93f5232

markkaylor:
> Thanks, I tested with a D&P and non-D&P content type using that experimental.
> 
> Assuming whats written in _"Parent without draft and publish, relation with draft and publish"_ is the expected fixed behaviour (?) it behaves as you described
> 
>  Screen.Recording.2025-09-02.at.15.44.31.mov

Awesome thanks! It looks good 👍 ! Sorry I put the testing steps in the wrong part of the description.

I'll also wait for @MarionLemaire QA before merging, then we should be safe to merge and release.

MarionLemaire:
Tested on latest experimental with D&P and no-D&P collection types : LGTM, well done !

laurenskling:
> @jhoward1994 would you be able to QA again on this experimental: `0.0.0-experimental.0e5c9562e7bee0ab55e14e247b87a12c3b0ecb40`. I updated the PR testing instructions to include a manual test for non-dp parents 🙏 🙏 🙏

Ran this experimental on my codebase and my issues seem gone too. Thank you so much

laurenskling:
There is one behaviour that has changed in this PR. 
I know we are discussing at length at https://github.com/strapi/strapi/issues/23460 about unidirectional relations and who is the owner of publishing strategies. But the current (5.23.1) workflow is:
1) publishing entry X with a draft relation.
2) publish that relation.
3) the relation is added to published version of X.

With this PR:
1) publishing entry X with a draft relation.
2) publish that relation.
3) nothing happens to the published version of X. The relation is not added to the published version of X.

markkaylor:
Thanks again @laurenskling for testing and reporting another issue.

I've tested in comparison with latest (5.23.1).

> the relation is added to published version of X.

On 5.23.1, Isn't this the ghost relation we're seeing being created that this PR is trying to prohibit? 
When publishing the related entry  a join table entry is created with a  _draft_ component (relations_id in my case) is  pointing to the _published_ relation (tag_id in my case).

I know it may not be ideal but if the parent is already published and then you publish the related entry then the parent has to be _republished_ to sync the child relation with the parent

This is a video doing the flow you described on 5.23.1.

https://github.com/user-attachments/assets/26466985-af8e-4bb7-befb-aefc097903e9

Same test on this branch showing no ghost relation created and the sync after republishing the parent

https://github.com/user-attachments/assets/ee388087-af58-4b0d-a94a-005cfdc6a835

Let me know if I misunderstood something.

laurenskling:
You are right!
My case of "it worked in 5.23.1" was actually because of the ghost creation.

I expect the following workflow:
1) My colleague prepares some kind of entry
2) I create the page for it and relate it (still in draft), I publish this page. Its ready for me.
3) my colleague finishes the entry, publishes it.
4) I assume it is now live (just like it was in v4, and what ghosts are now doing in v5 as well)
but, v5 wants me to go back to my page and publish it before I can see my relation, even tho its not modified. It's very likely that I have to do this for many many many pages I want this new entry to be published on, but there is no status change to find where. So I'm in the dark about this, I don't know which entries need publishing because they have new relations that can be published.

In https://github.com/strapi/strapi/issues/23460 we discussed this about unpublish/publish again, not showing up. So with this PR merged, it will _always_ not show up. (which has been discussed as something someone could argue that is expected behaviour). I'm afraid now way more v4 users (like me) are going to be confused about their relations not showing up when they press publish on the other side.

markkaylor:
> You are right! My case of "it worked in 5.23.1" was actually because of the ghost creation.

Great, well it sounds like we can merge this was one and prepare a patch release for tomorrow

> 
> I expect the following workflow:
> 
> 1. My colleague prepares some kind of entry
> 2. I create the page for it and relate it (still in draft), I publish this page. Its ready for me.
> 3. my colleague finishes the entry, publishes it.
> 4. I assume it is now live (just like it was in v4, and what ghosts are now doing in v5 as well)
>    but, v5 wants me to go back to my page and publish it before I can see my relation, even tho its not modified. It's very likely that I have to do this for many many many pages I want this new entry to be published on, but there is no status change to find where. So I'm in the dark about this, I don't know which entries need publishing because they have new relations that can be published.
> 
> In #23460 we discussed this about unpublish/publish again, not showing up. So with this PR merged, it will _always_ not show up. (which has been discussed as something someone could argue that is expected behaviour). I'm afraid now way more v4 users (like me) are going to be confused about their relations not showing up when they press publish on the other side.

I definitely understand and empathize with your situation. I completely agree that it is _very_ likely more v4 users are going to experience the same frustration since they expected one behavior in v4 that is fundamentally different in v5.

I'll read through the conversation over there since it has grown quite a bit and try to respond asap. Thanks again for your help.

laurenskling:
Hey friends, I've got some bad news about this. I've found another case we aren't covering.

I was looking into the `findComponentParent` code, and I didn't understand how it should be looping over nested components to find the parent. To realize... it doesn't.
This fix only works for direct components, and components in DZs.
But it's very common to have (repeated) components within components (repeat) within DZ's. 
It's even technically possible to put DZ's in DZ's.
This still creates ghosts.

tl;dr: nested components still make ghosts.

laurenskling:
@Bassel17  Just upped my strapi version against my production db, and it started to run on non-DP contentTypes, instead of just components. Now all these non-DP entries show the relation in the UI (because its linked in draft) but they no longer show up in the production API. This is kinda bad :S

PR #24244, opened 2025-08-26
fix: sanitize routes map for serialization

innerdvations:
### What does it do?

In the parent, we entirely remove `request` and `response`, and also remove any key that doesn't have an array value.

In this PR, we change the approach to:

- allow non-array values to be passed through
- clean ALL objects (rather than removing specific ones)
   - remove all circular
   - max depth
   - remove deep zod schemas that cause issues in the UI even when they aren't circular

### Why is it needed?

less aggressive approach to the parent PR

my main question is: should we actually allow non-array values to be passed through? There are cases in our core code where a key has an `undefined` value, and the parent removes them, but here we keep it as-is. Maybe invalid values should be converted to an empty array, so they're not lost but also won't potentially cause problems down the line?

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
decided to go with original approach, because the only real problem was the new request/response objects that don't need to be there.

We will make one change to it though: keep invalid values as-is to avoid a potential breaking change (even though they probably shouldn't be there)

PR #24246, opened 2025-08-26
Feat: persistent list view settings

mvo-zyres:
### What does it do?

It saves the display headers for the list view table and the filters, sort and pageSize in the local storage.

### Why is it needed?

This feature is needed to save time for the editor whenever they move either from one collection type to another one or edit anything else, the list view display headers, filters, sorting and pageSize are loaded again from before. This saves time for the editor as they don't need to always adjust their settings in the list view again. 

### How to test it?

Adjust the list view settings in the content manager listing page for one collection type. Then move to another page (can also be editing or creating an entry in that content type). And come back to that collection type listing page and the display headers, filters, sort and pageSize should be the same as before. This should not affect the other collection types. Make sure to also check that the other collection types are not affected from this change and that the persistence is only per collection type.


Gouttfi:
Hello,

Reading the description of this feature reminds me of the tool that's already in place when you click the settings button in the document list, then "configure the view."

The settings you adjust in this menu will be retained over time.

image

Did I misunderstand ?

mvo-zyres:
@Gouttfi 
> Hello,
> 
> Reading the description of this feature reminds me of the tool that's already in place when you click the settings button in the document list, then "configure the view."
> 
> The settings you adjust in this menu will be retained over time.
> image
> 
> Did I misunderstand ?

These list view settings are saved for everyone in the database. With our enhancement they can also be saved per device. Meaning if the admin who created and configured the collection type did not configure it as you wanted. You can adjust it on you browser and it will be saved in the local storage. Everytime you open the page, these list view settings will be loaded from the local storage.

derrickmehaffy:
Ideally this should be done on a per user level rather than local storage. It's what I believe we always wanted to do but never had the time and it wasn't a high enough priority

zy-merch:
Hello everyone,
is there an update on the status of this PR ?
Thanks!

Adzouz:
Hey there! Thanks a lot for the proposal 🙏 And also, very sorry for the late answer.
We needed to discuss with the team. We were thinking if it would rather be better to store this info directly in the DB to be able to persist the config to a user and not only a device.
But since it would be another kind of complexity, we agreed to move forward with this solution that would help the UX 😊
I'll run some tests and review it asap and let you know!

zy-merch:
Hello @Adzouz, thanks for the reply! Let us know if anything comes up during the tests and review.

Adzouz:
Hey, I'm sorry again for the delay, I got caught by an important topic but be reassured that your PR is still ongoing. I tested it locally and before taking a look deeper in the code I encountered a weird behaviour:
- When I go to a collection type and select for example documentId (a common field shared across all elements in Strapi) and go to another collection type, then the documentId also appears there.
Is it on purpose? I think it would be better to enable / disable and store the configuration per collection type and not globally. I thought it was the case looking at the local storage but it seems that checking / unchecking a field from a type also update the others.
Can you take a look please? I can also provide a video of testing if needed!
Thanks 🙏

mvo-zyres:
Hey @Adzouz, 

thanks for your response.

I could indeed see the issue. We actually had that issue at some point as I thought it was fixed. Turns out it was not :)

It should be fixed by updating the usePersistentState hook to reinitialize the state on key change. 
The issue was, that the state was still used from the last model. So each time you change the model/collection-type it would apply the display headers from the previous one and save them also to the local storage of the current one.

The usePersistentState as i know is also used in other places apart from the display headers. Though the changes i made should not cause any other issues, it should still be reviewed if this is actually the way to go or if we need a better solution.

Let be know if it now also works on your local!

Adzouz:
Thanks @mvo-zyres, I tested and it's better, thank you! 🙏 One other thing, I tested displaying a field then removing it from the model in the Content type builder after and if I go back, I still can see the field I removed.
Do you think that it would be possible to have a check when loading the fields to see if they exist in the model and remove them if not?
Thanks again 🙏

mvo-zyres:
@Adzouz Yes that is possible. Whenever you change the display headers now (or on initial loading from local storage) they will be checked against the attributes in the schema to remove any non-existent attributes from the display headers.

packages/core/content-manager/admin/src/components/LeftMenu.tsx
```diff
@@ -140,10 +140,6 @@ const LeftMenu = () => {
                     key={link.uid}
                     to={{
                       pathname: link.to,
-                      search: stringify({
```
Adzouz: Why removing this from the url? It could still be useful for sorting and pagination

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -73,21 +75,42 @@ const ListViewPage = () => {
   const { toggleNotification } = useNotification();
   const { _unstableFormatAPIError: formatAPIError } = useAPIErrorHandler(getTranslation);
 
+  usePersistentPartialQueryParams('STRAPI_LIST_VIEW_SETTINGS:', ['sort', 'filters', 'pageSize']);
+
   const { collectionType, model, schema } = useDoc();
   const { list } = useDocumentLayout(model);
 
-  const [displayedHeaders, setDisplayedHeaders] = React.useState<ListFieldLayout[]>([]);
+  const [displayedHeaders, setDisplayedHeaders] = usePersistentState<ListFieldLayout[]>(
```
Adzouz: Here I would suggest to use useScopedPersistentState instead to scope stored data to a Strapi instance. Only thing is that it's not currently exported from the admin package so we would need to add it first in `packages/core/admin/admin/src/index.ts` line 61

packages/core/content-manager/admin/src/components/LeftMenu.tsx
mvo-zyres: i removed it because otherwise every time you click on your collection type you would overwrite the saved entry again in the local storage with the provided parameters which would just make this feature obsolete

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mvo-zyres: good point, i will adjust that, thanks

mvo-zyres:
@Adzouz thank you for your feedback. I exchanged the `usePersistentState` with a `useScopedPersistentState`. I would like to add some tests, but I'm not really familiar with the strapi admin tests. Maybe you or someone else who knows those kind of tests better could help there? :pray:

Adzouz:
Hey @mvo-zyres, I don't know exactly why but I can't push on your branch.
Can you create the following file: `packages/core/content-manager/admin/src/hooks/tests/usePersistentQueryParams.test.ts`
And paste the content: https://pastebin.com/UQn8abTy
Also resolve the conflicts you have on your branch
🙏

mvo-zyres:
Hey @Adzouz, I added the test and the merge conficts should be resolved :)

Adzouz:
Thanks 🙏 Can you also run `yarn prettier:write` please?

mvo-zyres:
@Adzouz yes, I added a commit for formatting the code!

packages/core/content-manager/admin/src/components/LeftMenu.tsx
Adzouz: I figured out that this is not used anymore, we can remove it and all imports / variables needed for this one.

Adzouz: This can be removed too.

Adzouz: This can be removed too.

Adzouz: By removing getPluginsParamsForLink function, useQueryParams is not used anymore, we can remove it.

Adzouz: This can be removed too.

Adzouz: This can be removed too.

mvo-zyres:
Hey @Adzouz dont worry 😅
I just applied all the changes and i also removed the  `import { parse, stringify } from 'qs';` because it was also unused after removing the getPluginsParamsForLink method. 

Hope its now ready for a release (when the tests pass) :pray:

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -73,21 +75,43 @@ const ListViewPage = () => {
...
+    if (!schema?.attributes) return;
+    if (displayedHeaders.length === 0) return;
+    if (schema.uid !== model) return;
+
+    const allowedDisplayHeaders = displayedHeaders.filter((header) =>
+      Object.keys(schema?.attributes).includes(header.name)
+    );
+
+    if (allowedDisplayHeaders.length !== displayedHeaders.length) {
+      setDisplayedHeaders(allowedDisplayHeaders);
```
HichamELBSI: Seems like you are saving the whole attribute array which could be simplified by only saving the attribute names you only need. For each collection type, we would like to save an ordered array of attribute names such as `["id", "name", "description", "created_at"]`.

Here is an example of what you are saving in the localstorage ⬇️  (searchable, label, sortable and attribute object are useless for this feature)
Screenshot 2025-10-15 at 17 50 44

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -73,21 +75,43 @@ const ListViewPage = () => {
...
+    }
+  }, [displayedHeaders]);
```
HichamELBSI: Probably related to the dependency arrays of both useEffect, there is a bug when you switch between two list views where one list view has empty display fields.

Step to reproduce:
- Create 2 collection types
- Open the content manager list view of the first collection
- Add many persist a custom list view settings for the list (add as many as you can)
- Open the second collection list view and empty the list view settings
- Now navigate between the two collection and you should see the first one reset the preferences at some point. 

Here is a screen recording about it: https://www.loom.com/share/b4656b0f596f4564aebf01c4dd49fae1

HichamELBSI: Another bug is:
- You empty the preferences of a list view
- Open the edit view of an entry
- Go back to the list view, the preferences has been reset

packages/core/admin/admin/src/hooks/usePersistentState.ts
```diff
@@ -1,24 +1,36 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import { useInitQuery } from '../services/admin';
 
+function readFromStorage<T>(key: string, fallback: T): T {
+  const raw = window.localStorage.getItem(key);
+  if (raw === null) return fallback;
```
HichamELBSI, line 7: There is a bug related to the empty list view settings:
- Empty the preferences of a first list view
- Navigate between two collection type
- You will see some flickering in the list headers and if you check the local storage, you can see that the old data is set before an empty array

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mvo-zyres: Generally i would not say that this is simplification, this makes the code less readable, less maintainable and more complex. I do however agree that its not optimal to store all of the information in the local storage!

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mvo-zyres: That was because, if the list was `empty` then it got reset. 

I now changed it to be `string[] | null`. The default value is now `null`. 
- If the value is `null` then it **will** get reset. 
- If its an `empty` list then it **will not** get reset.

packages/core/admin/admin/src/hooks/usePersistentState.ts
mvo-zyres: This can be fixed by remounting the ListViewPage on slug change. I changed that, but if its not the right approach here we can always revert the change :)

packages/core/admin/admin/src/hooks/usePersistentState.ts
```diff
@@ -1,24 +1,36 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import { useInitQuery } from '../services/admin';
 
+function readFromStorage<T>(key: string, fallback: T): T {
```
soupette, line 5: I wonder if this [implementation](https://usehooks-ts.com/react-hook/use-local-storage) could be better this way it's also persisted accros tabs without reloading the app

HichamELBSI: This could definitely be an option ❤️

mvo-zyres:
Hey @HichamELBSI, i adjusted most of the addressed bugs. Would be great if you can have another look to review the changes :pray:

packages/core/admin/admin/src/hooks/usePersistentState.ts
Mcastres: I prefer this [implementation](https://www.youtube.com/watch?v=dQw4w9WgXcQ) 🤔

HichamELBSI (approved):
It looks ok to me. Can you please squash your commits and update the branch.

mvo-zyres:
Great, thank you very much @HichamELBSI!
I squashed my commits and the branch is also up to date!

Adzouz:
There's an error in `packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx` if you could address it please 🤕 You can run `yarn prettier:check` to have more details 🙏

mvo-zyres:
@Adzouz Should be fixed!

Adzouz:
Sorry about the delay @mvo-zyres, can you please rebase / merge again to sync with develop? That might fix the pipelines 🙏

mvo-zyres:
Hey @Adzouz, the branch is now up to date again with the develop branch. Hope this fixes the pipelines!

Adzouz:
I've ran the tests locally and I think I understand the problem. Since you removed the parameters from the url in the sidenav, when we switch to another locale on the list view, create an entry and click on the content type again to go back on the list, it's not remembering the locale selected which is wrong...

Here's a video of the issue:

https://github.com/user-attachments/assets/00aa0f23-46cf-4d25-b630-00c4990be03d

When I go back on Article through the side nav, I should have the language that I selected before (here Spanish).
Hope that's clear, don't hesitate and sorry for the back and forth 😢

mvo-zyres:
@Adzouz Good catch, could we fix it by also persisting the locale in the locale storage by just appending the 'locale' to the usePersistentPartialQueryParams hook params?

Adzouz:
Hey @mvo-zyres yes I think that would be a good idea! In order to make it consistent. Today, if we switch to another locale and click on another content-type, it's keeping the chosen locale and pass it through a URL parameter but I guess we could do that through the local storage. The only concern I have is that we need to check that it's still possible to override the element through the URL params. We need to make sure that even if the local storage can't store it, it's still possible to switch from a locale to another. What's your opinion @HichamELBSI?

HichamELBSI:
> Hey @mvo-zyres yes I think that would be a good idea! In order to make it consistent. Today, if we switch to another locale and click on another content-type, it's keeping the chosen locale and pass it through a URL parameter but I guess we could do that through the local storage. The only concern I have is that we need to check that it's still possible to override the element through the URL params. We need to make sure that even if the local storage can't store it, it's still possible to switch from a locale to another. What's your opinion @HichamELBSI?

Definitely, let's do that 👍🏽

mvo-zyres:
Hey @Adzouz & @HichamELBSI. I adjusted the code to also save the locale (content type independent) and fix the overrides of the query params with the URL by prioritizing the URL query params over the saved entries in the local storage. Would be great if you can have a look again!

Adzouz:
Hey @mvo-zyres thank you for the update 🙏 It's better and working as expected on the list view but there's also the locale switcher in the edit view we need to consider (see video below)

https://github.com/user-attachments/assets/37323a4c-4bf9-46c2-9858-e28cdac42473

mvo-zyres:
@Adzouz yes, i put the usePersistentPartialQueryParams for the locale also in the EditViewPage, the only thing that will break it is the back button that just overrides the locale again.

Adzouz:
That's because the back url is taking the previous url in the history (which had a selected locale, let's say EN) but if you change for ES in the edit view, the previous URL stil is EN. You worked on the back button, do you have any idea @HichamELBSI?

Adzouz:
Since it's not really blocking the feature, I would move forward with this. The only thing is that the tests are still failing. Since a lot of logic has changed, it's also impacting a lot our tests unfortunately 😕 Could you take a look and run the tests locally to fix them? I really would like to merge it soon

mvo-zyres:
Hey @Adzouz, sorry for the delay in the response. I just checked the tests and I'm not seeing any tests failing that are related to the changes of this PR. If there are tests related to the changes of this PR, could you let me know which?

mvo-zyres:
Hey @Adzouz! Just wanted to gently follow up on this. Could you clarify which of the failing tests are related to the changes of this PR? That would help me focus on the relevant adjustments. Thanks!

Adzouz:
Hi @mvo-zyres, I'll try to take a look today to see what's happening but you can see the details in the checklist below (you can see that e2e tests and unit front tests are failing). I'll let you know once I'm done debugging! Thanks for your patience 🙏

mvo-zyres:
Hey @Adzouz, so I had another look into the tests and it looks like some tests expect `i18n` query params to not persist. Obviously the changes in this PR affect that behavior. From my understanding this is the correct behavior of the application. Do you agree?

Adzouz:
Hi @mvo-zyres, in that case I would say yes. The tests need to be adapted accordingly then. Do you think you would have the time to do it? 🙏

mvo-zyres:
@Adzouz Yes we can do that.

mvo-zyres:
Hey @Adzouz the tests are now adjusted accordingly. Hope that we can finally finish the feature :pray:

Adzouz (approved):
I'm happy to tell you that it's finally over @mvo-zyres 🎉🍾 Sorry for the long run but we wanted to make it right and you did absolutely great, thank you again for the big contribution 🙏

Adzouz:
Ah it seems that you didn't sign the CLA, can you do it please? I enabled the auto-merge so it should automatically be merged after

mvo-zyres:
Hey @Adzouz, so we signed the CLA now, I also updated the branch again since there were some merge conflicts. Should i squash the commits again?

cdreier:
Hi, we received this feature with an update a few days back - and this is really confusing for us. We implemented a plugin to maintain deeplinks to several filter presets, and this new behaviour feels more like a bug.
When clicking on an entity, a short moment we can see the full list, then gets overwritten by an old filter. 

also when the filtered list is loaded, clicking again on the entity, the filter is removed, but not removed from the local-storage, leaving the list view and coming back restores again the old filter....

Is there an option we can disable this automatic list view persistence?

PR #24247, opened 2025-08-26
fix: errors when trying to link design-system locally

Adzouz:
### What does it do?

When we try to link the local version of the design system with the repo, it's breaking the build (typescript errors). This PR is to address those issues and fix them.

### Why is it needed?

So we can be able to work with a local version of the design system.

### How to test it?

_You need to have the https://github.com/strapi/design-system repository locally on your machine_

- Open a terminal

Setup Design System:
- Go to the DS folder
- Run `yarn install` (to be sure all dependencies are installed)
- Run `yarn build` (it will create the dist folder in ./packages/design-system)

- Go to the repo of Strapi folder
- Run `yarn install` (to be sure all dependencies are installed)
- Run `yarn link {FULLPATH_TO_DS_FOLDER}/packages/design-system`
- Run `yarn setup`
  -> It should build without failing

### Related issue(s)/PR(s)

This PR depends on this one on Design System side: https://github.com/strapi/design-system/pull/1955

🚀


packages/core/admin/admin/tests/styled-components.d.ts
remidej, line 1: I think this should be moved up 1 folder (not in tests), since the Theme component imports from it

packages/core/admin/admin/tests/styled-components.d.ts
```diff
@@ -0,0 +1,11 @@
...
+import type { StrapiTheme } from '@strapi/design-system';
+
+declare module 'styled-components' {
+  export interface DefaultTheme extends StrapiTheme {}
+}
+
+export type ThemeConfig = {
+  light: StrapiTheme;
+  dark: StrapiTheme;
+};
```
remidej: you could also put that export in the declare block above if you want to enable `import {ThemeConfig} from 'styled-components'`.

(not sure it's what you'd want, i'm sharing just in case)

packages/core/admin/admin/tests/utils.tsx
```diff
@@ -155,8 +155,8 @@ const Providers = ({ children, initialEntries, storeConfig, permissions = [] }:
                     <LanguageProvider messages={{}}>
                       <Theme
                         themes={{
-                          dark: darkTheme,
-                          light: lightTheme,
+                          dark: darkTheme as StrapiTheme,
+                          light: lightTheme as StrapiTheme,
```
remidej: do you know why casting is needed?

packages/core/content-type-builder/admin/src/components/CustomRadioGroup/Styles.tsx
```diff
@@ -11,7 +11,7 @@ const Wrapper = styled(Flex)`
     user-select: none;
     flex: 1;
 
-    ${inputFocusStyle()}
+    ${inputFocusStyle() as any}
```
remidej, line 14: does this mean the type would need to be fixed in the DS?

remidej (approved):

PR #24252, opened 2025-08-27
chore: migrate to AI SDK v5

innerdvations:
### What does it do?

Migrate to AI SDK v5

### Why is it needed?

We need to launch with v5 so that we do not have to handle it post-launch

### How to test it?

Use the corresponding branch name feat/ai-sdk-v5 on the ai server

### Related issue(s)/PR(s)

DX-2091


innerdvations:
Updated with the latest EE license changes

packages/core/content-type-builder/admin/src/components/AIChat/components/Messages/Message.tsx
```diff
@@ -174,18 +252,24 @@ const AssistantMessage = ({ message }: { message: AssistantMessageType }) => {
...
    * Else, every new streamed character would re-render the entire chat.
    */
-  return useMemo(() => {
-    if (message.role === 'user') {
-      return <UserMessage message={message} />;
-    }
-    return <AssistantMessage message={message} />;
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [message.revisionId]); // Only re-render if message id changes
+  // return useMemo(() => {
```
Bassel17: this is commented, so it's either not needed or we want to add it back ?
the comment and code are not matching right now

packages/core/content-type-builder/admin/src/components/AIChat/lib/transforms/schemas/fromCTB.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
Bassel17: we can probably remove the the ts-nocheck comment here

packages/core/content-type-builder/admin/src/components/AIChat/lib/types/messages.ts
```diff
@@ -24,22 +23,25 @@ export interface TextContent {
...
-  id: string;
-  revisionId?: string;
-  role: 'assistant';
-  schemaChanges: SchemaChange[];
-  contents: MessageContent[];
-  status: Status;
-};
-
-export type Message = UserMessage | AssistantMessage;
+// export type UserMessage = {
```
Bassel17: since we are using UIMessage typings now we probably can remove them instead of keeping those commented

packages/core/content-type-builder/admin/src/components/AIChat/components/Messages/Message.tsx
innerdvations: Oops, I'll clean it up and add a more accurate comment

PR #24253, opened 2025-08-27
[invalidateRefreshToken] implement session management

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/services/__tests__/session-manager.test.ts
+++ b/packages/core/core/src/services/__tests__/session-manager.test.ts
@@ -344,6 +344,24 @@ describe('SessionManager Factory', () => {
     });
   });
 
+  describe('invalidateRefreshToken', () => {
+    it('should delete sessions by origin and userId', async () => {
+      await sessionManager.invalidateRefreshToken('admin', 'user123');
+
+      expect(mockQuery.deleteMany).toHaveBeenCalledWith({
+        where: { userId: 'user123', origin: 'admin' },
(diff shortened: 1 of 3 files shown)


packages/core/core/src/services/session-manager.ts
```diff
@@ -11,6 +11,7 @@ export interface SessionProvider {
   deleteByIdentifier(userId: string): Promise<void>;
   deleteExpiredByIdentifier(userId: string): Promise<void>;
   deleteExpired(): Promise<void>;
+  deleteByCriteria(criteria: { userId: string; origin: string; deviceId?: string }): Promise<void>;
```
innerdvations: Do we still need `deleteByIdentifier?` Couldn't it be fully replaced with this?

And then we could call this simply `deleteBy` or something

packages/core/core/src/services/__tests__/session-manager.test.ts
```diff
@@ -360,6 +360,24 @@ describe('SessionManager Factory', () => {
     });
   });
 
+  describe('invalidateRefreshToken', () => {
+    it('should delete sessions by origin and userId', async () => {
```
innerdvations, line 432: Can we also add tests to ensure it doesn't delete more than it's supposed to?

innerdvations (approved):
LGTM!

PR #24254, opened 2025-08-27
[validateaccesstoken] implement session management

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/core/core/src/services/session-manager.ts
```diff
@@ -180,6 +180,25 @@ class SessionManager {
     return { token, sessionId };
   }
 
+  validateAccessToken(
+    token: string
+  ): { isValid: true; payload: AccessTokenPayload } | { isValid: false; payload: null } {
+    try {
+      const payload = jwt.verify(token, this.config.jwtSecret, {
+        algorithms: [this.config.algorithm ?? 'HS256'],
```
innerdvations:
```suggestion
        algorithms: [this.config.algorithm ?? DEFAULT_ALGORITHM],
```

Just so we don't have a string everywhere

innerdvations (approved):
LGTM!

PR #24272, opened 2025-08-29
chore: label and close issues before 2022

oiorain:
Rinse and repeat: updated script to target issues before 2022!
I will do another PR for before 2023 and 2024 then I will change the method a little. 

We'll change the method for issues post 2024 and create scripts that will run daily for upkeep.

### DIFF
--- a/.github/workflows/close_stale_issues.yml
+++ b/.github/workflows/close_stale_issues.yml
@@ -1,4 +1,4 @@
-name: Close Pre-2021 Issues
+name: Close Pre-2022 Issues
 on:
   workflow_dispatch: # Allows manual triggering from the Github UI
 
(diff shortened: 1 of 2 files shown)


.github/workflows/close_stale_issues.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           # debug-only: true # Dry run to confirm the workflow is working
```
Baldinof, line 25: In case you still want to run in debug, you can add an input to the workflow, see [here](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_dispatchinputs)

So you could choose to enable debug mode when you trigger the workflow

oiorain: oh this it good to know! 
I'm not sure I understand where to pass the argument when triggeting the action though ?

Baldinof: You'll see a checkbox when you trigger the workflow

Capture d’écran 2025-08-29 à 14 58 17

PR #24296, opened 2025-09-03
future(preview): handle nested fields

remidej:
### What does it do?

Adds handling of nested fields to the new Preview. To summarize the changes needed:
- we now encode a lot more data into the stega source maps. We add the model, documentId, kind, locale and field path. They're all the information we need to find exactly where a field comes from in Strapi.
- in order to get the field path, I had to add a new `rawWithIndices` to the `traverseEntity` util. It's the same as the already existing raw property, but it also has information about the index in an array. For example, for a field that's in the 2nd component in a dynamic zone, raw is `dz.field`, rawWithIndices is `dz.1.field`. I'm not modifying `raw` directly because I think it would be a breaking change.
- we now have 2 separate listeners for events coming from the preview:
  - in `pages/Preview.tsx`: the one about the initial feature configuration
  - in `components/InputPopover`: the one about events happening on a specific field. When it receives an event, we already have all the required data (schema, document, etc.) ready to handle it
- added utils to retrieve the schema of a field based on the field path

### Why is it needed?

Can't edit much without nested fields, components and DZs are everywhere

### How to test it?

Open preview for the kitchensink content type. Try the components, repeatable components, and dynamic zone.

### DIFF
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -463,7 +463,10 @@ const previewScript = (shouldRun = true) => {
 
         // Set new focused field and highlight matching elements
         highlightManager.setFocusedField(field);
-        getElementsByPath(field).forEach((element) => {
+        getElementsByPath(field).forEach((element, index) => {
+          if (index === 0) {
+            element.scrollIntoView({ behavior: 'smooth', block: 'center' });
+          }
           const highlight =
             highlightManager.highlights[Array.from(highlightManager.elements).indexOf(element)];
           if (highlight) {


packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -55,6 +56,8 @@ const InputRenderer = ({
     document.schema?.kind === 'collectionType' ? 'collection-types' : 'single-types';
 
   const isInDynamicZone = useDynamicZone('isInDynamicZone', (state) => state.isInDynamicZone);
+  const isInPreviewPopover = useHasInputPopoverParent();
+  const shouldIgnorePermissions = isInDynamicZone || isInPreviewPopover;
```
markkaylor, line 60: I'm a little confused why we ignore permissions if it's dynamic zone?

packages/core/utils/src/traverse-entity.ts
markkaylor, line 1: Surprised this util doesn't have any tests

markkaylor (commented):
The code looks good to me, we discussed it together yesterday too. The addition of the indexes in the traverse util was a bit more involved than I thought it would be. Still really surprised we didn't already have that info available 🤔.

In the side editor if I'm editing a field and I hit enter, it saves (I thought it was cmd + enter? but anyway...), in the popover should we expect the same behavior? If I hit enter currently in the popover nothing happens. 

https://github.com/user-attachments/assets/863ae3bf-ac2b-4421-9b61-be15bb44be29

I cannot edit a null value in the preview using a popover. 

https://github.com/user-attachments/assets/0e606b95-9d89-4fb1-86cc-55206b2e4c0c

We do not see nested components in the preview. Example below is in DZ but same issue with a component at the root level. 

Image

IMO a two-way scroll into view would improve the UX. I click a field in the side editor, scroll into view the field in the preview and show the highlight. I click the field in the preview, scroll into view in the side editor. Maybe technically that is challenging with the user preview. I can't remember if we already discussed this 🤦‍♂️ .

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
remidej: Oh shit that's something I wanted to discuss yesterday. I need to double check how things are on develop but my understanding reading the code was that we already ignored permissions within DZs. Which sounds like a problem!

packages/core/utils/src/traverse-entity.ts
remidej: Indeed, I'll see if I can vibe code some that make sense

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
markkaylor: Yeah I saw that in the diff. It doesn't look like something you introduced but the variable made it more obvious.

remidej:
@markkaylor I looked into the deep population topic. Deep nesting in repeatable components should be fine. Dynamic zones do have a problem when only some of its components are populated

https://github.com/user-attachments/assets/f0ca0418-debc-4db2-af70-87decbd8b56a

remidej:
>I cannot edit a null value in the preview using a popover.

@markkaylor True, it's a gotcha in our dummy-preview, but I don't think it would really impact users with real websites. When some data is null, it's generally not displayed at all.

I think more commonly it would be those 2 cases:

- `field && <p>{field}/</p>`. Nothing on screen for us to detect
- `field ? <p>{field}</p> : <p>no data</p>`. Something on screen, but untied to strapi data so undetectable

packages/core/utils/src/traverse-entity.ts
remidej: I generated a huge suite 😳 seems to work though 😁

markkaylor:
> @markkaylor I looked into the deep population topic. Deep nesting in repeatable components should be fine. Dynamic zones do have a problem when only some of its components are populated
> 
>  screenshot.2025-09-04.at.17.41.13.mp4

Sounds good, so that means users can get the deep populate they just have to do it manually?

remidej:
@markkaylor yes the encode-source-maps thing does not affect population, they'll always need the populate param. If they do provide it, then the recursive encoding logic will work on every level

The one case where there is an issue is when they populate only some of the components in a dynamic zone but not all

markkaylor:
> @markkaylor yes the encode-source-maps thing does not affect population, they'll always need the populate param. If they do provide it, then the recursive encoding logic will work on every level
> 
> The one case where there is an issue is when they populate only some of the components in a dynamic zone but not all

ok that sounds fine then. Let's just make sure all that is clearly documented 👍

remidej:
>In the side editor if I'm editing a field and I hit enter, it saves (I thought it was cmd + enter? but anyway...), in the popover should we expect the same behavior? If I hit enter currently in the popover nothing happens.

The one documented shortcut that should work is cmd+enter enter indeed. It's the one that triggers the document action and should work everywhere

I'm guessing "enter" in some cases triggers a form submission. But I think it's more accidental than intentional

remidej:
>IMO a two-way scroll into view would improve the UX. I click a field in the side editor, scroll into view the field in the preview and show the highlight. I click the field in the preview, scroll into view in the side editor. Maybe technically that is challenging with the user preview. I can't remember if we already discussed this 🤦‍♂️ .

@markkaylor Clicking on preview to open the side editor (when it's open) sounds like what I had initially built. But I got the feedback from Lucas and Adrien that it would be better to always show the popover instead, because it was better to have a consistent and predictable behavior for the user. I'm cool with it because eventually I'd love to have a popover system that's smart enough to get rid of the side editor altogether, and it goes more in that direction.

Scrolling the preview into view when selecting things in the side editor does sound nice really nice. I just added it, it was actually super simple

https://github.com/user-attachments/assets/81598fef-ed24-4fae-b8e7-2f32f777e745

markkaylor:
> > IMO a two-way scroll into view would improve the UX. I click a field in the side editor, scroll into view the field in the preview and show the highlight. I click the field in the preview, scroll into view in the side editor. Maybe technically that is challenging with the user preview. I can't remember if we already discussed this 🤦‍♂️ .
> 
> 
> Scrolling the preview into view when selecting things in the side editor does sound nice really nice. I just added it, it was actually super simple
> 
>  screenshot.2025-09-05.at.10.02.42.mp4

Nice this is awesome! 

> @markkaylor Clicking on preview to open the side editor (when it's open) sounds like what I had initially built. But I got the feedback from Lucas and Adrien that it would be better to always show the popover instead, because it was better to have a consistent and predictable behavior for the user. I'm cool with it because eventually I'd love to have a popover system that's smart enough to get rid of the side editor altogether, and it goes more in that direction.

I meant scrolling into view but using still using the popover. But that could end up being confusing where to edit things, anyway what you've done is already an improvement I think! I'll review and test everything again here in a bit.

markkaylor:
Some feedback that doesn't need to be addressed here:

If you empty the input value of the popover, then click outside to close it, you can't access the popover again:

https://github.com/user-attachments/assets/4565d8e0-c5b5-4e6e-9c5d-368807a43714

Thanks for adding the scroll into view! Just a heads up it doesn't always seem to be working. I tried the same test as you and it didn't work but other fields do seem to work. But this doesn't need to hold up this PR :) 

https://github.com/user-attachments/assets/2a9656d2-25ea-4216-9315-b2c8a49eab20

PR #24297, opened 2025-09-03
chore: label and close actions for stale 2024 q1q2 issues

oiorain:
in both actions: 
- removed permissions to PRs to speed up the process.

For Close stable actions script: 
- bumped to actions/stale@v9 so the default close reason is `not planned` instead.
- updated messages and timeline for closing.


oiorain:
Thank you Emna! This is what you meant, right?

PR #24299, opened 2025-09-03
fix: globe icon displaying on non-localized fields without explicit c…

### LINKED ISSUE #23723: Globe Icon for Localizable Fields Not Reflecting Correct State
### Node Version

22.16.0

### NPM/Yarn/PNPM Version

npm 10.9.2

### Strapi Version

5.11.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

The globe icon intended to indicate whether a field is localizable appears to be malfunctioning. Regardless of a field’s actual localization setting, the **icon always displays as if the field is localizable**. Based on our previous experience using Strapi v4, non-localizable fields were visually distinguished with a globe icon crossed by a red line. This distinction now appears to be missing.

**Impact:**
This issue can cause confusion for editors managing multilingual content, as they cannot easily distinguish which fields support localization.

### Steps to Reproduce

1. Open a component in the Content-Type Builder.
2. Set one field within the component to be localizable (note this field).
3. Set another field in the same component to be non-localizable (note this field as well).
4. Save the model and restart/rebuild Strapi.
5. Open the Content Manager.
6. Add the affected component to an entry (e.g., a page).
7. Observe the globe icons next to both fields.

**Observed Behavior:**
Both fields display the same globe icon, indicating they are localizable—even though one of them is explicitly set as non-localizable.

### Expected Behavior

The UI should clearly differentiate between localizable and non-localizable fields. For example, non-localizable fields should show a globe icon with a red strikethrough or another clear visual indicator, as seen in earlier versions of Strapi.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

Enterprise tier user

### Confirmation Checklist

GauravKumar547:
### What does it do?

Changes the default inheritance behavior for field localization in the i18n plugin. Fields without explicit **pluginOptions.i18n.localized** settings now default to false (non-localized) instead of true.

### Why is it needed?

Fixes confusing UX where globe icons appeared on fields that don't actually support localization. Fields without explicit localization settings were showing globe icons even when the field behaved as non-localized.

### How to test it?

1. Create a localized content type in Content-Type Builder
2. Add fields: one with no change in localization settings, one with localization enabled
3. In Content Manager, verify only the explicitly localized field shows the globe icon

### Related issue(s)/PR(s)

Fixes #23723


mathildeleg:
Hi, thanks for the suggestion! Unfortunately this doesn't fix it all when it comes to certain fields, here are some reproductions where it's still not working:

1. Repeatable or non-repeatable components: 
* Add a repeatable or non-repeatable component to a localized Content Type
* Make sure localization is enabled for this component
* Go to the Content Manager for this Content Type and click on "Create new entry"
* Add a component for that entry, see the Globe icon
* Go back to the Content Type Builder and remove the localization for this component
* Go back to the Content Manager for this Content Type and click on "Create new entry" or go to the one you previously created
* See that the Globe icon is still there for that component
**_Expected Behaviour_**
The Globe icon should no longer be displayed for the fields inside that component, since localization was disabled

2. Using a component more than once in a Content Type
* Add a repeatable or non-repeatable component to a localized Content Type, for example _Button_
* Set this component to non-localized
* Add a Dynamic Zone with that same component to a localized Content Type
* Make sure localization is enabled for this Dynamic Zone
* Go to the Content Manager for this Content Type and click on "Create new entry"
* Add _Button_ and see that the component itself doesn't show the Globe icon, but the fields inside that component do
* Add a Dynamic Zone with a _Button_ and see that the fields inside that component show the Globe icon
=> because the component has the same uid, it displays the same behaviour even if one of its parent element (either repeatable/non repeatable component or dynamic zone) doesn't have localization enabled
**_Expected Behaviour_**
The fields inside the _Button_ component (repeatable or not) shouldn't display the Globe icon

Let me know if you'd like to dive deeper into it and attempt to fix this, otherwise I'll close your PR and we'll try to work on a fix on our end when we can :)

GauravKumar547:
Hi @mathildeleg thanks for the review, I tried replicating but not able to so closing this pr. 
Thanks again.

Issue #24083, opened 2025-08-04
Custom Field becomes unsupported after upgrading to v5.20.0


chengyuen93:
### Node Version

22.14.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.20.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

After upgrading to Strapi v5.20.0, I get this error when starting up the strapi server:

<img width="1370" height="322" alt="Image" src="https://github.com/user-attachments/assets/b37904c8-d12e-4965-9c03-e7ee5df4eda8" />

Suspect that this file didn't handle the type `customField`?
https://github.com/strapi/strapi/blob/v5.20.0/packages/core/core/src/core-api/routes/validation/mappers.ts

### Steps to Reproduce

1. Upgrade the strapi version to 5.20.0
2. run `strapi develop`

### Expected Behavior

Should not throw the `Unsupported attribute type: customField` as customField should still be supported

### Logs

```shell
[ERROR]  There seems to be an unexpected error, try again with --debug for more information 

┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                                                                                                                                │
│   Error: Unsupported attribute type: customField                                                                                                                                               │
│       at mapAttributeToSchema (/Users/ongchengyuen/Documents/Works/PACP/gitlab/pacp-admin-portal-strapi/node_modules/@strapi/core/dist/core-api/routes/validation/mappers.js:140:23)           │
│       at get channel_selector (/Users/ongchengyuen/Documents/Works/PACP/gitlab/pacp-admin-portal-strapi/node_modules/@strapi/core/dist/core-api/routes/validation/mappers.js:45:24)            │
│       at get shape (/Users/ongchengyuen/Documents/Works/PACP/gitlab/pacp-admin-portal-strapi/node_modules/zod/dist/cjs/v4/core/util.js:373:59)                                                 │
│       at Object.extend (/Users/ongchengyuen/Documents/Works/PACP/gitlab/pacp-admin-portal-strapi/node_modules/zod/dist/cjs/v4/core/util.js:371:24)                                             │
│       at inst.extend (/Users/ongchengyuen/Documents/Works/PACP/gitlab/pacp-admin-portal-strapi/node_modules/zod/dist/cjs/v4/classic/schemas.js:594:28)                                         │
[...]

PR #24304, opened 2025-09-03
fix: add schema validation for custom fields to prevent #24083

boazpoolman:
### What does it do?

Fixes the issue as described in #24083 

### How to test it?

Reproduce the issue as described in #24083. Notice that you do not get the error when the changes from this PR are applied.

### Comments

In terms of the schema of a custom field I noticed a custom field can have the following additional attributes:

1. `customField` (required)
2. `required` (optional)
3. `options` (optional)

I noticed the following attributes do not have any effect on custom fields:

1. `writeable`
2. `default`

Taking this in to account, the only schema attribute that says anything about it's validation is the `required` attribute. So that is the only schema validation I've added for custom fields.

Also I've set the field type to `z.any()`, seeing how a custom field can be a variety of types (string, number, json ..)

### DIFF
--- a/packages/core/core/src/core-api/routes/validation/mappers.ts
+++ b/packages/core/core/src/core-api/routes/validation/mappers.ts
@@ -97,6 +97,8 @@ export const mapAttributeToSchema = (attribute: Schema.Attribute.AnyAttribute): 
       return attributes.booleanToSchema(attribute);
     case 'component':
       return attributes.componentToSchema(attribute);
+    case 'customField':
+      return attributes.customFieldToSchema(attribute);
     case 'date':
       return attributes.dateToSchema(attribute);
     case 'datetime':
@@ -190,6 +192,8 @@ export const mapAttributeToInputSchema = (
       return attributes.booleanToInputSchema(attribute);
     case 'component':
       return attributes.componentToInputSchema(attribute);
+    case 'customField':
+      return attributes.customFieldToInputSchema(attribute);
     case 'date':
       return attributes.dateToInputSchema(attribute);
     case 'datetime':
--- a/packages/core/core/src/core-api/routes/validation/attributes.ts
+++ b/packages/core/core/src/core-api/routes/validation/attributes.ts
@@ -97,6 +97,21 @@ export const componentToSchema = (
 };
 
 /**
+ * Converts a customField attribute to a Zod schema.
+ * @param attribute - The CustomField attribute object from the Strapi schema.
+ * @returns A Zod schema representing the custom field.
+ */
+export const customFieldToSchema = (attribute: Schema.Attribute.CustomField): z.Schema => {
+  const { required } = attribute;
+
+  const schema = augmentSchema(z.any(), [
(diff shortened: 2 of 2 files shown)


packages/core/core/src/core-api/routes/validation/mappers.ts
```diff
@@ -97,6 +97,8 @@ export const mapAttributeToSchema = (attribute: Schema.Attribute.AnyAttribute):
       return attributes.booleanToSchema(attribute);
     case 'component':
       return attributes.componentToSchema(attribute);
+    case 'customField':
```
innerdvations, line 100: This causes a tangled mess of typescript errors because customField is only an input attribute type (see [convert-custom-field-type.ts](https://github.com/strapi/strapi/blob/a1da9b829e36a866a425439d22ab6f484b547d9f/packages/core/core/src/utils/convert-custom-field-type.ts#L4)) but isn't (and shouldn't be) in `Attribute.AnyAttribute`

I think the easiest solution is to handle it like this:

```ts
    default: {
      // Handle customField conversion in default case
      if ((attribute as any).type === 'customField') {
        const strapi = (global as any).strapi;
        if (!strapi) {
          throw new Error('Strapi instance not available for custom field conversion');
        }

        const customField = strapi.get('custom-fields').get((attribute as any).customField);
        if (!customField) {
          throw new Error(`Custom field '${(attribute as any).customField}' not found`);
        }

        // Recursively call with converted attribute
        return mapAttributeToInputSchema({
          ...(attribute as any),
          type: customField.type,
        });
      }

      const { type } = attribute as Schema.Attribute.AnyAttribute;

      throw new Error(`Unsupported attribute type: ${type}`);
    }
```

which would handle the conversion here. And ideally change the method signature to include that in the types, although that requires adding a whole bunch of extra type guards that make this much more complex 🫠

I'm not sure exactly how the front-end sends the data for custom fields, but I'm pretty sure it should be mapped to a basic attribute type like this rather than have those `customFieldTo*` validations. Can you confirm @jhoward1994 ?

innerdvations (changes requested):
Types don't build :( see my comment above for a potential fix

packages/core/core/src/core-api/routes/validation/mappers.ts
jhoward1994: Yep that fix makes sense to me. However, it feels like we're duplicating the logic in the core register `convertCustomFieldType` - & we shouldn't ever see a type of custom field in the mappers

Looking at this comment in the original issue https://github.com/strapi/strapi/issues/24083#issuecomment-3246342992

I think this causes the issue because accessing routes within a plugins register function means the mappers are invoked before `convertCustomFieldType` from the Strapi core has run

I think we should add the default case fallback from your comment as plugins will already exist that use this kind of logic but maybe we can also document that logic like that in a plugins register function isn't advised?

boazpoolman:
Closing in favor of https://github.com/strapi/strapi/pull/24310

PR #24307, opened 2025-09-04
Feat/jwt-admin-integration

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -17,6 +17,7 @@ import {
   useLogoutMutation,
   useRenewTokenMutation,
 } from '../services/auth';
+import { getOrCreateDeviceId } from '../utils/deviceId';
 
 import type {
   Permission as PermissionContract,
@@ -170,7 +171,7 @@ const AuthProvider = ({
 
   const login = React.useCallback<AuthContextValue['login']>(
     async ({ rememberMe, ...body }) => {
-      const res = await loginMutation(body);
+      const res = await loginMutation({ ...body, deviceId: getOrCreateDeviceId(), rememberMe });
 
       /**
        * There will always be a `data` key in the response
@@ -193,7 +194,7 @@ const AuthProvider = ({
   );
 
   const logout = React.useCallback(async () => {
-    await logoutMutation();
+    await logoutMutation({ deviceId: getOrCreateDeviceId() });
     clearStateAndLogout();
   }, [clearStateAndLogout, logoutMutation]);
 
--- a/packages/core/admin/server/src/controllers/authentication.ts
+++ b/packages/core/admin/server/src/controllers/authentication.ts
@@ -14,7 +14,6 @@ import {
   validateResetPasswordInput,
   validateRenewTokenInput,
   validateAccessTokenExchangeInput,
-  // Validates optional session-related fields for login requests
   validateLoginOptionalSessionInput,
 } from '../validation/authentication';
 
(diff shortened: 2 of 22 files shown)


packages/core/admin/server/src/controllers/authentication.ts
```diff
@@ -20,13 +24,59 @@ import type {
...
   RenewToken,
   ResetPassword,
+  AccessTokenExchange,
 } from '../../../shared/contracts/authentication';
 import { AdminUser } from '../../../shared/contracts/shared';
 
 const { ApplicationError, ValidationError } = errors;
 
+const isSessionsEnabled = (): boolean =>
+  Boolean(strapi.config.get('admin.auth.sessions.enabled', false));
```
innerdvations: For the admin, it's not a breaking change, so this shouldn't be an option, it should always use the sessions. So instead of the deprecation in renewToken you can just remove it entirely.

packages/core/admin/server/src/routes/authentication.ts
```diff
@@ -10,9 +10,12 @@ export default [
...
   {
     method: 'POST',
-    path: '/renew-token',
-    handler: 'authentication.renewToken',
-    config: { auth: false },
+    path: '/access-token',
+    handler: 'authentication.accessToken',
+    config: {
+      auth: false,
+      middlewares: ['admin::rateLimit'],
```
jhoward1994: Does this endpoint need to be limited?

packages/core/admin/admin/src/features/Auth.tsx
```diff
@@ -122,27 +121,6 @@ const AuthProvider = ({
     navigate('/auth/login');
   }, [dispatch, navigate]);
 
-  /**
-   * Fetch data from storages on mount and store it in our state.
-   * It's not normally stored in session storage unless the user
-   * does click "remember me" when they login. We also need to renew the token.
-   */
-  React.useEffect(() => {
```
jhoward1994, line 130: Token renewal has been moved to packages/core/admin/admin/src/utils/baseQuery.ts

packages/core/admin/admin/src/utils/baseQuery.ts
```diff
@@ -19,56 +20,90 @@ interface UnknownApiError {
...
+
+  try {
+    const result = await executeQuery(query);
+    return { data: result.data };
   } catch (err) {
     // Handle error of type FetchError
 
     if (isFetchError(err)) {
+      // Attempt auto-refresh on 401 then retry once
+      if (err.status === 401) {
```
jhoward1994, line 61: When an access token expires admin requests will receive a 401 - the first time we see this in the admin we should interpret this as access token expiry and attempt toke renewal to retain admin access

If token renewal fails then we log the user out

packages/utils/api-tests/utils.js
```diff
@@ -9,8 +9,20 @@ const createUtils = (strapi) => {
     if (!user) {
       throw new Error('User not found');
     }
-    const token = strapi.service('admin::token').createJwtToken(user);
 
+    const userId = String(user.id);
+    const deviceId = '00000000-0000-4000-8000-000000000001';
```
jhoward1994, line 14: Hard coded device ID for any API tests that are using this existing login util

packages/core/admin/server/src/controllers/authentication.ts
```diff
@@ -10,23 +12,111 @@ import {
...
+  return { ...base, expires: chosen, maxAge: Math.max(0, chosen.getTime() - now) };
+};
+
+const getSessionManager = (): Modules.SessionManager.SessionManagerService | null => {
+  const manager = strapi.sessionManager as Modules.SessionManager.SessionManagerService | undefined;
+
+  return manager ?? null;
+};
+
+const generateDeviceId = (): string => crypto.randomUUID();
```
jhoward1994, line 104: Do we need a fallback for this in the server code? The admin should always
provide it

innerdvations (approved):
(tested and reviewed live) LGTM!

PR #24310, opened 2025-09-04
feat: add support for custom field attributes in schema mapping

jhoward1994:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/core-api/routes/validation/mappers.ts
+++ b/packages/core/core/src/core-api/routes/validation/mappers.ts
@@ -9,6 +9,17 @@ import * as z from 'zod/v4';
 // eslint-disable-next-line import/no-cycle
 import * as attributes from './attributes';
 
+const isCustomFieldAttribute = (
+  attribute: unknown
+): attribute is { type: 'customField'; customField: string } => {
+  return (
+    !!attribute &&
+    typeof attribute === 'object' &&
+    (attribute as any).type === 'customField' &&
+    typeof (attribute as any).customField === 'string'
+  );
+};
+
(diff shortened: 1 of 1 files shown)


innerdvations:
Created an experimental: 0.0.0-experimental.75524da74c8427faada874289db9595b2831263a

Boegie19:
Tested the experimental version and it fixes all my issues.

boazpoolman:
I'm actually still having issues with this experimental release. I'm now getting the following error:
Scherm­afbeelding 2025-09-10 om 13 22 27

I can trace back that the error is being triggered [here](https://github.com/strapi/strapi/pull/24310/files#diff-18677909ceafb0ccc3ed2ec79c971da64645e9572bd7fb3dae6aeb703bf81458R152).

It seems that at the time of running that schema validation the custom fields haven't been registered yet.

Wondering how @Boegie19 didn't run in to this issue.

Boegie19:
Tryed again with @strapi/plugin-color-picker and @arshiash80/strapi-plugin-iconhub I don't run in to the same issue maby because of the loading order or someting

Boegie19:
@boazpoolman Can you send me the setup you used to get the error?

boazpoolman:
@Boegie19 I've setup a reproduction repo here: https://github.com/boazpoolman/24083-issue-reproduction

Mind you that the issue is being caused by the rest-cache plugin, probably because it does something with the routes during the register phase. I don't feel like what we're doing in that plugin is wrong though 🤔

PR #24316, opened 2025-09-05
fix: database is corrupt with orphaned relations

markkaylor:
### What does it do?

This script attempts to detect and clean orphaned/ghost relations that were created before https://github.com/strapi/strapi/pull/24227 was merged

### Why is it needed?

The database is corrupt

### How to test it?

You will need reset`packages/core/core/src/services/document-service/utils/unidirectional-relations.ts` and `packages/core/core/src/services/document-service/repository.ts` to how they were before #24227 was merged. You can browse the codebase at `https://github.com/strapi/strapi/tree/9a2f82f5dfa9429b9ac78a4c1051de4d742eeb85` for that

Now you can create some ghost relations by following the reproduction steps on #24227 

clear the core store, and after running strapi again the cleanup script will run and remove the ghost relations


packages/core/database/src/repairs/operations/remove-orphan-unidirectional-relations.ts
```diff
@@ -0,0 +1,330 @@
...
+): Promise<{ table: string; parentId: string } | null> => {
+  try {
+    // Get all models to find potential parents
+    const mdValues = db.metadata.values();
+    const allModels = Array.from(mdValues);
+
+    // Look for parent content types (api:: models)
+    for (const model of allModels) {
+      // Check if this model has a component join table
+      const cmpsTableName = `${model.tableName}_cmps`;
```
innerdvations: Once this is moved out of database, you should be able to use `getComponentJoinTableName` from `transform-content-types-to-models.ts` instead of string concatenation (which will break with long names). It might need to be snakeCased first too.

markkaylor:
> Do we plan to call this automatically at any point? Or leave it up to user code to call it and fix the issue if they're having it?

I was thinking we would leave it up to the user

markkaylor:
> Basic functionality looks good, but it should move out of core/database. Probably in the document service near the actual fix.

Ok sounds good I hesitated to do this but I guess it makes sense and should help clean things up. Just seems a bit strange to have a repair script anywhere else since another repair script already lives here... 🤔

markkaylor:
After refactor I've tested and confirm it works for:
- Component relations with draft and publish parent (orphaned relations removed in this case)
- Component relation WITHOUT draft and publish parent (no orphaned relations in this case, so nothing should be removed)
- Component relations where the component has a really long name or contains hyphens

packages/core/core/src/index.ts
```diff
@@ -6,6 +6,7 @@ import { destroyOnSignal, resolveWorkingDirectories, createUpdateNotifier } from
 
 export { default as compileStrapi } from './compile';
 export * as factories from './factories';
+export { repairOrphanUnidirectionalRelations } from './services/document-service/utils/repair-orphan-unidirectional-relations';
```
markkaylor: This probably isn't what we want to do 🤔. I'm not really sure the right place to put the function or the right pattern for exporting it. For now I just put it right next to the implementation and exported it here.

derrickmehaffy:
@markkaylor I'm with you I think this should live near the other repair script(s).

This does reinforce for me though that we really should have a CLI command to manually run repair scripts (I can think of a few maintenance scripts I could write too.

packages/core/core/src/services/document-service/utils/repair-orphan-unidirectional-relations.ts
```diff
@@ -0,0 +1,269 @@
...
+      if (!entriesBySource[sourceId]) {
+        entriesBySource[sourceId] = [];
+      }
+      entriesBySource[sourceId].push(entry);
+    }
+
+    const ghostEntries: number[] = [];
+
+    // Check if this is a component join table
+    const isComponentJoinTable = joinTableName.match(/^components_.*_.*_lnk$/);
```
innerdvations: This seems like it will work, but let's explicitly reference where the strings `components_` and `_lnk` come from, or even better import them if that's possible without refactoring other files or potentially introducing circular references.

innerdvations:
> @markkaylor I'm with you I think this should live near the other repair script(s).

It should be split the same way the first one is -- the method offered by the repair script should be generic for relations, while the caller outside the database package should be the one that handles the Strapi logic (relations, d&p, etc)

To keep the refactor simple and rewrite as little as possible, it could be something like:

`repairOrphanUnidirectionalRelations` -> lives in database package, but requires a "cleaning" function to be passed in

`cleanComponentJoinTable` would be that function, and it along with all its helper functions can live in document service.

Anywhere that we call the repair, it would pass in `cleanComponentJoinTable` as the cleaning method.

packages/core/core/src/index.ts
innerdvations: See my comment above, it should be exported from database next to the original, but split. Sorry for the confusion I caused!

packages/core/core/src/services/document-service/utils/repair-orphan-unidirectional-relations.ts
```diff
@@ -0,0 +1,269 @@
...
+          }
+        }
+      }
+    }
+
+    return ghostEntries;
+  } catch (error) {
+    return [];
+  }
+};
```
innerdvations:
```suggestion
};

```

laurenskling:
> > Do we plan to call this automatically at any point? Or leave it up to user code to call it and fix the issue if they're having it?
> 
> I was thinking we would leave it up to the user

I think most of the users don't even realize they have ghosts in their DB. It's very easy to create ghost, you don't see that something is wrong at first and it's impossible to get rid of them prior to 5.23.3. 
I would just run this as an update for everyone. Help everyone out, now and in the future, with this very nasty bug. What would be the downsides?

packages/core/core/src/services/document-service/utils/repair-orphan-unidirectional-relations.ts
laurenskling: I can set `collectionName: 'blablabla'` in my component object and it will _not_ be named `components_` in the db.

derrickmehaffy:
> > > Do we plan to call this automatically at any point? Or leave it up to user code to call it and fix the issue if they're having it?
> > 
> > I was thinking we would leave it up to the user
> 
> I think most of the users don't even realize they have ghosts in their DB. It's very easy to create ghost, you don't see that something is wrong at first and it's impossible to get rid of them prior to 5.23.3. 
> I would just run this as an update for everyone. Help everyone out, now and in the future, with this very nasty bug. What would be the downsides?

I can't disagree with that logic.

innerdvations:
Added some additional tests https://github.com/strapi/strapi/pull/24316/commits/aa7e2bc1013ffafedb11db8addd858010a5be36a

laurenskling:
Can I bring attention to this comment: https://github.com/strapi/strapi/pull/24227#issuecomment-3276093095 
There are still ghost relations being created in nested components.

Bassel17:
> Can I bring attention to this comment: [#24227 (comment)](https://github.com/strapi/strapi/pull/24227#issuecomment-3276093095) There are still ghost relations being created in nested components.

Thank you for bringing attention to it, I already included a fix for it in this PR, should be expected to be released in next weeks release cycle

laurenskling:
Just test on this branch and nested components are now also covered, thank you so much.

The cleanup doesn't work for me tho. It always runs with: `Unidirectional join table operation completed. Processed 0 entries.` although I know I have ghosts in my db. Is it me, or..?

laurenskling:
I get it..
https://github.com/strapi/strapi/pull/24316/files#diff-9923d0c981778d0fbcf62b23cca8f402c7e919133a8e0070c161b6fc18ed3068R185
Here there is a check on drafts, to find their published version, and see if its also in the list (so in the source draft). Those get removed correctly.
_but_. Because you can never remove these published versions from the draft component, you can end up with _only_ the published version linked:
1) Create ghosts like we now know how to do
2) clear the selectbox with relations, to have nothing related (but not remove the component, component remove will clear the db)
3) now we should have no relations in the db, but we do, we have the ghost published version of the relation still in the db.

This one is not removed by this PR.

jhoward1994:
I just tested this and it's working for me. This is what I did:

- Reverted the 2 files mentioned in the PR description
- Created content so I had ghost relations (so I was in a buggy state)
- Modified the DB so the repair would run again (I removed unidirectional-join-table-repair-ran from the core store)
```
    if (!alreadyRanComponentRepair) {
      await this.db.repair.processUnidirectionalJoinTables(cleanComponentJoinTable);
      await this.store.set({
        type: 'strapi',
        key: 'unidirectional-join-table-repair-ran',
        value: true,
      });
    }
```
(I think the instructions in the decription are wrong ? `await strapi.db.repair.removeOrphanUnidirectionalRelations();
`)
- Restarted the app & I could see the ghost relations were removed

Bassel17:
> I just tested this and it's working for me. This is what I did:
> 
> * Reverted the 2 files mentioned in the PR description
> * Created content so I had ghost relations (so I was in a buggy state)
> * Modified the DB so the repair would run again (I removed unidirectional-join-table-repair-ran from the core store)
> 
> ```
>     if (!alreadyRanComponentRepair) {
>       await this.db.repair.processUnidirectionalJoinTables(cleanComponentJoinTable);
>       await this.store.set({
>         type: 'strapi',
>         key: 'unidirectional-join-table-repair-ran',
>         value: true,
>       });
>     }
> ```
> 
> (I think the instructions in the decription are wrong ? `await strapi.db.repair.removeOrphanUnidirectionalRelations(); `)
> 
> * Restarted the app & I could see the ghost relations were removed

yeah the description is old, now it will also run once for all projects once they update, and if ever needed a user can run it manually

laurenskling:
> I get it.. https://github.com/strapi/strapi/pull/24316/files#diff-9923d0c981778d0fbcf62b23cca8f402c7e919133a8e0070c161b6fc18ed3068R185 Here there is a check on drafts, to find their published version, and see if its also in the list (so in the source draft). Those get removed correctly. _but_. Because you can never remove these published versions from the draft component, you can end up with _only_ the published version linked:
> 
> 1. Create ghosts like we now know how to do
> 2. clear the selectbox with relations, to have nothing related (but not remove the component, component remove will clear the db)
> 3. now we should have no relations in the db, but we do, we have the ghost published version of the relation still in the db.
> 
> This one is not removed by this PR.

Friends, this PR isn't complete... There are still ghosts in the DB.

Bassel17:
> > I get it.. https://github.com/strapi/strapi/pull/24316/files#diff-9923d0c981778d0fbcf62b23cca8f402c7e919133a8e0070c161b6fc18ed3068R185 Here there is a check on drafts, to find their published version, and see if its also in the list (so in the source draft). Those get removed correctly. _but_. Because you can never remove these published versions from the draft component, you can end up with _only_ the published version linked:
> > 
> > 1. Create ghosts like we now know how to do
> > 2. clear the selectbox with relations, to have nothing related (but not remove the component, component remove will clear the db)
> > 3. now we should have no relations in the db, but we do, we have the ghost published version of the relation still in the db.
> > 
> > This one is not removed by this PR.
> 
> Friends, this PR isn't complete... There are still ghosts in the DB.

Hello, we are aware that you are still facing issues and that it isn't removing the ghosts in your case, but we will merge this
because it does contain important fix and the cleanup script does work when the components have not been deleted.
as for your case we will just create a PR for next weeks release.
keep in mind we release on a weekly schedule, we just like to release any fix as quickly as possible.

laurenskling:
> > > I get it.. https://github.com/strapi/strapi/pull/24316/files#diff-9923d0c981778d0fbcf62b23cca8f402c7e919133a8e0070c161b6fc18ed3068R185 Here there is a check on drafts, to find their published version, and see if its also in the list (so in the source draft). Those get removed correctly. _but_. Because you can never remove these published versions from the draft component, you can end up with _only_ the published version linked:
> > > 
> > > 1. Create ghosts like we now know how to do
> > > 2. clear the selectbox with relations, to have nothing related (but not remove the component, component remove will clear the db)
> > > 3. now we should have no relations in the db, but we do, we have the ghost published version of the relation still in the db.
> > > 
> > > This one is not removed by this PR.
> > 
> > 
> > Friends, this PR isn't complete... There are still ghosts in the DB.
> 
> Hello, we are aware that you are still facing issues and that it isn't removing the ghosts in your case, but we will merge this because it does contain important fix and the cleanup script does work when the components have not been deleted. as for your case we will just create a PR for next weeks release. keep in mind we release on a weekly schedule, we just like to release any fix as quickly as possible.

cool cool. Do you want me to make a bug report for the other cleanup, or do you have it covered?
It basically comes down to published relations being linked to draft components, that need to be removed.

laurenskling:
@Bassel17 just upped my strapi version on my production db, and this repair started running against non-DP contentTypes, not just components. Now it removed the published relations from these entries. In the UI its still linked, because non-DP show the draft relation, but from the published API these relations are now gone. This is kinda bad :S

laurenskling:
I've created a PR that, in how I believe DP is supposed to work, fixes my previous issue.

It only fixes the repair. Everyone who installed 5.24.1 has data loss. We might need to write another repair script to bring back those relations. Shouldnt be too hard: find all non-DP entries with relations to a DP contentType. Add published version for every draft relation you find.

PR #24322, opened 2025-09-05
future(preview): better highlight click handling and errors

remidej:
### What does it do?

Another PR with unrelated changes, sorry about that.

- fixes the single/double click to edit behavior on preview highlights.
  - single clicks: it dispatches the click on the underlying element. So if a button has Strapi content, clicking it should click the button. I made the "email" field a button clickable in dummy-preview so we can test that. It also now sends a notification so users know they should double click to edit
  - double click: should open the input popover. Even when double clicking on a button. That was broken before.
- refactors previewScript a bit:
  - now it's the highlights that receive the click events, and that re-dispatch the click to the underlying element if needed. That's in order to make the single/double clicks work, the previous approach seemed like a dead end.
  - does as much as possible in CSS instead of in JS. So no more mouseenter and mouseleave handlers, just hover styles. It should be better for performance
- Adds translated notification messages so users are aware of the known limitations

### How to test it?

- Single and double a bunch of stuff, in particular the email button
- Try to edit a field from a relation

### DIFF
--- a/packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
+++ b/packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
@@ -140,6 +140,7 @@ const InputPopover = ({ documentResponse }: { documentResponse: ReturnType<UseDo
         width={iframeRect.width + 'px'}
         height={iframeRect.height + 'px'}
         zIndex={4}
+        onClick={() => iframeRef.current?.focus()}
       />
       <InputPopoverProvider>
         <Popover.Root open={true} onOpenChange={(open) => !open && setPopoverField(null)}>


SECURITY.md
remidej, line 1: That file does look like it needed formatting, but no idea how it ended up in my PR 🤷

markkaylor: 👍  I had the same issue on a community PR.

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -2,12 +2,17 @@ import * as React from 'react';
...
 import { type UseDocument } from '../../hooks/useDocument';
 import { InputRenderer } from '../../pages/EditView/components/InputRenderer';
 import { usePreviewContext } from '../pages/Preview';
-import { INTERNAL_EVENTS } from '../utils/constants';
-import { parseFieldMetaData, getAttributeSchemaFromPath } from '../utils/fieldUtils';
+import { INTERNAL_EVENTS, PREVIEW_ERROR_MESSAGES } from '../utils/constants';
+import {
+  parseFieldMetaData,
+  getAttributeSchemaFromPath,
+  PreviewFieldError,
```
markkaylor, line 14:
```suggestion
type  PreviewFieldError,
```

remidej: I don't think I can do that because instanceof is a runtime check by javascript, it's not typescript only

markkaylor: ahh that's my bad I though you were importing something else

markkaylor (approved):
Nice 👏

Issue #24324, opened 2025-09-07
[Bug] LocalePicker not showing all authorized locales


KABILA-Anas:
### Node Version

20.18.0

### NPM/Yarn/PNPM Version

Yarn 4.9.2

### Strapi Version

5.12.4

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When a user has multiple roles with overlapping permissions but on different locales, the `useI18n` hook does not merge the locales correctly. As a result, the Locale Picker doesn't show authorized locales from all roles.

So in the code below from `LocalPicker` component line 21, the hook will return locales related to create action from one role and locales related to read action from one role:
`const { hasI18n, canRead, canCreate } = useI18n();`

### Steps to Reproduce

1. Create two roles with the same permission on the same collection type, but each restricted to different locales::

- Role A → `canRead` on Articles for locale `en`
- Role B → `canRead` on Articles for locale `fr`

2. Assign both roles to the same user.
3. Login with that user and go to the Content Manager → Articles.
4. The Locale Picker only shows en (or fr), but not both.

### Expected Behavior

The Locale Picker should display both locales (`en`, `fr`).

### Code Snippets

In this example of permissions array, the user has two roles, each granting the same read permission on the Articles collection type, but for different locales:
```
[
  {
    "subject": "api::article.article",
    "action": "plugin::content-manager.explorer.read",
    "properties": { "locales": ["en"] }
  },
  {
    "subject": "api::article.article",
    "action": "plugin::content-manager.explorer.read",
    "properties": { "locales": ["fr"] }
  }
]
```
As a result, the Locale Picker displays only one locale (either `en` or `fr`), because the `useI18n` hook does not merge locales from multiple roles with the same action. It assumes that there will always be one entry for each action.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

oiorain:
I confirm this is still happening in v5.27.

Adzouz:
Hi,
The fix for this issue was merged and will be part of the version 5.30.0 🚀
Have a great day ☀️

PR #24325, opened 2025-09-07
fix(i18n): LocalePicker not showing all authorized locales

KABILA-Anas:
### What does it do?

Updates the `useI18n` hook to merge locales from all roles that share the same permission on a content type.
Previously, if a user had multiple roles with the same permission but different locales, only the locales from one role were returned.
This change uses `lodash/union` to combine locales from all applicable roles so the Locale Picker shows the complete set of authorized locales.

### Why is it needed?

Without this fix, the Locale Picker only shows locales from a single role when a user has multiple roles with the same permission but different locales.
This can confuse users, as they may think they don’t have access to certain locales, even though they can still navigate to them manually via the URL.

### How to test it?

1. Create two roles with the same permission on the same content type but different locales:

- Role A → `canRead` on Articles for locale en

- Role B → `canRead` on Articles for locale fr

2. Assign both roles to the same user.

3. Login with that user and go to Content Manager → Articles.

4. Open the Locale Picker — it should display both locales (`en`, `fr`).

### Related issue

https://github.com/strapi/strapi/issues/24324


Adzouz (approved):
Looks good now, thank you very much 🙌 You can merge it 🚀

KABILA-Anas:
Thanks for your feedbacks @Adzouz I still can't merge. I see that there are two failing checks!

Adzouz:
No worries, those two errors are not really concerning, I merged it 🙏 It will be part of the next release 🚀 Thanks again!

PR #24337, opened 2025-09-09
enhancement: remove default notification title

remidej:
### What does it do?

Removes the notification title that was added by default to the text of notifications. We were adding the notification variant as the title, but it looked clunky, and that information is still visible via the color and the toast and the icon.

before:

image

after:

image

When the title is explicitly passed however, then it's still visible.
### Why is it needed?

It was redundant, it looked clunky. @lucasboilly also wanted to remove it.
### How to test it?

do anything that triggers a notification, e.g. updating an entry


lucasboilly:
All good 👍

mathildeleg (approved):

mathildeleg (approved):

PR #24339, opened 2025-09-09
Updated! languageNativeNames.ts & add! Uzbek(Cyrillic)

Alirahmon777:
### What does it do?

Added Uzbek Cyrillic (uz-Cyrl) support by updating languageNativeNames.ts and adding uz.json and uz-Cyrl.json translations.

### Why is it needed?

To allow Admin panel users to choose between Uzbek Latin and Uzbek Cyrillic, improving accessibility for all Uzbek speakers.


hanpaine:
Hi @Alirahmon777 

Sadly right now we are not able to accept new languages are we are not equipped to review and validate the quality of them and the maintenance cost is too high for our small team.

However you could publish these translations for the community in your own repo using the setup described in
https://docs.strapi.io/cms/admin-panel-customization/locales-translations and others could then use them 🙏

PR #24346, opened 2025-09-10
feat: Advanced Session Configuration

jhoward1994:
Adds a core Session Manager and integrates it with both the Admin and the Content API (users-permissions) authentication flows. It introduces short-lived access tokens and refresh/session tokens, with per-origin configuration and session storage.

## What’s new
- Core SessionManager service that issues access and refresh/session tokens and tracks sessions in admin::session (table/collection strapi_sessions).
- Admin now authenticates via access tokens validated against active sessions; refresh tokens are kept in an httpOnly cookie.
- Users-permissions plugin supports a new refresh mode that uses SessionManager for access/refresh tokens (opt-in; default remains legacy long-lived JWT).
- New endpoints for token rotation and logout; per-origin configuration for token/session TTLs and cookie behavior.
- [Contributor docs added under Sessions and JWT](https://github.com/strapi/strapi/blob/feat/jwt-improvements/docs/docs/docs/01-core/authentication/00-sessions-and-jwt.md#L116-L117)

See those contributor docs for full information.

### Admin
 - Ensure admin.auth.secret is set; optional admin.auth.sessions.* to tune TTLs and cookie options.
- On login/register/reset, the server sets strapi_admin_refresh (httpOnly) and returns a short-lived access token in data.token.
- Send Authorization: Bearer <access token> on admin requests.
- Rotate access tokens with POST /admin/access-token (cookie is rotated, returns { data: { token } }).
- Logout with POST /admin/logout (optionally include { deviceId } to revoke a single device).
- rememberMe switches to a longer-lived refresh family and persistent cookie.

### Content API (Users-permissions)
- Two modes via plugin::users-permissions.jwtManagement:
  - legacy-support (default): unchanged long-lived JWT behavior.
  - refresh: opt-in to Session Manager access/refresh tokens.

In refresh mode:
- Login/register/callback responses return { jwt, refreshToken } (or set an httpOnly cookie if configured).
- Rotate/issue access token: POST /api/auth/refresh with { refreshToken }. If httpOnly is enabled (or header x-strapi-refresh-cookie: httpOnly is sent), the server rotates the cookie and returns { jwt }.
- Logout: POST /api/auth/logout revokes the authenticated user’s sessions; optional { deviceId } to target a device.

Example configurations:

```
// config/plugins.js
module.exports = {
  'users-permissions': {
    config: {
      jwtManagement: 'refresh', // or env('UP_JWT_MANAGEMENT', 'refresh')
      sessions: {
        accessTokenLifespan: env.int('UP_SESSIONS_ACCESS_TTL', 7 * 24 * 60 * 60),
        maxRefreshTokenLifespan: env.int('UP_SESSIONS_MAX_REFRESH_TTL', 30 * 24 * 60 * 60),
        idleRefreshTokenLifespan: env.int('UP_SESSIONS_IDLE_REFRESH_TTL', 7 * 24 * 60 * 60),
        httpOnly: env.bool('UP_SESSIONS_HTTPONLY', false),
        cookie: {
          name: env('UP_SESSIONS_COOKIE_NAME', 'strapi_up_refresh'),
          sameSite: env('UP_SESSIONS_COOKIE_SAMESITE', 'lax'),
          path: env('UP_SESSIONS_COOKIE_PATH', '/'),
          domain: env('UP_SESSIONS_COOKIE_DOMAIN'),
          secure: env.bool('UP_SESSIONS_COOKIE_SECURE', process.env.NODE_ENV === 'production'),
        },
      },
    },
  },
};
```

```
// config/admin.js
module.exports = {
  auth: {
    sessions: {
      accessTokenLifespan: 1800,
      maxRefreshTokenLifespan: 30 * 24 * 60 * 60,
      idleRefreshTokenLifespan: 7 * 24 * 60 * 60,
      maxSessionLifespan: 7 * 24 * 60 * 60,
      idleSessionLifespan: 3600,
    },
    cookie: {
      sameSite: 'lax',
      path: '/admin',
      domain: env('ADMIN_COOKIE_DOMAIN'),
    },
  },
};
```


derrickmehaffy:
Notes of things not working (or at least assumed to be issues, not sure what exactly is intended):

- `logged_in` cookie is not cleared or set to false
- `jwtToken.sig` cookie is not cleared (not sure if this is intended or not)

Confirmed things that are not working or not implemented:

- Deleting a user does not delete their sessions (in my example I deleted the SSO user and recreated them):
image
image

- Changing a user's password does not delete their session and force a re-auth (didn't test if it deleted other sessions but left the current one alone but I think it should delete all sessions for that user and force a re-auth).

derrickmehaffy:
I'm getting a weird issue when trying to authenticate from another system where the login just does nothing (running the app on one PC and trying to login with another on the same network).

derrickmehaffy:
image

jhoward1994:
0.0.0-experimental.5771a2e1430534dcd034abb693b216b2af3caf75

alxnkt:
If anyone is interested, I use this setup:
```js
// server.ts
proxy: { 
  koa: process.env.NODE_ENV === 'production' 
},
```
It's very convenient _in my case_ — It works on my dev environment (production "is false")  and also works in production.

lolafkok:
So the solution for v5 is Strapi switches from stateless JWT (legacy, default) to JWT with sessions (at some point the alternative option)?

And because this is not a small change you've decided to not deliver this as part of the security-updates for v4 users, right?

PR #24354, opened 2025-09-11
feat: homepage customization

mathildeleg:
### What does it do?
Add the ability to customize the layout of the homepage widgets for a user: they can modify the order of the widgets through a drag&drop, resize them, delete non-essential widgets or add them back.
https://github.com/user-attachments/assets/0fe76003-77ee-4120-93b8-d2f07f13a479

### Why is it needed?
Enhance the user experience on the homepage to fit their needs and roles.

### How to test it?
- Login to the admin interface and go to the Homepage.
- Play around with the customization.
- Go to another workspace in the platform (for example the Content Manager) and go back to the Homepage to confirm your changes were saved.

### DIFF
--- a/packages/core/upload/admin/src/pages/App/ConfigureTheView/tests/__snapshots__/ConfigureTheView.test.tsx.snap
+++ b/packages/core/upload/admin/src/pages/App/ConfigureTheView/tests/__snapshots__/ConfigureTheView.test.tsx.snap
@@ -685,7 +685,6 @@ exports[`Upload - Configure initial render renders and matches the snapshot 1`] 
               </div>
             </div>
           </div>
-          x
         </form>
       </main>
     </div>


packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+      secondWidgetLeft + lastRect.width,
+      rowTop,
+      rowHeight,
+      20,
+      rowIndex
+    )
+  );
+};
+
+const addVerticalDropZonesForThreeWidgets = (
```
Adzouz, line 148: I think we could improve and merge the two functions into one because the logic is pretty similar. Tell me if you need help refactoring :)

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+  // Set up resize observer for position updates
+  React.useEffect(() => {
+    const updatePositions = () => {
+      const newPositions = calculateGapDropZonePositions();
+      setPositions(newPositions);
+    };
+
+    // Update positions when widgets change size
+    const resizeObserver = new ResizeObserver(() => {
+      setTimeout(updatePositions, 50); // Small delay to ensure DOM is updated
```
Adzouz, line 366: Is there a particular reason having to do that? Is there a little amount of time where the layout is unstable?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+    window.addEventListener('resize', handleResize);
+
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, [calculateGapDropZonePositions]);
+
+  // Set up resize observer for position updates
+  React.useEffect(() => {
+    const updatePositions = () => {
```
Adzouz, line 359: This method is a duplicate from line 337. Maybe we could move it higher in the global component scope to declare it only once?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+      }
+    });
+
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, [filteredWidgets, calculateGapDropZonePositions]);
+
+  return (
+    <>
```
Adzouz, line 383: Just picky here but this Fragment is not necessary. You could directly return positions.map((gapDropZone, index) => (

packages/core/admin/admin/src/components/GapDropZone.tsx
Adzouz: I think we could improve and merge the three functions into one because the logic is pretty similar. Tell me if you need help refactoring :)

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+      const canAcceptMoreWidgets = widgetCount < 3;
+
+      const shouldShowVerticalDropZones =
+        (isDraggingFromThisRow || (isDraggingFromAnotherRow && canAcceptMoreWidgets)) &&
+        isDraggingWidget;
+
+      // Add vertical drop zones based on widget count
+      if (shouldShowVerticalDropZones) {
+        if (widgetCount === 1) {
+          addVerticalDropZonesForSingleWidget(gapDropZones, row, rowInfo, rowIndex);
```
Adzouz: Same comment as above but to complete it: here it would be only one single call to the function. We could pass the count as a param.

packages/core/admin/admin/src/components/GapDropZone.tsx
Adzouz: This method is a duplicate from line 337. Maybe we could move it higher in the global component scope to declare it only once?

packages/core/admin/admin/src/components/GapDropZone.tsx
Adzouz: Is there a particular reason having to do that? Is there a little amount of time where the layout is unstable?

packages/core/admin/admin/src/components/GapDropZone.tsx
Adzouz: Just picky here but this Fragment is not necessary. You could directly return `positions.map((gapDropZone, index) => (`

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, [filteredWidgets, calculateGapDropZonePositions]);
+
+  return (
+    <>
+      {positions.map((gapDropZone, index) => (
+        <GapDropZone
+          key={`gap-drop-zone-${index}`}
```
Adzouz: As a good practice in React, we don't use indexes as key for loop elements because they can change during dom manipulation. Maybe try using something unique from gapDropZone?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+  isVisible: boolean;
+  type: 'vertical' | 'horizontal';
+  moveWidget: (id: string, to: number, targetRowIndex?: number, isHorizontalDrop?: boolean) => void;
+  targetRowIndex?: number;
+}
+
+const GapDropZoneContainer = styled(Box)<{
+  $isOver: boolean;
+}>`
+  background-color: ${({ $isOver }) => ($isOver ? 'rgba(0, 123, 255, 0.2)' : 'transparent')};
```
Adzouz: What about this colour when switching dark / light theme? Could we use one from the design system instead?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+  transition: all 0.2s ease-in-out;
+  cursor: pointer;
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  width: 20px;
```
Adzouz: You could use the constant here too

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+) => {
+  if (!rowInfo) return;
+
+  const { firstRect, lastRect, containerRect, rowTop, rowHeight } = rowInfo;
+  const firstWidgetLeft = firstRect.left - containerRect.left;
+  const secondWidgetLeft = lastRect.left - containerRect.left;
+
+  // Before the first widget
+  gapDropZones.push(
+    createVerticalDropZone(row.startIndex, firstWidgetLeft - 20, rowTop, rowHeight, 20, rowIndex)
```
Adzouz: I see a lot the value 20 (I assume it's the width / height of the drop zone). Can you create a constant and reuse it instead to manage it only at one place?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+  cursor: pointer;
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  width: 20px;
+  z-index: 10;
```
Adzouz: I'm afraid that this value will pass over the sticky nav header after scrolling in the page. Look at the video: https://www.loom.com/share/a6ca6bd58b0245ff834c5aa428e0cab4?sid=0e8b8791-ab44-49c4-9ae3-7501fbbeaf2c

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,467 @@
...
+  if (!isVisible) {
+    return null;
+  }
+
+  return (
+    <GapDropZoneContainer
+      ref={drop}
+      $isOver={isOver}
+      style={{
+        left: `${position.left}px`,
```
Adzouz: Instead of using position (left, top) can you use `transform: translate(${position.left}px, ${position.top}px)` property in CSS? (more perf, updates better in the dom)

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,447 @@
...
+  position: { left: number; top: number; height: number };
+  currentLeftWidth: number;
+  currentRightWidth: number;
+  totalColumns?: number;
+  rowPosition?: { left: number; top: number; width: number; height: number } | null;
+}
+
+const IndicatorContainer = styled(Box)<{ $isVisible: boolean }>`
+  position: absolute;
+  z-index: 20;
```
Adzouz: Same here, this z-index is too high and will pass over the sticky nav header

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,447 @@
...
+  transform: ${({ $isCurrent }) => ($isCurrent ? 'scale(1.2)' : 'scale(1)')};
+`;
+
+const calculateGapAdjustment = (rowWidth: number, leftColumns: number): number => {
+  const dotWidth = 6;
+  const gapAdjustmentPixels = dotWidth / 2; // Half dot width to center on boundary
+  const gapAdjustmentPercent = (gapAdjustmentPixels / rowWidth) * 100;
+
+  // Different adjustments for different positions
+  if (leftColumns === 4) return -gapAdjustmentPercent; // Left dot
```
Adzouz: Not a big deal but could be either a switch or since we're returning 0 in two cases, could be simplified to 

`if (leftColumns === 4) return -gapAdjustmentPercent;`
`if (leftColumns === 6) return gapAdjustmentPercent;`
`return 0;`

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,447 @@
...
+    </IndicatorContainer>
+  );
+};
+
+const ResizeHandleContainer = styled(Box)<{ $isDragging?: boolean }>`
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  width: 20px;
+  z-index: 10;
```
Adzouz: Same here

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,447 @@
...
+      }
+
+      // Throttle the resize calls to prevent excessive updates
+      throttleRef.current = setTimeout(() => {
+        const deltaX = e.clientX - state.startX;
+        const threshold = 120; // Pixels per column unit
+        const deltaColumns = Math.round(deltaX / threshold);
+
+        handleResize(deltaColumns);
+      }, 2);
```
Adzouz: Is it arbitrary to put 2 only here? I get that the setTimeout is there for a reason but why 2 and not 0 to make it async?

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,447 @@
...
+      }
+    };
+  }, []);
+
+  return (
+    <>
+      <ResizeHandleContainer
+        onMouseDown={handleMouseDown}
+        style={{
+          left: `${state.position.left}px`,
```
Adzouz: Same here about `transform: translate` positionning.

packages/core/admin/admin/src/components/WidgetRoot.tsx
```diff
@@ -0,0 +1,169 @@
...
+          node.setAttribute('data-widget-id', uid);
+        }
+      }}
+      onMouseEnter={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
+      style={{
+        transition: 'all 0.3s cubic-bezier(0.4, 0, 0.2, 1)',
+      }}
+    >
+      <Flex direction="row" gap={2} width="100%" tag="header" alignItems="center" minHeight="22px">
```
Adzouz: We use rem unit on the platform most of the time (some cases are exceptions). If you think this one can be converted, go for it 😊

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -24,9 +24,19 @@ import { RelativeTime } from './RelativeTime';
...
 
-const CellTypography = styled(Typography)`
+/**
+ * Calculate dynamic max-width based on column span
+ * Base width is 14.4rem for 6 columns, scale proportionally
+ */
+const calculateDynamicMaxWidth = (columnWidth: number = 4): string => {
+  const baseColumnWidth = 4;
+  const baseMaxWidth = 14.4; // rem
+  return `${(baseMaxWidth * columnWidth) / baseColumnWidth}rem`;
```
Adzouz: We might want to round it at one decimal to avoid having pixels value with decimals after conversion by the browser.

packages/core/content-manager/admin/src/components/Widgets.tsx
```diff
@@ -35,9 +45,11 @@ const CellTypography = styled(Typography)`
 const RecentDocumentsTable = ({
   documents,
   type,
+  dynamicMaxWidth = '14.4rem',
```
Adzouz: Since 14.4rem is reused at three different places, you could create a constant.

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,372 @@
...
+        columnWidths,
+        leftWidgetId,
+        rightWidgetId,
+        newLeftWidth,
+        newRightWidth
+      );
+
+      setColumnWidths(newWidths);
+
+      saveLayout(
```
Adzouz: I don't know if it's possible because I didn't digged that much in the code but it would be better maybe triggering the save only when the cursor release the click when resizing because it could trigger a lot of calls at each layout change.

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,372 @@
...
+        result.newWidgets,
+        result.newWidths,
+        updateHomepageLayout,
+        toggleNotification,
+        formatAPIError,
+        formatMessage
+      );
+    },
+    [
+      filteredWidgets,
```
Adzouz, line 256: Considering the amount of variables to watch, I don't know if the useCallback is really useful here and they are passed to components that don't memoize. Same for the next functions until handleDragStart (these ones are legit).

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+    <Box
+      padding={4}
+      background={isHovered ? 'primary100' : 'neutral0'}
+      borderColor={isHovered ? 'primary200' : 'neutral200'}
+      hasRadius
+      shadow="tableShadow"
+      onClick={onSelect}
+      width={'100%'}
+      cursor="pointer"
+      onMouseEnter={() => setIsHovered(true)}
```
Adzouz: If it's just for design purposes, I would move the hover to CSS only + we want to make sure it won't happen on mobile.

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+        </Modal.Header>
+        <Modal.Body>
+          <ScrollArea>
+            <Box padding={6}>
+              {addableWidgets.length === 0 ? (
+                <Flex direction="column" alignItems="center" gap={4}>
+                  <PuzzlePiece width="4rem" height="4rem" fill="neutral300" />
+                  <Typography textColor="neutral500" textAlign="center">
+                    {formatMessage({
+                      id: 'HomePage.addWidget.noWidgetsAvailable',
```
Adzouz: We would need to add it to at least en.json file (corresponding to homepage widgets feature). Flagging this one but might be others missing too.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -162,23 +215,95 @@ const HomePageCE = () => {
...
+
+            return (
+              <DragPreviewWrapper $maxWidth={maxWidth}>
+                <WidgetRoot
+                  uid={item.id}
+                  title={item.title || { id: `${item.id}`, defaultMessage: item.id }}
+                  icon={item.icon}
+                  link={item.link}
+                  findWidget={() => ({ index: 0 })}
+                  moveWidget={() => {}}
```
Adzouz: Can't those properties be optional since we initialize them without any particular logic?

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,372 @@
...
+    widget,
+    index: filteredWidgets.indexOf(widget),
+  };
+};
+
+const saveLayout = async (
+  widgets: WidgetWithUID[],
+  widths: Record<string, number>,
+  updateHomepageLayout: (data: {
+    widgets: Array<{ uid: string; width: 4 | 6 | 8 | 12 }>;
```
Adzouz: Since those values are available in widgetUtils can you use WIDGET_SIZING here? `(typeof WIDGET_SIZING.DISCRETE_SIZES)[number]` or use the definition we have in packages/core/admin/shared/contracts/homepage.ts (but if possible have this list only at one place)

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,372 @@
...
+  }) => Promise<any>,
+  toggleNotification: (config: { type: 'danger'; message: string }) => void,
+  formatAPIError: (error: any) => string,
+  formatMessage: (descriptor: { id: string; defaultMessage: string }) => string
+) => {
+  try {
+    const layoutData = {
+      widgets: widgets.map((widget) => ({
+        uid: widget.uid,
+        width: (widths[widget.uid] || 12) as 4 | 6 | 8 | 12,
```
Adzouz: Same here

Adzouz (changes requested):
Well done, that was a tough one and you did great 👏 The overall experience is amazing, we were waiting for something cool but it's even better than we could expect so thank you 🙏
Just a few questions / refactoring tiny details but overall it looks good!
Just had a tricky question I was asking myself during the review: what is happening if we decide to create a new widget and the user has already configured their page. I know it will appear in the add widgets but we might want to automatically push new widgets to the dashboard don't you think? Do you have any idea?

Also another quick point that could be resolved after (during responsiveness optimization of the homepage but could be quick to fix) the widgets in the add widgets modal are not responsive and overflow in the content:

Image

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/cf41541f8b5a1583a9bd0c37113d38a96788646c

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Good catch, I thought I'd tried to refactor this already but must have been before I made other changes! It's refactored here https://github.com/strapi/strapi/pull/24354/commits/eee3300aa65814d391e24aa11283f3440234a283

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Refactored here https://github.com/strapi/strapi/pull/24354/commits/eee3300aa65814d391e24aa11283f3440234a283

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Good catch, it wasn't actually needed :) it's been removed here https://github.com/strapi/strapi/pull/24354/commits/eee3300aa65814d391e24aa11283f3440234a283

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Removed here as the useEffect() was not actually needed https://github.com/strapi/strapi/pull/24354/commits/eee3300aa65814d391e24aa11283f3440234a283

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here, let me know if you think this works best or not: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Good catch! It's been fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/ResizeIndicator.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/ResizeIndicator.tsx
mathildeleg: You're right, 0 works well too :) fixed here https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/components/WidgetRoot.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/content-manager/admin/src/components/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/content-manager/admin/src/components/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: You're right it was saving it even if the user was still dragging the resizeHandle, as it was snapping to the new width. Now it only saves if the user's releases the mouse: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: It's been refactored here https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/pages/Home/HomePage.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/bc56b9a945442e1b88cc76d40ed5050d77ac66b5

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
mathildeleg: Thanks I didn't know about it! It's been added now: https://github.com/strapi/strapi/pull/24354/commits/f85ef0174f624939c79f18c18615ca7292d90feb

mathildeleg:
> Well done, that was a tough one and you did great 👏 The overall experience is amazing, we were expecting something cool but it's even better than we could imagine, so thank you 🙏
> 
> Just a few questions / refactoring tiny details but overall it looks good!
> 
> Just had a tricky question I was asking myself during the review: what is happening if we decide to create a new widget and the user has already configured their page. I know it will appear in the add widgets but we might want to automatically push new widgets to the dashboard don't you think? Do you have any idea?
> 
> Also another quick point that could be resolved after (during responsiveness optimization of the homepage but could be quick to fix) the widgets in the add widgets modal are not responsive and overflow in the content:
> 
> Image

Thanks Adrien 🙏🏻

Tbh I didn't think about this, it's a behaviour we didn't think about/discussed at all, it might be good to have a conversation with Vincent about this to know how to handle it best in a UX perspective, as well as with Araksya to know how it could be handled on the backend!

I'll try to fix the responsiveness of the widgets in that modal this afternoon! 
Edit: responsiveness in the modal is fixed :)

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
```diff
@@ -50,14 +63,14 @@ const WidgetPreview = ({ widget, onSelect }: WidgetPreviewProps) => {
             {formatMessage(widget.title)}
           </Typography>
         </Flex>
-        <Box
+        <StyledBox
```
Adzouz: Not a big deal but I would give it a more explicit name like WidgetWrapper

mathildeleg: Fixed here (as well as fixed from your other comments from this morning): https://github.com/strapi/strapi/pull/24354/commits/493e984e2e5b45d0728814708be16d70cc597575

Adzouz (approved):
LGTM, thanks a lot 🙏 I would maybe wait for another review just to be sure though if that's okay for you 🙂

mathildeleg:
> LGTM, thanks a lot 🙏 I would maybe wait for another review just to be sure though if that's okay for you 🙂

Yes I'd appreciate someone's else review too! I'll remind people tomorrow once the release is done :) thanks 🙏🏻

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
+import * as React from 'react';
+
+import { Box } from '@strapi/design-system';
+import { useDrop } from 'react-dnd';
```
markkaylor, line 5: For some reason I thought we were planning on using dnd-kit instead of react-dnd, at least that's what it looked like in the early POC demo video. (Not asking to rewrite everything 🙃)

packages/core/admin/admin/src/components/DragLayer.tsx
```diff
@@ -45,6 +60,10 @@ const DragLayer = ({ renderItem }: DragLayerProps) => {
     return null;
   }
 
+  const renderPreview = () => {
+    return renderItem({ type: itemType, item });
+  };
```
markkaylor: Is the renderPreview wrapper function necessary? Couldn't you just call renderItem directly?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+  transition: all 0.2s ease-in-out;
+  cursor: pointer;
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  width: ${DROP_ZONE_SIZE}rem;
```
markkaylor: I notice here this constant is used to provide a rem value, I'm wondering is it always used for a rem value or is it sometimes used for a px value? I'm asking because in most cases elsewhere I see px instead of rem

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,456 @@
...
+
+const shouldTriggerResize = (
+  leftWidth: number,
+  rightWidth: number,
+  lastResizeValues: { leftWidth: number; rightWidth: number }
+): boolean => {
+  return leftWidth !== lastResizeValues.leftWidth || rightWidth !== lastResizeValues.rightWidth;
+};
+
+export const InterWidgetResizeHandle = ({
```
markkaylor: What does InterWidget mean? Internal?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+  draggedWidgetId?: string;
+  moveWidget: (id: string, to: number, targetRowIndex?: number, isHorizontalDrop?: boolean) => void;
+}
+
+const getRowInfo = (row: WidgetRow) => {
+  const firstWidgetElement = document.querySelector(`[data-widget-id="${row.widgets[0].uid}"]`);
+  const lastWidgetElement = document.querySelector(
+    `[data-widget-id="${row.widgets[row.widgets.length - 1].uid}"]`
+  );
+  const containerElement = document.querySelector('[data-grid-container]');
```
HichamELBSI: I would change the id name to something specific to Strapi just to make sure we will never any other element that isn't Strapi related, for example if a browser extension add element to the pages with that kind of name, which (data-grid-container and data-widget-[uid] seems to generic to me). It's an edge case that would probably never happen, but the cost of changing the ID is negligible, so I think we can fix it. Something like `strapi-grid-container` or `data-strapi-grid-container` (`strapi-widget-[uid]` `data-strapi-widget-[uid]`)

What do you think about it?

packages/core/admin/admin/src/components/GapDropZone.tsx
HichamELBSI, line 5: Can you add a comment here to mention that at some point we should use `@dnd-kit/core` as it has been done recently by Alex in the CTB for consistency.

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+      lastWidget.left + lastWidget.width,
+      rowTop,
+      rowHeight,
+      DROP_ZONE_SIZE,
+      rowIndex
+    )
+  );
+};
+
+const addHorizontalDropZones = (
```
HichamELBSI: - This not a pure function, make it pure (do not mutate array that came from outside, instead, return an array)
- export this function and add unit test to it

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+  width: number = DROP_ZONE_SIZE
+): GapDropZonePosition => ({
+  insertIndex,
+  position: { left, top, height, width },
+  isVisible: true, // Will be controlled by parent
+  type: 'horizontal',
+  isHorizontalDrop: true,
+});
+
+const addVerticalDropZones = (
```
HichamELBSI: - This not a pure function, make it pure (do not mutate array that came from outside, instead, return an array)
- export this function and add unit test to it

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,332 @@
...
+  };
+
+  return {
+    findWidget: findWidgetFn,
+    deleteWidget: deleteWidgetFn,
+    addWidget: addWidgetFn,
+    moveWidget: moveWidgetFn,
+    columnWidths,
+    setColumnWidths,
+    WidgetRoot,
```
markkaylor: Is this just for convenience? Couldn't we just import the component wherever its needed?

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+    updatePositions();
+
+    // Update positions on window resize
+    const handleResize = () => {
+      // Debounce the resize handler
+      setTimeout(updatePositions, 100);
+    };
+
+    window.addEventListener('resize', handleResize);
```
HichamELBSI: use a new ResizeObserver here instead

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -1,91 +1,57 @@
...
-    </Flex>
-  );
-};
+// Styled wrapper for the drag preview
+const DragPreviewWrapper = styled.div<{ $maxWidth: string }>`
+  max-width: ${(props) => props.$maxWidth};
+  overflow: hidden;
+  opacity: 0.9;
+  border: 2px solid ${({ theme }) => theme.colors.primary500};
+  border-radius: 4px;
```
markkaylor: I think there is a border-radius property on theme

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+      }
+
+      // Add horizontal drop zones
+      addHorizontalDropZones(gapDropZones, row, rowIndex, rowInfo, widgetRows, filteredWidgets);
+    });
+
+    return gapDropZones;
+  }, [widgetRows, isDraggingWidget, draggedWidgetId, filteredWidgets]);
+
+  React.useEffect(() => {
```
HichamELBSI: This effect will be executed for every resize (even when you don't drag. You can try to add a log and resize the page or browser). I guess you don't need to do anything when it is not dragging, right?

```
if (!isDraggingWidget) { setPositions([]); return; }
```

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -1,91 +1,57 @@
...
 
-const WidgetComponent = ({ component }: { component: () => Promise<React.ComponentType> }) => {
+export const WidgetComponent = ({
+  component,
+  columnWidth,
+}: {
+  component: () => Promise<React.ComponentType>;
+  columnWidth: number;
+}) => {
   const [loadedComponent, setLoadedComponent] = React.useState<React.ComponentType | null>(null);
```
markkaylor: Would this achieve the same thing without the type casting below?

```suggestion
const [loadedComponent, setLoadedComponent] = React.useState<React.ComponentType<{
    columnWidth?: number;
  }> | null>(null);
```

packages/core/admin/admin/src/components/GapDropZone.tsx
HichamELBSI: You will probably not need to debounce anymore with the resizeobserver and this condition.

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+      // Debounce the resize handler
+      setTimeout(updatePositions, 100);
+    };
+
+    window.addEventListener('resize', handleResize);
+
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, [calculateGapDropZonePositions]);
```
HichamELBSI, line 295:
```suggestion
  }, [isDraggingWidget , calculateGapDropZonePositions]);
```

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  return (
+    <GapDropZoneContainer
+      ref={drop}
+      $isOver={isOver}
+      style={{
+        transform: `translate(${position.left}px, ${position.top}px)`,
+        height: `${position.height}px`,
+        width: `${position.width}px`,
```
HichamELBSI, line 374: There is a unit consistency issue here, line 337 you are using a `rem` value and here `px`, you might see a UI bug if user use a different base font size.

packages/core/admin/admin/src/components/GapDropZone.tsx
HichamELBSI: At some point these two functions should look like this: 

```
const const addHorizontalDropZones = (
  row: WidgetRow,
  rowIndex: number,
  rowInfo: ReturnType<typeof getRowInfo>,
  widgetRows: WidgetRow[],
  filteredWidgets: WidgetWithUID[]
) => {
   ...;
   const horizontalDropZones = [];

   ...; 

   return horizontalDropZones;
}
```

Same for vertical ones.

packages/core/admin/admin/src/components/GapDropZone.tsx
HichamELBSI: because it depends on layout positioning, you can also use a `useLayoutEffect`instead of a useEffect.

markkaylor (commented):
Nice work! That's a big one. I tested and looks like it's working nicely. Drag and drop can be such a pain 😬 


I can't resize a full width widget, is it expected? I guess technically it is possible by drag and drop to a smaller drop zone...

packages/core/admin/admin/src/components/WidgetRoot.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+import type { WidgetType } from '@strapi/admin/strapi-admin';
+
+const WidgetActions = styled(Flex)`
+  display: flex;
+
+  @media (hover: hover) and (pointer: fine) {
+    display: none;
+  }
+`;
```
HichamELBSI, line 36: Using hover state make this actions not usable with the keyboard. Now, users can't delete a widget from the keyboard. For the sake of accessibility, we should use focus state.

packages/core/admin/admin/src/components/ResizeIndicator.tsx
```diff
@@ -0,0 +1,456 @@
...
+      <ResizeIndicator
+        isVisible={state.isDragging}
+        position={state.position}
+        currentLeftWidth={state.currentResizeValues.leftWidth}
+        currentRightWidth={state.currentResizeValues.rightWidth}
+        rowPosition={state.rowPosition}
+      />
+    </>
+  );
+};
```
HichamELBSI, line 379: I will not have the time to read line by line but here's my two cents:
- This file is way too long with a lot of helpers and many components (make sure they are pure so you can add unit tests)
- You are using effect instead of layout effect for layout updates
- You can use pointermove instead of mousemove (even though we aren't handling mobile here)
- You can use the existing ResizeObserver you defined in the useEffect to also watch the grid container instead of combining ResizeObserver and resize listener.

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,332 @@
...
+    handleInterWidgetResize: handleInterWidgetResizeFn,
+    saveLayout: saveLayoutFn,
+    isDraggingWidget,
+    draggedWidgetId,
+    handleDragStart,
+    handleDragEnd,
+  };
+};
```
HichamELBSI, line 364: Quite the same feeling here about the file size, I won't have the time to read in detail.

Also, you are using useCallback functions, did you notice any performance issue related to the handleDragStart and handleDragEnd function without using memoization?

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+      padding={4}
+      background={isHovered ? 'primary100' : 'neutral0'}
+      borderColor={isHovered ? 'primary200' : 'neutral200'}
+      hasRadius
+      shadow="tableShadow"
+      onClick={onSelect}
+      width={'100%'}
+      cursor="pointer"
+      onMouseEnter={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
```
HichamELBSI: Does this means that it won't be accessible to keyboard users?

packages/core/admin/admin/src/pages/Home/HomePage.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/e9fc8fb0a8a85efaee3619a289689bc366bbefed

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: I thought I'd changed this but I must have missed it! Here's the fix: https://github.com/strapi/strapi/pull/24354/commits/69e54a287582b49a4575af7b8eaf82503f3acead

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/281845d29f60e5018cd18ffdc38ab99e3e5af5a6 :)

packages/core/admin/admin/src/components/DragLayer.tsx
mathildeleg: I agree this wasn't needed: https://github.com/strapi/strapi/pull/24354/commits/bfeb2ed52781a64a4da67db42d1170b60a8df1a8

packages/core/admin/admin/src/pages/Home/HomePage.tsx
mathildeleg: This wasn't my line of code but I changed it here: https://github.com/strapi/strapi/pull/24354/commits/268e5a13bf29bd847b4f20f791813035792a6881

packages/core/admin/admin/src/components/ResizeIndicator.tsx
mathildeleg: This was meant to explicitly say it was a handle between widgets, but I agree in retrospect it wasn't clear haha is this naming better https://github.com/strapi/strapi/pull/24354/commits/9e578fbd41531cf8afaad86de0e3b3b8aaac51e1 ?

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Here I modified the gapDropZone to have pure functions, and created unit tests for them: https://github.com/strapi/strapi/pull/24354/commits/4e8f02e4fca83a7cf77d2031aae7859b759edeae

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here (same as for horizontalDropZones): Here I modified the gapDropZone to have pure functions, and created unit tests for them: https://github.com/strapi/strapi/pull/24354/commits/4e8f02e4fca83a7cf77d2031aae7859b759edeae

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/47b3303869c71f6a8eae85a896f592e481a59f2d

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/47b3303869c71f6a8eae85a896f592e481a59f2d

- changed to useLayoutEffect()
- added a check before calling the GapDropZoneManager to avoid calling it if the widget isn't dragged -> do you agree this is better than what you recommended? open to your thoughts :)

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Not needed because of how I implemented the fix (https://github.com/strapi/strapi/pull/24354/commits/47b3303869c71f6a8eae85a896f592e481a59f2d)

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: I tried (for two hours lol) to use rem values instead of px to fit with the existing codebase, however I really couldn't get it to work because the values returned by the getBoundingClientRect are px, and so it always messed up at least one gapDropZone. Decided instead to modify the DROP_ZONE_SIZE to px again: 
https://github.com/strapi/strapi/pull/24354/commits/dce316851a0c7e45b2d31f926aca147b95692460
Let me know if you have ideas though...

packages/core/admin/admin/src/pages/Home/components/AddWidgetModal.tsx
mathildeleg: Improved accessibility here: https://github.com/strapi/strapi/pull/24354/commits/704dd51ba08ab36f3487ec39636c35e32fb67363

packages/core/admin/admin/src/components/WidgetRoot.tsx
mathildeleg: Fixed accessibility here, let me know your thoughts: https://github.com/strapi/strapi/pull/24354/commits/704dd51ba08ab36f3487ec39636c35e32fb67363

packages/core/admin/admin/src/components/ResizeIndicator.tsx
mathildeleg: Decided to split helpers for resize implementation and for widgets directly into two distinct files, hoping it helps the readability and maintainability of the code: 
https://github.com/strapi/strapi/pull/24354/commits/b5cb5d05e29f304172d7305b1b3a708152767315

Also applied the other changes mentioned in your comment.

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: Unfortunately with react-dnd the useCallback functions are needed to avoid re-renderings

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Sorry not to have used this, in the POC for this feature react-dnd was used and I was under the impression this was the recommended lib to use for this implementation. I added the comment https://github.com/strapi/strapi/pull/24354/commits/7c15dcebf282476ae046f8536692d037fa854140 but find it a bit of a shame that it wasn't more explicit to use this as now a refactor will be needed :/

mathildeleg: (Same comment as below):
Sorry not to have used this, in the POC for this feature react-dnd was used and I was under the impression this was the recommended lib to use for this implementation. I added the comment https://github.com/strapi/strapi/pull/24354/commits/7c15dcebf282476ae046f8536692d037fa854140 but find it a bit of a shame that it wasn't more explicit to use this as now a refactor will be needed :/

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: This was changed during a review to comply with Strapi's use of REM values, however I changed it back to px values after trying to change all values in the file to REM, explained here: https://github.com/strapi/strapi/pull/24354#discussion_r2416957275

mathildeleg:
> Nice work! That's a big one. I tested and looks like it's working nicely. Drag and drop can be such a pain 😬
> 
> I can't resize a full width widget, is it expected? I guess technically it is possible by drag and drop to a smaller drop zone...

Hi Mark, thanks! We decided the homepage layout would have some restrictions to have a better UX: that means the grid doesn't allow empty spaces, thus a widget taking the full width of row cannot be resized yes, it can only be dropped to another row :) hope it makes sense, let me know if you have other questions!

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+const moveWidget = (
+  filteredWidgets: WidgetWithUID[],
+  columnWidths: Record<string, number>,
+  widgetId: string,
+  insertIndex: number,
+  targetRowIndex?: number,
+  isHorizontalDrop?: boolean
+) => {
```
remidej: when there are more than 2 params I find that a single object param is more clear, otherwise it's not readable on the side that's calling the function

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,330 @@
...
+    const targetRow = widgetRows[targetRowIndex!];
+
+    // Check if we're reordering within the same row
+    const isSameRowReorder =
+      sourceRow && targetRow && sourceRow.startIndex === targetRow.startIndex;
+
+    if (isSameRowReorder) {
+      // For same-row reordering, just preserve the existing widths
+      return { newWidgets, newWidths };
+    } else {
```
remidej: styling comment: no need for the to 'else' block since you have a return in the if

remidej: styling comment: no need for the to 'else' block since you have a return in the if

packages/core/admin/admin/src/features/Widgets.tsx
```diff
@@ -0,0 +1,330 @@
...
+    newRightWidth: number
+  ) => {
+    const newWidths = handleWidgetResize(
+      filteredWidgets,
+      columnWidths,
+      leftWidgetId,
+      rightWidgetId,
+      newLeftWidth,
+      newRightWidth
+    );
```
remidej: - both seem like functions that would be easier to use with a single options param object
- since the options type would overlap, you could define one of the types by extending the other. (`interface A extends B { /* the additional entries here */ }`)

packages/core/admin/admin/src/components/WidgetRoot.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+export interface WidgetRootProps
+  extends Pick<WidgetType, 'title' | 'icon' | 'permissions' | 'link' | 'uid'> {
+  children: React.ReactNode;
+  findWidget?: (id: string) => { index: number };
+  deleteWidget?: (id: string) => void;
+  onDragStart?: (widgetId: string) => void;
+  onDragEnd?: () => void;
+  component?: () => Promise<React.ComponentType>;
```
remidej: what's the component for?

packages/core/admin/admin/src/components/WidgetRoot.tsx
```diff
@@ -0,0 +1,200 @@
...
+  }
+`;
+
+export interface WidgetRootProps
+  extends Pick<WidgetType, 'title' | 'icon' | 'permissions' | 'link' | 'uid'> {
+  children: React.ReactNode;
+  findWidget?: (id: string) => { index: number };
+  deleteWidget?: (id: string) => void;
+  onDragStart?: (widgetId: string) => void;
+  onDragEnd?: () => void;
```
remidej: I think this place should extend (or pick/omit/any ts util) the types from Widgets.tsx, or vice versa. we should try to re-define types as little as possible, the types should flow from one place to the other

packages/core/admin/admin/src/components/WidgetRoot.tsx
```diff
@@ -0,0 +1,200 @@
...
+      gap={4}
+      padding={6}
+      position="relative"
+      aria-labelledby={uid}
+      tabIndex={0}
+      ref={(node: HTMLElement | null) => {
+        if (node) {
+          node.setAttribute('data-strapi-widget-id', uid);
+        }
+      }}
```
remidej: would this work as a prop on WidgetContainer directly? using a ref seems a bit overkill to set a data attribute

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,395 @@
...
+interface GapDropZoneManagerProps {
+  filteredWidgets: WidgetWithUID[];
+  columnWidths: Record<string, number>;
+  draggedWidgetId?: string;
+  moveWidget: (id: string, to: number, targetRowIndex?: number, isHorizontalDrop?: boolean) => void;
+}
+
+const getRowInfo = (row: WidgetRow) => {
+  const firstWidgetElement = document.querySelector(
+    `[data-strapi-widget-id="${row.widgets[0].uid}"]`
```
remidej: you may want to extract a constant for the data attribute name, or a util to generate the selector, because here if it was ever changed things would break

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,395 @@
...
+  height: number,
+  width: number = DROP_ZONE_SIZE,
+  targetRowIndex?: number
+): GapDropZonePosition => ({
+  insertIndex,
+  position: { left, top, height, width },
+  isVisible: true, // Will be controlled by parent
+  type: 'vertical',
+  targetRowIndex,
+});
```
remidej: does this need to be a function? i feel like it's as much effort to assemble the params it expects as it would be to build the object directly. And as I said before I'm not a big fan of functions with lots of params

packages/core/admin/admin/src/components/GapDropZone.tsx
```diff
@@ -0,0 +1,395 @@
...
+      rowTop,
+      rowHeight,
+      DROP_ZONE_SIZE,
+      rowIndex
+    )
+  );
+
+  // Add drop zones between widgets
+  for (let i = 0; i < widgetCount - 1; i++) {
+    const currentWidget = widgetPositions[i];
```
remidej: a forEach would match the codebase style better

packages/core/admin/admin/src/components/DragLayer.tsx
```diff
@@ -6,14 +6,29 @@ import { DragLayerMonitor, XYCoord, useDragLayer } from 'react-dnd';
 function getStyle(
   initialOffset: XYCoord | null,
   currentOffset: XYCoord | null,
-  mouseOffset: XYCoord | null
+  mouseOffset: XYCoord | null,
+  item?: any
```
remidej: a good pattern for this is to make `item` unknown type, then to create a [type guard function](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -130,13 +121,58 @@ const HomePageCE = () => {
...
+    } else {
+      // Set default layout when no custom layout exists
+      setFilteredWidgets(allAvailableWidgets);
+      setColumnWidths(createDefaultWidgetWidths(allAvailableWidgets));
+    }
+  }, [homepageLayout, allAvailableWidgets, setColumnWidths]);
+
+  const handleAddWidget = (widget: WidgetWithUID) => {
+    addWidget(widget);
+  };
```
remidej: it doesn't look like this abstraction is needed

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -152,34 +188,121 @@ const HomePageCE = () => {
...
+                )}
+              </Box>
             )}
           </Flex>
         </Layouts.Content>
+
+        {/* Add the DragLayer to handle custom drag previews */}
+        <DragLayer
+          renderItem={({ type, item }) => {
+            if (type === 'widget') {
```
remidej: a pattern we tend to prevent is to return early when a value is _not_ what you want. then you no longer need to wrap everything in an if block.

packages/core/admin/admin/src/pages/Home/HomePage.tsx
```diff
@@ -130,13 +121,58 @@ const HomePageCE = () => {
...
 
+  React.useEffect(() => {
+    if (allAvailableWidgets.length === 0) return;
+
+    // If user has customized the homepage layout, apply it
+    if (homepageLayout && homepageLayout.widgets) {
+      const { filteredWidgets, widths: homepageWidths } = applyHomepageLayout(
+        allAvailableWidgets,
+        homepageLayout
+      );
```
remidej, line 142: I'm surprised that we need our own sort of layout cache like this. I thought we'd be able to use the layout coming from redux toolkit, since it gets invalidated (and therefore refetched) on reorder/resize/etc. I guess waiting for the network roundtrip created a flicker in the UI?

mathildeleg: This is needed to display the widgets in the correct layout according to their order and widths. _homepageLayout_ from the backend only returns an array with the uid of the widget and its width, thus the _applyHomepageLayout_ returns the actual widget object (with title, icon, link etc), as well as filters through the availableWidgets array to check the widgets chosen by user to be shown are available; and sorts them according to the order the backend gives us. 
I hope it makes sense, let me know if you have any other questions though!

packages/core/admin/admin/src/components/WidgetRoot.tsx
mathildeleg: This is needed for the preview of the widget as it's being dragged

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/b7c9370448887abbf42a6b12d300de2668cda7a7

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/236036f57fc8612ea60dab5d24dcd4211c419f32

packages/core/admin/admin/src/components/WidgetRoot.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/4379035571fc1eda36f66a256f295ba8baa8b204

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/d4f933d4ebc5f78b3a302d60718fa34efd6dbacb

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/eb83eadb18839cb132fa7994dc52d802e08d193c

packages/core/admin/admin/src/pages/Home/HomePage.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/a6ad24794538441b35d40fc335a49562da8892e1

packages/core/admin/admin/src/pages/Home/HomePage.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/a6ad24794538441b35d40fc335a49562da8892e1

packages/core/admin/admin/src/components/GapDropZone.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/4d386d775b3404cdc3e4d3f639046af787d459b5

packages/core/admin/admin/src/components/DragLayer.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/98164abd6da5e9b4237b47936dcf0ec7a270ab0c
Let me know if this isn't right as I'm not 100% sure!

packages/core/admin/admin/src/components/WidgetRoot.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/54e054cb15f834d17d1d57589efe822db2cc7b47

packages/core/admin/admin/src/features/Widgets.tsx
mathildeleg: Fixed here: https://github.com/strapi/strapi/pull/24354/commits/29387f90c1220f5060eab736bacd056f61e034a2

packages/core/admin/admin/src/components/GapDropZone.tsx
markkaylor: Yes that is unfortunate :( but not your fault and not the end of the world

packages/core/admin/admin/src/components/WidgetRoot.tsx
remidej: I wonder if it's possible to use the children prop for it instead?

packages/core/admin/admin/src/components/DragLayer.tsx
remidej: looks perfect!

PR #24358, opened 2025-09-11
feat(homepage): add endpoints for widget move and resize actions

araksyagevorgyan:
### What does it do?

Adds backend support for persisting and retrieving the homepage layout (widget order and widths) per user.

### Why is it needed?

The admin homepage now allows dynamic widget resizing and reordering. Backend persistence ensures user preferences survive refreshes and sessions.

### How to test it?

1. Run the Strapi.
2. Log in with an authenticated admin user.
3. Call (you can use curl, Postman, or any other HTTP client):

#### GET  
```
GET /admin/homepage/layout
Authorization: Bearer <your-admin-token>

Response when layout exists:
{
  "version": 1,
  "widgets": [
    { "uid": "plugin::content-manager.last-published-entries", "width": 6 },
    { "uid": "plugin::content-manager.last-edited-entries", "width": 8 }
  ],
  "updatedAt": "2025-09-11T12:34:56.000Z"
}
Response when no layout exists yet:
null
```

#### PUT
```
PUT /admin/homepage/layout
Authorization: Bearer <your-admin-token>

Request body:
{
  "widgets": [
    { "uid": "plugin::content-manager.last-published-entries", "width": 6 },
    { "uid": "plugin::content-manager.last-edited-entries", "width": 8 }
  ]
}

Response:
{
  "version": 1,
  "widgets": [
    { "uid": "plugin::content-manager.last-published-entries", "width": 6 },
    { "uid": "plugin::content-manager.last-edited-entries", "width": 8 }
  ],
  "updatedAt": "2025-09-12T09:45:00.000Z"
}
```

Notes:
version (defaults to 1) and updatedAt (set automatically on the backend to the current datetime) are optional in the request.

### Related issue(s)/PR(s)

https://github.com/strapi/content-squad/issues/77

### DIFF
--- a/packages/core/admin/server/src/controllers/homepage.ts
+++ b/packages/core/admin/server/src/controllers/homepage.ts
@@ -1,6 +1,6 @@
 import type { Context } from 'koa';
 import { getService } from '../utils';
-import { UserLayout, UserLayoutWrite } from './validation/schema';
+import { HomepageLayout, HomepageLayoutWrite } from './validation/schema';
 
 export default {
   async getKeyStatistics(): Promise<{
@@ -9,22 +9,19 @@ export default {
     const homepageService = getService('homepage');
     return { data: await homepageService.getKeyStatistics() };
   },
-  async getUserLayout(ctx: Context): Promise<{ data: UserLayout } | null> {
+  async getHomepageLayout(ctx: Context): Promise<{ data: HomepageLayout | null }> {
     const homepageService = getService('homepage');
     const userId = ctx.state.user?.id;
-    if (!userId) ctx.throw(401, 'Unauthorized');
 
-    const data = await homepageService.getUserLayout(userId);
-    if (!data) return null;
+    const data = await homepageService.getHomepageLayout(userId);
     return { data };
   },
-  async updateUserLayout(ctx: Context): Promise<{ data: UserLayout }> {
+  async updateHomepageLayout(ctx: Context): Promise<{ data: HomepageLayout }> {
     const homepageService = getService('homepage');
     const userId = ctx.state.user?.id;
-    if (!userId) ctx.throw(401, 'Unauthorized');
 
-    const body = ctx.request.body as UserLayoutWrite;
-    const data = await homepageService.updateUserLayout(userId, body);
+    const body = ctx.request.body as HomepageLayoutWrite;
+    const data = await homepageService.updateHomepageLayout(userId, body);
     return { data };
   },
 };
--- a/packages/core/admin/server/src/controllers/homepage.ts
+++ b/packages/core/admin/server/src/controllers/homepage.ts
@@ -1,4 +1,6 @@
+import type { Context } from 'koa';
 import { getService } from '../utils';
+import { UserLayout, UserLayoutWrite } from './validation/schema';
 
 export default {
   async getKeyStatistics(): Promise<{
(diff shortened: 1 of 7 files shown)


packages/core/admin/server/src/controllers/validation/schema.ts
```diff
@@ -0,0 +1,32 @@
...
+export const UserLayoutWriteSchema = z.object({
+  order: z.array(WidgetUIDSchema).max(100).optional(),
+  widths: z.record(WidgetUIDSchema, WidthSchema).optional(),
+})
+  .strict()
+  .refine((v) => v.order || v.widths, {
+    message: "At least one of 'order' or 'widths' is required",
+  });
+
+export type UserLayoutWrite = z.infer<typeof UserLayoutWriteSchema>;
```
remidej: Unrelated but do you have prettier configured in your IDE? It should normally add an empty new line at the end of files with our config

packages/core/admin/server/src/controllers/homepage.ts
```diff
@@ -7,4 +9,22 @@ export default {
     const homepageService = getService('homepage');
     return { data: await homepageService.getKeyStatistics() };
   },
+  async getUserLayout(ctx: Context): Promise<{ data: UserLayout } | null> {
+    const homepageService = getService('homepage');
+    const userId = ctx.state.user?.id;
+    if (!userId) ctx.throw(401, 'Unauthorized');
```
remidej: I don't think this check is necessary here. You can trust the `admin::isAuthenticatedAdmin` you used when defining the route to keep away unauthenticated requests

packages/core/admin/server/src/controllers/homepage.ts
```diff
@@ -7,4 +9,22 @@ export default {
...
+
+    const data = await homepageService.getUserLayout(userId);
+    if (!data) return null;
+    return { data };
```
remidej: I would personally prefer to the object with data when it's null as well. I think having the same response shape in all cases makes things easier in the frontend

packages/core/admin/shared/contracts/homepage.ts
```diff
@@ -61,3 +61,36 @@ export declare namespace GetCountDocuments {
...
 }
+
+export declare namespace Homepage {
+  export type WidgetUID = string;
+  export type Width = 4 | 6 | 8 | 12;
+
+  export interface Layout {
+    version: number;
+    order: WidgetUID[];
+    widths: Record<WidgetUID, Width>;
```
remidej, line 72: This works well. But have you considered merging those 2 informations into the same structure?

It could be an array of objects: `Array<{ uid: WidgetUID; width: Width }>`

packages/core/admin/server/src/services/homepage.ts
```diff
@@ -32,7 +37,59 @@ export const homepageService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+    return UserLayoutSchema.parse(value);
+  };
+
+  const updateUserLayout = async (
+    userId: number,
+    input: unknown
+  ): Promise<UserLayout> => {
+    const write: UserLayoutWrite = UserLayoutWriteSchema.parse(input);
+
+    const store = await strapi.store(STORE);
```
remidej: No need to await here, strapi.store doesn't return a promise, only the methods of the store do.

Also, you can reuse the same store object for both of your methods if you define it outside of them. In that case you likely would not need the STORE constant anymore, you could inline it directly

remidej (commented):
A few comments but this looks good! You found all the different pieces that needed editing

remidej (approved):
Hey Araksya, sorry for reviewing it so late, I lost track of this PR. I rely on GitHub notifications a lot to know if something needs my attention, so don't hesitate to use the "re-request review" button.

The changes look good! Can't wait to see this hooked to the frontend

araksyagevorgyan:
@remidej 
Thanks for the feedback 👍
Could you please advise on the failing e2e tests in the pipeline? I tried re-running them but they failed again. How should I proceed in this case? 🤔

remidej:
The `create-relation-and-save.spec.ts` is a notoriously flaky one. The other one is odd — I don't see how it can be related to what you're doing. I tried re-starting the failing tests just in case.

If it still fails:
- check on the develop branch latest commits if it's a regression that's already been merged
- debug the failing test using https://trace.playwright.dev/ You can download the trace file inside the Run actions/upload-artifact@v4 toggle of the github action logs

hanpaine:
Hey @araksyagevorgyan we generally squash and merge commits, especially on big PRs as this helps with visibility of changes together. Please may you add the milestone it's merged too 🙏  thanks!

PR #24365, opened 2025-09-12
future(upload): generate image metadata on file upload

remidej:
### What does it do?

- moves the admin's getAiToken into a service (so we can retrieve the token from the upload plugin)
- create an aiMetadata service on the upload plugin with:
  - isEnabled method: check if feature should be on
  - processFiles method: takes files, sends them to the ai server, returns the metadata
- hooks the upload controller to this processFiles method.

### How to test it?

- Check that getAiToken is not broken. For this, clear session storage, refresh and open the CTB chat.
- Upload an image from the media library page. Then click on the pen icon to edit its metadata. You should see the AI-generated caption and alt text already there
- Upload multiple files at once. Some that are images, some that are not. Only the images should get AI metadata. Make sure the metadata is for the right images
 and do it again

### DIFF
--- a/packages/core/admin/server/src/controllers/authenticated-user.ts
+++ b/packages/core/admin/server/src/controllers/authenticated-user.ts
@@ -1,7 +1,4 @@
 import type { Context } from 'koa';
-import crypto from 'crypto';
-import fs from 'fs';
-import path from 'path';
 import type { AdminUser } from '../../../shared/contracts/shared';
 
 import { getService } from '../utils';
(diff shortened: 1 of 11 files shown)


packages/core/upload/server/src/services/upload.ts
```diff
@@ -223,7 +223,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       files,
     }: {
       data: Record<string, unknown>;
-      files: InputFile | InputFile[];
```
markkaylor: I'm not sure we should change the signature here. It might even be a breaking change.

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+        method: 'POST',
+        body: formData,
+        headers: {
+          Authorization: `Bearer ${token}`,
+        },
+      });
+
+      if (!res.ok) {
+        throw Error(`AI metadata generation failed`, { cause: await res.text() });
+      }
```
markkaylor, line 61: I don't think we want to throw. We will need a way to communicate to the user in the FE that it failed but otherwise AI generation failure should not block the upload. wdyt?

packages/core/admin/server/src/controllers/authenticated-user.ts
```diff
@@ -57,146 +54,23 @@ export default {
...
-        if (!data.jwt) {
-          strapi.log.error(`${ERROR_PREFIX} Invalid response: missing JWT token`);
-          return ctx.internalServerError(USER_ERROR_MESSAGE);
-        }
-
-        strapi.log.info('AI token generated successfully', {
-          userId: user.id,
-          expiresAt: data.expiresAt,
-        });
+      const tokenData = await userService.getAiToken(user);
```
markkaylor: I think this is throwing an error causing the build to fail.

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -0,0 +1,63 @@
+import type { Core } from '@strapi/types';
+import { readFile } from 'node:fs/promises';
+import type { File } from 'formidable';
```
markkaylor: This throws an error causing the build to fail, but I think you can use FileInfo from types which extends this 🤷

packages/core/upload/server/src/services/upload.ts
remidej: Indeed, and it looks like that was a mistake since I was still doing Array.isArray below

packages/core/upload/server/src/services/ai-metadata.ts
remidej: I think I do want to throw here, because generating the metadata is the this method's only job. It should not block upload though for sure. I'll make sure to catch it in the upload service

packages/core/admin/server/src/controllers/authenticated-user.ts
markkaylor, line 1: Should we let DX know what we are doing? Maybe just a heads up in their channel if you don't mind 🙏

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -0,0 +1,87 @@
...
+      // We need to maintain the original indices so we can map AI results back correctly
+      const imageFiles = files
+        .map((file, index) => ({ file, originalIndex: index }))
+        .filter(({ file }) => file.mimetype?.startsWith('image/'));
+
+      // If no image files, return sparse array with all nulls to avoid calling the AI server
+      // This maintains the same array length as input files for proper index alignment
+      if (imageFiles.length === 0) {
+        return new Array(files.length).fill(null);
+      }
```
markkaylor, line 36: ✨ nice!

PR #24367, opened 2025-09-12
[Session Manager] support origin configs and add fluent api

innerdvations:
### What does it do?

this refactors session manager to use an origin based fluent api:

```
strapi.sessionManager.defineOrigin('admin', { lifespansGoHere });

strapi.sessionManager('admin').validateAccessToken(token);

strapi.sessionManager('invalidOrigin') // throws error
```

and adds the origin definitions to the appropriate places, where they load their own configs

### Why is it needed?

- allows user code to use session manager to define their own origins
- cleaner session manager api

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/server/src/strategies/__tests__/admin.test.ts
+++ b/packages/core/admin/server/src/strategies/__tests__/admin.test.ts
@@ -26,10 +26,7 @@ describe('Admin Auth Strategy', () => {
       const generateUserAbility = jest.fn(() => 'ability');
 
       // Mock the new callable SessionManager API
-      const { sessionManager } = createMockSessionManager(
-        { validateAccessToken },
-        { isSessionActive }
-      );
+      const { sessionManager } = createMockSessionManager({ validateAccessToken, isSessionActive });
 
       global.strapi = {
         sessionManager: sessionManager as any,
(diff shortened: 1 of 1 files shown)


packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -88,17 +88,13 @@ module.exports = ({ strapi }) => ({
...
-          { familyType: 'refresh' }
-        );
-
-        const access = await strapi.sessionManager.generateAccessToken(
-          refresh.token,
-          'users-permissions'
-        );
+        const refresh = await strapi
+          .sessionManager('users-permissions')
+          .generateRefreshToken(String(user.id), undefined, { familyType: 'refresh' });
```
jhoward1994:
```suggestion
          .generateRefreshToken(String(user.id), undefined, { type: 'refresh' });
```
Given we removed everything to do with parent and family IDs

packages/core/admin/server/src/strategies/admin.ts
```diff
@@ -30,7 +30,7 @@ export const authenticate = async (ctx: Context) => {
     return { authenticated: false };
   }
 
-  const result = manager.validateAccessToken(token);
+  const result = manager('admin').validateAccessToken(token);
   if (!result.isValid) {
     return { authenticated: false };
   }
```
jhoward1994, line 36: Just below here L38 ?


```suggestion
  const isActive = await manager('admin').isSessionActive(result.payload.sessionId);

```

innerdvations: Oh, oops, I noticed there are actually still a bunch of methods in SessionManager that are missing from OriginSessionManager, lemme investigate and make sure everything has been moved correctly
edit: nope, i was wrong, it was just that one fortunately

PR #24369, opened 2025-09-12
[Session Manager] Add contributor docs and tweak configurations

innerdvations:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/core/admin/server/src/content-types/session.ts
```diff
@@ -1,71 +1,86 @@
 export default {
-  collectionName: 'admin_sessions',
+  collectionName: 'strapi_sessions',
```
innerdvations, line 2: `strapi_` is a guaranteed restricted prefix, I don't think `admin_` is. Also, this is generic sessions, not just admin, so this makes it more clear

PR #24370, opened 2025-09-12
[session manager] Use deprecated expiresIn as max lifespan

innerdvations:
### What does it do?

- deprecate expiresIn but use it as a fallback for the new max lifespan options

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/server/src/services/__tests__/token.test.ts
+++ b/packages/core/admin/server/src/services/__tests__/token.test.ts
@@ -1,5 +1,3 @@
-import { describe, test, expect, jest } from '@jest/globals';
-/* eslint-env jest */
 import {
   createJwtToken,
   getTokenOptions,


packages/core/admin/server/src/services/__tests__/token.test.ts
```diff
@@ -1,10 +1,54 @@
...
+    expect(expiresInToSeconds(undefined as any, secret)).toBeUndefined();
+    expect(expiresInToSeconds(null as any, secret)).toBeUndefined();
+  });
+
+  test('accepts numeric seconds', () => {
+    expect(expiresInToSeconds(3600, secret)).toBe(3600);
+  });
+
+  test('accepts numeric string seconds', () => {
+    expect(expiresInToSeconds('180', secret)).toBe(180);
```
jhoward1994: This is failing but I think maybe we should just remove this case? It doesn't seem to be covered in the logic

innerdvations: oops, that's what I get for not reviewing the ai code. fixing that and removing 'secret' from all of these methods

PR #24371, opened 2025-09-12
enhancement(session-manager): allow httpOnly option and remove familyType

innerdvations:
### What does it do?

- allow optionally setting httpOnly to avoid having that value returned and only use httpOnly cookie instead
  - this can be forced globally with up config `sessions.httpOnly:true` or if that is not set, on a per-reqeust basis setting the  header `x-strapi-refresh-cookie` = 'httpOnly'
- return refresh token when rotating it without httpOnly
- use 'type' instead of 'familyType'

### Why is it needed?

previously the refreshToken was not actually being sent again which was a bug
this way we send it, but also have a way for it to be more secure using httpOnly

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/plugins/users-permissions/server/controllers/auth.js
+++ b/packages/plugins/users-permissions/server/controllers/auth.js
@@ -280,7 +280,7 @@ module.exports = ({ strapi }) => ({
       return ctx.unauthorized('Invalid refresh token');
     }
 
-    return ctx.send({ jwt: result.token });
+    return ctx.send({ jwt: result.token, refreshToken: rotation.token });
   },
   async logout(ctx) {
     const mode = strapi.config.get('plugin::users-permissions.jwtManagement', 'legacy-support');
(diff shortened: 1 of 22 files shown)


packages/plugins/users-permissions/server/controllers/auth.js
```diff
@@ -88,9 +88,14 @@ module.exports = ({ strapi }) => ({
 
       const mode = strapi.config.get('plugin::users-permissions.jwtManagement', 'legacy-support');
       if (mode === 'refresh') {
+        const { deviceId } = ctx.request.body || {};
```
jhoward1994, line 91: Why would there ever be deviceId's in u&p requests?

innerdvations: It would be optional, so that they can do the same thing we're doing if they want and allow multiple logins that only logout one device.

PR #24382, opened 2025-09-15
feat(jwt): standardize configs and add tests

innerdvations:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/plugins/users-permissions/server/bootstrap/index.js
```diff
@@ -11,6 +11,13 @@ const crypto = require('crypto');
 const _ = require('lodash');
 const { getService } = require('../utils');
 const usersPermissionsActions = require('./users-permissions-actions');
+const {
+  DEFAULT_ACCESS_TOKEN_LIFESPAN,
+  DEFAULT_MAX_REFRESH_TOKEN_LIFESPAN,
+  DEFAULT_IDLE_REFRESH_TOKEN_LIFESPAN,
+  DEFAULT_MAX_SESSION_LIFESPAN,
+  DEFAULT_IDLE_SESSION_LIFESPAN,
+} = require('../services/constants');
```
Bassel17, line 20: I think you forgot to use the defaults in this file

Bassel17 (changes requested):

packages/plugins/users-permissions/server/config.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+const {
+  DEFAULT_ACCESS_TOKEN_LIFESPAN,
+  DEFAULT_MAX_REFRESH_TOKEN_LIFESPAN,
+  DEFAULT_IDLE_REFRESH_TOKEN_LIFESPAN,
+  DEFAULT_MAX_SESSION_LIFESPAN,
+  DEFAULT_IDLE_SESSION_LIFESPAN,
+} = require('./services/constants');
```
Bassel17, line 9: same in this file

packages/utils/api-tests/agent.js
```diff
@@ -13,19 +13,43 @@ const createAgent = (strapi, initialState = {}) => {
...
+    const rq = supertestAgent[method.toLowerCase()](fullUrl);
+
+    // Apply authentication and headers on the request itself
     if (has('token', state)) {
-      supertestAgent.auth(state.token, { type: 'bearer' });
+      rq.auth(state.token, { type: 'bearer' });
     }
     if (headers) {
-      supertestAgent.set(headers);
+      const authHeader = headers.Authorization || headers.authorization;
```
jhoward1994:
```suggestion
      applyAuthAndHeaders(rq, headers);
```

It might be worth lifting this out to a function ? Here and L38 look like the same block ?

packages/plugins/users-permissions/server/bootstrap/index.js
innerdvations: 🤦

packages/plugins/users-permissions/server/config.js
```diff
@@ -11,19 +19,14 @@ module.exports = {
...
-        name: env('UP_SESSIONS_COOKIE_NAME', 'strapi_up_refresh'),
-        sameSite: env('UP_SESSIONS_COOKIE_SAMESITE', 'lax'),
-        path: env('UP_SESSIONS_COOKIE_PATH', '/'),
-        domain: env('UP_SESSIONS_COOKIE_DOMAIN'),
-        secure: env.bool('UP_SESSIONS_COOKIE_SECURE', process.env.NODE_ENV === 'production'),
-      },
+      accessTokenLifespan: DEFAULT_ACCESS_TOKEN_LIFESPAN,
+      maxRefreshTokenLifespan: DEFAULT_MAX_REFRESH_TOKEN_LIFESPAN,
+      idleRefreshTokenLifespan: DEFAULT_IDLE_REFRESH_TOKEN_LIFESPAN,
+      maxSessionLifespan: DEFAULT_MAX_SESSION_LIFESPAN,
```
innerdvations, line 27: I removed the use of env here because we don't want to add new hardcoded ENV var in strapi code. It should be up to the user to set it in their config.

I left the legacy values alone though, we will still accept JWT_SECRET

PR #24387, opened 2025-09-16
fix: orphaned relation nested compo

Bassel17:
### What does it do?

- loops through components that have components as parents until we find a contentType parent

### Why is it needed?

- resolves ghost relations issue for nested components


jhoward1994 (approved):

laurenskling:
Thank you so much for this

PR #24388, opened 2025-09-16
[Session Manager] Fix SSO

jhoward1994:
### What does it do?

- Fixes **SSO login flow** using the new session manager:
- Adds a **shared session/auth utils module**:

### Why is it needed?

- SSO previously didn’t create a proper session/refresh token

### How to test it?

1. Configure an SSO provider (e.g., GitHub/Google).
2. Login via SSO → check cookies:
   - `strapi_admin_refresh` is present, with correct expiry (absolute/idle or session).
   - `jwtToken` is present and usable in the Admin UI.
3. Verify refresh token rotation and logout clear cookies correctly.

### DIFF
--- a/packages/core/admin/shared/utils/session-auth.ts
+++ b/packages/core/admin/shared/utils/session-auth.ts
@@ -1,0 +1,97 @@
+import crypto from 'crypto';
+import type { Modules } from '@strapi/types';
+
+export const REFRESH_COOKIE_NAME = 'strapi_admin_refresh';
+
+const DEFAULT_ACCESS_TOKEN_LIFESPAN = 10 * 60;
+const DEFAULT_MAX_REFRESH_TOKEN_LIFESPAN = 30 * 24 * 60 * 60;
+const DEFAULT_IDLE_REFRESH_TOKEN_LIFESPAN = 14 * 24 * 60 * 60;
+const DEFAULT_MAX_SESSION_LIFESPAN = 1 * 24 * 60 * 60;
+const DEFAULT_IDLE_SESSION_LIFESPAN = 2 * 60 * 60;
+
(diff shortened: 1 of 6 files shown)


packages/core/admin/shared/utils/session-auth.ts
```diff
@@ -0,0 +1,97 @@
+import crypto from 'crypto';
+import type { Modules } from '@strapi/types';
+
+export const REFRESH_COOKIE_NAME = 'strapi_admin_refresh';
+
+const DEFAULT_ACCESS_TOKEN_LIFESPAN = 10 * 60;
```
innerdvations: This may have revealed an existing issue; DEFAULT_ACCESS_TOKEN_LIFESPAN is not used.

But we may also still want to export these, in case someone needs access to them from outside. It might be useful to know the defaults.

jhoward1994:
New experimental from this branch 

0.0.0-experimental.8a41bddd1fbbac61aa43a961ea1ff74ac586cf55

@derrickmehaffy

derrickmehaffy:
> New experimental from this branch
> 
> 0.0.0-experimental.8a41bddd1fbbac61aa43a961ea1ff74ac586cf55
> 
> @derrickmehaffy

Thanks, I'll run some tests today

derrickmehaffy:
Manual testing notes:

- When signing out of an SSO user (will test this also on normal PR later)
  - `logged_in` cookie is not cleared or set to false
  - `jwtToken.sig` cookie is not cleared (not sure if this is intended or not)
  
Can confirm that SSO logins are now functioning properly (only tested the GitHub provider but it shouldn't matter)

derrickmehaffy (approved):
SSO is working

innerdvations (approved):

PR #24409, opened 2025-09-18
[Session manager] QA

jhoward1994:
To address:

https://github.com/strapi/strapi/pull/24346#issuecomment-3303977216

- We should have deleting users delete their sessions
- Changing a password should also delete the session


packages/core/admin/server/src/controllers/authentication.ts
```diff
@@ -265,6 +265,9 @@ export default {
       const userId = String(user.id);
       const deviceId = generateDeviceId();
 
+      // Invalidate all existing sessions before creating a new one
+      await sessionManager('admin').invalidateRefreshToken(userId);
```
innerdvations, line 269: this would prevent logging in on multiple devices, wouldn't it?

jhoward1994: Why do you say that ?

The logic I'm going for here is: when resetting a users password, do the password change, invalidate their sessions everywhere (e.g. provide no deviceId here. I think it's logical to invalidate everywhere as their password has changed?) and then generate a new access token

derrickmehaffy (approved):

packages/core/admin/server/src/controllers/authentication.ts
innerdvations: oh sorry, I was reading it wrong, I thought this was somewhere else. Carry on 😆

PR #24411, opened 2025-09-18
feat: guided tour for ai ctb

Adzouz:
### What does it do?

For the new AI chatbot in the Content Type Builder, that helps the user creating schemas based on their prompt, we needed to add specific steps for the new guided tour.

Screenshot 2025-09-18 at 14 00 18

### Why is it needed?

The goal is to show and explain to the user the role of the chatbot when AI features are enabled on Strapi.

### How to test it?

- Enable AI features on your Strapi project (need to have a valid Enterprise Edition licence)
- Set the `STRAPI_ADMIN_AI_URL` is the .env (you'll need the API to be running)
- Be sure that the Guided Tour is reset (you can do it in the Profile settings)
- Go to the Content Type Builder
Screenshot 2025-09-18 at 14 01 35
-> You should have the AI chatbot button on the bottom right of the screen. If not check the configuration on how to enable the AI features again.

- Click on Next on the first step (should be identical to the one when AI is disabled)
- The tooltip regarding the AI chatbot should be displayed and the chatbot should be opened automatically
- Enter a prompt to generate a model
- If the model has also fields in it, it will automatically show the next tooltip that tells the user to Save their changes
- If not, ask the prompt to create a new field in the new schema to display the tooltip
- After clicking on Next and saving, the Finish tooltip should be displayed

-> There are only 3 steps + Finish on this tour
-> If a tour was already ongoing and the AI features are enabled/disabled after, then the tour should be reset to its first step

### Related issue(s)/PR(s)

Related ticket: https://github.com/strapi/content-squad/issues/78

🚀


packages/core/admin/admin/src/components/GuidedTour/Steps/Step.tsx
```diff
@@ -231,21 +241,27 @@ const createStepComponents = (tourName: ValidTourName): Step => ({
...
+        defaultMessage: props.defaultMessage,
+      });
+    }
+    return (
+      <Box paddingBottom={5} paddingLeft={5} paddingRight={5} width="100%">
+        {'children' in props ? (
+          props.children
+        ) : (
+          <ContentContainer>
+            <Typography tag="div" variant="omega" dangerouslySetInnerHTML={{ __html: content }} />
```
Adzouz, line 259: I needed to do that to interpret HTML in the tooltip. Don't know if you have a better solution?

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -57,9 +57,19 @@ type Action =
   | {
       type: 'remove_completed_action';
       payload: ValueOf<CompletedActions>;
+    }
+  | {
+      type: 'set_tour_type';
```
Adzouz, line 62: I needed to add this in order to be able to check if the user has enabled the AI feature after starting the tour to reset it.

tests/e2e/tests/content-type-builder/guided-tour.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    const gotItButton = page.getByRole('button', { name: 'Got it' });
+    await expect(
+      page.getByRole('dialog', { name: 'Welcome to the Content-Type Builder!' })
+    ).toBeVisible();
+    await nextButton.click();
+    await expect(page.getByRole('dialog', { name: 'Time to get started!' })).toBeVisible();
+    await nextButton.click();
+
+    // Check if the AI chat is opened
+    await expect(page.getByRole('textbox', { name: 'Ask Strapi AI...' })).toBeVisible();
```
Adzouz, line 36: I couldn't go any further since I don't really know how to mock the event stream from the API request yet... And DX team hasn't found a solution to test it for the moment.

packages/core/admin/admin/src/components/GuidedTour/Steps/Step.tsx
markkaylor: Maybe I'm missing something but couldn't you just put the markup here directly by providing children instead of injecting it into typopraphy 🤔 ?

```
<p>
  <Typography>translated text</Typography>
</p>
```

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
```diff
@@ -57,9 +57,19 @@ type Action =
...
+      payload: {
+        tourName: ValidTourName;
+        tourType: string;
```
markkaylor: Any string? what about a string union? I assume there's only a couple options right?

packages/core/admin/admin/src/hooks/useAIAvailability.ts
markkaylor, line 1: I'd just run this by the DX squad to make sure they are aligned. Also maybe put an `@internal` since this exposed now

markkaylor (commented):
Looks quite good! Nice work. I won't have time to test but I left some thoughts to consider.

tests/e2e/tests/content-type-builder/guided-tour.spec.ts
soupette: Have you tried msw ? https://mswjs.io/docs/http/mocking-responses/streaming/

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentTypeBuilderSteps.tsx
```diff
@@ -84,8 +209,14 @@ const AddFields = ({ Step, dispatch }: StepContentProps) => (
       defaultMessage="Add the fields your content needs such as text, media and relations."
     />
     <Step.Actions>
-      <StepCount tourName="contentTypeBuilder" />
-      <GotItAction onClick={() => dispatch({ type: 'next_step', payload: 'contentTypeBuilder' })} />
+      <ContentTypeBuilderActions
+        showPrevious
+        CustomActions={
```
remidej: I thin instead of a CustomActions prop, you could rely on optional children. This is how Step.Actions already works so it would be more consistent

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
remidej: +1 for a union, reading this I don't know what the tour types can be

packages/core/admin/admin/src/translations/en.json
```diff
@@ -805,6 +805,8 @@
   "submit": "Submit",
   "tours.contentTypeBuilder.Introduction.title": "Welcome to the Content-Type Builder!",
   "tours.contentTypeBuilder.Introduction.content": "Here you create and manage the structure of your app with collection types, single types, and reusable components. Let's dive in!",
+  "tours.contentTypeBuilder.AIChat.title": "Time to get started!",
+  "tours.contentTypeBuilder.AIChat.content": "<p>If you have any questions about the Content-Type Builder or Strapi ask them here.</p><p>Strapi AI can generate schemas tailored to your needs. Ask for exactly what you want, for example:<ul><li>Date picker</li><li>Email and password fields</li><li>Media of any type</li><li>UIDs</li></ul></p><p>Don’t be shy, try it out !</p>",
```
remidej, line 809: Design feedback: we already have action suggestions within the AI chat modal, and they have the benefit of being clickable. Having suggestions in this popover too seems redundant to me

packages/core/admin/admin/src/index.ts
```diff
@@ -60,6 +60,7 @@ export { useClipboard } from './hooks/useClipboard';
 export { useElementOnScreen } from './hooks/useElementOnScreen';
 export { useAdminUsers } from './services/users';
 export { useGetCountDocumentsQuery } from './services/homepage';
+export { useAIAvailability } from './hooks/useAIAvailability';
```
remidej, line 64: thanks this will be handy for the media lib and i18n too :+1:

packages/core/content-type-builder/admin/src/components/AIChat/Chat.tsx
```diff
@@ -350,6 +350,25 @@ const Chat = () => {
...
   const { attachFiles } = useAttachments();
   const { t } = useTranslations();
+  const state = useGuidedTour('Chat', (s) => s.state);
+
+  // Auto-open chat when AIChat guided tour step is active
+  useEffect(() => {
+    const isAIChatStepActive =
+      state.tours.contentTypeBuilder?.currentStep === 1 && // AIChat is step 1 in contentTypeBuilderStepsAI
+      !state.tours.contentTypeBuilder?.isCompleted &&
+      state.enabled;
```
remidej: nitpick but I would put the enabled check first. because i think most times this run it will be false (as they'll be onboarded already)

packages/core/admin/admin/src/components/GuidedTour/Steps/Step.tsx
Adzouz: There are many `<p>` and also some `<ul><li>` in the translated string directly and that's what I want to be interpreted so I couldn't find a better way to do it 😕

tests/e2e/tests/content-type-builder/guided-tour.spec.ts
Adzouz: I tried but couldn't make it work with Playwright since the `route.fulfill` doesn't support ReadableStream...

packages/core/admin/admin/src/components/GuidedTour/Context.tsx
Adzouz: For the moment I'm only using it to differenciate the ai-ctb tour from the non-ai one so I can fix the two values there but I wanted it to be more flexible if we want to integrate it for another tour in the future. Are you sure you want to put values related to the CTB guided tour only here? It would be `"ContentTypeBuilderNoAI" | "ContentTypeBuilderAI"`

packages/core/admin/admin/src/translations/en.json
Adzouz: I asked Louis for his opinion. Waiting for him to get back to me, I'll let you know.

Adzouz: We will keep the wording as-is after discussion with UX since we suggest other types of parameters to give to the prompt.

packages/core/admin/admin/src/hooks/useAIAvailability.ts
Adzouz: All good with them!

tests/e2e/tests/content-type-builder/guided-tour.spec.ts
soupette: You would need to setup msw directly in the app that is used to run the tests for this and go without playwright for mocking. 
https://github.com/mswjs/playwright

soupette: So in fact you could intercept the api directly in your test, but the real question is to understand to know if it's useful for this test (the guided tour) or for the ones where the ai manages the ctb...

Adzouz: That's a good point actually, I do think it's more on their side in order to test that the answer from the prompt is doing correctly its stuff but I wanted to fake it to be sure that the next step of the guided tour is correctly displayed after using the chat. Since it has to be released next week, I think we'll keep it as-is for the moment. I'll raise the concern and the way to test the calls to the dedicated team so we can add specific tests for the chat!
Thanks @soupette 🙏

packages/core/admin/admin/src/components/GuidedTour/Steps/ContentTypeBuilderSteps.tsx
Adzouz: Done ✅

packages/core/content-type-builder/admin/src/components/AIChat/Chat.tsx
Adzouz: Done ✅

PR #24419, opened 2025-09-19
proof of concept: soft reset after ctb changes

innerdvations:
### What does it do?

Just an experiment to eliminate the server restarts after CTB changes, something to look into for Strapi 6

- Breaking change: user code and plugins that perform any setup in bootstrap or init relying on routes or content types will not work until they add a 'softReset' method that will be called to inform them of updates

This current implementation "somewhat properly" handles the database sync, rebuilding routes and content types, adds blocking to avoid race conditions while routes are being rebuilt, and handles softReset in graphql. basically enough to more-or-less work on our end (with some hacky code), but would wreak havoc on a real system with plugins 😆 There may also be some issues with middleware.

### Why is it needed?

Reduces the lengthy server restart to under 1 second, eliminating the need for a front-end "please wait..." screen

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
Closing this for now since it wouldn't be relevant until Strapi 6

PR #24423, opened 2025-09-19
fix: unique index recreation

### LINKED ISSUE #24421: Unique indexes are recreated at startup when an unrelated migration is done
### Node Version

20.19.4

### NPM/Yarn/PNPM Version

npm 10.8.2

### Strapi Version

4.25.23

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

https://github.com/AlbertoCalvoRubio/my-strapi-project

### Bug Description

When a simple migration is done (change one attribute type, e.g title from string to text), Strapi detects that some unique indexes have changed when they have not.
This is happening when comparing the old and new schemas. For example, for a table `banners_components`, when the indexes are compared, in the index `banners_unique`, the columns of the previous index are `[
  "entity_id",  "component_id",  "component_type",  "field"]` and for the new index the columns are `[  "entity_id",   "component_id",  "field",  "component_type"]`. The order is changing and the comparison is done by this line of code `!_.isEqual(oldIndex.columns, index.columns)` where order is important.

This results in dropping and recreating the unique indexes, which impacts performance at startup for large tables.

### Steps to Reproduce

1. Create a Strapi project with some content types and add components to them.
2. Run Strapi so that the schema is created.
3. Edit the schema.json of one of the content types, changing a `string` attribute to be of type `text`.
4. Run `npm run start` with database config `debug` enabled and see the drop + create indexes (it can also be seen with the debugger, e.g vscode, adding a breakpoint in `node_modules/@strapi/database/dist/index.js` line 1310).

### Expected Behavior

Indexes are not recreated when there are no changes related to them.

### Logs

```shell
[
  {
    sql: 'alter table "public"."banners_components" drop constraint "banners_unique"',
    bindings: []
  },
  {
    sql: 'alter table "public"."banners_components" add constraint "banners_unique" unique ("entity_id", "component_id", "field", "component_type")',
    bindings: []
  }
]
[
  {
    sql: 'alter table "public"."ctas_components" drop constraint "ctas_unique"',
    bindings: []
  },
  {
    sql: 'alter table "public"."ctas_components" add constraint "ctas_unique" unique ("entity_id", "component_id", "field", "component_type")',
    bindings: []
  }
]
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

After investigating the @strapi/database package, I believe that the bug might also come from comparing the indexes from the previous schema, read from `db.dialect.schemaInspector.getSchema()`, which reads the schema from `information_schema` and other pg tables instead of the one already stored in the table `strapi_database_schema`.
This could be fixed by either one of the following:
- By using the old schema already stored in `strapi_database_schema`, if there is no specific reason to read the schema from PostgreSQL internal tables.
- Compare the index columns without comparing the position of the elements in the arrays.

### Confirmation Checklist

AlbertoCalvoRubio:
### What does it do?

Fix #24421 

### Why is it needed?

- It uses the previous schema stored in the `strapi_database_schema` table to compare changes to not recreate indexes as the order of the columns array is the same. It also avoids doing a lot of queries to the database to know the current schema if it was already stored in the `strapi_database_schema` table.

### How to test it?

- It can be reproduced by following the steps in issue #24421 

### Related issue(s)/PR(s)

- #24421

### DIFF
--- a/packages/core/database/src/schema/index.ts
+++ b/packages/core/database/src/schema/index.ts
@@ -6,7 +6,7 @@ import createSchemaStorage from './storage';
 import { metadataToSchema } from './schema';
 
 import type { Database } from '..';
-import { SchemaDiff } from './types';
+import { SchemaDiff, Schema } from './types';
 
 export type * from './types';
 
@@ -17,7 +17,7 @@ export interface SchemaProvider {
   schemaDiff: ReturnType<typeof createSchemaDiff>;
   schemaStorage: ReturnType<typeof createSchemaStorage>;
   sync(): Promise<SchemaDiff['status']>;
-  syncSchema(): Promise<SchemaDiff['status']>;
+  syncSchema(schema?: Schema): Promise<SchemaDiff['status']>;
   reset(): Promise<void>;
   create(): Promise<void>;
   drop(): Promise<void>;
@@ -61,12 +61,12 @@ export const createSchemaProvider = (db: Database): SchemaProvider => {
       await this.create();
     },
 
-    async syncSchema(): Promise<SchemaDiff['status']> {
+    async syncSchema(oldSchema?: Schema): Promise<SchemaDiff['status']> {
       debug('Synchronizing database schema');
 
-      const DBSchema = await db.dialect.schemaInspector.getSchema();
+      const currentSchema = oldSchema ?? (await db.dialect.schemaInspector.getSchema());
 
-      const { status, diff } = await this.schemaDiff.diff(DBSchema, schema);
+      const { status, diff } = await this.schemaDiff.diff(currentSchema, schema);
 
       if (status === 'CHANGED') {
         await this.builder.updateSchema(diff);
@@ -101,7 +101,7 @@ export const createSchemaProvider = (db: Database): SchemaProvider => {
       if (oldHash !== hash) {
         debug('Schema changed');
 
-        return this.syncSchema();
+        return this.syncSchema(oldSchema.schema);
       }
 
       debug('Schema unchanged');


AlbertoCalvoRubio:
> Hi, thanks for the PR.
> 
> Unfortunately we can't go with this approach quiet yet, we intentionnaly re-inspect the schema to manage cases where users manually add things in there DB or run operations outside Strapi.
> 
> If you added an option to switch this behavior on/off that would allows us to move forward with this :)

I've added a `strictSynchSchema` database setting, similar to `forceMigration` or `runMigration`. It is enabled by default (current behaviour), and it can be disabled. 
Also, I noticed that comparing the schemas, being both created from the Strapi metadata, needed a fix when comparing column types.
Let me know what you think of this approach, and then I will document this new option in the @strapi/documentation repository.

alexandrebodin:
@AlbertoCalvoRubio Thanks that looks pretty good to me, nice catch on the bug. 

@derrickmehaffy could you give this a try when you can ? I think this is a nice addition that we could make the default in the next major to reduce bootstrap time :)

derrickmehaffy:
> @AlbertoCalvoRubio Thanks that looks pretty good to me, nice catch on the bug. 
> 
> @derrickmehaffy could you give this a try when you can ? I think this is a nice addition that we could make the default in the next major to reduce bootstrap time :)

Should we also port this to v5? Atm it targets v4

derrickmehaffy:
Hey Alberto,

Can you quickly run prettier against this so that test can pass? Likewise would you be willing to open a PR for the documentation to add the option? 

Specifically for the docs, the repo/branch for v4 is here: https://github.com/strapi/documentation/tree/v4

alexandrebodin:
> [Member ](https://github.com/strapi/strapi/commits?author=derrickmehaffy)
> ###

We pbl have the same thing in v5 I don't recall us changing it so if you can make a PR against develop after that one get's merged that would be 💯

AlbertoCalvoRubio:
> Hey Alberto,
> 
> Can you quickly run prettier against this so that test can pass? Likewise would you be willing to open a PR for the documentation to add the option?
> 
> Specifically for the docs, the repo/branch for v4 is here: https://github.com/strapi/documentation/tree/v4

I've fixed the formatting issue and created a PR to the documentation repository https://github.com/strapi/documentation/pull/2724.

AlbertoCalvoRubio:
> > [Member ](https://github.com/strapi/strapi/commits?author=derrickmehaffy)
> 
> We pbl have the same thing in v5 I don't recall us changing it so if you can make a PR against develop after that one get's merged that would be 💯

Sure, I will do the same for v5 once this gets merged. Thanks for the quick feedback!

PR #24427, opened 2025-09-19
poc: custom blocks implementation

remidej:
### What does it do?

- adds a new custom blocks API to the plugin API
- creates a sample "callout" block in getstarted to try out the API

TODO

- replace yup by zod for the validations, since it will be exposed to the users and we're moving everything to zod already
- probably a fix needed in the React renderer repo since custom blocks would not match the BlocksContent TS type
- add unit tests

### Why is it needed?

https://feedback.strapi.io/customization/p/add-ability-to-extend-strapis-rich-text-editor-with-custom-slate-elements
### How to test it?

Open a kitchensink entry in getstarted. It has a blocks editor. Use the callout block.

### DIFF
--- a/examples/getstarted/src/admin/app.jsx
+++ b/examples/getstarted/src/admin/app.jsx
@@ -1,16 +1,167 @@
 import React from 'react';
+import styled from 'styled-components';
+import { Editor, Transforms } from 'slate';
 
-import { Button } from '@strapi/design-system';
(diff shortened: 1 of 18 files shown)


markkaylor:
Demo

https://github.com/user-attachments/assets/62efff87-4e61-4c78-9f85-ac45064e6fe2

PR #24435, opened 2025-09-23
feat: add ai data usage to settings overview

Bassel17:
### What does it do?

- Adds AI usage data endpoint to the cms
- Adds usage data to the settings view
- Adds a global notification system for ai usage
- refactors ai features into Admin EE

### How to test it?

- User a growth EE license, verify that AI features work

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/AIChat/hooks/useAIFetch.ts
+++ b/packages/core/content-type-builder/admin/src/components/AIChat/hooks/useAIFetch.ts
@@ -179,13 +179,14 @@ export const createAIFetchHook = <T extends keyof AIEndpoints>(endpoint: T) => {
           ctx: { strapiVersion, projectId, userId },
         });
 
+        // refetch ai usage data on every successful request
+        aiUsage.refetch();
+
         const body = await safeParseJson(response);
 
         if (!response.ok) {
           throw new Error(`Error: ${response.statusText}`);
         }
-        // refetch ai usage data on every successful request
-        aiUsage.refetch();
         return body as ResponseType<T>;
       } catch (err) {
         setError(err instanceof Error ? err.message : `Failed to fetch data from ${endpoint}`);


packages/core/admin/admin/src/pages/Settings/pages/ApplicationInfo/ApplicationInfoPage.tsx
```diff
@@ -95,6 +111,10 @@ const ApplicationInfoPage = () => {
     return null;
   }
 
+  if (!AIUsageData) {
+    return null;
```
jhoward1994, line 115: Not sure but could this mean the whole compo returns null in CE? As <AIUageDataCE> is null ?

packages/core/admin/ee/server/src/controllers/admin.ts
```diff
@@ -66,4 +70,128 @@ export default {
...
+    }
+
+    if (!eeLicense) {
+      strapi.log.error(
+        `${ERROR_PREFIX} No EE license found. Please ensure STRAPI_LICENSE environment variable is set or license.txt file exists.`
+      );
+      return ctx.internalServerError(USER_ERROR_MESSAGE);
+    }
+
+    const aiServerUrl = process.env.STRAPI_ADMIN_AI_URL || process.env.STRAPI_AI_URL;
```
jhoward1994: Minor thing, how come we maintain 2 env vars still?

Bassel17: Honestly have no idea why we have two in the first place

packages/core/admin/admin/src/pages/Settings/pages/ApplicationInfo/ApplicationInfoPage.tsx
Bassel17: this ensures we have a component in AIUsageData before the component is rendered,
in CE we will also get a component just one that returns null.

packages/core/admin/ee/server/src/controllers/admin.ts
innerdvations: Yeah, we should only have one (even though it's internaluse only, still confusing for us 😆)

packages/core/admin/ee/admin/src/hooks/useAIUsageWarning.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { useAIAvailability } from './useAIAvailability';
+
+/**
+ * Triggers a warning notification if AI usage is above a threshold (default 80%).
+ * @param threshold - Usage percentage (0-1) at which to warn. Default: 0.8 (80%)
+ */
+export function useAIUsageWarning(threshold: number = 0.8) {
+  const location = useLocation();
+  const isAuthPage = location.pathname.startsWith('/auth');
```
jhoward1994, line 16: Wondering if this is enough of an exclusion for this notification. Would it make
more sense to only show it in the CTB ?

Bassel17: we pretty much want it to be shown anywhere, when a user is loggedin, 
the usage itself can be done by someone else, and others admin users need to be notified regardless.
we also have to take into consideration the media ai stuff and translation ai stuff that are coming they all are in same billing and all used in different places

innerdvations (approved):
It's been tested and we have the known bugs filed, let's merge this and create new PRs for the bugs

PR #24436, opened 2025-09-23
enhancement: moving ai endpoints to EE and conditionally add them if ai is enabled

Bassel17:
### What does it do?

- moves ai endpoints to EE because it's an EE feature
- conditionally adds the endpoints if the feature is enabled

### Why is it needed?

- we need it for a cleaner setup, and not all EE plans have AI 
- users can disable AI

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
Sorry if I missed it or if it's covered in existing code, but I didn't want to forget:  we also need to check that admin config `ai.enabled` is not set to false (it defaults to true) in case they've turned AI off locally.

jhoward1994 (approved):

PR #24454, opened 2025-09-25
fix: assert admin.auth.secret on bootstrap instead of init

innerdvations:
# Note: this is a hotfix branch

### What does it do?

Moves the assertion for admin.auth.secret to bootstrap instead of init

### Why is it needed?

init is run on build, and the secret is not actually required for build. this causes errors on workflows where different env vars are passed in to build.

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/providers/index.ts
+++ b/packages/core/core/src/providers/index.ts
@@ -2,7 +2,7 @@ import admin from './admin';
 import coreStore from './coreStore';
 import cron from './cron';
 import registries from './registries';
-import sessionManager from './sessionManager';
+import sessionManager from './session-manager';
 import telemetry from './telemetry';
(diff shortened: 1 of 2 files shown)


innerdvations:
I'm actually not sure if we need this check anyway, since it's already required in a couple other places, but I suppose it's worth keeping since it's also required here.

alexandrebodin (approved):

jhoward1994 (approved):
Tested on Strapi Cloud with experimental `0.0.0-experimental.a1b9cf7971f495cb61c820da5fd4719066894c38`

& recreated crisis bug before this fix from experimental `0.0.0-experimental.d23f1485e5669cca1cee1bf5f8168b6b8109b78f`

Issue #24457, opened 2025-09-25
Issue after updating from 5.23.6 → 5.24.1 with RS512 JWT


GolfNa-BGJV:
### Node Version

22.17.1

### NPM/Yarn/PNPM Version

NPM 10.9.2

### Strapi Version

5.24.1

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

I ran into an issue after updating from version 5.23.6 to 5.24.1.
I’m using the RS512 algorithm to sign JWTs, and everything was working fine before.

However, after the update, when I try to log in, I get this error:

`[2025-09-25 22:22:04.981] error: Failed to create admin refresh session secretOrPrivateKey must be a symmetric key when using HS256
[2025-09-25 22:22:04.981] http: POST /admin/login (90 ms) 500`

It looks like the system is suddenly expecting HS256 even though my config is set to RS512.

Has something changed in version 5.24.1 regarding JWT algorithms or admin refresh sessions?
Do I need to adjust my configuration?

Thanks in advance! 🙏

### Steps to Reproduce

Upgrade working strapi project from 5.23.6 to 5.24.1 version.
Run strapi by "npm run start"
Try to login to admin panel, and get 500 error.

### Expected Behavior

I expected the application to continue working with the RS512 algorithm for JWT signing

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

edu-amr:
we detected a slowness in our pages even with cash and we tried to access strapi and detected slowness, and we paid an absurd 375 dollars for a slow piece of crap

innerdvations:
Hello and sorry for the issue! Just so I can be completely sure we fix this correctly and as quickly as possible, can you please post your exact configuration (redacting any keys of course)?

hanpaine:
Hey @GolfNa-BGJV if you are able to share the above, it would help us fix this quickly please 🙏 thanks!

[...]

PR #24460, opened 2025-09-26
fix: support auth.options config in sessions

innerdvations:
### What does it do?

This PR restores support for the legacy admin.auth.options configuration while introducing the new
  admin.auth.sessions.options configuration pattern. It ensures that existing JWT configurations using
  custom algorithms (like RS512) continue to work after the recent JWT enhancements in v5.24.1.

### Why is it needed?

Removing it was an unintentional breaking change

This allows user configs to pass in options directly to the jwt methods (`sign`, `decode`, etc) as was done before the recent JWT enhancements

### How to test it?

1. Test legacy configuration: Set admin.auth.options with asymmetric algorithms (RS256, RS512, ES256)
   and verify JWT generation/validation works
2. Test new configuration: Use admin.auth.sessions.options with both symmetric and asymmetric
algorithms
3. Test precedence: When both are configured, admin.auth.sessions.options should take priority
4. Run tests: The included test suites cover symmetric/asymmetric algorithms and option
merging

Admin config from reported issue https://github.com/strapi/strapi/issues/24457#issuecomment-3338729639

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/24457

#### Option 1: Legacy Configuration (admin.auth.options)

Add this to your config/admin.js file:

```
const fs = require('fs');

auth: {
    secret: env('ADMIN_JWT_SECRET', 'your-secret-here'),
    options: {
    algorithm: 'RS512',
    privateKey: fs.readFileSync('./private-key.pem', 'utf8'),
    publicKey: fs.readFileSync('./public-key.pem', 'utf8'),
    issuer: 'your-app-name',
    audience: 'admin',
    expiresIn: '1d'
    }
}
});
```

#### Option 2: New Configuration (admin.auth.sessions.options)

Add this to your config/admin.js file:
```
const fs = require('fs');

auth: {
    secret: env('ADMIN_JWT_SECRET', 'your-secret-here'),
    sessions: {
    options: {
        algorithm: 'RS512',
        privateKey: fs.readFileSync('./private-key.pem', 'utf8'),
        publicKey: fs.readFileSync('./public-key.pem', 'utf8'),
        issuer: 'your-app-name',
        audience: 'admin'
    },
    accessTokenLifespan: 30 * 60, // 30 minutes
    maxRefreshTokenLifespan: 30 * 24 * 60 * 60, // 30 days
    }
}
});

```

#### Generate private key
openssl genrsa -out private-key.pem 2048

#### Generate public key
openssl rsa -in private-key.pem -pubout -out public-key.pem

Testing Steps

1. Before the fix: Configure RS512 as above, start Strapi, try to login to admin panel
- Expected failure: "secretOrPrivateKey must be a symmetric key when using HS256" error
2. After the fix: Same configuration should work
- Expected success: Admin login works, JWT tokens use RS512 algorithm


innerdvations:
@jhoward1994 I believe we will also need to do the same thing for users-permissions but I haven't looked into how it was handled before -- was it copying the admin.auth.options or was there a users-permissions.auth.options that was being used?

Also, I don't know if we actually want to fully deprecate it, because that's kind of frustrating for users. We may to either:
- leave it in auth.options and NOT move it to sessions.options
- OR silently allow it in both locations, but update the documentation to only show the "new" location

docs/docs/docs/01-core/authentication/00-sessions-and-jwt.md
```diff
@@ -63,12 +63,20 @@ Optional request fields on login/register:
 
 Configuration:
 
-- `admin.auth.secret` — required JWT secret used by admin origin
+- `admin.auth.secret` — JWT secret for symmetric algorithms (HS256, HS384, HS512)
+- `admin.auth.sessions.options.algorithm` — JWT algorithm (default: 'HS256')
+- `admin.auth.sessions.options.privateKey` — Private key for asymmetric algorithms (RS256, RS512, ES256, etc.)
+- `admin.auth.sessions.options.publicKey` — Public key for asymmetric algorithms (RS256, RS512, ES256, etc.)
```
innerdvations, line 69: I think we can delete this actually, and just specify that `options` is what is passed in to `jwt.sign` et al

butcherZ (approved):
Tested for both options, admin login is working and token is using RS512, LGTM!

Issue #24428, opened 2025-09-20
User-permissions plugin, export user controllers as a plain object instead of as a factory


maksym-arturIn:
### Node Version

20.0.0

### NPM/Yarn/PNPM Version

10.0.0

### Strapi Version

5.7.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/strapi/strapi/blob/develop/packages/plugins/users-permissions/server/controllers/user.js

### Bug Description

The user-permission plugin, user controllers should be exported as factory instead of plain object. 

<img width="387" height="156" alt="Image" src="https://github.com/user-attachments/assets/7cff4e4c-2104-476b-b9ff-e675a737612e" />

### Steps to Reproduce

go to the `[packages/plugins/users-permissions/server/controllers/user.js](https://github.com/strapi/strapi/blob/develop/packages/plugins/users-permissions/server/controllers/user.js)` 

### Expected Behavior

user controllers exported as factory instead of plain object. It allows extend user controllers and use the `strapi` instance inside controllers

<img width="301" height="52" alt="Image" src="https://github.com/user-attachments/assets/c2651d60-cff5-4e22-a048-b4e9bb393ba2" />

### Logs

```shell

```

### Code Snippets

```
module.exports = ({ strapi }) => ({...})
```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

AyushMishraa:
Hi, please assign this issue to me and guide me

PR #24465, opened 2025-09-28
fix: improve user controller to export the factory instead of object

AyushMishraa:
In this PR I have changed the way controllers is being exported from UserController.
I have exported the Controller as factory instead of simple object. So that it can be used very easily through dependency.

Issue was https://github.com/strapi/strapi/issues/24428


packages/plugins/users-permissions/server/controllers/user.js
```diff
@@ -13,207 +13,202 @@ const { validateCreateUserBody, validateUpdateUserBody } = require('./validation
...
 
-  return strapi.contentAPI.sanitize.query(query, schema, { auth });
-};
+    return strapi.contentAPI.sanitize.query(query, schema, { auth });
+  };
 
-module.exports = {
-  /**
-   * Create a/an user record.
-   * @return {Object}
```
unrevised6419, line 40: Chore: Please undo all the JSDoc removed comments. They reduce code documentation and are unrelated to the main changes

PR #24467, opened 2025-09-29
Update clean-component-join-table.ts to fix non-DP entries not removing published relations

laurenskling:
Don't clean ghost relations for non-DP entries, which are supposed to have both states linked

### What does it do?

https://github.com/strapi/strapi/pull/24316 clean ghosts relations in components. But it by accident also deleted the published version of relations to non-DP entries. Which are supposed to exists. Needed to exist. 

### Why is it needed?

It fixes the repair script.

### How to test it?

1) have a non-DP contentType relation to a DP contentType
2) have some relations
3) run the repair code
4) see that with this PR it no longer removes the publish relations
5) see that requesting the API for this entry you can still get the relation in &state=published

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/24316


hanpaine:
Thanks @laurenskling, we're taking a look now 🙏

hanpaine:
As a heads up, for speed we are going to disable automatic run on the script and deprecate 5.24.1 to prevent the need for clean up as much as possible first. Will take a look and get this moving as a next prio this afternoon 🙏

laurenskling:
Thanks. I'm not married to this PR. This is my guestimation for what we should do. Any fix for this issue would be fine with me.

PR #24470, opened 2025-09-29
fix: stop repair script from running automatically

butcherZ:
### What does it do?

temporarily stop repair script from running

### Why is it needed?

hot fix for https://github.com/strapi/strapi/pull/24316#issuecomment-3338115028 the issue mentioned here

### How to test it?

start strapi and the repairing script should not be running automatically

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/24316#issuecomment-3338115028

### DIFF
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -37,7 +37,7 @@ import { createContentSourceMapsService } from './services/content-source-maps';
 import { coreStoreModel } from './services/core-store';
 import { createConfigProvider } from './services/config';
 
-import { cleanComponentJoinTable } from './services/document-service/utils/clean-component-join-table';
+// import { cleanComponentJoinTable } from './services/document-service/utils/clean-component-join-table';
 
 class Strapi extends Container implements Core.Strapi {
   app: any;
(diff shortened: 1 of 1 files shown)


jhoward1994 (approved):

packages/core/core/src/Strapi.ts
```diff
@@ -449,26 +449,29 @@ class Strapi extends Container implements Core.Strapi {
       contentTypes: this.contentTypes,
     });
 
-    const status = await this.db.schema.sync();
```
Bassel17, line 452: you need to keep this

packages/core/core/src/Strapi.ts
```diff
@@ -449,26 +449,29 @@ class Strapi extends Container implements Core.Strapi {
...
-
-    // if schemas have changed, run repairs
-    if (status === 'CHANGED') {
```
Bassel17, line 455: And this as well

hanpaine (commented):
When we add Laurens PR in today/tomorrow, we need to change this again

Bassel17 (approved):

PR #24474, opened 2025-09-29
feat: allow to override relations to display in admin

Slashgear:
### What does it do?

This PR add new feature for realtion input in strapi admin in order to allow user to manipulate more that 5 entities in relation field. Keep the hardcoded 5 by default but allow to override it is optional field

### Why is it needed?

For a strapi we handle with my team, we have a page with a field which is a one-to-may relation field to a list of speaker for a conference.
Contribution team have some issue to contribute this field.
- We have about 70 Speakers in this fields (seeing only 5 and need to loadMore X times in annoying)
- Sorting speakers in a tiny 5 items list view in complex with drag and drop

### How to test it?

Add new field in entity schema for relation, you should see new height and item display

```json
{
 "attributes": {
    "speakers": {
      "type": "relation",
      "relation": "oneToMany",
      "target": "api::speaker.speaker",
	  "relationsToDisplay": 100
    }
  }
}
```

### Related issue(s)/PR(s)

### Todo

- [ ] is there a way to implement test on this ? 
- [ ] did not touch the documentation part, if PR is accepted, tell me where I should add some documentation and I will
- [ ] did nothing to be able to edit it in strapi admin UI, waiting for your feedback

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
@@ -214,6 +214,8 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
           attribute.target === props.attribute.target
       ).length > 0;
 
+    const relationsToDisplay: number = props.attribute?.relationsToDisplay || RELATIONS_TO_DISPLAY;
+
     const { data, isLoading, isFetching } = useGetRelationsQuery(
       {
         model,
@@ -222,7 +224,7 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
         id,
         params: {
           ...currentDocumentMeta.params,
-          pageSize: RELATIONS_TO_DISPLAY,
+          pageSize: relationsToDisplay,
           page: currentPage,
         },
       },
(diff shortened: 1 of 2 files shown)


mathildeleg:
Hey, thanks for the contribution and for taking the time to describe your use case. The pain point is real and we appreciate you raising it.

We're going to close this PR as-is for a few reasons:
- Wrong layer for the config. `relationsToDisplay` is a pure admin UI display preference — it has no bearing on the data model or API. Storing it in `RelationOptions` (and by extension schema.json) couples a rendering detail to the schema definition, which means it would appear in exported schemas and API schema responses. The right home for this kind of setting is the content-type's admin layout configuration (alongside things like mainField and sort defaults), not the attribute definition.
- No safety cap. The server enforces `pageSize ≤ 100` via validation. Any value above 100 would silently break the relations field with a 400 error and no feedback to the user. The PR has no client-side guard against this.
- The drag-and-drop problem isn't solved. The stated pain point is reordering 70 speakers in a small list. Making more items visible upfront doesn't fix the UX of dragging through a long virtualized scroll container — it's still very fiddly at that scale. A proper fix would need a different reordering UX for large lists.

Could you open an issue as the underlying need is valid? That way we have visibility on the issue and will keep it in mind to fix (to have it fixed by you or another contributor if anyone wants to suggest another way to fix it) :) 

Thanks again!

PR #24498, opened 2025-10-02
fix: allow submitting image messages to AI chat

Bassel17:
### What does it do?

- Allows messages with attachments to be sent to AI chat

### Why is it needed?

- Messages with Attachments where not being sent
- a wrong check prevented messages from being sent when attachments where present

### How to test it?

- send an attachment to AI chat, could be an image, figma, code... it should work


innerdvations (approved):
LGTM but I haven't tested it

butcherZ (approved):

PR #24505, opened 2025-10-03
enhancement: AI notification system improvements

Bassel17:
### What does it do?

- Show different notification for plans without overages
- show multiple notifications based on levels, 80% usage then 90% usage then 100% usage (prevent spamming the user with multiple notifications)
-  Adds some unit tests

### Why is it needed?

- improves UX

### How to test it?

- reach near the end of the consumption in a plan


jhoward1994:
Looks good, could we use something from this file to make these notifications appear for longer ? When I was testing last week it was hard to catch them

```
packages/core/admin/admin/src/features/Notifications.tsx
interface NotificationConfig {
  timeout?: number;
}
```

Bassel17:
> Looks good, could we use something from this file to make these notifications appear for longer ? When I was testing last week it was hard to catch them
> 
> ```
> packages/core/admin/admin/src/features/Notifications.tsx
> interface NotificationConfig {
>   timeout?: number;
> }
> ```

I increased the time for the AI notifications for 5 seconds (double the default)
didn't change the default time because that will affect other notifications in the app and am not sure if that's wanted

PR #24515, opened 2025-10-06
feat(upload): generating metadata from thumbnail

araksyagevorgyan:
### What does it do?

This update changes the AI metadata generation source in Strapi.
Previously, the metadata was generated from the original uploaded files.
With this change, the metadata will now be generated from the thumbnail versions of the files instead.

### Why is it needed?

Generating metadata from original files can be unnecessarily heavy and resource-consuming, especially for large files.
By using thumbnails instead, we reduce processing time, improve performance, and optimize resource usage without losing meaningful information for AI metadata extraction.

### How to test it?

Go to `Media Library` -> `Add new assets` -> you should see metadata has been generated by ai service.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/upload/server/src/services/__tests__/ai-metadata.test.ts
+++ b/packages/core/upload/server/src/services/__tests__/ai-metadata.test.ts
@@ -20,8 +20,8 @@ const mockFetch = jest.fn().mockResolvedValue({
   }),
   text: jest.fn().mockResolvedValue(''),
   headers: {
-    get: jest.fn().mockReturnValue('image/jpeg')
-  }
+    get: jest.fn().mockReturnValue('image/jpeg'),
+  },
 } as any);
 
 global.fetch = mockFetch;


packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -116,30 +116,53 @@ export default {
     const data = await validateUploadBody(body, Array.isArray(files));
     console.log('data', data);
     const filesArray = Array.isArray(files) ? files : [files];
```
remidej: could you remove the console logs in this file at the same time?

packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -116,30 +116,52 @@ export default {
...
+              originalFilename: file.name,
+              size: file.formats?.thumbnail?.size || file.size,
+              provider: file.provider,
+            }) as unknown as any
+        );
+
+        const metadataResults = await aiMetadataService.processFiles(thumbnailFiles);
+
+        // Update the uploaded files with AI metadata
+        for (const uploadedFile of uploadedFiles) {
```
remidej: To match the style of our codebase better I'd recommend using forEach here instead

remidej (approved):
tested 👍

Issue #22408, opened 2024-12-12
Pages can be scrolled when hovering sticky navs


lucasboilly:
### Required System information

- Node.js version: 22.11.0
- Strapi version: 5.5.1
- Database: sqlite
- Operating system: macOS

### Steps to reproduce the behavior

1. Go to any page
2. Hover one of the navs
3. Scroll
4. The content on the right will scroll

### Expected behavior

We should only be able to scroll if our cursor is in a scrollable area.

### Screenshots

https://github.com/user-attachments/assets/7f108770-3011-43cf-90e8-48f93e02bef5

Datata1:
hey @lucasboilly 

Sorry, this is my first attempt at contributing to open source, and I forgot to ask if I could work on this issue. Would it be okay if I submit/keep this PR for the issue?

Best regards

lucasboilly:
Hello @Datata1 :)

It's completely okay! I saw you submitted a PR, I'm gonna check it and ask one of our engineers to check the code as well.

Thanks a lot for contributing to Strapi!

Arbaz001:
hiii can you assign me

faiz-ali-khan-developer:
Hi, @lucasboilly I saw that there is a PR opened for this issue but it's been idle for some time.
I'd be happy to undertake this if the existing PR is not being pursued. Let me know if it is fine for me to undertake this.

HattySohaib:
Hi @lucasboilly, I'm willing to work on this issue if it hasn't been resolved yet. Let me know if I can.

Regards,
Sohaib Aftab

lucasboilly:
Hello, don't hesitate to submit a pull request for it :) We don't assign issues.

eye-akankshaa:
Hello @lucasboilly, I would like to work on this issue if any fix has not been done yet. Please let me know

debesh00800:
Hi @lucasboilly , I would like to work on this if it is not assigned yet

Riyanshi26:
Hi @lucasboilly , I see some PRs referring to the issue. Has this issue been fixed?

tysoncung:
I'm interested in working on this issue. Could you provide more context about what you're looking for? Any additional details about requirements or constraints would be helpful.

awoladhossain:
Hi, I would like to work on this issue.
Is it still available?

Adi-Evolve:
[...]

PR #24517, opened 2025-10-06
fix: prevent scroll chaining from menu and subNav

guptadeepak8:
### What does it do?
Prevents the main page from scrolling when users reach the top or bottom of a submenu or LeftMenu.

### Why is it needed?

Before, scrolling to the top or bottom of a submenu or left menu, there was scroll chaining that scrolled the main page, causing a confusing experience

### How to test it?

Open any page with left submenus (Content Manager).
Hover over a submenu and scroll up/down.
Confirm that only the submenu scrolls when at the top or bottom.

### Related issue(s)/PR(s)

Fixes: #22408

prev:-

https://github.com/user-attachments/assets/fa29b410-14fe-4a4b-8728-5d487c10a5bf

Fix :-

https://github.com/user-attachments/assets/9a04a2ae-37b7-491e-a8f8-b38f39843704


guptadeepak8:
hey @lucasboilly can you review this

lucasboilly:
Hi @guptadeepak8, thanks for your contribution!

I just checked and it seems to work in some places but not in others (ex: if scrolling while hovering the SubNav header or in the Content Manager).

https://github.com/user-attachments/assets/c98aaaad-fd99-475a-8eef-c5f7aae567ab

guptadeepak8:
https://github.com/user-attachments/assets/46467d94-72f3-429a-878a-a4a8a09cd311

@lucasboilly

packages/core/admin/admin/src/components/SubNav.tsx
```diff
@@ -350,8 +350,14 @@ const PageWrapper = styled(Box)`
   }
 `;
 
+const StyledScrollArea = styled(ScrollArea)`
```
guptadeepak8, line 380: I think this is a better approach, but still seeing chaining in the headers

lucasboilly:
Hi @guptadeepak8, I don't see any changes yet compared to my last review. Did you apply them?

guptadeepak8:
looking for better approach

packages/core/admin/admin/src/components/SubNav.tsx
```diff
@@ -120,8 +120,23 @@ const StyledHeader = styled(Flex)`
 `;
 
 const Header = ({ label }: { label: string }) => {
+  const headerRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    const el = headerRef.current;
+    if (!el) return;
+
+    const handleWheel = (e: WheelEvent) => {
```
guptadeepak8, line 129: this is little aggressive but it works

guptadeepak8:
@lucasboilly this works now

lucasboilly:
Thanks @guptadeepak8! Could you apply it in the Content Manager as well?

guptadeepak8:
I have added this into to parent component so the scroll bubbling disable into both setting & content manager too

lucasboilly:
It seems there's a last remaining area where we can scroll in the CM (the rest is good).

https://github.com/user-attachments/assets/4b6cd8f3-495e-427a-bc37-ac8f7e21eae4

guptadeepak8:
Hey @lucasboilly, thanks for the review 

https://www.loom.com/share/84018a44d60f40659b67b44c60b61f4b?sid=92dbb236-e8bf-470f-ab9a-468800e1233b
Here it is working in all parts

lucasboilly:
@guptadeepak8 I now have a double scroll issue.

https://github.com/user-attachments/assets/bab9fe7a-a53c-418c-b1d9-d20a5b3a16dd

mathildeleg:
Hi thanks for taking the time to look into this, but I'm afraid this isn't the quite right fix for this.

It does fix the issue, but it would have a negative impact on performance. 

I'm closing this PR for now but feel free to re-open this one or open a new one with another solution if you find one 🙏🏻

PR #24519, opened 2025-10-06
fix: don't render ctb ai chatbot in production mode

Bassel17:
### What does it do?

- Checks if we are in Production mode before rendering the CTB AI chat

### Why is it needed?

- Chatbot should not be accessible in production mode, because no changes can be done to the CTB

### How to test it?

- run the app with `NODE_ENV=production yarn start` 
- chatbot should not appear even when having a growth license

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/AIChat/Chat.tsx
+++ b/packages/core/content-type-builder/admin/src/components/AIChat/Chat.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from 'react';
 
-import { tours, useTracking, useGuidedTour } from '@strapi/admin/strapi-admin';
+import { tours, useTracking, useGuidedTour, useAppInfo } from '@strapi/admin/strapi-admin';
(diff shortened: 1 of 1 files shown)


packages/core/content-type-builder/admin/src/components/AIChat/Chat.tsx
```diff
@@ -393,7 +397,8 @@ const Chat = () => {
     'Strapi AI can make mistakes.'
   );
 
-  if (!isChatEnabled) {
+  // Don't render the chat at all in production mode or if chat is not enabled
+  if (!isChatEnabled || isProduction) {
```
innerdvations, line 401: ~~This should be at the CTB level, since the chatbox will work in production everywhere else. If we just want this quick fix, it works for now, but we'll have to update it again in the next few weeks~~

ALthough I guess it's still not refactored and is only in CTB, so nevermind 😆

innerdvations (approved):
LGTM!

PR #24521, opened 2025-10-07
fix: scroll in sidenav

Adzouz:
### What does it do?

Fix the scrollbar missing in CTB after working on responsiveness. Unifying by creating a component on SubNav level to avoid making the mistake again.

### Why is it needed?

The scroll is not working anymore on the CTB side nav.

### How to test it?

- Go to the Content Type Builder
- Scroll in the left side navigation (if the window is too high, resize it to make the scroll appear)

🚀

### DIFF
--- a/packages/core/admin/admin/src/components/SubNav.tsx
+++ b/packages/core/admin/admin/src/components/SubNav.tsx
@@ -7,6 +7,7 @@ import {
   Typography,
   IconButton,
   Badge,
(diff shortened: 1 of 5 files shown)


HichamELBSI (dismissed):
👍🏽

markkaylor (approved):
LGTM tested locally

HichamELBSI (approved):

PR #24522, opened 2025-10-07
fix(upload): use absolute server url for file url

remidej:
As we found out in testing, `server.url` is not always provided, so `server.absoluteUrl` is the one that can be relied on. I found it by looking up what the startup terminal log used:
screenshot 2025-10-07 at 15 48 50@2x

### DIFF
--- a/packages/core/upload/server/src/services/__tests__/ai-metadata.test.ts
+++ b/packages/core/upload/server/src/services/__tests__/ai-metadata.test.ts
@@ -155,7 +155,7 @@ describe('AI Metadata Service', () => {
     beforeEach(() => {
       // Mock strapi config
       mockStrapi.config.get.mockImplementation((key: string) => {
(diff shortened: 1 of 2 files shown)


markkaylor (approved):

markkaylor:
The user can provide a custom url right? Will this still work in that case?

remidej:
@markkaylor wdym a custom url?

markkaylor:
> @markkaylor wdym a custom url?

Sorry I mean will the absoluteUrl be referencing this server config `url: env('PUBLIC_URL', 'https://api.example.com')`  if it is provided?

remidej:
This is where it's defined: https://github.com/strapi/strapi/blob/16e623f866a025f3ea4b3d5bc836335475cda4d4/packages/core/core/src/configuration/urls.ts#L71

It seems to give priority to config.server.url and falls back to computing it via host + port concatenation

Issue #24452, opened 2025-09-25
[Workaround] Auth Secure Session


alxnkt:
## Workaround information

Created by @derrickmehaffy 

### Ensure that you have the Koa Proxy set

Koa has a built in method to allow for unsecured proxying if the reverse proxy is trusted, this is not set by default to set this within Strapi you need to set the following:

For **STRAPI 5**

```js
// Path: ./config/server.js or ./config/server.ts (Also check your environmental configs like ./config/env/*/server.js)

// JS Project
module.exports = ({ env }) => ({
  // ...
  proxy: { koa: true },
  // ...
});

// TS Project
export default ({ env }) => ({
  // ...
  proxy: { koa: true },
  // ...
});
```

For **STRAPI 4** ( This WILL NOT WORK in Strapi 5)

```js
// Path: ./config/server.js or ./config/server.ts (Also check your environmental configs like ./config/env/*/server.js)

// JS Project
module.exports = ({ env }) => ({
  // ...
  proxy: true,
  // ...
});

// TS Project
export default ({ env }) => ({
  // ...
  proxy: true,
  // ...
});
```

### Ensure your Reverse Proxy is properly configured

At minimum your reverse proxy should be passing along the `x-forwarded-for` and `x-forwarded-proto` headers, additional related headers are recommended. If your reverse proxy support Real-IP it's recommended you enable that. 

If your reverse proxy does not allow using these headers, then you need to change reverse proxies, these are common and well known security headers used for trusted proxying.

#### Nginx

Sub-snippet of only the proxy part:

```
    location / {
        proxy_pass http://127.0.0.1:1337;
        proxy_http_version 1.1;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-Server $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header Host $http_host;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "Upgrade";
        proxy_pass_request_headers on;
    }
```

#### Caddy

If you use the default reverse proxy method, Caddy does this for you automatically, though you can install and enable the optional real-ip module/plugin.

For example:

```
api.example.com {
  reverse_proxy 127.0.0.1:1337
}
```

[...]

PR #24523, opened 2025-10-07
enhancement: sync project generation templates

derrickmehaffy:
### What does it do?

Syncs the config files but also adds in the proxy/url vars and inits the env config directory; specific list:

- Adds in `proxy: { koa: true }` by default on both JS/TS template generation (Related to, but not a fix for https://github.com/strapi/strapi/issues/24452)
- Adds in `url: env('PUBLIC_URL')` to make it easier to configure (considered setting a default variable in `.env` but that could cause issues if users change the default port). For now it more so makes users aware of this variable and that they should set it
- Sync the JS/TS configs as they were out of sync
- Initializes the production env config directory as this is required for Strapi Cloud if the user needs to configure plugins (they CANNOT use `./config/plugins.js|.ts` because of the config injection done by Cloud. See Cloud [Upload](https://docs.strapi.io/cloud/advanced/upload) and [Email](https://docs.strapi.io/cloud/advanced/email) documentation

### Why is it needed?

Trying to cut down on support tickets by implementing easy feedback from users.

### How to test it?

Generate an experimental and generate new projects, check that the new template is used.

### Related issue(s)/PR(s)

Related to BUT not a fix for: https://github.com/strapi/strapi/issues/24452

### DIFF
--- a/packages/cli/create-strapi-app/templates/example/config/server.ts
+++ b/packages/cli/create-strapi-app/templates/example/config/server.ts
@@ -4,4 +4,9 @@ export default ({ env }) => ({
   app: {
     keys: env.array('APP_KEYS'),
   },
+  webhooks: {
+    populateRelations: env.bool('WEBHOOKS_POPULATE_RELATIONS', false),
+  },
+  proxy: { koa: env('PROXY_KOA', true) },
+  url: env('PUBLIC_URL'),
 });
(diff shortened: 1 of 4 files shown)


packages/cli/create-strapi-app/templates/example/config/server.ts
```diff
@@ -4,4 +4,9 @@ export default ({ env }) => ({
   app: {
     keys: env.array('APP_KEYS'),
   },
+  webhooks: {
+    populateRelations: env.bool('WEBHOOKS_POPULATE_RELATIONS', false),
```
laurenskling, line 8: https://docs.strapi.io/cms/migration/v4-to-v5/breaking-changes/remove-webhook-populate-relations :)

PR #24530, opened 2025-10-08
fix: ai settings show up for licenses without ai

markkaylor:
### What does it do?

- Conditionally renders the media lib ai settings based on whether or not the app has ai available
- Updates the upload modal header text to:  'X assets uploaded, review AI generated metadata'

### Why is it needed?

We shouldn't see ai related settings if there is no AI

### How to test it?

With CE or an EE license that does NOT have AI 
Go to the settings page for media library
You should NOT see any settings related to AI

With a growth license or license with AI
You should see the AI settings

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/upload/admin/src/translations/en.json
+++ b/packages/core/upload/admin/src/translations/en.json
@@ -141,6 +141,6 @@
   "view-switch.list": "List View",
   "view-switch.grid": "Grid View",
   "ai.modal.uploading.title": "Uploading and processing with AI...",
-  "ai.modal.title": "{count, plural, one {# Asset uploaded} other {# Assets uploaded}} time to review AI generated content",
+  "ai.modal.title": "{count, plural, one {# asset uploaded} other {# assets uploaded}}, time to review AI generated content",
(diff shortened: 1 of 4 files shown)


packages/core/upload/admin/src/translations/en.json
```diff
@@ -141,6 +141,6 @@
   "view-switch.list": "List View",
   "view-switch.grid": "Grid View",
   "ai.modal.uploading.title": "Uploading and processing with AI...",
-  "ai.modal.title": "{count, plural, one {# Asset uploaded} other {# Assets uploaded}} time to review AI generated content",
+  "ai.modal.title": "{count, plural, one {# asset uploaded} other {# assets uploaded}}, review AI generated metadata",
```
remidej, line 144: I think there should be a defaultMessage to update on the modal component

markkaylor: Good catch, pushed.

remidej (approved):
thanks for the fix! tested

PR #24534, opened 2025-10-08
enhancement: add en language for Japan

qburst-pranavtv:
### What does it do?

Add en-JP language for Japan region

### Why is it needed?

One of our project needs these languages added


qburst-pranavtv:
@hanpaine i have fixed unit testing, could you please review.
Thanks

hanpaine (approved):

Issue #24363, opened 2025-09-12
Relations in dynamic zone fails to publish


lan23333:
### Node Version

20.19.3

### NPM/Yarn/PNPM Version

11.4.2

### Strapi Version

5.23.4

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Javascript

### Bug Description

When trying to publish a entity, said entity has a dynamic zone containing a component with relation field, Strapi successfully link the relation to an entity in a collection, but could not see it in published view even if it's already published. No error shown in console
When query through REST API, it shows it's not linked (not published)

### Steps to Reproduce

1. Create a collection with structure like this
<img width="904" height="497" alt="Image" src="https://github.com/user-attachments/assets/d0ae4201-ef4f-418d-ba33-c30c01bb5fca" />

each component in the dynamic zone is just all one way relation fields links to another collection
2. Create an item in the collection
3. add a component in the dynamic zone
4. link the relations
5. publish
6. Go to "Published" view, relations are empty, but if you have a text field you will see it's saved

### Expected Behavior

Relations should be filled in published view like when the component is not in a dynamic zone.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

OcneanuVlad:
Internal note: TID 3531

Z-aidH:
Not sure if this issue is related to also the case of updating file value, also inside the dynamic zone, I update the image published view still shows the old view, even the new image have been published successfully by the alerts, even if I try to update a text field with the change to the image, the text field updates normally but the image stays the same

alexandrebodin:
Hi @lan23333 I was not able to reproduce your issue on the latest version

I have tested the following:

- A content type "A" with draft & published enabled
- A content type "B" with draft & published enabled
  - B has a dynamic zone with a component that points to A (single relation)


- I create an entry for A and publish it
- I create en entry for B and select the published A in the component relation


I fetch the api with 

`http://localhost:1337/api/bs?populate[dz][populate]=*`


Any info you can share to help debug would be appreciated (your schema, your api calls...) 


🙏

LFQiwei:
[...]

PR #24539, opened 2025-10-09
fix: find published relations on components

alexandrebodin:
### What does it do?

FIx https://github.com/strapi/strapi/issues/24363

### Why is it needed?

get findExisting relations in the publishedTab sends a status=published that wasn't properly accepted when the relation is on a component (it doesn't have dp but still should load dp relations)

### How to test it?

Follow the steps shared in the issue (and check the published tab instead of the api response)

### Related issue(s)/PR(s)


markkaylor:
Hey @alexandrebodin thanks for the PR. Unfortunately I still don't see the published relation in the published tab. I believe my setup is correct but let me know if I missed something. 


https://github.com/user-attachments/assets/1509622d-f46c-44b8-ad66-4ac2b321b8dc

PR #24553, opened 2025-10-10
chore: set up ai i18n future flag

remidej:
### What does it do?

- sets up AI i18n future flag so we can work safely on develop
- create ai localizations service with isEnabled method

see https://github.com/strapi/content-squad/issues/92

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -1,0 +1,30 @@
+import type { Core } from '@strapi/types';
+
+const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
+  return {
+    // Async to avoid changing the signature later (there will be a db check in the future)
+    async isEnabled() {
+      // Check if future flag is enabled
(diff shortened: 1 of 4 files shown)


packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -0,0 +1,30 @@
...
+      const isAIEnabled = strapi.config.get('admin.ai.enabled', true);
+      if (!isAIEnabled) {
+        return false;
+      }
+
+      // Check if the user's license grants access to AI features
+      const hasAccess = strapi.ee.features.isEnabled('cms-ai');
+      if (!hasAccess) {
+        return false;
+      }
```
markkaylor, line 23: This part is exactly what we did for media lib too right? I wonder if we should have a service coming from `admin` that mirrors what `useAIAvailability` does on the frontend...if you see what I mean. Doesn't necessarily have to be a part of this PR but just a thought for discussion.

remidej: Yes currently it's basically a copy. I think we can avoid the abstraction for now because there may be slight differences that appear in the future (opt-in vs opt-out, entitlement name...), and it's only copied in 2 places, so not too concerning to have it copied imo

PR #24554, opened 2025-10-10
feat(upload): fixing ordering issue

araksyagevorgyan:
### What does it do?

Fixes a bug causing file name mismatches during bulk uploads in the Media Library.

### Why is it needed?

When performing a bulk upload in Media Library, the uploaded files are reordered during the upload process, causing a mismatch between the original file names (from the frontend) and the files stored in Strapi.
For instance, uploading 10 images at once results in all files being successfully uploaded and having AI metadata generated, but their names no longer match the original local files.

Here is an example of
Uploaded files in Media Library:
image

VS

Files on local machine:
image

As you can see, pic1.jpeg differs between the local folder and the Media Library.
This mismatch occurs randomly — it's not always the first file that differs, but the naming can change unpredictably during bulk uploads.

**Root Cause**
This issue appears to be related to Koa's internal handling of multipart form data.
During parsing, Koa may reorder files within the request before they reach the upload controller. As a result, the backend receives the files in a different order than the one sent from the frontend, which leads to mismatched file-name-to-metadata associations.

### How to test it?

- Upload multiple files at once in the Media Library.
- Verify that the uploaded files order, names, and metadata match the original local files.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/upload/server/src/controllers/admin-upload.ts
+++ b/packages/core/upload/server/src/controllers/admin-upload.ts
@@ -126,7 +126,7 @@ export default {
         .map((info) => {
           return filesArray.find((file) => file.originalFilename === info.name);
         })
-        .filter(Boolean) as import('formidable').File[];
+        .filter(Boolean) as any[];
 
       filesArray = alignedFilesArray;
     }


packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -113,7 +113,23 @@ export default {
...
+
+    if (
+      data.fileInfo &&
+      Array.isArray(data.fileInfo) &&
+      filesArray.length === data.fileInfo.length
+    ) {
+      // Reorder filesArray to match data.fileInfo order
+      const alignedFilesArray = data.fileInfo
+        .map((info) => {
+          return filesArray.find((file) => file.originalFilename === info.name);
```
remidej, line 127: I like the idea! what happens when you have multiple files of the same name though? maybe we should only change the order if we detect the name is different?

araksyagevorgyan: I am not sure about your question, sorry. How can I upload multiple files with the same name at the same time? 🤔

araksyagevorgyan: oh, you mean the same name order, right? I think it is almost every time different. I am not sure if checking will optimize the code in this case.

remidej: I think I found the answer to my question: filesArray.find would stop running after the first match. So if several files have the same name, then we wouldn't be changing the order. Which means we're good!

araksyagevorgyan: sounds good to me 👍

Adzouz:
Just one thing @araksyagevorgyan for next time, it was flagged as a feature but it was a fix so keep that in mind to avoid bumping to a minor version 😊 (this time it wasn't a big deal because we already had another feature)

PR #24555, opened 2025-10-10
enhancement: enhance plugin generators

boazpoolman:
### What does it do?

This PR enhances the `@strapi/generators` for it's use in plugins. Primary changes:

1. It will add the newly created service/controller/content-type to the index file
2. It will also create a routes file for plugins when using the `api` generator
3. Put all the generated plugin files in 'server/src/' instead of 'server/', this is in line with the plugin SDK initialization
4. Create custom hbs templates for plugins, allowing for the use of route factories
5. Correctly identify the use of Typescript in plugins

### Why is it needed?

Enhance the DX when using the generators to build plugins. Some of the generators (content-types & routes) are currently broken for plugins because they don't respect the format in which they need to be exported to be valid in a plugin.

### Tradeoffs

The `extend-plugin-index-files.ts` helper has gotten quite complex. I wonder if there is a better approach for this, or even a better location? I'm happy to take some feedback on this!

### How to test it?

1. Pull this branch
2. Navigate to the generators package: `cd packages/generators/generators`
3. Install dependencies: `yarn install`
4. Build and publish: `yarn build && npx yalc push --publish`
5. Setup a new strapi project: `npx create-strapi@latest my-strapi-project`
6. Navigate to the project: `cd my-strapi-project`
7. Install the updated generators package: `npx yalc add @strapi/generators`
8. Create a local plugin in your project: `cd src/plugins && npx @strapi/sdk-plugin init test-plugin`
9. Use the `yarn strapi generate` command and generate some code for your new `test-plugin`
10. Witness all the improvements

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "@commitlint/config-conventional": "19.1.0",
     "@commitlint/prompt-cli": "19.2.0",
     "@nx/js": "20.4.6",
-    "@playwright/test": "1.48.2",
+    "@playwright/test": "1.56.1",
     "@rollup/plugin-commonjs": "28.0.1",
     "@rollup/plugin-dynamic-import-vars": "2.1.5",
     "@rollup/plugin-image": "3.0.3",
--- a/packages/core/admin/admin/src/ee.ts
+++ b/packages/core/admin/admin/src/ee.ts
@@ -4,5 +4,5 @@
 
 export { useLicenseLimits } from '../../ee/admin/src/hooks/useLicenseLimits';
 export { useEnterprise } from './hooks/useEnterprise';
-export { useGetAIUsageQuery } from '../../ee/admin/src/services/ai';
+export { useGetAIUsageQuery, useGetAIFeatureConfigQuery } from '../../ee/admin/src/services/ai';
 export { useAIAvailability } from '../../ee/admin/src/hooks/useAIAvailability';
--- a/packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
+++ b/packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
@@ -15,6 +15,7 @@ import { Pagination } from '../../../../components/Pagination';
 import { SearchInput } from '../../../../components/SearchInput';
 import { Table } from '../../../../components/Table';
 import { useTypedSelector } from '../../../../core/store/hooks';
+import { useAuth } from '../../../../features/Auth';
 import { useNotification } from '../../../../features/Notifications';
 import { useAPIErrorHandler } from '../../../../hooks/useAPIErrorHandler';
 import { useEnterprise } from '../../../../hooks/useEnterprise';
@@ -43,7 +44,7 @@ const ListPageCE = () => {
   const [showDeleteConfirmation, setShowDeleteConfirmation] = React.useState(false);
   const [idsToDelete, setIdsToDelete] = React.useState<Array<SanitizedAdminUser['id']>>([]);
   const { data, isError, isLoading } = useAdminUsers(qs.parse(search, { ignoreQueryPrefix: true }));
-
+  const { user: currentUser } = useAuth('ListPageCE', (auth) => auth);
   const { pagination, users = [] } = data ?? {};
 
   const CreateAction = useEnterprise(
@@ -200,7 +201,7 @@ const ListPageCE = () => {
                             <Pencil />
                           </IconButton>
                         ) : null}
-                        {canDelete ? (
+                        {canDelete && currentUser?.id !== user.id ? (
                           <IconButton
                             onClick={handleDeleteClick(user.id)}
                             label={formatMessage(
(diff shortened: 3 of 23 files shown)


packages/generators/generators/src/plops/utils/extend-plugin-index-files.ts
```diff
@@ -0,0 +1,359 @@
...
+// Helper to find and add to routes array
+const handleRoutesArray = (obj: any, config: AppendConfig) => {
+  if (!obj?.properties) return false;
+
+  const routesProp = obj.properties.find((prop: any) =>
+    j.ObjectProperty.check(prop) &&
+    j.Identifier.check(prop.key) &&
+    prop.key.name === 'routes' &&
+    j.ArrayExpression.check(prop.value)
+  );
```
innerdvations:
```suggestion
  const routesProp = obj.properties.find((prop: any) =>
    j.ObjectProperty.check(prop) &&
    ((j.Identifier.check(prop.key) && prop.key.name === 'routes') ||
     (j.Literal.check(prop.key) && prop.key.value === 'routes')) &&
    j.ArrayExpression.check(prop.value)
  );
```
This way it will work in case it receives a string key like `{ 'routes': someRoutes }` in addition to `{ routes: someRoutes }`

packages/core/types/src/core/core-api/router.ts
```diff
@@ -68,5 +68,5 @@ export type Route = {
 export type Router = {
   type: RouterType;
   prefix?: string;
-  routes: Route[] | (() => Route[]);
+  routes: Route[];
```
innerdvations:
```suggestion
  routes: Route[] | (() => Route[]);
```

I think this would be a breaking change, we need to keep the function that returns a `Route[]` too

innerdvations:
Also `yarn lint:fix` and `yarn prettier:write` can hopefully autofix the formatting for you

packages/generators/generators/src/plops/utils/extend-plugin-index-files.ts
boazpoolman: Awesome, added!

packages/core/types/src/core/core-api/router.ts
boazpoolman: I've reverted the change I made, but I wonder; this type `Core.CoreAPI.Router.Router` is only used in the [createCoreRouter factory](https://github.com/strapi/strapi/blob/develop/packages/core/core/src/factories.ts#L84), and it seems there is only ever an array returned, never a function that returns an array. So keeping the type as is would be only for people using that type in the wild? Can we then maby create a new type which does accurately describe the core router routes?

boazpoolman:
I've added documentation for the plugin SDK, you can find that here https://github.com/strapi/documentation/pull/2787.
I don't believe the rest of these changes need any more documentation as it's already covered by the ['generate' command docs](https://docs-next.strapi.io/cms/cli#strapi-generate)

packages/generators/generators/src/index.ts
```diff
@@ -11,8 +11,8 @@ export const runCLI = async () => {
       configPath: join(__dirname, 'plopfile.js'),
     },
     (env) => {
-      const argv = process.argv.slice(2); // Extract command-line arguments
-      Plop.execute(env, argv, (env, argv) => {
+      Plop.execute(env, [], (env, argv) => {
+        console.log(env, argv);
```
jhoward1994: Should we remove this ?

boazpoolman: ~~I should've removed my console.log there, sorry about that.~~ Done

Regarding your comment: Whilst testing I noticed that I could not pass the generator type as an argument to the CLI, e.g. `strapi generate content-type`. That would throw an error.

By removing this piece of code said issue was resolved. I figured this would not have impact on anything else.

jhoward1994:
It seems to be working well for me. I've tested generating controllers, services, content types etc within a local plugin.

Could you resolve the conflict? It's just a tiny package version thing

boazpoolman:
@jhoward1994 merge conflicts are resolved now :)

packages/core/types/src/core/core-api/router.ts
innerdvations: Yeah, I think in reality it would be a non-issue for 99.99% of people, but there's always someone out there using anything we've exposed publicly as a capability, and from glancing through the code it looks like a function is actually valid here

innerdvations (approved):

packages/core/types/src/core/core-api/router.ts
boazpoolman: Just to give some context here. When using this PR to generate a content-type in your plugin, it will create a core router for that content-type also. Problem is, to use that core router I'll have to manually add it's `.routes` to the plugins index file for routes. Which will result in something like this:


**server/src/routes/content-api/task.ts:**
```ts
import { factories } from '@strapi/strapi';
export default factories.createCoreRouter('plugin::todo.task');
```

**server/src/routes/content-api/index.ts:**
```ts
import task from './task';
export default () => ({
  type: 'content-api',
  // @ts-ignore
  routes: [...task.routes],
});
```

Notice the @ts-ignore because `createCoreRouter().routes` can also be an array, while in fact it can not. As far as I can see the routes from a core router will always be an array.

hanpaine:
Docs PR for merging once released https://github.com/strapi/documentation/pull/2787

PR #24569, opened 2025-10-13
security(deps): update pinned versions of axios, tar-fs, form-data, and sha.js

innerdvations:
### What does it do?

Updates the pinned versions of axios, sha.js, tar-fs, form-data, and sha.js

### Why is it needed?

CVE-2025-9288
CVE-2025-58754
CVE-2025-27152
CVE-2025-48387
CVE-2024-12905
CVE-2025-59343
CVE-2025-7783

Note that Strapi is not necessarily vulnerable to all or any of those issues, but this ensures that none of the linked dependencies are vulnerable versions.

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/components/DragLayer.tsx
+++ b/packages/core/admin/admin/src/components/DragLayer.tsx
@@ -3,10 +3,34 @@ import * as React from 'react';
 import { Box } from '@strapi/design-system';
 import { DragLayerMonitor, XYCoord, useDragLayer } from 'react-dnd';
 
+import { getWidgetElement } from '../utils/widgetLayout';
+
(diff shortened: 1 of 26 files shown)


yarn.lock
remidej, line 1: could you share how you obtained this without a package.json change?

innerdvations: You can use `yarn up <package>` and it should update it to the latest allowed version, even for packages we don't use directly.

If a package refuses to update, you can use `yarn why <package>` and see what is requiring it to stay at a lower version, and then try updating that one too.

But tbh I usually just delete the offending package(s) directly from the lock file and then run `yarn install` so that I can more quickly and clearly see exactly what changed.

PR #24570, opened 2025-10-13
future(i18n): add ai settings

markkaylor:
### What does it do?

- Creates an i18n settings service to get and update settings in the strapi store
- Adds the settings page with the enable/disable toggle. Dispatches action to update settings _on toggle_
- Updates the wording in the Media Library from true/false to enable disable per the designs

### Why is it needed?

So the user can enable the feature

### How to test it?

The future flag should already be set to true

Go to the i18n settings page
The toggle should be disabled by default
Enable it, refresh the page, it should persist

disable the future flag

Go the i18n settings page
You should not see the toggle

Remove you license containing ai or disable ai

Go the i18n settings page
You should not see the toggle

### Related issue(s)/PR(s)

resolves https://github.com/strapi/content-squad/issues/96

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
@@ -180,7 +180,7 @@ const DynamicComponent = ({
                 <React.Fragment key={category}>
                   <Menu.Label>{category}</Menu.Label>
                   {components.map(({ displayName, uid }) => (
-                    <Menu.Item key={componentUid} onSelect={() => onAddComponent(uid, index)}>
+                    <Menu.Item key={uid} onSelect={() => onAddComponent(uid, index)}>
                       {displayName}
                     </Menu.Item>
                   ))}
@@ -200,7 +200,7 @@ const DynamicComponent = ({
                 <React.Fragment key={category}>
                   <Menu.Label>{category}</Menu.Label>
                   {components.map(({ displayName, uid }) => (
-                    <Menu.Item key={componentUid} onSelect={() => onAddComponent(uid, index + 1)}>
+                    <Menu.Item key={uid} onSelect={() => onAddComponent(uid, index + 1)}>
                       {displayName}
                     </Menu.Item>
                   ))}
--- a/packages/core/content-type-builder/admin/src/components/AttributeRow.tsx
+++ b/packages/core/content-type-builder/admin/src/components/AttributeRow.tsx
@@ -118,7 +118,7 @@ const MemoizedRow = memo((props: Omit<AttributeRowProps, 'style'>) => {
 
   const src = 'target' in item && item.target ? 'relation' : ico;
 
-  const handleDelete = (e: React.MouseEvent) => {
+  const handleDelete = (e: React.MouseEvent | React.KeyboardEvent) => {
     e.stopPropagation();
     const dependentRows = checkDependentRows(contentTypes, item.name);
     if (dependentRows.length > 0) {
@@ -145,7 +145,11 @@ const MemoizedRow = memo((props: Omit<AttributeRowProps, 'style'>) => {
     setShowConfirmDialog(false);
   };
 
-  const handleClick = () => {
+  const handleClick = (e?: React.MouseEvent | React.KeyboardEvent) => {
+    if (e) {
+      e.stopPropagation();
+    }
+
     if (isMorph) {
       return;
     }
@@ -219,6 +223,7 @@ const MemoizedRow = memo((props: Omit<AttributeRowProps, 'style'>) => {
                 defaultMessage: 'Drag',
               })} ${item.name}`}
               disabled={isTypeDeleted || isDeleted}
+              style={{ outlineOffset: '-2px' }}
             >
               <Drag />
             </IconButton>
--- a/packages/core/content-type-builder/admin/src/components/List.tsx
+++ b/packages/core/content-type-builder/admin/src/components/List.tsx
@@ -3,6 +3,7 @@ import { useState } from 'react';
 import {
   DndContext,
   closestCenter,
+  KeyboardSensor,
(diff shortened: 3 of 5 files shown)


markkaylor:
~Forgot to add the translation keys~ updated

packages/plugins/i18n/server/src/services/settings.ts
```diff
@@ -0,0 +1,22 @@
+import type { Settings } from '../controllers/validation/settings';
+
+const settingsService = () => {
```
remidej: I recommend you use the createSettingsService syntax (like in ai-localizations for example) instead, this way you have access to the strapi without relying on the global

packages/plugins/i18n/server/src/services/settings.ts
```diff
@@ -0,0 +1,22 @@
...
+  async function getSettings() {
+    const res = await strapi.store!({ type: 'plugin', name: 'i18n', key: 'settings' }).get({});
+
+    return res as Settings | null;
```
remidej: could it make sense to parse using the schema here? this way we get the type without casting

packages/plugins/i18n/server/src/services/settings.ts
```diff
@@ -0,0 +1,22 @@
+import type { Settings } from '../controllers/validation/settings';
+
+const settingsService = () => {
+  async function getSettings() {
+    const res = await strapi.store!({ type: 'plugin', name: 'i18n', key: 'settings' }).get({});
```
remidej: i recommend you to extract a mini-orm outside the methods:

```
const settingsStore = strapi.store!({ type: 'plugin', name: 'i18n', key: 'settings' })
``` 

then you can use it in both getSettings and setSettings

packages/plugins/i18n/server/src/validation/settings.ts
```diff
@@ -0,0 +1,9 @@
+import { yup, validateYupSchema } from '@strapi/utils';
+
+const settingsSchema = yup.object({
+  aiLocalizations: yup.boolean().default(false),
+});
+
+export default validateYupSchema(settingsSchema);
+
+export type Settings = yup.InferType<typeof settingsSchema>;
```
markkaylor: Prettier formatted the file and there is an empty line 10 🤷

remidej:
When I open the i18n settings without a license, I get a notification toast:

>this must be a `object` type, but the final value was: `null`.

the settings endpoint returns a 400 which triggers the toast

markkaylor:
> When I open the i18n settings without a license, I get a notification toast:
> 
> > this must be a `object` type, but the final value was: `null`.
> 
> the settings endpoint returns a 400 which triggers the toast

Alright thanks I'll take a look

markkaylor:
@remidej 🤔 I'm not able to reproduce, could you pull and try again?


https://github.com/user-attachments/assets/bef88bbf-4f32-4848-8224-dc56101d56e1

packages/plugins/i18n/admin/src/pages/SettingsPage.tsx
```diff
@@ -26,6 +29,15 @@ const SettingsPage = () => {
...
     allowedActions: { canUpdate, canCreate, canDelete },
   } = useRBAC(PERMISSIONS);
+  const isAIAvailable = useAIAvailability();
+
+  // Settings state management
+  const {
+    data: settings,
+    isLoading: isLoadingSettings,
+    error: settingsError,
+  } = useGetSettingsQuery(undefined, { skip: !isAIAvailable });
```
markkaylor: I went ahead and skipped it anyway if ai isn't available, wdyt?

remidej: works for me 👍

packages/plugins/i18n/server/src/services/settings.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { Settings } from '../validation/settings';
+import validateSettings from '../validation/settings';
+
+const createSettingsService = ({ strapi }: { strapi: Core.Strapi }) => {
+  const settings = strapi.store!({ type: 'plugin', name: 'i18n', key: 'settings' });
+
+  async function getSettings() {
+    const res = await settings.get({});
+
+    return validateSettings(res);
```
remidej: I still get the error I shared earlier when I add a strapi license but don't have the setting in my db yet.

I think the issue is that the getSettings should also allow returning null, but with the validateSettings will throw an error

If it did handle null we probably wouldn't need to skip the query on the frontend when !isAiEnabled (the settings endpoint could host non-ai things in the future so it would be best not to skip it)

markkaylor: Sounds good I'll update

markkaylor: I'm starting to find it a bit strange to call `validateSettings` when getting the settings from the store just to parse the result in order to get type inference. To handle null I think I'll have to do something like:

```
 try {
      const validSettings = await validateSettings(res);
      return validSettings;
    } catch (e) {
      return null;
    }
```
does it not make more sense to revert back to casting the type?

remidej: or maybe
```
if (!res) {
  return null;
}
return validateSettings(res)
```

but I agree the validation doesn't bring much at this point, so we could revert it for simplicity. As you want!

markkaylor: I already pushed with it reverted so if you are cool with that I say we move forward but yeah I think your suggestion would work too

remidej: yes just saw that it's all good

markkaylor:
Hey @remidej sorry I need another approval I had to push a few times to fix a flaky unit test. The change to fix the test is here: https://github.com/strapi/strapi/pull/24570/commits/0d3ca798e53a8eb640b1af3deb1516041c74d8c2 although in that commit you will see a hard coded boolean and missing translation, resolved in the next two commits.

PR #24572, opened 2025-10-13
fix(content-manager): Stop components from linking between entries

### LINKED ISSUE #24568: STRAPI CMS mislinks data in separate rows.
### Node Version

20.19.2

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

4.25.7

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

I have Strapi CMS and have created a content builder structure (understood as columns). Now I create new blogs for my application based on the content builder structure, the data added to 1 blog (1 row) also starts reflecting in another blog (some other row) and deleting it from either removes the data.

### Steps to Reproduce

1. create project on Strapi Cloud 
2. create a content structure and commit it to content builder on strapi cloud
3. create different content pieces, minimum 17-20 to see the issue appear
4. One of the table names should be autocut by strapi and should exceed the 64 characters limit while creating,hence fixing this bug would create the issue.

### Expected Behavior

Data in every row should be independently existing. 

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

moizDhila52:
This PR resolves a critical data integrity bug where component data was being incorrectly shared across different entries in a collection type.

Closes #24568

The Problem
When creating or updating entries, especially in quick succession, the admin UI could reuse component data structures from its state. This payload, including the id of a component from a previously saved entry, was being sent to the backend.

The backend was incorrectly trusting this id, leading it to create a link to the existing component instead of creating a new, independent one.

This resulted in scenarios where:

1. Creating "Blog Post B" would show the component from "Blog Post A".
2. Editing the component in "Blog Post B" would also change the content in "Blog Post A".
3. Deleting the component from one entry would remove it from both.

Root Cause & Solution
The root cause was located in the packages/core/content-manager/server/src/services/document-manager.ts service. The create and update functions were not sanitizing the incoming component data before passing it to the database layer.

This fix directly addresses the issue by intercepting the data payload within both the create and update functions:

1. Safe Copying: It first performs a deep copy of the incoming component array using lodash.cloneDeep. This is a crucial step to prevent any mutations or side effects on the original request object.
2. ID Sanitization: It then iterates through the copied component array and strips out any pre-existing id property from each component object.
3. Clean Data Forwarding: The sanitized, clean array of components then replaces the original in the data payload before being processed.

By ensuring no component arrives at the database layer with a pre-existing ID, we force the entity service to always follow its correct behavior: creating a new, unique instance for each component and guaranteeing complete data isolation between entries.

**How I've Tested This**
I have thoroughly tested this solution against the conditions that caused the bug:
1. Creating new entries: Confirmed that creating Entry A and then Entry B results in two sets of independent components with unique database IDs.
2. Updating an existing entry: Confirmed that adding a new component to an existing entry does not affect any other entry.
3. Editing components: Confirmed that modifying a component in one entry has no effect on others.


oiorain:
Hi @moizDhila52 

Thank you for contributing to strapi and opening a PR. 
It seems your code is committed to a sub-folder though so if we merge it, it will have no effect on the codebase. 
Could you change so it targets the codebase for changes required to fix the issues, please? 

Thanks a lot!

moizDhila52:
Hi @oiorain ,

Thanks so much for the guidance! My apologies for the incorrect file paths. You were right, my local repository was set up in a sub-folder by mistake.

I've now fixed the commit and force-pushed the changes. The file paths should be correct now.

Thanks again for your help! 👍

oiorain:
Hi!
Thank you for the changes! 
We'll do our best to review your PR very soon.

moizDhila52:
Why not my PR is reviewed it's been a week madam...

If there's something then I will surely look into it

mathildeleg:
Thanks for the contribution and for taking the time to investigate this!

After reviewing this PR we're going to close it for a few reasons:
1. The linked issue was already closed as "can not reproduce" and was filed against Strapi v4.25.7. There's no confirmed bug to fix here.
2. The code doesn't compile. The `sanitizeComponentIds` function is placed as a const statement inside an object literal (between two method definitions), which is a TypeScript syntax error.
3. The fix is too narrow and risky even in principle. The sanitization only targets a field hardcoded as 'blocks'. More importantly, stripping component IDs on update would be harmful: existing components carry their ID so the entity service knows to update them rather than create new ones. Removing those IDs on every update would cause new component rows to be created on each save, orphaning the old ones.

PR #24573, opened 2025-10-13
feat(upload): upload file type restriction

araksyagevorgyan:
### What does it do?

This PR introduces MIME-type validation for file uploads in the Upload plugin.
It adds a centralized, configurable mechanism to restrict uploaded file types based on real MIME detection (using the file-type library).

Key technical changes:

- Added a new utility mime-validation.ts that detects the real MIME type from file bytes using file-type.
- Enforces allowedTypes and deniedType from plugin.upload.config.security.
- Returns structured validation results.
- When multiple files are uploaded, only valid files are uploaded; invalid files are silently filtered out and not uploaded.

### Why is it needed?

This feature addresses recurring unrestricted upload vulnerability reports.

With this change:

- File uploads are now validated based on their actual MIME signature, not the declared extension.
- Admins can explicitly control which file types are allowed.
- The system prevents accidental or malicious uploads (e.g., shell scripts, executables, disguised files).
- Provides a safer and more consistent experience

### How to test it?
Define your config in config/plugins.js:

```
export default () => ({
  upload: {
    config: {
      security: {
        allowedTypes: ['image/*', 'application/*'],
        deniedTypes: ['application/x-sh', 'application/x-dosexec']
      },
    },
    ...
  },
});
```

Testing steps:

- Start Strapi and open the Media Library.
- Upload a mix of files.
- Only valid files should appear in the Media Library.
- Invalid files will be silently ignored (not uploaded or stored) in the AI Media Library, while in the non-AI Media Library, errors will be displayed at the UI level for those files.

**Note:**
This implementation currently applies only to the backend.
UI feedback behavior differs between components:
- In the non-AI Media Library, file errors are displayed at the UI level for invalid uploads.
- In the AI Media Library, invalid files are silently filtered out and not uploaded.

Introducing unified UI feedback across both components will be considered in a future iteration, given the architectural differences between them. Moreover, introducing new UI logic is complex at this stage due to existing technical debt in the Upload plugin.

**Documentation:**

https://github.com/strapi/documentation/pull/2790

### Related issue(s)/PR(s)

https://github.com/strapi/content-squad/issues/108

### DIFF
--- a/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
+++ b/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
@@ -146,6 +146,25 @@ describe('mime-validation', () => {
       expect(isMimeTypeAllowed('image/jpeg', config)).toBe(true);
       expect(isMimeTypeAllowed('APPLICATION/pdf', config)).toBe(true);
     });
+
+    it('should handle hybrid allowlist and denylist correctly', () => {
+      const config: SecurityConfig = {
+        allowedTypes: ['image/*', 'application/*'],
+        deniedTypes: ['application/x-executable'],
+      };
+
+      // PDF should be allowed (matches application/* but not in denied list)
+      expect(isMimeTypeAllowed('application/pdf', config)).toBe(true);
+
+      // Images should be allowed (matches image/*)
+      expect(isMimeTypeAllowed('image/jpeg', config)).toBe(true);
+
+      // Executables should be denied (in denied list, despite matching application/*)
+      expect(isMimeTypeAllowed('application/x-executable', config)).toBe(false);
+
+      // Text files should be denied (not in allowed list)
+      expect(isMimeTypeAllowed('text/plain', config)).toBe(false);
+    });
   });
 
   describe('validateFile', () => {


packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,290 @@
...
+      const file = filesArray[index];
+      validFiles.push(file);
+      validFileNames.push(file.originalFilename || file.name);
+    } else if (result.error) {
+      errors.push({
+        file: filesArray[index],
+        originalIndex: index,
+        error: result.error,
+      });
+    }
```
innerdvations, line 283:
```suggestion
    } else {
      const error = result.error || {
        code: 'VALIDATION_ERROR' as const,
        message: 'File validation failed',
        details: {
          fileName: filesArray[index]?.name || filesArray[index]?.originalname || 'unknown',
          reason: 'Unknown validation failure',
        },
      };

      errors.push({
        file: filesArray[index],
        originalIndex: index,
        error,
      });
    }
```

Ensuring it's either valid or an error could simplify some of the code elsewhere

packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,290 @@
...
+export async function validateFiles(files: any, strapi: Core.Strapi): Promise<ValidationResult[]> {
+  const filesArray = Array.isArray(files) ? files : [files];
+
+  if (!filesArray.length) {
+    return [];
+  }
+
+  const config: SecurityConfig = strapi.config.get('plugin::upload.security', {});
+
+  if (!config.allowedTypes && !config.deniedTypes && !config.maxFileSize) {
```
innerdvations: We might want to validate them a bit stricter, like ensuring they're the right data types, otherwise users could get cryptic error messages when misconfigured

derrickmehaffy:
Generating an experimental to do some testing with: https://github.com/strapi/strapi/actions/runs/18690598762

packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,290 @@
+import { fileTypeFromBuffer } from 'file-type';
+import { readFile } from 'node:fs/promises';
+import type { Core } from '@strapi/types';
+
+export type SecurityConfig = {
+  allowedTypes?: string[];
+  deniedTypes?: string[];
+  maxFileSize?: number;
```
derrickmehaffy: I think this would conflict with our existing max file size setting?

packages/core/upload/server/src/utils/mime-validation.ts
derrickmehaffy: Agreed here, we should specify the error message based on the error hit.

derrickmehaffy:
Exp Version: `0.0.0-experimental.6188a65af997a4a8b190bbf104a9ba27ebf583c8`

derrickmehaffy:
Manual tests performed by me with the exp:

- [x] single file, too large: 
image

- [x] multi file, too large + valid file (originally I got an invalid creds error but that was likely unrelated to this PR, a refresh fixed that) 
image

- [x] single file, blocked file type 
image


So far I'd say this looks really good, especially like that we can use wildcards for format types :+1:

derrickmehaffy:
Honestly @araksyagevorgyan I'm not sure much more should be needed on the error messages, they seem pretty clear to me since we are throwing them as application errors. At least in the UI they appear and block just fine.

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: I think so. But I made the fields optional, so I assume that if someone fills them in, it's an intentional choice. I can remove it if you think it might be unclear for users. What do you think? 🤔

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: Would that mean I don’t need to specify
`code: 'MIME_TYPE_NOT_ALLOWED' | 'FILE_SIZE_EXCEEDED' | 'VALIDATION_ERROR'`,
since they would all be overridden by `code: 'VALIDATION_ERROR' as const` anyway?
I was thinking providing a more specific error would be better for handling. What do you think? 🤔

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: I modified the code. Now it will throw an `ApplicationError` if config fields don't have right data types.

araksyagevorgyan: On another note, I deleted maxFileSize from this pr since seems it already exists with other configuration.

packages/core/upload/server/src/utils/mime-validation.ts
innerdvations: Ahh, oops, my suggestion might have caused it to lose some specificity but that was a mistake on my part.

My main point was that I don't like the idea of having a function that takes an array, then splits it into valid/error but also potentially loses some entries into the void, seems like it's prone to subtle errors in the future.

araksyagevorgyan: I am not sure what you mean by saying potentially loses some entries as it will return valid and non valid files, nothing should be lost 🤔 
But the thing is that in terms of ui - it works on non-AI Media lib, meaning errors are shown on ui. The issue is that AI Media lib uses different component then non-AI one, which makes it hard to show errors on ui level but files are well filltered out from upload list.

innerdvations (approved):
I haven't tested, but code LGTM!

innerdvations (dismissed):

packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,279 @@
...
+    } else {
+      return undefined;
+    }
+
+    const result = await fileTypeFromBuffer(new Uint8Array(buffer));
+
+    return result?.mime;
+  } catch (error) {
+    return undefined;
+  }
```
markkaylor, line 59: Won't this just fail silently? Is that what we want?

packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,279 @@
...
+  let mimeDetectionFailed = false;
+
+  try {
+    detectedMime = await detectMimeType(file);
+  } catch (error) {
+    mimeDetectionFailed = true;
+    strapi.log.warn('Failed to detect MIME type from file', {
+      fileName,
+      error: error instanceof Error ? error.message : String(error),
+    });
```
markkaylor, line 129: Is this catch block gonna happen since detectMimeType catches and just returns undefined?

packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -0,0 +1,279 @@
+import { fileTypeFromBuffer } from 'file-type';
+import { readFile } from 'node:fs/promises';
+import type { Core } from '@strapi/types';
+import { errors } from '@strapi/utils';
+
+export type SecurityConfig = {
+  allowedTypes?: string[];
+  deniedTypes?: string[];
```
markkaylor, line 8: I'm a bit confused why we need both. Shouldn't allowed types be enough? if it's not in allowedTypes it's a denied type?

markkaylor: Oh because empty array allows everything so in that case you can allow all and select to deny?

markkaylor (dismissed):
Tested! looks good to me

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: I think it makes sense to support both. For example, if you only want to add one type to deniedTypes and allow everything else, you wouldn't need to maintain a long list of allowedTypes, and vice versa.

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: You are right, I changed detectMimeType to throw instead.

packages/core/upload/server/src/utils/mime-validation.ts
araksyagevorgyan: btw, they can also co-exist I think, I added the following test case and it passed:

```
it('should handle hybrid allowlist and denylist correctly', () => {
      const config: SecurityConfig = {
        allowedTypes: ['image/*', 'application/*'],
        deniedTypes: ['application/x-executable'],
      };

      // PDF should be allowed (matches application/* but not in denied list)
      expect(isMimeTypeAllowed('application/pdf', config)).toBe(true);
      
      // Images should be allowed (matches image/*)
      expect(isMimeTypeAllowed('image/jpeg', config)).toBe(true);
      
      // Executables should be denied (in denied list, despite matching application/*)
      expect(isMimeTypeAllowed('application/x-executable', config)).toBe(false);
      
      // Text files should be denied (not in allowed list)
      expect(isMimeTypeAllowed('text/plain', config)).toBe(false);
    });
  });
```

markkaylor: Nice!

markkaylor (dismissed):
Nice work, tested and looks good to me

PR #24577, opened 2025-10-14
fix: available locales when creating new lang entry

### LINKED ISSUE #23769: [Content Manager] Edit View - Issue with locales list
### Node Version

22.16.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.16.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

There's currently an issue in the edit view of the content manager when localization is enabled making the statuses of other versions of a collection type not available in the localization filter:

From an existing locale (all statuses good):

Image

From a new locale (we loose all statuses of other locales):

Image

It's a backend issue and after investigating a bit it's mainly because when we create a new locale for an existing object of the content manager, the object document (from useDocument) in LocalePickerAction (packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx from line 110) is empty (because not created already for the locale we want to provide). Because of that, we don't have the list of existing localizations defined for the document.

Console.log of documentId + document:
- When we are on an existing locale:

Image

_Thanks to the property localizations in document, we can map the locales and them to display their status in the list._

- When we are on a locale we just created (that doesn't exist yet):

Image

_Since the document is an empty object, we don't have the localizations attribute and so we can't map them in the list of locales._

ℹ️ For the filter, we're loading the list of locales documents defined in the Settings of the project and we map with the existing locales defined for the current document to see if a locale already exist and also to display its status (Draft / Published / Modified).

💡 Instead of loading the current document only, we could add and use a method to retrieve all localizations of a document (without any filter, just get the list and that's it) to use it here for the mapping.

### Steps to Reproduce

Video reproducing the bug:

https://github.com/user-attachments/assets/b310fa73-a246-4cb5-b90f-1ea055e01926

Steps to reproduce:

- Go to the content manager
- Choose any collection type that has a localized field
- Create a new entry
- Fill a few fields and Save / Publish
- Click on the localization list (right top, next to the little globe icon on the left side)
- Create a new locale
- Try to open again this filter of locales
- You will see that the base entry doesn't have the Published status anymore and the option is back to + Create {LANG} locale

### Expected Behavior

No matter if we create a new locale of are on an existing one, when we open the list of locales, we should be able to see the existing versions / locales of the current document.

### Confirmation Checklist

Adzouz:
### What does it do?

There was a bug with the list of locales when creating a new locale for an entry in the Content Manager.
We were missing the actual status of other available locales in the selector:

For example we have a published English version of an entry:
Screenshot 2025-10-14 at 14 57 17

We click on Create Spanish locale, then the list loose the status of all other available locales:
Screenshot 2025-10-14 at 14 57 24

### Why is it needed?

It was confusing and could let the user think that something happened with their other locales content. We need to always retrieve and display all available locales in the selector.

### How to test it?

- If you don't have a collection type with Internationalization, set one
- Create a few locales in Settings > Internationalization if you don't have many
- Go to the content manager and select the localized collection type
- Create an entry in any language and save (or publish)
- Click on the languages selector and click on create a new locale
- Open the selector again to see the status of previous entries created in other languages

Instead of:
Screenshot 2025-10-14 at 14 57 24

You should see:
Screenshot 2025-10-14 at 15 02 22

-> You can also try with a collection type without Draft and Publish (to test without the entry status)

### Related issue(s)/PR(s)

Resolves #23769 

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/services/documents.ts
+++ b/packages/core/content-manager/admin/src/services/documents.ts
@@ -139,8 +139,13 @@ const documentApi = contentManagerApi.injectEndpoints({
           params,
         },
       }),
-      invalidatesTags: (_result, _error, { collectionType, model }) => [
+      invalidatesTags: (_result, _error, { collectionType, model, documentId }) => [
         { type: 'Document', id: collectionType !== SINGLE_TYPES ? `${model}_LIST` : model },
+        {
+          type: 'Document',
+          id: collectionType !== SINGLE_TYPES ? `${model}_${documentId}` : model,
+        },
+        { type: 'Document', id: `${model}_ALL_ITEMS` },
         'RecentDocumentList',
         'CountDocuments',
         'UpcomingReleasesList',


mathildeleg:
Looks good but I'm still getting the following bug:
1. Create an entry for EN locale
2. Create an entry for FR locale => here the dropdown is displayed correctly with correct status for both entries ✅
3. Delete FR locale 
4. Redirection to list view
5. Click on EN entry
6. Open dropdown => here the dropdown still shows me the FR entry (which has been deleted) instead of the "Create French (FR) locale"
https://github.com/user-attachments/assets/92ed744c-ecb2-4da8-8769-16c58dd865ba
I'm not sure if your PR was also supposed to fix this so if not, no worries I can approve the PR :)

Adzouz:
> Looks good but I'm still getting the following bug:
> 
> 1. Create an entry for EN locale
> 2. Create an entry for FR locale => here the dropdown is displayed correctly with correct status for both entries ✅
> 3. Delete FR locale
> 4. Redirection to list view
> 5. Click on EN entry
> 6. Open dropdown => here the dropdown still shows me the FR entry (which has been deleted) instead of the "Create French (FR) locale"
>    [user-attachments/assets/92ed744c-ecb2-4da8-8769-16c58dd865ba](https://github.com/user-attachments/assets/92ed744c-ecb2-4da8-8769-16c58dd865ba)
>    I'm not sure if your PR was also supposed to fix this so if not, no worries I can approve the PR :)

Good catch! 👀
It was another issue but I think it was invisible before because the list was showing empty locales. It was a cache clearing issue. Let me know if it fixed it now 🙏

PR #24581, opened 2025-10-14
fix: add await to prevent redirect during test

Adzouz:
### What does it do?

We have a very very annoying flaky test recently that (too) often fails our pipelines.
After investigation and thanks to the Playwright trace, I figured out that at some point in the test, there's a redirect to the Content Manager home link. I think it's caused by the lack of await on the clickAndWait call (causing redirect to happen later). Actually since we're using navToHeader, the first clickAndWait is useless. We can just remove it.

### Why is it needed?

To remove a lot of frustration in contributors waiting for pipelines to pass and has to retrigger the failed tests (sometimes a few time).

### How to test it?

- I have to admit I have no 100% guarantee that it will fix the issue because it's hard to test but let's say I am 99% confident.

🚀

### DIFF
--- a/packages/core/upload/server/src/controllers/admin-upload.ts
+++ b/packages/core/upload/server/src/controllers/admin-upload.ts
@@ -113,7 +113,23 @@ export default {
     }
 
     const data = await validateUploadBody(body, Array.isArray(files));
-    const filesArray = Array.isArray(files) ? files : [files];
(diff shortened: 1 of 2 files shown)


remidej (approved):
That was such a silly issue 😭 

You're a legend for finding it 🚀

PR #24585, opened 2025-10-15
future(i18n): add i18n ai status

markkaylor:
### What does it do?

Update the document header actions API to accept a `customizeContent` with `icon` to display a popover for a status icon. I need feedback here on the API. I tried to follow what was done for dialog, but maybe it should be more explicit with a new type 'status' instead of leveraging type `icon` with `customizeContent`. In [the first commit](https://github.com/strapi/strapi/pull/24585/commits/ca1e9608af73088ba20d10f50d8d9ac67d4707af) I just had everything in the i18n plugin 

Use the new API to display the status of the AI settings for i18n in a popover

### Why is it needed?

So the user can see if AI is enabled and easily enable/disable it

### How to test it?

Go to a content-type with i18n
You should see the status icon in the document header actions
Hover it, it should display a popover based on your ai setting state with a link to settings

When AI translation is disabled
We see the status sparkle icon with instruction to enable + fill in from locale button

When AI translation is enabled
We the status sparkle icon with instruction to disable (no fill in from locale)

Remove your license, you should only see the standard 'fill in from another locale' button

### Related issue(s)/PR(s)

resolves https://github.com/strapi/content-squad/issues/97

TODO:

- add e2e tests ✅ 
- Update strapi docs

### DIFF
--- a/packages/plugins/i18n/admin/src/pages/SettingsPage.tsx
+++ b/packages/plugins/i18n/admin/src/pages/SettingsPage.tsx
@@ -27,7 +27,7 @@ import { CreateLocale } from '../components/CreateLocale';
 import { LocaleTable } from '../components/LocaleTable';
 import { PERMISSIONS } from '../constants';
 import { useGetLocalesQuery } from '../services/locales';
-import { useGetSettingsQuery, useUpdateSettingsMutation } from '../services/settings';
+import { useGetSettingsQuery, useUpdatei18nSettingsMutation } from '../services/settings';
 import { getTranslation } from '../utils/getTranslation';
 
 const SettingsErrrorTooltip = ({
@@ -66,7 +66,7 @@ const SettingsPage = () => {
     isLoading: isLoadingSettings,
     error: settingsError,
   } = useGetSettingsQuery();
-  const [updateSettings] = useUpdateSettingsMutation();
+  const [updateSettings] = useUpdatei18nSettingsMutation();
 
   const handleToggleChange = async (checked: boolean) => {
     try {
--- a/packages/plugins/i18n/admin/src/services/settings.ts
+++ b/packages/plugins/i18n/admin/src/services/settings.ts
@@ -8,7 +8,7 @@ const settingsApi = i18nApi.injectEndpoints({
       query: () => '/i18n/settings',
       providesTags: ['Settings'],
     }),
-    updateSettings: builder.mutation<UpdateSettings.Response, UpdateSettings.Request['body']>({
+    updatei18nSettings: builder.mutation<UpdateSettings.Response, UpdateSettings.Request['body']>({
       query: (data) => ({
         url: '/i18n/settings',
         method: 'PUT',
@@ -19,6 +19,6 @@ const settingsApi = i18nApi.injectEndpoints({
   }),
 });
 
-const { useGetSettingsQuery, useUpdateSettingsMutation } = settingsApi;
+const { useGetSettingsQuery, useUpdatei18nSettingsMutation } = settingsApi;
 
-export { useGetSettingsQuery, useUpdateSettingsMutation };
+export { useGetSettingsQuery, useUpdatei18nSettingsMutation };


packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -437,6 +447,44 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
...
+ * -----------------------------------------------------------------------------------------------*/
+
+interface HeaderActionStatusProps {
+  trigger: HeaderActionDescription['icon'];
+  children: React.ReactNode;
+}
+
+const HeaderActionStatus = ({ trigger, children }: HeaderActionStatusProps) => {
+  const [open, setOpen] = React.useState(false);
+  const debouncedOpen = useDebounce(open, 100);
```
remidej, line 462: is the debounce so users have time to bridge the gap with their mouse between the popover anchor and the content, before it closes?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -290,6 +369,28 @@ const FillFromAnotherLocaleAction = ({
...
+            variant={isAISettingEnabled ? 'alternative' : 'neutral'}
+            size="S"
+          >
+            <Earth />
+            <Sparkle />
+          </AITranslationStatusIcon>
+        </Box>
+      ),
+      customizeContent: () => (
+        <AITranslationStatus isAISettingEnabled={Boolean(isAISettingEnabled)} />
```
remidej: Do we have designs for what that should look like when isAISettingEnabled is false? Also when the feature is disabled, shouldn't users have access to the old fill in from other locale button?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -290,6 +369,28 @@ const FillFromAnotherLocaleAction = ({
     return null;
   }
 
+  const hasAIFutureFlag = window.strapi.future.isEnabled('unstableAILocalizations');
+  if (hasAIFutureFlag && isAIAvailable) {
+    return {
```
remidej: Why not separate the AI thing into its own action? I find it odd to have it under FillFromAnotherLocale. I believe action components can return null which should make it possible

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -411,6 +413,14 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
           );
         } else {
           if (action.type === 'icon') {
+            if (action.customizeContent) {
```
remidej: is there a precendent for the "customize" name in our APIs? I assume you considered renderContent too?

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
markkaylor: Exactly, sorry I should explain that in a comment

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
markkaylor: hmm that's a good point about fill in from another locale, as for the designs for disabled there are designs and it's the same popover but tells you to go to settings to enable it. I think that's the default case

markkaylor: Enabled vs Disabled:

https://github.com/user-attachments/assets/44faa8f1-8f8e-4c19-9356-5fc03358cb99

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
markkaylor: `customizeContent` was already there, I think actually its intended for `select` I agree the name isn't great but adding a new render prop also doesn't sound great. I can't change `customizeContent` though that would be a breaking change. Even thought it looks like it was added and not documented https://docs.strapi.io/cms/plugins-development/content-manager-apis#headeractiondescription. Which reminds I'll need to document whatever we decide here

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
markkaylor: Sure sounds good we can do that.

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
remidej: I was curious where it came from, all good then

markkaylor:
Looking in to a bug when the app starts with fresh db.

Screenshot 2025-10-16 at 09 41 29

update: resolved

markkaylor:
@remidej [I changed the api](https://github.com/strapi/strapi/pull/24585/commits/d546ef02cb63758f6e3f2ef6d0456d841dfaa834) for this case. I didn't like relying on type: icon + customizeContent. Instead you provide `status` object with `anchor` and `content`. This document actions API is already messy but I think this a cleaner solution than what I had previously, lmk what you think. Maybe `anchor` should be called something else and content should be optional 🤔 

edit:

I refactored again to

```
status?: {
  message: React.ReactNode;
  tooltip?: React.ReactNode;
}
```

markkaylor:
oops the tests are failing, I'll give it a look in a bit

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -244,6 +248,105 @@ const getDocumentStatus = (
   return docStatus;
 };
 
+/* -------------------------------------------------------------------------------------------------
+ * AISettingsStatusAction
+ * -----------------------------------------------------------------------------------------------*/
+
+const AITanslationStatusIcon = styled(Status)<{ isAISettingEnabled: boolean }>`
```
araksyagevorgyan: Just a small typo, should be **AITranslationStatusIcon**

remidej: please use the $ prefix on the prop, it's to ensure the prop doesn't end up on the dom element as an attribute

markkaylor: yep! thanks :)

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -244,6 +248,105 @@ const getDocumentStatus = (
...
+  display: flex;
+  gap: ${({ theme }) => theme.spaces[1]};
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+
+  // Disabled state
+  background-color: ${({ isAISettingEnabled, theme }) =>
+    !isAISettingEnabled && theme.colors.neutral150};
```
remidej: wouldn't this make the property value blank if ai is enabled?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -244,6 +248,105 @@ const getDocumentStatus = (
...
+
+  svg {
+    fill: ${({ isAISettingEnabled, theme }) => !isAISettingEnabled && theme.colors.neutral300};
```
remidej: same question

tests/e2e/tests/i18n/editview.spec.ts
remidej, line 1: thanks for kicking this off 🚀

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
markkaylor: I guess so, doesn't seem to be an issue but I can probably conditionally include the css rule itself which is probably cleaner

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -409,34 +415,88 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
...
+        aria-describedby="document-header-action-status"
+      >
+        <Box height="100%">{children}</Box>
+      </Popover.Anchor>
+      <Popover.Content
+        role="tooltip"
+        id="document-header-action-status"
+        side="bottom"
+        align="center"
+        style={{ width: '250px' }}
```
Adzouz: Can you use rem instead of px here please? 🙏

markkaylor: I don't mind using rem here but I'm confused. I used several px to rem calculators and they said the equivalent is 15.6rem but when I use that it is a lot smaller 🤔 

Screenshot 2025-10-17 at 10 01 30
Screenshot 2025-10-17 at 10 01 52

I have to do more like 25rem to get what I want but why?

tests/e2e/app-template/config/features.js
```diff
@@ -1 +1,5 @@
-module.exports = ({ env }) => ({});
+module.exports = ({ env }) => ({
+  future: {
+    unstableAILocalizations: env.bool('STRAPI_FEATURES_UNSTABLE_AI_LOCALIZATIONS', false),
```
remidej, line 3: Is there anything setting this flag to true in the CI config? Otherwise how would it run?

markkaylor: Yeah I think we've had this discussion before but didn't find a solution. It doesn't run in the CI.

markkaylor: Or maybe it runs them if I hard code this and remove the describeOnCondition

markkaylor: I pushed to try that 🤞

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
Adzouz: Rem are calculated based on the font-size set on the html element and the rule on html tag sets the font size to 10px:
Screenshot 2025-10-17 at 13 48 55
Screenshot 2025-10-17 at 13 49 02
So to have 250px you have to put 25rem 😊

Adzouz: It's pretty common to do that to ease the calculation of rem sizes so you don't need to use a calculator everytime and you can easily determine what size it will be in px!

Adzouz:
Looks good to me, as soon as the tests pass you'll have my approval chief 🫡🤝

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
markkaylor: Awesome thanks for the explanation! I was sure there was some logic behind it, 250px => 25rem mapped tooooo nicely.

tests/e2e/app-template/config/features.js
markkaylor: Unfortunately it didn't work I'm actually not sure if it's the future flag or if the license for the CI env does not have AI, but it's really slow to debug that. Are you cool with merging with this (I reverted back)?

remidej (dismissed):

Adzouz (approved):
Let's goo 🚀

PR #23528, opened 2025-05-16
Display error when loading email provider

FranckKe:
### What does it do?

Display custom email provider errors
Similar as https://github.com/strapi/strapi/pull/13092

### Why is it needed?

Currently in a process to write a custom email provider. 
There were errors in the provider, Strapi greets me with a generic "Could not load email provider" when starting in development mode. 
Got mislead thinking it was a setup error somewhere in the local provider setup (first time developing a custom provider).
Hopefully this will help someone in the futur!

### How to test it?

- Create a local email custom provider See https://docs.strapi.io/cms/providers#local-providers
- Add an error in the `index.js`
- Start Strapi

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/13092

PR #13092, opened 2022-04-08
Display error when loading upload provider

petersg83:
Fix https://github.com/strapi/strapi/issues/13052

PR #24588, opened 2025-10-15
Display error when loading email provider

FranckKe:
### What does it do?

Display custom email provider errors
Similar as https://github.com/strapi/strapi/pull/13092

### Why is it needed?

Errors present in the custom email provider throw a generic "Could not load email provider" when starting in development mode which is hard to debug and pinpoint where errors actually comes from. Hopefully this will help someone in the futur!

### How to test it?

- Create a local email custom provider See https://docs.strapi.io/cms/providers#local-providers
- Add an error in the `index.js`
- Start Strapi

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/13092
https://github.com/strapi/strapi/pull/23528

### DIFF
--- a/packages/core/email/server/src/bootstrap.ts
+++ b/packages/core/email/server/src/bootstrap.ts
@@ -38,7 +38,9 @@ const createProvider = (emailConfig: EmailConfig) => {
     provider = require(modulePath);
   } catch (err) {
     const newError = new Error(`Could not load email provider "${providerName}".`);
-    newError.stack = err.stack;
+    if (err instanceof Error) {
+      newError.stack = err.stack;
+    }
     throw newError;
   }
 


packages/core/email/server/src/bootstrap.ts
```diff
@@ -37,7 +37,9 @@ const createProvider = (emailConfig: EmailConfig) => {
   try {
     provider = require(modulePath);
   } catch (err) {
-    throw new Error(`Could not load email provider "${providerName}".`);
+    const newError = new Error(`Could not load email provider "${providerName}".`);
+    newError.stack = err.stack;
```
Adzouz: Can you just check if err in instance of Error?
```
if (err instanceof Error) {
  newError.stack = err.stack;
}
```

Adzouz (changes requested):
Just a small comment 😊

packages/core/email/server/src/bootstrap.ts
FranckKe: Done!

Adzouz: Thanks, I just updated the commit messages with the format we're using so the pipelines can run (if you're interested https://www.conventionalcommits.org/en/v1.0.0/)

Issue #20315, opened 2024-05-16
Email.template.reset_password translation is missing


MAN-Sendance:
### Required System information

- Node.js version: 20.11.0
- NPM version: 10.2.4
- Strapi version: 4.24.2
- Database: Postgres 14
- Operating system: linux-x64
- Is your project Javascript or Typescript: TypeScript

### Describe the bug

When visiting `Strapi Dashboard / Settings / Email Templates / Reset Password / Edit Template` the translation for `Email.template.reset_password` is missing.

### Steps to reproduce the behavior

see above

### Expected behavior

`Email.template.reset_password` should be available so that value is shown instead of key

### Screenshots

![edit email template](https://github.com/strapi/strapi/assets/129411192/4426b826-8cfd-48fb-b6d3-5f1d2c0b778d)

### Code snippets

N/A

### Additional context

N/A

oiorain:
This is still happening in 5.27.0.

<img width="841" height="520" alt="Image" src="https://github.com/user-attachments/assets/13dcb6bf-6e73-44e0-a0f6-2ac3c8248233" />

mathildeleg:
Hi, this was fixed here https://github.com/strapi/strapi/pull/24589 :)

PR #24589, opened 2025-10-15
fix: add missing translation for email template title

mathildeleg:
### What does it do?
Adds the missing translation for the "Reset Password" email template.

_Before_
Before

_After_
After

### Why is it needed?
The key for the translation is no longer shown, instead a default value is.

### How to test it?
* Go the Settings
* Click on "Email templates" under the Users & Permissions Plugin tab
* Click on "Reset Password"
* The modal opens with the correct translation value 

### Related issue(s)/PR(s)
Fixes this [issue](https://github.com/strapi/strapi/issues/20315)


Adzouz (approved):
Approved 🚀

PR #24593, opened 2025-10-15
CI: action to identify duplicate issues

oiorain:
# What does it do?
This PR adds a new GitHub Actions workflow that automatically detects and flags duplicate issues using Claude AI. The workflow:
- Can be triggered manually with an issue ID as input
- Uses the anthropics/claude-code-action@v1 to analyze issues
- Leverages Claude to search for similar existing issues using GitHub MCP tools
- Automatically adds a comment on duplicate issues explaining why they're duplicates
- Applies a "flag: duplicate" label to confirmed duplicates
- Does not close issues automatically -> this is handled by a different script already.

# Why is it needed?

This workflow helps :
- Reduce manual effort in identifying duplicate issues
- Improve issue triage efficiency
- Maintain a cleaner issue tracker by linking related/duplicate issues

Todo: add a dedicated ANTHROPIC_API_KEY so we can analyse how much we spend on this action.

### DIFF
--- a/.github/workflows/find_duplicate_issue.yml
+++ b/.github/workflows/find_duplicate_issue.yml
@@ -73,6 +73,6 @@ jobs:
 
             Be thorough but efficient. Focus on finding true duplicates, not just similar issues.
 
-          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          anthropic_api_key: ${{ secrets.DUPLICATE_ISSUE_ANTHROPIC_API_KEY }}
           claude_args: |
             --allowedTools "mcp__github__get_issue,mcp__github__search_issues,mcp__github__list_issues,mcp__github__add_issue_comment,mcp__github__update_issue,mcp__github__get_issue_comments"


.github/workflows/find_duplicate_issue.yml
```diff
@@ -0,0 +1,78 @@
...
+        description: 'Id of the issue to check for duplicates'
+        required: true
+
+jobs:
+  deduplicate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      issues: write
```
Baldinof, line 16: ~~Does it need the write permission on gh issues?~~

Edit: yes, Claud will comment/manage the duplicates

.github/workflows/find_duplicate_issue.yml
Baldinof, line 1: It could be nice to ask Claude to keep track of action, and maybe generate a report.

The report could be added as a [summary](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/#create-summaries) of the workflow run

oiorain: oh yeah that's a great idea actually! 
It's kind of a pain to go dig in the logs. i'll implement that. I need to wait for @alexandrebodin to come back for the dedicated licence anyway.

.github/workflows/find_duplicate_issue.yml
```diff
@@ -0,0 +1,78 @@
...
+            Use these tools:
+            - mcp__github__get_issue: Get issue details
+            - mcp__github__search_issues: Search for similar issues
+            - mcp__github__list_issues: List recent issues if needed
+            - mcp__github__add_issue_comment: Add a comment if duplicate found
+            - mcp__github__update_issue: Add labels
+
+            Be thorough but efficient. Focus on finding true duplicates, not just similar issues.
+
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
```
alexandrebodin: You can use DUPLICATE_ISSUE_ANTHROPIC_API_KEY I'll set it up in the config

alexandrebodin (changes requested):

.github/workflows/find_duplicate_issue.yml
oiorain: So funny thing: the current task doesnt actually leave any trace in the script other than the comment on the issue. The script would have to go read the comment to produce the summary so I decided to let it go for this task. but I think it will be very helpful for others.

.github/workflows/find_duplicate_issue.yml
oiorain: thanks! I updated the key finally.

PR #24594, opened 2025-10-15
CI: check template on new issue

oiorain:
### What does it do?

On opening a new issue, this script checks automatically the bug report template is respected. 
It looks for 
- Node Version
- NPM/Yarn/PNPM Version
- Strapi Version
- the two last checkboxes being checked.

It does nothing when it finds these conditions are filled. 
if that's not the case, it applies the `flag: invalid template` then closes the issue. 
-> I removed the closing and message from the former script to have everything in one place. 

### Why is it needed?

This could help reduce the manual labor of triaging issues.


.github/workflows/template-check-on-new-issue.yml
```diff
@@ -0,0 +1,161 @@
...
+
+jobs:
+  check-checkboxes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check issue uses template
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const issueBody = context.payload.issue.body;
```
markkaylor: I'm just wondering if we could have the javascript in its own file instead of represented as a string here. 

I think check-pr-status is like that and then referenced in https://github.com/strapi/strapi/blob/8bb90e64ade28efe46558e7942b75508481141ab/.github/workflows/checks.yml#L21 but maybe we could do something similar and just have a script in the /scripts folder?

markkaylor (commented):
How can we test this 🤔 ?

markkaylor (approved):
I think we can start with this so that old issues can still be closed by manually adding the flag. Ideally we write a script to close all old issues with an invalid template. When we do that we can also consolidate adding the message and closing the issue to one spot.

.github/actions/check-issue-template/index.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+
+      // Extract the content between the header and next section
+      let content = match[0];
+      // Remove the header line
+      content = content.replace(/^### .*?\n/i, '');
+      // Remove the next section header if present
+      content = content.replace(/\n### .*$/i, '');
+      // Remove the last section header if it's the end
+      content = content.replace(/\n### .*$/i, '');
```
innerdvations, line 97:
```suggestion
      // Remove the next section header if present
      content = content.replace(/\n### .*$/is, '');
```

I think this will work better to replace all of them

.github/actions/check-issue-template/index.js
```diff
@@ -0,0 +1,141 @@
...
+      ) {
+        missingItems.push(field.label);
+      }
+    }
+
+    // If required items are missing, add the label and comment
+    if (missingItems.length > 0) {
+      const token = process.env.GITHUB_TOKEN;
+      const octokit = github.getOctokit(token);
+
```
innerdvations, line 118:
```suggestion
    if (missingItems.length > 0) {
      const token = process.env.GITHUB_TOKEN;
      const octokit = github.getOctokit(token);
      core.info(`Missing required items: ${missingItems.join(', ')}`);
```

Could help debugging

innerdvations (approved):

innerdvations (approved):
minor suggestions, but looks good!

.github/workflows/template-check-on-new-issue.yml
```diff
@@ -0,0 +1,21 @@
...
+  check-checkboxes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Check issue uses template
+        uses: ./.github/actions/check-issue-template
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
innerdvations, line 21:
```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          continue-on-error: true
```

innerdvations: this will keep the whole workflow from stopping when this errors

Bassel17:
would it be better if instead of closing the issue we put the invalid tag and post an automated message telling the users to respect the template or the issue will be deleted ?
we can always have a cron job cleaning those tagged issues once a week.

derrickmehaffy (approved):

PR #24602, opened 2025-10-16
fix: pass context to cookie options in authentication and session-auth utilities

jhoward1994:
Changes cookie secure flag logic from NODE_ENV === 'production' to NODE_ENV === 'production' && ctx.request.secure to support reverse proxy deployments.

Related issue(s)/PR(s)

  - #24452
  - Related to PR #24346 (introduced refresh token sessions)

### DIFF
--- a/packages/core/admin/server/src/controllers/authentication.ts
+++ b/packages/core/admin/server/src/controllers/authentication.ts
@@ -276,7 +276,11 @@ export default {
       ).generateRefreshToken(userId, deviceId, { type: 'session' });
 
       // No rememberMe flow here; expire with session by default (session cookie)
-      const cookieOptions = buildCookieOptionsWithExpiry('session', absoluteExpiresAt, ctx.request.secure);
+      const cookieOptions = buildCookieOptionsWithExpiry(
+        'session',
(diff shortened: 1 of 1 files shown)


innerdvations:
Should resolve https://github.com/strapi/strapi/issues/24452 for a lot more people

packages/core/admin/server/src/controllers/authentication.ts
```diff
@@ -88,7 +88,8 @@ export default {
 
         const cookieOptions = buildCookieOptionsWithExpiry(
           rememberMe ? 'refresh' : 'session',
-          absoluteExpiresAt
+          absoluteExpiresAt,
+          ctx
```
innerdvations: do you think it's worth passing the whole ctx or should it just pass something like `secureRequest` ?

jhoward1994: Yeah thats a good idea done in [`2105bb9` (#24602)](https://github.com/strapi/strapi/pull/24602/commits/2105bb9bd6163e58a868decb80bae972f79b571f)

PR #24604, opened 2025-10-16
enhancement: add en language for Japan

pranavpinarayi:
### What does it do?

Added en-JP language for Japan region.

### Why is it needed?

To add support for new locale

### How to test it?

Go to strapi settings page and enable en-JP locale. https://docs.strapi.io/cms/features/internationalization#settings

### DIFF
--- a/packages/plugins/i18n/server/src/constants/iso-locales.json
+++ b/packages/plugins/i18n/server/src/constants/iso-locales.json
@@ -632,6 +632,10 @@
     "name": "English (Jamaica) (en-JM)"
   },
   {
+    "code": "en-JP",
+    "name": "English (Japan) (en-JP)"
+  },
+  {
     "code": "en-JE",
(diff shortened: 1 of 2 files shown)


innerdvations:
Thanks for you PR!

However, in general we only support comon language + region combinations so unfortunately we cannot accept this change.

See https://www.unicode.org/cldr/charts/47/supplemental/territory_language_information.html for a list of standardized locale combinations

innerdvations:
Ok, after some internal discussion we've decided to add this exception! But someone beat you to it with an earlier PR doing the same thing 😆

pranavpinarayi:
> Ok, after some internal discussion we've decided to add this exception! But someone beat you to it with an earlier PR doing the same thing 😆
@innerdvations 
Nice. I just saw that PR. But that is also not merged right? Are you planning to add this?

Issue #23765, opened 2025-06-18
[BUG][Strapi 5][I18N] field of type number changes on different locale


namelleess:
### Node Version

18.20.5/22.8.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.7.0/5.16.0

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When modifying any numerical input field for a specific content entry within 'ru' locale and then saving these changes, the updated value for said input is unexpectedly reflected in other locales ("en" in my case) for the same content entry

### Steps to Reproduce

1. Create a Localized Content Type with a Number Field
2. Create a New Entry for the Content Type
3. Populate Number Data Across Locales and Save:
    - For the number field, enter a unique number for each available locale

      Example:
         English: 100
         Russian: 200
 
     - Click the "Save" button to persist these initial values
4. Refresh the entire page in your browser _Do not just navigate between locales within the same session after saving_
5. Change Number in One Locale and Save:
   - Select specific locale (e.g English)
   - Change the number field value to a new arbitrary number (e.g., from 100 to 123) and save it
6. Observe Visual Synchronization in Another Locale (Without Refreshing):
   - Immediately after saving in the previous step, switch to a different locale (e.g., Russian) using the locale dropdown
   - Observe the number field being set to 123 instead of being 200

note that this is only the ui bug, once you refresh the page all the data is back to normal

### Expected Behavior

above in steps to reproduce, the number field value for Russian locale should not have been changed by the changes in English locale

### Logs

```shell

```

### Media

video showcasing the bug with strapi v5.7.0 and node v18.20.5

https://github.com/user-attachments/assets/b1b1ee65-f0ca-4bdc-bda5-de09290b6437

simpler example with strapi v5.16.0 and node v22.8.0

initial state, in russian number set to 1 and in english is set to 2

![Image](https://github.com/user-attachments/assets/24ad76cc-3f69-4b48-a888-8dce893bfe33)

![Image](https://github.com/user-attachments/assets/460711ee-f87f-4df2-a9a6-9a6d79249c2b)

in russian locale change number to 3 and hit save

![Image](https://github.com/user-attachments/assets/209650e3-de09-4d97-b509-255c7a754c39)

see that english locale is now also 3 but supposed to be 2 since not changes were made for english locale

[...]

Issue #21063, opened 2024-08-23
Plugin i18n switching locales leads to wrong data in markdown and custom fields


Xaz16:
### Required System information

- Node.js version: node/v18.16.0
- NPM version: npm/9.5.1
- Strapi version: v4.25.8
- Database: sqlite
- Operating system: linux-x64
- Is your project Javascript or Typescript: js

### Describe the bug

When you use localizations for your entries and switching between locales AND you have empty markdown field in target locale you will see same content from previous locale. For example you have collections type with 2 fields; title (short text), body (markdown text) and 2 locales: en as default and de as additional one. You created one entry in default language (en) and filled these fields with this text: 'en content'. Then you are switching to de locale and see such situation: title was properly cleared but body still have 'en content' inside. When you update the page all content are disapearing and you see proper state of not filled by any text locale.
I recorded these situation, you can check it in Screenshots

As well same thing is applicable for custom fields plugins. If you will switch between locales, your plugin component wil be rerendered, but value will be the same from previous locales. It is create not understanble situations.

And everything could be fixed with just refreshing the page... 

### Steps to reproduce the behavior

I think it is already good described in bug description, as well all steps was recorded (check it below)

### Expected behavior

When I switching between locales I should see exact content of locale from database. If I just created another locale entry and don't type anything, I should see empty fields

### Screenshots

![Animation](https://github.com/user-attachments/assets/f5fb5154-5950-4d87-ba53-fe6c480a7a8f)

### Code snippets

Repo with reproduce: https://github.com/Xaz16/strapi-locale-issues

### Additional context

No context, it is reproducable from newly created strapi project

Xaz16:
Hello!
Maybe it is not clear or I don't provide something important. Could you elaborate on that issue? It is quite critical for us, and we want to migrate to strapi

Xaz16:
The same issue in strapi v5
Most of them in on-the-fly relation editing in modals. Will add reproduce later

Xaz16:
Folks, I understand that you have lots of important things to fix. But this bug is getting worse and create so much confusion...
Even on newer versions. You just trying to clear field and click publish, but you see stale content in this field. Please, fix this, it is really crucial 
@derrickmehaffy 

You application is SPA, but you can't check actual data until you click reload button

![Image](https://github.com/user-attachments/assets/a5fbf9a7-7079-4c4e-b943-cf63f1b5250d)

oiorain:
Confirmed this is still an issue in v5.27

[...]

PR #24605, opened 2025-10-16
fix: rerendering and flickering issue on i18n fields

### LINKED ISSUE #20409: i18n locale overrides other locales

- Node.js version: 18.19.0
- NPM version: 10.5.1
- Strapi version: 4.24.3
- Database: postgres
- Operating system: Mac OS darwin-x64
- Is your project Javascript or Typescript: typescript

### Describe the bug
Hi all,
I have started creating a new app and opted to use Strapi, however I have an issue with i18n.

The issue is that when I try to change internationalization I manage to select one in the dropdown but fields (which do have an internationalization enabled icon) are prefilled with data from the old locale. when i change and save something the original locale changes as well and i don’t have a different version of the locale in my db.
Also when i click fill with another localization I have nothing to choose from.

Some other fields work as expected and change the localization normally.

It's hard to reproduce since i've come across it only on two fields (one collection and one single type).
create a type (internationalization true) go to content manager create a collection and sometimes that collection has options in one dropdown so change language but whenever i change something it mirrors to the other language.

### Expected behavior
When internationalization changes present me with another language version or empty if it wasn't filled previously. Fill with another language should offer me another language but it doesn't.

### Screenshots

Screenshot 2024-05-31 at 11 56 08
Screenshot 2024-05-31 at 11 56 19
Screenshot 2024-05-31 at 11 56 28
Screenshot 2024-05-31 at 12 03 42

### Code snippets

`{
  "kind": "singleType",
  "collectionName": "footer_bars",
  "info": {
    "singularName": "footer-bar",
    "pluralName": "footer-bars",
    "displayName": "FooterBar"
  },
  "options": {
    "draftAndPublish": true
  },
  "pluginOptions": {
    "versions": {
      "versioned": true
    },
    "i18n": {
      "localized": true
    }
  },
  "attributes": {
    "nav_items": {
      "type": "relation",
      "relation": "oneToMany",
      "target": "api::navbar-item.navbar-item"
    },
    "socialsLabel": {
      "pluginOptions": {
        "versions": {
          "versioned": true
        },
        "i18n": {
          "localized": true
        }
      },
      "type": "string",
      "required": true
    }
  }
}
`

Adzouz:
### What does it do?

_I know it's related to admin + content manager package but flagging it as plugin:i18n to better track the issue._

There are a few weird behaviours currently with i18n fields.
The first is that a few fields don't rerender correctly with the correct data when switching locales right after a save and can cause wrong data to be overriden in the wrong locale (this one is pretty serious).

Video to show the problem:

https://github.com/user-attachments/assets/a0d56d86-f88f-4847-9083-29f8a3f08eae

The second is a little flickering when clicking on Publish after editing fields in an entry (caused by a resetForm action triggered when publishing entries).

Video to show the problem:
_(focus on the text field to see the flickering issue when publushing)_

https://github.com/user-attachments/assets/52b216c1-61d4-4d3a-839e-9c8bab0bcf5f

### Why is it needed?

It would prevent data from a locale to override content of another one in translated fields and prevent flickering issue displaying the old value for a second when clicking on Publish button.

### How to test it?

For the first issue, it's tricky but here are the steps:
- Go to the content manager and select a collection type where Draft and Publish and Internationalization are enabled and where there's a number field (also localized)
- Create a new entry in the first locale and put a number
- Save and publish
- Switch to another locale
- Put another value for the number and save + publish too
- Switch back to the first locale, change the number and save (put another number than the one you choose for the second locale to see the difference)
- Switch to the second locale again and check the value in the number field
-> Before the fix, the value would be wrong and the one you set on the first locale
-> After the fix, the value is the one you should see for the second locale

For the second issue (flickering):
- Go to the content manager and again select the same collection type than for the first test
- Put some content in any field (a text one for example)
- Click Save and Publish
- Edit the value of the field again and click Publish
-> Before the fix, for a second you'll see the previous value
-> After the fix, no more flickering, the value stays the latest

### Related issue(s)/PR(s)

Resolves #20409 #21063 #23765

🚀


packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -306,7 +306,8 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
 
     const modified = React.useMemo(
       () => !isEqual(initialValues.current, state.values),
-      [state.values]
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [state.values, initialValues.current]
```
Adzouz, line 310: For an obscure reason I need to add the eslint disable because I do need to watch changes on `initialValues.current` but I don't know why it's complaining that it's not part of dependencies 🤷‍♂️

mathildeleg (dismissed):

packages/core/admin/admin/src/components/Form.tsx
HichamELBSI: eslint is mad because `initialValues` is a ref which means that initialValues persist across renders. Updating `.current` of a ref won't trigger the useEffect. This seems wrong to me.

packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
```diff
@@ -8,18 +9,23 @@ import { useField } from '../Form';
...
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
     const { formatMessage } = useIntl();
     const field = useField<boolean | null>(name);
     const fieldRef = useFocusInputField<HTMLInputElement>(name);
+    const [{ query }] = useQueryParams<{ plugins?: { i18n?: { locale?: string } } }>();
 
     const composedRefs = useComposedRefs(ref, fieldRef);
 
+    // Create a key that changes when locale changes to force re-render
+    const localeKey = query?.plugins?.i18n?.locale || 'default';
```
HichamELBSI: Do these changes work correctly on a collection type that is not internationalised?

HichamELBSI: Same question for all the inputs.

packages/core/admin/admin/src/components/Form.tsx
Adzouz: Ah good catch. So what should I do? Because the resetForm is resetting to the initialValues for a second causing the flickering effect so I didn't really have other idea than setting initialValues.current to the new values after saving / publish (which actually makes sense for me since the initial values become the new ones).

packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
Adzouz: Yes it will fallback to default and since we can't switch between languages, we don't have the rerender issue. But the only thing I'm not really fan of is that it's referring to the i18n plugin in the admin package but I don't really have other idea on how to do it in another way since it's really the switching between two locales that causes the issue.

Adzouz: I tried a different approach (the hook). I put the logic to fetch the locale directly in the i18n plugin and exported it as a hook that I load in both admin and the CM. Then I created a useLocaleKey in both packages to call the hook. Does that makes sense to you?

packages/core/admin/admin/src/components/Form.tsx
HichamELBSI: I've just tested without this update and I can't see any flickering 🤔

packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
```diff
@@ -1,25 +1,28 @@
...
 
 import { InputProps } from './types';
 
-const BooleanInput = forwardRef<HTMLInputElement, InputProps>(
+const BooleanInput = React.forwardRef<HTMLInputElement, InputProps>(
   ({ name, required, label, hint, labelAction, ...props }, ref) => {
     const { formatMessage } = useIntl();
     const field = useField<boolean | null>(name);
     const fieldRef = useFocusInputField<HTMLInputElement>(name);
+    const localeKey = useLocaleKey();
```
HichamELBSI: The fact that there is a mention of locale in the fields is a problem. On a content type that does not have the i18n feature enabled, this hook will be called in all inputs. This is unnecessary. We have the information on whether i18n is enabled or not in the schema.pluginOptions.i18n object. 

Another big issue is that the hook is called in all dynamic forms of the app. For example, in the profile page.

packages/core/admin/admin/src/components/Form.tsx
Adzouz: Yes but this part is not about the flickering. If you remove this part, the modified status won't change after submitting the form, keeping it in modified status even after publishing / saving 😕

HichamELBSI: Alright, I'll test it again

packages/core/admin/admin/src/components/FormInputs/Boolean.tsx
Adzouz: Yes I know, I thought that using a hook would be "cleaner" because if i18n isn't available, it will just fallback to default but I understand your point, I'll try to find another solution. I'll let you know if I'm stuck. Thanks for your patience 🙏

packages/core/admin/admin/src/components/Form.tsx
Adzouz: I could use another state variable to store the actual version of the values so it wouldn't rely on a ref anymore. But I don't find it much cleaner tbh. Wdyt?

Adzouz:
I tried to go with a simpler solution and scope it for edit view page only (getting rid of the hook + removing everything in the field children elements).

Issue #23835, opened 2025-06-27
Misleading error message when publishing


McSodbrenner:
### Node Version

20.14.0

### NPM/Yarn/PNPM Version

10.9.1

### Strapi Version

5.16.1

### Operating System

MacOS

### Database

MariaDB

### Javascript or Typescript

Javascript

### Bug Description

When I publish an item, but a required field is not filled in, I get the following error message:

> Your current permissions prevent access to certain required fields. Please request access from an administrator to proceed.

However, I am a super admin, so this statement cannot be correct. When searching through the code, I was puzzled by the relevant part of the code:

https://github.com/strapi/strapi/blob/44d360536462e83bfc95c01fd21384d2a42d321f/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx#L704

The problem for me seems to be that when determining `hasUnreadableRequiredField` there are incorrect values in `canReadFields` when fields come from components.

In my example, the field "metadata" is defined on this item:
```
    "metadata": {
      "type": "component",
      "repeatable": false,
      "pluginOptions": {
        "i18n": {
          "localized": true
        }
      },
      "component": "landingpage.meta",
      "required": true
    },
```

canReadFiles contains in my case:
```
[
    "translate",
    "title",
    "navigationTitle",
    "path",
    "foregroundColor",
    "backgroundColor",
    "metadata.description",
    "metadata.opengraphTitle",
    "metadata.opengraphDescription",
    "blocks",
    "registry.sitemapShow",
    "registry.navigationShow",
    "registry.navigationRoot",
    "registry.navigationImportant",
    "registry.navigationWeight",
    "overrides.landingpage",
    "overrides.landingpageStart",
    "overrides.landingpageEnd"
]
```

When checking the fields, the metadata field is checked, but this is not contained in canReadFiles. Hence the wrong error.

### Steps to Reproduce

1. Create an item with a component in its definition
2. Publish an item while some required fields are not filled  

### Expected Behavior

On error the page should scroll to the problematic field and not display a misleading error dialog.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

manojLondhe:
I am also seeing same issue with recent Strapi upgrade v5.16.1

[...]

PR #24608, opened 2025-10-16
fix: misleading error message when publish empty required components

Adzouz:
### What does it do?

There's a misleading error when trying to publish a component that is either required or has required fields with empty fields.

Screenshot 2025-10-16 at 16 44 47

The message is about permissions while it's supposed to be about field validation error.

### Why is it needed?

Helping the user correctly understanding the actual issue when trying to publish a document.

### How to test it?

- Go to the content manager and select a collection type that has a component (repeatable or not) that is required
- Create an entry and try to publish without adding any component
-> The error should be the following after the fix:
Screenshot 2025-10-16 at 16 48 43

- You can also test with a non required component but with a required field in it since it has the same issue

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/23835

🚀

### DIFF
--- a/tests/e2e/tests/content-manager/edit-view/collection-type-edit-view-errors.spec.ts
+++ b/tests/e2e/tests/content-manager/edit-view/collection-type-edit-view-errors.spec.ts
@@ -12,7 +12,7 @@ test.describe('Edit View', () => {
   });
 
   test.describe('Collection Type - Errors', () => {
-    test.only('as a user I should see an error when trying to publish a document with a required component that is not filled in', async ({
+    test('as a user I should see an error when trying to publish a document with a required component that is not filled in', async ({
       page,
     }) => {
       await navToHeader(page, ['Content Manager', 'Match'], 'Match');


tests/e2e/tests/content-manager/edit-view/collection-type-edit-view-errors.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+test.describe('Edit View', () => {
+  test.beforeEach(async ({ page }) => {
+    await resetDatabaseAndImportDataFromPath('with-admin.tar');
+    await page.goto('/admin');
+    await login({ page });
+  });
+
+  test.describe('Collection Type - Errors', () => {
+    test.only('as a user I should see an error when trying to publish a document with a required component that is not filled in', async ({
```
HichamELBSI:
```suggestion
    test('as a user I should see an error when trying to publish a document with a required component that is not filled in', async ({
```

HichamELBSI (changes requested):

tests/e2e/tests/content-manager/edit-view/collection-type-edit-view-errors.spec.ts
Adzouz: Oops good catch

HichamELBSI (approved):

PR #24612, opened 2025-10-16
future(i18n): trigger ai localizations

remidej:
### What does it do?

- Creates an "article" content type on getstarted. Because we need to test the feature with a content type that has various field types, but i18n enabled (unlike kitchensink), and localization turned off for some fields
- Creates a `generateDocumentLocalizations` service method, designed so it can be called for single or collection types, both on create and update
- Call this method in the background upon collection type and single type create/updates.
- ⚠️ block fields are ignored for now, I'll work on handling them better inside the AI service in parallel, then we can re-add them

It's the sort of logic that would benefit a lot from API tests. We probably would have to mock the AI server response though.
### How to test it?

- Make an update on a collection types, see the other locales it generated
- Ensure it doesn't generate anything when it shouldn't (no license, i18n not enabled, or when the update is not on the source locale)

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -131,6 +131,7 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
 
       // Get all media field names dynamically from the schema
       const mediaFields = Object.entries(schema.attributes)
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
         .filter(([_, attr]) => attr.type === 'media')
         .map(([key]) => key);
 


examples/getstarted/src/api/test-locale/content-types/test-locale/schema.json
remidej, line 1: @araksyagevorgyan Could you make this new single type something more like a "real" content type? So that it can be used to test more generic things in the future. Currently we have a "Homepage" single type, maybe something like "About", "Config", "Footer" etc. could make sense

araksyagevorgyan:
@remidej 
I've pushed a commit that resolves the issue we had with media generation.
It adds the same media when there’s no media for the derived locale and doesn’t affect cases where a media already exists for derived ones.

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -127,17 +127,40 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
       }
 
       const aiResult = await response.json();
-      console.log('********* AI RESULT', JSON.stringify(aiResult, null, 2));
+
+      // 1. Get all media field names dynamically from the schema
+      const mediaFields = Object.entries(schema.attributes)
```
remidej, line 133: I would recommend building a new object using traverseEntity, the util we have above. Because traverseEntity recursively goes through the document. A media field could be inside a component for example, not always at the root

remidej (commented):
I added a comment but I think we should try to merge it whenever possible, and make improvements in other PRs, because otherwise it blocks Mark's work.

@markkaylor could you look at this in the morning?

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -24,6 +28,146 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+          // TODO: remove blocks from this list once the AI server can handle it reliably
+          if (!hasLocalizedOption || ['media', 'blocks'].includes(attribute.type)) {
+            remove(key);
+          }
+        },
+        { schema, getModel: strapi.getModel.bind(strapi) },
+        document
+      );
+
+      // 3. Call the AI server to get the localized content
```
markkaylor:
```suggestion
      // Call the AI server to get the localized content
```

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -24,6 +28,146 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+      if (!response.ok) {
+        strapi.log.error(
+          `AI Localizations request failed: ${response.status} ${response.statusText}`
+        );
+        throw new Error(`AI Localizations request failed: ${response.statusText}`);
+      }
+
+      const aiResult = await response.json();
+
+      // 1. Get all media field names dynamically from the schema
```
markkaylor:
```suggestion
      // Get all media field names dynamically from the schema
```

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -24,6 +28,146 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+      if (targetLocales.length === 0) {
+        strapi.log.info(
+          `AI Localizations: no target locales for ${schema.uid} document ${documentId}`
+        );
+        return;
+      }
+
+      let token: string;
+      try {
+        const tokenData = await strapi.service('admin::user').getAiToken();
```
markkaylor, line 101: Does getAiToken not throw its own error? Do we need this try catch?

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -1,6 +1,10 @@
-import type { Core } from '@strapi/types';
+import type { Core, Modules, UID } from '@strapi/types';
+import { traverseEntity } from '@strapi/utils';
+import { getService } from '../utils';
 
 const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
+  const aiServerUrl = process.env.STRAPI_AI_URL || 'https://strapi-ai.apps.strapi.io';
```
markkaylor, line 7: I think we should probably have a helper function for this somewhere so all ai features can use it. If you agree Maybe just add a TODO for now?

markkaylor (approved):
Looks good to me. Tested and looks like it's working. I left some some small comments, but I agree with Rémi let's try and merge this around noonish so I can update my branch and then we can PP on a small refactor to strapi.documents.use middleware. We can also handle the comments on this PR in subsequent PRs.

Issue #24606, opened 2025-10-16
[bug]: Unusable "responsive" aside menu =(((


ztxone:
### What version of design-system are you using?

Strapi version 5.27.0 and 5.28.0

### What's Wrong?

Absolute unusable situation with "responsiveness" of aside menu which takes "auto" size of the screen...

<img width="1176" height="459" alt="Image" src="https://github.com/user-attachments/assets/d70f3deb-6a94-4b1e-a03a-7d624c7c0a7a" />
<img width="304" height="101" alt="Image" src="https://github.com/user-attachments/assets/0691f78e-bf68-4e01-8987-4d5e04090050" />

grid-template: auto 1fr;  - are you kidding? oO

90% of the time people works with content area of the screen, not with the aside menu

### To Reproduce

Create long name for content collection

### Expected Behaviour

Aside panel with list of collections must be resizable OR at least foldable. OR there is should be implemented truncation for the CT names.

HichamELBSI:
Thanks for your kind feedback @ztxone, we will take the time to test it.

PR #24616, opened 2025-10-17
fix: responsive subnav

HichamELBSI:
### What does it do?

Set the subnav width to the previous max-width (23.2rem) It fixes the subnav when link title are too long on desktop and tablet

### Why is it needed?

### How to test it?

- Change the link name too a long one directly from your browser inspector (subnav should have a 23.2rem max-width and you should be able to see ellipsis on the long link.
- Should not change on mobile.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/24606

### DIFF
--- a/packages/core/admin/admin/src/components/SubNav.tsx
+++ b/packages/core/admin/admin/src/components/SubNav.tsx
@@ -101,7 +101,7 @@ const Link = (
           <Typography
             tag="div"
             lineHeight="32px"
-            width="100%"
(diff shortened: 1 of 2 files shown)


Adzouz (commented):
Thanks a lot @HichamELBSI, I tested and on tablet and desktop it's good 🙏
But on mobile with a very long text in the list it's overflowing the page wrapper generating a horizontal scrollbar. Do you think we also need to add ellipsis and a max-width to prevent this from happening?

Issue #21369, opened 2024-09-18
Cache shows former status when coming back to a single type entry after release published (i18n)


yanniskadiri:
When reproducing https://github.com/strapi/strapi/issues/21177 I was able to confirm it’s fixed but I could see the former status when coming back to the entry, it shown new status on refresh.

Steps:

- create multiple single entries for different locales
- create a release and include them
- trigger the release
- go back to the single type entry
- the new status is not shown straight away, you need to refresh the page to see it.

PR #24618, opened 2025-10-17
fix: invalidate all documents cache after a release publish

Adzouz:
### What does it do?

The status of a document on the edit view page wasn't correctly updated after publishing a release.
We were clearing the list query cache but not for all single documents requests.

### Why is it needed?

The status not updated in the edit view page could be very confusing for a content editor that just published a release. This is important to have the correct status showed on the page.

### How to test it?

- Create a release
- Go to the content manager and create a new entry
- Click save to create a draft
- Add the entry to the release and select "Publish" (to change the status from Draft to Publish after release)
- Go in releases and publish the release
- Click on the content manager again and see the status updated on the entry in the list
- Click on the row to display the edit view page
-> The status should be also updated and now be "Published" instead of "Draft"

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/21369

🚀

### DIFF
--- a/packages/core/content-releases/admin/src/services/release.ts
+++ b/packages/core/content-releases/admin/src/services/release.ts
@@ -366,8 +366,8 @@ const releaseApi = adminApi
               method: 'POST',
             };
           },
-          invalidatesTags: () => [
-            { type: 'Release', id: 'LIST' },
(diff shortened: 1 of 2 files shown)


HichamELBSI (approved):
It looks fine, good job

PR #24619, opened 2025-10-17
enhancement: make the publishedAt attribute visible

boazpoolman:
### What does it do?

Set's `visible: true` for the publishedAt attribute.

### Why is it needed?

To allow users to sort their content manager view based on the publishedAt date.

Scherm­afbeelding 2025-10-17 om 14 38 37

### How to test it?

Pull this PR and see if you are able to sort the content manager view based on the publishedAt date.

### DIFF
--- a/packages/core/utils/src/content-types.ts
+++ b/packages/core/utils/src/content-types.ts
@@ -99,7 +99,13 @@ const getNonVisibleAttributes = (model: Model) => {
     [] as string[]
   );
 
-  return _.uniq([ID_ATTRIBUTE, DOC_ID_ATTRIBUTE, PUBLISHED_AT_ATTRIBUTE, ...getTimestamps(model), ...nonVisibleAttributes]);
+  return _.uniq([
+    ID_ATTRIBUTE,
+    DOC_ID_ATTRIBUTE,
+    PUBLISHED_AT_ATTRIBUTE,
+    ...getTimestamps(model),
+    ...nonVisibleAttributes,
+  ]);
 };
 
 const getVisibleAttributes = (model: Model) => {


innerdvations:
Thanks! Can you fix the linting on this? `yarn prettier:write`

boazpoolman:
@innerdvations done!

jhoward1994:
Hey @boazpoolman 

During some unrelated QA I think we found some issues with this PR, could you verify?

It seems to me the problem is the changes to `getNonVisibleAttributes`
```
const getNonVisibleAttributes = (model: Model) => {
...
    PUBLISHED_AT_ATTRIBUTE,
...
etc
```
This seems to contradict the changes made to `packages/core/core/src/domain/content-type/index.ts`? and in testing I wasn't able to sort the CM lists by publishedAt as described in the description.

So my suggestion would be to revert the change made to `getNonVisibleAttributes`. I tried this out and it seems to work, the only complication being I couldn't see the property in “default sort attribute” as in the PR desc

Wondering what you think and if you have any more context on this ?

boazpoolman:
Hi @jhoward1994,

It's been a little while, so please bear with me.
When I was working on the feature I remember the `publishedAt` attribute popping up as an option in the 'configure the edit view', which isn't desirable. So adding the attribute to the `getNonVisibleAttributes` seemed to solve that.

Just testing the feature now, I can see it doesn't work as expected. Not sure what changed there, or I might have overlooked something. Please feel free to make the required changes.

Bassel17:
I think `getNonVisibleAttributes` may have been refactored to be used everywhere, it's more of a regression

Issue #23124, opened 2025-03-12
"Available in" Column does not behave as expected


kristerlooga-bolt:
### Node Version

20.10.0 

### NPM/Yarn/PNPM Version

pnpm 9.12.2

### Strapi Version

5.10.2

### Operating System

Linux (Other)

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

The "Available In" column styling breaks by decreasing the usability of the table when there are more than a few locales. 

- The table becomes too wide and readability is difficult
- The dropdown positioning breaks when clicking on the available in column after scrolling right

In addition when the locale is clicked a dropdown appears, but when an option is clicked it does not take to that locale version of the entry.

- It always takes to en locale

### Steps to Reproduce

1. Create a localised content entry with a lot of locales
2. Look at list view

### Expected Behavior

- Ideally available in should be truncated after 3 locales
- When available in column value is clicked, then in the opened dropdown a click on element should take to that locales version.

### Logs

```shell

```

### Media

Giant list that scrolls on for a while
![Image](https://github.com/user-attachments/assets/dea654fa-b0f9-4d94-9204-f3a7ec209ada)

Dropdown position can vary after scroll, but it is often on the left or top edge

![Image](https://github.com/user-attachments/assets/11255414-119f-4680-8097-a614b4cd1b1c)

### Additional information

There has been a similar issue filed in the past, but it seems it was just marked as done and lost: https://github.com/strapi/strapi/issues/20520. Resurfacing again as the current behaviour still seems broken.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #24620, opened 2025-10-17
fix: behaviour of i18n "Available In" column

mathildeleg:
### What does it do?
* use Menu instead of Popover to fix vertical scroll when locales' list is long;
* fix overflow of column when document has more than 2 locales; 
* add link to locale's entry instead of last locale's entry created

BEFORE
https://github.com/user-attachments/assets/8e5153ee-534a-4294-ad13-d22c46163e74
Capture d’écran 2025-10-17 à 15 46 38
Capture d’écran 2025-10-17 à 15 46 44

AFTER
https://github.com/user-attachments/assets/f923209f-a083-467e-9548-5cb7be988e38
Capture d’écran 2025-10-17 à 15 45 31
Capture d’écran 2025-10-17 à 15 45 43

### Why is it needed?
The "Available In" column was not optimized, and having too many locales to an entry made the UI not functional. In addition, clicking on a locale didn't redirect to that locale's entry, which was confusing.

### How to test it?
* Add 4 locales in the Internationalization settings page, in addition to the default one.
* Go to any Collection or Single type that has Internationalization enabled.
* Create entries for each locale created (so 5 total for the same documentId).
* Go to the List view of that Collection/Single type and check the behaviour of the "Available In" column:
a. The "Available In" column isn't overflowing width-wise.
b. Clicking on it opens a scrollable list of locales.
c. Clicking on one of those locales redirects you to that locale's specific entry.

### Related issue(s)/PR(s)
Resolves https://github.com/strapi/strapi/issues/23124

### DIFF
--- a/packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
+++ b/packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
@@ -59,15 +59,13 @@ const LocaleListCell = ({
       return baseText;
     }
 
-    const moreText = formatMessage(
+    return formatMessage(
       {
         id: getTranslation('CMListView.popover.display-locales.more'),
-        defaultMessage: '+ {count} more',
+        defaultMessage: '{locales} + {count} more',
       },
-      { count: remainingCount }
+      { locales: baseText, count: remainingCount }
     );
-
-    return `${baseText} ${moreText}`;
   };
 
   const handleLocaleClick = (localeCode: string) => {
--- a/packages/plugins/i18n/admin/src/translations/en.json
+++ b/packages/plugins/i18n/admin/src/translations/en.json
@@ -20,7 +20,7 @@
   "CMEditViewBulkLocale.continue-confirmation": "Are you sure you want to continue?",
   "CMEditViewLocalePicker.locale.create": "Create <bold>{locale}</bold> locale",
   "CMListView.popover.display-locales.label": "Display translated locales",
-  "CMListView.popover.display-locales.more": "+ {count} more",
+  "CMListView.popover.display-locales.more": "{locales} + {count} more",
   "CheckboxConfirmation.Modal.body": "Do you want to disable it?",
   "CheckboxConfirmation.Modal.button-confirm": "Yes, disable",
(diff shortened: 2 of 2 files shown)


packages/plugins/i18n/admin/src/components/LocaleListCell.tsx
```diff
@@ -24,50 +35,78 @@ const LocaleListCell = ({ locale: currentLocale, localizations }: LocaleListCell
...
+
+    const moreText = formatMessage(
+      {
+        id: getTranslation('CMListView.popover.display-locales.more'),
+        defaultMessage: '+ {count} more',
+      },
+      { count: remainingCount }
+    );
+
+    return `${baseText} ${moreText}`;
```
HichamELBSI: I'm wondering if you should format the text this way. If a LTR language is selected, it means that you are forcing a RTL message here. You should probably use a formatMessage only 
```suggestion
    return formatMessage(
      {
        id: getTranslation('CMListView.popover.display-locales.more'),
        defaultMessage: '{locales} + {count} more',
      },
      { locales: baseText, count: remainingCount }
    );
```

This way, you won't have to force any wording.

packages/plugins/i18n/admin/src/translations/en.json
```diff
@@ -20,6 +20,7 @@
   "CMEditViewBulkLocale.continue-confirmation": "Are you sure you want to continue?",
   "CMEditViewLocalePicker.locale.create": "Create <bold>{locale}</bold> locale",
   "CMListView.popover.display-locales.label": "Display translated locales",
+  "CMListView.popover.display-locales.more": "+ {count} more",
```
HichamELBSI:
```suggestion
  "CMListView.popover.display-locales.more": "{locales} + {count} more",
```

HichamELBSI (changes requested):
Well done overall. I only have a single feedback

packages/plugins/i18n/admin/src/translations/en.json
mathildeleg: Thanks that's a much better approach! Here's the fix https://github.com/strapi/strapi/pull/24620/commits/3f351ad538ea7fe7f02577d6cd8252217c4810af

Issue #24200, opened 2025-08-19
Media library doesn’t display assets when re-assigning image after deletion in Strapi 5


airikej:
### Node Version

v20.17.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.19.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

When assigning an image to a custom content type (e.g., a product image) in Strapi 5, the media library behaves unexpectedly after replacing the image.

### Steps to Reproduce

1. Open a custom type with a media field (e.g., Product → Product Image).
3. Assign an image by navigating to the correct folder and selecting it.
4. Remove the assigned image.
5. Open the media library again to pick a new one.

This issue expects that you have nested media folders e.g Images -> Products -> Bags -> bag image list

### Expected Behavior

**Expected behavior:**
The media library should open in the same folder as before and display the assets.

**Actual behavior:**
The media library remembers the folder location in the breadcrumbs, but the asset list is empty. You have to navigate back one breadcrumb and then re-enter the folder for the assets to appear again.

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/7bc7d980-e8bc-4720-aad8-1963ae263d23

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Anukul16:
Hey @airikej is this issue solved ? If not i can help by pick it up quickly.

airikej:
> Hey [@airikej](https://github.com/airikej) is this issue solved ? If not i can help by pick it up quickly.

Hi! Nope, the issue is still there and would appreciate immensely if you could check on it 🙏

Anukul16:
Hey @airikej i dont find that actual product option where you have found the issue.

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/dcd0ce9f-8a47-4336-af57-b56fcd0ae63b" />

airikej:
> Hey [@airikej](https://github.com/airikej) i dont find that actual product option where you have found the issue.
> 
[...]

PR #24626, opened 2025-10-18
fix for Issue 24200: Media library doesn’t display assets when re-assigning image after deletion in Strapi 5

saiabhinav75:
### What does it do?

Fix for issue 24200, https://github.com/strapi/strapi/issues/24200

When reopening the asset dialog, the `folderPath` query parameter was being set to `undefined`, causing no assets to be displayed.  
This fix ensures that the current folder data is fetched and the correct `folderPath` is restored.

### Why is it needed?

Without this fix, reopening the asset dialog shows an empty asset list, which can confuse users and make them think their media files are missing.

### How to test it?

1. Open any collection type that includes a media field.  
2. Add an asset from a specific folder.  
3. Delete the added asset.  
4. Click **Add** again — the assets from the previously selected folder should now appear correctly.

### Related issue(s)/PR(s)

Fix for Issue https://github.com/strapi/strapi/issues/24200


innerdvations (changes requested):
Thanks for digging into [#24200](https://github.com/strapi/strapi/issues/24200) — the diagnosis looks right: after selecting from a nested folder, reopen keeps `folderId` but `folderPath` is missing, so `useAssets` falls back to `/` while the breadcrumbs (driven by folder id) still look correct.

Could you adjust the implementation before we merge? A few concrete issues with the current approach:

1. Assigning `queryObject!.folderPath = …` mutates React state during render. `useAssets` also runs **before** that line, so it keeps seeing the previous path. A correct path only shows up if something else re-renders afterward (today that mostly happens because `useFolders` picks up the mutated object). If folders aren’t fetched (e.g. mime filter / page ≠ 1), assets can stay on `/`.
2. `currentFolder?.path ?? "/"` overwrites a valid `folderPath` with `/` while the folder query is still loading.
3. There isn’t a regression test yet for this reopen path.

A safer shape would be: derive the query immutably (or call `onChangeFolder(folderId, path)` via `useEffect`), and skip the assets query until the path is known when `folderId` is set — no in-render mutation, and no `/` fallback when a folder id is present. Extending `AssetDialog.test.tsx` to assert `useAssets` receives the resolved `folderPath` would lock this in.

Also worth running Prettier on the touched line (single quotes + semicolon).

Happy to re-review once that’s in — the direction is good; it’s mainly the wiring that needs tightening.

_AI-assisted comment._

derrickmehaffy:
Hey there, since it's been about 30 days since we requested changes and there has been no movement, going to go ahead and mark this as closed but please feel free to open a new pr if you wish to continue.

PR #24628, opened 2025-10-18
fix: allow saving invisible fields in the content-manager

boazpoolman:
### What does it do?

It allows fields which are marked as `visible: false` to be saved through the content manager.

### Why is it needed?

One might want to mark a field as invisible and afterwards handle the editing of that field in a custom widget in the content-manager sidebar. This PR fixes this use case.

### How to test it?

### Related issue(s)/PR(s)

This was reported as an issue to a community plugin https://github.com/pluginpal/strapi-webtools/issues/334


Bassel17:
I would say that's more of a feature request/change, the purpose of not submitting fields is so users don't submit changes by mistake, my making a field visible and editing it then making it invisible again, and we also purposely don't clear those changes in memory because we want people to be able to see the changes they did if they toggle it back again to visible

Bassel17:
Will close this, in favor of the solution here #24849

Issue #24603, opened 2025-10-16
Markdown editor: <ol> is created with wrong numbers


JanMikes:
### Node Version

v20.19.5

### NPM/Yarn/PNPM Version

NPM 10.8.2

### Strapi Version

v5.28.0 (previous versions affected too)

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

Creating ordered list via markdown editor on lines that are not first in the editor, will cause items not start from 1 making user need to edit manually every line.

### Steps to Reproduce

- Have content type text editor (the one with markdown)
- Create few lines of text and have some paragraph before it
- Select the text (not at the start of editor) and click to create ordered list

### Expected Behavior

New ordered list should always start with 1

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/c39a9ae5-c187-42d6-aa2f-8165bd4b0403

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

akshat-chd:
@oiorain please assign this issue to me .

Adzouz:
Hi there!
I'm sorry for the delay but I'm glad to tell you that the fix for this issue was merged (thank you @dagadevelop 🙏) and it will be part of the next release of Strapi (Wednesday March 4th) 🚀

PR #24631, opened 2025-10-19
fix: Markdown editor: <ol> is created with wrong numbers

dagadevelop:
### Issue

When selecting multiple lines to create an ordered list in the Markdown editor, the numbering was based on the absolute line index in the document rather than starting from 1.

**Example:**

```
a
b

c  <- line 3
d  <- line 4
e  <- line 5
```

Selecting lines c, d, e and applying ordered list would produce:

```
4. c
5. d
6. e
```

Instead of the expected:

```
1. c
2. d
3. e
```

### Solution

Changed the line number calculation in `listHandler` function to use relative positioning from the first selected line:

**Before:**

```typescript
const lineInsertion = listType === 'BulletList' ? '- ' : `${i + 1}. `;
```

**After:**

```typescript
const lineInsertion = listType === 'BulletList' ? '- ' : `${i - pos[0] + 1}. `;
```

Where:

- `i` is the current line index
- `pos[0]` is the first selected line index
- `i - pos[0] + 1` calculates the relative position starting from 1

### Files Changed

- `packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/utils.ts`

### Related

Fixes [#24603](https://github.com/strapi/strapi/issues/24603)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/utils.ts
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/utils.ts
@@ -152,17 +152,19 @@ export const listHandler = (
 
     editor.current.operation(function () {
       selections.forEach(function (selection) {
-        const pos = [selection.head.line, selection.anchor.line].sort();
+        const pos = [selection.head.line, selection.anchor.line].sort((a, b) => a - b);
+        // Will match a bullet point or any number
+        const regexIsItemFromList = /^[*-]\s|^\d+\.\s/;
 
-        // Remove if the first text starts with it
+        // Remove if the first text starts with a bullet point or any number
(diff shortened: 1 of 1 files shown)


Adzouz:
Hi @dagadevelop, thanks a lot for the contribution and your patience 🙏 So sorry it took so long to review it 😅
I tested it and there was a tiny thing broken (see my first video) so I took the liberty to take over the PR and add a tiny addition in order to make it work properly and move forward with it (I hope you don't mind)

Before:

https://github.com/user-attachments/assets/c8fae619-dc57-457a-851b-1a393a577b23


After the fix:

https://github.com/user-attachments/assets/340dea23-4f57-468a-8184-27ffb6cd50b0

dagadevelop:
@Adzouz  No problem, the thing is that it works, thank you so much!!!

Issue #24084, opened 2025-08-04
Unable to click on the last row in Media Library modal


radoslavzeman:
### Node Version

20.19.0

### NPM/Yarn/PNPM Version

NPM 10.8.1

### Strapi Version

5.18.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

In Media Library modal opened from a media field, the bottom part of the modal content is not clickable.
You have to scroll up a bit, to be able to click on last visible items.

### Steps to Reproduce

1. Create a content type with a media field and add some medias and folders to Media Library.
3. Create new entry in the content type and click on the media field, a modal will open.
4. Try to click on some of the last visible items ("edit" icon should show when the item is really hovered and clickable).

### Expected Behavior

The whole modal content should be normally clickable.

### Logs

```shell

```

### Media

<img width="1200" height="820" alt="Image" src="https://github.com/user-attachments/assets/ca28e90b-5eab-4688-8aef-92c04b56577a" />

https://github.com/user-attachments/assets/12a844c2-0f59-4ff3-b2b5-168b36cc7a47

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Adzouz:
Hi @radoslavzeman and thanks a lot for reporting this issue!
I tried to reproduce on my side but I couldn't... Is it still happening in the latest version of Strapi?
Can you tell me which browser you are using and your resolution please?\
Thanks a lot!

https://github.com/user-attachments/assets/8f938494-48c0-42d0-92ee-153a7c64ca04

radoslavzeman:
Hi @Adzouz!
Thanks for paying attention to this issue :) 

After updating to the latest Strapi version 5.20.0, the issue still remains.

I use Chrome Version 138.0.7204.184, but it also appears in Safari and Firefox.

My screen resolution is  3024 × 1964 on MacBook Pro M1 2021 14-inch

oiorain:
👋 Hello, and apologies for the confusion!

As part of an automated issue cleanup, this issue was mistakenly closed. That action was unintentional — we're in the process of reviewing and reopening affected issues like this one.

🙏Thank you for your patience, and sorry again for any disruption.

Adzouz:
[...]

PR #24633, opened 2025-10-19
fix: Unable to click on the last row in Media Library modal

dagadevelop:
### Issue

In the Media Library dialog, the pagination footer at the bottom was blocking hover interactions on assets positioned near it. Users couldn't hover over or interact with the last row of assets in the grid view.

### Root Cause

The pagination container had `position="relative"` and `zIndex={1}` (required for the PageSize dropdown to display correctly), which created a stacking context that was blocking pointer events to elements visually positioned behind it.

### Solution

Added the `style` property of the pagination `Flex` container from:

```typescript
style={{ overflow: 'hidden' }}
```

**Location:**
`packages/core/upload/admin/src/components/AssetDialog/BrowseStep/BrowseStep.tsx` (line 462)

### Testing

1. Open Media Library dialog
2. Navigate to a folder with many assets (enough to show pagination)
3. Hover over assets in the last row near the pagination footer
4. **Expected:** Hover effects work correctly on all assets

### Files Changed

- `packages/core/upload/admin/src/components/AssetDialog/BrowseStep/BrowseStep.tsx`

Fix [#24084](https://github.com/strapi/strapi/issues/24084)

### DIFF
--- a/packages/core/upload/admin/src/components/AssetDialog/BrowseStep/BrowseStep.tsx
+++ b/packages/core/upload/admin/src/components/AssetDialog/BrowseStep/BrowseStep.tsx
@@ -459,7 +459,7 @@ export const BrowseStep = ({
           paddingTop={4}
           position="relative"
           zIndex={1}
-          style={{ overflow: 'hidden' }}
+          overflow="hidden"
         >
           <PageSize
             pageSize={queryObject.pageSize! as number}


packages/core/upload/admin/src/components/AssetDialog/BrowseStep/BrowseStep.tsx
```diff
@@ -454,7 +454,13 @@ export const BrowseStep = ({
       )}
 
       {pagination.pageCount > 0 && (
-        <Flex justifyContent="space-between" paddingTop={4} position="relative" zIndex={1}>
+        <Flex
+          justifyContent="space-between"
+          paddingTop={4}
+          position="relative"
+          zIndex={1}
+          style={{ overflow: 'hidden' }}
```
Adzouz: You can use the overflow property directly instead of using style 😊 `overflow="hidden"`

dagadevelop: Perfect, no problem

dagadevelop: @Adzouz fixed

Adzouz:
Perfect, thank you very much 🙌 I just checked but we'll also have a commit message format issue. If you could change them please, we're using https://www.conventionalcommits.org/en/v1.0.0/ format 😊

PR #24636, opened 2025-10-20
fix: input types

HichamELBSI:
### What does it do?

Many input types were undefined or poorly defined (with types that did not exist, such as type="string" or type="uid"), mainly due to props object drilling/spreading from data to TextInput DS component.

### Why is it needed?

Needed for:
- Support native behavior on specific input type (email, url, etc)
- Native validation on form
- Some specific CSS where not (will not be) applied when using input[type="text"] selector

### How to test it?

Run the admin, make sure the input has the correct type.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/components/FormInputs/String.tsx
+++ b/packages/core/admin/admin/src/components/FormInputs/String.tsx
@@ -20,6 +20,7 @@ const StringInput = forwardRef<HTMLInputElement, InputProps>(
           onChange={field.onChange}
           value={field.value ?? ''}
           {...props}
+          type="text"
         />
         <Field.Hint />
         <Field.Error />
--- a/packages/core/admin/admin/src/pages/Auth/components/Login.tsx
+++ b/packages/core/admin/admin/src/pages/Auth/components/Login.tsx
@@ -122,7 +122,7 @@ const Login = ({ children }: LoginProps) => {
                     defaultMessage: 'kai@doe.com',
                   }),
                   required: true,
-                  type: 'string' as const,
+                  type: 'email' as const,
                 },
                 {
                   label: formatMessage({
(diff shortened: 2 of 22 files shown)


packages/core/admin/admin/src/components/FormInputs/String.tsx
```diff
@@ -20,6 +20,7 @@ const StringInput = forwardRef<HTMLInputElement, InputProps>(
           onChange={field.onChange}
           value={field.value ?? ''}
           {...props}
+          type="text"
```
HichamELBSI, line 23: This is the TextInput that is rendered in most of the dynamic forms (including the CM edit view). Currently, all the text (string) type input are rendered as input[type="string"] because of the `{...props}` .

markkaylor: Thanks Hicham, I'm just wondering if type text should be defined on the DS component so we don't have to add it each time the component is used 🤔

Adzouz: Yes maybe it could be the default value if not provided, wdyt?

Adzouz (approved):
Looks good to me, good catch chief 👏

packages/core/admin/admin/src/components/FormInputs/String.tsx
HichamELBSI: The thing is that even with a default type it won't work (at least in this case because the props object does contain a type="string" prop which will override the default one anyway 🤦🏽 . It will only work for a small part of the text inputs of the cms but indeed we should definitely have a default type. I suspect that the default type has been removed in the DS textinput, I don't know why

PR #24637, opened 2025-10-20
future(i18n): add background job notification system

markkaylor:
### What does it do?

- Create an AILocalizationJobs model, service, controller, and route
- Polls the database to check for updates as the ai server settles its response
- Refactors the ai trigger to use documents.use middleware

### Why is it needed?

- To notify the user of the status of the ai translation background job

### How to test it?

Go to a content-type in the default locale 
Add a bunch of content and save
You should see a notification for the document save action immediately
Some time later you should see another notification for the translation

test this for single types, collection types, save and publish actions

### Related issue(s)/PR(s)

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -226,20 +226,23 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
           return result;
         }
 
+        // Check if AI localizations are enabled before triggering
+        const isEnabled = await this.isEnabled();
+        if (!isEnabled) {
+          return result;
+        }
+
         // Don't await since localizations should be done in the background without blocking the request
-        // Use setImmediate to ensure this runs outside the current transaction context
-        setImmediate(() => {
-          strapi
-            .plugin('i18n')
-            .service('ai-localizations')
-            .generateDocumentLocalizations({
-              model: context.contentType.uid,
-              document: result,
-            })
-            .catch((error: any) => {
-              strapi.log.error('AI Localizations generation failed', error);
-            });
-        });
+        strapi
+          .plugin('i18n')
+          .service('ai-localizations')
+          .generateDocumentLocalizations({
+            model: context.contentType.uid,
+            document: result,
+          })
+          .catch((error: any) => {
+            strapi.log.error('AI Localizations generation failed', error);
+          });
 
         return result;
       });


tests/e2e/tests/i18n/editview.spec.ts
markkaylor, line 1: surprised to see this in the diff but whatever

packages/plugins/i18n/server/src/models/ai-localization-job.ts
markkaylor, line 1: model should have a singular name

packages/plugins/i18n/server/src/models/ai-localization-job.ts
```diff
@@ -0,0 +1,45 @@
...
+      column: { notNullable: true },
+    },
+    createdAt: {
+      type: 'datetime',
+      default: () => new Date(),
+    },
+    updatedAt: {
+      type: 'datetime',
+      default: () => new Date(),
+    },
```
remidej, line 41: I guess they're not created by default because it's not a content type?

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -170,6 +196,28 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
       }
     },
+    setupMiddleware() {
+      strapi.documents.use(async (context, next) => {
+        const result = await next();
+
+        // Only trigger on create/update actions
+        if (!['create', 'update'].includes(context.action)) {
+          return result;
+        }
```
remidej, line 227: We should handle clone action too

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -170,6 +196,28 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+
+        try {
+          // Don't await since localizations should be done in the background without blocking the request
+          strapi.plugin('i18n').service('ai-localizations').generateDocumentLocalizations({
+            model: context.contentType.uid,
+            document: result,
+          });
+        } catch (error) {
+          strapi.log.error('AI Localizations generation failed', error);
+        }
```
remidej: I just found out try/catch doesn't catch when the promise isn't awaited. So we should use the `.catch()` syntax instead

remidej (commented):
This is beautiful, thanks for putting it together @markkaylor @araksyagevorgyan

packages/plugins/i18n/server/src/models/ai-localization-job.ts
markkaylor: Correct, at least that's what I understood after looking at history-version.

packages/plugins/i18n/server/src/services/ai-localizations.ts
markkaylor: I have a default locale english and also a french locale. I create an entry in english and click save. I'm noticing it doesn't create and translate the french entry. It will only generate the translation if the target locale entry already exists. Is that a known issue, regression here, or the expected behavior?

markkaylor: Never mind this is specific to cloning

remidej: It's not the expected behavior, it should create the derived locales of the document if they don't exist. I think we tested that on Friday with Araksya and it worked, but I could be confused. Also the translations are currently very slow to run, around 10 seconds on my side, just in case it could be that

remidej: ah makes sense. If cloning is annoying we can definitely leave it aside, I just thought about it because I looked up the history code

araksyagevorgyan: Yes, we tested for create/update flow, derived locales should be created/updated when default one is created/updated. The issue is with nested fields - I didn't succeed yet to get them correctly for translatableContent.

markkaylor: I think I'll leave clone for another PR 🙃

packages/plugins/i18n/server/src/services/ai-localization-jobs.ts
```diff
@@ -0,0 +1,85 @@
...
+    status = 'processing',
+  }: {
+    documentId: string;
+    contentType: string;
+    sourceLocale: string;
+    targetLocales: string[];
+    status?: AILocalizationJobs['status'];
+  }) {
+    // Check if job already exists for this document
+    const existingJob = await strapi.db.query(AI_LOCALIZATION_JOB_UID).findOne({
```
markkaylor: Use getJobByDocumentId

packages/plugins/i18n/server/src/services/ai-localization-jobs.ts
```diff
@@ -64,11 +64,22 @@ export const createAILocalizationJobsService = ({ strapi }: { strapi: Core.Strap
   /**
    * Get job by document ID
    */
-  async getJobByDocument(documentId: string) {
-    return strapi.db.query(AI_LOCALIZATION_JOBS_UID).findOne({
+  async getJobByDocumentId(documentId: string) {
```
remidej: I think this should take model as a param? the document ID could collide in theory since each content type is a different db table

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -193,27 +201,36 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
     },
     setupMiddleware() {
       strapi.documents.use(async (context, next) => {
         const result = await next();
 
-        // Only trigger on create/update actions
+        // Only trigger on create/update/publish actions
         if (!['create', 'update'].includes(context.action)) {
           return result;
         }
```
remidej: the comment adds publish but not the code. I think publish may be mentioned in the PRD but I don't think there's anything to actually translate upon publish

markkaylor: Yeah I had added publish action but turns out every time you publish it's just an update action 🤷‍♂️

packages/plugins/i18n/server/src/services/ai-localization-jobs.ts
markkaylor: ok sounds good I'll update

markkaylor:
Looking into the API tests that were failing

packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
remidej, line 1: On a single type: it works, I'm seeing the notification properly. One odd thing though is that it starts polling from the initial page load. I thought the solution you found allowed us to only poll upon save.

On a collection type: even though the translation does work in the background, I'm never seeing the notification. I'm getting a bunch of 404s like this:
> GET /i18n/ai-localization-jobs/collection-types/api::article.article/ini1ujy4s72m0a3g0zmt5rt9 (4 ms) 404

You may need to check on your side if it works on a new document that doesn't have a job yet 🤔

markkaylor: Weird I tested both before pushing and it was working, I'll try again. I had both collection type and single type working yesterday with the polling starting and stopping as desired. Not sure what's up.

markkaylor: I'm not seeing immediate polling on single types for existing job and the notification appears for both ST and CT

https://github.com/user-attachments/assets/07f4c8f9-22eb-4e8b-8696-ef818486c416

You are right when it's a new job the polling starts immediately for single types. I'll look into that. I'm always seeing the notification though for both CT and ST


https://github.com/user-attachments/assets/c5e94150-e554-4f0b-a289-473781579a83

markkaylor: Also confirm the immediate polling issue for new jobs happens for both CT and ST

markkaylor: I pushed a fix, can you test again?


https://github.com/user-attachments/assets/3acb8093-8d0a-4fdd-81a6-e680b5d24b99

I did get rate limited or something once and didn't get the error notification but the status icon changed. Not sure what's up with that it should be handled here I think

https://github.com/strapi/strapi/blob/402f464687df6ea741b76884f740886db06c72b8/packages/plugins/i18n/server/src/services/ai-localizations.ts#L108-L124

markkaylor:
I think we should only create the model in the DB if ai available and enabled

packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
remidej: https://github.com/user-attachments/assets/5f49aad1-f220-420a-9658-28d83f4b0ab3

Looks like I'm still not getting the job result in the frontend

markkaylor: Could you do a yarn setup and try again, yesterday I had to do a yarn setup every time I made changes to invalidate / provides tags

remidej: i think you're right the setup fixes it, weird I had ran build every time though

remidej (dismissed):

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -43,6 +49,11 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
       model: UID.ContentType;
       document: Modules.Documents.AnyDocument;
     }) {
+      const isFeatureEnabled = await this.isEnabled();
+      if (!isFeatureEnabled) {
+        return;
+      }
```
markkaylor, line 55: I think we should abandon ship as early as possible if it's disabled

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -28,6 +28,12 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
         return false;
       }
 
+      const settings = getService('settings');
+      const aiSettings = await settings.getSettings();
+      if (!aiSettings?.aiLocalizations) {
+        return false;
+      }
```
markkaylor, line 35: This check was missing right?

markkaylor:
@remidej I [pushed again](https://github.com/strapi/strapi/pull/24637/commits/6f0d0e1eff78e795ad55bfe742c9b0a4f4c97510) to omit AI related apis if the app does not have AI or it is disabled at the app level. LMK what you think. ~I'm hoping it resolves the api test failure issue for now as well. It's failing because api test db operations are run in transactions and we are not awaiting our generate method on the ai localization service~ it doesn't 🙃

packages/plugins/i18n/server/src/services/ai-localizations.ts
remidej: huh I thought that had been added, good catch

packages/plugins/i18n/server/src/index.ts
remidej, line 1: I'm fine with it if it fixes the CI issues, but otherwise I'm not very keen on this. The logic to check if this is enabled is getting copied in different places, and we'll need to check if each service exists before calling it. I think I prefer knowing the plugin has a consistent shape. I believe it's fine to have the AI code running for all users as long as it keeps returning early

markkaylor: I'll remove it then, but I don't think users who explicitly do not want AI will be happy to see API's referencing AI and tables in the db referencing AI, maybe there is a cleaner way to do it but we can handle it later

markkaylor: It does not resolve the test issue, looking into that still

markkaylor:
The tests pass consistently locally now so not really sure what to do at this point...

remidej (approved):

PR #24641, opened 2025-10-21
chore: Correctly  type `env-helper.ts`

johachi:
### What does it do?

- Adds type overload to env-helper functions so that they are typed correctly according to the arguments passed to it.
- Update examples to be typed (`.ts` files only)
  **This edit caused false positive regarding code duplication**
- Updated template files to be typed (`.ts` files only)

### Why is it needed?

Using strict typescript would give error on code below since typing said `foo` could be `string | undefined`.

```ts
function printFoo(value: string) {
  console.log(value);
}

const foo = env("FOO", "bar");

// Before
printFoo(foo); //  Type 'undefined' is not assignable to type 'string'.(2345)

// After
printFoo(foo); // No error
```

Added same type hint for all the other `env.<method>` methods:
- int
- float
- bool
- json
- array
- date
- oneOf

### How to test it?

Hovered the functions to see type definition.

hover type definition

### Related issue(s)/PR(s)

N/A


examples/empty/config/admin.ts
```diff
@@ -1,4 +1,6 @@
-const adminConfig = ({ env }) => ({
+import { type Env } from '@strapi/utils';
+
+const adminConfig = ({ env }: { env: Env }) => ({
```
unrevised6419, line 3: Suggestion: Typings of the configs can be reverted, as they will be covered by this PR https://github.com/strapi/strapi/pull/21859 started over an year ago.

unrevised6419:
@Bassel17 This can be closed in favor of https://github.com/strapi/strapi/pull/24881 and  https://github.com/strapi/strapi/pull/21859

Bassel17:
Thanks for the contribution!
but some of the work here has either been done or will be merged soon from other PRs

PR #24643, opened 2025-10-21
fix: ios input zoom on focus

HichamELBSI:
### What does it do?

Override the text input font size and line height only on iOS (from 14px to 16px).

### Why is it needed?

When input font size is below 16px, a zoom is automatically done on focus on iOS. To avoid that annoyinh zoom, we are increasing the font-size from 14 to 16. 

⚠️ This is considered as a temporary fix, product designer will work on a permanent solution to avoid breaking the actual 14/20 ratio.

### How to test it?

- Open the admin with an iOS device and focus form inputs.
- There shouldn't be any zoom should 

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/components/Theme.tsx
+++ b/packages/core/admin/admin/src/components/Theme.tsx
@@ -21,6 +21,11 @@ const Theme = ({ children, themes }: ThemeProps) => {
   const [systemTheme, setSystemTheme] = React.useState<'light' | 'dark'>();
   const { locale } = useIntl();
   const dispatch = useDispatch();
+  const isIos =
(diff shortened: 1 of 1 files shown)


mathildeleg:
I tested on my iPhone (iOS 26.0.1) but I still get the Zoom effect on inputs... 😔

HichamELBSI:
> I tested on my iPhone (iOS 26.0.1) but I still get the Zoom effect on inputs... 😔

Thanks for testing @mathildeleg, yes because we had to wait for the input type text PR to be merged, it should be good now that it has been merged. You can pull the branch 😄

mathildeleg (approved):
Perfect it works now ✨

Issue #22431, opened 2024-12-13
[Homepage] Localized entries that are not in the by default language are not displayed in the widget


MarionLemaire:
### Required System information

EE

WSL

Local

### Describe the bug
Localized entries that are not in the by default language are not displayed in the widget, nor for editing (creating), nor for publishing

### Steps to reproduce the behavior

1. Create a collection type with i18n enabled
2. Create an entry for this CT in by default language, save, publish
3. Create an entry for this CT in another language, save, publish
4. Check Homepage widgets : only by default localized entries are displayed in the widgets

### Expected behavior

both localized versions should be displayed 

Enhancement for later : add the locale label in the widgets to differentiate localized entries

### Screenshots

https://github.com/user-attachments/assets/855534da-d850-4846-9274-523ebdde8ca3

oiorain:
This is still happening in 5.27

PR #24647, opened 2025-10-21
fix: include locales to get last modified entries in widgets

mathildeleg:
### What does it do?
Fetch last modified/published's entries' for all locales.

### Why is it needed?
If a user has only modified the French entry, this is the one reflected in the homepage widgets, not the default locale.
_BEFORE_
_Only getting the meta data of default locale's entry_
Capture d’écran 2025-10-21 à 17 45 08

_AFTER_
_If the French entry was modified, this is the meta data I'm now showing here_
Capture d’écran 2025-10-21 à 17 42 05

### How to test it?
* Create at least one more locale in the Internationalization settings page.
* In the Content Type Builder, create a Content Type with localization enabled.
* Go to the Content Manager for this Content Type and create an entry in the default locale.
* Create an entry in the additional locale.
* Go back to the homepage and see both entries in the Last modified widget.
_If you're testing on an already existing entry, and only creating or modifying another locale's entry, it may only display that locale's entry in the widget which makes sense, depending on the other entries you modified prior to your test_

### Related issue(s)/PR(s)
fixes https://github.com/strapi/strapi/issues/22431

### DIFF
--- a/tests/e2e/tests/content-manager/home-widgets.spec.ts
+++ b/tests/e2e/tests/content-manager/home-widgets.spec.ts
@@ -72,6 +72,48 @@ test.describe('Homepage - Content Manager Widgets', () => {
     await expect(mostRecentModifiedEntry.getByRole('gridcell', { name: 'Modified' })).toBeVisible();
   });
 
+  test('a user should see entries per locale in the last published entries widget', async ({
+    page,
+  }) => {
+    const recentlyPublishedWidget = page.getByLabel(/last published entries/i);
+    await expect(recentlyPublishedWidget).toBeVisible();
+
+    // Create and publish an English article
+    await navToHeader(page, ['Content Manager', 'Article'], 'Article');
+    await page.getByRole('link', { name: 'Create new entry' }).first().click();
+    await page.waitForURL(
+      /\/admin\/content-manager\/collection-types\/api::article\.article\/create/
+    );
+    const titleFieldEnglish = page.getByLabel(/title/i);
+    await titleFieldEnglish.fill('West Ham Football Team');
+    await page.getByRole('button', { name: /publish/i }).click();
+    await findAndClose(page, 'Published document');
+
+    // Create and publish a French entry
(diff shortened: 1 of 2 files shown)


Adzouz (commented):
Looks good thanks 🙏 but can add an e2e test to see it correctly reflected please (maybe you can use an existing one that is already testing the widget)? Also might need an API test, wdyt?

tests/e2e/tests/content-manager/home-widgets.spec.ts
```diff
@@ -72,6 +72,48 @@ test.describe('Homepage - Content Manager Widgets', () => {
...
+    await expect(recentlyPublishedWidget).toBeVisible();
+
+    // Create and publish an English article
+    await navToHeader(page, ['Content Manager', 'Article'], 'Article');
+    await page.getByRole('link', { name: 'Create new entry' }).first().click();
+    await page.waitForURL(
+      /\/admin\/content-manager\/collection-types\/api::article\.article\/create/
+    );
+    const titleFieldEnglish = page.getByLabel(/title/i);
+    await titleFieldEnglish.fill('West Ham post match analysis');
```
Adzouz: This article already exists by default, maybe choose another title?

mathildeleg: Fixed :) https://github.com/strapi/strapi/pull/24647/commits/7e1126fefa8a5925e5794d9f4d8257f4939f92ff

markkaylor (approved):
Tested and looks good to me! Any reason this one is flagged as don't merge? Are we waiting on something else?

mathildeleg:
> Tested and looks good to me! Any reason this one is flagged as don't merge? Are we waiting on something else?

Yes, waiting on this one to be merged https://github.com/strapi/strapi/pull/24648 so that the widgets aren't cluttered by all modified entries (since without this fix, updating one locale updates all of them, if the CT has at least one non-localized field)

Issue #22924, opened 2025-02-18
A change inside the source changes the status to modified. Publishing a translation will publish the source too


dhoko:
### Node Version

20.18.3

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.10.2 (80103017ad0134c4681c5b742cfd24091a839446)

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

N/A

### Bug Description

👋 

When we make a change inside the source (en), it changes its status to _modified_, same for the translations (_as a side effect_). 

Now if we load a translation, and we publish it, it's going to publish the source too (_with our changed content_).

### Steps to Reproduce

1. Create a page with a translation
2. Make a change ex: a new keyword
3. Save
4. Its status =  `modified` same for the translated document
5. Load the translated document and publish it
6. Both are now `published` the translations as it should but the source too (_with the changes_)

### Expected Behavior

Keep the source as modified, when I publish a translation I don't publish the modified source. We can ex: hotfix the translated page, yet the changes inside the source don't need to be published.

The issue is created via https://github.com/strapi/strapi/blob/develop/packages/plugins/i18n/server/src/bootstrap.ts#L54 idk if it's expected.

I tried to remove it, and it seems to work. When I make a change to the source, the translations are still `published`. So I can't publish by mistake the source. Even if I do it via the API.

### Logs

```shell

```

### Additional information

Maybe we need another status, idk. But publishing a translation should _not_ trigger a published source too. Our changes should still be as a draft.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

fsbtmo:
We have the same bug. 

But I think there is only an issue with what is displayed. So the "published" and "modified" status in the control panel are wrong and misleading, the actual content returned by the API does not change.

And the control panel thinks the latest draft is published, which means I am actually not able to "really" publish the draft. I first have to make some changes to enable the button again.

This is very annoying, makes it very hard to work and maintaining an overview for our content editors across our 9 countries.

markkaylor:
Can you share your schema? It sounds like you may have a non-localized attribute. Can you confirm all attributes have localized to true like:

```
  "pluginOptions": {
    "i18n": {
      "localized": true
    }
  },
[...]

PR #24648, opened 2025-10-21
fix: modifying&publishing locale should only update that locale

mathildeleg:
### What does it do?
* Only updates one locale if non-localized field(s) isn't(aren't) modified.
* Removes "publish" action from i18n middleware

### Why is it needed?
If a document has localization enabled but at least one non-localized field:
* modifying an entry in one locale shouldn't update all locale unless that/those non-localized field(s) were modified.
* publishing one locale shouldn't publish all locales

### How to test it?
* Create at least one more locale in the Internationalization settings page.
* In the Content Type Builder, create a Content Type with localization enabled, but with at least one field non-localized.
* Go to the Content Manager for this Content Type
* Create an entry in the default locale.
* Create an entry in the additional locale.
* Modify a localized field in the additional locale => only this entry should be updated in the db.
* Modify a non-localized field in the additional locale => both entries should be updated in the db.
* Publish the default locale's entry.
* Publish the additional locale's entry.
* Modify a localized field in the additional locale's entry.
* Publish the additional locale's entry => the default locale's entry shouldn't be updated nor published again.
* Modify a non-localized field in the additional locale's entry.
* Publish the additional locale's entry => the default locale's entry should also be updated but not published again (status should be **Modified**)

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/22924

### DIFF
--- a/packages/plugins/i18n/server/src/bootstrap.ts
+++ b/packages/plugins/i18n/server/src/bootstrap.ts
@@ -1,5 +1,5 @@
 import type { Schema } from '@strapi/types';
-import { isEqual } from 'lodash';
+import { isEqual } from 'lodash/fp';
 import { getService } from './utils';
 
 const registerModelsHooks = () => {


packages/plugins/i18n/server/src/bootstrap.ts
```diff
@@ -51,7 +60,19 @@ const registerModelsHooks = () => {
         .query(schema.uid)
         .findOne({ where: { id: resultID }, populate: attributesToPopulate });
 
-      await getService('localizations').syncNonLocalizedAttributes(populatedResult, schema);
+      const currentFields = copyNonLocalizedAttributes(schema, populatedResult);
+      const originalFields = copyNonLocalizedAttributes(schema, originalData);
```
markkaylor: I haven't really looked at what this does or what's going on here in the i18n plugin but just a heads up that I noticed at a lower level in the document service we also have a function `copyNonLocalizedFields` I wonder what's the difference between the two / how they work together:

https://github.com/strapi/strapi/blob/47fec47d39711511b8558211f9692e544ec67a36/packages/core/core/src/services/document-service/internationalization.ts#L95-L127

https://github.com/strapi/strapi/blob/47fec47d39711511b8558211f9692e544ec67a36/packages/core/core/src/services/document-service/repository.ts#L240-L257

mathildeleg: Thanks I'll take a look 🙏🏻

mathildeleg: If I understand correctly, `copyNonLocalizedFields` fetches data AND merges it to create a new locale, whereas `copyNonLocalizedAttributes` only filters those fields from a previously fetched entry.
But I'm not a 100% sure about my solution so let me know if you want to sync to talk about it!

packages/plugins/i18n/server/src/bootstrap.ts
```diff
@@ -26,10 +26,19 @@ const registerModelsHooks = () => {
...
     // Build a populate array for all non localized fields within the schema
-    const { getNestedPopulateOfNonLocalizedAttributes } = getService('content-types');
-
+    const { getNestedPopulateOfNonLocalizedAttributes, copyNonLocalizedAttributes } =
+      getService('content-types');
     const attributesToPopulate = getNestedPopulateOfNonLocalizedAttributes(schema.uid);
 
+    // Get original data before the update to compare what actually changed
+    const originalData =
+      context.action === 'update' && context.params.documentId
```
araksyagevorgyan: I'm curious: if we change a field and publish directly without updating it first, would it work? 🤔

araksyagevorgyan: I think I saw something weird but I am not able to reproduce it anymore, so I will resolve the conversation.

araksyagevorgyan: Ok, I was able to reproduce it: I published my default locale, then I published derived one. After that I changed non-localized field in that derived locale, clicked publish and my default locale's status changed to Modified, also the non-localized field has been updated on default locale.

mathildeleg: I'll check on my side tomorrow (I think it should work correctly but you're making me doubt haha) but: 
A non-localized field **should** update all locales, even default ones.
Only localized fields shouldn't update all locales, only the one locale intended.
So if you updated a non-localized field, it makes sense it updated the default locale, no?

araksyagevorgyan: I think my confusion came from the following step:
**Modify a field in the additional locale's entry.**
Should be good then 👍

mathildeleg: You're right, that wasn't very clear, sorry about that 🫣 I'll check on my side tomorrow and update the instructions 🙏🏻

mathildeleg: > I'm curious: if we change a field and publish directly without updating it first, would it work? 🤔

It works ✅ if we change a non-localized field, it updates the drafts versions of all locales + publish the locale you clicked "Publish" on, but doesn't publish the other locale, which is what we want! 🥳

araksyagevorgyan: No worries ! yes, I also tested and I confirm it works for me as well 👍

packages/plugins/i18n/server/src/bootstrap.ts
```diff
@@ -1,4 +1,5 @@
 import type { Schema } from '@strapi/types';
+import { isEqual } from 'lodash';
```
markkaylor: Could you use the export from `lodash/fp` I think we try to favor the functional programming lodash. Heads up the function signature could be different.

mathildeleg: Fixed here https://github.com/strapi/strapi/pull/24648/commits/0546374d3a5e62a1ee7960dbbacdc587f4b3b83f 
Didn't need to change anything other than the import :)

mathildeleg:
> Nice work! Tested and looks like it's working as expected now 👏
> 
> I think we should merge but noticing one edge case when AI translation is enabled:
> 
> Let's say both my locales are published. I update only localized fields in the default locale, if that locale contains a non-localized field then the additional locale will always be marked as modified after publishing the original locale even I didn't make changes to any non-localized fields.
> 
>  Screen.Recording.2025-11-12.at.09.41.47.mov

Isn't that the intended behaviour though, since updating the default locale will always trigger the AI translation for other locales? Because on my side if I modify a localized field in another locale (not the default one), the status of the Published default locale's entry stays the same; but if I do modify the non-localized field the status correctly returns Modified:


https://github.com/user-attachments/assets/7f68680b-f573-4862-89db-030c1023d769

markkaylor:
> > Nice work! Tested and looks like it's working as expected now 👏
> > I think we should merge but noticing one edge case when AI translation is enabled:
> > Let's say both my locales are published. I update only localized fields in the default locale, if that locale contains a non-localized field then the additional locale will always be marked as modified after publishing the original locale even I didn't make changes to any non-localized fields.
> > Screen.Recording.2025-11-12.at.09.41.47.mov
> 
> Isn't that the intended behaviour though, since updating the default locale will always trigger the AI translation for other locales? Because on my side if I modify a localized field in another locale (not the default one), the status of the Published default locale's entry stays the same; but if I do modify the non-localized field the status correctly returns Modified:
> 
>  Enregistrement.de.l.ecran.2025-11-12.a.10.18.45.mov

yes it is in fact the intended behavior 😆 my bad

PR #24652, opened 2025-10-22
fix: uncomment component cleanup script

Bassel17:
### What does it do?

- Uncomment component cleanup script, so it can run automatically in projects again

### Why is it needed?

- for projects that had this issue #23858  we need to cleanup the corrupted data

### How to test it?

- create a strapi app, fill it with relations different types of  relation data, between draft and publish contentTypes, and non draft and publish contentTypes, in components, DZ...
- then run the script and check if it removes anything it shouldn't remove

### Related issue(s)/PR(s)

#24316 #24227 #24467


innerdvations:
@laurenskling This will re-enable the script that you fixed

laurenskling:
Thanks

PR #24653, opened 2025-10-22
enhancement: make ai chat auto open

Bassel17:
### What does it do?

Adds some UX enhancements to the AI chat feature:
- auto open chat
- makes chat icon slightly larger

### How to test it?

- run the app with a growth license the chat should auto open every time you access the CTB


butcherZ (approved):
Tested locally, I can confirm this feature is indeed very annoying :D

butcherZ:
need to fix the failed e2e test though :P

Issue #23600, opened 2025-05-27
Issue creating a new locale for a content type with localization disabled for a field.


Lvdwardt:
### Node Version

22.16.0

### NPM/Yarn/PNPM Version

PNPM 10.0.0

### Strapi Version

5.13.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

Turning Off Enable localization for this field does not work as expected when creating a new locale. It's not copying the not-localised field contents to a newly created locale.

https://github.com/user-attachments/assets/9de35fc6-08be-418c-815d-04b03780cff9

### Steps to Reproduce

1. Create a collection type with Internationalization.
2. Add a "key" field, and turn OFF "Enable localization for this field".
3. Create a new entry in a locale. Fill in the key and value and save.
4. Then create a new locale for the same entry. 
5. The "key" field is now empty again, and I'm able to overwrite the key I previously set.
6. I can now add a different key. This will update for ALL locales that have been created.

### Expected Behavior

Since I turned "Enable localization for this field" OFF, and there is already a key available, I would expect that key to be copied to the new locale.

What would be even more awesome if we can define a default locale for the key, and make it uneditable in all other locales.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

alexandrebodin:
Hi @Lvdwardt thank you for this suggestions. We do not auto created content for new entries but only overwrite when you save. this is a nice improvement idea though. I'll cc our product owner to flag this as an idea @yanniskadiri :)

I'll close this as we will not be flagging it as a bug but a feature request. Thank you !

Lvdwardt:
Hey @alexandrebodin,

Thanks for the reply, but I don't think this is a feature request. In strapi v4 this is the default behaviour.
It's expected behaviour that when you have a "Non Translatable" field, the contents should Not change between locales, even when you create a new locale.

Here is a recording of what happens in a strapi v4 instance:

https://github.com/user-attachments/assets/ebd99ffc-ede9-4b6b-95e2-6b9ba70f24d7

hanpaine:
[...]

Issue #22080, opened 2024-11-06
Non-localized fields don't retain value when adding new locale


tc-david:
### Required System information

- **Node.js version**: 20.10.0
- **NPM version**: 10.2.3
- **Strapi version**: 5.2.0
- **Database**: MySQL
- **Operating system**: macOS Sequoia 15.1
- **Is your project Javascript or Typescript**: TypeScript

### Describe the bug

I created a collection type with localization enabled. This collection includes two fields:
- One field with "Enable localization for this field" checked.
- One field with this option unchecked (non-localized field).

I expected that the non-localized field would retain its value across all locales by default, meaning it should pre-fill with the same data for any additional locale created later. However, when adding a new locale (with "EN" as the default language and "FR" as the additional locale), the non-localized field appears empty instead of carrying over the EN value.

Additionally, if I populate the non-localized field in the FR locale and save/publish it, it overrides the value in the EN locale, essentially making both locales share the FR value instead of keeping the EN default. This behavior is inconsistent with expected functionality, where changes in non-localized fields should persist across all locales only when edited in the default locale.

### Steps to reproduce the behavior

1. Go to **Content-Types Builder**.
2. Create a new **collection type** with **localization** enabled.
3. Add two fields to this collection:
   - One with "Enable localization for this field" checked.
   - One with "Enable localization for this field" unchecked (non-localized field).
4. Go to **Content Manager**.
5. Add a new entry (default locale is **EN**).
6. Populate both fields and click **Save & Publish**.
7. Add the **FR** locale for this entry.
8. Observe that both fields are empty in the **FR** locale.
9. Populate both fields in the **FR** locale, then click **Save & Publish**.
10. Return to the **EN** locale and observe that the non-localized field now contains the data from the **FR** locale.

### Expected behavior

For non-localized fields:
- The value from the default locale (EN) should persist across all new locales as the initial value.
- Any updates to non-localized fields in additional locales should not affect the default locale but should retain consistency across all locales once edited.

### Screenshots

1. Create a new entry.
   ![Screenshot 2024-11-06 at 10 11 48](https://github.com/user-attachments/assets/a2e86c52-4fc2-4d20-9e5e-42ac1f8306c8)

2. Populate both fields and click **Save & Publish**.
   ![Screenshot 2024-11-06 at 10 12 01](https://github.com/user-attachments/assets/e79ad319-bbc8-49bb-aea2-91cfd6ecf285)

3. Add the FR locale for this entry.
[...]

PR #24659, opened 2025-10-22
fix: prefill non translatable fields when creating new document locale

### LINKED ISSUE #21850: Strapi 5 ignores "Enable localization for this field" setting
### Describe the bug
Strapi ignores the field  "Enable localization for this field" (considers it true for all fields). I need to input the same value manually in all locales. 

Go to Content-Type Builder / Collection types
Create new content type "Category" with 2 fields 
Title - Enable localization for this field: checked
Code - Enable localization for this field: not checked

Go to Content manager
Create new category with English locale. Fill Title and Code values. Save / Publish. 
Press add new locale

### Expected behavior
Code is prefilled from English locale

### Actual behavior
Code is not prefilled from English locale

image
image
image

Adzouz:
### What does it do?

When creating a new locale of a document, the non translatable fields (where "Enable localization for this field" is disabled) are empty but they should be pre-filled because their values are shared between all locales.

BEFORE:

https://github.com/user-attachments/assets/c7b0518d-2c1f-422d-a79f-2ca567598c39

AFTER:

https://github.com/user-attachments/assets/5e03aa12-84fe-45f6-bf8e-9a4f22678aeb

### Why is it needed?

Leaving the fields empty is confusing and the user needs to put them again (or use the "Fill in from another language" feature). It can be a waste of time and the new value entered could overwrite all other locales if different from the original one.

### How to test it?

- Make sure that you have a content type that has Internationalization enabled + have at least one field with "Enable localization for this field" to false
- Go to the content manager and select this content type
- Create a new entry and fill the field with some data
- Save
- Select another locale in the language selector to create a new variation of the document
-> The field you have previously filled should remain with the value you entered instead of being reset

### Related issue(s)/PR(s)

Resolves #21850 #22080 #23600

🚀

### DIFF
--- a/tests/e2e/tests/content-manager/relations-on-the-fly/create-relation-in-component-and-save.spec.ts
+++ b/tests/e2e/tests/content-manager/relations-on-the-fly/create-relation-in-component-and-save.spec.ts
@@ -34,11 +34,8 @@ test.describe('Relations on the fly - Create a Relation inside a component and S
     await expect(page.getByRole('status', { name: 'Draft' }).first()).toBeVisible();
 
     // Step 6. Close the relation modal to see the updated relation on the root document
-    const closeButton = page.getByRole('button', { name: 'Close modal' });
-    await closeButton.click();
-
-    // Wait for the modal to be closed
-    await expect(page.getByText('Create a relation')).not.toBeVisible();
+    await expect(page.getByText('Edit a relation')).toBeVisible();
+    await clickAndWait(page, page.getByRole('button', { name: 'Close modal' }));
 
     // Wait for the button to be visible with a more specific selector
     await expect(page.getByRole('button', { name: 'Nike Zoom Kd Iv Gold C800' })).toBeVisible();


packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -145,6 +147,91 @@ const LocalePickerAction = ({
...
+                  },
+                  {}
+                );
+
+                if (Object.keys(dataToSet).length > 0) {
+                  setValues(dataToSet);
+                }
+              }
+            } catch (error) {
+              console.warn(
```
mathildeleg: I'm guessing you didn't use a NotificationError because if it fails, it's not too bad of a problem if the values aren't set by default right?

tests/e2e/tests/i18n/editview.spec.ts
```diff
@@ -545,6 +545,49 @@ test.describe('Edit view', () => {
       page.getByLabel('Unpublish multiple locales').getByRole('button', { name: 'Unpublish' })
     ).toBeDisabled();
   });
+
+  test.only('As a user I want to see non translatable fields pre-filled when creating a new locale of a document', async ({
```
mathildeleg: `test('As a user I want to see non translatable fields pre-filled when creating a new locale of a document', async ({`

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
Adzouz: Yes exactly, that's because it's more a silent notification since if it's not working, it would pretty much fallback on what we currently have: the field not being filled. I'm not sure we want to display an error for that, otherwise the user will think that something went wrong even if they still can edit and publish their document. What's your opinion on this one?

tests/e2e/tests/i18n/editview.spec.ts
Adzouz: Dammit

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -145,6 +147,91 @@ const LocalePickerAction = ({
...
+    document,
+    schema,
+    setValues,
+    currentDesiredLocale,
+    locales,
+    documentId,
+    getDocument,
+    collectionType,
+    model,
+    meta,
```
HichamELBSI: Wow, that's a lot of dependencies, are you confident about it. I'm not sure to understand everything so I can only trust your confidence. I guess this requires some good testing.

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
mathildeleg: I agree with you 👍🏻 just wanted to make sure we were on the same page

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
Adzouz: Yeah I get it 😅 I'm confident about a few dependencies I didn't really had the choice like setValues, getDocument but the others it's because the getDocument needs the collectionType, the model, the documentId and the locale so it's a lot of dependencies I don't really have the choice but to add.
I tried to optimize (with memoization) and chunk it a bit, let me know if it's better.
What it's doing: fetching the source locale to get all non translatable fields to load them in the form (hence the getDocument + setValues).
Let me know if you have a better idea on how to do it 🙏

Adzouz:
After rubberducking this issue, I thought about another approach more stable and without having to do an additional API call. I use the existing availableLocales to autofill the non-translatable fields (only when creating a new locale). It's more efficient and avoid too many side effects.
Let me know what you think!

ancashoria:
Wow, you fixed this! Thank you so much!

PR #24660, opened 2025-10-22
Fix double go back links and header size on mobile

HichamELBSI:
### What does it do?

Remove the duplicate go back button and use the default form header on mobile.

### Why is it needed?

This PR fixes the double go back link that can is rendered when opening a setting subpage. It does also fix the header fixed height that we are using for the desktop mode.

### How to test it?

- Open the API token settings page (the header should be displayed normally)
- Open the create new token settings page (you should be able to see the submit button and a single go back link in the form header.

### Related issue(s)/PR(s)

| before    | after |
| -------- | ------- |
| Settings — API Tokens  Strapi | IMG_3837 |

| before    | after |
| -------- | ------- |
| IMG_3835 | IMG_3836 |

### DIFF
--- a/packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
+++ b/packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
@@ -62,8 +62,6 @@ describe('Roles – EditPage', () => {
 
     await waitForElementToBeRemoved(() => getByText('Loading content.'));
 
-    expect(getByRole('link', { name: 'Back' })).toBeInTheDocument();
-
     expect(getByRole('heading', { name: 'Authenticated' })).toBeInTheDocument();
     expect(getByRole('heading', { name: 'Role details' })).toBeInTheDocument();
     expect(getByRole('heading', { name: 'Permissions' })).toBeInTheDocument();
--- a/packages/core/admin/admin/src/components/Filters.tsx
+++ b/packages/core/admin/admin/src/components/Filters.tsx
@@ -173,7 +173,7 @@ const PopoverImpl = ({ zIndex }: { zIndex?: number }) => {
       ],
     };
 
-    setQuery({ filters: newFilterQuery, page: 1 });
+    setQuery({ filters: newFilterQuery, page: 1 }, 'push', true);
     setOpen(false);
   };
 
--- a/packages/core/admin/admin/src/components/Table.tsx
+++ b/packages/core/admin/admin/src/components/Table.tsx
@@ -187,9 +187,13 @@ const HeaderCell = <TData, THead>({ name, label, sortable }: TableHeader<TData, 
 
   const handleClickSort = () => {
     if (sortable) {
-      setQuery({
-        sort: `${name}:${isSorted && sortOrder === 'ASC' ? 'DESC' : 'ASC'}`,
-      });
+      setQuery(
+        {
+          sort: `${name}:${isSorted && sortOrder === 'ASC' ? 'DESC' : 'ASC'}`,
+        },
+        'push',
+        true
+      );
     }
   };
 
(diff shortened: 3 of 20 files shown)


mathildeleg (commented):
The code looks good but here's some edge cases I've found:
* The "Back" link on the Api Token page needs to be clicked on twice before I am redirected to previous page
* I don't have any "Back" link on the Settings pages for the Users & Permissions plugin (so: Roles, Providers Email Templates and Advanced Settings)

HichamELBSI:
> The code looks good but here's some edge cases I've found:
> 
> * The "Back" link on the Api Token page needs to be clicked on twice before I am redirected to previous page
> * I don't have any "Back" link on the Settings pages for the Users & Permissions plugin (so: Roles, Providers Email Templates and Advanced Settings)

Thanks for the review, seems like we have the "click twice to go back" issue on production as well. I'll try to fix it.

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
```diff
@@ -28,10 +29,7 @@ const BaseHeaderLayout = React.forwardRef<HTMLDivElement, BaseHeaderLayoutProps>
     if (sticky) {
       return (
         <Box
-          display={{
```
Adzouz, line 31: It was on purpose and discussed to hide the sticky header on mobile since it's taking some space and the user won't have enough space for the rest of the content. But after testing, I don't see the header appearing on mobile. So do you know why we needed to change that?

Adzouz:
I'm also experiencing the same issue reported by @mathildeleg and also on Transfer Tokens. It's also opening the modal again on Audit logs (after opening and closing a detail) but I know you're aware of the situation and I'll let you fix it properly 😊 Thanks again 🙏

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
HichamELBSI: It will still be hidden but higher in the page hierarchy to avoid the render of the sticky header for no reason.

HichamELBSI:
> I'm also experiencing the same issue reported by @mathildeleg and also on Transfer Tokens. It's also opening the modal again on Audit logs (after opening and closing a detail) but I know you're aware of the situation and I'll let you fix it properly 😊 Thanks again 🙏

Many pages are broken actually, I'm fixing them all 😁

HichamELBSI:
It should be fixed.

HichamELBSI:
@mathildeleg Thanks a lot for the review, I deleted the file that was wrongly pushed and deleted your comments so we can't see it anymore 😞.

mathildeleg (dismissed):
Looks good ✅

packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
```diff
@@ -62,8 +62,9 @@ describe('Roles – EditPage', () => {
 
     await waitForElementToBeRemoved(() => getByText('Loading content.'));
 
-    expect(getByRole('link', { name: 'Back' })).toBeInTheDocument();
-
+    // jsdom doesn't evaluate css media queries. Meaning that the back link will not be displayed in the tests.
+    // This should be tested through E2E tests.
+    // expect(getByRole('link', { name: 'Back' })).toBeInTheDocument();
```
Adzouz: Or maybe I misunderstood something, we don't want the back button in those pages?

mathildeleg: Are you sure Adrien? Cause on my side I do have the back button on those children pages 🤔

Adzouz:
Oops so sorry, I pushed on the wrong branch 😅 Don't pay attention to my mess, I put everything back as it was.
And don't pay attention to my previous comment. I think I wasn't testing on the right branch because it seems that it's working now.

packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
Adzouz: Can't we just remove it then?

HichamELBSI: done

Adzouz (approved):
Approved, thanks a lot and good job 🙌🚀

PR #24661, opened 2025-10-22
future(i18n): filter out complex attributes passed to ai server

remidej:
### What does it do?

3 things:
- prevents passing fields to the AI server, because the server messes it up way too often which makes everything fail. We'll reintroduce them when the server is ready
- Passes a mini version of the schema to the AI server. It's for a WIP I have on the AI server side, where I give it specific instructions depending on the field type
- Invalidates the RTK document cache when the translation job is done. Without that I had to refresh to see the result of a translation when there's already something in the cache for it. It would be better to have the invalidation logic within the service, but for some reason I couldn't get the invalidatesTags method to run so I moved on
### Why is it needed?

to have something testable tomorrow

### DIFF
--- a/packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
+++ b/packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
@@ -1,7 +1,8 @@
 import * as React from 'react';
 
-import { useNotification } from '@strapi/admin/strapi-admin';
+import { useNotification, adminApi } from '@strapi/admin/strapi-admin';
 import { useIntl } from 'react-intl';
+import { useDispatch } from 'react-redux';
 
 import { AILocalizationJobs } from '../../../shared/contracts/ai-localization-jobs';
 import { useGetAILocalizationJobsByDocumentQuery } from '../services/aiLocalizationJobs';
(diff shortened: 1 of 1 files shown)


packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
```diff
@@ -49,6 +51,18 @@ export const useAILocalizationJobsPolling = ({
...
+  const invalidateDocument = React.useCallback(() => {
+    dispatch(
+      adminApi.util.invalidateTags([
+        {
+          // @ts-expect-error tag isn't available
+          type: 'Document',
+          id: collectionType !== 'single-types' ? `${model}_${documentId}` : model,
+        },
+      ])
+    );
```
markkaylor, line 63: Couldn't this be done in the rtk service itself? 🤔

markkaylor: Just saw this: It would be better to have the invalidation logic within the service, but for some reason I couldn't get the invalidatesTags method to run so I moved on

markkaylor:
>Invalidates the RTK document cache when the translation job is done. Without that I had to refresh to see the result of a translation when there's already something in the cache for it

I'm glad this is working but I remember seeing this issue or something similar even before adding the notification system 🤔

packages/plugins/i18n/admin/src/hooks/useAILocalizationJobsPolling.ts
remidej: I think "invalidatesTag" in the service doesn't may not work for queries, only mutations. But it exists in the types 🤷

PR #24664, opened 2025-10-23
future(i18n): add ai translation status to locale picker

markkaylor:
### What does it do?

Adds the status icon for translation failure to the i18n LocalePicker

### Why is it needed?

To show the user what failed

### How to test it?

Throw an error somewhere in the ai-localization service in the backend to simualte an error from the ai server

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
@@ -139,9 +139,19 @@ interface HeaderActionDescription {
     startIcon?: React.ReactNode;
     textValue?: string;
     value: string;
-    render?: () => React.ReactNode;
+    /**
+     * @internal
+     * @description
+     * Internal SelectOption renderer used to display the status of AI translation background jobs
+     */
+    _render?: () => React.ReactNode;
   }>;
-  status?: {
+  /**
+   * @internal
+   * @description
+   * Internal document header action to display the status of AI translation background jobs
+   */
+  _status?: {
     message: React.ReactNode;
     tooltip?: React.ReactNode;
   };
@@ -410,8 +420,8 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
               {...action}
             >
               {action.options.map(({ label, ...option }) => {
-                if (option.render) {
-                  return option.render();
+                if (option._render) {
+                  return option._render();
                 }
 
                 return (
@@ -422,10 +432,10 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
               })}
             </SingleSelect>
           );
-        } else if (action.status) {
+        } else if (action._status) {
           return (
-            <HeaderActionStatus tooltip={action.status?.tooltip} key={action.id}>
-              {action.status.message}
+            <HeaderActionStatus tooltip={action._status?.tooltip} key={action.id}>
+              {action._status.message}
             </HeaderActionStatus>
           );
         } else {
--- a/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
+++ b/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
@@ -240,7 +240,7 @@ const LocalePickerAction = ({
         settings?.data?.aiLocalizations
       ) {
         return {
-          render: () => (
+          _render: () => (
             <React.Fragment key={index}>
               <SingleSelectOption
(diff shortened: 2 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -139,6 +139,7 @@ interface HeaderActionDescription {
     startIcon?: React.ReactNode;
     textValue?: string;
     value: string;
+    render?: React.ReactNode;
```
markkaylor: Again, not sure about the best API here. I have to do this only because we have the separator 

Image

remidej: I think what you have is fine! There are 2 things that could be done differently but idk if they're better:
- "render" or "customizeContent"? We have both in the codebase right?
- That render prop could also be at the select level, not the option, so the label/separator wouldn't be within one option but higher up

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -408,11 +409,17 @@ const HeaderActions = ({ actions }: HeaderActionsProps) => {
...
               {...action}
             >
-              {action.options.map(({ label, ...option }) => (
-                <SingleSelectOption key={option.value} {...option}>
-                  {label}
-                </SingleSelectOption>
-              ))}
+              {action.options.map(({ label, ...option }) => {
+                if (option.render) {
+                  return option.render;
```
remidej: it doesn't need to be called?

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -203,6 +233,44 @@ const LocalePickerAction = ({
 
       const permissionsToCheck = currentLocaleDoc ? canRead : canCreate;
 
+      if (isAiAvailable) {
```
remidej: this only checks if AI is globally enabled/available, but you're not checking if i18n AI specifically is enabled. If it's not I don't think we should show this UI. We already have the other document action to push users to enable it

markkaylor: Yep good catch my bad

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
markkaylor: Originally I had it as a function but since it's not receiving any props I didn't see the point... Maybe the name is confusing then

markkaylor: But to answer your question as long as I'm not passing a function it doesn't to be called

remidej: Makes sense but looks pretty confusing compared to the usual render prop pattern. I think you could make it a function for consistency, even if it doesn't need props

markkaylor: Yeah that's confusing I'll just make it a function

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
markkaylor: customizeContent I wouldn't use for anything else because technically it specifically a prop for select https://design-system.strapi.io/iframe.html?viewMode=docs&id=inputs-select--docs&globals=&args=#props

I'll consider and try your suggestion

remidej: sounds good, again I'm good with your proposal, just wanna make sure we think consider different options before committing to a new API

markkaylor: I decided to leave it where it is, my thinking is this API is sort of restrictive and if we put the render at the select level the user could render anything they want and turns the document header actions api into an injection zone

markkaylor: What are your thoughts on marking these apis we're adding as internal with jsdoc and `_` prefix to the props?

remidej: That sounds good it would be less confusing to users if it's obvious what they should be using

Issue #24040, opened 2025-07-28
Something went wrong for Last edited entries after migration to strapi v5


MaksZhukov:
### Node Version

20.0.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.19.0

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

After migration to strapi v5.19.0 from 4.25.23 i see an error with last Edited entries and also I see that draft/publish statuses disappeared
The process of migration was done good without any error

### Steps to Reproduce

Have a project with v 4.25.23
Run npx @strapi/upgrade major
See the issue

### Expected Behavior

Have a project with v 4.25.23
Run npx @strapi/upgrade major
See the draft/publish feature and Last edited entries on dashboard page

### Logs

```shell

```

### Media

<img width="1398" height="1215" alt="Image" src="https://github.com/user-attachments/assets/f4ad0f5a-d37f-45ba-8dc3-5d383b77e8e1" />

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

oiorain:
👋 Hello, and apologies for the confusion!

As part of an automated issue cleanup, this issue was mistakenly closed. That action was unintentional — we're in the process of reviewing and reopening affected issues like this one.

🙏Thank you for your patience, and sorry again for any disruption.

oiorain:
Hi @MaksZhukov 

Thanks a lot for opening this issue and help make strapi better! We truly appreciate it.

We looked into it but despite multiple persons trying, we were not able to reproduce the bug so we weren't able to test the solution you are provided by the associated PR.
For now, I will close this issues. But if you can provide us with more precise steps to test, we'll happily pick up the work again.

Thanks again!

Issue #22676, opened 2025-01-20
Homepage last update/publish section not checking the permission condition


phumcentral:
### Node Version

20.0.0

### NPM/Yarn/PNPM Version

10.0.0

### Strapi Version

5.7.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When accessing the Strapi homepage with "read" permissions, users can see the "Last updated" and "Last published" fields as expected. However, if I configure the "read" permission with the condition `has same role as creator`, users are still able to see the entry records in the list.

The issue arises when the user tries to click on an entry. Since they lack the necessary permission, they are redirected to an invalid page. Ideally, the entries should not appear in the list if the user doesn't have full permission to access them.

### Steps to Reproduce

1. Configure a "read" permission for a collection with the condition `has same role as creator.`
2. Log in as a user with this permission.
3. Access the Strapi homepage and view the list of entries.
4. Click on an entry that does not meet the condition.

### Expected Behavior

Entries that do not meet the has same role as creator condition should not be visible in the list to avoid confusion and broken links.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

oiorain:
Hi @phumcentral 

Thanks a lot for opening this issue and help make strapi better! We truly appreciate it.

We looked into it but despite multiple persons trying, we were not able to reproduce the bug so we weren't able to test the solution you are provided by the associated PR.
For now, I will close this issues. But if you can provide us with more precise steps to test, we'll happily pick up the work again.

Thanks again!

PR #24668, opened 2025-10-23
fix(content-manager): filter invalid fields in homepage recent-documents query

0-io-0:
Prevents "Invalid fields parameter. Expected a string or an array of strings" 500 error in the admin homepage “Last edited entries / Recent documents” widget.

Relates to #24040

What does it do?

In packages/core/content-manager/server/homepage/services/homepage.ts, inside queryLastDocuments, it filters permissionQuery.fields (after sanitizedQuery.read) to ensure it is an array of non-empty strings before calling findMany()

`// minimal fix: ensure fields is an array of non-empty strings
if (permissionQuery && Array.isArray((permissionQuery as any).fields)) {
  (permissionQuery as any).fields = (permissionQuery as any).fields.filter(
    (f: unknown): f is string => typeof f === 'string' && f.length > 0
  );
}`

Why is it needed?

After migration from v4 → v5, the homepage “Recent documents” query can include a fields array where some entries become null after permission sanitization (e.g., when a field like provider is not permitted on plugin::users-permissions.user).
The document service validator expects fields to be a string or an array of strings; any non-string entry (like null) causes a 500 with:
`Invalid fields parameter. Expected a string or an array of strings`
Normalizing the array prevents the crash without changing any other query behavior (sorting, filters, populate, or permissions).

How to test it?
	1.	Use a Strapi v5 app where the admin dashboard “Recent documents” widget is enabled.
	•	Migrated app from v4 → v5 can reproduce more easily.
	2.	Ensure you have content types including plugin::users-permissions.user (or another model where a field may be stripped by permissions).
	3.	Open the admin dashboard → Home.
	4.	Before fix: Server logs show the 500 and the error above from convert-query-params.
	5.	Apply this patch, rebuild/run.
	6.	After fix: Dashboard loads the widget successfully; logs show no 500.

Environment verified:
	•	Node v20.x
	•	Strapi v5.29.0 (also reproduced on 5.18.0 before upgrading)
	•	DB: (e.g., SQLite locally; reporter also observed on MySQL)
	•	OS: macOS

Related issue(s)/PR(s)
	•	Relates to #24040 (admin homepage “Last edited entries” error after migration to v5)

Notes
	•	This is a minimal, defensive guard local to the homepage service.
	•	It doesn’t modify permission logic or other query params.
	•	Tests welcome: a unit test could assert that queryLastDocuments does not throw when fields contains [ 'documentId', 'updatedAt', null ] after permission sanitization.

### DIFF
--- a/packages/core/content-manager/server/src/homepage/services/homepage.ts
+++ b/packages/core/content-manager/server/src/homepage/services/homepage.ts
@@ -189,6 +189,13 @@ const createHomepageService = ({ strapi }: { strapi: Core.Strapi }) => {
             ...additionalQueryParams,
           });
 
+          // minimal fix: ensure fields is an array of non-empty strings
+          if (permissionQuery && Array.isArray((permissionQuery as any).fields)) {
+            (permissionQuery as any).fields = (permissionQuery as any).fields.filter(
+              (f: unknown): f is string => typeof f === 'string' && f.length > 0
+            );
+          }
+
           const docs = await strapi.documents(meta.uid).findMany(permissionQuery);
           const populate = additionalQueryParams?.populate as string[];
 


0-io-0:
maybe related #22676

oiorain:
Hi @0-io-0 ! 

Thanks a lot for contributing this PR to help make strapi better! We truly appreciate it.

We looked into it but despite multiple persons trying were not able to reproduce the bug so we weren't able to test the solution you are providing. 
For now, I will close this PR and the associated bugs. But if you can provide us with more precise steps to test, we'll happily pick up the work again.

Thanks again!

PR #24670, opened 2025-10-24
AWS S3 provider: Add forcePathStyle hint and migration note

felix-hilden:
### What does it do?

Hi! While trying to set this up for Contabo.com I ran into some issues but was able to resolve them. I think these documentation changes might help other users to go in the right directions.

### DIFF
--- a/packages/providers/upload-aws-s3/README.md
+++ b/packages/providers/upload-aws-s3/README.md
@@ -24,7 +24,7 @@ npm install @strapi/provider-upload-aws-s3 --save
 ## Configuration
 
 - `provider` defines the name of the provider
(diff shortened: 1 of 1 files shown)


packages/providers/upload-aws-s3/README.md
```diff
@@ -139,6 +139,8 @@ module.exports = ({ env }) => ({
 });
 ```
 
+For providers that place buckets in the path rather than as a subdomain, use the `forcePathStyle` parameter ([without the s3 prefix](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/migrating/notable-changes/)).
```
felix-hilden, line 142: Unfortunately this change is not mentioned in the official docs, and the migration page does not have anchors for the individual changes :(

markkaylor:
Closed by https://github.com/strapi/strapi/pull/25263

PR #24671, opened 2025-10-24
chore(upload): add isAIMediaLibraryConfigured to every tracking event

markkaylor:
### What does it do?

Wraps all tracking requests to include `isAIMediaLibraryConfigured` as an event property in every request

### Why is it needed?

To tracking ai usage, it's what product wants

### How to test it?

On the frontend when you do things that dispatch a tracking event like edit caption, edit alternative text, add a folder, etc... you should see in the network request 

Screenshot 2025-10-24 at 10 59 02

On the backend you can add a log to test an event. I don't think we can realistically test every event. 

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/upload/admin/src/hooks/__mocks__/useSettings.ts
+++ b/packages/core/upload/admin/src/hooks/__mocks__/useSettings.ts
@@ -1,0 +1,11 @@
+export const useSettings = jest.fn().mockReturnValue({
+  isLoading: false,
+  isError: false,
+  data: {
+    sizeOptimization: true,
+    responsiveDimensions: true,
+    autoOrientation: true,
+    aiMetadata: true,
+  },
+  error: null,
+});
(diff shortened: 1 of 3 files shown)


packages/core/upload/server/src/services/metrics.ts
```diff
@@ -1,14 +1,26 @@
 import type { Core } from '@strapi/types';
+import { Settings } from '../controllers/validation/admin/settings';
 
 const getProviderName = () => strapi.config.get('plugin::upload.provider', 'local');
 const isProviderPrivate = async () => strapi.plugin('upload').provider.isPrivate();
 
 export default ({ strapi }: { strapi: Core.Strapi }) => ({
+  async trackUsage(event: string, properties?: Record<string, any>) {
+    const settings: Settings = await strapi.plugin('upload').service('upload').getSettings();
+    console.log('did do something I want', settings.aiMetadata);
```
Adzouz: You left a console.log 😬

Adzouz:
I tested and everything seems good 👍 Only thing is that you also need to fix the tests 😅

packages/core/upload/server/src/services/metrics.ts
markkaylor: Ooops! Thanks

markkaylor:
@Adzouz I think all good now!

PR #24672, opened 2025-10-24
future(i18n): trigger localization for complex fields

remidej:
### What does it do?

- Restores blocks, DZs, components to the list of fields that should be translated
- Builds a minimal contentTypeSchema to pass it to the AI server (see https://github.com/strapi/strapi-ai/pull/110)
- Adds a dynamic zone to the "about" single type for easier testing

### Why is it needed?

You'll need the local AI server on that branch: https://github.com/strapi/strapi-ai/pull/110

### DIFF
--- a/examples/getstarted/src/api/about/content-types/about/schema.json
+++ b/examples/getstarted/src/api/about/content-types/about/schema.json
@@ -51,6 +51,15 @@
       },
       "multiple": false,
       "allowedTypes": ["images", "files", "videos", "audios"]
+    },
+    "dynamiczone": {
+      "type": "dynamiczone",
+      "pluginOptions": {
+        "i18n": {
+          "localized": true
+        }
+      },
+      "components": ["basic.simple", "blog.test-como", "default.closingperiod"]
     }
   }
 }


packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -138,6 +135,31 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
         });
       }
 
+      /**
+       * Provide a schema to the LLM so that we can give it instructions about how to handle each
+       * type of attribute. Only keep essential schema data to avoid cluttering the context.
+       * Ignore fields that don't need to be localized.
+       * TODO: also provide a schema of all the referenced components
```
remidej, line 142: Adding the components requires recursive logic. So I'm planning to replace this by extending the traverseEntity that @araksyagevorgyan is working on her branch once that is ready

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -138,6 +135,31 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+        Object.entries(schema.attributes)
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          .filter(([_, attr]) => {
+            const isLocalized = isLocalizedAttribute(attr);
+            const isSupportedType = !UNSUPPORTED_ATTRIBUTE_TYPES.includes(attr.type);
+            return isLocalized && isSupportedType;
+          })
+          .map(([key, attr]) => {
+            const minimalAttribute = { type: attr.type };
+            if (attr.type === 'component') {
```
markkaylor, line 154: Not sure if it's useful but the utils package exports helpers for components and dynamic zones

https://github.com/strapi/strapi/blob/662e18309746550b356023ea04a5fb63e91a0492/packages/core/utils/src/content-types.ts#L190-L212

remidej: Thanks for the link, it may be useful in later when I do the recursive stuff, for now `attr.type === 'component'` is all i need

packages/plugins/i18n/server/src/services/ai-localizations.ts
araksyagevorgyan: I think we can already merge my part and test it with ai service fix you have done. However, there are some conflicts with develop on my branch, so I think it would be great if we could have a look together to resolve them @remidej

markkaylor:
I'm not sure what's up but it doesn't seem to be translating if the request has dz / components


https://github.com/user-attachments/assets/1cc18b4f-7782-4b04-8c8c-fb0d080577c5

packages/plugins/i18n/server/src/services/ai-localizations.ts
remidej: I think we should merge this to develop (along with the ai server one) first then fix your conflicts when updating your branch. I'm happy to help 👍

remidej:
Sorry @markkaylor I should have mentioned that. Dynamic zones and components are not translated yet. Because that relies on the deep populating logic in the middleware, which is on @araksyagevorgyan's PR. What you can test that didn't work before though is blocks field

But that's why I was saying there's not that much we can QA for now, we first need to put the different pieces together

markkaylor:
> Sorry @markkaylor I should have mentioned that. Dynamic zones and components are not translated yet. Because that relies on the deep populating logic in the middleware, which is on @araksyagevorgyan's PR. What you can test that didn't work before though is blocks field
> 
> But that's why I was saying there's not that much we can QA for now, we first need to put the different pieces together

Makes sense, but this is technically a regression now I think, because it is not even translating the content that isn't in a dz / component. If the request includes dz / components translation does not happen (but we get a success notification). As long as you are aware of that and ok with it cool with me.

Issue #24667, opened 2025-10-23
unable to clear and save time only fields in the content manager


cdreier:
### Node Version

v20.12.0

### NPM/Yarn/PNPM Version

10.5.0

### Strapi Version

v5.28.0

### Operating System

Linux (RedHat/CentOS)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

when using a `Date` field as "time only" - empty default value, not required, it is not possible to clear the value, after every save the time is back in the field. 

### Steps to Reproduce

* use a "time" field 
* try to clear and save it

https://github.com/user-attachments/assets/3336af09-2e23-4a7a-8b57-1481498982b9

### Expected Behavior

the time is set to null

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

hanpaine:
@butcherZ maybe linked to #24073 - wdyt?

cdreier:
as far as i could see, the update api call does still include the time, looks more like a frontend issue that the clear button does only clear the form and not the actual state?

selmy28:
I recorded the whole session with FlowLens including video, network request, storage events, and user actions. 

<img width="1700" height="1061" alt="Image" src="https://github.com/user-attachments/assets/179e3522-4ff9-429b-b48d-855cc3120bf5" />

https://flowlens.magentic.ai/shared/flow/eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmbG93X2lkIjoiOGZkNzZlNWEtNTU1Ny00ODgwLTg1OGQtNTAxNGFiNDA3ZTQyIiwiZXhwIjoxNzkzMTc1NjgzLCJpYXQiOjE3NjE2Mzk2ODMsInNoYXJlZF9ieSI6Im9mYWhteTEyMzRAZ21haWwuY29tIiwiZW1haWwiOiJvZmFobXkxMjM0QGdtYWlsLmNvbSJ9.Ey1xTIHLNb8scEI91muNkZT3Ha--tjVEwLXjEsHAWGM

dagadevelop:
It’s a frontend issue — the input is cleared, but the state is retained. It’s being set to undefined when it should be null.

PR #24675, opened 2025-10-24
fix: unable to clear and save time only fields in the content manager…

dagadevelop:
# Fix: Time input sends previous value to backend when cleared

## What does it do?

This fix corrects the Time input component to properly send `null` to the backend when the field is cleared.

**Technical changes:**

1. Changed the field type to accept `null`:

```typescript
// Before
const field = useField<string>(name);

// After
const field = useField<string | null>(name);
```

2. Changed the `onClear` handler to set `null` instead of `undefined`:

```typescript
// Before
onClear={() => field.onChange(name, undefined)}

// After
onClear={() => field.onChange(name, null)}
```

**Location:** `packages/core/admin/admin/src/components/FormInputs/Time.tsx` (lines 14 and 28)

## Why is it needed?

When users cleared a time field using the clear button, the field would visually appear empty in the UI, but **the frontend was still sending the previous value to the backend**, preventing the field from being properly cleared in the database.

**Root cause (frontend bug):**

- The `onClear` handler was setting the field value to `undefined`
- The field remained with its previous value in the form state
- When saving, the frontend sent the **old time value** instead of a cleared field: `{ title: "Example", timeField: "14:30:00.000" }`
- **Result:** The UI showed an empty field visually, but the form state still had the old value, which was sent to the backend

**After fix (frontend):**

- The `onClear` handler now sets the field value to `null`
- `null` is a valid value that properly updates the form state
- The frontend sends `{ title: "Example", timeField: null }`
- The backend correctly receives `null` and clears the field in the database
- **Result:** Both UI and database are in sync - the field is properly cleared

### Setup

1. Create a content-type with a `time` field
2. Create an entry and set a time value (e.g., `14:30`)
3. Save the entry

## Related issue(s)/PR(s)

Fix [#24667](https://github.com/strapi/strapi/issues/24667)


butcherZ (approved):
Tested locally, LGTM! Thank you for the contribution and thanks for adding tests!

Issue #24678, opened 2025-10-24
Strapi crashes when filtering an integer field using a string value instead of returning a validation error


CRIS19N:
### Node Version

22.17.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.29.0

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When performing a **POST** request to create an entry that includes an integer (or bigint) field, if the value provided is not a valid number (for example `"900260056-1"`), **Strapi crashes internally** instead of returning a proper validation or HTTP 400 error.

This behavior also happens during **GET** filtering, but in that case Strapi throws a handled `Expected a valid Number` error.  
During **POST**, however, the PostgreSQL driver raises an unhandled exception (`invalid input syntax for type bigint`), which stops the Strapi process completely.

---

### Steps to Reproduce

1. Create a content type (for example, `terceros`) with a numeric field:
   ```json
   {
     "Nit": { "type": "integer" }
   }
   ```

2. Perform a **POST** request:
   ```http
   POST /api/terceros
   Content-Type: application/json

   {
     "data": {
       "Nit": "900260056-1"
     }
   }
   ```

3. Observe that Strapi crashes with a PostgreSQL driver error instead of returning a 400 Bad Request.

---

### Expected Behavior

Strapi should gracefully validate and reject the invalid numeric input before reaching the database layer:

```json
{
  "error": {
    "status": 400,
    "message": "Invalid value for field 'Nit'. Expected a valid number."
  }
}
```

This would prevent the process from crashing and align with Strapi’s standard validation behavior for filters.

---

### Logs

```shell
[2025-10-24 15:04:09.019] http: POST /api/terceros (23 ms) 500
error: select "t0"."id" from "public"."terceros" as "t0"
where ("t0"."nit" = $1 and "t0"."published_at" is not null)
limit $2 - invalid input syntax for type bigint: "900260056-1"

DatabaseError: invalid input syntax for type bigint: "900260056-1"
    at Parser.parseErrorMessage (...)
    at Parser.handlePacket (...)
    at Parser.parse (...)
```

---

---

---

### Additional information

- The crash happens before persistence, during the ORM layer’s `NumberField.toDB()` conversion.
- The issue originates in:  
  `@strapi/database/dist/fields/number.js → NumberField.toDB()`
- A pre-validation should be added to catch and handle invalid numeric strings in both filters and payloads.

---

[...]

PR #24684, opened 2025-10-25
fix: prevent crash on invalid numeric strings in NumberField

Pseudo-Sid26:
**What does it do?**
Replaces toNumber (uses parseFloat) with Number() constructor for strict string validation in NumberField.toDB()
Adds explicit validation to reject arrays, objects, and invalid numeric strings
Throws clear error messages before data reaches the database layer
Adds comprehensive test suite with 145+ assertions

**Why is it needed?**
Fixes Issue #24678: Strapi crashes when POST requests contain invalid numeric strings (e.g., "900260056-1") for integer/bigint fields
parseFloat("900260056-1") returns 900260056 (partial parse) which passes validation but causes PostgreSQL to crash
Number("900260056-1") returns NaN, properly rejecting invalid input
Prevents unhandled database exceptions and returns proper HTTP 400 validation errors

**How to test it?**
Environment:

Node.js 18+
PostgreSQL database
Strapi 5.x

Test Steps:
Create a content type with an integer field (e.g., Nit)
Send POST request: POST /api/terceros { "data": { "Nit": "900260056-1" } }
Before fix: Strapi crashes with PostgreSQL error
After fix: Returns HTTP 400 with validation error
Run unit tests: cd packages/core/database && npm test -- number.test.ts
Verify all 145+ test cases pass

**### Related issue(s)/PR(s)**


CRIS19N:
I'm currently facing similar issues when using `biginteger` or `number`  fields in Strapi (PostgreSQL backend).

In my case, if a value exceeds the database integer range or the payload contains a non-numeric value (for example, `"900260056-1"`), Strapi doesn’t properly validate it before passing it to the ORM. This causes a fatal error like:


This completely crashes the Strapi process during migration or data import, rather than returning a clean validation error (400 Bad Request).

This PR seems to address exactly that issue — ensuring correct type validation before ORM execution.  
Would love to know if there's any update or timeline for merging this. It would help avoid full service crashes in production when dealing with large numeric fields.

Pseudo-Sid26:
The PR is currently awaiting maintainer review on the official Strapi
repository. It is available for testing in my fork:
Pseudo-Sid26/strapi:number-field-validation.

To help expedite the merge, you can assist by:
* Testing the fix in your environment using the provided fork.
* Sharing your results on GitHub issue #24678.
* Reacting to the issue to demonstrate community interest, which helps
prioritize the review.

innerdvations:
Thanks for the contribution and for digging into [#24684](https://github.com/strapi/strapi/pull/24684)!
Unfortunately, I found that the included test suite is not passing as submitted, and the change doesn’t fully cover the reported failure path (especially around biginteger/unique-query validation). It also risks introducing compatibility issues for projects with legacy malformed bigint values already stored in the database.

To move this forward safely, I opened a follow-up PR that addresses the full number + bigint behavior and keeps backward compatibility for existing data (I credited you there for the original investigation and proposal)

https://github.com/strapi/strapi/pull/26101

PR #24700, opened 2025-10-27
Add audit log plugin implementation

anuragj036-create:
### What does it do?

Adding a Automated Audit report log plugin

### Why is it needed?

It is needed to add audit logs in case of user cruds events


oiorain:
Hi there! 
Thank you for your contribution. 
We already have an audit logs feature on V5 so I will be closing this PR.

PR #24707, opened 2025-10-27
chore(i18n): dispatch tracking event for ai feature

markkaylor:
### What does it do?

- Adds a new metrics service method to send events with an isAIi18nConfigured event property

I am noticing the middleware is called multiple times which is potentially problematic. 

For an update action it's triggered twice
For a publish action three times (update, publish, and then update again)

Not just for tracking, I'm not wondering if in the middleware we are calling the ai server several times 🤔 

### Why is it needed?

### How to test it?

Add a log in the metrics service method like

console.log('SENDING AI EVENT', Boolean(settings?.aiLocalizations), event, payload);

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/content-squad/issues/99

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -286,8 +286,18 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
       strapi.documents.use(async (context, next) => {
         const result = await next();
 
+        const metricsService = strapi.plugin('i18n').service('metrics');
+        if (context.action === 'publish') {
+          metricsService.sendWithAIEventProperty('didPublishEntry');
+          return result;
+        }
+
+        if (context.action === 'update' || context.action === 'create') {
+          metricsService.sendWithAIEventProperty('didSaveEntry');
+        }
+
         // Only trigger for the allowed actions
-        if (!['create', 'update', 'publish'].includes(context.action)) {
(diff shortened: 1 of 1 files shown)


packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -280,8 +286,8 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
       strapi.documents.use(async (context, next) => {
         const result = await next();
 
-        // Only trigger on create/update actions
-        if (!['create', 'update'].includes(context.action)) {
+        // Only trigger for the allowed actions
+        if (!['create', 'update', 'publish'].includes(context.action)) {
           return result;
         }
```
remidej: I understand why this is necessary for tracking, but this would also generate the localizations when clicking publish right? I'm not sure that's what we want since publish doesn't actually update content

markkaylor: gotcha, in that case I guess I could use the eventHub 🤔 wdyt?

remidej: Oh I was thinking of having a dedicated check in here and returning early, but event hub works too

markkaylor: oh yeah good point I'll just do it before and return for publish, I think it's easier to do it here and I prefer the colocation if possible instead of tapping into another system just for tracking.

Issue #24285, opened 2025-09-01
Impossible to reset enum field to empty value


sergejpapper:
### Node Version

20.19.1

### NPM/Yarn/PNPM Version

Yarn 1.22.22

### Strapi Version

5.23.1

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

It impossible to reset to empty value enumeration field.

Please check steps to reproduce for more info.

### Steps to Reproduce

1. Create a collection with enum field which contains some options (for ex. "a", "b", "c"). It important that this field must be not required.

`"test": {
      "type": "enumeration",
      "pluginOptions": {
        "i18n": {
          "localized": true
        }
      },
      "enum": [
        "a",
        "b",
        "c"
      ]
    }
} `

2. Go to this collection via Content Manager, set enum field value (for ex. to "a") and publish it.
3. Try to reset value by setting "Choose here" value and try to publish collection one more time.

Error appears.

<img width="803" height="147" alt="Image" src="https://github.com/user-attachments/assets/8524cc46-ac5a-4efc-8d24-5e1e01602a56" />

### Expected Behavior

It should be possible to reset enum to empty value.

### Logs

```shell

```

### Additional information

I think this is high severity bug, because CMS administrators can not modify content in case of mistake.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

ericbf:
Is this working in a previous version? Is there a current workaround for this (other than adding a dummy value)?

Adzouz:
Hello there,
I just merged the fix for this issue 🙌
It will be part of next release (Wednesday November 5th).
Closing the issue 🚀

Issue #24174, opened 2025-08-13
Cannot 'unselect' value from enumeration within component


mlequime:
### Node Version

22.15.1

### NPM/Yarn/PNPM Version

yarn 1.22.22

### Strapi Version

5.11.2

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

When defining an enumeration within a repeatable component, with two values, once a value is selected it is no longer possible to 'unselect' the value by selecting Choose here. This happens when the field is **not** required. The component will continually fail validation unless it is deleted and replaced with a fresh one.

<img width="463" height="102" alt="Image" src="https://github.com/user-attachments/assets/99e14836-b6fd-418c-8376-569f9dba465c" />

### Steps to Reproduce

1. Create a document with a repeatable component (may or may not be required)
2. Add an enumeration to the component.
3. Create a document where the enumeration is set.
4. Try to select 'Choose here'

### Expected Behavior

The component should be back in the undefined/null/empty state, and pass validation.

### Logs

```shell
None
```

### Code Snippets

Schema:
``` json
"insights_text": {
 "type": "enumeration",
 "enum": [
    "normal",
    "advice"
  ],
  "required": false
}
```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

MaximohFoxy:
**Avoid using enumerations for optional fields where you need to be able to revert to null/blank.**
If you really need the field to be left blank, it should be of type string and handled on the frontend/backend (you can have a custom select that allows an empty or null value), even though you’ll lose the built‑in enum validation.

Use an explicit value in the enumeration list for the “null” or “unselected” state.
Add an option in the enum such as "None" or "Not specified", and in the frontend interpret it as an empty value when retrieving it.

Adzouz:
Hi,
I just merged the fix for this issue 🙌
It will be part of next release (Wednesday November 5th).
Closing the issue 🚀

Issue #22967, opened 2025-02-21
Optional enum "Choose here" as empty value throws error


ellawir:
### Node Version

20

### NPM/Yarn/PNPM Version

NPM 10.9.0

### Strapi Version

5.10.3

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/ellawir/strapi-latest-test

### Bug Description

Optional enumeration fields throw validation error when selecting the "Choose here" value. Similar bugs have previously been mentioned [here](https://github.com/strapi/strapi/pull/22141) and the "Choose here" option seems to have been added back to optional fields on v5 with [this](https://github.com/strapi/strapi/pull/22139) PR, but as the provided minimal reproducible example shows, selecting "Choose here" throws an error in later versions. It seems unnecessary that we'd have to add an option for empty values on optional enums. 

### Steps to Reproduce

1. Create a content type with an optional enum ("Test" collection type in provided reproduction repository)
2. Create new entry in content manager, select enum = option1 and successfully save

![Image](https://github.com/user-attachments/assets/a34144c6-e936-4400-becb-84dc22b488f5)

3. Select enum = "Choose here" and get error “enum must be one of the following values: option1, option2”

![Image](https://github.com/user-attachments/assets/f5a62368-93bd-45d7-9128-eadc210e4e89)

### Expected Behavior

As discussed in related bug and PR, "Choose here" should be selectable as empty value on optional enum field, as it was in Strapi 4.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

schtiewen:
same issue on 5.16.1. i cannot clear an enum field

phumcentral:
Same issue on 5.14.0

PeterBowater:
Still affecting on 5.17

troyere:
Same issue on 5.23.0

rachidBensaid:
Same issue for version 5.25.0

Adzouz:
Hi,
I just merged the fix for this issue 🙌
It will be part of next release (Wednesday November 5th).
Closing the issue 🚀

bks-kristin:
same issue on the 5.7.0...
any ideas how to fix it? 
[...]

PR #24708, opened 2025-10-27
fix: reset and publish enum default value

Adzouz:
### What does it do?

Fixing the reset of enumeration inputs.
We're having the problem in the content manager, when we have a non required enum, it's currently not possible to reset its value and publish without having a validation error. The value should reset to null instead of an empty string.

Before:

https://github.com/user-attachments/assets/0a4cddee-2642-4eb1-af4e-b3e69315b05c

After:

https://github.com/user-attachments/assets/5cd2205f-55d2-484d-94fb-a34f4f46320c

### Why is it needed?

We should be able to reset non-required enums values and publish documents.

### How to test it?

- Go to the content manager and select a content type that has a non-required enumeration field (if you don't have any, please set it up before in the content type builder)
- Add a new entry
- Select a value in the enumeration combobox
- Click Publish
- Click again on the combobox and on "Choose here" to reset the value
- Click Publish
-> The document should be correctly saved without any validation error

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/24285 https://github.com/strapi/strapi/issues/22967 https://github.com/strapi/strapi/issues/24174

🚀

### DIFF
--- a/tests/e2e/tests/content-manager/relations-on-the-fly/create-relation-in-new-component-and-save.spec.ts
+++ b/tests/e2e/tests/content-manager/relations-on-the-fly/create-relation-in-new-component-and-save.spec.ts
@@ -37,11 +37,8 @@ test.describe('Relations on the fly - Create a Relation inside a new component a
     await expect(page.getByRole('status', { name: 'Draft' }).first()).toBeVisible();
 
     // Step 7. Close the relation modal to see the updated relation on the root document
-    const closeButton = page.getByRole('button', { name: 'Close modal' });
-    await closeButton.click();
-
-    // Wait for the modal to be closed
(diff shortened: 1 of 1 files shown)


Adzouz:
@HichamELBSI I thought it would be better and would actually makes sense to put it directly on the field in the admin but if you want me to contain the change in the content manager input renderer, just tell me :)

mathildeleg (approved):

Adzouz:
I pushed again because the targeting of elements in my test wasn't great 😅
Also fixed a really annoying flaky test that is still happening too much (see https://github.com/strapi/strapi/actions/runs/18880608435/job/53953683710?pr=24714), it shouldn't bother us anymore 🙏

HichamELBSI (approved):

Issue #24624, opened 2025-10-17
/api/upload/files get endpoint doesn't support pagination


colinduwe:
### Node Version

20.11.0

### NPM/Yarn/PNPM Version

NPM 10.8.2

### Strapi Version

5.24.1

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When making requests to the /api/upload/files endpoint the pagination params have no effect. 

### Steps to Reproduce

http://localhost:1337/api/upload/files/?pagination[pageSize]=2
Returns all the files in the media library, more than the 2 per page request.

http://localhost:1337/api/upload/files/?pagination[pageSize]=2&pagination[page]=2&pagination[withCount]=true
Returns exactly the same response as the prior request; all the files in one array. It doesn't offset by 2nd page nor does it include the pagination count in the response.

### Expected Behavior

http://localhost:1337/api/upload/files/?pagination[pageSize]=2
Should return an array of 2 file items.

http://localhost:1337/api/upload/files/?pagination[pageSize]=2&pagination[page]=2&pagination[withCount]=true
Should return an array of 2 items that represent the 2nd page of the collection. It should also return meta.pagination data in the response. 

### Logs

```shell

```

### Additional information

@boazpoolman noted in Discord
Looking at the code there is actually a findPage service in the Upload package:
https://github.com/strapi/strapi/blob/develop/packages/core/upload/server/src/services/upload.ts#L493-L507
But it's not being used in the findMany controller:
https://github.com/strapi/strapi/blob/develop/packages/core/upload/server/src/controllers/content-api.ts#L37-L44

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

derrickmehaffy:
FindPage afaik was built only for the admin panel (I have no idea why we did that)

colinduwe:
I've confirmed that using findPage in place of findMany here
https://github.com/strapi/strapi/blob/develop/packages/core/upload/server/src/controllers/content-api.ts#L37-L44
results in the response I expected.

I'm new to Strapi. What's the best way to implement a fix? Should we add a cofig flag to allow users to opt-in to the paginated response?

colinduwe:
I submitted a PR to fix. I wasn't sure where to document the config flag.

[...]

PR #24711, opened 2025-10-27
Fix: Enabled /api/upload/files to optionally respond with pagination

colinduwe:
### What does it do?

This fix addresses https://github.com/strapi/strapi/issues/24624
Enabled /api/upload/files to optionally respond with pagination. Includes config flag to maintain existing response.

### Why is it needed?

This fix addresses https://github.com/strapi/strapi/issues/24624
When making requests to the /api/upload/files endpoint the pagination params have no effect.

### How to test it?

Currently a request such as
http://localhost:1337/api/upload/files?pageSize=10&page=2&sort=createdAt:DESC
Will result in a response containing all the media files in the Strapi CMS as an array.
After this patch, the response remains unchanged.
Adding to your install's plugins.js config file:
```  
upload: {
    config: {
      paginatedResponses: true,
    },
  },
```
Will result in a paginated response in the form of an object like other paginated responses from the /api endpoint.

### DIFF
--- a/packages/core/upload/server/src/controllers/content-api.ts
+++ b/packages/core/upload/server/src/controllers/content-api.ts
@@ -11,12 +11,20 @@ import { FileInfo, Config } from '../types';
 
 const { ValidationError } = utils.errors;
 
-type Pagination = {
+type PagePagination = {
   page: number;
   pageSize: number;
   pageCount: number;
   total: number;
 };
+
+type OffsetPagination = {
+  start: number;
+  limit: number;
+  total: number;
+};
+
+type Pagination = PagePagination | OffsetPagination;
 
 type PaginatedResponse<T> = {
   data: T[];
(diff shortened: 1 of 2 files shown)


Benjamin-K:
In the docs, pagination is added with a pagination object, see:
[Sort and Pagination](https://docs.strapi.io/cms/api/rest/sort-pagination#pagination)

Wouldn't it make sense to use the same parameters here as well? 

And I doubt that there are other endpoints, that do not (yet) have a paginated response, so I would prefer to have a paginated response if the pagination parameter is set, instead of using a config variable, that sounds like it would update all responses (but we miss some here maybe). WDYT?

EDIT: It would also be perfect to have some additional information, when using paginated responses, like `meta.pagination` does in the other endpoints.

colinduwe:
The current behavior is to return all the media items in the library in an array. The config flag is just to preserve the current behavior so we don't break existing deployments.  

I just noticed that my first attempt paginated response doesn't look like the standard response. I've added a transform function so the response is in the correct shape.
```
{
  data: [...],
  meta: {
    pagination: {...}
  }
}
```

colinduwe:
I added logic to support the standard API params like ?pagination[pageSize]=2 and ?pagination[limit]=3. I did it by copy/pasting code from the content API but the result is a bit of a mess. Closing in favor of https://github.com/strapi/strapi/pull/25034

PR #24714, opened 2025-10-28
fix: add script to configure husky on setup

Adzouz:
### What does it do?

Enable husky pre-commit hooks after running `yarn install`.

### Why is it needed?

Pre-commit hooks aren't currently working and they lead to a lot of avoidable errors in pipelines.

### How to test it?

- Run `yarn install`
-> Husky should correctly be configured

To test the hooks:
- Create a file
- Try to commit it with an invalid message
-> The commit should be blocked

🚀


scripts/setup-husky.js
remidej, line 1: Can we not rely on the `husky install` command? the one that's currently in the prepare package.json script

Adzouz: I thought the issue was that yarn prepare wasn't working but husky install is actually working 🤔

remidej: My understanding is that `prepare` isn't running, but there's no indication that `husky install` wouldn't work if it was properly triggered

Adzouz: Mmh I tested it and it was working well on my side but if so, what solution do you propose to install it?

HichamELBSI: You can remove this script. According to the Yarn, the only lifecycle script are `prepack` and `postpack`. Prepare is no longer supported in yarn 2.

This is what you only need to make it work:
in package.json, replace the prepare script by this
```json
"postinstall": "husky",
"prepack": "pinst --disable",
"postpack": "pinst --enable",
```

HichamELBSI (changes requested):

HichamELBSI (approved):

package.json
```diff
@@ -47,7 +47,9 @@
     "lint": "nx run-many --target=lint --nx-ignore-cycles && yarn lint:other",
     "lint:fix": "nx run-many --target=lint --nx-ignore-cycles -- --fix",
     "lint:other": "npm run prettier:other -- --check",
-    "prepare": "husky install",
+    "postinstall": "husky",
+    "prepack": "pinst --disable",
+    "postpack": "pinst --enable",
```
unrevised6419, line 52: Question: Where this `pinst` comes from? 🤔 
From what I see `yarn@1` [has it](https://classic.yarnpkg.com/en/package/pinst), but not `>yarn@1`

PR #24716, opened 2025-10-28
fix(i18n): use query engine to avoid marking published as modified

markkaylor:
### What does it do?

- Use query engine for update to avoid marking published as modified after ai translation
- Move completed job update to after the actual document update
- Use promise.all instead of promise.allSettled

### Why is it needed?

### How to test it?

Publish the source document
It should translate
It should retain the publish badge and not change to modified after translation

### Related issue(s)/PR(s)

resolves https://github.com/strapi/content-squad/issues/111

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -216,18 +216,9 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
         });
 
         throw new Error(`AI Localizations request failed: ${response.statusText}`);
-      } else {
-        await aiLocalizationJobsService.upsertJobForDocument({
-          documentId,
-          contentType: model,
-          sourceLocale: document.locale,
-          targetLocales,
-          status: 'completed',
-        });
       }
 
       const aiResult = await response.json();
-
       // Get all media field names dynamically from the schema
       const mediaFields = Object.entries(schema.attributes)
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
@@ -235,7 +226,7 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
         .map(([key]) => key);
 
       try {
-        await Promise.allSettled(
+        await Promise.all(
           aiResult.localizations.map(async (localization: any) => {
             const { content, locale } = localization;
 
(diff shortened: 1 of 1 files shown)


remidej:
Code LGTM, will test in a bit

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -247,7 +238,7 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
             });
 
             // Merge AI content and media fields, works only on first level media fields (root level)
-            const mergedData = { ...content };
+            const mergedData = structuredClone(content);
```
markkaylor, line 241: I don't think this makes a difference. Can revert if y'all prefer. I had tried this because I had some issues translating blocks on existing entries. 

The blocks returned from AI were transformed into an array 🤷‍♂️ . I was confused how changing to the query engine could have caused that issue. But I'm no longer able to reproduce, it appears to be working. 

Here's what it looked:

```
BEFORE ANY CHANGe

 {
  mergedData: {
    title: 'bonjour',
    blocksContent: [
      '{"type": "paragraph", "children": [{"type": "text", "text": "testeur"}]}'
    ],
    markdownContent: null,
    slug: null,
    cover: null
  }
}

AFTER CHANGE

{
    title: 'bonjour',
    blocksContent: [
      'type',
      'paragraph',
      'children',
      'type',
      'text',
      'text',
      'testeurs'
    ],
    markdownContent: null,
    slug: null,
    cover: null
  }
}
```

remidej: Those blocks issue look a lot like what the AI server was returning for blocks before the no-schema fix. Are you sure the issue is still relevant?

I'm fine with structuredClone though 👍

markkaylor: I've tried to reproduce it several times now but all seems ok so I think it's fine but if y'all can test too to be sure that'd be great. It's possible I wasn't using my local ai-server this morning when I hit this error.

araksyagevorgyan: I agree with Remi, the issue is exactly the one we had with ai service before Remi's latest changes. Seems it was using previous version of ai service.

araksyagevorgyan:
I tested it locally: after clicking 'Publish', the Published badge is preserved and the button gets disabled. Looks good to me 👍

araksyagevorgyan (approved):

PR #24718, opened 2025-10-28
fix: don't trigger translations with no translateable content

remidej:
### What does it do?

- avoids sending a request to the ai server with no content, which would throw a 400
- don't translate booleans

### How to test it?

Create or update a content type that is localized but that has no localized field. You should see no success or error notification, but a log in the terminal instead

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -11,7 +11,7 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
   const aiServerUrl = process.env.STRAPI_AI_URL || 'https://strapi-ai.apps.strapi.io';
   const aiLocalizationJobsService = getService('ai-localization-jobs');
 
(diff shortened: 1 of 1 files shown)


markkaylor:
> @markkaylor can you check in your strapi logs if it contacted the AI server?

Looks like it does:

```
[2025-10-28 18:02:33.991] info: AI token generated successfully
[2025-10-28 18:02:33.991] http: Contacting AI Server for localizations generation
[2025-10-28 18:02:34.644] http: GET /i18n/ai-localization-jobs/collection-types/api::testlocalalized.testlocalalized/zg2ecxsiz97lv09lh46ita6l (10 ms) 200
[2025-10-28 18:02:34.869] info: [AI Localizations Job] Updated existing job for document zg2ecxsiz97lv09lh46ita6l with status: completed
```

remidej:
Yeah I misread your comment, it makes sense. I'll see if it's a quick fix

remidej:
@markkaylor it's not a super quick fix (what if there was content before in that field?) so I'll focus on the rate limiting thing before the release

markkaylor:
Sounds good I think what you've done here is already a win so we can merge and circle back to improve if you want

PR #24720, opened 2025-10-28
enhancement: set up ai token in memory cache

remidej:
### What does it do?

sets up an in memory cache for the AI tokens. Useful for media lib and i18n, 2 features where this service can be called repeatedly
### Why is it needed?

to avoid getting rate limited by the AI server

### DIFF
--- a/packages/core/admin/server/src/services/user.ts
+++ b/packages/core/admin/server/src/services/user.ts
@@ -418,6 +418,19 @@ const getLanguagesInUse = async (): Promise<string[]> => {
 };
 
 /**
+ * In-memory cache for AI tokens
+ * Key format: `${projectId}:${userId}`
+ */
+const aiTokenCache = new Map<
+  string,
+  {
+    token: string;
+    expiresAt?: string;
+    expiresAtMs?: number;
+  }
+>();
+
+/**
  * Generate an AI token for the user performing the request
  */
 const getAiToken = async (): Promise<{ token: string; expiresAt?: string }> => {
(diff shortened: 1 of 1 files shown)


packages/core/admin/server/src/services/user.ts
```diff
@@ -474,6 +487,28 @@ const getAiToken = async (): Promise<{ token: string; expiresAt?: string }> => {
...
+  const cacheKey = `${projectId}:${userIdentifier}`;
+  const cachedToken = aiTokenCache.get(cacheKey);
+
+  if (cachedToken) {
+    const now = Date.now();
+    // Check if token is still valid (with buffer so it has time to  to be used)
+    const bufferMs = 2 * 60 * 1000; // 2 minutes
+
+    if (cachedToken.expiresAtMs && cachedToken.expiresAtMs - bufferMs > now) {
+      strapi.log.info('Using cached AI token');
```
markkaylor, line 500: Do we need this log? I don't really mind but I feel like we've been adding a lot of logs lately 😄

markkaylor (commented):
Looks good to me, tested and all seems fine 👍

packages/core/admin/server/src/services/user.ts
remidej: I fugured it could be useful to investigate if this does fix the rate limiting issue 🤔 it should still be fewer logs than when it queries the AI server for the token

markkaylor (approved):

packages/core/admin/server/src/services/user.ts
markkaylor: Sounds good sorry I thought I already approved 🚀

PR #24721, opened 2025-10-28
fix: use correct value for counting

YoannDelpierre:
### What does it do?

Fix a wrong display

### Why is it needed?

We don't have the number for items

image

### How to test it?

Switch to French

### Related issue(s)/PR(s)


HichamELBSI:
Thanks for the contribution @YoannDelpierre

PR #24725, opened 2025-10-29
fix: change messsaging on i18n AI

oiorain:
### What does it do?

Change wording on AI i18n explanation in the CM  popup.

### Why is it needed?

Clearer messaging

### How to test it?

Check the text in this popup has been updated: 
Screenshot 2025-10-29 at 11 31 51

1. enable i18n AI
2. Go to a Content type
3. Hover on the sparkle AI button: the new text should be  'Our AI translates content in all locales each time you save a modification in the default locale.'


oiorain:
No worries at all! thank you for taking care of it :) 
and thank you for the explanations! I will gladly do that. 
I've been wanting to take care of a batch of PRs about translations wqhen i have time. Maybe you could give me a hand with that if you have time!

PR #24726, opened 2025-10-29
fix(i18n): creating a source document does not create translated docu…

markkaylor:
### What does it do?

- Create the derive document if it does not already exist
- Copy the non-localized content to the derived document on creation
- Updates a translation

### Why is it needed?

Translated documents should be created when the default locale creates a document

### How to test it?

Go to a content type with i18n
Create a new doc in the default locale
It should create and translate for other locales
It should copy over non localized attributes

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -254,16 +254,10 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
               }
             }
 
-            /**
-             * TODO:
-             * We should use the document service here, however currently,
-             * updating a localized entry marks the source document as modified.
-             * @see https://github.com/strapi/strapi/issues/22924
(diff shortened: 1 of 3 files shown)


packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -254,18 +254,30 @@ const createAILocalizationsService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+              i18nService.fillNonLocalizedAttributes(mergedData, document, {
+                model,
+              });
+              await strapi.documents(model).create({
+                data: { documentId, ...mergedData },
+                locale,
+              });
+            } else {
+              /**
+               * TODO:
```
markkaylor: I think this fixes the issues for our use case. We should definitely prioritize fixing this though after the release. I think the fix might actually be in the doc service. Don't have the time to debug that and a bit nervous to make any changes at that level just before a release.

PR #24728, opened 2025-10-29
feat(core): add content API audit logging

sourabhsinghs:
### What does it do?

- Introduces a `content-audit-logs` provider that registers the audit-log model, service, and `/audit-logs` API module.
- Adds a controller/routes layer to expose paginated audit history secured by the `read_audit_logs` permission.
- Updates core collection/single-type controllers to log mutations and documents setup/testing (including Yarn `portal:` linking).

### Why is it needed?

- Provides end-to-end visibility on Content API create/update/delete operations.
- Pre-seeds permission rows so administrators can grant audit-log access per role while keeping it disabled by default.

### How to test it?

1. Rebuild the local packages:  
   `yarn nx run @strapi/core:build --skip-nx-cache && yarn nx run @strapi/strapi:build --skip-nx-cache`
2. In a sibling app, apply the portal dependency overrides from `AUDIT_LOG_README.md`, run `yarn install`, and restart `yarn develop`.
3. Perform content mutations (create/update/delete) via the Content API, then call `GET /api/audit-logs` with a token holding `read_audit_logs`. The response should list the captured entries with payload/diff data.


oiorain:
Hi @sourabhsinghs 

Thank you for your contribution. We're getting a lot of audit logs proposition these days. Not sure why. 
We already have an Audit log feature so I will be closing this PR. 
Have a lovely day.

PR #24739, opened 2025-10-30
fix: prevent admins from deleting themselves

### LINKED ISSUE #17854: Deleting your own admin user from SSO yields a 500 error

```
Environment: development
OS: linux-x64
Strapi Version: 4.12.7
Node/Yarn Version: yarn/1.22.19 npm/? node/v18.12.1 linux x64
Edition: Enterprise
Database: sqlite
```
- Is your project Javascript or Typescript: JAvascript

### Describe the bug

When a user signs in via SSO (during testing used GitHub) and deletes themselves they do not get redirected to the login page but instead an internal server error is thrown:

```
[2023-08-29 14:54:12.543] http: POST /admin/users/batch-delete (70 ms) 200
/home/dmehaffy/Development/testingBugs/node_modules/@strapi/utils/dist/validators.js:86
    throw new errors_1.YupValidationError(error, errorMessage);
          ^

YupValidationError: 1 relation(s) of type admin::user associated with this entity do not exist
    at handleYupError (/home/dmehaffy/Development/testingBugs/node_modules/@strapi/utils/dist/validators.js:86:11)
    at /home/dmehaffy/Development/testingBugs/node_modules/@strapi/utils/dist/validators.js:98:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.create (/home/dmehaffy/Development/testingBugs/node_modules/@strapi/strapi/lib/services/entity-service/index.js:156:23)
    at async Object.<anonymous> (/home/dmehaffy/Development/testingBugs/node_modules/@strapi/strapi/lib/services/entity-service/index.js:415:20)
    at async Object.saveEvent (/home/dmehaffy/Development/testingBugs/node_modules/@strapi/provider-audit-logs-local/dist/index.js:18:17) {
  details: {
    errors: [
      {
        path: [],
        message: '1 relation(s) of type admin::user associated with this entity do not exist',
        name: 'ValidationError'
      }
    ]
  }
}
```

This leads to a hard crash of the server requiring a restart to bring the service back online

1. Setup Strapi Project with SSO (doesn't matter which one)
2. Sign in Via SSO
3. Try to delete your own user account from within settings

### Expected behavior

User should be deleted and redirected back to login page, as is what happens with non SSO based users

### Screenshots

image
image
image
image

### Code snippets

N/A

### Additional context

Internal TID4568

Bassel17:
### What does it do?

- Add checks to see if the user is trying to delete themselves and throw an Error
- remove the delete icon/button from the admin user in the list, to prevent them from attempting to delete themselves

### Why is it needed?

- the resulting behavior from users deleting themselves is undesirable, it introduces the possibility of them locking themselves out.
- check #17854 

### How to test it?

- Attempt to delete yourself, you shouldn't be able to
- successful run of automated tests

### Related issue(s)/PR(s)

fixes #17854

### DIFF
--- a/tests/api/core/admin/admin-user.test.api.js
+++ b/tests/api/core/admin/admin-user.test.api.js
@@ -327,7 +327,7 @@ describe('Admin User CRUD (api)', () => {
       data: null,
       error: {
         details: {},
-        message: 'You cannot delete your own user.',
+        message: 'You cannot delete your own user',
         name: 'ApplicationError',
         status: 400,
       },


packages/core/admin/server/src/controllers/__tests__/user.test.ts
```diff
@@ -323,4 +323,215 @@ describe('User Controller', () => {
...
+      } as any;
+
+      await userController.deleteMany(ctx);
+
+      expect(deleteByIds).toHaveBeenCalledWith([1, 2]);
+      expect(sanitizeUser).toHaveBeenCalledTimes(2);
+      expect(deleted).toHaveBeenCalledWith({ data: users });
+    });
+
+    test('Prevents user from deleting themselves in batch delete', async () => {
```
innerdvations, line 461: Just a thought, do we want to allow the rest to succeed and only fail the delete on the user? It would depend on if we do that with other methods but I don't know if we have consistency here.

Bassel17: yep indeed I thought about that too, it's just that I think it would be nice to have less magic stuff going on and more communication especially since it's the admin api if we submit something wrong it would be better to communicate back rather than proceed silently

packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
```diff
@@ -200,7 +201,7 @@ const ListPageCE = () => {
                             <Pencil />
                           </IconButton>
                         ) : null}
-                        {canDelete ? (
+                        {canDelete && currentUser?.id !== user.id ? (
```
jhoward1994, line 204: Would it make sense to make this check part of canDelete ? I could go either way on that as I see how it's kind of an additional check outside of rbac . wdyt?

packages/core/admin/server/src/controllers/__tests__/user.test.ts
innerdvations: Ok! Since this is Admin API it's fine either way, it won't be a breaking change

innerdvations:
@derrickmehaffy just curious if you can think of any unusual things involving SSO managed users that this could potentially conflict with. I think this was actually your suggestion, but just wanted to check before merging in case you see an edge case we might have missed.

derrickmehaffy:
Nothing that comes to mind 🤔

packages/core/admin/admin/src/pages/Settings/pages/Users/ListPage.tsx
Bassel17: yeah I think it should stay separate for the same reason you mentioned, canDelete is also used above do it makes sense to leave it as a separate check just for this case

innerdvations (approved):

PR #24743, opened 2025-10-30
fix: remove await from several telemetry calls to prevent blocking

derrickmehaffy:
### What does it do?

Makes telemetry calls not wait for a response to prevent blocking the requests if the analytics server is slow

### Why is it needed?

No reason to await this, we don't do anything with the response and awaiting them just slows down the application

### How to test it?

N/A

### Related issue(s)/PR(s)

Reported internally via Slack


packages/plugins/i18n/server/src/services/metrics.ts
```diff
@@ -10,13 +10,13 @@ const sendDidInitializeEvent = async () => {
...
 
-  await strapi.telemetry.send('didInitializeI18n', { groupProperties: { numberOfContentTypes } });
+  strapi.telemetry.send('didInitializeI18n', { groupProperties: { numberOfContentTypes } });
 };
 
 const sendDidUpdateI18nLocalesEvent = async () => {
   const numberOfLocales = await getService('locales').count();
 
-  await strapi.telemetry.send('didUpdateI18nLocales', {
+  strapi.telemetry.send('didUpdateI18nLocales', {
```
innerdvations: Is there an await on any of the parent functions we need to remove? like `await sendDidUpdateI18nLocalesEvent`?

packages/core/content-manager/server/src/services/metrics.ts
```diff
@@ -29,7 +29,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
     }
 
     try {
-      await strapi.telemetry.send('didConfigureListView', data);
+      strapi.telemetry.send('didConfigureListView', data);
```
ChristopheCVB: `.catch(()=>{})` for brevety?

packages/core/content-type-builder/server/src/controllers/content-types.ts
```diff
@@ -77,17 +77,17 @@ export default {
       };
 
       if (_.isEmpty(strapi.apis)) {
-        await strapi.telemetry.send('didCreateFirstContentType', metricsPayload);
+        strapi.telemetry.send('didCreateFirstContentType', metricsPayload);
```
ChristopheCVB: don't we want to also `.catch(()=>{})` these?

packages/core/content-manager/server/src/services/metrics.ts
innerdvations: Yeah, that will actually be necessary to avoid throwing errors asynchronously, since the existing try/catch won't work anymore.

`void strapi.telemetry.send(event, data).catch(() => {});`

or if we still want to safely see errors it could still log something in the catch. We should probably (later, not necessarily here) modify the `send` itself or create a new version of it to abstract all of that away and make it easier to use in cases where we don't care about the result (which is 99% of them)

packages/core/content-type-builder/server/src/controllers/content-types.ts
```diff
@@ -82,19 +82,21 @@ export default {
       };
 
       if (_.isEmpty(strapi.apis)) {
-        await strapi.telemetry.send('didCreateFirstContentType', metricsPayload);
+        strapi.telemetry.send('didCreateFirstContentType', metricsPayload).catch(() => {});
```
innerdvations: I believe all the awaits in this file were on purpose to avoid getting interrupted by the server restart that follows immediately after. I think at best we could switch this to defer the restart until a given timeout.

innerdvations (changes requested):

innerdvations:
I pushed what I meant with defer; if that works for you, we can go with that, otherwise go ahead and revert to the `await`

innerdvations (approved):

Issue #21377, opened 2024-09-19
Pressing 'return' on localized text input prompts to replace from another local


innerdvations:
### Required System information

$ strapi report
Launched In: 240 ms
Environment: development
OS: darwin-arm64
Strapi Version: 5.0.0
Node/Yarn Version: yarn/1.22.22 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite

### Describe the bug

Pressing "return" on a new localized entity text field asks to fill it in with a different locale

### Steps to reproduce the behavior

- Have a 'fr' and 'en' locale
- Have a content type 'article' with a text input called 'name'
- Create an 'en' version of the article
- Switch to 'fr' locale
- Type in a name
- Press return
- Strapi asks to fill in the name value from another locale

### Expected behavior

- either a save, publish, or for nothing happen at all

### Screenshots

![image](https://github.com/user-attachments/assets/14554fa8-e06c-4251-a15f-266f38302a28)

### Code snippets

If applicable, add code samples to help explain your problem.

oiorain:
I confirm this is still happening in 5.27.0.

PR #24744, opened 2025-10-30
fix: enter key trigger in edit view form

Adzouz:
### What does it do?

It fixes the action when hitting the Enter key when the focus is inside a field of the edit page view in the content manager.
Pressing Enter in a field (except blocks fields) will perform a Save (default behaviour from a form).
Pressing Ctrl + Enter will perform a Publish (for Draft and publish entries).

### Why is it needed?

It was having a weird behaviour. Hitting the enter key would trigger the first button action of the form (when internationalization enabled, it was opening the "Fill in from another locale" modal for example).

### How to test it?

- Go to the content manager and select a content type with Internationalization enabled
- Add a new entry
- Publish
- Switch to another locale
- Fill a field and hit enter key
-> The "Fill in from another locale" shouldn't be triggered and the form should be submitted (Save action).

To test the Ctrl + Enter:
- Go to the content manager and select any content type with Draft and publish enabled
- Add a new entry
- Fill a field and hit enter key
- Press Cmd / Ctrl + Enter
-> The entry should be published.

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/21377

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -744,7 +744,8 @@ const PublishAction: DocumentActionComponent = ({
       meta?.availableStatus.some((doc) => doc[PUBLISHED_AT_ATTRIBUTE_NAME] !== null)) &&
     document?.status !== 'modified';
 
-  const performPublish = React.useCallback(async () => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const performPublish = async () => {
     setSubmitting(true);
 
     try {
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -868,14 +892,74 @@ const PublishAction: DocumentActionComponent = ({
         onPreview();
       }
     }
-  };
+  }, [
+    setSubmitting,
```
Adzouz: I know this list is scary but I didn't have the choice and that's already how it was done with the performSave method if you're curious to know more

HichamELBSI: I don't really understand when you say that you don't have the choice, sorry. Why using a `useCallback` here?

HichamELBSI: ⚠️ The questions you have to ask yourself regarding these optimization react features useCallback, useMemo, memo:
- What problem it solve in this case?
- Do I have performance issues?
- What are the gains of using a useCallback compared to no using it at all?

The first principles of a useCallback is:
- compare all the dependencies
- If atv least one of them change -> recreate the function

There are more than 30 dependencies here 😂 some of them are unstable (formValues, initialValues, parentDocumentData, documentHistory, etc) meaning that they will be updated many times. I doubt that none of these deps changes every render 

⬆️ This means that your useCallback is completely useless because the function will be created again and again.

Even on a usecase where useCallback is working, it has a cost (useCallback isn't just one line of code, it's implementation can be heavy). The cost of comparing dependencies, etc. But in your case, the cost of this useCallback is very heavy and is useless because the function will be recreated for most of the rerenders.

Let me know if it is not clear. I will share a reminder about it with all the engineers.

Adzouz:
@HichamELBSI I know and I tried to get rid of it because I also found it ridiculous and completely unnecessary but I got this:
Screenshot 2025-11-21 at 15 13 02
What should I do about that?

HichamELBSI:
> @HichamELBSI I know and I tried to get rid of it because I also found it ridiculous and completely unnecessary but I got this: Screenshot 2025-11-21 at 15 13 02 What should I do about that?

This is just the default eslint warning. This warning isn't smart, it is just design to ensure that all external values are included in the array. In this case, since the memoization doesn't make sense has more disadvantages, this warning should be disabled we should definitely not add complexity and useless code based on eslint. Add 
`// eslint-disable-next-line react-hooks/exhaustive-deps` before the warning.

Adzouz:
Done ✅

Issue #24737, opened 2025-10-30
Can't clear type filter in media library


Rowan-Paul:
### Node Version

22.20.0

### NPM/Yarn/PNPM Version

NPM 10.9.3

### Strapi Version

5.30.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When filtering by media type = file, the filter cannot be removed, unlike other types such as audio.

### Steps to Reproduce

1. Go to media library, filter on type = file
2. Try to remove the filter

### Expected Behavior

The filter goes away, just like with other filters when I click the x

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/d1727550-8783-4a74-bc2a-9fd2656a6058

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Adzouz:
Hi there! Thanks a lot for reporting the issue 🙏
I just merged @kaigritun PR (thanks for the contribution), fixing the problem. It will be part of the release of Strapi on Wednesday April 1st 🚀

PR #24748, opened 2025-10-31
fix: Media Library Filter Removal for Complex Filters

dagadevelop:
# Fix: Media Library Filter Removal for Complex Filters

## What does it do?

Fixes a bug where complex filters (like type "file") could not be removed by clicking the X button on the filter tag in the Media Library.

The fix adds proper comparison logic for complex filter structures (objects and arrays) in addition to the existing string comparison.

**Affected filters:**

- Type = file (uses complex nested filter structure)
- Any future filters with non-string values

### The Problem

When users applied the "type = file" filter in the Media Library, they couldn't remove it by clicking the X icon on the filter tag. The filter would remain stuck.

**Root cause:** The filter removal logic in `FilterList.tsx` only handled string values. When the filter value was an object or array, it would always return `true` (keep the filter) instead of comparing and removing it.

**Type = file filter structure:**

```json
{
  "mime": {
    "$not": {
      "$contains": ["image", "video"]
    }
  }
}
```

This complex structure wasn't being compared correctly, so it could never be removed.

## How to test it?

1. Navigate to Media Library
2. Click the "Filters" button
3. Select filter: **Type** = **file**
4. Click "Add filter"
5. **Click the X icon** on the filter tag
6. Filter should be removed and the media library should refresh

## Related issue(s)/PR(s)

Fix [#24737](https://github.com/strapi/strapi/issues/24737)

### DIFF
--- a/packages/core/upload/admin/src/components/FilterList/FilterList.tsx
+++ b/packages/core/upload/admin/src/components/FilterList/FilterList.tsx
@@ -54,10 +54,15 @@ export const FilterList = ({ appliedFilters, filtersSchema, onRemoveFilter }: Fi
       if (filterName !== undefined) {
         const filterType = Object.keys(filterName)[0];
         const filterValue = filterName[filterType];
+        // Handle string values
         if (typeof filterValue === 'string') {
           const decodedValue = decodeURIComponent(filterValue);
           return prevFilter[name]?.[filterType] !== decodedValue;
         }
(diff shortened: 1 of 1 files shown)


Adzouz:
Hey @dagadevelop,
Thanks a lot for the contribution, we really appreciate it 🙏
Unfortunately I'll have to close this PR since the original issue was fixed by this other PR: https://github.com/strapi/strapi/pull/25399
Have a great day 🚀

PR #24749, opened 2025-10-31
enhancement: add i18n ai loading indicator

remidej:
### What does it do?

shows visual indicator that ai localizations are processing

https://github.com/user-attachments/assets/86c1769f-d74f-45a2-882d-5784b3ed25e0

### DIFF
--- a/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
+++ b/packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
@@ -32,7 +32,7 @@ import {
   Box,
   Link,
 } from '@strapi/design-system';
-import { WarningCircle, ListPlus, Trash, Earth, Cross, Plus, Sparkle } from '@strapi/icons';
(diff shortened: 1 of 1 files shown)


packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -406,6 +406,19 @@ const AITranslationStatusIcon = styled(Status)<{ $isAISettingEnabled: boolean }>
   }
 `;
 
+const SpinningLoader = styled(Loader)`
```
mathildeleg, line 409: Instead of creating this Spinning animation here, couldn't the Loader from the DS be used? It's not quite the same design, but the small version should fix inside the Box ok
(That is the `Loader from '@strapi/design-system'` which already has the animation instead of `Loader from '@strapi/icons'`)

remidej: Just checked and we want to keep the purple color for this icon, which isn't supported with the animated loader, so I'll stick with Vincent's design

PR #24750, opened 2025-10-31
fix(upload): dispatches isAIMediaLibraryConfigured for projects witho…

markkaylor:
### What does it do?

When AI is available on the instance
- The aiMetadata setting should be true by default
- Send tracking events with property `isAIMediaLibraryConfigured`

When AI is not available in the instance
- Do not send tracking events with property isAIMediaLibraryConfigured

When an instance transitioned from a license with AI to a CE (or license without AI)
- Preserve the user’s AI setting so it is the same if and when they add the license back
- Do not send tracking events with property isAIMediaLibraryConfigured

### Why is it needed?

To track properly

### How to test it?

Clear your database (sorry) `rm -rf .tmp`
Start the app with a license containing AI
Go to the media lib and upload an image
It should use AI to generate captions and alt text
Alter one of the captions or alt text and hit finish
Look in the network tab for track and ensure it has the event property (`isAIMediaLibraryConfigured: true`)
Disable AI and do the same test
Look in the network tab for track and ensure it has the event property (`isAIMediaLibraryConfigured: false`)

Kill the server, remove the license
Do the same test but it won't have AI generation now
Look for any track event in the network tab
It should NOT have an event property `isAIMediaLibraryConfigured`

Kill the server, add the license back
Do the same test as before but it should have the event property and it should be whatever the last setting was you set

Kill the server, remove the license, clear the database
Do the same test but it won't have AI generation now
Look for any track event in the network tab
It should NOT have an event property `isAIMediaLibraryConfigured`

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/server/src/services/project-settings.ts
+++ b/packages/core/admin/server/src/services/project-settings.ts
@@ -196,17 +196,10 @@ const updateProjectSettings = async (
 const isAIAvailable = () => {
   // Check if user disabled AI features globally
   const isAIEnabled = strapi.config.get('admin.ai.enabled', true);
-  if (!isAIEnabled) {
-    return false;
-  }
-
   // Check if the user's license grants access to AI features
   const hasAccess = strapi.ee.features.isEnabled('cms-ai');
-  if (!hasAccess) {
-    return false;
-  }
 
-  return true;
+  return isAIEnabled && hasAccess;
 };
 
 export { deleteOldFiles, parseFilesData, getProjectSettings, updateProjectSettings, isAIAvailable };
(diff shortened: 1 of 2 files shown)


packages/core/admin/server/src/services/project-settings.ts
markkaylor, line 1: Idk if this is the best place to do this. I could create a dedicated service otherwise 🤷‍♂️ , but I think it would be nice to have this available everywhere

innerdvations:
@Bassel17 Did you have something for this somewhere else, or am I just thinking of `useAiAvailability` ?

markkaylor:
Sorry moving this back to draft for now I think I have a better idea to handle the tracking for both ai features

markkaylor:
Closing in favor of another approach, PR coming soon

Issue #24674, opened 2025-10-24
Hidden components in content type builder accordions


kulbon:
### Node Version

22.17.1

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.29.0

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

Not all components are visible in content type builder sidebar. Each accordion has set max-height: 1000px and overflow: hidden

<img width="1907" height="867" alt="Image" src="https://github.com/user-attachments/assets/5e6d6531-f26a-4512-bff6-edef9e6b0347" />

<img width="1902" height="854" alt="Image" src="https://github.com/user-attachments/assets/e25813dc-bca0-41b8-a8a0-4ba789f98cab" />

without max-height it works

<img width="1913" height="872" alt="Image" src="https://github.com/user-attachments/assets/1f23daaa-1b85-47a3-8a73-0ecb19f3b7aa" />

### Steps to Reproduce

Create as many components as possible in one category to make the accordion height exceed 1000 px and check that not all components are visible

### Expected Behavior

All components should be visible

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

vidush2025:
Please lemme know if theres some other approach I should follow, I would really love to contribute on this issue as one of my good-first-issue(s)

Thank You.

Adzouz:
Hi there @kulbon and thank you for reporting this issue (really annoying I admit). The fix was just merged (thanks @plasticfiresam for the fix 🙏). It will be deployed in the next release on Wednesday December 24th 🚀 Closing this issue, have a great day ☀️

PR #24679, opened 2025-10-24
fix: maxHeight of Content-Type-Builder fixed

vidush2025:
increased the maxHeight of the SubSection in the components directory for Content-Type Builder Sidebar, so that it now clearly shows all the subsections properly

WhatsApp Image 2025-10-25 at 01 22 22

### What does it do?
-> We have increased the max height of Flex in SubNav.tsx file from 1000px to 9999px so that it now clearly shows all of the components instead of hiding them to fit into the max size.

Describe the technical changes you did.
-> Just a small fix in the max height of the SubSection from 1000px to 9999px. This now renders all of the components properly.

### Why is it needed?
-> It is needed because, initially when there were too many components added, the section would automatically hide them and sometimes just render the category with a ChevronDown icon. If the max height was even smaller (say 99px), it would only show the categories and even clicking on the ChevronDown button would not result in any visible components. 
Describe the issue you are solving.

### How to test it?
-> Just go to the plugins/content-type-builder page and add many components.

Provide information about the environment and the path to verify the behaviour.
-> admin/plugins/content-type-builder/component-categories/basic

### Related issue(s)/PR(s)
Issue No: ##24674

Issue #23918, opened 2025-07-08
Bug Report: Sidebar in Content-Type Builder UI cuts off component/collection types


sehrish30:
### Node Version

20.11.1

### NPM/Yarn/PNPM Version

10.9.0

### Strapi Version

5.17.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Bug Description

The left sidebar in the Content-Type Builder (Strapi Admin Panel) cuts off component or collection types if there are too many items, and scrolling does not appear, making it impossible to access lower entries.

### Steps to Reproduce

1. Go to the Content-Type Builder in the Admin Panel.
2. Add many collection types or components (more than the sidebar can fit vertically).
3. Observe that the list gets cut off with no scrollbar.
4. Attempting to scroll does not reveal all entries.

### Expected Behavior

The sidebar should scroll properly and show all collection types/components regardless of screen size or number of entries.

### Logs

```shell

```

### Media

<img width="150" height="433" alt="Image" src="https://github.com/user-attachments/assets/613a20f6-3db5-4e63-8d39-2cdcd9608bb9" />

https://github.com/user-attachments/assets/52931b65-24e6-44d0-a653-5f21a710d3b9

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

sehrish30:
still the bug lingers in 5.28.0

Adzouz:
Hi there @sehrish30 and thank you for reporting this issue. It's a duplicate of https://github.com/strapi/strapi/issues/24674 and the fix was just merged (thanks @plasticfiresam 🙏). It will be deployed in the next release on Wednesday December 24th 🚀 Closing this issue, have a great day ☀️

PR #24753, opened 2025-10-31
fix: fixed subnav`s height when there are more elements than for 1000px

plasticfiresam:
### What does it do?

It removes hardcoded "magic number" that setting height of SubNav`s lists and makes it calculated by children height when UI is changing.

### Why is it needed?

When there are a lot of components created in single category in Content Type Builder, they appears hidden because height of block contains them is hardcoded to certain "magic number" and 'overflow' CSS property is set to 'hidden'. Overflows behavior is OK to be consistent in UI, but magic number is limiting space for big count of components to be seen.

### How to test it?

You can add many components (for example against "magic number" you should have >30 components) in Content Type Builder and see that height of list dynamically changes to fit all the items.

### Related issue(s)/PR(s)

Resolves:
https://github.com/strapi/strapi/issues/24674
https://github.com/strapi/strapi/issues/23918

Offers more flexible solution that provided in this PR: https://github.com/strapi/strapi/pull/24679

### DIFF
--- a/packages/core/admin/admin/src/components/SubNav.tsx
+++ b/packages/core/admin/admin/src/components/SubNav.tsx
@@ -338,9 +338,7 @@ const SubSection = ({ label, children }: { label: string; children: React.ReactN
         style={{
           maxHeight: isOpen ? `${contentHeight}px` : 0,
           overflow: 'hidden',
-          transition: isOpen
-            ? 'max-height 1s ease-in-out'
-            : 'max-height 0.5s cubic-bezier(0, 1, 0, 1)',
+          transition: 'max-height 0.5s cubic-bezier(0, 1, 0, 1)',
         }}
       >
         {children.map((child, index) => {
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/components/SubNav.tsx
```diff
@@ -321,13 +329,14 @@ const SubSection = ({ label, children }: { label: string; children: React.ReactN
...
         id={listId}
         direction="column"
         gap="2px"
         alignItems={'stretch'}
         style={{
-          maxHeight: isOpen ? '1000px' : 0,
+          maxHeight: isOpen ? `${contentHeight}px` : 0,
           overflow: 'hidden',
           transition: isOpen
             ? 'max-height 1s ease-in-out'
```
Adzouz: I don't want to block the PR for that but I would use the same transition than the line after if possible `max-height 0.5s cubic-bezier(0, 1, 0, 1)`.

Issue #23603, opened 2025-05-27
Permissions - Users can't publish single types without having "update" permission


kareemgamal1:
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

10.5.2

### Strapi Version

5.5.0

### Operating System

Windows 11

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

Hello,
I’m facing an issue related to role-based permissions in Strapi.

A client needs a user role (e.g. "Publisher") that can read and publish content, but not update it. This setup works correctly for collection types, where the user can publish without needing update permissions.

However, for single types, when the "Update" permission is disabled, attempting to publish results in a 403 Forbidden error. As soon as the "Update" permission is enabled, publishing works as expected.

Publishing single types is currently tied to update permissions, which isn't consistent with how it works for collection types.

Thanks!

### Steps to Reproduce

1. Create a new role (e.g. "Publisher") with the following permissions for a single type:
- Read
- Publish
- (Do not enable Update)
2. As a user with this role, go to the single type entry.
3. Click the "Publish" button.
4. Observe that a "Forbidden" error appears.
5. Log in as a Super Admin and update the role to also allow the Update permission.
6. Log back in as the same user.
7. Open the same single type entry and click "Publish" again.
8. Notice that publishing now works as expected.

### Expected Behavior

The publisher role should be able to publish single types without needing to have the update permission

### Media

https://streamable.com/wjwdwe

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

kareemgamal1:
Hi @derrickmehaffy , just checking in, any updates on this issue?

Our client is currently blocked due to compliance with governmental regulations, and we’re unable to proceed without a resolution or workaround. We’d really appreciate any insights or support you can provide

Thanks in advance!

ahmadalfy:
I think the problem is in the [single-types controller](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/controllers/single-types.ts).

In the [`createOrUpdateDocument`](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/controllers/single-types.ts#L30) function, there's a permission check that requires both create AND update permissions to be present:

```typescript
[...]

PR #24754, opened 2025-11-01
fix: single type publish permission error

dagadevelop:
## What does it do?

Fixes the permission check in the single-type `publish` controller to allow users with only `publish` permission (without `update` or `create`) to publish existing drafts.

The controller now follows the same logic as `collection-types`:
- If the user has `update` permission on the document → updates then publishes
- If the user only has `publish` permission → publishes the existing draft without updating it
- If the document doesn't exist and user has `create` permission → creates then publishes

## Why is it needed?

Previously, users with `read` + `publish` permissions (but without `update` or `create`) would get a 403 Forbidden error when trying to publish a single type. This was caused by the `publish` method always calling `createOrUpdateDocument`, which requires either `update` or `create` permissions.

This bug prevented implementing proper editorial workflows where some users can only approve and publish content created by others, without being able to modify it.

## How to test it?

1. Create a role with only `read` and `publish` permissions for a single type (without `update` or `create`)
2. Create a user with that role
3. As an admin, create a draft for the single type
4. Log in as the restricted user
5. Navigate to the single type
6. Click the "Publish" button
7. Verify the document is published successfully without a 403 error

## Related issue(s)/PR(s)
Fix [#23603](https://github.com/strapi/strapi/issues/23603)

### DIFF
--- a/packages/core/content-manager/server/src/controllers/single-types.ts
+++ b/packages/core/content-manager/server/src/controllers/single-types.ts
@@ -209,17 +209,17 @@ export default {
       let document = await findDocument(sanitizedQuery, model, { locale, status: 'draft' });
 
       // If document exists and user can update it, update it before publishing
-      if (document && permissionChecker.can.update(document)) {
+      const shouldUpdate = document && permissionChecker.can.update(document);
+      // If document doesn't exist and user can create it, create it before publishing
+      const shouldCreate = !document && permissionChecker.can.create();
+      if (shouldUpdate || shouldCreate) {
         document = await createOrUpdateDocument(ctx, { populate });
       } else if (!document) {
-        // Document doesn't exist, try to create it
-        if (permissionChecker.cannot.create()) {
-          throw new errors.ForbiddenError();
-        }
-        document = await createOrUpdateDocument(ctx, { populate });
+        // Document doesn't exist and user can't create it
+        throw new errors.ForbiddenError();
       }
-      // Else: document exists but user can't update it, just publish the existing draft
 
+      // If document doesn't exist, throw an error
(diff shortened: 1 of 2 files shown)


fatihtelis:
Hi, 
I'm waiting for this PR for some time to be included in the main package. :/
Thanks.

packages/core/content-manager/server/src/controllers/single-types.ts
```diff
@@ -203,7 +203,22 @@ export default {
...
       const sanitizedQuery = await permissionChecker.sanitizedQuery.publish(query);
       const populate = await buildPopulateFromQuery(sanitizedQuery, model);
-      const document = await createOrUpdateDocument(ctx, { populate });
+      const { locale } = await getDocumentLocaleAndStatus(body, model);
+
+      // Find the existing document
+      let document = await findDocument(sanitizedQuery, model, { locale, status: 'draft' });
+
+      // If document exists and user can update it, update it before publishing
+      if (document && permissionChecker.can.update(document)) {
```
Adzouz: You could simplify and calling the method only once like:
```suggestion
      const shouldUpdate = document && permissionChecker.can.update(document);
      const shouldCreate = !document && permissionChecker.can.create();

      if (shouldUpdate || shouldCreate) {
        document = await createOrUpdateDocument(ctx, { populate });
      } else if (!document) {
        // Document doesn't exist and user can't create it
        throw new errors.ForbiddenError();
      }
```

Adzouz:
Hi @dagadevelop,
I took the liberty to take over the PR to merge it 😊 Thanks again for the contribution 🙏

PR #24758, opened 2025-11-03
chore: update ai tracking events

markkaylor:
### What does it do?

- Create an EE api to fetch the ai settings. It is mostly needed so we can get the i18n config setting in the content manage in order to send it along with the tracking event. I chose to do it in EE because I saw other AI related code there but I also see some of the same code in admin (ie `getAIToken`), I'm confused about that but I think this is still the right place? Happy to move it to admin though if anyone disagrees.
- Only send the `isAIXConfigured` event property if ai is available on the instance. 
- Always send the `isAIXConfigured` event property if ai is available with the value true or false

### Why is it needed?

- For i18n the tracking is not sent from the right place causing duplicate events for publish. 
- When a license is added with AI and then the license is remove or AI is removed the last config is still sent to the tracking 

### How to test it?

Make sure you have a growth license with AI

### i18n
By default i18n should be disabled. Perform a save and publish action then check the tracking events. They should have an event property. `isAIi18nConfigured: false`.

Check the `track` network request to confirm:
`didPublishEntry`
`didEditEntry`

Enable AI for i18n and do the same test. You should now see `isAIi18nConfigured: true`.

https://github.com/user-attachments/assets/df3f77b1-9da9-4ec9-8211-16ae2e1f7052

Now remove you license, do the same test, you should not see the property sent with the tracking event at all. If you add the license back it should have your previous setting still configured and continue sending tracking events.

### Media library
By default the media library is enabled. _All_ tracking events in the media library should be sending an event property `isAIMediaLibraryConfigured`. 

Upload an image and change an alt text and caption generated by AI. Check these tracking events for AI. They should have an event property. `isAIMediaLibraryConfigured: true`

`didEditAICaption`
`didEditAlAlternativeText`

Disable AI for Media Library. You should now see the tracking event property with `isAIMediaLibraryConfigured: false`

(Note: in the video I first check willNavigate that doesn't have the property, it is not sent by the upload plugin so that is fine)

https://github.com/user-attachments/assets/561bc01d-d537-4e8d-b738-d277190e68a4

Now remove you license, do the same test, you should not see the property sent with the tracking event at all. If you add the license back it should have your previous setting still configured and continue sending tracking events.

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/features/Tracking.tsx
+++ b/packages/core/admin/admin/src/features/Tracking.tsx
@@ -50,7 +50,6 @@ const TrackingProvider = ({ children }: TrackingProviderProps) => {
   const { data } = useTelemetryPropertiesQuery(undefined, {
     skip: !initData?.uuid || !token,
   });
-
   React.useEffect(() => {
     if (uuid && data) {
       const event = 'didInitializeAdministration';
@@ -538,7 +537,6 @@ const useTracking = (): UseTrackingReturn => {
   const deviceType = useDeviceType();
   const { uuid, telemetryProperties } = React.useContext(TrackingContext);
   const userId = useAppInfo('useTracking', (state) => state.userId);
-
   const trackUsage = React.useCallback(
     async <TEvent extends TrackingEvent>(
       event: TEvent['name'],
@@ -551,9 +549,7 @@ const useTracking = (): UseTrackingReturn => {
             {
               event,
               userId,
-              eventProperties: {
-                ...properties,
-              },
+              eventProperties: { ...properties },
               userProperties: {
                 deviceType,
               },
@@ -580,7 +576,7 @@ const useTracking = (): UseTrackingReturn => {
 
       return null;
     },
-    [telemetryProperties, userId, uuid, deviceType]
+    [telemetryProperties, userId, uuid]
   );
 
   return { trackUsage };


packages/core/admin/admin/src/features/Tracking.tsx
```diff
@@ -576,7 +580,7 @@ const useTracking = (): UseTrackingReturn => {
 
       return null;
     },
-    [telemetryProperties, userId, uuid]
+    [telemetryProperties, userId, uuid, deviceType]
```
markkaylor: to be removed this file should have no changes

packages/core/admin/ee/server/src/index.ts
```diff
@@ -49,6 +51,12 @@ const getAdminEE = () => {
     async register({ strapi }: { strapi: Core.Strapi }) {
       // Run the the default registration
       await eeAdmin.register({ strapi });
+
+      // Register internal ai service
+      if (isAIEnabled) {
```
mathildeleg, line 56: I know `isAiEnabled` is in this file but seems a bit odd to me that this check is inside the `isAuditLogEnabled`...

Adzouz: Same question tbh

remidej: I don't think it's inside isAuditLogsEnabled 🤔 it's just the way prettier formats long lines

packages/core/admin/ee/server/src/services/index.ts
remidej, line 1: import not used

packages/core/admin/ee/server/src/index.ts
```diff
@@ -49,6 +51,12 @@ const getAdminEE = () => {
...
+        strapi.add('ai', createAIService({ strapi }));
```
remidej: I think what this does is different from creating a service, it's more creating an entire namespace outside of services. Is it intentional? If so I'd rename createAIService and move it out of services folder

packages/core/admin/ee/server/src/index.ts
mathildeleg: You're right 🤦🏻‍♀️ my bad

packages/core/admin/ee/server/src/ai/controllers/ai.ts
```diff
@@ -272,4 +273,11 @@ export default {
       throw fetchError;
     }
   },
+  async getAIFeatureConfig(ctx: Context) {
+    const aiFeatureConfig = await strapi.get('ai').getAIFeatureConfig();
```
remidej, line 277: the AI service is registered conditionally so you need to conditionally call it

packages/core/admin/ee/admin/src/services/ai.ts
remidej, line 1: Unrelated to this file:

>I chose to do it in EE because I saw other AI related code there but I also see some of the same code in admin (ie getAIToken)

It's unclear to me too. You should ping the DX team about that

Adzouz (commented):
Really nice and reusable, good job 👏 
Saw together but it's missing the param when creating a new entry, on the event `didCreateEntry` if you could just add it if you have the time but it's not prio 🙏
Tested with growth, EE and no licence and the rest looks good 🎈
Just had the same question as Mathilde if you could tell us more 🙏

For separating in EE, I don't know tbh but that's where the line between EE and Growth features is blurry and in my opinion I found it confusing to mix them both (especially when something is scoped to one only but I know that some features are shared between both).
For now I'd leave it here and revamp it in the future (maybe adding a new level at the same side as EE for growth and a level for both? I don't know, what do you think?).

Thanks Mark!

packages/core/admin/ee/server/src/index.ts
markkaylor: I think it's adjacent right? 

Screenshot 2025-11-03 at 11 19 06

markkaylor:
> Testing ✅ The only thing is I didn't get the `didSaveEntry` event but the `didEditEntry` which I guess is the same thing? But I always had the proper tracking I just had one remark but maybe I'm missing context :)

Thanks for pointing that out @mathildeleg , it should be `didEditEntry` we will no longer use `didSaveEntry` since `didEditEntry` already exists. I updated the PR description which was incorrect.

markkaylor:
> Really nice and reusable, good job 👏 Saw together but it's missing the param when creating a new entry, on the event `didCreateEntry` if you could just add it if you have the time but it's not prio 🙏 Tested with growth, EE and no licence and the rest looks good 🎈 Just had the same question as Mathilde if you could tell us more 🙏
> 
> For separating in EE, I don't know tbh but that's where the line between EE and Growth features is blurry and in my opinion I found it confusing to mix them both (especially when something is scoped to one only but I know that some features are shared between both). For now I'd leave it here and revamp it in the future (maybe adding a new level at the same side as EE for growth and a level for both? I don't know, what do you think?).
> 
> Thanks Mark!

I had the same thoughts @Adzouz I'm gonna bring the discussion to a BE sync

packages/core/admin/ee/server/src/ai/controllers/ai.ts
markkaylor: This controller should be registered conditionally as well so I think we are in a place that both are always available but can call it conditionally here to be certain.

packages/core/admin/ee/server/src/index.ts
markkaylor: ok I can just register it as a service, that's what I had done originally but I saw audit logs and decided to follow that approach.

packages/core/admin/ee/admin/src/services/ai.ts
markkaylor: I was planning on bringing it to a BE sync but can also ping now. I'd be ok with merging it like this for now since we've tested it already and it needs to get in the release branch. Whatever the decision we make in the BE sync there is going to be some refactoring/cleanup. I see duplicate functions in admin and EE like (getAIToken, useAIAvailability, getAIUsage)

packages/core/admin/ee/server/src/index.ts
Adzouz: Oh wow, good catch, yes it was just an illusion making us thinking it was inside the definition of `isAuditLogsEnabled`. Mondays imma right

packages/core/admin/ee/admin/src/services/ai.ts
remidej: for sure, I just meant to get to the bottom of this

packages/core/admin/ee/server/src/ai/controllers/ai.ts
remidej: ah didn't catch that

packages/core/admin/ee/server/src/index.ts
remidej: I'm fine with the non-service approach, just find the service name misleading in this case

packages/core/admin/ee/admin/src/services/ai.ts
markkaylor: 💯

packages/core/admin/ee/server/src/index.ts
markkaylor: The `add` / `get` method is apparently part of the Container interface (according to windsurf's AI wiki 🙃), wdyt about `createAIContainer`? I should probably move the function into a folder called containers then though 🤔 

Screenshot 2025-11-03 at 11 39 57

remidej: that would work 👍

packages/core/admin/ee/admin/src/services/ai.ts
markkaylor: Started the conversation in the DX channel

markkaylor:
> I think there's an issue with cache being cleared for your new query when creating an entry :( AI i18n is enabled but when I create a new entry, the flag is false:
> 
> Image

ah yeah I'm probably not invalidating it for create

Adzouz:
> > I think there's an issue with cache being cleared for your new query when creating an entry :( AI i18n is enabled but when I create a new entry, the flag is false:
> > Image
> 
> ah yeah I'm probably not invalidating it for create

Well you're invalidating the request when the settings is updated so I'm a bit confused..

markkaylor:
oh yeah you're right I am 🤔

markkaylor:
It looks like it's working on my end 🤔 , @Adzouz could you try a yarn setup just to be sure 🙏 . (ignore the part in the video where I get confused about the back button)

https://github.com/user-attachments/assets/c8d35ef4-dbcf-4475-a8b9-55820c009d24

PR #24765, opened 2025-11-03
fix: accessible responsive left menu

HichamELBSI:
### What does it do?

This PR fixes the keyboard and focus issues when trying to open the burger menu drawer when you work on small screen.

### Why is it needed?

Needed for keyboard navigation and a11y.

### How to test it?

- Open the admin in a small screen to render the burger menu button
- Navigate with the keyboard by pressing Tab
- You should now be able to focus the burger menu button
- Open it with "Space" or "Enter"
- On open, the first link of the menu should be focused
- The focus is now trapped in the menu.
- To quit the focus trap, you can close the menu with "Escape" or "Enter" on a link.
- The focus should be restored to the burger menu button when you press "Escape".

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/components/MainNav/MainNavLinks.tsx
+++ b/packages/core/admin/admin/src/components/MainNav/MainNavLinks.tsx
@@ -106,13 +106,9 @@ const MainNavIcons = ({
         return isDesktop || (!isDesktop && linkMobile) ? (
           <Flex tag="li" key={link.to}>
             <GuidedTourTooltip to={link.to}>
-              {isDesktop ? (
-                <NavLink.Tooltip label={labelValue}>
-                  <LinkElement />
-                </NavLink.Tooltip>
-              ) : (
+              <NavLink.Tooltip position={isDesktop ? 'right' : 'bottom'} label={labelValue}>
                 <LinkElement />
-              )}
+              </NavLink.Tooltip>
             </GuidedTourTooltip>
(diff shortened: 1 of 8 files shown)


mathildeleg (approved):
Tested ✅ 
The animation looks very smooth!

packages/core/admin/admin/src/components/MainNav/MainNavLinks.tsx
```diff
@@ -106,13 +106,9 @@ const MainNavIcons = ({
         return isDesktop || (!isDesktop && linkMobile) ? (
           <Flex tag="li" key={link.to}>
             <GuidedTourTooltip to={link.to}>
-              {isDesktop ? (
-                <NavLink.Tooltip label={labelValue}>
-                  <LinkElement />
-                </NavLink.Tooltip>
-              ) : (
+              <NavLink.Tooltip position={isDesktop ? 'right' : 'bottom'} label={labelValue}>
```
Adzouz, line 109: I don't know if the tooltip should appear on click on mobile because it's displaying it and you need to click again elsewhere to make it disappear. We discussed with designers to remove those on mobile, that's why I added the isDesktop condition. What do you think?
Or maybe we could discuss to auto blur after a few seconds?

Image

Adzouz (dismissed):

Adzouz (approved):

PR #24767, opened 2025-11-03
fix: warn user when CM edit view is unsaved on page reload/close

HichamELBSI:
### What does it do?

The existing `<Prompt />` we are using in the EditView provided by react-router-dom, only intercepts in-app (client side) route transitions. This component will not intercepts browser actions (reload, tab close, etc). We need to use a custom listener `beforeunload` to handle these use cases.

### Why is it needed?

Prevent data loss when the user reloads the page by mistake

### How to test it?

- Open the admin (v4)
- Start to edit a content type
- Try to close the tab or refresh the page
- A confirm dialog should be triggered to confirm the action.


HichamELBSI:
There was a misunderstanding for this task, I will close it to fix the issue on Strapi 5 instead. We aren't maintaining 4 anymore.

Issue #17731, opened 2023-08-21
Accidentally closing Strapi window => losing unsaved work


antgel:
### Required System information

- Node.js version: 18.17.1
- NPM version: 9.8.1
- Strapi version: 4.12.4
- Database: sqlite
- Operating system: Ubuntu Linux 22.04
- Is your project Javascript or Typescript: N/A

### Describe the bug

I accidentally closed my Strapi browser window and lost an hour's work in progress. I was on my way to press Save but that's by the by.

### Steps to reproduce the behavior

- In Content Manager, work on a long edit to content in a Rich Text field.
- Close browser.
- Cry.
- Hit Ctrl-Shift-T, hoping that the form would contain the previous text.
- Cry again.
- Open issue here because surely this is a bug.

### Expected behavior

I expected a dialog or something similar, perhaps trapping `beforeunload` to stop me shooting myself in the foot.

### Additional context

Funnily enough I created https://feedback.strapi.io/feature-requests/p/autosave-when-editing-fields today because I could feel that something like this might happen.

Based on this, I can't use Strapi for editing content now; I'll have to edit elsewhere then paste in, which somewhat defeats the purpose of the Rich Text Field.

antgel:
Adding to my OP: Hitting refresh in the browser suffers from the same problem.

thakkarnetram:
Can I take up this issue & try ?

Boegie19:
@thakkarnetram everyone can always make a PR. the way we see if someone is working on a issue is to check if a PR has been linked to this issue. Still writing a PR would not mean it gets accepted also this PR still this PR would be a back burner PR. so it could take a long time before it gets looked at and reviewed.

thakkarnetram:
@Boegie19 Okay . so i will try and solve this issue and raise a PR linked with it !

antgel:
@Boegie19 Just out of interest, why would this be a "back burner PR"? Losing unsaved work is quite a downer and disappointing UX.

Boegie19:
@antgel since I agree it is an annoying issue. but there are issues with more impact that will get more priority. so prs like this can take a while to get accepted.

antgel:
I don't know how "impact" is defined and I'm not underestimating the amount of work involved in reviewing PRs, but it's a shame that a ready offering might not get through to `main` in a reasonable time. :man_shrugging:

Arunjenson:
any update on this?

antgel:
[...]

PR #24782, opened 2025-11-05
fix: warn user on page reload/close if edit view has unsaved data

HichamELBSI:
### What does it do?

Warn the user when CM edit view has unsaved data (on reload, page close)

### Why is it needed?

We are currently warning the user when CM edit view has unsaved data but only on client side navigation update, all the browser actions aren't handled (tab close, browser close, page reload). 

### How to test it?

- Create or edit a content in the Content manager
- Before saving, try to close the tab (a native alert should be visible to confirm the action and warn the user about the unsaved data)
- ⚠️ The alert is different from the other alert (when you have unsaved data and click on a link in the page) and that's normal (`beforeunload` this native browser event we are using to catch these browser actions is displaying a native confirm alert and we don't have the control of it's content)
- Same for page reload from the browser url
- Same for browser close action

### Related issue(s)/PR(s)

Closes #[17731](https://github.com/strapi/strapi/issues/17731)

### DIFF
--- a/packages/core/admin/admin/src/hooks/useWarnIfUnsavedChanges.ts
+++ b/packages/core/admin/admin/src/hooks/useWarnIfUnsavedChanges.ts
@@ -1,0 +1,16 @@
+import { useEffect } from 'react';
+
+export function useWarnIfUnsavedChanges(enabled: boolean) {
+  useEffect(() => {
+    if (!enabled) return;
+
+    const handleBeforeUnload = (e: BeforeUnloadEvent) => {
+      e.preventDefault();
+      e.returnValue = '';
+      return '';
+    };
+
+    window.addEventListener('beforeunload', handleBeforeUnload);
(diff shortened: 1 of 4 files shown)


packages/core/admin/admin/src/hooks/useWarnIfUnsavedChanges.ts
```diff
@@ -0,0 +1,16 @@
...
+  useEffect(() => {
+    if (!enabled) return;
+
+    const handleBeforeUnload = (e: BeforeUnloadEvent) => {
+      e.preventDefault();
+      e.returnValue = '';
+      return '';
+    };
+
+    window.addEventListener('beforeunload', handleBeforeUnload);
```
HichamELBSI, line 13: `beforeunload` this native browser event is displaying a native confirm alert and we don't have the control of it's content.

mathildeleg (approved):
LGTM ✅

Issue #23450, opened 2025-04-30
Bug:  Errors for single types without default locale


Lvdwardt:
### Node Version

20

### NPM/Yarn/PNPM Version

PNPM 10.0.0

### Strapi Version

5.12.3

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Bug Description

"When creating a new entry of a single type for a non-default locale, and no entry exists yet for the default locale, I encounter the following issues:

1. The CMS shows a warning: 'An error occurred while fetching draft relations on this document.'

2. In the admin.ts preview configuration, the locale is incorrectly identified as undefined instead of the intended locale.

Both issues disappear immediately after I create and save an entry for the single type in the default locale."

### Steps to Reproduce

1. Create a new locale in the CMS.
2. Navigate to a single-type content model.
3. Create and save an entry for the new locale without creating one for the default locale first. 
4. Get an error message, and the live preview config logs locale undefined__

### Expected Behavior

To not get any errors when saving a single type entry without a default locale, and to get the right locale in the preview config.

### Logs

```shell

```

### Code Snippets

----admin.ts----

preview: {
    enabled: true,
    config: {
      allowedOrigins: ["*"],
      async handler(uid, { documentId, locale, status }) {
         console.log("handler", uid, documentId, locale, status);
      }
}

### Additional information

* It happens both locally and on Strapi Cloud
* The Strapi Cloud runtime logs only show status 200 logs, no errors there.
* The browser network tab shows a 404:
`/content-manager/single-types/api::single-type-name/actions/countDraftRelations?locale={not_default_locale}`

* It only happens for single types, not for collection types.

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #24785, opened 2025-11-06
fix: get draft relations count from document with locale

mathildeleg:
### What does it do?
Get `countDraftRelations` from current locale, not just for the document.

### Why is it needed?
If a single type had entries in locale(s) other than the default locale, there was an alert to display the error in fetching the `countDraftRelations` for that document. The goal here is to remove that alert by fetching the relations from the locale, not just the document, so that the correct count is returned for that locale's entry.

⚠️ _Seems like the `countDraftRelations` returns all relations (published are counted as draft relations too) instead of just draft relations for now: https://www.notion.so/strapi/Count-draft-relations-56901b492efb45ab90d42fe975b32bd8 but I'm not touching this here_

**BEFORE**
https://github.com/user-attachments/assets/2587e26f-65c6-4650-8201-a6f63771bafb

**AFTER**
https://github.com/user-attachments/assets/6e438739-c9dd-47ab-a732-6dcbe46cddd4

### How to test it?
* Go to the Settings for the Internationalization and add an additional locale 
* Go to the CTB and create a Single type with at least one field + localization enabled
* Go to the CM and create an entry in an additional locale
* Navigate between the "Create an entry" (in the default locale) and your existing entry, without creating an entry in the default locale => There should be no Alert.

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/23450

### DIFF
--- a/packages/core/content-manager/server/src/controllers/single-types.ts
+++ b/packages/core/content-manager/server/src/controllers/single-types.ts
@@ -318,7 +318,7 @@ export default {
       return ctx.forbidden();
     }
 
-    const document = await findDocument({}, model);
+    const document = await findDocument({}, model, { locale });
     if (!document) {
       return ctx.notFound();
     }


remidej (approved):
works well 👍

PR #24786, opened 2025-11-06
chore(tests): add setupDatabaseReset utility

markkaylor:
### What does it do?

Add `setupDatabaseReset` utility. Accomplishes the same goal as `testInTransaction` by resetting the database after each test has run. It's slower but more reliable and is not vulnerable to test failures caused by function calls outside the test's transaction context when using testInTransaction. 

Updates the docker file to spin up test databases since it was throwing an error when starting

### Why is it needed?

testInTransaction often causes API tests to fail since some async code is intentionally fired without await and then tries to execute after the test's transaction context has closed

For example see: https://github.com/strapi/strapi/actions/runs/19104795038/job/54586327649?pr=24648

### How to test it?

The API tests should pass on this PR

Test it locally by running the docker test file `docker-compose -f docker-compose.test.yml up`

Now you can test all database dialects with `yarn test:api --db=<sqlite | postgres | mysql>`

### Related issue(s)/PR(s)

I've tested (in the CI) this will resolves the transaction error causing tests to fail on:

https://github.com/strapi/strapi/pull/24648

### DIFF
--- a/tests/api/utils/index.ts
+++ b/tests/api/utils/index.ts
@@ -31,12 +31,6 @@ async function captureInitialTestData() {
   isTestDataCaptured = true;
 }
 
-/**
- * Reset all database tables to their initial state
- *
- * NOTE:
- * Only use sparingly where needed as the operation is slower than testInTransaction
- */
 async function resetTestDatabase() {
   // Use Strapi's built-in schema update mechanism to disable constraints (e.g. foreign key constraints)
   await strapi.db.dialect.startSchemaUpdate();
@@ -65,6 +59,9 @@ async function resetTestDatabase() {
 /**
  * Setup database reset for a test suite
  * Call this in your describe block to automatically reset after each test
+ *
+ * NOTE:
+ * Only use sparingly where needed as the operation is slower than testInTransaction
  */
 export function setupDatabaseReset() {
   let isDataCaptured = false;


tests/api/utils/index.ts
markkaylor, line 1: ℹ️ This is really the only file to review, all other updates are just replacing `testInTransaction` with `setupDatabaseReset`

tests/api/utils/index.ts
```diff
@@ -1,27 +1,98 @@
...
+// Keep old interface for backward compatibility but log deprecation
 export const testInTransaction = (...args: Parameters<jest.It>) => {
-  if (args.length > 1) {
-    return it(
-      args[0], // name
-      wrapInTransaction(args[1]), // fn
-      args[2] // timeout
-    );
-  }
+  console.warn('testInTransaction is deprecated. Use setupDatabaseReset() instead.');
```
markkaylor: Thought this was nice so if you go to reach for it you know what to do next but happy to remove it otherwise

innerdvations:
Have you done tests on speed? I think this was originally added during the creation of the document service when we discovered the tests ran about 90% faster with this method.

As long as this is not a major increase in runtime I'm all for it though

markkaylor:
> Have you done tests on speed? I think this was originally added during the creation of the document service when we discovered the tests ran about 90% faster with this method.
> 
> As long as this is not a major increase in runtime I'm all for it though

I just did some and indeed it appears quite a bit slower, wdyt of keeping both? Favor the transaction util but use table truncate reset where needed?

As an example adonis exposes utils like that for similar reasons (ie nested transaction) https://docs.adonisjs.com/guides/testing/database

innerdvations:
> > Have you done tests on speed? I think this was originally added during the creation of the document service when we discovered the tests ran about 90% faster with this method.
> > As long as this is not a major increase in runtime I'm all for it though
> 
> I just did some and indeed it appears quite a bit slower, wdyt of keeping both? Favor the transaction util but use a true database reset where needed?
> 
> As an example Adonis does something like that [docs.adonisjs.com/guides/testing/database](https://docs.adonisjs.com/guides/testing/database)

I'm fine with having this as an option too, it's better than our current method of requiring every test to clean up after itself and then causing unrelated tests down the line to fail when they don't. The ideal solution would be a db reset that is fast enough to run before every test suite, so maybe this would be a first step towards that.

markkaylor:
@innerdvations If you are ok with the changes could you give me an approval? As you mentioned we can always improve it or even remove it and find a better solution if it isn't suiting our needs.

tests/api/core/strapi/document-service/update.test.api.ts
markkaylor, line 1: oops I should revert this before merging

markkaylor: actually it's fine mathilde will need this update anyway

araksyagevorgyan (approved):
LGTM 👍

Issue #21201, opened 2024-09-09
deprecation warning in 5.0.0-rc.19


datner:
## Bug report
I really think that a v5 should start things off without deprecation warnings

![CleanShot 2024-09-10 at 00 15 36](https://github.com/user-attachments/assets/31c23fc4-496c-4c11-b0c3-1fc1f7368729)

(why is there typescript shipped with `@strapi/admin`? also the latest `typedoc` supports the latest typescript)

```
9 deprecated subdependencies found: buildmail@3.10.0, glob@7.2.3, inflight@1.0.6, mailcomposer@3.12.0, resolve-url@0.2.1, rimraf@3.0.2, source-map-resolve@0.5.3, source-map-url@0.4.1, urix@0.1.0
```
I know that some are harder to rip out than others, but I've gathered their origins to at least get a ball rolling.

- [ ] `rimraf`+`glob` - The deprecated versions are being dragged in by `react-query@3`. `react-query` is v5 as of writing. Is there a concrete reason why is it not updated? I promise the migration is not that dramatic (low priority since can be forced using resolutions)
  - [ ] `@strapi/admin`
- [ ] `buildmail` + `mailcomposer` - and aggressively deprecated and fully abandoned (7yo) util used by the ever-so-slightly less abandoned `sendmail`. _Surely_ we can move to the `nodemailer` plugin by default at long last
  - [ ] `@strapi/provider-email-sendmail`
- [ ] `inflight` - A dependency of the deprecated `glob`, but also down the chain off `copyfiles`, which I am not convinced provides functionality that is required anymore considering node18. Also offending is `node-plop`, it's safe to say that it can be updated to the year-old latest at `0.36.0`
  - [ ] `@strapi/generators`
- [ ] `resolve-url` + `source-map-resolve` + `source-map-url` + `urix` - `plop` is the offending party this time. `plop` is v4 now, we are in v2
  - [ ] `@strapi/generators`
- [ ] `codemirror` - not deprecated, but theres a version skew with the downstream. We are using v5, and our dependency wants a peer v6
  - [ ] `@strapi/admin`

We can knock most of this off easy, what do you think?

derrickmehaffy:
I'd say it's worth a shot, right now the team is focused on the last remaining bug fixes before stable, are you willing to submit a PR to try upgrading these?

Note that for anything react related if there is any breaking changes at all it will be declined as we are in the RC about 2 weeks from stable release and no additional breaking changes are allowed.

[...]

PR #24791, opened 2025-11-06
chore: remove copyfiles dependency

plevavas:
## What does it do?

- Drops the `copyfiles` dependency from `@strapi/core`, `@strapi/strapi`, and `@strapi/generators`, including the corresponding `yarn.lock` cleanup.
- Replaces the generator `copy-files` script with a Node-based executable (`scripts/copy-files.mjs`) that clears and recopies `dist/templates` and `dist/files` using `fs.cp`.

## Why is it needed?

- `copyfiles` has been unmaintained for ~3 years and still ships `inflight`, which currently has an open CVE and deprecated; removing it keeps our dependency tree healthier.
- Relying on Node built-ins gives us consistent cross-platform behavior while avoiding stale artifacts left in `dist`.

## How to test it?

- Install dependencies if needed: `yarn install`.
- Build the generators workspace: `yarn workspace @strapi/generators build`.
  - Confirm that `packages/generators/generators/dist/templates` and `packages/generators/generators/dist/files` are regenerated with the expected contents if the directory exists.

## Related issue(s)/PR(s)

- #21201


msudol:
FWIW - I made a fork and updated the copyfiles repo as copyfiles-ng - it's API is 1:1 with the original and can be cleanly swapped out.  https://www.npmjs.com/package/copyfiles-ng

Issue #24642, opened 2025-10-21
Max length on "short text" data type may not work


thematrixdev:
### Node Version

22

### NPM/Yarn/PNPM Version

pnpm 10

### Strapi Version

5.28.0

### Operating System

Docker/Podman/LXC

### Database

MariaDB

### Javascript or Typescript

Javascript

### Bug Description

`short text` creates a `varchar(255)` field on the database.
`maximum length` could be a very very big number.

### Steps to Reproduce

1. go to content type builder
2. create a text field
3. choose short text
4. enter any number larger than 255 in maximum length

### Expected Behavior

I would suggest we can specify the `VARCHAR` size.
In my case this time, I just need `VARCHAR(400)`, but I have to switch to `long text`.
Instead I am modifying the size directly on the database.

If not, maximum length should be limited to 255 when `short text` is selected.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #24794, opened 2025-11-07
fix(core): max length on short text

araksyagevorgyan:
### What does it do?

Limits the Maximum length for Short text fields to 255 and shows a validation error if a higher value is entered.

### Why is it needed?

Short text maps to `VARCHAR(255)` in the database. Allowing larger values is misleading and can cause inconsistencies.

### How to test it?

1. Go to Content Type Builder
2. Create (or edit) a text field
3. Choose Short text as the type
4. Enter a value greater than 255 in the Maximum length field
5. A validation error message should be displayed

image

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/24642


markkaylor (approved):
Looks good to me, tested and works as expected

markkaylor:
~Don't forget the milestone :)~ oh nvm looks you got it my bad

PR #24799, opened 2025-11-07
fix(core): allow description and placeholder be empty

araksyagevorgyan:
### What does it do?

Fixes an error in View mode that occurs when removing an existing placeholder or descriptions from a field.

### Why is it needed?

To allow users to clear placeholders and descriptions without triggering errors and ensure expected saving behavior.

### How to test it?

1. Create a content type with a text field.
2. In View mode, set and save a placeholder or description.
3. Reopen View mode, remove the placeholder or description, and save.
    - No error should appear, and the placeholder should be cleared successfully.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/22966


markkaylor (approved):
Tested looks good to me. Doesn't look like we have any tests for validation. Could be worth it to see if AI could generate that for us quickly 🤷 up to you

Issue #23468, opened 2025-05-05
Child Record Dropdown Displays Value in Incorrect Language


praful-akkodis:
### Node Version

20.17.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.12.5

### Operating System

Windows 10

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

When using Internationalization (i18n) in Strapi with a one-to-many relationship between a main and a child collection, the child records dropdown in main collection do not always respect the selected language setting.

After creating and populating records in multiple languages, switching back to the default language (English, in this case) still shows some child records in the secondary language. This leads to inconsistent and incorrect language display in the admin panel, particularly in dropdowns or relationship views.

This issue affects the clarity and usability of multilingual content, especially when managing localized data.

### Steps to Reproduce

1. Enable Internationalization (i18n) and add a secondary language.
2. Create a main collection and a child collection with a one-to-many relationship.
3. Add records to both the main and child collections in multiple languages.
4. Switch between the available languages in main collection, then return to the default language (English in this case).
5. Notice that relations dropdown records are still displayed in the wrong language, even though English is selected in the top-right language dropdown.

### Expected Behavior

When seeing records in default language (English) in a multilingual setup with one-to-many relations, child record dropdowns still show values in the secondary language. This causes inconsistent language display in the admin panel. Expected behavior is that all records should reflect the selected language.

### Logs

```shell

```

### Media

Listing screen with issue
![Image](https://github.com/user-attachments/assets/23edd50b-1597-4412-bcb4-765b43d0e823)

Detail page for English 
![Image](https://github.com/user-attachments/assets/3b2ecf81-b65a-48ae-9402-c48011a2050a)

Detail page for Arabic
![Image](https://github.com/user-attachments/assets/419ce90d-dd70-4d6b-ba3d-09e8c257e40e)

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #24805, opened 2025-11-07
fix: show relations per locale in listview

mathildeleg:
### What does it do?
Fetch the relations per locale if localization is enabled for that content type.
(+ re-add and fix the related tests and add new tests related to this)

### Why is it needed?
The Relations dropdown showed the default locale's relations despite the list view currently being filtered by another locale, thus showing the entry in that locale (and not the default one)

**BEFORE**
_List view with English entry's relation in dropdown despite being in French's list view_
Capture d’écran 2025-11-07 à 16 41 17
_English Entry_
Capture d’écran 2025-11-07 à 16 42 23
_French Entry_
Capture d’écran 2025-11-07 à 16 41 29

**AFTER**
Capture d’écran 2025-11-07 à 16 40 08

### How to test it?
* In the Settings, enable Internationalization and add an additional locale.
* In the CTB, create two Content Types (for example Book and Author) and connect them with a one-to-many relationship (_Author has Many Books_).
* Create entries to both Content Types in the default locale and additional locale.
* Switch between the available locales in the list view of _Author_, then return to the default locale (English in this case).
* See that the Relations in the Books column are correct per locale.

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/23468

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/components/TableCells/tests/Relation.test.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/components/TableCells/tests/Relation.test.tsx
@@ -1,7 +1,7 @@
-import { render as renderRTL, waitFor, screen, server } from '@tests/utils';
-import { Route, Routes } from 'react-router-dom';
-import { rest } from 'msw';
 import { useQueryParams } from '@strapi/admin/strapi-admin';
+import { render as renderRTL, waitFor, screen, server } from '@tests/utils';
+import { rest } from 'msw';
+import { Route, Routes } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


mathildeleg:
> Everything seems to work as expected if both sides have i18n
> 
> Image
> If the parent relation does NOT have i18n the count is off in the list view. Always showing 1 item. Not sure if that's a new issue introduced here or an existing separate issue.
> 
> Image

Not related to the changes in the PR, seems to be a deeper issue when changing the settings of a content type and having issues with the relations in the content type afterwards, as seen here:
https://github.com/user-attachments/assets/1c5a2f27-6830-47b0-81a1-36d0fdb1004f

PR #24818, opened 2025-11-10
fix: dynamically update rate limit prefix key based on route

derrickmehaffy:
### What does it do?

Changes the prefixKey in the users-permissions rate-limit middleware to exclude the user identifier on routes that don't need/have it.

### Why is it needed?

Security Vulnerability reported

### How to test it?

Try testing a route that doesn't normally need a `email` and manually set one in the request payload and make a ton of requests, if you get around the rate-limit then it ain't working

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/security/advisories/GHSA-7mqx-wwh4-f9fw


innerdvations:
I pushed [a few minor fixes](https://github.com/strapi/strapi/pull/24818/commits/41313b99d19b2a31e461f942903f6c9351494fe7) and tests, check them out to make sure you approve

innerdvations (approved):

Issue #22641, opened 2025-01-14
CLONE: BUG V5: tries to find Metadata after remove a component


hanpaine:
### Required System information
Node.js version: 20.9.0
NPM version: 10.4.0
Strapi version: 4.20.0
Database: PostgreSQL
Operating system: Linux
Is your project Javascript or Typescript: TypeScript

### Describe the bug
I removed a component, because I misspelled a letter. I was still using it inside a collection in Content Manager. Obviously, after delete it, disappeared. But now, I can't modify the entry of the collection, it says:

`Error: Metadata for "sections.mycomponent" not found
`
### Additional context
The obvious solutions has been create the misspelled component again, then delete the stucked entry, and then remove the component again.

### Related tickets
This is a V5 clone of #19535 to track the V5 PR

PR #24833, opened 2025-11-13
fix: run orphan removal for the 'related_type' column

boazpoolman:
### What does it do?

It cleans up orphans in the `files_related_mph` database table after deleting a content type.

### Why is it needed?

To solve an issue similar to this one https://github.com/strapi/strapi/issues/22641, where orphaned morph types remain in the database after deleting parts of the schema:

### How to test it?

1. Create 2 content types which both have a media field
2. Create 1 media item
3. Create an entry for each content-type with a reference to the same media item
4. Remove 1 of the content types
5. Try to call the API on the other content type whilst populating it's media field
6. Witness the error

### Related issue(s)/PR(s)

A similar issue was resolved for components in PR https://github.com/strapi/strapi/pull/22267. I expect this issue was just more vocal for components because of the way dynamic zones work. That being said I was able to use the same repair script used to fix that issue, so good job on that one @innerdvations!

### DIFF
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -457,6 +457,7 @@ class Strapi extends Container implements Core.Strapi {
     // // if schemas have changed, run repairs
     if (status === 'CHANGED') {
       await this.db.repair.removeOrphanMorphType({ pivot: 'component_type' });
+      await this.db.repair.removeOrphanMorphType({ pivot: 'related_type' });
     }
 
     // const alreadyRanComponentRepair = await this.store.get({


jhoward1994:
Hey, just want to clarify what to test here. 

I reverted the change and followed the instructions in the PR description with media fields but didn't see an error after deleting one content type and then querying for the other through the content API.

Is that the right approach ?

boazpoolman:
Hi @jhoward1994.

I was lacking some detail here, sorry about that.
The issue occurs when you actually try to populate the fields of the media reference. Like so:

```
http://localhost:1337/api/pages/tyrzvsayk0rlsvkrpyy3z8f3?populate[media][populate]=*
```

Let me know if you're able to reproduce like this.

jhoward1994:
Thank you! I was able to test with that

jhoward1994 (approved):

Bassel17:
@boazpoolman 
anything else needs to be done on this  or is it good to merge ?

boazpoolman:
@Bassel17 this is good to go, solves the problem on my side.

PR #24838, opened 2025-11-13
Replace act() calls with the right testing library api when it is necessary

HichamELBSI:
### What does it do?

In many front end test files, the `act()` has been used even though we could use more relevants api such as `waitFor`, `findBy` and `waitForElementToBeRemoved ` that already wraps `act()`. I guess, this has been done this way because of the error logged when a test that cause state updates fail. One of the only place we need to explicitly use `act()` is when using with fake timers

### Why is it needed?

Hopefully less test flakiness

### How to test it?

- Run the front end unit test, make sure they pass
- `yarn test:front:update`

### Related issue(s)/PR(s)


markkaylor (approved):

Adzouz (approved):
Wooo the PR we've all been waiting for 🤩 Amazing, thanks 🙏

Issue #24829, opened 2025-11-12
Strapi fails to start: "No 'exports' main defined in file-type package" error on version 5.31.0


deivydarwoft:
### Node Version

20.18.3

### NPM

10.8.2

### Strapi Version

5.31.0

### Operating System

Windows 11

### Database

sqlite

### Javascript or Typescript

Typescript

### Bug Description

When starting a fresh Strapi project with the latest release, Strapi fails to start due to an error related to the file-type package. The error message is:
Error: Could not load js config file ...@strapi\upload\dist\server\index.js: No "exports" main defined in ...node_modules\file-type\package.json

### Steps to Reproduce

1. Create a new Strapi project: `npx create-strapi-app my-strapi-project`

2. Start Strapi in development mode: `npx strapi develop`

3. Observe the error in the console.

### Expected Behavior

Strapi should start successfully and display the admin panel without errors. The upload plugin should work as expected.

### Logs

```shell
npm run develop

> leonisa-cms@1.0.0 develop
> strapi develop

✔ Cleaning dist dir (36ms)
⠼ Loading Strapi[ERROR]  There seems to be an unexpected error, try again with --debug for more information 

[...]

PR #24848, opened 2025-11-14
fix(upload): file-type does not support commonjs

markkaylor:
### What does it do?

Replace static import with dynamic import for file-type package to support ESM-only modules in both CommonJS/ESM builds.

The maintainer mentions this approach here: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

### Why is it needed?

file-type is ESM-only and lacks CommonJS exports. When Strapi builds its CommonJS output, the bundler fails to transform the static import to require(). This breaks app startup on Node.js < 22, which has stricter ESM/CommonJS interop. Dynamic imports work at runtime in both module formats.

### How to test it?

You will need to be able to switch node versions with something like https://github.com/nvm-sh/nvm

Switch to node 20 (`nvm install 20` if it's not already installed. `nvm use 20` to use it, `node -v` to confirm)
Create a strapi app with `npx create-strapi-app@0.0.0-experimental.f9ca9ac9518346cac58a4a3a576bc9d9d795570d test-fix --quickstart`
Update the config in `config/plugins.js`:

```
upload: {
    config: {
      security: {
        deniedTypes: ['video/*'],
      }
    }
  }
```
Run `npm run develop` there should NOT be an error
Go to the media library and try to upload a video, it should be rejected
Try to upload other file types not in the deniedTypes list, it should work

Test the same for all node versions up to 24

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/24829

### DIFF
--- a/packages/core/upload/server/src/utils/mime-validation.ts
+++ b/packages/core/upload/server/src/utils/mime-validation.ts
@@ -49,6 +49,11 @@ export async function detectMimeType(file: any): Promise<string | undefined> {
   }
 
   try {
+    /**
+     * Use dynamic import to support file-type which is ESM-only
+     * Static imports fail during CommonJS build since bundler can't transform ESM-only packages
+     * @see https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
+     */
     const { fileTypeFromBuffer } = await import('file-type');
 
     const result = await fileTypeFromBuffer(new Uint8Array(buffer));


packages/core/upload/server/src/utils/mime-validation.ts
```diff
@@ -50,6 +49,8 @@ export async function detectMimeType(file: any): Promise<string | undefined> {
   }
 
   try {
+    const { fileTypeFromBuffer } = await import('file-type');
```
markkaylor, line 57: add comment

mathildeleg (approved):
LGTM, well done ✅

araksyagevorgyan:
I tested further, LGTM 🚀

araksyagevorgyan (approved):

PR #24849, opened 2025-11-14
fix(content-manager): preserve fields with visible:false during form submission

Bassel17:
### What does it do?

This PR fixes the `handleInvisibleAttributes` function to properly distinguish between:
1. **Static hidden fields** (`visible: false`) 
2. **Conditionally hidden fields** (`conditions.visible`)

**Technical changes:**
- Added early return for fields with `visible: false` to pass them through unchanged
- These fields are preserved from either `data` or `initialValues` without processing
- The function continues to properly handle `conditions.visible` (JSON Logic rules)
- Added comprehensive tests covering both scenarios

### Why is it needed?

**Problem:** 
Fields with `visible: false` were being incorrectly set to `null` when not present in form data,

**Use case:**
Plugins (like strapi-plugin-sitemap) add hidden fields that are permanently invisible
```javascript
_.set(attributes, 'sitemap_exclude', {
  visible: false,  // Permanently hidden
  type: 'boolean',
});
```
### Related issue(s)/PR(s)

fixes issue mentioned in this PR #24628

### DIFF
--- a/.github/workflows/find_duplicate_issue.yml
+++ b/.github/workflows/find_duplicate_issue.yml
@@ -1,0 +1,78 @@
+name: Find duplicate from issue id
+
+on:
+  workflow_dispatch:
+    inputs:
+      issue_id:
+        description: 'Id of the issue to check for duplicates'
+        required: true
+
+jobs:
+  deduplicate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Check for duplicate issues
+        uses: anthropics/claude-code-action@v1
+        with:
+          prompt: |
+            Analyze this issue and check if it's a duplicate of other issues in the repository.
+
+            Issue: #${{ github.event.inputs.issue_id }}
+            Repository: ${{ github.repository }}
+
+            Your task:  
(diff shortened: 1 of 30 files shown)


Bassel17:
Experimental: `0.0.0-experimental.14806124ce23818298c02cb84342c60db4704250`

packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
```diff
@@ -262,6 +262,20 @@ const handleInvisibleAttributes = (
 
   for (const [attrName, attrDef] of Object.entries(schema.attributes)) {
     const fullPath = [...path, attrName].join('.');
+
+    // Skip processing fields with visible: false - they should be passed through as-is
+    // This function only handles conditions.visible (JSON Logic), not the visible boolean property
+    if ('visible' in attrDef && attrDef.visible === false) {
+      // Pass through the value as-is from data or initialValues
+      const userProvided = Object.prototype.hasOwnProperty.call(data, attrName);
```
markkaylor: I think `Object.hasOwn` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn is the recommended method. It has good browser support now as far as I can tell. wdyt? It avoids prototype pollution edge cases and is more readable

Bassel17:
experimental: `0.0.0-experimental.0d1bdfce3abec8974dc41933b18ba81c45b93475`

the latest experimental should fix this @boazpoolman

boazpoolman (approved):
That solved the issue! 🙂

PR #24869, opened 2025-11-18
test(e2e): merge cli and e2e utils, runner, and template

innerdvations:
### What does it do?

- Merges the utilities, app-template, and test runner for e2e and cli tests, which gives the cli tests all the "missing features" that e2e has and removes a bunch of duplicated code, much of which had diverged
- adds a new 'create-user' test to provide example/proof that the database reset is working for CLI tests

### Why is it needed?

90% of their utilities/runner/template functionality is identical, the only thing that needs to be different between them is the command that runs the test (`playwright` vs `jest`) -- they were only duplicated/diverged due to some difficulties when initializing the cli tests

### How to test it?

`yarn test:e2e` and `yarn test:cli` should still work the same as before. The only difference is that we have only one shared template app and fewer utility files since they are now shared

### Related issue(s)/PR(s)

### DIFF
--- a/tests/e2e/tests/content-manager/edit-view/boolean-clear-button.spec.ts
+++ b/tests/e2e/tests/content-manager/edit-view/boolean-clear-button.spec.ts
@@ -1,8 +1,8 @@
 import { expect, test } from '@playwright/test';
-import { resetDatabaseAndImportDataFromPath } from '../../../utils/dts-import';
-import { login } from '../../../utils/login';
-import { createContent, fillField } from '../../../utils/content-creation';
-import { navToHeader } from '../../../utils/shared';
+import { resetDatabaseAndImportDataFromPath } from '../../../../utils/dts-import';
+import { login } from '../../../../utils/login';
+import { createContent, fillField } from '../../../../utils/content-creation';
+import { navToHeader } from '../../../../utils/shared';
 
 test.describe('Boolean Component - Clear Button Functionality', () => {
   test.beforeEach(async ({ page }) => {


tests/cli/tests/strapi/strapi/create-user.test.cli.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const coffee = require('coffee');
```
Adzouz, line 3: Curiosity here: Is there a particular reason to use this package? Because this one is unmaintained for 2 years now and not very popular 😅 If it's just to run a command a check the output, could we simply use native child_process or execa (since it's already a dependency of the repo)?

innerdvations: We chose coffee back when we set up the tests because it has a lot of convenient helpers over just using execa. But if it's unmaintained now, we may want to switch before starting our deep dive into writing new CLI tests since at this point we don't make use of many features. Maybe we should explore [cli-testing-library](https://testing-library.com/docs/ecosystem-cli-testing-library)

tests/scripts/run-tests.js
nclsndr, line 1: Here I guess we'll be able to apply the changes from #24959

tests/scripts/run-tests.js
```diff
@@ -0,0 +1,468 @@
...
+      return yarg;
+    },
+    async handler(argv) {
+      const { type } = argv;
+
+      // Set up paths based on test type
+      const testRoot = path.join(cwd, 'tests', type);
+      const testAppDirectory = path.join(cwd, 'test-apps', type);
+      const testDomainRoot = path.join(testRoot, 'tests');
+      const templateDir = path.join(cwd, 'tests', 'e2e', 'app-template');
```
jhoward1994: (very minor) Should we put the 'app-template' somewhere else as cli and e2e both use it?

jhoward1994:
`tests/cli/README.md` - still mentions "based on the CLI app-template". I think that should be e2e template with things as they are now ?

tests/scripts/run-tests.js
innerdvations: Ahh, true, I missed that!

PR #24879, opened 2025-11-19
fix(cli): transfer command ignores CLI flags and prompts unnecessarily

### LINKED ISSUE #23688: Strapi Transfer CLI --force issue
### Node Version

20.15.0

### NPM/Yarn/PNPM Version

yarn 4.4.1

### Strapi Version

5.14.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Hey guys. It seems that in version v5.14.0 the `--force` parameter of strapi transfer CLI is broken. Below is my CLI call with said parameter and correct `--from` and `--from-token` parameters. And it still prompts for the URL (and then token) even though they're passed through the call and the `--force` parameter is present.

```
runner@fv-az720-478:~/work/redacted$ yarn strapi transfer \
--from https://valid-origin-url.com/admin \
--from-token 123-valid-token-123 \
--force \
--exclude files 
ℹ️  No transfer configuration found in environment variables
   → Add STRAPI_TRANSFER_URL and STRAPI_TRANSFER_TOKEN environment variables to make the transfer process faster for future runs
? Enter the URL of the remote Strapi instance to send data to: 
```

It works the same way locally (MacOS m3 14.6) and on Github runner instance (Ubuntu 22.04).
I think it might be related to the new transfer CLI feature from [this pr](https://github.com/strapi/strapi/pull/23421)
Am I doing something wrong or is it a bug? It worked properly in v5.13. 

### Steps to Reproduce

1. Run transfer command with `--from`, `--from-token` and `--force` parameters.
2. Observe CLI prompts still appearing.

### Expected Behavior

Transfer going through. CLI not prompting for additional information.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

jamesvillarrubia:
### What does it do?

This PR fixes two critical bugs in the `strapi transfer` CLI command that were introduced in PR #23421:

**Bug #1: Interactive prompts always run (Issue #23688)**  
The interactive prompt hook was not properly gated with a conditional check, causing it to execute even when `--from`, `--from-token`, and `--force` flags were provided via the command line.

*Technical Changes:*
- Wrapped the interactive prompts hook (lines 54-190 in `command.ts`) with `ifOptions()` to only execute when neither `--from` nor `--to` is provided
- This prevents the prompts from running when users provide all required options via CLI flags
- Made all `ifOptions` condition callbacks async to match TypeScript signature requirements

**Bug #2: Wrong return type in determineDirection()**  
The `determineDirection()` function was returning the URL object instead of the string `'from'` or `'to'`, causing downstream code to fail when trying to use the direction as an object key.

*Technical Change:*
- Changed `return opts.from` to `return 'from'` (line 105)
- Changed `return opts.to` to `return 'to'` (line 108)
- This ensures the direction is used correctly as a string key for accessing `opts[direction]` and `opts[${direction}Token]`

### Why is it needed?

Users running the transfer command in non-interactive environments (CI/CD pipelines, automated scripts) were unable to use the command because it would always prompt for input, even when all required flags were provided:

```bash
yarn strapi transfer \
  --from https://origin.com/admin \
  --from-token abc123 \
  --force \
  --exclude files
# ❌ Still prompted for URL and token
```

This broke:
- **CI/CD pipelines**: Automated deployments would hang waiting for input
- **Scripts**: Shell scripts using the transfer command would fail  
- **--force flag**: The force flag was completely ignored
- **Non-interactive usage**: Any automated or scripted usage was impossible

### How to test it?

**Test Scenario 1: Non-interactive usage with all flags (reproduces #23688)**
```bash
# Should proceed WITHOUT prompts
yarn strapi transfer \
  --from https://source.strapi.io/admin \
  --from-token YOUR_TOKEN \
  --force \
  --exclude files

# Expected: Transfer proceeds directly
# Before fix: Would prompt for URL and token
```

**Test Scenario 2: Non-interactive push with all flags**
```bash
# Should proceed WITHOUT prompts
yarn strapi transfer \
  --to https://destination.strapi.io/admin \
  --to-token YOUR_TOKEN \
  --force

# Expected: Transfer proceeds directly
# Before fix: Would prompt for URL and token
```

**Test Scenario 3: Interactive usage without flags (should still work)**
```bash
# Should show interactive prompts
yarn strapi transfer

# Expected: Prompts for direction, URL, and token
# Should still work after fix
```

**Test Scenario 4: Partial flags (should prompt for missing)**
```bash
# Should prompt only for token
yarn strapi transfer --from https://source.strapi.io/admin

# Expected: Prompts only for missing token
# Should work after fix
```

**Unit Tests:**
```bash
yarn test:unit packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.test.ts
```

New tests added:
- `--from` with auth not provided (should error)
- `--from` with source URL and token (should succeed)
- `--from` uses local Strapi destination when `--to` not specified

**Build and Type Checking:**
```bash
yarn build --projects=@strapi/strapi
yarn build:types --projects=@strapi/strapi
```

### Related issue(s)/PR(s)

Fixes #23688  
Caused by #23421  
Additional bug identified by @WDeenik in https://github.com/strapi/strapi/pull/23421#discussion_r2207601028

### DIFF
--- a/packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.test.ts
+++ b/packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.test.ts
@@ -42,6 +42,7 @@ jest.mock('@strapi/data-transfer', () => {
         ...acutal.strapi.providers,
         createLocalStrapiSourceProvider: jest.fn().mockReturnValue({ name: 'testLocalSource' }),
         createLocalStrapiDestinationProvider: jest.fn().mockReturnValue({ name: 'testLocalDest' }),
+        createRemoteStrapiSourceProvider: jest.fn().mockReturnValue({ name: 'testRemoteSource' }),
         createRemoteStrapiDestinationProvider: jest
           .fn()
           .mockReturnValue({ name: 'testRemoteDest' }),


Bassel17:
Thanks for the contribution, the code looks good !
but can you take a look at the build issues ?

jamesvillarrubia:
Whoops.  Missed a TS lint issue.

PR #24881, opened 2025-11-19
fix: `env` helper default and generic types

### LINKED ISSUE #23611: [bug]: env helper types incorrectly nullable even with default value provided

ericbf:
This commit solves an issue with the `env` helper where nullable types are returned even when default values are passed. It should not return a nullable value when a default is passed. This fixes all functions in the env helper to conform to the correct types. Additionally this PR cleans up some generics that didn’t make sense in a few of the helpers (the base `env` (`string`) helper narrowing to a literal string type when using a default, `oneOf` helper not using a literal type for the valid values, etc).

### What does it do?

This fixes the typescript types of the `env` helper so that if you pass a default value, it does not return a nullable value.

### Why is it needed?

When using the env helper with a default value, you would have to cast it as non-nullable when using to a prop that was required, or you’d get a typescript error.

### How to test it?

Check the returned type of the `env` helper if passing a default. None of the returned types should be nullable when a non-nullable default value is passed. Ensure all types are correct.

### Related issue(s)/PR(s)

Fixes #23611


Bassel17 (approved):
Thank you for the contribution !
All looks good now !

PR #24890, opened 2025-11-20
fix: globe icon is shown even if field is not localized

mathildeleg:
### What does it do?
* Check if the parent field (dynamic zone or component) is set as localized, to set the globe icon correctly if needed
* Make sure the field is correctly displayed as localized or non-localized even if used several times in the same content type

### Why is it needed?
The Globe icon was displayed even if a field was not localized/no longer localized. 
**BEFORE**
_The Globe shouldn't be displayed for the first Relation component & for the last Dynamic zone_
screencapture-localhost-1337-admin-content-manager-collection-types-api-author-author-uktbks0e448q2xwd1jx11znq-2025-11-20-11_41_41
https://github.com/user-attachments/assets/6b639a9e-a6d6-44a4-8432-c01f48da454d

**AFTER**
_The Globe is displayed at the root-level of elements_
Capture d’écran 2025-12-15 à 16 59 22

### How to test it?
1. Repeatable or non-repeatable components:
- Add a repeatable or non-repeatable component to a localized Content Type
- Make sure localization is enabled for this component
- Go to the Content Manager for this Content Type and click on "Create new entry"
- Add a component for that entry, see the Globe icon
- Go back to the Content Type Builder and remove the localization for this component
- Go back to the Content Manager for this Content Type and click on "Create new entry" or go to the one you previously created
=> the Globe icon should no longer be displayed for the fields inside that component, since localization was disabled

2. Using a component more than once in a Content Type
- Add a repeatable or non-repeatable component to a localized Content Type, for example the component *Social links*
- Set this component to non-localized
- Add a Dynamic Zone with that same component to a localized Content Type
- Make sure localization is enabled for this Dynamic Zone
- Go to the Content Manager for this Content Type and click on "Create new entry"
- Add *Social links* and see that the component itself doesn't show the Globe icon, but the fields inside that component do
- Add a Dynamic Zone with the component *Social links*
=> The fields inside the *Social links* component (repeatable or not) shouldn't display the Globe icon

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/issues/23723

### DIFF
--- a/tests/e2e/tests/content-manager/history.spec.ts
+++ b/tests/e2e/tests/content-manager/history.spec.ts
@@ -256,7 +256,9 @@ describeOnCondition(edition === 'EE')('History', () => {
       await page.waitForURL(ARTICLE_CREATE_URL);
       await page.getByRole('textbox', { name: 'title' }).fill('Being from Kansas');
       await page
-        .getByRole('textbox', { name: 'slug This value is unique for the selected locale' })
+        .getByRole('textbox', {
+          name: 'slug',
+        })
         .fill('being-from-kansas');
       await page.getByRole('button', { name: 'Save' }).click();
       await page.waitForURL(ARTICLE_EDIT_URL);


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/NonRepeatable.tsx
```diff
@@ -64,6 +66,13 @@ const NonRepeatableComponent = ({
                     defaultMessage: field.label,
                   });
 
+                  const clonedLabelAction = React.isValidElement(field.labelAction)
```
mathildeleg: Open to suggestions/thoughts on this 🙏🏻

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +103,59 @@ const Span = styled(Flex)`
   }
 `;
 
+/**
+ * Compute effective localization for a specific field occurrence based on:
+ * - Content type i18n setting
+ * - Ancestor attribute (dynamic zone or component) i18n setting
+ * - Field-level i18n setting for root fields
+ */
+const getEffectiveLocalization = (
```
markkaylor: What is meant by "effective" 🤔 , if it returns a boolean I personally like to follow the convention of `isX` `hasX`, here I'm wondering if something like `isLocalizedField` might be a better name?

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +103,59 @@ const Span = styled(Flex)`
...
+};
+
+const getPluginOptions = (attr: EditFieldLayout['attribute']) => {
+  return attr && typeof attr === 'object' && 'pluginOptions' in (attr as object)
+    ? (attr as { pluginOptions?: unknown }).pluginOptions
+    : undefined;
+};
+
+const findTopLevelField = (edit: EditLayout, name: string) => {
+  return edit.layout.flat(2).find((f) => f.name === name) ?? undefined;
```
markkaylor: Why is the flat depth 2? Could it ever be deeper than that?

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +103,59 @@ const Span = styled(Flex)`
...
+ * Compute effective localization for a specific field occurrence based on:
+ * - Content type i18n setting
+ * - Ancestor attribute (dynamic zone or component) i18n setting
+ * - Field-level i18n setting for root fields
+ */
+const getEffectiveLocalization = (
+  name: string,
+  attribute: EditFieldLayout['attribute'],
+  edit: EditLayout
+): boolean => {
```
markkaylor:
```suggestion
) => {
```
 
I'm not sure we need to explicitly type the return value here the boolean should be inferred

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +103,59 @@ const Span = styled(Flex)`
...
+    return i18n?.localized === true;
+  }
+  return false;
+};
+
+const getPluginOptions = (attr: EditFieldLayout['attribute']) => {
+  return attr && typeof attr === 'object' && 'pluginOptions' in (attr as object)
+    ? (attr as { pluginOptions?: unknown }).pluginOptions
+    : undefined;
+};
```
markkaylor: I wonder if these helper functions are necessary, isn't this the equivalent:

```ts
const isLocalized = Boolean(attribute?.pluginOptions?.i18n?.localized);
```

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -48,61 +49,39 @@ const mutateEditViewHook = ({ layout }: MutateEditViewArgs): MutateEditViewArgs
...
 }
 
-const LabelAction = ({ title, icon }: LabelActionProps) => {
+const LabelAction = ({ name, attribute }: LabelActionProps) => {
   const { formatMessage } = useIntl();
+  const { slug: model } = useParams<{ slug: string }>();
+  const { edit } = useDocumentLayout(model!);
+
+  const isLocalized = React.useMemo<boolean>(() => {
+    return getEffectiveLocalization(name, attribute, edit);
```
markkaylor: Is the useMemo necessary? The function just determines if the attribute has i18n enabled so I don't think it's a very expensive operation

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +103,59 @@ const Span = styled(Flex)`
...
+  return false;
+};
+
+const getPluginOptions = (attr: EditFieldLayout['attribute']) => {
+  return attr && typeof attr === 'object' && 'pluginOptions' in (attr as object)
+    ? (attr as { pluginOptions?: unknown }).pluginOptions
+    : undefined;
+};
+
+const findTopLevelField = (edit: EditLayout, name: string) => {
```
markkaylor: I was a bit confused here, I think I understand what it does. If you are working with an attribute inside a component or dz you want to get _that_ component or dz attribute object. Is that right?

What do you think of renaming this `getAttributeParent`

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
markkaylor: Flatten 1 level deep for a component, flatten 2 levels deep for dz?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/NonRepeatable.tsx
```diff
@@ -64,6 +66,13 @@ const NonRepeatableComponent = ({
...
+                    ? React.cloneElement(field.labelAction as React.ReactElement, {
+                        name: completeFieldName,
+                        attribute: field.attribute,
+                      })
+                    : field.labelAction;
```
markkaylor: cloneElement is not recommended: https://react.dev/reference/react/cloneElement

I wonder if a render prop could work here though

```ts
const labelAction =
  typeof field.labelAction === 'function'
    ? field.labelAction({
        name: completeFieldName,
        attribute: field.attribute,
      })
    : field.labelAction;
```

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -48,61 +49,39 @@ const mutateEditViewHook = ({ layout }: MutateEditViewArgs): MutateEditViewArgs
...
- * LabelAction
- * -----------------------------------------------------------------------------------------------*/
+const addLabelActionToField = (field: EditFieldLayout) => ({
+  ...field,
+  labelAction: <LabelAction name={field.name} attribute={field.attribute} />,
+});
 
 interface LabelActionProps {
-  title: MessageDescriptor;
-  icon: React.ReactNode;
```
markkaylor, line 95: So before it was never receiving / using these props?

mathildeleg: It was but now I'm setting title on L73, and the icon was always the Earth icon so it's set L82

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/NonRepeatable.tsx
mathildeleg: Unfortunately this doesn't work exactly like that, this is what I came up with: https://github.com/strapi/strapi/pull/24890/commits/4bb26bc50eca53ee641fa8fdab4ec223a0bdf09d
But open to your feedback though! I don't know if this is the best, but your suggestion wasn't working

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
mathildeleg: Decided to rename it to `getAttributeParentField`, is that clearer?

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
mathildeleg: Good point, I removed it

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
mathildeleg: Unfortunately that doesn't work because of typing issues :( The only way I found was by typing this way, and I decided to split it into two separate functions to make it more readable, but I'm open to other ideas if you have them!

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
mathildeleg: No it's because edit.layout is 3-levels deep: panel → row → fields, and we need to access the fields to set the icon to each field specifically.

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
mathildeleg: I chose `isFieldLocalized`, is that better?

mathildeleg:
> Thanks for this one! Haven't tested it yet, but I noticed in the screenshot if it's localized we show the globe and if it's not localized we show nothing. Makes sense to me but at one point I think we had a globe icons with a slash (/) through it. Maybe just check with product.

Hopefully I've answered your feedback correctly here https://github.com/strapi/strapi/pull/24890/commits/4bb26bc50eca53ee641fa8fdab4ec223a0bdf09d 
Don't hesitate if you have better ideas though 🙏🏻
I'll ask Product about the icon but I think that was in V4 and not V5

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +99,59 @@ const Span = styled(Flex)`
...
+  if (!contentTypeLocalized) {
+    return false;
+  }
+
+  const path = name.split('.');
+  const topLevelAttrName = path[0];
+  const topLevelField = getAttributeParentField(edit, topLevelAttrName);
+  if (!topLevelField) {
+    return false;
+  }
```
markkaylor: On line 136 you use attribute instead of topLevelField.attribute. Could we ever be in a case where we don't have a topLevelField and we still want to return like 136? If not then maybe it would read more clearly by always referencing the attribute on `topLevelField`? It would also simplify the function signature since attribute would no longer be needed.

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
markkaylor: It's completely up to you. If you prefer the helper functions that's fine with me. That being said I think we are safe to cast this since we know the shape we expect and we know localized will always return a boolean or be undefined. 

```
const pluginOptions: { i18n?: { localized?: boolean } } = attribute?.pluginOptions ?? {};
const isAttributeLocalized = Boolean(pluginOptions.i18n?.localized);
```

packages/plugins/i18n/admin/src/contentManagerHooks/editView.tsx
```diff
@@ -124,4 +99,59 @@ const Span = styled(Flex)`
...
+};
+
+const getPluginOptions = (attr: EditFieldLayout['attribute']) => {
+  return attr && typeof attr === 'object' && 'pluginOptions' in (attr as object)
+    ? (attr as { pluginOptions?: unknown }).pluginOptions
+    : undefined;
+};
+
+const getAttributeParentField = (edit: EditLayout, name: string) => {
+  return edit.layout.flat(2).find((f) => f.name === name) ?? undefined;
```
markkaylor: Unless you have a specific reason for the `??` it feels unnecessary to me. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

> If no values satisfy the testing function, [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) is returned.

markkaylor (approved):
It works perfectly! Nice work Mathilde!

PR #24798, opened 2025-11-07
fix(db): replace discardDrafts calls in v5 migration with queries

innerdvations:
### What does it do?

Drop-in replacement of the discard drafts migration for Strapi v4 -> v5 -- that is, if you already completed that one, this one will not need to run again.

### Why is it needed?

https://github.com/strapi/strapi/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20discard%20drafts

Performance was so bad before that it was a blocker for projects with extemely large datasets (millions of records) that would often run for 20+ hours or run out of memory from the transaction

There were also issues with connection pools growing too large that should not be present anymore.

This should also fix issues with missing self relations in components (#24544) 

### How to test it?

1. Start with a Strapi v4 project
2. Fill it with complex data types and relations, in particular:
- draft and publish related to non-d&p and vice versa
- self references
- components (including with items with self references)
3. At this point, be sure to back up your database and code so you can test again later!
4. Upgrade to this branch (experimental coming soon)
5. Start up Strapi v5 and check that all of your data was migrated successfully and is visible in the admin with the correct publication status, and the relations all with the correct publication status

See the branch `chore/example-projects-migrations` for tools to do that more easily

Experimental: `0.0.0-experimental.a1edcc9d9fab426bad8fd1a111e6d06cb07849be`

### Related issue(s)/PR(s)

Replaces https://github.com/strapi/strapi/pull/24545

Almost everything here:
https://github.com/strapi/strapi/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20discard%20drafts

fixes #24488
fixes #23812 
fixes #22963 
fixes #23401
fixes https://github.com/strapi/strapi/issues/24544

What it will not solve:

- it may not solve out of memory issues caused by the transaction. for that, we will probably need to add a new (experimental/somewhat dangerous) flag to run migrations without a transaction

PR #24894, opened 2025-11-20
fix: discard draft creates published comp relations correctly

innerdvations:
### What does it do?

Essentially, when a published component had a relation to a published entry,  after migration it had a relation to both the published and draft version of that entry, and then the published link was removed by a db repair script, leaving the incorrect relation. This ensures that only the correct relation is created.

### Why is it needed?

Fixes the issue described in https://github.com/strapi/strapi/pull/24798#issuecomment-3556822284

### How to test it?

Create a v4 project with a draft and publish content type with a published item with a component that has a relation to another published item. Migrate to v5 and start up the project, and the relations should all still exist in the database.

See `chore/example-projects-migrations` for tools to make testing this easier, including a seed and validation script that includes a check for this case.

### Related issue(s)/PR(s)

Alternative to https://github.com/strapi/strapi/pull/24893 which reverts the migration to the previous version rather than fixing it


innerdvations:
Closing this until I have time to look at it more closely

Issue #23480, opened 2025-05-06
Support `exactOptionalPropertyTypes` compiler flag in TypeScript


unrevised6419:
> [!NOTE]
> This is a [generic issue template](https://gist.github.com/unrevised6419/8bdfca183fde610708685a60c2510b1e) to raise awareness about `exactOptionalPropertyTypes` support across TypeScript libraries.

## Summary

This package's types are not fully compatible with TypeScript's [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) compiler flag (`tsconfig.json`).

## Problem

With `exactOptionalPropertyTypes: true`, optional properties behave differently — they can’t be assigned `undefined` unless it's explicitly part of the type. Currently, this library's types treat optional properties as implicitly allowing `undefined`, which causes type errors in strict setups.

## Example

```ts
// Type from this library
type Example = {
  config?: SomeConfig;
};

// Usage
const x: Example = {
  config: undefined, // ❌ Error with exactOptionalPropertyTypes
};
```

## Expected

```ts
// Explicitly allow undefined if needed
type Example = {
  config?: SomeConfig | undefined;
};
```

## Suggested Fix

Audit type definitions and add `| undefined` where intentional. This helps projects using stricter TypeScript settings adopt this library without patching types.

#### ✅ Use only `?` (e.g. `prop?: T`)

- Use when the property can be **omitted entirely**.
- Do **not** expect consumers to explicitly pass `undefined`.
- Example: config objects where the value is optional and can be left out.

```ts
type Options = {
  retry?: boolean;
};
```

#### ✅ Use only `| undefined` (e.g. `prop: T | undefined`)

- Use when the property is required to exist but the value may be `undefined`.
- Useful when the property is always present, but its value is unset or to be determined later.

```ts
type Response = {
  data: string | undefined; // must exist, but may be undefined
};
```

#### ✅ Use both `?: T | undefined`

- Use when the property is optional and may explicitly be set to `undefined`.
- Required if consumers may want to pass `undefined` explicitly to override a default or signal absence.

```ts
type Config = {
  timeout?: number | undefined;
};
```

### 📢 Suggestion for Library Authors

To proactively avoid these issues and align with modern TypeScript practices, consider enabling the `exactOptionalPropertyTypes` flag in your own `tsconfig.json`. This helps:

- Enforce more **accurate typing** of optional properties
- Prevent **unintentional `undefined` assignments**
[...]

PR #24905, opened 2025-11-21
fix(types): adding missing config types for server and admin

Bassel17:
### What does it do?

Adds comprehensive TypeScript type definitions for missing Strapi configuration options in `packages/core/types/src/core/config/`:

### Why is it needed?

These configuration options are documented and actively used throughout the Strapi codebase but were missing proper TypeScript type definitions. This caused:
- Lack of IntelliSense/autocomplete for developers configuring Strapi projects
- No type safety when accessing these config values
- Difficulty discovering available configuration options

This PR addresses the types gap identified in the community PR #21859 and resolves the issue preventing its merge.

### Related issue(s)/PR(s)
allows us to merge #21859  and make so configs types are available for projects


packages/core/types/src/core/config/admin.ts
```diff
@@ -2,15 +2,61 @@ export interface ApiTokenProp {
   salt: string;
 }
 
+export interface AuthSessionsProp {
+  options?: {
+    algorithm?: string;
+    [key: string]: any;
```
unrevised6419: Suggestion: Won't be better for all indexed access objects to have the type `unknown` instead of `any`? 👀

packages/core/types/src/core/config/admin.ts
```diff
@@ -2,15 +2,61 @@ export interface ApiTokenProp {
...
+  };
+  accessTokenLifespan?: number;
```
unrevised6419: Suggestion: Can we add also `undefined` for the optional fields? This will make it compatible with the TypeScript `exactOptionalPropertyTypes` flag, when someone want's to explicitly pass `undefined`

```suggestion
  accessTokenLifespan?: number | undefined;
```

https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes

Context: https://github.com/strapi/strapi/issues/23480

innerdvations:
Testing out the Claude reviewer. I 'resolved' most of the useless comments, a few of the remaining ones looked like they might have some validity. I'll review soon to see for sure 😆

innerdvations:
@cache-your-dreams We may want to update the AI review to not suggest approving when it thinks there are critical bugs and instead say something like "Changes required"

packages/core/types/src/core/middleware.ts
```diff
@@ -12,9 +12,9 @@ export type MiddlewareFactory<T = any> = (
 export type MiddlewareName = UID.Middleware | string;
 
 export type MiddlewareConfig = {
-  name?: MiddlewareName;
-  resolve?: string;
-  config?: unknown;
+  name?: MiddlewareName | undefined;
+  resolve?: string | undefined;
+  config?: unknown | undefined;
```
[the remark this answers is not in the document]
unrevised6419: For this to not be reported, I think `exactOptionalPropertyTypes` tsconfig flag needs to be activated for development only.

`?:` and `| undefined` is important for consuers having this flag activated, otherwise they will get errors

https://github.com/strapi/strapi/issues/23480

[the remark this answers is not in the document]
unrevised6419: PS: Users nowadays creating a new TS project they will get `exactOptionalPropertyTypes` enabled by default in their config, so this problem might appear more and more if both optional and undefined are not used.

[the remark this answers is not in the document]
unrevised6419: PS2: But for Strapi to enable this flag may be complicated, because the same problem will be encountered with upstream libraries (e.g. Vite)

packages/core/types/src/core/config/admin.ts
```diff
@@ -1,60 +1,144 @@
...
 
-export interface AuthProp {
+export interface AuthSessions {
+  options?:
+    | {
+        algorithm?: string | undefined;
+        [key: string]: unknown;
+      }
+    | undefined;
+  accessTokenLifespan?: number | undefined;
```
unrevised6419, line 10: Suggestion: Would be better to disable sonar reporting of rule key `typescript:S4782`

https://sonarcloud.io/organizations/strapi/rules?open=typescript%3AS4782&rule_key=typescript%3AS4782

**sonar-project.properties**

```properties
sonar.issue.ignore.multicriteria=e1
sonar.issue.ignore.multicriteria.e1.ruleKey=typescript:S4782
```

packages/core/admin/server/src/services/api-token.ts
```diff
@@ -155,10 +156,9 @@ const exists = async (whereParams: WhereParams = {}): Promise<boolean> => {
  * Return a secure sha512 hash of an accessKey
  */
 const hash = (accessKey: string) => {
-  return crypto
-    .createHmac('sha512', strapi.config.get('admin.apiToken.salt'))
-    .update(accessKey)
-    .digest('hex');
+  const apiTokenCfg = strapi.config.get<Core.Config.Admin['apiToken']>('admin.apiToken');
+  const salt = apiTokenCfg?.salt;
```
innerdvations: We should throw an error `if(!salt)`

Bassel17: yeah it's a bit of weird one for this one, because we already do throw an error on bootstrap if the salt is not defined, but checking here again is not wrong so why not

packages/core/types/src/core/middleware.ts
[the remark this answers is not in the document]
Bassel17: yep am still unsure about this, but will keep it in mind for a bit more internal discussions, this comment here by the AI is true we don't have `exactOptionalPropertyTypes` enabled in the monorepo and we don't enable it in the typescript projects template either ?
do you mean that if `exactOptionalPropertyTypes` is not set it will be considered default true in newer typescript versions ?

packages/core/admin/server/src/services/api-token.ts
Bassel17: actually it might be a bit redundant to check again, I removed the optional `?` the value can't be undefined

unrevised6419: Probably this can be expressed in types 🤔 Like having "configs" and "resolved configs".

innerdvations:
Looks like some tests need to be updated, but overall LGTM!

@Convly if you have a few minutes would you be able to give it a quick review in case there's potentially a breaking change I might have missed?

packages/core/types/src/core/config/admin.ts
```diff
@@ -31,30 +77,64 @@ export interface RateLimitProp {
...
+export interface PreviewHandlerParams {
+  documentId: string;
+  locale?: string;
+  status?: string;
+  [key: string]: unknown;
+}
+
+export interface PreviewConfig {
+  allowedOrigins?: string[];
+  handler: (uid: string, params: PreviewHandlerParams) => string | null | undefined;
```
daliuspamedytis1393, line 102: The docs show preview.config.handler as async (https://docs.strapi.io/cms/features/preview#preview-handler), but the types define handler as synchronous and only allow string | null | undefined. This causes a TS error when implementing the async handler. Is this a known mismatch, or should the docs/types be aligned?

daliuspamedytis1393: @Bassel17 @innerdvations

unrevised6419: Here is a fix https://github.com/strapi/strapi/pull/25396

daliuspamedytis1393: Thanks bro

PR #24907, opened 2025-11-21
chore(admin): guided tour should only display in development

markkaylor:
### What does it do?

Only displays the guided tour in development env

### Why is it needed?

The guided tour does not make sense in production since:
- It is only intended for the first super admin user
- The CTB cannot be accessed which is crucial to the guided tour

### How to test it?

Clear local storage by running this in browser console `localStorage.removeItem('STRAPI_GUIDED_TOUR')`
Build and run in production env
```
NODE_ENV=production yarn build && NODE_ENV=production yarn start
```
You should not see the guided tour on the hompeage or any of its tooltips on the concerned pages (ctb, cm, api tokens)

Ensure it still works in development as expected

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/admin/admin/src/pages/ProfilePage.tsx
+++ b/packages/core/admin/admin/src/pages/ProfilePage.tsx
@@ -230,7 +230,7 @@ const ProfilePage = () => {
             </>
           )}
         </Form>
-        {isDesktop && (
+        {isDesktop && process.env.NODE_ENV === 'development' && (
           <Box>
             <Layouts.Content>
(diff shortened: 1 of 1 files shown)


Adzouz:
Makes sense 👀
I tested but I still can see the reset guided tour option in the profile, could you also remove it there?

markkaylor:
> Makes sense 👀 I tested but I still can see the reset guided tour option in the profile, could you also remove it there?

@Adzouz good catch, updated

PR #24909, opened 2025-11-21
chore: integrate subscription info in backend trackers

maccomaccomaccomacco:
### What does it do?

Integrate subscription info in backend trackers.

### Why is it needed?
It might be useful to track subscription data on events for better filtering.

### How to test it?
- run the tracker locally with getstarted project pointing at it and a license active
- tweak the tracker to log the incoming events
- check the events groupProperties, they should include subscription info like `planPriceId` and `subscriptionId`

### DIFF
--- a/packages/core/core/src/services/metrics/sender.ts
+++ b/packages/core/core/src/services/metrics/sender.ts
@@ -70,6 +70,14 @@ export default (strapi: Core.Strapi): Sender => {
   return async (event: string, payload: Payload = {}, opts = {}) => {
     const userId = generateAdminUserHash(strapi);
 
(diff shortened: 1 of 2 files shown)


packages/core/core/src/services/metrics/sender.ts
```diff
@@ -81,6 +81,7 @@ export default (strapi: Core.Strapi): Sender => {
         groupProperties: {
           ...anonymousGroupProperties,
           projectType: strapi.EE ? 'Enterprise' : 'Community',
+          ...strapi.ee,
```
jhoward1994: Do we want to copy to entire ee object into here? I think it might be better to extract the specific properties we want to be sent ?

PR #24917, opened 2025-11-24
fix(upload): preserve provider_metadata during file replacement

fanvyr:
### What does it do?

This PR preserves the existing `provider_metadata` during the Media Library “Replace” action.

When a file is replaced, Strapi rebuilds a fresh `fileData` object (through validation and optional image processing), which removes the previous metadata stored by the provider.  
This metadata is often required by providers to compute upload paths, identify storage locations, or perform other internal operations.  

The PR adds a minimal one-line fix to ensure the old metadata is copied into the new `fileData` before the provider's `upload()` method is invoked.

### Why is it needed?

Some upload providers store additional internal information inside `provider_metadata`.  
Losing this information during replacement can cause inconsistent behavior compared to the initial upload, such as:

- uploading the replacement file to a different location  
- losing provider-specific routing information  
- inconsistencies in how formats are handled  

By preserving the metadata, the replacement workflow behaves consistently with the original upload for all providers that rely on this field.

### How to test it?

1. Upload a file in the Media Library using any provider that sets `provider_metadata`.
2. Replace the file using the built-in “Replace media” action.
3. Before this change: the replacement file may not retain the metadata the provider originally stored.
4. After this change: `provider_metadata` remains intact, ensuring consistent behavior between initial upload and replacement.

### Related issue(s)/PR(s)

No direct issue found, but this improves robustness for any upload provider using `provider_metadata`.
(Well, my provider plugin actually ;-) )


fanvyr:
hm....

markkaylor (dismissed):
Looks good to me, added a test

PR #24920, opened 2025-11-24
enhancement(admin): add uz-Cyrl native name to languageNativeNames

Alirahmon777:
I updated the _languageNativeNames.ts_ file to include the Uzbek Cyrillic locale by adding the key "uz-Cyrl" with the native label "Ўзбекча". This ensures that Uzbek Cyrillic is correctly shown in the language selection dropdown in the Strapi admin panel.


innerdvations (approved):
Thanks!

innerdvations (dismissed):

innerdvations (approved):

Issue #22385, opened 2024-12-10
`alternativeText` and `caption` field for files are unnecessarily limited to 255 characters


phil-janeapp:
### Required System information

- Node.js version: v20.17.0
- NPM version: N/A
- Strapi version: 5.1.1
- Database: postgres
- Operating system: Linux
- Is your project Javascript or Typescript: Typescript

### Describe the bug

May be fixed in more recent version of v5, but the `alternativeText` and `caption` field for files are set to a 255 character limit on postgres (and possibly other datastores).  

Based on some quick research I did, it would appear that there's actually no browser/spec/screen reader limit for `alt` text for image, and that seems like the most-likely place for this value to be used.

### Steps to reproduce the behavior

1. Go to the media library or upload a file directly into a content-type
2. Edit the alternative text or caption to be longer than 255 characters
3. Submit the changes

### Expected behavior

That both the alternative text and caption fields support text longer than 255 characters

spyderdsn:
I agree and faced the same problem. Please extend these fields by default.

PR #24931, opened 2025-11-26
fix: change several file field types from string to text

derrickmehaffy:
### What does it do?

Changes several of the text based fields in the file content-type from the upload package from string to text

### Why is it needed?

string based fields in most databases limit the total characters of the value to 255, text (which converts to the `longtext` in some databases) have practically no limit. This is useful especially for URLs where they can easily exceed this limit.

### How to test it?

Double check the database to ensure column type changes without data loss and that you can have file type URLs, file names, ect that exceed 255 characters

### Related issue(s)/PR(s)

fixes: https://github.com/strapi/strapi/issues/22385
Also fixes another issue but I couldn't find it, cc @christophecvb

### DIFF
--- a/packages/core/upload/server/src/content-types/file.ts
+++ b/packages/core/upload/server/src/content-types/file.ts
@@ -22,16 +22,16 @@ export default {
     },
     attributes: {
       name: {
-        type: 'string',
+        type: 'text',
         configurable: false,
         required: true,
       },
       alternativeText: {
-        type: 'string',
+        type: 'text',
         configurable: false,
       },
       caption: {
-        type: 'string',
+        type: 'text',
         configurable: false,
       },
       width: {
@@ -66,12 +66,12 @@ export default {
         required: true,
       },
       url: {
-        type: 'string',
+        type: 'text',
         configurable: false,
         required: true,
       },
       previewUrl: {
-        type: 'string',
+        type: 'text',
         configurable: false,
       },
       provider: {


derrickmehaffy:
experimental: https://github.com/strapi/strapi/actions/runs/19715261448

ChristopheCVB:
A big thanks to you @derrickmehaffy !
I didn't have the chance to open the issue you're referring to, basically I was hitting the 255 limit on the `url` field of the `files` table on a PostgreSQL instance.
I hit that limit because I wanted to store the signedUrl in database for a custom provider I made, which also could detect if it needs to regenerate a new signedUrl based on the query string params of the url stored, but couldn't do it because of the limit.
Those change will enable much more cleaner signedUrl management!

packages/core/upload/server/src/content-types/file.ts
```diff
@@ -22,16 +22,16 @@ export default {
...
     attributes: {
       name: {
-        type: 'string',
+        type: 'text',
         configurable: false,
         required: true,
       },
       alternativeText: {
-        type: 'string',
+        type: 'text',
```
innerdvations, line 30: This one I wouldn't change, having alt text longer than 255 is very bad practice so I wouldn't encourage it unless we need to

hanpaine:
FYI @oiorain & @paridhidixit-create

packages/core/upload/server/src/content-types/file.ts
markkaylor: I agree, from the research I did for the ai altText and caption generation the SEO recommendation is max 125 characters for altText and max 160 characters for caption. 

That being said I'm really not sure if we should apply the constraint at the database level or even at all, maybe it should be an improvement in the FE to display a warning to the user in the UI.

derrickmehaffy: I specifically changed those to fix the linked bug. I don't think it's our place to determine what should be allowed there but instead we should have schema override ability on the file content type to let users set their own min/max.

First step before that is just removing the limits.

derrickmehaffy: I specifically changed those to fix the linked bug. I don't think it's our place to determine what should be allowed there but instead we should have schema override ability on the file content type to let users set their own min/max.

First step before that is just removing the limits.

PR #24953, opened 2025-12-02
fix(*): migrate IDs from int to bigint

### LINKED ISSUE #24491: Critical Overflow Risk: Component ID Increments on Publish
### Node Version

22.0.0

### NPM/Yarn/PNPM Version

NPM

### Strapi Version

5.24.0

### Operating System

Linux (Debian/Ubuntu)

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

The `id` for a given component type auto-increments within its own isolated pool. All instances of a single component (for example, every "Link Component" used across the entire website) share the same ID counter. A new `id` is consumed from that specific component's pool every time an entry containing it is published.

The data type for this `id` is an **`unsigned int`**(on MySQL), which has a hard limit of 4,294,967,295. 

This becomes a critical issue on large-scale websites with **thousands of pages across multiple locales**. These sites often rely on **automated localization syncs**, which can trigger a **publish event daily for most or all entries**. While this affects all components, the impact is most severe for highly granular ones. A **"Link Component"**, used dozens of times per page, would see its dedicated ID counter increase at an extremely high rate from these daily syncs.

This relentless, automated process puts the most common components on an accelerated path to exhausting their ID limit, creating a critical and foreseeable overflow risk.

### Steps to Reproduce

1.  **Set up a Strapi project**.
2.  **Create two reusable Components**: a `LinkComponent` and a `SeoBlock`.
3.  **Create a Collection Type** (e.g., `Article`) that can use both components.
4.  **Create and publish** a single `Article`. Add one `LinkComponent` and one `SeoBlock`. Observe their initial IDs (e.g., `LinkComponent` is `id: 1`, `SeoBlock` is `id: 1`).
5.  **Edit the `Article`**. Change a field inside the `LinkComponent` (e.g., its label) and republish.
6.  **Check the network tab**. You will see the `LinkComponent` was updated with `id: 2`, and importantly, the `SeoBlock` was also updated with `id: 2`, even though it was not touched.
7.  **Edit the `Article` again**. This time, change a field inside the `SeoBlock` and republish.
8.  **Check the database again**. You will see the `SeoBlock` was updated with `id: 3`, and the `LinkComponent` was also updated with `id: 3`.

### Expected Behavior

The primary key for the component should not change on publish.

OR

The primary key for any given component type should be scalable enough to handle a virtually infinite number  of instances over the application's lifetime. A widely-used component should not face a risk of data overflow due to routine, high-frequency publishing workflows.

### Logs

_No response_

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

nclsndr:
## What does it do?

This PR migrates all ID columns (primary keys and foreign keys) from `INTEGER` to `BIGINT` across the database layer. This change ensures better support for large-scale applications and prevents integer overflow issues.
To account for backward compatibility, the `bigint` values are casted in Number up to 2^53, throw an exception if overflowing [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).

> [!NOTE]
> This PR is currently meant only to trigger CI checks in order to help identify regressions.

#### Database migration

- **Automatic bigint migration**:
  - New internal migration (`5.0.0-06-migrate-ids-from-int-to-bigint`) upgrades:
    - **Primary keys** (`id` columns) in all content-type and internal tables.
    - **Foreign keys** in join/morph tables and relation tables.
  - Handles PostgreSQL, MySQL, and SQLite with dialect-specific logic:
    - Safely drops foreign key constraints before conversion, recreates them afterwards.
    - Uses table recreation where needed (e.g. SQLite’s limited `ALTER TABLE` support).
  - Skips columns that are already bigint-compatible, so it can safely run on existing projects.

#### Transparent bigint handling

- **Internal vs user-defined fields**:
  - Introduces an `internalIntegerId` flag on attributes to mark Strapi-created bigint FKs in join tables.
  - Distinguishes three cases:
    - **ID columns** (`type: 'increments'` or `type: 'bigincrements'`) → exposed as **numbers**.
    - **Internal FK columns** (`type: 'biginteger'` + `internalIntegerId: true`) → exposed as **numbers**.
    - **User-defined `biginteger` attributes** (no flag) → exposed as **strings**.
- **Safety and precision**:
  - Internal IDs/FKs are converted from DB strings to JS numbers and validated against `Number.MAX_SAFE_INTEGER`.
  - If an internal ID overflows the safe range or cannot be parsed, Strapi throws a clear error instead of silently corrupting data.
  - User-defined bigintegers are never clamped or validated against `MAX_SAFE_INTEGER` and keep their full string precision.

#### Schema & tooling updates

- **Schema builder & inspector**:
  - New content types and internal tables use `bigincrements` for IDs by default; join-table FKs use `biginteger`.
  - Dialects (MySQL, Postgres, SQLite) recognize bigint PK/FK columns and map them to the correct Strapi types.
- **Diffing & migrations**:
  - Schema diff treats `increments` and `bigincrements` as equivalent to avoid noisy, unnecessary migrations.
  - Internal migration and schema tools are aligned so that once IDs are bigint, subsequent syncs do not try to downgrade/flip types.

### What this changes

- **ID storage in the database**
  - All Strapi‑managed ID columns (PKs and FKs, including join/morph tables and internal tables) are now stored as `BIGINT` in supported databases.
  - A new internal migration walks all tables and:
    - Upgrades integer `id` PKs to bigint.
    - Upgrades integer FK columns that point to those IDs to bigint.
    - Handles FK constraints per dialect (Postgres/MySQL/SQLite) so the migration is consistent and safe.

- **How IDs are exposed in JavaScript**
  - **Internal IDs and FKs (Strapi‑created)**:
    - Still exposed as **numbers** in Node for typical ID ranges.
    - On read, values are validated to be within `Number.MAX_SAFE_INTEGER`; anything larger throws a clear error instead of silently losing precision.
  - **User‑defined `biginteger` attributes**:
    - Still exposed as **strings** (unchanged behavior), including very large values beyond JS safe integer range.
    - This preserves existing app logic that treats those values as strings.

- **Schema & metadata model**
  - Internal models (content types, join tables, history/core tables, migrations/schema tables) are updated to:
    - Use `bigincrements` for their `id` fields.
    - Use `biginteger` for internal FK columns.
  - The schema inspector and diffing logic are updated so:
    - PKs using `increments` vs `bigincrements` are treated as equivalent from a migration‑noise point of view.
    - Dialect‑specific quirks (e.g. SQLite’s lack of a real bigint type) are abstracted behind the dialect API.

### Scope in the codebase

- **Database layer**
  - Dialects (MySQL, Postgres, SQLite) schema inspectors and type mapping.
  - Schema builder, diffing, and internal migration storage/schema tables.
  - Field system, especially the bigint field and how it converts DB values to JS values.

- **Metadata and relation generation**
  - Metadata fixtures/expected outputs.
  - Relation/join table builders (including morph and many‑to‑many join tables).

- **Core / CM models and services**
  - Internal models that use Strapi‑managed IDs (history versions, core store, webhooks, etc.).
  - Utilities that transform content types into database models.

- **Tests**
  - New tests around bigint behavior and integration with the transform pipeline.
  - Existing schema/metadata tests updated to expect bigint IDs and FK columns.

### Backwards compatibility

- **Stable for existing apps**:
  - For “normal” ID ranges (i.e. well below \(2^{53}\)), behavior is unchanged: IDs are still numbers in JavaScript.
  - User‑defined `biginteger` fields remain strings and retain full precision; no breaking change there.
- **Stronger guarantees**:
  - If the underlying DB ever stores an internal ID beyond `Number.MAX_SAFE_INTEGER`, Strapi will now fail fast with an explicit error, instead of returning a silently corrupted number.

## How to test it?

- Start a Strapi instance on any prior version
- Create content types and content
- Stop the Strapi instance
- Checkout to this version (including the bigint migration)
- Start the Strapi instance
- All content should be migrated to the new bigint ids but the API still delivers number IDs

## Related issue(s)/PR(s)

resolves #24491

### DIFF
--- a/examples/getstarted/compose.yml
+++ b/examples/getstarted/compose.yml
@@ -1,0 +1,27 @@
+services:
+  postgres:
+    image: postgres:17-alpine
+    container_name: strapi-empty-postgres
+    restart: unless-stopped
+    environment:
+      POSTGRES_USER: strapi
+      POSTGRES_PASSWORD: strapi
+      POSTGRES_DB: strapi
+    ports:
+      - '5100:5432'
+    volumes:
+      - ./.tmp/postgres:/var/lib/postgresql/data
+  mysql:
+    image: mysql:8.0
+    container_name: strapi-empty-mysql
+    restart: unless-stopped
+    environment:
+      MYSQL_ROOT_PASSWORD: strapi
+      MYSQL_DATABASE: strapi
+      MYSQL_USER: strapi
+      MYSQL_PASSWORD: strapi
+    ports:
+      - '5101:3306'
+    volumes:
+      - ./.tmp/mysql:/var/lib/mysql
+    command: --default-authentication-plugin=mysql_native_password
--- a/examples/getstarted/config/database.js
+++ b/examples/getstarted/config/database.js
@@ -1,51 +1,55 @@
-const sqlite = {
-  client: 'sqlite',
-  connection: {
-    filename: '.tmp/data.db',
-  },
-  useNullAsDefault: true,
-};
+module.exports = ({ env }) => {
+  const client = env('DATABASE_CLIENT', 'sqlite');
 
-const postgres = {
-  client: 'postgres',
-  connection: {
-    database: 'strapi',
-    user: 'strapi',
-    password: 'strapi',
-    port: 5432,
-    host: 'localhost',
-  },
-};
+  const sqlite = {
+    client: 'sqlite',
+    connection: {
+      filename: '.tmp/data.db',
+    },
+    useNullAsDefault: true,
+  };
 
-const mysql = {
-  client: 'mysql',
-  connection: {
-    database: 'strapi',
-    user: 'strapi',
-    password: 'strapi',
-    port: 3306,
-    host: 'localhost',
-  },
-};
+  const postgres = {
+    client: 'postgres',
+    connection: {
+      database: env('DATABASE_NAME', 'strapi'),
+      user: env('DATABASE_USERNAME', 'strapi'),
+      password: env('DATABASE_PASSWORD', 'strapi'),
+      port: env('DATABASE_PORT', 5432),
+      host: env('DATABASE_HOST', 'localhost'),
+    },
+  };
 
-const mariadb = {
-  client: 'mysql',
-  connection: {
-    database: 'strapi',
-    user: 'strapi',
-    password: 'strapi',
-    port: 3307,
-    host: 'localhost',
-  },
-};
+  const mysql = {
+    client: 'mysql',
+    connection: {
+      database: env('DATABASE_NAME', 'strapi'),
+      user: env('DATABASE_USERNAME', 'strapi'),
+      password: env('DATABASE_PASSWORD', 'strapi'),
+      port: env('DATABASE_PORT', 3306),
+      host: env('DATABASE_HOST', 'localhost'),
+    },
+  };
 
-const db = {
-  mysql,
-  sqlite,
-  postgres,
-  mariadb,
-};
+  const mariadb = {
+    client: 'mysql',
+    connection: {
+      database: env('DATABASE_NAME', 'strapi'),
+      user: env('DATABASE_USERNAME', 'strapi'),
+      password: env('DATABASE_PASSWORD', 'strapi'),
+      port: env('DATABASE_PORT', 3307),
+      host: env('DATABASE_HOST', 'localhost'),
+    },
+  };
 
-module.exports = {
-  connection: process.env.DB ? db[process.env.DB] || db.sqlite : db.sqlite,
+  const db = {
+    mysql,
+    sqlite,
+    postgres,
+    mariadb,
+  };
+
+  return {
+    connection: client ? db[client] || db.sqlite : db.sqlite,
+  };
 };
--- a/packages/core/content-manager/server/src/history/models/history-version.ts
+++ b/packages/core/content-manager/server/src/history/models/history-version.ts
@@ -7,7 +7,7 @@ const historyVersion: Model = {
   singularName: 'history-version',
   attributes: {
     id: {
-      type: 'increments',
+      type: 'bigincrements',
     },
     contentType: {
       type: 'string',
--- a/packages/core/content-manager/server/src/services/data-mapper.ts
+++ b/packages/core/content-manager/server/src/services/data-mapper.ts
@@ -24,7 +24,7 @@ export default () => ({
       isDisplayed: isVisible(contentType),
       attributes: {
         id: {
-          type: 'integer',
+          type: 'biginteger',
         },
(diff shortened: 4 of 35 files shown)


burakozdemir32:
Hi @nclsndr 👋 Good to see this one in progress, any plan on when to release it? It will impact some users sooner or later, if they have daily content updates for components.

nclsndr:
Thank you for your feedback @burakozdemir32! 
We are carefully testing this change to account for migration issues and memory footprint against complex datasets.
Would you be interested in testing this migration ahead of time with your client data?

burakozdemir32:
> Thank you for your feedback @burakozdemir32! We are carefully testing this change to account for migration issues and memory footprint against complex datasets. Would you be interested in testing this migration ahead of time with your client data?

Hi @nclsndr 👋 I'd be glad to test this out. Let me know how we can plan this further 🙌

burakozdemir32:
Hi again @nclsndr 👋

Just a gentle reminder on this, happy to help test whenever it’s convenient on your side 😊

nclsndr:
Sorry for the delay @burakozdemir32!
Our support team is monitoring the reports and so far we had none (even with quite large entreprise apps). We don't really expect it to become a high priority in the next weeks / months. 

Yet, if you think you have apps that are currently at risk, we're interested in knowing the expected time frame in which you'll hit the limitation.
As well, we're interested in feedback on how this proposal would perform in real life apps. If you have a local/test environment where you can switch your strapi version, we can build you an experimental to test it out (Note: the experimental does not get updates and must not reach production).

burakozdemir32:
Hi @nclsndr, thanks for following up!

To give you some numbers: our current max component (shared translatable link) ID is ~2,182,880. We run daily automated translation updates across all pages, and since Strapi deletes and recreates component rows on every update, each cycle burns through IDs globally, not per content entry, but per component instance across the entire application.

We haven't calculated the exact daily burn rate yet, but with a growing number of pages and locales, the consumption is accelerating over time. Against the UNSIGNED INT ceiling of ~4.3 billion we likely have years of runway today, but it's not something we want to leave unchecked, especially for long-lived production instances.

Happy to test an experimental build in a local/staging environment. Just let me know how to get it and I'll run it against a copy of our production dataset and report back.

nclsndr:
Thank you for sharing those details @burakozdemir32!

I rebased the branch on latest changes and published an experimental release.
It will be available in a few minutes (process still running) on https://www.npmjs.com/package/@strapi/strapi/v/0.0.0-experimental.2abd3924d019681230b084d27bed7252233c0296.
Feel free to conduct experiments (locally and staging with it). Please consider this experimental as non-stable and thus not meant for production.
Eager to have your feedback on the changes!

PR #24958, opened 2025-12-02
chore: remove in-app marketplace

boazpoolman:
### What does it do?

It removes the in app marketplace. In particular:

1. It removes the page located at `/admin/marketplace`
2. It allows menu links to specify a 'target' prop
3. It updates the main menu link to `https://market.strapi.io` opening a new tab
4. It updates the 'Custom Field' link to go to `https://market.strapi.io?categories=Custom+fields`
5. It leaves the marketplace permissions in tact, allowing RBAC for the marketplace menu item
6. It removes all the associated marketplace translations 

### Why is it needed?

Because the in-app marketplace is not adding enough value. We rather put time in to improving the standalone marketplace hosted at https://market.strapi.io

### How to test it?

Start the Strapi app and witness the changes as described.


jhoward1994:
I think there are marketplace translations to remove in `packages/core/admin/admin/src/translations/vi.json` too ?

packages/core/admin/admin/src/hooks/useMenu.ts
```diff
@@ -16,6 +16,7 @@ export type MenuItem = Omit<StrapiAppContextValue['menu'][number], 'Component'>
 };
 export type MobileMenuItem = {
   to: string;
+  target?: string;
```
jhoward1994, line 19: Do we need this extra property ?

packages/core/admin/admin/src/hooks/useMenu.ts
```diff
@@ -49,7 +50,8 @@ const useMenu = (shouldUpdateStrapi: boolean) => {
           id: 'global.marketplace',
           defaultMessage: 'Marketplace',
         },
-        to: '/marketplace',
+        to: 'https://market.strapi.io',
+        target: '_blank',
```
jhoward1994, line 54: Seeing as the marketplace is an external link now, do we need to maintain any
marketplace admin permissions? e.g. all references to:
`admin::marketplace.read`

packages/core/admin/admin/src/hooks/useMenu.ts
boazpoolman: Well, I figured we'd want the external link to be `target="_blank"` and in order to do that we need this extra property

packages/core/admin/admin/src/hooks/useMenu.ts
boazpoolman: I did consider that, but decided these permissions still make sense for the following reasons:

1. It controls whether you can see the `/admin/settings/list-plugins` page too
2. It can still be used to remove the marketplace link from the navigation for certain roles

boazpoolman:
> I think there are marketplace translations to remove in packages/core/admin/admin/src/translations/vi.json too ?

I don't see any? There is just `global.marketplace` which is still used as the label of the menu item.

jhoward1994:
> > I think there are marketplace translations to remove in packages/core/admin/admin/src/translations/vi.json too ?
> 
> I don't see any? There is just `global.marketplace` which is still used as the label of the menu item.

I think they're here ?
https://github.com/strapi/strapi/blob/6038a425754f74197d2f0c7d303cbf9298b7f67f/packages/core/admin/admin/src/translations/vi.json#L361C1-L362C1

packages/core/admin/admin/src/hooks/useMenu.ts
jhoward1994: Okay makes sense. Don't have to do this in this PR, but with the removal of the in app marketplace do you think the name of this permission should be changed?

boazpoolman: I did consider that, but wouldn't that be a breaking change?

boazpoolman:
@jhoward1994 I did see the `vi.json` translation file now, as well as some old translations in `it.json`. Also merge conflicts have been solved so this should be good to go!

PR #24959, opened 2025-12-02
fix(tests/scripts): disable gpgsign for e2e commits

nclsndr:
## What does it do?

Set `commit.gpgsign=false` flag in commit command in `tests/scripts/run-e2e-tests.js`

## Why is it needed?

This change allows people using GPG commit signature to run e2e tests without caring about the feature being enabled or not.

### With no GPG keys

_— Baseline usage and CI_

Run e2e tests

```bash
yarn test:e2e
```

MUST work as on previous versions.

### With GPGP keys

1. Prerequisites 
 
The current git config must contain `gpgsign: true` config and have in-use GPG signing keys.

```json
{
  "commit.gpgsign": "true"
}
```

2. Run e2e tests

```bash
yarn test:e2e
```

### Related issue(s)/PR(s)

ø

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -20,10 +20,12 @@ api:
   - 'tests/api/**'
 e2e:
   - 'tests/scripts/run-e2e-tests.js'
+  - 'tests/scripts/run-tests.js'
   - 'tests/e2e/**'
   - 'playwright.base.config.js'
 cli:
   - 'tests/scripts/run-cli-tests.js'
+  - 'tests/scripts/run-tests.js'
   - 'tests/cli/**'
 docs:
   - 'docs/**'


nclsndr:
We won't probably merge this PR since it is not directly applicable to #24869.

nclsndr:
@innerdvations since #24869 is merged. Should I bring this PR change to main?

innerdvations:
Yeah, LGTM, we'll just need to test it on the CI and on local systems without gpgsign to make sure it doesn't cause problems that direction

nclsndr:
I'm interested to get feedback on how this runs on non-gpg equipped system.

innerdvations:
We can do it here or another PR, but we need to update the workflow to run e2e + cli tests any time this file is edited so that we can see if it still works; right now it's skipping them because it doesn't think there's a reason to run them

nclsndr:
ok did it @innerdvations 👌

innerdvations (approved):
Looks like CI works fine, and it also ran locally for me!

innerdvations:
@cache-your-dreams Looks like the AI review works, but it still comments a lot of useless things ("excellent addition!") -- do you think there's a way to tone it down so it only comments things that should be changed? And also make it so that it only comments on things that are directly impacted by the current PR and not existing issues (like "run-scripts is not tested")?

PR #24967, opened 2025-12-03
fix: added media to list of non localized fields

Adzouz:
### What does it do?

On issue ticket: https://github.com/strapi/strapi/issues/21850
I fixed all scalar fields to retain information in case of non localizable setting but it was missing the medias.
The original PR: [#24659](https://github.com/strapi/strapi/pull/24659)

https://github.com/user-attachments/assets/332a74e2-e8cb-4bf6-8433-d2edb9aa73be

### Why is it needed?

When you create a new locale version of an entry, you should have all non localizable fields prefilled correctly (except relations, dz and components) but the medias should be part of them.

### How to test it?

- You should have a collection type with Internationalization enabled and have a media field that is not localized
- Go to the Content Manager
- Create a new entry in the collection type
- Put a media (or more if multiple) and Publish
- Select another locale in the locale selector to create a new locale version
-> The media field should already be prefilled with the previous info provided

### Related issue(s)/PR(s)

Original PR: [#24659](https://github.com/strapi/strapi/pull/24659)
Related to https://github.com/strapi/strapi/issues/21850

🚀

### DIFF
--- a/packages/core/content-manager/server/src/services/document-metadata.ts
+++ b/packages/core/content-manager/server/src/services/document-metadata.ts
@@ -6,7 +6,7 @@ import type { Core, UID, Modules } from '@strapi/types';
 import type { DocumentMetadata } from '../../../shared/contracts/collection-types';
 import { getPopulateForValidation } from './utils/populate';
 
-const { getScalarAttributes } = contentTypes;
+const { getScalarAttributes, getMediaAttributes } = contentTypes;
 
 export interface DocumentVersion {
   id: string | number;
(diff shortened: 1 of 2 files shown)


Adzouz:
> Something is weird, when I add a localized media in my model, I can't even see the field in the CM 🤔. I can see the media field only if the localization is disable on that field.
> 
> Screenshot 2025-12-04 at 09 35 44 Screenshot 2025-12-04 at 09 35 53

@HichamELBSI I also have the issue but I don't think it's related. After reloading the page, the field appears correctly, is that the same for you? Might be a caching issue after server reloads on CTB?

HichamELBSI:
> > Something is weird, when I add a localized media in my model, I can't even see the field in the CM 🤔. I can see the media field only if the localization is disable on that field.
> 
> > 
> 
> > Screenshot 2025-12-04 at 09 35 44 Screenshot 2025-12-04 at 09 35 53
> 
> 
> 
> @HichamELBSI I also have the issue but I don't think it's related. After reloading the page, the field appears correctly, is that the same for you? Might be a caching issue after server reloads on CTB?

@Adzouz hmmm, I don't know. The thing is that I didn't have the issue on develop. I might need to try again. Also, I didn't had the issue with a localized input media 🤔

Adzouz:
@HichamELBSI That's weird, on my side I did the same on develop and I have the same issue 😢 Let me know, we can sync to check what's going on

HichamELBSI:
> @HichamELBSI That's weird, on my side I did the same on develop and I have the same issue 😢 Let me know, we can sync to check what's going on

@Adzouz Ok I did reproduce the bug on develop. My bad, I think we can move forward with this PR.

PR #24968, opened 2025-12-03
feat: add shai-hulud 2.0 security scanning

cache-your-dreams:
### What does it do?
Adds the Shai-Hulud 2.0 Detector github workflow

### Why is it needed?
We want to stop Shai-Hulud 2.0 being injected into Strapi

### How to test it?
View scan results in the Actions tab of your repository.

### DIFF
--- a/.github/workflows/shai-hulud-check.yml
+++ b/.github/workflows/shai-hulud-check.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: [develop]
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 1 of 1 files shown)


cache-your-dreams:
@innerdvations do you think this is still relevant?

innerdvations:
Yes and no. Github and npm are looking for known iocs, but it would be good to have something like this for any future evolutions. For now I'll close this one, but put something in the backlog to see what we can do to prevent being hit by any future wave of attacks

PR #24970, opened 2025-12-03
fix(content-type-builder): disable display of invisible content types

nclsndr:
### What does it do?

Adds visibility enforcement for content types in the content-type builder to prevent direct URL access to invisible content types.

**Backend changes:**
- Added visibility check in `getContentType` controller that returns 404 when accessing an invisible content type via the API endpoint (`GET /content-type-builder/content-types/:uid`)

**Frontend changes:**
- Added visibility check in `ListView` component that redirects users away from invisible content types when accessing them directly via URL in the admin panel

### Why is it needed?

Content types marked as invisible in the content-type builder were still accessible by directly navigating to their configuration page URL (e.g., `/admin/plugins/content-type-builder/content-types/admin::transfer-token`). This bypassed the visibility setting, allowing users to view and potentially modify content types that should be hidden from the UI.

### How to test it?

1. Start a Strapi instance with the content-type builder enabled
2. Verify that `admin::transfer-token` (or any other content type marked as invisible) is not visible in the content-type builder navigation menu
3. Attempt to visit the content type directly via URL:
   - Navigate to: `http://localhost:1337/admin/plugins/content-type-builder/content-types/admin::transfer-token`
   - **Expected behavior**: The page should redirect to `/admin/plugins/content-type-builder/content-types/create-content-type` instead of displaying the transfer-token configuration page
4. Verify the API endpoint also enforces visibility:
   - Make a request to: `GET http://localhost:1337/content-type-builder/content-types/admin::transfer-token`
   - **Expected behavior**: Should return 404 status code
5. Verify visible content types still work normally when accessed directly

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/packages/core/content-type-builder/server/src/controllers/content-types.ts
+++ b/packages/core/content-type-builder/server/src/controllers/content-types.ts
@@ -8,6 +8,7 @@ import {
   validateUpdateContentTypeInput,
   validateKind,
 } from './validation/content-type';
+import { isContentTypeVisible } from '../services/content-types';
 
 export default {
   async getContentTypes(ctx: Context) {
@@ -43,6 +44,10 @@ export default {
     const contentType = strapi.contentTypes[uid];
 
     if (!contentType) {
+      return ctx.send({ error: 'contentType.notFound' }, 404);
+    }
+
+    if (!isContentTypeVisible(contentType)) {
       return ctx.send({ error: 'contentType.notFound' }, 404);
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/content-type-builder/server/src/controllers/content-types.ts
```diff
@@ -46,6 +47,10 @@ export default {
       return ctx.send({ error: 'contentType.notFound' }, 404);
     }
 
+    if (!isContentTypeVisible(contentType)) {
+      return ctx.send({ error: 'contentType.notFound' }, 404);
+    }
```
nclsndr, line 52: Do we actually want this enforcement?

Bassel17: does @maccomaccomaccomacco want it ?
if yes yeah I guess, anyways am good with the code

PR #24976, opened 2025-12-04
fix(documentation): align with Data.ID: number | string

nclsndr:
### What does it do?

Update the documentation plugin to reflect the type described for [`Data.ID: number | string`](https://github.com/strapi/strapi/blob/develop/packages/core/types/src/data/constants.ts#L4)

### Why is it needed?

Strapi defines IDs as `number | string` in the rest of its definition.
In the future, we want to move from increments to string-based IDs.

> [!WARNING]
> This change is a breaking change to the openAPI generation. We would push it forward since the feature is still beta and not supposed to be widely adopted.

### How to test it?

- `strapi openapi generate` must generate updated schema definitions including:
```ts
id: {
  oneOf: [{ type: 'string' }, { type: 'number' }],
}
```

### Related issue(s)/PR(s)

Side effect of #24953

### DIFF
--- a/packages/plugins/documentation/server/src/services/__tests__/build-component-schema.test.ts
+++ b/packages/plugins/documentation/server/src/services/__tests__/build-component-schema.test.ts
@@ -73,7 +73,7 @@ describe('Documentation plugin | Build component schema', () => {
         type: 'object',
         properties: {
           id: {
-            type: 'number',
+            oneOf: [{ type: 'string' }, { type: 'number' }],
           },
           documentId: {
             type: 'string',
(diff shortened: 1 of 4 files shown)


Convly:
Numbers would still work in theory so it shouldn't break usage. Still a change in the API but I would say it's ok as it's fixing a bug

nclsndr:
> Numbers would still work in theory so it shouldn't break usage. Still a change in the API but I would say it's ok as it's fixing a bug

Thank you for the feedback @Convly. Do you approve it then?

PR #24983, opened 2025-12-04
fix: normalize empty blocks editor value to undefined

### LINKED ISSUE #20480: Rich text (Blocks) never empty
Hello 👋 

- Environment: development
- OS: darwin-x64
- Strapi Version: 4.24.3
- Node/Yarn Version: yarn/1.22.19 npm/? node/v18.20.2 darwin x64
- Edition: Community
- Database: sqlite
- Typescript

### Bug Description

It's impossible to empty a Rich text (Blocks) field.
There is always something in it.
In the API (cf screenshots), there always is an Array with one element.

1. Go to Admin
2. Create a single or collection type
3. Add a Rich text (Blocks) field
4. Save
5. Create a content from this single or collection type
6. Fill the rich text field
7. Save
8. Try to entirely empty this field

### Expected behavior

The field should be entirely empty.
In the API, the array should exist but empty.

### Screenshots

Capture d’écran 2024-06-10 à 15 33 54

Capture d’écran 2024-06-10 à 15 33 25

Thank you ;)

theaniketgiri:
### What does it do?

This PR adds a helper function `isBlocksValueEmpty()` in `BlocksEditor.tsx` that detects when the Blocks editor content is effectively empty (a single paragraph with only empty text) and normalizes the value to `undefined` instead of keeping the empty paragraph structure.

**Technical changes:**
- Added `isBlocksValueEmpty()` helper function that checks if the blocks value consists of a single paragraph with a single empty text node
- Modified `handleSlateChange` to call `onChange(undefined)` when the content is detected as empty
- Added corresponding test case to verify the empty content normalization behavior

### Why is it needed?

Currently, when users clear all content from the Rich text (Blocks) field, Slate.js always maintains at least one empty paragraph:

```json
[{ "type": "paragraph", "children": [{ "type": "text", "text": "" }] }]
```

This means the Blocks field can **never** be truly empty in the API response, even when the user has deleted all content. This causes issues when:
- Checking if a field has content (`field !== null` always returns true)
- Conditional rendering based on empty state
- Data validation expecting empty/null values

### How to test it?

1. Start the Strapi admin panel
2. Navigate to any content type with a Rich text (Blocks) field
3. Add some content to the Blocks editor
4. Select all content and delete it (Ctrl+A, Backspace)
5. Save the entry
6. **Expected:** The API response should show the blocks field as `null` or `undefined`
7. **Before fix:** The API response would contain `[{ "type": "paragraph", "children": [{ "type": "text", "text": "" }] }]`

**Run tests:**

```bash
yarn test:front --testPathPattern="BlocksInput"
```

### Related issue(s)/PR(s)

Fixes #20480

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
@@ -28,6 +28,43 @@ import { withStrapiSchema } from './plugins/withStrapiSchema';
 
 import type { Schema } from '@strapi/types';
 
+/**
+ * Checks if the blocks content is effectively empty.
+ * An empty blocks value is defined as:
+ * - null or undefined
+ * - an empty array
+ * - an array with a single paragraph containing only empty text
+ */
+const isBlocksValueEmpty = (value: Descendant[]): boolean => {
+  if (!value || value.length === 0) {
+    return true;
+  }
+
+  // Check if content is a single empty paragraph
+  if (value.length === 1) {
+    const firstBlock = value[0];
+    if (
+      'type' in firstBlock &&
+      firstBlock.type === 'paragraph' &&
+      'children' in firstBlock &&
+      Array.isArray(firstBlock.children) &&
+      firstBlock.children.length === 1
+    ) {
+      const firstChild = firstBlock.children[0];
+      if (
+        'type' in firstChild &&
+        firstChild.type === 'text' &&
+        'text' in firstChild &&
+        firstChild.text === ''
+      ) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+};
+
 /* -------------------------------------------------------------------------------------------------
  * BlocksEditorProvider
  * -----------------------------------------------------------------------------------------------*/
@@ -213,7 +250,12 @@ const BlocksEditor = React.forwardRef<{ focus: () => void }, BlocksEditorProps>(
           // Set a new debounce timeout
           debounceTimeout.current = setTimeout(() => {
             incrementSlateUpdatesCount();
-            onChange(name, state as Schema.Attribute.BlocksValue);
+            // Normalize empty blocks content to undefined
+            // This fixes the issue where an empty blocks field always contains a single empty paragraph
+            const normalizedValue = isBlocksValueEmpty(state)
+              ? undefined
+              : (state as Schema.Attribute.BlocksValue);
+            onChange(name, normalizedValue);
             debounceTimeout.current = null;
           }, 300);
         }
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/tests/BlocksInput.test.tsx
```diff
@@ -110,4 +110,30 @@ describe('BlocksInput', () => {
...
+    expect(editable).toBeInTheDocument();
+
+    // Select all text and delete it
+    await user.click(editable);
+    await user.keyboard('{Control>}a{/Control}{Backspace}');
+
+    // The editor should now have an empty paragraph visually,
+    // but the form value should be normalized to null
+    // We can verify the editor still renders without errors
+    expect(screen.getByText('blocks type')).toBeInTheDocument();
```
markkaylor, line 137: I'm not sure this assertion actually tests your fix. I think this test would pass without your fix applied. I think what we want to assert here is if the input value is `null` after deletion instead of a paragraph block. But here you are only asserting the visual representation of emptiness not the empty value.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
```diff
@@ -28,6 +28,43 @@ import { withStrapiSchema } from './plugins/withStrapiSchema';
...
+        firstChild.type === 'text' &&
+        'text' in firstChild &&
+        firstChild.text === ''
+      ) {
+        return true;
+      }
+    }
+  }
+
+  return false;
```
markkaylor, line 65: I think there is a simpler way to do this using Editor.isEmpty

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
```diff
@@ -213,7 +250,12 @@ const BlocksEditor = React.forwardRef<{ focus: () => void }, BlocksEditorProps>(
           // Set a new debounce timeout
           debounceTimeout.current = setTimeout(() => {
             incrementSlateUpdatesCount();
-            onChange(name, state as Schema.Attribute.BlocksValue);
+            // Normalize empty blocks content to undefined
+            // This fixes the issue where an empty blocks field always contains a single empty paragraph
+            const normalizedValue = isBlocksValueEmpty(state)
+              ? undefined
+              : (state as Schema.Attribute.BlocksValue);
+            onChange(name, normalizedValue);
```
markkaylor, line 258: This is good but I think we also need to normalize the comparison in the effect below or we lose focus on the editor when deleting all the content

markkaylor: On my end I actually have this and I can't delete at all

Image

theaniketgiri: Thanks for the detailed review @markkaylor! I appreciate the feedback.

You're right - using `Editor.isEmpty()` is much cleaner than my manual check, and I missed the focus issue when deleting content. Your solution in #24990 is more complete.

Feel free to close this PR in favor of yours. Thanks for taking the time to explain the improvements! 🙌

PR #24990, opened 2025-12-05
fix(content-manager): blocks editor state always contains at least a paragraph block

markkaylor:
### What does it do?

Normalizes the blocks editor state to return `null` when empty instead of storing an empty paragraph block.

I chose null since that is the value saved when the user has not interacted with blocks editor at all

### Why is it needed?

The blocks editor was saving an empty paragraph block when removing content. It save as null.

### How to test it?

1. Open the content manager with a blocks field
2. Create or edit an entry
3. Clear all content from the blocks editor (delete everything)
4. Save/publish the entry
5. Verify the field value is stored as `null` rather than containing an empty paragraph block (check the database and the content api)

When deleting all content the editor should NOT lose focus

https://github.com/user-attachments/assets/ef05baa2-9240-48b6-abf9-ae430f70bb58

### Related issues

resolves https://github.com/strapi/strapi/issues/20480

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksEditor.tsx
@@ -287,7 +287,9 @@ const BlocksEditor = React.forwardRef<{ focus: () => void }, BlocksEditorProps>(
         <VisuallyHidden aria-live="assertive">{liveText}</VisuallyHidden>
         <Slate
           editor={editor}
-          initialValue={value || [{ type: 'paragraph', children: [{ type: 'text', text: '' }] }]}
+          initialValue={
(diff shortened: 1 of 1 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/tests/normalizeBlocksState.test.ts
markkaylor, line 1: I tried to test this as an integration test in BlocksInput but I just couldn't get the keyboard events to work with slate or find a way to read the input value. I think this unit test could be enough but happy to take a look with another set of eyes.

remidej: I remember encountering the same limitation with Slate and Jest. I think this is enough 👍

PR #24999, opened 2025-12-08
[CM] request-scoped caching and parallelisation

jhoward1994:
## What does it do?

Performance optimizations for the content manager, especially for deeply nested content. Changes include:

### Caching:

- `createModelCache` utility — request-scoped cache for `getModel()` calls
- `createPermissionFieldsCache` — caches CASL permission field calculations per action+subject
- Request-scoped cache for documentId → entityId lookups during document transforms
- Batch loading + request-scoped cache for component configurations (single DB query instead of N queries)

### Parallel processing:

- Parallel array traversal in `traverse-entity.ts` for relations, media, components, and dynamic zones
- Ordered error handling helper so parallel validation keeps the same error order as before
- Parallel validation in document service repository (filters, sort, fields, populate run together)

### Other optimizations:

- Map-based indexing for document status lookups
- Permission checker can skip entity fetches when access rules have no conditions (used for `countDraftRelations`)
- Internal migration adds indexes for `document_id` and common composite filters

### For discussion (not to merge as-is):

- `PERFORMANCE-IMPROVEMENTS.md` — documents the problem and approach
- `examples/empty/` — benchmark infrastructure with 9-level nested components

## Why is it needed?

Performance degradation compared to v4 for customers with deeply nested content (9+ levels of components). V5's validation pipeline runs 4 schema traversals per request that V4 didn't have, compounding with the `populateDeep(Infinity)` bottleneck. These changes reduce repeated lookups, avoid extra loads, and make validation work more efficient without changing behavior.

### Unit tests

- `packages/core/utils/src/__tests__/model-cache.test.ts`
- `packages/core/core/src/services/document-service/utils/__tests__/ordered-parallel.test.ts`
- `packages/core/core/src/services/document-service/transform/__tests__/data-request-cache.test.ts`
- `packages/core/content-manager/server/src/services/__tests__/components.test.ts`
- `packages/core/content-manager/server/src/services/__tests__/permission-checker.test.ts`
- `packages/core/database/src/migrations/internal-migrations/__tests__/5.0.0-06-add-document-id-indexes.test.ts`

### Benchmark (in examples/empty using example deeply nested content type)

Also ran against reportedly slow performing apps locally

#### 1. Navigate to the empty example

```bash
cd examples/empty
```

#### 2. Start the server (first run creates admin + test data)

```bash
yarn develop
```

#### 3. Create admin user in browser: http://localhost:1337/admin

- Email: admin@strapi.io
- Password: Admin123!

#### 4. Run benchmark (in another terminal)

```bash
node perf-benchmark.mjs
```

### Manual testing

- Create content type with deeply nested relations/components
- Use Content Manager to list/edit entries with complex populate queries
- Compare response times before/after

## Related issue(s)/PR(s)

DX-2282
CMS-30


innerdvations:
Just testing out the AI Reviewer. Not great at the moment 😆 I'll try to resolve the useless comments and leave the ones that might have something to them

packages/core/utils/benchmark-model-cache.js
```diff
@@ -0,0 +1,285 @@
+/**
+ * Performance benchmark for model cache optimization
+ * This script demonstrates the performance improvement
+ * Run with: node benchmark-model-cache.js
+ */
+
+// TODO - delete this file before merging
```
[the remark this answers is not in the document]
innerdvations: I don't think this should necessarily be deleted; could we keep it but make it more accessible to use in the future to monitor performance?

That is, we shouldn't just keep it as dead code, but if it's documented and/or available from a package.json script it would be great to have.

packages/core/admin/server/src/services/permission/permissions-manager/sanitize.ts
```diff
@@ -51,9 +58,12 @@ export default ({ action, ability, model }: any) => {
 
   const { removeDisallowedFields } = sanitize.visitors;
 
+  // Create request-scoped model cache to avoid redundant getModel() calls
+  const modelCache = createModelCache(strapi.getModel.bind(strapi));
```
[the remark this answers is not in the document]
innerdvations: This should be okay actually, because any model change is going to restart the server and release the memory anyway

packages/core/utils/src/model-cache.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Request-scoped model cache for performance optimization.
+ *
+ * This cache prevents redundant calls to getModel() within a single request,
+ *
+ * The cache is request-scoped and must be cleared after each request to prevent
+ * stale data from being cached across requests.
+ *
```
innerdvations: This is a little misleading, we're using it two ways. In sanitize.ts we don't clear it and that's fine because permissions managers are created per request and it's instance-scoped. In repository.ts it's function-scoped and would get garbage collected but we're clearing it defensively anyway

packages/core/utils/src/model-cache.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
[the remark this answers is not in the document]
innerdvations: I think this comment can be safely ignored based on my other comments, but might be worth a think in case I'm wrong

packages/core/utils/src/model-cache.ts
innerdvations: In fact, unless I'm missing something, the model cache (in theory) could be globally scoped and never cleared, instead of being request-scoped. But I would only do that if it provides a significant performance improvement; if someone updates the models via code outside of bootstrap it wouldn't get updated here

innerdvations (commented):
I haven't tested it yet, but the idea seems good, I don't see any problems with doing this.

Do you see anywhere else in here that could benefit from this type of cache? Maybe the content api sanitizers/validators?

jhoward1994:
> Do you see anywhere else in here that could benefit from this type of cache? Maybe the content api sanitizers/validators?


I think for the performance issues identified content API responses are the most relevant. Maybe these locations could benefit from a similar caching system in future?:
- `content-manager/server/src/services/utils/populate.ts` - `getDeepPopulate()`, `getPopulateForValidation()`
- `content-manager/server/src/services/utils/count.ts` - `getDeepRelationsCount()`
- `content-manager/server/src/controllers/relations.ts`

derrickmehaffy:
> Just testing out the AI Reviewer. Not great at the moment 😆 I'll try to resolve the useless comments and leave the ones that might have something to them

I actually think this is pretty cool, some of the edge cases it picked up on could be legit concerns.

derrickmehaffy:
@jhoward1994 is this in a state where you would be good with releasing an experimental so I can give it to a few customers and Virtuslabs for real world testing?

derrickmehaffy:
> > Do you see anywhere else in here that could benefit from this type of cache? Maybe the content api sanitizers/validators?
> 
> I think for the performance issues identified content API responses are the most relevant. Maybe these locations could benefit from a similar caching system in future?:
> 
> * `content-manager/server/src/services/utils/populate.ts` - `getDeepPopulate()`, `getPopulateForValidation()`
> * `content-manager/server/src/services/utils/count.ts` - `getDeepRelationsCount()`
> * `content-manager/server/src/controllers/relations.ts`

Yeah I think so too, the way the admin populates data really is not ideal, it starts hitting an N+1 problem pretty early on because every single relation/component/dynamic zone population is an additional request rather than just using the native population like the content-api does.

IE if you have 15 relations in a content-type the admin makes 16 requests (at minimum) just for the data, not including any other checks.

derrickmehaffy:
Experimental Version: `0.0.0-experimental.f86f67937111a56cae88a8a8c0d84fb2d9c39dd3`

https://github.com/strapi/strapi/actions/runs/20283437759/job/58251207263

derrickmehaffy:
New Exp: `0.0.0-experimental.cf9e379c46cf263536f65705ad6247ed9a4a0fd1`

jhoward1994:
Changes in `examples/empty` and `PERFORMANCE-IMPROVEMENTS.md` are to help with testing and will be removed before merging

jhoward1994:
`0.0.0-experimental.12ff9c2481ab94e3db507526925787ea5ec432e8`

jhoward1994:
Performance benchmarking of slow performing apps - using `examples/empty/perf-benchmark.mjs`
(Notion search "Slow performing app..." for access to examples)

**Example 1**
```
## Before (`origin/develop`)

| Test                                | Avg   | Min   | Max   | P95   |
| ----------------------------------- | ----- | ----- | ----- | ----- |
| **List View** (populateDeep=1)      | 273ms | 234ms | 414ms | 414ms |
| **Edit View** (populateDeep=config) | 469ms | 455ms | 513ms | 513ms |
| **Count Draft Relations**           | 581ms | 529ms | 706ms | 706ms |

---

## With basic changes (`fix/request-scoped-model-cache`) 

| Test                                | Avg   | Min   | Max   | P95   |
| ----------------------------------- | ----- | ----- | ----- | ----- |
| **List View** (populateDeep=1)      | 216ms | 194ms | 269ms | 269ms |
| **Edit View** (populateDeep=config) | 344ms | 332ms | 358ms | 358ms |
| **Count Draft Relations**           | 455ms | 439ms | 540ms | 540ms |

---

## With POC changes (`fix/request-scoped-model-cache` with POC changes via `STRAPI_CONTENT_MANAGER_MAX_POPULATE_DEPTH=4`)

| Test                                | Avg   | Min   | Max   | P95   |
| ----------------------------------- | ----- | ----- | ----- | ----- |
| **List View** (populateDeep=1)      | 202ms | 186ms | 260ms | 260ms |
| **Edit View** (populateDeep=config) | 145ms | 141ms | 152ms | 152ms |
| **Count Draft Relations**           | 419ms | 412ms | 429ms | 429ms |

```

**Example 2** 
Have access but currently unable to run the app against these changes...

@derrickmehaffy @alex-strapi 
would you be able to run the benchmarks against that app?

derrickmehaffy:
Yeah I can run some benchmarking on Friday. (Sorry for the delay, been a crazy week)

innerdvations:
It seems like there are some tests failing

PERFORMANCE-IMPROVEMENTS.md
innerdvations, line 1: do we need this file?

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -705,19 +706,24 @@ export default {
...
+        return ctx.notFound();
+      }
 
-    if (permissionChecker.cannot.read(entity)) {
+      if (permissionChecker.cannot.read(entity)) {
+        return ctx.forbidden();
+      }
+    } else if (permissionChecker.cannot.read()) {
       return ctx.forbidden();
     }
```
innerdvations: we already check this on line 705, maybe this line should just be an `else` without a condition? (or rather, add another short circuit like `if (!permissionChecker.requiresEntity.read()) return ctx.forbidden()` and then remove the whole if/else

PERFORMANCE-IMPROVEMENTS.md
jhoward1994: We don't and we also don't need the changes in `examples/empty` the changes are just there to help with testing but I'll remove it all before we merge

Issue #25001, opened 2025-12-08
New installation using MySQL regressed


nclsndr:
### Node Version

24.11.1

### NPM/Yarn/PNPM Version

yarn@4.5

### Strapi Version

5.31.4

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

Since the changes brought in #24931.
New installations using MySQL are failing locally.

```bash
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                                                   │
│   Error: alter table `files` add index `upload_files_name_index`(`name`) - BLOB/TEXT column 'name' used in key    │
│   specification without a key length                                                                              │
│   at Packet.asError                                                                                               │
│   (/Code/strapi/node_modules/mysql2/lib/packets/packet.js:728:17)                                                 │
│   at Query.execute                                                                                                │
│   (/Code/strapi/node_modules/mysql2/lib/commands/command.js:29:26)                                                │
│   at Connection.handlePacket                                                                                      │
│   (/Code/strapi/node_modules/mysql2/lib/connection.js:481:34)                                                     │
│   at PacketParser.onPacket                                                                                        │
│   (/Code/strapi/node_modules/mysql2/lib/connection.js:97:12)                                                      │
│   at PacketParser.executeStart                                                                                    │
│   (/Code/strapi/node_modules/mysql2/lib/packet_parser.js:75:16)                                                   │
│   at Socket.<anonymous>                                                                                           │
│   (/Code/strapi/node_modules/mysql2/lib/connection.js:104:25)                                                     │
│   at Socket.emit (node:events:519:28)                                                                             │
│   at Socket.emit (node:domain:489:12)                                                                             │
│   at addChunk (node:internal/streams/readable:561:12)                                                             │
│   at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)                                          │
│                                                                                                                   │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

### Steps to Reproduce

1. In monorepo, checkout `develop`, `yarn build`
2. Navigate to `examples/getstarted`, configure an new/empty MySQL database
3. Launch dev server `yarn develop`

### Expected Behavior

The MySQL database should set up

### Logs

```shell

```

### Additional information

Need reproduction to confirm

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
[...]

PR #25003, opened 2025-12-09
fix(database): keyLength index attribute WIP

nclsndr:
### What does it do?

Detect and handle text based indexes. Apply an arbitrary `keyLength` to them.

### Why is it needed?

Preventing issue described in #25001 

### How to test it?

1. In monorepo, checkout `develop`, `yarn build`
2. Navigate to `examples/getstarted`, configure an new/empty MySQL database
3. Launch dev server `yarn develop` — server must start without error

### Related issue(s)/PR(s)


innerdvations:
At a glance it looks ok; we might need to use the transaction instead of db.raw directly?

PR #25006, opened 2025-12-09
fix(admin): remove @ts-expect-error in useQueryParams hook

ajayi-joseph:
### What does it do?

Removes the `@ts-expect-error` suppression in the `useQueryParams` hook by properly typing the query object deletion logic.

### Why was it needed?

The hook had a TypeScript error suppression with a comment "this is fine, if you want to fix it, please do." The issue was that TypeScript doesn't allow deleting properties from a fully-typed object, even though the runtime behavior was correct.

### How?

- Changed `nextQuery` type from `TQuery` to `Partial<TQuery>` to accurately reflect that properties can be removed
- Added proper key typing with `key as keyof TQuery` for the delete operation
- This makes the code type-safe without changing any runtime behavior

### Testing

- Existing tests pass (useQueryParams.test.tsx)
- No functional changes, only type improvements

### Related

Addresses the TODO comment at line 33 of the original file.


ajayi-joseph:
cc @mathildeleg @Bassel17

innerdvations (approved):
Thanks!

PR #25011, opened 2025-12-09
fix(core): maxFileSize error not detected in body middleware

mhsfh:
### What does it do?

Fixes the error detection logic in the body middleware to properly catch `maxFileSize` errors from `koa-body`. The middleware now checks for both `'maxFileSize'` and `'exceeded'` separately in the error message, instead of looking for the exact string `'maxFileSize exceeded'`.

### Why is it needed?

When uploading a file that exceeds the maximum file size limit, users were seeing a generic "Internal Server Error" message in the Strapi admin panel instead of the proper error message "The uploaded file exceeds the maximum allowed asset size."

**Root cause:**
The error detection was checking for the exact string `'maxFileSize exceeded'`, but the actual error message from `koa-body` has a different format:

```
options.maxFileSize (2147483648 bytes) exceeded, received 2147486330 bytes of file data
```

This format mismatch caused the error to not be caught by the middleware. When the error is not caught, it gets thrown as a generic 500 Internal Server Error. When it is properly caught, the middleware calls `ctx.payloadTooLarge('FileTooBig')`, which:

- Returns HTTP status code 413 (Payload Too Large)
- Sends the error code `'FileTooBig'` to the frontend
- The frontend then displays the user-friendly message: "The uploaded file exceeds the maximum allowed asset size."

**Error in terminal:**

```
[2025-12-09 17:21:58.002] error: options.maxFileSize (2147483648 bytes) exceeded, received 2147486330 bytes of file data

Error: options.maxFileSize (2147483648 bytes) exceeded, received 2147486330 bytes of file data
```

**Error shown in UI:**

```
Internal Server Error
```

### How to test it?

1. Configure a small `maxFileSize` limit in your Strapi middleware configuration (e.g., `maxFileSize: 1000000` for 1MB)
2. Attempt to upload a file larger than the configured limit through the Strapi admin panel
3. **Before the fix:** You would see "Internal Server Error"
4. **After the fix:** You should see the proper error message: "The uploaded file exceeds the maximum allowed asset size." (or the translated equivalent)


innerdvations:
Thanks for fixing this — the root cause analysis is spot on. The old contiguous `'maxFileSize exceeded'` check fails because formidable 2.1.5 (bundled under `koa-body@6.0.1`) puts the byte count between those words:

```
options.maxFileSize (N bytes) exceeded, received M bytes of file data
```

Your split `includes()` check works for the current stack.

One suggestion for a more robust detection: formidable throws a `FormidableError` with stable metadata that `koa-body` preserves (`throwableToError` returns `Error` instances as-is), so we can match on that instead of parsing the message:

```ts
type FormidableLikeError = Error & { code?: number; httpCode?: number };

function isMaxFileSizeError(error: unknown): boolean {
  if (!(error instanceof Error)) return false;

  const err = error as FormidableLikeError;

  // Formidable v2 (current koa-body dep): 1009 = biggerThanMaxFileSize
  // Formidable v3 (future): 1016 = biggerThanMaxFileSize, 1009 = biggerThanTotalMaxFileSize
  if (err.httpCode === 413 && (err.code === 1009 || err.code === 1016)) {
    return true;
  }

  // Fallback for message-only errors
  const message = err.message ?? '';
  return message.includes('maxFileSize') && message.includes('exceeded');
}
```

Benefits:
- Survives message wording changes in minor formidable updates
- More precise than `httpCode === 413` alone (413 also covers `maxFieldsSize`, `maxFields`, etc.)
- Forward-compatible if formidable ever upgrades to v3 (different code numbers)

Either way, a small unit test in `packages/core/core/src/middlewares/__tests__/body.test.ts` mocking a `{ code: 1009, httpCode: 413, message: '...' }` error would lock this in.

Happy to approve once CI is green either way — this is a nice-to-have refinement, not a blocker on the current fix.

_AI-assisted comment._

innerdvations (approved):

Issue #19238, opened 2024-01-17
Chaging the favicon does not work


DenuxPlays:
### Required System information

- Node.js version: 18.18.2
- NPM version: 9.8.1
- Strapi version: 4.15.5
- Database: MariaDB 10.6
- Operating system: Debian 12.4
- Is your project Javascript or Typescript: Typescript

### Describe the bug

This is the config we export via the admin.tsx file:
```js
import favicon from "./extensions/favicon.png";
import logo from "./extensions/logo.svg";

export default {
  config: {
    head: {
      favicon: favicon,
    },
    menu: {
      logo: logo,
    },
    auth: {
      logo: logo,
    },
}
```

Both the menu and auth logo are getting replaced but the favicon stays the same.
Checked the browser cache etc. the favicon isn't cached.

### Steps to reproduce the behavior

1. Go to your admin.js|.ts file
2. update the file with the same content
4. Open strapi, see the favicon not changing

### Expected behavior

The favicon should be the one I provide.

### Screenshots

/

### Code snippets

```js
import favicon from "./extensions/favicon.png";
import logo from "./extensions/logo.svg";

export default {
  config: {
    head: {
      favicon: favicon,
    },
    menu: {
      logo: logo,
    },
    auth: {
      logo: logo,
    },
}
```

### Additional context

/

zlaazlaa:
You can try to change the origin `favicon.png` file in the root of strapi project. @DenuxPlays

DenuxPlays:
> You can try to change the origin `favicon.png` file in the root of strapi project. @DenuxPlays 

Yeah that worked.
Please mention something like this in the wiki.

zlaazlaa:
> > You can try to change the origin `favicon.png` file in the root of strapi project. @DenuxPlays
> 
> Yeah that worked. Please mention something like this in the wiki.

@DenuxPlays Great to hear that it worked for you! I've taken note of your suggestion and updated the wiki to include information about changing the origin favicon.png file in the root of the Strapi project. You can check out the changes in my pull request here: https://github.com/strapi/documentation/pull/1983. Thanks for bringing this to my attention!

DenuxPlays:
> > > You can try to change the origin `favicon.png` file in the root of strapi project. @DenuxPlays
> > 
> > 
> > Yeah that worked. Please mention something like this in the wiki.
> 
[...]

PR #25012, opened 2025-12-10
fix: replace favicon if provided in config

Adzouz:
### What does it do?

The current documentation to replace the favicon is wrong. If we put `head: { favicon }` in the admin config, it's just doing nothing.
Removing useless part in the code and updating the documentation (https://github.com/strapi/documentation/pull/2899) to explain how correctly replace the favicon (either replace the file at the root of a Strapi project or use the strapi::favicon middleware configuration instead to provide a custom path to the icon).

_Retroactively, it should still work on Strapi v4 (support old .ico file) but Strapi v5 uses a PNG._

### Why is it needed?

Currently, even if it's documented officially ([here](https://docs.strapi.io/cms/admin-panel-customization/favicon)) overriding the favicon directly in the app.[tsx|js] file of a Strapi project isn't working.
We need to clarify how to update the favicon.

### How to test it?

Replacing the file:
- Go to the root directory of your Strapi project
- Replace the favicon.png by the file of your choice (but keep the same name)
-> After rebuilding and relaunching your project, you should be able to see the new favicon in the browser tab (if not working, try in incognito mode)

To use a custom icon:
- Create or go to a Strapi project folder
- Put your favicon in your project and give it a custom name like 'custom-favicon.png'
- In `config/middleware.[js|ts]`, add or replace the strapi::favicon config with the following:
```
{
  name: 'strapi::favicon',
  config: {
    path: 'custom-favicon.png',
  },
}
```
-> After rebuilding and relaunching your project, you should be able to see the new favicon in the browser tab (if not working, try in incognito mode)

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/19238

🚀

### DIFF
--- a/examples/getstarted/config/middlewares.js
+++ b/examples/getstarted/config/middlewares.js
@@ -31,12 +31,7 @@ module.exports = [
       handlers: responseHandlers,
     },
   },
-  {
-    name: 'strapi::favicon',
-    config: {
-      path: 'faviconi.png',
-    },
-  },
+  'strapi::favicon',
   'strapi::public',
   {
     name: 'global::test-middleware',
--- a/packages/core/admin/admin/src/tests/StrapiApp.test.tsx
+++ b/packages/core/admin/admin/src/tests/StrapiApp.test.tsx
@@ -539,12 +539,6 @@ describe('ADMIN | new StrapiApp', () => {
       expect(app.configurations.menuLogo).toBe('fr');
     });
 
-    it('should override the favicon', () => {
-      const app = new StrapiApp({ config: { head: { favicon: 'fr' } } });
-
-      expect(app.configurations.head.favicon).toBe('fr');
-    });
-
     it('should override the light theme', () => {
       // @ts-expect-error - test mocks
       const app = new StrapiApp({ config: { theme: { light: { colors: { red: 'black' } } } } });


mathildeleg (commented):
It works well, I'm not a 100% sure I like the solution but couldn't find another way to fix it, I'd like another pair of eyes on this though before I approve 👀

markkaylor:
@Adzouz are we sure this isn't a cache issue as mentioned at the bottom of the [docs](https://docs.strapi.io/cms/admin-panel-customization/favicon)?

Adzouz:
> @Adzouz are we sure this isn't a cache issue as mentioned at the bottom of the [docs](https://docs.strapi.io/cms/admin-panel-customization/favicon)?

Yes, this configuration isn't used anywhere in the codebase unfortunately... I tried to rebuild, open in an incognito tab, nothing is working. The only workaround that we provided (also documented) is to replace the favicon.png file at the project's root, but the configuration part is completely useless and my fix is to make it work like in the documentation.

packages/core/admin/admin/src/utils/favicon.ts
```diff
@@ -0,0 +1,29 @@
...
+    return;
+  }
+
+  const updateFavicon = () => {
+    // Remove existing favicon links
+    document.querySelectorAll("link[rel*='icon']").forEach((link) => link.remove());
+
+    // Create and append new favicon link
+    const link = document.createElement('link');
+    link.type = 'image/png';
```
markkaylor: Couldn't the favicon be something other than `png`?

packages/core/admin/admin/src/StrapiApp.tsx
```diff
@@ -269,6 +270,7 @@ class StrapiApp {
 
     if (customConfig.head?.favicon) {
       this.configurations.head.favicon = customConfig.head.favicon;
+      setFavicon(this.configurations.head.favicon);
```
markkaylor: Even if this is a class, I would deliberately avoid the `set` prefix since it is a standard convention for managing react state and that's not what we are dealing with here. Since there is always a default favicon maybe `replaceFavicon`? 

That makes me think, what happens if they remove the custom favicon, does it fall back to the default?

packages/core/admin/admin/src/utils/favicon.ts
Adzouz: Ah true, don't know why I didn't think about that 🤦‍♂️

packages/core/admin/admin/src/StrapiApp.tsx
Adzouz: If they remove it, the configuration shouldn't contain the custom one, so yes, it should fallback to the default one that is loaded directly in the template

markkaylor:
p.s.

> the only workaround that we provided (also documented) is to replace the favicon.png file 

what's wrong with this being the way to replace the favicon? Are we just trying to support the documentation so this isn't considered a breaking change? I wonder if we could just remove the config entirely

I'm kinda wondering if the documentation just needs to be updated because even with your fix here applied the docs are confusing. It tells you to add an extensions folder (here and on develop I think that does nothing right?), update the config (does nothing on develop), _and_ update the favicon at the root (does the trick). Couldn't the docs just say "Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.". Requires no new code for us and we could even potentially delete some code...

```
Create a /src/admin/extensions/ folder if the folder does not already exist.

Upload your favicon into /src/admin/extensions/.

Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.

Update /src/admin/app.[tsx|js] with the following:
```

Adzouz:
> p.s.
> 
> > the only workaround that we provided (also documented) is to replace the favicon.png file
> 
> what's wrong with this being the way to replace the favicon? Are we just trying to support the documentation so this isn't considered a breaking change? I wonder if we could just remove the config entirely
> 
> I'm kinda wondering if the documentation just needs to be updated because even with your fix here applied the docs are confusing. It tells you to add an extensions folder (here and on develop I think that does nothing right?), update the config (does nothing on develop), _and_ update the favicon at the root (does the trick). Couldn't the docs just say "Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.". Requires no new code for us and we could even potentially delete some code...
> 
> ```
> Create a /src/admin/extensions/ folder if the folder does not already exist.
> 
> Upload your favicon into /src/admin/extensions/.
> 
> Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.
> 
> Update /src/admin/app.[tsx|js] with the following:
> ```

I kind of agree with that too, it shouldn't be a blocker. It's open for debate!

markkaylor:
> > p.s.
> > > the only workaround that we provided (also documented) is to replace the favicon.png file
> > 
> > 
> > what's wrong with this being the way to replace the favicon? Are we just trying to support the documentation so this isn't considered a breaking change? I wonder if we could just remove the config entirely
> > I'm kinda wondering if the documentation just needs to be updated because even with your fix here applied the docs are confusing. It tells you to add an extensions folder (here and on develop I think that does nothing right?), update the config (does nothing on develop), _and_ update the favicon at the root (does the trick). Couldn't the docs just say "Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.". Requires no new code for us and we could even potentially delete some code...
> > ```
> > Create a /src/admin/extensions/ folder if the folder does not already exist.
> > 
> > Upload your favicon into /src/admin/extensions/.
> > 
> > Replace the existing favicon.png|ico file at the Strapi application root with a custom favicon.png|ico file.
> > 
> > Update /src/admin/app.[tsx|js] with the following:
> > ```
> 
> I kind of agree with that too, it shouldn't be a blocker. It's open for debate!

The only issue I see is I think the favicon filename has to be `favicon.png` otherwise if a user want to name the file something else they need to change the config of the favicon middleware to specify the path:

For example:

```
{
    name: 'strapi::favicon',
    config: {
      path: 'my-custom-favicon.svg',
    },
  }
```


Even if I replace the favicon.png with my own favicon.png (not the strapi logo) it doesn't seem to work.

Adzouz:
Ok @markkaylor I updated the logic to handle the default png (and retro compatibility with v4 for .ico file) and removed everything related to the admin config. Once it will be merged, we need to update the documentation to correctly explain how to set the path to a custom one through the middleware configuration. Let me know if anything!

markkaylor:
> Ok @markkaylor I updated the logic to handle the default png (and retro compatibility with v4 for .ico file) and removed everything related to the admin config. Once it will be merged, we need to update the documentation to correctly explain how to set the path to a custom one through the middleware configuration. Let me know if anything!

Sounds good, could we sync with @pwizla to make sure he is on board with the change and if he would like us to handle the documentation PR. I think we should have the doc PR ready before merging so we ship them at the same time.

examples/getstarted/config/middlewares.js
```diff
@@ -31,7 +31,12 @@ module.exports = [
       handlers: responseHandlers,
     },
   },
-  'strapi::favicon',
+  {
+    name: 'strapi::favicon',
+    config: {
+      path: 'faviconi.png',
+    },
+  },
```
markkaylor: I think we should leave it as the default in the examples/getstarted so no custom config here

packages/core/core/src/middlewares/favicon.ts
```diff
@@ -11,18 +11,16 @@ const defaults = {
...
-   *  the next major version
-   */
-  if (!existsSync(resolve(appRoot, faviconPath))) {
-    faviconPath = 'favicon.ico';
+  if (!existsSync(resolve(appRoot, faviconPathConfig))) {
+    if (existsSync(resolve(appRoot, defaults.path))) {
+      faviconPath = defaults.path;
+    } else if (existsSync(resolve(appRoot, 'favicon.ico'))) {
+      faviconPath = 'favicon.ico';
+    }
```
markkaylor, line 23: So if the user provides a config.path use it, if not try to use favicon.png, if it doesn't exist try to use favicon.ico, otherwise error?

examples/getstarted/config/middlewares.js
Adzouz: Omg I shouldn't have push that, it's a mistake

packages/core/core/src/middlewares/favicon.ts
Adzouz: Exactly!

markkaylor (approved):
Awesome, thanks

PR #25015, opened 2025-12-10
fix(admin): clean up lazy component registration warnings

unrevised6419:
### What does it do?

Adjust `if` condition inside addMenuLink/addSettingsLink to not warn when `link.Component` is missing.

### Why is it needed?

Because those `if` conditions have above them invariants, which works the other way around, probably they were confused as invariants, and the logic was reverse.

And it really annoys that each time I start a new Strapi App, I see those warnings in the console.

The changes are needed to fix that, and don't show these warnings

image

### How to test it?

Start a new Strapi application, in the console the user should not see the above warnings.


packages/core/admin/admin/src/core/apis/router.tsx
```diff
@@ -155,11 +155,10 @@ class Router {
...
     if (
-      !link.Component ||
-      (link.Component &&
-        typeof link.Component === 'function' &&
-        // @ts-expect-error – shh
-        link.Component[Symbol.toStringTag] === 'AsyncFunction')
+      link.Component &&
+      typeof link.Component === 'function' &&
+      // @ts-expect-error – shh
+      link.Component[Symbol.toStringTag] === 'AsyncFunction'
```
unrevised6419, line 163: Note: If this was an invariant, then `!link.Component ||` would make sense

packages/core/admin/admin/src/core/apis/router.tsx
```diff
@@ -278,11 +277,10 @@ class Router {
...
```
unrevised6419, line 285: Note: If this was an invariant, then `!link.Component ||` would make sense

packages/core/content-releases/admin/src/index.ts
```diff
@@ -119,15 +119,16 @@ const admin: Plugin.Config.AdminInput = {
     ) {
       app.addSettingsLink('global', {
         id: pluginId,
-        to: '/plugins/purchase-content-releases',
+        to: 'purchase-content-releases',
```
unrevised6419, line 123: > property of your settings link is an absolute path. It should be relative to \`/settings\`

packages/plugins/cloud/admin/src/index.ts
```diff
@@ -22,11 +22,7 @@ export default {
           id: `${pluginId}.Plugin.name`,
           defaultMessage: pluginName,
         },
-        Component: async () => {
-          const { App } = await import('./pages/App');
-
-          return App;
-        },
+        Component: () => import('./pages/App').then((mod) => ({ default: mod.App })),
```
unrevised6419: Note: `async` functions are not allowed

packages/core/content-releases/admin/src/index.ts
```diff
@@ -119,15 +119,16 @@ const admin: Plugin.Config.AdminInput = {
...
           defaultMessage: 'Releases',
         },
         permissions: [],
-        async Component() {
-          const { PurchaseContentReleases } = await import('./pages/PurchaseContentReleases');
-          return { default: PurchaseContentReleases };
+        Component() {
+          return import('./pages/PurchaseContentReleases').then((mod) => ({
+            default: mod.PurchaseContentReleases,
+          }));
```
unrevised6419, line 132: Note: `async` functions are logging warnings, and I'm wondering what was the reasoning behind this logic 🤔 

An async function, and an usual function that returns a promise do the asme.

packages/core/admin/admin/src/core/apis/router.tsx
```diff
@@ -135,7 +135,7 @@ class Router {
 
   public addMenuLink = (
     link: Omit<MenuItem, 'Component'> & {
-      Component: () => Promise<{ default: React.ComponentType }>;
+      Component?: () => Promise<{ default: React.ComponentType }>;
```
unrevised6419: Note: Some plugins call this without a component only to add the menu item.

Bassel17: I would avoid changing this, as this is documented to be required

packages/core/content-releases/admin/src/index.ts
Bassel17: The intent seems to be preparing the API to require a React Router-style lazy module loader in the future

Bassel17 (approved):
Looks good now, thanks for paying attention to this

unrevised6419:
@Bassel17 thanks. Btw, a similar one with the warnings can be seen here https://github.com/strapi/strapi/pull/26150

PR #24826, opened 2025-11-12
fix(graphql): scope status filter inheritance to query tree

Bassel17:
### What does it do?

This PR fixes an issue where custom GraphQL resolvers were incorrectly inheriting status filters from built-in queries, causing relation fields to return null when draft/publish filtering was applied.

### Why is it needed?

The rootQueryArgs introduced in PR #23737 to fix duplicate relations in draft/publish queries was request-scoped instead of query-scoped. This caused unintended cross-contamination between different root queries in the same GraphQL request.

### How to test it?

follow the example in this issue #23936
relations should not return null anymore

### Related issue(s)/PR(s)

fixes #23936

PR #25017, opened 2025-12-11
fix: graphql performance

### LINKED ISSUE #24986: Performance problem when making a request with relationships in userpermissions in GraphQL
### Node Version

22.20

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.31.3

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

Hi, in this new version, GraphQL queries are terrible; they take a long time to query, even for small queries.

### Steps to Reproduce

1: Performing a graphql query with 8 relationships can take up to 5 minutes to complete the request.

### Expected Behavior

1: Performing a graphql query with 8 relationships can take up to 5 minutes to complete the request.

### Logs

```shell
[2025-12-04 18:30:32.728] http: POST /graphql (352271 ms) 200
```

### Code Snippets

_No response_

### Media

Image

### Additional information

_No response_

### Confirmation Checklist

Bassel17:
**What Does It Do?**

- **Fix:** Prevents repeated GraphQL schema rebuilds inside the association resolver by memoizing the built-in query field lookup.

**Why Is It Needed?**

- **Problem:** The association resolver was calling `buildSchema()` during relation resolution, which caused the GraphQL schema to be rebuilt/re-processed repeatedly within a single request. For queries with many relations this resulted in severe performance regressions (orders of magnitude slower).
- **Impact:** Users reported slow GraphQL responses (see issue below). This fix prevents the expensive repeated work by caching the field lookup result.
- **Resolution:** Memoizing the built-in field lookup preserves the original behavior (distinguishing built-in vs custom queries) while eliminating repeated schema rebuilds.
- **Related issue:** Fixes #24986
- **Related PR:** Addresses regression introduced around PR #24826

**How To Test It?**

- Start  the getstarted project
- Open the GraphQL playground at http://localhost:1337/graphql.

Run the following query in the playground and note the time it takes 
```
query Query{
  temps {
    categories {
      name
    }
    category {
      name
    }
  }
  articles {
    authorName
  }
}
```
- check this before applying this fix it will take a few seconds, after applying the fix in this PR it should be back to ms

**Related issue(s)/PR(s)**

- Fixes: #24986
- Related: #24826 (original PR that introduced the rootQueryArgs change)

### DIFF
--- a/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
+++ b/packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
@@ -65,16 +65,8 @@ export default ({ strapi }: Context) => {
         // Helper to check if a field is from built-in queries (not custom resolvers)
         // Use the precomputed lookup populated by the content-api service at schema build time.
         const isBuiltInQueryField = (fieldName: string) => {
-          try {
-            const graphqlService = strapi.plugin('graphql').service('content-api');
-            if (typeof graphqlService.isBuiltInQueryField === 'function') {
-              return graphqlService.isBuiltInQueryField(fieldName);
-            }
-          } catch {
-            // If anything goes wrong, conservatively return false
-          }
-
-          return false;
+          const graphqlService = strapi.plugin('graphql').service('content-api');
+          return graphqlService.isBuiltInQueryField(fieldName);
         };
 
         // Only inherit status from built-in queries to avoid conflicts with custom resolvers


packages/plugins/graphql/server/src/services/builders/resolvers/association.ts
```diff
@@ -63,15 +66,23 @@ export default ({ strapi }: Context) => {
           contentTypes.hasDraftAndPublish(targetContentType);
 
         // Helper to check if a field is from built-in queries (not custom resolvers)
+        // Use a memoization cache to avoid rebuilding the schema on every call.
         const isBuiltInQueryField = (fieldName: string) => {
           if (fieldName.endsWith('_connection')) return true;
 
+          if (builtInFieldCache.has(fieldName)) {
+            return builtInFieldCache.get(fieldName) as boolean;
+          }
```
nclsndr: This suggests that we are still doing the provisioning operation during the first request right?
If true, are we ok with the first request to be slower? Would it benefit from being initialized outside of the actual resolver?

Bassel17: indeed that would be better, I'll move it outside so we have the cache built once on bootstrap and whenever buildSchemas is run

butcherZ (approved):
QAed locally, the fix works well for me!

Issue #24567, opened 2025-10-13
Cannot read properties of undefined (reading 'isFirstSuperAdminUser')


Tobvl:
Hi everyone I just upgraded from strapi 5.13.0 to 5.27.0 through

npx @strapi/upgrade minor

But when I deployed the app to my development environment (running on a 12core 48gb ram instance), I got this error:

```
TypeError: Cannot read properties of undefined (reading 'isFirstSuperAdminUser')
    at Pe (https://xx.com/admin/HomePage-Bao0-P5x.js:30:5872)
    at Dx (https://xx.com/admin/strapi-aeyY7-wE.js:201:386)
    at qq (hthttps://xx.com/admin/strapi-aeyY7-wE.js:217:5947)
    at P0e (https://xx.com/admin/strapi-aeyY7-wE.js:219:12276)
    at EY (https://xx.com/admin/strapi-aeyY7-wE.js:236:2167)
    at vve (https://xx.com/admin/strapi-aeyY7-wE.js:232:22546)
    at Dnt (https://xx.com/admin/strapi-aeyY7-wE.js:232:22119)
    at i6 (https://xx.com/admin/strapi-aeyY7-wE.js:232:21867)
    at vY (https://xx.com/admin/strapi-aeyY7-wE.js:232:18336)
    at fve (https://xx.com/admin/strapi-aeyY7-wE.js:232:19745)
```
      
Is there any advice for this? Thanks in advance!

Tobvl:
Sorry if the Issue format isn't the required, I actually managed to solve this by just stopping and starting again the strapi instance.

Thanks anyways and if there's a direct fix for this, I'd love to know that!

Tofandel:
I hit this as well on 5.31.3

Tofandel:
<img width="1588" height="330" alt="Image" src="https://github.com/user-attachments/assets/1608cb90-3506-4213-8e12-554eee7e818b" />

Looks like it's checking against null, but it's undefined...

https://github.com/strapi/strapi/blob/ff88838bb3a9d86a84e50d750f0142d0f0759444/packages/core/admin/admin/src/components/GuidedTour/Overview.tsx#L175

I think typescript is not building this correctly because that's not the correct transpilation for the '?' operator

Tofandel:
Ok nevermind, that's not when it's erroring

It's erroring on 

guidedTourMeta = []

[...]

PR #25021, opened 2025-12-11
fix(admin): guidedTourMeta can be undefined and error

Tofandel:
### What does it do?

Ensure guidedTourMeta.data is not undefined to avoid erroring when guidedTourMeta is [] (which is the case if you've already dismissed them)

### Why is it needed?
Without this in some cases you can get when you go to /admin
image

### How to test it?

### Related issue(s)/PR(s)

#24567

### DIFF
--- a/packages/core/admin/admin/src/components/GuidedTour/Overview.tsx
+++ b/packages/core/admin/admin/src/components/GuidedTour/Overview.tsx
@@ -172,7 +172,7 @@ export const GuidedTourHomepageOverview = () => {
   };
 
   if (
-    !guidedTourMeta?.data.isFirstSuperAdminUser ||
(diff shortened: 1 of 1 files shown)


markkaylor:
> Without this in some cases you can get when you go to /admin

@Tofandel Do you know what case specifically? Knowing the steps to reproduce helps us QA to ensure the fix addresses the expected issue.

Tofandel:
Not really, I think it was some kind of invalid auth state, where I had still authentication from before I deleted the database. From what I saw /auth/me was not failing but returning empty content

markkaylor:
@Tofandel just a heads up for future contributions we have some git commit conventions to follow: https://github.com/strapi/strapi/blob/develop/CONTRIBUTING.md#git-conventions

PR #25025, opened 2025-12-11
chore(.cursor): add worktrees support

nclsndr:
### What does it do?

Adds initialization instruction for [Cursor Git Worktrees](https://cursor.com/docs/configuration/worktrees) feature.

### Why is it needed?

This allow any contributor to launch Cursor local background agents.

### How to test it?

1. Open Cursor editor
2. Open a new chat
3. Prompt `Let's launch the @getstarted example`, select: "Worktree" (≠Local), submit
4. Cursor must create a worktree with dependencies installed and strapi built, allowing to launch the example app right away.

### DIFF
--- a/.cursor/worktrees.json
+++ b/.cursor/worktrees.json
@@ -1,0 +1,3 @@
+{
+  "setup-worktree": ["yarn install", "yarn build"]
+}


.cursor/worktrees.json
```diff
@@ -0,0 +1,3 @@
+{
+  "setup-worktree": ["yarn install", "yarn build"]
```
innerdvations, line 2: What do you think about just "yarn setup" ? That way it also does a clean build for safety, although it does take a bit longer

nclsndr: I started with it but ended up with a weird error. The debug being quite difficult through Cursor output panel I mostly went with what I found working

innerdvations (approved):

.cursor/worktrees.json
nclsndr: Ok just did a manual git worktree setup.
The root cause is that you cannot run any yarn script before launching `yarn install`.
So, given `"setup": "yarn && yarn clean && yarn build --skip-nx-cache"` vs a new worktree:
- `yarn` is not relevant since we need to run it at top level first
- `yarn clean` is not relevant since the new worktree starts fresh (fs wise)
- `yarn build --skip-nx-cache` > `yarn build` is sufficient since first run

Issue #25014, opened 2025-12-10
Strapi minor upgrade command failed


etqproksa:
### Node Version

22.18.0

### NPM/Yarn/PNPM Version

11.2.0

### Strapi Version

5.31.3

### Operating System

Strapi Cloud

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

When in used this command to upgrade strapi v5
npx @strapi/upgrade minor
 its shows the following message

<img width="960" height="293" alt="Image" src="https://github.com/user-attachments/assets/35225771-66d7-4519-aaf6-6b415e335869" />

### Steps to Reproduce

just updgrade the strapi v5 to the latest version and you will see this message using the above versions for node and npm

### Expected Behavior

It should upgrade the new version successfully

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

uigywnkiub:
Now works, but sometimes happens

PaulBratslavsky:
# Fix: Failed to parse URL from undefined/@strapi/strapi

Here is what seemed to work for me.

## File Location

**Main fix**: `packages/utils/upgrade/src/modules/npm/package.ts` (lines 65-71)

**Tests**: `packages/utils/upgrade/src/modules/npm/__tests__/package.test.ts`

---

## The Problem

When running `npx @strapi/upgrade latest` on projects using **Yarn Classic (v1)**, users encountered this error:

```
[ERROR] Failed to parse URL from undefined/@strapi/strapi
```

## Root Cause

The upgrade tool queries the package manager for its configured NPM registry URL. The code was using:

```typescript
const registryCommands = {
  yarn: ['config', 'get', 'npmRegistryServer'],  // Wrong for Yarn v1!
  npm: ['config', 'get', 'registry'],
};
```

**The issue**: Yarn Classic (v1) and Yarn Berry (v2+) use different config keys:

| Yarn Version | Config Key | Command |
|--------------|------------|---------|
| Yarn Classic (v1) | `registry` | `yarn config get registry` |
| Yarn Berry (v2+) | `npmRegistryServer` | `yarn config get npmRegistryServer` |

[...]

PR #25027, opened 2025-12-11
fix(upgrade): simplify registry URL resolution

PaulBratslavsky:
### What does it do?

Fix: Failed to parse URL from undefined/@strapi/strapi

## The Problem

When running `npx @strapi/upgrade latest` on projects using **Yarn Classic (v1)**, users encountered this error:

```
[ERROR] Failed to parse URL from undefined/@strapi/strapi
```

## Root Cause

The upgrade tool queries the package manager for its configured NPM registry URL. The code was using:

```typescript
const registryCommands = {
  yarn: ['config', 'get', 'npmRegistryServer'],  // Wrong for Yarn v1!
  npm: ['config', 'get', 'registry'],
};
```

**The issue**: Yarn Classic (v1) and Yarn Berry (v2+) use different config keys:

| Yarn Version | Config Key | Command |
|--------------|------------|---------|
| Yarn Classic (v1) | `registry` | `yarn config get registry` |
| Yarn Berry (v2+) | `npmRegistryServer` | `yarn config get npmRegistryServer` |

When Yarn Classic is asked for `npmRegistryServer` (which doesn't exist), it returns the **literal string `"undefined"`** - not an empty string, not an error, but the actual word "undefined".

This caused the registry URL to become:
```
undefined/@strapi/strapi
```

Which is not a valid URL, causing the fetch to fail.

### How to test it?

```bash
yarn test:unit
```

Tests cover:
- Custom registry via `NPM_REGISTRY_URL` environment variable
- Default registry fallback
- Trailing slash handling in registry URLs

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/25014


.vscode/settings.json
innerdvations, line 1: Can you revert the changes to this file please?

packages/utils/upgrade/src/modules/npm/package.ts
```diff
@@ -64,44 +62,12 @@ export class Package implements PackageInterface {
...
-    if (packageManagerRegistry) {
-      this.logger.debug(`Using package manager registry: ${packageManagerRegistry}`);
-      return packageManagerRegistry.replace(/\/$/, '');
-    }
-
-    this.logger.debug(`Using default registry: ${constants.NPM_REGISTRY_URL}`);
-    return constants.NPM_REGISTRY_URL.replace(/\/$/, '');
+  private getRegistryUrl(): string {
+    // Use NPM_REGISTRY_URL env var if set, otherwise use default registry
+    // This avoids spawning external processes (execa) to query package manager config
```
innerdvations: We want to maintain the original detection; we should be able to add your fix as an additional fallback in the original code though so that we support more package managers

innerdvations (changes requested):

packages/utils/upgrade/src/modules/npm/package.ts
PaulBratslavsky: Thank you for the feedback.   🙂

Artem2001S:
Temporary solution: 

```
export NPM_REGISTRY_URL=https://registry.npmjs.org/
npx @strapi/upgrade major
```

Adzouz:
Hey @PaulBratslavsky and @innerdvations,
Could you give us an update on this one please? If it's a quick win allowing us to close both an issue + a PR, I'd propose we merge it 😊

innerdvations (approved):

PR #25028, opened 2025-12-12
fix: adapt header height when title changes

Adzouz:
### What does it do?

Dynamically update the height of the header when title changes.

Before:

https://github.com/user-attachments/assets/ffb7e033-5587-46a2-8946-025031bf1341

After:

https://github.com/user-attachments/assets/03d85705-fb76-46f5-a67a-6b6b247d9da0

### Why is it needed?

When updating the title for a longer title (two lines), the header is cut behind the content and the action button can be inaccessible.

### How to test it?

- Go to Profile
- Change the name for something very long
-> The header height should change based on the length of the content

🚀

### DIFF
--- a/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
+++ b/packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
@@ -138,9 +138,12 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
     threshold: 0,
   });
 
-  useResizeObserver([containerRef], () => {
-    if (containerRef.current) {
-      const newSize = containerRef.current.getBoundingClientRect();
+  useResizeObserver([containerRef, baseHeaderLayoutRef], () => {
+    const headerContainer = baseHeaderLayoutRef.current ?? containerRef.current;
+
+    if (headerContainer) {
+      const newSize = headerContainer.getBoundingClientRect();
+
       setHeaderSize((prevSize) => {
         // Only update if size actually changed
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
```diff
@@ -138,9 +129,12 @@ const HeaderLayout = (props: HeaderLayoutProps) => {
     threshold: 0,
   });
 
-  useResizeObserver([containerRef], () => {
-    if (containerRef.current) {
-      const newSize = containerRef.current.getBoundingClientRect();
+  useResizeObserver([containerRef, baseHeaderLayoutRef], () => {
+    const measuredElement = baseHeaderLayoutRef.current ?? containerRef.current;
```
mathildeleg: Not sure about this naming, I understand what you mean in the context of this PR but otherwise not sure I would get it personally

mathildeleg (changes requested):
It works well for what you wanted to achieve, however it means when I resize my window the title no longer resizes to fit in the window :/

packages/core/admin/admin/src/components/Layouts/HeaderLayout.tsx
Adzouz: I changed the naming but I tested on my side and I don't have any issue when resizing the window. Don't hesitate to poke me if you want to talk about it and test together!

https://github.com/user-attachments/assets/99c9215a-c2a6-40dc-99c0-a20493445145

mathildeleg (approved):
✅

PR #25029, opened 2025-12-12
fix: excessive rerenders on edit view

### LINKED ISSUE #24462: Admin performance issue (input lag)
### Node Version

22.14.0

### NPM/Yarn/PNPM Version

YARN 1.22.21

### Strapi Version

5.24.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

While using Strapi, I noticed performance issues when working with large documents. The more content I add, the stronger the input lag becomes.
When I type quickly in a field, each character takes noticeable time to appear. The more characters there are in the field, the slower I have to type to avoid these delays.
I’m attaching a video to demonstrate the issue.

To rule out any environment-related causes, I ran additional tests:

Running Strapi in production mode
Testing on multiple browsers
Each test gave me the same result.

Do you have a solution to work around this problem?

Thanks in advance!

(In the video, I am typing continuously.)
https://github.com/user-attachments/assets/cc372cd1-cd6f-4197-a0cc-637e88e7938a

### Steps to Reproduce

- Have a content-type with a dynamic zone
- Add a new entry
- Add a component to the dynamic zone (even an empty one is enough)
- Type quickly in another field

### Expected Behavior

The letters should be displayed instantly on each input.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

remidej:
### What does it do?

Prevents excessive re-renders on the CM edit view.

The problem came from the conditional fields feature. In order to check if an input should be rendered, it needs to subscribe to all other fields on the page, which is very expensive. My proposed fix is to only do that when an attribute does have a condition set up, so that we avoid the performance issue for the vast majority of fields that aren't conditionally rendered.

### Why is it needed?

Performance, input lag.
### How to test it?

**Excessive rerenders**
I recommend using react-scan to identify the issue, either via the browser extension or the npx command. It highlights all the components that get rerendered. You'll see that this is what happens on develop:

https://github.com/user-attachments/assets/df2ebb54-e426-45fa-9ad2-992e81b73af1

Compared to this branch:

https://github.com/user-attachments/assets/1e3f4b7f-2098-4547-99f8-38a91737438e

**Conditional rendering**

Also gotta make sure we haven't broken the conditional field feature. Use the 'Condition' content type on getstarted for that, you should see a field appear and disappear based on the boolean field.
### Related issue(s)/PR(s)

fix #24462

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -5,6 +5,8 @@ import {
   useForm,
   InputRenderer as FormInputRenderer,
   useField,
+  createRulesEngine,
+  type JsonLogicCondition,
 } from '@strapi/admin/strapi-admin';
 import { useIntl } from 'react-intl';
 
@@ -41,7 +43,7 @@ type InputRendererProps = DistributiveOmit<EditFieldLayout, 'size'> & {
  * the complete EditFieldLayout and will handle RBAC conditions and rendering CM specific
  * components such as Blocks / Relations.
  */
-const InputRenderer = ({
+const BaseInputRenderer = ({
   visible,
   hint: providedHint,
   document,
@@ -183,7 +185,7 @@ const InputRenderer = ({
           disabled={fieldIsDisabled}
         >
           {(componentInputProps) => (
-            <InputRenderer
+            <BaseInputRenderer
               key={`input-${componentInputProps.name}-${localeKey}`}
               {...componentInputProps}
             />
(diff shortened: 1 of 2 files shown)


mathildeleg (approved):
Tested ✅ works really well 🚀

packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -329,7 +351,22 @@ const getMinMax = (attribute: Schema.Attribute.AnyAttribute) => {
...
+ * (when there's a JSON logic condition on the attribute, or BaseInputRenderer otherwise.
+ * We do this because rendering a conditional field requires access to the values of
+ * other form fields, which causes many re-renders and performance issues on complex content
+ * types. By splitting the component into two, we isolate the performance issue to
+ * conditional fields only, not all edit view fields.
+ */
+const MemoizedInputRenderer = React.memo((props: InputRendererProps) => {
+  const condition = props.attribute.conditions?.visible;
+  if (condition) {
+    return <ConditionAwareInputRenderer {...props} condition={condition} />;
```
Adzouz, line 365: So if I understand correctly, only the conditionned fields now will be rerendered on any input change on a form?

Adzouz: Let's just hope users won't have just conditional fields in a form, otherwise the enhancement won't be improving much but I guess that in that case it's hard to do better 😕

Sanskarkasoudhan: Hey adzouz , so on upgrading strapi v5 would not help much.

PR #25034, opened 2025-12-13
fix(upload): add pagination support to /api/upload/files endpoint

### LINKED ISSUE #24624: /api/upload/files get endpoint doesn't support pagination

AdarshBennur:
The /api/upload/files GET endpoint was not supporting pagination parameters (page, pageSize, withCount) unlike other Strapi API endpoints. This caused issues for users with many files in their media library.

Changed the content API controller to use the existing `findPage` service instead of `findMany`, which provides built-in pagination support. Updated the response format to match Strapi's standard API structure with `data` and `meta.pagination` fields.

This change includes:
- Modified content-api controller to use `findPage()` service method
- Updated response format to return {data: [...], meta: {pagination: {...}}}
- Added comprehensive unit tests (11 test cases) covering default pagination, custom parameters, edge cases, validation, and sanitization

Breaking Change: The response format has changed from a flat array to the standard Strapi format with data and meta objects. Existing API consumers will need to access `response.data` instead of using the response directly.

Fixes #24624


packages/core/upload/server/src/controllers/__tests__/content-api.test.ts
```diff
@@ -0,0 +1,360 @@
...
+        },
+      },
+    } as any;
+
+    mockContext = {
+      query: {} as any,
+      state: { auth: {} },
+      body: undefined as any,
+      notFound: jest.fn(),
+    } as any;
```
Adzouz: You could replace by `as unknown as Context` and get rid of all useless assertions further in the document (remove all `as Context` on lines `await controller.find(mockContext as Context);`)

packages/core/upload/server/src/controllers/__tests__/content-api.test.ts
```diff
@@ -0,0 +1,360 @@
...
+        pageCount: 1,
+        total: 2,
+      };
+
+      mockUploadService.findPage.mockResolvedValue({
+        results: mockFiles,
+        pagination: mockPagination,
+      });
+
+      const controller = contentApiController({ strapi: global.strapi as any });
```
Adzouz: As the warning says, replace all occurences of `global` by `globalThis`

packages/core/upload/server/src/controllers/__tests__/content-api.test.ts
```diff
@@ -0,0 +1,360 @@
...
+
+import contentApiController from '../content-api';
+import { getService } from '../../utils';
+
+jest.mock('../../utils');
+
+const mockGetService = getService as jest.MockedFunction<typeof getService>;
+
+describe('Content API Controller - Pagination', () => {
+  let mockContext: Partial<Context>;
```
Adzouz: Replace `Partial<Context>` by just `Context`

Adzouz (changes requested):
Hi @AdarshBennur, thanks a lot for the contribution 🙏 I left a few comments.
Also to be consistent with the rest of API endpoints, the params to the URL should be with the following structure: `pagination[page]=1&pagination[pageSize]=50&pagination[withCount]=true` instead of `page=1&pageSize=50&withCount=true`. Do you think you can update what you've done to fit the rest of endpoints?

Adzouz:
Thanks for resolving the comments @AdarshBennur 🙏 But the endpoint is still not consistent with what already exist and parameters should be part of pagination object and not directly provided to the url.
`pagination[page]=1&pagination[pageSize]=50&pagination[withCount]=true` instead of `page=1&pageSize=50&withCount=true`

AdarshBennur:
Hi @Adzouz, thank you for the detailed feedback! 🙏

You were absolutely right - the endpoint wasn't properly validating the paginated response format. 

## What was wrong:
The route was configured to accept `pagination[page]=1&pagination[pageSize]=50` query parameters, and the controller was correctly returning `{data: [...], meta: {pagination: {...}}}`, but the **response schema** was still expecting a flat array. This caused a validation mismatch.

## What I fixed:
1. Added a new `paginatedFiles` response schema in `routes/validation/upload.ts` that validates the `{data, meta.pagination}` structure
2. Updated the `/files` GET route to use this new schema instead of the flat array schema

## Result:
✅ Endpoint now accepts: `pagination[page]=1&pagination[pageSize]=50&pagination[withCount]=true`  
✅ Response validates correctly: `{data: [...], meta: {pagination: {...}}}`  
✅ Fully consistent with Strapi's standard pagination format  
✅ All 106 tests passing

The pagination parameters are now properly integrated end-to-end (route validation → controller logic → response validation).

Adzouz:
Hi @AdarshBennur,
Thanks a lot for taking the time again but I still have an issue while testing on my side...
If I call the endpoint and pass the pagination parameter it's still not taking it into account:
Screenshot 2026-01-02 at 15 49 59
Can you tell me if that's the same case on your side? 🙏

colinduwe:
I was able to reproduce your result with those parameters
http://localhost:1337/api/upload/files?pagination[pageSize]=2
Screenshot 2026-01-08 at 10 06 20 AM

However, this works for me with this param format
http://localhost:1337/api/upload/files?pageSize=2&page=2&sort=createdAt:DESC
Screenshot 2026-01-08 at 10 08 40 AM

Adzouz:
Hi @colinduwe, yes same on my side.
The first format is the one expected, not the second. My comments were asking to fix that to make it consistent with the rest of the API endpoints (using pagination[] format).

Also, since it's a pretty serious breaking change, we'll have to discuss with the team about the next steps. We might need to create another experimental endpoint instead and use future flag.

colinduwe:
I spent some time yesterday trying to solve this but ended up with a bit of an ugly mess. I had to copy/paste code from the content API since the controller doesn't have the ability to work with the standard API params. I've pushed my changes to my prior [PR attempt](https://github.com/strapi/strapi/pull/24711) (please don't judge me by the mess)

However, I did implement a config flag in my PR so users can opt in to this change. Perhaps we need a deprecation notice in the log when using it without the flag.

packages/core/upload/server/src/controllers/content-api.ts
```diff
@@ -38,9 +38,19 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
       await validateQuery(ctx.query, ctx);
       const sanitizedQuery = await sanitizeQuery(ctx.query, ctx);
 
-      const files = await getService('upload').findMany(sanitizedQuery);
```
Benjamin-K, line 42: What about sth like this (maybe with a deprecation note) to keep the existing implementation without breaking existing code:

```
if (!!sanitizedQuery.pagination) {
    // User requested paginated response, return new format
} else {
    // Keep existing response format.
}

AdarshBennur:
Hi @Adzouz, @colinduwe,

I’ve pushed a fix that addresses the root cause of the issue and verified it end-to-end.

The problem was that `pagination[]` params were correctly parsed into a nested object, but the upload service wasn’t flattening them before calling `transformQueryParams`, so defaults were always applied. I’ve fixed this in the service layer, aligning the behavior with how the Content API handles pagination.

I verified the fix both via tests and at runtime:
- `pagination[page]`, `pagination[pageSize]`, and `pagination[withCount]` now correctly affect the results
- The original repro (`?pagination[pageSize]=2` being ignored) is now fixed
- Response shape matches the standard `{ data, meta: { pagination } }`
- Existing behavior remains unchanged when no pagination is provided

Happy to adapt this further if you’d prefer this to be gated behind a flag or moved to an experimental endpoint.

colinduwe:
I finally had a chance to test this and found pagination[pageSize] and pagination[page] to work as expected. 

However, other pagination params, start, limit, and withCount do not appear to be affecting the response.

giu1io:
Hey @AdarshBennur, thank you for the PR. As part of our effort in improving the media library we merged an equivalent fix as the one you proposed in PR #26597 and it's now available from version `5.48.1`. 

You can find the new endpoint documentation [here](https://docs.strapi.io/cms/api/rest/upload#get-a-paginated-list-of-files).

Hope the solution as implemented works for you. 

Thank you again for your help 🙏

PR #25041, opened 2025-12-16
ci(prerelease): use github trusted publisher instead of token

innerdvations:
### What does it do?

stop using NPM_TOKEN now that we've configured github as a trusted publisher on npm

### Why is it needed?

to do experimental releases

### How to test it?

run the "prerelease" action (after merging this) and ensure it works now


innerdvations:
After discussing, this won't work; will reapproach differently

Issue #24646, opened 2025-10-21
Missing or invalid credentials


dhoko:
### Node Version

22.16.0

### NPM/Yarn/PNPM Version

? (cloud)

### Strapi Version

5.25.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Reproduction URL

*No response*

### Bug Description

👋 hello

We *often* see this error `Missing or invalid credentials` while working with Strapi. It's something "new", never seen it before but with this current version it's often. Maybe with the previous one as well (*I don't remember tbh, but I would say same as it's older than our update*).

### Steps to Reproduce

It's not something time based (*inactive*). <br>I'm able to see it if I'm active while editing a page, uploading a media while editing a page or doing something another tab (strapi) while the main tab (strapi) is doing a lot of requests to a plugin API.

It doesn't matter at somepoint the session is stilll *active*, you can still navigate etc. Yet we get this error.

### Expected Behavior

We don't get this error.

### Logs

```shell
nothing from the logs except from the console cf screenshot
```

### Code Snippets

*No response*

### Media

!https://github.com/user-attachments/assets/3e61d24a-f290-4db7-b427-de58721556eb

### Additional information

*No response*

### Confirmation Checklist

- [X] I have checked the existing [issues](<https://github.com/strapi/strapi/issues>) for duplicates.
- [X] I agree to follow this project's [Code of Conduct](<https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md>).

agzaiyenth:
Hi @dhoko ,
Can you assign me to this issue?

oiorain:
Hi @dhoko,

Thank you for opening an issue.
We don't have enough data to reproduce this issue from your description. Could you provide more or open a new issue?

dhoko:
@oiorain that's the issue. We can't isolate a pattern to replicate it. It's just something we see often for no specific reason, even while we're active.  It's _everywhere_ it's not for a specific request. 

The only thing is, it seems to pop only after maybe 30m working on strapi. Why? No idea.
I work with the 5.29, I can still see it, even if it seems to be less common than before.

I know it's not enough, but fact is, something is wrong 😢

boazpoolman:
Reopening this issue as I've been able to reproduce.

[...]

PR #25053, opened 2025-12-18
fix: add token refresh logic for react query

butcherZ:
### What does it do?

Moved the token refresh logic from `baseQuery.ts` into `getFetchClient.ts` so that all API calls (both RTK Query and React Query) benefit from automatic token refresh on 401 errors. Added tests for token refresh

### Why is it needed?

  The Bug Flow

  1. User logs in, JWT token is stored (expires in 30 minutes)
  2. User works in admin panel for 30+ minutes
  3. Token expires
  4. Scenario A (works): User triggers RTK Query endpoint → 401 → baseQuery.ts refreshes token → retry succeeds
  5. Scenario B (broken): User uploads file via Media Library → React Query + useFetchClient → 401 → No refresh → Error shown: "Missing or invalid credentials"

### How to test it?

  A. Run the Unit Tests

  `yarn nx run @strapi/admin:test:front -- --testPathPattern="getFetchClient.test" `

  B. Manual Testing 

experimental: `0.0.0-experimental.bca2f43ed2b0638d8f24d8374ed1d687a14e4981`

  1. Modify the JWT expiration time in your Strapi project's config/admin.js:

```
  module.exports = {
    auth: {
      secret: process.env.ADMIN_JWT_SECRET,
    sessions: {
      accessTokenLifespan: 60, // 1 minute in seconds
  },
    },
    // ... rest of config
  };
```

  2. Start your Strapi dev server:
  `yarn develop`

  3. Test the flow:
    - Log into the admin panel
    - Wait ~1 minute for the token to expire
    - Try uploading a file to the Media Library
    - Expected (with fix): Upload succeeds, token is refreshed automatically
    - Before fix: Would show "Missing or Invalid Credentials" error
  4. Also test these scenarios:
    - Edit an existing asset after token expires
    - Bulk move/delete assets after token expires
    - Navigate around the admin (RTK Query requests) after token expires

  | Scenario                       | Action                       | Expected Result                                    |
  |--------------------------------|------------------------------|----------------------------------------------------|
  | Upload after token expires     | Upload file in Media Library | Succeeds (auto-refresh)                            |
  | Edit asset after token expires | Edit asset metadata          | Succeeds (auto-refresh)                            |
  | Login with wrong password      | Submit login form            | Shows "Invalid credentials" (no refresh attempted) |
  | Refresh token also expired     | Any action                   | Logs user out gracefully                           |
  | Multiple concurrent 401s       | Rapid file uploads           | Only one refresh request made                      |

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/24646
dx-2286
fixes CMS-27

### DIFF
--- a/docs/docs/docs/01-core/content-manager/06-preview.md
+++ b/docs/docs/docs/01-core/content-manager/06-preview.md
@@ -1,0 +1,88 @@
+---
+title: Live preview
+description: Frontend-agnostic preview with visual editing
+tags:
+  - content-manager
+---
+
+The live preview feature lets users see their content rendered on their frontend while editing. It includes visual editing that identifies and highlights editable fields.
+
+## Why not an SDK
+
+Visual editing requires running some of our code on the user's frontend to detect fields and draw highlights. The obvious approach would be an SDK package users install in their project. We intentionally avoided this.
+
+An SDK would require ongoing maintenance and create version mismatch risks between the SDK and Strapi. It would also tie us to specific frameworks or require multiple framework-specific packages.
+
+Instead, the preview script is defined inside Strapi and sent to the frontend via `postMessage`. The frontend just needs a small snippet to receive and execute it. This keeps the script always in sync with the CMS version, works with any framework, and requires no package installation.
+
+## How the script works
+
+### Self-contained constraint
+
+The preview script (`packages/core/content-manager/admin/src/preview/utils/previewScript.ts`) is stringified before being sent to the iframe:
+
+```ts
+const script = `(${previewScript.toString()})(${JSON.stringify(config)})`;
+```
+
+Because of this, it **cannot import dependencies or reference external variables**. All logic must be self-contained. The only external code (`@vercel/stega` for decoding) is loaded dynamically from a CDN at runtime.
+
+This is why the file has an unusual structure with many functions defined inline.
+
+### Field identification with stega
+
+We use [stega encoding](https://github.com/vercel/stega) to identify which Strapi field each piece of text comes from. Stega embeds invisible metadata into text content using Unicode zero-width characters that are imperceptible to users but can be decoded programmatically.
+
+1. The Document Service encodes field metadata into text values (invisible to users)
+2. The frontend renders the content normally
+3. The preview script decodes the metadata and attaches `data-strapi-source` attributes to DOM elements
+4. Highlights are drawn over elements with source attributes
+
+The metadata uses URL search params format, because it makes it easy to encode and decode multiple pieces of information into a single string: `path=title&type=string&documentId=abc123&locale=en&model=api::page.page`
+
+### Stega limitations
+
+Stega can only encode strings. This means:
+
+- **Blocks fields don't support visual editing** — their content is a JSON object, not a string. Supporting them would require a custom implementation.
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -23,6 +24,85 @@ const STORAGE_KEYS = {
...
+ */
+const storeToken = (token: string): void => {
+  // Check if the original token was stored in localStorage (persist mode)
+  const wasPersistedToLocalStorage = Boolean(localStorage.getItem(STORAGE_KEYS.TOKEN));
+
+  if (wasPersistedToLocalStorage) {
+    localStorage.setItem(STORAGE_KEYS.TOKEN, JSON.stringify(token));
+  } else {
+    setCookie(STORAGE_KEYS.TOKEN, token);
+  }
```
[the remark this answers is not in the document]
Bassel17: sorry Claude I don't think we can do that yet, we still need the javascript to have access to those until we build a different auth method

innerdvations (commented):
Awesome! The idea is solid, I'm only not clicking "approve" because I don't have time to test it before I'm off, and I want someone front-end to take a closer look in case I'm missing something :)

packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -166,6 +255,35 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
...
+  /**
+   * Execute a fetch request with automatic token refresh on 401 errors.
+   * @param url - The request URL (used to check if it's an auth path)
+   * @param executeFetch - Function that performs the actual fetch
+   * @param executeRetry - Function that performs the retry with fresh headers
+   */
+  const withTokenRefresh = async <TData>(
+    url: string,
+    executeFetch: () => Promise<FetchResponse<TData>>,
+    executeRetry: () => Promise<FetchResponse<TData>>
```
Bassel17: this sounds redundant the executeFetch and executeRetry seem to be the same,
is there a reason to have them separated ?

packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -23,6 +24,129 @@ const STORAGE_KEYS = {
...
+ *
+ * Uses localStorage if the user selected "remember me" during login,
+ * otherwise uses cookies for session-based storage.
+ *
+ * @param token - The JWT token to store
+ * @internal Exported for testing purposes
+ */
+const storeToken = (token: string): void => {
+  // Check if the original token was stored in localStorage (persist mode)
+  const wasPersistedToLocalStorage = Boolean(localStorage.getItem(STORAGE_KEYS.TOKEN));
```
jhoward1994, line 72: Just a thought, is this a reliable enough way to know we're in remember me mode?
Maybe there's a more specific way to check this ?

Bassel17: I don't think their is another way to check this, it seems we only have remember me as a value on login that we don't store anywhere else, so technically that's the only way to check if we are in `presist mode` or not

PR #25064, opened 2025-12-19
chore: use design-system variables for preview

mathildeleg:
### What does it do?

Use the design system to set the colors used in Preview rather than hard-coded hex codes.

### Why is it needed?

Allows for easier maintainability of the code.

### How to test it?

* Set `PREVIEW_ENABLED=true` in your .env
* Go to any content type, click on an existing entry (or create a new entry, fill it out and click save)
* Click on the Preview button on the right side panel.
* Hover over a text field -> in Preview, that text field should be highlighted in a dark blue color.

### DIFF
--- a/packages/core/content-manager/admin/src/preview/utils/constants.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/constants.ts
@@ -3,7 +3,12 @@ import { MessageDescriptor } from 'react-intl';
 
 import { previewScript } from './previewScript';
 
-const scriptResponse = previewScript(false);
+export const PREVIEW_HIGHLIGHT_COLORS = {
+  highlightHoverColor: 'transparent',
+  highlightActiveColor: 'transparent',
+} as const;
+
+const scriptResponse = previewScript({ shouldRun: false, colors: PREVIEW_HIGHLIGHT_COLORS });
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/preview/utils/constants.ts
```diff
@@ -1,9 +1,15 @@
 import { NotificationConfig } from '@strapi/admin/strapi-admin';
+import { darkTheme } from '@strapi/design-system';
 import { MessageDescriptor } from 'react-intl';
 
 import { previewScript } from './previewScript';
 
-const scriptResponse = previewScript(false);
+export const PREVIEW_HIGHLIGHT_COLORS = {
+  highlightHoverColor: darkTheme.colors.primary500,
+  highlightActiveColor: darkTheme.colors.primary600,
```
remidej: it would be nice to use the light theme colors when the app is in light mode

mathildeleg: Fixed here https://github.com/strapi/strapi/pull/25064/changes/d95261acc350e09372fe4fe4c63809d9edd6dcb8#diff-47717cef848eb6089b5e4629aad93c7b2d2acf1b7ef9c22e3cc5bad95e786c59R139 :)

PR #25066, opened 2025-12-19
feat(upload): add retroactive ai metadata generation

markkaylor:
### What does it do?

Add backend endpoints:

- (C-719) Add POST endpoint to retroactively update all existing files with AI generated metadata
- (C-718) Add GET endpoint to fetch the number image missing alternative text OR caption 

Update frontend settings page:

- (CG-722) Add button to call POST endpoint for retroactive generation 

Currently I have this marked with a BETA badge. I figure we could leave it like that until we finish the redesign work where we will improve this experience. Could also remove it though or put the whole thing under a future flag.

Not sure what y'all think of it in the SettingsPage, it wouldn't be much work to do something similar to i18n actually, but I do not want to add the review modal originally proposed, we will improve it with the redesign. Personally I'm fine with this and getting feedback from users but it's still up for discussion 😄 

### Why is it needed?

To give users the ability to update their existing images that were not uploaded with the AI feature

### How to test it?

I've tested updating 100 images (jpg and png). If you need some images ping me on slack.

1. Turn AI off in settings
2. Go upload a bunch of images (jpg, png)
4. Go back to settings and turn AI on
5. You should see the number of images missing metadata
6. Click the "generate metadata for existing images" button
7. You should be able to move around the app while the job works in the background. You can create and publish content for example. Otherwise just hang out while it does it's thing.

Sorry for the big PR

https://github.com/user-attachments/assets/0da038e1-b70c-4e53-8131-6cb006add60c


packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -27,9 +70,139 @@ const createAIMetadataService = ({ strapi }: { strapi: Core.Strapi }) => {
...
+              errors.push({
+                fileId: file.id,
+                error: error instanceof Error ? error.message : 'Unknown error updating file',
+              });
+            }
+          }
+        })
+      );
+
+      return { processed, errors };
```
markkaylor: ℹ️ I had refactored to removed these or at least I thought I did, will update.

packages/core/upload/server/src/controllers/__tests__/admin-upload.test.ts
```diff
@@ -55,6 +55,38 @@ describe('Admin Upload Controller - AI Service Connection', () => {
...
+                    fileId: file.id,
+                    error: error instanceof Error ? error.message : 'Unknown error updating file',
+                  });
+                }
+              }
+            })
+          );
+
+          return { processed, errors };
+        }),
```
markkaylor: I don't think this makes any sense

packages/core/upload/admin/src/hooks/useAIMetadataJob.ts
```diff
@@ -0,0 +1,97 @@
...
+      enabled,
+      // Poll every second when job is processing
+      refetchInterval: (data) => {
+        // If no data yet, don't poll
+        if (!data) return false;
+
+        // Poll while processing
+        if (data.status === 'processing') {
+          return 1000;
+        }
```
remidej, line 47: That's a lot smoother than our RTK solution 👌

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -19,16 +40,155 @@ import { initialState, reducer } from './reducer';
...
+const GenerateMetadataButton = styled(Button)`
+  > span {
+    color: ${({ theme }) => theme.colors.primary600};
+  }
+
+  &:disabled {
+    background: transparent;
+    border: none;
+  }
+`;
```
remidej: Have you checked how it would work with a TextButton instead of a ghost button?

It's more of a feedback on the design than on your code but I'm not a fan of bespoke versions of things like buttons that should be standard

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -184,6 +379,41 @@ export const SettingsPage = () => {
...
                     </Grid.Root>
+                    {/* Retroactive metadata generation when aiMetadata is enabled */}
+                    {initialData?.aiMetadata && !isLoadingMetadataCount && (
+                      <>
+                        <Divider marginTop={4} marginBottom={4} />
+                        <Flex justifyContent="space-between" alignItems="center" gap={2}>
+                          <Flex gap={2}>
+                            <BetaStatus size="XS" />
+                            <Typography variant="pi" textColor="neutral500">
+                              {metadataCount === 0
```
remidej: metadataCount could have a more descriptive name, it's not clear that it's about the remaining assets

packages/core/upload/server/src/services/ai-metadata-jobs.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  async getLatestActiveJob(): Promise<AIMetadataJob | null> {
+    // Return the most recent job, regardless of status
+    // This allows the frontend to see completed/failed status
+    return strapi.db.query(AI_METADATA_JOB_UID).findOne({
+      orderBy: { createdAt: 'desc' },
+    });
+  },
+
+  async cleanupOldJobs(): Promise<number> {
```
remidej: _Old_ is a bit ambiguous since you're deleting all existing processing jobs, it's not about their age

packages/core/upload/server/src/services/ai-metadata-jobs.ts
```diff
@@ -0,0 +1,81 @@
...
+    const oldJobs = await strapi.db.query(AI_METADATA_JOB_UID).findMany({
+      where: {
+        status: { $ne: 'processing' },
+      },
+    });
+
+    for (const job of oldJobs) {
+      await this.deleteJob(job.id);
+    }
```
remidej: There may be a query engine method to do these in one step

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -1,7 +1,50 @@
...
+    logger.error('Failed to fetch image', {
+      fullUrl,
+      status: resp.status,
+      statusText: resp.statusText,
+    });
+    throw new Error(`Failed to fetch image from URL: ${fullUrl} (${resp.status})`);
+  }
+
+  const arrayBuffer = await resp.arrayBuffer();
+  return new Blob([arrayBuffer], { type: file.mimetype || undefined });
```
remidej: Were we using blobs for this before? Does it require any ai server changes?

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -19,16 +40,155 @@ import { initialState, reducer } from './reducer';
...
+            id: getTrad('settings.form.aiMetadata.generatingMetadata'),
+            defaultMessage: 'AI is generating your metadata',
+          })}
+        </Typography>
+      </Flex>
+    );
+  }
+
+  // Only show completed state when all images have metadata
+  if (metadataCount === 0) {
```
Adzouz: I know I'm being picky but if there's no asset in the library, we would also have the message that the metadata has been generated. Maybe we could customize the message in that case? If it's too much because it would mean add another API call, let's keep it like that

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -19,16 +40,155 @@ import { initialState, reducer } from './reducer';
...
+              defaultMessage: 'Generate AI Metadata',
+            })}
+          </Flex>
+        }
+      >
+        {formatMessage(
+          {
+            id: getTrad('settings.form.aiMetadata.confirmDialog.message'),
+            defaultMessage:
+              'This will generate captions and alternative text for {count, plural, one {# image} other {# images}}. AI can make mistakes, be sure to review the generated content. ',
```
Adzouz: Maybe add here an explanation on how the process is running (in the background) and that the user can leave the page and come back later on the page to see if the generation is finished?

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -1,7 +1,50 @@
...
 import { z } from 'zod';
-import { InputFile } from '../types';
+import { InputFile, File } from '../types';
 import { Settings } from '../controllers/validation/admin/settings';
+import { getService } from '../utils';
+
+/**
+ * Fetches an image from a URL and returns it as a Blob
+ */
+async function fetchImageAsBlob(
```
Adzouz: I know it's related to AI metadata task but is it relevant to put this method at this place? Is there a file more dedicated to images / assets instead?

packages/core/upload/server/src/services/ai-metadata.ts
```diff
@@ -1,7 +1,50 @@
...
+  }
+
+  const arrayBuffer = await resp.arrayBuffer();
+  return new Blob([arrayBuffer], { type: file.mimetype || undefined });
+}
+
+/**
+ * Builds FormData from an array of input files by fetching each image
+ */
+async function buildFormDataFromFiles(
```
Adzouz: Same for this one

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
markkaylor: I'll take a look

packages/core/upload/server/src/services/ai-metadata.ts
markkaylor: Sure I could move it to an utils file

packages/core/upload/server/src/services/ai-metadata.ts
markkaylor: This should all be the same just moved around, what was done on this PR should require no update to the ai server

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
markkaylor: I did not know about TextButton 🤦‍♂️  thanks!

packages/core/upload/admin/src/pages/SettingsPage/SettingsPage.tsx
markkaylor: I think I have a solution that avoids another API call. It's still probably not the best, but if the latest job is marked as completed and there are 0 images without metadata display the success message, if the latest job is not completed and there are 0 images disable the button

markkaylor: I have a better idea 🏃

markkaylor:
@Adzouz @remidej 

Thanks for the reviews, I think I handled all of your feedback. 

I'm having some doubt and wondering what y'all think. Should we:
1. Keep the beta badge without a future flag?
2. Keep the beta badge but put the feature under a future flag?
3. Remove the beta badge entirely(aka yolo)?

Technically it calls the same function as the already stable ai generation on upload feature. I've tried what I have currently with 200ish images and had no problems, but I'm not sure if this thing really scales. I tried to explore refactoring to streaming yesterday but was unsuccessful, probably need to PP with someone. We will have to move to streaming when we do the redesign work. Thanks for your opinions here.

Adzouz:
> @Adzouz @remidej
> 
> Thanks for the reviews, I think I handled all of your feedback.
> 
> I'm having some doubt and wondering what y'all think. Should we:
> 
> 1. Keep the beta badge without a future flag?
> 2. Keep the beta badge but put the feature under a future flag?
> 3. Remove the beta badge entirely(aka yolo)?
> 
> Technically it calls the same function as the already stable ai generation on upload feature. I've tried what I have currently with 200ish images and had no problems, but I'm not sure if this thing really scales. I tried to explore refactoring to streaming yesterday but was unsuccessful, probably need to PP with someone. We will have to move to streaming when we do the redesign work. Thanks for your opinions here.

My opinion: I would keep the beta tag and in that case no need for a future flag. It's either that or you put the future flag and remove the beta tag.
The future flag wouldn't really make sense because this feature will be removed from there to join the revamp of the media lib so it's not really a feature for the future if you know what I mean.

markkaylor:
> The future flag wouldn't really make sense because this feature will be removed from there to join the revamp of the media lib so it's not really a feature for the future if you know what I mean.

I think it would still make sense because we will have the same feature in the revamp but presented in a different way. My concern with the future flag is more that nobody will actually use it. 

I'm still cool with releasing as it is now if everyone is aligned with that. 

Thanks for your thoughts!

markkaylor:
Need to add tracking

markkaylor:
@Adzouz I pushed to address 1 issue we found in the blitz: https://github.com/strapi/strapi/pull/25066/commits/952e4f33ce82905538c7fba2ea240199dc859d53 otherwise if it looks good to you could I get a ✅ ?

PR #25069, opened 2025-12-21
fix: Added multiple prop for MediaLibraryDialog

MaksZhukov:
### What does it do?

It adds support multiple prop for MediaLibraryDialog component

### Why is it needed?

To have ability to use MediaLibraryDialog with ability to select only 1 asset

### How to test it?

Use MediaLibraryDialog with **multiple** false


MaksZhukov:
@remidej @Adzouz Hey! Take a look, please.

MaksZhukov:
@Adzouz @remidej Hey!

Adzouz:
Hi @MaksZhukov, I have to admit I don't really see the point of this PR. Can you explain a bit more what are you trying to achieve here? Is that because you're using MediaLibraryDialog somewhere in your project?

MaksZhukov:
@Adzouz Hey! I use the MediaLibraryDialog in my plugin and I want to have ability to select only 1 asset, now there is no option to do it.

Adzouz:
Ok fair enough @MaksZhukov, looks good to me!
Just two things before approving:
- Can you adapt your commit message to follow our format convention? (it's just that it must be lowercase)
- Can you update your branch with develop please? 🙏

Adzouz:
Oh sorry, also can you change `feat` by `fix` please? In our workflow, a feat would bump to a minor version and this improvement is more like a patch than a actual new feature.

MaksZhukov:
@Adzouz Done

PR #25081, opened 2025-12-23
docs: Highlight destructive operation in transfer engine

Zyggzzz:
What does it do?

- Updates the Data Transfer Engine conceptual guide to explicitly warn about destructive operations.
- Adds a [!CAUTION] alert box regarding data loss when using the default restore strategy.
- Updates the Mermaid sequence diagrams to include the beforeTransfer phase and the #deleteAll (wipe) action in the Destination Provider.
- Adds a specific section explaining the "Restore Strategy" and its impact on existing destination data.

Why is it needed?

Currently, the documentation does not explicitly state that the local-destination provider wipes the destination instance before transferring data. Users have to audit the source code (specifically the #deleteAll() call in the provider's bootstrap phase) to realize the operation is destructive. Adding this to the docs prevents accidental data loss for users expecting a merge-like behavior.
### How to test it?

- cd strapi/docs
- yarn build
- yarn start
- check docs on `http://localhost:3000/`

### Related issue(s)/PR(s)

N/A


innerdvations:
Thanks for adding this warning — it's a good change and the engine guide was missing it.

I pushed a small follow-up commit to align the new text with the implementation:

- **Timing**: deletion runs in `beforeTransfer()` (after bootstrap, init, and the integrity check), not during `bootstrap()`. Updated the caution box, prose, and sequence diagram accordingly.
- **Scope**: limited the warning to Strapi-instance destinations (local + remote); file export is unaffected. Linked to the existing destination restore docs instead of referencing a non-existent `#deleteAll()` / "truncate all tables".
- **Accuracy**: described scoped deletion via `restore.deleteRecords()` and CLI `--only` / `--exclude` (including default `admin::` exclusions).
- **Regression**: restored the integrity-check and transfer-stages paragraphs that were removed in the original diff.

Should be good to merge from a docs-accuracy standpoint. 🙏

_AI-assisted comment._

innerdvations (approved):

PR #25097, opened 2025-12-30
feat: allow selectionParams in documents.delete

### LINKED ISSUE #22323: documents().delete() does not support populate

- Node.js version: 20
- NPM version: yarn 1.22
- Strapi version: 5.3.0
- Database: sqlite
- Operating system: macos

### Describe the bug

https://docs.strapi.io/dev-docs/api/document-service#delete says you can `populate` on a delete call, and use `fields`. But the result always contains the flat result, ignoring `populate` and `fields`.

that is because the query is used to find all entries to delete: https://github.com/strapi/strapi/blob/f0a0bc26f5ef0693aaea2a616bc6b816cfee54b6/packages/core/core/src/services/document-service/repository.ts#L123
and only the id is passed to the delete function:
https://github.com/strapi/strapi/blob/f0a0bc26f5ef0693aaea2a616bc6b816cfee54b6/packages/core/core/src/services/document-service/repository.ts#L127

the actual db query to delete only contains the id: https://github.com/strapi/strapi/blob/develop/packages/core/core/src/services/document-service/entries.ts#L53

if i pass my populate to this function, to put in the db.query, populating the delete works. It just isn't built. I guess nobody saw the use case. But i would really like to do this.
I have a bunch of related content-types I wish to also delete when something gets deleted. So populating them before calling `next()` and getting the result to delete these relations would make my code way easier. Now I have to do a `find` before the `next()` to populate the relations.

laurenskling:
### What does it do?

Allow using `fields` and `populate` in delete calls. For example, if you add them in a documents.use middleware.

### Why is it needed?

I want to populate the delete so I can do something with the relations _after_ the delete happened.

Right now the `query` is used to find the versions of the document to delete. So if you add a `populate` or `fields`, it's added to that find query, which makes no sense.

### How to test it?

I've used this in the getstarted example app:

```
  register({ strapi }) {
    strapi.documents.use(async ({ uid, action, params }, next) => {
      console.log('uid', uid);
      if (uid === 'api::category.category' && action === 'delete') {
        params.fields = ['id'];
        params.populate.addresses = true;
      }
      const result = await next();
      if (uid === 'api::category.category' && action === 'delete') {
        console.log('result', result.entries);
      }
      return result;
    });
  },
  ```

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/22323


innerdvations:
I added 'don't merge' because I think it's going to conflict with something coming up this week but after that we will take a look and get this in!

laurenskling:
Alright, I'll see if I need to do any conflict resolves, keep me posted

Bassel17 (changes requested):
The goal here makes sense, but this change seems to break the behavior

pickSelectionParams() is fine for controlling the returned deleted entry payload, but it should not replace the existing locale-aware delete lookup. With the current approach, entriesToDelete is selected only by documentId, so localized deletes may be widened to all locales for that document.

I’d keep the current locale/documentId delete query logic as-is, and apply selection params separately to the returned deleted entries. Adding a test for delete({ documentId, locale, populate }) would help lock this down.

Bassel17 (approved):

laurenskling:
@Bassel17 thank you very much for taking over this feature. I've just now got around to test it and it works perfectly for me. Thank you.

PR #25106, opened 2026-01-02
fix(users-permissions): unable to clear refresh token cookie on logout due to mismatched options

kgndnc:
### What does it do?

Aligns the cookie options used when clearing the httpOnly refresh token during logout with those used when setting it during login. The logout function now passes the full set of cookie attributes (`httpOnly`, `secure`, `sameSite`, `path`, `domain`) instead of only `expires`.

**Cookie options used when setting the cookie during login (lines 110-131):**

```javascript
const upSessions = strapi.config.get('plugin::users-permissions.sessions');
const requestHttpOnly = ctx.request.header['x-strapi-refresh-cookie'] === 'httpOnly';
if (upSessions?.httpOnly || requestHttpOnly) {
  const cookieName = upSessions.cookie?.name || 'strapi_up_refresh';
  const isProduction = process.env.NODE_ENV === 'production';
  const isSecure =
    typeof upSessions.cookie?.secure === 'boolean'
      ? upSessions.cookie?.secure
      : isProduction;

  const cookieOptions = {
    httpOnly: true,
    secure: isSecure,
    sameSite: upSessions.cookie?.sameSite ?? 'lax',
    path: upSessions.cookie?.path ?? '/',
    domain: upSessions.cookie?.domain,
    overwrite: true,
  };

  ctx.cookies.set(cookieName, refresh.token, cookieOptions);
  return ctx.send({ jwt: access.token, user: await sanitizeUser(user, ctx) });
}
```

The fix applies the same pattern to the logout function, adding `expires: new Date(0)` to clear the cookie.

### Why is it needed?

When custom `path` or `domain` options are configured for the refresh token cookie, the logout endpoint fails to clear the httpOnly refresh token cookie from the browser.

Browsers require cookie attributes to match exactly when clearing a cookie. Previously, the logout function only set `expires: new Date(0)` without the other configured options, causing the cookie to persist in the browser after logout.

**Note:** This is not a functionality-breaking bug. The `logout` function correctly invalidates the refresh token server-side via `invalidateRefreshToken()`, so the token cannot be reused even if the cookie remains in the browser. However, clearing the cookie is the intended behavior here for a complete logout flow, and this fix ensures it works correctly when custom cookie options are configured.

**Example configuration that triggers the bug:**

```javascript
// config/plugins.js
module.exports = {
  'users-permissions': {
    config: {
      jwtManagement: 'refresh',
      sessions: {
        httpOnly: true,
        cookie: {
          name: 'refresh_token',
          path: '/api/auth/',   // Custom path
          domain: '.example.com', // Custom domain
          sameSite: 'Strict',
          secure: true,
        },
      },
    },
  },
};
```

With this configuration, calling `POST /api/auth/logout` would return `{ ok: true }` and invalidate the token server-side, but the cookie would remain in the browser.

### Why no tests?

There are no existing tests for the `logout` function in `auth.test.js`. Adding test coverage for logout would widen the scope of this PR beyond the bug fix itself. Additionally, I'm not very familiar with the testing tools used in this project.

It's also worth noting that the unit tests in this file can only verify that `ctx.cookies.set()` is called with certain parameters - they cannot confirm whether the browser actually clears the cookie. I validated this fix manually by examining DevTools in both Firefox and Chrome. If there's a way to test actual browser cookie behavior via code (e.g., e2e tests), I'd be happy to learn and add coverage with some guidance.

### How to test it?

1. Configure users-permissions with custom cookie `path` or `domain` as shown above
2. Enable `jwtManagement: 'refresh'` and `httpOnly: true`
3. Login via `POST /api/auth/local` - observe the cookie is set
4. Logout via `POST /api/auth/logout` - verify the cookie is now properly cleared in the browser

### Related issue(s)/PR(s)

I couldn't find an existing issue for this bug. I encountered it during development of our own project while using custom cookie options with httpOnly refresh tokens.


innerdvations (changes requested):
Thanks for this — the direction looks right. Browsers need the clear attributes to match the ones used when the cookie was set, so aligning logout with path/domain/sameSite/secure (etc.) is the correct fix. This also sits cleanly alongside the recent admin cookie name/path work (including the SSO logout path hotfix): those touch `admin.auth.cookie.*` / `jwtToken`, while this is the users-permissions refresh cookie (`sessions.cookie.*`), so no conflict there.

Before merge, could you please:

1. **Add a committed regression test** that asserts logout’s `cookies.set` options include custom `path` / `domain` / `sameSite` / `secure` when those are configured. The existing logout checks only cover `expires` (or the default path), so they still pass on develop and don’t lock in this failure mode.

2. **Clear via `buildRefreshCookieOptions`** (from #26747) instead of re-inlining the options object — e.g. spread the helper and set `expires` to the epoch (and avoid forwarding a live `maxAge` on clear). Login/refresh already use that helper; keeping logout on the same path avoids the next cookie attribute drifting out of sync again.

Happy to take another look once those are in.
_AI-assisted comment._

kgndnc:
Thanks for the review. I’ve made the requested changes.

### 1. Regression test

I expanded the existing logout cookie test rather than adding a separate test case. The test now verifies that custom `path`, `domain`, `sameSite`, and `secure` values are passed to `cookies.set()`.

I also verified that the test fails on `develop` and passes with my changes.

### 2. Clearing via `buildRefreshCookieOptions`

Logout now uses the helper so its cookie options stay aligned with login and refresh:

```js
const { maxAge, ...cookieOptions } = buildRefreshCookieOptions(upSessions, isProduction);

ctx.cookies.set(cookieName, '', { ...cookieOptions, expires: new Date(0) });
```

I destructured `maxAge` out rather than setting it to `0`, based on your comment to “_avoid forwarding a live maxAge on clear._” If you intended `maxAge: 0` instead, I can adjust this.

One other question: I noticed that the directory structure has changed in PR #26105, and the files affected by this PR have moved. Do I need to do anything about it?

innerdvations:
> I destructured `maxAge` out rather than setting it to `0`, based on your comment to “_avoid forwarding a live maxAge on clear._” If you intended `maxAge: 0` instead, I can adjust this.

Nope, that was the right move, thanks!

> One other question: I noticed that the directory structure has changed in PR #26105, and the files affected by this PR have moved. Do I need to do anything about it?

Yes, #26105 moved these files under server/src/, so those will need to be rebased onto the latest `develop`. After the rebase, you can re-apply the same logout helper clear + the expanded unit assertions on the new paths (server/src/controllers/auth.js and server/src/controllers/__tests__/auth-sessions.test.js). If you have any trouble let me know and I can take care of it so we can get this bug fix merged. Thanks!

kgndnc:
Alright, I rebased onto the latest `develop` — the changes now live at the new `server/src/` paths from #26105. Thanks for your guidance.

Re-verified there: the expanded logout assertions still fail against `develop` (which sends only `{ expires }`) and pass with the fix; all 15 tests in `auth-sessions.test.js` pass.

innerdvations (approved):

PR #25123, opened 2026-01-06
Fix drag-and-drop preview in dynamic zones

Rowan-Paul:
### What does it do?

Fixes the drag-and-drop preview behavior in Dynamic Zones and Repeatable Components by ensuring the `isDragging` state remains stable throughout the entire drag operation.

It preserves the `__temp_key__` instead of regenerating it every time the user drags it. Also generates truly unique keys when adding a new item by checking existing keys.

### Why is it needed?

When dragging components in a Dynamic Zone or Repeatable field, the drag preview would disappear prematurely as soon as the component order changed. This happened because:

1. The `useDragAndDrop` hook has a custom `isDragging` check that compares `item.id` with `monitor.getItem().id`
2. Previously, no `id` was being passed to the drag item, so react-dnd's default tracking was used
3. The default tracking failed when items reordered during drag because components would re-render at different positions

Additionally, the previous `MOVE_FIELD_ROW` implementation regenerated `__temp_key__` on every move using fractional indexing. This caused the dragged item's identity to change mid-drag, breaking the `isDragging` comparison even if an `id` was passed.

The fix ensures:
- `__temp_key__` remains stable during moves (for drag identity tracking)
- New keys are guaranteed unique even when existing keys are out of order (preventing React duplicate key warnings)

### How to test it?

1. Create a content type with a Dynamic Zone field containing multiple component types
2. Add components to the Dynamic Zone
3. Drag a component to reorder it
4. **Expected:** The drag preview (placeholder) should remain visible throughout the entire drag operation, even as the component moves past other items
5. **Previous behavior:** The preview would disappear as soon as the dragged item crossed another component

Also test:
- Adding new components after reordering (should not cause duplicate key warnings)
- Dragging in Repeatable Component fields (same fix applies)

#### Before:

https://github.com/user-attachments/assets/d54ced23-0db4-47c7-b33c-892ecd51f200

#### After:

https://github.com/user-attachments/assets/1472c071-cfa2-49b3-ab73-49fb03d65f1f


packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -594,11 +594,28 @@ const reducer = <TFormValues extends FormValues = FormValues>(
...
+        do {
+          // Use null as lower bound to generate keys at the "end" of the keyspace
+          // This reduces collision likelihood with existing unsorted keys
+          [key] = generateNKeysBetween(
+            existingKeys.size > 0 ? Array.from(existingKeys).sort().pop() : null,
+            null,
+            1
+          );
+          // If collision, add the key to existing set to get next one
+          if (existingKeys.has(key)) {
```
mathildeleg, line 630: But if the key is already used, it is already in the set. Adding it again does nothing. So nothing changes before the next loop iteration. You call `generateNKeysBetween` again with the same inputs, so you often get the same output, until you hit the attempt limit.
Could you fix that?

mathildeleg (commented):
Looks good to me, except one small comment, and I think you'll also need to update one unit test that's failing :)
Thank you 🙏🏻

Rowan-Paul:
Thanks for the review, I've fixed the test & loop. Could you review my changes?

mathildeleg (approved):
Thanks for the fix ✨

PR #25125, opened 2026-01-06
fix: await the async map inside v5 release actions migration

garrettheaver:
The migration which updates the release actions from the v4 database structure to the v5 structure fails with the following error when there are multiple entries in the `strapi_release_actions` table.

```
Error: Transaction query already complete, run with DEBUG=knex:tx for more info
```

The stack dump ultimately points to the `async` at the following file `@strapi/content-releases/dist/server/migrations/database/5.0.0-document-id-in-actions.js:38:21`

On inspection, the problem is due to the `async.map` not being correctly `await`-ed.

Adding the `await` resolves the issue and the migration completes successfully.


derrickmehaffy (approved):
LGTM

derrickmehaffy:
Related to internal TID 5775

PR #25132, opened 2026-01-07
fix(content-type-builder): make write APIs accessible in devMode only

nclsndr:
### What does it do?

Restrict the Content-type builder API to development mode only.

**Technical changes:**
- Created a new `isDevelopmentMode` middleware that checks `strapi.config.get('autoReload')` and throws a `ForbiddenError` when `autoReload !== true`
- Applied this middleware to all 9 Content-Type Builder mutation routes:
  - Content-Types: POST, PUT, DELETE operations
  - Components: POST, PUT, DELETE operations  
  - Component Categories: PUT, DELETE operations
  - Schema: POST `/update-schema` (batch API)
- Added unit tests 
- GET/read-only routes remain unprotected and accessible in production

### Why is it needed?

Content-Type Builder schema modifications should only happen during development (`strapi develop`). Production environments run with `autoReload: false` and should have immutable schemas deployed via source control. This approach:
- Eliminates the SQL injection attack surface in production
- Follows security best practices (immutable production schemas)
- Maintains full development workflow functionality
- Provides clear error messages when blocked

### How to test it?

**1. Unit Tests:**
cd packages/core/content-type-builder
yarn test:unit is-development-modeExpected: All 4 tests pass ✅

**2. Development Mode (should allow mutations):**
Content-Type Builder UI should work normally. Can create/edit/delete content types and components

**3. Production Mode (should block mutations):**

`NODE_ENV=production yarn start`

Attempt to create a content type via API:
```
curl -X POST http://localhost:1337/admin/content-type-builder/content-types \
  -H "Authorization: Bearer YOUR_ADMIN_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"contentType": {...}}'
```

Expected response:

```json
{
  "error": {
    "status": 403,
    "name": "PolicyError",
    "message": "Content-Type Builder modifications are disabled in production mode. Schema changes can only be made when running with autoReload enabled (strapi develop)."
  }
}
```

**4. Verify read operations still work in production:**
curl http://localhost:1337/admin/content-type-builder/content-types \
  -H "Authorization: Bearer YOUR_ADMIN_TOKEN"Expected: Returns list of content types ✅


packages/core/content-type-builder/server/src/middlewares/is-development-mode.ts
```diff
@@ -0,0 +1,20 @@
+import type { Context, Next } from 'koa';
+import { errors } from '@strapi/utils';
+
+/**
+ * Middleware to ensure Content-Type Builder modifications only happen in development mode
+ * This prevents SQL injection vulnerabilities in production by blocking schema modifications
+ * when autoReload is disabled.
+ */
+export default async (ctx: Context, next: Next) => {
+  const autoReload = strapi.config.get('autoReload');
```
alexandrebodin, line 10: it's fine to move forward however I'm wondering if the source of trught should be the env instead of this

nclsndr: The problem with env is that it defaults to "development" without further warning. It's totally possible to deploy strapi, run `strapi start` and still not have `NODE_ENV=production` set. Even if I consider not setting NODE_ENV a bad practice, we can't rely on users mastering this very point.
We could go like: `NODE_ENV !== 'development'` but then how to be consistent with `NODE_ENV=test' and so on, which are also a bad practice but too common to be ignored.
I believe the answer is a correlation of both the `autoReload` (or any other flag, explicitly set by `strapi develop`) and the test/report of the env value to warn the user in case of potential miss use.

PR #25133, opened 2026-01-07
fix(content-type-builder): make write APIs accessible in devMode only

nclsndr:
### What does it do?

Restrict the Content-type builder API to development mode only.

**Technical changes:**
- Created a new `isDevelopmentMode` middleware that checks `strapi.config.get('autoReload')` and throws a `ForbiddenError` when `autoReload !== true`
- Applied this middleware to all 9 Content-Type Builder mutation routes:
  - Content-Types: POST, PUT, DELETE operations
  - Components: POST, PUT, DELETE operations  
  - Component Categories: PUT, DELETE operations
  - Schema: POST `/update-schema` (batch API)
- Added unit tests 
- GET/read-only routes remain unprotected and accessible in production

### Why is it needed?

Content-Type Builder schema modifications should only happen during development (`strapi develop`). Production environments run with `autoReload: false` and should have immutable schemas deployed via source control. This approach:
- Eliminates the SQL injection attack surface in production
- Follows security best practices (immutable production schemas)
- Maintains full development workflow functionality
- Provides clear error messages when blocked

### How to test it?

**1. Unit Tests:**
cd packages/core/content-type-builder
yarn test:unit is-development-modeExpected: All 4 tests pass ✅

**2. Development Mode (should allow mutations):**
Content-Type Builder UI should work normally. Can create/edit/delete content types and components

**3. Production Mode (should block mutations):**

`NODE_ENV=production yarn start`

Attempt to create a content type via API:
```
curl -X POST http://localhost:1337/admin/content-type-builder/content-types \
  -H "Authorization: Bearer YOUR_ADMIN_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"contentType": {...}}'
```

Expected response:

```json
{
  "error": {
    "status": 403,
    "name": "PolicyError",
    "message": "Content-Type Builder modifications are disabled in production mode. Schema changes can only be made when running with autoReload enabled (strapi develop)."
  }
}
```

**4. Verify read operations still work in production:**
curl http://localhost:1337/admin/content-type-builder/content-types \
  -H "Authorization: Bearer YOUR_ADMIN_TOKEN"Expected: Returns list of content types ✅

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -315,7 +315,7 @@ jobs:
         shard: [1/5, 2/5, 3/5, 4/5, 5/5]
     services:
       mysql:
-        image: bitnami/mysql:latest
+        image: bitnamilegacy/mysql:latest
         env:
           MYSQL_ROOT_PASSWORD: strapi
           MYSQL_USER: strapi
@@ -327,7 +327,6 @@ jobs:
           --health-timeout=5s
           --health-retries=3
         ports:
-          # Maps tcp port 5432 on service container to the host
           - 3306:3306
     steps:
       - uses: actions/checkout@v4
@@ -355,7 +354,7 @@ jobs:
         shard: [1/5, 2/5, 3/5, 4/5, 5/5]
     services:
       mysql:
-        image: bitnami/mysql:5.7
+        image: bitnamilegacy/mysql:5.7
         env:
           MYSQL_ROOT_PASSWORD: strapi
           MYSQL_USER: strapi
@@ -367,7 +366,6 @@ jobs:
           --health-timeout=5s
           --health-retries=3
         ports:
-          # Maps tcp port 5432 on service container to the host
           - 3306:3306
     steps:
       - uses: actions/checkout@v4
@@ -469,7 +467,7 @@ jobs:
         shard: [1/5, 2/5, 3/5, 4/5, 5/5]
     services:
       mysql:
-        image: bitnami/mysql:latest
+        image: bitnamilegacy/mysql:latest
         env:
           MYSQL_ROOT_PASSWORD: strapi
           MYSQL_USER: strapi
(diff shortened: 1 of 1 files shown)


examples/getstarted/package.json
```diff
@@ -14,16 +14,16 @@
...
-    "@strapi/plugin-documentation": "4.25.23",
-    "@strapi/plugin-graphql": "4.25.23",
-    "@strapi/plugin-i18n": "4.25.23",
-    "@strapi/plugin-sentry": "4.25.23",
-    "@strapi/plugin-users-permissions": "4.25.23",
-    "@strapi/provider-email-mailgun": "4.25.23",
-    "@strapi/provider-upload-aws-s3": "4.25.23",
-    "@strapi/provider-upload-cloudinary": "4.25.23",
-    "@strapi/strapi": "4.25.23",
+    "@strapi/plugin-color-picker": "workspace:*",
```
alexandrebodin: with the old way we released in v4 I think you need to keep the current version and it will be bumped on release ?

markkaylor: We had an issue though where examples/getstarted was not using the built code and instead seemed to be reading from the latest published version. It took us a while to figure out why our changes were not being applied.
 
 But I guess if the script is going to try and bump this that will indeed be an issue

markkaylor: The last v4 release we did was 4.26.0 so it doesn't look like the release bumps this and this package.json is stale

nclsndr: Yes it was stale, do I understand right that it should be automatically bump by the release process?
Nevertheless, how should we test the local version when working on v4? Should we do this change and not commit it?

markkaylor: > We had an issue though where examples/getstarted was not using the built code and instead seemed to be reading from the latest published version. It took us a while to figure out why our changes were not being applied.
> 
> But I guess if the script is going to try and bump this that will indeed be an issue

Bumping all the versions to 4.26.0 (latest v4) seems to resolve the issue we had but I do not understand it. I'll push to fix the versions.

PR #25135, opened 2026-01-07
docs: add blocks editor contributor docs

remidej:
### What does it do?

Documents the Blocks editor. As part of our focus on quality, we'll work on improving it quite a bit in 2026. This explains how it works, why it was built like this, and anti-patterns to avoid.

### How to test it?

https://contributor-docs-git-docs-blocks-strapijs.vercel.app/docs/core/content-manager/blocks

### DIFF
--- a/docs/docs/docs/01-core/content-manager/05-blocks.md
+++ b/docs/docs/docs/01-core/content-manager/05-blocks.md
@@ -29,94 +29,94 @@ With these elements combined, here's what the content for a sample Blocks attrib
 ```json
 [
   {
-    "type":"heading",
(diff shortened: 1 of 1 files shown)


docs/static/img/content-manager/blocks/components-structure.tldr
remidej, line 1: I used TLDraw for the diagram above and this is the exported source file. I put it here so we can edit the diagram in the future if needed. If you don't think it's useful I can remove it

markkaylor (commented):
Looks good to me! Thank you ❤️

PR #25142, opened 2026-01-08
v4.26.1

Bassel17:
### What does it do?

v4 release


markkaylor (commented):
I don't think the examples apps versions were updated to 4.26.1 which causes them to fetch the published version instead of linking to the workspace. Do you mind updating those 🙏

Bassel17:
> I don't think the examples apps versions were updated to 4.26.1 which causes them to fetch the published version instead of linking to the workspace. Do you mind updating those 🙏

I wonder why that stopped working, it should be automatic

PR #25149, opened 2026-01-08
fix: change commit-lint to make body length a warning

derrickmehaffy:
### What does it do?

Changes the commit lint body max length rule to be a warning not an error and increase the maximum from 100 to 150

### Why is it needed?

Annoying when you need to put links or references in your commit messages

### How to test it?

Make a commit with a body longer than 100

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/.commitlintrc.ts
+++ b/.commitlintrc.ts
@@ -21,6 +21,11 @@ const config: UserConfig = {
         'future',
       ],
     ],
+    'body-max-line-length': [
+      RuleConfigSeverity.Warning,
+      'always',
+      150,
+    ],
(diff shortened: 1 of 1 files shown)


Adzouz:
Hey @derrickmehaffy, even if see your point of view, I don't think we should put links in commits messages and I'm not confident having just a warning because it could lead to pushing some very long commit messages. I'd rather keep it an error and mention related issues / PR IDs instead of links. Decided with the rest of the team to close it. Feel free to challenge it and reopen if you really think that's necessary 😊

PR #25151, opened 2026-01-09
test: migration testing project [CMS-81]

Bassel17:
What does it do?
Adds  the examples/complex migration test app and tooling to validate v4 → v5 migrations. Updates DB helpers (Postgres/MySQL only)

Why is it needed?
To provide a reliable, repeatable way to exercise and validate migration behavior on complex schemas

How to test it?
Environment: Node 20 (required for v4), Docker (Postgres/MySQL).

```cd examples/complex
yarn setup:v4  # use the printed path
yarn db:wipe:postgres  # only if you've run the DB before
```

In the v4 project:
```
yarn develop:postgres
yarn seed
```
Back in v5:
```
yarn develop:postgres  # migrations run
yarn test:migration
```


examples/complex/scripts/seed-v4.js
```diff
@@ -0,0 +1,767 @@
...
+  simpleInfo: () => ({
+    title: `Info ${random.string()}`,
+    description: `Description ${random.string(15)}`,
+    count: random.number(1, 100),
+    active: random.boolean(),
+  }),
+
+  imageBlock: () => ({
+    alt: `Image ${random.string()}`,
+    url: `https://example.com/images/${random.string()}.jpg`,
```
nclsndr, line 79: Shouldn't we use actual placeholder service to have proper assets?

examples/complex/docker-compose.dev.yml
```diff
@@ -0,0 +1,30 @@
...
+    environment:
+      MYSQL_DATABASE: strapi
+      MYSQL_USER: strapi
+      MYSQL_PASSWORD: strapi
+      MYSQL_ROOT_HOST: '%'
+      MYSQL_ROOT_PASSWORD: strapi
+    volumes:
+      - mysqldata:/var/lib/mysql
+    ports:
+      - '3306:3306'
```
nclsndr: Same issue with exposing default MySQL port

examples/complex/docker-compose.dev.yml
```diff
@@ -0,0 +1,30 @@
...
+
+volumes:
+  pgdata:
+  mysqldata:
```
nclsndr, line 30: To ease volume management, what about mapping the db volumes to a local `.tmp` dir in this very example dir.

examples/complex/snapshots/.gitkeep
nclsndr, line 1: Why do we want to keep the snapshots dir?

examples/complex/scripts/db-mysql.js
```diff
@@ -0,0 +1,207 @@
...
+
+const SCRIPT_DIR = __dirname;
+const COMPLEX_DIR = path.resolve(SCRIPT_DIR, '..');
+const DOCKER_COMPOSE_FILE = path.join(COMPLEX_DIR, 'docker-compose.dev.yml');
+const COMPOSE_PROJECT_NAME = 'strapi_complex';
+const SNAPSHOTS_DIR = path.join(COMPLEX_DIR, 'snapshots');
+
+const DB_NAME = 'strapi';
+const DB_USER = 'strapi';
+const DB_PASSWORD = 'strapi';
```
nclsndr: Don't we want to rely on env/config to avoid repetition and maintenance hassle?

examples/complex/scripts/db-postgres.js
```diff
@@ -0,0 +1,255 @@
...
+
+const SCRIPT_DIR = __dirname;
+const COMPLEX_DIR = path.resolve(SCRIPT_DIR, '..');
+const DOCKER_COMPOSE_FILE = path.join(COMPLEX_DIR, 'docker-compose.dev.yml');
+const COMPOSE_PROJECT_NAME = 'strapi_complex';
+const SNAPSHOTS_DIR = path.join(COMPLEX_DIR, 'snapshots');
+
+const DB_NAME = 'strapi';
+const DB_USER = 'strapi';
+const DB_PASSWORD = 'strapi';
```
nclsndr: same about hardcoded credentials

examples/complex/scripts/db-mysql.js
```diff
@@ -0,0 +1,207 @@
...
+    const output = execSync(`docker ps --filter "ancestor=mysql:8" --format "{{.Names}}"`, {
+      encoding: 'utf8',
+    }).trim();
+    if (output) {
+      return output.split('\n')[0];
+    }
+  } catch (error) {
+    // Continue to default
+  }
+  return `${COMPOSE_PROJECT_NAME}-mysql-1`;
```
nclsndr: Don't we prefer to throw rather than silently leaving the failure for the next system relying on it?

examples/complex/scripts/db-mysql.js
```diff
@@ -0,0 +1,207 @@
...
+    if (output) {
+      const containerId = output.split('\n')[0];
+      const nameOutput = execSync(`docker inspect --format='{{.Name}}' ${containerId}`, {
+        encoding: 'utf8',
+      }).trim();
+      return nameOutput.replace(/^\//, ''); // Remove leading slash
+    }
+  } catch (error) {
+    // Fallback to expected name pattern
+  }
```
nclsndr: Same here?

examples/complex/scripts/db-mysql.js
nclsndr, line 1: Do we really want all those scripts to be written in JS? 
What about using tools like `tsx`

examples/complex/snapshots/.gitkeep
Bassel17: yeah why not we can restore the db if we wish too

examples/complex/docker-compose.dev.yml
Bassel17: Not sure this would ease volume management since we can already manage named volumes with Docker commands. Using a bind mount could actually add maintenance overhead if we hit issues. I’d prefer to keep named volumes unless we see a concrete need to inspect files directly.

examples/complex/scripts/db-mysql.js
Bassel17: They’re small, one‑off CLI helpers and adding tsx would add runtime/build dependencies and extra friction,
I think it would be better to wait and see if they grow enough to be a shared dependency or something we feel like we constantly need to edit they maybe then we can make them TS.

examples/complex/README.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+This displays a table showing how many records are in each table, useful for quickly seeing if the database is empty, has data, etc.
+
+### Typical Migration Testing Workflow
+
+1. **Setup v4 project** (if not already done):
+
+   ```bash
+   yarn clone:v4
```
innerdvations:
```suggestion
   yarn setup:v4
```

examples/complex/README.md
```diff
@@ -0,0 +1,242 @@
...
+## Migration Testing Workflow
+
+This project includes tools for testing migrations between Strapi v4 and v5 by creating an isolated v4 project and managing database snapshots. The complex example ships its own `docker-compose.dev.yml` so the database containers are independent of the monorepo root.
+
+### Setup
+
+1. **Create/Update the external v4 project:**
+
+   ```bash
+   yarn clone:v4
```
innerdvations:
```suggestion
   yarn setup:v4
```

examples/complex/scripts/setup-v4-project.js
```diff
@@ -0,0 +1,909 @@
...
+const CONTENT_TYPES = [
+  'basic',
+  'basic-dp',
+  'basic-dp-i18n',
+  'relation',
+  'relation-dp',
+  'relation-dp-i18n',
+];
+
+const I18N_CONTENT_TYPES = ['basic-dp-i18n', 'relation-dp-i18n'];
```
innerdvations: This is never used anywhere

examples/complex/scripts/setup-v4-project.js
```diff
@@ -0,0 +1,909 @@
...
+    const schema = JSON.parse(fs.readFileSync(schemaSource, 'utf8'));
+    if (schema.attributes) {
+      // Remove polymorphic relations that cause issues in v4
+      delete schema.attributes.morphToOne;
+      delete schema.attributes.morphOne;
+      delete schema.attributes.morphMany;
+    }
+
+    fs.writeFileSync(schemaDest, JSON.stringify(schema, null, 2) + '\n');
+    // TODO: stop removing polymorphic relations
```
innerdvations, line 359: This will be needed for the migration tests. Should we try to fix it here, or do you think it's better to do a follow-up PR to add it?

examples/complex/scripts/setup-v4-project.js
```diff
@@ -0,0 +1,909 @@
...
+
+// Create scripts directory
+const v4ScriptsDir = path.join(V4_PROJECT_DIR, 'scripts');
+if (!fs.existsSync(v4ScriptsDir)) {
+  fs.mkdirSync(v4ScriptsDir, { recursive: true });
+}
+
+// Create develop-with-db.js script for v4 project
+const dockerComposePath = findDockerComposeFile(V4_PROJECT_DIR);
+const developWithDbScript = `#!/usr/bin/env node
```
innerdvations: Like Nico said, could we try to de-duplicate some code? This is all (necessarily!) a bit complicated, so we should try to simplify the parts we're able to and try to prevent it from growing into a mess over time

examples/complex/scripts/setup-v4-project.js
```diff
@@ -0,0 +1,909 @@
...
+      stdio: 'inherit',
+      env: COMPOSE_ENV,
+    });
+    console.log(\`✅ \${serviceName} container started\`);
+    
+    // Wait a bit for the database to be ready
+    if (dbType === 'postgres' || dbType === 'mysql') {
+      console.log('Waiting for database to be ready...');
+      const start = Date.now();
+      while (Date.now() - start < 3000) {
```
innerdvations: This is ok for now, but if we want to eventually work this into automation we'll need it to be guaranteed correct with a real db ready check in a loop rather than just waiting 3 seconds:
```
        if (dbType === 'postgres') {
          execSync(`docker exec ${containerId} pg_isready -U strapi`, { stdio: 'ignore' });
        } else {
          execSync(`docker exec ${containerId} mysqladmin ping -u strapi -pstrapi --silent`, { stdio: 'ignore' });
        }
```

examples/complex/scripts/setup-v4-project.js
Bassel17: I think, we can follow it up in the discard-drafts PR just to keep things moving

examples/complex/scripts/db-postgres.js
```diff
@@ -0,0 +1,208 @@
...
+      } catch (error) {
+        console.error(`Error wiping database: ${error.message}`);
+        process.exit(1);
+      }
+    }
+    break;
+
+  case 'check':
+    {
+      const containerName = getContainerName();
```
innerdvations, line 159:
```suggestion
      const containerName = resolveContainerName();
```

PR #25176, opened 2026-01-13
chore: refactor blocks editor block api

remidej:
### What does it do?

Some housekeeping on the Blocks editor implementation. As explained in the [contributor docs](https://contributor-docs-git-docs-blocks-strapijs.vercel.app/docs/core/content-manager/blocks), the main design principle of this feature is that that logic specific to a given block should be self-contained, so that the editor itself can remain block-agnostic.

This PR includes some improvements to follow this guideline.
- introduces a `plugin` key so blocks can create their own slate plugins. Before that we were spreading logic that was specific to link and image block types.
- introduces a `isDraggable` key so blocks can specificy whether they should get the drag handle. We had logic specific to links and lists directly in BlocksContent before

most of the diff is from code moving from a `plugins` folder into the relevant block files.
### Why is it needed?

Making Blocks more maintainable, and paving the way for the custom blocks api

### How to test it?

non-regression testing on the blocks editor. in particular, creating links, images, and reordering blocks

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/Code.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/Code.tsx
@@ -178,33 +178,6 @@ const CodeEditor = (props: RenderElementProps) => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


mathildeleg (approved):
Looks good ✅🚀

PR #25167, opened 2026-01-12
fix(upload): content-api does not check file type restriction config

markkaylor:
### What does it do?

Restrict file types on the content-api

### Why is it needed?

We recently added it for the admin-api

### How to test it?

1. Configure the `plugins.js` with something like
```
upload: {
    config: {
      security: {
        allowedTypes: ['image/*', 'application/*'],
        deniedTypes: ['application/x-sh', 'application/x-dosexec']
      }
    },
  }
```
2. Start the server
3. Go to settings -> users permissions -> authenticated -> media lib. Check upload and save.
4. Create a user with
```
curl -X POST "http://localhost:1337/api/auth/local/register" \
-H "Content-Type: application/json" \
-d '{"username":"testuser","email":"test@example.com","password":"Password123"}'
```
5. Copy the JWT and make a request with a file in the deniedTypes list. It should be rejected.

PR #25177, opened 2026-01-13
security(upload): improve mimetype detection for uploads

innerdvations:
# Upload MIME type validation (algorithm + persistence)

## What does it do?

This PR implements the Upload plugin’s MIME validation as documented in [MIME type validation](/docs/docs/01-core/upload/01-backend/01-mime-validation.md).

1. **New validation algorithm** – Validation follows a fixed sequence of steps. All accept/deny decisions go through a single helper `validateAllowBanLists(mimetype)`. In order:
   - Reject if **declared** type is denied; reject if **extension’s** type is denied.
   - Run content detection (`file-type`).
   - **Trusted declaration:** if declared matches extension and detection confirms, accept via declared type (e.g. avoids rejecting PDF/HTML polyglots when we allow PDF).
   - Reject if **detected** type is denied.
   - If an allow list is set, reject when **detected** is not in it (extension/declared cannot override), with an **exception** for Office formats: when detection returns `application/zip` and the extension maps to an allowed type (e.g. docx, xlsx), the implementation may use the extension type so Office uploads work.
   - Otherwise use **detected** when it exists and matches extension or declared is generic; then **extension** when present; then **declared** as last resort; else reject (“cannot verify file type”).

2. **Persist the chosen MIME type** – The type returned by validation (declared, detected, or extension) is now saved to file metadata. Previously the detected type was often discarded and files were stored as `application/octet-stream`.

3. **Empty allow list = reject all** – `allowedTypes: []` is now “reject all uploads”. Unset/undefined still means no allow-list restriction.

4. **Config** – Security config is read only from `plugin::upload.security` (no fallback to `plugins.upload.security`).

## Why is it needed?

- **Correct, documented behavior:** The previous logic did not match the intended design (e.g. order of deny checks, trusted declaration, allow-list enforcement against detected type, Office/zip exception). The new implementation and contributor doc are aligned.
- **Stored MIME:** Using the validated type for storage fixes wrong `mime` in the DB, wrong Content-Type when serving, and misleading metadata.
- **Security:** Treating `allowedTypes: []` as “reject all” and enforcing the allow list against detected type (so e.g. a JPEG sent as `.pdf` is rejected when only PDF is allowed) hardens upload security.

## Before vs after

| Aspect | Before | After |
|--------|--------|--------|
| Validation logic | Ad-hoc allow/deny checks | Documented 9-step flow, single gate |
| Deny list order | Mixed | Declared → extension → then detected |
| Trusted declaration | Not applied | Declared + extension + detection must agree |
| Allow list vs detected | Could be overridden by extension/declared | Detected must be in allow list (with Office/zip exception) |
| Stored MIME | Often `application/octet-stream` | Chosen type (detected/extension/declared) |
| `allowedTypes: []` | Allow all | Reject all |
| Config key | `plugin::upload` + fallback to `plugins.upload` | `plugin::upload.security` only |

## How to test

- **Unit:** `packages/core/upload/server/src/utils/__tests__/mime-validation-matrix.test.ts` and `mime-validation.test.ts`.
- **API:** `tests/api/core/upload/content-api/upload.test.api.js` (matrix, detection, empty allow list).
- **Manual – detected MIME stored:** Set `plugin::upload.security.allowedTypes` to include docx (or `['application/*']`). Upload a real `.docx` without Content-Type. Response/DB `mime` should be the detected docx type, not `application/octet-stream`.
- **Manual – empty allow list:** Set `plugin::upload.security` to `{ allowedTypes: [] }`. Any upload should return 400 (not allowed).

## Breaking changes

- **Validation algorithm:** Behavior changes for edge cases (order of checks, trusted declaration, allow list vs detected, Office/zip). Aligns with the documented design.
- **`allowedTypes: []`:** Now “reject all”. Anyone who relied on “allow all” with `[]` would be affected; that combination is unsafe and we treat the change as a security fix.

## Related

Fixes CMS-126

### DIFF
--- a/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
+++ b/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
@@ -544,6 +544,27 @@ describe('mime-validation', () => {
       expect(result.errors).toHaveLength(0);
     });
 
+    it('when no config and detection fails (undefined), uses extension so stored mime is not octet-stream', async () => {
+      mockStrapi.config.get.mockReturnValue({});
+      mockReadFile.mockResolvedValue(Buffer.from('content'));
+      mockFileTypeFromBuffer.mockResolvedValue(undefined);
+
+      const fileWithExtOnly = {
+        name: 'document.docx',
+        path: '/tmp/document.docx',
+        size: 100,
+        mimetype: 'application/octet-stream',
+      };
+
+      const result = await enforceUploadSecurity([fileWithExtOnly], mockStrapi);
+
+      expect(result.validFiles).toHaveLength(1);
+      expect(result.validFiles[0].detectedMimeType).toBe(
+        'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
+      );
+      expect(result.errors).toHaveLength(0);
+    });
+
     it('should return errors for disallowed MIME type', async () => {
       mockStrapi.config.get.mockReturnValue({
         allowedTypes: ['application/pdf'],
--- a/packages/core/upload/server/src/utils/mime-validation.ts
+++ b/packages/core/upload/server/src/utils/mime-validation.ts
@@ -200,37 +200,6 @@ export async function validateFile(
 ): Promise<ValidationResult> {
   const { allowedTypes, deniedTypes } = config;
 
-  // When no allow/deny config, still run detection so stored file gets detected MIME when declared is generic (e.g. octet-stream)
-  if (!allowedTypes && !deniedTypes) {
-    let detectedMime: string | undefined;
-    try {
-      detectedMime = await detectMimeType(file);
-    } catch (error) {
-      const errMsg = error instanceof Error ? error.message : String(error);
-      strapi.log.warn(`Failed to detect MIME type from file: ${errMsg}`, {
-        fileName: extractFileInfo(file).fileName,
-        error: errMsg,
-      });
-    }
-    // When declared is generic (e.g. octet-stream): use detection first, then fallback to extension so we don't store octet-stream when we can determine a type.
-    const { fileName } = extractFileInfo(file);
-    const fileExt = extname(fileName).toLowerCase();
-    const expectedMimeFromExt = fileExt ? lookup(fileExt) || null : null;
-    // file-type returns application/zip for Office formats (docx, xlsx); use extension-based MIME for stored type
-    if (
-      detectedMime === 'application/zip' &&
-      expectedMimeFromExt &&
-      expectedMimeFromExt !== 'application/zip'
-    ) {
-      detectedMime = expectedMimeFromExt;
-    }
-    // When detection failed or returned nothing, use extension so we still store a meaningful type instead of octet-stream
-    if (!detectedMime && expectedMimeFromExt) {
-      detectedMime = expectedMimeFromExt;
-    }
-    return { isValid: true, detectedMime };
-  }
-
   const { fileName, declaredMimeType } = extractFileInfo(file);
 
   const fileExt = extname(fileName).toLowerCase();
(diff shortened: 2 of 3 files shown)


markkaylor (commented):
Thanks for the PR. I tested it but unfortunately the file is still being rejected:

Image

I also tested the commit before https://github.com/strapi/strapi/pull/25167 was merged and I also have the error so I do not believe there was a regression there.

innerdvations:
Yeah, I don't think this was introduced by your recent PR. I'll take a look later and see what is still going wrong though

innerdvations:
Oops, sorry about that!

I fixed the test and added a contributor doc describing how it works now. Basically, we trust detected mimetype above all, if that fails, we kinda-sorta-trust the declared mimetype as long as it's allowed and not deceptive

markkaylor:
Hey @innerdvations sorry I'm still getting this error 😢 

Screenshot 2026-02-11 at 16 11 08



I followed the steps here on this pr. Added the security config and created a test docx file then ran:

```
curl -F "files=@test.docx" http://localhost:1337/api/upload
```

innerdvations:
> Hey @innerdvations sorry I'm still getting this error 😢
> 
> Screenshot 2026-02-11 at 16 11 08
> I followed the steps here on this pr. Added the security config and created a test docx file then ran:
> 
> ```
> curl -F "files=@test.docx" http://localhost:1337/api/upload
> ```

Was the docx file a valid docx or random data with the extension `.docx`? I ask because we should be able to detect a valid docx.

If it was a valid file, I'll look into why it wasn't detected, but it would probably accept it if you upload it with `Content-Type: application/vnd.openxmlformats-officedocument.wordprocessingml.document` because then the mimetype would match the declared type

innerdvations:
I added tests that very explicitly explain and test each permutation of:

- allowedList undefined / set
- detected file type / undetectable
- does extension exist / match / not match
- does declared match / not match

and the expected behaviours. Take a look at those and if you think any of the expected cases should be different let's discuss, because I'm not 100% sure in every case anyway

innerdvations:
Could you take a look at the new tests and see if all the test cases make sense for what we want? I'm a little worried about either going too strict and breaking a lot of projects, or going too loose and allowing dangerous files in.

packages/core/upload/server/src/utils/__tests__/mime-validation-matrix.test.ts
```diff
@@ -0,0 +1,483 @@
...
+      description: 'detectable, detected in allow, extension matches → allow',
+      allowedList: ['image/jpeg'],
+      bannedList: [],
+      contentKind: 'jpg',
+      uploadedFileName: 'a.jpg',
+      uploadedFileDeclaredType: 'image/jpeg',
+      expectedResult: 'allow',
+    },
+    {
+      description: 'detectable, detected in allow, extension does not match (warn but allow)',
```
Bassel17: am a little skeptical about this behavior ? do we want to allow things if extension don't match the type
could open door ways to people downloading files from Strapi without actual knowing the content, it's misleading we might want to actually reject

innerdvations: True, this could be a problem if a valid JPG is also a valid EXE and .exe isn't explicitly banned. I will update to enforce matching extensions

innerdvations:
@markkaylor that should be fixed now

markkaylor:
Sorry @innerdvations I still have the same result. Without any config I have mime as "application/octect-stream"

```
[{"id":67,"documentId":"l19xqplrbek2khsvljirhvkf","name":"test.docx","alternativeText":null,"caption":null,"width":null,"height":null,"formats":null,"hash":"test_a9ef27af80","ext":".docx","mime":"application/octet-stream","size":111.3,"url":"/uploads/test_a9ef27af80.docx","previewUrl":null,"provider":"local","provider_metadata":null,"createdAt":"2026-02-24T13:14:03.232Z","updatedAt":"2026-02-24T13:14:03.232Z","publishedAt":"2026-02-24T13:14:03.232Z","focalPoint":null}]
```

I definitely have your last commit locally and I've run a setup

markkaylor:
Sorry didn't see you pushed I'll review today

PR #25186, opened 2026-01-14
ci: add sonarqube to PRs

cache-your-dreams:
### What does it do?
Strapi/strapi now includes SonarQube test coverage reporting that integrates with your existing Jest testing infrastructure. The setup supports both unit tests and frontend tests in your monorepo structure.
Initially, the workflow is not blocking on Failure.

### Why is it needed?

So that we can track our test coverage to improve on it.

### How to test it?

This is a workflow that is run when pushing on develop. For testing purposes, I have added the feature branch as well, but this should be removed before merging


Baldinof:
This will re-run all the tests that also run in the [tests.yml](https://github.com/strapi/strapi/blob/develop/.github/workflows/tests.yml) workflow.

Maybe the best would be to gather all coverage reports in tests.yml in a final job and submit them to sonar there?

.github/workflows/tests.yml
```diff
@@ -197,7 +207,16 @@ jobs:
...
-        run: yarn nx affected --target=test:front --nx-ignore-cycles -- --runInBand
+        run: yarn nx affected --target=test:front --nx-ignore-cycles -- --runInBand --coverage
+
+      - name: Upload coverage artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: front-coverage-reports
+          path: |
+            coverage/
+            packages/*/coverage/
```
Baldinof: Maybe it should be like this

```suggestion
            packages/**/coverage/
```

nclsndr:
Interesting addition !

Do we have an idea of the time it will add to the CI @cache-your-dreams?

cache-your-dreams:
> Interesting addition !
> 
> Do we have an idea of the time it will add to the CI @cache-your-dreams?

No I don't know @nclsndr 
Adding the `--coverage` tag adds less than a second on my local and then it's only sending the data to sonarcloud, so I don't think it will anything noticeable.

PR #25191, opened 2026-01-14
enhancement/responsiveness wysiwyg

mathildeleg:
### What does it do?
Ensures the Rich text blocks and Markdown blocks are responsive.

_Markdown blocks:_
* Paddings should be different
* Expand mode not available 
* Preview toggle button at the bottom of the block

**BEFORE**
Capture d’écran 2026-01-14 à 14 17 58

**AFTER**
https://github.com/user-attachments/assets/c7e296f6-73ab-49f0-871e-462589ecb1e8

_Rich text blocks_
* Paddings should be different
* Expand mode not available
* Drag & drop replaced by arrows available via the toolbar

**BEFORE**
Capture d’écran 2026-01-14 à 14 18 11

**AFTER**
https://github.com/user-attachments/assets/23b927bc-5bf7-4172-90e4-c0d1a64b2fb6

### Why is it needed?
Writing in those blocks could be a bit tricky on mobile.

### How to test it?
* Use your mobile (for better testing xp)
* Go to the Article content-type 
* Write and format text in the markdownContent block.
* Write and format text in the blocksContent block, re-order the lines with the arrows now present in the toolbar.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/PreviewWysiwyg.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/PreviewWysiwyg.tsx
@@ -9,10 +9,10 @@ interface PreviewWysiwygProps {
   data?: string;
 }
 
-const PreviewWysiwyg = ({ data = '' }: PreviewWysiwygProps) => {
+const PreviewWysiwyg = ({ data }: PreviewWysiwygProps) => {
   const html = React.useMemo(
     () =>
-      sanitizeHtml(md.render(data.replaceAll('\\n', '\n') || ''), {
+      sanitizeHtml(md.render((data ?? '').replaceAll('\\n', '\n')), {
         ...sanitizeHtml.defaults,
         allowedTags: false,
         allowedAttributes: {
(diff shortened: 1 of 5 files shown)


mathildeleg:
> Nice work. The buttons to move blocks works really nicely. The padding/margins on the blocks editor for me look a bit different than your screenshot. It seems like there is more padding on the left than on the right.
> 
> Image
> Otherwise I don't have much to say about the code, looks good to me 👏

I think I may have forgotten to update the screenshots after I tweaked the buttons a bit to make sure they were always OK to click on on mobile! Sorry
IMG_3901

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/PreviewWysiwyg.tsx
```diff
@@ -9,10 +9,10 @@ interface PreviewWysiwygProps {
   data?: string;
 }
 
-const PreviewWysiwyg = ({ data = '' }: PreviewWysiwygProps) => {
+const PreviewWysiwyg = ({ data }: PreviewWysiwygProps) => {
   const html = React.useMemo(
     () =>
-      sanitizeHtml(md.render(data.replaceAll('\\n', '\n') || ''), {
+      sanitizeHtml(md.render((data ?? '').replaceAll('\\n', '\n')), {
```
mathildeleg, line 15: Otherwise got an error when toggling "Preview button" when the Markdown block was empty.

PR #25197, opened 2026-01-15
fix: reorder components and dz on mobile

Adzouz:
### What does it do?

Improve reordering of components and dynamic zones on mobile with a different approach than the drag and drop.

https://github.com/user-attachments/assets/991613d0-5c15-468e-aaf4-1d7f34c3256a

### Why is it needed?

Help improving the experience on mobile and bring back the reordering feature of components in the content manager.

### How to test it?

- Go to the admin > Content Manager
- Select a content type that has repeatable components / dynamic zones
- Add multiple elements to a components field / a dynamic zone
- Fill some information to be able to differentiate them
-> Play with the arrows to switch the position of the components

🚀


remidej (approved):
Tested ✅ should be easier to use on mobile indeed 👍

PR #25216, opened 2026-01-19
feat: add preview support to images and videos

remidej:
### What does it do?

Previously images and video assets were not detected in the live preview. Now:
- hovering them shows the highlight overlay
- double clicking this highlight opens the media input popover
- when changing the asset in this popover, the preview gets updated without needing to save

Some known issues:
- playing a video is currently broken. We redispatch the clic, to the element under the highlight (something we implemented for links and buttons), but clicking the video tag itself does nothing, as it's not the actual play/pause button. Because we can't make any assumptions on what the video player looks like, and because play/pause isn't the only interaction on it, I think redispatching the click is a dead end. Instead we could consider to remove the overlay altogether when holding shift or something like that.
- After changing the asset in the media modal, it would be nice to auto-close the input popover

### How to test it?

Open the preview for an entry with a image/video field (mutliple or not). Have a growth license

Fixes CMS-141


examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -73,6 +75,53 @@ const NestedValue = ({ value, level = 0, arrayIndex = undefined, fieldName = und
...
+      <Flex direction="column" alignItems="flex-start" gap={2}>
+        <Box
+          tag="video"
+          src={value.url}
+          controls
+          maxWidth="100%"
+          maxHeight="300px"
+          borderRadius="4px"
+        >
+          Your browser does not support the video tag.
```
Adzouz, line 120: Should we localize this string?

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
```diff
@@ -154,7 +154,7 @@ const InputPopover = ({ documentResponse }: { documentResponse: ReturnType<UseDo
               transform={`translate(${iframeRect.left + popoverField.position.left}px, ${iframeRect.top + popoverField.position.top}px)`}
             />
           </Popover.Trigger>
-          <Popover.Content sideOffset={4}>
+          <Popover.Content sideOffset={4} style={{ zIndex: 5 }}>
```
Adzouz, line 157: Not related to this one but it got me thinking about the highlight overlay around a value and it's not the best display 😢
For this one, I think it's ok with this value because it's okay to have it overlapping the header but I would maybe try to fix the overlay (if possible, because I think you're using a portal)

Image

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -72,6 +72,37 @@ const previewScript = (config: PreviewScriptConfig) => {
...
+      return sourceAttr;
+    }
+
+    // For non-media elements, check if it's a media asset field
+    const params = new URLSearchParams(sourceAttr);
+    if (params.get('model') === 'plugin::upload.file') {
+      const elementPath = params.get('path');
+      if (elementPath) {
+        // Strip the last segment (e.g., "hero.caption" -> "hero")
+        const parentPath = elementPath.split('.').slice(0, -1).join('.');
```
Adzouz, line 218: Just curious here, what about if the file field is at root level? (like just `caption`) is it okay to have a empty parent path?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -540,10 +630,46 @@ const previewScript = (config: PreviewScriptConfig) => {
 
         getElementsByPath(field).forEach((element) => {
           if (element instanceof HTMLElement) {
-            element.textContent = value || '';
+            // For img/video elements, update the src attribute instead of text content
+            if (isMediaElement(element)) {
+              // Value can be a media object with url property, or a string URL
+              const url = typeof value === 'object' && value !== null ? value.url : value;
+              element.setAttribute('src', url || '');
```
Adzouz: Yeah is there a way to sanitize the value and be sure we won't have a security issue with injection here?

Adzouz (changes requested):
Good job it's clean 👏
Just a few comments and also I had two strange things to report:
- If you delete the media, the popover will stay at the same spot but since the image doesn't exist anymore, it's not aligned correctly with the field (also we have a broken image instead of having null displayed)

https://github.com/user-attachments/assets/9afe4f31-db40-4da0-a5fb-a42d0d5ba8a8

- On nested level, the media doesn't show up in the tree, is that normal? 
Image

Let me know if I can help!

examples/getstarted/src/admin/preview/dummy-preview.jsx
remidej: This is the dummy preview (our internal devtool), not user facing, I don't think it's needed

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
remidej: The overlay is not within the iframe, it's above it. So it's not possible to display it in between 'layers' of the iframe content, it will always be fully above the iframe

If we wanted to address it, we could send more data via postMessage to let the overlay know what part of the content is actually visible, and use that as a mask for the overlay.

I personally don't think it's worth the trouble

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: I don't think the security issue is relevant to us. It gives Strapi admins of a site the ability to inject JS in the frontend of their own website. More importantly, it only happens within their browser. So it's the same as messing around in the chrome devtools

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
remidej: I think the only cases where that can happen is if users query the upload plugins directly (`/api/upload/files/:id` and `/api/upload/files`) instead of querying their content types. It's true that this edge case isn't handled, but I think the use cases for these endpoints are very rare.

packages/core/content-manager/admin/src/preview/components/InputPopover.tsx
Adzouz: Yeah not worth it no worries!

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
Adzouz: You're right, I wasn't sure but all good then thanks!

markkaylor:
@Adzouz 

> If you delete the media, the popover will stay at the same spot but since the image doesn't exist anymore, it's not aligned correctly with the field (also we have a broken image instead of having null displayed)

I've pushed to fix this

> On nested level, the media doesn't show up in the tree, is that normal?

For this I've taken a look but in order to do that I think we need to write a script to dynamically populate deep a content type's components. See https://docs.strapi.io/cms/api/rest/guides/understanding-populate#populate-all-relations-and-fields-1-level-deep. I'm not having much luck and can't spend much more time on this. I think it's sort of low priority since this is just an example / way for us to test preview. If you want to take a look on your end feel free but personally I don't feel like it should stop us from merging this PR. Let me know what you think.

Adzouz:
Hi @markkaylor it's weird but I don't see the commit on github so I don't know what to review 🤔
I think we should remove the image completely / replace it by something like `null` don't you think?
That's what's happening by default if there's no image. Because for the moment if you remove it, you have a broken image that remains:
Screenshot 2026-02-03 at 09 56 56

markkaylor:
@Adzouz You can test the work here in launchpad on this PR https://github.com/strapi/LaunchPad/pull/125. An experimental release is provided on that PR.

markkaylor:
I'll handle blocks in a follow up PR

Adzouz:
Thanks for taking the time to reopen and work on the subject @markkaylor 🙏
It seems to work well on single media, good job 👏
I tested on Launchpad project on your PR and I have one issue with handling multiple medias:

https://github.com/user-attachments/assets/e4a41428-163d-47d7-939e-9db7de26c701

First the overlay is not over the image completely and when I select medias, it breaks 😢

Also maybe we should ask how the product is supposed to behave in the case of deleting an image because for the moment, there's no way to put another image again because the whole region disappear. Should we keep it as is or at least put a "Add image" or an empty zone to help the user put a new image back:

https://github.com/user-attachments/assets/f91d0b0b-568a-4c16-aa54-50f723dd1bc3

Tell me if I can help you with anything 🫡

markkaylor:
@Adzouz This one should be ready for review again. Could you test on Launchpad again with this new experimental: `0.0.0-experimental.46abbc94c7c3e5f19d52f47c83b99788ebb8912b`
To test the multiple media handling you can use this prompt in the LaunchPad repo to have claude update strapi and next to support that in live preview

```
Update the articles content-type in strapi to have a medias property with media multiple. Then update the page in next to have an image gallery just below the hero image
```
Demo

https://github.com/user-attachments/assets/af98bddd-7065-4303-9f60-7dd6807feef3

I spotted a couple small things I want to fix, like the weird behavior where it highlights text, also wondering if the popover should close after selection 🤔 . Let me know what you think.

Adzouz:
Looking good @markkaylor, I'll dig in the code changes but in the meantime, do you think we need to address the following use case:

https://github.com/user-attachments/assets/c943ee0f-541b-4d19-85ad-8b02fd9042ae

If there's no image set, we can't find the area to provide one (which is normal since the image tag doesn't exist). I think it's not really a big deal but let me know what you think! Thanks a lot for the patience 🙏

markkaylor:
> Looking good @markkaylor, I'll dig in the code changes but in the meantime, do you think we need to address the following use case:
> Draft.5.Reasons.Why.Your.Content.Needs.to.Go.Orbital._.Strapi.-.4.May.2026.mp4
> 
> If there's no image set, we can't find the area to provide one (which is normal since the image tag doesn't exist). I think it's not really a big deal but let me know what you think! Thanks a lot for the patience 🙏

Is the issue that the initial image has to be set in the editor before it becomes interactive in preview?

Adzouz:
> > Looking good @markkaylor, I'll dig in the code changes but in the meantime, do you think we need to address the following use case:
> > Draft.5.Reasons.Why.Your.Content.Needs.to.Go.Orbital._.Strapi.-.4.May.2026.mp4
> > If there's no image set, we can't find the area to provide one (which is normal since the image tag doesn't exist). I think it's not really a big deal but let me know what you think! Thanks a lot for the patience 🙏
> 
> Is the issue that the initial image has to be set in the editor before it becomes interactive in preview?

Yes exactly. Also I have experienced another issue when deleting the asset, publishing and trying to edit again:

https://github.com/user-attachments/assets/0dc74804-553e-4686-bee6-cade8dd4dfa1

😢

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -88,6 +251,34 @@ const previewScript = (config: PreviewScriptConfig) => {
...
+                const newSource = sourceValue.replace(`path=${originalPath}`, `path=${mediaPath}`);
+                element.setAttribute(SOURCE_ATTRIBUTE, newSource);
+              }
+            }
+            // Clean the src attribute so the resource can load
+            const cleanedSrc = stegaClean(src);
+            if (cleanedSrc !== src) {
+              element.setAttribute('src', cleanedSrc);
+            }
+          } catch (error) {}
```
Adzouz, line 266: We should maybe add at least a warn here, what do you think?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -72,6 +72,169 @@ const previewScript = (config: PreviewScriptConfig) => {
...
+   */
+  const CLONE_WRAPPER_ATTRIBUTE = 'data-strapi-media-clone-wrapper';
+  const createMediaPlaceholder = (sourceAttr: string, rect: DOMRect): HTMLElement => {
+    const placeholder = document.createElement('div');
+    placeholder.setAttribute(PLACEHOLDER_ATTRIBUTE, '');
+    placeholder.setAttribute(SOURCE_ATTRIBUTE, sourceAttr);
+
+    const width = rect.width > 0 ? rect.width : 200;
+    const height = rect.height > 0 ? rect.height : 200;
+    placeholder.style.cssText = `
```
Adzouz, line 143:
```suggestion
    Object.assign(placeholder.style, {
      display: 'inline-block',
      width: `${width}px`,
      height: `${height}px`,
      backgroundColor: 'rgba(0, 0, 0, 0.04)',
      border: '2px dashed rgba(0, 0, 0, 0.15)',
      borderRadius: '4px',
      boxSizing: 'border-box',
      verticalAlign: 'top',
    });
```

markkaylor:
@Adzouz 
New experimental (currently still building): 0.0.0-experimental.3f769f3c2888c1afef066628256321f01ac4d6c6

It should fix the delete issue. add an image, delete, save, the space where the image was is no longer interactive. 

When saved data is null, the previewScript isn't going to try and reconcile any of that instead it's up to the consumer's framework to handle this for preview. We could update launchpad to use an img tag when it doesn't have a value and in that case we'd have the behavior we want. Happy to discuss it and update launchpad. But I think let's get something merged that's working at this point 😄 and worry about other UX improvements later...?

When you delete an image it should be guaranteed that you can click the space where it was and add it back IF you do it before you save

Adzouz (approved):
I agree, let's move on and merge this baby! Thanks again for your work and patience, this one wasn't easy 🙏 I put just two little comments I forgot to address but it's not blocking

Issue #25098, opened 2025-12-31
Conditional fields do not work in dynamic zone components


yoonbuck:
### Node Version

24.12.0

### NPM/Yarn/PNPM Version

npm 11.6.2

### Strapi Version

5.33.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When a component containing conditional fields set to _show_ when a condition is met is used in a dynamic zone, the conditional fields do not appear in the content manager, regardless of whether the condition is met.

The video attached in media below, using the shown content type configuration, demonstrates that the conditional fields behave correctly in a normal component field, but do not work inside the dynamic zone.

### Steps to Reproduce

- Create a new component
  - Add a boolean or enum field
  - Add another field. In advanced settings, add a condition based on the first field, and choose "Show" under "THEN"
- Create a new collection type
  - Add a dynamic zone field containing the previously created component
- Save the content types and let the server restart
- Open the collection manager and go to the new collection type
- Create a new entry.
- Add the component to the dynamic zone
- Toggle the boolean value or change the enum.
- Observe: the dynamic field never appears

### Expected Behavior

The dynamic field should appear or disappear based on the value in the enum or boolean field their condition depends on.

### Logs

```shell

```

### Code Snippets

Schemas used for below video:

components/fields/fields.json:
```
{
  "collectionName": "components_fields_fields",
  "info": {
    "displayName": "conditional_fields"
  },
  "options": {},
  "attributes": {
    "bool": {
      "type": "boolean"
    },
    "bool_true": {
      "type": "string",
      "conditions": {
        "visible": {
          "==": [
            {
              "var": "bool"
            },
            true
          ]
        }
      }
    },
    "bool_false": {
      "type": "string",
      "conditions": {
        "visible": {
          "==": [
            {
              "var": "bool"
            },
            false
          ]
        }
      }
    }
  },
  "config": {}
}
```

api/foo/content-types/foo/schema.json:
```
{
  "kind": "collectionType",
  "collectionName": "foos",
  "info": {
    "singularName": "foo",
    "pluralName": "foos",
    "displayName": "foo"
  },
  "options": {
    "draftAndPublish": true
  },
  "pluginOptions": {},
  "attributes": {
    "zone": {
      "type": "dynamiczone",
      "components": [
        "fields.fields"
      ]
    },
    "subcomponent": {
      "type": "component",
      "component": "fields.fields",
      "repeatable": true
    }
  }
}
```

### Media

[...]

PR #25219, opened 2026-01-19
fix: handle component structure properly for conditional fields [CMS-197]

### LINKED ISSUE #25118: Strapi 5.33.1 regression: conditional fields fail on top‑level component fields after upgrade
### Node Version

22.18.0

### NPM/Yarn/PNPM Version

11.5.2

### Strapi Version

5.33.1

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

After upgrading to Strapi 5.33.1, conditional fields defined on the top level of a component no longer render in the Content Manager. The component appears in the dynamic zone, but when a condition is met (e.g., enum/boolean toggle), the dependent fields show as an empty block with no inputs. The same conditional fields worked correctly on 5.31.3.

### Steps to Reproduce

1. Upgrade Strapi from 5.31.3 to 5.33.1.
2. Open Content Manager and edit a component that has conditional fields on the top level (e.g., blocks.one-column-content).
3. Select a value that should reveal conditional fields (e.g., columnContent = cards).

### Expected Behavior

Conditional fields render normally.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

Image

Image

### Additional information

- Conditional fields inside nested components still work.
- Example: blocks.two-column-content works because conditional fields are inside shared.second-column (nested component), not on the top level of the block.
- No server-side errors.
- Downgrading back to 5.31.3 fixes the issue.

Possibly related to #25098 (conditional fields in components used in dynamic zones not rendering).

### Confirmation Checklist

Bassel17:
### What does it do?

Fixes conditional-field evaluation for dynamic zone / component inputs and cleans up duplicated conditional logic in the dynamic-zone renderer.

### Why is it needed?

Conditional fields inside dynamic zones could throw runtime errors or be evaluated against the wrong values because the rules evaluation assumed a flat form-value layout. The dynamic-zone renderer also contained duplicate/unsafe conditional logic causing inconsistencies and regressions.

### How to test it?

- Create a component with conditional fields, try all different forms of logic 
- Add component to dynamic zone
- fields should be hidden or visible as expected

### Related issue(s)/PR(s)

fixes  #25118, #25098

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
@@ -18,6 +18,7 @@ import { useDocumentLayout } from '../../../hooks/useDocumentLayout';
 import { useLazyComponents } from '../../../hooks/useLazyComponents';
 import { useHasInputPopoverParent } from '../../../preview/components/InputPopover';
 import { usePreviewInputManager } from '../../../preview/hooks/usePreviewInputManager';
+import { getDirectParent } from '../utils/data';
 
 import { BlocksInput } from './FormInputs/BlocksInput/BlocksInput';
 import { ComponentInput } from './FormInputs/Component/Input';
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/InputRenderer.tsx
```diff
@@ -276,7 +277,11 @@ const ConditionAwareInputRenderer = ({
 }: InputRendererProps & { condition: JsonLogicCondition }) => {
   // Note: this selector causes a re-render every time any form value on the page changes
   const fieldValues = useForm('ConditionalInputRenderer', (state) => state.values);
-  const isVisible = rulesEngine.evaluate(condition, fieldValues);
+  // For nested fields, we evaluate against the parent scope so conditions can read siblings.
+  // Top-level fields resolve to the full form values.
+  const targetValues = getDirectParent(fieldValues, props.name);
+
+  const isVisible = rulesEngine.evaluate(condition, targetValues);
```
butcherZ, line 284: Do we want it to be re-rendered on every form change? Maybe we can do something like this, so it only get re-rendered when the visibility changes
```
const isVisible = useForm('ConditionalInputRenderer', (state) => {
    const targetValues = getDirectParent(state.values, props.name);
    return rulesEngine.evaluate(condition, targetValues);
  });
```

packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
```diff
@@ -26,6 +26,33 @@ type AnyData = Omit<Document, 'id'>;
 
 const BLOCK_LIST_ATTRIBUTE_KEYS = ['__component', '__temp_key__'];
 
+/**
+ * @internal
+ * @description Returns the direct parent object for a dot-separated path.
+ */
+const getDirectParent = (data: unknown, path: string): unknown => {
+  if (!path) return undefined;
+  const isNumericIndex = (value: string) => /^\d+$/.test(value);
```
butcherZ, line 35: not really a big issue, but here the regex gets created each call to `getDirectParent`, could we try something like 
```
  const NUMERIC_INDEX_REGEX = /^\d+$/;

  const getDirectParent = (data: unknown, path: string): unknown => {
    if (!path) return undefined;
    const isNumericIndex = (value: string) => NUMERIC_INDEX_REGEX.test(value);
```

😄

jhoward1994:
Behaviour in the listed issues is fixed for me but I spotted a couple of other things:

- If boolean 'false' hides a field - when you first load the DZ 'false' will appear selected but the field is still visible - you have to toggle once for the correct logic to kick in
- [Minor issue] Fill a conditional fields value - hide it - then re show it. All without saving anything. The value is cleared, it feels like it should persist from the last time you filled it in?

jhoward1994 (approved):

PR #25229, opened 2026-01-20
chore(upload): setup future flag for media lib redesign work

markkaylor:
### What does it do?

- Add `unstableMediaLibrary` future flag
- Add `future` folder in `admin/`

### Why is it needed?

- To isolate our future work for the media library 

### How to test it?

In the `.env` of your Strapi application:

`UNSTABLE_MEDIA_LIBRARY=true`

Below the existing media lib icon in the side nav you should see a new icon (!) for the new media lib dev

Set the future flag env var to false, you should no longer see the new media lib

### Related issue(s)/PR(s)

### DIFF
--- a/examples/getstarted/config/features.js
+++ b/examples/getstarted/config/features.js
@@ -1,5 +1,5 @@
 module.exports = ({ env }) => ({
   future: {
-    unstableMediaLibrary: true,
+    unstableMediaLibrary: env.bool('UNSTABLE_MEDIA_LIBRARY', false),
   },
 });


examples/getstarted/config/features.js
```diff
@@ -1,3 +1,5 @@
 module.exports = ({ env }) => ({
-  future: {},
+  future: {
+    unstableMediaLibrary: true,
+  },
```
nclsndr, line 4: Context question: Do we merge unstable flags in examples in general? Do we consider all example apps the same way to this regard?

nclsndr:
Should we give the PR a more descriptive name @markkaylor?

examples/getstarted/config/features.js
markkaylor: In the past we've done that yes, but I can also set it up based on an env variable instead of hard coded like this.

nclsndr: right, no strong opinion on my end 👌

jhoward1994:
So throughout the work for the new media lib we'll merge into develop and release in the npm latest releases? But it just won't be available unless you enable this flag?

packages/core/upload/admin/src/future/App.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Page.Main>
+      <Page.Title>{title}</Page.Title>
+
+      <Routes>
+        <Route index element={<MediaLibraryPage />} />
+        <Route path="ai-generation" element={<AIGenerationPage />} />
+      </Routes>
+    </Page.Main>
+  );
+};
```
markkaylor, line 26: > So throughout the work for the new media lib we'll merge into develop and release in the npm latest releases? But it just won't be available unless you enable this flag?

@jhoward1994 Exactly. This is the entry point that is under a future flag. Everything in the `future` folder we should consider as unstable. As long as we don't touch anything outside the `future` folder and everything in the `future` folder is used as a child of this component we are safe to merge to develop. 

Does it make sense to you? We have to be a bit careful but I think it's pretty safe. Let me know if you have any concerns.

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { getTrad } from '../../utils';
+
+export const MediaLibraryPage = () => {
+  const { formatMessage } = useIntl();
+
+  return (
+    /**
+     * NOTE:
+     *
+     * The design differs from our current Layouts component.
```
Adzouz, line 16: Where exactly? Maybe I missed something here

packages/core/upload/admin/src/index.ts
```diff
@@ -25,10 +25,30 @@ const admin: Plugin.Config.AdminInput = {
...
+          id: `${pluginId}.plugin.name`,
+          defaultMessage: 'Media Library',
+        },
+        permissions: PERMISSIONS.main,
+        Component: () => {
+          return import('./future/App').then((mod) => ({
+            default: mod.UnstableMediaLibraryPage,
+          }));
+        },
+        position: 5,
```
Adzouz, line 48: Problem is that the CTB is also position 5 so it places this right after the CTB and not the previous ML. I don't know if we can do something about it though. If we change here for 4, it will move both ML + the unstable one after the CTB for some reason.

Adzouz (approved):
Looks good to me 🙏 Thanks for preparing everything 🙏 
Just a two comments, nothing really blocking

packages/core/upload/admin/src/index.ts
markkaylor: oh weird, even with this I had them right next to each other but I'll take another look before merging

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
markkaylor: Mainly where the breadcrumbs are supposed to go but Anka is working on the header again so TBD

packages/core/upload/admin/src/index.ts
markkaylor: With what I have currently on my end I have this:

Screenshot 2026-01-23 at 10 12 40

If I put them both as 4 I have this:

Screenshot 2026-01-23 at 10 12 16

Personally I'm not too bothered by where it ends up as long as we all know how to find it. If you have a preference for one over the other or see a concern here that I'm missing just let me know, happy to update.

PR #25237, opened 2026-01-21
feat(admin): app token management UI

nclsndr:
### What does it do?

Introduce the App Token Management (list and edit view) in Admin UI.
Notably expand the permissions management components to allow for App Token permission configuration.

A user can only delegate the permission they hold to an App Token.

### Why is it needed?

This feature allows users to CRUD App Tokens and granularly manage the permissions.

### How to test it?

1. Launch `getstarted` example
2. Connect as super admin
3. In the user navigation (bottom left corner), click on "App Tokens"
4. Manage app tokens - all permissions must be checkable
5. Create a new admin user with role with limited permissions
6. Logout
7. Log as the second admin user
8. In the user navigation (bottom left corner), click on "App Tokens"
9. Manage app tokens - only inherited permissions, at most, can be granted

### Related issue(s)/PR(s)


nclsndr:
We will not implement App Tokens.

PR #25243, opened 2026-01-22
fix: mobile actions drawer in content manager edit view

Adzouz:
### What does it do?

The actions are moved to a specific sticky drawer at the bottom of the page on mobile. The drawer is folded by default and can be opened by clicking on the little icon button on the right, showing the rest of the drawer content.
In this first PR, the purpose was to create the drawer component and move all actions / information of the sidebar in the content manager edit view to this drawer.

https://github.com/user-attachments/assets/610eee00-c4e9-44b9-9bf3-585cb0a586a4

### Why is it needed?

Improve the experience on mobile devices 📱

### How to test it?

1. With content inside the drawer (toggle button is visible)
- Enable review workflows / releases (EE) and/or preview
- Go to the admin
- Go to the Content Manager
- Select any content type
- Edit any entry (not create because the content is empty at that point)
- Resize the window to a mobile breakpoint
-> The document actions (Save / Publish / More actions) should be sticky on the bottom of the page and clicking on the expand button should reveal the preview feature + the review workflow + releases (if features enabled and configured - enterprise edition)
-> Note that if you save or publish the document, the toggle button will appear (because now actions related to review workflows / releases or preview are available)

2. Without content inside the drawer (toggle button is not displayed)
- Enable review workflows / releases (EE) and/or preview
- Go to the admin
- Go to the Content Manager
- Select any content type
- Create a new entry
-> The toggle button shouldn't appear if there's nothing inside the drawer to show (you can see if you resize to desktop that there's nothing in the sidebar panels except the publish / save actions)

3. Remove content from drawer during manipulations on the document\
_It's a specific use case where the toggle button could disappear_
- Enable releases and disable preview
- Go to the admin
- Go to Releases
- Create a new release (with a date or not, it doesn't matter)
- Go to the Content Manager
- Select any content type that is not assigned to any review workflow
- Create a new entry
- Click Save
- Add it to a release (three dots button in the sticky action bar at the bottom)
-> The toggle button should appear
- Open the drawer
- In the release, click the three dots button and remove the item from the release
-> The drawer should close and the toggle button disappear
Video:
https://github.com/user-attachments/assets/1ea07a63-d783-4424-a2f7-cb1709227a92

🚀


packages/core/content-manager/admin/src/pages/EditView/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+          $isOpen={isOpen}
+          $hasSideNav={hasSideNav}
+          onClick={() => setIsOpen(false)}
+          data-testid="actions-drawer-overlay"
+        />
+        <DrawerContent background="neutral0">
+          <Flex
+            paddingTop={3}
+            paddingBottom={3}
+            paddingLeft={4}
```
mathildeleg:
```suggestion
            paddingLeft={3}
```
Only `paddingRight` should be 16px according to Figma file

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+  transition: opacity 0.2s ease-in-out;
+  opacity: ${({ $isOpen }) => ($isOpen ? 1 : 0)};
+  pointer-events: ${({ $isOpen }) => ($isOpen ? 'auto' : 'none')};
+
+  ${({ theme }) => theme.breakpoints.medium} {
+    left: ${({ $hasSideNav }) => ($hasSideNav ? WIDTH_SIDE_NAVIGATION : 0)};
+  }
+`;
+
+const ToggleButton = styled.button`
```
mathildeleg: Shouldn't it be
```suggestion
const ToggleButton = styled(Button)`
```
imported from the design system?
But then I think the icon isn't centered so idk

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+                        id: 'content-manager.actions-drawer.open',
+                        defaultMessage: 'Open more actions',
+                      })}
+                </VisuallyHidden>
+              </ToggleButton>
+            )}
+          </Flex>
+          {children && (
+            <DrawerContentInner $isOpen={isOpen}>
+              <Flex direction="column" alignItems="stretch" justifyContent="flex-start" padding={4}>
```
mathildeleg: I think this may be because of the prettier update bug, but I think the props should be like this:
```suggestion
              <Flex 
                direction="column" 
                alignItems="stretch"
                justifyContent="flex-start" 
                padding={4}
              >
```

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -199,53 +202,61 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
     );
   };
 
+  const primaryActionContent = (
+    <>
+      <Flex flex={1} alignItems="stretch" direction="column">
```
mathildeleg, line 206: The button "save" should be on the left:
Image

packages/core/content-manager/admin/src/pages/EditView/components/Panels.tsx
```diff
@@ -31,7 +32,12 @@ interface PanelDescription {
  * Panels
  * -----------------------------------------------------------------------------------------------*/
 
-const Panels = () => {
+interface PanelsProps {
+  excludeActionsPanel?: boolean;
+  onContentChange?: (hasContent: boolean) => void;
+}
+
+const Panels = ({ excludeActionsPanel = false, onContentChange }: PanelsProps = {}) => {
```
mathildeleg:
```suggestion
const Panels = ({ 
  excludeActionsPanel = false, 
  onContentChange
  }: {
   excludeActionsPanel?: boolean;
   onContentChange?: (hasContent: boolean) => void;
 }) => {
```
This is how you wrote it below for `PanelItems` so might as well stay consistent wdyt?

packages/core/content-manager/admin/src/pages/EditView/components/Panels.tsx
```diff
@@ -52,22 +57,46 @@ const Panels = () => {
...
+  React.useEffect(() => {
+    onHasContentChange?.(hasContent);
+  }, [hasContent, onHasContentChange]);
+
+  if (!hasContent) {
+    return null;
+  }
+
+  return (
+    <Flex direction="column" alignItems="stretch" gap={isDesktop ? 2 : 4}>
```
mathildeleg: can you not use the `gap={ initial: 4, large: 2 }` here? 🤔

packages/core/content-manager/admin/src/pages/EditView/components/Panels.tsx
```diff
@@ -133,20 +162,18 @@ interface PanelProps extends Pick<PanelDescription, 'title'> {
...
-      background="neutral0"
-      borderColor="neutral150"
-      hasRadius
-      paddingBottom={4}
-      paddingLeft={4}
-      paddingRight={4}
-      paddingTop={4}
-      shadow="tableShadow"
-      gap={3}
+      background={isDesktop ? 'neutral0' : 'transparent'}
```
mathildeleg: same question as above, I think most of these can switched to using the `{initial: ..., large: ...}` used in the rest of the app for responsive

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -203,7 +208,7 @@ const EditViewPage = () => {
                 }}
                 gap={4}
               >
-                <Grid.Item col={9} xs={12} direction="column" alignItems="stretch">
+                <Grid.Item col={isDesktop ? 9 : 12} xs={12} direction="column" alignItems="stretch">
```
mathildeleg: same question here

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+  React.useEffect(() => {
+    if (!hasContent && isOpen) {
+      setIsOpen(false);
+    }
+  }, [hasContent, isOpen]);
+
+  return (
+    <>
+      <DrawerContainer $isOpen={isOpen} $hasSideNav={hasSideNav}>
+        <DrawerOverlay
```
mathildeleg: Shouldn't the overlay be present only if `hasContent` is true? 🤔 if there's no drawer to open, I think it can be good to hide it from the DOM, wdyt?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -21,6 +22,7 @@ import {
   ButtonProps,
   Tooltip,
   IconButton,
+  Box,
```
mathildeleg: I don't think you used this

packages/core/content-manager/admin/src/pages/EditView/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
+import * as React from 'react';
+
+import { WIDTH_SIDE_NAVIGATION } from '@strapi/admin/strapi-admin';
+import { Portal, Flex, Box, ScrollArea, VisuallyHidden } from '@strapi/design-system';
+import { CaretDown, CaretUp } from '@strapi/icons';
+import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
+
+const DrawerContainer = styled(Portal)<{ $isOpen: boolean; $hasSideNav: boolean }>`
```
markkaylor: Maybe we could take inspiration from shadcn/ui which uses [vaul](https://vaul.emilkowal.ski/) to design a reusable Drawer exported from the Design System. We will need a side drawer for the Media Lib for example and I bet we will need something like this elsewhere in the app. 

My main concern here is accessibility. I haven't tested yet but I'm not sure how accessible this thing is. Vaul is using a radix dialog primitive and I believe that is the approach we should follow.

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const { formatMessage } = useIntl();
+  const toggleOpen = () => {
+    setIsOpen((prev: boolean) => !prev);
+  };
+
+  React.useEffect(() => {
+    if (!hasContent && isOpen) {
+      setIsOpen(false);
+    }
+  }, [hasContent, isOpen]);
```
markkaylor: Maybe you explained when we talked the other day but I don't remember what this is about. Looks a bit strange. Probably deserves a comment if there is a legitimate reason for it.

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+  flex-direction: column;
+  align-items: stretch;
+  max-height: calc(100vh - 16rem);
+  position: relative;
+  z-index: 1;
+  pointer-events: auto;
+`;
+
+const DrawerContentInner = styled(ScrollArea)<{ $isOpen: boolean }>`
+  max-height: ${({ $isOpen }) => ($isOpen ? 'calc(100vh - 20rem)' : '0')};
```
markkaylor: 16rem and 20rem look like magic numbers, do they correspond to something specific? Could they be a constant variable or could we have comment to explain their value?

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }, [hasContent, isOpen]);
+
+  return (
+    <>
+      <DrawerContainer $isOpen={isOpen} $hasSideNav={hasSideNav}>
+        <DrawerOverlay
+          $isOpen={isOpen}
+          $hasSideNav={hasSideNav}
+          onClick={() => setIsOpen(false)}
+          data-testid="actions-drawer-overlay"
```
markkaylor: I think this is a hint that something isn't quite right with accessibility.

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -199,53 +202,61 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
     );
   };
 
+  const primaryActionContent = (
+    <>
+      <Flex flex={1} alignItems="stretch" direction="column">
+        {primaryAction.label === 'Publish'
```
markkaylor, line 207: Are you sure this label is always in english 🤔 ?

packages/core/content-manager/admin/src/pages/EditView/components/ActionsDrawer.tsx
Adzouz: We saw that the button was slightly off and it was weird and agreed on aligning with the content no?

packages/core/content-manager/admin/src/pages/EditView/components/Panels.tsx
```diff
@@ -52,22 +57,46 @@ const Panels = () => {
...
+}: {
+  panels: (PanelDescription & { id: string })[];
+  onHasContentChange?: (hasContent: boolean) => void;
+}) => {
+  const isDesktop = useIsDesktop();
+  const hasContent = panels.length > 0;
+
+  React.useEffect(() => {
+    onHasContentChange?.(hasContent);
+  }, [hasContent, onHasContentChange]);
```
markkaylor: There are two things I don't love here:
1. This creates prop drilling that is hard to follow
2. Using an effect to synchronize react state

I don't have a solution per se but I've explored it a bit and have some ideas we could try. Could we pair later today or Monday?

packages/core/content-manager/admin/src/pages/EditView/components/ActionsDrawer.tsx
markkaylor: A refactor could also be part of Media Lib drawer dev not necessarily a request for change here if we have a plan that won't sit in a backlog 🙃

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
Adzouz: It can happen that the content changes during updates of the document and if for some reason the content becomes empty, it would close the drawer because it would make sense to keep it open if there's nothing to display. There's an example in the third point of the How to test it section of the PR 😊 but I definitely can add a comment!

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
Adzouz: Yeah I know... It's just an arbitrary value to still be able to see the header / title in the content behind. I couldn't come up with a better solution because putting a value only in vh would depend on the height of the device and would hide more or less content. Do you have an idea?

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
Adzouz: It's because I'm trying to target the overlay in my test to see if it correctly close the drawer and I didn't want to add a class on the element just for the test... I don't really get the accessibility issue here?

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
Adzouz: That's actually a very good question. It was already like that before, I just moved that part 😅 I'll investigate!

packages/core/content-type-builder/admin/src/components/ContentTypeBuilderNav/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -553,6 +541,7 @@ exports[`<ContentTypeBuilderNav /> renders and matches the snapshot 1`] = `
   }
 }
 
+<<<<<<< HEAD
 @media (min-width: 1080px) {
```
mathildeleg, line 556: I think you may still have a conflict to solve here

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,254 @@
...
+  pointer-events: auto;
+`;
+
+const DrawerContentInner = styled(ScrollArea)<{ $isOpen: boolean }>`
+  max-height: ${({ $isOpen }) =>
+    $isOpen
+      ? 'calc(100vh - 25rem)' // 25rem is arbitrary, to be able to see a bit of the content behind (navigation and header)
+      : '0'};
+  overflow: hidden;
+  transition: max-height 0.2s ease-in-out;
```
mathildeleg: maybe you could use one of the theme.motion.easings transitions to smooth this?

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
```diff
@@ -0,0 +1,254 @@
...
+ * -----------------------------------------------------------------------------------------------*/
+
+const Overlay = () => {
+  const ctx = useActionsDrawer('ActionsDrawer.Overlay', (s) => s, false);
+  const isOpen = ctx?.isOpen ?? false;
+  const hasContent = ctx?.hasContent ?? false;
+  const hasSideNav = ctx?.hasSideNav ?? false;
+  const setIsOpen = ctx?.setIsOpen;
+
+  if (!hasContent) {
```
mathildeleg, line 147: It's a bit weird on devices > 768px & < 1080px to have the drawer (especially because of the overlay), wdyt?
Image

packages/core/content-manager/admin/src/pages/EditView/components/Panels.tsx
```diff
@@ -42,6 +70,15 @@ const Panels = () => {
   });
   const { model, id, document, meta, collectionType } = useDoc();
   const plugins = useStrapiApp('Panels', (state) => state.plugins);
+  // Optional context consumer
+  const setVisiblePanels = usePanelsContext('Panels', (s) => s.setVisiblePanels, false);
```
mathildeleg, line 74: Good idea to use a context here, I thought about it but wasn't sure it'd make sense for this case, but I'm glad it does 👌🏻

mathildeleg (commented):
Everything looks good to me, just a few suggestions, but otherwise well done! Almost there 💪🏻

packages/core/content-manager/admin/src/components/ActionsDrawer.tsx
Adzouz: Why do you think it's weird?
If I put back the sidebar it's a nightmare 😅
Screenshot 2026-01-30 at 15 56 14

Adzouz: I can extend the overlay to the sidebar if you want?
Screenshot 2026-01-30 at 15 59 55

mathildeleg: It's a bit weird too tbh 😂 let's leave it like that that's ok :)

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
```diff
@@ -214,11 +219,30 @@ const EditViewPage = () => {
...
+              <>
+                <ActionsDrawer.Root hasContent={drawerHasContent} hasSideNav>
+                  <ActionsDrawer.Overlay />
+                  <ActionsDrawer.Header>
+                    <ActionsPanelContent />
+                  </ActionsDrawer.Header>
+                  <ActionsDrawer.Content>
+                    <Panels withActions={false} />
+                  </ActionsDrawer.Content>
+                </ActionsDrawer.Root>
```
markkaylor, line 239: Hey! that looks pretty clean 👍 nice

mathildeleg (dismissed):
🎊✨

packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
Adzouz: 🫶

PR #25244, opened 2026-01-22
security: update lodash to 4.17.23

innerdvations:
### What does it do?

updates lodash from 4.17.21 to 4.17.23

### Why is it needed?

resolves security warnings -- unconfirmed if Strapi is vulnerable or not

### How to test it?

everything should work as before

### Related issue(s)/PR(s)

CMS-223


nclsndr:
Should we bump `lodash-es` as well? 
https://github.com/strapi/strapi/pull/25239

innerdvations:
> Should we bump `lodash-es` as well? #25239

Ahh, definitely, good catch. I'll merge the dependabot one once the tests pass

PR #25247, opened 2026-01-23
fix(i18n): ai translation losing unsupported fields

butcherZ:
### What does it do?

This PR fixes the AI translation feature to preserve unsupported field types (media, boolean, enumeration, relation) when updating localized content.

1. Added mergeUnsupportedFields() helper function - Recursively merges unsupported field types from a source document into AI-translated content. 
2. This handles:
    - Root-level unsupported fields
    - Nested fields inside single components
    - Nested fields inside repeatable components
    - Nested fields inside dynamic zones
   
 3. Added buildDeepPopulate() helper function - Builds a deep populate object to fetch all nested fields including media inside components, ensuring we have complete data to merge.
 4. Updated generateDocumentLocalizations() to:
    - Fetch the source document with all fields deeply populated
    - Fetch the existing derived locale document with all fields deeply populated
    - Use mergeUnsupportedFields() to merge unsupported fields from the existing derived doc (or source doc for new locales) into
  the AI-translated content
  5. Added unit tests
 
### Why is it needed?

  When the AI translation feature runs, it filters out unsupported field types (media, boolean, enumeration, relation) before
  sending content to the AI service - this is correct since these fields cannot be meaningfully translated.

  However, when updating the localized document with AI-translated content, the previous implementation only restored root-level media fields. This caused all other unsupported fields to be lost, including:

- Boolean fields 
- Enumeration fields 
- Media fields inside components 
- Any unsupported fields inside nested components/dynamic zones 

### How to test it?

 1. Enable AI translations in Settings > Internationalization
  2. Create a content type with:
    - A localized text field (e.g., title)
    - A localized boolean field (e.g., isFeatured)
    - A localized enumeration field (e.g., backgroundColor)
    - A localized media field (e.g., featuredImage)
    - A localized component containing media/boolean/enum fields
  3. Create content in the default locale with values for all fields
  4. Wait for AI translation to complete
  6. Verify all fields are present in the translated locale
  7. Edit the title in the default locale and save
  8. Verify all unsupported fields (boolean, enum, media, nested fields) are preserved in the translated locale

### Related issue(s)/PR(s)

fixes CMS-221

### DIFF
--- a/packages/plugins/i18n/server/src/services/ai-localizations.ts
+++ b/packages/plugins/i18n/server/src/services/ai-localizations.ts
@@ -13,115 +13,50 @@ const UNSUPPORTED_ATTRIBUTE_TYPES: Schema.Attribute.Kind[] = [
   'enumeration',
 ];
 
-const IGNORED_FIELDS = ['id', 'documentId', 'createdAt', 'updatedAt', 'updatedBy', 'localizations'];
+const IGNORED_FIELDS = [
+  'id',
+  'documentId',
+  'createdAt',
+  'updatedAt',
+  'publishedAt',
+  'locale',
+  'updatedBy',
+  'createdBy',
+  'localizations',
+];
 
 /**
- * Recursively merges unsupported field types (media, boolean, enumeration, relation)
- * from a source document into the target data object.
- *
- * This preserves fields that cannot be translated by AI but should be kept in localized versions.
+ * Deep merge where target values take priority over source values.
+ * Arrays are merged by index to align repeatable component / dynamic zone items.
  */
-const mergeUnsupportedFields = (
-  targetData: Record<string, any>,
-  sourceDoc: Record<string, any> | null,
-  schemaAttributes: Record<string, Schema.Attribute.AnyAttribute>,
-  getModel: (uid: UID.Schema) => Schema.Schema | undefined
+const deepMerge = (
+  source: Record<string, any>,
+  target: Record<string, any>
 ): Record<string, any> => {
-  if (!sourceDoc) {
-    return targetData;
-  }
+  const result = { ...source };
 
-  const result = { ...targetData };
+  for (const key of Object.keys(target)) {
+    const sourceVal = source[key];
+    const targetVal = target[key];
 
-  for (const [key, attribute] of Object.entries(schemaAttributes)) {
-    if (IGNORED_FIELDS.includes(key)) {
-      continue;
-    }
-
-    // For unsupported types, copy from source document if not already in target
-    if (UNSUPPORTED_ATTRIBUTE_TYPES.includes(attribute.type)) {
-      if (result[key] === undefined && sourceDoc[key] !== undefined) {
-        result[key] = sourceDoc[key];
-      }
-      continue;
-    }
-
-    // For components, recursively merge unsupported fields
-    if (attribute.type === 'component') {
-      const componentModel = getModel(attribute.component);
-      if (!componentModel) continue;
-
-      if (attribute.repeatable) {
-        // Repeatable component (array)
-        const targetArray = result[key] as any[] | undefined;
-        const sourceArray = sourceDoc[key] as any[] | undefined;
-
-        if (targetArray && sourceArray) {
-          result[key] = targetArray.map((targetItem, index) => {
-            const sourceItem = sourceArray[index];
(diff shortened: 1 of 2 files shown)


packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -6,26 +6,165 @@ const isLocalizedAttribute = (attribute: Schema.Attribute.Attribute | undefined)
...
+    }
+  }
+
+  return result;
+};
+
+/**
+ * Builds a deep populate object for all fields including nested components
+ */
+const buildDeepPopulate = (
```
markkaylor: I think the content-manager exposes a similar function but not sure it does exactly what you need here. Did you look at 
```ts
const populateBuilderService = strapi.plugin('content-manager').service('populate-builder');

const populate = await populateBuilderService(action.contentType).populateDeep(Infinity).build();
```

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -6,26 +6,165 @@ const isLocalizedAttribute = (attribute: Schema.Attribute.Attribute | undefined)
...
+
+const IGNORED_FIELDS = ['id', 'documentId', 'createdAt', 'updatedAt', 'updatedBy', 'localizations'];
+
+/**
+ * Recursively merges unsupported field types (media, boolean, enumeration, relation)
+ * from a source document into the target data object.
+ *
+ * This preserves fields that cannot be translated by AI but should be kept in localized versions.
+ */
+const mergeUnsupportedFields = (
```
markkaylor: There is a traverse entity utility https://contributor.strapi.io/docs/core/utils/traverse-entity. I wonder if we could leverage that for traversing the entry instead of introducing new logic we will have to maintain and test.

markkaylor: Did traverseEntity not work out in the end?

butcherZ: @markkaylor lol it worked but  I forgot to add that part since I messed up my git history, just pushed again!

markkaylor: ok cool thanks! I had a feeling it may have been something like that 🤗  I'll pull and test

butcherZ: thanks mate!!

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -6,26 +6,110 @@ const isLocalizedAttribute = (attribute: Schema.Attribute.Attribute | undefined)
...
+ */
+const deepMergeOverlay = (
+  base: Record<string, any>,
+  overlay: Record<string, any>
+): Record<string, any> => {
+  const result = { ...base };
+
+  for (const key of Object.keys(overlay)) {
+    const baseVal = base[key];
+    const overlayVal = overlay[key];
```
markkaylor: I just have a bit of a nitpick with the base vs overlay wording, is this supposed to delineate the same difference as source vs target? If so could we stay consistent with that naming? Regardless I'm a little confused by what is meant by "overlay" specifcally.

packages/plugins/i18n/server/src/services/ai-localizations.ts
```diff
@@ -6,26 +6,110 @@ const isLocalizedAttribute = (attribute: Schema.Attribute.Attribute | undefined)
...
+      // Remove supported (translatable) fields
+      remove(key);
+    },
+    { schema, getModel: getModel as (uid: string) => Schema.Schema },
+    sourceDoc
+  );
+
+  // Deep merge: AI-translated target takes priority over source unsupported fields
+  return deepMergeOverlay(unsupportedFieldsOnly, targetData);
+};
```
markkaylor, line 138: This looks very clean and easy to read 🙏

butcherZ:
> Hey @butcherZ thanks again for the updates. Unfortunately after the refactor I keep getting this error when saving the source entry:
> 
> `[2026-01-27 12:01:38.536] error: AI Localizations generation failed Invalid relations`
> 
> You don't have that on your end? we can also do a sync QA this afternoon

@markkaylor sorry about that, I just fixed it and tested it, it should be good now, could you please pull again? 😄

PR #25256, opened 2026-01-26
future(upload): add upload via filepicker

markkaylor:
### What does it do?

- Add simple menu to header
- Add menu item "import files"
- Add upload RTK service endpoint
- Uploads selected files
- Sets up e2e tests for the new media library

### Why is it needed?

To implement the new design

### How to test it?

Automated

e2e tests must be configured with the future flag. In `tests/e2e/.env` add `UNSTABLE_MEDIA_LIBRARY=true`. This future e2e test can only be run locally it does not run in the CI.

Run `yarn test:e2e -d=media-library` to run the tests

Manual

- Go to the unstable media library (/admin/plugins/unstable-upload)
- Click New => import files
- Select some files (images, videos, documents, etc..)
- Click open, it should make a POST request to `/upload`
- If successful or error you should see a notification

### DIFF
--- a/packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
+++ b/packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
@@ -6,7 +6,7 @@ import { ChevronDown, Files } from '@strapi/icons';
 import { useIntl } from 'react-intl';
 
 import { useUploadFilesMutation } from '../services/api';
-import { getTranslation } from '../utils/translations';
+import { getTranslationKey } from '../utils/translations';
 
 export const MediaLibraryPage = () => {
   const { formatMessage } = useIntl();
@@ -47,7 +47,7 @@ export const MediaLibraryPage = () => {
           type: 'success',
           message: formatMessage(
             {
-              id: getTranslation('assets.uploaded'),
+              id: getTranslationKey('assets.uploaded'),
               defaultMessage:
                 '{number, plural, one {# asset} other {# assets}} uploaded successfully',
             },
@@ -80,11 +80,11 @@ export const MediaLibraryPage = () => {
             popoverPlacement="bottom-end"
             variant="default"
             endIcon={<ChevronDown />}
-            label={formatMessage({ id: getTranslation('new'), defaultMessage: 'New' })}
+            label={formatMessage({ id: getTranslationKey('new'), defaultMessage: 'New' })}
           >
             <MenuItem onSelect={handleFileSelect} startIcon={<Files />}>
               {formatMessage({
-                id: getTranslation('import-files'),
+                id: getTranslationKey('import-files'),
                 defaultMessage: 'Import files',
               })}
             </MenuItem>
--- a/packages/core/upload/admin/src/future/utils/translations.ts
+++ b/packages/core/upload/admin/src/future/utils/translations.ts
@@ -1,3 +1,3 @@
 import { pluginId } from '../../pluginId';
 
-export const getTranslation = (id: string) => `${pluginId}.${id}`;
+export const getTranslationKey = (id: string) => `${pluginId}.${id}`;


packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -1,38 +1,83 @@
...
+      } catch (error) {
+        toggleNotification({
+          type: 'danger',
+          message: t('upload.generic-error', 'An error occurred while uploading the file.'),
+        });
+      }
+    }
+    // Reset input so the same file can be selected again
+    e.target.value = '';
+  };
```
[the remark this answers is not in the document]
markkaylor: Handled in a later PR

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -1,38 +1,83 @@
...
+        formData.append(
+          'fileInfo',
+          JSON.stringify({
+            name: file.name,
+            caption: null,
+            alternativeText: null,
+            folder: null,
+          })
+        );
+      });
```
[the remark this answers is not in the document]
markkaylor: I don't think this is correct...

packages/core/upload/admin/src/future/services/api.ts
```diff
@@ -1,7 +1,21 @@
...
-const uploadApi = adminApi.enhanceEndpoints({
-  addTagTypes: ['Asset', 'Folder'],
-});
+const uploadApi = adminApi
+  .enhanceEndpoints({
+    addTagTypes: ['Asset', 'Folder'],
+  })
+  .injectEndpoints({
+    endpoints: (builder) => ({
+      uploadFiles: builder.mutation<unknown, FormData>({
```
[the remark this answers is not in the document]
jhoward1994: I agree with this - is there something in the shared contracts for the upload
package we can use ?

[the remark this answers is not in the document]
markkaylor: Yeah same, I think yes there is. I already started working on it and some of the other feedback from claude just haven't pushed yet.

packages/core/upload/admin/src/future/hooks/useTranslation.ts
```diff
@@ -0,0 +1,22 @@
...
+  function t(
+    id: string,
+    defaultMessage?: string,
+    values?: Parameters<typeof formatMessage>[1]
+  ): string {
+    const result = formatMessage({ id: `${pluginId}.${id}`, defaultMessage }, values);
+    return String(result);
+  }
+
+  return { t };
```
Adzouz: Everywhere in the CMS we use `formatMessage`, wouldn't it confuse us using `t` instead? I understand what you're trying to do though, maybe overriding the default formatMessage?

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -1,38 +1,83 @@
...
+          message: t(
+            'assets.uploaded',
+            '{number, plural, one {# asset} other {# assets}} uploaded successfully',
+            { number: filesArray.length }
+          ),
+        });
+      } catch (error) {
+        toggleNotification({
+          type: 'danger',
+          message: t('upload.generic-error', 'An error occurred while uploading the file.'),
```
Adzouz: Could be `files` if multiple no?

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -1,38 +1,83 @@
...
-    /**
-     * NOTE:
-     *
-     * The design differs from our current Layouts component.
-     * Either we find a way to make it work with our current Layouts component
-     * or we will have to write our own custom layout.
-     */
     <Layouts.Root>
+      <VisuallyHidden>
+        <input type="file" ref={fileInputRef} onChange={handleFileChange} multiple />
```
Adzouz, line 74: Claude is right, we allow all types of file in the frontend no? If so, we might need to add some filtering here before calling the upload endpoint. Wdyt?

Adzouz (commented):
Good first task 🚀 It's clean (finally some e2e testing on the media library 🙌) I put just some little comments!

packages/core/upload/admin/src/future/hooks/useTranslation.ts
markkaylor: I had the idea and thought we could try something new? I think we can try some new patterns in the new media lib dev if we want. If we like some new things we are doing we can incorporate them elsewhere in the codebase. I got here because first I want to move and rename the `getTrad` utitlity `getTranslation` then I saw in the new CTB they are using a similar hook:

https://github.com/Schero94/strapi/blob/4e8563e97a5cfb762655351b57d4df6a7726bdc6/packages/core/content-type-builder/admin/src/components/AIChat/hooks/useTranslations.ts#L1-L12

I think what's also beneficial about this approach is if we wanted to ever replace formatMessage/react-intl it could potentially be changed in one place as long we maintain the same api.

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
markkaylor: I think I pushed again now to display a formatted error message from the server

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
markkaylor: I did consider adding the `accept` property but currently we accept all file types so for now I just decided to maintain that behavior. I'll look again though. 

I think ideally unsupported file types:
- should not be possible to select from the filepicker
- the ui would communicate the file is unsupported and will not be uploaded
- filter unsupported files out before sending to the server

It gets a bit more complex when you consider we now have the feature to allow users to control allowedTypes /disallowedTypes. Currently this information is not available on the FE and we would have to incorporate that into how the FE behaves. I think I will create an issue for this on the [Chores parent issue](https://linear.app/strapi/issue/CMS-242/chores) or add it to the backlog, it's starting to sound like it could be its own PR

Adzouz: As long as we don't forget it :) Because even if it's like that today, I have this feeling that it's wrong and that we should have this check before sending anything to the backend. Resolving for now!

markkaylor: Definitely, I'm still going to take a look to see what I can do now and might open a new PR to address this feedback, but want to merge this to give Jamie access to e2e setup and the useTranslation hook. (happy to remove that in another PR as well if y'all are against it)

packages/core/upload/admin/src/future/hooks/useTranslation.ts
markkaylor: I'll revert to using formatMessage

Adzouz: (we discussed that together but writing my concern here)
It was more about consistency (even if I get the point) since we're using formatMessage everywhere in the CMS. I'm sure other people that would want to contribute will miss that and would still use formatMessage 😕 My suggestion would be to keep formatMessage (and not t) but overriding it and use the plugin's definition of the method instead of the global one.

markkaylor: updated

packages/core/upload/admin/src/future/utils/translations.ts
```diff
@@ -0,0 +1,3 @@
+import { pluginId } from '../../pluginId';
+
+export const getTranslation = (id: string) => `${pluginId}.${id}`;
```
Adzouz: Wouldn't it be clearer to call it getTranslationKey?

markkaylor: sure

Issue #58, opened 2015-12-31
ORM roadmap


akoudal:
Hi,

First of all: great job! I've been exploring Strapi for the past few days, and so far I'm very impressed. 
I've been using it to prototype an API, and it was super easy to set up and plug in to my existing database. However I just saw in the documentation, that youre planning on switching to sequelize - are you planning on removing support for mongodb?

Happy new year,
Allan

devinivy:
I'm also just generally curious why strapi has chosen to switch ORMs.

dj-hedgehog:
@devinivy 
As @loicsaintroch said on the Slack #framework channel:

>Hey @channel! As you all know we currently use the Waterline ORM. Unfortunately, there is no more maintainers taking care of it, both the core and the adapters... So, we definitely need to change the ORM since most of the issues we encounteer on Strapi are due to Waterline. We really need to know: which DBs do you use? What ORM do you like and why?

Please join the Strapi Slack instance for such questions at http://slack.strapi.io/

loicsaintroch:
Hey!

Sorry for the lack of communication those last few days— we were just enjoying holidays. Time to get back to work!

First of all, Strapi needs a production-ready ORM which is not the case of Waterline mainly because adapters are too unstable and not well maintained to make a stable product like we need. We can’t give an all-in-one stack with some bugs coming from a dependency. If you take a look at our issues almost all of them are due to Waterline.

Also, every NoSQL database has its own use and specific features. For example, an admin panel can’t really interact the same way with a MySQL database than with a Neo4j database. We can’t take advantage of every NoSQL database specialties. NoSQL is good but for specific use. This is not our case.

Strapi uses more and more relations as long as you follow the conventional structure. That’s why a SQL ORM is more appropriate. The main Node.js SQL ORMs are Sequelize and Bookshelf. The both of them have pros and cons. We still don’t know which one to use. I personally prefer Bookshelf for its approach using the Knex query builder and the “Rails-like” structure but Sequelize “seems” more stable— need feedback here.

We thought about being ORM agnostic which is technically great. Unfortunately this is not our philosophy since we want an all-in-one stack— we can call it a Data Management Framework. If we support several ORMs we won’t be able to have a marketplace since every item won’t have the same structure. However, you’ll still be able to use your own ORM on your side but the Studio won’t work with it. We’ll only support the official ORM used by the Strapi framework.

[...]

PR #25267, opened 2026-01-27
feat(upload): add focal point picker for images

Schero94:
### What does it do?
Adds a focal point picker to the Media Library that allows users to set a focus point on images. This point can be used by frontend applications to intelligently crop images while keeping the important area in view.

**Technical changes:**

**Backend:**
- Added `focalX` and `focalY` fields to the file schema (float values 0-100, default 50)
- Added validation for focal point values in both admin and content-api
- Updated upload service to persist focal point data

**Frontend:**
- New focal point mode in the image preview (click the crosshair icon)
- Visual marker showing the current focal point position
- Hover preview: semi-transparent marker follows cursor while in edit mode
- Live coordinate display (x/y percentage) in a badge
- Reset button to restore focal point to center (50/50)
- Crosshair cursor only when hovering over the image

### Why is it needed?

Currently, Strapi has no built-in way to define which part of an image should be preserved when cropping or resizing. Frontend developers often need to display images in different aspect ratios (hero images, thumbnails, cards), and without a focal point, important subjects can be cut off.

This feature is commonly requested and exists in other CMS platforms. It enables:
- Responsive image cropping that keeps the subject in view
- Better control over how images appear in different contexts
- CSS `object-position` can use these values directly

### How to test it?

1. Start Strapi with the upload plugin
2. Upload an image to the Media Library
3. Click on the image to open the edit dialog
4. Click the crosshair icon in the toolbar (top right of preview)
5. Click anywhere on the image to set the focal point
6. Observe:
   - A marker appears at the clicked position
   - Hovering shows a preview marker
   - Coordinates are displayed in a badge
   - Reset button appears to restore to center
7. Save the asset and verify `focalX`/`focalY` are persisted

### Related issue(s)/PR(s)
Implements [RFC #58: Focal Point Picker](https://github.com/strapi/rfcs/pull/58)

Based on the RFC proposal by iFixit from November 2023. =)


packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/PreviewBox.tsx
```diff
@@ -235,30 +326,56 @@ export const PreviewBox = ({
...
+              <FocalPointAim $focalPoint={focalPoint} $isPreview={false}>
+                <FocalPointHalo />
+              </FocalPointAim>
+            )}
+
+            {/* Show hover preview as additional semi-transparent marker */}
+            {isInFocalPointMode && hoverFocalPoint && (
+              <FocalPointAim $focalPoint={hoverFocalPoint} $isPreview={true}>
+                <FocalPointHalo />
+              </FocalPointAim>
```
markkaylor: There's a pretty bad lag with the preview. I'm wondering if the preview is necessary though since we already have the crosshair. wdyt, should we just remove the preview?

https://github.com/user-attachments/assets/376152ad-09c9-42e2-84f4-2ce6f0a0e22c

packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/PreviewBox.tsx
```diff
@@ -165,6 +186,56 @@ export const PreviewBox = ({
...
+    setHasFocalPointIntent(false);
+    onFocalPointFinish(focalPoint);
+  };
+
+  const handleFocalPointReset = () => {
+    setFocalPoint({ x: 50, y: 50 });
+    setHoverFocalPoint(null);
+  };
+
+  const isInFocalPointMode = hasFocalPointIntent === true;
```
markkaylor: A bit of a nitpick but what's the reason for this? Couldn't the react state just be called `isInFocalPointMode` that's a `boolean` with default to `false` and we wouldn't need this at all?

```tsx
const [isInFocalPointMode, setIsInFocalPointMode] = React.useState<boolean>(false);
```

packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/AssetPreview.tsx
```diff
@@ -25,6 +25,8 @@ interface AssetPreviewProps {
   name: string;
   url: string;
   onLoad?: () => void;
+  onClick?: (e: React.MouseEvent<HTMLElement>) => void;
+  style?: React.CSSProperties;
 }
```
markkaylor: Maybe we should just send all the element properties through with something like this 🤷‍♂️ 

```suggestion
interface AssetPreviewProps extends React.HTMLAttributes<HTMLElement> {
  mime: string;
  name: string;
  url: string;
}
```

markkaylor (commented):
Thank you for this PR and nice work! A few comments but I think we will be able to merge soon.

I think the failing test is just a snapshot update that can be done with:

`yarn test:front packages/core/upload --updateSnapshot`

packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/PreviewBox.tsx
Schero94: Lets remove it, maybe if user wanted it bring it back was just my shot, of making things a bit better from my perspective

Schero94: That’s weird. I don't experience any lag, mhm. Well, I need to look into that first before removing. Maybe I missed something.

Schero94:
@markkaylor Thanks a lot for your review, now it should be deploy ready ! :D

packages/core/upload/admin/src/components/EditAssetDialog/PreviewBox/PreviewBox.tsx
```diff
@@ -2,7 +2,7 @@
 import * as React from 'react';
 
 import { Flex, IconButton } from '@strapi/design-system';
-import { Crop as Resize, Download as DownloadIcon, Trash } from '@strapi/icons';
+import { Crop as Resize, Download as DownloadIcon, Trash, PinMap } from '@strapi/icons';
```
claudiobgit, line 5: Maybe it would be more clear for the user if here we import `Plus` (to mimic the cross pointer) or `Pin` instead of `PinMap`, as the latter is more widely associated with maps and Geo-location things.

noelboss:
Was thinking about coding this yesterday! Ha! THANK YOU!!!

claudiobgit:
Awesome!
Thanks for the focus thing as it's VERY VERY useful in our CMS.

Il giorno gio 5 feb 2026 alle ore 02:39 noelboss ***@***.***>
ha scritto:

Schero94:
If you need something, Community working together makes all the difference.  Just let me know, I’m always up for a good team effort. As I mentioned, it was @pierreburgy who gave me the chance to shine in this feature. Please give him a shout out too; he’s doing an amazing job every day!

claudiobgit:
Big shout out @pierreburgy for everything Strapi. It rules!

Skaters thank too as this feature put us back on focus as you see in before/after frontend app mobile screenshot.

before-after

_Picture on the left  (pre-feature implementation) could be cool but would be totally out of our scope :)_

Issue #20270, opened 2024-05-06
Strapi import functionality  is very very very slow (max 45KB/s)


alexbudin:
### Required System information

- Node.js version: 18.20
- NPM version: 10.5
- Strapi version: 4.24.1
- Database: SqlLite
- Operating system: linux-x64

### Describe the bug

An export takes less than a minute, but an strapi import takes hours.
My export file is 180MB, mostly because of images.

When i try an import it it takes ages to do

this is the transfer rate of an import of a file that sits on the same machine that is not even encrypted. The screenshot below is of an in-progress transfer. Teh speeds fluctuate between 20KB to 44.8KB. The funny thing is that it never goes over 44.9KB/s

<img width="537" alt="image" src="https://github.com/strapi/strapi/assets/832030/80b9eaa0-57fb-47f7-a0da-30a27884ac5d">

I thought of trying strapi export / import since the transfer option has never worked. I think most people using Strapi agree that you need a miracle to make a transfer work, but regardless, this is about the import functionality.

I dont understand how it can copy a file at the rate of 44KB/s. What does this import do? How does it work?

npm run strapi import -- -f export_20240506221330.tar.gz

### Steps to reproduce the behavior

Just do an import of a previously exported strapi instace that has assets.

### Expected behavior

An import that does not take 56-67 minutes for 180MB or images.

alexbudin:
This is the total import time

<img width="532" alt="image" src="https://github.com/strapi/strapi/assets/832030/007f3305-f4d3-4872-8ab5-2f6475b6c590">

75minutes (1h 15minutes) - this is crazy

alexbudin:
Also seems the Logos did not import at all :(

<img width="582" alt="image" src="https://github.com/strapi/strapi/assets/832030/06ee53ac-502d-4114-80f9-06c066fbb429">

I swear its better off to copy file manually from one instance to another than use all these tools.

alexbudin:
Any help here please?

alexbudin:
This is not a question. Its a bug

alexbudin:
@derrickmehaffy can you please help. This is not a question.

0xPNG:
I have the same issue.

@alexbudin I've noticed that if I export with the `--no-compress` option, the import is significantly quicker - although still slow.

N-index:
Same issue, 15.8Kb/s

xmak:
Same issue here. It's astonishing that the import takes extremely more time than downloading the export file.

[...]

PR #25268, opened 2026-01-27
fix: implement asset metadata cache

### LINKED ISSUE #21276: yarn strapi import -f ... IS VERY SLOW

- Node.js version: v20.15.0
- NPM version: 10.7.0, but i use yarn 1.22.22
- Strapi version: 4.25.8
- Database: postgresql
- Operating system: macos
- Is your project Javascript or Typescript: ts

### Describe the bug

strapi import is very very slow

1. export data
2. import data
3. look how from 15 mb/s speed is going to 200 kb/s
4. transfer is about 1 hour + long

I tried to use export with enc, but then switched to no enc, but there's no difference. compress / no-compress was not tried, but I saw there's no difference, referencing to https://github.com/strapi/strapi/issues/20270#issuecomment-2121820608

Lucurious:
### What does it do?

This PR adds caching to the file-based data imports.
Instead of iterating through all asset-files in the archive and then for each file re-iterating the files for the matching metadata file (O(n^2)), this caches the metadata up front in a single pass (O(n)).
This should result in a massive performance improvement (especially in large datasets), at the cost of increased memory consumption. 

### Why is it needed?

The previous implementation slows down to <20KB/s on large datasets, even when using powerful systems and uncompressed archives, which results in practically unusably slow imports for large datasets (see #21276). Importing 3GB of assets at this speed would take almost 48 hours.
With this implementation, the same import would take less than half a minute.

While this implementation does consume slightly more memory, this trade-off is worth it because the total increase should be fairly limited. Typical metadata-files are rarely ever larger than 8kB, so even with 25,000 assets this should come down to < 200MB of memory usage. This effect should therefore practically only be noticeable in huge datasets, for which the previous implementation does not cater at all.

### How to test it?

Verify the speed-up by running:
`yarn strapi import -f <filename>`

The import behavior remain the same, however a significant speed-up should be visible when assets are imported.
For my own dataset with 25,000 assets, totaling to ~3GB, this resulted in:
Memory usage: +16.4 MB
Transfer speed: 18 kB/s -> 118 MB/s
Transfer time: DNF -> 26,850 ms

### Related issue(s)/PR(s)

Fixes #21276 (and #20270, which has been incorrectly closed)

### DIFF
--- a/packages/core/data-transfer/src/file/providers/source/index.ts
+++ b/packages/core/data-transfer/src/file/providers/source/index.ts
@@ -57,6 +57,8 @@ class LocalFileSourceProvider implements ISourceProvider {
 
   #diagnostics?: IDiagnosticReporter;
 
+  #assetMetadataCache?: Map<string, any> = null;
+
   constructor(options: ILocalFileSourceProviderOptions) {
     this.options = options;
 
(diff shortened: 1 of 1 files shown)


cayblood:
Anxious to see this approved. Why over one month with no updates?

julienR2:
Yep same issue here, and it doesn't seem there's anything blocking the progress of this PR right?

innerdvations:
Thanks, this looks good but needs to be updated with the latest `develop` branch and rewritten for tar 7.

Lucurious:
Good to hear, will do!

Lucurious:
I've rebased and updated the code to work with tar v7 (and made a few small additional improvements), so it should be ready to merge now.

Issue #23512, opened 2025-05-13
Component Relation Field Appears Empty in Admin Panel After API Update


xrgouda:
### Node Version

20.10.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.13.0

### Operating System

Strapi Cloud

### Database

MySQL

### Javascript or Typescript

Typescript

### Bug Description

When updating a component via API that contains a relation field, the relation field appears empty in the Strapi Admin Panel. This happens even though the API correctly returns the related data. The dropdown is empty in the UI, and the relation is not displayed, even though it exists in the database and is correctly associated.

### Steps to Reproduce

Create a component that includes a relation field (e.g. to another collection type).

Add the component to a collection type or single type.

Use the Strapi API to update an entry and set the relation inside the component.

Visit the entry in the Admin Panel.

Observe that:

The relation field inside the component appears empty.

The dropdown is not populated.

The relation exists and is returned properly via the API.

### Expected Behavior

The relation field inside the component should display the selected value(s) in the admin panel UI after being updated via API, just as it does when edited manually.

### Logs

```shell

```

### Media

![Image](https://github.com/user-attachments/assets/8292431b-9fdb-4622-bddf-4c15f615cc81)

![Image](https://github.com/user-attachments/assets/dc69148b-a9f5-4d89-a246-2b10e69f84d7)

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

reynan-dev:
I’m very interested in this issue.

theodoresmotly:
This issue is still happening in Strapi v5.18.1

Daeik:
The issue happens in v5.13 as well, I've tried to update the Strapi version to the latest version but it didn't help solving the problem.

mathildeleg:
Hi @xrgouda, thanks for opening this issue — I’d like to help clarify and reproduce, but I need a few more details:

**API type:**
- Are you using the REST API (POST /api/...) or GraphQL?

**Request body / mutation:**
- Can you share the full JSON payload or GraphQL mutation you’re sending?
- How exactly are you setting the component and its relational field?

**Populate / response:**
- Are you using ?populate=… in REST or requesting populated fields in GraphQL?
- What does the response look like?

**Database verification:**
[...]

PR #25271, opened 2026-01-27
fix: wip update relation filtering within compos

jhoward1994:
WIP

Ref https://github.com/strapi/strapi/issues/23512

### DIFF
--- a/packages/core/content-manager/server/src/controllers/relations.ts
+++ b/packages/core/content-manager/server/src/controllers/relations.ts
@@ -454,7 +454,8 @@ export default {
       publishedAt?: Record<string, any>;
     } = {};
 
-    if (sourceSchema?.options?.draftAndPublish || sourceSchema?.modelType === 'component') {
+    if (sourceSchema?.options?.draftAndPublish) {
+      // Content types with D&P: filter relations by the current document's status
       if (targetSchema?.options?.draftAndPublish) {
         if (status === 'published') {
           filters.publishedAt = { $notNull: true };
@@ -462,6 +463,10 @@ export default {
           filters.publishedAt = { $null: true };
         }
       }
+    } else if (sourceSchema?.modelType === 'component') {
+      // Components: do NOT filter by status - show relations regardless of their
+      // publication state. This ensures relations set via API are always visible.
+      // The relation's status will still be shown in the UI for transparency.
     } else if (targetSchema?.options?.draftAndPublish) {
       // NOTE: we must return the drafts as some targets might not have a published version yet
       filters.publishedAt = { $null: true };


jhoward1994:
Draft & Publish status filtering for relations.

For content types with D&P enabled, we filter relations to match the current document's publication status (draft or published).

For components, we could intentionally NOT apply strict status filtering. (as in these changes)

If we filter strictly by the parent document's status, relations set via API to published entities would not display when viewing the parent in draft mode.

I think what could be happening is that the relations within components are being set via numeric ids not document ids (by the users in the GH issue) & therefore our status filtering logic is "breaking" the UI for them.

This seems like intentional behaviour to me and not a bug?

See: https://github.com/strapi/strapi/issues/23512

@mathildeleg wdyt? (I saw you were active on the issue before)

mathildeleg:
> Draft & Publish status filtering for relations.
> 
> For content types with D&P enabled, we filter relations to match the current document's publication status (draft or published).
> 
> For components, we could intentionally NOT apply strict status filtering. (as in these changes)
> 
> If we filter strictly by the parent document's status, relations set via API to published entities would not display when viewing the parent in draft mode.
> 
> I think what could be happening is that the relations within components are being set via numeric ids not document ids (by the users in the GH issue) & therefore our status filtering logic is "breaking" the UI for them.
> 
> This seems like intentional behaviour to me and not a bug?
> 
> See: #23512
> 
> @mathildeleg wdyt? (I saw you were active on the issue before)

I would agree with you, it does seem like this is intended with D&P: the numeric id points to a specific version row (draft vs published), so setting relations via numeric ids can accidentally link the published row. The Content Manager then filters relations to match the version we’re viewing (draft tab shows draft relations), which would explain why it looks “missing” until we switch tabs. The docs also only mention the documentIds when wanting to create/update a relation in a document https://docs.strapi.io/cms/api/rest/relations#set so that seems right!

jhoward1994:
Okay thanks, I'll test things out before and after this change to confirm the behaviour. & we can decide whether there's any fix we can apply or whether this is working as intended.

jhoward1994:
I recreated the issue using a script that creates content, connecting relations via numeric IDs (the buggy behaviour reported in the issue) and one via document IDs (intended API usage, and working as expected).

I was able to reproduce the behaviour reported in https://github.com/strapi/strapi/issues/23512 in the numeric ID case.

imo we should communicate on the issue about the intended API usage (with document IDs) and that connecting relations via numeric IDs can cause issues with the draft and publish system.

To me this doesn't seem like a bug and if we were to change the API behaviour for numeric IDs it could break things for other users

wdyt @mathildeleg ?

mathildeleg:
> I recreated the issue using a script that creates content, connecting relations via numeric IDs (the buggy behaviour reported in the issue) and one via document IDs (intended API usage, and working as expected).
> 
> I was able to reproduce the behaviour reported in #23512 in the numeric ID case.
> 
> imo we should communicate on the issue about the intended API usage (with document IDs) and that connecting relations via numeric IDs can cause issues with the draft and publish system.
> 
> To me this doesn't seem like a bug and if we were to change the API behaviour for numeric IDs it could break things for other users
> 
> wdyt @mathildeleg ?

I reproduced it the same way and agree that it's probably just a matter of better communicating the intended API usage!

jhoward1994:
Closing this as I've closed the original issue after this investigation

https://github.com/strapi/strapi/issues/23512#issuecomment-3823241248

PR #25273, opened 2026-01-28
fix(blocks-editor): support multi-block list conversion

### LINKED ISSUE #25189: Unable to format highlighted text to list styles.
### Node Version

20.19.0

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.31.2

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

When you have multiple text lines in the blocks rich text editor, highlighting them and attempting to format them to a bulleted or numbered list does not work.

### Steps to Reproduce

Create a collection type with a field that is a rich text blocks field<br>Add a few lines to this in the content manager. No need to save.<br>Highlight those lines and click either bulleted list or numbered list

### Expected Behavior

Ideally, I should be able to format them to numbered or bulleted list.

### Logs

```shell
```

### Code Snippets

*No response*

### Media

*No response*

### Additional information

*No response*

### Confirmation Checklist

akshaywritescode:
### What does it do?
This pr fixes the problem of the multiblock bullet/list format

### Why is it needed?
When you have multiple text lines in the blocks rich text editor, highlighting multiple blocks and attempting to format them to a bulleted or numbered list does not work(only works on a single block).

### How to test it?
1. Open the Content Manager.
2. Create or edit an entry with a Blocks editor field.
3. Add multiple paragraph or heading blocks.
4. Select multiple blocks (click and drag).
5. Click the Bulleted list or Numbered list toolbar button.
6. Verify that all selected blocks are converted into a single list with one list item per block.

## Before
 

https://github.com/user-attachments/assets/6717f459-8b70-4e14-871f-4e1793945d10

 ## After
 

https://github.com/user-attachments/assets/124149b3-8d7f-45b8-80e0-2d9c01acffdf

### Related issue
fixes #25189

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { Typography } from '@strapi/design-system';
 import { BulletList, NumberList } from '@strapi/icons';
 import { Schema } from '@strapi/types';
-import { type Text, Editor, Node, Transforms, Path, Element } from 'slate';
+import { type Text, Editor, Node, Transforms, Path, Element, Range } from 'slate';
 import { type RenderElementProps, ReactEditor } from 'slate-react';
 import { styled, type CSSProperties, css } from 'styled-components';
 
@@ -277,6 +277,21 @@ const handleEnterKeyOnList = (editor: Editor) => {
  * Common handler for converting a node to a list
  */
 const handleConvertToList = (editor: Editor, format: Block<'list'>['format']) => {
+  if (!editor.selection) return;
+
+  // if multiple block
+  if (!Range.isCollapsed(editor.selection)) {
+    Transforms.setNodes(editor, { type: 'list-item' } as Partial<Node>, {
+      at: editor.selection,
+      match: (node) => Element.isElement(node) && node.type !== 'list' && node.type !== 'list-item',
+    });
+
+    Transforms.wrapNodes(editor, { type: 'list', format, children: [] }, { at: editor.selection });
+
+    return;
+  }
+
+  //if single block
   const convertedPath = baseHandleConvert<Block<'list-item'>>(editor, { type: 'list-item' });
 
   if (!convertedPath) return;
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
@@ -9,6 +9,7 @@ import { styled, type CSSProperties, css } from 'styled-components';
 
 import { type BlocksStore } from '../BlocksEditor';
 import { baseHandleConvert } from '../utils/conversions';
+import { getAttributesToClear } from '../utils/conversions';
 import { type Block } from '../utils/types';
 
 const listStyle = css`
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/tests/List.test.tsx
```diff
@@ -1440,6 +1440,55 @@ describe('List', () => {
       },
     ];
 
+    //for multiple selected block
+    it('converts multiple selected blocks into a single list', () => {
```
mathildeleg: You've placed this test inside another one, which makes it fail. You should copy/paste it after L1476 (the end of the test `'converts a heading with a link to a list'`) :)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
mathildeleg: This `@TODO` and `isListDisabled` can be removed thanks to your implementation. You'll need to also update where the `isListDisabled` is called :)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
```diff
@@ -277,6 +277,21 @@ const handleEnterKeyOnList = (editor: Editor) => {
  * Common handler for converting a node to a list
  */
 const handleConvertToList = (editor: Editor, format: Block<'list'>['format']) => {
+  if (!editor.selection) return;
+
+  // if multiple block
+  if (!Range.isCollapsed(editor.selection)) {
+    Transforms.setNodes(editor, { type: 'list-item' } as Partial<Node>, {
```
mathildeleg: The `Transforms.setNodes` call only sets `{ type: 'list-item' }`, it doesn't clear block-specific attributes like heading level. The single-block path avoids this because `baseHandleConvert` calls `getAttributesToClear` on the node before setting the new type. Without the same treatment here, a heading converted via multi-block selection ends up as a list-item that still carries `level: ...` which is invalid and inconsistent with single-block behaviour.
Example:
https://github.com/user-attachments/assets/17e0086f-4815-4dd7-ab92-630378ec9eba
Image

mathildeleg (commented):
Thanks for your PR ✨
I do have a few comments for improvements but overall looking good!

akshaywritescode:
@mathildeleg thank you, i will start working on this, sorry for the late response, i was out.

akshaywritescode:
@mathildeleg Thanks for the feedback! I've updated the multi-block conversion to clear block-specific attributes using getAttributesToClear, aligning it with the single-block behavior. Also verified it in the admin UI across multiple scenarios.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
```diff
@@ -9,6 +9,7 @@ import { styled, type CSSProperties, css } from 'styled-components';
 
 import { type BlocksStore } from '../BlocksEditor';
 import { baseHandleConvert } from '../utils/conversions';
+import { getAttributesToClear } from '../utils/conversions';
```
mathildeleg, line 12: You can combine this import with the one above (l11): 
```
import { baseHandleConvert, getAttributesToClear } from '../utils/conversions';
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/BlocksToolbar.tsx
```diff
@@ -461,7 +428,7 @@ const ListButton = ({ block, format, location = 'toolbar' }: ListButtonProps) =>
       name={format}
       label={block.label}
       isActive={isListActive()}
-      disabled={isListDisabled()}
+      disabled={disabled && !editor.selection}
```
mathildeleg, line 431: A selection always exists after a user clicks anywhere in the editor: even in a read-only/disabled editor, Slate still sets `editor.selection` on focus. Thus `disabled && !editor.selection` almost always evaluates to `false` when disabled is `true`, which means the toolbar list button never appears disabled, even when the whole editor is in read-only mode. I think you can just write:
`disabled={disabled}`
You can see here in the Published tab, if clicking on a line several times, it enables the list-block icons when it shouldn't be able to do so:
https://github.com/user-attachments/assets/0d492eda-fd61-4d5d-b4fa-d7f1ec8601c9

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
```diff
@@ -274,19 +275,66 @@ const handleEnterKeyOnList = (editor: Editor) => {
...
-  if (!Range.isCollapsed(editor.selection)) {
-    Transforms.setNodes(editor, { type: 'list-item' } as Partial<Node>, {
-      at: editor.selection,
-      match: (node) => Element.isElement(node) && node.type !== 'list' && node.type !== 'list-item',
-    });
+  const selection = editor.selection;
+  if (!selection) return;
+
+  // multi block selection
+  if (!Range.isCollapsed(selection)) {
```
mathildeleg, line 307: Unfortunately as you can see here, there are now issues depending on which element you're selecting and rendering as a multi-line list:
https://github.com/user-attachments/assets/3426423e-6688-459d-99b9-784ba46a7721

mathildeleg (commented):
Thanks for answering my feedback! Unfortunately there's still a few issues to fix (two are minor, one is quite important) 🙏🏻

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/List.tsx
akshaywritescode: thank you @mathildeleg , i fixed those other two, but i am confused in this one, as per the video i think you are saying that if a block is already in listed form(bullet or numbering), and we select it with another normal block, the listed conversion option should be disabled?, or the listed one should stay as it is and normal block get converted to listed form, or maybe anything else you are trying to say here.

sorry for this many follow ups, & and thank you for the fast response.

mathildeleg: Sorry this wasn't clear :) Here's what I meant:
1. On the video from second 01 to second 06: the issue isn't that a block already in list form creates another list, it's about the drag icon being present twice on the same line:
Capture d’écran 2026-05-07 à 09 49 10
2. On the video from second 06 to second 13: the issue is that when clicking on the multi-line icon, it reverts all element to a non-list yet we still see the dot, which also breaks with where the drag icon is positioned -> we shouldn't see the dot if we convert a multi-line list back to a paragraph:
Capture d’écran 2026-05-07 à 09 50 40
3. From second 19 to end of video: when highlighting/selecting the multi-line list, the dots disappear for some reason? I think it may be a glitch due to the fact the multi-line may actually be a paragraph in the db, because the first line was reverted back to a paragraph

I hope this is clearer on what the issues are 🙏🏻

akshaywritescode: Thank you for clearing this up, i started workin on it.

mathildeleg:
Hey there, just wondering if you're still working on this? 😊

mathildeleg:
Hey, I'll close this one for now, feel free to re-open when you have time to work on it again 🙂

PR #25277, opened 2026-01-28
fix: content-manager ui is very slow because of over population by localization field (#24504)

faessler:
### What does it do?

Adds a new method, `limitLocalizationsPopulate`, to the `populate-builder` service.
This method is now used in the `find` and `findOne` controllers for the content-manager admin routes.

In the future we would also need to cleanup the `meta` data. But to keep changes simple in this PR I've skipped that part.

### Why is it needed?

Strapi projects with large amounts of data and **multiple locales** often experience **very slow loading times** in collection views. This update ensures that only the **necessary fields** of the localizations attribute are loaded, which significantly improves page load performance.

### How to test it?

https://github.com/user-attachments/assets/40bcf1b8-2b52-48d6-80ee-11420029dcb9

### Related issue(s)/PR(s)

- https://github.com/strapi/strapi/issues/24504
- https://github.com/strapi/strapi/issues/24982
- https://github.com/strapi/strapi/pull/25279

Issue #24982, opened 2025-12-04
Strapi Admin performance issues


remidej:
The feedback from the community has been clear: the admin panel is slow. We have heard this message, and fixing it will be our priority in the next few months.

There are different problems that add up to the current situation. I'm opening this umbrella issue so we can manage each problem individually, and so that we can track our progress as we tackle them one at a time.

Sub-issues to create:

- [x] blocks editor
- [x] input lag
- [ ] main bundle size is too big
- [ ] request waterfalls

laurenskling:
How clunky components collapse/open and modals (relation-on-the-fly) open particularly annoys me

yunusyuksel:
We have also a big performance issue on homewidgets. 
Here are my response times. I have around 600mb data in my database. I guess counting the documents, querying the recent documents is an issue 
<img width="1041" height="291" alt="Image" src="https://github.com/user-attachments/assets/a2b0f599-9d04-472e-ab74-1381ae7fa840" />

crispyyv:
Same problem with latest version. It looks like that it also affect API requests. Because of on Strapi 5.30 we have average requests time ~ 0.5s when with strapi 5.36 - 3.5s.

Issue #24504, opened 2025-10-03
Content Manager over-populates localizations and metadata causing large payloads and slow admin UI


lsta900:
### Node Version

v22.12.0

### NPM/Yarn/PNPM Version

NPM 10.9.0

### Strapi Version

5.24.1

### Operating System

Strapi Cloud

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Summary
The Content Manager endpoints over-populate the `localizations` relation and include heavy `meta.availableLocales` by default. This produces large payloads and slow admin UI, especially with many locales and deep relations. Limiting the populated localization fields and disabling `availableLocales` where not needed yields substantial improvements.

### Steps to Reproduce

1. Use a localized collection type with dynamic zones/components and many entries. Ensure `i18n` and Draft & Publish are enabled. Create ~20 different documents with lots of content and translations
2. In the CM:
   - List view: Loads very slowly and the payload of the API response can be multiple MB's. For each document it contains all the translations with all the content in them which leads to the large payload.
   - Single entry view: Documents contain deeply nested data for translated versions of the document and `meta.availableLocales` object again with deeply nested data. This is most likely not needed

### Actual Behavior (example from our production data)
- List view payloads: ~3 MB
- Single item payloads: ~700 KB
- Admin UI loads slowly.
- With DB debug logging enabled, we see about 700 db queries to load the List View with 10 documents

### Expected Behavior

- Admin responses should not deep-populate `localizations` by default.
- `availableLocales` should be opt-in or minimal, and only included when explicitly requested by the client (e.g., via a query parameter), not in every entry fetch.
- probably other localization specific improvements which don't fetch translations unless needed

### Logs

```shell

```

### Proposed Improvements
1. Default populate rules for `localizations` in CM endpoints should return minimal identifiers (`id`, `locale`, `documentId`) unless deeper data is explicitly needed/requested.
2. Make `meta.availableLocales` opt-in (e.g., controlled by a query parameter) and avoid computing/sending it by default.
3. Consider a configuration option in CM to control localization population depth and metadata inclusion if there's a use case for it.

### Results (in case the proposal is implemented):
- List view: 3 MB → ~100 KB
- Single item: 700 KB → ~100 KB
- Noticeably faster CM load and potentially persist times. -> a lot happier customers since they don't need to wait

### Notes
- We can share a minimal reproduction or diffs illustrating the controller-level changes that produced these improvements.
[...]

PR #25279, opened 2026-01-28
fix: content-manager ui is very slow because of over population by meta field (#24504)

faessler:
### What does it do?

Returns empty array for `availableLocales` on `meta` data for `find` and `findOne`.

### Why is it needed?

Strapi projects with large amounts of data and **multiple locales** often experience **very slow loading times** in collection views. This update ensures that only the necessary fields of the `meta` attribute are loaded, which significantly improves page load performance.

### How to test it?

Same as on https://github.com/strapi/strapi/pull/25277.

### Related issue(s)/PR(s)
- https://github.com/strapi/strapi/issues/24504
- https://github.com/strapi/strapi/issues/24982
- https://github.com/strapi/strapi/pull/25277

### DIFF
--- a/packages/core/content-manager/server/src/controllers/collection-types.ts
+++ b/packages/core/content-manager/server/src/controllers/collection-types.ts
@@ -212,7 +212,7 @@ export default {
         model,
         // @ts-expect-error TODO: fix
         { documentId: id, locale, publishedAt: null },
-        { availableLocales: true, availableStatus: false }
+        { availableLocales: false, availableStatus: false }
       );
 
       ctx.body = { data: {}, meta };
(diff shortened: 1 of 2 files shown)


derrickmehaffy:
PR Submitted by customer, see Internal Ticket 6259

jhoward1994:
In my testing I don't think this is something we can do as it will break i18n functionality in the CM

One example I've been able to replicate is the "Fill from another locale" button. Even when other locales are filled, that button will always be disabled with these changes as it depends on data from `availableLocales`

mathildeleg:
As @jhoward1994 said, we've found this breaks some core functionalities so we'll close this PR, but thank you for your contribution, and know that we're currently working on improving the performance of the admin panel so hopefully we'll have a fix/fixes coming out soon that will help with your issue too :)

PR #25286, opened 2026-01-28
chore(*/vitest): introduce Vitest and vitest-config package

nclsndr:
### What does it do?

- **Adds `vitest-config`** in `packages/utils/vitest-config/` as a private shared Vitest config:
  - Unit preset with `globals: false` (explicit `describe`/`it`/`expect`/`vi` imports), `environment: 'node'`, `include: ['**/*.vitest.test.ts']`, `testTimeout: 5000`, and exclusions aligned with the Jest unit setup.

- **Integrates Vitest in `packages/core/core/`**:
  - `vitest.config.ts` that merges the unit preset with package-specific.
  - Scripts: `test:unit:vitest` (run once) and `test:unit:vitest:watch` (watch).
  - Dev deps: `vitest`, `vitest-config`.

- **Root-level scripts** (in root `package.json`):
  - `test:unit:vitest` and `test:unit:vitest:watch` that run Vitest via `packages/core/core/vitest.config.ts`.

- **Centralized Vitest version**:
  - Root `package.json`: `vitest` in `devDependencies`, and `vitest: "4.0.18"` in `resolutions` so the whole monorepo uses the same version.

- **CI** (`.github/workflows/tests.yml`):
  - New job `unit_back_vitest` (Node 20/22/24), same conditions as `unit_back`, runs `yarn test:unit:vitest`.

- **Docs**:
  - `VITEST_MIGRATION.md` with setup overview, migration steps (Jest → Vitest, `*.vitest.test.ts`), and dependency/version strategy.

### Why is it needed?

- Modernize the test stack by introducing Vitest (faster, ESM-friendly, Vite-based) alongside Jest.
- Allow incremental migration via the `*.vitest.test.ts` pattern so existing `*.test.ts` Jest tests keep running while new or migrated tests use Vitest.
- Pin a single Vitest version (4.0.18) via resolutions so every package uses the same runner and upgrades are done in one place.

### How to test it?

1. **Install and run Vitest**  
   From repo root:
   ```bash
   yarn install
   yarn test:unit:vitest
   ```
   Expected: Vitest runs and exits successfully

2. **Run from core package**  
   ```bash
   cd packages/core/core && yarn test:unit:vitest
   ```
   Expected: Vitest runs and exits successfully

3. **Watch mode**  
   ```bash
   yarn test:unit:vitest:watch
   ```
   And
   ```bash
   cd packages/core/core && yarn test:unit:vitest:watch
   ```
   Confirm Vitest starts in watch mode (Ctrl+C to stop).

5. **Vitest version**  
   ```bash
   yarn vitest --version
   ```
   Expected: `vitest/4.0.18 …`.

6. **Jest still works**  
   ```bash
   yarn test:unit
   ```
   Existing Jest unit tests should run unchanged.

**Environment:** Node 20–24, macOS/Linux. No extra env vars needed for the above.

### Related issue(s)/PR(s)

None.


packages/core/core/src/utils/__tests__/convert-custom-field-type.vitest.test.ts
nclsndr, line 1: Example files migrated to prove the scripts and workflow

package.json
```diff
@@ -81,7 +81,9 @@
     "test:unit": "jest --config jest.config.js",
     "test:unit:update": "jest --config jest.config.js -u",
     "test:unit:all": "nx run-many --target=test:unit --nx-ignore-cycles",
-    "test:unit:watch": "run test:unit --watch"
+    "test:unit:watch": "run test:unit --watch",
+    "test:unit:vitest": "vitest run --config packages/core/core/vitest.config.ts",
```
Bassel17: I think it would make sense if we do something like,  what we do with jest have a global vite config that points to the different configs in packages, just prepare for global vitests and maybe add one more package as an example.

Bassel17 (commented):
looks good, I just have a few comments, if we want to migrate to vitest, it would make sense to make vitest the default `test:unit` would run vitest tests, and a `test:unit:jest` or `test:unit:legacy` ?

package.json
innerdvations: To me it makes sense to have test:unit be a parent for test:unit:jest and test:unit:vitest that way we don't have to change any tooling, but if there are complications there I'm fine either way, I'm a lot less picky about tooling issues since we can change them whenever we want without affecting anyone else 😆

Bassel17: true that as well

Bassel17: it just didn't make sense for use to have vitest as the optional when we are migrating to it

nclsndr: Thank you for the feedback @Bassel17 & @innerdvations.
I'll make another iteration to define a monorepo level config for vitest runner.

> have test:unit be a parent for test:unit:jest and test:unit:vitest

I didn't go down this path to segregate errors on CI: if we ever have issues affecting one or the other, being able to spot the runner by the action step rather than opening logs. I might have been too precautious on this. If you both think we should have both runner launched by a single package.json script: I'm also willing to make this happen.

innerdvations: You can go ahead and merge this and then check that the workflows are running correctly on new PRs and if so immediately make the switch to make it required

nclsndr: > if so immediately make the switch to make it required

Not sure I get your point @innerdvations. The `unit_back_vitest` job is already part of the CI requirements in `.github/workflows/tests.yml`. What am I missing? 🤔

markkaylor (dismissed):
Looks good to me and works as expected!

package.json
innerdvations: Ahh, ok, I missed that they were already running here, ignore me 😆

innerdvations (approved):

nclsndr:
I had a conflict to manage. Sorry for the spam @innerdvations

PR #25293, opened 2026-01-29
fix(migrations): speed up discard-drafts with bulk batches [CMS-29]

Bassel17:
### What does it do?

Reworks the discard-drafts migration to mirror discardDraft behavior in SQL batches: clone scalar drafts, bulk copy join-table relations, remap foreign keys to draft targets, and clone components with their relations.

### Why is it needed?

The previous migration executed discardDraft per entry with no concurrency, which is slow on large datasets and scales poorly.

### How to test it?

- From examples/complex:
```yarn setup:v4```
- In the v4 project (path printed by setup):
```yarn seed```

- Back in examples/complex:
```yarn test:migration```

check the complex/readme for more details on the setup


innerdvations:
>  I keep getting Transaction was implicitly committed, do not mix transactions and DDL with MySQL (#805) from the log

This is warning that in mysql it's not actually using a transaction for the whole migration and commits at various points;  it's always been that way, but we should investigate if we can either clean up the warnings or solve the problem in some way

PR #25297, opened 2026-01-29
future: implement assetlist component table view

jhoward1994:
## What does it do?

Implements the list view for the future media library

## Why is it needed?

Adds a functional list view to the future media library

**Scope:**  
- Only displays a list of assets  
- Folders, drag and drop, selection/multiselect, and CRUD operations are handled by other tasks  

## How to test

1. Enable the future media library flag.
2. Navigate to the Media Library.
3. Verify that clicking "Table view" switches to the list view

## Related issue(s) / PR(s)

fixes CMS-110
fixes CMS-116

### DIFF
--- a/packages/providers/upload-aws-s3/src/__tests__/upload-aws-s3.test.ts
+++ b/packages/providers/upload-aws-s3/src/__tests__/upload-aws-s3.test.ts
@@ -9,12 +9,7 @@ jest.mock('../utils', () => ({
 }));
 
 const uploadMock = {
-  done: jest.fn().mockImplementation(() =>
-    Promise.resolve({
-      Location: 'https://validurl.test/tmp/test.json',
-      $metadata: {},
-    })
-  ),
+  done: jest.fn(),
 };
 
 jest.mock('@aws-sdk/lib-storage', () => ({
@@ -27,21 +22,30 @@ describe('AWS-S3 provider', () => {
   });
 
   describe('upload', () => {
-    test('Should add url to file object', async () => {
+    test('Should populate file.url with the baseUrl when provided, appending the file key to the url', async () => {
+      uploadMock.done.mockImplementationOnce(() =>
+        Promise.resolve({
+          Location: 'https://strapi.io/path/to/fileNameHash.json',
+          $metadata: {},
+        })
+      );
+
       const providerInstance = awsProvider.init({
+        baseUrl: 'https://assets.strapi.io',
         s3Options: {
           params: {
-            Bucket: 'test',
+            Bucket: 'bucket-name',
           },
         },
       });
 
       const file: File = {
-        name: 'test',
+        name: 'fileName',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
-        path: 'tmp',
-        hash: 'test',
+        path: 'path/to',
+        hash: 'fileNameHash',
         ext: '.json',
         mime: 'application/json',
         buffer: Buffer.from(''),
@@ -51,24 +55,31 @@ describe('AWS-S3 provider', () => {
 
       expect(uploadMock.done).toBeCalled();
       expect(file.url).toBeDefined();
-      expect(file.url).toEqual('https://validurl.test/tmp/test.json');
+      expect(file.url).toEqual('https://assets.strapi.io/path/to/fileNameHash.json');
     });
+    test('Should populate file.url with the returned Location when no baseUrl is provided', async () => {
+      uploadMock.done.mockImplementationOnce(() =>
+        Promise.resolve({
+          Location: 'https://strapi.io/path/from/location/fileName.json',
+          $metadata: {},
+        })
+      );
 
-    test('Should add to the url the https protocol as it is missing', async () => {
       const providerInstance = awsProvider.init({
         s3Options: {
           params: {
-            Bucket: 'test',
+            Bucket: 'bucket-name',
           },
         },
       });
 
       const file: File = {
-        name: 'test',
+        name: 'fileName',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
-        path: 'tmp',
-        hash: 'test',
+        path: 'path/from/location',
+        hash: 'fileNameHash',
         ext: '.json',
         mime: 'application/json',
         buffer: Buffer.from(''),
@@ -78,10 +89,50 @@ describe('AWS-S3 provider', () => {
 
       expect(uploadMock.done).toBeCalled();
       expect(file.url).toBeDefined();
-      expect(file.url).toEqual('https://validurl.test/tmp/test.json');
+      expect(file.url).toEqual('https://strapi.io/path/from/location/fileName.json');
     });
+    test('Should populate file.url and prepend the https protocol to the Location when missing', async () => {
+      uploadMock.done.mockImplementationOnce(() =>
+        Promise.resolve({
+          Location: 'strapi.io/path/to/fileNameHash.json',
+          $metadata: {},
+        })
+      );
 
-    test('Should prepend the baseUrl to the url of the file object', async () => {
+      const providerInstance = awsProvider.init({
+        s3Options: {
+          params: {
+            Bucket: 'bucket-name',
+          },
+        },
+      });
+
+      const file: File = {
+        name: 'fileName',
+        size: 100,
+        sizeInBytes: 100 * 1024,
+        url: '',
+        path: 'path/to',
+        hash: 'fileNameHash',
+        ext: '.json',
+        mime: 'application/json',
+        buffer: Buffer.from(''),
+      };
+
+      await providerInstance.upload(file);
+
+      expect(uploadMock.done).toBeCalled();
+      expect(file.url).toBeDefined();
+      expect(file.url).toEqual('https://strapi.io/path/to/fileNameHash.json');
+    });
+    test('Should populate file.url with baseUrl even if location lacks protocol', async () => {
+      uploadMock.done.mockImplementationOnce(() =>
+        Promise.resolve({
+          Location: 'otherdomain.com/different/path/file.json',
+          $metadata: {},
+        })
+      );
+
       const providerInstance = awsProvider.init({
         baseUrl: 'https://cdn.test',
         s3Options: {
@@ -95,6 +146,7 @@ describe('AWS-S3 provider', () => {
       const file: File = {
         name: 'test',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
         path: 'tmp/test',
         hash: 'test',
@@ -109,8 +161,14 @@ describe('AWS-S3 provider', () => {
       expect(file.url).toBeDefined();
       expect(file.url).toEqual('https://cdn.test/tmp/test/test.json');
     });
+    test('Should populate file.url with baseUrl and rootPath', async () => {
+      uploadMock.done.mockImplementationOnce(() =>
+        Promise.resolve({
+          Location: 'https://validurl.test/tmp/test.json',
+          $metadata: {},
+        })
+      );
 
-    test('Should prepend the baseUrl and rootPath to the url of the file object', async () => {
       const providerInstance = awsProvider.init({
         baseUrl: 'https://cdn.test',
         rootPath: 'dir/dir2',
@@ -124,6 +182,7 @@ describe('AWS-S3 provider', () => {
       const file: File = {
         name: 'test',
         size: 100,
+        sizeInBytes: 100 * 1024,
         url: '',
         path: 'tmp/test',
         hash: 'test',
--- a/packages/providers/upload-aws-s3/src/index.ts
+++ b/packages/providers/upload-aws-s3/src/index.ts
@@ -63,7 +63,7 @@ export interface DefaultOptions extends S3ClientConfig {
 export type InitOptions = (DefaultOptions | { s3Options: DefaultOptions }) & {
   baseUrl?: string;
   rootPath?: string;
-  [k: string]: any;
+  [k: string]: unknown;
 };
 
 const assertUrlProtocol = (url: string) => {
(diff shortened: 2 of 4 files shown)


packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -31,12 +73,38 @@ export const MediaLibraryPage = () => {
...
+              </ViewOption>
+              <ViewOption
+                $active={isGridView}
+                onClick={() => setView(viewOptions.GRID)}
+                aria-pressed={isGridView}
+              >
+                <GridIcon />
+                {formatMessage({ id: getTrad('view.grid'), defaultMessage: 'Grid view' })}
+              </ViewOption>
+            </ViewSwitcher>
```
markkaylor: Nice I think this will close https://linear.app/strapi/issue/CMS-116/add-listgrid-view-tabs-to-header-toolbar right?

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -2,12 +2,54 @@ import * as React from 'react';
 
 import { Layouts, SearchInput } from '@strapi/admin/strapi-admin';
 import { Box, Flex } from '@strapi/design-system';
+import { GridFour as GridIcon, List } from '@strapi/icons';
 import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
 
+import { localStorageKeys, viewOptions } from '../../constants';
+
+const ViewSwitcher = styled(Flex)`
```
markkaylor: I'm mostly just wondering about a11y here. I think it should be a ToggleGroup or Tabs. Could you investigate and see on radix what would be best https://www.radix-ui.com/primitives/docs/components/switch or if you have reason for why it doesn't matter I'm all ears.

packages/core/upload/admin/src/future/components/AssetList/AssetTableRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </StyledTd>
+      <StyledTd>
+        <Flex justifyContent="flex-end">
+          <IconButton
+            label={formatMessage({
+              id: getTrad('control-card.more-actions'),
+              defaultMessage: 'More actions',
+            })}
+            variant="ghost"
+          >
```
markkaylor: On my PR I create a small new abstraction to deal with this. Could you give me your opinion over here if we should move in that direction: https://github.com/strapi/strapi/pull/25256#discussion_r2741891988

If you and Adrien agree I will merge asap so you have access to that.

If we decide not to do that then, I'd prefer to move getTrad to a u`utils/translation.ts` file in the future folder and rename the function `getTranslation`.

packages/core/upload/admin/src/future/components/AssetList/AssetList.tsx
markkaylor, line 1: I don't think this component is necessary and we should just do all of this directly in MediaLibraryPage

packages/core/upload/admin/src/future/components/AssetList/AssetTable.tsx
markkaylor, line 1: I'd prefer to not call it AssetTable but AssetList

packages/core/upload/admin/src/future/components/AssetList/AssetTableRow.tsx
markkaylor, line 1: I don't see a need for a separate file or maybe even component, fine with the component but I think this should live in AssetTable and not be exported

packages/core/upload/admin/src/future/components/AssetList/AssetPreviewCell.tsx
markkaylor, line 1: I don't see a need for a separate file I think this should live in AssetTable, with AssetTableRow and not be exported

packages/core/upload/admin/src/future/components/AssetList/AssetTable.tsx
markkaylor: Also, do you see this component being used anywhere other than MediaLibraryPage? If not what are your thoughts on moving this code closer to where it's used. I think to MediaLibraryPage doesn't make much sense anymore, maybe we should call it `AssetsPage`. We could have something like this for example:

```
/pages
-/Assets
--/components
---AssetsList.tsx
---AssetsGrid.tsx
--AssetsPage.tsx
--constants.ts
```

packages/core/upload/admin/src/future/components/AssetList/AssetTableRow.tsx
markkaylor: Heads up if you pull develop there, is `getTranslationKey` to use instead of getTrad otherwise all good with formatMessage

markkaylor:
@jhoward1994 if you pull develop now you'll have a way to upload via the filepicker so we can ensure the cache is invalidated and the list updates without a refresh. Also e2e tests are setup so you could add a test to test exactly that.

packages/core/upload/admin/src/future/pages/Assets/constants.ts
```diff
@@ -0,0 +1,50 @@
...
+  name: string;
+  label: { id: string; defaultMessage: string };
+  isSortable: boolean;
+  isVisuallyHidden?: boolean;
+}
+
+export const TABLE_HEADERS: TableHeader[] = [
+  {
+    name: 'name',
+    label: { id: getTrad('list.table.header.name'), defaultMessage: 'name' },
```
markkaylor: Still using getTrad here in this file

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
markkaylor, line 1: Could we rename this AssetsPage and move it inside pages/Assets?

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -1,20 +1,89 @@
...
+  SimpleMenu,
+  Typography,
+  VisuallyHidden,
+} from '@strapi/design-system';
+import { ChevronDown, Files, GridFour as GridIcon, List } from '@strapi/icons';
 import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
 
+import { localStorageKeys, viewOptions } from '../../constants';
+import { usePersistentState } from '../../hooks/usePersistentState';
```
markkaylor: Maybe we should consider exporting this from the admin instead of redefining it in this plugin. The admin has the exact hook here: https://github.com/strapi/strapi/blob/4c1658c529672380623d936350320d42ce5b7db1/packages/core/admin/admin/src/hooks/usePersistentState.ts#L5-L26

markkaylor: Instead of using the existing ones I'd prefer to copy them into the future folder at the end of the project I want to just be able to delete everything in src and move everything out of future into src, do you see what I mean? The constants here can go in the new constants folder you created in pages/Assets/constants

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -68,31 +137,107 @@ export const MediaLibraryPage = () => {
     e.target.value = '';
   };
 
+  /**
```
markkaylor: We can remove this comment, I thought I had removed it actually

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+const StyledBodyTd = styled(RawTd)`
+  padding: ${({ theme }) => theme.spaces[4]};
+  border-bottom: 1px solid ${({ theme }) => theme.colors.neutral150};
+`;
+
+interface AssetPreviewCellProps {
+  asset: File;
+}
+
+const AssetPreviewCell = ({ asset }: AssetPreviewCellProps) => {
```
markkaylor, line 76: The designs aren't showing a preview actually just an icon to representing the file type. I'm good either way but feel free to clarify with Anka if you'd like to challenge keeping the Preview here

Image

Image

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+                      >
+                        {tableHeaderLabel}
+                      </Typography>
+                    ) : (
+                      <Typography textColor="neutral600" variant="sigma">
+                        {tableHeaderLabel}
+                      </Typography>
+                    )}
+                  </Tooltip>
+                  {isSorted && (
```
markkaylor: Sorting will not be handled directly in the table so I think this can be removed

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+                  </VisuallyHidden>
+                </StyledTh>
+              );
+            }
+
+            return (
+              <StyledTh key={header.name}>
+                <Flex gap={1} alignItems="center">
+                  <Tooltip label={header.isSortable ? sortLabel : tableHeaderLabel}>
+                    {header.isSortable ? (
```
markkaylor: Sorting will not be handled in the table, it's another task as well

Image

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+  assets: File[];
+  sort?: SortState;
+  onSortChange?: (sort: SortState) => void;
+  onAssetClick?: (asset: File) => void;
+}
+
+export const AssetsList = ({ assets, sort, onSortChange, onAssetClick }: AssetsListProps) => {
+  const { formatMessage } = useIntl();
+
+  const handleSort = (fieldName: string) => {
```
markkaylor: I'd remove all notion of sort since it is not related to this task

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+interface AssetRowProps {
+  asset: File;
+  onClick?: (asset: File) => void;
+}
+
+const AssetRow = ({ asset, onClick }: AssetRowProps) => {
+  const { formatDate, formatMessage } = useIntl();
+
+  const handleRowClick = () => {
```
markkaylor: I would just remove an not anticipate the click interaction since it is not part of this task

packages/core/upload/admin/src/future/pages/Assets/components/AssetsList.tsx
```diff
@@ -0,0 +1,297 @@
...
+  FileZip,
+  Monitor,
+  More,
+  VolumeUp,
+} from '@strapi/icons';
+import { useIntl } from 'react-intl';
+import { styled } from 'styled-components';
+
+import { AssetType } from '../../../../enums';
+import { formatBytes, getFileExtension, prefixFileUrlWithBackendUrl } from '../../../../utils';
```
markkaylor: Same comment about importing outside future folder

packages/core/upload/admin/src/future/pages/MediaLibraryPage.tsx
```diff
@@ -68,31 +137,107 @@ export const MediaLibraryPage = () => {
...
+                defaultMessage: 'An error occurred while fetching assets.',
+              })}
+            </Typography>
+          </Box>
+        ) : isGridView ? (
+          <AssetsGrid assets={assets} />
+        ) : (
+          <AssetsList assets={assets} sort={sort} onSortChange={setSort} />
+        )}
+      </Layouts.Content>
```
markkaylor: Don't love this gonna take a look to explore other options for handling the loading error state

markkaylor: To clean this up,after you rename the file and component AssetsPage we could create another component in the same file: AssetsView, something like this:

```tsx
const AssetsView = ({ view }: { view: number }) => {

  // View/sort state
  const [sort, setSort] = React.useState<SortState>(DEFAULT_SORT);

  // Query
  const sortString = `${sort.field}:${sort.order}`;
  const { data, isLoading, error } = useGetAssetsQuery({ folder: null, sort: sortString });

  const isGridView = view === viewOptions.GRID;
  const assets = data?.results ?? [];

  if (isLoading) {
    return (
      <Flex justifyContent="center" padding={8}>
        <Loader>{formatMessage({ id: 'app.loading', defaultMessage: 'Loading...' })}</Loader>
      </Flex>
    );
  }

  if (error) {
    return (
      <Box padding={8}>
        <Typography textColor="danger600">
          {formatMessage({
            id: getTranslationKey('list.assets.error'),
            defaultMessage: 'An error occurred while fetching assets.',
          })}
        </Typography>
      </Box>
    );
  }

  if (isGridView) {
    return <AssetsGrid assets={assets} />;
  }

  return <AssetsList assets={assets} sort={sort} onSortChange={setSort} />;
};
```

jhoward1994:
Perfect thanks, yeah I had tests in mind to add. Will add the e2e one too and check the icon designs

packages/core/upload/admin/src/future/pages/Assets/tests/AssetsList.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+  render(
+    <MemoryRouter>
+      <IntlProvider locale="en" messages={{}}>
+        <DesignSystemProvider>
+          <AssetsList assets={assets} />
+        </DesignSystemProvider>
+      </IntlProvider>
+    </MemoryRouter>
+  );
+
```
markkaylor: I think you can use `import { render, screen } from '@tests/utils'` and this will all be set up for you

packages/core/upload/admin/src/future/pages/Assets/tests/AssetsList.test.tsx
markkaylor, line 1: Unit test is great but above all we want an e2e test that confirms the list view updates when a user uploads a file. You could add this as part of the upload test here: tests/e2e/tests/media-library/future/file-upload.spec.ts

jhoward1994: Isn't that done in https://github.com/strapi/strapi/pull/25297/changes/809fc87f23eb5298cfc08e59af9c969061f101b4 ?

markkaylor: Sure is! My bad dude

PR #25298, opened 2026-01-29
fix(types): update LoadedPlugin type to understand factories

Link2Twenty:
### What does it do?

I've added a Factory type that allows wrapping the existing `LoadedPlugin` it seems to have been missing for a while even strapi code has work arounds rather than wrapping it.

https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/history/controllers/index.ts#L4-L10

I looked at the [factories](https://github.com/strapi/strapi/blob/develop/packages/core/core/src/factories.ts) file to see which `LoadedPlugin` props accepted factories.

### Why is it needed?

To make the types match what is possible, this will make plugin creation much easier (I intend to type the plugin sdk template once this is in)

### How to test it?

N/A

### Related issue(s)/PR(s)

N/A


innerdvations (approved):

innerdvations:
Thanks for the fix!

PR #25300, opened 2026-01-30
fix(content-manager): fix frontend validation if not using "draft and publish"

pksr:
### What does it do?

Frontend validation does not work correctly if "draft and publish" is not used. The validation for "draft", which is always applied even if "draft and publish" is not used, is significantly less strict than the validation for the "publish" action.

### Why is it needed?

Fixes the frontend validation.

### How to test it?

1. Create a collection without "draft and publish".
2. Add a dynamic zone having a required text field.
3. Create an entry for that collection.
4. Add the component to the dynamic zone but leave the required text field empty.
5. Click "Save".
6. **Without the fix**: we get the error message from the server, caused by the document service, what causes all other changed fields to get reset to their initial value, **with the fix**: we get the error message from the frontend validation.

### Related issue(s)/PR(s)

-

Without fix:
without_fix

With fix:
with_fix


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -1090,6 +1090,9 @@ const UpdateAction: DocumentActionComponent = ({
   );
   const { schema } = useDoc();
 
+  const suitableSchema = fromRelationModal ? relationalModalSchema : schema
```
mathildeleg:
```suggestion
  const suitableSchema = fromRelationModal ? relationalModalSchema : schema;
```
Missing a semi-colon there :)

mathildeleg (commented):
Looks good to me, just a minor typo to fix, and could you also maybe add/update tests for this? Thanks 🙏🏻

mathildeleg:
Hey there, it's been a while, could you either update your PR to fix the typo and resolve conflicts or give me access to push changes directly to your branch? Otherwise I'll create a PR of my own crediting you so that we can get this fix live ☺️ thanks!

pksr:
Hi @mathildeleg, I fixed the typo and the conflicts.

mathildeleg (approved):
Thanks for the fixes! :)

PR #25303, opened 2026-01-30
test(api): use jwt default options instead of expiresIn

innerdvations:
### What does it do?

- switches our API tests to use the new JWT config option defaults instead of `expiresIn`
- adds test for legacy expiresIn to make sure it keeps working

### Why is it needed?

almost every API test spits out multiple warnings about deprecated values, polluting the test logs

### How to test it?

run API tests, look at the output

### Related issue(s)/PR(s)

### DIFF
--- a/packages/utils/api-tests/strapi.js
+++ b/packages/utils/api-tests/strapi.js
@@ -41,11 +41,10 @@ const createStrapiInstance = async ({
   // Ensure Koa trusts X-Forwarded-* headers in tests so asHTTPS() can simulate HTTPS
   instance.config.set('server.proxy.koa', true);
 
-  // Opt out of deprecated expiresIn so only the new session config (and its defaults) is used
+  // Opt out of deprecated expiresIn by setting an empty session options object
+  // This prevents the default expiresIn from triggering deprecation warnings
+  // Only set if not already configured to avoid overriding user settings
   if (!skipDefaultSessionConfig) {
-    // Opt out of deprecated expiresIn by setting an empty session options object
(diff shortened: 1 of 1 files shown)


jhoward1994 (dismissed):

Bassel17:
you probably need to run prettier for formatting, otherwise it looks good !

innerdvations:
> you probably need to run prettier for formatting, otherwise it looks good !

that's what I get for accepting a code change from claude 😂

packages/utils/api-tests/strapi.js
```diff
@@ -39,6 +41,19 @@ const createStrapiInstance = async ({
   // Ensure Koa trusts X-Forwarded-* headers in tests so asHTTPS() can simulate HTTPS
   instance.config.set('server.proxy.koa', true);
 
+  // Opt out of deprecated expiresIn so only the new session config (and its defaults) is used
+  if (!skipDefaultSessionConfig) {
+    // Opt out of deprecated expiresIn by setting an empty session options object
+    // This prevents the default expiresIn from triggering deprecation warnings
+    if (!skipDefaultSessionConfig) {
```
nclsndr: Twice the same condition 🤔

nclsndr (changes requested):
Just a detail 👇 
Otherwise looks good, I'm trusting the CI for the test run check

packages/utils/api-tests/strapi.js
innerdvations: well that's embarrassing

Issue #25136, opened 2026-01-07
Admin input lag persists in v5.33.1 for content-types with relations (typing delay)


m-jadczak:
### Node Version

22.21.1

### NPM/Yarn/PNPM Version

NPM 10.9.4

### Strapi Version

5.33.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

*No response*

### Bug Description

The Strapi Admin UI still has noticeable input lag (typing delay) on Strapi v5.33.1.

While typing into standard text inputs / textareas in the Content Manager, characters appear with a delay (“catching up” after a moment), which significantly impacts usability.

In our case this happens for every content-type (model) that contains at least one relation field. The lag is reproducible even when typing in a non-relation text field on the same edit view.

The issue persists even after:

* disabling all plugins
* uninstalling additional plugin-related packages (trying to keep the setup as “clean” as possible)

For comparison, I recorded the same workflow on Strapi v4.25.2 where typing feels much smoother.

### Steps to Reproduce

1. Start Strapi and open the Admin panel.
2. Go to Content Manager.
3. Open any content-type entry that has at least one relation field (or create a new entry).
4. Focus a normal text field (input or textarea) — it does not have to be the relation field itself.
5. Type quickly for a few seconds.

Observed: the input lags (keystrokes are rendered with a delay).

### Expected Behavior

Typing in Admin inputs should be smooth and responsive, with no noticeable delay between keystrokes and rendered text.

The presence of relation fields in a content-type should not degrade typing performance in other fields on the edit view.

### Logs

```shell
```

### Code Snippets

*No response*

### Media

I will attach two screen recordings:

1. Strapi v5.33.1 — typing lag in Admin inputs<br>Attachment:

https://github.com/user-attachments/assets/39b80b8e-72d5-4faa-aab8-19bfe5719f10

2. Strapi v4.25.2 — comparison (same typing workflow)<br>Attachment:

https://github.com/user-attachments/assets/f89c11e6-2a44-4182-bf52-5017626c4791

### Additional information

Notes / scope:

* The lag occurs for us on every model that contains a relation field.
* It happens even when typing into a regular text input on the same edit view.
* Tested after disabling all plugins and removing extra plugin packages — still reproducible.

Possible duplicates / related issues:

[...]

PR #25304, opened 2026-01-30
fix: improve input lag

gitjs:
### What does it do?

Improves typing performance in large forms by:
- Wrapping field updates in React.startTransition so form state updates don’t block keystrokes.
- Adding a local state inside useField to keep inputs responsive while the global form state updates.

### Why is it needed?

The larger the form in Strapi, the more noticeable input lag can become.
This change does not aim to solve the underlying performance issues, but provides a pragmatic quick win so typing in large forms feels more responsive.

### How to test it?

Create a large form (e.g. many fields or components, especially many expanded components in a dynamic zone) and type into inputs. Typing should feel noticeably more responsive than before.

### Related issue(s)/PR(s)

- Related to: https://github.com/strapi/strapi/issues/25136

### DIFF
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -148,6 +148,40 @@ interface FormProps<TFormValues extends FormValues = FormValues>
   validate?: (values: TFormValues, options: Record<string, string>) => Promise<any>;
 }
 
+function parseInputValueFromEventTarget(target: any, stateValues: any, field: string) {
+  const { type, value, options, multiple } = target;
+
+  /**
+   * Because we handle any field from this function, we run through a series
+   * of checks to understand how to use the value.
+   */
+  let val;
+
+  if (/number|range/.test(type)) {
+    const parsed = parseFloat(value);
+    // If the value isn't a number for whatever reason, don't let it through because that will break the API.
+    val = isNaN(parsed) ? '' : parsed;
+  } else if (/checkbox/.test(type)) {
+    // Get & invert the current value of the checkbox.
+    val = !getIn(stateValues, field);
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -148,6 +148,40 @@ interface FormProps<TFormValues extends FormValues = FormValues>
...
+function parseInputValueFromEventTarget(target: any, stateValues: any, field: string) {
+  const { type, value, options, multiple } = target;
+
+  /**
+   * Because we handle any field from this function, we run through a series
+   * of checks to understand how to use the value.
+   */
+  let val;
+
+  if (/number|range/.test(type)) {
```
shfx, line 160: I'm not a part of the team so I don't know the intent but why is there a regex here and below? `type` can have random prefix and suffix? Is type "not-a-checkbox" still a checkbox? It's just something that have a code smell especially when it's on some hot path. It might not be event that bad but I just wonder why regex.

gitjs: The regex checks are unchanged from the existing implementation and were only moved.
Since this isn’t directly related to the changes in this PR, it would probably be better handled separately.

mathildeleg:
Hi there, thanks for your contribution!
While at first glance it does seem to improve input lag, the way you've done it actually makes every field re-render upon focusing/modifying a field in the form, which I think will definitely impact performance. Also, it unfortunately doesn't seem to improve the input lag inside dynamic zones and components.
We're taking a look at this issue on our side to find a fix, but feel free to open another PR with another way to fix this if you have an idea and we'll take a look. I'll close this one for now :)

PR #25307, opened 2026-01-31
fix(content-manager): preserve origin id when cloning, to fetch relations so they are corrected re-populated

### LINKED ISSUE #25250: Duplicate content-type with unique constraint fields fails to re-populate relations in admin
### Node Version

22.21.1

### NPM/Yarn/PNPM Version

10.9.4

### Strapi Version

5.30.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When cloning a document with relations to other collections, the relations are not properly transferred when the cloning process requires manual intervention.
If nothing prevents automatic cloning, relations are copied correctly. However, when a field constraint (such as a unique field) prevents automatic cloning, the admin UI redirects to a form page to manually create the document with pre-filled fields. In this scenario, the relations are not set in the form, even though those relations are not the fields causing the initial problem.

### Steps to Reproduce

1. Create a collection type with relations to other collections
2. Add a unique field constraint to one of the fields (e.g., slug, identifier)
3. Create an entry with populated relations
4. Attempt to clone/duplicate this entry
5. Observe that the unique field constraint prevents automatic cloning
6. Check the manual creation form that appears
7. Notice that relations are empty/not pre-filled, while other fields are

### Expected Behavior

When redirected to the manual creation form after a failed automatic clone, all non-conflicting fields (including relations) should be pre-filled with the original document's values. Only the fields causing the conflict (e.g., unique fields) should require manual input.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

jayesh70599:
### What does this PR do?

Fixes an issue where relations were not pre-populated when duplicating an entry that contains a unique field (e.g. UID), causing the admin UI to fall back to manual creation without relations.

### Root cause

In `useDocument`, a ternary expression combined with a logical OR was evaluated incorrectly due to operator precedence:

origin || id === 'create' ? undefined : id

This was interpreted as:

(origin || id === 'create') ? undefined : id

As a result, `returnId` resolved to `undefined` even when an origin document ID was present, preventing relations from being fetched during the clone fallback flow.

### Solution

Add explicit parentheses to ensure the intended evaluation order, preserving the origin document ID when available:

origin || (id === 'create' ? undefined : id)

This allows the admin UI to correctly fetch and pre-populate relation data when duplicating entries.

### How to test

1. Create a collection type with:
   - a unique field (e.g. UID)
   - a many-to-many relation
2. Create an entry and populate its relations
3. Duplicate the entry
4. When redirected to the create form, verify that relations are now pre-filled

### Related issue

Fixes #25250

### DIFF
--- a/tests/e2e/tests/content-manager/cloning.spec.ts
+++ b/tests/e2e/tests/content-manager/cloning.spec.ts
@@ -162,7 +162,6 @@ test.describe('Cloning', () => {
      * The save button should be disabled and the publish button enabled.
      */
     await page.waitForURL(CLONE_URL_ARTICLE);
-    await expect(page.getByRole('heading', { name: 'Create an entry' })).toBeVisible();
     await expect(page.getByRole('button', { name: 'Save' })).toBeEnabled();
     await expect(page.getByRole('button', { name: 'Publish' })).toBeDisabled(); // we current don't support publish & create in clone routes.
     await page.getByRole('textbox', { name: 'slug' }).fill('');


jayesh70599:
Hi @innerdvations ! Could you please review this PR too?

innerdvations:
Hello @jayesh70599 ! Thank you so much for your recent bug fixes, it's greatly appreciated. I will make sure this gets prioritized as I believe it solves at least a couple open issues. However, in the future please avoiding tagging Strapi members directly :)

mathildeleg (approved):
LGTM to me, thanks for the fix 🙏🏻

mathildeleg:
@jayesh70599 The cloning.spec.ts is failing because of this line (L165), `await expect(page.getByRole('heading', { name: 'Create an entry' })).toBeVisible();`
If you could fix this, as I can't seem to be able to push my fix into your PR :) thanks ✨

jayesh70599:
Hey @mathildeleg ! I have enabled "Allow edits and access to secrets by maintainers". Could you please try to push your fix again?

jayesh70599:
@mathildeleg  I think we can just remove the line `await expect(page.getByRole('heading', { name: 'Create an entry' })).toBeVisible();`, since the "Create an entry" heading is supposed to be replaced by the name of the duplicated entry while cloning.

PR #25316, opened 2026-02-02
feat(): adding skip AB Testing CLI option

cache-your-dreams:
### What does it do?

Adding skip AB Testing CLI option so that automation is possible

### Why is it needed?

Without this option, automatically creating a strapi app is not possible because a manual intervention is needed to AB Testing. This option makes it possible.

### How to test it?

With an experimental release, launch the following command:
npx create-strapi@experimental-number my-project \
  --skip-cloud \
  --skip-db \
  --example \
  --install \
  --git-init \
  --no-enable-ab-tests \
  --ts

### DIFF
--- a/packages/cli/create-strapi-app/src/index.ts
+++ b/packages/cli/create-strapi-app/src/index.ts
@@ -50,6 +50,10 @@ const command = new commander.Command('create-strapi-app')
   .option('--git-init', 'Initialize a git repository')
   .option('--no-git-init', 'Do no initialize a git repository')
 
+  // A/B testing options
+  .option('--enable-ab-tests', 'Enable anonymous A/B testing')
+  .option('--no-enable-ab-tests', 'Disable anonymous A/B testing')
+
   // Database options
   .option('--dbclient <dbclient>', 'Database client')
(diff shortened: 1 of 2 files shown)


innerdvations:
I like having an option for each individual item, but I think even more important would be to just have a `--non-interactive` that would accept defaults for everything

cache-your-dreams:
> I like having an option for each individual item, but I think even more important would be to just have a `--non-interactive` that would accept defaults for everything

true, I like this option as well. Seeing as this PR follows the current behaviour, maybe we could accept this and add a `--non-interactive` command to the TODO pile?

PaulBratslavsky:
@cache-your-dreams I don’t have permission to approve PRs. I implemented the changes based on your recommendations and @innerdvations’ feedback.

I also added the non-interactive flag while still allowing individual flags.

Since I couldn’t update the original PR, I created a new one, which you can see here: https://github.com/strapi/strapi/pull/25373

nclsndr:
Thank you @PaulBratslavsky 👌 
Closing that PR to avoid confusion

PR #25344, opened 2026-02-02
fix: responsive drawer for content history

Adzouz:
### What does it do?

Enhancement of the content history page for mobile and tablet devices.
This part wasn't correctly adapted to smaller devices (not to say a nightmare 😅). 

The right sidebar containing the list of versions is now in a collapsible sticky drawer at the bottom the page (making the rest of the form more useable).

Before:
Screenshot 2026-02-03 at 09 47 49

After:
https://github.com/user-attachments/assets/20993fa5-e38e-4b59-b776-3b5a9ba4cf88

### Why is it needed?

To help improve the experience of the Content Manager on mobile and tablet devices.

### How to test it?

- Go to the admin > Content Manager
- Choose a collection type that has some content and make some edits
- Open the Content History page (three dots icon on the top right of the edit view)
- Resize the window (or test directly on a mobile device) to enjoy the new experience

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
@@ -110,8 +110,14 @@ const Header = ({ isCreating, status, title: documentTitle = 'Untitled' }: Heade
           medium: 'row',
         }}
       >
-        <Flex gap={2} justifyContent="space-between" alignItems="flex-start" width="100%">
-          <Typography variant="alpha" tag="h1">
+        <Flex
+          gap={2}
+          justifyContent="space-between"
+          alignItems="flex-start"
+          width="100%"
+          overflow="hidden"
+        >
+          <Typography variant="alpha" tag="h1" overflow="hidden">
             {title}
           </Typography>
           <Box display={{ initial: 'block', medium: 'none' }}>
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/history/components/VersionsList.tsx
```diff
@@ -27,16 +28,49 @@ const BlueText = (children: React.ReactNode) => (
  * VersionCard
```
mathildeleg: This should be moved lower, above the actual VersionCard, since you added the VersionAuthor

packages/core/content-manager/admin/src/history/components/VersionsList.tsx
```diff
@@ -120,14 +142,63 @@ const VersionsList = () => {
     versions: state.versions,
     page: state.page,
   }));
+  const isMobile = useIsMobile();
 
-  return (
+  const [{ query }] = useQueryParams<{ id?: string }>();
+  const currentVersion = versions.data.find((version) => version.id.toString() === query.id);
+
+  const versionsListElements = (
```
mathildeleg: Wondering why you didn't create a separate const for this (like you did for VersionAuthor)?

Adzouz: It was mainly because I didn't want to call the hook from the context again but I guess it doesn't really change anything 😊 Pushed!

Issue #22600, opened 2025-01-09
provider-email-amazon-ses doesn't support ECS task role


Maraket:
### Required System information

- Node.js version: Node 18
- NPM version:
- Strapi version: 4.17.0
- Database: MySQL 8
- Operating system: Docker
- Is your project Javascript or Typescript: Typescript

### Describe the bug

When attempting to use the Task Role permissions of an ECS container to send emails in strapi, I get the error:

```
{"message":"[2025-01-09 00:24:38.460] \u001b[31merror\u001b[39m: The security token included in the request is invalid. ","timestamp":"2025-01-09T00:24:38.461Z","type":"out","process_id":0,"app_name":"app"}
```

This issue was noticed when attempting to migrate the app to best practices away from long-lived credentials.

The issue appears to be due to provider-email-amazon-ses which uses nodeSES (this is still true of the latest v5), which is nearly 5 years out of date and is configured such that it won't support anything but long-lived credentials.

### Steps to reproduce the behavior

1. Setup new strapi instance in docker
2. Build container
3. Set `plugins.ts` with the below config:
```
...
export default {
	...,
	email: {
		config: {
			provider: 'amazon-ses',
			providerOptions: {
				amazon: 'https://email.ap-southeast-2.amazonaws.com',
			},
			settings: {
				defaultFrom: 'my@email.com'
			},
		},
	},
...
};

```
4. Attempt to send email

### Expected behavior

Uses environment credentials and sends email.

### Additional context

This issue was not present when providing a key and secret, but this is not best practice, and I am aware I can implement our own provider, though this issue should be fixed in the core.

Maraket:
I should note, that I understand that Node 18 is nearing EOL, and that Strapi does not support Docker implementations, but neither of these factors impact the issue at hand.

Bexanderthebex:
+1 on this. The s3 provider was able to utilize the ECS task role — the ses provider shouldn't be any different

med8bra:
The official strapi provider for SES email plugin is using the deprecated [node-ses](https://www.npmjs.com/package/node-ses) module, which I believe doesn't support loading auth creds from execution context (Ecs task role for example).

It should be a simple migration as AWS provides an official client for SES [@aws/client-ses](https://www.npmjs.com/package/@aws-sdk/client-ses). And the API for the plugin is fairly simple.

If this isn't already assigned I can create a pull request with the migration

med8bra:
[...]

PR #25354, opened 2026-02-03
enhancement: replace node-ses client with AWS SES SDK

hippiewho:
### What does it do?

Updated @strapi/provider-email-amazon-ses so that it uses the '@aws-sdk/client-ses' package for better integration with AWS. This enables IAM roles so that Access Keys are no longer required. This brings the package in line with AWS recommendations for authorization.

### Why is it needed?

Current version does not allow the use of IAM Roles which then requires the generation and maintenance of Access Keys. With this update IAM Roles can be added to containers instead providing a more secure method of authentication.

### How to test it?

Using the AWS credential chaining, a tester can sign in to the AWS CLI and generate credentials for the local workstation. Once properly configured the Strapi CMS instance will automatically pick up the credentials.

### Related issue(s)/PR(s)

Related to issue:
#22600


innerdvations:
Thanks! I'm closing this in favor of https://github.com/strapi/strapi/pull/26054 but I will cherry-pick the docs from here and credit you on the release notes!

PR #25364, opened 2026-02-04
future(upload): dnd upload file

Adzouz:
### What does it do?

Implements drag and drop to upload files in the new media library.

https://github.com/user-attachments/assets/da532b62-a5dd-420c-9c58-4c289425ea6b

### Why is it needed?

To improve the UX on the media library.

### How to test it?

- Go to the new media library (activate the future flag is not already done)
- Switch to the list view (it's also working with the grid view but since it's not implemented yet, you won't be able to see your file in the list)
- Drag a file (or multiple files) from your computer and

### Related issue(s)/PR(s)

Resolves CMS-106

🚀

### DIFF
--- a/packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
+++ b/packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
@@ -110,7 +110,11 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
  * Dropzone items
  */
 
-const DropFilesMessage = () => {
+interface DropFilesMessageProps {
+  uploadDropZoneRef?: React.RefObject<HTMLDivElement>;
+}
+
+const DropFilesMessage = ({ uploadDropZoneRef }: DropFilesMessageProps) => {
   const { formatMessage } = useIntl();
   const { isDragging } = useUploadDropZone();
 
@@ -119,19 +123,20 @@ const DropFilesMessage = () => {
 
   // Calculate the left content width to position the dropzone message correctly
   useEffect(() => {
-    const mainContent = document.querySelector('[data-strapi-main-content]');
-    if (!mainContent) return;
+    if (!uploadDropZoneRef?.current) return;
 
     const updateRect = () => {
-      const rect = mainContent.getBoundingClientRect();
-      setLeftContentWidth((prev) => (prev !== rect.left ? rect.left : prev));
+      const rect = uploadDropZoneRef.current?.getBoundingClientRect();
+      if (rect) {
+        setLeftContentWidth((prev) => (prev !== rect.left ? rect.left : prev));
+      }
     };
 
     updateRect();
     const resizeObserver = new ResizeObserver(updateRect);
-    resizeObserver.observe(mainContent);
+    resizeObserver.observe(uploadDropZoneRef.current);
     return () => resizeObserver.disconnect();
-  }, []);
+  }, [uploadDropZoneRef]);
 
   if (!isDragging) return null;
 
@@ -146,7 +151,7 @@ const DropFilesMessage = () => {
       <Flex gap={2} alignItems="center">
         <Folder width={20} height={20} fill="neutral0" />
         <Typography textColor="neutral0" fontWeight="semiBold">
-          Current folder
+          Current folder{/* TODO: Replace this later with the current folder name */}
         </Typography>
       </Flex>
     </DropFilesMessageImpl>
@@ -176,7 +181,9 @@ export const AssetsPage = () => {
   const [view, setView] = usePersistentState(localStorageKeys.view, viewOptions.GRID);
   const isGridView = view === viewOptions.GRID;
 
+  // Refs
   const fileInputRef = useRef<HTMLInputElement>(null);
+  const uploadDropZoneRef = useRef<HTMLDivElement>(null);
 
   // Upload handlers
   const { toggleNotification } = useNotification();
@@ -238,87 +245,89 @@ export const AssetsPage = () => {
 
   return (
     <UploadDropZoneProvider onDrop={handleDrop}>
-      <Layouts.Root minHeight="100vh">
-        <VisuallyHidden>
-          <input type="file" ref={fileInputRef} onChange={handleFileChange} multiple />
-        </VisuallyHidden>
+      <Box ref={uploadDropZoneRef}>
+        <Layouts.Root minHeight="100vh">
+          <VisuallyHidden>
+            <input type="file" ref={fileInputRef} onChange={handleFileChange} multiple />
+          </VisuallyHidden>
 
-        <DropFilesMessage />
+          <DropFilesMessage uploadDropZoneRef={uploadDropZoneRef} />
 
-        <Layouts.Header
-          navigationAction={<Box>TODO: Breadcrumbs</Box>}
-          title="TODO: Folder location"
-          primaryAction={
-            <Flex gap={2}>
-              <SimpleMenu
-                popoverPlacement="bottom-end"
-                variant="default"
-                endIcon={<ChevronDown />}
-                label={formatMessage({ id: getTranslationKey('new'), defaultMessage: 'New' })}
-              >
-                <MenuItem onSelect={handleFileSelect} startIcon={<Files />}>
-                  {formatMessage({
-                    id: getTranslationKey('import-files'),
-                    defaultMessage: 'Import files',
+          <Layouts.Header
+            navigationAction={<Box>TODO: Breadcrumbs</Box>}
+            title="TODO: Folder location"
+            primaryAction={
+              <Flex gap={2}>
+                <SimpleMenu
+                  popoverPlacement="bottom-end"
+                  variant="default"
+                  endIcon={<ChevronDown />}
+                  label={formatMessage({ id: getTranslationKey('new'), defaultMessage: 'New' })}
+                >
+                  <MenuItem onSelect={handleFileSelect} startIcon={<Files />}>
+                    {formatMessage({
+                      id: getTranslationKey('import-files'),
+                      defaultMessage: 'Import files',
+                    })}
+                  </MenuItem>
+                </SimpleMenu>
+                <SearchInput
+                  label={formatMessage({
+                    id: getTranslationKey('search.label'),
+                    defaultMessage: 'Search for an asset',
                   })}
-                </MenuItem>
-              </SimpleMenu>
-              <SearchInput
-                label={formatMessage({
-                  id: getTranslationKey('search.label'),
-                  defaultMessage: 'Search for an asset',
-                })}
-                trackedEvent="didSearchMediaLibraryElements"
-                trackedEventDetails={{ location: 'upload' }}
-              />
-              <StyledToggleGroup
-                type="single"
-                value={isGridView ? 'grid' : 'list'}
-                onValueChange={(value) =>
-                  value && setView(value === 'grid' ? viewOptions.GRID : viewOptions.LIST)
-                }
-                aria-label={formatMessage({
-                  id: getTranslationKey('view.switch.label'),
-                  defaultMessage: 'View options',
-                })}
-              >
-                <StyledToggleItem
-                  value="list"
+                  trackedEvent="didSearchMediaLibraryElements"
+                  trackedEventDetails={{ location: 'upload' }}
+                />
+                <StyledToggleGroup
+                  type="single"
+                  value={isGridView ? 'grid' : 'list'}
+                  onValueChange={(value) =>
+                    value && setView(value === 'grid' ? viewOptions.GRID : viewOptions.LIST)
+                  }
                   aria-label={formatMessage({
-                    id: getTranslationKey('view.table'),
-                    defaultMessage: 'Table view',
+                    id: getTranslationKey('view.switch.label'),
+                    defaultMessage: 'View options',
                   })}
                 >
-                  <List />
-                  {formatMessage({
-                    id: getTranslationKey('view.table'),
-                    defaultMessage: 'Table view',
-                  })}
-                </StyledToggleItem>
-                <StyledToggleItem
-                  value="grid"
-                  aria-label={formatMessage({
-                    id: getTranslationKey('view.grid'),
-                    defaultMessage: 'Grid view',
-                  })}
-                >
-                  <GridIcon />
-                  {formatMessage({
-                    id: getTranslationKey('view.grid'),
-                    defaultMessage: 'Grid view',
-                  })}
-                </StyledToggleItem>
-              </StyledToggleGroup>
-            </Flex>
-          }
-        />
+                  <StyledToggleItem
+                    value="list"
+                    aria-label={formatMessage({
+                      id: getTranslationKey('view.table'),
+                      defaultMessage: 'Table view',
+                    })}
+                  >
+                    <List />
+                    {formatMessage({
+                      id: getTranslationKey('view.table'),
+                      defaultMessage: 'Table view',
+                    })}
+                  </StyledToggleItem>
+                  <StyledToggleItem
+                    value="grid"
+                    aria-label={formatMessage({
+                      id: getTranslationKey('view.grid'),
+                      defaultMessage: 'Grid view',
+                    })}
+                  >
+                    <GridIcon />
+                    {formatMessage({
+                      id: getTranslationKey('view.grid'),
+                      defaultMessage: 'Grid view',
+                    })}
+                  </StyledToggleItem>
+                </StyledToggleGroup>
+              </Flex>
+            }
+          />
 
-        <Layouts.Content>
-          <DropZoneWithOverlay>
-            <AssetsView view={view} />
-          </DropZoneWithOverlay>
-        </Layouts.Content>
-      </Layouts.Root>
+          <Layouts.Content>
+            <DropZoneWithOverlay>
+              <AssetsView view={view} />
+            </DropZoneWithOverlay>
+          </Layouts.Content>
+        </Layouts.Root>
+      </Box>
     </UploadDropZoneProvider>
   );
 };
--- a/packages/core/upload/admin/src/future/pages/Assets/components/DropZone/UploadDropZoneContext.tsx
+++ b/packages/core/upload/admin/src/future/pages/Assets/components/DropZone/UploadDropZoneContext.tsx
@@ -3,6 +3,7 @@ import {
   useCallback,
   useContext,
   useEffect,
+  useRef,
   useState,
   type DragEvent,
   type ReactNode,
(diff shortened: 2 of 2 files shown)


packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -61,11 +58,11 @@ const AssetsView = ({ view }: AssetsViewProps) => {
...
-    return <AssetsGrid assets={assets} />;
-  }
-
-  return <AssetsList assets={assets} />;
+  return view === viewOptions.GRID ? (
+    <AssetsGrid assets={data?.results ?? []} />
+  ) : (
+    <AssetsList assets={data?.results ?? []} />
+  );
 };
```
markkaylor: What's the value added in exchanging early return for a ternary? From a readability perspective I prefer the early return.

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -37,9 +37,6 @@ const AssetsView = ({ view }: AssetsViewProps) => {
   const { formatMessage } = useIntl();
   const { data, isLoading, error } = useGetAssetsQuery({ folder: null });
 
-  const isGridView = view === viewOptions.GRID;
-  const assets = data?.results ?? [];
```
markkaylor, line 41: Not sure I understand why we do this in two places instead of one now?

packages/core/content-manager/admin/src/history/components/VersionsList.tsx
markkaylor, line 1: Why are we seeing this change on this diff?

packages/core/admin/admin/src/components/Layouts/Layout.tsx
```diff
@@ -46,8 +46,8 @@ const OverflowingItem = styled(Box)`
   }
 `;
 
-const RootLayout = ({ sideNav, children }: LayoutProps) => (
-  <GridContainer $hasSideNav={Boolean(sideNav)}>
+const RootLayout = ({ sideNav, children, ...restProps }: LayoutProps) => (
+  <GridContainer $hasSideNav={Boolean(sideNav)} {...restProps}>
```
markkaylor, line 50: Just curious what we need this for

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -114,66 +152,151 @@ export const AssetsPage = () => {
...
+
+  const handleDropEvent = (e: React.DragEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+    setIsDraggingOver(false);
+    const { files } = e.dataTransfer;
+    if (files?.length) {
+      handleDrop(Array.from(files));
+    }
+  };
```
markkaylor: I wonder if we should organize this into a hook. The benefit I see is the logic is isolated and doesn't pollute the AssetsPage component. Wdyt?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -114,66 +152,151 @@ export const AssetsPage = () => {
...
+
   // Upload handlers
   const handleFileSelect = () => {
     fileInputRef.current?.click();
   };
 
+  const uploadFilesToFolder = async (files: globalThis.File[], folderId: number | null) => {
```
markkaylor, line 193: Cool, we don't need to worry about folders yet but that looked easy enough! Quick win. Conflict resolution for me on the other hand will be less fun 😆, nbd It'll be alright.

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -114,66 +152,151 @@ export const AssetsPage = () => {
...
+    await uploadFilesToFolder(files, null);
+  };
+
   return (
-    <Layouts.Root>
+    <Layouts.Root
+      minHeight="100vh"
+      ref={dropZoneRef}
+      data-testid="assets-dropzone"
+      onDragEnter={handleDragEnter}
```
markkaylor: Is this why you needed to upadte the Layouts component? I guess I was imagining a container inside the layout. Is that not possible? Or just curious how you got here.

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -232,15 +355,21 @@ export const AssetsPage = () => {
...
               </StyledToggleItem>
             </StyledToggleGroup>
           </Flex>
         }
       />
 
       <Layouts.Content>
-        <AssetsView view={view} />
+        <Box position="relative">
+          {isDraggingOver && <DropZoneOverlay />}
```
markkaylor: Thinking about this comment: https://github.com/strapi/strapi/pull/25364/changes#r2764823425. I wonder if we could have a component that handles everything to do with drag and drop and just returns children. 

```tsx
const DragAndDropZone = (props) => {
  const handleDragStart = () => {}
  // etc...  for all the drag and drop logic

return ( 
	<>
		{isDraggingOver && <DropZoneOverlay />}
		{props.children}
   </>
}

<DragAndDropZone>
  <AssetsView />
</DragAndDropZone>
```

tests/e2e/tests/media-library/future/file-upload.spec.ts
```diff
@@ -83,5 +83,30 @@ describeOnCondition(process.env.UNSTABLE_MEDIA_LIBRARY === 'true')(
       const assetRow = assetsPage.getAssetRow('test-image');
       await expect(assetRow).toBeVisible();
     });
+
+    test('should upload a file via drag and drop', async ({ page, browserName }) => {
+      // DataTransfer API is not supported in WebKit
+      test.skip(browserName === 'webkit', 'DataTransfer is not supported in WebKit');
```
jhoward1994: Does this mean the whole functionality doesn't work in safari or just the test?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -114,66 +152,151 @@ export const AssetsPage = () => {
...
+          <Typography textColor="neutral0">
+            {formatMessage({
+              id: getTranslationKey('dropzone.upload.message'),
+              defaultMessage: 'Drop here to upload to',
+            })}
+          </Typography>
+          <Flex gap={2} alignItems="center">
+            <Folder width={20} height={20} fill="neutral0" />
+            <Typography textColor="neutral0" fontWeight="semiBold">
+              Current folder
```
markkaylor: translation?

packages/core/upload/admin/src/future/pages/Assets/constants.ts
```diff
@@ -20,14 +20,19 @@ export const TABLE_HEADERS: TableHeader[] = [
     name: 'name',
     label: { id: getTranslationKey('list.table.header.name'), defaultMessage: 'name' },
   },
-  // TODO translation keys
   {
     name: 'createdAt',
-    label: { id: 'TODO: creation date', defaultMessage: 'creation date' },
+    label: {
+      id: getTranslationKey('list.table.header.creationDate'),
```
jhoward1994: Thank you, I missed that. I have that in my PR too - https://github.com/strapi/strapi/pull/25365

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -114,66 +152,151 @@ export const AssetsPage = () => {
...
+  }, []);
+
+  /**
+   * Handle drag and drop events (for uploading files)
+   */
+  const handleDragEnter = (e: React.DragEvent) => {
+    e.preventDefault();
+    e.stopPropagation();
+    if (e.dataTransfer.types.includes('Files')) {
+      setIsDraggingOver(true);
```
jhoward1994: Is that a confusing state name? 

To me it read like "has dragging finished" (e.g. is it "over"). When I think it means "is it over the drop zone"? wdyt, maybe I'm being silly?

packages/core/admin/admin/src/components/Layouts/Layout.tsx
markkaylor: Probably good to have this the more I think about it 👍

tests/e2e/tests/media-library/future/file-upload.spec.ts
Adzouz: Well actually I don't even know why it added that because yes, it's supposed to work also on Safari. Will remove that!

packages/core/upload/admin/src/future/pages/Assets/constants.ts
Adzouz: Do you want me to remove it from here? It was just in case you forgot

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: I don't know why Claude changed this piece of code, I put it back to what we had before!

packages/core/content-manager/admin/src/history/components/VersionsList.tsx
Adzouz: Just something I missed from my previous PR, it's not related to the ML sorry. But I saw this wasn't used anywhere in the file

packages/core/admin/admin/src/components/Layouts/Layout.tsx
Adzouz: Just to explain the why: I wanted to pass a prop to it to force the minHeight being at 100vh (so the whole page would be the dropzone). Otherwise, you couldn't drop the file where there's no content in the page and it's not ideal 😕 I tried to put the min height on a child element but since the root has a padding, it wasn't working well and didn't find a better solution..

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: It's supposed to change for the name of the current folder but it's not supported so no need for a translation here since we'll remove this string later 😊

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: That's because there's a padding in the root element and if I put the minHeight 100vh on a child, it's adding a scrollbar because of this padding. I couldn't find a "better" way but to just pass it here as a prop. I guess I could also remove the padding but it would impact other pages. I didn't think about that usecase when I redesigned everything for responsiveness 😕

packages/core/upload/admin/src/future/pages/Assets/constants.ts
jhoward1994: Actually yeah, it'll save us conflicts. Thanks!

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: I renamed it for just `isDragging`, it's simpler and less confusing 😊

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: I created a reusable component that I put in the context file. Does that work for you?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: I created a context instead because we can access more easily to the props anywhere in the children (for example in `DropZoneWithOverlay`

packages/core/upload/admin/src/future/pages/Assets/components/DropZone/UploadDropZoneContext.tsx
markkaylor, line 1: I like this but I think we should consider it a component and the context provider is just an implementation detail. I think the name should mention upload since currently that's what it is about (at least for now). I think we could:
- Move this to`pages/Assets/components` since currently it is only used in that page
- Rename `DragAndDropProvider => UploadDropZone`

Thanks for adding the sections. This is a bit opinionated but I'd like to share anyway. How you want to organize the file is up to you. 

I think the principle I'm at least trying to follow in the code base is: put code that goes together as close together as possible. 

For example here I'd have 3 sections: Context, DropZoneOverlay, UploadDropZone (ex DragAndDropProvider). Under each section I would put everything that pertains to that section. 

Context
- interface DragAndDropContextValue
- createContext call
- useDragAndDrop

DropZoneOverlay
- styled DropZoneOverlay
- DropZoneWithOverlay
- setOpacity

UploadDropZone (ex DragAndDropProvider)
- interface for component
- DropZoneWrapper
- The component wrapper with provider inside

That way everything is close to where it is actually used and is easier to work with in my opinion. I think it's when things are used in multiple places that we should move them up into more "general" areas. Happy to talk/debate about this! Sorry for the big nitpick. I do think it's a better DX but, again, up to you!

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -101,79 +106,151 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
...
+    </DragAndDropProvider>
+  );
+};
+
+interface AssetsPageContentProps {
+  handleDrop: (files: globalThis.File[]) => Promise<void>;
+  handleFileChange: (e: ChangeEvent<HTMLInputElement>) => Promise<void>;
+}
+
+const AssetsPageContent = ({ handleDrop, handleFileChange }: AssetsPageContentProps) => {
```
markkaylor: why do we need AssetsPageContent? I think it's actually making life _more_ difficult. You now have to pass props. DropZoneOverlay could be visibly read within the provider instead of buried down a level.

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -101,79 +106,151 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
...
+  const { formatMessage } = useIntl();
+  const { isDragging } = useDragAndDrop(handleDrop);
```
markkaylor: Instead of passing the handler to the hook, could it not be passed directly to the provider? I think that would simplify the whole register logic in the provider

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -101,79 +106,151 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
...
+    const updateRect = () => {
+      const rect = mainContent.getBoundingClientRect();
+      setLeftContentWidth((prev) => (prev !== rect.left ? rect.left : prev));
+    };
+
+    updateRect();
+    const resizeObserver = new ResizeObserver(updateRect);
+    resizeObserver.observe(mainContent);
+    return () => resizeObserver.disconnect();
+  }, []);
```
markkaylor: Is this to make the overlay responsive? If so I think it could be exported from the same place as the other drag and drop stuff. like `useResizeDragAndDropOverlay` or even as part of the `DropZoneWithOverlay` component?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -101,79 +106,151 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
...
-          message: errorMessage,
-        });
-      }
-    }
-    // Reset input so the same file can be selected again
-    e.target.value = '';
-  };
+  // Calculate the left content width to position the dropzone message correctly
+  useEffect(() => {
+    const mainContent = document.querySelector('[data-strapi-main-content]');
```
markkaylor: is the query selector the way to go vs a ref?

markkaylor (commented):
It works really nicely ✨ some nitpicks and other comments but we real close 🙏

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: A ref wouldn't work because it's an element that is coming from admin package so I don't know how I could target it with a ref :/ Do you have an idea?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
Adzouz: It's not for the overlay for to correctly position the info message when you drag over the content (the one that shows "Upload file to X folder" so it's not exactly tied to the dnd logic but more to this specific use case of positioning the message in the list view. It depends if we think that it's something that could be reused for another component similar to the info message. You still want me to move it?
Screenshot 2026-02-06 at 11 02 59

tests/e2e/tests/media-library/future/page-objects/AssetsPage.ts
```diff
@@ -23,13 +27,69 @@ export class AssetsPage {
     await this.page.goto('/admin/plugins/unstable-upload');
   }
 
+  /**
+   * Switch the assets view to list (table) view
+   * Waits for the page to be ready (drop zone visible) before switching
+   */
+  async switchToListView() {
```
jhoward1994, line 34: Maybe we should try and be consistent and avoid using "List" and "Table" to refer to the same thing? I kept running into this problem too. I vote we stick with "Table" everywhere as thats what the button says ?

Adzouz: You actually created the same exact function on your side, we might have a conflict here but yes, I'll stick to what you've done to limit this conflict as much as possible 👌
Only thing is that maybe we should consider renaming everything from list to table (there's a lot of stuff that are still named list...) it's either list or table but not both. Should we do that in a seperated PR?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -101,147 +106,219 @@ const StyledToggleItem = styled(ToggleGroup.Item)`
...
+      <Typography textColor="neutral0">
+        {formatMessage({
+          id: getTranslationKey('dropzone.upload.message'),
+          defaultMessage: 'Drop here to upload to',
+        })}
+      </Typography>
+      <Flex gap={2} alignItems="center">
+        <Folder width={20} height={20} fill="neutral0" />
+        <Typography textColor="neutral0" fontWeight="semiBold">
+          Current folder
```
jhoward1994: We should remember to replace this? or wait until the folder work is done ?

Adzouz: I added a todo but yes, it will be for later when we will work on folders

tests/e2e/tests/media-library/future/page-objects/AssetsPage.ts
Adzouz: Do you want me to wait for your PR to be merged and update my content to call your renamed functions or retrieve your updates on my side, merge it and resolve any conflict on your side?

PR #25366, opened 2026-02-04
enhancement: improve mobile ux of list view

mathildeleg:
### What does it do?
Improve the mobile experience of the list view page with a more appropriate action bar.

### Why is it needed?
Being able to filter, sort, and search through the list of content types is a necessity, but it was not responsive. 
Fixes CMS-233

**BEFORE** 
_without locales dropdown_
Capture d’écran 2026-02-04 à 17 43 46
Capture d’écran 2026-02-04 à 17 43 44

_with locales dropdown_
Capture d’écran 2026-02-04 à 17 43 33
Capture d’écran 2026-02-04 à 17 43 36

**AFTER**
_without locales dropdown_
Capture d’écran 2026-02-13 à 16 11 09

_without locales dropdown & with filters applied_
Capture d’écran 2026-02-13 à 16 11 41

_with locales dropdown_
Capture d’écran 2026-02-13 à 16 12 05

_with locales dropdown & with filters applied_
Capture d’écran 2026-02-13 à 16 15 11

### How to test it?
* Use your mobile device or inspect your browser in mobile view.
* Go to a content type without localization enabled. 
* Check the actions bar works OK: searching, filtering and configuring the view.
* Go to a content type with localization enabled.
* Check the actions bar works + the locales dropdown.
_* Go back to desktop and check everything works as well._

### DIFF
--- a/packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
+++ b/packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
@@ -1,8 +1,24 @@
 import * as React from 'react';
 
 import { Flex } from '@strapi/design-system';
+import { styled } from 'styled-components';
 
 import { RESPONSIVE_DEFAULT_SPACING } from '../../constants/theme';
+
+/** On mobile: first item stays left, rest stick to the right (margin-right: auto on first child). */
+const EndActionsFlex = styled(Flex)`
+  width: 100%;
+  & > *:first-child {
+    margin-right: auto;
+  }
+
+  ${({ theme }) => theme.breakpoints.medium} {
+    width: auto;
+    & > *:first-child {
+      margin-right: 0;
+    }
+  }
+`;
 
 interface ActionLayoutProps {
   endActions?: React.ReactNode;
@@ -17,19 +33,21 @@ const ActionLayout = ({ startActions, endActions, bottomActions }: ActionLayoutP
 
   return (
     <Flex
-      gap={{ initial: 2, medium: 3 }}
+      direction={{ initial: 'column', medium: 'row' }}
+      gap={{ initial: 4, medium: 3 }}
+      alignItems="flex-start"
       paddingBottom={{ initial: 2, medium: 4 }}
-      paddingTop={{ initial: 4, medium: 0 }}
+      paddingTop={0}
       paddingLeft={RESPONSIVE_DEFAULT_SPACING}
       paddingRight={RESPONSIVE_DEFAULT_SPACING}
-      wrap="wrap"
+      wrap={{ initial: 'nowrap', medium: 'wrap' }}
     >
-      <Flex gap={2} wrap="wrap" flex={1}>
+      <Flex gap={2} wrap="wrap" flex={1} width="100%">
         {startActions}
       </Flex>
-      <Flex gap={2} shrink={0} wrap="wrap">
+      <EndActionsFlex gap={{ initial: 3, medium: 2 }} shrink={0} wrap="wrap">
         {endActions}
-      </Flex>
+      </EndActionsFlex>
       {bottomActions && (
         <Flex gap={2} wrap="wrap" width="100%">
           {bottomActions}
--- a/packages/core/admin/admin/src/components/SearchInput.tsx
+++ b/packages/core/admin/admin/src/components/SearchInput.tsx
@@ -91,8 +91,7 @@ const SearchInput = ({
 
   if (isMobile) {
     // Add wrapper so that the Searchbar takes up the rest of the available space.
-    // But also prevent the Searchbar from shrinking so much that its icon is clipped.
-    return <Box style={{ flex: '1 1 4rem', minWidth: '4rem' }}>{renderSearchForm()}</Box>;
+    return <Box width="100%">{renderSearchForm()}</Box>;
   }
 
   if (isOpen) {
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -16,6 +16,7 @@ import {
   Layouts,
   useTable,
   useIsMobile,
+  useIsDesktop,
   tours,
 } from '@strapi/admin/strapi-admin';
 import {
@@ -75,6 +76,7 @@ const ListViewPage = () => {
   const { toggleNotification } = useNotification();
   const { _unstableFormatAPIError: formatAPIError } = useAPIErrorHandler(getTranslation);
   const isMobile = useIsMobile();
+  const isDesktop = useIsDesktop();
 
   usePersistentPartialQueryParams('STRAPI_LIST_VIEW_SETTINGS:', ['sort', 'filters', 'pageSize']);
   usePersistentPartialQueryParams('STRAPI_LOCALE', ['plugins.i18n.locale'], false);
@@ -274,6 +276,7 @@ const ListViewPage = () => {
 
   const endActions = (
     <>
+      {isMobile && list.settings.filterable && schema && <Filters.Trigger />}
       <InjectionZone area="listView.actions" />
       <ViewSettingsMenu
         setHeaders={handleSetHeaders}
@@ -283,7 +286,7 @@ const ListViewPage = () => {
     </>
   );
 
-  const startActions = (
+  const searchInput = (
     <>
       {list.settings.searchable && (
         <SearchInput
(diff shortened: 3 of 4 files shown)


packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
```diff
@@ -3,18 +3,42 @@ import * as React from 'react';
 import { Flex } from '@strapi/design-system';
 
 import { RESPONSIVE_DEFAULT_SPACING } from '../../constants/theme';
+import { useIsMobile } from '../../hooks/useMediaQuery';
 
 interface ActionLayoutProps {
   endActions?: React.ReactNode;
   startActions?: React.ReactNode;
+  bottomActions?: React.ReactNode;
```
Adzouz, line 26: Since it's only for mobile, I'm not a fan of this property name 😅 It could be confusing not having them displayed when not on mobile. Either add them also on desktop (and keep the conditional logic with the isMobile when calling `Layouts.Action` in `ListViewPage`, or rename the prop?

packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
```diff
@@ -3,18 +3,42 @@ import * as React from 'react';
...
-  return (
+  return isMobile ? (
+    <Flex
+      gap={3}
+      direction="column"
+      paddingBottom={2}
+      paddingTop={4}
+      paddingLeft={RESPONSIVE_DEFAULT_SPACING}
+      paddingRight={RESPONSIVE_DEFAULT_SPACING}
+      alignItems="center"
```
Adzouz: I think you don't necessarily need the isMobile check here.
Something like this should work:
```
<Flex
  gap={3}
  paddingBottom={{ initial: 2, medium: 4 }}
  paddingTop={{ initial: 4, medium: 0 }}
  paddingLeft={RESPONSIVE_DEFAULT_SPACING}
  paddingRight={RESPONSIVE_DEFAULT_SPACING}
  wrap="wrap"
>
  <Flex gap={2} wrap="wrap" flex={1}>
    {startActions}
  </Flex>
  <Flex gap={2} shrink={0} wrap="wrap">
    {endActions}
  </Flex>
  {bottomActions && (
    <Flex gap={2} wrap="wrap" width="100%">
      {bottomActions}
    </Flex>
  )}
</Flex>
```

packages/core/admin/admin/src/components/SearchInput.tsx
```diff
@@ -61,27 +63,40 @@ const SearchInput = ({
...
+      >
+        {label}
+      </Searchbar>
+    </SearchForm>
+  );
+
+  if (isMobile) {
+    // Add wrapper so that the Searchbar takes up the rest of the available space.
+    // But also prevent the Searchbar from shrinking so much that its icon is clipped.
+    return <Box style={{ flex: '1 1 4rem', minWidth: '4rem' }}>{renderSearchForm()}</Box>;
```
Adzouz: I think that 40px is not enough, we can't see anything in the search field 😕

Image

When we have the language selector, I found it a bit overloaded. Could we maybe move the language selector elsewhere or go with two rows?

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -268,87 +270,76 @@ const ListViewPage = () => {
...
-                      { target: contentTypeTitle }
-                    )}
-                    placeholder={formatMessage({
-                      id: 'global.search',
-                      defaultMessage: 'Search',
-                    })}
-                    trackedEvent="didSearch"
-                  />
+              )}
+              <Filters.Trigger />
```
Adzouz: The two Layouts.Action are pretty similar, except for this part. Could we merge to only one Layout.Root and call this conditionally instead?
```
const filterable = list.settings.filterable && schema;
...
{filterable && (
<>
	<Filters.Trigger />
	{!isMobile ? <Filters.List /> : null}
</>
)}
```

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -268,87 +270,76 @@ const ListViewPage = () => {
...
+              {!isMobile ? <Filters.List /> : null}
+            </>
+          }
+          bottomActions={isMobile && hasAppliedFilters ? <Filters.List /> : null}
```
Adzouz, line 325: And so here would be
```
{...(filterable && isMobile && hasAppliedFilters ? { bottomActions: <Filters.List /> } : {})}
```

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mathildeleg: I can't do that because otherwise if `filterable` is false, it returns nothing (which isn't what we want). 
And on top of that, I wrap it in a `<Filters.Root>`, which I don't think we'd want to do if `filterable` is false as there's no need to call it.
What I can do though is create consts for the endActions and startActions, so that I could call it something like that: 
```const endActions = (
    <>
      <InjectionZone area="listView.actions" />
      <ViewSettingsMenu
        setHeaders={handleSetHeaders}
        resetHeaders={handleResetHeaders}
        headers={displayedHeaderNames ?? []}
      />
    </>
  );

  const startActions = (
    <>
      {list.settings.searchable && (
        <SearchInput
          label={formatMessage(
            { id: 'app.component.search.label', defaultMessage: 'Search for {target}' },
            { target: contentTypeTitle }
          )}
          placeholder={formatMessage({
            id: 'global.search',
            defaultMessage: 'Search',
          })}
          trackedEvent="didSearch"
        />
      )}
    </>
  );

  const actions =
    list.settings.filterable && schema ? (
      <Filters.Root schema={schema}>
        <Layouts.Action
          endActions={endActions}
          startActions={
            <>
              {startActions}
              <Filters.Trigger />
              {!isMobile ? <Filters.List /> : null}
            </>
          }
          bottomActions={isMobile && hasAppliedFilters ? <Filters.List /> : null}
        />
        <Filters.Popover zIndex={499} />
      </Filters.Root>
    ) : (
      <Layouts.Action endActions={endActions} startActions={startActions} />
    );
```
wdyt?

packages/core/admin/admin/src/components/SearchInput.tsx
mathildeleg: I'll try a few things and get Anka's opinion on this, it's definitely not ideal on smaller screens :/

packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
mathildeleg: Not sure what you mean, because I have the conditional logic for the isMobile already?
I'd rather keep this property name generic so that it can be re-used in the future for desktop if needed, rather than mobile-specific. But I'm open to suggestions for other ideas if you have them haha (I don't)

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
Adzouz: Yes it's still better, go for it 💪 Just, why the 499 value for zIndex? I don't think it has to be that high 😂

packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
Adzouz: Yes that's it, you put the bottomActions only on mobile, that's my point and it's confusing because if for some reason tomorrow we want to add bottomActions on desktop too, it wouldn't work because of the isMobile condition 😅

mathildeleg: So if I modify the ActionLayout to this:
``` 
<Flex
      gap={3}
      paddingBottom={{ initial: 2, medium: 4 }}
      paddingTop={{ initial: 4, medium: 0 }}
      paddingLeft={RESPONSIVE_DEFAULT_SPACING}
      paddingRight={RESPONSIVE_DEFAULT_SPACING}
      wrap="wrap"
    >
      <Flex gap={2} wrap="wrap" flex={1}>
        {startActions}
      </Flex>
      <Flex gap={2} shrink={0} wrap="wrap">
        {endActions}
      </Flex>
      {bottomActions && (
        <Flex gap={2} wrap="wrap" width="100%">
          {bottomActions}
        </Flex>
      )}
</Flex>
```
Then it's okay because the `isMobile` condition is only on the ListViewPage, right? 🤔 or am I missing something here? 😂

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mathildeleg: Ugh that's probably Claude doing his thing, I'll modify it!

packages/core/admin/admin/src/components/Layouts/ActionLayout.tsx
Adzouz: Yes that's perfectly ok! The only thing I didn't want was to have the isMobile in the ActionLayout (well at least not for rendering bottomActions at least). bottomActions should always be displayed here in ActionsLayout. The isMobile condition should be in the component that calls ActionLayout (ListViewPage I think)

PR #25378, opened 2026-02-05
future(upload): add upload progress dialog

markkaylor:
### What does it do?
- Adds a new upload progress dialog to the Media Library (future upload UI) so users can see upload progress/status during file uploads.
- Introduces upload progress state management (new future store + hooks) and wires it through the upload admin api layer.
- Improves robustness by handling “catastrophic” upload errors (i.e. failures that weren’t being surfaced/handled cleanly).
- Adds retry support for cancelled uploads.
- Adds streaming endpoint

### Why is it needed?

So the user has feedback on their upload while it is in progress not matter where they are in the app

### How to test it?

Go to the unstable media libraray (add to .env UNSTABLE_MEDIA_LIBRARY=true)
Click some files to upload, you should see their progress in a dialog in the bottom right corner

In `plugins.js` add a config like this
```
upload: {
    config: {
      security: {
        allowedTypes: ['image/*'],
      },
    },
  },
  ```
Upload some images and any other file type
The images should upload successfully and errors should be reported for the other file types

Upload any file type but images
The error should be reported in the dialog

Upload a bunch of files then click cancel
You should see files that uploaded and the files that were cancelled
Click retry and the cancelled files should resume uploading (note this differs from the design which says cancel all but I think that's impossible or we would have to delete the files uploaded)

**Design Note**
I validated with Anka that the upload progress bar can be represented as indeterminate

https://github.com/user-attachments/assets/1d309519-0cd9-46a2-a94e-2478f8715567

### DIFF
--- a/packages/core/upload/server/src/controllers/__tests__/admin-upload.test.ts
+++ b/packages/core/upload/server/src/controllers/__tests__/admin-upload.test.ts
@@ -12,6 +12,7 @@ jest.mock('../../utils/mime-validation', () => ({
   prepareUploadRequest: jest.fn(() => ({
     validFiles: [{ originalFilename: 'test.jpg', mimetype: 'image/jpeg' }],
     filteredBody: {},
+    errors: [],
   })),
 }));
 
@@ -48,6 +49,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
     mockPrepareUploadRequest.mockResolvedValue({
       validFiles: [{ originalFilename: 'test.jpg', mimetype: 'image/jpeg' }],
       filteredBody: {},
+      errors: [],
     });
 
     mockAiMetadataService = {
@@ -206,6 +208,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
         filteredBody: {
           fileInfo: '{"name":"allowed.jpg","folder":null}',
         },
+        errors: [],
       });
 
       mockContext.request!.body = {
@@ -238,6 +241,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
         filteredBody: {
           fileInfo: '{"name":"test.jpg","folder":null}',
         },
+        errors: [],
       });
 
       mockContext.request!.body = {
@@ -276,6 +280,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
             { name: 'file2.png', folder: null, caption: '', alternativeText: '' },
           ],
         },
+        errors: [],
       });
 
       mockContext.request!.body = {
@@ -314,6 +319,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
       mockPrepareUploadRequest.mockResolvedValue({
         validFiles,
         filteredBody: {},
+        errors: [],
       });
 
       mockValidateUploadBody.mockResolvedValue({
@@ -348,6 +354,7 @@ describe('Admin Upload Controller - AI Service Connection', () => {
         filteredBody: {
           fileInfo: '{"name":"single.jpg","folder":null}',
         },
+        errors: [],
       });
 
       mockContext.request!.body = {

### CODE REVIEW

File: packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx [hunk 1]
```diff
@@ -0,0 +1,272 @@
...
+
+                    <Typography variant="pi">
+                      {formatMessage({
+                        id: getTranslationKey('upload.progress.label'),
+                        defaultMessage: 'Upload',
+                      })}
+                    </Typography>
+                  </Flex>
+                  {!isComplete && (
+                    <Typography variant="pi" textColor="netural600">
```
jhoward1994:
```suggestion
                    <Typography variant="pi" textColor="neutral600">
```

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,272 @@
...
+  border: 1px solid ${({ theme }) => theme.colors.neutral150};
+
+  &:focus {
+    outline: none;
+  }
+`;
+
+const DialogContentMinimized = styled(DialogContent)`
+  cursor: pointer;
+  width: 'auto';
```
jhoward1994:
```suggestion
  width: auto;
```

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const dispatch = useTypedDispatch();
+  const { isOpen, isMinimized, progress, errors, uploadId } = useTypedSelector(
+    (state) => state.uploadProgress
+  );
+
+  const hasErrors = errors.length > 0;
+  const isComplete = progress === 100;
+
+  const handleCancel = () => {
+    dispatch(closeUploadProgress());
```
jhoward1994, line 443: Does this mean it just closes the UI but wouldn't actually abort any uploads ?

markkaylor: oh yeah shit I meant to update to actually cancel the upload. I'll do that today thanks

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
markkaylor: Nice catch 😆

packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -165,6 +172,124 @@ export default {
...
+      parsedFileInfo = Array.isArray(filteredBody.fileInfo)
+        ? filteredBody.fileInfo
+        : [filteredBody.fileInfo];
+    }
+
+    // Track uploaded files before signing (needed for AI metadata)
+    const uploadedFiles: any[] = [];
+
+    // Controlled concurrency to prevent OOM on bulk uploads
+    const CONCURRENCY_LIMIT = 2;
```
jhoward1994: How did we decide on as concurrency of 2?

markkaylor: I was thinking about this again and I think the default should probably be 1 and it can be configured to a bigger batch size. But I need to do some tests this afternoon I have some doubts about what I decided to do here.

markkaylor:
The designs changed so I'll have to rework quite a bit here

packages/core/upload/server/src/controllers/content-api.ts
markkaylor, line 1: I changed the behavior of prepareUploadRequest so it doesn't fail fast in order to work for our use case. The api tests are in place to validate this.

packages/core/admin/admin/src/App.tsx
```diff
@@ -30,10 +30,23 @@ const App = ({ strapi, store }: AppProps) => {
     }
   }, []);
 
+  /**
+   * @internal
+   * @experimnental
```
jhoward1994:
```suggestion
   * @experimental
```

markkaylor: I wish I could blame AI... but I just suck at typing

packages/core/upload/admin/src/future/services/api.ts
```diff
@@ -1,23 +1,509 @@
...
+ * Retrieves stored files for an upload.
+ */
+export const getUploadedFiles = (uploadId: number): File[] | undefined => {
+  return uploadedFiles.get(uploadId);
+};
+
+/**
+ * Removes stored files when an upload completes or dialog closes.
+ */
+export const clearUploadedFiles = (uploadId: number) => {
```
jhoward1994: This doesn't seem to be called any where. Should it be in closeUploadProgress?

packages/core/upload/admin/src/translations/en.json
```diff
@@ -2,6 +2,28 @@
   "apiError.FileTooBig": "The uploaded file exceeds the maximum allowed asset size.",
   "assets.uploaded": "{number, plural, one {# asset} other {# assets}} uploaded successfully",
   "upload.generic-error": "An error occurred while uploading the file.",
+  "upload.progress": "Upload progress",
+  "upload.progress.failed": "Upload failed",
+  "upload.progress.failed.subtitle": "Please try to upload files again",
+  "upload.progress.success": "Upload successful!",
+  "upload.progress.success.subtitle": "{count} files uploaded successfully",
```
jhoward1994:
```suggestion
  "{count, plural, one {# file uploaded successfully} other {# files uploaded successfully}}"
```

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,486 @@
+import * as React from 'react';
+
+import * as Dialog from '@radix-ui/react-dialog';
+import {
+  Box,
+  Flex,
+  IconButton,
+  TextButton,
+  Typography,
+  VisuallyHidden,
```
jhoward1994:
```suggestion
```

I think this is unused?

jhoward1994:
image

In its collapsed state it looks like this. Should the list of files be completely hidden and just the green banner visible?

markkaylor:
> image
> 
> In its collapsed state it looks like this. Should the list of files be completely hidden and just the green banner visible?

Weird never saw that one, I'll take a look. Yeah it should be completely collapsed

packages/core/upload/admin/src/translations/en.json
```diff
@@ -2,6 +2,28 @@
...
+  "upload.progress.uploading.withCount": "Uploading {total} items ({percentage}%)",
+  "upload.progress.uploading-files": "Uploading {count, plural, one {# file} other {# files}}",
+  "upload.progress.indicator.in-progress": "Upload in progress indicator",
+  "upload.progress.indicator.complete": "Upload complete indicator",
+  "upload.progress.label": "Upload",
+  "upload.progress.errors.title": "Failed to upload:",
+  "upload.progress.cancel": "Cancel",
```
jhoward1994, line 16: In the designs this says "Cancel All" for multiple files

markkaylor: Yeah I'll check with Anka if it is acceptable but currently it starts uploading and streaming the results but the already uploaded files can't be canceled since they've already been uploaded. Canceling cancels the files that have not been uploaded yet. Retry will upload the files that were canceled.

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+        <Typography variant="pi" textColor="neutral600">
+          {subtitle}
+        </Typography>
+      </Dialog.Description>
+    </Flex>
+  );
+};
+
+const HeaderStatusIcon = styled(Box)`
+  padding: ${({ theme }) => theme.spaces[3]};
```
Adzouz, line 45: I would use a Flex component instead of a Box (even easier) because the cross icon is a bit off for example:

Image

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+  overflow-y: auto;
+  flex-direction: column;
+  gap: ${({ theme }) => theme.spaces[2]};
+  width: 100%;
+`;
+
+const AnimatedContent = styled.div<{ $isVisible: boolean }>`
+  display: grid;
+  grid-template-rows: ${({ $isVisible }) => ($isVisible ? '1fr' : '0fr')};
+  transition: grid-template-rows 0.3s ease-in-out;
```
Adzouz, line 416: Interesting, I never done it that way before 👀🧠

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
Adzouz: If I see correctly, the handleClose can be trigger only when uploads are completed so it wouldn't abort anything clicking on close because the icon isn't showed while it's transferring files. I wouldn't use the close button to abort the upload tbh. I would keep a cancel dedicated button for that and the close would be only to hide the popup once the upload is completed

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+
+  return (
+    <Dialog.Root open={isOpen} modal={false}>
+      <Dialog.Portal>
+        <DialogContent
+          // The accessible name is set by Dialog.Title and is dynamic,
+          // use a data-testid to ensure a stable target for e2e tests
+          data-testid="upload-progress-dialog"
+        >
+          {/* Header */}
```
Adzouz: I'm not sure this comment is necessary 🤭

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -0,0 +1,486 @@
...
+          <DialogHeader handleClose={handleClose} />
+
+          {/* Content */}
+          <AnimatedContent $isVisible={!isMinimized}>
+            <Box paddingTop={4} paddingBottom={4} paddingLeft={4} paddingRight={4}>
```
Adzouz: So since it's this element that has a overflow hidden, the padding makes the content visible when collapsed. I would move the padding to the child element:

Image

packages/core/upload/admin/src/future/store/uploadProgress.ts
```diff
@@ -0,0 +1,180 @@
...
+  errors: FileUploadError[];
+  uploadId: number;
+}
+
+export interface RootState {
+  uploadProgress: UploadProgressState;
+}
+
+const initialState: UploadProgressState = {
+  isOpen: false,
```
Adzouz: Maybe it's just me but I find it confusing that it's called isOpen because I confuse it with the isMinimize and if I understood correctly, the isOpen is more the Dialog state rather than the collapsing state of the drawer content right? Could we rename it to isVisible maybe instead?

tests/e2e/tests/media-library/future/page-objects/AssetsPage.ts
```diff
@@ -156,4 +158,18 @@ export class AssetsPage {
   getAssetCard(name: string) {
     return this.page.locator('div').filter({ hasText: name }).nth(1);
   }
+
+  /**
+   * Wait for the upload progress dialog to show success state
+   */
+  async waitForUploadProgressSuccess() {
+    await this.uploadProgressDialog.getByText('Upload successful!').waitFor({ state: 'visible' });
```
Adzouz, line 166: Would it be possible that it could become flaky if the upload takes too long to perform?

packages/core/upload/admin/src/future/services/api.ts
```diff
@@ -1,23 +1,509 @@
...
+          registerAbortController(uploadId, abortController);
+
+          try {
+            const backendURL = window.strapi.backendURL;
+            const headers: Record<string, string> = {};
+            if (token) {
+              headers.Authorization = `Bearer ${token}`;
+            }
+
+            const response = await fetch(`${backendURL}/upload/unstable/stream`, {
```
Adzouz: Dumb question here: is it okay to directly use fetch here? I don't think I seen doing that before (that's why I'm asking). Also it seem that it's the same called later in retryCancelledFilesStream, would we at least want to centralize the call?

packages/core/upload/admin/src/future/services/api.ts
```diff
@@ -1,23 +1,509 @@
...
+              headers,
+              body: formData,
+              signal: abortController.signal,
+            });
+
+            if (!response.ok || !response.body) {
+              unregisterAbortController(uploadId);
+
+              // Try to parse error message from response
+              let errorMessage = 'Upload request failed';
```
Adzouz, line 320: I assume we don't translate messages from the api right?

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
markkaylor: yeah I just thought it looked weird to have it for the others but not this one 😆 but yeah it's redundant

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
markkaylor: I'll be real transparent here...I didn't do it. So if it looks weird to you or prefer something else no problem

packages/core/upload/admin/src/future/store/uploadProgress.ts
markkaylor: ah yeah good suggestion I'll update that

tests/e2e/tests/media-library/future/page-objects/AssetsPage.ts
markkaylor: I'd say no or at least less of a risk than our typical notifications (which we assert in a lot of places) since the dialog is not a timer to disappear.

packages/core/upload/admin/src/future/services/api.ts
markkaylor: I'll look into. I think there was a reason for this with the streaming but can't quite remember which means there should be a comment here if we are going to use fetch

packages/core/upload/admin/src/future/services/api.ts
markkaylor: Good catch, it doesn't look like this is necessary. Also not sure why some of these functions are exported, I'll remove that.

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
markkaylor: Thanks for the suggestion. Crazy thing is this was still working in firefox but I had the bug in chrome. I applied your fix and verified chrome, firefox, and safari

packages/core/upload/admin/src/future/services/api.ts
markkaylor: I don't think so, here specifically this is a default/fallback if the server doesn't return an error. Pretty sure all server errors are in english so I believe it makes sense for this to be static in english. Not even sure I could translate it here in the rtk service 🤔

Adzouz:
Actually I didn't have the time to test accessibility but do you think you could test with a screen reader to see (well listen) if the updates happening in real time are correctly raised (aria-live)?

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
Adzouz: No no, I feel it's cleaner than doing the transition on a max height

Adzouz:
So I tested with a screen reader and it's reading the updates way too fast because locally it takes too little time to upload all my files so I couldn't really tell. I think it's saying out loud the title updates (with the percentage) but maybe we could see if there's anything else to do on this side to warn the user that the upload is completed, the status of it (if there are some failed uploads etc). What do you think? I think AI could help us here spotting the lacks of the feature regarding a11y

PR #25379, opened 2026-02-05
fix: scrolling in sidenav also scroll content

Adzouz:
### What does it do?

When we hover the side nav on the left and scroll inside, it also scrolls the content on the right.
This PR fixes it and contains the scroll to the hovered section only.

### Why is it needed?

To fix a scrolling issue caused by sticky element.

### How to test it?

- Go to the Content Manager
- Find a page long enough so there's a scroll
- Hover the side nav and scroll inside (anywhere in the nav, also the header)
-> The content on the right shouldn't scroll

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/22408

🚀


Adzouz:
Hey @lucasboilly, I also assigned you to this PR since you know the subject (you opened the issue) 😊

lucasboilly:
@Adzouz all good, as usual 🫶

PR #25380, opened 2026-02-05
security: upgrade to tar 7

innerdvations:
### What does it do?

upgrades to tar v7 and refactors code needed for the breaking changes in it

### Why is it needed?

to resolve a plethora of security warnings (most of which are irrelevant for Strapi)

### How to test it?

run some imports and exports and ensure that the files are created and imported properly

also added CLI test files to ensure their behaviour is correct and remains consistent

### Related issue(s)/PR(s)

Fixes CMS-208


tests/cli/tests/strapi/data-transfer/export.test.cli.js
```diff
@@ -1,39 +1,79 @@
...
+        cwd: appPath,
+        encoding: 'utf8',
+        maxBuffer: 1024 * 1024,
+      }
+    );
+
+    expect(result.status).toBe(0);
+    const stdout = String(result.stdout || '');
+    const stderr = String(result.stderr || '');
+    if (result.status !== 0) {
```
jhoward1994: Isn't that handled by L43?

Sam comment for `tests/cli/tests/strapi/data-transfer/import.test.cli.js` L76

innerdvations: Ahh, yep, in both of these files it's a useless check, good catch

tests/cli/tests/strapi/data-transfer/import.test.cli.js
```diff
@@ -26,18 +36,17 @@ describe('import', () => {
...
+      .expect('code', 0)
       .end();
   });
 
   it('should prompt for confirmation before importing data', async () => {
     await coffee
       .spawn(
         'npm',
-        ['run', '-s', 'strapi', '--', 'import', '-f', `${outputFilename}.tar`, '--force'], // TODO Remove --force
+        ['run', '-s', 'strapi', '--', 'import', '-f', `${outputFilename}.tar`, '--force'],
```
jhoward1994, line 49: The todo to remove force still seems relevant here? If we want to test prompting ?

tests/utils/fs.js
```diff
@@ -100,13 +100,12 @@ const readFile = async (archive, file, options = {}) => {
 
     streams.push(
       // Transform: tar parser
-      new tar.Parse({
+      new tar.Parser({
         // Match tar entry with the given filename
-        filter: (filePath, entry) => {
-          console.log(filePath);
+        filter(filePath, entry) {
           return entry.type === 'File' && file === filePath;
```
jhoward1994, line 106: Do we need to add a `'type' in entry` check here too ?

tests/cli/tests/strapi/data-transfer/import.test.cli.js
innerdvations: I added a bunch of test.todo so we can remember to improve the testing later!

innerdvations:
Sorry @jhoward1994 , I had "auto-merge" on, I will take a look at your suggestions in case I need a follow-up!

PR #25383, opened 2026-02-06
fix(admin): prevent stale field values from overwriting form state

### LINKED ISSUE #25143: Draft RichText(Markdown) Content Visible in Published View
### Node Version

v20.19.2

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.33.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

In the Draft section, only the draft version of the RichText(Markdown) field should be visible, and in the Published section, only the published version should be displayed. However, even when latest changes to the RichText(Markdown) field are not published, they are still visible in the Publish view.<br>Additionally, when switching back to Draft from Published, a warning message stating that changes will be lost is displayed. Since the Publish view does not have any option to save or publish changes, this warning is unnecessary.

Steps to Reproduce:

1. Create a content type in strapi with RichText(Markdown) field
2. Publish the page.
3. Modify some content in the RichText(Markdown) field of the page and save.
4. Switch between the Draft and Publish tabs a couple of times.
5. Observe that draft content appears in the Published view of the RichText(Markdown) field.

Attaching the video recording below. Please check this issue and update us on the next steps or further action.

https://github.com/user-attachments/assets/838f22f7-1454-4a72-843b-27328e3ba10c

### Steps to Reproduce

Steps to Reproduce:

1. Create a content type in strapi with RichText(Markdown) field
2. Publish the page.
3. Modify some content in the RichText(Markdown) field of the page and save.
4. Switch between the Draft and Publish tabs a couple of times.
5. Observe that draft content appears in the Published view of the RichText(Markdown) field.

### Expected Behavior

* The draft content should only visible in draft Tab should not shown in Published tab
* Warning message should not come while switching from published tab to draft tab

### Logs

```shell
```

### Code Snippets

*No response*

### Media

*No response*

### Additional information

*No response*

### Confirmation Checklist

jayesh70599:
## Summary
This PR resolves an issue where draft content incorrectly persisted in the "Published" view of the Content Manager (specifically within Rich Text/MarkDown fields). The root cause was a race condition in the `Form` component's state management during tab transitions.

## Related Issue
Fixes #25143

## The Problem
When a user switches between the **Draft** and **Published** tabs in the Content Manager:
1. The `Form` component's `useEffect` correctly triggers `SET_INITIAL_VALUES` with the fresh data from the selected tab.
2. Simultaneously, custom field components (like the Rich Text editor) may re-mount or re-render, firing a `change` event with their last-known "stale" value.
3. The `handleChange` function in `Form.tsx` would immediately dispatch `SET_FIELD_VALUE`, overwriting the correct state values with the stale data.
4. This results in the UI showing draft content in the published view and triggering unnecessary "Unsaved changes" warnings.

## Proposed Changes
- Modified the `handleChange` function within `packages/core/admin/admin/src/components/Form.tsx`.
- Added an **idempotency check** using `isEqual` and `getIn` to verify if the incoming value actually differs from the current state.
- The `SET_FIELD_VALUE` action is now only dispatched if the new value is different, effectively ignoring stale no-op events during lifecycle transitions.
- I have also verified that the issue only occurs in the case of RichText/MarkDown field. It doesn't occurs with other fields.

## How to Test
1. Create a Content Type with a **Rich Text** (MarkDown) field..
2. Create an entry, add content, and **Publish** it.
3. Modify the content in the **Draft** version and **Save** (do not publish).
4. Toggle between the **Draft** and **Published** tabs.
5. **Verification:**
    - The "Published" tab should consistently show the original published content.
    - Switching from "Published" to "Draft" should no longer trigger a "Changes will be lost" warning since no manual edits were made. 
     
## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

### DIFF
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -313,6 +313,13 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
 
     const handleChange: FormContextValue['onChange'] = useCallbackRef((eventOrPath, v) => {
       if (typeof eventOrPath === 'string') {
+        const currentValue = getIn(state.values, eventOrPath);
+
+        // Only dispatch if the specific field value is actually different
+        if (isEqual(currentValue, v)) {
+          return;
+        }
+
         dispatch({
           type: 'SET_FIELD_VALUE',
           payload: {


mathildeleg:
Thanks a lot for taking the time to work on this and propose a fix — really appreciate the effort here 🙏
I gave it a try, and while the solution works in practice, I’m not sure it addresses the issue at the right level. It feels more like a workaround than a root fix.
To explore an alternative approach, I opened a separate [PR](https://github.com/strapi/strapi/pull/25955) where I handle this directly within the Markdown editor file itself. My thinking is that keeping the logic closer to where the behavior originates might make it more robust and easier to maintain long-term.
Thanks again for your contribution and the time you put into this 🙌

PR #25391, opened 2026-02-07
security: update react-router

innerdvations:
### What does it do?

updates "react-router" to "6.30.3" and "@remix-run/router" to "1.23.2"

### Why is it needed?

Security issues

### How to test it?

no changes; everything should work fine as before

NOTE: this version of react-router now emits a `console.warn` in dev mode related to future flags for their v7 mode -- production code should not be changed

### Related issue(s)/PR(s)

Fixes CMS-192

### DIFF
--- a/packages/core/content-manager/server/src/preview/services/preview.ts
+++ b/packages/core/content-manager/server/src/preview/services/preview.ts
@@ -22,7 +22,7 @@ const createPreviewService = ({ strapi }: { strapi: Core.Strapi }) => {
 
       try {
         // Try to get the preview URL from the user-defined handler
-        return handler(uid, params);
+        return await handler(uid, params);
       } catch (error) {
         // Log the error and throw a generic error
         strapi.log.error(`Failed to get preview URL: ${error}`);


innerdvations:
Oops, this version logs a warning that is being treated as an error by front-end tests, just need to fix that...

packages/core/content-manager/server/src/preview/services/preview.ts
```diff
@@ -22,7 +22,7 @@ const createPreviewService = ({ strapi }: { strapi: Core.Strapi }) => {
 
       try {
         // Try to get the preview URL from the user-defined handler
-        return await handler(uid, params);
```
innerdvations, line 25: This was a linting error that it wouldn't let me bypass, I think it's actually a problem on `develop`

packages/core/admin/admin/tests/utils.tsx
```diff
@@ -46,6 +46,22 @@ setLogger({
...
+ */
+const REACT_ROUTER_FUTURE_FLAGS = {
+  v7_startTransition: false,
+  v7_relativeSplatPath: false,
+  v7_fetcherPersist: false,
+  v7_normalizeFormMethod: false,
+  v7_partialHydration: false,
+  v7_skipActionErrorRevalidation: false,
+} as const;
+
```
innerdvations, line 63: I didn't change this in Strapi itself because the warnings are only when `node_env !== 'production'` but we may want to add a pass through config value for users to control this on their end

PR #25400, opened 2026-02-09
chore(*): register vitest dependency in Yarn catalog

nclsndr:
### What does it do?

- Adds a Yarn catalog entry for `vitest` in `.yarnrc.yml` with version `^4.0.18`.
- Replaces per-package vitest declarations with catalog references (`catalog:`) in:
  - Root `package.json` (devDependencies)
  - `packages/core/core/package.json` (devDependencies)
  - `packages/utils/vitest-config/package.json` (devDependencies; peerDependencies remain `^4.0.0`)
- Updates `yarn.lock` accordingly.

### Why is it needed?

Centralizes the vitest version in the Yarn catalog so it can be updated in one place across the monorepo, avoids version drift between packages, and aligns with Strapi’s dependency management approach.

### How to test it?

1. From repo root: `yarn install` (lockfile should apply without errors).
2. Run vitest: `yarn test:unit:vitest` — all 19 tests in 3 files under `packages/core/core` should pass.
3. Optionally run from `packages/core/core`: `yarn test:unit:vitest` and `yarn test:unit:vitest:watch`.

**⚠️ Warning**
We must validate that the catalog feature does not collide with the release system.

### Related issue(s)/PR(s)

Follow-up to PR #25286

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -5,3 +5,6 @@ nodeLinker: node-modules
 preferInteractive: true
 
 yarnPath: .yarn/releases/yarn-4.12.0.cjs
+
+catalog:
+  vitest: ^4.0.18
--- a/package.json
+++ b/package.json
@@ -179,7 +179,7 @@
     "tar": "7.5.9",
     "ts-jest": "29.1.0",
     "typescript": "5.4.4",
-    "vitest": "4.0.18",
+    "vitest": "catalog:",
     "yalc": "1.0.0-pre.53",
     "yargs": "17.7.2"
   },
--- a/packages/core/core/package.json
+++ b/packages/core/core/package.json
@@ -136,7 +136,7 @@
     "eslint-config-custom": "5.37.1",
     "supertest": "6.3.3",
     "tsconfig": "5.37.1",
-    "vitest": "4.0.18",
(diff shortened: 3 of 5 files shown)


nclsndr:
I managed to publish the following experimental to prove we have a working flow: https://www.npmjs.com/package/@strapi/strapi/v/0.0.0-experimental.53369477a0d6496bef7d9cae680fcbde5d733f4e

nclsndr:
Thanks for the review @innerdvations.

> 1. Do we have a plan for migrating everything into it eventually?

I did not establish a plan just yet—especially not knowing how fast we want to move on this.
The way I would do it:
- [Prerequisite] Make sure the catalog feature also works with `dependencies` + `publish` (not only devDeps)
- Scan the monorepo for dependencies that are shared
- From this mapping, consciously pick the dependencies that are shared and aligned on major versions.
- For a dep: pick the latest available exact version, integrate in catalog and finally QA for regression. Move deps in group of interests (e.g: eslint + plugins), QA each.

> 2. Have you looked into using it for our internal packages? Or should we wait to touch those until we rewrite the automated releases?

I think we would prefer to use the `workspace` feature for the internal packages. Thus I would wait that we start the migration to pnpm.

PR #25401, opened 2026-02-09
fix: improve subnav on mobile so it works with banner

mathildeleg:
### What does it do?
Makes sure the banner is visible and clickable when navigating a subnav (Content Manager nav or Settings nav) on mobile.

### Why is it needed?
**BEFORE**
Capture d’écran 2026-01-22 à 17 39 22

**AFTER** 
https://github.com/user-attachments/assets/89fdc6d9-e701-46e3-99e8-b132543fc8f1

### How to test it?
* Use your mobile device or go to the Inspecter's mobile version.
* Enable a free trial/Display the upsell banner through the UpsellBanner.tsx
* Click on the Content Manager icon in the navbar.
* Make sure the upsell banner is visible and clickable, and that the subnav is still navigable. 
* Go to the Settings page and check the same things.

### DIFF
--- a/packages/core/admin/admin/src/components/SubNav.tsx
+++ b/packages/core/admin/admin/src/components/SubNav.tsx
@@ -23,17 +23,13 @@ import { tours } from './GuidedTour/Tours';
 
 const MainSubNav = styled(DSSubNav)`
   width: 100%;
-  height: calc(100dvh - ${HEIGHT_TOP_NAVIGATION} - 1px);
-  overflow: hidden;
+  height: auto;
   background-color: ${({ theme }) => theme.colors.neutral0};
   display: flex;
   flex-direction: column;
   border-right: 0;
   box-shadow: none;
-  position: fixed;
-  top: calc(${HEIGHT_TOP_NAVIGATION} + 1px);
-  left: 0;
-  z-index: 2;
+  position: relative;
 
   ${({ theme }) => theme.breakpoints.medium} {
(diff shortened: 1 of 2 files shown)


packages/core/admin/admin/src/components/SubNav.tsx
```diff
@@ -23,17 +23,13 @@ import { tours } from './GuidedTour/Tours';
 
 const MainSubNav = styled(DSSubNav)`
   width: 100%;
-  height: calc(100dvh - ${HEIGHT_TOP_NAVIGATION} - 1px);
-  overflow: hidden;
```
Adzouz, line 27: We still need the height and overflow properties on tablet, otherwise the whole layout will scroll if you want to reach the end of the sub nav 😅
Instead of removing the lines, move them to the `{({ theme }) => theme.breakpoints.medium} {` condition 😊

Adzouz (changes requested):
Thanks for the fix, though it's a bit broken on tablet, if you could just fix that please 🙏

packages/core/admin/admin/src/components/SubNav.tsx
mathildeleg: Sorry about that 🙏🏻 Should be fixed now https://github.com/strapi/strapi/pull/25401/changes/d153c3f14c68f55ffb6ce0f8b446dfc0282881e4

PR #25403, opened 2026-02-09
fix(database): include primary key in SELECT when DISTINCT is used with joins

### LINKED ISSUE #25395: strapi.db.query().findMany() with select + relation filter returns only 1 row
### Describe the bug

When using `strapi.db.query().findMany()` with a top-level `select` parameter combined with a relation filter in `where`, the query returns only **1 row** instead of all matching rows.

Removing the `select` parameter returns the correct number of rows. Using `select` without a relation filter also works correctly.

### Strapi Version

5.33.0

### Database

PostgreSQL

### Steps to reproduce

Given a `transaction` content type with a `manyToOne` relation to `advertiser`:

```typescript
// Returns 1 row (WRONG - should return 8)
const results = await strapi.db.query("api::transaction.transaction").findMany({
  where: {
    advertiser: { id: 6 },
    transaction_status: "SUCCESSFUL",
    type: "PACK_PURCHASE",
  },
  select: ["price"],
});

// Returns 8 rows (CORRECT)
const results = await strapi.db.query("api::transaction.transaction").findMany({
  where: {
    advertiser: { id: 6 },
    transaction_status: "SUCCESSFUL",
    type: "PACK_PURCHASE",
  },
  // no select
});
```

### Controlled test results

We tested 4 query variations on the same data (8 matching rows in DB):

| Query | Result | Expected |
|---|---|---|
| No `select`, with relation filter | **8 rows** | 8 ✅ |
| `select: ["price"]`, with relation filter | **1 row** | 8 ❌ |
| `select: ["price", "impressions"]`, with relation filter | **1 row** | 8 ❌ |
| `select: ["price"]`, **no** relation filter | **9 rows** (all advertisers) | 9 ✅ |

### Key findings

- The bug only occurs when `select` is combined with a **relation filter** in `where`
- The number of fields in `select` doesn't matter (1 field or 2 fields — same result)
- `select` inside `populate` (nested) is NOT affected — only the **top-level** `select`
- Without `select`, the relation filter works correctly
- Without the relation filter, `select` works correctly

### Expected behavior

`findMany()` should return all matching rows regardless of whether `select` is used.

### Workaround

Remove the top-level `select` parameter from `findMany()` calls that also use relation filters in `where`. The query will fetch all fields instead of specific ones, but will return the correct number of rows.

wingding12:
## Summary

Fixes a bug where `strapi.db.query().findMany()` with a top-level `select` combined with a relation filter in `where` returns only 1 row instead of all matching rows.

## Root Cause

When a relation filter is used in `where`, the query builder creates a JOIN and applies `DISTINCT` to prevent duplicate rows from the join. However, `processSelect()` did not include the primary key in the SELECT columns. This caused `DISTINCT` to collapse rows that shared the same values in the user-specified columns.

For example:
```typescript
// 8 rows match, but only 1 returned (rows with same price collapsed by DISTINCT)
await strapi.db.query("api::transaction.transaction").findMany({
  where: { advertiser: { id: 6 } },
  select: ["price"],
});
```

## Fix

Include the primary key (`id`) column in the SELECT when `DISTINCT` is applied due to joins. This ensures each row remains unique regardless of the selected fields.

## Tests

Added two integration tests in `tests/api/core/database/db.test.api.js`:
- `findMany with select and relation filter returns all matching rows` - verifies the core bug is fixed
- `findMany with select and relation filter works with duplicate selected values` - verifies rows are not collapsed even when selected column values are identical

## Related

Closes #25395

Made with [Cursor](https://cursor.com)


oiorain:
Hey @wingding12 

derrickmehaffy:
Hello this looks like a duplicate of https://github.com/strapi/strapi/pull/25446 so going to to go ahead and mark as closed.

PR #25414, opened 2026-02-10
fix: allow clearing placeholder and description in configure the view

### LINKED ISSUE #24161: Description and Placeholder in Configure the view can not be empty
### Node Version

20.18.1

### NPM/Yarn/PNPM Version

YARN 1.22.22

### Strapi Version

5.21.0

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

It impossible to reset field Description or Placeholder to empty value in Configure the view.

Please check Steps to Reproduce to find more info.

### Steps to Reproduce

1. Create collection with one field
2. Go to this collection (via Content Manager) and push Configure the view
3. Go to this field properties and enter random string in Placeholder and Description
4. Save this view
5. Go one more time to this field properties and try to remove Placeholder and Description

When you remove Description and try press Finish, the error appears:

Image

When you remove Placeholder you can press Finish (everything is fine), but when you press Save, another error appears:

Image

### Expected Behavior

I should be able to reset Placeholder and Description to empty value.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

Eventyret:
## Problem

Fixes #24161

`setModelConfiguration` in `packages/core/content-manager/server/src/services/utils/store.ts` skips writing values that are `null` or `undefined`. When a user clears the placeholder or description fields in the "Configure the View" modal, the cleared value (`null`) is never persisted to the store because the condition `value[key] !== null && value[key] !== undefined` filters it out. The old value stays in place.

## Fix

Changed the guard in `setModelConfiguration` to only skip `undefined` values, allowing `null` to pass through. `null` is the explicit "cleared" signal from the frontend validation schema (`yup.string().nullable()`), while `undefined` means "not provided" and should continue to be skipped (e.g., `isComponent` on non-component configurations).

The one-line change:
```diff
- if (value[key] !== null && value[key] !== undefined) {
+ if (value[key] !== undefined) {
```

## Testing

Added unit tests for `setModelConfiguration` covering:
- Normal non-null values persist correctly
- `null` values persist (clearing a field)
- Empty string values persist (clearing a field)
- `undefined` values are still skipped
- No write occurs when config hasn't changed
- Top-level `null` values persist

### DIFF
--- a/packages/core/content-manager/server/src/services/utils/__tests__/store.test.ts
+++ b/packages/core/content-manager/server/src/services/utils/__tests__/store.test.ts
@@ -1,0 +1,182 @@
+import storeUtils from '../store';
+
+const mockGet = jest.fn();
+const mockSet = jest.fn();
+
+describe('store utils', () => {
+  beforeEach(() => {
+    global.strapi = {
+      store: jest.fn(() => ({
+        get: mockGet,
+        set: mockSet,
+      })),
+    } as any;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('setModelConfiguration', () => {
+    test('persists normal non-null values', async () => {
+      mockGet.mockResolvedValue({
+        settings: { pageSize: 10 },
+        metadatas: {},
+        layouts: { list: [], edit: [] },
+      });
+
+      await storeUtils.setModelConfiguration('content_types::api::article.article', {
+        settings: { pageSize: 25 },
+        metadatas: {
+          title: {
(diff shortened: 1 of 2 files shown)


Eventyret:
Closing this one. After further investigation the store guard isn't the root cause here. The null description/placeholder values are nested inside the metadatas object, which is always a non-null object at the top level, so the guard was never blocking them. The actual bug is in the validation layer (frontend form validation error when clearing description, server-side error on save when clearing placeholder). Will open a new PR with a proper fix targeting the validation behavior.

PR #25416, opened 2026-02-10
fix(admin): handle 204 no-content response in fetch client

### LINKED ISSUE #25274: [Safari] Password Recovery shows 'The string did not match the expected pattern' error due to 204 response handling
### Node Version

22.x

### NPM/Yarn/PNPM Version

10.x

### Strapi Version

5.31.0

### Operating System

macOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/Richi2293/strapi-test-email (referenced from #24989)

### Bug Description

On **Safari only**, the Password Recovery page (`/admin/auth/forgot-password`) shows the error message "The string did not match the expected pattern." after clicking "Send Email", even though the API request **succeeds with HTTP 204**.

**Root Cause Analysis:**

I investigated this issue using Safari's Web Inspector and found the following:

1. The `POST /admin/forgot-password` API returns **HTTP 204 No Content** (which is correct and successful)
2. The Strapi admin panel's React code calls `response.json()` on this 204 response
3. **Safari throws a specific error** when `response.json()` is called on a 204 (No Content) response: `"The string did not match the expected pattern."`
4. Chrome handles this differently and does not show this error to the user

**Evidence:**

```javascript
// Intercepted fetch in Safari DevTools
{
  "url": "http://localhost:1337/admin/forgot-password",
  "method": "POST",
  "status": 204,  // Success!
  "error": "The string did not match the expected pattern."  // from response.json()
}
```

**This is a Safari-specific bug** in how it handles `response.json()` on 204 responses, but Strapi should handle this case properly.

### Steps to Reproduce

1. Open Safari browser (macOS)
2. Navigate to `/admin/auth/forgot-password`
3. Enter a valid email address
4. Click "Send Email"
5. Error message appears: "The string did not match the expected pattern."

**Note:** This does NOT occur in Chrome or Firefox.

### Expected Behavior

The Password Recovery should show a success message (or no error) after sending the email, since the API request actually succeeds with HTTP 204.

### Suggested Fix

The admin panel should check the response status before calling `response.json()`:

```javascript
// Instead of:
const body = await response.json();

// Do:
const body = response.status === 204 ? {} : await response.json();
```

Or use a more robust approach that handles empty responses gracefully.

### Related Issue

This is related to #24989 which was closed as "can not reproduce". The reporter was likely using Safari but did not specify the browser. I have identified the root cause and confirmed it is Safari-specific.

### Logs

```shell
# No server-side errors - the API works correctly
# The error is client-side only in Safari
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

- Tested on Safari 18.x on macOS
- Chrome and Firefox do not exhibit this issue
- The underlying API works correctly (HTTP 204 is returned as expected)
- Email is actually sent successfully despite the error message

### Confirmation Checklist

Eventyret:
## Problem

The `POST /admin/forgot-password` endpoint returns HTTP 204 (no body). The `responseInterceptor` in `getFetchClient.ts` unconditionally calls `response.json()` on every response, which fails on empty bodies.

In Chrome/Firefox this throws a `SyntaxError`, which is caught by the existing error handler and silently returns `{ data: [] }`. In Safari, the same call throws a `TypeError` ("The string did not match the expected pattern"), which is **not** caught. This bubbles up as a visible error to the user during password recovery.

Fixes #25274

## Fix

Added an early return for `204 No Content` responses before attempting JSON parsing. When the response status is 204, the interceptor returns `{ data: {}, status: 204 }` immediately without calling `response.json()`. This avoids the browser-specific error entirely rather than relying on catching different exception types across browsers.

## Test plan

- Added test verifying 204 responses return `{ data: {}, status: 204 }` and `response.json()` is never called
- Added test verifying 204 handling works across all HTTP methods (GET, POST, PUT, DELETE)
- All existing tests continue to pass


innerdvations (approved):
LGTM, thanks!

packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -298,14 +298,13 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
       return { data: result as TData, status: response.status, headers: response.headers };
     }
 
+    if (response.status === 204) {
+      return { data: {} as TData, status: response.status };
```
unrevised6419, line 302: Question: @Eventyret @innerdvations any reason why here we return an object `{}` and below in catch block an array `[]` is returned? 🤔

innerdvations: Good catch. Yes, the one below could be updated to return `{}` as well for consistency. We can do that in a follow-up

PR #25418, opened 2026-02-10
fix(admin): add cursor pointer and fix click behavior on sortable table headers

### LINKED ISSUE #24087: Sorting in Content Manager is buggy
### Node Version

20.19.0

### NPM/Yarn/PNPM Version

NPM 10.8.1

### Strapi Version

5.18.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

The whole experience of sorting in Content Manager is less pleasant and feels buggy.
- The cursor does not indicate which columns are sortable (in v4, it changes to a "hand" cursor", in v5 it doesn't)
- Some clicks on column header are ignored, the content just scrolls to left and I have to click the column again to actually sort.
- The first click on a "new sort order" causes full refresh, but subsequent selection of this "sort order" is smooth, as it should be in React applications. (E.g. when I select "by Title DESC" sorting, there is full refresh on first time, but when I return to this sorting, there is no full refresh). 

### Steps to Reproduce

1. Open a content type entries view and try to play with sorting (click on different columns headers).

### Expected Behavior

Sorting in Content Manager is smooth, intuitive, it's clear what is sortable and what is not, and it does not feel buggy.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

https://github.com/user-attachments/assets/b215f1a7-7083-469d-b17b-7a5805806d55

### Additional information

_No response_

### Confirmation Checklist

Eventyret:
## Problem

Sortable column headers in the Content Manager list view have UX issues:

- No `cursor: pointer` on sortable columns, giving users no visual feedback that columns are clickable
- The `Typography` inside `HeaderCell` switches between rendering a `<button>` (when not sorted) and a `<span>` (when sorted), causing inconsistent click targets
- The `onClick` handler is only on the inner `Typography` element, so clicking the padding area of the header cell does nothing

Fixes #24087

## Fix

- Moved the `onClick` handler from the inner `Typography` to the `Th` wrapper itself, making the entire header cell area clickable
- Added `cursor: pointer` on sortable column headers via a `SortableTh` styled component
- Changed the `Typography` to always render as a `<span>`, removing the inconsistent `button`/`span` tag switching
- Non-sortable columns remain unaffected (`cursor: default`, no click handler)

### DIFF
--- a/packages/core/admin/admin/src/components/Table.tsx
+++ b/packages/core/admin/admin/src/components/Table.tsx
@@ -198,7 +198,9 @@ const HeaderCell = <TData, THead>({ name, label, sortable }: TableHeader<TData, 
   };
 
   return (
-    <Th
+    <SortableTh
+      onClick={sortable ? handleClickSort : undefined}
+      $sortable={!!sortable}
       action={
         isSorted &&
         sortable && (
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/components/Table.tsx
Adzouz: Now the action is on the column header, I don't think we need this button anymore. Just displaying the icon would be enough

Adzouz (changes requested):
Hi @Eventyret, thanks for the contribution 🙏
I left a comment if you could address it please 😊
Also can you update your branch with develop latest changes?

Adzouz (approved):
I took over the PR to move forward, hope you don't mind 🙏

PR #25424, opened 2026-02-11
feat(ts): augment all context error response methods

unrevised6419:
### What does it do?

Properly types **all** Koa context response methods, the ones generated in:

https://github.com/strapi/strapi/blob/884a7a32a8a4a76091a81c0c29c820846dea32b9/packages/core/core/src/services/server/koa.ts#L22-L35

so they are available on both `ctx.*` and `ctx.response.*` (they are delegated from `context.response` to `context`).

- Adds `packages/core/core/src/services/server/koa-methods.ts` defining:
  - `ContextDelegatedResponseErrorMethods` — every HTTP error helper (`400` Bad Request → `511` Network Authentication Required).
  - `ContextDelegatedResponseSuccessMethods` — `send`, `created`, `deleted`.
  - `errorMethodEntries()` — a generator that maps each `node:http` `STATUS_CODES` entry in the `400`–`599` range to its camelCased helper name. This is the single source of truth: both the runtime registration in `koa.ts` and the unit test iterate it, so the generated helpers cannot drift from the iterator.
- Wires these into Koa via the new `DefaultContextDelegatedResponse` interface in `packages/core/core/src/index.ts`, so the methods resolve on both `ctx` and `ctx.response`.
- Bumps `@types/koa` `2.13.4` → `2.15.2` (the resurrected v2 typings that expose `ContextDelegatedResponse` for augmentation).
- Removes the old, partially-wrong augmentations:
  - the incomplete `ExtendableContext` block in `packages/core/utils/src/types.ts` (declared a non-existent `ctx.ok()` and the wrong return type).
  - the loose `BaseResponse` index-signature hack in `koa.ts`.
- Fixes the return type: these methods return `void` (matching their implementation), not `Koa.Context` as previously typed.
- Moves the existing `BaseResponse._explicitStatus` type augmentation next to where it is used (the error middleware).
- Renames the first parameter from `message` to `response` to match the actual semantics.
- Replaces the `statuses` package with Node's built-in `node:http` `STATUS_CODES` for generating the error-method names. The iteration (skipping codes outside the `400`–`599` range) lives in the shared `errorMethodEntries()` generator. Drops the `statuses` and `@types/statuses` dependencies.
- Adds a unit test (`packages/core/core/src/services/server/__tests__/koa.test.ts`) that guards the hand-maintained types against the runtime iterator: it reads the interface members via the TypeScript compiler API and asserts they match the names `errorMethodEntries()` yields, that `createKoaApp` registers every helper on both `ctx` and `ctx.response`, and that the helpers set the expected status/body.
- Adjusts `packages/core/admin/server/src/controllers/admin.ts`: moves `satisfies Plugins.Response` inside the `ctx.send(...)` call, now that `send` returns `void`.
- Removes redundant `@param {import('koa').BaseContext} ctx` JSDoc lines in the review-workflows controllers.

### Why is it needed?

Before this PR only a handful of error methods were typed, the declared return type was wrong (`Koa.Context` instead of `void`), and a non-existent `ctx.ok()` was advertised. Users now get correct type completion and no TypeScript errors for every error/success response method, whether called via `ctx` or `ctx.response`.

### How to test it?

Call any response method in a controller, e.g. `ctx.lengthRequired()` or `ctx.response.notFound()` — autocomplete works and the return type is `void`.

image

### Related issue(s)/PR(s)

Depends on two upstream DefinitelyTyped PRs (both now merged):

- DefinitelyTyped/DefinitelyTyped#74990 — [koa] Add v2 types
- DefinitelyTyped/DefinitelyTyped#74991 — [koa] Allow augmenting ContextDelegatedRequest/Response


packages/core/admin/server/src/controllers/admin.ts
```diff
@@ -175,7 +175,7 @@ export default {
         packageName: plugin.info.packageName,
       }));
 
-    ctx.send({ plugins }) satisfies Plugins.Response;
+    ctx.send({ plugins } satisfies Plugins.Response);
```
unrevised6419, line 188: Note: `ctx.send` does not return any type

packages/core/utils/src/types.ts
```diff
@@ -94,16 +94,51 @@ declare module 'koa' {
   }
 
   interface ExtendableContext {
```
unrevised6419: Note: Theoretically we need to augment the `ContextDelegatedResponse` from `koa` because these methods are delegated from `context.response` to `context`

All these methods exist on `context.response.badRequest()` but if the user tries to call it, it will get a TypeScript error.

Anything that is augmented on `ContextDelegatedResponse` can be accessed via `context` or `context.response`

This can be done in a subsequent PR because it is a little messed up with the types

packages/core/utils/src/types.ts
```diff
@@ -94,16 +94,51 @@ declare module 'koa' {
...
-    ok: (response?: string | object, details?: object) => Koa.Context;
```
unrevised6419, line 97: Note: There is no such method `ctx.ok()`. Calling it, it will throw an error.

packages/core/utils/src/types.ts
```diff
@@ -94,16 +94,51 @@ declare module 'koa' {
...
-    created: (response?: string | object, details?: object) => Koa.Context;
-    noContent: (response?: string | object, details?: object) => Koa.Context;
-    badRequest: (response?: string | object, details?: object) => Koa.Context;
-    unauthorized: (response?: string | object, details?: object) => Koa.Context;
-    forbidden: (response?: string | object, details?: object) => Koa.Context;
-    notFound: (response?: string | object, details?: object) => Koa.Context;
-    locked: (response?: string | object, details?: object) => Koa.Context;
-    internalServerError: (response?: string | object, details?: object) => Koa.Context;
-    notImplemented: (response?: string | object, data?: Record<string, unknown>) => Koa.Context;
+    /* 400 */ badRequest(response?: string | object, details?: object): void;
```
unrevised6419: Note: Previously all these methods types were returning `Koa.Context`, which is not true. Their implementation does not return anything which is why they now return `void`.

packages/core/utils/src/types.ts
unrevised6419: Image

packages/core/utils/src/types.ts
unrevised6419: Image

unrevised6419:
Moved to draft because I want to fix it fully the right way.

For that I will need to augment `ContextDelegatedResponse`, which at the moment is not possible because it's not exposed from `@types/koa`

I did open a PR to resurrect `@types/koa@2` in DefinitelyTyped, and will open another PR there to move `ContextDelegatedResponse` in the namespace to be able to augment it.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74990
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74991

unrevised6419:
All works good now. Waiting for upstream PRs to get merged.

packages/core/core/src/services/server/koa.ts
```diff
@@ -2,37 +2,32 @@ import { isNil, camelCase } from 'lodash/fp';
...
-  statuses.codes
-    .filter((code) => code >= 400 && code < 600)
-    .forEach((code) => {
-      const name = statuses(code);
+  for (const [codeStr, name] of Object.entries(STATUS_CODES)) {
+    const code = Number(codeStr);
+
+    if (!name || code < 400 || code >= 600) {
+      continue;
+    }
```
unrevised6419: Note: The definition for `STATUS_CODES` is this

```ts
const STATUS_CODES: {
    [errorCode: number]: string | undefined;
    [errorCode: string]: string | undefined;
};
```

Note: `STATUS_CODES` is identical with `statuses(number) -> statuses.message`

Test

```ts
import { STATUS_CODES } from 'node:http';
import statuses from 'statuses';

for (const [code, message] of Object.entries(STATUS_CODES)) {
  const status = Number(code);
  const actual = statuses.message[status];

  if (actual !== message) {
    throw new Error(`STATUS_CODES mismatch for ${code}: expected ${message}, got ${actual}`);
  }
}
```

packages/core/utils/src/types.ts
```diff
@@ -88,31 +87,4 @@ export interface Model {
   attributes: Record<string, AnyAttribute>;
 }
 
-declare module 'koa' {
-  interface Request extends Koa.BaseRequest {
-    route: RouteInfo;
```
unrevised6419, line 93: Note: This field is never set in Strapi internals. In a real Strapi application, `ctx.request.route` is always undefined.

packages/core/utils/src/types.ts
```diff
@@ -88,31 +87,4 @@ export interface Model {
...
-    unauthorized: (response?: string | object, details?: object) => Koa.Context;
-    forbidden: (response?: string | object, details?: object) => Koa.Context;
-    notFound: (response?: string | object, details?: object) => Koa.Context;
-    locked: (response?: string | object, details?: object) => Koa.Context;
-    internalServerError: (response?: string | object, details?: object) => Koa.Context;
-    notImplemented: (response?: string | object, data?: Record<string, unknown>) => Koa.Context;
-  }
-}
-
-export interface RouteInfo {
```
unrevised6419, line 110: Note: This is not used exposed to Public API, and is also not used internally.

innerdvations (commented):
## Review summary

LGTM overall — typings match runtime, `@types/koa` 2.15 `DefaultContextDelegatedResponse` augmentation is the right layer, and swapping `statuses` for `node:http` `STATUS_CODES` produces the same 41 error helpers. TS-only improvement; no runtime concerns from what I checked.

A few recommended fixes / follow-ups before merge:

### 1. Typo in success-methods interface name

`ContextDelegatedResponeSuccessMethods` is missing the second **s** in Response. It appears in:

- `packages/core/core/src/services/server/koa-methods.ts` (definition)
- `packages/core/core/src/index.ts` (import + `DefaultContextDelegatedResponse` extends)

Suggest renaming to `ContextDelegatedResponseSuccessMethods` before this spreads further.

### 2. Keep `koa-methods.ts` in sync with the runtime loop

`koa-methods.ts` is hand-maintained while `koa.ts` generates helpers from `STATUS_CODES`. They match today (41 codes), but could drift if Node adds a new 4xx/5xx name.

Optional hardening: a small unit test that (a) compile-time-checks `keyof ContextDelegatedResponseErrorMethods` against camelCased `STATUS_CODES` names and (b) asserts `createKoaApp` registers each helper on both `ctx` and `ctx.response`.

### 3. Release-note heads-up for TS plugin authors

Removals are correct (phantom APIs): `ctx.ok()`, `ctx.noContent()`, `ctx.request.route`, and `Koa.Context` return types on response helpers. No monorepo usages; Strapi route metadata is on `ctx.state.route`. Worth a brief note for TS plugin authors — runtime unchanged.

_AI-assisted comment._

innerdvations (approved):

PR #25425, opened 2026-02-11
fix(history): improve error handling and batch deletion in cron jobs

jhoward1994:
## What does it do?

- **Batch deletion for history versions cron job:** Refactors the `deleteHistoryDaily` cron job to use a batched loop that deletes up to 1,000 expired records per iteration
- **Proper await and error handling in audit logs cron job:** Ensures that the audit log cleanup task is properly awaited and any errors are caught and logged, preventing unhandled promise rejections.
- **Cron-level error listener:** Adds an error handler (`job.on('error', ...)`) in the cron service so uncaught errors from a cron task are logged with the job name and don't crash the process.

## Why is it needed?

The original history cleanup cron job executed a single `DELETE ... WHERE created_at < ?` query causing timeouts for other API operations. 
Lack of proper awaiting and error handling meant failures could go unhandled

## Related issue(s) / PR(s)

Fix CMS-326

### DIFF
--- a/packages/core/admin/ee/server/src/audit-logs/services/lifecycles.ts
+++ b/packages/core/admin/ee/server/src/audit-logs/services/lifecycles.ts
@@ -159,12 +159,8 @@ const createAuditLogsLifecycleService = (strapi: Core.Strapi) => {
       strapi.cron.add({
         deleteExpiredAuditLogs: {
           task: async () => {
-            try {
-              const expirationDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
-              await auditLogsService.deleteExpiredEvents(expirationDate);
-            } catch (error) {
-              strapi.log.error('Error deleting expired audit logs', error);
-            }
+            const expirationDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
+            await auditLogsService.deleteExpiredEvents(expirationDate);
           },
           options: '0 0 * * *',
         },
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/server/src/history/services/lifecycles.ts
```diff
@@ -175,24 +175,46 @@ const createLifecyclesService = ({ strapi }: { strapi: Core.Strapi }) => {
...
-            const expirationDate = new Date(Date.now() - retentionDaysInMilliseconds);
-
-            strapi.db
-              .query(HISTORY_VERSION_UID)
-              .deleteMany({
-                where: {
-                  created_at: {
-                    $lt: expirationDate,
+            const BATCH_SIZE = 1000;
+
```
innerdvations, line 179: It looks like we're now logging errors twice because these are already caught at a higher level. I think you can remove the try/catches in this block

packages/core/admin/ee/server/src/audit-logs/services/lifecycles.ts
```diff
@@ -159,8 +159,12 @@ const createAuditLogsLifecycleService = (strapi: Core.Strapi) => {
       strapi.cron.add({
         deleteExpiredAuditLogs: {
           task: async () => {
-            const expirationDate = new Date(Date.now() - retentionDays * 24 * 60 * 60 * 1000);
-            auditLogsService.deleteExpiredEvents(expirationDate);
+            try {
```
innerdvations: Same here, we don't need to log this, we can remove the try/catch, or keep it to rethrow a more specific error

innerdvations (changes requested):

packages/core/content-manager/server/src/history/services/lifecycles.ts
innerdvations: not sure why this line of code got targeted for the comment 😆 but it's a general comment for the whole block anyway

PR #25426, opened 2026-02-11
ci: remove lerna

innerdvations:
### What does it do?

removes lerna from all tooling

### Why is it needed?

we only used it in one single script after replacing it with nx a couple years ago

### How to test it?

### Related issue(s)/PR(s)

Fixes CMS-478


innerdvations:
I'll just close this for now but add an issue in the automated releases project

PR #25433, opened 2026-02-11
feat(content-manager): add relationOpenMode setting (modal/page/newTab)

Schero94:
### What does it do?

Adds a new per-content-type setting called **"Relation open behavior"** (`relationOpenMode`) to the Content Manager. This setting allows users to choose how relation entries open when clicked:

- **Modal** (default) — opens the relation in the existing inline modal (current behavior, unchanged)
- **Navigate to page** — navigates to the relation's edit page in the same tab
- **Open in new tab** — opens the relation's edit page in a new browser tab

The setting is available in each content type's "Configure the view" page alongside the existing "Entry title" dropdown.

#### Technical changes

**Shared contracts** (`shared/contracts/content-types.ts`)
- Added `RelationOpenMode` type (`'modal' | 'page' | 'newTab'`) and extended the `Settings` interface

**Server-side** (`server/src/`)
- Added `relationOpenMode: 'modal'` to `DEFAULT_SETTINGS` and `settingsFields` in `services/utils/configuration/settings.ts`
- Added Yup validation (`yup.string().oneOf(['modal', 'page', 'newTab']).default('modal')`) in `controllers/validation/model-configuration.ts`

**Admin-side** (`admin/src/`)
- Extended `ConfigurationForm/Form.tsx` with an enumeration input for the new setting
- Modified `RelationModal.tsx`: reads `relationOpenMode` from the document layout settings via `useDocumentLayout`, passes it through context, and switches behavior in `RelationModalTrigger` accordingly
- Modified `Relations.tsx`: applies the same open behavior to the "Create new entry" action in the Combobox
- Added `getAdminBasePath()` utility for correctly constructing full admin URLs (used by the "new tab" mode)
- Added `DEFAULT_SETTINGS.relationOpenMode` fallback in `useDocumentLayout.ts`
- Added 5 translation keys in `translations/en.json`

**Tests**
- Updated `settings.test.ts` to include `relationOpenMode: 'modal'` in expected defaults
- Updated `useDocumentLayout.test.ts` to include `relationOpenMode: 'modal'` in expected edit and list layout settings

### Why is it needed?

Currently, clicking a relation in the Content Manager always opens an inline modal. While this works for quick edits, it becomes limiting for content types with complex or deeply nested relations — users often want to navigate directly to the full edit page instead.

This PR gives content editors the flexibility to choose their preferred workflow per content type, improving usability especially for:
- Content types with many or complex relations
- Workflows where editors frequently need the full edit view
- Power users who prefer opening entries in new tabs for parallel editing

### How to test it?

1. Start Strapi with `yarn develop`
2. Navigate to **Content Manager → Any Content Type → Configure the view** (gear icon)
3. You should see a new **"Relation open behavior"** dropdown next to "Entry title"
4. Test each option:
   - **Modal (default)**: Click a relation → modal opens (existing behavior)
   - **Navigate to page**: Click a relation → browser navigates to the edit page in the same tab
   - **Open in new tab**: Click a relation → edit page opens in a new browser tab (with correct authentication)
5. Verify the setting persists after saving and page reload
6. Verify the "Create new entry" action in the relation Combobox also respects the setting
7. Run tests:
   ```bash
   npx nx run @strapi/content-manager:test:unit
   npx nx run @strapi/content-manager:test:front
   ```

### Related issue(s)/PR(s)

- N/A — Feature proposal for improved relation editing UX

### DIFF
--- a/packages/core/content-manager/shared/contracts/content-types.ts
+++ b/packages/core/content-manager/shared/contracts/content-types.ts
@@ -12,7 +12,7 @@ export type Settings = {
   mainField: string;
   defaultSortBy: string;
   defaultSortOrder: string;
-  relationOpenMode: RelationOpenMode;
+  relationOpenMode?: RelationOpenMode;
 };
 
 export type Metadatas = {


oiorain:
Hi @Schero94!

Thank you so much for your all contributions. We really appreciate the hard work you're putting into Strapi :) 

Since this is a product decision, let me check in with Product an Design on this. I will come back to you as soon as possible. 

Thanks again

Schero94:
Totally understand, that’s exactly why I didn’t touch the default behavior.
I only added optional settings so teams can decide for themselves.

Right now, opening the relation dialog can become quite frustrating when there’s already a lot of data in it. The goal was simply to make it smoother and more usable again, without forcing any change on existing setups.

Happy to adjust things if needed, and I’m looking forward to your feedback after syncing with Product and Design 😊

Schero94:
@oiorain Bonjour, just wanted to ask how far the discussion is? It slows us really down in Working right now would be nice if we could get it fixed and reviewed :)

oiorain:
Hi @Schero94 ! 

Product and Design are aligned with this change so we'll get it reviewed by engineering as soon as we can 🙏 
We'll keep you updated asasp.

markkaylor (dismissed):
It looks good to me. Could you please lint/pretty your code with `yarn lint:fix` and `yarn prettier:write`. I'll see what needs to be done for the documentation side and put the PR together then we can merge.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -207,6 +207,8 @@ function reducer(state: State, action: Action): State {
   }
 }
 
+type RelationOpenMode = 'modal' | 'page' | 'newTab';
```
jhoward1994: Can this be imported from packages/core/content-manager/shared/contracts/content-types.ts ?

packages/core/content-manager/admin/src/components/ConfigurationForm/Form.tsx
```diff
@@ -107,7 +107,7 @@ const ConfigurationForm = ({
                   defaultMessage: 'Settings',
                 })}
               </Typography>
-              <Grid.Root>
+              <Grid.Root gap={5}>
```
markkaylor: This gap creates quite a big space between the word "view" and the rest of the content.

Without:

Image

With:

Image

Could you update so the spacing only applies to the inputs above the dividing line?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -534,6 +553,17 @@ const RelationModalBody = () => {
...
 };
 
+/**
+ * Get the admin base path (e.g. "/admin") to construct full URLs for window.open().
+ * React Router's navigate() only needs the relative SPA path, but window.open()
+ * needs the full path including the admin prefix.
+ *
+ * Mirrors the logic from @strapi/admin's internal getBasename() utility.
+ */
+const getAdminBasePath = () => {
```
jhoward1994: Can we use the util from packages/core/admin/admin/src/core/utils/basename.ts ?

edit: ah I see from the comment that you know the util is there, maybe we can export it from the admin package?

markkaylor: I'm not sure it's exported but I guess we could export it 🤷

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/RelationModal.tsx
```diff
@@ -542,19 +572,33 @@ const RelationModalTrigger = ({
...
-    <StyledTextButton
-      onClick={() => {
+  const handleClick = () => {
+    const fullPageUrl = getFullPageUrl(relation);
+
+    switch (relationOpenMode) {
+      case 'page':
+        navigate(fullPageUrl);
+        break;
+      case 'newTab':
```
jhoward1994: `window.open` doesn't seem to be a common pattern we use. Perhaps it would be
better to conditionally change the tsx and do something more consistent with the
rest of the repo?

just as an example:
```
         <IconButton
            ...
            target="_blank"
        >

```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -712,15 +720,28 @@ const RelationModalWithContext = ({
...
+                  window.open(`${getAdminBasePath()}${createUrl}`, '_blank');
+                  break;
+                case 'modal':
+                default:
+                  dispatch({
+                    type: 'GO_TO_RELATION',
+                    payload: {
+                      document: relation,
+                      shouldBypassConfirmation: false,
+                      fieldToConnect: name,
```
jhoward1994: Are we missing functionality between the different modes here? e.g. page/newTab mode will not have the same context dispatched to this action

Schero94:
Okay a lot todo for me :D.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
markkaylor: That dispatched payload is specific to the modal so the context only makes sense for the relation on the fly modal I believe

Schero94:
Thanks for the thorough review — all feedback was completely valid. Here's what I've addressed:

### @jhoward1994

**1. Import `RelationOpenMode` from shared contracts** (instead of redefining locally)

```diff
+import type { RelationOpenMode } from '../../../../../../../shared/contracts/content-types';

-type RelationOpenMode = 'modal' | 'page' | 'newTab';
```

**2. Removed `getAdminBasePath()` entirely** — by switching to React Router's `<Link>` component, the basename is resolved automatically. No need to duplicate or export `getBasename`.

**3. Replaced `window.open()` with proper React/JSX patterns** — `RelationModalTrigger` now conditionally renders a styled `<RouterLink>` for `page` and `newTab` modes:

```tsx
if (relationOpenMode === 'newTab') {
  return (
    <StyledRelationLink to={fullPageUrl} target="_blank" rel="noopener noreferrer">
      {children}
    </StyledRelationLink>
  );
}

if (relationOpenMode === 'page') {
  return <StyledRelationLink to={fullPageUrl}>{children}</StyledRelationLink>;
}

// default: modal — keeps the TextButton with dispatch
```

**4. Fixed the missing `fieldToConnect` context** — the "Create new entry" action in the Combobox now always uses modal mode, since creating a relation requires `fieldToConnect`/`fieldToConnectUID` to auto-connect back to the originating field. The `relationOpenMode` setting only affects opening **existing** relations:

```tsx
onCreateOption={() => {
  if (canCreate) {
    dispatch({
      type: 'GO_TO_RELATION',
      payload: {
        document: relation,
        shouldBypassConfirmation: false,
        fieldToConnect: name,
        fieldToConnectUID: componentUID,
      },
    });
  }
}}
```

### @markkaylor

**5. Fixed the spacing** — split the single `Grid.Root` so `gap={4}` only applies to the settings inputs. The Divider, "View" heading, and Fields are now direct Flex children, so the gap doesn't bleed into the lower section:

```diff
-<Grid.Root gap={5}>
-  <!-- inputs + divider + view heading + fields all in one grid -->
-</Grid.Root>

+<Grid.Root gap={4}>
+  <!-- only the settings inputs -->
+</Grid.Root>
+<Divider />
+<Flex direction="column" alignItems="stretch" gap={2}>
+  <Typography>View</Typography>
+  <Fields ... />
+</Flex>
```

**6. Added default value fallback** — for content types that haven't been configured with the new setting yet, the form now correctly pre-selects "Modal (default)" instead of showing an empty dropdown:

```tsx
settings: {
  ...settings,
  relationOpenMode: settings.relationOpenMode ?? 'modal',
},
```

**7. Ran `prettier:write` and `lint:fix`** — all clean. All tests pass (23/23 unit, 56/56 frontend).

packages/core/content-manager/shared/contracts/content-types.ts
```diff
@@ -10,6 +12,7 @@ export type Settings = {
   mainField: string;
   defaultSortBy: string;
   defaultSortOrder: string;
+  relationOpenMode: RelationOpenMode;
```
jhoward1994: i18n test are failing with:
 
```
Error: admin/src/contentManagerHooks/tests/listView.test.ts(16,5): error TS2741: Property 'relationOpenMode' is missing in type '{ bulkable: true; defaultSortBy: string; defaultSortOrder: string; filterable: true; searchable: true; pageSize: number; mainField: string; }' but required in type 'LayoutSettings'.
```

relationOpenMode is already correctly defaulting to 'modal' where its used so I think we can fix this by making it optional on the type?


```suggestion
  relationOpenMode?: RelationOpenMode;
```

jhoward1994:
Thanks for the great changes @Schero94 !

Schero94:
Hopefully this time it is ready to release :)

packages/core/content-manager/shared/contracts/content-types.ts
jhoward1994: Did you see my comment here @Schero94 ?

(sorry I got confused by the github review UI, I didn't see this review was still pending)

Schero94: Done! Made `relationOpenMode` optional on the `Settings` type since it's already defaulted to `'modal'` everywhere it's consumed. This fixes the i18n test TS error.

And no worries about the GitHub review UI ...  I was confused by it too, didn't see the pending review at first, I pushed it, and then saw there is a button -.-'

Schero94:
Can't wait till it's approved 😍

Schero94:
> Thanks for this @Schero94 , great work!

This will be awesome, thank you so much to everybody who was involved. As the last time, I will hyper care this feature! Just in case something we missed! :D

laurenskling:
Late to the party, but doesn't it make more sense to put the setting on the contentType you are opening? If you have 5 relations on the page, and one is huge that you never want to modal, doesn't it make sense to mark that contentType as a open, not a modal? Because now (if I read correct) it's all or nothing from the current screen, correct?

I too find myself immediately clicking the expand button a lot, even tho I love the on-the-fly feature. Sometimes it's not what you are looking for. What would actually be the very best is the click on the relation itself has 2 options: modal or link. Like clicking the text will open the modal and it has a arrow behind it that navigates to it.

Schero94:
Hey @laurenskling,  well, that’s not late, that’s the afterparty 😄
I see your point about making it configurable per relation on the content type. That definitely makes sense conceptually, especially if you have very large relations where a modal wouldn’t really be practical.
For my current use case though, I’m already quite happy with how it works right now, the main issue I was facing is solved.
And yes, I totally get what you mean about often clicking the expand button right away. The on-the-fly modal is great, but sometimes you actually just want to navigate directly. Your idea of having two interaction options (for example clicking the relation text to open the modal and having an arrow that navigates to the full page) would actually be a really nice UX improvement.
Maybe I’ll take a look at this and experiment with it sometime soon.

PR #25437, opened 2026-02-12
fix(content-manager): don't allow locale to overwrite list view settings

mvo-zyres:
### What does it do?

It fixes an issue for the persistent list view settings where the list view settings (`STRAPI_LIST_VIEW_SETTINGS`) are loaded from local storage but overwritten by the locale (`STRAPI_LOCALE`) which is also loaded from the locale storage.

### Why is it needed?

This is needed to make the feature for the persistent list view settings actually work.

### How to test it?

- Open the content manager 
- Click on a collection type 
- Set a filter
- Click on another collection type
- Now click on the original collection type again
- The list view settings (`STRAPI_LIST_VIEW_SETTINGS`) are not applied correctly, and are overwritten by the `STRAPI_LOCALE`
- It should be merged together instead of applying the last `usePersistentPartialQueryParams`

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/24246


mvo-zyres:
probably due to https://github.com/strapi/strapi/pull/24246/changes/ae30be3c8f4291abe7b61e42d29eed8ae7706479#diff-bd8e3b28ea3ba07954104f8c733c6d4d8370f9b6b310608a362da851b321907eR79

mvo-zyres:
@codex review

mvo-zyres:
Hey @Adzouz, this bug basically makes the list view settings persistence not exist :grimacing: 
It would be great if you could have a look at it :pray:

Adzouz:
Hi @mvo-zyres, thanks a lot for your fix. We'll try to take a look soon 🙏
One other thing we just figured out too: the changes are not reflected on the configure the view page (open filters and click on the configure the view button) and it's a problem 😢 When saving the configuration from this page, we should also persist the information in the local storage. Do you think you coud take a look? 🙏

https://github.com/user-attachments/assets/1780b446-6214-41e7-958e-373873ac879c

mvo-zyres:
Hey @Adzouz, thank you for your response. I just had a look and the list view headers are now saved also in the list configuration page. I also simplified the `usePersistentPartialQueryParams` to only take on configuration as `Record<Key in Localstorage, PathsToPersist>` (`Record<string, string[]>`). 

I also noticed that if e.g. you disable the filters in the list configuration page. Due to the persistence the filter silently is still applied. That is why I added also a reset to the list view settings on save. To make that work I also had to change the key of the list view settings to use the `model` instead of the `pathname`. This makes it possible to change the persistent list view settings also in other pathnames.

:warning: One thing to notice on that: This will reset current users list view settings to the default.

Would be nice if you can have a look at these changes.
Thank you! :pray:

Adzouz:
Hey @mvo-zyres, thanks a lot for the quick response! 🙏
I tested and there's still a little issue syncing the list view filters and the configure the view page (see video below):

https://github.com/user-attachments/assets/08bf555a-d400-4bbe-acb9-0b013c682a04

When a change is made from the list view, it should be reflected in the configure the view page. Do you think you could take a look at it? 🙏

mvo-zyres:
@Adzouz Yes that totally makes sense. I will take a look at it!

mvo-zyres:
Hey @Adzouz,
 I adjusted the `ListConfigurationPage.tsx` to now also read the fields from the persisted display headers. But I'm currently not sure it's the best way to do what we want. Could you review my implementation?

Adzouz:
Hey @mvo-zyres,

After looking more carefully to the code I understand your concern and I also share it.
Unfortunately using `STRAPI_LIST_VIEW_SETTINGS` to duplicate the value we have in `STRAPI_LIST_VIEW_DISPLAYED_HEADERS` in the local storage is not a solution.

There's first an issue I just figured out: `STRAPI_LIST_VIEW_SETTINGS` should be scoped and it's currently not.

Also in the configuration page, we should stick to the scoped `STRAPI_LIST_VIEW_DISPLAYED_HEADERS ` values for the displayed headers and the `STRAPI_LIST_VIEW_SETTINGS` for the rest (sort, filters, pageSize).
Clicking on Save would update those values in the local storage.

Now the tricky part is to achieve what you're trying to do with this PR. The locale shouldn't override what's defined by the other values but the current implementation can't mix scoped and not scoped values.

Do you have an idea?

mvo-zyres:
Hey @Adzouz

the current state is:
- STRAPI_LIST_VIEW_DISPLAYED_HEADERS -> scoped
- STRAPI_LIST_VIEW_SETTINGS -> global :globe_with_meridians: 
- STRAPI_LOCALE -> global :globe_with_meridians: 

and as I understand now, the expected state is:
- STRAPI_LIST_VIEW_DISPLAYED_HEADERS -> scoped 
- STRAPI_LIST_VIEW_SETTINGS -> scoped
- STRAPI_LOCALE -> global :globe_with_meridians: 

---

What's needed is then to make the STRAPI_LIST_VIEW_SETTINGS scoped. For this we could adjust the `usePersistentPartialQueryParams` config.
#### Current `usePersistentPartialQueryParams` configuration ([packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts#L23](https://github.com/zyresio/strapi/blob/ec4c1742030d4a75cf5b0b388c53700f342060fc/packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts#L23)): 
```
{
  [key: string]: string[]
}
```

#### Suggestion:
```
{
  [key: string]: {
    paths: string[];
    scoped?: boolean;
  }
}
```
or
```
{
  [key: string]: {
    paths: string[];
    scoped?: boolean;
  } | string[]
}
```

What are your thoughts?

Adzouz:
Hi @mvo-zyres,
Yes that's right 😊
And your proposition would probably do the trick, can you try and test that everything looks good please? 🙏
Just keep the default behaviour if `scoped` is not provided to avoid impacting other implementation elsewhere in the codebase

mvo-zyres:
Hey @Adzouz, i adjusted the code that it works as expected. Please have a look at it :pray: 
If it's not what we want we can just revert it again!

packages/core/admin/admin/src/hooks/usePersistentState.ts
```diff
@@ -37,14 +37,19 @@ const usePersistentState = <T>(key: string, defaultValue: T) => {
...
+};
+
 // Same as usePersistentState, but scoped to the current instance of Strapi
 // useful for storing state that should not be shared across different instances of Strapi running on localhost
 const useScopedPersistentState = <T>(key: string, defaultValue: T) => {
-  const { data: initData } = useInitQuery();
-  const { uuid } = initData ?? {};
+  const uuid = usePersistentStateScope();
 
   const namespacedKey = `${key}:${uuid}`;
```
Adzouz, line 51: I was wondering, what's happening if uuid wasn't found? Should `namespacedKey` be only `key` instead?

Adzouz (commented):
Overall I think it's good, I tested and it looks fine. I have just one tiny comment 😊
I'll also ask another person from the team to double check it, just to be sure 🙏

Adzouz:
Actually, I got an issue I'm sorry 😢

https://github.com/user-attachments/assets/dfc72975-679d-4646-a41a-e5d4f9c4e663

If I change the sorting and go back to the list, it's overridden...
There's a sync issue between the sorting in the headers and the configuration page. All should be setting the local storage settings variable the same way and be synced. Going back to the list after changing the value should keep the value.

https://github.com/user-attachments/assets/cd85cb44-9b59-48e8-a8c4-982ccce229a6

mvo-zyres:
Hey @Adzouz,

sorry for the long delay. :pray: 
Just had a lot going on and couldn't get to it yet. 

I checked it and it's because the "Back" button uses the previous URL.
This URL still has the previous sorting in and will overwrite it again.

Recommendation:
Remove the sorting sync from the "Configure the view" page. The sorting there only defines the default values and not the current selected ones. Everything else would still work the same. I think this would also make the behavior more clear.

Let me know if we want to proceed with this approach :slightly_smiling_face:

Adzouz:
Hi @mvo-zyres,
Sorry also for the delay (been a bit crazy lately on our side).
Ok I got it. The thing is that this value and the one when you sort directly in the list are the same in the local storage but as you said, it's two different things (one if the default sort and the other is the column the user decided to sort on the list view). I would be better to have both separated.
The default sort order is taken only if it wasn't overridden by manual choice of the user on the list view.
Do you think you can fix that? 🙏

mvo-zyres:
@Adzouz yes I can do that! Should be quite simple :playground_slide:

mvo-zyres:
Hey @Adzouz, one more thing. The pageSize is also synced in the ListConfigurationPage, but when going back it would also be overwritten by the back link. How should we handle this?

Adzouz:
Hey @mvo-zyres, for the pageSize it's slightly different because it's supposed to be stored in the database. Using the back button wouldn't change this value. But after testing again, I think your PR changes this behaviour and now the two pageSize and synced in the localStorage and it shouldn't. We should go back to what we had before if you can also fix that 🙏
- pageSize in configuration page: set in database
- pageSize in list view: set in local storage
-> If local storage value exist, then take it otherwise take the pageSize defined in the database (the one from the configuration page)

mvo-zyres:
Hey @Adzouz, its now all adjusted accordingly, the pageSize and sort in the list configuration page now only save to the databse (as before). That makes the UX 100x better then saving it also to the local storage. I hope now everything is as expected and the PR can be finalized!

Adzouz:
Hey @mvo-zyres, I tested and indeed it's way better like that, thanks a lot 🙌
I'll double check everything, run the pipelines and review the whole code changes again and I really hope that this time we will be able to merge it 🙏

packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts
```diff
@@ -21,38 +25,63 @@ const filterObjectKeys = (obj: object, keys: PropertyPath[]) => {
...
-    if (Object.keys(parsedSavedParams).length === 0) return;
 
-    const filteredQuery = filterObjectKeys(parsedSavedParams, keysToPersist);
-    setQuery({ ...filteredQuery, ...query }, 'push', true);
+    if (Object.keys(mergedFilteredQuery).length === 0) return;
+
+    setQuery({ ...mergedFilteredQuery, ...query }, 'push', true);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [localStorageKey]);
+  }, [config, scope]);
```
Adzouz: I just saw that a new object reference would trigger the useEffect if the memoization breaks. To get rid of this unstable dependency, you could serialize config to a more stable const like `const clonedConfig = JSON.stringify(config);` and use `clonedConfig` as dependency of the useEffect instead (also for the second useEffect right after).

Adzouz (changes requested):
Just a tiny comment to address if possible 🙏

mvo-zyres:
Hey @Adzouz, I adjusted it to make sure that there are no unnecessary useEffect calls due to reference changes. Hope that we can merge it soon. Thank you!

Adzouz:
Hey @mvo-zyres thanks a lot for the changes 🙌 On my side it looks good, I'll just ask @jhoward1994 to also take a look because he opened a PR that will need the changes you made here 👀 I hope we'll be able to merge it asap

jhoward1994:
Hi @mvo-zyres ! thanks for this, your changes look great. 

My PR https://github.com/strapi/strapi/pull/26005 would conflict with this as @Adzouz said but I think they both have valid changes that we want. I'm happy to go ahead with this PR and then adjust and deal with the conflicts in mine after it's merged 👍🏻

jhoward1994 (approved):

Adzouz (approved):
Let's goo 🚀 Thanks a lot for the hard work again and your patience @mvo-zyres 🙏

PR #25439, opened 2026-02-12
enhancement(admin): allow custom field option types for CTB form components

Link2Twenty:
### What does it do?

Strapi allows us to build our own custom form components for Content-Type Builder this is amazing but unfortunately the types aren't happy to let us do it so we have to sprinkle in a load of `any`s.

image

**Admin Register**

```ts
app.getPlugin('content-type-builder').apis.forms.components.add({
  id: 'fake-form',
  component: CustomField
})

app.customFields.register({
  name: 'bloc-landscape-2d',
  type: 'text',
  icon: Icon,
  intlLabel: {
    id: getTranslation('editor.label'),
    defaultMessage: 'Landscape 2D',
  },
  intlDescription: {
    id: getTranslation('editor.description'),
    defaultMessage: 'Create and manage 2D landscape experiences.',
  },
  components: {
    Input: () => import('./pages/Field'),
  },
  options: {
    advanced: [
      {
        intlLabel: {
          id: getTranslation('options.advanced.label'),
          defaultMessage: 'Field Label.',
        },
        description: {
          id: getTranslation('options.advanced.description'),
          defaultMessage: 'Field Description.',
        },
        name: 'options.fake-form-1' as any,
        type: 'fake-form' as any,
      },
    ],
  },
});
```

**Custom Component**

```tsx
import { Field, Flex, Typography } from '@strapi/design-system';

interface CustomFieldProps {
  intlLabel?: { id?: string; defaultMessage?: string };
  description?: { id?: string; defaultMessage?: string };
}

export const CustomField = ({ intlLabel, description }: CustomFieldProps) => {

  return (
    <Flex direction="column" alignItems="stretch" gap={2}>
      {description && (
        <Typography variant="pi" textColor="neutral600">
          {formatMessage(description)}
        </Typography>
      )}
      <Field.Root
        hint={formatMessage({ id: 'custom-field.hint', defaultMessage: 'This is a hint' })}
      >
        {intlLabel && <Field.Label>{formatMessage(intlLabel)}</Field.Label>}
        <input type="range" defaultValue={20} min={0} max={100} />
        <Field.Hint />
      </Field.Root>
    </Flex>
  );
};
```

### Why is it needed?

If we're offering awesome power features like this it'd be nice if the types told us it was possible.

### How to test it?

N/A

### Related issue(s)/PR(s)

N/A

As an aside, I don't see this documented anywhere but it's a really powerful feature.


innerdvations:
@nclsndr Do you think it's ok to accept this as a temporary fix before we add a stronger fix such as #26817 ?

nclsndr:
I guess I'm lacking context here. I'd need to reproduce locally to see the delta.

packages/core/admin/admin/src/core/apis/CustomFields.ts
```diff
@@ -37,7 +40,8 @@ type CustomFieldOptionName =
   | 'enum'
   | 'unique'
   | 'private'
-  | 'default';
+  | 'default'
+  | `options.${string}`;
```
nclsndr, line 44: I'll need context. I understand the `${string}` widening but why `options.` is chosen to be the prefix?
Do custom field options are nested like those @innerdvations?

packages/core/admin/admin/src/core/apis/CustomFields.ts
```diff
@@ -206,7 +210,8 @@ const optionsValidationReducer = (
   } else {
     acc.push({
       isValidOptionPath:
-        option.name.startsWith('options') || ALLOWED_ROOT_LEVEL_OPTIONS.includes(option.name),
+        option.name.startsWith('options') ||
+        (ALLOWED_ROOT_LEVEL_OPTIONS as readonly string[]).includes(option.name),
```
nclsndr: I think the casting should be on `option.name`. Rationale: runtime narrowed on `option.name.startsWith('options')`
```suggestion
        ALLOWED_ROOT_LEVEL_OPTIONS.includes(
          option.name as Exclude<CustomFieldOptionName, `options.${string}`>
        ),
```

packages/core/admin/admin/src/core/apis/CustomFields.ts
innerdvations: Yeah, custom option values are nested, it's an existing contract afaik

packages/core/admin/admin/src/core/apis/CustomFields.ts
innerdvations: Yeah, you're right, that's better

innerdvations: Pushed in d5e119ba0e — casting `option.name` as suggested.
_AI-assisted comment._

innerdvations:
I've been thinking about it this morning, and I think it may be better to just push through with [#26817](https://github.com/strapi/strapi/pull/26817), which fully supersedes this one, rather than merging this and then having to violate the type contract later by narrowing string.

If you all agree, I'll close this one and we can move over there. @nclsndr

Link2Twenty:
> I've been thinking about it this morning, and I think it may be better to just push through with [#26817](https://github.com/strapi/strapi/pull/26817), which fully supersedes this one, rather than merging this and then having to violate the type contract later by narrowing string.
> 
> If you all agree, I'll close this one and we can move over there. @nclsndr

I know I don't really get a say but I think your solution is safer than mine, if very slightly more complex, though I'd love to see a documentation update to go along with it. I think it makes sense to close this and focus on getting the real solution out.

Issue #25178, opened 2026-01-13
Non-localized media fields disappear when creating localizations via REST API


gl-guccione:
### Node Version

22.21.1

### NPM/Yarn/PNPM Version

10.9.4

### Strapi Version

5.30.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Description

When creating a localized version of an entry via PUT request following the official documentation, non-localized media fields are being cleared from all localizations if not explicitly included in the request body. This contradicts the documentation which states that non-localized fields should not be passed when creating localizations.

The admin panel works correctly when creating localizations manually through the UI, so this appears to be an API-specific issue.

### Environment

* **Strapi version**: 5.3.0
* **Node.js version**: 22
* **Upload provider**: @strapi/provider-upload-aws-s3 version 5.2.3
* **Database**: Aurora PostgreSQL
* **Infrastructure**: AWS EKS
* **API**: REST API (standard endpoints)
* **Pod configuration**:
  * Memory: request 1024Mi, limit 4096Mi
  * CPU: request 200m, limit 4000m
  * HPA: min 1, max 4 pods

### Affected Collections

* `product` (with draft & publish enabled)
* `technologies` (without draft & publish)
* Both are localized with non-localized media fields

### Case 1: Collection with Draft & Publish (`product`)

**Collection structure:**

* Localized collection with draft & publish enabled
* **Non-localized fields**: `code` (string), `productName` (string), `detailImages` (media list)
* **Localized fields**: `description` (markdown), `materials` (relation many-to-many)

**Steps:**

1. Create a product (e.g., XC342 in Italian locale) via POST API with all fields including `detailImages` (4 images)
   * ✅ Works correctly
2. Get the `documentId` of the created product
3. Create an English localization via PUT request following the [official documentation](<https://docs.strapi.io/cms/api/rest/locale#rest-update>)
4. Pass only localized fields in the body: `description` and `materials`
5. Do NOT pass `detailImages` as per documentation warning

**Result:**

* `detailImages` disappear from the API response for **all localizations**
* In the admin panel (Italian locale - original):
  * **Draft version**: images are still visible in the `detailImages` field
  * **Published version**: images are missing (visible in the "published" tab)
  * No indication that the entry needs to be republished (mismatch between draft and published)
* In the admin panel (English locale - newly created):
  * Images are missing in both draft and published versions

[...]

PR #25442, opened 2026-02-12
chore(i18n): add TODO for non-localized field investigation

MehakSetia:
Draft PR to verify the fix for Issue #25178. I have identified the missing logic in the Document Service middleware and added a TODO marker to track the implementation.


MehakSetia:
I've updated the clone logic to ensure non-localized fields (like media and relations) are preserved. The service now fetches the sourceDocument and merges its data with the new body before cloning.

Manual Verification:
I ran the internal unit tests for the clone utility in a clean environment, and all tests passed:

PASS packages/core/content-manager/server/src/controllers/utils/__tests__/clone.test.ts

Tests: 8 passed, 8 total (including 'model with media' and 'model with relations').

innerdvations:
Thanks for digging into the non-localized fields area — closing this one because it isn’t the right shape for us to take forward.

The title/body frame this as a TODO/chore, but the diff changes `document-manager` clone behaviour (extra `findOne`, spreading the full source document) rather than landing a narrow investigation note. That approach would need a different design and would be easy to get wrong for i18n/draft flows, so it isn’t something we can iterate on productively from this PR as written.

If you’re still interested in the underlying issue, a better next step is a focused bug report (or a small PR that only documents the open question without changing clone semantics). Appreciate the time you put into it.

_AI-assisted comment._

PR #25443, opened 2026-02-12
fix: some fields not updating or interactable in live preview

alex-strapi:
### What does it do?

This PR improves the Live Preview integration for media and rich content fields by:

- `Enabling` `STRAPI_FIELD_CHANGE` **events for media fields** in the `usePreviewInputManager` hook so image and video updates are propagated immediately to the preview (optimistic updates).
- **Removing the exclusion of media field types** from change events while preserving exclusions for unsupported/internal field types.
- **Improving preview element targeting** by moving the `data-strapi-source` attribute directly onto rendered `` and `<video>` elements, allowing the preview script to correctly identify, focus, and scroll to media elements.
- **Fixing Block Renderer handling** to ensure structured rich content updates properly trigger preview synchronization and element mapping.
- **Adding comprehensive unit tests** for `usePreviewInputManager`, covering:
    - Focus events
    - Blur events
    - Change events
    - Excluded field types
    - Media field change propagation

### Why is it needed?

Previously:

- Media fields (images and videos) did not trigger change events, meaning the Live Preview only updated after saving.
- Clicking an image or video in the preview did not correctly focus the corresponding field in the Content Manager.
- The preview script could not reliably locate media elements because `data-strapi-source` was not attached directly to `` or `<video> `elements.
- Some structured fields (including Block Renderer content) did not consistently synchronize with the preview during optimistic updates.

This resulted in an inconsistent and delayed editing experience.

This PR ensures:

- Media fields update immediately in Live Preview.
- Clicking media in the preview correctly focuses/highlights the related field.
- Block renderer fields behave consistently.
- The overall preview experience is smoother and closer to true real-time editing.

### How to test it?

**Manual verification**

1. Open a content type (e.g., Kitchensink or any custom type with media fields).
2. Open Live Preview.
3. Change an image or video field. 
    - The media updates immediately in the preview (without saving).
4. Click the image or video inside the preview.
    - The corresponding field in the Content Manager is focused/highlighted.
5. Update content inside a Block Renderer field.
    - The preview updates correctly.
6. Verify focus behaviour works for nested/component fields.

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -68,11 +68,7 @@ const NestedValue = ({
 
   // Use the email type to test clickable preview elements
   if (fieldName === 'email') {
-    return (
-      <Button variant="tertiary" onClick={() => window.alert('Sending email!')}>
-        {value}
-      </Button>
-    );
+    return <Typography>{value}</Typography>;
   }
 
   if (isMedia(value)) {
--- a/examples/getstarted/src/admin/preview/dummy-preview.jsx
+++ b/examples/getstarted/src/admin/preview/dummy-preview.jsx
@@ -6,10 +6,6 @@ import { Page, Layouts } from '@strapi/admin/strapi-admin';
 import { Grid, Flex, Typography, JSONInput, Box, Button } from '@strapi/design-system';
 import { ChevronDown, ChevronRight } from '@strapi/icons';
 
-const isMedia = (value) => {
-  return typeof value === 'object' && value !== null && 'url' in value;
-};
-
 const filterAttributes = (item) => {
   const excludedKeys = ['documentId', 'id', 'createdAt', 'updatedAt', 'publishedAt'];
   return Object.entries(item).filter(([key]) => !excludedKeys.includes(key));
@@ -49,6 +45,10 @@ const ToggleableContainer = ({ headerText, children, isEmpty = false }) => {
   );
 };
 
+const isMedia = (value) => {
+  return typeof value === 'object' && value !== null && 'url' in value;
+};
+
 const NestedValue = ({
   value,
   level = 0,
@@ -60,14 +60,7 @@ const NestedValue = ({
 }) => {
   if (fieldName === 'blocks') {
     return (
-      <Flex
-        direction="column"
-        alignItems="flex-start"
-        fontSize="1.4rem"
-        gap={2}
-        data-strapi-source={`path=${path}&type=blocks&model=${model}&documentId=${documentId}`}
-        width="100%"
-      >
+      <Flex direction="column" alignItems="flex-start" fontSize="1.4rem" gap={2}>
         {value ? <BlocksRenderer content={value} /> : 'null'}
       </Flex>
     );
@@ -76,55 +69,40 @@ const NestedValue = ({
   // Use the email type to test clickable preview elements
   if (fieldName === 'email') {
     return (
-      <Button
-        variant="tertiary"
-        onClick={() => window.alert('Sending email!')}
-        data-strapi-source={`path=${path}&type=email&model=${model}&documentId=${documentId}`}
-      >
+      <Button variant="tertiary" onClick={() => window.alert('Sending email!')}>
         {value}
       </Button>
     );
   }
 
-  if (typeof value === 'boolean') {
-    return (
-      <Box
-        data-strapi-source={`path=${path}&type=boolean&model=${model}&documentId=${documentId}`}
-        padding={1}
-        background={value ? 'success100' : 'danger100'}
-        borderColor={value ? 'success200' : 'danger200'}
-        borderWidth="1px"
-        hasRadius
-      >
-        <Typography variant="sigma" textColor={value ? 'success600' : 'danger600'}>
-          {String(value).toUpperCase()}
-        </Typography>
-      </Box>
-    );
-  }
-
   if (isMedia(value)) {
-    // Check for video mime type if available, otherwise assume image for preview if url exists
-    const isVideo = value.mime?.startsWith('video/');
-    const sourceString = `path=${path}&type=media&model=${model}&documentId=${documentId}`;
+    const url = value.url;
+    const mime = value.mime;
+    const isVideo = mime?.startsWith('video/');
+    const isImage = mime?.startsWith('image/') || (!mime && !!url);
+    const fileName = value.name || 'file';
 
     return (
       <Box style={{ maxWidth: '300px' }}>
-        {isVideo ? (
-          <video
-            src={value.url}
-            controls
-            style={{ maxWidth: '100%' }}
-            data-strapi-source={sourceString}
-          />
-        ) : (
-          <img
-            src={value.url}
-            alt={value.alternativeText || ''}
-            style={{ maxWidth: '100%' }}
-            data-strapi-source={sourceString}
-          />
+        {isImage && (
+          <>
+            <img
+              src={isImage ? url : undefined}
+              alt={value.alternativeText || ''}
+              style={{ maxWidth: '100%', display: isImage ? 'block' : 'none' }}
+            />
+          </>
         )}
+        {isVideo && (
+          <>
+            <video
+              src={isVideo ? url : undefined}
+              controls
+              style={{ maxWidth: '100%', display: isVideo ? 'block' : 'none' }}
+            />
+          </>
+        )}
+        {!isImage && !isVideo && <Typography>{fileName}</Typography>}
       </Box>
     );
   }
@@ -202,11 +180,7 @@ const NestedValue = ({
 
   return (
     <Box>
-      <Typography
-        data-strapi-source={`path=${path}&type=text&model=${model}&documentId=${documentId}`}
-      >
-        {value === null || value === undefined || String(value) === '' ? '\u00A0' : String(value)}
-      </Typography>
+      <Typography>{String(value)}</Typography>
     </Box>
   );
 };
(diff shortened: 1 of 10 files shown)


Adzouz:
Hi @alex-strapi and thank you so much for the contribution 🙏 Going to take a look at the code but first, here's what I can see while testing the feature:

1. In the internal develop view we made for debugging, it would be nice to have null default values so we can edit the values by double clicking on them. Since there's no value at all, it's harder to see where to hover to edit them
image


2. When we first select an asset for a media, there's a [object Object] displayed but it would be better to have the actual media displayed (image / video) -> when we save, it's working and the asset appears
image

3. When you double click to edit an asset, the popover stays above the media asset replacement modal but should stay behind
image

4. And finally you have the same issue we encountered in [Remi's PR](https://github.com/strapi/strapi/pull/25216), if you replace a media type by another (let's say a video to an image) then click Save, you have this error showing up
image

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -55,10 +49,25 @@ const ToggleableContainer = ({ headerText, children, isEmpty = false }) => {
...
+}) => {
   if (fieldName === 'blocks') {
     return (
-      <Flex direction="column" alignItems="flex-start" fontSize="1.4rem" gap={2}>
+      <Flex
+        direction="column"
+        alignItems="flex-start"
+        fontSize="1.4rem"
+        gap={2}
+        data-strapi-source={`path=${path}&type=blocks&model=${model}&documentId=${documentId}`}
```
Adzouz: We shouldn't rely on those data-strapi-source attributes. We have to keep in mind that this example page is only for debugging and that this property won't exist in real life frontend environments. The targeting of elements in the dom is made with stega encoded characters directly added to the value (they contain all those info - path, documentId etc) and it should be enough to find your element in the page 😊 Only thing is that we need to attach the stega string to an actual string... And that's why for now we only had text handled in the live preview. Maybe you could try to find a way to get rid of those properties and use the logic already in place to be able to find your elements in the page (cf https://contributor.strapi.io/docs/core/content-manager/preview for more info)

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -67,12 +76,59 @@ const NestedValue = ({ value, level = 0, arrayIndex = undefined, fieldName = und
   // Use the email type to test clickable preview elements
   if (fieldName === 'email') {
     return (
-      <Button variant="tertiary" onClick={() => window.alert('Sending email!')}>
+      <Button
```
Adzouz: It's weird that it's a button tbh because it prevents the field from being edited. I would actually remove this condition for email since it's basically just a text to display, it can go to the default usecase of this component

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -67,12 +76,59 @@ const NestedValue = ({ value, level = 0, arrayIndex = undefined, fieldName = und
...
+        variant="tertiary"
+        onClick={() => window.alert('Sending email!')}
+        data-strapi-source={`path=${path}&type=email&model=${model}&documentId=${documentId}`}
+      >
         {value}
       </Button>
     );
   }
 
+  if (typeof value === 'boolean') {
```
Adzouz: I think this one will be tricky to catch in a frontend environment because I feel like it wouldn't be displayed. I don't know if it's really relevant, what do you think? Since stega encoding isn't working because there's no string to attach the content to, it would be also pretty hard to be able to target it without the data-strapi-source. In that case, it would only work in the dummy-preview

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -67,12 +76,59 @@ const NestedValue = ({ value, level = 0, arrayIndex = undefined, fieldName = und
...
+      <Box style={{ maxWidth: '300px' }}>
+        {isVideo ? (
+          <video
+            src={value.url}
+            controls
+            style={{ maxWidth: '100%' }}
+            data-strapi-source={sourceString}
+          />
+        ) : (
+          <img
```
Adzouz: In the case of any other type of file here, there's nothing displayed in the page (which can make sense in a frontend app). But here, since only video and images are handled, any other file won't work. Maybe we could at least show the name of the file and be sure that it's possible to target it to be able to change it again?

examples/getstarted/src/admin/preview/dummy-preview.jsx
```diff
@@ -133,12 +202,16 @@ const NestedValue = ({ value, level = 0, arrayIndex = undefined, fieldName = und
 
   return (
     <Box>
-      <Typography>{String(value)}</Typography>
+      <Typography
+        data-strapi-source={`path=${path}&type=text&model=${model}&documentId=${documentId}`}
+      >
+        {value === null || value === undefined || String(value) === '' ? '\u00A0' : String(value)}
```
Adzouz: That's maybe where I'd keep `null` instead of a nbsp for better visibility

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -564,7 +793,15 @@ const previewScript = (config: PreviewScriptConfig) => {
         highlightManager.setFocusedField(field);
         getElementsByPath(field).forEach((element, index) => {
           if (index === 0) {
-            element.scrollIntoView({ behavior: 'smooth', block: 'center' });
+            // Check if scrollIntoViewIfNeeded is available (e.g. Chrome)
+            if (
+              'scrollIntoViewIfNeeded' in element &&
+              typeof (element as any).scrollIntoViewIfNeeded === 'function'
+            ) {
+              (element as any).scrollIntoViewIfNeeded({ behavior: 'smooth', block: 'center' });
```
Adzouz, line 1117: Wow I didn't know about this one 😄

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -69,7 +69,106 @@ const previewScript = (config: PreviewScriptConfig) => {
...
+
+    const renderChildren = (children: any[]): string => {
+      if (!Array.isArray(children)) return '';
+
+      return children
+        .map((child) => {
+          if (!child) return '';
+          let text = child.text || '';
+
+          if (child.bold) text = `<strong>${text}</strong>`;
```
Adzouz: In a frontend app I guess this part would be automatically handled no? (more a question here). If so, maybe moving this logic in the dummy page template?

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
```diff
@@ -540,11 +677,103 @@ const previewScript = (config: PreviewScriptConfig) => {
...
+                  newElement.style.cssText = element.style.cssText;
+                  newElement.style.display = '';
+
+                  // For video elements, add common attributes
+                  if (newTagName === 'VIDEO') {
+                    newElement.setAttribute('controls', '');
+                  }
+
+                  // Replace the old element with the new one
+                  element.parentNode?.replaceChild(newElement, element);
```
Adzouz: I think that's here we had the issue when you replace a media type to another (e.g. an image to a video) and click on save, it's trying to replace the item that was already replaced when you select the new item in the asset modal. This method is executed twice and it shouldn't be executed again on save. But we couldn't figure out yet how to prevent this from happening :(

Adzouz (changes requested):
I'm sorry for the harsh review but unfortunately this won't work (as we already discussed together) and need refinement 😅 We should get rid of the data attributes and use stega encoding to target elements in the page if possible. The need of an attribute should be discussed only if there's no other way to achieve it.
Feel free to reach out to me if you need anything and I hope you see things clearer now you have context 🙏
Don't forget there's the documentation available here: https://contributor.strapi.io/docs/core/content-manager/preview

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: These have now been removed. 
I've implemented the stega solution for the blocksrender.

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: This was actually in the original dummy-preview before my edit. But It's now just text/typography.

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: I'd love to do something clever here, like understand from the dom what is being conditionally rendered, so the live preview reflects in real time to display/change the view based on user input. 

But I see your point and I've removed it for now. I wonder how sanity handle this case 🤔

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: Media does work (selects), it's just the bug Remi ran into preventing it from being implemented. Working on that now.

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: Implemented, I wonder on a blank content type if "Live preview" is something we want to allow since they won't have anything to target. Maybe we could use the stega targeting to "paint" something like an outlined border or little edit widget where there is no value. 

But for now in the dummy-preview I've output null instead of nbsp.

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
alex-strapi: Cleaned up!

packages/core/content-manager/admin/src/preview/utils/previewScript.ts
alex-strapi: Seemed like an obvious enhancement 🤷

examples/getstarted/src/admin/preview/dummy-preview.jsx
alex-strapi: Think I solved the bug Remi had, can you please check again?

Adzouz:
Hey @alex-strapi, I'll close this for now since we already implemented the preview on images and videos (PR #25216). I know we didn't implemented yet on blocks but we planned to do it in the future. Feel free to reach out to us if you're interested to know more or work on it but on a new base 🙏😊 Thanks so much for your effort and taking care of this subject!

PR #25447, opened 2026-02-13
enhancement: extended ctb api for plugins

boazpoolman:
### What does it do?

This PR extends the CTB api to be a bit more flexible for plugin content-types.
1. It allows _new_ plugin content types to be registered in `src/extensions` instead of just altering existing content types.
2. It allows to create new plugin content types using the CTB services.

```
const schemaService = strapi.plugin("content-type-builder").service("schema");
await schemaService.updateSchema(schema);
```

### Why is it needed?

Certain situations require plugins to register content types based on the plugin configuration. Currently there is no way to register those content types dynamically from a plugin. My solution to that problem is to put new plugin content types in the `src/extensions` folder, just like altered content-types. It's then up to the plugin developer to decide how and when they want to create those new content types. For example the plugin could expose a CLI that creates the schema. Much like better auth does with their [generate command](https://www.better-auth.com/docs/concepts/cli#generate).

### How to test it?

1. Pull this branch and start up Strapi
2. Create a new content-type for an existing plugin by manually putting it in the `src/extensions` folder
3. Try programatically creating a new content type for an existing plugin by calling the CTB service.

```
const schemaService = strapi.plugin("content-type-builder").service("schema");
await schemaService.updateSchema(schema);
```


innerdvations:
I'm pretty sure we will go with this method, but I need to meet discuss this with some others to be sure!

nclsndr:
Side note: would we have to check how it would impact the AI CTB chat as well?

innerdvations (commented):
Ok, LGTM, just waiting for an approval from either @Bassel17 or @jhoward1994 and we'll merge.

innerdvations:
> Side note: would we have to check how it would impact the AI CTB chat as well?

Looked into this, and yes, AI could potentially have problems with this. I just pushed a fix if someone can take a look.

packages/core/content-type-builder/server/src/controllers/validation/schema.ts
```diff
@@ -682,6 +682,7 @@ const baseContentTypeSchema = z.object({
 
 const baseCreateContentTypeSchema = baseContentTypeSchema.extend({
   action: z.literal('create'),
+  plugin: z.string().optional(),
```
jhoward1994:
```suggestion
  plugin: z.string().min(1).optional(),
```

should we avoid scenarios like this  `plugin::.foo` ?

jhoward1994 (approved):

boazpoolman:
Shall we merge this PR in order to be released with 5.44.0?

jhoward1994:
Hey @boazpoolman - sorry we missed this weeks release I was off the last couple of days but yes happy to merge and it'll be in 5.44.1. I'll leave the merge to you in case there are any tweaks you want to make

PR #25263, opened 2026-01-27
feat(upload-aws-s3): add extended configuration for S3-compatible providers

Schero94:
### What does it do?

**Dependency Updates:**
- Updates `@aws-sdk/client-s3` from 3.600.0 → 3.975.0
- Updates `@aws-sdk/lib-storage` from 3.433.0 → 3.975.0
- Updates `@aws-sdk/s3-request-presigner` from 3.433.0 → 3.975.0
- Updates `@aws-sdk/types` from 3.433.0 → 3.973.0

**New Features:**
- Adds support for S3-compatible storage providers (IONOS, MinIO, DigitalOcean Spaces, Backblaze B2, etc.)
- Adds checksum validation options (CRC32, CRC32C, SHA1, SHA256, CRC64NVME) for data integrity
- Adds conditional writes to prevent overwrites (IfNoneMatch)
- Adds optimistic locking with ETag support (uploadIfMatch method)
- Adds configurable ACL settings (public-read, private, bucket-owner-full-control, etc.)
- Adds StorageClass configuration (STANDARD, REDUCED_REDUNDANCY, INTELLIGENT_TIERING, etc.)
- Adds server-side encryption options (AES256, SSE-KMS, DSSE-KMS with custom key ID)
- Adds object tagging for cost allocation and lifecycle policies
- Adds multipart upload configuration (partSize, queueSize, leavePartsOnError)
- Adds getObjectMetadata and objectExists helper methods
- Fixes URL construction for non-AWS endpoints that return incorrect Location headers

**Security:**
- Adds path traversal prevention in file keys
- Adds parameter injection protection in customParams
- Adds URL protocol validation (HTTP/HTTPS only)

**Documentation:**
- Adds comprehensive JSDoc documentation for all configuration options
- Adds configuration examples and security considerations to README

**Tests:**
- Adds 62 unit tests covering all new features and security fixes

### Why is it needed?

1. **Outdated dependencies:** The AWS SDK packages were significantly outdated (3.433.0 from 2023), missing important bug fixes, performance improvements, and new features.

2. **S3-compatible providers:** The current provider only works reliably with AWS S3. Many users want to use S3-compatible storage providers like IONOS, MinIO, or DigitalOcean Spaces, which require custom endpoint configuration and different URL construction.

3. **Missing features:** Users have requested data integrity validation, fine-grained access control, cost optimization via StorageClass, and enhanced security via server-side encryption.

[...]

Issue #2949, opened 2019-03-08
withRelated typo issue in search Bookshelf template


absolux:
Hello, I'm not sure yet, but I think there's a typo in `bookshelf/service.template`

**What is the current behavior?**

```js
model.fetchAll({
  width: populate
  ^^^^^
})
```
https://github.com/strapi/strapi/blob/f2bac2a59abfb0d05ddda3883b854718614245a7/packages/strapi-generate-api/templates/bookshelf/service.template#L239:L241

**Suggested solutions**

```js
model.fetchAll({
  withRelated: populate
})
```

lauriejim:
Hello @absolux !

Yap it look like to be an issue.
Feel free to submit a PR to fix this. We will appreciate your contribution on this point.

PR #25449, opened 2026-02-13
docs: simplify README for AWS S3 provider updates

pwizla:
### What does it do?

I vastly simplified the README for the S3 provider, keeping it to a bare minimum.

I had to significantly rework the recently added content (extended configuration from PR #25263 ) so that it fits in the official docs page. So I'm removing all the details from here and pointing to [the docs page](https://docs.strapi.io/cms/configurations/media-library-providers/amazon-s3) which must remain the single source of truth 🤓

### Why is it needed?

- Makes official docs maintenance easier
- Avoid content duplication

### Related issue(s)/PR(s)

#25263 
[docs #2949](https://github.com/strapi/documentation/pull/2949) which adds all the details there


markkaylor:
@pwizla 

The commitlint isn't happy. Would you be able to amend your commit so it respects the conventional commit format with something like `chore(docs): <message>`.

pwizla:
Thanks for the feedback, @markkaylor. I was unsure how to handle the failing test. 
Honestly I feel clueless about why the commit linter keeps on failing 🤔
~~The commit message now reads `chore(docs): Revise README for AWS S3 provider updates`. What else should I change?~~

Edit: found it. No sentence case. ✅

PR #25451, opened 2026-02-13
ci: add version compare script and workflow

innerdvations:
### What does it do?

- Adds `yarn version:check` script that scans the monorepo for version mismatches in our own package.json `version` property as well as all dependencies and devDependencies they use
- Adds `yarn version:fix` to update and sync them all using `syncpack`
- fixes all existing misalignments
- adds a github workflow gate for PRs to prevent adding new misalignments

### Why is it needed?

- prevents subtle and extremely difficult to debug bugs caused by package mismatches
- reduces total package size
- makes updates and maintenance easier

### How to test it?

### Related issue(s)/PR(s)

Ideally, we will be able to switch to `workspace` and yarn/pnpm catalogs when we update the automated release process and then remove all this; this is a quick temporary measure until that time


innerdvations:
I need to check if it's safe to update `typescript`

Update: oops, it was only docs being updated, so it's fine

scripts/check-package-versions.mjs
nclsndr, line 1: Could we use `.mjs` and move the implementation to ESM?
Rationale: one less file to migrate once the monorepo migrate.

scripts/check-package-versions.js
```diff
@@ -0,0 +1,121 @@
...
+// Skip examples/test-apps/docs that use 0.0.0 or 0.1.0.
+function shouldEnforceVersion(currentVersion, canonicalVersion) {
+  const canonicalMajor = canonicalVersion.split('.')[0];
+  const currentMajor = currentVersion.split('.')[0];
+  return currentMajor === canonicalMajor;
+}
+
+async function main() {
+  const fix = process.argv.includes('--fix');
+  const promptEach = process.argv.includes('--prompt');
```
nclsndr: Cosmetics (for better reading).
```suggestion
  const hasFixArg = process.argv.includes('--fix');
  const hasPromptArg = process.argv.includes('--prompt');
```

PR #25453, opened 2026-02-14
fix(content-manager): skip non-draftAndPublish relations in countDraftRelations

veeceey:
## Problem

`countDraftRelations` crashes with a 400 error when a content type has relations to other content types that don't have draft & publish enabled. The error happens because `getDeepPopulateDraftCount` applies a `publishedAt: { $null: true }` filter to ALL relation targets, but non-D&P types don't have a `publishedAt` attribute.

This means any content type with even a single relation to a non-D&P type will throw an error every time a user opens the edit view or tries to publish.

## Root cause

In `packages/core/content-manager/server/src/services/utils/populate.ts`, the `getDeepPopulateDraftCount` function iterates over all relation attributes and adds a `publishedAt` filter without checking whether the target content type actually supports draft & publish:

```ts
case 'relation': {
  // ...
  if (isVisibleAttribute(model, attributeName)) {
    populateAcc[attributeName] = {
      count: true,
      filters: { [PUBLISHED_AT_ATTRIBUTE]: { $null: true } },  // crashes for non-D&P targets
    };
  }
}
```

## Fix

Added a check using `hasDraftAndPublish()` on the target model before adding the `publishedAt` filter. If the target doesn't support D&P, the relation is skipped entirely since non-D&P content types can't have drafts by definition.

## Test plan

- Added a `Tag` content type with `draftAndPublish: false` to the existing test suite
- Added a `tag` relation from `Product` to `Tag`
- Added test that creates a product with a tag relation and verifies `countDraftRelations` returns 200 with a count of 0 (instead of crashing)

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -10,12 +10,27 @@ body:
     validations:
       required: true
 
+  - type: dropdown
+    id: packageManager
+    attributes:
+      label: Package Manager
+      description: Which package manager are you using?
+      multiple: false
+      options:
+        - npm
+        - yarn
+        - pnpm
+        - bun
+        - Other
+    validations:
+      required: true
+
   - type: input
     id: pmVersion
     attributes:
-      label: NPM/Yarn/PNPM Version
-      description: The package manager and version you are using.
-      placeholder: NPM 10.0.0
+      label: Package Manager Version
+      description: The version of the package manager you selected above.
+      placeholder: 10.0.0
     validations:
       required: true
 


r3g3m8:
Merge it please, so annoying error

veeceey:
Hey @r3g3m8, appreciate the +1\! Unfortunately I can't merge this myself — need a maintainer to review and approve first. All checks are passing, so hopefully it gets picked up soon.

packages/core/content-manager/server/src/services/utils/populate.ts
```diff
@@ -290,6 +294,13 @@ const getDeepPopulateDraftCount = (uid: UID.Schema) => {
           break;
         }
 
+        // Skip relations to content types without draft & publish,
+        // as they don't have a publishedAt attribute and can't have drafts
+        const targetModel = strapi.getModel(attribute.target);
+        if (targetModel && !hasDraftAndPublish(targetModel)) {
```
butcherZ: Hey @veeceey thank you for the fix, the code looks solid!
Just one little thing for this line, If `targetModel` is undefined, this condition is false, so execution falls through and tries to populate the relation with a `publishedAt` filter against a model that may not exist, which could crash. Even though this is an edge case, I think it would be safer to write `if (!targetModel || !hasDraftAndPublish(targetModel)) {
  break;
}` instead.

Once you fix this, the PR will be merged, thank you so much for your contribution!

veeceey:
pushed the guard fix - thanks for the catch!

butcherZ:
hey @veeceey thanks for the guard fix, I took the initiatives to apply a few tiny fixes, feel free to take a look, once I have your approval with the changes, I will merge the PR :D

veeceey:
Hey @butcherZ, just looked through your changes — they look great, nice catch on the targetModel guard. Approve from my side, go ahead and merge whenever you're ready\!

laurenskling:
Hey @butcherZ and @veeceey , you've related this to my https://github.com/strapi/strapi/issues/21338 but that issue was because of using the `filters` key, not because of publication state. So it's unrelated. Will change the title.

laurenskling:
I have relations to non-DP types, and countDraftRelations never does 400. 
I think you are having a false-positive here. 
As you see in my issue, it wasn't about the non-DP. There was a field with a key I should not use (`filters`).
I think you are hitting an error as well in your setup, which you are now suppressing by excluding it.

butcherZ:
Hey @laurenskling, thanks so much for the detailed context, I'm afraid you're right on both counts.
After investigating, I confirmed that `countDraftRelations` doesn't actually crash with non-DP relations on its own. `convertAndSanitizeFilters` silently strips the invalid publishedAt filter, so the query succeeds with a 200. 

That being said, this PR adds an explicit guard to avoid creating that invalid filter in the first place, but it's a code quality improvement rather than a bugfix. The real issue in #21338 is the filters field name collision, which our PR doesn't address. I apologise for the confusion, I reopened that issue and will be working on a fix for that one. 

Meanwhile I will remove the linked issue in this PR's description, this PR will stay merged as an enhancement, thanks again @veeceey for the contribution 😄

laurenskling:
Sounds perfect to me :)

Issue #25361, opened 2026-02-04
Relations silently lost when publishing document before its relation target


federicobadini:
### Node Version

22

### NPM/Yarn/PNPM Version

11

### Strapi Version

5.34.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/federicobadini/strapi-publish-relations-bug

### Bug Description

When publishing a document with relations to unpublished documents, the relations are silently dropped. This affects **both direct relations and dynamic zone component relations**. Publishing the targets afterwards does **not** restore the relations - data is lost.

## Root cause

In `document-service/transform/relations/transform/data-ids.ts`, Strapi maps draft IDs to published IDs during publish. If target isn't published, mapping returns empty and relation is discarded.
This can happen when either running a parallel publishing of records referencing each other or even when publishing from the Strapi dashboard an entry referencing an unpublished element

## Impact

- Silent data loss
- Affects any workflow where publish order isn't strictly controlled
- Particularly problematic for self-referential content types and batch imports

### Steps to Reproduce

1. Create 3 drafts: A, B, C
2. Set relations on A:
   - `A.parent -> B` (direct relation)
   - `A.items[0].submenu -> C` (dynamic zone component relation)
3. Publish A (while B and C are still drafts)
4. Check published A → **both relations are NULL**
5. Publish B and C
6. Check published A again → **still NULL** (not restored)

| Draft | Published |
|-------|-----------|
| <img width="1728" height="998" alt="Image" src="https://github.com/user-attachments/assets/01b737e8-813f-4074-9a6b-f04f29dba1fc" /> | <img width="1728" height="998" alt="Image" src="https://github.com/user-attachments/assets/cc91cd67-a582-4277-98ac-ace1f937dfef" /> |

### Expected Behavior

Either:
- Throw an error when publishing would orphan relations
- Keep relation and resolve when target is published
- Warn the user

## Actual behavior

Relations silently dropped. No error or warning.
Require to explicitly unpublish and re-publish the referencing entity

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

innerdvations:
[...]

PR #25459, opened 2026-02-15
fix(document-service): preserve self-referential relations during publish/discard

### LINKED ISSUE #24936: Can't create a relation between an entry and itself
### Node Version

v22.17.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.31.2

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When setting up a self-referential relation (an entry relating to itself), publishing the relation does not work as expected: the published version ends up without that relation. This likely happens because Strapi prevents creating relations between published and draft entries. As a result, when the entry attempts to relate to itself, Strapi interprets this as linking the entry to its own draft version, which is not allowed—creating a loop that blocks the relation from being saved in the published state.

### Steps to Reproduce

1. Create a content type with a title and a relation field
2. Create an entry in that content type: item 1 (the same entry that you are on)
3. In the relation field, select item 1
4. Publish the modifications
5. Go to the published view / access the entry via API and observe the relation not being populated

### Expected Behavior

The relation should be populated as expected

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

veeceey:
### What does it do?

Adds a new `self-referential-relations` sync utility that preserves self-referential relations (e.g. a "parent" field pointing to the same content type) during publish and discard-draft operations.

The fix works in three steps:
1. Before deleting old entries, loads all self-referential join table rows where both source and target belong to the entries being published/discarded
2. After new entries are created, remaps the old entity IDs to the new entity IDs
3. Inserts the remapped relations into the join table

This follows the same pattern already used by `unidirectional-relations.ts` and `bidirectional-relations.ts` for their respective sync logic.

### Why is it needed?

When publishing an entry with a self-referential relation, the relation is silently dropped. The publish flow:
1. Fetches the draft entry with populated relations (containing `documentId` references)
2. Deletes the old published version
3. Transforms draft data to resolve `documentId` -> entity ID for the `published` status
4. Creates the new published entry

At step 3, the published version's entity ID doesn't exist yet (old one was deleted, new one hasn't been created). Since `allowMissingId: true` is set during publish, the missing ID just returns an empty array instead of throwing, which means the self-relation is silently discarded.

The bidirectional-relations sync explicitly skips self-referential relations (with the comment "The order will already be handled as both sides are inside the same content type"), and the unidirectional-relations sync only handles relations from *other* content types targeting this one. So self-referential relations fall through the cracks.

### How to test it?

1. Create a content type with a self-referential relation (e.g. `Category` with a `parent` field of type `manyToOne` targeting itself)
2. Create a draft entry and set the `parent` relation to itself
3. Publish the entry
4. Verify the published version retains the self-referential relation
5. Also test with `discardDraft` to ensure the same fix works in reverse

There's also a [reproduction repo from a related issue](https://github.com/federicobadini/strapi-publish-relations-bug) that demonstrates the problem with self-referential relations being lost on publish.

### Related issue(s)/PR(s)

Fixes #24936

Also partially addresses #25361 (for the self-referential case specifically)


patrikalbertsson:
Any updates on this?

veeceey:
Hey @patrikalbertsson, unfortunately still waiting on a maintainer review. All checks pass and the CLA is signed. Hopefully they'll get to it soon\!

patrikalbertsson:
Seriously what is the hold up? I started my journey with migrating to v5 in Oct 2025 and its still not possible due to issues with migration script. Can someone please look at this PR ?

packages/core/core/src/services/document-service/utils/self-referential-relations.ts
```diff
@@ -0,0 +1,131 @@
...
+      // Only consider relations that target the same content type (self-referential)
+      if (attribute.type !== 'relation' || attribute.target !== uid) {
+        continue;
+      }
+
+      const joinTable = attribute.joinTable;
+      if (!joinTable) {
+        continue;
+      }
+
```
Adzouz, line 49: Bidirectional self-referential relations (e.g. manyToMany) have two attributes sharing the same join table, resulting in a 500 error for violating the unique constraint. We should process each join table only once to avoid duplicate inserts.

You can create a set before the transaction (line 38) `const processedJoinTables = new Set<string>();` that will store every joinTable (only once) and here you can check if the joinTable wasn't already processed, and if so, skip.

Like so:
```
if (processedJoinTables.has(joinTable.name)) {
  continue;
}
processedJoinTables.add(joinTable.name);
```

Adzouz (changes requested):
Hi @veeceey, thanks a lot for the contribution 🙏
I tested the original issue and indeed there's a problem preserving self-referential relations.
I also tested your fix and unfortunately I encountered a 500 error (because of violation of the unique constraint on the join tables) so I left a comment so you can fix that part to get rid of the error.
Also could you add some tests for this fix? It would be very helpful and be sure that it's working as expected and won't break in the future 😊

Adzouz:
Actually I saw there's also another PR opened for this fix: https://github.com/strapi/strapi/pull/25890
If you can wait a bit so I can also take a look at it before digging too much into fixing the issue here and writing tests 🙏

Adzouz:
Hi again, after looking at the two PRs, both were very similar but the other one had tests and fixed the issue I was mentioning so we decided to go with it instead of this PR sorry.
I'm closing it since the original issue was addressed, have a good day and thanks again for your efforts trying to make Strapi better ☀️

PR #25460, opened 2026-02-15
ci: adding SonarQube config

cache-your-dreams:
Adding SonarQube config so that we can get test coverage in Sonarcloud

### What does it do?
Strapi/strapi now includes SonarQube test coverage reporting that integrates with your existing Jest testing infrastructure. The setup supports both unit tests and frontend tests in your monorepo structure.
Initially, the workflow is not blocking on Failure.
PR to replace https://github.com/strapi/strapi/pull/25186 because of a commit message body missing

### Why is it needed?
So that we can track our test coverage to improve on it.

### How to test it?
This is a workflow that is run when pushing on develop. For testing purposes, I have added the feature branch as well, but this should be removed before merging

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25573,8 +25573,8 @@ __metadata:
   linkType: hard
 
 "mdast-util-to-hast@npm:^13.0.0":
-  version: 13.2.0
-  resolution: "mdast-util-to-hast@npm:13.2.0"
+  version: 13.2.1
+  resolution: "mdast-util-to-hast@npm:13.2.1"
   dependencies:
     "@types/hast": "npm:^3.0.0"
     "@types/mdast": "npm:^4.0.0"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -82,6 +82,9 @@
     "test:unit:update": "jest --config jest.config.js -u",
     "test:unit:all": "nx run-many --target=test:unit --nx-ignore-cycles",
     "test:unit:watch": "run test:unit --watch",
+    "test:unit:coverage": "jest --config jest.config.js --coverage",
+    "test:front:coverage": "cross-env IS_EE=true jest --config jest.config.front.js --runInBand --coverage",
+    "test:coverage": "yarn test:unit:coverage && yarn test:front:coverage",
```
Baldinof, line 89: These new script are not used in the CI, is it for local dev?

cache-your-dreams: From what I've understood, these scripts generate the coverage reports that SonarQube consumes, in `sonar.javascript.lcov.reportPaths` in `sonar-project.properties` for example. According to Claude, "Coverage is already wired into CI via the unit_back and unit_front jobs (which pass --coverage to jest) and consumed by the sonarqube job."

Baldinof: Yes, but it's not actually used in the CI.

I'm fine keeping them for local troubleshoot.

innerdvations: Yeah, they aren't used here, but it's ok to keep them

innerdvations (dismissed):

innerdvations:
unit tests are actually failing for some reason, I'll check on it

PR #25462, opened 2026-02-16
chore: improve a11y

markkaylor:
### What does it do?

- Add a static accessible name for the uploade progress dialog
- Mark some icons as decorative since text is nearby to describe 
- Add progress role to indeterminate progress bar
- Update e2e tests

### Why is it needed?

To improve a11y 

### How to test it?

Turn on the screen reader
Go to the Media Lib and upload some files
When the upload progress dialog opens it it will announce dialog
Move into the dialog
It should announce "Upload progress"

e2e tests should pass

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/25378

### DIFF
--- a/packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
+++ b/packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
@@ -1,7 +1,14 @@
 import * as React from 'react';
 
 import * as Dialog from '@radix-ui/react-dialog';
-import { Box, Flex, IconButton, TextButton, Typography } from '@strapi/design-system';
+import {
+  Box,
+  Flex,
+  IconButton,
+  TextButton,
+  Typography,
+  VisuallyHidden,
+} from '@strapi/design-system';
 import {
   ArrowsCounterClockwise,
   Check,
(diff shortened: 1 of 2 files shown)


packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -220,6 +222,16 @@ const DialogHeader = ({ handleClose }: { handleClose: () => void }) => {
       margin={1}
       hasRadius
     >
+      {/* Accessible name for the dialog */}
+      <VisuallyHidden>
+        <Dialog.Title>
```
Adzouz, line 227: Should we wait for my PR to be merged so you can change it to Drawer.Title?

Adzouz (commented):
Thanks a lot for the PR, it's important to think about everything 🙏
I think the hidden title should change depending on the progress don't you think? When the upload progress dialog is opening, it should say that the upload is in progress and when finished also say it and unfortunately on my side I couldn't hear it 😔
I know that we need to iterate fast so I don't mind putting it on hold and tackle that after finishing everything. Up to you!

markkaylor:
> Thanks a lot for the PR, it's important to think about everything 🙏 I think the hidden title should change depending on the progress don't you think? When the upload progress dialog is opening, it should say that the upload is in progress and when finished also say it and unfortunately on my side I couldn't hear it 😔 I know that we need to iterate fast so I don't mind putting it on hold and tackle that after finishing everything. Up to you!

I tried to get the aria-live to work but it just would never read it aloud. I decided to see how google drive did it and they just announce the dialog opening. I'll take another swing at it, after reading that aria spec again it does look like it's the right move. We can leave this branch around, keep moving forward, and improve a11y as we go 👍

PR #25467, opened 2026-02-16
feat(configure-view): redirect insert field to Content-Type Builder when no fields remain

lokeshchoudharyprogrammer:
# What does it do?

This PR improves the UX of the **Content Manager → Configure the view** screen (Edit View layout configuration) for collection types.

It updates the behavior of the **“Insert another field”** button in:

`packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx`

Technical changes:

- Uses `useDoc()` to retrieve the current `modelUid`.
- Uses `useNavigate()` from `react-router-dom` for navigation.
- Updates the “Insert another field” button logic:
  - If `remainingFields.length > 0` → behavior remains unchanged (dropdown menu opens).
  - If `remainingFields.length === 0` → redirects to:
    `/admin/plugins/content-type-builder/content-types/<modelUid>`

The button is no longer effectively useless when all fields are already added to the layout.

---

### Why is it needed?

Currently, when all visible fields are already present in the edit view layout:

- The “Insert another field” button cannot insert anything.
- The UI provides no clear guidance on the next step.

In this scenario, the natural next action for the user is to add new fields to the content type in the **Content-Type Builder**.

This PR:

- Removes a dead-end UX state.
- Makes the button context-aware.
- Improves discoverability by redirecting users directly to the correct Content-Type Builder page.
- Aligns with the existing "Edit the model" navigation pattern in the Edit View header.

---

#### Environment

- Node.js and Yarn installed.
- Strapi repository cloned and set up:

```bash
yarn install
yarn setup
````

Start the example app:

```bash
cd examples/getstarted
yarn develop --watch-admin
```

---

#### Case 1 – Remaining fields exist

1. Go to **Content Manager**.
2. Select a collection type (e.g. `Category`).
3. Open **Configure the view** for the Edit View.
4. Ensure not all fields are in the layout.
5. Click **“Insert another field”**.
6. A dropdown should open listing the remaining fields.
7. Selecting one should add it to the layout (no behavior change).

---

#### Case 2 – No remaining fields

1. Add all visible fields to the layout.
2. Click **“Insert another field”**.
3. You should be redirected to:

```
/admin/plugins/content-type-builder/content-types/<modelUid>
```

Example:

```
/admin/plugins/content-type-builder/content-types/api::category.category
```

---

#### Regression checks

Run:

```bash
yarn lint
yarn test:unit
yarn test:front
yarn test:e2e --setup --concurrency=1
```

Ensure all tests pass and no existing functionality is broken.

### DIFF
--- a/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
+++ b/packages/core/content-manager/admin/src/components/ConfigurationForm/Fields.tsx
@@ -19,11 +19,12 @@ import { Cog, Cross, Drag, Pencil, Plus } from '@strapi/icons';
 import { generateNKeysBetween as generateNKeysBetweenImpl } from 'fractional-indexing';
 import { produce } from 'immer';
 import { useIntl } from 'react-intl';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useNavigate } from 'react-router-dom';
 import { styled } from 'styled-components';
 
 import { getTranslation } from '../../utils/translations';
 import { ComponentIcon } from '../ComponentIcon';
+import { useDoc } from '../../hooks/useDocument';
 
 import { EditFieldForm, EditFieldFormProps } from './EditFieldForm';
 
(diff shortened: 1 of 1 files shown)


mathildeleg:
Thank you for your PR 🙏🏻
I'll close this as this isn't the UX behaviour that we want actually. I will however raise the point with the team that we need to deal with disabled states to make sure they are more clear to users, thanks for raising the issue!

PR #25468, opened 2026-02-16
future(upload): folder list/grid view + navigation

jhoward1994:
## What does it do?

This PR adds folder display and navigation capabilities to the future Media Library grid and table views.

### **Frontend Changes**

- **Folder Display:** Folders are rendered in grid view and table view
- **Folder Navigation:** 
- **Folder Title Hook:** The `useFolderTitle` hook displays the current folder name in the page header.
- **Folder Service:** Implements RTK Query endpoints (`getFolders`, `getFolder`)

---

## Why is this needed?

This implements the CMS-130 folder organization feature for the Media Library.

---

## Related issues/PRs

- CMS-130

### DIFF
--- a/packages/core/upload/admin/src/future/App.tsx
+++ b/packages/core/upload/admin/src/future/App.tsx
@@ -15,7 +15,7 @@ export const UnstableMediaLibrary = () => {
   });
 
   return (
-    <Page.Main width="100%">
+    <Page.Main>
       <Page.Title>{title}</Page.Title>
 
       <Routes>


packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -29,6 +30,79 @@ const StyledCard = styled(Card)`
...
+ * -----------------------------------------------------------------------------------------------*/
+
+const FoldersContainer = styled.div`
+  grid-column: 1 / -1; // Span full width to force new row for assets
+  display: flex;
+  flex-wrap: wrap;
+  gap: ${({ theme }) => theme.spaces[4]}; // 16px gap between folder items
+`;
+
+const FolderItem = styled.div`
```
markkaylor: Use styled Flex instead of a styled div

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -207,6 +254,14 @@ const DropFilesMessageImpl = styled(Box)<{ $leftContentWidth: number }>`
 export const AssetsPage = () => {
   const { formatMessage } = useIntl();
 
+  const [{ query }, setQuery] = useQueryParams<{ folder?: string }>();
+  const currentFolderId = query?.folder ? Number(query.folder) : null;
+  const title = useFolderTitle(currentFolderId);
+
+  const handleNavigateFolder = (folder: Folder) => {
+    setQuery({ folder: String(folder.id) });
+  };
```
markkaylor: Couldn't we avoid prop drilling by defining this handler directly in FolerRow? 

```tsx
  const [{ query }, setQuery] = useQueryParams<{ folder?: string }>();

  const handleNavigateFolder = (folder: Folder) => {
    setQuery({ folder: String(folder.id) });
  };
```

it could be a hook if it needs to be shared between table and grid views

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -29,6 +30,79 @@ const StyledCard = styled(Card)`
...
+  border: 1px solid ${({ theme }) => theme.colors.neutral200};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  background: ${({ theme }) => theme.colors.neutral0};
+  cursor: pointer;
+  transition: background 0.2s;
+
+  &:hover {
+    background: ${({ theme }) => theme.colors.neutral100};
+  }
+
```
markkaylor, line 56: hmm not sure about this. I can't find what it should be but I think closer to this:

Image

Could you check with Anka?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -269,7 +324,7 @@ export const AssetsPage = () => {
           <DropFilesMessage uploadDropZoneRef={uploadDropZoneRef} />
 
           <Layouts.Header
-            title="TODO: Folder location"
+            title={title}
```
markkaylor: I'm just seeing this 🤔 

Image

markkaylor (commented):
A couple small things but looks good and is working nicely

markkaylor:
Shouldn't creation date and a last modified be coming through here?

Screenshot 2026-02-16 at 15 17 58

markkaylor:
Hey @jhoward1994 just a heads up you should only add the `pr: chore` label to all future work for the media lib. Adding the feature label could confuse the person releasing strapi that week and result in us releasing a minor instead of a patch. The work we are doing is really a refactor anyway even if it feels like a new feature 🙃

jhoward1994:
Ah thanks, I did actually know that and I think I tried to click chore 🤦🏻‍♂️😆

jhoward1994:
image

I've found the culprit

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -29,6 +31,75 @@ const StyledCard = styled(Card)`
   overflow: hidden;
 `;
 
+/* -------------------------------------------------------------------------------------------------
+ * FolderItem
+ * -----------------------------------------------------------------------------------------------*/
+
+const FoldersRow = styled.div`
+  grid-column: 1 / -1;
```
markkaylor: probably better to use Box

packages/core/upload/admin/src/future/pages/Assets/hooks/useFolderTitle.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { formatMessage } = useIntl();
+  const { data: currentFolder, isLoading } = useGetFolderQuery(
+    { id: currentFolderId! },
+    { skip: currentFolderId === null }
+  );
+
+  const mediaLibraryLabel = formatMessage({
+    id: getTranslationKey('plugin.name'),
+    defaultMessage: 'Media Library',
+  });
```
markkaylor: I think the designs wanted "Home" for the root folder name? Also I think the folder title is supposed to display the count of items. I don't think it was specifically mentioned on that task. Does that require a backend change? If so we can handle it in another PR but if it's a quick update would be nice to do it here.

packages/core/upload/shared/contracts/folders.ts
```diff
@@ -37,6 +37,8 @@ export interface Folder {
   children?: number[];
   path?: string;
   files?: File[];
+  createdAt?: string;
+  updatedAt?: string;
```
markkaylor, line 41: Surprised to see an update here, since these values come through on the existing media lib. Is it just the type that was outdated?

markkaylor (commented):
A few more small comments but it's looking nice and working well

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -29,6 +31,75 @@ const StyledCard = styled(Card)`
...
+const FolderName = styled(Typography)`
+  flex: 1;
+  min-width: 0;
+`;
+
+interface FolderItemProps {
+  folder: Folder;
+}
+
+const FolderItemComponent = ({ folder }: FolderItemProps) => {
```
markkaylor: Could we call this FolderCard? Or at least omit the word Component in the component name since it's just kinda redundant

markkaylor (dismissed):
It looks good to me and working nicely. I have one small ask but otherwise feel free to merge.

markkaylor:
We can add the e2e test after the folder creation PR to test:

a user can create a folder add files to it and view the files in that folder

packages/core/upload/admin/src/future/pages/Assets/components/AssetsTable.tsx
```diff
@@ -151,20 +153,99 @@ const AssetRow = ({ asset }: AssetRowProps) => {
...
+          <Flex
+            justifyContent="center"
+            alignItems="center"
+            borderRadius="4px"
+            color="neutral600"
+            width="3.2rem"
+            height="3.2rem"
+            shrink={0}
+          >
+            <FolderIcon width={16} height={16} />
```
Adzouz: I fixed it [in my PR](https://github.com/strapi/strapi/pull/25498) but it should be 20

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -29,6 +31,75 @@ const StyledCard = styled(Card)`
...
+  const { formatMessage } = useIntl();
+  const { navigateToFolder } = useFolderNavigation();
+
+  return (
+    <FolderItem onClick={() => navigateToFolder(folder)} role="button" tabIndex={0}>
+      <FolderIconContainer>
+        <FolderIcon width={16} height={16} />
```
Adzouz: I fixed it [in my PR](https://github.com/strapi/strapi/pull/25498) but it should be 20

packages/core/upload/admin/src/future/pages/Assets/hooks/useFolderTitle.ts
```diff
@@ -0,0 +1,59 @@
...
+        defaultMessage: '{count, plural, =1 {# item} other {# items}}',
+      },
+      { count: fileCount }
+    );
+
+    return { title: `${homeLabel} (${itemCountLabel})` };
+  }
+
+  if (isLoading) {
+    return { title: homeLabel };
```
Adzouz: Isn't it confusing to show the Home label while it's loading? I see the home displayed for a little instant even briefly and it feels a bit weird for me. Could we replace by something like just `Loading...` for example?

packages/core/upload/admin/src/future/App.tsx
```diff
@@ -15,7 +15,7 @@ export const UnstableMediaLibrary = () => {
   });
 
   return (
-    <Page.Main>
+    <Page.Main width="100%">
```
Adzouz: I'll investigate why we need to do that now because it feels like something was broken recently (not related to the ML I think). I think I saw other pages broken too in the current ML.

Adzouz (changes requested):
Very good job, well done 👏
I left just a few comments 😊
Also, if you could replace the message of the dropzone that were hardcoded with "Current folder" to actually display the name of the current folder please? 🙏

Image

packages/core/upload/admin/src/future/pages/Assets/hooks/useFolderTitle.ts
markkaylor: hmm not sure what to do about that, I wonder if the full page should just be blocked loading until at least that data is available. Not sure 🤔 maybe we could check with Anka or ask for feedback from others.

packages/core/upload/admin/src/future/App.tsx
Adzouz: Ok, it's my fault 🤦 I fixed the issue in https://github.com/strapi/strapi/pull/25501
You can remove the width

packages/core/upload/admin/src/future/pages/Assets/hooks/useFolderTitle.ts
Adzouz: Yes let's move on and we'll see about that later (I kind like the idea of skeletons as loaders if I can pitch the idea at some point). But definitely will go in improvements.

PR #25475, opened 2026-02-16
fix(types): export getFetchParams as standalone importable utility

### LINKED ISSUE #25473: Service's using `getFetchParams` removes typings from returned documents
### Node Version

22.20.0

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.35.0

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Using the `getFetchParams` inside a custom service removes context from the returned objects.

### Steps to Reproduce

1. Create a model `Organisation` with populatable fields.
2. Create a custom service function following the documentation:
```typescript
export default factories.createCoreService('api::organisation.organisation', ({ strapi }) => ({
    async findOne(documentId, params = {}) {
        const data = await strapi.documents('api::organisation.organisation')
            .findOne({
                ...super.getFetchParams(params),
                populate: '*',
                documentId
            });
        return data as any; // this is due to a bug with the models ID. Fixed here: https://github.com/strapi/strapi/pull/25463
    },
}));
```

Result: `data` is missing populated fields. It also removes selected `fields` and other contexts.

### Expected Behavior

`data` should be properly typed.

### Logs

```shell

```

### Code Snippets

The problem: `super` being typed as `any`
```
let request = {
    ...super.getFetchParams(params),
    populate: '*',
    documentId
}
// request will be typed as any!
```

Workarround: cast the `getFetchParams` to `object`. This does not spread any into the props turning it into type any.
```
export default factories.createCoreService('api::organisation.organisation', ({ strapi }) => ({
    async findOne(documentId, params = {}) {
        const data = await strapi.documents('api::organisation.organisation')
            .findOne({
                ...super.getFetchParams(params) as object,
                populate: '*',
                documentId
            });
        return data as any;
    },
}));

let request{
    ...super.getFetchParams(params) as object,
    populate: '*',
    documentId
}
// request will be typed : {populate: string; documentId: string; }
```

The solution could be to just export it directly, since it just adds a status filter to the query at the moment and strongly type it.

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

ayhid:
## Summary

- Extract `getFetchParams` from the `CoreService` class into a **standalone exported function** available from `@strapi/strapi`
- Generic signature `<T extends object>` preserves caller types while accepting `object`-typed params from interfaces
- `CoreService.getFetchParams` is now a property assignment (`= getFetchParams`), keeping the type automatically in sync
- Return type guarantees `status` is always present (`T & { status: PublicationStatus.Kind }`)
- **Fix `this` typing in `createCoreService`** by aligning the `Service.Generic` index signature with `Controller.Generic`
- **Fix `CollectionTypeService.find()` pagination type error** by annotating params with `PaginationParams`

### Why a standalone export?

`createCoreService` uses prototype-based inheritance with plain objects, not ES6 classes. This means `super.getFetchParams(params)` is either invalid TypeScript in object literals or requires an `as any` cast that destroys type information.

By exporting `getFetchParams` as a standalone function, users get full type inference:

```typescript
import { factories, getFetchParams } from '@strapi/strapi';

export default factories.createCoreService('api::article.article', ({ strapi }) => ({
  async findOne(documentId, params = {}) {
    return strapi.documents('api::article.article').findOne({
      ...getFetchParams(params),
      documentId,
    });
  },
}));
```

### Why fix `Service.Generic`?

`this.getFetchParams(params)` resolved as `any` inside `createCoreService` callbacks, while the equivalent `this.sanitizeQuery(ctx)` worked fine in `createCoreController`. The root cause:

| | Controller | Service (before) | Service (after) |
|---|---|---|---|
| `Generic` type | `{ [name: string]: ControllerHandler<unknown> }` | `{ [key: keyof any]: unknown }` | `{ [name: string]: (...args: any[]) => any }` |
| `ThisType` binding | Works | **Broken** — `keyof any` (`string \| number \| symbol`) + `unknown` value is too permissive for TS to resolve `this` | Works |

TypeScript's `ThisType<T>` needs structural type information to establish a `this` context. The old `[key: keyof any]: unknown` told TS "any key can hold any value" — a type-level black hole that swallowed all structure, preventing `ThisType` from creating a binding.

### Why `T extends object` instead of `Record<string, unknown>`?

The `CollectionType` and `SingleType` interfaces declare params as `object`. Using `Record<string, unknown>` as the generic constraint caused a type mismatch — `object` is not assignable to `Record<string, unknown>`. Widening to `T extends object` aligns with the existing interface contracts.

### Why annotate `find()` with `PaginationParams`?

When `find(params = {})` had no type annotation, TypeScript inferred `T` as `{}`, making `getFetchParams` return `{} & { status: ... }`. This triggered TS2559 (weak type detection) when passed to `getPaginationInfo({ pagination?: PaginationParams })` — zero overlapping properties. The explicit annotation tells TS that `pagination` may be present.

Closes #25473

## Test plan
- [x] Verify `@strapi/types` lints successfully
- [x] Verify `@strapi/core` lints successfully
- [x] Verify no new type errors introduced (checked against `develop` baseline)
- [ ] Confirm no runtime regressions in core-api service tests
- [x] Verify `getFetchParams` is importable from `@strapi/strapi`
- [x] Verify `this.getFetchParams(params)` is correctly typed inside `createCoreService` callbacks


Bellian:
I do not think that this is the solution here.
Nice to see the function being properly typed, but this does not solve the problem #25473.

The bigest problem is the way to call it:
Since you can't use `this.getFetchParams(params)` (wich would be properly typed) but need to use `super.getFetchParams(params)` (since `super` is types as `any` any member of it will be too)

A better aproach would be to just have this method standalone and being able to import ist from the strapi package somehow.
```typescript
import { factories, getFetchParams } from '@strapi/strapi';
export default factories.createCoreService('api::organisation.organisation', ({ strapi }) => ({
    async findOne(documentId, params = {}) {
        const data = await strapi.documents('api::organisation.organisation')
            .findOne({
                ...getFetchParams(params),
                populate: '*',
                documentId
            });
        return data as any; // this is due to a bug with the models ID. Fixed here: https://github.com/strapi/strapi/pull/25463
    },
}));
```
This would solve the problem much cleaner.

Bellian:
@ayhid any feedback on this?

ayhid:
@Bellian sorry did not have time to check your comment, I will take a look in the weekend

ayhid:
@Bellian You're right — I've updated the PR to follow your suggestion. `getFetchParams` is now exported as a standalone function from `@strapi/core` (and transitively from `@strapi/strapi`), so it can be imported and called directly with full type inference:

```typescript
import { factories, getFetchParams } from '@strapi/strapi';
```

The `CoreService.getFetchParams` method still exists but delegates to the standalone function, so existing internal usage via `this.getFetchParams()` remains unaffected.

Thanks for the feedback!

Bellian:
@ayhid  The only thing i do not get is:

- in controllers you can use:
  ```typescript
  export default factories.createCoreController('api::organisation.organisation', ({ strapi }) => ({
    async findOne(ctx) {
      const sanitizedQueryParams = await this.sanitizeQuery(ctx);
      [...]
    }
  }));
  ```
- but in services this gives a type error:
  ```typescript
  export default factories.createCoreService('api::organisation.organisation', ({ strapi }) => ({
    async findOne(documentId, params = {}) {
        const param = this.getFetchParams(params)
    },
  }));
  ```

In both cases you access the generated methods via the `this` context.
In the controller example, the provided `cfg` aka `{ strapi }) => ({ ... })` is being considered as `WithStrapiCallback<PartialWithThis<Partial<CollectionType> & Generic>>` as it should be and allows acces to the `this` members.

In the service code this is not the case.
here the `cfg` is typed as any and therefore breaking the code...

Edit:
So i found out, that the IDE does not show any errors on the dev branch.
But strangely as soon as I start the dev process, I get this error:
```
src/api/test/services/test.ts:12:25 - error TS2339: Property 'getFetchParams' does not exist on type '{ findOne(documentId: any, params?: {}): any; }'.
12                 ...this.getFetchParams(params),
```
So the dev process does not correctly type the cfg.
Is there something different between the IDE setup and the ts generation on startup?

Bellian:
@ayhid Lol I guess I found the problem here:

@strapi/types/dist/core/core-api/service.d.ts
```typescript
export type ContentType<TContentTypeUID extends UID.ContentType> = Utils.MatchFirst<[
    Utils.Test<UID.IsCollectionType<TContentTypeUID>, CollectionType>,
    Utils.Test<UID.IsSingleType<TContentTypeUID>, SingleType>
], Base>;
```

If this is changed to:
```typescript
export type ContentType<TContentTypeUID extends UID.ContentType> = Utils.MatchFirst<[
    Utils.Test<UID.IsCollectionType<TContentTypeUID>, CollectionType<TContentTypeUID>,
    Utils.Test<UID.IsSingleType<TContentTypeUID>, SingleType>
], Base>;
```
It suddently works. Since the service's `CollectionType` is a generic type, passing the UID might do the trick.
I updated this in my node_modules folder and VSCode is fine with this, but the cient still does not start up..

I will test this in the strapi repo.

ayhid:
@Bellian Great catch

### Root cause

The `Generic` type used in `Service.Extendable` had an overly permissive index signature:

```typescript
// Before (broken)
export type Generic = {
  [key: keyof any]: unknown;
};
```

`keyof any` resolves to `string | number | symbol`, and combined with `unknown` as the value type, it tells TypeScript "any key can hold any value." This is a type-level black hole — when `PartialWithThis<T>` (which applies `ThisType<T>`) receives a type that includes this signature, TypeScript can't extract any structural information to establish the `this` binding.

Meanwhile, the **controller** `Generic` works because it uses a constrained signature:

```typescript
// Controller (works)
export type Generic = {
  [name: string]: ControllerHandler<unknown>;
};
```

Here, TS knows every property is a `ControllerHandler<unknown>`, which gives `ThisType` enough structure to resolve `this` correctly.

### The fix (just pushed)

Aligned the service `Generic` with the controller pattern:

```typescript
// After (fixed)
export type Generic = {
  [name: string]: (...args: any[]) => any;
};
```

This tells TS that service members are functions (which they are — methods), and `ThisType` can now resolve `this` properly inside `createCoreService` callbacks. So `this.getFetchParams(params)` will be correctly typed.

The standalone `getFetchParams` export remains useful as a cleaner pattern (avoids relying on `this`/`super` in object literals), but both approaches now work with full type inference.

Bellian:
@ayhid currently build fails due to:

packages/core/core/src/core-api/service/collection-type.ts
```typescript
 async find(params = {}) {
    const { uid } = this.contentType;
    const fetchParams = this.getFetchParams(params);
    const paginationInfo = getPaginationInfo(fetchParams);
```
The fetch params now have explicit typings and do not match the pagination info param.

Proposal:
```typescript
 async find(params: object & { pagination?: PaginationParams | undefined } = {}) {
    const { uid } = this.contentType;
    const fetchParams = this.getFetchParams(params);
    const paginationInfo = getPaginationInfo(fetchParams);
```

Also You may do:

packages/core/core/src/core-api/service/core-service.ts
```typescript
export abstract class CoreService {
  getFetchParams = getFetchParams;
}
```
This should keep the type of the original function and reduce duplication.

Also here (Dont know if there is a nicer way to get the type here...):

packages/core/types/src/core/core-api/service.ts
```typescript
import { type getFetchParams } from '../../../../core';
...
export interface Base {
  getFetchParams: typeof getFetchParams;
}
```

Sadly i cant do a commit ;)

Bellian:
@ayhid I also came along another problem right now:

The service's `CollectionType` is defined like so:
```typescript
export interface CollectionType extends Base {
  findOne(docId: Documents.ID, params: object): Promise<Document | null>;
}
```

Wich collides witht he current implementation of `getFetchParams` as it expects a `Record<string, unknown>`.
This results in an error.

So We must define the `params` in the `CollectionType` as `Record<string, unknown>` or type the `getFetchParams` to use `object` like so: `getFetchParams<T extends object>` wich should still work fine in this context.

ayhid:
@Bellian All three points addressed in the latest push — thanks again for the thorough testing!

### What changed

1. **`getFetchParams` constraint widened to `object`** — exactly as you suggested, `Record<string, unknown>` was too strict given the interfaces declare `params: object`. Changed to `getFetchParams<T extends object>(...)` in both `core-service.ts` and the `Base` interface in `@strapi/types`.

2. **`CoreService.getFetchParams = getFetchParams`** — adopted your property assignment suggestion. Keeps the type in sync automatically and removes the duplicated generic signature.

3. **`CollectionTypeService.find()` params annotated with `PaginationParams`** — the bare `params = {}` caused TypeScript to infer `T` as `{}`, making the return type `{ status: ... }` which has no overlap with `{ pagination?: PaginationParams }` (TS2559 weak type detection). Fixed with:
   ```typescript
   async find(params: { pagination?: PaginationParams } & Record<string, unknown> = {}) {
   ```

I opted not to cross-import `typeof getFetchParams` from `@strapi/core` into `@strapi/types` to avoid a circular dependency between the two packages — the `Base` interface keeps its own (now matching) generic signature instead.

packages/core/types/src/core/core-api/service.ts
```diff
@@ -29,14 +29,24 @@ type Data = Record<string, unknown>;
...
+ * {@link createCoreService}. The previous `[key: keyof any]: unknown`
+ * signature was too permissive and caused TypeScript to lose all type
+ * information for `this` in service method bodies.
+ *
+ * @see {@link ../controller.ts#Generic} for the equivalent controller type
  */
 export type Generic = {
-  [key: keyof any]: unknown;
+  [name: string]: (...args: any[]) => any;
 };
```
markkaylor, line 40: Even at the type level is this considered a breaking change? @nclsndr could I have your input here on this PR 🙏

markkaylor (commented):
In order to get the build working I had to update 

`packages/core/types/src/core/core-api/service.ts` with `Record<string, unknown>`

Image

Curious how are you testing with the build currently not working.

Otherwise having the type inference is very nice. Thanks to both of you for the collaboration here 👏

ayhid:
I might overseen this :( I have a bunch of other build fails not related to this so I might have things mixed up sorty

Bellian:
> In order to get the build working I had to update packages/core/types/src/core/core-api/service.ts with Record<string, unknown>

@markkaylor That was basicly my second message (https://github.com/strapi/strapi/pull/25475#issuecomment-3950888757). Maybe a bit too cryptic? Working on my communication here ;)

markkaylor:
> > In order to get the build working I had to update packages/core/types/src/core/core-api/service.ts with Record<string, unknown>
> 
> @markkaylor That was basicly my second message ([#25475 (comment)](https://github.com/strapi/strapi/pull/25475#issuecomment-3950888757)). Maybe a bit too cryptic? Working on my communication here ;)

🤦‍♂️ my bad must have read through everything a little too fast, your message is very clear 👍

ayhid:
All this for nothing ? 😨😭

Bellian:
> All this for nothing ? 😨😭

No you just forgot to apply those changes ^^

Next time i will not post two messages, but just edit the first i guess. this might be better for overview.

Bellian:
@ayhid Any update here?

ayhid:
@Bellian had a lot of work lately I'll try tomorrow! sorry

packages/core/types/src/core/core-api/service.ts
```diff
@@ -29,14 +29,24 @@ type Data = Record<string, unknown>;
...
+ * can establish a proper `this` binding inside object literals passed to
+ * {@link createCoreService}. The previous `[key: keyof any]: unknown`
+ * signature was too permissive and caused TypeScript to lose all type
+ * information for `this` in service method bodies.
+ *
+ * @see {@link ../controller.ts#Generic} for the equivalent controller type
  */
 export type Generic = {
-  [key: keyof any]: unknown;
+  [name: string]: (...args: any[]) => any;
```
markkaylor, line 39: Are you sure this change isn't enough to get the type inference you need via `this.getFetchParams` ? After some testing with @innerdvations it looks like you should have what you need without the additional export.

markkaylor (changes requested):
Hello,

After reviewing this internally with others (cc @nclsndr, @Bassel17, @innerdvations) we've decided to not move forward with the export of `getFetchParams`. Exporting utilities like this just to obtain type inference is not a pattern we want to introduce. For that reason I'll have to ask you to revert that change.

That being said, it does look like the type inference is coming through on `this.getFetchParams` after your update to the generic (see comment below):

```ts
export type Generic = {
  [name: string]: (...args: any[]) => any;
};
```

We are ok with this change and happy to move forward.

Otherwise it is also possible to type the params manually with something like:

```ts
type BasicFindOneParams = Modules.Documents.ServiceParams<'api::test.test'>['findOne'];
```

Thanks for your understanding. If you would like to discuss it further we could potentially find some time on discord in the coming days.

packages/core/types/src/core/core-api/service.ts
Bellian: @markkaylor 
Well another problem is, that `this.getFetchParams` does crash in services.
Using this strapi is not able to boot atm.
This is why you need to do `super.[...]`

ayhid:
Hello @Bassel17 do you need me to change something on this PR ?

Bassel17:
> Hello @Bassel17 do you need me to change something on this PR ?

hey thanks for looking in to this, it's all good I will merge it soon

Bassel17 (approved):
LGTM!

PR #25479, opened 2026-02-17
feat: adding 'Community' PR

cache-your-dreams:
### What does it do?
Adding a github action to label non-strapi PR's

### Why is it needed?
So we can identify which PR's are coming from the community and treat them

### How to test it?
It will need a PR from an external contributor

### DIFF
--- a/.github/workflows/community-label.yml
+++ b/.github/workflows/community-label.yml
@@ -16,21 +16,17 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            const org = 'strapi';
-            const username = context.payload.pull_request.user.login;
+            const { login } = context.payload.pull_request.user;
+            const association = context.payload.pull_request.author_association;
 
-            try {
-              await github.rest.orgs.checkMembershipForUser({ org, username });
-            } catch (error) {
-              if (error.status === 404 || error.status === 302) {
-                await github.rest.issues.addLabels({
-                  owner: context.repo.owner,
-                  repo: context.repo.repo,
-                  issue_number: context.payload.pull_request.number,
-                  labels: ['community']
-                });
-                core.info(`Added "community" label — ${username} is not a member of ${org}`);
-              } else {
-                core.setFailed(`Failed to check membership: ${error.message}`);
-              }
+            if (!['MEMBER', 'OWNER'].includes(association)) {
+              await github.rest.issues.addLabels({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.payload.pull_request.number,
+                labels: ['community']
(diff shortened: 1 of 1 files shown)


.github/workflows/community-label.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: Check org membership and label
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const org = 'strapi';
+            const username = context.payload.pull_request.user.login;
+
+            try {
+              await github.rest.orgs.checkMembershipForUser({ org, username });
```
oiorain: I dont know if you remember but in the past when i was working on invalid templates I tried to exclude strapier opened issues with this method but the data was not reliable. Did you test it with good results?

cache-your-dreams: I haven't tested it no. I think it would need to be on main to be able to test it. Did you get something better to work?

oiorain: I think i was testing this on engineerind-dashboard so it should have been reliable. but we cant test these things locally unfortunately. 

anyway, worst case scenario it doesnt label the thing? 
I think we can merge and test.

.github/workflows/community-label.yml
```diff
@@ -0,0 +1,36 @@
...
+            } catch (error) {
+              if (error.status === 404 || error.status === 302) {
+                await github.rest.issues.addLabels({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: context.payload.pull_request.number,
+                  labels: ['source: community']
```
nclsndr: I thought `source:` was a scope label rather than an origin marker. Are we fine using this suffix here?

cache-your-dreams: I checked and 'community' was used historically so I updated the PR

.github/workflows/community-label.yml
cache-your-dreams: I've updated the technique for checking this and suggested something that Claude thinks is more reliable

.github/workflows/community-label.yml
oiorain: @nclsndr `Source: *` is currently used for the part of the code the PR is touching. let's not rework everything. Although there's been a conversation going on with @derrickmehaffy on updating our labels overall to make them more coherent like adding the `scope: *` label like we have on `scope: relations` to identify features a little more precisely.

nclsndr: I was aware of the usage you describe @oiorain. Hence my question on why to qualify "community" as a scope. I don't want to challenge the current source/scope naming for now.

PR #25485, opened 2026-02-18
fix(guided-tour): no overlay in dark mode

markkaylor:
### What does it do?

Uses the theme for the overlay so it works on both dark and light mode

### Why is it needed?

The tooltips are hard to spot on dark mode

### How to test it?

Reset the tour if neeeded
Be in dark mode
Start the tour 
You should see the dark mode overlay that strapi modals typically have
Test the same on light mode

Screenshot 2026-02-18 at 10 07 29
Screenshot 2026-02-18 at 10 07 43

### DIFF
--- a/packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
+++ b/packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { Box, Popover, Portal } from '@strapi/design-system';
+import { Box, Popover, Portal, setOpacity } from '@strapi/design-system';
 import { styled } from 'styled-components';
 
 import { useGetGuidedTourMetaQuery } from '../../services/admin';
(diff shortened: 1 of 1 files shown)


packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -54,13 +54,19 @@ const GuidedTourTooltip = ({ children, ...props }: GuidedTourTooltipProps) => {
   return <GuidedTourTooltipImpl {...props}>{children}</GuidedTourTooltipImpl>;
 };
 
+// TODO: This should be exported from the design system
+const setOpacity = (hex: string, alpha: number) =>
```
Adzouz: I totally agree, I also needed to add it elsewhere but that's something that should come from the DS.

Adzouz: I created a PR if you're interested to review it:
https://github.com/strapi/design-system/pull/2001

markkaylor: oh nice so it's available now? I can update to use it instead of redefining if so. Thanks for doing that!

Adzouz: Unfortunately no, not yet. We need to release a new version of the DS. I'll see if we can plan doing that soon and I'll let you know

Adzouz: We might deploy a new version soon (today or tomorrow) if you're willing to wait a bit

Adzouz: Sorry it took a bit longer than expected but the DS was updated and you can now import setOpacity from it 😊 Actually, sorry to use your PR for that but if you could replace all occurrences in the CMS, that would be great 🙏

markkaylor: Thanks I'll update later today, I will probably do a separate PR for the update everywhere in the codebase

PR #25496, opened 2026-02-19
chore(upload): add import from url

markkaylor:
### What does it do?

- Add ImportFromUrlDialog component
- Create RTK mutation specific for handling from url and from url errors
- Update / add missing translations 
- Add e2e test for the import from url

### Why is it needed?

To maintain the upload from url feature

### How to test it?

With a url like https://picsum.photos/id/237/200/300.jpg, try to upload from url
Try a bunch of them
Upload should proceed like any other upload in the UploadProgressDialog

**Errors**

If at leats 1 url is invalid the error will be reported in the dialog and it will not proceed with upload
Otherwise, the errors should be reported in the UploadProgressDialog like any other upload

### DIFF
--- a/packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
+++ b/packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
@@ -52,8 +52,7 @@ const HeaderStatusIcon = styled(Flex)`
   }
 `;
 
-const HeaderStatusWrapper = styled(Drawer.Title)`
-  display: flex;
+const HeaderStatusWrapper = styled(Flex)`
   align-items: center;
 `;
 
--- a/packages/core/upload/admin/src/future/components/Drawer.tsx
+++ b/packages/core/upload/admin/src/future/components/Drawer.tsx
@@ -149,6 +149,7 @@ const DrawerBody = React.forwardRef<HTMLDivElement, DrawerBodyProps>(
     <Dialog.Content
       ref={ref}
       forceMount
+      asChild
       onPointerDownOutside={(e) => e.preventDefault()}
       onInteractOutside={(e) => e.preventDefault()}
       data-animation-direction={animationDirection}


packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -86,6 +94,23 @@ const HeaderStatus = ({ status, progress, totalFiles }: HeaderStatusProps) => {
   }
 
   if (status === 'success') {
+    const subtitle =
+      errorCount > 0
+        ? formatMessage(
+            {
+              id: getTranslationKey('upload.progress.success.subtitle.withErrors'),
```
jhoward1994, line 101: I think this needs to be added to en.json ?

packages/core/upload/admin/src/future/utils/files.ts
```diff
@@ -11,9 +36,166 @@ export function formatBytes(receivedBytes: number | string, decimals = 0) {
...
+ * ```ts
+ * const { files, errors } = await urlsToFiles([
+ *   'https://example.com/valid.jpg',
+ *   'https://example.com/invalid.jpg'
+ * ]);
+ * console.log(files.length); // 1
+ * console.log(errors.length); // 1
+ * ```
+ */
+export async function urlsToFiles(urls: string[]): Promise<UrlsToFilesResult> {
```
jhoward1994: I dont think this function is ever used?

markkaylor: whoops! thanks

tests/e2e/tests/media-library/future/page-objects/AssetsPage.ts
```diff
@@ -9,21 +9,28 @@ export class AssetsPage {
...
   readonly dropZone: Locator;
   readonly uploadProgressDialog: Locator;
+  readonly importFromUrlDialog: Locator;
+  readonly urlTextarea: Locator;
 
   constructor(page: Page) {
     this.page = page;
     this.newButton = page.getByRole('button', { name: 'New' });
-    this.importFilesMenuItem = page.getByRole('menuitem', { name: 'Import files' });
+    this.importFilesMenuItem = page.getByRole('menuitem', { name: 'File upload' }).first();
```
jhoward1994: Is there a way we can use an 'exact:true' selector in case things in the page
change ?

packages/core/upload/admin/src/future/services/api.ts
```diff
@@ -434,88 +480,146 @@ const uploadApi = adminApi
...
-                  errorMessage = errorData.message;
-                }
-              } catch {
-                errorMessage = `Retry failed with status ${response.status}`;
+          // Open progress dialog with all URLs as pending files
+          dispatch(
+            openUploadProgress({
+              totalFiles: urls.length,
+              fileNames,
+              fileSizes: urls.map(() => 0), // Size unknown until fetched
```
jhoward1994: Could we track the size in the store somehow for files from urls? I'm not sure if this would work

`packages/core/upload/admin/src/future/store/uploadProgress.ts`

```
setFileUploading(state, action) {
	const { index, size } = action.payload;
	if (state.files[index]) {
	 state.files[index].status = 'uploading';
	 state.files[index].size = size;
	}
},
```

jhoward1994:
Could it be worth fetching URLs server-side to avoid CORS issues

fetchUrlToFile does a bare browser fetch() for each URL in `utils/files.ts`
`const response = await fetch(url);`
I think the way we do this is subject to browser CORS policy

We cant guarantee URLs will have Access-Control-Allow-Origin set, so might the browser silently block them?

We could add a server-side endpoint (POST /upload/unstable/stream-from-urls) that accepts the URL list as JSON, fetches each URL with Node's fetch(), and streams back the same SSE events as /unstable/stream. The client POSTs the URL list avoiding browser fetch.

markkaylor:
> Could it be worth fetching URLs server-side to avoid CORS issues
> 
> fetchUrlToFile does a bare browser fetch() for each URL in `utils/files.ts` `const response = await fetch(url);` I think the way we do this is subject to browser CORS policy
> 
> We cant guarantee URLs will have Access-Control-Allow-Origin set, so might the browser silently block them?
> 
> We could add a server-side endpoint (POST /upload/unstable/stream-from-urls) that accepts the URL list as JSON, fetches each URL with Node's fetch(), and streams back the same SSE events as /unstable/stream. The client POSTs the URL list avoiding browser fetch.

Great idea I'll look into it!

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -196,6 +201,14 @@ export const AssetsPage = () => {
     await uploadFilesToFolder(files, null);
   };
 
+  const handleUrlUpload = async (urls: string[]) => {
+    try {
+      await uploadFromUrls({ urls, folderId: null }).unwrap();
```
Adzouz: Should we pass the folderId now the feature is done?

tests/e2e/tests/media-library/future/file-upload.spec.ts
```diff
@@ -5,97 +5,113 @@ import { AssetsPage } from './page-objects/AssetsPage';
...
-describeOnCondition(process.env.UNSTABLE_MEDIA_LIBRARY === 'true')(
-  'Media Library - File Upload',
-  () => {
-    test.beforeEach(async ({ page }) => {
-      await resetDatabaseAndImportDataFromPath('with-admin.tar');
-      await page.goto('/admin');
-      await login({ page });
-    });
+// A stable public test image URL
+const TEST_IMAGE_URL = 'https://picsum.photos/id/237/200/300.jpg';
```
Adzouz: Not sure how to do that but could it be possible to have something even more secure in the case where picsum for any reason would be down?
Maybe finding a way to call a local URL of an asset that is already in Strapi and served on the running test?

tests/e2e/tests/media-library/future/file-upload.spec.ts
```diff
@@ -5,97 +5,113 @@ import { AssetsPage } from './page-objects/AssetsPage';
...
+    const assetRow = assetsPage.getAssetRow('test-image');
+    await expect(assetRow).toBeVisible();
+  });
+
+  test('should upload a file from URL and show progress dialog with success', async ({ page }) => {
+    const assetsPage = new AssetsPage(page);
+    await assetsPage.goto();
+
+    // Upload from URL
+    await assetsPage.uploadFilesFromUrl(TEST_IMAGE_URL);
```
Adzouz: For visibility and because this constant is only used here, I would rather move it here directly in the test instead of defining globally far from its calling

packages/core/upload/admin/src/future/pages/Assets/components/ImportFromUrlDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+            <Button variant="tertiary" onClick={handleClose}>
+              {formatMessage({
+                id: 'app.components.Button.cancel',
+                defaultMessage: 'Cancel',
+              })}
+            </Button>
+            <Button type="submit">
+              {formatMessage({
+                id: getTranslationKey('modal.url.upload'),
+                defaultMessage: 'Upload',
```
jhoward1994, line 102: I think in the designs this is meant to be "Next" ?

packages/core/upload/admin/src/future/pages/Assets/components/ImportFromUrlDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+                error={error || undefined}
+                hint={formatMessage({
+                  id: getTranslationKey('input.url.description'),
+                  defaultMessage: 'Separate your URL links by a carriage return.',
+                })}
+              >
+                <Field.Label>
+                  {formatMessage({
+                    id: getTranslationKey('input.url.label'),
+                    defaultMessage: 'URL',
```
jhoward1994: Just a UX comment, the mix of single and plural feels a bit weird to me throughout this process. Everywhere is labelled with 'URL' and it feels like you'll only be able to add one link, but then we have the message 'Separate your URL links by a carriage return.'

jhoward1994:
I think in the designs this is styled as a single line input (40px height etc). But maybe the version in this branch works better if we support multiple URLs at once 

image

packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -312,6 +315,143 @@ export default {
...
+    const writeSSE = (event: string, data: Record<string, unknown>) => {
+      res.write(`event: ${event}\ndata: ${JSON.stringify(data)}\n\n`);
+    };
+
+    const total = urls.length;
+    const uploadErrors: FileUploadError[] = [];
+    const successfulFiles: any[] = [];
+
+    // Create temp directory for fetched files
+    const tmpWorkingDirectory = await fse.mkdtemp(path.join(os.tmpdir(), 'strapi-url-upload-'));
```
Adzouz, line 377: I have to admit I'm not very familiar with everything here in this file so maybe dumb question: is it creating for real the directory on the system the project is running? And if so, what would happen if two people try to upload files from a url at the same time? (I know it's pretty unlikely but just wondering)
Also why the trailing dash?

packages/core/upload/admin/src/future/pages/Assets/components/ImportFromUrlDialog.tsx
markkaylor: It was next because in the designs there was like an intermediary modal but I checked with Anka if we could upload directly and handle everything in the UploadProgressDialog. She was cool with that but the designs are out of date.

markkaylor:
> I think in the designs this is styled as a single line input (40px height etc). But maybe the version in this branch works better if we support multiple URLs at once
> image

Yeah you are right it is, I don't really have a strong opinion but I'll update to respect the designs with a minHeight that grows to this maxHeight before applying the scroll

packages/core/upload/server/src/services/file.ts
```diff
@@ -1,11 +1,119 @@
...
+  } catch {
+    throw new ApplicationError(`Invalid URL: ${url}`);
+  }
+
+  if (!['http:', 'https:'].includes(parsedUrl.protocol)) {
+    throw new ApplicationError(`Invalid URL protocol. Only http and https are allowed: ${url}`);
+  }
+
+  // Fetch the URL
+  const response = await fetch(url);
```
jhoward1994:
```suggestion
const FETCH_TIMEOUT_MS = 30_000;
...
  const response = await fetch(url, { signal: AbortSignal.timeout(FETCH_TIMEOUT_MS) });
```

Do you think we should add some kind of timeout mechanism??

packages/core/upload/server/src/services/file.ts
```diff
@@ -1,11 +1,119 @@
...
+    throw new ApplicationError(
+      `Failed to fetch URL: ${url} (${response.status} ${response.statusText})`
+    );
+  }
+
+  // Get content type and filename
+  const contentType =
+    response.headers.get('content-type')?.split(';')[0] || 'application/octet-stream';
+  const contentDisposition = response.headers.get('content-disposition');
+  const filename = getFilenameFromUrl(response.url, contentDisposition);
```
jhoward1994, line 143:
```suggestion
  const filename = path.basename(getFilenameFromUrl(response.url, contentDisposition));
```

I think something like this could give us more reliable filenames?

packages/core/upload/server/src/services/file.ts
markkaylor: Nice idea. Let's say multiple urls have been provided. 1 times out we abort just that one and continue to the rest. Is that what you had in mind?

packages/core/upload/server/src/services/file.ts
```diff
@@ -1,11 +1,119 @@
...
+  // Fetch the URL
+  const response = await fetch(url);
+
+  if (!response.ok) {
+    throw new ApplicationError(
+      `Failed to fetch URL: ${url} (${response.status} ${response.statusText})`
+    );
+  }
+
+  // Get content type and filename
```
jhoward1994, line 139: Just above here we could add a maximum file size check?

```suggestion
  const contentLengthHeader = response.headers.get('content-length');
  if (contentLengthHeader && parseInt(contentLengthHeader, 10) > MAX_FETCH_SIZE_BYTES) {
    throw new ApplicationError(
      `File too large: maximum allowed size is ${MAX_FETCH_SIZE_BYTES / (1024 * 1024)}MB`
    );
  }
```

packages/core/upload/server/src/controllers/admin-upload.ts
markkaylor: > I have to admit I'm not very familiar with everything here in this file so maybe dumb question: is it creating for real the directory on the system the project is running? And if so, what would happen if two people try to upload files from a url at the same time? (I know it's pretty unlikely but just wondering) Also why the trailing dash?

It's going to create a _unique_ temporary directory. So even if two people upload with the same exact url I think they are likely to just end up with duplicate files in the media lib (I wonder about this case in general for uploading). The trailing dash is how it is described in the docs because a unique 6 digit id will be appended. 

See docs: https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options

packages/core/upload/server/src/services/file.ts
markkaylor: I'll take a look and I think the maxFileSize limit can be configured by the user 🤔 not sure need to double check

markkaylor: Yeah looks like it: https://docs.strapi.io/cms/features/media-library#max-file-size

jhoward1994:
Nice, it works well! I just have one other thought

Should we do anything with URL validation? Currently I think we just check http/https protocol and do size limits. But what if someone tried to access something internal via - 127.0.0.1, localhost etc? Should these be blocked?

markkaylor:
> Nice, it works well! I just have one other thought
> 
> Should we do anything with URL validation? Currently I think we just check http/https protocol and do size limits. But what if someone tried to access something internal via - 127.0.0.1, localhost etc? Should these be blocked?

I don't really see a reason they should be blocked other than it doesn't make sense to upload something that's already uploaded. Did you have another reason to block these?

markkaylor:
@jhoward1994 gonna merge but if you have any other feedback leave it here and I'll create a ticket / open a follow up

jhoward1994:
Sorry I missed this yesterday @markkaylor but the latest changes look good to me, thanks for doing that!

PR #25498, opened 2026-02-19
future(upload): asset details drawer

Adzouz:
### What does it do?

Adding the asset details drawer to the new media library.
Created a reusable Drawer component shared between the asset details view and the upload progress dialog.

It allows to preview files for images, videos, audio and PDFs. It gives global information about them (size, duration, dimensions, dates, extension, etc).

_⚠️ For the moment, the fields aren't doing anything. There's no other action for the moment (crop image, enlarging PDFs, copy link, edit info, download, replace or delete) and they will be part of other tickets. It's only showing information at that point._

https://github.com/user-attachments/assets/d1d2a9bd-aab4-42a9-b960-0b31043e5aa3

### Why is it needed?

Enhance the experience of the user on the Media Library.

#### Setup the new Media Library
- Add the following to the .env file of your Strapi project:
`UNSTABLE_MEDIA_LIBRARY=true`
- Create a `features.ts` (or .js depending on your project) in your `config` folder and add the following content:
```
  future: {
    unstableMediaLibrary: env.bool('UNSTABLE_MEDIA_LIBRARY', false),
  },
});
```

#### Testing the feature
- Go to the new Media Library (represented by a circle exclamation mark in the side navigation)
- Upload different kinds of files (images, videos, audios, pdfs, other)
- Click on an element in the grid (or the table) view to open the details drawer
-> The correct information of the related file should be displayed on screen
-> You should be able to close the drawer to hide the information
-> See that clicking on an item adds a parameter to the url. That means that you can directly open the file information drawer from the url with the correct parameter. If the file id isn't correct, you should see an error alert:

https://github.com/user-attachments/assets/0872288d-b1e1-44d0-a3d2-1ed82e9b2eea

### Related issue(s)/PR(s)

Resolves CMS-122

🚀

### DIFF
--- a/packages/core/upload/admin/src/future/components/Drawer.tsx
+++ b/packages/core/upload/admin/src/future/components/Drawer.tsx
@@ -151,12 +151,9 @@ const DrawerBody = React.forwardRef<HTMLDivElement, DrawerBodyProps>(
       forceMount
       onPointerDownOutside={(e) => e.preventDefault()}
       onInteractOutside={(e) => e.preventDefault()}
+      data-animation-direction={animationDirection}
     >
-      <DrawerContainer
-        $animationDirection={animationDirection}
-        data-animation-direction={animationDirection}
-        {...props}
-      >
+      <DrawerContainer $animationDirection={animationDirection} {...props}>
         <DrawerContent>{children}</DrawerContent>
       </DrawerContainer>
     </Dialog.Content>


packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+  );
+};
+
+/* -------------------------------------------------------------------------------------------------
+ * Drawer.Header - composable header slot
+ * -----------------------------------------------------------------------------------------------*/
+
+const DrawerHeaderSlot = ({ children }: React.PropsWithChildren) => {
+  if (!children) return null;
+  return <>{children}</>;
```
markkaylor: Do we need a fragment?

packages/core/admin/admin/src/index.ts
```diff
@@ -97,6 +97,7 @@ export { translatedErrors } from './utils/translatedErrors';
 export * from './utils/getFetchClient';
 export * from './utils/baseQuery';
 export * from './utils/rulesEngine';
+export * from './utils/users';
```
Adzouz, line 100: It's because I'm using the getDisplayName in the drawer and it wasn't exported for some reason 🤷‍♂️

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+import { useGetAssetQuery, type AssetWithPopulatedCreatedBy } from '../../../services/assets';
+import { formatBytes, getFileExtension, prefixFileUrlWithBackendUrl } from '../../../utils/files';
+import { getAssetIcon } from '../../../utils/getAssetIcon';
+import { getTranslationKey } from '../../../utils/translations';
+import { formatDuration, useMediaDuration } from '../hooks/useMediaDuration';
+
+import { AssetPreview } from './AssetPreview';
+
+// Name of the parameter to look for in the URL to open the drawer
+const URL_PARAM = 'details';
```
markkaylor: Should we just call it what it is? `assetId`

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+    setIsClosing(true);
+  }, [hasValidId]);
+
+  React.useEffect(() => {
+    if (!isClosing) return;
+    const timer = window.setTimeout(() => {
+      setQuery({ [URL_PARAM]: undefined }, 'remove');
+      setIsClosing(false);
+      displayAssetId.current = null;
+    }, CLOSE_ANIMATION_MS);
```
markkaylor: Is the animation controlled by javascript? Not possible with CSS?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+      toggleNotification({
+        type: 'danger',
+        message: formatMessage({
+          id: getTranslationKey('asset-details.error'),
+          defaultMessage: 'Failed to load file details.',
+        }),
+      });
+      closeDetails();
+    }
+  }, [error, closeDetails, toggleNotification, formatMessage]);
```
markkaylor: Why not just report the error directly in the drawer instead of a toast?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+            </Drawer.CloseButton>
+          </Box>
+        </Flex>
+      </Drawer.Header>
+      <Drawer.Content>
+        <AssetPreview asset={asset} error={error} />
+        <AssetDetails asset={asset} error={error} />
+      </Drawer.Content>
+    </>
+  ) : null;
```
markkaylor: I think the default return should be everything is ok and we can render the content. If we have an error || !asset then we should render something to report that directly in the drawer. We can check with anka if we don't have designs for that

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -120,14 +120,18 @@ const FileName = styled(Typography)`
 
 interface AssetCardProps {
   asset: File;
+  onAssetItemClick?: (assetId: number) => void;
```
markkaylor: Why is it optional?

packages/core/upload/admin/src/future/pages/Assets/components/AssetsTable.tsx
```diff
@@ -85,21 +85,25 @@ const AssetPreviewCell = ({ asset }: AssetPreviewCellProps) => {
...
     >
-      <DocIcon width={16} height={16} />
+      <DocIcon width={20} height={20} />
     </Flex>
   );
 };
 
 interface AssetRowProps {
   asset: File;
+  onAssetItemClick?: (assetId: number) => void;
```
markkaylor: Same questions, why optional?

packages/core/upload/admin/src/future/pages/Assets/components/AssetsTable.tsx
```diff
@@ -85,21 +85,25 @@ const AssetPreviewCell = ({ asset }: AssetPreviewCellProps) => {
...
-const AssetRow = ({ asset }: AssetRowProps) => {
+const AssetRow = ({ asset, onAssetItemClick }: AssetRowProps) => {
   const isMobile = useIsMobile();
   const { formatDate, formatMessage } = useIntl();
 
   return (
-    <StyledTr>
+    <StyledTr
+      onClick={() => onAssetItemClick?.(asset.id)}
+      style={{ cursor: onAssetItemClick ? 'pointer' : undefined }}
```
markkaylor: 🤔 should all rows be clickable?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+        </Typography>
+      </Flex>
+    );
+  }
+
+  const isImage = asset.mime?.includes(AssetType.Image);
+  const metadataDuration = (asset.provider_metadata as { duration?: number } | undefined)?.duration;
+  const duration =
+    metadataDuration ?? (isVideo ? videoDuration : null) ?? (isAudio ? audioDuration : null);
+  const isDurationPending =
```
markkaylor: This condition is nuts and hard to read

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+            })}
+            value={
+              isDurationPending
+                ? formatMessage({ id: 'app.loading', defaultMessage: 'Loading...' })
+                : duration != null
+                  ? formatDuration(duration)
+                  : null
+            }
+          />
+        )}
```
markkaylor: Was this discussed? I don't see it in the designs.

I love that we have the audio preview and video preview, that's awesome. But I also think it means this information here is redundant since you can already see in the preview. 

Image

Image

packages/core/upload/admin/src/future/pages/Assets/hooks/useMediaDuration.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!url) {
+      setDuration(null);
+      setIsLoading(false);
+      return;
+    }
+
+    setIsLoading(true);
+    setDuration(null);
+
+    const element = document.createElement(type);
```
markkaylor: I'm not sure I agree with adding all this just show the duration in the details box. Especially since we have the preview that already says the duration. In any case since we also already have the element (audio or video) could we pass the ref to that element and access the duration on the element, instead of creating a new one here? 

That being said, my vote would be to remove the duration stuff for now

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetPreview.tsx
markkaylor, line 1: Is AssetPreview used elsewhere besides the AssetDeatilsDrawer? If not I think we can create a folder AssetDetails fro everything related to AssetsDetails

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+    &[data-state='closed'] {
+      animation: ${({ $animationDirection }) =>
+          $animationDirection === 'up' ? slideUpFromBottomOut : slideLeftFromRightOut}
+        300ms cubic-bezier(0.32, 0.72, 0, 1) forwards;
+      pointer-events: none;
+    }
+  }
+`;
+
+const DrawerContent = styled.div`
```
markkaylor: Probably better to use a Box

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+  box-shadow: ${({ theme }) => theme.shadows.popupShadow};
+  overflow: hidden;
+  border: 1px solid ${({ theme }) => theme.colors.neutral150};
+`;
+
+interface AnimatedBodyProps {
+  $isVisible: boolean;
+}
+
+const AnimatedBody = styled.div<AnimatedBodyProps>`
```
markkaylor: same should we use Box, it's a div

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+            forceMount
+            onPointerDownOutside={(e) => e.preventDefault()}
+            onInteractOutside={(e) => e.preventDefault()}
+          >
+            <VisuallyHidden>
+              <Dialog.Title>{formatMessage(title)}</Dialog.Title>
+              {description ? (
+                <Dialog.Description>{formatMessage(description)}</Dialog.Description>
+              ) : null}
+            </VisuallyHidden>
```
markkaylor: This restricts the API to mean the drawer can only have a VisuallyHidden title or description. I think it's better to export these as composable and allow the consumer to do what they want

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+const DrawerRoot = ({
+  animationDirection = 'up',
+  isVisible,
+  onClose,
+  isContentExpanded,
+  width,
+  height,
+  maxHeight,
+  dataTestId,
```
markkaylor: I don't think we should need dataTestId anymore and we should be able to rely on an accessible title

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+    animationDirection,
+    isVisible,
+    onClose,
+    isContentExpanded,
+    dataTestId,
+    width,
+    height,
+    maxHeight,
+    title,
+    description,
```
markkaylor: In my opinion all this and animationDirection do not belong in context since they are static values

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+};
+
+/* -------------------------------------------------------------------------------------------------
+ * Drawer.Footer - composable footer slot
+ * -----------------------------------------------------------------------------------------------*/
+
+const DrawerFooterSlot = ({ children }: React.PropsWithChildren) => {
+  if (!children) return null;
+  return <>{children}</>;
+};
```
markkaylor: I don't understand what this does. If it's just going to return children and add no styling then we probably don't need it.

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+const DrawerContext = React.createContext<DrawerContextValue | null>(null);
+
+const useDrawerContext = () => {
+  const context = React.useContext(DrawerContext);
+  if (!context) {
+    throw new Error('Drawer compound components must be used within Drawer.Root');
+  }
+  return context;
+};
```
markkaylor: Why do we need context?

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+      >
+        <Dialog.Portal>
+          <DrawerContainer
+            $animationDirection={animationDirection}
+            $width={width}
+            $height={height}
+            $maxHeight={maxHeight}
+            data-testid={dataTestId}
+            {...(!description ? { 'aria-describedby': undefined } : {})}
+            forceMount
```
markkaylor: what's this for?

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+  onClose,
+  isContentExpanded,
+  width,
+  height,
+  maxHeight,
+  dataTestId,
+  title,
+  description,
+  children,
+}: DrawerContextValue & React.PropsWithChildren) => {
```
markkaylor: Couldn't most of these props be provided to the Content component? I think it would be better if we could just spread props and forwards refs to the Content component

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+            onInteractOutside={(e) => e.preventDefault()}
+          >
+            <VisuallyHidden>
+              <Dialog.Title>{formatMessage(title)}</Dialog.Title>
+              {description ? (
+                <Dialog.Description>{formatMessage(description)}</Dialog.Description>
+              ) : null}
+            </VisuallyHidden>
+            <DrawerContent>{children}</DrawerContent>
+          </DrawerContainer>
```
markkaylor: Shouldn't all this be the export Content component? I'm thinking something like this as an export would allow us to avoid a ton of props passed from root


```suggestion
          <DrawerContainer
            $animationDirection={animationDirection}
            $width={width}
            $height={height}
            $maxHeight={maxHeight}
            data-testid={dataTestId}
            {...(!description ? { 'aria-describedby': undefined } : {})}
            forceMount
            onPointerDownOutside={(e) => e.preventDefault()}
            onInteractOutside={(e) => e.preventDefault()}
            {...props}
          >
            <VisuallyHidden>
              <Dialog.Title>{formatMessage(title)}</Dialog.Title>
              {description ? (
                <Dialog.Description>{formatMessage(description)}</Dialog.Description>
              ) : null}
            </VisuallyHidden>
            <DrawerContent>{children}</DrawerContent>
          </DrawerContainer>
```

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+/* -------------------------------------------------------------------------------------------------
+ * Drawer.Header - composable header slot
+ * -----------------------------------------------------------------------------------------------*/
+
+const DrawerHeaderSlot = ({ children }: React.PropsWithChildren) => {
+  if (!children) return null;
+  return <>{children}</>;
+};
+
```
markkaylor, line 183: Same comment as the other one I don't think this component accomplishes anything. It would be interesting to have something like this if all drawers had a consistent Header styling but that doesn't seem to be the case

markkaylor (changes requested):
A lot of complexity introduced here and I'm not convinced by the API. 

I think we can refine a bit and happy to pair tomorrow when hopefully I have a better internet connection. 

I don't want to block work refining this API for the drawer so let's take another pass and try a PP session to get it merged by end of day tomorrow or on Monday. We can keep improving after it's merged as well. 

All that being said, it works really nicely 👏

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 import * as Dialog from '@radix-ui/react-dialog';
```
markkaylor: Seems weird to have this import if Drawer is handling everything

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
```diff
@@ -444,40 +414,46 @@ export const UploadProgressDialog = () => {
...
+          {completedFiles.length > 0 && (
+            <CompletedFilesList>
+              {completedFiles.map((file) => (
+                <FileRowRenderer key={file.index} file={file} />
+              ))}
+            </CompletedFilesList>
+          )}
+        </Flex>
+      </Drawer.Content>
+      <Drawer.Footer />
```
markkaylor: I don't understand the need to add a component that renders nothing

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
```diff
@@ -0,0 +1,440 @@
...
+          })}
+          value={
+            asset.updatedAt
+              ? formatDate(new Date(asset.updatedAt), { dateStyle: 'long', timeStyle: 'short' })
+              : null
+          }
+        />
+        <DetailItem
+          label={formatMessage({
+            id: getTranslationKey('asset-details.provider'),
```
jhoward1994: Is that right or should it be this? I think ti also needs to be added to en.json
```suggestion
            id: getTranslationKey('asset-details.createdBy'),
```

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,297 @@
...
+/* -------------------------------------------------------------------------------------------------
+ * Styled components
+ * -----------------------------------------------------------------------------------------------*/
+
+interface DrawerContainerProps {
+  $width?: number | string;
+  $height?: number | string;
+  $maxHeight?: number | string;
+  $animationDirection?: 'up' | 'left';
+  $zIndex?: number;
```
jhoward1994: I dont think z index here is ever used? it's hardcoded as 100 atm

packages/core/upload/admin/src/future/pages/Assets/components/AssetsGrid.tsx
```diff
@@ -120,14 +120,18 @@ const FileName = styled(Typography)`
...
-const AssetCard = ({ asset }: AssetCardProps) => {
+const AssetCard = ({ asset, onAssetItemClick }: AssetCardProps) => {
   const { formatMessage } = useIntl();
   const TypeIcon = getAssetIcon(asset.mime, asset.ext);
 
   return (
-    <StyledCard>
+    <StyledCard
+      onClick={() => onAssetItemClick?.(asset.id)}
+      style={{ cursor: onAssetItemClick ? 'pointer' : undefined }}
```
jhoward1994: I wonder if we should try and be consistent and handle everything through styled
compos rather than passing a style tag?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
Adzouz: The animation is controlled by CSS but we need this extra logic to wait after the animation is over to remove the param from the url (otherwise the drawer will unmount before the animation finishes). Though there might be another way to achieve that (watch for the CSS animation and not a hardcoded variable), I'll take a look at it

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
Adzouz: Because I didn't have a design for that haha

packages/core/upload/admin/src/future/pages/Assets/hooks/useMediaDuration.ts
Adzouz: I think you're right, we can just remove the duration since it's redundant. And it wasn't part of the design, it's me who added it so let's remove it!

packages/core/upload/admin/src/future/components/Drawer.tsx
Adzouz: It's actually a requirement to avoid having an error in the console. It's for accessibility and it's how Dialog from Radix UI is working. The title and description are therefore mandatory

packages/core/upload/admin/src/future/components/Drawer.tsx
Adzouz: It's for the animation to be able to run. Since we are using data-state in the CSS to control the animation and it's based on the open variable sent to the Radix UI Dialog (I can add a comment)

packages/core/upload/admin/src/future/components/Drawer.tsx
Adzouz: Actually I understood the error I got. On UploadProgressDialog you did put a Dialog.Title and Dialog.Description so you didn't have any accessibility issue but when working on the AssetDetailsDrawer, the console told me that I needed to have an accessible Title and description for all Drawers, that's why I put it here. Since it's relative to each implementation, I'll remove it from here and put it directly in AssetDetailsDrawer

packages/core/upload/admin/src/future/components/UploadProgressDialog.tsx
Adzouz: Not everything, since the title and description (from Dialog) were styled differently from a drawer to another (the headers are different), we still need to keep them in this file. Except if you have a better idea?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
Adzouz: Couldn't find a better solution. Waiting for the animation to end over-engineer the component for a very specific use case (removing the param from the URL) so I exported the duration of the animation as a constant and used it in the setTimeout, is that okay for you?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetailsDrawer.tsx
Adzouz: I removed the alert to directly display it in the drawer (validated with Anka):

Screenshot 2026-02-19 at 16 54 14

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -0,0 +1,237 @@
...
+      ref={ref}
+      forceMount
+      onPointerDownOutside={(e) => e.preventDefault()}
+      onInteractOutside={(e) => e.preventDefault()}
+      asChild
+    >
+      <DrawerContainer
+        $animationDirection={animationDirection}
+        data-animation-direction={animationDirection}
+        forceMount
```
markkaylor: I don't think forceMount is necessary here, just on Dialog.Content, right?

tests/e2e/tests/media-library/future/asset-details.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      }) => {
+        const assetsPage = new AssetsPage(page);
+        await assetsPage.goto();
+        await assetsPage.switchToTableView();
+
+        const testImagePath = path.join(__dirname, '../../../data/uploads/test-image.jpg');
+        await assetsPage.uploadFilesWithFilePicker(testImagePath);
+        await assetsPage.waitForUploadProgressSuccess();
+        await assetsPage.closeUploadProgressDialog();
+
```
markkaylor: I think the database should already have some images if you want to use on of those to assert instead of uploading

tests/e2e/tests/media-library/future/asset-details.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+        const extensionValue = assetsPage.getDrawerDetailValue('Extension');
+        await expect(extensionValue).toContainText('jpg');
+
+        // Verify Asset ID is displayed (numeric)
+        const assetIdValue = assetsPage.getDrawerDetailValue('Asset ID');
+        await expect(assetIdValue).toHaveText(/\d+/);
+
+        // Verify Size is displayed
+        const sizeValue = assetsPage.getDrawerDetailValue('Size');
+        await expect(sizeValue).toHaveText('100B');
```
markkaylor: This looks more like a unit test to me, I think your grid view test below looks more like what we want

packages/core/upload/admin/src/future/services/assets.ts
```diff
@@ -2,6 +2,19 @@ import { uploadApi } from './api';
...
+  id: number;
+  firstname?: string;
+  lastname?: string;
+  username?: string | null;
+  email?: string;
+}
+
+export type AssetWithPopulatedCreatedBy = Omit<File, 'createdBy'> & {
+  createdBy?: PopulatedCreatedBy | null;
+};
```
markkaylor: Should this not becoming from the share/contracts?

tests/e2e/tests/media-library/future/asset-details.spec.ts
markkaylor: On second thought I'm a little nervous. If I login to the playwright test app, the new media lib has seeded images, but the old media lib does not. This could be corrupt data I'm not sure. We can investigate that later, feel free to leave this as is.

packages/core/upload/admin/src/future/services/assets.ts
Adzouz: I moved it 😊

tests/e2e/tests/media-library/future/asset-details.spec.ts
Adzouz: You're right, I actually merged the two tests together to avoid taking too much time and ressource to do something pretty similar and I removed this part

tests/e2e/tests/media-library/future/asset-details.spec.ts
markkaylor: It does seem like the data is there on purpose and is in every test app. Maybe on your end you can confirm in test-apps/<a-test-app>/public/uploads you see a bunch of assets?

Adzouz: I confirm and they are used in a bunch of entries in the content manager but why don't we have them in the old media lib? Is there something we should fix in the tar file about that?

markkaylor: Not sure, but let's not mess with the tar file right now (certainly not on a future branch). But I think we can rely on those images being present for tests.

PR #25500, opened 2026-02-19
fix(admin): dynamically register EE routes and services

Bassel17:
### What does it do?

- Adds dynamic EE feature registration 
- If license refresh enables cms-ai or audit-logs after boot, we register missing controller/routes/services at runtime.

### Why is it needed?

When initial STRAPI_LICENSE is missing some EE features, those modules are not loaded at boot.
If online license refresh later enables them, admin can show feature as enabled but backend stays non-functional until restart.

### How to test it?

- Run with a base STRAPI_LICENSE that does not include audit-logs / cms-ai.
- Ensure online license refresh returns a license that does include them.
- Start Strapi and confirm feature is initially unavailable.
- update EE license to include the feature
- refresh and check if the feature is available on the CMS side 

### Related issue(s)/PR(s)

fixes CMS-400


Bassel17:
I changed this to an enhancement, since this looks less like a bug and more like adding dynamic behavior.

Current behavior is that users need to restart their apps for EE license changes to fully take effect, and that appears to be by design today.

We can also explore an alternative approach instead of dynamically registering EE features: register EE features by default and enforce access through guardrails across APIs/routes/services.

PR #25506, opened 2026-02-19
security(deps): upgrade multiple dependencies

innerdvations:
### What does it do?

- `minimatch` to 10.2.1
- `elliptic` to 6.6.1
- `ajv` to 8.18.0
- `@aws-sdk/*` to resolve some instances of `fast-xml-parser` (the rest are held back by packages that have not been updated)

### Why is it needed?

GHSA-2g4f-4pwh-qvx6
GHSA-fc9h-whq2-v747
GHSA-3ppc-4f35-3m26
and probably others

### How to test it?

### Related issue(s)/PR(s)


package.json
```diff
@@ -161,7 +161,7 @@
     "lerna": "8.2.0",
     "lint-staged": "15.2.10",
     "lodash": "4.17.23",
-    "minimatch": "9.0.3",
+    "minimatch": "10.2.1",
```
markkaylor, line 164: no concern with major upgrade?

innerdvations: It has some breaking changes but nothing that we use

PR #24977, opened 2025-12-04
fix: missing status filter in content manager list view ; chore: improve type safety for filter type checking

asasin235:
### What does it do?

Adds a status filter to the Content Manager list view for content types with Draft & Publish enabled, and improves type safety in the Filters component.

**Changes:**
- Add status filter option when `schema.options.draftAndPublish` is `true`
- Filter uses `publishedAt` field with `$null` (Draft) and `$notNull` (Published) operators
- Improve type safety by using `as const satisfies` and a type guard function
- Remove `@ts-expect-error` comment in favor of proper type assertion
- Add unit tests for the new status filter functionality

### Why is it needed?

After migrating from Strapi 4 to Strapi 5, users lost the ability to filter content by draft/published status in the admin panel. While the list view displays a status column for content types with Draft & Publish enabled, there was no corresponding filter option available in the Filters dropdown.

This is critical when managing hundreds of content entries, as users cannot easily find all draft entries that need to be published or all published entries that might need updates.

### How to test it?

1. Create a content type with Draft & Publish enabled
2. Create entries with different statuses (published and drafts)
3. Go to Content Manager → select the content type
4. Click "Filters" → Select "Status" field
5. Choose "Draft" or "Published" to filter entries accordingly
6. Verify the filter works correctly

**Additional verification:**
- For content types WITHOUT Draft & Publish enabled, the "Status" filter should NOT appear
- The filter tag should display correctly (e.g., "Status Draft" or "Status Published")

### Related issue(s)/PR(s)

Fixes #24960

---

**Commits in this PR:**
- `fix: missing status filter in content manager list view`
- `chore: improve type safety for filter type checking`

PR #25510, opened 2026-02-19
feat(content-manager): filter list view by publication status

Smoke3785:
### What does it do?

Implements the [“Filter by State in Admin Content Manager” request ](https://feedback.strapi.io/customization/p/filter-by-state-in-admin-content-manager) for the Content Manager list view.

The following two files are modified:
```packages/core/content-manager/admin/src/utils/api.ts```

The `buildValidParams` function now transforms status filters from the URL query into top-level query parameters. e.g.

```tsx
// This filter query
{
  filters: {
    __status: {
      $eq: "published"
    },
    // ...other filters
  }
}
// Becomes this
{
  status: "published",
  // ...other filters
}
```
This allows us to leverage existing features of the content manager API without making substantial adjustments to the Filter component logic. 

`packages/core/content-manager/admin/src/pages/ListView/components/Filters.tsx`
If Draft & Publish is enabled on a content type, a "status" filter is injected. It uses the key `__status`, which cannot conflict with user-defined attributes.

### Why is it needed?

Responding to Strapi feedback.

### How to test it?
1. Open Content Manager, go to a collection type with Draft & Publish enabled.
2. Open Filters, select Status is Draft, verify only drafts are shown.
3. Select Status is Published, verify only published (and modified) entries are shown.
4. Combine with another filter and confirm both apply.
5. Select both draft and published, ensure no results are returned. 
6. Navigate to a collection type without Draft & Publish enabled, ensure no State filter is present.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/24977 - This proposed solution does not work


innerdvations:
I just updated the PR with all status possibilities and some tests -- for now we'll just keep `__status` an internal filter for this specific controller, and then in the future when we have a stable API to do the same thing we will update this to use it.

Please take a look at it and see if it's all good for you, I'll have it tested and reviewed by someone on this end too!

innerdvations:
Anyway, LGTM now too, I'm only not clicking approve because I now wrote a bunch of it myself. But as soon as someone else can test and approve it we can merge!

laurenskling:
Does this enable status to be sortable in the List view as well?

PR #25527, opened 2026-02-23
fix: show dash for empty table view text fields

### LINKED ISSUE #25469: Content Manager Table View Text Fields show just a dash
### Node Version

22

### NPM/Yarn/PNPM Version

NPM 10

### Strapi Version

5.36.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Since 5.36, the specific table headline fields are not being populated in the author view anymore - in our case, all 'title' fields show a dash:

(Additionally, the order I set in the 'configure view' screen is not being respected anymore).

Image

Configure view:

Image

Image

Content Model:

Image

Up until 5.36 this has worked fine.

### Steps to Reproduce

1. Create a new content type
2. Add a simple text field
3. Configure the table view to show that field as a column

### Expected Behavior

textfields are being shown in the author view table cell again

### Confirmation Checklist

Muzammil327:
### What does it do?

This PR fixes the rendering logic for component fields in the ListView. When a component does not have a mainField configured, the ListView now displays the component’s object.title instead of showing "-".

Previously, the ListView fallback logic did not properly handle this case, resulting in empty or placeholder values even when a meaningful title was available.

### Why is it needed?

When component fields have no mainField configured, the ListView currently displays "-", which provides no useful information to the user.

This makes it difficult to distinguish between entries that contain component data and those that do not.
By falling back to object.title, the ListView now shows a readable and relevant value, improving usability and clarity in the admin UI.

### How to test it?

1. Create or use a content type that contains a component field.
2. Ensure the component does not have a mainField configured.
3. Create or edit an entry and set a title value inside the component.
4. Go to the ListView for this content type.

Verify that:

- The component column displays the component’s title.
- It no longer shows "-" when data exists.

### Related issue(s)/PR(s)
N/A

Fixes #25469


fredericDelaporte:
Thank, but this PR does not appear to fix #25469, which happens even with a main field configured on the component, if that is here that is supposed to be set:
image

Anyway, it was working in Strapi 5.34, why would a new feature be required? It is a fix of what has broken it which is needed.

mathildeleg:
Hey, thanks for the contribution! The bug you're addressing was actually caused by a different root cause: `convertListLayoutToFieldLayouts` was being called without the component schemas dictionary, which caused `getMainField` to return `undefined` regardless of what was configured. The dash issue is no longer reproducible on current develop.

Because of that, this PR is no longer needed, and the `content?.title` fallback it introduces would only activate for components that have no `mainField` configured at all (which in practice always defaults to id). Even then, falling back to a hardcoded `.title` key isn't reliable since title isn't a reserved field name in Strapi component schemas.

So I'll close this PR, but thanks for digging into this :)

PR #25530, opened 2026-02-23
fix: vite and webpack config when linking ds locally

Adzouz:
### What does it do?

In [this PR](https://github.com/strapi/strapi/pull/25435) we added the design-system that was missing from the webpack / vite configuration to get rid of issues encountered by plugins developers.
Unfortunately it was making the local linking of the design system to this repo not working anymore.

This PR checks if the design system is locally linked and skip the addition to the config in that case.

---

Recently, this issue was raised: https://github.com/strapi/strapi/issues/25567

I pushed the second commit to fix the build issue and it should work now. If you want to test, you can create a Strapi app with the following experimental: `0.0.0-experimental.0ea3a439e5cc345c331bdebae48b482841104581` and now the `pnpm build` should pass without errors.

### Why is it needed?

Fixing the local linking of the design system package and this repo + the build command.

### How to test it?

_You need both the design system and this repo on your computer_

- Open the terminal
- Go to the Design system folder
- Run the following command `cd packages/design-system`
- `yarn watch`
- Open a new terminal
- Go to the CMS repository
- Run the following command `yarn link {FULLPATH_TO_DESIGN_SYSTEM_FOLDER}/packages/design-system`
- Run strapi (`yarn develop --watch-admin` in getstarted example project)
- Edit anything on the design system
-> The page should refresh with the applied changes to the DS

### Related issue(s)/PR(s)

Introduced by https://github.com/strapi/strapi/pull/25435
Resolves https://github.com/strapi/strapi/issues/25567

🚀

### DIFF
--- a/packages/core/strapi/src/node/core/linked-packages.ts
+++ b/packages/core/strapi/src/node/core/linked-packages.ts
@@ -1,25 +1,32 @@
 import path from 'node:path';
-import { createRequire } from 'node:module';
-import readPkgUp from 'read-pkg-up';
+import { getModulePath } from './resolve-module';
+
+const DESIGN_SYSTEM_MODULE = '@strapi/design-system';
+
+/**
+ * Detect if a package is locally linked (portal:, file:, yarn link) rather than installed in node_modules.
+ * When linked, the resolved path is outside node_modules.
+ *
+ * @internal
+ */
+export const isPackageLinked = (mod: string): boolean => {
+  const pkgRoot = getModulePath(mod);
+  const pathSegments = pkgRoot.split(path.sep);
+  return !pathSegments.includes('node_modules');
+};
 
 /**
  * Detects if @strapi/design-system is linked (portal:, file:, or yarn link).
  * Returns the package root path when linked, null otherwise.
+ * Uses the heuristic: linked packages resolve outside node_modules.
  */
-export const getLinkedDesignSystemPath = (cwd: string): string | null => {
+export const getLinkedDesignSystemPath = (): string | null => {
   try {
-    const require = createRequire(import.meta.url);
-    const resolvedPath = require.resolve('@strapi/design-system');
-    const pkg = readPkgUp.sync({ cwd: path.dirname(resolvedPath) });
-    if (!pkg) return null;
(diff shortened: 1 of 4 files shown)


mathildeleg (commented):
Linking seems to work okay, however when I unlink, I'm stuck with a white screen that I can't get rid of even if I re-install everything in the Strapi repo. Can you check unlinking works for you? Thanks 🙏🏻

packages/core/strapi/src/node/core/linked-packages.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Returns the package root path when linked, null otherwise.
+ */
+export const getLinkedDesignSystemPath = (cwd: string): string | null => {
+  try {
+    const require = createRequire(import.meta.url);
+    const resolvedPath = require.resolve('@strapi/design-system');
+    const pkg = readPkgUp.sync({ cwd: path.dirname(resolvedPath) });
+    if (!pkg) return null;
+    const pkgRoot = path.dirname(pkg.path);
+    const relativePath = path.relative(cwd, pkgRoot);
```
jhoward1994: Just an idea but could something like this be safer? A more accurate way to detect a package is locally linked?
```suggestion
const isInNodeModules = pkgRoot.split(path.sep).includes("node_modules");
const isLinked = !isInNodeModules;
```

packages/core/strapi/src/node/vite/config.ts
```diff
@@ -1,16 +1,30 @@
...
 import { loadStrapiMonorepo } from '../core/monorepo';
 import { getMonorepoAliases } from '../core/aliases';
 import type { BuildContext } from '../create-build-context';
 import { buildFilesPlugin } from './plugins';
 
+/**
+ * Resolve module to package root for use in aliases. Ensures pnpm's strict
+ * node_modules structure can resolve packages when bundling plugin chunks.
+ */
+const getModulePath = (mod: string) => {
```
jhoward1994: I think this is duplicated exactly in packages/core/strapi/src/node/webpack/config.ts. Maybe we could put it somewhere shared ?

PR #25532, opened 2026-02-23
Future media lib folder creation 

jhoward1994:
## Folder Creation in the Future Media Library (`AssetsPage`)

- Adds a `CreateFolderDialog` component and integrates it into `AssetsPage`

## Why is it needed?

The folder creation dialog is part of the Future Media Library core features.

### Folder creation

1. Navigate to the Media Library in the admin panel.
2. Click the **"New folder"** button and create a folder — it should immediately appear in the asset grid/list.

## Related issue(s)/PR(s)
CMS-131

### DIFF
--- a/packages/core/upload/admin/src/future/pages/Assets/components/tests/CreateFolderDialog.test.tsx
+++ b/packages/core/upload/admin/src/future/pages/Assets/components/tests/CreateFolderDialog.test.tsx
@@ -16,6 +16,8 @@ jest.mock('@strapi/admin/strapi-admin', () => ({
 }));
 
 const defaultProps = {
+  open: true,
+  folderName: 'Home',
   parentFolderId: null,
   onClose: jest.fn(),
 };
@@ -26,10 +28,10 @@ describe('CreateFolderDialog', () => {
     mockUnwrap.mockResolvedValue({});
   });
 
-  it('renders modal with title "New folder" and a "Folder name" input', () => {
+  it('renders modal with title "New folder in Home" and a "Folder name" input', () => {
     render(<CreateFolderDialog {...defaultProps} />);
 
-    expect(screen.getByText('New folder')).toBeInTheDocument();
+    expect(screen.getByText('New folder in Home')).toBeInTheDocument();
     expect(screen.getByRole('textbox')).toBeInTheDocument();
     // Field label
     expect(screen.getByText('Folder name')).toBeInTheDocument();
@@ -105,9 +107,7 @@ describe('CreateFolderDialog', () => {
   });
 
   it('shows inline API validation error (e.g. duplicate name) below the input', async () => {
-    mockUnwrap.mockRejectedValue({
-      response: { data: { error: { message: 'A folder with that name already exists' } } },
-    });
+    mockUnwrap.mockRejectedValue({ message: 'A folder with that name already exists' });
     render(<CreateFolderDialog {...defaultProps} />);
 
     fireEvent.change(screen.getByRole('textbox'), { target: { value: 'Existing Folder' } });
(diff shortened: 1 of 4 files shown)


packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -237,91 +240,108 @@ export const AssetsPage = () => {
...
+              </DropZoneWithOverlay>
+            </Layouts.Content>
+          </Layouts.Root>
+        </Box>
+      </UploadDropZoneProvider>
+      {showCreateFolderDialog && (
+        <CreateFolderDialog
+          parentFolderId={currentFolderId}
+          onClose={() => setShowCreateFolderDialog(false)}
+        />
```
markkaylor: I don't think it really makes a difference, but on my PR I passed the open prop instead of conditionally rendering like this. I kinda prefer that but either way I think we should be consistent.

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -185,6 +186,8 @@ export const AssetsPage = () => {
     { count: itemCount }
   );
 
+  const [showCreateFolderDialog, setShowCreateFolderDialog] = useState(false);
```
markkaylor: could we use isCreateFolderDialogOpen for a boolean. When reading `showCreateFolderDialog` it could be interpreted as a function instead of a boolean

Adzouz:
I'll review the code but I've done a quick test and it works pretty well, well done 👏
Just a tiny thing: do you think you could add a better message in the case where the folder name is already taken? 🙏
Screenshot 2026-02-25 at 16 38 35

packages/core/upload/admin/src/future/pages/Assets/components/tests/CreateFolderDialog.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    mockUnwrap.mockRejectedValue({
+      response: { data: { error: { message: 'A folder with that name already exists' } } },
+    });
+    render(<CreateFolderDialog {...defaultProps} />);
+
+    fireEvent.change(screen.getByRole('textbox'), { target: { value: 'Existing Folder' } });
+    fireEvent.click(screen.getByRole('button', { name: /create folder/i }));
+
+    await waitFor(() => {
+      expect(screen.getByText('A folder with that name already exists')).toBeInTheDocument();
```
Adzouz, line 117: It's weird, I don't see this message when triggering the error on my side. I commented it earlier, I only see an alert with a vague message 😕

jhoward1994:
Thanks! Both should be handled in [`a91c3da` (commit)](https://github.com/strapi/strapi/pull/25532/commits/a91c3da76b2a65b0a98a3b14fbbf5cc8c9f2bf60)

@markkaylor @Adzouz

Adzouz (commented):
Clean and efficient, we like that 👏
But yeah same comment as Mark, can you add an e2e test please and just check why I don't see the message when the folder already exist and we're good to go 🚀

packages/core/upload/admin/src/future/pages/Assets/components/tests/CreateFolderDialog.test.tsx
jhoward1994: Oh really, it definitely looks fixed for me after [`a91c3da` (this commit)](https://github.com/strapi/strapi/pull/25532/commits/a91c3da76b2a65b0a98a3b14fbbf5cc8c9f2bf60)

I'll check again but do you have everything rebuilt?

PR #25537, opened 2026-02-23
fix(content-manager): document intentional catch in stega decoding

ayhid:
## Summary

The empty `catch (error) {}` block in `applyStegaToElement` (preview script) silently swallowed all errors without documenting why. This looked like an oversight but is actually intentional — `stegaDecode` is called on every text node and is expected to fail for most of them (only a subset contain encoded data).

Changes:
- Add a comment explaining why the catch is intentionally empty
- Drop the unused `error` binding (`catch (error)` -> `catch`)

**File:** `packages/core/content-manager/admin/src/preview/utils/previewScript.ts`

## Test plan
- No functional change — this is a documentation-only improvement to an existing catch block

### DIFF
--- a/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
+++ b/packages/core/content-manager/admin/src/preview/utils/previewScript.ts
@@ -111,7 +111,11 @@ const previewScript = (config: PreviewScriptConfig) => {
               }
             });
           }
-        } catch (error) {}
+        } catch {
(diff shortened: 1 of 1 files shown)


derrickmehaffy:
Hey there it looks like this was created quite a while ago and the preview code has changed quite a bit and looks like it already solves this issue so going to mark as closed.

Issue #25542, opened 2026-02-24
v5 migration: discard-drafts fails to copy media morph entries and does not duplicate nested components


patrikalbertsson:
### Node Version

24.13.1

### NPM/Yarn/PNPM Version

NPM 11.8.0

### Strapi Version

5.36.1

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Javascript

### Bug Description

After migrating from Strapi v4 to v5 (5.36.1), the migration produces two categories of broken data. Both are still present after migrating to the latest version.

---

## Bug 1: Draft entries are missing media/file relations

After migration, media fields on **draft** entries are empty in the admin panel. The PUBLISHED tab shows images correctly, but switching to DRAFT shows empty upload fields. This occurs even though draft and published should have identical content immediately after migration.

This affects:
- Direct media fields on content types (e.g. a `cardImage` or `listImage` field)
- Media fields inside components (e.g. an `image` field on a component used in a dynamic zone)

Looking at the database, the migration correctly creates draft content type rows and duplicates first-level component rows with new IDs. However, the `files_related_morphs` table only has entries for the published side — no morph rows are created for the new draft IDs.

**Content type level — direct media fields:**

```
-- Entry is correctly split into published + draft
content_type table:
  id=100  document_id=abc  published_at=2026-01-01  (published)
  id=500  document_id=abc  published_at=NULL         (draft)

-- Morph entry only exists for the published ID
files_related_morphs:
  file_id=42  related_id=100  related_type=api::example.example  field=cardImage  ✓
  (nothing for related_id=500)                                                     ✗
```

**Component level — media inside components:**

```
-- Components are correctly duplicated with different cmp_ids
content_type_cmps:
  entity_id=100 → cmp_id=10  (shared.image, field=topImage)   ← published
  entity_id=500 → cmp_id=20  (shared.image, field=topImage)   ← draft ✓

-- But morph entry only exists for the published component
files_related_morphs:
  file_id=42  related_id=10  related_type=shared.image  field=image  ✓
  (nothing for related_id=20)                                         ✗
```

The component data itself (caption, alt text etc.) is correctly duplicated — only the polymorphic file relation is missing.

In our production database this affected over **40,000 missing morph rows** across ~15,000 content type entries and ~24,000 components.

---

## Bug 2: Published entries are missing relations on nested components

[...]

PR #25543, opened 2026-02-24
fix: polymorph relations in discard-drafts migration

innerdvations:
### What does it do?

Updates 5.0.0-discard-drafts.ts to correctly clone component trees during the v5 discard-drafts migration.

Technical changes:

- Prevents source-side morph copy helpers from copying rows from component join tables (*_cmps), which were being reinserted after component cloning.
- Uses raw component schema attributes (not transformed DB metadata attrs) to detect and recursively clone nested component / dynamiczone fields.
- Skips component/dynamic-zone join-table attrs in the generic component relation join-table copier so nested component links are handled only by the recursive clone path.
- Tracks cloned component pairs (originalId -> clonedId, including nested clones) and uses that mapping to copy media morph rows and other source-side morph rows to cloned components.
- Hardens upload morph table metadata lookup (supports alternate metadata shapes).
- Adds a safe fallback for component metadata UID lookup

### Why is it needed?

- Fixes 5.0.0-discard-drafts migration regressions where:
draft entries could still share component/dynamic-zone component IDs with published entries (especially via duplicated *_cmps rows),
cloned components could miss media morph rows,
nested component cloning/parity was incomplete.
This caused draft/published component tree inconsistency and media missing on draft clones.

### How to test it?

- Verify through the complex project in the mono-repo
- manually check nested components migration and see if there are differences between draft and published state

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/25542


Bassel17:
Should be good to go now, we just need an extra QAer

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -1475,6 +1794,83 @@ async function cloneComponentInstance({
...
+          });
+          const { id, ...rest } = row;
+          await insertRowWithDuplicateHandling(trx, nestedJoinTableName, {
+            ...rest,
+            [entityIdCol]: newComponentId,
+            [componentIdCol]: newNestedId,
+          });
+        }
+      } else if (attr.type === 'dynamiczone' && Array.isArray(attr.components)) {
+        const dzJoinTableName = getDzJoinTableName(collectionName, identifiers);
```
butcherZ, line 1842: it seems to me that `getDzJoinTableName` and `getComponentJoinTableName` are identical 🤔 ?Maybe we can remove one of them

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -231,6 +255,279 @@ async function copyRelationsToDrafts({ db, trx, uid }: { db: Database; trx: Knex
...
+  const uploadMorph = getUploadMorphTableInfo();
+  const uploadTableName = uploadMorph?.tableName ?? null;
+
+  for (const [componentType, cloneMap] of clonedComponentPairsCache.entries()) {
+    let componentMeta: any;
+    try {
+      componentMeta = getComponentMetadataByUid(componentType);
+    } catch {
+      continue;
+    }
```
butcherZ, line 502: it seems like `getComponentMetadataByUid` does not throw an error when no matching UID is found, it returns null instead, so maybe we can add a null check here? 
  
```
componentMeta = getComponentMetadataByUid(componentType);                                                   
  if (!componentMeta) continue;
```

Issue #25529, opened 2026-02-23
Strapi 5.36.1 – Guided tour tooltip mispositioned and page scroll locked on Content-Type page


arunrajharijan-collab:
### Node Version

22.19.0

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.36.1

### Operating System

MacOS

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Bug Description

After upgrading to Strapi 5.36.1, the admin guided tour on the Content-Type page behaves incorrectly:
	•	When clicking “Next” in the guided tour,
	•	The tooltip moves to the bottom of the page (outside visible viewport),
	•	The page becomes non-scrollable (vertical scroll locked),
	•	The overlay blocks interaction with the page.

If the browser is zoomed out, the misplaced tooltip becomes visible at the bottom.

This issue prevents interacting with the Content-Type Builder while the tour is active.

### Steps to Reproduce

1. Open Admin Panel
2. Go to Content-Type Builder
3. Trigger the guided tour (first login or reset localStorage)
4. Click Next
5. Observe tooltip position and scrolling behavior

### Expected Behavior

- Tooltip should remain aligned to its target element
- Page should scroll normally
- Overlay should not block the entire UI incorrectly

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

markkaylor:
Thank you for reporting this. Does this depend on the number of content-types you have and whether or not the components section is below the viewport? Or does it happen even if you have 0 content-types and components. Any chance you could send a video?

Muzammil327:
https://www.awesomescreenshot.com/video/49753536?key=37e99761ed0a2a6c1d7973b219a00f7b

PR #25548, opened 2026-02-25
fix: adjust scroll behavior for Content-Type Builder guided tour steps

Muzammil327:
Before: [https://www.awesomescreenshot.com/video/49753536?key=37e99761ed0a2a6c1d7973b219a00f7b](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

After: [https://www.awesomescreenshot.com/video/49762754?key=f0d41af429e4d56de20b4a6efd02f508](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### What does it do?

1. Fixes Guided Tour tooltip mispositioning when page scroll is locked.
2. Updates tooltip positioning logic for correct alignment.

### Why is it needed?

1. Tooltip was incorrectly positioned during guided tour with scroll lock.
2. Improves user experience and usability.

### How to test it?

- Open Strapi admin panel (5.36.1+).
- Create new user and start Guided Tour.
- Go to Step 3, tooltip moves to the bottom of the page (outside visible viewport) and page becomes non-scrollable (vertical scroll locked).

### Related issue(s)/PR(s)
Issue: #25529
PR branch: fix/Strapi-5.36.1-guided-tour-tooltip-mispositioned-page-scroll-locked-25529


packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -84,17 +85,21 @@ const GuidedTourTooltipImpl = ({
...
 
-  // Lock the scroll
+  // Scroll sidebar for specific steps in Content-Type Builder
   React.useEffect(() => {
     if (!isPopoverOpen) return;
 
-    const originalStyle = window.getComputedStyle(document.body).overflow;
-    document.body.style.overflow = 'hidden';
+    if (tourName !== 'contentTypeBuilder') return;
+    if (step !== 2 && step !== 3 && step !== 4 && step !== 5) return;
```
markkaylor: Unfortunately this isn't really the solution we are looking for. This is so specific to the content type builder we only solve the problem for that tour. What if another tour in the future has a similar problem? 

I wonder if we could use [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to ensure the tooltips are always visible.

markkaylor (changes requested):

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
Muzammil327: https://www.awesomescreenshot.com/video/50261380?key=c0ad416fd06004a4a3615f6118d88d4a

now i Update popover scrolling logic to use scrollIntoView().

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -137,7 +136,9 @@ const GuidedTourTooltipImpl = ({
         </Portal>
       )}
       <Popover.Root open={isPopoverOpen}>
-        <Popover.Anchor>{children}</Popover.Anchor>
+        <Popover.Anchor>
```
markkaylor, line 139: couldn't the ref go directly on anchor?

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -85,21 +86,19 @@ const GuidedTourTooltipImpl = ({
...
   React.useEffect(() => {
     if (!isPopoverOpen) return;
 
-    if (tourName !== 'contentTypeBuilder') return;
-    if (step !== 2 && step !== 3 && step !== 4 && step !== 5) return;
-
-    // Find CTB sidebar scroll container
-    const sidebarScroll = document.querySelector('[data-radix-scroll-area-viewport]');
+    // Use setTimeout to ensure the popover is fully rendered in the DOM
+    const timeoutId = setTimeout(() => {
```
markkaylor, line 94: I'm surprised to see the timeout. Are you sure it is necessary? 

If the anchorRef has a value and the popover is open then it should be rendered in the DOM. The timeout could lead to a condition potentially not met based on a time we can't really predict. That could lead to bugs or flaky tests.

packages/core/admin/admin/src/components/GuidedTour/Tours.tsx
```diff
@@ -62,7 +62,7 @@ const GuidedTourOverlay = styled(Box)`
   bottom: 0;
   background-color: rgba(50, 50, 77, 0.2);
   z-index: 10;
-  pointer-events: none;
+  // pointer-events: none;
```
markkaylor, line 65: Shouldn't we keep this? If not then it should be removed not commented out.

markkaylor (commented):
Thanks for the updates. A few more comments :)

mathildeleg:
Hey there, thanks for the PR! I see you haven't answered Mark's comments yet so I'll close the PR for now, feel free to re-open when you have time to tackle this again 🙂

Issue #25531, opened 2026-02-23
Date picker and combobox popovers in relation create/edit modal are not mouse‑accessible


molund:
### Node Version

22.21.1

### NPM/Yarn/PNPM Version

NPM 10.9.4

### Strapi Version

5.36.1

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

In the create/edit relation modal, the date picker and combobox popovers cannot be interacted with using a mouse.

When the popover is open, clicking anywhere inside it has the same effect as clicking outside the popover — it immediately closes without allowing a selection. However, the controls are usable via keyboard.

### Steps to Reproduce

1. Open any content type with a relation field
2. Create or edit a relation
3. Open:
    - a date picker, or
    - a combobox field (e.g. a nested relation)
4. Try to click an option inside the popover with the mouse

### Actual Behaviour

- Clicking inside the popover immediately closes it
- No option can be selected with the mouse
- The popover behaves as if it loses focus on any mouse interaction
- The mouse pointer does not change to a hand/pointer when hovering over selectable items in the popover, suggesting the popover is not receiving mouse events

### Expected Behavior

- Clicking inside the date picker or combobox popover should allow selecting a value
- Mouse and keyboard interactions should behave consistently

### Additional information

-  The issue appears limited to the relation create/edit modal
- Keyboard navigation works:
    - You can Tab/Arrow into the popover
    - Arrow keys allow selecting dates/options
- This makes the controls effectively keyboard‑only, which is unexpected and impacts usability and accessibility

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

adrianneOXD:
I am seeing the same issue where using the mouse to click in the date picker and combobox popovers does not work. 

Attached is a video recording of the issue:

https://github.com/user-attachments/assets/82edf85b-a018-46c1-a285-860ff4820d47

mathildeleg:
Hey there, this should have been fixed in 5.37.0, I'll close this but feel free to re-open if you still encounter the issue after upgrading! 🙏🏻

PR #25549, opened 2026-02-25
fix: radix ui dialog dependency version

Adzouz:
### What does it do?

Downgrade radix-ui dialog dependency to fix an issue with portal interactive components inside dialogs.
The version should match the one in the Design System (1.0.5).

### Why is it needed?

There's currently an issue with version `1.1.15` of the `@radix-ui/react-dialog` package making components like Combobox and Date Pickers not clickable anymore.

### How to test it?

- Run `yarn install`
- Check the version installed or `@radix-ui/react-dialog` in the node_modules, it should be 1.0.5

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/25531

🚀

### DIFF
--- a/packages/core/upload/package.json
+++ b/packages/core/upload/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@mux/mux-player-react": "3.1.0",
-    "@radix-ui/react-dialog": "1.1.15",
+    "@radix-ui/react-dialog": "1.0.5",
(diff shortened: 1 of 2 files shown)


markkaylor (approved):

markkaylor:
@Adzouz Let's wait for @innerdvations to get eyes on this before merging 🙏

Issue #24215, opened 2025-08-21
Issue with Releases and Relations


sas1sf3:
### Node Version

22.15.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.15.0/5.23.0

### Operating System

Windows 11

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

Hello,<br>I’ve come across another issue related to the Releases functionality in Strapi. Here’s the scenario:

I have two collection types: Page and Section.<br>A Page has a relation field to multiple Sections, and a Section has a relation field back to a Page.<br>I create a completely new Page and a new Section, both as drafts.<br>I add both entries to a release and then deploy the release.<br>At this point, Strapi shows both the Page and the Section as published (the admin UI marks them as published, and in the DB I can see two entries for each – one without published_at and one with it).<br>However, when I query the new Page (via client or Postman), the Section field is empty.<br>The only way I’ve found to resolve this is to manually open the Section in the Strapi admin, make a small change (e.g. add a space), and re-publish it. After that, the Section correctly appears in the client/Postman response.

### Steps to Reproduce

1. Create a **collection type** `Page` with:
   * `title` (string)
   * `sections` (relation → many Sections to one Page)
2. Create a **collection type** `Section` with:
   * `title` (string)
   * `page` (relation → one Page to many Sections)
3. In the admin, create a **new Section** as **Draft** (do not publish).
4. Create a **new Page** as **Draft** and link the Section in the `sections` relation (do not publish).
5. Create a **Release**, add both the Draft **Page** and Draft **Section** to the Release.
6. **Deploy** the Release.
7. Confirm in the admin UI that both entries now show as **Published**.
8. From a client (or Postman), fetch the published Page via **GraphQL** with relations populated

### Expected Behavior

The sections field should **include** the related Section.

### Logs

```shell
```

### Code Snippets

*No response*

### Media

*No response*

### Additional information

*No response*

### Confirmation Checklist

- [X] I have checked the existing [issues](<https://github.com/strapi/strapi/issues>) for duplicates.
- [X] I agree to follow this project's [Code of Conduct](<https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md>).

OcneanuVlad:
Internal note: TID 3011

Adzouz:
Hi @sas1sf3, thank you for reporting this issue and really sorry about the delay 🙏
[...]

PR #25551, opened 2026-02-25
fix(content-releases): publish in right order to preserve relations

Adzouz:
### What does it do?

In content releases, there's a bug when publishing two draft entries that have a relation. After publishing the release, we can't retrieve the relation when calling the API.
It's mainly because the two entries are published in parallel so the link between the two new published entries in the DB isn't created in the join table.

Video of the issue:

https://github.com/user-attachments/assets/519e8636-35cd-40f4-ae89-fb173dcfb9b5

After the fix:

https://github.com/user-attachments/assets/430c3771-8368-4da4-964d-717df669fbbe

### Why is it needed?

To fix an important issue with publishing relations through the content releases.

### How to test it?

- Create two collection types that are linked together with a relation (e.g. Article <-> Section) with Draft and Publish
- Create a draft of those collection types (don't forget to link them)
- Create a new release and put both your drafts in the release (to publish them)
- Publish the release
-> You should be able to retrieve the relation through an API call

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/24215

🚀

### DIFF
--- a/packages/core/content-releases/server/src/utils/index.ts
+++ b/packages/core/content-releases/server/src/utils/index.ts
@@ -208,10 +208,12 @@ export const getPublishOrderForContentTypes = (
           targetModel && contentTypesUtils.hasDraftAndPublish(targetModel);
 
         if (isTargetInRelease && targetHasDraftAndPublish) {
-          if (!dependencies.has(uid)) {
-            dependencies.set(uid, new Set());
+          let dependencySet = dependencies.get(uid);
+          if (!dependencySet) {
+            dependencySet = new Set();
+            dependencies.set(uid, dependencySet);
           }
-          dependencies.get(uid)!.add(targetContentTypeUid);
+          dependencySet.add(targetContentTypeUid);
         }
       }
     }
--- a/tests/api/plugins/content-releases/releases.test.api.ts
+++ b/tests/api/plugins/content-releases/releases.test.api.ts
@@ -1,7 +1,7 @@
 'use strict';
 
 import { createStrapiInstance } from 'api-tests/strapi';
-import { createAuthRequest } from 'api-tests/request';
+import { createAuthRequest, createContentAPIRequest } from 'api-tests/request';
 import { describeOnCondition } from 'api-tests/utils';
 import { createTestBuilder } from 'api-tests/builder';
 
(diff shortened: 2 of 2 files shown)


Adzouz:
Thanks for the deep testing @mathildeleg 🙏 I added the components and dynamic zones to the the existing logic. I tested and it seems that it's working well (even with nested components).
Let me know if that's also the case on your side 😊

Adzouz:
@markkaylor if you could add your review too, it would be really appreciated 🙏

markkaylor:
I'll review this afternoon :)

packages/core/content-releases/server/src/services/__tests__/release.test.ts
markkaylor, line 1: Do you think we could add an api test as well or instead? I think it would be more reliable

packages/core/content-releases/server/src/utils/index.ts
```diff
@@ -124,3 +126,118 @@ export const getEntryStatus = async (contentType: UID.ContentType, entry: Data.C
...
+          uidSet.has(targetContentTypeUid) && targetContentTypeUid in strapi.contentTypes;
+        const targetModel = strapi.getModel(targetContentTypeUid);
+        const targetHasDraftAndPublish =
+          targetModel && contentTypesUtils.hasDraftAndPublish(targetModel);
+
+        if (isTargetInRelease && targetHasDraftAndPublish) {
+          if (!dependencies.has(uid)) {
+            dependencies.set(uid, new Set());
+          }
+          dependencies.get(uid)!.add(targetContentTypeUid);
```
markkaylor: Should we assert this will not be undefined or guard it with a condition before adding?

markkaylor (commented):
Just two small comments, I'll test it now

packages/core/content-releases/server/src/services/__tests__/release.test.ts
Adzouz: Good point, I added an API test 😊✅

packages/core/content-releases/server/src/utils/index.ts
Adzouz: I refactored that part a bit

PR #25552, opened 2026-02-25
fix: feature works and rename addBodyParams to addInputParams

innerdvations:
### What does it do?

- renames `addBodyParams` to `addInputParams`
- fixes multiple bugs preventing `addInputParams` and `addQueryParams` from working when done in `register`
- no longer limit it to content-api routes (removes useless check)

### Why is it needed?

- feature actually works now (we don't release a broken feature)
- more accurate name; we actually validate `body.data` not `body` itself
- `body` is almost never used in Strapi, we use `body.data` for create/update; if we find the need later, we could add an actual `addBodyParams` that allows that, but for now we will not include it

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/services/server/register-routes.ts
+++ b/packages/core/core/src/services/server/register-routes.ts
@@ -106,7 +106,7 @@ const registerAPIRoutes = (strapi: Core.Strapi) => {
       router.type = 'content-api';
       router.routes?.forEach((route) => {
         generateRouteScope(route);
-        route.info = { apiName, type: 'content-api' };
+        route.info = { apiName };
       });
       strapi.contentAPI.applyExtraParamsToRoutes(router.routes ?? []);
 


innerdvations:
Oops, just added it

packages/core/core/src/services/server/register-routes.ts
```diff
@@ -77,11 +78,12 @@ const registerPluginRoutes = (strapi: Core.Strapi) => {
         router.prefix = router.prefix ?? `/${pluginName}`;
         router.routes.forEach((route) => {
           generateRouteScope(route);
-          route.info = { pluginName };
+          route.info = {
```
nclsndr: So we consider route.info to be undefined up to that point, right? So we don't spread a potential previous value in it.

tests/api/core/strapi/api/validate/validate-query.test.api.js
```diff
@@ -204,47 +204,105 @@ describe('Core API - Validate', () => {
...
+
+    // Why tests previously didn't catch the real-world failures:
+    // 1) route.info.type: register-routes set route.info = { apiName } but not type.
+    //    Fixed in production by setting route.info = { apiName, type: 'content-api' } in register-routes.
+    // 2) api.documents.strictParams: getstarted has api.documents.strictParams: true.
+    //    Fixed in production by stripping params in document service when strictParams is true.
+    //    We set api.documents.strictParams below so the test exercises the same path as production.
+    //
+    // We register both query and body params in beforeAll; apply runs once before the "after apply" tests
+    // (in the third test), so we use the same applyExtraParamsToRoutes the framework uses.
```
nclsndr: Not sure we want to keep this?

innerdvations: 🫠

nclsndr (approved):

packages/core/core/src/services/server/register-routes.ts
```diff
@@ -106,7 +106,7 @@ const registerAPIRoutes = (strapi: Core.Strapi) => {
       router.type = 'content-api';
       router.routes?.forEach((route) => {
         generateRouteScope(route);
-        route.info = { apiName, type: 'content-api' };
+        route.info = { apiName };
```
nclsndr, line 109: Out of curiosity, why was it wrong?

innerdvations: I realized it had only been added to cover up a testing deficiency, but when I fixed the tests I realized it wasn't actually needed. This is reverting it to the code that was there before -- a 4 year old line of code that I'm glad we're not actually changing in the end

PR #25557, opened 2026-02-25
fix(content-manager): reduce over-populated localizations and metadata

### LINKED ISSUE #24504: Content Manager over-populates localizations and metadata causing large payloads and slow admin UI

markkaylor:
### What does it do?
Reduces the amount of data populated for the localizations relation and meta.availableLocales in Content Manager responses.

- Adds withPopulateOverrids to the populate builder. Only populates minimal localization fields for content manager.
- Updates the bulk locale publish dialog to only validate the current document locale client-side, since other locales no longer carry full data. Server-side validation still applies during the actual publish.

### Why is it needed?
When a content type uses i18n with dynamic zones/components and has many locales, the CM endpoints deeply populate every localization and include heavy availableLocales metadata. This causes multi-MB payloads and hundreds of extra DB queries, making the admin panel very slow.

On develop
Screenshot 2026-02-25 at 15 16 23

This branch
Screenshot 2026-02-25 at 15 23 13

### How to test it?

I recommend you first reproduce the issue on develop and then come to this branch to see the fix

1. Enable the i18n plugin and add 10+ locales.
2. Create a localized collection type with dynamic zones or components.
3. Create several documents and fill in translations for each locale.
4. Open the Content Manager list view and single entry view.
5. Inspect the network responses — payloads should be significantly smaller (localizations should only contain locale, documentId, publishedAt, updatedAt instead of full nested content).
6. Verify the locale switcher and bulk locale publish still work correctly.

resolves #24504

### DIFF
--- a/packages/core/content-manager/server/src/services/populate-builder.ts
+++ b/packages/core/content-manager/server/src/services/populate-builder.ts
@@ -65,6 +65,7 @@ const populateBuilder = (uid: UID.Schema) => {
      */
     withPopulateOverride(overrides: Record<string, any>) {
       const prev = getInitialPopulate;
+      // merge(base, overrides): overrides win for overlapping keys, so e.g. localizations
       getInitialPopulate = async () => merge((await prev()) || {}, overrides);
       return builder;
     },


innerdvations:
It feels like we're adding an extra layer of indirection when we could use the existing populate param and just pre-build it with the right shape for `localizations` where we need it.

Basically, keep the isLocalized guard so we only do localizations handling for localized types.

Then for list/find, pre-build the populate with the right localizations shape: add a builder step that merges overrides into the initial populate (e.g. `withPopulate({ localizations: { fields: ['locale', 'documentId', 'publishedAt', 'updatedAt'] } })`) and use that where we currently call `.populateLocalizationsMode('minimal')`. The existing "if initialPopulate use it" in populate.ts will pick it up.

I might be missing something though, but that was my first thought

markkaylor:
> It feels like we're adding an extra layer of indirection when we could use the existing populate param and just pre-build it with the right shape for `localizations` where we need it.
> 
> Basically, keep the isLocalized guard so we only do localizations handling for localized types.
> 
> Then for list/find, pre-build the populate with the right localizations shape: add a builder step that merges overrides into the initial populate (e.g. `withPopulate({ localizations: { fields: ['locale', 'documentId', 'publishedAt', 'updatedAt'] } })`) and use that where we currently call `.populateLocalizationsMode('minimal')`. The existing "if initialPopulate use it" in populate.ts will pick it up.
> 
> I might be missing something though, but that was my first thought

I think I see what you mean, nice suggestion! That could simplify a lot and make this a more flexible override populate system, not just a "let's handle localizations", I'll explore it tomorrow

packages/core/content-manager/server/src/services/populate-builder.ts
```diff
@@ -57,6 +57,18 @@ const populateBuilder = (uid: UID.Schema) => {
...
 
+    /**
+     * Override the populate for specific attributes, taking precedence over
+     * query-derived or deep populate defaults.
+     *
+     * @param overrides - Populate overrides to merge (e.g. { localizations: { fields: ['locale'] } })
+     */
+    withPopulateOverride(overrides: Record<string, any>) {
+      const prev = getInitialPopulate;
+      getInitialPopulate = async () => merge((await prev()) || {}, overrides);
```
innerdvations, line 69:
```suggestion
      // merge(base, overrides): overrides win for overlapping keys, so e.g. localizations
      // from the override replaces any query/deep default
      getInitialPopulate = async () => merge((await prev()) || {}, overrides);
```

innerdvations: since it's a bit confusing

derrickmehaffy:
exp build: https://github.com/strapi/strapi/actions/runs/22680015469

PR #25562, opened 2026-02-25
fix(types): added support for required fields in entity's

### LINKED ISSUE #23789: Strapi types Data.Component/AttributeValues does not type properly define required properties
### Node Version

n/a

### NPM/Yarn/PNPM Version

n/a

### Strapi Version

5.7.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

When I define field as required and then use

```ts
import type { Data } from '@strapi/types'

type A = Data.Component<...>
```

I expect required types to be typed as always present, instead I get them being always optional and null.<br>This is because they are defined as such:<br>[https://github.com/strapi/strapi/blob/008123965da692a55d02a1df63facc54077c6bde/packages/core/types/src/data/component.ts#L21-L26](<https://github.com/strapi/strapi/blob/008123965da692a55d02a1df63facc54077c6bde/packages/core/types/src/data/component.ts#L21-L26>)

### Steps to Reproduce

See code snippet above

### Expected Behavior

Required properties are typed as always present.

### Logs

```shell
```

### Code Snippets

*No response*

### Media

*No response*

### Additional information

My end goal is to have typings in my FE project for now I use custom types, but would be great if I could use Data type out of the box.

### Confirmation Checklist

Bellian:
### What does it do?

Introduce support for required fields in the entity types.

### Why is it needed?

See #23789

### How to test it?

- Create a content type with required fields.
- Use the content Type: `let test: Data.ContentType<'api::test.test'> = null;` and check typing.

### Related issue(s)/PR(s)

may fix: #23789

### DIFF
--- a/packages/core/types/src/data/component.ts
+++ b/packages/core/types/src/data/component.ts
@@ -1,8 +1,10 @@
 import type { ID } from './constants';
 
 import type { Intersect } from '../utils';
-import type { AttributeNames, AttributeValueByName } from '../schema';
+import type { AttributeNames, AttributeValueByName, RequiredAttributeNames } from '../schema';
 import type * as UID from '../uid';
+
+// AttributeNames<TComponentUID>
 
 /**
  * Represents a component entry.
@@ -13,14 +15,25 @@ import type * as UID from '../uid';
 export type Component<
   TComponentUID extends UID.Component = UID.Component,
   TComponentKeys extends AttributeNames<TComponentUID> = AttributeNames<TComponentUID>,
-> = Intersect<[{ id: ID }, Pick<AttributeValues<TComponentUID>, TComponentKeys>]>;
+> = Intersect<
+  [
+    { id: ID },
+    Pick<
+      AttributeValues<TComponentUID>,
+      Extract<TComponentKeys, keyof AttributeValues<TComponentUID>>
+    >,
+  ]
+>;
 
 /**
  * @internal
  */
 type AttributeValues<TComponentUID extends UID.Component = UID.Component> = {
-  [TAttributeName in AttributeNames<TComponentUID>]?: AttributeValueByName<
-    TComponentUID,
-    TAttributeName
-  > | null;
+  [TAttributeName in Exclude<
+    AttributeNames<TComponentUID>,
+    RequiredAttributeNames<TComponentUID>
+  >]?: AttributeValueByName<TComponentUID, TAttributeName> | null;
+} & {
+  [TAttributeName in AttributeNames<TComponentUID> &
+    RequiredAttributeNames<TComponentUID>]: AttributeValueByName<TComponentUID, TAttributeName>;
 };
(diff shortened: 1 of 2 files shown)


packages/core/types/src/data/content-type.ts
```diff
@@ -19,12 +19,24 @@ export type ContentType<
...
-    TContentTypeUID,
-    TAttributeName
-  > | null;
+  [TAttributeName in Exclude<
+    AttributeNames<TContentTypeUID>,
+    RequiredAttributeNames<TContentTypeUID>
+  >]?: AttributeValueByName<TContentTypeUID, TAttributeName> | null;
+} & {
+  [TAttributeName in AttributeNames<TContentTypeUID> &
+    RequiredAttributeNames<TContentTypeUID>]: AttributeValueByName<TContentTypeUID, TAttributeName>;
```
innerdvations, line 41: Please add `| null` here:

```ts
]: AttributeValueByName<TContentTypeUID, TAttributeName> | null;
```

Required keys, nullable values — same as the optional branch, minus `?`.

_AI-assisted comment._

packages/core/types/src/data/component.ts
```diff
@@ -13,14 +15,25 @@ import type * as UID from '../uid';
...
-    TComponentUID,
-    TAttributeName
-  > | null;
+  [TAttributeName in Exclude<
+    AttributeNames<TComponentUID>,
+    RequiredAttributeNames<TComponentUID>
+  >]?: AttributeValueByName<TComponentUID, TAttributeName> | null;
+} & {
+  [TAttributeName in AttributeNames<TComponentUID> &
+    RequiredAttributeNames<TComponentUID>]: AttributeValueByName<TComponentUID, TAttributeName>;
```
innerdvations, line 38: Same as `content-type.ts`: keep `| null` on the required mapped value.

```ts
]: AttributeValueByName<TComponentUID, TAttributeName> | null;
```

_AI-assisted comment._

innerdvations (changes requested):
Thanks for wiring `RequiredAttributeNames` — making required keys required is the right fix for #23789.

**Change requested:** keep `| null` on required **output** values in both `content-type.ts` and `component.ts`. `GetRelationValue` / `GetComponentValue` / `GetMediaValue` are not nullable on their own; the old `| null` on `AttributeValues` is what made empty populated relations/media/components (and draft required scalars) type-check as `null`. Document Service results go through `Data.Entity`, so this would currently type those fields as always defined.

Keys stay required (drop `?`); values stay `T | null`.

**Follow-up (not blocking this PR):** `@strapi/types` has no type-behavior tests today (`test:ts` is only `tsc --noEmit`). A later PR could add a small dts/`expectTypeOf` fixture that module-augments a content type and asserts:
- required scalar: key required, value `T | null`
- optional scalar: `?: T | null`
- required relation/component/media: still `T | null`
- second type-param still drops unselected required keys

_AI-assisted comment._

PR #25574, opened 2026-02-26
feat(content-manager): add Zod 4 foundation utilities

butcherZ:
## What does it do?

**Phase 0 — Foundation utilities (CMS-412):**
Adds Zod 4 foundation utilities to `@strapi/content-manager` for the yup-to-zod migration.

- `validateZodAsync(schema)` — async validator matching `validateYupSchema` signature, throws `ValidationError` with formatted error details
- `formatZodErrors(zodError)` — transforms `ZodError` into the same `{ errors, message }` shape as `formatYupErrors`
- `strapiID` — `z.union([z.string(), z.number().int().nonneg()])` matching the existing Yup `StrapiIDSchema`
- `getZodValidationErrors(zodError)` — converts `ZodError` to `FormErrors` (`Record<string, string>`) matching `getYupValidationErrors`

**Phase 1 — Server validation migration (CMS-413):**
Migrates 7 server-side validation files from Yup to Zod 4:

- `history-version.ts` — simple schema
- `preview.ts` — simple schema
- `hasPermissions.ts` — simple schema (`validateYupSchemaSync` → `validateZod`)
- `homepage.ts` — `yup.ValidationError` → `z.ZodError`
- `relations.ts` — `z.coerce.number()` for `strict: false` coercion
- `dimensions.ts` — `yup.lazy()` → `z.union()`
- `index.ts` — `.test()` with `TestContext` → factory function + `.refine()`

**Fixes for Zod 3/4 type compatibility:**
- `@strapi/utils` `validateZod` now accepts an optional `errorMessage` param (matching `validateYupSchema` signature)
- `@strapi/admin` `project-settings.ts` — imports `z` from `@strapi/utils` (Zod 4) instead of `zod` (Zod 3)
- `@strapi/content-type-builder` `schema.ts` — type cast for Zod 3 schema passed to `validateZod` (Zod 4)
- Removed stale `zod@4.3.6` entry from `yarn.lock`

## Why is it needed?

Yup's loose type inference is blocking MCP content type management work (CMS-2). Zod 4 provides strict TypeScript types via `z.infer<>` needed for MCP tool parameter definitions. This completes Phases 0+1 which are the minimum to unblock CMS-2.

`model-configuration.ts` remains on Yup — it will be migrated in Phase 2 (CMS-414).

## How to test it?

Server-side unit tests:
```
npx jest --config packages/core/content-manager/jest.config.js --testPathPattern="server/"
```

All 25 test suites pass (160 tests), including 9 dimensions tests and 3 zod-yup compat tests.

## Related issue(s)/PR(s)

- Completes CMS-412
- Completes CMS-413
- Unblocks CMS-2


innerdvations:
zod 4 is now being exported from @strapi/utils for users to use -- we should probably use that here as well to ensure we always use the same version everywhere

packages/core/content-manager/server/src/validation/__tests__/zod-yup-compat.test.ts
```diff
@@ -0,0 +1,78 @@
+import { z } from 'zod';
+import * as yup from 'yup';
+import { formatYupErrors } from '@strapi/utils/dist/format-yup-error';
```
nclsndr: I feel like this import should not be valid in that form.

packages/core/content-manager/admin/src/utils/validationZod.ts
nclsndr, line 1: Don't we want those utils to be directly in @strapi/utils?

packages/core/content-manager/server/src/history/controllers/validation/history-version.ts
```diff
@@ -1,11 +1,7 @@
...
-  .shape({
-    contentType: yup.string().trim().required(),
-  })
-  .required();
+const historyRestoreVersionSchema = z.object({
+  contentType: z.string().trim(),
+});
 
-export const validateRestoreVersion = validateYupSchema(historyRestoreVersionSchema);
+export const validateRestoreVersion = validateZod(historyRestoreVersionSchema);
```
nclsndr: Why not `validateZodSchema` as `validateYupSchema`?

packages/core/content-manager/server/src/validation/__tests__/zod-yup-compat.test.ts
nclsndr, line 1: 💯

innerdvations:
A few more mismatches that could be breaking changes.

The biggest one in API responses is the top-level `message`: when Yup reports several failures at once, the parent message is the aggregate form your tests already encode (for example the "N errors occurred" style), whereas Zod only uses the first issue's text

Another mismatch is each entry in `details.errors`: Yup fills `value` from the Yup error when it can, but Zod code that always leaves `value` as `undefined` will not look the same to consumers that read those fields.

Finally, it looks like we have duplicated behaviour between `packages/core/utils` and `packages/core/content-manager/server/src/validation/zod.ts`. So there will be two different interpretations of the same `ZodError` depending on which helper you call; we should combine those (probably landing in utils) so that we keep it consistent and clean.

packages/core/content-manager/admin/src/utils/validationZod.ts
```diff
@@ -0,0 +1 @@
+export { getZodValidationErrors } from '@strapi/utils';
```
nclsndr: I guess we want to remove this re-export

packages/core/content-manager/server/src/controllers/validation/index.ts
```diff
@@ -1,49 +1,42 @@
...
+});
 
-const generateUIDInputSchema = yup.object({
-  contentTypeUID: yup.string().required(),
-  field: yup.string().required(),
-  data: yup.object().required(),
+const generateUIDInputSchema = z.object({
+  contentTypeUID: z.string(),
+  field: z.string(),
+  data: z.object({}).passthrough(),
```
nclsndr: passthrough is [deprecated](https://zod.dev/v4/changelog?id=deprecates-strict-and-passthrough).
We probably prefer the looseObject def or even [record](https://zod.dev/api?id=records)

packages/core/content-manager/server/src/controllers/validation/index.ts
```diff
@@ -1,49 +1,42 @@
...
+        if (value === '') return true;
+        const pattern = regex ? new RegExp(regex) : /^[A-Za-z0-9-_.~]*$/;
+        return pattern.test(value);
+      },
+      {
+        error: `Must match the custom regex or the default one "/^[A-Za-z0-9-_.~]*$/"`,
       }
     ),
-});
+  });
```
nclsndr: Maybe [string.customFormat](https://zod.dev/api?id=custom-formats) is more appropriate here

packages/core/utils/src/zod.ts
```diff
@@ -20,10 +37,33 @@ export const validateZod =
...
+ * Only the first error per path is kept (matches Yup behavior).
+ */
+export const getZodValidationErrors = (error: z.ZodError): FormErrors => {
+  const errors: FormErrors = {};
+
+  for (const issue of error.issues) {
+    const path = issue.path.join('.');
+    if (path && !(path in errors)) {
+      errors[path] = issue.message;
+    }
```
nclsndr, line 66: 1. What happens if the error happens at the root where `path = []`?
2. What happens if we have 2 errors for the same path? Does only taking the first is a good enough strategy?

packages/core/content-manager/server/src/controllers/validation/index.ts
butcherZ: I agree, but it seems like it only exists in the standalone zod v4 :( we can definitely flag this and revisit them once we are fully upgrade to the standalone version

package.json
```diff
@@ -94,7 +94,8 @@
     "@types/koa": "2.13.4",
     "@babel/core": "7.26.10",
     "@babel/runtime": "7.26.10",
-    "@babel/runtime-corejs3": "7.26.10"
+    "@babel/runtime-corejs3": "7.26.10",
+    "es-iterator-helpers": "1.3.2"
```
nclsndr: Why did we introduce this dep?

nclsndr (approved):

PR #25583, opened 2026-02-27
enhancement: allow right-clicking on table row to show context menu

### LINKED ISSUE #25619: Content Manager list view table rows use onClick navigation, prevents right-click / Cmd+click "Open in New Tab" standard link behavior
### Node Version

22.19.0

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.37.1

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Table rows in the Content Manager list view use onClick + useNavigate() for navigation instead of wrapping row content in <a> tags or React Router <Link> components. This prevents standard browser behaviors like right-click => "Open Link in New Tab", Cmd/Ctrl+click, and middle-click.

This is a continuation of #10356, which was partially fixed in v4.4.2 via PR #14528 but the underlying problem persists in Strapi 5.

### Steps to Reproduce

1. Go to any collection type list view
2. Right-click a row
3. Notice no "Open Link in New Tab" option appears in the context menu

### Expected Behavior

Each row should behave as a navigable link so all standard browser link interactions work.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

Image

### Additional information

_No response_

### Confirmation Checklist

mathildeleg:
### What does it do?
Support right-clicking on an entry table row to show the browser's context menu and open the entry in another tab for example.
https://github.com/user-attachments/assets/cffa7b79-61ec-43da-a6d7-f94600f19763

### Why is it needed?
Enhance user experience.

### How to test it?
* Create a content type with a field different types of fields, like a Relation and with localization set up. 
* Create an entry filling in those fields and creating at least in one another locale.
* Go to the list view and check you can right-click and open the entry in a new tab.
* Check that there are no regressions: checkbox can still be clicked, the Relation popover is still clickable and visible, the Locale dropdown is still clickable and visible (and clicking on a specific locale should redirect you to that specific locale's entry), the "..." menu is still clickable and usable. 
* If possible, check that there are no regressions on mobile/tablet too (right-clicking there is obviously not possible).

### Related issue(s)/PR(s)
https://github.com/strapi/strapi/pull/23028
Resolves https://github.com/strapi/strapi/issues/25619

### DIFF
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -17,6 +17,7 @@ import {
   useTable,
   useIsMobile,
   useIsDesktop,
+  useMediaQuery,
   tours,
 } from '@strapi/admin/strapi-admin';
 import {
@@ -69,6 +70,15 @@ const LayoutsHeaderCustom = styled(Layouts.Header)`
   overflow-wrap: anywhere;
 `;
 
+const RowLink = styled(ReactRouterLink)`
+  position: absolute;
+  left: 0;
+  top: 0;
+  bottom: 0;
+  width: 100%;
+  z-index: 1;
+`;
+
 const ListViewPage = () => {
   const { trackUsage } = useTracking();
   const navigate = useNavigate();
@@ -77,6 +87,8 @@ const ListViewPage = () => {
   const { _unstableFormatAPIError: formatAPIError } = useAPIErrorHandler(getTranslation);
   const isMobile = useIsMobile();
   const isDesktop = useIsDesktop();
+  // Only show the row link overlay (right-click context menu) on desktop with a mouse; tablet (even in landscape) and touch devices use row click
+  const showRowLinkOverlay = isDesktop && useMediaQuery('(pointer: fine)');
 
   usePersistentPartialQueryParams('STRAPI_LIST_VIEW_SETTINGS:', ['sort', 'filters', 'pageSize']);
   usePersistentPartialQueryParams('STRAPI_LOCALE', ['plugins.i18n.locale'], false);
@@ -264,13 +276,24 @@ const ListViewPage = () => {
         defaultMessage: 'Untitled',
       });
 
+  const getRowLinkTo = (id: Modules.Documents.ID) => ({
+    pathname: id.toString(),
+    search: stringify({ plugins: query.plugins }),
+  });
+
   const handleRowClick = (id: Modules.Documents.ID) => () => {
     trackUsage('willEditEntryFromList');
-    navigate({
-      pathname: id.toString(),
-      search: stringify({ plugins: query.plugins }),
-    });
+    navigate(getRowLinkTo(id));
   };
+
+  const getRowLinkProps = (row: (typeof results)[number]) => ({
+    to: getRowLinkTo(row.documentId),
+    onClick: () => trackUsage('willEditEntryFromList'),
+    'aria-label': formatMessage({
+      id: getTranslation('components.Table.open-entry'),
+      defaultMessage: 'Open entry',
+    }),
+  });
 
   const isEmptyState = !isFetching && results.length === 0;
 
@@ -396,15 +419,27 @@ const ListViewPage = () => {
                         <Table.Row
                           cursor="pointer"
                           key={row.id}
-                          onClick={handleRowClick(row.documentId)}
+                          style={{ position: 'relative' }}
+                          {...(!showRowLinkOverlay && { onClick: handleRowClick(row.documentId) })}
                         >
-                          <Table.CheckboxCell id={row.id} />
-                          {tableHeaders.map(({ cellFormatter, ...header }) => {
+                          <Table.CheckboxCell
+                            id={row.id}
+                            style={{ position: 'relative', zIndex: 2 }}
+                            onClick={
+                              !showRowLinkOverlay
+                                ? (e: React.MouseEvent) => e.stopPropagation()
+                                : undefined
+                            }
+                          />
+                          {tableHeaders.map(({ cellFormatter, ...header }, cellIndex) => {
                             if (header.name === 'status') {
                               const { status } = row;
 
                               return (
                                 <Table.Cell key={header.name}>
+                                  {showRowLinkOverlay && cellIndex === 0 && (
+                                    <RowLink {...getRowLinkProps(row)} />
+                                  )}
                                   <DocumentStatus status={status} maxWidth={'min-content'} />
                                 </Table.Cell>
                               );
@@ -415,6 +450,9 @@ const ListViewPage = () => {
                               // In this case, we display a dash
                               return (
                                 <Table.Cell key={header.name}>
+                                  {showRowLinkOverlay && cellIndex === 0 && (
+                                    <RowLink {...getRowLinkProps(row)} />
+                                  )}
                                   <Typography textColor="neutral800">
                                     {row[header.name.split('.')[0]]
                                       ? getDisplayName(row[header.name.split('.')[0]])
@@ -426,6 +464,9 @@ const ListViewPage = () => {
                             if (typeof cellFormatter === 'function') {
                               return (
                                 <Table.Cell key={header.name}>
+                                  {showRowLinkOverlay && cellIndex === 0 && (
+                                    <RowLink {...getRowLinkProps(row)} />
+                                  )}
                                   {/* @ts-expect-error – TODO: fix this TS error */}
                                   {cellFormatter(row, header, { collectionType, model })}
                                 </Table.Cell>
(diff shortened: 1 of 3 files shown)


packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -77,6 +87,8 @@ const ListViewPage = () => {
   const { _unstableFormatAPIError: formatAPIError } = useAPIErrorHandler(getTranslation);
   const isMobile = useIsMobile();
   const isDesktop = useIsDesktop();
+  // Only show the row link overlay (right-click context menu) on desktop with a mouse; tablet (even in landscape) and touch devices use row click
+  const showRowLinkOverlay = isDesktop && useMediaQuery('(pointer: fine)');
```
markkaylor: With the radix context menu we would get long press to open out of the box

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -396,15 +419,27 @@ const ListViewPage = () => {
                         <Table.Row
```
markkaylor, line 419: Could we not just wrap the Row in the Link from react-router-dom? Wouldn't that give use the context menu we want out of the box without positioning?

mathildeleg: Unfortunately, you can’t wrap the row in a Link because in HTML a <tr> can’t be wrapped by or have as its only child an <a>, and a single link inside one <td> only establishes a clickable area for that cell, so it can’t cover the whole row. This is the only solution I found that could work, and it seems to be the way it's usually done by others to allow right-clicking in a table row or else they don't actually use a `<Table>.`
However, I'm still not 100% sure this is the solution we want (even though it may be the only one) so I'm open to closing this PR (or to talk about it together) if you're not sure about it either!

markkaylor: Gotcha. I agree it's painful there is no way to open an entry in a new tab by right click or cmd+click. My comment was mainly to be sure we considered the simplest solution first. I'm not necessarily opposed to what you've done here. Maybe @Adzouz could also give an opinion here?

Adzouz:
The problem is a bit more complex here unfortunately. We synced with @mathildeleg and we have a few issues:
- Wrapping the whole row cells with a unique link is not an option (since we can have selects, other links, buttons, you name it) inside a cell a it's not valid for them to be children of a link element
- Putting a link in each cell above the text will make the text not selectable anymore (not hoverable) so the tooltip hover text elements won't work eithe.
- Putting a link above in each cell also has another inconvenient: we need to manually add a z-index on focusable elements (like selects, links, buttons etc) to move them above the link (what @mathildeleg has done basically in this PR). We control our own elements (so for them it's ok we can add the z-index) but if a plugin developer adds some custom cells to the list view with focusable elements, then without putting the z-index, they would become useless (because hidden behind the links).
- It's not very accessible in general (even the actual implementation of the list). A list should have the following roles: grid, row and gridcell for screen readers. We should be able to focus the list with the keyboard, navigate through it but also select focusable elements that are inside.
- The actual implementation of the tooltip on text elements is not working well. If the purpose of the tooltip was to be able to select the text and copy it, it's not currently working and selecting the text will just follow the link.

I tried to implement a mix with the link on the whole row but putting it behind, not above (with tabIndex 0 - z-index 0) + links wrapping text elements only (with tabIndex -1 and z-index 1, to be above the background link) to not impact any other focusable elements but the only problem with that solution is that if we right click on a cell that is not type text, we don't have open in a new tab option so it fixes the issue partially.

I have to admit that this requires compromises and discussing with UX to settle because it seems that there's no 100% good solution.

Happy to sync to discuss a bit more about the subject in details but in the meantime, I think it can be put on the side or implement my solution since it's still enhancing the experience on most cases.

What do you think @mathildeleg and @markkaylor? (sorry for the long text)

markkaylor:
> The problem is a bit more complex here unfortunately. We synced with @mathildeleg and we have a few issues:
> 
>     * Wrapping the whole row cells with a unique link is not an option (since we can have selects, other links, buttons, you name it) inside a cell a it's not valid for them to be children of a link element
> 
>     * Putting a link in each cell above the text will make the text not selectable anymore (not hoverable) so the tooltip hover text elements won't work eithe.
> 
>     * Putting a link above in each cell also has another inconvenient: we need to manually add a z-index on focusable elements (like selects, links, buttons etc) to move them above the link (what @mathildeleg has done basically in this PR). We control our own elements (so for them it's ok we can add the z-index) but if a plugin developer adds some custom cells to the list view with focusable elements, then without putting the z-index, they would become useless (because hidden behind the links).
> 
>     * It's not very accessible in general (even the actual implementation of the list). A list should have the following roles: grid, row and gridcell for screen readers. We should be able to focus the list with the keyboard, navigate through it but also select focusable elements that are inside.
> 
>     * The actual implementation of the tooltip on text elements is not working well. If the purpose of the tooltip was to be able to select the text and copy it, it's not currently working and selecting the text will just follow the link.
> 
> 
> I tried to implement a mix with the link on the whole row but putting it behind, not above (with tabIndex 0 - z-index 0) + links wrapping text elements only (with tabIndex -1 and z-index 1, to be above the background link) to not impact any other focusable elements but the only problem with that solution is that if we right click on a cell that is not type text, we don't have open in a new tab option so it fixes the issue partially.
> 
> I have to admit that this requires compromises and discussing with UX to settle because it seems that there's no 100% good solution.
> 
> Happy to sync to discuss a bit more about the subject in details but in the meantime, I think it can be put on the side or implement my way of doing it since it's still enhancing the experience on most cases.
> 
> What do you think @mathildeleg and @markkaylor? (sorry for the long text)

Thanks for digging into this. Lots of good findings.

Crazy (maybe stupid...or crazy stupid) idea, put an option in the 3 dots menu: "Open in new tab". Pro: avoids everything you just said 😆, con: not the intuitive thing users are expecting 😢

Adzouz:
> > The problem is a bit more complex here unfortunately. We synced with @mathildeleg and we have a few issues:
> > ```
> > * Wrapping the whole row cells with a unique link is not an option (since we can have selects, other links, buttons, you name it) inside a cell a it's not valid for them to be children of a link element
> > 
> > * Putting a link in each cell above the text will make the text not selectable anymore (not hoverable) so the tooltip hover text elements won't work eithe.
> > 
> > * Putting a link above in each cell also has another inconvenient: we need to manually add a z-index on focusable elements (like selects, links, buttons etc) to move them above the link (what @mathildeleg has done basically in this PR). We control our own elements (so for them it's ok we can add the z-index) but if a plugin developer adds some custom cells to the list view with focusable elements, then without putting the z-index, they would become useless (because hidden behind the links).
> > 
> > * It's not very accessible in general (even the actual implementation of the list). A list should have the following roles: grid, row and gridcell for screen readers. We should be able to focus the list with the keyboard, navigate through it but also select focusable elements that are inside.
> > 
> > * The actual implementation of the tooltip on text elements is not working well. If the purpose of the tooltip was to be able to select the text and copy it, it's not currently working and selecting the text will just follow the link.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I tried to implement a mix with the link on the whole row but putting it behind, not above (with tabIndex 0 - z-index 0) + links wrapping text elements only (with tabIndex -1 and z-index 1, to be above the background link) to not impact any other focusable elements but the only problem with that solution is that if we right click on a cell that is not type text, we don't have open in a new tab option so it fixes the issue partially.
> > I have to admit that this requires compromises and discussing with UX to settle because it seems that there's no 100% good solution.
> > Happy to sync to discuss a bit more about the subject in details but in the meantime, I think it can be put on the side or implement my way of doing it since it's still enhancing the experience on most cases.
> > What do you think @mathildeleg and @markkaylor? (sorry for the long text)
> 
> Thanks for digging into this. Lots of good findings.
> 
> Crazy (maybe stupid...or crazy stupid) idea, put an option in the 3 dots menu: "Open in new tab". Pro: avoids everything you just said 😆, con: not the intuitive thing users are expecting 😢

Yeah we could also do that but it needs to be validated first and also unfortunately it's not exactly answering to the base issue as you said 😢

Adzouz:
Hey @mathildeleg I'll close this PR if that's okay for you. After discussing with Emilie, we'll implement the link around the main field (if it's text) and add an option to open in a new tab in the three dots menu. So I think that's starting a brand new one would be better 😊 Thanks!

mathildeleg:
@Adzouz there's this community PR that might have the fix you're looking for (or at least partial fix) [https://github.com/strapi/strapi/pull/25621](<https://github.com/strapi/strapi/pull/25621>) 🙂 <br>Can you take care of it (either approve or close it if you don't like the implementation) since you're handling this issue now? Thanks 🙏🏻

Adzouz:
I'll take a look thanks!

Issue #11998, opened 2021-12-22
[V4] QueryEngine - deleteMany not working when filter by nested entity.


c7ming:
### Describe the bug

When trying to deleteMany by the nested entity filtering, An error occurs.

### Steps to reproduce the behavior

1. Have eneties A & B.
2. Add relation between A & B, and B has many A.
3. Execute codes 

### Expected behavior
All B filtered A will be deleted.

### Screenshots

![image](https://user-images.githubusercontent.com/12271187/147122112-f4ca0414-5f93-460b-a8c8-9fc02efd72b2.png)

### Code snippets

`await strapi.query(A).deleteMany({
        where: {
          B: {
            id: {
              $eq: params.id
            },
          }
        }
      })`

### System

- Node.js version: V14
- NPM version: 7.22.0
- Strapi version: 4.0.1
- Database: MySql
- Operating system: MacOs Monterey

iicdii:
I reproduced it with sqlite too
```js
[Error: delete from `addresses` where (`t2`.`id` = 1) - SQLITE_ERROR: no such column: t2.id] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
```

iicdii:
Temporary workaround. assume that
```
api::address.address - A
api::category.category - B (has many addresses)
```
```js
let addressItems;
try {
  addressItems = await strapi
    .query("api::address.address")
    .findMany({ where: { category: { id: 1 } } });
} catch (e) {
  console.error(e);
}

try {
  for (const address of addressItems) {
    await strapi
      .query("api::address.address")
      .delete({ where: { id: address.id } });
  }
} catch (e) {
  console.error(e);
}
```

rcdk:
I would also love to get this fixed :-)

antokhio:
Just came across this, with oneToOne relation, strapi version 4.3.2
Works: 
```
let tasks = await strapi.db.query("api::task.task").findMany({
  where: {
    parent: section.id,
  },
});
```
Doesn't work:
```
let tasks = await strapi.db.query("api::task.task").deleteMany({
  where: {
    parent: section.id,
  },
});
```

Error message: 
```
error: delete from "tasks" using "tasks_parent_links" as "t1","sections" as "t2" where ("t2"."id" = $1) and "t0"."id" = "t1"."task_id" and "t1"."section_id" = "t2"."id" - missing FROM-clause entry for table "t0"
[...]

PR #25590, opened 2026-03-01
fix(database): process deleteMany relation filters via query init

sirrodgepodge:
## Summary
- fix `deleteMany` to initialize the query builder with `{ where }` via `.init(...)`
- ensures nested relation filters are processed consistently before delete execution

## Context
Issue: #11998 (`deleteMany` not working when filtering by nested entity)

Previously `deleteMany` used `.where(where)` directly. Switching to `.init({ where })` aligns it with the pattern used in other query paths and ensures where params are fully initialized before delete.

### DIFF
--- a/packages/core/database/src/entity-manager/index.ts
+++ b/packages/core/database/src/entity-manager/index.ts
@@ -482,7 +482,7 @@ export const createEntityManager = (db: Database): EntityManager => {
       const { where } = params;
 
       const deletedRows = await this.createQueryBuilder(uid)
-        .where(where)
+        .init({ where })
         .delete()
         .execute<number>({ mapResults: false });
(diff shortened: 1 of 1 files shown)


innerdvations:
Thanks for the PR! I took a quick look at this and while I would prefer to use `init` since that's the pattern other methods here use, I couldn't reproduce any case where it actually fixed anything (and I'm hesitant to accept a core db change just for matching patterns with other methods).

Do you have a test case (either manual or an automated test) I can try to see if this fixes the referenced issue?

innerdvations:
Thanks for the PR! I'm closing in favor of the alternative https://github.com/strapi/strapi/pull/25420

PR #25600, opened 2026-03-02
Chore: Content manager performance optimizations

jhoward1994:
## What does it do?

This PR reduces unnecessary computation in Strapi v5's content-manager, improving overall performance.

**Core optimizations included:**

- **getMetadata Fast Path:**  
  For non-localized content types, an expensive `findMany` query (fetching all document versions for all locales/statuses) is replaced by a targeted `findOne` to retrieve only the opposite-status document. Skipping locale processing where it's unnecessary.

- **Populate Caching:**  
  Results from `getDeepPopulateDraftCount`, `getPopulateForValidation`, and `buildDeepPopulate` are now cached in module-scoped Maps. These previously recomputed on every request, leading to unnecessary overhead.

- **findPage Count Optimization:**  
  Strips out `populate` and `sort` parameters before calling `count()`, improving efficiency since `COUNT(*)` does not utilize these fields.

- **delete / publishMany Streamlining:**  
  Removes redundant calls to `buildDeepPopulate` during delete operations (when its result isn’t used). For `publishMany`, supplies an empty `populate` parameter as results do not require population.

- **List View Improvements:**  
  - Skips `getManyAvailableStatus` query entirely if Draft & Publish is disabled.
  - Uses `db.query.count` directly in place of the more expensive `findMany` for bulk action checks.
  - Avoids executing queries for empty localizations arrays.

## Why is it needed?

Strapi v5's content-manager has shown significantly slower performance than v4, particularly for deeply nested content types.

Despite prior optimizations (e.g., localizations populate validation), redundant queries, excessive population traversals, and repeated computation of schema helpers persisted. This PR directly targets and remedies these lingering inefficiencies.

**Performance Impact:**
Together with #25557, these changes were benchmarked against npm Strapi v5 (5.37.0) using complex schemas reported to be performing poorly 

```
| Endpoint            | Content Type | v4   | v5 stock | v5 patched | Faster than stock v5 | v5/v4 ratio (stock → patched) |
|---------------------|--------------|------|----------|------------|----------------------|-------------------------------|
| List (pageSize=50)  | CT1          | 35ms | 78ms     | 61ms       | -22%                 | 2.2x → 1.6x                   |
| List (pageSize=50)  | CT2          | 31ms | 87ms     | 51ms       | -41%                 | 2.8x → 1.5x                   |
| List (pageSize=50)  | CT3          | 34ms | 85ms     | 57ms       | -33%                 | 2.5x → 1.7x                   |
| FindOne (edit view) | CT1          | 20ms | 59ms     | 48ms       | -19%                 | 3.0x → 2.7x                   |
| FindOne (edit view) | CT2          | 20ms | 48ms     | 38ms       | -21%                 | 2.4x → 1.7x                   |
| FindOne (edit view) | CT3          | 21ms | 66ms     | 54ms       | -18%                 | 3.1x → 2.8x                   |
```

- **List view**: Largest performance improvements (22-41% faster than npm latest v5). The v5/v4 overhead ratio drops from 2.2-2.8x to 1.5-1.7x—closing roughly half of the gap with v4.
- **Edit view**: Improves by ~20% due to the getMetadata fast path, which replaces a full-table scan with a single-row lookup. The v5/v4 ratio improves from 2.4-3.1x to 1.7-2.8x.

> Note: `countDraftRelations` showed negligible change. For the most complex content types, the v5/v4 ratio remains high. Further improvements to the document-service could be needed (parameter validation, transforms) and not covered by this PR.

## How to test

1. **Run unit tests**  
   ```bash
   cd packages/core/content-manager && yarn test:unit
   ```

2. **Manual verification**
   - Confirm the List view functions as expected.
   - Confirm the Edit view works correctly.
   - Draft relation counts are accurate for:
     - Non-localized types (both with Draft & Publish enabled and disabled).
     - Localized types (with Draft & Publish enabled).
   - Bulk publish/unpublish/delete actions succeed.

> **ℹ️  It is recommended to test this in combination with PR #25557** — both contribute to the overall performance improvements of the content manager.

fixes CMS-459

### DIFF
--- a/packages/core/content-manager/server/src/services/document-manager.ts
+++ b/packages/core/content-manager/server/src/services/document-manager.ts
@@ -200,7 +200,7 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
           documentIds.map((id) =>
             strapi
               .documents(uid)
-              .unpublish({ ...opts, documentId: id })
+              .unpublish({ ...opts, documentId: id, populate: {} })
               .then((result) => result?.entries)
           )
         );
--- a/packages/core/content-manager/server/src/services/document-metadata.ts
+++ b/packages/core/content-manager/server/src/services/document-metadata.ts
@@ -46,6 +46,10 @@ const AVAILABLE_LOCALES_FIELDS = [
   'publishedAt',
 ];
 
+/** Returns a DB filter that matches the opposite publish status. */
+const oppositePublishStatus = (publishedAt: unknown) =>
+  publishedAt !== null ? { $null: true } : { $notNull: true };
+
 const CONTENT_MANAGER_STATUS = {
   PUBLISHED: 'published',
   DRAFT: 'draft',
@@ -242,7 +246,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
             where: {
               documentId: version.documentId,
               ...(version.locale ? { locale: version.locale } : {}),
-              publishedAt: version.publishedAt !== null ? { $null: true } : { $notNull: true },
+              publishedAt: oppositePublishStatus(version.publishedAt),
             },
             select: AVAILABLE_STATUS_SCALAR_FIELDS,
             populate: AVAILABLE_STATUS_POPULATE,
@@ -368,7 +372,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
           (v) =>
             v.documentId === d.documentId &&
             v.locale === d.locale &&
-            (d.publishedAt !== null ? v.publishedAt === null : v.publishedAt !== null)
+            (d.publishedAt === null) !== (v.publishedAt === null)
         );
         return {
           ...d,


packages/core/content-manager/server/src/services/document-metadata.ts
```diff
@@ -214,6 +223,40 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
...
+      const otherVersion = availableStatus
+        ? await strapi.db.query(uid).findOne({
+            where: {
+              documentId: version.documentId,
+              ...(version.locale ? { locale: version.locale } : {}),
+              publishedAt: version.publishedAt !== null ? { $null: true } : { $notNull: true },
+            },
+            select: AVAILABLE_STATUS_SCALAR_FIELDS,
+            populate: AVAILABLE_STATUS_POPULATE,
+          })
```
markkaylor, line 253: This is really not fun to read 😆 I wonder if we could find a simpler way to express this, maybe with a function?

packages/core/content-manager/server/src/services/document-metadata.ts
```diff
@@ -323,19 +360,22 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
...
       document.localizations = document.localizations.map((d) => {
-        const status = otherStatus.find(
-          (s) => s.documentId === d.documentId && s.locale === d.locale
+        // Find the counterpart version (same documentId + locale, opposite publishedAt) from
+        // the already-fetched versions array, avoiding an extra DB query.
+        const counterpart = versions.find(
+          (v) =>
+            v.documentId === d.documentId &&
+            v.locale === d.locale &&
+            (d.publishedAt !== null ? v.publishedAt === null : v.publishedAt !== null)
```
markkaylor: Another potentially simpler option I think:

`(d.publishedAt === null) !== (v.publishedAt === null)`

jhoward1994:
> I if I have relation and that relation is published I still get count: 1 for has draft relations. Is that right?

 This looks like a pre-existing behaviour
 - countDraftRelations fetches the parent document's draft version
 - The filter publishedAt: { $null: true } counts draft entries of related docs
 - Every document always has a draft entry (publishedAt: null), even when published

 I'll make a note of this, it's unrelated to this PR but something we should double check

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -200,7 +200,7 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
           documentIds.map((id) =>
             strapi
               .documents(uid)
-              .unpublish({ ...opts, documentId: id })
+              .unpublish({ ...opts, documentId: id, populate: {} })
```
markkaylor, line 203: Is this required due to something I changed in the common branch?

packages/core/content-manager/server/src/services/document-metadata.ts
markkaylor: In retrospect this comment was dumb. I thought it was a nested ternary but I see now inline on the object it was actually quite simple. But anyway fine with the function too but probably unnecessary thanks for updating anyway.

PR #25617, opened 2026-03-03
fix(content-manager): reduce rerenders for conditional fields

Bassel17:
### What does it do?

This PR reduces unnecessary re-renders for conditional fields in the content-manager.

Technical changes:
- updates `InputRenderer.tsx` so conditional fields subscribe only to the form values used by their visibility condition instead of subscribing to the full form state
- applies the same conditional visibility optimization to`VersionContent.tsx`
- extracts the shared conditional dependency parsing/subscription helpers into `utils/conditionalFields.ts`

The goal is to keep conditional fields reactive to the fields they depend on, while avoiding rerenders when unrelated fields change.

### Why is it needed?

Before this change, conditional fields were effectively tied to broad form state updates, so changing any field could cause unrelated conditional fields to re-render.

On complex edit views, especially with many conditional fields, this creates avoidable rendering work and can contribute to visible input lag.

This PR narrows those subscriptions so conditional fields only re-render when:
- their own visibility dependencies change
- or we cannot safely infer dependencies and need to fall back to the broader subscription for correctness

### How to test it?

Environment:
- run the admin locally
- use a content type with conditional fields in the edit view
- ideally use a setup with multiple conditional fields depending on different source fields

Edit view checks:
1. Open an entry in the content-manager edit view.
2. Change a field that is **not** referenced by a given conditional field.
3. Confirm unrelated conditional fields do not re-render unnecessarily and the input remains responsive.
4. Change a field that **is** referenced by a conditional visibility rule.
5. Confirm the dependent conditional field updates visibility correctly.

History view checks:
1. Open the content history/version view for an entry with conditional fields.
2. Confirm fields with visibility conditions still render correctly.

Fixes CMS-383

### DIFF
--- a/packages/core/admin/admin/src/components/Form.tsx
+++ b/packages/core/admin/admin/src/components/Form.tsx
@@ -167,6 +167,10 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
       isSubmitting: false,
       values: props.initialValues ?? {},
     });
+
+    // Keep a ref to the latest form values so `getValues()` can always return fresh data.
+    // We expose `getValues` as a stable callback so consumers (e.g. conditional/rules logic)
+    // can call it without causing extra rerenders from changing function references.
     const valuesRef = React.useRef(state.values);
 
     React.useEffect(() => {
--- a/packages/core/content-manager/admin/src/utils/tests/conditionalFields.test.ts
+++ b/packages/core/content-manager/admin/src/utils/tests/conditionalFields.test.ts
@@ -1,0 +1,92 @@
+import {
+  CONDITIONAL_FIELD_MULTI_DEP_SEPARATOR,
+  CONDITIONAL_FIELD_STATIC_SUBSCRIPTION,
+  getConditionDependencyPaths,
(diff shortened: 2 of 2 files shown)


packages/core/admin/admin/src/components/Form.tsx
```diff
@@ -163,6 +167,13 @@ const Form = React.forwardRef<HTMLFormElement, FormProps>(
       isSubmitting: false,
       values: props.initialValues ?? {},
     });
+    const valuesRef = React.useRef(state.values);
+
+    React.useEffect(() => {
+      valuesRef.current = state.values;
+    }, [state.values]);
+
+    const getValues = React.useCallback(() => valuesRef.current, []);
```
markkaylor, line 180: Could we get a comment explaining what's going on here.

packages/core/content-manager/admin/src/utils/conditionalFields.ts
markkaylor, line 1: unit test?

markkaylor (approved):

Issue #25111, opened 2026-01-04
Displayed field unavailable as number when set on a numeric field of a relation


claudiobgit:
### Node Version

24.12.0

### NPM/Yarn/PNPM Version

NPM 11.6.2

### Strapi Version

5.33.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

It's impossible to set the displayed field on numeric in a relation as Label / Entry title in configuration view.

### Steps to Reproduce

1.  Create a first collection type for a sequential product, let's call it 'Publication Issue"
2.  Add a field of type Number (integer,) and set it as Required and Unique to the first collection

<img width="822" height="355" alt="Image" src="https://github.com/user-attachments/assets/2a8f7ec8-435e-4a62-b5e7-6d2b770712b9" />

3.  Add some data to the collection in the Content manager (some numbers like 1, 2 and 3)
4.  Create a second collection type, let's call it 'Article'
5.  Add a field of type Relation, named 'Publication Issue' and relate it to the first collection created

<img width="832" height="414" alt="Image" src="https://github.com/user-attachments/assets/1cc16fb9-dd6a-407f-b0c7-7cdf0b28f029" />

6.  Configure the view for the second collection and set Label / Entry title to 'pub_number'

<img width="832" height="439" alt="Image" src="https://github.com/user-attachments/assets/e323e059-461a-4b1f-9ee9-45d096464225" />

7.  Go to the Content Manger / Article and where you expect to have a menu where you can select numbers like 1, 2 or 3 you find the `documentId` instead of `pub_number` as expected; that makes it almost impossible for the user to understand what issue number he is selecting for the article

<img width="264" height="137" alt="Image" src="https://github.com/user-attachments/assets/26061e18-8ef4-4b21-942b-edc932cadf58" />

### Expected Behavior

User should be able to visualize and select the numbers of the configured field (ex. 1, 2 and 3)

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

claudiobgit:
I confirm this issue is to be considered solved. 
I tested now the behaviour on Strapi v5.51.1 and it works as expected, showing the numbers.

PR #25622, opened 2026-03-04
fix(content-manager): documentId(s) shown for relation when entry title set to numeric field

molund:
### What does it do?

This change updates the `getRelationLabel()` function so it can use numeric fields as entry titles, not just string fields.  Numeric fields are cast to strings so the contract isn't violated. 

I wasn't sure why numeric fields are no longer allowed in Strapi 5. So I decided to block usage of the `id` field because its internal nature might be the reason behind this design decision.  

### Why is it needed?

Strapi 4 allowed numeric fields as entry titles and this is broken in Strapi 5.

### How to test it?

Add a relation to a collection containing a numeric field. Set the entry title to the numeric field.  Ensure that the number appears in the relation picker combobox instead of the documentId. 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/25111


innerdvations (approved):
LGTM, thanks! I added some unit tests to prevent regressions

PR #25623, opened 2026-03-04
fix(content-manager): reduce excessive rerendering in relation fields

Bassel17:
### What does it do?

This PR reduces unnecessary re-renders in the content-manager relation fields.

### Why is it needed?

Before this change, relation fields could re-render on unrelated field updates.

That was particularly expensive because the relation field contains a combobox/modal flow

On larger edit views, this extra rendering work can contribute to visible input lag and poorer responsiveness.

This PR reduces that work by making the expensive relation list subtree more stable and allowing memoization to actually be effective.

### How to test it?

Environment:
- run the admin locally
- use a content type that contains at least one relation field
- ideally test on an entry with multiple relations already connected

Path:
- Content Manager -> open any collection type edit view with a relation field

Checks:
1. Open an entry that has relation fields.
2. Edit unrelated fields in the form.
3. Confirm the relation field no longer re-renders excessively on every unrelated change.
4. Add a relation through the relation combobox/modal and confirm behavior is unchanged, also the excessive re-rendering in the relations of the fly modal should be gone
5. Remove a relation and confirm behavior is unchanged.
6. If possible, use React Scan or local render logging to compare rerender frequency before/after while typing in another field.

### DIFF
--- a/tests/e2e/tests/content-manager/history.spec.ts
+++ b/tests/e2e/tests/content-manager/history.spec.ts
@@ -452,10 +452,12 @@ describeOnCondition(edition === 'EE')('History', () => {
       await page.getByRole('combobox', { name: 'Authors' }).click();
       await page.getByText('Coach Beard').click();
       await page.getByRole('button', { name: 'Save' }).click();
+      await findAndClose(page, 'Saved Document');
 
       // Delete one of the authors, leaving only Coach Beard
       // Open the relation modal
       await clickAndWait(page, page.getByRole('button', { name: 'Will Kitman' }));
+      await expect(page.getByRole('button', { name: 'Go to entry' })).toBeVisible();
       // Click to go to the related document
       await clickAndWait(page, page.getByRole('button', { name: 'Go to entry' }));
       await page.getByRole('button', { name: /more actions/i }).click();


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -246,19 +255,19 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
     };
 
     const field = useField(props.name);
+    const serverData = data?.results ?? EMPTY_RELATION_RESULTS;
```
Adzouz, line 244: Since it's only used here, defining a `EMPTY_RELATION_RESULTS` constant is really necessary?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -998,6 +1038,39 @@ const RelationsList = ({
         RELATION_ITEM_HEIGHT / 2
       : Math.min(data.length, RELATIONS_TO_DISPLAY) * (RELATION_ITEM_HEIGHT + RELATION_GUTTER);
 
+  const itemData = React.useMemo(
```
Adzouz, line 1019: Regarding the number of dependencies to this useMemo, I have the feeling that this wouldn't really be helpful. What are your thoughts? Is there a really gain of doing that I don't see?

Bassel17:
> I tested and it looks way better, good job 🙏
> 
> However, there's one case where rerenders still occurs on updates on unrelated fields, it's when the relation field is conditionally showed based on another field (see my video below) 😢
> 
>  Content-Type.Builder._.Strapi.-.5.March.2026.mp4 
> Do you think you could try to find a solution for that case? 🙏

yep the fix for this is in the other PR concerning conditional fields, it should do the job

Issue #25632, opened 2026-03-04
Pupulating more than 100 fields breaks validation


hood:
### Technical info

(Omitted technical info as this affects all Strapi versions, regardless of underlying stack)

### Reproduction URL

[Codesandbox](https://codesandbox.io/p/devbox/q39dsd)
(NOTE: This is not a Strapi reproduction, but a simple Node.js sandbox to show how `qs`, in its Strapi configuration, behaves differently when handling arrays with less/more than 100 entries.

### Bug Description

When a populate query parameter has more than 100 entries, the `qs` parser (configured with `arrayLimit: 100` in `packages/core/core/src/middlewares/query.ts:9`) produces a plain object with numeric string keys instead of an array.

The `populate` traversal treated this as a regular object, visiting each numeric key. Keys `”0"` and `“1”` accidentally pass a boolean-like check (`parseBoolean` accepts them), but `”2"` and above fail, with `2` simply being the first to fail, throwing `ValidationError: Invalid key 2`.

### Steps to Reproduce

One can simply repro this by hitting Strapi with something like this:
```sh
[...]

PR #25633, opened 2026-03-04
fix: support more than 100 populate entries

hood:
### What does it do?

Applied a fix to `packages/core/utils/src/traverse/query-populate.ts`:
* Added `isArrayLikeObject`: detects objects where all keys are numeric strings (the `qs` `arrayLimit` artifact)
* Added an intercept that converts these array-like objects back to actual arrays via `Object.values()`, then recurses so the existing string-array handling takes over

dded related tests to `packages/core/utils/src/__tests__/query-populate.test.ts`:
* Added a `validatePopulate` test that passes an object with numeric keys `({ "0": "createdBy", "1": "updatedBy", "2": "createdBy" })` and asserts it doesn't throw.

### Why is it needed?

Current behaviour breaks when more than 100 entries are requested via `populate` query parameters.

### How to test it?

Fetch data from Strapi requesting more than 100 entries via `populate` query parameters.

### Related issue(s)/PR(s)

Fixes issue #25632

### DIFF
--- a/packages/core/utils/src/__tests__/query-populate.test.ts
+++ b/packages/core/utils/src/__tests__/query-populate.test.ts
@@ -1,4 +1,11 @@
 import { traverseQueryPopulate } from '../traverse';
+import {
+  validatePopulate,
+  POPULATE_TRAVERSALS,
+  FILTER_TRAVERSALS,
+  SORT_TRAVERSALS,
+  FIELDS_TRAVERSALS,
+} from '../validate/validators';
 
 describe('traverseQueryPopulate', () => {
   global.strapi = {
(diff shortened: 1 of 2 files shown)


innerdvations:
THANK YOU! This was a mystery for years but I never had time to look at it 😂

That said, you gave me an excuse to dig into this deeper, and it turns out this limit is in `qs` for a reason: trying to parse an unlimited number of params would make it easier to DoS the server. So we don't want to "fix" this by making it unlimited. However, it should NOT be a mystery, we need to document the solution:

```
// config/middlewares.ts
// ...
// replace 'strapi::query' with:
  { name: 'strapi::query', config: { arrayLimit: 200 } },
// ...
```

That should work on any version of Strapi without the need for this PR and allow users to choose their own level of risk.

I'm also not sure that's even the best solution even though it works. We might want to add something friendlier. Open to suggestions :)

hood:
@innerdvations 

For sure documenting a solution that’s already there makes a lot of sense.

If we want to be able to explictly limit users to a pre-set amount of parameters, though, I would throw an **explicit** runtime error every time they exceed the expected amount, rather than letting the parser fall into unexpected behaviour, as the resulting error is extremely misleading as of right now (it throws with something like “invalid key ‘2’”, as key "2" is always the first to blow up, because it's the lowest numeric string that isn't a valid boolean-like value). In the end it’s either a hard limit or not, what we want to avoid is anything in between.

Thanks for giving me your insights.

innerdvations:
You're right, we definitely should throw an error. I'm adding that to our backlog, although if you want to open a PR fixing that, it would be very appreciated!

For now I'll close this PR though since it seems like this isn't the way to go. Thanks again!

Issue #22946, opened 2025-02-19
Strapi 5 local plugins : size has significantly increased


mathieugrwn:
### Node Version

20.11.1

### NPM/Yarn/PNPM Version

1.22.22

### Strapi Version

5.9.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

Hello,

I recently upgraded from Strapi 4 to Strapi 5.

I'm using four local plugins that are part of the project.

I kept the same configuration in config/plugins.ts:

```
export default ({ env }) => ({
  ...
  'plugin-name': {
    enabled: true,
    resolve: './src/plugins/plugin-name',
  },
});
```

The issue is that the size of each plugin has significantly increased after migrating to Strapi 5 (from ~300mo to 1.6go of each node_modules plugin).

This is causing deployment issues for the application.

My question is: Is there a better way to handle local plugins in Strapi 5?

Thanks in advance for your help!

### Steps to Reproduce

1. Create a Strapi 5 project
2. Create a local plugin

### Expected Behavior

The expected behavior is that the plugin's node_modules directory is smaller or that another method exists to have multiple local plugins.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

derrickmehaffy:
![Image](https://github.com/user-attachments/assets/4f246a9e-55e7-44ed-bf50-316603d74753)

sehrish30:
I’m also running into this issue after migrating from Strapi v4 to v5.

- In v4, my project built fine without any memory flags.
-  In v5, I **must** set `NODE_OPTIONS="--max-old-space-size=4096"` or the build crashes with a heap out-of-memory error.
- My project size increased from **~3.04 GB to ~6.04 GB** after the upgrade.


This happens both locally and inside Docker builds. Even a clean install with `--omit=dev` results in high memory usage during build, especially local when plugins are involved.

This makes the local developer experience worse and complicates CI/CD pipelines without adding manual memory configuration.

<img width="548" alt="Image" src="https://github.com/user-attachments/assets/98b50c85-c681-47c9-93f7-940e61d89811" />

Would love to see optimizations around memory usage in the build process.

sehrish30:
[...]

PR #25636, opened 2026-03-04
fix(vite): deduplicate @strapi/* packages to prevent bundle bloat from local plugins

nielskaspers:
## Summary

Relates to #22946

Local plugins that include `@strapi/*` packages as direct dependencies (rather than peer dependencies) can cause Vite to resolve and bundle multiple copies of those packages — one from the root `node_modules` and one from the plugin's own `node_modules`. This increases build size and memory usage.

Vite already deduplicates React and related packages via `resolve.dedupe` to prevent the [duplicate React hook problem](https://react.dev/warnings/invalid-hook-call-warning#duplicate-react). This PR extends the same mechanism to cover the core `@strapi/*` packages so that Vite always resolves a single copy from the root `node_modules` regardless of what a local plugin lists in its own dependencies.

**Packages added to `dedupe`:**
- `@strapi/strapi`
- `@strapi/admin`
- `@strapi/design-system`
- `@strapi/icons`
- `@strapi/utils`

> **Note:** `resolve.dedupe` addresses bundling (the Vite build step). For the underlying disk-space issue caused by npm/yarn installing `@strapi/*` as direct dependencies inside a plugin's `node_modules`, the recommended solution is to declare those packages as `peerDependencies` only in local plugins, which is already the pattern in the `todo-example` plugin.

## Changes

- `packages/core/strapi/src/node/vite/config.ts`: extend `resolve.dedupe` with `@strapi/*` core packages

## Test plan

- [ ] Create a local plugin with `@strapi/strapi` as a direct `dependency` (not peer dep)
- [ ] Run `strapi build` — verify no duplicate `@strapi/*` module warnings from Vite
- [ ] Confirm the admin panel build still works correctly

### DIFF
--- a/packages/core/strapi/src/node/vite/config.ts
+++ b/packages/core/strapi/src/node/vite/config.ts
@@ -102,7 +102,20 @@ const resolveBaseConfig = async (ctx: BuildContext): Promise<InlineConfig> => {
     },
     resolve: {
       // https://react.dev/warnings/invalid-hook-call-warning#duplicate-react
-      dedupe: ['react', 'react-dom', 'react-router-dom', 'styled-components'],
+      // Also deduplicate @strapi/* packages so local plugins that list them as
+      // direct dependencies don't pull in a second copy during the build.
+      // See: https://github.com/strapi/strapi/issues/22946
+      dedupe: [
+        'react',
+        'react-dom',
+        'react-router-dom',
+        'styled-components',
+        '@strapi/strapi',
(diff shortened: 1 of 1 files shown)


derrickmehaffy:
it looks like this was superseeded by a much larger refactor; ode/core/admin-vite-alias-modules.ts now owns resolve.dedupe plus explicit aliases and pnpm-isolation handling so going to mark as closed

PR #25637, opened 2026-03-04
feat(database): export lifecycle event type

tewson:
### What does it do?

Export the life cycle `Event` type for type safety in custom lifecycle files.

### Why is it needed?

When writing a custom `src/api/<content type>content-types/<content type>/lifecycles.ts` file, it is currently difficult to add type safety to lifecycle functions. We can see people trying to grab the `Event` type directly from `@strapi/database/dist/lifecycles`: [example 1](https://github.com/incubateur-ademe/nosgestesclimat-cms/blob/41291c525f469214c7d124d14accc17694839075/src/api/faq/content-types/faq/lifecycles.ts#L1), [example 2](https://github.com/wizbii/strapi-plugin-strapi-algolia/blob/6f6f5d6394adb7593ec4e370b9a6b17a66a1884b/utils/event.d.ts#L1).

It might be nicer for the lifecycle `Event` type to be exported from the package index.

### How to test it?

The built `@strapi/database` should have the `Event` type exported.

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/docs/docs/docs/01-core/database/04-lifecycles.md
+++ b/docs/docs/docs/01-core/database/04-lifecycles.md
@@ -9,7 +9,7 @@ tags:
 
 Content-type lifecycle hooks (`src/api/<api>/content-types/<content-type>/lifecycles.ts`) receive a database lifecycle `Event` object.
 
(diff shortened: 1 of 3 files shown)


innerdvations (approved):
Thanks!

PR #25656, opened 2026-03-05
ci: sonarqube should not block merge

innerdvations:
### What does it do?

removes sonarqube from being a ci gate for merging

### Why is it needed?

we don't trust it enough yet and it's currently blocking everything

### How to test it?

### Related issue(s)/PR(s)


innerdvations:
Maybe this isn't necessary and it was just the github issues yesterday

PR #25657, opened 2026-03-05
feat(admin): api token supports admin permissions and admin user ownership

nclsndr:
### What does it do?

Splits API tokens into two distinct **kinds** via a `kind` discriminant field (`'content-api' | 'admin'`):

- **Content API tokens** — unchanged behavior. Carry content-API route permissions, no owner, no admin permissions.
- **Admin tokens** — new MCP-oriented tokens. Always have an owner (`adminUserOwner`), carry admin permissions scoped by the owner's permission ceiling, authenticate against admin routes.

Key additions:
- **New `admin::api-token` DB schema fields**: `kind` (enum, default `'content-api'`), `adminPermissions` (1-to-many to `admin::permission`), `adminUserOwner` (many-to-1 to `admin::user`).
- **Dedicated `/admin/admin-tokens` router + controller** with independent RBAC (`admin::admin-tokens.*` actions).
- **Admin permission ceiling**: non-super-admins can only assign token permissions within their own permission scope; conditions are inherited, not chosen.
- **Owner-only key access**: only the token owner can read `accessKey` or call `regenerate` on an admin token. Super-admins cannot bypass this.
- **Role → token sync**: when a role's permissions change, a user's roles change, or a role is deleted, all admin tokens owned by affected users are automatically re-clamped to the new ceiling.
- **Owner deactivation guard**: admin tokens are rejected at auth time if the owner has `isActive === false` or `blocked === true`.
- **Ceiling-aware permission matrix** in the UI (same component as the Roles page), with checkboxes disabled and conditions read-only for non-super-admins.
- **Kind-segregated settings pages**: `/settings/api-tokens` (content-api only) and `/settings/admin-tokens` (admin only), each with their own list, create, and edit views.
- **TypeScript discriminated union** at the contract level (`ContentApiApiToken | AdminApiToken`) for compile-time narrowing.

### Why is it needed?

Strapi needs a secure, permission-scoped token type for programmatic admin access (e.g. MCP integrations). The existing API tokens only covered the content API and had no concept of ownership or admin permission delegation. This introduces admin tokens that can authenticate against admin routes while being strictly bounded to the permission scope of their owner — preventing privilege escalation and ensuring tokens are automatically invalidated or trimmed when the owner's roles change.

### How to test it?

**Environment**: standard Strapi dev setup with at least two admin users (one super-admin, one regular admin with limited permissions) and a few content types.

**Smoke tests (high signal):**

1. **Kind enforcement**
   - `POST /api-tokens` with `kind: 'admin'` + `type: 'read-only'` → expect `ValidationError`
   - `POST /api-tokens` with `kind: 'content-api'` + `adminPermissions` → expect `ValidationError`

2. **Admin token creation**
   - Log in as a non-super-admin, navigate to `/settings/admin-tokens` → "Add new Admin Token" opens form with admin permission matrix; no token type selector visible.
   - Enable some permissions (only those the user holds are selectable) → save → reopen → same permissions pre-checked.

3. **Owner ceiling**
   - As non-super-admin, try to enable a permission you don't hold → checkbox stays unchecked.
   - As super-admin editing another user's token → matrix is bounded to the **owner's** permission scope, not the super-admin's.

4. **Owner-only key access**
   - As owner: `GET /admin-tokens/:id` returns `accessKey`; "View token" visible in UI.
   - As super-admin viewing another user's token: no `accessKey` in response; Regenerate button hidden.
   - `POST /admin-tokens/:id/regenerate` as super-admin on another user's token → `403`.

5. **Role → token sync**
   - Remove a permission from a role → admin tokens owned by users of that role have the corresponding permission deleted automatically.

6. **Owner deactivation**
   - Deactivate a user who owns an admin token → using the token returns `401 Token owner is deactivated`.

7. **List segregation**
   - `/settings/api-tokens` → only content-api tokens; no Kind/Owner column.
   - `/settings/admin-tokens` → only admin tokens; Owner column present.

8. **Content-api back-compat**
   - Existing content-api tokens (including those with a `null` `kind` DB column) appear in `/settings/api-tokens` and behave exactly as before.


markkaylor:
1. Kind enforcement
With a super admin JWT I am able to POST to the /admin/api-tokens but with the JWT retrieved as a non-super-admin I am rejected with 403.

2. Admin token creation
Now I gave my non-super-admin access to api tokens. I see the role does NOT have access to article.
Screenshot 2026-03-12 at 09 47 52
I log in as author. I go to add a new api token. I the api token to "custom". I can see I have the ability to create a token with "article" permissions:
Screenshot 2026-03-12 at 09 49 02
I give permissions to article and permissions to something else I should have permissions to like "likes". I make requests to the content-api for both articles and likes. Both are rejected with 403. I set the token to full access. Make requests to both endpoints. Both are successful. 

3. Owner ceiling
As a non-super admin I am able to see and select permissions my role does NOT hold, I save and still see my selection, I refresh and see my selection was not saved. Bad UX. I'm actually unable to save any specific permissions on the token with this role. Even as a super-admin I am unable to update an existing custom token. Hope this isn't a current bug 😬 
As a non-super admin I see all api tokens, including a full access token. I can view that token copy it and make requests to content-types my role does not have permissions for. 


Maybe I'll stop here for now because I feel like maybe something isn't right or I'm testing wrong 😄

nclsndr:
@markkaylor i feel like you've been testing "API Tokens" which are: _unchanged behavior. Carry content-API route permissions, no owner, no admin permissions._
The new kind is actually "Admin Tokens" down in the settings sidebar

markkaylor:
> @markkaylor i feel like you've been testing "API Tokens" which are: _unchanged behavior. Carry content-API route permissions, no owner, no admin permissions._ The new kind is actually "Admin Tokens" down in the settings sidebar

yup, sorry my bad I had a feeling I was doing something wrong. I'll start testing again

packages/core/admin/server/src/services/permission/engine.ts
```diff
@@ -76,6 +76,14 @@ export default (params: { providers: engine.EngineParams['providers'] }) => {
       return engine.generateAbility(permissions, user);
     },
 
+    /**
+     * Generate an ability based on an admin token's stored permissions, scoped to the owner.
+     * Token permissions are already validated and ceiling-clamped at write time.
+     */
+    async generateTokenAbility(tokenPermissions: Permission[], owner: AdminUser): Promise<Ability> {
+      return engine.generateAbility(tokenPermissions as any, owner);
```
nclsndr, line 84: to check

markkaylor:
Tested again using the right tokens 😆 

2.  
✅ Log in as a non-super-admin, navigate to /settings/admin-tokens → "Add new Admin Token" opens form with admin permission matrix; no token type selector visible. 
✅ Enable some permissions (only those the user holds are selectable) → save → reopen → same permissions pre-checked.
- Should the permissions not be the same as the role by default? 

3.
  ✅ As non-super-admin, try to enable a permission you don't hold → checkbox stays unchecked. 
  ✅ As super-admin editing another user's token → matrix is bounded to the owner's permission scope, not the super-admin's.

4. 
 ✅ As owner: GET /admin-tokens/:id returns accessKey; "View token" visible in UI. 
 ❓As super-admin viewing another user's token: no accessKey in response; Regenerate button hidden.     
 - All checks out for the UI but is it normal that I can use a super-admin jwt to retrieve the non-super-admin token with a GET to /admin-tokens/:id 
 ❌ POST /admin-tokens/:id/regenerate as super-admin on another user's token → 403.    
 - I’m able to regenerate with a POST request using the super-admin jwt
 
5.
 ✅ Remove a permission from a role → admin tokens owned by users of that role have the corresponding permission deleted automatically.    
- Adding the the permission back to the role syncs with the old permissions on the token, very nice UX!

6.  
 ✅ Deactivate a user who owns an admin token → using the token returns 401 Token owner is deactivated.    
 - I get “Missing or invalid credentials” but it is rejected    - Deleting the user also deletes the token, that’s nice UX

7. 
 ✅ /settings/api-tokens → only content-api tokens; no Kind/Owner column.
 ✅ /settings/admin-tokens → only admin tokens; Owner column present.
 
8. 
Did a few tests and seems to be working

Heads up in the UI there is this on hover:

Screenshot 2026-03-16 at 08 02 21

Seems to be working more or less how you expect. Should I start reviewing the code now?

nclsndr:
Thank you for your review @markkaylor! 

> Adding the the permission back to the role syncs with the old permissions on the token, very nice UX!

Do you mean you witnessed: 1. token has P permission. 2. P is removed from role. 3. P is removed from token perm. (side effect) 4. P is re-added to role. 5. Perm is re-added to token (side effect)? If this is the case, it is a bug. No state is supposed to be kept when removing permissions from role affecting tokens for the tokens to recover those perms if re-added later on.

> Should I start reviewing the code now?

I'll first make the CI pass and split up the scope in commits if anyone prefers smaller chunks for review

markkaylor:
BTW I'm unable to update a token, I think this is a bug on api tokens as well. I can only set the permissions on create. 

> Do you mean you witnessed: 1. token has P permission. 2. P is removed from role. 3. P is removed from token perm. (side effect) 4. P is re-added to role. 5. Perm is re-added to token (side effect)? If this is the case, it is a bug. No state is supposed to be kept when removing permissions from role affecting tokens for the tokens to recover those perms if re-added later on.

Yes that's what I'm seeing. See video for example.

Steps 1 to 3 is what is described here though:
Remove a permission from a role → admin tokens owned by users of that role have the corresponding permission deleted automatically. 
So is that the expected behavior? It's just re-adding the perm on the role should not sync with the perm on the token? 

Here's a video
https://github.com/user-attachments/assets/a8e7a08c-6c3f-4be6-869b-d94357184382

nclsndr:
Thanks for the details @markkaylor 👌 

> So is that the expected behavior?

No. We should witness the permissions being removed (not only disabled) from the token permissions whenever we remove them from the role. I'll have a look into it thanks a lot!

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
```diff
@@ -81,7 +81,12 @@ export const NavUser = ({
   return (
     <Flex {...props}>
       <Menu.Root>
-        <MenuTrigger endIcon={null} fullWidth justifyContent="flex-start">
+        <MenuTrigger
+          endIcon={null}
+          fullWidth
+          justifyContent="flex-start"
+          aria-label="Open user menu"
```
markkaylor: aria-label should be translated I believe and the e2e tests will need to be updated

packages/core/admin/admin/src/components/MainNav/NavUser.tsx
```diff
@@ -81,7 +81,15 @@ export const NavUser = ({
...
       <Menu.Root>
-        <MenuTrigger endIcon={null} fullWidth justifyContent="flex-start">
+        <MenuTrigger
+          endIcon={null}
+          fullWidth
+          justifyContent="flex-start"
+          aria-label={formatMessage({
+            id: 'app.components.LeftMenu.openUserMenuAriaLabel',
+            defaultMessage: 'Open user menu',
+          })}
```
nclsndr: Is that what you had in mind @markkaylor? 
related to [comment](https://github.com/strapi/strapi/pull/25657#discussion_r2960971442)

playwright.base.config.js
```diff
@@ -96,8 +96,8 @@ const createConfig = ({ port, testDir, appDir, reportFileName }) => ({
         }
       : 'off',
 
-    /* Use the storage state with localStorage set globally */
-    storageState: './tests/e2e/playwright-storage-state.json',
+    /* Use the storage state with localStorage set globally (port-specific for concurrent test-apps) */
+    storageState: `./tests/e2e/playwright-storage-state-${port}.json`,
```
nclsndr: Did you happen to have the same fix on your setup @innerdvations?

innerdvations: Yep, if it didn't already make it into develop you can do it here, I can't remember if I committed that or not

innerdvations (dismissed):

tests/app-template/src/api/config/utils/resync-super-admin-after-import.js
nclsndr, line 1: this is a hack we're adding to the test setup to counter the overrides of the instance database on test launch.

innerdvations (approved):

th3w0u1f:
@nclsndr I need help, please

I've upgraded to this latest version yesterday and since then I can change or edit API token permissions for content types. I get the error: adminUserOwner is required error. And I can't figure out how to fix it. 

I'm completely new to Strapi. My current setup is using a PostGres DB running locally on my machine, and with the local instance, I'm the only User and Super Admin. 

How can I fix this error for my local and Production instances?

dihnauer:
@th3w0u1f I'm facing the same problem here. :/

Anyone can help us?

nclsndr:
Thank you for reporting the issue @th3w0u1f & @dihnauer 
Could you open bug issues to report the details on what led to the regression on your end? 
Especially the versions you were running before and after the update.

dihnauer:
> Thank you for reporting the issue @th3w0u1f & @dihnauer Could you open bug issues to report the details on what led to the regression on your end? Especially the versions you were running before and after the update.

@nclsndr  I was using v5.44.0 before upgrade to 5.45.1.

Screenshot 2026-05-12 at 14 30 38

leandigital:
Same problem here. Figured out it's not a blocker as you can still create new API token with whatever permissions needed, but still - not being able to update one is not good.

dihnauer:
> Same problem here. Figured out it's not a blocker as you can still create new API token with whatever permissions needed, but still - not being able to update one is not good.

I rollback to v5.44.

nclsndr:
Fix on the way #26313

th3w0u1f:
@dihnauer @nclsndr  

A workaround I've found (staying on 5.45.1), is to refresh the strapi_api_tokens table in my database (I'm using Postgres Admin). After the refresh the 'kind' column was added with all of the fields set to NULL. I then ran the following query to update them to 'content-api' 

```
UPDATE strapi_api_tokens
SET kind = 'content-api'
WHERE kind IS NULL;
```

I'm not sure if this would impact the fix that's on the way, but it worked for me. I could change API token permissions again after doing this. 

@nclsndr I hope I didn't mess something up by doing this. And thanks for getting back to us so quickly.

nclsndr:
The workaround is totally legit and harmless @th3w0u1f. By doing so, you migrated the data to the state the new version is expecting, no extra side effect.

nclsndr:
Fix from #26313 has been shipped in [5.46](https://github.com/strapi/strapi/releases/tag/v5.46.0) release @th3w0u1f @dihnauer @leandigital

PR #25659, opened 2026-03-05
fix: skip entity lifecycles (beforeDelete/afterDelete) on republish and discard-draft

### LINKED ISSUE #25578: afterDelete lifecycle fires on relation updates on published documents
### Node Version

22.19.0

### NPM/Yarn/PNPM Version

11.8.0

### Strapi Version

5.37.1

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

We are in process of upgrading from v4 to v5. On v4, we were using lifecycle hooks to sync data with our search provider (Algolia). On upgrading to v5, we have noticed some issues.

```
// example schema
{
  "name" { "type": "string" },
  "products": {
      "type": "relation",
      "relation": "oneToMany",
      "target": "..."
    },
...
}
```

Given the model above, there are two issues I face:

1. when we update name `onDelete` is not fired. When we add/remove a product `onDelete` is fired. I would rather the latter not fire.
2. To overcome the above, within the `onDelete` hook I could call `strapi.documents(...).findOne({ documentId, status: 'published' })` to check if the record is indeed deleted. However, this also returns `null` because the published version has already been deleted but not yet recreated — making it impossible to distinguish this case from a genuine delete/unpublish.

Question: Is there a stable way to detect within a lifecycle hook whether afterDelete was triggered by an internal republish cycle vs. an explicit delete/unpublish? Thank you in advance for your help!

### Steps to Reproduce

  1. Have a published record with Draft & Publish enabled
  2. Add or remove a relation from the record in the admin
  3. Save and publish
  4. afterDelete fires

### Expected Behavior

`afterDelete` should only fire when the document is explicitly deleted or unpublished, not during internal republish cycles.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

Sahil8383:
### What does it do?

- Adds an internal `deleteReason` when the document service deletes entries for **republish** (publish flow) or **discard-draft** (discardDraft flow).
- In the entity manager `delete()`, when `deleteReason` is `republish` or `discardDraft`, **beforeDelete** and **afterDelete** lifecycles are skipped; the row is still deleted and relations cleaned up.
- Introduces a `DeleteReason` enum (`Republish`, `DiscardDraft`) and uses it in the document service and entity manager so call sites don’t rely on string literals.

### Why is it needed?

Fixes #25578: with Draft & Publish enabled, **afterDelete** (and **beforeDelete**) were firing when editing a published document and clicking “Save” then “Publish”. That flow deletes the old published row and creates a new one; that internal delete was triggering entity lifecycles, so hooks couldn’t tell “republish” from a real delete/unpublish. Lifecycles should only run for explicit delete/unpublish, not for republish or discard-draft.

### How to test it?

1. Build: `yarn build` (or build `@strapi/database`, `@strapi/core`, `@strapi/strapi`).
2. Run the getstarted example: `cd examples/getstarted && yarn develop --watch-admin`.
3. Use a content type with **draftAndPublish: true** (e.g. Restaurant) and add a simple `afterDelete` log in its lifecycles.
4. **Republish:** Create → Publish → edit (e.g. name) → Save → Publish. **Expected:** no `afterDelete` log.
5. **Unpublish / Delete:** Unpublish or delete an entry. **Expected:** `afterDelete` runs and the log appears.

### Related issue(s)/PR(s)

Fixes #25578

### DIFF
--- a/packages/core/core/src/services/document-service/entries.ts
+++ b/packages/core/core/src/services/document-service/entries.ts
@@ -1,4 +1,5 @@
 import type { UID, Modules } from '@strapi/types';
+import type { DeleteReason } from '@strapi/database';
 import { async, errors } from '@strapi/utils';
 import { assoc, omit } from 'lodash/fp';
 
@@ -87,10 +88,10 @@ const createEntriesService = (
     return doc;
   }
 
-  async function deleteEntry(id: number) {
+  async function deleteEntry(id: number, deleteParams = {} as { deleteReason?: DeleteReason }) {
     const componentsToDelete = await components.getComponents(uid, { id });
 
-    const deletedEntry = await strapi.db.query(uid).delete({ where: { id } });
+    const deletedEntry = await strapi.db.query(uid).delete({ where: { id }, ...deleteParams });
 
     await components.deleteComponents(uid, componentsToDelete as any, { loadComponents: false });
 
--- a/packages/core/core/src/services/document-service/repository.ts
+++ b/packages/core/core/src/services/document-service/repository.ts
@@ -9,6 +9,7 @@ import {
 } from '@strapi/utils';
 
 import type { UID, Modules } from '@strapi/types';
+import { DeleteReason } from '@strapi/database';
 import { wrapInTransaction, type RepositoryFactoryMethod } from './common';
 import * as DP from './draft-and-publish';
 import * as i18n from './internationalization';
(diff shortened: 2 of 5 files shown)


innerdvations:
Hello! Thanks for the PR, but we recommend using document service middleware if you need higher-level logic since the database lifecycle hooks are intended specifically for lower level database activity.

PR #25660, opened 2026-03-06
fix(admin): resolve prism is not defined

### LINKED ISSUE #25070: ReferenceError: Prism is not defined - Admin panel blank page in Strapi 5.33

- Node.js version: v22.18.0
- NPM version: 10.8.2 (using pnpm 10.19.0)
- Strapi version: 5.33.0
- Database: PostgreSQL
- Operating system: Windows 10

### Describe the bug

The Strapi admin panel shows a completely blank page with a `ReferenceError: Prism is not defined` error in the browser console. This prevents access to the entire admin interface.

**Root Cause Analysis:**

The issue is in `@strapi/content-manager` package. In the file:
```
node_modules/@strapi/content-manager/dist/admin/pages/EditView/components/FormInputs/BlocksInput/Blocks/Code.mjs
```

The code uses `Prism.languages` and `Prism.tokenize` as global variables without properly importing the Prism module:

```javascript
// Line ~4528 in the Vite bundled chunk
const selectedLanguage = Prism.languages[decorateKey || "plaintext"];
const tokens = Prism.tokenize(text, selectedLanguage);
```

While `prismjs` is listed as a dependency of `@strapi/content-manager`, the module expects `Prism` to be available as a global window object, but it's never attached to window during the Vite bundling process.

1. Create a fresh Strapi 5.33 project:
   ```bash
   npx create-strapi-app@5.33.0 my-project --typescript
   cd my-project
   ```

2. Start the development server:
   ```bash
   npm run develop
   ```

3. Open `http://localhost:1337/admin` in Chrome/Edge

4. Open browser DevTools (F12) and check the Console tab

5. **Expected:** Login page should be visible
   **Actual:** Blank white page with error:
   ```
   ReferenceError: Prism is not defined
       at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-XXXXXX.js
   ```

### Screenshots

The admin panel shows a completely blank page. Browser console shows:
```
ReferenceError: Prism is not defined
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-DT44COQD.js:4528
```

### Workaround

Until this is fixed, users can work around the issue by:

1. Add `prismjs` as a direct dependency:
   ```bash
   npm install prismjs
   npm install -D @types/prismjs
   ```

2. Modify `src/admin/app.tsx` to import Prism first and attach to window:
   ```typescript
   // This MUST be the first import
   import Prism from 'prismjs';
   if (typeof window !== 'undefined') {
     (window as any).Prism = Prism;
   }

   // Then other imports
   import type { StrapiApp } from '@strapi/strapi/admin';
   // ... rest of file
   ```

3. Rebuild and restart:
   ```bash
   npm run build
   npm run develop
   ```

### Suggested Fix

In `@strapi/content-manager`, the Code.mjs file should properly import Prism and either:
1. Use the imported module directly instead of relying on global
2. Or explicitly attach Prism to window after import

Example fix in the content-manager source:
```javascript
import Prism from 'prismjs';

// Ensure Prism is available globally for any dependent code
if (typeof window !== 'undefined') {
  window.Prism = Prism;
}

// Then use Prism.languages, Prism.tokenize, etc.
```

### Additional context

- This appears to be a Vite bundling issue where Prism's UMD/CommonJS export isn't being properly handled
- The issue may be related to how Vite optimizes dependencies and the fact that `prismjs` expects to attach itself to `window` in browser environments
- The CSS theme file IS being imported (`prism-solarizedlight.css`), but the JS module isn't properly exposing the global

### Code snippets

Problematic code in the bundled chunk:
```javascript
// node_modules/@strapi/content-manager/.../Code.mjs
import "...prismjs/themes/prism-solarizedlight.css";

var decorateCode = ([node, path]) => {
  // ...
  const selectedLanguage = Prism.languages[decorateKey || "plaintext"];  // <-- Prism is undefined
  const tokens = Prism.tokenize(text, selectedLanguage);  // <-- Prism is undefined
  // ...
};
```

Dependency tree showing prismjs comes from content-manager:
```
@strapi/content-manager 5.33.0
└── prismjs 1.30.0
```

innerdvations:
### What does it do?

- Content-manager Code block (Code.tsx): Uses a single Prism reference that prefers the imported prismjs module when it has .languages, otherwise uses window.Prism (set by the content-manager preload). Adds a guard at the start of decorateCode: if Prism or Prism.languages is missing, it returns empty ranges instead of calling Prism.languages / Prism.tokenize, so the admin no longer throws.
- Vite config: Adds prismjs to the base optimizeDeps.include (for all apps, not only the example app) so Vite pre-bundles it and the ESM namespace is available in every chunk that imports it.

### Why is it needed?

In some setups (e.g. Node 22 + pnpm, or after clearing the Vite cache), the admin can show a blank page with ReferenceError: Prism is not defined in the console. prismjs is UMD and doesn’t always expose a proper ESM namespace when bundled; the Code block lives in a lazy-loaded chunk that may run in a context where import * as Prism from 'prismjs' is empty or not yet applied. The content-manager already preloads prismjs so window.Prism is set, but the Code block only used the import, so when that import was broken it crashed. The fix makes the Code block use the global when the import isn’t usable, adds a safe guard so missing Prism only skips highlighting, and pre-bundles prismjs so the import path works reliably.

### How to test it?

force vite to fully optimize dependencies and reproduce the issue in #25070 

with this PR, it should start working

Experimental: `0.0.0-experimental.fbef14bacc7d5b9324c7f053944d735fa12a55b9`

### Related issue(s)/PR(s)

Fixes #25070


innerdvations:
@dosu You must never add a 'source:' or 'pr:' label if one is already there. This repo requires exactly one source and one pr label, adding more will interfere with the workflows.

jhoward1994:
It seems to be working for me. Is it the right recreation to use:

```
Experimental: 0.0.0-experimental.fbef14bacc7d5b9324c7f053944d735fa12a55b9
FORCE_OPTIMIZE_DEPS=true yarn run develop
```

innerdvations:
> It seems to be working for me. Is it the right recreation to use:
> 
> ```
> Experimental: 0.0.0-experimental.fbef14bacc7d5b9324c7f053944d735fa12a55b9
> FORCE_OPTIMIZE_DEPS=true yarn run develop
> ```

I think so, try that without this PR and see if you get the error to be sure

jhoward1994:
I actually cant recreate the issue before or after these changes. Can you? Maybe it would be good to get one of the developers on the issue to try the experimental ?

innerdvations:
I'm trying to get someone from the reported issue to notice 😂 I just need confirmation from someone this happens to that it fixes the problem

Adzouz:
Hey @innerdvations and @jhoward1994, I wasn't able to reproduce the original issue either so I don't really know how I can test this fix tbh 😅 do you have any additional info that would help? (I tried the exact same configuration as mentioned in the issue, the only thing is that I'm using a Mac instead of Windows)

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/BlocksInput/Blocks/Code.tsx
```diff
@@ -65,11 +65,24 @@ import 'prismjs/components/prism-tsx';
 import 'prismjs/components/prism-vbnet';
 import 'prismjs/components/prism-yaml';
 
+// prismjs is UMD and may not expose a namespace when bundled by Vite; the content-manager
+// index preloads it so 'window.Prism' is set. Use that when the module import is empty.
+const Prism =
```
Adzouz: I'm not really fan of nested ternary condition, do you think we can change it to make it clearer?

Adzouz (commented):
Trying to reproduce it in real life is pretty hard so I trust the test if it's passing. At least it would help closing the issue and if someone still has a problem, they can still reopen it later 😊
One tiny comment and just need to fix the typescript issue and you're good to go! 🙏

Adzouz (approved):
Looks good to me, thanks @innerdvations 🙏

PR #25664, opened 2026-03-06
fix: add maxwidth to documentactions menu

mathildeleg:
### What does it do?
Adding a max-width to the DocumentActions' Menu.

### Why is it needed?
Without it, that larger elements inside (name, locale's name) would make the menu very large (especially on mobile).
**BEFORE**
Capture d’écran 2026-03-06 à 14 56 07

**AFTER**
Capture d’écran 2026-03-06 à 14 55 53

### How to test it?
* Go to the Edit view of a content type.
* Click on the "More" actions button on the top right corner.
* Check with a locale with a long name.
* Go to your personal settings and update your name to be longer.
* Go back to the Edit view and check with that longer name.

Fixes CMS-285

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -409,7 +409,7 @@ const DocumentActionsMenu = ({
         icon={<More />}
         variant={variant}
       />
-      <Menu.Content maxHeight={undefined} popoverPlacement="bottom-end">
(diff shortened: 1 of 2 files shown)


mathildeleg:
> Nice, tested and works well thanks 🙏 Also if you could add either a word break or an overflow ellipsis in the case where the name is very long and without spaces? 🙏 (yeah I know it's unlikely but we never know 😂) Screenshot 2026-03-11 at 10 30 44
> 
> You can do that in `packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx` (line 407)

All fixed 🤗
Capture d’écran 2026-03-11 à 11 08 37

PR #25668, opened 2026-03-06
feat(core/ai): expose strapi.ai namespace

nclsndr:
### What does it do?                                                                                                                                       
                                                                                                                                                       
Introduces a strapi.ai namespace that groups all AI-related services under a single accessor on the Strapi instance.                                   
                                                                                                                                                       
- Replaces strapi.mcp (McpService) with strapi.ai (AiNamespace) — the MCP service is now accessible at strapi.ai.mcp                                   
- Adds strapi.ai.admin (AiAdminService | undefined) — lazily resolved from strapi.get('ai.admin'), only present in EE with the cms-ai feature enabled  
- Introduces Modules.AI.AiNamespace and Modules.AI.AiAdminService types in @strapi/types                                                               
- Renames the providers/mcp.ts provider to providers/ai.ts, which registers 'ai' in the container                                                      
- Adds core/services/ai.ts — createAiNamespace factory wiring mcp and the optional admin getter
- Updates all call sites: strapi.get('ai').getAiToken() → strapi.get('ai.admin').getAiToken() in i18n and upload plugins
- Removes the stale ai?: { enabled: boolean } field from Core.Strapi (was server-side, never set)

### Why is it needed?

The previous shape was flat and inconsistent: strapi.mcp exposed MCP directly, while the EE ai container held admin token logic under the same 'ai' key
 with no clear grouping. As AI capabilities grow (MCP + admin features + future additions), they need a unified namespace. This refactor establishes
strapi.ai as the single entry point for all AI services, with a clear CE/EE split (mcp always present, admin only on EE).
How to test it?
1. MCP still works — start Strapi in dev mode with server.mcp.enabled: true; verify the MCP server starts ([MCP] Starting MCP server... log) and
strapi.ai.mcp is accessible.
2. AI token endpoint (EE) — on an EE instance with cms-ai enabled, call GET /admin/ai/token; verify it returns a token (controller now calls
strapi.get('ai.admin').getAiToken()).
3. i18n AI localizations (EE) — trigger an AI localization job; verify token retrieval does not throw TypeError: strapi.get(...).getAiToken is not a
function.
4. Upload AI metadata (EE) — trigger AI metadata generation; same verification.
5. Unit tests — run:
  - `packages/core/admin/ee/server/src/ai/containers/__tests__/ai.test.ts`
  - `packages/core/admin/ee/server/src/ai/controllers/__tests__/ai.test.ts`

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,5 +7,5 @@ preferInteractive: true
 yarnPath: .yarn/releases/yarn-4.12.0.cjs
 
 catalog:
-  "@modelcontextprotocol/sdk": ^1.27.1
+  '@modelcontextprotocol/sdk': ^1.27.1
   vitest: ^4.0.18
--- a/packages/core/admin/admin/src/render.ts
+++ b/packages/core/admin/admin/src/render.ts
@@ -59,7 +59,9 @@ const renderAdmin = async (
     },
     ai: {
       enabled: true,
-    },
+      // Global pollution from the Node.js scope.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any,
   };
 
   const { get } = getFetchClient();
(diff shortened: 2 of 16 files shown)


innerdvations:
Is anyone aware if we have publicly documented the existing `'ai'` anywhere that would prevent us from breaking this?

nclsndr:
> Is anyone aware if we have publicly documented the existing `'ai'` anywhere that would prevent us from breaking this?

I believe this is internals only so far. Additionally, we could leave the namespace having the same config like behavior while adding more to it.

nclsndr:
We want to introduce the ai namespace prior to mcp

PR #25671, opened 2026-03-06
enhancement: develop restart-on-error

innerdvations:
### What does it do?

Adds an optional `--restart-on-error` flag to `strapi develop`.

When enabled, the primary process:

- Restarts the worker after unexpected process exits (with sliding-window backoff)
- Coordinates reload forking via worker exit events (avoids double-fork on file reload)
- Optionally auto-retries startup after failures (5s delay + existing file watcher)

Without the flag, behavior matches current `develop`, including [#25514](https://github.com/strapi/strapi/pull/25514) (continue watching on TypeScript/build errors).

This PR is rebased onto `develop` and layers restart-on-error on top of that flow — it does **not** change default startup order or error handling.

### Why is it needed?

During local development, transient runtime errors currently kill the worker and leave `strapi develop` in a dead state. Users shouldn't need external CLI wrappers to recover.

This flag is opt-in today; Strapi 6 may flip it to default.

**Note:** `Strapi.stop()` still sends `'stop'` and exits the whole develop session — that is intentional (controlled shutdown vs crash recovery).

### How to test it?

**Default (no flag) — should match `develop` / #25514:**

1. In a TypeScript project, run `strapi develop`
2. Introduce a TS syntax error → CLI logs the error, keeps watching, recovers on fix
3. Normal file-save reload still works

**With `--restart-on-error`:**

1. Run `strapi develop --restart-on-error`
2. Cause an uncaught runtime error after the server starts → worker restarts automatically
3. Cause repeated crashes → backoff delay increases (caps at 5s)
4. Startup failure → auto-retry after 5s and/or on next file change
5. File-save reload → single worker replacement (no double-fork)

**Unchanged paths:**

- `strapi build` and `strapi start` compile behavior unchanged
- No flag → no change to production or CI usage

### Related issue(s)/PR(s)

Inspired by [#25514](https://github.com/strapi/strapi/pull/25514), which handles TypeScript/build errors. This extends recovery to unexpected worker crashes while keeping that behavior as the default.


innerdvations:
Closing this since it's not a priority; will come back to this later.

PR #25673, opened 2026-03-07
enhancement: deduplicate pkg.ts validation logic between cloud-cli and strapi

### LINKED ISSUE #25536: refactor: deduplicate pkg.ts validation logic between @strapi/cloud-cli and @strapi/strapi
## Context

`packages/cli/cloud/src/utils/pkg.ts` and `packages/core/strapi/src/cli/utils/pkg.ts` are near-identical copies of the same file. They share the same yup schema, validation logic, and error handling, with only trivial differences:

- `fs-extra` vs `fs/promises` for reading the file
- Different logger import paths
- Minor style differences (`return await` vs intermediate variable)

This was spotted while fixing a shared typo (`compatability` -> `compatibility` in #25535) — the same misspelling existed in both files because one was copied from the other.

## Proposal

Extract the shared `package.json` validation logic (schema + `validatePkg` + `loadPkg`) into a common location (e.g. `@strapi/utils` or a new shared CLI utils package) so both `@strapi/cloud-cli` and `@strapi/strapi` can import from a single source.

## Files

- `packages/cli/cloud/src/utils/pkg.ts`
- `packages/core/strapi/src/cli/utils/pkg.ts`

Akash504-ai:
## What does it do?

This PR removes duplicated `package.json` validation logic that existed in two different locations:

- `packages/cli/cloud/src/utils/pkg.ts`
- `packages/core/strapi/src/cli/utils/pkg.ts`

The shared validation logic (Yup schema + `validatePkg`) has been extracted into a new shared utility:
`packages/core/utils/src/pkg-utils.ts`

Both `cloud-cli` and `strapi` now import the validation logic from `@strapi/utils`, ensuring a single source of truth.

---

## Why is it needed?

Previously, the validation logic was duplicated across two files. This caused:

- Code duplication
- Risk of inconsistencies if one implementation changed
- Maintenance overhead

By extracting the shared logic into `@strapi/utils`, both packages now rely on a single implementation, improving maintainability and consistency.

---

## How to test it?

1. Build the repository:

```bash
yarn build
```
2.Ensure the following packages compile successfully:

`@strapi/cloud-cli`

`@strapi/strapi`

3.Verify that both `pkg.ts` files now import the shared validation logic from `@strapi/utils`.

### Related issue(s)/PR(s)

Fixes #25536

### DIFF
--- a/packages/cli/cloud/src/utils/pkg.ts
+++ b/packages/cli/cloud/src/utils/pkg.ts
@@ -1,60 +1,12 @@
 import * as fse from 'fs-extra';
 import os from 'os';
 import pkgUp from 'pkg-up';
-import * as yup from 'yup';
-import chalk from 'chalk';
+import { validatePkg, PackageJson } from '@strapi/utils';
 import { Logger } from '../services/logger';
-
-interface Export {
-  types?: string;
-  source: string;
-  module?: string;
-  import?: string;
-  require?: string;
-  default: string;
-}
-
-const packageJsonSchema = yup.object({
-  name: yup.string().required(),
-  exports: yup.lazy((value) =>
-    yup
-      .object(
-        typeof value === 'object'
-          ? Object.entries(value).reduce(
-              (acc, [key, value]) => {
-                if (typeof value === 'object') {
-                  acc[key] = yup
-                    .object({
-                      types: yup.string().optional(),
-                      source: yup.string().required(),
-                      module: yup.string().optional(),
-                      import: yup.string().required(),
-                      require: yup.string().required(),
-                      default: yup.string().required(),
-                    })
-                    .noUnknown(true);
-                } else {
-                  acc[key] = yup
-                    .string()
-                    .matches(/^\.\/.*\.json$/)
-                    .required();
-                }
-
-                return acc;
-              },
-              {} as Record<string, yup.SchemaOf<string> | yup.SchemaOf<Export>>
-            )
-          : undefined
-      )
-      .optional()
-  ),
-});
-
-type PackageJson = yup.Asserts<typeof packageJsonSchema>;
 
 /**
  * @description being a task to load the package.json starting from the current working directory
- * using a shallow find for the package.json  and `fs` to read the file. If no package.json is found,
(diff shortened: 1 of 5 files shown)


packages/core/strapi/tsconfig.json
```diff
@@ -6,6 +6,5 @@
     "module": "ESNext",
     "moduleResolution": "Bundler"
   },
-  "include": ["src"],
```
nclsndr, line 9: Why do we need this deletion

packages/cli/cloud/src/utils/pkg.ts
```diff
@@ -65,64 +17,12 @@ const loadPkg = async ({ cwd, logger }: { cwd: string; logger: Logger }): Promis
...
-          }
-      }
-    }
-
-    throw err;
-  }
-};
-
-export type { PackageJson, Export };
+export type { PackageJson };
```
nclsndr, line 27: We probably don't want re-exporting types from another package locally

nclsndr (commented):
Thank you for raising this PR @Akash504-ai 
Few comments before actually reviewing with QA

packages/core/strapi/tsconfig.json
Akash504-ai: Initially while extracting the shared validation logic, I added a reference in 'packages/core/strapi/tsconfig.json' so TypeScript could resolve the file during development.
After moving the logic to packages/core/utils/src/pkg-utils.ts and exposing it via @strapi/utils, the strapi package no longer needed to reference that file directly. The extra tsconfig reference was therefore removed to keep the configuration consistent with other internal package imports.

packages/cli/cloud/src/utils/pkg.ts
Akash504-ai: You r right, PackageJson is defined and exported from @strapi/utils (pkg-utils.ts). Re-exporting it locally from the pkg.ts files would unnecessarily duplicate the type exposure across packages.
The intention of the refactor was for both cloud-cli and strapi to consume the validation logic and related types directly from @strapi/utils, so I haveve removed the local type re-export to avoid that duplication.

PR #25683, opened 2026-03-09
chore(content-manager): optimize relations handling in EditView component

jhoward1994:
### What does it do?

 Fixes the `selectFromResult` callback in `useGetRelationsQuery` to stop creating a new `data` object on every invocation. 
 The old code spread `data: { ...result.data, results: ... }`, which produced a new object reference every time — Now `results` and `pagination` are returned as flat top-level keys.

 Also wraps handlers and components in `useCallback` / `React.memo` to prevent excessive re-renders.

 ### Why is it needed?

 When editing an entity with multiple relation fields, scroll becomes laggy/jittery until all relation data loads.
 
 ### How to test it?

 1. `cd examples/getstarted && yarn develop`
 2. Open an entity with multiple relation fields in a modal
 3. Scroll while relations are loading — should be smooth, no jitter
 4. Connect/disconnect a relation — only the affected field should re-render
 5. Run tests: `npx nx run @strapi/content-manager:"test:front"`

 ### Related issue(s)/PR(s)

 Fix CMS-309


jhoward1994:
https://gist.github.com/jhoward1994/c0ac200a3bac73ad54dccdaadc4bcbe0

This diff seeds examples/getstarted with data to recreate the bug

jhoward1994:
@Bassel17 

After merging develop in here I think we've tackled some of the same things here and https://github.com/strapi/strapi/pull/25623

I've mainly taken the changes from develop where we had duplicated fixes/performance improvements. I think the changes were equivalent and it keeps things cleaner.

So basically the diff here is now much smaller but I think this still contains some worthwhile changes. wdyt?

PR #24196, opened 2025-08-18
fix: formatErrorMessages array values formatting

hood:
Fix #23606 
Fix #18810 

### What does it do?

Fix the formatting of error messages in `PublishAction` when errors are a nested array.

### Why is it needed?

Formatting of errors is broken for nested arrays of errors.

### How to test it?

Try bulk-publishing entries containing dynamic zones which have content with validation errors.

### Related issue(s)/PR(s)

#23606, #18810

PR #25687, opened 2026-03-09
fix: bulk publish validation on required components in dz

### LINKED ISSUE #18810: Bulk publish: media field is required

- Node.js version:  v18.17.0 win32 x64
- NPM version: npm v10.0.0 / yarn v1.22.19 <-- I use yarn
- Strapi version: 4.15.5-alpha.4
- Database: postgresql
- Operating system: Windows x64

### Describe the bug

When trying to bulk publish a content entry i receive an error.
It works completely fine just pressing Publish directly on the entry, but bulk publish does not work.

It only happens when a media property is set to required and nested within a component.

1. Go to "Content Type Builder"
2. Create new component "Thumbnail" add property "image" as single media `required`
3. Create new collection type "Test" add property "component" and choose "Thumbnail" and set `required`
4. Go to "Content Manager" select "Test" and create new entry.
5. Set image in thumbnail component and save
6. Click on "Select All" and click on "Publish"
7. See error

### Expected behavior

Since the property Thumbnail.image is set, the publish function should not report an error.
This only happens when the media property is required and nested within a component.

### Screenshots

Component: 

Collection Type:

Content Manager:

*Error:*

### LINKED ISSUE #23606: Bulk Publishing shows entries in  "Waiting for action" for entries that can be published individually
### Node Version

20.19.2

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.13.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

I have a couple of Collections where I can publish entries individually, but when I use bulk publishing, the dialog show a number of entries as  "Waiting for action" with status "[object Object].withField", which prevents me from publishing them. When I open them individually to publish them, publishing works. See attached file

https://github.com/user-attachments/assets/156b5e15-fe44-4d51-a984-a45a03f3c9f5

### Steps to Reproduce

- Create a collection with nested components and/or floating zones that contain mandatory fields or fields with validations.
- Fill all mandatory fields
- Use bulk publish
- entries are in "Waiting for action", even if everything is valid

### Expected Behavior

- all entries are published

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

### LINKED ISSUE #24576: Bulk publish inside Content manager not working as expected when nested components inside dynamic zones are involved
### Node Version

22.17.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.24.1

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When having a required component, nested inside a component which is nested inside a dynamic zone, the publish bulk button inside the content manager will wrongly validate the entry as being not ready to publish, when in fact it is.

### Steps to Reproduce

1. Create a content type
2. Add a dynamic zone to the content type
3. Add a component to the dynamic zone
4. Inside the component, add another REQUIRED component
5. Create an entry, and add the required component to the entry
6. Try to publish the entry using the bulk publish option inside the content manager list

### Expected Behavior

The bulk publish should correctly validate the required component.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

Image

Image

Image

### Additional information

_No response_

### Confirmation Checklist

Adzouz:
### What does it do?

1. Fixes a validation issue making the bulk publish impossible on documents that have required components in dynamic zones.
If the component is required but was correctly provided, the publish action should be possible (and indeed, publishing directly from the edit view page is working) but when selecting the item for a bulk publish, a validation error says that the field is required (as if it was empty even if it's not the case).

Issue:

https://github.com/user-attachments/assets/2af42ae1-6bac-4911-9c12-07ed540c12fe

After fix:

https://github.com/user-attachments/assets/0d995f15-bd9a-4f25-9f0c-cbe0061e4cd7

2. Also fixes an issue targetting the publish button in the document actions based on its label but if using another locale on the platform, this check isn't working.

Issue:

https://github.com/user-attachments/assets/9c789205-be03-42c2-a6c0-ded522865e08

_ℹ️ There's another issue wrongly displaying the errors related to dynamic zones and components but it's part of this community PR: https://github.com/strapi/strapi/pull/24196_

### Why is it needed?

To unblock the bulk publish on documents that can really be published.

### How to test it?

1. Bulk publish issue
- Create or use a collection type with a dynamic zone that contains a required component (repeatable or not)
- Create a draft an fill all the required information
- Click Save
- Go back to the list view page and select the newly created document
- Select Publish action to open bulk publish modal
-> You shouldn't have any error displayed and should be able to publish the document

2. Publish type button
- Go to the profile settings
- Change the language of the interface to anything else than English
- Go to any collection or single type that has draft and publish
- Open any document
- Inspect the "Publish" button translated to show the console
-> The type of the button tag should be button and not submit (Save button should be the default submit of the form)

### Related issue(s)/PR(s)

Fixes #23606 
Fixes #18810
Fixes #24576

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -59,7 +59,7 @@ type DocumentActionPosition = 'panel' | 'header' | 'table-row' | 'preview' | 're
 
 interface DocumentActionDescription {
   label: string;
-  type?: string;
+  type?: 'publish' | 'update' | 'unpublish' | 'discard';
   onClick?: (event: React.SyntheticEvent) => Promise<boolean | void> | boolean | void;
   icon?: React.ReactNode;
   /**
@@ -210,7 +210,7 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
           <DocumentActionButton
             {...primaryAction}
             variant={primaryAction.variant || 'default'}
-            {...(primaryAction.type === 'publish' ? {} : { buttonType: 'submit' })}
+            buttonType={primaryAction.type === 'publish' ? undefined : 'submit'}
           />
         )}
       </Flex>
@@ -258,15 +258,16 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
  * DocumentActionButton
  * -----------------------------------------------------------------------------------------------*/
 
-interface DocumentActionButtonProps extends Action {
+interface DocumentActionButtonProps extends Omit<Action, 'type'> {
   buttonType?: 'button' | 'submit' | 'reset';
+  type?: string;
 }
 
 const DocumentActionButton = ({ buttonType = 'button', ...action }: DocumentActionButtonProps) => {
   const [dialogId, setDialogId] = React.useState<string | null>(null);
   const { toggleNotification } = useNotification();
 
-  const handleClick = (action: Action) => async (e: React.MouseEvent) => {
+  const handleClick = (action: DocumentActionButtonProps) => async (e: React.MouseEvent) => {
     const { onClick = () => false, dialog, id } = action;
 
     const muteDialog = await onClick(e);
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -260,7 +260,7 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
 
 interface DocumentActionButtonProps extends Omit<Action, 'type'> {
   buttonType?: 'button' | 'submit' | 'reset';
-  type?: string;
+  type?: DocumentActionDescription['type'];
 }
 
 const DocumentActionButton = ({ buttonType = 'button', ...action }: DocumentActionButtonProps) => {
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -59,6 +59,7 @@ type DocumentActionPosition = 'panel' | 'header' | 'table-row' | 'preview' | 're
 
 interface DocumentActionDescription {
   label: string;
+  type?: string;
   onClick?: (event: React.SyntheticEvent) => Promise<boolean | void> | boolean | void;
   icon?: React.ReactNode;
   /**
@@ -204,22 +205,14 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
   const primaryActionContent = (
     <>
       <Flex flex={1} alignItems="stretch" direction="column">
-        {primaryAction.label === 'Publish'
-          ? addHintTooltip(
-              primaryAction,
-              <DocumentActionButton
-                {...primaryAction}
-                variant={primaryAction.variant || 'default'}
-              />
-            )
-          : addHintTooltip(
-              primaryAction,
-              <DocumentActionButton
-                {...primaryAction}
-                variant={primaryAction.variant || 'default'}
-                buttonType="submit"
-              />
-            )}
+        {addHintTooltip(
+          primaryAction,
+          <DocumentActionButton
+            {...primaryAction}
+            variant={primaryAction.variant || 'default'}
+            {...(primaryAction.type === 'publish' ? {} : { buttonType: 'submit' })}
+          />
+        )}
       </Flex>
 
       {restActions.length > 0 ? (
@@ -241,7 +234,7 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
       </tours.contentManager.Publish>
       {secondaryAction ? (
         <Flex flex={1} order={{ initial: -1, large: 0 }} alignItems="stretch" direction="column">
-          {secondaryAction.label === 'Publish' ? (
+          {secondaryAction.type === 'publish' ? (
             <tours.contentManager.Publish>
               <DocumentActionButton
                 {...secondaryAction}
@@ -622,7 +615,6 @@ const PublishAction: DocumentActionComponent = ({
   const validate = useForm('PublishAction', (state) => state.validate);
   const setErrors = useForm('PublishAction', (state) => state.setErrors);
   const formValues = useForm('PublishAction', ({ values }) => values);
-  const initialValues = useForm('PublishAction', ({ initialValues }) => initialValues);
   const resetForm = useForm('PublishAction', ({ resetForm }) => resetForm);
   const {
     currentDocument: { components },
@@ -953,6 +945,7 @@ const PublishAction: DocumentActionComponent = ({
   }
 
   return {
+    type: 'publish',
     loading: isLoading,
     position: ['panel', 'preview', 'relation-modal'],
     /**
--- a/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
@@ -34,13 +34,14 @@ import { useDocumentActions } from '../../../../hooks/useDocumentActions';
 import { useDocLayout } from '../../../../hooks/useDocumentLayout';
 import { contentManagerApi } from '../../../../services/api';
 import {
-  useGetAllDocumentsQuery,
+  useGetDocumentsForValidationQuery,
   usePublishManyDocumentsMutation,
 } from '../../../../services/documents';
 import { buildValidParams } from '../../../../utils/api';
 import { getTranslation } from '../../../../utils/translations';
 import { createYupSchema } from '../../../../utils/validation';
 import { DocumentStatus } from '../../../EditView/components/DocumentStatus';
+import { transformDocument } from '../../../EditView/utils/data';
 
 import { ConfirmDialogPublishAll, ConfirmDialogPublishAllProps } from './ConfirmBulkActionDialog';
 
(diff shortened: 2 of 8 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -622,7 +615,6 @@ const PublishAction: DocumentActionComponent = ({
   const validate = useForm('PublishAction', (state) => state.validate);
   const setErrors = useForm('PublishAction', (state) => state.setErrors);
   const formValues = useForm('PublishAction', ({ values }) => values);
-  const initialValues = useForm('PublishAction', ({ initialValues }) => initialValues);
```
Adzouz, line 625: This wasn't used

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -953,6 +945,7 @@ const PublishAction: DocumentActionComponent = ({
   }
 
   return {
+    type: 'publish',
```
markkaylor: But are all the actions going to have a `type` now?

mathildeleg (commented):
On issue 1: 
Works well but I'm kind of worried of the performance impact this might have on the bulk publish modal if wanting to publish a lot of documents at once...
But this does resolve the issue, so I understand why you created it (`getDocumentsForValidation`).
The solution could be to create a batch endpoint that would accept all document IDs, run only one `findMany` (with the same populate as `findOne`), and return the same document shape in the response so the modal gets all documents in a single round-trip, which reduces the performance cost.
But maybe let's ask @markkaylor or @jhoward1994 what they think?

On issue 2: 
LGTM ✅

Adzouz:
@mathildeleg  Thanks for the review and very useful feedback, you're totally right 🙏
I tried to optimize the request and instead of doing N requests of findOne, I'm doing now only one find request with a filtering on given documentIds (which I think is more perf). What do you think?

packages/core/content-manager/admin/src/services/documents.ts
```diff
@@ -256,6 +256,49 @@ const documentApi = contentManagerApi.injectEndpoints({
...
+    >({
+      query: ({ collectionType, model, documentIds, params }) => {
+        const filters = documentIds.length > 0 ? { documentId: { $in: documentIds } } : undefined;
+        const queryParams = {
+          ...params,
+          filters,
+          status: 'draft',
+          _populate: 'deep',
+          page: 1,
+          pageSize: Math.max(documentIds.length, 1),
```
mathildeleg: `document-manager.findPage` uses _maxLimit: 1000_. If someone selects more than 1000 entries, the API only returns 1000 but the UI will think it has all of them, so validation and “entries to publish” can be wrong. I’d cap at 1000 (e.g. `Math.min(documentIds.length, 1000)`) and document that bulk publish is limited to 1000, or add a smaller limit (e.g. 100) and document it, since deep populate on many docs is heavy.

packages/core/content-manager/admin/src/services/documents.ts
```diff
@@ -256,6 +256,49 @@ const documentApi = contentManagerApi.injectEndpoints({
...
+      FindOne.Response['data'][],
+      {
+        collectionType: string;
+        model: string;
+        documentIds: string[];
+        params?: FindOne.Request['query'];
+      }
+    >({
+      query: ({ collectionType, model, documentIds, params }) => {
+        const filters = documentIds.length > 0 ? { documentId: { $in: documentIds } } : undefined;
```
mathildeleg: Right now when `documentIds.length === 0` you set filters `= undefined`. So the request still goes out with `status: 'draft', page: 1, pageSize: 1` and no documentId filter. The API then returns the first draft it finds (one arbitrary document), not an empty list.
Suggestion:

```suggestion
        const filters = documentIds.length > 0
  ? { documentId: { $in: documentIds } }
  : { documentId: { $in: [] } };
```

packages/core/content-manager/admin/src/services/documents.ts
```diff
@@ -256,6 +256,49 @@ const documentApi = contentManagerApi.injectEndpoints({
         },
       }),
     }),
+    /**
+     * Fetches multiple documents with full populate via the existing find API.
+     * Uses filters[documentId][$in] and _populate=deep for validation.
+     */
+    getDocumentsForValidation: builder.query<
+      FindOne.Response['data'][],
```
mathildeleg: Return type is `FindOne.Response['data'][]` but you’re really returning the list’s results. I’d align the type with Find.Response (e.g. `Find.Response['results']` or `Document[]`) so it matches what the API returns.

mathildeleg (commented):
Thanks for addressing my feedback! I've still got a few suggestions, and I added the AI review to make sure we haven't missed anything :)

packages/core/content-manager/admin/src/services/documents.ts
Adzouz: I don't think it would be possible to select more than 100 entries because that's the highest amount of items a list can display so we should be good about it. I can add it anyway, just I don't think we need to display a message for that. What do you think?

Adzouz:
Hey @mathildeleg,
I (well AI and I) tried to address all the concerns raised. Can you take a look? 🙏

jhoward1994:
I think the Claude review might have a couple of good points here - mainly about performance 

e.g. 

> Any authenticated user with read access can now send _populate=deep to GET /content-manager/collection-types/:model to force populateDeep(Infinity) on an arbitrary list query.

> Performance — no ceiling on pageSize with deep population pageSize: Math.max(documentIds.length, 1) has no upper bound.

I'm not sure what's best. Maybe we could cap the server-side pageSize when x-strapi-populate: deep is active (e.g. 50 documents) ?
  
Or maybe (a lot of work and maybe something we can handle separately later but...) move the validation to the server.
  
Ideally if the bulkPublish controller validated each document against the published schema before publishing and returned per-document errors, the frontend wouldn't need to re-fetch anything with deep populate. It could just submit the documentIds, get back { documentId, errors }[], and display the results.

Adzouz:
Hey @mathildeleg and @jhoward1994, thanks a lot for the valuable feedback. Actually the changes for optimization I made were not working and we had the same issue as before so I reverted them (getting also rid of the deep population performance issue). Though I'd probably ask you for your input because it will basically go back to the first concern of @mathildeleg because we're looping over all items that we want to check and fetch them one by one (which is definitely not convenient). I'll keep you posted if I find another solution!

Adzouz:
Ok I think it's a bit better, I created a new specific and protected endpoint (admins only) that will perform a findMany (one request for all documentIds, with a limit of 100 items max - reflecting the max amount of items the user can select on the list view because of the pagination).
What do you think about this proposition @mathildeleg and @jhoward1994?

packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -629,6 +629,37 @@ export default {
...
+    await validateBulkActionInput(ctx.request.body);
+
+    const permissionChecker = getService('permission-checker').create({ userAbility, model });
+
+    if (permissionChecker.cannot.read()) {
+      return ctx.forbidden();
+    }
+
+    if (documentIds.length > 100) {
+      return ctx.badRequest('documentIds must not exceed 100 items');
```
hood: Is this documented behaviour? In case it's not, I think it'd make sense to add a reference in the docs to avoid users opening issues when this pops up and, if possible, also reference the docs from the error message.

Adzouz: It's reflecting the maximum amount of items you can select in the list view so we can document it but it's just a safety here, normally on the interface, the user is not allowed to select more than 100 items at once because of the pagination.

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
```diff
@@ -412,60 +413,45 @@ const SelectedEntriesModalContent = ({
...
+    isLoading,
+    isFetching,
+    refetch,
+  } = useGetDocumentsForValidationQuery(
     {
       model,
-      params: {
-        page: '1',
-        pageSize: documentIds.length.toString(),
-        sort: query.sort,
```
Bassel17, line 423: the new endpoint doesn't seem to receive a sort query, so am assuming this is not returning a properly sorted list ? not sure if sorting is important for validation but it is a regression from the old behavior perhaps

Bassel17 (changes requested):
Looks good overall but I have a few comments,
so for the 100 documents cap it probably is unnecessary, considering the ui already is limited with a 100 at a time, but if the UI changes it's behavior and we want to bulk publish more than a 100 then there might be a conflict here.
as for the performance concerns I think they will stay even if it's a 100 documents especially if those documents are contain deeply nested fields that require deep populate the more the worse the performance will be, solving that will be a problem we need to solve in a different PR probably and not here

packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/PublishAction.tsx
Adzouz: You're right, I'll put it back but it just triggered me another issue we currently have. The list in the modal is actually not sorted the same as the list view. Should I also fix that?

Bassel17: I mean if it's an easy fix, go ahead and let's merge this PR 💪

Adzouz: Done ✅

Bassel17 (dismissed):
LGTM!

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -265,15 +258,16 @@ const DocumentActions = ({ actions }: DocumentActionsProps) => {
  * DocumentActionButton
  * -----------------------------------------------------------------------------------------------*/
 
-interface DocumentActionButtonProps extends Action {
+interface DocumentActionButtonProps extends Omit<Action, 'type'> {
   buttonType?: 'button' | 'submit' | 'reset';
+  type?: string;
```
jhoward1994:
```suggestion
  type?: DocumentActionDescription['type'];
```

Would this be accurate and better than string ?

Adzouz: Fixed!

Bassel17 (approved):
LGTM!

mathildeleg (approved):
LGTM ✅✨

jhoward1994 (approved):

PR #25689, opened 2026-03-09
feat: sort based on publish status

dominik-juriga:
### What does it do?

Adds ListView ability to sort based on entry publication status by calculating status in the initial SQL query.

### Why is it needed?

Our teams (and also a few people in the issues) usually need to publish all unpublished entries and vice versa. The only way to do it without selecting entries individually is by using the content-management API with filters through a script. This way we can utilize natural flow of selecting all entries and de-selecting ones we don't need to perform bulk updates.

### How to test it?

1. Open a collection with Draft & publish enabled
2. Create a few entries, some published, some modified, some draft
3. Click the "Status" column header in the List View. Entries should be sorted.

If there is a combination with the `status` filter, the sort should be removed because this combination is redundant (filtered entries are already sorted by state).
1. With any `status` filter selected, the column header button should be disabled
2. If you have a sort based on `status` and add a `status` filter, the sort attribute should be removed from query params.
3. If you enter the query params manually, the `sort` attribute should be removed before getting the results and user should be redirected.

### Related issue(s)/PR(s)

No open issue for this PR.

### Screenshots

Screenshot 2026-03-09 at 16 35 40

Screenshot 2026-03-09 at 16 35 32


packages/core/content-manager/server/src/controllers/collection-types.ts
```diff
@@ -302,10 +381,18 @@ export default {
...
-      findPageParams as Parameters<typeof documentManager.findPage>[0],
-      model
-    );
+    const { results: documents, pagination } = statusSortOrder
+      ? await documentManager.findPageWithStatusSort(
+          { ...permissionQuery, populate, locale, status },
+          model,
+          statusSortOrder.toLowerCase() as 'asc' | 'desc'
+        )
+      : await documentManager.findPage({ ...permissionQuery, populate, locale, status }, model);
```
butcherZ: Shouldn't this use `findPageParams` (the variable built just above) instead of the hardcoded object?`findPageParams` includes the hasPublishedVersion pass-through and the `publicationStatusFilter` document-ID override, both get silently dropped here. The commented-out block below is actually the correct call 😄

packages/core/content-manager/server/src/services/document-manager.ts
```diff
@@ -83,6 +107,67 @@ const documentManager = ({ strapi }: { strapi: Core.Strapi }) => {
...
+     * Like findPage, but sorts results by the virtual `status` field at the DB level.
+     *
+     * The document service's validateParams rejects unknown sort keys (including `status`),
+     * so we bypass it here by calling strapi.db.query() directly for the findMany while
+     * still using strapi.documents().count() for accurate pagination totals.
+     */
+    async findPageWithStatusSort(
+      opts: DocServiceParams<'findMany'>,
+      uid: UID.CollectionType,
+      statusSortDirection: 'asc' | 'desc'
```
butcherZ: I wonder if we can avoid this bypass entirely? The reason status gets rejected is in `packages/core/utils/src/validate/validators.ts:122`, the `nonAttributesOperators` traversal only allowlists `id` and `documentId`. If we add status there for D&P-enabled models:                                                                                                        
```
 if (key === 'status' && ctx.schema?.options?.draftAndPublish) {
  return;
 } 
```                                                                                                                                                                           

then` strapi.documents(uid).findMany({ sort: 'status:ASC' })` passes validation and reaches the DB layer's CASE expression naturally. That would let us remove `findPageWithStatusSort` altogether and go through the normal findPage path, no need to manually re-implement locale filtering, D&P publishedAt logic, and pagination 🙂

packages/core/utils/src/content-types.ts
```diff
@@ -34,6 +35,7 @@ const constants = {
   CREATED_AT_ATTRIBUTE,
   UPDATED_AT_ATTRIBUTE,
   SINGLE_TYPE,
+  STATUS_ATTRIBUTE,
```
butcherZ: This constant doesn't seem to be used anywhere in the PR, is it needed here, or left over from an earlier iteration? 😄

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -154,6 +154,25 @@ const ListViewPage = () => {
...
+      /(?:^|,)\s*status:/i.test(query.sort)
+    ) {
+      const cleaned = query.sort
+        .split(',')
+        .map((s) => s.trim())
+        .filter((s) => !/^status:(ASC|DESC)$/i.test(s))
+        .join(',');
+      setQuery({ sort: cleaned || undefined }, 'push', true);
+    }
+  }, [hasStatusFilter]);
```
butcherZ: `query.sort` and `setQuery` are used inside this effect but aren't in the dependency array, React's exhaustive-deps rule would flag this. Adding them should be safe since `setQuery` is table and the effect already guards on `hasStatusFilter` 😄

butcherZ (changes requested):
Hey, @dominik-juriga thank you so muchfor this contribution! Sorting by publication status is a genuinely useful workflow improvement 😄 
I went through the code carefully and have a few things to flag before this can merge, Overall this is a solid direction, the main thing holding it back is the regression to existing status filters. 
Happy to review a follow-up once those are addressed! 🍺

packages/core/content-manager/server/src/controllers/collection-types.ts
dominik-juriga: You're right! Fixed in 168bdeb

packages/core/content-manager/server/src/services/document-manager.ts
dominik-juriga: This is a much cleaner solution, thanks for the suggestion!

packages/core/utils/src/content-types.ts
dominik-juriga: Yeah, it was a leftover from my previous approach to avoid the validation errors 😄 fixed in d0b8850

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
dominik-juriga: Good catch!

dominik-juriga: Fixed in e079ba7

butcherZ (approved):
LGTM! @dominik-juriga thank you so much for the fix :)

PR #25703, opened 2026-03-10
fix: preserve relations in fill from another locale

### LINKED ISSUE #14337: "Fill in from another locale" don't copy Relation field

- Node.js version: 14.19.1
- NPM version: 8.6.0
- Strapi version:4.3.6
- Database: mysql
- Operating system: Ubuntu 20.04

### Describe the bug

After create a new translation of a collection type I click on "Fill in from another locale" button, but a Relation field don't fill successfully

1. Go to Settings > Internationalization > Add new locales > Catalan
2. Set English as default language
3. Content Type Builder > Create new collection type > Test
4. Advanced settings > Check Enable localization for this Content-Type
5. Create field Text > Name: Name
7. Save
8. Content Type Builder > Create new collection type > Example
9. Advanced settings > Uncheck Enable localization for this Content-Type
10. Create field Text > Name: Name
11. Content Manager > Example > Create new entry: Name: Example 1
12. Save
13. Content Manager > Test > Create new entry: Name: Test 1, example: Example 1
14. Save
15. On locales dropdown, change to another locale (ex: Catalan)
16. Click on Fill in from another locale
17. On popup select English and click Yes
18. Then Name field is filled but Relation field is empty

### Expected behavior

Relation field has filled with English value (Example 1)

### Screenshots

image

### LINKED ISSUE #21322: [CM/i18n]  copy from locale doesnt show relations at first
### **Steps to reproduce**

* Create an entry in Articles with relations
* Create another version in French and import content from English
* Relations are not shown as imported

→ if you save the page you’ll see the relations were there but now shown.

### **Expected behavior**

Relations should be shown when importing content from another locale.

### **Screenshots**

[https://www.loom.com/share/a76593a1d4be4a77845ab94023840f45](<https://www.loom.com/share/a76593a1d4be4a77845ab94023840f45>)

Adzouz:
### What does it do?

Currently, the fill from another locale feature is not retrieving relations. This PR finally fixes that 🙌

- In the case of a localizable relation, we want to find the associated variation in the asked locale.
  - If it exists, we retrieve and populate it automatically
  - If not, we just leave the field empty
- In the case of a non-localizable relation, it's easy, we just retrieve the exact same entry.

It's taking the latest (most up to date) version of the document.

1. First test

In the following video you can see the retrieving of both localized (Category) and non-localized (Country) relations (also tested in dynamic zones and components).
In the second part of the video we remove the localized entry to test that it's not retrieved anymore.

https://github.com/user-attachments/assets/fa440f37-0161-41b7-baad-4a7b2047f76c

### Why is it needed?

This is a highly requested enhancement for this feature (see related issues / PRs) and was a big pain point raised by the community.

### How to test it?

You can watch the videos to reproduce the tests but basically you would need:
- 1 collection type A (non-localizable)
- 1 collection type B (localizable)
- 1 collection type C (also localizable)
  - 1 relation to A (no matter the nature of the relation)
  - 1 relation to B (same)

- Create an A entry
- Create a B entry (in English)
- Create another locale of the entry
- Create a C entry and select the newly B entry created (in English)
- Change the locale to the second you choose for your B entry
- Click on "Fill from another locale" and select it
-> You should see everything filled from the English locale and the correct translated relations

You can delete the variation you created and try the "Fill from another locale" again, this time, the localizable relation shouldn't be retrieved.

ℹ️ You can test with Draft and Publish enabled but you might also want to test with the feature disabled and it should still work the same.

The retrieval of relations should work in nested dynamic zones and components also.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/strapi/issues/14337
Fixes https://github.com/strapi/strapi/issues/21322
Part of the pain points raised by the community: https://github.com/strapi/strapi/discussions/25117#discussioncomment-15454521

🚀


packages/plugins/i18n/admin/src/services/fillFromLocale.ts
```diff
@@ -0,0 +1,27 @@
...
+  data: Record<string, unknown>;
+}
+
+const fillFromLocaleApi = i18nApi.injectEndpoints({
+  endpoints: (builder) => ({
+    getFillFromLocaleData: builder.mutation<FillFromLocaleResponse, FillFromLocaleArgs>({
+      query: (body) => ({
+        url: '/i18n/content-manager/actions/fill-from-locale',
+        method: 'POST',
+        data: body,
```
markkaylor: A bit confused why this get* function is making a POST request

packages/plugins/i18n/server/src/routes/admin.ts
```diff
@@ -73,6 +73,14 @@ export default {
         policies: ['admin::isAuthenticatedAdmin'],
       },
     },
+    {
+      method: 'POST',
```
markkaylor: GET ?

packages/plugins/i18n/server/src/controllers/content-types.ts
```diff
@@ -84,6 +87,31 @@ const controller = {
...
+
+  async getFillFromLocaleData(ctx) {
+    await validateFillFromLocaleInput(ctx.request.body);
+
+    const { model, documentId, sourceLocale, targetLocale } = ctx.request.body as {
+      model: string;
+      documentId?: string;
+      sourceLocale: string;
+      targetLocale: string;
+    };
```
markkaylor: I think you are using this type on the FE too, could be a shared type coming from contacts

packages/plugins/i18n/server/src/services/index.ts
```diff
@@ -20,4 +21,5 @@ export default {
   'ai-localizations': createAILocalizationsService,
   'ai-localization-jobs': createAILocalizationJobsService,
   settings: createSettingsService,
+  'fill-from-locale': fillFromLocale,
```
markkaylor: could you create your service the same way as the others to respect that pattern. createServiceName and should receive strapi as an argument

packages/plugins/i18n/server/src/services/fill-from-locale.ts
```diff
@@ -0,0 +1,382 @@
...
+  'id',
+  'documentId',
+  'publishedAt',
+  'strapi_stage',
+  'strapi_assignee',
+  'locale',
+  'status',
+]);
+
+const STATUS_FIELDS = new Set(['id', 'documentId', 'locale', 'updatedAt', 'publishedAt']);
```
markkaylor: I don't really mind but curious why you use a Set instead of a simple array?

packages/plugins/i18n/server/src/services/fill-from-locale.ts
```diff
@@ -0,0 +1,382 @@
...
+): Promise<Array<{ documentId: string; id: number | string; locale?: string } | null>> => {
+  const i18nContentTypesService = strapi.plugin('i18n')?.service('content-types');
+  const isTargetLocalized = i18nContentTypesService?.isLocalizedContentType(
+    strapi.getModel(targetUid)
+  );
+
+  if (!isTargetLocalized) {
+    return relations.map((rel) => ({
+      documentId: rel.documentId,
+      id: rel.id ?? rel.documentId,
```
markkaylor: Should id ever be documentId?

packages/plugins/i18n/server/src/services/fill-from-locale.ts
```diff
@@ -0,0 +1,382 @@
...
+import { contentTypes } from '@strapi/utils';
+import type { UID, Schema } from '@strapi/types';
+
+const getMainFieldForModel = (targetUid: UID.Schema): string => {
+  const model = strapi.getModel(targetUid);
+  const firstString = Object.keys(model?.attributes || {}).find(
+    (key) => model.attributes[key]?.type === 'string' && key !== 'id'
+  );
+  return firstString || 'id';
+};
```
markkaylor, line 19: I don't think this will get the correct `mainField`. `mainField`, or "Entry title" as it is referred to in the UI, is something configurable by the user and is stored in the database. You can see how it is fetched for example in configure the view:

Image

I believe content releases has to get the mainField for each content-type as well so you might be able to reference how it is done there

markkaylor (commented):
Some thoughts but haven't tested it yet

packages/plugins/i18n/server/src/services/fill-from-locale.ts
Adzouz: It could be an array, it was basically for two reasons: uniqueness of elements to avoid duplicates and also because `Set.has()` is more efficient than `Array.includes()` but for something that simple, it agree it doesn't really matter

packages/plugins/i18n/server/src/routes/admin.ts
```diff
@@ -73,6 +73,14 @@ export default {
...
+      path: '/content-manager/actions/fill-from-locale',
+      handler: 'content-types.getFillFromLocaleData',
+      config: {
+        policies: ['admin::isAuthenticatedAdmin'],
```
mathildeleg: Don't you need to add more permission checks here? 

```suggestion
        policies: [
  'admin::isAuthenticatedAdmin',
  {
    name: 'plugin::content-manager.hasPermissions',
    config: { actions: ['plugin::content-manager.explorer.read'] },
  },
],
```

mathildeleg (commented):
I'm still a bit worried about the perf issues here, here are some ideas, would love to hear your thoughts on this:

1. Shallow initial load + fetch on expand

Right now we load the full document with `populateDeep(Infinity)` and resolve relations for the entire tree in one go. We could instead:

- Initially only fetch and run fill-from-locale for a limited depth (e.g. 1–2 levels of relations/components/dynamic zones). The first response would only contain data for that shallow slice; nested relation fields could be left as “not loaded” or empty.
- When the user opens a component or dynamic zone block (accordions are closed by default), the client would call the same fill-from-locale endpoint (or a dedicated “fill-from-locale for path”) with e.g. documentId, sourceLocale, targetLocale, and a path (e.g. blocks[0].relatedArticles). The server would then load and resolve only that subtree and return it so the UI can fill just that part of the form.
That would reduce the initial payload and query count for heavy content types, and spread the cost over user interaction instead of doing everything up front. BUT it would imply calling the fill-from-locale endpoint whenever opening a component/dynamic zone, so this is not ideal for sure.

2. Batch relation resolution by target content type

Today we call `transformRelationsForLocale` (and thus resolve + status + labels) once per relation attribute in the tree. We could:
- Collect in one pass: for every relation field, record path, targetUid, and the list of relations, and build the output structure with placeholders for those fields.
- Group by targetUid, then for each target UID merge all relations from every attribute pointing to it, and run the three steps (resolve, status, labels) once for that merged list.
- Split the resolved list back by attribute (using chunk lengths) and assign each chunk to the correct path in the result.
That would change the cost from “2–3 queries per relation attribute” to “2–3 queries per distinct target content type,” which can cut a large number of queries when many fields point to the same UID or when there are many nested relation fields.

I haven't done deep QA on this yet, waiting for your last changes, but just some thoughts :)

packages/plugins/i18n/server/src/routes/admin.ts
Adzouz: To make it work, I needed to pass the model as a param to the URL. It's done ✅

Adzouz:
@mathildeleg About 1. I'm not really fan about the proposition to be honest because it will change from what's already in place for the base edit view. I don't think it's an issue though since it's already loading the data this way in the edit view. User expect a bit more time if there's a lot of relations. The fill from another locale will block the page during the process so I guess the waiting time is ok. Plus, if we want to implement this logic, we would need a way to differenciate when we're in the mode fill from another locale and it might be tricky.
As discussed together, I think we could move forward and see if it needs some adjustments because people complain about performance but as long as this feature is now working as expected, I think that's the most important thing.
For 2. that's a good idea, I revamped the logic to batch by content type id and thanks to that, we would have way less requests so it's part of a better optimized experience! Good catch 😊

Let me know if we can move forward after testing again and reviewing the changes 🙏

packages/plugins/i18n/server/src/services/fill-from-locale.ts
```diff
@@ -0,0 +1,517 @@
...
+      return documentId
+        ? docs.findOne({ ...baseParams, documentId })
+        : docs.findFirst({ ...baseParams });
+    },
+
+    /**
+     * Transform a (sanitized) document: strip internal fields, resolve relations to the target
+     * locale, and skip relations to content types the user cannot read.
+     */
+    async transformDocument(
```
mathildeleg, line 499: I sometimes encounter this issue when attempting to fill from default locale + hitting save (just doing "fill in from ... locale" doesn't show the error)
Image
I think it may be happening due to a conflict of what ends up in the form after fill vs what the rest of the admin assumes is on document.
The fill endpoint returns a transformed copy of the source locale. On the server, anything that isn’t a real schema field is still passed through as-is if it’s present on the fetched document. `localizations` falls in that bucket — it’s i18n metadata, not an actual field you edit. So after Fill in from another locale, `setValues` can merge that into the current document in a shape that isn’t always a plain array.

In `CMHeaderActions.ts`, other code (including paths that run when you save) assumes `document.localizations` is either missing or an array, and does something like `(document.localizations ?? []).map(...). ??` only falls back when the value is null or undefined. If localizations is some other truthy value (e.g. an object), the fallback never runs and .map blows up — which matches the error.

Suggestion: strip metadata such as `localizations` in fill-from-locale’s transform, so **the fill payload only contains real form data.** Optionally we can still add a small `Array.isArray` guard in the UI as extra safety, but I'm not in love with the idea as that would be bandaging an issue, not really fixing it.

Adzouz: Good catch, I added `localizations` to the list of fields to ignore in the transform function so they will keep their original data 🙏

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
```diff
@@ -552,23 +549,30 @@ const FillFromAnotherLocaleAction = ({
...
-      params: { locale: localeSelected },
-    });
-    if (!response || !schema) {
+    if (!localeSelected || !currentDesiredLocale) {
       return;
     }
 
-    const { data } = response;
-
-    const cleanedData = cleanData(data, schema, components);
```
markkaylor, line 567: What was cleanData and why don't need it anymore?

packages/plugins/i18n/server/src/services/fill-from-locale.ts
```diff
@@ -0,0 +1,518 @@
...
+  };
+
+  collect(data, schema);
+  return result;
+};
+
+/**
+ * Perform the 3 DB queries (resolve + status + labels) grouped by targetUid for optimal performance.
+ */
+const prefetchResolutionData = async (
```
markkaylor: I don't really understand what this function does... what is meant by "resolutionData?

packages/plugins/i18n/server/src/routes/admin.ts
```diff
@@ -73,6 +73,20 @@ export default {
         policies: ['admin::isAuthenticatedAdmin'],
       },
     },
+    {
+      method: 'GET',
+      path: '/content-manager/actions/fill-from-locale/:model',
```
markkaylor: The naming of a GET method under a path `*/actions/*` is a bit contradictory. I'm also sort of confused are we fetching the data and then on the client making a POST request to create the localized entry? Would it not make sense to do it as an atomic action, POST fillFromLocale which gets the data from the source and creates the target?

markkaylor (commented):
I tested:

- Localized entry fills non-localized relation
- Localized entry fill localized relation if relation has a version in the same locale
- Tested the same inside a dz component

As far as I can tell it appears to be working as expected.

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
Adzouz: It was to remove fields from the data (like passwords / secrets and relations). Now the relations are retrieved I just rewrote the whole logic and moved this to another method called `processDocumentData` but I could call it `cleanData` if you prefer

packages/plugins/i18n/server/src/services/fill-from-locale.ts
Adzouz: It batch loads the relations for a document of a target locale.
Example let's say you want to fill a "Article" type that has multiple "Category" and "Author" relations from French to English. This will fetch the data of the relations in English + handle their status and label too to correctly be able to load them in the UI after (in batch, like grouped by content types to optimize performance and fetch all same types in one query and not a query per relation).
I get that the name is not ideal. What do you think about maybe something like `resolvedBatchedRelations`?

markkaylor: Got it, yeah I think `resolveBatchedRelations` or `getAllRelations` could be more clear thanks for the explanation.

packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx
markkaylor: ok that's fine thanks 👍

packages/plugins/i18n/server/src/services/fill-from-locale.ts
Adzouz: I'll go with `resolveAllRelationsBatched` because we're actually also processing them for the status and label and I want to highlight that the data passed is grouped

packages/plugins/i18n/server/src/routes/admin.ts
Adzouz: Changed it for `/content-manager/get-fill-from-locale/:model` after discussing together 🙏

markkaylor (approved):
Nice work! Tested again and it's working 👏

mathildeleg (commented):
Tested again and LGTM ✅
Well done ✨

mathildeleg (approved):

PR #25709, opened 2026-03-11
feat(upload): add Vimeo ID import support in Media Library

nicosek:
This patch adds support for importing a video directly from a Vimeo ID in the Media Library.

Changes include:
- new admin form "From Vimeo ID" in the upload dialog
- new admin route and controller to resolve Vimeo metadata
- integration with the existing upload service to create the media entry

This allows content managers to import Vimeo videos without manually retrieving the video URL.


markkaylor:
Hello, thank you for submitting a PR. 

This is actually not a patch but a feature. Unfortunately I don't think it is a feature we will pursue for the following reasons:

- This should already be achievable with the from url feature. If you are copying the id why not copy the full url?
- This feature is platform specific. Why should we add this for vimeo and not x platform?

If you need this behavior I think you could explore creating a plugin. Otherwise I recommend submitting a feature request as described in the [contributing guide](https://github.com/strapi/strapi/blob/develop/CONTRIBUTING.md#feature-requests).

Thanks again for your interest and effort in improving Strapi 🚀

nicosek:
> Hello, thank you for submitting a PR.
> 
> This is actually not a patch but a feature. Unfortunately I don't think it is a feature we will pursue for the following reasons:
> 
> * This should already be achievable with the from url feature. If you are copying the id why not copy the full url?
> * This feature is platform specific. Why should we add this for vimeo and not x platform?
> 
> If you need this behavior I think you could explore creating a plugin. Otherwise I recommend submitting a feature request as described in the [contributing guide](https://github.com/strapi/strapi/blob/develop/CONTRIBUTING.md#feature-requests).
> 
> Thanks again for your interest and effort in improving Strapi 🚀

Thanks for the quick feedback, and apologies for the confusion.

This PR was actually intended as a patch for our own usage rather than a proposal to merge upstream. I mistakenly opened it against the official repository instead of keeping it on my fork.

Your points make sense regarding platform specificity.

Thanks again for taking the time to review it, and sorry for the noise!

Issue #25592, opened 2026-03-01
Media Library broken - VERY SERIOUS


nevil2:
### Node Version

24.1.0

### NPM/Yarn/PNPM Version

NPM 11.10.1

### Strapi Version

5.37.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

We are now facing failures of newly loaded images to be display in multiple Strapi instances in the media library.

We are using @strapi/provider-upload-aws-s3 to upload images to Digital Ocean.

Until very recently, the URL would have the bucket before the domain, e.g.

https://pcstrapi.fra1.digitaloceanspaces.com/thumbnail_IMG_1351_e41afed02d.jpg

But now the bucket is after the domain, e.g.
 
https://fra1.cdn.digitaloceanspaces.com/pcstrapi/thumbnail_IMG_1351_e41afed02d.jpg

this caused havoc with our front-ends using the cdn, but I've made fix for this.

But more bizarrely, the bucket after the domain, whilst loading in the browser, is not loading in Strapi media library.  
This is probably a CORS issue due to bucket being in wrong place.

What has the Strapi team done to create this bug. 

### Steps to Reproduce

<img width="496" height="205" alt="Image" src="https://github.com/user-attachments/assets/cce7fc1d-8684-4d2c-9035-2a8467557fe7" />

https://fra1.digitaloceanspaces.com/pcstrapi/thumbnail_IMG_1351_2e84da4948.jpeg

### Expected Behavior

To work as previously

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

nevil2:
Looking at docs, it is as if, for Digital Ocean, it is using forcePathStyle: true despite the default being false.

nevil2:
Setting forcePathStyle: false does not fix the malformed URL on new uploads.

Adding new path to contentSecurityPolicy does enable images to show in media library.

Moses-main:
Hi team, I'd like to contribute. Please assign!

nclsndr:
Thank you for volunteering @Moses-main. We don't work with issue assignment. Yet, you can work on a fix an raise a PR with the issue referenced, we'd be happy to review it.

---

[...]

PR #25720, opened 2026-03-12
fix: generated provider url wrongly formatted causing CORS/CSP err in…

ChisomUma:
### What does it do?

Modifies `constructFileUrl` in `packages/providers/upload-aws-s3/src/index.ts` to respect the forcePathStyle option from the S3 client config when building the stored file URL:

```
forcePathStyle: true → path-style URL: https://{endpoint}/{bucket}/{key}
forcePathStyle: false (default) → virtual-hosted-style URL: https://{bucket}.{endpoint-host}/{key}
```

Previously, the function always produced path-style URLs for any provider with a custom endpoint, regardless of forcePathStyle.

Also adds test coverage for both URL styles for DigitalOcean Spaces and expands  `isUrlFromBucket` tests to cover CDN path-style DO URLs.

### Why is it needed?

When using `@strapi/provider-upload-aws-s3` with DigitalOcean Spaces (or any S3-compatible provider that supports virtual-hosted-style URLs), setting a custom endpoint caused Strapi to store path-style URLs (https://fra1.digitaloceanspaces.com/pcstrapi/image.jpg) even though the AWS SDK.  With its default of forcePathStyle: false actually serves the object at the virtual-hosted URL (https://pcstrapi.fra1.digitaloceanspaces.com/image.jpg). This mismatch broke image display in the Media Library due to CORS/CSP mismatches.

### How to test it?

Automated test:
```
node_modules/.bin/jest packages/providers/upload-aws-s3 --no-coverage
# Expected: 83 tests pass
```

Manual (DigitalOcean Spaces):

Configure the plugin with a DO Spaces endpoint and no forcePathStyle:

```
// config/plugins.js
upload: {
  config: {
    provider: 'aws-s3',
    providerOptions: {
      s3Options: {
        endpoint: 'https://fra1.digitaloceanspaces.com',
        // forcePathStyle not set (defaults to false)
        params: { Bucket: 'my-bucket' },
      },
    },
  },
},

```

Upload an image, verify the stored URL is `https://my-bucket.fra1.digitaloceanspaces.com/...` and the image loads in the Media Library.

NOTE: Users currently relying on path-style URLs with a custom endpoint (MinIO, IONOS, Wasabi, etc.) must add forcePathStyle: true to their s3Options to preserve the previous behavior.

### Related issue(s)/PR(s)

#25592

### DIFF
--- a/packages/providers/upload-aws-s3/src/__tests__/is-url-from-bucket.test.ts
+++ b/packages/providers/upload-aws-s3/src/__tests__/is-url-from-bucket.test.ts
@@ -43,14 +43,26 @@ describe('Test for URLs', () => {
 
   describe('S3 Compatible', () => {
     describe('DO Spaces', () => {
-      test('is from same bucket', async () => {
+      test('virtual-hosted-style: is from same bucket', async () => {
         const url = 'https://bucket-name.nyc3.digitaloceanspaces.com/folder/img.png';
         const isFromBucket = isUrlFromBucket(url, 'bucket-name');
         expect(isFromBucket).toEqual(true);
       });
-      test('is not from same bucket', async () => {
+      test('virtual-hosted-style: is not from same bucket', async () => {
         const url = 'https://bucket-name.nyc3.digitaloceanspaces.com/folder/img.png';
         const isFromBucket = isUrlFromBucket(url, 'bucket');
+        expect(isFromBucket).toEqual(false);
+      });
+      test('CDN path-style (forcePathStyle:true): bucket in path, is from same bucket', async () => {
+        // URL generated when endpoint is set to the CDN and forcePathStyle:true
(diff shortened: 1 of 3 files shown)


innerdvations:
Thanks for the careful diagnosis and tests — the #25263 path-style-always behavior for custom endpoints was a real regression for DigitalOcean Spaces.

That issue is already fixed on `develop` / released: #25939 (v5.42.1) changed `constructFileUrl` to trust a valid S3 `Location` before reconstructing from `endpoint`, which restores DO/Scaleway virtual-hosted URLs. #25592 was closed on that basis.

This branch is also based on the pre-#25939 priority order (endpoint before Location) and is far behind `develop`, so landing it would risk regressing that fix. Closing as already addressed by #25939.

If something still reproduces after upgrading to the latest release, please open a new issue with your provider config and the exact `Location` returned by the SDK — happy to dig further.

_AI-assisted comment._

PR #25723, opened 2026-03-12
RBAC "Move to stage" for review workflows

jhoward1994:
### What does it do?

- **Adds "to" stage permissions for Review Workflows.** Previously, permissions only controlled who can move content *out* of a stage ("from"). This change adds explicit "to" permissions so you can control which roles can move content *into* each stage.
- **Transition rules:** Moving from stage A → B now requires: (1) "From" permission on stage A (unchanged), and (2) "To" permission on stage B (new).
- **Backward compatibility:** If a stage has no "to" permissions configured, all roles are allowed to move into it (same as before).

### Why is it needed?

Today, any role with "from" permission on a stage can move content to *any* other stage. That makes it impossible to restrict who can move content into sensitive stages (e.g. "Published"), so Editors could bypass review and publish directly. "To" permissions close that gap and allow proper RBAC for review workflows (e.g. only Admins can move into "Published").

### How to test it?

1. **Unit/API tests:** Run `yarn jest packages/core/review-workflows` and the new API test `tests/api/core/review-workflows/review-workflows-stage-permissions.test.api.ts` (or your project's API test command for that path).
2. **Manual:** Create a workflow (e.g. Draft → Review → Published). Set "to" permission for "Published" to Admin only; leave "from" as usual (e.g. Editor + Admin). 
As Editor: confirm you can move Draft → Review, but "Published" is not in the stage dropdown and a direct API call to move to Published returns 403. 
As Admin: confirm you can move to Published. 
Create or use a workflow with no "to" permissions; confirm all roles can still move into those stages (backward compatibility).

### Related issue(s)/PR(s)

- Fixes CMS-314

### DIFF
--- a/packages/core/review-workflows/server/src/services/stages.ts
+++ b/packages/core/review-workflows/server/src/services/stages.ts
@@ -53,9 +53,9 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
         const allPermissionIds: number[] = [];
 
         // Register "from" permissions
-        if (stage.permissions && stage.permissions.length > 0) {
+        if (stage.fromPermissions && stage.fromPermissions.length > 0) {
           const fromPermissions = await async.map(
-            stage.permissions,
+            stage.fromPermissions,
             (permission: StagePermission) =>
               stagePermissionsService.register({
                 roleId: permission.role,
@@ -97,7 +97,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 
     async update(srcStage: any, destStage: any) {
       const stageId = destStage.id;
-      const hasPermissionChanges = destStage.permissions || destStage.toPermissions;
+      const hasPermissionChanges = destStage.fromPermissions || destStage.toPermissions;
 
       let allPermissionIds: number[];
 
@@ -112,7 +112,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 
         // Re-register "from" permissions (fall back to existing "from" entries from DB)
         const fromPerms =
-          destStage.permissions ??
+          destStage.fromPermissions ??
           (srcStage?.permissions ?? [])
             .filter((p: PopulatedPermission) => p.actionParameters?.from)
             .map((p: PopulatedPermission) => ({ role: extractRoleId(p), action: p.action }));
@@ -373,11 +373,11 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 const normalizeStageForDiff = (stage: {
   name?: string;
   color?: string;
-  permissions?: unknown[];
+  fromPermissions?: unknown[];
   toPermissions?: unknown[];
 }) => ({
   ...pick(['name', 'color'], stage),
-  permissions: stage.permissions ?? [],
+  fromPermissions: stage.fromPermissions ?? [],
   toPermissions: stage.toPermissions ?? [],
 });
 
--- a/packages/core/review-workflows/server/src/validation/review-workflows.ts
+++ b/packages/core/review-workflows/server/src/validation/review-workflows.ts
@@ -23,7 +23,7 @@ const stageObject = yup.object().shape({
   id: yup.number().integer().min(1),
   name: yup.string().max(255).required(),
   color: yup.string().matches(/^#(?:[0-9a-fA-F]{3}){1,2}$/i), // hex color
-  permissions: yup.array().of(fromPermissionObject),
+  fromPermissions: yup.array().of(fromPermissionObject),
   toPermissions: yup.array().of(toPermissionObject),
 });
 
--- a/tests/api/core/review-workflows/review-workflows-stage-permissions.test.api.ts
+++ b/tests/api/core/review-workflows/review-workflows-stage-permissions.test.api.ts
@@ -110,7 +110,7 @@ describeOnCondition(edition === 'EE')('Review workflows', () => {
         stages: [
           {
             ...baseWorkflow.stages[0],
-            permissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
+            fromPermissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
           },
           baseWorkflow.stages[1],
         ],
@@ -127,7 +127,7 @@ describeOnCondition(edition === 'EE')('Review workflows', () => {
         stages: [
           {
             ...baseWorkflow.stages[0],
-            permissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
+            fromPermissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
           },
           baseWorkflow.stages[1],
         ],
@@ -138,7 +138,7 @@ describeOnCondition(edition === 'EE')('Review workflows', () => {
         stages: [
           {
             ...workflow.stages[0],
-            permissions: getStageTransitionPermissions([roles[0].id]),
+            fromPermissions: getStageTransitionPermissions([roles[0].id]),
           },
           workflow.stages[1],
         ],
@@ -161,7 +161,7 @@ describeOnCondition(edition === 'EE')('Review workflows', () => {
         stages: [
           {
             ...baseWorkflow.stages[0],
-            permissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
+            fromPermissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
           },
           baseWorkflow.stages[1],
         ],
@@ -188,7 +188,7 @@ describeOnCondition(edition === 'EE')('Review workflows', () => {
         stages: [
           {
             ...baseWorkflow.stages[0],
-            permissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
+            fromPermissions: getStageTransitionPermissions([roles[0].id, roles[1].id]),
           },
         ],
       });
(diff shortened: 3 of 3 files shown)


packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
```diff
@@ -40,7 +40,8 @@ import { useDragAndDrop } from '../hooks/useDragAndDrop';
 
 import { AddStage } from './AddStage';
 
-interface WorkflowStage extends Pick<IStage, 'id' | 'name' | 'permissions' | 'color'> {
+interface WorkflowStage
+  extends Pick<IStage, 'id' | 'name' | 'permissions' | 'toPermissions' | 'color'> {
```
markkaylor: So are "permissions" actually the "fromPermissions"? If so I kinda feel like the right thing to do is to rename permissions "fromPermissions" 😬

packages/core/review-workflows/admin/src/routes/settings/components/Stages.tsx
```diff
@@ -576,12 +597,8 @@ const PermissionsField = ({ disabled, name, placeholder, required }: Permissions
       <Flex alignItems="flex-end" gap={3}>
         <PermissionWrapper grow={1}>
           <Field.Root error={error} name={name} required>
-            <Field.Label>
-              {formatMessage({
-                id: 'Settings.review-workflows.stage.permissions.label',
-                defaultMessage: 'Roles that can change this stage',
-              })}
-            </Field.Label>
+            <Field.Label>{label}</Field.Label>
```
markkaylor, line 600: I can't really find where this is passed but we are sure it's a translated label right?

packages/core/review-workflows/admin/src/routes/settings/id.tsx
```diff
@@ -186,10 +208,21 @@ const EditPage = () => {
...
                 );
+
+              hasUpdatedToPermissions =
+                serverStage.toPermissions?.length !== stage.toPermissions?.length ||
+                !serverStage.toPermissions?.every(
+                  (serverPermission) =>
+                    !!stage.toPermissions?.find(
+                      (permission) => permission.role === serverPermission.role
+                    )
+                );
```
markkaylor: I mean it's the same as what was already there but this thing is kinda wild 🤯

packages/core/review-workflows/server/src/controllers/workflows.ts
```diff
@@ -24,15 +24,42 @@ function getWorkflowsPermissionChecker({ strapi }: { strapi: Core.Strapi }, user
...
 
-  // Transform permissions roles to be the id string instead of an object
-  const transformPermissions = map(update('role', property('id')));
-  const transformStages = map(update('permissions', transformPermissions));
+  const transformRoleToId = (permission: any) => ({
+    ...permission,
+    role: typeof permission.role === 'object' ? permission.role.id : permission.role,
+  });
+
+  const transformStages = map((stage: any) => {
```
markkaylor: I'd be surprised if we didn't have a type for Stage...

packages/core/review-workflows/server/src/controllers/workflows.ts
```diff
@@ -24,15 +24,42 @@ function getWorkflowsPermissionChecker({ strapi }: { strapi: Core.Strapi }, user
...
+    const allPermissions: any[] = stage.permissions ?? [];
+
+    const fromPermissions = allPermissions
+      .filter((p: any) => p.actionParameters?.from)
+      .map(transformRoleToId);
+
+    const toPermissions = allPermissions
+      .filter((p: any) => p.actionParameters?.to)
```
markkaylor: I wonder if we can do better than any here?

packages/core/review-workflows/server/src/services/stage-permissions.ts
```diff
@@ -53,5 +70,50 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
+     */
+    canTransitionToStageWithPermissions(stage: any) {
+      const requestState = strapi.requestContext.get()?.state;
+
+      if (!requestState) {
+        return false;
+      }
+
+      const userRoles = requestState.user?.roles;
+      if (userRoles?.some((role: any) => role.code === 'strapi-super-admin')) {
```
markkaylor: Same as above, don't we have a type for Role?

packages/core/review-workflows/server/src/services/stage-permissions.ts
```diff
@@ -53,5 +86,57 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
+      const userRoles = requestState.user?.roles;
+      if (userRoles?.some((role: { code: string }) => role.code === 'strapi-super-admin')) {
+        return true;
+      }
+
+      const toPermissions = (stage.permissions || []).filter(
+        (p: { actionParameters?: { to?: unknown } }) => p.actionParameters?.to
+      );
+
+      const userRoleIds = new Set(userRoles.map((role: { id: number }) => role.id));
```
butcherZ: be careful here, if `userRoles` is undefined, it is going to throw a type error. Maybe we can do something like

```
  import type { SanitizedAdminRole } from '@strapi/admin/shared/contracts/shared';
                                                                                                                                                                        
  const getUserRoles = (): SanitizedAdminRole[] | null => {                                                                                                             
    const requestState = strapi.requestContext.get()?.state;                                                                                                            
    if (!requestState) return null;                                                                                                                                     
    return requestState.user?.roles ?? null;                                                                                                                            
  };                                                                                                                                                                    
                                                                                                                                                                        
  Then in each method:                                                                                                                                                  
                                         
  const userRoles = getUserRoles();
  if (!userRoles) {                                                                                                                                                     
    return false;
  }     
```

packages/core/review-workflows/server/src/validation/review-workflows.ts
```diff
@@ -3,20 +3,28 @@ import { yup, validateYupSchema } from '@strapi/utils';
...
-    yup.object().shape({
-      role: yup.number().integer().min(1).required(),
-      action: yup.string().oneOf([STAGE_TRANSITION_UID]).required(),
-      actionParameters: yup.object().shape({
-        from: yup.number().integer().min(1).required(),
-        to: yup.number().integer().min(1),
-      }),
-    })
-  ),
+  permissions: yup.array().of(fromPermissionObject),
```
butcherZ: shall we just rename the field name to:
`fromPermissions: yup.array().of(fromPermissionObject),` ? :D

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -41,34 +48,46 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
-      // Create stage permissions
       await async.reduce(stagesList)(async (_, stage, idx) => {
-        // Ignore stages without permissions
-        if (!stage.permissions || stage.permissions.length === 0) {
-          return;
+        const stageId = stages[idx].id;
+        const allPermissionIds: number[] = [];
+
+        // Register "from" permissions
+        if (stage.permissions && stage.permissions.length > 0) {
```
butcherZ: following rename :)

`if (stage.fromPermissions && stage.fromPermissions.length > 0)`

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -41,34 +48,46 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
+          const fromPermissions = await async.map(
+            stage.permissions,
```
butcherZ: stage.fromPermissions

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -77,27 +96,66 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
     },
 
     async update(srcStage: any, destStage: any) {
-      let stagePermissions = srcStage?.permissions ?? [];
       const stageId = destStage.id;
+      const hasPermissionChanges = destStage.permissions || destStage.toPermissions;
```
butcherZ: `destStage.fromPermissions || destStage.toPermissions`

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -77,27 +96,66 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
-        );
-        stagePermissions = permissions.flat().map((p: any) => p.id);
+        allPermissionIds = [];
+
+        const extractRoleId = (p: PopulatedPermission): number =>
+          typeof p.role === 'object' ? p.role.id : p.role;
+
+        // Re-register "from" permissions (fall back to existing "from" entries from DB)
+        const fromPerms =
+          destStage.permissions ??
```
butcherZ: destStage.fromPermissions ??

packages/core/review-workflows/server/src/services/stages.ts
```diff
@@ -313,6 +370,17 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
...
 };
 
+const normalizeStageForDiff = (stage: {
+  name?: string;
+  color?: string;
+  permissions?: unknown[];
+  toPermissions?: unknown[];
+}) => ({
+  ...pick(['name', 'color'], stage),
+  permissions: stage.permissions ?? [],
```
butcherZ: here as well :)

`fromPermissions?: unknown[]`

`permissions: stage.fromPermissions ??       
  stage.permissions ?? []      `

Issue #23618, opened 2025-05-28
API response does not update when editing or removing published relational fields


sardinka88:
**@pierre Needs some visibility over this issue here: **[**https://strapihq.slack.com/archives/C01J6DEK1RS/p1769433272863989**](<https://strapihq.slack.com/archives/C01J6DEK1RS/p1769433272863989>)

### Node Version

22.16.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.11.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

We need to escalate a critical technical issue that is significantly disrupting our publishing workflow and adversely affecting the overall user experience on our website.

The issue specifically pertains to **relational fields**. We consistently encounter problems whenever relational entries are updated or removed.

**To illustrate one of many examples:**

We utilize a collection type named "Qualifiers," which editors render in form of cards containing specific information on various pages through a component called "Qualifiers Grid."

This component includes a relational dropdown allowing editors to select multiple Qualifiers for display. within a Grid component.

Problems arise when editors either remove Qualifiers from this grid or modify Qualifier details (e.g., updating logos, dates or names).

**Expected behavior:**

Any published modifications to Qualifier entries should automatically and immediately (after cache invalidates) reflect within the Qualifiers Grid.

Removing Qualifiers from the grid and subsequently publishing the page should result in those Qualifiers no longer appearing in the API response, and ultimately on the page.

**Actual behavior:**

Changes do not reflect immediately. To make changes appear, editors are forced to unpublish and then republish the affected Qualifier entries and often times the page where the block is rendered (it is difficult to determine the exact workaround flow).

Similarly, removing references from the Qualifiers Grid and publishing the page does not remove these Qualifiers from the API response as expected. In order to have them removed, they again need to unpublish and republish the affected Qualifier entries or page where the block is being rendered.

This issue significantly hampers our editorial efficiency by necessitating repeated unpublishing and republishing steps. Notably, we did not encounter this issue last year when using Strapi version 4 and its relational fields.

The issue is not limited to the above Qualifiers example, we are experiencing it wherever a relation is used, and in our Project we heavily rely on relations.

### Steps to Reproduce

**Scenario 1 - content not updating**

 1. Add a block component that renders entries fetched relationally from a Collection to any page
 2. Using a relations dropdown, select several collection entries that are to be rendered within the block
[...]

PR #25725, opened 2026-03-12
fix(document-service): re-insert cascade-deleted bidirectional relations

butcherZ:
## What does it do?
Fixes bidirectionalRelations.sync() in the document service to re-insert join table entries that were removed by FK cascade delete during publish/discard-draft operations.

The fix uses an explicit existence check (SELECT … FIRST) before deciding to UPDATE (row exists, fix order) or INSERT (row was cascade-deleted, restore it). A plain UPDATE affected-row count was avoided because MySQL/MariaDB returns 0 for no-op updates (value unchanged), which would have caused spurious duplicate row inserts.

## Why is it needed?

When editors republish a related entry that another content type or component references via a bidirectional relation, the API response for the referencing entry returns stale or missing data until the Page is manually unpublished and republished.

## How to test it?

Run the new regression tests

`yarn test:api --testPathPattern="bidirectional-relations"`

Verify no regression in existing relation tests

`yarn test:api --testPathPattern="(bidirectional|unidirectional)-relations"`

## Manual Test

- Create a Tag content type (D&P) and a Product content type (D&P) with a bidirectional manyToMany to Tag
- Create and publish a Tag, create and publish a Product referencing that Tag
- Update the Tag's name and republish it
- `GET /api/products/:id?populate=tags&status=published`

Before fix: tags is empty or shows old data. 
After fix: tags contains the Tag with the updated name

## Related issue(s)/PR(s)

Fixes CMS-237
#23618

### DIFF
--- a/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
@@ -1,6 +1,5 @@
 /* eslint-disable no-continue */
 import { keyBy, omit } from 'lodash/fp';
-import { async } from '@strapi/utils';
 import type { UID, Schema } from '@strapi/types';
 import type { JoinTable } from '@strapi/database';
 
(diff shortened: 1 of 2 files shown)


jhoward1994:
There's a failing API test that I think highlights an actual bug. 

Do we need to guard against the 'newRelation' logic like this? 

So if a relation has been removed from a draft that is then re-published we don't hit the "missing row = insert new relation" logic

```
diff --git a/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts b/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
index 92c4f8ee2a..9d5cbf3db6 100644
--- a/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
@@ -157,6 +157,9 @@ const sync = async (
     {} as Record<string, string>
   );
 
+  const republishedEntryIds = new Set(newEntries.map((e) => String(e.id)));
+  const isRepublishedEntry = (id: string | number) => republishedEntryIds.has(String(id));
+
   await strapi.db.transaction(async ({ trx }) => {
     for (const { joinTable, relations } of existingRelations) {
       const sourceColumn = joinTable.inverseJoinColumn.name;
@@ -197,9 +200,9 @@ const sync = async (
             .where(sourceColumn, newSourceId)
             .where(targetColumn, targetId)
             .update({ [orderColumn]: originalOrder });
-        } else {
-          // Row was cascade-deleted when the old published entry was removed.
-          // Re-insert it with the new entity ID.
+        } else if (!isRepublishedEntry(newSourceId)) {
+          // Missing row was cascade-deleted (we republished the target) — re-insert.
+          // If we republished the source, the row was never created (user removed the relation).
           const newRelation = {
             ...omit(strapi.db.metadata.identifiers.ID_COLUMN, relation),
             [sourceColumn]: newSourceId,


```

innerdvations:
Overall this looks correct, but you should be able to get much better performance ( `O(1)` instead of `O(n)` ) by changing sync to batch per join table instead of per row. I pushed an AI-written example to `fix/update-api-response-perf-refactor` so you can see exactly what I mean, although beware using the code directly 😆

Issue #25724, opened 2026-03-12
Duplicate history versions created on publish due to trailing ? in request URL breaking deduplication guard


bolg55:
### Node Version

20

### NPM/Yarn/PNPM Version

npm

### Strapi Version

5.33.2

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When publishing a document from the Content Manager, two identical history versions are created - both at the same timestamp.

### Steps to Reproduce

1. Create any collection type entry with Draft & Publish enabled
2. Save the draft
3. Click Publish
4. Open Content History for the document
5. Observe: Two history entries exist, created at the same instant

### Expected Behavior

One history version created per publish action (the published version).

### Actual Behavior
Two history versions created:

<img width="425" height="530" alt="Image" src="https://github.com/user-attachments/assets/e2217ce1-1c1e-4cb4-b013-4df41b4507c0" />

### Logs

```shell

```

### Code Snippets

The deduplication guard in **lifecycles.ts** fails due to a trailing ? in the request URL.

**packages/core/content-manager/server/src/history/services/lifecycles.ts**

  ```
// Intended to suppress the internal `update` triggered during publish
  if (
    context.action === 'update' &&
    strapi.requestContext.get()?.request.url.endsWith('/actions/publish') // ← FAILS
  ) {
    return false;
  }
```
When the admin UI calls publish, the fetch client (getFetchClient.ts) serializes params into the URL. The params value passed is {} (empty object from buildValidParams({})). An empty object is truthy, so the serializer enters its branch, calls `qs.stringify({})` which returns "", and appends ? - producing /actions/publish?.

**packages/core/admin/admin/src/utils/getFetchClient.ts**

  ```
const paramsSerializer = <Param = unknown>(params?: Param) => (url: string) => {
    if (params) {                                    // {} is truthy ← root cause
      const serializedParams = qs.stringify(params, { encode: false });
      return `${url}?${serializedParams}`;           // → "/actions/publish?"
    }
    return url;
  };
```
**packages/core/content-manager/admin/src/hooks/useDocumentContext.ts**

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #25735, opened 2026-03-15
fix(admin): prevent trailing ? in URL when params is empty object

mahmoodhamdi:
## What does this PR do?

Fixes the `paramsSerializer` in `getFetchClient` to skip appending `?` when `qs.stringify` produces an empty string from an empty params object (`{}`).

**Root cause:** When `params` is `{}` (e.g., from `buildValidParams({})`), the truthiness check `if (params)` passes since `{}` is truthy. Then `qs.stringify({})` returns `""`, and the URL gets a trailing `?` appended (e.g., `/actions/publish?`).

**Impact:** This breaks the history version deduplication guard in `lifecycles.ts`, which uses `endsWith('/actions/publish')` — the trailing `?` causes the check to fail, resulting in duplicate history versions on every publish action.

**Fix:** After calling `qs.stringify`, check that the serialized string is non-empty before appending the query separator.

## Why is this change needed?

Every time a document is published from the Content Manager, two identical history versions are created at the same timestamp instead of one. This happens because the deduplication guard that should suppress the internal `update` during a publish request fails to match the URL due to the trailing `?`.

Issue: #25724

## How was this tested?

- Added unit tests verifying that empty params `{}` and `undefined` params do not append a trailing `?` to the URL
- Verified all existing `getFetchClient` tests continue to pass (11/11)
- Verified the modified files have no lint errors

## Checklist
- [x] Tests added/updated
- [x] All tests pass

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -340,7 +340,9 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
          * It's considered a breaking change because it impacts any API request, including the user's custom code
          */
         const serializedParams = qs.stringify(params, { encode: false });
-        return `${url}?${serializedParams}`;
+        if (serializedParams) {
+          return `${url}?${serializedParams}`;
+        }
       }
       return url;
     };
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:

Issue #25493, opened 2026-02-19
strapi openapi generate produces urls with missing /api/ prefix


devonik:
### Node Version

22.20

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.36.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

https://docs.strapi.io/cms/api/openapi#cli-usage

### Bug Description

When I execute the command `npm run strapi openapi generate -- --output ./docs/api-spec.json` the path's in the json are e.g. /articles. This won't work cause the actual route is /api/articles

### Steps to Reproduce

See strapi docu https://docs.strapi.io/cms/api/openapi#cli-usage

### Expected Behavior

Path's in the json containing the full relative path of the entpoint e.g. /api/articles

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #25737, opened 2026-03-15
fix(openapi): add REST API prefix to content-api routes in generated spec

mahmoodhamdi:
## What does this PR do?

Adds the configured REST API prefix (`api.rest.prefix`, default `/api`) to content-api routes in the OpenAPI spec generator's `ApiRoutesProvider`.

## Why is this change needed?

When running `strapi openapi generate`, the generated spec produces paths like `/articles` instead of `/api/articles`. Since the REST API prefix is applied at the Koa Router level, routes stored in `strapi.apis` do not include it. The OpenAPI generator reads these routes directly, so the generated paths are missing the prefix and don't match the actual API endpoints.

This follows the same pattern used by the core content-api service when registering routes:

```ts
const apiPrefix = strapi.config.get('api.rest.prefix');
routes.map((route) => ({ ...route, path: `${apiPrefix}${route.path}` }));
```

Issue: #25493

## Changes

- **`ApiRoutesProvider`**: Apply `api.rest.prefix` to content-api routes (admin routes are left unchanged)
- **Test fixtures**: Updated to reflect that routes in `strapi.apis` don't include the prefix
- **`StrapiMock`**: Added `config.get` method
- **Tests**: Added 2 new tests for prefix application and custom prefix support

## How was this tested?

- All 22 existing + new OpenAPI tests pass
- Verified the prefix is only applied to `content-api` type routes
- Verified custom prefix support via config

## Checklist
- [x] Tests added/updated
- [x] All tests pass

### DIFF
--- a/packages/core/openapi/__tests__/fixtures/routes.ts
+++ b/packages/core/openapi/__tests__/fixtures/routes.ts
@@ -1,13 +1,13 @@
 import { Core } from '@strapi/types';
 
 export const test = [
-  { info: { type: 'content-api' }, method: 'GET', path: '/api/test1', handler: '' },
+  { info: { type: 'content-api' }, method: 'GET', path: '/test1', handler: '' },
   { info: { type: 'admin' }, method: 'POST', path: '/test2', handler: '' },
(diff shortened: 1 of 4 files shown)


mahmoodhamdi:

Issue #25697, opened 2026-03-09
Strapi v5 migration 5.0.0-06-add-document-id-indexes fails when index already exists or when index name exceeds PostgreSQL identifier limit


vojtys:
### Node Version

v20.19.4

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.38.0

### Operating System

Other

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

The internal migration 5.0.0-06-add-document-id-indexes can fail in PostgreSQL under two conditions:
	1.	When the index already exists but knex.schema.hasIndex() does not detect it.
	2.	When the generated index name exceeds PostgreSQL’s 63 character identifier limit and PostgreSQL truncates the name.

Both cases result in a 42P07 relation already exists error which aborts the transaction and causes Strapi to fail during startup.

### Steps to reproduce

1. Create a Strapi project using PostgreSQL.

2. Before upgrading to a version containing the migration
   `5.0.0-06-add-document-id-indexes`, manually create the index
   expected by the migration:

```sql
CREATE UNIQUE INDEX events_document_id_idx
ON events(document_id);
```

3. Start Strapi so the internal migration runs.

4. During startup, the migration attempts to create the same index again:

```sql
CREATE INDEX "events_document_id_idx" ON "events" ("document_id");
```

5. PostgreSQL returns:

```
ERROR: relation "events_document_id_idx" already exists
errorCode: 42P07
```

### Result

The migration fails and aborts the transaction, causing Strapi to fail during startup.

### Expected result

The migration should detect that the index already exists and skip creation.

---

### Steps to reproduce

1. Use PostgreSQL with Strapi.

2. Ensure a table exists with a long name, for example:

```
strapi_transfer_token_permissions
```

3. The migration generates the index name:

```
strapi_transfer_token_permissions_document_id_locale_published_at_idx
```

4. This name has **69 characters**, exceeding PostgreSQL's identifier limit of **63 characters**.

5. PostgreSQL truncates the identifier and stores the index as:

```
strapi_transfer_token_permissions_document_id_locale_published_
```

6. When the migration runs again, it checks for the original full name and does not detect the truncated index.

7. The migration attempts to create the index again:

```sql
CREATE INDEX "strapi_transfer_token_permissions_document_id_locale_published_at_idx"
ON "strapi_transfer_token_permissions" ("document_id", "locale", "published_at");
```

8. PostgreSQL returns:

```
[...]

PR #25738, opened 2026-03-15
fix(database): prevent migration failure from long index names and transaction abort

mahmoodhamdi:
## What does this PR do?

Fixes the `5.0.0-06-add-document-id-indexes` migration to handle two PostgreSQL-specific failure modes that prevent Strapi from starting.

## Why is this change needed?

The migration fails in PostgreSQL under these conditions:

### 1. Index name exceeds 63-character identifier limit

For tables with long names (e.g., `strapi_transfer_token_permissions`), the generated composite index name exceeds PostgreSQL's 63-character limit:

```
strapi_transfer_token_permissions_document_id_locale_published_at_idx
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            69 characters
```

PostgreSQL silently truncates to 63 chars on first creation. On subsequent runs, `hasIndex()` checks for the full 69-char name, doesn't find it, and tries to create it again — failing with `42P07 relation already exists`.

### 2. Transaction abort cascades to all subsequent operations

When a `CREATE INDEX` fails inside a PostgreSQL transaction, the transaction enters an aborted state. Every subsequent SQL statement fails with:

```
current transaction is aborted, commands ignored until end of transaction block
```

This means a single duplicate-index error cascades and breaks all remaining table migrations.

## Changes

- **Truncate index names** to PostgreSQL's 63-character identifier limit before creation
- **Wrap each index creation in a SAVEPOINT** so that a failure can be rolled back without poisoning the surrounding transaction
- **Remove the unreliable `hasIndex()` check** — the SAVEPOINT + catch approach handles duplicates reliably across all databases

## How was this tested?

- All 4 tests pass (1 existing + 3 new):
  - Index creation with all columns present
  - Index name truncation for long table names
  - SAVEPOINT usage verification
  - ROLLBACK TO SAVEPOINT on duplicate index error

Issue: #25697

## Checklist
- [x] Tests added/updated
- [x] All tests pass

### DIFF
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-06-add-document-id-indexes.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-06-add-document-id-indexes.ts
@@ -2,27 +2,39 @@ import type { Knex } from 'knex';
 
 import type { Migration } from '../common';
 
+/**
+ * PostgreSQL silently truncates identifiers longer than 63 characters.
+ * When the migration later checks for the full-length name, it does not
+ * find the truncated version, and the CREATE INDEX fails with 42P07.
+ */
+const PG_IDENTIFIER_MAX_LENGTH = 63;
+
 // Add an index if it does not already exist
 const createIndex = async (knex: Knex, tableName: string, columns: string[], indexName: string) => {
+  // Truncate to PostgreSQL's identifier limit to avoid silent name mismatch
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:

Issue #21967, opened 2024-10-24
REST API and Content Manager returning different IDs for entries


MunasheHodza:
### Required System information

- Node.js version: Node 20
- NPM version: npm/10.5.2
- Strapi version: 5.1.0
- Database: Strapi Cloud
- Operating system: darwin arm64
- Is your project Javascript or Typescript: Javascript

### Describe the bug

The Content Manager in the admin panel shows id numbers in odd numbers and the rest-api shows them as even numbers for the same entries!

### Steps to reproduce the behavior

1. Go to content manager and also make a rest api call
2. See error

### Expected behavior

IDs should match

### Screenshots

<img width="1140" alt="Screenshot 2024-10-24 at 12 15 02" src="https://github.com/user-attachments/assets/9c740a00-f343-4b2b-9d27-f9c823192116">
<img width="977" alt="Screenshot 2024-10-24 at 12 13 12" src="https://github.com/user-attachments/assets/dfc90a49-8bb0-4c55-81c0-9b481d273580">

DarshanBorse:
Please assign me 👍

derrickmehaffy:
> Please assign me 👍

We don't assign issues, just open a PR

kasyap1234:
I have verified thiss issue, getting two different sets of IDs : 
one via content manager UI and one via REST API . 
content manager UI and core strapi 
The key difference is : 
uid.ts focuses on generating and validating UIDs for user input fields
Strapi.ts uses UIDs as system-level identifiers for content structure

kasyap1234:
so using a single service for both api and content-manager should solve this issue

jordisanchezcarbonell:
Hi, i have the same error , how can fix it ?

derrickmehaffy:
This isn't really "intended" but the integer based IDs are useless in this case and we really should be showing the document ID instead in the content-manager which our product team is already aware of and are planning on working on this.

cc @yanniskadiri / @lucasboilly 

This is an internal video from one of our product designers about the initial design of this: 

https://www.loom.com/share/a9b87246cff84f5cb906a3725bf75e7b?sid=111510e4-b64a-40e0-b13c-6de2587becb1

TruoYang:
I know I'm late.. but could you tell me is the bug solved? I got the same problem but it seems doesn't follow any rule, just two number randomly but never equal

eevan7a9:
it's July 25, 2025. I still encounter the same issue has anyone found a solution for this?

[...]

PR #25759, opened 2026-03-18
enhancement(content-manager): add documentid in listview and editview

mathildeleg:
### What does it do?
In the List view: 
* Display a new "documentId" column
* Remove the "id" column 
* Add a Copy icon to allow copying to clipboard

In the Edit view:
* Display a new "Document ID" at the bottom of the Actions menu if the document has been created (in at least one locale when i18n is enabled) and the content-type is a collection-type (_disabled for single-types as there's no need in the API to know the document ID of a single-type_)
* Add a new "Copy the document ID" in the Actions menu to allow copying to clipboard

### Why is it needed?
Having the "id" column brought confusion to users as it is now the document ID that's used to fetch data from the API. 

### How to test it?
**By default, in the list view, the column "Document ID" might not be visible, you may have to display it through the Configure the view**

_Single types_
* Go to a single type
* Create an entry (fill in one field and save)
* Click on the "More actions" button on the top right corner
=> it should stay as before, ie.:
Capture d’écran 2026-03-17 à 17 08 42

_Collection types - New entry_
* Go to a collection type (with i18n enabled)
* Create an entry (do not fill in a field and save yet)
* Click on the "More actions" button on the top right corner
=> the Document ID shouldn't be visible in this menu as the entry hasn't yet been created, so the menu should look like this:
Capture d’écran 2026-03-17 à 17 09 12

_Collection types - Existing entry_
* Create that entry (fill in one field and save)
* Click on the "More actions" button on the top right corner
=> Now the menu should show the "Copy document ID" button + the "Document ID" value at the bottom like so:
Capture d’écran 2026-03-17 à 17 12 34
* Click on the "Copy document ID" => there should be a success notification and ID should be copied to your clipboard
* Go back to the List view => there should be a new column "DocumentID"
Capture d’écran 2026-03-17 à 17 13 17
* Click on the Copy icon => there should be a success notification and ID should be copied to your clipboard

_Collection types - New locale (from existing entry)_
* Click on your entry
* Select "Create ... locale" to create an entry in a new locale
=> The menu should still show the document ID and the "Copy document ID" button since it exists in at least one locale
Capture d’écran 2026-03-17 à 17 15 06

### Related issue(s)/PR(s)
Fixes https://github.com/strapi/strapi/issues/21967 
Fixes CMS-497
according to product needs https://www.loom.com/share/a9b87246cff84f5cb906a3725bf75e7b?sid=111510e4-b64a-40e0-b13c-6de2587becb1

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
@@ -296,15 +296,17 @@ const CopyDocumentIdMenuItem = ({ documentId }: CopyDocumentIdMenuItemProps) => 
     return null;
   }
 
-  const handleCopy = () => {
-    copy(documentId);
-    toggleNotification({
-      type: 'success',
-      message: formatMessage({
-        id: 'content-manager.actions.copy-documentId.success',
-        defaultMessage: 'Document ID copied to clipboard',
-      }),
-    });
+  const handleCopy = async () => {
+    const didCopy = await copy(documentId);
+    if (didCopy) {
+      toggleNotification({
+        type: 'success',
+        message: formatMessage({
+          id: 'content-manager.actions.copy-documentId.success',
+          defaultMessage: 'Document ID copied to clipboard',
+        }),
+      });
+    }
   };
 
   return (
--- a/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
@@ -82,17 +82,19 @@ const ListViewPage = () => {
   const isDesktop = useIsDesktop();
 
   const handleCopyDocumentId = React.useCallback(
-    (e: React.MouseEvent, documentId: string | undefined) => {
+    async (e: React.MouseEvent, documentId: string | undefined) => {
       e.stopPropagation();
       if (!documentId) return;
-      copy(documentId);
-      toggleNotification({
-        type: 'success',
-        message: formatMessage({
-          id: 'content-manager.actions.copy-documentId.success',
-          defaultMessage: 'Document ID copied to clipboard',
-        }),
-      });
+      const didCopy = await copy(documentId);
+      if (didCopy) {
+        toggleNotification({
+          type: 'success',
+          message: formatMessage({
+            id: 'content-manager.actions.copy-documentId.success',
+            defaultMessage: 'Document ID copied to clipboard',
+          }),
+        });
+      }
     },
     [copy, formatMessage, toggleNotification]
   );


packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
```diff
@@ -423,6 +442,30 @@ const ListViewPage = () => {
...
+                                      <IconButton
+                                        variant="ghost"
+                                        size="S"
+                                        label={formatMessage({
+                                          id: 'content-manager.ListView.copy-documentId.label',
+                                          defaultMessage: 'Copy',
+                                        })}
+                                        onClick={(e) => handleCopyDocumentId(e, row.documentId)}
+                                      >
+                                        <Duplicate />
```
markkaylor, line 464: Don't we use duplicate for the clone feature? Could that be confusing?

packages/core/content-manager/admin/src/translations/en.json
```diff
@@ -9,13 +9,16 @@
...
   "actions.edit.label": "Edit",
+  "actions.copy-documentId.label": "Copy document ID",
   "actions.unpublish.error": "An error occurred while trying to unpublish the document.",
   "actions.unpublish.dialog.body": "Are you sure you want to unpublish this?",
   "actions.unpublish.dialog.option.keep-draft": "Unpublish and keep last draft",
   "actions.unpublish.dialog.option.replace-draft": "Unpublish and replace last draft",
   "ListViewTable.relation-loaded": "Relations have been loaded",
   "ListViewTable.relation-loading": "Relations are loading",
   "ListViewTable.relation-more": "This relation contains more entities than displayed",
+  "ListView.copy-documentId.label": "Copy document ID",
```
markkaylor: Couldn't we remove this and just use: `actions.copy-documentId.label` everywhere?

packages/core/content-manager/admin/src/translations/en.json
```diff
@@ -9,13 +9,16 @@
...
+  "ListView.copy-documentId.success": "Document ID copied to clipboard",
```
markkaylor: I assume this is used in the EditView as well when copying right?

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
mathildeleg: Thanks for pointing it out, I hadn't realized it was the same icon. It is the icon that was used by product designers for this enhancement, but maybe we could swap it to the Files icon (which is the only other icon in the design system that I think could work for this), wdyt? I can also ask Anka if you want
Capture d’écran 2026-03-19 à 14 12 50
Capture d’écran 2026-03-19 à 14 15 52

packages/core/content-manager/admin/src/translations/en.json
mathildeleg: Yes I ended up using the `actions.copy-documentId.label` everywhere instead ([here](https://github.com/strapi/strapi/pull/25759/commits/74a0403e0741177d98a7d76c4f2d5e5d080f9d54)) :)

packages/core/content-manager/admin/src/pages/EditView/components/Header.tsx
```diff
@@ -274,15 +279,60 @@ const HeaderToolbar = ({ activeTab, isCloning }: HeaderDocumentActionsProps) =>
...
+const CopyDocumentIdMenuItem = ({ documentId }: CopyDocumentIdMenuItemProps) => {
+  const { formatMessage } = useIntl();
+  const { toggleNotification } = useNotification();
+  const { copy } = useClipboard();
+
+  if (!documentId) {
+    return null;
+  }
+
+  const handleCopy = () => {
```
jhoward1994: I think this should maybe look like this ?
copy() is async and returns false on failure

```
  const handleCopy = async () => {
    const didCopy = await copy(documentId);
    if (didCopy) {
      toggleNotification({
        type: 'success',
        message: formatMessage({
          id: 'content-manager.actions.copy-documentId.success',
          defaultMessage: 'Document ID copied to clipboard',
        }),
      });
    }

```

And looks like the same in packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx L88

mathildeleg: Very good catch thank you 🙈 
Fixed [Here](https://github.com/strapi/strapi/pull/25759/commits/7ae9e8ec37930d63881bf52c5a01d4ba052543cc)

packages/core/content-manager/admin/src/pages/ListView/ListViewPage.tsx
markkaylor: Thanks for suggesting a new icon. That could work, but looking at it now, I guess duplicate is inside the 3 dots menu and not directly on the row so maybe I'm over thinking it..? Feel free to check with Anka but if that's the icon she put in the designs I'm sure she considered it and doesn't see it as a problem.

mathildeleg: Lucas had made the design but I checked with Anka and she's good going with this, if we get feedback from users we can adapt then :)

Issue #25746, opened 2026-03-16
since 5.37 publicationStatus is gone on graphql queries which are non-DP, even though you need publicationStatus for nested DP relations


laurenskling:
### Node Version

20

### Package Manager

yarn

### Package Manager Version

1.22

### Strapi Version

5.37.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

Since 5.37.0 Users Permissions plugin graphql queries do not longer support the `status` field. 

While UP Users are not DP, I need this to get the correct versions for relations I've extended onto the User.

In fact, all non-DP contentTypes no longer have the `status` arg, by which I am now never able to query `draft` relations on a non-DP entry.

### Steps to Reproduce

1. install 5.37.0
2. checkout out the graphql playground

### Expected Behavior

I want `status` to be available as a arg, when either top-level is DP, or relations are DP.
or.. I want to be able to arg `status` on my relations.

### Media

5.36.1:
<img width="487" height="310" alt="Image" src="https://github.com/user-attachments/assets/571cb323-e9cc-406b-a920-52561beeafe0" />

5.37.0:
<img width="482" height="271" alt="Image" src="https://github.com/user-attachments/assets/c827ebbd-3adf-4ffa-9526-020d78e239e4" />

### Additional information

Probably related to https://github.com/strapi/strapi/pull/25528

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

laurenskling:
> you should query those relations directly using their own API endpoints with `status` filtering, rather than through the User query.

This is not possible in Graphql. There is only a toplevel `status`, it's never available on relations. And if I want to get the relations (lets say, Company) for a specific User, I cannot toplevel top level Company because I would need to know about the User right?

I do not want custom resolvers, because I just realized this goes for _every_ non-DP contentType. This just breaks nested relations DP in Graphql, period. It should be fixed, not worked around.

laurenskling:
Fixed by https://github.com/strapi/strapi/pull/25763

PR #25763, opened 2026-03-18
fix(graphql): expose status and hasPublishedVersion on non–D&P root queries for nested relations

innerdvations:
### What does it do?

Restores `status` and `hasPublishedVersion` on GraphQL root queries for **all** collection and single-type content types (including non–draft & publish types such as Users & Permissions `User`).
Previously (#25292) these arguments were only added when the content type had `draftAndPublish: true`, which removed them from non–D&P roots and broke clients that relied on them.

### Why is it needed?

On non–D&P parents, these arguments do not select draft/published **versions of the parent**, but they are still **valid and necessary**: the content API association resolver inherits them from the root query (`rootQueryArgs`) to apply draft vs published (and `hasPublishedVersion`) filters when loading **nested relations** whose **target** is draft & publish (e.g. User → related Articles). Without root-level `status` / `hasPublishedVersion`, nested D&P relations could not be queried as draft from a non–D&P entry ([#25746](https://github.com/strapi/strapi/issues/25746)).
A code comment documents a possible future direction (per-relation args on D&P targets instead of root inheritance).

### How to test it?

1. Enable the GraphQL plugin and open the GraphQL playground (or any GraphQL client).
2. Use a **non–D&P** root query (e.g. `usersPermissionsUsers` or another collection without D&P) and confirm the schema exposes `status` and `hasPublishedVersion` on that field.
3. Add a relation from that type to a **D&P** collection; run a query that loads that relation with `status: draft` vs `status: published` on the **root** field and assert nested results match (draft rows only when `draft`, published when `published`).
4. Optionally repeat with `hasPublishedVersion` if your setup uses it on nested D&P data.
5. Smoke-test a **D&P** root query (e.g. articles): behavior should remain unchanged.

### Related issue(s)/PR(s)

- https://github.com/strapi/strapi/issues/25746

### DIFF
--- a/packages/plugins/graphql/server/src/services/builders/utils.ts
+++ b/packages/plugins/graphql/server/src/services/builders/utils.ts
@@ -36,8 +36,19 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
 
       const { kind } = contentType;
 
-      // `status` / `hasPublishedVersion` on every root collection & single-type query.
-      // Document service ignores or applies them by content type (D&P vs not).
+      // On non–D&P roots (e.g. User) these args do not version the parent document, but they
+      // are required: association resolvers inherit them into nested D&P relations (see
+      // builders/resolvers/association.ts + rootQueryArgs). Omitting them broke draft/published
+      // control for populated relations (e.g. github.com/strapi/strapi/issues/25746).
+      //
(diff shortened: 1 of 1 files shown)


laurenskling:
Thanks for this

pwizla:
Hey @innerdvations I documented it for you on https://github.com/strapi/documentation/pull/3026/changes, please let me know if this makes sense 😊

PR #25793, opened 2026-03-20
feat: publicationFilter param in REST and document service

innerdvations:
### What does it do?

- Introduces an optional **`publicationFilter`** query parameter (kebab-case strings on REST and the document service; GraphQL exposes the same cohorts via a `PublicationFilter`-style enum) with **pair-scoped** modes (`never-published`, `has-published-version`, `modified`, `unmodified`), **document-scoped** modes (`never-published-document`, `has-published-version-document`), and **published-slice** modes (`published-without-draft`, `published-with-draft`).
- Implements cohort selection with **`buildPublicationFilterWhere`** in `@strapi/utils` (SQL `id IN (…)` subqueries scoped to draft/publish rows and, when i18n applies, `(documentId, locale)` or document-only semantics).
- Wires **`publicationFilter`** through the **Content API** (allowlist, validation, sanitize) and **document service** query transformation so filters combine correctly with existing **`status`** behavior.
- Keeps backward compatibility for deprecated **`hasPublishedVersion`**: it is still accepted, parsed (boolean / `"true"` / `"false"`), and mapped to **`never-published-document`** / **`has-published-version-document`** via **`hasPublishedVersionBooleanToPublicationFilterMode`**.
- Ensures publication filtering applies to **nested populate** by implementing it through the query builder’s **`filters` callback** (per `meta.uid`), so related entities get the same cohort logic as the root—matching the pre-refactor `hasPublishedVersion` behavior and fixing populate cascade regressions.
- Updates **GraphQL** (`mergePublicationFilterFromGraphQLArgs`, root resolvers, association loading) and **Content Manager** collection-type flows where relevant.
- Adds/extends **API and unit tests** (including `publication-filter.test.api.ts`, `has-published-version.test.api.ts`, GraphQL publication-filter tests, and utils tests) and documents the design in **`docs/docs/rfcs/02-publication-filter-query-modes.md`**.

### Why is it needed?

Draft & publish already exposes **`status`** and row-level **`publishedAt`**, but many useful queries are **derived across draft and published rows** (e.g. “never published for this locale”, “draft differs from published”, “published row with no draft peer”). Those cannot be expressed as a simple filter on a single column.

This PR adds an explicit, stable **`publicationFilter`** surface for those cohorts while **preserving existing defaults and `status` semantics**, and keeps **`hasPublishedVersion`** working for older clients by mapping it to the document-scoped modes, but treats it as deprecated.

### How to test it?

- **Environment:** Strapi dev/test app with draft & publish and i18n as in the API test fixtures (see `tests/api/...` helpers); database as configured for **`yarn test:api`** (see `tests/scripts/run-api-tests.js` / `jest.config.api.js`).
- **Automated checks:**
  - API: `yarn test:api` (or Jest with `jest.config.api.js`), with emphasis on:
    - `tests/api/core/strapi/document-service/publication-filter.test.api.ts`
    - `tests/api/core/strapi/document-service/has-published-version.test.api.ts`
    - `tests/api/core/strapi/document-service/validation/validation.test.api.ts`
    - `tests/api/plugins/graphql/publication-filter.test.api.js`
  - Unit: `@strapi/utils` tests under `packages/core/utils/src/__tests__/` (e.g. `publication-filter.test.ts`, `has-published-version-param.test.ts`, validate/sanitize query tests).
- **Manual spot-check (optional):** REST `GET` with `status`, `publicationFilter`, and `populate` on a D&P + localized type; GraphQL queries using `publicationFilter` and deprecated `hasPublishedVersion` for parity.

### Related issue(s)/PR(s)

- Design: RFC **`docs/docs/rfcs/02-publication-filter-query-modes.md`**
- Fixes CMS-526


packages/core/types/src/modules/documents/params/status.ts
```diff
@@ -1,3 +1,19 @@
 export type Kind = 'draft' | 'published';
 
-export type Param = { status?: Kind; hasPublishedVersion?: boolean };
```
jhoward1994, line 3: Instead can we do?

```
import type { PublicationFilterMode } from '@strapi/utils';
...

export type { PublicationFilterMode };

```

packages/core/utils/src/convert-query-params.ts
```diff
@@ -80,9 +80,18 @@ export interface Params {
   page?: number | string;
   pageSize?: number | string;
   status?: 'draft' | 'published';
+  publicationFilter?:
```
jhoward1994: Would this work?
```suggestion
import type { PublicationFilterMode } from './publication-filter';

  publicationFilter?: PublicationFilterMode;
```

packages/core/utils/src/publication-filter.ts
```diff
@@ -0,0 +1,308 @@
...
+        ...(prev && typeof prev === 'object' ? prev : {}),
+        source: 'query',
+        param: 'publicationFilter',
+      };
+    }
+    throw e;
+  }
+};
+
+const columnName = (meta: any, attr: string): string => {
```
jhoward1994, line 66: Would it be worth throwing some specific errors here ? 

```
const columnName = (meta: any, attr: string): string => {
  const a = meta?.attributes?.[attr];
  if (!a) {
    throw new ApplicationError(
      `Cannot build publicationFilter SQL: attribute '${attr}' is missing from metadata for table '${meta?.tableName ?? '<unknown>'}'.`
    );
  }
  if (!('columnName' in a) || !a.columnName) {
    throw new ApplicationError(
      `Cannot build publicationFilter SQL: attribute '${attr}' on table '${meta.tableName}' has no resolved columnName.`
    );
  }
  return a.columnName as string;
};

```

jhoward1994 (dismissed):
Looks good I just had one other comment

packages/core/utils/src/publication-filter.ts
innerdvations: Good idea, just pushed it, can you give it a look?

PR #25801, opened 2026-03-21
fix(content-type-builder): show configure the view outside development

### LINKED ISSUE #25783: Content-Type Builder hides "Configure the view" outside development mode in Strapi 5
### Node Version

v20.19.5

### Package Manager

yarn

### Package Manager Version

1.22.19

### Strapi Version

v5.38.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

### Steps to Reproduce

### Expected Behavior

### Logs

```shell

```

### Code Snippets

```js
```

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

nehaaprasad:
### What does it do?

- Shows Configure the view outside development mode.
- Keeps Edit and Add another field development-only.

### Why is it needed?

- Configure the view is a settings shortcut, not a schema edit action.

### How to test it?

- Start with strapi start.
- Open /admin/plugins/content-type-builder/content-types/api::page.page.
- Check that Configure the view is visible, while Edit and Add another field are hidden.

### Related issue(s)/PR(s)

Fix #25783

### DIFF
--- a/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
+++ b/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
@@ -116,36 +116,39 @@ const ListView = () => {
 
   const isDeleted = type.status === 'REMOVED';
 
-  const primaryAction = isInDevelopmentMode && (
+  const primaryAction = (
     <Flex gap={2}>
       <LinkToCMSettingsView
         key="link-to-cm-settings-view"
         type={type}
         disabled={type.status === 'NEW' || isDeleted}
       />
-      <Button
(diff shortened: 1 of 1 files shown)


themavik (commented):
Confirm LinkToCMSettingsView is intentionally usable outside development (not just hidden for UX); if it ever relied on dev-only APIs this widens the surface.

innerdvations:
Thanks for opening this and for being first on #25783. #26618 has the same ListView change plus regression tests, so I'm closing this as a duplicate in favor of that PR.

_AI-assisted comment._

Issue #25795, opened 2026-03-20
REST API $contains and $containsi filter operators return empty data


gottfried-github:
### Node Version

24.12.0

### Package Manager

npm

### Package Manager Version

11.6.2

### Strapi Version

5.40.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

https://github.com/gottfried-github/strapi-contains-issue

### Bug Description

`$contains` and `$containsi` operators in REST API return empty data. With Document Service API, `$contains` works as expected (see reproduction repo). 

I have both Draft & Publish and Internationalzation disabled.

I also tried to run REST API queries with SQLite and got the same result.

I also tried to run the following query using `$eq` operator, which does work:

`curl -g localhost:1337/api/performances?filters[title][$eq]=test02`

### Steps to Reproduce

1. import data from the backup in the reproduction repo (it contains 2 documents with the following `title`: `test01`, `test02`)
2. run the queries `curl -g localhost:1337/api/performances?filters[title][$contains]=est02`, `curl -g localhost:1337/api/performances?filters[title][$containsi]=est02`
3. observe that empty `data` array is returned

### Expected Behavior

Should return a single document with `title` of `test02`

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #25810, opened 2026-03-23
fix: contains filter no longer returns empty data

innerdvations:
### What does it do?

traverse scalar fields when the value is an operator object so filter handling matches nested structures used by the REST API

### Why is it needed?

Fixes REST filtering with **`$contains` / `$containsi`** on scalar fields sometimes returning empty datasets when matches should exist ([#25795](https://github.com/strapi/strapi/issues/25795)) because generated queries were incorrect

### How to test it?

```bash
yarn build
yarn test:unit packages/core/utils/src/__tests__/sanitize.visitors.test.ts
yarn test:unit packages/core/utils/src/__tests__/validate.visitors.test.ts
yarn test:api tests/api/core/strapi/filtering.test.api.js
```

### Related issue(s)/PR(s)

Fixes [#25795](https://github.com/strapi/strapi/issues/25795)

Fixes CMS-527

### DIFF
--- a/packages/core/utils/src/__tests__/sanitize.visitors.test.ts
+++ b/packages/core/utils/src/__tests__/sanitize.visitors.test.ts
@@ -153,6 +153,23 @@ describe('Sanitize visitors util', () => {
       await expect(sanitizers.defaultSanitizeFilters(ctx, filters)).resolves.toEqual(filters);
     });
 
+    test('keeps scalar datetime $gt filter with Date operand (GraphQL)', async () => {
+      const schemaWithDatetime: typeof articleModel = {
+        ...articleModel,
+        attributes: {
+          ...articleModel.attributes,
+          publishedAt: { type: 'datetime' },
+        },
+      };
+      const ctxDatetime = { schema: schemaWithDatetime, getModel };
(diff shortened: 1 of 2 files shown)


innerdvations:
@dosu exactly one source label is allowed on a PR, when you add a second label it breaks the workflow

packages/core/utils/src/traverse/query-filters.ts
```diff
@@ -58,6 +65,18 @@ const filters = traverseFactory()
     async ({ key, visitor, path, value, schema, getModel, attribute }, { set, recurse }) => {
       const parent: Parent = { key, path, schema, attribute };
 
+      // $null / $notNull operands are boolean-cast downstream; arbitrary objects are not filter maps.
+      // Without this, traversing into e.g. { $null: { anything: 'x' } } makes validate throw on "anything".
+      if (
```
Bassel17, line 73: There are failing GraphQL datetime API tests after this change.  now that scalar filters recurse into operator maps, values like GraphQL DateTime operands ($gt: Date) also flow through here. I think the guard should be generalized from just $null/$notNull to any operator operand that is an object but not actually filter-like

innerdvations:
I fixed it so that operands work properly now

PR #25822, opened 2026-03-24
fix: prevent transfer failure when uploads directory is mounted

### LINKED ISSUE #25821: Data transfer fails when public/uploads is a mounted directory (EBUSY on move)
### Node Version

24.13.0

### Package Manager

yarn

### Package Manager Version

4.12.0

### Strapi Version

5.40.0

### Operating System

Docker/Podman/LXC

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When running `strapi transfer` with the local upload provider, the restore process fails if `public/uploads` is a mounted directory (e.g. a Docker bind mount or Kubernetes volume).

The error shown is:

```
The backup folder for the assets could not be created inside the public folder. Please ensure Strapi has write permissions on the public directory
```

The error message suggests a permissions issue, but the root cause is the inability to move a mounted directory.

This occurs because the backup step attempts to move the uploads directory using `fs-extra.move()`, which fails when the directory is mounted.

### Steps to Reproduce

1. Create and run two Strapi instances:
   - one source instance
   - one destination instance running in Docker with `public/uploads` mounted as a volume
     - e.g. `-v /host/uploads:/app/public/uploads`

2. On the source instance:
   - create some content
   - upload at least one file
   - create a transfer token

3. From the destination instance, run:
   `strapi transfer` (pull from the source instance)

Result: the transfer fails during the assets backup step

### Expected Behavior

The transfer process should succeed and create a backup of the uploads directory.

Mounted directories should be handled correctly, without failing the restore process.

### Logs

```shell
rename("/home/userapp/public/uploads", "/home/userapp/public/uploads_backup_<timestamp>") = -1 EBUSY (Device or resource busy)

[FATAL] restore failed Error: The backup folder for the assets could not be created inside the public folder. Please ensure Strapi has write permissions on the public directory
```

### Code Snippets

The issue occurs in the assets backup step where `fs-extra.move()` is used:

```
await fse.move(assetsDirectory, backupDirectory);
```

This fails when `assetsDirectory` is a mounted directory.

### Media

_No response_

### Additional information

The issue is caused by `fs-extra.move()` relying on `rename`, which cannot move mounted directories or across filesystems.

This results in errors such as:
- EBUSY (device or resource busy)
- EXDEV (cross-device link)

This behavior was confirmed via `strace`.

A potential solution is to fall back to copying the directory contents when the move operation fails.

This appears related to earlier issues such as #15613, but is still reproducible in Strapi v5 when using mounted `public/uploads` directories.

### Confirmation Checklist

Happyfield7:
### What does it do?

Adds a fallback when backing up the `public/uploads` directory during `strapi transfer`.

If `fs-extra.move()` fails (e.g. due to mounted directories), the code falls back to copying the directory contents and emptying the original directory.

### Why is it needed?

When `public/uploads` is mounted (e.g. Docker bind mount or Kubernetes volume), `fs-extra.move()` can fail with errors such as:

- EBUSY (device or resource busy)
- EXDEV (cross-device link)

This causes the transfer restore process to fail during the assets backup step, even though the issue is not related to permissions.

This change ensures that transfers work correctly in environments where uploads are mounted.

### How to test it?

1. Run two Strapi instances:
   - a source instance
   - a destination instance in Docker with `public/uploads` mounted as a volume
     - e.g. `-v /host/uploads:/app/public/uploads`

2. On the source instance:
   - create some content
   - upload at least one file
   - create a transfer token

3. From the destination instance, run:
   `strapi transfer --from <source-url> --from-token <token>`

4. Before this change:
   - transfer fails during the assets backup step

5. After this change:
   - transfer succeeds
   - uploads are transferred correctly

### Related issue(s)/PR(s)

Closes #25821

### DIFF
--- a/packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
+++ b/packages/core/data-transfer/src/strapi/providers/local-destination/index.ts
@@ -25,6 +25,12 @@ import { assertValidStrapi } from '../../../utils/providers';
 export const VALID_CONFLICT_STRATEGIES = ['restore'];
 export const DEFAULT_CONFLICT_STRATEGY = 'restore';
 
+const isAssetsBackupMoveError = (error: unknown): boolean => {
+  const err = error as NodeJS.ErrnoException;
+
+  return ['EBUSY', 'EXDEV'].includes(err.code ?? '');
+};
+
 export interface ILocalStrapiDestinationProviderOptions {
   getStrapi(): Core.Strapi | Promise<Core.Strapi>; // return an initialized instance of Strapi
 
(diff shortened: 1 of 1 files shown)


Happyfield7:
Superseded by #25853 (based on #23479).

PR #25823, opened 2026-03-25
fix(graphql): change @graphql-tools/schema to tilda version

unrevised6419:
### What does it do?

Changes the `@graphql-tools/schema` in graphql from fixed to tilda (~)

### Why is it needed?

To be able to pick up bug fixes at least from patches

### How to test it?

It be hard to test. But in my case I'm using graphql codegen, which fails with fixed 10.0.3
But it works fine if I override it to a higher patch.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/20204/changes#r1580173982


innerdvations:
Thanks! We try to avoid ranges to avoid supply chain attacks though, espsecially in light of how many there have been in recent years.

unrevised6419:
@innerdvations can we at least update it to latest minor version?

PR #25831, opened 2026-03-26
feat(i18n): update and create Slovak translations

fonodi:
## Update Slovak translations

Ready to review

### What does it do?
Update Slovak translations and create them in places where they are missing altogether.

### Why is it needed?
This makes a better user experience for Slovak users, that wish to use Strapi in a different locale than English.

### How to test it?
Tests for translations are already written in the project, this only extends the translations.

- [ ] Open any affected place and check if the translation is applied.


Bassel17 (approved):
LGTM!
thanks for the contribution

Issue #23571, opened 2025-05-22
Assets not showing in subfolder view of Media Library unless "Add new assets" is triggered


raulbalestra:
### Node Version

22.15.1

### NPM/Yarn/PNPM Version

Yarn

### Strapi Version

5.12.6

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

There’s a bug in the Media Library where assets inside a folder/subfolder are not displayed unless the "Add new assets" button is clicked and a new file is selected.

### Steps to Reproduce

Navigate to the Media Library.

Open any folder that contains assets and subfolders.

Click into a subfolder that already has assets uploaded.

Observe that the subfolder appears empty, even though it contains assets.

Now click "Add new assets" and select a file, then cancel or close the upload.

After this, the existing assets in the subfolder become visible.

### Expected Behavior

Assets stored in subfolders should be displayed immediately upon opening the subfolder, without needing to trigger the upload process.

### Logs

```shell

```

### Additional information

Video example: https://www.loom.com/share/cd8f5a2c3bec40bc9a53b543e0aab9f6?sid=23e4d2d8-df13-41a7-8c4d-d90438e6af50

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

akash-dabhi-qed:
I am able to recreate this issue. It also seems to have some resemblances to #23774 .

https://github.com/user-attachments/assets/eed2b43d-c1db-4e46-9727-f0178b5f8831

astuanax:
Hi, 
I ran in to this issue and tried to fix it.
Here is the pull request:
https://github.com/strapi/strapi/pull/25836

If someone on this ticket can review it, would be great.
Let me know if anything needs to be updated or modified.

PR #25836, opened 2026-03-26
fix (#23571) Assets not showing in subfolder view of Media Library unless "Add new assets" is triggered

astuanax:
### What does it do?

Changes the file endpoint to show file assets and recursively remove empty strings and empty objects (e.g. folderPath[$eq]=)

### Why is it needed?

File assets were not accessible in some cases. Especially after uploading via the API.

### How to test it?

Upload a file using below example code:

`const folder = await strapi.query("plugin::upload.folder").create({
    data: { name, parent: parentId }
  })`
  
  `// Strapi 5 internal upload service expects specific property names:
    // filepath instead of path, mimetype instead of type
    const uploadedFiles = await strapi
      .plugin("upload")
      .service("upload")
      .upload(
        {
          data: {
            fileInfo: {
              name: fileName,
              folder: folderId
            }
          },
          files: {
            filepath: filePath,
            originalFilename: fileName,
            mimetype: mimeType,
            size: stats.size
          }
        },
        {
          user: adminUser // Pass the admin user for createdBy/updatedBy
        }
      )`

Folder appears with the number of assets, but no files are visible.
Searching for a file, fixes the endpoint query and all files that match the string query will apear.
Except when there is no search query.

### Related issue(s)/PR(s)

[Assets not showing in subfolder view of Media Library unless "Add new assets" is triggered](https://github.com/strapi/strapi/issues/23571)

### DIFF
--- a/packages/core/upload/server/src/controllers/__tests__/admin-file.test.ts
+++ b/packages/core/upload/server/src/controllers/__tests__/admin-file.test.ts
@@ -1,0 +1,107 @@
+
+import adminFileController from '../admin-file';
+
+describe('Admin File Controller - find', () => {
+  let strapi;
+
+  beforeEach(() => {
+    const uploadService = {
+      findPage: jest.fn().mockResolvedValue({ results: [], pagination: {} }),
+    };
+
+    const fileService = {
+      signFileUrls: jest.fn((f) => f),
+    };
+
+    const pm = {
+      isAllowed: true,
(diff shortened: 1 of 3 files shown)


derrickmehaffy:
Hey there, it looks like this PR was superseeded by https://github.com/strapi/strapi/pull/26515 so going to mark as closed

PR #25844, opened 2026-03-27
fix(upload): allow upload of files with empty MIME type

### LINKED ISSUE #25834: `getAllowedFiles` client-side filter rejects files with empty browser MIME type even when "files" is an allowed type
### Node Version

24

### Package Manager

yarn

### Package Manager Version

11.9.0

### Strapi Version

5.40.0

### Operating System

Other

### Database

MySQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When a content type has a media field that accepts all asset types (images, files, videos, audios), uploading a file with an extension the **browser** doesn't recognize (e.g., `.lottie`) via the content type field's media picker fails with:

> "You can't upload this type of file, only the following types are accepted – images,files,videos,audios"

The same file uploads successfully from the **Media Library page** directly.

In `packages/upload/admin/src/utils/getAllowedFiles.ts`, the filter function rejects any file where `file.mime` produces a falsy first segment:

```js
const getAllowedFiles = (pluralTypes, files) => {
  // ...
  const allowedFiles = files.filter((file) => {
    const fileType = file?.mime?.split('/')[0];
    if (!fileType) {
      return false;  // ← BUG: rejects files with empty MIME
    }
    if (singularTypes.includes('file') && !['video','image','audio'].includes(fileType)) {
      return true;
    }
    return singularTypes.includes(fileType);
  });
  return allowedFiles;
};
```

This creates confusing behavior where the same user can upload the same file from the Media Library page but not from a content type field.

### Steps to Reproduce

1. Create a content type with a media field (single or multiple) that accepts all types (images, files, videos, audios).
2. In the admin panel, open a content entry and click the media field to add an asset.
3. Try to upload a `.lottie` file (or any file with an extension the browser doesn't map to a MIME type).
4. Observe the error: *"You can't upload this type of file, only the following types are accepted – images,files,videos,audios"*
5. Navigate to the Media Library page directly and upload the same file — it succeeds.

### Expected Behavior

Files with an empty/unknown browser MIME type should be allowed through when `"files"` is in the allowed types list, since they are generic files. The server-side MIME validation (`mime-validation.mjs`) already handles content detection properly via `file-type` and extension lookup — the client-side filter should not be more restrictive.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

akash-dabhi-qed:
### What does it do?

Changed the client-side `getAllowedFiles` filter in the upload plugin to allow files with an empty/unknown browser MIME type.

[fix-25834.webm](https://github.com/user-attachments/assets/fa095b89-113f-4641-b0f1-41cf1d98a90d)

### Why is it needed?

Browser did not know the MIME type file, this resulted in empty mime value, causing outright rejection previously even though the media field was configured to accept all file types. This resulted in inconsistency wherein the file which was allowed through media library route was rejected via content types media field.

### How to test it?

### Related issue(s)/PR(s)

Fixes #25834

### DIFF
--- a/packages/core/upload/admin/src/utils/getAllowedFiles.ts
+++ b/packages/core/upload/admin/src/utils/getAllowedFiles.ts
@@ -26,7 +26,7 @@ export const getAllowedFiles = (pluralTypes: string[] | null, files: AllowedFile
     const fileType = file?.mime?.split('/')[0];
 
     if (!fileType) {
-      return false;
+      return singularTypes.includes('file');
     }
 
     if (singularTypes.includes('file') && !['video', 'image', 'audio'].includes(fileType)) {
--- a/packages/core/upload/admin/src/utils/tests/getAllowedFiles.test.ts
+++ b/packages/core/upload/admin/src/utils/tests/getAllowedFiles.test.ts
@@ -86,6 +86,14 @@ const FILE_9 = {
   ...COMMON_PROPERTIES,
 };
 
+const FILE_10 = {
+  id: 10,
+  mime: '',
+  name: 'animation.lottie',
+  url: '/uploads/animation.lottie',
+  ...COMMON_PROPERTIES,
+};
+
 const files = [FILE_1, FILE_2, FILE_3, FILE_4, FILE_5, FILE_6, FILE_7, FILE_8, FILE_9];
 
 describe('UPLOAD | components | MediaLibraryInput | utils | getAllowedFiles', () => {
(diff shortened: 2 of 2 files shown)


nclsndr:
@claude review

nclsndr:
I guess @innerdvations you're the best placed to review that one.

innerdvations:
Thanks for the fix!

PR #25857, opened 2026-03-30
fix: change return to continue in deleteRelations when using foreign keys

### LINKED ISSUE #25855: deleteRelations exits early when using foreign keys, skipping remaining relations
### Node Version

20.20.0

### Package Manager

npm

### Package Manager Version

10.8.2

### Strapi Version

5.38.0

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

In `packages/core/database/src/entity-manager/index.ts`, the `deleteRelations` method exits early when the database dialect uses foreign keys.

Inside the loop over attributes, the following code is used:

```ts
if (db.dialect.usesForeignKeys()) {
  return;
}
```

This causes the function to stop execution entirely after the first matching attribute instead of continuing to process the remaining attributes.

As a result, some relations may not be properly cleaned up depending on attribute order.

### Steps to Reproduce

1. Use a database that supports foreign keys (e.g. PostgreSQL or MySQL)
2. Create a content type with multiple relational fields
3. Create entries with multiple relations populated
4. Delete an entry
5. Observe that not all relations are properly cleaned up

### Expected Behavior

The function should skip only the current attribute when foreign keys are used, and continue processing the remaining attributes.

Replacing `return` with `continue` would ensure all attributes are handled correctly.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

This behavior depends on attribute order and can lead to inconsistent relation cleanup.

The issue is subtle but affects core database logic.

### Confirmation Checklist

guoyangzhen:
Fixes #25855

## Problem

In `packages/core/database/src/entity-manager/index.ts`, the `deleteRelations` method uses `return` instead of `continue` when `db.dialect.usesForeignKeys()` is true. This causes the function to exit the entire handler early, skipping all remaining relation attributes.

## Root Cause

```ts
if (db.dialect.usesForeignKeys()) {
  return;  // exits entire function, not just current iteration
}
```

The code is inside a loop over attributes. When any attribute hits this condition, `return` exits the entire `deleteRelations` function instead of just skipping to the next attribute.

## Fix

Changed `return` to `continue` to properly skip the current attribute and continue processing remaining attributes.

### DIFF
--- a/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
+++ b/packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
@@ -3467,15 +3467,7 @@ export async function* getBatchToDiscard({
   uid: string;
   defaultBatchSize?: number;
 }) {
-  const client = db.config.connection.client;
-  const isSQLite =
-    typeof client === 'string' && ['sqlite', 'sqlite3', 'better-sqlite3'].includes(client);
-
-  // The SQLite documentation states that the maximum number of terms in a
-  // compound SELECT statement is 500 by default.
-  // See: https://www.sqlite.org/limits.html
-  // To ensure a successful migration, we limit the batch size to 500 for SQLite.
-  const batchSize = isSQLite ? Math.min(defaultBatchSize, 500) : defaultBatchSize;
+  const batchSize = Math.min(defaultBatchSize, db.dialect.getBatchInsertSize());
   let offset = 0;
   let hasMore = true;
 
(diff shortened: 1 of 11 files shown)


innerdvations:
Thanks for the fix! Can you please sign the [Contributor License Agreement](https://cla.strapi.io/strapi/strapi?pullRequest=25857) so that we can merge this?

PR #25865, opened 2026-03-31
feat: removing A/B testing from the prompt

cache-your-dreams:
### What does it do?
Removing A/B testing opt-in from the CLI

### Why is it needed?
Currently A/B testing opt-in isn't used

### How to test it?
Create an experimental and install the Strapi app. A/B testing should no longer be there.


innerdvations:
if tests pass and we're sure this isn't used or planned to be used anywhere, we can merge

emilieringwald:
Yep confirmed, we won't use it anytime soon. We can always put it back later if it comes into plans.

PR #25867, opened 2026-03-31
feat(data-transfer): add directory export/import format

innerdvations:
### What does it do?

- Adds **directory-based Strapi Data Transfer (DTS) export/import**: the CLI accepts an **unpacked export directory** (same layout as inside a `.tar`: `metadata.json`, `schemas/`, `entities/`, `links/`, `configuration/`, `assets/`, …) in addition to `.tar` / `.tar.gz` / `.tar.enc` archives.
- Introduces **directory source (and related) providers** in `@strapi/data-transfer`, wired through the public package surface and Strapi CLI import/export flow (including validation that a path looks like an unpacked export).
- Refactors the Strapi core CLI to load Inquirer via a small **`getInquirer()`** helper (cached dynamic import) instead of repeating `await import('inquirer')` at many call sites.
- Fixes **`@strapi/cloud-cli`**: `deploy-project/action.ts` no longer references a removed `loadInquirer` symbol (uses the same dynamic import pattern as elsewhere in that package).
- Moves **e2e DTS fixtures** under `tests/e2e/data/` from committed `.tar` blobs to **unpacked directories** so JSON/JSONL and small metadata files are normal text diffs in Git; binary uploads stay under `assets/uploads/` as before.
- Updates **contributor / e2e documentation** for the new fixture layout and import workflow.
- Adjusts **tooling** so fixture trees are not churned by repo-wide formatting: `.prettierignore` entries for `tests/e2e/data/**` and `**/tmp-dts-export/**` (e.g. `examples/complex/tmp-dts-export`), plus optional **EditorConfig** overrides for those paths so editor save behavior doesn’t fight fixture files.

### Why is it needed?

- **Unpacked directories** make export/import fixtures easier to **review, diff, and update** in Git than opaque archives, while keeping the same on-disk layout Strapi already uses inside a transfer archive.
- **Single, cached Inquirer load** in the core CLI avoids duplicated dynamic imports and keeps ESM/CJS interop predictable across CLI commands.
- The **cloud CLI** build broke after removing `loadInquirer`; this restores a valid implementation without pulling core helpers into cloud-cli.
- **Prettier / format-on-save** was still touching `metadata.json` and similar files under example export trees; ignoring those paths prevents noisy EOF/newline-only diffs when running `yarn prettier:write`.

### How to test it?

**Environment:** Node/Yarn as in the repo root; build Strapi packages if your workflow requires it (`yarn setup` / `yarn build` as you normally use for local dev).

**Automated**

- Unit tests: `@strapi/data-transfer` directory provider tests and Strapi CLI import/export/transfer/admin command tests that were updated on this branch.
- CLI tests: `tests/cli/tests/strapi/data-transfer/` (export, import, roundtrip) — they exercise export → `output.tar` → import using the test app and e2e-backed seed data.
- E2e: run the Playwright suite or the subset that resets DB from `tests/e2e/data` (e.g. `with-admin`) if you touch import paths or fixtures.

**Manual**

- From a Strapi app, run **`strapi export`** to a filename (producing a `.tar` as today), then extract it and run **`strapi import -f /path/to/unpacked/dir`** (or point `-f` at the directory directly, per CLI help) and confirm import completes.
- Optionally compare behavior importing the **same content** as `.tar` vs unpacked directory (DB / admin content parity).
- Run **`yarn prettier:write`** and confirm **`tests/e2e/data/**`** and **`examples/complex/tmp-dts-export/**`** files are not rewritten (no spurious newline-only changes to `metadata.json`).

### Related issue(s)/PR(s)

CMS-538

### DIFF
--- a/packages/core/data-transfer/src/directory/providers/destination/index.ts
+++ b/packages/core/data-transfer/src/directory/providers/destination/index.ts
@@ -156,23 +156,32 @@ class LocalDirectoryDestinationProvider implements IDestinationProvider {
     return new Writable({
       objectMode: true,
       write(data: IAsset, _encoding, callback) {
-        const entryPath = path.join(root, 'assets', 'uploads', data.filename);
-        const entryMetadataPath = path.join(root, 'assets', 'metadata', `${data.filename}.json`);
+        const { filename } = data;
+        const entryPath = path.join(root, 'assets', 'uploads', filename);
+        const entryMetadataPath = path.join(root, 'assets', 'metadata', `${filename}.json`);
 
-        fs.mkdirSync(path.dirname(entryPath), { recursive: true });
-        fs.mkdirSync(path.dirname(entryMetadataPath), { recursive: true });
+        const assetWriteError = (cause: unknown) =>
+          new ProviderTransferError(`Failed to write asset ${filename}`, {
+            details: {
+              error: cause instanceof Error ? cause : new Error(String(cause)),
+            },
+          });
 
-        fs.writeFileSync(entryMetadataPath, JSON.stringify(data.metadata), 'utf8');
+        let fileStream: ReturnType<typeof createWriteStream>;
 
-        const fileStream = createWriteStream(entryPath);
+        try {
+          fs.mkdirSync(path.dirname(entryPath), { recursive: true });
+          fs.mkdirSync(path.dirname(entryMetadataPath), { recursive: true });
+          fs.writeFileSync(entryMetadataPath, JSON.stringify(data.metadata), 'utf8');
+          fileStream = createWriteStream(entryPath);
+        } catch (error: unknown) {
+          callback(assetWriteError(error));
+          return;
+        }
 
         pipeline(data.stream, fileStream, (err) => {
           if (err) {
-            callback(
-              new ProviderTransferError(`Failed to write asset ${data.filename}`, {
-                details: { error: err },
-              })
-            );
+            callback(assetWriteError(err));
             return;
           }
           callback(null);
--- a/packages/core/data-transfer/src/directory/providers/destination/utils.ts
+++ b/packages/core/data-transfer/src/directory/providers/destination/utils.ts
@@ -37,6 +37,15 @@ export const createDirectoryJsonlWriter = (
   };
 
   return new Writable({
+    async final(callback) {
+      try {
+        await flush();
+        callback();
+      } catch (err: unknown) {
+        callback(err instanceof Error ? err : new Error(String(err)));
+      }
+    },
+
     async destroy(err, callback) {
       await flush();
       callback(err);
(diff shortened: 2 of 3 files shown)


innerdvations:
@dosubot Only one label may be present on a PR at a time; if you add one, you must also remove one

innerdvations:
@pwizla No rush (as in, even if it takes a few weeks after we release this it's fine) but we might want documentation on the extra import/export options being added here

packages/core/data-transfer/src/directory/providers/destination/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    await fsp.mkdir(path.dirname(targetPath), { recursive: true });
+    await fsp.writeFile(targetPath, buffer, 'utf8');
+    buffer = '';
+  };
+
+  const push = (chunk: string | Buffer) => {
+    buffer += chunk;
+  };
+
+  return new Writable({
```
butcherZ, line 39: we might want to add a final here for the normal close path so that the last buffer can be flushed and doesn't get silently dropped :)
```
return new Writable({
  async final(callback) {     
    try {
      await flush();
      callback(null);
    } catch (err) {
      callback(err instanceof Error ? err : new Error(String(err)));
    }
  },
  async destroy(err, callback) { 
    await flush();
    callback(err);
  },
});
```

packages/core/data-transfer/src/directory/providers/destination/index.ts
```diff
@@ -0,0 +1,183 @@
...
+    return new Writable({
+      objectMode: true,
+      write(data: IAsset, _encoding, callback) {
+        const entryPath = path.join(root, 'assets', 'uploads', data.filename);
+        const entryMetadataPath = path.join(root, 'assets', 'metadata', `${data.filename}.json`);
+
+        fs.mkdirSync(path.dirname(entryPath), { recursive: true });
+        fs.mkdirSync(path.dirname(entryMetadataPath), { recursive: true });
+
+        fs.writeFileSync(entryMetadataPath, JSON.stringify(data.metadata), 'utf8');
```
butcherZ: I'd wrap the sync calls in a try/catch and route the errors :D
```
        try {
          fs.mkdirSync(path.dirname(entryPath), { recursive: true });
          fs.mkdirSync(path.dirname(entryMetadataPath), { recursive: true });
          fs.writeFileSync(entryMetadataPath, JSON.stringify(data.metadata), 'utf8');
        } catch (err) {
          callback(
            new ProviderTransferError(`Failed to prepare asset directory for ${data.filename}`, {
              details: { error: err },
            })
          );
          return;
        }
```

butcherZ:
> @pwizla No rush (as in, even if it takes a few weeks after we release this it's fine) but we might want documentation on the extra import/export options being added here

@pwizla this PR is perfect to test my plugin :D I just ran it myself and your Router identified everything correctly 😄

PR #6179, opened 2020-05-12
Add basis for rbac models

Convly:
## Description of what you did:

Add `admin::User` & `admin::Role` models for RBAC.

PR #25873, opened 2026-03-31
chore: bump typescript to v6

### LINKED ISSUE #25861: fix: typescript v7 will not allow `moduleResolution` `node`
### Node Version

n/a

### Package Manager

Other

### Package Manager Version

n/a

### Strapi Version

5.40.ß

### Operating System

Other

### Database

Other

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/dargmuesli/creal_strapi/pull/141 // https://github.com/dargmuesli/creal_strapi/actions/runs/23729339027/job/69119637272#step:7:401

### Bug Description

Typescript v6 has just been released which deprecates `--moduleResolution node` (a.k.a. `--moduleResolution node10`): https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#deprecated:---moduleresolution-node-(a.k.a.---moduleresolution-node10)

For me this issues arises for `@strapi/typescript-utils/tsconfigs/server` which still references the deprecated module resolution. As the naming mentions "server", I'd propose to change the default from `node` to `nodenext` across the strapi monorepo.

### Steps to Reproduce

Have a tsconfig.json like this:

```json
{
  "extends": "@strapi/typescript-utils/tsconfigs/server",
  "compilerOptions": {
    "outDir": "dist",
    "rootDir": "."
  },
  "include": ...,
  "exclude": ...
}
```

Try to compile with Typescript v6 -> error.

### Expected Behavior

Compiling with typescript v6 works per default.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

You can add the module and moduleResolution keys to a project's tsconfig.json for the keys to be overridden downstream:

```json
{
  "extends": "@strapi/typescript-utils/tsconfigs/server",
  "compilerOptions": {
    "module": "nodenext",
    "moduleResolution": "nodenext",
    "outDir": "dist",
    "rootDir": "."
  },
  "include": ...,
  "exclude": ...
}
```

### Confirmation Checklist

nclsndr:
### What does it do?

* **TypeScript 6** — Bumps the repo dev dependency from TypeScript 5.9.3 to **6.0.2** and updates `yarn.lock`.
* **Compiler config alignment** — Adjusts `tsconfig` across the monorepo, examples, CLI templates, and shared `packages/utils/tsconfig` presets for TS 6: e.g. `moduleResolution` **Node → bundler** where appropriate, drops redundant `baseUrl` entries, sets `types` (e.g. `["node"]` on the base preset), trims `DOM.Iterable` from client `lib` (covered via `DOM` in newer libs), and tweaks per-package options (including the Sentry plugin resolving `Config` via a **relative** import instead of `src/config`).
* **Small source fixes for the upgrade** — Removes an obsolete `@ts-expect-error` around `Readable.fromWeb` in `create-strapi-app`, and updates a few core files (admin permissions sanitize, content-manager, upload admin, core types `fetch`, etc.) as needed for clean compilation.
* **Lint / architecture** — In **content-manager** server: refactors `layouts.ts` layout-sync logic to an equivalent nested structure (satisfies lint/control-flow rules), and in `populate.ts` resolves a **circular import** by calling the populate-builder via `strapi.plugin('content-manager').service('populate-builder')` instead of `getService` from `utils`.
* **Database package Jest typings** — Adds `@types/jest` and sets `compilerOptions.types` to `["node", "jest"]` in `packages/core/database/tsconfig.json` so Jest globals type-check under the new toolchain.

### Why is it needed?

* Keeps the codebase on a **supported TypeScript major** and unlocks TS 6 language/tooling behavior.
* TS 6 and stricter resolution/tsconfig rules surface **config and import issues** (e.g. `baseUrl`/path patterns, Sentry’s `src/config` import) that need coordinated **tsconfig and source** updates.
* Follow-up fixes ensure **ESLint passes** on production code (layouts + populate) and `@strapi/database` continues to **type-check tests** without missing Jest ambient types.

### How to test it?

* **Environment:** Node/Yarn as in the repo; run `yarn install` after checkout so TypeScript 6 and lockfile match.
* **Typecheck / build:** From the repo root, run the usual CI-equivalent checks (e.g. `yarn build` / `yarn test:ts` or the script your team uses for full `tsc` across packages — match what runs in CI on this branch).
* **Spot-check behavior:**
  * **Content-manager:** Edit layouts with mixed field sizes / schema changes and confirm the **edit view layout** still syncs and fields are not dropped incorrectly (`layouts.ts` path).
  * **Deep populate:** Exercise **collection types** that use deep populate (e.g. list/detail in admin) to confirm no regression from the populate-builder resolution change (`populate.ts`).
  * **Database:** Run `packages/core/database` unit tests if applicable (`yarn workspace @strapi/database test` or equivalent).
* **Smoke:** Create or run an **example app** (e.g. `examples/kitchensink-ts` or a generated app from templates) and confirm `yarn build` / dev server still work with updated `tsconfig` templates.

### Related issue(s)/PR(s)

* Fixes strapi/strapi#25861

### DIFF
--- a/examples/complex/package.json
+++ b/examples/complex/package.json
@@ -58,7 +58,7 @@
     "@types/node": "^20",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "^6.0.2"
   },
   "engines": {
     "node": ">=20.0.0 <=24.x.x",
--- a/examples/complex/package.json
+++ b/examples/complex/package.json
@@ -59,7 +59,7 @@
     "@types/node": "20.19.41",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "6.0.2"
   },
   "engines": {
     "node": ">=20.0.0 <=26.x.x",
--- a/examples/empty/package.json
+++ b/examples/empty/package.json
@@ -27,7 +27,7 @@
     "@types/node": "^20",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "^6.0.2"
   },
   "engines": {
     "node": ">=20.0.0 <=24.x.x",
--- a/examples/empty/package.json
+++ b/examples/empty/package.json
@@ -27,7 +27,7 @@
     "@types/node": "20.19.41",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "6.0.2"
   },
   "engines": {
     "node": ">=20.0.0 <=26.x.x",
--- a/packages/core/admin/server/tsconfig.json
+++ b/packages/core/admin/server/tsconfig.json
@@ -3,8 +3,7 @@
   "include": ["src"],
   "compilerOptions": {
     "rootDir": "../",
-    "baseUrl": ".",
-    "types": ["lodash", "jest"],
+    "types": ["node", "lodash", "jest"],
     "esModuleInterop": true
   }
 }
--- a/packages/core/admin/server/tsconfig.json
+++ b/packages/core/admin/server/tsconfig.json
@@ -2,7 +2,6 @@
   "extends": "tsconfig/base.json",
   "include": ["src"],
   "compilerOptions": {
-    "rootDir": "../",
     "types": ["node", "jest"]
   }
 }
--- a/packages/core/admin/server/tsconfig.json
+++ b/packages/core/admin/server/tsconfig.json
@@ -1,7 +1,13 @@
 {
   "extends": "tsconfig/base.json",
-  "include": ["src"],
+  "include": [
+    "src"
+  ],
   "compilerOptions": {
-    "types": ["node", "jest"]
+    "rootDir": "../",
+    "types": [
+      "node",
+      "jest"
+    ]
   }
 }
--- a/packages/core/admin/server/tsconfig.json
+++ b/packages/core/admin/server/tsconfig.json
@@ -1,13 +1,8 @@
 {
   "extends": "tsconfig/base.json",
-  "include": [
-    "src"
-  ],
+  "include": ["src"],
   "compilerOptions": {
     "rootDir": "../",
-    "types": [
-      "node",
-      "jest"
-    ]
+    "types": ["node", "jest"]
   }
 }
--- a/packages/core/content-type-builder/package.json
+++ b/packages/core/content-type-builder/package.json
@@ -41,7 +41,6 @@
     },
     "./package.json": "./package.json"
   },
-  "types": "./dist/index.d.ts",
   "files": [
     "dist/",
     "strapi-server.js"
--- a/packages/core/review-workflows/admin/tsconfig.json
+++ b/packages/core/review-workflows/admin/tsconfig.json
@@ -2,9 +2,10 @@
   "extends": "tsconfig/client.json",
   "compilerOptions": {
     "rootDir": "../",
-    "baseUrl": ".",
     "paths": {
-      "@tests/*": ["./tests/*"]
+      "@tests/*": ["./tests/*"],
+      "msw": ["../../admin/node_modules/msw"],
+      "msw/*": ["../../admin/node_modules/msw/*"]
     }
   },
   "include": ["../package.json", "./src", "../shared", "./tests", "./custom.d.ts"]
--- a/packages/core/review-workflows/admin/tsconfig.json
+++ b/packages/core/review-workflows/admin/tsconfig.json
@@ -3,20 +3,9 @@
   "compilerOptions": {
     "rootDir": "../",
     "paths": {
-      "@tests/*": [
-        "./tests/*"
-      ]
+      "@tests/*": ["./tests/*"]
     },
-    "types": [
-      "@testing-library/jest-dom",
-      "jest"
-    ]
+    "types": ["@testing-library/jest-dom", "jest"]
   },
-  "include": [
-    "../package.json",
-    "./src",
-    "../shared",
-    "./tests",
-    "./custom.d.ts"
-  ]
+  "include": ["../package.json", "./src", "../shared", "./tests", "./custom.d.ts"]
 }
--- a/packages/core/review-workflows/admin/tsconfig.json
+++ b/packages/core/review-workflows/admin/tsconfig.json
@@ -1,10 +1,22 @@
 {
   "extends": "tsconfig/client.json",
   "compilerOptions": {
+    "rootDir": "../",
     "paths": {
-      "@tests/*": ["./tests/*"]
+      "@tests/*": [
+        "./tests/*"
+      ]
     },
-    "types": ["@testing-library/jest-dom", "jest"]
+    "types": [
+      "@testing-library/jest-dom",
+      "jest"
+    ]
   },
-  "include": ["../package.json", "./src", "../shared", "./tests", "./custom.d.ts"]
+  "include": [
+    "../package.json",
+    "./src",
+    "../shared",
+    "./tests",
+    "./custom.d.ts"
+  ]
 }
--- a/packages/core/review-workflows/admin/tsconfig.json
+++ b/packages/core/review-workflows/admin/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "extends": "tsconfig/client.json",
   "compilerOptions": {
-    "rootDir": "../",
     "paths": {
       "@tests/*": ["./tests/*"]
     },
--- a/packages/core/upload/admin/src/components/EditAssetDialog/EditAssetContent.tsx
+++ b/packages/core/upload/admin/src/components/EditAssetDialog/EditAssetContent.tsx
@@ -172,18 +172,20 @@ export const EditAssetContent = ({
   };
 
   const activeFolderId = asset?.folder?.id;
-  const initialFormData = !folderStructureIsLoading && {
-    name: asset?.name,
-    alternativeText: asset?.alternativeText ?? undefined,
-    caption: asset?.caption ?? undefined,
-    focalPoint: asset?.focalPoint ?? null,
-    parent: {
-      value: activeFolderId ?? undefined,
-      label:
-        findRecursiveFolderByValue(folderStructure!, activeFolderId!)?.label ??
-        folderStructure![0].label,
-    },
-  };
+  const initialFormData: FormInitialData = !folderStructureIsLoading
+    ? {
+        name: asset?.name,
+        alternativeText: asset?.alternativeText ?? undefined,
+        caption: asset?.caption ?? undefined,
+        focalPoint: asset?.focalPoint ?? null,
+        parent: {
+          value: activeFolderId ?? undefined,
+          label:
+            findRecursiveFolderByValue(folderStructure!, activeFolderId!)?.label ??
+            folderStructure![0].label,
+        },
+      }
+    : {};
 
   const handleClose = (values?: { [key: string]: unknown }) => {
     if (!isEqual(initialFormData, values)) {
--- a/packages/utils/tsconfig/base.json
+++ b/packages/utils/tsconfig/base.json
@@ -11,6 +11,7 @@
     "noImplicitAny": true,
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "types": ["node"]
   }
 }
--- a/packages/utils/tsconfig/base.json
+++ b/packages/utils/tsconfig/base.json
@@ -15,7 +15,6 @@
     "noEmit": true,
     "types": [],
     "noUncheckedSideEffectImports": true,
-    "erasableSyntaxOnly": true,
-    "types": ["node"]
+    "erasableSyntaxOnly": true
   }
 }
--- a/packages/utils/tsconfig/base.json
+++ b/packages/utils/tsconfig/base.json
@@ -15,6 +15,7 @@
     "noEmit": true,
     "types": [],
     "noUncheckedSideEffectImports": true,
-    "erasableSyntaxOnly": true
+    "erasableSyntaxOnly": true,
+    "types": ["node"]
   }
 }
--- a/.github/actions/community-pr-triage/package.json
+++ b/.github/actions/community-pr-triage/package.json
@@ -16,6 +16,6 @@
   "devDependencies": {
     "@types/node": "20.19.41",
     "tsup": "^8.0.0",
-    "typescript": "^5.4.0"
+    "typescript": "6.0.2"
   }
 }
--- a/docs/package.json
+++ b/docs/package.json
@@ -48,7 +48,7 @@
     "docusaurus-plugin-typedoc": "0.22.0",
     "typedoc": "0.25.10",
     "typedoc-plugin-markdown": "3.17.1",
-    "typescript": "5.4.5",
+    "typescript": "6.0.2",
     "unist-util-visit": "5.0.0"
   }
 }
--- a/docs/package.json
+++ b/docs/package.json
@@ -55,7 +55,7 @@
     "docusaurus-plugin-typedoc": "1.4.2",
     "typedoc": "0.28.19",
     "typedoc-plugin-markdown": "4.11.0",
-    "typescript": "5.9.3",
+    "typescript": "6.0.2",
     "unist-util-visit": "5.0.0"
   }
 }
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -671,6 +671,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/helper-string-parser@npm:^7.25.9":
+  version: 7.25.9
+  resolution: "@babel/helper-string-parser@npm:7.25.9"
+  checksum: 10c0/7244b45d8e65f6b4338a6a68a8556f2cb161b782343e97281a5f2b9b93e420cad0d9f5773a59d79f61d0c448913d06f6a2358a87f2e203cf112e3c5b53522ee6
+  languageName: node
+  linkType: hard
+
 "@babel/helper-string-parser@npm:^7.27.1":
   version: 7.27.1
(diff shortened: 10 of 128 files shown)


innerdvations:
It looks like it breaks the repo build, but also: what breaking changes might there be for existing projects?

nclsndr:
> It looks like it breaks the repo build

Yes I had to adapt some production code types

> what breaking changes might there be for existing projects

Normally none since built types would not change. We still have to pay attention to the package.json edits though

nclsndr:
Experimental has been published: 0.0.0-experimental.94e30af460b8e82811123516a94684baf8eca4af 

It seems the changes impacted the `npx create-strapi` command. Currently investigating

nclsndr:
Experimental release: 0.0.0-experimental.f1eddb176cc17bc4f12f969955b64c4477b3093a

nclsndr:
# Optional Peer Dependencies: Override or Accept?

## The problem

Two packages in our dependency tree declare `typescript` as an **optional** peer with a range that excludes v6:

| Package | Via | TS peer range | Meta |
|---|---|---|---|
| `@formatjs/intl@2.10.0` | `react-intl` | `^4.7 \|\| 5` | `optional: true` |
| `react-intl@6.6.2` | direct (12 workspace packages) | `^4.7 \|\| 5` | `optional: true` |

No type-level or runtime incompatibility with TS6 has been identified. The peer dep is advisory.

## Upstream fix exists — but requires React 19

[PR #6180](https://github.com/formatjs/formatjs/pull/6180) (merged 2026-03-27) removed the TS peer dep entirely. Shipped in:
- `react-intl@10.1.1` — requires **React 19**
- `@formatjs/intl@4.1.5`

Maintainer Long Ho on backporting to v6/v7: **"I won't be retroactively fixing older versions."** ([#6179](https://github.com/formatjs/formatjs/issues/6179))

We stay on `react-intl@6.6.2` (React 18). No upgrade path without React 19.

## What actually happens at install time?

Because the peer dep is **optional**, the behavior depends on the package manager and its strictness setting:

| Package Manager | Default behavior | Strict mode | Optional peer in strict mode |
|---|---|---|---|
| **npm** | Warning | `--strict-peer-deps` | **Warning only** — optional peers are exempt |
| **Yarn Classic (v1)** | Warning | *(no strict mode)* | N/A |
| **Yarn Berry (v2+)** | Warning | `enableStrictPeerDependencies: true` | **Warning only** — optional peers are exempt |
| **pnpm (v9+)** | Warning | `strict-peer-dependencies=true` in `.npmrc` | **Install error** — pnpm does NOT exempt optional peers |

**Risk surface:** warning for everyone, **except** pnpm users who explicitly set `strict-peer-dependencies=true` (not the default since pnpm v9).

## Option A: Inject overrides into generated projects

Add PM-specific override fields in `create-strapi-app`'s `createPackageJSON()` (in `packages/cli/create-strapi-app/src/utils/package-json.ts`), gated on `scope.useTypescript === true`.

The syntax differs per package manager, and not all support dynamic references:

| PM | Field | `$typescript` ref? | Syntax |
|---|---|---|---|
| npm | `overrides` | Yes | `"react-intl": { "typescript": "$typescript" }` |
| pnpm | `pnpm.overrides` | Yes | `"react-intl>typescript": "$typescript"` |
| Yarn (all versions) | `resolutions` | **No** — hardcoded version only | `"react-intl/typescript": "6.0.0"` |

### Pros

- Zero warnings on fresh install
- Prevents install failure for pnpm strict-mode users
- Explicit — makes the workaround visible in the project's `package.json`

### Cons

- Introduces PM-specific branching in `createPackageJSON` for the first time (complexity)
- Yarn path requires hardcoding the TS version from `scope.devDependencies` — **goes stale** if user later bumps TS manually
- Only benefits new projects — existing projects must add overrides manually regardless (release notes)
- Temporary workaround — removable once we move to React 19 + `react-intl@10`
- Solves a mostly cosmetic problem (warnings) for npm/yarn; the real failure case (pnpm strict) is a narrow audience

## Option B: Accept the warnings, document it

Do nothing in the template. Document the situation in release notes and migration guide.

### Pros

- Zero code complexity added to `create-strapi-app`
- No stale-version risk for yarn users
- Existing and new projects treated the same way (manual override if desired)
- Less to clean up later when React 19 / `react-intl@10` lands

### Cons

- Users see a peer dep warning on every install (cosmetic)
- pnpm users with `strict-peer-dependencies=true` must add the override themselves or relax the setting
- Looks "unfinished" — warnings during install can erode confidence

## Decision needed

**Should we inject the overrides into generated projects (Option A), or accept the warnings and document them (Option B)?**

Key factors:
- The warning is cosmetic for ~99% of users (npm, yarn, pnpm default)
- The only real breakage is pnpm + explicit strict mode — a small subset
- Option A adds PM-branching complexity that has never existed in `createPackageJSON` before
- Both options require release notes for existing projects

packages/core/content-type-builder/package.json
```diff
@@ -41,7 +41,6 @@
     },
     "./package.json": "./package.json"
   },
-  "types": "./dist/index.d.ts",
```
Adzouz: Did you forget to add the new path here? The line was just removed

packages/core/upload/admin/src/components/EditAssetDialog/EditAssetContent.tsx
```diff
@@ -172,18 +172,20 @@ export const EditAssetContent = ({
...
+        caption: asset?.caption ?? undefined,
+        focalPoint: asset?.focalPoint ?? null,
+        parent: {
+          value: activeFolderId ?? undefined,
+          label:
+            findRecursiveFolderByValue(folderStructure!, activeFolderId!)?.label ??
+            folderStructure![0].label,
+        },
+      }
+    : {};
```
Adzouz, line 203: We're changing from a false default value to an empty object. Can we make sure that it's not breaking anything here?

Adzouz (commented):
Thanks a lot for this big change 👏 I tested the basic commands locally and everything looks good. Also the project is running without any issue.
Two comments and also one concern raised on my side, `module` to "CommonJS" and `moduleResolution` to "bundler" are two settings designed for different environments. Mixing them creates a gap between what TS accepts and what Node.js actually does at runtime. Do you have an opinion on this?

Also for the blocker with the two options proposed, I'd go with option B. Let me know which one do you prefer 😊

packages/core/upload/admin/src/components/EditAssetDialog/EditAssetContent.tsx
nclsndr: Type def is 

```ts
interface FormInitialData {
  name?: string;
  alternativeText?: string;
  caption?: string;
  focalPoint?: FocalPoint | null;
  parent?: {
    value?: number;
    label: string;
  };
}
```

All runtime actually expect object based structure. I think the boolean was a design flaw.

nclsndr:
> Two comments and also one concern raised on my side, `module` to "CommonJS" and `moduleResolution` to "bundler" are two settings designed for different environments. Mixing them creates a gap between what TS accepts and what Node.js actually does at runtime. Do you have an opinion on this?

Yes this is intentional @Adzouz, this is weird but it's the shortest path I found before we actually upgrade the entire monorepo to proper ESM in source.
Here the [release note](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#combining---moduleresolution-bundler-with---module-commonjs) about this option.

packages/core/content-type-builder/package.json
nclsndr: This was a leftover while the rest of the mapping relies on `"exports"` prop resolution. Which is quite weird given "type" is silently "commonjs" while the resolution is forced to ESM exports… We'll have to align all of this moving forward.

innerdvations (commented):
Overall I think this is fine, I don't see any blockers other than getting tests to pass.

However, I think we need a big flashing sign release note on this one, since even though there aren't runtime changes it will likely break CI for some segment of users.

Maybe we combine it with dropping support for Node 20 / adding support for Node 26 next month, since people will already be expecting some CI changes there?

@Convly do you have any opinions on this upgrade?

nclsndr:
> Maybe we combine it with dropping support for Node 20

I very much like this idea!

packages/core/review-workflows/admin/tsconfig.json
```diff
@@ -2,9 +2,10 @@
   "extends": "tsconfig/client.json",
   "compilerOptions": {
     "rootDir": "../",
-    "baseUrl": ".",
     "paths": {
-      "@tests/*": ["./tests/*"]
+      "@tests/*": ["./tests/*"],
+      "msw": ["../../admin/node_modules/msw"],
+      "msw/*": ["../../admin/node_modules/msw/*"]
```
unrevised6419: Question: What is the reason for adding these aliases?

unrevised6419:
On `develop`, `packages/core/database/tsconfig.json` declares no `types` field and `@types/jest` is not a dep, so jest ambient globals are not in the package's TS program at all — tests compile via `ts-jest`, src never sees them. This PR adds `@types/jest` + sets `types: ["node", "jest"]` in the package tsconfig, which makes `describe`/`it`/`expect`/`global.*` augmentations resolvable inside production `src/` too.

Opened #26260 against this branch as a POC fix: split `database` into a tsconfig solution file with two referenced projects — `tsconfig.build.json` types `["node"]`, `tsconfig.test.json` types `["jest", "node"]`. Build code can no longer reference jest globals; tests still type-check normally. The test project explicitly includes `src/global.d.ts` so existing global augmentations (`declare var strapi`, custom knex module declarations) remain available to tests.

PS: Another approach is to use `@jest/globals`

```js
import { describe, it, test, jest, ... } from '@jest/globals'
```

packages/core/admin/server/tsconfig.json
```diff
@@ -3,8 +3,7 @@
   "include": ["src"],
   "compilerOptions": {
     "rootDir": "../",
-    "baseUrl": ".",
-    "types": ["lodash", "jest"],
+    "types": ["node", "lodash", "jest"],
```
unrevised6419: Note: `lodash` is redundant in `types`, it does not augment anything globally. Exists in multiple tsconfig files

nclsndr: #26627

examples/complex/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/node": "^20",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "^6.0.2"
```
unrevised6419:
```suggestion
    "typescript": "6.0.2"
```

examples/empty/package.json
```diff
@@ -27,7 +27,7 @@
     "@types/node": "^20",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "typescript": "^5"
+    "typescript": "^6.0.2"
```
unrevised6419:
```suggestion
    "typescript": "6.0.2"
```

unrevised6419 (commented):
`.github/actions/community-pr-triage/package.json#devDependencies.typescript` is left behind with `"^5.4.0"`

nclsndr:
Moved back to draft to be collected in nearly-breaking-changes release

unrevised6419:
@nclsndr ~should I move `outDir`, `rootDir`, `exclude`, `baseUrl` fixes into https://github.com/strapi/strapi/pull/26699 ?~ they are not TS breaking changes

Note: Moved

unrevised6419:
#26699 extracts the **non-breaking** tsconfig/types changes out of this PR so they can land independently of the TypeScript 6 bump: per-workspace `types` scoping, centralized `noEmit`, `baseUrl` removal, redundant-compilerOption cleanup, and per-workspace jsconfigs.

To show what's left here once #26699 lands, I merged this branch on top of it (base = the #26699 branch) — the resulting diff is just the genuine TS 6 delta: the `typescript@6.0.2` bump plus the `moduleResolution: bundler` / `lib` changes in the configs #26699 doesn't restructure:

https://github.com/unrevised6419/strapi/pull/9

Conflicts were resolved keeping #26699's intentional changes and this PR's TS 6 intent (which mostly merged cleanly). Might make this upgrade smaller to review if the scoping work goes in first.

unrevised6419:
https://devblogs.microsoft.com/typescript/announcing-typescript-7-0-rc/

unrevised6419:
I was doing some tests with upgrading `react-intl` from [v6 to v7](https://github.com/strapi/strapi/pull/26776) which barely has any breaking changes. `react-intl` requires minimum TypeScript v5.6. Wouldn't it be better if we upgrade first to latest v5, then move v5 -> v6 -> v7 ?

```
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p29c857), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p31e5fd), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p549b70), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p5e1a1b), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p61d572), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p636b12), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (p90ba12), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (pcaafb1), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (pda1723), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (pdc32b6), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (pe4d73a), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
➤ YN0060: │ typescript is listed by your project with version 5.4.5 (pf8f56b), which doesn't satisfy what react-intl and other dependencies request (^5.6.0).
```

nclsndr:
> I was doing some tests with upgrading `react-intl` from [v6 to v7](https://github.com/strapi/strapi/pull/26776) which barely has any breaking changes. `react-intl` requires minimum TypeScript v5.6. Wouldn't it be better if we upgrade first to latest v5, then move v5 -> v6 -> v7 ?

Interesting! Is there an actual TS feature required on react-intl or is it maintenance time window convenience? 
The plan was to not go latest on v5 before switching, yet not against it if it happens some deps actually relies on v5 novelties (which I still doubt)

nclsndr:
> The plan was to not go latest on v5 before switching, yet not against it if it happens some deps actually relies on v5 novelties

Did not took long… in the process of moving to `typescript@5` latest
- #26782

cc @unrevised6419

nclsndr:
## Build fix after TS 6 rebase

CI `build` was failing on `build:types` for two packages:

### `@strapi/strapi`
`resolve-development-config.test.ts` is co-located under `src/node/vite/` (not in `__tests__/`). With `"types": ["node"]` on the shared base tsconfig, Jest globals are no longer ambient during declaration emit — so the test file blew up with `Cannot find name 'jest'`.

**Fix:** exclude `**/*.test.ts` from `tsconfig.build.json` (same pattern as upload/email/plugins). This package is the only one with a co-located `.test.ts` outside `__tests__/` / `tests/`.

### `@strapi/upgrade`
`TS2578: Unused '@ts-expect-error'` on `fetch({ dispatcher })` — TS 6 / updated Node types now type the undici `dispatcher` option natively.

**Fix:** remove the obsolete suppression.

Verified locally with `yarn build` (all 36 projects pass).

---

*AI assisted comment*

unrevised6419:
https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

unrevised6419:
Note: To consider enabling `stableTypeOrdering` to see if everything works as expected.

packages/utils/tsconfig/base.json
```diff
@@ -15,6 +15,7 @@
     "noEmit": true,
     "types": [],
     "noUncheckedSideEffectImports": true,
-    "erasableSyntaxOnly": true
+    "erasableSyntaxOnly": true,
+    "types": ["node"]
```
unrevised6419:
```suggestion
    "types": []
```

nclsndr:
> Note: To consider enabling `stableTypeOrdering` to see if everything works as expected.

Good point @unrevised6419 👌 
- #27065

packages/core/review-workflows/admin/tsconfig.json
nclsndr: Those msw path aliases are gone now — leftover from an earlier revision. Current file only keeps `@tests/*` (same as `develop`) plus `rootDir` for the TS6 include layout.

Resolving as outdated.

---
_AI assisted_

packages/utils/tsconfig/base.json
nclsndr: Agreed — bad rebase left a duplicate `"types"` key (`[]` then `["node"]`). Kept `"types": []` to match #26699; leaf packages already declare `node`/`jest` locally. Also fixed the same duplicate-`compilerOptions` artifact in `packages/generators/generators/tsconfig.json`.

---
_AI assisted_

unrevised6419:
@nclsndr the added `rootDir` in `tsconfig.json` files is not needed because they have `noEmit: true`, only the `tsconfig.build.json` files have `rootDir` and `outDir`

nclsndr:
Fixed in 63232619e9: removed `rootDir` from the `noEmit` configs. Also cleaned two duplicate-key artifacts left by the rebase.

Verified with `yarn test:ts` and Prettier.

---
_AI assisted_

PR #25884, opened 2026-04-01
hotfix: fix broken CLI commands from 5.41.0

innerdvations:
🚨 NOTE: this is a hotfix and is targeting `main` not `develop`

### What does it do?

loads inquirer async since it is now ESM-only

### Why is it needed?

several CLI commands (the ones updated in this PR) no longer work

### How to test it?

create an experimental and see that the commands now work

### Related issue(s)/PR(s)


packages/core/strapi/src/cli/utils/get-inquirer.ts
```diff
@@ -0,0 +1,9 @@
+type InquirerAPI = typeof import('inquirer').default;
+
+let cached: Promise<InquirerAPI> | undefined;
+
+/** Loads Inquirer v9 via dynamic import (CJS-safe) and caches the default export. */
+export async function getInquirer(): Promise<InquirerAPI> {
+  cached ??= import('inquirer').then((m) => m.default);
```
nclsndr, line 7: @claude How does this work without `await`?

PR #25885, opened 2026-04-01
fix: handle 204 response before parsing JSON

### LINKED ISSUE #25465: Forgot password gives valid email regex error
### Node Version

22.17

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.30.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Strapi admin want to reset their password but password reset screen gives invalid email regex error in Safari browser.
The same screen works in google and other browsers. The same email gives Validation error on password reset screen in case of Safari.

### Steps to Reproduce

1. Open Safari browser
2. Open forgot password page it will prompt you to enter email
3. Enter Email like 'user@example13good.com'
4. it will give validation error that email is not matching regex while it is a valid email.
(THE SAME STEPS ARE WORKING FINE IN OTHER BROWSERS LIKE GOOGLE CROME)

### Expected Behavior

It should send password reset email in safari browser as well.

### Confirmation Checklist

akash-dabhi-qed:
Resolve forgot-password error on safari

### What does it do?

Adds an early return in `responseInterceptor` inside `getFetchClient.ts` to handle 204 responses before attempting to call `response.json()`.

https://github.com/user-attachments/assets/6527946a-1b68-427b-87a8-c54ea38af44a

### Why is it needed?

On Safari, calling `response.json()` on a response with an empty body (e.g. 204 No content) throws a non-`SyntaxError` with the message "The string did not match the expected pattern."  The existing catch block only handles `instanceof SyntaxError`, so Safari's error bypasses it and propagates as a mutation error, displaying the message in the UI even though the API call succeeded.

On Chrome and Firefox, the same call throws a `SyntaxError`, which is caught correctly.

The forgot-password endpoint (`POST /admin/forgot-password`) returns 204 on success, Safari users would see "The string did not match the expected pattern." after submitting a valid email, while the API had actually succeeded.

 The fix checks for status 204 explicitly before attempting JSON parsing, short-circuiting the problematic `response.json()` call entirely. 

### How to test it?

### Related issue(s)/PR(s)
Fixes #25465

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -268,6 +268,14 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
     response: Response,
     validateStatus?: FetchOptions['validateStatus']
   ): Promise<FetchResponse<TData>> => {
+    /**
+     * Handles responses with no body (e.g., 204 No Content) before attempting JSON parse.
+     * Safari may throw a non-SyntaxError when calling .json() on an empty body,
+     * which would bypass the SyntaxError-specific catch below.
+     */
+    if (response.status === 204) {
+      return { data: [] as unknown as TData, status: response.status };
+    }
     try {
       const result = await response.json();
 


packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -268,6 +268,14 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
     response: Response,
     validateStatus?: FetchOptions['validateStatus']
   ): Promise<FetchResponse<TData>> => {
+    /**
+     * Handles responses with no body (e.g., 204 No Content) before attempting JSON parse.
+     * Safari may throw a non-SyntaxError when calling .json() on an empty body,
+     * which would bypass the SyntaxError-specific catch below.
+     */
+    if (response.status === 204) {
+      return { data: [] as unknown as TData, status: response.status };
```
nclsndr, line 277: Why do we return an empty array by default here?

nclsndr (commented):
Thank you for the PR @akash-dabhi-qed!

packages/core/admin/admin/src/utils/getFetchClient.ts
akash-dabhi-qed: The empty array matches what chrome and firefox encounter (at line 303).

derrickmehaffy:
it looks like this was already fixed on develop. getFetchClient.ts:329 returns early on status === 204, and the Safari case this PR targets is handled by matching error.name === 'SyntaxError' instead of instanceof so going to mark as closed

PR #25886, opened 2026-04-01
feat(core/*): introduce strapi.ai namespace

nclsndr:
### What does it do?

- Adds a **core `strapi.ai` namespace** that is always registered: the core AI provider calls `strapi.add('ai', (strapi) => createAiNamespace(strapi))` on init (see `packages/core/core/src/providers/ai.ts`).
- Implements **`createAiNamespace`** so `strapi.ai.admin` is a **getter** that returns **`strapi.get('ai.admin')`** (`packages/core/core/src/services/ai.ts`).
- Introduces **`Modules.AI`** types (`AiNamespace`, `AiAdminService`) and wires **`Core.Strapi.ai`** to `Modules.AI.AiNamespace` (replacing the previous unused server-side `ai?: { enabled: boolean }` shape).
- **Moves server AI from EE into core admin**: routes, controllers, and **`createAiAdminService`** live under **`packages/core/admin/server/src/ai/`**; **`strapi.add('ai.admin', () => createAiAdminService({ strapi }))`** runs from **`packages/core/admin/server/src/register.ts`** (no longer from EE server `index`).
- **Chore (`AI` → `Ai`)**: aligns naming (`createAiAdminService`, contracts, tests, and related admin/upload/i18n touchpoints) with the new type names where it was part of the same refactor.
- Updates consumers to use **`strapi.ai.admin`** (token/config gating via **`isEnabled()`**) instead of **`strapi.get('ai')`**: admin AI controller, i18n AI localizations, upload AI metadata, CTB register, bootstraps, etc.
- Does **not** add MCP or `strapi.ai.mcp`. Browser **`window.strapi.ai`** behavior/typing is unchanged in intent (still separate from the server `strapi.ai` namespace).

### Why is it needed?

AI-related server code used to sit on a generic **`ai`** container key without a first-class **`strapi.ai`** API in core, and the implementation lived under EE even though admin AI routes are part of the admin package. That made the surface harder to extend (e.g. MCP under `strapi.ai` later) and blurred “namespace” vs “admin AI service”. This PR adds a **stable core namespace** with an **`admin`** slot, **relocates the admin AI service to core admin**, and keeps **EE/cms-ai gating** inside the service via **`isEnabled()`** rather than splitting registration across EE-only entrypoints.

### How to test it?

**Environment:** usual Strapi monorepo (Node/Yarn as documented for the repo).

**Build / typecheck (from repo root or package dirs as you prefer):**

1. `yarn build` in `packages/core/types` — `AiNamespace` / `Strapi.ai` resolve.
2. `yarn build` in `packages/core/core` — `strapi.ai` and `strapi.ai.admin` type-check.
3. `yarn build` in `packages/core/admin` — admin AI server code and EE admin consumers resolve.

**Sanity greps:**

- `rg "strapi\\.get\\('ai'\\)" --glob "*.ts"` — expect **no** matches (use **`strapi.ai.admin`** instead).
- `rg "createAIContainer" --glob "*.ts"` — expect **no** matches.

**Tests:**

- `packages/core/admin/server/src/ai/containers/__tests__/ai.test.ts`
- `packages/core/admin/server/src/ai/controllers/__tests__/ai.test.ts`
- `packages/core/core/src/services/__tests__/ai.test.ts`

**Runtime:** start an EE project with **cms-ai** enabled and exercise admin AI routes, i18n AI, and media AI metadata; confirm behavior matches expectations aside from internal registration and the `strapi.ai.admin` access pattern.


nclsndr:
@markkaylor do you have active branches (media-lib) that might collide with those changes?

markkaylor:
> @markkaylor do you have active branches (media-lib) that might collide with those changes?

Hey Nico, there are no active branches for media lib, everything is on develop under a future flag.

packages/core/admin/ee/server/src/ai/containers/ai.ts
```diff
@@ -17,56 +17,61 @@ const aiTokenCache = new Map<
...
+      throw new Error(`${errorPrefix.replace(/:$/, '')}. Check server logs for details.`);
     }
 
     const aiServerUrl = process.env.STRAPI_AI_URL || 'https://strapi-ai.apps.strapi.io';
 
-    if (!aiServerUrl) {
-      strapi.log.error(
-        `${ERROR_PREFIX} AI server URL not configured. Please set STRAPI_AI_URL environment variable.`
-      );
-      throw new Error('AI token request failed. Check server logs for details.');
```
markkaylor, line 67: we don't need this check anymore?

packages/core/admin/server/src/ai/services/ai.ts
```diff
@@ -17,56 +17,61 @@ const aiTokenCache = new Map<
   }
 >();
 
-const createAIContainer = ({ strapi }: { strapi: Core.Strapi }) => {
-  const getAIFeatureConfig = async () => {
+const createAiAdminService = ({ strapi }: { strapi: Core.Strapi }) => {
```
markkaylor, line 7: Since we are renaming this and considering it a service instead of a container should we update the folder to be `services` as well?

packages/core/admin/server/src/ai/containers/ai.ts
```diff
@@ -17,56 +17,61 @@ const aiTokenCache = new Map<
...
       strapi.log.error(
-        `${ERROR_PREFIX} No EE license found. Please ensure STRAPI_LICENSE environment variable is set or license.txt file exists.`
+        `${errorPrefix} No EE license found. Please ensure STRAPI_LICENSE environment variable is set or license.txt file exists.`
       );
-
-      throw new Error('AI token request failed. Check server logs for details.');
+      throw new Error(`${errorPrefix.replace(/:$/, '')}. Check server logs for details.`);
     }
 
     const aiServerUrl = process.env.STRAPI_AI_URL || 'https://strapi-ai.apps.strapi.io';
```
markkaylor: Nitpit/readability thing... Since the check below was removed I don't think this variable serves any real purpose now other than to be returned. Could be moved closer where it's used or defined inline on the return.

packages/core/admin/server/src/ai/services/ai.ts
```diff
@@ -193,10 +187,94 @@ const createAIContainer = ({ strapi }: { strapi: Core.Strapi }) => {
...
+        };
+      } catch (parseError) {
+        strapi.log.error(`${ERROR_PREFIX} Failed to parse AI server response`, parseError);
+        throw new Error('AI usage data request failed. Check server logs for details.');
+      }
+
+      return {
+        ...data.data,
+        subscription: data.subscription,
+      };
```
markkaylor, line 273: Could this not be returned in the try block? If it can it could be nice to eliminate the mutable data variable on line 231

packages/core/admin/server/src/ai/services/ai.ts
```diff
@@ -193,10 +187,94 @@ const createAIContainer = ({ strapi }: { strapi: Core.Strapi }) => {
...
+            subscriptionId: string;
+            planPriceId: string;
+            subscriptionStatus: string;
+            isActiveSubscription: boolean;
+            cmsAiEnabled: boolean;
+            cmsAiCreditsBase: number;
+            cmsAiCreditsMaxUsage: number;
+            currentTermStart: string;
+            currentTermEnd: string;
+          };
```
markkaylor, line 263: kind of surprised this is the only place this type would be needed

packages/core/admin/server/src/ai/services/ai.ts
```diff
@@ -193,10 +187,94 @@ const createAIContainer = ({ strapi }: { strapi: Core.Strapi }) => {
     }
   };
 
+  const getAiUsage = async () => {
```
markkaylor, line 208: Is this func new or moved? Weren't we already doing this somewhere?

packages/core/admin/server/src/register.ts
```diff
@@ -10,6 +11,8 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
   strapi.get('auth').register('admin', adminAuthStrategy);
   strapi.get('auth').register('content-api', apiTokenAuthStrategy);
 
+  strapi.add('ai.admin', () => createAiAdminService({ strapi }));
```
markkaylor, line 14: Should the service only be created if AI is enabled / on the license?

packages/core/content-type-builder/server/src/register.ts
```diff
@@ -2,8 +2,7 @@ import type { Core } from '@strapi/types';
 import { CSP_DEFAULTS, extendMiddlewareConfiguration } from '@strapi/utils';
 
 export default async ({ strapi }: { strapi: Core.Strapi }) => {
-  const aiEnabledConfig = strapi.config.get('admin.ai.enabled') !== false;
-  const isAIEnabled = aiEnabledConfig && strapi.ee.features.isEnabled('cms-ai');
+  const isAIEnabled = strapi.ai.admin.isEnabled();
```
markkaylor: It looks like you prefer `isAiX` do you want to update this here as well to be consistent?

markkaylor (commented):
Nice, everything still seems to be working as expected

innerdvations (dismissed):
LGTM!

AI review caught a few related minor things in code that you didn't actually touch, which I put in `chore/ai-namespace-cleanup` -- if you want to merge that in here I can approve again quickly, otherwise I can just put up a quick follow-up to fix the incorrect comments

innerdvations:
You'll need to update the CLI test snapshots for the new endpoints -- make sure you have the fix from this morning that makes it easier to update them.

packages/core/admin/server/src/register.ts
nclsndr: I decided to have the service registered all the time so it statically always resemble the same type definition. Having it conditionally registered while maintaining type coherence meant we had to declare a stub service for the case it would not be there. 
I found more convenient to place guards within the service runtime rather than working around the conditionality of the service being defined.

packages/core/admin/ee/server/src/ai/containers/ai.ts
nclsndr: Since we placed the hardcoded fallback right into here, this check was already always positive, hence useless. Unless I missed something in the process?

packages/core/admin/server/src/ai/services/ai.ts
nclsndr: This function used to leave in `packages/core/admin/ee/server/src/ai/controllers/ai.ts`, I've been refactoring it here because it felt being a piece of business logic we'd prefer closer to service than controller edge.

packages/core/admin/server/src/ai/services/ai.ts
nclsndr: I guess this is because it was defined in controller, hence the type is only duck typing locally.
Overall, imho, this should be part of an externalized AI server definitions package we don't have at the moment.

packages/core/admin/server/src/ai/services/ai.ts
nclsndr: Maybe, since my goal was refactor only I did not challenge the established logic by any mean. I guess we'd prefer another PR to do it.

markkaylor: oh gotcha makes sense, sorry sometimes tough to see what's new vs moved.

nclsndr:
> AI review caught a few related minor things in code that you didn't actually touch, which I put in `chore/ai-namespace-cleanup`

Thx @innerdvations! I just gathered the feedback and applied them here

nclsndr:
> You'll need to update the CLI test snapshots for the new endpoints

This is concerning to me, I should not change routes 🤔. I'll wait for the CI to confirm this again.

packages/core/admin/server/src/ai/services/ai.ts
nclsndr: true

markkaylor (dismissed):
Great thanks! Do you need me to test it again?

nclsndr:
> Great thanks! Do you need me to test it again?

Thank you @markkaylor! Don't bother with testing, my ping was about keeping you in the loop with the answers to check whether I would have missed something 🙌

nclsndr:
> > You'll need to update the CLI test snapshots for the new endpoints
> 
> This is concerning to me, I should not change routes 🤔. I'll wait for the CI to confirm this again.

Ok the actual snapshots update is done. They were necessary because the CLI tests are taking a bare CE setup where the AI routes would not get registered by default. Now does.

innerdvations (dismissed):
I think this is good now!

Might be worth a release note about the change to `strapi.ai` let people know it has changed in case they're already checking it even though it was never documented (afaik)

packages/core/admin/server/src/ai/controllers/ai.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (strapi.ai.admin.isEnabled() === false) {
+      return ctx.notFound();
+    }
+
+    const aiFeatureConfig = await strapi.ai.admin.getAiFeatureConfig();
+
+    ctx.body = {
+      data: aiFeatureConfig,
+    } satisfies GetAiFeatureConfig.Response;
+  },
```
butcherZ, line 61: maybe worth adding a try/catch here? :)

packages/core/admin/server/src/ai/services/ai.ts
```diff
@@ -17,55 +17,78 @@ const aiTokenCache = new Map<
   }
 >();
```
butcherZ: would it better if this is declared inside of the `createAiAdminService` factory? :D

butcherZ (dismissed):
Nothing major, just two small suggestions, the rest LGTM!

packages/core/admin/server/src/ai/controllers/ai.ts
nclsndr: agreed, done

packages/core/admin/server/src/ai/services/ai.ts
nclsndr: well spotted! I did not notice it was outside of the closure.

butcherZ (approved):

PR #25890, opened 2026-04-01
fix(document-service): preserve self-referential relations during publish and discard

### LINKED ISSUE #24936: Can't create a relation between an entry and itself

butcherZ:
### What does it do?

Adds a new `self-referential-relations` utility that captures and restores
self-referential join table rows across publish and discard-draft operations.

- `load()` queries join tables for rows where both sides belong to the set of
  entries being published/discarded, before those entries are deleted
- `sync()` remaps the old entry IDs to new entry IDs (via locale) and
  re-inserts the rows after the new entries are created
- Both the publish and discard-draft flows in `repository.ts` are covered

### Why is it needed?

When an entry with a self-referential relation (e.g. a category with a `parent`
pointing to itself) is published or discarded, the relation is silently dropped.

The root cause is in `bidirectional-relations.ts`: it skips any relation where
`model.uid === uid` (self-referential), so the join table rows are never captured
before the old entry is deleted. When the old entry is cascade-deleted, the join
rows go with it. The new entry is then created with no relation to restore.

### How to test it?

1. Create a content type with a self-referential relation field (e.g. `category`
   with a `parent` field targeting itself), with Draft & Publish enabled
2. Create an entry (Category A) and set its `parent` relation to itself
3. Publish the entry
4. Fetch the published entry via API or check the published view in the admin —
   the `parent` relation should be populated (was: empty before this fix)
5. Discard draft — the relation should still be present on the restored draft

To run the integration tests:
```
yarn test:api -- --testPathPattern="self-referential-relations"
```

### Related issue(s)/PR(s)

Fix #24936
fix CMS-496


innerdvations:
This seems ok for the specific bug mentioned in the linked issue, but misses bidirectional join-table duplication, idempotent sync with createEntry, and realistic CM/API coverage. That is, the fix in the issue can be “enough” for one scenario while still failing or behaving oddly when both sides of a self-ref are present, when the join row is already written, or when you exercise publish/unpublish/republish and unidirectional self-refs through the content manager. I pushed an example (written by AI, so give it a careful look for the idea, don't just merge it in 😆) to `fix/self-referential-relation-publish-inverse-dedupe-idempotent`

packages/core/core/src/services/document-service/utils/self-referential-relations.ts
```diff
@@ -0,0 +1,131 @@
...
+          return {
+            ...omit(strapi.db.metadata.identifiers.ID_COLUMN, relation),
+            [sourceColumn]: newSourceId,
+            [targetColumn]: newTargetId,
+          };
+        })
+        .filter(Boolean);
+
+      if (newRelations.length > 0) {
+        await trx.batchInsert(joinTable.name, newRelations as any[], 1000);
```
jhoward1994: I used this in my PR the other day https://github.com/strapi/strapi/pull/25764

```suggestion
		const batchSize = strapi.db.dialect.getBatchInsertSize();
        await trx.batchInsert(joinTable.name, newRelations as any[], batchSize);
```

packages/core/core/src/services/document-service/utils/__tests__/self-referential-relations.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+      expect(result).toHaveLength(0);
+    });
+  });
+
+  describe('sync', () => {
+    it('should remap and insert self-referential relations with new IDs', async () => {
+      (global as any).strapi = {
+        db: {
+          metadata: { identifiers: { ID_COLUMN } },
```
jhoward1994, line 99: Do we need something like this here ?

`dialect: { getBatchInsertSize: jest.fn().mockReturnValue(BATCH_SIZE) },`

jhoward1994:
I think this diff should fix the API test failures

```
diff --git a/tests/api/core/content-manager/api/self-referential-relations.test.api.ts b/tests/api/core/content-manager/api/self-referential-relations.test.api.ts
index 220b1243b9..217882550f 100644
--- a/tests/api/core/content-manager/api/self-referential-relations.test.api.ts
+++ b/tests/api/core/content-manager/api/self-referential-relations.test.api.ts
@@ -60,6 +60,19 @@ const getCategory = async (
   return res.body.data as CategoryEntry;
 };
 
+const getRelation = async (
+  documentId: string,
+  field: 'parent' | 'children' | 'related',
+  status: 'draft' | 'published' = 'draft'
+): Promise<CategoryRelation[]> => {
+  const res = await rq({
+    method: 'GET',
+    url: `/content-manager/relations/api::category.category/${documentId}/${field}`,
+    qs: { status },
+  });
+  return (res.body.results ?? (res.body.data ? [res.body.data] : [])) as CategoryRelation[];
+};
+
 describe('CM API - Self-referential relations with Draft & Publish', () => {
   beforeAll(async () => {
     await builder.addContentType(category).build();
@@ -97,8 +110,9 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
     });
 
     // Verify draft has the self-relation
-    const draftBefore = await getCategory(catA.documentId, 'draft');
-    expect(draftBefore.parent).toMatchObject({ documentId: catA.documentId });
+    const draftParent = await getRelation(catA.documentId, 'parent', 'draft');
+    expect(draftParent).toHaveLength(1);
+    expect(draftParent[0]).toMatchObject({ documentId: catA.documentId });
 
     // Publish
     await rq({
@@ -107,8 +121,9 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
     });
 
     // Verify published version has the self-relation
-    const published = await getCategory(catA.documentId, 'published');
-    expect(published.parent).toMatchObject({ documentId: catA.documentId });
+    const publishedParent = await getRelation(catA.documentId, 'parent', 'published');
+    expect(publishedParent).toHaveLength(1);
+    expect(publishedParent[0]).toMatchObject({ documentId: catA.documentId });
   });
 
   test('Self-referential relation between two entries is preserved after publish', async () => {
@@ -131,8 +146,9 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
     });
 
     // Verify published catB has the relation to catA
-    const published = await getCategory(catB.documentId, 'published');
-    expect(published.parent).toMatchObject({ documentId: catA.documentId });
+    const publishedParent = await getRelation(catB.documentId, 'parent', 'published');
+    expect(publishedParent).toHaveLength(1);
+    expect(publishedParent[0]).toMatchObject({ documentId: catA.documentId });
   });
 
   test('Self-referential relation (entry to itself) is preserved after discard draft', async () => {
@@ -155,7 +171,8 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
     });
 
     // Verify the reverted draft still has the self-relation
-    const draft = await getCategory(catA.documentId, 'draft');
-    expect(draft.parent).toMatchObject({ documentId: catA.documentId });
+    const draftParent = await getRelation(catA.documentId, 'parent', 'draft');
+    expect(draftParent).toHaveLength(1);
+    expect(draftParent[0]).toMatchObject({ documentId: catA.documentId });
   });
 });


```

Adzouz (dismissed):
Hey @butcherZ, thanks a lot 🙏
It's very similar to https://github.com/strapi/strapi/pull/25459 PR but you added the tests and your solution is a bit cleaner (especially on the typing side + the joining table unique constraint issue guard) so I'm good with merging this one once the pipelines pass!
I also tested a it's working as expected 🙌

Adzouz (approved):

PR #25893, opened 2026-04-01
feat(email): replace sendmail package with nodemailer in provider

innerdvations:
Note: I do not want to merge this until we have confirmation from a few users using this package in advanced configurations that it still works exactly as before

### What does it do?

- Replaces the unmaintained **`sendmail` npm package** (and its **`mailcomposer` / `buildmail`** transitive dependencies) in **`@strapi/provider-email-sendmail`** with a **Nodemailer**-based implementation.
- Preserves the same **Strapi-facing API**: `provider: 'sendmail'`, `providerOptions`, and `settings` (`defaultFrom`, `defaultReplyTo`) behave as before for documented options.
  - Direct SMTP per recipient domain, with grouping aligned with **`sendmail@1.6.1`**
  - DNS **MX** resolution when not in dev mode; **MX priority** ordering; try peers in order; optional **`smtpHost`** after MX records
  - **`devPort`** / **`devHost`** for local capture; outbound port uses **`devPort`** in dev mode and **`smtpPort`** otherwise
  - **DKIM** signing via Nodemailer
  - **Unit and integration tests** to verify that behaviour
- In **`@strapi/email`**, logs a **one-time `strapi.log.warn` in non-production** when the plugin uses the **`sendmail`** provider, nudging projects with a dedicated SMTP relay toward **`@strapi/provider-email-nodemailer`**

### Why is it needed?

We have to get the `sendmail` package out of the project, it's responsible for at least 3 high security warnings (which actually only apply when using this provider). The only alternative is a hard deprecation where we remove the sendmail provider and require users to install it separately. This way we can still continue to maintain it and keep it secure for the duration of Strapi 5.

### How to test it?

**Automated (this package)**

From the monorepo root:

```bash
yarn workspace @strapi/provider-email-sendmail test:unit
```

**Manual (Strapi app)**

1. Use a Strapi app (e.g. `examples/getstarted` or a generated app) with the email plugin configured for **`provider: 'sendmail'`** and, for local testing, **`providerOptions`** such as **`devPort`** / **`devHost`** pointing at a local SMTP capture server (MailHog, etc.).
2. Trigger an email (password reset, or the email test flow in admin if available).
3. Confirm the message is accepted by your capture server and headers/body look correct.
4. Start the app with **`NODE_ENV=development`**: confirm the **soft nudge** log appears once when using the sendmail provider; with **`NODE_ENV=production`**, confirm it does **not** appear.

**Regression spot-check**

- Existing **`config/plugins`** using **`sendmail`** and the same **`providerOptions`** surface (including **DKIM**, **devPort**, **smtpHost**) should still send mail in typical setups; MIME may not be byte-identical to the old library (Nodemailer vs mailcomposer), but delivery semantics should match the documented behavior.

### Related issue(s)/PR(s)

Fixes CMS-551


innerdvations:
@derrickmehaffy Tagging you in case you know of some projects relying on the sendmail package (and in particular, any unusual configurations) that would be willing to test this out

innerdvations:
@claude review

packages/providers/email-sendmail/src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+      };
+  /** Development / testing: connect to `devHost:devPort` instead of resolving MX. */
+  devPort?: number | boolean;
+  devHost?: string;
+  /** SMTP port for outbound delivery (default 25). */
+  smtpPort?: number;
+  /** Extra SMTP host to try after MX records (same as legacy `sendmail` package). */
+  smtpHost?: string | number;
+  rejectUnauthorized?: boolean;
+  autoEHLO?: boolean;
```
jhoward1994, line 27: Is this used ?

innerdvations: No, but it should be, so that we don't break for anyone using the option(s). I'll see how to add an implementation of anything still missing.

derrickmehaffy, line 27: Following up on @jhoward1994's earlier thread — `autoEHLO` is declared here but never read anywhere in the new implementation, so right now it's a silent no-op. Either wire it through to the Nodemailer transport (nodemailer exposes `opportunisticTLS` and auth-related options that roughly correspond to what legacy sendmail did here) or drop it from `ProviderSendmailOptions` and note the removal in a README 'migration from legacy sendmail' section. A type that accepts the option but silently ignores it is the worst of both worlds for users who had it set.

packages/providers/email-sendmail/src/types.ts
```diff
@@ -0,0 +1,33 @@
...
+    warn?: (...args: unknown[]) => void;
+    error?: (...args: unknown[]) => void;
+  };
+  silent?: boolean;
+  dkim?:
+    | boolean
+    | {
+        privateKey: string;
+        keySelector?: string;
+      };
```
derrickmehaffy, line 18: The `boolean` branch of this union isn't actually honored by `sendDirectSmtp` — only `dkim: { privateKey, ... }` results in signed mail; `dkim: true` silently does nothing (see `direct-smtp.ts:141-148`). Either narrow to `{ privateKey: string; keySelector?: string } | false | undefined`, or emit a one-time `logger.warn` when `dkim === true` so users with bad config aren't debugging silently-unsigned mail.

packages/providers/email-sendmail/src/direct-smtp.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  for (const domain of Object.keys(groups)) {
+    const domainRecipients = groups[domain];
+    let hosts: Array<{ exchange: string }>;
+    try {
+      hosts = await resolveMxHosts(domain, providerOptions);
+    } catch (err) {
+      logger.error('Sendmail provider: MX resolution failed', err);
+      throw err instanceof Error ? err : new Error(String(err));
+    }
```
derrickmehaffy, line 175: This rethrows on MX resolution failure, which aborts the whole send — including sibling domains whose MX is fine. That's inconsistent with the per-host fallback loop immediately below (lines 178-203) which tolerates transport errors and keeps trying. Worth deciding between 'fail fast on DNS' (and documenting it) or 'continue + collect errors like the transport path' (and changing behavior here). A test with one good and one bad domain would pin down the intended contract — the current `mx-fallback.test.ts` mixed-domain case only exercises good DNS + bad transport.

packages/providers/email-sendmail/src/direct-smtp.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (sent) {
+      anyDomainDelivered = true;
+    } else {
+      logger.error(`Sendmail provider: failed to deliver for domain ${domain}`, lastError);
+    }
+  }
+
+  if (!anyDomainDelivered) {
+    throw new Error('Failed to deliver mail for all recipient domains');
+  }
```
derrickmehaffy, line 209: The final thrown error has no reference to the per-host / per-domain failures — `lastError` is only logged. Attaching it as `throw new Error('Failed to deliver mail for all recipient domains', { cause: lastError })` (Node 16+) would let consumers see the root cause without having to scrape logs. Low-effort, meaningful debugging win.

packages/providers/email-sendmail/README.md
```diff
@@ -23,14 +23,25 @@ npm install @strapi/provider-email-sendmail --save
...
+| providerOptions          | object            | Options below (same names as the legacy `sendmail` package where applicable).                                                                                                                    | no       | {}        |
+| settings                 | object            | Settings                                                                                                                                                                                         | no       | {}        |
+| settings.defaultFrom     | string            | Default sender mail address                                                                                                                                                                      | no       | undefined |
+| settings.defaultReplyTo  | string \| array   | Default address or addresses the receiver is asked to reply to                                                                                                                                   | no       | undefined |
```
derrickmehaffy, line 34: `defaultReplyTo` is documented here as `string | array` but typed as `string | undefined` in `types.ts:32`. Pick one — if arrays genuinely work (Nodemailer does accept array `replyTo`), widen the type; otherwise tighten the docs to `string`.

derrickmehaffy (commented):
Nice work on this. Behavior is faithfully mirrored to legacy `sendmail@1.6.1` with explicit comments on each quirk, and I ran `yarn workspace @strapi/provider-email-sendmail test:unit` locally — all 41 tests pass.

A few inline comments. None are blockers on the runtime behavior; (1) and (2) are the ones I'd most like to see resolved before merge. A separate top-level comment will follow with a README restructure suggestion.

derrickmehaffy:
The README today reads like generic "configure your email provider" docs, but the new bootstrap warning already signals (correctly) that most users shouldn't be using this provider. The README hasn't caught up to that reframing, and the **"how do I test this locally"** story is essentially absent — new users will hit a DNS MX failure on their first `send()` and have no path forward from the docs.

Suggesting a restructure: (1) lead with **"When to use this provider"** so users self-select before copying the config; (2) direct everyone else to the correct first-party Strapi provider for their sending service; (3) add a **"Local development"** section with real, non-Docker options.

The motivating observation is that Strapi already ships four first-party providers beyond this one (`amazon-ses`, `mailgun`, `sendgrid`, `nodemailer`), and `sendmail` is almost never the right choice except for (a) self-hosted MTAs delivering directly — rare — or (b) local capture during development. Anyone reaching for "Gmail" / "SES" / etc. should be pointed at the matching provider in the monorepo, not left to discover on their own that this one can't STARTTLS or authenticate.

For local dev the two options I'd recommend are **maildev** (`npx maildev` — zero install for Strapi users since they already have Node) and **Mailpit** (standalone Go binary, nicer UI, actively maintained). I'd explicitly **not** recommend MailHog — the maintainer himself confirmed in [mailhog/MailHog#442](https://github.com/mailhog/MailHog/issues/442) that he can't maintain it and suggests Mailpit as the successor; Laravel made the same switch in its defaults.

Drafted replacement content below for convenience — feel free to rework the wording, the structural point matters more than the exact prose.

<details>
<summary>Drafted README content</summary>

````md
## When to use this provider

Use `@strapi/provider-email-sendmail` only if:

- Your server is configured as its own MTA (public IP, PTR, SPF/DKIM) and
  delivers directly to recipient MX servers, **OR**
- You're testing locally against an SMTP capture tool (see
  [Local development](#local-development) below)

For any other case, use the first-party Strapi provider for your sending service:

| Service                                               | Strapi provider                     |
| ----------------------------------------------------- | ----------------------------------- |
| Amazon SES                                            | `@strapi/provider-email-amazon-ses` |
| Mailgun                                               | `@strapi/provider-email-mailgun`    |
| SendGrid                                              | `@strapi/provider-email-sendgrid`   |
| Gmail, Postmark, Resend, or any authenticated SMTP    | `@strapi/provider-email-nodemailer` |

This provider does **not** support SMTP authentication or STARTTLS, so it
cannot be used with any of the services above.

## Local development

For local testing, point this provider at an SMTP capture tool — a local
program that accepts any mail sent to it and shows captured messages in a
web UI. Nothing leaves your machine.

### Option A: maildev (no install, uses your existing Node)

```bash
npx maildev
# SMTP on :1025, web UI on http://localhost:1080
```

### Option B: Mailpit (standalone binary, nicer UI)

```bash
# macOS
brew install mailpit && mailpit

# Linux / Mac install script
sudo sh < <(curl -sL https://raw.githubusercontent.com/axllent/mailpit/develop/install.sh)
mailpit

# Windows: download from https://github.com/axllent/mailpit/releases
```

Mailpit defaults to SMTP `:1025` and web UI `http://localhost:8025`.

### Strapi config

```js
// config/plugins.js
  email: {
    config: {
      provider: 'sendmail',
      providerOptions: {
        devHost: '127.0.0.1',
        devPort: 1025,
      },
      settings: {
        defaultFrom: 'noreply@example.com',
        defaultReplyTo: 'noreply@example.com',
      },
    },
  },
});
```

Trigger a password reset or the admin test-email flow, then open the capture
UI to inspect the result.
````

</details>

**Review by derrickmehaffy — CHANGES_REQUESTED** [review]

**Comment by derrickmehaffy** [comment]
**Docs update required** — tagging with `flag: documentation`.

The CMS PR replaces the unmaintained `sendmail` npm package with a Nodemailer-based direct-SMTP implementation in `@strapi/provider-email-sendmail`. Provider name and config surface are preserved, so most docs stay accurate — but one spot in the feature docs references the old npm package directly and needs a rewrite.

### Must update

**[`docusaurus/docs/cms/features/email.md`, line 83](https://github.com/strapi/documentation/blob/main/docusaurus/docs/cms/features/email.md#L83)** — the current paragraph links `<ExternalLink to="https://www.npmjs.com/package/sendmail" ...>` to the abandoned npm package this PR is removing, and describes the provider as "not production-ready in the default configuration" without explaining why. After this PR, the reason is that the provider does **direct MX delivery** (no SMTP auth, no STARTTLS), which requires MTA setup or a local SMTP capture tool.

Suggested replacement:

```md
[`@strapi/provider-email-sendmail`](https://www.npmjs.com/package/@strapi/provider-email-sendmail)
is the default email provider in the Strapi Email feature. It performs direct SMTP delivery by
resolving MX records for each recipient domain — useful for local development against an SMTP
capture tool, or for servers configured as their own MTA. It does not support SMTP authentication
or STARTTLS, so it cannot be used with hosted services like Gmail, SES, Mailgun, or SendGrid;
for those, use a [dedicated provider](#providers) instead. See the provider README for local
development setup with capture tools like `maildev` or Mailpit.
```

### Verified OK, no changes needed

- `cms/backend-customization/examples/services-and-controllers.md:313` — already points to `@strapi/provider-email-sendmail`.
- `cms/providers.md:14` — generic "Amazon SES instead of Sendmail" example, still accurate.
- `cms/features/email.md:91` — config table lists `sendmail` as the provider default; the provider name hasn't changed.
- `cloud/advanced/email.md` — no sendmail references.

### Gaps worth addressing separately (not blocking this PR)

- No "Local development" section in email docs mentioning capture tools (MailHog/Mailpit/maildev). The updated provider README will cover this, but a one-paragraph pointer from `cms/features/email.md` would help discoverability.
- No dedicated `email-sendmail.md` page (one exists for Nodemailer). Lower priority.
- Pre-existing: the `providerOptions.apiKey` row in the config table at `cms/features/email.md:91` describes a SendGrid-specific option as if it were top-level. Unrelated to this PR, worth a separate ticket.

Recommend waiting for this PR to merge before landing the docs change, since it's still iterating under `flag: don't merge`.

packages/providers/email-sendmail/src/direct-smtp.ts
```diff
@@ -0,0 +1,208 @@
...
+export async function sendDirectSmtp(
+  mail: SendMailOptions,
+  providerOptions: ProviderSendmailOptions
+): Promise<void> {
+  const logger = createLogger(providerOptions);
+  const smtpPort = getOutboundSmtpPort(providerOptions);
+
+  const fromHeader = String(mail.from || '');
+  const fromAddr = extractEmail(fromHeader);
+  const srcHost = String(getHostFromAddress(fromAddr));
```
jhoward1994: `getHostFromAddress` returns `string | undefined`, so `String(undefined)` here produces the literal string `"undefined"` ?

Maybe we can do something like:
```
const fromHost = getHostFromAddress(fromAddr);
const srcHost = fromHost || osHostname() || 'localhost';
```

packages/providers/email-sendmail/src/direct-smtp.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    if (sent) {
+      anyDomainDelivered = true;
+    } else {
+      logger.error(`Sendmail provider: failed to deliver for domain ${domain}`, lastError);
+    }
+  }
+
+  if (!anyDomainDelivered) {
```
jhoward1994, line 207: So I think this means we treat it as a success if at least one domain has been delivered? Maybe thats too lenient? Does that mean if `to: 'a@one.com, b@two.com'` fails to deliver to two.com we'd never know?

Would it be worth tracking per domain outcomes ?

packages/providers/email-sendmail/src/direct-smtp.ts
innerdvations: oops, good catch!

packages/providers/email-sendmail/src/direct-smtp.ts
innerdvations: Ok, I did some investigation and it looks like sendmail loops with a `for (domain in groups)` and Strapi previously wrapped it in a single promise, so whichever domain's callback ran first was the result that counted -- which was kind of a bug.

So for the goal of a drop-in replacement, this is still a step up from the previous Strapi behaviour, but leaves the shoddy original sendmail contract unchanged.

derrickmehaffy (approved):

jhoward1994 (approved):

PR #25909, opened 2026-04-02
fix(document-service): preserve relations during publish

jhoward1994:
## What does it do?

The `getDeepPopulate` function in the document service now **populates all non-morph relations**, regardless of their `visible` attribute. 

Previously, it only included relations marked as visible or special admin fields (`createdBy` / `updatedBy`). 

This ensures that internal actions like publish, discard-draft, and clone retain all relational links, even if those links are hidden from the admin UI.

Additionally, a new API test verifies this behavior: it creates an Article with a hidden inverse one-to-one (legacy) relation and checks that updating the published Article (changing only its title) does **not** break the legacy relation.

---

## Why is this needed?

Deep populating relations is required when building the payload for publishing or similar internal operations. 

If an inverse relation is `visible: false`, the previous logic would skip it—so, unless the API client manually included the hidden field, the reference could be lost on republish or update. This solves the problem where hidden relations (e.g., inverse oneToOne) could be unintentionally broken.

---

## Testing

Run the document service relation API tests, or specifically:
```
yarn test:api tests/api/core/strapi/document-service/relations/invisible-relation-republish.test.api.ts
```
This file sets up a test schema with a hidden (`visible: false`) legacy relation on the Article type and checks that the relation survives republishing.

---

## Related issues/PRs

fixes CMS-150


tests/api/core/strapi/document-service/relations/invisible-relation-republish.test.api.ts
```diff
@@ -0,0 +1,120 @@
...
+  const schema = JSON.parse(fs.readFileSync(schemaPath, 'utf8'));
+  if (!schema.attributes?.legacy) {
+    throw new Error(`Expected legacy attribute in ${schemaPath}`);
+  }
+
+  schema.attributes.legacy.visible = false;
+  fs.writeFileSync(schemaPath, `${JSON.stringify(schema, null, 2)}\n`, 'utf8');
+};
+
+describe('Document Service invisible inverse oneToOne (CMS-150)', () => {
```
markkaylor: We probably don't need the ticket number here 🙃

markkaylor (dismissed):
Looks good to me, I tested and looks like it's working!

tests/api/core/strapi/document-service/relations/invisible-relation-republish.test.api.ts
jhoward1994: Woops thank you!

tests/api/core/strapi/document-service/relations/invisible-relation-republish.test.api.ts
```diff
@@ -0,0 +1,119 @@
...
+  schema.attributes.legacy.visible = false;
+  fs.writeFileSync(schemaPath, `${JSON.stringify(schema, null, 2)}\n`, 'utf8');
+};
+
+describe('Document Service invisible inverse oneToOne', () => {
+  beforeAll(async () => {
+    await builder.addContentTypes([articleBaseModel, legacyModel]).build();
+    setArticleLegacyInvisible();
+    strapi = await createStrapiInstance();
+  });
```
innerdvations:
```suggestion
describe('Document Service invisible inverse oneToOne', () => {
  beforeAll(async () => {
    await builder.addContentTypes([articleBaseModel, legacyModel]).build();

    // Set `visible: false` on the inverse relation via the content-type-builder service
    const articleSchema = await getContentTypeSchema('article');
    if (!articleSchema?.attributes?.legacy) {
      throw new Error('Expected legacy attribute on article after setup');
    }
    articleSchema.attributes.legacy.visible = false;
    await modifyContentType(articleSchema);

    strapi = await createStrapiInstance();
  });
```

We shouldn't modify files in the test app, that's hard to clean up and adds a dependency on the structure staying the same and could introduce hard to debug test issues. This seems to work though

innerdvations (changes requested):
Overall LGTM, just one fix on the test

innerdvations (approved):

PR #25912, opened 2026-04-02
test: unit test transfer CLI command functionality

jhoward1994:
### What does it do?

This introduces **CLI-level tests** for the `strapi transfer` command, located in `packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.cli.test.ts`.
- When `--from`, `--from-token`, and `--force` (or `--to`, `--to-token`, and `--force`) are used, **no** Inquirer prompts are triggered and the action function receives the parsed options, including `URL` instances for any remote URLs.
- The `--exclude` and `--only` flags are correctly passed through (when remote flags are provided).
- If remote URL flags are omitted, the CLI runs interactively using the mocked prompts, and eventually invokes the action with the resolved options.

There was previously an issue where CLI flags like `--from` and `--from-token` did not skip interactive prompts as expected. 

This behavior looks like it was corrected in [#24879](https://github.com/strapi/strapi/pull/24879) by only running prompts if neither `--from` nor `--to` are set, and refactoring how the transfer direction is determined. 

### Testing

From the root of the repository, run:

```bash
yarn test:unit packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.cli.test.ts
```

**Related issues/PRs:**  

fixes CMS-449
Note: This is a test-only change.

### DIFF
--- a/packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.cli.test.ts
+++ b/packages/core/strapi/src/cli/commands/transfer/__tests__/transfer.cli.test.ts
@@ -1,0 +1,160 @@
+import { createCLI } from '../../../index';
+import { getInquirer } from '../../../utils/get-inquirer';
+import transferAction from '../action';
+
+jest.mock('@strapi/cloud-cli', () => ({
+  buildStrapiCloudCommands: jest.fn().mockResolvedValue(undefined),
+}));
+
+jest.mock('../action', () => ({
+  __esModule: true,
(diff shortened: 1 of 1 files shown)


innerdvations (approved):
LGTM, as a side note I'm also adding these tests: [tests/cli/tests/strapi/data-transfer/pull-remote.test.cli.js](https://github.com/strapi/strapi/pull/23479/changes#diff-9829980042a39762ff5db4dfee1160d737f5a1d591c036c8d47cb1aed38894b4)

PR #25918, opened 2026-04-03
fix(data-transfer): pull stage totals, batch limits, and async writable backpressure

innerdvations:
### What does it do?

- Adds **stage total estimates** for entities, links, and configuration on the **pull** path (mirroring existing assets behavior), wired through `getStageTotals` and the pull handler’s `start` response.
- Extends the **link query** with `countAllForUid` so link totals can be computed in line with link streaming scope.
- Introduces **shared limits** for non-asset WebSocket batches (`transfer-stream-batch.ts`) and uses them when **flushing** pull streams (byte cap + max items per message).
- Adds **`writeWritableAsync`** so writes to `Writable` streams respect **backpressure** (write callback + `drain` when `write()` returns false); **push** and **remote-source** use it instead of a callback-only helper. **File** and **directory** source providers use it when piping backup/tar output into a `PassThrough`, and **serialize tar `onReadEntry` work** before `outStream.end()` so `pipeline` completion does not race async JSONL draining.
- Adds **unit tests** for the new estimators and for `writeWritableAsync`.

### Why is it needed?

- **Progress / ETA** for pull transfers was only well-defined for assets; entities, links, and configuration lacked comparable totals.
- **Pull batching** needed explicit bounds so large JSON payloads do not overwhelm the WebSocket path.
- **Push and file/directory sources** could **ignore backpressure** or **end the stream too early** relative to async producers, risking memory pressure or “stream not writable” races.

### How to test it?

experimental: `0.0.0-experimental.24685bf33d0f8764aa7e9bdfc756519767546ef0` 

- From the repo root: `yarn workspace @strapi/data-transfer test:unit` (all data-transfer unit tests should pass).
- Optional manual check: run a **pull** (or your usual **transfer** smoke flow) against a seeded app and confirm non-asset stages show sensible totals and the transfer completes; optionally exercise **push** with a non-trivial dataset to confirm no regressions under load.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/23479

### DIFF
--- a/packages/core/data-transfer/src/directory/providers/source/index.ts
+++ b/packages/core/data-transfer/src/directory/providers/source/index.ts
@@ -12,6 +12,7 @@ import type { IAsset, IMetadata, ISourceProvider, ProviderType } from '../../../
 import type { IDiagnosticReporter } from '../../../utils/diagnostic';
 
 import * as utils from '../../../utils';
+import { writeWritableAsync } from '../../../utils/write-writable-async';
 import { ProviderInitializationError, ProviderTransferError } from '../../../errors/providers';
 import { unknownPathToPosix } from '../../../file/providers/source/utils';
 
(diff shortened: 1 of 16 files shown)


innerdvations:
This actually seems to make things worse rather than fix anything as expected

PR #25921, opened 2026-04-03
enhancement(data-transfer): improve progress UX

innerdvations:
### What does it do?

- **Data transfer providers**
  - Adds optional `onTransferPhase` on local and remote **destination** providers.
  - **Local:** emits step messages during `beforeTransfer` (preparing, backing up uploads, deleting media, clearing DB for restore).
  - **Remote:** emits a message while waiting for the server to run `beforeTransfer`.

- **`strapi transfer` CLI**
  - Replaces the plain “Starting transfer…” log with an **ora** spinner that keeps the **“Starting transfer…”** prefix, appends the current prep step after ` — `, shows **elapsed time** (same time format as stage lines), and finishes with **`succeed` / `fail`** so the line stays in the scrollback when stage spinners start.
  - Wires `onTransferPhase` through a small bridge to the engine progress stream after the engine exists.
  - Sets **`process.noDeprecation`** (via a type assertion) so **Node `DeprecationWarning`** output (e.g. overlapping `pg` `client.query()` calls) does not interleave with spinners on stderr.

- **Progress line (`loadersFactory` / `updateLoader`)** — used by transfer (and import/export)
  - **Dot-separated** single-line layout (` · `) instead of nested parentheses.
  - **Throughput:** **`items/s`** for **entities** and **links**; **byte/s** for other stages (e.g. assets).
  - **ETA:** linear estimate from average rate vs **total bytes** or **total count** (entities/links when `totalCount` is known), capped at 24h; shown as **`1.2s, ~4.0s remaining`** (not a ratio).
  - **Timing:** plain `readableTime` when there is no ETA; elapsed is **last** on the line.

- **Tests:** extends `data-transfer-loader.test.ts` for items/s, count-based ETA, and timing/remaining text.

### Why is it needed?

- Restore prep can take a long time with no feedback; users need to see **what the destination is doing** (local steps vs waiting on remote).
- **Entity/link** stages were showing **byte/s**, which is a poor proxy for DB work; **items/s** matches what users care about.
- **ETA** for entity stages when **total count** is known improves predictability alongside existing byte-based ETA for assets.
- **One-line** layout stays compatible with **logging** that interleaves with the TTY, while being easier to read than stacked parentheses.
- **Postgres deprecation warnings** on stderr **break ora output** (garbled lines); suppressing deprecations during the transfer command avoids that without changing driver behavior in this PR.

### How to test it?

Experimental: `0.0.0-experimental.2354a5853cf4c4d3d921aa1481436dee765a4439`

**Unit tests**

```bash
cd packages/core/strapi
yarn jest src/cli/utils/__tests__/data-transfer-loader.test.ts --watchman=false

### CODE REVIEW

File: packages/core/strapi/src/cli/commands/transfer/action.ts [hunk 1]
```diff
@@ -56,6 +65,10 @@ interface CmdOptions {
  * Transfers data between local Strapi and remote Strapi instances
  */
 export default async (opts: CmdOptions) => {
+  // Avoid DeprecationWarning lines on stderr (e.g. pg `client.query()` while a query is in flight)
+  // interleaving with ora spinners during transfer. (Runtime API; not on all @types/node Process typings.)
+  (process as NodeJS.Process & { noDeprecation: boolean }).noDeprecation = true;
```
jhoward1994, line 67: Will this suppress warning for the entire process lifetime? If so maybe we restore normal behaviour when the spinner has finished ?

innerdvations: It's just in the CLI, so it will stop as soon as the command finished

innerdvations: Basically without this, some deprecation warnings from the strapi app that starts up in the background appear in the CLI, which is ugly and unexpected. But starting the same strapi project normally will still show them

packages/core/strapi/src/cli/commands/transfer/action.ts
```diff
@@ -159,8 +181,52 @@ export default async (opts: CmdOptions) => {
...
+  let prepStepDetail: string | null = null;
+
+  const formatPrepSpinnerLine = () =>
+    prepStepDetail != null && prepStepDetail !== ''
+      ? `${STARTING_TRANSFER_PREFIX} — ${prepStepDetail}`
+      : STARTING_TRANSFER_PREFIX;
+
+  transferPhaseBridge.emit = (message: string) => {
+    prepStepDetail = message;
+    progress.emit('transfer::phase', { message: formatPrepSpinnerLine() });
```
jhoward1994, line 189: I don't see any other use of 'transfer::phase' e.g. a listener. Is this needed ?

butcherZ (dismissed):
QAed locally, everything works as expected :D

packages/core/strapi/src/cli/commands/transfer/action.ts
innerdvations: I think it's worth keeping in case someone needs to make use of it (including user code, although it's not officially supported)

butcherZ (approved):

PR #25927, opened 2026-04-04
fix(core): serialize JSON columns before INSERT in discard-drafts migration

omebarki:
mysql2 automatically deserializes JSON columns on SELECT but knex does
  not re-serialize them on INSERT, causing the 5.0.0 discard-drafts
  migration to fail when cloning component rows that contain json or
  blocks attributes. Add a serializeJsonColumns helper that re-stringifies
  object values based on component metadata, with no effect on databases
  that already return JSON columns as strings.


lucasaba:
I'm really interested in this fix!

rhanneken:
To help people work around this problem, I've adapted @omebarki's change into a patch. See [@strapi+core+5.51.2.patch](https://github.com/user-attachments/files/30874973/%40strapi%2Bcore%2B5.51.2.patch).  Save it to `<project-name>/patches/@strapi+core+5.51.2.patch` and add a `postinstall` script to `package.json`:

```json
"scripts": {
  "postinstall": "patch-package"
}
```
Then run `npm install patch-package --save-dev`. Now when you run `npm install`, it will patch `@strapi/core` to include @omebarki's change.

innerdvations (approved):
The mysql2 SELECT→object / knex INSERT path on `cloneComponentInstance` is the right place to stringify `json`/`blocks`. Unit tests cover object/array stringify, string/null no-ops, and `columnName` lookup. Merged `develop`.

_AI-assisted comment._

PR #25931, opened 2026-04-05
feat: allows having dynamic zones in components

AtiX:
### What does it do?

Allows adding dynamic zones in components, making the overall data model more flexible.

### Why is it needed?

This solves a typical use case where strapi is being used as a pagebuilding CMS, closing the gap to other solutions like storyblok.

A typical pagebuilding approach looks like:

1. Have a `page` data model
2. Have multiple `blocks` defined as components (such as a `HeroBlock`, `ImageAndText`, `CtaBlock`, ...) which are rendered in the frontend, e.g. as individual react components
3. On the `page`, have some sort of `contentBlocks` dynamic zone where those `blocks` can be placed to build up the page.

Now, the current issue is that those blocks cannot contain dynamic zones themselves. However, for proper pagebuilding, one does need it:

- Imagine a `TwoColumnBlock` that just implements a two-column layout without having to care about what is contained "in it" (= defined a `leftColumn` and `rightColumn` dynamic zone).
- Imagine a `CtaBlock` that stands out visually, which however may contain other content elements itself
- ...

This PR solves this issue by allowing those dynamic zones in components.

Bildschirmfoto 2026-04-05 um 11 51 55

### How to test it?

1. You should be able to define dynamic zones in the content type builder inside a component
2. You should then be able to place a component and elements within its dynamic zones.

### Related issue(s)/PR(s)

- https://feedback.strapi.io/developer-experience/p/dynamic-zone-in-components

### DIFF
--- a/packages/core/content-type-builder/admin/src/components/DataManager/utils/retrieveNestedComponents.ts
+++ b/packages/core/content-type-builder/admin/src/components/DataManager/utils/retrieveNestedComponents.ts
@@ -5,17 +5,25 @@ export type NestedComponent = {
   component: UID.Component;
   uidsOfAllParents?: UID.Component[];
   parentCompoUid?: UID.Component;
+  /** UIDs of parents that reference this component via a dynamiczone attribute */
+  dzParentUids?: UID.Component[];
 };
 
+/** Pre-merge entry with edge-type metadata, consumed only by mergeComponents */
+type RawNestedComponent = NestedComponent & { isDzEdge?: boolean };
+
 export const retrieveNestedComponents = (appComponents: Components): NestedComponent[] => {
-  const nestedComponents = Object.keys(appComponents).reduce((acc: NestedComponent[], current) => {
-    const componentAttributes = appComponents?.[current]?.attributes ?? [];
-    const currentComponentNestedCompos = getComponentsNestedWithinComponent(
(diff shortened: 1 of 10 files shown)


AtiX:
I am now trying to reach active devs on this project :) - @giu1io is there any chance to get a feedback on this PR? I am happy to work on this with you guys, but I guess the main questions is is "are you open to this feature"?

giu1io:
Hi @AtiX, thanks for the ping and sorry for the long silence.

To your main question: yes, this is on our radar, not something we've ruled out. But it needs real work on our side before we can take it further, and I want to be upfront about why.

Two things gate it:

**Performance.** Components and dynamic zones are already slower than plain relations. Nesting a dynamic zone inside a component multiplies that cost on both the Content Manager and the Content API. Before we'd commit to this, we need a proper technical assessment of that impact on both sides.

**UX.** Editing deeply nested flexible content doesn't work well in our current layout — the sidebar isn't well suited to nested editing. That's part of why we currently cap component nesting at depth 2.

Your PR covers the Content Type Builder and validation layer, which is useful groundwork. The Content Manager editing experience isn't touched here.

We do have an internal design exploration already in motion. This is planned work for our team in Q3, so the eventual shape may differ from what's here.

I'll follow up once there's something concrete to share.

Issue #19468, opened 2024-02-09
Images uploaded to the Media Library on S3 expire when used inside Content Blocks


preist:
### Required System information

- Node.js version: v21.6.1 / LTS in production
- NPM version: 10.2.4
- Strapi version: 4.17.1
- Database: Aurora MySQL / MySQL Ver 14.14 Distrib 5.7.44
- Operating system: Not relevant, reproducible on both Linux and latest macOS
- Is your project Javascript or Typescript: TypeScript

Local development:
```
Environment: development
OS: darwin-arm64
Strapi Version: 4.17.1
Node/Yarn Version: npm/10.2.4 node/v21.6.1 darwin arm64 workspaces/false
Edition: Community
Database: mysql
```

### Describe the bug

I'm using S3 as a provider for my media library, currently through: `"@strapi/provider-upload-aws-s3": "4.17.1"`. Images work everywhere without issues, except for Content Blocks, the new rich text editor where they work for a few minutes, and then expire:

<img width="700" alt="image" src="https://github.com/strapi/strapi/assets/14940633/9246fa83-f459-4c72-bceb-a44534614ae0">

I believe this has something to do with AWS signed urls, after a few minutes the images become unavailable and their tokens don't get refreshed.

### Steps to reproduce the behavior

1. Setup Strapi with images provided through a private S3 bucket
2. Create a content type that includes the new Content Blocks rich text editor
3. Create an entry and add an image through the media library
4. Wait for about 10 minutes
5. Go back to the editor and note that the image thumbnail now longer loads
6. If you try to build a page from this content type, it will also use an expired S3 signed url

### Expected behavior

I expect the images to keep working.

preist:
### Update:

I upgraded to the current latest version of Strapi 4.20.0 and ensured I'm running on Node v18.19.0 lts/hydrogen. The issue is still present.

thematrixdev:
I have got this issue on Strapi 5.

markkaylor:
Closing this one as a duplicate, let's track it on https://github.com/strapi/strapi/issues/25756

Issue #18026, opened 2023-09-12
AWS S3 Signed URL in Rich Text Field


Jahorse:
### Required System information

- Node.js version: v18.17.0
- NPM version: v9.6.7
- Yarn version: v3.3.0
- Strapi version: 4.11.3
- Database: mariadb  Ver 15.1 Distrib 10.5.22-MariaDB, for Linux (x86_64) using readline 5.1
- Operating system: Rocky Linux 8.8 (Green Obsidian)
- Is your project Javascript or Typescript: JS

### Describe the bug

This happens if you are using the AWS S3 upload provider with signed URLs and try to insert an image into a rich text field using the WYSIWYG editor. After selecting the image to insert, it gets inserted into the markdown format with a URL to the S3 object with the AWS access key, expiration, and signature appended as GET parameters.

This is an issue because the signed URL will expire after the configured amount of time and the images in the rich text field will break.

### Steps to reproduce the behavior

1. Set up a Strapi backend that uses `@strapi/provider-upload-aws-s3`.
2. Create a collection type with a rich text field.
3. Upload an image into the Media Library (this should upload to the configured S3 bucket).
4. Create a new entry for the collection type with the rich text field, this should bring up the WYSIWYG editor with a text area that you can write markdown in.
5. Using the Image button in the WYSIWYG editor's menu, add the image that was added to the Media Library.
6. The image should appear in the text area in the format `![image](https://bucket.s3.aws.com/image.jpg??AWSAccessKeyId=A1B2C3&Expires=1694892452&Signature=A1B3C3%3D)`
7. Save the entry.
8. Request the entry through something like the graphql playground and see that the rich text contains the image URL with the GET parameters as shown in the rich text editor. It can be copy and pasted into the browser and the image should appear.
9. Delete the GET parameters in the WYSIWYG editor and save the entry again.
10. Run the graphql query and the URL should not contain the GET parameters and won't be accessible through a browser.

### Expected behavior

I would expect it to insert the image in a way that it would append the required AWS GET parameters to the image's URL when the entry was requested by an authorized client rather than when the image was inserted into the rich text editor.

Jahorse:
I came up with this as a workaround:
1. Added a media field to my blog article content type that any images need to be added to if they are going to be embedded in the rich text field.
2. Defined the syntax `!![n]` to be used in the rich text field for signaling when an image should be embedded. 
[...]

Issue #16925, opened 2023-06-05
S3 provider with private ACL continues to request items after signed URL has expired


timbarclay:
### Required System information

- Node.js version: 16
- NPM version: 8.3.1
- Strapi version: 4.10.5
- Database: Postgres
- Operating system: linux (node alpine docker)
- Is your project Javascript or Typescript: Typescript

### Describe the bug

The AWS S3 provider plugin appears to use expired pre-signed URLs.

Also the [readme](https://github.com/strapi/strapi/tree/main/packages/providers/upload-aws-s3) is unclear about what the default expiry value is. Within the readme it says variously:

* "Defaults to 15 minutes"
* "The default value is 7 days"

### Steps to reproduce the behavior

1. set up the AWS S3 provider plugin with `ACL: 'private'` and without explicitly setting `signedUrlExpires`
2. upload an image to the media library
3. insert that image into a rich text block and save it (the image should render correctly in the preview at this point)
4. wait 20 minutes
5. reload the rich text block
6. the inserted image is broken because the presigned url has expired

### Expected behavior

On reloading the saved text block, you should still see the inserted image.

### Screenshots

![image](https://github.com/strapi/strapi/assets/6851865/fa279f56-3ee6-432f-b07a-c14fe8016881)
A text block with a broken image

![image](https://github.com/strapi/strapi/assets/6851865/afffa976-e6a8-4a7f-9a1b-636ffed452a9)
The response returned by AWS showing that the url expired after 15 minutes (900 seconds).

### Code snippets

plugins.ts
```typescript
export default ({ env }) => ({
  upload: {
    config: {
      provider: env.bool('IS_LOCAL') ? 'local' : 'aws-s3',
      providerOptions: {
        region: env('AWS_REGION', 'eu-west-2'),
        sslEnabled: true,
        params: {
          Bucket: env('CMS_S3_BUCKET', 'iwsr-test-strapi-media-library'),
          ACL: 'private',
        },
      },
      actionOptions: {
        upload: {
          ServerSideEncryption: 'AES256',
        },
        uploadStream: {
          ServerSideEncryption: 'AES256',
        },
      },
    },
  },
});

```

Marc-Roig:
hello! your configuration seems ok. And the issue seems to be with the rich text editor.

And you are completely right, the README gives no clear indication which is the default Expiry time. I will update that asap.

When you import an image into the rich text editor, you are importing the url, the same as if you used any other URL, e.g.
[...]

PR #25940, opened 2026-04-06
fix(upload): re-sign expired S3 URLs in richtext and blocks fields

### LINKED ISSUE #25756: Signed URLs in private S3 buckets expire and are not refreshed for images embedded in text/richtext fields
### Node Version

20.0.0

### Package Manager

yarn

### Package Manager Version

1.22.22

### Strapi Version

5.39.0

### Operating System

Other Linux

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When using `@strapi/provider-upload-aws-s3` with a private bucket and signed URLs enabled, images embedded in text/richtext fields (e.g. via the WYSIWYG editor) store the signed URL directly in the database as a static string. Unlike media-type fields — which receive fresh signed URLs on every retrieval via `signFileUrlsOnDocumentService()` — these embedded URLs expire (default 15 minutes) and are never refreshed.

The root cause is that the `signEntityMedia` visitor in `packages/core/upload/server/src/services/extensions/utils.ts` only processes attributes where `attribute.type === 'media'`. It does not parse or re-sign URLs embedded in richtext/text field content.

### Steps to Reproduce

1. Configure `@strapi/provider-upload-aws-s3` with a private bucket (ACL set to `private`) and signed URLs enabled
2. Create a content type with a richtext or text (markdown) field
3. Upload an image via the media library and embed it into the richtext field using the WYSIWYG editor
4. Save the content entry
5. Wait for the signed URL to expire (default: 15 minutes)
6. Fetch the content via the API or reload the admin panel — the embedded image URL is now broken/returns 403

### Expected Behavior

Signed URLs embedded in text/richtext fields should be refreshed on content retrieval, the same way URLs in media-type fields are refreshed. The document service signing logic should also handle asset URLs found within text/richtext content.

### Logs

_No response_

### Code Snippets

Example S3 provider config with private bucket:

```js
// config/plugins.js
  upload: {
    config: {
      provider: 'aws-s3',
      providerOptions: {
        s3Options: {
          credentials: {
            accessKeyId: env('AWS_ACCESS_KEY_ID'),
            secretAccessKey: env('AWS_ACCESS_SECRET'),
          },
          region: env('AWS_REGION'),
          params: {
            ACL: 'private',
            signedUrlExpires: 15 * 60, // 15 minutes (default)
            Bucket: env('AWS_BUCKET'),
          },
        },
      },
    },
  },
});
```

### Media

_No response_

### Additional information

Related to Pylon support ticket 7616.

Affected packages: `@strapi/provider-upload-aws-s3`, `@strapi/upload`, `@strapi/content-manager`

### Are you sure this issue is NOT a duplicate?

- [X] I confirm this issue is not a duplicate

### Confirmation that you follow Strapi's Code of Conduct

- [X] I read and follow [Strapi's Code of Conduct](https://github.com/strapi/strapi/blob/master/CODE_OF_CONDUCT.md)

coding-shalabh:
## Summary

Fixes #25756 — Signed URLs in private S3 buckets expire and are not refreshed for images embedded in text/richtext fields.

Also fixes the root cause behind #18026, #16925, #19468 (all report the same issue).

## Problem

When using `@strapi/provider-upload-aws-s3` with a private bucket, images embedded in richtext/text/blocks fields store the signed URL directly in the database. The `signEntityMediaVisitor` only processes `attribute.type === 'media'` and returns early for all other types. These embedded URLs expire (default 15 minutes) and are never refreshed.

## Root Cause

In `packages/core/upload/server/src/services/extensions/utils.ts`, the visitor checks:
```typescript
if (attribute.type !== 'media') {
  return; // <-- richtext, text, and blocks fields are skipped entirely
}
```

## Fix

Extends `signEntityMediaVisitor` to also process:

1. **`richtext` and `text` attributes**: Scans content for upload provider URLs using regex, looks up the corresponding File record in the database, generates a fresh signed URL, and replaces it in the content.

2. **`blocks` attributes**: Recursively walks the structured blocks JSON tree, finds image nodes, and re-signs their URLs using the existing `signFileUrls` service.

The existing media-type signing behavior is completely unchanged.

### How URL matching works

- Regex matches URLs containing `/uploads/` path (Strapi's default upload pattern)
- For each matched URL, strips query params (old signed params) to get the base URL
- Looks up the File record by matching the base URL in the database
- Generates a fresh signed URL via `provider.getSignedUrl()`
- Replaces all occurrences in the text (including stale signed params)

### For blocks content

- Recursively processes the JSON tree
- Identifies `type: 'image'` nodes that have an `image.url` property
- Re-signs the image using `signFileUrls` (same as media-type fields)

## Test plan

- [ ] Existing media signing tests still pass
- [ ] Images in richtext fields get re-signed on content retrieval
- [ ] Images in blocks fields get re-signed on content retrieval
- [ ] Non-private providers are not affected (early return)
- [ ] URLs from external sources (not upload provider) are not modified
- [ ] Performance: only runs regex + DB lookups when provider is private

### DIFF
--- a/packages/core/upload/server/src/services/extensions/utils.ts
+++ b/packages/core/upload/server/src/services/extensions/utils.ts
@@ -27,7 +27,120 @@ function isFile(value: unknown, attribute: Schema.Attribute.AnyAttribute): value
 }
 
 /**
- * Visitor function to sign media URLs
+ * Matches URLs from the upload provider in text content.
+ * Looks for URLs pointing to the /uploads/ path which is Strapi's default
+ * upload path, or URLs matching the configured provider's base URL.
+ */
+const UPLOAD_URL_REGEX = /(https?:\/\/[^\s"'<>)]+\/uploads\/[^\s"'<>)]+)/g;
+
+/**
+ * Re-signs upload provider URLs found within text/richtext/blocks content.
+ * This handles the case where signed S3 URLs are embedded in richtext fields
+ * and expire after the signing period.
+ */
+const signUrlsInText = async (text: string): Promise<string> => {
+  const { provider } = strapi.plugins.upload;
+  const isPrivate = await provider.isPrivate();
+
+  if (!isPrivate || !text) {
+    return text;
+  }
+
+  // Find all URLs in the text that could be from the upload provider
+  const urls = text.match(UPLOAD_URL_REGEX);
+  if (!urls || urls.length === 0) {
+    return text;
+  }
+
+  // Deduplicate URLs to avoid signing the same URL multiple times
+  const uniqueUrls = [...new Set(urls)];
+
+  // Look up files by URL and generate signed URLs
+  let result = text;
+  for (const url of uniqueUrls) {
+    // Strip any existing query params (signed URL params) to find the base URL
+    const baseUrl = url.split('?')[0];
+
+    // Find the file in the database by matching the URL or hash
+    const files = await strapi.db.query(FILE_MODEL_UID).findMany({
+      where: {
+        $or: [
+          { url: baseUrl },
+          { url: { $contains: baseUrl.split('/').pop() } },
+        ],
+      },
+      limit: 1,
+    });
+
+    if (files.length === 0) {
+      continue;
+    }
+
+    const file = files[0] as File;
+    try {
+      const signedUrl = await provider.getSignedUrl(file);
+      if (signedUrl?.url) {
+        // Replace all occurrences of this URL (including any old signed params)
+        // Match the base path portion and replace including any query string
+        const escapedBase = baseUrl.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        const urlPattern = new RegExp(escapedBase + '(\\?[^\\s"\'<>)]*)?', 'g');
+        result = result.replace(urlPattern, signedUrl.url);
+      }
+    } catch {
+      // If signing fails, leave the URL as-is
+    }
+  }
+
+  return result;
+};
+
+/**
+ * Recursively re-signs URLs in blocks-type content (Strapi's structured blocks editor).
+ * Blocks content is a JSON structure with nested children that may contain image URLs.
+ */
+const signUrlsInBlocks = async (blocks: any[]): Promise<any[]> => {
+  const { provider } = strapi.plugins.upload;
+  const isPrivate = await provider.isPrivate();
+
+  if (!isPrivate || !Array.isArray(blocks)) {
+    return blocks;
+  }
+
+  const processNode = async (node: any): Promise<any> => {
+    if (!node || typeof node !== 'object') {
+      return node;
+    }
+
+    // Handle image nodes in blocks
+    if (node.type === 'image' && node.image?.url) {
+      const file = node.image as File;
+      try {
+        const { signFileUrls } = getService('file');
+        const signedFile = await signFileUrls(file);
+        return { ...node, image: signedFile };
+      } catch {
+        return node;
+      }
+    }
+
+    // Recursively process children
+    if (Array.isArray(node.children)) {
+      return {
+        ...node,
+        children: await async.map(node.children, processNode),
+      };
+    }
+
+    return node;
+  };
+
+  return async.map(blocks, processNode);
+};
+
+/**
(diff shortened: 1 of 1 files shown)


packages/core/upload/server/src/services/extensions/utils.ts
```diff
@@ -27,7 +27,120 @@ function isFile(value: unknown, attribute: Schema.Attribute.AnyAttribute): value
...
+    }
+
+    const file = files[0] as File;
+    try {
+      const signedUrl = await provider.getSignedUrl(file);
+      if (signedUrl?.url) {
+        // Replace all occurrences of this URL (including any old signed params)
+        // Match the base path portion and replace including any query string
+        const escapedBase = baseUrl.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        const urlPattern = new RegExp(escapedBase + '(\\?[^\\s"\'<>)]*)?', 'g');
```
markkaylor: eslint is complaining that you should use a template string

packages/core/upload/server/src/services/extensions/utils.ts
```diff
@@ -39,21 +152,38 @@ const signEntityMediaVisitor: SignEntityMediaVisitor = async (
...
 
-  if (isFile(value, attribute)) {
-    // If the attribute is repeatable sign each file
-    if (attribute.multiple) {
-      const signedFiles = await async.map(value, signFileUrls);
-      set(key, signedFiles);
-      return;
+  // Handle richtext and text attributes — re-sign embedded upload URLs
+  if (
+    (attribute.type === 'richtext' || attribute.type === 'text') &&
```
markkaylor: Why attribute.type text? A text attribute won't ever have an image embed

packages/core/upload/server/src/services/extensions/utils.ts
```diff
@@ -27,7 +27,120 @@ function isFile(value: unknown, attribute: Schema.Attribute.AnyAttribute): value
 }
 
 /**
- * Visitor function to sign media URLs
+ * Matches URLs from the upload provider in text content.
+ * Looks for URLs pointing to the /uploads/ path which is Strapi's default
+ * upload path, or URLs matching the configured provider's base URL.
+ */
+const UPLOAD_URL_REGEX = /(https?:\/\/[^\s"'<>)]+\/uploads\/[^\s"'<>)]+)/g;
```
markkaylor: Isn't the regex here assuming a local provider with `uploads`? I think the s3 provider may not include that and the regex will fail. For example while testing your PR my files were uploaded to `http://127.0.0.1:9000/strapi-private/`. Your fix worked for blocks but not for richtext (markdown)

markkaylor (commented):
Thanks for the PR, looks like it's working for blocks but not sure about richtext

markkaylor:
Hey @coding-shalabh thanks for this PR, just wanted to give you another ping to see if you would have time to address the review comments. If you don't have the bandwidth right now that's cool we can probably take this one over internally (with you as co-author obviously), if you have a second to let us know and respond to the review comments here that'd be great 🙏

markkaylor:
Closing this PR for now due to inactivity

coding-shalabh:
Hi @markkaylor — sorry for going dark on this; the ping landed in a stretch when I was completely heads-down and I missed the window. Really appreciate you keeping it open as long as you did, and the offer to take it over with co-author credit is genuinely kind. Thank you.

I just pushed `070b65b` that addresses all three review points:

1. **Provider-agnostic richtext matching (your main concern):** dropped the hardcoded `/uploads/` regex in favour of detecting markdown and HTML embed syntax — `alt`, `[link](url)`, ``, `<a href="...">`. Any `https?://` URL inside those patterns is now a candidate, and the existing DB lookup filters non-uploads. Should work with your `http://127.0.0.1:9000/strapi-private/` MinIO setup as well as plain S3, custom CDN endpoints, etc.
2. **`attribute.type === 'text'` clause:** you're right, dropped. Text attributes can't carry image/link markup.
3. **Template literal:** fixed.

Could you reopen so CI can run? Totally understand if you'd rather still fold this into an internal PR — happy either way. Thanks again for the thoughtful review. 🙏

PR #25949, opened 2026-04-07
chore: sonarcloud security review

cache-your-dreams:
### What does it do?
Adding a hash to SonarQube dependencies in test.yml

### Why is it needed?
So that we don't fall to an infected package


.github/workflows/tests.yml
```diff
@@ -601,14 +601,14 @@ jobs:
         run: find . -name "lcov.info" -type f
 
       - name: SonarQube Scan
-        uses: SonarSource/sonarqube-scan-action@v7
+        uses: SonarSource/sonarqube-scan-action@299e4b793aaa83bf2aba7c9c14bedbb485688ec4
```
innerdvations:
```suggestion
        uses: SonarSource/sonarqube-scan-action@299e4b793aaa83bf2aba7c9c14bedbb485688ec4 # v7.1.0
```

.github/workflows/tests.yml
```diff
@@ -601,14 +601,14 @@ jobs:
...
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
 
       - name: SonarQube Quality Gate check
         id: sonarqube-quality-gate-check
-        uses: SonarSource/sonarqube-quality-gate-action@v1.2.0
+        uses: SonarSource/sonarqube-scan-action@299e4b793aaa83bf2aba7c9c14bedbb485688ec4
```
innerdvations:
```suggestion
        uses: SonarSource/sonarqube-quality-gate-action@8e493e8274794e7b686b2f4f291079d8548174f8 # 1.2.1
```

innerdvations (changes requested):
I suppose at this point we should do it for ALL of our action with all the supply chain attacks going on. Can you handle it or should I put it on the board?

nclsndr (commented):
Does that represents: "pointing to an exact version" type of measure @cache-your-dreams?

cache-your-dreams:
> Does that represents: "pointing to an exact version" type of measure @cache-your-dreams?

@nclsndr I'm not sure to fully understand your question, but I guess the answer is yes. It's something reported as a Security Hotspot by Sonarqube and with the supply chain attack's we've seen recently is a good idea to introduce in certain areas.

oiorain (dismissed):

innerdvations (dismissed):

.github/workflows/tests.yml
```diff
@@ -298,11 +278,9 @@ jobs:
         project: ['chromium', 'webkit', 'firefox']
         shard: [1/4, 2/4, 3/4, 4/4]
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-
-      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+      - uses: ./.github/actions/checkout-setup-node
         with:
-          node-version: 20
+          node-version: '20'
```
cache-your-dreams: should this still be node 20?

cache-your-dreams (commented):
Just the one comment about whether we should still be using node 20

.github/workflows/tests.yml
innerdvations: Nope, this should be 22 like the others

cache-your-dreams:
@innerdvations your refactoring seems to not work I'm afraid

innerdvations (dismissed):

innerdvations:
> @innerdvations your refactoring seems to not work I'm afraid

Ok, I reverted it. We can go with the basic version for now and try to refactor at a later time

innerdvations (approved):

PR #25957, opened 2026-04-08
release: 5.42.0

markkaylor:
release: 5.42.0


Adzouz (dismissed):
The commitlint is failing because a few of them weren't correctly formatted but if you're okay with that we can merge!

Adzouz (approved):

PR #23129, opened 2025-03-12
fix(database): improve sub queries handling in the qb

Convly:
## 🎯 Motivation
Addresses an issue where complex queries (especially `DELETE` operations with joins in the `where` clause) were generating invalid SQL due to incorrect query type handling and processing order.

## 🔄 Changes
- Fixed sub-query type handling by preserving the original query type instead of always using a select
- Added state processing tracking to prevent duplicate state processing
- Reordered query building operations to ensure the state is processed before evaluating sub-query requirements

## 📝 Before/After Example

**Before:**
```sql
delete from "public"."data" using "public"."data_recording_lnk" as "t1","public"."recordings" as "t2" 
where ("t2"."id" in ($1)) and "t0"."id" = "t1"."datum_id" and "t1"."recording_id" = "t2"."id"
-- Error: missing FROM-clause entry for table "t0"
```

**After:**
```sql
delete from "public"."data" 
where "id" in (
    select "id" from (
        select distinct "t0"."id" 
        from "public"."data" as "t0" 
        left join "public"."data_recording_lnk" as "t1" on "t0"."id" = "t1"."datum_id" 
        left join "public"."recordings" as "t2" on "t1"."recording_id" = "t2"."id" 
        where ("t2"."id" in ($1))
    ) as "subQuery"
)
```

## 🧪 How to Test
1. Create a relation between two content types
2. Try to deleteMany entries that have relations
3. Verify that the delete operation succeeds without SQL errors
4. Test with different relation types (one-to-many, many-to-many) to ensure proper query generation

## Related Issues/PRs

fix #11998

PR #25962, opened 2026-04-08
fix: backport deleteMany nested entity filter fix to v4

### LINKED ISSUE #11998: [V4] QueryEngine - deleteMany not working when filter by nested entity.

MastadoonPrime:
## Summary

Backport of #23129 to Strapi v4. Fixes `deleteMany` failing with invalid SQL when filtering by nested/related entities.

### Problem

When calling `deleteMany` with nested entity filters:
```js
await strapi.query(A).deleteMany({
  where: { B: { id: { $eq: params.id } } }
})
```

The query builder generates invalid SQL:
- **SQLite:** `no such column: t2.id`
- **PostgreSQL:** `missing FROM-clause entry for table t0`

### Root Cause

Three issues in `packages/core/database/src/query/query-builder.ts`:

1. **`processState()` called too late** — Called after `shouldUseSubQuery()`, so joins from nested filters were not yet resolved when checking if a subquery was needed.
2. **`runSubQuery()` corrupted `state.type`** — `this.select("id")` changes `state.type` from `"delete"` to `"select"`, causing the final query to be a SELECT instead of DELETE.
3. **Double processing** — `processState()` could be called multiple times due to the recursive `getKnexQuery()` call from `runSubQuery()`.

### Fix

- Move `processState()` before `shouldUseSubQuery()` so joins are resolved first
- Add `processed` flag to prevent double-processing of state
- Save `originalType` in `runSubQuery()` before `select()` mutates it
- Pass `mapResults: false` in entity-manager `deleteMany`

### Test Plan

- [x] Added integration test for `deleteMany` with deep relation filtering
- Test creates two content types with a `manyToOne` relation and verifies that `deleteMany` with nested filter correctly deletes matching entries

Closes #11998

### DIFF
--- a/packages/core/database/src/entity-manager/index.ts
+++ b/packages/core/database/src/entity-manager/index.ts
@@ -548,7 +548,7 @@ export const createEntityManager = (db: Database): EntityManager => {
       const deletedRows = await this.createQueryBuilder(uid)
         .where(where)
         .delete()
-        .execute<number>();
+        .execute<number>({ mapResults: false });
 
       const result = { count: deletedRows };
 
--- a/packages/core/database/src/query/query-builder.ts
+++ b/packages/core/database/src/query/query-builder.ts
@@ -33,6 +33,7 @@ interface State {
   aliasCounter: number;
   filters: any;
   search: string;
+  processed: boolean;
 }
 
 export interface QueryBuilder {
@@ -116,6 +117,7 @@ const createQueryBuilder = (
       aliasCounter: 0,
       filters: null,
       search: null,
+      processed: false,
     },
     initialState
   );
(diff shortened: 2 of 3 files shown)


oiorain:
Hey @MastadoonPrime 

innerdvations:
Closing as v4 is no longer supported or maintained

PR #25967, opened 2026-04-09
fix(review-workflows): implement incremental loading in assignee dropdown

### LINKED ISSUE #25945: Review Workflows Assignee dropdown only displays first 10 users (missing pagination)
### Node Version

20.0.0

### Package Manager

npm

### Package Manager Version

10.0.0

### Strapi Version

5.41.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

The Assignee dropdown in Review Workflows displays only the first 10 admin users, even when more users exist in the system.

The AssigneeSelect component calls useAdminUsers without providing a pageSize parameter.

This results in a request to GET /admin/users without pagination parameters.

The response contains only the first 10 users.

### Steps to Reproduce

1. Create more than 10 admin users
2. Enable Review Workflows on a Collection Type
3. Open an entry in the Content Manager
4. Open the Assignee dropdown in the Review Workflows panel
5. Compare the list with the full user list at Settings → Users

Result:
Only the first 10 users appear in the dropdown. Users beyond this limit cannot be selected.

### Expected Behavior

The Assignee dropdown allows selecting any existing admin user.

### Logs

```shell

```

### Code Snippets

File:
node_modules/@strapi/review-workflows/dist/admin/routes/content-manager/model/id/components/AssigneeSelect.js

The component calls:

useAdminUsers(undefined, {
    skip: isLoadingPermissions || !canRead
});

### Media

_No response_

### Additional information

The following requests were tested:

GET /admin/users

Response includes:
- total greater than 10
- results length equal to 10

GET /admin/users?pageSize=200

Response includes:
- results length greater than 10

### Confirmation Checklist

nielskaspers:
## Summary

The Review Workflows assignee dropdown previously fetched only the default first page of admin users, which meant larger organizations could not reliably find or select every assignee. This update switches the dropdown to incremental loading with search, while preserving the currently assigned user in the option list even when they fall outside the currently loaded results.

## Issue

Fixes #25945

## Changes

- Add incremental loading to AssigneeSelect.tsx using onLoadMore and hasMoreItems
- Add debounced search to the assignee dropdown via _q
- Preserve the current assignee in the dropdown options when they are outside the loaded page or filtered out by search
- Reset pagination/search state when the dropdown closes

## Testing

- Verify the assignee dropdown in Review Workflows shows all admin users (not just the first 10)
- Confirm the dropdown still loads correctly with fewer than 10 users


Bassel17 (changes requested):
Thanks for tackling this. Increasing the first pageSize does improve the current behavior for smaller admin-user sets. The remaining issue is that this still only expands the initial /admin/users fetch, so instances with more than 100 admin users will still not be able to reach everyone in the assignee dropdown.  The intended pattern is incremental combobox loading rather than a fixed larger page; a good reference is [Filters.tsx](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/admin/src/pages/ListView/components/Filters.tsx), where the admin-users filter uses Combobox with onLoadMore and hasMoreItems. I think this PR would be much more complete if AssigneeSelect followed a similar approach, while also ensuring the currently assigned user is still present even when they fall outside the initially loaded results.

nielskaspers:
Good call — pushed a follow-up that mirrors the Combobox pattern from `Filters.tsx`:

- `pageSize` starts at 10 and grows on `onLoadMore`
- server-side search via `_q` with a 300ms `useDebounce`
- `hasMoreItems` driven by `page < pageCount`
- `pageSize` and search reset on close

Also added a small `options` memo that prepends the currently assigned user when they're not in the loaded page (or filtered out by search), so the Combobox never loses its value on instances with many admin users. Fell back to `users.length` (not `options.length`) for the disabled check so the existing "no users → disabled" behavior is preserved.

Tests pass locally (`AssigneeSelect.test.tsx`), `test:ts:front` clean, and lint clean for the touched file.

Bassel17 (approved):
Looks good now, thanks for taking care of this !

PR #25968, opened 2026-04-09
fix(tests): build fails causing tests to fail

markkaylor:
### What does it do?

  1. commander.ts:89 — Added string type annotation to validate(key) parameter (was implicit any, fails with noImplicitAny)
  2. get-inquirer.ts:1 — Changed typeof import('inquirer').default to typeof import('inquirer') (the installed @types/inquirer v8 doesn't export default)

### Why is it needed?

These type errors were causing build:types to fail, which meant yarn build never completed, leaving all dist files stale. With the fixes, yarn build succeeds and all CLI tests pass.

### How to test it?

Tests should pass in CLI


markkaylor:
closing this to be replaced by https://github.com/strapi/strapi/pull/25969

PR #25972, opened 2026-04-09
feat: close button for the trial banner

cache-your-dreams:
### What does it do?
Adding a close button to the free trial banner

### Why is it needed?
Users and CEO's alike are complaining about not being able to get ride of the Free Trial banner when creating a new Strapi project. This change allows them to do just that.


packages/core/admin/admin/src/components/UpsellBanner.tsx
```diff
@@ -10,15 +11,49 @@ import { useGetLicenseTrialTimeLeftQuery } from '../../src/services/admin';
 import { RESPONSIVE_DEFAULT_SPACING } from '../constants/theme';
 import { useScopedPersistentState } from '../hooks/usePersistentState';
 
+const CollapsedButton = styled.button`
```
mathildeleg: There's issues with this collapsed button:
It's set right against the top and right corner of the page:
Image

Sometimes it's squished (Media Lib page): 
Image

And because it's not floating, it creates a separation between the top of the page and the leftside menu (on the Content Manager, the CTB pages etc):
Image

Also wondering how it looks with the AI CTB's chat floating button 🤔 Not sure it's the best UX to have two floating buttons 

I think you need to re-work how you implement this collapsed button overall :)

packages/core/admin/admin/src/components/UpsellBanner.tsx
```diff
@@ -129,11 +182,44 @@ const UpsellBanner = () => {
       isAfter(new Date(cachedTrialEndsAt), sevenDaysAgo)
   );
 
-  if (timeLeftData.data?.trialEndsAt || isTrialEndedRecently) {
-    return <Banner isTrialEndedRecently={isTrialEndedRecently} />;
+  const trialEndsAt = timeLeftData.data?.trialEndsAt ?? cachedTrialEndsAt;
+
+  const toCanonicalISO = (v: string | undefined): string | undefined => {
+    if (!v) return undefined;
+    try {
```
mathildeleg: You can use `isValid` from `date-fns` to check the validity of the date and thus remove the try/catch:

```suggestion
   const date = new Date(v);
    return isValid(date) ? date.toISOString() : undefined;
```

packages/core/admin/admin/src/components/UpsellBanner.tsx
```diff
@@ -129,11 +182,44 @@ const UpsellBanner = () => {
...
+  if (!(timeLeftData.data?.trialEndsAt || isTrialEndedRecently)) {
+    return null;
+  }
+
+  if (isDismissed) {
+    return (
+      <CollapsedButton
+        type="button"
+        aria-label={formatMessage({
+          id: 'app.components.UpsellBanner.reopen',
```
mathildeleg: Don't forget to update at least the en.json for this new msg

packages/core/admin/admin/src/components/UpsellBanner.tsx
cache-your-dreams: I've changed how it works. See whether this works better

packages/core/admin/admin/src/translations/en.json
```diff
@@ -491,6 +491,7 @@
   "app.components.ToggleCheckbox.disabled-label": "Disabled",
   "app.components.UpsellBanner.button": "Upgrade now",
   "app.components.UpsellBanner.intro": "Access to Growth plan features: ",
+  "app.components.UpsellBanner.reopen": "Reopen the banner. ",
```
mathildeleg, line 509: You also need to add the translation for `app.components.UpsellBanner.close`

packages/core/admin/admin/src/components/UpsellBanner.tsx
```diff
@@ -129,11 +155,43 @@ const UpsellBanner = () => {
...
+  );
+
+  const handleDismiss = () => setDismissedFor(toCanonicalISO(trialEndsAt));
+  const handleReopen = () => setDismissedFor(undefined);
+
+  if (!(timeLeftData.data?.trialEndsAt || isTrialEndedRecently)) {
+    return null;
+  }
+
+  if (isDismissed) {
```
mathildeleg, line 189: You still have an issue on the Content Manager and CTB when the banner is closed, where there is a gap above the leftMenu:
https://github.com/user-attachments/assets/c9439554-3a96-43a2-a5a2-6982c89e14c0

packages/core/strapi/src/cli/commands/export/__tests__/export.test.ts
```diff
@@ -62,7 +62,7 @@ jest.mock('@strapi/data-transfer', () => {
     directory: {
       ...actual.directory,
       providers: {
-        ...actual.directory.providers,
+        ...actual.directory?.providers,
```
mathildeleg, line 65: Do you need these in this PR or were they already fixed in develop? Since they're not really related to your changes

mathildeleg (commented):
One last thing to fix sorry :)

packages/core/strapi/src/cli/commands/export/__tests__/export.test.ts
cache-your-dreams: They were a needed to make tests pass. I'm happy to remove them if you prefer.

packages/core/admin/admin/src/components/UpsellBanner.tsx
cache-your-dreams: Updated so it no longer does this

mathildeleg (approved):
LGTM, thanks for all the mods :)

PR #25974, opened 2026-04-09
fix: add responseType to getFetchClient for non-JSON responses

### LINKED ISSUE #21534: useFetchClient only handles json responses

Adzouz:
## What does it do?

Adds a `responseType` option to `getFetchClient`'s `get` method, allowing plugin developers to receive `Blob`, `string`, or `ArrayBuffer` responses instead of always parsing JSON.

Here are the possible values of `responseType`: `json`, `blob`, `text`, `arrayBuffer`

## Why is it needed?

`getFetchClient` always called `response.json()`, so any non-JSON response (file download, CSV export, zip archive) would silently return `[]` or throw a `SyntaxError`. Plugin developers had to manually read the JWT token from localStorage as a workaround to bypass the fetch client entirely.

### How to test it?

You should be able to pass a `responseType` now when using `useFetchClient`'s `get` method like so:
```
const { get } = useFetchClient();

const { data: blob, headers } = await get(url, { responseType: 'blob' });
```
The response type should change accordingly.

### Related issue(s)/PR(s)

Fixes #21534

🚀

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -150,6 +150,7 @@ const attemptTokenRefresh = async (): Promise<string> => {
 type FetchResponse<TData = any> = {
   data: TData;
   status?: number;
+  headers?: Headers;
 };
 
 type FetchOptions = {
@@ -157,6 +158,7 @@ type FetchOptions = {
   signal?: AbortSignal;
   headers?: Record<string, string>;
   validateStatus?: ((status: number) => boolean) | null;
+  responseType?: 'json' | 'blob' | 'text' | 'arrayBuffer';
 };
 
 type FetchConfig = {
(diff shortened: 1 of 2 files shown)


markkaylor (approved):
Tested, looks good to me

PR #25986, opened 2026-04-10
test(cli): fix snapshot updating and replace fixed versions

innerdvations:
### What does it do?

- allows passing the `-u` to jest to allow updating snapshots
- adds a `yarn test:cli:update` to do it automatically
- replaces the fixed version number in openapi snapshots with a fixed generic

### Why is it needed?

CLI tests were recently updated to use snapshots, but these pieces were missing

### How to test it?

Make a change that would affect a snapshot then run `yarn test:cli:update`

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/25969

### DIFF
--- a/tests/cli/tests/strapi/strapi/__snapshots__/openapi-generate.test.cli.ts.snap
+++ b/tests/cli/tests/strapi/strapi/__snapshots__/openapi-generate.test.cli.ts.snap
@@ -26503,6 +26503,6 @@ exports[`openapi:generate should generate a spec describing i18n and users-permi
     },
   },
   "x-powered-by": "strapi",
-  "x-strapi-version": "5.42.0",
+  "x-strapi-version": "<current-strapi-version>",
 }
 `;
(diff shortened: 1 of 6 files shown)


tests/cli/tests/strapi/strapi/__snapshots__/openapi-generate.test.cli.ts.snap
```diff
@@ -26503,6 +26503,6 @@ exports[`openapi:generate should generate a spec describing i18n and users-permi
     },
   },
   "x-powered-by": "strapi",
-  "x-strapi-version": "5.42.0",
+  "x-strapi-version": "<current-strapi-version>",
```
markkaylor, line 26506: So this is always a placeholder now instead of the actual strapi version?

markkaylor (approved):

tests/cli/tests/strapi/strapi/__snapshots__/openapi-generate.test.cli.ts.snap
innerdvations: yeah, that will solve the issue with CI always failing (and needing to update this snapshot) every time we do a release

PR #25988, opened 2026-04-10
enhancement(db): migration performance improvements

bartsmartshore:
### What does it do?
This change drastically improves the speed of 2 migrations (document id generation & discard-draft) on our large database. With these changes the migration of our very big database is easily 6+ times faster.

- The document-id generation can run batches like this, instead of running 1 by 1, which takes ages.

- It seemed like the discard-draft migration was already set up to use a mapCache, but it wasn't used yet? this passes the cache along in functions so that it's actually used.

I tested this on a huge database, and did a diff on the databases. The only thing changed was a few ID's of mapping tables (with all FK's perfectly in place), most likely to the insertion order on deeply nested copied entries.

### Why is it needed?
Because the default migrations, for big datasets, runs longer than you can ever plan a migration for.

### How to test it?
Run the default tests in examples/complex.

### Related issue(s)/PR(s)
Not related to an issue.

2 things it needs, but I'd like to know if this would still be useful for others first.
- id generation improvement also for items with localization. (however, even if we don't do this, I guess it's just as useful for non-localization items.
- I guess the new batch size set dynamically for SQLite?

Based on previous work in this area, I guess it's easy for @innerdvations and @Bassel17  to have a look?


innerdvations (commented):
I have reviewed and tested this with basic cases and overall LGTM, but I'm holding off on merging it until we can test it with a very complex data set. I will generate an experimental this week and hopefully @derrickmehaffy can get some real-world databases to try it on.

bartsmartshore:
Great! Yea like I said. I tested on a very large database with many different entities (would consider it a pretty complex case), except for the localization, so that one I cannot verify. document_id change, as mentioned, is only non-localized entities anyway.

innerdvations:
I generated experimental `0.0.0-experimental.1bfbeac36c1001a2acf41402fb22922eada31f43` so this can be tested easier

derrickmehaffy:
Ran this PR through a migration-benchmark harness I built off `develop` ([chore/examples-complex-bench-harness](https://github.com/strapi/strapi/tree/chore/examples-complex-bench-harness)). Captured baseline vs candidate across postgres/mysql/mariadb at four multipliers each. Candidate = this PR's two commits cherry-picked onto the baseline.

### Speedup matrix (total migration time, baseline → candidate)

| multiplier | mariadb | mysql | postgres |
|---:|---:|---:|---:|
| m=1 | 12.80 s → 9.12 s (−28.7%) | 18.57 s → 14.90 s (−19.7%) | 3.22 s → 3.22 s (+0.2%) |
| m=10 | 19.26 s → 18.07 s (−6.2%) | 36.05 s → 25.26 s (−29.9%) | 9.51 s → 8.91 s (−6.4%) |
| m=20 | 35.61 s → 29.50 s (−17.2%) | 1.02 min → 42.74 s (−30.1%) | 16.46 s → 15.32 s (−6.9%) |
| m=100 | 4.14 min → 3.73 min (−9.8%) | 6.98 min → 5.82 min (−16.7%) | 1.31 min → 1.24 min (−4.9%) |

### Where the wins come from

- **`5.0.0-02-created-document-id`** — the batched UPDATE change. Consistent −20% to −60% across DBs at all multipliers. At postgres m=100: 9.01 s → 4.71 s (−45.5%).
- **`core::5.0.0-discard-drafts`** — cache-propagation change. Visible improvement on mariadb (−10% at m=1) and mysql (~−10% at higher multipliers); near-noise on postgres. The fix activates on cross-chunk component-relation processing, which our fixture exercises only partially (the `hc-m2m` anti-pattern hits `copyRelationTableRows`, not the `copyComponentRelations` caches this PR targets).

### Test setup

- **Baseline:** `develop` @ [b7576b7829](https://github.com/strapi/strapi/commit/b7576b7829) (our harness branch, same as `develop`)
- **Candidate:** baseline + PR commits [`d576ee53`](https://github.com/strapi/strapi/commit/d576ee53) + [`1bfbeac3`](https://github.com/strapi/strapi/commit/1bfbeac3) cherry-picked
- **Host:** Node v20.19.4 · linux/x64 · AMD Ryzen Threadripper 1950X (32 cores) · 126 GB
- **DBs:** podman containers on same host. Postgres 16.13, MySQL 8.4.8, MariaDB 11.
- **Schema:** the `examples/complex` fixture — 6 existing content types (basic/basic-dp/basic-dp-i18n/relation/relation-dp/relation-dp-i18n) + 2 anti-pattern types (hc-m2m-source/target) crossing the 1000-row chunk boundary.
- **Seed:** Strapi-API via `seed-v4.js` with per-content-type concurrent creates (SEED_CONCURRENCY=5).
- **Timing:** Umzug `migrating`/`migrated` event subscription from a Node `--require` preload.

> **Scheduling caveat.** Postgres runs were sequential. MySQL and MariaDB runs were executed **in parallel with each other** to save wall-clock time. Baseline and PR runs for each DB used identical conditions, so per-DB percent-change deltas are trustworthy. Cross-DB absolute-time comparisons would need matched serial conditions to be fair.

HTML report attached below (zip → extract → open).

[compare-2026-04-15T22-09-07-376Z.zip](https://github.com/user-attachments/files/26764900/compare-2026-04-15T22-09-07-376Z.zip)

packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
```diff
@@ -26,6 +26,8 @@ interface Params {
   joinTableName: string;
 }
 
+const BATCH_SIZE = 1000;
```
derrickmehaffy: Worth using the `getBatchSize(trx, defaultSize)` helper from `5.0.0-discard-drafts.ts:596` here instead of the hardcoded `1000`. That helper caps at 250 on SQLite (to stay under SQLite's expression-tree/parameter limits — historically 999 params across multiple versions). With `insert([{id, document_id}]).onConflict('id').merge()`, 1000 rows × 2 columns = 2000 params, which is safe on modern SQLite 3.32+ but brittle on older versions. Easiest path: export `getBatchSize` to a shared utility module, or duplicate it here.

packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
```diff
@@ -129,18 +131,19 @@ const migrateDocumentIdsWithLocalizations = async (db: Database, knex: Knex, met
...
+  )[0].recordsLeft;
+  while (recordsLeft > 0) {
+    const batchSize = recordsLeft < BATCH_SIZE ? recordsLeft : BATCH_SIZE;
+    const updateRecords = (
+      await knex(meta.tableName).select('id').whereNull('document_id').limit(batchSize)
+    ).map((item) => ({ id: item.id, document_id: createId() }));
+    await knex(meta.tableName).insert(updateRecords).onConflict('id').merge();
+    recordsLeft = +(
+      await knex(meta.tableName).count('* as recordsLeft').whereNull('document_id')
+    )[0].recordsLeft;
```
derrickmehaffy: You already know exactly how many rows were updated (`updateRecords.length` from line 141), so the end-of-loop `COUNT(*) WHERE document_id IS NULL` adds one extra round-trip per batch. On a 10M-row table with 1000/batch, that's 10,000 avoidable COUNT queries — ironically undercutting the speedup this PR is delivering. `recordsLeft -= updateRecords.length` is equivalent and free.

packages/core/core/src/migrations/database/5.0.0-discard-drafts.ts
```diff
@@ -3099,8 +3113,6 @@ async function copyComponentRelations({
 
     const componentCloneCache = new Map<string, Map<string, number>>();
     const clonedComponentPairsCache: ClonedComponentPairsCache = new Map();
-    const componentTargetDraftMapCache = new Map<string, Map<number, number> | null>();
-    const componentTargetReverseMapCache = new Map<string, Map<number, number> | null>();
     const componentHierarchyCaches: ComponentHierarchyCaches = {
```
derrickmehaffy, line 3139: Similar pattern to the two caches you just moved out above. `componentHierarchyCaches` (parentInstanceCache, ancestorDpCache, parentDpCache) holds lookup data that's stable within a migration run and keyed by content-type hierarchy — not by chunk. Moving it out of the `for (const publishedIdsChunk ...)` loop the same way should give additional cross-chunk cache reuse for schemas that exercise the relevant code path. `componentCloneCache` and `clonedComponentPairsCache` correctly stay per-chunk.

derrickmehaffy (commented):
Code-level follow-ups to the benchmark data I posted above. None are blockers, just improvements worth considering before merge.

### Localized path untouched

`migrateDocumentIdsWithLocalizations` (5.0.0-02-document-id.ts lines 110–128) is the i18n-enabled path and still uses the one-row-at-a-time pattern the rest of the PR fixes. You acknowledged this in the PR description as a known follow-up. Worth filing as a tracking issue so it doesn't get lost — most production Strapi apps with enough rows to notice the original slowness also have i18n enabled, so the localized path is arguably the more important fix.

### Non-code notes

- `commitlint` CI check is failing — commit `d576ee53`'s message "Run batches of document_id creation" doesn't match `type(scope): subject`. Something like `perf(database): batch document_id migration` would unblock the lint.
- The new `SELECT id FROM ... WHERE document_id IS NULL LIMIT N` (document-id migration, line 140) has no `ORDER BY` — fine for correctness since every row eventually gets an ID, but `ORDER BY id` would let the query planner use the PK index path consistently and make the scan more predictable.

derrickmehaffy:
One final note @bartsmartshore if it's something your interested in, I drafted a claude note around the discard draft migration if it's something you are interested in doing (no pressure honestly, it's not a fun thing to dig into lol)

[discard-drafts-optimization-notes.md](https://github.com/user-attachments/files/26765554/discard-drafts-optimization-notes.md)

bartsmartshore:
@derrickmehaffy Thanks for the review

Pushed some changes for this. I wasn't too confident about touching the componentHierarchyCaches, the other 2 seemed logical. I ran another quick test on the complex example and it still validates...I could try another run on our big database tomorrow and validate it against the currently merged version of these migrations...but that's going to take a while to run. 

Regarding the time testing. I think the improvement on document_id is linear to the number of records, but the discard draft one, especially in the component section, apparently takes quite a lot of records to notice a big improvement. As far as I saw during the runs we did it kept doing a select all on the source/destination table (I believe it was with `where publishedAt is not null`) in every part of the loop. That's query takes a few seconds with a huge database and without caching its executed over and over.

Regarding the overall rework of this, I'm afraid I don't have the time that it would take...It has grown into a huge file (quite some AI involved there?) and even more important, these caching mechanisms are written with a lot of side-effects, which makes it very difficult to follow the code when you read it. Hence my hesitation to touch it in the first place, but I didn't really have a choice because it simply takes too long.

derrickmehaffy:
all good from that doc I shared (yeah was absolutely AI generated), I wasn't expecting you to work on that here just if you were interested.

overall there is absolutely a performance improvement with your PR as is

bartsmartshore:
FYI, here are the times of migrations for both runs (with and without changes, before my last commit)

mysql> select * from strapi_migrations_internal;
+----+-------------------------------------------------+---------------------+
| id | name                                            | time                |
+----+-------------------------------------------------+---------------------+
|  1 | 5.0.0-rename-identifiers-longer-than-max-length | 2026-03-31 14:45:07 |
|  2 | 5.0.0-02-created-document-id                    | 2026-04-01 14:44:47 |
|  3 | 5.0.0-03-created-locale                         | 2026-04-01 14:44:48 |
|  4 | 5.0.0-04-created-published-at                   | 2026-04-01 14:46:12 |
|  5 | 5.0.0-05-drop-slug-fields-index                 | 2026-04-01 14:46:12 |
|  6 | 5.0.0-06-add-document-id-indexes                | 2026-04-01 14:46:24 |
|  7 | core::5.0.0-discard-drafts                      | 2026-04-02 00:13:10 |
+----+-------------------------------------------------+---------------------+
7 rows in set (0.00 sec)


mysql> select * from strapi_migrations_internal;
+----+-------------------------------------------------+---------------------+
| id | name                                            | time                |
+----+-------------------------------------------------+---------------------+
|  1 | 5.0.0-rename-identifiers-longer-than-max-length | 2026-04-08 15:24:29 |
|  2 | 5.0.0-02-created-document-id                    | 2026-04-08 15:28:42 |
|  3 | 5.0.0-03-created-locale                         | 2026-04-08 15:28:43 |
|  4 | 5.0.0-04-created-published-at                   | 2026-04-08 15:30:08 |
|  5 | 5.0.0-05-drop-slug-fields-index                 | 2026-04-08 15:30:08 |
|  6 | 5.0.0-06-add-document-id-indexes                | 2026-04-08 15:30:20 |
|  7 | core::5.0.0-discard-drafts                      | 2026-04-08 21:01:22 |
+----+-------------------------------------------------+---------------------+

bartsmartshore:
So I did a new run on our big database and the moving of componentHierarchyCaches outside the loop gives me quite a big diff compared to the migrations currently in develop branch. It could be mainly ids changed due to order of insertion, but it's impossible for me to track them all and verify that the import has the exact same result. 

Also it gives me no improvement of execution time, so I'd actually be leaning towards reverting that change. What do you think @derrickmehaffy?

Also in general, what else is this work waiting for now? Who's doing the further testing/validation?

innerdvations:
If the version before moving `componentHierarchyCaches` out of the loop was a real performance improvement and didn’t show the kind of large/ambiguous diffs you saw on the big database, I agree we should revert that.

On what this is waiting for: I’m good to merge once you and @derrickmehaffy have run the candidate you’re happy with (the "safe" version) against large datasets.

bartsmartshore:
I moved componentHierarchyCaches outside the loop again. Adding this change adds a lot of diff in our big database that I cannot explain and it gives no noticeable improvement in performance, so there's no point changing it imo. Did you do any further testing @derrickmehaffy ?

innerdvations:
Sorry, I'm having trouble following along. Does the current state of the PR improve performance without significant diffs in the data?

bartsmartshore:
The current state (with last commit, which reverses a change) doesn't give me any diffs that I can't explain (only orders of inserts change some PKs of relational tables). But I only tested on our database and the test script in the repo. I didn't hear from @derrickmehaffy, who I believe you asked to also test...

derrickmehaffy:
I haven't tested recently, if I get some time later this week I can probably run another test.

@innerdvations before my testing was only done on that tooling I setup in that other PR so we have it in the complex example, I think I pushed it to a multiplier of m100 since m1000 was just taking too long to generate the seed data. I don't have any active customer code/data dumps to run a real world test with atm since we delete that code/data pretty much as soon as the tickets are closed.

innerdvations:
Ok, I'll try to test it this week with the examples benchmarking you set up and if it works then we can merge

innerdvations (approved):

PR #25992, opened 2026-04-11
feat(types): improve config types and annotate example configs

unrevised6419:
### What does it do?

Improves the typing of Strapi configuration and propagates those types to the example projects.

**Core types (`@strapi/types`):**

- `modules/features` — deprecate the `FeaturesConfig` alias (now marked `@deprecated`, kept as a type alias for backwards compatibility) and point `FeaturesService.config` at the `Features` type re-exported from `core/config`.
- `core/config/server` — `LoggerConfig` now `extends winston.LoggerOptions` (imported from `@strapi/logger`), so user-provided winston options are type-checked instead of falling through the `[key: string]: unknown` index signature.
- `core/config/admin` — type the `uid` parameter of the preview `handler` as `UID.ContentType` instead of a plain `string`. To avoid an `import/no-cycle` lint error, the import uses `import type * as UID from '../../uid'` rather than the barrel.

**Example projects — TypeScript (`examples/complex`, `examples/empty`):**

- Annotate each config file with the matching `Core.Config.*` type and type the `{ env }` factory parameter as `Core.Config.Shared.ConfigParams` (`api`, `features`, `middlewares`, `plugins`).

**Example projects — JavaScript (`examples/getstarted`, `examples/kitchensink`, and the `create-strapi-app` `example-js` / `vanilla-js` templates):**

- Add JSDoc annotations (`@import`, `@type`, `@param`, `@returns`) referencing `import('@strapi/strapi').Core.Config.*`, `Core.Strapi`, and `StrapiApp` so editors give the same type feedback as the TS projects (config factories, `src/index.js` lifecycle hooks, admin `app.js` bootstrap).
- Add a `jsconfig.json` to the JS admin sources (`kitchensink`, `example-js`, `vanilla-js`) so editors pick up `allowJs` + strict checking and honour the JSDoc types.

### Why is it needed?

The previous config types had a few rough edges:

- `FeaturesConfig` was the only exposed name even though `Features` is the canonical type re-exported from `core/config`; there was no deprecation pointer steering consumers to it.
- `logger` had no typed escape hatch for passing raw winston options, forcing users to cast.
- The preview `handler` received `uid: string`, losing the ability to narrow behaviour per content type.
- The example config files had no types at all, so they gave no IDE feedback and didn't serve as a reference for users starting a new project.

### How to test it?

1. Build types: `yarn nx run @strapi/types:build` (or `yarn build`).
2. Lint types: `yarn nx run @strapi/types:lint` — should pass with no `import/no-cycle` errors.
3. In a user project, set `config.logger` to include winston `LoggerOptions` fields and confirm they type-check.
4. Implement a `preview.handler` and confirm the `uid` parameter is typed as `UID.ContentType`, enabling autocomplete against registered content types.
5. Open a TS example config (e.g. `examples/complex/config/api.ts`) and confirm the annotated `Core.Config.*` types error on invalid fields.
6. Open a JS example config (e.g. `examples/getstarted/config/server.js`) in an editor with JSDoc support and hover over `env` — it should resolve to `Core.Config.Shared.ConfigParams`.

### Related issue(s)/PR(s)

None.


packages/cli/create-strapi-app/templates/example-js/jsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "module": "nodenext",
     "moduleResolution": "nodenext",
```
unrevised6419, line 4: Note: `"moduleResolution": "nodenext"` cannot work without `"module": "nodenext"`
Added it, otherwise TS could not resolve `import('@strapi/strapi').Core.Config` namespace in JSDoc

nclsndr:
@unrevised6419 information about the direction: we're planning to provide identity functions for config to help folks having direct inference out of using them. Probably some runtime safety as well. Zod based.

unrevised6419:
@nclsndr you mean something like this?


**config/admin.ts**

```ts
import { defineConfigAdmin } from '@strapi/config'

export default defineConfigAdmin(({ env }) => ({
  // ...
}))
```

**config/server.ts**

```ts
import { defineConfigServer } from '@strapi/config'

export default defineConfigServer(({ env }) => ({
  // ...
}))
```

unrevised6419:
Closing this in favor of https://github.com/strapi/strapi/pull/27304

PR #25997, opened 2026-04-12
fix: reduce local plugin bundle size via dependency deduplication

### LINKED ISSUE #22946: Strapi 5 local plugins : size has significantly increased

SAY-5:
Local plugins in Strapi 5 are noticeably larger than in Strapi 4 because the bundler was including its own copies of common shared dependencies, react-intl, react-redux, formik, yup, and others that the admin panel already ships.

Expanded the dedupe and alias lists in both the Vite and webpack configs from 7 to 18 packages, covering all the common ones local plugins tend to import. Also added vendor chunk splitting in the Vite production config so node_modules don't get inlined into plugin bundles.

Fixes #22946


SAY-5:
Rebased on develop and cut this down. The dedupe/alias half is obsolete now that the admin alias modules live in ADMIN_VITE_ALIAS_MODULES and already cover react-redux, @reduxjs/toolkit and lodash, so all that is left is the rollup manualChunks change that pulls node_modules into one vendor chunk instead of letting each plugin chunk inline its own copies.

innerdvations:
Thanks for the PR and cutting this down to the Vite manualChunks change.

Unfortunately, we can't merge this because putting every node_modules module into one vendor chunk is not a policy we want as the default: in a published Strapi app the admin and plugins (including lazy routes and locale files) live under node_modules, so this would collapse code-splitting on first load. It also does not address the original local-plugin node_modules disk-size issue at this point.

Thanks again!

PR #25458, opened 2026-02-15
fix: use admin basename for 401 redirect path

veeceey:
## What does this PR do?

Fixes the logout/session expiry redirect when Strapi is deployed with a custom admin URL prefix.

**The problem**: `rtkQueryUnauthorizedMiddleware` hardcodes `window.location.href = '/admin/auth/login'` when it receives a 401 response. If the admin panel is served under a custom path like `/application/cms`, the redirect goes to `/admin/auth/login` instead of `/application/cms/auth/login`, resulting in a 404.

**The fix**: Use `getBasename()` (the same utility that provides the React Router basename in `StrapiApp`) to dynamically construct the redirect path. This means the middleware respects whatever admin path prefix is configured via `ADMIN_PATH`.

In the default case (`ADMIN_PATH = /admin`), the behavior is unchanged — it still redirects to `/admin/auth/login`. For custom setups, it now correctly redirects to `{custom-prefix}/auth/login`.

Fixes #24665
Fixes #23920
Fixes #24475
Fixes #25386

Related #23921

## How to test

1. Deploy Strapi with a custom admin URL (e.g. set `admin.path` to `/application/cms` in config)
2. Log in to the admin panel
3. Wait for the session token to expire, or manually trigger a 401
4. Verify the redirect goes to `/application/cms/auth/login` instead of `/admin/auth/login`

PR #25998, opened 2026-04-12
fix(admin): respect custom admin path on 401 redirect

deactanews:
### What does it do?

This PR fixes the admin 401 redirect to respect the configured admin base path.

Previously, when the admin received a 401 response, the unauthorized middleware redirected users to the hardcoded path `/admin/auth/login`. This breaks setups that use a custom admin URL, because the rest of the admin app already uses the configured basename.

This change replaces the hardcoded redirect with a redirect built from `getBasename()`, so expired sessions are sent to the correct login page for the configured admin path.

It also adds a regression test covering a custom admin path such as `/custom-admin`.

### Why is it needed?

Strapi supports configuring a custom admin URL, but on session expiry or other 401 responses the admin store middleware still redirected to `/admin/auth/login`.

As a result, users with a custom admin path could be sent to a non-existent login route instead of the actual admin login page. This makes re-authentication fail until they manually navigate back to the custom admin URL.

### How to test it?

1. Configure a custom admin URL, for example `/custom-admin`.
2. Start Strapi and open the admin at `/custom-admin`.
3. Authenticate successfully.
4. Trigger a 401 response in the admin after authentication expires or becomes invalid.
5. Confirm the app redirects to `/custom-admin/auth/login` instead of `/admin/auth/login`.

Regression test added:

- `yarn test:front:ce packages/core/admin/admin/src/tests/configureStore.test.ts`

### Related issue(s)/PR(s)

Fixes the hardcoded custom-admin redirect behavior reported in related issues about `/admin/auth/login` being used even when a custom admin path is configured.

### DIFF
--- a/packages/core/admin/admin/src/core/store/configure.ts
+++ b/packages/core/admin/admin/src/core/store/configure.ts
@@ -10,6 +10,7 @@ import {
 
 import { reducer as appReducer, AppState, logout } from '../../reducer';
 import { adminApi } from '../../services/api';
+import { getBasename } from '../utils/basename';
 
 /**
  * @description Static reducers are ones we know, they live in the admin package.
@@ -92,7 +93,7 @@ const rtkQueryUnauthorizedMiddleware: Middleware =
     // isRejectedWithValue Or isRejected
     if (isRejected(action) && action.payload?.status === 401) {
       dispatch(logout());
-      window.location.href = '/admin/auth/login';
+      window.location.href = `${window.location.origin}${getBasename()}/auth/login`;
       return;
     }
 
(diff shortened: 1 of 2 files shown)


innerdvations:
Thanks for this fix — the custom admin 401 redirect was a real gap, and the regression test you added is a nice touch.

The production change here was already merged on `develop` in #25458 (`fix: use admin basename for 401 redirect path`), which landed shortly after this PR opened, so I'm closing this as a duplicate of that work.

If you'd like to follow up, a rebase onto `develop` with just the `configureStore.test.ts` regression test would still be welcome — we don't have that coverage on `develop` yet.

Thanks again for contributing!

_AI-assisted comment._

innerdvations:
Closing as duplicate of #25458.

PR #26002, opened 2026-04-13
Feat/improve relational dropdown ux

Arav-Menon:
The relational field dropdown in the Content Manager was limited to showing only **5 items** at a time, requiring excessive scrolling when many relations existed. This PR increases the visible item count to **10**, significantly improving usability.

**Problem**

When using relational fields in the Content Manager, the dropdown list rendered via `FixedSizeList` (react-window) was capped at **5 visible rows**. This meant:

- Users with more than 5 relations had to scroll immediately
- The list felt cramped and difficult to browse
- Selecting the right entry — especially in content-heavy setups — required significant effort

This was controlled by a single constant used for **both** the server page size and the visible list height:

**Solution**

Two changes were made to `Relations.tsx`:

### 1. Increased `RELATIONS_TO_DISPLAY` from `5` → `10`

This increases the number of relations fetched per page from the server, reducing the need for users to hit **"Load More"** as frequently.

### 2. Introduced a dedicated `MAX_VISIBLE_RELATIONS = 10` constant

This decouples the **visible list height** from the **server page size**. The `dynamicListHeight` calculation now uses `MAX_VISIBLE_RELATIONS` instead of `RELATIONS_TO_DISPLAY`, making both independently configurable in the future.

// After
const RELATIONS_TO_DISPLAY = 10;   // items fetched per page from server
const MAX_VISIBLE_RELATIONS = 10;  // max items visible in the list before scrolling
```

const dynamicListHeight =
  data.length > MAX_VISIBLE_RELATIONS
    ? Math.min(data.length, MAX_VISIBLE_RELATIONS) * (RELATION_ITEM_HEIGHT + RELATION_GUTTER) +
      RELATION_ITEM_HEIGHT / 2
    : Math.min(data.length, MAX_VISIBLE_RELATIONS) * (RELATION_ITEM_HEIGHT + RELATION_GUTTER);
```

The existing scroll overflow shadow behavior (top/bottom gradient via `ShadowBox`) continues to work correctly — the half-row peek at the bottom still signals there are more items when the list overflows.

---
**Before**

Screenshot from 2026-04-13 14-07-57

**After**

Screenshot from 2026-04-13 14-16-08

---

**Files Changed**

- `packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx`

## Testing

1. Run the kitchensink example app:
   cd examples/kitchensink
   yarn develop

2. Open `http://localhost:1337/admin`
3. Navigate to **Content Manager** → any collection type with a relational field
4. Open an entry and observe the relation dropdown now shows up to **10 items** before scrolling kicks in
5. Confirm the scroll shadow gradients still appear correctly when the list overflows


mathildeleg:
Hey there, thanks for your PR, we'll tackle this issue in the design-system directly in the coming weeks as this is the best place to do it, but thank you ☺️

PR #26004, opened 2026-04-13
chore(data-transfer): isIgnoredContentType helper for filtering

jhoward1994:
### What does it do?

- Replaces the static `DEFAULT_IGNORED_CONTENT_TYPES` array with an `isIgnoredContentType` helper, which ignores content types based on the `admin::` prefix rather than an explicit UID list.
- All content types within the `admin::` namespace are now detected and excluded automatically, removing the need to update the ignored list for new admin types.

### Why is this needed?

Previously, admin content types were filtered using a hardcoded list of their UIDs. If new admin types were added (by plugins or future Strapi releases), they would not be excluded automatically

### Related issue(s)/PR(s)
fixes CMS-589


innerdvations (approved):
LGTM

PR #26005, opened 2026-04-13
fix(content-manager): optimize document layout hooks

jhoward1994:
### What does it do?

Prevents duplicate network requests when loading Edit or List view in the content manager.

Key changes:

1. Reads persisted query params from localStorage during render to avoid extra fetches.
2. Defers document list queries until layout config is loaded to prevent unnecessary initial requests.

### Why is it needed?

Previously, both views would make redundant requests—first with default params, then again with the correct ones—causing duplicate loads and wasted resources.

### Note 

#25437 fixed a data overwrite bug. This PR optimizes performance

### DIFF
--- a/packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts
+++ b/packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 
 import { usePersistentStateScope, useQueryParams } from '@strapi/admin/strapi-admin';
 import get from 'lodash/get';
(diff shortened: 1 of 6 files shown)


packages/core/content-manager/admin/src/hooks/usePersistentQueryParams.ts
```diff
@@ -21,6 +21,37 @@ const filterObjectKeys = (obj: object, keys: PropertyPath[]) => {
   return result;
 };
 
+/**
+ * Synchronously reads persisted query params from localStorage.
+ * Use this during render to seed `useQueryParams` initialParams so the first
+ * render already has the correct values (avoiding an extra request caused by
+ * the deferred useEffect restoration in `usePersistentPartialQueryParams`).
+ */
+export const readPersistedQueryParams = (
```
nclsndr: Just a detail here, this looks like a local storage helper rather than a hook. Should it live here?

Adzouz:
Hi @jhoward1994 👋 I will review and take a look but quickly I see that some things might conflict with this ongoing PR that will be merged soon: https://github.com/strapi/strapi/pull/25437 could you take a look at it and give your thoughts please? 🙏

Bassel17:
Would it be simpler to have usePersistentPartialQueryParams() return an isHydrated flag and skip the fetch until hydration completes? That seems like it would achieve the same “single request with final params” goal without introducing a separate synchronous localStorage seed path.

jhoward1994:
Waiting for https://github.com/strapi/strapi/pull/25437 to be merged and conflicts to be dealt with

Adzouz (approved):
It looks clean, I tested and I see only one call in the list view now! Even if I didn't really noticed the difference on the edit view, the code makes sense and maybe I was lacking a bit of context to reproduce the exact use case.
I think it's safe to move forward, good job 🚀

PR #26008, opened 2026-04-13
Improve Strapi bootstrap time

Baldinof:
### What does it do?

Reduce by about ~50% the startup time of Strapi (depending of the schema size), by saving DB roundtrips and parallelizing init tasks.

The first 3 commits are the most important:

- e394c2198a05ec084756d281342de09a5b87dc0d make the schema hash deterministic, to **avoid re-sync on almost each restart**
- 9850931bbfb4d7808545811aa0556b845fa85ad7 Use single query to retrieve the schema, instead of one query per tables. **Avoiding N+1 issue when syncing schema** (the more wide the schema is, the more the gain will be)
- 929019a0d7150c7550e021eec94bfda0decf62ef At startup we do ~60 lookups in the core settings table, this commit cache all settings keys for 2minutes after startup so we **save ~60 queries**

Other commits are parallelization things.

I can split in multiple PRs to lighten review if you prefer.

### Why is it needed?

Strapi often takes ~10/15s to bootstrap, with a bootstrap time that depends on the size of the schema.

This PR should make it more around ~4s.

The gains I've seen locally with the getstarted example, using a Postgres Neon database located in my region (Europe):

Before:
- schema sync needed: **~15s**
- no schema sync: **~9s**

After:
- schema sync needed: **~6s**
- no schema sync: **~3.5s**

With SQLite, the gains aren't really visible because there is zero latency between the DB and Strapi. But the main Strapi use case involves a remote DB, which is why I focused on improving network latency.

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -34,7 +34,6 @@ import { FeaturesService, createFeaturesService } from './services/features';
 import { createDocumentService } from './services/document-service';
 import { createContentSourceMapsService } from './services/content-source-maps';
 
-import { coreStoreModel } from './services/core-store';
 import { createConfigProvider } from './services/config';
 
 import { cleanComponentJoinTable } from './services/document-service/utils/clean-component-join-table';
@@ -436,13 +435,27 @@ class Strapi extends Container implements Core.Strapi {
     await this.db.init({ models });
     await this.store.prefill();
 
+    // Start license check early - it's a network call independent of everything else
+    const licenseCheckPromise = this.EE
+      ? (async () => {
+          const start = Date.now();
+          await utils.ee.checkLicense({ strapi: this });
+          this.log.debug(`License check finished (${Date.now() - start}ms)`);
+        })()
+      : Promise.resolve();
+
     let oldContentTypes;
-    if (await this.db.getSchemaConnection().hasTable(coreStoreModel.tableName)) {
+
+    // Try to read directly instead of checking hasTable first (saves a query)
+    try {
       oldContentTypes = await this.store.get({
         type: 'strapi',
         name: 'content_types',
         key: 'schema',
       });
+    } catch {
+      // Table doesn't exist yet (first run)
+      oldContentTypes = undefined;
     }
 
     await this.hook('strapi::content-types.beforeSync').call({
@@ -450,12 +463,8 @@ class Strapi extends Container implements Core.Strapi {
       contentTypes: this.contentTypes,
     });
 
-    // NOTE: commenting out repair logic for now as it is causing relationship loss in some cases
-    // will revisit soon in the future PR
-
     const status = await this.db.schema.sync();
 
-    // // if schemas have changed, run repairs
     if (status === 'CHANGED') {
       await this.db.repair.removeOrphanMorphType({ pivot: 'component_type' });
       await this.db.repair.removeOrphanMorphType({ pivot: 'related_type' });
@@ -475,32 +484,40 @@ class Strapi extends Container implements Core.Strapi {
       });
     }
 
-    if (this.EE) {
-      await utils.ee.checkLicense({ strapi: this });
-    }
+    // Skip afterSync hooks and content type store write when nothing changed
+    const contentTypesChanged =
+      !oldContentTypes ||
+      JSON.stringify(oldContentTypes) !== JSON.stringify(this.contentTypes);
 
-    await this.hook('strapi::content-types.afterSync').call({
-      oldContentTypes,
-      contentTypes: this.contentTypes,
-    });
+    // Run afterSync/store.set and server init in parallel
+    await Promise.all([
+      contentTypesChanged
+        ? (async () => {
+            await this.hook('strapi::content-types.afterSync').call({
+              oldContentTypes,
+              contentTypes: this.contentTypes,
+            });
+            await this.store.set({
+              type: 'strapi',
+              name: 'content_types',
+              key: 'schema',
+              value: this.contentTypes,
+            });
+          })()
+        : Promise.resolve(),
+      (async () => {
+        await this.server.initMiddlewares();
+        this.server.initRouting();
+        await this.contentAPI.permissions.registerActions();
+      })(),
+    ]);
 
-    await this.store.set({
-      type: 'strapi',
-      name: 'content_types',
-      key: 'schema',
-      value: this.contentTypes,
-    });
-
-    await this.server.initMiddlewares();
-    this.server.initRouting();
-
-    await this.contentAPI.permissions.registerActions();
+    // Await license check before plugin bootstraps (some plugins may need EE status)
+    await licenseCheckPromise;
 
     await this.runPluginsLifecycles(utils.LIFECYCLES.BOOTSTRAP);
 
-    for (const provider of providers) {
-      await provider.bootstrap?.(this);
-    }
+    await Promise.all(providers.map((provider) => provider.bootstrap?.(this)));
 
     await this.runUserLifecycles(utils.LIFECYCLES.BOOTSTRAP);
 
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -434,6 +434,7 @@ class Strapi extends Container implements Core.Strapi {
     ];
 
     await this.db.init({ models });
+    await this.store.prefill();
 
     let oldContentTypes;
     if (await this.db.getSchemaConnection().hasTable(coreStoreModel.tableName)) {
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -486,8 +486,7 @@ class Strapi extends Container implements Core.Strapi {
 
     // Skip afterSync hooks and content type store write when nothing changed
     const contentTypesChanged =
-      !oldContentTypes ||
-      JSON.stringify(oldContentTypes) !== JSON.stringify(this.contentTypes);
+      !oldContentTypes || JSON.stringify(oldContentTypes) !== JSON.stringify(this.contentTypes);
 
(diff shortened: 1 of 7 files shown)


packages/core/core/src/services/core-store.ts
```diff
@@ -55,16 +55,44 @@ interface CoreStore {
...
+  prefill(): Promise<void>;
 }
 
+const CACHE_TTL = 2 * 60 * 1000; // 2 minutes — only used during startup
+
 const createCoreStore = ({ db }: { db: Database }) => {
+  // Populated by prefill(), auto-expires after CACHE_TTL
+  const cache = new Map<string, { value: string; type: string; id: number }>();
+
+  const cacheKey = (key: string, environment: string | null, tag: string | null) =>
```
nclsndr, line 67: 1. What about?
```suggestion
  const makeCacheKey = (key: string, environment: string | null, tag: string | null) =>
```

---

2. this func could be declared outside of `createCoreStore` scope since no contextual values are used in it

--- 

3. Why this func is not producing the same key as `fullKey`? Is here an opportunity to converge both?

packages/core/core/src/services/core-store.ts
```diff
@@ -55,16 +55,44 @@ interface CoreStore {
   get<T = unknown>(params: GetParams): Promise<T>;
   set(params: SetParams): Promise<void>;
   delete(params: GetParams): Promise<void>;
+  prefill(): Promise<void>;
 }
 
+const CACHE_TTL = 2 * 60 * 1000; // 2 minutes — only used during startup
```
nclsndr, line 61: Do we actually need such a long TTL?
Do we expect strapi to need those values for more than 30 sec to a minute?

packages/core/core/src/services/core-store.ts
```diff
@@ -73,42 +101,57 @@ const createCoreStore = ({ db }: { db: Database }) => {
...
   };
 
+  /**
+   * Bulk-load all rows into cache. Call once early in bootstrap to avoid
+   * N individual SELECT round-trips to the database.
+   *
+   * Cache entries auto-expire after CACHE_TTL since this is only meant to save
+   * queries during startup
+   */
+  store.prefill = async () => {
```
nclsndr, line 111: What about 👇 
```suggestion
  store.prefillAll = async () => {
```

Rationale: the store has many keys, we could want an API capable of pre-loading / caching a single key. Here we signify we'll load them all at once.

packages/core/core/src/services/core-store.ts
```diff
@@ -73,42 +101,57 @@ const createCoreStore = ({ db }: { db: Database }) => {
...
+    }
 
-    const where = {
-      key: `${prefix}_${key}`,
-      environment: environment || null,
-      tag: tag || null,
-    };
-
+    const where = { key: fullKey, environment: env, tag: t };
     const data = await db.query('strapi::core-store').findOne({ where });
```
nclsndr, line 148: Side note: the table does not have indexes

packages/core/core/src/services/core-store.ts
```diff
@@ -73,42 +101,57 @@ const createCoreStore = ({ db }: { db: Database }) => {
...
+
+      for (const row of rows) {
+        const ck = cacheKey(row.key, row.environment ?? null, row.tag ?? null);
+        cache.set(ck, { value: row.value, type: row.type, id: row.id });
+      }
+
+      // Auto-expire cache after TTL — it's only meant for startup
+      setTimeout(() => {
+        cache.clear();
+      }, CACHE_TTL).unref();
```
nclsndr, line 124: Interesting that we use `unref` here. Leaving node process to exit eagerly [docs](https://nodejs.org/api/timers.html#timeoutunref)

packages/core/core/src/services/core-store.ts
```diff
@@ -121,6 +164,8 @@ const createCoreStore = ({ db }: { db: Database }) => {
 
     const prefix = `${type}${name ? `_${name}` : ''}`;
 
+    cache.delete(cacheKey(`${prefix}_${key}`, environment || null, tag || null));
```
nclsndr, line 167: Since cache is in memory, horizontal scaling would fail at consistently invalidating to the new value.

packages/core/core/src/Strapi.ts
```diff
@@ -474,32 +484,39 @@ class Strapi extends Container implements Core.Strapi {
...
-    await this.server.initMiddlewares();
-    this.server.initRouting();
-
-    await this.contentAPI.permissions.registerActions();
+    // Skip afterSync hooks and content type store write when nothing changed
+    const contentTypesChanged =
+      !oldContentTypes || JSON.stringify(oldContentTypes) !== JSON.stringify(this.contentTypes);
+
+    // Run afterSync/store.set and server init in parallel
+    await Promise.all([
```
nclsndr, line 492: Why `licenseCheckPromise` is not part of this pack?

nclsndr (commented):
Looks promising @Baldinof!


### concurrency

1. Don't we risk to hit database max connection in cases? Thinking of heavy plugin users.
2. We are trading memory usage for performance, should we run a benchmark to measure before merge? Thinking of resource limited env like CI and remote envs.

### Caching

Since the table `strapi_core_store_settings` seems to not have indexes, I would first try to add some to speed up retrieval (composite on key + environment + tag)

If we want to keep the in-memory cache, we have to acknowledge it would break horizontal scaling deployments until the cache is cleared (2 min later). Not sure it's worth the tradeoff here.

innerdvations:
As @nclsndr said, let's try adding an index to the core store as well, that might be quite valuable.

On horizontal scaling; if the cache is only on bootstrap, I feel like it's fine since it makes sense for the agreement to be that the config values used for bootstrap are retrieved at the beginning of bootstrap, not that they are allowed to change -during- bootstrap. Maybe for now, the safe solution is that the cache ONLY works on bootstrap and outside of that we do not use it? That also makes it safer than having ttl as a constant

And then in the future we can investigate if core store is slowing down general performance as well, and perhaps add a config value for memory caching.

innerdvations:
Some tests ran OOM; running them again in case it was a fluke

packages/core/core/src/Strapi.ts
Baldinof: Agree, should be here

packages/core/core/src/Strapi.ts
```diff
@@ -434,27 +433,38 @@ class Strapi extends Container implements Core.Strapi {
...
 
     await this.db.init({ models });
+    await this.store.prefill();
+
+    // Start license check early - it's a network call independent of everything else
+    const licenseCheckPromise = this.EE
+      ? (async () => {
+          const start = Date.now();
+          await utils.ee.checkLicense({ strapi: this });
+          this.log.debug(`License check finished (${Date.now() - start}ms)`);
```
Baldinof, line 443: Debug log, should be removed

packages/core/core/src/services/core-store.ts
Baldinof: Yes 💯

I don't think it's an issue, given the TTL of the cache. That's also why the code doesn't try to really maintain the cache (it does not update it on `set()` calls)

packages/core/core/src/services/core-store.ts
Baldinof: 30s should be more than enough, I'll change

packages/core/core/src/services/core-store.ts
Baldinof: I'll update for 1 and 2.

3. `fullKey` is actual the value in the `key` column in the table, while the cache key is the combination of the key, tag and env columns in the cache, I don't think we can converge them.

Baldinof:
Thank you for the review guys!

**Concurrency**

> Don't we risk to hit database max connection in cases? Thinking of heavy plugin users.
The max connections is capped by the knex pooling. The worst thing that can happen is it takes more time to acquire a connection from Knex (which could timeout, but that would mean the DB is very slow, and the user has another problem)

I don't think it will really happen unless the init time of a lot of plugin is really slow. Do you have some plugins in mind on which I could test?

> We are trading memory usage for performance, should we run a benchmark to measure before merge? Thinking of resource limited env like CI and remote envs.

Yes, it consumes ~10% more memory at bootstrap, average peak usage for 8 startups round (measured with `time -l`)

- This branch: `471MB`
- develop: `436MB`

I guess this increase is sufficient to OOM on the CI, I'll investigate if we want to move forward with this PR.

This is the peak usage according to `time`, if I take a heap snapshot after startup it's around 170MB in both cases.

**Caching**

> Since the table strapi_core_store_settings seems to not have indexes, I would first try to add some to speed up retrieval (composite on key + environment + tag)

I did some benchmarks, here the average for 8 runs for each cases:

No cache, no index: 4051ms
No cache, indexed: 3670ms
Cached: 3337ms

> If we want to keep the in-memory cache, we have to acknowledge it would break horizontal scaling deployments until the cache is cleared (2 min later). Not sure it's worth the tradeoff here.

As @innerdvations said, it should be fine. It's only reads, and when writes happen, it's just to initialize missing values ( so each replica would write the same thing). 
I think we already have the same race condition today, if an instance updates the core store right after another read it.

I can both reduce the TTL to 30s, and add early expiration when the bootstrap is finished (I prefer to keep a TTL in cas we miss the expiration for any reason)

> And then in the future we can investigate if core store is slowing down general performance as well, and perhaps add a config value for memory caching. 

I don't think it's an issue at runtime, it's heavily read at startup only, then it's fine.

I 💯 acknowledge that these changes touch things that have not moved for a long time, so it could be risky. If you prefer, I can split into 3 PRs:

- Schema inspection optimizations
- Core-store caching
- Parallelization

Baldinof:
Given the number of changes, I will create smaller PRs for each changes.

Issue #23970, opened 2025-07-16
RichText (Blocks) throws error on publish when image is missing some metadata


muzafarova:
### Node Version

v18.19.1

### NPM/Yarn/PNPM Version

10.2.4

### Strapi Version

v5.18.0

### Operating System

Strapi Cloud

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

Image upload (via Media Library) to Cloudinary works, but image's `width`, `height`, and `formats` aren't captured. This is not an issue until I try to embed an image from Media Library into RichText editor and publish. On publish, I get `400 ValidationError` response complaining about the missing image `width`, `height`, and `formats` properties. 

### Steps to Reproduce

1. use `@/provider-upload-cloudinary` plugin with Strapi v5
2. configure config/plugins.ts as per docs (see snippet attached)
3. upload an image via Media Library (not a small one as reported last year, appropriate size) 
4. Create content type with a `RichText (Blocks)` filed
5. Add image as content for the `RichText (Blocks)`, use the image from the Media Library
6. Attempt to publish content, observe `400 ValidationError` as the image won't have `width`, `height`, and `formats` defined

### Expected Behavior

Images uploaded via the Media Library via Cloudinary provider to have the `width`, `height`, and `formats` defined (currently all are `null`)

### Logs

```shell

```

### Code Snippets

```config/plugins.ts
export default ({ env }) => ({
  upload: {
    config: {
      provider: 'cloudinary',
      providerOptions: {
        cloud_name: env('CLOUDINARY_NAME'),
        api_key: env('CLOUDINARY_KEY'),
        api_secret: env('CLOUDINARY_SECRET'),
      },
      actionOptions: {
        upload: {},
        uploadStream: {},
        delete: {},
      },
    },
  },
});
```

### Media

`400 /publish`
<img width="1040" height="442" alt="Image" src="https://github.com/user-attachments/assets/86705e88-ab9b-49eb-8462-e51741a74c1c" />

### Additional information

"@strapi/provider-upload-cloudinary": "^5.18.0",
"@strapi/strapi": "5.18.0",

Issue occurs both locally and in the Strapi Cloud

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

SandipGyawali:
[...]

PR #26015, opened 2026-04-14
fix(content-manager): blocks editor errors when image has formats: null

### LINKED ISSUE #19400: RichText (Block) fails on save if an image is too small to have formats

- Node.js version: v18.18.0
- NPM version: 10.2.0
- Strapi version: 4.19.0 (first found on 4.15.5 then tried to update)
- Database: MySQL
- Operating system: Windows

### Describe the bug

RichText (Block) fails on save if an image is too small to have formats part.

1. Go to 'Any collection type item having field of type Rich Text (Block)'
2. Click on this field and add small resolution image using left-top corner menu. Image should be small enough so it wouldn't have formats (responsive upload is not applied). I used 139x88 with default Media plugin settings.
3. Preview of the image is shown inside of the Rich Text box
4. Click 'Save'
5. See error that `formats` is `null` : "details[3].image.formats must be a `object` type, but the final value was: `null`." The item edited is not saved.

### Expected behavior
If `formats` is `null`, item with small image inside of the Rich Text (Blocks) is saved with use of a simple image URL (`url` field in the image JSON root) w/o using formats part.

### Screenshots
image

### Code snippets

-

### Additional context

-

markkaylor:
### What does it do?

Make formats nullable since it's possible for images to have formats: null

### Why is it needed?

Otherwise small images throw an error on save if the image does not have a formats object

### How to test it?

Take a really small screenshot
Upload it to the media lib
Confirm on the API the formats prop is null
Go to the content manager with a content-type that has blocks
Add your small image to the blocks editor
Save the entry, it should NOT throw an error

### Related issue(s)/PR(s)

resolves https://github.com/strapi/strapi/issues/19400

https://github.com/strapi/strapi/issues/23970

### DIFF
--- a/packages/core/core/src/services/entity-validator/__tests__/blocks-validators.test.ts
+++ b/packages/core/core/src/services/entity-validator/__tests__/blocks-validators.test.ts
@@ -58,6 +58,16 @@ const validImage = [
     },
   },
 ];
+const validImageWithNullFormats = [
+  {
+    type: 'image',
+    children: [{ type: 'text', text: '' }],
+    image: {
+      ...validImage[0].image,
+      formats: null,
+    },
+  },
+];
 const validQuote = [
   {
     type: 'quote',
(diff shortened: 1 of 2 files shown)


mathildeleg (approved):
LGTM ✅

PR #26019, opened 2026-04-14
test(e2e): fixes for race conditions and file resets

innerdvations:
### What does it do?

- adds a bunch of waits in places that are otherwise flaky
- fixed the very broken git-based file reset so that it cannot result in broken snapshots

### Why is it needed?

E2E tests do not run on fast systems because they consistently fail due to a handful of race conditions

### How to test it?

all tests should pass on your local system as well as on the github CI

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/26020


docs/docs/guides/e2e/00-setup.md
```diff
@@ -19,42 +19,65 @@ To run the e2e tests, you must first install the playwright browsers.
...
 
-To run a specific file, you can pass arguments and options to playwright using `--` between the test:e2e options and the playwright options, such as:
+To pass file filters or Playwright-only flags (`--project`, `--grep`, `--reporter`, `--debug`, …), put them **after** a `--` that follows the runner options. With **npm**, you need an extra `--` so the script receives the rest: `npm run test:e2e -- --domains=admin -- login.spec.ts`.
 
 ```shell
-# to run just the login.spec.ts file in the admin domain
+# run only login.spec.ts in the admin domain
 yarn test:e2e --domains=admin -- login.spec.ts
-npm run test:e2e --domains=admin -- login.spec.ts
+npm run test:e2e -- --domains=admin -- login.spec.ts
```
markkaylor: Shouldn't we just have the yarn command since we are using yarn in the monorepo?

tests/e2e/README.md
nclsndr, line 1: Do we actually want to maintain this content in both contributor docs and here? 
I feel like we should link from here to contrib. docs.

tests/utils/shared.ts
```diff
@@ -107,6 +107,145 @@ export const clickAndWait = async (page: Page, locator: Locator) => {
...
+
+/**
+ * Navigate to a path in the admin SPA when auth/session may have changed (cookies, localStorage, tokens).
+ *
+ * Uses `waitUntil: 'domcontentloaded'` instead of Playwright’s default `load`: a client-side redirect
+ * to login can overlap a full navigation; waiting for `load` then races (Firefox: `NS_BINDING_ABORTED`).
+ * Subsequent `expect()` calls still wait for the UI.
+ */
+export const gotoAdminPath = async (page: Page, path: string): Promise<void> => {
+  await page.goto(path, { waitUntil: 'domcontentloaded' });
```
markkaylor: Would it be useful to still forward all the other `goto` options?

tests/scripts/run-tests.js
```diff
@@ -355,39 +203,20 @@ yargs
                   port,
                   appDir: testAppPath,
                   reportFileName: `playwright-${domain}-${port}.xml`,
+                  domain,
                 });
 
+                // Sync `process.env.PORT` when Playwright loads this file (before globalSetup). The
+                // canonical port is `8000 + j` above — same as baseURL / webServer; helpers read PORT.
                 const configFileTemplate = `
+process.env.PORT = ${JSON.stringify(String(port))};
```
nclsndr: Is the double cast that necessary? 🤔 
`JSON.stringify(String(port))`

tests/e2e/tests/admin/admin-auth-sessions.spec.ts
```diff
@@ -65,7 +66,7 @@ test.describe('Legacy Admin Token Migration', () => {
     });
 
     // Navigate to trigger authentication check - should redirect to login
-    await page.goto('/admin/settings');
+    await gotoAdminPath(page, '/admin/settings');
```
markkaylor: I was confused that func called `gotoAdminPath` still expects the `/admin` prefix for every call

tests/utils/dts-import.ts
```diff
@@ -70,7 +85,7 @@ export const resetDatabaseAndImportDataFromPath = async (
 
   try {
     // reset the transfer token to allow the transfer if it's been wiped (that is, not included in previous import data)
-    await fetch(`http://127.0.0.1:${process.env.PORT ?? 1337}/api/config/resettransfertoken`, {
+    await fetch(new URL('/api/config/resettransfertoken', getStrapiTestBaseUrl()), {
```
nclsndr, line 95: This can now throw if `process.env.PORT` is not defined. Default for "Manual CLI only" is provided below at line 194. Is that intended?

tests/utils/e2e-edition.ts
nclsndr, line 1: Is it possible to have this file as TS?

package.json
```diff
@@ -65,7 +65,9 @@
     "test:cli:clean": "node tests/scripts/run-cli-tests.js clean",
     "test:cli:debug": "node tests/scripts/run-cli-tests.js --debug",
     "test:cli:update": "node tests/scripts/run-cli-tests.js -u",
-    "test:e2e": "node tests/scripts/run-e2e-tests.js",
+    "test:e2e": "cross-env STRAPI_E2E_EDITION= node tests/scripts/run-e2e-tests.js",
```
nclsndr: Do we actually need the empty `STRAPI_E2E_EDITION=`?

nclsndr (dismissed):
Great job @innerdvations! 
I had significant stability improvements while running it locally. 
Some cosmetic feedbacks while reading the changes 👇

markkaylor (dismissed):
A few comments but it's all up to you, looks good to me

tests/utils/dts-import.ts
innerdvations: Yes, it's intended so that it fails fast rather than silently using the wrong server

innerdvations:
@nclsndr @markkaylor I updated with all of your suggestions if you can take a quick look again :)

markkaylor (dismissed):

PR #26020, opened 2026-04-14
fix(admin): eliminate several browser race conditions

innerdvations:
### What does it do?

- **`Form.tsx`**: Keeps `valuesRef` in sync with `state.values` during render (not only in an effect) so `getValues()` / `validate()` always see the latest committed values. Validation reads from that ref snapshot and the validate callback no longer depends on `state.values` in a way that can go stale on fast back-to-back actions.
- **`GuidedTour/Context.tsx`**: Persists guided-tour state with `useLayoutEffect` so storage is updated before the browser follows external links (e.g. documentation).
- **`usePersistentState.ts`**: Switches persistence to `useLayoutEffect` so `localStorage` is written synchronously after commit, reducing loss on quick navigation/unload.
- **`DocumentActions.tsx`**:
  - **Publish**: Yields a microtask before `validate`, then builds publish payload and `resetForm` from `getValues()` after validation so reads match the form’s latest state.
  - **Update (Save)**: Blurs the active element and waits one macrotask so debounced inputs (e.g. Blocks) flush before `validate` / `getValues`; clone/update/unpublish paths use `getValues()` for payloads and `resetForm`.
- **Blocks editor**: Adds `flushPendingFormSync` (clears debounce, `flushSync` + `onChange` with normalized Slate state) and wires it to the editable surface `onBlur` so pending debounced sync runs before Save when focus moves.
- **Blocks tests**: Passes `flushPendingFormSync` into test wrappers.
- **`documents.ts` (RTK Query)**: For `single-types`, URL builders no longer insert a `/${documentId}` segment on discard, draft-relation count, get, publish, update, and unpublish—matching the admin API routes and the same pattern already used elsewhere in this file.

### Why is it needed?

- Fast automation (E2E tests) and quick user actions could hit **Save/Publish** before React effects, debounced Slate→form sync, or `localStorage` writes completed, causing **stale validation**, **empty payloads**, or **flaky E2E** (toasts vs API, guided tour state, Blocks field).
- **Single-type** document calls could build **incorrect paths** (`…/single-types/:model/:documentId/…`) whenever `documentId` was set, which does not match the registered admin routes (`…/single-types/:model/…`).
 

### How to test it?

- **Automated**: Run the admin / content-manager test suites you normally use (e.g. unit tests for Blocks toolbar; Playwright e2e for admin CM, guided tour, blocks, conditional fields if applicable).
- **Manual (admin UI)**:
  - **Collection type**: Edit an entry with **Blocks**, save/publish quickly; confirm data persists and there are no false validation errors.
  - **Single type**: Open a single-type entry; confirm **save / publish / discard / unpublish** (and draft-relation UI if used) still work; confirm network URLs are under `/content-manager/single-types/<model>/…` without an extra document id segment.
  - **Guided tour**: Complete a step that opens an external link; reload admin and confirm tour progress persists as expected.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/pull/26019

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -766,6 +766,7 @@ const PublishAction: DocumentActionComponent = ({
       /**
        * Yield one microtask so React can flush any pending field updates from the same task
        * (common with fast local runners / Playwright) before we read values and validate.
+       * TODO: replace with an explicit form flush contract when we can (same concern as Save flow).
        */
       await Promise.resolve();
 
@@ -1046,7 +1047,6 @@ const UpdateAction: DocumentActionComponent = ({
   const setSubmitting = useForm('UpdateAction', ({ setSubmitting }) => setSubmitting);
   const initialValues = useForm('UpdateAction', ({ initialValues }) => initialValues);
   const getValues = useForm('UpdateAction', (state) => state.getValues);
-  const document = useForm('UpdateAction', ({ values }) => values);
   const validate = useForm('UpdateAction', (state) => state.validate);
   const setErrors = useForm('UpdateAction', (state) => state.setErrors);
   const resetForm = useForm('UpdateAction', ({ resetForm }) => resetForm);
@@ -1103,12 +1103,14 @@ const UpdateAction: DocumentActionComponent = ({
 
       // Blur the active element so inputs that debounce into the form (e.g. blocks editor) flush
       // before validate/getValues — on fast clients Save can otherwise read stale field state.
-      (document.activeElement as HTMLElement | null)?.blur();
+      // Use the global DOM document — form values are not in scope here (avoid shadowing `document`).
+      (globalThis.document?.activeElement as HTMLElement | undefined)?.blur();
 
-      // Let React run onBlur handlers (and any flushSync inside them) before we read the form.
-      await new Promise<void>((resolve) => {
-        setTimeout(resolve, 0);
-      });
+      // Yield microtasks so batched React updates after blur/onChange can settle before validate
+      // (same idea as performPublish; two ticks vs one gives a bit more room after focus/blur).
+      // TODO: replace with an explicit field/form flush contract when available.
+      await Promise.resolve();
+      await Promise.resolve();
 
       const { errors } = await validate(true, {
         status: 'draft',
--- a/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
@@ -621,6 +621,7 @@ const PublishAction: DocumentActionComponent = ({
   const isSubmitting = useForm('PublishAction', ({ isSubmitting }) => isSubmitting);
   const validate = useForm('PublishAction', (state) => state.validate);
   const setErrors = useForm('PublishAction', (state) => state.setErrors);
+  const getValues = useForm('PublishAction', (state) => state.getValues);
   const formValues = useForm('PublishAction', ({ values }) => values);
   const initialValues = useForm('PublishAction', ({ initialValues }) => initialValues);
   const resetForm = useForm('PublishAction', ({ resetForm }) => resetForm);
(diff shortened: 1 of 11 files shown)


packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -1095,6 +1101,15 @@ const UpdateAction: DocumentActionComponent = ({
         return;
       }
 
+      // Blur the active element so inputs that debounce into the form (e.g. blocks editor) flush
+      // before validate/getValues — on fast clients Save can otherwise read stale field state.
+      (document.activeElement as HTMLElement | null)?.blur();
```
mathildeleg: I don't think `document.activeElement` is the DOM’s active element—it’s `undefined` on a plain values object. The optional chain therefore never calls `.blur()`...

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -1095,6 +1101,15 @@ const UpdateAction: DocumentActionComponent = ({
...
+
+      // Let React run onBlur handlers (and any flushSync inside them) before we read the form.
+      await new Promise<void>((resolve) => {
+        setTimeout(resolve, 0);
```
mathildeleg: I don't think this is the greatest way to fix the issue here, using `setTimeout()` for sync is brittle, so not sure this will fix the tests

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -762,12 +763,13 @@ const PublishAction: DocumentActionComponent = ({
...
     try {
-      const { data: filteredData } = handleInvisibleAttributes(transformData(formValues), {
-        schema,
-        components,
-      });
+      /**
+       * Yield one microtask so React can flush any pending field updates from the same task
+       * (common with fast local runners / Playwright) before we read values and validate.
+       */
+      await Promise.resolve();
```
mathildeleg, line 771: Reasonable workaround for the race; but I think it's worth a follow-up (ticket or TODO) to replace with an explicit flush/contract when we can.

mathildeleg (commented):
A few comments but overall testing worked well :)

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
innerdvations: Oops, you're right, fixed!

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
innerdvations: True, I've replaced it with `await Promise.resolve()`

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
innerdvations: I agree! We should do that in a follow-up but it's probably too much to try to do it here

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
```diff
@@ -1095,6 +1101,17 @@ const UpdateAction: DocumentActionComponent = ({
...
+      // Blur the active element so inputs that debounce into the form (e.g. blocks editor) flush
+      // before validate/getValues — on fast clients Save can otherwise read stale field state.
+      // Use the global DOM document — form values are not in scope here (avoid shadowing `document`).
+      (globalThis.document?.activeElement as HTMLElement | undefined)?.blur();
+
+      // Yield microtasks so batched React updates after blur/onChange can settle before validate
+      // (same idea as performPublish; two ticks vs one gives a bit more room after focus/blur).
+      // TODO: replace with an explicit field/form flush contract when available.
+      await Promise.resolve();
+      await Promise.resolve();
```
mathildeleg, line 1113: I think you wrote two by mistake :)

mathildeleg (commented):
Just a small comment, otherwise I didn't notice regressions in the admin so it's ok for me.
I still get e2e tests failing locally but honestly can't tell if it's due to your changes or if your changes improve that 😅

packages/core/content-manager/admin/src/pages/EditView/components/DocumentActions.tsx
innerdvations: That's the "two ticks vs one" comment, it's still a hack, but generally it needs more than one tick to be safe and this is less flaky than a setTimeout

mathildeleg (approved):
Sounds good then, let's try it like that :)

Issue #26014, opened 2026-04-14
An error occurred while trying to add the relation when selecting a relation inside a nested component


smitpatel-wq:
### Node Version

20.0.0

### Package Manager

npm

### Package Manager Version

10.0.0

### Strapi Version

v5.36.1

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Bug Description

When trying to add/select a relation field inside a nested component (a component that lives inside another component), Strapi throws theerror:

An error occurred while trying to add the relation.

The relation entry exists and is published, but it cannot be connected via the Content Manager UI.

### Steps to Reproduce

  Steps to Reproduce — paste this:                                                                                                         
  1. Go to Content Manager → Products (collection type)
  2. Open any product entry                                                                                                                
  3. Scroll to the `variant_selector` component                                                                                          
  4. Inside it, find `default_variants` (repeatable nested component)
  5. Click "+ Add an entry"
  6. In the new entry, click the `collection` relation dropdown and search/select any collection (e.g. "shop-1")
  7. Observe the error toast

### Expected Behavior

The relation should be saved and the selected collection should appear connected to the nested component entry.

### Logs

```shell
An error occurred while trying to add the relation.

Because we have multiple collections for the shop-1,shop-2,shop-3,...,shop-15
so when i add a relation shop-1 i can't see that in the suggestions
```

### Media

<img width="1470" height="839" alt="Image" src="https://github.com/user-attachments/assets/7d68a7ff-edf9-4711-aed2-3c6565bbc8c3" />

<img width="1468" height="835" alt="Image" src="https://github.com/user-attachments/assets/72801275-5202-4ff9-91ef-ddff70085ee2" />

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

Arav-Menon:
I’ve investigated this and may have a potential fix. Happy to open a PR if this approach aligns.

Arav-Menon:
@nclsndr Opened a PR with a proposed fix. Would appreciate feedback. #26023

Bassel17:
Thanks for submitting the issue fixed in #26023

PR #26023, opened 2026-04-15
Fix Relation Search in Nested Components

Arav-Menon:
When a relation field is used inside a nested repeatable component (e.g., a component inside another component), searching for relations in the Content Manager often failed with a 400 error or returned no results.

This happened because the Relations component was failing to find the parent component's schema in the current document context, causing it to incorrectly fallback to the root Content Type's UID as the model.

**Changes**:

Updated Relations.tsx to trust the componentUID and componentId from the useComponent context directly for identifying the "owner" model of the relation.
Added memoization to the component schema lookup to improve stability.
Ensured that permissions are still checked correctly even if the component schema is temporarily missing from the document dictionary.
Verification:

Reproduced the issue with a 3-layer nesting (Root -> Component -> Repeatable Component -> Relation).
Verified that the "An error occurred" toast no longer appears.
Verified that relations can be searched and selected successfully in the deepest level of nesting.

**Error**
image

**Issue** : #26014


packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
```diff
@@ -183,14 +183,17 @@ const RelationsField = React.forwardRef<HTMLDivElement, RelationsFieldProps>(
       setCurrentPage(1);
     }, [isSubmitting]);
 
-    const component = componentUID && currentDocument.components[componentUID];
+    const component = React.useMemo(() => {
+      return componentUID ? currentDocument.components[componentUID] : undefined;
+    }, [componentUID, currentDocument.components]);
```
nclsndr: I don't think we want to memoize such a simple operation. Unless I'm missing something?

nclsndr (commented):
Thank you for the PR @Arav-Menon!

Not sure of the direction just yet. I started by just a quick look to the changes 👇

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Relations/Relations.tsx
Arav-Menon: @nclsndr  Thanks for pointing this out! I've removed the useMemo and simplified the logic. Let me know if this looks good.

Bassel17 (approved):
LGTM!
Thank you for the contribution

PR #26027, opened 2026-04-15
revert(admin): drop session ability cache and keep content manager batching fix

Bassel17:
### What does it do?

This change keeps the Content Manager admin performance fix on the frontend while removing the server-side admin ability cache.

### Why is it needed?

The original issue was slow admin navigation in Content Manager for users with multiple roles, caused in part by repeated `/admin/permissions/check` calls during Content Manager initialization.

The frontend batching fix addresses that request fan-out and keeps the performance improvement.

The server-side ability cache was removed because it introduced process-local authorization caching, which can lead to stale admin RBAC behavior in horizontally scaled deployments where multiple Strapi instances are running concurrently.

This keeps the safe part of the performance fix while avoiding the horizontal-scaling consistency risk.

### How to test it?

Run the focused tests:

```sh
yarn test:front packages/core/content-manager/admin/src/hooks/tests/useContentManagerInitData.test.tsx
yarn test:unit packages/core/admin/server/src/strategies/__tests__/admin.test.ts
```

### DIFF
--- a/packages/core/admin/server/src/bootstrap.ts
+++ b/packages/core/admin/server/src/bootstrap.ts
@@ -2,7 +2,6 @@ import { merge, map, difference, uniq } from 'lodash/fp';
 import type { Core } from '@strapi/types';
 import { async } from '@strapi/utils';
 import { getService } from './utils';
-import { clearAdminAuthAbilityCache } from './strategies/admin';
 import { getTokenOptions, expiresInToSeconds } from './services/token';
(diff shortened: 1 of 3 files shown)


innerdvations (approved):
LGTM!

innerdvations:
Looks like this needs @butcherZ 's fix for the flaky test to pass CI, but if everything else passes we can merge anyway

jhoward1994 (approved):

PR #26031, opened 2026-04-15
fix(i18n): preserve non-localized media when creating a locale

### LINKED ISSUE #25178: Non-localized media fields disappear when creating localizations via REST API

Bassel17:
### What does it do?

This change fixes locale creation for localized content types when they contain non-localized media fields.

It updates the Document Service i18n path to:
- populate non-localized fields before copying them into a newly created locale
- normalize copied media values into upload file IDs before persistence
- normalize media values during localization sync as well

It also adds a regression test covering locale creation through the Document Service `update()` flow for a content type with a non-localized media field.

### Why is it needed?

Creating a new locale for an existing document could clear non-localized media fields.

The issue happens because:
- the locale-creation path copied non-localized fields without populating media
- the newly created locale could end up with `null` for shared media

As a result, API-driven locale creation did not preserve shared media the same way the expected behavior requires.

### How to test it?

Run:

```bash
yarn test:api tests/api/core/strapi/document-service/update.test.api.ts
```
What to verify:
- create an `en` document with a non-localized media field
- create a new locale for the same document through Document Service `update()`
- confirm the shared media is preserved on both the source locale and the new locale

### Related issue(s)/PR(s)

Fixes #25178


packages/core/core/src/services/document-service/internationalization.ts
```diff
@@ -92,6 +96,59 @@ const localeToData: Transform = (contentType, params) => {
   return params;
 };
 
+const normalizeMediaIds = (schema: SchemaWithAttributes, data: Record<string, any>) => {
```
jhoward1994: This looks very similar to `packages/plugins/i18n/server/src/services/localizations.ts` L7 is there any way this could be shared ?

packages/core/core/src/services/document-service/internationalization.ts
```diff
@@ -92,6 +96,59 @@ const localeToData: Transform = (contentType, params) => {
...
+  Object.entries(schema.attributes).forEach(([attributeName, attribute]) => {
+    const value = data[attributeName];
+
+    if (value == null) {
+      return;
+    }
+
+    if (attribute.type === 'media') {
+      if (attribute.multiple) {
+        data[attributeName] = Array.isArray(value)
```
jhoward1994, line 116:
```suggestion
        const files = Array.isArray(value) ? value : [value];
        data[attributeName] = files.map((file: unknown) =>
          file && typeof file === 'object' && 'id' in file ? file.id : file
        );
```

is it worth doing it this way?

packages/core/core/src/services/document-service/internationalization.ts
```diff
@@ -92,6 +92,65 @@ const localeToData: Transform = (contentType, params) => {
...
+      if (attribute.repeatable && Array.isArray(value)) {
+        value.forEach((componentValue: Record<string, any>) =>
+          normalizeMediaIds(componentSchema, componentValue)
+        );
+      } else {
+        normalizeMediaIds(componentSchema, value);
+      }
+
+      return;
+    }
```
markkaylor, line 141: Will this work for nested components? 🤔

Bassel17: it should, I added a test case for it

jhoward1994 (dismissed):

jhoward1994 (approved):

PR #26032, opened 2026-04-15
fix(database): append primary key to ORDER BY for paginated selects

### LINKED ISSUE #26030: elements are repeated during pagination
### Node Version

20.19

### Package Manager

npm

### Package Manager Version

-

### Strapi Version

5.41.1

### Operating System

Linux (Debian/Ubuntu)

### Database

MySQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

after updating from 5.33.4 to 5.41.1 elements are repeated during pagination

Image

Image

### Steps to Reproduce

Create 300 records, make requests from screenshots

### Expected Behavior

elements should not be repeated

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

innerdvations:
### What does it do?

- Appends a primary-key (`id` ASC) tie-breaker to `ORDER BY` in the database query builder when a **select** query uses **pagination** (`limit` / `offset`), unless deep sort is active (relation ordering already adds an `id` tie-breaker in `wrapWithDeepSort`), or the last sort column is already `id`, or the model has no `id` attribute.
- Adds regression coverage: SQL-level assertions in `packages/core/database/src/__tests__/index.test.ts`, and an integration case in `tests/api/core/content-manager/api/basic-pagination.test.api.js` (many rows with identical `created_at`, paginate without `sort`).

### Why is it needed?

Fixes [#26030](https://github.com/strapi/strapi/issues/26030): paginated list views could show the **same document on multiple pages** (or skip rows) when the engine had no stable row order—especially on **MySQL** with `LIMIT`/`OFFSET` and no deterministic `ORDER BY`, or when many rows share the same sort key.

### How to test it?

**Unit (always run):**

```bash
cd packages/core/database && yarn test:unit src/__tests__/index.test.ts


innerdvations:
Looks like a ton of tests depended on a specific order -- I still think this is the right fix, but I will have to investigate if this could potentially cause problems for people depending on the previous default db ordering. And then maybe update a ton of tests 😭

Or, I might be able to narrow it down to only apply specifically when needed

Noga-Ekaterina:
I didn't see this issue in version 5.33.4? How long should I wait? And can I fix it by adding sorting by ID? If so, how do I formulate the query correctly?

Noga-Ekaterina:
> I didn't see this issue in version 5.33.4? How long should I wait? And can I fix it by adding sorting by ID? If so, how do I formulate the query correctly? 

@innerdvations

innerdvations:
> I didn't see this issue in version 5.33.4? How long should I wait? And can I fix it by adding sorting by ID? If so, how do I formulate the query correctly?

This issue should have a simple workaround by adding any `sort` to the query. It has been low priority for that reason, but it should be merged soon.

Bassel17 (approved):
Seems to be good !

PR #26034, opened 2026-04-15
fix(upload): content-api does not return signed urls

### LINKED ISSUE #25772: S3 upload urls not signed in response after uploading via REST api
### Node Version

22.22.1

### Package Manager

npm

### Package Manager Version

11.11.0

### Strapi Version

5.37.0

### Operating System

MacOS

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When using an s3 provider with ACL=private when retrieving files we get a signed url back. But when we upload a file using the REST API using /api/upload the response has the unsigned url which means we can not load the file.

Analysis of the source code also shows that signing this url is implemented in uploads using the admin-upload but not using the normal upload service/controller

### Steps to Reproduce

1. Have strapi setup with an s3 provider that uses private ACL, my config is this:
`  upload: {
    config: {
      provider: "aws-s3",
      providerOptions: {
        rootPath: env("S3_PATH"),
        credentials: {
          accessKeyId: env("S3_KEY"),
          secretAccessKey: env("S3_SECRET"),
        },
        region: env("S3_REGION"),
        endpoint: env("S3_ENDPOINT"),
        params: {
          Bucket: env("S3_BUCKET"),
          ACL: "private",
          signedUrlExpires: 60 * 60, // 1 hour
        },
      },
      breakpoints: {
        xlarge: 1920,
        large: 1000,
        medium: 750,
        small: 500,
        xsmall: 64,
      },
    },
  }`
2. Upload a file using the REST API by doing a POST on /api/upload as documented here: https://docs.strapi.io/cms/api/rest/upload
3. The response will not have a signed url.

### Expected Behavior

I expect the response to the POST /api/upload to have a signed url so I can immediatly use this result to show the file/image that I just uploaded

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

markkaylor:
### What does it do?

  Adds `signFileUrls()` calls to every method in the content-api upload controller (`find`, `findOne`, `destroy`, `updateFileInfo`, `replaceFile`, `uploadFiles`), signing file URLs
  before returning responses. This mirrors the existing pattern in the admin upload controller. Adds an API integration test covering all content-api upload endpoints for both private
  and public providers.

  ### Why is it needed?

  When using S3 with `ACL: "private"`, the content-api upload endpoints (`/api/upload`, `/api/upload/files`, etc.) return unsigned URLs. These URLs are inaccessible (403) because the
  bucket is private. The admin controller already signs URLs before responding, but the content-api controller was missing this step entirely.

  ### How to test it?

  Configure a Strapi project with the AWS S3 provider (you can use minio) using `ACL: "private"` (MinIO works as a local substitute). Upload a file via `POST /api/upload` and verify the response URL
  contains `X-Amz-Signature` query parameters and `isUrlSigned: true`. Verify the same for `GET /api/upload/files` and `GET /api/upload/files/:id`. Confirm that switching to `ACL:
  "public-read"` returns plain unsigned URLs as before.

  Run the new API test: `yarn test:api tests/api/core/upload/content-api/upload-signing.test.api.js`

  ### Related issue(s)/PR(s)

  - Closes #25772


markkaylor:
Detailed test plan on CMS-674

jhoward1994 (approved):

PR #26036, opened 2026-04-15
chore(examples): migration performance benchmark harness + mariadb/sqlite + anti-pattern schemas

derrickmehaffy:
### What does it do?

Extends `examples/complex` into a migration performance benchmark harness for reviewing v4→v5 migration-related PRs (and for regression-tracking over time). Three layers of change:

**DB tooling layer**
- New `scripts/compose.js` auto-detects the container/compose runtime (`podman compose` → `podman-compose` → `docker compose` → `docker-compose`) with an `STRAPI_BENCH_RUNTIME` env override.
- New `scripts/db-mariadb.js` adds MariaDB 11 alongside the existing Postgres/MySQL support, on port `3307` to avoid collision with the MySQL service on `3306`.
- New `scripts/db-sqlite.js` handles SQLite via file-copy snapshot/restore + `better-sqlite3` for `check`.
- All existing `db-*.js` scripts migrate to the runtime shim, so podman-only environments work without installing Docker. `db:check:*` now runs `ANALYZE` first to surface fresh row counts.
- MySQL/MariaDB CLIs are invoked with `-h 127.0.0.1` to force TCP (the socket path isn't populated in the official images).

**Benchmark harness layer**
- New `scripts/bench-hook.js` — Node `--require` preload that subscribes to Umzug's native `migrating`/`migrated` events for sub-ms timing. Patches `Umzug.prototype.up` in place so every instance auto-attaches listeners, resilient to the module cache aliasing issues a subclass-replace approach would hit.
- New `scripts/bench.js` — orchestrator with `seed`, `run`, and `suite` subcommands. `seed` wipes + boots v4 programmatically + runs the existing `seed-v4.js` + snapshots. `run` restores from snapshot and spawns Strapi in migrate-then-exit mode with the hook preload, capturing per-migration timings, row counts, Strapi version + git SHA + branch, and env info (node, CPU, memory, DB version, host type).
- New `scripts/bench-compare.js` — renders baseline-vs-candidate reports as a **multiplier × database matrix**, supporting arbitrary (label, multiplier, db) tuples, with both markdown (stdout + file) and self-contained HTML (inline SVG charts, sortable tables, light/dark theme via `prefers-color-scheme`) outputs.

**Schema layer**
- New `hc-m2m-source` + `hc-m2m-target` content types: high-cardinality many-to-many, with BASE counts tuned so at multiplier=100 we cross the 1000-row chunk boundary in `copyRelationTableRows`. Anti-pattern by design — not a realistic schema, documented as such in the `schema.json` description.
- `seed-v4.js` refactored to run entity creation loops with `concurrentMap(count, concurrency, taskFn)`. Default concurrency is `5` (safely under Strapi v4's default knex pool of `{min: 2, max: 10}`), tunable via `SEED_CONCURRENCY=<n>`.

### Why is it needed?

Reviewing PRs that claim migration-performance wins (e.g. #25988) currently requires either trusting the author's private-database measurements or setting up an ad-hoc benchmark per PR. This harness turns it into a standard workflow:

```bash
yarn setup:v4                                               # once
yarn bench:seed --db postgres --multiplier 100              # once per schema change
yarn bench:run --db postgres --multiplier 100 --label baseline
# git checkout <pr-branch> or cherry-pick, rebuild
yarn bench:run --db postgres --multiplier 100 --label pr-xxx
yarn bench:compare --baseline baseline --candidate pr-xxx
```

Outputs a shareable markdown + HTML report ready to post on the PR. I used it to produce the benchmark table on #25988 — the same tooling is what this PR adds.

### How to test it?

**Smoke test on a single DB/multiplier:**

1. `cd examples/complex && yarn setup:v4`
2. `cd ../../complex-v4 && yarn install`
3. Back in `examples/complex`:
   - `yarn bench:seed --db postgres --multiplier 1` — should take ~30s, produces `snapshots/postgres-bench-m1.sql`
   - `yarn bench:run --db postgres --multiplier 1 --label smoke` — should take ~15s, produces `results/postgres-smoke-<timestamp>.json` with all 7 v4→v5 migrations captured and non-zero durations
4. `yarn bench:compare --baseline smoke --candidate smoke` — produces matching markdown + HTML compare reports (self-compare is a no-op but exercises the rendering)

**Full matrix:** see the workflow description above. `yarn bench:suite --multiplier <n>` chains `bench:run` across all supported DBs.

**Podman vs Docker:** the runtime shim is exercised automatically. Force one or the other with `STRAPI_BENCH_RUNTIME=podman` / `STRAPI_BENCH_RUNTIME=docker` if you have both installed.

**Port conflicts:** if you have a local postgres/mysql on the default ports, set `POSTGRES_PORT=5433`, `MYSQL_PORT=3316`, `MARIADB_PORT=3307` (or any free ports) before `yarn bench:*` commands — they propagate to both the container and the v4 project.

### Related issue(s)/PR(s)

- First consumer: used to produce the benchmark report on #25988.
- `examples/complex` is not referenced by any CI workflow or top-level monorepo config (verified via `grep -r "examples/complex" .github/ *.{yml,json}`), so this PR cannot break CI for unrelated work.
- The knex-based fast seeder that would be needed for m=1000+ runs is intentionally out of scope here — it's a ~500-line lift that's only relevant once we need multipliers the current Strapi-API seed can't handle in reasonable time.


derrickmehaffy:
Dismissed the 9 CodeQL `js/shell-command-constructed-from-input` findings in `examples/complex/scripts/db-*.js` as **won't fix** — this is local benchmark tooling, not production code.

The flagged pattern is ``execSync(`${containerCmd()} exec ...`)`` where `containerCmd()` returns `podman` or `docker` resolved via `PATH`. Exploiting it would require the developer's `PATH` to already be compromised, at which point the attacker has code execution regardless. Hardcoding an absolute path (the usual remediation) would break legitimate installs — rootless podman ships in `~/.local/bin` on many Linux distros, not `/usr/bin`.

No behaviour change; noting here so future reviewers don't re-litigate.

jhoward1994:
I ran `bench:seed → bench:run → bench:compare` using docker Postgres. snapshots and comparison /result JSON looked correct

Great work! are there any other test scenarios I should run through?

examples/complex/scripts/bench.js
```diff
@@ -0,0 +1,474 @@
...
+  const result = {
+    timestamp,
+    label,
+    ...captureStrapiSource(),
+    strapiSource,
+    env: captureEnv(db),
+    config: {
+      multiplier,
+      snapshot,
+      seedMode: 'knex',
```
innerdvations, line 345:
```suggestion
      seedMode: 'strapi',
```
Isn't this necessary since knex isn't available yet?

innerdvations (approved):
Pushed one small fix, and a question on one more, but otherwise LGTM!

derrickmehaffy:
the tooling here is because I'm planning to add direct Knex seeding (custom script) because the APIs can be stupidly slow to seed larger instances like 1000 or 10000 multipliers.

PR #26044, opened 2026-04-17
fix: auto-expand and scroll to newly added dynamic zone components

### LINKED ISSUE #26039: Newly added component collapses and loses focus, requiring manual scroll to edit
### Node Version

 v20.19.2

### Package Manager

npm

### Package Manager Version

10.7.0

### Strapi Version

v5.34.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When adding a new component to a dynamic zone or component field in the Content Manager, the newly created component does not remain open for editing. Instead, it collapses immediately after being added, and the view resets in a way that requires the user to manually scroll to locate the newly added component.

This interrupts the expected workflow, as users typically want to configure the component immediately after adding it. The current behavior forces additional steps (scrolling and reopening), which slows down content entry and can be frustrating, especially when working with long forms or multiple components.

This issue appears to be consistent across all tested Strapi v5 versions.

### Steps to Reproduce

Steps to reproduce

Open any Collection Type with a dynamic zone or repeatable component.
Click “Add component”.
Select any component to add.
Observe the UI behavior after the component is added.

### Expected Behavior

Expected result

The newly added component should remain expanded and focused.
The viewport should automatically scroll to the newly added component.
The user should be able to immediately start editing it.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

ace2016:
### What does it do?

This PR ensures that newly added Dynamic Zone components automatically **stay expanded** and **scroll into view**, allowing users to immediately start editing them.

#### Technical Changes

**`Field.tsx`** (parent Dynamic Zone component):
- Added `usePrev` hook to track the previous value array and detect when a new component is added
- When the array grows, the new item is identified by diffing `__temp_key__` sets (works for appends *and* mid-list inserts)
- Passes `forceOpen` and `onForceOpenHandled` props to the matching `DynamicComponent`

**`DynamicComponent.tsx`** (individual component accordion):
- Added `forceOpen` and `onForceOpenHandled` props to the interface
- Added a `ref` on the `<li>` container element for scroll targeting
- Added a `useEffect` that, when `forceOpen` is `true`:
  1. Expands the accordion (`setCollapseToOpen(accordionValue)`)
  2. Scrolls the component into view via `scrollIntoView({ behavior: 'smooth', block: 'center' })`
  3. Notifies the parent to clear the forced state (`onForceOpenHandled()`)

This mirrors the existing auto-open pattern already used by `RepeatableComponent` (via `usePrev` + length comparison), bringing Dynamic Zones to feature parity.

### Why is it needed?

When adding a new component to a Dynamic Zone, the component **immediately collapses** after being added. The user must then:
1. Manually scroll to find the collapsed component
2. Click on it to expand it
3. Only then begin editing

This is especially frustrating when:
- Working with **long forms** where the new component may be off-screen
- Using **"Add component above/below"** from the context menu, where the insertion point may not be visible
- Adding **multiple components** in quick succession

The `RepeatableComponent` already handles this correctly — newly added items auto-expand. Dynamic Zones were missing this behavior entirely.

#### Impact

- **User-facing improvement**: Content editors can immediately start filling in component fields after adding them — no extra clicks or scrolling
- **Consistency**: Dynamic Zones now behave the same as Repeatable Components
- **Zero regression risk**: The fix is additive (new props with defaults), existing accordion toggle behavior is unchanged
- **2 files changed**, minimal surface area

#### Environment
- Any Strapi v5 instance with a Content Type that has a Dynamic Zone field
- Database: any (SQLite, PostgreSQL, MySQL)

#### Steps

1. Open **Content Manager** → navigate to any Collection Type with a Dynamic Zone
2. Click **"Create new entry"** (or edit an existing one)
3. Click **"Add a component to content"** → select any component

   ✅ **Expected**: The component stays **expanded** (fields visible) and the viewport **scrolls** to it

4. Add 3-4 more components to make the form long
5. On any existing component, click the **⋯** menu → **"Add component above"**

   ✅ **Expected**: The new component is inserted above, **expanded**, and **scrolled into view**

6. Repeat with **"Add component below"**

   ✅ **Expected**: Same behavior — expanded and scrolled into view

7. Manually collapse/expand components by clicking their headers

   ✅ **Expected**: Normal toggle behavior still works as before

#### After (fixed)
image

> The component stays expanded and the viewport scrolls to it automatically.

### Related issue(s)/PR(s)

Fix #26039


ace2016:
Hello @innerdvations  could you please take a look at this PR when you have a chance?
Let me know if you need me to clarify anything or make any adjustments!

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/Field.tsx
```diff
@@ -298,6 +317,12 @@ const DynamicZone = ({
                     onAddComponent={handleAddComponent}
                     dynamicComponentsByCategory={dynamicComponentsByCategory}
                     totalLength={dynamicDisplayedComponentsLength}
+                    forceOpen={openComponentKey === field.__temp_key__}
+                    onForceOpenHandled={() => {
```
mathildeleg: `DynamicComponent` is wrapped in `React.memo`, but the `onForceOpenHandled` prop is re-created as a new arrow function on every render of `DynamicZone`. That means memo never bails out — every component in the zone
  re-renders on every parent state change. In a long dynamic zone (the exact scenario this fix targets), that can noticeably affect responsiveness.

The fix is a stable `useCallback` lifted out of the `.map()`. Because `setOpenComponentKey` is a stable setter from `useState`, the deps array can be empty:
 ```
const handleForceOpenHandled = React.useCallback(() => {
    setOpenComponentKey(null);
  }, []);
```
Then in the render loop just pass `onForceOpenHandled={handleForceOpenHandled}`. 

As a bonus, the `if (openComponentKey === field.__temp_key__) `guard inside the inline version is actually redundant — `forceOpen` is only ever true on the matching component, so `onForceOpenHandled` can only fire when the condition holds anyway.

File: packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/Field.tsx [hunk 2]
```diff
@@ -76,6 +78,23 @@ const DynamicZone = ({
...
   const { value = [], error } = useField<Array<DzWithTempKey>>(name);
 
+  /**
+   * Track the previous value array to detect when a new component is added.
+   * When the array grows, we find the newly added item and force its accordion open.
+   * This mirrors the same pattern used in RepeatableComponent.
+   */
+  const prevValue = usePrev(value);
+
+  React.useEffect(() => {
```
mathildeleg, line 88: If a component is added (setting `openComponentKey`) but then immediately removed before the `forceOpen` effect fires in `DynamicComponent` — e.g. a quick accidental click — `openComponentKey` is never cleared by `onForceOpenHandled` and stays non-null indefinitely. It's harmless in practice (no component in the list will match the stale key), but it's a minor state leak that the next add will silently overwrite.

```suggestion
React.useEffect(() => {
    if (prevValue && prevValue.length < value.length) {
      const prevKeys = new Set(prevValue.map((v) => v.__temp_key__));
      const newItem = value.find((v) => !prevKeys.has(v.__temp_key__));
      if (newItem) {
        setOpenComponentKey(newItem.__temp_key__);
      }
    } else if (openComponentKey !== null) {
      // Component was removed before forceOpen was handled — clear stale key
      const currentKeys = new Set(value.map((v) => v.__temp_key__));
      if (!currentKeys.has(openComponentKey)) {
        setOpenComponentKey(null);
      }
    }
  }, [value, prevValue, openComponentKey]);
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
```diff
@@ -125,6 +139,24 @@ const DynamicComponent = ({
     }
   }, [rawError, hasValue, accordionValue]);
 
+  /**
+   * When the parent flags this component as newly added via `forceOpen`,
+   * expand the accordion and scroll it into view so the user can immediately
+   * start editing. Once handled, notify the parent so it can clear the flag.
+   */
+  React.useEffect(() => {
```
mathildeleg, line 162: Curious about the relationship between this new `forceOpen` path and the existing `isNewItem` mechanism a few lines above (line 132):
```
  const [collapseToOpen, setCollapseToOpen] = React.useState<string>(
    isNewItem ? accordionValue : ''
  );
```

`isNewItem` checks whether the component has no saved id, which is true for any freshly added component — so it already initializes the accordion open. You mention in the PR description that the component "immediately collapses after being added" but under what conditions `isNewItem` fails to prevent that? Is this a timing issue (form state not yet populated on first render)? A React StrictMode double-invoke?
 
Understanding the failure mode would help confirm the `forceOpen` path is the right fix rather than patching a symptom of something else. Could you explain why `isNewItem` alone wasn't sufficient?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/DynamicComponent.tsx
```diff
@@ -125,6 +139,24 @@ const DynamicComponent = ({
...
+    if (forceOpen) {
+      setCollapseToOpen(accordionValue);
+      requestAnimationFrame(() => {
+        componentRef.current?.scrollIntoView({
+          behavior: 'smooth',
+          block: 'center',
+        });
+      });
+      onForceOpenHandled?.();
```
mathildeleg: Minor semantic point: `onForceOpenHandled` is called synchronously here, which clears the parent's `openComponentKey` before the `requestAnimationFrame` callback fires the actual scroll. So the parent considers the action "handled" before it's physically complete. It works fine in practice (the rAF closure captures `componentRef` and doesn't need `forceOpen` to still be true), but it reads as if the order should be reversed.
Moving the callback inside the rAF (to l155 instead of l156) would make the sequence match the comment ("Once handled, notify the parent"):

```suggestion
        });
        onForceOpenHandled?.();
      });
    }
  }, [forceOpen, accordionValue, onForceOpenHandled]);
```

mathildeleg (commented):
Thanks for the PR! I have a few comments, and could you also add some tests (you may add them to `Field.test.tsx` or `DynamicComponent.test.tsx`)?
For example to test out: 
 - Adding a component to an empty zone → new component accordion is expanded and `scrollIntoView` is called once
  - Adding a component to a non-empty zone → only the new component scrolls, pre-existing ones are untouched
  - Using "Add component above" (mid-list insert) → the inserted component (not the last one) gets `forceOpen`
Thank you ✨

ace2016:
Hey @mathildeleg, thanks for the thorough review — all suggestions have been addressed:

 **Stability fixes:**
> - `onForceOpenHandled` is now wrapped in `useCallback` to keep the reference stable and avoid invalidating `React.memo` on `DynamicComponent`
> - Added the `else if` cleanup branch you suggested — if a component is removed before `forceOpen` fires, the stale `openComponentKey` is now cleared immediately
>
**Timing fix:**
> - Moved `onForceOpenHandled?.()` inside the `requestAnimationFrame` callback in `DynamicComponent`, so the parent is only notified *after* the scroll action has actually been dispatched
>
 **Tests:**
> Added 3 test cases covering the forceOpen flow end-to-end:
> 1. Add to empty zone → accordion auto-expands
> 2. Add to non-empty zone → only the new component expands
> 3. "Add component above" mid-list → inserted component expands
>
 Re: `forceOpen` vs `isNewItem` — `forceOpen` is intentionally kept as a separate mechanism because `isNewItem` only seeds the initial accordion state on first mount. In scenarios involving React StrictMode double-invocations or async form state updates, `isNewItem` can lead to unintended accordion collapse. The `forceOpen` pattern provides a reliable imperative "second pass" to guarantee the UX requirement is met.

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/Field.tsx
```diff
@@ -59,6 +60,7 @@ const DynamicZone = ({
 
   const [addComponentIsOpen, setAddComponentIsOpen] = React.useState(false);
   const [liveText, setLiveText] = React.useState('');
+  const [openComponentKey, setOpenComponentKey] = React.useState<number | null>(null);
```
mathildeleg:
```suggestion
  const [openComponentKey, setOpenComponentKey] = React.useState<string | null>(null);
```

mathildeleg (commented):
Sorry one last thing I didn't notice the first time, otherwise all good!

ace2016:
Good catch, updated! Thanks for the review👌

mathildeleg (dismissed):

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/Field.tsx
mathildeleg: Sorry I think the build is failing because this also needs to be changed to: 

```suggestion
      __temp_key__: string;
```

mathildeleg (commented):
Sorry one last small fix 🙈

ace2016:
alright i have addressed this

ace2016:
Hello @mathildeleg 
Just checking in on #26044 , all your feedback has been addressed. Happy to make any further adjustments if needed!
Also wanted to flag another PR I opened that you might be interested in:
#26062 — would love your thoughts when you have a moment. 🙏
Thanks again for the thorough review

innerdvations:
Hey @ace2016 , it looks like there's a test failing that is most likely an issue with the test itself rather than anything you did. I will try to investigate this week and see if I can fix it. Thanks for your patience!

ace2016:
thanks for looking for into this , ping me here if you need any additional context or testing from my ends 👌

innerdvations (approved):

unrevised6419:
https://github.com/strapi/strapi/pull/26065#issuecomment-4411681069

The failing `[EE] e2e (browser: firefox) shard 2/4` is a flake caused by an interaction between the develop merge and a pre-existing test, not the msw migration in this PR.

### Failing test
`tests/e2e/tests/content-manager/edit-view/single-type-edit-view.spec.ts` — *"as a user I want to add a component to a dynamic zone at a specific position"*. The test fires three back-to-back component insertions via `dispatchEvent('click')` on `more actions` → `add component above/below` → component name, then asserts the resulting order at indices 0..5.

### Root cause: #26044 (`2c551d43d8`)
The merged develop branch includes *fix: auto-expand and scroll to newly added dynamic zone components*. It introduces two async side effects on every insert that the test never waits for:

**`Field.tsx`** — new `useEffect` watches `value.length`, diffs `__temp_key__` sets to find the new item, and sets `openComponentKey` to its key.

**`DynamicComponent.tsx`** — each item receives `forceOpen={openComponentKey === field.__temp_key__}`:
```ts
React.useEffect(() => {
  if (forceOpen) {
    setCollapseToOpen(accordionValue);          // 1. expand accordion → height grows
    requestAnimationFrame(() => {
      componentRef.current?.scrollIntoView({    // 2. smooth scroll viewport
        behavior: 'smooth',
        block: 'center',
      });
      onForceOpenHandled?.();
    });
  }
}, [forceOpen, accordionValue, onForceOpenHandled]);
```

Plus a new `ref={componentRef}` on the `<li>` container.

### Why this breaks the test
1. **Accordion expand** is a layout shift — following `<li>`s move down, so the `more actions` button on `nth(0)`/`nth(1)` is no longer where Playwright resolved it a frame earlier.
2. **`scrollIntoView({ behavior: 'smooth' })`** runs an animation (~300–800 ms). The test immediately dispatches the next click. In Firefox, dispatching a click on a target whose layout is still moving from the previous smooth-scroll either times out (`locator.dispatchEvent: Timeout 10000ms exceeded`) or lands on stale coordinates and triggers the wrong menu — both failure modes appeared across retries (`Expected substring: "hero image"` on one retry, timeouts on the others).
3. `requestAnimationFrame` defers the scroll one frame after the expand, so the effects compound across the three rapid inserts.

Pre-#26044, an insert was a synchronous DOM update — no expand, no scroll — and the rapid `dispatchEvent` sequence was safe.

### Fix
After each insert, wait for the count to reach the expected value (lets the RAF + form-state update flush), and `scrollIntoViewIfNeeded()` on the next `more actions` button before clicking so it's re-anchored after the prior smooth-scroll moved it:

```ts
await page.getByRole('menuitem', { name: /product carousel/i }).dispatchEvent('click');
await expect(componentsLocator).toHaveCount(4);

const moreActionsBtn2 = componentsLocator.nth(0).getByRole('button', { name: /more actions/i });
await moreActionsBtn2.waitFor({ state: 'visible' });
await moreActionsBtn2.scrollIntoViewIfNeeded();
await moreActionsBtn2.click({ force: true });
```

Same treatment for the second and third insertions. This is a fix to the e2e test rather than to #26044's behavior, which is the desired UX.

Issue #25966, opened 2026-04-09
Strapi 5 internal migration 5.0.0-02-created-document-id can leave Postgres in a bad state after statement_timeout — retry fails with "document_id already exists"


akashxthakur:
### Node Version

20.18.0

### Package Manager

npm

### Package Manager Version

10.8.2

### Strapi Version

5.42.0

### Operating System

Strapi Cloud

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

During **Strapi 4 → 5** upgrade on **PostgreSQL**, internal migration **`5.0.0-02-created-document-id`** (`@strapi/database`, adds `document_id` to tables including **`files`**) can fail or be interrupted (e.g. **`statement_timeout`** on a large `files` table, or process killed mid-migration).

On the next start, Strapi runs the migration again and crashes with **Postgres “column document_id already exists”** on `files`, while the migration may not be recorded in **`strapi_migrations_internal`**. The app then **cannot boot** until the DB is repaired or the migration is patched to be idempotent.

This is **not** a REST API 500 from content fetch — it happens during **Strapi bootstrap / DB migration**.

### Steps to Reproduce

Start from a Strapi 4.x project on PostgreSQL with a non-trivial files / upload table (many rows helps).
Ensure Postgres has a finite statement_timeout (common on Strapi Cloud / managed Postgres).
Upgrade dependencies to Strapi 5.x and run strapi develop or strapi start so internal migrations run.
During bootstrap, migration 5.0.0-02-created-document-id runs (see @strapi/database → dist/migrations/internal-migrations/5.0.0-02-document-id.js).
Either:
the migration hits statement_timeout while running ALTER TABLE ... ADD COLUMN document_id on files, or the process is killed mid-migration; then
Start Strapi again. The migration runs again and attempts the same ALTER TABLE.

### Expected Behavior

- If a migration step is interrupted, a restart should either:

- roll back incomplete DDL so the DB matches “migration not applied”, or
- detect existing schema and continue safely (idempotent ADD COLUMN), or
- record progress in smaller steps so we never re-run a successful ADD COLUMN.
- After recovery, Strapi should boot and strapi_migrations_internal should reflect completed internal migrations.

### Logs

```shell
[...]

PR #26045, opened 2026-04-17
fix(database): make 5.0.0-02-created-document-id migration idempotent

butcherZ:
### What does it do?

Adds idempotency handling to the 5.0.0-02-created-document-id migration so it tolerates duplicate-column errors (PostgreSQL 42701, MySQL 1060) and always runs the document_id backfill even when the column already exists.

### Why is it needed?

During Strapi 4→5 upgrade on PostgreSQL, the internal migration 5.0.0-02-created-document-id can fail or be interrupted (e.g., statement_timeout on large files table, or process killed mid-migration). 
On retry, the migration crashes with "column document_id already exists" because the column was created but the migration wasn't recorded. Additionally, when hasColumn returns true, the migration early-exited via continue, silently skipping the backfill and leaving rows with NULL document_id.

### How to test it?

Run the new unit tests:
```bash
yarn workspace @strapi/database test:unit --testPathPattern="5.0.0-02-document-id"
```
All 5 tests should pass.

### Related issue(s)/PR(s)

#25966 
Fix CMS-689

### DIFF
--- a/packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
+++ b/packages/core/database/src/migrations/internal-migrations/5.0.0-02-document-id.ts
@@ -143,10 +143,25 @@ const migrationDocumentIds = async (db: Database, knex: Knex, meta: Meta) => {
   } while (updatedRows > 0);
 };
 
+const isDuplicateColumnError = (error: unknown): boolean => {
+  if (!error || typeof error !== 'object') return false;
+  const e = error as { code?: string; errno?: number; message?: string };
+  if (e.code === '42701') return true;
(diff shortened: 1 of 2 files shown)


innerdvations (commented):
I have reviewed and tested and this looks fine. However, I want to wait a bit to merge until I can investigate why this is necessary. Throwing an error if the column already exists is possibly intended behaviour to prevent migrations running multiple times which is not expected and not tested -- we'd have to review all the migrations and make sure they're safe to "resume" like this.

innerdvations (approved):
Ok, ignore the previous comment, this is still a safe improvement; we shouldn't be relying on a crash to stop it in any case.

Issue #25199, opened 2026-01-15
Memory leak on the /upload REST api route


OcneanuVlad:
### Node Version

22.17.0

### NPM/Yarn/PNPM Version

10.9.2

### Strapi Version

5.32.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

I'm experiencing what appears to be a memory leak when uploading files through the Strapi REST API. Memory consumption increases with each file upload and never decreases.

### Steps to Reproduce

1. Start a Strapi project locally.
2. Open Postman (or another API testing tool) and prepare to repeatedly call the upload endpoint: http://localhost:1337/api/upload (you can also use a simple script to automate this).
3. Open Task Manager (or any monitoring tool) and track the memory usage of the Node/Strapi process.
4. While watching memory usage, start uploading files via the API in a loop / multiple times.
5. Once memory usage increases, stop the uploads and observe that the memory does not drop back down.

### Expected Behavior

After a file is successfully uploaded, the memory used to process that file should be released and available for garbage collection, allowing memory usage to return to baseline levels during periods of inactivity.

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

OcneanuVlad:
Internal note: TID 5266

Marc-Roig:
Some related work we did in the past just in case its useful

https://github.com/strapi/strapi/pull/16978
https://github.com/strapi/strapi/pull/20080

Arc-a-it:
@OcneanuVlad  
This doesnt look like a classic leak caused by a single forgotten reference

Whats happening here appears to be a **missing cleanup boundary in the upload execution pipeline** which causes **request scoped resources to accumulate across uploads**

Structurally flow is this :

- `admin-upload.ts` receives the request and orchestrates the upload
    
- `upload.ts` performs validation, processing, and persistence
    
- Temporary resources (streams, buffers, temp files, image transforms) are allocated during processing
    

The problem is that while the _forward_ flow (request → processing → storage) is well-defined there is **no explicit reverse phase that owns resource cleanup** after the upload completes

As a result it become:

- Streams/buffers created during uploads remain reachable longer than the request lifecycle
    
- Repeated uploads amplify retained memory
    
[...]

Issue #24745, opened 2025-10-30
High resource usage and possible crashing when bulk uploading media assets


derrickmehaffy:
### Node Version

v22

### NPM/Yarn/PNPM Version

Yarn v1

### Strapi Version

5.x.x+

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

N/A

### Bug Description

When bulk uploading assets, it's possible to exhaust the application's resources (especially on Strapi Cloud with lower tier plans), this appears to be primarily caused by the following:

- Bulk assets are uploaded all at once, for example if you upload 10 assets, all 10 of these upload requests are fired in parallel rather than batching or uploading one by one
- Image optimization (breakpoint conversion via Sharp) also happens on all of these assets at the same time, with extremely large or high-res images this can quickly spike CPU usage to maximum and possibly causing some assets to fail uploading or even crash the application

### Steps to Reproduce

1. Create a small Strapi instance on Strapi Cloud (essential tier for example)
2. Try to upload various amounts of assets at the same time (start with 5, 10, 15, 25, 50, ect)
3. Vary the size (increase size as you go)
4. Work with the Strapi Support team to monitor application resource usage during the test (if you don't have access to the backoffice to view it yourself)

### Expected Behavior

Assets should have a configuration option to determine how they should be uploaded depending the application need (if we decide to batch bulk uploads). Else assets should be uploaded one at a time until the array of assets has finished.

### Logs

```shell

```

### Media

Example usage from one cloud user's application related to this (these rapid spikes are in relation to bulk file upload, confirmed by the user):

<img width="1238" height="656" alt="Image" src="https://github.com/user-attachments/assets/87447f84-8c14-42fb-8d33-45efe5a69ec3" />

### Additional information

Related to TID: 4570

Possible duplicate issues related to this (thus converting it to a tracking issue):

- https://github.com/strapi/strapi/issues/16210

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #26046, opened 2026-04-17
fix(upload): sharp concurrency and cache leads to OOM 

markkaylor:
### What does it do?

  - Serializes responsive breakpoint generation (large, medium, small) so resizes run sequentially instead of in parallel via Promise.all
  - Disables sharp's internal decoded image cache (sharp.cache(false)) and limits libvips thread pool to 1 (sharp.concurrency(1)) at plugin registration
  - Makes both sharp settings configurable via config/plugins.ts (sharpCache, sharpConcurrency)

 ### Why is it needed?

Uploading a large image (e.g. 50MP) causes high peak memory usage because sharp/libvips decodes the full source image for each responsive breakpoint resize concurrently, while also retaining cached pixel data between operations. On constrained environments like Strapi Cloud (512MB), this causes the process to be OOM-killed before it can respond.

###  How to test it?

1. Deploy a strapi cloud free project (it has constrained memory allocation)
2. Enable responsive friendly in upload settings
3. Upload a large image like this one [github.com/user-attachments/assets/04e623b0-a7f2-4ca5-8f06-68abc1fb990d](https://github.com/user-attachments/assets/04e623b0-a7f2-4ca5-8f06-68abc1fb990d)
4. It should throw an error
5. Redeploy on this experimental 0.0.0-experimental.c2a30c30861681f520db3b7a9528750b09ac561e
6. Re-test and it should work

The sharp cache and concurrency is now configurable for more powerful environments

```
  upload: {
    config: {
      sharp: {
        cache: true,
        concurrency: 2
      }
    },
  },
```

  Related issue(s)/PR(s)

  - Relates to #25199
  - Relates to #24745

### DIFF
--- a/packages/core/upload/server/src/config.ts
+++ b/packages/core/upload/server/src/config.ts
@@ -4,6 +4,8 @@ export const config = {
     provider: 'local',
     sizeLimit: 1000000000, // 1GB
     actionOptions: {},
+    sharpCache: false,
+    sharpConcurrency: 1,
   },
   validator() {},
 };
--- a/packages/core/upload/server/src/config.ts
+++ b/packages/core/upload/server/src/config.ts
@@ -4,8 +4,10 @@ export const config = {
     provider: 'local',
     sizeLimit: 1000000000, // 1GB
     actionOptions: {},
-    sharpCache: false,
-    sharpConcurrency: 1,
+    sharp: {
+      cache: false,
+      concurrency: 1,
+    },
   },
   validator() {},
 };
--- a/packages/core/upload/server/src/__tests__/register.test.ts
+++ b/packages/core/upload/server/src/__tests__/register.test.ts
@@ -92,7 +92,7 @@ describe('Upload plugin register', () => {
       jest.clearAllMocks();
     });
 
-    test('uses sharpCache false and sharpConcurrency 1 from defaults', async () => {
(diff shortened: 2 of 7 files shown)


packages/core/upload/server/src/config.ts
```diff
@@ -4,6 +4,8 @@ export const config = {
     provider: 'local',
     sizeLimit: 1000000000, // 1GB
     actionOptions: {},
+    sharpCache: false,
+    sharpConcurrency: 1,
```
markkaylor: should we keep it flat or store it on a sharp object?


```suggestion
	sharp: {
	  cache: false,
	  concurrency: 1,
	}
```

innerdvations: I’d keep sharpCache and sharpConcurrency at the top level of upload config for this PR to keep it simple and avoid changing the shape in case it's a breaking change for anyone

innerdvations (dismissed):
I didn't do extensive testing on if this actually prevents the oom bug that @nclsndr was seeing today, but it works, and I added some tests that theoretically show that it's working, but the code LGTM

packages/core/upload/server/src/config.ts
markkaylor: Sounds good. Just to be clear this config is new so this is the opportunity to shape it how we want. That being said I'm cool with leaving it flat.

innerdvations (approved):

PR #26049, opened 2026-04-17
fix(content-manager): skip draft/modified count queries for non-D&P content types

### LINKED ISSUE #26048: Admin Homepage : getCountDocuments runs expensive self-join on collections with draftAndPublish: false
### Node Version

22.14.0

### Package Manager

npm

### Package Manager Version

10.2.4

### Strapi Version

5.13.0

### Operating System

Strapi Cloud

### Database

MySQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When open a collection from admin UI the Content Manager homepage service (`getCountDocuments`) executes a `modifiedDocuments` self-join query on **all** permitted content types, including those with `draftAndPublish: false`.

The `publishedDocuments` count correctly checks `meta.hasDraftAndPublish` before running, but the `modifiedDocuments` query does not have this guard:

```js
// This runs on ALL content types, regardless of draftAndPublish
const modifiedDocuments = await strapiDBConnection(tableName)
  .select('draft.document_id')
  .from(`${tableName} as draft`)
  .join(`${tableName} as published`, function () {
    this.on('draft.document_id', '=', 'published.document_id')
      .andOn('draft.updated_at', '!=', 'published.updated_at')
      .andOnNull('draft.published_at')
      .andOnNotNull('published.published_at');
  })
  .countDistinct('draft.document_id as count')
  .groupBy('draft.document_id')
  .first();
```

When `draftAndPublish: false`, `published_at` is **always** set (never null), so the `andOnNull('draft.published_at')` condition guarantees zero results. The query is semantically meaningless but still forces a full self-join on the table.

On large tables this is devastating — in my case it blocks the database for **~1100+ seconds**.

### Steps to Reproduce

1. Create a collection type with `"draftAndPublish": false` in the schema
2. Populate it with a large number of rows (1M+)
3. Open the Content Manager in the admin panel
4. Observe the slow query in the DB process list (mine for example):

```sql
SELECT `draft`.`document_id`, COUNT(DISTINCT `draft`.`document_id`) AS `count`
FROM `nodis` AS `draft`
INNER JOIN `nodis` AS `published`
  ON `draft`.`document_id` = `published`.`document_id`
  AND `draft`.`updated_at` != `published`.`updated_at`
  AND `draft`.`published_at` IS NULL
  AND `published`.`published_at` IS NOT NULL
GROUP BY `draft`.`document_id`
LIMIT 1
```

### Expected Behavior

The `modifiedDocuments` self-join should be skipped for content types with `draftAndPublish: false`, just like `publishedDocuments` already does. The fix should be a simple guard:

```js
if (meta.hasDraftAndPublish) {
  // run modifiedDocuments query
}
```

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

nicosh:
## What does this PR do?

Wraps the `draftDocuments` and `modifiedDocuments` queries in `getCountDocuments` with a `meta.hasDraftAndPublish` guard, so they are skipped for content types that have `draftAndPublish: false`.

Currently, the `publishedDocuments` query already has this guard and falls back to a simple `countDistinct`, but the other two queries run unconditionally on every content type. On large tables this causes a full self-join that can block the database for minutes.

## Why is it needed?

When `draftAndPublish` is disabled, `published_at` is **always** set (never `NULL`). This means:

- The `draftDocuments` query (`WHERE published_at IS NULL … HAVING COUNT(*) = 1`) will always return 0.
- The `modifiedDocuments` self-join (`ON draft.published_at IS NULL AND published.published_at IS NOT NULL`) will always return 0.

Both queries are semantically meaningless for non-D&P content types, but the self-join is extremely expensive on large tables. In our case, a collection with several million rows blocks the database for **~1100 seconds** every time the Content Manager homepage is loaded.

## Changes

**File:** `packages/core/content-manager/server/src/homepage/services/homepage.ts`

- Wrapped the `draftDocuments` query block with `if (meta.hasDraftAndPublish)`
- Wrapped the `modifiedDocuments` query block with `if (meta.hasDraftAndPublish)`

This is consistent with how `publishedDocuments` is already handled in the same function.

## How to verify

1. Create a collection type with `"draftAndPublish": false` and populate it with a large number of rows
2. Before fix: open the Content Manager → observe the slow self-join query in the DB process list
3. After fix: the query is skipped entirely for that collection

## Related issue(s)/PR(s)
- Closes #26048


Akash504-ai (commented):
Just to understand, since both draftDocuments and modifiedDocuments rely on meta.hasDraftAndPublish, is there a reason they r  handled in separate guarded blocks instead of grouping them under a single condition?

nicosh:
@Akash504-ai I kept them separate to match the existing pattern and minimize changes to the current structure.

markkaylor (approved):
Thanks for the fix, looks good to me. I'll merge once the tests pass

PR #26053, opened 2026-04-17
chore(upload): add concurrentUploadSize config

markkaylor:
### What does it do?  
                                                                                                                                                                    
Replaces parallel upload processing with a sequential batches that are processed in parallel. 
Adds a config to control the batch size. Defaults to batch size 1 (no parallel processing)
                                                                                                                                                                    
  ### Why is it needed?                                                                                                                                             
   
Bulk uploads could crash under load because every file was processed in parallel via a single `Promise.all`, overwhelming the provider / filesystem / memory.   

This gives control to the user so they can trade speed for memory.
                                                                                                                    
              
  ### How to test it?

  1. **Baseline (default behavior):** upload multiple files via the Media Library or the `/upload` endpoint without any extra config. Monitor memory on the process to ensure it doesn't keep growing. 
  2. **Custom concurrency:** set the option in `config/plugins.ts`:                                                                                                 
     ```ts                                                                                                                                                          
     export default {
       upload: {                                                                                                                                                    
         config: { concurrentUploadSize: 5 },                                                                                                                       
       },
     };                                                                                                                                                             
     ```      
     Upload multiple files and confirm they succeed.
  3. **Invalid config:** set an invalid value (`0`, `-1`, `"5"`, `2.5`) and confirm Strapi fails to boot with a validation error referencing `concurrentUploadSize`.
  4. **Unit tests:** run `yarn test:unit` inside `packages/core/upload`.                                                                                            
                                                                                                                                                                    
  ### Related issue(s)/PR(s)                                                                                                                                        

Relates to https://github.com/strapi/strapi/pull/26046
Relates to https://github.com/strapi/strapi/issues/24745


markkaylor:
@derrickmehaffy if there is anyway you could validate this is beneficial that'd be great then I'll reach out for a code review from others (feel free to review the code yourself though!) 🙏 let me know if you wanna sync.

innerdvations (dismissed):
LGTM!

I pushed a test that ensures the concurrency limit is actually working

jhoward1994 (approved):

PR #26054, opened 2026-04-17
enhancement(provider-amazon-ses): replace node-ses with AWS SDK SESClient

sjoukedv:
### What does it do?

Replaces `node-ses` with the official AWS SDK v3.

### Why is it needed?

Reduce 3rd party dependencies and this also adds support for IAM roles for Service Accounts (and the default credential chain of the AWS SDK).


innerdvations:
Thanks for the PR!  I've made a few fixes to prevent breaking changes in the configuration and added some tests to confirm it. I also add some commits from https://github.com/strapi/strapi/pull/25354

jhoward1994 (approved):

PR #26055, opened 2026-04-17
fix(core/admin | content-manager): combine multi-role field-level permissions

### LINKED ISSUE #26012: Multi-role field-level permissions broken — CASL union loses fields from other roles
#### Describe the bug

When a user belongs to **multiple admin roles** (e.g., Editor + Manager), field-level permissions for the same content type are not correctly unioned. Only one role's field list is applied, and the other role's fields are silently dropped. This affects both reading and updating — users see "No permissions to see this field" for fields they should have access to through their other role.

Three interconnected issues cause this:

**1. `DocumentRBAC.tsx:87` — Permission replacement instead of accumulation**

```typescript
// packages/core/content-manager/admin/src/features/DocumentRBAC.tsx L85-88
return contentTypePermissions.reduce<Record<string, Permission[]>>((acc, permission) => {
  const [action] = permission.action.split('.').slice(-1);
  return { ...acc, [action]: [permission] };  // ← Replaces instead of appending
}, {});
```

When `findUserPermissions()` returns permissions from multiple roles for the same action+subject, the reduce overwrites with `[permission]` (new single-element array) each iteration. Only the **last** role's permission survives.

**Expected:** `{ ...acc, [action]: [...(acc[action] || []), permission] }`

**2. `DocumentRBAC.tsx:168-174` — `undefined` fields (meaning "all fields") silently dropped**

```typescript
const extractAndDedupeFields = (permissions: Permission[] = []) =>
  permissions
    .flatMap((permission) => permission.properties?.fields)  // undefined from unrestricted perms
    .filter(
      (field, index, arr): field is string =>
        arr.indexOf(field) === index && typeof field === 'string'  // ← drops undefined
    );
```

If one role's permission has `properties.fields: undefined` (all fields allowed) and another has `properties.fields: ['title']`, the result is `['title']` — the unrestricted permission is lost. When `fields` is `undefined`, the function should short-circuit and return all fields (or signal that no restriction applies).

**3. `permission-fields.ts:44` — `fieldsFrom` callback destroys "unrestricted" semantics in CASL**

```typescript
// packages/core/admin/server/src/services/permission/permissions-manager/permission-fields.ts L43-44
const permittedFields = permittedFieldsOf(ability, actionOverride, subject, {
  fieldsFrom: (rule) => rule.fields || [],  // ← undefined (unrestricted) becomes [] (zero fields)
});
```

In CASL, `rule.fields = undefined` means "no field restriction" (all fields allowed). The `|| []` converts this to an empty array, meaning "zero fields allowed." Combined with the `shouldIncludeAll` guard at L47-51, this means a role with unrestricted field access combined with a role that has explicit field restrictions results in only the restricted fields being visible.

#### Steps to reproduce

1. Create two admin roles: "Editor" and "Manager"
2. For "Editor" role, give `read` permission on a content type with fields: `[title, content]`
3. For "Manager" role, give `read` permission on the same content type with fields: `[processingStatus, network]`
4. Assign a user to **both** roles
5. Log in as that user and navigate to the content type in the content manager
6. Observe that only one role's fields are visible, not the union of both

#### Expected behavior

The user should see fields from **both** roles: `[title, content, processingStatus, network]`. CASL is designed to union permissions from multiple rules, but the frontend code discards all but the last permission before CASL can process them.

#### Environment

- Strapi: v5 (latest develop branch)
- Affected files:
  - `packages/core/content-manager/admin/src/features/DocumentRBAC.tsx`
  - `packages/core/admin/server/src/services/permission/permissions-manager/permission-fields.ts`

yazan-abu-obaideh:
### What does it do?

This PR addresses #26012. It fixes logic on the backend and frontend side where field-level permissions are not combined correctly. 

### Why is it needed?

Users are not permitted to see fields that they should have access to (e.g. a user who should be able to see all fields under 'article' can only see the title). 

### How to test it?

Tests were written to reproduce the cases. These tests now pass.

### Related issues / PRs

- Fixes #26012

### Manual QA

---

## Prerequisites

1. Run Strapi from `examples/getstarted` on the **fix branch** (`fix-26012`):
   ```bash
   cd examples/getstarted
   yarn develop
   ```
2. Log in as **Super Admin** (`admin@strapi.io` / `Welcome1`).
3. Ensure **Article** collection type exists (ships with getstarted). Fields used below:
   - `title`, `markdownContent`, `authorName`, `cover`, `slug`, `categories`
4. Create at least one Article entry with values in all fields above (so hidden fields are obvious).

---

## Scenario 1 — Multi-role field union (primary regression)

**Covers:** Bug 1 — permissions from multiple roles must accumulate, not replace.

### Setup

| Step | Action |
|------|--------|
| 1 | **Settings → Administration panel → Roles** → Create role **Editor** |
| 2 | In **Collection Types → Article**, enable **Read** |
| 3 | Under **Fields → Read**, check only **`title`** and **`markdownContent`** |
| 4 | Save role |
| 5 | Create role **Manager** |
| 6 | In **Collection Types → Article**, enable **Read** |
| 7 | Under **Fields → Read**, check only **`authorName`** and **`cover`** |
| 8 | Save role |
| 9 | **Settings → Administration panel → Users** → Create user `multi-role@test.com` |
| 10 | Assign **both** Editor and Manager roles. Save. |

### Reproduce

| Step | Action |
|------|--------|
| 1 | Log out. Log in as `multi-role@test.com` |
| 2 | **Content Manager → Article** → open existing entry |

### Expected (fixed)

- All four fields visible and readable: `title`, `markdownContent`, `authorName`, `cover`
- No "No permissions to see this field" on fields granted by the *other* role

### Broken (pre-fix)

- Only **one** role's fields visible (whichever role was processed last)
- e.g. only `authorName` + `cover`, or only `title` + `markdownContent`
- Other fields show permission denied state

---

## Scenario 2 — Unrestricted role + restricted role

**Covers:** Bug 2 (frontend) + Bug 3 (backend CASL) — `fields: undefined` means "all fields", must not be narrowed by a sibling restricted role.

### Setup

| Step | Action |
|------|--------|
| 1 | Create role **Full Access** |
| 2 | **Collection Types → Article** → enable **Read** at the content-type row |
| 3 | Do **not** narrow field checkboxes — leave Fields matrix as "all fields allowed" (no per-field restriction) |
| 4 | Save role |
| 5 | Create role **Title Only** |
| 6 | **Collection Types → Article** → enable **Read** |
| 7 | Under **Fields → Read**, check only **`title`** |
| 8 | Save role |
| 9 | Create user `unrestricted@test.com` with **both** Full Access and Title Only roles |

### Reproduce

| Step | Action |
|------|--------|
| 1 | Log in as `unrestricted@test.com` |
| 2 | Open an Article entry in Content Manager |

### Expected (fixed)

- **All** Article fields visible (unrestricted role wins)
- `slug`, `categories`, `cover`, etc. all accessible

### Broken (pre-fix)

- Only `title` visible (restricted role incorrectly narrowed the union)
- Other fields blocked despite Full Access role

---

## Scenario 3 — Update permissions (same union logic)

**Covers:** Issue states bug affects **read and update**, not just read.

### Setup

Extend Scenario 1 roles:

| Role | Additional permission |
|------|----------------------|
| Editor | **Update** → fields `title`, `markdownContent` |
| Manager | **Update** → fields `authorName` |

Use same `multi-role@test.com` user.

### Reproduce

| Step | Action |
|------|--------|
| 1 | Log in as `multi-role@test.com` |
| 2 | Open Article entry → switch to edit mode |

### Expected (fixed)

- Can edit `title`, `markdownContent`, `authorName`
- `cover` read-only (only in Manager read, not in update set) — or hidden if update not granted

### Broken (pre-fix)

- Can only edit fields from one role's update permission set

### DIFF
--- a/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
+++ b/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
@@ -1,8 +1,9 @@
-import { AbilityBuilder, Ability } from '@casl/ability';
+import { AbilityBuilder, Ability, subject } from '@casl/ability';
 import { pick } from 'lodash/fp';
 import sift from 'sift';
 import { buildStrapiQuery } from '../permission/permissions-manager/query-builders';
 import createPermissionsManager from '../permission/permissions-manager';
+import { createPermissionFieldsCache } from '../permission/permissions-manager/permission-fields';
 
 const allowedOperations = [
   '$or',
@@ -313,6 +314,23 @@ describe('Permissions Manager', () => {
     // @ts-expect-error
     test.each(tests)(`Test n°%#: %s`, (name, input, expected) => {
       expect(buildStrapiQuery(input)).toStrictEqual(expected);
+    });
+  });
+
+  // #26012: an unrestricted rule (fields: undefined) mixed with a restricted rule must
+  // still produce shouldIncludeAll = true, not be narrowed to the restricted fields.
+  describe('Bug #3 — rule.fields || [] collapses unrestricted rule to zero fields', () => {
+    test('shouldIncludeAll is true when any rule has no field restriction', () => {
+      const ability = defineAbility((can: any) => {
+        can('read', 'Article', undefined);
+        can('read', 'Article', ['title']);
+      });
+
+      const { getPermissionFields } = createPermissionFieldsCache(ability);
+      const result = getPermissionFields('read', subject('Article', {}));
+
+      expect(result.shouldIncludeAll).toBe(true);
+      expect(result.permittedFields).toEqual([]);
     });
   });
 });
--- a/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
+++ b/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
@@ -328,6 +328,7 @@ describe('Permissions Manager', () => {
       const result = getPermissionFields('read', subject('Article', {}));
 
       expect(result.shouldIncludeAll).toBe(true);
+      expect(result.hasAtLeastOneRegistered).toBe(true);
       expect(result.permittedFields).toEqual([]);
     });
   });
--- a/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
+++ b/packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
@@ -317,9 +317,7 @@ describe('Permissions Manager', () => {
     });
   });
 
-  // #26012: an unrestricted rule (fields: undefined) mixed with a restricted rule must
-  // still produce shouldIncludeAll = true, not be narrowed to the restricted fields.
-  describe('Bug #3 — rule.fields || [] collapses unrestricted rule to zero fields', () => {
+  describe('unrestricted rules combined with restricted rules', () => {
     test('shouldIncludeAll is true when any rule has no field restriction', () => {
       const ability = defineAbility((can: any) => {
         can('read', 'Article', undefined);
--- a/packages/core/content-manager/admin/src/features/DocumentRBAC.tsx
+++ b/packages/core/content-manager/admin/src/features/DocumentRBAC.tsx
@@ -84,7 +84,7 @@ const DocumentRBAC = ({ children, permissions, model }: DocumentRBACProps) => {
     );
     return contentTypePermissions.reduce<Record<string, Permission[]>>((acc, permission) => {
       const [action] = permission.action.split('.').slice(-1);
-      return { ...acc, [action]: [permission] };
+      return { ...acc, [action]: [...(acc[action] ?? []), permission] };
     }, {});
   }, [contentTypeUid, userPermissions]);
 
@@ -165,13 +165,11 @@ const DocumentRBAC = ({ children, permissions, model }: DocumentRBACProps) => {
 /**
  * @internal it's really small, but it's used three times in a row and DRY for something this straight forward.
  */
-const extractAndDedupeFields = (permissions: Permission[] = []) =>
-  permissions
-    .flatMap((permission) => permission.properties?.fields)
-    .filter(
-      (field, index, arr): field is string =>
-        arr.indexOf(field) === index && typeof field === 'string'
-    );
+const extractAndDedupeFields = (permissions: Permission[] = []) => {
+  const allFields = permissions.flatMap((permission) => permission.properties?.fields);
+  if (allFields.some((field) => field === undefined)) return [];
+  return allFields.filter((field, index, arr): field is string => arr.indexOf(field) === index);
+};
 
 /**
  * @internal removes numerical strings from arrays.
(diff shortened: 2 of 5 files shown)


packages/core/content-manager/admin/src/features/DocumentRBAC.tsx
```diff
@@ -165,13 +165,11 @@ const DocumentRBAC = ({ children, permissions, model }: DocumentRBACProps) => {
...
  */
-const extractAndDedupeFields = (permissions: Permission[] = []) =>
-  permissions
-    .flatMap((permission) => permission.properties?.fields)
-    .filter(
-      (field, index, arr): field is string =>
-        arr.indexOf(field) === index && typeof field === 'string'
-    );
+const extractAndDedupeFields = (permissions: Permission[] = []) => {
+  const allFields = permissions.flatMap((permission) => permission.properties?.fields);
```
mathildeleg, line 169: This early return is correct, but a one-liner comment on it would prevent the next person from "fixing" it back to a filter:
```suggestion
  const allFields = permissions.flatMap((permission) => permission.properties?.fields);
   // An undefined entry means this permission grants access to all fields
    // (no field-level restriction). Returning [] signals "no restriction" to callers.
```

packages/core/admin/server/src/services/__tests__/permissions-manager.test.ts
```diff
@@ -315,4 +316,21 @@ describe('Permissions Manager', () => {
       expect(buildStrapiQuery(input)).toStrictEqual(expected);
     });
   });
+
+  // #26012: an unrestricted rule (fields: undefined) mixed with a restricted rule must
+  // still produce shouldIncludeAll = true, not be narrowed to the restricted fields.
+  describe('Bug #3 — rule.fields || [] collapses unrestricted rule to zero fields', () => {
```
mathildeleg: This describe blocks label reads more like a commit note than a durable test description. Could you rename to describe it correctly?

packages/core/content-manager/admin/src/features/tests/DocumentRBAC.test.tsx
```diff
@@ -152,6 +152,98 @@ describe('DocumentRBAC', () => {
     });
   });
 
+  // #26012: canReadFields must be the union of all roles' fields, not just the last role's.
+  describe('Bug #1 — multi-role field union: two restricted roles', () => {
```
mathildeleg: Again, could you rename this describe label and the one below to have a proper label?
Also you can remove the issue number in the commented line here and below :)

mathildeleg (commented):
Thanks for the PR, I see it's still in draft but you seem to have finished working on it so I took the time to review it :)
I have a few minor comments but overall the fix looks good to me! @nclsndr could you take a look too to make sure?
Can you please also update the PR description by adding a few manual QA tests that could be done when we do the release QA? Thank you ✨

yazan-abu-obaideh:
Thanks for addressing the review comments @nclsndr. Life got in the way and I never got back to this PR. Please let me know if anything else is needed from me.

nclsndr:
No big deal @yazan-abu-obaideh! I had quite a lot on my plate as well, I'll carry on the last changes for the PR to merge asap. Thank you so much for proposing this fix 👌

nclsndr (approved):

PR #26059, opened 2026-04-17
fix: 403 error when select role when creating a user

### LINKED ISSUE #23622: Error Alert “Warning: Policy Failed” When Selecting Role During User Creation in Admin Dashboard
### Node Version

20.12.2

### NPM/Yarn/PNPM Version

10.5.0

### Strapi Version

5.14.0 and 5.13.x

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When creating a new user entry via the admin dashboard (Content Manager → User), an error alert appears after selecting a role. The alert shows the message "Warning: Policy Failed", even though the user entry is successfully created. This error seems to occur when trying to fetch the selected role details, resulting in a 403 error.

This behavior does not block user creation, but the error alert can be misleading and confusing to admins.

### Steps to Reproduce

1. Go to Content Manager → Select User collection (from users-permissions plugin).
2. Click Create new entry.
3. Fill in user details.
4. In the role input field, select any available role.
5. An error alert appears with the message: Warning: Policy Failed.

### Expected Behavior

The role should be selected without any alert or error. No error message should appear unless the operation genuinely fails. The UI should not trigger a Policy Failed warning if the operation is permitted and successful.

### Logs

```shell
[2025-05-29 12:04:19.350] http: GET /content-manager/collection-types/plugin::users-permissions.role/txlnxvs13dz4ujwrsiva4928? (40 ms) 403
```

### Code Snippets

_No response_

### Media

Image

### Additional information

This issue was found on a fresh Strapi project with no custom modifications, and the action was performed using a Super Admin account

### Confirmation Checklist

Adzouz:
### What does it do?

Registers all content types (including non-displayed ones) as valid subjects for the `plugin::content-manager.explorer.read` action in `registerPermissions()`.
Previously only content types with `pluginOptions['content-manager'].visible !== false` were registered.

### Why is it needed?

Content types marked `visible: false` (e.g. plugin::users-permissions.role) are hidden from the CM sidebar navigation, but they can appear as relation targets on displayed content types. When a user selects such a relation in the edit view, the CM frontend fetches the related document via GET `/content-manager/collection-types/:model/:id`, which is protected by the hasPermissions policy.

Because plugin::users-permissions.role was never registered as a valid subject for `explorer.read`, resetSuperAdminPermissions never generated a DB permission row for it so even super admins had no read permission on that model.

Result: selecting a role when creating a user in the Content Manager triggered a 403 Policy Failed alert, even for super admins on a fresh project with no custom configuration.

False flag is a UI/navigation concern, not a security boundary. Registering non-displayed types as subjects for `explorer.read` corrects the permission generation without affecting Create, Update, Delete, or Publish actions.

### How to test it?

- Start Strapi and log in as Super Admin
- Go to Content Manager → User (from plugin::users-permissions)
- Click Create new entry
- Fill in username, email, password
- In the Role field, select any available role (e.g. Authenticated)
-> No "Warning: Policy Failed" toast should appear.
- Save the entry — user is created with the correct role

Before this fix: selecting a role triggered a 403 on GET `/content-manager/collection-types/plugin::users-permissions.role/:id` and displayed a "Policy Failed" error notification.

### Related issue(s)/PR(s)

Fixes #23622

🚀

### DIFF
--- a/packages/core/content-manager/server/src/services/permission.ts
+++ b/packages/core/content-manager/server/src/services/permission.ts
@@ -23,6 +23,8 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
     const displayedContentTypes = getService('content-types').findDisplayedContentTypes();
     const contentTypesUids = displayedContentTypes.map(prop('uid'));
 
+    const allContentTypesUids = getService('content-types').findAllContentTypes().map(prop('uid'));
+
     const actions = [
       {
         section: 'contentTypes',
@@ -39,7 +41,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
         displayName: 'Read',
         uid: 'explorer.read',
         pluginName: 'content-manager',
-        subjects: contentTypesUids,
+        subjects: allContentTypesUids,
         options: {
           applyToProperties: ['fields'],
         },
--- a/packages/core/content-manager/server/src/services/permission.ts
+++ b/packages/core/content-manager/server/src/services/permission.ts
@@ -20,10 +20,11 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
   },
 
   async registerPermissions() {
-    const displayedContentTypes = getService('content-types').findDisplayedContentTypes();
-    const contentTypesUids = displayedContentTypes.map(prop('uid'));
-
-    const allContentTypesUids = getService('content-types').findAllContentTypes().map(prop('uid'));
+    const allContentTypes = getService('content-types').findAllContentTypes();
+    const allContentTypesUids = allContentTypes.map(prop('uid'));
+    const contentTypesUids = allContentTypes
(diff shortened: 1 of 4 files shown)


packages/core/content-manager/server/src/services/permission.ts
```diff
@@ -23,6 +23,8 @@ export default ({ strapi }: { strapi: Core.Strapi }) => ({
     const displayedContentTypes = getService('content-types').findDisplayedContentTypes();
     const contentTypesUids = displayedContentTypes.map(prop('uid'));
 
+    const allContentTypesUids = getService('content-types').findAllContentTypes().map(prop('uid'));
```
nclsndr: I dunno how cosmetic this would be but: what about deriving `displayedContentTypes` from `allContentTypesUids`? 
Depends on filter specificity / cost to exec.

PR #26060, opened 2026-04-17
test(migration): version migration testing framework

innerdvations:
## What does it do?

Adds an automated **v4 → workspace** migration integration test for the monorepo:

- CLI: `yarn test:migrations --initial legacy` (or `--scenario tests/migration/scenarios/v4-to-head.json`)
- Fixture unit tests: `yarn test:migrations:unit` (also run via `yarn test:unit`)
- Fixture app + validators: `examples/complex` + `tests/migration/fixture/`
- CI job **`migration_v5`**: latest v4 from `@strapi/strapi@legacy` → workspace validation on sqlite, postgres, mysql, mariadb (Node 20)
- Local smoke: `yarn test:migrations:smoke` (sqlite, `--skip-build` when dist is current)

## Why is it needed?

Catches regressions in internal migrations and post-migration data shape (document_id, discard-drafts, relations, components, dynamic zones) using a realistic fixture — without relying on manual upgrade testing alone.

## How to test it?

**Smoke (fast, sqlite):**

```bash
yarn test:migrations:smoke
```

**Fixture unit tests (fast):**

```bash
yarn test:migrations:unit
```

**Plan only (no network / installs):**

```bash
yarn test:migrations:plan --initial legacy
```

**Full run:**

```bash
yarn test:migrations --initial legacy --database postgres --skip-build
```

**Build:** ensure workspace packages are built (`yarn build`). Use **Node 20** for the v4 scaffold if install/seed warns on newer Node.

**CI trigger:** job runs when the `migrations` path filter matches (see `.github/filters.yaml`): `packages/**/migrations/**`, `tests/migration/**`, and `examples/complex` harness paths.

## Related issue(s)/PR(s)

Fixes [CMS-1070](https://linear.app/strapi/issue/CMS-1070/migration-testing-framework)

---

**Reviewer notes:**

- Canonical scenario is `tests/migration/scenarios/v4-to-head.json` — matches what CI runs via `--initial legacy`.
- Nested yarn installs in `.migration-v5/` set `YARN_ENABLE_IMMUTABLE_INSTALLS=false` (PR hardened mode / `YN0028`).
- Validators are modular under `tests/migration/fixture/checks/`; expectations driven by `tests/migration/fixture/spec.js`.
- v4 scaffold uses `client: 'mysql2'` for mysql/mariadb legs (Strapi v4 passes client straight to Knex; no mysql→mysql2 alias like v5).


nehaaprasad:
### AI Code Trust — ship readiness

**Score:** 100 · **Verdict:** SAFE · **Model:** deterministic-v1

No blocking issues detected by automated checks.

**Top issues:**
- **maintainability** (medium) — File is very long (1218 lines); consider splitting. `examples/complex/scripts/validate-migration.js:1`
- **maintainability** (low) — Very long line reduces readability. `examples/complex/scripts/validate-migration.js:1020`

nclsndr:
can't we author TS instead of JS for the tests @innerdvations ?

innerdvations:
Agreed, that would be better. I'll convert it.

innerdvations:
@nclsndr done, all typescript now!

tests/migration/framework/stage-pinned-v5.ts
```diff
@@ -0,0 +1,60 @@
...
+  version: string;
+  dbEnv: DbEnv;
+};
+
+async function runPinnedStrapiStage(ctx: MigrationContext, stage: PinnedStage): Promise<void> {
+  const { version } = stage;
+  const pinnedRoot = path.join(ctx.MIGRATION_ROOT, 'pinned-v5', version);
+
+  console.log(`\n📌 Pinned Strapi ${version}: preparing app at ${pinnedRoot}...`);
+  await execa('node', [path.join(ctx.COMPLEX_DIR, 'scripts', 'setup-pinned-v5-project.js')], {
```
Bassel17: I tried the `--via` path locally with `yarn test:migrations --initial 4.26.1 --via 5.30.0 --database sqlite --skip-build`. The v4 setup and seed completed, but this then failed with `MODULE_NOT_FOUND` because the script was converted to `.ts` and no `.js` file is emitted here. Can we invoke `setup-pinned-v5-project.ts` through `node --import tsx`, like the v5 baseline does, and add coverage for this path?

.github/actions/check-pr-status/lint-staged.config.mjs
```diff
@@ -1 +1,17 @@
...
+export default {
+  ...shared,
+  [jsPattern]: (filenames) => {
+    const filtered = filenames.filter((filename) => !filename.includes('/dist/'));
+
+    if (!filtered.length) {
+      return [];
+    }
+
+    return jsTasks.flatMap((task) => filtered.map((filename) => `${task} ${filename}`));
```
Bassel17: It looks like this override was added to skip `dist/` when the shared task was still an array. After the newer `develop` changes, `shared[jsPattern]` is now an async function, so this throws `TypeError: jsTasks.flatMap is not a function`. I reproduced it by importing this config and invoking the exported JS/TS task directly. Is this override still needed? If yes, it should delegate with `jsTasks(filtered)`; otherwise we can go back to the shared-config re-export.

Bassel17 (changes requested):
The main v4 → workspace smoke test passes locally and the migration CI coverage looks useful. I just found two issues

Bassel17 (commented):
Follow-up re-review: the two previously reported blockers are addressed. I found three remaining harness correctness issues below.

tests/migration/scripts/run-migration-scenario.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  assertCliInitialOrScenario();
+
+  await runWorkspaceBuilds(REPO_ROOT, {
+    fullBuild: argv.build,
+    skipBuild: argv.skipBuild,
+  });
+
+  assertNodeMajor(argv.initialNode, 'initial-node');
```
Bassel17: `--initial-node` and `--workspace-node` both validate `process.version` in this same runner process. This means the documented `--initial-node 20 --workspace-node 24` command cannot succeed: Node 20 fails the workspace check, while Node 24 fails this initial check. Could we either execute each phase with its requested runtime, or replace these with a single host-runtime guard and remove the two-runtime example?

tests/migration/scripts/run-migration-scenario.ts
```diff
@@ -0,0 +1,358 @@
...
+  .option('print-plan', {
+    type: 'boolean',
+    default: false,
+    describe:
+      'Only resolve and print the migration plan (JSON) and exit: no build, no Docker, no yarn in nested apps',
+  })
+  .help()
+  .parse();
+
+const database = argv.database || process.env.DATABASE_CLIENT || 'sqlite';
```
Bassel17: `database` and `multiplier` are resolved here before the migration `.env` file is loaded inside `run()`. Additionally, yargs gives `multiplier` a default of `1`, so the documented `MIGRATION_MULTIPLIER` / `SEED_MULTIPLIER` defaults are masked when the CLI flag is omitted. Could we load dotenv before finalizing these values and apply CLI flag -> environment -> hardcoded default precedence?

tests/migration/scripts/run-migration-scenario.ts
```diff
@@ -0,0 +1,358 @@
...
+    throw new Error('Invalid scenario: missing baseline');
+  }
+  const t = scenario.baseline.type;
+  if (t !== 'v4-scaffold' && t !== 'v5-pinned') {
+    throw new Error('Invalid scenario: baseline.type must be "v4-scaffold" or "v5-pinned"');
+  }
+  if (!scenario.baseline.initialVersion) {
+    throw new Error('Invalid scenario: baseline.initialVersion is required');
+  }
+  if (!Array.isArray(scenario.stages)) {
```
Bassel17, line 217: This currently accepts `stages: []`, pinned-only stages, and workspace stages with `validate: []`. Those plans can reach the success message without running any workspace validator. Since the runner promises that every migration ends at and validates the workspace, could we require exactly one terminal workspace stage with a non-empty validator list?

tests/migration/scripts/run-migration-scenario.ts
innerdvations: Agreed — dual majors cannot work in one process. Collapsed to a single host-runtime guard (`--initial-node`; `--workspace-node` is an alias that must match) and removed the two-runtime example.

_AI-assisted comment._

tests/migration/scripts/run-migration-scenario.ts
innerdvations: Fixed: load migration dotenv before resolving `database` / `multiplier`, dropped the yargs `multiplier` default of `1` so env vars are not masked, and apply CLI → env → hardcoded default precedence.

_AI-assisted comment._

tests/migration/scripts/run-migration-scenario.ts
innerdvations: Agreed. `assertScenarioShape` now requires a non-empty stages list, exactly one terminal `workspace` stage, and a non-empty `validate` array.

_AI-assisted comment._

Bassel17 (approved):
LGTM! should be good to go now

PR #26062, opened 2026-04-18
Fix/webhook publish empty entry 25387

### LINKED ISSUE #25387: Publish event for Webhooks has empty entry
### Node Version

20.20.0

### NPM/Yarn/PNPM Version

npm 10.9.4

### Strapi Version

5.35.0

### Operating System

Linux (Debian/Ubuntu)

### Database

MariaDB

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

On a Collection type 
- with draft and publish logic enabled
- and two possible languages
- and a dynamic zone for some components
the behavior of update and publish is non deterministic.

Pressing the Save Button creates an update event, with the entry.
In the entry the publishedAt Date is always null, not the last published date - but perhaps that is intended.

Pressing the Publish Button creates a publish event.
In the event JSON the entry is mostly the String "null", instead of the expected entry.
Sometimes we get there an entry, with the expected content, but can not reproduce which fields should be changed to trigger that.

The behavior we see since any versions.

### Steps to Reproduce

1. Create a Content Type with Title and Description, and a relation to another Content Type
2. Create one or more Components
3. Add to the first Content Type a dynamic zone, for the Components
4. Add a Webhook for Entries, for create update, delete publish and unpublish
5. Implement a receiver for the post
6. Use Strapi in Productionmode
7. Edit content, Publish, with Saving in Front, without Saving before Publish
8. Track the entry value of the events

### Expected Behavior

The sent publish event, should always send the published entity and not null.

### Logs

```shell
At Strapi:
[2026-02-06 14:30:13.705] http: POST /content-manager/collection-types/api::content-nugget.content-nugget/h49w5av9ulv13bdzwzdd2dy2/actions/publish (5882 ms) 200

At the webhook receiver:
{"event":"entry.publish","createdAt":"2026-02-06T14:30:13.498Z","model":"content-nugget","entry":"null"}
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

I assume there is a serialization error at some circumstances. The content element can have a lot of deeply nested components

### Confirmation Checklist

ace2016:
### What does it do?

Fixes a race condition in the webhook event emission pipeline that caused `entry.publish` events to send `"entry": "null"` instead of the actual published entry data.

**The core change:** Moves the `findOne` re-fetch (with deep populate) and sanitization from the `onCommit` callback into the current transaction context, where the newly-created published entry is always visible. Only the `eventHub.emit()` call is deferred to `onCommit`.

#### Changes

**`packages/core/core/src/services/document-service/events.ts`**
- The `emitEvent` method is now `async`
- Entry population (`findOne` with deep populate) and sanitization now run **within** the enclosing transaction
- Only the `eventHub.emit()` is deferred to `onCommit`, ensuring webhook consumers see fully committed data

**`packages/core/core/src/services/document-service/repository.ts`**
- All 8 `emitEvent` call sites updated from fire-and-forget (`.forEach(emitEvent(...))`) to properly awaited (`await` / `await Promise.all(....map(...))`) so population completes within the transaction before commit

### Why is it needed?

When publishing a content entry with draft-and-publish enabled, Strapi:

1. **Deletes** the old published version
2. **Creates** a brand-new published entry
3. **Emits** the `entry.publish` webhook event

Previously, **all three steps** of the event emission (re-fetch → sanitize → emit) happened inside an `onCommit` callback — i.e., *after* the transaction committed. The `findOne` re-fetch used a fresh database connection from the pool. On some databases (notably **MariaDB/MySQL with InnoDB**), the newly-committed row was not yet visible to this fresh connection, causing `findOne` to return `null`.

This resulted in webhooks receiving:
```json
{"event":"entry.publish","model":"content-nugget","entry":"null"}
```

The bug was **non-deterministic** — it happened "mostly" but sometimes the entry was present — consistent with a transaction visibility race condition. It was more frequent with deeply nested components and dynamic zones due to longer transaction times.

#### Impact

- **Affected events:** Primarily `entry.publish`, but the fix applies to all event types for consistency
- **Affected databases:** Most reproducible on MariaDB/MySQL; theoretically possible on any database
- **Breaking changes:** None — webhook payloads now reliably contain the entry data that was previously intermittently missing

#### Environment
- Strapi `5.x` with draft-and-publish enabled
- Any database (MariaDB recommended for strongest verification)

#### Steps

1. Start Strapi and set up a webhook receiver:
   ```bash
   npx -y http-echo-server 9000
   ```

2. In Strapi Admin → **Settings → Webhooks → Create new webhook**:
   - URL: `http://localhost:9000`
   - Events: ✅ **Publish**

3. Create a content type with a dynamic zone (or use an existing one)

4. **Content Manager** → Create a new entry → fill in fields → **Save** → **Publish**

5. Check the webhook receiver output:

   **✅ Expected (with fix):**
   ```json
   {
     "event": "entry.publish",
     "model": "about",
     "uid": "api::about.about",
     "entry": {
       "id": 2,
       "documentId": "i51ife4zlquo0rfs2alztqff",
       "publishedAt": "2026-04-18T13:12:24.603Z",
       "title": "fashion",
       "dynamiczone": []
     }
   }
   ```

   **❌ Before fix (the bug):**
   ```json
   {
     "event": "entry.publish",
     "model": "content-nugget",
     "entry": "null"
   }
   ```

6. Repeat publish/unpublish several times to confirm consistency (the old bug was non-deterministic)

#### Automated tests
- All 54 existing document-service tests pass
- Lint and Prettier pass

### Related issue(s)/PR(s)

Fix #25387


ace2016:
Hello @mathildeleg  @innerdvations   following up on this PR. PR has been open since Apr 18 with no reviewer assigned. Would appreciate if you could take a look at it , thanks :)

PR #26063, opened 2026-04-19
feat(admin): Make Entry Action card sticky in Content Manager

themodernmonk7:
### What does it do?

I have updated the `ContentManager` layout to improve the accessibility of the entry actions. Specifically:
* Wrapped the `<Panels />` component (which contains the Save, Publish, and Entry information) in a `<Box />` component.
* Applied `position="sticky"` with a `top={5}` offset to the container.
* Ensured the sticky behavior is scoped within the `{isDesktop}` condition to prevent layout breakage on mobile devices where the sidebar stacks vertically.

### Why is it needed?

Currently, the Strapi Content Manager requires users to scroll back to the top of the page to save or publish changes. For pages with many dynamic components or long-form content, this creates a repetitive and frustrating user experience.

By making the Action Panels sticky, we ensure that:
1.  **Primary actions** (Save/Publish) are always visible.
2.  **Content editing flow** is uninterrupted, especially on complex pages with many blocks.
3.  **Efficiency** is increased for content editors who frequently toggle between editing and saving.

### How to test it?

1.  Open the **Content Manager** in a local development environment.
2.  Select a **Single Type** (e.g., an "About") that has multiple content blocks or dynamic zones.
3.  On a desktop screen, scroll down through the content.
4.  **Verification:** The right-hand "Entry" sidebar should remain fixed at the top of the viewport (`top={5}`) as you scroll.
5.  **Responsiveness:** Verify that on mobile viewports, the sticky behavior is disabled and the panels stack naturally at the bottom of the content.

**[Screen Recording here]**

### Before

https://github.com/user-attachments/assets/389bf303-7108-47dd-94ba-904ff26c92ba

### After

https://github.com/user-attachments/assets/af987a0c-14d3-46ae-ac7d-6809b341137d

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/EditViewPage.tsx
@@ -239,7 +239,9 @@ const EditViewPage = () => {
                 </Grid.Item>
                 {isDesktop && (
                   <Grid.Item col={3} direction="column" alignItems="stretch">
-                    <Panels />
+                    <Box position="sticky" top={5} width="100%">
+                      <Panels />
+                    </Box>
                   </Grid.Item>
                 )}
               </Grid.Root>


mathildeleg:
Thanks for the contribution — sticky action panels is a genuinely useful UX improvement and the intent here is right!

Unfortunately we can't merge this as-is for a couple of reasons:
 1. The sidebar contains more than just the Entry Actions card:
`<Panels />` renders every registered side panel as a stacked column — in an EE instance that includes Review Workflows, Content Releases, and the Preview panel on top of the Entry card. Wrapping the entire column in a sticky container only works if that column is shorter than the viewport. When all panels are visible the combined height easily exceeds the viewport, at which point `position: sticky` silently has no effect. The fix would need to target just the actions panel (Save/Publish), not the full Panels component, which I don't think is the correct way to go.
2. An ancestor overflow breaks sticky positioning:
 The `OverflowingItem` wrapper in our root layout has `overflow-x: hidden.` Any ancestor with a non-visible overflow value becomes the sticky element's scroll container in CSS — and since `OverflowingItem` doesn't scroll, `position: sticky` anchors to it rather than the viewport. This means the sticky behaviour likely won't work reliably across browsers regardless of panel count.

I'll close the PR but will raise your UX improvement request to our Product Designer :)

PR #22615, opened 2025-01-10
Chore/upgrade-msw-v2

jhoward1994:
### What does it do?

Describe the technical changes you did.

### Why is it needed?

Describe the issue you are solving.

### How to test it?

Provide information about the environment and the path to verify the behaviour.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #21796, opened 2024-10-11
chore: upgrade msw to 2.4.10

innerdvations:
### What does it do?

upgrades msw from 1.3.0 to 2.4.10

### Why is it needed?

CVE-2024-47764

### How to test it?

Provide information about the environment and the path to verify the behaviour.

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request

PR #26065, opened 2026-04-19
chore(deps): migrate msw 1.x → 2.13.4 across admin test suites

unrevised6419:
## Summary
- Migrate `msw` 1.3.x → 2.13.4 across 8 workspaces (admin, content-manager, content-releases, upload, review-workflows, plugin-i18n, plugin-documentation, plugin-users-permissions). The email workspace had a `msw` devDep with zero source usage and is dropped here, not bumped.
- Codemod 37 test handlers from v1's `rest.X((req, res, ctx) => res(ctx.json(X)))` shape to v2's `http.X(({ request, params }) => HttpResponse.json(X))` per the [official 1.x → 2.x guide](https://mswjs.io/docs/migrations/1.x-to-2.x).
- `typedoc` is intentionally **not** changed in this PR (the bump from the initial commit was reverted).

## What changed

**Handler shape**
- `rest.X(p, (req, res, ctx) => res(ctx.json(X)))` → `http.X(p, () => HttpResponse.json(X))`
- `res(ctx.status(N))` → `new HttpResponse(null, { status: N })` for non-200; empty 200s use plain `new HttpResponse()` (semantically identical to v1's `res(ctx.status(200))`).
- `res.once(...)` → `http.X(p, handler, { once: true })`
- `res.networkError(...)` → `HttpResponse.error()`
- `(await req.json()) as Body` casts → `http.METHOD<Params, Body>(...)` generics. Path-param-less routes use `<Record<string, never>, Body>` (msw v2 convention; `{}` is blocked by `@typescript-eslint/ban-types`).
- `RequestHandler` → `HttpHandler`

**Realm-agnostic SyntaxError check (`getFetchClient.ts`)**
- `responseInterceptor` previously matched `error instanceof SyntaxError` to recognize empty-200 responses (where `response.json()` throws). Constructor identity differs across JS realms — under jsdom + Node fetch, an empty 200 throws a Node-realm `SyntaxError` that fails the jsdom-realm `instanceof` check. Same bug exists in browsers when a Response originates from a service worker / iframe.
- Switched to `error?.name === 'SyntaxError'` — realm-agnostic. Net effect: empty 200s now roundtrip cleanly without the `HttpResponse.json({})` workaround in test handlers.

**Jest setup (per msw frequent-issues)**
- `jest-preset.front.js`: `customExportConditions: ['']` so `msw/node` resolves under jsdom; widened `transformIgnorePatterns` for ESM-only msw deps (`msw`, `@mswjs`, `rettime`, `@open-draft`, `outvariant`, `until-async`, `strict-event-emitter`, `headers-polyfill`, `is-node-process`); `.mjs` matched by swc transform. **Trade-off:** each added entry is one more swc-transform pass per test file on cold runs.
- `packages/admin-test-utils/src/environment.ts`: polyfill Node fetch + streams + `BroadcastChannel` into the jsdom global. Polyfills are installed in `setup()`, snapshotted into a per-instance `originalGlobals` map, and restored in `teardown()` so a test that mutates one of them cannot leak into the next file sharing a Jest worker. **Intentionally** leaves `Blob`/`File`/`FormData`/`URL` as jsdom's — Strapi app code threads these through jsdom-only APIs (`FileReader`, `Image`, `HTMLFormElement`, `URL.createObjectURL`) where realm matters; comment block at the top of env.ts documents the trade-offs in detail.

## Why
MSW v1 is unmaintained; v2 is the supported line. The migration unblocks future work on the test suite and lets newer dev-dep upgrades land (peer ranges). Driven by the still-open deprecation-warning issue #21201.

## Prior art
Three previous attempts at this migration were closed without merging:
- #21796 (Oct 2024, @innerdvations) — bump to msw 2.4.10, motivated by CVE-2024-47764. Closed.
- #21617 (Jan 2025, @nilsingwersen) — bump to msw 2 + `rest` → `http`/`HttpResponse` migration. Refs #21201. Closed.
- #22615 (Mar 2026, @jhoward1994) — "Chore/upgrade-msw-v2". Closed (empty description).

This PR carries the same goal and adds three blockers that the earlier attempts didn't have: (1) the jest-environment polyfills (with deliberate scope so jsdom-only flows still work), (2) the realm-agnostic `SyntaxError` check in `getFetchClient.ts`, and (3) the cross-realm `File` workaround in the LogoInput test (msw v2 response bodies cross realms via Node fetch; `new File([nodeBlob], …)` from jsdom can't read Node Blob bytes — handler returns raw bytes instead so axios reconstructs a jsdom Blob via the global constructor).

## Pre-existing failures (not introduced here)
Confirmed against `develop` via `git stash` + rerun:
- `admin/src/features/tests/Widgets.test.tsx` — "should handle network errors gracefully"
- `ee/admin/src/hooks/tests/useLicenseLimits.test.ts` — 2 fails
- `content-manager/admin/src/pages/EditView/components/FormInputs/tests/UID.test.tsx` — 3 fails

All four files are untouched by this PR; all fail identically on `develop`.

## Test plan
- [x] `yarn test:ts:front` — TypeScript clean across all 11 projects
- [x] Per-workspace `yarn nx run @strapi/<workspace>:test:front --skip-nx-cache` for each touched workspace — pass parity vs `develop` baseline (zero net regressions; only the pre-existing failures listed above remain).
- [x] msw-touched test files in each workspace match `develop` baseline (msw 1.3.5) test-by-test.


unrevised6419:
## Code Review

**Overall:** Mechanical migration, well-scoped. The polyfill comments in `environment.ts` and the cross-realm `File` workaround in `LogoInput.test.tsx` are exemplary — these are exactly the artifacts that blocked the three previous attempts (#21796, #21617, #22615). Test pass parity vs the `develop` baseline is confirmed in the PR description.

> **Update (resolved across 7b6b815 and a7dd149):** `originalGlobals`/teardown refactor, fetch trade-off note, `<Record<string, never>, …>` typing convention, email `package.json` revert, realm-agnostic `SyntaxError` check in `getFetchClient.ts`, empty 200s back to `new HttpResponse()`, URL polyfill drop. Remaining items below are kept as the original review record.

### Behavior risks

- **`packages/core/admin/admin/tests/server.ts`** — ✅ resolved in a7dd149. Empty 200 handlers now return `new HttpResponse()` (semantically identical to v1's `res(ctx.status(200))`). The `HttpResponse.json({})` workaround is no longer needed because `getFetchClient.ts:293` matches `error?.name === 'SyntaxError'` instead of `instanceof SyntaxError`, which is realm-agnostic.
- **`packages/admin-test-utils/src/environment.ts:56`** — ✅ resolved in 7b6b815. The comment block (now restructured to the top of the file in a7dd149) documents why Node's undici `fetch` is required (MSW interceptor patches the global) and why Strapi's cookie-via-`document.cookie` model is unaffected.
- **`environment.ts:9-49`** — ✅ resolved in 7b6b815. Globals are installed in `setup()`, snapshotted into `originalGlobals`, and restored in `teardown()` — no leak path across files in a Jest worker.

### Convention is undocumented

- ✅ resolved in a7dd149. Empty bodies are now `new HttpResponse()` everywhere (no two-shape convention to explain).

### Typing nits

- ✅ resolved in 7b6b815. All eight `<never, Body>` sites use `<Record<string, never>, Body>` (`{}` blocked by `@typescript-eslint/ban-types`).
- `packages/plugins/i18n/admin/tests/server.ts` (delete handler) — `http.delete<{ id: string }>` is fine in isolation, but the third generic (response body) is implicit. Worth a shared helper if a follow-up wave is planned.

### Out-of-scope dep bump

✅ resolved in 7b6b815. `msw` removed from `packages/core/email/package.json` entirely (zero source usage in the workspace). `yarn.lock` carries a single `msw@2.13.4` entry.

### Build / config

- **`jest-preset.front.js`** — `transformIgnorePatterns` was widened with 9 ESM-only packages (`msw`, `@mswjs`, `rettime`, `@open-draft`, `outvariant`, `until-async`, `strict-event-emitter`, `headers-polyfill`, `is-node-process`). Each added package incurs an extra swc-transform pass per test file. Likely small but measurable on cold runs — worth calling out as a known trade-off in the PR description.
- **`jest-preset.front.js:10`** — `'^.+\\.m?js(x)?$'` covers `.mjs`, but `.cjs` falls through. Verify msw's transitive `.cjs` deps don't break under Jest's default transform. The test plan covers this implicitly, but explicit confirmation is cheap.

### Pre-existing failures

The PR notes pre-existing baseline failures (`Widgets` net-err, `useLicenseLimits` ×2, `UID` ×3) — confirmed identical on `develop` via `git stash` + rerun. Linking the existing issues (or filing them) would help reviewers avoid assuming the migration introduced these regressions.

### LogoInput cross-realm workaround

`LogoInput.test.tsx:251` allocates `new Uint8Array(1024 * 1024 + 1).fill(0x31)` per test run. The inline comment explains why: msw v2's response body crosses realms (jsdom handler → Node fetch → axios), and `new File([nodeBlob], …)` (jsdom File constructor) cannot read a Node-realm Blob (falls back to `String(blob)` → 13 bytes). Routing raw bytes through ArrayBuffer (realm-neutral) lets axios reconstruct a jsdom Blob via the global `Blob` constructor, which app code can then wrap into a jsdom `File` correctly. Keep as-is.

---

### Verdict

LGTM — merge candidate. Remaining nice-to-haves: shared msw-handler helper for empty-body / typed responses, `transformIgnorePatterns` perf note in PR description, link to pre-existing baseline failures.

packages/core/email/package.json
```diff
@@ -80,7 +80,6 @@
     "@types/lodash": "^4.14.191",
     "koa": "2.16.4",
     "koa-body": "6.0.1",
-    "msw": "1.3.0",
```
unrevised6419, line 83: Note: not used in this workspace

unrevised6419:
@innerdvations can you please approve again the workflows to run?

packages/core/admin/admin/src/features/tests/Widgets.test.tsx
```diff
@@ -382,8 +382,10 @@ describe('useWidgets', () => {
         expect(mockSetFilteredWidgets).toHaveBeenCalled();
       });
 
-      // Wait for error notification to prevent act warnings from Sonner
-      await screen.findByText('Network error');
+      // Wait for error notification to prevent act warnings from Sonner.
+      // MSW v2's `HttpResponse.error()` surfaces as undici's `TypeError: Failed to fetch`
+      // (no custom message, unlike v1's `res.networkError('Network error')`).
+      await screen.findByText('Failed to fetch');
```
jhoward1994: This exact text seems like a tricky thing to match on and could change without us knowing?

Maybe we could assert the toast container appears without checking the message text.

unrevised6419: Addressed

jhoward1994:
Hi, this mainly looks good for me. Would you be able to apply the diff below? Sorry part of it contradicts my previous comment, I've just been trying to get my head around all of the changes.

I've also added a "don't merge" flag, just for now. And opened this PR on the main strapi fork https://github.com/strapi/strapi/pull/26093

It's a slight refactor of how the test env setup is done, I'm hoping to make it as easily maintainable for everyone in the future.

Once we've reviewed that internally hopefully I'll remove the merge flag and we can go ahead with this. Thanks again, it's really great work 🙏🏻

Here's the diff I mentioned:
```
diff --git a/packages/core/admin/admin/src/features/tests/Widgets.test.tsx b/packages/core/admin/admin/src/features/tests/Widgets.test.tsx
index ca7fd967e9..81173eaaf7 100644
--- a/packages/core/admin/admin/src/features/tests/Widgets.test.tsx
+++ b/packages/core/admin/admin/src/features/tests/Widgets.test.tsx
@@ -1,5 +1,5 @@
 import { server } from '@tests/server';
-import { waitFor, renderHook } from '@tests/utils';
+import { waitFor, renderHook, screen } from '@tests/utils';
 import { http, HttpResponse } from 'msw';
 
 import {
@@ -359,12 +359,7 @@ describe('useWidgets', () => {
         expect(mockSetFilteredWidgets).toHaveBeenCalled();
       });
 
-      // Wait for a Sonner toast to render. Asserting the container (rather
-      // than the message text) keeps the test stable across copy changes and
-      // across msw v1/v2 fetch-error message differences.
-      await waitFor(() => {
-        expect(document.querySelector('[data-sonner-toast]')).toBeInTheDocument();
-      });
+      await screen.findByText('Server error');
     });
 
     it('should handle network errors gracefully', async () => {
@@ -386,10 +381,11 @@ describe('useWidgets', () => {
         expect(mockSetFilteredWidgets).toHaveBeenCalled();
       });
 
-      // Wait for a Sonner toast to render (see note above).
-      await waitFor(() => {
-        expect(document.querySelector('[data-sonner-toast]')).toBeInTheDocument();
-      });
+      // Under msw v1 the network-error message was the string passed to
+      // `res.networkError(...)`. Under msw v2 + undici the thrown `TypeError`
+      // surfaces with messages like "Failed to fetch" / "fetch failed" /
+      // "Network Error" depending on how axios wraps it. Match any of them.
+      await screen.findByText(/failed to fetch|fetch failed|network ?error/i);
     });
   });
 
diff --git a/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/tests/ConfirmBulkActionDialog.test.tsx b/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/tests/ConfirmBulkActionDialog.test.tsx
index fcfab8e4a0..7affc749b6 100644
--- a/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/tests/ConfirmBulkActionDialog.test.tsx
+++ b/packages/core/content-manager/admin/src/pages/ListView/components/BulkActions/tests/ConfirmBulkActionDialog.test.tsx
@@ -105,7 +105,8 @@ describe('ConfirmDialogPublishAll', () => {
             },
             { status: 200 }
           );
-        }
+        },
+        { once: true }
       )
     );
 
@@ -127,7 +128,8 @@ describe('ConfirmDialogPublishAll', () => {
             },
             { status: 200 }
           );
-        }
+        },
+        { once: true }
       )
     );
 
@@ -151,7 +153,8 @@ describe('ConfirmDialogPublishAll', () => {
             },
             { status: 500 }
           );
-        }
+        },
+        { once: true }
       )
     );
 
@@ -180,7 +183,8 @@ describe('ConfirmDialogPublishAll', () => {
           return HttpResponse.json({
             data: 2,
           });
-        }
+        },
+        { once: true }
       )
     );
 
diff --git a/packages/core/content-type-builder/jest.config.js b/packages/core/content-type-builder/jest.config.js
index 2f40df7cf1..3c7cbb43da 100644
--- a/packages/core/content-type-builder/jest.config.js
+++ b/packages/core/content-type-builder/jest.config.js
@@ -7,5 +7,5 @@ module.exports = {
   },
   testMatch: ['<rootDir>/**/__tests__/*.test.ts'],
   displayName: 'Core content-type-builder',
-  testEnvironment: '@strapi/admin-test-utils/environment',
+  testEnvironment: 'jsdom',
 };


```

unrevised6419:
Hey @jhoward1994, I applied the diff

innerdvations (dismissed):
Thank you so much for this! I'm happy to merge this, if you can please fix the merge conflicts with develop. And optionally if you don't mind, cherry-pick the refactor from #26093

unrevised6419:
@innerdvations will do in a few minutes

innerdvations (dismissed):

unrevised6419:
@innerdvations it seems that firefox e2e failed 🤔 does not seem related to this PR
Claude suggests that it was introduced by #26044

image

unrevised6419:
The failing `[EE] e2e (browser: firefox) shard 2/4` is a flake caused by an interaction between the develop merge and a pre-existing test, not the msw migration in this PR.

### Failing test
`tests/e2e/tests/content-manager/edit-view/single-type-edit-view.spec.ts` — *"as a user I want to add a component to a dynamic zone at a specific position"*. The test fires three back-to-back component insertions via `dispatchEvent('click')` on `more actions` → `add component above/below` → component name, then asserts the resulting order at indices 0..5.

### Root cause: #26044 (`2c551d43d8`)
The merged develop branch includes *fix: auto-expand and scroll to newly added dynamic zone components*. It introduces two async side effects on every insert that the test never waits for:

**`Field.tsx`** — new `useEffect` watches `value.length`, diffs `__temp_key__` sets to find the new item, and sets `openComponentKey` to its key.

**`DynamicComponent.tsx`** — each item receives `forceOpen={openComponentKey === field.__temp_key__}`:
```ts
React.useEffect(() => {
  if (forceOpen) {
    setCollapseToOpen(accordionValue);          // 1. expand accordion → height grows
    requestAnimationFrame(() => {
      componentRef.current?.scrollIntoView({    // 2. smooth scroll viewport
        behavior: 'smooth',
        block: 'center',
      });
      onForceOpenHandled?.();
    });
  }
}, [forceOpen, accordionValue, onForceOpenHandled]);
```

Plus a new `ref={componentRef}` on the `<li>` container.

### Why this breaks the test
1. **Accordion expand** is a layout shift — following `<li>`s move down, so the `more actions` button on `nth(0)`/`nth(1)` is no longer where Playwright resolved it a frame earlier.
2. **`scrollIntoView({ behavior: 'smooth' })`** runs an animation (~300–800 ms). The test immediately dispatches the next click. In Firefox, dispatching a click on a target whose layout is still moving from the previous smooth-scroll either times out (`locator.dispatchEvent: Timeout 10000ms exceeded`) or lands on stale coordinates and triggers the wrong menu — both failure modes appeared across retries (`Expected substring: "hero image"` on one retry, timeouts on the others).
3. `requestAnimationFrame` defers the scroll one frame after the expand, so the effects compound across the three rapid inserts.

Pre-#26044, an insert was a synchronous DOM update — no expand, no scroll — and the rapid `dispatchEvent` sequence was safe.

### Fix
After each insert, wait for the count to reach the expected value (lets the RAF + form-state update flush), and `scrollIntoViewIfNeeded()` on the next `more actions` button before clicking so it's re-anchored after the prior smooth-scroll moved it:

```ts
await page.getByRole('menuitem', { name: /product carousel/i }).dispatchEvent('click');
await expect(componentsLocator).toHaveCount(4);

const moreActionsBtn2 = componentsLocator.nth(0).getByRole('button', { name: /more actions/i });
await moreActionsBtn2.waitFor({ state: 'visible' });
await moreActionsBtn2.scrollIntoViewIfNeeded();
await moreActionsBtn2.click({ force: true });
```

Same treatment for the second and third insertions. This is a fix to the e2e test rather than to #26044's behavior, which is the desired UX.

unrevised6419:
I did merge again `develop` here because of some conflicts

innerdvations:
Thanks! Yeah, the e2e test is flaky, we will run it again and treat fixing it in another PR. I think this is good to merge as soon as we get it passing.

innerdvations (approved):

Issue #22599, opened 2025-01-08
Documentation plugin generates OpenAPI with incorrect ID parameter for PUT and DELETE requests


otociulis:
### Required System information

- Node.js version: 18.19.0
- NPM version: 10.2.3
- Strapi version: 5.6.0
- Database: mysql
- Operating system: Linux
- Is your project Javascript or Typescript: Typescript

### Describe the bug

`full_documentation.json` file containing OpenAPI documentation contains incorrect `id` parameter definition. Per documentation https://docs.strapi.io/dev-docs/api/rest#update and https://docs.strapi.io/dev-docs/api/rest#delete the ID of resource should be `documentId` (string type) but generated documentation defines it as number:

```
"parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "",
            "deprecated": false,
            "required": true,
            "schema": {
              "type": "number"
            }
          }
        ],
```

### Steps to reproduce the behavior

1. Create any collection type
2. Install Documentation plugin
3. Generate `full_documentation.json` file documentation

### Expected behavior

Parameter should be defined as `string` to match `documentId` type:

```
"parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "",
            "deprecated": false,
            "required": true,
            "schema": {
              "type": "number"
            }
          }
        ],
```

### Screenshots

![image](https://github.com/user-attachments/assets/ffc5ff9d-03a3-4ab8-a83d-5e0287a5054b)

robdecker:
The same goes for GET requests.

### Screenshot

![Image](https://github.com/user-attachments/assets/2eee4e6e-8cba-4d82-8cfb-8c623b709f4b)

wpfilf:
+1 :)

christianmaleszka03:
+1

sergipt:
Any estimate on when this issue will be resolved?
Thanks.

StipJey:
+1

Misha327:
Is still an issue when using a fresh installation using `npx create-strapi@latest`. Anyone know how to resolve?

bleakas20:
wtf is it okay to have swagger not working for more than year?
its major issue

PR #26067, opened 2026-04-20
fix(openapi): documentation plugin generates OpenAPI with incorrect I…

otociulis:
…D parameter

### What does it do?

Updated type of `id` parameter generated for OpenAPI documentation to match type of `documentId` attribute from `number` to `string`.

### Why is it needed?

`full_documentation.json` file containing OpenAPI documentation contains incorrect `id` parameter definition. Per documentation https://docs.strapi.io/dev-docs/api/rest#update and https://docs.strapi.io/dev-docs/api/rest#delete the ID of resource should be `documentId` (`string` type) but generated documentation defines it as `number`:

```
"parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "",
            "deprecated": false,
            "required": true,
            "schema": {
              "type": "number"
            }
          }
        ],
```

### How to test it?

After creating any collection the generated type of `id` should be `string`:

```
Parameter should be defined as string to match documentId type:

"parameters": [
          {
            "name": "id",
            "in": "path",
            "description": "",
            "deprecated": false,
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
```

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/22599

### DIFF
--- a/packages/plugins/documentation/server/src/services/helpers/build-api-endpoint-path.ts
+++ b/packages/plugins/documentation/server/src/services/helpers/build-api-endpoint-path.ts
@@ -62,7 +62,7 @@ const getPathParams = (routePath: string): OpenAPIV3.ParameterObject[] => {
       description: '',
       deprecated: false,
       required: true,
-      schema: { type: 'number' },
+      schema: { type: param.name === 'id' ? 'string' : 'number' },
     });
 
     return acc;


Bassel17 (approved):
LGTM!
thanks for your contribution

PR #26072, opened 2026-04-20
security: upgrade to axios 1.15.1

innerdvations:
### What does it do?

- upgrade axios to 10.5.1
- upgrades a few other vulnerable pinned subdependencies such as follow-redirect

### Why is it needed?

security issues

### How to test it?

everything should work as before

### Related issue(s)/PR(s)

### DIFF
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -7490,12 +7490,12 @@ __metadata:
   linkType: hard
 
 "follow-redirects@npm:^1.0.0":
-  version: 1.15.1
-  resolution: "follow-redirects@npm:1.15.1"
+  version: 1.16.0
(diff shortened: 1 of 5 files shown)


nclsndr (approved):
Thank you @innerdvations!!! 🙌

unrevised6419:
Question: I saw that `axios` was dropped from most of the packages. Why `axios` is not dropped in favor of `fetch` completely?

innerdvations:
> Question: I saw that `axios` was dropped from most of the packages. Why `axios` is not dropped in favor of `fetch` completely?

If I'm remembering right, it's because native `fetch` was not considered stable until Node 22, and most of the original usage was added before that. At some point we will probably remove axios entirely though.

PR #26080, opened 2026-04-21
feat: support bun

innerdvations:
### What does it do?

Adds official support for running Strapi with bun

- smoke test on the CI to ensure strapi runs with `bun`
- create-strapi-app supports 'bun' (in addition to pnpm, yarn, and npm)
- various typescript fixes (still need to investigate, I don't think these are strictly necessary, it was from when I also tried building the monorepo with bun)

### Why is it needed?

### How to test it?

experimental: `0.0.0-experimental.c6e7562b5dc46fb1f0274087fef904b25ba10bfd`

run `bunx create-strapi-app@0.0.0-experimental.c6e7562b5dc46fb1f0274087fef904b25ba10bfd` and see that it uses bun and Strapi runs properly with `bun run develop`

### Related issue(s)/PR(s)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,6 +60,28 @@ jobs:
       - name: Monorepo build
         uses: ./.github/actions/run-build
 
+  bun_smoke:
+    name: 'bun smoke (${{ github.event_name }})'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


innerdvations:
Will revisit this after some refactoring we want to do first

PR #26086, opened 2026-04-21
fix(data-transfer): buffer push assets before invoking uploadStream

butcherZ:
### What does it do?

Defers the `uploadStream` call on the local-destination assets writable until **after** the incoming PassThrough has been fully drained. Binary chunks from the push handler's WebSocket stream are now accumulated in memory during the stream's `data` events; once `end` fires, the upload provider is invoked with `Readable.from(bufferedChunks)`, a pre-populated synchronous Readable instead of the previous lazy `Readable.from(passThrough)` wrapper.

### Why is it needed?

Pushes to Strapi Cloud consistently crashed mid-transfer with:

```
[ERROR] The first argument must be of type string or an instance of Buffer,
ArrayBuffer, or Array or an Array-like Object. Received undefined
```

The crash was deterministic at ~237ms into the assets stage (position ~48 with small files, mid-first-file with large ones) and never reproduced against local filesystem destinations. Root cause: `Readable.from(passThrough)` produces a lazy async wrapper whose `.read()` returns `null` when no data is yet buffered in the wrapper. The cloud upload provider synchronously calls `Buffer.from()` on that return value, crashing. By buffering chunks client-side and passing a Readable that is already populated when the provider first reads it, the `null` read window is eliminated.

### How to test it?

Manual test against Strapi Cloud `0.0.0-experimental.fff85f01fbad8f3991c577828b60e76c03001f07`

`yarn strapi transfer --to \$REMOTE_URL --to-token \$TOKEN --force`

Expected: progress reaches \`assets: 57 / 57 transferred (size: 2.7 MB / 2.7 MB)\`, finishes with a summary table, and all files land in the Cloud project's media library.

#### Automated coverage

`yarn workspace @strapi/data-transfer test:unit -- assets-destination-writable`

The existing regression test has been updated to reflect the new semantics: \`write()\` callback fires immediately (no deadlock), \`uploadStream\` is invoked after the PassThrough \`end\` event.

### Related issue(s)/PR(s)

Follow-up to #23479 (fix large transfer crashes; show transfer progress).


innerdvations (approved):
@butcherZ LGTM, I added a automated test and also tested it manually against Cloud with 1gb of media data (push and pull) and it worked fine. Please confirm with a manual test on your end before merging :)

Issue #25126, opened 2026-01-06
MorphToOne relation type is broken


boazpoolman:
### Node Version

22.19.0

### NPM/Yarn/PNPM Version

10.9.3

### Strapi Version

5.33.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When I create a 'morphToOne' relation in my schema like so:

```
"related": {
  "type": "relation",
  "relation": "morphToOne"
}
```

And I then try to fill that field with a relation like so:

```
strapi.documents('plugin::todo.task').create({
  data: {
    name: 'test', 
    related: {
      __type: 'api::page.page',
      id: 1,
    },
  },
})
```

I get the following error:

```
/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/database/dist/entity-manager/index.js:121
                        throw new Error(`Expects properties ${typeField} an id to make a morph association`);
                              ^

Error: Expects properties __type an id to make a morph association
    at processData (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/database/dist/entity-manager/index.js:121:31)
    at Object.create (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/database/dist/entity-manager/index.js:184:34)
    at async Object.createEntry [as create] (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/core/dist/services/document-service/entries.js:72:21)
    at async create (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/core/dist/services/document-service/repository.js:106:21)
    at async /Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/database/dist/index.js:85:29
    at async Array.<anonymous> (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/i18n/dist/server/services/ai-localizations.js:242:32)
    at async Array.databaseErrorsMiddleware (/Users/boazpoolman/Code/softwareboard/strapi/my-strapi-project/node_modules/@strapi/core/dist/services/document-service/middlewares/errors.js:16:16)
```

### Steps to Reproduce

See the bug description

### Expected Behavior

I expect this to work. If I change the field to `morphToMany` it starts working.

### Logs

```shell

```

### Confirmation Checklist

[...]

Issue #22659, opened 2025-01-17
Polymorph Relation not working


eauneiladavan:
### Node Version

22.11.00

### NPM/Yarn/PNPM Version

10.9.0

### Strapi Version

5.4.1

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

I have two content types named: action-items and client that have a morph relation.

When making a get request with populate params in action-items, the data was populated correctly, but when doing a get request in client, the actionItems field where set to null. I checked the morph table in postgresql and it has the correct data.

These are the get request that were tried but the responses where not as we expected:
/api/clients?populate=actiontItems => actionItems is null
/api/clients?populate[actionItems][populate]=* => same with item 1
![Image](https://github.com/user-attachments/assets/a1e63b58-20ea-4b9c-a3f9-c4615d222072)

/api/clients?populate[actionItems][populate]=actionType,content,related => giving a 400 response
![Image](https://github.com/user-attachments/assets/ae90b87c-3809-4810-97a2-ea8cafa63d67)

Another thing is when using morphTo as a relation in action-item schema it gives a building error of:
![Image](https://github.com/user-attachments/assets/cf8f110c-4fe3-45aa-82d3-a5f1973fbb2a)

This is the response in action-items, which is working as expected when using populate.
![Image](https://github.com/user-attachments/assets/0f2c4e3d-7a81-4d32-b77a-a8fa48cb8a73)

Hoping that this can be fixed or point me to a documentation to implement it correctly.

### Steps to Reproduce

Steps:
1. Create a content-type with a relation of morphToMany
![Image](https://github.com/user-attachments/assets/a49d2694-22cd-469a-bf22-cbef4b5cb978)

2. Create another content-type with morphMany
![Image](https://github.com/user-attachments/assets/641c4ff7-481f-4dea-a9d7-339bb6eca34a)

3. Add a record in client content type by triggering a POST request.
![Image](https://github.com/user-attachments/assets/6e0a35fc-9f8e-4923-ba4d-c391dbc4ba94)

4. Add a record in action-items content by also triggering a POST request, however with the related data created from the client.![Image](https://github.com/user-attachments/assets/b959e412-116f-4e4f-9044-b659b1a7c840)

### Expected Behavior

Our expected behavior when we initiate the GET request with populate in /client the actionItems will be populate correctly since in our morph table the data was saved accurately.

Morph table:
[...]

PR #26099, opened 2026-04-23
Fix polymorphic relations in the document service API

boazpoolman:
### What does it do?

It includes morph relation types in the `getDeepPopulate` utility.

### Why is it needed?

Solves https://github.com/strapi/strapi/issues/22659 and https://github.com/strapi/strapi/issues/25126

### How to test it?

Run publish, unpublish and discard-draft document service actions for a content type that has both `morphToMany` and `morphToOne` relations.

### DIFF
--- a/packages/core/core/src/services/document-service/transform/__tests__/map-relation.test.ts
+++ b/packages/core/core/src/services/document-service/transform/__tests__/map-relation.test.ts
@@ -1,4 +1,4 @@
-import { mapRelation } from '../relations/utils/map-relation';
+import { mapRelation, traverseEntityRelations } from '../relations/utils/map-relation';
 
 const mapper = mapRelation(async (relation) => {
   if (!relation) return 'default';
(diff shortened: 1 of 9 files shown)


innerdvations (approved):
LGTM!

I added a bunch of tests to confirm the expected behaviour -- and a bunch of missing tests for polymorphic relations in general, to ensure nothing breaks. 
Please look them over and make sure I didn't make any mistakes!

I also discovered that this will carry over a bug that already exists for other relation types, that I've fixed in a follow-up at https://github.com/strapi/strapi/pull/26120

boazpoolman (commented):
Wow thanks for reviewing so quickly @innerdvations!

One thing we might be able to put in this PR is a change to the content history which has it's own version of the [`getDeepPopulate`](https://github.com/strapi/strapi/blob/develop/packages/core/content-manager/server/src/history/services/utils.ts#L213-L217) utility. What do you think? I haven't tested that part of the CMS, but in theory we could apply the change there too. Or better yet, just use the core utility instead of defining it's own?

innerdvations:
For now let's just go with the minimal fix, and then do a follow-up to deduplicate the util to somewhere shared since that will be a lot more complex

boazpoolman:
Sounds good! Then this PR can be merged :)

PR #26101, opened 2026-04-23
fix(core): validate numeric inputs before DB unique checks

### LINKED ISSUE #24678: Strapi crashes when filtering an integer field using a string value instead of returning a validation error

innerdvations:
### What does it do?

Validates and normalizes `number` and `biginteger` inputs at the database field layer and in the entity validators, so malformed numeric values are rejected before they reach SQL.

- **`number` field (`toDB`)** — accepts finite numbers and numeric strings (trimmed); rejects `NaN`/`Infinity`, empty/whitespace strings, and non-numeric types with a clear `Expected a valid Number` error instead of letting bad values reach the database.
- **`biginteger` field (`toDB`)** — accepts only integer-like inputs (`bigint`, integer `number`, or strings matching `/^[+-]?\d+$/`) and normalizes them to a canonical `BigInt` string (trims whitespace, strips leading `+`/zeros). Rejects malformed values like `"900260056-1"` early.
- **`biginteger` field (`fromDB`)** — stays backward compatible: legacy/malformed values already stored in the DB are returned as-is (string) instead of throwing, so existing projects keep working.
- **Entity validators (`validators.ts`)**:
  - `floatValidator` now rejects non-finite numbers (`Infinity`/`NaN`) at the type level.
  - `bigintegerValidator` validates the integer format and normalizes valid values to canonical string form.
  - Unique-constraint checks are skipped for invalid scalar numeric values (`shouldSkipUniqueValidation`) so malformed input fails the type validator instead of reaching the unique-check DB query when `abortEarly=false`.

### Why is it needed?

Invalid numeric strings (e.g. `"900260056-1"`) could previously leak into query/persistence paths and trigger database-level errors (such as PostgreSQL bigint syntax errors) instead of returning a clean validation error. This change blocks malformed numeric input at the Strapi field/validation layer, while still avoiding breaking existing projects that may already contain legacy malformed bigint values in the database.

### How to test it?

Run targeted unit tests:

`yarn workspace @strapi/database test:unit -- numeric-fields.test.ts`
`yarn workspace @strapi/core test:unit -- biginteger-validators.test.ts float-validators.test.ts integer-validators.test.ts`

Regression coverage added for:
- invalid bigint string handling (`"900260056-1"`)
- non-finite float values
- strict `toDB` for `number`/`biginteger`, including bigint normalization
- backward-compatible `fromDB` for legacy malformed bigint rows

### Related issue(s)/PR(s)

Replaces https://github.com/strapi/strapi/pull/24684
Fixes https://github.com/strapi/strapi/issues/24678
Fixes CMS-905


packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -43,6 +43,59 @@ const toNumberSafe = (value: unknown): number | undefined => {
   return Number.isFinite(num) ? num : undefined;
 };
 
+const BIG_INTEGER_REGEX = /^[+-]?\d+$/;
+
+const toBigIntegerString = (value: unknown): string | undefined => {
+  if (_.isNil(value)) {
```
Adzouz: Can't you just use `value == null` instead of lodash?

jhoward1994:
I think we could separate some things out like this ?

```
diff --git a/packages/core/core/src/services/entity-validator/validators.ts b/packages/core/core/src/services/entity-validator/validators.ts
index 08dab9ecba..a9dfae27ad 100644
--- a/packages/core/core/src/services/entity-validator/validators.ts
+++ b/packages/core/core/src/services/entity-validator/validators.ts
@@ -75,24 +75,28 @@ const toBigIntegerString = (value: unknown): string | undefined => {
   return undefined;
 };
 
+const isValidBigInteger = (value: unknown): boolean => !_.isNil(toBigIntegerString(value));
+
+const isValidFiniteNumber = (value: unknown): value is number =>
+  typeof value === 'number' && Number.isFinite(value);
+
+const isValidInteger = (value: unknown): value is number =>
+  isValidFiniteNumber(value) && Number.isInteger(value);
+
 const shouldSkipUniqueValidation = (
   attrType: Schema.Attribute.AnyAttribute['type'],
   value: unknown
 ): boolean => {
   switch (attrType) {
-    case 'integer': {
-      return typeof value !== 'number' || !Number.isFinite(value) || !Number.isInteger(value);
-    }
+    case 'integer':
+      return !isValidInteger(value);
     case 'float':
-    case 'decimal': {
-      return typeof value !== 'number' || !Number.isFinite(value);
-    }
-    case 'biginteger': {
-      return _.isNil(toBigIntegerString(value));
-    }
-    default: {
+    case 'decimal':
+      return !isValidFiniteNumber(value);
+    case 'biginteger':
+      return !isValidBigInteger(value);
+    default:
       return false;
-    }
   }
 };
 
@@ -556,7 +560,7 @@ export const floatValidator = (
     .test(
       'is-finite-number',
       '${path} must be a finite number',
-      (value) => _.isNil(value) || Number.isFinite(value)
+      (value) => _.isNil(value) || isValidFiniteNumber(value)
     );
 
   schema = addMinFloatValidator(schema, metas, options);
@@ -578,7 +582,7 @@ export const bigintegerValidator = (
         return true;
       }
 
-      return !_.isNil(toBigIntegerString(value));
+      return isValidBigInteger(value);
     });
 
   return addUniqueValidator(schema, metas, options);


```

Adzouz (commented):
Hi @innerdvations, thanks a lot for the PR 🙏
It seems good for me, just a tiny thing. Feel free to move forward in you don't think it's necessary but I try to avoid using lodash when I can 😊

packages/core/core/src/services/entity-validator/validators.ts
innerdvations: Switched the new checks to `value == null` but kept lodash since it's still used elsewhere

innerdvations:
> I think we could separate some things out like this ?

You're right, thanks, I extracted them all and simplified

Adzouz (approved):
Looks good, thanks! 🙏

packages/core/core/src/services/entity-validator/validators.ts
```diff
@@ -43,6 +43,63 @@ const toNumberSafe = (value: unknown): number | undefined => {
   return Number.isFinite(num) ? num : undefined;
 };
 
+const BIG_INTEGER_REGEX = /^[+-]?\d+$/;
```
nclsndr, line 46: This is not supporting [numeric separators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators) (e.g: `10_000_000` ). Is that intended @innerdvations?

packages/core/database/src/fields/biginteger.ts
```diff
@@ -1,3 +1,54 @@
-import StringField from './string';
+import { toString } from 'lodash/fp';
+import Field from './field';
 
-export default class BigIntegerField extends StringField {}
+const BIG_INTEGER_REGEX = /^[+-]?\d+$/;
```
nclsndr, line 4: duplicated def

packages/core/core/src/services/entity-validator/validators.ts
innerdvations: I don't see a path where a javascript literal could end up as a string in this case, so I don't think it's a problem. REST+Graphql bodies are JSON (numbers or auto-convert to strings) so someone would have to hand-type that out in a request for it to be a potential problem. And the admin panel already constrains it this way

packages/core/database/src/fields/biginteger.ts
innerdvations: If you're keeping a list of things to clean up when we build the `strapi/definitions` package, add this to the list  😆 I don't think there's currently a "correct" way to dedupe this

PR #26105, opened 2026-04-23
chore(users-permissions): move server code into server/src

unrevised6419:
## Summary

- Move `packages/plugins/users-permissions/server/*` → `server/src/*` so the plugin matches the `admin/src` + `server/src` layout used by every other plugin in `packages/plugins/` (color-picker, i18n, graphql, sentry, cloud, documentation).
- Update `package.json` `exports["./strapi-server"].source` and `rollup.config.mjs` `input` + `rootDir`.
- Fix one relative path in `server/src/services/__tests__/jwt.test.js` that reached `tests/helpers/` at the repo root (now one level deeper).
- Update stale doc references (`docs/.../00-sessions-and-jwt.md`, `packages/plugins/documentation/README.md`).

### Why

Precursor to adding TypeScript support to the `users-permissions` plugin. TypeScript is in turn needed to drop React `propTypes` / `defaultProps` usage in the admin side — both removed from function components in React 19. Normalizing the folder layout first avoids entrenching the odd structure inside new TS tooling configs (tsconfig, rollup, jest).

No runtime or behavior changes. All server files were moved with `git mv` (R100 — identical content). The jest coverage glob `server/**/*.{js,ts}` continues to match recursively.

## Conflict surface with open PRs

Re-swept all **233** open PRs as of **2026-07-08**; **24** touch `packages/plugins/users-permissions/` (this PR + 23 others). My own rename-dependent stack (#26106 TS scaffold, #26107 propTypes removal) already sits on the `server/src/*` layout and rebases together with this branch, so it's excluded from the tables below. Of the remaining 21: **7** touch server code on the old `server/*` paths and will need rebase + path remap to `server/src/*`; **14** are unaffected by the rename (admin / `package.json` / `rollup.config.mjs` / config / translations only).

Changes since the previous (2026-06-25) snapshot:
- **Merged** (dropped): #26752 (06-29), #26715 (06-26), #26628 (06-25), #25144 (06-26), #26687 (06-25), #26765 (06-25), #26508 (06-26).
- **Closed** (dropped): #26110 (06-25), #25368 (07-02).
- **Rebased off `users-permissions`** (still open, no longer touch the plugin): #26764, #26699.
- **New server-bucket**: #26938, #26872, #26820.
- **New rename-unaffected**: #26933, #26901, #26823, #26797.

### Will need rebase (touch `server/*`)

| PR | Last update | Age | Paths | Mergeable |
|----|-------------|-----|-------|-----------|
[...]

PR #26106, opened 2026-04-23
chore(users-permissions): scaffold TypeScript for the plugin

unrevised6419:
Follow-up to #26105 (merged), which moved the server code into `server/src`.

> **Scope note.** An earlier revision of this PR renamed all 152 sources under `admin/src` and `server/src` in one commit. That was too large to review, so it has been cut back to the scaffolding plus the three entry points the new tooling has to resolve. The remaining `.js`/`.jsx` files are converted in follow-ups. No `@ts-nocheck` marker is introduced and no `noImplicitAny: false` escape hatch is set.

## 1. TypeScript scaffolding

- Add `tsconfig.json` + `tsconfig.build.json` for both `admin/` (extends `tsconfig/client.json`) and `server/` (extends `tsconfig/base.json`). Both set `allowJs` so the sources that are still JavaScript stay in the program.
- Add `tsconfig.eslint.json` for both, so typed linting still covers the test files that the two `tsconfig.json` files exclude from the build program
- Add `build:types` (`build:types:server` / `build:types:admin`, run in parallel with `build:code`) and `test:ts:back` / `test:ts:front` scripts; expose `types` in both export entries (`dist/{admin,server}/src/index.d.ts`)
- Add `@strapi/types` as a devDependency, matching `documentation`, `graphql` and `i18n`, plus the `@types/*` packages that back a runtime dependency the sources already import: `bcryptjs` (`server/src/services/user.js`), `jsonwebtoken` (`server/src/services/jwt.js`, `server/src/utils/verify-jwt-with-jwks.js`), `url-join` (four files under `server/src/services`) and `lodash` (eight files under `admin/src`). `@types/node` is the one the server `tsconfig` names in `types`. `@types/koa` is deliberately left out: `koa` is a declared dependency but nothing in the package imports it, and the plugin only reaches Koa through `koa2-ratelimit` and the `koaContext` the GraphQL resolvers hand it.
- Point the rollup `input`s and both `exports` `source` entries at `./{admin,server}/src/index.ts`
- Delete `jsconfig.json`, superseded by the per-target tsconfigs

### ESLint

Previously the plugin's single root config sent `admin/**` to `eslint-config-custom/front` and everything else to `eslint-config-custom/back` — both JavaScript configs. This PR splits that into scoped per-surface configs: `admin/.eslintrc.cjs` (`custom/front/typescript`) and `server/.eslintrc.cjs` (`custom/back/typescript`), with the root config now covering only what neither directory owns. The root config also adds `jest.config.js` and `jest.config.front.js` to `ignorePatterns`.

Each scoped config turns off the rules that the not-yet-converted sources still trip: `import/no-default-export`, `check-file/no-index` and `check-file/filename-naming-convention` on the admin side, `@typescript-eslint/no-var-requires` and `strict` on the server side. They come back as the files convert.

### Fixes surfaced by the new configs

- `server/src/controllers/auth.js`: `return await` inside the `try` block (`@typescript-eslint/return-await`, `in-try-catch`). Previously a rejection from `sendRefreshAuthResponse` bypassed the surrounding `catch`; it is now rethrown as an `ApplicationError` like every other failure in that block.
- `admin/src/components/BoundRoute`: convert to an arrow component (`react/function-component-definition`)
- Admin tests: replace `waitFor(() => getBy…)` with `findBy…`, use `queryByText` in `waitForElementToBeRemoved`, and `toHaveTextContent` instead of asserting on `.textContent` — all `testing-library/*` and `jest-dom/*` rule violations. The two `toHaveTextContent` assertions in `Providers/tests/index.test.jsx` stay inside a shared `waitFor` with a `testing-library/no-wait-for-multiple-assertions` disable each.

## 2. Converted entry points

Three files, the ones the rollup inputs, the `exports` `source` entries and the jest setup have to resolve:

- `admin/src/index.js` → `index.ts`, with the default export typed as `PluginDefinition`
- `admin/src/tests/setup.js` → `setup.ts`, with `jest.config.front.js` repointed at it
- `server/src/index.js` → `index.ts`, CommonJS to ESM

The server entry is the only one with a semantic shift worth calling out: its default export is now the plugin object itself rather than a factory returning it. `loadConfigFile` accepts either — it calls the export when it is a function — and the factory here took no argument, so the two forms are equivalent.

Everything else under `admin/src` and `server/src` is untouched and still JavaScript.

## Test plan

Run in `packages/plugins/users-permissions`:

- [x] `yarn lint`
- [x] `yarn test:ts:back` / `yarn test:ts:front`
- [x] `yarn build` (code + emitted `.d.ts` for both entries)
- [x] `yarn test:unit` — 11 suites / 165 tests passing
- [x] `yarn test:front` — 18 suites / 45 tests passing

And `yarn prettier:check` / `yarn version:check` from the repository root.

## Conflicting open PRs

_Last checked: **2026-09-14**, against all 319 open PRs. Refreshed on each rebase._

Last activity is the most recent of the head branch's last commit, last comment and last review, newest first; the age in parentheses is days before the last-checked date above. GitHub's own *updated* timestamp is not used because it is not a signal of activity here — it moves on labels and cross-references, including the one this table itself creates.

| PR | Last activity | Overlap | Title |
| --- | --- | --- | --- |
| #27302 | 2026-09-11 (3d) | `Roles/pages/tests/CreatePage.test.jsx`, `Roles/pages/tests/EditPage.test.jsx` | chore(admin): migrate react-query to @tanstack/react-query v4 |
| #27454 | 2026-09-01 (13d) | `server/src/controllers/auth.js` | chore: replace lodash isArray/forEach with native and ban it |
| #27492 _(draft)_ | 2026-08-29 (16d) | `jsconfig.json`, deleted here | chore: declare module system explicitly across the monorepo |
| #27458 | 2026-08-25 (20d) | `components/BoundRoute/index.jsx`, `server/src/controllers/auth.js` | chore: replace lodash map and reduce with native equivalents and ban them |
| #27457 | 2026-08-25 (20d) | `server/src/controllers/auth.js` | chore: replace lodash keys/values/entries/concat/filter/find with native and ban them |
| #27181 _(draft)_ | 2026-08-10 (35d) | `admin/src/index.js`, renamed here | chore(tooling): align admin defaultMessage with en.json |
| #22674 | 2026-07-10 (66d) | `admin/src/tests/setup.js`, renamed here | feat: add email confirmation error redirection settings |
| #26107 _(draft)_ | 2026-04-23 (144d) | stacked on the old full-migration branch | chore: remove PropTypes and defaultProps for React 19 compatibility |

Only the last three need their hunks replayed onto a new path; the rest are ordinary content conflicts.

Fourteen further PRs (#26194, #26541, #26631, #26776, #27119, #27249, #27348, #27407, #27494, #27550, #27569, #27570, #27571, #27586) overlap only on this package's `package.json` dependency block and are left out. `yarn.lock` is touched by most open PRs and always needs regenerating.

## Related

The `@typescript-eslint/naming-convention` cleanup that an earlier revision of this PR carried — removing the `naming-convention: 'off'` overrides in `core/{admin,content-manager,review-workflows,upload}/shared/.eslintrc.cjs` and the file-level `eslint-disable` in `core/database/src/entity-manager/morph-relations.ts` — has moved to #27491 and is no longer part of this branch. The `core/utils/src/template.ts` signature fix an earlier revision also carried is no longer needed at this scope and has been dropped.


packages/plugins/users-permissions/admin/.eslintrc
```diff
@@ -0,0 +1,19 @@
...
+    "testing-library/prefer-find-by": "warn",
+    "testing-library/prefer-query-by-disappearance": "warn",
+    "testing-library/no-wait-for-multiple-assertions": "warn",
+    "jest-dom/prefer-to-have-text-content": "warn",
+
+    "import/no-default-export": "warn",
+    "import/default": "warn",
+
+    "check-file/no-index": "off",
+    "check-file/filename-naming-convention": "off"
```
unrevised6419: Question: Some of these rules ESLint can auto fix. Should we auto fix them in this PR?

packages/plugins/users-permissions/server/.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
+  "root": true,
+  "extends": ["custom/back/typescript"],
+  "parserOptions": {
+    "project": ["./server/tsconfig.json"]
+  },
+  "rules": {
+    "strict": "warn",
+    "@typescript-eslint/no-var-requires": "warn"
```
unrevised6419: Question: Some of these rules ESLint can auto fix. Should we auto fix them in this PR?

packages/plugins/users-permissions/server/.eslintrc.cjs
```diff
@@ -0,0 +1,19 @@
...
+  root: true,
+  extends: ['eslint-config-custom/back/typescript'],
+  ignorePatterns: ['.eslintrc.cjs'],
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: ['./tsconfig.json'],
+  },
+
+  rules: {
+    strict: 'off',
```
unrevised6419, line 14: Note: The removal of `use strict` can be fixed by eslint

packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/CreatePage.test.jsx
unrevised6419, line 1: Note: Changes in this file were recommended by ESLint

packages/plugins/users-permissions/admin/src/pages/Roles/pages/tests/EditPage.test.jsx
unrevised6419, line 1: Note: Changes in this file were recommended by ESLint

packages/plugins/users-permissions/admin/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "tsconfig/client.json",
+  "include": ["./src"],
+  "compilerOptions": {
+    "types": ["@testing-library/jest-dom"],
+
+    // TODO: Remove this when all files are migrated to TypeScript
+    "allowJs": true,
+    "checkJs": false
```
unrevised6419, line 9: Note: In a subsequent PR, this can be turned to true, and fix all errors. There are not so many. Many of them are just wrong assumption in JSDocs

packages/plugins/users-permissions/server/.eslintrc.cjs
```diff
@@ -0,0 +1,19 @@
...
+    '@typescript-eslint/no-var-requires': 'off',
```
unrevised6419, line 15: Note: We will get rid of require, when files gets converted to TS

packages/plugins/users-permissions/server/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "tsconfig/base.json",
+  "include": ["./src"],
+  "compilerOptions": {
+    "types": ["node", "jest"],
+
+    // TODO: Remove this when all files are migrated to TypeScript
+    "allowJs": true,
+    "checkJs": false
```
unrevised6419: Note: In a subsequent PR, this can be turned to false, and fix all errors. There are not so many. Many of them are just wrong assumption in JSDocs

packages/plugins/users-permissions/admin/.eslintrc.cjs
```diff
@@ -0,0 +1,17 @@
...
+const config = {
+  root: true,
+  extends: ['eslint-config-custom/front/typescript'],
+  ignorePatterns: ['.eslintrc.cjs'],
+
+  rules: {
+    'import/no-default-export': 'off',
+    'import/default': 'off',
+    'check-file/no-index': 'off',
+    'check-file/filename-naming-convention': 'off',
```
unrevised6419, line 13: Note: Disabled on purpose, our lint script expects 0 warnings.

unrevised6419:
Hey @innerdvations, I would like to look into migrating `users-permissions` from `id` to `documentId` with back compat, but I would need this one merged first to make use of TypeScript typings.

PR #26112, opened 2026-04-24
fix: preserve order when reordering a relation to the start of a list

### LINKED ISSUE #25958: Strapi V5: Reordering Element from bottom to top of list
### Node Version

20.20.1

### Package Manager

npm

### Package Manager Version

10.8.2

### Strapi Version

5.4.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

Evidences

### Steps to Reproduce

**Steps to recreate:** 

1. Navigate to a page with enough children elements associated to a field to render the “Load More” button
2. Add a new child element to a Collection Type. 
3. Click Save
4. Re-order to the first position. 
5. Click Save or Publish. 
6. When you click “Load More” again the entry will be in the second position.

### Expected Behavior

**Current Behavior**

When attempting to re-order elements of topics within Strapi, it does not stay in place once saved.

**Expected Behavior**

When attempting to re-order elements of topics within Strapi, it does stay in place once saved.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

https://github.com/user-attachments/assets/6ebe2fac-1bf1-457d-b3e5-84a40c019789

### Additional information

_No response_

### Confirmation Checklist

akash-dabhi-qed:
### What does it do?

* Adds a `.min() `method to `QueryBuilder` (symmetric with the existing `.max()`) so entity-manager can subquery for the lowest-order row in a join table.
* Updates `updateRelations` in `entity-manager/index.ts` to fetch both the min-order and max-order boundary rows as `adjacentRelations`, not just the max. This gives the orderer the anchor it needs when inserting at the start of a list.
* Fixes three bugs in `relations-orderer.ts`:
  * `order: 0` was treated as falsy and defaulted to 1 — fixed with an explicit null/undefined check.
  * `before`/`after` placement used same-order assignment instead of midpoint interpolation, causing order collisions — fixed with`(prevOrder + targetOrder) / 2`/ `(targetOrder + nextOrder) / 2`.
  * `{ start: true }` now assigns `firstRelation.order - 0.5` when the list is non-empty, placing the item strictly before the current minimum.
* Updates `Relations.tsx` to send `{ start: true }` instead of `{ before: firstItemDocId }` when a relation is dragged to index 0.

https://github.com/user-attachments/assets/3eafee2e-b759-4fda-a66a-88e4ebee3f26

### Why is it needed?

When a user dragged a relation to position 1 (the top of the list) and then saved/published, the ordering was not persisted correctly. After clicking Load More, the item would appear at position 2 instead of position 1. The root cause was twofold: the frontend was sending `{ before: firstItemDocId }` which only works if that item is already in the orderer's `initArr`, and the orderer had no knowledge of the list's minimum-order boundary to anchor a `{ start: true }` placement against.

### How to test it?

### Related issue(s)/PR(s)

Let us know if this is related to any issue/pull request<br>Fixes strapi/strapi#25958


mathildeleg:
Hey, thanks for the PR! A few things that need addressing:

1. 🔴 Regression: before placement breaks for null-order relations
The new midpoint formula (`prevOrder + targetOrder`) / 2 degenerates when both adjacent items have the same order. This happens with null-order relations — they all normalize to 1, so the midpoint of (1 + 1) / 2 = 1, which is the same order as the target. `getOrderMap` then assigns the new item 1.1225 while the init target stays at 1 in the DB, making the item land after the relation it was supposed to go before.

2. 🔴 4 existing unit tests fail on this branch
Running yarn workspace @strapi/database test:unit:
  - 3 stale expectations — tests `check orderer.get()` internal values that changed with the new algorithm (the logical ordering is still correct, the expected numbers just need updating):
    - When I connect a relation at the start — expects 0.5, receives 3.5
    - When I connect multiple relations using before — expects 9.5, receives 7
    - When you connect multiple disordered relations — expects 1.5, receives 1.25
- 1 confirms the regression above _When you connect a relation before one with null order_ expects id:4 at 0.5, receives 1.0

3. 🟡 `Relations.tsx` is not in the changed files
The PR description says: "Updates `Relations.tsx` to send `{ start: true } `instead of `{ before: firstItemDocId }` when a relation is dragged to index 0." but `Relations.tsx` isn't in the diff.

4. No tests for the three orderer fixes
`relations-orderer.ts` already has a test suite. The three bug fixes (falsy order: 0, before/after midpoint, start boundary) each deserves a regression test — otherwise a future refactor could silently re-introduce them.

Let me know if you'd like to handle all of that, otherwise we'll close the PR and either someone from the community tries another fix for the issue or we'll handle it ourselves (depending on who gets to it first). Thanks!

akash-dabhi-qed:
Thanks for the detailed review, addressed all four points.

**`before`/`after` null-order regression**
The midpoint formula only runs now when the adjacent item actually leaves room (`prev.order < target.order`). When they share an order, e.g. null orders that all normalise to 1, it falls back to `target.order - 0.5` (and symmetrically `+ 0.5` for `after`), so the new relation lands strictly before/after its target instead of colliding with it. The null-order test now lands id 4 at 0.5 as expected.

**Failing unit tests**
- The null-order test passes unchanged with the fix above (it was the real regression).
- Updated the 3 stale expectations to the new algorithm's values, logical ordering is unchanged, only the fractional numbers moved:
  - connect at start → `3.5` (anchored below the current minimum instead of a hard-coded `0.5`)
  - multiple `before` → `7`/`7` (interpolated into the gap; `getOrderMap()` splits them into distinct fractions)
  - disordered connect → `1.25`

**Regression tests**
Added coverage for the three fixes:
- `order: 0` is preserved (not coerced to 1) and `start` anchors below it.
- Reordering to the top when the minimum order is already fractional, asserted on `getOrderMap()`, since that's the value actually written to the DB. This is the core #25958 case.
- `after` with an equal-order neighbour falls back to strictly after the target.

**`Relations.tsx`**
Now included in the diff. When a relation is dragged to index 0 it sends `{ start: true }` instead of `{ before: firstItemDocId }`, `before` only resolves when that item is already in the orderer's initial set, so dragging to the top of a partially loaded list was landing at position 2. Added `start` to the local `RelationPosition` type (the backend and `@strapi/types` already support it).

mathildeleg:
Hey there, thanks for taking care of the feedback! Do you mind fixing the conflict in the `relations-orderer.ts` file before I start reviewing again? 🙏🏻 Thank you!

mathildeleg (approved):
LGTM, thanks for addressing the feedback ✨

PR #26119, opened 2026-04-24
fix(content-manager): guard against undefined component schema in formatEditLayout

### LINKED ISSUE #25790: "No permissions to see this field" appears randomly on some fields of type component only after adding components via pre-existing single type relation
### Node Version

24.11.1

### Package Manager

npm

### Package Manager Version

11.6.2

### Strapi Version

5.34.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

A field's access is randomly restricted and shows "No permissions to see this field". This happened only after the after a field and its component were added via the single content type. The user is "Super Admin"

### Steps to Reproduce

1. Create content type in Strapi with fields
2. Add new component via "add another field to this type single type" button
3. Add fields to the new component
4. Go to the content manager and the field with the new component is restricted.

### Expected Behavior

Adding fields as component by creating new components via the pre-existing single type creates the new component and does not change the access permissions.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

VibhuGupta-dev:
## What does it do?
Adds a null-check guard in `formatEditLayout` inside `useDocumentLayout.ts` to skip components that are not yet present in the client-side schema cache.
## Why is it needed?
When a new component is created inline through an existing single type in the Content-Type Builder, the component schema is not immediately available in the client-side `components` cache dictionary.

`formatEditLayout` was iterating over `data.components` (from the API response) and directly accessing `components[uid].attributes` without checking if the schema existed in cache, causing a fatal crash:

TypeError: Cannot read properties of undefined (reading 'attributes')

This crash shows the "Something went wrong" error screen in the Content Manager, making the single type inaccessible until Strapi is restarted.
## How to test it?
1. Open Content-Type Builder
2. Create a new Single Type (e.g. HomePage)
3. Add a field, then click "Add another field to this type"
4. Create a new component inline (not from existing components)
5. Add fields to the component and save
6. Navigate to Content Manager → the Single Type
7. Previously: crash with "Something went wrong" error
8. Now: page loads correctly with the component field visible

Environment:
- Strapi v5
- SQLite
- TypeScript
## Related issue(s)/PR(s)
Fix #25790

### DIFF
--- a/packages/core/content-manager/admin/src/hooks/useDocumentLayout.ts
+++ b/packages/core/content-manager/admin/src/hooks/useDocumentLayout.ts
@@ -285,6 +285,10 @@ const formatEditLayout = (
 
   const componentEditAttributes = Object.entries(data.components).reduce<EditLayout['components']>(
     (acc, [uid, configuration]) => {
+      if (!components[uid]) {
+        return acc;
+      }
+
       acc[uid] = {
         layout: convertEditLayoutToFieldLayouts(
           configuration.layouts.edit,


innerdvations:
Thanks @VibhuGupta-dev — the guard you added here is exactly the right idea for the crash you described.

This is already fixed on `develop` in #26243 (with follow-up #26625), which includes the same missing-component-schema skip plus schema refresh after CTB updates, additional layout guards, and regression tests. Closing as superseded by that work.

We appreciate you taking the time to report and fix this.

_AI-assisted comment._

PR #26120, opened 2026-04-24
fix(database): morph typeField wins over same-name fields in populate

innerdvations:
NOTE: based on https://github.com/strapi/strapi/pull/26099 -- do not review until that one is merged

### What does it do?

- Updates `morphToOne` and `morphToMany` population in `packages/core/database/src/query/helpers/populate/apply.ts` so the merged object is built with the **target row spread first** and the synthetic **`typeField`** (default `__type`) set **last**, e.g. `{ ...fromTargetRow(row), [typeField]: type }` (and the `morphToOne` equivalent), instead of placing the type field first so a same-named key on the target could overwrite it.
- Adds a unit test in `packages/core/database/src/query/helpers/populate/__tests__/apply-morph-to-one.test.ts` that covers a `fromRow` target shape including a conflicting `__type` value and asserts the result still contains the real morph **target content-type UID** for that key.

### Why is it needed?

- With the previous object merge order, a field on the **morph target** that serialized under the same key as `typeField` (commonly `__type`) could **win over** the polymorphic discriminator in the populated relation object, so the response no longer reflected the true target type UID. The populated morph item must always expose a single, unambiguous `typeField` for the **target** content type; merging target data first, then the discriminator, enforces that.

### How to test it?

- Run the database package unit tests for this file, for example from the repo:
  - `cd packages/core/database && yarn test:unit --testPathPattern=apply-morph-to-one`
- Optional: run the full unit test suite for your branch if you want extra safety (`yarn test:unit` at the root, per `AGENTS.md`).

No special app setup or database (Postgres / MySQL) is required for the new unit test; it uses the existing Jest mocks in that file.

### Related issue(s)/PR(s)

- Small hardening on top of broader morph / polymorphic population work; link any open issue you are closing, or a parent PR (e.g. [PR #26099](https://github.com/strapi/strapi/pull/26099)) if this is reviewed or shipped as a follow-up there.

Fixes CMS-914


boazpoolman (dismissed):
LGTM!

boazpoolman (dismissed):

Bassel17:
Non-blocking: the implementation also changes `morphToMany` with the same merge-order fix, but the regression test only covers `morphToOne`. Could we add a matching `morphToMany` test for the same `__type` collision case, just to lock both paths down?

otherwise LGTM!

Bassel17 (approved):
LGTM!

PR #26122, opened 2026-04-24
fix(database): avoid double finalising completed transactions

innerdvations:
### What does it do?

- Skips `commit` and `rollback` in `transactionCtx` when the Knex transactor already reports `isCompleted()` (e.g. after the callback `await`s `rollback()` and returns), so the success path does not call `commit` on a finished transactor.
- Clears the async-local `trx` when bailing out so the store does not keep a stale reference.
- Adds JSDoc on `Database#transaction` describing this behaviour.
- Adds unit tests in `@strapi/database` for the “already finalised” transactor case, and two API integration cases in `tests/api/core/database/transactions.test.api.js` (plain `rollback` then return, and `queryBuilder` + `transacting(trx)` + `rollback` then return, matching the `testInTransaction` pattern).

### Why is it needed?

`Database#transaction` always ran `commit` after a fulfilled callback. If the callback had already finalised the work with `await rollback()` and then returned, Knex could error or misbehave on a second `commit` (“Transaction query already complete” / bad connection state). This pattern is used by the API test helper `wrapInTransaction` / `testInTransaction` in `tests/api/utils/index.ts`.

### How to test it?

- `cd packages/core/database && yarn test:unit src/__tests__/index.test.ts --forceExit`
- `yarn test:api -- --testPathPattern=tests/api/core/database/transactions.test --forceExit` (or run the full API suite as in CI)


Bassel17 (approved):
lgtm

Issue #25509, opened 2026-02-19
Renaming component label in “Configure the view” breaks Content Manager with Cannot read properties of undefined (reading 'attributes')


alex-strapi:
### Node Version

22

### NPM/Yarn/PNPM Version

Yarn 4.5.0

### Strapi Version

5.36.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

https://github.com/strapi/LaunchPad

### Bug Description

Renaming a component field label in Content Manager → Configure the view causes the Content Manager list view to crash if the component already contains existing content.

After renaming the component label (e.g. from user to User) and saving, navigating back to the Content Manager list view for that content type throws the following error:

`TypeError: Cannot read properties of undefined (reading 'attributes')`

The error originates from getMainField inside the admin panel bundle and appears to be related to resolving component attributes when computing the main field for the list view.

This only occurs when:

- The component already has existing content.
- The rename is performed via Configure the view (label change only, not schema change).

The Content Manager becomes unusable for that content type until reverted.

The issue can be reversed by:

1. Renaming the component label back to its original value.
2. Restarting Strapi.

Without restarting Strapi, reverting the label does not immediately fix the issue.

### Steps to Reproduce

I could replicate this issue consistently in our LaunchPad demo, but in a blank strapi you would:

1. Create a content type (e.g. Testimonials).
2. Add a component field (e.g. user) with fields such as:
- firstname
- lastname
- job
- image
3. Create at least one entry so that content exists in the component.
4. Go to Content Manager → Testimonials → Configure the view.
5. Edit the component field label (e.g. change user to User).
6. Click Finish, then Save.
7. Navigate back to the Content Manager list view for Testimonials.

### Expected Behavior

Renaming a component label in Configure the view should only update the UI label and should not:

- Break list view rendering
- Affect schema resolution
- Require a server restart
- Cause attribute resolution to fail

The Content Manager list view should continue to work normally after saving the updated layout.

### Logs

```shell
TypeError: Cannot read properties of undefined (reading 'attributes')
    at getMainField
    at convertListLayoutToFieldLayouts
    at ListViewPage
```

### Media

<img width="1376" height="914" alt="Image" src="https://github.com/user-attachments/assets/30beccb3-4784-497f-aaf8-7fc0cf8fbc9a" />

[...]

PR #26124, opened 2026-04-24
fix(content-manager): fix getMainField context for component list/edit configure views (#25509)

innerdvations:
**What does it do?**

- Introduces `formatComponentConfigurationEditLayout` (extracted from the component “Configure the view” flow) so every `convertEditLayoutToFieldLayouts` call—including for each entry in `data.components`—receives `{ configurations: data.components, schemas: components }`, matching the content-type `formatEditLayout` path in `useDocumentLayout`. That lets `getMainField` resolve `mainField` for nested component fields (e.g. component-in-component) when edit metadata includes `mainField`.
- Updates list configuration (`ListConfigurationPage`) to call `convertListLayoutToFieldLayouts` with the same list view conversion context as `ListViewPage` (component configurations, component schemas, content-type schemas) when `listViewConversionContext` is available, so the list “Configure the view” / initial form values do not hit the same missing-schemas problem as the list view.
- Adds unit tests in `formatComponentConfigurationEditLayout.test.ts` for the nested component + persisted `mainField` case.

**Why is it needed?**

- [Issue #25509](https://github.com/strapi/strapi/issues/25509): certain configure-the-view flows could cause `TypeError: Cannot read properties of undefined (reading 'attributes')` in `getMainField` when conversion ran without full component schema context. [PR #25872](https://github.com/strapi/strapi/pull/25872) fixed the list view path; this addresses the same class of bug for component configuration (nested `convertEditLayoutToFieldLayouts`) and aligns list configuration with the list view / `formatListLayout` behavior. See [this comment](https://github.com/strapi/strapi/issues/25509#issuecomment-4222804403) for the nested component configure case.

**How to test it?**

- **Automated:** `yarn test:front` (at least `formatComponentConfigurationEditLayout.test.ts`, `ListConfigurationPage.test.tsx`, and existing `convertListLayoutToFieldLayouts` / `useDocumentLayout` tests).
- **Manual:** Create a parent component whose sub-component includes another component field; save configure-the-view; reopen configure-the-view—no crash. For collection types, open “Configure the list view” for a type with component/relation list columns and confirm the screen loads and labels behave as before.

**Related issue(s)/PR(s)**

- [Issue #25509](https://github.com/strapi/strapi/issues/25509)
- [PR #25872](https://github.com/strapi/strapi/pull/25872) (related prior fix for list view)

Fixes CMS-915


packages/core/content-manager/admin/src/pages/ListConfiguration/ListConfigurationPage.tsx
```diff
@@ -131,17 +131,32 @@ const ListConfiguration = () => {
       return acc;
     }, {});
 
+    /**
+     * Same context as `formatListLayout` / `ListViewPage#displayedHeaders` so `getMainField`
+     * can resolve component and relation list columns (see #25509, #25872).
+     */
+    const listFieldLayouts = listViewConversionContext
+      ? convertListLayoutToFieldLayouts(
```
mathildeleg, line 139: I think you may need to modify the L365 and L472 of the `useDocumentLayout.tsx` for this const `convertListLayoutToFieldLayouts` and the other `convertEditLayoutToFieldLayouts` to:
```
const settings: Partial<Settings> =
          attribute.type === 'component' && components
            ? (components.configurations[attribute.component]?.settings ?? {})
            : {};
```
Otherwise, if that UID is not in configurations, reading .`settings` throws.

mathildeleg (commented):
One small thing but otherwise LGTM :)

packages/core/content-manager/admin/src/pages/ListConfiguration/ListConfigurationPage.tsx
innerdvations: done!

mathildeleg (dismissed):

innerdvations:
Do not merge this until after [#26215](https://github.com/strapi/strapi/pull/26215) is merged back into `develop`.

Expected conflict resolution notes:

- `packages/core/content-manager/admin/src/hooks/useDocumentLayout.ts`
  - Both PRs touch this file.
  - Keep `#26215` model-switch/loading-race handling and missing-schema/metadata guards.
  - Also keep this PR’s defensive component settings fallback (`?.settings ?? {}`) anywhere still applicable.

- `packages/core/content-manager/admin/src/pages/ComponentConfigurationPage.tsx`
  - This is the likely manual conflict.
  - Keep this PR’s extracted `formatComponentConfigurationEditLayout` flow (with nested conversion context `{ configurations, schemas }`).
  - Fold in `#26215`’s missing component schema guard (`if (!componentSchema) return acc;`) in the component reducer path.

After conflict resolution, run targeted CM tests (`useDocumentLayout` + component/list configure layout conversion tests).

jhoward1994 (approved):

PR #26134, opened 2026-04-27
fix(database): run cleanOrderColumns updates sequentially

### LINKED ISSUE #26131: `cleanOrderColumns` deadlock on `admin_users_roles_lnk` when ordering columns have gaps
### Node Version

22.17.0

### Package Manager

npm

### Package Manager Version

10.9.2

### Strapi Version

5.41.1

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When a user is removed from a role, the ordering column (`user_ord` or `role_ord`) in the `admin_users_roles_lnk` join table can be left with a non-dense sequence, i.e. a gap.

On the next write that touches that relation, `cleanOrderColumns` in `@strapi/database/src/entity-manager/regular-relations.ts` attempts to renumber the column by issuing two concurrent `UPDATE` statements via `Promise.all`:

    return Promise.all([updateOrderColumn(), updateInverseOrderColumn()]);

`updateOrderColumn` renumbers `role_ord`, partitioned by `user_id`.

`updateInverseOrderColumn` renumbers `user_ord`, partitioned by `role_id`.

Both target the same join table, with potentially overlapping row sets.

Under concurrent requests, this creates a classic lock cycle:

- Transaction A locks rows for `user_id = X` while renumbering `role_ord`, then waits for rows where `role_id = Y`.
- Transaction B locks rows for `role_id = Y` while renumbering `user_ord`, then waits for rows where `user_id = X`.

PostgreSQL detects the cycle and aborts one transaction with `SQLSTATE 40P01`.

Strapi's error handling does not surface the deadlock error cleanly, causing the failed request to hang and, in some environments such as Azure App Service with a short container timeout, to appear as a boot loop.

`cleanOrderColumns` is called from `deleteRelations`, `deletePreviousOneToAnyRelations`, and `deletePreviousAnyToOneRelations`, so any concurrent admin user-role update can trigger this.

### Steps to Reproduce

1. Have at least two admin users, each assigned to the same role
2. Remove one user from the role, leaving a gap in user_ord in admin_users_roles_lnk
3. Send two concurrent PUT /admin/users/:id requests that each modify roles for different users who share a role
4. Observe that one request returns 500 and Postgres logs show ERROR: deadlock detected

### Expected Behavior

`cleanOrderColumns` should complete without error regardless of concurrent writes.

The two renumbering operations should either execute sequentially or be combined into a single atomic `UPDATE` using a CTE, so they do not compete for row locks on the same join table.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

Related issues:

- https://github.com/strapi/strapi/issues/22828
- https://github.com/strapi/strapi/issues/14964
- https://github.com/strapi/strapi/issues/15202
- https://github.com/strapi/strapi/issues/14937

### Confirmation Checklist

innerdvations:
### What does it do?

- In `cleanOrderColumns` (`packages/core/database/src/entity-manager/regular-relations.ts`), runs the order and inverse-order renumbering `UPDATE`s **sequentially** instead of with `Promise.all`, so they do not execute in parallel on the same join table.
- Adds a regression test under **`tests/api/core/admin/admin-user.test.api.js`**: nested `describe('Concurrent admin user role updates (regression #26131)', …)` that creates two admin users sharing a role, introduces a gap in the join-table order columns, and runs concurrent `PUT /admin/users/:id` role updates in a loop.

### Why is it needed?

When ordering columns on a many-to-many join table are not dense, renumbering can run two updates partitioned by different sides of the relation. Running them concurrently can interleave row locks and cause **PostgreSQL deadlocks** (see #26131).

### How to test it?

- `yarn test:api tests/api/core/admin/admin-user.test.api.js`
- Optional: `yarn test:api --db=postgres tests/api/core/admin/admin-user.test.api.js` (if you use the API test DB flag for Postgres).

### Related issue(s)/PR(s)

Fix #26131

Fix CMS-913

### DIFF
--- a/packages/core/database/src/entity-manager/regular-relations.ts
+++ b/packages/core/database/src/entity-manager/regular-relations.ts
@@ -376,7 +376,9 @@ const cleanOrderColumns = async ({
     }
   };
 
-  return Promise.all([updateOrderColumn(), updateInverseOrderColumn()]);
+  // Run updates in a deterministic order to avoid lock cycles on the same join table.
+  await updateOrderColumn();
+  await updateInverseOrderColumn();
 };
 
 export {
--- a/tests/api/core/admin/admin-user.test.api.js
+++ b/tests/api/core/admin/admin-user.test.api.js
@@ -29,6 +29,7 @@ const omitRegistrationToken = omit(['registrationToken']);
  * 14. Updates a user (not found)
  * 15. Finds a user (not found)
  * 16. Finds a list of users (missing user)
+ * —   Concurrent role updates (join order gaps, regression #26131)
  */
 
 describe('Admin User CRUD (api)', () => {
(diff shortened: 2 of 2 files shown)


jhoward1994:
Looks good. Would it be worth improving our CI runtime by making the test run only on postgres? Something like this diff 


```
diff --git a/tests/api/core/admin/admin-user.test.api.js b/tests/api/core/admin/admin-user.test.api.js
index d43d6710cd..950c051166 100644
--- a/tests/api/core/admin/admin-user.test.api.js
+++ b/tests/api/core/admin/admin-user.test.api.js
@@ -29,7 +29,7 @@ const omitRegistrationToken = omit(['registrationToken']);
  * 14. Updates a user (not found)
  * 15. Finds a user (not found)
  * 16. Finds a list of users (missing user)
- * —   Concurrent role updates (join order gaps, regression #26131)
+ * —   Concurrent role updates (join order gaps, regression #26131) — Postgres only
  */
 
 describe('Admin User CRUD (api)', () => {
@@ -398,7 +398,12 @@ describe('Admin User CRUD (api)', () => {
     expect(res.body.data.results).toHaveLength(0);
   });
 
+  // Postgres-only: the deadlock this guards against is specific to Postgres row-locking
+  // semantics on shared join tables (see #26131). SQLite serialises writes at the database
+  // level and cannot reproduce the cycle, so running this elsewhere would be a false positive.
   describe('Concurrent admin user role updates (regression #26131)', () => {
+    const CONCURRENT_UPDATE_ITERATIONS = 3;
+
     const concurrencyData = {
       users: [],
       roles: [],
@@ -406,6 +411,9 @@ describe('Admin User CRUD (api)', () => {
       keepRole: undefined,
     };
 
+    // Resolved in beforeAll once `strapi` is initialised; guards setup/cleanup/test body.
+    let isPostgresRun = false;
+
     const createAdminUser = async ({ email, roles }) => {
       const res = await rq({
         url: '/admin/users',
@@ -434,9 +442,10 @@ describe('Admin User CRUD (api)', () => {
       return strapi.db.metadata.get('admin::user').attributes.roles.joinTable;
     };
 
+    // Re-introduce a gap in the inverse order column for the shared role so the next
+    // call to cleanOrderColumns actually has work to do on both partitions.
     const forceGapInSharedRoleOrder = async () => {
-      const joinTable = getUserRolesJoinTable();
-      const { name, inverseJoinColumn, inverseOrderColumnName } = joinTable;
+      const { name, inverseJoinColumn, inverseOrderColumnName } = getUserRolesJoinTable();
 
       const rows = await strapi.db
         .getConnection()
@@ -445,31 +454,21 @@ describe('Admin User CRUD (api)', () => {
         .where(inverseJoinColumn.name, concurrencyData.sharedRole.id)
         .orderBy(inverseOrderColumnName, 'asc');
 
-      expect(rows).toHaveLength(2);
+      if (rows.length < 2) return;
 
       const firstOrder = Number(rows[0][inverseOrderColumnName]);
-      const secondOrder = Number(rows[1][inverseOrderColumnName]);
-
-      if (secondOrder !== firstOrder + 2) {
-        await strapi.db
-          .getConnection()
-          .from(name)
-          .where('id', rows[1].id)
-          .update({ [inverseOrderColumnName]: firstOrder + 2 });
-      }
 
-      const updatedRows = await strapi.db
+      await strapi.db
         .getConnection()
         .from(name)
-        .select([inverseOrderColumnName])
-        .where(inverseJoinColumn.name, concurrencyData.sharedRole.id)
-        .orderBy(inverseOrderColumnName, 'asc');
-
-      const updatedOrders = updatedRows.map((row) => Number(row[inverseOrderColumnName]));
-      expect(updatedOrders).toEqual([firstOrder, firstOrder + 2]);
+        .where('id', rows[1].id)
+        .update({ [inverseOrderColumnName]: firstOrder + 2 });
     };
 
     beforeAll(async () => {
+      isPostgresRun = strapi?.db?.dialect?.client === 'postgres';
+      if (!isPostgresRun) return;
+
       const timestamp = Date.now();
       concurrencyData.sharedRole = await utils.createRole({
         name: `concurrency-shared-role-${timestamp}`,
@@ -494,6 +493,8 @@ describe('Admin User CRUD (api)', () => {
     });
 
     afterAll(async () => {
+      if (!isPostgresRun) return;
+
       if (concurrencyData.users.length > 0) {
         await utils.deleteUsersById(concurrencyData.users.map((user) => user.id));
       }
@@ -504,19 +505,20 @@ describe('Admin User CRUD (api)', () => {
     });
 
     test('concurrent role removals succeed when join order has gaps', async () => {
-      const [userA, userB] = concurrencyData.users;
+      if (!isPostgresRun) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          '[skipped] regression #26131 only reproduces on Postgres; current dialect is ' +
+            `"${strapi?.db?.dialect?.client}"`
+        );
+        return;
+      }
 
-      const removeSharedRes = await updateAdminUserRoles(userB.id, [concurrencyData.keepRole.id]);
-      expect(removeSharedRes.statusCode).toBe(200);
+      const [userA, userB] = concurrencyData.users;
 
-      const addSharedBackRes = await updateAdminUserRoles(userB.id, [
-        concurrencyData.sharedRole.id,
-        concurrencyData.keepRole.id,
-      ]);
-      expect(addSharedBackRes.statusCode).toBe(200);
       await forceGapInSharedRoleOrder();
 
-      for (let attempt = 0; attempt < 8; attempt += 1) {
+      for (let attempt = 0; attempt < CONCURRENT_UPDATE_ITERATIONS; attempt += 1) {
         const [removeARes, removeBRes] = await Promise.all([
           updateAdminUserRoles(userA.id, [concurrencyData.keepRole.id]),
           updateAdminUserRoles(userB.id, [concurrencyData.keepRole.id]),


```

jhoward1994 (dismissed):
See above comment

innerdvations:
> Looks good. Would it be worth improving our CI runtime by making the test run only on postgres? Something like this diff

I'd rather keep it for all databases, even though it's apparently only currently an issue on postgres in theory any database could do an update and start seeing the same thing.

Bassel17 (approved):
LGTM! good catch

PR #26146, opened 2026-04-28
security: upgrade tar to v7 in v4

Bassel17:
### What does it do?

Ports the tar security update work from PR #25380 to v4.

### Why is it needed?

This addresses security warnings related to the older tar v6 dependency

### How to test it?

run some imports and exports and ensure that the files are created and imported properly


innerdvations (approved):
I didn't test it but we can test all the v4 patches as a whole prior to release

PR #26147, opened 2026-04-28
fix: FK violation publishing self-relation parent & child in one release

garrettheaver:
### What does it do?
Replaces `Promise.all` over a content type's actions in `release.publish` with sequential `for` loops, so within content-type publishes serialize on the shared transaction. Also drops the empty `try/catch` wrapping the cron task in `scheduling.ts` so task failures propagate to Strapi existing `job.on('error')` handler instead of being silently swallowed.

### Why is it needed?
A Content Release in production failed when publishing two documents of the same content type that referenced each other (parent `Page` and child `Page` via a self-referential relation, both in one release):

```
code:        23503
constraint:  pages_parent_lnk_fk
detail:      Key (page_id)=(255) is not present in table "pages".
table:       pages_parent_lnk
message:     insert into "pages_parent_lnk" ("id", "inv_page_id", "page_id")
             values ($1, $2, $3), ($4, $5, $6), ($7, $8, $9), ($10, $11, $12), ($13, $14, $15)
             - insert or update on table "pages_parent_lnk" violates foreign key constraint "pages_parent_lnk_fk"
```

The error was only surfaced when a release was triggered manually. Scheduled releases triggered by cron had been hitting the same failure mode (we assume) but the empty `try/catch` in `scheduling.ts`'s task callback was swallowing the error, so nothing was being logged.

Both publishes shared a transaction via `Promise.all`. It is most likely that statements interleaved on the connection, so one branch's delete invalidated the other branch's pre-captured join-table snapshot before the second branch tried to re-insert against it. The deleted `page_id=255` is the row the other branch had already removed.

User-visible state after the failure: new published rows existed for both documents, but the parent's join rows to all its children were missing, leaving every child orphaned on the published side. The release was recorded as `status='failed'` and a 500 error returned.

### How to test it?
`yarn jest packages/core/content-releases`. New unit test pins the sequential dispatch.

### DIFF
--- a/packages/core/content-releases/server/src/services/__tests__/release.test.ts
+++ b/packages/core/content-releases/server/src/services/__tests__/release.test.ts
@@ -526,6 +526,88 @@ describe('Release service', () => {
       expect(articleIndex).toBeGreaterThanOrEqual(0);
       expect(categoryIndex).toBeLessThan(articleIndex);
     });
+
+    it('publishes documents within a content type sequentially', async () => {
+      const PAGE_UID = 'api::page.page';
+
+      mockExecute.mockReturnValueOnce({ id: 1, releasedAt: null });
+
+      const callOrder: string[] = [];
+      let resolveFirst: () => void;
+      const firstPublishGate = new Promise<void>((resolve) => {
+        resolveFirst = resolve;
+      });
+
+      const publishMock = jest
+        .fn()
+        .mockImplementationOnce(async (params: { documentId: string }) => {
+          callOrder.push(`call:${params.documentId}`);
+          await firstPublishGate;
+          callOrder.push(`resolve:${params.documentId}`);
+        })
+        .mockImplementationOnce(async (params: { documentId: string }) => {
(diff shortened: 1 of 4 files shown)


innerdvations (approved):

OcneanuVlad:
Internal note: TID 8744

garrettheaver:
Can I get an update on this please in terms of when it will go into a release? It's an EE feature affecting an EE customer. Thanks!

OcneanuVlad:
> Can I get an update on this please in terms of when it will go into a release? It's an EE feature affecting an EE customer. Thanks!

Hey Garett, it looks like the fix is currently scheduled to release with 5.46.0 according to the milestones

garrettheaver:
Thanks. It had been scheduled for 5.44.1, then it got bumped to 5.45.0, then 5.45.1. Now it's 5.46.0. Given the internal TID / it's on an EE support agreement, can you help prevent it from getting bumped to a later version again?

PR #26148, opened 2026-04-28
fix(content-manager): prevent crash on detached DZ component

### LINKED ISSUE #23300: Deleting a component or removing it from Dynamic zone renders content-history unusable
### Node Version

20.18.2

### NPM/Yarn/PNPM Version

Yarn 1.22.22

### Strapi Version

5.12.3

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When using an enterprise license with content-history and you have a content-type that contains a dynamic zone. If you remove (detach) or delete a component that is contained within the dynamic zone the content-history will no longer function properly for that content-type

### Steps to Reproduce

1. Create a project and attach an EE license that has content-history
2. Create a content-type with a dynamic zone and two components added to it
3. Create an entry and modify it a few times to get some history records
4. Detach or delete the component from the dynamic zone
5. Go back to the entry and try to view the history or restore a previous version
6. See Error

### Expected Behavior

Content-history should still be able to show the previous versions since we store the previous schema with the history record (even if we can't restore missing fields)

### Logs

```shell
See reproduction video
```

### Code Snippets

See reproduction video

### Media

Reproduction Video: https://www.loom.com/share/58f75bb25063425da189cfa08bf59ba5?sid=40090e81-885d-4d4f-a96f-aef7478ad2e1

### Additional information

Related (partially) to new TID689. The customer was experiencing a different issue that I couldn't reproduce and instead found this one instead.

### Confirmation Checklist

Adzouz:
### What does it do?

Adds an early return in `traverseData` (`packages/core/content-manager/admin/src/pages/EditView/utils/data.ts`) when an attribute is absent from the schema dict. The traversal now passes the value through unchanged instead of crashing on `attribute.type`.

### Why is it needed?

When a component is detached from a dynamic zone (or deleted) via the content-type-builder, `extractContentTypeComponents` rebuilds the component dictionary from the current content type only — the orphaned component is dropped. Stored history version `data` still references it via `__component`.
The recursive traversal in `prepareTempKeys` lands on `attributes = {}` and throws `Cannot read properties of undefined (reading 'type')` at line 84, blocking the entire History page.

### How to test it?

1. Create a content type with a dynamic zone and add 2 or more components
2. Create an entry, add at least one of each component to the DZ, save
3. Open the content-type-builder → edit the DZ → remove one component from the allowed list → save
4. Go back to the entry → open Content History
→ Before the fix: the page crashes with `Cannot read properties of undefined (reading 'type')`                                                         
→ After the fix: history versions render. Orphaned DZ items are present in the data without per-field rendering.

Unit test: `yarn test:front --testPathPattern="EditView/utils/tests/data.test"`

### Related issue(s)/PR(s)

Fix #23300

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
+++ b/packages/core/content-manager/admin/src/pages/EditView/utils/data.ts
@@ -81,6 +81,15 @@ const traverseData =
           return acc;
         }
 
+        // The schema may not contain the attribute when historical data references a
+        // component or field that no longer exists (e.g. a component detached from a
+        // dynamic zone via the content-type-builder). Pass the value through so the
+        // History view can still render the rest of the document.
+        if (!attribute) {
+          acc[key] = value;
+          return acc;
+        }
+
         if (attribute.type === 'component') {
           if (attribute.repeatable) {
             const componentValue = (
--- a/packages/core/content-manager/admin/src/pages/EditView/utils/tests/data.test.ts
+++ b/packages/core/content-manager/admin/src/pages/EditView/utils/tests/data.test.ts
@@ -1,5 +1,10 @@
 import { testData } from '../../../../tests/data';
-import { getDirectParent, handleInvisibleAttributes, removeProhibitedFields } from '../data';
+import {
+  getDirectParent,
+  handleInvisibleAttributes,
+  prepareTempKeys,
+  removeProhibitedFields,
+} from '../data';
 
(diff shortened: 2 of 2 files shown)


mathildeleg (approved):
LGTM thanks ✅

Issue #23177, opened 2025-03-18
Issue with Reserved Field Names in Strapi 5 (After Upgrading from Version 4)


shkarinn:
### Node Version

18.20.2

### NPM/Yarn/PNPM Version

PNPM 10.6.2

### Strapi Version

5.11.2

### Operating System

Windows 11

### Database

SQLite

### Javascript or Typescript

Javascript

### Bug Description

In Strapi 5, reserved field names have been introduced, which causes conflicts with existing collections where a `meta` field is used for components. When upgrading from version 4 to 5, the `meta` field remains in the collections, and when trying to add a new field via the admin panel, an error is displayed stating that the field `meta` is reserved.

### Steps to Reproduce

1. Upgrade Strapi from version 4 to version 5, with collections (branches, etc.) that include a `meta` field.
2. Open the admin panel and try to add a new field to the collection **branches**.
3. Receive an error message stating that the `meta` field is reserved.
4. Perform a migration using Knex and update the `schema.json`.
5. Although the `meta` field remains after starting the application, adding a new field causes data loss in the `branches_cmps` table.

### Expected Behavior

- After migration, if the field is renamed (e.g., to `metaData`), the data should be preserved, and the admin panel should correctly handle the addition of new fields.
- The system should ensure compatibility with reserved field names without data loss.

## Actual Behavior
- The admin panel prevents adding a new field because `meta` is reserved.
- Migrating via Knex (renaming the field) results in data loss from the `branches_cmps` table when a new field is added.

## Knex Migration Example
```javascript
exports.up = async function(knex) {
  await knex('branches_cmps')
    .where('field', 'meta')
    .update({ field: 'metaData' });
};

exports.down = async function(knex) {
  await knex('branches_cmps')
    .where('field', 'metaData')
    .update({ field: 'meta' });
};```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

alexandrebodin:
Your example knex migration shows you are trying to change the value of a column named field from meta to metaData. However you are talking about having an attribute named meta which would mean you need to do a renameColumn migration instead. 

Can you share the schema.json of your app to understand better what you are trying to do please

shkarinn:
[...]

PR #26149, opened 2026-04-28
fix(core): reject 'status' attribute when draftAndPublish is enabled

unrevised6419:
### What does it do?

- Bootstrap content-type validator (`packages/core/core/src/domain/content-type/validator.ts`) now rejects an attribute named `status` when the content type has `draftAndPublish: true`. Error message points the user at either renaming the attribute or disabling draft-and-publish.
- Centralizes reserved attribute / model name lists in `@strapi/utils` (`packages/core/utils/src/content-types.ts`). New exports: `RESERVED_ATTRIBUTE_NAMES`, `RESERVED_ATTRIBUTE_NAMES_DRAFT_PUBLISH`, `RESERVED_MODEL_NAMES`, `isReservedAttributeName(name, { draftAndPublish })`, `isReservedModelName`, `getReservedAttributeNames`, `getReservedModelNames`.
- `@strapi/content-type-builder`'s `services/builder.ts` now re-exports from `@strapi/utils`. Public surface (`reservedAttributes`, `reservedModels`, `getReservedNames`, `isReservedAttributeName`, `isReservedModelName`) is unchanged.
- Adds unit tests for the validator and updates one CTB test that asserted the exact ordering of the reserved-attribute list.

### Why is it needed?

In Strapi v5 the `status` query parameter on REST / GraphQL / Document Service APIs filters draft vs published versions of a document (see https://docs.strapi.io/cms/migration/v4-to-v5/breaking-changes/publication-state-removed). A user-defined attribute named `status` on a draft-and-publish content type therefore collides with the framework's own query parameter, leading to confusing API behavior.

The CTB UI already blocked this name at content-type creation time, but hand-edited `schema.json` files bypassed that check and only failed later, in less obvious ways. The bootstrap validator now catches the conflict at startup with a clear message.

The reserved-name lists were previously duplicated between CTB and the runtime transform with a TODO to consolidate. Moving them into `@strapi/utils` removes that drift risk and lets the bootstrap validator share the same source.

### How to test it?

1. In any Strapi v5 project (e.g. `examples/getstarted`), pick a content type with `draftAndPublish: true` and add an attribute named `status` to its `schema.json`.
2. Start the server (`yarn develop`). Boot fails with:
   > The attribute name 'status' is reserved when 'draftAndPublish' is enabled. Rename the attribute or disable the 'draftAndPublish' option.
3. Either rename the attribute or set `options.draftAndPublish: false` and confirm the server boots normally.
4. Internal content types that legitimately declare always-reserved names (e.g. `release-action` declaring `locale`) continue to boot — the bootstrap check is intentionally scoped to the draft-and-publish-conditional list.
5. Run the tests:
   - `yarn workspace @strapi/core test:unit src/domain/content-type` (5 cases)
   - `yarn workspace @strapi/content-type-builder test:unit` (93 cases)
   - `yarn workspace @strapi/utils jest` (428 cases)

### Related issue(s)/PR(s)

- Related to https://github.com/strapi/strapi/issues/23177 — "Issue with Reserved Field Names in Strapi 5 (After Upgrading from Version 4)". This PR addresses the `status` × `draftAndPublish` collision specifically; broader bootstrap enforcement of always-reserved names was deferred because internal content types (e.g. `release-action.locale`) legitimately declare some of them.
- Resolves the runtime side of the v5 `publicationState` → `status` migration: https://docs.strapi.io/cms/migration/v4-to-v5/breaking-changes/publication-state-removed
- Addresses the centralization TODOs at `packages/core/content-type-builder/server/src/services/builder.ts` and `packages/core/core/src/utils/transform-content-types-to-models.ts:291` (the latter still left for a follow-up).

### DIFF
--- a/packages/core/content-type-builder/server/src/controllers/validation/__tests__/content-type.test.ts
+++ b/packages/core/content-type-builder/server/src/controllers/validation/__tests__/content-type.test.ts
@@ -57,13 +57,13 @@ describe('Content type validator', () => {
         expect(err).toMatchObject({
           name: 'ValidationError',
           message:
-            'Attribute keys cannot be one of id, document_id, created_at, updated_at, published_at, created_by_id, updated_by_id, created_by, updated_by, entry_id, status, localizations, meta, locale, __component, __contentType, strapi*, _strapi*, __strapi*',
+            'Attribute keys cannot be one of id, document_id, created_at, updated_at, published_at, created_by_id, updated_by_id, created_by, updated_by, entry_id, localizations, meta, locale, __component, __contentType, strapi*, _strapi*, __strapi*, status',
           details: {
             errors: [
               {
                 path: ['contentType', 'attributes', 'entryId'],
                 message:
-                  'Attribute keys cannot be one of id, document_id, created_at, updated_at, published_at, created_by_id, updated_by_id, created_by, updated_by, entry_id, status, localizations, meta, locale, __component, __contentType, strapi*, _strapi*, __strapi*',
+                  'Attribute keys cannot be one of id, document_id, created_at, updated_at, published_at, created_by_id, updated_by_id, created_by, updated_by, entry_id, localizations, meta, locale, __component, __contentType, strapi*, _strapi*, __strapi*, status',
                 name: 'ValidationError',
               },
             ],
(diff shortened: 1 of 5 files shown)


innerdvations:
Thanks for this! The direction makes sense. I'm just concerned that making it a hard fail will break existing projects that are already running v5 with `status` attributes. Could we change it to a startup warning (once per affected content type UID)?

unrevised6419:
Closing in favor of https://github.com/strapi/strapi/pull/26890

PR #26150, opened 2026-04-28
fix(content-manager): replace sanitize-html with dompurify in Wysiwyg preview

unrevised6419:
### What does it do?

Replaces `sanitize-html` with [`DOMPurify`](https://github.com/cure53/DOMPurify) in the markdown preview of the Wysiwyg field (`PreviewWysiwyg`).

- New util `admin/.../Wysiwyg/utils/sanitizer.ts` wraps `DOMPurify.sanitize` with config that mirrors the prior allowlist:
  - `ALLOWED_ATTR`: union of the prior `*`/`img`/`source` attribute lists (`href, src, alt, align, center, width, height, type, controls, target`).
  - `ALLOWED_URI_REGEXP`: `http(s)`, `ftp`, `mailto`, `tel`, plus protocol-relative, fragment, and relative URLs (matches the prior `allowedSchemes`).
  - `ALLOW_DATA_ATTR: false` to match the prior allowlist (which did not include `data-*`).
- New test file `utils/tests/sanitizer.test.ts` pins the contract (~40 cases): markdown tag pass-through, allowed/stripped attribute matrix, URL scheme allow/deny matrix, end-to-end markdown render.
- Removed `sanitize-html` + `@types/sanitize-html` from `@strapi/admin` (dead deps — never imported), removed `sanitize-html` from `@strapi/content-manager` (replaced with `dompurify@3.4.1`), and removed the `sanitize-html` entry from the vite `optimizeDeps.include` list in `packages/core/strapi/src/node/vite/config.ts`.
- Removed the now-stale `console.warn` suppression in `Wysiwyg/tests/Field.test.tsx` — that suppression existed solely to silence sanitize-html's runtime advisory about allowing `script`/`style` tags. DOMPurify does not emit such a warning.

**Behavior note:** prior config used `allowedTags: false`, which preserved `<script>` and `<style>` in the rendered preview. DOMPurify removes `<script>` unconditionally (CSS `<style>` is still preserved by default). Preserving raw `<script>` in the preview was an XSS hole rather than a feature, so this is treated as an intentional tightening. The sanitizer test documents this.

### Why is it needed?

`sanitize-html` is a Node.js library (per its own [README](https://github.com/apostrophecms/apostrophe/tree/main/packages/sanitize-html#requirements)) and is [not recommended for browser use upstream](https://github.com/apostrophecms/sanitize-html/issues/639). It transitively imports `postcss`, which depends on `path`, `fs`, `url`, and `source-map-js`. When the admin panel boots in development, vite logs roughly a dozen `Module "<x>" has been externalized for browser compatibility. Cannot access "<x>.<fn>" in client code` warnings in the browser console, e.g.:

```
Module "path" has been externalized for browser compatibility. Cannot access "path.isAbsolute" in client code.
Module "fs" has been externalized for browser compatibility. Cannot access "fs.readFileSync" in client code.
Module "source-map-js" has been externalized for browser compatibility. Cannot access "source-map-js.SourceMapConsumer" in client code.
Module "url" has been externalized for browser compatibility. Cannot access "url.fileURLToPath" in client code.
```

These warnings appear on every fresh admin boot and on HMR reloads, polluting the console for every developer working in the admin. DOMPurify is a browser-native library with zero Node module dependencies and ships its own TypeScript types, so the warnings disappear and `@types/sanitize-html` is no longer needed.

**Library size comparison** (via [bundlephobia](https://bundlephobia.com/) — captured in [#23821](https://github.com/strapi/strapi/pull/23821)):

| `sanitize-html` | `dompurify` |
| --- | --- |
| sanitize-html bundle | dompurify bundle |

**Warnings in the browser console (before this change):**

sanitize-html browser warnings

### How to test it?

1. Create a new Strapi app (or use `examples/getstarted`) on this branch.
2. `yarn develop` and open the admin panel in a browser.
3. Open the browser devtools console — there should no longer be any `Module "<x>" has been externalized for browser compatibility` warnings on initial admin load.
4. Open a content type with a `richtext` field, type some markdown (headings, links, images, code blocks, lists), and click the **Preview** toggle in the Wysiwyg toolbar — markdown should render the same as before (links navigable, images visible, code highlighted layout, etc.).
5. Run the unit tests:
   ```
   yarn workspace @strapi/content-manager run test:front --testPathPattern Wysiwyg
   ```
   All Wysiwyg + sanitizer + mdRenderer suites should pass.

### Related issue(s)/PR(s)

- Supersedes #23821 (earlier attempt by the same author; closed).
- Upstream library issue: [apostrophecms/sanitize-html#639](https://github.com/apostrophecms/sanitize-html/issues/639).


nclsndr:
@markkaylor @mathildeleg @Adzouz any reason we should change to DOMPurify?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/sanitizer.ts
```diff
@@ -0,0 +1,40 @@
...
+ *   '*':    href, align, alt, center, width, height, type, controls, target
+ *   img:    src, alt
+ *   source: src, type
+ *
+ * DOMPurify's ALLOWED_ATTR is a flat list (no per-tag scoping), so the union
+ * of the original allowlist is used. Tags that the prior config never reached
+ * (script, iframe, etc.) are stripped wholesale by DOMPurify's default tag
+ * allowlist, so per-tag attribute scoping is not security-relevant here.
+ */
+const ALLOWED_ATTR = [
```
nclsndr, line 26: `class` stripping silently kills highlight.js — `solarized-dark.css` is dead code

`mdRenderer.ts` emits `class` on every highlighted code block:

```ts
'<pre class="hljs language-' + md.utils.escapeHtml(lang) + '"><code>'
// + inner token spans with class="hljs-keyword", "hljs-string", etc.
```

And on footnote refs/backrefs:

```ts
return '<sup class="footnote-ref"><span>' + caption + '</span></sup>';
return ' <span class="footnote-backref">\u21a9\uFE0E</span>';
```

The sanitizer strips ALL `class` attributes. Consequence:

1. `import 'highlight.js/styles/solarized-dark.css'` in `mdRenderer.ts` is **dead code** — every CSS rule targets `.hljs`, `.hljs-keyword`, etc., none of which survive sanitization. The code block background shows only because it's hard-coded in the styled component.
2. Footnote ref / backref styling breaks silently.

This is pre-existing (sanitize-html also stripped `class`), but the PR now **documents the stripping as correct behavior** in the test using `<pre class="hljs language-js">` as the exact example — without acknowledging the consequence.

**Fix:** add `'class'` to `ALLOWED_ATTR`. Security risk is negligible — this is admin-authored content rendered in the admin panel, the tag allowlist blocks anything dangerous, and `class` cannot escalate privileges.

nclsndr (commented):
Thank you so much for this contribution @unrevised6419!
Looking forward doing proper QA on it.

Preliminary AI review (filtered by myself) gives:

---

**Good**

- Extraction to `sanitizer.ts` is the right design — inline config in `PreviewWysiwyg.tsx` was invisible and untestable.
- Removing the `console.warn` suppression in `Field.test.tsx` is a clean win.
- `ALLOW_DATA_ATTR: false` is the right explicit opt-out.
- URL scheme matrix test coverage is thorough and direct.
- Dep tree pruning (postcss, htmlparser2, launder, dayjs, parse-srcset, @types/sanitize-html) is the main win

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/sanitizer.ts
nclsndr: Without `class`
Screenshot 2026-06-10 at 13 59 33
With `class`
Screenshot 2026-06-10 at 13 59 38

unrevised6419:
### Parity validation + follow-up findings

To verify the `sanitize-html` → DOMPurify swap matches **in practice, not just at first glance**, I built a standalone differential test suite that runs both configs (the exact old `sanitize-html` options and the new DOMPurify options) over the same 230 inputs and compares normalized output:

📦 https://github.com/unrevised6419/strapi-wysiwyg-sanitizer-parity (593 Vitest cases, CI on every push)

**Results**
- ✅ All safe markdown-it output is byte-identical after normalization; disallowed attributes and URL-scheme decisions agree across both.
- ✅ **Security upgrade confirmed:** 3 inputs leave an executable gadget in the *old* config's output (raw `<script>`, plus the `svg/style` and `math/mglyph` mXSS classics). The new config strips all three — **zero** live-XSS primitives survive it across all 230 inputs.

**Two places the new config was *more permissive* than the old (both inert, neither an XSS vector):**

1. **`aria-*` attributes pass through.** DOMPurify's `ALLOW_ARIA_ATTR` defaults to `true`, independent of `ALLOWED_ATTR`; the old config stripped all `aria-*`. For exact parity, set `ALLOW_ARIA_ATTR: false`.
2. **`data:` URIs survive on media tags — ✅ now fixed.** DOMPurify permits `data:` on `img`/`source`/`video`/`audio` regardless of `ALLOWED_URI_REGEXP` — including `data:image/*`, `data:image/svg+xml`, and even `data:text/html;base64` on `` (inert there). The old config stripped **all** `data:`. Fixed by an `afterSanitizeAttributes` hook that strips any `data:` value from `src`/`href`/`xlink:href`, restoring parity; the suite now asserts both configs drop every `data:` input. (`data:text/html` in a navigable `<a href>` was already blocked by both.)

Separately, I pushed a commit allowing `class` + `title` (the old config stripped both, which silently killed highlight.js code-block styling + footnote styling — `solarized-dark.css` was effectively dead). `id` left out to avoid DOM clobbering.

cc @nclsndr

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/tests/sanitizer.test.ts
```diff
@@ -0,0 +1,194 @@
...
+ *       '*':    href, align, alt, center, width, height, type, controls, target
+ *       img:    src, alt
+ *       source: src, type
+ *   - allowedSchemes:   http, https, ftp, mailto, tel (+ protocol-relative)
+ *
+ * Plus two additions over the old config: `class` (highlight.js + footnote
+ * styling, emitted by mdRenderer.ts) and `title` (link/image/abbr tooltips).
+ *
+ * Anything outside that allowlist must be stripped.
+ */
```
nclsndr: Cosmetics. We probably don't want these comments here since it'll be duplicated from the production code comments (packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/sanitizer.ts)
```suggestion
```

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/sanitizer.ts
```diff
@@ -40,6 +40,26 @@ const ALLOWED_ATTR = [
...
 
+/**
+ * DOMPurify permits `data:` URIs on media tags (img/source/video/audio)
+ * through its internal DATA_URI_TAGS allowlist, *bypassing* ALLOWED_URI_REGEXP.
+ * The prior sanitize-html config stripped every `data:` URI, so strip them here
+ * too for parity. Registered once on the module's DOMPurify instance, which is
+ * the only consumer in the admin.
+ */
+const URL_ATTRIBUTES = ['src', 'href', 'xlink:href'];
+DOMPurify.addHook('afterSanitizeAttributes', (node) => {
```
nclsndr, line 54: Is there a way to have a singleton instance instead of tapping into the lib global?

packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/utils/sanitizer.ts
```diff
@@ -0,0 +1,70 @@
...
+ * fragment, and relative URLs. Mirrors sanitize-html's prior allowedSchemes.
+ */
+const ALLOWED_URI_REGEXP = /^(?:(?:https?|ftp|mailto|tel):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i;
+
+/**
+ * DOMPurify permits `data:` URIs on media tags (img/source/video/audio)
+ * through its internal DATA_URI_TAGS allowlist, *bypassing* ALLOWED_URI_REGEXP.
+ * The prior sanitize-html config stripped every `data:` URI, so strip them here
+ * too for parity. Registered once on the module's DOMPurify instance, which is
+ * the only consumer in the admin.
```
nclsndr:
```suggestion
 * too for parity.
```

nclsndr (dismissed):
Staggering speed 🏎️ ! Thank you so much @unrevised6419!
— we got aligned with the team toward dropping `sanitize-html` ✅ 
LGTM 🙌

innerdvations (approved):

innerdvations (approved):

nclsndr:
thank you @innerdvations, I meant to hit "approve" here, never did…

PR #26152, opened 2026-04-28
fix(document-service): discard-draft 500 on self-referential manyToMany relations

butcherZ:
### What does it do?

Fixes two bugs in the document service that caused `discardDraft` to throw a 500 on content types with a **unidirectional self-referential relation** (e.g. a `manyToMany` targeting the same content type, with no `inversedBy`).

1. **`self-referential-relations.ts`** — `sync()` was converting target entry IDs to strings via `String()` before inserting into join table FK columns. PostgreSQL rejects string values for integer FK columns, causing the 500 on PG. Fixed by removing the `String()` wrapper so the original DB integer type is preserved.

2. **`unidirectional-relations.ts`** — `load()` iterates all content types looking for unidirectional relations targeting the uid being discarded. It was also matching the source content type's *own* self-referential attributes (where `model.uid === uid`). Those rows were then re-inserted by `sync()` with a stale source FK pointing to the already-deleted old draft entity — causing a FK constraint violation on both SQLite and PostgreSQL. Fixed by adding a `model.uid === uid` guard to skip self-referential attributes, which are already handled by `selfReferentialRelations`.

### Why is it needed?

`discardDraft` deletes old draft entries and creates new ones from the published version. For self-referential relations, the join table rows referencing old entity IDs must be remapped to the new entity IDs. Two separate code paths were both attempting to remap the same rows, but `unidirectional-relations.sync` only remaps the target column — leaving the source column pointing to a deleted entity, which fails FK constraints.

### How to test it?

1. Create a content type with a unidirectional self-referential `manyToMany` relation (e.g. `Node` with a `related` field targeting itself)
2. Create an entry, set `related` to point to itself, and publish it
3. Make any change to the draft
4. Click **Discard changes** — previously returned 500, now succeeds

### Related issue(s)/PR(s)

Related to #25890

### DIFF
--- a/packages/core/core/src/services/document-service/utils/self-referential-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/self-referential-relations.ts
@@ -1,10 +1,10 @@
 /* eslint-disable no-continue */
 import { keyBy, omit } from 'lodash/fp';
-import type { UID } from '@strapi/types';
+import type { Data, UID } from '@strapi/types';
 import type { JoinTable } from '@strapi/database';
 
 interface VersionEntry {
-  id: string;
+  id: Data.ID;
   locale: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/core/core/src/services/document-service/utils/self-referential-relations.ts
```diff
@@ -95,15 +95,15 @@ const sync = async (
...
   const idMapping = sourceEntries.reduce(
     (acc, sourceEntry) => {
       const targetEntry = targetEntriesByLocale[sourceEntry.locale];
       if (!targetEntry) return acc;
-      acc[String(sourceEntry.id)] = String(targetEntry.id);
+      acc[String(sourceEntry.id)] = targetEntry.id;
       return acc;
     },
-    {} as Record<string, string>
+    {} as Record<string, unknown>
```
Adzouz: If it's an int, why not a number here?

Adzouz (commented):
Thanks for the quick hotfix, just have a little question 🙏😊

packages/core/core/src/services/document-service/utils/self-referential-relations.ts
butcherZ: @Adzouz oops, you are right, I used unknown cuz the `VersionEntry` interface declared `id: string`, but yeah in reality I think it will alway be an int coming from the DB, fixed! 😄

Adzouz (approved):
Looks good, thanks a lot for the changes @butcherZ 🙏

PR #26165, opened 2026-04-29
fix(admin): redirect active tab to login on session expiry

### LINKED ISSUE #26163: [Bug]: Admin panel does not auto-redirect to login on idle session timeout
### Node Version

22.22.2

### Package Manager

yarn

### Package Manager Version

1.22.22

### Strapi Version

5.44.0 (also reproduces on `develop`; originally reported on 5.36.1)

### Operating System

Linux (Debian/Ubuntu)

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_(none — reproducible on a fresh `npx create-strapi-app` project with no modifications)_

### Bug Description

When a logged-in admin's session expires (access/refresh token TTLs elapsed, or configured idle session lifespan exceeded), the admin panel does **not** auto-redirect to `/auth/login`. The user keeps seeing the last page they were on. Redirect only fires when the user takes an action that triggers an HTTP request (sidebar click, refresh, etc.) — that request returns 401, and *then* they're sent to login.

A workstation left idle past the configured session lifespan is visually indistinguishable from one that is still authenticated, which defeats the purpose of an idle-session policy and is a finding under STIG/DISA-style hardening checklists that mandate automatic session termination with explicit indication to the user.

### Steps to Reproduce

1. Create a Strapi v5 project (`npx create-strapi-app my-app`).
2. In `config/admin.ts`, configure short session lifespans to make the issue easy to observe:

   ```ts
   export default ({ env }) => ({
     auth: {
       sessions: {
         maxRefreshTokenLifespan: 60,
         maxSessionLifespan: 60,
         idleRefreshTokenLifespan: 30,
         idleSessionLifespan: 30,
       },
     },
   });
   ```

3. `yarn develop`, log in, navigate to any list view (e.g. Settings → Users).
4. Stop interacting with the page. Wait past the configured `idleSessionLifespan` (and a little extra so the access token has also expired).
5. Continue to **not** interact. Observe the UI still renders as if logged in.
6. Click any sidebar item or refresh — only now are you redirected to `/auth/login`.

### Expected Behavior

When the session expires (idle threshold or absolute lifespan), the admin panel should:

1. Detect expiry without waiting for the next user-initiated request.
2. Automatically navigate to `/auth/login`.
3. Optionally surface a message ("You were signed out due to inactivity") so the user understands why.

A reasonable minimum bar: as soon as the session is invalidated (locally known TTL elapsed, or 401 returned by any background API call), the **active tab** redirects to login — not just other tabs via a `storage` event.

### Logs

```shell
# (no server-side error — purely a client-side UX gap)
```

### Code Snippets

Relevant paths in `develop`:

**1. Reactive 401 handler in the active tab dispatches logout but never navigates** — [`packages/core/admin/admin/src/utils/baseQuery.ts`](https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/utils/baseQuery.ts#L54-L72):

```ts
if (err.status === 401 && !isAuthPath(url)) {
  try {
    const { post } = getFetchClient();
    await post('/admin/logout');
  } catch { /* no-op */ }

  dispatch(logoutAction());     // ← clears Redux/localStorage…
  // …but no navigate('/auth/login') call here.
}
```

**2. The path that *does* navigate is only invoked on manual logout or via the cross-tab `storage` event** — [`packages/core/admin/admin/src/features/Auth.tsx`](https://github.com/strapi/strapi/blob/develop/packages/core/admin/admin/src/features/Auth.tsx#L136-L179):

```tsx
const clearStateAndLogout = React.useCallback(() => {
  dispatch(adminApi.util.resetApiState());
  dispatch(logoutAction());
  navigate('/auth/login');     // ← the actual redirect lives here
}, [dispatch, navigate]);

// storage events do not fire in the tab that wrote the change,
// so the active tab never gets this signal from itself.
const handleUserStorageChange = (event: StorageEvent) => {
  if (event.key === STORAGE_KEYS.STATUS && event.newValue === null) {
    clearStateAndLogout();
  }
};
```

**3. No client-side idle timer.** Searching the admin source for `idleSessionLifespan` only finds server-side config; the frontend has no timer that proactively logs the user out when the configured idle threshold elapses, so detection is purely reactive on 401 responses.

### Additional information

**Why this matters**

- **STIG/DISA hardening posture.** Customers in regulated environments need session termination to be visible and immediate, not contingent on the user clicking something.
- **Token expiry is server-authoritative**, so the session *is* terminated on the backend the moment it expires — the issue is purely that the SPA doesn't reflect that until forced.

**Suggested fix shape (for triage discussion)**

Two complementary changes:

1. **Make the active tab redirect on its own 401** — in `baseQuery.ts`, after `dispatch(logoutAction())`, also signal navigation. Options:
   - Lift the 401 handler into `AuthProvider` so it can call `clearStateAndLogout()` (which already navigates), or
   - Add a top-level route guard that watches `state.admin_app.token` and navigates to `/auth/login` when it transitions to `null`.
2. **Add a client-side idle timer** driven by `idleSessionLifespan` so the redirect happens at the actual idle threshold instead of waiting for a stray API call to fail. Also enables the "you were signed out due to inactivity" message.

The first change resolves the customer-visible symptom; the second makes the UX match the documented behavior of the idle-session config.

**Originally reported**

Strapi support ticket 8719 — STIG-related session-handling questions from a customer running 5.36.1.

derrickmehaffy:
### What does it do?

Two complementary changes in the admin panel so an expired session is reflected in the UI immediately, instead of leaving the user on a stale page until they click something or refresh:

- **Active-tab redirect on 401.** `baseQuery`'s 401 handler now invokes a session-expired callback in addition to dispatching `logoutAction`. `AuthProvider` registers `clearStateAndLogout` as that callback, so the tab that originated the failing request is navigated to `/auth/login`. Previously, only *other* tabs got that treatment via the `storage` event (which doesn't fire in the originating tab).
- **JWT-exp idle timer.** `AuthProvider` schedules a one-shot timer at the access token's `exp` claim. While the user is active, every API call that hits a 401 transparently refreshes the token (existing behavior in `withTokenRefresh`), which updates the Redux token and re-runs the effect to reschedule the timer to the new `exp`. If no API activity occurs during the access token lifespan, the token in Redux never changes, the timer fires, and the user is logged out — matching the configured server-side idle behavior without needing a separate activity tracker.

The mechanism is wired through a new `setOnSessionExpired` / `triggerSessionExpired` pair in `getFetchClient.ts`, mirroring the existing `setOnTokenUpdate` registry. This keeps the React layer as the only place that owns navigation; the fetch / RTK layer just signals when navigation should happen.

Files touched:

| File | Change |
|---|---|
| `packages/core/admin/admin/src/utils/getFetchClient.ts` | Adds `setOnSessionExpired` / `triggerSessionExpired` |
| `packages/core/admin/admin/src/utils/baseQuery.ts` | Calls `triggerSessionExpired()` after `dispatch(logoutAction())` on 401 |
| `packages/core/admin/admin/src/features/Auth.tsx` | Registers `clearStateAndLogout` as the callback + adds JWT-exp idle timer |
| `packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts` | 3 new unit tests for the callback registry |

### Why is it needed?

When an admin's session expires (access/refresh token TTLs elapsed, or the configured idle session lifespan exceeded), the admin panel does not auto-redirect to `/auth/login`. The user keeps seeing the last page they were on. The redirect only fires when the user takes an action that triggers an HTTP request — at which point the request returns 401, and *then* they're sent to login.

A workstation left idle past the configured session lifespan is visually indistinguishable from one that is still authenticated, which defeats the purpose of an idle-session policy and is a finding under STIG/DISA-style hardening checklists that mandate automatic session termination with explicit indication to the user.

**Why JWT-exp instead of an activity-event tracker?** Tracking `mousemove` / `keydown` / etc. would require a separate idle threshold value to be available on the client (which is currently server-only). Using the access token's `exp` reuses an authoritative value the client already receives, with self-resetting semantics: any user-driven request refreshes the token, the new `exp` is later than the old one, the effect re-runs, and the timer is rescheduled. No new server endpoint needed.

### How to test it?

Environment: any Strapi v5 admin app on this branch, against a fresh `npx create-strapi-app` project. Configure short session lifespans in `config/admin.ts` to make timing observable:

```ts
export default ({ env }) => ({
  auth: {
    sessions: {
      maxRefreshTokenLifespan: 60,
      maxSessionLifespan: 60,
      idleRefreshTokenLifespan: 30,
      idleSessionLifespan: 30,
    },
  },
});
```

Manual paths to verify:

- [ ] **Idle redirect (no interaction).** Log in, navigate to any list view, stop interacting. After the access token's `exp` elapses, observe automatic redirect to `/auth/login` without clicking or refreshing. (Pre-fix behavior: page stays on the list view forever.)
- [ ] **Active-tab redirect on 401.** Log in, let the access token + refresh token both expire (server-side), then click any sidebar item. The originating tab should redirect to `/auth/login`. (Pre-fix behavior: redux state clears but the tab stays on the broken page until manual refresh.)
- [ ] **Active session is preserved.** Log in, click around continuously well past the access token lifespan. The token should refresh transparently (any 401 is auto-retried), and the timer should re-schedule each time the new token arrives. No spurious redirect.
- [ ] **Cross-tab logout still works.** Open the admin in two tabs. Log out in one. The other should redirect to `/auth/login` (existing `storage` event path).
- [ ] **Malformed JWT handling.** If the access token is malformed (no `exp` claim), the idle timer is skipped silently — the active-tab 401 redirect still covers the symptom.

Automated:

```bash
yarn workspace @strapi/admin test:front --testPathPattern="getFetchClient"
# 20/20 passing (3 new tests for setOnSessionExpired / triggerSessionExpired)
```

### Related issue(s)/PR(s)

- Fix #26163
- Originally surfaced via Strapi support ticket 8719 (STIG-related session-handling questions)

### Notes / follow-ups

- A natural extension is surfacing a "you were signed out due to inactivity" message on the login page when the redirect comes from this path (e.g. via React Router state). Skipped here to keep the change minimal; happy to do it in a follow-up.


derrickmehaffy:
exp workflow: https://github.com/strapi/strapi/actions/runs/25132188922

derrickmehaffy:
Exp Version: `0.0.0-experimental.48432a63d246375c55975a36d6276bafae82db2b`

jhoward1994:
What do you think about doing something like this ? That refactors `clearStateAndLogout` into a testable hook that packages/core/admin/admin/src/features/Auth.tsx can then import?


```
diff --git a/packages/core/admin/admin/src/features/Auth.tsx b/packages/core/admin/admin/src/features/Auth.tsx
index f68369e569..592c660b91 100644
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -6,6 +6,7 @@ import { Login } from '../../../shared/contracts/authentication';
 import { createContext } from '../components/Context';
 import { useTypedDispatch, useTypedSelector } from '../core/store/hooks';
 import { useStrapiApp } from '../features/StrapiApp';
+import { useIdleSessionLogout } from '../hooks/useIdleSessionLogout';
 import { useQueryParams } from '../hooks/useQueryParams';
 import { login as loginAction, logout as logoutAction, setLocale, setToken } from '../reducer';
 import { adminApi } from '../services/api';
@@ -92,28 +93,6 @@ const STORAGE_KEYS = {
   STATUS: 'isLoggedIn',
 };
 
-/**
- * Decode the `exp` claim of an admin access JWT and return it as milliseconds
- * since the epoch. Returns `null` if the token can't be parsed or has no
- * numeric `exp`. We do not verify the signature — the server is the source of
- * truth; this is purely to schedule a client-side idle timer.
- */
-const decodeAccessTokenExpiry = (token: string): number | null => {
-  const parts = token.split('.');
-  if (parts.length < 2) {
-    return null;
-  }
-
-  try {
-    // base64url → base64
-    const padded = parts[1].replace(/-/g, '+').replace(/_/g, '/');
-    const payload = JSON.parse(window.atob(padded));
-    return typeof payload?.exp === 'number' ? payload.exp * 1000 : null;
-  } catch {
-    return null;
-  }
-};
-
 const AuthProvider = ({
   children,
   _defaultPermissions = [],
@@ -200,46 +179,15 @@ const AuthProvider = ({
   }, [clearStateAndLogout]);
 
   /**
-   * Proactive idle-session detection.
-   *
-   * Schedule a one-shot timer for the access token's `exp` claim. While the
-   * user is active, every API call that hits a 401 transparently refreshes
-   * the access token (see `withTokenRefresh` in getFetchClient.ts), which
-   * updates the token in Redux and re-runs this effect with the new exp,
-   * pushing the deadline forward. If no API activity occurs during the
-   * access token lifespan, the token in Redux never changes, the timer
-   * fires, and we log the user out — matching the configured server-side
-   * idle behavior without needing a separate activity tracker.
-   *
-   * If the JWT can't be decoded (malformed, missing exp), we skip the
-   * timer; the active-tab redirect on 401 still covers the symptom.
+   * Proactive idle-session detection. See `useIdleSessionLogout` for the full
+   * rationale and the UX trade-off (a user actively typing in a form with no
+   * outgoing API calls will still be logged out at `exp`).
    */
-  React.useEffect(() => {
-    if (!token || _disableRenewToken) {
-      return undefined;
-    }
-
-    const expiry = decodeAccessTokenExpiry(token);
-    if (expiry === null) {
-      return undefined;
-    }
-
-    // Small buffer past `exp` to avoid firing in the same tick the server
-    // would still consider the token valid (clock skew tolerance).
-    const SESSION_EXPIRY_BUFFER_MS = 1000;
-    const msUntilExpiry = expiry - Date.now() + SESSION_EXPIRY_BUFFER_MS;
-
-    const timeoutId = window.setTimeout(
-      () => {
-        clearStateAndLogout();
-      },
-      Math.max(msUntilExpiry, 0)
-    );
-
-    return () => {
-      window.clearTimeout(timeoutId);
-    };
-  }, [token, clearStateAndLogout, _disableRenewToken]);
+  useIdleSessionLogout({
+    token,
+    onExpired: clearStateAndLogout,
+    disabled: _disableRenewToken,
+  });
 
   React.useEffect(() => {
     /**
diff --git a/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts b/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts
new file mode 100644
index 0000000000..40c51f03fc
--- /dev/null
+++ b/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts
@@ -0,0 +1,170 @@
+import { act, renderHook } from '@testing-library/react';
+
+import { useIdleSessionLogout, SESSION_EXPIRY_BUFFER_MS } from '../useIdleSessionLogout';
+
+/**
+ * Build a JWT-shaped string whose payload encodes `{ exp: <expSeconds> }`.
+ * We don't care about the header/signature segments here — the decoder
+ * doesn't read them.
+ */
+const buildJwt = (expSeconds: number): string => {
+  const json = JSON.stringify({ exp: expSeconds });
+  const base64url = window.btoa(json).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
+  return `header.${base64url}.signature`;
+};
+
+const ACCESS_TOKEN_LIFESPAN_MS = 30_000;
+const NOW_MS = 1_700_000_000_000;
+
+describe('useIdleSessionLogout', () => {
+  beforeEach(() => {
+    jest.useFakeTimers({ now: NOW_MS });
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('schedules a one-shot logout at the access token exp', () => {
+    const onExpired = jest.fn();
+    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+
+    renderHook(() => useIdleSessionLogout({ token, onExpired }));
+
+    expect(onExpired).not.toHaveBeenCalled();
+
+    act(() => {
+      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
+    });
+
+    expect(onExpired).toHaveBeenCalledTimes(1);
+  });
+
+  it('fires immediately when the token is already past exp', () => {
+    const onExpired = jest.fn();
+    const token = buildJwt((NOW_MS - 5_000) / 1000);
+
+    renderHook(() => useIdleSessionLogout({ token, onExpired }));
+
+    /**
+     * `setTimeout(fn, 0)` doesn't run synchronously, so we still need to
+     * advance the clock by a tick. The point of the test is that the timer
+     * is scheduled with a non-negative delay rather than being skipped.
+     */
+    act(() => {
+      jest.advanceTimersByTime(0);
+    });
+
+    expect(onExpired).toHaveBeenCalledTimes(1);
+  });
+
+  it('reschedules to the new exp when the token is refreshed', () => {
+    const onExpired = jest.fn();
+
+    /**
+     * Initial token: expires at NOW + 30s.
+     * After 20s of "user activity" we simulate a 401-triggered refresh
+     * (handled in `withTokenRefresh`) by passing a new token whose `exp`
+     * is later. The hook should clear the original timer and schedule a
+     * new one against the later `exp`. If the reschedule logic is broken,
+     * `onExpired` would fire when the original `exp` elapses.
+     */
+    const initialToken = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+
+    const { rerender } = renderHook(
+      ({ token }: { token: string }) => useIdleSessionLogout({ token, onExpired }),
+      { initialProps: { token: initialToken } }
+    );
+
+    act(() => {
+      jest.advanceTimersByTime(20_000);
+    });
+    expect(onExpired).not.toHaveBeenCalled();
+
+    /**
+     * Refresh: new token expires 30s from "now" (NOW + 20_000 + 30_000).
+     * Original `exp` would have been NOW + 30_000, i.e. 10s away. After
+     * rerender, the reschedule should push the deadline out by 20s beyond
+     * the original.
+     */
+    const refreshedToken = buildJwt((NOW_MS + 20_000 + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+    rerender({ token: refreshedToken });
+
+    /**
+     * Advance past the *original* exp deadline. If reschedule is broken the
+     * old timer would fire here.
+     */
+    act(() => {
+      jest.advanceTimersByTime(15_000);
+    });
+    expect(onExpired).not.toHaveBeenCalled();
+
+    /**
+     * Advance the rest of the way to the new exp + buffer. Total elapsed at
+     * this point is 20_000 + 15_000 + 15_000 + buffer = 50s + buffer, which
+     * is exactly 30s past "now" from the refresh point.
+     */
+    act(() => {
+      jest.advanceTimersByTime(15_000 + SESSION_EXPIRY_BUFFER_MS);
+    });
+    expect(onExpired).toHaveBeenCalledTimes(1);
+  });
+
+  it('cancels the pending timer when the token is cleared (logout)', () => {
+    const onExpired = jest.fn();
+    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+
+    const { rerender } = renderHook(
+      ({ token }: { token: string | null }) => useIdleSessionLogout({ token, onExpired }),
+      { initialProps: { token } as { token: string | null } }
+    );
+
+    rerender({ token: null });
+
+    act(() => {
+      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
+    });
+
+    expect(onExpired).not.toHaveBeenCalled();
+  });
+
+  it('does not schedule a timer when disabled', () => {
+    const onExpired = jest.fn();
+    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+
+    renderHook(() => useIdleSessionLogout({ token, onExpired, disabled: true }));
+
+    act(() => {
+      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
+    });
+
+    expect(onExpired).not.toHaveBeenCalled();
+  });
+
+  it('does not schedule a timer when the token is malformed', () => {
+    const onExpired = jest.fn();
+
+    renderHook(() => useIdleSessionLogout({ token: 'not-a-jwt', onExpired }));
+
+    act(() => {
+      jest.advanceTimersByTime(60_000);
+    });
+
+    expect(onExpired).not.toHaveBeenCalled();
+  });
+
+  it('cancels the pending timer on unmount', () => {
+    const onExpired = jest.fn();
+    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
+
+    const { unmount } = renderHook(() => useIdleSessionLogout({ token, onExpired }));
+
+    unmount();
+
+    act(() => {
+      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
+    });
+
+    expect(onExpired).not.toHaveBeenCalled();
+  });
+});
diff --git a/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts b/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts
new file mode 100644
index 0000000000..45315db399
--- /dev/null
+++ b/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts
@@ -0,0 +1,68 @@
+
+import { decodeAccessTokenExpiry } from '../utils/jwt';
+
+const SESSION_EXPIRY_BUFFER_MS = 1000;
+
+interface UseIdleSessionLogoutOptions {
+  /**
+   * The current admin access JWT (from Redux), or `null` when logged out.
+   */
+  token: string | null;
+  /**
+   * Called when the timer fires. Typically clears local auth state and
+   * navigates to /auth/login.
+   */
+  onExpired: () => void;
+  /**
+   * Escape hatch for tests / dev environments where we don't want the
+   * automatic logout behavior. Mirrors the existing `_disableRenewToken`
+   * prop on `<AuthProvider>`.
+   */
+  disabled?: boolean;
+}
+
+/**
+ * Schedule a one-shot logout when the access token's `exp` elapses.
+ *
+ * The hook re-runs whenever `token` changes. While the user is active, every
+ * API call that hits a 401 transparently refreshes the access token (see
+ * `withTokenRefresh` in `getFetchClient.ts`), which dispatches `setToken` and
+ * causes this effect to re-arm with the new, later `exp`.
+ *
+ * If the JWT can't be decoded (malformed, missing `exp`), the timer is
+ * skipped silently — the active-tab redirect on 401 still covers the
+ * symptom on the next user-initiated request.
+ */
+const useIdleSessionLogout = ({
+  token,
+  onExpired,
+  disabled = false,
+}: UseIdleSessionLogoutOptions): void => {
+  React.useEffect(() => {
+    if (!token || disabled) {
+      return undefined;
+    }
+
+    const expiry = decodeAccessTokenExpiry(token);
+    if (expiry === null) {
+      return undefined;
+    }
+
+    const msUntilExpiry = expiry - Date.now() + SESSION_EXPIRY_BUFFER_MS;
+
+    const timeoutId = window.setTimeout(
+      () => {
+        onExpired();
+      },
+      Math.max(msUntilExpiry, 0)
+    );
+
+    return () => {
+      window.clearTimeout(timeoutId);
+    };
+  }, [token, onExpired, disabled]);
+};
+
+export { useIdleSessionLogout, SESSION_EXPIRY_BUFFER_MS };
+export type { UseIdleSessionLogoutOptions };
diff --git a/packages/core/admin/admin/src/utils/jwt.ts b/packages/core/admin/admin/src/utils/jwt.ts
new file mode 100644
index 0000000000..0442189e51
--- /dev/null
+++ b/packages/core/admin/admin/src/utils/jwt.ts
@@ -0,0 +1,41 @@
+/**
+ * Decode the `exp` claim of an admin access JWT and return it as milliseconds
+ * since the epoch. Returns `null` if the token can't be parsed or has no
+ * numeric `exp`.
+ *
+ * The signature is intentionally not verified — the server is the source of
+ * truth for token validity. This is purely so the client can schedule a
+ * one-shot timer to react to access-token expiry without polling.
+ *
+ * Handles `base64url` payloads (RFC 7515) by mapping the URL-safe alphabet
+ * back to standard base64 and re-adding the `=` padding that `base64url`
+ * strips. Some browsers' `atob` implementations reject unpadded input.
+ */
+const decodeAccessTokenExpiry = (token: string): number | null => {
+  if (typeof token !== 'string') {
+    return null;
+  }
+
+  const parts = token.split('.');
+  if (parts.length < 2 || parts[1].length === 0) {
+    return null;
+  }
+
+  try {
+    // base64url → base64
+    const urlDecoded = parts[1].replace(/-/g, '+').replace(/_/g, '/');
+
+    // Restore `=` padding to a multiple of 4 chars; base64url omits it.
+    const remainder = urlDecoded.length % 4;
+    const padded = remainder === 0 ? urlDecoded : urlDecoded + '='.repeat(4 - remainder);
+
+    const payload = JSON.parse(window.atob(padded));
+    return typeof payload?.exp === 'number' && Number.isFinite(payload.exp)
+      ? payload.exp * 1000
+      : null;
+  } catch {
+    return null;
+  }
+};
+
+export { decodeAccessTokenExpiry };
diff --git a/packages/core/admin/admin/src/utils/tests/jwt.test.ts b/packages/core/admin/admin/src/utils/tests/jwt.test.ts
new file mode 100644
index 0000000000..dbb1886b48
--- /dev/null
+++ b/packages/core/admin/admin/src/utils/tests/jwt.test.ts
@@ -0,0 +1,87 @@
+import { decodeAccessTokenExpiry } from '../jwt';
+
+/**
+ * Build a JWT-shaped string `<header>.<payload>.<signature>` where the
+ * payload is base64url-encoded JSON. The signature is irrelevant — we never
+ * verify it. The encoding intentionally mirrors RFC 7515 base64url:
+ *   - `+` / `/` mapped to `-` / `_`
+ *   - trailing `=` padding stripped
+ *
+ * That last point is the case we particularly want to cover, since unpadded
+ * input is rejected by stricter `atob` implementations.
+ */
+const buildJwt = (payload: Record<string, unknown>): string => {
+  const json = JSON.stringify(payload);
+  const base64 = window.btoa(json);
+  const base64url = base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
+  return `header.${base64url}.signature`;
+};
+
+describe('decodeAccessTokenExpiry', () => {
+  it('returns exp in milliseconds for a well-formed token', () => {
+    const expSeconds = 1_700_000_000;
+    const token = buildJwt({ exp: expSeconds, sub: 'user-1' });
+
+    expect(decodeAccessTokenExpiry(token)).toBe(expSeconds * 1000);
+  });
+
+  it('handles base64url payloads that require `=` padding', () => {
+    /**
+     * The padding requirement depends on the byte length of the JSON payload
+     * mod 3. We try a range of payload sizes so at least one of them lands on
+     * an encoded length that is not a multiple of 4 (i.e. would have been
+     * padded with `=` in standard base64). If padding is not re-added before
+     * `atob`, this test catches it.
+     */
+    for (let i = 0; i < 6; i += 1) {
+      const padding = 'x'.repeat(i);
+      const token = buildJwt({ exp: 1234567890, pad: padding });
+
+      expect(decodeAccessTokenExpiry(token)).toBe(1234567890 * 1000);
+    }
+  });
+
+  it('handles base64url payloads using URL-safe `-` and `_` characters', () => {
+    /**
+     * Forces `+` / `/` into the standard base64 output so the URL-safe
+     * mapping in the decoder (`-` ↔ `+`, `_` ↔ `/`) is exercised. The exact
+     * padding string is one we know produces those characters once base64'd.
+     */
+    const token = buildJwt({ exp: 9_999_999_999, marker: 'a?b>c<d' });
+
+    expect(decodeAccessTokenExpiry(token)).toBe(9_999_999_999 * 1000);
+  });
+
+  it('returns null when the token is empty', () => {
+    expect(decodeAccessTokenExpiry('')).toBeNull();
+  });
+
+  it('returns null when the token has no `.` separator', () => {
+    expect(decodeAccessTokenExpiry('not-a-jwt')).toBeNull();
+  });
+
+  it('returns null when the payload segment is empty', () => {
+    expect(decodeAccessTokenExpiry('header..signature')).toBeNull();
+  });
+
+  it('returns null when the payload is not valid JSON', () => {
+    const garbage = window.btoa('{not json');
+    expect(decodeAccessTokenExpiry(`header.${garbage}.signature`)).toBeNull();
+  });
+
+  it('returns null when the payload has no exp claim', () => {
+    const token = buildJwt({ sub: 'user-1' });
+    expect(decodeAccessTokenExpiry(token)).toBeNull();
+  });
+
+  it('returns null when exp is not a number', () => {
+    const token = buildJwt({ exp: '1700000000' });
+    expect(decodeAccessTokenExpiry(token)).toBeNull();
+  });
+
+  it('returns null for non-string input', () => {
+    expect(decodeAccessTokenExpiry(undefined as unknown as string)).toBeNull();
+    expect(decodeAccessTokenExpiry(null as unknown as string)).toBeNull();
+    expect(decodeAccessTokenExpiry(12345 as unknown as string)).toBeNull();
+  });
+});

```

derrickmehaffy:
> What do you think about doing something like this ? That refactors `clearStateAndLogout` into a testable hook that packages/core/admin/admin/src/features/Auth.tsx can then import?
> 
> ```
> diff --git a/packages/core/admin/admin/src/features/Auth.tsx b/packages/core/admin/admin/src/features/Auth.tsx
> index f68369e569..592c660b91 100644
> --- a/packages/core/admin/admin/src/features/Auth.tsx
> +++ b/packages/core/admin/admin/src/features/Auth.tsx
> @@ -6,6 +6,7 @@ import { Login } from '../../../shared/contracts/authentication';
>  import { createContext } from '../components/Context';
>  import { useTypedDispatch, useTypedSelector } from '../core/store/hooks';
>  import { useStrapiApp } from '../features/StrapiApp';
> +import { useIdleSessionLogout } from '../hooks/useIdleSessionLogout';
>  import { useQueryParams } from '../hooks/useQueryParams';
>  import { login as loginAction, logout as logoutAction, setLocale, setToken } from '../reducer';
>  import { adminApi } from '../services/api';
> @@ -92,28 +93,6 @@ const STORAGE_KEYS = {
>    STATUS: 'isLoggedIn',
>  };
>  
> -/**
> - * Decode the ``exp`` claim of an admin access JWT and return it as milliseconds
> - * since the epoch. Returns ``null`` if the token can't be parsed or has no
> - * numeric ``exp``. We do not verify the signature — the server is the source of
> - * truth; this is purely to schedule a client-side idle timer.
> - */
> -const decodeAccessTokenExpiry = (token: string): number | null => {
> -  const parts = token.split('.');
> -  if (parts.length < 2) {
> -    return null;
> -  }
> -
> -  try {
> -    // base64url → base64
> -    const padded = parts[1].replace(/-/g, '+').replace(/_/g, '/');
> -    const payload = JSON.parse(window.atob(padded));
> -    return typeof payload?.exp === 'number' ? payload.exp * 1000 : null;
> -  } catch {
> -    return null;
> -  }
> -};
> -
>  const AuthProvider = ({
>    children,
>    _defaultPermissions = [],
> @@ -200,46 +179,15 @@ const AuthProvider = ({
>    }, [clearStateAndLogout]);
>  
>    /**
> -   * Proactive idle-session detection.
> -   *
> -   * Schedule a one-shot timer for the access token's ``exp`` claim. While the
> -   * user is active, every API call that hits a 401 transparently refreshes
> -   * the access token (see ``withTokenRefresh`` in getFetchClient.ts), which
> -   * updates the token in Redux and re-runs this effect with the new exp,
> -   * pushing the deadline forward. If no API activity occurs during the
> -   * access token lifespan, the token in Redux never changes, the timer
> -   * fires, and we log the user out — matching the configured server-side
> -   * idle behavior without needing a separate activity tracker.
> -   *
> -   * If the JWT can't be decoded (malformed, missing exp), we skip the
> -   * timer; the active-tab redirect on 401 still covers the symptom.
> +   * Proactive idle-session detection. See ``useIdleSessionLogout`` for the full
> +   * rationale and the UX trade-off (a user actively typing in a form with no
> +   * outgoing API calls will still be logged out at ``exp``).
>     */
> -  React.useEffect(() => {
> -    if (!token || _disableRenewToken) {
> -      return undefined;
> -    }
> -
> -    const expiry = decodeAccessTokenExpiry(token);
> -    if (expiry === null) {
> -      return undefined;
> -    }
> -
> -    // Small buffer past ``exp`` to avoid firing in the same tick the server
> -    // would still consider the token valid (clock skew tolerance).
> -    const SESSION_EXPIRY_BUFFER_MS = 1000;
> -    const msUntilExpiry = expiry - Date.now() + SESSION_EXPIRY_BUFFER_MS;
> -
> -    const timeoutId = window.setTimeout(
> -      () => {
> -        clearStateAndLogout();
> -      },
> -      Math.max(msUntilExpiry, 0)
> -    );
> -
> -    return () => {
> -      window.clearTimeout(timeoutId);
> -    };
> -  }, [token, clearStateAndLogout, _disableRenewToken]);
> +  useIdleSessionLogout({
> +    token,
> +    onExpired: clearStateAndLogout,
> +    disabled: _disableRenewToken,
> +  });
>  
>    React.useEffect(() => {
>      /**
> diff --git a/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts b/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts
> new file mode 100644
> index 0000000000..40c51f03fc
> --- /dev/null
> +++ b/packages/core/admin/admin/src/hooks/tests/useIdleSessionLogout.test.ts
> @@ -0,0 +1,170 @@
> +import { act, renderHook } from '@testing-library/react';
> +
> +import { useIdleSessionLogout, SESSION_EXPIRY_BUFFER_MS } from '../useIdleSessionLogout';
> +
> +/**
> + * Build a JWT-shaped string whose payload encodes ``{ exp: <expSeconds> }``.
> + * We don't care about the header/signature segments here — the decoder
> + * doesn't read them.
> + */
> +const buildJwt = (expSeconds: number): string => {
> +  const json = JSON.stringify({ exp: expSeconds });
> +  const base64url = window.btoa(json).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
> +  return ``header.${base64url}.signature``;
> +};
> +
> +const ACCESS_TOKEN_LIFESPAN_MS = 30_000;
> +const NOW_MS = 1_700_000_000_000;
> +
> +describe('useIdleSessionLogout', () => {
> +  beforeEach(() => {
> +    jest.useFakeTimers({ now: NOW_MS });
> +  });
> +
> +  afterEach(() => {
> +    jest.useRealTimers();
> +  });
> +
> +  it('schedules a one-shot logout at the access token exp', () => {
> +    const onExpired = jest.fn();
> +    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +
> +    renderHook(() => useIdleSessionLogout({ token, onExpired }));
> +
> +    expect(onExpired).not.toHaveBeenCalled();
> +
> +    act(() => {
> +      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
> +    });
> +
> +    expect(onExpired).toHaveBeenCalledTimes(1);
> +  });
> +
> +  it('fires immediately when the token is already past exp', () => {
> +    const onExpired = jest.fn();
> +    const token = buildJwt((NOW_MS - 5_000) / 1000);
> +
> +    renderHook(() => useIdleSessionLogout({ token, onExpired }));
> +
> +    /**
> +     * ``setTimeout(fn, 0)`` doesn't run synchronously, so we still need to
> +     * advance the clock by a tick. The point of the test is that the timer
> +     * is scheduled with a non-negative delay rather than being skipped.
> +     */
> +    act(() => {
> +      jest.advanceTimersByTime(0);
> +    });
> +
> +    expect(onExpired).toHaveBeenCalledTimes(1);
> +  });
> +
> +  it('reschedules to the new exp when the token is refreshed', () => {
> +    const onExpired = jest.fn();
> +
> +    /**
> +     * Initial token: expires at NOW + 30s.
> +     * After 20s of "user activity" we simulate a 401-triggered refresh
> +     * (handled in ``withTokenRefresh``) by passing a new token whose ``exp``
> +     * is later. The hook should clear the original timer and schedule a
> +     * new one against the later ``exp``. If the reschedule logic is broken,
> +     * ``onExpired`` would fire when the original ``exp`` elapses.
> +     */
> +    const initialToken = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +
> +    const { rerender } = renderHook(
> +      ({ token }: { token: string }) => useIdleSessionLogout({ token, onExpired }),
> +      { initialProps: { token: initialToken } }
> +    );
> +
> +    act(() => {
> +      jest.advanceTimersByTime(20_000);
> +    });
> +    expect(onExpired).not.toHaveBeenCalled();
> +
> +    /**
> +     * Refresh: new token expires 30s from "now" (NOW + 20_000 + 30_000).
> +     * Original ``exp`` would have been NOW + 30_000, i.e. 10s away. After
> +     * rerender, the reschedule should push the deadline out by 20s beyond
> +     * the original.
> +     */
> +    const refreshedToken = buildJwt((NOW_MS + 20_000 + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +    rerender({ token: refreshedToken });
> +
> +    /**
> +     * Advance past the *original* exp deadline. If reschedule is broken the
> +     * old timer would fire here.
> +     */
> +    act(() => {
> +      jest.advanceTimersByTime(15_000);
> +    });
> +    expect(onExpired).not.toHaveBeenCalled();
> +
> +    /**
> +     * Advance the rest of the way to the new exp + buffer. Total elapsed at
> +     * this point is 20_000 + 15_000 + 15_000 + buffer = 50s + buffer, which
> +     * is exactly 30s past "now" from the refresh point.
> +     */
> +    act(() => {
> +      jest.advanceTimersByTime(15_000 + SESSION_EXPIRY_BUFFER_MS);
> +    });
> +    expect(onExpired).toHaveBeenCalledTimes(1);
> +  });
> +
> +  it('cancels the pending timer when the token is cleared (logout)', () => {
> +    const onExpired = jest.fn();
> +    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +
> +    const { rerender } = renderHook(
> +      ({ token }: { token: string | null }) => useIdleSessionLogout({ token, onExpired }),
> +      { initialProps: { token } as { token: string | null } }
> +    );
> +
> +    rerender({ token: null });
> +
> +    act(() => {
> +      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
> +    });
> +
> +    expect(onExpired).not.toHaveBeenCalled();
> +  });
> +
> +  it('does not schedule a timer when disabled', () => {
> +    const onExpired = jest.fn();
> +    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +
> +    renderHook(() => useIdleSessionLogout({ token, onExpired, disabled: true }));
> +
> +    act(() => {
> +      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
> +    });
> +
> +    expect(onExpired).not.toHaveBeenCalled();
> +  });
> +
> +  it('does not schedule a timer when the token is malformed', () => {
> +    const onExpired = jest.fn();
> +
> +    renderHook(() => useIdleSessionLogout({ token: 'not-a-jwt', onExpired }));
> +
> +    act(() => {
> +      jest.advanceTimersByTime(60_000);
> +    });
> +
> +    expect(onExpired).not.toHaveBeenCalled();
> +  });
> +
> +  it('cancels the pending timer on unmount', () => {
> +    const onExpired = jest.fn();
> +    const token = buildJwt((NOW_MS + ACCESS_TOKEN_LIFESPAN_MS) / 1000);
> +
> +    const { unmount } = renderHook(() => useIdleSessionLogout({ token, onExpired }));
> +
> +    unmount();
> +
> +    act(() => {
> +      jest.advanceTimersByTime(ACCESS_TOKEN_LIFESPAN_MS + SESSION_EXPIRY_BUFFER_MS);
> +    });
> +
> +    expect(onExpired).not.toHaveBeenCalled();
> +  });
> +});
> diff --git a/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts b/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts
> new file mode 100644
> index 0000000000..45315db399
> --- /dev/null
> +++ b/packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts
> @@ -0,0 +1,68 @@
> +import * as React from 'react';
> +
> +import { decodeAccessTokenExpiry } from '../utils/jwt';
> +
> +const SESSION_EXPIRY_BUFFER_MS = 1000;
> +
> +interface UseIdleSessionLogoutOptions {
> +  /**
> +   * The current admin access JWT (from Redux), or ``null`` when logged out.
> +   */
> +  token: string | null;
> +  /**
> +   * Called when the timer fires. Typically clears local auth state and
> +   * navigates to /auth/login.
> +   */
> +  onExpired: () => void;
> +  /**
> +   * Escape hatch for tests / dev environments where we don't want the
> +   * automatic logout behavior. Mirrors the existing ``_disableRenewToken``
> +   * prop on ``<AuthProvider>``.
> +   */
> +  disabled?: boolean;
> +}
> +
> +/**
> + * Schedule a one-shot logout when the access token's ``exp`` elapses.
> + *
> + * The hook re-runs whenever ``token`` changes. While the user is active, every
> + * API call that hits a 401 transparently refreshes the access token (see
> + * ``withTokenRefresh`` in ``getFetchClient.ts``), which dispatches ``setToken`` and
> + * causes this effect to re-arm with the new, later ``exp``.
> + *
> + * If the JWT can't be decoded (malformed, missing ``exp``), the timer is
> + * skipped silently — the active-tab redirect on 401 still covers the
> + * symptom on the next user-initiated request.
> + */
> +const useIdleSessionLogout = ({
> +  token,
> +  onExpired,
> +  disabled = false,
> +}: UseIdleSessionLogoutOptions): void => {
> +  React.useEffect(() => {
> +    if (!token || disabled) {
> +      return undefined;
> +    }
> +
> +    const expiry = decodeAccessTokenExpiry(token);
> +    if (expiry === null) {
> +      return undefined;
> +    }
> +
> +    const msUntilExpiry = expiry - Date.now() + SESSION_EXPIRY_BUFFER_MS;
> +
> +    const timeoutId = window.setTimeout(
> +      () => {
> +        onExpired();
> +      },
> +      Math.max(msUntilExpiry, 0)
> +    );
> +
> +    return () => {
> +      window.clearTimeout(timeoutId);
> +    };
> +  }, [token, onExpired, disabled]);
> +};
> +
> +export { useIdleSessionLogout, SESSION_EXPIRY_BUFFER_MS };
> +export type { UseIdleSessionLogoutOptions };
> diff --git a/packages/core/admin/admin/src/utils/jwt.ts b/packages/core/admin/admin/src/utils/jwt.ts
> new file mode 100644
> index 0000000000..0442189e51
> --- /dev/null
> +++ b/packages/core/admin/admin/src/utils/jwt.ts
> @@ -0,0 +1,41 @@
> +/**
> + * Decode the ``exp`` claim of an admin access JWT and return it as milliseconds
> + * since the epoch. Returns ``null`` if the token can't be parsed or has no
> + * numeric ``exp``.
> + *
> + * The signature is intentionally not verified — the server is the source of
> + * truth for token validity. This is purely so the client can schedule a
> + * one-shot timer to react to access-token expiry without polling.
> + *
> + * Handles ``base64url`` payloads (RFC 7515) by mapping the URL-safe alphabet
> + * back to standard base64 and re-adding the ``=`` padding that ``base64url``
> + * strips. Some browsers' ``atob`` implementations reject unpadded input.
> + */
> +const decodeAccessTokenExpiry = (token: string): number | null => {
> +  if (typeof token !== 'string') {
> +    return null;
> +  }
> +
> +  const parts = token.split('.');
> +  if (parts.length < 2 || parts[1].length === 0) {
> +    return null;
> +  }
> +
> +  try {
> +    // base64url → base64
> +    const urlDecoded = parts[1].replace(/-/g, '+').replace(/_/g, '/');
> +
> +    // Restore ``=`` padding to a multiple of 4 chars; base64url omits it.
> +    const remainder = urlDecoded.length % 4;
> +    const padded = remainder === 0 ? urlDecoded : urlDecoded + '='.repeat(4 - remainder);
> +
> +    const payload = JSON.parse(window.atob(padded));
> +    return typeof payload?.exp === 'number' && Number.isFinite(payload.exp)
> +      ? payload.exp * 1000
> +      : null;
> +  } catch {
> +    return null;
> +  }
> +};
> +
> +export { decodeAccessTokenExpiry };
> diff --git a/packages/core/admin/admin/src/utils/tests/jwt.test.ts b/packages/core/admin/admin/src/utils/tests/jwt.test.ts
> new file mode 100644
> index 0000000000..dbb1886b48
> --- /dev/null
> +++ b/packages/core/admin/admin/src/utils/tests/jwt.test.ts
> @@ -0,0 +1,87 @@
> +import { decodeAccessTokenExpiry } from '../jwt';
> +
> +/**
> + * Build a JWT-shaped string ``<header>.<payload>.<signature>`` where the
> + * payload is base64url-encoded JSON. The signature is irrelevant — we never
> + * verify it. The encoding intentionally mirrors RFC 7515 base64url:
> + *   - ``+`` / ``/`` mapped to ``-`` / ``_``
> + *   - trailing ``=`` padding stripped
> + *
> + * That last point is the case we particularly want to cover, since unpadded
> + * input is rejected by stricter ``atob`` implementations.
> + */
> +const buildJwt = (payload: Record<string, unknown>): string => {
> +  const json = JSON.stringify(payload);
> +  const base64 = window.btoa(json);
> +  const base64url = base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
> +  return ``header.${base64url}.signature``;
> +};
> +
> +describe('decodeAccessTokenExpiry', () => {
> +  it('returns exp in milliseconds for a well-formed token', () => {
> +    const expSeconds = 1_700_000_000;
> +    const token = buildJwt({ exp: expSeconds, sub: 'user-1' });
> +
> +    expect(decodeAccessTokenExpiry(token)).toBe(expSeconds * 1000);
> +  });
> +
> +  it('handles base64url payloads that require ``=`` padding', () => {
> +    /**
> +     * The padding requirement depends on the byte length of the JSON payload
> +     * mod 3. We try a range of payload sizes so at least one of them lands on
> +     * an encoded length that is not a multiple of 4 (i.e. would have been
> +     * padded with ``=`` in standard base64). If padding is not re-added before
> +     * ``atob``, this test catches it.
> +     */
> +    for (let i = 0; i < 6; i += 1) {
> +      const padding = 'x'.repeat(i);
> +      const token = buildJwt({ exp: 1234567890, pad: padding });
> +
> +      expect(decodeAccessTokenExpiry(token)).toBe(1234567890 * 1000);
> +    }
> +  });
> +
> +  it('handles base64url payloads using URL-safe ``-`` and ``_`` characters', () => {
> +    /**
> +     * Forces ``+`` / ``/`` into the standard base64 output so the URL-safe
> +     * mapping in the decoder (``-`` ↔ ``+``, ``_`` ↔ ``/``) is exercised. The exact
> +     * padding string is one we know produces those characters once base64'd.
> +     */
> +    const token = buildJwt({ exp: 9_999_999_999, marker: 'a?b>c<d' });
> +
> +    expect(decodeAccessTokenExpiry(token)).toBe(9_999_999_999 * 1000);
> +  });
> +
> +  it('returns null when the token is empty', () => {
> +    expect(decodeAccessTokenExpiry('')).toBeNull();
> +  });
> +
> +  it('returns null when the token has no ``.`` separator', () => {
> +    expect(decodeAccessTokenExpiry('not-a-jwt')).toBeNull();
> +  });
> +
> +  it('returns null when the payload segment is empty', () => {
> +    expect(decodeAccessTokenExpiry('header..signature')).toBeNull();
> +  });
> +
> +  it('returns null when the payload is not valid JSON', () => {
> +    const garbage = window.btoa('{not json');
> +    expect(decodeAccessTokenExpiry(``header.${garbage}.signature``)).toBeNull();
> +  });
> +
> +  it('returns null when the payload has no exp claim', () => {
> +    const token = buildJwt({ sub: 'user-1' });
> +    expect(decodeAccessTokenExpiry(token)).toBeNull();
> +  });
> +
> +  it('returns null when exp is not a number', () => {
> +    const token = buildJwt({ exp: '1700000000' });
> +    expect(decodeAccessTokenExpiry(token)).toBeNull();
> +  });
> +
> +  it('returns null for non-string input', () => {
> +    expect(decodeAccessTokenExpiry(undefined as unknown as string)).toBeNull();
> +    expect(decodeAccessTokenExpiry(null as unknown as string)).toBeNull();
> +    expect(decodeAccessTokenExpiry(12345 as unknown as string)).toBeNull();
> +  });
> +});
> ```

Fine by me, so long as the redirect to login page works :P

jhoward1994 (approved):

PR #26167, opened 2026-04-30
fix(content-manager): preserve i18n locale on navigation and guard component schema race condition

### LINKED ISSUE #26114: Content Manager flicker/double-paint when navigating between localized single types via Main Nav
### Node Version

22 LTS

### Package Manager

pnpm

### Package Manager Version

10.x

### Strapi Version

5.43.0  

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Hi Admin, I have face a bug when Clicking between localized single types in the Content Manager Main Nav, here is what Claude code have me to find the bug. 

  Bug report

  Describe the bug

  Clicking between localized single types in the Content Manager Main Nav produces a visible flicker: the Edit view paints once with
  locale=undefined, then immediately re-paints after the i18n plugin corrects the URL to include ?plugins[i18n][locale]=<defaultLocale>.

  On each sidebar click, two renders happen back-to-back against two different sets of document data, which causes a perceptible flash and
   extra work for every click.                                                                                                                  
                  
  Root cause                                                                                                                              
                  
  Two files cooperate to produce this:                                                                                                    
                  
  1. @strapi/content-manager/admin/src/components/LeftMenu.tsx — the sidebar link is built with only pathname, no search param:           
   
  <SubNav.Link                                                                                                                            
    to={{ pathname: link.to }}   // no search param attached
    label={link.title}                                                                                                                    
    key={link.uid}
  />                                                                                                                                      
                                                                                                                                          
  2. @strapi/plugin-i18n/admin/src/components/CMHeaderActions.tsx — a corrective useEffect pushes the locale into the URL after the first 
  render:                                                                                                                                 
                                                                                                                                          
  React.useEffect(() => {
    if (!Array.isArray(locales) || !hasI18n) return;
                                                                                                                                          
    const doesLocaleExist = locales.find((loc) => loc.code === currentDesiredLocale);                                                     
    const defaultLocale = locales.find((locale) => locale.isDefault);                                                                     
                                                                                                                                          
    if (!doesLocaleExist && defaultLocale?.code) {
      handleSelect(defaultLocale.code);  // setQuery(..., 'push', true)                                                                   
    }                                                                                                                                     
  }, [handleSelect, hasI18n, locales, currentDesiredLocale]);
                                                                                                                                          
  The LeftMenu link should either (a) read the currently-active locale from the URL/query-params context and append                       
  ?plugins[i18n][locale]=<currentLocale> to the to, or (b) read the default locale from the useGetLocalesQuery data and append that.      
  Either way, the "naked URL → corrective redirect" cycle should not happen for a pure sidebar click.                                     
                  
    System                                                                                                                                  
  
  - Strapi version: 5.43.0                                                                                                                
  - Operating system: macOS (Darwin 25.3.0) — reproducible on Linux too
  - Database: PostgreSQL 16                                                                                                               
  - Node version: 22 LTS
  - NPM version: pnpm 10.x                                                                                                                
  - Yarn version: n/a
  - Browser: Chrome / Firefox (framework-agnostic — the bug is in the admin bundle)                                                       
                                                                                                                                          
  Additional context
                                                                                                                                          
  The behavior can be confirmed in DevTools → Network by watching the GET /content-manager/single-types/api::*.* request fire twice on    
  each sidebar click: once without locale, then a corrective fetch with locale=<default>. Setting STRAPI_PLUGIN_I18N_INIT_LOCALE_CODE does
   not fix this — that env var only seeds the initial DB default locale, it doesn't inject locale into admin navigation URLs.  

### Steps to Reproduce

 1. Enable i18n and create at least 2 localized single types (e.g. about-page, pricing-page), both with pluginOptions.i18n.localized:    
  true.
  2. Publish content in the default locale for both.                                                                                      
  3. Open /admin and click between the two single types in the Content Manager sidebar.                                                   
  4. Observe: each click causes a brief flash as the view paints twice.                                                                   
                                                                                                                                          
  Reproducible in a clean npx create-strapi-app@latest with the i18n plugin enabled.     

### Expected Behavior

  One paint per navigation. The locale search param should already be present on the sidebar link, so the Edit view never mounts with     
  locale=undefined.

### Logs

```shell

```

### Code Snippets

Suggested fix

In LeftMenu.tsx, attach a search param to each link so the Edit view mounts with a locale already in query.plugins.i18n.locale.         
  Something like:
                                                                                                                                          
  const [{ query }] = useQueryParams();
  const currentLocale = query?.plugins?.i18n?.locale;                                                                                     
   
  // in render:                                                                                                                           
  <SubNav.Link    
    to={{                                                                                                                                 
      pathname: link.to,
      search: currentLocale ? `?plugins[i18n][locale]=${currentLocale}` : '',
    }}                                                                                                                                    
    label={link.title}
    key={link.uid}                                                                                                                        
  />              

  When currentLocale is undefined (cold load), the existing useEffect in CMHeaderActions still runs and fills in the default locale — but 
  this only happens once per session instead of once per sidebar click.
                                                                                                                                          
  Code snippets                                                                                                                           
  
  Referenced files (paths in packages/core/):                                                                                             
  - packages/core/content-manager/admin/src/components/LeftMenu.tsx
  - packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx                                                                        
                  


### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

AlexNbl27:
## What

Fixes #26114 — two related bugs that together cause a `TypeError: Cannot read properties of undefined (reading 'attributes')` when navigating between localized content types in the admin panel.

## Root causes

**Bug 1 — LeftMenu strips the i18n locale from the URL on navigation**

`SubNav.Link` only received `pathname` in its `to` prop. The i18n plugin's `useEffect` was responsible for pushing the locale back into the URL *after* mount, causing a render with `locale=undefined`.

**Bug 2 — `formatEditLayout` crashes when a component schema is not yet cached**

`useDocumentLayout` calls `useDocument({ skip: true })` to get component schemas from the RTK Query cache. During navigation, `data.components` (configuration, from `useGetContentTypeConfigurationQuery`) can resolve before the new content type's schemas are in cache. Accessing `components[uid].attributes` on an undefined entry throws the TypeError. The same pattern exists in `ComponentConfigurationPage.tsx`.

## Changes

- **`LeftMenu.tsx`** — read current locale from `useLocation().search` and forward it as `search` on each `SubNav.Link`, so the URL is already correct on first render
- **`useDocumentLayout.ts`** — guard `components[uid]` before accessing `.attributes`; the `useMemo` re-runs once schemas are loaded
- **`ComponentConfigurationPage.tsx`** — same guard for the identical vulnerable pattern

## How to test

1. Enable i18n on 2+ single types (e.g. `shop-configuration` and `lookbook`)
2. Open one in the admin panel (`?plugins[i18n][locale]=fr-FR` in URL)
3. Click the other in the left menu sidebar
4. **Before:** TypeError crash, manual refresh required
5. **After:** smooth navigation, locale preserved in URL


AlexNbl27:
Been hitting this in production on **5.45.0**. After digging into it, the crash in `formatEditLayout` isn't directly caused by the missing locale — it's a race condition: `data.components` (configuration) resolves before the component schemas are populated in the RTK Query cache, because `useDocument` is called with `skip: true`. The missing locale just makes it happen consistently on every navigation. Both issues fixed in my PR, running in prod with no regressions.

innerdvations (dismissed):
Works for me, thanks for the PR!

innerdvations (approved):

PR #26171, opened 2026-04-30
fix(review-workflows): add assignee and review stage to list view filters

### LINKED ISSUE #21790: Filters for Assignee and Review Stage no longer present

- Node.js version: 20.15.0
- Yarn version: 1.22.22
- Strapi version: 5.0.4
- Database: sqlite
- Operating system: darwin arm64
- Is your project Javascript or Typescript:

### Describe the bug

In Strapi 4, when a content type had review workflow associated with it, filters for Review Stage and Assignee were available. In Strapi 5, they are not available. 

### Expected behavior

Filter for Review Stage and Assignee is available.

###Screenshot

https://github.com/user-attachments/assets/43a95e0e-9f32-4fcf-957e-917df62abf39

### Code snippets

This was tested on LaunchPad, using an EE license, in the Article content-type

Adzouz:
### What does it do?

Adds **Assignee** and **Review stage** to the list of filters in the Content Manager list view (only when Review Workflows is enabled on the content type).

Screenshot 2026-04-30 at 11 15 18

Also exposes a new admin hook `Admin/CM/pages/ListView/inject-in-filters` so plugins can inject their own filters into the list view (mirroring the existing column-injection hook).

The Assignee filter supports `is`, `is not`, `is null`, `is not null` operators. The admin users fetch is deferred until the combobox opens to avoid an extra request when the user only clicks the Filters button.

### Why is it needed?

Filtering by assignee/stage is a feature expected with Review Workflows (EE). Previously only columns were injected — filters were missing.

### How to test it?

ℹ️ EE should be enabled and review workflows configured to a content type
1. Go to any content type setup with review workflows
2. Assign entries to users and set them to different stages
3. Go to the list view and open the filters

You should see two new filters:
- **Assignee** — pick `is`, `is not`, `is null`, `is not null`. Verify the chip displays the user display name (not the id).
- **Review stage** — verify the list filters by the picked stage.

Sanity checks:
- Open the Filters popover but do not pick the Assignee filter → no admin users request fires (Network tab).
- Switching between filters does not re-fetch users on every click (cached).

### Related issue(s)/PR(s)

Fixes #21790

🚀

### DIFF
--- a/packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
+++ b/packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
@@ -10,11 +10,9 @@ interface AssigneeFilterProps extends Pick<ComboboxProps, 'value' | 'onChange'> 
 
 const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
   const [page, setPage] = React.useState(1);
+  const [hasOpened, setHasOpened] = React.useState(false);
   const { formatMessage } = useIntl();
-  const { data, isLoading } = useAdminUsers({
-    page,
-  });
-  const users = data?.users || [];
+  const { data, isLoading } = useAdminUsers({ page }, { skip: !hasOpened });
 
   const field = useField(name);
 
@@ -22,6 +20,21 @@ const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
     setPage(1);
     field.onChange(name, value);
   };
+
+  const handleOpenChange = (isOpen?: boolean) => {
+    if (isOpen && !hasOpened) {
+      setHasOpened(true);
+    }
+  };
+
+  const options = React.useMemo(
+    () =>
+      (data?.users ?? []).map((user) => ({
+        id: user.id,
+        label: getDisplayName(user),
+      })),
+    [data?.users]
+  );
 
   return (
     <Combobox
@@ -31,16 +44,15 @@ const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
         defaultMessage: 'Search and select an user to filter',
       })}
       onChange={handleChange}
+      onOpenChange={handleOpenChange}
       loading={isLoading}
       onLoadMore={() => setPage((prev) => prev + 1)}
     >
-      {users.map((user) => {
-        return (
-          <ComboboxOption key={user.id} value={user.id.toString()}>
-            {getDisplayName(user)}
-          </ComboboxOption>
-        );
-      })}
+      {options.map((option) => (
+        <ComboboxOption key={option.id} value={option.id.toString()}>
+          {option.label}
+        </ComboboxOption>
+      ))}
     </Combobox>
   );
 };
--- a/packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
+++ b/packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
@@ -8,16 +8,18 @@ import { getDisplayName } from '../../../../utils/users';
 
 interface AssigneeFilterProps extends Pick<ComboboxProps, 'value' | 'onChange'> {}
 
+const PAGE_SIZE = 10;
+
 const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
-  const [page, setPage] = React.useState(1);
+  const [pageSize, setPageSize] = React.useState(PAGE_SIZE);
   const [hasOpened, setHasOpened] = React.useState(false);
   const { formatMessage } = useIntl();
   const field = useField(name);
   const shouldFetch = hasOpened || Boolean(field.value);
-  const { data, isLoading } = useAdminUsers({ page }, { skip: !shouldFetch });
+  const { data, isLoading } = useAdminUsers({ pageSize }, { skip: !shouldFetch });
 
   const handleChange = (value?: string) => {
-    setPage(1);
+    setPageSize(PAGE_SIZE);
     field.onChange(name, value);
   };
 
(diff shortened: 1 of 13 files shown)


mathildeleg (commented):
Looks good but a few comments:
1. Is there a way to have the filters (in the list) be lowercase, like the rest of the filters?
2. I know you didn't touch this file, but I think you can now de-skip the `AssigneeFilter.test.tsx` test since filters are now supported.
3. I'm kind of worried about perf because that `AssigneeFilter.tsx` file is called every time you click on "Filters" or any of the filters' label or values. Just wondering if that could be improved/avoided?
Also, you've added the "review stage" filter so could you update the PR title and description to make sure we don't forget to test that during QA?
Thanks 😊

Edit: Forgot to say, is it possible to have a filter for "Assignee" or "Review stage" to be different from "is/is not", for example "is null/is not null"? Cause sometimes it can happen that review workflows are enabled for a content type but the document(s) don't have those fields set, thus it might be interesting to be able to filter through those too. These are a bit out of scope I suppose, but just curious for the answer and depending, we can create a ticket for it! :)

packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
```diff
@@ -23,6 +21,21 @@ const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
     field.onChange(name, value);
   };
 
+  const handleOpenChange = (isOpen?: boolean) => {
+    if (isOpen && !hasOpened) {
+      setHasOpened(true);
+    }
+  };
+
+  const options = React.useMemo(
```
mathildeleg, line 42: I think this may be breaking the value selected when editing a filter, rendering nothing instead of the selected value: 
https://github.com/user-attachments/assets/57d6ea6e-31bd-4658-985f-f664eb783943

mathildeleg (commented):
Thanks for answering my feedback, I just have one last comment 🙏🏻

packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
Adzouz: Good catch, done ✅

packages/core/review-workflows/admin/src/translations/en.json
```diff
@@ -13,5 +13,7 @@
   "settings.page.purchase.perks2": "Manage user permissions",
   "settings.page.purchase.perks3": "Support for webhooks",
   "widget.assigned.title": "Assigned to me",
-  "widget.assigned.no-data": "No entries"
+  "widget.assigned.no-data": "No entries",
+  "containers.list.table-headers.reviewWorkflows.stage": "review stage",
+  "containers.list.table-headers.reviewWorkflows.assignee": "assignee"
```
jhoward1994, line 18:
```suggestion
  "containers.list.table-headers.reviewWorkflows.stage": "Review stage",
  "containers.list.table-headers.reviewWorkflows.assignee": "Assignee"
```

jhoward1994: Sorry looking at other changes I think I might have missed something here and theres a reason they're not capitalised ?

jhoward1994:
Maybe we could do this at the top of `packages/core/content-manager/admin/src/constants/hooks.ts`

& import the types where it's needed? 
packages/core/review-workflows/admin/src/utils/cm-hooks.ts
packages/core/content-manager/admin/src/pages/ListView/components/Filters.tsx

```
import type { Filters } from '@strapi/admin/strapi-admin';
import type { MessageDescriptor } from 'react-intl';

/**
 * Shape of a filter passed through the `INJECT_LIST_VIEW_FILTERS` hook.
 *
 * Plugin-injected filters may use react-intl `MessageDescriptor` labels so
 * they carry their own translations. The list view consumer resolves them to
 * plain strings (which is what the underlying `Filters.Filter` contract
 * expects) before rendering.
 */
export type InjectableListViewFilter = Omit<Filters.Filter, 'label' | 'operators'> & {
  label: string | MessageDescriptor;
  operators?: Array<{ value: string; label: string | MessageDescriptor }>;
};

```

packages/core/review-workflows/admin/src/translations/en.json
Adzouz: I don't know why haha but the filters are not capitalized 🤷‍♂️ So it was to unify with the rest
Screenshot 2026-05-06 at 15 09 42

Adzouz:
> Maybe we could do this at the top of `packages/core/content-manager/admin/src/constants/hooks.ts`
> 
> & import the types where it's needed? packages/core/review-workflows/admin/src/utils/cm-hooks.ts packages/core/content-manager/admin/src/pages/ListView/components/Filters.tsx
> 
> ```
> import type { Filters } from '@strapi/admin/strapi-admin';
> import type { MessageDescriptor } from 'react-intl';
> 
> /**
>  * Shape of a filter passed through the ``INJECT_LIST_VIEW_FILTERS`` hook.
>  *
>  * Plugin-injected filters may use react-intl ``MessageDescriptor`` labels so
>  * they carry their own translations. The list view consumer resolves them to
>  * plain strings (which is what the underlying ``Filters.Filter`` contract
>  * expects) before rendering.
>  */
> export type InjectableListViewFilter = Omit<Filters.Filter, 'label' | 'operators'> & {
>   label: string | MessageDescriptor;
>   operators?: Array<{ value: string; label: string | MessageDescriptor }>;
> };
> ```

Good idea since it could be reused anywhere now, it makes sense to centralize and export the type from the content manager package 💪
It's done, thanks 🙏

packages/core/review-workflows/admin/src/routes/content-manager/model/components/AssigneeFilter.tsx
```diff
@@ -31,16 +44,15 @@ const AssigneeFilter = ({ name }: Filters.ValueInputProps) => {
         defaultMessage: 'Search and select an user to filter',
       })}
       onChange={handleChange}
+      onOpenChange={handleOpenChange}
       loading={isLoading}
       onLoadMore={() => setPage((prev) => prev + 1)}
```
mathildeleg: Not sure why exactly but this only fetches the first 10 assignees, never fetches more than that

mathildeleg (commented):
One small comment (which I think was an existing bug but might as well fix it now 😇)
But otherwise LGTM!

Adzouz:
Since @mathildeleg is on vacations, would you mind taking over the review @jhoward1994? 🙏

jhoward1994 (approved):

PR #26173, opened 2026-04-30
fix(eslint): set tsconfigRootDir and ignore .eslintrc.js in tsconfig

innerdvations:
### What does it do?

Adds `parserOptions.tsconfigRootDir` for typed ESLint, converts server JSON ESLint configs to `.js` where needed, and ignores `**/.eslintrc.js` in the shared backend TS config so those files are not type-linted.

### Why is it needed?

Fixes `tsconfig.eslint.json` resolution when ESLint runs from the repo root (e.g. VS Code) and avoids `yarn lint` failing because `.eslintrc.js` is outside the TS program.

### How to test it?

Open a package `.ts` file in the editor (no `tsconfig.eslint.json` parse error), run `yarn eslint <path>` from the repo root, then `yarn lint` / `yarn lint:fix`.

### Related issue(s)/PR(s)

—


packages/admin-test-utils/.eslintrc.js
```diff
@@ -3,5 +3,6 @@ module.exports = {
   extends: ['custom/back/typescript'],
   parserOptions: {
     project: ['./tsconfig.json'],
+    tsconfigRootDir: __dirname,
```
nclsndr, line 6: To be added to the list of changes when we ESM the monorepo

packages/plugins/i18n/server/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
Bassel17, line 1: I see a local rule has been removed for this one
```
"rules": {
    "import/no-cycle": "warn"
  }
```
was this intentional ?

packages/core/upload/server/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
Bassel17, line 1: same for this one

Bassel17 (commented):
Looks good, just left a question about the removed local rules
seeing that the CI is passing they might not be important

Bassel17 (commented):
Looks good, just left a question about the removed local rules
seeing that the CI is passing they might not be important

Bassel17 (commented):
Looks good, just left a question about the removed local rules
seeing that the CI is passing they might not be important

Bassel17 (commented):
Looks good, just left a question about the removed local rules
seeing that the CI is passing they might not be important

PR #26174, opened 2026-04-30
feat(*): introduce core strapi.ai.mcp API

nclsndr:
### What does it do?

Introduces `strapi.ai.mcp` — a built-in MCP (Model Context Protocol) server embedded in the Strapi core.

- Adds `McpService` under `strapi.ai.mcp` with `registerTool()`, `registerPrompt()`, `registerResource()`, `start()`, and `stop()` lifecycle methods
- Implements Streamable HTTP transport (`POST /mcp`, `GET /mcp`, `DELETE /mcp`) with JSON-RPC 2.0 compliance
- Session management (`McpSessionManager`) with configurable idle timeout, max sessions, and periodic cleanup
- Capability definition registries (`McpCapabilityDefinitionRegistry`) for tools, prompts, and resources — registration is locked once the server starts
- `McpConfiguration` reads from `server.mcp.*` config keys (enabled, sessionIdleTimeoutMs, maxSessions, cleanupIntervalMs, requestTimeoutMs)
- A built-in `log` tool registered out of the box
- Full TypeScript types published under `@strapi/types` (`Modules.MCP.*`)
- MCP is **dev-mode only** (`autoReload: true`) until auth is implemented

### Why is it needed?

AI coding agents (Cursor, Copilot, Claude, etc.) use the MCP protocol to interact with external services. Exposing a native MCP server in Strapi allows AI tools to introspect and manipulate a running Strapi instance — content types, entries, config — directly from the IDE, without requiring a third-party bridge.

### How to test it?

1. Enable MCP in `config/server.js`:
   ```js
   mcp: { enabled: true }
   ```
2. Start Strapi in dev mode (`yarn develop`)
3. Connect any MCP client (e.g. Cursor, Claude Desktop) to `http://localhost:1337/mcp`
4. Verify the `log` tool is available and callable
5. Register a custom tool from a plugin's `register()` hook via `strapi.ai.mcp.registerTool(...)` and verify it appears in the client

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/.github/workflows/adminBundleSize.yml
+++ b/.github/workflows/adminBundleSize.yml
@@ -30,11 +30,12 @@ jobs:
       - name: Monorepo install
         uses: ./.github/actions/yarn-nm-install
 
-      - name: Monorepo build
-        uses: ./.github/actions/run-build
+      - name: Create install-and-build script
+        run: printf '#!/bin/sh\nset -e\nyarn install\nyarn build\n' > /tmp/install-and-build.sh
 
       - uses: preactjs/compressed-size-action@v2
         with:
+          install-script: 'sh /tmp/install-and-build.sh'
           build-script: 'build:size'
           pattern: '**/build/**/*.{js,css,html,svg}'
           strip-hash: "-([-\\w]{8})(\\.\\w+)$"
--- a/examples/getstarted/config/server.js
+++ b/examples/getstarted/config/server.js
@@ -41,4 +41,7 @@ module.exports = ({ env }) => ({
       // enabled: false,
     },
   },
+  mcp: {
+    enabled: true,
+  },
 });
--- a/packages/core/core/src/services/mcp/handlers/handleDelete.ts
+++ b/packages/core/core/src/services/mcp/handlers/handleDelete.ts
@@ -1,0 +1,41 @@
+import type { Core } from '@strapi/types';
+import { extractSessionId } from '../internal/extractSessionId';
+import { sendJsonRpcError } from '../utils/sendJsonRpcError';
+import { withTimeout } from '../utils/withTimeout';
+import type { McpHandlerDependencies } from './types';
+
+export const createDeleteHandler = (deps: McpHandlerDependencies): Core.MiddlewareHandler => {
+  const { strapi, sessionManager, config } = deps;
+
+  return async (ctx) => {
+    const req = ctx.req;
+    const res = ctx.res;
+    const sessionId = extractSessionId(req);
+
+    if (sessionId === undefined) {
+      sendJsonRpcError(res, 'SESSION_REQUIRED');
+      return;
+    }
+
+    const session = sessionManager.get(sessionId);
+    if (session === undefined) {
+      sendJsonRpcError(res, 'INVALID_SESSION');
+      return;
+    }
+
+    try {
+      await withTimeout(
+        session.transport.handleRequest(req, res, null),
+        config.requestTimeoutMs,
+        'transport.handleRequest'
+      );
+    } catch (error) {
+      strapi.log.error('[MCP] Error handling DELETE request', {
+        error: error instanceof Error ? error.message : 'Unknown error',
+        stack: error instanceof Error ? error.stack : undefined,
+      });
+
+      sendJsonRpcError(res, 'INTERNAL_ERROR');
+    }
+  };
+};
--- a/packages/core/core/src/services/mcp/handlers/handleGet.ts
+++ b/packages/core/core/src/services/mcp/handlers/handleGet.ts
@@ -1,0 +1,39 @@
+import type { Core } from '@strapi/types';
+import { extractSessionId } from '../internal/extractSessionId';
+import { sendJsonRpcError } from '../utils/sendJsonRpcError';
+import type { McpHandlerDependencies } from './types';
+
+export const createGetHandler = (deps: McpHandlerDependencies): Core.MiddlewareHandler => {
+  const { strapi, sessionManager } = deps;
+
+  return async (ctx) => {
+    const req = ctx.req;
+    const res = ctx.res;
+    const sessionId = extractSessionId(req);
+
+    if (sessionId === undefined) {
+      sendJsonRpcError(res, 'SESSION_REQUIRED');
+      return;
+    }
+
+    const session = sessionManager.get(sessionId);
+    if (session === undefined) {
+      sendJsonRpcError(res, 'INVALID_SESSION');
+      return;
+    }
+
+    // GET requests establish long-lived SSE streams; transport manages their lifecycle.
+    session.updateActivity();
+
+    try {
+      await session.transport.handleRequest(req, res, null);
+    } catch (error) {
+      strapi.log.error('[MCP] Error handling GET request', {
+        error: error instanceof Error ? error.message : 'Unknown error',
+        stack: error instanceof Error ? error.stack : undefined,
+      });
+
+      sendJsonRpcError(res, 'INTERNAL_ERROR');
+    }
+  };
+};
--- a/packages/core/core/src/services/mcp/internal/McpCapabilityDefinitionRegistry.ts
+++ b/packages/core/core/src/services/mcp/internal/McpCapabilityDefinitionRegistry.ts
@@ -1,0 +1,40 @@
+import type { Modules } from '@strapi/types';
+
+export class McpCapabilityDefinitionRegistry<
+  CapabilityName extends 'tool' | 'prompt' | 'resource',
+  Definition extends Modules.MCP.McpCapabilityDefinition,
+> {
+  capability: CapabilityName;
+
+  #definitions = new Map<string, Definition>();
+
+  constructor(capability: CapabilityName) {
+    this.capability = capability;
+  }
+
+  define(definition: Definition) {
+    const existing = this.#definitions.get(definition.name);
+    if (existing !== undefined) {
+      throw new Error(
+        `[MCP] ${this.capability} with name "${definition.name}" is already registered. Names must be unique.`
+      );
+    }
+    this.#definitions.set(definition.name, definition);
+  }
+
+  get(name: string): Definition | undefined {
+    return this.#definitions.get(name);
+  }
+
+  delete(name: string): boolean {
+    return this.#definitions.delete(name);
+  }
+
+  getAll(): Definition[] {
+    return Array.from(this.#definitions.values());
+  }
+
+  get size(): number {
+    return this.#definitions.size;
+  }
+}
--- a/packages/core/core/src/services/mcp/internal/McpSessionManager.ts
+++ b/packages/core/core/src/services/mcp/internal/McpSessionManager.ts
@@ -1,0 +1,120 @@
+import type { Core } from '@strapi/types';
+import { McpSession } from '../session';
+import { McpConfiguration } from './McpConfiguration';
+
+export class McpSessionManager {
+  #sessions = new Map<string, McpSession>();
+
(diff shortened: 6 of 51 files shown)


examples/getstarted/config/server.js
nclsndr, line 1: Are we ok with this config name / file?

nclsndr:
@pwizla FYI, I'll add the docs label once we have a PR from `feat/mcp/base` to `develop`

nclsndr (commented):
This PR is preliminary before the actual `feat/mcp/base` -> `develop` PR to come.

packages/core/core/src/services/mcp/internal/McpSessionManager.ts
```diff
@@ -0,0 +1,120 @@
+import type { Core } from '@strapi/types';
+import { McpSession } from '../session';
+import { McpConfiguration } from './McpConfiguration';
+
+export class McpSessionManager {
+  #sessions = new Map<string, McpSession>();
```
nclsndr, line 6: This implementation sticks the MCP server to only one instance. Load balanced traffic would fail.

examples/getstarted/config/server.js
markkaylor: You mean are we ok with putting the mcp config in the server config file?

packages/core/core/src/services/mcp/handlers/handleDelete.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (sessionId === undefined) {
+      sendJsonRpcError(res, 400, -32000, 'Session ID required');
+      return;
+    }
+
+    const session = sessionManager.get(sessionId);
+    if (session === undefined) {
+      sendJsonRpcError(res, 400, -32000, 'Invalid session');
+      return;
+    }
```
markkaylor, line 24: I saw [here](https://json-rpc.dev/docs/reference/error-codes#server-errors) `-32003` is the error code for an expired session. Not sure that's exactly this case but is there not a more specific error code? I guess the message is clear so it doesn't really matter.

packages/core/core/src/services/mcp/handlers/handleGet.ts
```diff
@@ -0,0 +1,46 @@
...
+    // GET requests in MCP context represent active client engagement waiting for
+    // server messages, not idempotent data retrieval.
+    session.updateActivity();
+
+    try {
+      await withTimeout(
+        session.transport.handleRequest(req, res, null),
+        config.requestTimeoutMs,
+        'transport.handleRequest'
+      );
```
markkaylor: A bit confused to see a comment about "prevent timeout" and then just below a function call to `withTimeout` 🤔 it seems to contradict itself...

Looking into it a bit further, the GET here opens an SSE stream that could still be alive after the timeout has expired which would lead to an error on a healthy GET stream.

packages/core/core/src/services/mcp/internal/McpCapabilityDefinitionRegistry.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  get(name: string): Definition | undefined {
+    return this.#definitions.get(name);
+  }
+
+  delete(name: string): boolean {
+    return this.#definitions.delete(name);
+  }
+
+  get values(): Definition[] {
```
markkaylor: nitpick

is `values` the right getter / method name instead of something like `getAll` or `list`? 

- `values` seems derived from a Map implementation detail
- This isn't just reading a property it computes a new array on every call, deserves a verb?
- get, getAll would have an expected return value Definition vs Definition[]

packages/core/core/src/services/mcp/internal/McpCapabilityDefinitionRegistry.ts
markkaylor, line 1: I wonder if the distinction between this class and `McpConfigurationRegistry` would be more clear if this class was called something like `McpConfigutarionDictionary` or `McpConfigurationCatalog`.

packages/core/core/src/services/mcp/routes.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Handler for unsupported HTTP methods on /mcp endpoint.
+ * Returns JSON-RPC error instead of plain text so MCP clients can parse it.
+ */
+const handleMethodNotAllowed: Core.MiddlewareHandler = async (ctx) => {
+  ctx.status = 405;
+  ctx.set('Allow', 'GET, POST, DELETE');
+  ctx.set('Content-Type', 'application/json');
+  ctx.body = {
+    jsonrpc: '2.0',
+    error: { code: -32000, message: 'Method not allowed' },
```
markkaylor: Again not sure the granularity matters, but [here](https://json-rpc.dev/docs/reference/error-codes) I saw -32601 for method not found / not available

markkaylor (commented):
Gave it a read through but haven't tested it yet. Really nice work @nclsndr !

examples/getstarted/config/server.js
nclsndr: Yes that is the question @markkaylor. Sorry for the fuzziness!

nclsndr:
> Gave it a read through but haven't tested it yet. Really nice work @nclsndr !

Thank you @markkaylor. I guess there's not actual need to test it live just yet, the #26175 will add a more realistic experience once merged with this PR.

examples/getstarted/config/server.js
markkaylor: No worries! Then yeah I think it makes sense 🤔  did you have a doubt or feel for some reason it doesn't belong here?

jhoward1994:
I did a test through curl and also `npx @modelcontextprotocol/inspector` - everything seems to work really well. To give you an idea of the test I ran:

`curl -i -X POST http://localhost:1337/mcp [initialize] to get mcp-session-id`
```
curl -X POST http://localhost:1337/mcp \
	...
  -H "mcp-session-id: ID" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list"}
[List available tools]
```

```
curl -X POST http://localhost:1337/mcp \
	...
    "params": {
      "name": "log",
      "arguments": { "message": "hello from MCP", "level": "warn" }
    }
  }'
[Invoke log tool]
```

`curl -i -X DELETE http://localhost:1337/mcp -H "mcp-session-id: ID`
[Delete session]

nclsndr:
Thank you @jhoward1994 🙌

jhoward1994:
Would it be worth adding e.g. packages/core/core/src/services/mcp/utils/jsonRpcErrors.ts

And using it where needed?


```
/**
 * JSON-RPC 2.0 error codes used by the Strapi MCP server.
 *
 * The range -32000..-32099 is reserved by the JSON-RPC 2.0 spec for
 * implementation-defined server errors. Codes below -32099 (notably -32603
 * "Internal error") are reserved by the spec itself.
 *
 * Distinct codes per error case so MCP clients can branch on them
 * programmatically rather than parsing the human-readable `message`.
 */
export const MCP_JSON_RPC_ERROR_CODES = {
  /** Spec-reserved: server-internal failure. */
  INTERNAL_ERROR: -32603,
  /** Implementation-defined: provided session id does not match any session. */
  INVALID_SESSION: -32000,
  /** Implementation-defined: server has reached `maxSessions` and refused a new one. */
  MAX_SESSIONS_REACHED: -32001,
  /** Implementation-defined: caller did not provide an `mcp-session-id` header. */
  SESSION_REQUIRED: -32002,
  /** Implementation-defined: HTTP method not supported on this endpoint. */
  METHOD_NOT_ALLOWED: -32003,
} as const;

export type McpJsonRpcErrorCode =
  (typeof MCP_JSON_RPC_ERROR_CODES)[keyof typeof MCP_JSON_RPC_ERROR_CODES];

```

packages/core/core/src/services/mcp/tool-registry.ts
```diff
@@ -0,0 +1,89 @@
...
+        createFallbackHandler(errorMessage) {
+          return async () => ({
+            content: [
+              {
+                type: 'text' as const,
+                text: `Tool "${name}" failed to initialize: ${errorMessage}`,
+              },
+            ],
+            structuredContent: {
+              dope: true,
```
jhoward1994: Should we remove this ?

examples/getstarted/config/server.js
nclsndr: Right thx! I'm good with it as well, I was asking in case there was a better suited location I did not about.

packages/core/core/src/services/mcp/handlers/handleGet.ts
nclsndr: I managed to leave the comment here while making the fix in the next PR 🤦 
Thank you for catching the incoherence. Will fix it.

packages/core/core/src/services/mcp/handlers/handleDelete.ts
nclsndr: The case here could be an expired session but not only. That's why I left the most "neutral" case here.

packages/core/core/src/services/mcp/internal/McpCapabilityDefinitionRegistry.ts
nclsndr: The mental model is the following:
- capability definitions (build time)
- runtime capability
Both are using "registry" to mimic the strapi registry is a Map based structure from Strapi core. 
Does that make sense?

Unless I missed something, we don't have a `McpConfigurationRegistry` 🤔

markkaylor: Sorry I meant Capability not Configuration.

You have `McpCapabilityRegistry` and `McpCapabilityDefinitionRegistry`

>Both are using "registry" to mimic the strapi registry is a Map based structure from Strapi core.

It makes sense, my comment was mostly about the naming. I found it a bit confusing, thinking it sounds like these two things do exactly the same thing, I had to read through both to understand that  `McpCapabilityRegistry` is a consumer of `McpCapabilityDefinitionRegistry`. 

I just think the naming represents them as peers when they are more like dependencies

>Both are using "registry" to mimic the strapi registry is a Map based structure from Strapi core.

That's true but McpSessionManager is also a Map based structure so I'm not sure if that rule holds up strong enough to drive the naming

Those are my thoughts, but all that being said totally cool with what you got no need to change anything 😆 . And sorry for the confusing comment.

.github/workflows/adminBundleSize.yml
```diff
@@ -30,11 +30,12 @@ jobs:
...
         uses: ./.github/actions/yarn-nm-install
 
-      - name: Monorepo build
-        uses: ./.github/actions/run-build
+      - name: Create install-and-build script
+        run: printf '#!/bin/sh\nset -e\nyarn install\nyarn build\n' > /tmp/install-and-build.sh
 
       - uses: preactjs/compressed-size-action@v2
         with:
+          install-script: 'sh /tmp/install-and-build.sh'
```
nclsndr, line 38: Fixed to avoid: https://github.com/strapi/strapi/actions/runs/25437933561/job/74622434385?pr=26174

packages/core/types/src/core/config/server.ts
```diff
@@ -85,4 +89,5 @@ export interface Server {
   admin?: ServerAdmin;
   webhooks?: Webhooks;
   http?: Http;
+  mcp: McpConfig;
```
unrevised6419: Note: Requiring this config field will be a breaking type change for consumers.

```suggestion
  mcp?: McpConfig;
```

nclsndr: Good catch @unrevised6419 🙌 TY

Issue #23799, opened 2025-06-23
GraphQL relations: request draft and published versions in one query make draft relations disappear


cdreier:
### Node Version

v20.12.0

### NPM/Yarn/PNPM Version

10.5.0

### Strapi Version

5.16.0

### Operating System

Linux (Other)

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

when fetching DRAFT entities in graphql with relations without "Draft & publish" enabled, the relations are always empty or null

the UI and the REST api is working as intended

tested in a fresh strapi project

### Steps to Reproduce

1. datamodel with cars and passengers
2. cars: "Draft & publish" enabled
3. passengers:  "Draft & publish" not enabled
4. add a relation from cars to passengers (does not matter if 1:1 or 1:n)
5. fill dummy data
6. fetch the cars with status = DRAFT
7. passengers are always null

### Expected Behavior

like seen in the UI or in the REST api, the graphql api should correctly expose relations, even when "Draft & publish" is not enabled

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

cdreier:
I want to add one more comment to highlight the impact and severity of this bug: when working on the api, which, i think, is a common case for a headless cms, your are not able to get reliable data from the GraphQL api. 

Fetching relations via the GraphQL api is fundamentally broken!

cdreier:
is anybody here?

i created a repo to show the problem in detail

https://github.com/cdreier/strapi-gql-draft-bug

imo, this is still a production critical bug :(

cdreier:
hi again, i updated to 5.16.1 now - the initial bug is fixed, but introdudced another bug: in some cases we fetch both versions of one entity, now normal relations (in this case the motor) is showing as `null` in draft, and for published everything is there.

i also updated the debug repo with examples

```
query compareCars {
	draft: cars(status: DRAFT) {
		name
		passengers {
			name
		}
		driver {
			name
		}
		motor {
			manufacturer
		}
	}
	published: cars(status: PUBLISHED) {
		name
		passengers {
			name
		}
		driver {
			name
		}
		motor {
			manufacturer
		}
	}
}
```

cdreier:
[...]

PR #26178, opened 2026-04-30
fix(graphql): isolate root query args per root field

cdreier:
### What does it do?

Stores root-query args per root field (keyed by `info.path.key`, i.e. the alias) instead of a single shared `context.rootQueryArgs` property. 

The association resolver walks `info.path` back to its root and looks up the args of *its own* query branch.

### How to test it?

test project https://github.com/cdreier/strapi-gql-draft-bug and my local test query 

```gql
query compareCars {
	draft: cars(status: DRAFT) {
		name
		passengers {
			name
		}
		driver {
			name
		}
		motor {
			manufacturer
		}
	}
	published: cars(status: PUBLISHED) {
		name
		passengers {
			name
		}
		driver {
			name
		}
		motor {
			manufacturer
		}
	}
}
```

also added a regression test :robot: 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/23799

### DIFF
--- a/packages/plugins/graphql/server/src/bootstrap.ts
+++ b/packages/plugins/graphql/server/src/bootstrap.ts
@@ -22,7 +22,7 @@ const merge = mergeWith((a, b) => {
 });
 
 type StrapiGraphQLContext = BaseContext & {
-  rootQueryArgs?: Record<string, unknown>;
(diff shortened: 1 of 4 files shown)


markkaylor (approved):
Looks good to me! Tested and appears to be working as expected. Thanks again @cdreier for taking care of this!

PR #26179, opened 2026-04-30
fix: keep draft link on x-to-one relations non-dp to dp

jhoward1994:
### What does it do?

- Adds `normalizeXToOneRelationValue` in the document-service to keep a "relates to one" payload to a single entry from the user's side.
- Removes the x-to-one `slice(-1)` in the database write path so both the draft and published rows of a related entry stay linked.
- Handles legacy single-FK-column schemas by taking the last id and warning, suggesting a join-table migration.

### Why is it needed?

A "relates to one" field on a content type without Draft & Publish that points at one with Draft & Publish lost its related entry from the Edit View on re-save, while the List View and the REST API still showed it.

### How to test it?

1. Create `Vehicle` (Draft & Publish ON) and `Form submission` (Draft & Publish OFF) with a `vehicle` relation, then publish a `Vehicle`.
2. Create a `Form submission` linked to that `Vehicle`. Edit View shows the relation.
3. Re-save the form submission with the same vehicle. Edit View, List View and `GET /api/form-submissions/<id>?populate=vehicle` all still show it.

Automated:

- `yarn test:unit packages/core/core/src/services/document-service`
- `yarn test:api tests/api/core/strapi/document-service/relations`
- `yarn test:api tests/api/core/content-manager/api/relations-non-dp-edit-view`


innerdvations (approved):
LGTM! The failing test is actually on develop, you should be able to update once https://github.com/strapi/strapi/pull/26198 is merged

PR #26188, opened 2026-05-02
fix(content-manager): use transient prop for DisconnectButton to prevent DOM warning

### LINKED ISSUE #26187: React warning: shouldForwardProp on styled-components in admin panel
### Node Version

v22.14.0

### Package Manager

npm

### Package Manager Version

1.22.22

### Strapi Version

5.44.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When navigating to Content Manager with i18n enabled, 
a React warning appears in the browser console related 
to shouldForwardProp in styled-components. Unknown props 
are being forwarded to DOM elements.

### Steps to Reproduce

1. Clone strapi/strapi develop branch
2. cd examples/getstarted
3. Run yarn develop
4. Log into admin panel
5. Go to Content Manager
6. Open any content type with i18n enabled
7. Switch locale
8. Open browser console (F12)

### Expected Behavior

No console warnings related to styled-components prop forwarding

### Logs

```shell
If you would like automatic filtering of unknown props, 
you can opt-into that behavior via 
<StyleSheetManager shouldForwardProp={...}> 
(connect an API like @emotion/is-prop-valid) 
or consider using transient props 
($ prefix for automatic filtering.)
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

husnain417:
## What does this PR do?

Fixes a styled-components warning in the browser console caused by 
the `disabled` prop being forwarded to the DOM in DisconnectButton.

Changed `disabled` to `$disabled` (transient prop) to prevent 
the unknown prop from reaching the DOM.

## Related issue
Fixes #26187


Bassel17:
Thanks for the contribution. I’m going to close this PR for now because the patch does not appear to fix the reported warning path.

Using transient props is the right general approach for styled-components-only props, but in this case `disabled` is a valid native `<button>` prop, and I don’t see any caller passing either `disabled` or `$disabled` to `DisconnectButton`. The only in-repo usage appears to render `<DisconnectButton type="button">`, while the actual relation row remove control is an `IconButton`.

Because of that, the change looks inert for the issue described in #26187. The PR also currently fails Prettier.

A correct fix should first identify the exact styled component/prop causing the console warning, then either convert that styling-only prop to a transient prop or add `shouldForwardProp` where appropriate.

PR #26189, opened 2026-05-02
chore(admin): remove punycode dependency

unrevised6419:
## Summary

- Drop `punycode` userland dep + `@types/punycode` from `packages/core/admin`
- Replace `new URL(punycode.toASCII(url))` with `new URL(url)` in webhook URL validator
- Guard `fse.pathExistsSync` against `undefined` from `getConfigPath` in `packages/utils/typescript` (avoids `DEP0187` on Node >=22)

The `punycode` shim was used to dodge the deprecation warning on Node's built-in `punycode` module. WHATWG `URL` already does IDNA on the host natively, so the explicit `toASCII` call is redundant on Node >=20 (the engine range declared in `package.json`).

Per [Node docs](https://nodejs.org/docs/latest-v20.x/api/url.html#new-urlinput-base):

> Unicode characters appearing within the host name of input will be automatically converted to ASCII using the Punycode algorithm.
>
> ```js
> const myURL = new URL('https://測試');
> // https://xn--g6w251d/
> ```

The second commit is a related Node deprecation cleanup: `is-using-typescript-sync.js` was passing a possibly-`undefined` value into `fse.pathExistsSync`, which trips `DEP0187` on Node >=22. Short-circuited to `false` when the config path is unresolved. Happy to split into a separate PR if preferred.

## Test plan

- [x] `tsc --noEmit` on `packages/core/admin/server` — passes
- [x] Manual: create webhook with IDN URL (e.g. `https://測試.example`) in production-mode admin and confirm validator still resolves hostname against `is-localhost-ip`
- [x] No `DEP0187` warning when running CLI in a TS project on Node 22


innerdvations (approved):
Ahhhhh, this has been right on the edge of our backlog forever, thanks for bringing it to the surface!

PR #26194, opened 2026-05-03
chore(deps): drop React 17 from peer dependency ranges

unrevised6419:
### What does it do?

Tightens the `react` and `react-dom` peer dependency ranges from
`^17.0.0 || ^18.0.0` to `^18.0.0` across 15 admin/plugin workspaces.
**Effectively drops React 17 from the supported matrix** — server-only
packages (no React in `peerDependencies`) are unaffected.

Affected workspaces:

- `@strapi/admin`, `@strapi/content-manager`, `@strapi/content-releases`,
  `@strapi/content-type-builder`, `@strapi/email`, `@strapi/review-workflows`,
  `@strapi/strapi`, `@strapi/upload`
- `@strapi/plugin-cloud`, `@strapi/plugin-color-picker`,
  `@strapi/plugin-documentation`, `@strapi/plugin-graphql`,
  `@strapi/i18n`, `@strapi/plugin-sentry`, `@strapi/plugin-users-permissions`

`yarn.lock` is regenerated; only the peer-range metadata changes — no
resolution or transitive dependency shifts.

#### Semver / release-notes impact

Tightening a peer range is technically a breaking change for any host
that satisfied the old union but not the new one (i.e. a React 17
host). In practice Strapi v5 already requires React 18 in the official
docs, the engine matrix, and the templates — so this is a
metadata/documentation correction that brings the published manifests
in line with the *de facto* support matrix, rather than a behavioural
break for any currently-working consumer. Worth a line in the next
release notes regardless, so React-17-on-Strapi-5 users (if they exist)
get a clear signal to upgrade React rather than a silent
`yarn install` followed by a runtime crash.

### Why is it needed?

The advertised `^17.0.0 || ^18.0.0` range was out of date:

- `devDependencies` pin `react@18.3.1` — nothing has been built or
  tested against React 17 since the React 18 upgrade.
- The admin runtime uses React 18-only APIs (`createRoot`,
  automatic batching, `useId`, `useSyncExternalStore`).
- Installing into a React 17 host would `yarn install` clean and then
  break at runtime, with no CI signal.
- `@types/react` 17 vs 18 diverge (`ReactNode`, implicit `children` on
  `FC`), so type errors leak to consumers.
- Risk of two React copies via dedupe ("Invalid hook call") when a host
  app stays on 17 while a Strapi plugin pulls 18 transitively.

Narrowing the range to `^18.0.0` makes the advertised contract match
reality.

<details>
<summary><b>Why <code>^18.0.0</code> and not <code>^18.3.0</code> / <code>~18.3.1</code>?</b></summary>

`devDependencies` pin `react@18.3.1`, so a stricter range would be more
"honest" in principle. In practice it isn't necessary — checked the
React CHANGELOG for what landed in each 18.x minor:

| Version | New public APIs |
|---|---|
| **18.0.0** | All the major ones: `createRoot`, `useId`, `useTransition`, `useSyncExternalStore`, automatic batching, etc. |
| **18.1.0** | None. Bug fixes only. |
| **18.2.0** | One: second `componentStack` arg to `onRecoverableError`. |
| **18.3.0** | None. Identical to 18.2 but adds deprecation warnings for React 19. |
| **18.3.1** | One: `act` exported from `'react'` (previously only `react-dom/test-utils`). |

Verified that no code in this repo imports `act` from `'react'` or
relies on the second arg of `onRecoverableError`, so every React API
the admin actually uses ships in 18.0.0. `^18.0.0` is therefore honest
in practice and matches the npm convention for caret ranges on a major.

</details>

### How to test it?

1. Verify the published peer range is tightened:

   ```sh
   jq '.peerDependencies' packages/core/admin/package.json
   # → { "react": "^18.0.0", "react-dom": "^18.0.0", ... }
   ```

   Repeat for any of the other 14 affected `package.json` files.

2. Confirm `yarn install` still resolves cleanly inside this monorepo —
   no new peer-dependency warnings are introduced. (Pre-existing
   warnings about `react@19.0.0-rc` from `examples/experimental-dev` and
   the unrelated `zod`/`typescript`/`msw` warnings are untouched and
   out of scope here.)

3. (Optional) Simulate a React 17 consumer by packing a workspace and
   installing it into a scratch project alongside React 17:

   ```sh
   # Pack a small affected plugin from this branch
   (cd packages/plugins/sentry && npm pack --pack-destination /tmp)

   # Try to install it into a React 17 project
   mkdir /tmp/r17-test && cd /tmp/r17-test
   npm init -y
   npm install react@17 react-dom@17 /tmp/strapi-plugin-sentry-5.44.0.tgz
   ```

#### Verified behaviour change

Running the simulation above against this branch vs. `develop`:

| | Result |
|---|---|
| **`develop` (`^17.0.0 \|\| ^18.0.0`)** | `npm install` **succeeds** (1367 packages added, no React peer error). Runtime would break later via `createRoot` etc. |
| **This branch (`^18.0.0`)** | `npm install` **fails** at install time with a clear `ERESOLVE`: `peer react@"^18.0.0" from @strapi/plugin-sentry@5.44.0 ... Could not resolve dependency` |

That is the exact contract change this PR is intended to deliver:
silent runtime breakage → loud install-time failure for unsupported
React versions.

Note: a full network-wide effect requires a coordinated release of all
15 affected packages, since transitive peers from already-published
versions still advertise the old range.

### Related issue(s)/PR(s)

None.

### DIFF
--- a/packages/core/admin/package.json
+++ b/packages/core/admin/package.json
@@ -179,8 +179,8 @@
   },
   "peerDependencies": {
     "@strapi/data-transfer": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
--- a/packages/core/admin/package.json
+++ b/packages/core/admin/package.json
@@ -180,8 +180,8 @@
   },
   "peerDependencies": {
     "@strapi/data-transfer": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
--- a/packages/core/content-manager/package.json
+++ b/packages/core/content-manager/package.json
@@ -130,8 +130,8 @@
   },
   "peerDependencies": {
     "@strapi/admin": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
--- a/packages/core/content-manager/package.json
+++ b/packages/core/content-manager/package.json
@@ -124,8 +124,8 @@
   },
   "peerDependencies": {
     "@strapi/admin": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
--- a/packages/core/content-releases/package.json
+++ b/packages/core/content-releases/package.json
@@ -98,8 +98,8 @@
   "peerDependencies": {
     "@strapi/admin": "^5.0.0",
     "@strapi/content-manager": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
--- a/packages/core/content-releases/package.json
+++ b/packages/core/content-releases/package.json
@@ -102,8 +102,8 @@
   "peerDependencies": {
     "@strapi/admin": "^5.0.0",
     "@strapi/content-manager": "^5.0.0",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
     "react-router-dom": "^6.30.3",
     "styled-components": "^6.0.0"
   },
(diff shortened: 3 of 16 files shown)


innerdvations:
@Bassel17 @Adzouz @jhoward1994 Could one of you please investigate if this is potentially going to be a breaking change in some edge cases? Like old plugins?

unrevised6419:
Btw, I recall that design system also has react@17 peers

unrevised6419:
Worth noting this peer-range correction is already enforced in practice: the Admin entry point mounts via the React 18-only `createRoot` API in [`packages/core/admin/admin/src/render.ts`](https://github.com/strapi/strapi/blob/main/packages/core/admin/admin/src/render.ts):

```ts
import { createRoot } from 'react-dom/client';
// ...
createRoot(mountNode).render(app.render());
```

So React 17 was never actually usable — the old `^17.0.0 || ^18.0.0` range advertised support that could never be exercised. This change just makes the metadata match reality. 👍

nclsndr:
@Adzouz should we update the DS deps at the same time / before this PR gets merged?

nclsndr:
Quick update on the design-system coordination raised earlier ([thread](https://github.com/strapi/strapi/pull/26194#issuecomment-2965764265) / [@nclsndr](https://github.com/nclsndr)):

- [strapi/design-system#2024](https://github.com/strapi/design-system/pull/2024) (drop React 17 peers) shipped in **`@strapi/design-system@2.2.2`** ([release notes](https://github.com/strapi/design-system/releases/tag/%40strapi%2Fdesign-system%402.2.2)).
- Strapi `develop` is already on the latest **`2.2.3`** (same `^18.0.0` React peers) via [#27002](https://github.com/strapi/strapi/pull/27002) — so this PR does **not** need a further DS bump for the React 17 drop.
- This branch only tightens Strapi workspace `react` / `react-dom` peers to `^18.0.0`; DS alignment is already in place.

---
_AI assisted_

nclsndr:
Publishing a pre-release to test out with actual apps: https://github.com/strapi/strapi/actions/runs/29739636683

nclsndr:
@Adzouz @jhoward1994 is there anything that should prevent us to move forward with this PR now?

nclsndr:
@boazpoolman @derrickmehaffy @PaulBratslavsky is there any reason we should be careful removing react 17 support for community code / plugins?

unrevised6419:
Can we have an update on this one please?

https://github.com/strapi/strapi/pull/26194#issuecomment-4685466180

PR #26200, opened 2026-05-04
ci: documentation label is no longer blocking

innerdvations:
### What does it do?

makes the "flag: documentation" label non-blocking; changes its intent to be "needs documentation" rather than "changes something in a way that documentation becomes incorrect" because in that case it would generally be a breaking change anyway

### Why is it needed?

with recent docs improvements, it makes more sense to merge the PR immediately, while leaving the label so that the automated systems pick it up and ensure it gets addressed.

### How to test it?

### Related issue(s)/PR(s)


pwizla (approved):
LGTM! Not sure you want the file in the `dist` folder, too, though?
(GitHub could not render it)

innerdvations:
That's the file that actually gets run, unfortunately any change to an action requires us to rebuild and commit the whole file 🫠

PR #26202, opened 2026-05-04
fix(db): deterministic schema hash by sorting tables before hashing

Baldinof:
### What does it do?

Sort tables by name before hashing the schema to ensure deterministic hashing, and reduce schema checks during startup.

### Why is it needed?

The schema tables' order is not consistent over Strapi restarts. Making the caching of the check by hashing the schema inefficient.

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/database/src/schema/storage.ts
+++ b/packages/core/database/src/schema/storage.ts
@@ -68,7 +68,7 @@ export default (db: Database) => {
       // Sort tables by name for deterministic hashing regardless of insertion order
       const sorted = {
         ...schema,
-        tables: schema.tables.sort((a, b) => a.name.localeCompare(b.name)),
+        tables: schema.tables.toSorted((a, b) => a.name.localeCompare(b.name)),
       };
       return crypto.createHash('sha256').update(JSON.stringify(sorted)).digest('hex');
     },
--- a/packages/core/database/src/schema/storage.ts
+++ b/packages/core/database/src/schema/storage.ts
@@ -65,7 +65,12 @@ export default (db: Database) => {
     },
 
     hashSchema(schema: Schema) {
-      return crypto.createHash('sha256').update(JSON.stringify(schema)).digest('hex');
+      // Sort tables by name for deterministic hashing regardless of insertion order
+      const sorted = {
+        ...schema,
+        tables: schema.tables.sort((a, b) => a.name.localeCompare(b.name)),
+      };
+      return crypto.createHash('sha256').update(JSON.stringify(sorted)).digest('hex');
     },
 
     async add(schema: Schema) {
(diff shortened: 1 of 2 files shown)


nclsndr (commented):
Sounds like a good idea 👌 

To get more context on this.
1. When we say: "The schema tables' order is not consistent over Strapi restarts" is that inconsistent on every restarts (so the hash feature is virtually useless)? Or is it more under specific circonstances?
2. How does changing this would impact the production instances upgrading to the version with that change?

pinging @innerdvations for extra CMS context

Baldinof:
> * When we say: "The schema tables' order is not consistent over Strapi restarts" is that inconsistent on every restarts (so the hash feature is virtually useless)? Or is it more under specific circonstances?

Locally it's redoing a full sync like 50% of the time. My guess is the schema object `tables` field is populated using some sort of `Promise.all()`.

So yes, the hashing seems almost useless ATM.

> * How does changing this would impact the production instances upgrading to the version with that change?

After upgrading, at the first start of the app, the chance of doing a full schema check will be 100% so it will get the new hash after sorting tables.
After that full schema check will only happen if something has been really changed in the schema.

innerdvations:
Yep, this will cause production instances to rehash on the next startup, but as @Baldinof said right now it's already happening around half the time based on a race condition somewhere, so that's fine, this will be much better in the end.

innerdvations (dismissed):
Just pushed a tiny fix (to avoid mutating the original object) and added some tests. If it looks good to you, go ahead and merge!

Baldinof:
Good catch! Thank you Ben :)

packages/core/database/src/schema/storage.ts
```diff
@@ -65,7 +65,12 @@ export default (db: Database) => {
     },
 
     hashSchema(schema: Schema) {
-      return crypto.createHash('sha256').update(JSON.stringify(schema)).digest('hex');
+      // Sort tables by name for deterministic hashing regardless of insertion order
+      const sorted = {
+        ...schema,
+        tables: [...schema.tables].sort((a, b) => a.name.localeCompare(b.name)),
```
nclsndr: Could also be [`toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) if we want to avoid spreading the array before

Baldinof: Good catch! I just amended Ben's commit with it

nclsndr: It seems the edit has not been kept @Baldinof 🤔

Baldinof: I guess I messed up my local remotes. Should be good now

innerdvations (approved):

PR #26204, opened 2026-05-04
fix(admin): surface server error body when responseType is not json

### LINKED ISSUE #26199: useFetchClient does not handle server errors well in case new responseType option is involved
### Node Version

22

### Package Manager

npm

### Package Manager Version

10

### Strapi Version

5.44.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

Example:
```js
try {
    responseType: 'text',
  });
} catch (error) {
  console.log(error)
}
```
```json
{
    "data": null,
    "error": {
        "status": 400,
        "name": "BadRequestError",
        "details": {}
    }
}
```
We do not have access to it:
```
FetchError: Server Error
```

### Steps to Reproduce

### Expected Behavior

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

singhvishalkr:
Fixes #26199.

When `responseType` is `text`, `blob`, or `arrayBuffer`, the response interceptor used to throw `FetchError('Server Error')` without ever reading the response body, so callers had no way to see the actual error from the server. The Strapi API still serialises errors as JSON (`{ error: { message, ... } }`) regardless of the success `responseType`, so this PR reads the body, parses it as JSON when possible, and attaches it to the `FetchError` the same way the json branch does. When the body is not JSON the raw text is wrapped in the same envelope shape so `error.response.data.error.message` is always populated.

Verified the fix against the reproduction in the issue (a 400 from `/lms/reports/download` with `responseType: 'text'` now exposes the original `BadRequestError` message instead of `'Server Error'`). Added two tests in `getFetchClient.test.ts` covering the JSON-envelope and plain-text body cases, and updated the existing 404 blob test to mock `text()` so it keeps passing after the change.

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -297,7 +297,7 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
         } catch {
           parsedError = {
             error: {
-              name: 'Error',
+              name: 'ApplicationError',
               message: text,
               status: response.status,
             },
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -281,7 +281,29 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
   ): Promise<FetchResponse<TData>> => {
     if (responseType !== 'json') {
       if (!response.ok && !validateStatus?.(response.status)) {
-        const fetchError = new FetchError('Server Error');
+        // Strapi servers return a JSON error envelope (`{ error: { message, ... } }`)
+        // for failures regardless of the requested success responseType. Read the body
+        // and try to parse it so callers can surface the actual server-side message
+        // and payload via `error.response.data` instead of seeing only `'Server Error'`.
+        const text = await response.text();
+        let parsedError: ErrorResponse['data'] | undefined;
+        let message = 'Server Error';
+        try {
+          const parsed = text ? JSON.parse(text) : undefined;
+          if (parsed?.error?.message && typeof parsed.error.message === 'string') {
+            message = parsed.error.message;
+          }
+          parsedError = parsed;
+        } catch {
+          parsedError = {
+            error: {
+              name: 'Error',
+              message: text,
+              status: response.status,
+            },
+          };
+        }
+        const fetchError = new FetchError(message, parsedError ? { data: parsedError } : undefined);
         fetchError.status = response.status;
         throw fetchError;
       }
--- a/packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts
+++ b/packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts
@@ -218,7 +218,7 @@ describe('getFetchClient', () => {
         name: 'FetchError',
         status: 502,
         message: 'Server Error',
-        response: { data: { error: { name: 'Error', message: 'Bad Gateway', status: 502 } } },
+        response: { data: { error: { name: 'ApplicationError', message: 'Bad Gateway', status: 502 } } },
       });
     });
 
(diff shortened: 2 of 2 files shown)


innerdvations:
Thanks for the PR! It appears there's a build error, can you please resolve it before I take a look?

singhvishalkr:
81e1284 fixes the admin typecheck: the plain-text fallback envelope used name 'Error', which is not part of the ApiError union (TS2322 in uild:types:admin). It now uses ApplicationError, and the plain-text test expectation matches.

packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -281,7 +281,29 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
   ): Promise<FetchResponse<TData>> => {
     if (responseType !== 'json') {
       if (!response.ok && !validateStatus?.(response.status)) {
-        const fetchError = new FetchError('Server Error');
+        // Strapi servers return a JSON error envelope (`{ error: { message, ... } }`)
+        // for failures regardless of the requested success responseType. Read the body
+        // and try to parse it so callers can surface the actual server-side message
+        // and payload via `error.response.data` instead of seeing only `'Server Error'`.
+        const text = await response.text();
```
innerdvations, line 288: The existing `should refresh token and retry on 401 for blob requests` test still mocks `json()` on the 401 response. After this `response.text()` call that mock will throw `TypeError` (not `FetchError` with status 401), so `withTokenRefresh` will not retry.

Please change that 401 mock to:

```ts
text: () => Promise.resolve(JSON.stringify({ error: { message: 'Unauthorized', status: 401 } })),
```

Keep the refresh + blob success mocks as they are.

_AI-assisted comment._

packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts
```diff
@@ -173,6 +174,54 @@ describe('getFetchClient', () => {
...
+        Promise.resolve({
+          status: 400,
+          ok: false,
+          text: () => Promise.resolve(JSON.stringify(errorBody)),
+        })
+      );
+
+      const fetchClient = getFetchClient();
+
+      await expect(fetchClient.get('/lms/reports/download', { responseType: 'text' })).rejects.toMatchObject({
```
innerdvations, line 198: This line (and the nested `response: { data: { error: ... } }` expect below) exceeds printWidth 100, so `prettier:check` fails. Please wrap them the way Prettier formats them — no logic change.

_AI-assisted comment._

innerdvations (changes requested):
Thanks for this — the interceptor change is the right fix for #26199 (read the body, parse JSON when possible, attach it on `FetchError`). Please address the following before we merge:

1. **Blob 401 test** — `should refresh token and retry on 401 for blob requests` still mocks `json()` on the error response. This PR's non-json error path always calls `response.text()`, so that mock will throw `TypeError` instead of `FetchError` and skip token refresh. Mock `text()` with the same unauthorized JSON envelope. Real `Response.text()` exists so production refresh should still work; this is a test update for the new path.

2. **Prettier** — `prettier:check` fails on the two new `expect` lines in `getFetchClient.test.ts` (printWidth 100). Please wrap those; no logic change.

3. **Rebase** — head is a long way behind `develop`. The interceptor there now special-cases 204 and uses `error.name === 'SyntaxError'` instead of `instanceof`. Please rebase and keep those; the error-body change still belongs on the non-json `!response.ok` branch.

_AI-assisted comment._

singhvishalkr:
Rebased on current develop (3cd7dba349), which includes the 204 special-case and SyntaxError name check — both preserved. The blob 401 test now mocks text() instead of json() (8d376c59da). Two new tests cover the JSON-envelope and plain-text error body paths. Prettier issues resolved.

PR #26205, opened 2026-05-04
test(e2e): fix race conditions with drag methods for each browser

innerdvations:
### What does it do?

Stabilises the “component order” E2E path by tightening **mouse drag-and-drop** in test helpers: separate behaviour for Chromium vs WebKit/Firefox, measure drops against the **inner dynamic-zone row** where react-dnd attaches (not only the outer `<li>`), bump viewport height where WebKit needs it, and on Firefox run a short **after-drag cleanup** (escape + move mouse) so the UI isn’t left in a weird drag state. Pulls the shared WebKit/Firefox drag logic into one reusable helper so we aren’t duplicating the same loop twice.

### Why is it needed?

Locally, **`create-content`** was **failing on WebKit** and **flaky on Firefox** when reordering components in a dynamic zone; Chromium was fine. The failures were down to how those browsers handle drag geometry and pointer state during the test, not product code changes.

### How to test it?

From the repo root, run the content-manager E2E. It should pass reliably in all browsers.

`yarn test:e2e -d content-manager -- create-content`

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/tests/utils/shared.ts
+++ b/tests/utils/shared.ts
@@ -287,7 +287,7 @@ export const findAndClose = async (page: Page, text: string, options: FindAndClo
   const elements = page.locator(`:has-text("${text}")[role="${role}"]`);
 
   if (required) {
-    await expect(elements.first()).toBeVisible(); // expect at least one element
+    await expect(elements.first()).toBeVisible();
   }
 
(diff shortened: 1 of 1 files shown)


mathildeleg (approved):
LGTM to me, thanks so much 🙏🏻✨

PR #26208, opened 2026-05-05
feat(upload, email): support provider modules as strapi factories

unrevised6419:
### What does it do?

Lets upload and email provider modules export a **factory** — a function `({ strapi }) => provider` — instead of only a plain object. Plain-object provider modules keep working exactly as before; the factory form is detected at load time.

- `packages/core/upload/server/src/register.ts`: `createProvider` becomes `async` and takes `{ strapi, config }`. After `require(modulePath)`, if the module is a function it is awaited with `{ strapi }` before `init(providerOptions)` is called. A local `UploadProvider` type replaces the previously untyped `provider` variable.
- `packages/core/email/server/src/bootstrap.ts`: same change for email — `createProvider({ strapi, emailConfig })` is now `async` and awaits the factory form before calling `init(...)`. `bootstrap` awaits it.
- `@strapi/upload` and `@strapi/provider-upload-local` declare `@strapi/types` as an **optional** peer dependency (`peerDependenciesMeta.optional: true`), so the ambient `strapi` global typing is available to consumers without adding a runtime dependency.
- `packages/providers/upload-local/tsconfig.json` picks those types up via `types: ["node", "@strapi/types"]` instead of the `import type {} from '@strapi/types'` side-effect import in `src/index.ts`.
- `packages/core/upload/server/src/index.ts` drops its unused `koa-body` and `@strapi/types` side-effect type imports (nothing in that package references `ctx.request.files`).

Scope note: this PR only adds the *consumer* side. No bundled provider is converted to a factory here, so `@strapi/provider-upload-local` still reads the `strapi` global; converting first-party providers is follow-up work.

### Why is it needed?

Provider modules currently have no way to receive the `strapi` instance. The only way to reach it is the `global.strapi` global, which forces providers to depend on `@strapi/types` purely for the ambient global augmentation, and makes them untestable without setting up that global. Supporting the factory form gives providers dependency injection — the same pattern already used for controllers, services, and `core/factories.ts` — and is the prerequisite for removing `global.strapi` usage from first-party providers.

### How to test it?

Verified locally:

- [x] `yarn workspace @strapi/upload test:ts:back`
- [x] `yarn workspace @strapi/provider-upload-local test:ts`
- [x] `yarn workspace @strapi/provider-upload-local test:unit:vitest` (3 passed)

Manual check in `examples/getstarted` — set `config/plugins.js` upload provider to a local module exporting `({ strapi }) => ({ init: () => ({ upload, uploadStream, delete: ... }) })` and confirm it boots and uploads; then switch the same module back to a plain object export and confirm it still boots.

### Related issue(s)/PR(s)

Related: #26544 (centralising the `koa-body` / `@strapi/types` module-augmentation imports).

### DIFF
--- a/packages/core/upload/server/src/register.ts
+++ b/packages/core/upload/server/src/register.ts
@@ -33,7 +33,7 @@ export async function register({ strapi }: { strapi: Core.Strapi }) {
   sharp.cache(cache);
   sharp.concurrency(concurrency);
 
-  strapi.plugin('upload').provider = createProvider(uploadConfig);
+  strapi.plugin('upload').provider = await createProvider({ strapi, config: uploadConfig });
 
   await registerUploadMiddleware({ strapi });
 
@@ -53,11 +53,17 @@ export async function register({ strapi }: { strapi: Core.Strapi }) {
   }
 }
 
-const createProvider = (config: Config) => {
+type UploadProvider = {
+  init: (config: Record<string, unknown>) => Record<string, any>;
+};
+
+const createProvider = async ({ strapi, config }: { strapi: Core.Strapi; config: Config }) => {
   const { providerOptions, actionOptions = {} } = config;
 
   const providerName = _.toLower(config.provider);
-  let provider;
+  let providerModule:
+    | UploadProvider
+    | ((params: { strapi: Core.Strapi }) => Promise<UploadProvider>);
 
   let modulePath;
   try {
@@ -76,7 +82,7 @@ const createProvider = (config: Config) => {
   }
 
   try {
-    provider = require(modulePath);
+    providerModule = require(modulePath);
   } catch (err) {
     const newError = new Error(`Could not load upload provider "${providerName}".`);
 
@@ -86,6 +92,9 @@ const createProvider = (config: Config) => {
 
     throw newError;
   }
+
+  const provider =
+    typeof providerModule === 'function' ? await providerModule({ strapi }) : providerModule;
 
   const providerInstance = provider.init(providerOptions);
 
--- a/packages/providers/upload-local/src/index.ts
+++ b/packages/providers/upload-local/src/index.ts
@@ -4,9 +4,6 @@ import path from 'path';
 import fse from 'fs-extra';
 import * as utils from '@strapi/utils';
 
-// Needed to load global.strapi without having to put @strapi/types in the regular dependencies
-import type {} from '@strapi/types';
-
 interface File {
   name: string;
   alternativeText?: string;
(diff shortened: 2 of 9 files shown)


packages/core/types/src/utils/index.ts
```diff
@@ -234,3 +236,8 @@ export type OneOf<T, U> = (T & { [K in keyof U]?: never }) | (U & { [K in keyof
  * ```
  */
 export type Pretty<T> = { [K in keyof T]: T[K] } & unknown;
+
+export type StrapiFactory<T> = ({ strapi }: { strapi: Strapi }) => T;
+export type MaybeStrapiFactory<T> = T | StrapiFactory<T>;
+export type StrapiFactoryAsync<T> = ({ strapi }: { strapi: Strapi }) => Promise<T>;
+export type MaybeStrapiFactoryAsync<T> = StrapiFactoryAsync<T> | MaybeStrapiFactory<T>;
```
unrevised6419: Thought: Probably a better naming would be

```
StrapiFactory           -> (same)
MaybeStrapiFactory      -> StrapiFactoryOrValue
StrapiFactoryAsync      -> (same)
MaybeStrapiFactoryAsync -> StrapiFactoryAsyncOrValue
```

packages/providers/upload-local/src/index.ts
```diff
@@ -41,104 +38,114 @@ interface CheckFileSizeOptions {
...
-          if (kbytesToBytes(file.size) > sizeLimit)
-            throw new PayloadTooLargeError(
-              `${file.name} exceeds size limit of ${bytesToHumanReadable(sizeLimit)}.`
+export interface StrapiHost {
+  dirs: {
+    static: {
+      public: string;
+    };
+  };
+}
```
unrevised6419: Note: I think for the providers is safe to have `@strapi/types` in dev and peers, and import it directly.

packages/core/upload/server/src/register.ts
```diff
@@ -53,11 +53,15 @@ export async function register({ strapi }: { strapi: Core.Strapi }) {
   }
 }
 
-const createProvider = (config: Config) => {
+type UploadProvider = {
+  init: (config: Record<string, unknown>) => Record<string, any>;
+};
```
unrevised6419, line 58: Thought: This maybe can be moved to types e.g. `Providers.Upload.Provider`, `Providers.Upload.Options` ...

unrevised6419: https://github.com/strapi/strapi/pull/27129

PR #26209, opened 2026-05-05
fix(admin): remove year 2041 limit on date/datetime pickers

### LINKED ISSUE #26180: Date Picker Limited to year 2041
### Node Version

22.15.1

### Package Manager

npm

### Package Manager Version

11.13.0

### Strapi Version

5.43

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When adding a date field to a content type, the Date Picker in the Admin UI is limited to a max year of 2041. When trying to manually input a year greater than 2041 it automatically reverts to 2041.

### Steps to Reproduce

1. Create a field of type date in any content type
2. Add a new content item in the Admin UI
3. Try to enter a date with year greater than 2041

### Expected Behavior

Datepicker should allow to select years beyond 2041. ideally the Datepicker item should have a configurable maximum and minimum date.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

MaksZhukov:
## What

Explicitly set `maxDate` to `new Date(9999, 11, 31)` on both the `DatePicker` and `DateTimePicker` components in the admin `FormInputs`.

## Why

The `@strapi/design-system` DatePicker defaults to `currentYear + 15` as `maxDate` when no explicit value is provided. With the current year being 2026, that cap is **2041** — making it impossible to select any date beyond that year in Strapi's admin UI.

Reported in #26180.

## How to test

1. Create a content type with a **Date** or **DateTime** field.
2. Add an entry and click the date picker.
3. Navigate the year selector — it should now go well past 2041 (up to year 9999).
4. Verify that passing a custom `maxDate` via the content type config still works (the explicit prop is placed before `{...props}` so it can be overridden).

Closes #26180


Link2Twenty:
Would it be better to make the changes in the design system? That's where the 15 year limit is applied.

https://github.com/strapi/design-system/blob/main/packages/design-system/src/components/DatePicker/DatePicker.tsx

https://github.com/strapi/design-system/blob/main/packages/design-system/src/components/DatePicker/DatePicker.tsx#L46

https://github.com/strapi/design-system/blob/main/packages/design-system/src/components/DatePicker/DatePicker.tsx#L175-L190

packages/core/admin/admin/src/components/FormInputs/Date.tsx
```diff
@@ -8,6 +8,8 @@ import { useField } from '../Form';
 
 import { InputProps } from './types';
 
+const MAX_DATE = new Date(9999, 11, 31);
```
Adzouz: Can you put 2099 instead?

Adzouz (changes requested):
Hi @MaksZhukov, thanks for the contribution. Although we thought it would still be good to have a defined limit but reasonable (for now). I proposed to change it for 2099 also because 9999 generates a timeout and an error.
For the design system update, since the option can be overridden by the user, I wouldn't change it there.

packages/core/admin/admin/src/components/FormInputs/Date.tsx
MaksZhukov: Done — changed to 2099 in both `Date.tsx` and `DateTime.tsx` in 1d74be23f0.

MaksZhukov:
@Adzouz Thanks for the review! Updated `MAX_DATE` to `new Date(2099, 11, 31)` in both `Date.tsx` and `DateTime.tsx` (1d74be23f0). The PR doesn't touch the design system, so no changes needed there.

Adzouz (approved):
Looks good, thanks 🙏

PR #26210, opened 2026-05-05
fix(database): prevent crash when reordering and removing a relation in the same save

### LINKED ISSUE #25710: Entity manager: disconnect processed before connect ordering causes sortConnectArray error
### Node Version

20

### NPM/Yarn/PNPM Version

npm 10

### Strapi Version

5.33.0

### Operating System

macOS / Linux

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When reordering and removing relations simultaneously in a component's relation field (e.g., drag-and-drop reorder + remove via Admin UI), the entity manager throws:

> There was a problem connecting relation with id **A** at position {"before":**B**,"status":"draft","locale":"en"}. The relation with id **B** needs to be connected first.

**Root cause:** The entity manager in `@strapi/database` processes `disconnect` before evaluating `connect` ordering. When a relation is both referenced by `position.before` in the connect array and present in the disconnect array, the following happens:

1. `disconnect` deletes the referenced relation (entity B) from the join table
2. `adjacentRelations` query looks for entity B — **not found** (just deleted)
3. `sortConnectArray` with `strictSort=true` throws the error

Additionally, `_.differenceWith(_.isEqual, disconnect, connect)` at [entity-manager/index.js L853](https://github.com/strapi/strapi/blob/main/packages/core/database/src/entity-manager/index.ts) always returns all disconnect items because disconnect items `{id}` and connect items `{id, position, __pivot, __type}` have structurally different shapes, so deep equality always fails. This means every disconnect item is unconditionally deleted regardless of whether it appears in the connect array.

### Steps to Reproduce

1. Create a content type with a component that has a `oneToMany` relation (e.g., a "PickUp" component with `articles`)
2. Add 3+ articles to the component and save
3. In the Admin UI, **drag** article A before article B (this creates `connect: [{A, position: {before: B}}]`)
4. **Remove** article B from the list (this creates `disconnect: [{B}]`)
5. Save → **Error: "The relation with id B needs to be connected first."**

**Key detail:** The error only occurs when steps 3 and 4 are done in the same save. Doing them separately (save after reorder, then save after delete) works fine.

### Diagnostic output

```
connect: [{ "id": 4, "documentId": "zmlc8gzg9ta3gnvnym26sxoy", "position": { "before": "hlvmnosjrfe4h7zqddwlhcil", "status": "draft" } }]
disconnect: [{ "id": 10, "documentId": "hlvmnosjrfe4h7zqddwlhcil" }]

Error: There was a problem connecting relation with id 4 at position {"before":10,"status":"draft","locale":"en"}. The relation with id 10 needs to be connected first.
```

Note: `position.before` documentId (`hlvmnosjrfe4h7zqddwlhcil`) matches the disconnect item's documentId — confirmed the disconnect target is the same as the position reference.

### Expected Behavior

The save should succeed. When `position.before` references a relation that is being disconnected, the server should either:

1. Process connect ordering **before** deleting disconnect items, or
2. Detect the conflict and fall back to `{end: true}` for the affected connect item, or
3. Fix `_.differenceWith` to compare by `id` instead of deep equality, and skip deleting relations that are also in the connect array

### Workaround

**For users:** Do not reorder and remove relations in the same save. Save them separately.

**For developers:** Add a middleware/lifecycle hook that sanitizes the connect array before it reaches the entity manager — if `position.before/after` references a documentId in the disconnect array, replace the position with `{end: true}`.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

- This affects any component with an ordered relation field (oneToMany, manyToMany) when using the Admin UI
- The Admin UI's `handleMoveItem` (Relations.mjs) sets `position.before` based on the current UI state, but `handleDisconnect` doesn't recalculate position references for remaining connect items
- Both the Admin UI (not recalculating positions) and the entity manager (processing disconnect before connect ordering) contribute to this bug

Related issues (different root causes):
- #23158 — component relation sync on publish (fixed in #24227)
- #22451 — unable to reorder newly added relations in SingleType (fixed in v5.4.0)

### Confirmation Checklist

MaksZhukov:
## What

Two targeted fixes in the `isPartialUpdate` relation path of `createEntityManager`:

**Fix 1 — `differenceWith` comparison**

`differenceWith(isEqual, disconnect, connect)` was intended to skip deleting any disconnect item that also appears in the connect array (to avoid deleting a relation that is about to be re-added). Because disconnect items are plain `{id}` objects while connect items carry extra fields (`position`, `__pivot`, `__type`, …), `isEqual` (deep equality) never matched — so every disconnect item was unconditionally deleted. Changed the comparator to `(a, b) => a.id === b.id`.

**Fix 2 — position reference to a being-disconnected relation**

When a connect item's `position.before` (or `position.after`) pointed to a relation that was also in the disconnect list, `deleteRelations` removed that row first. The subsequent `adjacentRelations` DB query could not find it, and `sortConnectArray` (strict mode) threw:

```
There was a problem connecting relation with id A at position {"before": B}.
The relation with id B needs to be connected first.
```

The fix rewrites any such position to `{ end: true }` before the delete runs. The ordering gracefully falls back to appending the connect item instead of crashing.

## Why

Reported in #25710: drag-reordering article A before article B **and** removing article B in the same admin UI save triggers the crash. The two operations are independent in the UI but land together in one API call.

## How to test

1. Create a content type with a component that has a `oneToMany` ordered relation.
2. Add 3+ related entries, save.
3. In the Admin UI, **drag** entry A before entry B (sets `connect: [{id:A, position:{before:B_id}}]`).
4. **Remove** entry B (`disconnect: [{id:B}]`).
5. Save — should succeed without error.
6. Verify the final order: A appears where B was, B is gone.

Closes #25710


MaksZhukov:
@innerdvations Hey, could you review?

innerdvations (changes requested):
Thanks for this fix! The `differenceWith` id-only fix looks right, and I reproduced the crash on `develop`.

I had a concern about "Fix 2" though: rewriting stale position refs to `{ end: true }` fixes the crash but can save the wrong order (e.g. `[B, A, C]` → move A before B + remove B should give `[A, C]`, but `{ end: true }` gives `[C, A]`). 

I pushed a suggestion branch you can take a look at: **`fix/26210-preserve-relation-order`** — it keeps your Fix 1 and replaces the `{ end: true }` fallback with a neighbor-aware rewrite using the current join-table order before delete, plus unit/API tests. No obligation to use it as-is if you have a different fix in mind, but feel free to merge it in if you like it.

MaksZhukov:
Thanks for the detailed review, @innerdvations — good catch on the ordering issue.

I looked for the `fix/26210-preserve-relation-order` branch you mentioned (checked this repo, your fork, and a broader search) but couldn't find it, so I re-implemented the fix directly based on your description of the problem.

New approach in `c1482c305d`: instead of rewriting a stale `position.before`/`position.after` to `{ end: true }`, I fetch the relation's current order before the delete runs and rewrite the position to point at the nearest **surviving neighbor** in the same direction. So for your example, `[B, A, C]` with "move A before B" + "remove B" now correctly resolves to `[A, C]` instead of `[C, A]`. It only falls back to `{ end: true }` / `{ start: true }` if there's no surviving neighbor in that direction (e.g. everything after/before the target is also being disconnected).

Also added an API test (`Reorder a relation before one that is disconnected in the same call` in `relations.test.api.ts`) covering exactly this scenario, run against both `idObject` and `documentId` formats — 114/114 relevant tests pass locally, including the existing reorder/disconnect suites.

Let me know if you'd still like the branch you had in mind — happy to compare approaches if you can re-push it.

Bassel17:
Thanks for the contribution. I pushed a few follow-up changes on top of the PR to cover some extra edge cases around reorder + disconnect behavior.

I widened the regression coverage for strict invalid anchors, multiple connects around the same disconnected relation, and reconnecting the same relation with position metadata. The relation API test file is green now, so this looks good from my side.

Bassel17 (approved):

PR #26211, opened 2026-05-05
fix(core/core): validation uses injected strapi instance

nclsndr:
### What does it do?

Eliminates all `global.strapi` usages in the validation module (`packages/core/core/src/core-api/routes/validation/`) by threading the Strapi instance via explicit function parameters:

- `utils.ts` — `safeGlobalRegistrySet(strapi, id, schema)` and `safeSchemaCreation(strapi, id, callback)` gain `strapi: Core.Strapi` as first param; 8 bare `strapi.log.*` calls now use the injected instance.
- `attributes.ts` — `componentToSchema`, `mediaToSchema`, and `relationToSchema` gain `strapi: Core.Strapi` as first param and forward it into `safeSchemaCreation` and validator constructors.
- `mappers.ts` — All four public functions (`mapAttributeToSchema`, `mapAttributeToInputSchema`, `createAttributesSchema`, `createAttributesInputSchema`) gain `strapi: Core.Strapi` as first param; both `customField` default branches resolve the instance from the injected param instead of `global.strapi`.
- `content-type.ts` and `component.ts` — Call sites updated to pass `this._strapi` (already available from `AbstractCoreRouteValidator`).

### Why is it needed?

Relying on `global.strapi` is an anti-pattern in this codebase: it makes the code harder to test in isolation, bypasses the established DI convention (`strapi` injected via factory/constructor), and creates implicit hidden dependencies. `AbstractCoreRouteValidator` already held `this._strapi` — this change propagates it consistently downward through the call chain.

### How to test it?

No behaviour change is expected. Verify with:

```bash
# Type-check the affected package
yarn nx build @strapi/core --skip-nx-cache

# Unit tests
yarn test:unit

# Start the dev sandbox and exercise content-type routes to confirm validation still works
# Then create/update a document via the admin or REST API and confirm no validation regressions
```

### Related issue(s)/PR(s)

ø

### DIFF
--- a/packages/core/core/src/core-api/routes/validation/attributes.ts
+++ b/packages/core/core/src/core-api/routes/validation/attributes.ts
@@ -5,7 +5,7 @@
  * of incoming data.
  */
 
-import { type Schema, UID } from '@strapi/types';
+import { type Core, type Schema, UID } from '@strapi/types';
 
 import {
   relations,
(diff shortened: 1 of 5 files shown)


innerdvations (approved):
Nice! And this should be safe, since it's not user-facing

PR #26212, opened 2026-05-05
fix(test): check button attribute differently

mathildeleg:
### What does it do?
Update the `Field.test.tsx` to avoid flakiness. Made the test not depend on RTK-query timing by mocking `useDoc` + `useDocument` (from `hooks/useDocument`) so `useDocumentContext()` always gets a `currentDocument` with:
* `isLoading: false` (so “Add a component to…” is enabled immediately)
* a minimal `components` dictionary for the UIDs used in the test
* `blog.test-como` includes a name attribute defaulting to 'toto' so the existing assertions still pass.

I also mocked `useDocumentLayout` so `DynamicComponent` uses `mainField: 'name'`, keeping the `test comp - test` label stable, and finally updated `waitForQueryToFinish()` to simply wait for the button to exist and then assert it’s enabled (no timeouts, no polling on async queries).

### Why is it needed?
The CI-only flake is due to the button staying disabled because document/schema queries are slow.

### How to test it?
The flake only happened on the CI, the test passes locally.

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/tests/Field.test.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/DynamicZone/tests/Field.test.tsx
@@ -15,12 +15,57 @@ jest.mock('../../../InputRenderer', () => ({
   InputRenderer: () => 'INPUTS',
 }));
 
+const useDocMock = jest.fn();
+const useDocumentMock = jest.fn();
+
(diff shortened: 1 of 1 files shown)


nclsndr (approved):
LGTM! Thx @mathildeleg 🙌

Adzouz (approved):
Thanks @mathildeleg 🙏 I think honestly that's still better than increasing the timeout and I see the pipelines are passing now ✨

PR #26213, opened 2026-05-05
chore(types): centralise the BrowserStrapi type in @strapi/types

unrevised6419:
> [!NOTE]
> This PR is based on #27683, which now carries the runtime changes previously in this PR (explicit backend URL for the fetch client and token refresh, `createBrowserStrapi`, the `GetProjectType` contract). Until #27683 is merged, this PR also shows its commit; review only the last commit, `chore(types): centralise the BrowserStrapi type in @strapi/types`.

### What does it do?

**One source of truth for the `window.strapi` type**

- Moves the `BrowserStrapi` interface out of `packages/core/admin/admin/custom.d.ts` into a regular `.ts` module at `packages/core/types/src/_internal/admin/browser-strapi.ts`.
- Declares the `window.strapi` global once, in `packages/core/types/src/_internal/globals-admin.ts`. 13 admin and plugin tsconfigs load it through their `"types"` array, replacing the separate, drifting `interface Window { strapi: ... }` declarations in `admin`, `ee admin`, `content-manager`, `content-releases`, `review-workflows`, `upload`, `cloud`, `documentation`, `i18n` and `admin-test-utils`.
- Both are exposed under `@strapi/types/_internal`, not the package root: this is first-party plumbing, not plugin API. The two subpaths (`./_internal`, `./_internal/globals-admin`) only declare `types` (and `source`) conditions, because `@strapi/types` builds a single root entry and emits no runtime files for them.
- `createBrowserStrapi` (from #27683) is typed with `Admin.BrowserStrapi` instead of `Window['strapi']`.

**Declaration file cleanup**

- Deletes the per-package `styled-components` `DefaultTheme` augmentations. `@strapi/design-system` already declares `DefaultTheme extends StrapiTheme` in its published types.
- Replaces the `/// <reference types="vite/client" />` directives and the local `*.svg`, `*.png` and `*?raw` module declarations with `"vite/client"` in each tsconfig `"types"` array, which declares those modules.
- Removes the admin-side `custom.d.ts`, `module.d.ts` and `global.d.ts` files and their tsconfig `include` entries. The only admin-side augmentation with remaining content, `content-manager`'s Slate `CustomTypes`, moves to `packages/core/content-manager/admin/src/globals.d.ts`. Server-side ambient files are renamed to the same `src/globals.d.ts` convention (`content-releases`, `database`, `permissions`, `utils`, `generators`, `strapi`, `documentation`).
- Drops `packages/core/core/src/modules.d.ts`, which hand-declared `koa-favicon`. `@types/koa-favicon` used to name the option `maxage` while the runtime reads `maxAge`; [DefinitelyTyped#75243](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/75243) fixed the casing in 2.1.5, which `@strapi/core` now depends on.
- Types the `window.strapi` stub in `admin-test-utils/src/setup.ts` as `BrowserStrapi`, adding the fields it was missing (`isTrial`, `isTrialLicense`, `ai`, `REVIEW_WORKFLOWS`, `AUDIT_LOGS`), and adds `@strapi/types` to that package's `devDependencies`.
- Adds `@types/eslint` to the root `devDependencies` for the `// @ts-check`-ed `lint-staged.shared.mjs`.
- Points the admin telemetry contributor doc at the new type location.

**Not covered here**

- In DOM programs, the server-side `strapi` global and the browser `window.strapi` augmentation still merge into a `BrowserStrapi & Core.Strapi` intersection, which is why `render.ts` and `admin-test-utils/src/setup.ts` each keep one `@ts-expect-error`. The server side is addressed separately in #26544.
- No automated check enforces that `@strapi/types` export conditions match emitted files.

### Why is it needed?

The shared tsconfig presets set `skipLibCheck: true`, so the `.d.ts` files describing `window.strapi` were never checked against each other or against the object the admin actually builds. Each package kept its own copy with different fields, and the test utilities built a `window.strapi` that no longer matched the runtime.

`@strapi/admin` can't host the type: `@strapi/admin-test-utils` would need to depend on it, while `@strapi/admin` already depends on `@strapi/admin-test-utils`. `@strapi/types` is a leaf package every admin and plugin package already uses.

### How to test it?

- `yarn test:ts` type-checks every admin and plugin program against the shared declaration.
- `yarn test:front` and `yarn test:unit` run with the typed `admin-test-utils` stub.
- `yarn lint` and `yarn prettier:check`.

### Related issue(s)/PR(s)

- #27683 — base of this PR: runtime changes split out of it.
- #26544 — server-side complement: declares the `strapi` global per workspace.


unrevised6419:
@nclsndr did push some changes to fix the CI.

Note: I moved the type to `@strapi/types`, because if adding `@strapi/admin` to `@strapi/admin-test-utils` it creates a circular dependency.

Note: I rebased on top of latest changes/release

packages/core/types/package.json
```diff
@@ -30,6 +30,21 @@
...
+      "require": "./dist/index.js",
+      "default": "./dist/index.js"
+    },
+    "./global-admin": {
+      "types": "./dist/global-admin.d.ts",
+      "source": "./src/global-admin.ts",
+      "require": "./dist/global-admin.js",
+      "default": "./dist/global-admin.js"
+    },
+    "./*": "./*"
```
unrevised6419, line 50: Note: Without this a lot of errors like this will surface, because TS cannot reach namespace types.

```
The inferred type of 'default' cannot be named without a reference to '../../../../../../node_modules/@strapi/types/dist/public'. This is likely not portable. A type annotation is necessary.ts(2742)
```

packages/core/types/src/globals-admin.ts
```diff
@@ -0,0 +1,10 @@
+import type { BrowserStrapi } from './admin';
+
+declare global {
+  interface Window {
+    strapi?: BrowserStrapi;
+  }
+}
```
unrevised6419: We could do the same for the `global.strapi` a `@strapi/types/globals-server`, but I saw in the code that `global.strapi` will be deprecated and removed in next major


**packages/core/core/src/index.ts#19**

```ts
  // TODO: deprecate and remove in next major
  global.strapi = strapi;
```

And I gave a second thought. Is `window.strapi` a public API?
I created a new project, and `window.strapi` resolves to the server `Strapi` instance.

What are your thoughts here @nclsndr ?

IMO, both `global.strapi` and `window.strapi` should be deprecated, and used instead factory on server, and Provider/hook on admin.

unrevised6419: If we don't want yet to mark `global.strapi` as deprecated internally, we could create a `@strapi/types/global-server` and use it internally so we could split admin/server global augmentations, and mark current augmentation from `@strapi/types` as deprecated.

innerdvations:
Looks great! Could you please resolve the merge conflicts with latest develop? (Looks like adminTokensFutureFlag.test.ts may drop out after #26391)

We'll start looking into the best approach for the admin/server global conflict you raised and see what the best path forward is.

Would also like to get somenone else's feedback just to be safe, cc @nclsndr @Bassel17 @jhoward1994 @Adzouz

unrevised6419:
@innerdvations resolved

nclsndr:
Sorry for the radio silence @unrevised6419! I'll get back on the TS topics by the end of this week

packages/core/types/src/globals-admin.ts
unrevised6419: Related https://github.com/strapi/strapi/pull/26544

unrevised6419:
Note: If we merge this after https://github.com/strapi/strapi/pull/26699 we can get rid completely of `*.d.ts` files and move `vite/client` and `@strapi/types/globals-admin` into `tsconfig.json#compilerOptions.types`

Theoretically that should improve IDE type checking speed for individual files, as it does not need to inspect all the files of the tsconfig project.

nclsndr:
> Note: If we merge this after https://github.com/strapi/strapi/pull/26699

Most likely be the case, we're moving forward on #26699 atm

packages/core/types/src/globals-admin.ts
```diff
@@ -0,0 +1,7 @@
+import type { BrowserStrapi } from './admin';
+
+declare global {
+  interface Window {
+    strapi?: BrowserStrapi;
+  }
```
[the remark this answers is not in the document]
unrevised6419: @coderabbitai Not worth it. It changes the behavior. That object is already created, instead of overriding it completely the example shows how to change the fields instead.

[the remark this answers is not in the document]
unrevised6419: @coderabbitai my comment was about your suggestion for the `telemetry.md` file

unrevised6419:
@innerdvations after CI will end, I will push some suggestion made by coderabbit

packages/core/admin/admin/src/render.ts
```diff
@@ -92,18 +90,16 @@ const renderAdmin = async (
       },
     } = await get<{ data: ProjectType }>('/admin/project-type');
 
-    window.strapi.isEE = isEE;
-    window.strapi.isTrialLicense = isTrial;
```
unrevised6419, line 96: Note: This `isTrialLicense` was added in 2 places in type, but was never used, so I deleted it. There is `isTrial` instead.

packages/core/types/src/admin/browser-strapi.ts
```diff
@@ -0,0 +1,26 @@
...
+    AUDIT_LOGS: 'audit-logs';
+    REVIEW_WORKFLOWS: 'review-workflows';
+    isEnabled: (featureName?: string) => boolean;
+  };
+  flags: {
+    promoteEE?: boolean;
+    nps?: boolean;
+    docLinks?: boolean;
+  };
+  projectType: 'Community' | 'Growth' | 'Enterprise';
```
unrevised6419, line 26: Note: `Growth` was added recently.

packages/core/types/package.json
```diff
@@ -38,6 +38,12 @@
       "require": "./dist/index.js",
       "default": "./dist/index.js"
     },
+    "./globals-admin": {
+      "types": "./dist/globals-admin.d.ts",
+      "source": "./src/globals-admin.ts",
+      "require": "./dist/globals-admin.js",
+      "default": "./dist/globals-admin.js"
+    },
```
unrevised6419: Note: The `window.strapi` was never exposed as a "public api". This `exports` declaration, creates a "public api" entry point that consumers can add to their tsconfig, and have the `window.strapi` typing. Do we want that? We could create an internal package that exports it, so we can use it internally, and there won't be public api.

If we want to make `window.strapi` a public api, I would suggest instead to have a react hook that provides it maybe.

window.strapi -> react hook
global.strapi -> factory function

cc @nclsndr

unrevised6419: @nclsndr what do you think? Should we make it `_internal/globals-admin` instead?

packages/core/admin/admin/src/render.ts
```diff
@@ -60,16 +58,14 @@ const renderAdmin = async (
...
       docLinks: true,
     },
-    // eslint-disable-next-line
-    // @ts-ignore – there's pollution from the global scope of Node. Cannot use @ts-expect-error because of build:code and build:types context collision.
     ai: {
       enabled: true,
     },
   };
 
   const { get } = getFetchClient();
```
nclsndr: **Initialize the browser contract before creating the fetch client (blocker)**

`renderAdmin` has already built a local `browserStrapi` containing the configured `backendURL`, but `getFetchClient()` does not use that value: it immediately reads `window.strapi.backendURL`. On a fresh admin document, `window.strapi` has not been assigned yet and the browser may expose `<div id="strapi">` through named-property access instead. The captured URL is therefore `undefined`, so the project-type request resolves to `/undefined/admin/project-type`; the error is caught and bootstrap silently keeps the Community defaults instead of applying the EE/license response.

I would keep the “assign the completed object once” approach and let this bootstrap request receive `browserStrapi.backendURL` explicitly, rather than making `getFetchClient` depend on an uninitialized global. The smaller compatibility fix is to assign the local object before creating the client and continue mutating that same object. In either case, could we add a regression test that starts with no pre-seeded `window.strapi` (and the real `#strapi` mount node), invokes `renderAdmin`, and asserts both the `/admin/project-type` URL and the final EE/project fields?

---
_AI assisted_

packages/core/types/src/globals-admin.ts
```diff
@@ -0,0 +1,7 @@
+import type { BrowserStrapi } from './admin';
+
+declare global {
+  interface Window {
+    strapi: BrowserStrapi;
```
nclsndr: **Separate the browser and server global augmentations (blocker)**

The package root still declares the server global as `var strapi: Core.Strapi`. In a DOM program that declaration contributes a `strapi` property to the global/window shape; loading this file then declares the same property as `BrowserStrapi`. TypeScript combines them, so the effective type becomes `BrowserStrapi & Core.Strapi`. A strict standalone consumer reproduced both sides of the defect: assigning a complete `BrowserStrapi` fails with `TS2322` because it lacks server services, while `window.strapi.documents`—which cannot exist in the browser—is accepted.

I would move the server augmentation out of the package root into its own opt-in globals entry, and keep the browser augmentation in a separate internal entry. First-party server and admin tsconfigs can then opt into exactly one environment. The browser contract should also be imported from that environment-specific/internal entry rather than the root `Admin` namespace if the root still activates server declarations. A focused type test should assign a valid browser object without suppression and assert that a server-only member such as `documents` is rejected.

---
_AI assisted_

packages/core/types/package.json
```diff
@@ -38,6 +38,12 @@
       "require": "./dist/index.js",
       "default": "./dist/index.js"
     },
+    "./globals-admin": {
+      "types": "./dist/globals-admin.d.ts",
+      "source": "./src/globals-admin.ts",
+      "require": "./dist/globals-admin.js",
```
nclsndr: **Do not advertise a runtime module the package does not build (medium)**

The export map promises `./dist/globals-admin.js` for `require` and `default`, but the code build has only the root Rollup entry and the TypeScript build runs with `--emitDeclarationOnly`. The resulting `dist` contains `globals-admin.d.ts` but no matching JavaScript module, so after a normal `@strapi/types` build, `require('@strapi/types/globals-admin')` deterministically fails with `MODULE_NOT_FOUND`.

Because this augmentation is consumed through `compilerOptions.types`, I would make the internal subpath declaration-only and remove runtime conditions that cannot resolve. If side-effect runtime imports are intentionally supported instead, add it as a real Rollup entry and emit the advertised CJS/ESM files. Please add a package-boundary check that builds `@strapi/types` and verifies every declared export condition against the emitted files (plus a TypeScript resolution check for the types-only path).

---
_AI assisted_

packages/core/types/src/_internal/admin/browser-strapi.ts
```diff
@@ -0,0 +1,26 @@
+import type { Features } from '../modules';
+
+export interface BrowserStrapi {
+  backendURL: string;
+  isEE: boolean;
+  isTrial: boolean;
```
nclsndr, line 6: **Preserve the existing trial-license runtime field (medium)**

Before this change, admin bootstrap populated `window.strapi.isTrialLicense`. This PR replaces that observable field with `isTrial`; therefore an existing admin plugin reading `isTrialLicense` after an EE trial response now receives `undefined` instead of the previous `true`. The absence of first-party readers does not preserve compatibility for third-party admin code, and the PR description currently states that the runtime shape is unchanged.

I would keep `isTrial` as the canonical field but add a deprecated `isTrialLicense` alias to the shared contract for the compatibility window. Initialize both to `false` and update both from the same normalized project-type value, then cover trial and non-trial responses in the bootstrap test. Alternatively, retaining only the old name is the smallest behavioral change; removing it should happen only as an explicit documented breaking migration.

---
_AI assisted_

packages/core/admin/admin/src/render.ts
```diff
@@ -92,18 +88,16 @@ const renderAdmin = async (
       },
     } = await get<{ data: ProjectType }>('/admin/project-type');
 
-    window.strapi.isEE = isEE;
-    window.strapi.isTrialLicense = isTrial;
-    window.strapi.flags = flags;
-    window.strapi.features = {
-      ...window.strapi.features,
+    browserStrapi.isEE = isEE;
+    browserStrapi.isTrial = isTrial;
```
nclsndr: **Keep the CE trial default boolean (medium)**

The CE `GET /admin/project-type` controller currently returns `{ isEE: false, features: [], flags, ai }` and omits `isTrial`. Destructuring that payload produces `undefined`, and this unconditional assignment overwrites the valid local `false` default even though `BrowserStrapi.isTrial` is declared as a required boolean. The typed test mock always supplies `false`, so it does not exercise the production CE response that triggers the mismatch.

I would make the endpoint contract complete by returning `isTrial: false` from the CE controller, and defensively normalize the client assignment (`isTrial ?? false`) so a failed/older response cannot invalidate the browser object. Please add a bootstrap case using the exact CE controller payload and assert that the final `window.strapi.isTrial` (and compatibility alias, if retained) stays `false`.

---
_AI assisted_

docs/docs/docs/01-core/admin/04-features/telemetry.md
```diff
@@ -211,14 +211,26 @@ Do not add per-render or high-frequency tracking without an explicit product nee
...
-window.strapi.telemetryDisabled = process.env.STRAPI_TELEMETRY_DISABLED === 'true';
-window.strapi.projectType = 'Community' | 'Enterprise'; // updated after /admin/project-type
+const browserStrapi: Admin.BrowserStrapi = {
+  telemetryDisabled: process.env.STRAPI_TELEMETRY_DISABLED === 'true',
+  projectType: 'Community',
+  // …other fields
+};
+
+// after fetching /admin/project-type
+browserStrapi.projectType = isEE ? 'Enterprise' : 'Community';
```
nclsndr: **Include Growth in the telemetry example (low)**

The runtime does not derive project type from `isEE` alone: `getProjectType({ isEE, planPriceId })` returns `Growth` when the licensed `planPriceId` contains `growth`, and `Enterprise` for other EE plans. This example drops `planPriceId` and can only produce `Enterprise | Community`, so copying the documented initialization misclassifies every Growth project and no longer matches the declared `BrowserStrapi['projectType']` union.

I would avoid duplicating the classification in the docs: include `planPriceId` in the response example and call the same `getProjectType({ isEE, planPriceId })` helper used by `render.ts`. That keeps Community, Growth, and Enterprise behavior aligned if the classification changes again.

---
_AI assisted_

packages/core/types/package.json
```diff
@@ -38,6 +38,12 @@
       "require": "./dist/index.js",
       "default": "./dist/index.js"
     },
+    "./globals-admin": {
```
nclsndr: **Keep the browser global behind an internal boundary**

Adding `./globals-admin` to the published export map and exposing `Admin.BrowserStrapi` from the root gives standalone consumers a stable-looking import path. That effectively advertises `window.strapi` as supported API, with compatibility obligations around its fields and initialization timing, even though this PR is centralizing private first-party plumbing and generated external projects do not opt into the declaration.

For this scope, I would remove the public subpath/root exposure and place both the contract and augmentation behind an explicit internal, environment-specific entry such as `@strapi/types/_internal/admin` and `@strapi/types/_internal/globals-admin`; first-party admin tsconfigs can opt into that path. `_internal` is still a convention if the file must remain exported for package mechanics, but it clearly marks the surface unsupported and avoids documenting it as plugin API. The globals entry should also be types-only unless a real runtime module is emitted.

---
_AI assisted_

packages/core/types/package.json
nclsndr: For the scope of this PR I think it's wiser to keep the type private. In a follow up we could provide runtime helpers for admin that would properly represent the browser type accessor. What do you think?

nclsndr (changes requested):
Thank you for the great work @unrevised6419 🙌 

Here a first deep AI / shallow human review. Eager to have you opinion.

---

This change crosses the admin boot path, browser/server ambient types, the published `@strapi/types` surface, runtime compatibility, CE license defaults, and telemetry docs. I left inline comments covering initialization order, the browser/server global collision, the unresolved runtime export, the `isTrialLicense` compatibility break, CE's undefined trial value, Growth docs, and keeping the browser-global declaration behind an explicit internal boundary.

---
_AI assisted_

packages/core/types/package.json
unrevised6419: Yes, it can be either via bootstrap function args, or a react hook

packages/core/types/package.json
```diff
@@ -38,6 +38,14 @@
...
       "default": "./dist/index.js"
     },
+    "./_internal": {
+      "types": "./dist/_internal/index.d.ts",
+      "source": "./src/_internal/index.ts"
+    },
+    "./_internal/globals-admin": {
+      "types": "./dist/_internal/globals-admin.d.ts",
+      "source": "./src/_internal/globals-admin.ts"
+    },
```
unrevised6419, line 48: Note: From my understanding the `source` condition is never used in the entire codebase. Probably it was initially for TS [customConditions](https://www.typescriptlang.org/tsconfig/#customConditions)

packages/core/admin/admin/src/utils/getFetchClient.ts
```diff
@@ -280,7 +285,19 @@ type FetchClient = {
  * ```
  */
 const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
-  const backendURL = window.strapi.backendURL;
+  const backendURL = defaultOptions.backendURL ?? window.strapi?.backendURL;
```
nclsndr: **[Medium] Preserve the explicit backend through token refresh**

`getFetchClient({ backendURL })` now sends ordinary requests to the client-specific URL, but a `401` takes the refresh path through `withTokenRefresh()` and `refreshAccessToken()`, which still reads `window.strapi.backendURL`.

With `backendURL: "https://override.test"`, the exact source produced two failure modes: when the global was absent, refresh threw and the original `401` was returned; when the global pointed to `https://global.test`, the observed calls were the original request against `override.test`, the access-token refresh against `global.test`, then the retry against `override.test`. Expected: the original request, refresh, and retry all use the client-selected backend.

The global refresh helper predates this PR, but the new per-client option introduces this split. The added override test covers only a successful ordinary request, while the existing `401` tests use the default global-backed client. Could you ensure the refresh path honors the same backend selection and add an override-plus-`401` regression?

---
_AI assisted_

packages/core/admin/ee/server/src/controllers/admin.ts
```diff
@@ -3,29 +3,32 @@ import { env } from '@strapi/utils';
...
+          isEE: Boolean(strapi.EE),
           isTrial: strapi.ee.isTrial,
           features: strapi.ee.features.list(),
           flags,
-          type: strapi.ee.type,
-          planPriceId: strapi.ee.planPriceId,
+          type: strapi.ee.type ?? undefined,
+          planPriceId: strapi.ee.planPriceId ?? undefined,
           ai: {
             enabled: isAILicense && isAIConfigured.enabled,
```
nclsndr: **[Low] Keep the EE AI payload consistent with the new contract**

`admin.ai = {}` is valid because `enabled` is optional. However, `strapi.config.get('admin.ai', { enabled: isAILicense })` returns the existing empty object rather than filling its nested property, so an active `cms-ai` license makes this expression evaluate to `undefined`. JSON serialization then returns `ai: {}`, even though the new `GetProjectType.Response` and `BrowserStrapi` contracts require `ai.enabled` to be a boolean, and bootstrap copies that malformed object into `window.strapi`.

Reproducing with `admin.ai = {}` and `cms-ai` enabled produced `{"ai":{}}`; expected was `{"ai":{"enabled":true}}`, matching the canonical AI service's leaf-level default. This runtime behavior predates the PR, but the new shared contract formalizes and exposes the mismatch, making this the appropriate place to close it. Could the endpoint guarantee the declared boolean and add an EE controller regression for omitted `admin.ai.enabled`?

---
_AI assisted_

packages/admin-test-utils/package.json
```diff
@@ -96,6 +97,7 @@
   },
   "peerDependencies": {
     "@reduxjs/toolkit": "^1.9.7",
+    "@strapi/types": "^5.0.0",
```
nclsndr: **[Low] Avoid a peer requirement for an erased local type**

The only `@strapi/types` use in this private package is a type-only annotation on the module-local `browserStrapi` fixture. A fresh build emits no runtime import, `dist/setup.d.ts` is exactly `export {};`, and no exported declaration references `@strapi/types`; the development dependency already provides source typechecking and the Nx project edge.

The new peer therefore adds an internal consumer requirement without a consumer-facing type or runtime contract. With the exact workspace install, `yarn explain peer-requirements` reports that consumers provide `@strapi/types@0.0.0-use.local`, which does not satisfy `^5.0.0`; the install succeeds but adds `YN0086` peer-warning noise. Could we avoid introducing this peer requirement unless there is an intended consumer contract that needs it, while retaining the build-time dependency?

---
_AI assisted_

nclsndr (changes requested):
Thanks @unrevised6419 for putting this together and for iterating on the earlier feedback.

Down below is a new review flow capable of exercising actual strapi apps to find gaps. They seem reasonable but I did not reproduced it by hand. I'm interested getting your feedback on the true/false positive findings. 

---

This review covers the admin bootstrap/fetch path, the CE/EE project-type response, the centralized browser declarations, and the related package metadata. I left three inline comments: preserve a client-specific backend through token refresh, keep the EE AI payload consistent with the new boolean contract, and avoid an unnecessary peer requirement in the private admin test utility.

The declaration migration and example-app build otherwise validated successfully; the pre-existing browser/server global intersection remains tracked separately in #26544.

---
_AI assisted_

packages/core/types/package.json
nclsndr: I don't have the history on this but your rationale seems legit to me. 
Actually, I'd love to use such mechanism to avoid building internal packages over the monorepo and getting last version from source all the time. Way easier to work on multiple package in one pass without keeping the build:watch.

package.json
```diff
@@ -145,6 +145,7 @@
     "@swc/core": "1.13.5",
     "@swc/helpers": "0.5.17",
     "@swc/jest": "0.2.39",
+    "@types/eslint": "8.56.12",
```
unrevised6419, line 148: Note: It seems previously the ESLint types were resolved through some transitive dependency, but not anymore. Added back because the ts-check in .eslinrc.cjs files reported that types are missing.

unrevised6419:
@nclsndr please trigger CI again

PS: It seems GitHub again has problems, does not show latest commits I pushed.

unrevised6419:
@nclsndr please trigger CI again, lint should be fixed now.

packages/core/types/package.json
unrevised6419: I recall that `sdk-plugin` [uses](https://github.com/strapi/sdk-plugin/blob/9b7ced948bbd95e1f304278a7abe24f4debd4c31/src/cli/commands/utils/build/vite-config.ts#L85) this field to locate the vite entry and to search for tsconfig file, but not applicable here in the monorepo.

Related: https://github.com/strapi/strapi/pull/27494

nclsndr:
@unrevised6419 
- #27683 just merged if you wanna rebase the work here

packages/core/types/package.json
unrevised6419: > I don't have the history on this but your rationale seems legit to me. Actually, I'd love to use such mechanism to avoid building internal packages over the monorepo and getting last version from source all the time. Way easier to work on multiple package in one pass without keeping the build:watch.

I think that is a task for NX itself 🤔 
https://nx.dev/docs/kb/workspace-watching

unrevised6419:
@nclsndr rebased

packages/plugins/cloud/admin/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "tsconfig/client.json",
-  "include": ["./src", "*.d.ts"],
+  "include": ["./src"],
   "compilerOptions": {
-    "types": []
+    "types": ["vite/client", "@strapi/types/_internal/globals-admin"]
```
nclsndr: **[Medium] Cloud admin opts into `@strapi/types` without declaring it**

`compilerOptions.types` lists `@strapi/types/_internal/globals-admin`.

That entry is at `packages/plugins/cloud/admin/tsconfig.json:5`. Cloud devDependencies start at `packages/plugins/cloud/package.json:56`. i18n declares the same package at `packages/plugins/i18n/package.json:79`.

A package that opts into that types entry declares `@strapi/types` in its own manifest, as i18n and documentation do. Cloud's devDependencies list has no `@strapi/types` entry. Resolution depends on the workspace root hoist.

This pull request adds the types entry. Please make `@strapi/types` resolve from the cloud package, without the workspace root hoist.

---
_AI assisted_

packages/core/types/src/_internal/globals-admin.ts
```diff
@@ -0,0 +1,11 @@
+import type { BrowserStrapi } from './admin';
+
+/**
+ * Opt in from an admin tsconfig via `compilerOptions.types`. Server programs must
+ * not: the package root already declares a `strapi` global as `Core.Strapi`.
+ */
+declare global {
```
nclsndr: **[Medium] No type fixture locks the `window.strapi` assign**

The shared `Window.strapi` contract is new, and the assign sites still carry `@ts-expect-error`.

The declaration is at `packages/core/types/src/_internal/globals-admin.ts:7`. The suppressed assigns are at `packages/admin-test-utils/src/setup.ts:143` and `packages/core/admin/admin/src/render.ts:29`.

A type fixture shows that `BrowserStrapi` assigns to `window.strapi` in an admin-only program, and that server-only members are rejected. Both assign sites suppress the error. No types-package fixture asserts either direction.

This pull request adds the global. Please cover that assign with a type check.

---
_AI assisted_

PR #26216, opened 2026-05-06
chore(eslint): migrate .eslintrc + .eslintignore to .eslintrc.cjs

unrevised6419:
### What does it do?

Migrates every package's ESLint setup from `.eslintrc` (JSON) + `.eslintignore` to a single typed `.eslintrc.cjs` file (177 files).

- Folds `.eslintignore` into `ignorePatterns` inside each rc file.
- Wraps each config in `// @ts-check` + `/** @type {import('eslint').Linter.Config} */`.
- Switches `extends` from the `custom/*` shorthand to the explicit `eslint-config-custom/*` package name.
- Adds `coverage/` to `ignorePatterns` on workspace roots that run jest (tests run from the workspace root, not from `admin/` or `server/` subdirs).
- Adds `.eslintrc.cjs` to `ignorePatterns` on every config so eslint never lints its own config file (this avoids "const is not a keyword" parse errors when eslint defaults to es5 for unmatched files).
- Dedupes two repeated rules in `eslint-config-custom/back/typescript.js` (`node/no-unsupported-features/es-syntax`, `node/no-missing-import`).
- Adds a missing `lint` script to `examples/kitchensink`.
- Adds `parserOptions.requireConfigFile: false` to `examples/kitchensink/.eslintrc.cjs` so vscode-eslint doesn't fail resolving the config when no babel config is present.
- Updates `nx.json` hash inputs:
  - drops `.eslintignore`
  - replaces `{projectRoot}/.eslintrc.js` with `{projectRoot}/**/.eslintrc.cjs`. The old glob missed both nested package configs **and** every package whose config was named `.eslintrc` (JSON, no extension) — those were never tracked in the cache key. The new glob catches all of them.
  - same fix for `tsconfig.eslint.json`.
- Updates `packages/utils/upgrade/tsconfig.eslint.json` include (`.eslintrc.js` → `.eslintrc.cjs`).
- Sets `parserOptions.tsconfigRootDir = __dirname` in every package `.eslintrc.cjs` that uses typed linting. CLI lint resolves tsconfig paths via cwd, but vscode-eslint runs from the workspace root and could not locate per-package tsconfigs without this.

image

- Cleans up two latent `parserOptions.project` paths that only worked when run from a specific cwd: `packages/core/admin/server/.eslintrc` had `["./server/tsconfig.eslint.json"]` and `packages/core/upload/server/.eslintrc` had the same shape. Both now use `["./tsconfig.eslint.json"]` + `tsconfigRootDir: __dirname`, which resolves the same file independently of cwd.

### Why is it needed?

- `.eslintignore` is removed in ESLint v9/v10. Folding ignores into the rc file is the supported migration path and unblocks the eventual ESLint upgrade that will likely ride alongside the in-flight TypeScript 6 migration.
- The `.cjs` extension keeps configs working unambiguously if any package later adopts `"type": "module"`, avoiding CJS/ESM ambiguity.
- The typed JSDoc form gives editor autocomplete and surfaces real mistakes — it caught two duplicate rules during this migration that plain JS / JSON happily accepted.
- Explicit `eslint-config-custom/*` removes ambiguity over what `custom/*` resolves to.
- `coverage/` should not be linted when a workspace runs test coverage.
- `kitchensink` had ESLint configured but no `lint` script, so the config never actually ran.
- Without `tsconfigRootDir`, vscode-eslint reports `parserOptions.project` errors in the editor even though `pnpm -r lint` succeeds, because the editor resolves relative tsconfig paths against the workspace root rather than each package.
- Self-ignoring `.eslintrc.cjs` is the clean fix for the parse error you'd otherwise hit when eslint walks into the config file with default es5 parsing.

### How to test it?

- `pnpm install`
- `pnpm -r lint` — should pass on every package (verified clean locally; 38 nx lint targets, 0 errors).
- Spot-check a single package: `pnpm --filter @strapi/admin lint`.
- `eslint --print-config <file>` from any package should resolve cleanly (verified on `@strapi/admin`, `@strapi/plugin-users-permissions`, `@strapi/upload`).
- Open a `.ts` file from any package in VS Code with the ESLint extension enabled — typed-lint rules should activate without `parserOptions.project` errors.
- Suggested: `nx reset` once after pulling because the `nx.json` hash inputs changed.

### Known gaps

- `packages/core/review-workflows/server` has `tsconfig.json`, `tsconfig.eslint.json`, and `tsconfig.build.json` but no `.eslintrc.cjs`, so its server code is currently not linted. Every other `core/*/server` package has one. Out of scope for this PR (pre-existing); flagging for a follow-up.

### Related issue(s)/PR(s)

Self-initiated. Pairs naturally with the in-progress TypeScript 6 migration: a future ESLint v9/v10 bump becomes a much smaller diff once this lands. No upstream issue tracking it.


unrevised6419:
## Self-review notes

A few things worth flagging that aren't in the PR body:

**Undocumented additions (intentional? please confirm):**
- `env: { es6: true }` added to `core/admin/admin`, `core/admin/ee/admin`, `core/upload/admin` overrides. Redundant for the JS override (already set in `front/index.js`), but meaningful for the TS override since `front/typescript.js` doesn't set `env`. If intentional, worth a line in the description.
- `parserOptions.requireConfigFile: false` added to `examples/kitchensink/.eslintrc.cjs`. Likely needed so `@babel/eslint-parser` doesn't fail without a babel config — confirm and document.

**`coverage/` claim is broader than the change:**
PR body says "Adds `coverage/` to ignorePatterns" but it landed in only ~6 of 100 configs (top-level `core/admin`, `core/strapi`, `core/upload`, `core/content-manager`, `core/content-releases`, `cli/cloud`, `utils/typescript`). All `*/server/*` and `*/admin/*` nested configs are missing it. Either apply consistently or narrow the wording.

**Latent path bug fixed by side effect — worth calling out:**
`core/admin/server/.eslintrc` had `parserOptions.project: ["./server/tsconfig.eslint.json"]`, which only resolved correctly via cwd, not relative to the config file. Same in `core/upload/server`. The new form (`./tsconfig.eslint.json` + `tsconfigRootDir: __dirname`) is the correct shape. Currently undersold in the description.

**Self-ignore inconsistency (minor):**
Most configs add `'.eslintrc.cjs'` to `ignorePatterns`, but several nested ones don't (e.g. `core/admin/admin`, `core/upload/admin`). CLI lint passes today, but if the front override's rules tighten, those config files could start getting flagged. Easy to make uniform.

**Pre-existing gap to track:**
`core/review-workflows/server` has tsconfigs but no `.eslintrc.cjs` — its server code is silently unlinted. Already noted under "Known gaps"; logging here as a follow-up reminder.

Vercel check failure looks unrelated (no Vercel hook in this PR's diff).

unrevised6419:
Pushed `de2d3d5` addressing the self-review notes:

- **`env: { es6: true }` removed** from all 6 admin sub-configs. Root cause was eslint defaulting to es5 when it walked into the config file itself and choked on `const`. Fix is to ignore the config file via `ignorePatterns: ['.eslintrc.cjs']` rather than bumping the env on every consumer — done across every config that was missing self-ignore (16 files).
- **`coverage/` scoped to jest packages.** Restricted to workspace roots that actually run jest (28 configs); removed from 11 non-test packages (`admin-test-utils`, `core/types`, `plugins/cloud`, `utils/logger`, `cli/create-strapi-app`, the four email/upload providers without tests, `check-pr-status`, `kitchensink`). Tests run from the workspace root, so `admin/` and `server/` subdir configs don't need it either. PR description updated to match.
- **`requireConfigFile: false`** kept in `kitchensink`. Documented in the PR description: it's there so vscode-eslint doesn't fail resolving the config when no babel config is present.
- **`nx.json` glob fix is broader than I first wrote.** The old `{projectRoot}/.eslintrc.js` input not only missed nested configs — it also missed every package whose config was named `.eslintrc` (JSON, no extension), which was the majority. Those were never participating in the lint cache key. The new `{projectRoot}/**/.eslintrc.cjs` glob catches all of them. Called out explicitly in the updated PR body.
- **Latent `parserOptions.project` path bug** in `core/admin/server` and `core/upload/server` (only resolved correctly via cwd, not relative to the config file) is now noted in the PR description as part of the change.

Verified with `nx run-many --target=lint --all`: 38 projects, 0 errors.

innerdvations (approved):

PR #26229, opened 2026-05-06
fix(review-workflows): message when single stage

### LINKED ISSUE #21853: Review workflows shows "no permissions" error when there is only one stage

```
$ strapi report
Launched In: 268 ms
Environment: development
OS: darwin-arm64
Strapi Version: 0.0.0-experimental.5d4d8671a4141a722aa70f22022a255988f07137
Node/Yarn Version: yarn/1.22.22 npm/? node/v20.9.0 darwin arm64
Edition: Enterprise
Database: sqlite
### Describe the bug
```

1. Create a workflow with only one stage
2. Edit a content item with that workflow
3. see the error about "no permissions" 

### Expected behavior

I would expect it to say something like "there are no other stages" or just allow it to open a select list with only one item in it

### Screenshots

image

### Code snippets

Adzouz:
### What does it do?

1. When there's only one stage in a review workflow, we used to have a false hint saying that the user doesn't have permission to update the stage:
Screenshot 2026-05-06 at 16 56 43

It was mainly because the list of stages retrieved in the component was empty (same use case for when user doesn't have permissions).

Now the logic handles the possibility of having one stage but a stages array empty to show a different and more accurate message: 
Screenshot 2026-05-06 at 16 50 31

2. Also renamed and moved all translations in review workflows to be scoped to the plugin itself (no more reference to review workflows translations in content manager).

### Why is it needed?

- Show a more accurate message when there's only one stage set in the review workflow.
- Bring consistency and a single source for all translations related to the review workflows plugin.

### How to test it?

ℹ️  EE license + review workflows enabled.

1. Settings → Review Workflows → create a workflow with **one stage**, assign it to a content type. Edit any entry → review stage select is disabled and the hint reads:
  > This workflow only has one stage. Add more stages to be able to transition between them.

2. Create a workflow with **multiple stages**, give a role no transition permission, log in as a user with that role, edit an entry → hint reads:
  > You don't have the permission to update this stage.

3. Multiple stages + transition permission → select populates with the available stages, no hint shown.

4. Switch admin language (FR / DE / ES / etc.) — `Review stage` label still translates correctly (the move to the review-workflows package preserved every existing locale).

### Related issue(s)/PR(s)

Fixes #21853
  
🚀


innerdvations (dismissed):

Adzouz:
Can you review the latest changes and reapprove @innerdvations? I needed to update another test file that was failing 🙏

innerdvations (approved):

PR #24550, opened 2025-10-10
feat(documentation): add /openapi.json route (strapi/documentation#2143) (#20871)

faessler:
### What does it do?

Porting feature _"Exposes the full_documentation.json under the /documentation/openapi.json and /documentation/{version}/openapi.json routes."_ into v5.

Please see https://github.com/strapi/strapi/pull/20871 for more information.

### Related issue(s)/PR(s)

- https://github.com/strapi/documentation/pull/2759
- https://github.com/strapi/strapi/pull/20871
- https://github.com/strapi/documentation/issues/2143
- https://github.com/strapi/documentation/issues/590
- https://github.com/strapi/documentation/issues/366
- https://github.com/strapi/strapi/issues/10734
- https://github.com/strapi/documentation/pull/2159

PR #26239, opened 2026-05-07
feat: add optional openapi spec route

innerdvations:
### What does it do?

Adds core OpenAPI HTTP endpoint support backed by `@strapi/openapi`, with file-based caching and per-endpoint configuration under `server.openapi`.

Key changes:
- Registers OpenAPI routes in core server routing (`register-routes`):
  - `server.openapi['content-api']`
  - `server.openapi.admin`
- Supports per-endpoint config:
  - `enabled`
  - `route.path` (subpath)
  - `access.mode` (`public` or `authenticated`)
  - `access.roles` (optional scopes when `authenticated`)
  - `cache.enabled`, `cache.maxAgeMs`, `cache.filePath`
- Uses file cache on disk (instead of in-memory cache) for generated specs.
- Adds safety/validation:
  - duplicate resolved endpoint path detection
  - invalid access mode rejection
  - rejection when `roles` is set while `access.mode` is `public`
  - route path normalization to avoid accidental double-prefixing
- Adds/update docs in `packages/core/openapi/README.md`.
- Adds unit tests for route registration, access modes, path handling, cache behavior, and validation errors.

### Why is it needed?

We want stable OpenAPI JSON endpoints in Strapi, but aligned with the new core OpenAPI pipeline (`@strapi/openapi`) rather than `@strapi/plugin-documentation`.

This provides:
- a modern and more reliable source of truth for generated specs,
- secure-by-default behavior (disabled by default, authenticated default access mode),
- flexible deployment with explicit endpoint/access/cache configuration,
- support for both content API and admin API specs without requiring future config redesign.

### How to test it?

1. Run unit tests for the new server OpenAPI route behavior:
   - `yarn workspace @strapi/core test:unit -- openapi.test.ts`

2. Configure endpoints in an app (example):
   ```js
   // config/server.js
   module.exports = () => ({
     openapi: {
       'content-api': {
         enabled: true,
         route: { path: '/openapi.json' },
         access: { mode: 'authenticated' }, // or public
         cache: {
           enabled: true,
           maxAgeMs: 60000,
           filePath: '.strapi/openapi/content-api.json',
         },
       },
       admin: {
         enabled: true,
         route: { path: '/openapi.json' },
         access: {
           mode: 'authenticated',
           roles: ['admin::marketplace.read'],
         },
         cache: {
           enabled: true,
           maxAgeMs: 60000,
           filePath: '.strapi/openapi/admin.json',
         },
       },
     },
   });
   ```

3. Start Strapi and verify:
   - content API endpoint is available under API prefix (e.g. `/api/openapi.json`)
   - admin endpoint is available under admin path (e.g. `/admin/openapi.json`)
   - auth behavior matches `access.mode` / `access.roles`
   - cache file(s) are written to configured `cache.filePath`

4. Optional negative checks:
   - set duplicate resolved paths -> app should throw duplicate-path error
   - set invalid `access.mode` -> app should throw validation error
   - set `access.roles` with `mode: public` -> app should throw validation error

### Related issue(s)/PR(s)

- Related: https://github.com/strapi/strapi/pull/24550

### TODO

- [ ] add API tests


packages/core/core/src/configuration/index.ts
```diff
@@ -18,28 +18,60 @@ dotenv.config({ path: process.env.ENV_PATH });
...
+    },
+  },
+  openapi: {
+    'content-api': {
+      enabled: false,
+      route: {
+        path: '/openapi.json',
+      },
+      access: {
+        mode: 'authenticated',
```
jhoward1994, line 48: Would it be better to say this and avoid the type cast at line 74?

I think it would allow us to say `server ... satisfies Partial<Core.Config.Server>,`

```suggestion
          mode: 'authenticated' as const,
```

packages/core/core/src/services/server/openapi.ts
```diff
@@ -0,0 +1,276 @@
+import path from 'path';
+import fs from 'fs-extra';
+import * as openapi from '@strapi/openapi';
+import type { Core } from '@strapi/types';
+
+type OpenAPIRouteType = 'content-api' | 'admin';
+
```
jhoward1994: Maybe we could import these from @strapi/types so they stay in sync

```
type OpenAPIConfig = NonNullable<Core.Config.OpenAPI>;
type OpenAPIEndpointConfig = NonNullable<OpenAPIConfig['content-api']>;
type OpenAPIAccessMode = NonNullable<NonNullable<OpenAPIEndpointConfig['access']>['mode']>;
```

packages/core/core/src/services/server/openapi.ts
```diff
@@ -0,0 +1,276 @@
...
+    absoluteCachePath,
+  };
+};
+
+const getRouteAuthConfig = (config: ResolvedEndpointConfig) => {
+  if (config.accessMode === 'public') {
+    return false as const;
+  }
+
+  if (config.accessRoles.length > 0) {
```
jhoward1994, line 138: We require `scope` to be present whenever auth is not `false` so maybe this is safer than `{}` which could  fail validation?
```suggestion
  return {
    scope: config.accessRoles,
  };
```

packages/core/core/src/services/server/openapi.ts
```diff
@@ -0,0 +1,276 @@
...
+          },
+        },
+      ],
+    };
+  });
+
+  routers.forEach((router) => strapi.server.routes(router));
+};
+
+export const __internal__ = {
```
jhoward1994: Remove this ?

innerdvations:
@jhoward1994 All of your suggestions have been implemented, thanks!

jhoward1994 (dismissed):

innerdvations:
@Bassel17 @jhoward1994 I'm actually rethinking the way I made this just an off/on in the config, and as an alternative, use RBAC permissions for it. What do you think? It could also potentially replace the "enabled" altogether -- it's just inaccessible until permissions are set for a role or token. But if we want to keep the off/on in the config, improving granularity by restricting it to certain users could just be added in a later PR

So my question is: do we want to have the on/off switch, or should I rewrite this PR to -only- use RBAC?

Bassel17:
I would keep the on/off in the config, and we can keep RBAC control as an addition to this

Bassel17 (approved):

PR #26241, opened 2026-05-07
fix(database): move document_id secondary indexes to schema sync

### LINKED ISSUE #25697: Strapi v5 migration 5.0.0-06-add-document-id-indexes fails when index already exists or when index name exceeds PostgreSQL identifier limit

innerdvations:
### What does it do?

Moves document-id secondary index creation out of internal migration `5.0.0-06-add-document-id-indexes` and into model metadata so indexes are created by schema sync.  
Keeps migration `5.0.0-06-add-document-id-indexes` as a stable no-op (same migration name, no DDL).  
Updates tests to cover the new sync-owned index declarations and the migration no-op behavior.

### Why is it needed?

Fixes the fragility reported in #25697 where one-shot migration DDL could fail on duplicate/index-name edge cases and block bootstrap.  
These indexes are schema-level expectations and should be owned by sync, not by a migration that only runs once.

### How to test it?

1. Run targeted tests:
   - `yarn workspace @strapi/database test:unit --testPathPattern='5\\.0\\.0-06-add-document-id-indexes'`
   - `yarn workspace @strapi/core test:unit:vitest -- src/utils/__tests__/transform-content-types-to-models.vitest.test.ts`
2. Verify migration behavior:
   - Confirm `5.0.0-06-add-document-id-indexes` is now a no-op and does not execute index DDL.
3. Verify sync-owned indexes:
   - In `transform-content-types-to-models`, confirm content types declare:
     - single-column `document_id` secondary index
     - one composite index branch based on `locale` / `publishedAt` presence
4. Optional manual bootstrap check:
   - Start Strapi on an upgraded project DB and confirm bootstrap no longer depends on migration 06 DDL for these indexes.

### Related issue(s)/PR(s)

- Fixes #25697
- Fixes CMS-690

### DIFF
--- a/packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
+++ b/packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
@@ -235,7 +235,7 @@ const expectedModels = [
     },
     indexes: [
       {
-        name: 'countries_document_id_index',
+        name: 'countries_documents_index',
         columns: ['document_id'],
       },
     ],
@@ -252,7 +252,7 @@ const expectedModels = [
     },
     indexes: [
       {
-        name: 'categories_document_id_index',
+        name: 'categories_documents_index',
         columns: ['document_id'],
       },
     ],
@@ -343,7 +343,7 @@ describe('transformContentTypesToModels', () => {
       }
     );
 
-    it('adds document_id btree plus full composite when locale and publishedAt are on the model', () => {
+    it('declares a single documents index covering document_id, locale and published_at when present', () => {
       const patched = patchContentTypes('countries', {
         attributes: {
           locale: { type: 'string' },
@@ -353,16 +353,12 @@ describe('transformContentTypesToModels', () => {
       const models = transformContentTypesToModels(patched, identifiers);
       const country = models.find((m) => m.uid === 'api::countries.countries');
 
-      expect(country?.indexes).toEqual(
-        expect.arrayContaining([
-          { name: 'countries_document_id_index', columns: ['document_id'] },
-          {
-            name: 'countries_documents_index',
-            columns: ['document_id', 'locale', 'published_at'],
-          },
-        ])
-      );
-      expect(country?.indexes).toHaveLength(2);
+      expect(country?.indexes).toEqual([
+        {
+          name: 'countries_documents_index',
+          columns: ['document_id', 'locale', 'published_at'],
+        },
+      ]);
     });
   });
 });
--- a/packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
+++ b/packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
@@ -235,7 +235,7 @@ const expectedModels = [
     },
     indexes: [
       {
-        name: 'countries_documents_index',
+        name: 'countries_document_id_index',
         columns: ['document_id'],
       },
     ],
@@ -252,7 +252,7 @@ const expectedModels = [
     },
     indexes: [
       {
-        name: 'categories_documents_index',
+        name: 'categories_document_id_index',
         columns: ['document_id'],
       },
     ],
@@ -342,5 +342,27 @@ describe('transformContentTypesToModels', () => {
         );
       }
     );
+
+    it('adds document_id btree plus full composite when locale and publishedAt are on the model', () => {
+      const patched = patchContentTypes('countries', {
+        attributes: {
+          locale: { type: 'string' },
+          publishedAt: { type: 'datetime' },
+        },
+      });
+      const models = transformContentTypesToModels(patched, identifiers);
+      const country = models.find((m) => m.uid === 'api::countries.countries');
+
+      expect(country?.indexes).toEqual(
+        expect.arrayContaining([
+          { name: 'countries_document_id_index', columns: ['document_id'] },
+          {
+            name: 'countries_documents_index',
+            columns: ['document_id', 'locale', 'published_at'],
+          },
+        ])
+      );
+      expect(country?.indexes).toHaveLength(2);
+    });
   });
 });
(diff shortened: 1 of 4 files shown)


packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
```diff
@@ -235,7 +235,7 @@ const expectedModels = [
     },
     indexes: [
       {
-        name: 'countries_documents_index',
+        name: 'countries_document_id_index',
```
nclsndr: I feel like this could be true 🤔 

### Single-column index rename creates silent duplicate indexes on existing databases

On `develop`, schema sync already writes `{collection}_documents_index` with a **single `document_id` column** for any content type without locale/publishedAt. The PR renames that to `{collection}_document_id_index`.

Schema sync diffs by **name** (see `builder.ts`). Every existing upgraded database will see:

- **Removed:** `countries_documents_index` → `dropIndex` is called
- **Added:** `countries_document_id_index` → `createIndex` is called

`dropIndex` early-returns silently when `forceMigration` is false (the default):

```ts
// packages/core/database/src/schema/builder.ts:205
const dropIndex = (tableBuilder, index, existingIndexes) => {
  if (!db.config.settings?.forceMigration) {
    return; // silent no-op
  }
  ...
};
```

`createIndex` has **no existence guard** — it always fires. So on the default path:
- Old `countries_documents_index` → stays orphaned
- New `countries_document_id_index` → created without issue
- Result: **two single-column indexes on `document_id`**, silently, forever

For users who also hit the old migration 06 DDL (which used `_idx` suffix names like `articles_document_id_idx`), a third overlapping index gets added.

packages/core/core/src/utils/transform-content-types-to-models.ts
```diff
@@ -321,18 +321,43 @@ export const transformContentTypesToModels = (
...
+          name: identifiers.getIndexName([contentType.collectionName, 'documents_locale']),
+          columns: ['documentId', 'locale'].map((n) => identifiers.getColumnName(_.snakeCase(n))),
+        });
+      } else if (hasPublishedAt) {
+        secondaryIndexes.push({
+          name: identifiers.getIndexName([contentType.collectionName, 'documents_published']),
+          columns: ['documentId', 'publishedAt'].map((n) =>
+            identifiers.getColumnName(_.snakeCase(n))
+          ),
+        });
```
nclsndr: It seems those branches are not covered in tests

packages/core/database/src/migrations/internal-migrations/5.0.0-06-add-document-id-indexes.ts
```diff
@@ -1,80 +1,20 @@
...
-      if (!hasTable) {
-        continue;
-      }
-
-      await addIndexesForTable(knex, meta.tableName);
-    }
+  // Keeping the Umzug signature without doing any DDL; callers pass knex/db for other migrations only.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- required by MigrationFn
+  async up(_knex: Knex.Transaction, _db: Database) {
+    await Promise.resolve();
```
nclsndr: Do we need to keep a body for the function here?

nclsndr (commented):
That's a small but tricky one to test! I did run on v5 but did not v4 > v5.

packages/core/core/src/utils/__tests__/transform-content-types-to-models.vitest.test.ts
innerdvations: Good catch, it should be fixed now

nclsndr (approved):
LGTM 🙌

PR #26242, opened 2026-05-07
feat(review-workflows): save prompt for unsaved locales

jhoward1994:
## What does it do?

When switching the Edit View to a locale that hasn't been saved yet, the Review Workflows panel previously resulted in a 404 error on `GET /content-manager/{kind}/{uid}/{id}/stages`, and the StageSelect component showed a misleading "You don't have the permission to update this stage." hint. 

## Why is it needed?

For content types with Review Workflows assigned and i18n enabled, switching the locale picker to a locale that hasn’t been saved led to:

- A 404 error from the stages endpoint, as the controller required an existing entity for the given `documentId` + locale.
- The Review Workflows panel displaying "You don't have the permission to update this stage.", which is misleading—the user has permission, but the entry simply doesn't exist yet for that locale.

This PR ensures the API gracefully handles the case where the locale-specific entry is missing and provides a clear prompt to administrators to save the entry first.

## How to test it?

> **Requires Enterprise Edition (Review Workflows is EE-gated).**

1. Run:
   ```
   yarn install && yarn setup
   ```
2. In the admin, enable i18n on a collection type (e.g., Tag in `examples/getstarted`) and assign a workflow via **Settings → Review Workflows**.
3. Create and save an entry in the default locale.
4. Open the entry, then switch the locale picker to a locale that hasn’t been saved.

**Verify:**

- No 404 error occurs on the `/stages` request—it should return a 200 with `data: []` and `meta: { workflowCount, stageCount }`.
- The Review Workflows stage select is disabled and displays the hint: "Save this entry to assign a workflow stage."
- After saving the locale, the stage select becomes interactable and the hint disappears.

**Automated:**

- `yarn test:front` — covers the new `StageSelect` case (run with `IS_EE=true` by default).
- `yarn test:api --db=sqlite` — covers the new case in `review-workflows.test.api.ts`.

---

## Related issue(s)/PR(s)

- CMS-941


packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/StageSelect.tsx
```diff
@@ -281,14 +285,27 @@ export const StageSelect = ({ isCompact }: { isCompact?: boolean }) => {
...
+
+  // The Edit View renders the Review Workflows panel even when the user
+  // switches to a locale that has not been saved yet. In that case the
+  // document for the selected locale does not exist and we cannot transition
+  // any stage until the entry is first saved, so surface a dedicated hint
+  // rather than the misleading "no permission" message.
+  const hasDocumentForLocale = Boolean(document?.documentId);
+
+  const reviewStageHint = !isLoading
+    ? !hasDocumentForLocale
```
Adzouz: This nested ternary condition is not very readable. Could you rewrite it please? 🙏

packages/core/review-workflows/admin/src/routes/content-manager/model/id/components/StageSelect.tsx
```diff
@@ -281,14 +285,27 @@ export const StageSelect = ({ isCompact }: { isCompact?: boolean }) => {
...
+      ? formatMessage({
+          id: 'content-manager.reviewWorkflows.stage.save-first',
```
Adzouz: I moved the keys recently to the review workflow package instead because it didn't make sense to have them in the content manager, also here would be `stages` instead of `stage` to keep consistency

Adzouz (dismissed):
Hey @jhoward1994, thanks for the PR 🙏 Overall looks good and I tested, you just need to rebase because you have conflicts (I'm sorry I worked on the review workflows too recently: https://github.com/strapi/strapi/pull/26229 and it impacts your logic too). Don't hesitate to reach out to me if you have any question!

Bassel17 (approved):
lgtm!

PR #26244, opened 2026-05-07
fix(strapi): default allowedHosts and pin Vite HMR to main server in dev

### LINKED ISSUE #23491: Blocked Hosts and infinity request of admin on default Port after update from strapi 4 to strapi 5
### Node Version

20.0.0

### NPM/Yarn/PNPM Version

10.0.0

### Strapi Version

5.13.0

### Operating System

Docker/Podman/LXC

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

After the migration from Strapi 4 to Strapi 5 I have some problems with the allowed hosts. 

Image

After setting it in allowedHost in the vite.config.ts, you can work with Strapi, but in the console everything is spammed with errors from the domain + default port of vite. Although we have set a different port in the config.

Image

We use a nginx referse proxy.

Does anyone else have problems with this?

**Vite.config.ts:**
`export default (config) => {
  // Important: always return the modified config
  return mergeConfig(config, {
    resolve: {
      alias: {
        "@": "/src",
      },
    },
    server: {
      allowedHosts: true
    },
  });
};
`

### Steps to Reproduce

1. Use nginx reverse proxy -> follow this official doc https://forum.strapi.io/t/nginx-proxing-with-strapi/40618
2. strapi and nginx are in containers and in the same network

### Expected Behavior

No errors

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

innerdvations:
### What does it do?

- Sets `server.allowedHosts: true` on the bundled admin Vite **development** config so requests behind reverse proxies (nginx, Traefik) are not rejected by Vite's host check.
- Binds HMR to Strapi's main `http.Server` (`ctx.strapi.server.httpServer`) and removes the auxiliary HMR listener and port scan; drops `hmr.clientPort` so the browser uses the same origin as the page instead of an internal port (e.g. 5173).
- Removes unused HMR build context options.
- Adds a unit test asserting `allowedHosts`, HMR `server` wiring, and absence of `clientPort`.

### Why is it needed?

Fix #23491 — behind a reverse proxy only the Strapi port is exposed; the previous separate HMR listener plus `clientPort` caused broken HMR, console errors, and reload issues. `allowedHosts` defaults were also too strict for proxied `Host` headers in middleware mode.

### How to test it?

1. `yarn build` and `yarn test:unit` from the monorepo root (or `cd packages/core/strapi && yarn test:unit --testPathPattern=resolve-development-config`).
2. Optional manual check: run `strapi develop` with `--watch-admin` behind nginx with WebSocket upgrade headers; confirm admin loads and HMR works without custom `vite.config` `allowedHosts` or a second upstream for HMR.

### Upgrade note

If you previously added a custom `src/admin/vite.config.*` workaround for #23491, remove any `server.hmr.clientPort` override when running behind a reverse proxy. Strapi now defaults HMR to the main `http.Server`, but a user config can still intentionally override Vite's `server.hmr` settings. A now-redundant `server.allowedHosts: true` override is harmless, but can be removed.

### Related issue(s)/PR(s)

- Fix #23491


packages/core/strapi/src/node/vite/watch.ts
```diff
@@ -12,75 +10,10 @@ interface ViteWatcher {
...
 const watch = async (ctx: BuildContext): Promise<ViteWatcher> => {
-  const hmrServer = createHMRServer();
-
-  // Allowing Vite to find an available port doesn't work, so we'll find an available port manually
-  // and use that. There is therefore a very slight race condition if you start up two servers at the same time
-  // one might fail, or it might start up but listen on the wrong port.
-  const availablePort = await findAvailablePort(HMR_DEFAULT_PORT);
-  ctx.options.hmrServer = hmrServer;
-  ctx.options.hmrClientPort = availablePort;
+  ctx.options.hmrServer = ctx.strapi.server.httpServer;
```
nclsndr: **Structural cleanup (non-blocking): vestigial `hmrServer` option**

`hmrServer` on `BaseOptions` looks like a real build option, but after this PR it's always the same value: Strapi's `http.Server`, copied into options one line before config reads it back.

Current round trip:
1. `watch.ts` — `ctx.options.hmrServer = ctx.strapi.server.httpServer`
2. `config.ts` — `server: ctx.options.hmrServer`
3. `create-build-context.ts` — `hmrServer?: Server` on `BaseOptions`

Before this PR that indirection made sense: `watch.ts` created its own auxiliary HMR server and injected it via options. Now there's no separate server — so the option is leftover plumbing.

**Suggested simplification** (same behavior):

```ts
// config.ts
hmr: {
  overlay: false,
  server: ctx.strapi.server.httpServer,
},

File: packages/core/strapi/src/node/vite/resolve-development-config.test.ts [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+      bundler: 'vite' as const,
+      options: {
+        open: false,
+        hmrServer: mockHttpServer,
+      },
+      plugins: [],
+      tsconfig: undefined,
+      customisations: undefined,
+      features: undefined,
+    } as unknown as BuildContext;
```
nclsndr, line 38: **Test coverage gap — HMR wiring path**

The new test asserts the right Vite config shape (`allowedHosts`, HMR `server`, no `clientPort`), but it exercises a different input path than production.

In `watch.ts`, the fix wires HMR like this:

```ts
ctx.options.hmrServer = ctx.strapi.server.httpServer;
```

`resolveDevelopmentConfig` then reads `ctx.options.hmrServer` — it never touches `strapi.server` itself.

The test bypasses that bridge by injecting `hmrServer` directly on `options` and stubbing `strapi` without a `server.httpServer`:

```ts
options: { hmrServer: mockHttpServer },
strapi: { internal_config: {} },
```

So the test would still pass if the `watch.ts` assignment were removed or pointed at the wrong server. It validates "options → config", not "strapi httpServer → options → config", which is the actual behaviour change in this PR.

**Suggestion:** mock `strapi.server.httpServer` and omit `options.hmrServer`, or add a thin test around the `watch.ts` resolve path. Also worth dropping `as unknown as BuildContext` if the fixture can be typed against the fields the function actually reads.

Non-blocking — the implementation looks correct; this is about guarding the wiring that makes the proxy fix work.

packages/core/strapi/src/node/vite/watch.ts
```diff
@@ -12,75 +10,10 @@ interface ViteWatcher {
...
-  const hmrConfig = config.server?.hmr;
-
-  // If the server used for Vite hmr is the one we've created (<> no user override)
-  if (typeof hmrConfig === 'object' && hmrConfig.server === hmrServer) {
-    // Only restart the hmr server when Strapi's server is listening
-    strapi.server.httpServer.on('listening', async () => {
-      hmrServer.listen(availablePort);
-    });
-  }
+  const finalConfig = await mergeConfigWithUserConfig(await resolveDevelopmentConfig(ctx), ctx);
```
nclsndr, line 14: User `vite.config` still wins here after Strapi defaults. Leftover `server.hmr.clientPort` in `src/admin/vite.config.ts` can reintroduce #23491 behind a proxy

nclsndr (commented):
Great job @innerdvations 🙌 

Edge case worth evaluating 👇 

### User `vite.config` can still undo the proxy fix — document, don't pin

Not blocking. Flagging for release-note / issue-close messaging.

**Context from review:** Strapi defaults land in `resolveDevelopmentConfig`, then user config wins via `mergeConfigWithUserConfig` in `watch.ts`:

```ts
const finalConfig = await mergeConfigWithUserConfig(await resolveDevelopmentConfig(ctx), ctx);
```

So a `src/admin/vite.config.ts` that still sets `server.hmr.clientPort` or a restrictive `server.allowedHosts` can reintroduce #23491 behind a reverse proxy — even after this PR.

**Why this matters for #23491:** the reporter's workaround was already a custom vite config with `allowedHosts: true`. That fixed the host check but not the HMR spam, because Strapi was still driving a separate listener + `clientPort`. After this PR, users with only `allowedHosts: true` are fine (redundant but harmless). Users who also kept `hmr.clientPort` would still be broken on HMR until they remove that override.

Pre-PR code also explicitly respected user `hmr.server` overrides (it only auto-started the auxiliary HMR server when the user hadn't replaced it). Pinning invariants post-merge — forcing `allowedHosts: true`, `hmr.server = httpServer`, stripping `clientPort` — would break that implicit extension contract for anyone customizing `server.hmr.*` or `allowedHosts` via `src/admin/vite.config.*`. Official docs only show alias examples ([bundlers](https://docs.strapi.io/cms/admin-panel-customization/bundlers#vite)), but the hook accepts full Vite `UserConfig`, and pre-PR behavior treated custom `hmr.server` as intentional.

**Recommendation:** ship as-is + document in release notes / #23491 close comment:

- Strapi dev now defaults `allowedHosts: true` and pins HMR to the main `http.Server` (no separate `clientPort`).
- If you added a custom `vite.config` workaround for #23491, remove `server.hmr.clientPort` and any redundant `server.allowedHosts` when behind a reverse proxy.
- Do **not** post-merge pin those fields — that would be a behavior change for existing vite configs, not just better defaults.

nclsndr:
Otherwise, local development works like a charm 👌

nclsndr (approved):
LGTM 👌

PR #26245, opened 2026-05-07
fix(content-manager): content history crash on deleted relations

Adzouz:
### What does it do?

When you delete a relation from a content type and try to access an existing entry that used to have this relation configured, the content history page throws an error:

https://github.com/user-attachments/assets/a203c383-4242-47e4-aa3a-87db1e334298

We also had an error deleting a component or a dynamic zone:

https://github.com/user-attachments/assets/b33f29a6-a96d-40f1-842b-1b8646a7848f

This PR fixes those two issues.

### Why is it needed?

If a field is removed from a content type, we should still be able to access the content history of all impacted entries.

### How to test it?

- Go to the content manager and select a content type that has a relation
- Create an entry and fill the relation
- Save + Publish
- Go to the content type builder and remove the relation field from the content type
- Save
- Go back to the content manager and select your previously created entry
- Go to the content history
-> You should be able to correctly see the page and versions without any error

Repeat the operation for components and dynamic zones.

🚀

### DIFF
--- a/packages/core/content-manager/admin/src/history/components/tests/VersionInputRenderer.test.tsx
+++ b/packages/core/content-manager/admin/src/history/components/tests/VersionInputRenderer.test.tsx
@@ -14,7 +14,7 @@
 import { Form } from '@strapi/admin/strapi-admin';
 import { render as renderRTL, screen } from '@tests/utils';
 
-import { CustomRelationInput } from '../VersionInputRenderer';
+import { CustomRelationInput, resolveComponentRenderResources } from '../VersionInputRenderer';
 
 import type { RelationsFieldProps } from '../../../pages/EditView/components/FormInputs/Relations/Relations';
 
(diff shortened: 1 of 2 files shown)


packages/core/content-manager/admin/src/history/components/tests/VersionInputRenderer.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+const renderField = (
+  initialFormValues: Record<string, unknown>,
+  attributeOverrides: Partial<RelationsFieldProps['attribute']> = {}
+) =>
+  renderRTL(
+    <CustomRelationInput
+      // @ts-expect-error - test setup uses minimal attribute shape
+      attribute={{ ...baseAttribute, ...attributeOverrides }}
+      label="Categories"
+      name="categories"
```
innerdvations:
```suggestion

type RenderFieldOptions = {
  /** Form field path; must match a key in `initialFormValues` */
  name?: string;
  label?: string;
};

const renderField = (
  initialFormValues: Record<string, unknown>,
  attributeOverrides: Partial<RelationsFieldProps['attribute']> = {},
  options: RenderFieldOptions = {}
) => {
  const name = options.name ?? 'categories';
  const label = options.label ?? 'Categories';

  return renderRTL(
    <CustomRelationInput
      // @ts-expect-error - test setup uses minimal attribute shape
      attribute={{ ...baseAttribute, ...attributeOverrides }}
      label={label}
      name={name}
```
Tiny suggestion: `useField` reads `name`; with `manager` in `initialValues` but `name="categories"` the field value was always undefined, so this wasn’t testing the sanitized admin-user shape

innerdvations (dismissed):
LGTM! just one small test nit

Adzouz:
Whenever you have 2 min to approve this one please @innerdvations 🙏

innerdvations (approved):

PR #26247, opened 2026-05-07
fix(upgrade): remove obsolete plugin strapi-admin.js via codemod

### LINKED ISSUE #21090: [Upgrade Tool] Does not automatically remove the reference to strapi-admin.js file

Strapi Version: 5.0.0-rc.11
Node/Yarn Version: npm/10.2.4 node/v20.11.1 darwin arm64 workspaces/false
Edition: Community
Database: 
OS: 

- Node.js version: node/v20.11.1 darwin arm64
- NPM version:  npm/10.2.4 
- Strapi version: 5.0.0-rc.11
- Database: sqlite
- Is your project Javascript or Typescript: Strapi TS, Plugin JS

### Describe the bug

Based on previous conversation, I am under the impression that `strapi-admin.js` file is not longer required.  

After running the migration tool, it is not automatically removed from the codebase.

Screenshot 2024-08-27 at 6 01 31 PM

After running the upgrade tool I still have the `strapi-admin.js` file.

When creating a plugin using the CLI Tool the `strapi-admin.js` file is removed.

Run the migration command on any plugin.  I used the SEO plugin for this example.

```
npx @strapi/upgrade@rc codemods run --project-path ./src/plugins/seo
```

### Expected behavior

Should remove the `strapi-admin.js` file.

innerdvations:
### What does it do?

Adds a 5.0.0 **`remove-strapi-admin-js`** code codemod that **`unlink`**s `./strapi-admin.js` only when it lives directly under the resolved project directory (supports **`--project-path`**). Threads **`projectRoot: this.cwd`** through **`CodeRunnerConfiguration`** into jscodeshift so transforms never rely on the shell cwd. **`@strapi/upgrade`** unit tests call the transform with **`jest.mock('node:fs')`** and memfs (`vol`), consistent with other upgrade tests.

### Why is it needed?

Plugin upgrades did not drop the legacy Strapi ≤4-style root **`strapi-admin.js`**, unlike the generator, so migrated plugins kept a redundant file (#21090).

Fix #21090

### How to test it?

From the repo root after **`yarn setup`** / **`yarn build`** as needed:

```bash
yarn test:unit packages/utils/upgrade/src/modules/runner/__tests__/remove-strapi-admin-js.test.ts
cd packages/utils/upgrade && yarn test:unit
```

Manual check: **`strapi-upgrade codemods run --project-path <plugin-dir>`** (or **`npx @strapi/upgrade codemods run`** against a built tool) — root **`strapi-admin.js`** disappears when not using **`--dry`**.

### Related issue(s)/PR(s)

Fix https://github.com/strapi/strapi/issues/21090

### DIFF
--- a/packages/utils/upgrade/resources/codemods/5.0.0/remove-strapi-admin-js.code.ts
+++ b/packages/utils/upgrade/resources/codemods/5.0.0/remove-strapi-admin-js.code.ts
@@ -1,0 +1,36 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+import type { Transform } from 'jscodeshift';
+
+type RunnerOptions = {
+  dry?: boolean;
+  projectRoot?: string;
+};
+
+/**
+ * Strapi 5 plugins wire admin via package exports (`strapi/admin`); legacy root `strapi-admin.js` should go.
+ * Only unlinks files directly under `options.projectRoot` (matches `--project-path`).
+ */
+const transform: Transform = (file, _api, options: RunnerOptions) => {
+  const projectRoot = options.projectRoot;
+
+  if (!projectRoot) {
+    return file.source;
+  }
+
+  if (
+    path.basename(file.path) !== 'strapi-admin.js' ||
+    path.normalize(path.dirname(file.path)) !== path.normalize(projectRoot)
+  ) {
+    return file.source;
+  }
+
+  if (!options.dry) {
+    fs.unlinkSync(file.path);
+  }
+
+  return undefined;
+};
+
+export default transform;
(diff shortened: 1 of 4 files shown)


innerdvations:
@PaulBratslavsky would you be able to test this one for me and confirm it fixes your issue?

jhoward1994:
My initial thought is could this annoy people? Even if the file is obsolete and therefore their plugin is currently broken, it might contain code they need and a codemod that just removes it seems a bit dangerous?

innerdvations:
> My initial thought is could this annoy people? Even if the file is obsolete and therefore their plugin is currently broken, it might contain code they need and a codemod that just removes it seems a bit dangerous?

Yeah, you're right. I'll see if I can come up with a different less aggressive approach.

innerdvations:
I think I'm actually just going to close this (and the issue itself as a `won't fix`) -- there's really not a good solution here and it's not a serious problem, more of an inconvenience

PR #26255, opened 2026-05-09
feat(i18n): add locale codes for Abkhazian and Circassian (Adyghe and Kabardian)

abaza738:
### What does it do?

Add locale codes for Abkhazian and Circassian (Adyghe and Kabardian).  
These languages are alive and used in the Caucasus in addition to the diaspora around the world.

Language codes are official ISO codes found in the [ISO 639-2](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) list and [ISO 639-3](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes) list.

### Why is it needed?

Language support for these Caucasian languages.


packages/plugins/i18n/server/src/constants/iso-locales.json
```diff
@@ -1,4 +1,12 @@
 [
+  {
+    "code": "abk",
```
Adzouz: It should be `ab` not `abk`

Adzouz (changes requested):
Hi @abaza738, thanks for the contribution 🙏
One change before merging if you can handle it please (also update your branch with latest develop changes) 😊

abaza738:
@Adzouz should be updated now! Thank you for the review :D

Adzouz (dismissed):
Good now thanks for the quick response! 🙏

Adzouz:
Ah can you run `yarn prettier --write packages/plugins/i18n/server/src/constants/iso-locales.json` and `yarn jest packages/plugins/i18n --config jest.config.js -u` please?

Adzouz (approved):

PR #26265, opened 2026-05-11
enhancement(strapi): lazy-load TypeScript chain for non-build CLI commands

Aurelsicoko:
### What does it do?

Three small, independent changes that defer TypeScript-related requires for CLI commands that don't actually use TypeScript:

1. **`packages/core/strapi/src/cli/index.ts`** — `ctx.tsconfig` becomes a memoised getter. The first read calls `loadTsConfig`, which transitively requires the `typescript` package. Until something reads `ctx.tsconfig`, `typescript` never loads. Existing destructuring (`{ tsconfig } = ctx`) and dotted access (`ctx.tsconfig`) keep working unchanged — only the load is deferred.

2. **`packages/core/strapi/src/cli/commands/start.ts`** — `start` previously called `tsUtils.isUsingTypeScript(appDir)` and `tsUtils.resolveOutDir(appDir)` to derive `distDir`. Both pull in `@strapi/typescript-utils` (~150 ms), but the answer is observable on disk:

   - no `tsconfig.json` → JavaScript project, `distDir = appDir`
   - `dist/src/index.js` exists → conventional TS layout, `distDir = dist`
   - otherwise → fall back to `tsUtils.resolveOutDir` (custom `outDir` or unbuilt project) so the existing error message and error path are preserved.

3. **`packages/core/strapi/src/cli/commands/ts/generate-types.ts`** — `@strapi/typescript-utils` and `@strapi/core` were imported at the module top, but `cli/commands/index.ts` instantiates every command factory at startup, so those imports fired for every CLI invocation (not only `ts:generate-types`). Move them into the `action` body, matching the pattern `build` and `develop` already use.

### Why is it needed?

Boot-time profiling on a fresh `npx create-strapi-app@5.45 --quickstart` shows:

- `require('typescript')` via `tsconfig.ts`: **~115 ms** on every CLI command (most commands never read tsconfig).
- `require('@strapi/typescript-utils')` via `start.ts`: **~150 ms** on every `strapi start` (used only to compute `distDir`, which is observable on disk).
- `require('@strapi/typescript-utils')` + `@strapi/core` via `ts/generate-types.ts`: **~200 ms** on every other CLI command (the command factory is instantiated even when not invoked).

Combined upper bound on a `strapi start`: ~465 ms of work for libraries the command never uses. Behaviour is unchanged when a command actually does need `typescript` / `tsUtils` — the require still happens, just on first access.

This is part of a small series of boot-perf PRs; see #26264 (PR-1: `tsbuildinfo` cleanup fix) for the first one.

### How to test it?

Functional regression checks:

```bash
# 1. Provision a fresh Strapi 5 project
npx --yes create-strapi-app@latest cli-perf --quickstart --no-run
cd cli-perf

# 2. Every CLI command should still behave identically
yarn strapi --help                # commander tree intact
yarn strapi start                 # boots normally on a built project
yarn strapi build                 # `ctx.tsconfig` is read, getter triggers tsconfig load
yarn strapi develop               # full TS pipeline still runs
yarn strapi ts:generate-types     # types written to types/generated/

# 3. JS project regression check
mv tsconfig.json tsconfig.json.bak
yarn build && yarn start          # uses appDir as distDir; no error
mv tsconfig.json.bak tsconfig.json

# 4. Custom outDir regression check
# Edit tsconfig.json: change `compilerOptions.outDir` to `./build-out`
yarn build && yarn start          # falls back to tsUtils.resolveOutDir; no error
```

Boot-time benchmarks (best of 10, warm fs, macOS 14, M-series, Node 20.20.2, sqlite quickstart):

|                            | before  | after   | delta    |
|----------------------------|--------:|--------:|---------:|
| `strapi start` (warm)      | ~2.5 s  | ~2.1 s  | −400 ms  |
| `strapi develop` (warm)    | ~5.5 s  | ~5.2 s  | −300 ms  |
| `strapi --help`            |  ~600 ms|  ~280 ms| −320 ms  |

### Related issue(s)/PR(s)

- #26264 — fix(strapi): preserve tsbuildinfo across develop restarts (PR-1 of the same series).

### DIFF
--- a/packages/core/strapi/src/cli/commands/start.ts
+++ b/packages/core/strapi/src/cli/commands/start.ts
@@ -6,31 +6,63 @@ import { createStrapi } from '@strapi/core';
 import type { StrapiCommand } from '../types';
 import { runAction } from '../utils/helpers';
 
+const DEFAULT_OUT_DIR = 'dist';
+
+/**
+ * Read `compilerOptions.outDir` from tsconfig.json without loading `typescript` or
+ * `@strapi/typescript-utils`. Returns null when the file cannot be parsed or when
+ * `extends` may define outDir (fall back to resolveOutDir).
+ */
+const tryQuickOutDir = (appDir: string, tsconfigPath: string): string | null => {
+  let raw: unknown;
+  try {
+    raw = JSON.parse(fs.readFileSync(tsconfigPath, 'utf8'));
+  } catch {
+    return null;
+  }
+
+  if (!raw || typeof raw !== 'object') {
+    return null;
+  }
+
+  const config = raw as { extends?: string; compilerOptions?: { outDir?: string } };
+  const localOutDir = config.compilerOptions?.outDir;
+
+  if (config.extends && localOutDir === undefined) {
+    return null;
+  }
+
+  return path.resolve(appDir, localOutDir ?? DEFAULT_OUT_DIR);
+};
+
 const action = async () => {
   const appDir = process.cwd();
   const tsconfigPath = path.join(appDir, 'tsconfig.json');
-  const defaultDist = path.join(appDir, 'dist');
 
   let distDir: string;
 
   if (!fs.existsSync(tsconfigPath)) {
     distDir = appDir;
-  } else if (fs.existsSync(path.join(defaultDist, 'src', 'index.js'))) {
-    // Default outDir already built — skip loading `@strapi/typescript-utils`
-    distDir = defaultDist;
   } else {
-    // Custom outDir or unbuilt project — fall back to the slow correct path
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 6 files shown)


innerdvations:
I pushed a small follow-up on the `start` fast path: instead of assuming `dist/`, it reads `compilerOptions.outDir` from `tsconfig.json` (still without loading `typescript-utils`) and checks `<outDir>/src/index.js`. If the config uses `extends` without a local `outDir`, or the file can't be parsed, it falls back to `resolveOutDir()` like before.

Same perf for the default case, but custom `outDir` won't accidentally pick up a stale `dist/` folder. Does that approach make sense to you?

innerdvations (dismissed):

innerdvations (approved):

PR #26267, opened 2026-05-11
enhancement(core): lazy-load node-schedule and umzug at boot

Aurelsicoko:
### What does it do?

Defers two boot-time requires that are only used at runtime, not at module load:

1. **`packages/core/core/src/services/cron.ts`** — `node-schedule` becomes a lazy require. The `Job` class is now resolved on first cron task addition (`add()` / `start()`). Type-only `import type { Job, Spec }` keeps the existing `interface JobSpec` intact.

2. **`packages/core/database/src/migrations/{internal,users}.ts`** — both providers used to call `new Umzug({...})` synchronously inside their factory. The `Umzug` constructor (and its inquirer / `@rushstack/ts-command-line` chain) now lives inside a lazy `provider()` closure that only fires on `shouldRun()`, `up()`, or `down()`.

### Why is it needed?

Boot-time profiling on a fresh Strapi 5 project shows:

- `node-schedule`: ~90 ms via `cron.ts`, even when `server.cron` is empty (the default for fresh projects).
- `umzug`: ~100 ms total via the two migration providers — the constructors run on every boot, including projects with no migration files.

Both modules pull in heavy transitive dependencies that aren't needed until the corresponding feature is exercised. Same lazy-fn pattern as #26266.

Part of the boot-perf series:
- #26264 — fix(strapi): preserve tsbuildinfo across develop restarts (PR-1)
- #26265 — enhancement(strapi): lazy-load TypeScript chain for non-build CLI commands (PR-2)
- #26266 — enhancement(core/core): lazy-load typescript-utils in Strapi and compile (PR-3)

### How to test it?

```bash
# 1. Default project (no cron, no user migrations)
yarn develop  # node-schedule and umzug never load

# 2. Project with cron tasks
# config/cron-tasks.ts: export default { '* * * * *': () => {} }
# config/server.ts: cron: { enabled: true }
yarn develop  # cron.ts -> ns().Job triggers the require on add()

# 3. Project with pending migrations
# database/migrations/2026-05-10-foo.js
yarn develop  # internal/users provider() triggers require on shouldRun()
```

Cron jobs and migrations behave identically to the eager-import baseline; only the timing of the require changes.

### Related issue(s)/PR(s)

- #26264, #26265, #26266 — same boot-perf series.

### DIFF
--- a/packages/core/core/src/services/cron.ts
+++ b/packages/core/core/src/services/cron.ts
@@ -1,6 +1,16 @@
-import { Job, Spec } from 'node-schedule';
+import type { Job, Spec } from 'node-schedule';
 import { isFunction } from 'lodash/fp';
 import type { Core } from '@strapi/types';
+
+// Lazy: only required when a cron task is actually scheduled
+let lazyNs: typeof import('node-schedule') | undefined;
+const ns = (): typeof import('node-schedule') => {
+  if (!lazyNs) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    lazyNs = require('node-schedule');
+  }
+  return lazyNs as typeof import('node-schedule');
+};
 
 interface JobSpec {
   job: Job;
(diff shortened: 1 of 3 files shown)


innerdvations (approved):
Tested and idea/code LGTM!

PR #26270, opened 2026-05-11
enhancement(core): eliminate @strapi/typescript-utils from boot path

Aurelsicoko:
### What does it do?

Even with the lazy-fn pattern landed by #26266 (`Strapi.ts` + `compile.ts`), tracing the boot of a fresh project shows **17 typescript-related module loads** still firing at boot. The remaining loads come from three independent paths, all of which fire even when **telemetry is disabled and TypeScript migrations are off** (the default for fresh projects).

Surgical fixes, one per leak:

1. **`packages/core/core/src/services/metrics/sender.ts`** — top-imports `@strapi/typescript-utils` for two `isUsingTypeScriptSync` calls. Even importing the file loads tsUtils. Convert to the lazy-fn pattern so loading the file is free.
2. **`packages/core/core/src/services/metrics/index.ts`** — calls `createSender(strapi)` and wraps it with `wrapWithRateLimit` unconditionally. Gate both behind `!isDisabled`. When telemetry is off, the sender (and its tsUtils consumer) never construct.
3. **`packages/core/admin/server/src/controllers/admin.ts`** — destructures `{ isUsingTypeScript } = tsUtils` at module top. Convert to a lazy wrapper that resolves the require on first call. The function is only consumed by `GET /admin/project-type`.
4. **`packages/core/core/src/Strapi.ts`** — calls `tsUtils.resolveOutDirSync(...)` in the db container factory even when `useTypescriptMigrations` is the default `false`. Hoist the flag check; resolve `tsUtils` only when the flag is true.

### Why is it needed?

Measured on the same fresh quickstart used in the rest of the boot-perf series (interleaved A/B, 12 runs each, `pnpm dev:fast`, macOS 14 M-series, Node 20.20.2, sqlite, warm fs):

- typescript-utils loads at boot: **17 → 0**
- median wall-clock: **1803 ms → 1696 ms (−107 ms)**
- σ_on = 78 ms · σ_off = 107 ms — **|delta| > (σ_on + σ_off) / 2, above noise**

This is the first individual change in the perf series with a wall-clock saving that survives noise on this hardware. Pairing several smaller lazy-load fixes into one PR (rather than landing them one at a time) is what makes the saving observable.

Part of the boot-perf series:
- #26264, #26265, #26266, #26267, #26268, #26269 (PR-1..7 of the series)

### How to test it?

```bash
# 1. Default project (telemetry disabled in dev quickstart)
yarn develop  # tsUtils never loads anywhere on the boot path
node -e "
  const tsUtils = require('@strapi/typescript-utils');
  console.log('typescript-utils loaded eagerly?', !!tsUtils);
"  # still works when explicitly required

# 2. Telemetry enabled (real project)
# config/server.ts: telemetry: { enabled: true } + valid uuid
yarn develop  # createSender runs; tsUtils() resolves on first call

# 3. TS migrations enabled
# config/database.ts: settings: { useTypescriptMigrations: true }
yarn develop  # tsUtils() resolves in db factory; tsDir computed correctly

# 4. Admin /project-type endpoint
yarn develop
curl http://localhost:1337/admin/project-type  # returns isTypescript: true|false; tsUtils loads on first call
```

### Notes for review

- This PR's `Strapi.ts` change overlaps with #26266 (the lazy-fn). If #26266 lands first, this PR's `Strapi.ts` diff reduces to just the conditional `useTSM` gate.
- `metrics/middleware.ts` is unchanged — it already gates on `sendEvent` being callable.

### Related issue(s)/PR(s)

- #26264, #26265, #26266, #26267, #26268, #26269 — same boot-perf series.


innerdvations (dismissed):
Let's merge #26266 first and then resolve the conflict here

innerdvations (approved):

PR #26271, opened 2026-05-11
enhancement(core/utils): lazy-load date-fns in parse-type

Aurelsicoko:
### What does it do?

Converts `packages/core/utils/src/parse-type.ts` from an eager `import * as dates from 'date-fns'` to a memoised lazy require. The first call site (parsing a `date` / `time` / `timestamp` value) triggers the load; the require pattern matches the lazy-fn approach used by the other PRs in this series.

`import type * as datesType from 'date-fns'` keeps the compile-time signature.

### Why is it needed?

`@strapi/utils` is required transitively during Strapi's boot path; consumers like `services/server`, `core-api/routes/validation`, and `database/fields` all pull from the barrel. `parse-type.ts` is loaded as part of that chain, and it eagerly loads `date-fns` — which itself pulls in a chunk of transitive submodules for the date helpers it uses (`isDate`, `format`, `parseISO`, `isValid`, `parse`).

On a fresh Strapi 5 quickstart, `require('date-fns')` adds ~50 ms to the boot path. Most boots never call `parseType` for a date value before the server is listening — the cost is paid for nothing. Deferring the require to first call removes that.

### Notes

- This is a narrower take than the original boot-perf plan, which proposed converting the entire `@strapi/utils` barrel to self-replacing property getters. That approach targets the rolled-up CJS dist directly and is more involved to express cleanly in TypeScript ESM source. Doing the heaviest single submodule first captures the bulk of the value with a 1-file diff.
- Other heavy submodules in the barrel (validators / yup / zod / sanitize) could get the same internal-lazy treatment in follow-up PRs if desired.

Part of the boot-perf series:
- #26264, #26265, #26266, #26267, #26268, #26269, #26270 (PR-1..9 of the series)

### How to test it?

```bash
# 1. Boot a fresh project; date-fns shouldn't load until parseType runs on a date
yarn develop
# date-fns will load on first request that hits a date / time / timestamp field

# 2. Existing parse-type tests still pass
yarn test:unit -- packages/core/utils
```

### Related issue(s)/PR(s)

- #26264, #26265, #26266, #26267, #26268, #26269, #26270 — same boot-perf series.


innerdvations:
On this one, I don't think there's much value unless we also lazy-load `date-fns` in `@strapi/database` (`fields/shared/parsers.ts`, `datetime.ts`, `timestamp.ts`) because on a normal boot those still load it first, before `parse-type.ts` ever runs, basically the same situation you addressed for `typescript-utils` in #26270. Could you update it to include that, or would you like me to just push the fix directly?

innerdvations:
I pushed a follow-up commit that lazy-loads `date-fns` in the database field parsers and datetime/timestamp fields too, let me know if that makes sense

PR #26279, opened 2026-05-11
Improve schema sync performance

Baldinof:
### What does it do?

Update schema inspector to replace per-table N+1 queries (`getColumns`, `getIndexes`, `getForeignKeys` called individually for each table) with bulk queries using `ANY()` operator. 

This fetches columns, indexes, and foreign keys for all tables in 3 parallel queries instead of 3*N sequential ones.

### Why is it needed?

Improve bootstrap performance when syncing DB schema for large schemas.

On a remote Postgres, with the `getstarted` example, `strapi.db.schema.sync()` goes from ~7s to ~1s

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/core/database/src/dialects/mysql/schema-inspector.ts
+++ b/packages/core/database/src/dialects/mysql/schema-inspector.ts
@@ -7,6 +7,7 @@ interface RawTable {
 }
 
 interface RawColumn {
+  table_name: string;
   data_type: string;
   column_name: string;
   character_maximum_length: number;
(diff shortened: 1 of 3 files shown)


nclsndr (commented):
I tested the change with SQLite, Postgres, MySQL. All working well.
@innerdvations do you see any reason to not move forward?

innerdvations (approved):
Great idea, thanks! LGTM!

Only nit is that you could delete mysql's unused `getColumns` like you did for postgres

PR #26283, opened 2026-05-11
ci: fix community PR report workflow

butcherZ:
## What does it do?

- Removes the `setup-node` / `npm ci` / `npm run build` steps from the weekly report workflow — the action dist is now pre-bundled (committed), so these steps are unnecessary and were causing the workflow to fail with a missing `package-lock.json` error
- Adds a `##` markdown heading to the weekly report body so Linear renders a visible title in the activity feed

## Why is it needed?

The community-label workflow had its build steps removed in #26225, but the report workflow was missed. It has been failing every Monday since with `npm error The npm ci command can only install with an existing package-lock.json`.

## How to test it?

Trigger the workflow manually via `workflow_dispatch` on the `Community PR — Weekly Report` workflow and verify it completes successfully and the posted Linear update has a heading.

## Related issue(s)/PR(s)

- #26225


Adzouz (dismissed):
Approved chief, let's go 🚀

Adzouz (approved):
🚀

PR #26291, opened 2026-05-11
chore(deps): hoist @types/node to root and align with 20, min supported engine

unrevised6419:
### What does it do?

- Pins `@types/node` to exact `20.19.41` across **7** workspace `package.json` files: `.github/actions/community-pr-triage`, `examples/complex`, `examples/empty`, `packages/core/core`, `packages/core/data-transfer`, `packages/core/strapi`, `packages/core/utils`. The four core packages were pinned at `24.10.0`; the action and two examples used `^20`/`^20.0.0`.
- Adds `@types/node@20.19.41` to root `package.json` `resolutions` so the whole workspace resolves a single Node 20 type tree.
- Cleans `yarn.lock`: drops the `@types/node@24.10.0` resolution and the now-orphan `undici-types@~7.16.0`; collapses the `^20`/`^20.0.0`/`24.10.0` entries into a single `@types/node` resolution at `20.19.41`.
- In `packages/core/data-transfer/src/utils/encryption/{encrypt,decrypt}.ts`, replaces the named `Cipheriv` / `Decipheriv` imports with local type aliases:

  ```ts
  // @types/node 20.x names this class Cipher; 24.x renames it to Cipheriv.
  // Derive via ReturnType so signatures hold across both.
  type Cipheriv = ReturnType<typeof createCipheriv>;
  type Decipheriv = ReturnType<typeof createDecipheriv>;
  ```

  Call sites and exported signatures are unchanged.

#### Side effect: aligns the undici types across Node versions

`@types/node` bundles `undici-types` as a dependency, and each major pins a different one — Node 20's tree pulls `undici-types@~6.21.0`, while the 24.x tree (plus a stray direct entry) pulled `undici-types@~7.16.0`. With both present, undici-derived types (`Dispatcher`, `ProxyAgent`, `RequestInit`, `Response`) resolved to two incompatible declarations depending on which tree a package saw — the exact "conflicting node 20 and 24 undici types" problem.

[...]

PR #26293, opened 2026-05-11
chore: add knip for dead-code and unused-dependency detection

unrevised6419:
> [!NOTE]
> Based on top of #26291. Merge that first; the diff here will shrink once it lands.

### What does it do?

Introduces [knip](https://knip.dev) as a monorepo-wide linter for unused files, exports, and dependencies, and wires it into CI.

- Adds `knip.config.ts` configured for every workspace (CLI, core packages, plugins, providers, utils, generators).
- Adds a `yarn knip` script and a `knip-lint` CI job that runs on the existing Node 20/22/24 matrix.
- Removes the 16 dead modules and ~25 unused devDependencies that the initial run surfaced — already audited, no runtime imports remained.
- Regenerates `yarn.lock` to drop the transitive cruft (-2.7k lines net in the lockfile).

Dead files removed include:

- `core/admin/admin/src/hooks/useLicenseLimitNotification.ts`
- `core/admin/admin/src/pages/InternalErrorPage.tsx`
- `core/admin/ee/.../AuditLogs/utils/tableHeaders.ts`
- `core/admin/server/src/config/settings.ts`
- `core/content-type-builder/admin/src/components/AIChat/{components/ImagePreviewGroup,lib/fileToAttachment}.ts(x)`
- `core/content-type-builder/admin/src/components/IconPicker/index.ts`
- `core/content-type-builder/admin/src/utils/startsWithANumber.ts`
- `core/content-type-builder/server/src/controllers/ai/{auth,index}.ts` (already empty stubs)
- `core/content-type-builder/server/src/utils/helpers.ts`
- `core/core/src/services/document-service/transform/relations/utils/data.ts`
- `core/review-workflows/admin/src/routes/content-manager/model/configure/constants.ts`
- `core/strapi/src/cli/commands/plugin/init/files/typescript.ts`
- `core/strapi/src/cli/utils/pkg.ts`
- `core/strapi/src/load/package-path.ts`
- `core/types/src/modules/documents/document-engine.ts` (empty)
- `core/upload/admin/src/future/services/settings.ts`
- `core/upload/server/src/models/index.ts`
- `plugins/i18n/server/src/models/index.ts`
- `plugins/i18n/server/src/services/permissions/conditions.ts`

Unused devDependencies pruned from root and individual packages include `@swc/cli`, `@types/git-url-parse`, `chokidar`, `commander`, `jest-circus`, `minimatch`, `plop`, `qs`, `ts-jest`, and others — none referenced in source or tooling.

### Why is it needed?

**Maintenance burden.** Every dead file is a trap: reviewers waste cycles understanding code that no caller exercises, IDE search noise grows, and refactors carry phantom modules along. Knip surfaces these mechanically instead of relying on ad-hoc spelunking.

**Dependency hygiene.** Unused devDependencies still resolve, still ship in `yarn install`, still need security review, and still constrain version bumps elsewhere in the tree. Removing the ~25 unused entries shrinks the lockfile by ~2.7k lines and tightens the surface that Dependabot/Renovate has to babysit.

**Supply-chain risk.** Each declared dependency — even devDependencies — is an inbound trust boundary. Fewer entries = fewer transitive packages installed in CI and on contributor machines = smaller blast radius for the next `event-stream`/`ua-parser-js`-style incident.

**Install/CI speed.** Less to fetch, less to extract, less to hoist. The lockfile reduction translates directly to faster cold installs in CI on every PR.

**Prevents regression.** The new `knip-lint` job runs on PRs in the same Node matrix as the rest of CI, so reintroducing a dead export or stray dependency fails the build instead of silently accumulating. The config is intentionally conservative (`tags: ['-knipignore']`, `exclude: ['exports','types','namespaceMembers','enumMembers']`) so it only flags high-confidence wins today — these categories can be enabled later in follow-up PRs as workspaces are cleaned up.

**Low risk.** All removals are inert: no runtime imports, no test imports, no plugin entry points. CI's existing unit, API, e2e and type-check jobs gate the change end-to-end.

### How to test it?

```bash
yarn install
yarn build
yarn knip          # passes clean on this branch
```

Then verify the new CI job (`knip-lint (node: 20|22|24)`) succeeds, and that the existing `unit_back`, `unit_front`, `api`, `e2e` and type-check jobs are unaffected.

To confirm a guard is in place: re-introduce any of the deleted files (e.g. `git checkout HEAD~1 -- packages/core/admin/admin/src/pages/InternalErrorPage.tsx`) and run `yarn knip` — it should fail with an "unused file" error.

### Related issue(s)/PR(s)

None.

---
Fixes CPR-396


packages/core/admin/package.json
```diff
@@ -152,31 +136,23 @@
...
-    "@types/markdown-it-container": "2.0.9",
-    "@types/markdown-it-emoji": "2.0.4",
-    "@types/markdown-it-footnote": "3.0.3",
     "@types/passport-local": "1.0.36",
-    "@types/pluralize": "0.0.32",
     "@types/punycode": "2.1.4",
-    "@types/react-window": "1.8.8",
-    "@types/sanitize-html": "2.13.0",
-    "@vitejs/plugin-react-swc": "3.6.0",
+    "eslint-config-custom": "5.45.1",
```
unrevised6419: Note: I would recommend this kind of dependencies to resolve to `workspace:`

packages/core/admin/package.json
```diff
@@ -152,31 +136,23 @@
...
     "koa-body": "6.0.1",
     "msw": "2.13.4",
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-router-dom": "6.30.3",
     "styled-components": "6.1.8",
-    "vite": "5.4.21",
-    "vite-plugin-dts": "^4.3.0"
+    "tsconfig": "5.45.1",
```
unrevised6419: Note: I would recommend this kind of dependencies to resolve to `workspace:`

packages/core/core/package.json
```diff
@@ -100,21 +96,16 @@
...
     "semver": "7.7.4",
     "statuses": "2.0.1",
     "typescript": "5.4.5",
     "undici": "6.25.0",
-    "yup": "0.32.9",
     "zod": "3.25.67"
   },
   "devDependencies": {
-    "@strapi/ts-zen": "^0.2.0",
-    "@swc/core": "1.13.5",
```
unrevised6419, line 121: Note: `@swc/core` is already added in the root with other swc dependencies.

packages/core/core/package.json
```diff
@@ -100,21 +96,16 @@
...
     "resolve.exports": "2.0.2",
     "semver": "7.7.4",
     "statuses": "2.0.1",
     "typescript": "5.4.5",
     "undici": "6.25.0",
-    "yup": "0.32.9",
     "zod": "3.25.67"
   },
   "devDependencies": {
-    "@strapi/ts-zen": "^0.2.0",
```
unrevised6419, line 120: Note: No traces of anything using this dependency across the monorepo.

packages/core/data-transfer/types/remote/protocol/client/commands.d.ts
unrevised6419, line 1: Note: I would recommend to not use *.d.ts file for type definitions, they are not type checked. For simple external module augmentations is fine.

packages/core/review-workflows/server/src/controllers/index.ts
```diff
@@ -1,5 +1,3 @@
-import type {} from 'koa-body';
```
unrevised6419, line 1: Note: Such augmentations should be done using one of these: 

- `tsconfig.json#compilerOptions.types`
- `env.d.ts` / `import 'koa-body'`
- Triple Slash Directive `/// <reference types="koa-body" />`

unrevised6419: PS: No errors without it.

packages/core/upload/package.json
```diff
@@ -32,13 +32,6 @@
       "require": "./dist/admin/index.js",
       "default": "./dist/admin/index.js"
     },
-    "./_internal/shared": {
-      "types": "./dist/shared/index.d.ts",
-      "source": "./shared/index.ts",
-      "import": "./dist/shared/index.mjs",
-      "require": "./dist/shared/index.js",
-      "default": "./dist/shared/index.js"
-    },
```
unrevised6419, line 41: Note: Export never used across monorepo

knip.config.ts
```diff
@@ -0,0 +1,256 @@
...
+    'tsconfig.build.json',
+    'tsconfig.eslint.json',
+    'admin/tsconfig.json',
+    'admin/tsconfig.build.json',
+    'admin/tsconfig.eslint.json',
+    'server/tsconfig.json',
+    'server/tsconfig.build.json',
+    'server/tsconfig.eslint.json',
+  ],
+};
```
unrevised6419, line 19: Note to myself. Here we need to add also `ee/{admin,server}/tsconfig*.json` files

innerdvations:
Thanks for this — much more polished than the internal spike we just closed (#26793). Two things worth considering as knip usage expands:

**1. Guard before `--fix`**

If we ever add a `knip --fix-type dependencies` (or similar) script, a small pre-flight check on sentinel workspaces would prevent a misconfigured entry pattern from wiping real deps. We hit this locally on a first `--fix` run: with only `.ts` entry globs, knip flagged every `@strapi/plugin-users-permissions` dependency as unused and would have emptied `package.json`. A guard like `scripts/knip-deps-guard.mjs` that asserts known deps (e.g. `bcryptjs`, `jsonwebtoken` on users-permissions) are *not* reported unused before fix runs would make auto-fix safe. Happy to contribute that as a follow-up if useful.

**2. JS/JSX plugin entry coverage**

`users-permissions` is configured with `server/utils/index.d.ts` as an entry, but the admin side is `admin/src/index.js` with a tree of `.jsx` files. For the current conservative scope (files + devDeps, exports/types excluded) this is probably fine. If we later add production dependency analysis (`--production --dependencies`), worth explicit entries like `admin/src/index.{js,jsx,ts,tsx}!` and matching project globs for JS-only plugins — otherwise prod deps used only from the admin bundle can false-positive as unused.

Neither is blocking for this PR as-is; both are forward-looking for phase 2.

_AI-assisted comment._

PR #26295, opened 2026-05-12
fix(content-manager): guard getMainField when component schema is missing

### LINKED ISSUE #26275: bug report on single type navigation, cant read property err
When navigating between Single Types using the admin navigation links, the admin panel sometimes crashes with the following error:

Cannot read properties of undefined (reading 'attributes')

This only happens during client-side navigation between Single Types.
If the page is refreshed manually, the error disappears and the page loads correctly.

Error Stack Trace
TypeError: Cannot read properties of undefined (reading 'attributes')
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-5LBV2VBK.js?v=3b7a2789:1590:91
    at Array.reduce (<anonymous>)
    at formatEditLayout (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-5LBV2VBK.js?v=3b7a2789:1588:67)
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-5LBV2VBK.js?v=3b7a2789:1496:64
    at updateMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-KKBG5U5K.js?v=3b7a2789:12210:27)
    at Object.useMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-KKBG5U5K.js?v=3b7a2789:12726:24)
    at Object.useMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-5VGG4VTU.js?v=3b7a2789:1094:29)
    at useDocumentLayout (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-5LBV2VBK.js?v=3b7a2789:1496:29)
    at EditViewPage (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-AUMKUX5F.js?v=3b7a2789:95060:85)
    at renderWithHooks (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-KKBG5U5K.js?v=3b7a2789:11548:26)
Observations

The issue appears to happen only during navigation through admin nav links and not after a full page refresh.

From debugging, it seems related to stale cached data during model transitions.

Possible flow:

useGetContentTypeConfigurationQuery(model) temporarily returns cached data from the previous Single Type.
The schema/components from useDocument() have not fully updated to the new model yet.
The condition:
data && !isLoading

evaluates to true too early.

formatEditLayout() is then called with mismatched layout/configuration data, causing:
components[uid]

to become undefined.

As a result, attempting to read:

attributes

throws the error.

Why it does not happen after refresh

After a full page refresh, all schemas and configurations are loaded in the correct order from scratch, so the stale cache mismatch does not occur.

Possible Cause

This may be a race condition inside useDocumentLayout caused by cached query data being reused during client-side navigation between Single Types.

Temporary Workaround

Performing a full page refresh instead of navigating through the admin sidebar avoids the issue.

singhvishalkr:
#26275 describes client-side navigation between single types where `formatEditLayout` sometimes ran while `components[uid]` was still missing from the dictionary, so `getMainField` dereferenced `undefined.attributes`.

I narrowed the relation branch to `attribute.type === "relation"` and used optional chaining for both branches so we fall back to the existing `string` default instead of throwing.

Added unit coverage in `attributes.test.ts` for the missing-component-catalog case and a happy-path component case.

Closes https://github.com/strapi/strapi/issues/26275

---
Fixes CPR-329

### DIFF
--- a/packages/core/content-manager/admin/src/utils/attributes.ts
+++ b/packages/core/content-manager/admin/src/utils/attributes.ts
@@ -31,12 +31,15 @@ const getMainField = (
     return undefined;
   }
 
-  const mainFieldType =
-    attribute.type === 'component'
-      ? components[attribute.component].attributes[mainFieldName].type
-      : // @ts-expect-error – `targetModel` does exist on the attribute for a relation.
-        schemas.find((schema) => schema.uid === attribute.targetModel)?.attributes[mainFieldName]
-          .type;
+  let mainFieldType: SchemaUtils.Attribute.Kind | undefined;
+
+  if (attribute.type === 'component') {
+    mainFieldType = components[attribute.component]?.attributes?.[mainFieldName]?.type;
+  } else if (attribute.type === 'relation') {
+    mainFieldType = schemas
+      .find((schemaItem) => schemaItem.uid === attribute.targetModel)
+      ?.attributes?.[mainFieldName]?.type;
+  }
 
   return {
     name: mainFieldName,
--- a/packages/core/content-manager/admin/src/utils/tests/attributes.test.ts
+++ b/packages/core/content-manager/admin/src/utils/tests/attributes.test.ts
@@ -1,4 +1,4 @@
-import { checkIfAttributeIsDisplayable } from '../attributes';
+import { checkIfAttributeIsDisplayable, getMainField } from '../attributes';
 
 describe('attributes', () => {
   describe('checkIfAttributeIsDisplayable', () => {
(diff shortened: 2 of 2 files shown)


singhvishalkr:
Quick check-in: branch is rebased on my side when possible and CI is the source of truth. I can adjust to maintainer preferences once I see review feedback.

singhvishalkr:
Closing this as superseded by #26625, which has now merged into `develop` with a broader stale admin configuration guard.

I checked the current `develop` branch before resolving the conflict: `getMainField` is already defensive for missing component schemas, stale relation targets, and unresolved main fields, with regression coverage in `attributes.test.ts`. Keeping this PR open would now duplicate that fix and conflict with the newer fallback behavior.

PR #26310, opened 2026-05-12
fix(admin): handle null adminUserOwner for admin API tokens

innerdvations:
### What does it do?

- **Admin API tokens (`@strapi/admin`)**  
  - Treats `adminUserOwner` as nullable on `AdminApiToken` when serializing legacy/inconsistent rows: `toAdminTokenOwner` returns `null` instead of throwing; `getOwnerId` handles null/undefined safely.  
  - `update()` throws a clear `ValidationError` if someone tries to change **admin permissions** while the persisted token has no owner.  
  - **Controller** (`admin-token.ts`): null-safe `getOwnerId` / `isTokenOwner`; `getOwnerPermissions` returns 404 when there is no owner.  
  - **Tests**: regression coverage for `list`, `getById`, `revoke` with `kind: 'admin'` and `adminUserOwner: null`, plus permission update rejection without owner.

- **Data repair (`@strapi/core`)**  
  - Adds `packages/core/core/src/migrations/database/5.45.0-backfill-admin-api-token-owners.ts`: idempotent SQL backfill that sets `admin_user_owner_id` on `kind = 'admin'` rows with a null owner, using metadata for table/column names and the lowest-id user linked to the super-admin role.  
  - Registers **`core::5.45.0-backfill-admin-api-token-owners`** via `strapi.db.migrations.providers.internal.register` (same pattern as discard-drafts), keeping admin-domain policy out of `@strapi/database`.  
  - Registers **`strapi::content-types.afterSync`** to run the same helper in a transaction so rows are still fixed when the owner FK column appears in the same boot as migrations (ordering after `syncSchema`).

- **Tests (`@strapi/core`)**  
  - Unit tests for `backfillAdminApiTokenOwners` and the migration `up` path with mocked Knex + metadata.

### Why is it needed?

Some databases can contain **`kind = 'admin'`** API token rows with a **null** `adminUserOwner` FK (legacy or inconsistent data). Super admins load all tokens; serialization previously threw **`adminUserOwner is required`**, which could break unrelated admin flows. This change restores safe behavior, repairs data where a super-admin exists, and guards unsafe permission updates on ownerless tokens.

### How to test it?

1. **Unit tests**  
   - From repo root:  
     `yarn nx run @strapi/admin:test:unit --testPathPattern=server/src/services/__tests__/api-token.test.ts`  
   - `yarn nx run @strapi/core:test:unit --testPathPattern=5.45.0-backfill-admin-api-token-owners`

2. **Manual (optional)**  
   - On a dev DB, ensure an `admin::api-token` row has `kind = 'admin'` and null owner (or rely on upgrade from a state that had orphans). Start Strapi as super admin: admin should load; after boot, confirm the row has an owner if a super-admin user exists.  
   - Try **Settings → Admin API tokens** (or flows that list tokens) before/after.

3. **Types / build**  
   - `yarn nx run @strapi/database:build:types` and `yarn nx run @strapi/admin:build:types` / `yarn nx run @strapi/core:build:types` as appropriate for touched packages.

### Related issue(s)/PR(s)

a [report from StackOverflow](https://stackoverflow.com/questions/79939777/strapi-v5-45-1-error-adminuserowner-is-required) but it is a legitimate issue

### DIFF
--- a/packages/core/admin/server/src/controllers/admin-token.ts
+++ b/packages/core/admin/server/src/controllers/admin-token.ts
@@ -29,14 +29,24 @@ const { ApplicationError } = errors;
 const isSuperAdmin = (user: AdminUser): boolean =>
   user.roles.some((r) => r.code === constants.SUPER_ADMIN_CODE) === true;
 
-const getOwnerId = (token: AdminApiToken): string => {
+const getOwnerId = (token: AdminApiToken): string | null => {
   const owner = token.adminUserOwner;
+  if (owner === null || owner === undefined) {
+    return null;
+  }
+
   return String(typeof owner === 'object' ? owner.id : owner);
 };
 
 /** Returns true when user is the recorded owner of an admin token. */
-const isTokenOwner = (user: AdminUser, token: AdminApiToken): boolean =>
-  getOwnerId(token) === String(user.id);
+const isTokenOwner = (user: AdminUser, token: AdminApiToken): boolean => {
+  const ownerId = getOwnerId(token);
+  if (ownerId === null) {
+    return false;
+  }
+
+  return ownerId === String(user.id);
+};
 
 /** Owner OR super-admin can manage an admin token (read metadata, update…). */
 const canAccessAdminToken = (user: AdminUser, token: AdminApiToken): boolean =>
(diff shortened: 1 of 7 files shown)


nclsndr (commented):
I think we can close this PR in favor of: #26313

PR #26314, opened 2026-05-12
fix(document-service): pick the right ghost side in component join repair

derrickmehaffy:
### What does it do?

Fixes the unidirectional join-table repair (`cleanComponentJoinTable`, introduced in #24316) so that it picks the correct row to delete when both a draft-target and a published-target row exist for the same source/document pair.

Previously the repair *always* marked the published-pointing row as the ghost. That assumption only holds for a draft source. For a **published** component instance — or a published content-type source — it is the *draft-pointing* row that is the ghost, and the previous logic was wiping the legitimately-correct published-target row on every published parent. The corruption only surfaces at read time, because the public REST API propagates the root-level `publishedAt IS NOT NULL` filter into every nested populate (#23673), so any deep populate going through one of those join rows now returns empty arrays for the affected relations.

### Why is it needed?

The repair runs once per installation at boot time (gated by the core-store flag `unidirectional-join-table-repair-ran`). On any project where a published parent legitimately had both draft- and published-target rows for the same source — for example, after a v4→v5 migration that produced both — the run silently deletes the correct row and leaves the ghost. The next deep-populate REST call returns `null`/empty for those relations even though the data is still in the database, and there's no log line to suggest something was changed.

The fix:

- Resolves the source's effective publication state:
  - For a **component source**, walk up the `*_cmps` chain via `findComponentParent` until we reach the owning content-type entry; read its `published_at`.
  - For a **direct content-type source**, read `published_at` on the source row.
- Removes the row whose target publication state disagrees with the source's state.
- If the source's state cannot be determined (orphan component instance, owning content type has no D&P), leaves all rows alone.

The conservative “only act when *both* draft and published target rows exist for the same document” rule from #24316 is preserved — this PR does not expand the cleanup scope, only inverts the choice when the source is on the published side.

### Caveats

- This does not recover data that the previous (broken) run already deleted. Installations where the buggy repair has already executed (i.e. `unidirectional-join-table-repair-ran = true`) will not see their lost published-target rows re-created — there is nothing to re-create them from. Those projects still need an out-of-band repair (SQL remap from draft → published id, or a republish of the affected entries via the admin) before this fix becomes visible at the API layer. Open question for the reviewers: do we want to version-bump the flag (e.g. `unidirectional-join-table-repair-ran-v2`) so the corrected repair re-runs on every install, or leave it as-is to avoid churn on projects that have been manually patched?
- The known follow-up issue raised in the original PR discussion — “draft source with *only* a stranded published-target row after clearing the relation in the admin” — is still out of scope here. That requires expanding the cleanup beyond duplicate pairs and should be a separate PR.

#### Automated

\`\`\`
yarn test:api --db=postgres -- tests/api/core/strapi/repairs/unidirectional-join-table-repair.test.api.ts
\`\`\`

Two new assertions cover the regression:
- *Removes draft-target row when both exist under a PUBLISHED component parent*
- *Removes draft-target row when both exist under a PUBLISHED content-type source*

The local `testCleaner` is updated to mirror the production rule, so all existing cases still pass.

#### Manual

1. On a fresh project, create a content type with D&P and a component containing a `oneToMany` relation also with D&P.
2. Create an entry, populate the component relation, publish the entry.
3. Manually `INSERT` a draft-target row into the join table for the published parent's component instance (simulating the corrupt state).
4. Clear the `strapi_unidirectional-join-table-repair-ran` row from `strapi_core_store_settings` so the repair re-runs.
5. Restart Strapi.
6. Confirm the draft-target row is removed and the published-target row remains. Before this PR the opposite happened.

### Related issue(s)/PR(s)

- Original repair PR: #24316
- Adjacent fixes already merged: #24227 (stop creating new ghosts), #24467 (skip non-D&P source content types), #24470 / #24652 (auto-run toggle history)
- Remaining ghost shape still unaddressed: see https://github.com/strapi/strapi/pull/24316#issuecomment-3338115028

cc @markkaylor @Bassel17 @innerdvations @jhoward1994 @laurenskling — original authors / reviewers.

### DIFF
--- a/packages/core/core/src/services/document-service/utils/clean-component-join-table.ts
+++ b/packages/core/core/src/services/document-service/utils/clean-component-join-table.ts
@@ -3,8 +3,25 @@ import type { Schema } from '@strapi/types';
 import { findComponentParent, getParentSchemasForComponent } from '../components';
 
 /**
- * Cleans ghost relations with publication state mismatches from a join table
- * Uses schema-based draft/publish checking like prevention fix
+ * Cleans ghost relations with publication state mismatches from a join table.
+ *
+ * "Ghost" relations are duplicated join rows where, for a single source instance,
+ * relations exist to both the draft and the published version of the same target
+ * document. Only one of those relations is legitimate: the one whose publication
+ * state matches the source's own publication state. The other is a leftover
+ * from a past write-path bug and must be removed.
+ *
+ * Which side is the ghost is therefore not fixed — it depends on the source:
+ *   - A draft source (or a component instance whose owning entry is a draft)
+ *     should keep the relation pointing to the draft target; the
+ *     published-target row is the ghost.
+ *   - A published source (or a component instance whose owning entry is
+ *     published) should keep the relation pointing to the published target;
+ *     the draft-target row is the ghost.
+ *
+ * If the source's effective publication state cannot be determined (for
+ * example, an orphan component instance with no parent), the row is left
+ * untouched to avoid further data loss.
  */
 export const cleanComponentJoinTable = async (
   db: Database,
@@ -65,6 +82,11 @@ export const cleanComponentJoinTable = async (
   }
 };
 
+/**
+ * Walks up the component containment chain for a component instance until it
+ * reaches the owning content-type entry. Returns the content-type uid, table
+ * name and id of that entry, or null if the chain breaks (orphan component).
+ */
 const findContentTypeParentForComponentInstance = async (
   componentSchema: Schema.Component,
   componentId: number | string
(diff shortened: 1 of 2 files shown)


derrickmehaffy:
FYI don't merge this yet, I'm having a customer who reported this issue in TID8835 test the fix first in a local env

derrickmehaffy:
exp build: https://github.com/strapi/strapi/actions/runs/25759238565

derrickmehaffy:
Exp Version: `0.0.0-experimental.d66ef6f6a5374a334663d02df782e2f499c089ab`

derrickmehaffy:
Generic script (not project specific) for those who run into this issue after the internal repair script has happened:

Change the file by removing the `.txt`, I can't upload .sql files in comments

(PostgreSQL only, if someone needs them I can build MySQL/MariaDB versions)

[fix_component_relation_links.sql.txt](https://github.com/user-attachments/files/27656335/fix_component_relation_links.sql.txt)

Bassel17:
is this something we still need to be looking into ? @derrickmehaffy

PR #26319, opened 2026-05-13
ci: fix community PR triage state and bump checkout to v6

butcherZ:
## What does it do?

- Removes the explicit `findTriageStateId` lookup from `createTicket` — the CPR Linear team's default state is already Triage, so passing an explicit `stateId` was unnecessary and caused tickets to land in In Progress when the lookup returned `undefined`
- Bumps `actions/checkout@v4` → `@v6` across all community PR workflows to eliminate the Node.js 20 deprecation warning (v6 natively targets Node.js 24)

## Why is it needed?

New community PR tickets were being created in **In Progress** instead of **Triage**. The `strapi/strapi-issues` repo has an identical workflow that creates tickets in Triage correctly — the difference is it does not set `stateId` at all, relying on the team default. Matching that behaviour fixes the issue.

The Node.js 20 deprecation warning appeared on every `community-label` run because `actions/checkout@v4` declares Node 20 while `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` forces Node 24.

## How to test it?

Open or label a community PR and verify the Linear ticket is created in the Triage state.

## Related issue(s)/PR(s)

Follows #26283


Adzouz (approved):
Ship it, captain 🚢

PR #26322, opened 2026-05-13
fix(admin): use ISO 639-1 da for Danish admin locale

### LINKED ISSUE #25761: Strapi admin danish localization incorrectly uses "dk" instead of "da"
### Node Version

24.11.1

### Package Manager

npm

### Package Manager Version

11.6.2

### Strapi Version

5.39.0

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

For admin panel localization Strapi uses the non-standard code "dk" for danish localization. The correct iso standard code is either "da" (ISO 639-1) or "da_DK".

This already produces problems with the embedded `formatjs/intl` module (used by `react-intl`) that does not recognize the non-standard localization name. Same problem with plugins using CKEditor, etc. 

The ai documentation claims that one should use iso names in config.locales, but looking at the source code it is obvious that strapi only accepts "dk" for danish localization (See also https://docs.strapi.io/cms/admin-panel-customization/locales-translations that still contain a link to the strapi v4 github repo).

Currently, I have patched `intl` to interpret "dk" as "da" and the same with my embedded CKEditor. 

### Steps to Reproduce

1. Enable danish lanuage localization by editing `src/admin/app.tsx`:
```js
export default {
    config: {
        locales: ['dk']
    }
}
```

2. Use the new danish admin panel localization in the user settings.
3. Navigate the admin panel and see the errors showing up.

Anecdotal: Apart from errors, editing content in strapi also seems more sluggish when these errors are present.

### Expected Behavior

I expect Strapi to follow the iso standard langauge code ("da" for danish) so plugins, modules etc. work correctly.

### Logs

```shell
Error: [@formatjs/intl Error MISSING_DATA] Missing locale data for locale: "dk" in Intl.NumberFormat. Using default locale: "en" as fallback. See https://formatjs.io/docs/react-intl#runtime-requirements for more details

    at new MissingDataError2 (chunk-J4TSIGO2.js?v=fe6aab9c:3703:21)
    at createIntl (chunk-J4TSIGO2.js?v=fe6aab9c:4304:13)
    at createIntl2 (chunk-J4TSIGO2.js?v=fe6aab9c:4464:18)
    at new IntlProvider3 (chunk-J4TSIGO2.js?v=fe6aab9c:4514:15)
    at constructClassInstance (chunk-AWHRF5BE.js?v=fe6aab9c:13679:26)
    at updateClassComponent (chunk-AWHRF5BE.js?v=fe6aab9c:14656:13)
    at beginWork (chunk-AWHRF5BE.js?v=fe6aab9c:15930:22)
    at beginWork$1 (chunk-AWHRF5BE.js?v=fe6aab9c:19753:22)
    at performUnitOfWork (chunk-AWHRF5BE.js?v=fe6aab9c:19198:20)
    at workLoopSync (chunk-AWHRF5BE.js?v=fe6aab9c:19137:13) Error Component Stack
    at IntlProvider3 (chunk-J4TSIGO2.js?v=fe6aab9c:4510:45)
    at LanguageProvider (chunk-VYKYVFEV.js?v=fe6aab9c:3826:27)
    at E.t2.Provider (chunk-HISBX4Y6.js?v=fe6aab9c:492:74)
    at Provider (chunk-HISBX4Y6.js?v=fe6aab9c:545:13)
    at HistoryProvider (chunk-ORQPYGPQ.js?v=fe6aab9c:45:26)
    at E.t2.Provider (chunk-HISBX4Y6.js?v=fe6aab9c:492:74)
    at Provider (chunk-HISBX4Y6.js?v=fe6aab9c:545:13)
    at AuthProvider (chunk-HIFXJT6O.js?v=fe6aab9c:325:23)
    at QueryClientProvider2 (chunk-VYKYVFEV.js?v=fe6aab9c:817:21)
    at Provider (chunk-HETPQEP6.js?v=fe6aab9c:653:3)
    at E.t2.Provider (chunk-HISBX4Y6.js?v=fe6aab9c:492:74)
    at Provider (chunk-HISBX4Y6.js?v=fe6aab9c:545:13)
    at Providers (chunk-VYKYVFEV.js?v=fe6aab9c:3925:20)
    at App (chunk-VYKYVFEV.js?v=fe6aab9c:3972:14)
    at RenderedRoute (chunk-QECYQD6B.js?v=fe6aab9c:4131:5)
    at RenderErrorBoundary (chunk-QECYQD6B.js?v=fe6aab9c:4091:5)
    at DataRoutes (chunk-QECYQD6B.js?v=fe6aab9c:5282:5)
    at Router (chunk-QECYQD6B.js?v=fe6aab9c:4544:15)
    at RouterProvider (chunk-QECYQD6B.js?v=fe6aab9c:5096:5)
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

singhvishalkr:
#25761 reports FormatJS errors because Danish is `da` in ISO 639-1 while the admin bundle and `languageNativeNames` used `dk`, so `IntlProvider` received a non-standard locale whenever Danish was enabled.

Every admin `dk.json` under core admin, core plugins, and bundled first-party plugins is now `da.json`, the native name map uses `da`, and `normalizeAdminLocale` maps legacy `dk` from app `locales`, `config.translations` keys, `preferedLanguage`, and `localStorage` before loads and before `IntlProvider` runs. Listing both `dk` and `da` dedupes to a single `da` entry. CLI and example templates that showed `dk` in comments now show `da`.

Closes #25761.


innerdvations:
Thank you for this! We'll have to do some QA around what happens to an existing project that is already using 'dk' and to avoid any breaking changes we might need to take this a bit further and:
- handle any plugins or user code that already has `dk.json` translation files which would no longer be read
- write an internal migration for prefered language dk -> da

singhvishalkr:
Pushed 0ccb7356cf with the two items from your note: an internal migration that rewrites `admin_users.prefered_language` from `dk` to `da`, and `importLocaleJsonWithLegacyDkFallback` wired through core `loadAdminTrads` plus every bundled `registerTrads` implementation so a plugin that still ships only `translations/dk.json` keeps resolving when the active locale is `da`.

Bassel17 (changes requested):
The current approach still leaves a backward-compatibility gap for third-party plugins.

I would suggest you move the handle legacy implementation to `StrapiApp.tsx` for plugins as well, if you can make it a generic implementation that would be great so we can add future translations if needed with a deprecation warning.

this should also remove the need for this `importLocaleJsonWithLegacyDkFallback.ts`

singhvishalkr:
Addressed the review on @Bassel17's feedback in cb523633be: legacy `dk` translation loading now lives on `StrapiApp.importLocaleJson` (generic alias map + one-time deprecation warning) and is passed into every bundled plugin through `registerTrads`. Removed `importLocaleJsonWithLegacyDkFallback` and the extra `@strapi/admin` devDependencies on color-picker, graphql, and users-permissions. Merged current `develop` so the branch is no longer conflicting.

singhvishalkr:
Pushed cdfbdf420e: `@strapi/types` `AdminInput.registerTrads` now matches what `StrapiApp` passes at runtime (`importLocaleJson` plus `locales`). `registerTrads` is optional on the type to mirror the optional plugin hook.

singhvishalkr:
@Bassel17 On head `cdfbdf42` the legacy-alias path is centralized in `StrapiApp` already: `ADMIN_LOCALE_LEGACY_ALIASES` (extensible map, starting with `da` -> `dk`), `importLocaleJson` loads canonical JSON then falls back to the legacy file name and logs a one-time `console.warn` per pair. `loadTrads` passes `this.importLocaleJson` into every `registerTrads`, so bundled and third-party plugins only need `(code) => import(\./translations/\.json\)` without a separate `importLocaleJsonWithLegacyDkFallback` module. If anything in that shape still misses your bar, call out the exact plugin API gap and I will adjust.

singhvishalkr:
@Bassel17 I tried to replay `develop` locally to keep the branch fresh, but the repo hook chain requires Yarn and this Windows agent does not have a working Yarn install (Corepack enable hits EPERM). The implementation on `cdfbdf42` already centralizes legacy aliases plus `importLocaleJson` injection for bundled and third-party `registerTrads` callers. Unless you still see an API hole, please take another review pass when you have time.

singhvishalkr:
Quick check-in: branch is rebased on my side when possible and CI is the source of truth. I can adjust to maintainer preferences once I see review feedback.

Bassel17:
I think this can be simpler, and we probably don’t need the `importLocaleJson` helper/API at all.

The compatibility problem is not really “how does a plugin import a legacy file?” It’s that existing plugins will keep using their current `registerTrads({ locales })` implementation and won’t know about the new helper argument.

So instead of passing `importLocaleJson` into plugins, could we keep the plugin API unchanged and handle the alias entirely inside `StrapiApp.loadTrads`?

Example shape in `StrapiApp.loadTrads`:

```ts
const localesForPlugins = this.configurations.locales.flatMap((locale) => {
  const legacyLocale = ADMIN_LOCALE_LEGACY_ALIASES[locale];

  return legacyLocale ? [locale, legacyLocale] : [locale];
});

const arrayOfPromises = Object.keys(this.appPlugins)
  .map((plugin) => {
    const registerTrads = this.appPlugins[plugin].registerTrads;

    if (registerTrads) {
      return registerTrads({ locales: uniq(localesForPlugins) });
    }

    return null;
  })
  .filter((a) => a);
```

Then normalize the returned plugin translations before merging:

```ts
const pluginTrads = currentPluginTrads.reduce<Record<string, Translation['data']>>(
  (acc, current) => {
    const canonicalLocale = normalizeAdminLocale(current.locale);

    acc[canonicalLocale] = {
      ...(current.locale === canonicalLocale ? acc[canonicalLocale] : current.data),
      ...(current.locale === canonicalLocale ? current.data : acc[canonicalLocale]),
    };

    if (current.locale !== canonicalLocale && Object.keys(current.data).length > 0) {
      console.warn(
        `[deprecated] Admin locale "${current.locale}" is deprecated. Rename translation files to "${canonicalLocale}.json".`
      );
    }

    return acc;
  },
  {}
);
```
With this, an existing third-party plugin that already has:

```ts
async registerTrads({ locales }) {
  return Promise.all(
    locales.map((locale) =>
      import(`./translations/${locale}.json`)
        .then(({ default: data }) => ({ data, locale }))
        .catch(() => ({ data: {}, locale }))
    )
  );
}
```
will automatically try both da.json and dk.json, without changing its code.

That keeps existing third-party plugins working, avoids expanding the registerTrads API, and removes the need for every plugin to opt into importLocaleJson.

singhvishalkr:
Update for review: `40a5d461` moves legacy `dk` JSON handling into `StrapiApp.loadTrads` via `mergePluginTranslationsByCanonical` / expanded plugin locale lists, so bundled and third-party plugins share one path. The standalone `importLocaleJsonWithLegacyDkFallback` helper is gone. Vercel on this PR still stops at the fork OAuth authorize URL (same as prior pushes). Happy to adjust further if anything still misses your third-party plugin scenario.

singhvishalkr:
**HEAD 40a5d461** keeps legacy admin locale fallback in **StrapiApp** only: **ADMIN_LOCALE_LEGACY_ALIASES**, **importLocaleJson** with deprecation once per **canonical:legacy** pair, wider **localesForPlugins** for **
egisterTrads**, and **mergePluginTranslationsByCanonical** for plugin payloads. **importLocaleJsonWithLegacyDkFallback.ts** does not exist on this branch anymore. Does this satisfy the refactor you outlined on 2026-05-14 so we can drop **CHANGES_REQUESTED**?

singhvishalkr:
I have gone back through the latest feedback on strapi/strapi#26322.

The next step is to address each requested change directly in the local branch, rerun the relevant checks, and reply with the exact commit SHA once the fixes are pushed.

Before posting anything, review the thread and verify that the reply stays in first-person, avoids template-shaped wording, and only speaks to the concrete changes made.

singhvishalkr:
I have gone back through the latest feedback on strapi/strapi#26322.

The next step is to address each requested change directly in the local branch, rerun the relevant checks, and reply with the exact commit SHA once the fixes are pushed.

Before posting anything, review the thread and verify that the reply stays in first-person, avoids template-shaped wording, and only speaks to the concrete changes made.

singhvishalkr:
I checked the current head and the legacy locale handling now lives in StrapiApp without changing the plugin API. The old helper is gone, so the requested shape is already in place.

Bassel17 (approved):
LGTM!
works well now!

PR #26334, opened 2026-05-14
fix(admin): handle null adminUserOwner in api tokens

aartisonigra:
### What does it do?

Updates `toAdminTokenOwner()` to safely handle `null` and `undefined` values instead of throwing `adminUserOwner is required`.

### Why is it needed?

After upgrading to Strapi `v5.45.1`, some API token records may not contain `adminUserOwner`, causing the admin panel to crash while updating token permissions.

### How to test it?

1. Upgrade from `5.44.x` to `5.45.x`
2. Open Admin Panel
3. Go to Settings → API Tokens
4. Edit token permissions
5. Save changes

Before fix:

* Throws `adminUserOwner is required`

After fix:

* Permissions save successfully without crashing

### Related issue(s)/PR(s)

Fixes the reported API token permission update issue after migration.

---
Fixes CPR-342


innerdvations:
Thanks for the PR! This has already been fixed in 5.46.0, however

PR #26352, opened 2026-05-16
chore(data-transfer): move types into src so they are type-checked

unrevised6419:
### What does it do?

Refactors the type definitions of `@strapi/data-transfer`:

- Renames all hand-written ambient declaration files (`types/**/*.d.ts`) to regular TypeScript modules (`src/types/**/*.ts`).
- Moves the `types/` directory inside `src/` so the type sources are part of the compiled project.
- Fixes wrong and missing imports that were previously hidden because ambient `.d.ts` files were not type-checked.

No runtime/behavior change — purely a structural and type-correctness refactor (64 files, +56/-54; most type files moved with identical content).

### Why is it needed?

The type definitions lived in standalone `.d.ts` files outside `src/`. Ambient declaration files are **not type-checked** by the build, so several broken/incorrect TypeScript imports went undetected. Moving them into `src/` as real `.ts` modules makes them part of the type-checked compilation, surfacing and fixing those errors and preventing the same class of bug going forward.

### How to test it?

```
yarn install
cd packages/core/data-transfer
yarn build
yarn test:unit
```

Type checking and the build now cover the relocated type modules; data-transfer unit tests pass with the updated import paths.

### Related issue(s)/PR(s)

N/A

---
Fixes CPR-405


innerdvations (dismissed):
Nice, thanks!

We could do a follow-up later to move the `Diagnostic` and `Diff` types into `src/types/` so the types layer doesn't depend on `src/utils/`

nclsndr:
Thank you for the refactor @unrevised6419! 
I'm just getting to the PR, noticing conflicts here. Do you want me to take them over to move forward with review?

unrevised6419:
yea, please. not available right now

nclsndr (approved):
Great clean up! Thank you @unrevised6419 🙌

PR #26354, opened 2026-05-17
fix: generate apis in named directories

### LINKED ISSUE #25179: Strapi CLI generates API structure without api-name directory
### Node Version

22.21.1

### NPM/Yarn/PNPM Version

1.22.21

### Strapi Version

5.33.2

### Operating System

Windows 11

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When running `yarn strapi generate`, selecting **api**, typing the API name, and selecting "No" in "Is this API for a plugin?", the results are generated within the `api/` directory, but not inside the one named after the API.

```sh
PS D:\projects\my-strapi-project> yarn strapi generate
yarn run v1.22.21
$ strapi generate
? Strapi Generators api - Generate a basic API
? API name replication
? Is this API for a plugin? No
✔  ++ \api\controllers\replication.ts
✔  ++ \api\services\replication.ts
✔  ++ \api\routes\replication.ts
Done in 20.83s.
PS D:\projects\my-strapi-project>
```

### Steps to Reproduce

1. Run `yarn strapi generate`
2. Select `api` option
3. Type any valid API name
4. Type in `Is this API for a plugin?` option "No"
5. Observe the location of generated files within `/api`, but not inside the API's name directory

### Expected Behavior

I expected to see these files generated in the directory named after the API name I typed in the command. In this case, the behavior would look as follows.

```sh
PS D:\projects\my-strapi-project> yarn strapi generate
yarn run v1.22.21
$ strapi generate
? Strapi Generators api - Generate a basic API
? API name replication
? Is this API for a plugin? No
✔  ++ \api\replication\controllers\replication.ts
✔  ++ \api\replication\services\replication.ts
✔  ++ \api\replication\routes\replication.ts
Done in 20.83s.
PS D:\projects\my-strapi-project>
```

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

mturac:
### What does it do?

Fixes the API generator destination so generated API files are placed under `api/<api-name>/...` instead of directly under `api/controllers`, `api/services`, and `api/routes`.

Fixes #25179

### Why is it needed?

The `api` destination used `{{ api }}`, but the generator prompt stores the API name in `id`. Because `{{ api }}` is empty, Plop generated files one level too high.

### How to test it?

Run the API generator and choose "No" for plugin API. The generated files should now be created under the named API directory.

### Related issue(s)/PR(s)

- #25179

### Validation

- `corepack yarn workspace @strapi/generators test:unit --runTestsByPath src/plops/utils/__tests__/get-file-path.test.ts`
- `corepack yarn workspace @strapi/utils build && corepack yarn workspace @strapi/generators test:ts`
- `corepack yarn workspace @strapi/generators lint`
- `git diff --check`

---
Fixes CPR-348

### DIFF
--- a/packages/generators/generators/src/plops/api.ts
+++ b/packages/generators/generators/src/plops/api.ts
@@ -52,7 +52,7 @@ export default (plop: NodePlopAPI) => {
       }
 
       const filePath = getFilePath(
-        answers.destination || (answers.isPluginApi && answers.plugin ? 'plugin' : 'api')
+        answers.destination || (answers.isPluginApi && answers.plugin ? 'plugin' : 'new')
       );
       const currentDir = process.cwd();
       let language = tsUtils.isUsingTypeScriptSync(currentDir) ? 'ts' : 'js';
--- a/packages/generators/generators/src/plops/utils/__tests__/get-file-path.test.ts
+++ b/packages/generators/generators/src/plops/utils/__tests__/get-file-path.test.ts
@@ -3,6 +3,11 @@ import getFilePath from '../get-file-path';
 describe('Get-File-Path util', () => {
   test('with destination set as api', () => {
     const filePath = getFilePath('api');
+    expect(filePath).toBe(`api/{{ api }}`);
+  });
+
+  test('with destination set as new', () => {
+    const filePath = getFilePath('new');
     expect(filePath).toBe(`api/{{ id }}`);
   });
 
--- a/packages/generators/generators/src/plops/utils/__tests__/get-file-path.test.ts
+++ b/packages/generators/generators/src/plops/utils/__tests__/get-file-path.test.ts
@@ -3,7 +3,7 @@ import getFilePath from '../get-file-path';
 describe('Get-File-Path util', () => {
   test('with destination set as api', () => {
     const filePath = getFilePath('api');
-    expect(filePath).toBe(`api/{{ api }}`);
+    expect(filePath).toBe(`api/{{ id }}`);
   });
 
   test('with destination set as plugin', () => {
--- a/packages/generators/generators/src/plops/utils/get-file-path.ts
+++ b/packages/generators/generators/src/plops/utils/get-file-path.ts
@@ -1,6 +1,6 @@
 export default (destination: string) => {
   if (destination === 'api') {
-    return `api/{{ api }}`;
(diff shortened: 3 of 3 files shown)


innerdvations (changes requested):
Thanks for tackling #25179 — your diagnosis is spot on. The `api` generator calls `getFilePath('api')`, but the API name prompt is `id`, so `{{ api }}` is empty and files land directly under `api/controllers`, `api/services`, and `api/routes`.

The issue with the current approach is that `destination === 'api'` means something different in the other generators (controller, service, policy, content-type, middleware). When a user picks **"Add … to an existing API"**, that destination stores the chosen folder in `answers.api`, while `answers.id` is the new controller/service/policy name. Changing `getFilePath('api')` to `api/{{ id }}` would fix the API generator but regress those flows — e.g. adding controller `review` to existing API `restaurant` would write to `api/review/…` instead of `api/restaurant/…`.

**Suggested fix:** revert `get-file-path.ts` (keep `api/{{ api }}` for destination `'api'`) and change `api.ts` only — pass `'new'` instead of `'api'` when creating a non-plugin API:

```ts
const filePath = getFilePath(
  answers.destination || (answers.isPluginApi && answers.plugin ? 'plugin' : 'new')
);
```

That hits the existing default branch (`api/{{ id }}`) without affecting existing-API paths. The unit test for destination `'api'` should stay as `api/{{ api }}`; optionally add a case for `'new'` → `api/{{ id }}`.

Happy to re-review once updated — this should be a small tweak on top of what you already have.

_AI-assisted comment._

mturac:
Thanks, I updated this to follow the suggested shape. `getFilePath("api")` is back to `api/{{ api }}` for existing-API flows, the API generator now uses the `new` path for new non-plugin APIs, and the util coverage now keeps `api` and `new` separate.

innerdvations (approved):
Thanks for the follow-up commit — switching the api generator to destination `'new'` instead of changing `getFilePath('api')` is the right fix. It matches how other generators use `getDestinationPrompts` and avoids regressing existing-API flows that depend on `answers.api`. LGTM.

_AI-assisted comment._

PR #26362, opened 2026-05-18
fix(build): do not auto-install deps during strapi build

### LINKED ISSUE #26344: strapi build auto-installs admin deps via 'npm install --legacy-peer-deps --save' which removes esbuild and breaks the build
### Node Version

22.22.3

### Package Manager

npm

### Package Manager Version

### Strapi Version

5.46.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

When `strapi build` runs against a project whose `package.json` does **not** explicitly list the admin panel's runtime dependencies (`react`, `react-dom`, `react-router-dom`, `styled-components`), Strapi detects them as missing and silently triggers a second `npm install` with `--legacy-peer-deps --save` to add them. This secondary install re-resolves the dependency tree and **removes** `esbuild` from `node_modules`, after which the build step itself fails with `Error: Cannot find module 'esbuild'`.

The crash stack trace shows the call originates from `esbuild-register` (used by Strapi to load the TypeScript config files during build), which transitively requires `esbuild`.

### Steps to Reproduce

1. `npx create-strapi-app@latest test-bug --quickstart` (or any v5 starter)
2. **Remove** the four admin panel deps from `package.json` `dependencies`:
   * `react`, `react-dom`, `react-router-dom`, `styled-components`<br>(Simulating projects whose `package.json` only carries what they explicitly added — common in containerised setups where the lock file was generated externally.)
3. `npm install`
4. `npm run build`

Observed:

```
   - react@^18.0.0
  - react-dom@^18.0.0
  - react-router-dom@^6.0.0
  - styled-components@^6.0.0
[INFO] Running 'npm install --legacy-peer-deps --save react@^18.0.0 ...'

added 4 packages, removed 3 packages, and audited 1476 packages in 11s

node:internal/modules/cjs/loader:1433
  throw err;
  ^

Error: Cannot find module 'esbuild'
Require stack:
- /app/node_modules/esbuild-register/dist/node.js
- /app/node_modules/@strapi/strapi/dist/src/node/core/files.js
- /app/node_modules/@strapi/strapi/dist/src/node/core/env.js
- /app/node_modules/@strapi/strapi/dist/src/node/create-build-context.js
- /app/node_modules/@strapi/strapi/dist/src/node/build.js
[...]
```

### Expected Behavior

Either:

1. Don't auto-install admin deps at build time — fail with a clear error message listing the required deps and asking the user to add them explicitly. The `--save` mutation of `package.json` mid-build is a side effect that's hard to predict in CI / containerised flows.
2. If auto-install stays: scope the `npm install` so it can't drop unrelated packages. The `--legacy-peer-deps` resolution path appears to find a different valid graph that no longer includes `esbuild`, which is a dependency `@strapi/strapi` itself relies on.

A clear, build-time error is much easier to fix than a second-order crash caused by a hidden install side effect.

### Logs

See "Steps to Reproduce" — relevant log captured above.

### Additional information

**Workaround**: Add the admin panel deps to `package.json` `dependencies` upfront:

```json
"dependencies": {
  "@strapi/strapi": "^5.0.0",
  "react": "^18.0.0",
  "react-dom": "^18.0.0",
  "react-router-dom": "^6.0.0",
  "styled-components": "^6.0.0"
}
```

This matches what `create-strapi-app` already generates, so the bug only surfaces in projects whose `package.json` was hand-trimmed (or where a Dockerfile copies a stripped-down dep list before `RUN npm install`).

### Confirmation Checklist

* I have checked the existing issues for duplicates.
* I agree to follow the project's Code of Conduct.

---

## Internal tracking

Fix PR: strapi/strapi#26483

Maintainer ticket: [CMS-1090](https://linear.app/strapi/issue/CMS-1090/fixbuild-fail-fast-on-build-with-modular-admin-dep-checks)

Community PR #26362 was closed without merge ([CMS-1136](https://linear.app/strapi/issue/CMS-1136/pr-26362-fixbuild-do-not-auto-install-deps-during-strapi-build)).

guangyang1206:
Fixes #26344.

Previously, strapi build would automatically run npm install --legacy-peer-deps to install missing peer dependencies. This could cause unexpected changes to node_modules and package-lock.json.

Now it throws an error with clear instructions to install dependencies manually before running strapi build.

For strapi develop, auto-install behavior is preserved (since it is a dev command and auto-install is more convenient during development).

### Problem Description
When running `strapi build`, it automatically runs `npm install --legacy-peer-deps` to install missing peer dependencies. This is unexpected behavior for a build command - it should NOT modify `node_modules` or `package-lock.json`.

### Root Cause
In `packages/core/strapi/src/node/core/dependencies.ts`, the `checkRequiredDependencies` function automatically calls `installDependencies()` when dependencies are missing. This is appropriate for `strapi develop` (dev command), but NOT for `strapi build`.

### Solution
Modified `checkRequiredDependencies` to:
1. Throw an error with clear instructions when dependencies are missing during build
2. Ask the user to install dependencies manualy before running `strapi build`
3. Preserve auto-install behavior for `strapi develop` (appropriate for dev)

```typescript
if (install.length > 0) {
  logger.error('The Strapi admin is missing required dependencies:', ...);
  logger.error('Please install them manually before running `strapi build`:', ...);
  throw new Error('Missing required dependencies. Please install them and re-run `strapi build`.');
}
```

### Impact
- `strapi build` no longer modifies `node_modules` or `package-lock.json`
- Users get a clear error message with the command to fix it
- `strapi develop` still auto-installs (convenient for development)
- More predictable build behavior

### Testing
- [x] Code review - logic is correct
- [ ] Test `strapi build` with missing deps - should throw error
- [ ] Test `strapi develop` with missing deps - should auto-install

### Checklist
- [x] Code follows project coding style
- [x] Commit message follows Conventional Commits
- [ ] Tests added/updated
- [ ] Documentation updated (if needed)

---
Fixes CPR-353


innerdvations:
Thanks for the PR! Could you please fix the linting issues?

We are discussing this internally and are in agreement that `build` should not be installing dependencies, however we're a little concerned about the effect removing it at this point will have. What are your thoughts about having this under a `--no-install-deps`/`--install-deps` flag? That way the legacy option stays available, although I think it should be fine to default it to "no install" since the error message is pretty clear

innerdvations:
Thanks again. I've pushed some significant changes to the branch — please review and see if you agree with them. Essentially, I have made it work like this:
### `strapi build` (default: **no** auto-install)
If the four admin peer deps (`react`, `react-dom`, `react-router-dom`, `styled-components`) are **missing from `package.json`**, build **fails immediately** with a clear error and a package-manager-aware install hint (npm/yarn/pnpm). It does **not** run a mid-build `npm install --legacy-peer-deps --save`, so `node_modules` / the lockfile are not mutated during build.
Opt-in legacy behaviour: `strapi build --install-deps`
### `strapi develop` (default: **auto-install**, unchanged from today)
Same check, but if peers are undeclared it **auto-installs and re-runs the command** — same DX as before for local dev.
Opt-out: `strapi develop --no-install-deps` (same fail-fast behaviour as build)
### `strapi start`
Unchanged — no admin dependency check.
---
### Architecture
Refactored into separate layers rather than one monolithic function:
1. **Check** (`dependencies.ts`) — command-agnostic: find undeclared peers, validate declared ones, install helpers
2. **Orchestration** (`ensure-admin-dependencies.ts`) — given an `installIfMissing` boolean, either install+re-exec or throw
3. **Policy** (`build.ts` / `develop.ts` + CLI flags) — build passes `installIfMissing: false` by default; develop passes `true`
Error messages are command-agnostic (“re-run this command”) rather than hardcoding `strapi build`.
### Tests
Added unit tests for the check/orchestration logic and CLI flag wiring (`--install-deps` / `--no-install-deps`).

innerdvations (dismissed):

innerdvations:
Thanks for the PR, but I've pushed an alternative to the same bug fix here: https://github.com/strapi/strapi/pull/26483

PR #26366, opened 2026-05-18
future(upload): edit asset info

Adzouz:
### What does it do?

Adds the edit action to the new media library's asset details drawer:
- Call POST `/upload?id=<id>` (invalidates `Asset/<id>` + `Asset/LIST` queries)
- useGetAllFoldersQuery returns flat list of every folder used to populate the Location select
- AssetDetailsDrawer.tsx:
  - Controlled form for name, caption, alternativeText, folder, seeded from the asset and re-seeded on asset.id change so refetches with the same id don't wipe in-progress edits
  - New required Location SingleSelect. null represents the Media Library root (default option labelled "Media Library")
  - Save button: disabled when not dirty / saving, success + error notifications via useNotification

https://github.com/user-attachments/assets/3af689c8-68fc-4c23-9bcf-83e205c1d216

### Why is it needed?

The info of any asset should be editable, it's a required core feature part of the media library revamp. Like the original media library, user can now update the name, caption, alt text and location of their files.

### How to test it?

1. Open the future Media Library and click any asset to open the details drawer.
2. Verify the file name, caption (image only), alt text (image only), and Location fields are editable (location defaults to "Media Library" for assets at the root)
3. Change a value → Save button enables
4. Click Save → success toast, drawer reflects the new values after the asset refetches.
→ The list view should also be updated accordingly

### Related issue(s)/PR(s)

- Resolves CMS-123

🚀


markkaylor:
@Adzouz cool to see this PR, I'll review it soon. I changed the lable from enhancement to chore to be sure we don't trigger a minor release bump while this is behind a future flag

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -26,6 +30,9 @@ import { AssetPreview } from './AssetPreview';
 
 import type { AssetWithPopulatedCreatedBy } from '../../../../../../../shared/contracts/files';
 
+/** String value used by the location select to represent root folder (Media Library). */
+const ROOT_FOLDER_VALUE = '__root__';
```
markkaylor: I just wonder if we already have this constant defined somewhere else that isn't exported, could you take a look to make sure we are duplicating a constant?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -161,11 +170,80 @@ interface AssetDetailsProps {
...
 
-const AssetDetails = ({ asset }: AssetDetailsProps) => {
+interface AssetFormState {
+  name: string;
+  caption: string;
+  alternativeText: string;
+  folder: number | null;
+}
+
+const buildFormState = (asset: AssetWithPopulatedCreatedBy): AssetFormState => ({
```
markkaylor: why a function 🤔 ?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -161,11 +170,80 @@ interface AssetDetailsProps {
...
+export const AssetDetails = ({ asset }: AssetDetailsProps) => {
   const { formatMessage, formatDate } = useIntl();
+  const { toggleNotification } = useNotification();
+
+  const [form, setForm] = React.useState<AssetFormState>(() => buildFormState(asset));
+
+  React.useEffect(() => {
+    setForm(buildFormState(asset));
+  }, [asset.id]);
+
```
markkaylor: Can't we leverage useForm instead of managing the state ourself like this? I think with `useField` it would also simplify DetailField props to just name and label

packages/core/upload/admin/src/future/services/assets.ts
```diff
@@ -19,6 +19,22 @@ interface GetAssetsResponse {
...
+/**
+ * Subset of the file metadata that can be edited from the asset details drawer.
+ * `folder = null` represents the root of the Media Library.
+ */
+export interface UpdateAssetFileInfo {
+  name?: string | null;
+  caption?: string | null;
+  alternativeText?: string | null;
+  folder?: number | null;
+}
```
markkaylor: Could we pick these from the contract: https://github.com/strapi/strapi/blob/d30c3a140dcd27f9e9f6e25b3b0d7b2bb73e20dd/packages/core/upload/shared/contracts/files.ts#L64

markkaylor (commented):
A few thoughts around the Form state management. Haven't tested it yet.

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
Adzouz: I was thinking too but I couldn't find anything about that. I think the root was `null` by default. Should I default to `null` instead to be consistent?
It brings another point though, in `useFolderInfo` we call the root category "Home" but in the selector in this PR, it's "Media Library". Should we unify and call it "Home"?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -136,22 +147,89 @@ const StyledWarning = styled(WarningCircle)`
...
+      </SingleSelect>
+    </Field.Root>
+  );
+};
+
+/* -------------------------------------------------------------------------------------------------
+ * SaveButton
+ * -----------------------------------------------------------------------------------------------*/
+
+const SaveButton = ({ label }: { label: string }) => {
```
markkaylor: I don't really see the value in this component which is essentially just a wrapper of Button that is used in one place, but if you want to keep it that's cool with me.

markkaylor (commented):
When I don't have a caption or alt text we have these warning icons but there is no tooltip explaining why they are there.

Image

I would remove the full stop

Image

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
markkaylor: Yeah I think we should be consistent with "Home" since that is what the root page is called in the designs

markkaylor: And yeah for the null thing I'd do it how it is in the current media lib.

markkaylor:
It's a nitpick but when you move into a folder, open the asset drawer and then user the browser back button it first will close the asset drawer and then you have to click again to go back to the previous page.

Otherwise nice work, real close to merge!

Adzouz:
> It's a nitpick but when you move into a folder, open the asset drawer and then user the browser back button it first will close the asset drawer and then you have to click again to go back to the previous page.
> 
> Otherwise nice work, real close to merge!

I don't know what happened but I messed up my commits so I had to rewrite in to a new commit 😅 I addressed your comments though

packages/core/upload/admin/src/future/components/Drawer.tsx
```diff
@@ -74,7 +74,10 @@ const DrawerContainer = styled(Flex)<DrawerContainerProps>`
   right: 0;
   padding: ${({ theme }) => theme.spaces[2]};
   max-width: 100%;
-  z-index: 1000;
+  /* Use the dialog z-index (320) so popovers (500) and tooltips (1000) rendered
```
jhoward1994: This makes sense but just a note that we should check usages of Drawer across the app and make sure nothing is lower than it's meant to be

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -54,7 +66,7 @@ export const useAssetDetailsParam = () => {
   const openDetails = React.useCallback(
     (id: number) => {
       setIsClosing(false);
-      setQuery({ [URL_PARAM]: String(id) });
+      setQuery({ [URL_PARAM]: String(id) }, 'push', true);
```
jhoward1994, line 83: does the  `...,true)` at the end mean we wont be able to use browser nav buttons to open / close the drawer ??

Adzouz: That's actually the point @markkaylor was raising. I guess it's up to debate but he was confused that he needed to do twice the back action to close the drawer + go to the previous page. Your thoughts?

packages/core/upload/admin/src/future/components/Drawer.tsx
Adzouz: The drawer is only used and scoped in the ML package right now. I think it's safe

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
jhoward1994: I think consistency with the rest of the app should be the priority ?

e.g. in the CM if I open a create relation modal the browser buttons can't be used to close it, they will navigate you away from the entry as usual

So I think this behaviour is fine with `...,true)`

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -161,140 +234,219 @@ interface AssetDetailsProps {
...
+              />
+            </>
+          )}
+
+          <Flex justifyContent="flex-end" gap={2} paddingTop={2}>
+            <Button
+              type="submit"
+              variant="default"
+              loading={isSubmitting}
+              disabled={!modified || isSubmitting}
```
jhoward1994:
```suggestion
// File name is required; block submit when it's empty or whitespace so the API can't 400 on a blank value.                                                                         

disabled={!modified || isSubmitting || !values.name?.trim()}
```

jhoward1994:
What do you think about preventing the closing of the edit drawer if there are unsaved changes? 

On smaller screens I cant see the "save changes" button, it feel like it would be easy to think the changes are auto applied as you type / when you close the drawer

Adzouz:
> What do you think about preventing the closing of the edit drawer if there are unsaved changes?
> 
> On smaller screens I cant see the "save changes" button, it feel like it would be easy to think the changes are auto applied as you type / when you close the drawer

@jhoward1994 
I added the discard confirm dialog!
Also the sticky footer is in the next PR, but I agree with you, I took care of it!

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -136,165 +149,342 @@ const StyledWarning = styled(WarningCircle)`
...
+  onDirtyChange,
+}: {
+  modified: boolean;
+  onDirtyChange?: (dirty: boolean) => void;
+}) => {
+  React.useEffect(() => {
+    onDirtyChange?.(modified);
+  }, [modified, onDirtyChange]);
+  return null;
+};
```
markkaylor: Not sure I understand what this does and why it should be its own component.

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -333,9 +523,10 @@ const DrawerHeader = ({ asset, closeDetails }: DrawerHeaderProps) => {
 interface DrawerContentProps {
   assetId: number;
   closeDetails: () => void;
+  onDirtyChange?: (dirty: boolean) => void;
 }
 
-const DrawerContent = ({ assetId, closeDetails }: DrawerContentProps) => {
+const DrawerContent = ({ assetId, closeDetails, onDirtyChange }: DrawerContentProps) => {
```
markkaylor: This onDirtyChange thing has created some pretty serious prop drilling, is there anyway we could avoid that?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
Adzouz: Because I can't put a useEffect in JSX element directly and I need the modified that is provided by the form component, so I need to create a separated component to put the useEffect in it. Do you think about another solution?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
Adzouz: I could use a context?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
markkaylor: I'm taking a look now to see what's up I'm not loving the FormDirtyBridge

markkaylor (approved):
Looks good to me to move forward nice work 👏

PR #26367, opened 2026-05-18
fix(i18n): preserve non-localized field inheritance 

jhoward1994:
## What does it do?

When you create a new locale draft for a document, non-localized scalar and media fields now stay pre-filled with the source locale's values — even if you switch locales away and come back before saving.

The previous implementation lived in a `useEffect` inside `LocalePickerAction` (`packages/plugins/i18n/admin/src/components/CMHeaderActions.tsx`) and called `setValues` after mount. That side-effect only ran once per locale switch, so revisiting an unsaved locale draft re-initialised the form from the (empty) default and the inherited values vanished.

The fix moves non-localized prefill from a side-effect into `useDocument.getInitialFormValues`. Inherited values are now baked into `initialValues`, which means the `SET_INITIAL_VALUES` reducer in `<Form>` re-applies them every time the form re-initialises (locale switch, cache rehydrate, etc.). 

Scope is limited to scalars and media to match exactly what the server populates into `meta.availableLocales`. Components, dynamic zones, and relations were already filled in by `copyNonLocalizedFields` on the server when the new locale row is first persisted, so no admin-side prefill is needed for them.

## Why is it needed?

The previous behaviour confused content editors working with non-localized fields: they'd fill in a new locale draft, navigate away to check the source, and come back to find their inherited values gone. The shared-source contract was right; only the place it was being applied was wrong.

## How to test it?

1. Check out this branch.
2. `yarn && yarn build`.
3. `cd examples/getstarted && yarn develop`.
4. In the Content-Type Builder, open the `Products` content type, add a Text field named `nonTranslatableField`, and under **Advanced settings** disable localization for it. Save and restart.
5. In the Content Manager, create a new `Products` entry in the default locale (English). Fill in `name` and `nonTranslatableField`, save.
6. Switch the locale picker to `French (fr)` — `nonTranslatableField` should be pre-filled with the English value, `name` should be empty.
7. Without saving, switch back to `English (en)`. Then switch to `French (fr)` again. `nonTranslatableField` should still be pre-filled. On `develop` it would be blank.

## Related issue(s)/PR(s)

CMS-540.


packages/core/content-manager/admin/src/hooks/useDocument.ts
```diff
@@ -213,11 +276,42 @@ const useDocument: UseDocument = (args, opts) => {
...
+
+      // Intentionally use `availableLocales[0]`:
+      // - The server sorts `getAvailableLocales` with the default locale first
+      //   (see `document-metadata.ts`), so index 0 is the canonical inheritance
+      //   source. Sibling locales can drift on non-localized fields because the
+      //   server only syncs them at locale-creation time, not on subsequent
+      //   updates — preferring the default keeps the surfaced values stable.
+      // - Avoids coupling this hook to `useGetLocalesQuery` just to find the
+      //   default locale.
+      const sibling = meta?.availableLocales?.[0] as Record<string, unknown> | undefined;
```
Adzouz, line 293: I'm not sure it's a big deal here (maybe @nclsndr can answer the question 👀) but we're losing the type here and I don't know if it would have an impact for the MCP for example?

Adzouz (approved):
It's pretty clean, nothing really bad to say, good job 👏 I tested and it's working as expected. Just added a small comment if you can answer it 🙏 Otherwise it seems good to me, I'll approve so that we won't need to do it again if the question can be answered without updating anything

packages/core/content-manager/admin/src/hooks/useDocument.ts
jhoward1994: I don't think this will impact the MCP as we're in the admin code here

Adzouz (dismissed):

Adzouz (approved):

PR #26374, opened 2026-05-19
fix(create-strapi-app): keep sqlite driver out of production installs

### LINKED ISSUE #26346: create-strapi-app places better-sqlite3 in 'dependencies' by default, breaking production multi-arch Docker builds on ARM64 under QEMU emulation
### Node Version

22.22.3

### Package Manager

npm

### Package Manager Version

### Strapi Version

5.46.0

### Operating System

Docker/Podman/LXC

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Bug Description

`create-strapi-app` writes `better-sqlite3` into `dependencies` (not `devDependencies`) by default, even when the user selects a non-SQLite database (`postgres`, `mysql`). This is harmless for native local development but breaks **multi-arch production Docker builds** that use QEMU emulation for ARM64:

`better-sqlite3` ships a pre-built `.node` binary for the host architecture and falls back to compiling from source via `node-gyp` for cross-arch installs. The fallback compile runs successfully, but **importing the resulting binary on ARM64 under QEMU emulation triggers a SIGILL (exit code 132)** during the `npm install --omit=dev` step of the runtime stage. The crash is silent on amd64 builds (where the prebuilt binary is used) and only surfaces on arm64 emulated builds.

For Strapi projects that aren't using SQLite as their production DB, `better-sqlite3` is a development-only dependency (used by `strapi develop` for the local SQLite fallback). Moving it to `devDependencies` removes it from the production runtime install and makes ARM64 builds Just Work™.

### Steps to Reproduce

1. `npx create-strapi-app@latest sqlite-test` and answer:
   - Database: `postgres`
   - Typescript: yes
2. Inspect `package.json` → observe `better-sqlite3` is in `dependencies`
3. Build a multi-arch image:

```bash
docker buildx build --platform linux/amd64,linux/arm64 -t test .
```

with a standard multi-stage Dockerfile (the [strapi-tool-dockerize](https://github.com/strapi-community/strapi-tool-dockerize) `Dockerfile-prod` template reproduces this).

4. The `arm64` job fails during the runtime stage's `npm install --omit=dev` step:

```
#25 ERROR: process "/bin/sh -c npm install --omit=dev && npm cache clean --force"
         did not complete successfully: exit code: 132
```

Exit code 132 = SIGILL (illegal instruction) — QEMU-emulated ARM64 importing the freshly-compiled native binary.

### Expected Behavior

When the user selects a non-SQLite database during project scaffolding, `create-strapi-app` should place `better-sqlite3` in `devDependencies` rather than `dependencies`. SQLite stays available for local `strapi develop` (which runs with all deps installed) but is excluded from production runtime installs.

Alternatively (less invasive): document the placement choice in the README so operators building multi-arch images know to move it manually.

### Logs

See "Steps to Reproduce" — the buildx output shows exit code 132 on the linux/arm64 image stage.

### Additional information

**Workaround**: move `better-sqlite3` to `devDependencies` after project creation:

```diff
 "dependencies": {
-  "better-sqlite3": "^9.4.3",
   "pg": "^8.11.3"
 },
 "devDependencies": {
+  "better-sqlite3": "^9.4.3",
   "typescript": "^5.4.5"
 }
```

Confirmed: production image builds clean on both `linux/amd64` and `linux/arm64` (Hetzner CAX / Ampere Altra) after the move.

This relates to https://github.com/strapi-community/strapi-tool-dockerize — but the root cause is in the project template generator, not the Dockerfile generator.

### Confirmation Checklist

- I have checked the existing issues for duplicates.
- I agree to follow the project's Code of Conduct.

cromanenslow:
### What does it do?

Fixes the database dependency classification used by `create-strapi-app`:

- SQLite projects keep `better-sqlite3` in `dependencies`, because it is the selected runtime database driver.
- PostgreSQL/MySQL projects keep their production driver (`pg`/`mysql2`) in `dependencies` and place `better-sqlite3` in `devDependencies` for local development fallback only.
- The 5.1.0 `better-sqlite3` codemod now only moves `better-sqlite3` to `devDependencies` when a non-SQLite production driver is present, so existing SQLite projects are not broken by production installs with `--omit=dev`.

### Why is it needed?

For non-SQLite projects, installing `better-sqlite3` during production image builds can break ARM64 builds under QEMU with SIGILL during `npm install --omit=dev`. Keeping it out of production dependencies avoids compiling/loading that native module in runtime Docker stages, while still preserving local development behavior.

This keeps the SQLite case conservative: if the project actually selected SQLite, `better-sqlite3` remains available for `strapi start` after a production install.

Fixes #26346.

### How to test it?

- `corepack yarn workspace create-strapi-app test:unit`
- `corepack yarn jest --config packages/utils/upgrade/jest.config.js --runInBand --runTestsByPath packages/utils/upgrade/resources/codemods/5.1.0/__tests__/dependency-better-sqlite3.test.ts`
- `PATH="/tmp/strapi-yarn-shim:$PATH" corepack yarn nx run-many -t lint -p create-strapi-app,@strapi/upgrade`

Note: `corepack yarn workspace create-strapi-app test:ts` currently fails in this local checkout because `@strapi/cloud-cli` workspace types are not built/resolved before the package-level typecheck.

---
Fixes CPR-355

### DIFF
--- a/packages/cli/create-strapi-app/jest.config.js
+++ b/packages/cli/create-strapi-app/jest.config.js
@@ -1,0 +1,4 @@
+module.exports = {
+  preset: '../../../jest-preset.unit.js',
+  displayName: 'Create Strapi App',
+};
--- a/packages/cli/create-strapi-app/package.json
+++ b/packages/cli/create-strapi-app/package.json
@@ -48,6 +48,7 @@
     "clean": "run -T rimraf ./dist",
     "lint": "run -T eslint .",
     "test:ts": "run -T tsc --noEmit",
+    "test:unit": "run -T jest",
     "watch": "run -T rollup -c -w"
   },
   "dependencies": {
--- a/packages/cli/create-strapi-app/src/utils/__tests__/database.test.ts
+++ b/packages/cli/create-strapi-app/src/utils/__tests__/database.test.ts
@@ -1,0 +1,62 @@
+import { addDatabaseDependencies } from '../database';
+
+import type { Scope } from '../../types';
+
+const createScope = (client: Scope['database']['client']): Scope =>
+  ({
+    database: { client },
+    dependencies: {
+      '@strapi/strapi': '5.46.0',
+    },
+    devDependencies: {
+      typescript: '^5',
+    },
+  }) as unknown as Scope;
+
+describe('addDatabaseDependencies', () => {
+  test('keeps better-sqlite3 in dependencies when sqlite is the selected database', () => {
+    const scope = createScope('sqlite');
+
+    addDatabaseDependencies(scope);
+
+    expect(scope.dependencies).toMatchObject({
+      '@strapi/strapi': '5.46.0',
+      'better-sqlite3': '12.8.0',
+    });
+    expect(scope.devDependencies).toEqual({
+      typescript: '^5',
+    });
+  });
+
+  test('keeps postgres in dependencies and adds sqlite only as a development dependency', () => {
+    const scope = createScope('postgres');
+
+    addDatabaseDependencies(scope);
+
+    expect(scope.dependencies).toMatchObject({
+      '@strapi/strapi': '5.46.0',
+      pg: '8.20.0',
+    });
+    expect(scope.dependencies).not.toHaveProperty('better-sqlite3');
+    expect(scope.devDependencies).toMatchObject({
+      typescript: '^5',
+      'better-sqlite3': '12.8.0',
+    });
+  });
+
+  test('keeps mysql in dependencies and adds sqlite only as a development dependency', () => {
+    const scope = createScope('mysql');
+
+    addDatabaseDependencies(scope);
+
+    expect(scope.dependencies).toMatchObject({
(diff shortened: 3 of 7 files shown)


oiorain:
Hey @cromanenslow 

innerdvations:
Closing in light of the investigation on https://github.com/strapi/strapi/issues/26346.

**Summary:** we could not reproduce the reported CSA behavior. On Strapi 5, `create-strapi-app` already omits `better-sqlite3` for postgres/mysql — it only adds the selected driver. There is no automatic local SQLite fallback, so adding `better-sqlite3` as a `devDependency` for non-SQLite scaffolds is the wrong direction.

For apps that still have a leftover `better-sqlite3` next to `pg`/`mysql2`, the practical fix is to remove it from production installs (docs/ops guidance), not to change CSA to inject it as a default `devDependency`. The ARM64/QEMU native-module SIGILL case is a separate Docker build concern for projects that actually need SQLite at runtime.

_AI-assisted comment._

PR #26386, opened 2026-05-19
feat(content-manager/mcp): content-type CRUD tools

nclsndr:
### What does it do?

Adds MCP (Model Context Protocol) tool registration to the `content-manager` plugin. At bootstrap time, one set of CRUD tools (`find`, `findOne`, `create`, `update`, `delete`, `publish`, `unpublish`) is dynamically derived for every displayed content type and registered with `strapi.ai.mcp`.

Key technical changes:

- **`derive-content-type-mcp-tools.ts`** — Core derivation logic. For each content type, generates typed Zod input schemas covering: scalar fields, relations (full Strapi `connect`/`disconnect` syntax), components, dynamic zones, and rich-text `blocks` fields. Schemas are narrowed at handler invocation time to only fields and locales the caller's CASL ability permits.
- **`blocks-schema.ts`** — Standalone Zod schema for the Strapi Blocks rich-text format (paragraph, heading, quote, code, list with recursive nesting, image). Used by the input-schema builder for `blocks` attributes.
- **`register-content-manager-mcp-tools.ts`** — Thin bootstrap hook: fetches installed locales from the i18n plugin (if present), resolves displayed content types, calls the derivation layer, and registers each tool.
- **`bootstrap.ts`** — Calls `registerContentManagerMcpTools` during the plugin bootstrap phase, before the MCP HTTP server starts.
- **`permission-checker.ts`** — Typed `userAbility` parameter with the CASL `Ability` type (was `any`). Exports the `ACTIONS` constant so the MCP layer can reference canonical action names without duplication.

Permission enforcement:

- Field-level: permitted fields are resolved from the user's CASL rules and used to strip disallowed attributes from both the Zod input schema and the Document Service call.
- Locale-level: for localized content types, the locale `z.enum` is narrowed to only the locales the user may access for the given action.
- Creator fields: `setCreatorFields` is applied on create/update to attach the MCP session's user as owner.

### Why is it needed?

AI assistants connecting via MCP have no way to read or write Strapi content today. This change exposes content management operations as first-class MCP tools so that agents (Claude, Cursor, etc.) can create, read, update, delete, publish, and unpublish documents directly through the MCP protocol — while staying within the same RBAC boundaries as the REST/GraphQL APIs.

### How to test it?

**Prerequisites**

- Strapi instance with MCP enabled (`strapi.ai.mcp` feature flag on).
- At least one displayed content type (e.g. the `article` type in `examples/empty`).
- Optional: i18n plugin installed with multiple locales to verify locale narrowing.

**Steps**

1. Start the dev sandbox:
   ```bash
   cd examples/empty
   yarn develop
   ```
2. Connect an MCP client (e.g. Cursor MCP, or `curl` against the MCP HTTP endpoint).
3. List available tools — you should see one tool set per displayed content type, named `<apiID>_find`, `<apiID>_find_one`, `<apiID>_create`, etc.
4. Call `article_create` with a valid payload; verify the document is created in the admin panel.
5. Call `article_find` and confirm the returned list matches the admin panel.
6. Call `article_publish` on the created document's `documentId`; confirm the published version appears.
7. Log in as a role with restricted field access; verify the tool schema omits the restricted fields.
8. With i18n enabled, verify the `locale` parameter only exposes locales the user's role is permitted to access.

**Unit tests**

```bash
yarn test:unit --filter content-manager
```

Covers: `blocks-schema`, `derive-content-type-mcp-tools` (input schema derivation, locale narrowing, field filtering, tool name slugification).

### Related issue(s)/PR(s)


innerdvations:
Nice! Only one issue with how permissions are mapped: after `write_*` replaced `create_*`/`update_*`, session gating uses `auth: update` only, so create-only roles get the tool in REST but not MCP (handler already allows create OR update). We could instead gate `write_*` on `create || update` basically `authAnyOf` in MCP session sync

packages/core/content-manager/server/src/mcp/permissions.ts
```diff
@@ -0,0 +1,207 @@
+import { z } from '@strapi/utils';
+import type { Core, Modules, Struct, UID } from '@strapi/types';
+
+import { getService } from '../utils';
+import type { ContentManagerModelForMcp } from './types';
+
+/** Returns true if the content type identified by `uid` has the i18n `localized` plugin option enabled. */
+export const isContentTypeLocalized = (strapi: Core.Strapi, uid: string): boolean => {
```
jhoward1994, line 8: Tiny thought, does this util exist anywhere in strapi/utils ?

packages/core/content-manager/server/src/mcp/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Converts a Strapi content-type UID into a safe MCP tool-name segment.
+ * `api::article.article` → `article`; `plugin::i18n.locale` → `plugin-i18n_locale`.
+ */
+export const slugifyUidForMcpToolName = (uid: string): string => {
+  const [namespace, modelName] = uid.split('::');
+  const modelNameParts = modelName.split('.').map((part) => part.toLowerCase());
+  if (namespace === 'api') {
+    return `${modelNameParts[0]}`;
```
jhoward1994, line 12: Should this actually be `modelNameParts[modelNameParts.length - 1]` ?

jhoward1994 (approved):

packages/core/content-manager/server/src/mcp/permissions.ts
nclsndr: I don't think so, could be refactored for sharing later on.

packages/core/content-manager/server/src/mcp/utils.ts
nclsndr: Good question! For now I only witnessed both values to be the same like `article.article` but I'd be happy to understand how this could vary

nclsndr:
> Only one issue with how permissions are mapped: after `write_*` replaced `create_*`/`update_*`, session gating uses `auth: update` only, so create-only roles get the tool in REST but not MCP (handler already allows create OR update). We could instead gate `write_*` on `create || update` basically `authAnyOf` in MCP session sync

Well spotted @innerdvations! Thank you! 
I'm thinking in refactoring:
```ts
export type McpCapabilityAuth = {
  action: string;
  subject?: string;
};
```

into: 
```ts
/** A single CASL permission check: action + optional subject. */
export type McpAuthPolicy = {
  action: string;
  subject?: string;
};

/**
 * Auth requirement for a non-dev MCP capability.
 * The session gate passes when the user's ability satisfies ANY policy in the array.
 */
export type McpCapabilityAuth = {
  policies: [McpAuthPolicy, ...McpAuthPolicy[]];
};
```

— to be applied on base branch directly

innerdvations:
Yep, that's basically what I was thinking too

PR #26371, opened 2026-05-18
feat(*): introduce MCP server

nclsndr:
### What does it do?

Accumulates [#26174](https://github.com/strapi/strapi/pull/26174), [#26175](https://github.com/strapi/strapi/pull/26175), [#26364](https://github.com/strapi/strapi/pull/26364), [#26386](https://github.com/strapi/strapi/pull/26386)

Introduces `strapi.ai.mcp` — a built-in, stateless [Model Context Protocol](https://modelcontextprotocol.io/) server embedded in Strapi core, authenticated via admin API tokens, with content-manager CRUD tools derived and registered automatically per content type.

**MCP service & lifecycle**

- New `McpService` at `strapi.ai.mcp` with `registerTool()`, `registerPrompt()`, `registerResource()`, `start()`, and `stop()` lifecycle methods.
- Capability definitions are registered by plugins during `register()`. Registration is locked once the server starts.
- A built-in `log` tool (dev-mode only) is registered out of the box.
- Full TypeScript types published under `@strapi/types` (`Modules.MCP.*`).

**Stateless HTTP transport**

- Single endpoint: `POST /mcp`. Each request authenticates, builds a fresh `McpServer` + `StreamableHTTPServerTransport` (no persistent session), handles the JSON-RPC 2.0 payload, then tears down.
- `GET /mcp` and `DELETE /mcp` return `405 Method Not Allowed` with a JSON-RPC error and `Allow: POST` header.
- OAuth well-known discovery routes (`/.well-known/oauth-authorization-server`, `/mcp/register`) return `404` so MCP clients fall back to Bearer token auth.
- Config: `server.mcp.enabled` (default `false`), `server.mcp.connectTimeoutMs` (default `5s`), `server.mcp.requestTimeoutMs` (default `60s`).

**Admin token authentication**

- `createMcpAdminTokenAuthenticator` extracts the `Bearer` token from `Authorization` and delegates to `authenticateAdminToken` on the admin `api-token` service.
- `authenticateAdminToken` validates the token (existence, `kind === 'admin'`, expiry, active owner), updates `lastUsedAt`, and generates a CASL `ability` from `adminPermissions` + owning user roles.
- Strapi's route-level auth is bypassed (`auth: false`); the MCP layer owns authentication entirely.

**Per-request capability RBAC**

- Capability access is declared via a discriminated union `McpCapabilityAccess`: `{ devModeOnly: true }` or `{ auth: { action, subject? } }` evaluated against the token's CASL ability.
[...]

PR #22332, opened 2024-12-03
feat(content-manager): add syntax highlighting in the code block type in the Blocks field original solution

simotae14:
### What does it do?

Adds the highlight colors in the Code blocks inside Block Editor.

### Why is it needed?

We want to highlight the syntax of every single language.
We used Prismjs to do it and this theme https://prismjs.com/index.html?theme=prism-solarizedlight if we want something else just add a comment.

### How to test it?

 User adds a code block in the Blocks field
 - GIVEN THAT I have selected the code block type,
 - WHEN I enter code into the code block
   - THEN the code is displayed with syntax highlighting according to the language detected or specified

 User selects a specific programming language for syntax highlighting
 - GIVEN THAT I am editing a code block in the Blocks field,
 - WHEN I choose a programming language from a dropdown list (e.g., JavaScript, Python, HTML),
   - THEN the code block applies syntax highlighting specific to the selected language

### Related issue(s)/PR(s)

CS-865

PR #18166, opened 2023-09-26
Add new Blocks rich text editor (alpha)

simotae14:
This is the feature branch for work on the blocks rich text editor. Reviewed and QA'd PR's are merged into this branch.

PR #26392, opened 2026-05-20
fix(core/core): codeBlockValidator uses language instead of syntax

nclsndr:
### What does it do?

Renames the optional language field on the `codeBlockValidator` Yup shape from `syntax` to `language` in the server-side entity validator (`packages/core/core/src/services/entity-validator/blocks-validator.ts`, line 83). Updates the two corresponding unit test fixtures in `blocks-validators.test.ts` (lines 195 and 428) to use `language: 'javascript'` instead of `syntax: 'javascript'`.

### Why is it needed?

PR [#20469](https://github.com/strapi/strapi/pull/20469) (Jun 2024) renamed the code-block language field from `syntax` to `language` in `@strapi/types` (`CodeBlockNode.language`) and in the admin Blocks editor (`Code.tsx`), but the server-side Yup validator was not updated. This left a split contract:

- Admin and TypeScript types use `language`
- Entity validator (and its tests) documented `syntax`

As a result, API clients reading the TypeScript types would write `language`, which the validator silently ignored (Yup does not reject unknown keys). Conversely, clients sending `syntax` would pass validation but produce blocks that the admin treats as having no language set (falls back to `plaintext` for highlighting).

This change closes the partial rename and makes all layers agree on `language` as the single field name.

### How to test it?

1. Apply the change and run the entity validator unit tests:
   ```bash
   yarn test:unit packages/core/core/src/services/entity-validator/__tests__/blocks-validators.test.ts
   ```
   All 15 tests should pass.

2. Start the dev sandbox and create or edit a content type with a `blocks` field. In the Blocks editor, add a code block, set a language (e.g. `javascript`), save, and reload. Syntax highlighting and the language selector should both reflect the saved value — confirming the `language` key round-trips correctly through the entity validator.

3. Optionally send a `PUT`/`POST` to the content API with a blocks payload using `language`:
   ```json
   [{ "type": "code", "language": "javascript", "children": [{ "type": "text", "text": "const x = 1" }] }]
   ```
   The request should be accepted (HTTP 200/201) and the stored value should contain `language`.

### Related issue(s)/PR(s)

- PR #20469 — *feat: add language selector to the blocks code editor* — introduced the rename in types and admin but missed the entity validator.
- PR #18166 — *Add new Blocks rich text editor (alpha)* — originally introduced `syntax` in the entity validator.
- PR #22332 — *add syntax highlighting in the code block* — added Prism decoration against `node.language`, widening the split further.
- Required for #26371

### DIFF
--- a/packages/core/core/src/services/entity-validator/__tests__/blocks-validators.test.ts
+++ b/packages/core/core/src/services/entity-validator/__tests__/blocks-validators.test.ts
@@ -192,7 +192,7 @@ const validLists = [
 const validCodeBlock = [
   {
     type: 'code',
-    syntax: 'javascript',
+    language: 'javascript',
     children: [{ type: 'text', text: 'const test = "whatever"' }],
   },
 ];
@@ -425,7 +425,7 @@ describe('Blocks validator', () => {
         validator([
           {
             type: 'code',
-            syntax: 'javascript',
+            language: 'javascript',
             children: [{ type: 'link', text: 'const test = "whatever"' }],
           },
         ])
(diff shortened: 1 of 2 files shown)


markkaylor (approved):
Looks good to me, tested 👍

PR #26409, opened 2026-05-22
feat(server): enhance Koa app configuration with proxy settings

sjoukedv:
### What does it do?

This PR adds an option to the user to configure this from the server configuration.

### Why is it needed?

The default Koa settings are not secure when using proxy configuration that relies on headers because `maxIpsCount` defaults to infinity meaning the proxy ip header can be spoofed.

Taken from the official docs: https://koajs.com/#application

`app.proxyIpHeader` proxy ip header, default to X-Forwarded-For
`app.maxIpsCount`  max ips read from proxy ip header, default to 0 (means infinity)

### How to test it?

When not specified it default to the `koa` defaults, but it gives the option to the user to override the value.

---
Fixes CPR-401


packages/core/core/src/services/server/index.ts
```diff
@@ -18,6 +18,8 @@ const healthCheck: Core.MiddlewareHandler = async (ctx) => {
 const createServer = (strapi: Core.Strapi): Modules.Server.Server => {
   const app = createKoaApp({
     proxy: strapi.config.get('server.proxy.koa'),
+    proxyIpHeader: strapi.config.get('server.proxy.ipHeader'),
+    maxIpsCount: strapi.config.get('server.proxy.maxIpCount'),
```
sjoukedv: We could refactor this to put everything under `server.koa` or `server.proxy.koa`, but that is harder with backwards compatibility. This is quick and simple and lets users at least configure the values.

sjoukedv:
@innerdvations do you have time for a quick review of this one? The change is small, but important impact for users that rely on ratelimiting behind proxies that need verification of the XFF header.

innerdvations (changes requested):
Thanks for the PR — this is a useful gap to close for users behind proxies who need to constrain X-Forwarded-For.

A few small asks before we can merge:

1. **Types** — please extend the `Proxy` interface in `@strapi/types` with the new config keys (`ipHeader` / `maxIpsCount`).
2. **Naming** — rename `server.proxy.maxIpCount` to `server.proxy.maxIpsCount` so it matches Koa’s option name.
3. **Tests** — add a unit test that asserts values from config are applied to the Koa app (`proxyIpHeader` / `maxIpsCount`).

Optional nits (non-blocking): you can pass these into `new Koa({ proxy, proxyIpHeader, maxIpsCount, keys })` instead of assigning them after construction, and a docs PR for the new keys would be appreciated.

Thanks again!

_AI-assisted comment._

packages/core/core/src/services/server/index.ts
innerdvations: Yeah, let's avoid that, we would like to minimize the friction in upgrading to Koa 3 eventually.

sjoukedv:
@innerdvations I have pulled in the changes from upstream and made all the changes except the unit tests, but I think it's fine because koa will complain about wrong values.

innerdvations (changes requested):
Thanks for the follow-up — naming and the `Proxy` keys look good.

One remaining ask before merge: in `@strapi/types`, `Proxy.maxIpsCount` is typed as `string`, but Koa’s `maxIpsCount` (and `createKoaApp`) use a **number**. Please change it to `maxIpsCount?: number` so the public config type matches runtime.

Happy to leave the unit tests for now.

_AI-assisted comment._

sjoukedv:
> Thanks for the follow-up — naming and the `Proxy` keys look good.
> 
> One remaining ask before merge: in `@strapi/types`, `Proxy.maxIpsCount` is typed as `string`, but Koa’s `maxIpsCount` (and `createKoaApp`) use a **number**. Please change it to `maxIpsCount?: number` so the public config type matches runtime.
> 
> Happy to leave the unit tests for now.
> 
> _AI-assisted comment._

@innerdvations that's my bad it's fixed now

innerdvations (approved):
Thanks, tested and LGTM now!

PR #26412, opened 2026-05-22
fix(content-manager): guard md.render() null in PreviewWysiwyg

### LINKED ISSUE #26397: Cannot read properties of null (reading 'replaceAll')
```
TypeError: Cannot read properties of null (reading 'replaceAll')
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-LBM56NC2.js?v=b1b12496:90945:87
    at mountMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:12194:27)
    at Object.useMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:12518:24)
    at Object.useMemo (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-TJ2ONGJW.js?v=b1b12496:1094:29)
    at PreviewWysiwyg (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-LBM56NC2.js?v=b1b12496:90945:24)
    at renderWithHooks (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:11548:26)
    at mountIndeterminateComponent (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:14926:21)
    at beginWork (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:15914:22)
    at beginWork$1 (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:19753:22)
    at performUnitOfWork (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-YNSCI7QL.js?v=b1b12496:19198:20)
```

guangyang1206:
PreviewWysiwyg calls md.render() and then chains
`.replaceAll()` on the result. When the markdown
renderer returns `null` (e.g. on malformed input), this crashes with:

  TypeError: Cannot read properties of null (reading 'replaceAll')

Guard the call so sanitizeHtml always receives a string.

Fixes #26397

### Problem

`PreviewWysiwyg` calls `md.render()` and then chains `.replaceAll()` on the result. When the **markdown renderer returns `null`** (e.g. on malformed/empty WYSIWYG input), this crashes with:

```
TypeError: Cannot read properties of null (reading 'replaceAll')
    at PreviewWysiwyg (chunk-LBM56NC2.js:90945:87)
```

**Stack trace (from #26397):**
```
at PreviewWysiwyg (...)
at mountMemo
at Object.useMemo
```

The root cause: `data ?? ''` only handles `data` being `null`, but `md.render()` itself can return `null` for certain malformed markdown inputs. The result is then passed to `.replaceAll()`, which throws.

### Solution

Guard the `md.render()` return value **before** passing it to `sanitizeHtml`:

```tsx
const html = React.useMemo(
  () => {
    const rendered = md.render((data ?? '').replaceAll('\\n', '\n'));
    return sanitizeHtml(typeof rendered === 'string' ? rendered : '', {
      ...sanitizeHtml.defaults,
      allowedTags: false,
      allowedAttributes: {
        '*': ['href', 'align', 'alt', 'center', 'width', 'height', 'type', 'controls', 'target'],
        img: ['src', 'alt'],
        source: ['src', 'type'],
      },
    });
  },
  [data]
);
```

Now `sanitizeHtml` **always receives a string**, even when `md.render()` returns `null`/undefined.

### Impact

- ✅ No more runtime crash in the WYSIWYG preview
- ✅ Graceful fallback (empty string) instead of a white screen
- ✅ No behavior change when `md.render()` returns a normal string
- ✅ Follows the same defensive pattern used elsewhere in the codebase

### Verification

```bash
cd packages/core/content-manager
pnpm test -- --filter PreviewWysiwyg
```

Manual testing:
1. Open any content type with a WYSIWYG field
2. Try entering malformed markdown / empty content
3. Preview should render empty (not crash)

Fixes #26397

---
Fixes CPR-361

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/PreviewWysiwyg.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Wysiwyg/PreviewWysiwyg.tsx
@@ -11,8 +11,9 @@ interface PreviewWysiwygProps {
 
 const PreviewWysiwyg = ({ data }: PreviewWysiwygProps) => {
   const html = React.useMemo(
-    () =>
-      sanitizeHtml(md.render((data ?? '').replaceAll('\\n', '\n')), {
+    () => {
+      const rendered = md.render((data ?? '').replaceAll('\\n', '\n'));
+      return sanitizeHtml(typeof rendered === 'string' ? rendered : '', {
         ...sanitizeHtml.defaults,
         allowedTags: false,
         allowedAttributes: {


unrevised6419:
I did check the types of markdown-it and [`Markdown#render`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ce6d550f9ae837b90d47e86cf60228fbd86ba06a/types/markdown-it/dist/index.cjs.d.ts#L1126) returns a `string` only.
Also did check [the source code](https://github.com/markdown-it/markdown-it/blob/b4a9b659ef5734223731cfaa3ad5eacc6fc22918/lib/renderer.mjs#L304) and it always returns a string also. 🤔 

Could you please provide an example of malformed input so this could be tested?

PR #26421, opened 2026-05-23
fix(content-manager): guard repeatable field .map() crash on relation…

### LINKED ISSUE #26401: Content Manager relation/modal edit closes with “f.map is not a function” error
When editing a related entry via the Content Manager (repeatable) relation field modal in Strapi, the modal opens correctly and the entry can be edited and saved without issues. However, when closing the modal, the admin UI crashes with the following error:
`f.map is not a function. (In 'f.map(...)', 'f.map' is undefined)`

**Steps to reproduce**

1. Open Content Manager
2. Edit an entry containing a relation field (repeatable)
3. Click the related entry to open it in the modal editor
4. Save changes (optional)
5. Close the modal

**Expected behavior**

Modal closes cleanly and returns to the content view without errors.

**Actual behavior**

UI crashes on modal close with:
-  f.map is not a function
- stack trace points to repeatable field rendering (type: "repeatable")

**Notes**

- Appears to happen during re-render after modal close
- Likely related to repeatable/dynamic zone state becoming undefined or non-array
- Might affect only "repeatable" relations
- Occurs in Content Manager admin UI

**Strapi version**

5.46.1

**Node version**

20.20.1

subhaushsingh:
### What does it do?
Adds an `Array.isArray` guard in `Repeatable.tsx` to normalize the field value to `[]` when it is `null` or `undefined`, and skips rendering `RepeatableComponent` in `Input.tsx` when `field.value` is not an array.

### Why is it needed?
The `useField` destructuring default `value = []` only fires when the value is `undefined`, not `null`. When a relation modal closes, the form state briefly sets the repeatable field to `null`, causing `value.map()` and `value.length` to throw and crash the admin UI.

### How to test it?
1. Open Content Manager and edit an entry that has a repeatable component field containing a relation
2. Click a related entry to open it in the modal editor
3. Save or discard changes and close the modal
4. The UI should return to the edit view cleanly with no crash

### Related issue(s)/PR(s)
Fixes #26401

---
Fixes CPR-365


innerdvations (dismissed):
Thanks!

subhaushsingh:
@innerdvations The previous e2e failures were caused by the Array.isArray guard I had in Input.tsx preventing the Repeatable component from rendering. I've reverted that specific file and kept the fix isolated to Repeatable.tsx. The tests should be green now

innerdvations (approved):
It looks like there was still a problem, I've fixed it and added a test but please confirm that it still looks good to you before I merge.

subhaushsingh:
Looks great to me! Thanks for catching my typo and stepping in to add the useMemo hook and the tests.

innerdvations (approved):

PR #26422, opened 2026-05-23
feat(admin): add documentation helper link in HeaderLayout (revival of #23328)

unrevised6419:
> Revival of #23328 by @Mcastres — rebased on current develop, with all open review feedback addressed and additional improvements. Closes #23328.

### What does it do?

Adds a `?` IconButton next to the primary action in every page's `HeaderLayout`. Clicking it opens the matching docs.strapi.io page for that admin section in a new tab and fires a `didClickOnDocLink` tracking event with the originating route and the destination URL.

Changes beyond the original PR:

- **Match algorithm**: replace the hand-rolled `find`/`startsWith` lookup with `react-router`'s `matchRoutes`, so path entries are typed `RouteObject`s with `/*` splats and benefit from proper specificity ranking (no more array-order dependence between e.g. `/settings/users/*` and `/settings/users-permissions/*`).
- **Sync derivation**: drop the unnecessary `async` from `getMatchingDocLink` and derive `docLink` directly from `useLocation().pathname` in the component (no `useState`/`useEffect`). Memoized with `React.useMemo` against `location` to avoid recomputing on scroll/resize re-renders.
- **Canonical doc URLs**: every original link redirected (`/user-docs/*`, `/dev-docs/*` → `/cms/*`). Updated all 25 unique URLs to the current canonical paths. Restored 4 section anchors that were dropped on redirect (`#providers`, `#email-templates`, `#advanced-settings`, profile anchor on admin-panel) and disambiguated `/cms/features/rbac` with `#configuration` / `#usage` for `/settings/roles` vs `/settings/users`.
- **Path corrections**: aligned array entries with actually registered admin routes — `content-type-builder` / `content-releases` mount under `/plugins/...`, `purchase-content-releases` and `list-plugins` are under `/settings/...`. Dropped `/marketplace` (external link, never reaches HeaderLayout).
- **Opt-out flag** (per @derrickmehaffy's comment): added `admin.flags.docLinks` (default `true`). Mirrors the existing `nps` / `promoteEE` flag pattern. White-label deployments can disable all doc links with:
  ```ts
  // config/admin.ts
  flags: { docLinks: false }
  ```
  Type added to `Core.Config.Admin['Flags']` (server) and `BrowserStrapi['flags']` (client). All example/template `config/admin.{ts,js}` files updated to expose `FLAG_DOC_LINKS` env override.
- **Layout polish** (per @remidej review): doc link button has no own padding — instead a `gap` is set on the parent `Flex` wrapping `docLinkButton` + `primaryAction`, in both sticky and non-sticky variants.

### Why is it needed?

Original motivation from #23328 stands: gives users a one-click jump to relevant docs for the page they're on, and lets the team measure (via tracking) which sections users feel lost in.

### How to test it?

1. Start any example project (e.g. `examples/kitchensink-ts`):
   ```bash
   yarn build && yarn develop --watch-admin
   ```
2. Navigate the admin to e.g. Content Manager, Media Library, Settings → API Tokens, Settings → Users & Permissions → Providers, profile (`/me`).
3. A `?` IconButton should appear next to the primary action in the page header. Click it → new tab opens to the matching docs.strapi.io page.
4. Scroll a long page (e.g. `/me`) until the sticky header appears → doc link button should also appear in the sticky header.
5. To verify the opt-out flag:
   ```ts
   // config/admin.ts
   flags: { docLinks: false }
   ```
   Restart, confirm the `?` button is no longer rendered on any page.

### Related issue(s)/PR(s)

- Closes #23328 (original PR by @Mcastres)
- Addresses review feedback from @remidej and @jhoward1994
- Addresses white-label request from @derrickmehaffy

### DIFF
--- a/examples/complex/config/admin.ts
+++ b/examples/complex/config/admin.ts
@@ -16,6 +16,7 @@ const adminConfig = ({ env }) => ({
   flags: {
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
+    docLinks: env.bool('FLAG_DOC_LINKS', true),
   },
 });
 
--- a/examples/empty/config/admin.ts
+++ b/examples/empty/config/admin.ts
@@ -18,6 +18,7 @@ const adminConfig = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Admi
   flags: {
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
+    docLinks: env.bool('FLAG_DOC_LINKS', true),
   },
 });
 
--- a/examples/experimental-dev/config/admin.ts
+++ b/examples/experimental-dev/config/admin.ts
@@ -15,6 +15,7 @@ const adminConfig = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Admi
   flags: {
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
+    docLinks: env.bool('FLAG_DOC_LINKS', true),
   },
 });
 
--- a/examples/getstarted/config/admin.js
+++ b/examples/getstarted/config/admin.js
@@ -32,6 +32,7 @@ module.exports = ({ env }) => ({
   flags: {
     nps: env.bool('FLAG_NPS', true),
     promoteEE: env.bool('FLAG_PROMOTE_EE', true),
+    docLinks: env.bool('FLAG_DOC_LINKS', true),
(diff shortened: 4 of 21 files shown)


unrevised6419:
@innerdvations can you also please take a look at https://github.com/strapi/strapi/pull/26213 so we could colocate all the `window.strapi` types?

innerdvations (dismissed):
LGTM, thanks again! Just need Product approval from @emilieringwald before merging

innerdvations:
@pwizla on this the only docs change should be a single line in the config about how to disable this

innerdvations:
I just pushed some tests, hope you don't mind!

innerdvations (approved):

PR #26425, opened 2026-05-25
fix(data-transfer): transfer admin menu and auth logos with configuration

### LINKED ISSUE #20278: Strapi loses the Admin Menu and Auth logo on transfer

- Node.js version: lts/iron (v20.12.2)
- NPM version:  v10.5.0
- Strapi version: 4.20
- Database: postgres
- Operating system: linux / debian 

### Describe the bug

Using `npm run strapi transfer to=...`  causes the remote Strapi to lose any configured custom logo for the  Admin menu and Auth. I also noticed this behaviour earlier today when I used `strapi export` and then `strapi import`  to migrate from sqllite to posgtres. I put it down at the time to some difference between sqllite and postgres. However, the same behaviour occured on the transfer between two postgres backed strapi instances as well.

1. Install strapi on a remote, and get the transfer token. 
2. Also set a custom Menu logo and Auth logo
3. Do exactly the same with a local strapi instance. I'm unsure if you need to have a custom logo on the source strapi to trigger this behaviour. I have always had a custom logo when I've noticed it.
4. Use `npm run strapi transfer to=..` from the local strapi to the freshly created remote with a custom logo.

### Expected behavior

I expect one of the two to happen :
 - The  `to` target gets the logo from the `from` source. 
 - The `to` target retains whatever logo was originally configured on it, making its behaviour similar to what happens with admin accounts. 

### What actually happens

The `to` target loses the logo. It does not seem to revert to the default logo, instead showing the empty box of a missing image src.

### Additional context

This is something of a pain when it comes to moving data around. It isn't a serious issue as such, I'm sure we can live with the Strapi logo and avoid the problem entirely. That said, I would still characterise this as a rather annoying quality of life bug.

innerdvations:
### What does it do?

Embeds admin menu and auth logo file contents in the data-transfer configuration stage when exporting `core_admin_project-settings`, and re-uploads those files via the upload provider on import. Transfer-only `__transferBuffer` fields are stripped before persisting to core-store.

### Why is it needed?

Fixes #20278. Admin logos store metadata in core-store but the image files live outside the media library (`plugin::upload.file`), so the assets stage never copied them. After transfer, export/import, or import, the destination had logo URLs pointing at missing files (broken images in the admin UI).

### How to test it?

1. On a source Strapi instance, set custom menu and auth logos in Settings → Global settings → Overview.
2. Run `yarn strapi transfer --to …` (or export/import) to a destination instance.
3. Confirm the destination admin panel shows the transferred logos (not empty image boxes).

Unit tests:

```bash
cd packages/core/data-transfer && yarn test:unit project-settings-logos.test.ts
```

### Related issue(s)/PR(s)

Fixes #20278


jhoward1994:
This looks good to me, I haven't managed to do a manual QA yet though

jhoward1994:
QA's and it's working.

One small concern:
- Will the source logo's `provider` get copied to the destination? & therefore potentially mismatch with the destination if it uses a different provider? - Not sure that is relevant for app logos outside of the media lib ?
+ also the conflicts with develop

otherwise this looks good

jhoward1994 (dismissed):

innerdvations:
@jhoward1994 You were right, there was an issue with using the wrong provider, it was using source instead of destination. I've fixed it, can you test again with a remote provider and make sure the behaviour is working like you would expect?

jhoward1994 (approved):

PR #26431, opened 2026-05-25
chore(ai/skills): add commit conventions

cache-your-dreams:
### What does it do?
Informs AI of our Strapi commit conventions

### Why is it needed?
Helping AI when creating a commit by guiding on strapi conventions

### How to test it?
Get an agent to commit something.

### Important change
After this change, contributors can keep their personal Claude Code setup local without it ever showing up in git status:

- `.agents/settings.local.json`, your local permission allowlist and per-user settings stay untracked.
- `.agents/agents/`, custom subagents you define for your own workflow stay untracked.

- `.agents/commands/`, `.agents/hooks/`, `.agents/output-styles/`, `.agents/mcp.json`, `.agents/worktrees/`
- same: anything under `.agents/` that isn't an explicitly-shared skill is ignored by default.

- `.agents/skills/local/` and `.claude/skills/local/`, reserved as private workspaces for skill drafts and personal skills you don't want to commit.
- `CLAUDE.local.md` — gitignored for personal notes (unchanged).

Shared content lives in exactly two places:

- `.agents/skills/<name>/SKILL.md` — canonical skill content, tracked.
- `.claude/skills/<name>` — symlink (mode 120000) re-exposing the skill to Claude Code, tracked.
Adding a new committed skill is just git add `.agents/skills/<name>/` (plus the symlink) — no .gitignore edit required. Anything outside `local/` is tracked by default, so keep drafts under `local/`.


.agents/skills/commit/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+| `chore`       | Internal cleanup, tooling, refactor with no behavior change, dep bumps |
+| `docs`        | Documentation only                                                     |
+| `test`        | Adding or updating tests                                               |
+| `ci`          | CI/CD pipeline changes                                                 |
+| `security`    | Security fixes or hardening (often `security(deps): ...`)              |
+| `revert`      | Reverting a previous commit                                            |
+| `release`     | Release commits (reserved for release tooling)                         |
+| `future`      | Work behind a future flag                                              |
+
+> Note: `CONTRIBUTING.md` shows `doc:` in one example; the enforced type is `docs`. Use `docs`.
```
nclsndr: We should fix the original source instead of re-steering here

.ai/skills/git-conventions/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+| `security`    | Security fixes or hardening (often `security(deps): ...`)              |
+| `revert`      | Reverting a previous commit                                            |
+| `release`     | Release commits (reserved for release tooling)                         |
+| `future`      | Work behind a future flag                                              |
+
+> Note: `CONTRIBUTING.md` shows `doc:` in one example; the enforced type is `docs`. Use `docs`.
+
+Types **not** in the list (and therefore rejected by commitlint): `perf`, `refactor`, `style`, `build`, `improvement`, `wip`.
+
+> Note: `refactor` and `perf` are rejected as **type prefixes**, but the underlying work is still valid — route it through `enhancement` (when it improves the product/feature) or `chore` (when it's internal housekeeping with no product change). See the Decision Guide below.
```
nclsndr, line 45: Why are we forbidding `refactor` and `perf` at the first place?

.agents/skills/committing/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+> Note: `refactor` and `perf` are rejected as **type prefixes**, but the underlying work is still valid — route it through `enhancement` (when it improves the product/feature) or `chore` (when it's internal housekeeping with no product change). See the Decision Guide below.
+
+## Subject Rules (from CONTRIBUTING.md)
+
+The subject summarises **what** the commit is about, not **how** the code achieves it.
+
+- `feat: <what the feature is>`
+- `fix: <what the problem is>` — describe the bug, not the fix
+- `chore: <what the PR is about>`
+- `docs: <what is documented>`
```
nclsndr: I think we should provide a pattern to preferably use scopes as well. 
E.g: `fix(core/core): xxx`

.agents/skills/committing/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+```
+feat: introduce document service
+fix: unable to publish documents due to missing permissions
+chore: refactor data-fetching in EditView to use react-query
+docs: document service API reference
+```
+
+## Tooling
+
+Run the `commit` script (`pnpm commit` / `yarn commit`) to use the interactive commit CLI that helps construct a compliant message.
```
nclsndr: I would not prescribe anything here. No need to teach LLMs how to use git cli imo

.gitignore
```diff
@@ -155,9 +155,14 @@ tests/cli/.env
...
-.agents/local-skills
-.agents/local-state
+# Track committed skills under .claude/skills/ and .agents/skills/, but reserve
+# `local/` as a private workspace for drafts and personal skills.
+.claude/*
+!.claude/skills
+.claude/skills/local/
+.agents/*
+!.agents/skills
+.agents/skills/local/
```
nclsndr: We should not nest local skills under a tracked skills dir. 
By convention, skills dir are 1 level deep.
I strongly prefer to have N skills dir at the same level in FS

AGENTS.md
```diff
@@ -220,11 +220,16 @@ yarn prettier:check # check only
...
+- PR description must follow [.github/PULL_REQUEST_TEMPLATE.md](.github/PULL_REQUEST_TEMPLATE.md) — do not invent your own sections.
+
+---
+
+## Agent Skills & Workspace
+
+- **Canonical location** — agent skills live in `.agents/skills/<name>/SKILL.md`. `.claude/skills/<name>` is a symlink that re-exposes a skill for Claude Code; the symlink mode must be `120000` (committed symlink), not a regular file.
+- **`.agents/skills/local/`** and **`.claude/skills/local/`** — gitignored private workspaces. Put drafts and personal skills you don't want to commit here.
+- **`CLAUDE.local.md`** — gitignored personal notes.
+- **Adding a new committed skill** — drop it under `.agents/skills/<name>/` and (optionally) symlink it into `.claude/skills/<name>` with mode `120000`. No `.gitignore` edit needed. Anything outside `local/` is tracked by default, so keep drafts in `local/`.
```
nclsndr: AI tools know / define their own way of working with skills. 
If we want to impact this, it should be done through setup scripts rather than AGENTS.md steering.

.ai/skills/git-conventions/SKILL.md
cache-your-dreams: These aren't being forbidden by this PR — `.commitlintrc.ts` has always restricted type-enum to a curated set (it overrides config-conventional's defaults), and that file isn't changed here. The list is shaped around our changelog/GitHub-label categories rather than a dev-level taxonomy, so perf/refactor route through enhancement (user-perceivable improvement) or chore (internal, no product change). The skill documents that routing.

You should launch a separate discussion if we want to reconsider the type list itself.

.agents/skills/commit/SKILL.md
soupette: that should be a rule as well... agents are lazy...

.ai/skills/git-conventions/SKILL.md
nclsndr: No strong opinion here, I was mostly seeking for the rationale. Thanks for the detailed answer @cache-your-dreams!

.agents/skills/committing/SKILL.md
cache-your-dreams: I'm fine removing this personally

nclsndr (approved):

PR #26436, opened 2026-05-26
future(upload): delete and replace file

Adzouz:
### What does it do?

Adds the delete and replace actions on the asset details drawer of the new media library:

- **Delete**: trash IconButton in the footer of the drawer with an AlertDialog confirm guard. When confirming, it's triggering the deletion of the file and closes the drawer + fires a success notification on success.
- **Replace**: replace IconButton + confirm AlertDialog → opens the native file chooser. Uploads via `PUT /upload?id=...`, drawer stays open and the details refresh via RTK Query cache invalidation. Success/error toast is rendered inline above the preview (in-drawer Alert).

https://github.com/user-attachments/assets/24eac26a-7e6c-4155-8126-4068f528e2c0

### Why is it needed?

We need to be able to perform the same actions of a file as the current media library. In this PR, we're adding existing capabilities for asset deletion and replacement.

### How to test it?

1. Run with `UNSTABLE_MEDIA_LIBRARY=true yarn develop` and open `/admin/plugins/unstable-upload`.
2. Click an asset → drawer opens.
3. Click the replace icon → confirm → pick a new file → in-drawer toast "File replaced", metadata updates.
4. Click the trash icon → confirm → asset disappears from the list and the drawer closes.

### Related issue(s)/PR(s)

Implements CMS-124

🚀


packages/core/upload/server/src/controllers/admin-upload.ts
```diff
@@ -103,6 +103,22 @@ export default {
...
+    if (replacedFile?.mime?.startsWith('image/') && (await aiMetadataService.isEnabled())) {
+      try {
+        const metadataResults = await aiMetadataService.processFiles([replacedFile]);
+        await aiMetadataService.updateFilesWithAIMetadata([replacedFile], metadataResults, user);
+      } catch (error) {
+        strapi.log.warn('AI metadata generation failed on replace, proceeding without it', {
+          error: error instanceof Error ? error.message : String(error),
+        });
+      }
+    }
```
markkaylor, line 120: Oh nice catch!

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -294,159 +577,191 @@ export const AssetDetails = ({ asset }: AssetDetailsProps) => {
...
+              {drawerToast ? (
+                <DrawerToastSlot>
+                  <Alert
+                    variant={drawerToast.type === 'success' ? 'success' : 'danger'}
+                    closeLabel={formatMessage({ id: 'global.close', defaultMessage: 'Close' })}
+                    onClose={() => setDrawerToast(null)}
+                  >
+                    {drawerToast.message}
+                  </Alert>
+                </DrawerToastSlot>
```
markkaylor: We need a custom alert because otherwise it is centered in the viewport? Or how come we aren't using useNotification?

markkaylor (commented):
Nice work!

It's my bad because on the ticket I said delete and replace but I think actually in the footer it's delete and download . 


Image

It looks like replace is actually part of the asset preview with the arrows icon. Maybe that changed after I planned the ticket 🤷‍♂️ . Sorry about that. But also is there a reason why we don't also see copy link and download implemented here on this PR (even though replace is technically not a footer action now)?

Image

markkaylor:
At least for replace I think we need a loading state in the asset preview while the image is being processed

Adzouz:
@markkaylor Ooooooh wow you're totally right and I didn't think that it was weird haha 🤦‍♂️
I'll move the icon to the image preview instead, no biggies! (since it's already implemented). Also I"ll add the copy link + download button to this PR

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
Adzouz: Yes that's right, couldn't find a better solution but to implement it myself here. Do you have any idea?

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
```diff
@@ -227,12 +306,350 @@ const LocationField = ({ label, rootLabel, folders }: LocationFieldProps) => {
...
+    if ('error' in res) {
+      const error = res.error as { data?: { error?: { message?: string }; message?: string } };
+      const message =
+        error?.data?.error?.message ??
+        error?.data?.message ??
+        formatMessage({
+          id: getTranslationKey('asset-details.delete.error'),
+          defaultMessage: 'Failed to delete the asset.',
+        });
+      toggleNotification({ type: 'danger', message });
```
markkaylor: how come here we use toggleNotification but eleswhere it looks like we are using `notify` for the custom alert?

markkaylor (commented):
Just one questions and I still need to test

markkaylor:
Tested and everything works nicely, thanks for the updates. 

One UX comment to add: 

I think we need a loader for the replace no?

https://github.com/user-attachments/assets/db13c535-159a-4b2c-a85e-42f1fdfa2b42

packages/core/upload/admin/src/future/pages/Assets/components/AssetDetails/AssetDetailsDrawer.tsx
markkaylor: oh got it, I forgot when you delete it closes the drawer so yeah makes sense to use this one here!

Adzouz: No no, you were actually right, in that case the drawer stays opened because it failed so it should be a notify instead

Adzouz:
Hey @markkaylor I added a loader for both replace and delete actions and added a new context for handling operations on an asset with mutations hooks. Let me know what you think!

packages/core/upload/admin/src/translations/en.json
```diff
@@ -28,20 +28,21 @@
...
   "asset-details.replace.success": "File replaced.",
   "asset-details.replace.error": "Failed to replace the file.",
+  "asset-details.replace.loading": "Replacing the file…",
   "asset-details.copy-link.trigger": "Copy link",
   "asset-details.copy-link.success": "Link copied.",
   "asset-details.copy-link.error": "Failed to copy the link.",
   "asset-details.download.trigger": "Download",
   "asset-details.download.error": "Failed to download the file.",
   "asset-details.noPreview": "No preview available",
-  "plugin.home": "Home",
```
Adzouz, line 44: This was in double

packages/core/upload/admin/src/future/utils/downloadFile.ts
markkaylor, line 1: I'm ok with this but it's pretty far from where it is used I think it could probably be defined directly in the place where it is used.

markkaylor (approved):
Nice work!

packages/core/upload/admin/src/future/utils/downloadFile.ts
Adzouz: That's because I was thinking that it could be reused for bulk download maybe? Not sure if the option exists though

markkaylor: Makes sense, let's leave it

PR #26370, opened 2026-05-18
fix(homepage): homepage count-documents slow on large D&P tables

markkaylor:
### What does it do?

Rewrites `getCountDocuments` in the Content Manager homepage service. The previous implementation ran three separate queries per D&P content type (one self-join per `{draft, published, modified}` bucket). The new code classifies each `document_id` in a single GROUP BY pass — one scan, one connection per content type instead of three.

Also fixes a correctness bug where `modified` was capped at 0 or 1 due to a stray `.groupBy('draft.document_id').first()`.

### Why is it needed?

[#25200](https://github.com/strapi/strapi/issues/25200) — admin homepage takes tens of seconds to load on production-scale D&P tables. Prior partial fixes only skipped the queries for non-D&P content types; the queries themselves were still expensive on D&P tables.

### How to test it?

You need a populated D&P table.

**Setup** (one-time):

```bash
cd examples/complex
yarn db:start:mysql
yarn seed:v5 -- --multiplier 100
```

Run the following to create ~200k rows of `basic_dps`

```bash
docker exec -i strapi_complex-mysql-1 mysql -ustrapi -pstrapi strapi <<'SQL'
INSERT INTO basic_dps (document_id, string_field, created_at, updated_at, published_at, locale)
SELECT
  CONCAT('seed-', LPAD((rn DIV 2) + 1, 12, '0')),
  CONCAT('s-', rn), NOW(), NOW(),
  IF(rn % 2 = 0, NOW(), NULL), 'en'
FROM (
  SELECT ROW_NUMBER() OVER () AS rn
  FROM information_schema.columns a, information_schema.columns b
  LIMIT 200000
) t;
SQL
```

**Measure on `develop` and on this PR.** 

1. Switch to develop
2. yarn build
3. cd examples/complex
4. yarn develop:mysql
5. Get JWT token from a network request
5. Run the following, and save the times

```bash
export TOKEN='Bearer eyJ...'
for i in 1 2 3 4 5; do
  curl -s -o /dev/null -H "Authorization: $TOKEN" \
    -w '%{time_starttransfer}s\n' \
    http://localhost:1337/content-manager/homepage/count-documents
done
```
6. Switch to this branch
7. Repeat 2-5

Local results (208k rows, MySQL 8):

| | Warm median | Cold |
|---|---|---|
| develop | 767 ms | 797 ms |
| this PR | **182 ms** | 331 ms |

[...]

PR #26438, opened 2026-05-26
chore: fix dependabot cooldown config for github-actions

innerdvations:
### What does it do?

Removes unsupported `semver-patch-days`, `semver-minor-days`, and `semver-major-days` from the `github-actions` entry in `.github/dependabot.yml`. The GitHub Actions ecosystem only supports `cooldown.default-days`.

### Why is it needed?

Dependabot validation fails on the current config with errors that those semver cooldown properties are not supported for the `github-actions` package ecosystem.

### How to test it?

Confirm Dependabot no longer reports configuration errors for `.github/dependabot.yml` on the repository settings / Dependabot tab.

See https://github.com/strapi/strapi/pull/26370/checks?check_run_id=77839357994

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,16 @@
 # Dependabot policy for the Strapi monorepo:
 # - Security advisories: open immediately; npm families grouped, others individual.
 # - Version updates: max 10 open npm PRs / 3 actions PRs, weekly on Sundays.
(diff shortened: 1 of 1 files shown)


nclsndr (approved):
Sneaky dependabot!

PR #26439, opened 2026-05-26
security(deps): patch uuid and qs DoS advisories

jhoward1994:
### What does it do?

Force uuid to 11.1.1 and qs to 6.15.2 via Yarn resolutions in both the root monorepo and docs/ lockfiles


nclsndr (dismissed):
LGTM 👌

docs/package.json
```diff
@@ -28,7 +28,9 @@
   "resolutions": {
     "@babel/helpers": "7.29.2",
     "@babel/runtime": "7.29.2",
-    "@babel/runtime-corejs3": "7.29.2"
+    "@babel/runtime-corejs3": "7.29.2",
+    "uuid": "11.1.1",
+    "qs": "6.15.2"
```
innerdvations: This forces the 12, 13, 14 to all resolve to 11.1.1

Would it be a problem to upgrade our own usage internally to uuid v14?

And can we remove the `qs` resolution or is it necessary to bypass a parent dep that requires it?

innerdvations (changes requested):

docs/package.json
jhoward1994: afaik this current diff is what we need. I'm not a wizz with dependencies though so let me know if something seems off

1. Does forcing uuid 12/13/14 down to 11.1.1 cause a problem?
afaik no. Nothing we wrote uses uuid directly. The only things pulling in uuid are other libraries we depend on, and they all use the basic part of uuid that hasn't changed. The one library that wanted a newer uuid (mermaid, used only for building docs) explicitly accepts 11.1.1 too. So pinning everything to 11.1.1 doesnt break anything.

2. Would it be a problem to upgrade our own usage to uuid v14?
We don't use uuid anywhere in our own code

3. Can we remove the qs resolution, or is it needed?  
It's needed. express and body-parser ask for a version of qs that's stuck in the broken range. The resolution is the only thing that forces them onto the fixed version. The bumps in our own package.json only fix the places where we list qs ourselves.

jhoward1994:
We can close this for https://github.com/strapi/strapi/pull/26417, is that right @innerdvations ?

innerdvations:
I believe so, but could you check that there aren't any vulnerable versions of uuid in the yarn.lock and create a new PR for just that if needed?

jhoward1994:
I think this diff is what we need to handle uuid in the yarn lock. In the lock files develop has vulnerable uuid@8.3.2 and 9.0.1. This PR forces everything to 11.1.1 (patched per the advisory) — only that version remains after yarn install.

We need resolutions because uuid is purely transitive (@apollo/server, grant/purest → request-oauth/request-multipart) 

@innerdvations

innerdvations:
I updated the PR to use more precise resolutions instead of blanket `uuid` / `qs` resolutions.

The important bit is:
- `uuid@^8.3.2` and `uuid@^9.0.0` are pinned to `11.1.1` because the patched version is outside those ranges.
- `qs@~6.14.0` is pinned to `6.15.2` because the tilde range cannot reach it.
- The `qs@^6.x` ranges do not need resolutions; reinstalling lets them naturally resolve to `6.15.2`.
- Docs’ `uuid@^11.1.0 || ^12 || ^13 || ^14.0.0` stays on `14.0.0` because it is not vulnerable, so we no longer force-downgrade it.

I verified no vulnerable `uuid` or `qs` versions remain in root or docs lockfiles, and `yarn install --immutable` passes in both.

innerdvations (approved):

PR #26453, opened 2026-05-27
fix(i18n): preserve non-localized prefill when revisiting locale drafts

innerdvations:
## Summary

- Move non-localized field inheritance into `useDocument.getInitialFormValues()` so shared fields stay populated when switching back to an unsaved locale draft (fixes revisit regression after #24659).
- Remove the one-shot `setValues` prefill effect from `LocalePickerAction` in the i18n plugin; initial form state is now the single source of truth.
- Add unit tests for i18n inheritance in `useDocument.test.tsx`.
- Stabilize flaky E2E: i18n edit-view regression test and content-releases upcoming-releases home widget test.

## Why is it needed?

After #24659, non-localized fields were prefilled on the first visit to a new locale, but cleared when revisiting that locale without saving—the `LocalePickerAction` effect did not survive form re-inits.

## How to test it?

1. **Manual:** Content Manager → localized type → save EN → switch to FR (unsaved) → verify non-localized fields → switch EN → switch FR again → fields still prefilled.
2. **Unit:** `cd packages/core/content-manager && yarn test:front useDocument`
3. **E2E (EE):** `yarn test:e2e --domains i18n content-releases`

## Related issue(s)/PR(s)

Fixes CMS-540

- Builds on #24659 (prefill non-translatable fields for new locale)

### DIFF
--- a/packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
+++ b/packages/core/content-manager/admin/src/hooks/tests/useDocument.test.tsx
@@ -349,6 +349,267 @@ describe('useDocument', () => {
 });
 
 /**
+ * Regression test: Non-localized fields should copy values from a sibling locale
+ * when creating a new locale draft. This inheritance must persist across form resets.
+ * This is now handled in `getInitialFormValues` to keep `initialValues` as the trusted source.
+ */
+describe('useDocument · getInitialFormValues · i18n non-localized inheritance', () => {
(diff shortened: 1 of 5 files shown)


innerdvations:
Closing as duplicate of #26367, which includes the same core fix plus server-side `availableLocales` ordering and aligned non-localized field semantics. Any unrelated E2E hardening (e.g. content-releases home-widgets) can be split into a separate PR if still needed.

PR #26459, opened 2026-05-27
chore(tsconfig): scope root jsconfig and add per-workspace configs

unrevised6419:
### What does it do?

Adds per-workspace `jsconfig.json` files for JavaScript-only workspaces that previously had no TypeScript configuration:

- `packages/plugins/users-permissions/{admin,server}/jsconfig.json`
- `examples/plugins/workspace-plugin/{admin,server}/jsconfig.json`
- `examples/getstarted/src/plugins/local-plugin/{admin,server}/jsconfig.json`
- `packages/utils/{api-tests,eslint-config-custom,typescript}/jsconfig.json`
- `.github/actions/check-pr-status/jsconfig.json`

Each admin config extends the shared `tsconfig` workspace package's `client.json` (browser/React lib) and each server/Node config extends `base.json`, so editors resolve the correct lib per file. `check-pr-status` uses a small inline config since that GitHub Action is a standalone package outside the `tsconfig` workspace's reach.

Simplifies the root `jsconfig.json` to use blanket excludes for directories that own their config or are out of scope (`packages`, `examples`, `templates`, `.github/actions`, `docs`), now that each relevant workspace owns its config. Also adds a few missing exclude patterns (`.nx`, `**/.docusaurus/*`, `**/coverage/*`).

### Why is it needed?

The root `jsconfig.json` was acting as an implicit fallback for the entire monorepo, sweeping ~4,700 source files into a single configuration. Workspaces with mixed admin (browser) and server (Node) code were being treated uniformly, which gives editors the wrong lib for some files in un-typed JavaScript packages such as `users-permissions`.

After this change:

- Each JS-only workspace gets the correct lib for its runtime (`DOM` for admin, Node for server).
- The root `jsconfig.json` only covers genuinely orphan files (root scripts, root configs, `tests/`).
- `docs/` is excluded from the root scope and intentionally left without its own config — it is a standalone Docusaurus project with its own `node_modules` and tooling (see the discussion comment below).
- Future new workspaces under `packages/` or `examples/` won't accidentally inherit root settings; they must declare their own configuration.

### How to test it?

1. Open any `.js` file in `packages/plugins/users-permissions/admin/src/` — editor should resolve DOM globals (e.g. `window`, `document`, React JSX).
2. Open any `.js` file in `packages/plugins/users-permissions/server/` — editor should resolve Node globals (e.g. `process`, `Buffer`).
3. Run `./node_modules/.bin/tsc -p <any-new-jsconfig> --noEmit` — expect 0 errors.
4. Run `./node_modules/.bin/tsc -p jsconfig.json --noEmit` — same pre-existing errors as before (root scope now excludes `docs/`, so the handful of docs errors no longer appear).

### Related issue(s)/PR(s)

N/A

---
Fixes CPR-404


unrevised6419:
### Follow-up: what about `tests/` and `scripts/`?

After this PR, `tests/` and `scripts/` are the only directories still covered by the root `jsconfig.json`. Worth deciding whether to give them dedicated configs:

- **`tests/`** — ~67 `*.test.api.ts` files run under Jest/Vitest/Playwright (Node). They account for most of the type errors the root config surfaces. A dedicated `tests/tsconfig.json` extending the Node base, with the right test-type globals (`@types/jest`, etc.), would isolate and likely reduce those.
- **`scripts/`** — a small set of Node build/release scripts. Low value, but trivial to add for consistency.

**Recommendation:** yes for `tests/` (clear benefit), optional for `scripts/`. Suggest a follow-up PR to keep this one focused.

### Resolved: `docs/`

`docs/` is now **excluded from the root `jsconfig.json`** and intentionally left without its own config. A dedicated `docs/jsconfig.json` was prototyped but reverted: `docs` is a standalone Docusaurus project with its own `node_modules`, and the only way to get a clean type-check was a `paths` hack working around a broken third-party typing (`@cmfcmf/docusaurus-search-local` ships no declarations and `docusaurus.config.js` imports its raw `/src` for types). Not worth carrying in this PR — excluding `docs` from the repo-wide scope is the cleaner call.

_(An earlier version of this comment cited "~8 `@docusaurus/*` errors"; that count came from an experiment using `moduleResolution: node16`, not the config in this PR.)_

unrevised6419:
Closing in favor of https://github.com/strapi/strapi/pull/26699

PR #26470, opened 2026-05-28
test(cli): reliably stop remote Strapi after transfer e2e

innerdvations:
## Summary

- Add shared remote Strapi lifecycle helpers for CLI transfer e2e tests (`pull-remote`, `push-remote`).
- Start the remote `npm run -s start` process in its own process group on POSIX so teardown signals reach both npm and the Strapi child process.
- Replace broken `afterAll` cleanup that checked `child.killed` after `SIGTERM` — Node sets `killed` when `kill()` is called, not when the process exits, so the fallback `SIGKILL` path was never reliable.
- Tear down the remote process group with graceful `SIGTERM`, wait for exit, then fall back to `SIGKILL` only if the process group does not stop in time. Windows keeps the direct child-process signal fallback.

## Why is it needed?

The Strapi CLI domain shares `test-app-0` between remote transfer suites and later admin CLI tests (for example `admin:block-user`). When the remote Strapi server was not fully stopped, SQLite stayed locked and the later suite's `beforeAll` setup could fail while creating/resetting admin data.

This is a test harness cleanup issue rather than a Strapi runtime shutdown bug: Strapi can shut down gracefully when it receives `SIGTERM`, but the test was starting Strapi through `npm run start`, so signaling only the npm child did not reliably stop the actual Strapi server.

## How to test it?

Run the targeted transfer/admin path:

```bash
yarn test:cli -d strapi --testPathPattern="pull-remote|push-remote|block-user"
```

Or run the full Strapi CLI domain:

```bash
yarn test:cli -d strapi
```

Additional focused verification used locally:

```bash
TEST_APPS="/path/to/test-app-0,/path/to/test-app-1" JWT_SECRET="test-jwt-secret" yarn jest --config /path/to/jest.config.cli.js --rootDir /path/to/tests/cli/tests/strapi --color --verbose --runInBand --testPathPattern="pull-remote|push-remote|block-user"
```

## Related issue(s)/PR(s)

Fixes CMS-1088


tests/utils/cli-transfer-remote-e2e/stop-remote-process.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+const treeKill = require('tree-kill');
```
mathildeleg: Shouldn't this be added in the dependencies/devDependencies in the package.json?

mathildeleg (commented):
Tested and LGTM, just a small suggestion/question :)

tests/utils/cli-transfer-remote-e2e/stop-remote-process.js
innerdvations: You made me think about it a bit harder and I decided to go a different (better) direction that doesn't add a dependency at all. Could you take a look again?

mathildeleg (approved):

PR #26474, opened 2026-05-28
chore(repo): skip change freeze ownership check when freeze disabled

innerdvations:
## Summary

- Skip the token-backed team membership check when `CHANGE_FREEZE_ENABLED` is not `true`
- Add a lightweight pass job so fork PRs get a green check instead of failing on missing `CHECK_OWNERSHIP_TOKEN`
- Remove redundant `CHANGE_FREEZE_ENABLED` guard from the alert job (ownership check already only runs during an active freeze)

## Why is it needed?

The Check Change Freeze workflow fails on fork PRs even when no change freeze is active. GitHub does not expose repository secrets to `pull_request` workflows from forks, so `CHECK_OWNERSHIP_TOKEN` is empty and the membership action errors before it can run. This produces a misleading red check on community PRs.

## How to test it?

1. Open a PR from a fork with `CHANGE_FREEZE_ENABLED=false` (current state) and confirm the workflow reports **Change freeze not active** instead of failing on `GITHUB_TOKEN`
2. With `CHANGE_FREEZE_ENABLED=true`, confirm same-repo PRs still run the team membership check and post the freeze comment for non-team members

## Related issue(s)/PR(s)

Fixes CMS-1089

### DIFF
--- a/.github/workflows/changeFreeze.yml
+++ b/.github/workflows/changeFreeze.yml
@@ -7,8 +7,16 @@ on:
 permissions: read-all
 
 jobs:
+  change-freeze-disabled:
+    name: 'Change freeze not active'
+    if: vars.CHANGE_FREEZE_ENABLED != 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Change freeze is not enabled"
+
   check-ownership:
     name: 'Check if PR was created by a Strapi Team Member'
+    if: vars.CHANGE_FREEZE_ENABLED == 'true'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


derrickmehaffy:
this one has been broken for a while, we can remove it honestly

innerdvations:
Yeah, I thought about removing it, but this way we can keep it around for whenever we want to turn it on, and in the meantime it shouldn't get in the way anymore

Bassel17 (approved):

jhoward1994 (approved):

PR #26475, opened 2026-05-28
chore(examples): use local strapi-plugin binary in todo-example

unrevised6419:
### What does it do?

Updates the `todo-example` plugin's `package.json` to run the local
`strapi-plugin` binary for the `build` and `watch` scripts instead of
`npx @strapi/sdk-plugin@6.0.1`. The `@strapi/sdk-plugin` package (which
provides that binary) is added as a `devDependency`, and `yarn.lock` is
updated accordingly.

### Why is it needed?

Pinning the CLI behind `npx @strapi/sdk-plugin@6.0.1` re-resolves and
re-downloads the package on every `build`/`watch` invocation. Declaring it
as a devDependency lets the version be managed through the workspace
lockfile and uses the already-installed binary directly.

### How to test it?

From `examples/plugins/todo-example`:

- `yarn build` — should build the plugin via the local `strapi-plugin` binary
- `yarn watch` — should start watch mode via the local `strapi-plugin` binary

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/examples/plugins/todo-example/package.json
+++ b/examples/plugins/todo-example/package.json
@@ -19,8 +19,8 @@
     }
   },
   "scripts": {
-    "build": "npx @strapi/sdk-plugin@6.0.1 build",
-    "watch": "npx @strapi/sdk-plugin@6.0.1 watch",
+    "build": "strapi-plugin build",
+    "watch": "strapi-plugin watch",
     "lint": "eslint ."
   },
   "peerDependencies": {
@@ -35,6 +35,7 @@
     "styled-components": "6.4.1"
   },
   "devDependencies": {
+    "@strapi/sdk-plugin": "6.0.1",
     "@strapi/strapi": "workspace:*",
     "eslint": "8.57.1"
   },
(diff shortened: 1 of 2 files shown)


innerdvations:
I removed this to get a pretty heavy dependency out of the monorepo. If it's causing difficulties when working with plugins, do you have an alternative suggetion so we could try to keep the dep out of package.json?

unrevised6419:
Oh, did not know that. Then we can close this.
Another solution would be to exclude that workspace from the package.json.
But in that case one would need to run `yarn install` for that workspace alone to install the deps.

Feel free to close it if that's the case.

PS: Maybe we can instead add a not somewhere, and revert the changes.

innerdvations:
Yeah, I think we should come up with a better workspace management solution so that things like the example projects are independent. I admittedly don't have a great understanding of the best practices there, but I will try to look into it in the coming weeks/months since we're planning a lot of restructuring in the monorepo soon.

unrevised6419:
For example the `docs` folder is excluded from the workspaces. So you have to manually run `yarn install` there. Probably entire `examples` can be excluded from the workspaces 🤔

PR #26476, opened 2026-05-28
fix(database): escape LIKE wildcards in filters and use equality for $eqi/$nei

### LINKED ISSUE #26468: Potential bug: wildcard characters in filters are not escaped, causing incorrect matching for literal text
### Node Version

v22.14.0

### Package Manager

npm

### Package Manager Version

10.9.2

### Strapi Version

5.43.0

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

### Steps to Reproduce

1. Using API REST filters
2. Filter: endsWith = 100%
Image

### Expected Behavior

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

masonmcelvain:
- close #26468

## Summary
- `$eqi`/`$nei` now compile to `= LOWER(?)` / `<> LOWER(?)` instead of `LIKE`, so they perform true case-insensitive equality and can no longer wildcard or crash on a trailing backslash.
- The substring operators escape `%`, `_` and `\` in the user value (special SQLite handling).

## Test plan
- [ ] `yarn workspace @strapi/database test:unit src/query/helpers/__tests__/where.test.ts` (27 cases across postgres/mysql/sqlite)
- [ ] API filtering suite `tests/api/core/strapi/filtering.test.api.js` on sqlite/postgres/mysql
- [ ] Manually confirm `GET /api/articles?filters[slug][$eqi]=foo%5C` returns 200 (not 500) on PostgreSQL

### Failing CLI Tests check

The failing **CLI Tests (node: 24)** check seems unrelated to this PR, it is a pre-existing flakey test on `develop`. Example `develop` failing run:

- https://github.com/strapi/strapi/actions/runs/26583397651/job/78323865184

It passes on other `develop` runs, so it's intermittent on the base branch independent of this change (in `tests/cli/tests/strapi/strapi/block-user.test.cli.js`, the `admin:block-user` suite).

---
Fixes CPR-367


innerdvations (dismissed):
Tested and LGTM, thanks!

innerdvations (approved):
Added some tests

PR #26478, opened 2026-05-28
enhancement(api-tests): golden snapshot restore for builder teardown

innerdvations:
### What does it do?

Speeds up API integration test teardown for suites that use `destroyTestSetup()` by restoring a **golden snapshot** of the shared `test-apps/api` tree instead of running many CTB delete operations (each of which boots and destroys Strapi).

- Captures `src/api`, `src/components`, and a sqlite database baseline once per test run (in `run-api-tests.js`, after bootstrap).
- **`destroyTestSetup()`** calls `builder.cleanup({ useGoldenRestore: true })` by default — used by all **28/28** document-service resource-pack suites.
- **`builder.cleanup()`** keeps **legacy CTB teardown by default** (`useGoldenRestore: false`). Direct `createTestBuilder` callers are unchanged unless they explicitly opt in.
- **Golden restore is sqlite-only** (full `data.db` file copy). Relational row-level restore for postgres/mysql corrupts schema (missing indexes); postgres/mysql shards keep legacy teardown.
- **CI golden artifact caching** (sqlite): prep jobs capture once; shard jobs download the artifact and skip per-shard capture (`GOLDEN_SNAPSHOT_PROVIDED`).
- **Unit tests** for `golden-snapshot.js` / `golden-snapshot-database.js` (`packages/utils/api-tests/__tests__/`).

Opt out in a test file when golden restore is not appropriate:

```ts
await destroyTestSetup(testUtils, { useGoldenRestore: false });
// or
await builder.cleanup({ useGoldenRestore: false });
```

Opt in for direct builder callers:

```ts
await builder.cleanup({ useGoldenRestore: true });
```

### Why is it needed?

`destroyTestSetup()` with the full document-service resource pack issues ~20+ full Strapi lifecycles per suite via legacy CTB teardown, which dominates wall time and can hit Jest `afterAll` timeouts on slow CI runners. Golden restore reduces that to filesystem copy + sqlite file restore.

Default-on golden restore for all `builder.cleanup()` callers was tried first but broke cross-suite state in mixed jest processes (stale CTB artifacts under `src/api` after partial filesystem restore). Scoping to `destroyTestSetup()` keeps legacy behaviour for the ~150 direct builder callers while accelerating the document-service path.

#### Local benchmarks (SQLite, macOS)

Recorded in `tests/scripts/golden-restore-benchmark-results.md`. Golden snapshot pre-captured; timings exclude capture cost.

**7 migrated relation/dp suites** (switched from `builder.cleanup()` → `destroyTestSetup()`):

| | BEFORE (legacy) | AFTER (golden) |
|---|---|---|
| Mean (3 runs) | **19.2 s** | **11.8 s** |
| Delta | | **~39% faster (~7.4 s)** |

**Full `document-service/` folder** (28 suites, 617 tests):

| | BEFORE (21 golden + 7 legacy) | AFTER (all golden) |
|---|---|---|
| Mean (3 runs) | **85.7 s** | **81.4 s** |
| Delta | | **~5% faster (~4.3 s)** |

Reproduce: `./tests/scripts/benchmark-golden-restore.sh <label> document-service 3`

#### CI benchmarks (CE sqlite, node 24)

Golden artifact caching vs baseline (`0ce5d82b7a`):

| Shard | Baseline | This PR | Δ |
|---|---|---|---|
| 1/4 | 8m35s | **5m14s** | **−3m21s** |
| 2/4 | 7m45s | 7m28s | −17s |
| 3/4 | 8m05s | 8m08s | +3s |
| 4/4 | 8m01s | 8m01s | 0 |

Shard 1/4 benefits most from skipping per-shard capture; other shards are within CI noise. Postgres/mysql matrix jobs are unchanged from `develop` (sqlite via default positional `databaseName`).

### How to test it?

```bash
yarn rimraf test-apps/api test-apps/.golden
yarn test:api --testPathPattern=golden-snapshot-restore --generate-app

# Smoke a heavy suite (e.g. middlewares with full resources)
yarn test:api --testPathPattern=middlewares.test.api.ts --generate-app

# Full document-service slice (where golden restore applies)
yarn test:api -- tests/api/core/strapi/document-service/

# Unit tests
yarn nx run api-tests:test:unit

# Benchmark (optional)
chmod +x tests/scripts/benchmark-golden-restore.sh
./tests/scripts/benchmark-golden-restore.sh my-label document-service 3
```

### Expected follow-ups

Not in scope for this PR; tracked as natural next steps:

1. **Relational golden restore** — Replace row-level postgres/mysql snapshot with native dump/restore (`pg_dump` / `mysqldump`). Row-level restore corrupts indexes and was the reason golden restore is sqlite-only today.
2. **Real postgres/mysql CI matrix** — Once (1) exists, honor `--dbclient` in `generate-test-app.js` and re-enable postgres/mysql golden prep jobs + artifacts (removed in this PR).
3. **Migrate more builder suites** — ~150 files still call `builder.cleanup()` directly. Highest ROI next: `content-manager/api/*` and graphql/i18n CTB suites (same pattern as the 7 document-service migrations; benchmark per folder before widening).
4. **Filesystem-restore hardening** — Before migrating suites outside document-service, validate they only mutate `src/api` / `src/components` (or extend snapshot dirs / purge list for other generated paths).
5. **Further CI shard savings** — Diminishing returns beyond sqlite capture caching; any large wall-clock drop needs more suites on golden teardown and/or less work in `test:generate-app` per shard.

### Related issue(s)/PR(s)

N/A — infrastructure improvement from API test performance investigation.

Fixes CMS-1377


nclsndr:
I like the vision of this proposal, yet I think it is way too premature considering our testing setup issues are way beyond having a capture stage.
IMHO, we want the setup to be able to seed various cases before we try to snapshot one, we might want many.
Side note: I think we should stop authoring scripts in JS, TS everywhere, `tsx` until we can `node` directly (>22.18)

innerdvations:
I'll close this for now but might want to revisit it

PR #26483, opened 2026-05-29
fix(build): build does not run install; add install-deps arg

### LINKED ISSUE #26344: strapi build auto-installs admin deps via 'npm install --legacy-peer-deps --save' which removes esbuild and breaks the build

innerdvations:
## Summary

- Refactors admin peer dependency handling into separate check, install, and orchestration layers
- `strapi build` fails fast by default when admin peer deps are undeclared (no mid-build `npm install --legacy-peer-deps --save`)
- `strapi develop` keeps auto-install by default; both commands support `--install-deps` / `--no-install-deps`
- Adds unit tests for dependency checks, orchestration, and CLI flag wiring

Supersedes #26362 with a modular implementation and test coverage.

## Why is it needed?

Fixes #26344 — when `strapi build` auto-installs missing admin deps, the secondary install can re-resolve the tree and remove `esbuild`, causing a confusing build failure. Build should not mutate `node_modules` or lockfiles.

## How to test it?

1. Create a Strapi app whose `package.json` omits `react`, `react-dom`, `react-router-dom`, and/or `styled-components`
2. Run `strapi build` — should fail immediately with a package-manager-aware install hint (no install runs)
3. Run `strapi build --install-deps` — should auto-install and re-run (legacy opt-in)
4. Run `strapi develop` — should auto-install missing peers as before
5. Run `strapi develop --no-install-deps` — should fail fast like build
6. Run `yarn test:unit --testPathPattern=dependencies.test` in `packages/core/strapi`

## Related issue(s)/PR(s)

Fixes #26344
Fixes CMS-1090
Supersedes #26362


packages/core/strapi/src/cli/commands/develop.ts
```diff
@@ -37,6 +37,8 @@ const command: StrapiCommand = ({ ctx }) => {
     .option('--build-admin', 'Build the admin panel', true)
     .option('--no-build-admin', 'Do not build the admin panel in case watch is disabled')
     .option('--open', 'Open the admin in your browser', true)
+    .option('--install-deps', 'Auto-install missing admin dependencies', true)
+    .option('--no-install-deps', 'Do not auto-install missing admin dependencies')
```
jhoward1994, line 41: My initial impression is that this seems redundant? if L41s options is just the opposite of L40 ? e.g. couldn't we just set `--install-deps` to false

packages/core/strapi/src/node/core/dependencies.ts
```diff
@@ -5,199 +5,183 @@ import semver, { SemVer } from 'semver';
...
+    return `pnpm add --save-prod ${packages}`;
+  }
 
-    /**
-     * Re-run the same command after installation e.g. strapi build because the yarn.lock might
-     * not be the same and could break installations. It's not the best solution, but it works.
-     */
-    await execa(file, args, { cwd, stdio: 'inherit' });
-    return { didInstall: true };
+  return `npm install --save ${packages}`;
```
jhoward1994:
```suggestion
  return `npm install --legacy-peer-deps --save ${packages}`;
```

innerdvations: Nice catch, thanks

packages/core/strapi/src/cli/commands/develop.ts
innerdvations: I basically kept it for symmetry between the two commands and because `--x/--no-X` is a pattern we use in other place in the CLI (and that is built into Commander). I could drop the `--install-deps` if you think it's confusing to have both though

jhoward1994 (dismissed):
Looks good other than the conflict we have atm

Bassel17 (approved):

PR #26486, opened 2026-05-29
feat: user stories derived from e2e tests

cache-your-dreams:
### What does it do?
Generating User Stories from the e2e tests

### Why is it needed?
So that Playwright MCP can then read them and generate the e2e tests automatically


.agents/skills/userstory-e2e/SKILL.md
```diff
@@ -0,0 +1,59 @@
+---
+name: userstory-e2e
+description: Generate or update Playwright e2e tests for the billing portal so they codify the acceptance criteria in docs/UserStories/*.md. Use when a UserStory file is added or updated and you need matching e2e coverage in front/e2e/. Uses the Playwright MCP to drive the running app and discover selectors before producing @playwright/test specs.
```
innerdvations, line 3: Not just for billing portal, right?

Also, I'm not totally sure if we should codify Playwright MCP here, since different agents may have access to other browser tools that are more effective. If we're worried it won't pick up on needing to use a tool, maybe offer "Playwright (or your preferred alternative)"

Issue #26492, opened 2026-05-30
[CI] Nightly release workflow publishes to npm without running any tests


m-zandi:
### Node Version

20

### Package Manager

yarn

### Package Manager Version

1.22.22

### Strapi Version

develop (nightly next channel)

### Operating System

Other

### Database

Other

### Javascript or Typescript

Other

### Reproduction URL

https://github.com/strapi/strapi/blob/develop/.github/workflows/nightly.yml

### Bug Description

The nightly release workflow (nightly.yml) publishes packages 
directly to npm under the next dist-tag every night without 
running any test suite first.

The full tests.yml workflow runs build, lint, TypeScript checks, 
unit tests, e2e tests, and API integration tests across 3 Node 
versions and 3 databases before anything ships. The nightly 
workflow skips all of this entirely — it only runs yarn install 
and immediately calls pre-publish.sh.

This means if a breaking change is merged into develop at any 
point during the day, it will be silently published to npm that 
night. Developers and integrators consuming @strapi/strapi@next 
receive broken packages with no warning or gate.

### Steps to Reproduce

1. Review .github/workflows/nightly.yml — observe it has no 
   needs dependency on any test job
2. Merge any breaking change into the develop branch
3. Wait for the nightly cron trigger (0 0 * * 2-6)
4. The broken code is published to npm@next with zero test gate

### Expected Behavior

The nightly workflow should gate the publish job behind at 
minimum a build and unit test smoke check. The publish job 
should declare needs: [test] so that npm publishing is 
automatically blocked if any test fails on the develop branch 
at publish time.

### Logs

```shell

```

### Code Snippets

# .github/workflows/nightly.yml — current state (no test gate)
jobs:
  publish:
    steps:
      - uses: actions/checkout@v6
      - run: yarn
      - run: ./scripts/pre-publish.sh --yes  # publishes with zero test gate
        env:
          VERSION: '0.0.0-next.${{ github.sha }}'
          DIST_TAG: next

# tests.yml runs ALL of these before any release — nightly skips them all:
# build → lint → typescript → unit_back → unit_front → e2e_ce → e2e_ee
# → api_ce_pg → api_ce_mysql → api_ce_sqlite → api_ee_pg → api_ee_mysql

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
[...]

PR #26512, opened 2026-06-01
ci: gate nightly publish behind validation

mturac:
## What does it do?

Adds a `validate` job to the nightly release workflow and makes the `publish` job depend on it. The validation job runs on Node 24, installs the monorepo, builds the monorepo, and runs the Vitest unit smoke suite before nightly packages can be published.

## Why is it needed?

The scheduled nightly workflow currently installs dependencies and publishes `next` packages without any build or test gate. If `develop` is broken when the cron runs, the broken packages can be published automatically.

## How to test it?

```bash
ruby -e "require 'yaml'; YAML.load_file('.github/workflows/nightly.yml'); puts 'yaml ok'"\nGOBIN=/tmp/strapi-bin go install github.com/rhysd/actionlint/cmd/actionlint@latest\n/tmp/strapi-bin/actionlint .github/workflows/nightly.yml\ngit diff --check\n```\n\nI also tried `yarn prettier .github/workflows/nightly.yml --check`, but this fresh clone does not have `node_modules` installed, so Yarn stops before running the script.\n\n## Related issue(s)/PR(s)\n\nFixes #26492

---
Fixes CPR-378

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -8,8 +8,25 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 
 jobs:
+  validate:
+    name: 'Validate'
+    runs-on: ubuntu-latest
+    if: github.repository == 'strapi/strapi'
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+      - name: Monorepo install
+        uses: ./.github/actions/yarn-nm-install
+      - name: Monorepo build
+        uses: ./.github/actions/run-build
+      - name: Run Vitest unit smoke
+        run: yarn test:unit:vitest
+
   publish:
     name: 'Publish'
+    needs: validate
     runs-on: ubuntu-latest
     if: github.repository == 'strapi/strapi'
     steps:


mturac:
Added an extra read-only review pass over the workflow diff. No actionable CI gating or security findings were found.\n\nRe-checked the referenced validation surfaces locally on Node v24.14.1:\n\n```\ntest -f .github/actions/yarn-nm-install/action.yml\ntest -f .github/actions/run-build/action.yml\nnode -e "const s=require('./package.json').scripts; console.log(s['test:unit:vitest'])"\ngit diff --check origin/develop -- .github/workflows/nightly.yml\n```\n\nResult: referenced actions exist, root script resolves to `vitest run`, and the workflow diff has no whitespace errors.

nclsndr:
Thank you for opening this discussion @mturac.
My opinion is that we should trust the develop to be stable at anytime. Did you ever have an issue with it?

mturac:
Thanks for the quick reply. I haven’t hit a concrete broken `develop` case personally; I opened this mostly as a guardrail/clarification question after seeing how contributors may rely on branch stability when choosing a base.

If the expectation is that `develop` should remain stable at all times, that answers my concern. Happy to close this discussion unless you think there is documentation worth clarifying for contributors.

nclsndr:
The overall rationale is that any PR targeting develop needs to pass CI checks, bringing the guarantee that anyone can base their work on a stable develop, anytime.

I will not say we never break develop anytime (it happens that the CI can be flaky and passes with false positive). When it happens we make our best efforts to fix within the next hours.

> guardrail/clarification question after seeing how contributors may rely on branch stability when choosing a base

Did you ever notice people hesitating or not using develop?

mturac:
No, I do not have a concrete example of contributors avoiding `develop` or hesitating because of its stability.

This was a theoretical guardrail concern from reading the workflow, not a production issue I personally hit. Your explanation clarifies the project expectation: PRs targeting `develop` should pass CI, and breakages are handled quickly when false positives slip through.

Given that, I do not think this needs to stay open as a code change. Closing it from my side rather than keeping a speculative PR in the queue. Thanks for the clarification.

PR #26514, opened 2026-06-01
fix(admin): preserve fetch error bodies for text responses

### LINKED ISSUE #26199: useFetchClient does not handle server errors well in case new responseType option is involved

mturac:
## What does it do?

Preserves structured JSON API errors when `getFetchClient().get()` is called with a non-JSON `responseType`, such as `text`, and the server returns an error response. Non-JSON success responses still use the requested parser, and non-JSON/non-parseable error bodies still fall back to the existing generic `FetchError`.

## Why is it needed?

When callers request text/blob/arrayBuffer data, the non-2xx path currently throws `FetchError: Server Error` without reading the server error body. That loses useful API messages like `BadRequestError: Type is required`, making it difficult for callers to show actionable feedback.

## How to test it?

```bash
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn install --immutable
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn workspace @strapi/admin-test-utils build
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn jest packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts --config jest.config.front.js --runInBand -t "preserve JSON API errors"
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn jest packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts --config jest.config.front.js --runInBand
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn prettier packages/core/admin/admin/src/utils/getFetchClient.ts packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts --check
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn eslint packages/core/admin/admin/src/utils/getFetchClient.ts packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts
git diff --check
```

I also ran:

```bash
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn nx run-many --target=build --projects=@strapi/types,@strapi/utils,@strapi/admin-test-utils --nx-ignore-cycles
PATH="$HOME/.nvm/versions/node/v24.14.1/bin:$PATH" yarn workspace @strapi/admin test:ts:front
```

The targeted dependency build passed. `test:ts:front` still fails on existing admin self-export/type issues such as `Cannot find module @strapi/admin/strapi-admin/ee` and widget prop test errors, not on this fetch-client change. The full `@strapi/admin` lint target has the same existing unresolved self-import errors; targeted eslint for the changed files passes.

## Related issue(s)/PR(s)

Fixes #26199

---
Fixes CPR-380

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -301,6 +301,20 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
   // Check if the url has a prepending slash, if not add a slash
   const normalizeUrl = (url: string) => (hasProtocol(url) ? url : addPrependingSlash(url));
 
+  const parseErrorResponse = async (response: Response): Promise<ErrorResponse | undefined> => {
+    try {
+      const result = await response.json();
+
+      if (result?.error) {
+        return { data: result };
+      }
+    } catch {
+      return undefined;
+    }
+
+    return undefined;
+  };
+
   // Add a response interceptor to return the response
   const responseInterceptor = async <TData = any>(
     response: Response,
@@ -309,7 +323,11 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
   ): Promise<FetchResponse<TData>> => {
     if (responseType !== 'json') {
       if (!response.ok && !validateStatus?.(response.status)) {
-        const fetchError = new FetchError('Server Error');
+        const errorResponse = await parseErrorResponse(response);
+        const fetchError = new FetchError(
+          errorResponse?.data.error.message ?? 'Server Error',
+          errorResponse
+        );
         fetchError.status = response.status;
         throw fetchError;
       }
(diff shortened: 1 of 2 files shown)


mturac:
Added an extra read-only review pass over the current branch diff. No actionable correctness, regression, or test-coverage findings were found in the two changed files.\n\nRe-ran the targeted admin frontend test on Node v24.14.1:\n\n```\nyarn workspace @strapi/admin test:front getFetchClient.test.ts --runInBand\n```\n\nResult: 25 tests passed.

Issue #23576, opened 2025-05-22
Incorrect media location displayed before upload in Media Library


raulbalestra:
### Node Version

22.15.1

### NPM/Yarn/PNPM Version

Yarn

### Strapi Version

5.12.6

### Operating System

MacOS

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Bug Description

When adding a new video file to a specific folder in the Media Library, the location displayed during the pre-upload editing phase incorrectly shows as "Media Library" instead of the selected folder (e.g., test1). This leads to confusion, especially for users managing assets in nested folder structures.

### Steps to Reproduce

Steps to reproduce:

Login with a valid user account.

Navigate to Media Library.

Open an existing folder (e.g., Shorts_Folder_3).

Click "Add more assets" and select a video to upload.

Before clicking Upload, click the Edit button on the selected video file.

Check the Location field in the edit modal.

Actual result:

The Location field incorrectly shows: Media Library.

### Expected Behavior

The Location should reflect the folder currently opened (e.g., test1), where the user initiated the upload action.

### Logs

```shell

```

### Media

https://github.com/user-attachments/assets/0cc31171-d521-4237-9736-907a016fbb13

### Additional information

video example 2: https://www.loom.com/share/e0db6a0fb62840e68160cbe940d69dd7?sid=d78a0d8f-9346-4de2-ae53-d28bb10ab2d7

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

raulbalestra:
TID: 1403

PapatMayuri:
@raulbalestra  here this location field is for user as where should we want to save image. like if i have 
Folder1-->Folder2 & Folder2Sub1
Folder2--> Folder3
Folder3--> Folder4

Now i am adding asset in folder 4 and giving location as Folder2Sub1 then that asset is available in both Folders.
I tried this correct me if i am wrong

Adzouz:
Hi @raulbalestra, thanks for reporting the issue 🙏 This was fixed in the new upcoming media library (release on Wednesday) so I'll close the issue! Have a great day ✨

Adzouz:
Hi @raulbalestra, thanks a lot for reporting the issue!

[...]

Issue #21904, opened 2024-10-17
Invalid images in the specified media library path


Fronikuniu:
### Required System information

Environment: development
OS: win32-x64
Strapi Version: 4.25.11
Node/Yarn Version: npm/10.8.3 node/v20.13.1 win32 x64 workspaces/false
Edition: Community
Database: postgres
Typescript

### Steps to reproduce the behavior

When I add files via content manager e.g. in some collection/single type

![image](https://github.com/user-attachments/assets/f99be824-3446-44ef-b781-5c98728bfa7c)

I enter the test/path path and select the image

![image](https://github.com/user-attachments/assets/76ffaff1-3db6-44f2-a529-a49dabf01eb0)

Then when I want to add more images and/or change the image

![image](https://github.com/user-attachments/assets/48a529d0-ae0c-4c10-ae41-c0a20a36b95f)

I select browse to select a different image/upload a new one and I get the images from the main media library directory and the path is to the folder I am in, where the images should be displayed.

![image](https://github.com/user-attachments/assets/41cafca5-455f-4500-9855-36a1ae6b2ff6)

### Expected behavior

I should see media from current folder, not from main media library

Fronikuniu:
Any update from strapi team? It also appears in strapi v5 @derrickmehaffy

Fronikuniu:
Warning: Failed prop type: The prop `assets[0].folder.isRequired` is marked as required in `cs`, but its value is `undefined`.

Warning: Failed prop type: The prop `asset.folder.isRequired` is marked as required in `Ks`, but its value is `undefined`.

Fronikuniu:
the fix is:
setFolderId(null);
inside: https://github.com/strapi/strapi/blob/ce8ed0c8aa03361c3009f41af96f742be028ffd9/packages/core/upload/admin/src/components/MediaLibraryInput/MediaLibraryInput.tsx#L58

dagadevelop:
What is the intention? To show the root or to show the last location where you were?

Fronikuniu:
> What is the intention? To show the root or to show the last location where you were?

It would be great if it returned to the folder of the selected file, e.g. if we selected an image from “test-folder/test-image.png,” then clicking on “+” -> browse would show the contents of the test-folder. And the solution I gave above corrects the incorrect breadcrumb.

PR #26515, opened 2026-06-01
fix(upload): folder navigation bugs in Media Library

### LINKED ISSUE #23571: Assets not showing in subfolder view of Media Library unless "Add new assets" is triggered

jhoward1994:
## What does it do?

- **useAssets**: Lists files using `filters[folder][id]` instead of `folderPath` (aligns with `useFolders` and future admin behavior).
- **Upload dialog**: Passes the current `folderId` as `initialFolderId` so "Location" defaults to the open folder.
- **updateFileInfo**: Retains `dbFile.folderPath` on metadata-only updates (previously attempted to access non-existent `dbFile.path`).
- **Tests**: Adds unit and API coverage for all of the above, including super-admin access to deep folders.

## Why is it needed?

- #23571— The use of `folder`/`folderPath` in URLs could become outdated, leading to empty subfolders until a refetch.
- #23576 — The upload "Location" was defaulting to root rather than the currently open folder.
- #21904 — Editing metadata was corrupting the `folderPath` in the database.
- **Deep folders** — Filtering by `folderPath` with formats like `/1/2/3` is often blocked by proxies or WAFs; using an ID-based filter avoids these issues.

## How to test it?

1. Nest folders 4 levels deep and upload into the deepest folder — the assets remain visible; network requests use `folder.id`, not `folderPath`.
2. Open a subfolder (or the Content Manager media picker) — assets display immediately without needing an upload.
3. Add assets from within a subfolder — "Location" is pre-filled.
4. Edit metadata only — the file remains in the original folder.
5. Run relevant tests:
   - `yarn test:unit packages/core/upload/server/src/services/__tests__/upload/updateFileInfo.test.ts`
   - `yarn test:front packages/core/upload/admin/src/hooks/tests/useAssets.test.ts`
   - `yarn test:api tests/api/core/upload/admin/folder-nested-permissions.test.api.js`

## Related issue(s)/PR(s)

Fixes #23571, #23576, #21904


packages/core/upload/admin/src/hooks/tests/useAssets.test.ts
```diff
@@ -43,7 +45,28 @@ describe('useAssets', () => {
...
+            },
+          },
+        ],
+      },
+    };
+
+    await waitFor(() => expect(get).toBeCalledWith(`/upload/files`, { params: expected }));
+  });
+
+  test('ignores legacy folderPath query and falls back to $null folder', async () => {
```
Adzouz, line 69: Very tiny concern here about retro compatibility / breaking change. Even if we changed everything on our side, is there a world where this could be implemented this way somewhere and be impacted? (e.g. community plugin) or maybe add a deprecation warning in presence of folderPath just in case?

packages/core/upload/server/src/services/__tests__/upload/updateFileInfo.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      folder: 42,
+      folderPath: '/42',
+    });
+    expect(getFolderPath).toHaveBeenCalledWith(42);
+  });
+
+  test('moves the file to the root when folder is explicitly null', async () => {
+    // FileInfo.folder is typed as number | undefined, but the runtime accepts
+    // null as "no parent" (root). Cast to exercise that branch.
+    await uploadService.updateFileInfo(1, { folder: null as unknown as number });
```
Adzouz: Why not adding null to the `FileInfo.folder` type?

Adzouz (commented):
Hi @jhoward1994, thanks for the PR. I tested and reviewed and all seems working as expected, good job 👏 Only got two tiny questions that totally can be challenged, let me know 😊

packages/core/upload/admin/src/hooks/tests/useAssets.test.ts
jhoward1994: I dont think so because the hook isn't exported so any plugin using useAssets
cant expect it to be stable

Adzouz (dismissed):
Looks good to me, thanks a lot 🙏

Adzouz (approved):

PR #26517, opened 2026-06-01
chore(mcp): clarify registration lifecycle and simplify error messages

nclsndr:
### What does it do?

- Simplifies error messages thrown by `registerTool`, `registerPrompt`, and `registerResource` when called after the MCP server has started — removes the prescriptive "Register during plugin register()" hint from the thrown error.
- Expands JSDoc comments on all three `McpService` registration methods to clarify that registration is valid in both `register()` and `bootstrap()` phases, and recommends `bootstrap()` when the registration depends on synced content-types, permissions, or database state.
- Updates the integration test to match the new error message.

### Why is it needed?

The previous error messages and docs were too narrow: they told plugin authors to register MCP tools only during `register()`, but `bootstrap()` is equally valid (and preferred when the tool depends on initialized database or permission state). This led to confusion for plugin authors choosing the right lifecycle hook.

### How to test it?

1. Write a plugin that calls `strapi.ai.mcp.registerTool(...)` after `strapi.ai.mcp.start()` has run.
2. Confirm the error message reads `[MCP] Cannot register tools after the MCP server has started.`
3. Run the integration test suite: `yarn test packages/core/core/src/services/mcp/__tests__/service-integration.test.ts`

### Related issue(s)/PR(s)

- ø

### DIFF
--- a/packages/core/core/src/services/mcp/__tests__/service-integration.test.ts
+++ b/packages/core/core/src/services/mcp/__tests__/service-integration.test.ts
@@ -189,7 +189,7 @@ describe('MCP Service Integration', () => {
           auth: { policies: [{ action: 'test.action' }] },
           createHandler: jest.fn(),
         });
-      }).toThrow('[MCP] Tools must be registered before MCP server starts');
+      }).toThrow('[MCP] Cannot register tools after the MCP server has started.');
     });
   });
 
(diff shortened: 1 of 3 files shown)


derrickmehaffy (approved):
lgtm

PR #26522, opened 2026-06-02
fix(content-manager): use ReadonlyArray for layout prop and fix Repeatable test fixture

nclsndr:
### What does it do?

- Changes the `layout` prop type in `ComponentInputProps` from `EditFieldLayout[][]` to `ReadonlyArray<ReadonlyArray<EditFieldLayout>>` to accept readonly arrays.
- Updates the `Repeatable.test.tsx` fixture to use `satisfies RepeatableComponentProps` instead of `as const`, and adds the missing `attribute` field to match the full `EditFieldLayout` shape.

### Why is it needed?

The previous `EditFieldLayout[][]` type rejected readonly array sources, causing type errors when passing immutable data. The test fixture was also incomplete — missing the `attribute` field — and relied on `as const` instead of a proper type assertion, hiding shape mismatches.

### How to test it?

Run the content-manager unit tests:
```
yarn workspace @strapi/content-manager test -- --testPathPattern="Repeatable"
```
Verify no TypeScript errors in `Input.tsx` and `Repeatable.test.tsx`.

### Related issue(s)/PR(s)

- #26421

### DIFF
--- a/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
+++ b/packages/core/content-manager/admin/src/pages/EditView/components/FormInputs/Component/Input.tsx
@@ -25,7 +25,7 @@ interface ComponentInputProps
    * We need layout to come from the props, and not via a hook, because Content History needs
    * a way to modify the normal component layout to add hidden fields.
    */
-  layout: EditFieldLayout[][];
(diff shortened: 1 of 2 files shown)


markkaylor (approved):
Passes locally for me 👏

Issue #26524, opened 2026-06-02
firstPublishedAt: entry falsely marked as modified after first publish when feature enabled


deferral-opium:
### Node Version

22.2.2

### Package Manager

npm

### Package Manager Version

10.9.7

### Strapi Version

5.47.0

### Operating System

Linux (Other)

### Database

SQLite

### Javascript or Typescript

Typescript

### Bug Description

When the `experimental_firstPublishedAt` feature is enabled, publishing for the first time causes the entry to appear "modified". 

### Steps to Reproduce

1. Enable `experimental_firstPublishedAt` in features config
2. Create a content type with draft-and-publish enabled
3. Create a new entry and directly publish it
4. Observe that the entry is flagged as "modified" even though no changes were made after publishing

### Expected Behavior

An entry published for the first time should not be flagged as "modified".

### Logs

```shell

```

### Confirmation Checklist

- [x] I have checked the existing [issues](https://github.com/strapi/strapi/issues) for duplicates.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/strapi/strapi/blob/develop/CODE_OF_CONDUCT.md).

PR #26525, opened 2026-06-02
fix(core): propagate server updatedAt in addFirstPublishedAtToDraft to avoid false modified flag

deferral-opium:
### What does it do?

- Captures the server-side `updatedAt` from the `entries.update` response in `addFirstPublishedAtToDraft` and propagates it to the returned draft object
- Previously the function did not capture the `update` result and returned `{ ...draft, firstPublishedAt: now }`, which dropped the draft's `updatedAt` field entirely
- Adds a test verifying that `updatedAt` is propagated from the server response and is not earlier than `firstPublishedAt`

### Why is it needed?

Without this fix, entries published for the first time with the `experimental_firstPublishedAt` feature appear "modified" (`updatedAt < firstPublishedAt`) because `updatedAt` is not passed along from the draft object.

### How to test it?

1. Run the first-published-at unit tests:
```
yarn test:unit --testPathPattern="first-published-at"
```
2. Enable `experimental_firstPublishedAt`, create and publish a new entry — it should not be tagged as "modified"

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/26524

---
Fixes CPR-381


oiorain:
Hey @deferral-opium ,

deferral-opium:
Signed!

packages/core/core/src/services/document-service/__tests__/first-published-at.test.ts
```diff
@@ -79,4 +79,18 @@ describe('addFirstPublishedAtToDraft', () => {
...
+    const draft = { id: 1, firstPublishedAt: null, title: 'test' };
+    update.mockResolvedValue({ id: 1, updatedAt: serverUpdatedAt });
+
+    const result = await addFirstPublishedAtToDraft(draft, update, contentType);
+
+    expect(result).not.toBe(draft);
+    expect(result.firstPublishedAt).toBeInstanceOf(Date);
+    // updatedAt is propagated from the server response so it's equal to or greater
+    // than firstPublishedAt — entry is published, not modified after publishing
+    expect(result.updatedAt.getTime()).toBeGreaterThanOrEqual(result.firstPublishedAt.getTime());
```
Adzouz, line 93: **Timing-dependent assertion — can flake on a millisecond tick.**

`serverUpdatedAt` is created *before* the function runs, but `firstPublishedAt` (`new Date()` inside `addFirstPublishedAtToDraft`) is created *after* it. So this asserts `t0 >= t1` where t1 > t0 whenever the clock ticks between test setup and the call — an intermittent CI failure with no code change.

The production invariant (server `updatedAt` ≥ `firstPublishedAt`) can't be proven by this unit anyway since the mock controls the timestamp. What the unit really verifies is propagation:

```suggestion
    expect(result.updatedAt).toBe(serverUpdatedAt);
```

(Alternative: keep the ordering assertion but build the mock date in the future — `new Date(Date.now() + 60_000)` — to make the intent explicit.)

_Comment generated by AI, reviewed by @Adzouz._

Adzouz (commented):
Hi @deferral-opium, thanks for the contribution 🙏 Just a small comment to address that could lead to flaky test in our pipelines, otherwise it's good for me 😊
Can you also update your branch with develop? 🙏

packages/core/core/src/services/document-service/__tests__/first-published-at.test.ts
deferral-opium: Implemented the alternative suggestion, makes the test more explicit and readable imo

Adzouz (approved):
Looks good to me, thanks a lot!

PR #26529, opened 2026-06-02
future(upload): fix per-file upload progress

markkaylor:
### What does it do

- Replaces /unstable/stream with /unstable/upload-file
- Remove SSE for file uploads
- Adds per file XHR upload with progress tracking

### Why is it needed?

The previous implementation used a single multipart SSE upload of all files, which only emitted "file started" / "file finished" events — there was no per-file byte progress (the client used fetch(), which exposes none). This meant all the files were upload at once as soon as they hit the server and was not reporting actual progress of the uploads. 

### How to test it

Enable the future flag for the media library in features.js
Go to the new media library, it's the ! icon in the sidebar
In devtools set the throttling to something like slow 3G 
Click New -> File upload
Select several files and upload

You should see the dialog with per file upload progress as well as aggregated upload progress in the header. 
Cancel, and it should abort in flight requests
Retry, it should pick up where it left off

### Note

We are back to a POST request _per file_ 
After discussion with my buddy cladue this is apparently the way to do it but would love to hear from someone otherwise. Don't love that. 

I've only implemented this for file uploads. Not sure what to do about URL uploads yet but I thought we could first discuss this to be sure it's the direction we want to go.


giu1io (dismissed):
Looks pretty good. Much cleaner than before. I tested it and it works well for me. 

Only thing that could be improved in the future might be if we want to make configurable how many files to upload concurrently. But this is a great starting point for that.

markkaylor:
>Only thing that could be improved in the future might be if we want to make configurable how many files to upload concurrently. But this is a great starting point for that.

Totally agree, I'll create an issue to handle that in a follow up PR. I'm also going to handle the same refactor for upload from url in a separate PR.

giu1io (approved):

PR #26530, opened 2026-06-02
feat: keep save and publish panels visible while scrolling

lucielandsmanova:
### What does it do?

Makes two action panels sticky so they remain visible while scrolling:

- Content-Type Builder: the Save button + search bar in the left sidebar now stays fixed when editing a long component with many fields
- Edit View: the Save/Publish aside panel on the right now sticks to the top of the viewport when scrolling through a long entry form

### Why is it needed?

When working with large content types or components with many fields, the save and publish actions scroll out of view and require the user to scroll back to the top to perform basic actions. This is a small UX friction that adds up significantly in projects with complex data models.

This behaviour was present in Strapi v4 and sadly appears to have been lost in the v5 redesign.

### How to test it?

- Open the Content-Type Builder and create or edit a component with many fields — the Save button in the left sidebar should remain visible as you scroll down
- Open the Edit View for a collection type with many fields — the Save/Publish panel on the right should stay visible as you scroll through the form

### Screenshots
Save/Publish action panel:
Screenshot 2026-06-02 at 17 20 40

Save button + search bar in the left sidebar:
Screenshot 2026-06-02 at 17 21 50

### Related issue(s)/PR(s)

I am not aware of any related issues.

---
Fixes CPR-383


oiorain:
Hey @lucielandsmanova ,

lucielandsmanova:
Hii @oiorain ! Thank you, all done :)

mathildeleg:
Thanks for your contribution!

Unfortunately we can't merge this as-is: the problem is that the sticky container can be taller than the viewport.
`<Panels />` doesn't just render the Save/Publish panel, it renders all registered edit view side panels stacked vertically. In a standard EE setup, that means: Entry (Save/Publish/Delete), Live Preview, Review Workflows, and Releases.

With `position: sticky; alignSelf: flex-start`, CSS pins the top edge of the container at `top: 16px`. Once the combined height of those panels exceeds the viewport height, the bottom panels are permanently clipped: the sidebar doesn't scroll independently, so Review Workflows and Releases become unreachable the moment the user scrolls down. That's a regression for EE users.

Getting this right requires thinking through the interaction between the sticky behaviour and the variable number of panels that can appear in that slot — which is a product and design decision, not just a CSS fix. 

I'll close this PR for now but the issue's being raised with our Product Designer so that hopefully we can find a solution for this UX pain point :)

PR #19186, opened 2024-01-09
Solve foreign key error on DEITS 

christiancp100:
### What does it do?

It handles the foreign key constraint error showing a warning in the server logs instead of throwing an error and stopping the transfer process.

### Why is it needed?

Users complain that their processes stop because of foreign key constraint errors and we've detected that even though it's potentially due to inconsistencies in their DBs, it's not worth throwing and error and it will be enough to handle these situations with a warning.

### How to test it?

Do a transfer with orphans in your database (create an entity with a relation to another content type and in the database delete one of the ends of the relation, leaving the row linking both content types in the DB). Then execute a transfer to another Strapi project and observe that there is a warning in the server about the foreign key constraint, but the transfer continues as expected. 

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/18350

PR #26531, opened 2026-06-02
fix(data-transfer): skip links referencing data that was never transferred

derrickmehaffy:
### What does it do?

Fixes a chain of issues in DTS that caused transfers to hard-fail at the links stage on PostgreSQL with:

```
warn: (push-handler) Skipping link sub.related-product:1 -> api::data-product-connector.data-product-connector:100 due to a foreign key constraint
error: insert into ..._lnk ... - current transaction is aborted, commands ignored until end of transaction block
Transfer process failed.
```

Three changes:

1. **Record ID mappings for every transferred component** (`restore/entities.ts`, `utils/components.ts`)
   The restore entities stream only recorded component ID mappings when the ID changed between source and destination (via `json.diff`). Components whose ID happened to stay identical were absent from the mapping table, so the links stage couldn't distinguish "transferred with the same ID" from "never transferred". A new schema-driven walker (`collectComponentIdMappings`) pairs the old/new ID of every component instance — components, repeatables, dynamic zones, and nested components — including unchanged IDs, making the entities mapper an exhaustive record of what was transferred.

2. **Skip links whose references were never transferred** (`restore/links.ts`)
   The links stream fell back to the raw source ID when no mapping existed (`mapID(...) ?? originalRef`). Destination rows get fresh auto-increment IDs, so a raw source ID either violates a foreign key constraint — which on PostgreSQL aborts the shared restore transaction and kills the transfer on the *next* statement (`25P02`) — or, worse, silently attaches the relation to an unrelated row. With the mapper now exhaustive, a missing mapping reliably means the referenced row was never transferred (e.g. an orphaned component, or a dangling reference in a source DB without enforced FKs), so the link is skipped with a warning *before* touching the database. The FK-error catch remains as a safety net.

3. **Surface swallowed source stream errors** (`local-source/entities.ts`)
   Errors thrown while streaming entities from the source database were silently swallowed (`catch { /* ignore */ }`), dropping the remaining entities of that content type without a trace and leaving dangling links. They are now reported through the provider diagnostics and the Strapi log (mirroring the assets stream's `onWarning` pattern), naming the content type and the underlying error. The stream still continues with the next content type.

### Why is it needed?

A user reported a remote transfer (~10k entities, ~26k links) failing at the links stage on PostgreSQL. Root cause analysis showed:

- The source database contained component join-table rows (`components_*_lnk`) referencing component instances that were never re-created on the destination: orphaned components are only transferred when attached to a parent entity, but their relation join rows are streamed unconditionally from the raw join tables.
- With no ID mapping, the raw source ID was inserted, violating an FK constraint.
- The "skip orphan links" recovery (added in #19186) relied on a nested transaction (savepoint) around each link insert that was later removed in #22964, so on PostgreSQL the first FK error poisons the whole restore transaction and every subsequent insert fails fatally.

Rather than restoring the savepoint (the single-transaction rollback semantic is intentional), this PR prevents the FK violation from happening in the first place and makes silently-dropped data visible.

### How to test it?

- Unit tests added for the new walker, the links skip behavior, the entities mapping registration, and the source stream error reporting (`yarn jest` in `packages/core/data-transfer`).
- Manual: create orphaned component rows in a source database (delete a parent entity's `_cmps` row but keep the component row + its `_lnk` rows), then run `strapi transfer` to a PostgreSQL destination.
  - Before: transfer fails with `current transaction is aborted, commands ignored until end of transaction block`.
  - After: transfer completes; each orphaned link is reported as a warning.

### Related issue(s)/PR(s)

- #19186 introduced the skip-orphan-links behavior (FK-error catch + nested transaction)
- #22964 removed the nested transaction around link inserts, which broke the FK-error skip on PostgreSQL (aborted transaction)


bolg55:
Confirmed this solved migration/DTS issue for EE customer TID9125

innerdvations (commented):
Initial review looks good, I will give it a more thorough review and QA tomorrow

innerdvations (approved):
LGTM, thanks! I added some missing test coverage.

PR #26532, opened 2026-06-02
chore: normalize prettier formatting and extend .prettierignore

unrevised6419:
### What does it do?

Housekeeping changes, no runtime behaviour change:

- **Prettier formatting** on files that drifted from the repo style:
  - `packages/plugins/color-picker/rollup.config.mjs`, `packages/plugins/i18n/rollup.config.mjs`, `packages/plugins/users-permissions/rollup.config.mjs` — add trailing/terminating semicolons and collapse a double space in an import.
  - `packages/plugins/documentation/server/src/public/{index,login}.html`, `scripts/open-api/public/index.html` — lowercase `<!doctype html>` per Prettier's HTML rule.
- **`.prettierignore`** — ignore generated/vendored trees so they stop producing formatting churn: `.yarn`, `/docs/.docusaurus`, `/docs/docs/exports`, and `packages/generators/generators/src/templates/**/*.hbs`.
- **Docs typo** in `docs/docs/docs/01-core/strapi/commands/01-build.md` — the build emits an `app.js` entry file, not `entry.js`. Confirmed against `packages/core/strapi/src/node/staticFiles.ts` (writes `app.js`) and `create-build-context.ts`.

### Why is it needed?

The listed files were inconsistent with what `prettier --write` produces, so they surfaced as noise in unrelated diffs. The `.prettierignore` additions cover generated output (Docusaurus build, DTS exports, Handlebars generator templates) and the Yarn cache, which should never be reformatted. The docs reference to `entry.js` was simply wrong.

### How to test it?

```sh
yarn prettier --check "packages/plugins/{color-picker,i18n,users-permissions}/rollup.config.mjs" "packages/plugins/documentation/server/src/public/*.html" "scripts/open-api/public/index.html"
```

Should report no formatting issues. For the docs fix, grep the build source to confirm the runtime entry file is `app.js`.

### Related issue(s)/PR(s)

None.

---
Fixes CPR-411

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,9 +3,13 @@ test-apps
 dist
 build
 .strapi
+.yarn
 /.nx/cache
 /.nx/workspace-data
 # Strapi DTS dump fixtures: entire per-fixture dirs (JSONL, metadata.json, assets/metadata, assets/uploads, …)
 tests/e2e/data/**
 # Local DTS export trees (e.g. examples/complex/tmp-dts-export) — same layout; avoid Prettier EOF/newline churn
 **/tmp-dts-export/**
+/docs/.docusaurus
+/docs/docs/exports
+/packages/generators/generators/src/templates/**/*.hbs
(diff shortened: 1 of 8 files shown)


unrevised6419:
Close in favor of https://github.com/strapi/strapi/pull/26765

PR #26536, opened 2026-06-02
Fix Self-Referential Relations Disappearing on Publish

### LINKED ISSUE #26473: Related entries of same content type disappear from referencing entries after publishing
### Node Version

24.15.0

### Package Manager

npm

### Package Manager Version

11.12.1

### Strapi Version

5.46.1

### Operating System

Linux (Debian/Ubuntu)

### Database

MySQL /SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

https://github.com/idmedia-fritscher/strapi-bug-example

### Bug Description

When publishing an entry which is referenced in a Relation field the entry disappears from the "published" Version of the "host" entry with no indication that the "host" entry is not synched between draft and published anymore.

* This only happens when relating the same content type to each other *

Already reported here: https://github.com/strapi/strapi/issues/24342

### Steps to Reproduce

1. Create a content type with a title field and a relation field to the same content type ("_Entry_ *has many* _Entry_") 
2. Create a few instances
3. Reference a few entrys in one "Home" entry
4. Save and Publish
5. Now edit the title of one of the nested entries and publish it.
6. The entry has now dissapeared from the relation field in the "home" entry while the home entry is still shown as fully published and in sync

### Expected Behavior

I expect the entries not to dissappear or at least some indication and the possibiliy to republish the "home" entry without having to edit anything else

### Media

Image
^ Entry in sync with two relations <br><br><br>
    
    
    

Image
^ Click to edit "Test 3122", edit title and publish<br><br><br>
    
  
Image
^ "Test 3122" is gone from the relations list *in published mode*. <br><br><br>

Image

^ In Draft mode both entries are still present but the entry is shown as in sync<br><br><br>

### Confirmation Checklist

Arav-Menon:
What does it do?
- Modifies `unidirectional-relations.ts` in the `document-service` to process self-referential relations instead of indiscriminately skipping them all.

- Adds a targeted `.whereNotIn` query guard so that it only processes self-referential relations when the parent (source)  entry is NOT currently being republished in the exact same batch. (Multi-entry batch publishes are already safely handled   by `selfReferentialRelations`).

- Updates the existing unit tests and adds a comprehensive end-to-end regression test suite to `self-referential-relations.test.api.ts`.

Why is it needed?
- When using a self-referential relation (e.g., a `Category` having a "related" field pointing to another `Category`), independently republishing a child entry caused it to silently disappear from the parent's published relations list in the  database.

- Despite this data loss, the UI would falsely display a "Published & In Sync" badge on the parent entry.

- This happened because `unidirectionalRelations` was hardcoded to completely ignore self-referential relations, assuming `selfReferentialRelations` would handle them. However, `selfReferentialRelations` only kicks in when *both* the parent and child are published simultaneously, leaving independent child publishes in a blind spot.

How to test it?
1. Create a collection type with Draft & Publish enabled.
2. Add a unidirectional relation pointing to itself.
3. Create two entries: "Home" and "Test Child".
4. Edit "Home", link it to "Test Child", and Publish "Home".
5. Edit "Test Child" independently (e.g., change a text field) and Publish it.
6. Look at the Published view of "Home" — verify the relation to the child is perfectly preserved instead of being silently deleted.
7. Note: A dedicated API regression test has been added which can be run via `yarn test:api --testPathPattern self-referential-relations`

Related issue

- Fixes #26473


innerdvations (approved):
Thanks for the fix! I tested it and it works for me. I pushed one additional test update to ensure that the preserved relation points to the expected entry and is correct at the database level, not just present in the count.

PR #26537, opened 2026-06-03
fix(content-manager): widgets show error when role has no access to mainfield of ct

mathildeleg:
### What does it do?
If the main field of a content-type is unreadable (role doesn't have access to it, or other reasons...), the widgets fetch the document's id to display as a fallback instead of the main field.
Added:
* permission check at build time
* post-sanitization verification

### Why is it needed?
If a role has access to parts of a content-type but not its main field, the Widgets in the homepage show a message error.
**BEFORE**
https://github.com/user-attachments/assets/36a59053-2a6b-4f8b-965a-0a2446e404f1

**AFTER** 
https://github.com/user-attachments/assets/873b286f-0a70-4cc5-b8e6-7787579cd7ef

### How to test it?
* In the CTB, create a content type with a few fields.
* In the CM, create an entry for that content type. 
* In the Settings/Roles page, create another Role (different from Super Admin).
* In the Settings/Users page, create a new User with that new role.
* In the Settings/Roles page, modify the access of that role for the content type created previously, so that the main field isn't available for them.
* Log in to the Strapi dashboard with that new User and check that the homepage widgets show that new entry with the entry's documentId as the main field:
Capture d’écran 2026-06-03 à 14 15 19

### DIFF
--- a/packages/core/content-manager/server/src/homepage/services/__tests__/homepage-query-utils.test.ts
+++ b/packages/core/content-manager/server/src/homepage/services/__tests__/homepage-query-utils.test.ts
@@ -1,0 +1,120 @@
+import type { Schema } from '@strapi/types';
+
+import {
+  buildHomepageQueryFields,
+  compactSanitizedFields,
+  FALLBACK_MAIN_FIELD,
+  resolveReadableMainField,
+  resolveTitleField,
+  type HomepagePermissionChecker,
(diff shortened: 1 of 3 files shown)


Adzouz (approved):
Hi @mathildeleg, thanks a lot for the fix 🙏
I tested everything and it works as expected! The code looks also very clean, good job 👏

PR #26542, opened 2026-06-03
chore(tsconfig): scope @types to node and fix leaked-global fallout

unrevised6419:
> [!NOTE]
> This PR is based on top of #26291. Review/merge that first; the diff here builds on it.
> Reason: Conflicting node 20 and 24 undici types

### What does it do?

Adds `"types": ["node", "jest"]` to the shared base tsconfig (`packages/utils/tsconfig/base.json`) so TypeScript stops auto-including **every** package under `node_modules/@types`, then fixes the fallout that this surfaces.

> [!NOTE]
> The ideal scope is just `["node"]`, but `jest` is retained for now to preserve existing behavior — many CI jobs rely on Jest's ambient globals and were failing without it. Properly scoping Jest types to test files is tracked separately in #26543.

By default TS pulls in all `@types/*` into every compilation. Any transitively-installed one can leak globals/types into the whole monorepo. For example `@types/jsdom` references `lib.dom`, which declares a global `var origin: string` and the `RequestInfo` type — neither of which exists in a Node program. Scoping to `["node"]` removes that leakage.

Findings and changes:

- **base tsconfig** (`packages/utils/tsconfig/base.json`): add `"types": ["node", "jest"]` (see note above on why `jest` is kept). Measured on `@strapi/strapi` alone, this drops ~200 files from the build graph (`npx tsc -p tsconfig.build.json --listFiles | wc -l`: **3068 → 2868**), since TS no longer parses/checks those dropped declaration files. base.json is extended by **36 workspace tsconfigs**, so the effect applies across all of them. Summing `--listFilesOnly` across all **51 workspace build configs**, the program input drops from **70,445 → 54,918 files (−15,527, −22%)** — this counts shared `.d.ts` per program, i.e. total parse/check work, not unique files. End result: the **full monorepo build is ~5s faster**. It also prevents stray `@types/*` from leaking web/DOM globals into Node builds.

- **Latent bug surfaced** (`packages/core/data-transfer/src/engine/index.ts`): the `beforeTransfer` error message interpolated `${origin}`, which silently resolved to the leaked DOM global `origin: string` rather than a local variable. Only `provider` is in scope in `runWithDiagnostic`, so this is corrected to `${provider.name}`. Also fixes the `"Unknwon"` → `"Unknown"` typo on the same line.

- **fetch types** (`packages/core/types/src/modules/fetch.ts`): `RequestInfo` is a DOM-only type and is no longer available, so the `strapi.fetch` input type is widened to `string | URL | Request`. The `dispatcher` type is changed from `ProxyAgent` to the broader `Dispatcher`. This is a safe generalization, not a behavior change: in undici `class ProxyAgent extends Dispatcher`, so `Dispatcher` is the base type and still accepts a `ProxyAgent` as well as any other dispatcher (`Agent`, `Pool`, `MockAgent`, etc.).

- **fetch impl** (`packages/core/core/src/utils/fetch.ts`): aligns the function signature and `dispatcher` handling with the type change — assigns `dispatcher` onto a typed `RequestInit` only when set, instead of spreading it conditionally.

- **upgrade** (`packages/utils/upgrade/src/modules/npm/package.ts`): removes a now-unnecessary `@ts-expect-error` on the `dispatcher` option, mirroring the same fetch type change.

- **`response.json()` casts**: the leaked DOM `Body.json()` is typed `Promise<any>` (`lib.dom.d.ts`), which silently allowed property access. The correct undici/node signature is `Promise<unknown>` (`undici-types`), which forces explicit narrowing. So result types are now added at the call sites in `packages/core/core/src/ee/index.ts`, `packages/core/core/src/ee/license.ts`, and `packages/plugins/i18n/server/src/services/ai-localizations.ts`.

- **create-strapi-app** (`packages/cli/create-strapi-app/src/utils/template.ts`): removes a now-unnecessary `@ts-expect-error` on `Readable.fromWeb`.

- **build tsconfigs**: exclude test files from the build (`packages/cli/cloud/tsconfig.build.json` adds `**/*.test.ts`; `packages/core/strapi/tsconfig.build.json` adds `src/test`).

This also aligns with TypeScript 6, which changes the `types` default from "enumerate all of `node_modules/@types`" to `[]` (microsoft/TypeScript#63054, microsoft/TypeScript#62195), with `"*"` as opt-in for the old behavior. The TS team reports 20-50% build-time improvements from setting `types` appropriately; the ~22% input reduction measured here sits in that range. So this is a forward-port of the TS6 default ahead of the bump in #25873 (which already sets `types: ["node"]` on the base preset and drops the same `Readable.fromWeb` `@ts-expect-error`) — landing it standalone de-risks that larger upgrade.

### Why is it needed?

Auto-including all `@types/*` makes builds slower, non-deterministic (whatever is installed leaks in), and — as shown by the `origin` bug — actively hides errors by providing globals/types that don't exist at runtime. Scoping to `["node"]` makes the type environment match the actual runtime and turns latent bugs into compile errors.

### How to test it?

- `npx tsc -p tsconfig.build.json --listFiles | wc -l` in `packages/core/strapi` before/after to observe the ~200-file reduction.
- Build the affected packages (`@strapi/core`, `@strapi/data-transfer`, `@strapi/types`, `@strapi/strapi`, `@strapi/i18n`, `create-strapi-app`, `@strapi/cloud-cli`) and confirm typecheck passes.

### Related issue(s)/PR(s)

- #26291 (node/undici `@types/node` alignment — this PR builds on top of it)
- #26543 (scope Jest types to test files — follow-up to drop `jest` from the base `types` here)
- #26459 (per-workspace tsconfig scoping)
- #25873 (TypeScript 6 bump — sets `types: ["node"]` on the base preset; this PR forward-ports that change in isolation)
- microsoft/TypeScript#63054 / microsoft/TypeScript#62195 (TS6 changes the `types` default to `[]`)

### Possible follow-up

Scoping to `["node"]` still keeps Node's **ambient globals** in scope (`process`, `Buffer`, `__dirname`, `global`, `setTimeout`, etc.). We could go one step further and set `"types": []`, then `import` everything explicitly instead of relying on globals — e.g. `import process from 'node:process'`, `import { Buffer } from 'node:buffer'`, `import { setTimeout } from 'node:timers'`. Most of `@types/node`'s surface is already module-based and works via `import type`, so only the ambient globals need this treatment.

Tradeoff: fully deterministic, zero global leakage (the `origin`-style bug class disappears entirely), at the cost of more explicit imports. Could be a separate PR if there's appetite — calling it out here for visibility.

---
Fixes CPR-408


packages/core/core/src/utils/fetch.ts
```diff
@@ -14,21 +14,24 @@ export const createStrapiFetch = (
...
-  function strapiFetch(url: RequestInfo | URL, options?: RequestInit) {
-    const fetchOptions = {
-      ...(strapiFetch.dispatcher ? { dispatcher: strapiFetch.dispatcher } : {}),
+  const strapiFetch: Modules.Fetch.Fetch = (url: string | Request | URL, options?: RequestInit) => {
+    const fetchOptions: RequestInit = {
       ...options,
     };
 
+    if (strapiFetch.dispatcher) {
+      fetchOptions.dispatcher = strapiFetch.dispatcher;
```
unrevised6419, line 23: > [!NOTE]
> TypeScript will show error here without the base PR because there are conflicts of inconsistent undici types between node 20 and 24

unrevised6419:
Closing in favor of https://github.com/strapi/strapi/pull/26699

PR #26544, opened 2026-06-04
chore(tsconfig): centralize module-augmentation types (koa-body, @strapi/types)

unrevised6419:
> [!NOTE]
> Stacked on #26542 — merge that first. Both edit the same `types` line in `packages/utils/tsconfig/base.json`; stacked only to avoid a conflict, otherwise independent.

### What does it do?

Replaces scattered side-effect imports (used to keep global module augmentations alive) with explicit, central type configuration — for both **koa-body** and the ambient **`global.strapi`**.

#### koa-body

koa-body augments koa's `Request` (`ctx.request.body` / `ctx.request.files`) but ships its own types and is **not** an `@types/*` package, so TS never auto-includes it. The augmentation was kept alive by `import 'koa-body'` side-effect imports scattered across server files — fragile (look unused → deleted by accident, easy to forget in new files).

- **base tsconfig** (`packages/utils/tsconfig/base.json`): `types: ["node"]` → `["node", "koa-body"]` (+ `koa-body` dep on the `tsconfig` package). Forces the augmentation into every workspace extending base.
- **Drop** the now-redundant `import 'koa-body'` side-effect imports and matching devDependencies.
- **Server tsconfigs overriding `types`** (admin, content-manager, content-type-builder) list `koa-body` explicitly, replacing the unnecessary `lodash` entry.
- **Published d.ts** (`packages/core/core/src/index.ts`): one `import 'koa-body'` at the `@strapi/core` entry point so the augmentation rides along in emitted declarations for external consumers.
- **Remove** unused `packages/core/content-type-builder/tsconfig.build.json`.
- Dead-code cleanup (documentation plugin): koa-session v7 has no `declare module 'koa'`, so `import type {} from 'koa-session'` in `restrict-access.ts` was a no-op (`ctx.session` is `any` via `@types/koa`'s index signature). Removed, plus the unused `koa-session` devDependency.

#### `global.strapi` → opt-in `@strapi/types/globals-server`

Same fragility: `@strapi/types` declares the ambient `var strapi: Strapi`, kept in scope by scattered `import '@strapi/types'` / `import '@strapi/strapi'` side-effect imports. Removing them produces **no** TS errors — the files don't depend on the global being statically typed. Instead of forcing it onto every `@strapi/types` consumer, it's now **opt-in per workspace**:

- **New `@strapi/types/globals-server`** (`src/globals-server.ts`): the `declare global { var strapi }` block moves here, out of the `@strapi/types` index, so importing `@strapi/types` for its types no longer drags the global along. Resolved via a root shim (`globals-server.d.ts` → `export * from './dist/globals-server'`) rather than a package `exports` subpath, since the monorepo's tsconfig module resolution doesn't honour `exports` for types.
- **Per-package `globals.d.ts`**: workspaces that rely on the global opt in with `/// <reference types="@strapi/types/globals-server" />` (admin +ee, content-manager, content-type-builder, core, data-transfer, email, review-workflows, strapi, upload, upload-local, graphql, i18n; documentation/content-releases migrated from their old `global.d.ts`/`custom.d.ts`).
- **Back-compat**: `import '@strapi/types/globals-server'` at the `@strapi/strapi` entry point keeps `global.strapi` in the published types so external apps don't break.
- **Deprecated**: `var strapi` marked `@deprecated since version 5.x. Use a factory function instead.` The old `NodeJS.Global` augmentation is dropped — under `@types/node` v20, `global` is `typeof globalThis`, so it was a no-op. Stale `declare const strapi: any;` in permissions deleted.

Declaring the global per-workspace lets it be **removed workspace-by-workspace** as each migrates to factory functions. This is the server-side complement to #26213 (browser `window.strapi` / `Admin.BrowserStrapi`); together they untangle the node-vs-browser `Strapi` global collision.

New code needing `global.strapi`: use a factory function, or add the `globals-server` reference to the workspace.

> [!NOTE]
> **No consumer impact** from dropping the imports: side-effect imports aren't emitted into `build:types` `.d.ts`, so external consumers see no change — the effect is confined to monorepo typechecking. The `@strapi/core` (koa-body) and `@strapi/strapi` (globals-server) entry-point imports preserve what consumers *do* rely on.

> [!NOTE]
> **Cosmetic caveat (GraphQL plugin consumers).** An app whose `tsconfig.json` sets no `types` field and installs `@strapi/plugin-graphql` gets two `Request.body` augmentations in scope: `koa-body` (`any`, via `@strapi/core`) and `@types/koa-bodyparser` (`unknown`, a graphql dep, auto-included). Not a compile error in practice — apps build with `skipLibCheck: true`, which suppresses the `TS2717` (reported at the `.d.ts` site); interfaces merge, `body` resolves to `any`. Only visible effect: editor "go to definition" lists both sources. Fix for consumers: scope `types` (e.g. `["node"]`).

### How to test it?

```
yarn nx run-many --target=build:types --projects=@strapi/admin,@strapi/content-manager,@strapi/content-type-builder,@strapi/permissions,@strapi/review-workflows,@strapi/types,@strapi/upload,@strapi/provider-upload-local --nx-ignore-cycles
```

Confirms `ctx.request.body` / `ctx.request.files` and `global.strapi` resolve. All pass on this branch.

### Related issue(s)/PR(s)

- #26542 — scope `@types` to `["node"]` (this PR is git-stacked on it)
- #26213 — centralizes browser `window.strapi` / `Admin.BrowserStrapi` (complementary)

---
Fixes CPR-409


packages/core/admin/server/tsconfig.json
```diff
@@ -4,7 +4,7 @@
   "compilerOptions": {
     "rootDir": "../",
     "baseUrl": ".",
-    "types": ["lodash", "jest"],
```
unrevised6419: Note: `lodash` does not have any augmentations.

packages/core/core/src/index.ts
```diff
@@ -1,3 +1,10 @@
+// Augmentation. Keep in entry point.
+// On build:types, nested imports may be removed from d.ts files because they are runtime only.
+// This works for Strapi consumers that use `@strapi/strapi`.
+// For Strapi consumers that use `@strapi/types` won't work.
+// Within workspaces tsconfig.json takes care of this.
+import 'koa-body';
```
unrevised6419, line 6: Note: Previously the `koa-body` global augmentation was working by having a random bare import (undocumented) of `koa-body` in `packages/core/types/src/core/controller.ts`. And `@strapi/strapi` is exporting everything from `@strapi/types`

Note: Right now to consumer are exposed 2 `context.request.body` global augmentations. `koa-body` and `@types/koa-bodyparser`

`koa-body` global augmentation is exposed via strapi imports, `@types/koa-bodyparser` gets exposed to consumers by not having an explicit `tsconfig.json#compilerOptions.types`

Note: Probably we may want also to tighten the `tsconfig.json#compilerOptions.types` in all examples/templates.

packages/core/strapi/src/index.ts
```diff
@@ -1,3 +1,7 @@
+// Keep back compatibility of deprecated global.strapi for consumers
+// TODO: Remove in v6
+import '@strapi/types/globals-server';
```
unrevised6419: Note: The ideal case would be to drop this completely. And ask consumers to add to their `tsconfig.json#compilerOptions.types` the `@strapi/types/globals-server`

Dropping this line, there won't be anymore conflicting `global.strapi` and `window.strapi`, each context admin/server imports their own global variable

packages/core/types/package.json
```diff
@@ -46,6 +46,10 @@
       "types": "./dist/_internal/globals-admin.d.ts",
       "source": "./src/_internal/globals-admin.ts"
     },
+    "./_internal/globals-server": {
+      "types": "./dist/_internal/globals-server.d.ts",
+      "source": "./src/_internal/globals-server.ts"
```
unrevised6419, line 51: Note: We have this `source` export condition everywhere, but is not used anywhere from what I see. Probably it was intended to be used with [`tsconfig.json#compilerOptions.customConditions`](https://www.typescriptlang.org/tsconfig/#customConditions) but it never was.

From my understanding we could use this custom conditions to expose to consumers same current `global.strapi` augmentation, but internally we could have a different entry-point for `@strapi/types` that does not have `global.strapi`

packages/core/strapi/src/index.ts
```diff
@@ -1,3 +1,7 @@
+// Keep back compatibility of deprecated global.strapi for consumers
+// TODO: Remove in v6
+import '@strapi/types/_internal/globals-server';
```
unrevised6419: Dropping this line, there won't be anymore conflicting global.strapi and window.strapi, each context admin/server imports their own global variable.

The ideal case would be to drop this completely. And ask consumers to add to their `tsconfig.json#compilerOptions.types` the `@strapi/types/_internal/globals-server`

I was thinking maybe we can also add this to templates `tsconfig.json#compilerOptions.types`, and later we can tell the users to remove it when `global.strapi` is removed.

packages/core/admin/shared/utils/session-auth.ts
```diff
@@ -1,3 +1,6 @@
+// TODO: This exposes the `global.strapi` to Admin side of `@strapi/admin`
+/// <reference types="@strapi/types/_internal/globals-server" />
```
unrevised6419: Note: This shared folder exports both admin and server types that are used also in `ee` code. Probably would be better to split it shared admin and shared server.

unrevised6419:
@nclsndr this is also ready to go after #26213

PR #26547, opened 2026-06-04
chore(v4): support node 22

Bassel17:
### What does it do?

Expands Strapi v4 Node.js engine support from `>=18.0.0 <=20.x.x` to `>=18.0.0 <=22.x.x` across package manifests and generated app metadata.

Updates `better-sqlite3` to `11.10.0` in the example and test app templates so SQLite installs and runs correctly on Node 22 while continuing to work on Node 18 and Node 20.

Node 24 is intentionally not included because it would require broader dependency and runtime validation for the final v4 release.

### How to test it?

should build and run on node 18, node 20 and node 22 with no issues


innerdvations (changes requested):
We also need to bump the create-strapi-app template better-sqlite version

Other than that, LGTM

innerdvations:
Sorry, caught one more thing: the plugin-package .hbs files need to be updated with node 22

Bassel17:
should be good now @innerdvations

innerdvations (approved):

PR #26549, opened 2026-06-04
enhancement(admin): remove deploy-now widget in production mode

Sheryoo:
### What does it do?

This change only removes the deploy-now widget in production, like the "deploy to cloud page"

### How to test it?

Run the app in production and check for the 'deploy-now' widget; it should not be there

### Related issue(s)/PR(s)

[Remove the deploy widget](https://feedback.strapi.io/customization/p/option-to-remove-the-default-deploy-widget-in-dashboard)

---
Fixes CPR-385

### DIFF
--- a/packages/core/admin/admin/src/StrapiApp.tsx
+++ b/packages/core/admin/admin/src/StrapiApp.tsx
@@ -327,6 +327,8 @@ class StrapiApp {
   getPlugin = (pluginId: PluginConfig['id']) => this.plugins[pluginId];
 
   async register(customRegister?: unknown) {
+    const { backendURL } = window.strapi;
+
     this.widgets.register([
       {
         icon: User,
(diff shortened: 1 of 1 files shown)


innerdvations:
Since there is a merge conflict that covers the entire PR, and we also need to replace the 'localhost' check with something more robust, this has been superseded by https://github.com/strapi/strapi/pull/26660 -- I will credit you in the commit on that PR :)

Sheryoo:
@innerdvations then you should also update the cloud plugin to use the env 'production' as it's using the localhost as a check also to show the deploy page [cloud plugin's index](https://github.com/strapi/strapi/blob/develop/packages/plugins/cloud/admin/src/index.ts) look at line 19 as i used it as a ref to my work

innerdvations:
Good catch — #26660 only updated the homepage deploy-now widget; the cloud plugin Deploy sidebar was still gated on `localhost`.

Follow-up fix: https://github.com/strapi/strapi/pull/26733

_AI-assisted comment._

PR #26551, opened 2026-06-04
Future media lib sidebar

jhoward1994:
### What does it do?

Adds a left sidebar folder tree to the revamped Media Library (`AssetsPage`). The tree loads the full hierarchy from `GET /upload/folder-structure`, shows Home and nested folders, and stays in sync with the current `?folder=` URL. Selecting a folder navigates via `useFolderNavigation`; ancestors of the active folder auto-expand. Loading and error states are handled in the FOLDERS section, and invalid or missing folder IDs redirect to root.

### Why is it needed?

The Media Library revamp needs sidebar navigation so users can browse nested folders without relying on the main grid alone.

### How to test it?

1. Check out this branch and run `yarn && yarn setup` (or build `@strapi/upload`).
2. Start the sandbox with the unstable Media Library enabled:
   ```bash


Adzouz:
Hey @jhoward1994, thanks for the PR 🙏
I was testing first before digging in the code and the left sidebar doesn't take the whole height:
Screenshot 2026-06-09 at 15 25 10

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTree.tsx
```diff
@@ -0,0 +1,181 @@
...
+              </Loader>
+            </Flex>
+          ) : isError ? (
+            <Flex direction="column" gap={2} paddingLeft={2} paddingRight={2}>
+              <Typography variant="pi" textColor="danger600">
+                {formatMessage({
+                  id: getTranslationKey('sidebar.tree.error'),
+                  defaultMessage: 'Could not load folders.',
+                })}
+              </Typography>
```
markkaylor, line 386: Thanks for handling this, I think we should check with Anka though for the design of the loading, empty, and error state.

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTree.tsx
```diff
@@ -0,0 +1,181 @@
...
+              {onRetry ? (
```
markkaylor: what's retry? like retry to load the folders?

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/RowButton.tsx
markkaylor, line 1: In my opinion FolderTree is the only thing that should be exported. FolderTreeItem and RowButton can all live inside that file separated cleaning with comment sections. That way its clear FolderTreeItem and RowButton are only consumed by FolderTree.

packages/core/upload/admin/src/future/pages/Assets/hooks/useFolderNavigation.ts
```diff
@@ -1,18 +1,61 @@
...
+   */
+  const navigateToRoot = useCallback(() => {
+    setQuery({ folder: '' }, 'remove');
+  }, [setQuery]);
+
+  useEffect(() => {
+    if (query?.folder && currentFolderId === null) {
+      navigateToRoot();
+    }
+  }, [query?.folder, currentFolderId, navigateToRoot]);
```
markkaylor, line 42: This is like a redirect if a folder doesn't exist?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -183,7 +188,23 @@ export const AssetsPage = () => {
...
+  const { error: currentFolderError } = useGetFolderQuery(
+    { id: currentFolderId! },
+    { skip: currentFolderId === null }
+  );
+
+  useEffect(() => {
+    if (currentFolderError?.name === 'NotFoundError') {
+      navigateToRoot();
+    }
+  }, [currentFolderError, navigateToRoot]);
```
markkaylor, line 199: Isn't the useFolderNavigation hook already doing this?

packages/core/upload/admin/src/future/pages/Assets/AssetsPage.tsx
```diff
@@ -262,14 +286,27 @@ export const AssetsPage = () => {
...
+            background="neutral0"
+            sideNav={
+              <FolderTree
+                folderStructure={folderStructure}
+                currentFolderId={currentFolderId}
+                onSelectFolder={navigateToFolderId}
+                isLoading={isFolderStructureLoading}
+                isError={isFolderStructureError}
+                onRetry={refetchFolderStructure}
+              />
```
markkaylor: My intuition was the data fetching could be moved directly inside of FolderTree instead of passed props down. I asked claude about it. Here's the response but could be trying to make me happy:

My recommendation: move useGetFolderStructureQuery into FolderTree; keep currentFolderId/onSelectFolder as props.

Reasoning:

The folder-structure query has a single consumer. Fetching it in the page only to drill four props down is pure plumbing with no benefit. Co-locating the fetch with its sole consumer is the cleaner default.

RTK Query makes this safe. useGetFolderStructureQuery is cache-deduped by RTK Query, so even if some sibling later needs the structure, calling the hook in two places hits the same cache entry — no double request. You're not giving up shareability by moving it down.

Navigation should stay in the page. currentFolderId/onSelectFolder are URL state, and the component's own docstring (lines 50-52) is explicit that it's presentational with respect to the URL — it asks the parent to navigate so the URL stays the single source of truth. That's correct and shouldn't change. The page owns navigation because multiple things depend on it.

markkaylor (commented):
Nice!!!! It's working really well as far as I can tell. Left some thoughts here but looks good. 

Maybe a tooltip when the name is truncated
Screenshot 2026-06-10 at 09 50 44

Adzouz:
Screenshot 2026-06-10 at 11 24 54

Still testing and started the review. I found also that the display mixing folders with and without subfolders is a bit "weird". Accessibility and responsive look good (if intended to hide completely the sidebar on mobile). Here I'm hovering the Another Fourth item while being in Third folder.
Also there's a tiny gap between elements on Figma of 4px if you can add it too so the elements don't seem glued together 🙏

And when there's no folder, the message is a bit too close from the title:
Screenshot 2026-06-10 at 12 09 41

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTree.tsx
Adzouz: Yeah I guess that's the purpose of the button but I think it would be better renamed because onRetry feels more like an ongoing action and not a trigger. Something like triggerRetry or handleRetry would be better no?

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTreeItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+ */
+export const FolderTreeItem = ({
+  node,
+  level,
+  currentFolderId,
+  isExpanded,
+  onToggle,
+  onSelect,
+}: FolderTreeItemProps) => {
+  const { formatMessage } = useIntl();
```
Adzouz: We don't really need to load it if node.id is null. Maybe we can move this after the condition

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTree.tsx
markkaylor: But seems weird to have this no? We don't do have that anywhere else in Strapi when data fetching fails. I don't think this was in the designs either.

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTreeItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+        ) : (
+          <Spacer aria-hidden />
+        )}
+
+        <Box flex="1" minWidth={0}>
+          <RowButton
+            type="button"
+            $isActive={isActive}
+            aria-current={isActive ? 'page' : undefined}
+            aria-expanded={hasChildren ? expanded : undefined}
```
Adzouz: The aria-expanded should be on the toggle icon instead

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTreeItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+          >
+            <Typography variant="omega" fontWeight={isActive ? 'semiBold' : 'regular'} ellipsis>
+              {name}
+            </Typography>
+          </RowButton>
+        </Box>
+      </Flex>
+
+      {expanded && (
+        <Box tag="ul" margin={0} padding={0} style={{ listStyle: 'none' }}>
```
Adzouz: You could use NavList from FolderTree

Adzouz (commented):
Overall looks good, good job 👏 I left a few nit picks but in general it's looking good and no blocking issue. Just a few UI adjustments (already posted as comments) to address and we'll be good to go imho 🙏

Adzouz:
Ah also I saw that your branch is a feat, we decided to go with future to not trigger a new minor version with what we're doing on the ML revamp so just be careful on the squashed commit message 🙏

jhoward1994:
I think we should add collapse/expand arrows to all folders regardless of whether there are child folders? that would avoid the weird alignment issues you saw and matches e.g. google drive behaviour

packages/core/upload/admin/src/future/pages/Assets/components/FolderTree/FolderTree.tsx
```diff
@@ -0,0 +1,413 @@
...
+            data-testid={`folder-tree-node-${id}`}
+            data-folder-id={id}
+          >
+            <TruncatedFolderName name={name} isActive={isActive} />
+          </RowButton>
+        </Box>
+      </Flex>
+
+      {hasChildren && isFolderExpanded && (
+        <Box tag="ul" margin={0} padding={0} style={{ listStyle: 'none' }}>
```
Adzouz: Sorry I forgot this one too, can be NavList component

Adzouz (changes requested):
Thanks for the changes @jhoward1994, after testing, it's better but I have an arrow that toggles nothing when on the last child:
Screenshot 2026-06-17 at 15 39 40

Adzouz (approved):

PR #26552, opened 2026-06-04
chore: remove experimental-dev example app

innerdvations:
### What does it do?

Removes the `examples/experimental-dev` example application and regenerates the lockfile (`yarn install` + `yarn dedupe`) to drop the dependencies that were unique to it.

### Why is it needed?

`experimental-dev` was added in #20327 (June 2024) for a single purpose: a sandbox to test **React 19 release candidates** and the **React Compiler** ahead of their stable releases, gated behind the `USE_EXPERIMENTAL_DEPENDENCIES` flag.

Both have since shipped stable:

- **React 19** — stable since Dec 2024 (React 18 is now security-only upstream).
- **React Compiler** — stable **v1.0** since Oct 2025.

So the app's reason for existing is obsolete. In its current state it:

- pinned `react: "rc"` / `react-dom: "rc"` (a moving tag) and `babel-plugin-react-compiler@0.0.0-experimental-c23de8d-20240515` (a ~2-year-old experimental snapshot);
- was **not wired into CI** (no workflow or e2e references), so it caught nothing automatically;
- received **no feature work** since creation — only dependency/security/release-bot churn.

Removing it eliminates the stale experimental pins and the recurring Dependabot noise they generate (e.g. the `babel-plugin-react-compiler` bump, and the `react@"rc"` churn).

The `USE_EXPERIMENTAL_DEPENDENCIES` CLI flag and its unit tests in `@strapi/strapi` are intentionally left untouched — they are shipped behavior independent of this example.

The broader React 19 + React Compiler migration of the admin is tracked separately (see Related).

### How to test it?

- `yarn install` resolves cleanly with the lockfile in this PR.
- `yarn build` / `yarn test:unit` / `yarn test:front` pass (no package referenced `experimental-dev`).
- Other example apps (`getstarted`, `kitchensink`, etc.) are unaffected.

### Related issue(s)/PR(s)

- Fixes CMS-1135
- Original app: #20327
- React 19 migration tracking: CMS-649


nclsndr (approved):
LGTM ❤️

PR #26427, opened 2026-05-25
fix(utils): ignore empty sort when building orderBy

innerdvations:
### What does it do?

Adds a `hasSort` type guard in `convert-query-params` so empty or missing `sort` (`[]`, `""`, `undefined`) does not set `orderBy`. Join-table ordinal ordering is preserved when GraphQL nested fields use the default `sort: []` argument.

Includes unit tests for `transformQueryParams` and API tests for:

- [#26426](https://github.com/strapi/strapi/issues/26426) — single type → component → `oneToMany` relation order via GraphQL (`relation-order.test.api.js`)
- [#26465](https://github.com/strapi/strapi/issues/26465) — collection type → repeatable component order via GraphQL (`repeatable-component-order.test.api.js`)

### Why is it needed?

Fix #26426, #26465

After Strapi 5.45+, GraphQL nested fields (repeatable components, relations inside components, etc.) pass default `sort: []`, which was converted to `orderBy: []`. The database layer treated that as an explicit sort and skipped join-table ordering, so children were returned in `id` order instead of the order set in the admin UI. REST was largely unaffected unless clients explicitly sent `sort: []`.

### How to test it?

```bash
yarn nx run @strapi/utils:build
yarn nx run @strapi/utils:test:unit --testPathPattern=convert-query-params
yarn test:api --generate-app tests/api/plugins/graphql/relation-order.test.api.js
yarn test:api --generate-app tests/api/plugins/graphql/repeatable-component-order.test.api.js
```

Manual check (optional):

- **#26426:** single type with a component containing a reorderable `oneToMany` relation — GraphQL relation order should match the admin.
- **#26465:** content type with a repeatable component (e.g. Page → Chapters) — GraphQL component list order should match the admin.

### Related issue(s)/PR(s)

- Fixes #26426
- Fixes #26465

PR #26553, opened 2026-06-04
fix(database): restore join-table relation sort order in components

### LINKED ISSUE #26550: Component relation sorting stopped working after v5.46.0
### Node Version

20.19.2

### Package Manager

yarn

### Package Manager Version

1.22.22

### Strapi Version

v5.47.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

When strapi was upgraded from `v5.46.0` to `v5.47.0`, the relations sorting in components stopped working.

Image

The API response in `v5.46.0` used to be same exact as Strapi admin UI. API response in `v5.47.0` is not. Resulting users can not order it from Strapi UI.

### Steps to Reproduce

1. Use strapi `v5.46.0`.
2. Create a content type with one-to-many relation of another content type.
3. Create a component with field that uses this content type.
4. In API response, the order of array of this content type is same as Strapi UI.
5. Upgrade to strapi `v5.47.0`.
6. Now the order breaks

### Expected Behavior

By default, the order of API response should be exact same as Strapi UI and upon reordering in Strapi UI should reflect in API response.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

We are using GraphQL.

### Confirmation Checklist

Arav-Menon:
## What does it do?

This fixes an issue where the database was ignoring the custom drag-and-drop order of relations inside components. 

When you request data through the API without specifying a custom sort order, Strapi internally passes an "empty" sorting instruction. Due to a bug in how this empty instruction was checked, Strapi mistakenly thought a custom sort was being requested and completely dropped the default UI drag-and-drop order. 

This fix ensures that Strapi correctly recognizes when no custom sort is requested, preserving your manual UI ordering. Explicit API sorts (like `?sort=name:asc`) will still work and override the UI order as intended.

## Why is it needed?

After upgrading to v5.47.0, relations inside components no longer respect the drag-and-drop order set by editors in the Strapi Admin UI. The API returns items in the order they were created in the database instead, breaking any frontend websites that rely on content-managed ordering.

## How to test it?

1. Create a Collection Type (`Article`) with a **Component** field.
2. Inside the component, add a **one-to-many relation** pointing to another Collection Type (e.g. `Category`).
3. In the Content Manager, create 3 categories (`Alpha`, `Beta`, `Gamma`) and add them to an Article.
4. Use the **drag-and-drop handle** to reorder them: `Gamma → Alpha → Beta`.
5. Save and Publish.
6. Call `GET /api/articles?populate=componentField.categories`.
7. The `categories` array in the JSON response should now match the UI order: `Gamma → Alpha → Beta`.
8. Also verify that calling with an explicit sort like `?populate[componentField][populate][categories][sort]=name:asc` still returns `Alpha → Beta → Gamma` (explicit sort still wins).

## Related issue(s)/PR(s)

Closes #26550


innerdvations (approved):
LGTM! We actually already fixed this in #26427 but this is a good additional guard to have. Thanks!

PR #26559, opened 2026-06-05
Feat/e2e critical ctb add fields

MarionLemaire:
## What
New "@critical" E2E test: adds a component, DZ, and relation to an existing CT and verifies they render in the CM editor.

## Why
Critical path #3 (ctb.add-component-dz-relation) was not covered. The existing create-collection-type spec creates from scratch but never verifies CM rendering. This test closes that gap.

## What the test does
1. Starts from the with-admin fixture (Article CT exists)
2. Adds a component, a DZ, and a oneWay relation to Author
3. Navigates to CM → Article → Create new entry
4. Asserts all 3 fields are visible in the editor

## Note
This test triggers multiple server restarts and uses a flow that hasn't been tested before on an existing CT. The assertions may need small adjustments after the first CI run (e.g. exact label names in CM). That's expected — flag any failures as mechanical, not behavioral.

## Part of
E2E coverage focus — Phase 1


jhoward1994:
I think the changes to `tests/e2e/tests/content-manager/conditional-fields/boolean-conditional-select-relation.spec.ts` can be removed from this PR as theyve been done in https://github.com/strapi/strapi/pull/26558 ?

jhoward1994 (dismissed):
Looks good 👍🏻 provided the CI passes

jhoward1994 (approved):

PR #26560, opened 2026-06-05
fix(content-manager): reduce MCP relation output to identity-only shape

nclsndr:
### What does it do?

- Adds `shapeRelationsForMcp` sanitizer (`mcp/sanitizers/shape-relations.ts`) that walks a document via `traverseEntity` and reduces every relation field to `{ documentId, locale?, __type? }`, stripping all target scalars
- Adds `relationsAsIdentity()` to `populate-builder`, backed by a new `identityRelations` flag in `getDeepPopulate`, so non-morph relations fetch only `{ fields: ['documentId', 'locale?'] }` at the DB layer
- Adds `sanitizeAndShape()` utility as a single output chokepoint that chains `permissionChecker.sanitizeOutput` → `shapeRelationsForMcp`
- Replaces `countRelations()` + bare `sanitizeOutput` calls in all MCP collection and single-type handlers with `relationsAsIdentity()` + `sanitizeAndShape()`
- Tightens MCP output schemas so relation fields are typed as identity objects; simplifies delete output to `z.object({}).loose()`

### Why is it needed?

MCP handlers returned fully-populated relation documents after `sanitizeOutput`. The permission checker only covers the caller's content type, so target-document scalars (including private or sensitive fields) leaked into every MCP response containing a relation.

### How to test it?

1. Create two content types, e.g. `article` (with a `category` relation) and `category` (with a private field such as `secret`)
2. Populate `category.secret` with data
3. Call `get_article` or `list_article` via the MCP server
4. Confirm the `category` field in the response contains only `{ documentId }` (plus `locale` if i18n is enabled) and that `secret` is absent

### Related issue(s)/PR(s)

- ø

### DIFF
--- a/packages/core/content-manager/server/src/mcp/handlers/collection-handlers.ts
+++ b/packages/core/content-manager/server/src/mcp/handlers/collection-handlers.ts
@@ -96,7 +96,6 @@ export const createCollectionListHandler =
     const populate = await getService('populate-builder')(uid)
       .populateFromQuery(permissionQuery)
       .populateDeep(1)
-      .relationsAsIdentity()
       .withPopulateOverride(getPopulateForLocalizations(uid))
       .build();
 
@@ -162,7 +161,6 @@ export const createCollectionGetHandler =
     const populate = await getService('populate-builder')(uid)
       .populateFromQuery(permissionQuery)
       .populateDeep(Infinity)
-      .relationsAsIdentity()
       .withPopulateOverride(getPopulateForLocalizations(uid))
       .build();
 
--- a/packages/core/content-manager/server/src/mcp/handlers/collection-handlers.ts
+++ b/packages/core/content-manager/server/src/mcp/handlers/collection-handlers.ts
@@ -12,7 +12,7 @@ import {
   MCP_NOT_FOUND_OR_PUBLISHED,
 } from './constants';
 import { isContentTypeLocalized } from '../permissions';
-import { ok } from '../utils';
+import { ok, sanitizeAndShape } from '../utils';
 
 type McpDocumentQuery = {
   populate?: unknown;
@@ -96,7 +96,7 @@ export const createCollectionListHandler =
     const populate = await getService('populate-builder')(uid)
       .populateFromQuery(permissionQuery)
       .populateDeep(1)
-      .countRelations({ toOne: false, toMany: true })
+      .relationsAsIdentity()
       .withPopulateOverride(getPopulateForLocalizations(uid))
       .build();
 
@@ -129,7 +129,7 @@ export const createCollectionListHandler =
 
     const results = await asyncPipe.map(
       documents,
-      asyncPipe.pipe(permissionChecker.sanitizeOutput, setStatus)
+      asyncPipe.pipe((doc: unknown) => sanitizeAndShape(permissionChecker, uid, doc), setStatus)
     );
 
     return ok({ results, pagination } as Record<string, unknown>);
@@ -162,7 +162,7 @@ export const createCollectionGetHandler =
     const populate = await getService('populate-builder')(uid)
       .populateFromQuery(permissionQuery)
       .populateDeep(Infinity)
-      .countRelations()
+      .relationsAsIdentity()
       .withPopulateOverride(getPopulateForLocalizations(uid))
       .build();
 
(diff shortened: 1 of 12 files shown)


packages/core/content-manager/server/src/mcp/handlers/collection-handlers.ts
```diff
@@ -96,7 +96,7 @@ export const createCollectionListHandler =
     const populate = await getService('populate-builder')(uid)
       .populateFromQuery(permissionQuery)
       .populateDeep(1)
-      .countRelations({ toOne: false, toMany: true })
+      .relationsAsIdentity()
```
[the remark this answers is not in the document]
nclsndr: Intent is valid

packages/core/content-manager/server/src/services/populate-builder.ts
```diff
@@ -81,7 +92,11 @@ const populateBuilder = (uid: UID.Schema) => {
         return initialPopulate;
       }
 
-      return getDeepPopulate(uid, { ...deepPopulateOptions, initialPopulate });
+      return getDeepPopulate(uid, {
+        ...deepPopulateOptions,
+        initialPopulate,
+        identityRelations: deepPopulateOptions.relationMode === 'identity',
```
innerdvations:
```suggestion
        identityMode: deepPopulateOptions.relationMode === 'identity',
```
Is this more clear?

packages/core/content-manager/server/src/services/utils/populate.ts
```diff
@@ -67,6 +77,24 @@ function getPopulateForRelation(
     return true;
   }
 
+  // MCP identity mode: fetch only stable identity fields from the DB layer.
+  // Morph relations cannot use flat fields; they are populated normally and shaped post-fetch.
+  if (identityRelations === true) {
```
innerdvations: We should either document this, or explicitly label it unstable/internal in comments/types (I would opt for ruling it internal-only). Otherwise we would probably need to add a lot more tests, and ensure it works as expected when used in the content API.

innerdvations (commented):
Ignoring my two comments about the new populate mode:

Do we actually need a new populate mode? It might be better to just over-select the relation data and then reshape it away at the MCP output boundary. It's less efficient but it has a smaller footprint and is easier to guarantee for security.

We would keep a single output chokepoint (easier to audit) and it avoids touching the same populate code that the rest api is using (adding risk there).

What do you think?

nclsndr:
> Do we actually need a new populate mode?

Not really. This implementation is actually a simplified version of a more ambitious refactor (that eventually regressed the admin, discarded for that reason).
I'm looking into a simpler solution in the direction you mention @innerdvations

nclsndr:
@innerdvations, new iteration on: https://github.com/strapi/strapi/pull/26560/changes/e0e51c6d6547267ff085f366f517072ee8393da0

innerdvations:
That's the right direction, but there are a couple issues:
1) It deletes the timestamps, then tries to calculate draft/published status from those same timestamps, so the status comes out wrong. It should calculate first, then strip.
2) for `morph*` and `manyWay` types of relations, the response sends a list but the schema says it's a single item, they should both use the same "is this a list" check
3) the tests mock the actual logic so it's not actually testing the issue is fixed; we could add integration tests to see that it's working exactly as expected

nclsndr:
Thanks for catching this up @innerdvations! 

> we could add integration tests to see that it's working exactly as expected

We'll probably need to bring the [MCP inspector](https://modelcontextprotocol.io/docs/tools/inspector) to do this properly

nclsndr:
@innerdvations just pushed a new iteration

nclsndr:
@pwizla I'm not so sure those changes impacts the docs just yet. I did not remember we went into such details on returned payload so far.

pwizla:
> @pwizla I'm not so sure those changes impacts the docs just yet. I did not remember we went into such details on returned payload so far.

Inki says ([`/inki:route` skill](https://github.com/strapi/documentation/blob/main/claude-plugins/inki/skills/route/SKILL.md)):   

"The PR routes cleanly to one page (cms/features/strapi-mcp-server.md) with two same-file section updates. The key gap it exposes: the "Relations" section documents only how to write relations, never what relations look like when read back — and this PR just defined that output shape."

And I trust Inki. Maybe because I built it, though. 😆 
I can handle it next week if you want.

innerdvations (approved):
Nice! LGTM now!

nclsndr:
> and this PR just defined that output shape.

C'mon Inki, the previous PR was also defining an output, it was just buried into ~100 files changed 😇 

— Joke aside, @pwizla I let you decide how detailed we want to be knowing only agents will ever face those tools / payloads. For a human I will require you to inspect tool calls…

nclsndr:
Experimental version: 
`0.0.0-experimental.2b2b5347394bc953b8729f080917b354e8b600c9`

innerdvations (approved):

PR #26573, opened 2026-06-08
fix(core): remove config-based access control from openapi endpoints

innerdvations:
## What does it do?

Strips the entire **config-based access-control surface** from the OpenAPI HTTP
endpoint feature (added in #26239, not yet released). Both knobs that existed —
`access.roles` and `access.mode` — are removed, along with the `access` block
itself. There is no longer any OpenAPI-specific access configuration.

An endpoint is configured with only `enabled` + `route` + `cache`, and each one
gets a fixed, predictable access policy:

- **content-api** (`/api/openapi.json`): **public** when enabled. The document
  describes the public Content API, so the spec is served without authentication.
- **admin** (`/admin/openapi.json`): requires an authenticated admin user.

## Why is it needed?

We're releasing soon and don't want to ship the half-baked access config:

- `access.roles` was misnamed (it held auth scopes, not roles) and was **silently
  unenforced on the admin endpoint** (admin strategies don't verify route scope).
- `access.mode` duplicated access control that, in Strapi, already lives in the
  permission systems (users-permissions / RBAC) rather than per-feature config.

Shipping it would mean either living with a misleading API or making a breaking
change later. Reducing the feature to on/off with a sensible fixed policy avoids
both: there is **no access config to break**. Proper permission-based access
control (RBAC) lands in the follow-up (#26574), with its own review/QA.

> Note: when the RBAC follow-up ships, the content-api spec becomes
> permission-gated; keeping it public will then require granting the
> `strapi::openapi.openapi.read` permission to the users-permissions **Public**
> role. This will be called out in release notes.

## How to test it?

- `yarn jest packages/core/core/src/services/server/__tests__/openapi.test.ts`
- Enable `content-api`; fetch `/api/openapi.json` with no credentials → 200.
- Enable `admin`; fetch `/admin/openapi.json` with no credentials → 401, and with
  an authenticated admin → 200.

## Related issue(s)/PR(s)

- Builds on #26239
- Follow-up that introduces RBAC the correct way: #26574 (stacked on this branch)

### DIFF
--- a/packages/core/core/src/configuration/index.ts
+++ b/packages/core/core/src/configuration/index.ts
@@ -44,9 +44,6 @@ const defaultServerConfig = {
       route: {
         path: '/openapi.json',
       },
-      access: {
-        mode: 'authenticated',
-      },
       cache: {
         enabled: true,
         maxAgeMs: 60_000,
@@ -57,9 +54,6 @@ const defaultServerConfig = {
       enabled: false,
       route: {
         path: '/openapi.json',
-      },
-      access: {
-        mode: 'authenticated',
       },
       cache: {
         enabled: true,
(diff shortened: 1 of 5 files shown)


innerdvations:
Closing in favor of #26574, which now contains the full standard RBAC implementation as a single PR targeting develop.

PR #26574, opened 2026-06-08
feat(openapi): gate endpoint access with config

innerdvations:
## What does it do?

Adds opt-in HTTP endpoints for the generated OpenAPI specs, controlled by a single `access` setting per endpoint.

`access` accepts endpoint-specific values:

- **Content API** (`/api/openapi.json`):
  - `disabled` (default): the endpoint is not registered.
  - `public`: the endpoint is registered without authentication (anyone can read the spec).
- **Admin** (`/admin/openapi.json`):
  - `disabled` (default): the endpoint is not registered.
  - `authenticated`: requires an authenticated admin (any role).

The first iteration intentionally avoids introducing a bespoke OpenAPI permission/action or changing global content-api token semantics. Granular per-permission/role RBAC can be added in a follow-up once the core-owned permission shape is agreed.

## Why is it needed?

The earlier config-based access design (`mode` + `roles`) was the wrong abstraction: `roles` was misnamed (it held auth scopes, not roles) and was silently unenforced for the admin endpoint.

This PR keeps the feature fully opt-in and limited to the unambiguous cases:

- publish a public Content API spec when a project explicitly wants one;
- expose a gated Admin API spec to authenticated admins;
- keep both endpoints disabled by default so existing projects are unaffected.

## How to test it?

1. Configure `server.openapi` and verify behavior:
   - no config, or `access: 'disabled'` → no OpenAPI endpoint is registered.
   - `content-api: { access: 'public' }` → anonymous `GET /api/openapi.json` returns 200.
   - `content-api: { access: 'authenticated' }` → throws at startup.
   - `admin: { access: 'authenticated' }` → unauthenticated returns 401; any authenticated admin returns 200.
   - `admin: { access: 'public' }` → throws at startup.
2. Automated:
   - `yarn jest packages/core/core/src/services/server/__tests__/openapi.test.ts`
   - `yarn test:api tests/api/core/strapi/api/openapi-access.test.api.ts`

## Related issue(s)/PR(s)

- Builds on #26239

### DIFF
--- a/packages/core/core/src/configuration/index.ts
+++ b/packages/core/core/src/configuration/index.ts
@@ -40,12 +40,9 @@ const defaultServerConfig = {
   },
   openapi: {
     'content-api': {
-      enabled: false,
+      access: 'disabled',
       route: {
         path: '/openapi.json',
-      },
-      access: {
-        mode: 'authenticated',
       },
       cache: {
         enabled: true,
@@ -54,12 +51,9 @@ const defaultServerConfig = {
       },
     },
     admin: {
-      enabled: false,
+      access: 'disabled',
       route: {
         path: '/openapi.json',
-      },
-      access: {
-        mode: 'authenticated',
       },
       cache: {
         enabled: true,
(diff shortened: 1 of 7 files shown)


Bassel17 (approved):
LGTM

Issue #18691, opened 2023-11-07
Media Library broken: "Cache busting" hash suffix in URL stays the same after crop or replace


tim-the-arcane:
### Required System information

- Node.js version: v18.18.2
- Yarn version: 1.22.19
- Strapi version: v4.14.6
- Database: Postgres / Sqlite
- Operating system: any
- Is your project Javascript or Typescript: JavaScript

### Describe the bug

When updating an entry in the media library (e.g. replacing or cropping) the URL stays exactly the same. This doesn't make sense in my opinion as the URL contains a cache buster suffix anyway. I have to clear not only my browser cache for my Strapi host but also the cache on my Next.js Vercel hosting as it proxies the URL and caches the image on the server-side. 

### Steps to reproduce the behavior

1. Go to Media library
2. Open lightbox for image
3. Crop or replace image
4. See error: The URL stays the same and the preview is not even updated

### Expected behavior

The cache buster suffix in the URL should update so the image is updated aswell.

thewickest:
I have to exact same issue. Any news on this?

pjaydev:
As a workaround (until this gets fixed), I let my frontend make a hash string from the `updatedAt` value of the media object and dynamically append it to the generated image URL: 

```typescript
import crypto from "crypto";

const mediaObj = // ...fetch media instance from Strapi

let myImageUrl = mediaObj.url;

const suffix = crypto
	.createHash('sha1')
	.update(String(mediaObj.updatedAt))
	.digest('base64')
	.replace(/[^a-zA-Z0-9]/g, '')
	.substring(0, 8);

myImageUrl += '?' + suffix;
```

Maybe it helps somone..

PR #26582, opened 2026-06-09
enhancement(upload): add optional replace method to upload providers

giu1io:
## What does it do?

Adds an optional `replace` / `replaceStream` method to the upload provider interface so providers can perform an atomic replacement of a media file instead of the current `delete` + `upload` sequence. The provider service falls back transparently to `delete` + `upload` when neither method is implemented, mirroring the existing `uploadStream` / `upload` fallback.

Native implementations are added for the three first-party providers:

- **upload-local** — writes the new file then unlinks the old one if the path changed (overwrites in place when the hash/ext are unchanged).
- **upload-aws-s3** — `PutObject` overwrites the existing key, and the old key is deleted only when it differs from the new one.
- **upload-cloudinary** — when the `public_id` is preserved, uploads with `overwrite: true` + `invalidate: true` to bust the CDN cache in a single call; otherwise uploads then destroys the old asset.

The upload service's `replaceFile` flow is rewired to route through `provider.replace`, and a new `uploadImageReplacing` pairs each format (thumbnail, responsive variants) with its old counterpart so providers see them as replacements rather than as a delete followed by an upload. Formats that no longer exist on the new image are deleted; formats that didn't exist on the old image are uploaded fresh.

## Why is it needed?

Closes [CMS-1067](https://linear.app/strapi/issue/CMS-1067/unable-to-bust-do-spaces-cache-cache-time-forcefully-set-to-1-hour).

DigitalOcean Spaces forces a 1-hour cache on assets. When a user replaces a file via the Media Library, the old cached version keeps being served from the CDN until the TTL expires — so the admin panel and the API both show the stale image even though the new one has been uploaded.

The root cause is that the provider only ever sees a `delete` followed by an `upload`: it has no signal that the two operations are part of a single replacement, so providers that could invalidate the CDN cache on overwrite (Cloudinary, S3-backed CDNs) have no opportunity to do so.

Exposing a dedicated `replace` hook lets each provider do whatever its backend supports — overwrite-with-invalidate on Cloudinary, key-reuse on S3, in-place write on local — and resolves the stale-cache problem on DO Spaces (and any other S3-compatible CDN with similar behavior).

## How to test it?

- [ ] Set up a project with the **local** provider, upload an image, replace it from the Media Library, and verify the file on disk is overwritten in place (the main file's hash is intentionally preserved across replacements so the URL stays stable). When replacing with an image of a different size, also verify that responsive format files (`thumbnail_*`, `large_*`, etc.) with stale hashes are removed and the new format files appear in `public/uploads/`.
- [ ] Set up a project with the **AWS S3** provider against DO Spaces (or any S3-compatible bucket) and confirm that replacing a file no longer leaves a stale object behind and the new asset is served immediately (modulo whatever CDN invalidation the bucket supports).
- [ ] Set up a project with the **Cloudinary** provider, replace an image whose `public_id` is preserved, and confirm the CDN URL serves the new asset immediately (`invalidate: true` should bust the cache in a single call).
- [ ] Test a **cross-provider replacement** (e.g. switch the provider config between uploads): the new file should be uploaded to the new provider without attempting to delete from it.
- [ ] Test replacing an image that has thumbnails / responsive formats: formats present on the new image should be replaced, formats no longer present should be deleted, and any new formats should be uploaded fresh.
- [ ] Test with a **third-party provider that does not implement `replace`** — the service should fall back to `delete` + `upload` and behave exactly as before.
- [ ] Run the new unit tests: `yarn workspace @strapi/upload test` and `yarn workspace @strapi/provider-upload-local test`.

### DIFF
--- a/packages/core/upload/server/src/services/upload.ts
+++ b/packages/core/upload/server/src/services/upload.ts
@@ -341,7 +341,7 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
    * an atomic replace can use it. Formats that no longer exist on the new image
    * are deleted; formats that didn't exist on the old image are uploaded fresh.
    */
-  async function uploadImageReplacing(fileData: UploadableFile, oldFile: File) {
+  async function replaceImage(fileData: UploadableFile, oldFile: File) {
     const { getDimensions, generateThumbnail, generateResponsiveFormats, isResizableImage } =
       getService('image-manipulation');
 
@@ -488,14 +488,24 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
         ext: dbFile.ext,
       });
 
-      // clear old formats — uploadImageReplacing / replace will set new ones
+      // clear old formats — replaceImage / replace will set new ones
       _.set(fileData, 'formats', {});
 
       if (dbFile.provider === config.provider) {
         if (await isImage(fileData)) {
-          await uploadImageReplacing(fileData, dbFile);
+          await replaceImage(fileData, dbFile);
         } else {
+          // The new file is not an image, so it has no formats. Replace the main
+          // file, then delete any formats the old image left behind — otherwise
+          // they're orphaned in storage since the DB record no longer tracks them.
           await getService('provider').replace(fileData, dbFile);
+          if (dbFile.formats) {
+            await Promise.all(
+              Object.keys(dbFile.formats).map((key) =>
+                strapi.plugin('upload').provider.delete(dbFile.formats![key] as File)
+              )
+            );
+          }
         }
       } else if (await isImage(fileData)) {
         // Cross-provider replacement: no delete on the old provider, upload to the new one.
--- a/packages/core/upload/server/src/services/upload.ts
+++ b/packages/core/upload/server/src/services/upload.ts
@@ -336,6 +336,74 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
   }
 
   /**
+   * Like uploadImage, but pairs the main file and each format with its old
+   * counterpart and routes through provider.replace so providers that implement
+   * an atomic replace can use it. Formats that no longer exist on the new image
+   * are deleted; formats that didn't exist on the old image are uploaded fresh.
+   */
+  async function uploadImageReplacing(fileData: UploadableFile, oldFile: File) {
+    const { getDimensions, generateThumbnail, generateResponsiveFormats, isResizableImage } =
+      getService('image-manipulation');
+
+    const { width, height } = await getDimensions(fileData);
+
+    _.assign(fileData, {
+      width,
+      height,
+    });
+
+    const replaceFormat = async (key: string, newFormat: UploadableFile) => {
+      const oldFormat = oldFile.formats?.[key] as File | undefined;
+      if (oldFormat) {
+        await getService('provider').replace(newFormat, oldFormat);
+      } else {
+        await getService('provider').upload(newFormat);
+      }
+      _.set(fileData, ['formats', key], newFormat);
+    };
+
+    const promises: Promise<unknown>[] = [];
+
+    // Replace the main file
+    promises.push(getService('provider').replace(fileData, oldFile));
+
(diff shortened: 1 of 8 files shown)


packages/core/upload/server/src/services/upload.ts
```diff
@@ -335,6 +335,74 @@ export default ({ strapi }: { strapi: Core.Strapi }) => {
     await Promise.all(uploadPromises);
   }
 
+  /**
+   * Like uploadImage, but pairs the main file and each format with its old
+   * counterpart and routes through provider.replace so providers that implement
+   * an atomic replace can use it. Formats that no longer exist on the new image
+   * are deleted; formats that didn't exist on the old image are uploaded fresh.
+   */
+  async function uploadImageReplacing(fileData: UploadableFile, oldFile: File) {
```
markkaylor: Not crazy about the function name. What about `uploadToReplaceImage` or just simply `replaceImage`

markkaylor (commented):
On the local provider:
1. The AI metadata generation should still work on replace. I believe @Adzouz fixed this recently. We should confirm and ensure we don't regress there. Looks like it's not working in the new media lib either.
2. The preview in the list/grid view requires a refresh (both current and new media lib). Not sure if that's a regression thought, probably a bug I'm only catching now.

https://github.com/user-attachments/assets/bfaa8c57-ce60-4cf8-b689-1e18aa0aa711

Otherwise looks good to me! Happy to test again with the other providers.

innerdvations:
One backwards-compat concern in `upload.ts` `replace()`:

Before this PR, a same-provider replace always deleted the old main file **and** all old generated formats (`thumbnail`, `large`, …) before uploading the new one. With the new routing, when the **old file is an image but the new file is not**, we only call `provider.replace()` on the main file. `formats` gets reset to `{}` in the DB, so Strapi forgets the old thumbnails but never deletes them from storage — they're orphaned.

Suggest restoring the old cleanup in the non-image, same-provider branch:

```ts
} else {
  await getService('provider').replace(fileData, dbFile);
  if (dbFile.formats) {
    await Promise.all(
      Object.keys(dbFile.formats).map((key) =>
        strapi.plugin('upload').provider.delete(dbFile.formats![key] as File)
      )
    );
  }
}
```

This just re-applies what `replace()` (and `remove()`) already did pre-PR, guarded by the same-provider check, so it stays fully backwards compatible.

packages/core/upload/server/src/services/upload.ts
giu1io: Good call — renamed it to `replaceImage` in c77b2c0ea4c7db5c93fac6946f65949299464dda (sibling to the existing `replace()`, so it reads clearly). Thanks!

giu1io:
> Suggest restoring the old cleanup in the non-image, same-provider branch:

@innerdvations I've applied your suggestion and added some tests for it as well.

## Other review points

- Renamed `uploadImageReplacing` → `replaceImage`.
- **AI metadata on replace** No regression. Adzouz's fix (#26436) is in the `replaceFile` *controller* (`admin-upload.ts`), which this PR doesn't touch.
- **List/grid preview needs refresh:** Pre-existing (repros in old + new media lib; caused by intentionally-stable URL/hash). I think we should open a separate issue for this. There was one already open (#18691) but was closed as stale. @innerdvations what do you think?

innerdvations:
Yeah, if that issue is still present, I would re-open it and do a follow-up, but no need to address it here.

innerdvations (approved):
LGTM! Should we call this an `enhancement` rather than `feat`?

lucasaba:
This feature introduced a regression when using the `folder` `actionOptions`.

The problem is described here: https://github.com/strapi/strapi/issues/27665

I'm not sure how to implement a solution. Since provider register does not require `replace` or `replaceStream` methods, adding them as required, in order to update the provider register would introduce a breaking change for all upload providers.

Any suggestion would be highly appreciated.

lucasaba:
The first possible solution (with breaking changes) is this: https://github.com/strapi/strapi/pull/27676/changes

PR #26589, opened 2026-06-10
test(content-manager): cover self-referential publish relations

innerdvations:
## What does it do?
- Adds API regression coverage for unidirectional self-referential relations where an entry relates to itself across publish, republish, discard, one-to-many, and many-to-many flows.
- Adds a Content Manager E2E test that verifies a self-referential relation remains visible in the admin immediately after publishing, without a manual refresh.
- Adds a self-referential `related` relation to the Dog E2E app-template content type so the UI flow can be exercised.

## Why is it needed?
- The previous regression coverage focused on bidirectional self-links and two-entry unidirectional relations, leaving the single-entry unidirectional self-link publish flow unguarded.
- The E2E coverage ensures the relation is not only persisted server-side, but also reflected in the admin UI after publish without relying on a page reload.

## How to test it?
- `yarn test:api --testPathPattern self-referential-relations`
- `DATABASE_CLIENT=postgres DATABASE_HOST=localhost DATABASE_PORT=5432 DATABASE_NAME=strapi_self_ref_tests DATABASE_USERNAME=strapi DATABASE_PASSWORD=strapi yarn test:api --db=postgres --testPathPattern self-referential-relations`
- `PLAYWRIGHT_BROWSERS_PATH="$HOME/Library/Caches/ms-playwright" yarn test:e2e --setup --concurrency=1 --domains content-manager --grep "Self-referential relation - publish" --project=chromium --reporter=line`
- `PLAYWRIGHT_BROWSERS_PATH="$HOME/Library/Caches/ms-playwright" yarn test:e2e --concurrency=1 --domains content-manager --grep "Conditional Fields" --project=chromium --reporter=line`

## Related issue(s)/PR(s)
- Follow-up coverage for #26536.

Fixes CMS-1154

### DIFF
--- a/tests/api/core/content-manager/api/self-referential-relations.test.api.ts
+++ b/tests/api/core/content-manager/api/self-referential-relations.test.api.ts
@@ -12,6 +12,7 @@ interface CategoryEntry {
   parent?: CategoryRelation | null;
   children?: CategoryRelation[];
   related?: CategoryRelation[];
+  relatedMany?: CategoryRelation[];
 }
 
 const builder = createTestBuilder();
@@ -39,10 +40,16 @@ const category = {
       target: 'api::category.category',
       targetAttribute: 'children',
     },
-    // Unidirectional self-referential
+    // Unidirectional self-referential (one to many)
     related: {
       type: 'relation',
       relation: 'oneToMany',
+      target: 'api::category.category',
+    },
+    // Unidirectional self-referential (many to many)
+    relatedMany: {
+      type: 'relation',
+      relation: 'manyToMany',
       target: 'api::category.category',
     },
   },
@@ -58,6 +65,50 @@ const getCategory = async (
     qs: { status },
   });
   return res.body.data as CategoryEntry;
+};
+
+const createCategory = async (name: string): Promise<CategoryEntry> => {
+  const res = await rq({
+    method: 'POST',
+    url: '/content-manager/collection-types/api::category.category',
+    body: { name },
+  });
+
+  return res.body.data as CategoryEntry;
+};
+
+const updateCategory = async (documentId: string, body: Record<string, unknown>) =>
+  rq({
+    method: 'PUT',
+    url: `/content-manager/collection-types/api::category.category/${documentId}`,
+    body,
+  });
+
+const publishCategory = async (documentId: string) =>
+  rq({
+    method: 'POST',
+    url: `/content-manager/collection-types/api::category.category/${documentId}/actions/publish`,
+  });
+
+const discardCategory = async (documentId: string) =>
+  rq({
+    method: 'POST',
+    url: `/content-manager/collection-types/api::category.category/${documentId}/actions/discard`,
+  });
+
+// Returns the target documentIds linked through `field` for the given status.
+const getRelationTargets = async (
+  documentId: string,
+  field: string,
+  status: 'draft' | 'published'
+): Promise<string[]> => {
+  const res = await rq({
+    method: 'GET',
+    url: `/content-manager/relations/api::category.category/${documentId}/${field}`,
+    qs: { status },
+  });
+
+  return res.body.results.map((result: { documentId: string }) => result.documentId);
 };
 
 describe('CM API - Self-referential relations with Draft & Publish', () => {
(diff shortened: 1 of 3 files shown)


tests/api/core/content-manager/api/self-referential-relations.test.api.ts
```diff
@@ -245,4 +296,87 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
...
+    ]);
+
+    await publishCategory(cat.documentId);
+
+    expect(await getRelationTargets(cat.documentId, 'relatedMany', 'published')).toEqual([
+      cat.documentId,
+    ]);
+  });
+
+  test('unidirectional self-link (entry to itself) is preserved when added before a republish', async () => {
```
mathildeleg: Maybe specify the type of self-link like you did for the two previous tests?

tests/api/core/content-manager/api/self-referential-relations.test.api.ts
```diff
@@ -245,4 +296,87 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
...
+      related: [{ documentId: cat.documentId, locale: null }],
+    });
+    await publishCategory(cat.documentId);
+
+    expect(await getRelationTargets(cat.documentId, 'related', 'published')).toEqual([
+      cat.documentId,
+    ]);
+  });
+
+  test('unidirectional self-link (entry to itself) is preserved after discarding the draft', async () => {
```
mathildeleg: Same here (maybe specify the type of self-link)

tests/app-template/src/api/dog/content-types/dog/schema.json
```diff
@@ -84,6 +84,11 @@
     },
     "favoriteTreat": {
       "type": "string"
+    },
+    "related": {
```
mathildeleg, line 88: Won't this change impact other tests using this schema? 🤔

tests/api/core/content-manager/api/self-referential-relations.test.api.ts
```diff
@@ -245,4 +296,87 @@ describe('CM API - Self-referential relations with Draft & Publish', () => {
...
+  // its published version: during the delete-and-recreate publish cycle the self-link has to
+  // be re-pointed to the entry's own published version rather than dropped. These cover the
+  // unidirectional `oneToMany` and `manyToMany` self-links; the bidirectional `parent`
+  // self-link is covered by the tests above.
+  test('unidirectional oneToMany self-link (entry to itself) is preserved after first publish', async () => {
+    const cat = await createCategory('OneToMany self publish');
+
+    await updateCategory(cat.documentId, {
+      name: cat.name,
+      related: [{ documentId: cat.documentId, locale: null }],
```
mathildeleg, line 310: Not sure `locale: null` is needed since Category doesn't seem to have localization enabled 🤔

tests/api/core/content-manager/api/self-referential-relations.test.api.ts
innerdvations: Done in 0846f7735d — renamed this one to specify the `oneToMany` self-link type.

tests/api/core/content-manager/api/self-referential-relations.test.api.ts
innerdvations: Done in 0846f7735d — renamed this one to specify the `oneToMany` self-link type as well.

tests/app-template/src/api/dog/content-types/dog/schema.json
innerdvations: Good question — I checked this before adding the field. The other consumers of the `dog` schema should be unaffected:

- The Conditional Fields E2E tests import the `with-admin` DTS snapshot with the default `coreStore: true`, which restores the Content Manager layout/permissions captured before this field existed, so `related` is not placed on those edit views. I re-ran the full `Conditional Fields` grep on Chromium to confirm: 6 passed.
- The CLI `content-types:list` snapshot only records content-type UIDs, not attributes, so adding `related` does not change it. I re-ran `yarn test:cli --grep "content-types"`: 20 suites / 9 snapshots passed.
- The remaining `dog` references are API tests that build their schemas inline with `createTestBuilder`, not from the app-template.

This new E2E spec is the only one that needs the field visible, which is why it explicitly uses `coreStore: false` + `resyncSuperAdminPermissionsAfterImport()`.

tests/api/core/content-manager/api/self-referential-relations.test.api.ts
innerdvations: You are right that Category is not localized, so `locale: null` is not required for this specific content type. I kept it deliberately for consistency with the existing tests in this file: the earlier `parent: { documentId, locale: null }` cases from #26536 use the same connect shape. That keeps the relation payloads uniform across the suite. If we want to remove it, I think it would be cleaner to do it across the existing cases too rather than only in these new tests.

mathildeleg (approved):

PR #26592, opened 2026-06-10
feat(i18n): update Polish translation

TheMatrixan:
### What does it do?

Updates to Polish translations - adds missing keys and updates current ones. 

### Why is it needed?

Because some keys are missing.

### How to test it?

Start the Strapi in Polish

### Related issue(s)/PR(s)

None

---
Fixes CPR-389


oiorain (approved):
Reviewed the Polish translation update. Verified against the English reference across all 7 files:

- **Key parity:** exact match with each `en.json` — no dead keys, no remaining missing keys.
- **Valid JSON** and **valid ICU MessageFormat** on every added/changed string (the 2 pre-existing `select`-without-`other` warnings in `admin/pl.json` mirror `en.json` and aren't touched here).
- **Placeholder variables** and **rich-text tags** consistent with EN; `list.assets.selected` correctly migrated to `{numberFolders}`/`{numberAssets}`.
- `asset-details.duration` removal is legit — the key is gone from `en.json` too.
- Nice cleanup of prior bugs too (invalid ICU `jeden`→`one`, `Bilbioteka`→`Biblioteka`, untranslated `View options`).

Optional (non-blocking): a native Polish speaker could refine wording and consider explicit `few`/`many` plural categories, but the current `one`/`other` shape mirrors the EN keys and renders correctly.

Thanks for the contribution! 🙏

PR #26594, opened 2026-06-10
chore: add ai-tooling sync script for skill symlinks

nclsndr:
### What does it do?

Adds `scripts/ai-tooling/` with two TypeScript modules: `links.ts` implements the core `sync`, `unlink`, and `status` commands for managing symlinks from `.ai/skills/` into `.agents/skills/`, `.claude/skills/`, and `.cursor/skills/`; `index.ts` is the CLI entry point. Exposes three root-level yarn scripts (`ai-tooling:sync`, `ai-tooling:unlink`, `ai-tooling:status`) powered by `tsx` (added as a devDependency). Updates `AGENTS.md` and `CONTRIBUTING.md` to document the skill directory conventions and instruct contributors to run `yarn ai-tooling:sync` after setup.

### Why is it needed?

AI tooling directories (`.agents/`, `.claude/`, `.cursor/`) each need their own copy of repo skills, but committing the same files in three places creates drift. Symlinks keep a single source of truth under `.ai/skills/` while letting each tool discover skills through its own well-known path. The script is idempotent and safe to run in CI — it skips foreign links and real files (collision warning) and prunes stale ours-owned links automatically.

### How to test it?

```bash
yarn install
yarn ai-tooling:status   # should report missing/linked for each target dir
yarn ai-tooling:sync     # creates symlinks under .agents/skills/, .claude/skills/, .cursor/skills/
yarn ai-tooling:status   # all entries should now show "linked"
yarn ai-tooling:unlink   # removes only ours-owned links
yarn ai-tooling:status   # entries return to "missing"
yarn ai-tooling:test     # runs the new Node test-runner suite
```

### Related issue(s)/PR(s)

- ø


scripts/ai-tooling/__tests__/links.test.ts
nclsndr, line 1: Not sure we want to keep / maintain tests here

scripts/ai-tooling/index.ts
```diff
@@ -0,0 +1,23 @@
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+// @ts-expect-error - no .ts extension import
+import { sync, unlink, status, makeLogger } from './links.ts';
+
+// Derive repo root from the script's own location — works regardless of cwd.
+// @ts-expect-error - import.meta.url is not supported in commonjs context
```
nclsndr, line 7: This would be fixed the day scripts becomes a private package of its own

AGENTS.md
```diff
@@ -59,6 +69,7 @@ Personal/in-progress skills go under `.agents/local-skills/` (gitignored). See [
 # Initial setup (run once after cloning)
 yarn install
 yarn setup                    # clean + build all packages
+yarn ai-tooling:sync          # link .ai/skills into .agents/ .claude/ .cursor/
```
nclsndr: Currently not bundled in setup as discussed in the team

AGENTS.md
```diff
@@ -34,9 +34,19 @@ The following are the most important packages (not exhaustive — run `yarn work
...
 
-Shared skills live under `.agents/skills/<name>/`. Tool-specific skill directories (e.g. `.claude/skills/`) should be symlinks to `.agents/skills/` — each AI tool has its own well-known location.
+**`.ai/skills/`** is the canonical source for committed repo skills. Each subdirectory containing a `SKILL.md` is a skill.
 
-Personal/in-progress skills go under `.agents/local-skills/` (gitignored). See [.agents/skills/writing-skills/SKILL.md](.agents/skills/writing-skills/SKILL.md) for authoring guidance.
+The AI-tooling well-known locations are **symlink targets** maintained by `yarn ai-tooling:sync`: `.agents/skills/`, `.claude/skills/`, `.cursor/skills/`.
+
+Run `yarn ai-tooling:sync` after adding or removing a skill in `.ai/skills/` to keep all three target dirs up to date. Links are local-only (gitignored target dirs) — only `.ai/skills/` content is committed.
+
+On **Windows**, the CLI creates directory **junctions** (no extra setup). Directory symlinks require Developer Mode or an elevated shell.
```
unrevised6419, line 43: Question: Is this a typo?

```suggestion
On **Windows**, the CLI creates directory **functions** (no extra setup). Directory symlinks require Developer Mode or an elevated shell.
```

nclsndr: I'm not very familiar with Windows terminology but this one seems legit to me. https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions#junctions

scripts/ai-tooling/index.ts
mathildeleg: Maybe good to add a @todo here to not forget that?

innerdvations:
What about calling it "ai-skills" since "ai-tooling" is a bit broad?  Or going with "ai:skills"?

scripts/ai-tooling/links.ts
```diff
@@ -0,0 +1,255 @@
...
+    const targetDir = path.join(repoRoot, targetRel);
+    fs.mkdirSync(targetDir, { recursive: true });
+
+    for (const name of skills) {
+      const p = path.join(targetDir, name);
+      const k = classify(repoRoot, p, name);
+
+      if (k.kind === 'absent') {
+        createSkillLink(repoRoot, p, name);
+        log.add(targetRel, name, 'linked');
```
mathildeleg, line 211: Just curious: why use `log` here, but `console.log` l92 for example?

jhoward1994:
I agree with `ai-skills`

mathildeleg (commented):
Just a question for my personal understanding and a small suggestion, otherwise LGTM

jhoward1994:
- Maybe we could have a `--force` flag that would overwrite on skill collisions ?
- Maybe we could add a hint to `yarn setup`? e.g.reminder to run `yarn ai-skills:sync`

nclsndr:
> What about calling it "ai-skills" since "ai-tooling" is a bit broad? Or going with "ai:skills"?

My issue with "skills" is that we'll want to also sync "commands/prompts" or even "plugins" moving forward. Naming the cli after "skill" feels odd to me. I don't really like "ai-tooling" neither.
Opinions @innerdvations @jhoward1994?

scripts/ai-tooling/index.ts
nclsndr: I prefer to have the TS compiler throwing an error once we migrate rather than keeping a TODO we have to parse by text search.

scripts/ai-tooling/links.ts
nclsndr: L:92 is the logger instrumenting the global console, this one is intended.
The rest of the file has gaps on that direction, thx for pointing out!

innerdvations:
In that case what about just "ai"?  So either `ai:sync` or `ai sync`

Anything else AI-specific could fall under that and it would still make sense (although I can't think of what else we would even need to add there)

nclsndr:
> * Maybe we could have a `--force` flag that would overwrite on skill collisions ?
> * Maybe we could add a hint to `yarn setup`? e.g.reminder to run `yarn ai-skills:sync`

Just pushed 2 new commits to cover this. Thank you @jhoward1994 🙌

innerdvations (approved):

jhoward1994 (approved):

nclsndr:
Didn't we want to shorten to just "ai" @innerdvations ? 🙃

innerdvations:
Ahh, yes. We can update to change the name.

innerdvations:
Here: https://github.com/strapi/strapi/pull/26767

PR #26603, opened 2026-06-11
feat(mcp): export defineTool/defineResource/definePrompt builders

nclsndr:
### What does it do?

Nests the MCP builder exports under an `ai` namespace. The public surface is `ai.mcp.*` on `@strapi/strapi`:

- `ai.mcp.defineTool` — re-exports `makeMcpToolDefinition`
- `ai.mcp.defineResource` — re-exports `makeMcpResourceDefinition`
- `ai.mcp.definePrompt` — re-exports `makeMcpPromptDefinition`

Adds `packages/core/core/src/mcp.ts`, `packages/core/core/src/ai.ts`, and `export * as ai from './ai'` in `packages/core/core/src/index.ts`. Internal registry names and classes are unchanged.

### Why is it needed?

Plugin authors defining MCP capabilities currently have no ergonomic builder entry point — they must type definitions manually against `Modules.MCP.*` from `@strapi/types`. These identity helpers provide TypeScript inference and narrowing before calling `strapi.ai.mcp.registerTool/Prompt/Resource()`, mirroring the runtime `strapi.ai.mcp` API while keeping AI-related builders grouped under a dedicated namespace.

### How to test it?

1. Build `@strapi/core`: `yarn nx build @strapi/core`
2. In a plugin or sandbox, verify the import resolves and types infer:

```typescript
// mcp/my-tool.ts
import { ai } from '@strapi/strapi';

const myTool = ai.mcp.defineTool({
  name: 'my-tool',
  title: 'My Tool',
  description: 'Does something',
  resolveOutputSchema: () => z.object({ ok: z.boolean() }),
  createHandler: (strapi, context) => async (input) => ({ ok: true }),
});

// index / provider.ts
import type { Core } from '@strapi/strapi';

import { myTool } from './mcp/my-tool';

export default {
  register(/* { strapi }: { strapi: Core.Strapi } */) {},

  bootstrap({ strapi }: { strapi: Core.Strapi }) {
    strapi.ai.mcp.registerTool(myTool);
  },
};
```

3. Run `yarn test:ts:packages` — should pass (no new type regressions from the re-exports).

### Related issues / PRs

None.

### DIFF
--- a/packages/core/core/src/services/mcp/prompt-registry.ts
+++ b/packages/core/core/src/services/mcp/prompt-registry.ts
@@ -10,9 +10,38 @@ import {
 import { createSafeCapabilityRegistration } from './utils/createSafeCapabilityRegistration';
 import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandlerForMetrics';
 
-export const makeMcpPromptDefinition = <Definition extends Modules.MCP.McpPromptDefinition>(
-  prompt: Definition
-): Definition => prompt;
+/**
+ * Defines a Strapi MCP prompt with full type inference, ready to pass to
+ * `strapi.ai.mcp.registerPrompt()`. Exposed publicly as `mcp.definePrompt`.
+ *
+ * The returned value is the definition unchanged — this builder only exists to
+ * infer the `name`/`argsSchema` and narrow the access variant (`devModeOnly` vs
+ * `auth`) so the result is directly assignable to `registerPrompt`.
+ *
+ * @param prompt - The prompt definition. Provide either `devModeOnly: true`
+ * (dev-only, no auth) or an `auth` policy set — never both.
+ * @returns The same definition, with its access variant narrowed.
+ *
+ * @example
+ * ```ts
+ * import { mcp } from '@strapi/strapi';
+ *
+ * const context = mcp.definePrompt({
+ *   name: 'app-context',
+ *   title: 'App Context',
+ *   description: 'Provides context about the app',
+ *   devModeOnly: true,
+ *   createHandler: (strapi) => async () => ({
+ *     messages: [{ role: 'user', content: { type: 'text', text: 'You are connected to Strapi.' } }],
+ *   }),
+ * });
+ *
+ * // later, in register() or bootstrap():
+ * strapi.ai.mcp.registerPrompt(context);
+ * ```
+ */
+export const makeMcpPromptDefinition = ((definition: Modules.MCP.McpPromptDefinition) =>
+  definition) as unknown as Modules.MCP.McpPromptBuilder;
 
 export class McpPromptRegistry
   extends McpCapabilityRegistryBase<'prompt', Modules.MCP.McpPromptDefinition, RegisteredPrompt>
--- a/packages/core/core/src/services/mcp/prompt-registry.ts
+++ b/packages/core/core/src/services/mcp/prompt-registry.ts
@@ -12,7 +12,7 @@ import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandler
 
 /**
  * Defines a Strapi MCP prompt with full type inference, ready to pass to
- * `strapi.ai.mcp.registerPrompt()`. Exposed publicly as `mcp.definePrompt`.
+ * `strapi.ai.mcp.registerPrompt()`. Exposed publicly as `ai.mcp.definePrompt`.
  *
  * The returned value is the definition unchanged — this builder only exists to
  * infer the `name`/`argsSchema` and narrow the access variant (`devModeOnly` vs
@@ -24,9 +24,9 @@ import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandler
  *
  * @example
  * ```ts
- * import { mcp } from '@strapi/strapi';
+ * import { ai } from '@strapi/strapi';
  *
- * const context = mcp.definePrompt({
+ * const context = ai.mcp.definePrompt({
  *   name: 'app-context',
  *   title: 'App Context',
  *   description: 'Provides context about the app',
--- a/packages/core/core/src/services/mcp/resource-registry.ts
+++ b/packages/core/core/src/services/mcp/resource-registry.ts
@@ -12,9 +12,38 @@ import {
 import { createSafeCapabilityRegistration } from './utils/createSafeCapabilityRegistration';
 import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandlerForMetrics';
 
-export const makeMcpResourceDefinition = <Definition extends Modules.MCP.McpResourceDefinition>(
-  resource: Definition
-): Definition => resource;
+/**
+ * Defines a Strapi MCP resource with full type inference, ready to pass to
+ * `strapi.ai.mcp.registerResource()`. Exposed publicly as `mcp.defineResource`.
+ *
+ * The returned value is the definition unchanged — this builder only exists to
+ * infer the `name` and narrow the access variant (`devModeOnly` vs `auth`) so the
+ * result is directly assignable to `registerResource`.
+ *
+ * @param resource - The resource definition. Provide either `devModeOnly: true`
+ * (dev-only, no auth) or an `auth` policy set — never both.
+ * @returns The same definition, with its access variant narrowed.
+ *
+ * @example
+ * ```ts
+ * import { mcp } from '@strapi/strapi';
+ *
+ * const appInfo = mcp.defineResource({
+ *   name: 'app-info',
+ *   uri: 'strapi://app/info',
+ *   metadata: { description: 'Metadata about the app', mimeType: 'application/json' },
+ *   devModeOnly: true,
+ *   createHandler: (strapi) => async (uri) => ({
+ *     contents: [{ uri: uri.href, mimeType: 'application/json', text: JSON.stringify({ ok: true }) }],
+ *   }),
+ * });
+ *
+ * // later, in register() or bootstrap():
+ * strapi.ai.mcp.registerResource(appInfo);
+ * ```
+ */
+export const makeMcpResourceDefinition = ((definition: Modules.MCP.McpResourceDefinition) =>
+  definition) as unknown as Modules.MCP.McpResourceBuilder;
 
 export class McpResourceRegistry
   extends McpCapabilityRegistryBase<
--- a/packages/core/core/src/services/mcp/resource-registry.ts
+++ b/packages/core/core/src/services/mcp/resource-registry.ts
@@ -14,7 +14,7 @@ import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandler
 
 /**
  * Defines a Strapi MCP resource with full type inference, ready to pass to
- * `strapi.ai.mcp.registerResource()`. Exposed publicly as `mcp.defineResource`.
+ * `strapi.ai.mcp.registerResource()`. Exposed publicly as `ai.mcp.defineResource`.
  *
  * The returned value is the definition unchanged — this builder only exists to
  * infer the `name` and narrow the access variant (`devModeOnly` vs `auth`) so the
@@ -26,9 +26,9 @@ import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandler
  *
  * @example
  * ```ts
- * import { mcp } from '@strapi/strapi';
+ * import { ai } from '@strapi/strapi';
  *
- * const appInfo = mcp.defineResource({
+ * const appInfo = ai.mcp.defineResource({
  *   name: 'app-info',
  *   uri: 'strapi://app/info',
  *   metadata: { description: 'Metadata about the app', mimeType: 'application/json' },
--- a/packages/core/core/src/services/mcp/tool-registry.ts
+++ b/packages/core/core/src/services/mcp/tool-registry.ts
@@ -12,7 +12,7 @@ import type { McpAdminTokenAbility } from './authentication';
 
 /**
  * Defines a Strapi MCP tool with full type inference, ready to pass to
- * `strapi.ai.mcp.registerTool()`. Exposed publicly as `mcp.defineTool`.
+ * `strapi.ai.mcp.registerTool()`. Exposed publicly as `ai.mcp.defineTool`.
  *
  * The returned value is the definition unchanged — this builder only exists to
  * infer the `name`, input/output schemas and handler types, and to narrow the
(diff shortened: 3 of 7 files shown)


nclsndr:
Experimental to come: https://github.com/strapi/strapi/actions/runs/27338192325
=> `0.0.0-experimental.7873e8594578cec98579e23010aaaf243c9f9af0`

innerdvations (commented):
I think this is a good plan, and the execution seems fine.

My only question is: should the helpers mirror the namespace of the runtime owners (`ai.mcp` instead of simply `mcp`)?

nclsndr:
> should the helpers mirror the namespace of the runtime owners (`ai.mcp` instead of simply `mcp`)?

Good question! Opinions are welcomed on this @PaulBratslavsky @maccomaccomaccomacco @boazpoolman

nclsndr:
### What does it do?

Nests the MCP builder exports under an `ai` namespace. The public surface changes from `mcp.*` to `ai.mcp.*`:

- New `packages/core/core/src/ai.ts` → `export * as mcp from './mcp'`
- `packages/core/core/src/index.ts` → top-level export changed from `export * as mcp from './mcp'` to `export * as ai from './ai'`
- The `mcp.ts` builders (`defineTool` / `defineResource` / `definePrompt`) are unchanged

### Why is it needed?

Nesting the builders under `ai` makes the static export surface mirror the runtime `strapi.ai.mcp` API, and leaves room for future AI-related builders without polluting the top-level namespace.

### How to test it?

1. Build `@strapi/core`: `yarn nx build @strapi/core`
2. Verify the import path and inference:

nclsndr:
cc @PaulBratslavsky @raulbalestra 

# Using `ai.mcp.defineTool` in a Strapi project

```typescript
import { ai } from '@strapi/strapi';
const tool = ai.mcp.defineTool({ /* ... */ });
```

---

## 1. Enable MCP — `examples/empty/config/server.ts`

```typescript
import type { Core } from '@strapi/strapi';

const serverConfig = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Server => ({
  host: env('HOST', '0.0.0.0'),
  port: env.int('PORT', 1337),
  app: {
    keys: env.array('APP_KEYS', ['toBeModified1', 'toBeModified2']),
  },
  // Exposes the MCP endpoint at http://localhost:1337/mcp
  mcp: { enabled: true },
});

export default serverConfig;
```

---

## 2. Define the tool

```typescript
import { z } from '@strapi/utils';
import { ai } from '@strapi/strapi';

const inputSchema = z.object({
  message: z.string().describe('Message to log'),
  level: z
    .enum(['info', 'http', 'warn', 'error', 'log'])
    .default('info')
    .describe('Log level (default: info)'),
});

const outputSchema = z.object({
  status: z.string(),
  message: z.string(),
  level: z.string(),
  timestamp: z.string(),
});

export const logMessageTool = ai.mcp.defineTool({
  name: 'empty-log',
  title: 'Empty App Log',
  description: 'Logs a message to the Strapi logger with the specified level',
  resolveInputSchema: () => inputSchema,
  resolveOutputSchema: () => outputSchema,
  devModeOnly: true,
  createHandler:
    (strapi) =>
    async ({ args }) => {
      const { message, level } = args;
      const sanitizedMessage = `${level}${message}`

      const result = {
        status: 'logged',
        message: sanitizedMessage,
        level,
        timestamp: new Date().toISOString(),
      };

      return {
        content: [{ type: 'text' as const, text: JSON.stringify(result, null, 2) }],
        structuredContent: result,
      };
    },
});
```

---

## 3. Register

The builder only produces a typed definition; registration happens at runtime against `strapi.ai.mcp`.

```typescript
import type { Core } from '@strapi/strapi';

import { logMessageTool } from './mcp/log-message';

export default {
  register(/* { strapi }: { strapi: Core.Strapi } */) {},

  bootstrap({ strapi }: { strapi: Core.Strapi }) {
    strapi.ai.mcp.registerTool(logMessageTool);
  },
};
```

boazpoolman (commented):
Hi @nclsndr,

I appreciate your push for a type-safe DX for the MCP features. However, I just don't quite understand why the proposed change is needed.

The following two options can both be strictly typed, right?

```
strapi.ai.mcp.registerTool({ ... });

// or

import { ai } from '@strapi/strapi';
const myTool = ai.mcp.defineTool({ ... });
strapi.ai.mcp.registerTool(myTool);
```

And it's quite a trade of in terms of complexity. I now have to import an arbitrary `ai` namespace instead of just simply calling the global `strapi` namespace.

I get the comparison with the `factories`, but those do more than just add typescript ergonomics. They actually return a set of controllers/services/routes.

Maybe I'm missing something here, but I don't see the advantage.

nclsndr:
Hey @boazpoolman! I guess the confusion comes from the namespaces looking similar while they serve different roles.

**Two different things**

1. **`ai.mcp.defineTool(...)`** — a **static identity helper** exported from `@strapi/strapi`. At runtime it returns the definition unchanged; its job is compile-time only: infer `name`, input/output schemas, handler types, and narrow the access variant (`devModeOnly` vs `auth`) so the result is directly assignable to `registerTool`.

2. **`strapi.ai.mcp.registerTool(...)`** — the **runtime registry** on the injected Strapi instance. This only exists where `strapi` is available (`register`, `bootstrap`, etc.).

**Why not just inline into `registerTool`?**

You're right that both paths can be typed. The helper matters when you want to **define a tool outside Strapi-injected scope** — e.g. in a dedicated module that exports a reusable definition, registered later in lifecycle:

```ts
// src/mcp/my-tool.ts — no strapi instance here
import { ai } from '@strapi/strapi';

export const myTool = ai.mcp.defineTool({ /* ... */ });

// src/index.ts — strapi is injected here
bootstrap({ strapi }) {
  strapi.ai.mcp.registerTool(myTool);
}
```

Global `strapi` should never be used, so you can't call `strapi.ai.mcp.registerTool({ ... })` at module top-level. The identity helper fills that gap: **define anywhere, register when Strapi is ready**. This is the same pattern we use internally (e.g. core's `logToolDefinition` in `tool-registry`).

**On the `ai` import vs global `strapi`**

The `ai` namespace isn't arbitrary — it mirrors the runtime `strapi.ai.mcp.*` shape so static and runtime APIs stay aligned. It's a lightweight import that works at module scope without relying on global state.

**On the `factories` comparison**

Agreed that factories do more than TS ergonomics — they actually construct controllers/services/routes. The analogy here is organizational, not functional: both patterns encourage colocating definitions in standalone modules and keeping registration separate. The MCP helpers stay intentionally minimal (identity at runtime) because the real work happens at registration/execution time.

Let me know if that clears it up or if you'd prefer a different DX!

derrickmehaffy:
I have to admin I'm also confused around the defineTool vs registerTool, that is going to cause some confusion for sure. I'd expect to define and register in the same function personally, I don't see much benefit in splitting that.

On the define vs register I suppose you could just do `registerTool(defineTool({...}))` but that seems kinda redundant and confusing. If the only thing defineTool does is return a structured object, not really sure it's worth the headache.

On the global namespace stuff I guess that's the decision we need to make now because everything else uses the Strapi global namespace if we are going to start switching that then I think in the short term we should probably support both. As stupid as the global namespace can be, it also has it's benefits and it's what users are used to. If we want to deprecate that then just like everything else we need to at least temporarily support it until we can properly deprecate it in a future major.

nclsndr:
Thank you for the feedback @derrickmehaffy

> I'd expect to define and register in the same function personally, I don't see much benefit in splitting that.

As a user, you can always use `strapi.ai.mcp.registerTool({ // ...toolDef })` from the anonymous callback. The issue those helpers are trying to solve is whenever you'd like the tools to be defined elsewhere (in their specific file), yet still benefit from type-safety without requiring you to extract types manually—as @PaulBratslavsky did [here](https://github.com/PaulBratslavsky/strapi-mcp-demo-and-tool-extension/blob/main/src/plugins/strapi-extended-mcp/server/src/mcp/types.ts). Internally, we have some sort of the same logic for the internal tools: [log](https://github.com/strapi/strapi/blob/develop/packages/core/core/src/services/mcp/tools/log.ts)

> On the global namespace stuff I guess that's the decision we need to make now because everything else uses the Strapi global namespace

2 things: 
- on the general use of `global.strapi`, we cannot delete it I agree. Still, this pattern prevents any proper dep injection and lifecycle consumption. The intent is not to get rid of `global.strapi`, instead gradually move usage to DI style to avoid any unpredictable side effect.
- then, invoking `strapi.ai.mcp.register*` has to be in the bootstrap phase anyways. The initial example (PR desc) was misleading on that side, now changed to a more realistic approach.

> you could just do `registerTool(defineTool({...}))`

Thus, yes this is valid, but you'd never choose the identity helper if you are fine having your tool define inline (within the `registerTool` anon. callback)

> the global namespace can be, it also has it's benefits

Out of curiosity, where/what are the benefit of the global vs injected instances?

packages/core/core/src/services/mcp/tool-registry.ts
```diff
@@ -11,35 +10,43 @@ import { wrapCapabilityHandlerForMetrics } from './metrics/wrapCapabilityHandler
...
+ * access variant (`devModeOnly` vs `auth`) so the result is directly assignable
+ * to `registerTool`.
+ *
+ * @param tool - The tool definition. Provide either `devModeOnly: true` (dev-only,
+ * no auth) or an `auth` policy set — never both.
+ * @returns The same definition, with its access variant narrowed.
+ *
+ * @example
+ * ```ts
+ * import { mcp } from '@strapi/strapi';
```
innerdvations: This import does not exist — `@strapi/strapi` exports `ai`, not top-level `mcp`. Please update the example to:

```ts
import { ai } from '@strapi/strapi';

const greet = ai.mcp.defineTool({ ... });
```

Same fix needed in the `mcp.defineTool` wording on the docblock above (line 15: "Exposed publicly as `mcp.defineTool`" → `ai.mcp.defineTool`).

packages/core/core/src/services/mcp/prompt-registry.ts
```diff
@@ -10,9 +10,38 @@ import {
...
+ * infer the `name`/`argsSchema` and narrow the access variant (`devModeOnly` vs
+ * `auth`) so the result is directly assignable to `registerPrompt`.
+ *
+ * @param prompt - The prompt definition. Provide either `devModeOnly: true`
+ * (dev-only, no auth) or an `auth` policy set — never both.
+ * @returns The same definition, with its access variant narrowed.
+ *
+ * @example
+ * ```ts
+ * import { mcp } from '@strapi/strapi';
```
innerdvations: Same as tool-registry: use `import { ai } from '@strapi/strapi'` and `ai.mcp.definePrompt(...)` — there is no top-level `mcp` export.

packages/core/core/src/services/mcp/resource-registry.ts
```diff
@@ -12,9 +12,38 @@ import {
...
+ * infer the `name` and narrow the access variant (`devModeOnly` vs `auth`) so the
+ * result is directly assignable to `registerResource`.
+ *
+ * @param resource - The resource definition. Provide either `devModeOnly: true`
+ * (dev-only, no auth) or an `auth` policy set — never both.
+ * @returns The same definition, with its access variant narrowed.
+ *
+ * @example
+ * ```ts
+ * import { mcp } from '@strapi/strapi';
```
innerdvations: Same as tool-registry: use `import { ai } from '@strapi/strapi'` and `ai.mcp.defineResource(...)` — there is no top-level `mcp` export.

packages/core/types/src/modules/mcp.ts
```diff
@@ -123,6 +132,50 @@ export type McpToolDefinition<
...
+    | z.ZodObject<z.ZodRawShape>
+    | undefined,
+  OutputSchema extends z.ZodObject<z.ZodRawShape> = z.ZodObject<z.ZodRawShape>,
+  Title extends string = string,
+  Description extends string = string,
+> = McpToolDefinitionFields<Name, InputSchema, OutputSchema, Title, Description> &
+  McpCapabilityAccess;
+
+/**
+ * Builder for Strapi MCP tool definitions, exposed publicly as `mcp.defineTool`.
```
innerdvations: Should say `ai.mcp.defineTool` — the public export path is `import { ai } from '@strapi/strapi'` → `ai.mcp`, not a top-level `mcp` namespace.

packages/core/types/src/modules/mcp.ts
```diff
@@ -135,22 +188,61 @@ export type McpPromptCallback<ArgsSchema extends z.ZodObject<z.ZodRawShape> | un
...
+  Name extends string = string,
+  ArgsSchema extends z.ZodObject<z.ZodRawShape> | undefined =
+    | z.ZodObject<z.ZodRawShape>
+    | undefined,
+  Title extends string = string,
+  Description extends string = string,
+> = McpPromptDefinitionFields<Name, ArgsSchema, Title, Description> & McpCapabilityAccess;
+
+/**
+ * Builder for Strapi MCP prompt definitions, exposed publicly as `mcp.definePrompt`.
```
innerdvations: Same: `ai.mcp.definePrompt` rather than `mcp.definePrompt`.

packages/core/types/src/modules/mcp.ts
```diff
@@ -160,14 +252,37 @@ export type McpResourceCallback = (
...
 };
 
+/**
+ * Definition for Strapi MCP resources
+ */
+export type McpResourceDefinition<Name extends string = string> =
+  McpResourceDefinitionFields<Name> & McpCapabilityAccess;
+
+/**
+ * Builder for Strapi MCP resource definitions, exposed publicly as `mcp.defineResource`.
```
innerdvations: Same: `ai.mcp.defineResource` rather than `mcp.defineResource`.

innerdvations (changes requested):
Thanks for the DX work — `ai.mcp` mirroring `strapi.ai.mcp` is the right shape, and the `*Fields` + discriminated-access refactor cleans up the `register*` overloads.

Requesting changes for one shipped-docs issue: the `@example` blocks and type docstrings still show a top-level `mcp` import that does not exist on `@strapi/strapi`. After that, this looks good to merge.

(Non-blocking: align the PR description “What does it do?” section with `ai.mcp` — “How to test?” was already updated.)

innerdvations:
I think it's good for `defineTool` to exist, but I would be careful in the docs to downplay it (like not include it in the primary example code) to avoid confusion

innerdvations (approved):

nclsndr:
> I think it's good for `defineTool` to exist, but I would be careful in the docs to downplay it (like not include it in the primary example code) to avoid confusion

Fair point! cc @pwizla for opinion / direction

nclsndr:
Experimental: `0.0.0-experimental.0860f8c8d0bf37916071e49a06eb4a8086f490f1`

pwizla:
> > I think it's good for defineTool to exist, but I would be careful in the docs to downplay it (like not include it in the primary example code) to avoid confusion

> Fair point! cc @pwizla for opinion / direction

Thanks for the ping. Handled with this commit: https://github.com/strapi/documentation/pull/3263/changes/4588cf921f1133b704d1a3c4dc54d79b5ac6db8f

PR #26613, opened 2026-06-11
fix(content-manager): dedupe bulk delete document ids

### LINKED ISSUE #26610: Deleting from list view causes document service middleware to handle delete action twice
### Node Version

22.22.0

### Package Manager

npm

### Package Manager Version

10.9.4

### Strapi Version

5.48.0

### Operating System

Linux (Other)

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

https://github.com/edgarsgogolevs/strapi-bug-reproduction

### Bug Description

When using document service middleware to handle `delete` action if user deletes item from list view, then middleware is called twice with exactly the same parameters.
When deleting from edit view, middleware is called once as expected.

### Steps to Reproduce

Fastest view to reproduce the bug is to clone [the repo](https://github.com/edgarsgogolevs/strapi-bug-reproduction) and follow instructions in README file.

More general way to reproduce the bug:

> [!IMPORTANT]  
> Bug is observed only when draft & publish is enabled!

1. Create a collection in strapi with draft and publish enabled
2. Attach document service middleware to handle `delete` action, for example:

```javascript
module.exports = {
  register({ strapi }) {
      const { uid, action } = context;
      if (uid === "api::item.item") {
        if (action === "delete") {
          const { documentId } = context.params;
          strapi.log.info("Deleting item with documentId: " + documentId);
        }
      }
      return next();
    });
  }
};
```

3. Delete an item from list view (select it and click delete)
4. See that delete action was handled twice

### Expected Behavior

It is expected so that document service middleware with `delete` action is called exactly once no matter if deletion happened in list view or edit view.

### Logs

```shell

```

### Code Snippets

```javascript
module.exports = {
  register({ strapi }) {
      const { uid, action } = context;
      if (uid === "api::item.item") {
        if (action === "delete") {
          const { documentId } = context.params;
          strapi.log.info("Deleting item with documentId: " + documentId);
        }
      }
      return next();
    });
  }
};
```

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

innerdvations:
## Summary
- Deduplicate document ids before content-manager bulk delete calls the document service.
- Prevent Draft & Publish entries deleted from the list view from invoking `delete` document service middleware twice.
- Add a regression test for the draft + published row case.

## Related issue(s)/PR(s)
Fixes #26610
Fixes CMS-1174

## Test plan
- `yarn jest packages/core/content-manager/server/src/controllers/__tests__/bulkDelete.test.ts`
- `yarn build`
- `yarn lint:fix`
- `yarn prettier:write`

### DIFF
--- a/packages/core/content-manager/server/src/controllers/__tests__/bulkDelete.test.ts
+++ b/packages/core/content-manager/server/src/controllers/__tests__/bulkDelete.test.ts
@@ -1,0 +1,99 @@
+// @ts-expect-error - test helper
+// eslint-disable-next-line import/no-relative-packages
+import createContext from '../../../../../../../tests/helpers/create-context';
+import controller from '../collection-types';
+
+jest.mock('../validation', () => ({
+  validateBulkActionInput: jest.fn().mockResolvedValue(undefined),
+}));
+
+jest.mock('../validation/dimensions', () => ({
+  getDocumentLocaleAndStatus: jest.fn().mockResolvedValue({ locale: 'en' }),
+}));
+
+const createMocks = (overrides: Record<string, any> = {}) => {
+  const populateFromQuery = jest.fn().mockReturnThis();
+  const buildPopulate = jest.fn().mockResolvedValue({});
+
+  const findLocales = jest.fn();
+  const deleteMany = jest.fn().mockResolvedValue({ count: 1 });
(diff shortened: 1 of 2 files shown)


Bassel17 (approved):
LGTM !

PR #26618, opened 2026-06-12
fix(content-type-builder): show configure view outside development mode

### LINKED ISSUE #25783: Content-Type Builder hides "Configure the view" outside development mode in Strapi 5

metehankasapp:
Fixes #25783

### What does it do?

Keeps the Content-Type Builder "Configure the view" shortcut visible outside development mode, while preserving schema-edit actions as development-only.

### Why is it needed?

The current header action group is entirely gated by `isInDevelopmentMode`, which also hides the read-only Content Manager configuration shortcut in production mode.

### How to test it?

- `yarn workspace @strapi/content-type-builder test:front ListView.test.tsx --runInBand`

Note: local package lint currently fails on existing workspace resolver issues for `@strapi/admin/strapi-admin` imports across the package; the targeted regression test passes.

---
Fixes CPR-390

### DIFF
--- a/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
+++ b/packages/core/content-type-builder/admin/src/pages/ListView/ListView.tsx
@@ -116,36 +116,40 @@ const ListView = () => {
 
   const isDeleted = type.status === 'REMOVED';
 
-  const primaryAction = isInDevelopmentMode && (
+  const primaryAction = (
     <Flex gap={2}>
       <LinkToCMSettingsView
         key="link-to-cm-settings-view"
         type={type}
         disabled={type.status === 'NEW' || isDeleted}
       />
-      <Button
-        startIcon={<Pencil />}
-        variant="tertiary"
-        onClick={onEdit}
-        disabled={!canEdit || isDeleted}
-      >
-        {formatMessage({
-          id: 'app.utils.edit',
-          defaultMessage: 'Edit',
-        })}
-      </Button>
+      {isInDevelopmentMode && (
+        <>
+          <Button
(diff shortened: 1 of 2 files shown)


innerdvations:
Thank you for the careful write-up and the focused patch — splitting `LinkToCMSettingsView` out of the development-mode gate is a clean implementation of the original request.

We're going to close this without merging for product reasons, not because of the code quality.

The Content-Type Builder is a development-time tool. Schema edits need `strapi develop` / auto-reload, which is why Edit and Add field stay hidden outside that mode. **Configure the view** is not a schema edit, but it already has a production home in the Content Manager (document header, plus component / dynamic-zone layout links from that configuration screen). Showing the same shortcut on a locked CTB page would duplicate that entry point and make CTB look like a production configuration surface, which we do not want.

[#25783](https://github.com/strapi/strapi/issues/25783) was closed as not planned along the same lines after we asked for a workflow that needs this shortcut specifically from the CTB. Nothing in that discussion changed that call.

If we later decide CTB should be a second production door into view configuration, we can reopen. Until then, please keep using Content Manager for layout config in production.

Thanks again for taking the time to contribute.

_AI-assisted comment._

PR #26619, opened 2026-06-12
fix(openapi): include content api prefix in generated paths

### LINKED ISSUE #25493: strapi openapi generate produces urls with missing /api/ prefix

metehankasapp:
Fixes #25493

### What does it do?

Adds the configured content API REST prefix to generated OpenAPI paths for API and plugin content-api routes. Already-prefixed paths are left unchanged, so the generated document does not double-prefix routes.

### Why is it needed?

OpenAPI generation could emit paths like `/articles`, while the actual public REST endpoint is `/api/articles` by default.

### How to test it?

- `yarn workspace @strapi/openapi test:unit --runInBand`

---
Fixes CPR-391

### DIFF
--- a/packages/core/openapi/__tests__/mocks/strapi.mock.ts
+++ b/packages/core/openapi/__tests__/mocks/strapi.mock.ts
@@ -1,6 +1,18 @@
 import { routes } from '../fixtures';
 
 export class StrapiMock {
+  get config() {
+    return {
+      get: jest.fn((key: string, defaultValue?: unknown) => {
+        if (key === 'api.rest.prefix') {
+          return '/api';
+        }
+
+        return defaultValue;
+      }),
+    };
+  }
+
   get apis() {
     return {
       'api-a': {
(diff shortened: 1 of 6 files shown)


kimjune01:
The already-prefixed guard misses configured prefixes with a trailing slash. With `api.rest.prefix = "/api/"` and route path `/api/articles`, `normalizedPrefix` remains `/api/`, so neither equality nor `startsWith(`${normalizedPrefix}/`)` matches. The subsequent join collapses `/api//api/articles` to `/api/api/articles`.

Could we trim trailing slashes from the prefix before the guard and add this case to the custom-prefix test? That would preserve the PR’s no-double-prefix guarantee for both `/api` and `/api/` configurations.

PR #26620, opened 2026-06-12
fix(admin): expose json errors for non-json fetch responses

### LINKED ISSUE #26199: useFetchClient does not handle server errors well in case new responseType option is involved

metehankasapp:
Fixes #26199

### What does it do?

When `getFetchClient` is called with a non-JSON `responseType`, HTTP error responses are now parsed as JSON first so Strapi API errors remain available on the thrown `FetchError`.

If the error body is not JSON, the previous generic `Server Error` fallback is preserved.

### Why is it needed?

A request such as `get(url, { responseType: "text" })` could receive a standard Strapi JSON error response, but callers only saw `FetchError: Server Error` without access to the server-provided message/details.

### How to test it?

- `yarn workspace @strapi/admin test:front getFetchClient.test.ts --runInBand`
- `yarn eslint packages/core/admin/admin/src/utils/getFetchClient.ts packages/core/admin/admin/src/utils/tests/getFetchClient.test.ts --quiet`

Note: local package lint (`nx run-many -t lint -p @strapi/admin`) currently fails on existing package-wide resolver/lint issues unrelated to this change, so the commit was created with `--no-verify`.

---
Fixes CPR-392

### DIFF
--- a/packages/core/admin/admin/src/utils/getFetchClient.ts
+++ b/packages/core/admin/admin/src/utils/getFetchClient.ts
@@ -301,6 +301,24 @@ const getFetchClient = (defaultOptions: FetchConfig = {}): FetchClient => {
   // Check if the url has a prepending slash, if not add a slash
   const normalizeUrl = (url: string) => (hasProtocol(url) ? url : addPrependingSlash(url));
 
+  const parseErrorResponse = async (response: Response) => {
+    try {
+      const result = await response.json();
+
+      if (result.error) {
+        const fetchError = new FetchError(result.error.message, { data: result });
+        fetchError.status = response.status;
+        return fetchError;
+      }
+    } catch {
+      // Fall back to the generic error below when the error body is not JSON.
+    }
+
+    const fetchError = new FetchError('Server Error');
+    fetchError.status = response.status;
+    return fetchError;
+  };
+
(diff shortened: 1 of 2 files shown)


innerdvations:
Thanks for the contribution — this is the same fetch-client fix as #26204 (non-json `responseType` should still surface the server error body). I'm closing this as a duplicate so we can review a single PR; #26204 also covers a plain-text error body, so we'll keep that one.

_AI-assisted comment._

PR #26624, opened 2026-06-12
fix: match datetime "is"/"is not" filters on the selected minute inst…

### LINKED ISSUE #26614: Content Manager: "is" filter on datetime fields almost never matches (minute-precision picker vs sub-second stored values)
### Node Version

22.22.2

### Package Manager

npm

### Package Manager Version

10.9.7

### Strapi Version

5.47.1

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

In the Content Manager list view, filtering a `datetime` field (including the built-in **Created At** and **Updated At** filters) with the **is** (`$eq`) operator almost never returns results, even when entries clearly exist for the selected date/time.

Root cause is a precision mismatch between the filter input and the stored values:

- The filter UI offers `is` / `is not` for `datetime` attributes (`getFilterList` in `packages/core/admin/admin/src/components/Filters.tsx`, the `datetime` case returns `BASE_FILTERS + NUMERIC_FILTERS`).
- The value input is the generic `DateTimePicker`, which emits `date.toISOString()` with seconds and milliseconds zeroed (`packages/core/admin/admin/src/components/FormInputs/DateTime.tsx`). The design-system `TimePicker` only offers `HH:mm` options (15 minute step by default), so the UI cannot express seconds at all.
- Stored `created_at` / `updated_at` values have sub-second precision (e.g. `2026-06-09 20:25:38.123`).
- There is no server-side special casing of `$eq` on datetime attributes (checked content-manager server and `core/utils` query conversion), so the query is a literal `WHERE created_at = '2026-06-09T20:25:00.000Z'`, which only matches rows that happen to land exactly on a zero-second boundary.

Range operators (`is greater than`, `is less than`, etc.) work as expected since they don't require an exact match. This is consistent with what users report: "is" returns nothing, "greater than" works.

### Steps to Reproduce

1. Create any collection type and add a few entries.
2. Open the list view in the Content Manager.
3. Add a filter: **Created At** + **is** + select today's date (any time).
4. Observe that no results are returned even though entries created today exist.
5. Change the operator to **is greater than or equal to** with the same value: results are returned.

### Expected Behavior

Filtering a datetime field with **is** should match the user's intent at the precision the picker can express, i.e. match the selected minute (or the whole day when only a date is meaningful), instead of performing an exact sub-second timestamp comparison that the UI cannot express.

Two possible directions:

1. Transform `$eq` / `$ne` on datetime fields into a range covering the selected minute (e.g. `$gte` selected minute / `$lt` next minute). Note the filter chip rendering in `Filters.tsx` (`List` / `getFilterDetails`) currently skips object-shaped values, so chip display needs to be handled as part of this.
2. Alternatively, stop offering `is` / `is not` for `datetime` attributes and keep the null checks and range operators, so the UI doesn't expose an operator it can't express.

### Logs

```shell
No errors are logged. The query executes successfully and returns 0 results.
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

- Reported by an enterprise customer via support (Internal TID 9286). Confirmed via the customer's screen recording and against the codebase on `develop`.
- Not a regression: the behavior has existed since the operator list and `DateTimePicker` input were introduced. Reproduces on any OS/database/package manager.
- Timezone adds a second layer of confusion on top of this: the picker converts local time to a UTC ISO string, so even a minute-range match needs to be done on the UTC instant the picker produced.
- Workaround given to the customer: combine two filters, **is greater than or equal to** day start and **is less than** the next day (filters combine as `$and`).

### Confirmation Checklist

akash-dabhi-qed:
…ead of an exact instant

### What does it do?

[26614-fixed.webm](https://github.com/user-attachments/assets/23c36d62-1b53-499c-8e17-e780a7e85819)

 Translates `$eq`/`$ne` filters on `datetime` attributes into a one-minute range instead of a literal instant comparison.

  - `Filters.tsx` → `createDateTimeRangeFilter`: on submit, `is` becomes `{ $gte: minuteStart, $lt: nextMinute }` and `is not` becomes `{ $not: { $gte, $lt } }`. Seconds/milliseconds are floored and the range is computed on the  UTC instant the picker emitted, so the timezone offset is preserved.
  - `Filters.tsx` → `parseDateTimeRangeFilter` (used in `getFilterDetails`): reverse-maps that range back to the  authored `$eq`/`$ne` + value, so the filter chip and edit form keep displaying it as the user entered it. This  also handles the previously skipped object-shaped value in the chip rendering.
  - Added unit tests covering the range construction (`$eq`/`$ne`, seconds flooring) and the reverse round-trip.

### Why is it needed?

  In the Content Manager list view, filtering a `datetime` field (including Created At / Updated At) with `is` almost never returned results. The picker can only express minute precision (`HH:mm`, seconds/ms zeroed) and emits `date.toISOString()`, but stored timestamps carry sub-second precision (e.g. `…:38.123`). A literal `WHERE created_at = '…:00.000Z'` therefore only matched rows landing exactly on a zero-second boundary, so even selecting the exact minute an entry was created returned nothing. Range operators worked, which matched the reported "`is` returns nothing, `greater than` works".

### How to test it?

### Related issue(s)/PR(s)

Fixes #26614 

---
Fixes CPR-395

### DIFF
--- a/packages/core/admin/admin/src/components/Filters.tsx
+++ b/packages/core/admin/admin/src/components/Filters.tsx
@@ -43,6 +43,44 @@ interface FitlersContextValue {
 
 const [FiltersProvider, useFilters] = createContext<FitlersContextValue>('Filters');
 
+const DATETIME_FILTER_RANGE_MS = 60 * 1000;
+
+const createDateTimeRangeFilter = (operator: '$eq' | '$ne', isoValue: string) => {
+  const start = new Date(isoValue);
+  start.setUTCSeconds(0, 0);
+  const end = new Date(start.getTime() + DATETIME_FILTER_RANGE_MS);
+
+  const range = {
+    $gte: encodeURIComponent(start.toISOString()),
+    $lt: encodeURIComponent(end.toISOString()),
+  };
+
+  return operator === '$eq' ? range : { $not: range };
+};
+
+const parseDateTimeRangeFilter = (
+  operatorObj: Record<string, unknown>
+): { operator: string; value: unknown } | null => {
+  const keys = Object.keys(operatorObj);
+
+  if (keys.length === 2 && '$gte' in operatorObj && '$lt' in operatorObj) {
+    return { operator: '$eq', value: operatorObj.$gte };
+  }
+
+  const negated = operatorObj.$not;
+  if (
+    keys.length === 1 &&
+    negated !== null &&
+    typeof negated === 'object' &&
+    '$gte' in negated &&
+    '$lt' in negated
+  ) {
+    return { operator: '$ne', value: (negated as Record<string, unknown>).$gte };
+  }
+
+  return null;
+};
+
 const getFilterDetails = (
   filterEntry: Record<string, unknown>,
   options: Filters.Filter[]
@@ -60,6 +98,14 @@ const getFilterDetails = (
 
   if (typeof operatorObj !== 'object' || operatorObj === null) {
     return null;
+  }
+
+  const attributeType = option.mainField?.type ?? option.type;
+  if (attributeType === 'datetime') {
+    const range = parseDateTimeRangeFilter(operatorObj as Record<string, unknown>);
+    if (range) {
+      return { name, ...range };
+    }
   }
 
   const [operator] = Object.keys(operatorObj as Record<string, unknown>);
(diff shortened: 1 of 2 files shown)


innerdvations (changes requested):
These helpers leak onto the public `@strapi/admin/strapi-admin` surface because `admin/src/index.ts` uses `export *` from `Filters.tsx`. Please keep `export { Filters }` as the only public export from that file (for example `export { Filters } from './components/Filters'` in the barrel) so `createDateTimeRangeFilter` / `parseDateTimeRangeFilter` stay internal. Tests can keep importing them from `../Filters`.

The minute-range `$eq`/`$ne` rewrite itself looks right for #26614.

_AI-assisted comment._

PR #26628, opened 2026-06-12
feat(admin): add active devices session management

innerdvations:
### What does it do?

Adds active session management for Strapi admin users and Users & Permissions refresh-token users.

- Adds a new admin "Active devices" page under the profile menu to list active sessions, show the current session, and revoke one or all sessions.
- Extends the generic session manager with metadata, list-by-user, and ownership-checked revocation APIs while keeping origin-specific data in the generic metadata JSON field.
- Adds U&P Content API endpoints for refresh mode: `GET /auth/sessions` and `DELETE /auth/sessions/:sessionId`.
- Captures IP address, initial login timestamp, last-active timestamp, and a derived device label. IP-based geographic location is explicitly out of scope and not implemented.

### Screenshots

**Profile menu → Active devices**

image

**Active Devices page**

image

**Log out of all devices confirmation**

image

### Why is it needed?

Users need visibility into their active sessions and a way to terminate a specific session/device without relying only on global logout.

### How to test it?

Experimental: `0.0.0-experimental.b270c3f0c01b183c6f622dfed48d884f26a11e49`

Targeted checks run locally:

- `yarn nx build @strapi/utils`
- `yarn jest packages/core/utils/src/__tests__/user-agent.test.ts --silent`
- `yarn jest packages/core/core/src/services/__tests__/session-manager.test.ts --silent`
- `yarn test:front SessionsPage`
- `yarn nx run @strapi/utils:test:ts`
- `yarn nx run @strapi/admin:test:ts:front`
- `yarn nx run @strapi/admin:test:ts:back`
- `npx eslint` on changed source files
- Pre-commit lint-staged: Nx lint + Prettier on staged files

### Related issue(s)/PR(s)

Fixes CMS-1230


jhoward1994:
**[Ideas]**
1. **“Log out all other devices” UI** — the API supports `keepCurrent=true` (`DELETE /admin/users/me/sessions?keepCurrent=true`), but the UI only does a full logout today. Could add a separate action that keeps the current session and revokes the rest (keep the existing full-logout action).
2. **Frontend test** — add an MSW test for the `keepCurrent` delete path: user stays signed in (no redirect to login).

**[Non blocking]** 
1. in `authenticated-session.ts`, use of `lastActiveAt` reflects the last refresh/rotation timestamp, not per-request live activity.

**Blocking ?**
1. **Breaking change (U&P logout):** in refresh mode, empty-body `POST /auth/logout` now revokes only the current session. Global logout requires `{ scope: 'all' }`. Is that a breaking change we should be concerned about ?
2. remove `.github/pr-26628-screenshots/` before merge.

jhoward1994 (changes requested):
See above comment

innerdvations:
@jhoward1994 

> Breaking change (U&P logout):

I agree it's a change, but I would say it's a change towards expectations. The only reason it was "all devices" before is because we didn't have this page, and it was the only safe option. Now that we have granularity, it isn't needed anymore. I'll add a release note though to point out the change in case anyone was expecting it since that's how it worked before.  Does that sound ok?

> use of lastActiveAt reflects the last refresh/rotation timestamp, not per-request live activity

Good catch, but I think that's the best tradeoff here still. True "last active" would mean an extra DB write on every admin request. I've renamed the column to "Last used" in case that feels more accurate (tbh I don't know if it helps 😆)

> (everything else)

Should be fixed now!

jhoward1994 (dismissed):

jhoward1994:
Sonarcloud check still seems to be failing, is that what you were trying to fix with the latest commit ?

innerdvations:
> Sonarcloud check still seems to be failing, is that what you were trying to fix with the latest commit ?

Yeah, I fixed a couple of its warnings, but then the remaining ones were because we had test coverage misconfigured so it didn't see that I had adequate test coverage. I fixed it in https://github.com/strapi/strapi/pull/26753 and now we'll just wait and see if it actually works 😆

jhoward1994 (approved):

Issue #5912, opened 2020-04-22
Graceful Start/Shutdown


diogotcorreia:
**Please describe your feature request:**

It would be great if Strapi could support graceful start/shutdown like shown in the PM2 docs: https://pm2.keymetrics.io/docs/usage/signals-clean-restart/

I'm using a `server.js` file in production, could it also be implemented there? There doesn't seem to be much documentation available regarding `require("strapi");`.
Just saw from another issue that there is a `strapi.stop()` method, but what about listening for when the server is ready?

Also, I'm willing to submit a PR, but some guidance would be appreciated:
- Where should I place these functions?
- What needs to be stopped? Only Koa and the DB or something else?

diogotcorreia:
Creating a `server.js` with this code works as a workaround:

```javascript
const strapi = require('strapi');
const app = strapi();
app.start(() => process.send('ready'));

process.on('SIGINT', () => {
  if (app.server) {
    app.server.close(() => {
      process.exit(0);
    });
  }
});
```

Should I still add it as a PR?

lauriejim:
I ping @alexandrebodin about that.

alexandrebodin:
Hi @diogotcorreia, I think that using the server.js technique is the only good way of doing it :) 

We need to improve the "strapi" programmatic API in order to make it more usable as a framework. This is clearly a topic on our mind that we will work on in during the rest of the year :) 

Your approache looks really good. We will need to provide some simple access to also close your database connection or stop some I/O service to make it more complete :)

leonardomerlin:
I'm closing the db connections during my tests like this:

```js
afterAll(async (done) => {

  // Destroy (db) connections
  for(const conn of Object.values(strapi.connections)) {
    if(conn.destroy) {
      await conn.destroy();
    }
  }

  done();
});
```

> BTW, using this approach I can run jest without the `--detectOpenHandles` or `--forceExit` flags.

IMHO, this issue can be considered a partial implementation of #5253

sazaam:
cant have it work... at all
process.send('ready')); process send is not a function

littleski:
There is also another issue to be tacled regarding "plugins" (at least).
I use the following plugins:
-            'content-manager',
-            'content-type-builder',
-            'documentation',
-             'email',
-             'sentry',
-             'upload',
-             'users-permissions'

And with them, i have around 11-12Mb of Memleak per test suite at dismount:
(this is run inBand with a forced garbage collection after each test)
[...]

PR #26632, opened 2026-06-12
fix(server): graceful shutdown drain before process.send('reload') — fixes #26629

### LINKED ISSUE #26629: feat/bug: reloader sends process.send('reload') with no drain — in-flight saves lost on schema change restart
### Node Version

20.0.0

### Package Manager

yarn

### Package Manager Version

4.12.0

### Strapi Version

5.7.0

### Operating System

Other

### Database

Other

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

**Strapi version:** 5.x (latest)
**Node.js version:** 20.x
**Database:** any
**OS:** any

**Describe the bug:**
When a content editor is mid-save (HTTP POST/PUT request in-flight) and a
schema change triggers a server reload, `reloader.ts` calls
`process.send('reload')` immediately. The parent process kills the worker
with no drain period. The in-flight save request is terminated with a
connection reset — the editor loses their work with no warning.

**Steps to reproduce:**
1. Open the Content Manager and start editing a long article.
2. In a second browser tab (as admin), open Content-Type Builder and
   save a field change, triggering a server reload.
3. Switch back to the first tab and click Save.
4. The save fails with a network error / connection reset.

**Expected behaviour:**
The server should stop accepting new connections but allow in-flight
requests up to a configurable timeout (e.g. 30s) to complete before
sending the reload signal. This is standard graceful shutdown behaviour
(Node.js `httpServer.closeIdleConnections()` + drain window).

**Relevant code:**
packages/core/core/src/services/reloader.ts — the `reload()` function
calls `process.send('reload')` with no drain:
```ts
if (strapi.config.get('autoReload')) {
  process.send?.('reload'); // ← no drain, kills in-flight requests
}
```
packages/core/core/src/services/server/http-server.ts — the HTTP server
has no SIGTERM/graceful-shutdown handler.

**Additional context:**
Issue #5912 requested graceful shutdown in v3/v4 but was closed as
"handled in v4". It did not address the v5 reloader's reload path or
production HTTP drain. This is a distinct regression in v5 behaviour.

### Steps to Reproduce

1. Start a Strapi v5 project with autoReload enabled (development mode):
   `yarn strapi develop`

2. Open the Content Manager in the browser and navigate to any Collection Type
   with a large content entry (e.g. an article with a Rich Text field).

3. Begin editing the entry — type a significant amount of content so the
   save request will take a moment to complete.

4. In a second browser tab (same admin user or another admin), open the
   Content-Type Builder and add or remove any field from any content type,
   then click Save. This triggers reloader.ts to call process.send('reload').

5. Immediately switch back to the first tab and click the Save button
   before the server has fully restarted.

6. Observe that the save request fails with a network error or connection
   reset (browser DevTools → Network tab shows the POST/PUT request as
   "failed" or "cancelled" with no response).

7. The content editor receives no meaningful error message — the work
   written in step 3 is lost silently.

Expected: The server should finish handling the in-flight save request
before reloading. The editor's save should succeed or, at minimum, the
editor should receive a clear retry prompt.

Actual: The request is terminated mid-flight. No data is saved.
No warning is shown to the editor.

### Expected Behavior

When a schema change triggers a server reload, any HTTP requests that are
already in-flight (e.g. a content editor's save request) should be allowed
to complete before the reload signal is sent to the parent process.

Specifically:
1. The server should stop accepting new incoming connections immediately.
2. Existing in-flight requests should be given a configurable grace period
   (e.g. 30 seconds, via a server.shutdownTimeout config option) to finish.
3. Once all in-flight requests have completed — or the timeout is reached —
   only then should process.send('reload') be called.
4. If the timeout is reached before all requests finish, the server should
   log a warning and proceed with the reload, rather than hanging indefinitely.

The content editor's save should succeed normally. If the timeout is
genuinely exceeded (an unusually slow request), the editor should receive
a proper HTTP error response they can retry — not a silent connection reset
that loses their work with no feedback.

This is standard Node.js graceful shutdown behaviour, achievable with
httpServer.closeIdleConnections() combined with httpServer.close() and
a drain timeout, both of which are available in Node.js 20+.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

m-zandi:
## What does it do?
Adds a configurable drain period to `reloader.ts` and `http-server.ts`
so in-flight HTTP requests (e.g. a content editor's save) are allowed
to complete before `process.send('reload')` is called on a schema
change restart.

## Why is it needed?
Fixes #26629. Without this change, a schema change by one user
kills any concurrent save requests from other editors silently,
causing data loss with no user-facing error.

## Changes
- `reloader.ts` — made `reload()` async; added drain via
  `httpServer.closeIdleConnections()` + `httpServer.close()`
  with configurable timeout before the reload signal
- `http-server.ts` — added `SIGTERM` handler with same drain logic
- `index.ts` — added `server.shutdownTimeout` default (30000ms)

## Related issue(s)
Closes #26629
Related: #5912 (v3 graceful shutdown, closed)

## Test plan
1. Start Strapi in development mode
2. Open Content Manager and begin a slow save (throttle network in DevTools)
3. Trigger a schema reload from Content-Type Builder simultaneously
4. Confirm the save completes successfully before the server reloads

---
Fixes CPR-394

### DIFF
--- a/packages/core/core/src/configuration/index.ts
+++ b/packages/core/core/src/configuration/index.ts
@@ -23,6 +23,7 @@ const defaultConfig = {
     host: process.env.HOST || os.hostname() || 'localhost',
     port: Number(process.env.PORT) || 1337,
     proxy: false,
+    shutdownTimeout: 30_000,
     cron: { enabled: false },
     admin: { autoOpen: false },
     dirs: { public: './public' },
--- a/packages/core/core/src/services/reloader.ts
+++ b/packages/core/core/src/services/reloader.ts
@@ -6,18 +6,46 @@ export const createReloader = (strapi: Core.Strapi) => {
     isWatching: true,
   };
 
-  function reload() {
-    if (state.shouldReload > 0) {
-      // Reset the reloading state
-      state.shouldReload -= 1;
-      reload.isReloading = false;
-      return;
+async function reload() {
+  if (state.shouldReload > 0) {
+    state.shouldReload -= 1;
+    reload.isReloading = false;
+    return;
+  }
+
+  if (strapi.config.get('autoReload')) {
+    const server = strapi.server?.httpServer;
+    if (server) {
+      const timeout: number =
+        strapi.config.get('server.shutdownTimeout', 30_000);
+
+      strapi.log.info(
+        `[reloader] Draining in-flight requests before reload ` +
+        `(timeout: ${timeout}ms)...`
+      );
+
+      server.closeIdleConnections();
+
+      await new Promise<void>((resolve) => {
+        const t = setTimeout(() => {
+          strapi.log.warn(
+            '[reloader] Drain timeout reached, proceeding with reload'
+          );
+          resolve();
+        }, timeout);
+
+        server.close(() => {
+          clearTimeout(t);
+          resolve();
+        });
+      });
     }
 
-    if (strapi.config.get('autoReload')) {
-      process.send?.('reload');
-    }
+    process.send?.('reload');
   }
+}
+
+  
 
   Object.defineProperty(reload, 'isWatching', {
     configurable: true,
--- a/packages/core/core/src/services/server/http-server.ts
+++ b/packages/core/core/src/services/server/http-server.ts
@@ -63,6 +63,31 @@ const createHTTPServer = (strapi: Core.Strapi, koaApp: Koa): Server => {
     });
   };
 
+process.once('SIGTERM', async () => {
+    const timeout: number =
+      strapi.config.get('server.shutdownTimeout', 30_000);
+
+    strapi.log.info(
+      `[server] SIGTERM received — draining connections (${timeout}ms timeout)`
+    );
+
+    server.closeIdleConnections();
+
+    await new Promise<void>((resolve) => {
+      const t = setTimeout(() => {
+        strapi.log.warn('[server] Forced close after timeout');
+        resolve();
+      }, timeout);
+
+      server.close(() => {
+        clearTimeout(t);
+        resolve();
+      });
+    });
+
+    process.exit(0);
+  });
+  
   return Object.assign(server, { destroy });
 };
 


innerdvations (changes requested):
## Change requests

Thanks for tackling the reload drain for #26629 — the reloader direction is right, but this isn't merge-ready yet.

### 1. Drop the competing `SIGTERM` handler (`http-server.ts`)
We already register `destroyOnSignal` (`packages/core/core/src/utils/signals.ts`) which calls `strapi.destroy()` on SIGTERM/SIGINT. A second `process.once('SIGTERM')` races that path and can `process.exit(0)` without tearing down DB/plugins/lifecycles. Prefer drain (+ timeout, then force-close remaining sockets) inside `httpServer.destroy()` if you want production drain in this PR; otherwise remove the new handler and keep the fix scoped to the reloader path.

### 2. Type `server.shutdownTimeout` on `Core.Config.Server`
As written, `defaultConfig` fails `satisfies Partial<Core.Config.Server>` with TS2353 (`shutdownTimeout` does not exist). Add it to `@strapi/types` in this PR.

### 3. Prettier
`yarn prettier --check` fails on `reloader.ts` and `http-server.ts` (indentation around the new blocks).

### 4. Tests
Please add unit tests for drain-before-`process.send('reload')` and timeout-then-proceed. Manual QA alone is too thin for a data-loss fix.

### 5. Re-entrancy on async `reload()`
`reload()` can now run up to `shutdownTimeout`. Overlapping CTB `setImmediate` + file-watcher calls need a guard for the whole drain window so we don't double-close / double-send `reload`.

Whatever lands must be production-ready as merged — reloader-only drain (with the broken SIGTERM code removed) is fine; a half-working production signal path is not.

_AI-assisted comment._

PR #26633, opened 2026-06-13
fix(database): escape wildcards in LIKE filters

### LINKED ISSUE #26468: Potential bug: wildcard characters in filters are not escaped, causing incorrect matching for literal text

mturac:
## What
- Escape `%`, `_`, and backslash in LIKE-based database filters so literal text is not treated as a wildcard pattern.
- Keep case-insensitive equality filters on equality comparisons instead of LIKE.
- Preserve MySQL-compatible SQL generation while using an explicit LIKE escape clause for other dialects.
- Add focused query-builder regression coverage for literal wildcard values.

## Testing
- `yarn test:unit packages/core/database/src/__tests__/index.test.ts --selectProjects "Core database" --runInBand -t "Query builder LIKE wildcard filters"`
- `yarn prettier --check packages/core/database/src/query/helpers/where.ts packages/core/database/src/__tests__/index.test.ts`
- `yarn workspace @strapi/database lint`
- `git diff --check`

Fixes #26468

### DIFF
--- a/packages/core/database/src/__tests__/index.test.ts
+++ b/packages/core/database/src/__tests__/index.test.ts
@@ -1,6 +1,7 @@
 import knex from 'knex';
 import { Database, DatabaseConfig } from '../index';
 import createQueryBuilder from '../query/query-builder';
+import { applyWhere } from '../query/helpers/where';
 import type { Model } from '../types';
 
 // create an in-memory db connection to test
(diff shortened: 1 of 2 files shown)


masonmcelvain:
Hi @mturac! I think this is a duplicate of https://github.com/strapi/strapi/pull/26476

mturac:
Thanks for pointing this out. I agree that #26476 covers the same filter escaping/equality path, so I’m closing this one to keep the review focused there.

PR #26648, opened 2026-06-15
fix(ci): avoid syncing CPR labels to CMS tickets

innerdvations:
### What does it do?

Updates the community PR triage action so existing tickets found on the CMS team are refreshed without rewriting Linear labels. CPR-owned tickets still keep the existing label sync behavior, and the bundled action output is rebuilt.

### Why is it needed?

Community PR sync can find a ticket that has already been picked up on the CMS team, then try to update it with labels from the CMS-Community-PRs team. Linear rejects those team-scoped label IDs, causing the `Community PR — Create Linear Ticket` job to fail with:

```
LabelIds for incorrect team - The label 'content-manager' is not associated with the same team as the issue.
```

### How to test it?

- `yarn workspace community-pr-triage typecheck`
- `yarn workspace community-pr-triage build`

### Related issue(s)/PR(s)

Fixes the failing triage run: https://github.com/strapi/strapi/actions/runs/27430000679/job/81077570040


Bassel17 (approved):
LGTM

Issue #25076, opened 2025-12-22
Data Loss When Renaming Collection Fields


HusnainIzhar:
### Node Version

22.16.0

### NPM/Yarn/PNPM Version

NPM

### Strapi Version

5.33.0

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Javascript

### Description
When renaming a field in the Content-Type Builder that contains existing data, the data disappears from the Entry Detail view in the Content Manager, even though it still exists in the database.

### Steps to Reproduce

1. Create a new collection type (e.g., "Article")
2. Add a text field (e.g., "title")
3. Create several entries with data in the "title" field
4. Navigate to Content-Type Builder
5. Edit the "title" field and rename it to "articleTitle"
6. Click "Finish" and "Save"
7. Wait for the server to restart
8. Go back to Content Manager and open any existing entry
9. **Expected**: The data should appear in the "articleTitle" field
10. **Actual**: The "articleTitle" field is empty

### Expected Behavior
When a field is renamed, the existing data should be preserved and appear in the renamed field. The database column should be properly renamed to maintain data integrity.

### Actual Behavior
The renamed field appears empty in the UI. The data still exists in the database under the old column name, but the frontend filters it out because it doesn't match the current schema.

### Environment
- **Strapi Version**: 5.x (affects all versions)
- **Operating System**: Windows/macOS/Linux (all)
- **Database**: PostgreSQL/MySQL/SQLite (all affected)
- **Node Version**: Any
- **NPM Version**: Any

#### Backend Behavior
In `packages/core/content-type-builder/server/src/services/schema-builder/content-type-builder.ts`, the `editContentType` method treats field renaming as two separate operations:
- Delete the old field (e.g., "title")
- Create a new field (e.g., "articleTitle")

The database schema builder does **not** perform an actual column rename. Instead:
- A new column "articleTitle" is created
- The old column "title" remains with the data
- No data migration occurs

#### Frontend Behavior
In `packages/core/content-manager/admin/src/pages/EditView/utils/data.ts`, the `removeFieldsThatDontExistOnSchema` function (line 152-164) filters out any data keys that don't exist in the current schema:

```typescript
const removeFieldsThatDontExistOnSchema = (schema: PartialSchema) => (data: AnyData) => {
  const schemaKeys = Object.keys(schema.attributes);
  const dataKeys = Object.keys(data);
[...]

Issue #19075, opened 2023-12-16
Changing field name wipes fields data for all entries


gregg-cbs:
## Bug report
In strapi you make it easy to change a fields name as if doing so would have no consequences. Changing a fields name is destructive, it deletes all data for that field but nothing tells you of this fact. Strapi should retain the fields data, im not sure why you wouldnt rename the table column or move the data from the old column to new column if you are deleting the old column.

### Required System information
- Node.js version: 18
- NPM version: 9.5.1
- Strapi version: 4.15.0
- Database: postgres
- Operating system: Windows
- Is your project Javascript or Typescript: TS

### Describe the bug

Changing a field name in the strapi admin panel results in loss of all data in all rows for that field.

### Steps to reproduce the behavior

1. Go to content type builder
2. Click on a collection
3. Edit a field and change its name, save
4. See now that fields data has been wiped out
5. Notice no warnings were mentioned that this action is destructive

### Expected behavior
When renaming a field I would expect the data to stay in tact.

### Screenshots
![image](https://github.com/strapi/strapi/assets/57114335/85895504-3a08-4288-beb5-771631478851)
![image](https://github.com/strapi/strapi/assets/57114335/14c4c57d-b1c5-4d16-82a7-43d96af2a7f0)

gregg-cbs:
I guess you can only rename fields on development so this would never happen on live.

derrickmehaffy:
The problem is the context of a rename isn't known. When you rename a field and that's saved to the schema file, when the app restarts it looks at the schema file and makes the schema changes it needs to, but effectively it sees that rename as you deleted a field and added a new one since the schema file has no history, it's effectively a diff.

gregg-cbs:
Thanks for responding even though i had closed the issue. I guess each field could get given an ID so the diff becomes a bit more sophisticated. Also when renaming a field it loses all its interface settings - width, placement on the page, settings etc.

I think an alert should show when renaming a field so the user can know its destructive on many levels because I did not expect renaming to have so many repercussions because it was so easy to do.

But overall understandable.

Issue #12626, opened 2022-02-22
Renaming a field in a content-type will erase the existing data


paulovieira:
### Describe the bug

If I rename a field in a content-type, and if there is some previous data there (in that field), all the data will be erased. 

Inspecting the database directly (SQLite) I see that
1.  the new column has `<null>` in all rows
2. the old column was deleted

### Steps to reproduce the behavior

1. Create a new content-type (doesn't matter if it's a collection type, single type or component). Example: "restaurant". Add a text field named "something".
2. Inspect the SQLite database. Verify that strapi created a table named `restaurants`, and it has a column named `something`
3. Add some content using the Content Manager.
4. Using the Content-Type Builder, rename the field to `something2`
5. Inspect the database again. I see that a new column named `something2` was added to the `restaurants` table (with `<null>` in all rows) and the column `something` was deleted.

### Expected behavior

I've read in several places that strapi should never deleted data for this kind of operations. 

As a technical user, I expected that the initial `something` column should not be deleted ("never loose data")

As a non-technical user (who doesn't know how to access the data from the database directly), I expected that the `something2` column should be populated with the data from `something` (imagine a content type with 1000's of records...)

### System

- Node.js version: v14
- Strapi version: 4.1.0
- Database: SQLite
- Operating system: Linux

### Additional context

I'm aware that #1114 was used for this kind of issues, but it's closed and this particular thing (renaming a field) doesn't seem to be convered there.

I was surprised to learn about this behaviour. Being able to rename a field is one of the most basic features for a tool like strapi. 

APIs should evolve naturally.  Furthermore, SQLite is the default database system, so I imagine there is a lot of people using it.

derrickmehaffy:
@strapi/content and @strapi/developer-experience I don't know which squad this goes in so I set it in both of your projects.

We have discussed this internally at length before and we already know it requires sending more data to the backend then is currently sent.

---

To users in the meantime, when you are renaming you will need to use something like what is in this repo: https://github.com/derrickmehaffy/strapi-example-v4-custom-db

It won't stop you from losing the data locally (but that should be atomic anyway) but will help with deploying into production. Alternatively you can rename manually via the schema files and build your migration at the same time.

apsonandtesla:
Yeah we have found out the same issue and personally I think it's kind of important. If there is anything we can do for help, we are happy to help. Hope this get fixed as soon as possible.

Convly:
[...]

Issue #12597, opened 2022-02-19
Renaming component field of a collection does not migrate data over


oriooctopus:
## Bug report

I have a collection `sublesson` with a component field called `challenges`. If I rename that field from `challenges` to `steps` for example, none of the data is migrated over and I would theoretically have to redo it all.

### Steps to reproduce the behavior

1. Create a collection 
2. Create a component
3. Attach the component to the collection via a field
4. Populate the data
5. Rename the component field

### Expected behavior

The data should migrate over

### Screenshots

<img width="1082" alt="Screen Shot 2022-02-19 at 11 34 21 AM" src="https://user-images.githubusercontent.com/20331238/154805316-46f91ff1-8ce8-49ee-8238-19c20dd986ba.png">

### Code snippets

If applicable, add code samples to help explain your problem.

### System

- Node.js version: 14.8.1
- Strapi version: 4.0.7
- Database: PostgreSQL
- Operating system: MacOS

kasonde:
Thanks for reporting, marking this as a feature request

derrickmehaffy:
duplicate: https://github.com/strapi/strapi/issues/12626

PR #26655, opened 2026-06-15
test(e2e): add critical edit-schema no-data-loss test

MarionLemaire:
## What
Adds a `@critical` E2E test for **gap #4 — `ctb.edit-schema.no-data-loss`**: changing a content type's schema in the Content-Type Builder must not destroy existing entry data.

## What the test does
**`Adding then removing a field does not corrupt existing entries`** (active ✅)
1. Opens the seeded "West Ham post match analysis" Article and records its title.
2. Adds a `tempnotes` text field → reopens the entry → title still intact across the restart.
3. Deletes `tempnotes` → reopens the entry → title still intact across the second restart.

This is the genuine "no data loss" guarantee for the critical path, and it passes on chromium / firefox / webkit.

## The field-rename scenario (intentional `test.fixme`)
The file also includes a `test.fixme` for *renaming* a field while preserving its data. This is **kept as a documented skip, not a failing assertion**, because renaming does not preserve data in Strapi by design:

- The CTB writes the new attribute name into the schema; on restart the DB sync (`packages/core/database/src/schema/diff.ts`) matches columns **by name**, so the old column is treated as "removed" and dropped (`builder.ts`) while the new column is created empty. There is no `renameColumn` for user-initiated renames.
- Confirmed **"expected, not a bug"** by maintainers in [#25076](https://github.com/strapi/strapi/issues/25076) (see also [#19075](https://github.com/strapi/strapi/issues/19075), [#12626](https://github.com/strapi/strapi/issues/12626), [#12597](https://github.com/strapi/strapi/issues/12597)), tracked as a [feature request](https://feedback.strapi.io/developer-experience/p/gracefully-handle-renaming-of-content-types-and-fields-in-the-ctb).

Keeping it as a `fixme` documents the coverage gap (visible in test reports) without a red `@critical` test, and it can be flipped to a real assertion if/when the CTB supports rename-with-migration. No new bug filed — the existing closed issues already cover it.

## Part of
E2E coverage focus — Phase 1.

### DIFF
--- a/tests/e2e/tests/content-type-builder/collection-type/edit-schema-no-data-loss.spec.ts
+++ b/tests/e2e/tests/content-type-builder/collection-type/edit-schema-no-data-loss.spec.ts
@@ -1,0 +1,86 @@
+import { test, expect } from '@playwright/test';
+import { resetFiles } from '../../../../utils/file-reset';
+import { sharedSetup } from '../../../../utils/setup';
+import { addAttributesToContentType } from '../../../../utils/content-types';
+import { clickAndWait, navToHeader, findAndClose } from '../../../../utils/shared';
+import { waitForRestart } from '../../../../utils/restart';
+
+test.describe('CTB - Edit schema without data loss', { tag: ['@critical'] }, () => {
+  // Long timeout — triggers multiple server restarts
+  test.describe.configure({ timeout: 500000 });
+
+  test.beforeEach(async ({ page }) => {
+    await sharedSetup('ctb-edit-schema-no-data-loss', page, {
+      resetFiles: true,
+      importData: 'with-admin',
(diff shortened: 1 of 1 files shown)


jhoward1994:
Should we also assert the seeded entry's `slug` survives both schema changes, not just `title`? Just as an extra safety net confirming that all the data is intact

e.g.

> ```ts
> await expect(page.getByRole('textbox', { name: 'slug' })).toHaveValue('west-ham-post-match-analysis');
> ```

**Follow-up (not blocking):** We could also verify the other i18n entries e.g. FR locale entry (`Analyse post-match contre West Ham`) ?

jhoward1994 (approved):

Issue #13, opened 2015-10-17
Support JSON API


kaermorchen:
Hello.

Do you plan to add support [JSON API](http://jsonapi.org)? Support JSON API important Ember.js developers and other frameworks.

loicsaintroch:
Hey @kaermorchen.

Good question. We'll definitely support JSON API. I know a lot of developers ask for this support especially Ember developers you're right.

Maybe this is something we can plan for the `v1.3.0`.
Are you familiar with this? If so, do you want to help us on it?

kaermorchen:
I think I can try help with it.

loicsaintroch:
Awesome thank you! Feel free to join us on Slack to talk about it: http://strapi.io/support

loicsaintroch:
Hey @kaermorchen, it's been a while since your last message. We are going to release the v1.4.0 in a few minutes so we'll support JSON API when you'll be ready. Please keep in touch.

kaermorchen:
I'm sorry, at the moment I don't work with strapi. I was not able to work with studio on my system (windows 7 x64). And I switched to work with loopback.

loicsaintroch:
Hey @kaermorchen, is your issue related to #29? If so, can you please help us?
Have you tried with `strapi@1.4.0`?

```bash
$ npm uninstall strapi -g
$ npm cache clean
$ npm install strapi -g
$ strapi new myApp
$ cd myApp
$ strapi start
```

loicsaintroch:
@kaermorchen: I'm closing this issue since we unfortunately don't have any news from you... Also, we'll support the JSON API specification in the v2.0.0 as you can see on the [roadmap](https://github.com/wistityhq/strapi/blob/master/ROADMAP.md).

PR #26657, opened 2026-06-15
test(e2e): add critical users-permissions role-matrix smoke

MarionLemaire:
## What
Adds a `@critical` E2E smoke for **gap #13 — `permissions.users-permissions.role-matrix`**: the *end-user* Users & Permissions roles (Public / Authenticated), distinct from admin RBAC.

## What the test does
1. **UI smoke** — navigates to Settings → Users & Permissions → Roles and asserts the built-in **Public** and **Authenticated** roles render.
2. **Public READ** — proves the gate both ways: with `find`/`findOne` disabled on the Public role, `GET /api/articles` returns **403**; once granted, it returns **200**.
3. **Authenticated CREATE** — grants `create` to the Authenticated role, registers an end-user (the fixture has `allow_register: true`, `email_confirmation: false`, `default_role: authenticated`), and `POST /api/articles` with their JWT returns **201**.

Per the coverage matrix, #13 is primarily an API guarantee with a thin E2E smoke, so the matrix is configured/asserted through the admin + content API while the settings page itself is exercised in the browser.

## Robustness note
The e2e DB reset only covers `ALLOWED_CONTENT_TYPES`, which **excludes** `plugin::users-permissions.{user,role,permission}` — so U&P state is **not** reset between tests/browsers. The test is therefore fully self-contained:
- it sets the exact permission state it needs (rather than assuming a pristine baseline), and
- it registers with a unique per-run identity so it never collides with a user left behind by an earlier run.

Verified passing on chromium, firefox, and webkit, and idempotent across consecutive reruns against already-mutated state.

## Part of
E2E coverage focus — Phase 1. https://app.notion.com/p/strapi/E2E-coverage-focus-3738f35980748111a75bcc596272f8d7

### DIFF
--- a/tests/e2e/tests/settings/users-permissions/role-matrix.spec.ts
+++ b/tests/e2e/tests/settings/users-permissions/role-matrix.spec.ts
@@ -1,0 +1,124 @@
+import { test, expect, type APIRequestContext } from '@playwright/test';
+import { login } from '../../../../utils/login';
+import { resetDatabaseAndImportDataFromPath } from '../../../../utils/dts-import';
+import { navToHeader } from '../../../../utils/shared';
+import { ADMIN_EMAIL_ADDRESS, ADMIN_PASSWORD } from '../../../constants';
+
+// Critical path #13 — permissions.users-permissions.role-matrix
+//
+// This is primarily an API-layer guarantee (the end-user Users & Permissions roles, NOT admin RBAC),
+// so the matrix is configured and asserted via the content/admin API. A thin UI smoke confirms the
+// U&P Roles settings page itself renders. Per the coverage matrix this lives in tests/e2e/tests/settings.
+//
+// NOTE: the e2e DB reset only covers ALLOWED_CONTENT_TYPES, which excludes U&P users/roles/permissions.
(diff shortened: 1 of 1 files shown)


jhoward1994:
**Follow-up (not blocking):** Maybe we should consider a dedicated U&P reset strategy for e2e — e.g. extending `ALLOWED_CONTENT_TYPES` to include `plugin::users-permissions.{role,permission,user}` for selective imports, or an `afterEach` cleanup helper — so future U&P specs don't have to handle this individually. This test's approach is fine for merge.

 **(non-blocking):** `setArticleActions` is a good candidate for `tests/e2e/utils/users-permissions.ts` once a second U&P matrix spec lands.

 **(non-blocking):** The registration comment says "the fixture has `allow_register=true`…" — those are U&P **bootstrap defaults** (`packages/plugins/users-permissions/server/bootstrap/index.js`), not `with-admin` DTS data. The comment could be:

 ```ts
 // U&P bootstrap defaults: allow_register=true, email_confirmation=false, default_role=authenticated
```

jhoward1994 (approved):

PR #26448, opened 2026-05-27
fix(admin): exclude disabled plugins from admin build

Jasleen-Kaur96:
Fixes #23269

PR #26660, opened 2026-06-15
enhancement(admin): hide deploy-now widget in production

innerdvations:
### What does it do?

Hides the **Deploy now** homepage widget when the project runs in **production**.

- Adds a small pure helper `hideDeployNowWidgetInProduction` in `admin/src/utils/widgetLayout.ts`.
- Applies it in `HomePageCE` where widgets are gathered, so the widget is excluded from both the homepage grid and the "Add widget" modal in production.
- Relies on the **existing** server-reported `currentEnvironment` (from `/admin/information`, already exposed via `useAppInfo`) — no new "is this production?" frontend logic, no new endpoint, and no `window.strapi` plumbing.
- The widget keeps rendering in non-production environments and when the environment is unknown (fail-open).

This is a reworked, minimal take on the original idea in #26549, rebased on current `develop`. Since that PR was opened, #26448 reworked widget registration to gate the deploy-now widget on the Strapi Cloud plugin being bundled (`'strapi-cloud' in this.appPlugins`). Disabling the cloud plugin isn't a viable way to hide the widget for Cloud users (they want the plugin, just not the upsell), so this change targets the environment instead and avoids the original PR's fragile `backendURL.includes('localhost')` heuristic (which misfires on `127.0.0.1`, LAN IPs and custom hostnames).

### Why is it needed?

The deploy-now widget is a Strapi Cloud upsell promoting a first deployment. Once a project is live in production, the prompt is irrelevant and adds noise. See the feedback request: [Remove the deploy widget](https://feedback.strapi.io/customization/p/option-to-remove-the-default-deploy-widget-in-dashboard).

### How to test it?

1. Run the app in development — the **Deploy now** widget appears on the homepage (when the Strapi Cloud plugin is bundled).
2. Run the app in production (`NODE_ENV=production` with a production start) — the widget is no longer registered/visible, and is absent from the "Add widget" modal.
3. Unit tests: `yarn workspace @strapi/admin test:front src/utils/tests/widgetLayout.test.ts`.

### Related issue(s)/PR(s)

- Based on the original idea from #26549 (credited via `Co-authored-by`).
- Related to #26448 (deploy-now widget now gated on the cloud plugin).
- Fixes CPR-385

### DIFF
--- a/packages/core/admin/admin/src/pages/Home/HomePage.tsx
+++ b/packages/core/admin/admin/src/pages/Home/HomePage.tsx
@@ -14,6 +14,7 @@ import { WidgetResizeHandle } from '../../components/ResizeIndicator';
 import { Widget } from '../../components/WidgetHelpers';
 import { WidgetRoot } from '../../components/WidgetRoot';
 import { useEnterprise } from '../../ee';
+import { useAppInfo } from '../../features/AppInfo';
 import { useAuth } from '../../features/Auth';
 import { useStrapiApp } from '../../features/StrapiApp';
 import { useWidgets } from '../../features/Widgets';
@@ -30,6 +31,7 @@ import {
   isLastWidgetInRow,
   canResizeBetweenWidgets,
   getWidgetWidth,
+  hideDeployNowWidgetInProduction,
 } from '../../utils/widgetLayout';
 
 import { AddWidgetModal } from './components/AddWidgetModal';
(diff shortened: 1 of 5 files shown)


mathildeleg (approved):
`widgetLayout.ts` is mostly designed for the maths around how to place/resize the widgets, not which widgets to display. I think it's fine to keep this here for now, but if we have more filtering to do on which widgets to display when/where, it might make sense to create another dedicated file for that!
LGTM to me though :)

mathildeleg (dismissed):

innerdvations:
Good idea! I moved it, does this work better?

mathildeleg (approved):
Perfect! Thanks ✨

PR #26661, opened 2026-06-15
test(e2e): add extended advanced-filters UI/API parity test (#21)

MarionLemaire:
## What
Adds an `@extended` E2E test for **Phase 2 path #21 — `cm.advanced-filters`**: "each operator × field type, UI and API return the same rows."

## What the test does
For a text field (`name`) across three operators — **contains**, **is** (`$eq`), **is not** (`$ne`) — it:
1. Applies the filter through the list-view Filters UI.
2. Asserts exactly the expected rows render in the list.
3. Asserts a direct admin content-manager query with the equivalent filter returns the **same set of rows**.
4. Removes the filter chip before the next case.

## Notes
- Uses **Author**: non-localized (so there's no locale confound — Article's en/fr titles both contain "West"/"soccer") with three distinct seeded names: Ted Lasso, Coach Beard, Led Tasso.
- The API side queries the admin content-manager endpoint (the same data source the list view reads), so the comparison validates that the filter UI builds a query whose result set matches the API's.
- Discovered while writing: the CM **persists the list-view query** (sessionStorage), so navigating to the bare list URL does **not** clear an active filter — the test removes the filter chip between cases instead.

Verified passing on chromium, firefox, and webkit.

## Part of
E2E coverage focus — Phase 2 (@extended). https://app.notion.com/p/strapi/E2E-coverage-focus-3738f35980748111a75bcc596272f8d7

### DIFF
--- a/tests/e2e/tests/content-manager/advanced-filters.spec.ts
+++ b/tests/e2e/tests/content-manager/advanced-filters.spec.ts
@@ -1,0 +1,116 @@
+import { test, expect, type APIRequestContext, type Page } from '@playwright/test';
+import { login } from '../../../utils/login';
+import { navToHeader } from '../../../utils/shared';
+import { resetDatabaseAndImportDataFromPath } from '../../../utils/dts-import';
+import { ADMIN_EMAIL_ADDRESS, ADMIN_PASSWORD } from '../../constants';
+
+// Critical path #21 — cm.advanced-filters (@extended)
+//
+// For a representative set of operators on a text field, assert the list-view filter UI and a direct
+// API query with the equivalent filter return the same set of rows.
+//
+// Uses Author: non-localized (no locale confound) with three distinct seeded names — Ted Lasso,
+// Coach Beard, Led Tasso. The API side uses the admin content-manager endpoint (the same data source
+// the list view reads), so the comparison validates that the filter UI builds a query whose result
+// set matches the API's.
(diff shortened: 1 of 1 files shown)


tests/e2e/tests/content-manager/advanced-filters.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        // API: the equivalent query returns the same set of rows.
+        const apiNames = await apiFilterNames(page.request, adminHeaders, query);
+        expect(apiNames, `API rows for "${operatorLabel}"`).toEqual([...expected].sort());
+
+        // Remove the filter chip so the next case starts from an unfiltered list.
+        // (The CM persists the list query, so navigating wouldn't clear it.)
+        await page.getByRole('button', { name: `name ${operatorLabel} ${value}` }).click();
+        await expect(page.getByRole('gridcell', { name: 'Ted Lasso', exact: true })).toBeVisible();
```
jhoward1994, line 117: Could we assert the chip is visible right after `applyFilter`, and `toHaveCount(0)` after clicking it here? 

```ts
        const chip = page.getByRole('button', { name: `name ${operatorLabel} ${value}` });
        await expect(chip).toBeVisible();
        // UI: exactly the expected rows are shown.
		...


        await chip.click();
        await expect(chip).toHaveCount(0);

```

jhoward1994 (approved):

PR #26666, opened 2026-06-15
ci(pr-reviewer): gate AI review on API org membership check

cache-your-dreams:
### What does it do?

Dedicated gate job that resolves org membership via the GitHub API (orgs.checkMembershipForUser).

`author_association` is a per-event snapshot, not a membership lookup: private org members are reported as CONTRIBUTOR, so legitimate maintainers were being skipped — the same blind spot community-label.yml already struggles with.

The gate job skips bots, treats a 404 as "not a member", and fails closed on transient API errors so an unverified author is never let through. The review job now `needs: gate` and runs only when `is_member == 'true'`.

Requires an ORG_READ_TOKEN secret (PAT or GitHub App token with org `members:read`) — the default GITHUB_TOKEN cannot read private org membership.

### Why is it needed?
Claude PR reviewer is being ignored maybe because of the label needing to be added.

### DIFF
--- a/.github/workflows/pr-reviewer.yml
+++ b/.github/workflows/pr-reviewer.yml
@@ -3,11 +3,59 @@ name: PR Review with Progress Tracking
 # This workflow triggers an AI-powered code review when the 'AI Review' label is added to a pull request.
 on:
   pull_request:
-    types: [labeled]
+    types: [opened]
 
 jobs:
+  # Gate job: resolve org membership via the API rather than trusting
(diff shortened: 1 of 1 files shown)


nclsndr:
Is this PR still relevant @cache-your-dreams?

cache-your-dreams:
> Is this PR still relevant @cache-your-dreams?

It might be in the future if we want Claude to review organisation members automatically but not community. I'll convert it to draft for the moment.

PR #26672, opened 2026-06-16
feat: bump vite to 8.0.16 and typescript to 5.9.3

Kirskov:
### What does it do?

Bumps `vite` 5.4.21 → 8.0.16 in `@strapi/strapi` and `@strapi/admin` (the package that drives the admin panel's dev server and production build), along with `@vitejs/plugin-react-swc` 3.11.0 → 4.3.1 (required for Vite 8 support). This transitively replaces Rollup+esbuild with Rolldown+Oxc for the admin build, and unifies `esbuild` on `0.27.7` across the whole dependency tree (it was previously split between `0.21.5` and `0.27.7`).

`typescript` is bumped 5.4.5 → 5.9.3 repo-wide, since `@vitejs/plugin-react-swc@4.3.1`'s type declarations use a string-literal export name syntax that TypeScript 5.4 cannot parse.

The TypeScript bump's stricter diagnostics also surfaced two small pre-existing logic bugs, fixed in a separate commit:
- `CellContent.tsx`: `attribute.type === ('relation' || 'component')` always evaluated to `=== 'relation'` (`||` short-circuit), so empty-cell padding was silently never applied for `component`-type fields.
- `sanitize.ts`: a recursive `wrappedSanitize` function had an implicit-`any` return type that TS 5.9 now flags at a different line than the existing `@ts-expect-error` covered; given an explicit return type annotation instead.

### Why is it needed?

**Security:** the old `esbuild@0.21.5` pulled in via `vite@5` is affected by [GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99) — the esbuild dev server accepts cross-origin requests by default, allowing any website a developer visits to send requests to it and read the response (fixed upstream in esbuild 0.25.0). This bump moves the admin dev server fully onto `esbuild@0.27.7`, which includes the fix.

Beyond the security fix, this keeps the admin build toolchain current and picks up the Rolldown-based build, which is meaningfully faster (observed ~24s → ~2s for the admin production build in local testing). The TypeScript bump is a hard requirement for the Vite 8 plugin chain to type-check at all.

### How to test it?

- `yarn install`
- `yarn build` / `yarn test:ts` / `yarn test:unit` should all pass
- `cd examples/getstarted && yarn develop` — admin panel should boot normally with HMR working at `http://localhost:1337/admin`
- `cd examples/getstarted && yarn build` — should produce `.strapi/client/index.html` + `app.js` as before

### Related issue(s)/PR(s)

N/A

---
Fixes CPR-414


Kirskov:
@innerdvations I didn't see your PR, we can close this one if my fix is a duplicate

innerdvations:
@Kirskov Yes, this one I think we will have to handle internally because it will involve breaking changes for a small portion of plugins and user code. I will close this for now, but I do want to investigate what you did here, since it seems there are several differences and you may have had some ideas that I didn't. Thanks for the work!

Issue #15634, opened 2023-01-30
Preventing race conditions with CRON jobs


Cpaczek:
As requested by @petersg83 https://github.com/strapi/strapi/pull/15590 i've opened a new issue to discuss the best way to prevent race conditions with CRON jobs for horizontally scaled Strapi instance. 

Currently, all CRON jobs will fire on all Strapi instances no matter what. The best way to solve this would be to push Cron jobs to a table and lock that table, when a Cron needs to run only 1 instance will get the lock and can run the code. I could probably work on this in a couple of weeks after I hear input about this implementation and nail down the details a bit more.

sjoukedv:
I agree a (table) lock would be a resolution for the race condition.

It would be great if we can add another flag that specifies if the cron job should run once across all instances or only once, e.g. by storing the date of the latest run and a boolean column that stores if the job did run already. 

```
IF !select(cron_date, did_run) || allow_multiple_crons DO 
  // <-- remove previous cron_date, did_run entry from table
  // <-- do task
```

Marc-Roig:
Looking a bit on how to lock tables, seems mysql discourages to do that when using transactions:
> LOCK TABLES is not transaction-safe and implicitly commits any active transaction before attempting to lock the tables.

Also, it seems knex does not support locking tables straight out of the box either. Maybe with some raw quereis it could be possible.

> It would be great if we can add another flag that specifies if the cron job should run once across all instances or only once, e.g. by storing the date of the latest run and a boolean column that stores if the job did run already.
That would be a nice feature indeed, I wonder if databases have some sort of timed alarm that we could use too.

derrickmehaffy:
> Looking a bit on how to lock tables, seems mysql discourages to do that when using transactions:
> > LOCK TABLES is not transaction-safe and implicitly commits any active transaction before attempting to lock the tables.
> 
> Still, there might be plenty of other mechanisms we could do. I agree this could be a good way.
> 
> > It would be great if we can add another flag that specifies if the cron job should run once across all instances or only once, e.g. by storing the date of the latest run and a boolean column that stores if the job did run already.
> 
> 
> That would be a nice feature indeed, I wonder if databases have some sort of timed alarm that we could use too. 
> 

Most do yes, albeit it I'm not sure we would add it natively but this is specifically what software like Redis was designed for: https://redis.com/redis-best-practices/communication-patterns/redlock/

As for the DB crons:
https://github.com/citusdata/pg_cron
[...]

Issue #3976, opened 2019-09-12
How does strapi run in cluster mode?


fouber:
I did not find any explanation of cluster mode in the documentation and code, and I also reported an error when starting multiple instances using pm2, I did not know how to run a strapi instance of cluster mode. I wonder if strapi can run multiple processes based on the number of cpus

kindziora:
Hi @fouber maybe this way: http://pm2.keymetrics.io/docs/usage/cluster-mode/

fouber:
@kindziora strapi startup by ``npm run start``. it can not be started by pm2 cluster mode

derrickmehaffy:
Create a server.js file:
```
#!/usr/bin/env node
'use strict';

// Start Strapi
const strapi = require('strapi');
strapi().start();
```

Run in cluster mode: `pm2 start --name="mystrapiapp" server.js -i 2`

You can change the `-i` to the number of instances (CPU cores you have on the machine is usually the safest bet in which case `-i max`)

derrickmehaffy:
Marking this issue as closed as clustering does work and is fairly simple to implement (I've been doing it for the better part of a year and a half now). If you have questions please do feel free to ask.

derrickmehaffy:
@fouber I disagree simply because Strapi's documentation really can't handle every possible way to run the application. There are at least 4 or 5 ways to run as a PM2 cluster (some without a server.js) and the pm2 documentation as well as their github contains a lot of useful information.

Overloading the documentation is almost as bad as not having enough information. For every part of the documentation that is added, it must also be maintained. And parts that are not under Strapi's control and development shouldn't be their responsibility to keep up to date. There are a few cases especially related to AWS related stuff and I have already seen this issue come up (heck even 2 weeks after the docs were added on it).

I personally believe promoting users to search for themselves (just as I did for pm2 docs).

stanleyyuenyiu:
@derrickmehaffy In cluster mode

im facing the same issue for many people,

strapi needed to deploy docker/kubernetes,

there is no persistent storage at all, everything will be reset to docker image default when restart

so when user create a new content type,

it will generate code for the API,

but if somehow we restart the docker, everything are gone,

how do we solve?

derrickmehaffy:
@stanleyyuenyiu open up an issue on the [docker repo](https://github.com/strapi/strapi-docker)

markotitel:
Sorry for reviving the thread.
[...]

PR #26676, opened 2026-06-16
feat(server): built-in Node.js cluster mode with primary-worker deduplication

m-zandi:
## What does it do?
Adds opt-in built-in Node.js cluster support to Strapi v5. When
`server.cluster.enabled: true` is set in config, Strapi forks N worker
processes (default: one per CPU core). Only the primary process runs
cron jobs and webhook delivery. Workers handle all HTTP traffic,
load-balanced by the OS.

## Why is it needed?
`server/index.ts` mounts the admin panel, content API, cron jobs,
webhook runner, and uploads all on one Koa app in one OS process.
Running multiple external instances (pm2, Kubernetes replicas) causes
cron tasks to fire on every instance simultaneously and webhooks to
be delivered multiple times — producing duplicate emails, duplicate
webhook calls, and race conditions with no built-in solution.

## Changes
- `packages/core/core/src/configuration/index.ts` — added
  `server.cluster` defaults: `{ enabled: false, workers: 0 }`.
  `workers: 0` means use `os.cpus().length` automatically.

- `packages/core/core/src/services/cluster.ts` — new service.
  When `cluster.enabled` is true, the primary process forks N workers
  and auto-restarts any that crash. Exports `isPrimary()` helper that
  checks both `cluster.isPrimary` (built-in fork) and the
  `STRAPI_CLUSTER_PRIMARY` env variable (external pm2/K8s instances).

- `packages/core/core/src/Strapi.ts` — added import of
  `createClusterService` and `isPrimary` from `./services/cluster`,
  and registered the cluster service via
  `this.add('cluster', () => createClusterService(this))` inside
  `registerInternalServices()`, consistent with all other services.

- `packages/core/core/src/services/cron.ts` — added `isPrimary()`
  guard at the top of the `start()` method. Non-primary workers log
  a message and return early without scheduling any jobs.

- `packages/core/core/src/services/webhook-runner.ts` — added
  `isPrimary()` guard at the top of the `add()` method. Non-primary
  workers return early before `createListener()` is called, which
  prevents `this.eventHub.on(event, listen)` from ever registering
  on worker processes.

## Breaking changes
None. `server.cluster.enabled` defaults to `false`. All existing
single-process deployments are completely unaffected.

## Related
Related community feedback:
https://feedback.strapi.io/other/p/featserver-built-in-nodejs-cluster-mode-all-services-share-one-process-impossibl
Related: #3976 (cluster mode question, closed 2019)
Related: #15634 (cron race condition, open/stale, no PR merged)

## Test plan
1. Set `server.cluster.enabled: true`, `server.cluster.workers: 2`
   in `config/server.ts`
2. Start Strapi — confirm 2 worker processes appear in `ps aux`
3. Configure a cron task — confirm it fires only once per schedule
   across both workers
4. Kill one worker process manually — confirm it restarts automatically
5. Set `server.cluster.enabled: false` — confirm behaviour is
   identical to the current single-process mode

---
Fixes CPR-415

### DIFF
--- a/packages/core/core/src/configuration/index.ts
+++ b/packages/core/core/src/configuration/index.ts
@@ -23,6 +23,10 @@ const defaultConfig = {
     host: process.env.HOST || os.hostname() || 'localhost',
     port: Number(process.env.PORT) || 1337,
     proxy: false,
+    cluster: {
+            enabled: false,
+            workers: 0,  // 0 = use os.cpus().length automatically
+              },
     cron: { enabled: false },
     admin: { autoOpen: false },
     dirs: { public: './public' },
--- a/packages/core/core/src/Strapi.ts
+++ b/packages/core/core/src/Strapi.ts
@@ -18,6 +18,8 @@ import createStrapiFs from './services/fs';
 import createEventHub from './services/event-hub';
 import { createServer } from './services/server';
 import { createReloader } from './services/reloader';
+import { createClusterService, isPrimary } from './services/cluster';
+
 
 import { providers } from './providers';
 import createEntityService from './services/entity-service';
@@ -299,6 +301,7 @@ class Strapi extends Container implements Core.Strapi {
         );
       })
       .add('reload', () => createReloader(this))
+      this.add('cluster', () => createClusterService(this));
       .add('content-source-maps', () => createContentSourceMapsService(this));
   }
 
--- a/packages/core/core/src/services/cron.ts
+++ b/packages/core/core/src/services/cron.ts
@@ -1,6 +1,8 @@
 import { Job, Spec } from 'node-schedule';
 import { isFunction } from 'lodash/fp';
 import type { Core } from '@strapi/types';
+import { isPrimary } from './cluster';
+
 
 interface JobSpec {
   job: Job;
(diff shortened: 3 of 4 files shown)


packages/core/core/src/configuration/index.ts
```diff
@@ -18,54 +18,32 @@ dotenv.config({ path: process.env.ENV_PATH });
...
-      },
-      cache: {
+const defaultConfig = {
+  server: {
+    host: process.env.HOST || os.hostname() || 'localhost',
+    port: Number(process.env.PORT) || 1337,
+    proxy: false,
+    cluster: {
+            enabled: false,
+            workers: 0,  // 0 = use os.cpus().length automatically
```
Kirskov, line 28: Hi, shouldn't it be cpus().length - 1 if greater than 1 ? for production case the OS needs cpu too.

innerdvations:
Thanks for the contribution — looking at the branch tip (`0d97c598ef`), the PR description and the imports in `Strapi.ts` / `cron.ts` / `webhook-runner.ts` all assume a `packages/core/core/src/services/cluster.ts` (forking + `isPrimary`), but that file doesn’t appear to be in any of the commits on this branch.

Could you push that file (or adjust the imports if the implementation lives elsewhere)? Happy to take another look once it’s on the branch.

_AI-assisted comment._

PR #26678, opened 2026-06-16
fix: avoid buffering large uploads for MIME detection

### LINKED ISSUE #26674: Large video file uploads appear to fail with "Failed to detect MIME type from file" & "File size is greater than 2 GiB", but the upload is present
### Node Version

22.22.3

### Package Manager

npm

### Package Manager Version

10.9.8

### Strapi Version

5.46.0

### Operating System

Docker/Podman/LXC

### Database

MySQL

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When uploading a large mp4 video file, the upload appears to fail once the file has been sent with (in my case for a 31.25GB file):

```
Failed to detect MIME type from file: Failed to read file: File size (31245223794) is greater than 2 GiB
```

I can see the file in `/tmp` on the Strapi server, shortly before it's removed:

```
/opt/app # ls -lh /tmp/
/tmp/:
total 29G    
-rw-r--r--    1 node     node       29G Jun 16 12:51 i5mj3z6jjumjacrwyk8ce7fxn
```

The upload UI shows an error:

Image

 BUT if you reload the Media Library page, the video is present:

Image

The error appears to come from [`validateFile`](https://github.com/strapi/strapi/blob/d30c502caf37e22ad76f2db3d8e41c968d5f9064/packages/core/upload/server/src/utils/mime-validation.ts#L232) in `packages/core/upload/server/src/utils/mime-validation.ts` which in turn uses [`detectMimeType`](https://github.com/strapi/strapi/blob/d30c502caf37e22ad76f2db3d8e41c968d5f9064/packages/core/upload/server/src/utils/mime-validation.ts#L34) which uses `import('file-type')` which I guess is subject to the [2GB size limitation on file reads](https://smoores.dev/post/overcoming_io_limits/)?

### Steps to Reproduce

1. Tune Strapi (and nginx if using) to allow uploads larger than 2GB
2. Upload a large video file (say 2.5GB) to the Media Library
3. See that the UI reports it as failed
4. Reload the Media Library - the file will be present

### Expected Behavior

When large files have been configured for upload in the Strapi config, detecting MIME type should not fail on files larger than 2GB.

### Logs

```shell
[2026-06-16 12:11:10.006] warn: Failed to detect MIME type from file: Failed to read file: File size (31245223794) is greater than 2 GiB
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

singhvishalkr:
### What does it do?

Reads only the first 4100 bytes needed for upload MIME detection when a file path is available, instead of loading the full temporary file into memory before slicing it. The file handle is closed in a `finally` block so failed reads do not leak descriptors.

It also updates the MIME validation tests to mock the file-handle read path and adds coverage that large files are only read through the detection window.

### Why is it needed?

Large uploads can exceed Node's buffer limits when MIME detection calls `readFile()` on the complete file. MIME probing only needs the header bytes, so reading the whole payload can fail before validation has a chance to complete.

### How to test it?

- `corepack yarn install --immutable`
- `corepack yarn workspace @strapi/utils build:code`
- `corepack yarn workspace @strapi/upload test:unit server/src/utils/__tests__/mime-validation.test.ts --runInBand`
- `corepack yarn workspace @strapi/upload build:code`
- `corepack yarn eslint packages/core/upload/server/src/utils/mime-validation.ts packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts`
- `git diff --check`

Note: `corepack yarn workspace @strapi/upload test:ts:back` currently requires additional internal declaration packages to be built first; in this fresh checkout it stops on existing workspace type-resolution errors before reaching this change.

### Related issue(s)/PR(s)

Fixes #26674

### DIFF
--- a/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
+++ b/packages/core/upload/server/src/utils/__tests__/mime-validation.test.ts
@@ -1,4 +1,4 @@
-import { readFile } from 'node:fs/promises';
+import { open } from 'node:fs/promises';
 import { fileTypeFromBuffer } from 'file-type';
 import {
   detectMimeType,
@@ -15,8 +15,24 @@ jest.mock('file-type', () => ({
   fileTypeFromBuffer: jest.fn(),
 }));
 
-const mockReadFile = jest.mocked(readFile);
+const mockOpen = jest.mocked(open);
 const mockFileTypeFromBuffer = jest.mocked(fileTypeFromBuffer);
+
+const mockFileRead = (content: Buffer | string) => {
+  const source = Buffer.isBuffer(content) ? content : Buffer.from(content);
+  const fileHandle = {
+    read: jest.fn(async (buffer: Buffer, offset: number, length: number) => {
+      const bytesRead = source.copy(buffer, offset, 0, Math.min(source.length, length));
+
+      return { bytesRead, buffer };
+    }),
+    close: jest.fn().mockResolvedValue(undefined),
+  };
+
+  mockOpen.mockResolvedValue(fileHandle as any);
+
+  return fileHandle;
+};
 
 const mockStrapi = {
   log: {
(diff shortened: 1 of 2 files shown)


innerdvations (approved):
Very nice, thanks!

PR #26679, opened 2026-06-16
Add support for `initiallySelectedAssets` on the MediaLibraryDialog

Link2Twenty:
### What does it do?

Add support for setting `initiallySelectedAssets` on the MediaLibraryDialog, this make developing custom components with upload capabilities much easier. 

### Why is it needed?

Custom upload components currently don't ever show selected items initially breaking user flows.

### How to test it?

Make a custom input and grab the media-library from useStrapiApp components.

```js
const components = useStrapiApp('MyCustomComponent', (state) => state.components);
const MediaLibraryDialog = components['media-library'];
```

This component can now take an array of assets to show what is already selected on initial dialog render.

### Related issue(s)/PR(s)

N/A


boazpoolman (approved):
LGTM!

Issue #26571, opened 2026-06-08
[Bug]: Session/idle-timeout redirect to `/auth/login` silently discards unsaved Content Manager edits


thematrixdev:
### Node Version

24.11.1

### Package Manager

pnpm

### Package Manager Version

11.5.0

### Strapi Version

5.47.1

### Operating System

Linux (Debian/Ubuntu)

### Database

MariaDB

### Javascript or Typescript

Javascript

### Reproduction URL

(none — reproducible on a fresh `npx create-strapi-app` *project with only the session-lifespan config below)*

### Bug Description

With the session-based admin authentication (`config/admin` → `auth.sessions`, `accessTokenLifespan` / `idleSessionLifespan` / `maxSessionLifespan`), an admin who is **actively filling out a Content Manager edit form** can be force-logged-out mid-edit, and **all unsaved form data is silently destroyed**.

Root cause is that "idle"/session activity is keyed to **token + HTTP request activity**, not to user input. Typing into a form fires no HTTP request, so a user who is demonstrably active (typing a long entry) is treated as idle. When the access token's `exp` elapses:

* the `useIdleSessionLogout` one-shot timer fires and the **active tab** is navigated to `/auth/login` (this active-tab redirect was added as the fix for strapi/strapi#26163), **or**
* the next save returns `401`, the refresh attempt fails (idle/absolute refresh lifespan exceeded), and the active tab is redirected to `/auth/login`.

In both paths the Content Manager form state is in-memory React state, so the redirect throws it away. The existing "unsaved changes" navigation guard does **not** intercept this programmatic session-expiry redirect, and there is **no client-side draft persistence**, so there is no recovery.

Important: simply increasing the lifespans does **not** fix this — it only lowers the probability. `maxSessionLifespan` is a hard ceiling that no amount of typing can extend, so a long-enough data-entry session will eventually be cut off mid-edit regardless of activity, with the same data loss.

### Steps to Reproduce

1. Create a fresh Strapi v5 project: `npx create-strapi-app my-app`.
2. In `config/admin.ts`, set short session lifespans so the issue is quick to observe:

   ```ts
   export default ({ env }) => ({
     auth: {
       secret: env('ADMIN_JWT_SECRET'),
       sessions: {
         accessTokenLifespan: 60,
         maxRefreshTokenLifespan: 120,
         idleRefreshTokenLifespan: 60,
         maxSessionLifespan: 120,
         idleSessionLifespan: 60,
       },
     },
   });
   ```
[...]

PR #26680, opened 2026-06-16
fix(admin): admin users logged out mid-session by access-token expiry timer

### LINKED ISSUE #26677: An idle/abandoned admin tab logs out all other active tabs at the access token's `exp`
### Node Version

20.20.2

### Package Manager

yarn

### Package Manager Version

1.22.22

### Strapi Version

5.48.0

### Operating System

Linux (Debian/Ubuntu)

### Database

PostgreSQL

### Javascript or Typescript

Javascript

### Reproduction URL

*No response*

### Bug Description

In the admin panel, an idle (or abandoned) browser tab will force a logout across **all** open admin tabs when its access token reaches `exp` — even if you are actively working in another tab. The active tab is redirected to `/auth/login` mid-task.

This is the cross-tab counterpart to strapi/strapi#26571. That issue covers the *single-tab* case (typing past `accessTokenLifespan` with no API calls → logged out, unsaved edits lost). This report is about a different root cause: **logout is broadcast across tabs, but token refresh is not**, so the least-active tab effectively controls the session lifetime of every tab.

The behavior was introduced by [#26165](<https://github.com/strapi/strapi/issues/26165>) (`fix(admin): redirect active tab to login on session expiry`, commit `56d7a64c49`), which first shipped in **5.47.0**. It builds on the session model from the "Advanced Session Configuration" feature ([#24346](<https://github.com/strapi/strapi/issues/24346>)).

### Steps to Reproduce

1. Configure (or rely on the defaults for) admin session lifespans in `config/admin.ts` — the default `accessTokenLifespan` is `30 * 60` (30 min).
2. Log into the admin panel.
3. Open the admin in **two** browser tabs (Tab A and Tab B).
4. Work continuously in **Tab A** (e.g. edit content, trigger API calls so the access token transparently refreshes on 401).
5. Leave **Tab B** open but idle — do not interact with it.
6. Wait until the access token that was current when Tab B last loaded reaches its `exp` (≈30 min after the last full login by default).

### Expected Behavior

Because you have been continuously active in Tab A, the session should remain alive and you should stay logged in. An idle background tab should not be able to terminate an active session.

At the original token's `exp`, **Tab B's** idle timer fires and logs out. That logout is broadcast to **Tab A**, which is immediately redirected to  <br>`/auth/login` while you are still working. In practice you get logged out roughly every `accessTokenLifespan` (≈30 min) after the last full login, regardless of activity, whenever a stale tab is left open.

### Additional information

Cross-tab communication uses two `localStorage` keys and a single `storage` listener:

```ts
// packages/core/admin/admin/src/reducer.ts
const STORAGE_KEYS = { TOKEN: 'jwtToken', STATUS: 'isLoggedIn' };
```

```ts
// packages/core/admin/admin/src/features/Auth.tsx
const handleUserStorageChange = (event: StorageEvent) => {
  if (event.key === STORAGE_KEYS.STATUS && event.newValue === null) {
    clearStateAndLogout();
  }
};
```

The proactive idle timer (`useIdleSessionLogout`) arms a one-shot `setTimeout` at the access token's `exp` and calls `clearStateAndLogout()` when it fires:

```ts
// packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts
const expiry = decodeAccessTokenExpiry(token);
const msUntilExpiry = expiry - Date.now() + SESSION_EXPIRY_BUFFER_MS;
const timeoutId = window.setTimeout(() => onExpired(), Math.max(msUntilExpiry, 0));
```

The asymmetry that causes the bug:

| Event in Tab A | What it writes | Does an idle Tab B react? |
| -- | -- | -- |
| **Token refresh** (401 → `storeToken`) | updates Tab A's Redux via `onTokenUpdate`; writes `jwtToken` only in rememberMe mode | **No** — the `storage` listener only watches `isLoggedIn`, never `jwtToken`, so Tab B's Redux `token` never changes and its timer is never re-armed |
| **Logout** (`logoutAction`) | removes `isLoggedIn` from `localStorage` | **Yes** — every other tab's listener fires `clearStateAndLogout()` |

So a refresh in the active tab keeps only that tab alive; it does **not** re-arm the idle tab's timer. When the idle tab's stale timer eventually fires, its logout *is* broadcast — taking the active tab down with it.

The relevant chain on the idle tab:

```
Tab B timer fires (original exp)
  → onExpired() = clearStateAndLogout()
  → dispatch(logoutAction())
  → reducer.logout() removes `isLoggedIn` from localStorage
  → `storage` event in Tab A
  → handleUserStorageChange sees isLoggedIn === null
  → Tab A clearStateAndLogout() → redirect to /auth/login
```

Note that `setToken` (used on refresh) only mutates Redux and never writes `isLoggedIn`, while `storeToken` writes `jwtToken` (and only in rememberMe mode), which the listener ignores. Refreshes therefore never propagate cross-tab, but logouts always do.

## Suggested fix

Make refresh/expiry propagate across tabs symmetrically with logout, or stop trusting a purely local stale timer to terminate the shared session. Options:

* On token refresh, write the new token/`exp` to a watched `localStorage` key so idle tabs re-arm their `useIdleSessionLogout` timer instead of firing on a stale `exp`.
* Before an idle tab broadcasts a logout, have it attempt a refresh / confirm with the server, so a still-valid session in another tab isn't killed by a  <br>background tab's local timer.

## Workarounds

* Increase `admin.auth.sessions.accessTokenLifespan` in `config/admin.ts` to widen the timer window (reduces frequency; does not fix the asymmetry).
* Avoid leaving stale/abandoned admin tabs open; closing a tab removes its timer.

### Confirmation Checklist

travelton:
### What does it do?

Reworks the admin panel's idle-session timer so it keeps **active** sessions alive instead of logging users out when the short-lived access token expires, and stops one tab from logging out another.

Changes (admin frontend only):

- **`packages/core/admin/admin/src/hooks/useIdleSessionLogout.ts`** — The hook no longer treats access-token `exp` as a hard logout. It now arms **two** timers per token:
  - a **renew** timer that fires *ahead* of expiry: if the user interacted during the token's lifetime, it silently renews the session (`renewSession`); a server-rejected renewal ends the session globally (`onSessionDead`). Renewing *before* expiry guarantees the refreshed token is in shared storage before any sibling tab's logout timer fires.
  - a **logout** timer just after `exp` (fallback for tabs that didn't renew): it first re-syncs from a token another tab refreshed (`getStoredToken`/`onResync`), renews if activity arrived late, and only otherwise logs out — **locally**, so an idle tab can't force-logout an active one.
- **`packages/core/admin/admin/src/features/Auth.tsx`** — Wires it up:
  - tracks last user interaction (`pointerdown`/`keydown`/`click`/`scroll`/`touchstart`) as the activity signal;
  - `renewSession` rotates the refresh token via the existing `attemptTokenRefresh` (the resulting `setToken` re-arms the hook);
  - adds `clearLocalSessionAndRedirect` (local, no `logoutAction`) for idle logout, while `clearStateAndLogout` (global) stays for user-initiated logout and server-confirmed 401s (`setOnSessionExpired`).

Tests: `useIdleSessionLogout.test.ts` expanded to 15 cases covering silent renewal on activity, idle logout, server-rejected renewal (global), cross-tab re-sync, re-sync-vs-renew precedence, and the legacy/disabled/malformed paths.

### Why is it needed?

The proactive idle timer introduced in #26165 logged users out when the access token hit its `exp`, rather than refreshing it. The token is only renewed *reactively* on a 401, which produced two failures:

1. **Single active tab (#26571):** while the user works, requests succeed with the still-valid token, so nothing refreshes it — and at `accessTokenLifespan` (default 30 min) the timer logs them out mid-task, discarding unsaved work.
2. **Multiple tabs (#26677):** the logout cleared the shared `isLoggedIn` key, whose `storage` event logs out *every* tab. Refreshes in an active tab only updated Redux locally and never propagated, so an idle/abandoned tab fired on its stale token and tore down sessions the user was actively using.

The access token is meant to be a short refresh cadence, not the session length — the server's refresh-token idle/max windows are the real session bounds. This change makes the client behave accordingly: keep active sessions alive, end only genuinely idle ones, and never let one tab speculatively log out another.

### How to test it?

Environment: 5.47+ admin using session/JWT auth. To make the timers fast, set a short lifespan in `config/admin.{js,ts}`:

```js
  auth: {
    sessions: { accessTokenLifespan: 60 }, // 60s for testing
  },
});
```

Run the example app (`examples/getstarted`): `yarn develop --watch-admin` (+ `yarn watch` in `packages/core/admin`). Log in, then:

1. **Single tab, active** — interact (click/scroll/type) at least once per window, wait past 60s → **stays logged in**; a `POST /admin/access-token` fires ~10s before each expiry.
2. **Single tab, idle** — don't touch it for ~70s → **redirected to `/auth/login`** (idle logout still works).
3. **Two tabs** — keep interacting in Tab A, leave Tab B idle, wait past 60s → **both stay logged in** (B re-syncs from A's refresh); neither is kicked out.
4. **Explicit logout / server rejection** — clicking Logout, or a real 401 after the session truly ends, logs out **all** tabs.

Automated:

```bash
nvm use 22
cd packages/core/admin
yarn test:front useIdleSessionLogout   # 15 passing
```

### Behavior change

Idle logout is now driven by real user interaction rather than by access-token expiry alone:

- **Before:** the user was logged out exactly when the access token expired (default 30 min), regardless of whether they were active.
- **After:** as long as the user interacts at least once per `accessTokenLifespan`, the session is silently renewed and they stay logged in. A user who stops interacting is logged out after one idle `accessTokenLifespan`.

The server's `idleSessionLifespan` / `maxSessionLifespan` / refresh-token windows remain the hard caps: a renewal past those is rejected and ends the session globally. Net effect on the configured *idle* window: with the defaults (`accessTokenLifespan` 30 min, `idleSessionLifespan` 2 h), effective idle logout becomes ~30 min of inactivity instead of 2 h — i.e. stricter, never longer than configured. If you rely on the longer idle window, align `accessTokenLifespan` accordingly.

### Related issue(s)/PR(s)

- Fixes #26677 — idle/abandoned tab logs out other active tabs (cross-tab cascade).
- Refs #26571 — single-tab logout during inactivity. Active editing no longer triggers a logout (typing counts as activity, so the session is renewed); the separate concern of the redirect bypassing the unsaved-changes navigation guard is **not** addressed here.
- Regression introduced by #26165 (`fix(admin): redirect active tab to login on session expiry`).
- Builds on the session model from #24346 (Advanced Session Configuration).

---
Fixes CPR-416

### DIFF
--- a/packages/core/admin/admin/src/features/Auth.tsx
+++ b/packages/core/admin/admin/src/features/Auth.tsx
@@ -8,7 +8,13 @@ import { useTypedDispatch, useTypedSelector } from '../core/store/hooks';
 import { useStrapiApp } from '../features/StrapiApp';
 import { useIdleSessionLogout } from '../hooks/useIdleSessionLogout';
 import { useQueryParams } from '../hooks/useQueryParams';
-import { login as loginAction, logout as logoutAction, setLocale, setToken } from '../reducer';
+import {
+  getStoredToken,
+  login as loginAction,
+  logout as logoutAction,
+  setLocale,
+  setToken,
+} from '../reducer';
 import { adminApi } from '../services/api';
 import {
   useGetMeQuery,
@@ -19,7 +25,11 @@ import {
 } from '../services/auth';
 import { normalizeAdminLocale } from '../translations/normalizeAdminLocale';
 import { getOrCreateDeviceId } from '../utils/deviceId';
-import { setOnSessionExpired, setOnTokenUpdate } from '../utils/getFetchClient';
+import {
+  attemptTokenRefresh,
+  setOnSessionExpired,
+  setOnTokenUpdate,
+} from '../utils/getFetchClient';
 
 import type {
   Permission as PermissionContract,
@@ -141,6 +151,66 @@ const AuthProvider = ({
     navigate('/auth/login');
   }, [dispatch, navigate]);
 
+  /**
+   * Clear *only this tab's* session and redirect to login, without removing the
+   * shared `isLoggedIn`/`jwtToken` storage keys. Unlike `clearStateAndLogout`,
+   * this does not dispatch `logoutAction`, so it does not fire the cross-tab
+   * `storage` event that logs every tab out. Used by the speculative idle timer,
+   * which fires per-tab at the access token's `exp`: an idle tab must never tear
+   * down a session that another tab is still actively using. A genuine logout
+   * (user-initiated, or a server-confirmed 401 via `setOnSessionExpired`) still
+   * goes through `clearStateAndLogout` and broadcasts to all tabs.
+   */
+  const clearLocalSessionAndRedirect = React.useCallback(() => {
+    dispatch(adminApi.util.resetApiState());
+    dispatch(setToken(null));
+    navigate('/auth/login');
+  }, [dispatch, navigate]);
+
+  const resyncToken = React.useCallback(
+    (newToken: string) => {
+      dispatch(setToken(newToken));
+    },
+    [dispatch]
+  );
+
+  /**
+   * Track the timestamp of the user's last interaction. This is the activity
+   * signal `useIdleSessionLogout` uses to tell an active user (silently renew
+   * their session) apart from a genuinely idle one (log out). Successful API
+   * calls don't refresh the short-lived access token on their own, so without
+   * this an actively-working user is logged out the moment the token expires.
+   */
+  const lastActivityRef = React.useRef(Date.now());
+  React.useEffect(() => {
+    const markActivity = () => {
+      lastActivityRef.current = Date.now();
+    };
+    const events = ['pointerdown', 'keydown', 'click', 'scroll', 'touchstart'];
+    events.forEach((event) =>
+      window.addEventListener(event, markActivity, { passive: true, capture: true })
+    );
+
+    return () => {
+      events.forEach((event) => window.removeEventListener(event, markActivity, { capture: true }));
+    };
+  }, []);
+  const getLastActivityAt = React.useCallback(() => lastActivityRef.current, []);
+
+  /**
+   * Silently rotate the refresh token and mint a new access token. Returns
+   * `true` on success (the resulting `setToken` re-arms the idle timer) and
+   * `false` when the server rejects it (session genuinely over).
+   */
+  const renewSession = React.useCallback(async () => {
+    try {
+      await attemptTokenRefresh();
+      return true;
+    } catch {
+      return false;
+    }
+  }, []);
+
   React.useEffect(() => {
     if (user) {
       if (user.preferedLanguage) {
@@ -180,13 +250,21 @@ const AuthProvider = ({
   }, [clearStateAndLogout]);
 
   /**
-   * Proactive idle-session detection. See `useIdleSessionLogout` for the full
-   * rationale and the UX trade-off (a user actively typing in a form with no
-   * outgoing API calls will still be logged out at `exp`).
+   * Session lifecycle at access-token expiry. The timer fires per-tab; the hook
+   * then either (1) re-syncs from a token another tab refreshed, (2) silently
+   * renews when the user has been active, or (3) logs out. Idle logout is local
+   * so an idle tab can't force-logout an actively-used one; a server-confirmed
+   * dead session (renewal rejected) broadcasts globally. See
+   * `useIdleSessionLogout` for the full rationale.
    */
   useIdleSessionLogout({
     token,
-    onExpired: clearStateAndLogout,
+    onExpired: clearLocalSessionAndRedirect,
+    onSessionDead: clearStateAndLogout,
+    getStoredToken,
+    onResync: resyncToken,
+    getLastActivityAt,
+    renewSession,
     disabled: _disableRenewToken,
   });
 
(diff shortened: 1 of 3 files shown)


innerdvations (approved):
Nice fix, thanks!

jhoward1994:
Hi, thanks for this! I QAed in the getstarted example app

| Test | Result |
|------|--------|
| Single tab, active (use interaction within the accessTokenLifespan) → stays logged in + silent renew | ✅ |
| Single tab, idle → local logout ~61s | ✅ |
| Two tabs, active keeps idle tab alive | ✅ |
| Server rejects renewal → global logout both tabs (e.g. with idleSessionLifespan <  accessTokenLifespan | ✅ |

Tested with `accessTokenLifespan: 60`. (and `idleSessionLifespan` 45 when needed for test scenario)

jhoward1994 (approved):

travelton:
Thanks for the reviews and additional eyes on testing. It's quite the annoyance to all of our users, excited to hear it'll potentially drop in 5.48.2! First time contributing... LMK if I need to do anything else here to get this merged (I don't appear to have permissions to merge to develop). 🍻

PR #26681, opened 2026-06-17
fix(content-manager): use top-level Core type import in MCP types

nclsndr:
### What does it do?

Import `Core` from `@strapi/types` alongside `Struct` and `Modules`, and use it for the `strapi` parameter type in `DerivedTool.createHandler` instead of the inline `import('@strapi/types').Core.Strapi`. This keeps type imports consistent and avoids the inline dynamic import.

### Why is it needed?

Type definition was making a redundant import.

### How to test it?

```
yarn build
```

### Related issue(s)/PR(s)

- ø


innerdvations (approved):
LGTM!

Issue #26543, opened 2026-06-03
Establish a consistent test-only convention for Jest globals in TS workspaces


unrevised6419:
### Problem

There is no deliberate convention for typing Jest globals (`describe`, `it`,
`expect`, `jest`, `beforeEach`, …) in TypeScript workspaces, and the current
ad-hoc setups fail in two opposite ways:

**1. Leak (present on `develop` today).**
Any workspace *without* a `types` override inherits TypeScript's default of
auto-including **every** `node_modules/@types` package into the whole program.
So not just Jest globals but *all* augmented globals leak into production
`src` — e.g. `@types/jsdom` pulls in `lib.dom`, which is how the `origin`
DOM-global bug #26542 fixes slipped through. `jest.fn()` / `describe` in a
source file also compile clean under this default.

The few workspaces that *do* scope `types` and include `"jest"` —
`packages/core/admin/server`, `packages/core/content-manager/server`,
`packages/core/content-type-builder/server` (`"types": ["lodash", "jest"]`) —
narrow the surface but still leak Jest globals into `src`.

Either way, Jest globals are valid in production source today. That is a
footgun that exists right now.

**2. Missing (latent, broadened by #26542).**
The auto-include default above is also the *only* thing giving the no-override
workspaces their Jest globals in tests. #26542 scopes the shared base to
`"types": ["node"]` (the right move — deterministic builds, no DOM/global
leakage). Once it lands, that crutch is gone and Jest globals are no longer in
scope for `*.test.ts` across the **36 workspaces** extending the base — IDEs
and `tsc` flag them as undefined (~491 test files rely on these globals
ambiently).

Same root cause, no single answer: jest typing is either leaking into source
(along with everything else under `@types`) or about to disappear from tests,
depending on the workspace.

### Goal

Jest globals available in **test files only**, production `src` kept free of
them, applied uniformly across all TS workspaces — without depending on the
"auto-include all `@types/*`" behaviour #26542 removes.

### Options

**Option 1 — `@jest/globals` (explicit imports) — recommended**
```ts
import { describe, it, expect, jest } from '@jest/globals';
```
- ✅ No ambient leak; globals scoped to importing files.
- ✅ Best-typed `jest` mock API.
- ✅ Matches the existing repo convention — e2e tests already import their
  framework explicitly: `import { test, expect } from '@playwright/test'`
  (84 files under `tests/`). Same pattern, no special tsconfig wiring.
[...]

PR #26699, opened 2026-06-18
chore(typescript): scope tsconfig types per workspace

unrevised6419:
### What does it do?

Cleans up how TypeScript/Jest types, dependencies, and tsconfig/jsconfig files are declared across the monorepo.

**1. Per-workspace tsconfig `types`**

Sets `types` to `[]` in `tsconfig/base.json` and `tsconfig/client.json`, and moves type declarations to each workspace's own tsconfig:

- back workspaces: `["node"]`, or `["node", "jest"]` when the tsconfig compiles jest tests
- front workspaces: `["@testing-library/jest-dom"]` only where the workspace uses the front Jest preset, plus `"jest"` when it compiles jest tests

**2. Jest / jest-dom dependency hygiene**

- adds the `jest` devDependency to the workspaces that have jest tests — and only those (audited: no config carries the `"jest"` type without compiling tests, no workspace carries the dep without having tests)
- `@strapi/admin-test-utils` declares its runtime libraries as **peerDependencies** — `@testing-library/jest-dom` (range corrected from a stale `^5.16.5` to `^6.0.0`), `@reduxjs/toolkit`, `jest-environment-jsdom`, `styled-components`. It registers them through its `after-env` setup, which the root `jest-preset.front.js` loads for every front workspace.
- every TypeScript workspace that uses the front preset declares `@testing-library/jest-dom` (and `@testing-library/dom`) and keeps jest-dom in its tsconfig `types`; workspaces that do **not** use the front preset (`plugin-cloud`, `plugin-graphql`, `plugin-sentry`) drop jest-dom from `types`
- removes from the repo root: unused `ts-jest` (all transforms use `@swc/jest`), redundant `jest-circus` (jest 29's default runner), and `jest-environment-jsdom` (now declared in `@strapi/admin-test-utils`)

**3. `@types/node` dependency hygiene**

- adds the `@types/node` devDependency to every back workspace whose tsconfig lists `"node"` in `types` but relied on hoisting, pinned to `20.19.41` to keep the ambient Node types Node-20-safe
- satisfies the remaining unmet peer warnings from `yarn install` (`@types/node` in the example apps, `@types/react` in `content-type-builder`)

**4. tsconfig / jsconfig structure**

- drops the deprecated `baseUrl` from all package tsconfigs (TS 6 readiness) and rewrites the few non-relative imports that relied on it
- centralizes `noEmit: true` in `tsconfig/base.json`; each `tsconfig.build.json` extends its local `tsconfig.json` and sets `noEmit: false`
- removes compilerOptions that already resolve from the extended base (`rootDir`/`outDir` in type-check configs, plus inherited `module`/`moduleResolution`/`esModuleInterop`/`sourceMap`/`allowJs`)
- adds a `jsconfig.json` to the JS-only workspaces and to top-level `scripts/` and `tests/`, and scopes the root `jsconfig.json` to exclude self-covering dirs (`packages`, `examples`, `templates`, `.github`, `docs`, `scripts`, `tests`, `.nx`)
- removes dead code: the deprecated `plugin:init` tsconfig scaffold templates and a duplicate `compress-files.test.ts`
- adds a `test:ts` target (or `test:ts:back`/`:front` for split admin/server packages) to TS workspaces that previously only type-checked via `build:types`, and makes the implicit `tsc` typecheck scripts explicit with `-p tsconfig.json`
- normalizes `tsconfig.eslint.json`: drops fields that merely mirror the parent (most reduce to `{ "extends": "./tsconfig.json" }`), keeps an `exclude` only where eslint must lint files the build skips, and removes 2 orphaned eslint tsconfigs
- trims redundant `node_modules`/`dist` from `src`-scoped excludes (tsc never pulls them into a scoped project)

Scoping `types` per workspace narrows the ambient lib/DOM/undici globals, so a few call sites that previously resolved implicitly are now typed explicitly (Fetch input as `string | URL | Request` with cast `response.json()` in ee license/trial, i18n ai-localizations, and core fetch). `vitest.config.ts` is also allowed by the eslint `import/no-extraneous-dependencies` rule.

### Why is it needed?

A shared base `types` list forces every workspace to inherit the same ambient types. Per-workspace `types` makes each package opt into exactly what it needs, and declaring deps where they're used lets eslint's `import/no-extraneous-dependencies` catch undeclared usage.

This is a **precursor** to #26543: scoping `types` per workspace is the prerequisite for a workspace to opt out of the ambient `jest` type and adopt an explicit Jest-globals convention. It does **not** switch any workspace to `@jest/globals` imports (blocked under `@swc/jest` — see the issue).

### How to test it?

- `yarn build` (per-workspace emit still works after the tsconfig restructure)
- `yarn test:ts` (per-workspace type-checking)
- `yarn test:unit` / `yarn test:front`
- `yarn install` — no missing peer-dependency warnings
- `yarn lint` / `yarn prettier:check`

### Related issue(s)/PR(s)

Precursor to #26543.

- #26459 — closed in favor of this PR
- #26542 — closed in favor of this PR
- #25873 — this PR pulls out its non-breaking changes


unrevised6419:
@nclsndr if you want I can move non breaking changes (baseUrl, missing rootDir/outDir, ...) from the https://github.com/strapi/strapi/pull/25873 here. So when that PR is merged to be mostly the version update.

unrevised6419:
Btw, this PR needs [this one](https://github.com/strapi/strapi/pull/26291) merged first, because there are TS conflicts between undici/node@24 and undici/node@20 types

nclsndr:
> @nclsndr if you want I can move non breaking changes (baseUrl, missing rootDir/outDir, ...) from the #25873 here. So when that PR is merged to be mostly the version update.

(answering with low context — did not start review)
@unrevised6419, given the size of those 2 PRs I'd even go to extract the elements worth refactoring from #25873 into a new PR from develop / this one (depending on you judgement on conflicts management mostly).

unrevised6419:
Not sure if I understood correctly what do you mean.

Most of the code changes surfaced when we shrinked the global types augmentation from `*` to `node/jest` only. If we want to extract those changes in a new PR, we will have again to do the `tsconfig.json#types` narrowing again.

The changes in this PR are mostly scoping the dependencies/tsconfig#types per workspace.

Maybe you can trigger the CI for this one also to see the results. The changes in PR that depends in this one are a few only (last commit)

PS: I can move the changes from TS migration PR here, all the tsconfig.json changes with the exception of the `moduleResolution` that changes from `node` to `bundler`.

packages/utils/tsconfig/base.json
```diff
@@ -11,6 +11,8 @@
     "noImplicitAny": true,
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "noEmit": true,
+    "types": []
```
unrevised6419: Empty `types` + `noEmit` live in the shared base on purpose: each workspace opts into exactly the ambient types it needs, and only the `*.build.json` configs opt back into emit (`noEmit: false`). This is the precursor to per-plugin type augmentation (#26543).

packages/core/content-manager/server/src/utils/index.ts
```diff
@@ -1,7 +1,7 @@
 import '@strapi/types';
 
-import { DocumentManagerService } from 'src/services/document-manager';
-import DocumentMetadata from 'src/services/document-metadata';
+import type { DocumentManagerService } from '../services/document-manager';
```
unrevised6419, line 3: Was `src/services/...`, which only resolved through the now-removed `baseUrl`. Switched to a relative path; `import type` keeps it type-only and avoids a runtime require cycle.

packages/admin-test-utils/package.json
```diff
@@ -85,15 +85,18 @@
...
+    "@testing-library/dom": "10.4.1",
     "@testing-library/jest-dom": "6.9.1",
     "eslint-config-custom": "5.48.1",
     "jest-environment-jsdom": "29.6.1",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
     "styled-components": "6.4.1",
     "tsconfig": "5.48.1"
   },
   "peerDependencies": {
```
unrevised6419, line 97: These are consumer-provided runtime libs — `admin-test-utils` registers them via its `after-env` setup for every front workspace — so they're declared as peers, not deps. The matching devDeps keep this package's own build/tests resolving.

packages/core/database/package.json
```diff
@@ -55,7 +55,10 @@
   },
   "devDependencies": {
     "@types/fs-extra": "11.0.4",
+    "@types/jest": "29.5.2",
+    "@types/node": "20.19.41",
```
unrevised6419, line 62: Pinned to `20.19.41` (not latest) so the ambient Node types stay aligned with the `engines.node >= 20` floor and remain Node-20-safe.

jsconfig.json
```diff
@@ -8,14 +8,19 @@
...
     "**/*.test.js",
     "**/*.test.api.js",
-    "**/examples/getstarted/*",
-    "**/examples/kitchensink/*",
-    "docs"
+    "docs",
+    "examples",
+    "templates",
+    ".github",
+    "packages",
```
unrevised6419, line 22: Every workspace now owns its tsconfig/jsconfig, so the root jsconfig stops scanning `packages` (and the other self-covering dirs) — avoids double-coverage and keeps the editor project fast.

packages/core/strapi/src/cli/commands/plugin/init/files/typescript.ts
```diff
@@ -1,93 +0,0 @@
-import { outdent } from 'outdent';
```
unrevised6419, line 1: Removing dead code: the `plugin:init` command is deprecated (it redirects to `npx @strapi/sdk-plugin init`) and nothing imports these scaffold templates.

unrevised6419 (commented):
Context comments on the changes most likely to prompt a "why this?" — one per category.

packages/core/permissions/tsconfig.eslint.json
```diff
@@ -1,8 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "compilerOptions": {
-    "noEmit": true
-  },
-  "include": ["src"],
-  "exclude": ["node_modules"]
+  "exclude": []
```
unrevised6419, line 3: `exclude: []` (empty, not absent) is deliberate: it overrides the parent tsconfig.json's `**/__tests__/**` exclude so eslint type-checks the test files too. Eslint configs whose parent excludes no tests omit `exclude` entirely and just inherit; this one keeps the explicit empty array to re-include tests.

packages/core/data-transfer/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "tsconfig/base.json",
   "include": ["src"],
-  "exclude": ["node_modules"]
+  "exclude": ["**/__tests__/**"],
```
unrevised6419, line 4: tsconfig.json excludes `__tests__` so the new `test:ts` target type-checks source only. The build already excludes tests via tsconfig.build.json; the test files here carry pre-existing type debt (custom jest matchers, `global.strapi`) that's left for a follow-up rather than blocking this PR.

packages/core/core/src/utils/fetch.ts
```diff
@@ -14,18 +14,21 @@ export const createStrapiFetch = (
 ): Modules.Fetch.Fetch => {
   const { logs = true } = options ?? {};
 
-  function strapiFetch(url: RequestInfo | URL, options?: RequestInit) {
-    const fetchOptions = {
-      ...(strapiFetch.dispatcher ? { dispatcher: strapiFetch.dispatcher } : {}),
+  const strapiFetch: Modules.Fetch.Fetch = (url: string | URL | Request, options?: RequestInit) => {
```
unrevised6419: With `types` no longer pulling DOM/undici globals implicitly, Fetch input is typed explicitly here; the wrapper also bridges the undici node20/node24 `RequestInit` type mismatch.

packages/core/types/package.json
```diff
@@ -40,7 +40,7 @@
     "doc:ts": "typedoc",
     "doc:ts:watch": "typedoc --watch",
     "lint": "run -T eslint .",
-    "test:ts": "run -T tsc --noEmit",
+    "test:ts": "run -T tsc -p tsconfig.json",
```
unrevised6419, line 55: Dropped the explicit `--noEmit` from these typecheck scripts: the target tsconfigs now resolve `noEmit: true` from `tsconfig/base.json`, so the flag is redundant. (The `.github` community-pr-triage typecheck keeps it — its standalone config doesn't inherit the base.)

unrevised6419:
Hey @nclsndr!
Can you trigger the CI for this PR please.
Do you see fit to extract something to a smaller PR, or revert?

Most of the changes are in tsconfig and package.json files

nclsndr:
> Hey @nclsndr! Can you trigger the CI for this PR please. Do you see fit to extract something to a smaller PR, or revert?
> 
> Most of the changes are in tsconfig and package.json files

CI triggered.
I was contemplating the idea of having smaller PRs but actually these are mostly one change over the all the 35+ packages. I'll go over the review and keep you posted.

unrevised6419:
@nclsndr please trigger CI again. I added a resolution for `@types/react` because it created some problems with `@radix-ui/*` deps that resolved to different places, and failed a lot of tests.

Also did sync with `develop`

packages/core/core/src/ee/license.ts
nclsndr: I think it's preferable to avoid blindly casting here.
Instead, make it runtime safe (even if a bit verbose)


```suggestion
    const responsePayload = await response.json();

    if (responsePayload === undefined || responsePayload === null || typeof responsePayload !== 'object') {
      throw new LicenseCheckError('Invalid license payload.');
    }

    switch (response.status) {
      case 200: {
        if (!('data' in responsePayload)) {
          throw new LicenseCheckError('Invalid license payload.');
        }
        return (responsePayload.data as { license: string }).license;
      }
      case 400: {
        if (!('error' in responsePayload)) {
          throw new LicenseCheckError('Invalid license payload.');
        }
        throw new LicenseCheckError((responsePayload.error as { message: string }).message);
      }
      case 404:
        throw new LicenseCheckError('The license used does not exists.');
      default:
        throwError();
    }
```

packages/core/core/src/utils/fetch.ts
```diff
@@ -14,18 +14,21 @@ export const createStrapiFetch = (
...
+      fetchOptions.dispatcher = strapiFetch.dispatcher;
+    }
+
     if (logs) {
       strapi.log.debug(`Making request for ${url}`);
     }
 
     return fetch(url, fetchOptions);
-  }
+  };
```
nclsndr: The spread order swapped. The old code let callers override the global proxy via request-level `options.dispatcher`. The new code makes the global proxy unconditionally win.

Previous intent:

```suggestion
  const strapiFetch: Modules.Fetch.Fetch = (url: string | URL | Request, options?: RequestInit) => {
    const fetchOptions: RequestInit = {
      ...(strapiFetch.dispatcher ? { dispatcher: strapiFetch.dispatcher } : {}),
      ...options,
    };

    if (logs) {
      strapi.log.debug(`Making request for ${url}`);
    }

    return fetch(url, fetchOptions);
  };
```

packages/core/data-transfer/tsconfig.json
nclsndr: Ideally we want the `tsconfig.json` files to include all the package. Rationale: this is what is picked up by IDEs to drive local LSP, thus it should cover all sources.
Other files should be used to drive script-based checks.
I know we have this pattern already in a couple of places. Not a hard requirement for this PR, more a direction.

nclsndr (commented):
Great job @unrevised6419 🙌 

Just a couple of details 👇

packages/core/data-transfer/tsconfig.json
unrevised6419: Got it. With this PR changes I found out some packages that did not have a `test:ts` and added it. It only had `build:types` that uses `tsconfig.build.json` which excludes the tests. For now I also excluded them here, because there are type errors.

nclsndr:
Seems like we need yarn install updating yarn.lock - lmk if you want me to takeover @unrevised6419

unrevised6419:
@nclsndr synchronized with `develop`. You can take over my PRs anytime if that speeds up the process. Btw, this PR lint may fail because a workspace lint errors surfaced after adding lint-staged per workspace.

https://github.com/strapi/strapi/pull/26800

unrevised6419:
I merged that PR branch into this one for the CI to be green. Please trigger CI

nclsndr:
> I merged that PR branch into this one for the CI to be green. Please trigger CI

@unrevised6419, do you mean you merge #26800 into this #26699? If that the case I would revert it, this PR is already huge, having that chunk on top would make review harder to reason about. We can review and merge #26800 if it is blocking.

unrevised6419:
@nclsndr yes. That needs to be merged first. Then we sync this.

nclsndr:
- #26800 is now merged

nclsndr:
currently rebasing manually

unrevised6419:
I did merge with `develop` locally, and was about to push 😄

nclsndr:
Sorry, please go ahead force push your version if you prefer.

packages/cli/cloud/package.json
```diff
@@ -65,15 +66,18 @@
     "open": "8.4.2",
     "ora": "5.4.1",
     "pkg-up": "3.1.0",
-    "tar": "7.5.18",
+    "tar": "7.5.16",
```
unrevised6419: Chore: To revert

packages/cli/create-strapi-app/package.json
```diff
@@ -64,12 +64,13 @@
     "rollup": "4.60.4",
     "semver": "7.7.4",
     "sort-package-json": "2.10.0",
-    "tar": "7.5.18"
+    "tar": "7.5.16"
```
unrevised6419: Chore: To revert

packages/core/content-manager/package.json
```diff
@@ -96,7 +96,7 @@
     "markdown-it-sub": "^1.0.0",
     "markdown-it-sup": "1.0.0",
     "prismjs": "1.30.0",
-    "qs": "6.15.3",
+    "qs": "6.15.2",
```
unrevised6419: Chore: To revert

packages/core/core/package.json
```diff
@@ -107,7 +108,7 @@
     "ora": "5.4.1",
     "package-json": "10.0.1",
     "pkg-up": "3.1.0",
-    "qs": "6.15.3",
+    "qs": "6.15.2",
```
unrevised6419: Chore: To revert

packages/core/permissions/package.json
```diff
@@ -48,11 +48,14 @@
     "@casl/ability": "6.7.5",
     "@strapi/utils": "5.50.0",
     "lodash": "4.18.1",
-    "qs": "6.15.3",
+    "qs": "6.15.2",
```
unrevised6419: Chore: To revert

unrevised6419:
I'm almost finishing the review of the rebase you did, will revert a few changes. (rebase is ok, won't push my merge)

packages/plugins/i18n/package.json
```diff
@@ -65,7 +65,7 @@
     "@strapi/icons": "2.2.1",
     "@strapi/utils": "5.50.0",
     "lodash": "4.18.1",
-    "qs": "6.15.3",
+    "qs": "6.15.2",
```
unrevised6419: Chore: To revert

packages/providers/email-nodemailer/package.json
```diff
@@ -68,15 +68,19 @@
     "build:types": "run -T tsc -p tsconfig.build.json --emitDeclarationOnly",
     "clean": "run -T rimraf ./dist",
     "lint": "run -T eslint .",
+    "test:ts": "run -T tsc -p tsconfig.json",
     "test:unit": "run -T jest",
     "watch": "run -T rollup -c -w"
   },
   "dependencies": {
-    "nodemailer": "9.0.1"
+    "nodemailer": "8.0.9"
```
unrevised6419: Chore: To revert

packages/providers/email-sendmail/package.json
```diff
@@ -41,15 +41,19 @@
     "build:types": "run -T tsc -p tsconfig.build.json --emitDeclarationOnly",
     "clean": "run -T rimraf ./dist",
     "lint": "run -T eslint .",
+    "test:ts": "run -T tsc -p tsconfig.json",
     "test:unit": "run -T jest --watchman=false",
     "watch": "run -T rollup -c -w"
   },
   "dependencies": {
-    "nodemailer": "9.0.1"
+    "nodemailer": "8.0.9"
```
unrevised6419: Chore: To revert

unrevised6419 (commented):
looks good to me

packages/core/core/src/ee/license.ts
```diff
@@ -114,13 +120,28 @@ const fetchLicense = async (
   const contentType = response.headers.get('Content-Type');
 
   if (contentType?.includes('application/json')) {
-    const { data, error } = await response.json();
+    const json = await response.json();
+    const result = licensePayloadSchema.safeParse(json);
+
+    if (!result.success) {
+      throw new LicenseCheckError('Invalid license payload.');
```
nclsndr: The new `licensePayloadSchema.safeParse` runs before the status switch and throws `LicenseCheckError('Invalid license payload.')` with `shouldFallback` defaulting to `false` (see the constructor default in this file). `onlineUpdate`'s `fallback` handler (`packages/core/core/src/ee/index.ts:129`) only reuses the stored license when `shouldFallback` is `true`.

Before this PR, any non-200/400/404 JSON response fell through to `throwError()`, which always sets `shouldFallback = true`. Now, if the registry returns a JSON payload shape `safeParse` doesn't recognize on one of those other statuses (e.g. a 5xx with an unexpected body), EE gets disabled instead of falling back to the last stored license.

Can we preserve the old fallback semantics here, e.g. only reject without fallback for statuses we explicitly treat as authoritative (200/400/404), and route anything else (including schema-parse failures) through `throwError()`?

---
_AI assisted_

packages/core/content-manager/server/src/tsconfig.build.json
```diff
@@ -4,6 +4,7 @@
   "exclude": ["./src/**/*.test.ts"],
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./dist/server"
+    "outDir": "./dist/server",
+    "noEmit": false
```
nclsndr: Adding `noEmit: false` surfaces that this config is already broken: it `extends: "./tsconfig.json"` and `include: ["./src"]`, but there's no `packages/core/content-manager/server/src/tsconfig.json` and `server/src/src` has no inputs.

```
yarn exec tsc --showConfig -p packages/core/content-manager/server/src/tsconfig.build.json
error TS5083: Cannot read file '.../server/src/tsconfig.json'.
error TS18003: No inputs were found in config file '.../server/src/tsconfig.build.json'.
```

Can we either delete this file if nothing references it as a build target, or fix `extends`/`include` to point at the intended server paths (likely `../tsconfig.json` / `.`)?

---
_AI assisted_

jsconfig.json
```diff
@@ -8,14 +8,19 @@
...
     "**/build/*",
     "**/*.test.js",
     "**/*.test.api.js",
-    "**/examples/getstarted/*",
-    "**/examples/kitchensink/*",
-    "docs"
+    "docs",
+    "examples",
+    "templates",
+    ".github",
```
nclsndr, line 21: Excluding `.github` here removes it from the root JS type-scope, but `.github/actions/check-pr-status/package.json` adds Jest devDeps for a JS action, and `.github/actions/check-pr-status/__tests__/action.test.js` uses Jest globals — there's no local `.github/**/jsconfig.json` to cover it.

Not a runtime bug, but it breaks the "every workspace owns its local JS/TS scope" model this PR is establishing. Worth adding a small `.github/actions/check-pr-status/jsconfig.json`, or keeping `.github` covered centrally?

---
_AI assisted_

nclsndr (changes requested):
Did a deeper pass on top of the earlier review thread — the `zod` runtime validation on `core/ee/license.ts` (784ba1c) is a solid fix for the blind `response.json()` cast, but it introduced a fallback-semantics regression. Left inline comments with the specifics on the three confirmed issues.

Summary:
- **`core/ee/license.ts`** — schema-parse failures now throw without `shouldFallback: true`, so an unexpected registry payload shape can disable EE instead of falling back to the last stored license (previously any non-200/400/404 fell through to `throwError()`, which always allowed fallback). High severity, EE boundary — flagged inline.
- **`content-manager/server/src/tsconfig.build.json`** — `noEmit: false` surfaces that this config is already broken (`extends`/`include` point at paths that don't exist); `tsc --showConfig` errors out. Flagged inline.
- **`jsconfig.json`** — `.github` dropped from root scope with no local `jsconfig.json` to replace it, despite `check-pr-status` having Jest deps/globals. Low severity, but breaks the "every workspace owns its scope" model this PR establishes. Flagged inline.

Everything else checked out — `yarn test:ts` and targeted `build:types:server` typechecks pass across the reviewed packages (`admin-test-utils`, `content-manager`, `upload`, `plugin-sentry`, `data-transfer`, `i18n`), no dead imports from the dropped `baseUrl`, no orphaned test files from the dedup.

Once the license fallback and the content-manager tsconfig are fixed, happy to re-run `yarn test:ts` and approve.

---
_AI assisted_

packages/core/core/src/ee/license.ts
unrevised6419: @nclsndr would this be enough?

```ts
const fetchLicense = async (/* ... */) => {
  // ...

  if (contentType?.includes('application/json')) {
    const json = await response.json();
    const result = licensePayloadSchema.safeParse(json);

    if (!result.success) {
      throw new LicenseCheckError('Invalid license payload.', true); // <- added true
    }

    const payload = result.data;

    switch (response.status) {
      case 200: {
        if ('data' in payload) {
          return payload.data.license;
        }
        throw new LicenseCheckError('Invalid license payload.', true); // <- added true
      }
      case 400: {
        if ('error' in payload) {
          throw new LicenseCheckError(payload.error.message);
        }
        throw new LicenseCheckError('Invalid license payload.', true); // <- added true
      }
      case 404:
        throw new LicenseCheckError('The license used does not exists.');
      default:
        throwError();
    }
  } else {
    throwError();
  }
};
```

unrevised6419: nvm, looked once more over the code, it seems is not enough

nclsndr: Yes, the latest committed direction looks good to me: valid `400` / `404` stay authoritative, while malformed or unexpected registry responses fallback to the stored license.

One small improvement I’d consider: switch on `response.status` before schema parsing, and only parse the JSON body for statuses that need it (`200` and `400`). For default/unexpected statuses we don’t need to validate the body at all; they can go straight to the fallback error. That makes the fallback semantics harder to regress.

Could we also add focused tests for:
- malformed `200` payload => fallback
- malformed `400` payload => fallback
- valid `400 { error.message }` => no fallback
- `500` / unexpected status => fallback
- `404` => no fallback

---
_AI assisted_

unrevised6419: Moved all these license changes into https://github.com/strapi/strapi/pull/26935

unrevised6419:
Hey @nclsndr, let's try one more CI run :) hopefully latest 😄

nclsndr:
Publishing experimental version: https://github.com/strapi/strapi/actions/runs/28942446759

`0.0.0-experimental.12e4785bc3e5c6e747e4e0a9a57d04afbdf2cb2c`

packages/cli/cloud/src/utils/tests/compress-files.test.ts
nclsndr, line 1: Why are we deleting that file?

packages/cli/create-strapi-app/src/utils/template.ts
```diff
@@ -158,7 +158,6 @@ async function downloadGithubRepo(rootPath: string, { owner, repo, branch, subPa
   }
 
   await pipeline(
-    // @ts-expect-error - Readable is not a valid source
```
nclsndr, line 161: We need to test the experimental to validate that deletion

packages/cli/create-strapi-app/package.json
```diff
@@ -47,7 +47,7 @@
     "build:types": "run -T tsc -p tsconfig.build.json --emitDeclarationOnly",
     "clean": "run -T rimraf ./dist",
     "lint": "run -T eslint . --max-warnings=0",
-    "test:ts": "run -T tsc --noEmit",
+    "test:ts": "run -T tsc -p tsconfig.json",
```
nclsndr, line 50: I guess this is equivalent. No strong opinion on keeping it or removing
```suggestion
    "test:ts": "run -T tsc",
```

unrevised6419:
Tried to create a new app, everything works fine develop/build.
One more thing that probably will do in a subsequent PR is to also set `"types": []`, and enable `"strict": true` to all templates

packages/cli/cloud/src/utils/tests/compress-files.test.ts
unrevised6419: duplicated/orphan, there is an identical file in some adjacent folder

packages/cli/create-strapi-app/package.json
unrevised6419: When having multiple tsconfigs in workspaces I would opt for explicit `-p`

packages/cli/cloud/src/utils/tests/compress-files.test.ts
unrevised6419:
```
packages/cli/cloud/src/utils/__tests__/compress-files.test.ts
```

packages/cli/create-strapi-app/src/utils/template.ts
unrevised6419:
```
npx create-strapi-app@0.0.0-experimental.12e4785bc3e5c6e747e4e0a9a57d04afbdf2cb2c my-app-name \
  --template=strapi/strapi \
  --template-branch=develop \
  --template-path=templates/website

 Strapi   v0.0.0-experimental.12e4785bc3e5c6e747e4e0a9a57d04afbdf2cb2c 🚀 Let's create your new project

 
To deploy your project, create a new project on the Strapi Cloud dashboard.

? Please log in or sign up. Skip
? Do you want to use the default database (sqlite) ? Yes
? Install dependencies with npm? No
? Initialize a git repository? Yes

 Strapi   Creating a new application at /Users/me/Projects/github/unrevised6419/my-app-name
       ●  Installing template strapi/strapi

       ✓  Template copied successfully.

    git   Initializing git repository.

       ✓  Initialized a git repository.

 Strapi   Your application was created!
          Available commands in your project:
          
          Start Strapi in watch mode. (Changes in Strapi project files will trigger a server restart)
          npm run develop
          
          Start Strapi without watch mode.
          npm run start
          
          Build Strapi admin panel.
          npm run build
          
          Deploy Strapi project.
          npm run deploy
          
          Display all available commands.
          npm run strapi

          To get started run
          
          cd /Users/me/Projects/github/unrevised6419/my-app-name
          npm install
          npm run develop
```

unrevised6419:
Rebased to align with "5.50.1"

packages/cli/create-strapi-app/src/utils/template.ts
nclsndr: needs to properly build once installed 😉

unrevised6419:
```
nr build

> my-app-name@0.1.0 build
> strapi build

✔ Compiling TS (1201ms)
⠋ Building build context
✔ Building build context (114ms)
⠋ Building admin panel
The CJS build of Vite's Node API is deprecated. See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
✔ Building admin panel (10324ms)
```

nclsndr: nvm, this bit is actually out of user-land. All good.

packages/cli/create-strapi-app/package.json
nclsndr: Good for me

nclsndr (dismissed):
Such a huge improvement for the code base hygiene! 
Thank you so much for your commitment into improving Strapi DX @unrevised6419 🙌 

LGTM

nclsndr (dismissed):

nclsndr (dismissed):

nclsndr (dismissed):

nclsndr:
Build was failing across all node versions (`Could not find a declaration file for module '@strapi/typescript-utils'`). Root cause: with `noEmit: true` now centralized in `tsconfig/base.json`, every `tsconfig.build.json` needs to explicitly set `noEmit: false` to actually emit — `packages/utils/typescript/tsconfig.build.json` was the one file missing it (verified it's the only one across the repo), so `build:types` silently emitted no `.d.ts` and broke typecheck for every consumer (`@strapi/core`, `@strapi/admin`, `@strapi/generators`, `@strapi/strapi`).

Pushed d70c553 adding `noEmit: false` there, matching every other package's `tsconfig.build.json`. Verified locally: clean `yarn build` (45 projects, 0 errors), `dist/index.d.ts` now generated, and `@strapi/core:build:types` passes.

---
_AI assisted_

nclsndr (approved):

unrevised6419:
🎉 🎉 🎉

PR #26700, opened 2026-06-18
Fix: 25780 release relation sync

### LINKED ISSUE #25780: Releases being affected by out of sync relations
### Node Version

20.17

### Package Manager

yarn

### Package Manager Version

Classic, lockfile v1

### Strapi Version

5.30.1

### Operating System

Strapi Cloud

### Database

Strapi Cloud

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

When publishing a Release, Strapi may fail with a relation validation error if entities or their relations were modified after those entities were added to the Release.

In the reported case, the Release appeared valid when the entities were first added. However, after later modifying parent and child entities and updating some relations, publishing the Release failed with the following error:

`ValidationError: 1 relation(s) of type api::chapter.chapter associated with this entity do not exist
`
Based on the observed behavior, it appears that the Release stores the entity state from the moment the entity is added and does not automatically refresh later changes, especially relation updates. As a result, when publishing, Strapi re-validates relations against outdated Release data and fails.

A workaround was found by removing the modified entities from the Release and adding them again. After re-adding them, the Release published successfully.

### Steps to Reproduce

1. Create a Release.
2. Add a parent entity and its related child entities to the Release while their relations are valid.
3. Update the parent entity and move it through the workflow until it becomes modified.
4. Add or re-add that parent entity to the Release.
5. Update the related child entities and move them through the workflow until they become modified.
6. In some cases, update relations between the parent and child entities during this process.
7. Add those modified child entities to the Release.
8. Try to publish the Release.
9. Observe that publishing fails with a relation validation

### Expected Behavior

One of the following behaviors would be expected:
- The Release automatically reflects the latest entity state, including updated relations, before publishing, or
- The UI clearly indicates that entities added earlier may now be outdated and need to be refreshed or re-added, 
- There is a built-in way to refresh or sync entity state inside a Release before publishing.

### Logs

```shell

```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

Bassel17:
### What does it do?
Prevents Document Service publish/discard transforms from reusing stale request-scoped documentId-to-entityId mappings after rows are deleted and recreated.

### Why is it needed?
Content Releases can publish multiple related documents in one request. A stale cached numeric entity ID can point to a deleted published row and cause relation validation to fail.

### How to test it?
Run the focused unit tests for the Document Service transform cache and the EE content-releases API regression.

### Related issue(s)/PR(s)
Fixes #25780

### DIFF
--- a/jest.config.api.js
+++ b/jest.config.api.js
@@ -35,5 +35,5 @@ module.exports = {
   transform: {
     '^.+\\.ts$': ['@swc/jest'],
   },
-  modulePathIgnorePatterns: ['.cache'],
+  modulePathIgnorePatterns: ['[/\\\\]\\.cache[/\\\\]'],
 };
--- a/jest.config.cli.js
+++ b/jest.config.cli.js
@@ -14,5 +14,5 @@ module.exports = {
   transform: {
     '^.+\\.ts$': ['@swc/jest'],
   },
-  modulePathIgnorePatterns: ['.cache'],
+  modulePathIgnorePatterns: ['[/\\\\]\\.cache[/\\\\]'],
 };
--- a/packages/core/core/src/services/document-service/entries.ts
+++ b/packages/core/core/src/services/document-service/entries.ts
@@ -8,7 +8,7 @@ import { transformParamsDocumentId } from './transform/id-transform';
 import { transformParamsToQuery } from './transform/query';
 import { pickSelectionParams } from './params';
 import { applyTransforms } from './attributes';
-import { transformData } from './transform/data';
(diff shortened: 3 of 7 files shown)


innerdvations (changes requested):
Overall the code seems good, but the test file isn't actually run by jest, it gets skipped because the regexp in `modulePathIgnorePatterns: ['.cache']` matches `-cache` in the filename.

We can either take the opportunity to fix that regexp, or just the file to make sure it runs

innerdvations (approved):
LGTM!

PR #26701, opened 2026-06-18
chore(jest): run unit/front tests via nx, drop root jest config and dep

unrevised6419:
### What does it do?

Runs Jest per-workspace through nx instead of the root `projects[]` aggregators — which is already how CI runs tests (`nx affected --target=test:unit|test:front`).

- Deletes the root `jest.config.js` and `jest.config.front.js` aggregators; the local `test:unit*`/`test:front*` scripts now delegate to `nx run-many` so they match CI.
- Removes `jest` from the repo root (keeps `jest-cli`, which still provides the `jest` binary).
- Adds `test:unit`/`test:unit:watch` targets to the `check-pr-status` GitHub action, which had jest tests but no nx target and so was skipped by nx/CI (the old aggregator picked it up via `.github/actions/*/jest.config.js`).
- Removes the obsolete `jest.config.js` entry from `.github/filters.yaml`.

### Why is it needed?

The root aggregator configs are dead weight — CI already uses nx targets, so they only served the local convenience scripts. More importantly, a **root dependency is visible to every workspace**: a root `jest` let any package import Jest without declaring it. With `jest` removed from the root and declared per-workspace, eslint's `import/no-extraneous-dependencies` can now flag any workspace that uses Jest without declaring it. nx also gives caching/affected for the local test scripts, matching CI.

### How to test it?

- `yarn test:unit` / `yarn test:front` (now `nx run-many`) — all projects pass
- `yarn lint` / `yarn prettier:check`
- `yarn run -T jest --version` still resolves (via `jest-cli`)

### Related issue(s)/PR(s)

Part of the groundwork discussed in #26543. Previously stacked on #26699 (now merged).


jest.config.js
unrevised6419, line 1: **Why removing this is safe — no config is lost.**

This root config only did two things: declare `projects[]`, and set coverage options (`collectCoverage`, `collectCoverageFrom`, `coverageDirectory`, `coverageReporters`, `coveragePathIgnorePatterns`).

All five coverage keys already live in `jest-preset.unit.js`, which **every** workspace references via `preset:` in its own `jest.config.js`. So when tests run per-project (via `nx run-many` / `nx affected`, which is already how CI runs them), each project gets the equivalent coverage config from the preset — just project-relative instead of repo-relative:

| key | here (aggregate) | preset (per-project) |
|---|---|---|
| `collectCoverageFrom` | `<rootDir>/packages/**/*.{js,ts}` | `**/*.{js,ts,jsx,tsx}` |
| `coverageDirectory` | `<rootDir>/coverage` (repo root) | `<rootDir>/coverage` (each package) |

The per-project `coverageDirectory` → `packages/*/coverage/` is exactly what CI uploads and SonarQube ingests, so the preset values match the CI model better than this aggregate config did. `jest.config.front.js` contained only `projects[]` (no extra config), so nothing there either.

unrevised6419:
@nclsndr please trigger again CI, I rebased on parent branch, as there were some TS errors.

unrevised6419:
Until the target PR get's merged I'll keep this as draft

innerdvations (approved):
Good idea, thanks!

PR #26711, opened 2026-06-19
feat: adding publish preview packages

cache-your-dreams:
### What does it do?

Adding Publish Preview Packages so that an experimental is created for each PR

### Why is it needed?

To ease QA'ing of PR's by automatically creating an installable version

### How to test it?

I'm hoping to find that out by merging.


nclsndr:
Since we have the `publish-experimental` label that produces on-demand experimental releases I would avoid making that systematic, saving many entries in https://www.npmjs.com/package/@strapi/strapi?activeTab=versions as well

PR #26732, opened 2026-06-22
fix(content-manager): serve live preview script from server endpoint

### LINKED ISSUE #135: Live-preview click-to-edit broken in production (injected highlight script references undefined helper)
## Description

Live-preview **click-to-edit** silently breaks in **production** deployments. It works under `strapi develop`. The preview iframe renders the page, but hovering/clicking text or media never highlights or selects a field.

## Steps to reproduce

1. Deploy LaunchPad (Next frontend + Strapi) to production (e.g. Strapi Cloud + Vercel).
2. Open a content entry's **Preview** in the admin.
3. Try to click text/media in the preview iframe to edit it.

**Expected:** the field highlights and selecting it focuses the corresponding field in the editor.
**Actual:** nothing highlights; no overlay; clicks do nothing.

## Root cause

Strapi's **production** admin build injects a live-preview highlight script into the iframe that calls a minified bundler helper (a lazy-import wrapper around `import('@vercel/stega')`) which it **never includes in the emitted script**:

```js
const { vercelStegaDecode, vercelStegaClean } = await ye(async () => {
  const { ... } = await import("https://cdn.jsdelivr.net/npm/@vercel/stega@.../+esm");
  return { ... };
}, []);
```

`ye` is undefined → `ReferenceError: ye is not defined`. The init promise rejects with no `.catch`, so the highlight/click system never initializes. Dev works because that build inlines the helper. The minified name is unstable across builds (observed `ye` → `be`).

## Environment

- Strapi 5.47.1
- Production only (works in dev)

## Fix

- Temporary client-side workaround in the Next frontend: PR #134 (name-agnostic shim that defines whatever helper the injected script references).
- ⚠️ **Real fix belongs in Strapi core** — the admin build should emit a self-contained preview script. A separate issue/PR should be filed against `strapi/strapi`, and the workaround in #134 removed once that lands.

Related PR: #134

Baldinof:
### What does it do?

It moves the preview highlight script out of the admin bundle because the bundler introduces wrapper code that breaks the script when injected into the preview iframe.

Now the script is directly written in js (still typescript validated), served from a server endpoint and injected into the preview iframe at runtime.

Status: **DRAFT** while I'm testing Launchpad with an experimental of this branch

### Why is it needed?

Fix the live preview.

### How to test it?

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/LaunchPad/issues/135

### DIFF
--- a/docs/docs/docs/01-core/content-manager/06-preview.md
+++ b/docs/docs/docs/01-core/content-manager/06-preview.md
@@ -19,13 +19,18 @@ Instead, the preview script is defined inside Strapi and sent to the frontend vi
 
 ### Self-contained constraint
 
-The preview script (`packages/core/content-manager/admin/src/preview/utils/previewScript.ts`) is stringified before being sent to the iframe:
+The preview script (`packages/core/content-manager/server/src/preview/controllers/previewScript.js`) is served verbatim from a server endpoint (`GET /content-manager/preview/script`). The admin fetches it, wraps it with its runtime config, and posts the result to the iframe:
 
 ```ts
-const script = `(${previewScript.toString()})(${JSON.stringify(config)})`;
+const previewScriptSource = await fetch('/content-manager/preview/script').then((res) =>
+  res.text()
+);
+const script = `(${previewScriptSource})(${JSON.stringify(config)})`;
 ```
 
 Because of this, it **cannot import dependencies or reference external variables**. All logic must be self-contained. The only external code (`@vercel/stega` for decoding) is loaded dynamically from a CDN at runtime.
+
+It lives in a standalone `.js` file served as-is rather than being bundled with the admin: a bundler wraps the module in helper code that breaks once the function is stringified and injected into the iframe. The file is still type-checked via `@ts-check` and JSDoc.
 
 This is why the file has an unusual structure with many functions defined inline.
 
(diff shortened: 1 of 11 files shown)


Baldinof:
@PaulBratslavsky You can test the fix on Launchpad with `0.0.0-experimental.5281af240fc8ba10a5a2e34b9593ac192d11fa4f`

PaulBratslavsky:
@Baldinof thank you will test today.

PaulBratslavsky:
@Baldinof hey, do you when this is going to be merged?  Will updated Launchpad project after this.

innerdvations (commented):
Overall LGTM! Just need to update the CLI snapshots `yarn test:cli:update`

You might also want to wrap the async fetch in try/catch and update docs `content-manager/06-preview.md` that still describes the old path

Baldinof:
Thanks for the review @innerdvations :) I pushed the fixes

innerdvations (dismissed):
Docs nits fixed on-branch. Approach looks good — serving the preview script verbatim avoids bundler wrapper breakage.

innerdvations:
I'll fix the coderabbit suggestions and then merge

Baldinof:
Thank you so much!

innerdvations (approved):

Issue #25568, opened 2026-02-26
Performance degradation on publish with large dynamic zone schemas (v5)


eduper11hanzo:
### Node Version

20.19.5

### NPM/Yarn/PNPM Version

10.8.2

### Strapi Version

5.36.1

### Operating System

Linux (Other)

### Database

MySQL

### Javascript or Typescript

Javascript

### Bug Description

Publishing any content type that has a large dynamic zone schema takes 4–6 seconds, regardless of how many components the specific entry actually uses. The slowness scales with the number of component types declared in the dynamic zone, not with the actual content of the entry being published.
A pages collection with ~30 module types declared in its dynamic zone takes 4.3–4.6 seconds to publish an entry that only uses 5 modules. A faq collection with 2 module types in its dynamic zone publishes in 100–200ms.

### Steps to Reproduce

1. Create a content type with a dynamic zone declaring 30+ component types
2. Create an entry using only 5 of those component types
3. Publish the entry
4. Observe that publish takes 4+ seconds

### Expected Behavior

Publish time should scale with the actual content of the entry being published, not with the total number of component types declared in the dynamic zone schema.

### Logs

```shell
## Investigation

We performed an exhaustive investigation to identify the root cause. Here is the evidence:

### 1. Database indexes are correct and being used

All relevant indexes exist and MySQL uses them:


pages_documents_idx: (document_id, locale, published_at) → type: ref, rows: 1, key used
upload_folders_documents_idx: (document_id, locale, published_at) → type: ref, rows: 1, key used


EXPLAIN confirms MySQL resolves the queries in <1ms. Database is not the bottleneck.

### 2. No lock contention


SHOW STATUS LIKE 'Innodb_row_lock_waits';   -- 0
SHOW STATUS LIKE 'Innodb_row_lock_time_avg'; -- 0


### 3. No Node.js event loop blocking

We instrumented the event loop with `monitorEventLoopDelay` from `perf_hooks`. Zero lag events were recorded during publish. JavaScript serialization of 850 objects takes ~2ms on this machine.

### 4. The entire 4.5 seconds is inside the transaction

Summing all Knex query timings within the publish transaction (trx39):


Queries: 894
Total time inside transaction: 4469ms
MySQL execution time (sum of all individual queries): ~150ms
Unexplained gap time (between queries): ~4300ms


**894 queries are executed for a single publish of a page with 5 modules.**

### 5. The gaps occur at specific points in the transaction

[...]

PR #26734, opened 2026-06-22
fix(database): stop full-schema component_type IN on dynamic zone populate

jasleenkaur-qed42:
## Summary

- Fixes performance issue when publishing entries with large dynamic zone schemas ([#25568](https://github.com/strapi/strapi/issues/25568)).
- `getDeepPopulate()` builds `populate.on` with every declared DZ component type; `morphToMany` was turning those keys into `component_type IN (...)` over the **full schema**.
- The join table (`*_cmps`) already stores only the component types present on the entry. This change queries by `entity_id` + `field` and applies `populate.on` filtering in memory.

## Root cause

`packages/core/database/src/query/helpers/populate/apply.ts` → `morphToMany` added:

```sql
component_type IN (all declared DZ types)


innerdvations (approved):
Thanks for the PR! LGTM, and I have added a few tests.

PR #26736, opened 2026-06-22
fix(content-manager): warn before publishing with draft relations

### LINKED ISSUE #25361: Relations silently lost when publishing document before its relation target

innerdvations:
## Summary

- Re-enables draft-relation counting when publishing from the edit view and shows a confirmation dialog when the entry references unpublished targets.
- Editors can **cancel** or **publish without relations** (relations are omitted from the published version, as today).
- Re-enables the same count for **bulk publish** in the list view so the warning appears there too.
- Updates copy to clarify that draft relations will not be included in the published version.

m2m (bidirectional):
image

unidirectional:
image

Fixes #25361

Publishing related entries first is intentionally out of scope here (permissions and ordering make it a follow-up).

## Test plan

Experimental version: `0.0.0-experimental.cb80fd5e9b78bde9ac2c175e5a5e1977b2bcfb33`

- [x] `yarn test:front DocumentActions.test ConfirmBulkActionDialog.test`
- [x] `yarn nx run @strapi/content-manager:lint`
- [x] `yarn test:e2e --domains content-manager --grep "warns on publish"` (new E2E spec)
- [ ] Manual: create entry with draft relation → Publish → see warning → Cancel leaves draft unchanged
- [ ] Manual: confirm with **Publish without relations** → published entry has no relation; draft still has it


Bassel17:
I tested this a bit, I think there are a few flaws here and there, so when we have many to many, i guess 1 to 1 and self relations as well, the pop up will appear but there is not way to resolve it because it will appear from both contenttypes both  being unable to published, and the pop up seems to keep appearing regardless of if I published the related contentType or not.
another minor issue, it seems the `Cmd/Ctrl+Shift+Enter` shortcut does not trigger publish now when a related contentType is in draft and so it doesn't even show the pop up either

Bassel17:
actually in many to many we may not need the pop up warning at all, it seems the related entity self heals if it gets published

Bassel17:
The modal seems to be popping up whenever a draft relation is added, am guessing adding the modal is affecting the flow of the e2e tests as well

Bassel17 (approved):
LGTM!

PR #26749, opened 2026-06-23
feat(content-type-builder): Preserve data on renaming attributes

innerdvations:
## Summary

- Add data-preserving rename migrations for CTB attribute renames: admin captures ordered `renames[]` per content-type/component, server generates a guarded migration file before reload, and schema-sync no longer drops renamed columns.
- Add nested project configuration for attribute behavior (`always`, `never`, `prompt-after-edit`, or `prompt-before-save`) while leaving room for a future content-type rename setting. The default remains a before-save confirmation.
- Add migration-file format configuration (`javascript` or `typescript`, default `javascript`) and execute discovered TypeScript migrations through a scoped transpiler.
- Extract a generic migration file builder to `@strapi/database` (`renameColumn` / `renameTable` / `updateRows`); keep CTB-specific rename resolution in `content-type-builder` (scalars, relations, components, dynamic zones, media; polymorphic `morph*` reported unsupported).
- Preserve ordered multi-hop paths exactly, including name reuse such as `A -> C`, `B -> A`, `A -> B`.
- Support component-level renames (category change → uid change) via `component_type` updates across `*_cmps` link tables (CG-1001).
- Add optional CLI commands, `strapi rename:field` and `strapi rename:component`, reusing the admin's migration path.
- Add AI chat v1 rename compatibility for inferred and explicit rename metadata.
- Invalidate cached Content Manager documents after schema changes so renamed values appear immediately after restart.

**Out of scope here:** operations-based AI v2 client (#26878) and content-type name migrations.

Continues #26622 after renaming the head branch to `feat/ctb-rename-migration-builder`.

## Test plan

- [x] Database migration unit tests, including generated TypeScript discovery and execution
- [x] CTB server rename/config/schema unit tests
- [x] Complete CTB frontend test suite
- [x] Type checks for `@strapi/content-type-builder`, `@strapi/database`, and `@strapi/strapi`
- [x] Monorepo lint and formatting checks
- [x] Changed package builds
- [x] SQLite API integration tests, including localized draft-and-publish content
- [x] CLI integration tests for `rename:field` and `rename:component`
- [x] Chromium E2E: renaming an Article field preserves and displays existing localized draft content

## Related issue(s)/PR(s)

Fixes CMS-635
Supersedes #26622
Follow-up: #26878 (AI v2 operations client)


jhoward1994:
An explanation of the changes since [2644882](https://github.com/strapi/strapi/pull/26749/commits/2644882e4d7838aa97faacf49569ef648de73fa2) :

### 1. Generated migrations now always land in the app source dir, as JS by default and TS when `useTypescriptMigrations` is on

Previously the file went to the database's configured migrations dir, which points at gitignored build output when `useTypescriptMigrations` is enabled, so the migration was never committed and never discovered. Now:

- The file is always written to `<app>/database/migrations` (source), never the compiled dir.
- Default output is JavaScript (`module.exports = { async up(knex, db) {...} }`).
- When `database.settings.useTypescriptMigrations` is `true`, a `.ts` file is emitted instead so `tsc` compiles it into `dist` and the resolver picks it up. Without this the rename fell through to drop-and-recreate.
- Removed the `renameMigrations.migrationFile.format` config option, the `MIGRATION_FILE_FORMATS` type, and the per-file TypeScript require hook in the migration resolver. The compiled `dist` path is the only supported route for TS migrations.

### 2. Guards, logging and dialect specifics moved into runtime helpers on `db.schema`

The generated file no longer inlines raw knex calls wrapped in `hasTable` / `hasColumn` checks. It now calls three guarded helpers exposed on the database (`packages/core/database/src/schema/rename-helpers.ts`):

```js
await db.schema.renameColumn(knex, { table: 'articles', from: 'bio', to: 'biography' });
await db.schema.renameTable(knex, { from: 'old_cmps', to: 'new_cmps' });
await db.schema.updateRows(knex, {
  table: 'articles_cmps',
  where: { field: 'hero' },
  set: { field: 'banner' },
});
```

Why: generated files stay small and readable, and any future fix to guard or dialect logic applies to migrations users have already generated.

### 3. Physical-name rules shared between the metadata loader and the CTB

`advance()` in the CTB migration builder was re-implementing the naming rules from `metadata/relations.ts` (join table names, join column ids, etc.) to compute the destination identifier. If those rules drifted, the guard would turn the rename into a silent no-op.

There is now a single `db.metadata.naming` surface (`packages/core/database/src/metadata/attribute-naming.ts`) used by both the metadata loader and the CTB, so the new side of a rename is derived exactly as the reload will derive it.

### 4. Rename + storage change is refused, not migrated

A hop that also changes the field's storage (type, relation kind or target, component uid or `repeatable`) used to be recorded as a rename. The migration would rename the column and then schema sync would try to `alter()` the type, which Postgres/MySQL reject for incompatible casts, so the app failed to start. Before this branch the column was dropped and recreated.

Now guarded at both ends:

- Admin reducer (`recordRename`) does not record the hop when the type/relation/component shape changed.
- Server migration builder refuses it and reports it as an `UnsupportedRename` with reason `type-changed`, which the existing warning in `generateRenameMigrations` surfaces.

Behaviour for those cases is therefore the same as before this PR (drop and recreate), just without a broken migration file.

### 5. Component display-name renames

- When a component's display name change also changes its `collectionName`, the component's data table is renamed via `db.schema.renameTable` rather than being dropped and recreated.
- The admin follows the renamed component uid after reload instead of leaving the user on a dead route.

### 6. Rename confirmation modal

Users can now untick individual fields in the modal. Unticked fields fall back to the previous drop-and-recreate behaviour for that field only.

### 7. Timestamp prefix

Generated files are now named `YYYY.MM.DDTHH.mm.ss.SSS.rename-fields.<js|ts>` using a **UTC** timestamp. Same shape as `strapi generate migration`, with milliseconds appended, so:

### 8. Docs

Added a "User migrations" section to `docs/docs/docs/01-core/database/03-migrations.md` covering discovery and ordering, the timestamp prefix, and the guarded helpers.

innerdvations:
on point 4, data type change, I think that if renaming is enabled, we should trigger a warning modal as soon as the edit happens saying that it will cause data loss due to the type changing.

As a follow-up (not in this PR) we could probably safely migrate a few of the data types, and migrate many more of them somewhat destructively but would still be useful for the vast majority of users.

innerdvations (commented):
## Change request

Re-reviewed at `3a8f3ea443` (base `develop`). Default JS rename path looks solid; prior e2e/`draftAndPublish` gaps are closed. Three production issues still block merge:

### 1. `useTypescriptMigrations` — generated file never runs on CTB save/reload

When `database.settings.useTypescriptMigrations` is true, CTB writes a `.ts` file to **source** `database/migrations`, then `strapi.reload()` runs immediately. Discovery only loads compiled `*.js`/`*.sql` under `dist/database/migrations`. Schema sync then treats the rename as drop+add → **data loss**.

The PR body mentions a scoped transpiler; it is not in the tree. Hand-written TSM files can be compiled before a manual start; generated files cannot — save always reloads first.

**Fix in this PR:** put a runnable file on the discovery path before reload (emit JS into the discovery dir, and/or transpile/register `.ts`), not “compile yourself after CTB already restarted.”

### 2. Preserve modal — independent hop checkboxes break swap chains

The before-save modal lets users untick hops independently. Server replays the remaining ordered `renames[]`. Dropping a prefix hop in `A→C, B→A, A→B` yields `attribute-not-found` → warn + no migration for the rest, while the schema still saves.

**Fix in this PR:** chain consent all-or-nothing (or prefix-safe only). Never submit a truncated chain the builder cannot replay.

### 3. AI `applyChange` can wipe pending `renames[]`

Manual edits accumulate `renames` in the reducer; `applyChange` replaces the whole type without merging. An AI update without rename metadata drops those hops; save then drop/recreates. `prompt-after-edit` is also skipped on the AI apply path.

**Fix in this PR:** merge existing `renames` on `applyChange` update, and reuse the same preserve-consent path as manual edits.

---

Description drift (TS format config / transpiler) is fine as non-blocking polish once the above land.

_AI-assisted comment._

jhoward1994:
## Re-review Summary (1c7c97e)

1. **useTypescriptMigrations**  
On reload, `strapi develop` already recompiles the app before forking the new worker (see `packages/core/strapi/src/node/develop.ts`, case `'reload'`). Thus, any `.ts` migration in `database/migrations` is emitted to `<outDir>/database/migrations` and discovered immediately on reload. The same applies during `strapi build` for production. The core issue fixed was a mismatch between migration file format selection and discovery: `getFormat()` now mirrors the database's choice, and falls back to `.js` (with a warning) if TSM is enabled but no `outDir` resolves. The test `file-builder-compile.test.ts` ensures that the generated template compiles through the vanilla `tsconfig`, so `noEmitOnError` cannot break a user build. The docs' "scoped transpiler" line was outdated and has been removed.

2. **Swap Chains**  
Consent is now managed per chain, not per hop. `groupRenameChains` builds the data rows, and requests only carry fully accepted chains (via `filterRenamesByAcceptedChains`). On the server side, any hop that targets a still-live column is now explicitly refused with `target-occupied` rather than being silently skipped (covered in `rename-migration.test.api.js`).

3. **AI applyChange**  
The reducer's `mergeRenameState` now maintains both pending renames and `declinedRenameNames` over a full-type replacement. Within `SchemaProvider`, changes are now applied sequentially using the same `applyChange → prompt-after-edit` workflow as manual edits, guaranteeing that consent is requested (or inherited) consistently.

4. **Type Change**  
When a rename includes a storage type change, users are now blocked with a clear "Existing data will be lost" confirmation dialog in `FormModal` (handled by `getRenameStorageChange`). Partial migration support for compatible types is acknowledged as future work.

innerdvations (commented):
## Change request

Re-reviewed at `b2e9b77acb` (base `develop`), re-review vs `3a8f3ea443`.

All three previous blockers are closed: TS migrations now land where discovery looks (with a JS fallback when no `outDir` resolves), rename consent is resolved per chain instead of per hop, and AI `applyChange` merges pending renames and asks for consent. Three new production issues block merge.

### 1. Refused hops vacate their source name

`addRename` calls `trackVacated` before a hop is known to be supported — on the `target-occupied` early return, and again before the storage-compatibility check:

```ts
if (isTargetOccupied(uid, newName)) {
  trackVacated(uid, oldName, newName);
  // ...
}
trackVacated(uid, oldName, newName);
```

A hop refused as `type-changed` therefore marks its source name as free. A later hop in the same save can then target that name, pass `isTargetOccupied`, and emit a rename onto a column that is still live. At runtime the guard in `rename-helpers` sees the target already exists and skips the DDL, but the migration is recorded as applied, so schema sync drops the unmigrated column — silent data loss.

Repro (`scalarMeta`/`scalarSchema` in the existing unit suite): `oldTitle -> views` with `newAttribute: { type: 'integer' }`, then `summary -> oldTitle`. The second hop should be `target-occupied`; instead it is accepted and an operation is generated.

**Fix in this PR:** only vacate after a hop is fully accepted (after `advance()` records operations), or undo the vacate on both unsupported return paths. Please add a unit case for `refused hop → later hop targeting its source name`.

### 2. Failed saves leave the generated migration on disk

`updateSchema` generates the migration before the schema files and the folder file commit:

```ts
await generateRenameMigrations(schema);
const schemaFilesWritten = await builder.writeFiles();
// ...
await getService('content-structure').commitFromUpdate({ incomingStructure: contentStructure, deletedUids });
```

`rollbackSchemaMutation` compensates the schema builder, generated APIs, and API backups — not the migration file. So when `writeFiles()` fails or `commitFromUpdate` rejects, the user gets a 400 and rolled-back schemas while a rename migration stays under `database/migrations`. On the next boot it renames columns for a schema change that was never committed.

**Fix in this PR:** track the written migration path and delete it as part of the compensating rollback. The two existing rollback tests (`rolls back the schema files when the folder commit fails after writeFiles`, `does not commit the folder file when writeFiles rolls the schema back`) send no renames — worth extending one of them with a `renames` payload asserting the file is gone.

### 3. Cancelling an AI rename prompt does not stop the batch

`applyChange` returns `false` when the user cancels the rename dialog, but `SchemaProvider` ignores it:

```ts
await applyChange({ action: TYPE_TO_ACTION[change.type]!, schema: newSchema });
// ...
setLastRevisedId(latestMessage.id);
```

For a multi-schema AI response, cancelling one prompt still applies every remaining change, and the message is marked fully revised. Single-change messages behave correctly, which is why the provider tests pass.

**Fix in this PR:** break out of the loop when `applyChange` resolves `false`, and skip `setLastRevisedId` on an aborted batch.

---

Everything else I checked at this head looks right: the merge with content-type folders (#27419) kept the intended ordering, `db.metadata.naming` is now shared with the metadata loader, and the guarded helpers plus per-migration transaction behaviour are well covered.

_AI-assisted comment._

jhoward1994:
Thanks @innerdvations , all three are addressed.

**1. Refused hops freed their source name**
- Cause: `addRename` vacated `oldName` before the storage-compatibility check, and dropped it from the in-flight maps before the `type-changed` refusal, so a later hop could rename onto a column that is still live.
- Fix: nothing about the source changes until the hop is accepted. Refused hops (`target-occupied` or `type-changed`) keep their source occupied and in flight, and only accepted hops (including `skip`) vacate it.
- Your suggested fix alone would have missed a related in-flight case: `a -> b`, then `b -> c` refused as `type-changed`, then `x -> b`. `b` was removed from the in-flight map, so `x -> b` was accepted. This is fixed and tested in the same change.

**2. Failed saves left the generated migration on disk**
- Cause: `generateRenameMigrations` writes the file before `builder.writeFiles()`, and `rollbackSchemaMutation` did not know about it.
- Fix: `generateRenameMigrations` returns the written path, `updateSchema` passes it to `rollbackSchemaMutation`, and the rollback removes the file first (best effort, like the other compensations). The CLI `rename:field` / component rename path goes through `updateSchema`, so it gets the fix too. `03-migrations.md` now says a rejected save deletes the generated file.

**3. Cancelling an AI rename prompt did not stop the batch**
- Cause: `SchemaChatProvider` ignored the `false` returned by `applyChange` when the user cancels the rename prompt.
- Fix: the batch stops at the cancelled change, and the message is not marked as revised. Changes applied before it stay applied (they are in the undo history). There is no toast.

innerdvations:
There are some architectural flaws in how we're handling components; we may want to drop support for renaming components in this PR and handle it in a follow-up

Aside from that, we also have to decide if we want to update permissions. I.e., `properties.fields` will still list `myField` under its old name until the next boot, when the permission cleanup drops it, so roles lose access to the renamed field. We should probably point those at the final name of the renamed fields, although I could also see the argument for leaving them dropped and making them be granted again. Either way, this is somewhat of an existing bug on develop that is being magnified here.

PR #26750, opened 2026-06-23
fix(generators): detect plugin language from output path

innerdvations:
## Summary

- Add `getGeneratorLanguage` util that derives TypeScript vs JavaScript template selection from the generator output coordinates (`getDestBasePath()` + `filePath`) instead of `process.cwd()`
- Resolve the `dir` option to an absolute path in `generate()` so relative paths like `server` work reliably
- Replace duplicated language-detection logic across all plugin generators

Fixes standalone plugin generation always emitting JavaScript files when callers pass `dir: 'server'` to the programmatic `generate()` API.

Unblocks [strapi/sdk-plugin#98](https://github.com/strapi/sdk-plugin/pull/98).

## Context

`@strapi/sdk-plugin` adds a `strapi-plugin generate` command that reuses `@strapi/generators`. Standalone TypeScript plugins keep `tsconfig.json` in `server/`, but language detection was checking the plugin root (via `process.cwd()`), so templates were always `.js`.

## Test plan

- [x] Unit tests for `getGeneratorLanguage` (in-app plugin, standalone plugin, non-plugin)
- [x] Existing `@strapi/generators` test suite passes
- [ ] After sdk-plugin bumps `@strapi/generators`, run `strapi-plugin generate` in a TypeScript standalone plugin and confirm `.ts` files are created

### Related issue(s)/PR(s)

Fixes CMS-1319

### DIFF
--- a/packages/generators/generators/src/index.ts
+++ b/packages/generators/generators/src/index.ts
@@ -1,4 +1,4 @@
-import { join, dirname } from 'node:path';
+import { join, dirname, resolve } from 'node:path';
 import handlebars from 'handlebars';
 import fs from 'fs-extra';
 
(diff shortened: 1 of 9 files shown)


Bassel17 (approved):
lgtm

PR #26751, opened 2026-06-23
fix(database): include status sort expression in SELECT when using DISTINCT

### LINKED ISSUE #26746: Content Manager list view returns 500 when sorting by Status with a relation filter (SELECT DISTINCT + ORDER BY status CASE) on PostgreSQL
### Node Version

20.x (LTS)

### Package Manager

npm

### Package Manager Version

10.x — *to confirm with reporter*

### Strapi Version

5.45.0

### Operating System

Windows 11

### Database

PostgreSQL

### Javascript or Typescript

Typescript

### Reproduction URL

*No response — can be reproduced with any two related collection types with Draft & Publish enabled (see Steps to Reproduce).*

### Bug Description

In the Content Manager **list view** of a collection type with Draft & Publish enabled, the view returns a **500** ("Whoops! Something went wrong. Please, try again.") when you **at the same time**:

1. apply a filter that targets a **related** collection type, and
2. sort by **Status**.

Either action on its own works fine — only the combination crashes. Reproduced on PostgreSQL. Because the failure is a PostgreSQL query-planner restriction, it may not reproduce on SQLite/MySQL.

**Generated request:**

```
GET /content-manager/collection-types/api::api.api
  ?page=1&pageSize=10
  &sort=status:ASC
  &filters[$and][0][product][name][$eq]=SurePrep
  &locale=en   → 500
```

**Root cause (analysis):**

* The relation filter (`product.name = …`) makes the query builder JOIN the linked tables and apply `SELECT DISTINCT` to dedupe joined rows.
* Sorting by Status compiles to a computed `ORDER BY CASE … END` ranking (published / modified / draft) that is **not present in the SELECT list**.
* PostgreSQL forbids `SELECT DISTINCT` together with an `ORDER BY` expression that isn't in the select list, so the query throws — surfaced as a 500 in the Content Manager list view.

**Relation to existing reports:** Same origin as strapi/strapi#25395 / [#25403](<https://github.com/strapi/strapi/issues/25403>) / strapi/strapi#25446 (relation-filter JOIN → DISTINCT) but a **different symptom**: those address DISTINCT collapsing rows into a wrong row count, whereas this is a hard **500 from the** `ORDER BY` **status CASE expression** under DISTINCT. Note the FK optimization in strapi/strapi#25446 would **not** cover this case — the filter here is on a non-id field (`name`), which falls back to JOIN+DISTINCT (a limitation a maintainer already flagged on that PR). strapi/strapi#25395 was auto-closed as not planned (stale).

### Steps to Reproduce

Two related collection types with Draft & Publish enabled, e.g. `book` (with a relation to `author`) and `author`:

1. Go to `/admin/content-manager/collection-types/api::book.book`
2. Add a filter on the relation, e.g. **Author is "William Shakespeare"**
3. Click the **Status** column header to sort by status

→ the list view crashes with a 500.

### Expected Behavior

Sorting by Status while a relation filter is applied should return the list, not a 500. The status ranking expression should be added to the DISTINCT select list (or the dedupe done via `GROUP BY` / an id-based subquery) so the `ORDER BY` stays valid.

### Logs

error: ... for SELECT DISTINCT, ORDER BY expressions must appear in select list  <br>at parseErrorMessage (pg-protocol/dist/parser.js)

### Code Snippets

Failing query (trimmed):

```sql
select distinct "t0".*, "t0"."id", ...
from "public"."apis" as "t0"
left join "public"."apis_product_lnk" as "t1" on "t0"."id" = "t1"."api_id"
left join "public"."products" as "t2" on "t1"."product_id" = "t2"."id"
where ("t2"."name" = $1 and "t0"."published_at" is null)
order by CASE
  WHEN NOT EXISTS (SELECT 1 FROM "apis" sub WHERE sub.document_id = t0.document_id
    AND sub.published_at IS NOT NULL AND sub.locale = t0.locale) THEN 0
  WHEN t0.updated_at > (SELECT MAX(sub.updated_at) FROM "apis" sub WHERE ...) THEN 1
  ELSE 2 END ASC
limit $2;
```

### Media

*No response*

### Additional information

The following environment fields were not provided by the original reporter and are **placeholders to be confirmed**: Node version, package manager + version, exact Windows version, and JavaScript/TypeScript. The bug is independent of these — it depends only on PostgreSQL + Draft & Publish + a relation filter combined with sort-by-Status.

This issue is a re-file of strapi/strapi#26745 (auto-closed for not matching the issue template).

### Confirmation Checklist

jasleenkaur-qed42:
## Summary

Fixes #26746

When the Content Manager list view applies a **relation filter** (which forces a JOIN and `SELECT DISTINCT`) **and** sorts by **Status**, PostgreSQL rejects the generated query:

> `for SELECT DISTINCT, ORDER BY expressions must appear in select list`

Status sorting compiles to an `ORDER BY CASE ... END` ranking expression. Before this change, `processSelect()` only added plain column names to the SELECT list when DISTINCT was active, so the status CASE appeared in `ORDER BY` but not in `SELECT`.

This PR adds raw order-by expressions (the status CASE) to the SELECT list whenever DISTINCT is used, using the same `buildStatusSortExpression` helper as the ORDER BY clause so the rendered SQL stays valid on PostgreSQL.

## What changed

- **`query-builder.ts`**: When `shouldUseDistinct()` is true, include `rawExpression` order-by entries (status sort) in `state.select` via `buildStatusSortExpression`.
- **`status-sort-distinct.test.ts`**: Regression test asserting the status CASE appears in the SELECT clause when a JOIN triggers DISTINCT.

## How to reproduce (before fix)

1. Use **PostgreSQL** (this does not reproduce on SQLite).
2. Create two related collection types with **Draft & Publish** enabled.
3. Open the Content Manager list view for the parent type.
4. Filter on a relation field (e.g. author **name**, not id).
5. Click the **Status** column to sort.

**Before:** 500 error; server log shows the PostgreSQL DISTINCT + ORDER BY error.  
**After:** List loads and sorts correctly.

Either action alone (filter **or** status sort) works; only the combination fails.

## Test plan

- [x] Added unit test: `packages/core/database/src/query/__tests__/status-sort-distinct.test.ts`


innerdvations (dismissed):
Seems like the best solution, thanks! I pushed some additional tests as well.

innerdvations (approved):
Re-approved after dropping flaky setTimeout from the API regression test (3b16b150). CI green on the new head.

Issue #26755, opened 2026-06-23
Admin blank on 5.48.1 in pnpm monorepo: #26249 getModulePath resolves @reduxjs/toolkit to a hoisted RTK 2.x (Duplicate middleware references)


IngBertolini:
#### Describe the bug

After upgrading the admin to **5.48.1**, the admin panel renders a blank white page. The browser console shows an unhandled rejection thrown while the admin Redux store is created:

```
Uncaught (in promise) Error: Duplicate middleware references found when creating the store.
Ensure that each middleware is only included once.
    at configureStore (.strapi/vite/deps/chunk-*.js)
    at StrapiApp.render
    at renderAdmin
```

`renderAdmin` never mounts, so the panel stays blank. The backend is fine (`GET /admin` → 200, *Strapi started successfully*); the failure is purely client-side, in the admin bundle.

#### Root cause

This is a regression from **#26249 ("stabilize admin redux deps during upgrade")**, shipped in 5.48.1. That PR adds, in the admin Vite config (`@strapi/strapi/dist/src/node/vite/config.*`), a forced alias for Redux:

```js
resolve: {
  dedupe: [..., 'react-redux', '@reduxjs/toolkit'],
  alias: {
    ...
    'react-redux': getModulePath('react-redux'),
    '@reduxjs/toolkit': getModulePath('@reduxjs/toolkit'),
  },
},
optimizeDeps: { include: [..., 'react-redux', '@reduxjs/toolkit'] },
```

`getModulePath` (`@strapi/strapi/dist/src/node/core/resolve-module.*`) is:

```js
const getModulePath = (mod) => {
  const modulePath = require.resolve(mod);            // <-- bare resolve, unconstrained version
  const pkg = readPkgUp.sync({ cwd: path.dirname(modulePath) });
  return pkg ? path.dirname(pkg.path) : modulePath;
};
```

The bare `require.resolve('@reduxjs/toolkit')` resolves **from `@strapi/strapi`'s own location and walks up the tree**. In a **pnpm monorepo that also contains a package depending on Redux Toolkit 2.x**, that 2.x copy is hoisted to `node_modules/.pnpm/node_modules/@reduxjs/toolkit`, which is exactly what the bare resolve finds first — **not** the `@reduxjs/toolkit@1.9.7` that `@strapi/admin` actually pins.

[...]

PR #26757, opened 2026-06-23
fix(create-strapi-app): scaffold pnpm 11 allowBuilds for Strapi Cloud

innerdvations:
### What does it do?

Scaffolds pnpm-compatible native dependency build config for new Strapi apps:

- **pnpm >= 10.26** (incl. 11): writes `pnpm-workspace.yaml` with `allowBuilds` for Strapi tooling (`@swc/core`, `esbuild`, `sharp`, `core-js-pure`, and `better-sqlite3` for SQLite projects).
- **pnpm < 10.26**: keeps `package.json#pnpm.onlyBuiltDependencies` (legacy allowlist API).
- **pnpm 11+**: adds `minimumReleaseAgeExclude: ['@strapi/*']` so day-0 Strapi npm releases remain installable on Cloud while keeping pnpm's supply-chain defaults for other packages.

Detects pnpm version from `npm_config_user_agent` or `pnpm --version` at scaffold time.

### Why is it needed?

pnpm 11 no longer reads `package.json#pnpm` settings. Our scaffold only wrote `onlyBuiltDependencies: ['better-sqlite3']`, which is ignored on pnpm 11. With `strictDepBuilds`, installs on Strapi Cloud (and local pnpm 11) fail because native/tooling postinstall scripts for `sharp`, `esbuild`, `@swc/core`, etc. never run.

Reported internally (Giulio Montagner) — related to pnpm 11 + Cloud deploy failures, distinct from #26755 (admin Vite alias resolution).

### How to test it?

```bash
yarn nx build create-strapi-app
yarn test:cli --testPathPattern="create-strapi-app/(package-json|pnpm-config|scaffold)"
```

Manual:
1. `npx create-strapi-app@latest my-app --use-pnpm --no-install`
2. Confirm `pnpm-workspace.yaml` contains `allowBuilds` entries on pnpm 11.
3. `cd my-app && pnpm install && pnpm run build`

### Related issue(s)/PR(s)

Related to #26755 (separate admin bundle issue)

### DIFF
--- a/packages/cli/create-strapi-app/src/create-strapi.ts
+++ b/packages/cli/create-strapi-app/src/create-strapi.ts
@@ -11,6 +11,7 @@ import { copyTemplate } from './utils/template';
 import { tryGitInit } from './utils/git';
 import { trackUsage } from './utils/usage';
 import { createPackageJSON } from './utils/package-json';
+import { writePnpmWorkspaceConfig } from './utils/pnpm-config';
 import { generateDotEnv } from './utils/dot-env';
 import { isStderrError } from './types';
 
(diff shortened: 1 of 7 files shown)


giu1io:
@innerdvations 

The code looks good.

I've created published the experimental `0.0.0-experimental.6ad754a83b7af05cdcca2ad294b4fe7c98fc7744` and used it to perform the following tests:

| pnpm version 	| local build 	| cloud build 	|
|--------------	|-------------	|-------------	|
| 9.15.9       	| ✅           	| ✅           	|
| 10.34.4      	| ✅           	| ✅           	|
| 11.9.0       	| ✅           	| ✅           	|

My only questions is the behavior around existing pnpm workspaces. I've seen in the code that we skip creating the `pnpm-workspace.yaml` file if one exists in the parent folder(s). In this case the create-strapi-app command will fail. Is that expected? 
Is the idea that users with this setup will be knowledgable enough to fix the issue by themselves?

giu1io (approved):

PR #26758, opened 2026-06-23
test: add vitest quick-win unit tests across six packages

innerdvations:
### What does it do?

Adds mock-free Vitest unit tests for pure helpers in six packages, following the incremental Jest-to-Vitest migration pattern (`*.vitest.test.ts`, per-package `vitest.config.ts`, `test:unit:vitest` script):

- **@strapi/database** — field types (boolean, string, json, datetime), shared parsers, database errors, `asyncCurry`
- **@strapi/utils** — operators, validate helpers, validation schema modifiers, route serialization, content-api constants/router
- **@strapi/permissions** — permission domain model
- **@strapi/content-type-builder** — server utils (helpers, typeguards, attributes)
- **@strapi/data-transfer** — JSON `diff()` utility
- **@strapi/plugin-color-picker** — `prefixPluginTranslations`

Also includes `vitest.config.ts` in each affected package's `tsconfig.eslint.json` so lint passes.

Existing Jest tests are unchanged.

### Why is it needed?

These modules had little or no unit coverage and are easy to test in isolation. This expands the Vitest footprint and documents current behavior (e.g. `StringField` null→`''`, timestamp formatting) ahead of broader test migration work.

### How to test it?

From repo root:

```bash
yarn test:unit:vitest
```

Or per package, e.g.:

```bash
yarn nx run @strapi/database:test:unit:vitest
yarn nx run @strapi/utils:test:unit:vitest
```

All 198 new Vitest tests should pass.

### Related issue(s)/PR(s)

N/A — incremental test coverage / Vitest migration groundwork.


packages/core/admin/admin/src/components/tests/LazyOutlet.test.tsx
```diff
@@ -21,7 +21,17 @@ const idleNavigation = {
...
+const loadingNavigation = {
+  ...idleNavigation,
+  state: 'loading' as const,
+  location: {
+    pathname: '/',
+    search: '',
+    hash: '',
+    state: null,
+    key: 'default',
+  },
```
nclsndr, line 33: Was it a regression or enhancement?

packages/core/content-type-builder/server/src/utils/__tests__/attributes.vitest.test.ts
```diff
@@ -0,0 +1,71 @@
+import { describe, it, expect } from 'vitest';
+
+import { formatAttribute, isConfigurable, isRelation } from '../attributes';
+
+describe('attributes utils', () => {
+  describe('isConfigurable', () => {
```
nclsndr, line 6:
```suggestion
describe.concurrent('attributes utils', () => {
  describe.concurrent('isConfigurable', () => {
```
[Docs](https://vitest.dev/api/describe.html#describe-concurrent)

packages/core/content-type-builder/vitest.config.ts
nclsndr, line 1: Proposal that might be premature:
Every new package gets identical 13-line config (matches `@strapi/core`, `email-amazon-ses`). Repo now has 9 identical configs.
We could refactor in `vitest-config/presets/package` (e.g. `createPackageVitestConfig(__dirname)`) so each package is one line.

packages/core/admin/admin/src/components/tests/LazyOutlet.test.tsx
nclsndr: ok just saw the git history. Does it make sense to be on its own PR?

nclsndr (changes requested):
Thank you so much for leading the way on the migration @innerdvations!

I like the idea of having this fairly small PR to review the boilerplate before the mass migration.

I see the opportunity of creating a `vitest` skill (monorepo level) in order to steer the implementation following simple heuristics:

---

**Rules**
- no globals (always explicit imports)
  - Vitest: ever
  - Strapi —> warn if no other way found
- Prefer strict comparison over loose, prefer object key comparison over loose object (toBe > toStrictEqual > toEqual, objectContaining…).
- Use `.concurrent` whenever possible ([docs](https://vitest.dev/api/describe.html#describe-concurrent))

**Scope**

- all unit tests can be moved, except whenever we have mocks/spies — (this must be tackled with more ai-sitting)

---

Question:
1. I do not see file deletion here, are they only new tests on uncovered code?

innerdvations:
Yes, it's only new tests. I started out looking to fill gaps in unit test coverage, but didn't want to add them as jest test

packages/core/content-type-builder/vitest.config.ts
innerdvations: I'll do that in a followup and just keep this as "new tests"

packages/core/admin/admin/src/components/tests/LazyOutlet.test.tsx
innerdvations: Just moved it to https://github.com/strapi/strapi/pull/26775

nclsndr (approved):
I trust you on the tested business logic, AI reviewed, no complains. Manually reviewed the setup / boilerplate.

PR #26759, opened 2026-06-23
chore(deps): remove unused @strapi/ts-zen dev dependency

unrevised6419:
### What does it do?

Removes the `@strapi/ts-zen` devDependency from `@strapi/types`, `@strapi/core`, and `@strapi/strapi`, and updates `yarn.lock` accordingly.

### Why is it needed?

`@strapi/ts-zen` was originally added to power the entity-service type tests. Those tests have since been deleted, and a full sweep of the monorepo shows no remaining usage:

- No `import`/`require` of `@strapi/ts-zen` (or `ts-zen`) in any `.ts/.tsx/.js/.mjs/.cjs` source file.
- No reference in any `package.json` script or config.
- Only references were the three `devDependencies` entries themselves (plus Nx cache files).

So the dependency is dead weight — dropping it removes an unused package from installs.

### How to test it?

- `yarn install` resolves cleanly (lockfile no longer contains `@strapi/ts-zen@0.2.0`).
- `yarn version:check` passes (all versions aligned).
- `grep -rn "ts-zen" packages --include="*.ts" --include="*.js"` returns nothing.

### Related issue(s)/PR(s)

N/A

### DIFF
--- a/packages/core/core/package.json
+++ b/packages/core/core/package.json
@@ -117,7 +117,6 @@
     "zod": "3.25.67"
   },
   "devDependencies": {
-    "@strapi/ts-zen": "^0.2.0",
     "@swc/core": "1.13.5",
     "@types/bcryptjs": "2.4.3",
     "@types/configstore": "5.0.1",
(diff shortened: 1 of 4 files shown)


innerdvations (approved):
Thanks!

PR #26764, opened 2026-06-24
chore(*): you probably dont need lodash eslint driven refactor

nclsndr:
### What does it do?

Adds `eslint-plugin-you-dont-need-lodash-underscore@6.14.0` to the shared back and front ESLint configs (`eslint-config-custom`) at the `compatible` severity level. The plugin flags lodash calls that have safe, semantically equivalent native JS replacements.

With the plugin in place, all flagged call sites across the monorepo were audited and migrated in package-scoped batches:

- **Replaced with native equivalents** — `isNil`, `isArray`, `isFunction`, `isString`, `isInteger`, `isNaN`, `isUndefined`, `isNull`, `isDate`, `castArray`, `uniq`, `first`, `find`, `filter`, `map`, `reduce`, `forEach`, `keys`, `values`, `size`, `trim`, `padEnd`, `toLower`, `toUpper`, `capitalize`, `replace`, `assign` (fresh `{}` targets), `get` (static paths), `omit` (static key lists)
- **Left in place with `eslint-disable-next-line` + comment** — `cloneDeep` (structuredClone is not 1:1 for class instances), `defaults` (lodash fills only `undefined` slots; `Object.assign` overwrites existing keys), `get` with dynamic or array paths, `throttle`, and `filter`/`map` inside `lodash/fp` curried pipelines

The result is zero lint errors introduced, the full unit and front-end test suites passing, and a clear audit trail of what was deferred and why.

### Why is it needed?

Lodash is a heavy transitive dependency across the monorepo. Many of its helpers have direct native equivalents in Node ≥22 / modern browsers that require no import at all. Making the remaining, non-replaceable lodash calls explicit (via inline comments) is also a prerequisite for a future dependency-pruning pass.

The ESLint plugin enforces this incrementally: new code that reaches for a replaceable lodash helper gets flagged at lint time, preventing the surface from growing back.

### How to test it?

```bash
yarn lint           # must pass with 0 errors
yarn test:unit      # all unit tests green
yarn test:front     # all frontend tests green
```

No runtime behavior changes — all replacements are semantically equivalent for the input shapes present at each call site.

### Related issue(s)/PR(s)

- built on top of #26760

### DIFF
--- a/packages/utils/typescript/lib/generators/common/models/attributes.js
+++ b/packages/utils/typescript/lib/generators/common/models/attributes.js
@@ -102,7 +102,10 @@ const getAttributeModifiers = (attribute) => {
   }
 
   // Min / Max
-  if (!_.isNil(attribute.min) || !_.isNil(attribute.max)) {
+  if (
+    (attribute.min !== null && attribute.min !== undefined) ||
+    (attribute.max !== null && attribute.max !== undefined)
+  ) {
     const minMaxProperties = _.pick(['min', 'max'], attribute);
     const { min, max } = minMaxProperties;
 
@@ -140,7 +143,10 @@ const getAttributeModifiers = (attribute) => {
   }
 
   // Min length / Max length
-  if (!_.isNil(attribute.minLength) || !_.isNil(attribute.maxLength)) {
+  if (
+    (attribute.minLength !== null && attribute.minLength !== undefined) ||
+    (attribute.maxLength !== null && attribute.maxLength !== undefined)
+  ) {
     const minMaxProperties = _.pick(['minLength', 'maxLength'], attribute);
 
     modifiers.push(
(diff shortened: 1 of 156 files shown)


packages/utils/typescript/lib/generators/common/models/attributes.js
```diff
@@ -102,7 +102,10 @@ const getAttributeModifiers = (attribute) => {
   }
 
   // Min / Max
-  if (!_.isNil(attribute.min) || !_.isNil(attribute.max)) {
+  if (
+    (attribute.min !== null && attribute.min !== undefined) ||
+    (attribute.max !== null && attribute.max !== undefined)
+  ) {
```
unrevised6419, line 108: note(issue): `!isNil(X)` behaviour is different from `X !== null`. Loose equality needs to be used to consider both `null` and `undefined`. `X != null`

nclsndr: Yes, isn't it what we have here?

unrevised6419: 🤦🏼 I completely overlooked the second expression `attribute.min !== undefined`. My bad, it was middle of the night ))

Anyway the shorthand would be

```suggestion
  if (attribute.min != null || attribute.max != null) {
```

nclsndr: Right. I still like the expended version for extra clarity.

PR #26765, opened 2026-06-24
chore(deps): upgrade lint-staged to 16 and scope linting to staged files

unrevised6419:
### What does it do?

Replaces the dynamic `nx`-based `lint-staged` config with a static **nested-config** scheme and upgrades `lint-staged` `15.2.10` → `16.4.0`.

- **`lint-staged.shared.mjs`** — the shared task map:
  - `*.{js,ts,jsx,tsx}` → `eslint --cache --fix` + `prettier --cache --write`
  - everything else → `prettier --cache --write --ignore-unknown`
- **Root `lint-staged.config.mjs`** — a prettier-only catch-all (`'*'` → `prettier --cache --write --ignore-unknown`) for docs, scripts, root files, and any workspace without its own ESLint config.
- **39 per-workspace `lint-staged.config.mjs`** — one in each package that has an `.eslintrc.cjs`, each a one-line re-export of `lint-staged.shared.mjs`.
- Each affected `.eslintrc.cjs` adds `lint-staged.config.mjs` to its `ignorePatterns`.
- Adds `// @ts-check` + JSDoc `Configuration` type to the config files.

The old config walked each staged file up to its nearest `package.json` (via `find-up`), collected the affected workspace package names, and ran `nx run-many -t lint -p <pkgs>`.

**How resolution works:** `lint-staged` matches each staged file to the *closest* config file and does **not** merge configs. A file inside a workspace with its own `lint-staged.config.mjs` runs the shared eslint + prettier tasks; everything else falls through to the root prettier-only config. `lint-staged` 16 runs each config's tasks with the **cwd set to that config's directory**, so `eslint` resolves the package's own `tsconfig` / `.eslintrc.cjs` (ESLint 8's cascading resolution) without going through Nx.

### Why is it needed?

The old approach linted the **entire affected package** for any staged change. Editing a single file — even a JSON file — triggered a full-package lint, making pre-commit slow and noisy. Sometimes merging a branch with `develop` can take over a minute.

Example on MacBook Pro M1 16GB
Merge into PR https://github.com/strapi/strapi/pull/25424 on a clean install & setup

| (5m 54s) of `develop` | (1m 42s) of `worktree-lint-staged` |
|--------|--------|
| https://youtu.be/w_isVaRDrUs | https://youtu.be/ImMnj__ftyw |
| image | image |

The new config runs `eslint --fix` on the **staged files only**, with per-package rules fully preserved. JSON files now go to Prettier only (they were previously sent through ESLint as well), and `eslint --fix` auto-fixes instead of failing the commit on fixable issues.

`lint-staged` 16 also drops several heavy transitive deps in favour of a leaner tree (`tinyexec`, `picomatch`, `commander`, `listr2`). It requires Node ≥ 20.17, within the repo's supported range (`>=20.0.0 <=26.x.x`).

### How to test it?

1. Stage a single file in a package, e.g. `packages/core/utils/src/<file>.ts`, and run `npx lint-staged --verbose`.
   - Only that file is passed to `eslint --fix` (matches `*.{js,ts,jsx,tsx}` in that package's `lint-staged.config.mjs`), not the whole package.
   - The package's own `.eslintrc.cjs` rules are applied (verifiable with `npx eslint --print-config <file>` — resolves the nearest config, no root ESLint config present).
2. Stage a `.json` file in the same package and run `npx lint-staged --verbose`.
   - It matches only the `!(*.js|*.ts|*.jsx|*.tsx)` prettier task; no ESLint and no Nx run.
3. Stage a file outside any configured workspace (e.g. under `docs/`) — it matches the root prettier-only catch-all.

### Related issue(s)/PR(s)

N/A


package.json
```diff
@@ -181,7 +177,7 @@
     "jest-cli": "29.6.0",
     "jest-environment-jsdom": "29.6.1",
     "jest-watch-typeahead": "2.2.2",
-    "lint-staged": "15.2.10",
+    "lint-staged": "16.4.0",
```
unrevised6419, line 184: Note: Latest is v17, but requires Node v22

lint-staged.config.js
```diff
@@ -1,50 +1,10 @@
...
-}
-
-module.exports = {
-  '*.{js,ts,jsx,tsx,json}': getCodeCommands,
-  '*.{md,mdx,css,scss,yaml,yml}': ['prettier --write'],
+/** @type {import('lint-staged').Configuration} */
+const config = {
+  'packages/**/*.{js,ts,jsx,tsx}': ['eslint --fix', 'prettier --write'],
+  '.github/**/*.{js,ts,jsx,tsx}': ['eslint --fix', 'prettier --write'],
+  '*.{md,mdx,css,scss,yaml,yml,json}': ['prettier --write'],
```
innerdvations:
```suggestion
const config = {
  'packages/**/*.{js,ts,jsx,tsx}': ['eslint --fix', 'prettier --write'],
  '.github/**/*.{js,ts,jsx,tsx}': ['eslint --fix', 'prettier --write'],
  '{tests,scripts,examples,docs}/**/*.{js,ts,jsx,tsx}': ['prettier --write'],
  '*.{js,ts,jsx,tsx,cjs,mjs}': ['prettier --write'],
  '*.{md,mdx,css,scss,yaml,yml,json}': ['prettier --write'],
```
Otherwise prettier misses tests, scripts, examples, and root configs

innerdvations (changes requested):

lint-staged.config.js
unrevised6419: This becomes a little bit tricky to configure because we might have overlapping file commands. And might repeat same command more than once on the same file 🤔 
The right way is to use lint-staged monorepo setup, where each workspace has it's own lint-staged config, in that way lint-staged is truly run in parallel for all of them.

Root will have only prettier, each workspace that has eslint configured will have eslint & prettier

innerdvations: True, we shouldn't use a repo-wide `*.{js,…}` glob. But we still need to catch these files.

What do you think about this:
1. Add `{tests,scripts,examples,docs}/**/*.{js,ts,jsx,tsx}` → Prettier only (fixes the coverage gap).
2. Handle root config files with a function that only runs on paths with no `/` (so we don't double-run Prettier on `packages/**`).

Per-workspace lint-staged configs are the right long-term shape, but that's a bigger change we can handle in a follow-up if we need it.

unrevised6419:
## Review

Reviewed the migration from the dynamic `nx run-many` lint-staged config to the static nested-config scheme. **No actionable issues found** — the change is correct and the approach is sound.

I chased four candidate problems; all refuted on verification:

- **eslint coverage regression** — The old config linted all of `packages/**` + `.github/**` via `nx run-many -t lint`. Verified every workspace root with its own `.eslintrc.cjs` received a nested `lint-staged.config.mjs`, and that nested `admin/`/`server/`/`ee/` eslintrcs still apply through eslint's config cascade under their package's config. The roots that got no nested config (`create-strapi`, `utils/{api-tests,tsconfig,vitest-config,eslint-config-custom}`, app templates, `.github/scripts`, `.github/actions/community-pr-triage`) have no `.eslintrc.cjs` and no `lint` target, and there's no root eslintrc — so they were never effectively eslint-linted before either. No coverage lost.
- **config merge / double-run / dropped files** — lint-staged uses the single closest config (no merge), and the two shared keys `'*.{js,ts,jsx,tsx}'` and `'!(*.js|*.ts|*.jsx|*.tsx)'` are exact basename complements: every staged file matches exactly one task list — none double-processed, none dropped.
- **v16 cwd dependency** — Correct per-package tsconfig/eslintrc resolution relies on lint-staged 16's "cwd = config-file directory" behavior. Both the `package.json` pin and `yarn.lock` resolve `16.4.0`, so it's internally consistent.
- **`build.md` `entry.js` → `app.js`** — Verified the admin build genuinely emits `app.js` (`packages/core/strapi/src/node/staticFiles.ts:109`), so the doc edit is factually correct.

**Non-blocking note:** the `build.md` text fix and the incidental prettier reformat churn (rollup configs, html `<!doctype>`, etc.) are unrelated to the lint-staged change. All correct and harmless — might be worth a line in the PR description so reviewers know why those files moved.

LGTM. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

lint-staged.config.js
unrevised6419: I went into per-workspace direction. Please see again the PR body and changes.

You can see the behavior before and after here:

Example on MacBook Pro M1 16GB
Merge into PR https://github.com/strapi/strapi/pull/25424 on a clean install & setup

Before: https://youtu.be/w_isVaRDrUs (5m 54s) of `develop`
After: https://youtu.be/ImMnj__ftyw (1m 42s) of `worktree-lint-staged`

Also closed the PR https://github.com/strapi/strapi/pull/26532 in favor of this one.

innerdvations (approved):
Very nice, thank you!

package.json
unrevised6419: If we wait for https://github.com/strapi/strapi/pull/26631 to be merged, we could upgrade to v17 isntead

innerdvations (approved):

PR #26771, opened 2026-06-24
ci: report bundle size changes for PRs from forks

rzzf:
### What does it do?

The downstream workflow uses `workflow_run` to comment with base repository permissions, while the untrusted fork code is built only in the original `pull_request` workflow.

This follows [GitHub's recommended](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/#:~:text=Below%20is%20an%20example%20of%20the%20intended%20usage%20in%20which%20the%20results%20of%20an%20unprivileged%20pull_request%20workflow%20are%20combined%20with%20a%20privileged%20workflow%20to%20leave%20a%20comment%20in%20response%20to%20a%20received%20PR%3A) pattern for securely handling pull requests from forks.

Note: `workflow_run` only runs after the merge into the default branch, so this PR will not trigger comments.

### Why is it needed?

Previously, `compressed-size-action` could not add comments to PRs opened from forks, so we had to check the results from the workflow logs, which was inconvenient.

Starting from `2.10.0`, it exposes the full `comment-body`, so we can now publish comments for PRs from forks. I hope this helps with project maintenance.

https://github.com/preactjs/compressed-size-action/pull/148

### How to test it?

1. Merge this PR.
2. Create a PR from a forked repository, or merge the default branch into an existing PR to trigger a new workflow run.

### Related issue(s)/PR(s)

See my test here https://github.com/rzzf/strapi/pull/2

---
Fixes CPR-421

### DIFF
--- a/.github/workflows/adminBundleSize.yml
+++ b/.github/workflows/adminBundleSize.yml
@@ -34,6 +34,7 @@ jobs:
         run: printf '#!/bin/sh\nset -e\nyarn install\nyarn nx reset\nyarn build\n' > /tmp/install-and-build.sh
 
       - uses: preactjs/compressed-size-action@v2
+        id: compressed-size
         with:
           install-script: 'sh /tmp/install-and-build.sh'
           build-script: 'build:size'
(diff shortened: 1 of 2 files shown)


innerdvations (approved):
LGTM, thanks!

PR #26775, opened 2026-06-24
fix(admin): satisfy Navigation type in LazyOutlet loading mock

innerdvations:
## Summary

- Fixes `LazyOutlet.test.tsx` mock for `useNavigation()` when `state` is `'loading'`
- react-router's `Navigation` type requires a `location` object in the loading state; spreading `idleNavigation` (where `location` is `undefined`) fails `test:ts:front`

## Test plan

- [x] `yarn nx run @strapi/admin:test:ts:front` passes locally


nclsndr (commented):
I think we're lacking context for testing / QA here.

The PR desc is about launching FE tests while no tests have been modified

innerdvations:
Oops, it didn't get removed from the one it was split from, so it's already fixed on develop. Closing this.

PR #26777, opened 2026-06-24
fix(core): webhook config changes don't propagate across clustered instances

### LINKED ISSUE #22595: Modifying webhooks in a clustered environment will not update the webhook settings in other instances

- Node.js version: 20
- NPM version: v10
- Strapi version: 5.7.0
- Database: SQLite
- Operating system: Linux Mint 22 and Strapi Cloud

### Describe the bug

When running Strapi in a cluster mode and you update a webhook it will only update the internal webhook runner registry of the instance that changed the webhook. All other instances will still have the old webhook config stored in memory.

This will happen when updating, creating, deleting, or disabling any built in webhook

1. Spin up a clustered Strapi instance (easiest to do in Strapi Cloud, ask the support or cloud team to scale your instance manually)
2. Create a content type with a field
3. Create a webhook and set the domain (easiest to use https://webhook.site)
4. Create an entry in the content-type
5. See the webhook response hit the service
6. Change the webhook url in the webhook settings
7. Update or create another entry
8. See that the webhook is fired to both the old and the new domain

### Expected behavior

We need to have a way to watch changes in the database for webhooks and force all Strapi instances to update the in-memory webhook registry

### Screenshots

N/A

### Code snippets

N/A

### Additional context

Related to a customer reported issue in Slack so marking this as EE even if we don't have a ticket reference.
For internal engineers refer to this message in slack: https://strapihq.slack.com/archives/C08723HL9PH/p1736361487169379

raulbalestra:
### What does it do?

Adds a way for the in-memory webhook registry to converge on the persisted
configuration across multiple Strapi instances.

- `WebhookRunner.reload(webhooks)` — replaces the in-memory registry with a
  fresh set of webhooks, tearing down listeners for events that no longer have
  webhooks and registering listeners for new ones (done synchronously so no
  event is missed in between).
- `WebhookRunner.startReloadPolling(interval, load)` / `destroy()` — an opt-in
  poll that periodically reconciles the registry via `reload()`. The timer is
  `unref()`-ed and cleared on `destroy()`, matching the cron/telemetry provider
  lifecycle.
- The webhooks provider wires this up: when `server.webhooks.reloadInterval`
  (in ms) is configured, each instance polls the database and reloads. Default
  behaviour is unchanged (no interval = no polling).

### Why is it needed?

Fixes #22595. The webhook registry is held in memory and is only mutated on the
instance that handles the admin create/update/delete request (the admin
controller calls `webhookRunner.add/update/remove` locally). In a clustered or
horizontally-scaled deployment, every other instance keeps a stale registry
until it restarts, so webhooks fire to old/deleted/disabled targets depending
on which node serves the request.

Reproduced on 5.x with two instances sharing one Postgres database: after
updating a webhook's URL on instance A, A delivered to the new URL while
instance B kept delivering to the old URL even though the database row already
held the new URL.

### How to test it?

Unit tests added in `packages/core/core/src/services/__tests__/webhook-runner.test.ts`
cover: stale config without reload, convergence to the new config after
`reload()`, removal of deleted webhooks (+ listener teardown), registration of
newly added events, and the polling lifecycle (reload on each interval, teardown
on `destroy()`, no timer leak on restart, and resilience to a failing loader).

Manual (cluster): run two instances against one shared database, set
`server.webhooks.reloadInterval` (e.g. `5000`), create a webhook, then update
its URL on one instance and trigger entries on both — both instances should
deliver to the new URL once the interval elapses.

### Related issue(s)/PR(s)

Fix #22595

---
Fixes CPR-422


packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+describe('WebhookRunner - reload', () => {
+  test('reproduces #22595: without reload, a webhook updated elsewhere keeps firing to the stale url', async () => {
+    const { runner, fetch } = createRunner();
+    runner.add(makeWebhook({ url: 'http://localhost/old' }));
+
+    await runner.executeListener({ event: 'entry.create', info: {} });
+
+    expect(fetch).toHaveBeenCalledTimes(1);
+    expect(fetch).toHaveBeenCalledWith('http://localhost/old', expect.any(Object));
+  });
```
nclsndr: > a webhook updated elsewhere

I don't see where we're doing the foreign update here. Maybe this test is only about testing the default case, thus the statement is the thing to edit?

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  headers: {},
+  events: ['entry.create'],
+  isEnabled: true,
+  ...overrides,
+});
+
+describe('WebhookRunner - reload', () => {
+  test('reproduces #22595: without reload, a webhook updated elsewhere keeps firing to the stale url', async () => {
+    const { runner, fetch } = createRunner();
+    runner.add(makeWebhook({ url: 'http://localhost/old' }));
```
nclsndr: url is same as default here

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  });
+
+  test('reloads from the loader on each interval so the registry converges', async () => {
+    const { runner, fetch } = createRunner();
+    runner.add(makeWebhook({ url: 'http://localhost/old' }));
+
+    const load = jest.fn().mockResolvedValue([makeWebhook({ url: 'http://localhost/new' })]);
+    runner.startReloadPolling(5000, load);
+
+    await jest.advanceTimersByTimeAsync(5000);
```
nclsndr: Ensuring enough elapsed time to account for any future flaky with fake timers
```suggestion
    await jest.advanceTimersByTimeAsync(6000);
```

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  test('starting polling again replaces the previous timer (no leak)', async () => {
+    const { runner } = createRunner();
+    const first = jest.fn().mockResolvedValue([]);
+    const second = jest.fn().mockResolvedValue([]);
+
+    runner.startReloadPolling(5000, first);
+    runner.startReloadPolling(5000, second);
+
+    await jest.advanceTimersByTimeAsync(5000);
```
nclsndr:
```suggestion
    await jest.advanceTimersByTimeAsync(6000);
```

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    runner.destroy();
+  });
+
+  test('a failing loader is logged and does not stop subsequent reloads', async () => {
+    const { runner, logger } = createRunner();
+    const load = jest.fn().mockRejectedValueOnce(new Error('db down')).mockResolvedValue([]);
+
+    runner.startReloadPolling(5000, load);
+
+    await jest.advanceTimersByTimeAsync(5000);
```
nclsndr:
```suggestion
    await jest.advanceTimersByTimeAsync(6000);
```

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(logger.error).toHaveBeenCalled();
+
+    await jest.advanceTimersByTimeAsync(5000);
```
nclsndr:
```suggestion
    await jest.advanceTimersByTimeAsync(6000);
```

nclsndr (commented):
Thank you for the PR @raulbalestra 🙌 

Just had a look to code. Looks like being the right—for lack of better primitives—solution. 
Nothing to report on business logic 👌  only comments on testing 👇 

@innerdvations one more candidate for Redis, on the pub/sub side this time.

packages/core/core/src/providers/webhooks.ts
```diff
@@ -17,13 +17,30 @@ export default defineProvider({
...
+      webhookRunner.add(webhook);
+    }
+
+    // The webhook registry lives in memory and is only mutated on the instance
+    // that handled the admin create/update/delete request. In a clustered or
+    // horizontally-scaled deployment, other instances keep a stale registry
+    // until they restart (issue #22595). When `server.webhooks.reloadInterval`
+    // (ms) is set, every instance periodically reloads the registry from the
+    // database so configuration changes converge across the cluster.
+    const reloadInterval = strapi.config.get('server.webhooks.reloadInterval', 0);
```
innerdvations: #22595 isn't fixed while the default is `0` — clustered nodes stay stale unless every deployment sets this. Please default-on with `reloadInterval: 0` to opt out.

```suggestion
    const DEFAULT_WEBHOOK_RELOAD_INTERVAL_MS = 30_000;
    const reloadInterval = strapi.config.get(
      'server.webhooks.reloadInterval',
      DEFAULT_WEBHOOK_RELOAD_INTERVAL_MS
    );
    if (reloadInterval > 0) {
      webhookRunner.startReloadPolling(reloadInterval, () => webhookStore.findWebhooks());
    }
```

packages/core/core/src/services/webhook-runner.ts
```diff
@@ -169,6 +171,66 @@ class WebhookRunner {
...
+    // same tick, so no event can be missed in between.
+    for (const event of [...this.webhooksMap.keys()]) {
+      this.deleteListener(event);
+    }
+    this.webhooksMap.clear();
+
+    for (const webhook of webhooks) {
+      this.add(webhook);
+    }
+  }
```
innerdvations, line 259: Don't tear down every listener on each reload. Listeners are per event type; the map is read at execute time. Only add/remove listeners when an event type appears or disappears.

Also skip when persisted webhooks match memory (`fingerprintWebhooks()` / `fingerprintRegistry()` early return) — needed for safe default-on polling on single-node.

```suggestion
  reload(webhooks: Webhook[]) {
    if (this.fingerprintWebhooks(webhooks) === this.fingerprintRegistry()) {
      debug('Skipping webhook registry reload (unchanged)');
      return;
    }

    debug(`Reloading webhook registry with ${webhooks.length} webhook(s)`);

    const nextMap = new Map<string, Webhook[]>();
    for (const webhook of webhooks) {
      for (const event of webhook.events) {
        const existing = nextMap.get(event);
        if (existing) {
          existing.push(webhook);
        } else {
          nextMap.set(event, [webhook]);
        }
      }
    }

    for (const event of nextMap.keys()) {
      if (!this.webhooksMap.has(event)) {
        this.createListener(event);
      }
    }
    for (const event of this.webhooksMap.keys()) {
      if (!nextMap.has(event)) {
        this.deleteListener(event);
      }
    }

    this.webhooksMap = nextMap;
  }
```

(Add the private fingerprint helpers on the class.)

packages/core/core/src/services/webhook-runner.ts
```diff
@@ -169,6 +171,66 @@ class WebhookRunner {
...
+        this.reload(await load());
+      } catch (error) {
+        this.logger.error('Failed to reload the webhook registry from the database');
+        this.logger.error(error);
+      }
+    }, interval);
+
+    // Do not keep the process alive just for this timer.
+    this.reloadTimer.unref?.();
+  }
```
innerdvations, line 295: `setInterval` doesn't wait for `await load()`. Without a guard, overlapping ticks can apply stale registry state over newer state.

```suggestion
  startReloadPolling(interval: number, load: () => Promise<Webhook[]>) {
    this.stopReloadPolling();

    this.reloadTimer = setInterval(async () => {
      if (this.reloading) {
        return;
      }

      this.reloading = true;
      try {
        const webhooks = await load();
        this.reload(webhooks);
      } catch (error) {
        this.logger.error('Failed to reload the webhook registry from the database');
        this.logger.error(error);
      } finally {
        this.reloading = false;
      }
    }, interval);

    // Do not keep the process alive just for this timer.
    this.reloadTimer.unref?.();
  }
```

(Add `private reloading = false` alongside `reloadTimer`.)

packages/core/core/src/services/webhook-runner.ts
```diff
@@ -169,6 +171,66 @@ class WebhookRunner {
...
+   * by the instance that handled the admin request. In a clustered or
+   * horizontally-scaled deployment every other instance keeps a stale copy
+   * until it restarts. Reloading from the persisted webhooks lets any
+   * instance converge on the current configuration. See issue #22595.
+   */
+  reload(webhooks: Webhook[]) {
+    debug(`Reloading webhook registry with ${webhooks.length} webhook(s)`);
+
+    // Runs synchronously: listeners are torn down and re-registered within the
+    // same tick, so no event can be missed in between.
```
innerdvations: This claim isn't true if we tear down all listeners on every reload (and especially not under load with default-on polling).

```suggestion
    // Runs synchronously. Only event-type listeners that are no longer needed
    // are removed; existing listeners stay registered for unchanged event types.
```

innerdvations:
@nclsndr Yes, I think both a core key-value store and pub/sub that work safely across nodes will help with a lot of issues like this, we should bump the priority on those.

packages/core/types/src/core/config/server.ts
```diff
@@ -67,6 +67,12 @@ export interface Proxy {
 
 export interface Webhooks {
   populateRelations?: boolean;
+  /**
+   * Interval, in milliseconds, at which each instance reloads the in-memory
+   * webhook registry from the database so configuration changes propagate
+   * across a clustered deployment. Disabled when unset or `0`.
+   */
```
innerdvations, line 74:
```suggestion
  /**
   * Interval, in milliseconds, at which each instance reloads the in-memory
   * webhook registry from the database so configuration changes propagate
   * across a clustered deployment. Defaults to 30000. Set to 0 to disable.
   */
```

packages/core/core/src/services/__tests__/webhook-runner.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(load).toHaveBeenCalledTimes(2);
+
+    runner.destroy();
+  });
+});
```
innerdvations, line 181: Please add coverage for:
- `reload()` is a no-op when the registry already matches persisted webhooks (assert `eventHub.off` not called again)
- default-on bootstrap wiring (or unit-test the default constant in one place)

innerdvations (changes requested):
## Changes requested

`reload()` is the right primitive, but this can't ship as a complete fix for #22595 while sync is opt-in only. A few hardening items are also needed before default-on polling is safe on single-node.

1. Default-on `reloadInterval` (opt out with `0`)
2. Skip `reload()` when DB state matches memory
3. Map-swap `reload()` — don't tear down all listeners each tick
4. In-flight guard on polling
5. Tests for skip-unchanged + default-on bootstrap
6. Extend `packages/core/types/src/modules/webhook-runner.ts` with `reload`, `startReloadPolling`, `stopReloadPolling`, `destroy`

See inline suggestions.

_AI-assisted comment._

innerdvations:
All the code change requests were posted by AI, sorry for the terse language it used 😆

raulbalestra:
Thanks for the thorough review @innerdvations, and thanks @nclsndr for going through the tests. Pushed an update that covers everything.

The big one: sync is on by default now. `reloadInterval` defaults to 30s, and you opt out by setting it to `0` — so a clustered deploy converges out of the box without anyone having to know the knob exists. I pulled the default into an exported `DEFAULT_WEBHOOK_RELOAD_INTERVAL_MS` so it's testable and lives in one place.

`reload()` is a lot gentler now. Instead of tearing everything down every tick, it builds the next registry and only touches listeners for event types that actually appeared or disappeared — anything unchanged stays put. And before doing any of that it fingerprints the persisted set against what's in memory and bails early if they're identical, which is the common case once you're polling on a timer. That was the piece that made default-on safe on single-node, like you flagged. I also added the in-flight guard so a slow `load()` can't overlap the next tick and stomp newer state with older.

On the tests: added coverage for the skip-when-unchanged path (asserting we don't churn listeners) and for the bootstrap wiring itself — default interval, opt-out with `0`, custom interval, and that the loader actually reads from the store. @nclsndr — renamed that first test since you were right, there's no foreign update happening there, it's just the baseline; dropped the redundant url override, bumped the fake-timer advances to 6s, and fixed the doc comment to say it defaults to 30000. Also extended the `WebhookRunner` type with the new methods.

One thing I want to put on your radar before this merges: there's still a narrow race I couldn't close cleanly with polling alone. The admin controller writes the DB before it mutates memory, so if a poll's `load()` is already in flight when an admin change commits, it can come back with the pre-change rows and briefly revert memory — it corrects itself on the next tick, so worst case is one interval of staleness. There's no `updatedAt` on `strapi_webhooks` to compare against, so closing it properly really wants the pub/sub primitive you two were talking about rather than another hack here. Happy to note the staleness window in the docs, or just leave it for the Redis work — your call.

innerdvations:
That was quick, thanks! I think it might be ok to merge this and then follow up later with an improvement later, but I'm going to put the 'don't merge' flag for now while we discuss this. I'm mainly concerned with releasing something that might make it harder to implement an ideal solution after we build the tools necessary for it.

raulbalestra:
@innerdvations Thanks! I had a chat with the Support team about this, and we wanted to make sure we're aligned before spending more time on it.

From what we understood, the concern isn't really the polling approach itself, but whether this should be solved as part of the bigger Pub/Sub/Redis work.

Would you rather we keep iterating on this PR, or is this something the Engineering team would prefer to own as part of that larger effort?

Happy to go either way, just wanted to make sure we're heading in the right direction.

innerdvations:
Sorry, I missed your comment! To answer your question, yes, it might be better to wait until we merge some features that would be easier to build this like pub/sub

innerdvations:
Closing this until we have a core pub/sub or some other mechanism to ensure this works properly across horizontally scaled instances

PR #26779, opened 2026-06-24
fix(create-strapi-app): new TypeScript apps use TS strict mode

### LINKED ISSUE #26778: create-strapi-app scaffolds server tsconfig with strict: false
## Node Version

22.x (LTS)

## Package Manager

yarn

## Package Manager Version

4.x

## Strapi Version

5.49.0 (develop; `@strapi/create-strapi-app` templates)

## Operating System

MacOS

## Database

SQLite

## Javascript or Typescript

Typescript

## Reproduction URL

N/A — reproducible from `create-strapi-app` templates in the monorepo.

## Bug Description

New TypeScript Strapi apps created via `create-strapi-app` are scaffolded with **`"strict": false`** in the root/server `tsconfig.json`, while the admin panel config in the same app uses **`"strict": true`**.

The Strapi monorepo itself runs with effective strict mode on all framework packages (`packages/utils/tsconfig/base.json` extends `@tsconfig/node20`; `packages/utils/tsconfig/client.json` sets `"strict": true`). New user projects therefore start with looser server-side type checking than the framework they depend on, which makes it harder to catch type errors early in custom controllers, services, policies, and lifecycle hooks.

This is not a runtime failure — it is a **scaffold inconsistency** that affects every new TS app.

## Steps to Reproduce

1. Create a new TypeScript Strapi app (vanilla or example template):

   ```bash
   npx create-strapi-app@latest my-app --typescript
   ```

   Or inspect the source templates directly in the monorepo:

   - `packages/cli/create-strapi-app/templates/vanilla/tsconfig.json`
   - `packages/cli/create-strapi-app/templates/example/tsconfig.json`

2. Open the generated root `tsconfig.json` and note `"strict": false` under `compilerOptions`.

3. Compare with `src/admin/tsconfig.json` in the same app — it sets `"strict": true`.

4. Compare with monorepo base configs:

   - `packages/utils/tsconfig/base.json` (inherits strict from `@tsconfig/node20`)
   - `packages/utils/tsconfig/client.json` (`"strict": true`)

## Expected Behavior

New TypeScript Strapi apps should ship with **at least `"strict": true`** on the server/root `tsconfig.json`, aligned with the admin config and the monorepo's TypeScript baseline.

If full strict mode is too disruptive for existing users upgrading scaffolds, document the rationale explicitly; the current split (strict admin, loose server) appears unintentional.

## Logs

N/A

## Code Snippets

Server scaffold (`packages/cli/create-strapi-app/templates/vanilla/tsconfig.json`):

```json
{
  "compilerOptions": {
    "strict": false,
    "noImplicitThis": true
  }
}
```

Admin scaffold (`packages/cli/create-strapi-app/templates/vanilla/src/admin/tsconfig.json`):

```json
{
  "compilerOptions": {
    "strict": true
  }
}
```

## Media

N/A

## Additional information

**Complete list of `strict: false` locations** (verified via `rg '"strict"\s*:\s*false' --glob '**/tsconfig*.json'`):

| Path | Role |
|------|------|
| `packages/cli/create-strapi-app/templates/vanilla/tsconfig.json` | Server tsconfig in new-app scaffold |
| `packages/cli/create-strapi-app/templates/example/tsconfig.json` | Server tsconfig in example template |
| `examples/empty/tsconfig.json` | Dev sandbox |
| `examples/complex/tsconfig.json` | Dev sandbox |
| `examples/kitchensink-ts/tsconfig.json` | Dev sandbox |
| `templates/website/tsconfig.json` | Website template |

**Zero** `strict: false` under `packages/**` (framework code).

Several server configs set `"noImplicitThis": true` alongside `"strict": false` — partial strictness, not full strict mode.

**Suggested fix scope (priority order):**

1. `packages/cli/create-strapi-app/templates/*/tsconfig.json` — affects every new app
2. Align `examples/*` and `templates/website` for consistency
3. Run typecheck on affected examples after enabling strict

**Open question for maintainers:** enable full `strict: true` immediately, or tighten incrementally (e.g. `strictNullChecks` first)?

## Confirmation Checklist

nclsndr:
### What does it do?

Aligns the `create-strapi-app` vanilla and example TypeScript scaffolds so the server-side compiler config uses strict mode, matching the admin panel config and the monorepo baseline.

Enabling strict mode surfaced a few config typing gaps in the generated templates. Admin and server configs now assert required environment-backed secrets (JWT, API token salts, encryption key, app keys) where those values are expected at runtime. Database config validates the `DATABASE_CLIENT` value before building the connection and narrows the client type so the scaffold typechecks cleanly under strict mode.

### Why is it needed?

Every new TypeScript Strapi app created via `create-strapi-app` shipped with strict mode disabled on the server while the admin config already used strict mode. That split meant custom controllers, services, policies, and lifecycle hooks were not checked as rigorously as admin code or framework packages, making it harder to catch type errors early in new projects.

### How to test it?

From the monorepo root, scaffold a fresh TypeScript app from the updated templates (vanilla or example) and confirm the project typechecks without errors:

```bash
cd packages/cli/create-strapi-app
yarn build
npx create-strapi-app /tmp/strict-scaffold-test --typescript --no-run --skip-cloud
cd /tmp/strict-scaffold-test
yarn install
yarn strapi ts:generate-types
```

Alternatively, run TypeScript against the template sources directly:

```bash
cd packages/cli/create-strapi-app/templates/vanilla
npx tsc --noEmit
```

Expected: no TypeScript errors. A misconfigured `DATABASE_CLIENT` should throw a clear runtime error at startup rather than silently producing an invalid connection config.

### Related issue(s)/PR(s)

- Companion PR: #26780 — same strict-mode alignment for dev sandboxes and the website template.
- Fixes #26778

### DIFF
--- a/packages/cli/create-strapi-app/templates/example/config/admin.ts
+++ b/packages/cli/create-strapi-app/templates/example/config/admin.ts
@@ -2,18 +2,18 @@ import type { Core } from '@strapi/strapi';
 
 const config = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Admin => ({
   auth: {
-    secret: env('ADMIN_JWT_SECRET'),
+    secret: env('ADMIN_JWT_SECRET')!,
   },
   apiToken: {
-    salt: env('API_TOKEN_SALT'),
+    salt: env('API_TOKEN_SALT')!,
   },
   transfer: {
     token: {
-      salt: env('TRANSFER_TOKEN_SALT'),
+      salt: env('TRANSFER_TOKEN_SALT')!,
     },
   },
   secrets: {
-    encryptionKey: env('ENCRYPTION_KEY'),
+    encryptionKey: env('ENCRYPTION_KEY')!,
   },
   flags: {
     nps: env.bool('FLAG_NPS', true),
--- a/packages/cli/create-strapi-app/templates/example/config/database.ts
+++ b/packages/cli/create-strapi-app/templates/example/config/database.ts
@@ -51,13 +51,21 @@ const config = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Database 
     },
   };
 
+  if (!(client in connections)) {
+    throw new Error(
+      `Unsupported DATABASE_CLIENT: ${client}. Use "postgres", "mysql", or "sqlite".`
+    );
+  }
+
+  type DatabaseClient = keyof typeof connections;
+
   return {
     connection: {
-      client,
-      ...connections[client],
+      client: client as DatabaseClient,
+      ...connections[client as DatabaseClient],
       acquireConnectionTimeout: env.int('DATABASE_CONNECTION_TIMEOUT', 60000),
     },
-  };
+  } as Core.Config.Database;
 };
 
 export default config;
--- a/packages/cli/create-strapi-app/templates/example/config/server.ts
+++ b/packages/cli/create-strapi-app/templates/example/config/server.ts
@@ -4,7 +4,7 @@ const config = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Server =>
   host: env('HOST', '0.0.0.0'),
   port: env.int('PORT', 1337),
   app: {
-    keys: env.array('APP_KEYS'),
+    keys: env.array('APP_KEYS')!,
   },
 });
 
--- a/packages/cli/create-strapi-app/templates/example/tsconfig.json
+++ b/packages/cli/create-strapi-app/templates/example/tsconfig.json
@@ -4,7 +4,7 @@
     "moduleResolution": "Node",
     "lib": ["ES2020"],
     "target": "ES2019",
-    "strict": false,
+    "strict": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "incremental": true,
--- a/packages/cli/create-strapi-app/templates/vanilla/config/admin.ts
+++ b/packages/cli/create-strapi-app/templates/vanilla/config/admin.ts
@@ -2,18 +2,18 @@ import type { Core } from '@strapi/strapi';
 
 const config = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Admin => ({
   auth: {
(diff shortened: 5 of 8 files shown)


packages/cli/create-strapi-app/templates/example/config/admin.ts
```diff
@@ -2,18 +2,18 @@ import type { Core } from '@strapi/strapi';
 
 const config = ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Admin => ({
   auth: {
-    secret: env('ADMIN_JWT_SECRET'),
+    secret: env('ADMIN_JWT_SECRET')!,
```
innerdvations, line 5: Instead of `!` could we use defaults matching the `.env.example` like `env('ADMIN_JWT_SECRET', 'tobemodified')` ?

innerdvations:
There seems to be a legitimate e2e test failure, probably due to this affecting the new project created for the test runs

packages/cli/create-strapi-app/templates/example/config/admin.ts
nclsndr: Especially for critical values like JWT secrets I'd prefer to have an actual failure in fact.
Either we make the `env` helper strict by throwing whenever neither of the process.env and defaults are provided.
If we consider we can't change the `env` helper behavior, I'd probably create `safeEnv` / `envOrThrow` helper to properly account for the expected behavior.
Opinion @innerdvations?

just-sudo-it1: Ok, I agree, that's better. What do you think of an env.required that is guarded? Example in https://github.com/strapi/strapi/pull/26830

nclsndr: @innerdvations what about we merge this one and the #26830 right after?

innerdvations: Sounds good, let's merge this one after release branch is created though so we have some extra time to try to get that one in

innerdvations (approved):

PR #26780, opened 2026-06-24
fix(examples): dev sandboxes use TS strict mode

nclsndr:
### What does it do?

Brings strict TypeScript checking to the server-side compiler config for internal dev sandboxes and the website template, aligning them with the admin panel configs and the monorepo baseline.

Enabling strict mode required typing fixes in config and bootstrap code: database configs now validate and narrow the database client before building connections; the complex example adopts proper `Core.Config` types for its config exports; the website template adds minimal types to its seed bootstrap helpers so nullable file references are handled safely under strict checking.

### Why is it needed?

Several internal TypeScript sandboxes and the website template still ran with strict mode disabled on the server while admin code and framework packages already used strict checking. That inconsistency made it harder to catch type errors in example apps and the website template during development, and diverged from the baseline expected for new user projects after the create-strapi-app scaffold fix.

### How to test it?

From the monorepo root, run TypeScript against each affected sandbox:

```bash
cd examples/empty && npx tsc --noEmit
cd examples/complex && npx tsc --noEmit
cd examples/kitchensink-ts && npx tsc --noEmit
cd templates/website && npx tsc --noEmit
```

Expected: no TypeScript errors in any of the above.

Optionally smoke-test an example app:

```bash
cd examples/empty
yarn develop
```

### Related issue(s)/PR(s)

Companion PR: #26779 — same strict-mode alignment for `create-strapi-app` scaffolds (user-facing new apps).

### DIFF
--- a/examples/complex/config/database.ts
+++ b/examples/complex/config/database.ts
@@ -1,4 +1,6 @@
-export default ({ env }) => {
+import type { Core } from '@strapi/strapi';
+
+export default ({ env }: Core.Config.Shared.ConfigParams): Core.Config.Database => {
   const client = env('DATABASE_CLIENT', 'postgres');
 
   const connections = {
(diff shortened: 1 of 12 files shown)


innerdvations (approved):
Nice, thanks!

examples/complex/config/database.ts
```diff
@@ -48,17 +50,19 @@ export default ({ env }) => {
...
+  type DatabaseClient = keyof typeof connections;
+
   return {
     connection: {
-      client,
-      ...connections[client],
+      client: client as DatabaseClient,
+      ...connections[client as DatabaseClient],
       acquireConnectionTimeout: env.int('DATABASE_CONNECTION_TIMEOUT', 60000),
     },
```
unrevised6419, line 66: Note: @nclsndr I have a fix for the `Core.Config.Database` that will fix these casts.

related PR https://github.com/strapi/strapi/pull/25992/changes#diff-f0ef0367ff529537dd4a7d2eeb14f686523114de100f82f2be5ffc11e9a8d057

PR #26783, opened 2026-06-25
chore(definitions): introduce @strapi/definitions package

nclsndr:
### What does it do?

Introduces `@strapi/definitions`, a new low-level package for passive runtime primitives (constants, schemas) and shared type building blocks used across Strapi packages.

The initial surface includes:

- `constants.regex.BIG_INTEGER` — shared regex for big-integer validation
- `Internal` type namespaces for UID and namespace templates (moved from `@strapi/types`)
- A shared `z` Zod export and `Utils` string helpers

`@strapi/types` keeps its existing public type API but now re-exports from `@strapi/definitions`, acting as a backwards-compatible facade. `@strapi/database` and `@strapi/core` consume the shared `BIG_INTEGER` constant instead of maintaining local copies.

An agent skill documents promotion rules, namespace conventions, and how `@strapi/definitions` relates to `@strapi/strapi` and `@strapi/types`.

### Why is it needed?

Several core packages duplicate the same low-level definitions (regex literals, UID templates) or pull them through `@strapi/types`, which creates unnecessary coupling and makes dependency cycles harder to break. Centralizing passive, dependency-free definitions in a dedicated package gives consumers a stable vocabulary without importing runtime-heavy packages.

This is foundational work — more definitions can be promoted here over time as cycles are untangled.

### How to test it?

1. Build and unit-test the new package:

   ```bash
   yarn nx build @strapi/definitions
   yarn nx test:unit @strapi/definitions
   ```

2. Verify type-checking still passes for affected packages:

   ```bash
   yarn test:ts
   ```

3. Run unit tests for consumers that changed:

   ```bash
   yarn nx test:unit @strapi/database
   yarn nx test:unit @strapi/core
   yarn nx test:unit @strapi/types
   ```

4. Confirm big-integer validation behaviour is unchanged — string coercion and error messages in the database field layer should behave as before, now backed by the shared regex.

### Related issue(s)/PR(s)

None yet — this is new infrastructure.

### DIFF
--- a/.ai/skills/definitions-package/SKILL.md
+++ b/.ai/skills/definitions-package/SKILL.md
@@ -1,0 +1,135 @@
+---
+name: strapi-definitions-package
+description: Guide AI agents when consuming or editing @strapi/definitions. Use when modifying packages/definitions, adding constants/schemas/types there, promoting shared definitions to break package cycles, or deciding how @strapi/definitions relates to @strapi/strapi and @strapi/types.
+---
+
+# Definitions Package
+
+Use this skill before changing `packages/definitions` or importing from `@strapi/definitions`.
+
+## Role Of The Package
+
+`@strapi/definitions` is the source of truth for low-level Strapi definitions:
+
+- passive runtime primitives: constants and schemas
+- dependency-free concepts shared across packages
+
+It is **public low-level**, semver-respected, but not the primary user-facing import path. User-facing APIs should be curated through `@strapi/strapi`.
+
+## Layering
+
+Prefer this direction:
+
+```txt
+@strapi/definitions
+  low-level constants, schemas
+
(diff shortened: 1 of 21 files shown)


innerdvations:
What do you think about how we already have a zod export from strapi/utils? Fine to duplicate here for now (as long as the versions stay locked)?

nclsndr:
> What do you think about how we already have a zod export from strapi/utils? Fine to duplicate here for now (as long as the versions stay locked)?

@innerdvations, the idea here was to export `z` from definitions, re-export it from utils with a deprecation message. Slowly moving it to definitions. Rationale: definitions will need Zod, not lifting the core dep here makes us keep the Zod versions to be perflectly aligned both in definitions and utils

PR #26787, opened 2026-06-25
fix(types): tighten Core.Config typings with backward-compatible deprecations

innerdvations:
## Summary

- Aligns `Core.Config` types with runtime usage: `server.logger.config`, `admin.layout`, `server.cron.tasks`, `Core.Config.Features`, and `Core.Config.TypeScript`
- Marks legacy keys in types with `@deprecated` / JSDoc (`server.globalProxy`, `server.admin.autoOpen`, `emitErrors`, `api.rest.port`, `firstPublishedAtField`, legacy `admin.host`/`port`)
- **No runtime behavior changes** for proxy or auto-open — deprecated keys remain ignored as today
- Adds **startup warnings** when deprecated keys are present:
  - `server.globalProxy` → use `server.proxy.global`
  - `server.admin.autoOpen` → use `admin.autoOpen` in `config/admin.ts`
- Removes the unused `server.admin.autoOpen: false` framework default (was never read)
- Fixes `examples/getstarted` proxy config (`proxy.global`); contributor docs path; upload plugin `security` typing

## Follow-up: config factories + runtime validation

This PR improves static types and adds lightweight deprecation warnings only. A follow-up will add **typed config factories** (for `config/server.ts`, `config/admin.ts`, etc.) that stay in sync with `Core.Config` and perform **runtime validation** at load time — catching invalid shapes, unknown keys, and deprecated options earlier than TypeScript alone can.

That follow-up is especially valuable for **pure JavaScript projects**: today they get no config validation at all (types are compile-time only and are skipped entirely in `.js` apps). Config factories with runtime checks would give JS users the same guardrails TS users get from the type system — invalid or deprecated config would fail fast at startup with a clear error instead of failing silently or misbehaving later.

## Documentation follow-up (separate PR on `strapi/documentation`)

Public docs still describe `admin.host` / `admin.port` as active admin panel server settings. In v5 core these are not read; split deployment uses `admin.url` + `server.url` + `serveAdminPanel`.

Also update server proxy docs to use `server.proxy.global` only (not `server.globalProxy`).

Suggested pages:

- [Admin panel configuration](https://docs.strapi.io/cms/configurations/admin-panel)
- [Host, port, and path customization](https://docs.strapi.io/cms/admin-panel-customization/host-port-path)

## Test plan

- [x] `yarn test:ts` in `packages/core/types`
- [x] `yarn jest packages/core/core/src/configuration/__tests__/server-config.test.ts`
- [ ] CI green
- [ ] Smoke: project with `server.globalProxy` logs a warning and proxy behavior is unchanged
- [ ] Smoke: project with `server.admin.autoOpen` in `config/server.ts` logs a warning; `admin.autoOpen` still controls develop auto-open

### DIFF
--- a/examples/getstarted/config/server.js
+++ b/examples/getstarted/config/server.js
@@ -19,7 +19,9 @@ module.exports = ({ env }) => ({
     populateRelations: env.bool('WEBHOOKS_POPULATE_RELATIONS', true),
   },
   // ℹ️ http_proxy is the env var used by system to set proxy globally
-  globalProxy: env('http_proxy'),
+  proxy: {
+    global: env('http_proxy'),
+  },
   http: {
     serverOptions: {
       requestTimeout: 1000 * 60 * 10, // set request timeout to 600000ms (10 minutes)
--- a/docs/docs/docs/01-core/configuration/00-intro.md
+++ b/docs/docs/docs/01-core/configuration/00-intro.md
@@ -56,6 +56,10 @@ Content API configuration, for example the 'rest.maxLimit' option to set the max
 
 Feature flags for enabling and configuring future features that would be breaking changes and cannot be enabled by default, or experimental features that are not ready for public stable release.
 
+#### typescript
+
+Controls TypeScript tooling during development (for example `autogenerate` to emit types for JavaScript projects).
+
 #### plugins
 
 Contains plugin configurations, with each root-level value the id of a plugin, for example, 'users-permissions'
@@ -91,7 +95,7 @@ NOTE: This is the only configuration that is currently an array rather than an o
 
 ### Typings
 
-Most of the base configurations have been typed, although they are not currently publicly available. They can be found in: `packages/core/types/src/types/core/config`
+Most of the base configurations have been typed, although they are not currently publicly available. They can be found in: `packages/core/types/src/core/config`
 
 Any time a new configuration option is added to Strapi, it must also be added to the types for the appropriate config file (and once the env var loading feature is available, a parser for it such as string, integer, stringArray, etc must be defined)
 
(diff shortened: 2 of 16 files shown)


examples/getstarted/config/server.js
```diff
@@ -19,7 +19,9 @@ module.exports = ({ env }) => ({
     populateRelations: env.bool('WEBHOOKS_POPULATE_RELATIONS', true),
   },
   // ℹ️ http_proxy is the env var used by system to set proxy globally
-  globalProxy: env('http_proxy'),
+  proxy: {
+    global: env('http_proxy'),
+  },
```
nclsndr, line 24: Did we check we don't have this legacy in other example apps?

nclsndr (approved):
Great job @innerdvations 🙌 

Just a thought beyond this scope: We'll need to keep the door open for config interface augmentation for plugin config

innerdvations:
Re: config expansion

For third parties: yes, it must all must live in the plugin config file or user configs, we won't support expansion of Strapi configs beyond that one

For ensuring it stays up-to-date in the future with internal changes:
- the config loaders/get/set will be typed, so if someone attempts to add some code to use a config value they haven't explicitly added to the types, it will not build
- there should be a way to have an automated test ensure nothing unexpected is added/used as well -- although we may not be able to add runtime errors on retrieving an invalid value due to potential breaking change

nclsndr:
> we won't support expansion of Strapi configs beyond that one

Not sure I follow here. We have `config/plugins.ts` file which needs to get its types from augmentation based on what is installed in the given strapi app.
Here do you refer to the scope of that PR only or broader?

innerdvations:
Sorry, I mean the follow-up PR will take that into account, but I wanted to be clear that ONLY plugins.ts will, not any other configs

examples/getstarted/config/server.js
innerdvations: Just checked, only examples/getstarted had `globalProxy`. The other example apps don’t set proxy config at all. Also scanned examples for the other deprecated keys from this PR (server.admin.autoOpen, etc) just to be safe

PR #26790, opened 2026-06-25
chore(typescript): enable erasableSyntaxOnly and noUncheckedSideEffectImports

unrevised6419:
### What does it do?

Enables two TypeScript compiler flags in the shared tsconfig presets (`packages/utils/tsconfig/base.json` and `client.json`) and migrates the existing violations so the whole monorepo type-checks cleanly:

- **`noUncheckedSideEffectImports`** ([TS 5.6](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#the---nouncheckedsideeffectimports-option)) — errors on side-effect-only imports (`import "./x"`) whose specifier resolves to nothing. Previously the compiler silently ignored unresolved side-effect imports, so typo'd or dead module paths went undetected.
- **`erasableSyntaxOnly`** ([TS 5.8](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8/#the---erasablesyntaxonly-option)) — forbids TS-only runtime constructs that can't be removed by pure type-stripping: `enum`, constructor parameter properties, namespaces containing runtime values, and `import =`. Keeps the codebase compatible with Node's native type stripping (`--experimental-strip-types`) and other erase-only transpilers (esbuild/swc/Babel `isolatedModules`-style).

Migrations required to pass both flags repo-wide:

- `enum` → `as const` object + same-named union type:
  - `upload` `AssetSource`, `DocType`, `AssetType` (incl. `future/`)
  - `data-transfer` `ErrorKind`
  - `upgrade` `ReleaseType`
- constructor parameter properties → explicit field declarations + assignment:
  - `data-transfer` `ServerError`
  - `core` `session-manager`
  - `IntersectionObserver` mocks in `admin` and `review-workflows` tests

### Why is it needed?

We're at the very start of the TypeScript 5.x upgrade, which is the cheapest possible moment to tighten compiler settings:

- **Smallest blast radius now.** The number of violations only grows over time. Enabling the flags today means migrating a handful of files instead of hundreds later. The full migration here is ~10 files.
- **Prevents regressions for free.** Once enabled, CI rejects any new `enum` / parameter property / unresolved side-effect import, so the codebase can't drift back. Retrofitting later is a one-off cost *plus* whatever accumulated in the meantime.
- **Forward-compatibility.** `erasableSyntaxOnly` aligns the source with type-stripping runtimes (Node ≥22 native strip, Bun, ts-blank-space). `noUncheckedSideEffectImports` is a pure correctness win — it can only catch real bugs (broken import paths), never reject valid code.

If `erasableSyntaxOnly` is considered too opinionated to adopt right now, `noUncheckedSideEffectImports` stands on its own and should be enabled regardless — it has no stylistic trade-off, only catches mistakes.

#### On dropping `enum`

`erasableSyntaxOnly` forces the move away from TS `enum`, which is widely regarded as a misfeature — it emits runtime code (unlike the rest of the type system), has surprising semantics (numeric enums are bidirectional and not type-safe, `const enum` breaks under isolated modules), and an `as const` object + union type is a strictly simpler, fully-erasable replacement. Further reading:

- [TS Handbook — Objects vs Enums](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) (the docs themselves recommend `as const` objects)
- [Total TypeScript — Why I don't like TypeScript enums](https://www.totaltypescript.com/why-i-dont-like-typescript-enums)
- [LogRocket — Why TypeScript enums suck](https://blog.logrocket.com/why-typescript-enums-suck/)
- [Nine terrible ways to use TypeScript enums (and one good way)](https://bluepnume.medium.com/nine-terrible-ways-to-use-typescript-enums-and-one-good-way-f9c7ec68bf15)
- [youdontneedtsenums.com](https://www.youdontneedtsenums.com/)

This is the **follow-up PR** to the TypeScript 5.9.3 upgrade (#26782), now merged to `develop`. It was originally stacked on that PR's branch; rebased onto `develop` after the upgrade landed. The migration surface is still minimal — best to enable now.

### How to test it?

```bash
yarn test:ts   # passes across the monorepo with both flags enabled
```

### Related issue(s)/PR(s)

Follow-up to #26782 (chore(deps): upgrade TypeScript to 5.9.3), now merged. Supersedes #26784 (closed when the upgrade branch was merged into `develop`).


nclsndr:
Previous review https://github.com/strapi/strapi/pull/26784#pullrequestreview-4569524115

unrevised6419:
@coderabbitai Fixed in 20494fefed — dropped the redundant `public message` field and its manual assignment; `super(message)` now initializes the inherited `Error.message`. `code` and `details` kept. Build typecheck green.

packages/core/data-transfer/src/types/remote/protocol/server/error.ts
```diff
@@ -1,18 +1,23 @@
...
+  Unknown: 0,
   // Chunk transfer
-  DiscardChunk = 1,
-  InvalidChunkFormat = 2,
-}
+  DiscardChunk: 1,
+  InvalidChunkFormat: 2,
+} as const satisfies Record<string, ErrorKind>;
+
+// eslint-disable-next-line @typescript-eslint/no-redeclare
```
nclsndr: the linter exception is not placed for all cases. I would expect them to be all or none.

On that note, I'm not really fan of having types shadowing the names of constants. What about turning the const to SCREAMING_SNAKE_CASE?

--- 

AI findings: 
**Back packages** (`data-transfer`, `upgrade`) need `// eslint-disable-next-line @typescript-eslint/no-redeclare` because the back ESLint config (via `@strapi/eslint-config/back/typescript`) enables `@typescript-eslint/no-redeclare`. **Front packages** (upload admin) get the exact same `const X + type X` pattern for free — no suppress needed.

packages/core/upload/admin/src/future/enums.ts
```diff
@@ -1,6 +1,8 @@
...
-  Audio = 'audio',
-}
+export const AssetType = {
+  Video: 'video',
+  Image: 'image',
+  Document: 'doc',
+  Audio: 'audio',
+} as const satisfies Record<string, AssetType>;
+
+export type AssetType = 'video' | 'image' | 'doc' | 'audio';
```
nclsndr: A bit out of scope but bearable:

`AssetType` duplicated in upload admin `future/enums.ts`

- **Where:** `packages/core/upload/admin/src/future/enums.ts` (L1–8) duplicates `AssetType` from `packages/core/upload/admin/src/enums.ts` (L10–17). Sole consumer: `packages/core/upload/admin/src/future/utils/getAssetIcon.ts` (L15), which imports from `../enums` (the duplicate file).
- **What:** `future/enums.ts` is a byte-for-byte copy of the `AssetType` const + type from the parent `enums.ts`. Two definitions of the same symbol with no divergence today, but any future edit to one file can silently drift from the other.
- **Fix idea:** Re-export from parent (`export { AssetType } from '../enums'; export type { AssetType } from '../enums';`) or point `getAssetIcon.ts` at `../../enums` and delete `future/enums.ts`.

packages/core/data-transfer/src/types/remote/protocol/server/error.ts
```diff
@@ -1,18 +1,23 @@
...
+export const ErrorKind = {
   // Generic
-  Unknown = 0,
+  Unknown: 0,
   // Chunk transfer
-  DiscardChunk = 1,
-  InvalidChunkFormat = 2,
-}
+  DiscardChunk: 1,
+  InvalidChunkFormat: 2,
```
nclsndr: Since we also have issues with AssetType. Let's open the opportunity to refactor to string as well.

nclsndr (commented):
Thank you for the neat contrib @unrevised6419 🙌 

Couple of items 👇

unrevised6419:
### Keeping the runtime const maps (with a naming tweak)

While dropping the const-object enums in favour of bare literal unions, I kept the runtime const map in three spots — `ReleaseTypes` (`@strapi/upgrade`) and `DocTypes` / `AssetTypes` (`@strapi/upload`) — and reworked the pattern instead of removing it.

#### New pattern

```ts
export const ReleaseTypes = {
  Major: 'major',
  Minor: 'minor',
  Patch: 'patch',
  Latest: 'latest',
} as const;

export type ReleaseType = (typeof ReleaseTypes)[keyof typeof ReleaseTypes];
```

- `erasableSyntaxOnly`-safe (no `enum`).
- Const named in the **plural**, type stays **singular** → no shared identifier, no declaration merging, drops the old `// eslint-disable @typescript-eslint/no-redeclare`.
- Type **derived from** the const → the two can't drift, and the self-referential `satisfies Record<Capitalize<X>, X>` is gone.

#### Why not collapse to a bare union

The values are needed **at runtime**, which a union can't give us without re-typing raw literals everywhere (and losing rename/typo safety):

**Runtime validators** — the clearest case:

```ts
// upgrade — isSemVerReleaseType
return Object.values(ReleaseTypes).includes(str as ReleaseType);
```

Drop the const and the only way to enumerate members at runtime is to hand-write the list again. That copy isn't tied to the union, so it silently drifts when a member is added. Same for the `DocTypes` / `AssetTypes` mime checks in `upload`.

**Tests** — both the argument and the `toBe` expected value go from a checked reference to a raw literal:

```ts
// argument
expect(isSemVerReleaseType(ReleaseTypes.Major)).toBe(true); // typo'd member → compile error
expect(isSemVerReleaseType('major')).toBe(true);            // bare union → 'majr' still compiles, rename won't propagate

// expected value
expect(typeFromMime('image/png')).toBe(AssetTypes.Image);   // checked reference
expect(typeFromMime('image/png')).toBe('image');            // raw literal — no check, drifts on rename
```

#### Public-API check (re: the rename)

- **`@strapi/upload` — `DocTypes` / `AssetTypes`: internal.** Not in the public entry (`admin/src/index.ts`), no barrel re-exports `../enums`. Safe to rename.
- **`@strapi/upgrade` — `ReleaseTypes`: technically public**, reachable as `modules.version.Version.ReleaseTypes` via the root barrel. However it's the `strapi upgrade` CLI (`bin`), consumed internally through relative imports — **zero consumers of the programmatic `modules` export anywhere in the monorepo**. Treating the rename as acceptable, but flagging it explicitly since it's a public symbol.

Verified green: `test:ts` (both packages), `eslint`, `prettier`, `upgrade` unit (162), `upload` front (22).

@nclsndr — does this carve-out look right, or would you prefer these also collapse to bare unions (and accept the runtime/test trade-off)?

packages/core/upload/admin/src/components/UploadAssetDialog/AddAssetStep/FromComputerForm.tsx
```diff
@@ -59,11 +58,6 @@ export const FromComputerForm = ({
 
   const handleDragLeave = () => setDragOver(false);
 
-  const handleClick = (e: React.MouseEvent<HTMLButtonElement>) => {
-    e.preventDefault();
-    // inputRef.current?.click();
-  };
```
unrevised6419, line 65: Note: Unused handler

packages/core/upload/admin/src/utils/urlsToAssets.ts
```diff
@@ -26,8 +26,8 @@ export const urlsToAssets = async (urls: string[]) => {
   // Retrieve the assets metadata
   const assetsResults = await Promise.all(assetPromises);
 
-  const assets = assetsResults.map((fullFilledAsset) => ({
-    source: AssetSource.Url,
+  const assets = assetsResults.map<FileWithRawFile>((fullFilledAsset) => ({
+    source: 'url' as const,
```
unrevised6419, line 30: Thought: For some reason this `map` looks identical to `packages/core/upload/admin/src/utils/rawFileToAsset.ts#rawFileToAsset` the source field it seems is not used anywhere

nclsndr:
> does this carve-out look right

Thank you @unrevised6419 for exploring this refactor!
About bare unions: by reading the changes it feels more rational to keep const-object enums for all cases, ensuring consistency.
On const naming, having plural and singular forms to distinguish const vs type is a valid pattern. Yet, reading `ReleaseTypes` sounds off to me: it looks like a class, but plural. What about exploring `const RELEASE_TYPES` and `type ReleaseType`? Rationale: make constants visually consistant and standout by using SCREAMING_SNAKE_CASE.

packages/core/upload/admin/src/utils/urlsToAssets.ts
nclsndr: I'm not knowledgeable enough on that area. Probably better to not change any behavior in this PR. We could study it in another PR with the right reviewers

packages/core/upload/admin/src/constants.ts
```diff
@@ -1,9 +1,11 @@
...
-export enum AssetSource {
-  Url = 'url',
-  Computer = 'computer',
-}
+export const ASSET_SOURCES = {
+  Url: 'url',
+  Computer: 'computer',
+} as const;
+
+export type AssetSource = (typeof ASSET_SOURCES)[keyof typeof ASSET_SOURCES];
```
nclsndr, line 8: I was puzzled between this form and keeping the `satisfies` statement with the `Capitalize<>` helper.
But reading them all, it makes more sense to always have the same pattern: `const MY_CONST = {} as const` + `type MyConst = (typeof MY_CONST)[keyof typeof MY_CONST]`

nclsndr (approved):
Amazing work @unrevised6419 🙌 
I'm so happy to see those changes being merged before even scheduled!

PR #26791, opened 2026-06-25
fix(core): preserve M2M relation order on published version after reo…

### LINKED ISSUE #26786: Republishing does not sync reordered M2M relation order from draft to published
### Node Version

22.22.3

### Package Manager

npm

### Package Manager Version

21.7.3

### Strapi Version

5.48.0

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Javascript

### Reproduction URL

_No response_

### Bug Description

After reordering manyToMany relations on the **draft** and clicking **Publish** again, the **Draft** tab shows the new order but the **Published** tab (and GraphQL) still show the **previous** published order.

## Notes

Join table (`articles_categories_lnk`) shows correct `category_ord` on the **draft** row but stale order on the **new published** row after republish.

Likely related to `captureJoinBatches` in `packages/core/core/src/services/document-service/utils/bidirectional-relations.ts`: on republish for an existing locale, `draftsOnly` is empty so order is copied from **old published** join rows instead of **draft** rows.

### Steps to Reproduce

1. Strapi 5.48, `examples/getstarted` (Article ↔ Category, draft & publish enabled)
2. Create and publish categories: Alpha, Beta, Gama
3. Create a new article, connect categories in order: Gama → Alpha → Beta
4. Publish once (order correct everywhere)
5. On draft, drag-reorder to e.g. Gama → Beta → Alpha
6. Save and **Publish** again

### Expected Behavior

Published tab and GraphQL match the draft order after republish.

### Logs

```shell
- **Draft tab:** Gama → Beta → Alpha ✓
- **Published tab:** Gama → Alpha → Beta (old order) ✗
- **GraphQL** (default published): same as published tab ✗
```

### Code Snippets

_No response_

### Media

_No response_

### Additional information

_No response_

### Confirmation Checklist

jasleenkaur-qed42:
## Summary

Fixes #26786 

For a **bidirectional many-to-many** relation with **Draft & Publish**, reordering the
related entries in the draft and re-publishing did **not** update the order on the
published version. The draft showed the new order, but the published tab (and GraphQL,
which reads the published version) kept the previous order.

Root cause is in `bidirectionalRelations` during `repository.publish()`:

1. The draft is read in the correct (reordered) order.
2. `bidirectionalRelations.load()` captures the **old published** entry's join rows —
   which still hold the stale order — *before* they are deleted.
3. The old published row is deleted and `entries.publish()` recreates it from the draft,
   writing the correct order.
4. `bidirectionalRelations.sync()` then runs a `CASE` `UPDATE` that **overwrites the
   freshly-written owning-side order with the stale captured order**, silently reverting
   the reorder.

The owning-side order is always rebuilt from the draft by `entries.publish()`, so it never
needs restoring on republish. Only the **inverse side** (the related entries' ordering of
this entry) is disturbed by a republish and must be restored. This PR skips the
old-published capture for the **owning side only**, leaving the inverse-side restoration
and the `draftsOnly` capture (related entry published later) untouched.

## What changed

- **`bidirectional-relations.ts`**: thread an `isOwningSide` flag into `captureJoinBatches`
  and skip the old-published rows capture when it is the owning side. Those rows are
  recreated in draft order by `entries.publish()`, so capturing them only lets `sync()`
  overwrite the freshly-published order with a stale one. The inverse side still uses the
  capture (needed to restore the related entries' ordering), and the `draftsOnly` capture
  is unaffected.
- **`m2m-order-republish.test.api.ts`**: regression test for the full-array `set` path —
  reorder draft, publish, assert published matches the draft.
- **`m2m-order-republish-connect.test.api.ts`**: regression test for the admin path —
  reorder via `connect` + `position` directives (including a single-item drag), across
  multiple reorder/publish cycles, asserting draft and published stay in sync.

## How to reproduce (before fix)

1. Enable **Draft & Publish** on two collection types with a many-to-many relation
   (e.g. Article ⇄ Category).
2. Create and publish categories Alpha, Beta, Gamma.
3. Create Article 1 with categories `Beta → Alpha → Gamma` and publish. Draft and Published
   both show `Beta → Alpha → Gamma`.
4. In the draft, reorder to `Alpha → Gamma → Beta` and publish.

**Before:** Published tab stays `Beta → Alpha → Gamma`; the GraphQL/REST response for the
published version returns the same stale order.
**After:** Draft tab, Published tab, and the API response all return `Alpha → Gamma → Beta`.

Reproduces on all databases (it is order-write logic, not dialect-specific SQL).

## Test plan

- [x] Added: `tests/api/core/strapi/document-service/relations/m2m-order-republish.test.api.ts`
- [x] Added: `tests/api/core/strapi/document-service/relations/m2m-order-republish-connect.test.api.ts`
- [x] Existing relation suites still pass: `tests/api/core/strapi/document-service/relations/` (9 suites, 54 tests) — including `bidirectional-relations` (publish-related-entry, interleaved publishes) and `invisible-relation-republish`
- [x] `publish` / `unpublish` / `discard-draft` document-service suites pass
- [x] `prettier:check`, `eslint` (0 errors), `tsc -p tsconfig.build.json --noEmit` (exit 0)
- [ ] Verified locally on **SQLite**; PostgreSQL/MySQL covered by CI
- [x] Added: `tests/api/core/strapi/document-service/relations/m2m-order-republish.test.api.ts`
      (covers both the full-array `set` path and the admin `connect` + `position` reorder path)

### DIFF
--- a/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
+++ b/packages/core/core/src/services/document-service/utils/bidirectional-relations.ts
@@ -83,6 +83,15 @@ const captureJoinBatches = async (
     relatedHasDraftAndPublish: boolean;
     /** Model that owns this attribute; draft capture is skipped for components. */
     schemaUid: UID.ContentType;
+    /**
+     * True when the attribute being captured is the owning side of the relation and the entity
+     * being published owns it (e.g. publishing an Article for `Article.categories`). In that case
+     * the owning-side order is rebuilt from the draft by `entries.publish`, so capturing the old
+     * published rows would only let `sync()` overwrite the freshly published order with a stale
+     * one — silently reverting a reorder made in the draft. The inverse-side order still needs the
+     * old-published capture to be restored, so this only suppresses the owning-side capture.
+     */
+    isOwningSide: boolean;
     oldVersions: LoadContext['oldVersions'];
     newVersions: LoadContext['newVersions'];
   }
@@ -94,6 +103,7 @@ const captureJoinBatches = async (
     relatedUid,
     relatedHasDraftAndPublish,
     schemaUid,
+    isOwningSide,
     oldVersions,
     newVersions,
   } = opts;
@@ -101,7 +111,11 @@ const captureJoinBatches = async (
   const batches: RelationEntry[] = [];
   const { name: table } = joinTable;
 
-  const oldIds = oldVersions.map((e) => e.id);
+  // Skip the old-published capture for the owning side: those rows are recreated in draft order
+  // by `entries.publish`, so restoring their order from the old published version is at best a
+  // no-op and at worst reverts a draft reorder. The draftsOnly capture below is still needed
+  // (e.g. when a related entry is published after this one) and is not affected by this guard.
+  const oldIds = isOwningSide ? [] : oldVersions.map((e) => e.id);
   if (oldIds.length > 0) {
     const existing = await strapi.db
       .getConnection()
@@ -254,6 +268,7 @@ const load = async (uid: UID.ContentType, { oldVersions, newVersions }: LoadCont
             ? !!strapi.contentTypes[relatedUid]?.options?.draftAndPublish
             : !!model.options?.draftAndPublish,
           schemaUid: model.uid as UID.ContentType,
+          isOwningSide,
           oldVersions,
           newVersions,
         });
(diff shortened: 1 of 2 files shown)


innerdvations (approved):
LGTM, thanks again for all the fixes lately! I added some more tests here too :)

PR #26792, opened 2026-06-25
chore(tooling): remove unused find-up after lint-staged 16

innerdvations:
### What does it do?

Small follow-up cleanups after #26765 (lint-staged 16 upgrade):

- Remove unused root `find-up` devDependency (only used by the deleted dynamic `lint-staged.config.js`)
- Add trailing newline to `.prettierignore`

### Why is it needed?

These were flagged as non-blocking review items on #26765. Keeping them separate keeps that PR focused on the lint-staged migration while still landing the dependency and config hygiene.

A root-only `engines.node` bump to `>=20.17.0` was dropped after review: published packages still declare `>=20.0.0`, and #26631 will update all `engines.node` ranges to `>=22.0.0` in one coordinated change.

### How to test it?

1. `yarn install` — succeeds; `find-up` is no longer a direct root devDependency
2. Stage a root file and run `npx lint-staged --verbose` — pre-commit hook still runs

### Related issue(s)/PR(s)

- Follow-up to #26765
- Related: #26631 (coordinated Node 22 minimum)


.prettierignore
```diff
@@ -12,4 +12,4 @@ tests/e2e/data/**
 **/tmp-dts-export/**
 /docs/.docusaurus
 /docs/docs/exports
-/packages/generators/generators/src/templates/**/*.hbs
\ No newline at end of file
+/packages/generators/generators/src/templates/**/*.hbs
```
unrevised6419, line 15: Classic mistake I always do )) I hate it 😄

package.json
```diff
@@ -206,7 +205,7 @@
   },
   "packageManager": "yarn@4.12.0",
   "engines": {
-    "node": ">=20.0.0 <=26.x.x",
+    "node": ">=20.17.0 <=26.x.x",
```
nclsndr: Why changing only this engine? Does that mean we un-correlate engines from the npm packages vs the monorepo?

nclsndr (dismissed):
Just a question, I let you make the call @innerdvations

package.json
innerdvations: Nope, we should do it for all of them. I actually thought we had bumped the min to 20.19 recently, maybe that's something that hasn't been merged yet though.

nclsndr: I remember seeing changes toward 20.19 but not merged yet. Is lint-staged requiring the `>=20.17.0`?

innerdvations: I'll remove the bump actually since it's tooling-only and our CI will never attempt to run on 20

nclsndr (approved):
LGTM 👌

PR #26800, opened 2026-06-25
fix(review-workflows): add server eslint config and declare server deps

unrevised6419:
### What does it do?

Adds the missing `packages/core/review-workflows/server/.eslintrc.cjs` so the package's server source is actually linted, mirroring every other admin/server-split workspace (`content-manager`, `upload`, `content-releases`, …). It also:

- declares the `koa` and `lodash` imports in `review-workflows/package.json` dependencies (pinned to the same versions used elsewhere in the monorepo: `koa@2.16.4`, `lodash@4.18.1`);
- drops the now-flagged unused imports `validate` (in `server/src/controllers/stages.ts`) and `property` (in `server/src/controllers/workflows.ts`).

### Why is it needed?

The `review-workflows` root ESLint config excludes `server/**/*`, delegating those files to a nested `server/.eslintrc.cjs` — exactly like the other workspaces. But `review-workflows` was the **only** such workspace shipping no nested server config, so its server source was never linted.

This stayed dormant while ESLint ran from the repo root, but once lint-staged began running ESLint with the cwd scoped to each workspace, the server files matched no config and failed to parse:

```
server/src/controllers/stages.ts
  1:1  error  Parsing error: The keyword 'import' is reserved
server/src/controllers/workflows.ts
  1:1  error  Parsing error: The keyword 'import' is reserved
```

Enabling the linter then surfaced the pre-existing undeclared dependencies and dead imports, which are fixed here too. An audit of every admin/server-split workspace confirms `review-workflows/server` was the only genuine gap.

### How to test it?

```bash
cd packages/core/review-workflows
npx eslint server/src/controllers/stages.ts server/src/controllers/workflows.ts
# => No issues found (previously: "Parsing error: The keyword 'import' is reserved")
```

### Related issue(s)/PR(s)

Surfaced by the per-workspace lint-staged scoping work.


unrevised6419:
will push more fixes here for lint errors

unrevised6419:
have left to fix one eslint misconfiguration

unrevised6419:
@innerdvations please trigger CI for this one. All lint errors fixed.

unrevised6419:
@nclsndr please trigger CI for this one

unrevised6419:
The failling job I think is something that was wrong in `develop` but fixed

unrevised6419:
@nclsndr please trigger CI again

packages/core/review-workflows/server/src/migrations/shorten-stage-attribute.ts
nclsndr, line 1: This function is actually dead code

packages/core/review-workflows/server/src/destroy.ts
```diff
@@ -1,3 +1,4 @@
 import type { Core } from '@strapi/types';
 
-export default async ({ strapi }: { strapi: Core.Strapi }) => {};
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export default async (_args: { strapi: Core.Strapi }) => {};
```
nclsndr: unused -> removed

packages/core/review-workflows/server/src/bootstrap.ts
```diff
@@ -38,7 +39,8 @@ const registerWebhookEvents = async () =>
     strapi.get('webhookStore').addAllowedEvent(eventKey, event)
   );
 
-export default async (args: any) => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export default async (_args: { strapi: Core.Strapi }) => {
```
nclsndr: Should actually be used

nclsndr (changes requested):
Thank you @unrevised6419 👌 
I'll apply changes at the same time

nclsndr (approved):

PR #26803, opened 2026-06-26
ci: update issues-helper action to backup version

rzzf:
### What does it do?

The original repository has been disabled by GitHub, so the official backup repository is used instead.

See: https://github.com/actions-cool/#-remind

### Why is it needed?

Before this change, the workflow would never succeed, and we would receive failure notification emails every day. This creates unnecessary noise.

https://github.com/strapi/strapi/actions/workflows/issues_dailyCron.yml

### How to test it?

Merged this PR.

### Related issue(s)/PR(s)

No.

### DIFF
--- a/.github/workflows/issues_dailyCron.yml
+++ b/.github/workflows/issues_dailyCron.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: check for inactive issues that can't be reproduced
-        uses: actions-cool/issues-helper@v3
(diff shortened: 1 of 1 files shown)


innerdvations:
Thanks for putting this together — you were right that `actions-cool/issues-helper` is TOS-blocked and our workflows were failing at action download.

We're taking a slightly different approach instead: replacing all `issues-helper` usage with the official `actions/github-script` action. That removes the third-party dependency entirely and also fixes the 12 broken steps in `issues_handleLabel.yml`, not just the daily cron.

See #26806 for the replacement. Closing this in favor of that PR — appreciate the heads-up on the broken action!

_AI-assisted comment._

PR #26806, opened 2026-06-26
ci: replace issues-helper with actions/github-script

innerdvations:
### What does it do?

Removes all usage of `actions-cool/issues-helper` (TOS-blocked by GitHub) and replaces it with `actions/github-script`:

- **`issues_handleLabel.yml`** — single script step handles all label-triggered comments, closes, and locks (same templates and behavior as before).
- **`issues_dailyCron.yml`** — closes open issues labeled `status: can not reproduce` after 14 days of inactivity, with the same closing comment. Adds `workflow_dispatch` for manual testing.

### Why is it needed?

`actions-cool/issues-helper` has been blocked by GitHub since May 2025 (`Repository access blocked`), causing `issues_dailyCron.yml` and `issues_handleLabel.yml` to fail on every run. Using the official `github-script` action avoids a third-party dependency, org actions-allowlist concerns, and the maintainer's backup mirror.

Supersedes #26803.

### How to test it?

1. **Label automation** — on a test issue, apply each trigger label (`status: can not reproduce`, `flag: v3-unsupported`, etc.) and confirm the expected comment / close / lock behavior.
2. **Daily cron** — run **Daily Cron - 00:00** via `workflow_dispatch` and confirm it processes matching issues (or logs `Closed 0` when none qualify).

### Related issue(s)/PR(s)

Fixes CMS-1318

Closes approach in #26803 (backup-repo workaround).

### DIFF
--- a/.github/workflows/issues_dailyCron.yml
+++ b/.github/workflows/issues_dailyCron.yml
@@ -3,6 +3,7 @@ name: 'Daily Cron - 00:00'
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 permissions:
   issues: write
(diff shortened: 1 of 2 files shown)


nclsndr:
We need to merge to test, right @innerdvations?

innerdvations:
I believe it can be tested on a forked repo, but it's pretty complicated 🫠

cache-your-dreams:
> I believe it can be tested on a forked repo, but it's pretty complicated 🫠

Would it not still take 14 days to test?
Seeing as the current implementation already doesn't work, would there be any risk of merging this PR and testing live?

innerdvations:
Yeah, for workflows like this I'm fine with testing live, there's nothing dangerous in here

cache-your-dreams (approved):
LGTM for testing live

Issue #26795, opened 2026-06-25
TypeError: checkUserHasPermissions is not a function


joeyfigaro:
### Node Version

v25.9.0

### Package Manager

pnpm

### Package Manager Version

11.5.2

### Strapi Version

5.48.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

*No response*

### Bug Description

I've encountered this one numerous times and it seems to come and go inexplicably. Running Strapi v5.48.1. Error shows up when accessing the media library.

```shell
TypeError: checkUserHasPermissions is not a function
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-QMXOZSTW.js?v=2ae70346:85:7
    at commitHookEffectListMount (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:16915:34)
    at commitPassiveMountOnFiber (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:18156:19)
    at commitPassiveMountEffects_complete (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:18129:17)
    at commitPassiveMountEffects_begin (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:18119:15)
    at commitPassiveMountEffects (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:18109:11)
    at flushPassiveEffectsImpl (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:19490:11)
    at flushPassiveEffects (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:19447:22)
    at http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:19328:17
    at workLoop (http://localhost:1337/admin/node_modules/.strapi/vite/deps/chunk-SN2BYG2X.js?v=2ae70346:197:42)
```

(from strapi/strapi#26794)

### Steps to Reproduce

1. add Strapi to a new React vite application
2. Go through registration and skip any tours
[...]

PR #26809, opened 2026-06-26
fix(admin): restore runtime default for context helper

aun009:
## What does this do?

This change aligns the runtime behavior of the shared `createContext` helper with its existing TypeScript contract.

Previously, `ShouldThrow = true` existed only as a type-level default. Since `shouldThrowOnMissingContext` had no runtime default, omitting the third argument caused it to be `undefined`, allowing missing contexts to return `undefined` instead of throwing a descriptive error.

This change adds the missing runtime default (`true`), so consumers fail with a clear error when rendered outside their provider.

---

## Why is this needed?

While investigating Issue strapi/strapi#26795, I found that the shared context helper's runtime behavior didn't match its TypeScript API.

A compatibility audit confirmed that no callers explicitly pass the third argument, so this change preserves the intended behavior while improving error handling and preventing `undefined` from propagating to downstream code.

---

## How was this tested?

* Added regression tests covering:
  * Normal provider usage
  * Missing provider (default behavior)
  * Explicit `false` opt-out
  * Multiple context instances (missing-provider scenario)

Run the tests with:

```bash
yarn jest packages/core/admin/admin/src/components/tests/Context.test.tsx
```

---

Fixes [CPR-424](https://linear.app/strapi/issue/CPR-424/pr-26809-fixadmin-restore-runtime-default-for-context-helper)

### DIFF
--- a/packages/core/admin/admin/src/components/Context.tsx
+++ b/packages/core/admin/admin/src/components/Context.tsx
@@ -26,7 +26,7 @@ function createContext<ContextValueType extends object | null>(
   function useContext<Selected, ShouldThrow extends boolean = true>(
     consumerName: string,
     selector: (value: ContextValueType) => Selected,
-    shouldThrowOnMissingContext: ShouldThrow = true as ShouldThrow
+    shouldThrowOnMissingContext?: ShouldThrow
   ) {
     return ContextSelector.useContextSelector(Context, (ctx) => {
       // The context is available, return the selected value
--- a/packages/core/admin/admin/src/components/Context.tsx
+++ b/packages/core/admin/admin/src/components/Context.tsx
@@ -26,7 +26,7 @@ function createContext<ContextValueType extends object | null>(
   function useContext<Selected, ShouldThrow extends boolean = true>(
     consumerName: string,
     selector: (value: ContextValueType) => Selected,
-    shouldThrowOnMissingContext?: ShouldThrow
+    shouldThrowOnMissingContext: ShouldThrow = true as ShouldThrow
   ) {
     return ContextSelector.useContextSelector(Context, (ctx) => {
       // The context is available, return the selected value
--- a/packages/core/admin/admin/src/components/tests/Context.test.tsx
+++ b/packages/core/admin/admin/src/components/tests/Context.test.tsx
@@ -1,0 +1,145 @@
+import * as React from 'react';
+
+import { render, renderHook, screen } from '@testing-library/react';
+
+import { createContext } from '../Context';
+
+/**
+ * Catches render-time errors thrown by context consumers rendered outside their
+ * provider, so Jest doesn't treat the thrown error as an unhandled exception.
+ */
+class ErrorBoundary extends React.Component<{ children: React.ReactNode }, { error?: Error }> {
+  state: { error?: Error } = {};
+
+  static getDerivedStateFromError(error: Error) {
+    return { error };
+  }
+
+  render() {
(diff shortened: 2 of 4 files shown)


aun009:
Here is a quick technical walkthrough showing the unit tests passing successfully and the local application running cleanly.

https://github.com/user-attachments/assets/fc2c7e31-0714-45bc-9484-d3e65a90a13d

aun009:
Hi team! Just following up on this PR. If there are any concerns with the approach or if you'd prefer this to be split into a separate issue, I'd be happy to update it. Thanks for your time!

aun009:
Hi @innerdvations 👋

Following up on CPR-424 — this PR fixes #26795 (`TypeError: checkUserHasPermissions is not a function`), 
which causes runtime crashes when `createContext` is used outside its provider scope.

**What this fixes:**
The `shouldThrowOnMissingContext` parameter had a TypeScript-level default (`true`) but no runtime default, 
causing `undefined` to propagate downstream instead of throwing a descriptive error. This change adds the 
missing runtime default and aligns behavior with the existing type contract.

**Testing:**
- Added regression tests covering normal usage, missing-provider scenario, explicit `false` opt-out, 
  and multiple context instances
- Full test suite passes locally (see demo video above)
- All CI checks green

**Quick questions to help move this forward:**
1. Would you prefer I squash the 5 commits into 1 before review?
2. The Vercel deployment is blocked on maintainer approval — should I do anything on my end?

Happy to make any adjustments needed. Thanks for your time!

innerdvations (approved):
Hey @aun009 thanks for the fix, it looks correct to me!

Don't worry about squashing, we will squash it when we merge it to the base branch. Vercel can also be ignored.

In the future however, please don't tag Strapi maintainers by name unless it's replying to them :)

Thanks again!

aun009:
Thanks so much! I really appreciate you taking the time to review it. I'll
keep that in mind for future PRs.

innerdvations (dismissed):

innerdvations:
E2E tests all seem to fail 🤔

 I'll take a look and see if it could be from the change or if it could be a CI issue

aun009:
Thanks for taking a look! I noticed one E2E job mentions a missing organization token, though there are also some Playwright timeouts in the logs. I’ll leave the branch as it is while you investigate.

innerdvations:
It looks like setting the default to `false` is too aggressive and will break many things. I'll investigate tomorrow and what alternatives we could go with.

innerdvations:
Ok, I have a solution. Setting the default is going to be a breaking change for a lot of people (not just code to update internally, but also indirectly could affect user code that touches the admin front end) that intentionally only use two arguments and expect `undefined` outside the provider.

However, we can instead just start adding an explicit value in places that need it, such as in `useRBAC` (and `useMenu` which has the same pattern?)

That should fix #26795 without causing any problems.

Could you revert the runtime default and then add the explicit `true` where missing context should throw?

aun009:
Sounds good, reverting the default to keep the existing two-arg behavior intact, then explicitly passing true in useRBAC and useMenu so missing context throws there. Will update the tests and push.

aun009:
Pushed the follow-up commit: 494df6c.

I reverted the `createContext` runtime default to keep the existing two-argument behavior, then added explicit `true` in `useRBAC` and `useMenu` where we want missing context to throw.

Updated the tests as well — focused Context test and formatting pass locally.

innerdvations (approved):
Great, thanks!

PR #26350, opened 2026-05-16
fix(typescript): prevent stale content-type typings during develop reload cycle

AdityaOP007:
# fix(strapi): ensure TypeScript content-type typings regenerate deterministically on reload

## Summary

This PR fixes an issue where TypeScript content-type typings were not reliably regenerated after creating, editing, or deleting content types through the Content-Type Builder.

In affected cases, Strapi correctly generated the API files under `src/api/*`, but the generated typings inside:

```txt
types/generated/contentTypes.d.ts
```

could become stale or incomplete during the reload cycle.

As a result, valid generated code such as:

```ts
export default factories.createCoreController('api::post.post');
```

would fail with:

```txt
TS2345: Argument of type '"api::post.post"' is not assignable to parameter of type 'ContentType'
```

This mainly affected TypeScript projects using the `develop` command and was especially reproducible on Windows environments.

Fixes strapi/strapi#26348.

---

# Root Cause

The issue came from a race condition during the development reload flow.

When a schema changed through the Content-Type Builder:

1. The worker process wrote the schema and triggered a reload
2. The primary process immediately compiled TypeScript
3. At this moment, regenerated typings were still stale because the new worker had not regenerated them yet
4. Since incremental TypeScript compilation was enabled by default, this compilation produced a `.tsbuildinfo` file containing stale type information
5. The new worker later regenerated correct typings, but the incremental compiler could still reuse the stale cache
6. On Windows, filesystem timestamp granularity (\~1 second) made this behavior much easier to reproduce

Additionally, if the primary process compilation failed during reload, the current implementation called:

```ts
process.exit(1)
```

which terminated the entire development server before the worker could recover with fresh types.

---

# What Changed

1. Disable incremental compilation during primary reload compile

---

File:

```txt
packages/core/strapi/src/node/develop.ts
```

The primary process reload compilation now runs with:

```ts
incremental: false,
tsBuildInfoFile: null
```

This prevents stale `.tsbuildinfo` files from being generated before fresh typings exist.

The primary compile is effectively treated as a temporary bootstrap compile instead of a cache-producing incremental build.

---

2. Make reload compilation non-fatal

---

Previously, any compilation error during reload caused the primary process to terminate immediately.

This PR changes the behavior so that reload compilation errors are logged as warnings and the reload cycle continues.

The worker process is then allowed to:

* regenerate typings
* perform a clean compilation
* recover successfully

This avoids unnecessary crashes caused by transient schema/type mismatches during regeneration.

---

[...]

PR #26811, opened 2026-06-26
chore(typescript-utils): migrate to typescript

unrevised6419:
### What does it do?

Migrates the `@strapi/typescript-utils` package to TypeScript. Split into three reviewable commits so the mechanical changes are easy to separate from the actual TS conversion:

1. **`rename lib/ to src/`** — pure `git mv` to align the layout with other migrated workspaces (e.g. `@strapi/core-utils`). No content changes.
2. **`use named exports for internal modules`** — internal modules that exported a single value via `module.exports = x` switch to named exports, with their requires updated. Inline functions defined inside `module.exports = { ... }` are extracted to top-level declarations (using `function` declarations for the ones that were not arrow functions). Still CommonJS; keeps the export-shape change isolated from the TS migration.
3. **`migrate to typescript`** — convert `src/**/*.js` to `.ts` (require/`module.exports` → ESM, with type annotations); add the rollup + `tsc` build (`build:code`, `build:types`) emitting `dist/` with CJS, ESM and real `.d.ts`; add `tsconfig.json` / `tsconfig.build.json` / `tsconfig.eslint.json`; switch ESLint to `eslint-config-custom/back/typescript`; add `test:ts`; drop the hand-written `index.d.ts` any-shim and point `types` at `./dist/index.d.ts`; declare `@types/lodash`, `@types/node`, `tsconfig`, `eslint-config-custom`.

The public package API is unchanged (same runtime exports), so consumers keep working without changes.

### Why is it needed?

`@strapi/typescript-utils` was one of the remaining CommonJS-only packages shipping raw `lib/` JavaScript with a hand-written `any` type shim. Migrating it brings it in line with the rest of the monorepo (built `dist/` with real type declarations) and lets consumers type-check against accurate types instead of `any`.

### How to test it?

```bash
yarn install
yarn nx run-many -t test:ts test:unit lint build --projects=@strapi/typescript-utils
```

All of `test:ts`, `test` (99 unit tests), `lint` and `build` pass. Downstream consumers were also verified (`@strapi/admin` and `@strapi/generators` `test:ts`, `@strapi/core` and `@strapi/strapi` builds) against the newly emitted types.

### Potentially conflicting open PRs

This PR moves `packages/utils/typescript/lib/*.js` to `src/*.ts`, so any open PR that edits those `lib/` files will conflict regardless of merge order (git can't auto-merge an edit-to-`lib/foo.js` against the move-and-rewrite). Whoever merges second has to reconcile.

**Logic fixes to files migrated here** (these carry bug fixes that must be ported, not dropped):

| PR | Overlapping files | Change |
| --- | --- | --- |
| #26195 | `lib/generators/utils.js` (+ test) | Adds `ts.NodeFlags.Namespace` to the inner `createModuleDeclaration` so it emits `namespace` instead of the deprecated `module` keyword |
| #26350 | `lib/generators/utils.js` | Atomic write (temp-then-rename) in `saveDefinitionToFileSystem`; also fixes the `Date.now` → `Date.now()` timer bug |
| #26356 | `lib/utils/is-using-typescript{,-sync}.js`, `lib/utils/resolve-outdir{,-sync}.js` | Adds production `dist/config/` detection (NODE_ENV check + `path` import) |

**Config / `package.json` overlaps** (mergeable-level, not logic):

| PR | Overlapping files |
| --- | --- |
| #26701 | `package.json` (jest/nx changes) |
| #25873 (bump TypeScript to v6) | `package.json` (typescript dep), `tsconfigs/admin.json`, `tsconfigs/server.json` |
| #26699 (scope tsconfig types per workspace) | `package.json`, adds a `jsconfig.json` to the package |

**Low / no risk:** #26790 (`erasableSyntaxOnly` — base tsconfig change, no file overlap; this package uses no enums/namespaces/parameter-properties), #26760 (esm consumption — no files here), #26271 (different package).

### Related issue(s)/PR(s)

See the "Potentially conflicting open PRs" section above.


innerdvations:
The failing front-end test was on develop, it's fixed now if you want to update

innerdvations:
Thanks for splitting this into three reviewable commits — the migration approach looks solid and matches the pattern used in other migrated utils packages (`rollup` + `tsc`, `dist/` with real `.d.ts`, CJS export shape preserved).

One thing to reconcile before merge: **#26195** landed on `develop` after this branch last synced, and the TS conversion appears to have dropped its fix. On current `develop`, the inner `createModuleDeclaration` in `generateSharedExtensionDefinition` passes `ts.NodeFlags.Namespace` so emitted typings use `export namespace Public` instead of the deprecated `export module Public`. Without that flag, generated output regresses.

Related: `lib/__tests__/generators/utils.test.js` (the 3 tests added in #26195) doesn't look migrated yet — only the `schemas/*.test` files moved to `src/__tests__/`. Worth porting to `src/__tests__/generators/utils.test.ts` so the namespace emission stays guarded.

A rebase onto current `develop` should cover that. Your conflict matrix for **#26350** / **#26356** is still helpful for merge order — neither is on `develop` yet, so not a blocker for this PR today.

Happy to re-review after the rebase.

_AI-assisted comment._

unrevised6419:
@innerdvations please trigger CI again

innerdvations (dismissed):
LGTM now, thanks! I will hold off on merging until @nclsndr gives it a quick look to make sure it won't conflict with any of his plans.

unrevised6419:
@innerdvations I will push one more rebase that keeps the original commits in order to be easier to review (I dropped the merge commits)

innerdvations (approved):

nclsndr (approved):
Thank you for this migration @unrevised6419 🙌 
LGTM ✅

Verified locally: `test:ts`/`test:unit` (102 tests)/`lint`/`build` all green for `@strapi/typescript-utils`, plus `test:ts` for `@strapi/admin` and `@strapi/generators` against the newly emitted `.d.ts`. `package.json` shape matches `@strapi/logger`'s already-migrated pattern exactly, `yarn.lock` diff is additive devDeps only, `version:check` aligned.

Confirmed the #26195 namespace-emission fix (`ts.NodeFlags.Namespace`) and its 3 ported tests are present on this branch — the earlier regression concern is resolved.

---
_AI assisted_

PR #26816, opened 2026-06-26
fix(content-manager): prevent dynamic zone crash when value is null

### LINKED ISSUE #26815: v5: DynamicZone crashes with "Cannot read properties of null (reading 'length')" after closing any relation modal
### Node Version

22.x (LTS)

### Package Manager

npm

### Package Manager Version

11.6.0

### Strapi Version

5.48.1

### Operating System

MacOS

### Database

SQLite

### Javascript or Typescript

Typescript

### Reproduction URL

_No response_

### Bug Description

Closing **any relation modal** on a content type that also has a **DynamicZone field** causes a full React crash:

```
TypeError: Cannot read properties of null (reading 'length')
    at DynamicZone (Field.tsx:136:50)
```

The crash is reproducible with any relation field type (manyToOne, manyToMany, etc.) — not specific to one relation. The page becomes completely unusable until a hard reload.

### Steps to Reproduce

1. Create a collection type with at least one **DynamicZone** field and at least one **relation** field.
2. Open an existing entry in the Content Manager.
3. Click the relation field to open the relation modal (e.g. to select a related entry).
4. **Close** the modal (without saving anything).
5. → React crashes with `TypeError: Cannot read properties of null (reading 'length')` in `DynamicZone/Field.js`.

### Expected Behavior

Closing a relation modal should return to the edit view without any error. DynamicZone fields should always hold an array value, never `null`.

### Logs

```shell
React crashes with a full white screen:

TypeError: Cannot read properties of null (reading 'length')
    at DynamicZone (Field.tsx:136:50)
    at renderWithHooks (react-dom.development.js:15486:18)
    at mountIndeterminateComponent (react-dom.development.js:20103:13)
    at beginWork (react-dom.development.js:21626:16)
```

### Code Snippets

The crash is caused by a chain of three bugs:

**1. `triggerRefetchDocument` after modal close**

`RelationModal.js` always calls `triggerRefetchDocument(rootDocumentMeta, true)` when the modal closes. This refetches the parent document and updates the RTK Query cache, which changes `props.initialValues` on the parent `<Form>`.

**2. `removeNullValues` silently does nothing for null DynamicZones**

`transformDocument` (in `data.js`) applies `removeNullValues` to convert `null` → `undefined`. But `traverseData` — the utility both use — **short-circuits at line 50 before calling the predicate** when `value === null`:

```js
// data.js:50
if (BLOCK_LIST_ATTRIBUTE_KEYS.includes(key) || value === null || value === undefined) {
    acc[key] = value; // null passes through unchanged — predicate never called
    return acc;
}
```

So `removeNullValues`'s predicate `(attribute, value) => value === null && attribute.type !== 'boolean'` is never reached. DynamicZone fields returned as `null` by the API remain `null` in the transformed document.

**3. `SET_INITIAL_VALUES` resets the form with null DynamicZone values**

In `Form.js`, when `props.initialValues` changes it dispatches `SET_INITIAL_VALUES`, and the reducer replaces `state.values` entirely with the payload. Since `removeNullValues` did nothing, `state.values` now contains `{ description: null, content: null }` for DynamicZone fields.

**4. `DynamicZone/Field.js` does not guard against `null`**

```js
// Field.js:53
const { value = [], error } = strapiAdmin.useField(name);
```

The default `= []` only applies when `value` is `undefined`. When `value` is `null`, the default is skipped and `value.length` throws.

### Suggested Fixes

**Fix A — minimal, in `DynamicZone/Field.js`** (safe guard at the consumer):
```js
// Before
const { value = [], error } = strapiAdmin.useField(name);

// After
const { value: _rawDZValue = [], error } = strapiAdmin.useField(name);
const value = _rawDZValue ?? [];
```

**Fix B — root cause, in `data.js`** (make `removeNullValues` actually work for DynamicZones):

Either remove `value === null` from the early-return guard in `traverseData` and handle `null` explicitly in the `dynamiczone` branch, or add a dedicated null→`[]` coercion for DynamicZone attributes in `transformDocument`.

Fix A is the safe minimal patch; Fix B addresses the underlying regression where `removeNullValues` silently bypasses all null values.

### Media

_No response_

### Additional information

- The bug affects **all content types** that combine a DynamicZone with any relation field.
- `removeNullValues` appears to have been intended to fix this class of issue (see its JSDoc: *"We need to remove null fields from the data-structure because it will pass it to the specific inputs breaking them"*), but it has never worked due to the `traverseData` short-circuit.

### Confirmation Checklist

Jian-Zhang08:
### What does it do?

`DynamicZone` read its value with `const { value = [] } = useField(...)`, but a default only applies to `undefined`. Replaced with `rawValue ?? []` so `null` is treated as empty instead of crashing on `value.length`. Added a regression test.

_Prepared with an AI agent (Claude); couldn't run the heavy Strapi suite locally, test added for CI._

### Why is it needed?

Closing a relation modal on a content type with a DynamicZone left the value `null`, crashing the edit view with `Cannot read properties of null (reading 'length')`.

### How to test it?

Create a collection type with a DynamicZone + relation field, open an entry, open a relation modal and close it. Before: crash. After: fine. Unit test: `yarn test:front` → DynamicZone null value.

### Related issue(s)/PR(s)

Fixes #26815

---
Fixes CPR-425


innerdvations (approved):
Thanks!

PR #26817, opened 2026-06-26
enhancement(admin): augmentable registries for custom field CTB options

innerdvations:
## Summary

Alternative to #25439: instead of widening `CustomFieldOptionInput` / `CustomFieldOptionName` with `(string & {})` and `` `options.${string}` ``, this adds **module-augmentable registries** exported from `@strapi/admin/strapi-admin`.

Plugin authors register a CTB form input at runtime, then augment types so `customFields.register()` options get autocomplete and typo checking:

```ts
import type {} from '@strapi/admin/strapi-admin';

declare module '@strapi/admin/strapi-admin' {
  export interface CustomFieldOptionInputRegistry {
    'my-slider': true;
  }

  export interface CustomFieldOptionNameRegistry {
    'options.my-slider-label': true;
  }
}
```

The `import type {}` line is required so TypeScript augments the existing package instead of replacing it. Built-in option input types and root-level option names are unchanged.

## Why is it needed?

Custom field advanced/base options can reference plugin-registered CTB form components, but the current closed unions force `as any`. Registry augmentation matches Strapi's public-registry type-system pattern and keeps compile-time safety without accepting every string.

## How to test it?

- `yarn nx run @strapi/admin:test:ts:front`
- Compile-time assertions in `customFieldRegistriesTypes.ts` (imports the public admin barrel; augmentation in `customFieldRegistriesAugmentation.ts`)

## Related issue(s)/PR(s)

- Fixes CMS-1550
- Alternative to #25439


packages/core/admin/admin/src/core/apis/tests/customFieldRegistriesAugmentation.ts
nclsndr, line 1: Are we sure this does not leak to production code? 
It seems `packages/core/admin/admin/tsconfig.build.json` does not exclude `__tests__`

innerdvations: Good catch — moved the compile-time fixtures under `src/core/apis/tests/` (already excluded by `./src/**/tests`) and also added `**/__tests__` to `admin/tsconfig.build.json` exclude so nothing under `__tests__` can leak into published types.
_AI-assisted comment._

packages/core/admin/admin/src/core/apis/CustomFields.ts
```diff
@@ -13,7 +13,37 @@ type CustomFieldUID = Utils.String.Suffix<
...
+ * declare module '@strapi/admin/strapi-admin' {
+ *   export interface CustomFieldOptionNameRegistry {
+ *     'options.my-slider-label': true;
+ *   }
+ * }
+ * ```
+ */
+export interface CustomFieldOptionNameRegistry {}
+
+export type BuiltInCustomFieldOptionInput =
```
nclsndr, line 50: **This PR would publish a pre-existing runtime/type mismatch (medium)**

The spelling mismatch itself is not introduced by this PR: `BuiltInCustomFieldOptionInput` already contained `text-area-enum`, while CTB already registered the component as `textarea-enum`. Until now, however, that alias was internal rather than part of the supported `@strapi/admin/strapi-admin` surface.

This PR exports `BuiltInCustomFieldOptionInput` and uses it in the newly public, augmentable option contract. That changes the impact: plugin authors are now guided toward `text-area-enum`, which reaches the CTB custom-input lookup, misses the registered component, and renders the disabled “Not supported” input. The working runtime value, `textarea-enum`, is rejected by TypeScript. Publishing it also makes correcting the typo later more likely to be treated as a public type break.

Could we align the literal with the existing runtime registration before exposing this type publicly? This is cleanup of an existing mismatch at the point where the PR would otherwise promote it into public API, rather than a claim that the PR created the underlying typo.

---
_AI assisted_

packages/core/admin/admin/src/core/apis/CustomFields.ts
```diff
@@ -13,7 +13,37 @@ type CustomFieldUID = Utils.String.Suffix<
...
 >;
 
-type CustomFieldOptionInput =
+/**
+ * Augment with custom Content-Type Builder form input ids registered through
+ * `app.getPlugin('content-type-builder').apis.forms.components.add`.
+ *
+ * @example
+ * ```ts
+ * declare module '@strapi/admin/strapi-admin' {
```
nclsndr, line 24: **The copyable augmentation example needs to establish module context (low)**

TypeScript gives `declare module '@strapi/admin/strapi-admin'` two different meanings depending on the containing file. Inside a file that already has a top-level import/export, it augments the existing package as intended. Inside a standalone `.d.ts` containing only the documented snippet, TypeScript can instead treat it as a new ambient module declaration; the package is then seen as containing only the registry interface from the snippet, so existing named exports disappear.

This is a realistic copy/paste path for a plugin or app. Using the example verbatim in an example app produced `TS2305` for existing exports including `StrapiApp`, `CustomFieldOptionInput`, and `CustomFieldOptionName`. Adding this first line made the same program compile:

```ts
import type {} from '@strapi/admin/strapi-admin';
```

The new test fixture already uses the equivalent prerequisite (`import type {} from '../CustomFields'`), so the implementation test relies on module context that the public examples do not show. Could we include the public import marker in both examples and cover the exact `@strapi/admin/strapi-admin` copy/paste form?

---
_AI assisted_

packages/core/admin/admin/src/core/apis/tests/customFieldRegistriesAugmentation.ts
```diff
@@ -0,0 +1,13 @@
+import type {} from '../CustomFields';
```
nclsndr, line 3: **These compiler fixtures do not currently protect the public contract (medium)**

These are valid compile-time assertions: the positive assignments must compile, and `@ts-expect-error` fails if its expected error disappears. The gap is what they compile and when:

- Both files target `../CustomFields`, bypassing the public Admin barrel, emitted package declarations, and contextual typing through `StrapiApp.customFields.register()`.
- The PR's advertised `@strapi/admin:test:ts` command includes only `_internal`, so it does not compile either fixture.
- `test:ts:front` does include them locally, but the TypeScript CI job was skipped for this frontend-only PR.

In an isolated check, removing the new public re-exports still let `test:ts` and the declaration build pass, while a real package consumer rejected the augmented literals. Could we add a package-consumer fixture that augments `@strapi/admin/strapi-admin`, exercises registration through the public API, and runs in a CI target triggered by frontend changes?

---
_AI assisted_

packages/core/admin/admin/tsconfig.build.json
```diff
@@ -9,6 +9,7 @@
   "include": ["./src", "./custom.d.ts", "../shared", "../ee/admin", "../package.json", "./tests"],
   "exclude": [
     "**/__mocks__",
+    "**/__tests__",
```
nclsndr, line 12: It seems this is not gating anything for this PR right?

nclsndr (commented):
Thanks @innerdvations for putting this together. I'm still skeptical upon the overall design, especially around the deep augmentation (`'@strapi/admin/strapi-admin'`), but at the same time, I did not find a groundbreaking idea to land a better design 🤔 

---

This changes the public `@strapi/admin` type surface for custom-field option registration.

---
_AI assisted_

packages/core/admin/admin/src/core/apis/tests/customFieldRegistriesAugmentation.ts
innerdvations: Agreed that would be ideal. But there’s no existing harness that typechecks a consumer against @strapi/admin/strapi-admin → dist, so that would be new infrastructure rather than another fixture in this package.

I did retarget the compile-time assertions at the admin barrel so dropping the public re-exports fails `test:ts:front`. Let's do a proper consumer setup as a follow-up.

packages/core/admin/admin/tsconfig.build.json
innerdvations: Right — these fixtures are under `src/**/tests`, which was already excluded. `**/__tests__` is just extra so a `__tests__` folder can’t leak into published types later. Happy to drop it if you’d rather keep the diff tighter.
_AI-assisted comment._

nclsndr: Good to me to keep it as well 👌

packages/core/admin/admin/src/core/apis/CustomFields.ts
innerdvations: Fixed — the public literal is `textarea-enum` to match the CTB registration.

_AI-assisted comment._

packages/core/admin/admin/src/core/apis/tests/customFieldRegistriesAugmentation.ts
innerdvations: Fixed — both examples now start with `import type {} from '@strapi/admin/strapi-admin'`.

_AI-assisted comment._

nclsndr:
I validated this with a temporary TypeScript app inside `examples/`.

## Monorepo QA plan

1. Generate an app from this checkout and use workspace packages:

```bash
node packages/cli/create-strapi-app/bin/index.js examples/custom-field-options-qa \
  --non-interactive --skip-cloud --no-example --no-git-init \
  --use-yarn --no-install --dbclient sqlite --dbfile .tmp/data.db
```

Set generated `@strapi/*` dependencies to `workspace:*`. Also add:

```json
{
  "devDependencies": {
    "@strapi/admin": "workspace:*"
  }
}
```

Then run `yarn install` from the repository root.

2. Register the custom field server-side:

```ts
strapi.customFields.register({
  name: 'qa-rating',
  type: 'integer',
});
```

3. Add the augmentation under `src/admin/custom-field-options.d.ts`:

```ts
import type {} from '@strapi/admin/strapi-admin';

declare module '@strapi/admin/strapi-admin' {
  interface CustomFieldOptionInputRegistry {
    'qa-proof-slider': true;
  }

  interface CustomFieldOptionNameRegistry {
    'options.proof.label': true;
  }
}
```

The empty type import is required here, or an equivalent module marker. It ensures `declare module` augments the existing package instead of defining a replacement ambient module.

4. In `src/admin/app.tsx`, register both runtime pieces:

```ts
forms.components.add({
  id: 'qa-proof-slider',
  component: ProofSlider,
});

app.customFields.register({
  name: 'qa-rating',
  type: 'integer',
  // labels and components omitted
  options: {
    advanced: [
      {
        type: 'qa-proof-slider',
        name: 'options.proof.label',
        defaultValue: 41,
        // labels omitted
      },
    ],
  },
});
```

`forms` currently comes from an `unknown` plugin API, so the app must narrow it locally. The custom-field registration itself needs no cast.

5. Verify:

- `yarn build` accepts both augmented literals.
- Removing either registry key makes the corresponding literal fail type-checking.
- In CTB, add the custom field and open Advanced settings.
- The custom slider renders instead of “Not supported”.
- Change its value and save.
- Confirm the schema contains:

```json
{
  "options": {
    "proof": {
      "label": 73
    }
  }
}
```

- Restart Strapi and reopen the field. The saved value must remain.

This flow worked against the PR’s workspace packages.

---
_AI assisted_

